PR #300, opened 2016-11-27
Add dev_start.sh to let people get a development environment started quickly

hach-que:
When I was first setting up a development environment for this, and attempting to get the application to connect to the database, it was confusing that `.env.production` wasn't something that the application was loading it, but rather dictated a list of environment variables that are required for the app to run.

This adds a `dev_start.sh` script which basically sets up all the required environment variables (including generating secrets if needed), installs any dependencies, runs migrations and then starts the rails server.  As long as the developer has PostgreSQL and has run `bundle install`, this should basically work out of the box.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -6,7 +6,11 @@ default: &default
 
 development:
   <<: *default
-  database: mastodon_development
+  database: <%= ENV['DB_DEV_NAME'] || 'mastodon_development' %>
+  username: <%= ENV['DB_USER'] || 'mastodon' %>
+  password: <%= ENV['DB_PASS'] || '' %>
+  host: <%= ENV['DB_HOST'] || 'localhost' %>
+  port: <%= ENV['DB_PORT'] || 5432 %>
 
 # Warning: The database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ public/assets
 .env.*
 node_modules/
 neo4j/
+.secret.keybase
+.secret.paperclip
--- a/dev_start.sh
+++ b/dev_start.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+# Environment variables
+export REDIS_HOST=localhost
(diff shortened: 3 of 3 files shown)


config/database.yml
```diff
@@ -6,7 +6,11 @@ default: &default
 
 development:
   <<: *default
-  database: mastodon_development
+  database: <%= ENV['DB_DEV_NAME'] || 'mastodon_development' %>
+  username: <%= ENV['DB_USER'] || 'mastodon' %>
+  password: <%= ENV['DB_PASS'] || '' %>
+  host: <%= ENV['DB_HOST'] || 'localhost' %>
```
hach-que: In my case, Rails couldn't connect to PostgreSQL over a UNIX socket.  This changes it so the default is to connect over localhost; I'm not sure what the correct setting should be here to maintain the existing behaviour (while still allowing `DB_HOST` to work as an environment variable).

Gargron:
This function is fulfilled by `.env` files. If you need to customize your development environment further, do not commit those changes to the repository - everyone's got a slightly different approach.

`.env.production` is only loaded in the production environment btw (i.e. when RAILS_ENV=production). `.env` would be loaded always, `.env.development` only in development, etc. These files are not meant to be committed into the repository. [See more](https://github.com/bkeepers/dotenv)

hach-que:
Yeah, it's just confusing that the `.env` files aren't something that the app loads for you, nor does it happen on `rails server`. Having to write out this script when not using Docker isn't something that's indicated elsewhere.

(i appreciate the convention of prepending them with the name of the `RAILS_ENV`, but since they're not automatically loaded by anything, that doesn't make much difference).

Gargron:
They are loaded automatically though

hach-que:
Oh, uh.  They haven't been loading for me.  Trust me, that was the first thing I tried since that's how I assumed it was supposed to work.  Only when it wasn't working and I saw the file used as a parameter to the Docker container I figured well maybe it's only purpose was for Docker, and these files weren't something I should expect the app to load.

I guess next weekend I'll poke around to see if I can figure out why the app isn't loading these files for me.

PR #301, opened 2016-11-27
Allow Esc to close modal lightbox

### LINKED ISSUE #298: Esc to close media lightbox
Closing a lightbox on escape is a common functionality and would be a nice addition, so that one doesn't have to switch away from keyboard context to close it.

hach-que:
Fixes #298 

Had to do some wrangling to get the event to the modal container, since `onKeyDown` on the lightbox itself never fired.


Gargron:
In Modal's `componentDidMount`, add an event listener to `window` for `keyup`. In its `componentWillUnmount`, remove that listener. No need for props or anything like that

PR #304, opened 2016-11-27
activate portuguese translation

ghost:


Gargron:
But just FYI the server-side strings for PT are *completely* missing right now.

PR #369, opened 2016-12-23
Add unlisted toggle

adbelle:
This version readds the unlisted toggle to the interface with a caveat that private posts will always be unlisted. The logic to send 'private'/'unlisted'/'public' to the API call is also implemented.

Suggestions for improvement:

* Automatically move Unlisted toggle to ON when post is selected as private & disable it
OR
* Hide Unlisted toggle when post is selected as private


Gargron:
Merged manually with squashed commits, so please get new master history from upstream before making any other changes in the future.

PR #386, opened 2017-01-03
On file-drag, show a border around textarea

mitchhentges:
How it should look:
2017-01-02-215237_290x371_scrot

out

~~However, I'm still wrangling the build, so I haven't been able to test it live, yet~~ __gotcha__

* You can still drop the file anywhere on the page
* Due to a limitation with the JS drag event API, I can't make a `hover` effect when the entire `window` is dragged over with a file: `dragexit`/`dragleave`/etc don't trigger when the mouse leaves the window. However:
* The animation works peachy when a file is dragged onto the Tooter textarea. I needed to do some redux magic to tell the component when the file was dropped, due to the "native" JS event being caught and stopped at `window`


mitchhentges:
Fo real now it's working.
out

Gargron:
This is awesome! Thank you!

PR #404, opened 2017-01-05
Display native emoji on browsers which support it

mistydemeo:
I like my OS's native emoji more than EmojiOne; I'd really like to be able to display it where it's supported. This PR adds a check to see if native emoji is available (via the mojibaka npm package), and only use EmojiOne fonts if native emoji isn't available.

Some possible questions / extra features:

* What's the minimum Unicode version we want to gate EmojiOne support on? This uses 6.1, but devices with Unicode 6.1 will lack support for a lot of common emoji. Maybe that should get gated to Unicode 8 or 9?
* If the minimum version is left low, is it okay for unsupported emoji to be left to the OS to render, or would we want to do something else with them?

### DIFF
--- a/app/assets/javascripts/components/emoji.jsx
+++ b/app/assets/javascripts/components/emoji.jsx
@@ -1,9 +1,18 @@
 import emojione from 'emojione';
+import detectVersion from 'mojibaka';
 
 emojione.imageType    = 'png';
 emojione.sprites      = false;
 emojione.imagePathPNG = '/emoji/';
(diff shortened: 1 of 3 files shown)


Gargron:
I just tried this out and the problem is that EmojiOne is years better than Firefox native emoji on Linux/Windows. I wanna stick to EmojiOne especially given they're replaced in the client rather than in the DB so any different client could choose to display them differently!

PR #417, opened 2017-01-05
Twitter Cards support

ineffyble:
Add twitter:card metatag to enable Twitter Cards support 🐦 🗂


ineffyble:
Cards don't have any kind of registration these days.

There's four types of Twitter Cards: 
- Summary Card: Title, description, thumbnail, and Twitter account attribution.
- Summary Card with Large Image: Similar to a Summary Card, but with a prominently featured image.
- App Card: A Card with a direct download to a mobile app.
- Player Card: A Card to provide video/audio/media

Summary is most suitable in general. Only possible deviation would be using summary with large image, when it's a post being shared with media, but that would involve a different image size in the metadata too I'm guessing.

FYI, the reason we only set this metatag, is that Twitter will use the Open Graph/Facebook tags. It just needs this extra one to say "show a card".

Issue #122, opened 2016-11-01
Add microformats-2 classes so webmentions look right


kevinmarks:
See [indiewebify.me](http://indiewebify.me/validate-h-entry/?url=https%3A%2F%2Fmastodon.social%2Fusers%2Fkevinmarks%2Fupdates%2F23949) for details
-when I send the webmention it [just says "unknown'](http://www.kevinmarks.com/distributed-verify.html) (Bottom right) because there is no h-entry markup. 
The classnames are really close, so it should just be a few extra classes

kevinmarks:
That's a good start, but adding h-card to the user page and adding h-entry to each post would also be great.

see https://indiewebify.me/validate-h-card/?url=https%3A%2F%2Fmastodon.social%2Fusers%2FGargron%2F
- it currently gets the first @ because the user info isn't marked up

and 
https://indiewebify.me/validate-h-entry/?url=https%3A%2F%2Fmastodon.social%2Fusers%2FGargron%2Fupdates%2F64002

shows that there's no h-entry markup on the post.

Comment markup like this is even better:

https://indieweb.org/comments-presentation#How_to_markup

PR #421, opened 2017-01-06
Add Microformats2 markup

valpackett:
Continuing #122, this PR adds full microformats2 h-feed, h-card and h-entry markup. This allows e.g. reading feeds in [Woodwind](https://woodwind.xyz) and displaying reply/repost/like contexts in [Sweetroll](https://github.com/myfreeweb/sweetroll).


kevinmarks:
Some nesting mixups here:
looking at https://mastodon.social/users/kevinmarks/
```<data class='p-name' value='kevinmarks on mastodon.social'>``` isn't setting the p-name of the feed, and is wrapping the whole h-feed div in a ```<data>``` element, as ```<data>``` isn't self-closing. 

The non-repost entries don't have class="h-entry" so aren't showing up as entries, but theor proerties are being put into the h-feed instead:


```<div class='entry'>``` should be ```<div class='entry h-entry'>```

PR #425, opened 2017-01-07
Add Vagrant support for easy development environments

ineffyble:
This adds a `Vagrantfile` to the Mastodon repository, which will allow developers to spin up an easy development environment using Vagrant and VirtualBox.

I've included some simple instructions via README (and hopefully can add more in-depth ones to the documentation), but it should be as simple as: install VirtualBox and Vagrant, clone repo, run `vagrant up`, wait a few minutes, go to http://mastodon.dev , and start editing files.

It feels slightly off to be doing this when the project is using Docker, but while in theory Docker should be able to support a good development environment with live-reloading etc, I've never found a good example. Vagrant is a popular choice for dev and I've found it really easy when working on Drupal, WordPress, and other projects.

Note that the provisioning in this is a little crude, since it's shell based (instead of using a provisioning tool such as Chef, Ansible, Puppet). That said, it's all working pretty well, and I've tested on a couple of devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ public/assets
 .env.production
 node_modules/
 neo4j/
+
+# Ignore Vagrant files
+.vagrant/
--- a/README.md
+++ b/README.md
@@ -139,12 +139,11 @@ Running `vagrant up` for the first time will run provisioning, which will:
 - Create a new VirtualBox virtual machine from that image
 - Run the provisioning script (located inside the Vagrantfile), which installs the system packages, Ruby gems, and JS modules required for Mastodon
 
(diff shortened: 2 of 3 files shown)


Gargron:
Any reason why it's 14.04 and not 16.04?

ineffyble:
I ran into [some problems](https://bugs.launchpad.net/cloud-images/+bug/1569237) with the ubuntu/xenial64 box, and didn't want to use a third-party alternative. It shouldn't have a notable impact, the Ruby and gem versions are presumably much more crucial.

Gargron:
Ah, I see. Hopefully it won't have an impact on the dev environment.

PR #432, opened 2017-01-08
Show correct volume icons on video player

blackle:


Gargron:
I suppose it was confusing the other way around?

blackle:
I think this is the canonical way, for example vine shows the icon as muted if it is currently muted

PR #437, opened 2017-01-08
Simplifying followers mappings

krainboltgreene:
Deduplicating some logic and switching to pluck, to avoid pulling the entire model in memory.


Gargron:
I'm all :+1: for DRYing up that code, but is there a reason you opted for `inject` instead of `Array#to_h`? I like how easy to understand the latter is.

krainboltgreene:
No need to retain that in-memory array of arrays, imo.

PR #438, opened 2017-01-08
Now we can use the new name in queries

krainboltgreene:


Gargron:
Ah, this isn't very obvious, but these methods (target, verb, etc) are implementations of `Streamable` and are used to generate Atom XML stuff. They don't need to be used in queries. Ideally I think I need to rewrite `Streamable` as a sort of DSL-like definition à la "belongs_to" etc, and avoid method overrides to not confuse other people.

krainboltgreene:
Gotcha.

PR #445, opened 2017-01-08
Switch to compose view when tapping 'mention' in dropdown on mobile

blackle:
puts the layout check in an `isMobile` function so we can use the idiom `isMobile(window.innerWidth)` elsewhere in the code to do mobile-only logic


blackle:
it looks like the travis check is hanging on this?

Gargron:
Can you move the `isMobile` check out of `/components` since it's not a component? Maybe a `utils.js`

blackle:
sure, which folder should that be in?

Gargron:
Top level where `emoji.jsx` lies

PR #457, opened 2017-01-11
Store JSON in HTML attribute to avoid <script> embedding problems

minitech:
JSON isn’t quite JavaScript-in-HTML, so a user setting a display name of `</script><script>alert(1)//` will see an alert on their dashboard. (Only on their own dashboard, though, so it isn’t really a vulnerability.) A more likely scenario is that invisible U+2028s or U+2029s might sneak in at some point and cause a mysterious blank page.


Gargron:
I decided to opt for the more simple way. I verified that it solved the issue. Sorry to reject your PR!

Issue #239, opened 2016-11-23
proper content warnings and more extensive sensitive content


hoodiek:
this is not saying that the current stuff is inadequate, it is merely stating that much more could be done. for example, talk of death, and suicide, can very easily trigger such people who might be wary of them in the first place. as a result, not only should images be able to be made able to be spoilered. i would suggest that if you click the sensitive content toggle, a text box could appear with a simple content warning of sorts. for example "CW food" can be placed in a textbox, and the first few words of the post would have that added, removing from the max character limit. this would make the whole toot act like the current sensitive image filter does, and better allow sensitive toots in general. thanks for listening.

EDIT: a suggestion to avoid the CW necessity is, sensitive content tweets would mute all except #hashtags and thus be able to show stuff selectively

Gargron:
I know some people might have an issue with TW/CW so let me just add that it'd also be perfect for hiding spoilers, something Twitter/FB have never done. I'll see if I can do something about this

hoodiek:
yep! i'd really appreciate this addition! thank you so much for caring and listening, @Gargron

Xkeeper0:
Perhaps having the "Sensitive content" box allow for a short free-form description of what's sensitive about it, while possibly marking the main text invisible instead of just the picture.

marrus-sh:
Was just thinking about this and definitely favor having a separate content description textbox over just muting everything except hashtags. Take the example post

> omg there was a giant tarantula in my bathtub today `#spiders`

By clicking on `#spiders`, which the author is required to include in order to indicate the manner of sensitive content, users can then view all spiders-related tweets in one place. While this isn't probably a big deal with arachnids, it would make it incredibly easy for people to target, say, abuse survivors simply by searching the relevant tags.

**TL:DR;** using something designed as a search feature for a sensitive content marker might not be the best idea? Making it a separate piece of data would probably be better.

NireBryce:
A thing i would like is a second type of mute, where instead of ignoring the tweet, it just spoilers it as if it was sensitive content

so I could say, block anything containing "CW: spiders" and it would just spoiler everything but the keyword, or maybe anything on the same line

hoodiek:
i know this thread is very old but i think it is a strong tool to give users. i think the best middle-ground would be a click to show/hide sensitive content, and it being expanded to work like this:

sensitive content toggle would be visible for all posts, and anything in a (cw[...]) type tag would be visible. clicking on the parenthesesed thing would allow you to hide or show the content

[...]

PR #460, opened 2017-01-12
Implement a click-to-view spoiler system

blackle:
this is to make spoilers/content warnings possible, and serves as a language independent solution to #239 and a replacement to rot13

here's a video of it in action https://www.youtube.com/watch?v=jQOapNNn8F0&

this might need some more testing, I haven't tested federation but I made it so it adds a "#spoiler" tag like sensitive content get the "#nsfw" tag, and consumes feeds with the same convention

I still don't know if this is the "right" solution to #239, so it might need some public vote? even if this doesn't get accepted it's still a good development exercise for me, so no skin off my back


blackle:
I've updated it quite a bit, you can see the changes here: https://www.youtube.com/watch?v=0Y3T6Xz5h7o

it's nearly finished, I just need to test how it federates. I also want to add a timeline filter so you can filter by content warning'd toots

Gargron:
This is *exactly* how I envisaged it! Or mostly, at least. Forgive for the delay in merging. I want to finish up my site settings feature first. As soon as I'm done with that I wanna merge this and polish it up a little. Like, the content warning input could have the same style as the regex filter input in home column settings, that would probably be less visually distracting. Also, I'm gonna see how you handled federation but I think it would be sensible to use an extra tag for the extra content, that way the normal ´<content />´ would contain the content warning (it's the tag that every OStatus implementation supports atm) and the actual content would be in something like `<extendedContent />` (or a different name, not sure). That way it would work like progressive enhancement and older platforms wouldn't see disturbing content accidentally. Make sense?

Gargron:
Alright, skimmed through changed files. The user facing part is, as I said, almost on point. Just gotta clean up the storage and federation. I think it would be worth it to add an extra column for the content warning rather than re-use `text`

blackle:
Thanks so much!

IMO It might be good to keep the style of the content warning box similar to the compose box, since it's a required field when visible and also will become part of the status itself

I'll likely push up a "Show posts with content warnings" filter for the timeline tonight, so don't merge before then

on the actual implementation side it works like described in the first description of this PR, the first line of the status is the CW and the rest is covered by a click-to-show. when it renders there's a function wrap_spoilers in formatter that basically adds the spoiler class to every line but the first, and that's how it knows to render it

blackle:
hmm yeah, maybe an extra text field for the spoiler could be better. I can put that in tonight if you'd like

blackle:
I added a "spoiler_text" column and used that instead of "text"

it's about functional now, I think it would be worth merging so you can look at how it federates

I'll make the "Show statuses with content warnings" filter a separate PR when I get to it.

I was also wondering what your opinion on putting the content warning box above the compose box would be, as that would make it clear to the user how it's formatted after posting (warning above content)

blackle:
What's the status on this, does it need more work before merging or are you waiting for a good time to merge and work on it yourself?

PR #461, opened 2017-01-12
fix Vagrantfile so that the correct ruby is installed

vjackson725:
Prior to this commit, the vagrantfile installs ruby 2.3.3. Due to commit tootsuite/mastodon@7951e7f, this no longer works.

Change the provision script so that 2.3.1 is installed via rbenv. Note this drastically increases provision time.


Gargron:
> Note this drastically increases provision time.

Sorry about that! The version was already pinned previously in `.ruby-version`, I had to add it to the Gemfile as well to facilitate Heroku deployments. I think there's barely a difference between 2.3.1 and 2.3.3 but just to be sure there are no compatibility issues I recommend using the exact same version.

ineffyble:
Cool, I originally went down the rbenv route but it seemed difficult. I might look at moving to rvm as I believe that may use pre compiled binaries in some cases.

PR #464, opened 2017-01-14
Make boost button spin around

blackle:
like so: https://mastodon.social/web/statuses/625716


Gargron:
I don't wanna be a killjoy but do we wanna have two wildly different animations next to each other? I need some convincing on this. What would be cool is if the retweet icon would have the two arrows move around, it'd be in the same direction as the star spin so it would look consistent and also hella cool, but I realize it's impossible since the retweet icon is a solid font symbol :(

blackle:
having the arrows move along the square would be really cool. I think in order to do that we'd have to do something like how twitter does it: iirc they have their like button animate by having a div with a really tall png with all the frames on it as the background and animate the background position attribute so each frame becomes visible as the position changes.

I think it would actually be really cool to do that. I'll try some things out today and show you them

blackle:
ok, it looks like this now: https://mastodon.social/web/statuses/631039
the real downside is that because the animation is a PNG, it doesn't scale very well if the user zooms in, compared to font awesome glyphs.

Gargron:
Not too concerned about zoom issues because in my experience zoom messes up lots of things *anyway*. This is awesome. The remounting is unfortunate... Should I change it so that the boost appears on top but doesn't replace the original? That being said, we did have it that way at the start and people complained about seeing the same thing multiple times in the feed.

blackle:
hmm, maybe if it remounted after a delay for the transition to finish? that sounds like it would be hard

Gargron:
Yeeaaaaaah, things like that would be hard because the DOM is always a reflection of state. Although Motion does have transition animations... But I haven't mastered that part of that library yet.

PR #472, opened 2017-01-14
Add source/app metadata to toots

### LINKED ISSUE #11: Track OAuth source in statuses
To help the 3rd party app ecosystem, add "posted from [Application Name]" type information to statuses.

ineffyble:
Add tracking of OAuth app that posted a status, extend OAuth apps to have optional website field, add application details to API, show application name and website on detailed status views. Resolves #11.


ineffyble:
This is still failing CI, as the dummy OAuth token created for tests doesn't have an application set.

I tried to modify the tests to fabricate a token, but my lack of experience with RSpec showed through, and somehow made things far worse.

app/views/api/v1/apps/show.rabl
```diff
@@ -0,0 +1,3 @@
+object @application
+
+attributes :id, :name, :website
```
Gargron: Do we need to expose the id here? (I don't think so). With the requirement of apps self-registering on different instances, there's probably little value to tracking numeric IDs, since two users using the same app from their different phones with the same instance will have different app credentials in the db.

Btw if anyone has ideas on how to improve that authorization flow, I'm all ears. Had to invent it since GNU social doesn't even try to do OAuth properly, they just use password grants/basic auth, and I haven't heard of any other federated networks with OAuth APIs.

db/schema.rb
```diff
@@ -153,6 +153,7 @@
     t.datetime "created_at"
     t.datetime "updated_at"
     t.boolean  "superapp",     default: false, null: false
+    t.string   "website"
```
Gargron: Can we use "url" for consistency across different models? Alternatively: "website_url"

ineffyble: I originally didn't go with URL because I didn't want it muddled with the redirect uri/url. If that's not likely, then sure, if not, website_url works.

app/views/api/v1/apps/show.rabl
ineffyble: I included ID as a default, but you're right, very little point at moment.

Biggest problem with authorization flow is just lack of documentation, both from Mastodon and Doorkeeper.

PR #478, opened 2017-01-15
Change boost sprite to look like fa-retweet

blackle:
looks like this now: https://i.imgur.com/TOt7jpe.png


blackle:
quick change: made it smaller to be the same width as FA https://i.imgur.com/ersFxpG.png

PR #489, opened 2017-01-17
Add Heroku deployment support

ineffyble:
This patch should add everything needed for (relatively) simple deployment to a free Heroku account.

Changes:
* Add a Procfile (used by Heroku and a few other services to tell it what it should be running).
* Add a Heroku button and instructions to README.
* Add an app.json file (used by Heroku Button and Heroku to build and deploy).
* Add support for a REDIS_PASSWORD environment variable.
* Look for a REDIS_URL environment variable - if found, set the component variables (host, port, password). This is required because Heroku's Redis service will set this.
* Add optional SMTP_DOMAIN variable (to facilitate testing with a trial Mailgun or other hosted service account, which often require sandbox domains)
* Detect HEROKU environment variable, and have Puma spawn a sidekiq process if found - this allows everything we need to operate in a single 'dyno', to avoid paid money.

This _shouldn't_ break anything, but there are a few changes. I have in the past run rspec etc, but will be good to see what CI says. Will also try and build a Vagrant environment from this version to test.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: bundle exec puma -C config/puma.rb
--- a/README.md
+++ b/README.md
@@ -118,6 +118,19 @@ Which will re-create the updated containers, leaving databases and data as is. D
 
 Docker is great for quickly trying out software, but it has its drawbacks too. If you prefer to run Mastodon without using Docker, refer to the [production guide](https://github.com/tootsuite/mastodon/wiki/Production-guide) for examples, configuration and instructions.
 
+## Deployment on Heroku (experimental)
+
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
+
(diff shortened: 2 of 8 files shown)


Procfile
```diff
@@ -0,0 +1 @@
+web: bundle exec puma -C config/puma.rb
```
Gargron: While I like the idea of saving money by running everything in one dyno, shouldn't we still define the correct jobs in the Procfile so people can scale the dynos dynamically from the Heroku dashboard/CLI?

ineffyble: Hmm, that could work. We'd want to rename the variable from HEROKU to HEROKU_SINGLE_DYNO or some such. Only thing I'm not sure of: will Heroku by default provision 1 web and 1 worker (which will exceed free capacity during the month)? I'll do some tests.

ineffyble:
Hey, CI passed!

Issue #423, opened 2017-01-07
user blocking/whitelisting of entire domains


deckycoss:
i have a proposal for an interface that would allow individual users to control which GNU social / mastodon instances they would see posts from on the public timeline.

the interface would be a checklist, on the user settings page, of all GNU social instances that the current instance knows about. instances that are checked are whitelisted; everything else is blocked. there could be buttons for things like sorting the list and checking/unchecking all entries at once. there should also be a toggle for whether or not to automatically whitelist new instances.

it seems there's currently no way to find a list of all existing instances of GNU social or even just mastodon. however, as soon as you view an account that belongs to another instance, a reference to that account (and thus the instance it's on) is added to the current instance's database. so, populating the checklist shouldn't be hard.

if people like the sound of this then i would be willing to do the work on it. i finally have my own dev instance working outside of docker and while i'm new to rails, i feel capable of adding this feature.

see issues #231 and #344 for previous discussion on similar topics.

krainboltgreene:
Would a "host file" like entry be a good first step?

Gargron:
What do you mean by host file-like entry?

krainboltgreene:
@Gargron:

``` html
<textarea />
```

Gargron:
Oh! Hm. No I actually really like the idea of listing all known domains next to toggles. Beyond moderation it's also a pretty good tool for people to learn more about the fediverse. For reference you'd get the domains with a `Account.remote.select('distinct domain')`. I don't think the overhead of querying that and displaying a table with check boxes is that much worse than a textarea entry!

krainboltgreene:
Got it! I'll work on this tonight if I get the ✅  from you.

Gargron:
Alright, @deckycoss expressed interest in doing it though so just make sure you don't do the same thing twice. Also, it'll be easier to make a static version in the settings area first rather than implementing it as REST API + React.js stuff, that can be refactored into later.

deckycoss:
actually, as i said in the first post, i would like to work on this. but if i need help from people more experienced with front-end then i'll say so

krainboltgreene:
Sweet! Let me know.

ineffyble:
Related: https://github.com/Gargron/mastodon/pull/302

deckycoss:
@ineffyble oh! i didn't know anyone had already started work on this. do you plan to resume the patch? if not, maybe i could continue where you left off?

ineffyble:
[...]

PR #501, opened 2017-01-19
user blocking of domains

daliacoss:
this is a patch for issue #423. it's still in the early stages as it only implements the AccountDomainBlock model, and filtering out toots by blocked domains from the user's public timeline. the finished patch will update the API and provide an interface for the user to create domain blocks.

this builds off of @hach-que 's unfinished patch #302.


app/controllers/api/v1/accounts_controller.rb
```diff
@@ -85,7 +85,7 @@ def unblock
   def relationships
     ids = params[:id].is_a?(Enumerable) ? params[:id].map(&:to_i) : [params[:id].to_i]
 
-    @accounts    = Account.where(id: ids).select('id')
+    @accounts    = Account.where(id: ids).select('id, domain')
```
Gargron: Dunno what you're doing here tho!

deckycoss: i think i was testing something that i didn't quite understand from @hach-que 's closed patch, and forgot to revert it

app/controllers/api/v1/domains_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    # helper object as there is currently no Domain model
+
+    attr_accessor :name, :current_account
+    
+    def initialize(name, current_account)
+        @name = name
+        @current_account = current_account
+    end
+    
+    def blocked
```
Gargron: If it returns a boolean, the convention is to end the method name with `?`

deckycoss:
@Gargron see my last few messages on masto

deckycoss:
@Gargron i think i finally managed to merge from upstream without breaking anything. this patch is done until i receive any feedback suggesting otherwise.

by the way, i never figured out why the tests originally passed "blocking?" an array, since whenever i do this the result is always false, even if the array matches the set of accounts that are being blocked. i was able to make the test failures go away by adding `return false unless other_account.is_a?(Account)` to my "blocking_account_or_domain?" method, thus preventing arrays from being used at all, but that felt like a weird solution so i haven't pushed it yet.

tobascodagama:
Is there a reason this was closed?

staticsafe:
This issue is being tackled in issue #2381. You can follow that one for updates.

PR #506, opened 2017-01-21
Add a default missing.png header image - fixes #492

### LINKED ISSUE #492: Missing header image is... missing
I'm seeing a lot of 404s, attempting to load /headers/original/missing.png, which looks like the URL that's provided for users without a header image.

Guessing we need something similar to /avatars/original/missing.png

ineffyble:
Fixes #492 by adding a default header image, inspired by the cute "Getting Started" mascot on /web.

Action shots:

screen shot 2017-01-22 at 5 17 58 am
screen shot 2017-01-22 at 5 17 51 am


Gargron:
If you just make it a transparent pixel I'll merge it in immediately (I don't think this image looks good)

ineffyble:
Changed now.

PR #513, opened 2017-01-23
Textarea: fix clicking on name suggestions

mistydemeo:
This was broken in c3e9ba6a, which added an onBlur to hide the suggestions when the textarea loses focus. Unfortunately, this fired even when the textarea lost focus to its own suggestions box, and that meant that onSuggestionClick is never called.

Setting a short window in which onSuggestionClick can still happen before hiding the suggestion area ensures that clicking still works, while still hiding the suggestions if the user clicks on something else.


mistydemeo:
CI failure looks unrelated to my branch.

PR #516, opened 2017-01-24
Make boost animation an SVG that uses SCSS colours

blackle:
had to make another variable because we don't have this colour yet


app/assets/stylesheets/variables.scss
```diff
@@ -6,3 +6,4 @@ $color5: #fff; // white
 $color6: #df405a; // error red
 $color7: #79bd9a; // succ green
 $color8: #000; // black
+$color9: #606984; // buttons
```
Gargron: This is `lighten($color1, 26%)`

Gargron:
I couldn't see an icon at all (Firefox 50.1) so not merging yet. I think that putting the SVG into JSX in its own component and doing animation via JS, and coloring via fill/stroke might be better

blackle:
that's strange because svg backgrounds should be supported (see http://caniuse.com/#feat=svg-css), I'll take a look into that. If I can't fix it I'll turn it into a JSX component

blackle:
Ok, fixed it, turned out I needed to escape the colours because firefox doesn't like hashes in data URIs 
got the friendly url function from http://stackoverflow.com/a/25478589/1073287

PR #520, opened 2017-01-24
Fix `invisible` CSS class

mistydemeo:
Commit 80cefd5b added the `invisible` attribute to parts of the URL by setting the `font-size` and `line-height` to 0; this doesn't actually make the space empty in all browsers, however. In some browsers, like Safari, the text will be rendered invisible but some space will still be taken up. `display: none` has the expected effect in all browsers that I've tested.

Before:
Before

After:
After

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -61,8 +61,7 @@
 }
 
 .invisible {
-  font-size: 0;
-  line-height: 0;
+  display: none;
 }
 
 .ellipsis {


Gargron:
`display: none` makes the text unselectable, which defeats the point of doing it this way in the first place. Sorry.

mistydemeo:
Can we find another solution for this? The current situation isn't really viable for Safari, since it doesn't look right.

Gargron:
The solution is inspired by how web Twitter does it. Did I miss something about that? I only use Firefox so I could miss a detail that's important for Chrome/Safari

blackle:
does adding "display: inline-block; width: 0px; height: 0px;" to the invisible class solve the problem? that might be an OK workaround?

mistydemeo:
Web Twitter may have other styles that are being inherited from elsewhere. This works in web Twitter on WekKit browsers, but not on Mastodon.

Adding two extra styles does fix this in all browsers, so I updated the commit.

mistydemeo:
By the way, I appreciate that you only have the resources to test in Firefox, but instantly closing the PR instead of suggesting changes (say, via the "PR review" feature) is sending me a signal to not bother trying to fix this.

Gargron:
I'm sorry about the miscommunication. There's two things here: the bug and the solution. I acknowledge the bug, and was simply unhappy with the proposed solution (hence PR close). If you are going to use the same PR for a different solution, I'll reopen ofc.

mistydemeo:
I've already posted changes to the branch this PR was from!

PR #533, opened 2017-01-26
Increase color contrast on buttons

yiskah:
Currently, the buttons on the login page fail WCAG color contrast standards for text. Someone with low vision may not be able to read these buttons. This change slightly darkens the blue so that it passes AA WCAG for Large Text (the most minimal amount you need to do to not be failing accessibility).


Gargron:
I solved this in 1c507ff17205e825e5f0b91fb44f4c4acdef8d2d but forgot to close this PR. Thank you for your time anyway!

PR #546, opened 2017-01-26
Swap out Roboto font for Droid Sans

PurpleBooth:
Roboto is a lovely font, however it doesn't have a complete set of
symbols. This means that things like Japanese characters, or unicode
symbols frequently don't display.

This change will swap out Robot for Droid Sans, which is a similar
looking font under the same license that does have a full character set.

You can compare the two here:
https://fonts.google.com/specimen/Droid+Sans
https://fonts.google.com/specimen/Roboto

This relates to issue tootsuite/mastodon#531

### DIFF
--- a/app/assets/stylesheets/about.scss
+++ b/app/assets/stylesheets/about.scss
@@ -12,7 +12,7 @@
   }
 
   h1 {
-    font: 46px/52px -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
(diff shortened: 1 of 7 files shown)


ineffyble:
Why was this closed?

PurpleBooth:
I closed this because the version of Droid Sans provided by google is also a subset of the complete character set.

Gargron:
I'll like to note that when I opened the linked toot in Firefox, Windows, I was able to see the glyphs.

PurpleBooth:
Yeah, the problem resides in Chrome and the usage of the local() call.

PR #570, opened 2017-01-28
Add missing back button to "Follow Requests"

PurpleBooth:
Currently we have the problem where the follow request button is missing
from the Follow request tab. This means that when you visit that tab you
either have to press the back button in the browser or re-enter the home
page, which breaks the app flow.

This will resolve that issue by adding a new back button to that tab,
in the same way that the public timeline, and favourites have.

Previously looked like this:
image

Now looks like this:
image


Gargron (changes requested):
The ColumnBackButton is not different to the one defined in components/column_back_button.jsx, please import that one instead and remove the copy.

PurpleBooth:
This should be resolved now.

PurpleBooth:
I've made this change across the board in #577

PR #572, opened 2017-01-29
Add 2-FA instructions for non-technical users

koboldunderlord:
Re-opened request with a tweak suggested by Effy


Gargron:
Can you move the file to docs/Using-Mastodon?

PR #573, opened 2017-01-29
Fix avatar scaling on high-DPI displays

mistydemeo:
Commit 00fa850bdc8e98dfb3e3772ec2b9a55e8538cadc updated the avatars to only play when hovered. This has led to a few display bugs with avatars.

The larger bug is that the width and height being set don't take into account that the resolution in CSS pixels and physical pixels might not align; for example, on the retina/high-DPI screens used on most modern laptops, phones, and tablets, these two things differ by a factor of 2 or more. The [`window.devicePixelRatio`](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio) property provides a way to request the ratio to convert between them. The result of this was that the canvas element was being told to render in "CSS pixels", which is significantly lower than the actual size on screen; this led to avatars being rendered very pixelated on high-DPI devices.

This PR updates the styles so that the canvas element renders at the real physical resolution using `window.devicePixelRatio`, while the size in CSS is set using CSS pixels.

Before:

Pixelated avatar

After:

Clean avatar

This leaves a couple of bugs that still need to get addressed after this:

1) Because the browser's native scaling is different than what we're doing here, the avatars will still look different after you hover over them. This isn't desirable. It's probably better to use a different method of handling this, and maybe to avoid invoking this logic on non-animated avatars.
2) On devices without hover, e.g. touch devices, there's no way to get the "hover" version of an avatar.

### DIFF
--- a/app/assets/javascripts/components/components/avatar.jsx
+++ b/app/assets/javascripts/components/components/avatar.jsx
@@ -25,7 +25,7 @@ const Avatar = React.createClass({
   },
 
   handleLoad () {
-    this.canvas.getContext('2d').drawImage(this.image, 0, 0, this.props.size, this.props.size);
+    this.canvas.getContext('2d').drawImage(this.image, 0, 0, this.props.size * window.devicePixelRatio, this.props.size * window.devicePixelRatio);
   },
 
   setImageRef (c) {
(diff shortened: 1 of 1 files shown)


blackle:
as a side note I looked into making it so that only avatars that end with the extension ".gif" get frozen, but it turns out that a user could get around this by uploading a gif file that has the extension ".jpg", and the server will accept it and save it with the .jpg extension

ineffyble:
I wonder if the best way forward is just to analyse avatars on upload, detect animation, create a static version if found, and have the API return an additional `animated_avatar` key if available? Then clients can choose what to load and render where.

mistydemeo:
> ...and have the API return an additional animated_avatar key if available?

I was thinking about that too! I think that's a good idea.

PR #577, opened 2017-01-29
Use same back component for all timelines

PurpleBooth:
Currently we're un-intuitively using the back button from the public
timeline. This is a duplicate of the component we have in the root
components directory.

This change will use the same back component for each of the timelines,
and change that component to be the global one.


Gargron:
Since there was a naming confusion between two different components, the global ColumnBackButton (wide, used instead of ColumnHeader) and the one defined as a public timeline component (slim, takes up part of the column header), I've resolved this issue myself when merging your other PR, so I will close this one.

PurpleBooth:
👍

Issue #462, opened 2017-01-13
Attaching an image, then video works, but not the opposite order


Vavassor:
So the normal behaviour on the site: when you attach a video to a toot, it greys out and disables the button. But if you attach 1-3 images first and then a video, it posts successfully. Here's [an example toot](https://mastodon.social/users/Vavassor/updates/710280) which shows 2 images, and a blank thumbnail that, when you click on it, shows the video (of a scorpion).

Gargron:
Good find!

PR #589, opened 2017-02-02
Fix for issue #462

rmhasan:
Modified `uploadCompose` action to send media ids of attached
media when sending a request. Modified `create` method in `MediaController`
to check if when posting a video, there are no other media attached
to the status by looking at the media ids sent from the `uploadCompose`
action. This is a fix for #462.

### DIFF
--- a/app/assets/javascripts/components/actions/compose.jsx
+++ b/app/assets/javascripts/components/actions/compose.jsx
@@ -119,7 +119,10 @@ export function uploadCompose(files) {
 
     let data = new FormData();
     data.append('file', files[0]);
-
+    data.append('media_ids', getState().getIn(
+      ['compose', 'media_attachments']
+    ).map(item => item.get('id')));
+    
     api(getState).post('/api/v1/media', data, {
       onUploadProgress: function (e) {
         dispatch(uploadComposeProgress(e.loaded, e.total));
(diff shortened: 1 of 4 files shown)


Gargron:
Hey! Thank you for taking the time to investigate this issue. I am not sure if this approach is ultimately correct - client apps could simply choose not to send other media IDs along. If it can be faked then it might as well be a client-side check (in so far as it would be possible to check the mime type of the selected file).

The final validation happens here: https://github.com/tootsuite/mastodon/blob/master/app/services/post_status_service.rb#L39 This is where it could really be fixed, as currently the only thing it does is ensure at most 4 media attachments.

You also don't need to worry about instantly deleting uploaded attachments. There is a cronjob that peridocally purges unattached media.

rmhasan:
Hi @Gargron, good point on client apps choosing not to send media ids. I moved the validation check  to `post_status_service.rb`. Please take a look at my latest commit.

Gargron:
Okay, final suggestion: `attach_media` happens *after* the status is created, so instead of raising an error I suggest silently discarding invalid media, i.e. if a video is present, discard the other media items.

rmhasan:
Hi @Gargron, In my last commit I moved the validation check before creating the status. What do you think about that? Or do you want me to implement what you suggested?

rmhasan:
Hi @gargon, I added a new commit to this PR, to use the exception handlers in api_controller.rb

PR #603, opened 2017-02-06
Expose ActivityStreams 2.0 representation of accounts

evanminto:
This is the first step in ActivityPub integration. Clients may now request the user's account URL with an Accept header set to "application/activity+json" to get the AS2 representation. This JSON object will later point to a set of ActivityPub AS2 Collections (Outbox, Inbox, etc.).

I set up some RABL base templates for future use, to consolidate some properties (@context, id) and reduce boilerplate.

I used "image" for the header image, but I'm not sure that's the best approach. Would be nice to ensure clients have a way of getting a header image, and "image" seems like the only good option within AS2/AP.

Technically the endpoint should be able to respond to 'application/ld+json; profile="https://www.w3.org/ns/activitystreams#"' in the Accept header, but Rails seems to have trouble recognizing Mime + Profile types, so I stuck with "application/activity+json".


app/views/accounts/show.rabl
```diff
@@ -0,0 +1,8 @@
+extends 'activitypub/types/person.rabl'
```
Gargron: It could be helpful to name this files `show.activitystreams2.rabl` because e.g. I would be confused what they're used for otherwise (name of file dot format dot processing language)

evanminto: I was thinking about doing this, so yeah, no problem.

PR #615, opened 2017-02-09
Added awoo.space

nevillepark:


yiskah:
It would appear that http://on.vu is now defunct, as the page does not load. Socially.constructed.space is a single-user instance, so it seems strange to have on this list. http://gay.crime.team is another instance, which I'm unsure if registration for it are open. It would make sense to include it if we are going to include all known instances, regardless of if they have open registration. (In which case, making a table with additional information would be productive.)

Sorry to glom this onto your PR. I don't want to create a conflicting PR with a grid of information w/o talking about it.

Gargron:
Btw I'm totally up for changing that page from a list to a table with more details.

PR #633, opened 2017-02-14
Add mastobot.club

yiskah:
Added mastobot.club to instance list


mitchellurgero:
That is not an instance?

Gargron:
Indeed, this is not an instance. Feel free to put it under apps, however (maybe separate section for app-websites)

yiskah:
Hmmm, interesting conundrum.

It is a server domain, right? Bots run off of it? And it's an application. But it's not an interface really... it's very gray. regardless it seems like it doesn't go here.

Gargron:
I believe that bots sign in into other accounts from this one using OAuth

PR #645, opened 2017-02-18
Fix #624 - Add localization for Bulgarian

### LINKED ISSUE #624: Add localization for Bulgarian
I open this issue and volunteer to work on it.

Is it enough to just submit a PR with the translations when ready?

svetlik:
Fix https://github.com/tootsuite/mastodon/issues/624

Preparing for translation: 
- Add translation files for Bulgarian
- Declare language

Will keep the commits coming.


svetlik:
Should this particular codeclimate error be ignored? It seems to detect code that is "similar", which I suspect is because of localization keys, which are normally the same for all locales.

svetlik:
@Gargron , @ineffyble , this one is finally happily awaiting review. Could any of you please do the honours? Or could you please include someone that could? Thank you!

Gargron (approved):

Issue #626, opened 2017-02-12
TOOT's do not get removed from "home" timeline & public timeline until after page refresh


mitchellurgero:
TOOT's do not get removed from "home" timeline & public timeline after deleting until after page refresh:
Replicate:
With account 1 while account 2 is logged in on another browser:
1. Post toot
2. Wait about 1 minute
3. Delete toot

With account 2:
1. View toot before deleted
2. go back to public timeline
3. try to view after delete and get a 404 error, but can still see the toot.

rmhasan:
Hi @Gargron and @mitchellurgero  PR #646 fixes this issue. Can someone please review it.

PR #646, opened 2017-02-19
Fix for #626

rmhasan:
If `fetchStatus` gets an error while sending sending request to `/api/v1/statuses/${id}/context`, then it will check if the error status is 404 and delete the status from the timeline if it is.  I created a new React class called `StatusNotFound`. In `status/indiex.jsx`, if status is null, then it will render the new React class with a back button on top.

### DIFF
--- a/app/assets/javascripts/components/actions/statuses.jsx
+++ b/app/assets/javascripts/components/actions/statuses.jsx
@@ -103,8 +103,12 @@ export function fetchContext(id) {
       dispatch(fetchContextSuccess(id, response.data.ancestors, response.data.descendants));
       dispatch(fetchStatusCard(id));
     }).catch(error => {
-      dispatch(deleteStatusSuccess(id));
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/actions/statuses.jsx
```diff
@@ -28,7 +28,6 @@ export function fetchStatus(id) {
     const skipLoading = getState().getIn(['statuses', id], null) !== null;
 
     dispatch(fetchContext(id));
-    dispatch(fetchStatusCard(id));
```
Gargron: Why have you decided to move this from here to the other callback?

rmhasan: @Gargron If a status was deleted in the database, but existed on a frontend and someone clicked on it, then you would get 2 error flashes, when you should be getting 1.

PR #652, opened 2017-02-25
Splitting frontend/backend??

marrus-sh:
Yea so this has been my working setup for about a week or so now and figured I'd submit a PR incase this is functionality desired upstream. Adds two environment variables (`SKIN` and `FRONTEND`) and packages the default masto skin/frontend into `/app/assets/skins/rooty/` and `/app/assets/frontends/tooty/` (feel free to change). See the file I added to `docs` for more info! The only tricky bit was splitting up the CSS files but I think I did alright.


yiskah:
Into This Idea

ProgVal:
merge conflict :/

Gargron:
I'm sorry it took so long to actually come out with a response (a lot of stuff got in the way and this wasn't exactly high prio). Basically, I do not intend to bundle any alternative frontends in the main repo. My position on the web UI is that while it is using the API like an app, it is a *default UI*, and every software comes with one. Usually they're static, this one's React. Its using the API exclusively just to make sure it has no advantages over 3rd party apps.

The alternative frontends don't actually need to live on the server. Like TweetDeck for Twitter, you can just make a webapp that works with any instance you throw at it.

There is another argument against it, and it's the high rate of churn during development. We're changing a lot of stuff, and having to support multiple variations would add a lot of wasteful overhead.

Sorry, and thank you for your effort. I imagine you will keep this in a fork of yours.

Issue #221, opened 2016-11-23
Allow pasting images from clipboard as well as uploading image file


stuartlangridge:
It is useful to be able to paste an image from the clipboard to post it (most IM/chat/etc systems support this, such as Telegram, Slack, G+, and the like). Twitter doesn't allow this, and Mastodon doesn't either, which means that one has to save a grabbed image to a temporary file and then load it again to toot it. It would be handy to allow pasting in this situation.

geckojsc:
I think it's just Tweetdeck that doesn't allow this, whereas vanilla Twitter does (tested in Firefox on Windows 10). Anyways I agree it would be very handy. :)

seanlinsley:
It'd still be nice to do this

seanlinsley:
https://mastodon.social/@Viss/1523942

PR #657, opened 2017-03-01
#221 – Support for pasting images from clipboard

marrus-sh:
Uploads in general weren't wanting to work on my development setup, so maybe test this one before deploying. It's pretty straightforward though.


Gargron:
`AutosuggestTextarea` is a "dumb" component. Avoid connecting it to redux directly. Use a prop for the callback to pass the files to, and assign that prop from `ComposeForm`/`ComposeFormContainer`

marrus-sh:
Done~

Issue #255, opened 2016-11-24
Hide bulk of longer posts from different sites


unsmell:
earlier today I saw a post in the public timeline, 2935 characters in length, from a different instance. It covered the entire public timeline.
I propose posts >500 characters long should only be displayed in full when the user opens it

josefkenny:
In my opinion, this should probably be implemented as a more general clientside solution to people posting toots with a very large number of lines. I.e., decide to truncate the display of posts based on the height of the `status__content` div element, or by counting `\n`s, whatever is least resource-intensive.

unsmell:
that would help in cases where users abuse the 500 character limit and make ridiculously long posts. Such as, 
![image](https://cloud.githubusercontent.com/assets/2267475/20604366/91a41628-b255-11e6-9cf6-ecd56d54736a.png)

msappir:
This aspect of federation can be confusing for newbies. For more transparent UX, I suggest the "read more" link in these cases, or a little question mark icon next to it, should have a tooltip explaining this. Something like:

> This toot is longer than Mastodon allows - the instance it comes from probably follows different rules. Click through to see the full toot.

jeroenpraat:
I didn't found this one while searching.

But it is really needed. See #5477

kr428:
Adding myself to this after having gone through some more or less rough discussion. For whatever it's worth: Would it be possible to remove the post length restriction from mastodon altogether and make this something "clients" (web application, mobile apps) should take care of? Thoughts on that:

- This restriction generally doesn't make much sense in a federated environment. There's Friendica, Diaspora or GNUSocial which allow for postings substantially longer than 500 chars.
- If you end up in an actual communication on some issue (which I definitely consider a *good* thing), a fixed limit only increases the amount of (split-up) messages as, at some point, you won't be able to write down your arguments in < 500 chars.
- Apart from this, the confortable length of posts seems to drastically differ depending upon context: At nights on my desktop, large screen and enough time for reading, it's way easier to deal with long posts than on my smartphone while commuting.

Thanks for considering this. :)

lanodan:
[Crosspost of https://github.com/tootsuite/mastodon/issues/7973#issuecomment-413060740 as it’s a duplicate issue]
Just so that the people here are aware of it.

I made a PR(#8205) that rebases an old one (#658) fixing this long post issue.

[...]

PR #658, opened 2017-03-01
#255 – Collapsable toots!

marrus-sh:
Automatically collapses toots in timelines whose text contents are taller than 200px tall, with a button at the bottom to expand. Doesn't look at toots with spoiler/content warnings, so hypothetically if someone *really* wanted to fill up your feed with long toots they could abuse that. I don't really see that being a problem though.


yiskah:
I imagine this would solve the problem of encountering instances without character limits posting entire articles as well?

yiskah:
Hey @Gargron is this Pull mergable right now? It would help address some of the harassment/spamming concerns that have been brought up in relation to someone making instances for the purpose of spamming. Since a post could not be too long if collapsed.

haniawni:
With the recent influx of new users remarking on the prevalence of bots, I dug through my blocklist- plenty of the ones I've blocked are blocked because they just post immense toots.

This pull is **exceedingly** relevant to new users and will help with harassment in the future; is there any technical reason why this has not been already merged?
Thanks!

SethMichaelLarson:
This feature is desperately needed.

app/assets/javascripts/components/components/status_content.jsx
```diff
@@ -87,6 +91,11 @@ const StatusContent = React.createClass({
     this.setState({ hidden: !this.state.hidden });
   },
 
+  handleCollapsedClick (e) {
+    e.preventDefault();
+    this.setState({ collapsed: !this.state.collapsed });
```
ineffyble: Is this setting `collapsed` to !null? Does that result in false? A little unsure of the value flow here.

Pajn: It flips the boolean. If collapsed is null, it will set collapsed to true as null is falsey.
However, as `handleCollapsedClick` only is used when collapsed is not null that case wont happen.

dcl:
Hey, what's the status on getting this merged?

yiskah:
We need it to be reviewed by probably gargron before merging

yiskah (approved):

ashfurrow:
@marrus-sh thanks for the pull request! Sorry it's bene hanging around with no resolution. It looks like there are a few merge conflicts. Would you be able to pull from latest and review the changes? There are a few CodeClimate issues to resolve too. Then we can ping Gargron and see about getting this merged. Thanks again!

PR #659, opened 2017-03-01
Convenient colour palettes

marrus-sh:
This doesn't change *anything* about the site, but it does add some handy colour palettes (Tango, Solarized, etc) in `variables/palettes.scss` to make customizing your instance as easy as substituting one variable name for another.

### DIFF
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -1,8 +1,12 @@
-$color1: #282c37; // darkest
-$color2: #d9e1e8; // lightest
-$color3: #9baec8; // lighter
-$color4: #2b90d9; // vibrant
-$color5: #ffffff; // white
-$color6: #df405a; // error red
-$color7: #79bd9a; // succ green
-$color8: #000000; // black
+@import 'variables/palettes';
+
+//  For some colour palettes to chose from, see `variables/palettes.scss`!!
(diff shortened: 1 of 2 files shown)


ineffyble:
Ooh, I like it.

expenses:
Looks good. I suppose there could be concerns about bloat with this sort of thing, but we definitely want to put colours into a separate file to avoid 'magic numbers'.

mjankowski (changes requested):
Thanks for collecting all of these -- can you explain a little more about how these would be intended to be used?

I have some concern that we should improve the variable usage throughout the project first before adding "themes" or color-sets.

Separately, this PR needs a rebase.

ashfurrow:
Cool idea! Thanks for the PR @marrus-sh. I think now that we have the ability to add custom CSS maybe some documentation could be added too. Could you rebase with the latest changes from master?

ashfurrow:
Any update on this @marrus-sh ?

ashfurrow:
Hi there, thanks for the pull request! It's been a few weeks with no updates, and https://github.com/tootsuite/mastodon/pull/2912 has been merged which addresses the same problem. I'm going to close this pull request, but if you'd like to discuss it further please don't hesitate to open a new one, or an issue. Thanks again!

Issue #467, opened 2017-01-14
Back button when viewing toot doesn't work


ineffyble:
Right now, if I navigate to https://mastodon.social/web/statuses/625716 , the Back button in the right column does nothing. No errors in console. Confirmed it's happening to @hoodiek too.

Gargron:
Back button = back button in browser

I think we need to replace that with some form of hierarchical navigation history instead

PR #660, opened 2017-03-01
#467 – Making the back button a little smarter

marrus-sh:
If the length of `window.history` is `1` then there isn't anywhere to `goBack()` to, and the back button just takes you to `/`.


Gargron:
For future reference: the string "fix #[number]" in a commit message (or pull request title, i think) will automatically close the referenced issue upon being merged.

PR #661, opened 2017-03-02
Add mute button

KitRedgrave:
This isn't entirely done... when you unmute somebody, you seem to never get their posts again until you unfollow and refollow.

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -22,18 +22,8 @@ class FeedManager
   end
 
   def push(timeline_type, account, status)
-    timeline_key = key(timeline_type, account.id)
-
-    if status.reblog?
-      # If the original status is within 40 statuses from top, do not re-insert it into the feed
-      rank = redis.zrevrank(timeline_key, status.reblog_of_id)
-      return if !rank.nil? && rank < 40
-      redis.zadd(timeline_key, status.id, status.reblog_of_id)
-    else
-      redis.zadd(timeline_key, status.id, status.id)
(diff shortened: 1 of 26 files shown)


app/lib/feed_manager.rb
```diff
@@ -95,31 +85,47 @@ def redis
   end
 
   def filter_from_home?(status, receiver)
-    should_filter = false
+    should_filter = receiver.muting?(status.account_id)                       # Filter if I'm muting this person
```
Gargron: If this is true, the method can return instantly, no need to go through all the other checks

app/lib/feed_manager.rb
```diff
@@ -22,18 +22,8 @@ def filter?(timeline_type, status, receiver)
...
-      # If the original status is within 40 statuses from top, do not re-insert it into the feed
-      rank = redis.zrevrank(timeline_key, status.reblog_of_id)
-      return if !rank.nil? && rank < 40
-      redis.zadd(timeline_key, status.id, status.reblog_of_id)
-    else
-      redis.zadd(timeline_key, status.id, status.id)
-      trim(timeline_type, account.id)
-    end
-
+    redis.zadd(key(timeline_type, account.id), status.id, status.reblog? ? status.reblog_of_id : status.id)
```
Gargron: For some reason your branch is "up to date" yet this area is outdated - did you overwrite it?

Gargron (commented):
If you are curious I have left a couple comments on problematic parts of the code. I'll take it from here though.

PR #676, opened 2017-03-23
:wrench: S3 protocol  from ENV

leopku:
Add support for reading S3 protocol from ENV
Also add S3_HOSTNAME in .env.production.sample
I used [minio](https://minio.io/) in my test environment(without `https`) and part of production. So I add no-ssl environment support for this great software.


Gargron:
You should always use SSL in production though.

Issue #679, opened 2017-03-25
publicity levels UI


hoodiek:
![visual](https://cloud.githubusercontent.com/assets/23704040/24326445/cee55c70-1184-11e7-9854-27821398b14c.png)

i understand this may just be a pet peeve of mine, but such an essential tool as making something a dm or a private post feels like it should only require one click. i would suggest modifying the UI to look like what i sloppily threw together above, as it allows an alternate use case, dragging to the appropriate publicity level would be incredibly intuitive, imo. im unsure just how easy this is to implement, so this might not be possible, but, if you could click and hold down on the globe, then drag up or down to select a publicity level different than the current, i think it would be incredible. if not, i would still have the UI change i show, because i think it is less of a visual mess than having it drop down and waste space that can eventually be used for something else (Recent hashtags you have viewed, maybe?)

while we're on the topic, unlisted and public do not have a simple difference explained. i am unsure of what should be the correct wording, but i think this is lacking

PR #683, opened 2017-03-26
Privacy Option Accessibility Improvements

yiskah:
This pull includes both color contrast improvements on the descriptions of privacy options as well as clarifications to what they do. This includes changing the "Unlisted" icon to be fa-icon unlock-alt rather than being the same as Public; and giving each options a two line description to maintain even height while properly conveying levels of privacy [and not implying anything is encrypted] and interactions with public timeline.

This is definitely open to changes and tweaks. 

This addresses one issue brought up in #679

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -1703,7 +1703,7 @@ button.active i.fa-retweet {
 
 .privacy-dropdown__option__content {
   flex: 1 1 auto;
-  color: #536f96;
+  color: darken($color3, 24%);
 
   strong {
     font-weight: 500;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components.scss
```diff
@@ -1703,7 +1703,7 @@ button.active i.fa-retweet {
 
 .privacy-dropdown__option__content {
   flex: 1 1 auto;
-  color: $color3;
+  color: #536f96;
```
Gargron: This could be `darken($color3, N%)` where N is how many percent darker your version of the color is. The code doesn't have explicit color definitions anywhere but in variables.scss

yiskah: Ah, I will go back and change that. thanks!

yiskah:
I have changed the hard color value to darken($color3, 24%) which has the same output

PR #685, opened 2017-03-26
Clarify post privacy warning

yiskah:
I was informed that the current warning if you @ a remote server in a private post is inadequate. These are suggested changes to better inform users.


yiskah:
Hey @Gargron I think this would be good to merge soon if I haven't made any errors (I was careful to copy your code closely) since it addresses safety and privacy concerns that were brought up recently.

PR #703, opened 2017-03-31
Mastodon is not an alternative, it is a solution

alexgleason:
This month I traveled to Cambridge, Massachusetts with my friends and attended LibrePlanet 2017. At the talk *Technology for direct actions*, Andrew Seeder pointed out that we often call free software an "alternative" to what exists. But an alternative is something that is mostly similar, and differentiated by preference rather than need. For instance, a trackball is an alternative to a mouse, but a jail is not an alternative to a bedroom. Andrew suggested that we start looking at free software as "solutions" rather than alternatives. So I propose boldly proclaiming that Mastodon is not an alternative to Twitter, but rather a solution to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Mastodon
 [travis]: https://travis-ci.org/tootsuite/mastodon
 [code_climate]: https://codeclimate.com/github/tootsuite/mastodon
 
-Mastodon is a free, open-source social network server. A decentralized alternative to commercial platforms, it avoids the risks of a single company monopolizing your communication. Anyone can run Mastodon and participate in the social network seamlessly.
+Mastodon is a free, open-source social network server. A decentralized solution to commercial platforms, it avoids the risks of a single company monopolizing your communication. Anyone can run Mastodon and participate in the social network seamlessly.
 
 An alternative implementation of the GNU social project. Based on ActivityStreams, Webfinger, PubsubHubbub and Salmon.
 


yiskah:
Um, this isn't a pull request? There's no commits. I'd avoid making such bold claims.

Gargron:
There is a commit, it's a one word change in the README. I agree with the reasoning, for PR purposes "solution" sounds better than "alternative"

yiskah:
Yeah I realized after I read that. Woops.

I still hate empty language like "solution" personally but it's just on the github so it's mostly harmless.

alexgleason:
I am not suggesting that Mastodon center its entire marketing around the word "solution," just as it does not center its entire marketing around the word "alternative." The word solution is no less full or empty than the word alternative, so there is no reason to hate either word. The solution to 3 plus 4 is 7, and an alternative to English is Spanish.

I think it makes sense for Mastodon to consider itself a solution to proprietary networks rather than an alternative, because Mastodon is a reaction to proprietary networks. Proprietary networks like Twitter are a threat to society, and Mastodon is a safe haven built against them. Ideally they should not coexist, and only free federated networks like Mastodon should be used.

Mastodon does offer a lot that Twitter doesn't in terms of features and ease of use, but this alone will not encourage people to move over. The "killer feature" of Mastodon is in its free and decentralized nature. The beautiful user interface, excellent UX, and charming illustrations substantially reduce the barrier of entry for new users which is unprecedented for any decentralized network. Diaspora made people turned off because it was hard to understand so they never made it past the home page. Mastodon is friendly and inviting, so it has tremendous potential to explode in popularity. But users will need a reason to explore in the first place, and I'm unconvinced that curiosity or the desire for an improvement is going to cut it. Users need to be aware of the problem proprietary networks like Twitter cause in order to feel compelled to move. It should be no secret that Mastodon is offering a solution to a problem rather than providing an alternative to something.

PR #707, opened 2017-03-31
Update components.scss

tomfhowe:
Hide the thing covering up a bit of the left of the post creation ui on small screens/windows


tomfhowe:
A bit of a kludge, but it seems to work.

PR #711, opened 2017-04-01
Better use of large screen

BrunoSpy:
Better use of large screen : make the two middle columns use the available space

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -719,8 +719,7 @@ a.status__content__spoiler-link {
   box-sizing: border-box;
   background: $color1;
   display: flex;
-  flex-grow: 1;
-  flex-shrink: 1;
+  flex: 1 1 0;
   flex-direction: column;
 }
 
@@ -751,19 +750,6 @@ a.status__content__spoiler-link {
   text-align: center;
   font-size: 16px;
   border-bottom: 2px solid transparent;
-}
-
-@media screen and (min-width: 2560px) {
-  .columns-area {
-    justify-content: center;
-  }
-
-  .column, .drawer {
-    width: 350px;
-    border-radius: 4px;
-    height: 90vh;
-    margin-top: 5vh;
-  }
 }
 
 .drawer__pager {
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -719,12 +719,9 @@ a.status__content__spoiler-link {
   box-sizing: border-box;
   background: $color1;
   display: flex;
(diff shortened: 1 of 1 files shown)


Gargron:
This is a somewhat aesthetic personal preference..

For the record, to properly make the columns fill in the window width, it needs to be

```scss
.column {
  flex: 1 1 0;
}
```
The 0 instead of auto at the end is important, because otherwise the columns will not have equal width, depending on the content inside (e.g. toots with long unbreakable words would stretch the column wider).

Somewhere in the CSS there is a responsive design specifically for very wide screens (~4K). Perhaps you would like to lower the threshold of when that design becomes active, instead of this? Check it out and let me know what you think of it. (It's here: https://github.com/BrunoSpy/mastodon/blob/13d8dc0c04868e5d0a25cb4bb5e3ea0df999f52e/app/assets/stylesheets/components.scss#L756)

Columns aren't very readable when they are too wide or too narrow, which is why I opted for 330px, which is optimal.

BrunoSpy:
As a GUI designer, I don't think it's an aesthetic issue : if the user has a wide screen and its browser maximized, there's a reason and we do not have to decide if it's a good reason or not.
As a consequence, not using the whole space available should have a functional reason, which I don't see here.

EzoeRyou:
I am affected by this issue and it doesn't match my taste. Not using entire space is a waste of screen space.

Although I can easily workaround by user CSS, most users don't have a knowledge of that.
Ideally, all users should have have such skill but it won't work that way.

Currently, one fork at dwango/mastodon is considering to fix this issue themselves but they decided to wait for upstream fix.

I see this issue is stalled for more than 2 weeks now.  I really want this issue is fixed ASAP.

Or close this as Won't Fix so it can be fixed by downstream.

mjankowski:
Related PR which just adjusts the breakpoints to help large screens: https://github.com/tootsuite/mastodon/pull/2021

ashfurrow:
Thanks for the PR @BrunoSpy. Since https://github.com/tootsuite/mastodon/pull/2021 was introduced that fixes this issue and has been approved, I'm going to recommend we close this pull request. Thanks again for the change, please feel free to open up issues with any other UI suggestions you have so we can discuss them 👍

BrunoSpy:
#2021 absolutely doesn't fix the issue : columns still have fixed width, which is a really bad design, and it's not just my point of view.

ashfurrow:
I appreciate that you disagree, why don't we continue the conversation on #2021? This is by no means a finished conversation – it's very much up for discussion. Additionally, instances can customize the CSS directly. Let me know if I can clarify anything. Thanks!

PR #743, opened 2017-04-02
Update French Translation

Kazhnuz:
Some more strings translated, and some correction made to the translation (for instance a typo have been fixed), and some modification to make some element easier to understand (for instance the local/federated timelines".

I've also changed the order of the lines in the code, to be more like the one of the english translation, in order to spot more easily differences between locales.


Kazhnuz:
Seems that I forgot a comma, I've added it. Sorry for the mistake !

PR #751, opened 2017-04-02
update Dockerfile

Wonderfall:
I've made some changes, the most important is the move to Alpine Linux.

gargron/mastodon size : 1.85GB
This Dockerfile size : 1.14GB (- ~700MB)

My instance social.targaryen.house is running with this image without any problem.

EDIT : Sorry for the mess! You can squash, thanks. :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,32 @@
-FROM ruby:2.3.1
+FROM ruby:2.3.1-alpine
 
-ENV RAILS_ENV=production
-ENV NODE_ENV=production
-
-RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main contrib non-free' >> /etc/apt/sources.list
-RUN curl -sL https://deb.nodesource.com/setup_4.x | bash -
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev libxml2-dev libxslt1-dev nodejs ffmpeg && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,24 +1,31 @@
...
+    $BUILD_DEPS \
+    nodejs \
+    libpq \
+    libxml2 \
+    libxslt \
+    ffmpeg \
+    file \
+    imagemagick \
+ && npm install -g npm@3 && npm install -g yarn \
+ && bundle install --deployment --without test development \
```
Gargron: I'm messing with Docker again and I noticed this is suboptimal. While this Dockerfile optimizes for smallest possible end size, having it setup this way means that every time *anything* in the source directory changes, an entirely new image is created, all of these dependencies are installed from scratch etc. It not only takes a long time, but also means you end up with more needlessly different images on your filesystem which in turn eats up loads of disk space.

Dockerfiles must be structured this way:

1. Things that never change, e.g. build deps, system libraries/packages, users, system settings, etc
2. Dependency lockfiles, but *only* lockfiles, because we only want to trigger 3 to re-run if they change and nothing else
3. Installation of dependencies based on lockfiles
4. The rest of application code, which changes more often but requires no reinstallations of anything

Wonderfall: No, it has been updated so we can benefit from Docker caching. This was a mistake I introduced but it was fixed (I don't have the PR in mind). The actual Dockerfile is optimized for caching.

PR #765, opened 2017-04-03
Update API docs

marrus-sh:
I noticed that I was receiving properties in my API responses that weren't documented, so I went through the Mastodon API views and controllers and updated the documentation as best I could.


docs/Using-the-API/API.md
```diff
@@ -33,295 +44,480 @@ API overview
...
+| `blocking`               | Whether the user is currently blocking the account |
+| `muting`                 | Whether the user is currently muting the account |
+| `requested`              | Whether the user has requested to follow the account |
+
+### Report
+
+| Attribute                | Description |
+| ------------------------ | ----------- |
+| `id`                     | The ID of the report |
+| `action_taken`           | The action taken in response to the report |
```
Gargron: This isn't implemented yet.

PR #773, opened 2017-04-03
Finnish Translation Pull Request

JantsoP:
This is currently still a bit working in progress. Only two more files needs to be edited so it will be done soon(tm) hehe


JantsoP:
Sorry. Need to still do some fixing :)

PR #839, opened 2017-04-04
Improve readability of text on profiles

ddevault:


app/assets/stylesheets/accounts.scss
```diff
@@ -34,6 +34,7 @@
     text-align: center;
     position: relative;
     z-index: 2;
+    text-shadow: 0 0 2px #000;
```
Gargron: Please use `$color8` instead of `#000` on all of these, and this is good to go. (color8 resolves to 000 but is configurable)

SirCmpwn: This is not really a color that should change along with the rest of the black in the application.

Gargron: There are other places where full black color is used, and it is using that variable. For me it's about consistency in the code. I don't expect anyone to actually change color8's value.

PR #842, opened 2017-04-04
Missing quotes

angristan:


dereckson (approved):
Looks good to me, it's indeed more coherent.

PR #846, opened 2017-04-04
Update social.diskseven.com's IPv6 status

TrollDecker:
As far as I'm aware, my instance should be supporting IPv6 now. Was a config error on my part that it wasn't working before. Whoops.


dereckson (approved):
Looks good to me, as the server indeed now resolves to 2a03:b0c0:1:a1::1039:4001

PR #852, opened 2017-04-04
[#817] Add email whitelist

### LINKED ISSUE #817: Option to whitelist email domains
The .env.production configuration file contains a space to blacklist email domains to forbid them from signing up to an instance.

For those persons who would like to run an instance associated with their specific organization, a whitelist might be more helpful here - so only those who have email accounts with that organization can be permitted to register.

Would this be a feasible addition?

peterkeen:
This adds the ability to filter user signup with a whitelist
instead of or in addition to a blacklist.

Fixes #817


peterkeen:
The test that's failing here is also failing on master. I'll look into it this evening if I have time.

PR #854, opened 2017-04-04
Fix crontab edit

angristan:
Missing -u parameter to specify the mastodon user.


dereckson (approved):
Looks good to me, -u is indeed required.

PR #855, opened 2017-04-04
We don't need the entire tag model for just the name

krainboltgreene:
Reduce memory overhead.


Gargron:
Oops, I have already done this :( Sorry!

PR #863, opened 2017-04-04
Add closed_registrations message on French language

Zoeille:
:'D


dereckson (changes requested):
Extraneous space and dot at the end of the line (translation looks good to me).

PR #871, opened 2017-04-04
Added my instance to the list

cmiles74:
Nothing exciting, there are probably a zillion of these by now.


dereckson (changes requested):
I'd recommend to split the Apache configuration from the instance request.

Generally, it's more coherent to have one idea, one commit.

cmiles74:
That makes sense, I can split them up.

PR #872, opened 2017-04-04
typo in admin doc

kwaio:
s/rails/rake/


dereckson (approved):

Gargron:
In rails 5 rake tasks can be invoked from the rails binary.

kwaio:
I'm a rather old-school sysadmin, I don't know much about all those fancy languages & technologies, so i'll belive you =)

PR #876, opened 2017-04-04
Quick attempt to get pull requests passing

krainboltgreene:
There are some pull requests that are failing despite not actually changing anything related to this spec. I propose pending the spec until a fix can be prioritized. If we let it keep failing we'll miss real failures.


krainboltgreene:
@Gargron You might find it super helpful to turn on branch protection. It will force branches to update to latest in order to be merged.

PR #887, opened 2017-04-04
add persistance to Postresql container

valentin2105:
Create a volume for Postgres datas. Because if you kill all containers without this volume, all users are losts.


Wonderfall:
Same goes for redis : 

```
  volumes:
    - ./redis:/data
```

Gargron:
Does docker-compose automatically create these folders on the host, or do you have to create them yourself? Or does docker-compose complain if they're not there?

For the record when this option is not specified, the contains still use volumes, they're just anonymous volumes that are hard to find later.

Wonderfall:
@Gargron docker-compose is smart and does all the magic for you. :)

Gargron:
In that case please add @Wonderfall's suggested addition and I will merge

valentin2105:
I think it's a better idea to specify the path for volumes, for easy backup for example.

krainboltgreene:
This will break on local development in a vm (aka mac), since they wont have the right permissions.

valentin2105:
So why not create a specific dev compose file ?

krainboltgreene (approved):

Gargron:
While this is a good change for new installations, I have realized (the hard way from an admin trying this out) that this will make docker-compose **break previous installations during update**.

I am reverting this until we can figure out how to not brick people's existing configurations.

jemus42:
Is @Wonderfall's redis addition really necessary? Because I'm one of those admins who lost their db because of a `docker-compose down`, but after I found the old postgres volume in `var/lib/docker/volumes` and mounted that for postgres, my database was intact again.

Wonderfall:
@jemus42 `/data` contains the dump.rdb file.

PR #888, opened 2017-04-04
Add more tests to models

fakenine:
Bonjour !

I began to add more tests for models, in particular for validations or scopes.

By doing this I also refactored the validations for associations requirement in the models, which can be done directly in the `belongs_to` lines via the `required: true` option instead of a new `validates :account, presence: true` for instance.


Gargron:
Thank you for writing tests!!

PR #891, opened 2017-04-04
Adds instructions for adding admin users for Heroku installs

ashfurrow:


Gargron:
You don't have to use the heroku CLI, you can use the rails task `USERNAME=yourUsername rails mastodon:make_admin` through the heroku runner

ashfurrow:
I didn't know that, would you like an amended pull request?

Gargron:
Yes! I would like that

ashfurrow:
Done 👍

PR #894, opened 2017-04-04
Adding https://mastodon.cc

asm:
Adding my instance to the list.  Thanks!


SirCmpwn:
Should the instance list just be moved to the github wiki?

PR #898, opened 2017-04-05
Remote follow improvements

ddevault:
This stores the @username@instance you provide in your session and
reuses it the next time you remote follow someone from this instance.

I'm not really hip on the Ruby scene, please forgive me if I've made any
mistakes here.

### DIFF
--- a/app/controllers/remote_follow_controller.rb
+++ b/app/controllers/remote_follow_controller.rb
@@ -8,6 +8,7 @@ class RemoteFollowController < ApplicationController
 
   def new
     @remote_follow = RemoteFollow.new
+    @remote_follow.acct = session[:remote_follow] if session.key?(:remote_follow)
   end
 
   def create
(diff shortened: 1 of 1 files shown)


app/views/remote_follow/new.html.haml
```diff
@@ -7,7 +7,7 @@
   = simple_form_for @remote_follow, as: :remote_follow, url: account_remote_follow_path(@account) do |f|
     = render 'shared/error_messages', object: @remote_follow
 
-    = f.input :acct, placeholder: t('remote_follow.acct')
+    = f.input :acct, placeholder: t('remote_follow.acct'), value: @remote_follow.acct
```
Gargron: This is unnecessary as simple_form should pull the :acct value from @remote_follow, which you already set in the controller.. I think. Have you tested it? Otherwise this looks good to me

SirCmpwn: I have tested it, it seems to work. I'll see if it works without this.

SirCmpwn:
Cheers!

PR #902, opened 2017-04-05
SSL best practices for nginx

asm:
These settings should be a good starting point.  See https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html for reference.


gdurelle:
https://mozilla.github.io/server-side-tls/ssl-config-generator/

PR #907, opened 2017-04-05
broadcast inside of workers [wip]

krainboltgreene:
By pushing this into a worker we can reduce the amount of time the feed manager using workers eat up a connection.

I've also attempted to reduce some code inside the feed manager, but it looks like it has some extra dependencies.


krainboltgreene:
Having FeedManager own building a inline rabl view seems like tight coupling and a misappropriation of concerns.

If building an inline rabl view is difficult enough to create a helper, then lets create a proper helper.

Gargron:
FeedManager is the only place that needs to do inline rendering of that view... But I more or less agree that something must be done about that, especially Rabl itself (400ms+ just for rendering JSON!!)

Do you have any suggestions for how I should structure serializations if I am to switch from rabl to something else, perhaps direct OJ building?

krainboltgreene:
@Gargron I think ActiveModel::Serializer is a really nice approach. jsonapi_resources is what I would go with in the future, though.

Gargron:
I think this is good to go now, isn't it?

Gargron:
Although, no, I suppose there are a couple of other areas that are broken because the broadcast method has been removed from FeedManager... Just a bit more work

krainboltgreene:
Yeah, I'm going to work on those tonight.

PR #909, opened 2017-04-05
Add https://securitymastod.one/

scriptjunkie:


alerque:
I think this entire static list should be replaced with some more dynamic registry such as the one at https://instances.mastodon.xyz. Such a dynamic list can give a lot more useful information and is easier to maintain than a stream of PR's.

PR #911, opened 2017-04-05
Added mastodon.network instance.

ian-mcdowell:
Added mastodon.network.


alerque:
I think this entire static list should be replaced with some more dynamic registry such as what's at https://instances.mastodon.xyz .

Gargron:
I agree with @alerque. I am getting in touch with the person who runs instances.mastodon.xyz because first I want to make sure it can handle the traffic and I won't just be pointing to a black hole. However for the time being gonna accept these PRs

PR #915, opened 2017-04-05
Add `mastodon.kida.io`

DK1DA:


alerque:
I think this entire static list should be replaced with some more dynamic registry such as the one at https://instances.mastodon.xyz. Such a dynamic list can give a lot more useful information and is easier to maintain than a stream of PR's.

kimar:
@alerque I agree, I'm closing this PR then in favour of the dynamic list

PR #947, opened 2017-04-05
Add Scalingo deployment support

johnsudaar:
This patch should add everything needed for (relatively) simple deployment to Scalingo.
Scalingo is a cool european PaaS provider.

Changes:

Add a Scalingo one click button to README.
Add an scalingo.json file (used by scalingo to determine the app environment to create).
Add .buildpacks file (used by scalingo to determine which buildpacks should be use).
Add .slugignore file (used by paas providers to prevent build dependencies to be pushed to prod).

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,0 +1,2 @@
+https://github.com/Scalingo/ruby-buildpack
+https://github.com/Scalingo/nodejs-buildpack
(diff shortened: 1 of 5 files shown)


Gargron:
Have you tested that it works?

Gargron:
I've received complaints from people running Mastodon on Scalingo, and I want to confirm:

1. Does streaming API work on Scalingo? I don't see any configs for it
2. Obviously the cronjobs are missing and it seems like that wasn't mentioned in the docs for Scalingo deployment

PR #966, opened 2017-04-05
Fix wrong url in scalingo.json

johnsudaar:
The url on the scalingo.json is still pointing on johnsudaar/mastodon instead of tootsuite/mastodon


Gargron:
Why did you close?

johnsudaar:
I created another PR which also fix the documentation see #1042

PR #975, opened 2017-04-05
Use autohiding scrollbars in Microsoft Edge

tomfhowe:
Use nicer scrollbars in MS edge


tomfhowe:
I know this is really minor, but they are _much_ less ugly and more touch friendly - a lot closer to the narrow scrollbars you are using in other browsers already.

yiskah (approved):
I was shown a demo of this code. It only affects MS Edge, and seems to work. It does not appear to affect accessibility or the apparentness of your ability to scroll.

PR #978, opened 2017-04-05
Add toot export

ddevault:
Import can come separately. Requires some discussion and buy-in from the maintainers on a strategy.

### DIFF
--- a/app/controllers/settings/exports_controller.rb
+++ b/app/controllers/settings/exports_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'csv'
+include RoutingHelper
 
 class Settings::ExportsController < ApplicationController
   layout 'admin'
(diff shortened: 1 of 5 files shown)


Gargron:
Commendable effort, but my concern is: how does this behave when someone has, say, 90,000 toots?

The file returned would be huge, but most importantly, the request to generate it would take so long it would time out. Can you think of any ways to mitigate this?

SirCmpwn:
Perhaps we can set an upper limit on it, like it only exports up to the most recent 5000 toots?

papjul:
Maybe we can get inspired by how Twitter handles it?
It generates in a background task and when it’s ready you receive an email to download zip file. You should also put a limit of a generation every X hours so the same file is served if requested multiple times during a certain period of time.

mjankowski (changes requested):
Needs rebase.

app/controllers/settings/exports_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'csv'
+include RoutingHelper
```
mjankowski: What is this adding?

ashfurrow:
It sounds like there are some technical details that need to be resolved before we can review this. I'm going to suggest we move this into a issue instead of a pull request, and we can resolve things there?

Gargron:
Closing this down, toot export is on the backburner, this is incomplete, and has been inactive for a while. But uhh, in case you are reading this, OP, I do thank you for the effort!

PR #998, opened 2017-04-05
update docker-compose.yml for better security

ghost:
Bind ports of "web" and "streaming" to 127.0.0.1 for better security, cause it is more likely to use Nginx as a reverse proxy.

Exposing 3000/4000 directly to public internet is a huge security impact.


wade-r:
Sorry, wrong branch.

PR #999, opened 2017-04-05
update docker-compose.yml for better security

ghost:
Bind ports of "web" and "streaming" to 127.0.0.1 for better security, cause it is more likely to use Nginx as a reverse proxy.

Exposing 3000/4000 directly to public internet is a huge security impact.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     env_file: .env.production
     command: bundle exec rails s -p 3000 -b '0.0.0.0'
     ports:
-      - "3000:3000"
+      - "127.0.0.1:3000:3000"
     depends_on:
       - db
       - redis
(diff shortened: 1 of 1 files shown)


Wonderfall:
If nginx is in a container, we don't even need to use "ports". We just have to link the containers.

wade-r:
@Wonderfall At this moment, `docker-compose.yml` does not contain a nginx, and [Production Guide](https://github.com/tootsuite/mastodon/blob/master/docs/Running-Mastodon/Production-guide.md) suggests a standalone Nginx.

So I think it's better to bind 3000/4000 to 127.0.0.1 by default in `docker-compose.yml`

wade-r:
BTW, CI failed, it's not my fault, development branch seems to be out of sync with master.

PR #1000, opened 2017-04-05
Adds Croatian language [Hrvatski]

diomed:
most of Croatian translation for mastodon is added in this PR.


yiskah:
I'm not sure how we should handle localizations. Should we trust that they're good translations or try to get another speaker of that language to vet?

diomed:
Con't know about others, but I've localised several open source projects so far.
You can see them in my repos. Also, the sooner you merge, the sooner someone
might start using language, and post fixes for what is wrong. 
I did say, I didn't translate everything, but what I translated can be trusted.
I'don't spend time writing some rubbish, then sending a PR.

Gargron (approved):

PR #1021, opened 2017-04-06
Add tooltips to Clear & Settings buttons

kili-ilo:
Add the same type of HTML `title`s (and `aria-label`s) as on other icon buttons.


yiskah (approved):
Looks good to me. I think someone else should check this too though

Gargron:
I believe this has been already done in #1101 with localizations, sorry! Thank you for your effort. The only thing the other PR was missing was `aria-label`, so if you wanna reopen your PR just to add those, you are welcome to!

PR #1025, opened 2017-04-06
switch development redis config to match style of production

foozmeat:
I needed password support in my development redis so i brought the connection config over from production.


ineffyble:
@foozmeat can you clean up this PR? There's a lot of extraneous commits in here.

Gargron:
This PR submits a lot of past commits from other people as new commits. I suggest that you note your actual change, and open a new PR based on the master branch.

I also suggest that you search configs for more places where `REDIS_HOST` is used, and add the password addition there, also, but that can be a note on a new, cleaner PR - I cannot merge this.

PR #1030, opened 2017-04-06
Add infinimatix.net to instance list

DerekNonGeneric:


yiskah (approved):
it sure is another instance

PR #1035, opened 2017-04-06
Add mastodon.elao.com instance

gfaivre:


yiskah (approved):
yup, it's an instance

PR #1038, opened 2017-04-06
Optimize Dockerfile

ghost:
This modification will make Docker cache an intermedia image with all depencencies installed but project source code not included.

Upon next build, as long as `Gemfile/Gemfile.lock`, `package.json/yarn.lock` are not modified, Docker will not rebuild previously cached intermedia image.

This will save a lot of time.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ public/assets
 node_modules
 storybook
 neo4j
+vendor/bundle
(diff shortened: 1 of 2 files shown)


thurloat:
I was going to submit one like this as well, this will cut down on build time substantially.

Wonderfall:
Sorry I've made this mistake : https://github.com/tootsuite/mastodon/pull/751/commits/92d35c52d9829fe628b4c980772eead9686e1c79
Thanks for the commit. It would be nice if Docker Hub could use the caching feature...

PR #1063, opened 2017-04-06
significant improvement in microformats markup

### LINKED ISSUE #965: microformats clean-up
As stated in #122 and #421 the original requester did not consider this 100% corrected.
Starting this issue as a new issue since old one was closed.

I noticed this particularly bad on main feeds where the h-entry class is missing on the main user feed. Thus making the page look list a single entry with lots of copies of the h-card and many different contents.

* * * *

dissolve:
This is a huge improvement and I believe will close #965.

Had these microformats reviewed by others in the community to help
ensure they are at least correct, if not complete.

I did not want to change the structure of the page, and so there it does
not fully mark up the entire ancestry chain, or reply chain, only the
direct decendants and direct ancestors are correctly associated, but
this is likely fine as the most important bit is to have access to the
urls for those toots which are now correctly fetchable.


skddc:
@dissolve Have you seen the CodeClimate failures? https://codeclimate.com/github/tootsuite/mastodon/pull/1063

(Not a maintainer here, just someone who like the PR. 😉)

dissolve:
keeping this PR up to date with master, its a bit of a nuisance to keep updating it.  Any idea what is needed to proceed here?

Gargron:
I'm reviewing this right now.

app/views/stream_entries/_detailed_status.html.haml
```diff
@@ -5,14 +5,15 @@
         = image_tag status.account.avatar.url(:original), width: 48, height: 48, alt: '', class: 'u-photo'
     %span.display-name
       %strong.p-name.emojify= display_name(status.account)
-      %span.p-nickname= acct(status.account)
+      %span= acct(status.account)
 
-  .status__content.e-content.p-name.emojify<
+  .status__content.p-name.emojify
```
Gargron: The line is missing `<` at the end. That removes extra whitespace from the output HTML, which *does* get picked up visually because we want to support ASCII art in toot text

Gargron (changes requested):

Gargron (approved):

Issue #1040, opened 2017-04-06
Profile and toot backgrounds


DeadSuperHero:
This seems like it could be a really easy fix, and I think it might actually make for a nice visual enhancement.

As a small concession for personalization, it would be nice if the background on user profiles and individual toots matched with the header image. So, if I navigate to my user page, I can see a blurred background that matches it. 

<img width="1054" alt="sean_tilley_-_social_nasqueron" src="https://cloud.githubusercontent.com/assets/401560/24745335/324214be-1a6a-11e7-83f9-7eb42f38fc94.png">

This is incidentally not that hard to create with CSS, and I think it can be accomplished with a pretty basic change to some templates. Rather than being a custom image, backgrounds can be created automatically from uploading a header picture.

The only caveat I can think of is that a basic fallback image might be necessray for new users who haven't set an image yet.

mkody:
This looks really nice, [I've tried it on my instance](https://im-in.space/@kdy).

I added those two lines to `app/views/accounts/show.html.haml` above the `.h-feed`
```
.bg-header
  .bg-header-img{ style: "background-image: url(#{@account.header.url( :original)})" }
```

And then some scss (in `app/assets/stylesheets/accounts.scss`)
```scss
.bg-header {
    filter: blur(8px);
    height: 100%;
    left: 0;
    position: fixed;
    top: 0;
    width: 100%;
    z-index: -1;

    .bg-header-img {
        background-size: cover;
        background-position: center;
        height: calc(100% + 30px);
        margin: -15px;
        width: calc(100% + 30px);
    }
}
```

The negative margins and the `calc(100% + 30px)` are to hide the black spots made by the blur effect.

When a user doesn't have a header image, we can see the default Mastodon background [(example)](https://im-in.space/@curv).  
Does that look good enough?

Gargron:
@mkody I like that A LOT!

mkody:
@Gargron May I open a PR?

I see that we might need to add a background to the `NEXT` link at the bottom, it's hard to see it with light backgrounds.

ariasuni:
According to discussion in #1064, it was abandoned because of technical issues. One year afterwards, is it still not possible to do that without it being too complex/slow?

mkody:
We could pre-generate the blurred picture (based on the static header file) and set it as a simple background-image.
This would be a solution to some browsers/devices being slower due to the CSS3 blur effect.

[...]

PR #1064, opened 2017-04-06
Make the header a blurred background on profiles and conversation pages

mkody:
Related to #1040

Some screenshots:
- [Profile](https://cloud.githubusercontent.com/assets/843787/24761827/d079fe80-1aec-11e7-9a4a-b1ee0ce70579.png)
- [The "NEXT" button was themed a bit to make it visible in case of light background](https://cloud.githubusercontent.com/assets/843787/24761834/d4cc551e-1aec-11e7-8a83-9af8e41aab10.png)
- [Toots](https://cloud.githubusercontent.com/assets/843787/24761836/d6db64bc-1aec-11e7-8bfc-bf5e57a64362.png)

_Edit: and live [here](https://im-in.space/@kdy) or [here](https://im-in.space/@E1000)._


riking (approved):

eramdam:
I'm taking liberties and commenting on this.

- You don't seem to use the prefixed version of `filter`. Some browsers don't support the unprefixed version yet IIRC.
- It doesn't seem like you have a fallback for browsers that just don't support CSS filters altogether. Edge is one of them and probably some mobile users. You should probably use [`@supports`](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) to detect them and hide the full bleed background image in that case
- Since GIFs are supported in headers, it might be very problematic in terms of performance since CSS filters are very costly in terms of computation in the way they're implemented in browsers as of right now. I know @Gargron has some ideas regarding that issue but that's definitely something to consider.

mkody:
Thanks for the comments!

Except IE, every browser seems to support filters without any prefix. http://caniuse.com/#feat=css-filters 
And you're right, `@supports` should be used! :+1: 

For the gifs in headers, if there's at least a way to only get the first frame, I'm pretty sure that'd do it. Like a `@account.header.url( :original_static )` or something? (I'm not used to Ruby to know how to implement that properly.)

Gargron:
Generating static versions is indeed what I had in mind. A bit worried about using more disk space, but statics would be just a fraction of the original filesize... Also, maybe could only generated static copies for gifs, not other formats - for other formats just link the same file in both original and original_static.

Gargron:
I gave this a go, and it introduces quite a noticeable lag on Firefox/Linux and even Chrome. So even though I added static avatar/header versions just for this to work, it's not looking good right now :/

expenses:
@Gargron Would it be ok to for the instance to generate the static, blurry backgrounds so the browser doesn't need to blur them?

mkody:
I'm a bit surprised for the lag, I didn't notice any, even when resizing, on my old laptop (macOS with Chrome), my desktop Linux or on mobile/tablet (Android).

Gargron:
In the end I think we're fine without this, sorry but thank you for the contribution.

PR #1074, opened 2017-04-06
Add an instance.

kadiix:


kadiix:
We are the first Belgian instance :-)

yiskah (approved):
Frittes

kadiix:
Hi yiskah. 

There is an error. 
It's mastodon.brussels and not mastodon.brussels.fr

Thank you,
I hope it works soon because we need beer and French fries!

PR #1087, opened 2017-04-06
Update en.jsx - "What's on your mind" string

pappasadrian:
as suggested by @rowletbot@octodon.social
https://octodon.social/web/statuses/106228


Gargron:
I feel like that's way too informal... Very very unsure about this. For the record, you didn't link to the permalink (e.g. the link that the relative time points to) so we can't open it without an octodon.social account.

yiskah (changes requested):
buddy? How does this help? I can see "What's" versus "what is" but "buddy" is very unncecessary

expenses:
I think the default is fine as it is. You can change the string on your own instance if you want to though.

yiskah:
Expenses makes a good point. I'm going to close this as-is.

PR #1089, opened 2017-04-06
Change from British to American way

ghost:
... It's the American way

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -105,8 +105,8 @@ en:
         one: "1 new notification since your last visit \U0001F418"
         other: "%{count} new notifications since your last visit \U0001F418"
     favourite:
-      body: 'Your status was favourited by %{name}:'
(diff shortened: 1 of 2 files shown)


thurloat:
👎 I much prefer the British / Canadian version.

Gargron:
Honestly the US spelling just looks strange to me.

Spunkie:
@Gargron It doesn't really matter which way it is as long as it is consistent across the project. If you like the british spelling than everything should use `en-GB `spelling and vise versa for `en-US` spelling.

The project should formally pick one and then make it's use clear in [CONTRIBUTING.md](https://github.com/tootsuite/mastodon/blob/master/CONTRIBUTING.md)

yiskah:
This is pedantic. If you want to create separate localizations for american versus UK english where the only difference is the existence of a "u" then be my guest but that would be a different pull request.

PR #1090, opened 2017-04-06
Added mastodon.host in the list of instances

gled-rs:
Added pericles.world in the list of instances.

Has a follow bot that populates its federated timeline instead of being focused on one subject or community.


yiskah (approved):
sure

PR #1108, opened 2017-04-06
Production-guide.md:Extend webserver configuration

adlerweb:
move to configuration, add apache (https://github.com/tootsuite/mastodon/issues/809#issuecomment-292019489) and several hints which might be helpful


Exagone313:
Please use interactive mode to create this commit as here it's clearly broken. Also, this Apache configuration is not equivalent to nginx's. Finally, why bothering to use Apache (that seems to have issues with websockets)?

Gargron:

PR #1109, opened 2017-04-06
Use image too in docker-compose

pierreozoux:
It is possible with version 2 of compose to use both `build` and `image`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
   web:
     restart: always
     build: .
-    image: tootsuite/mastodon
+    image: gargron/mastodon
     env_file: .env.production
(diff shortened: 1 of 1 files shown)


krainboltgreene:
Images are already defined in the dockerfiles?

Gargron:
What effect does this have?

ineffyble:
Relevant info [here](http://stackoverflow.com/questions/32230577/how-do-i-define-the-name-of-image-build-with-docker-compose)

Building with a defined image name means you can create a new container from that image without rebuilding.

I'm not certain if this affects the default docker-compose workflow for this project. My understanding is that docker-compose builds the images from source, and uses them for future commands, even without a defined tag (so you have to manually tell it to rebuild on changes either way).

Gargron (approved):

PR #1120, opened 2017-04-07
Add more specific class names to notification divs

chris-martin:
I found myself wanting to add some custom CSS to fav/boost notifications, but there weren't any CSS classes to distinguish them. This diff just adds some classes. It shouldn't directly affect how anything renders in the default style.


yiskah:
Unsure if we should merge this when the classes don't call anything in the official unmodified code. Seems messy.

ineffyble:
If the class names are similar to current ones, and don't clash, then adding them as "placeholder" classes doesn't feel inherently bad.

app/assets/javascripts/components/features/notifications/components/notification.jsx
```diff
@@ -21,7 +21,7 @@ const Notification = React.createClass({
 
   renderFollow (account, link) {
     return (
-      <div className='notification'>
+      <div className='notification notification--follow'>
```
Gargron: When it's a namespaced (nested) element: `parent__child`. When it's a modifier: `parent-modifier`. I see nothing wrong with this PR in general but the format of these needs to be fixed.

Gargron (changes requested):

app/assets/javascripts/components/features/notifications/components/notification.jsx
chris-martin: Oh sorry, when I saw the class names with `__` I assumed you were following [BEM](http://getbem.com/introduction/). Will update.

chris-martin:
@Gargron Updated with requested changes.

yiskah (approved):
These look to be the requested changes

Gargron (approved):

PR #1124, opened 2017-04-07
Rewrite Atom generation from stream entries to use Ox instead of Nokogiri

Gargron:
`StreamEntry` is now limited to only statuses, which allows some optimization. Removed extra queries on `AccountsController#show`. `AtomSerializer` instead of `AtomBuilderHelper` used in `AccountsController#show`, `StreamEntriesController#show`, `StreamEntryRenderer` and `PubSubHubbub::DistributionWorker`

`PubSubHubbub::DistributionWorker` moves n+1 `DomainBlock` check to `PubSubHubbub::DeliveryWorker` instead.

All Salmon slaps that aren't based on `StreamEntry` still use `AtomBuilderHelper` and Nokogiri.


app/lib/tag_manager.rb
```diff
@@ -78,6 +78,8 @@ def uri_for(target)
     case target.object_type
     when :person
       account_url(target)
+    when :note, :comment, :activity
+      unique_tag(target.created_at, target.id, 'Status')
```
Gargron: This is a cheat, of sorts. Relying on `status.created_at` always equaling `stream_entry.created_at` (but they run in the same transaction so it's a safe bet). This saves an extra query

app/models/status.rb
```diff
@@ -10,7 +10,7 @@ class Status < ApplicationRecord
 
   belongs_to :application, class_name: 'Doorkeeper::Application'
 
-  belongs_to :account, inverse_of: :statuses, counter_cache: true
+  belongs_to :account, inverse_of: :statuses, counter_cache: true, touch: true
```
Gargron: We wanna have an update timestamp in the Atom feed header, but we don't wanna additionally fetch which entry is the latest when rendering the feed (e.g. when paginating through the feed, wouldn't be able to use the loaded set, needs extra query). With the touch, `account.updated_at` will be up-to-date with latest status.

app/models/stream_entry.rb
```diff
@@ -5,62 +5,54 @@ class StreamEntry < ApplicationRecord
 
   belongs_to :account, inverse_of: :stream_entries
   belongs_to :activity, polymorphic: true
-
   belongs_to :status, foreign_type: 'Status', foreign_key: 'activity_id', inverse_of: :stream_entry
 
   validates :account, :activity, presence: true
 
-  STATUS_INCLUDES = [:account, :stream_entry, :media_attachments, :tags, mentions: :account, reblog: [:stream_entry, :account, mentions: :account], thread: [:stream_entry, :account]].freeze
+  STATUS_INCLUDES = [:account, :stream_entry, :media_attachments, :tags, mentions: :account, reblog: [:stream_entry, :account, :media_attachments, :tags, mentions: :account], thread: [:stream_entry, :account]].freeze
```
Gargron: I missed this before

app/models/stream_entry.rb
```diff
@@ -5,62 +5,54 @@ class StreamEntry < ApplicationRecord
...
 
+  default_scope { where(activity_type: 'Status') }
```
Gargron: I am phasing out `StreamEntry`-as-a-polymorphic-thing, because the only model connected to it is `Status` now. This allows me to simplify some rendering code and queries.

app/models/status.rb
krainboltgreene: This could be potentially a lot of new SQL queries.

app/models/stream_entry.rb
```diff
@@ -5,62 +5,54 @@ class StreamEntry < ApplicationRecord
...
   scope :with_includes, -> { includes(:account, status: STATUS_INCLUDES) }
 
   def object_type
-    if orphaned?
-      :activity
-    else
-      targeted? ? :activity : activity.object_type
-    end
+    orphaned? || targeted? ? :activity : status.object_type
```
krainboltgreene: Why the switch to a ternary operator?

Gargron: Simplification. Easier to read

app/models/status.rb
Gargron: Only when a status is created or deleted, no?

app/models/stream_entry.rb
```diff
@@ -5,62 +5,54 @@ class StreamEntry < ApplicationRecord
...
+    orphaned? ? nil : status.thread
   end
 
   def mentions
-    activity.respond_to?(:mentions) ? activity.mentions.map(&:account) : []
-  end
-
-  def activity
-    !new_record? ? send(activity_type.underscore) || super : super
+    orphaned? ? [] : status.mentions.map(&:account)
```
krainboltgreene: If a status has 100 mentions this is 201 objects in memory.

app/models/status.rb
krainboltgreene: Yes, but everytime a status changes anywhere it also does an `UPDATE` for the corresponding account.

app/models/stream_entry.rb
krainboltgreene: Makes sense.

app/models/status.rb
krainboltgreene: Saved or destroyed: http://guides.rubyonrails.org/association_basics.html#touch

app/models/stream_entry.rb
krainboltgreene: One way to make this easier would be to use: http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations

Gargron: Hm, how do I rewrite this?

Gargron: Actually, sec, I don't think this method is called by anything anymore. Sigh. I need to carefully deprecate a lot of this junk, hah.

krainboltgreene: No worries :D

Gargron:
Alright, I think I'm done

PR #1125, opened 2017-04-07
Email service options :P

jasonrhodes:
small addition in case people want email service options, sparkpost gives you 100k/mo free


Gargron:
Please either put a comment marker before the mailgun one, or after it. We want to either force people to specify this option, or to give a working default. Right now, the comma just breaks the config while being almost invisible.

jasonrhodes:
:+1: makes sense

Gargron (approved):

PR #1128, opened 2017-04-07
fixed a typo

343max:
`install` instead `intall`


yiskah:
This was already caught. thanks

PR #1129, opened 2017-04-07
Set columns in frontend to have fluid width

ghost:
Makes the 3 right columns expand equally to fill browser window. sidebar stays the same size.

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -716,6 +716,7 @@ a.status__content__spoiler-link {
 
 .column {
   width: 330px;
+  flex: 1;
   position: relative;
   box-sizing: border-box;
   background: $color1;
@@ -814,7 +815,6 @@ a.status__content__spoiler-link {
 .column, .drawer {
   margin-left: 5px;
   margin-right: 5px;
-  flex: 0 0 auto;
   overflow: hidden;
 }
 


etienne-gauvin:
A PR already exists for this : #711

yiskah:
I am against this. It violates the style guides. The current layout is intended to reduce the need for eye movement.

krainboltgreene (changes requested):
See @yiskah's comment.

rainyday:
Tweetdeck's column width makes sense when you're limited to 140 characters but not when messages can be fairly long. Narrow columns mean more scrolling.

BrunoSpy:
> The current layout is intended to reduce the need for eye movement.
It doesn't make sense. Even on Twitter tweets are 590px : so why limit to 330px when you can write up to 500 char. ?
And as I told earlier on #711, it's not up to you to decide to use only half of the user screen.

ericblade:
Tweetdeck also supports the addition of new columns, something that I have yet to see a function for here.  And we're burning the entire left column on things that don't really need to be visible at all times.  

While this is a good start to making a Tweetdeck like interface, the average user is completely mystified by the UI here. :-S

yiskah:
I'm not sure if the intention is to recreate Tweetdeck. Redesigning the interface UX is a larger conversation that needs to be had before merging a pull

ericblade:
Certainly, but leaving a full third of the screen space as completely empty, such as on a 1920 res display with normal font settings, doesn't make much sense.  I'm saying if we could fill up that space with additional columns, then it wouldn't be much of an issue at all.

If the left hand column didn't exist (the menu bar was not an entire full-width column, and the text-entry popped into view only when needed) the app would barely cover half the display, if that, on a 1920 screen. And as is, the left-hand column is functionally useless, as the menu bar is duplicated in it's entirety on the right hand column, and you only need the text entry box when you're actually entering text.

Now, on my 1200 width display, it'd actually be almost perfect to fit 3 columns, if the left hand column were a fly-out instead of a constantly available column.  

As is, I'd say that the larger length of messages compared to Tweetdeck, would be deserving of columns that are *slightly* wider, and if you have the screenspace, then a Timeline column should always be visible (barring the ability for the user to create custom columns happening)

PR #1137, opened 2017-04-07
Update List-of-Mastodon-instances.md

Noiwex:


Noiwex:
also maybe it should be noted that admin of this instance tends to use censorship

PR #1139, opened 2017-04-07
Fixed two translation errors

imolein:
Changed followers from Follower to Folger and following from Gefolgt to Folgt, to make it identical to the translation in de.jsx

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -7,8 +7,8 @@ de:
     terms: AGB
   accounts:
     follow: Folgen
-    followers: Follower
-    following: Gefolgt
+    followers: Folger
+    following: Folgt
     nothing_here: Hier gibt es nichts!
     people_followed_by: Nutzer, denen %{name} folgt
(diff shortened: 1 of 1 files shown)


Gargron (approved):

skaaarrr:
> Changed followers from Follower to Folger and following from Gefolgt to Folgt, to make it identical to the translation in de.jsx

This is a proven wrong translation of the English word follower. The word "Folger" does not exist. There is the subtle form of "folgern" which implies, however, conclude. That word ist "folger". See: https://de.wiktionary.org/wiki/folger

It will get worse in later releases. There is "Folger" and "Folger_innen". Both totally wrong. Try to find another word .... like "es folgen X" for example. That would be correct and gender neutral.

Sidenode: "Folger" was not included in the duden, because of that problem. https://www.duden.de/suchen/dudenonline/follower

ykzts:
With [Crowdin](https://crowdin.com/project/mastodon) you can discuss the translation.

Issue #1136, opened 2017-04-07
Encoding::UndefinedConversionError since today's upgrade.


Aldarone:
Hello I am the admin of witches.town and I updated my instance this morning. However, the toots aren't federating anymore, my users can't follow or boost anyone and I'm seeing this error all over the logs :

```
web_1        | 2017-04-07T07:54:55.285964614Z [54648852-d910-4b17-85b0-eeb750093bd8] method=POST path=/api/v1/statuses/120331/favourite format=html controller=Api::V1::StatusesController action=favourite status=500 error='Encoding::UndefinedConversionError: "\xC3" from ASCII-8BIT to UTF-8' duration=66.97 view=0.00 db=11.48
web_1        | 2017-04-07T07:54:55.287485626Z [54648852-d910-4b17-85b0-eeb750093bd8]   
web_1        | 2017-04-07T07:54:55.287554509Z [54648852-d910-4b17-85b0-eeb750093bd8] Encoding::UndefinedConversionError ("\xC3" from ASCII-8BIT to UTF-8):
web_1        | 2017-04-07T07:54:55.287708976Z [54648852-d910-4b17-85b0-eeb750093bd8]   
web_1        | 2017-04-07T07:54:55.287748374Z [54648852-d910-4b17-85b0-eeb750093bd8] app/services/favourite_service.rb:16:in `call'
web_1        | 2017-04-07T07:54:55.287866350Z [54648852-d910-4b17-85b0-eeb750093bd8] app/controllers/api/v1/statuses_controller.rb:95:in `favourite'
```
Do you know how I can fix this ?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Aldarone:
It seems the same issue affect mastodon.social too. I'm reverting 6d6a429af8fe4bd92ed497f401676353fdc603e0 to see if it comes from here.

Aldarone:
Yep, that was it.

nolanlawson:
Question: is now a good time to start thinking about [semantic versioning](http://semver.org/) for Mastodon? Pulling from master seems very risky, if for no other reason than that various instances could be on various commits of the codebase, making it hard to track down bugs.

Then again, setting up SemVer inherently creates extra maintenance burden for @Gargron and so maybe it wouldn't be possible without more committers. Let us know what we can do to help! 😊

(Context: I maintain https://toot.cafe/)

mkody:
[...]

PR #1140, opened 2017-04-07
Force UTF8 encoding on generated XML

Gargron:
#1124 has introduced an exception (#1136) that wasn't caught during development and unit testing because testing and development environments use inline Sidekiq instead of enqueue-into-redis-properly Sidekiq.

This marks the generated XML as UTF8 which allows Sidekiq to JSON-encode it properly for the queue

### DIFF
--- a/app/lib/atom_serializer.rb
+++ b/app/lib/atom_serializer.rb
@@ -7,7 +7,7 @@ class AtomSerializer
     def render(element)
       document = Ox::Document.new(version: '1.0')
       document << element
-      "<?xml version=\"1.0\"?>#{Ox.dump(element)}"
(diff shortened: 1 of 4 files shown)


app/lib/atom_serializer.rb
```diff
@@ -7,7 +7,7 @@ class << self
     def render(element)
       document = Ox::Document.new(version: '1.0')
       document << element
-      "<?xml version=\"1.0\"?>#{Ox.dump(element)}"
+      ('<?xml version="1.0"?>' + Ox.dump(element)).force_encoding('UTF-8')
```
Gargron: This is the main fix

PR #1147, opened 2017-04-07
Fix npm/yarn cache cleaning

fmauNeko:
Hello,

Here's a PR for the Dockerfile, to clean the right NPM/Yarn caches and not the wrong one.
This makes the related layer shrink from 932MB to 582MB.


fmauNeko:
Rebased onto latest master.

Gargron (approved):

PR #1159, opened 2017-04-07
Version bumps for ruby and misc gems

mjankowski:
Brings ruby to version 2.4.1 and misc gems to their latest stable releases.


Gargron:
This will require a lot of people to reinstall ruby/recompile their docker images completely. I suggest waiting on this for at least a week or two more as I do not want to stretch admins' nerves too much.

mjankowski:
No problem, I'll keep it around.

Is the issue for admins there the ruby version bump, or just any bundler bumps at all? (I'm deploying to Heroku w/out Docker, not totally sure on what the Docker issues are).

If ruby alone is the issue, I can pull that one commit out of here and just get the gem updates in, and save ruby for later.

Gargron:
Bundler bumps are OK! They don't break anything.

Correct me if I'm wrong, but changing `.ruby-version` will make Bundler complain about wrong Ruby version, right?

If I'm wrong about that then I retract my line of reasoning and the upgrade would probably be non-breaking.

mjankowski:
With the ruby version bump, here's stuff I can think of which would happen:

- People doing development will be mildly irritated, because they'll have to do an `rbenv install 2.4.1` (or whatever they use) and then `cd` in and out of the project dir, and then `bundle install` again ... this should be a familiar thing for most rails devs, won't take too long.
- People who are doing very very manual server admin work will need to do a manual ruby version upgrade prior to their next deploy, to have the right version in place. I suspect there are very few people just using system rubies instead of rbenv/rvm/etc (or something like heroku or docker)
- People using heroku will "automatically" upgrade on next deploy, because heroku detects the version change in Gemfile and updates ruby for you
- I _think_ that for people using Docker, that the "onbuild" instruction here which was already in place will tell Docker to put ruby in place, and will just use the newer ruby version the next time it builds. I am not a Docker expert though, and am not 100% sure that this is always the behavior, or how that works precisely.

Gargron (approved):

mjankowski:
Awesome, thanks.  I'll attempt to stand by if anyone has weird upgrade issues...

Fastidious:
Can the old image `2.3.1-alpine` be removed after this has installed `2.4.1-alpine`?

Wonderfall:
@Fastidious ```docker rmi ruby:2.3.1-alpine```
```docker system prune -a -f``` to remove all unused images.

PR #1166, opened 2017-04-07
Use HTTP Accept-Language to detect locale

Awea:
Hi there have you considered use the HTTP Accept-Language header to detect the locale? Here is a pull request in that way.


ineffyble:
@Awea are you able to fix the Gemfile order issue raised by codeclimate?

Awea:
@ineffyble Done, sorry for the delay. I don't understand, how codeclimate define the gem order?

Awea:
@ineffyble Oh it's rubocop... ^^

Awea:
The new added gem cause a warning with Rubocop because they were not alphabetically sorted. This doesn't appear using rubocop outside codeclimate :(

Gargron:
I don't care about gem ordering in the Gemfile so don't worry about that one (I wonder if that can be disabled on codeclimate/rubocop)

app/controllers/concerns/localized.rb
```diff
@@ -26,6 +26,6 @@ def set_locale
   end
 
   def default_locale
-    ENV.fetch('DEFAULT_LOCALE') { I18n.default_locale }
+    ENV.fetch('DEFAULT_LOCALE') { http_accept_language.compatible_language_from(I18n.available_locales) }
```
Gargron: What happens when there is no accept-language header at all? I don't see a fallback to `I18n.default_locale` in the block.

Awea: You're right. I updated it.

Gargron (approved):

PR #1169, opened 2017-04-07
Install Yarn dependencies during setup

JoelQ:
Previously the `bin/setup` script only installed the Ruby dependencies
via Bundler. This is not sufficient to get the test suite to pass. We
also need the JavaScript dependencies.

This commit installs the JavaScript dependencies as part of the
`bin/setup` script via Yarn.

This allows for a successful

    bin/setup && bundle exec rake

on a fresh clone of the repo.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -17,6 +17,7 @@ chdir APP_ROOT do
   puts '== Installing dependencies =='
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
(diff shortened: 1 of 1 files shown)


ineffyble:
Should we be adding another line to run an `npm install -g yarn` if npm is installed, but yarn is not? 

That would be consistent with the current script running a `gem install bundler`.

Gargron (approved):

PR #1176, opened 2017-04-07
Add specs for PostStatusService media attachment validations

cpytel:
There are currently not specs for the two media validations that are performed
by `PostStatusService`. This adds specs for the validations that ensure that you
cannot attach more than four files, and that a status cannot have both image and
video attachments.


cpytel:
This is superseded by #1183.

Issue #692, opened 2017-03-29
a way to update avatars and profiles with the API


tinruufu:
it would be nice if you could update user profiles and avatars via the public API

this is required for me to be able to make [this borb](https://mastodon.social/@rowletbot) reach feature parity with [its twitter counterpart](https://twitter.com/rowletbot) in a way that doesn't involve phantomjs or otherwise pretending to be a browser

tinruufu:
my awful phantomjs hack broke a few days ago and took rowletbot with it

i've not had time to look into exactly what caused it, but i would greatly prefer using a real API to debugging this nightmare i built

Gargron:
There's the `PATCH /api/v1/accounts/update_credentials` method

I'm surprised the person who added it didn't reference this issue, so it got lost. It was #1179

tinruufu:
this could be that im missing something, but im currently running into a situation where i can update my avatar through the web UI, but i can only update my avatar through the api once

every time i set my avatar to something different, it's being set to the first avatar i ever set through the api, regardless of its actual content. i've tried to work out what's going on, but i promise im sending a different image each time, and it's always getting set to the same thing

it's not transient, either; i can set an image in the web profile editor, but then when i set another avatar in the api, it gets set back to that same first one i set

could someone who understands ruby look into if this is just me being bad at programming or if this is actually how this endpoint behaves? im using mastodon.social, if that helps

Issue #742, opened 2017-04-02
Allow users to edit recently published toots 


davidcelis:
One of the most annoying things about Twitter is publishing a tweet, noticing that it has a typo or other dumb mistake, copying the tweet text, deleting it, and re-posting it (while potentially re-adding a picture) to fix the typo.

While editing toots could be problematic because of changing the context of replies, it could be completely reasonable to allow users to edit a toot that they've published in the last, say, five minutes.

blackle:
duplicate of #256, move discussion there

PR #1179, opened 2017-04-07
Allow users to update their Account in the API

davidcelis:
It would be nice for API clients to be able to allow users to update
their accounts without having to wrap Mastodon in a web view. This patch
adds an API endpoint to let users submit a PATCH for their account.

Quick note: I thought about including the ability for a user to update their
email address via this API endpoint, but ended up omitting it since it's
technically an attribute of the User instead of the Account. If that's
not a big deal and you'd like me to add that, I can include that
functionality 😄 

Signed-off-by: David Celis <me@davidcel.is>

### DIFF
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 class Api::V1::AccountsController < ApiController
-  before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
+  before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute, :update_credentials]
   before_action -> { doorkeeper_authorize! :follow }, only: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
+  before_action -> { doorkeeper_authorize! :write }, only: [:update_credentials]
   before_action :require_user!, except: [:show, :following, :followers, :statuses]
-  before_action :set_account, except: [:verify_credentials, :suggestions, :search]
+  before_action :set_account, except: [:verify_credentials, :update_credentials, :suggestions, :search]
 
   respond_to :json
 
@@ -12,6 +13,14 @@ class Api::V1::AccountsController < ApiController
 
   def verify_credentials
     @account = current_user.account
+    render action: :show
+  end
+
+  def update_credentials
+    @account = current_user.account
+
+    @account.update_attributes!(account_params)
+
     render action: :show
   end
 
@@ -134,5 +143,9 @@ class Api::V1::AccountsController < ApiController
 
   def statuses_pagination_params(core_params)
     params.permit(:limit, :only_media, :exclude_replies).merge(core_params)
+  end
+
+  def account_params
+    @account_params ||= params.permit(:display_name, :note, :avatar, :header)
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -163,6 +163,7 @@ Rails.application.routes.draw do
         collection do
           get :relationships
           get :verify_credentials
+          patch :update_credentials
           get :search
         end
 
--- a/docs/Using-the-API/API.md
+++ b/docs/Using-the-API/API.md
@@ -85,6 +85,17 @@ Returns an [Account](#account).
 
 Returns the authenticated user's [Account](#account).
 
+#### Updating the current user:
+
+    PATCH /api/v1/accounts/update_credentials
+
+Form data:
+
+- `display_name`: The name to display in the user's profile
+- `note`: A new biography for the user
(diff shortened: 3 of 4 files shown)


app/controllers/api/v1/statuses_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::V1::StatusesController < ApiController
   before_action -> { doorkeeper_authorize! :read }, except: [:create, :destroy, :reblog, :unreblog, :favourite, :unfavourite]
   before_action -> { doorkeeper_authorize! :write }, only:  [:create, :destroy, :reblog, :unreblog, :favourite, :unfavourite]
   before_action :require_user!, except: [:show, :context, :card, :reblogged_by, :favourited_by]
-  before_action :set_status, only:      [:show, :context, :card, :reblogged_by, :favourited_by]
+  before_action :set_status, only:      [:show, :update, :context, :card, :reblogged_by, :favourited_by]
```
Gargron: Why is there suddenly a `StatusController#update` method? I don't remember that being in there (and I am against status editing after-the-fact)

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -15,6 +15,18 @@ def verify_credentials
     render action: :show
   end
 
+  def update_credentials
+    @account = current_user.account
+
+    account_params = params.permit(:display_name, :note)
```
Gargron: This could (customarily) be extracted into a private method

Gargron: Also: what about letting you update avatar + header as well? Why not?

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::V1::AccountsController < ApiController
   before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
```
Gargron: With how these are laid out, `AccountsController#update_credentials` would require only "read" scope from apps. It should be "write", or perhaps even a new type of scope altogether, since no other API methods have allows you to modify your profile before

Gargron (changes requested):
Good initiate, and will fix #692, but requires some adjustments

app/controllers/api/v1/statuses_controller.rb
davidcelis: Oops, that's a mistake brought in by something else I was testing. I'll remove it. P.S: If you're against status editing, you might want to comment as such on #742!

app/controllers/api/v1/accounts_controller.rb
davidcelis: I haven't personally worked on an API that allowed you to update something like an avatar or header, since file uploads can be tricky with REST APIs.

davidcelis:
I updated the controller to require the `write` scope for this action. Using `write` seemed fine to me in lieu of more finer-grained scopes, though I'd love to see scopes for specific resources and permissions come later

davidcelis:
Now that I think about it, we could accept Base64 encoded images like Twitter's API…

Gargron:
@davidcelis The media upload API uses standard form uploads so you can totally just use that in this method, too. Also: I know that Paperclip can handle a variety of upload types, it *might* handle Base64 out of the box, I haven't tried.

davidcelis:
@Gargron Paperclip _does_ automatically support Base64 uploads! I added that functionality to the endpoint; one can add a base64 encoded image with this format:

```
avatar=data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAUoAAADrCAYAAAA…
```

Gargron (changes requested):

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 class Api::V1::AccountsController < ApiController
-  before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
+  before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute, :update]
```
davidcelis: where?

davidcelis:
Fixed the typo 😓

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -15,6 +16,16 @@ def verify_credentials
     render action: :show
   end
 
+  def update_credentials
+    @account = current_user.account
+
+    if @account.update_attributes(account_params)
```
Gargron: Okay one last thing I noticed!

You can use `@account.update!(account_params)` here without the if/else construct. It will throw an `ActiveRecord::RecordInvalid` exception if the update fails, which is caught in `ApiController` to display a standardized error message.

Gargron (changes requested):

app/controllers/api/v1/accounts_controller.rb
davidcelis: Ah cool! Done.

davidcelis:
Added the new method to the API docs, as well (and resolved the merge conflicts that came up)

Gargron (approved):

PR #1183, opened 2017-04-07
Add specs for PostStatusService

cpytel:
This implements all pending specs for `PostStatusService`, and adds additional coverage for the
following functionality:

* Normal status creation
* Creating a reply status
* Creating a sensitive status
* Creating a status with spoiler text
* A status with no spoiler text gets an empty string for spoiler text
* Creating a status with custom visibility
* Creating a status for an application
* Processing mentions
* Processing Hashtags
* Pinging PuSH hubs
* Crawling links
* Attaching media

Additionally, the user-facing error text for the media validation errors was extracted to a localized string, to allow them to be localized.

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -37,11 +37,11 @@ class PostStatusService < BaseService
   def validate_media!(media_ids)
     return if media_ids.nil? || !media_ids.is_a?(Enumerable)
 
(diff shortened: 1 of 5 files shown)


cpytel:
This builds upon #1176.

Gargron (approved):
:+1: Adding tests is 👌 👌

PR #1188, opened 2017-04-07
French translation update

trebmuh:
one typo + one punctuation + one consistency
hope that helps


Exagone313:
The second change (emails => courriels) is controversial.

Gargron (approved):

PR #1191, opened 2017-04-07
change suggested cipher for nginx

d3vgru:
The current value of secp384r1 for ssl_ecdh_curve in the suggested nginx configuration causes problems for Android 7.0 devices connecting with an app. Changing the value to prime256v1 fixes the issue.

See section 4.2.1 https://datatracker.ietf.org/doc/rfc7525/?include_text=1

### DIFF
--- a/docs/Running-Mastodon/Production-guide.md
+++ b/docs/Running-Mastodon/Production-guide.md
@@ -24,7 +24,7 @@ server {
 
   ssl_protocols TLSv1.2;
   ssl_ciphers EECDH+AESGCM:EECDH+AES;
-  ssl_ecdh_curve secp384r1;
+  ssl_ecdh_curve prime256v1;
   ssl_prefer_server_ciphers on;
   ssl_session_cache shared:SSL:10m;
 


Gargron:
I would like to ping @asm and @gdurelle from #902 who originally modified the example config. Is this a secure change?

gdurelle:
I'm not competent enough in cyphers to answer, but maybe @bortzmeyer@mastodon.gougere.fr can help. He kind of rocks at security.

asm:
The NodeJS community has explored this issue a bit: https://github.com/nodejs/node/issues/1495.  It looks like P-256 *and* P-384 are no longer safe.  Given this, I suppose I would opt for the curve that's supported on more devices/browsers.

asm (approved):
There seem to be issues with both these curves but I'm in favor of using the one more widely supported.

Gargron (approved):

felixsanz:
Please take a look at this. https://security.stackexchange.com/a/78624/71526

PR #1194, opened 2017-04-07
Use Setting.site_title value for `og:site_name` occurrences

mjankowski:
If someone renames their instance and shares a URL into a place which respects the `og:site_name` value, this is nicer!

### DIFF
--- a/app/helpers/site_title_helper.rb
+++ b/app/helpers/site_title_helper.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module SiteTitleHelper
+  def site_title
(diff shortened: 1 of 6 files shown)


app/helpers/site_title_helper.rb
```diff
@@ -0,0 +1,5 @@
+module SiteTitleHelper
+  def site_title
+    Setting.site_title.presence
```
mjankowski: Good catch, that was from an earlier version which made this more complicated than it had to be before I understood what `Setting` was doing. Updated to just ensure a string is there.

Gargron (approved):

PR #1196, opened 2017-04-07
Adds Oulipo.social to Mastodons list

mouse-reeve:
This is a Mastodon with a particular constraint about
what symbols you can post.


yiskah (approved):
good

Issue #1114, opened 2017-04-06
Boosted toot displays the wrong author


chris-martin:
This is a boost of my toot:

- https://mastodon.hasameli.com/users/gorzilla/updates/1019

But on icosahedron.website it shows up like this:

![screenshot-icosahedron website-2017-04-06-18-55-58](https://cloud.githubusercontent.com/assets/399718/24778988/8f813236-1afb-11e7-8daf-14de5608bf96.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

hishamhm:
Perhaps Gordon just copied and pasted your message? Did you check for this?

DeadSuperHero:
Seems like an attribution issue that's referencing the person who boosted, rather than the OG author?

chris-martin:
The URL I posted above, which says it's a boost, is the one that the screenshot's time ("8 hours ago") links to.

Dr-Syn:
@hishamhm Nope, he didn't copy/paste - I'm the admin of that instance, and if you go to that particular post, you can clearly see it's a boost and not a copy/paste; the boost metadata is there.

chris-martin:
In case it's relevant, the person who re-boosted it into my timeline was @nathanic@isohedron.website

chris-martin:
This is what `@gorzilla@hasameli.com`'s profile looks like from icosahedron:

![screenshot-icosahedron website-2017-04-06-19-12-14](https://cloud.githubusercontent.com/assets/399718/24779235/06093dda-1afd-11e7-99c0-741bdf6d0081.png)

The top link goes to https://mastodon.hasameli.com/users/gorzilla/updates/1019 and the bottom link goes to https://icosahedron.website/@chris_martin/202038

chris-martin:
And this is what `@gorzilla@hasameli.com`'s profile looks like on `hasameli.com`:

![screenshot-mastodon hasameli com-2017-04-06-19-14-56](https://cloud.githubusercontent.com/assets/399718/24779290/6ec6ceaa-1afd-11e7-93ac-6d3314c80a36.png)

nathanic:
Hi, I am `@nathanic@icosahedron.website`, booster of this toot.*  I can confirm that at the time I boosted this toot, it appeared in the timeline as authored by Gordon.  I was actually surprised to find out later it was by @chris-martin.  So whatever broke, it broke before I boosted.  It still appears as authored by Gordon on my timeline.

[...]

PR #1215, opened 2017-04-08
Fix #1141, fix #1126 - Avatar/profile info fetching

### LINKED ISSUE #1126: External user avatars not working.
I finished setting up a heroku instance today and I got s3 working for uploaded avatars but my instance is not pulling in external user avatars from other instances. The same thing is happening in my timeline so it's not just that 1 user, all external user avatars are broken.

https://social.spunkiedesign.com/users/Spunkie/following

Anyone know what I need to do to fix this?

* * * *

### LINKED ISSUE #1141: Avatar and stuff from other instance not fetched
My two instances do not fetch avatar and stuff from other instance since rev 5442083 :-/
Users profil are empty, no avatar, no description, no banner.

Reset to c19e0f1 solve this issue.

* * * *

Gargron:
Work through `UpdateRemoteProfileService` for both `<feed>` and `<entry>` top-level tags, as they both may contain an `<author>`.

It was previously discarding top-level `<entry>` tags because of historical usage in `ProcessFeedService`.


app/services/process_feed_service.rb
```diff
@@ -50,6 +49,7 @@ def create_status
       if verb == :share
         original_status, = status_from_xml(@xml.at_xpath('.//activity:object', activity: TagManager::AS_XMLNS))
         status.reblog    = original_status
+        status.text      = ''
```
blackle: what's this addition doing? fixing null text for reblogged entries? is this necessarily part of the fix?

app/services/update_remote_profile_service.rb
```diff
@@ -1,7 +1,12 @@
 # frozen_string_literal: true
 
 class UpdateRemoteProfileService < BaseService
-  def call(xml, account, resubscribe = false)
+  def call(body, account, resubscribe = false)
+    xml = Nokogiri::XML(body)
+    xml.encoding = 'utf-8'
+
+    xml = xml.at_xpath('/xmlns:feed', xmlns: TagManager::XMLNS) || xml.at_xpath('/xmlns:entry', xmlns: TagManager::XMLNS)
```
blackle: I'm glad to see that pulling the author info out of the body is now the callee's responsibility instead of the caller, it feels more DRY

app/services/update_remote_profile_service.rb
```diff
@@ -12,7 +17,7 @@ def call(xml, account, resubscribe = false)
       account.note         = author_xml.at_xpath('./poco:note', poco: TagManager::POCO_XMLNS).content unless author_xml.at_xpath('./poco:note', poco: TagManager::POCO_XMLNS).nil?
       account.locked       = author_xml.at_xpath('./mastodon:scope', mastodon: TagManager::MTDN_XMLNS)&.content == 'private'
 
-      unless account.suspended? || DomainBlock.find_by(domain: account.domain)&.reject_media?
+      if !account.suspended? && !DomainBlock.find_by(domain: account.domain)&.reject_media?
         account.avatar_remote_url = author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS)['href'] unless author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS).nil? || author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS)['href'].blank?
         account.header_remote_url = author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS)['href'] unless author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS).nil? || author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS)['href'].blank?
```
blackle: the repetition in these lines make me uncomfortable.. idk if ruby has pass by reference (I'm still kinda rusty) but I feel like this could be a helper function like populate_from_feed(account.avatar_remote_url, xml, "avatar"). not necessary for the fix tho I guess

app/services/process_feed_service.rb
Gargron: Not necessary for the fix, but I noticed that remote reblogs had their wrapper text stored, which is inconsistent with local reblogs (empty text). That storage space is wasted. Also, when a bug like #1114 happens, in order of best-caseness:

1. do not have that bug happen at all
2. display "RT [original author] [text]" instead of just "[text]"
3. display an empty toot to make it obvious it's broken and not misattribute content visually

app/services/update_remote_profile_service.rb
```diff
@@ -12,9 +17,9 @@ def call(xml, account, resubscribe = false)
       account.note         = author_xml.at_xpath('./poco:note', poco: TagManager::POCO_XMLNS).content unless author_xml.at_xpath('./poco:note', poco: TagManager::POCO_XMLNS).nil?
       account.locked       = author_xml.at_xpath('./mastodon:scope', mastodon: TagManager::MTDN_XMLNS)&.content == 'private'
 
-      unless account.suspended? || DomainBlock.find_by(domain: account.domain)&.reject_media?
-        account.avatar_remote_url = author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS)['href'] unless author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS).nil? || author_xml.at_xpath('./xmlns:link[@rel="avatar"]', xmlns: TagManager::XMLNS)['href'].blank?
-        account.header_remote_url = author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS)['href'] unless author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS).nil? || author_xml.at_xpath('./xmlns:link[@rel="header"]', xmlns: TagManager::XMLNS)['href'].blank?
+      if !account.suspended? && !DomainBlock.find_by(domain: account.domain)&.reject_media?
```
ineffyble: Ruby best practice doesn't like IF NOT tests, from memory (e.g. better to have unless X than if !X). Not sure if that matters too much, but worth noting (I've been bugged a lot by Rubocop about this in past).

Gargron: Afaik, "If not [thing]" is better as "unless [thing]", but if the thing is a more complex boolean operation, "if" is preferred.

PR #1223, opened 2017-04-08
Added mastodon spanish instance

carlosjs23:
my spanish instance :+1:


ineffyble (approved):
Neat

JantsoP:
As the file states

PR #1225, opened 2017-04-08
Close instance list to additions

yiskah:
The instance list is generating too many pull requests too quickly. We are currently working on another solution. I propose we stop maintaining this list on GitHub.

PR #1227, opened 2017-04-08
Added Hacker Tribe to the list

lumenwrites:
Added my instance, https://hackertribe.io/, to the list.


krainboltgreene:
See: #1225.

JantsoP:
As the file states: 

PR #1228, opened 2017-04-08
update Node to 6.x LTS

dunn:
The 4.x branch [entered maintenance](https://github.com/nodejs/LTS#lts-schedule1) on 1 April.


krainboltgreene:
Great call.

krainboltgreene:
Need to make sure everything works with the breaking changes from 4 -> 6. I'm pretty sure it'll be fine.

dunn:
Dunno why Travis' `nvm` is getting confused; that version-string works fine locally.

.nvmrc
```diff
@@ -1 +1 @@
-6.7.0
+lts/boron
```
Gargron: Perhaps it expects a number version? Travis version of nvm could be different from yours.. I have not really used nvm before either, so I am not sure.

.travis.yml
```diff
@@ -28,7 +26,7 @@ before_install:
   - sudo apt-get -qq update
   - sudo apt-get -qq install g++-4.8
 install:
-  - nvm install $TRAVIS_NODE_VERSION
+  - nvm install
   - npm install -g npm@3
```
Gargron: If Node 6+ comes with npm >=3, this line can be removed

dunn:
https://github.com/travis-ci/travis-ci/issues/7593

dunn:
The Travis people replied; with `language: ruby` they won't guarantee a modern `nvm`, but I forgot we can just do `6` in the `.nvmrc` instead of specifying the LTS name.

Gargron (approved):

PR #1231, opened 2017-04-08
update Docker section of README

ericblade:
Re-ordered the steps so it doesn't read "Do this, but first, do this
other step"
Added note about keeping the REDIS and DB settings as they are for
Docker use
Add which variables you will NEED to set to make the Mastodon work
Add how to generate the secrets
Add how to connect to your Mastodon
Add a note to read the Production-guide


HashDot (approved):
lgtm

ericblade:
Rebased, updated to account for some of the changes that have occurred in master since I wrote the changes originally.  Particularly removed the line about gmail and wow SMTP servers not working, because I have not tested them with the new SMTP options that have been added, so it may very well be possible to get them to work now . . and the new changes to docker-compose.yml

Gargron (approved):

mig5:
This looks wrong - you can't run docker-compose or the 'rake secret' tasks prior to creating the .env.production file.

```
mastodon@m:~$ docker-compose build
ERROR: Couldn't find env file: /opt/mastodon/.env.production
```

```
mastodon@m:~$ docker-compose run --rm web rake secret
ERROR: Couldn't find env file: /opt/mastodon/.env.production
```

The only way around seems to be to 

1) create .env.production (with empty values for those secrets) and then
2) run the rake secret commands x3 to generate the secrets, then 
3) populate .env.production with those secrets, then 
4) run docker-compose build

ericblade:
Bah. Thanks for noting that. I'll send up an update shortly to fix that.  I try to do better than that when I submit things to other people's projects, but i'll call it a newbie mistake :-)

mig5:
@ericblade no prob, it was still a big improvement on the original instructions :)

PR #1234, opened 2017-04-08
Block :tm:, :copyright: from emojification

riking:
Forbid ™ ® © so they don't turn into black-on-black emoji.

I haven't figured out the docker-compose stuff yet so these changes haven't been tested yet. If someone else could make sure this runs that would be nice.

### DIFF
--- a/app/assets/javascripts/components/emoji.jsx
+++ b/app/assets/javascripts/components/emoji.jsx
@@ -2,6 +2,8 @@ import emojione from 'emojione';
 
 const toImage = str => shortnameToImage(unicodeToImage(str));
 
+const blacklistedHex = {"00ae": true, "00a9": true, "2122": true};
+
 const unicodeToImage = str => {
   const mappedUnicode = emojione.mapUnicodeToShort();
 
@@ -13,6 +15,10 @@ const unicodeToImage = str => {
     const unicode  = emojione.jsEscapeMap[unicodeChar];
     const short    = mappedUnicode[unicode];
     const filename = emojione.emojioneList[short].fname;
+    // Blacklisted - these have better font support than images
+    if (blacklistedHex[filename])
+      return unicodeChar;
+    }
     const alt      = emojione.convert(unicode.toUpperCase());
 
     return `<img draggable="false" class="emojione" alt="${alt}" src="/emoji/${filename}.svg" />`;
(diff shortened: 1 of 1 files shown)


ineffyble:
Is there a better option? Such as using a light/white version of the EmojiOne images?

riking:
Emojione doesn't seem to provide that from what I can tell.

This is also an issue for emoji such as :heavy_check_mark:  :heavy_division_sign: :heavy_dollar_sign: :heavy_minus_sign: :heavy_multiplication_x: but those don't have quite as widespread font support for the Unicode versions.

yiskah (changes requested):
This does not seem like a good solution for this problem. What will happen is instead of returning an intended look it will return the string ":tm:" etc. which will be perceived as an error.

riking:
Really? In my browser, the content of the `alt` attribute is ™, not `:tm:`.

marrus-sh:
The Unicode solution for this is to use the Emoji Variation Selector U+FE0E to select text representations for glyphs; Apple etc will insert these for you whenever you insert a text version of an emoji, however EmojiOne doesn't respect them (https://github.com/Ranks/emojione/issues/301, https://github.com/Ranks/emojione/issues/415). Furthermore, EmojiOne defaults to emoji representations for *all* characters, including a number of characters which should (by the Emoji spec recommendations for web pages) default to text (for example, ™, ®, ©; again see https://github.com/Ranks/emojione/issues/415). So this PR is actually a workaround for EmojiOne's bug and a good approach IMO; the right-hand column of [this chart](http://unicode.org/emoji/charts/text-style.html) provides other characters which predate emoji or otherwise might be likely to have text representations. Of course, it would be up to us as to which (if any) of these we would implement.

(See also https://github.com/tootsuite/mastodon/issues/349, which is the Mastodon version of this Issue before we realized it was on EmojiOne's side.)

I agree that using the `alt` attribute is less-than-ideal though; I don't have push access to this PR but the fix that I would suggest is here: https://gist.github.com/marrus-sh/71b3b4813e648ae4263e6a4e4b78c19a; note that this doesn't affect emoji inserted with shortnames at all (the assumption being that if you're using a shortname then you intended the emoji presentation), however I think Masto converts shortnames to character points now anyway so maybe that doesn't even matter. All I did was add an extra check to the `str.replace` function.

marrus-sh:
This is something of a separate issue, but with respect to dark emoji not being visible, see also https://github.com/tootsuite/mastodon/issues/317, especially [this comment](https://github.com/tootsuite/mastodon/issues/317#issuecomment-263797072) where I suggest using EmojiOne's Project B&W emoji for otherwise-unreadable emoji forms. Mastodon already uses SVGs for its emoji, so the changes that would need to take place would be as follows:

1. Tell Mastodon to pull from Project B&W SVGs instead of regular ones for dark emoji
2. Insert emoji using an `<svg>` element directly (through `<use fill="currentColor" xlink:href="[emoji path]">` or similar) instead of through `` and style that element appropriately. I'm like 75% sure that this will work. If not, then the emoji would appear as plain black, which would be the opposite of help.

ticky:
It looks like emojione 3.0 fixes this (Ranks/emojione#415), but we can’t use it until emojione-picker updates its dependency.

mjankowski:
I agree that we should let the supporting lib fix this problem and benefit from it when they do.

PR #1235, opened 2017-04-08
Add mst3k.interlinked.me to list of instances

Elizafox:
It is what it says on the tin. It's an instance for the interlinked.me IRC community but is available for general purpose use.


ineffyble (approved):
Looks good

PR #1237, opened 2017-04-08
Error in list

kadiix:
Hi, 
Just change mastodon.brussels.fr to mastodon.brussels 
It was a mistake and the mobile application don't recognise the .fr

Thank you :-)


ineffyble (approved):
Makes sense to me

PR #1242, opened 2017-04-08
Update mastodon.nuzgo.net

Zoeille:
Add ipv6 support


ineffyble (approved):
Confirmed the instance has v6 connectivity via online checker.

PR #1252, opened 2017-04-08
Improve about page responsiveness

xPaw:
Turns into this:


ineffyble:
Ooh, looks good, tested it on my laptop and my iPhone 6S - definite improvement. Half left one minor comment to check about an extraneous line.

Gargron (approved):

PR #1253, opened 2017-04-08
Add (commented) volume in docker-compose && Mitigating the HTTPoxy Vulnerability

valentin2105:
Hi,

Here is the docker-compose.yml with volume commented (will not break any existing config). 

I also disable root config for Nginx pointing to assets that doesn't exist and mitigate the HTTPoxy Vulnerability in Nginx configuration. 

(and also added my instance 👍 )


docs/Running-Mastodon/Production-guide.md
```diff
@@ -49,7 +57,7 @@ server {
     proxy_set_header X-Real-IP $remote_addr;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto https;
-
+    proxy_set_header Proxy "";
```
Gargron: I would like to hear more about this!

docs/Running-Mastodon/Production-guide.md
```diff
@@ -34,9 +34,17 @@ server {
   keepalive_timeout    70;
   sendfile             on;
   client_max_body_size 0;
-  gzip off;
 
-  root /home/mastodon/live/public;
```
Gargron: You can't remove this line. We *want* to serve static assets through nginx, rather than making Ruby/Rails do it. Please explain your reasoning!

docs/Running-Mastodon/Production-guide.md
```diff
@@ -34,9 +34,17 @@ server {
...
+
+  gzip on;
```
Gargron: Is this tested in a live production? I remember disabling gzip on purpose due to some incompatibility very early on, however a lot has changed since then, so it might be okay now.

Gargron (changes requested):

docs/Running-Mastodon/Production-guide.md
valentin2105: It's to mitigate a Nginx vulnerability  : https://httpoxy.org/
Rail seem not impacted, but I think, it's a good practice.

docs/Running-Mastodon/Production-guide.md
valentin2105: Yeah, it's work fine on https://mastodon.cloud,

Better performance with Gzip enabled.  
https://gtmetrix.com/reports/mastodon.cloud/FBZCuj81

docs/Running-Mastodon/Production-guide.md
valentin2105: When I tried it, Got no CSS, no JS and no pictures, all in 404. (folder doesn't mapped correctly in /public/ folder.) 

Work great without this directive,

(maybe it's my fault).

valentin2105: I put again this line to not disable it. (:

Gargron (approved):

PR #1256, opened 2017-04-08
enable AOF redis mechanism

Wonderfall:
https://redis.io/topics/persistence
https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/RedisAOF.html

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ services:
   redis:
     restart: always
     image: redis:alpine
+    command: redis-server --appendonly yes
   web:
     restart: always
     build: .


Gargron:
AOF does disk I/O more often than RDB. Good for preventing data loss, but bad when most of the Redis operations are not meant for persistence. Here is how we use Redis in Mastodon:

* Home feeds (valuable on-write cache that is costly to recalculate)
* Publish/subscribe for streaming API (no persistence)
* On-read cache, each key expiring after 20 minutes

What this *ideally* means is that deployments should have two different Redis servers - one for the valuable cache, and one for the throwaway cache. But I believe that I cannot impose such on small deployments (this might become a configuration option, however). I think that RDB is completely alright for a default.

But open to hear about your experiences re: performance or other concerns!

Wonderfall:
From what you say, I think RDB is a better choice.
Ping @Hardware.

PR #1257, opened 2017-04-08
French typo

SansPseudoFix:
#1148 continuation (wasn't finish :P).

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -63,16 +63,22 @@ fr:
   datetime:
     distance_in_words:
       about_x_hours: "%{count}h"
-      about_x_months: "%{count}mo"
-      about_x_years: "%{count}y"
-      almost_x_years: "%{count}y"
+      about_x_months: "%{count}mois"
+      about_x_years:
+        one: un an
+        other: "%{count} ans"
+      almost_x_years:
+        one: un an
+        other: "%{count} ans"
       half_a_minute: A l'instant
-      less_than_x_minutes: "%{count}m"
+      less_than_x_minutes: "%{count}min"
       less_than_x_seconds: A l'instant
-      over_x_years: "%{count}y"
-      x_days: "%{count}d"
-      x_minutes: "%{count}m"
-      x_months: "%{count}mo"
+      over_x_years:
+        one: un an
+        other: "%{count} ans"
+      x_days: "%{count}j"
+      x_minutes: "%{count}min"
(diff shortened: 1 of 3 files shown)


config/locales/fr.yml
```diff
@@ -58,21 +58,21 @@ fr:
...
+    prompt_html: 'Vous (<strong>%{self}</strong>) avez demandé à suivre :'
     title: Suivre %{acct}
   datetime:
     distance_in_words:
       about_x_hours: "%{count}h"
-      about_x_months: "%{count}mo"
-      about_x_years: "%{count}y"
-      almost_x_years: "%{count}y"
+      about_x_months: "%{count}mois"
+      about_x_years: "%{count}an(s)"
```
Gargron: It is possible to localize pluralizations like this:

```yml
about_x_months:
  one: One month
  other: "%{count} months
```

Gargron (commented):
I can merge as-is but you can get rid of the ugly "an(s)" by localizing pluralizations (see other comment) if you want, first.

config/locales/fr.yml
SansPseudoFix: Right! I'm on it, thx :)

config/locales/fr.yml
```diff
@@ -58,21 +58,27 @@ fr:
...
   datetime:
     distance_in_words:
       about_x_hours: "%{count}h"
-      about_x_months: "%{count}mo"
-      about_x_years: "%{count}y"
-      almost_x_years: "%{count}y"
+      about_x_months: "%{count}mois"
+      about_x_years:
+        one: un an
+        other: %{count} ans
```
SansPseudoFix: Like this?

nightpool:
Looks like you have build errors!

```
I18n::InvalidLocaleData:
       can not load translations from /home/travis/build/tootsuite/mastodon/config/locales/fr.yml: 
       Psych::SyntaxError: (/home/travis/build/tootsuite/mastodon/config/locales/fr.yml):
       found character that cannot start any token while scanning for the next
       token at line 69 column 16
```

have you validated your YAML? It looks like you're missing some quotation marks on line 69

SansPseudoFix:
> have you validated your YAML? It looks like you're missing some quotation marks on line 69

 Mh, I didn't valide my code, it would be a good idea indeed…
Now, it's ok: validation, tests, and squash :)

Gargron (approved):

PR #1263, opened 2017-04-08
NFS for shared folder requires a private_network

happycoloredbanana:


ineffyble (commented):
Is the issue that if the hostsupdater plugin is not installed, the synced_folder config fails?

Wondering if a better approach would be to move the private_network line out of the plugin test.

If not, I think it would be better to explicitly add a different config.vm.synced_folder line for `/vagrant` in an `else` statement for consistency (as right now, `/vagrant` is still mounted as a Vagrant default, but that's not clear from the Vagrantfile).

happycoloredbanana:
Yes, this is regarding the failure when the plugin is not installed.

I think the second option you mention is better (because no matter which changes are done to the VM configuration, the shared folder config will succeed).

Fixed accordingly.

ineffyble (approved):
Looks good to me, a fallback to standard VirtualBox shared folder functionality makes sense.

PR #1267, opened 2017-04-08
Allow running mastodon on a different domain as the one used for identifying users

### LINKED ISSUE #1032: Subdomains
I haven't gotten to running my own instance yet (soon!), but I'm seeing a lot of `@username@mastodon.example.com` in toots, instead of `@username@example.com`.

With email, you can host a server and access your mail at `mail.example.com` but still address to `username@example.com` thanks to MX DNS records. Users get a simple email address, and the servers figure out the rest.

If you have to use a subdomain because you already have something else on at `example.com`, it seems like you end up having to use  `username@mastodon.example.com` as your username instead, unless you create a new domain, like `example.social`.

It looks like there's a `LOCAL_DOMAIN` variable, but I haven't seen anything akin to email's MX record. I can't find much documentation on `LOCAL_DOMAIN`. I'm not sure if this is already there, but hasn't been fully documented/communicated yet, or this just hasn't happened yet.

* * * *

ClearlyClaire:
This should fix #1032, hopefuly.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -11,6 +11,9 @@ DB_PORT=5432
 LOCAL_DOMAIN=example.com
 LOCAL_HTTPS=true
 
+# Use this if mastodon has to run on a different domain as the one used for federation
+#WEB_DOMAIN=mastodon.example.com
+
 # Application secrets
 # Generate each with the `rake secret` task (`docker-compose run --rm web rake secret` if you use docker compose)
 PAPERCLIP_SECRET=
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -11,8 +11,9 @@ DB_PORT=5432
 LOCAL_DOMAIN=example.com
 LOCAL_HTTPS=true
 
-# Use this if mastodon has to run on a different domain as the one used for federation
-#WEB_DOMAIN=mastodon.example.com
+# Use this only if you need to run mastodon on a different domain than the one used for federation.
+# Do not use this unless you know exactly what you are doing.
+# WEB_DOMAIN=mastodon.example.com
 
 # Application secrets
 # Generate each with the `rake secret` task (`docker-compose run --rm web rake secret` if you use docker compose)
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -163,7 +163,7 @@ class ProcessFeedService < BaseService
 
         url = Addressable::URI.parse(link['href'])
 
-        mentioned_account = if TagManager.instance.local_domain?(url.host)
+        mentioned_account = if TagManager.instance.web_domain?(url.host)
                               Account.find_local(url.path.gsub('/users/', ''))
                             else
                               Account.find_by(url: link['href']) || FetchRemoteAccountService.new.call(link['href'])
--- a/app/controllers/api/push_controller.rb
+++ b/app/controllers/api/push_controller.rb
@@ -30,7 +30,7 @@ class Api::PushController < ApiController
     params = Rails.application.routes.recognize_path(uri.path)
     domain = uri.host + (uri.port ? ":#{uri.port}" : '')
 
-    return unless TagManager.instance.local_domain?(domain) && params[:controller] == 'accounts' && params[:action] == 'show' && params[:format] == 'atom'
+    return unless TagManager.instance.web_domain?(domain) && params[:controller] == 'accounts' && params[:action] == 'show' && params[:format] == 'atom'
 
     Account.find_local(params[:username])
   end
--- a/app/lib/tag_manager.rb
+++ b/app/lib/tag_manager.rb
@@ -56,6 +56,10 @@ class TagManager
     id.start_with?("tag:#{Rails.configuration.x.local_domain}")
   end
 
+  def web_domain?(domain)
+    domain.nil? || domain.gsub(/[\/]/, '').casecmp(Rails.configuration.x.web_domain).zero?
+  end
+
   def local_domain?(domain)
     domain.nil? || domain.gsub(/[\/]/, '').casecmp(Rails.configuration.x.local_domain).zero?
   end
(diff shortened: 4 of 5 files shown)


ThibG:
With this change, having the following in .env.production allows to have user@example.org handled by mastodon running on social.example.org, as long as https://example.org/.well-known/host-meta redirects to or proxies https://social.example.org/.well-known/host-meta

> LOCAL_DOMAIN=example.org
> WEB_DOMAIN=social.example.org

.env.production.sample
```diff
@@ -11,6 +11,9 @@ DB_PORT=5432
 LOCAL_DOMAIN=example.com
 LOCAL_HTTPS=true
 
+# Use this if mastodon has to run on a different domain as the one used for federation
+#WEB_DOMAIN=mastodon.example.com
```
Gargron: I would add a note that people should leave it commented out unless they know exactly what they are doing - and add a space after `#` for consistency.

Gargron (changes requested):
:+1:

ThibG:
Actually, something might still be amiss with PubSubHubbub subscriptions… while two people followed me since I've re-set my instance, there doesn't seem to be any effective subscription, and they don't see my toots. People from that server where already following me before I re-set my instance, and the feeds URL has stayed the same, which may or may not explain the issue.
I'll investigate further…

ThibG:
It seems that even on instances that don't have faulty URLs in their database (#1276), PubSubHubbub fail with HTTP status 422. I don't know why yet.

Gargron:
Ah, I think I know why that might be. PubSubHubbub validates that the requested feed URL belongs to your domain. I need to double check which variable it uses to check the domain.

Gargron:
Indeed, `Api::PushController#topic_to_account` checks the domain with `TagManager.instance.local_domain?` which is otherwise used only for federation purposes. Perhaps worth it to just do the check locally in that controller instead, it just needs to be a case-insensitive compare against WEB_DOMAIN || LOCAL_DOMAIN

ThibG:
Thanks for the hint, my last patch fixed it :)
Now, I only have to deal with #1276 !

app/services/process_feed_service.rb
```diff
@@ -163,7 +163,7 @@ def mentions_from_xml(parent, xml)
 
         url = Addressable::URI.parse(link['href'])
 
-        mentioned_account = if TagManager.instance.local_domain?(url.host)
+        mentioned_account = if TagManager.instance.web_domain?(url.host)
```
Gargron: No no no, this is the opposite of what you want. Web domain = mastodon.example.com. You want your users to be on example.com. Revert this one

Gargron (changes requested):

app/services/process_feed_service.rb
ThibG: Indeed, misread this part, sorry about that! I have reverted it.

ThibG: Actually no, the URI in the “mentioned” links from other people's toots seem to be full URLs to the web front-end, thus using WEB_DOMAIN and not LOCAL_DOMAIN

Gargron (approved):

plinss:
This change is much appreciated. Question though: is it safe to update LOCAL_DOMAIN/WEB_DOMAIN on an already running instance? e.g. if I change from:
```
LOCAL_DOMAIN=social.example.com
```
to:
```
LOCAL_DOMAIN=example.com
WEB_DOMAIN=social.example.com
```
(and have the webfinger proxied from example.com, of course)
Will existing users on other instances who are already following @user@social.example.com still be following properly? Or do they all have to start over with @user@example.com?

ThibG:
I am sorry, but this change is not completely usable yet: indeed, remote Mastodon instances will fail at accepting salmon API requests (follow/unfollow notifications, private messages, any message if you're not followed by someone on the remote instance) from you (see #2012).

As for your migration question, this is tricky. The upside is that you will keep all your toots, and referencing them should still work. You will also keep your profile URL (at social.example.com) and URI.

However, your remote contacts will still be following user@social.example.com and think they are followed by user@social.example.com. So far, it's just standard migration issues. But worse, interaction with user@social.example.com will be broken (so you can't easily tell people to switch over to user@example.com, you won't have both at the same time). Even worse, remote instances that knew social.example.com will already have your URI associated to the obsolete user@social.example.com, which might break salmon requests and mentions processing even when #2012 gets fixed.
With this two issues, “starting over” won't help you, as the important thing here is the URI, which will be the same whether you're “starting other” or not. This is related to #2032 except in this case, the URI was “valid” at the moment the user was registered in both cases, I'm not sure how it should be handled…

michaeljdeeb:
This might be a niche use case, but I wanted to point it out. I'm currently hosting a static site at my root domain using GitHub Pages and a Mastodon instance at [mastodon.michaeljdeeb.com](https://mastodon.michaeljdeeb.com). I'm only able to use an http refresh to do the redirect and because there's no extension on host-meta the meta redirect file is downloaded instead of actually redirecting.

The workaround is to just host the root site myself, but if this is a solvable problem you might be helping out many other single user instances in the same situation.

plinss:
@michaeljdeeb if you have control of your DNS, then a SRV record (and support in Mastodon/the GnuSocial protocols) would solve your use case.

ThibG:
@michaeljdeeb I guess you can simply host the host-meta file on your static website without doing any redirection. This host-meta file will mention mastodon.michalejdeeb.com for webfinger queries, and that's it!

But as I've said earlier, remote Mastodon instances will currently fail to accept salmon requests from your instance, thus preventing you from interacting with private accounts and instances where noone follows you (I think).

michaeljdeeb:
Oh! I did try that, but it gave me errors when I would try to follow people or they me, but now I'm thinking that might have to do with an HTTP root domain and HTTPS Mastodon instance.

PR #1270, opened 2017-04-08
don't use a block when #fetching with a default value

dunn:
When we aren't manipulating the value fetched, it's clearer to use the
default-value parameter instead of passing a block.

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -17,9 +17,9 @@ class AboutController < ApplicationController
     @extended_description = Setting.site_extended_description
     @contact_account      = Account.find_local(Setting.site_contact_username)
     @contact_email        = Setting.site_contact_email
-    @user_count           = Rails.cache.fetch('user_count')            { User.count }
-    @status_count         = Rails.cache.fetch('local_status_count')    { Status.local.count }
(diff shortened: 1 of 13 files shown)


app/controllers/about_controller.rb
```diff
@@ -17,9 +17,9 @@ def more
     @extended_description = Setting.site_extended_description
     @contact_account      = Account.find_local(Setting.site_contact_username)
     @contact_email        = Setting.site_contact_email
-    @user_count           = Rails.cache.fetch('user_count')            { User.count }
```
nightpool: This defeats the point of using a cache, though?

nightpool:
In the case that the default value is expensive to compute, this is a performance regression.

dunn:
Oh, I didn't realize there was a performance difference between `{ }.fetch(:a) { "default" }` and `{ }.fetch(:a, "default")`.  If that's the reason for this style, feel free to close this PR>

Gargron:
Blocks are evaluated when they are explicitly called, unlike inline statements. Sorry!

nightpool:
@Gargron the ENV ones could probably still use the change?

nightpool:
There's no reason for `ENV.fetch('SMTP_FROM_ADDRESS') { 'notifications@localhost' }` vs `ENV.fetch('SMTP_FROM_ADDRESS', 'notifications@localhost')`, as far as I know.

PR #1272, opened 2017-04-08
WORK IN PROGRESS Create Ubuntu-guide.md

vielmetti:
Not ready yet, but when it is ready it will have a guide to running Mastodon on Ubuntu 16.04 LTS on both Intel and ARM servers.


vielmetti:
There's a better guide out there than this one, so contribute to it instead. This can be closed.

PR #1274, opened 2017-04-08
Apache2 configuration example

FlorianANAYA:
I added an example configuration for apache2, for people who run apache instead of Nginx.


Gargron:

PR #1278, opened 2017-04-08
Fix #795, fix #704, fix #835 - 2FA requires confirmation to be enabled

### LINKED ISSUE #704: Do not show the QR code when user has enabled 2FA
In 2FA, QR-code is a shared secret, that should only be shared once (when the user enables 2FA). Then, the authentication works as a challenge-response communication.

In the current Mastodon version, when a user visits the “2FA” page, and 2FA is already activated, Mastodon reminds her the shared secret (the QR). This would permit any attacker (by MitM or direct access of an unlocked session) to steal the secret, and generate valid tokens, thus heavily reducing the 2FA’s security.

The procedure for 2FA activation should:
- display the QR-code
- ask the user for a proof that she can generate valid TOTPs
- if we received a valid TOTP, 2FA is enabled
- 2FA settings page only says that 2FA is enabled, and don’t show the QR-code, nor stores it in database

The “Disable 2FA“ button is also vulnerable, because a simple valid POST request can disable it. The most simple workaround is to ask for the user’s password before accepting to disable 2FA. A more secure option would be to double the password with a token sent by mail or phone, but Mastodon doesn’t have this information (and should not have it without a good reason).

For information, this issue follows this discussion: https://mastodon.social/@gordonzola/1388199

### LINKED ISSUE #795: 2FA not functioning as expected
This may be a symptom of the unusually large server load happening today, but so far I am locked out of my account and I'm not sure what to do about it.

### Steps to reproduce:

1. Go to account settings
2. Select the option to set up 2FA
3. Trigger a server error somehow
4. Log out
5. Enter username & pass to log in
6. App asks for 2FA credentials, user doesn't have them
    * (Asking to resend confirmation also doesn't work)

Not sure how to recover my personal account, but it seems that there's also an issue with the app gracefully handling 2FA config.  Thank you!  I ❤️ 🐘

### LINKED ISSUE #835: 2FA page should require submitting a token before enabling 2FA for account
When clicking on the 2FA link in settings, 2FA is immediately enabled for the account. The expected behavior would be at least 1) a confirmation or better 2) submitting one 2FA code to confirm the device is functioning correctly before enabling 2FA for an account.

* * * *

Gargron:
TOTP secret is not shown again after 2FA is enabled


app/models/form/two_factor_confirmation.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Form::TwoFactorConfirmation
+  include ActiveModel::Model
+
+  attr_accessor :code
+
+  def self.from_hash(hash)
+    obj = new
+    hash.each { |key, value| obj.send("#{key}=", value) }
```
krainboltgreene: `ActiveModel::Model` already gives you this ability.

``` ruby
Form::TwoFactorConfirmation.new(a: 1, b: 2).b == 2
```

krainboltgreene (approved):

PR #1281, opened 2017-04-08
Adding toot.BERLIN

yeah:
Thanks!


JantsoP:
As it states on the file: 

yeah:
I saw that notice, but I also saw [new instances getting added after that](https://github.com/tootsuite/mastodon/commits/master/docs/Using-Mastodon/List-of-Mastodon-instances.md), so I thought I give it a try anyways.

PR #1282, opened 2017-04-08
Clean up about page

mjankowski:
- Introduces a `Summary` class which is used to reduce all the i-vars in the about controller, and consolidate "site stats" type stuff
- Move the registration form to a partial
- DRY up the "about site" metadata in the about page
- Update the about page and "more" page to use new Summary class

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -2,29 +2,24 @@
 
 class AboutController < ApplicationController
   before_action :set_body_classes
+  before_action :set_instance_presenter, only: [:show, :more]
 
-  def index
-    @description                  = Setting.site_description
-    @open_registrations           = Setting.open_registrations
-    @closed_registrations_message = Setting.closed_registrations_message
+  def show; end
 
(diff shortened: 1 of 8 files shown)


krainboltgreene:
I like this a lot, but I would rename `Summary` to `SummaryPresenter` to match the Rails Pattern.

krainboltgreene (dismissed):
I like this a lot, but I would rename Summary to SummaryPresenter to match the Rails Pattern.

mjankowski:
Fair enough, will do.

I thought about `AboutPage`, `SiteSummary`, etc, and did not like any, and also don't really like `Summary`.

Any pref amongst `SitePresenter`, `InstancePresenter`, `SummaryPresenter`? (I'll just pick one later if no one cares).

Gargron:
InstancePresenter seems consistent with presence of /api/v1/instance

mjankowski:
Renamed to `InstancePresenter`, reduced commits.

Gargron (approved):

PR #1289, opened 2017-04-08
Add comment to settings.yml to nudge admins towards editing values via Web UI

rtucker:
I totally failed to realize this file was just defaults when I was setting things up.  This message should provide a helpful nudge in the right direction for new instances.


kballard:
Seems kind of weird to have a message to the admin show up publicly on new instances. I'd suggest you just put a comment in there pointing at docs on how to set up an admin user, without actually changing the description.

rtucker:
@kballard Hmm, good point, and I think you're right about this.  I've updated the patch accordingly.  Thanks!

Gargron (approved):

PR #1293, opened 2017-04-08
Add env variable to disable prepared statements

Gargron:


krainboltgreene:
Definitely be careful with this one.

krainboltgreene (approved):

Gargron:
@krainboltgreene What do you mean? Is something wrong with this code? Or do you mean admins need to be careful with that setting?

krainboltgreene:
@Gargron Just that removing prepared statements in certain Rails/pg.gem/postgresql versions can be weird.

Should be fine.

ineffyble (approved):

PR #1296, opened 2017-04-08
Added instructions for installing on Centos/RHEL.

existentialcomics:
I added instructions alongside the Ubuntu instructions for Centos users, and tested it from a fresh centos 7 server. I also added a couple other notes on things that tripped me up, postgres not being initialized, and the systemd files not having the right PATH and GEM_PATH set.


Gargron:

PR #1299, opened 2017-04-08
block all robots in robots.txt

nolanlawson:
We can update later if users want to tweak this, but I figure defaulting to most-private is best.


nolanlawson:
I'm so sorry, I meant to do this on my own fork. Please ignore.

PR #1313, opened 2017-04-09
Fix issue in doc with nginx talking to sidekiq on v6 while it is only listening on v4

packetcat:
Sidekiq listens single stack on 127.0.0.1.

The nginx config in the Mastodon documentation proxy passes to sidekiq with the hostname localhost, which resolves dual stack to 127.0.0.1 and ::1. 

This causes timeout errors in the nginx error log as it tries to connect to ::1 fails and fall back to 127.0.0.1. 

To fix this, change your nginx proxy_pass statement to use 127.0.0.1 instead of localhost.


Gargron:

PR #1320, opened 2017-04-09
Focus textarea when text is inserted, fix space after uploaded filenames

ghost:
Should improve usability of compose form when uploading images.


app/assets/javascripts/components/components/autosuggest_textarea.jsx
```diff
@@ -138,6 +138,12 @@ const AutosuggestTextarea = React.createClass({
     }
   },
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.value !== this.props.value) {
+      this.textarea.focus();
```
Gargron: You can accomplish this by setting `focusDate` to `new Date()` in the compose reducer on the `COMPOSE_UPLOAD_SUCCESS` action. Shouldn't do this here since this type of functionality is already implemented that way.

Gargron (changes requested):

rainyday:
@Gargron Done

Gargron (approved):

PR #1323, opened 2017-04-09
Refresh webfinger

### LINKED ISSUE #1276: Update feeds/profile URLs from webfinger
Hi,
I have made a mistake when working on #1032 and incidently advertised incorrect URLs for my account in the XRD served on webfinger requests.
Therefore, the majority of the servers I'm interested in have incorrect URLs about my instance.

The current behaviour of mastodon seems to be looking up the user via webfinger, then checking if there is local information about that user. If not, local information is created with webfinger's response. Otherwise, there is no code path to update that informations, even if the URLs have changed (which is my case).

I'm not sure how to interpret the OStatus spec, but I think the local data for remote users should be updated with webfinger's reply, at least if one of the URLs errors out.

ClearlyClaire:
This is an attempt to fix #1276 that I've been running on my instance without any ill effect (that I noticed),
but I couldn't verify that it fixes my particular issue, as the instances having incorrect information about my account are remote ones on which I could not try those changes.

This is split in two commits:
* The first one allows to refresh some existing account if webfinger returns new information. This only performs additional requests if one of the URLs or the public key did actually change. This is essentially dead code in the absence of the second patch, as local information on the remote account will be returned if it exists, without querying webfinger.
* The second patch performs the webfinger query whenever we are looking up at a remote account we are not subscribed to. The rationale is that the most important aspect of a remote account is the ability to be subscribed to, and if that fails, the account info may need updating. On the other hand, we may miss things like the user's web-facing profile URL changing. Furthermore, it may generate additional unwanted requests, and may probably be restricted to remote users we tried to subscribe to but failed.

As evidenced by the second patch, I am not too sure *when* we should re-query webfinger, any input on this would be appreciated.

### DIFF
--- a/app/services/follow_remote_account_service.rb
+++ b/app/services/follow_remote_account_service.rb
@@ -28,21 +28,30 @@ class FollowRemoteAccountService < BaseService
 
     return Account.find_local(confirmed_username) if TagManager.instance.local_domain?(confirmed_domain)
 
-    confirmed_account = Account.find_remote(confirmed_username, confirmed_domain)
-    return confirmed_account unless confirmed_account.nil?
+    remote_url  = data.link('http://schemas.google.com/g/2010#updates-from').href
+    salmon_url  = data.link('salmon').href
+    url         = data.link('http://webfinger.net/rel/profile-page').href
+    public_key  = magic_key_to_pem(data.link('magic-public-key').href)
 
-    Rails.logger.debug "Creating new remote account for #{uri}"
+    account = Account.find_remote(confirmed_username, confirmed_domain)
+    return account unless account.nil? || remote_url != account.remote_url || salmon_url != account.salmon_url || url != account.url || public_key != account.public_key
 
-    domain_block = DomainBlock.find_by(domain: domain)
+    if account.nil?
+        Rails.logger.debug "Creating new remote account for #{uri}"
 
-    account = Account.new(username: confirmed_username, domain: confirmed_domain)
-    account.remote_url  = data.link('http://schemas.google.com/g/2010#updates-from').href
-    account.salmon_url  = data.link('salmon').href
-    account.url         = data.link('http://webfinger.net/rel/profile-page').href
-    account.public_key  = magic_key_to_pem(data.link('magic-public-key').href)
-    account.private_key = nil
-    account.suspended   = true if domain_block && domain_block.suspend?
-    account.silenced    = true if domain_block && domain_block.silence?
+        domain_block = DomainBlock.find_by(domain: domain)
+        account = Account.new(username: confirmed_username, domain: confirmed_domain)
+        account.suspended   = true if domain_block && domain_block.suspend?
+        account.silenced    = true if domain_block && domain_block.silence?
+        account.private_key = nil
+    else
+        Rails.logger.debug "Updating remote account #{uri}"
+    end
+
+    account.remote_url  = remote_url
+    account.salmon_url  = salmon_url
+    account.url         = url
+    account.public_key  = public_key
 
     body, xml = get_feed(account.remote_url)
     hubs      = get_hubs(xml)
--- a/app/services/follow_remote_account_service.rb
+++ b/app/services/follow_remote_account_service.rb
@@ -16,7 +16,7 @@ class FollowRemoteAccountService < BaseService
     return Account.find_local(username) if TagManager.instance.local_domain?(domain)
 
     account = Account.find_remote(username, domain)
-    return account unless account.nil? || !account.subscribed?
+    return account if account&.subscribed?
 
     Rails.logger.debug "Looking up webfinger for #{uri}"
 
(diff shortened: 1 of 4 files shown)


app/services/follow_remote_account_service.rb
```diff
@@ -28,21 +28,30 @@ def call(uri)
...
-    confirmed_account = Account.find_remote(confirmed_username, confirmed_domain)
-    return confirmed_account unless confirmed_account.nil?
+    remote_url  = data.link('http://schemas.google.com/g/2010#updates-from').href
+    salmon_url  = data.link('salmon').href
+    url         = data.link('http://webfinger.net/rel/profile-page').href
+    public_key  = magic_key_to_pem(data.link('magic-public-key').href)
 
-    Rails.logger.debug "Creating new remote account for #{uri}"
+    account = Account.find_remote(confirmed_username, confirmed_domain)
+    return account unless account.nil? || remote_url != account.remote_url || salmon_url != account.salmon_url || url != account.url || public_key != account.public_key
```
Gargron: I see no harm in skipping this long complicated check/early return, since you already put `Account.new` in an `if account.nil?` block below. At that point, you can go past all these URL assignments and then do a `account.save! if account.changed?`

I agree that this method's logic is getting out of hand

Gargron (changes requested):

app/services/follow_remote_account_service.rb
ThibG: The point of that was to avoid fetching the atom feed if nothing changed. But you're right, it can probably be done in a better way. I'm doing that right now.

Gargron: You're past the webfinger fetch here, at least (`Goldfinger.finger("acct:#{uri}")`).

The feed fetch is `get_feed(account.remote_url)`, perhaps you could make it conditional on `remote_url` having changed.

Generally, yes, I am concerned about this service's performance, because it is called quite a lot, and we want to minimize extra network requests. Perhaps some of this work could be off-loaded into a background worker, but it wouldn't do to queue up thousands of duplicates, so there should be some sort of check on that. Perhaps something like `last_webfingered_at` and doing it only once a day or week or month?

Just throwing ideas out there, this is not a definitive suggestion.

ThibG:
Implemented your suggestion (call webfinger at most once per remote account per day), but now the testsuite fails because it tries to refresh webfinger info :)

app/services/follow_remote_account_service.rb
```diff
@@ -16,7 +16,7 @@ def call(uri)
     return Account.find_local(username) if TagManager.instance.local_domain?(domain)
 
     account = Account.find_remote(username, domain)
-    return account unless account.nil?
+    return account unless account&.last_webfingered_at.nil? || account.last_webfingered_at + 86_400 < Time.now.utc
```
krainboltgreene: Okay, so I might be a little out of date on the Ruby syntax, but what is `x&.y`?

krainboltgreene: Also, for `account.last_webfingered_at + 86_400 < Time.now.utc` we should use rails helpers:

``` ruby
1.day.from_now(account.last_webfingered_at) < Time.now.utc
```

app/services/follow_remote_account_service.rb
```diff
@@ -29,20 +29,24 @@ def call(uri)
...
+      account = Account.new(username: confirmed_username, domain: confirmed_domain)
+      account.suspended   = true if domain_block && domain_block.suspend?
+      account.silenced    = true if domain_block && domain_block.silence?
+      account.private_key = nil
+    else
+      account = confirmed_account
+    end
 
-    domain_block = DomainBlock.find_by(domain: domain)
+    account.last_webfingered_at = Time.now.utc
```
krainboltgreene: Why the `utc`?

db/migrate/20170409170753_add_last_webfingered_at_to_accounts.rb
```diff
@@ -0,0 +1,5 @@
+class AddLastWebfingeredAtToAccounts < ActiveRecord::Migration[5.0]
+  def change
+    add_column :accounts, :last_webfingered_at, :datetime, null: true, default: nil
```
krainboltgreene: If `null: true` and `default: nil`, then we can actually remove both options. All columns are by definition without a null constraint and by default nil.

app/services/follow_remote_account_service.rb
ThibG: Safe navigation operator: returns null if account is null, account.last_webfingered_at otherwise.

db/migrate/20170409170753_add_last_webfingered_at_to_accounts.rb
krainboltgreene: We also really don't need to use `datetime`, just use `timestamp`.

krainboltgreene (changes requested):

app/services/follow_remote_account_service.rb
krainboltgreene: So the native version of

``` ruby
account.try(:last_webfingered_at).nil?
```

app/services/follow_remote_account_service.rb
ThibG: I used the subscription timeout as a model, I guess it makes sense to have everything as UTC to not have to mess with timezones

app/services/follow_remote_account_service.rb
krainboltgreene: If so, then we can further simplify to:

``` ruby
return account unless account.last_webfingered_at || 1.day.from_now(account.last_webfingered_at) < Time.now.utc
```

db/migrate/20170409170753_add_last_webfingered_at_to_accounts.rb
ThibG: Also modeled after subscription_expires_at (and db/migrate/20160919221059_add_subscription_expires_at_to_accounts.rb), I'm fine with timestamp if it's better

ThibG: If I understand https://www.postgresql.org/docs/9.1/static/datatype-datetime.html correctly, in Postgresql, those two should be the same thing. Furthermore, it seems only datetime is currently in Mastodon, not timestamp.

app/services/follow_remote_account_service.rb
ThibG: Won't that do the opposite? If last_webfingered_at is nil, it will be compared in the second condition, and if it's anything truthy, it will trigger a re-fetch, even if we haven't timed out yet?

ThibG: As for 1.day.from_now, that indeed seems much more readable, but I don't know RoR enough to understand how it handles timezones

db/migrate/20170409170753_add_last_webfingered_at_to_accounts.rb
krainboltgreene: Looks like they are definitely the same!

app/services/follow_remote_account_service.rb
krainboltgreene: I believe AR and postgres handle this, but if other parts of our code are doing it then go ahead.

app/services/follow_remote_account_service.rb
krainboltgreene: You're right!

ThibG: I'll trust RoR to do the right thing on this, and use your suggested 1.day.from_now :)

krainboltgreene (approved):

app/services/follow_remote_account_service.rb
ThibG: That is what I went for in this patch: re-fetch any remote account (both from webfinger and the feed) for which we have not refreshed for at least one day using a “last_webfingered_at” field.
I have not seen any ill effect on my small instance.

Gargron (approved):

PR #1330, opened 2017-04-09
Do not store last visited URL from API controllers

Gargron:
Sign-in redirects you back to last visited URL, but in case of API requests,
this sometimes redirected users to an API URL that, of course, greeted them
with an {"error":"The access token is invalid"}

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -7,6 +7,7 @@ class ApiController < ApplicationController
   protect_from_forgery with: :null_session
 
   skip_before_action :verify_authenticity_token
+  skip_before_action :store_current_location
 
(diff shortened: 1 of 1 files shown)


mjankowski:
Looks fine, and that issue has been confusing people!

Maybe add some coverage for api controller to avoid regression here?

koboldunderlord (approved):
LGTM +1

davidcelis:
I'd love to see this merged; it's affecting a bunch of users on my instance

davidcelis:
I'm still seeing "401 The access token is invalid" repeatedly after logging in, even after clearing my cookies. Is there something stored for me in the database that I need to clear out?

PR #1333, opened 2017-04-09
Search cleanup

mjankowski:
Two changes in here:

- Add basic coverage that hits each of `Tag.search_for`, `Account.search_for` and `Account.advanced_search_for` and covers what I think are the basic cases for each. Probably not exhaustive, but better than previous (zero) coverage.
- Add a `squish` to a few SQL blocks, which cleans up their output (formats down to one line) for logging purposes.

If this is accepted, I have a [potential followup](https://github.com/mjankowski/mastodon/compare/search-cleanup...mjankowski:pg-search) which replaces the custom search code with the `pg_search` gem, I'd be curious on thoughts about that before I go further there. Right now I've replaced the basic searches, but not yet the advanced.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -203,7 +203,7 @@ class Account < ApplicationRecord
     end
 
     def triadic_closures(account, limit = 5)
-      sql = <<SQL
+      sql = <<-SQL.squish
         WITH first_degree AS (
             SELECT target_account_id
             FROM follows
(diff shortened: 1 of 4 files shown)


Gargron:
Re: pg_search, I have *replaced* that gem with this custom code. It is a lot more performant, and I know exactly what it is doing. Additionally, in pg_search it was not possible to rank account results based on following/followed_by relations to the searching user, because it wasn't possible to fully modify search queries. I am happy with plain SQL in this case.

Gargron (approved):

mjankowski:
Ah, noted on it already being replaced. I won't pursue that.

If I do want to further refactor, can you confirm that the only difference in advanced search for Account is that it adds in a "rank the results which are followed/following the passed in account higher than the non-follows"? Or is there more than that that's different?

PR #1334, opened 2017-04-09
Add timeout functionality

### LINKED ISSUE #1284: Timed Silence
As an admin it would be great to be able to do a timed silence on a user. Like, silence a user for X days / hours

* * * *

coilysiren:
closes https://github.com/tootsuite/mastodon/issues/1284

currently WIP, no tests or styling yet

### DIFF
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -34,7 +34,7 @@ class Admin::AccountsController < ApplicationController
     redirect_to admin_accounts_path
   end
 
-  def timed_silence
+  def timeout
     @account.update(silenced: true)
     UnsilenceWorker.perform_in(params[:days].days, @account.id)
     redirect_to admin_accounts_path
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -36,7 +36,10 @@ class Admin::AccountsController < ApplicationController
 
   def timeout
     @account.update(silenced: true)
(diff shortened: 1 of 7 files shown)


lynnco:
Here's what it looks like at the moment. It's not great, but I'm not a designer 
image

Gargron:
I feel like I have misunderstood your original intent after all. I thought you wanted to implement something to stop users from posting (e.g. "Error! You are timed out. You can post again in x minutes") like in Twitch chat. I see only limited utility in applying existing silence function on a timed basis... All it does is sandbox the user so they can't be seen by non-followers. It's not even meant to be noticed, so what's the benefit of hiding someone only for 24h, if they do something wrong enough to be hidden?

app/controllers/admin/accounts_controller.rb
```diff
@@ -34,6 +34,15 @@ def silence
     redirect_to admin_accounts_path
   end
 
+  def timeout
+    @account.update(silenced: true)
+    UnsilenceWorker.perform_in(
+      (params[:time].to_i * params[:modifier].to_i).hours,
+      @account.id
+    )
+    redirect_to admin_accounts_path
```
bradurani: I feel like this is a dangerous way to mange unsilencing. If a job fails (and there are many reasons it could fail) a user could potentially never be unsilenced

bradurani (changes requested):

lynnco:
I'm actually going to let this idea rest for a bit, I need more experience with moderation on mastodon and how it could be improved

so basically what @Gargron said

Issue #1337, opened 2017-04-09
Support JRuby


fap-:
I would like to see the possibility to run mastodon on JRuby.
I'm not sure if this would help performance but I think it's worth a try.

JRuby doesn't support gems with C extensions.
To provide JRuby support we have to define alternatives to all incompatible gems.

fap-:
Linking issues to track progress of postgresql jdbc support for ActiveRecord 5:
https://github.com/jruby/activerecord-jdbc-adapter/issues/734
https://github.com/jruby/activerecord-jdbc-adapter/issues/708

gouessej:
It would help a lot to reduce the attack surface on the servers already using Java. I have nothing against Ruby in particular but I won't install PHP, Python, Java and Ruby on my low power server.

@fap-  Have you ever tried to deploy Mastodon in Wildfly, TomEE or Jetty?

fap-:
@gouessej Can you explain how this might help?
We are lacking a JRuby+AR5 compatible database adapter for postgres.

And then we would have to find a way to replace the other C extension dependent dependencies:
https://twitter.com/codinghorror/status/852313187017936896
https://twitter.com/codinghorror/status/852300775954853889

gouessej:
I use Jetty on my own server (Olimex Lime2), it doesn't have a lot of RAM, it can't run numerous application / web servers at the same time, running Mastodon with JRuby would allow to bundle it as a WAR and deploy it as a web application with my other ones with a single JVM for Jetty itself which would have a lower memory footprint than running Jetty on one side and running something using Ruby natively on the other side.

Moreover, each "technology" might have its own flaws. If I install PHP + Python + Java + Ruby, I will be exposed to all their flaws. As far as I know, Mastodon requires Node.JS and Ruby. I expect the combination of JRuby + Java from having less flaws than Java and Ruby but maybe I'm completely wrong.

Gargron:
It sounds like JRuby is still incompatible with Rails 5, and JRuby integration in general seems to be a PITA. I don't think we should have an open issue for it.

PR #1338, opened 2017-04-09
WIP #1337 Support JRuby

fapdash:
Started working on this. Since my experience with JRuby is limited, any help is highly appreciated.


fap-:
Currently the database schema can't be set up properly, I'm not sure if this is a problem with [activerecord-jdbc-adapter](https://github.com/jruby/activerecord-jdbc-adapter) (ActiveRecord 5.0 support is still WIP) or if we can do something about it.

```
ActiveRecord::StatementInvalid: ActiveRecord::JDBCError: org.postgresql.util.PSQLException: ERROR: multiple primary keys for table "statuses" are not allowed
  Position: 53: CREATE TABLE "statuses" ("id" bigserial PRIMARY KEY PRIMARY KEY  PRIMARY KEY, "uri" character varying, "account_id" integer NOT NULL, "text" text DEFAULT '' NOT NULL, "created_at" timestamp NOT NULL, "updated_at" timestamp NOT NULL, "in_reply_to_id" bigint, "reblog_of_id" bigint, "url" character varying, "sensitive" boolean DEFAULT 'f', "visibility" integer DEFAULT 0 NOT NULL, "in_reply_to_account_id" integer, "application_id" integer, "spoiler_text" text DEFAULT '' NOT NULL, "reply" boolean DEFAULT 'f', "favourites_count" integer DEFAULT 0 NOT NULL, "reblogs_count" integer DEFAULT 0 NOT NULL)
arjdbc/jdbc/RubyJdbcConnection.java:555:in `execute'
/home/mastodon/mastodon/db/schema.rb:229:in `block in (root)'
/home/mastodon/mastodon/db/schema.rb:13:in `<main>'
bin/rails:4:in `<main>'
ActiveRecord::JDBCError: org.postgresql.util.PSQLException: ERROR: multiple primary keys for table "statuses" are not allowed
  Position: 53
arjdbc/jdbc/RubyJdbcConnection.java:555:in `execute'
/home/mastodon/mastodon/db/schema.rb:229:in `block in (root)'
/home/mastodon/mastodon/db/schema.rb:13:in `<main>'
bin/rails:4:in `<main>'
Tasks: TOP => db:schema:load
(See full trace by running task with --trace)
```

Gargron:
This may be a helpful resource: <https://devcenter.heroku.com/articles/moving-an-existing-rails-app-to-run-on-jruby>

fap-:
@Gargron Thank you, I had already read that.
We probably have to wait until [activerecord-jdbc-adapter](https://github.com/jruby/activerecord-jdbc-adapter) supports AR5 with postgres.
SQLite integration might already work but Mastodons schema seems to be quite postgres specific so I think it's just not worth the trouble, just to get it working with JRuby NOW.
If not someone else comes up with a smart idea we'll just have to wait.

Gargron:
Is it possible to use a different pg gem, maybe?

fap-:
This seems to be the only alternative and is neither compatible with Rails 5 nor maintained: https://github.com/headius/jruby-pg/issues/25

There is a working adapter but only for Oracle DB: https://github.com/rsim/oracle-enhanced#rails-50
=\

Gargron:
Alright, I suggest putting this on hold until JRuby catches up with Rails 5, then. Are you okay with that?

fap-:
Sure. :+1:

gouessej:
I haven't found any updated roadmap for JRuby, do you know when it's planned to support Rails 5? Have you ever tried to use Warbler to make a war for Mastodon and deploy it on Jetty?

Issue #755, opened 2017-04-02
[BUG] if an excess of notifications of a hidden type exist, notifications do not load


yiskah:
I've been getting tons of Favs and Boosts lately and so I disabled seeing those notifications in column. There's so many that they span multiple "pages." What is happening is that the notifications column seems to load a "page" of all favs and boosts and then hides them and then concludes I have no notifications and will not continue loading "pages" until it finds a reply. The work-a-round is to re-enable seeing boosts or favs so that it starts loading through infinite scroll (no "load more" button...) until it hits a reply, and then I can re-hide boosts. Which defeats the point of hiding boosts to begin with.

Just a bug.

Gargron:
Fixed in #1341

PR #1341, opened 2017-04-09
API param to exclude notification types from response

Gargron:
* Adds `exclude_types` scalar param to `/api/v1/notifications`
* Makes web UI use it with column settings


Gargron:
This would fix #755

app/models/notification.rb
```diff
@@ -16,10 +16,17 @@ class Notification < ApplicationRecord
 
   validates :account_id, uniqueness: { scope: [:activity_type, :activity_id] }
 
+  TYPE_CLASS_MAP = {
+    mention:        'Mention',
+    reblog:         'Status',
+    follow:         'Follow',
+    follow_request: 'FollowRequest',
+    favourite:      'Favourite',
+  }.freeze
```
krainboltgreene: Instead of a mapping you can use `.classify`

krainboltgreene (approved):

app/models/notification.rb
Gargron: The names don't match exactly in all cases, and more might be added in the future. In fact at some point I would make the type an explicit column

PR #1343, opened 2017-04-09
Fix #1339 - better Atom titles

### LINKED ISSUE #1339: [standard][atom] Review title/content and summary
In atom link : all content is html encoded in title, then loosing whole html tag.

`<title>&lt;p&gt;A paragraph&lt;/p&gt;</title>`

But this don't respect the standard and broken the fediverse.

This must be more clean, example
```xml
<title>New note by user</title>
<content type="html">
<p>A paragraph</p>
</content>
```

or 
```xml
<title>New note by user</title>
<summary>A paragraph</summary>
<content type="html">
<p>A paragraph</p>
</content>
```

Linked bug : https://github.com/tootsuite/mastodon/issues/1057
* * * *

Gargron:


Gargron:
Tests are broken until #1357 is merged btw

jbond23:
If I understand it correctly, this change means that *every* atom item has the same title "New status by userid". This may work for federation but it's user unfriendly if you're reading the Atom in an Atom reader. I'd prefer to see something like the first 128 chars of an html_stripped version of the description. userid is already in the channel.title so that's redundant in the item. And "new entry" is obvious.

PR #1346, opened 2017-04-09
Production Guide clarifications

mbilokonsky:
- Added more detailed instructions for configuring Nginx
- Added instructions to encourage use of LetsEncrypt for SSL
- Added optional monitoring section using Netdata

### DIFF
--- a/docs/Running-Mastodon/Production-guide.md
+++ b/docs/Running-Mastodon/Production-guide.md
@@ -5,7 +5,7 @@ Production guide
 
 It is recommended to create a special user for mastodon on the server (you could call the user `mastodon`), though remember to disable outside login for it. You should only be able to get into that user through `sudo su - mastodon`.
 
-    $ adduser mastodon sudo
+    $ adduser mastodon
     $ echo "DenyUsers mastodon" >> /etc/ssh/sshd_config
     $ sudo systemctl restart ssh
     $ sudo su - mastodon
@@ -41,7 +41,9 @@ Once that's installed, run:
 
     certbot certonly
 
-and follow the prompts. It'll ask you what domain name you want to get a certificate for - put in the public facing domain name, which should be the same as your instance name. The one you set up DNS for. This installer is cool, it has the ability to stand up a temporary webserver on your host that it will use to verify that you own the domain name you're claiming. Once this is done, great! You can provide HTTPS access!
+and follow the prompts. It'll ask you what domain name you want to get a certificate for - put in the public facing domain name, which should be the same as your instance name. The one you set up DNS for. For now, use the standalone server to verify (option 2). Once this is done, great! You can provide HTTPS access! 
+
+*You will have to do one more thing, later, to make this certificate auto-renew - see 'Updating Your SSL Certificate' near the end of this guide.*
 
 ### Nginx
 To install nginx, run:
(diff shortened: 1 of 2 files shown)


ashfurrow:
This clarifies a few things I had questions about when I tried following the guide, thanks!

docs/Running-Mastodon/Production-guide.md
```diff
@@ -1,99 +1,150 @@
...
 
-Regardless of whether you go with the Docker approach or not, here is an example Nginx server configuration:
+It is recommended to create a special user for mastodon on the server (you could call the user `mastodon`), though remember to disable outside login for it. You should only be able to get into that user through `sudo su - mastodon`.
 
-```nginx
-map $http_upgrade $connection_upgrade {
-  default upgrade;
-  ''      close;
-}
+    $ adduser mastodon sudo
```
patf: I don't think `mastodon` user should be in the `sudo` group.

docs/Running-Mastodon/Production-guide.md
```diff
@@ -1,99 +1,150 @@
...
-    proxy_set_header X-Forwarded-Proto https;
+    certbot certonly
 
-    proxy_pass http://localhost:4000;
-    proxy_buffering off;
-    proxy_redirect off;
-    proxy_http_version 1.1;
-    proxy_set_header Upgrade $http_upgrade;
-    proxy_set_header Connection $connection_upgrade;
+and follow the prompts. It'll ask you what domain name you want to get a certificate for - put in the public facing domain name, which should be the same as your instance name. The one you set up DNS for. This installer is cool, it has the ability to stand up a temporary webserver on your host that it will use to verify that you own the domain name you're claiming. Once this is done, great! You can provide HTTPS access!
```
patf: There's one problem with using the standalone challenge here: Renewal won't work. That's because `nginx` will occupy port 80 and 443 once you're done with this guide. Probably the best option for this use-case is the webroot challenge. You can point it to nginx's default value for the `root` directive (probably `/var/www/html`?) and later change it to mastodon's public directory in the corresponding renewal file in `/etc/letsencrypt/renewal`.

mbilokonsky:
hey @patf if I understood you correctly, my updated solution should work (I add a note in the LetsEncrypt section and a new subsection at the bottom about re-validating certs after getting everything running). Does this solve that problem?

patf:
Sounds about right to me. If you have an instance ready based on this guide, you can run a quick `certbot renew --dry-run` to verify that renewal will work. This will simulate everything on Let's Encrypt's staging server (including domain validation), but won't actually save the generated certificate, so it's great for testing the renewal procedure.

mbilokonsky:
Cool, I added that to the documentation!

Gargron:
I am a bit concerned with the scope of this new guide. The thing is, usually you *don't* need to modify `nginx.conf`. You just modify `/etc/nginx/sites-available/default`

mbilokonsky:
That's the sort of thing I would have added had I known it. I'll take a look at that when I get home later.

docs/Running-Mastodon/Production-guide.md
```diff
@@ -239,6 +292,62 @@ You may want to run `which bundle` first and copypaste that full path instead of
...
+      gzip_types *;
+    }
+    ```
+
+Then `sudo systemctl restart nginx` and navigate to https://YOUR_DOMAIN/netdata to see realtime monitoring data.
+
+## Updating Your SSL Certificate
+Remember above when we set up our LetsEncrypt certificate using a standalone server? In three months, when it tries to auto-renew that certificate, it's going to fail because we now have something else listening on those ports. (namely, nginx and our mastodon app). 
+
+So technically your certificate is fine - it will work for three months. But if you want it to auto-renew (and, like, you do) then once everything is working go ahead and re-run `certbot certonly`. This time, choose option (1) (Place files in webroot directory (webroot)). It'll tell you that your cert isn't close to expiration, and ask you if you're sure - say yes. Enter a new webroot (press `1` again when prompted) and when it asks for the path, give it `/home/users/mastodon/live/public` (or whatever your equivalent path is). This will allow the whole thing to re-run and verify your mastodon instance.
```
patf: Turns out I forgot something else. You'll need to reload your `nginx` after renewing; otherwise the new certificate won't be used. To do that, you need to change the cronjob from:

    0 */12 * * * root test -x /usr/bin/certbot -a \! -d /run/systemd/system && perl -e 'sleep int(rand(3600))' && certbot -q renew

to:

    0 */12 * * * root test -x /usr/bin/certbot -a \! -d /run/systemd/system && perl -e 'sleep int(rand(3600))' && certbot -q renew --renew-hook "service nginx reload"

That's the default cronjob on Debian (from `/etc/cron.d/certbot`). Not 100% sure if it's the same on Ubuntu.

Gargron:

PR #1360, opened 2017-04-09
Add REDIS_DB env variable to configure Redis database

lmorchard:
Small tweak that seemed make config easier for me on a server with multiple instances - need to have each use a different Redis database for its sidekiq queue


lmorchard:
Oops, this didn't quite solve the problem yet. Back to the drawing board!

PR #1361, opened 2017-04-09
Main update from Ruby 2.3.1 to Ruby 2.4.1

nicolasblanco:
Some gems needed to be updated like `webmock`, `redis` and `simplecov`, mainly to remove MRI warnings on Fixnum.


mjankowski:
Dupe of https://github.com/tootsuite/mastodon/pull/1159 which I opened previously ... seems like there's hesitation to bump ruby right now.

Gargron:
I would like things to settle before requiring all the fresh admins to immediately reinstall Ruby. Give it a week or two, I think. Does that make sense?

Unless there's a way to make .ruby-version value not break existing installations that use 2.3.1.

nicolasblanco:
All right, I'll close this one, following #1159.

PR #1362, opened 2017-04-09
vector (svg) logo with correct inner shape and colors

xuv:
Redrew the logo from scratch as the previous svg version had wrong shapes and colors. 
Here is an image comparison between previous svg version and new one, compared to the original png logo.
 
logo-update

File size could be improved, but wanted to keep original SVG features such as Circle to discribe the dots. If indeed file size is very important, improvements can be made regarding this.


expenses:
Looks good. You may way to use a svg optimizer to compress it down a little though.

xuv:
@expenses: thanks for the suggestion. Indeed it's worth it. Now even smaller than the previous version. 1.2Ko with the help of SVGO.

moritzheiber:
Can we get this merged so that FontAwesome gets a Mastodon icon in its font :smiley: ?

wxcafe (approved):
:+1:

PR #1363, opened 2017-04-09
Add env var to disable Heroku spawn

ashfurrow:
To support running on a single Dyno, the `HEROKU` env var implies that the web Dyno should spawn a Sidekiq process. I'd like to propose adding another env var for disabling this (without removing the `HEROKU` one). That way, if you want, admins can run on two Heroku Dynos and separate the web and workers. /cc @ineffyble.


Gargron:
Afaik that spawn is the *only* thing that variable did. I am all in favour of removing it altogether, because spawning sidekiq like this always rubbed me the wrong way - can't be good for performance, either.

ashfurrow:
That's all correct; I've moved off hosting on heroku anyway, so feel free to close.

ineffyble:
It was added in the hopes of utilising entirely free resources for Mastodon. I think that's impractical, so we could take out that option, update the deploy config to spin up multiple, and update documentation.

PR #1366, opened 2017-04-09
Add REDIS_DB env variable to configure Redis database

lmorchard:
Small tweak that seemed make config easier for me on a server with multiple instances - need to have each use a different Redis database for caching & sidekiq queue. (Kind of a Rails & Ruby newb, but hopefully this does the trick.)

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -55,6 +55,8 @@ Rails.application.configure do
     ENV['REDIS_HOST'] = redis_url.host
     ENV['REDIS_PORT'] = redis_url.port.to_s
     ENV['REDIS_PASSWORD'] = redis_url.password
+    db_num = redis_url.path[1..-1]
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -55,14 +55,16 @@
     ENV['REDIS_HOST'] = redis_url.host
     ENV['REDIS_PORT'] = redis_url.port.to_s
     ENV['REDIS_PASSWORD'] = redis_url.password
+    db_num = redis_url.path[1..-1]
+    ENV['REDIS_DB'] = db_num if not db_num.empty?
```
dunn: This can be `if db_num.present?`

dunn: Alternatively, if you're defaulting to 0, `ENV['REDIS_DB'] = db_num || '0'`

lmorchard:
Updated to use `if db_num.present?`

Gargron (approved):

PR #1369, opened 2017-04-09
domain change

lfuelling:
I changed the domain.


yiskah:
we are no longer maintaining the instance list here

PR #1372, opened 2017-04-09
More SMTP customization

pointlessone:
Allows customizing:
* SMTP auth method
* SMTP cert verification mode
* SMTP STARTTLS auto enabling

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -35,6 +35,7 @@ SMTP_PORT=587
 SMTP_LOGIN=
 SMTP_PASSWORD=
 SMTP_FROM_ADDRESS=notifications@example.com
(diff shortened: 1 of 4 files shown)


Spunkie:
You will need to add defaults for this in 
```
/scalingo.json
/app.json
/.env.production.sample
```

Also I had planned on doing a similar PR but adding fields necessary to get self signed certs will work. If you think these look good would you mind adding them to your PR?
```
:openssl_verify_mode  => ENV['SMTP_OPENSSL_VERIFY_MODE'] || 'peer',
:enable_starttls_auto => ENV['SMTP_ENABLE_STARTTLS_AUTO'] || true,
```

wxcafe (approved):
Looks good to me

PR #1378, opened 2017-04-09
Define volume for postgres storage

jemus42:
As I suggest in [this issue](https://github.com/tootsuite/mastodon/issues/1376), I think it would be a good idea to define a db volume to make sure users don't lose their database if they shut down their server or run `docker-compose down`.

Please be aware I am new to docker and don't know where a "good" place for the volume would be, suggestions welcome.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ services:
   db:
     restart: always
     image: postgres:alpine
+    volumes:
+      - ./postgres:/var/lib/postgresql/data
   redis:
     restart: always
     image: redis:alpine


rainyday:
wont this nuke everyone's existing databases if they happen to pull this without knowing and run docker-compose up?

jemus42:
@rainyday For me it just complained that something changed and that the container will still use the old volume/storage, so it wouldn't delete the old db right away. It would however require manual tweaking I guess.

However, the current situations means that people running off of the default docker-compose.yml will lose their databases if they `docker-compose down` or reboot, and I don't think that's a lot better :/

Pajn:
See this https://github.com/tootsuite/mastodon/pull/1251 and the linked original PR.
Upgrades needs to be solved

Gargron:
> wont this nuke everyone's existing databases if they happen to pull this without knowing and run docker-compose up?

This is what happened to @JantsoP which led to #1251. So changing the default is now not possible. However we added warnings to the README and a commented out volume definition in the docker-compose.yml. I think that's the best non-breaking solution to this problem for now.

PR #1380, opened 2017-04-09
Edit nginx config to add extra security headers.

progval:
Content-Security-Policy to mitigate XSS.
X-Content-Type-Options to prevent use of forged media to exploit browsers
X-Frame-Options because I don't see how it could be a good idea to allow embedding Mastodon in an iframe.

I have been using this config on my 1k users instance for a week now; everything looks fine.

### DIFF
--- a/docs/Running-Mastodon/Production-guide.md
+++ b/docs/Running-Mastodon/Production-guide.md
@@ -39,6 +39,9 @@ server {
   root /home/mastodon/live/public;
 
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  add_header Content-Security-Policy "default-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self' https://example.com/ wss://example.com/; upgrade-insecure-requests";
+  add_header X-Content-Type-Options nosniff;
+  add_header X-Frame-Options DENY;
 
   location / {
     try_files $uri @proxy;
@@ -81,6 +84,8 @@ server {
   error_page 500 501 502 503 504 /500.html;
 }
 ```
+
+Make sure you remplace all instances of `example.com`.
 
 ## Running in production without Docker
 
--- a/docs/Running-Mastodon/Production-guide.md
+++ b/docs/Running-Mastodon/Production-guide.md
@@ -85,7 +85,7 @@ server {
 }
 ```
 
-Make sure you remplace all instances of `example.com`.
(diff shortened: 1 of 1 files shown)


docs/Running-Mastodon/Production-guide.md
```diff
@@ -82,6 +85,8 @@ server {
 }
 ```
 
+Make sure you remplace all instances of `example.com`.
```
jemus42: Typo, should probably be "replace".

ProgVal: Fixed, thanks

mkody:
I might ask, will images and videos hosted on insecure instances (because they don't or can't use TLS) still be able to load?

Gargron:
`X-Content-Type-Options` and `X-Frame-Options` are already set by Rails. There *is* an embeddable controller, you can embed toots on other pages using OEmbed.

ProgVal:
@mkody No. But the rest of the config won't work for them either.

@Gargron Fixed.

hrefhref:
>I might ask, will images and videos hosted on insecure instances still be able to load?

Maybe a little bit off-topic, but shouldn't media from other instances be proxied by the local one ? Like what github is doing when you link images in comments. It allows to ensure it's served over SSL, and avoid leak of client IP addresses

ProgVal:
@hrefhref Isn't it already?

docs/Running-Mastodon/Production-guide.md
```diff
@@ -39,6 +39,7 @@ server {
   root /home/mastodon/live/public;
 
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  add_header Content-Security-Policy "default-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self' https://example.com/ wss://example.com/; upgrade-insecure-requests";
```
Angristan: That's what I use : 
`default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self' wss://example.com;`

It's not a good idea to put `unsafe-inline` by default

ProgVal: Indeed.
Fixed, thanks.

mjankowski:
If still relevant, can you submit this as a PR against the docs repo?

Issue #1355, opened 2017-04-09
Automatically reserve some accounts at instance creation


wxcafe:
I believe some accounts (support@, help@, admin@, root@, ...) should be reserved when initializing the database (and maybe not counted in the user count?), and possibly aliased to the actual admin's account.

Please contribute if you have ideas of other @'s to reserve

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

James-Firth:
I think this is a pretty good idea to have some consistency and avoid people pretending they are the admin of a particular instance.

I would throw in:
* administrator (just to double-down on the admin idea)
* webmaster (for historical sake)

Tempted to put security@ but I think anyone reporting security issues would know how to find the email for the domain or post a bug here/message Gargron directly.

If possible the support/help accounts would also be bots to link to info pages (if help pages are compiled in the future)

ngerakines:
Implemented in https://github.com/tootsuite/mastodon/pull/1382 .

hikaru-light:
I suggest reserving these words "system", "host", "official", "owner", "master", "register", "staff".
At least, I added them as reserved usernames in developing another service.
Please let me know your opinions.

Gargron:
Implemented in #3566

dannythehacker:
i am so mad i want to set my user name to admin because i own the server how do i bypass the reserved bull shit i am so mad

dannythehacker:
and i cant even set the username to admin in command line what the f i understand at the sign up screen

trwnh:
edit config/settings.yml?

dannythehacker:
okay sorry

PR #1382, opened 2017-04-09
Added support for configurable reserved usernames

ngerakines:
This PR includes changes to the account model to support the `reserved_usernames` validation check for local accounts. With it, a list of usernames can be added to settings.yml as "reserved" for the purpose of preventing people from registering accounts with those usernames.


config/settings.yml
```diff
@@ -24,6 +24,9 @@ defaults: &defaults
   interactions:
     must_be_follower: false
     must_be_following: false
+  reserved_usernames:
+    - 'admin'
+    - 'support'
```
ngerakines: I added "admin" and "support" as default entries because those are two account names that I wouldn't want people to steal.

wxcafe: Please add the other ones that were discussed on #1355

ngerakines: Done.

Fastidious (commented):
How those get unreserved?

ngerakines:
Removing the configuration entry will allow the account to be created.

wxcafe:
It would be nice to be able to alias these accounts to another one, but this is a much bigger feature. I'll merge this when we know why the CI doesn't want to build (it's apparently not related to this tho) and I'll change the issue to reflect the change in scope

ngerakines:
Yeah, I saw the comment about aliasing, but decided to hold off on that one. A temporary solution for web visitors could be to have nginx/httpd rewrite rule samples.

Locally `rake` and spec file execution all looks good. I'm not sure what is going on with travis.

Thanks for taking a look and I'm glad it is useful.

ngerakines:
If there is anything that I can do to help this move along, please let me know. Looks like there is quite the backlog right now.

app/models/account.rb
```diff
@@ -8,7 +8,7 @@ class Account < ApplicationRecord
 
   # Local users
   has_one :user, inverse_of: :account
-  validates :username, presence: true, format: { with: /\A[a-z0-9_]+\z/i, message: 'only letters, numbers and underscores' }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, if: 'local?'
+  validates :username, presence: true, format: { with: /\A[a-z0-9_]+\z/i, message: 'only letters, numbers and underscores' }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, exclusion: { in: (Setting.reserved_usernames ||= []) }, if: 'local?'
```
bradurani: `Setting.reserved_usernames ||= []`
should be
`Setting.revserved_username || []`

bradurani (changes requested):

ashfurrow:
Thanks for the PR @ngerakines – looks good! Could you pull latest from master, review the changes, and address @bradurani 's comment above? Thanks again for your contribution!

app/models/account.rb
ngerakines: Done!

ngerakines:
The branch has been updated from current master and the change request applied.

ashfurrow:
The code looks good, but there's a CI error. I've not seen it before, anyone have any ideas?

config/settings.yml
hikaru-light: LGTM:cake:

hikaru-light:
I have another way because I have implemented reserving usernames in another service.
May I push my branch?

Gargron:
The failure is because Setting.reserved_usernames is evaluated at class load time, which is before anything else, and in this case before the migrations finish running, so it runs a SELECT against an incomplete database.

PR #1383, opened 2017-04-09
Update List-of-Mastodon-instances.md, add mastodon.engineering

skhameneh:


JantsoP:
As the file says

We are no longer maintaining this list as instances are popping up too quickly for using GitHub to be a tenable system for tracking them. Please standby while we work on another solution

PR #1387, opened 2017-04-10
Update Production-guide.md

bkeej:
a warning to use a swapfile for when installing dependencies on a small server

### DIFF
--- a/docs/Running-Mastodon/Production-guide.md
+++ b/docs/Running-Mastodon/Production-guide.md
@@ -88,6 +88,8 @@ It is recommended to create a special user for mastodon on the server (you could
 
 ## General dependencies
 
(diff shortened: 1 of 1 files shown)


skhameneh:
The server uses close to 512mb minimal at runtime; if you only have 512mb on your server it won't stay up long.  Maybe this message should be revisited to better convey requirements.

balamkej:
Cool! yeah, i really think there should be something in the guide about hardwear requirements. Maybe we can write up a section? I got an instance up and running on very minor hardware, but there were many mysterious hurdles. It could be nice to have minimum hardwear requirements + info about workarounds if you want to be daring.

Gargron:

Issue #1233, opened 2017-04-08
Emoji-dialog layout and icon size issue


ramlmn:
The emojione png images look way too small on a desktop. They are held inside a container of size 32x32px, padding 5px and are of size 16x16px and are positioned top left. This is a small layout quirk, making the emojies look weird and too small to distinguish.

And this is same for the `tab-bars`, the fa icons are of size 12px which is way too small considering accessibility. I needs to be atleast 14 or 16px.

![image](https://cloud.githubusercontent.com/assets/12368291/24826205/ee588b8a-1c4d-11e7-8cda-2cb8009495b2.png)
![image](https://cloud.githubusercontent.com/assets/12368291/24826281/9c59671c-1c4f-11e7-9b0d-8ba7b8082aca.png)

*****

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

ramlmn:
Closed by #1403

PR #1403, opened 2017-04-10
Fix icon and emoji sizes

ghost:
Fixes icons and emoji size issues described in #1233


ramlmn:
UI changes before and after.
emoji-before
emoji-after

header-before
header-after

header-before-2
header-after-2

Gargron (approved):

PR #1406, opened 2017-04-10
Dutch language files

meneer:


config/locales/doorkeeper.nl.yml
```diff
@@ -0,0 +1,130 @@
...
+    layouts:
+      admin:
+        nav:
+          oauth2_provider: 'OAuth2 Provider'
+          applications: 'Applicaties'
+          home: 'Home'
+application:
+        title: 'OAuth autorisatie vereist'
+    scopes:
+      follow: volg, blokkeer, deblokkeer en stop volgen accounts
```
Gargron: I'm not sure what this ^M is supposed to be.

Gargron (changes requested):

PR #1408, opened 2017-04-10
Dutch language added

meneer:


Gargron (changes requested):
I feel extremely uncomfortable with the entire file changing without any obvious changes. Are you working from Windows and committing Windows line breaks? Please don't do that, and please do not spread out your work on one thing between 5 different pull requests. It makes it much harder to review! :(

PR #1410, opened 2017-04-10
Dutch language added

meneer:


xvilo:
Thanks Mate! Was just looking into translating it.

blackle:
why are these changes spread out over multiple branches?

meneer:
> Pick your reaction
I have no clue about the inner workings of github. Will try to find sme more info to prevent my multiple branches

yookoala:
You're creating / editing files directly in Github's editor?

yookoala:
Two issues to your patches.
1. Unnecessarily breaking commits into different branch.
2. Your editor is using Windows EOL. That blocks normal workflow in code diff-ing.

PR #1415, opened 2017-04-10
Improve scrolling behaviour

### LINKED ISSUE #1412: Right-hand margin missing
#1312 removed the margin from the right-hand side of the desktop UI, entirely
screen-shot

* * * *

ticky:
- This switches from using `margin` to using `padding`.
  - This means column backgrounds need to be painted *inside* the scrollable area rather than on the background of the `column`.
  - As a side-effect, it seems to provide a _very minor_ boost to drawing performance, at the cost of a small amount of obviousness!
  - The padding also means the UI can scroll right to the end horizontally.
- This also fixes an issue where small browsers would encounter ~10px of horizontal scrolling.
- A few duplicate/overloaded CSS specifications were removed.
  - The ones that the browser used are the ones which stayed!
- The empty column indicator’s contents were set to flex to the centre so it can have a background applied, and could be squashed correctly by the column settings.
- Fixes #1412 

**Screenshots:**
320px wide:
screen shot 2017-04-10 at 8 45 39 pm

340px wide:
screen shot 2017-04-10 at 8 46 43 pm

Desktop:
screen shot 2017-04-10 at 8 47 14 pm

P.S. The “clear notifications” button remains in the right spot on mobile screens (it required the addition of a breakpoint-aware CSS class)
screen shot 2017-04-10 at 8 47 37 pm

**To-do:**
- [x] Fix horizontal scrolling margin
- [x] Make sure “clear column” button is still in the right place
  - [x] On desktop
  - [x] On mobile
- [x] Check column settings opening behaviour isn’t adversely affected

### DIFF
--- a/app/assets/javascripts/components/features/notifications/components/clear_column_button.jsx
+++ b/app/assets/javascripts/components/features/notifications/components/clear_column_button.jsx
@@ -4,16 +4,6 @@ const messages = defineMessages({
   clear: { id: 'notifications.clear', defaultMessage: 'Clear notifications' }
 });
 
-const iconStyle = {
-  fontSize: '16px',
-  padding: '15px',
-  position: 'absolute',
-  right: '48px',
-  top: '0',
-  cursor: 'pointer',
-  zIndex: '2'
-};
-
 const ClearColumnButton = React.createClass({
 
   propTypes: {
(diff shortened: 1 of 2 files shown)


ticky:
Updated, having verified the behaviour in all environments I can find it affects. Now with screenshots and rationale! :sparkles:

Gargron (approved):

PR #1429, opened 2017-04-10
Allow typing a toot while an image uploads

ticky:
This patch stops disabling the toot text field when an image is uploading. Instead, you can type to your heart's content and when the image uploads it'll append the image URL, and restore the position of your cursor to wherever it was in the toot text just prior. Effectively, the image URL is appended to the toot, and typing is not interrupted at all! ✨

### DIFF
--- a/app/assets/javascripts/components/features/compose/components/compose_form.jsx
+++ b/app/assets/javascripts/components/features/compose/components/compose_form.jsx
@@ -83,11 +83,23 @@ const ComposeForm = React.createClass({
     this.props.onChangeSpoilerText(e.target.value);
   },
 
+  componentWillReceiveProps (nextProps) {
+    // If this is the update where we've finished uploading,
+    // save the last caret position so we can restore it below!
+    if (!nextProps.is_uploading && this.props.is_uploading) {
+      this._restoreCaret = this.autosuggestTextarea.textarea.selectionStart;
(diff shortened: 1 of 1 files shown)


Gargron:
Hmmm, I'm having trouble confirming this. Locally images upload too quickly to be able to type at the same time for long enough, but I tried and I got some letters in the middle of the inserted URL. Could you give me some guidance on how you tested it?

Gargron:
Right, I can see that it does restore the caret position correctly at least if you aren't typing. There is probably a race condition if you type too fast at the very end of the upload. But it's good enough,

Gargron (approved):

ticky:
There’s actually an option in the Chrome dev tools (and I suspect in others!) to simulate a slow network connection. I set Chrome to GPRS speeds for testing this!

screen shot 2017-04-11 at 9 41 02 am

As far as race conditions go, this patch is using the React lifecycle to its advantage.

`componentWillReceiveProps` is called before the component renders, but after it’s been alerted to the new props. `componentDidUpdate` is called _immediately_ after the component renders to the DOM based on the new props.

There should be a very, very, _very_ small gap of time between them, if any. Most of the time between them should be taken up by a browser repaint, which will take care of passing keypresses after the DOM is ready again anyway.

PR #1445, opened 2017-04-10
Pagination improvements

mjankowski:
I noticed some duplication with how the various pages were passing around `pagination_options` and the controllers duplicating what could have been a global setting.

Changes include:

- Swap out `will_paginate` for the more actively maintained `kaminari` pagination gem
- Set as global options what used to be in the `pagination_options` helper
- Delete the AccountHelper, since it has nothing else in it
- Update views and stylesheets to account for the (slight) markup difference introduced
- Update some locales to remove what are now unused keys (fortunately, they all had the same exact value here)


app/controllers/accounts_controller.rb
```diff
@@ -35,11 +35,11 @@ def unfollow
   end
 
   def followers
-    @followers = @account.followers.order('follows.created_at desc').paginate(page: params[:page], per_page: 12)
+    @followers = @account.followers.order('follows.created_at desc').page(params[:page]).per(12)
   end
 
   def following
-    @following = @account.following.order('follows.created_at desc').paginate(page: params[:page], per_page: 12)
+    @following = @account.following.order('follows.created_at desc').page(params[:page]).per(12)
```
JoelQ: Thoughts on extracting this `12` to a constant?

app/controllers/accounts_controller.rb
```diff
@@ -35,11 +35,11 @@ def unfollow
   end
 
   def followers
-    @followers = @account.followers.order('follows.created_at desc').paginate(page: params[:page], per_page: 12)
+    @followers = @account.followers.order('follows.created_at desc').page(params[:page]).per(12)
```
JoelQ: Thoughts on extracting this `12` to a constant?

Gargron (approved):

PR #1448, opened 2017-04-10
Update List-of-Mastodon-instances.md

tuxx:
Added my public instance


mkody:
As the file says: 

PR #1455, opened 2017-04-10
Fix issue with missing emojify class in views

mjankowski:
I noticed that on the landing strip of user and status pages when signed out, and within the simple status area for the "boosted by" person, users who have `:emoji:` style naming were not being converted to actual emoji, but were just rendering the string as entered.

Before:

emojify before

After:

emojify after


Gargron (approved):

Fastidious:
Will this apply if the Emoji is after the username?

PR #1458, opened 2017-04-10
Added noagendasocial.com

CraigDamlo:
Added a new instance that has open registration.


mkody:
As the file says:

Issue #1044, opened 2017-04-06
Enabling CDN_HOST in .env.production using Docker not working


sys0dm1n:
[After creating account with the CDN enabled, everything works fine, once I activate my account as admin and relogin I get a 500 error:
```
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]   User Load (2.1ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["id", 1], ["LIMIT", 1]]
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]   Account Load (1.7ms)  SELECT  "accounts".* FROM "accounts" WHERE "accounts"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]   Doorkeeper::Application Load (1.0ms)  SELECT  "oauth_applications".* FROM "oauth_applications" WHERE "oauth_applications"."superapp" = $1 ORDER BY "oauth_applications"."id" ASC LIMIT $2  [["superapp", true], ["LIMIT", 1]]
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]    (0.3ms)  BEGIN
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]   Doorkeeper::AccessToken Exists (1.5ms)  SELECT  1 AS one FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."token" = $1 LIMIT $2  [["token", "12ff4d3707bc6cb1349924ccd665ad2c51be748f2cc6eaeb45d97teg57f7f6ad"], ["LIMIT", 1]]
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]   SQL (0.8ms)  INSERT INTO "oauth_access_tokens" ("resource_owner_id", "token", "created_at", "scopes") VALUES ($1, $2, $3, $4) RETURNING "id"  [["resource_owner_id", 1], ["token", "12ff4d3707bc6cb1349924ccd665ad2c51be748f2cc6eaeb45d97teg57f7f6ad"], ["created_at", 2017-04-06 09:19:23 UTC], ["scopes", "read write follow"]]
[e250e0dc-4609-42a7-86cb-d110e15fb7e6]    (53.2ms)  COMMIT
[...]

PR #1459, opened 2017-04-10
CDN_HOST routing_helper fix

josh04:
Two minor fixes to the behaviour of CDN_HOST:

- Currently, if you enter a CDN_HOST and build the Docker image, the application will throw the following error:

```
method=GET path=/web/accounts/166 format=html controller=HomeController action=index status=500 error='ActionView::Template::Error: undefined method 'protocol' for nil:NilClass' duration=35.83 view=0.00 db=27.18

 ActionView::Template::Error (undefined method 'protocol' for nil:NilClass):
     1: object false
     2:
     3: node(:meta) do
     4:   {
     5:     access_token: @token,
     6:     locale: I18n.locale,

 app/helpers/routing_helper.rb:15:in 'full_asset_url'
 app/views/api/v1/accounts/show.rabl:7:in 'block in eval_source'
 app/views/home/initial_state.json.rabl:3:in '_app_views_home_initial_state_json_rabl___4241867769042566489_47336508132160'
 app/views/home/index.html.haml:4:in 'block in _app_views_home_index_html_haml__1697788668079016006_47336538039340'
 app/views/home/index.html.haml:1:in '_app_views_home_index_html_haml__1697788668079016006_47336538039340'
 app/controllers/concerns/localized.rb:24:in 'block in set_locale'
 app/controllers/concerns/localized.rb:23:in 'set_locale'
```

This does not occur if the provided CDN_HOST has a protocol specified, so I've added one to .env.production.sample.

- In addition, once the page loads, assets which are loaded with full_asset_path such as avatars will have an incorrect URL, e.g. http://mastodon.social/http://static.mastodon.social/assets/avatar.png. I've tweaked routing_helper.rb in a way which appears to fix this on my instance.

Apologies if I've made any obvious slip-ups.

resolves issue #1044

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -36,8 +36,8 @@ SMTP_LOGIN=
 SMTP_PASSWORD=
 SMTP_FROM_ADDRESS=notifications@example.com
 
-# Optional asset host for multi-server setups
-# CDN_HOST=assets.example.com
+# Optional asset host for multi-server setups (inc. protocol)
+# CDN_HOST=http://assets.example.com
 
 # S3 (optional)
 # S3_ENABLED=true
--- a/app/helpers/routing_helper.rb
+++ b/app/helpers/routing_helper.rb
@@ -12,6 +12,6 @@ module RoutingHelper
   end
 
   def full_asset_url(source)
-    Rails.configuration.x.use_s3 ? source : File.join(root_url, ActionController::Base.helpers.asset_url(source))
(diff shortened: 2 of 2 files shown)


.env.production.sample
```diff
@@ -36,8 +36,8 @@ SMTP_LOGIN=
 SMTP_PASSWORD=
 SMTP_FROM_ADDRESS=notifications@example.com
 
-# Optional asset host for multi-server setups
-# CDN_HOST=assets.example.com
+# Optional asset host for multi-server setups (inc. protocol)
+# CDN_HOST=http://assets.example.com
```
Gargron: Very strange. I just checked, and this variable works without the protocol on mastodon.social.

josh04:
Having upgraded to v1.2.2, this has been fixed by some other change. Closing the pull request!

PR #1468, opened 2017-04-10
Trim long username on public grid card

### LINKED ISSUE #1426: Trim long usernames (visually)
I character limit on the showed description on a user must be set. [See](https://naganu.com/users/david/following) the overflow:

screen shot 2017-04-10 at 08 45 38

R0ckweb:
Fix #1426


riking:
I think this would be better as a CSS `overflow: ellipsis`.

Fastidious:
This isn’t the right fix (not even CSS), as it isn’t the problem. See my comment at #1426.

PR #1470, opened 2017-04-10
Remove order prior to .find_in_batches

mjankowski:
The `Status` class has a default order on it, so when this query gets built and
gets all the way to `find_in_batches` there is an order already there.

When `find_in_batches` is run it discards any existing order on the query, and
emits a warning to the logs if there is one there.

This change removes the order prior calling `find_in_batches`, which will stop
the logged warning from occurring as well.

Resolves https://github.com/tootsuite/mastodon/issues/1450

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -66,7 +66,7 @@ class FeedManager
     timeline_key = key(:home, into_account.id)
     oldest_home_score = redis.zrange(timeline_key, 0, 0, with_scores: true)&.first&.last&.to_i || 0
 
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Vertux:
I am not sure if these issues are related, but I get the cron info:

> Scoped order is ignored, it's forced to be batch order.

when I run the cron job:

`mastodon:media:remove_remote`

I guess it falls under #1604 - I am running the current v2.4.3 btw.

nightpool:
@Vertux that message is just a warning and it has nothing to do with this, i believe it's from the elasticsearch gem

Vertux:
@nightpool Thank you for the quick response, I was afraid, that something goes wrong, now I'm reassured.

PR #1479, opened 2017-04-10
Add email validation

### LINKED ISSUE #1286: Broken email validation in instances.
A user tried to create an account with "username@domain" missing the ".tld" part and surprisingly their account was created on my instance.

They of course now cannot create a valid account as their desired handle has been taken, suggesting that the basic email validation is broken.

On that note, any guidance on helping me edit the db at backend would be a sure help.

Thanks a lot.

* * * *

R0ckweb:
Fix #1286

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -149,9 +149,12 @@ Devise.setup do |config|
   # Range for password length.
   config.password_length = 8..72
 
-  # Email regex used to validate email formats. It simply asserts that
-  # one (and only one) @ exists in the given string. This is mainly
-  # to give user feedback and not to assert the e-mail validity.
+  # Email regex used to validate email formats. It asserts that
+  # - the localpart contain only word character - or dot
+  # - the domain contain only lower letter, number, - or dot (and don't start with dot).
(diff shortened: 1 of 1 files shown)


ericblade:
I can't comment on wether this is desireable or not, not knowing where this particular function is used, but the comment directly above it specifically states that that regex is not for checking that it's a valid email, only that it has an @ in it somewhere.

SO, if this change doesn't break anything, at the very minimum, it would need to have that comment altered.

R0ckweb:
@ericblade You're absolutely right !

expenses:
Seeing as we're validating emails anyway through confirmation, I think it's best to leave email checking to the client-side.

R0ckweb:
I think back-end validation is always a good idea.
Let client insert wrong values in DB make me nervous ^^

But, feel free to ignore and close this PR :)

Gargron:
Here is what an actual valid e-mail regex looks like: http://www.ex-parrot.com/pdw/Mail-RFC822-Address.html

The sanest thing to do is check if there's a @, and send a confirmation e-mail. Sorry. Thank you for your effort on this, anyways.

Issue #1456, opened 2017-04-10
Rake action for creating a user


codl:
#146 mentions manually creating user accounts in the administrative UI, which sounds very nice, but until that's done it would be nice to have a rake action to create a user and send them a password reset email, as a temporary solution for people who want to close registration and invite people manually onto their server

(I was going to do this myself but I don't know activerecord well enough & ended up spending a couple hours battling with it to no avail 😓)

wxcafe:
If someone knows how to do rake, this should probably be very easy - it's literally two inserts in the postgres DB

Gargron:
Fixed by #1482

PR #1482, opened 2017-04-10
 #1456 Added rake task to add a user.

debanshuk:
Pull request for issue: https://github.com/tootsuite/mastodon/issues/1456

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -19,6 +19,37 @@ namespace :mastodon do
     puts "User #{email} confirmed."
   end
 
+  desc 'Add a user by providing their email, username and initial password.' \
+       'The user will receive a confirmation email, then they must reset their password before logging in.'
+  task add_user: :environment do
+    print 'Enter email: '
+    email = STDIN.gets.chomp
+
+    print 'Enter username: '
+    username = STDIN.gets.chomp
+
+    print 'Create user and send them confirmation mail [y/N]: '
+    confirm = STDIN.gets.chomp
+    puts
+
+    if confirm.casecmp?('y')
+      password = SecureRandom.hex
+      user = User.new(email: email, password: password, account_attributes: { username: username })
+      if user.save
+        puts 'User added and confirmation mail sent to user\'s email address.'
+        puts "Here is the random password generated for the user: #{password}"
+      else
+        puts 'Following errors occured while creating new user:'
+        user.errors.each do |key, val|
+          puts "#{key}: #{val}"
+        end
+      end
+    else
+      puts 'Aborted by user.'
+    end
+    puts
+  end
+
   namespace :media do
     desc 'Removes media attachments that have not been assigned to any status for longer than a day'
     task clear: :environment do


lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,39 @@ namespace :mastodon do
     puts "User #{email} confirmed."
   end
 
+  desc 'Add a user by providing thier email, username and initial password. ' \
```
riking: nit: their

lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,39 @@ namespace :mastodon do
...
+       'The user would recieve a confirmation mail and then would be able to reset his password for logging in.'
```
riking: Grammar nits on this, but instead:

If you're expecting them to reset their password, use `SecureRandom.hex` for the password instead of prompting the admin.
Then replace the Enter Password loop with a Confirm Y/N.

riking (changes requested):

debanshuk:
@riking , Done all review changes. Please re-review.

lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,36 @@ namespace :mastodon do
     puts "User #{email} confirmed."
   end
 
+  desc 'Add a user by providing their email, username and initial password.' \
+       'The user would receive a confirmation mail and then would be able to reset his password for logging in.'
```
riking: those grammar nits I left out:

active voice, will instead of would

> The user will receive a confirmation email, then must reset their password before logging in.

lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,36 @@ namespace :mastodon do
...
+
+    print 'Create user and send them confirmation mail (y/n): '
+    confirm = STDIN.gets.chomp
+    puts
+
+    if confirm == 'y'
+      user = User.new(email: email, password: SecureRandom.hex, account_attributes: { username: username })
+      if user.save
+        puts 'User added and confirmation mail sent to user\'s email.'
+        puts 'Note: After confirming their email, the user would need to reset their password to be able to login.'
```
riking: same, will instead of would

debanshuk:
@riking, Done.

riking (approved):
LGTM, assuming you've tested?

debanshuk:
@riking, Yes, I have tested the functionality by adding a user and logging-in using that user (after confirming email and resetting password).

ashfurrow (approved):

ashfurrow:
I don't see any blockers here, would someone be available to merge?

lib/tasks/mastodon.rake
```diff
@@ -33,6 +33,36 @@ namespace :mastodon do
...
+
+    print 'Create user and send them confirmation mail (y/n): '
+    confirm = STDIN.gets.chomp
+    puts
+
+    if confirm == 'y'
+      user = User.new(email: email, password: SecureRandom.hex, account_attributes: { username: username })
+      if user.save
+        puts 'User added and confirmation mail sent to user\'s email.'
+        puts 'Note: After confirming their email, the user will need to reset their password to be able to login.'
```
Gargron: Just save the `SecureRandom.hex` into a variable and output it here. The admin will probably send the password to the user in question. This is pretty common practice for first logins (you get a random password, have to change it immediately). Better than having to reset password to get into the account.

lib/tasks/mastodon.rake
```diff
@@ -33,6 +33,36 @@ namespace :mastodon do
...
+    email = STDIN.gets.chomp
+
+    print 'Enter username: '
+    username = STDIN.gets.chomp
+
+    print 'Create user and send them confirmation mail (y/n): '
+    confirm = STDIN.gets.chomp
+    puts
+
+    if confirm == 'y'
```
Gargron: I mean, it's kinda more common to have these as `Y/n`, I'm not sure why but it would look more right here.

Gargron (changes requested):
This is okay, I was hesitant to merge this because I want to add an admin UI page for adding users, but on the other hand there is little harm in having two ways of doing it.

lib/tasks/mastodon.rake
debanshuk: @Gargron, As stated in this answer '[Y/n]' mean 'y' is default option, ie, if no character is entered and enter key is pressed, 'y' would be selected. Do you want me to change the current behavior (of 'no' being the default option) and make 'yes' the default action?

lib/tasks/mastodon.rake
debanshuk: @Gargron, Sure, I'll make this change.

ashfurrow:
Hi @debanshuk – thanks again for the pull request. Any update on the changes that were requested? No rush of course!

lib/tasks/mastodon.rake
debanshuk: @Gargron, I an capitalizing 'n', as no is the default action. So it would look like '[y/N]'.

debanshuk:
@ashfurrow @Gargron, I have done the requested changes.

ashfurrow:
Cool, looks good!

lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,37 @@ namespace :mastodon do
     puts "User #{email} confirmed."
   end
 
+  desc 'Add a user by providing their email, username and initial password.' \
+       'The user will receive a confirmation email, then must reset their password before logging in.'
```
neliyacf: `then they`

lib/tasks/mastodon.rake
```diff
@@ -19,6 +19,37 @@ namespace :mastodon do
...
+    print 'Create user and send them confirmation mail [y/N]: '
+    confirm = STDIN.gets.chomp
+    puts
+
+    if confirm.casecmp?('y')
+      password = SecureRandom.hex
+      user = User.new(email: email, password: password, account_attributes: { username: username })
+      if user.save
+        puts 'User added and confirmation mail sent to user\'s email address.'
+        puts "Here is the random password generate for the user: #{password}"
```
neliyacf: typo `generated`

neliyacf:
LGTM 👍

debanshuk:
@Gargron , It's been quite some time since I had done all requested changes. Could you re-review the issue and approve it?

Gargron (approved):

ressolu:
Does anyone know if a user can be added via API?

Gargron:
@ressolu No. Exposing e-mail address and password to a 3rd party app would not be safe, therefore the API does not have a method for registering.

PR #1484, opened 2017-04-10
[css] replace login box shadow with .wrapper background

dunn:
This removes the background and drop shadow from the login form and applies the same background (though without the drop shadow) to the `.wrapper` element.  Here's what it looks like:

screen shot 2017-04-10 at 3 37 24 pm

It's much more noticeable on instances with a custom background image.  Without this change, using a lighter image makes for very low contrast between it and the "about" text:
screen shot 2017-04-10 at 3 37 54 pm

But with this change it's perfectly usable:
screen shot 2017-04-10 at 3 38 22 pm

The change to the `.form-container` element is also to make light backgrounds usable:
screen shot 2017-04-10 at 3 39 37 pm

### DIFF
--- a/app/assets/stylesheets/about.scss
+++ b/app/assets/stylesheets/about.scss
@@ -1,10 +1,10 @@
 .about-body {
   .wrapper {
-    max-width: 600px;
-    margin: 0 auto;
+    background: rgba(darken($color1, 7%), 0.5);
     color: $color3;
-    padding-top: 50px;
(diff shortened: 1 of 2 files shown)


Gargron:
I do not officially support changing of the background image etc. If someone wants to customize their look, they can do that and do this, but I do not agree with it as a default look, sorry. Thank you for your time and effort anyway.

PR #1493, opened 2017-04-11
Improve emoji picker look

Gargron:
screenshot from 2017-04-11 01-59-37

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -667,6 +667,12 @@ a.status__content__spoiler-link {
     left: 8px;
   }
 
+  &.light {
+    &:before {
+      border-color: transparent transparent $color5 transparent;
+    }
+  }
+
   & > ul {
     list-style: none;
     background: $color2;
@@ -684,7 +690,7 @@ a.status__content__spoiler-link {
     }
 
     & > .emoji-dialog {
-      left: -249px;
+      left: -210px;
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components.scss
```diff
@@ -1575,11 +1597,15 @@ button.active i.fa-retweet {
...
   .emoji-search-wrapper + .emoji-categories-wrapper {
-    top: 83px;
+    top: 93px;
   }
 
   .emoji-row .emoji:hover {
     background: lighten($color2, 3%);
+
+    img, svg {
+      transform: translateZ(0) scale(1.2);
```
riking: the translateZ here is fishy - do you really need a separate render plane for EVERY image?

riking (changes requested):
Just a perf nit in the CSS.

app/assets/stylesheets/components.scss
Gargron: Hmm. I think I was misinformed about the performance implications - I believed that `translateZ(0)` activates GPU to make animations smoother (and I am planning to add a bouncy animation to this at some point.

riking:
Well, you can add the translateZ when it's actually used, right?
And yeah it creates a separate GPU context for the element which makes
animations easier.

Gargron:
@riking Added animation!

krainboltgreene:
This is so hecking legit.

krainboltgreene (approved):

PR #1498, opened 2017-04-11
Allow export of mutes list

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/1485


mjankowski:
This is not useful w/out an import as well ... closing for now, will reopen when import exists.

PR #1500, opened 2017-04-11
Fix broken mobile CSS and inconsistent margins

ghost:
Fix CSS issues causing horizontal scrolling on mobile and columns being mispositioned and not expanding to screen height on large monitors.


rainyday:
will update and re-push

PR #1502, opened 2017-04-11
ruby version was updated in .ruby-version but not Vagrant

halkeye:
Updated the vagrant file so it pulls the version from ruby versions file, this should prevent the same issue in the future.


halkeye:
:D I was pulling down the changes i got merged yesterday and my vagrant wouldn't boot

Gargron (approved):

PR #1506, opened 2017-04-11
Allow using .env to configure database in development

ghost:
Allows configuring postgres host in development (useful for people using windows subsystem for linux which can't run postgres). Also adds sourcemaps for js debugging when in development.


config/application.rb
```diff
@@ -45,6 +45,9 @@ class Application < Rails::Application
 
     config.middleware.use Rack::Attack
     config.middleware.use Rack::Deflater
+    
+    # Enable sourcemaps for debugging during development
+    config.browserify_rails.source_map_environments << "development"
```
bradurani: I feel like this would be better in a separate pull request

bradurani (changes requested):

ashfurrow (changes requested):
I've made these changes locally too, looks good 👍

Agree about moving the source map changes int their own PR, can you amend?

mecab:
I think it is good PR and I would need it to be merged. If @rainyday is busy, can I raise PR (without the change about source map) instead?

ashfurrow (approved):
Looks great!

mecab:
Cool, thanks!!

mjankowski (changes requested):
Could people get the same benefit by using `ENV['DATABASE_URL']` (which rails already supports) - or is there something in this change which that doesn't provide and needs custom env?

ashfurrow:
Hmm, I didn't know about `ENV['DATABASE_URL']`, if using it is compatible with this then maybe we should document that instead?

Gargron:
DATABASE_URL env variable is preferable for this, you can set any options in it and it will override other settings. I don't want to have too many custom vars. Sorry

Issue #621, opened 2017-02-11
confirm dialog for boosts


hoodiek:
if a yes/no confirm is possible, it would be extremely useful to be able to require a confirm when boosting things. whether that be all posts or just a setting to turn on isn't a big deal to me. i've seen many complaints for this

hoodiek:
referenced in #620

mitchellurgero:
Agreed - I will see if I can learn ruby just for this. lol

blackle:
fixed with referenced PR

PR #1507, opened 2017-04-11
Add boost confirm modal

blackle:
Fix for #621 

It looks like this: 
screenshot of modal

the modal is defined to be 480px but no wider than 90% of the screen size, and also the toot preview cannot be taller than 60% of the screen size (when this happens it becomes scrollable)

also pointer events for the toot preview are disabled thru css so the user doesn't click on, say, the username and get taken to their profile _behind_ the modal


yiskah:
Does this have the ability to use alt-click to bypass the modal?

blackle:
woops, forgot to implement that part
I can either modify this commit to add that or do it in a later one

yiskah:
If it's easy enough to add to this pull i think it would be ideal since it would be tedious for a lot of people to have to confirm every time

rainyday:
Can there be a way to turn this off? This just makes it more work to do anything and as someone who suffers from RSI I really don't like the idea of having to hold an extra key just to boost.

ghost (changes requested):
svg should be in its own file not crammed into the stylesheet

blackle:
the svg has to be in the stylesheet so admins can edit the colour scheme through variables.scss

blackle:
Ok, I added the alt+click to bypass the dialog, and also a setting in user preferences to disable it altogether

rainyday:
Is there a reason you can't just use css instead of a svg? This kind of thing is bad for performance and long term maintainability.

blackle:
I don't see how this is relevant to the current PR, I didn't add that in this commit, just changed the rules so it didn't apply to the boost button

blackle:
fuck, accidentally closed

Gargron (approved):

PR #1508, opened 2017-04-11
views/about: use Setting.site_title instead of hardcoding

dunn:


dunn:
Failure is from the API tests, might just be a flaky spec? https://travis-ci.org/tootsuite/mastodon/builds/220802750#L1349

Gargron (approved):

PR #1509, opened 2017-04-11
Lay out pages 2 through 5, add images, style modals

yiskah:
There are now images for pages 2 through 5 and appropriate style rules in components.scss for styling the images and some text changes. This is "version one."  At some point later we could replace the static images with some sort of illustrative animation, change or add pages, or slightly modify the copy to avoid orphan words on certain screen sizes. For now I think this is needed sooner than later and later fixes will be easy.

With this push the onboarding modals are *nearly* ready to be deployed, except for a bug which causes the modals to not launch.

something about `"values={{ admin: <Permalink href={admin.get('url')} to={`/accounts/${admin.get('id')}`}>@{admin.get('acct')}</Permalink> }} />"` causes a "Uncaught TypeError: Cannot read property 'get' of undefined" errors in the console. I hunted around all the controllers and checked things out but I couldn't figure out what about this line isn't defined properly. Once this bug gets fixed we should be able to merge this into Master


Gargron:
I don't understand, this just worked for me, no syntax errors. Do I need to do something to get a syntax error that you encountered? It doesn't look like anything is missing from the modal, either

yiskah:
As @blackle pointed out, the error occurs if the instance does not have an admin specified.

yiskah:
OK So the current bug is:

The modal does not appear if the instance does not have an admin/site contact. So I'm going to merge this code to the feature-onboarding branch and we can make one more pull to address that and then merge to master.

PR #1510, opened 2017-04-11
[css] reduce spacing between text elements on about pages

dunn:
The current margin is rather too large IMO.  Using awoo.space Guidelines as an example, this is what it's like currently:

screen shot 2017-04-10 at 8 07 53 pm

With the change, the spacing is more natural but not too tight:

screen shot 2017-04-10 at 8 08 27 pm

### DIFF
--- a/app/assets/stylesheets/about.scss
+++ b/app/assets/stylesheets/about.scss
@@ -64,7 +64,7 @@
   p, li {
     font: 16px/28px 'Montserrat', sans-serif;
     font-weight: 400;
-    margin-bottom: 26px;
+    margin-bottom: 12px;
 
     a {
       color: $color4;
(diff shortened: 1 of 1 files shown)


thoragio (changes requested):
@dunn Given that `em` units are not being used in the stylesheet, I suggest sticking with pixels and using a value like `14px` or `16px`.

dunn:
@thoragio is there a reason for avoiding ems?  I thought generally they were preferred over pixels as a unit.

thoragio:
@dunn It's not about avoiding them; it's about being consistent with the codebase and following the established convention. That being said, rems and ems don't really make much difference in many contexts.  See this article for details if you're interested: https://hackernoon.com/rems-and-ems-and-why-you-probably-dont-need-them-664b9ce1e09f

dunn:
Fair enough.  Switched from `.5em` to `12px`.

Gargron (approved):

PR #1511, opened 2017-04-11
views/about: move about_mastodon below site_description

dunn:
Information specific to the instance should be more prominent than
information about Mastodon in general.

Even when `site_description` isn't present, it's probably better to cut right to the chase with the signup form.

### DIFF
--- a/app/views/about/show.html.haml
+++ b/app/views/about/show.html.haml
@@ -19,7 +19,9 @@
     = image_tag 'logo.png'
     Mastodon
 
-  %p= t('about.about_mastodon').html_safe
+  - if @instance_presenter.site_description.present?
+    %h3= t('about.description_headline', domain: Rails.configuration.x.local_domain)
+    %p= @instance_presenter.site_description.html_safe
 
   .screenshot-with-signup
     .mascot= image_tag 'fluffy-elephant-friend.png'
@@ -38,6 +40,8 @@
           = link_to t('about.other_instances'), 'https://github.com/tootsuite/mastodon/blob/master/docs/Using-Mastodon/List-of-Mastodon-instances.md'
           ·
           = link_to t('about.about_this'), about_more_path
+
+  %p= t('about.about_mastodon').html_safe
 
   %h3= t('about.features_headline')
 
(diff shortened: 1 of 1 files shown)


yiskah:
I don't think this is very good for new users who do need to know what mastodon.social

dunn:
I'm not sure.  For the flagship instance, I think it looks fine, since there's no extra description (right hand image).  For others, "what is mastodon" will be below the fold (left hand):
awoo

If the order is going to remain how it is, then maybe the description could be shorter.  My main concern is that most instances look pretty much identical above the fold, so it's hard for them to indicate what is unique about _this_ instance.

Fastidious:
@dunn I prefer the layout you propose, indeed.

expenses:
I think it depends on how newcomers find Mastodon instances. If they first find an instance and _then_ learn about Mastodon, having the about text at the top is good. However, if they learn about Mastodon and _then_ look for an instance they want to join, the text becomes a little redundant.

I'm not sure which is better. I guess this 'Public Relations' kinda thing depends on the audience.

yiskah:
I think a middle ground could be good. The first thing should be "What is mastodon?" and then the second could be "What is $instance?". Followed by the rest of the descriptors. We want to try the features and signup form on one page. Right now this is the only landing page we have for people joining so it's unlikely someone will already know what mastodon is before joining. 

This layout makes sense for uses familiar with Mastodon. The current for users unfamiliar. Since familiar users will normally be logged in already, or know what to look for, I don't think we should de-prioritize software description.

dunn:
@yiskah a possible problem with that configuration would be that the signup form could fall beneath the fold (using witches.town as an example this time):
screen shot 2017-04-12 at 11 49 27 am

yiskah:
Right. We don't want the sign-up form falling down. The "What is mastodon" text is guaranteed to be a certain length, because that's universal, but we don't know how long another instance would make their text. If we want to guarantee "equality of quality" it may be wise to keep things as in, though I agree that having the about this instance thing but the very last paragraph feels odd.

dunn:
I can't think of a solution that balances these considerations, so I'll close this for now.  Thanks for the review!

yiskah:
It was a good idea ^^;; really. We just gotta figure it out more

Issue #1515, opened 2017-04-11
Password reset allows enumeration of used email addresses


Dr-Syn:
When in the password reset dialogue, entering an email address that is not in the database of users' passwords results in a specific error stating that the address has not been found.

This allows for enumeration of what emails are in-use by the users on a given instance, which could conceivably be leveraged to, e.g., determine which emails in a db dump located elsewhere are in-use by the users of a given instance, to take advantage of credential reuse or the like.

The best-practice behavior from a security POV would be to respond with something along the lines of

> A password reset email will be sent to this address if it is found in the database

regardless of whether the email address is present or not; this would nullify this kind of attack.

( Yes - I am aware that credential reuse is frowned upon; I'm also aware that it happens constantly and it's better to mitigate it when possible than to add to the endless lectures that regular users ignore. )

* * * *

- [ x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Dr-Syn:
Thank you very much!

gandaro:
You can also insert the email address in question into the registration form, and you will be told whether it is already taken.

From what I take from *devise.en.yml,* you can also use the “unlock account” form to find out which email addresses are taken on a server. See the “send me a confirmation email” form as well.

gandaro:
And an attacker can just set the preferred language (HTTP `Accept-Language`) to German, for instance, and then you can still use the “forgot password” form.

gandaro:
Please re-open.

PR #1516, opened 2017-04-11
Update language files

JantsoP:
Updated translation files according issue #1515


R0ckweb:
I don't think this fix completely #1515 as long as we return an error if email not found.

We probably need to use [paranoid mode from devise](https://github.com/plataformatec/devise/wiki/How-To:-Using-paranoid-mode,-avoid-user-enumeration-on-registerable)

expenses:
Good idea, it would be cleaner to use a single string though.

Gargron (approved):

PR #1520, opened 2017-04-11
Open sidekiq and pghero admin pages in new tab

bradurani:
On `settings` -> `administration` the sidekiq and pghero admin pages now open in a new tab


bradurani:
Whoops! guess I'm invalidating those secrets *facepalm*

PR #1526, opened 2017-04-11
Vagrant should install Ruby version set in the rbenv version file

happycoloredbanana:
Instead of hardcoded 2.3.1, retrieve the required ruby version from the rbenv version file in the mastodon repo.


mjankowski:
Similar to https://github.com/tootsuite/mastodon/pull/1502

PR #1527, opened 2017-04-11
Avoid user enumeration with devise paranoid mode

### LINKED ISSUE #1515: Password reset allows enumeration of used email addresses

R0ckweb:
close #1515

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -74,7 +74,8 @@ Devise.setup do |config|
   # It will change confirmation, password recovery and other workflows
   # to behave the same regardless if the e-mail provided was right or wrong.
   # Does not affect registerable.
-  # config.paranoid = true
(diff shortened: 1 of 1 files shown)


expenses:
Sounds like a good policy to have this enabled by default.

Gargron (approved):

PR #1532, opened 2017-04-11
fix #1475: create external volume for postgresql database

### LINKED ISSUE #1475: Documented process for updating Mastodon with Docker does not work
Based on README.md it seems we should be able to rebuild updates live. I have not been able to do this - 'docker-compose build' only works if I have first executed 'docker-compose down'. I decided to try this just to get the build working. Due to #1376, 'docker-compose down' killed my DB and I had to start over. Luckily I'm the only user on my instance so far.

I am guessing I missed a crucial step that would have enabled 'docker-compose build' to succeed without taking Mastodon down, since other instances are getting updates without wiping all user data. Documentation needs more detail on the correct procedure.

* * * *

ericblade:
This is to prevent data loss when doing docker-compose down

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,11 @@ Consult the example configuration file, `.env.production.sample` for the full li
 
 The project now includes a `Dockerfile` and a `docker-compose.yml` file (which requires at least docker-compose version `1.10.0`). You need to turn `.env.production.sample` into `.env.production` with all the variables set before you can:
 
+```
+    docker volume create --name=data
+
     docker-compose build
+```
 
 And finally
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ services:
   db:
     restart: always
     image: postgres:alpine
+    volumes:
+      - data:/var/lib/postgresql/data
   redis:
     restart: always
     image: redis:alpine
@@ -39,3 +41,7 @@ services:
       - redis
     volumes:
       - ./public/system:/mastodon/public/system
+
+volumes:
+  data:
+    external: true


Fastidious:
@ericblade wouldn’t that break up and running installations?

ericblade:
It does, but as soon as a running installation does docker-compose down, it gets broken anyway. I'm not super familiar with docker, so if there's a better way, I'm all ears. Solution cribbed from https://forums.docker.com/t/data-directory-var-lib-postgresql-data-pgdata-has-wrong-ownership/17963/25

Fastidious:
@ericblade the solution is not to type `docker-compose down`, or to set the path at *install* time. The main README.md states:

> The container has two volumes, for the assets and for user uploads. The default docker-compose.yml maps them to the repository's public/assets and public/system directories, you may wish to put them somewhere else. Likewise, the PostgreSQL and Redis images have data containers that you may wish to map somewhere where you know how to find them and back them up.

ericblade:
"may wish" seems a bit understated there, considering that it *will* lead to data loss when you shut it down, or is docker-compose down not the "proper" way to shut a set of images down to restart them later?  I suspect powering off the host machine over them is probably not cool, also, but I hadn't specifically tested that case.  I did, however, docker-compose down, before shutting down my computer last night during a storm, as i would normally do with any other images, and my database was lost.

Fastidious:
@ericblade you should never do a `docker-compose down`. It `stops and remove containers, networks, images, and volumes`. If you shutdown your server as you normally do, docker will be also shutdown nicely.

ericblade:
Fair enough, was not aware of start/stop vs up/down, and will withdraw this.  

However, I do feel it is worth some discussion -- wouldn't it be normal to store the database (as all data external to the database is stored in public/ already) rather than to expect the docker container to remain the same?  Isn't that at least part of the point of docker, is that you can re-create a running system from the Dockerfile/docker-compose.yml in short order, which would require any data that that depends on being externally mounted?

So, shouldn't it be the normal system configuration to export the database, as well as the other data?

Fastidious:
> wouldn't it be normal to store the database (as all data external to the database is stored in public/ already) rather than to expect the docker container to remain the same?

Agree! On my install, I have volumes for assets and DB in permanent, different locations.

ericblade:
So, it seems to me that it would be a good idea to set this for new installations.  I did just try it out without first creating the volume, and it complains with a warning that it's using the existing /var/lib/postgresql/data instead of mounting it externally. . . so it probably doesn't harm existing installs. 

I'm not sure exactly how to proceed -- I definitely feel that that warning is very understated, as I think it's fairly safe to assume that the default configuration would persist anything that needs to be persisted across instances, and even if your containers were completely destroyed, I would expect them to be able to be brought back up in a few minutes.  

And although I agree that it is best to have as in-depth of an understanding as you can of the software you depend upon, I think most people don't get that understanding until after they've performed database dropping mistakes ;-)

PR #1541, opened 2017-04-11
Allow import/export of mutes list

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/1485

### DIFF
--- a/app/workers/import_worker.rb
+++ b/app/workers/import_worker.rb
@@ -16,6 +16,8 @@ class ImportWorker
       process_blocks
     when 'following'
       process_follows
+    when 'muting'
+      process_mutes
     end
 
     @import.destroy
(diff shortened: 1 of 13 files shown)


app/workers/import_worker.rb
```diff
@@ -35,6 +37,14 @@ def import_rows
     CSV.new(import_contents).reject(&:blank?)
   end
 
+  def process_mutes
+    import_rows.each do |row|
+      target_account = Account.find_local(row.first)
```
Gargron: I think this isn't quite right... Mutes don't apply to local users only. This could should be more or less equivalent with the block import. If the account isn't there, look it up. It's counterintuitive, because if it's a "bad" account, why add it to the system? But if we don't, and it gets added later by someone else, we don't have the block/mute and that's worse.

Gargron: Again, made the mistake of commenting instead of reviewing

Gargron (changes requested):

app/workers/import_worker.rb
mjankowski: Ah, I hadn't thought about it that way.

I had thought that a person bringing an old mutes list from a former instance with them to a new instance would probably not need to worry about accounts that the new instance doesn't know about ... but you're right, we want to proactively mute them ahead of time in case the new instance learns about them.

Will update to match block behavior.

mjankowski:
Updated to call follow remote account service instead of just find local account.

Gargron (approved):

PR #1546, opened 2017-04-11
Quick best practice cleanup of views/helpers

mjankowski:
I gave a once over and removed some trailing whitespace, as well as updating various places to use attribute query methods instead of explicit blank checks.


app/lib/atom_serializer.rb
```diff
@@ -327,7 +331,9 @@ def add_namespaces(parent)
   end
 
   def serialize_status_attributes(entry, status)
-    append_element(entry, 'summary', status.spoiler_text) unless status.spoiler_text.blank?
+    if status.spoiler_text?
+      append_element(entry, 'summary', status.spoiler_text)
+    end
```
expenses: It might be cleaner to do `append_element(entry, 'summary', status.spoiler_text) if status.spoiler_text?`.

mjankowski: I prefer the wrapped style, but not super strongly. Main change was using the query method on the atribute. I'd be fine changing to inline if desired.

Gargron: Indeed, I don't know if you have rubocop notes appear in your editor but our style guide strongly prefers one-line statements to be in guard clauses rather than if-blocks.

Gargron (changes requested):

mjankowski:
Updated PR to restore inline conditionals.

Gargron (approved):

PR #1549, opened 2017-04-11
Make Reporting admin section translatable

tcitworld:
And translate it into english and french

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -7,9 +7,10 @@ class UserMailer < Devise::Mailer
   def confirmation_instructions(user, token, _opts = {})
     @resource = user
     @token    = token
+    @instance = Rails.configuration.x.local_domain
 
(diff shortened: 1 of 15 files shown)


app/mailers/user_mailer.rb
```diff
@@ -9,7 +9,7 @@ def confirmation_instructions(user, token, _opts = {})
     @token    = token
 
     I18n.with_locale(@resource.locale || I18n.default_locale) do
-      mail to: @resource.unconfirmed_email.blank? ? @resource.email : @resource.unconfirmed_email
+      mail to: @resource.unconfirmed_email.blank? ? @resource.email : @resource.unconfirmed_email, subject: I18n.t('devise.mailer.confirmation_instructions.subject')
```
tcitworld: I would like to inject the title of the instance here. Any idea on how I would do that ?

Gargron (approved):

PR #1551, opened 2017-04-11
Update fr.jsx

ghost:
fixing to minor typos


goofy-bz:
note there is no point to use gender-neutral thing for "personnes" which is a feminine word in French

tcitworld (approved):

wxcafe (approved):

PR #1553, opened 2017-04-11
Imports controller errors

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/1540 

The changes in https://github.com/tootsuite/mastodon/commit/e30bbb1cb097ded7db59e905287f85f9f2b5d1b8 broke the import functionality.

This rolls back those changes, and adds a spec which will fail if they are broken again.

### DIFF
--- a/app/workers/import_worker.rb
+++ b/app/workers/import_worker.rb
@@ -4,33 +4,41 @@ require 'csv'
 
 class ImportWorker
   include Sidekiq::Worker
-
   sidekiq_options queue: 'pull', retry: false
 
-  def perform(import_id)
-    import = Import.find(import_id)
+  attr_reader :import
 
-    case import.type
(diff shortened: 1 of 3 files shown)


Gargron:
That PR itself was a fix. This would revert the fix. Can we instead have some sort of solution that fixes both cases, both when files are stored locally, and when they are stored in the cloud?

mjankowski:
I didn't understand that original commit at first, but I see now it was handling the remote file storage (S3 or similar) situation.

I can update this to handle either remote or local file storage and preserve the prior fix.

mjankowski:
This has been updated in a way where I think it preserves the cloud storage fix, and fixes the local storage breakage.

Gargron (approved):

PR #1556, opened 2017-04-11
smtp delivery type fix

foozmeat:
# Conflicts:
#	config/environments/production.rb


tcitworld:
You should add a `SMTP_DELIVERY_METHOD` entry in the `.env.production` template.

foozmeat:
@tcitworld I added that stub entry in .env.production

tcitworld:
Maybe add also a comment to tell that the value can be `smtp` or `sendmail` ? And put this line before the other smtp-related ones ?

foozmeat:
@tcitworld alright I made those changes

gllmhyt (approved):
This would greatly improve the use of a local sendmail when we have a local send-only mail service.

Gargron (approved):

Issue #1469, opened 2017-04-10
Consider: SVG for icons?


jpdevries:
Really enjoying Mastodon's excellent user interface. I've noticed the interface uses icons in a critical manor, so I think browser support here is important.

## Support Analytics
SVG actually has a better caniuse score than webfonts. 

<details>
  <summary>SVG 97.96% safe (screenshot)</summary>
  <img src="http://j4p.us/0e082y2Q3V1p/Screen%20Shot%202017-04-10%20at%206.58.04%20PM.png" />
</details>
<br>

<details>
  <summary>Webfonts 95.3% safe (screenshot)</summary>
  <img src="http://j4p.us/2h1A3H0d1O2Z/Screen%20Shot%202017-04-10%20at%207.01.16%20PM.png" />
</details>
<br>

To be fair, while SVG itself is over 97% safe to use the SVG `<use>` syntax is not as safe:
> IE9-Edge12, Safari 5.1-6, and UCWeb 11 do not support [referencing external files](https://css-tricks.com/svg-use-external-source/) via <use xlink:href>. Polyfills are available: [server-side inlining + snippet](http://codepen.io/hexalys/pen/epErZj/) - [script](https://github.com/jonathantneal/svg4everybody)

But polyfills take SVG `<use>` back up to 97.96% safe. 

_Note: Font Awesome 5 will ship with an SVG framework_

## The Un&ndash;Detectables 
Users may leverage the user agent with user sheets or plugins that manipulate a web app's style to be more legible to them. Here is what the interface looks like when the FontAwesome font-face is overridden:
<details>
  <summary>UI w/ icon font overriden (screenshot)</summary>
  <img src="http://j4p.us/3k0y2m2B0x2J/Screen%20Shot%202017-04-10%20at%208.31.02%20PM.png" alt="Mastodon Interface with empty squares where Icons should be" />
</details>
<br>

We have no way of detecting if a user has done this, so truth be told, web fonts are undetectably less than 95% safe to use.

## Technical Difference

We'd be changing this:
```html
<i class="fa fa-fw fa-globe"></i>
```

to something like this:
```html
<svg class="icon icon-globe">
  <title>Federated Timeline</title>
  <use xlink:href="assets/img/icons.svg#icon-globe"></use>
</svg>
```

## Pros
 - better browser support
   - possible to [achieve 100% browser support with an `<image>` fallback](https://css-tricks.com/a-complete-guide-to-svg-fallbacks/)
 - better Art Direction
[...]

Issue #1349, opened 2017-04-09
Toolbar icons need accessible names


LJWatson:
The toolbar icons ("Getting started", "Local timeline", "Federated timeline", "Preferences", "Logout") do not have proper accessible names. For example:
```
<a title="Getting started" class="drawer__tab" href="/web/getting-started">
<i class="fa fa-fw fa-asterisk"></i>
</a>
```

This means that in some circumstances the purpose of the link is not being announced by screen readers. The problem is that the ```<i>``` element is the only content of the link, and it has no actual content for screen readers to use.

Note: When using the tab key to navigate the links, the `title` attribute does provide the necessary information. However, the `title` attribute should not be used in this context.

The solution is to remove the title attribute (to avoid the previous problem, and to avoid duplication with the correct solution), and use ARIA to enable screen readers to recognise the ```<i>``` element. For example:
```
<a class="drawer__tab" href="/web/getting-started">
<i role="img" aria-label="Getting started" class="fa fa-fw fa-asterisk"></i>
</a>
```

wxcafe:
Can you submit a PR for this?

LJWatson:
@wxcafe
I would, but I don't use React.js myself, sorry.

AntonioVdlC:
I can work on this if nobody else is! 🙂

wxcafe:
Sure yeah! 😊

Clément 'wxcafé' Hertling

fvsch:
I would recommend keeping the title attribute as well, since it was probably intended to make the interface more usable (at least for some mouse/trackpad users).

```html
<a title="Getting started" class="drawer__tab" href="/web/getting-started">
  <i role="img" aria-label="Getting started" class="fa fa-fw fa-asterisk"></i>
</a>
```

IanPouncey:
The problem with using `title` to provide necessary information is that it only works for pointing device users. People using a keyboard, touch screen, voice input, etc. will still not be able to discover what the link is for.

When icons are used but their meaning is not clear they should be accompanied by text to provide a usable experience. Beyond the need for an accessible name, this is a design problem as much as a technical implementation problem.

Note that the small screen size icon links do not have a title attribute but also lack an accessible name.

jpdevries:
I noticed this as well. Made a quick VoiceOver GIF showing the issue

![](http://j4p.us/1U2l1Y230H3J/mastodon_vo.gif)

[...]

PR #1562, opened 2017-04-11
DRYing out icon code

jpdevries:
This PR creates a component for icons so that changes to icons in the future can be made in just one&nbsp;place 💯 

Just removes duplication for now but does so as a step
towards #1469

also see #1349 which proposes adding descriptive text to icons (for the toolbar but eventually would be great across the board)

### Usage
The Icon component just takes the name of the icon now:
```js
  <Icon icon={this.props.icon} />
```

You can pass an `active` property which will add an `.active` class to the icon.
```js
  <Icon icon={this.props.icon} active={true} />
```

You can also pass a `fixedWidth` property which will add a `fa-fw` class to the icon:
```js
  <Icon icon={this.props.icon} fixedWidth={true} />
```

or override the `iconPrefix`:
```js
  <Icon icon={this.props.icon} iconPrefix="foo" />
```

The Icon component sets the `className` for you. If for some reason you want to override this, pass a `className` property.
```js
  <Icon icon={this.props.icon} className={`foo foo-${this.props.icon} foo-bar`} />
```

### DIFF
--- a/app/assets/javascripts/components/components/status_content.jsx
+++ b/app/assets/javascripts/components/components/status_content.jsx
@@ -5,6 +5,7 @@ import emojify from '../emoji';
 import { isRtl } from '../rtl';
 import { FormattedMessage } from 'react-intl';
 import Permalink from './permalink';
+import Icon from './icon';
 
 const StatusContent = React.createClass({
 
@@ -39,7 +40,7 @@ const StatusContent = React.createClass({
       } else if (link.textContent[0] === '#' || (link.previousSibling && link.previousSibling.textContent && link.previousSibling.textContent[link.previousSibling.textContent.length - 1] === '#')) {
         link.addEventListener('click', this.onHashtagClick.bind(this, link.text), false);
       } else if (media) {
-        link.innerHTML = '<i class="fa fa-fw fa-photo"></i>';
+        link.innerHTML = '<Icon icon="photo" />';
       } else {
         link.setAttribute('target', '_blank');
         link.setAttribute('rel', 'noopener');
--- a/app/assets/javascripts/components/components/status_content.jsx
+++ b/app/assets/javascripts/components/components/status_content.jsx
@@ -40,7 +40,7 @@ const StatusContent = React.createClass({
       } else if (link.textContent[0] === '#' || (link.previousSibling && link.previousSibling.textContent && link.previousSibling.textContent[link.previousSibling.textContent.length - 1] === '#')) {
         link.addEventListener('click', this.onHashtagClick.bind(this, link.text), false);
       } else if (media) {
-        link.innerHTML = '<Icon icon="photo" />';
+        link.innerHTML = '<i class="fa fa-fw fa-photo"></i>';
       } else {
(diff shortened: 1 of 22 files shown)


app/assets/javascripts/components/components/status_content.jsx
```diff
@@ -39,7 +40,7 @@ const StatusContent = React.createClass({
       } else if (link.textContent[0] === '#' || (link.previousSibling && link.previousSibling.textContent && link.previousSibling.textContent[link.previousSibling.textContent.length - 1] === '#')) {
         link.addEventListener('click', this.onHashtagClick.bind(this, link.text), false);
       } else if (media) {
-        link.innerHTML = '<i class="fa fa-fw fa-photo"></i>';
+        link.innerHTML = '<Icon icon="photo" />';
```
Gargron: This won't actually work. `Icon` is not html, it's a react component. It works with `i` because that *is* html.

Gargron (commented):
I realize this quite a lot of work, but I don't see *a lot* of benefit to doing this, when aria-labels should be added on buttons and links anyway, and not on the icons themselves (right?)

I fear this could introduce a lot of small bugs like the one I commented on.

app/assets/javascripts/components/components/status_content.jsx
jpdevries: @Gargron It works. `<Icon>` references the icon component just like any of our other components
https://github.com/tootsuite/mastodon/pull/1562/files#diff-38b8991419d1039766d703c68c4a4ebdR3

jpdevries:
@Gargron 
> when aria-labels should be added on buttons and links anyway, and not on the icons themselves (right?)

I assumed so also but see https://github.com/tootsuite/mastodon/issues/1349#issue-220477742

> but I don't see a lot of benefit to doing this
See https://github.com/tootsuite/mastodon/issues/1469

app/assets/javascripts/components/components/status_content.jsx
jpdevries: sorry @Gargron I stand corrected. I thought that was being parsed by React and assumed you didn't see the new React component. I'll work on fixing that.

jpdevries: went back to the static HTML string in https://github.com/tootsuite/mastodon/pull/1562/commits/b18d4752cd033322811c839462953a7304c67a4d

so icon code isn't 100% DRY 😞 due to the one spot `.innerHTML` is used

ashfurrow:
@jpdevries thanks for the pull request. It looks like it wasn't something that @Gargron wanted to include in the project, and there are a lot of merge conflicts since opening the pull request. I'm going to recommend we close the pull request, but it's totally open for discussion.

jpdevries:
@ashfurrow no problem. if you ever want me to rebase and resubmit it let me know. i think it could be a good step towards a larger theming and accessibility effort and eventually switching to SVG icons https://github.com/tootsuite/mastodon/issues/1469

ashfurrow:
Yup yup, thanks again!

PR #1567, opened 2017-04-11
Refactor exports controller

mjankowski:
A mix of updates:

- Adds spec coverage for the existing export controller actions
- Fixes a deprecation warning
- Refactor the routes/controllers to more RESTful style for exports area
- Adds `Export` class to handle the accounts-to-csv conversion (instead of in controller)

### DIFF
--- a/app/controllers/settings/exports_controller.rb
+++ b/app/controllers/settings/exports_controller.rb
@@ -17,7 +17,7 @@ class Settings::ExportsController < ApplicationController
     @accounts = current_account.following
 
     respond_to do |format|
-      format.csv { render text: accounts_list_to_csv(@accounts) }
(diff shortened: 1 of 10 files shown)


app/controllers/settings/exports_controller.rb
```diff
@@ -15,32 +12,18 @@ def show
   end
 
   def download_following_list
-    @accounts = current_account.following
+    export_data = Export.new(current_account.following).to_csv
```
Gargron: If you are replacing these with own controllers, why are these methods still in here?

mjankowski: Because I forgot to remove them! oops.

mjankowski: (Fixed in new push)

Gargron (approved):

Issue #1350, opened 2017-04-09
feature request: web ui: show username/address as a tooltip? on hover over 


ericblade:
Due to the limited width of columns, it is often not possible to see the full username and address of a user in the header of their messages, and the UI seems to strip the hostname part of mentions inside messages.  

Seems like it would be a good idea to have hover over the username part of a header show a tooltip with the user's full username and address, maybe even more details if they happen to be available on the client side already.  Same with hovering over a @ mention.

* * * *

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

ericblade:
The title for the mentions seems like it works great, I would like to extend it to other places where usernames are visible, as well.

Cassolotl:
This seems related to *[Feature request] Show profile preview on hover #6354*.

trwnh:
Probably a subset of #6354 -- which is for a richer profile preview, not just the username/handle.

PR #1568, opened 2017-04-11
Generate Colors for At-Mentions based on Instance Hash

mbilokonsky:
Not sure if there's interest in this at the project level...? But I've added this to my instance and folks seem to really like it.

If we did this the 'right' way we'd have ruby wrapping the '@' in a span with a class, so we avoid the weirdo DOM manipulation I'm doing. We could also look into potentially having a canonical list of colors for major instances, and falling back to hash-generated colors for unregistered instances...? 

Accessibility issues abound for color-blind etc too. 

But! This is an easy way to at-a-glance disambiguate the same username on different instances, and just in general showcase the diversity of the community. If it's something that y'all would like to see merged I'm happy to clean it up further etc. Just let me know! :)


mbilokonsky:
screen shot 2017-04-11 at 9 22 32 pm

mbilokonsky:
This is what two-color looked like. More 'secure', but ugly as hell. Undid it.

screen shot 2017-04-11 at 10 06 48 pm

ericblade:
Just my opinion, the inversed colors look really jarring, but i do like the slim underline.  Not sure if making it foreground for the entire thing would be potentially really awful or a better choice.  I like the idea, though . . .

yiskah (changes requested):
I'm uncomfortable merging this unless you can add something which would guarantee this will be implemented such that only colors safe for color-blind users would be used, or that it would otherwise not cause issues for color-blind users.

Fastidious:
I find it too busy, way too busy. Would like to have it as a setting that can be either disabled instance wide, on the user’s preferences, or the administrative interface.

mbilokonsky:
Disregard for now, this is having some (obvious) issues with merging and rebasing. Still working on it.

Spunkie:
> that it would otherwise not cause issues for color-blind users.

Well the username are the same color as before and the @ symbol is only displayed in black/white so at worst it's as readable as we have now without the benefit the colors are bringing. 

> implemented such that only colors safe for color-blind users would be used

Short of putting in a user setting asking each user what type of colorblindness they have I don't really see how this is even technically feasible? Putting in such a user setting would have far reaching implications outside this PR though.
___
I also agree allowing the user to disable/enable this as a user preferences would seem a good idea. Adding an admin option to disable it instance wide just seems silly to me though, at most there should be an option for the admins to change the default for the user preferences.

mbilokonsky:
Right, so the colorblind thing - honestly the point is that we're getting a _deterministic random color_ for each instance. I'm not sure how to restrict that color into something that's colorblind-friendly, but I'd love to hear ideas.

But as Spunkie points out above, this is at worst an augmentation that would add some color that may not be as useful to someone who is colorblind as it is to someone who is not. I'm not sure what it means for this to be colorblind friendly.

I definitely agree that it should be behind some sort of user-controlled setting (not admin controlled, I think, right?)

Paulloz:
Hi! This is your local colourblind (deuteranomal) developer speaking.

> Right, so the colorblind thing - honestly the point is that we're getting a deterministic random color for each instance. I'm not sure how to restrict that color into something that's colorblind-friendly, but I'd love to hear ideas.  

As said before, it doesn't seem to be technically feasible (as long as you want to cover the wide range of diverse colourblindness). Standard workaround would be the usage of different patterns (stripes, dots, etc.) in addition of colours. But here this does not seem feasible either.  
So, the question is: is this adding some interpretation difficulty to colourblind. I'd tend to say «no», if the local instance colour is distinguishable from all other instance colours.  

Well, I hope this will help the reasoning.

mbilokonsky:
Thank you so much for weighing in!

mbilokonsky:
I think I'm going to add a thin right border if the instance is local. That way the color doesn't change, but it DOES have an unambiguous "this is local" indicator as well. That satisfies @Paulloz's color-blind request above, too.

Gargron:
I think that what we want is some sort of tooltip rather than coloring text in this manner.

ericblade:
... i thought we had recently added UI title attributes to these, but I'm not seeing them in any browser when I hover.  Related request #1350 for tooltips.  

I do like the idea of different coloring per host, or even just different color for remote vs local, but unless you clamp the ranges, it could be pretty weird in some places, and there's no guarantee that color blindness would notice the significance of it depending on the randomness.  I also really like the little underlines you had there . . but of course, that could be an idea for someone else's client UI too :-D

yiskah:
My concern is less with being able to distinguish the color so much as ensuring the contrast is still there even if the colors meld. I've heard complaints from some color blind users that a lot of the instances with custom style-sheets are difficult to use because the contrasts are between colors that meld. If it's always a very bright color then I suppose it could be OK. I agree with gargron though that there is a more elegant solution to this.

mbilokonsky:
Point of this is that it's at-a-glance, as opposed to a tooltip which requires interaction. That feels valuable to me - tooltip to confirm instance name is also valuable. Two separate things, though. 

Re: colorblind, I get it, but it's deterministically randomly generated. I'm not sure how to do that into a bounded color space and anyway it's already going to have more shades than are easily distinguishable, color-blind or not. 

It's meant as an augmentation, not as a canonical feature that is the one true way to see instance alignment. 

But, I'm happy to keep it as a local feature and not merge if it's against the vibe you're going for - no worries!

yiskah:
If you can't make it a11y-safe because of how it works, then it needs to work differently.

mbilokonsky:
Can you explain what you mean by a11y-safe in this context? There is title text for screen readers and after reading the feedback of a colorblind user above I'm planning to add non-color based indication of 'local' toots, as I said above. 

What would it mean for it to be more a11y-friendly?

mbilokonsky:
The other approach, as I wrote in the initial PR, is to have a pre-defined set of colors hard-codes for major instances. But this seems like a tricky proposition because it would require instances to always have an updated list, and even then new or small instances wouldn't render correctly. 

Honestly just curious how best to improve this - I want to make it as useful as possible to as many people as possible. Thanks!

Artoria2e5:
Putting my grain of salt here. The main problem with a11y in this PR may be the unpredictability of the colors and, as a result, the unpredictability of the at/background contrast value. It can be a good idea to do some HSL (or HSLuv/LCh) magic here so you can limit the Lightness component to a range that guarantees some contrast. The Saturation component can be also limited on aesthetic grounds. The Hue component should ideally be the main value that allows one to differ one instance from another.

Issue #146, opened 2016-11-14
Admin UI for accounts


Gargron:
- [x] List local/remote accounts/users
- [x] Disable registrations
- [ ] Manually create user accounts
- [x] Reset an account's password
- [x] Toggle account silencing, NSFW marker

triplefox:
Forcing NSFW will be a help with porn bots

haniawni:
It'd be particularly useful for certain types of organizations to have member email addresses be unconstrained, but able to limit registrations, for example:
- require entering an invite code (ex. me and my friends, or members of a multiuniversity research organizing network)
- to require approval by an admin or by api, for semiclosed communities with an application process
- or up to a fixed total number of accounts

These suggestions would be particularly valuable for communities that are likely to be targeted by harassment, but which need registration open continuously for certain sorts of people.

wxcafe:
The suggestions in the last comment should be added to the main post

SherSpock:
> * [ ]  Manually create user accounts

Just curious, is the above still both missing and desired?

Checking out Mastodon for the first time and looking for something simple

PR #1569, opened 2017-04-11
Feature Create user account from admin interface

R0ckweb:
This allow to create new user account from administration web interface.

Related to #1456 #146 

screen shot 2017-04-11 at 20 06 08

screen shot 2017-04-11 at 20 06 19

screen shot 2017-04-11 at 20 06 47


halkeye:
Drat you are slightly ahead of me.

someone was reporting earlier that the forgot password functionality was breaking, but didn't look into it yet. As such, I was thiking it might be worth it to be able to specify password. Maybe default to random if not otherwise set.

Are you creating tests for this? Thats what I was getting around to right now.

R0ckweb:
I've been hesitant to add password fields too.
But I think use forgot password function force users to change their password. If  they have a ready to use password I think most of them don't take the time to change it.
Also, according to #146 next step is to implement account password modification from admin ui.

Password fields can easily be added if I'm the only one with this opinion.

Unit test added ;)

app/controllers/admin/accounts_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 module Admin
   class AccountsController < BaseController
-    before_action :set_account, except: :index
+    before_action :set_account, except: %i(index new create)
```
halkeye: New and create don't have account IDs in the URL so this function isn't needed to be called

spec/controllers/admin/accounts_controller_spec.rb
```diff
@@ -13,11 +15,104 @@
...
     it 'returns http success' do
-      get :show, params: { id: account.id }
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe 'POST #create' do
+    before do
+      request.env["devise.mapping"] = Devise.mappings[:user]
```
halkeye: If you are not using password in the form, why is the test providing it?

spec/controllers/admin/accounts_controller_spec.rb
```diff
@@ -13,11 +15,104 @@
...
+      expect(User.find_by(email: 'testadmin@example.com')).to_not be_nil
+    end
+  end
+
+  describe 'POST #create existing user' do
+    let(:user) { Fabricate(:user, email: 'testadmin@example.com', account: Fabricate(:account, username: 'testadmin')) }
+
+    before do
+      request.env["devise.mapping"] = Devise.mappings[:user]
+      post :create, params: { user: { account_attributes: { username: user.account.username }, email: user.email, password: '12345678', password_confirmation: '12345678' } }
```
halkeye: Same here, why password being provided

halkeye (changes requested):
Minor issues

app/controllers/admin/accounts_controller.rb
R0ckweb: This is why they are in 'except' filter !

halkeye: Oh shit. Shouldn't code review from bed without glasses. Sorry sorry

halkeye:
Makes sense if there is eventually going to be a way for admins to set passwords. I withdraw objections.

halkeye:
Maybe it should go to the show page after creation. I know it's kinda rails standard to go-to list though

halkeye (approved):

halkeye:
@R0ckweb ps, you seem to have merge conflicts now. The code moves so fast.

Issue #1565, opened 2017-04-11
Auto-remove unconfirmed users


Angristan:
I have a lot of users that just register with some random username and email, don't confirm the account and never come back.

![screenshot_11-04-2017_b hwqfvrkf](https://cloud.githubusercontent.com/assets/11699655/24925983/5dada11a-1efa-11e7-8826-7d17aada1423.png)

![screenshot_11-04-2017_b wimylqg8](https://cloud.githubusercontent.com/assets/11699655/24925992/636ae70c-1efa-11e7-88e2-e641e045d567.png)

It's faking the user stats, taking useless space in the DB and taking usernames that may have been used by others.

It would be great to have an option to delete the users manually, or having a cron that delete them after a certain delay, 7 days for example. I believe it does not cause any problem for the federation.

expenses:
You should be able to use cron to do the rake task now.

Fritillaria:
Hello, I was able to register for an account, and have never been able to get a confirmation email sent to me despite trying about 50 times. If there is going to be autodelete, there should be some attention to why the process is not being finished in the first place. How many of these users gave up trying to get confirmation? Why allow a user to open an account if I can't confirm it? I'm a user, that's all, but I'm frustrated.

Aldarone:
I'm on 8a58942 and when I tried the rake task I'm getting this exception : 

```
rails aborted!
NoMethodError: undefined method `delete_all' for #<Array:0x00564340b47998>
/mastodon/lib/tasks/mastodon.rake:85:in `block (4 levels) in <top (required)>'
/mastodon/vendor/bundle/ruby/2.4.0/gems/activerecord-5.0.2/lib/active_record/relation/batches.rb:121:in `block in find_in_batches'
/mastodon/vendor/bundle/ruby/2.4.0/gems/activerecord-5.0.2/lib/active_record/relation/batches.rb:214:in `block in in_batches'
/mastodon/vendor/bundle/ruby/2.4.0/gems/activerecord-5.0.2/lib/active_record/relation/batches.rb:198:in `loop'
/mastodon/vendor/bundle/ruby/2.4.0/gems/activerecord-5.0.2/lib/active_record/relation/batches.rb:198:in `in_batches'
/mastodon/vendor/bundle/ruby/2.4.0/gems/activerecord-5.0.2/lib/active_record/relation/batches.rb:120:in `find_in_batches'
[...]

PR #1571, opened 2017-04-11
Add task in order to delete unconfirmed users (older than 2days)

d0p1s4m4:
Related to #1565 

just add a new rake task in order to delete unconfirmed users

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -74,11 +74,11 @@ namespace :mastodon do
       end
     end
   end
-  
+
   namespace :users do
     desc 'clear unconfirmed users'
     task clear: :environment do
-      User.where('confirmed_at is NULL AND confirmation_sent_at <= ?', 7.days.ago).find_each(&:destroy)
+      User.where('confirmed_at is NULL AND confirmation_sent_at <= ?', 2.days.ago).find_each(&:destroy)
     end
   end
 
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -74,6 +74,13 @@ namespace :mastodon do
       end
     end
   end
+  
+  namespace :users do
+    desc 'clear unconfirmed users'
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -74,6 +74,13 @@ namespace :mastodon do
       end
     end
   end
+  
+  namespace :users do
+    desc 'clear unconfirmed users'
+    task clear: :environment do
+      User.where('confirmed_at is NULL AND confirmation_sent_at <= ?', 7.days.ago).find_each(&:destroy)
```
Gargron: Hmm maybe 24 hours or 2 days? Highly unlikely that someone would press "confirm" in an e-mail after waiting for so long.

Angristan: Yes, 2 days seems good.

Golgotha29: So, add a line on the first mail or on the web page : "You have two days for validate your account..."

Gargron (approved):

Fastidious:
Can you update the documentation on this one? Some administration document or similar, perhaps?

symac:
Hello,
will there be any way for a remote service to know that a user account has been deleted ? On http://mastory.social/ I am regularly looking at user pages on their instances to update the information I have and if I get a 404, at the moment, I consider it might be a temporary issue with the server and try again later. If the user has been deleted I would be interested to know it and be able to remove it from my database.

From what I can see in the diff it is just deleting the user from DB, is there any plan to change/improve that ?

Wonderfall:
@Fastidious `docker-compose run --rm mastodon-web rails mastodon:users:clear`

Gargron:
@symac You shouldn't be able to discover unconfirmed users, since they can't login (therefore, can't post)

Suspended accounts return 410.

valentin2105:
Can you release v1.1.2 with this merge ? 

Thanks a lot.

PR #1575, opened 2017-04-11
Add a max_users setting

aeonofdiscord:
Adds a max_users setting to settings.yml so admins can shut down new registrations automatically when the user count reaches a specified value.


expenses:
I think it would be better if you had the `max_users` setting (in `settings.yml`) be infinity by default instead of checking if it is set or not.

Fastidious:
I like this a lot. I was thinking about it today, precisely!

app/controllers/auth/registrations_controller.rb
```diff
@@ -28,7 +28,8 @@ def after_inactive_sign_up_path_for(_resource)
   end
 
   def check_enabled_registrations
-    redirect_to root_path if Rails.configuration.x.single_user_mode || !Setting.open_registrations
+    user_count = Rails.cache.fetch('user_count') { User.count }
+    redirect_to root_path if Rails.configuration.x.single_user_mode || !Setting.open_registrations || user_count >= (Setting.max_users || Float::INFINITY)
```
halkeye: would this not be cleaner to either use `InstancePresenter.new.open_registrations` or move into a model (settings?) so its not duplicated?

aeonofdiscord: Deduplication seems like a good idea. I'll just ref InstancePresenter unless anyone has any objections.

aeonofdiscord:
I'm not sure why the travis build is failing here but it seems to be happening because of something upstream?

halkeye:
Looks like the tests have been failing on the master branch too (looking at commit log)

cc @Gargron

app/controllers/auth/registrations_controller.rb
```diff
@@ -28,7 +28,7 @@ def after_inactive_sign_up_path_for(_resource)
   end
 
   def check_enabled_registrations
-    redirect_to root_path if Rails.configuration.x.single_user_mode || !Setting.open_registrations
+    redirect_to root_path if Rails.configuration.x.single_user_mode || !InstancePresenter.new.open_registrations
```
mjankowski: Maybe wrap this up into a private controller method like `single_user_or_registrations_closed?` to reveal the intention of the check here?

app/presenters/instance_presenter.rb
```diff
@@ -25,4 +24,10 @@ def status_count
   def domain_count
     Rails.cache.fetch('distinct_domain_count') { Account.distinct.count(:domain) }
   end
+
+  def open_registrations
```
mjankowski: Add some spec coverage for this method to show that various combination of settings and user counts are handled correctly?

mjankowski (changes requested):

jeanphilippeD:
Would it make sense to choose a default that is not infinity?

It might be good to choose a conservative default value that performs well in a system a reasonable admin is likely to initially setup, before seriously considering scaling for many thousands users?
From a certain number of users, the administrator needs to consider carefully the performance of the system, and so could update this setting.

aeonofdiscord:
@jeanphilippeD: I'm not sure what to use for a default value. The motivation behind picking infinity was to maintain continuity with the current behaviour, while giving an option to override it. The right setting is going to vary significantly based on the needs of each admin/instance, and I don't want to establish a default that causes problems further down the line.

app/presenters/instance_presenter.rb
aeonofdiscord: I've added some specs but I'm not sure how to mark the fact that I've done that.

app/presenters/instance_presenter.rb
```diff
@@ -26,6 +26,12 @@ def domain_count
     Rails.cache.fetch('distinct_domain_count') { Account.distinct.count(:domain) }
   end
 
+  def open_registrations
```
mjankowski: I think I would prefer to leave `open_registrations` as a delegation to Setting here in the InstancePresenter.

I'm not sure it needs to know about `max_users`, unless we're going to display that in a view somewhere.

We can probably skip instance presenter altogether and create a method in the registrations controller like `new_registrations_accepted?` that checks both open_registrations and max_users, and has similar spec coverage to what you added below. If it would be easier to test that via a helper or separate class, that's fine too.

spec/controllers/auth/registrations_controller_spec.rb
```diff
@@ -20,6 +20,7 @@
 
     before do
       Setting.open_registrations = true
+      Setting.max_users = nil
```
mjankowski: This should probably be the default value, and not need to be set?

spec/presenters/instance_presenter_spec.rb
```diff
@@ -32,6 +26,40 @@
 
     expect(instance_presenter.site_contact_email).to eq "admin@example.com"
   end
+  
+  context "when Setting.open_registrations is false" do
+    it "closes registrations" do
+      Setting.open_registrations = false
```
mjankowski: If you change `Setting...` values in specs, you should make sure to restore them to their original values to not have side effects elsewhere.

app/presenters/instance_presenter.rb
aeonofdiscord: That was what I had originally, but changed in response to this review: https://github.com/tootsuite/mastodon/pull/1575#pullrequestreview-32281280

If I move the test to RegistrationsController then it has to be duplicated in AboutController; it's in InstancePresenter to avoid that duplication.

spec/presenters/instance_presenter_spec.rb
aeonofdiscord: As far as I can tell none of the other specs adhere to this; is this something that should be being done across the board?

Gargron:
Closing due to inactivity.

PR #1582, opened 2017-04-11
Accessible Search Form Submit

jpdevries:
Does
 - Wraps the input.search__input in a form
 - Removes now unnecessary handleKeyDown (automatically triggered now by capturing the form onSubmit)
 - Turns div.search__icon into a button[type=“submit”]
   - add a little styling for the button
 - Adds role=‘search’ to new form (ARIA)
 - allow us to leverage the HTML5 Form Validation API down the line

Does Not
 - add a label or aria-label to the input or button

Assistive Technology may tap into standard `<form>` elements in various ways. Rather than explicitly listening to a keyUp for ENTER in the input, now listening to a generic form submit event.

### DIFF
--- a/app/assets/javascripts/components/features/compose/components/search.jsx
+++ b/app/assets/javascripts/components/features/compose/components/search.jsx
@@ -47,6 +47,7 @@ const Search = React.createClass({
         <input
           className='search__input'
           type='text'
+          required={true}
           placeholder={intl.formatMessage(messages.placeholder)}
           value={value}
           onChange={this.handleChange}
--- a/app/assets/javascripts/components/features/compose/components/search.jsx
+++ b/app/assets/javascripts/components/features/compose/components/search.jsx
@@ -29,15 +29,13 @@ const Search = React.createClass({
     this.props.onClear();
   },
 
-  handleKeyDown (e) {
-    if (e.key === 'Enter') {
-      e.preventDefault();
-      this.props.onSubmit();
-    }
-  },
-
   handleFocus () {
     this.props.onShow();
+  },
+
+  handleSubmit (e) {
+    e.preventDefault();
+    this.props.onSubmit();
   },
 
   render () {
@@ -45,22 +43,21 @@ const Search = React.createClass({
     const hasValue = value.length > 0 || submitted;
 
     return (
-      <div className='search'>
+      <form role='search' className='search' onSubmit={this.handleSubmit}>
         <input
           className='search__input'
           type='text'
           placeholder={intl.formatMessage(messages.placeholder)}
           value={value}
           onChange={this.handleChange}
-          onKeyUp={this.handleKeyDown}
           onFocus={this.handleFocus}
         />
 
-        <div className='search__icon'>
+        <button type='submit' className='search__icon'>
           <i className={`fa fa-search ${hasValue ? '' : 'active'}`} />
           <i className={`fa fa-times-circle ${hasValue ? 'active' : ''}`} onClick={this.handleClear} />
-        </div>
-      </div>
+        </button>
+      </form>
     );
   }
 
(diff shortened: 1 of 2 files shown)


Gargron:
I agree with adding some sort of accessibility like `aria-label`s or `role`s, but I do not agree with putting `<form>` tags into React.js and changing which events trigger what action. There are all sorts of things that can be done in the future with a dynamic UI and you *cannot* rely on the form/submit paradigm of static HTML pages.

jpdevries:
> cannot rely on the form/submit paradigm of static HTML pages

Unless I'm misunderstanding this, I don't think we are relying on the paradigm of static HTML pages. The Form API, is not inherently static. Aspects of it are designed to be used within async SPAs. For example, the Form Validation API is async which is why with this PR the browser will prevent you from submitting an empty (or otherwise invalid) search entry. The Form API also offers methods like `HTMLFormElement.reset()` to reset a form on a dynamic page.

@Gargron What is your concern here? Can you be more specific?

Gargron:
In React, everything can be anything and do anything depending on event bindings, so it really makes no sense to wrap anything into a `<form>` tag specifically or rely on the form tag's submit event. That is not idiomatic React code. Aria labels and roles are good and perfectly fine, but not forms.

jpdevries:
> so it really makes no sense to wrap anything into a <form> tag specifically or rely on the form tag's submit event

Might be worth labelling this accessibility and seeing what comes of the discussion. 

I'm not sure if ARIA roles add support for HTML5 Form Validation and Form Data APIs, but those are just nice to haves. The main idea is that users should be able to with dictation for example say "submit this form", the assistive technology fires the submit event of the form, and all is well rather than fingers crossed hoping it figures out how to submit a not-a-form.

marcysutton:
ARIA doesn't impact HTML5 behavior at all. It only impacts assistive technologies, so a keyboard user wouldn't be able to benefit from hitting the enter key elsewhere in the form if there is no form tag. One could argue that the absence of a form tag and the keyboard support that comes with it is a failure of [WCAG 2.1.1 for keyboard operability](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html).

MichielBijl:
The `button` doesn’t need a `type` attribute set to `submit` as that’s already the default value for it.

> The missing value default is the Submit Button state.

[The button element in HTML 5](https://www.w3.org/TR/html5/forms.html#the-button-element).

jpdevries:
@michaeljdeeb sure, and `<main>` doesn't need `<main role="main">` either, and labels that contain inputs don't need `<label for>`, but for backwards compatibility it is best to be verbose. 

What the HTML5 spec says user agents should do, and what they actually do, can be two different things. That said I don't know of any user agents that won't submit a form from a plain `<button>`. Likewise I don't know that they all do.

MichielBijl:
@jpdevries for what it’s worth, this has been in the spec as early as [HTML 4.01](https://www.w3.org/TR/html401/interact/forms.html#h-17.5) (1999).

jpdevries:
@michaeljdeeb There are things in the spec that still not all user agents respect, which is why I tend to favor being verbose. I'm fine with removing the `type="submit"` if it will move this PR along, because I think Mastodon is awesome and it is important for it to be accessible to everyone.

MichielBijl:
@jpdevries I wouldn’t hold the verboseness hold this PR back. It’s not that important. Also, I’m not @michaeljdeeb :P.

Gargron:
We've added a lot more a11y things into Mastodon now, but the reason I am against this PR is because it simply cannot be consistent throughout the application. You wrap the search input in a `<form>`, but there are a myriad of other inputs and dynamic elements (not at last the actual compose form) which would not be touched by this change. Since the search input listens on "enter" I think it's fine as-is.

Issue #1572, opened 2017-04-11
Vector favicon for Safari pinned tabs


daprice:
Safari on MacOS wants a single-layer, black .svg version of the favicon to show when the user “pins” a tab, along with a color to saturate the icon with on mouse hover (in the mean time, it just shows a square with the first letter of the instance’s domain, which isn’t bad, but a vector version of favicon.ico would probably be better). See [Apple’s docs on this](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html) and [this blog post with more info](http://blog.iconfactory.com/2015/11/the-new-favicon/).

I’d be happy to create this myself and add the `<link>` for it ~~if someone could point me to a vector version of the current favicon or Mastodon logo to work from~~.

Edit: just saw the SVG version linked in #962. I’ll go ahead and make a PR for this with the caveat that the logo itself might need to be updated after #962 is complete.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mjankowski:
SVG logo: https://github.com/tootsuite/mastodon/blob/master/app/assets/images/logo.svg

xuv:
Just wanted to point out there is a pull request in the queue that updates the SVG logo so that it corresponds better to the actual logo being used. 

https://github.com/tootsuite/mastodon/pull/1362

PR #1585, opened 2017-04-11
Add mask-icon.svg (#1572)

daprice:
Closes tootsuite/mastodon#1572

Added a 16x16 solid .svg version of the current logo for use as a tab icon in Safari, per [Apple’s docs](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html).

The rightmost (third) pinned tab in the attached screenshot shows a mastodon instance with this new icon; the middle one shows the default icon given to a mastodon instance prior to this change
screen shot 2017-04-11 at 5 35 33 pm

For reference in case this icon is replaced with a different circular one later on (e.g. as part of tootsuite/mastodon#962): to avoid a weird aliasing issue with the top, bottom, and sides of the circle, I had to make the circle 15.8 x 15.8 so it didn’t touch the edges of the 16x16 viewBox


expenses:
Looks good. I have this concern though we may end up with way too many copies of the same favicon, and that changing each one individually could become a pain. A rake task for copying the main favicon could be useful for this.

daprice:
My concern with automating it is that many of the icons should be slightly different from each other for various reasons. The iOS app one has to be a square with a solid background, the Windows one should be a circle (at least in the current iteration), the favicon.ico and mask-icon.svg need to have a lower level of detail since they’re displayed at smaller sizes (no dots inside the conversation bubbles), etc.

So I don’t think it's as simple as it seems. Depending on how the design changes, those variations might not even be the same – the smaller sizes might end up needing different simplifications, the shapes might change, etc. Any rake task might end up having to be rewritten each time the design undergoes a major change.

expenses:
@daprice Agreed, its not an easy thing to sort out conclusively.

xuv:
Maybe this PR request https://github.com/tootsuite/mastodon/pull/1362 needs to be merged before and then this icon updated accordingly.

wxcafe:
as @xuv said, and since the PR was since merged, please update this.

daprice:
Updated using icon from @xuv’s PR

wxcafe (approved):

xuv:
@daprice Thx. Could you please provide a screenshot with the updated Icon? It's not a show stopper but I want to see how it looks at this small size and in context of Safari. So we can use it as a reference for our discussion here: https://github.com/tootsuite/mastodon/issues/962 Thx.

daprice:
Focused:
screen shot 2017-04-12 at 2 40 40 pm

Unfocused:
screen shot 2017-04-12 at 2 40 59 pm

xuv:
@daprice Awesome. Thx.

PR #1587, opened 2017-04-12
CSS Variables for Sass $colors 💅

jpdevries:
One step closer to supporting “theming”. This would allow user agents and theme designers to customize the colors of the UI with very little CSS and no knowledge of the CSS selectors.
This is particularly useful for [supporting themes that respond to High Contrast Media queries](https://github.com/jpdevries/eureka#-accessible-themes).

CSS Variables aren’t completely safe to use. I’m not familiar with how the Sass is built, but thinking we’d need to:
- [ ] add a postcss plugin for CSS Variables fallbacks

CSS Tricks has a write up on [using CSS Custom Properties for Theming](https://css-tricks.com/css-custom-properties-theming/?utm_campaign=CSS%2BLayout%2BNews&utm_medium=email&utm_source=CSS_Layout_News_92)

<details open>
  <summary>Screenshots</summary >
  <div>
    
    
    
  </div>
</details>

### DIFF
--- a/app/assets/stylesheets/about.scss
+++ b/app/assets/stylesheets/about.scss
@@ -2,7 +2,7 @@
   .wrapper {
     max-width: 600px;
     margin: 0 auto;
-    color: $color3;
+    color: var(--color3, $color3);
     padding-top: 50px;
     padding-bottom: 50px;
 
@@ -15,7 +15,7 @@
     font: 46px/52px 'Roboto', sans-serif;
     font-weight: 600;
     margin-bottom: 20px;
-    color: $color4;
+    color: var(--color4, $color4);
     padding: 20px 0;
 
     img {
@@ -32,7 +32,7 @@
     line-height: 28px;
     font-weight: 400;
     margin-bottom: 20px;
-    color: $color5;
+    color: var(--color5, $color5);
   }
 
   h3 {
@@ -41,7 +41,7 @@
     line-height: 28px;
     font-weight: 400;
     margin-bottom: 20px;
-    color: $color2;
+    color: var(--color2, $color2);
   }
 
   ul, ol {
@@ -67,7 +67,7 @@
     margin-bottom: 26px;
 
     a {
-      color: $color4;
+      color: var(--color4, $color4);
       text-decoration: underline;
     }
   }
@@ -76,14 +76,14 @@
     display: inline-block;
     padding: 7px 7px 5px 7px;
     margin: 0 2px;
-    background: $color3;
-    color: $color1;
+    background: var(--color3, $color3);
+    color: var(--color1, $color1);
     font: 16px/16px 'Montserrat', sans-serif;
     font-weight: 300;
   }
 
   .screenshot {
-    box-shadow: 0 0 15px rgba($color8, 0.4);
+    box-shadow: 0 0 15px rgba(var(--color8, $color8), 0.4);
     margin-bottom: 26px;
 
     img {
@@ -103,7 +103,7 @@
       line-height: 36px;
 
       a {
-        color: $color3;
+        color: var(--color3, $color3);
         text-decoration: underline;
       }
     }
@@ -126,6 +126,7 @@
   justify-content: space-between;
   border-top: 1px solid lighten($color1, 10%);
   border-bottom: 1px solid lighten($color1, 10%);
+  border-color: var(--color1, $color1);
   padding-right: 14px;
 
   .section {
@@ -142,7 +143,7 @@
       font-size: 16px;
 
       &:last-child {
-        color: $color2;
+        color: var(--color2, $color2);
         font-size: 14px;
       }
     }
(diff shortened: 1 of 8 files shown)


expenses:
Sounds good. The support for css variables isn't totally there yet though (http://caniuse.com/#feat=css-variables), so we need to be able to work around that.

jpdevries:
@expenses yup. I believe postcss would turn 

```css
body {
  background: var(--color1, $color1);
}
```

to:
```css
body {
  background: $color1;
  background: var(--color1, $color1);
}
```

For backwards compatibility of the default colors

jpdevries:
@expenses CSS varibales landed in a Edge today. Woot! 
https://twitter.com/malyw/status/851913674021371904

wavebeem:
You can add a fallback but still have dynamic CSS variables in browser that support them? I didn't know that was possible.

wavebeem:
Also I found a good example of updating CSS variables using JS!

https://codepen.io/wesbos/pen/adQjoY?editors=1010

yookoala:
SASS / SCSS already has [variable support](http://sass-lang.com/guide#topic-2) that will naturally support all browser. So this sounds a bit strange to me. Do we really have a use case for this add-on?

marrus-sh:
@yookoala The use-case for this is users who are styling Mastodon with Stylish themes (ie https://userstyles.org), since there isn't currently a way for users to adjust Mastodon's appearance otherwise. Users don't have access to the original SASS so using CSS variables would make this process much simpler.

If Mastodon supported setting its colours through Settings or elsewhere then having CSS variables would be less of an issue.

yookoala:
But if Mastodon were to support theming through CSS variables, there will be no effect for older browsers. Plus theming through variables is, in my experience, usually inflexible. Would it be better off for Mastodon, as a platform, to better support theming through additional theme CSS files?

jpdevries:
@yookoala see https://philipwalton.com/articles/why-im-excited-about-native-css-variables/

> But if Mastodon were to support theming through CSS variables, there will be no effect for older browsers.

Correct. This feature would be considered progressive enhancement, just like the flexbox layout that is currently used.

> Would it be better off for Mastodon, as a platform, to better support theming through additional theme CSS files?

Probably not. CSS Variables are way easier and don't require the theme designer to have any knowledge of the CSS selectors or structure. They just override the variables. CSS Variables are now safe to use in all modern browsers
http://caniuse.com/#search=css%20variables

jpdevries:
I just tried out `postcss-custom-properties` for CSS Variable fallbacks in a project. It worked well, you just need to set the preserve option to true 

Turns
```css
.dangerous {
  color: var(--dangerous, red); }
```

into 
```css
.dangerous {
  color: red;
  color: var(--dangerous, red); } /* old browsers will ignore this */
```

yookoala:
Can we have PostCSS setup to work with our current SCSS?

wavebeem:
@yookoala `var(--blah)` syntax is already passed through Sass untouched, so you should be able to run PostCSS on top of the Sass results. If I recall correctly autoprefixer is actually a PostCSS plugin, and many people use that with Sass and LESS.

jpdevries:
Note: With PostCSS you want to make sure to set preserve to true to no flatten the variables
https://twitter.com/nucliweb/status/856060132102463488

mjankowski:
Closing after merge of https://github.com/tootsuite/mastodon/pull/2912

jpdevries:
@mjankowski #2912 is great but doesn't add CSS Properties for non-Sass related theming support. Would you like me to rebase and submit another PR that does?

mjankowski:
Sure!

yookoala:
I have studied the scss code in detail. I feel it is hard job to change all the variables to CSS variables.

There are around 460 color variable calls in the current SCSS files. And more than 100 of them involves [lighten()] or [darken()] SASS method call. From what I read, these preprocessing script call cannot be translated to CSS Variables until [color-mod()] function is adopted by major browsers. There are also occurrence of SVG in SCSS that uses the theme color variables with method call.

Also, the lighten / darken percentages of these are not entirely consistent. It is not like we can rewrite them into just a few color.

Until we rewrite all those method call into pure SASS-variables, there is no way we can convert the whole theme to fully support CSS variables.

[lighten()]: http://sass-lang.com/documentation/Sass/Script/Functions.html#lighten-instance_method
[darken()]: http://sass-lang.com/documentation/Sass/Script/Functions.html#darken-instance_method
[color-mod()]: https://drafts.csswg.org/css-color/#funcdef-color-mod

jpdevries:
@yookoala Yes you are correct that not all the current colors can be translated. In this PR, I just took a first step into translating all the simple color variables. Basically we would need to come up with names and variables for all the `lighten()` or `darken()` colors. I understand that would need to be part of a larger more organized effort. Would be willing to help out, as I think it would be a dream to be able to theme!

PR #1597, opened 2017-04-12
Allow using an SMTP server without authentication

mig5:
In Action Mailer, setting the 'user_name' or the 'authentication' configuration options to any value, enforces an attempt at authentication.

This means it's not possible to use, say, a Postfix smarthost running on the same host as your Docker instances, that doesn't require authentication (e.g instead relays mail for the Docker network ranges).

This change allows setting the SMTP_LOGIN and SMTP_AUTH_METHOD to 'none' respectively. If SMTP_LOGIN is 'none' then there is no password either.

Works for me on Debian 8 with standard Postfix install that allows relaying from the Docker network ranges/interfaces, with this .env.production config:

```
SMTP_SERVER=172.17.0.1
SMTP_PORT=25
SMTP_LOGIN=none
SMTP_PASSWORD=
SMTP_FROM_ADDRESS=notifications@example.com
SMTP_AUTH_METHOD=none
SMTP_OPENSSL_VERIFY_MODE=none
SMTP_ENABLE_STARTTLS_AUTO=false
```

Borrowed this fix from https://github.com/cantino/huginn/issues/1937 after reading that and https://stackoverflow.com/questions/26093514/rails-actionmailer-send-email-without-authentication


mig5:
I don't know why the test failed but I don't think it's anything to do with my change.

config/environments/production.rb
```diff
@@ -96,10 +96,10 @@
   config.action_mailer.smtp_settings = {
     :port                 => ENV['SMTP_PORT'],
     :address              => ENV['SMTP_SERVER'],
-    :user_name            => ENV['SMTP_LOGIN'],
-    :password             => ENV['SMTP_PASSWORD'],
+    :user_name            => ENV['SMTP_LOGIN'] == 'none' ? nil : ENV['SMTP_LOGIN'].presence,
+    :password             => ENV['SMTP_LOGIN'] == 'none' ? nil : ENV['SMTP_PASSWORD'].presence,
```
wxcafe: pretty sure this should say `ENV['SMTP_PASSWORD'] == 'none' ? nil : ENV['SMTP_PASSWORD'].presence,`

mig5:
No it was deliberate (if username is 'none', don't set the password even if there was one, because we are saying we don't want to auth at all). In other words a failsafe to ensure we set no credentials even if a password was accidentally left in the .env.production file.

But I suppose we could change it, just the user would have to remember to set password to 'none' too.

ineffyble:
I'm not sure I understand the problem entirely. What happens if you just don't sent the SMTP_USERNAME and SMTP_PASSWORD environment variables?

ineffyble:
@mig5 I'd prefer it if we only changed the SMTP_AUTH_METHOD to have a 'none' option. If you don't want to use auth, you set that to 'none', and you don't specify a username or password at all.

wxcafe (approved):

mig5:
@ineffyble changed as per your suggestion. Just be aware that if someone accidentally leaves a username in the settings, even if they set  the auth to 'none', Ruby may still attempt authentication.

This is the problem I was attempting to avoid (force username/password to be nil even if accidentally left in the settings). But if the user obeys the documentation I've added in the .env.production.sample properly, they'll be ok.

See this comment https://github.com/cantino/huginn/issues/1937#issuecomment-286929133 for a more in depth explanation of what I'm talking about.

Issue #1602, opened 2017-04-12
Occasional error in spec test against Api::V1::NotificationsController, excludes mentions


yookoala:
There is an error occasionally happen in Travis CI test:
```
Randomized with seed xxxxx

  1) Api::V1::NotificationsController GET #index with excluded mentions excludes mention
     Failure/Error: expect(assigns(:notifications).map(&:activity_id)).to_not include(@mention.mentions.first.id)
       expected [45, 12, 155] not to include 12
     # ./spec/controllers/api/v1/notifications_controller_spec.rb:63:in `block (4 levels) in <top (required)>'
```

A screen cap of Travis output:
![2017-04-12 00-54-14](https://cloud.githubusercontent.com/assets/91274/24957026/d99126ea-1fbc-11e7-9b39-121070fde62d.png)

Examples,
* [Build #2528](https://travis-ci.org/tootsuite/mastodon/builds/221015356#L1342-L1355) of #1551
* [Build #2616](https://travis-ci.org/tootsuite/mastodon/builds/221125582#L1332-L1344)
* [Build #2632](https://travis-ci.org/tootsuite/mastodon/builds/221187147#L1342-L1355) of #1591
* [Build #2654](https://travis-ci.org/tootsuite/mastodon/builds/221307239#L1343-L1356) of #1570

The problem occurs with no relation to the commit it concerns. It can be "fixed" by simply re-run the tests. This makes the CI results confusing to both contributors or reviewers. CI should be helpful in determining the code functionality and quality.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mjankowski:
Because this is intermittent and hard to reproduce it's hard to know exactly what's happening ... one theory I had is around database IDs from different models matching each other, attempted fix is here: https://github.com/tootsuite/mastodon/pull/1606

PR #1605, opened 2017-04-12
notifications test fail fix.

chaofan:
A pr for issue #1602

A before each clause run too many codes to mess cases asserting clauses.


chaofan:
@Gargron Not bad, mastodon source codes are really beautiful. I should be more familiar with the context of it. Thanks for bring us mastodon.

PR #1607, opened 2017-04-12
Webfinger resource to extract username from resource string

mjankowski:
I noticed the logic in this controller contained a reference to `params_path[:controller] == 'users'` which I think must be for outdated code (that route no longer exists?). I extracted the logic to pull out a username from a resource to it's own class, and added specs for the strings I'm guessing this is supposed to handle.

Feedback on whether that list of strings is exhaustive and sufficient would be appreciated!

### DIFF
--- a/app/lib/webfinger_resource.rb
+++ b/app/lib/webfinger_resource.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+class WebfingerResource
+  attr_reader :resource
+
+  def initialize(resource)
+    @resource = resource
+  end
+
+  def username
+    case resource
+    when /\Ahttps?/i
+      username_from_url
+    when /\@/
+      username_from_acct
+    else
+      raise(ActiveRecord::RecordNotFound)
+    end
+  end
+
+  private
+
+  def username_from_url
+    if account_show_page?
+      path_params[:username]
+    else
+      raise ActiveRecord::RecordNotFound
+    end
+  end
+
+  def account_show_page?
+    path_params[:controller] == 'accounts' && path_params[:action] == 'show'
+  end
+
+  def path_params
+    Rails.application.routes.recognize_path(resource)
+  end
+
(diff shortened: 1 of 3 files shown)


app/lib/webfinger_resource.rb
```diff
@@ -0,0 +1,60 @@
+class WebfingerResource
+  attr_reader :resource
+
+  def initialize(resource)
+    @resource = resource
+  end
+
+  def username
+    case resource
+    when /https/
```
Gargron: I'm thinking this should be `/https?/`

mjankowski: Ah I missed that in the original ... so both of these strings should work?

- `http://example.com/users/name` -> `name`
- `https://example.com/users/name` -> `name`

?

Gargron: Although your tests seem to account for the http case so I'm not sure what this regex does

mjankowski: PR updated to reflect that https is supported.

app/lib/webfinger_resource.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def initialize(resource)
+    @resource = resource
+  end
+
+  def username
+    case resource
+    when /\Ahttps?/i
+      username_from_url
+    when /acct/
```
Gargron: Thinking right now - You should be able to pass username@domain without the acct: in the front, and it should be able to deal with that. So maybe just make that the else condition.

Gargron: I should have submitted that as a change request - this should be able to handle `username@domain` without the `acct:` at the front

Gargron (changes requested):

mjankowski:
Updated to handle the simple `user@host` (w/out `acct:`) as well.

Gargron (approved):

PR #1623, opened 2017-04-12
fix Japanese translation

pinfort:
some mistake of translation was fixed


NeetShin:
This commit also fixed #1620 's incorrect notation. thanks @Grabacr07

Gargron (approved):

Issue #1628, opened 2017-04-12
Regression in newlines on profile page


nightpool:
This is a regression caused by #1397 

The toot displays fine in the web UI:
![image](https://cloud.githubusercontent.com/assets/233815/24973216/0130a90e-1f8c-11e7-8795-68561c3f8e4c.png)

but on the permalink page, it displays with way too many newlines:

![image](https://cloud.githubusercontent.com/assets/233815/24973243/142bcd90-1f8c-11e7-8f63-08118eb904f0.png)

I think the correct move here is to remove the `whitespace: prewrap` in the CSS for statuses.

(Also, I've seen lots of users complaining about this, because they're on instances that don't remove the newline in the web UI—did we know this was a breaking change when we merged it?)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mjankowski:
Probably resolved by https://github.com/tootsuite/mastodon/pull/1630

nightpool:
@mjankowski see also:

Toot:
```
test
test

test
```

in the web ui:
![image](https://cloud.githubusercontent.com/assets/233815/24973671/a134490a-1f8d-11e7-9dd9-768c75d86b7d.png)

on the profile page:
![image](https://cloud.githubusercontent.com/assets/233815/24973679/a62454e6-1f8d-11e7-8c94-5fe775710c91.png)

PR #1630, opened 2017-04-12
Apply white space pre-wrap to `p` instead of surrounding div

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/1609 and https://github.com/tootsuite/mastodon/issues/1628

Before with text like:

```text
Test

This

Testing
```

pre before separate

Before with text like:

```text
test
test

test
```

pre before together

After for those cases:

pre after separate

pre after together

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -272,7 +272,6 @@
   word-wrap: break-word;
   font-weight: 400;
   overflow: hidden;
-  white-space: pre-wrap;
 
   .emojione {
     width: 18px;


nightpool:
This doesn't fix #1628 if the text of the toot is
```
test
test

test
```

in the web ui:
image

on the profile page:
image

mjankowski:
Good catch ... will attempt to adjust for both cases.

mjankowski:
Original PR adjusted to handle both cases.

nightpool:
what happens when we get a message from a federated node that strips HTML?

Gargron:
Closing in favour of #1699

nightpool:
(for those following along at home: that was a typo: the correct PR is #1700)

PR #1633, opened 2017-04-12
Fix admin input display

SansPseudoFix:
Minor fix, but it bothered me (in french, but I think it was the same in deutsch etc…). 

Before:
french_before

Now:
french_after

English:
english

Deutsch:
deutsch


krainboltgreene (approved):

tcitworld:
Does it also work in mobile view ?

SansPseudoFix:
@tcitworld seems so. Just before breaking point:
pref_mobile

tcitworld (approved):

thomas-kuntz (approved):

PR #1648, opened 2017-04-12
Adjustments to the Norwegian translation

ghost:
This is an updated Norwegian translation of Mastodon. The adjustments have been tested on snabeltann.no and feedback has been gathered from the users there, who seem happy about it all.


thor-the-norseman:
The German translation has a problem. It's unrelated to this pull request, except for the fact that I have performed merges with the upstream, and the problematic code was included in those merges.

thor-the-norseman:
I don't know what to do about the codeclimate error, since it's not my code causing it to fail.

Gargron (approved):

PR #1653, opened 2017-04-12
Add joined at (created_at) on user profile

exadeci:
Displays the user joined date on their profile.


expenses:
Personally, I don't really consider this piece of information to be too helpful, so perhaps whether its shown on a users profile or not could be changed in the settings?

app/assets/stylesheets/accounts.scss
```diff
@@ -44,6 +44,15 @@
     }
   }
 
+  .joined-at {
+    float:left;
+    padding-left: 10px;
+
+    &:hover {
```
mjankowski: no styles here?

app/helpers/stream_entries_helper.rb
```diff
@@ -5,6 +5,10 @@ def display_name(account)
     account.display_name.blank? ? account.username : account.display_name
   end
 
+  def joined_at(account)
```
mjankowski: I'd probably just use i18n `t` for this, instead of a helper method?

app/assets/stylesheets/accounts.scss
exadeci: I changed my mind about a tooltip and apparently forgot to remove the hover, I'll change that.

app/helpers/stream_entries_helper.rb
exadeci: Ok I'll do that.

mjankowski (commented):
I'm not sure we actually want to include this on the page (I can see reasons why you don't want how new or how old you are on an instance to be public), but the actual implementation of this is fine.

exadeci:
@mjankowski How about making it a feature disabled by default on old users, enabled for new and with a toggle in the user account settings?

ashfurrow:
@exadeci thanks for the pull request! I like it: it shows how long of a history you have on Mastodon. But I can see it not being a welcome change for some users. I'm going to ask for a review from the Mastodon creator and then we can merge or close – our user settings are already pretty crowded, and this seems like something we'd want to have consistently present or not.

Gargron:
I think the current profile layout is better fit for this feature but it's really outdated now. Might add it myself later if there's actually a demand for it though.

PR #1655, opened 2017-04-12
Fix status whitespace

ghost:
Leave text formatting to client javascript other than linking/sanitization. fixes whitespace errors like leading spaces when toots with linebreaks are made.


rainyday:
Gonna add some more to this before submitting, actually

PR #1657, opened 2017-04-13
Fix status whitespace

ghost:
Removes p and br tags from incoming and outgoing statuses and lets client side css handle whitespace with newlines/spaces since it's already configured to do so.

### DIFF
--- a/app/assets/javascripts/components/components/status_content.jsx
+++ b/app/assets/javascripts/components/components/status_content.jsx
@@ -92,7 +92,7 @@ const StatusContent = React.createClass({
     const { status } = this.props;
     const { hidden } = this.state;
 
-    const content = { __html: emojify(status.get('content')).replace(/\n/g, '') };
+    const content = { __html: emojify(status.get('content')) };
     const spoilerContent = { __html: emojify(escapeTextContentForBrowser(status.get('spoiler_text', ''))) };
     const directionStyle = { direction: 'ltr' };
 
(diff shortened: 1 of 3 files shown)


Gargron:
There is absolutely no guarantee that HTML you receive from a remote status will contain any whitespace but the `<br />` and `<p>` tags.

rainyday:
Other servers like gnu social/postactive use whitespace for status formatting. The only instance im aware of that uses HTML for linebreaks is mastodon so this brings it in line with everything else.

Gargron:
I am right now considering whether merging #1397 was a mistake, as it clearly upset the balance of Everything Working Correctly. I feel like the expectation of paragraphs being delineated by newlines rather than html tags is misplaced in a content tag that is clearly marked as being of type html, rather than plaintext. Html does not respect newlines by default. That is why they must be encoded.

rainyday:
The current Mastodon CSS sets the status content to pre-wrap. The PR here: https://github.com/tootsuite/mastodon/pull/1663 removes the preformatted option so toots will render like normal HTML. To fix interoperability, use `<br />` tags  in place of `<p>` which GS also supports instead of newlines. The PR also converts every other space to &nbsp; at render time preserve intentional extra space on the part of the user to maintain expected behavior. Also stripped space around `<p>` tags from older instances since they tend to add extra space before and after html tags which could cause new paragraphs to be slightly indented due to spaces getting converted.

rainyday:
@gargron the problem is everything wasn't (and isn't) working correctly. If it was I wouldn't have made a pr

PR #1661, opened 2017-04-13
fix Portuguese translation

hugogameiro:
A couple of improvements and added a lot of missing values 
The en.jsx is also missing these values. I had to use fr.jsx to find the missing values because there were a lot of console notices and part of the web interface was still in English.

Still get a console notice:
[React Intl] Missing message: "getting_started.apps" for locale: "pt", using default message as fallback.
but have no idea of what getting_started.apps is because it only is used in the new zh-hk.jsx and I don't trust Google Translate.


rkarabut:
Just so you know, for now, you can find the variable names for default values by doing something like this:

```
tr@sflux:~/src/mastodon$ find -type f | xargs grep "Play sound"
./app/assets/javascripts/components/features/notifications/components/column_settings.jsx:    const soundStr = <FormattedMessage id='notifications.column_settings.sound' defaultMessage='Play sound' />;
```

So, say, translating "Play sound" (which, too, is unreferenced in any template aside from ru.jsx right now) would require placing `notification.column_settings.sound: "your translation"` into ./app/assets/javascripts/components/locales/pt.jsx .

Gargron:
@rkarabut Couldn't you grep for `defaultMessage`? Then if the results fit the `id` part as well, you'd get a full list of all localized strings in the JS files.

Gargron (approved):

rkarabut:
@Gargron Yes, of course, but I meant that for cases when you want to find a specific known string. Grepping for defaultMessage adds a lot of noise, you still need to filter out all the strings already present in the templates.

PR #1663, opened 2017-04-13
Normalize status whitespace between mastodon/gnu social

ghost:
Did some more checks, this should solve readability issues between mastodon/gnu social.


rainyday:
For clarification, Without this gnu social users can't see newlines in Mastodon and gnu social posts will have broken indentation when viewed in mastodon.

Gargron:
Isn't it better to patch GS to convert `<br />` to newlines on their end, if that's how they choose to represent that information?

rainyday:
They do support `br` but not `p` and right now mastdon uses both depending on if you put one or two newlines in your status. Switching to only using `br` solves interoperability issues.

yipdw:
FWIW, I've applied this patch on my own instance (which is running a modified `v1.1.2`) and it does indeed seem to fix whitespace formatting between GNU Social, Mastodon, and postActiv.  Here's an example:

- My Mastodon instance: https://sdfn-01.ninjawedding.org/@ninjawedding/49723
- GNU Social: https://quitter.no/notice/3811793
- postActiv: https://community.highlandarrow.com/notice/907565

If a pre-patch example is desired, I can back out the patch and make another update.

rainyday:
Bump

dcl:
Is there something blocking this change from being merged? It seems fairly simple, and would help a lot.

Gargron:
I don't understand why you aren't submitting a PR for GNU social to support `<p>` tags for paragraphs, one of the most basic tags ever. Why should we replace something semantic with stacked `<br />`s? Also, the commit that reverted #1397 clearly explained why exactly it was reverted: https://github.com/tootsuite/mastodon/commit/043862f4118a9ee5e6708ff9c144fa9e668ee06d #1397 visually broke a lot of stuff.

Let's say Mastodon and GNU social don't exist. Tell me why `one<br /><br />two` is more semantic than `<p>one</p><p>two</p>`?

app/assets/javascripts/components/components/status_content.jsx
```diff
@@ -93,7 +93,7 @@ const StatusContent = React.createClass({
     const { status } = this.props;
     const { hidden } = this.state;
 
-    const content = { __html: emojify(status.get('content')) };
+    const content = { __html: emojify(status.get('content').replace(/\s*(<p>|<\/p>)\s*/g, '$1').replace(/\s{2}/g, ' &nbsp;')) };
```
Gargron: If your changes remove `<p>` tags, what is this [regex-for-html](http://stackoverflow.com/a/1732454) doing here?

You also seem to be replacing two spaces with one non-breaking space. I am not aware of non-breaking spaces having a different width to normal ones, so I am puzzled by this.

app/lib/formatter.rb
```diff
@@ -14,8 +14,7 @@ def format(status)
 
     html = status.text
     html = encode(html)
-    html = simple_format(html, {}, sanitize: false)
-    html = html.delete("\n")
+    html = html.gsub(/(?:\n\r?|\r\n?)/, '<br />')
```
Gargron: Carriage return without line feed is not a newline, so why is `\n` optional in the second variant?

Gargron (changes requested):

app/lib/formatter.rb
rainyday: some older operating systems use \r as a new line. probably overkill in this day and age I admit.

app/assets/javascripts/components/components/status_content.jsx
rainyday: Maintains backwards compatibility with earlier versions of mastodon since not everyone is going to upgrade at the same time. Using an HTML parser seemed like overkill since we're just stripping whitespace next to certain patterns, not traversing a DOM or anything.

Using `&nbsp;` allows users to continue using custom spacing without relying on css `pre-wrap` which treats other characters besides ` ` as semantic.

Gargron: I know how non-breaking space works and why you only replace *starting* with two consecutive spaces. But you are replacing *two* spaces with *one* non-breaking.

I'm not seeing why stripping out `<p>` tags is necessary in this place, even if they're there. It's either the `<br />` from your version, or the `<p>` from older versions, but both should work the same, especially since we don't separate HTML tags with newlines there shouldn't be anything "extra" to strip out.

Gargron:
Behaviour like this is not acceptable. Your PR will not be merged from public callouts and ad-hominem attacks, only its technical merit. But that behaviour *will* disqualify you from contributing to the project, regardless of any technical merit, if you ever try doing that again.

app/assets/javascripts/components/components/status_content.jsx
rainyday: It's replacing two spaces with one regular space and one `&nbsp;` so the result is the same but without relying on `pre-wrap`

The regex does not strip out `p` tags, only whitespace that gets inserted before and after them in current versions of mastodon which can cause unintended indentation.

Gargron: Why not double `&nbsp;`?

Doing all of this in this JS piece, what happens to the public pages that don't use this code?

You may use [this](https://mstdn.jp/@Maxillaria/414078) as a landmark of what should not be broken.

Gargron: Actually, the linked toot is a bad example since it uses full-width spaces which seem to be preserved regardless of any CSS whitespace settings.

rainyday: turning every space into `&nbsp;` will prevent lines wrapping, but I do see what you mean with public timelines. I can add space formatting this to the PR in formatter.rb if that's okay with you, it's less brittle than using `pre-wrap` which will convert a string like `html <br /> string` to

```
html 
…string
```

instead of what's expected.

kevinmarks:
Why are you using retro xml style close tags? You're not using xhtml atom feeds so 
''''
<p> and <br> not <p></p> and <br />

Gargron:
I talked to MMN about this and he agreed with me that `<p>` is more semantic and GS should accept it. The reason GS doesn't itself use `<p>` tags is because it's simply putting text through the `nl2br` function in the standard php library. Sorry!

yookoala:
Submitted an issue here:
https://git.gnu.io/gnu/gnu-social/issues/272

PR #1677, opened 2017-04-13
Exclude javascript locale file from dup check

yookoala:
* Exclude javascript locale files form Code Climate's duplication
  engine. It is silly to have duplication check with locale files.
  They are supposed to look similar.

* Prevent unnecessary blocking for translation updates (like #1661)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,6 +1,8 @@
 engines:
  duplication:
    enabled: true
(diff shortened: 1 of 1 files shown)


rkarabut:
https://github.com/tootsuite/mastodon/issues/1052 can finally be closed then :D

Gargron (approved):

PR #1683, opened 2017-04-13
Use input type `number` for Two-factor code

watilde:
Since the two-factor code is always in number, it's better to use `<input type='number'>` instead of `text`.

### DIFF
--- a/app/views/auth/sessions/two_factor.html.haml
+++ b/app/views/auth/sessions/two_factor.html.haml
@@ -2,7 +2,7 @@
   = t('auth.login')
 
 = simple_form_for(resource, as: resource_name, url: session_path(resource_name), method: :post) do |f|
-  = f.input :otp_attempt, placeholder: t('simple_form.labels.defaults.otp_attempt'), input_html: { 'aria-label' => t('simple_form.labels.defaults.otp_attempt') }, required: true, autofocus: true, autocomplete: 'off'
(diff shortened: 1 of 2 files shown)


app/views/auth/sessions/two_factor.html.haml
```diff
@@ -2,7 +2,7 @@
   = t('auth.login')
 
 = simple_form_for(resource, as: resource_name, url: session_path(resource_name), method: :post) do |f|
-  = f.input :otp_attempt, placeholder: t('simple_form.labels.defaults.otp_attempt'), input_html: { 'aria-label' => t('simple_form.labels.defaults.otp_attempt') }, required: true, autofocus: true, autocomplete: 'off'
+  = f.input :otp_attempt, :type=>'number', placeholder: t('simple_form.labels.defaults.otp_attempt'), input_html: { 'aria-label' => t('simple_form.labels.defaults.otp_attempt') }, required: true, autofocus: true, autocomplete: 'off'
```
Gargron: Please use `type: :number` here (new Ruby hash syntax)

Gargron (changes requested):

app/views/auth/sessions/two_factor.html.haml
watilde: Opps, sure! I'm gonna update it.

Gargron (approved):

PR #1679, opened 2017-04-13
Added Italian language (it)

matteoaquila:
Added localization for Italian language (it)

PR #1694, opened 2017-04-13
Italian files are now in the right place

matteoaquila:
I made a mess and put the Italian localisation files in the wrong place...
I'm so sorry, please forgive me 😭
Now those files are finally il config/locales as they should...
And I've updated settings_helper.rb and application.rb as @yookoala told me (#1679). Still working on it.jsx and all the other Javascript stuff.


Gargron (approved):

app/helpers/settings_helper.rb
```diff
@@ -7,6 +7,7 @@ module SettingsHelper
     es: 'Español',
     eo: 'Esperanto',
     fr: 'Français',
+    it: 'Italiano'
```
Gargron: Missing `,` at end of line, syntax error

Gargron (changes requested):

app/helpers/settings_helper.rb
matteoaquila: Come on, one day I'll finally be able to submit something **totally** correct, I'm learning, never lose hope. 😂 ❤️ 
(Thank you for the correction)

yookoala:
@matteoaquila: You have syntax error in your `devise.it.yml` file.

Same issue. According to YAML specifications, all strings start with these characters must be quoted
`:`, `{`, `}`, `[`, `]`, `,`, `&`, `*`, `#`, `?`, `|`, `-`, `<`, `>`, `=`, `!`, `%`, `@`, `

Therefore,
1. [line 11](https://github.com/tootsuite/mastodon/blob/efcf9448da60574a41d79964cece87008667cb41/config/locales/devise.it.yml#L11) cannot have unquoted strings start with symbol. Need to quote this value with `"`.
1. [line 14](https://github.com/tootsuite/mastodon/blob/efcf9448da60574a41d79964cece87008667cb41/config/locales/devise.it.yml#L14) cannot have unquoted strings start with symbol. Need to quote this value with `"`.

Gargron (approved):

yookoala (commented):
You need to quote the whole value. Not just the variable.

E.g.

    "%{authentication_keys}" o password invalida.

should be 

    "%{authentication_keys} o password invalida."

PR #1696, opened 2017-04-13
Apply `i18n-tasks normalize` to locales

Gargron:


config/locales/fr.yml
```diff
@@ -39,6 +39,30 @@ fr:
     posts: Statuts
     remote_follow: Suivre à distance
     unfollow: Ne plus suivre
+  admin:
+    settings:
+      click_to_edit: Cliquez pour éditer
+      contact_information:
+        email: Entrez une adresse courriel publique
```
wxcafe: "addresse mail" is more common in french

tcitworld: Didn't someone changed them all to `courriel` elsewere some time ago ?

wxcafe: Ah, might be, I didn't see that. I agree we need consistency... I just think email is more common. But if everything else says courriel let's keep that I guess.

tcitworld: My guess is that at some we'll have a dedicated system to translate locales (transifex, weblate, pootle, others...) so the big decisions like this one will be taken by people who know the subject there. :)

koboldunderlord (approved):
LGTM +1

config/locales/fr.yml
Gargron: This PR is solely about normalizing the order of locale keys. You can and should do it yourself by running `i18n-tasks normalize` after translating something so a PR like this wouldn't be needed.

PR #1699, opened 2017-04-13
Fix #1220, fix #1671 - Hook up comment box to the Redux comment value

### LINKED ISSUE #1220: Reporting "comment" textbox repeatedly clears text.
When trying to report a user for their abusive posts, I am prompted to write comments. The issue is that as I type in the comment box, the text I have written is repeatedly deleted (I'd say every second or two) making it [almost] impossible to ensure that the comments are indeed submitted in the report.

* * * *

- [X ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
https://github.com/tootsuite/mastodon/issues/774

### LINKED ISSUE #1671: Report comment is never sent
When reporting a user, whatever is input into the report comment box is never sent to the server and reports always show up with an empty comment.

PR at #1672  

* * * *

Gargron:
Fix username styling regression introduced in #1063
Fix report screen background regression introduced in #1415

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -95,6 +95,6 @@ class Formatter
   end
 
   def mention_html(match, account)
-    "#{match.split('@').first}<span class=\"h-card\"><a href=\"#{TagManager.instance.url_for(account)}\" class=\"u-url mention\">@#{account.username}</a></span>"
+    "#{match.split('@').first}<span class=\"h-card\"><a href=\"#{TagManager.instance.url_for(account)}\" class=\"u-url mention\">@<span>#{account.username}</span></a></span>"
   end
 end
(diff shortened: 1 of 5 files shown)


app/lib/formatter.rb
```diff
@@ -95,6 +95,6 @@ def hashtag_html(match)
   end
 
   def mention_html(match, account)
-    "#{match.split('@').first}<span class=\"h-card\"><a href=\"#{TagManager.instance.url_for(account)}\" class=\"u-url mention\">@#{account.username}</a></span>"
+    "#{match.split('@').first}<span class=\"h-card\"><a href=\"#{TagManager.instance.url_for(account)}\" class=\"u-url mention\">@<span>#{account.username}</span></a></span>"
```
vahnj: Other than default span formatting, does adding <span> here do anything?  Maybe worth adding id=report-username or something like that

koboldunderlord (approved):
tiny comment, but LGTM +1

app/lib/formatter.rb
Gargron: Well, the h-card stuff has been added by microformat folks, but the span inside the a has a visual role (underline it on hover but do not underline the @ sign)

PR #1700, opened 2017-04-13
Fix #1609, fix #1628 - Revert #1397

### LINKED ISSUE #1609: Excessive line spacing in posts due to extra paragraph marks between <p> tags
Example: https://mastodon.social/users/Rushyo/updates/1800557 
Posts seem from instances on the latest version seem to be putting an extra paragraph mark in the post text between < p > tags, resulting in excessively large line spacing in most browsers. To clarify, it looks like this is happening: 
< p > paragraph 1 </ p > 

< p > paragraph 2 </ p > 

When really this should happen: 
< p > paragraph 1 </ p >
< p > paragraph 2 </ p >

See the following screenshot:
screen shot 2017-04-12 at 10 31 56 am

I might also suggest lowering the margin-bottom for paragraphs in posts to 10px, but that's just my opinion. I don't know enough about how the posts are made or saved to fix this myself, unfortunately. 

* * * *

- [✅ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

### LINKED ISSUE #1628: Regression in newlines on profile page

Gargron:
When transmitting data in a HTML-encoded element like <content type="html" />,
relying on newlines being preserved is not wise, since HTML by itself
does not care for newlines - it cares for `<p>` and `<br>`

Additional fix: reset NSFW toggle after sending toot

### DIFF
--- a/app/assets/javascripts/components/components/status_content.jsx
+++ b/app/assets/javascripts/components/components/status_content.jsx
@@ -92,7 +92,7 @@ const StatusContent = React.createClass({
     const { status } = this.props;
     const { hidden } = this.state;
 
-    const content = { __html: emojify(status.get('content')).replace(/\n/g, '') };
+    const content = { __html: emojify(status.get('content')) };
     const spoilerContent = { __html: emojify(escapeTextContentForBrowser(status.get('spoiler_text', ''))) };
     const directionStyle = { direction: 'ltr' };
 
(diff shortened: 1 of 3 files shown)


koboldunderlord (approved):
LGTM +1

PR #1702, opened 2017-04-13
Translate admin

tcitworld:
Finish #1549.


tcitworld:
As I've moved the `reports` section under `admin`, other translations are affected.
Also, will conflict #1696, so maybe get this in first. :)

Gargron (approved):

PR #1697, opened 2017-04-13
Fix minor jslint issue

yookoala:
* Undefined propType in react component `ColumnSettings`.
  Add proper PropTypes definition.

PR #1706, opened 2017-04-13
Fix "TypeError: Cannot read property 'isRequired' of undefined"

henrycatalinismith:
Looks like a recent commit introduced a reference to `ImmutablePropTypes.object.isRequired`, which looks absolutely fine on the surface, but turns out not to be a thing. This uncaught TypeError happens immediately on page load.

```
Uncaught TypeError: Cannot read property 'isRequired' of undefined
    at Object.93.../../../components/column_collapsable
```

screen shot 2017-04-13 at 18 15 31

All better now though 😄

### DIFF
--- a/app/assets/javascripts/components/features/notifications/components/column_settings.jsx
+++ b/app/assets/javascripts/components/features/notifications/components/column_settings.jsx
@@ -27,7 +27,7 @@ const ColumnSettings = React.createClass({
 
   propTypes: {
     settings: ImmutablePropTypes.map.isRequired,
-    intl: ImmutablePropTypes.object.isRequired,
+    intl: React.PropTypes.object.isRequired,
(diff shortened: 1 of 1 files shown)


rtucker:
#1697 (commit ce80d0b) was the breaking change, and this commit does fix it in my test environment

yookoala:
Sorry to have missed that out.

hnrysmth:
@yookoala lol no worries, when I figured out what was up it took me another few minutes to really believe it cos it just seems so surprising that that wouldn't be there! 😂

kinda tempted to look into seeing if it'd be possible to throw a PR together for that package that at least improves the error message or something

asonix:
It looks like this change is already in master

hnrysmth:
@asonix good spot, looks like it got fixed directly in there at about the same time as I sent this 😄

PR #1709, opened 2017-04-13
Add Support for Nanobox

danhunsaker:
Nanobox is a tool for creating consistent (but still disposable) development environments, deploying code to remote servers (with support for AWS [official], Digital Ocean [official], Linode [official], and Proxmox [via third party adapter], with others in development), and for ensuring that the development environment mirrors the production environment as closely as possible, so there are minimal surprises during deployment.

This PR adds support for developing Mastodon in a Nanobox environment, and/or for pushing it into a production Nanobox environment. Development with Nanobox doesn't require deploying via Nanobox, nor does deploying with Nanobox require developing with it. There are advantages to using it for both, though.


lib/tasks/db.rake
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+namespace :db do
```
ineffyble: This is the main change I can see that could affect non-Nanobox users/functionality.

Can you expand on the methodology behind this? Looks like it's overriding the default `db:migrate`

danhunsaker:
It's adding an additional command to the `db:migrate` namespace, as I understand it. Of course, I might be wrong about that, in which case the command could be renamed to sit outside that namespace. Given that the command itself still works, though, and actively references `db:migrate` to do it, I'm pretty sure it's not overriding anything.

lib/tasks/db.rake
danhunsaker: It's adding an additional command to the `db:migrate` namespace, as I understand it. Of course, I might be wrong about that, in which case the command could be renamed to sit outside that namespace. Given that the command itself still works, though, and actively references `db:migrate` to do it, I'm pretty sure it's not overriding anything.

danhunsaker: Does that address your concerns?

Gargron: Sorry this has been waiting for so long, it's partially because I haven't heard of Nanobox before this PR. Regarding this task.. Is it necessary to check the db existence "before_live"? Could it not be something that runs only once when the nanobox is being created?

danhunsaker: Nanobox doesn't support tasks that only run once on first deploy, unfortunately. I'm also not aware of a mechanism for determining whether the current deploy is the first (successful) one or not, other than the one given here which has Rake simply try to access the DB. I could try to approach this from a different angle that figures out whether this is an initial deploy or not some other way, but this is the cleanest approach since Rake tasks have access to the DB with the same settings as the rest of the application, without having to manually set them up externally.

The other alternatives I can think of, at the moment, are:

- make users do the initial DB setup manually (which is a bit involved, defeats some of the purpose of automated deploy tasks, and may actually work incorrectly given the automated deploy tasks would then run a `db:migrate` before the `db:setup` has a chance to be run)

- rename the task defined here so it doesn't sit inside an existing namespace (which would have no functional difference, but might be less scary to merge?)

Near as I can tell, though, this is the simplest and most effective option for running the `db:setup` task only when it's actually needed. So unless you really want me to change something, I'd feel safer leaving this as is.

bradurani:
Doesn't Docker give us reproducible production-like development environments? Nanobox doesn't seem to be a very well-known or widely-used tool. I suggesting trusting admins / developers to choose their own tooling, rather than trying to include every useful tool in this repo.

danhunsaker:
According to the Docker section of the documentation, using Docker in development is *not* supported, here. There are a number of elements that Docker just doesn't handle properly on its own. Nanobox does actually use Docker under the hood, but also handles all the song and dance needed to get development features working properly, without the dev having to know a single thing about Docker itself to get it going.

As to not including every possible deploy target in the repo, I can't speak to that. I agree that Nanobox isn't terribly widespread, yet, but I could argue that for Scalingo, too (I'd never heard of it before Mastodon), and Nanobox's spread is mostly so minimal due to being so recently out of private beta (I've been using it for about a year, but it's been publicly available only since February, IIRC). That part's not my call, but I thought it might be useful to at least offer the option, since initial configuration is actually a bit involved (as evidenced by the number of files involved, here), but actual use is so simple.

Gemfile
```diff
@@ -3,7 +3,9 @@
 source 'https://rubygems.org'
 ruby '>= 2.3.0', '< 2.5.0'
 
-gem 'pkg-config'
```
beatrix-bitrot: i'm concerned this will affect other things, can you tell me more about what's happening here/why?

danhunsaker: Sure.

So, it turns out that `pkg-config` isn't smart enough to properly find the right libraries when used with Nanobox. Since it is only included for building `nokogiri`, and it's possible to manually provide the correct paths while installing that, leaving it out of Nanobox deploys is preferable (especially since the manual config seems to be flat ignored if `pkg-config` is present).

This change lets me exclude `pkg-config` from the install list. My understanding is that, unless explicitly excluded, all package groups are treated as dependencies, so this shouldn't affect other platforms at all. That's part of the reason for using a unique name for the group - to prevent it from affecting any users except those who exclude it intentionally.

I am more than happy to use another approach instead, if anyone is aware of one.

beatrix-bitrot: That all seems very reasonable. This seems to be a minimally invasive changeset as far as non-nanobox installs go, and there are no merge conflicts present, so I am going to approve this. 

If you haven't rebased and tested this lately I recommend doing so in case there is anything that you need to change that isn't obvious/a merge conflict. That way if you find anything you can add it to this PR before it's merged (:

beatrix-bitrot (approved):
my only specific issue was addressed, and it seems that issues raised by other reviewers were a) also addressed and b) will only affect those using nanobox.

my overall concern is that it seems not a lot of people have experience with nanobox and so this code may be difficult to maintain. ultimately i don't think that's a reason not to approve this changeset though, so, looks good to me!

danhunsaker:
Rebased and tested, but it turns out the Nanobox team needs to build an updated `protobuf` package to get things working again with the current `cld3` Gem. That should happen in the next 12 hours or so, and I'll test again as soon as I can manage, then update here once things are stable again.

As to maintenance, I'm more than willing to continue to track changes needed to work with Nanobox for the foreseeable future.

danhunsaker:
Got everything resolved. finally. This is good to go!

Gargron:
Sorry for making you wait.

danhunsaker:
No worries! There's a lot going on.

PR #1710, opened 2017-04-13
Make contents of boost confirmation dialog scroll

tomfhowe:
Addresses https://github.com/tootsuite/mastodon/issues/1703 
This is necessary for boosting long posts on small screens - see this post for video: https://slime.global/@masklayer/102308


tomfhowe:
Example of a post that is impossible to boost from a small screen/window with the boost confirmation dialog on in it's current non-scrolling state: https://gnusocial.no/notice/1940320

Gargron (approved):

PR #1715, opened 2017-04-13
Fix text wrapping in buttons in some translations

tomfhowe:
Fixes https://github.com/tootsuite/mastodon/issues/1686. 
The max-width is not 100% necessary, but makes the button in the translation shown in that issue look better (without the max-width I'm adding, it is too wide to fit in the compose page) and not cause horizontal scrolling issues on small screens. 
Does pose an issue if there are ever any buttons wider than 75px that don't have that max-width overridden with their own class or inline style - but I don't think there currently are any.

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -18,7 +18,7 @@
   font-weight: 500;
   letter-spacing: 0;
   text-transform: uppercase;
-  padding: 0 6px;
+  padding: 0 8px;
   height: 36px;
   cursor: pointer;
   line-height: 36px;
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -25,6 +25,8 @@
   border-radius: 4px;
   text-decoration: none;
   transition: all 100ms ease-in;
+  white-space: nowrap;
+  max-width: 75px;
 
   &:hover, &:active, &:focus {
     background-color: lighten($color4, 7%);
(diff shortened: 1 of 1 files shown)


lae:
This still looks awkward/haphazard. It actually looks awkward for English as well prior to this fix (and still looks the same after this fix)


Could we maybe move the character count?

tomfhowe:
@lae Oh ok yeah, it didn't look so bad for me but I think I might have text scaling different or something in the browser. Maybe we can change padding on that button.

tomfhowe (commented):
Removed the max-width, since it wasn't a good solution really, decreased the padding a bit to address previous comments. Ideally, I would want to move the character count as mentioned, but it doesn't seem to have an id or a class and I don't know how to add one.

tomfhowe:
Now looks like this:
screen shot 2017-04-13 at 3 52 21 pm
A little better?

tomfhowe (commented):
That was too much.

tomfhowe (commented):
Move the character count over and add back some padding. Maybe this isn't a good solution anymore... idk

tomfhowe:
screen shot 2017-04-13 at 3 59 03 pm
Okay, well, I'm going to stop changing this, maybe someone else can make it better/less weird!

tomfhowe (commented):
Add a tiny bit of padding back. I'm done messing with this now I swear (sorry)

Gargron:
I believe this is no longer needed, thank you

PR #1729, opened 2017-04-13
Implementing Subresource Integrity

maxypy:
This PR allows automatic generation of integrity hash on all attached scripts.

Example : 
```
<script src="/assets/application-9156d64f4acb71c1d7a6e2b8ff8ea6750f008a5b85ae573e2718e9a3b9d5a032.js" integrity="sha256-kVbWT0rLccHXpuK4/46mdQ8AiluFrlc+Jxjpo7nVoDI="></script>
```

This is considered as good security practice by [Mozilla's Observatory](https://observatory.mozilla.org/) and [W3C](https://www.w3.org/TR/SRI/).
This modification requires a new gem : sprockets-rails.


Gargron:
Are you sure sprockets-rails is a new gem? Rails does use sprockets by default, and the Gemfile.lock doesn't look like you pulled a new version in. This is a good change, I think

Gargron (approved):

PR #1730, opened 2017-04-13
Remove unused methods

mjankowski:
Remove unused methods:

- api controller - `set_counters_maps` and `set_accounts_counters_maps`
- `Account#followers_domains`
- `User.prolific`

Also:

- Add coverage for `Account.triadic_closures`
- Add rake task to list all admin users

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -100,11 +100,4 @@ class ApiController < ApplicationController
     @reblogs_map    = Status.reblogs_map(status_ids, current_account)
     @favourites_map = Status.favourites_map(status_ids, current_account)
   end
-
-  def set_account_counters_maps(accounts) # rubocop:disable Style/AccessorMethodName
-    account_ids = accounts.compact.map(&:id).uniq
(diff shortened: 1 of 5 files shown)


expenses:
Might any of these be used in the future though?

mjankowski:
I'd generally prefer to not have unused code lying around ... it's easy enough to write code if we need it, and the git history is always there if we remember that we used to have something and want to see what it was.

In this case specifically, I'd like input into whether `Account.triadic_closures` and `User.admins` and `User.prolific` might be in place for use via the console or CLI. If they are, we should add spec coverage and/or rake task access and/or web UI for them.

The `set_*counters*` methods I believe used to be used and now are not.

Gargron:
I am not sure if `User.admins` is truly unused, but `Account.triadic_closures` is something I'd rather not remove because it took a while to figure out the correct SQL query. If we wanna re-add follow suggestions at some point, that's the method to power it.

The counters are deprecated though, you are correct.

mjankowski:
Thanks for clarification. I'll add ui/rake usage for `admins` and triadic and update this to just remove the unused ones.

mjankowski:
PR description and commits updated to reflect methods which are staying vs going.

Gargron (approved):

Issue #1736, opened 2017-04-14
docker-compose build results in .env.production error


ian-kelling:
Following the instructions of the readme, you get
```
docker-compose build
ERROR: Couldn't find env file: /a/hopt/mastodon/.env.production
```
This is easily resolved by doing the next step of creating that file first.

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1737, opened 2017-04-14
Fix getting started commands out of order

ian-kelling:
In current order, docker-compose build results in
ERROR: Couldn't find env file: /a/hopt/mastodon/.env.production


ian-kelling:
Fixes #1736

Gargron (approved):

Issue #1130, opened 2017-04-07
Invalid hostname sent on SMTP helo


plinss:
When sending confirmation emails, the SMTP helo (or elho) is using an invalid hostname, from my debugging logs it appears to be "unknown[&lt;actual IPv6 address&gt;]" rather than the server's hostname. This is causing email to be rejected by my Postfix's "reject_invalid_hostname" filtering rule.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

joshlbs:
Well done for logging this!
I had the same issue and, as a workaround, adding SMTP_DOMAIN=example.com to .env.production does the trick.

plinss:
Tried that and didn't help

wxcafe:
I'm pretty sure this is an issue with either your `/etc/hostname` or your `/etc/hosts`. Try working with these and see if that changes your outbound email domain.

plinss:
`/etc/hostname` and `/etc/hosts` are properly setup. Running:
`hostname -f`
yields the proper fqdn.

R0ckweb:
As joshlbs says, adding [SMTP_DOMAIN to .env.production](https://github.com/tootsuite/mastodon/blob/master/config/environments/production.rb#L101) solve this problem !

According to [the mailer documentation ](http://edgeguides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration) this config force a HELO domain.
Tested and approved.

plinss:
Adding SMTP_DOMAIN to .env.production did not solve the problem for either of the two instances I manage

ian-kelling:
@plinss, did you set SMTP_DOMAIN to a domain that would resolve to an ip you control, then restart mastodon? Setting SMTP_DOMAIN worked for me. What exact error message do you get?

ineffyble:
Hey all,

Just made a pull request to fix a bug that I discovered - the SMTP domain should in theory be set from the SMTP_DOMAIN environment variable, and fallback to the LOCAL_DOMAIN variable otherwise.

The problem was, the fallback wasn't working, it was referencing an internal variable that wasn't yet set. That would explain some of the issues mentioned.

plinss:
I was setting SMTP_DOMAIN in the .env.production file, not as an environment variable, perhaps that's why it didn't work for me. I wasn't getting an error from Mastodon, but a rejection from my smtp server due to malformed helo hostname.

Fixing the default sounds like the proper fix.

PR #1738, opened 2017-04-14
Add documentation of SMTP_DOMAIN

ian-kelling:
Without setting it, exim will reject the mail with a message like:

    rejected EHLO from [10.20.0.1]: syntactically invalid argument(s): {}


ian-kelling:
This helps people avoid issues like  #1130. Someone should take a look at what it defaults too, as it seems to me that setting it to LOCAL_DOMAIN is better. Right now, I see it's set to config.x.local_domain, but I don't know what that resolves to, and I'm not bothering to find out.

ineffyble:
Hey,

It should actually already set to LOCAL_DOMAIN.

`config.x.local_domain` is set to the LOCAL_DOMAIN by default, falling back to `localhost:PORT`. I'm just going to confirm this is working as intended, will comment again.

ineffyble:
Except that's not the case, because `config.x.local_domain` isn't set at the point where it's referenced... I suspect this may have changed/broken at some point in the past.

This actually explains a few other reported issues. I'm making a patch now.

ian-kelling:
I updated my pull request, simplifying the documentation, as this is rarely going to need to be set.

ineffyble (approved):
Looks good to me, approving for merge.

Gargron (approved):

PR #1741, opened 2017-04-14
Use mysql2 driver instead of postgresql.

furoshiki:
Fix migrate files to remove dependency of postgresql column type.


furoshiki:
Maybe next month.

Gargron:
We use a number of Postgres-specific SQL features. MySQL support will not happen, sorry.

PR #1536, opened 2017-04-11
Add a way to bind streaming service to localhost

maurice-audin:
adding Environment="BIND=127.0.0.1" to systemd service

PR #1747, opened 2017-04-14
Change default bind host to localhost

### LINKED ISSUE #1283: Default bind address should be 127.0.0.1
The mastodon-web and mastodon-streaming services appear to be binding to 0.0.0.0 and [::] respectively, which makes them bind to all available addresses and are publicly visible, even though they are meant to be proxied through Nginx.

It would be better to bind to 127.0.0.1 (and/or [::1]) by default, making them only accessible locally. At the least, documentation of a bind address configuration option for each would be great.

* * * *

pfigel:
Change the default bind host for web workers and streaming to localhost instead of 0.0.0.0. A new environment variable called HOST allows the bind host to be changed for deployments where load balancers are running on a separate host.

This is a combination/replacement of/for #999 (which seems to have gotten lost in the development branch) and #1536.

Fixes #1283

Note: I used 127.0.0.1 in `docker-compose.yml` because using `localhost` seems to cause docker to listen on `*:3000` instead, whereas `127.0.0.1` makes it bind to `localhost:3000` (as expected), both on IPv4 and IPv6. If someone more knowledgeable in docker thinks `localhost` would be more appropriate here and that this is somehow specific to my docker host (macOS 10.12.4, docker version 17.03.1-ce-mac5 (16048)), let me know.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,7 +1,7 @@
 threads_count = ENV.fetch('MAX_THREADS') { 5 }.to_i
 threads threads_count, threads_count
 
-port        ENV.fetch('PORT') { 3000 }
+port        ENV.fetch('PORT') { 3000 }, ENV.fetch('HOST') { 'localhost' }
 environment ENV.fetch('RAILS_ENV') { 'development' }
 workers     ENV.fetch('WEB_CONCURRENCY') { 2 }
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
     env_file: .env.production
     command: bundle exec rails s -p 3000 -b '0.0.0.0'
     ports:
-      - "3000:3000"
+      - "127.0.0.1:3000:3000"
     depends_on:
       - db
       - redis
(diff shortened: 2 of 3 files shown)


wxcafe:
> 127.0.0.1 makes it bind to localhost:3000 (as expected), both on IPv4 and IPv6. 

what the f*ck is up with docker, seriously?

This looks good to me apart from that docker being weird as heck

wxcafe (approved):

config/puma.rb
```diff
@@ -1,7 +1,7 @@
 threads_count = ENV.fetch('MAX_THREADS') { 5 }.to_i
 threads threads_count, threads_count
 
-port        ENV.fetch('PORT') { 3000 }
+port        ENV.fetch('PORT') { 3000 }, ENV.fetch('HOST') { 'localhost' }
```
Gargron: I am not sure if we need an env variable for host.. I think command-line arguments will override these options anyway, so that could be used instead of HOST=0.0.0.0 or whatever.

Gargron (changes requested):

Gargron:
Closing due to inactivity.

PR #1749, opened 2017-04-14
Adding first arabic string localization

ghost:
Hi,
I'm adding first arabic string localization for #Mastodon. Can we see how this RTL language behave on Mastodon before pushing more strings ?
Thank you.


ineffyble:
@Gargron Should we be making Code Climate ignore duplication in localisation files?

BoFFire:
Closing this PR for cleaning. I will submit another one.

PR #1752, opened 2017-04-14
Fix #1401 Add Instance API Version

### LINKED ISSUE #1401: Add the Instance API Version in GET /api/v1/instance
When writing a client you need to know what are the abilities of an Instance.
All instances will not update to the last update automatically so we need to know what version is in place.

GET api/v1/instance is the first call we make and can offer instance specifications, options/plugins infos.

Ex: Build 1.1.1 is available with public URI, don't know which one are up to date.

* * * *

I saw API versioning not instance deployment version

allanbreyes:
Fixes https://github.com/tootsuite/mastodon/issues/1401

This PR proposes a way to produce the build artifact `.version` via `rake mastodon:version:write`, and expose it via `/api/v1/instance`. It pulls the version from `git` tags at _build time_, but can fall back to `git` at _runtime_ if the executable is present.

Seems like there are a ton of different deploy options, so I didn't want to prescribe when/where to run that Rake task. Awesome project; hope this helps!

(Also added a few lines to the `bin/` scripts for initial setup.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ redis
 # Ignore vim files
 *~
 *.swp
+
+# Ignore version build file
+.version
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -39,6 +39,12 @@ class Setting < RailsSettings::Base
       records
     end
 
+    def version(fresh: false)
+      file = Rails.root.join('.version')
+      return File.read file if File.exist?(file) && !fresh
+      `git describe --tags`.strip if system('git --version')
+    end
+
     private
 
     def default_settings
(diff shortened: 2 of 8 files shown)


Gargron:
I mean, the idea makes sense but why not have the VERSION be a constant in the source code? And then just bump it when a new tagged release is coming. That's what I was planning to eventually do

allanbreyes:
@Gargron that works, too. My gripes with persisting a version file/constant:

1. It's more steps, and the actual commit that bumps it doesn't match the merge commit that's tagged for release.
2. Deploying and/or testing a build between releases/tags won't show the actual version (just the latest) vs. this showing `v1.1.2-3-gd7f673e`.

Your call, LMK or close at will. Thanks!

PR #1753, opened 2017-04-14
Code Climate - Change gsub to delete in formatter.rb

ineffyble:
Change usage of gsub to delete, as per Code Climate/Rubocop recommendation


Gargron (approved):
Hmm, looks like I was simply using the wrong quotation marks when I tried doing this when I first got that Rubocop warning and it didn't produce an equivalent result...

Issue #1394, opened 2017-04-10
Direct message still says "Boost" in tooltip


James-Firth:
Reproduction steps:
* Send Direct message
* View Direct message
* Hover over envelope Icon

Bug Result: "Boost" tooltip text should appear.

Expected Result:
Something like "Cannot boost a direct message"?
* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

rtucker:
i believe this is fixed in at least one or more of the above commits; 1.3.2 does not exhibit this behavior.

PR #1754, opened 2017-04-14
Tooltip for boosting depends on toot visibility (#1394)

expenses:
Fix for #1394. Looks like:
example
Note that this adds a new localization string.


thurloat:
There's an inconsistency between your default translation, and the English translation.

`app/assets/javascripts/components/locales/en.jsx` calls it a Toot, though there are other translations in the file that refer to toots as Posts. i.e. `"compose_form.privacy_disclaimer"`.

`app/assets/javascripts/components/components/status_action_bar.jsx` calls it a Post.

expenses:
@thurloat I think they're meant to be different.

yiskah (approved):
This looks to add a useful a11y feature

Gargron:
@expenses @thurloat For the record, the official UI does not refer to any items as "toots". They are "statuses" in most places, and "posts" in places where it needs to be shorter. Only the "publish" button is called "toot". This is deliberate, the UI can't get *too* silly.

app/assets/javascripts/components/locales/en.jsx
```diff
@@ -119,6 +119,7 @@ const en = {
   "status.mention": "Mention @{name}",
   "status.open": "Expand this status",
   "status.reblog": "Boost",
+  "status.cannot_reblog": "This toot cannot be boosted",
```
Gargron: Change this to status and this is good to go. Sorry it took so long

Gargron (changes requested):

app/assets/javascripts/components/locales/en.jsx
expenses: .. And done!

Gargron (approved):

PR #1759, opened 2017-04-14
Update Traditional Chinese, HK (zh-HK) translation

yookoala:
* Add missing translations.
* Update translation based on feedback on v1.1.2.


Gargron (approved):

Artoria2e5:
Hi @yookoala, are you interested in upstreaming the doorkeeper translation to https://github.com/doorkeeper-gem/doorkeeper-i18n? Since Mastodon's AGPL license is stricter than Doorkeeper's MIT (Expat) license, I don't have the legal power to loosen your copyright by sending it directly to doorkeeper.

yookoala:
@Artoria2e5: There you go: doorkeeper-gem/doorkeeper-i18n#30 :-)

Artoria2e5:
Hi @yookoala, it's me again. Thanks for the Doorkeeper PR! Are you interested in dropping https://github.com/tigrish/devise-i18n an issue with a link to the current [devise.zh-HK.yml](https://github.com/tootsuite/mastodon/blob/master/config/locales/devise.zh-HK.yml)? I guess they can use some zh-HK translation from here...

yookoala:
@Artoria2e5 Thanks for the invitation. It seems that there is already a [zh-HK](https://github.com/tigrish/devise-i18n/blob/master/rails/locales/zh-HK.yml) locale file there.

The existing translations there are much closer to the Cantonese spoken language. Whereas I use a written Chinese approach here. Mixing the translations style would be in-coherent. Overwriting their translation would be impolite, but I would keep the translation here as is to make the whole platform's translation coherent.

So I don't think we can use the same translations here as upstream. And it don't seems appropriate to backport, either :-(

Issue #861, opened 2017-04-04
Add a "Local timeline" privacy option


alex73630:
It would be useful for instance admins to send toots only to instance users.

It would look like this:

![image](https://cloud.githubusercontent.com/assets/7538405/24667620/60a7e538-1964-11e7-8aee-6f4ad1477380.png)

Is it possible in Mastodon ?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

yiskah:
It's possible but undesirable as it would promote users siloing and piling onto the same instance. This centralizes what should be a de-centralized distributed system. Users should not be afraid of migrating accounts due to missing out on particular content, this defeats a major point of federation.

alex73630:
Well, I would like to have this to toot when I'm doing an update without "spamming" federated timelines, as this only concerns my instance and not other ones.

seascape:
As a user (not admin) I would very much like the ability to occasionally post content only to users of my instance. This feels like an obvious missing feature to me.

I see your criticism, yiskah. Two thoughts: 

1) I don't think the existence of such a feature would lead to a sudden huge movement toward local-only posting. People, myself included, love Mastodon's federation; it's one of its key and strengths differentiators.

2) Allowing that some instances did choose to turn insular... why not? A few instances of many. The software can be used different ways by different people with different needs/wants/desires.

My own use case: marginalized identity, and fear external (originating outside Mastodon) harassment. I want to participate more openly (posting selfies I wouldn't want widely disseminated, for example) but hesitate to do so on an essentially worldwide basis. 

But my instance? It's small, well run, and I feel enough trust there. Local posting reduces tens of thousands of randos to tens of hundreds of self-selected people who tend to be more like me, sympathetic to me, supportive of me. 

Is my local-only content totally safe from bad actors? Of course not. But there would still be an overall increase of safety and trust (less randos) and I'd be able to feel pretty OK about the risk profile of sharing sensitive content when the occasion arises.

zyphlar:
I think this is a critical feature of federation: choosing how far to broadcast certain stuff. Otherwise certain sensitive or insular groups (think: an instance for a church, or for an addiction support group) will be forced to completely de-federate and/or block all other instances, which is an overreaction.

[...]

PR #1761, opened 2017-04-14
Add "local" visibility for public messages addressed only to instance members (& change default visibility by "local")

alkemist:
(Sorry my English level is too low)

Actuellement pour parler à de nouvelles personnes et faire de nouvelles rencontre, la visibilité "public" est très utile, mais elle ne permet pas de modérer cette visibilité à un publique plus restreint tel que les membres de l'instance.
Pourtant le fil "local" est très pratique pour les communautés, et la seule façon de communiquer entre membre est de publier des messages pour toutes les instances...

Avec cette visibilité "local" les membres d'une instance vont pouvoir communiquer entre eux sur le fil "local" sans noyer le fil "global".

De plus la visibilité par défaut est actuellement sur "global" et beaucoup d'utilisateurs ne changent pas ce paramètre est en viennent à partager par défaut leurs messages aux potentiels 200 000 utilisateurs des instances fédérés.
Le changement de ce paramètre par défaut sur "local" permet de modérer la visibilité des utilisateurs tout en nourrissant le fil "local" de leurs instances et en rendant le fil "global" moins dense.


gordonzola:
Translation:

Actually, in order to talk to new people and meet new faces, the “public” visibility is quite useful, but can’t be moderated to reach a more restricted public, like instance members.
However, the “local” thread is very practical for communities, and the only way to communicate between members is to publish messages between instances…

With this “local” visibility, instance members will be able to communicate between them without flooding the “global” thread.

Furthermore, the default visibility setting is actually on “global”, and many users don’t change this setting, and share by default their messages to the potential 200.000 users of the federated instances.
Changing this default value to “local” could allow to moderate the visibility of users while feeding the “local” thread of their instance, thus alleviating the “global” thread.

gordonzola:
Duplicate of #861 by the way

wxcafe:
It'd be nice if that option was automatically disabled in the UI for single-user instances

app/models/status.rb
```diff
@@ -6,7 +6,7 @@ class Status < ApplicationRecord
   include Streamable
   include Cacheable
 
-  enum visibility: [:public, :unlisted, :private, :direct], _suffix: :visibility
+  enum visibility: [:public, :local, :unlisted, :private, :direct], _suffix: :visibility
```
bradurani: I believe adding this enum here will shift the integer of the other values right, for example:

public: 1, unlisted: 2, private: 3, direct: 4

will become

public: 1, local: 2, unlisted: 3, private: 4, direct: 5

This will cause the visibility of all currently existing statuses in the db to change!

Unfortunately, you'll have to put `local` at the end

bradurani (changes requested):
Major data corruption warning above

app/models/status.rb
pixocode: Indeed I had not thought about that ^^ "... Thank you, it is fixed

Aldarone:
@pixocode Can you fix the conflict on ` app/assets/javascripts/components/locales/fr.jsx` ?

Aldarone:
D'ailleurs est-ce que ce ne serait pas plus clair de reprendre la formulation du niveau Unlisted ?

Quelque chose du genre « Do not show on federated timeline » ?

Gargron:
Nothing at the moment is intended to exclude followers from receiving your posts on purpose - following is a simple deal, "I want to see your posts". Adding this to exclude non-local followers would be a big change in that regard, which is why I still haven't merged this. Implementation-wise, this seems fine, but I think the effort is misguided. For example you do not accidentally send your message to 200,000 users on the global network, *unless* you have 200,000 followers, in which case it's intended.

pixocode:
@Gargron "you do not accidentally send your message to 200,000 users on the global network"
Currently, with the default timeline public, I think that's what happens
Moreover, the admin or the people who want to address their instances are obliged to go through the public timeline (not accidentally but in some way forced) :/

lune-sta:
I merged this PR to my instance (v1.2.2), but streaming does not work well.
When posting to local timeline, it is not published in local timeline.
When posting to public timeline, it is published in local timeline.
Reloading the browser will correctly take effect.

Is it not necessary to change the logic to redis?
https://github.com/pixocode/mastodon/blob/master/app/services/fan_out_on_write_service.rb#L24
I'm sorry if it was a misunderstanding.

pixocode:
@lune-sta Thank you, indeed there is little to be a lack in this file.
I tried to work on it but I had not succeeded, I was not sure that there really is need, but that may be the source of the problem. :-/

Aldarone:
@Gargron perhaps I misunderstood the way it works ? I thought it worked exactly like the unlisted option, but the toots are still visible in the local column (but not in the federated column)

Anyway I agree with you, followers should always see every toot except for the « direct » ones.

alex73630:
I just want to be sure about something, if someone follow _user1@instance.one_ from _user2@instance.two_. Do _user2@instance.two_ would see a toot from _user1@instance.one_ with local visibility in his Home TL ? I think this is important to keep this, or add an option to choose if this toot should be send or not to remote followers. (of course if it's possible to do it)

Gargron:
I believe locality should be a separate toggle from privacy setting. Furthermore, here are things to look out for:

* Do not include local toots in Atom feed
* Do not show local toots to logged-out users: public single view (html/atom), public profile view, public hashtag view, public/hashtag timelines API
* Skip PuSH and Salmon distributions for local toots

Gargron:
Closing as incomplete/inactive

colinstu12:
Was this ever going to be added?

PR #1766, opened 2017-04-14
Update nl.jxs

Fjoerfoks:
Some rewording and switching to infinitive localization.


xvilo (approved):
These are good changes language wise.

Gargron (approved):

PR #1767, opened 2017-04-14
CSP Javascript change

xvilo:
Since inline javascript is not 'safe' and blocked with the right CSP headers. I have changed this be loaded as an external file.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 
 class HomeController < ApplicationController
+  protect_from_forgery except: :streaming
   before_action :authenticate_user!
 
   def index
     @body_classes = 'app-body'
+  end
+  
+  def streaming
     @token        = find_or_create_access_token.token
     @web_settings = Web::Setting.find_by(user: current_user)&.data || {}
   end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -7,7 +7,7 @@ class HomeController < ApplicationController
   def index
     @body_classes = 'app-body'
   end
-  
+
   def streaming
     @token        = find_or_create_access_token.token
     @web_settings = Web::Setting.find_by(user: current_user)&.data || {}
--- a/app/views/home/index.html.haml
+++ b/app/views/home/index.html.haml
@@ -1,6 +1,6 @@
 - content_for :header_tags do
 
   = javascript_include_tag '/streaming'
-  = javascript_include_tag 'application'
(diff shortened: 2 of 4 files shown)


Gargron:
This isn't quite this easy - if the constants are not inlined, you have to think about asynchronous loading. Will they be available at the time the React components are instantiated?

ProgVal:
That's the patch for subresource integrity that was applied to https://social.targaryen.house/about ?
In my browser (Firefox 45 on Debian 9), it prevents loading resources with this error in the console:

```
None of the "sha256" hashes in the integrity attribute match the content of the subresource.
```

But @Wonderfall does not seem to be able to reproduce it

Gargron:
This should be resolved by #1806, I think?

Wonderfall:
Yes, #1806 also works for this.

xvilo:
I'm using this on my instance without any issues (mastodon.org.uk). If there is another way to fix the CSP issues, that's fine too!

ProgVal:
Actually, it depends on the Firefox version, for unknown reasons.

Wonderfall:
@ProgVal Found 2 bugs in Firefox resolved.
The first when you told me about HPKP, Firefox had a bug when certificate was using OCSP Must-Staple feature (off topic) : https://bugzilla.mozilla.org/show_bug.cgi?id=1278041
And this about SRI : https://bugzilla.mozilla.org/show_bug.cgi?id=1269241

So the only recommendation is to update Firefox to the latest version...

ProgVal:
> So the only recommendation is to update Firefox to the latest version...

I'm using the ESR version of Firefox; which is also used by half of Firefox users on Debian (according to popcon).
I wouldn't mind upgrading the version of my laptop, but Mastodon would still be broken for these people.

Wonderfall:
Yes, I know... We'll see how it goes when SRI will be available in a stable release. But you're not the only one who complained about that, I'm afraid many others will be disturbed.

Wonderfall:
I understand the bug now. I don't know why, but somehow CSS is compiled with BOM. Then the SHA-256 hash is calculated for SRI. Old versions of Firefox (Chrome never did that) are stripping BOM from CSS files, so indeed the hash changes and SRI goes boom. This behaviour was fixed in Firefox 52.

I'm not a developper (I mean, not as good as you all) so I'm wondering if we can strip BOM before the hash is calculated. If that's possible, we can enable SRI for CSS too.

xvilo:
I think this PR can be closed? There has been added another way to fix the CSP errors

Wonderfall:
But the discussion about SRI is still interesting, though it may be off-topic here.

PR #1770, opened 2017-04-14
[WAIT-FRONT] Add "status_max_length" instance parameter

alkemist:
Add the ability to edit the maximum number of characters per status


pixocode:
#1781

app/views/admin/settings/index.html.haml
```diff
@@ -38,3 +38,7 @@
         %strong= t('admin.settings.registrations.closed_message.title')
         %p= t('admin.settings.registrations.closed_message.desc_html')
       %td= best_in_place @settings['closed_registrations_message'], :value, as: :textarea, url: admin_setting_path(@settings['closed_registrations_message'])
+    %tr
+      %td
+        %strong Status max length
```
ineffyble: Can you change this to be internationalised?

ineffyble (changes requested):

app/views/admin/settings/index.html.haml
pixocode: Fixed :)

pixocode:
If someone is skilled reactjs, help would be greatly appreciated ^^"

ineffyble (approved):

pixocode:
It lacks a "front" part (which I did not succeed to do) so that the parameter is repercussion on the toot form :/

rixx:
So the only part missing is to replace the `500` [here](https://github.com/tootsuite/mastodon/blob/master/app/assets/javascripts/components/features/compose/components/compose_form.jsx#L176)?

pixocode:
@rixx yes, but I don't know how to retrieve the parameter from the back (with reactjs :/)

BrunoSpy:
you can look at 
https://gist.github.com/BrunoSpy/6581f106a15f5d830f3e2289002da2f5
We need a way to get the value of Setting.status_max_length.to_i in the two jsx files mentioned in the previous gist. But I don't know how.

BrunoSpy:
And every locale file must be changed too ("500" is hardcoded there too).

nightpool (changes requested):
needs the above concerns addressed, at least. (front-end changes, locale changes)

but I'm also not sure this is a good feature for the project to have, from a meta perspective. it's pretty easy for admins to change this themselves, to be sure, but I don't know that making this particular setting *easier* to change is a good idea in the long run.

rixx:
Could you expand upon this, @nightpool? Particularly on the "why"?

Gargron:
Closing due to inactivity.

PR #1771, opened 2017-04-14
Fix Japanese translation for remote_follow

tkusano:
Current remote_follow.acct translation inappropriate. Users may not input their own acccount. So fix the wording.


ohac:
Looks good to me. (I'm Japanese)

Gargron (approved):

PR #1772, opened 2017-04-14
Add language detection

Gargron:
- Language detection through WhatLanguage gem
- Stored as status metadata
- Serialized in Atom feeds
- Deserialized from Atom feeds

What else must be done? Expose language attribute in JSON API? Language filters for public timelines?


Gargron:
Hmm, surprised about `PG::NotNullViolation` tbh, I had the impression if a passed value is null and the column is not null, the default value would be used....

app/services/process_feed_service.rb
```diff
@@ -230,6 +231,10 @@ def content(xml = @xml)
       xml.at_xpath('./xmlns:content', xmlns: TagManager::XMLNS).content
     end
 
+    def content_language(xml = @xml)
+      xml.at_xpath('./xmlns:content', xmlns: TagManager::XMLNS)['xml:lang']&.presence || 'en'
```
mjankowski: Should that be hardcoded to `en` as default, or should it use whatever `I18n.default_locale` is?

db/migrate/20170414132105_add_language_to_statuses.rb
```diff
@@ -0,0 +1,5 @@
+class AddLanguageToStatuses < ActiveRecord::Migration[5.0]
+  def change
+    add_column :statuses, :language, :string, null: false, default: 'en'
```
mjankowski: Will this make all historical statuses on an instance retroactively call themselves `en` ?

Gargron: Yep, it will. That's the status quo - all current statuses are in the same pool

mjankowski: Is that true for a server running with a different default locale?

Like, if an admin has bothered to set a different default locale value, it might be a safe assumption that the statuses created on that instance are in that locale, and that this migration should use that value instead of hardcoding `en`?

When this is deployed it will be run on each instance separately, so it might be "more correct" to make the assumption based on that (ie, are the instances running in Japan, Hong Kong, etc - primarily using `en` or other locale?)

mjankowski:
Side question to the actual changes here, is this change being made in future support of something like - https://github.com/tootsuite/mastodon/issues/1094 ?

Gargron:
Yeah, pretty much, although unsure which UI path I'll take. But getting the language attribute out to clients via the API is first step - would allow some to display a "translate" button conditionally, perhaps.

PR #1773, opened 2017-04-14
Add recovery code support for two-factor auth

### LINKED ISSUE #563: Backup codes for 2FA
Many 2FA-enabled apps offer backup codes, which allow you to login, e.g. if you lost your phone or Yubikey and are not able to login via TOTP or any other 2FA mechanism. 

It would be great if Mastodon supported backup codes for 2FA in the next versions :-)

### LINKED ISSUE #987: Locked Out of Two-Factor Authentication
Somehow I seem to have screwed up when setting up Two-Factor Authentication... I don't receive a login token when attempting to sign in, so it seems I'm permanently locked out of my account. Any suggestions? 

* * * *

pfigel:
When users enable two-factor auth, the app now generates ten single-use recovery codes. Users are encouraged to print the codes and store them in a safe place.

The two-factor prompt during login now accepts both OTP codes and recovery codes.

The two-factor settings UI allows users to regenerated lost recovery codes. Users who have set up two-factor auth prior to this feature being added can use it to generate recovery codes for the first time.

Fixes #563 and fixes #987

PS: Recovery codes are 16 characters each by default (from `devise-two-factor`). I haven't found any best-practices on this, but it feels a bit long. I have left it at 16 for now because it's not something users will need to type often, but if the consensus is that less would be okay, I can change the value.


Gargron (changes requested):
Looks fine to me, except you need to make sure that your templates are named `something.html.haml` rather than just `something.haml`

patf:
Thanks for the review! Should be fixed now.

Gargron (approved):

PR #1786, opened 2017-04-14
adds basic microformats tests as suggested by @HongPong in #965

dissolve:


mjankowski:
Could these be view specs on the views/partials in question, rather than controller specs which check the response?

dissolve:
closing this PR and creating a new one based on suggestion here

PR #1791, opened 2017-04-14
Account search service refactor

mjankowski:
Prior to merging this, I'd like feedback as to whether the spec coverage seems adequate here.

I wanted to refactor the service class into smaller methods, but without full coverage beforehand it's hard to be confident that everything is handled well.

I added coverage for what seemed obvious to me, and the search UI still works as before after that changes, though.

### DIFF
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -37,15 +37,15 @@ class AccountSearchService < BaseService
   end
 
   def split_query_string
-    query.gsub(/\A@/, '').split('@')
+    @_split_query_string ||= query.gsub(/\A@/, '').split('@')
   end
 
   def query_username
-    split_query_string.first
+    @_query_username ||= split_query_string.first
   end
 
   def query_domain
-    query_without_split? ? nil : split_query_string.last
+    @_query_domain ||= query_without_split? ? nil : split_query_string.last
   end
 
   def query_without_split?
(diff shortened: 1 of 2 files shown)


app/services/account_search_service.rb
```diff
@@ -1,26 +1,90 @@
...
+      search_results_and_exact_match.compact.uniq
     end
+  end
+
+  def resolving_non_matching_remote_account?
+    resolve && !exact_match && !domain_is_local?
+  end
+
+  def search_results_and_exact_match
+    search_results.to_a + [exact_match]
```
Gargron: The exact match should probably be the first result

app/services/account_search_service.rb
```diff
@@ -1,26 +1,90 @@
...
+  def domain_is_local?
+    @_domain_is_local ||= TagManager.instance.local_domain?(query_domain)
+  end
 
-    results = [exact_match] + results.reject { |a| a.id == exact_match.id } if exact_match
+  def exact_match
+    @_exact_match ||= if domain_is_local?
+      Account.find_local(query_username)
+    else
+      Account.find_remote(query_username, query_domain)
```
Gargron: If you're nullifying query_domain, this method can be simplified to just Account.find_remote, because find_local is just an alias for find_remote with nil domain

Gargron (changes requested):
My mess of a service object has become much more readable! The logic checks out, too, I believe. Just a tiny ordering thing needs change

mjankowski:
Updated to make sure `exact_match` comes back first, and to shortcut right to `find_remote` with `nil` domain.

Gargron (approved):

PR #1796, opened 2017-04-14
Admin user listing

mjankowski:
Adds a user listing to the admin area. Users are filterable by admin status, and confirmed status. User show page has some basic details.

Might be useful for debugging sign up issues.

screen shot 2017-04-14 at 2 36 18 pm
screen shot 2017-04-14 at 2 36 29 pm
screen shot 2017-04-14 at 2 36 51 pm

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -3,6 +3,7 @@
 module Admin
   class UsersController < BaseController
     def index
+      @users = User.all.page(params[:page])
     end
 
     def show
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Admin
+  class UsersController < BaseController
+    def index
+    end
+
+    def show
+    end
+  end
+end
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -3,11 +3,25 @@
 module Admin
   class UsersController < BaseController
     def index
-      @users = User.all.page(params[:page])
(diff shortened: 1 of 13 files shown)


Gargron:
Hmm.. But that's part of accounts already. Could've added more filters to that one.

mjankowski:
I put it in a new/separate spot for a few reasons:

- The accounts filters are already sort of crowded!
- There is theoretically not a 1:1 mapping between users:accounts (I think you could have a user who doesn't have an account but is an admin and just does moderation?)
- The "I didn't get a confirmation email" and "was my email sent yet?" questions seem to come up a bunch, might be nice for admins to be able to see that yes it was sent
- A follow-up to this PR might include "reset password" or "re-send confirmation" workflows from the admin area, which make more sense to me to be on user than on admin

Either way, I'd be happy to rework this into the existing accounts area if you feel strongly. In that case, I'd probably just add the new filters, and then maybe add email address to the listing, and not much else.

Gargron:
Basically, no user record is ever created without an account record. The separation between the two is because I didn't want to store password and e-mail and login-time/IP related fields for remote users who would never have them - so it's a logical separation of the same entity.

Currently the detailed admin account view lists e-mail, last sign in, etc - so I would expect e-mail editing, password resets, and 2FA disables to be there, as well. I do get the point about being crowded, but from an end-user-admin's perspective, having pages for "accounts" and "users" would not be intuitive.

Sorry,  I hope I'm not being too bothersome/demanding about this!

mjankowski:
Makes sense re: separating user/account stuff, especially in light of remote accounts will never have all the stuff associated with signing in or local prefs and so on.

I'll close this one, and might add admin/confirmed filters to admin/accounts later.

I'll keep the slight helper refactor here in a branch and open as other PR.

PR #1797, opened 2017-04-14
Allow configuring SMTP certs by setting SMTP_CA_FILE

jviide:
This is another take on setting the SMTP certificates with the environment variable SMTP_CA_FILE. Functionally the changes are practically the same as in #1486 by @kyzh, except that:
 * the env variable is named SMTP_CA_FILE instead of SMTP_SSL_CERT_FILE
 * there is no default value for the environment variable
 * ~the Dockerfile now sets the value to `/etc/ssl/certs/ca-certificates.crt` by default~

The rationale is that this offers a way to enable certificate verification (SMTP_OPENSSL_VERIFY_MODE=peer) without affecting default behavior. Also based on my (admittedly slapdash) testing it seemed that the following alternatives didn't do the trick:
 * providing a valid CA bundle at `/etc/ssl/cert.pem` (where OpenSSL::X509::DEFAULT_CERT_FILE points at)
 * setting the SSL_CERT_FILE variable to `/etc/ssl/certs/ca-certificates.crt`


Dockerfile
```diff
@@ -4,7 +4,8 @@ LABEL maintainer="https://github.com/tootsuite/mastodon" \
       description="A GNU Social-compatible microblogging server"
 
 ENV RAILS_ENV=production \
-    NODE_ENV=production
+    NODE_ENV=production \
+    SMTP_CA_FILE=/etc/ssl/certs/ca-certificates.crt
```
ineffyble: We're not setting any other variables by default in the `Dockerfile` (the two above are the exception, because they have to be set _before_ dotenv runs).

Can we take this out of here, and instead add it to the example dotenv file for production?

ineffyble (changes requested):

Dockerfile
jviide: Sure!

jviide:
Squashed and rebased on top of the master branch.

jviide:
Updated the description to reflect the changes requested by @ineffyble.

Dockerfile
jviide: Commit f0fff31 implements the requested changes. I hope I didn't mess up the review workflow by rebasing.

akihikodaki:
I am sorry, but I have made a duplicate pull request (#2713) and it has been already merged. The only difference is that my code uses `ENV['SMTP_CA_FILE'].presence` instead of `ENV['SMTP_CA_FILE']`.

Gargron:
Sorry uwu

PR #1799, opened 2017-04-14
Add list of muted user to UI and Getting Started

pfigel:
Add the same UI that already exists for blocked users for muted ones and add it to the "Getting Started" menu.


Gargron (approved):
Thank you! Much needed addition

Issue #1236, opened 2017-04-08
Can't Verify CSRF token authenticity


ericblade:
Hi! I just setup a Docker built server, and created a couple of accounts.  When I try to login with them, I get the "Do you have cookies enabled?" message, and the web container logs "Can't verify CSRF token authenticity".

So, I'm trying to figure out how to get through that problem.

Somewhat more disturbing? perhaps, is that I found a way *around* that problem -- pressing Back from that error screen to go back to the login screen, then pressing Reload on my browser, gets me through.

Any ideas?  What more info can I provide to help?

rhaamo:
Same issue here, but without docker.

rhaamo:
Also duplicate of #1150

ericblade:
Per Technowix on #1150, "You can revert to 624a9a7, temp fix :°" .. which does seem to work.  It does leave me with a "streaming" module that has a LOT of logspam for "Error: Missing access token", but otherwise login and everything else seem to be working.  I'm thinking maybe not close this one as duplicate, though, since it has actual text of an error, and therefore is searchable :-D

ronilaukkarinen:
I see this all the time. What does it mean?

PR #1801, opened 2017-04-14
Introduce Capybara and first feature spec

cpytel:
This introduces Capybara and the first feature spec.

I focused on coverage for log in for the first feature spec because that would
have prevented 624a9a7 causing #1236.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,6 +70,7 @@ group :development, :test do
 end
 
 group :test do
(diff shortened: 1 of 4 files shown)


Gargron (approved):

Gargron:
Thanks a lot for this! Really good stuff

cpytel:
👍

PR #1806, opened 2017-04-14
Send initial state inside a <script type="application/json"> tag

### LINKED ISSUE #1656: Inline JS prevents people from setting up a safer CSP policy
More of an enhancement request than a real issue: I'm pretty sure simply defining `STREAMING_API_BASE_URL` and `INITIAL_STATE` in a different way (not inlining it in a `<script>`) would allow people to setup a CSP policy without an `unsafe-inline` directive.

* * * *

jviide:
This pull request is an attempt to fix #1656. It's based on the suggestion @riking's gave in the issue conversation, but instead of a `<div>` tag the state is sent inside a `<script type="application/json">` tag. This allows the state data chunk to be located inside `<head>`, just before the application code bundle.

The STREAMING_API_BASE_URL value is not the window object's attribute anymore, rather it's routed in the store's meta values similarly like access_token.

The json_escape function *should* take care of escaping problematic characters such as `<`, `>` and `&` in the JSON data.

### DIFF
--- a/app/assets/javascripts/components/containers/mastodon.jsx
+++ b/app/assets/javascripts/components/containers/mastodon.jsx
@@ -60,8 +60,8 @@ import { hydrateStore } from '../actions/store';
 import createStream from '../stream';
 
 const store = configureStore();
-
-store.dispatch(hydrateStore(window.INITIAL_STATE));
+const initialState = JSON.parse(document.getElementById("initial-state").textContent);
+store.dispatch(hydrateStore(initialState));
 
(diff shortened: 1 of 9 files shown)


jviide:
Hm, Code Climate seems to give failing marks. Did these changes make code duplication reach some critical mass?

Komic:
Pretty sure you can safely ignore the Code Climate issues. The `type="application/json"` looks like a good approach to me, thanks for taking a look! :)

Gargron (approved):

PR #1816, opened 2017-04-15
Rename admin/accounts_helper to admin/filter_helper

mjankowski:
- Add some spec coverage for both methods
- Add explicit constant call-outs for where the params are from


Gargron:
As an aside, if you're not already on my Discord, you can e-mail me so I can give you an invite link. (eugen@zeonfederated.com)

Gargron (approved):

PR #1818, opened 2017-04-15
Add some missing Spanish translations

007lva:
Hello @carlosjs23, I see that you contributed for these translations on the past, would be nice if you can check this :smiley:


carlosjs23 (commented):
two_factor_auth string break the page layout, this can be a style issue, I would prefer to keep it abbreviated, perhaps you could improve the abbreviation.

carlosjs23 (approved):

Gargron (approved):

PR #1819, opened 2017-04-15
Remove HTML formatting from portable contacts note.

ghost:
HTML isn't part of the portable contacts spec.

last available reference: https://web-beta.archive.org/web/20150923043617/http://portablecontacts.net:80/draft-spec.html


Gargron:
This change would break functionality. Linked entities in bios have been a requested and expected feature (and as always, formatting is a responsibility of the origin instance). I suggest adding a `type="html"` attribute on the `<poco:note />` tag, it is part of the more general Atom spec: <https://tools.ietf.org/html/rfc4287#section-3.1.1>

rainyday:
Unfortunately that doesn't solve the issue as poco:note isn't allowed to be HTML regardless of whether the encapsulating document allows it or not in other fields. Could you use the `Summary` element instead? You can use HTML there without breaking the spec. 

http://activitystrea.ms/specs/atom/1.0/#atom.object.components
http://activitystrea.ms/specs/atom/1.0/#rfc.section.2.1.7

1337god:
+1

PR #1820, opened 2017-04-15
Give SINGLE_USER a chance to register

### LINKED ISSUE #1817: Brand new instance with SINGLE_USER_MODE=true fails with an exception
Trying to access brand new `SINGLE_USER_MODE=true` instance right after installation causes:

````
[f593578a-1a02-4572-98a1-ff79126279df] method=GET path=/ format=html controller=HomeController action=index status=500 error='ActionController::UrlGenerationError: No route matches {:action=>"show", :controller=>"accounts", :username=>nil} missing required keys: [:username]' duration=99.90 view=0.00 db=20.01
[f593578a-1a02-4572-98a1-ff79126279df]   
[f593578a-1a02-4572-98a1-ff79126279df] ActionController::UrlGenerationError (No route matches {:action=>"show", :controller=>"accounts", :username=>nil} missing required keys: [:username]):
[f593578a-1a02-4572-98a1-ff79126279df]   
[f593578a-1a02-4572-98a1-ff79126279df] app/controllers/home_controller.rb:15:in `authenticate_user!'
[f593578a-1a02-4572-98a1-ff79126279df] app/controllers/concerns/localized.rb:24:in `block in set_locale'
[f593578a-1a02-4572-98a1-ff79126279df] app/controllers/concerns/localized.rb:23:in `set_locale'
````

### PR DESCRIPTION
**Pull request by saper**
An attempt to open a brand new Mastodon instance configured
as SINGLE_USER_MODE=true will cause an exception.

Enable temporary registration if we have no users in the database

Fixes #1817

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < ActionController::Base
   force_ssl if: "Rails.env.production? && ENV['LOCAL_HTTPS'] == 'true'"
 
   include Localized
-  helper_method :current_account
+  helper_method :current_account, :single_user_mode?
 
   rescue_from ActionController::RoutingError, with: :not_found
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
@@ -67,6 +67,10 @@ class ApplicationController < ActionController::Base
       format.any  { head 422 }
       format.html { render 'errors/422', layout: 'error', status: 422 }
     end
+  end
+
+  def single_user_mode?
+    @single_user_mode ||= Rails.configuration.x.single_user_mode && Account.first
   end
 
   def current_account
--- a/app/controllers/auth/registrations_controller.rb
+++ b/app/controllers/auth/registrations_controller.rb
@@ -28,7 +28,7 @@ class Auth::RegistrationsController < Devise::RegistrationsController
   end
 
   def check_enabled_registrations
-    redirect_to root_path if Rails.configuration.x.single_user_mode || !Setting.open_registrations
+    redirect_to root_path if single_user_mode? || !Setting.open_registrations
   end
 
   private
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -13,7 +13,7 @@ class HomeController < ApplicationController
   private
 
   def authenticate_user!
-    redirect_to(Rails.configuration.x.single_user_mode ? account_path(Account.first) : about_path) unless user_signed_in?
+    redirect_to(single_user_mode? ? account_path(Account.first) : about_path) unless user_signed_in?
   end
 
   def find_or_create_access_token
--- a/app/views/accounts/show.html.haml
+++ b/app/views/accounts/show.html.haml
@@ -14,7 +14,7 @@
   %meta{ property: 'og:image:height', content: '120' }/
   %meta{ property: 'twitter:card', content: 'summary' }/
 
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by saper** [comment]
Looks like controller helper_method is a trick that the test suite does not like, will have a look at it

**Comment by mjankowski** [comment]
Instead of doing this check on every request, could we improve the docs to make it more clear that single user instances need to create accounts prior to being able to use the app?

File: app/controllers/application_controller.rb [hunk 1]
```diff
@@ -69,6 +69,10 @@ def unprocessable_entity
     end
   end
 
+  def single_user_mode
```
Gargron: I suggest renaming this to `single_user_mode?`

app/controllers/application_controller.rb
```diff
@@ -69,6 +69,10 @@ def unprocessable_entity
...
+    Rails.configuration.x.single_user_mode && Account.first
```
Gargron: And this could be `@single_user_mode ||= Rails.configuration.x.single_user_mode && Account.first` for multiple calls

Gargron (approved):

Gargron (changes requested):
Didn't mean to click approve, meant to click reqest changes

saper:
I was thinking about spitting out the error message instead.

Is the Account.first check expensive? If yes, then we could store the result somewhere?

Gargron:
@mjankowski The DB check only occurs when single user mode is enabled, and it would occur anyway for the redirect to the right profile, so I think that's not a big deal

app/controllers/application_controller.rb
saper: It was that in my first draft :) thanks

Gargron (approved):

saper:
Wow, thanks.

I've tried to fix the tests by doing

````
diff --git a/spec/views/accounts/show.html.haml_spec.rb b/spec/views/accounts/show.html.haml_spec.rb
index 8265b2f4..1fc75da7 100644
--- a/spec/views/accounts/show.html.haml_spec.rb
+++ b/spec/views/accounts/show.html.haml_spec.rb
@@ -10,6 +10,8 @@ describe "stream_entries/show.html.haml" do
   end
 
   it "has valid author h-card and basic data for a detailed_status" do
+    allow(view).to receive(:single_user_mode?).and_return(false)
+
     alice  =  Fabricate(:account, username: 'alice', display_name: 'Alice')
     bob    =  Fabricate(:account, username: 'bob', display_name: 'Bob')
     status =  Fabricate(:status, account: alice, text: 'Hello World')
@@ -32,6 +34,8 @@ describe "stream_entries/show.html.haml" do
   end
 
   it "has valid h-cites for p-in-reply-to and p-comment" do
+    allow(view).to receive(:single_user_mode?).and_return(false)
+
     alice   =  Fabricate(:account, username: 'alice', display_name: 'Alice')
     bob     =  Fabricate(:account, username: 'bob', display_name: 'Bob')
     carl    =  Fabricate(:account, username: 'carl', display_name: 'Carl')
````
but causes the `<someverylangeclassdescription does not implement: single_user_mode?` which looks like https://github.com/rspec/rspec-rails/issues/1076 to me

Gargron:
@saper Right, I am still encountering the test issue. Really annoying. Any solution?

Gargron:
Never mind, found a hacky way to fix those tests.

saper:
Nice.

What would be the proper way ? Providing spec for application controller?

PR #1823, opened 2017-04-15
Add tasks for open/close registration

chr-1x:
It would be helpful for some server administrators to be able to open and close registrations from the command line. This PR adds two rake tasks, mastodon:settings:open_registrations and mastodon:settings:close_registrations, so that this can be automated easily.

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -86,7 +86,7 @@ namespace :mastodon do
     desc 'Open registrations on this instance'
     task open_registrations: :environment do
       setting = Setting.where(var: 'open_registrations').first
-      setting.value = 'true'
+      setting.value = true
       setting.save
     end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -82,6 +82,22 @@ namespace :mastodon do
     end
   end
 
+  namespace :settings do
+    desc 'Open registrations on this instance'
+    task open_registrations: :environment do
+      setting = Setting.where(var: 'open_registrations').first
+      setting.value = 'true'
```
blackle: why is one true in quotes and the other one without?

Chronister: ah, mistake on my part, appears to have been working anyway because 'true' is truthy. Pushed a fix for consistency.

Gargron (approved):

Aldarone:
Thank you. Very much.

Issue #1059, opened 2017-04-06
Integrate Snowflake IDs for statuses


Gargron:
Snowflake IDs are an idea that is used by Twitter, Discord, Instagram and many more. Here is the problem layout:

- Timelines are ordered by unique IDs, this allows keyset pagination (since_id, max_id)
- Timelines are meant to be chronological
- This is not an issue for local toots, because they get an assigned ID when they are created
- However, federated toots may arrive with various delays
- There are processes that "pull" remote toots into the system as well (thread resolving, the search for toot URL thing)
- In those cases, they also get an ID assigned when they are created *locally*, which can differ wildly from what their `created_at` timestamp is

A snowflake ID is a big integer that consists of a **precise timestamp**, a **sequence mask** and a couple more optional parts to ensure uniqueness, depending on how many different generators run at the same time (i.e. different workers/shards/regions etc)

Because the outermost part of a snowflake ID is a timestamp, chronological sorting is built in. And while the timestamp is most safely-unique when it contains precise milliseconds, I think those can be padded for remote toots (whose timestamp does not carry ms precision)

- [x] All status ID columns in the database schema are bigint
- [ ] A snowflake generator must be integrated for status IDs instead of psql's autoincrement
- [ ] The snowflake generator must accept a timestamp parameter from the past

This is a somewhat complicated area, and I welcome any help & expertise on the issue. Here are a couple of resources for reference:

- http://rob.conery.io/2014/05/29/a-better-id-generator-for-postgresql/
- https://github.com/twitter/snowflake/tree/b3f6a3c6ca8e1b6847baa6ff42bf72201e2c2231

Another note is that Twitter, Instagram &co can write separate high-performance daemons for ID generation. A separate process just for that would ensure that all the different web and sidekiq workers would be calling the same ID generator, thus ensuring that the sequence part in the ID is unique. That would be a simple thing to do, **however** there is a limit to how many different processes I want to impose on small instance admins (there is already: web, sidekiq, streaming)

Another approach is having the ID generator in Ruby code. However, because there is a variable amount of web and sidekiq workers that could call this code in each of their separate processes, they wouldn't share the sequence part. As I mentioned earlier, this could be mitigated by a "worker index" part in the snowflake ID. **However**, the problem is that the various web and sidekiq workers do not know information like "I am worker number 2 out of 5".

[...]

PR #1824, opened 2017-04-15
Favourite toot importing and exporting

expenses:
Toot exporting seemed like a handy feature for account migration.

An image of what the export looks like

The export returns the domain, author and id of each toot, like:

    localhost:3000, admin, 8
    localhost:3000, admin, 9
    localhost:3000, admin, 10

I think this is a good format for favourites because it doesn't depend on whether the domain being migrated to is local or not, and it retains the privacy of the authors toots.

I also added a spec file for checking.


expenses:
I've now added importing as well. This PR will need to be checked carefully because I haven't yet tested importing statuses from other instances.

app/models/export.rb
```diff
@@ -20,6 +20,16 @@ def to_following_accounts_csv
     to_csv account.following
   end
 
+  def to_favourite_posts_csv
+    CSV.generate do |csv|
+      account.favourites.each do |favourite|
+        domain = favourite.status.account.local? ? Rails.configuration.x.local_domain : favourite.status.account.domain
+
+        csv << [domain, favourite.status.account.username, favourite.status_id]
```
Gargron: This will export a local status ID, which is irrelevant to any other instance but the one where it comes from. It would make sense to just export a list of URIs here. Though mind you, it's impossible to do a look up by URI if you don't already have the status in the DB. So list of URLs would be the only thing that worked.

app/workers/import_worker.rb
```diff
@@ -71,4 +73,18 @@ def process_follows
       end
     end
   end
+
+  def process_favourites
+    import_rows.each do |row|
+      begin
+        domain, account, id = row
+
+        status = (domain == Rails.configuration.x.local_domain) ? Status.find(id) : FetchRemoteStatusService.new.call("http://#{domain}/users/#{account}/#{id}.atom")
```
Gargron: Yeah this doesn't work with any status that doesn't come from a Mastodon instance. Exporting URLs would make this code simpler.

Gargron: Keep in mind the reason we don't have backfilling of old statuses en masse is because we don't have #1059, so if you ran an import like this on 10k statuses from e.g. ~3 months ago all public timelines would be filled with 10k statuses from 3 months ago.

Gargron (changes requested):

app/workers/import_worker.rb
expenses: Oh, ok. So is importing impractical at the moment? Maybe I should revert to c7bf4794d7149df3d29f4cad98cc2d4c42fb346b then.

Gargron:
Closing due to inactivity. Without import this could've still been a useful feature, just a CSV of two columns, URI and URL, but alas.

PR #1826, opened 2017-04-15
Add a classname to the loading bar

Aldarone:
This will allow styling of the loading bar using Compass and thus, ease custom instance styling.


walf443:
It looks good to me

Gargron (approved):

PR #1838, opened 2017-04-15
Improve toot click box (#1673) (partial)

### LINKED ISSUE #1673: Allow the expansion of a toot by on the whole block instead of only the text
On a toot displayed in the timeline, it seems only the text is clickable to expand it (to see replies and such). From another social network tool I use (hint: it ends with Deck...), I am used to be able to click everywhere on the block, except the handle and avatar, to expand the message. 
I think it would be nice to have the same behaviour so the following red zones are clikcable:

Screenshot from Mastodon

I find it easier/quicker to click in the left zone (the reddest one on the screen) than aiming in the text, which can have links and images like in this example.

* * * *

expenses:
This PR mostly fixes #1673 by expanding the click-box of the toot to the whole thing while still allowing the buttons to be clicked without 'selecting' the toot by stopping the event propagation. 

Unfortunately the toot `DropdownTrigger` (`...`) class cannot be edited since it is comes from [an external toolkit](https://github.com/Fauntleroy/react-simple-dropdown), meaning that clicks on it cannot be stopped from propagating (as far as I know at least) and will 'select' the toot.

As far as I'm aware we have three options to sort this out:

1. Get a commit pushed to the `react-simple-dropdown` to add this `onClick` functionality.
2. Make a fork of the repository and do the above.
3. Implement dropdowns internally so we're not relying on the toolkit.


expenses:
Closing due to https://github.com/tootsuite/mastodon/issues/1673#issuecomment-294633851.

PR #1845, opened 2017-04-15
fix regex filter

abcang:
Fixed a bug that regex filter matches HTML tag generated by simple_format.

* completion of line breaks
* tag deletion
* unescape

### DIFF
--- a/app/assets/javascripts/components/features/ui/containers/status_list_container.jsx
+++ b/app/assets/javascripts/components/features/ui/containers/status_list_container.jsx
@@ -26,7 +26,7 @@ const makeGetStatusIds = () => createSelector([
     try {
       if (showStatus) {
         const regex = new RegExp(columnSettings.getIn(['regex', 'body']).trim(), 'i');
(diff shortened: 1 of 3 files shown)


Gargron:
Have you noticed if doing the DOM parsing in a loop makes the columns noticeably laggier?

abcang:
I confirmed that it will be delayed.
I would like to think of another way.
Thank you.

Gargron:
@abcang Perhaps something like an extra reducer where you would store the indexable plaintext at time of entry, and then that would become part of the status object in the container and could be filtered-on.

Don't worry if this sounds too complicated, I might do it later...

abcang:
Yes, I think so too.
Instead of deleting tags and escaping on the client side, I'd like to change the server side to send raw content.

abcang:
I changed to send raw content on the server side.
But, in the case of remote toot, I noticed that escaped content was sent.
Therefore, I am planning to return to the process to escape on the client side.

abcang:
I changed it to unescape in reducer.
Therefore, the lag problem was solved.

Gargron (approved):

Issue #1853, opened 2017-04-15
Part of the settings menu hidden


thomaskuntzz:
When the translation of a navigation item in the settings takes two lines, the "Authorized Apps" link is cut and hidden in part.

In this case, it's the 2FA link that takes two lines in the French translation instead of one in the English version:

![capture du 2017-04-15 16-17-39](https://cloud.githubusercontent.com/assets/16361470/25064286/ba34b16c-21f7-11e7-9221-a625cc91cbc9.png)

![capture du 2017-04-15 16-19-57](https://cloud.githubusercontent.com/assets/16361470/25064294/d7fc68ac-21f7-11e7-9470-94b47f7bdb0d.png)

I think the navigation links should not be stuck at the bottom of the screen as they are now, or **at least the content of the navigation menu has not to be hidden if its height exceeds the on it has in the English version.**

Also, maybe we could put less white space around the Mastodon logo at the top?

SansPseudoFix:
Regarding the right side, I have a PR ( #1633 ) waiting that fix it.

wxcafe:
Duplicate of #781, closing

PR #1347, opened 2017-04-09
enable to scroll the setting's page sidebar

857b:
add scroll:auto and use margin instead of padding for the bottom

PR #1855, opened 2017-04-15
Make settings nav scrollable if it exceeds viewport height

henrycatalinismith:
This is a fix for #1853. I just added `overflow-y: auto` to the preferences sidebar.

nav

It's possible that this scroll bar might be a little ugly on Windows, I dunno. Haven't got a Windows machine to test on. But it addresses the issue in the lowest-impact way possible.

An alternative fix for this would be to remove the `position: fixed` on `body.admin`. That'd be a slightly bigger change to the behaviour of the settings pages, since it'd mean that the whole nav sidebar scrolled alongside the main settings content, but maybe it's also better than having two independently scrollable divs alongside each other? Dunno... your call, and if you'd prefer fixing it that way I'd be happy to do that work too 😄


thomas-kuntz (approved):

thomaskuntzz:
Can someone merge this?

rhaamo:
There is also #1347 which seems to fix the same issue.

wxcafe:
This is a duplicate of #1347, closing

PR #1857, opened 2017-04-15
ignoring .DS_Store

mbilokonsky:
Is there a reason this isn't already ignored?


mbilokonsky:
oh weird nevermind it is

PR #1858, opened 2017-04-15
cache blockings for reducing queries.

walf443:


walf443:
Test failed. It seems that this change is not affected. Does master failing test now?

Gargron (approved):

PR #1864, opened 2017-04-15
ActiveRecord::NotFound -> ActiveRecord::RecordNotFound

alpaca-tc:
Fixed bug.

```
uninitialized constant ActiveRecord::NotFound
```

---

BTW, the `Rails.application.routes.recognize_path` is private API so we really shouldn't use it in application. The only reason it's not been removed already is there's a massive amount of legacy controller routing tests of rails that depend on it.


Gargron (approved):

Gargron:
It may be a private API but it's insanely useful and I don't know of an alternative

PR #1865, opened 2017-04-15
Domain block improvements

Gargron:
* Make file attachment on MediaAttachment optional
* Create MediaAttachment but without actual file download when domain is blocked with reject_media set to true
* Clean up old media files when creating a new domain block with reject_media set to true
* Return remote_url in media attachments API if local file is not present
* Undo domain block action in admin UI
* Ability to enable reject_media from admin UI

### DIFF
--- a/app/services/block_domain_service.rb
+++ b/app/services/block_domain_service.rb
@@ -3,12 +3,34 @@
 class BlockDomainService < BaseService
   def call(domain_block)
     if domain_block.silence?
-      Account.where(domain: domain_block.domain).update_all(silenced: true)
+      silence_accounts!(domain_block.domain)
+      clear_media!(domain_block.domain) if domain_block.reject_media?
     else
-      Account.where(domain: domain_block.domain).find_each do |account|
-        account.subscription(api_subscription_url(account.id)).unsubscribe if account.subscribed?
-        SuspendAccountService.new.call(account)
-      end
+      suspend_accounts!(domain_block.domain)
+    end
+  end
+
+  private
+
+  def silence_accounts!(domain)
+    Account.where(domain: domain).update_all(silenced: true)
+  end
+
+  def clear_media!(domain)
+    Account.where(domain: domain).find_each do |account|
+      account.avatar.destroy
+      account.header.destroy
+    end
+
+    MediaAttachment.where(account: Account.where(domain: domain)).find_each do |attachment|
+      attachment.file.destroy
+    end
+  end
+
+  def suspend_accounts!(domain)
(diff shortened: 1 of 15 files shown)


ashfurrow:
This is great, thank you!

mjankowski:
Domain block stuff looks good.

Is the view spec stuff (fix?) here on purpose?

Gargron:
@mjankowski I kinda had to add it here since I wanted the tests to pass. I didn't notice that the tests were still broken when merging the other thing, sadly - I just looked through the code and it looked good to me.

mjankowski:
Cool.

I just submitted this for view spec fix only: https://github.com/tootsuite/mastodon/pull/1888

If that looks good, maybe merge that one first and then rebase this with just the domain stuff?

app/services/block_domain_service.rb
```diff
@@ -3,12 +3,34 @@
 class BlockDomainService < BaseService
   def call(domain_block)
     if domain_block.silence?
-      Account.where(domain: domain_block.domain).update_all(silenced: true)
```
mjankowski: I like the change to move this to smaller methods.  Possible further refactor...

- Set `@domain_block` accessor right in call, and then call a method like `process_block` or something, and do nothing else in `call`
- `process_block` would have the if/else logic around silence vs suspend
- Add a method like `blocked_domain` which is `domain_block.domain` and replace a few of those
- Add something like `domain_accounts` which does the Account.where(...)` that's repeated a few times here

config/locales/en.yml
```diff
@@ -90,8 +92,22 @@ en:
           silence: Silence
           suspend: Suspend
         title: New domain block
+      reject_media: Reject media files
+      reject_media_hint: Removes locally stored media files and refuses to download any in the future. Irrelevant for suspensions
```
mjankowski: Period at end of sentence?

mjankowski (approved):
Left a note about a possible refactor, but looks good overall.

Angristan:
It would be great to have the "Reject media files" option without blocking or suspending the instance.

Angristan:
Another idea : proxying media files without caching them. Otherwise it will be a mess with CSP.

Gargron:
@Angristan How to deal with WebM/MP4? Proxying on the fly would be one heck of a load.

Angristan:
True... About that, we should at least have the ability of choosing the delay of caching. It's a huge waste of space, in a few feeks/month, every instance will have dozens of GBs of medias stored that aren't even accessed anymore. :/

PR #1866, opened 2017-04-15
Fix up recovery codes design a little

Gargron:


app/views/settings/two_factor_auths/_recovery_codes.html.haml
```diff
@@ -1,6 +1,5 @@
 %p.hint= t('two_factor_auth.recovery_instructions')
 
-%h3= t('two_factor_auth.recovery_codes')
```
mjankowski: Is this key used anywhere else?

mjankowski (approved):

app/views/settings/two_factor_auths/_recovery_codes.html.haml
Gargron: Don't think so, it was added in the pr that added this functionality

PR #1876, opened 2017-04-15
Fix #1870 - Strip control characters out of strings in AtomSerializer

### LINKED ISSUE #1870: Ox can not dump control character
`AtomSerializer` depends on [Ox](https://github.com/ohler55/ox).
Currently some user have saved control character. However `AtomSerializer` can not dump it.

```
CONTROL_CHARACTER = ["\v", "\f", "\b"]

CONTROL_CHARACTER.each do |string|
  Ox.dump(Account.new(note: string)) 
end
# "\v" => SyntaxError: '\#x08' is not a valid XML character.
# "\f" => SyntaxError: '\#x0c' is not a valid XML character.
# "\b" => SyntaxError: '\#x0b' is not a valid XML character.
```

Should we remove control character from string attribute?
I don’t know what to do about that. 😕 

```
*A* `SyntaxError` *occurred while* `GET </users/XXXXXXXXX.atom>` *was processed by* `accounts#show`
Exception
----------------
'\#x08' is not a valid XML character.

XXX/app/lib/atom_serializer.rb:10:in `dump'
XXX/app/lib/atom_serializer.rb:10:in `render'
XXX/app/controllers/accounts_controller.rb:20:in `block (2 levels) in show'
```

* * * *

Gargron:


blackle:
https://www.w3.org/TR/2000/REC-xml-20001006#charsets I found this spec for valid xml characters, so I suppose we should be excluding all characters _but_ these ones

nightpool:
shouldn't Ox just turn these into XML entities? why is it refusing to dump these?

app/lib/atom_serializer.rb
```diff
@@ -311,11 +311,17 @@ def unfavourite_salmon(favourite)
...
-    attributes.each { |k, v| element[k] = v.to_s }
-    element << content.to_s unless content.nil?
+    attributes.each { |k, v| element[k] = sanitize_str(v) }
+    element << sanitize_str(content) unless content.nil?
     parent  << element
   end
 
+  def sanitize_str(raw_str)
+    str = raw_str.to_s
+    ["\v", "\f", "\b"].each { |char| str = str.delete(char) }
```
alpaca-tc: `raw_str.to_s.tr!("\v\f\b", '')` 👍

Issue #1319, opened 2017-04-09
Clarify "private" status privacy text


duckinator:
When you adjust the privacy level for an individual toot, the description for "Private" is "Post to followers only." However, this is misleading because it _doesn't federate to other instances,_ so it's only posting to followers _on the same instance_.

![screenshot_2017-04-09_02-39-27](https://cloud.githubusercontent.com/assets/39698/24835355/2faea3aa-1cce-11e7-9f3e-deb4041aeb70.png)

The text should be updated to indicate this. Perhaps something like "Post only to followers on this instance." or something like that? (Or maybe something that's shorter but equivalent, it's 3am and I'm sleepy.)

(I've [read about why private toots don't federate](https://mastodon.social/@Skirmisher/2044254), and this issue is _not_ about making them do so &mdash; that's a totally separate issue.)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

wolfteeth:
Seconding this. The existing text and the existing documentation are very confusing.

msappir:
Thirded. A short way to amend the text would be "Post to local followers only" (local already denotes 'this instance' in the UI for "local timeline" so this should be comprehensible.)

yiskah:
something for the sake of UX is we want each of the descriptions to be the same height and to not be so long that it could go off the screen or cause the need to scroll to read the full menu. If you could mess around and inspect element and find some phrasings which have the same height between then we'd love to use those.

msappir:
My suggested minimal edit seems to do the trick without adding a line break.
![170411-privacy_text1](https://cloud.githubusercontent.com/assets/7591121/24913928/61bd745c-1edb-11e7-8753-7909612d8b9d.png)

msappir:
My minimal edit seems to do the trick without adding a line break:
![170411-privacy_text1](https://cloud.githubusercontent.com/assets/7591121/24913928/61bd745c-1edb-11e7-8753-7909612d8b9d.png)

yiskah:
Cool, someone could make a pull for this then. Hopefully at some point down the road we'll find a solution for federation using NodeInfo or a whitelist or somesuch

Cassolotl:
![screen shot 2017-09-26 at 00 51 29](https://user-images.githubusercontent.com/6607512/30836252-ddda2cc2-a254-11e7-911f-dbd0ac25d302.png)

I believe this issue can be closed now!

PR #1878, opened 2017-04-15
#1319 Specify local followers rather than all followers.

simonv3:
Addresses #1319, makes it clearer that Private means local followers only, rather than all followers across instances


yiskah:
It seems this may not be true soon if #2111 is merged

Gargron:
Indeed, that was a temporary limitation

simonv3:
That works too!

Gargron:
Current master federated private toots so I am closing this 👍

PR #1883, opened 2017-04-15
Onboarding modal

Gargron:
screenshot from 2017-04-16 01-53-17
screenshot from 2017-04-16 01-53-29
screenshot from 2017-04-16 01-53-39
screenshot from 2017-04-16 01-53-51
screenshot from 2017-04-16 01-54-00


adbelle:
I think it could do with an extra page explaining what's on page 1. It's not quite eighth-grade legible IMO that you're going to be talking to some people on -other- websites that your instance admin doesn't control.

ashfurrow:
Looks awesome! I wonder about this view in the modal:



Maybe the descriptions for local/federated timelines should be split and under their icons, to match the descriptions of "Home" and "Notifications"?

Gargron:
@ashfurrow Nope, that's how it was before this and it looked worse. Besides, local/federated are kind of the same "public timeline" just filtered differently.

blackle:
for page 4, I think you could go without the word "subtype" and just say "the local timeline is the same but limited to posts by users on <instance name>"

hnrysmth:
@adbelle this is a really excellent point. The words "decentralized federation" are quite a hardcore pair of tech jargon words to ask people to read as part of their very first post-signup interaction. Makes me wonder if there's a risk of alienating people who don't know what a "decentralized federation" is. Not so sure about adding more pages though. IMO the more you add to something like this the less people will read it.

I think Google Photos has a decent example of one of these: as few pages as possible, as little text as possible, and focuses on telling you _what_ the app can do rather than exactly _how_ to do it.

google

Slack's another good example. No mention of "the cloud" or any heavy-hitting words like "team-based collaboration".

welcome-pages-slack

I've been trying to find ways of simplifying the text for this modal but copywriting is _haaaaaaaaaaard_ and even though I have a lot of respect for it as a skillset & career, that's not the same as being actually _good_ at it, so I'm struggling! 😂 Here's the best I could do so far. It got longer cos I couldn't find a way to say something so complex in simpler words without also using _more_ words.

screen shot 2017-04-16 at 14 52 16

Gargron:
I updated the wording a little. This won't be perfect anyway, until dopatwo (my artist friend) is available (27th April).

yiskah (commented):
It's bad.

PR #1885, opened 2017-04-16
Remove unused gem: coffee-rails

007lva:
Because I don't see Coffeescript code:

screenshot from 2017-04-16 02 27 16


007lva:
Ups, I see the build is broken but not due to my changes. It also fails locally on master branch, ping @Gargron.

mjankowski:
We're working on those view spec failures -- this looks good, and will get merged after those are fixed.

Gargron (approved):

PR #1886, opened 2017-04-16
Leave out the "Expires" header from S3 uploads

jviide:
This pull request leaves out the "Expires" header from S3 uploads to improve compatibility with Google Cloud Storage.

Currently setting `S3_ENDPOINT=https://storage.googleapis.com` and uploading a file yields the following error message:

> The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.

Based on some trial and error it appears that Google Cloud Storage is having hard time with the "Expires" header, set in `config/initializers/paperclip.rb`. Leaving the header out seems to be enough to convince Cloud Storage to cooperate.


Gargron:
What's important to know is whether removing the Expires header will make browsers not cache the images. That could turn into unnecessary bandwidth costs for S3 users. Perhaps Cache-Control alone is sufficient but I am not sure.

jviide:
Based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires:
 > If there is a Cache-Control header with the "max-age" or "s-max-age" directive in the response, the Expires header is ignored.

Also https://tools.ietf.org/html/rfc7234#section-5.3 states:
> If a response includes a Cache-Control field with the max-age directive (Section 5.2.2.8), a recipient MUST ignore the Expires field.

I have no idea how well different browsers actually align with that requirement, though :neutral_face:

Gargron (approved):

PR #1887, opened 2017-04-16
Unite all mandatory rake tasks in mastodon:daily

Gargron:
Add mastodon:media:remove_remote task (this one would make more sense after #1865)
Make mastodon:maintenance:add_static_avatars more resilient to exceptions

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -44,7 +44,7 @@ namespace :mastodon do
       MediaAttachment.where(account: Account.silenced).find_each(&:destroy)
     end
 
-    desc 'Remote cached remote media attachments that are older than a week'
+    desc 'Remove cached remote media attachments that are older than a week'
     task remove_remote: :environment do
       MediaAttachment.where.not(remote_url: '').where('created_at < ?', 1.week.ago).find_each do |media|
         media.file.destroy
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -32,6 +43,13 @@ namespace :mastodon do
     task remove_silenced: :environment do
       MediaAttachment.where(account: Account.silenced).find_each(&:destroy)
     end
+
+    desc 'Remote cached remote media attachments that are older than a week'
```
mjankowski: Should be "Remove cached..." (instead of "Remote"?)

mjankowski (approved):

lib/tasks/mastodon.rake
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 namespace :mastodon do
+  desc 'Execute daily tasks'
+  task :daily do
```
ineffyble: Is it worth adding some type of output/logging, to iterate over the tasks?

ineffyble: e.g. Right now, running `rake mastodon:daily` returns no relevant output. It would be nicer if it had something like:

```
Clearing feeds...
Refreshing PuSH subscriptions...
```

If it was easy enough to do, could you perhaps put the list of tasks into an array, and iterate over it, printing the task `desc` before each one?

ineffyble (approved):
Looks good, doesn't error for me on an empty instance.

Issue #1898, opened 2017-04-16
Relaxing Ruby version requirement


akihikodaki:
Ruby version requirement has been updated in #1159. It is nice to have the latest version, but I suggest to relax the requirement because of the following concerns:

* Ruby 2.3.x is still common for some Linux distributions.

The Software Collections for Red Hat and CentOS and Ubuntu 16.04 LTS still provides Ruby 2.3.x. Though [Ruby 2.4 brings various changes](https://docs.ruby-lang.org/en/trunk/NEWS-2_4_0.html), Mastodon still works on Ruby 2.3.x for now.

* `Gemfile` specifies the `TEENY` version.

That means Ruby and Mastodon should be updated at the same time; you cannot update before a newer version of Mastodon gets released, and visa versa, even if the change is expected to be compatible.

I would like to have `ruby '>= 2.3.0', '< 2.5.0'` in `Gemfile` to solve them, but I would also like to hear your opinions.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

expenses:
why not just `ruby '>= 2.3.0'`? Also feel free to make the PR and move the discussion there.

yumetodo:
@expenses ``Ruby 2.4 brings various changes`` is the cause , I supposed.

akihikodaki:
No, I put the one because Ruby 2.5.0 and the later versions, which are not released yet, may bring breaking changes, considering the versioning scheme.

expenses:
@akihikodaki Oh, ok, fair enough.

wxcafe:
since #1901 has been merged, I'm closing this

PR #1901, opened 2017-04-16
Relax Ruby version requirement

akihikodaki:
This solves #1898.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
   postgresql: 9.4
 
 rvm:
+  - 2.3.4
   - 2.4.1
 
 services:
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-ruby '2.4.1'
+ruby '>= 2.3.0', '< 2.5.0'
 
 gem 'pkg-config'
 


ineffyble:
Will be interesting to see how the Travis builds go against multiple Ruby versions. Would also be neat to have some stats on what versions are in use out there (and what other technologies) - something like [stats.civicrm.org](http://stats.civicrm.org).

If we're going to be testing/supporting multiple versions, perhaps we need them better defined in the Travis file? The current build is running against `2.3.1`, but I believe `2.3.3` is the latest minor version of 2.3 (in fact, I can see a complaint in the Travis logs:
```warning: parser/current is loading parser/ruby23, which recognizes 2.3.3-compliant syntax, but you are running 2.3.1.****```

**Comment by akihikodaki** [comment]
This change includes update for `.travis.yml` as well. 

https://github.com/tootsuite/mastodon/pull/1901/files#diff-354f30a63fb0907d4ad57269548329e3R16
```diff
 rvm:
-  - 2.4.1
+  - 2.3
+  - 2.3.0
+  - 2.4
```

Travis says 2.3 (lastest 2.3) and 2.3.0 are OK, but 2.4 is failing :/ I'm investigating the issue.

ineffyble:
Yes, I'm not certain they're adequate though. I think we want to test against 2.3.3 which your Travis changes don't implement.

akihikodaki:
Looks like rvm in Travis CI does not implement 2.3.4 and 2.4.x. I'll try to use [the precompiled binaries](http://rubies.travis-ci.org/).

akihikodaki:
80037d9e448429dd5511c9d26143a6a8c1d78674 enumerates all versions supposed to be specified with `Gemfile`

.travis.yml
```diff
@@ -14,6 +14,12 @@ addons:
   postgresql: 9.4
 
 rvm:
+  - 2.3.0
```
mjankowski: It's probably not necessary to test every single patch release version via travis ... that will be a lot of builds for every PR/branch.

Specifying either just a) the minor release versions we want to support, or b) the point releases within the minor release we want to support - is probably fine.

Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-ruby '2.4.1'
+ruby '>= 2.3.0', '< 2.5.0'
```
mjankowski: This is probably fine, because we can continue to use `.ruby-version` to mean "the version we expect developers to be working against", but relaxing in `Gemfile` will allow admins who can't run very latest and don't have access to rvm/rbenv/etc to use whatever they have in some supported range.

.travis.yml
Gargron: Indeed, this will make every Travis build go a looooong time,. which means longer delay before I know that a PR can be safely merged. I suggest just 2.3.1 and 2.4.1

Gargron (changes requested):

ineffyble:
Not 2.3.3 instead? (Latest minor version of 2.3)

007lva:
Latest minor version of 2.3 is [2.3.4](https://www.ruby-lang.org/en/news/2017/03/30/ruby-2-3-4-released/).

Gargron (approved):

PR #1907, opened 2017-04-16
reduce unneed query when post without attachements.

walf443:
This commit reduce following query:

```
 MediaAttachment Load (0.9ms)  SELECT "media_attachments".* FROM "media_attachments" WHERE "media_attachments"."status
_id" IS NULL AND 1=0 ORDER BY id asc
```

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -39,7 +39,8 @@ class PostStatusService < BaseService
 
     raise Mastodon::ValidationError, I18n.t('media_attachments.validations.too_many') if media_ids.size > 4
 
-    media = MediaAttachment.where(status_id: nil).where(id: media_ids.take(4).map(&:to_i))
+    target_media_ids = media_ids.take(4).map(&:to_i)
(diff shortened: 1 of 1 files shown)


app/services/post_status_service.rb
```diff
@@ -39,15 +39,16 @@ def validate_media!(media_ids)
 
     raise Mastodon::ValidationError, I18n.t('media_attachments.validations.too_many') if media_ids.size > 4
 
-    media = MediaAttachment.where(status_id: nil).where(id: media_ids.take(4).map(&:to_i))
+    target_media_ids = media_ids.take(4).map(&:to_i)
+    media = target_media_ids.blank? ? [] : MediaAttachment.where(status_id: nil).where(id: target_media_ids)
```
ineffyble: Would it be possible to just set media to nil here, instead of an empty array, and avoid the change lower down?

walf443: @ineffyble fixed

Gargron (approved):

PR #1919, opened 2017-04-16
Remove API authentication for public statuses

### LINKED ISSUE #1760: Remove authentication requirement for /api/v1/statuses/ endpoint when passed ID is of a public toot
See title. A user wanting to programatically get a public toot from any instance currently must scrape the public /users/ page. Allowing use of the API endpoint for such activity would be cleaner.

* * * *

happycoloredbanana:
Fixes #1760

### DIFF
--- a/app/controllers/api/v1/statuses_controller.rb
+++ b/app/controllers/api/v1/statuses_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Api::V1::StatusesController < ApiController
-  before_action -> { doorkeeper_authorize! :read }, except: [:create, :destroy, :reblog, :unreblog, :favourite, :unfavourite]
+  before_action :authorize_if_got_token, except:            [:create, :destroy, :reblog, :unreblog, :favourite, :unfavourite]
   before_action -> { doorkeeper_authorize! :write }, only:  [:create, :destroy, :reblog, :unreblog, :favourite, :unfavourite]
   before_action :require_user!, except: [:show, :context, :card, :reblogged_by, :favourited_by]
   before_action :set_status, only:      [:show, :context, :card, :reblogged_by, :favourited_by]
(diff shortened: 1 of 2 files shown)


ineffyble (commented):
Lots of unrelated lines showing as changed - indentation/lint perhaps? Code Climate isn't showing any new *or* fixed errors, so I'm a little puzzled.

happycoloredbanana:
I have created two contexts for tests: one for tests with a proper OAuth token, another - with a missing token. All the previously existing tests were moved into the first context - so indents.

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -114,4 +114,9 @@ def status_params
   def pagination_params(core_params)
     params.permit(:limit).merge(core_params)
   end
+
+  def authorize_if_hidden
+    status = Status.find(params[:id])
+    doorkeeper_authorize! :read if status.hidden?
```
Gargron: If I'm not mistaken, current behaviour is to pretend to 404 when trying to view a status you don't have a permission to view - because we don't want to reveal the *existence* of such a status. Now this breaks that behaviour - if you can't view a status without credentials, but with credentials it throws 404, you know that it exists.

Instead, if this method is to be public, it should always be public.

Gargron (changes requested):

happycoloredbanana:
Fixed with one additional check added: if an OAuth token is present in the request - perform Doorkeeper authorization to provide consistent behavior for bad tokens. Does it make sense?

happycoloredbanana:
Ok, now I see that the hidden check is not needed, hold on.

happycoloredbanana:
So, my question is - what should happen if someone tries to call this API while providing an invalid token? Should we say "doesn't matter, the API is public anyway" or "if you are trying to behave according to the OAuth flow, we will provide the needed checks as well"?

ineffyble:
I feel like the latter - we should reject invalid tokens even if the token
wasn't needed.

Gargron:
Agreed with @ineffyble, that's the correct way to do it

happycoloredbanana:
Cool, so I think my last commit does exactly this - if the token is present in the request, perform Doorkeeper authorization as before. Otherwise, no checks are done. What do you think?

Gargron (approved):

PR #1925, opened 2017-04-16
Adds note for instance admins.

ashfurrow:
Based on my conversation with @wxcafe.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -3,3 +3,4 @@
 * * * *
 
 - [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
+- [ ] I am an instance admin who has checked that I'm running a [tagged release](https://github.com/tootsuite/mastodon/releases) and _not_ the latest master.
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -3,3 +3,4 @@
 * * * *
 
 - [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
+- [ ] I am an instance admin who has checked that I'm running a [tagged release](https://github.com/tootsuite/mastodon/releases) and _not_ the latest master.
```
wxcafe: we also want users to be able to contribute tho... I'd word it more like 

[ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this)

wxcafe (changes requested):

ISSUE_TEMPLATE.md
Gargron: :+1: on @wxcafe's suggestion, we don't want to necessarily discourage people from reporting issues on master, but we do want to know if it *is* on master, because issues in tagged releases are more dire

wxcafe (approved):

Gargron (approved):

nightpool:
My concern about this is that it makes it feel mandatory—should we not report bugs that only happen on master?

wxcafe:
you can just leave the box unchecked. At least that way we know what you're reporting, and bugs that are on a tagged release need a quicker fix

PR #1929, opened 2017-04-16
separate queue for RemovalWorker

walf443:
When some user that has many followers remove a status, many RemovalWorker job generated.
It make slow consuming default  queue tasks.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: bundle exec puma -C config/puma.rb
-worker: bundle exec sidekiq -q default -q push -q pull -q mailers
+worker: bundle exec sidekiq -q default -q push -q pull -q mailers -q remove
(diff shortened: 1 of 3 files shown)


ineffyble:
Are you able to make a pull request against the [documentation repo](https://github.com/tootsuite/documentation) and update all the examples there to include the extra queue?

This is a bit of a breaking change, so not sure how we could roll it out easily.

Gargron:
Yeah, no, this is not an easy change to do at all! When I had to introduce the `pull` queue I was getting reports of admins who didn't realize it, for almost a week. Besides, adding a new queue by itself won't change much unless you're running independent Sidekiq processes for specific queues.

Other solutions:

- Make `RemoveStatusService` run faster
- Move it to an existing non-default queue

walf443:
OK. I'll research RemoveStatusService to work more fast.

PR #1930, opened 2017-04-16
Puma worker killer

peterkeen:
This will help relieve memory pressure on smaller instances.


ineffyble:
I actually looked into this back when I was trying to get Heroku working well. It seems like a good idea to me, and this implementation looks okay, but I haven't tested. @Gargron?

peterkeen:
Not that it means a whole lot in all the diverse environments that Mastodon runs in, but I've run this exact config in a large production app (~30k monthly active users) running on Heroku and it's helped out immensely.

Gargron:
So uh, do you know where the memory leak is? Is it a general Rails/Puma issue? I can't say I noticed RAM usage growing over time on mastodon.social. Are you sure it's not just peaking after the entire application allocates all of its stuff over time?

ineffyble:
It could be Heroku-specific. There was definitely a slow but steady memory growth when I was first trialling it with Heroku.

peterkeen:
I don't have a specific leak, unfortunately. That said, every Rails + Puma app I've ever worked on has slow memory growth that eventually starts hitting quotas or swap. Deploying really often (I'm assuming mastodon.social runs master or close to it most of the time) will alleviate the symptoms since it doesn't have time to grow, but an instance running on a smaller VM that only deploys point releases will likely start having problems.

That said, I absolutely understand not wanting to include something like this "just because.

nateberkopec:
All Ruby applications have logarithmic memory usage, which will almost never completely flatten out due to memory fragmentation. Growing memory usage doesn't necessarily mean a leak.

Gargron:
I guess I want more opinions on whether this should be included in the box. Not sure whether this would be a performance improvement or the opposite (with new workers having to re-allocate memory)

Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
 gem 'jquery-rails'
 gem 'puma'
+gem 'puma_worker_killer'
```
walf443: `require: false` ?

config/puma.rb
```diff
@@ -12,3 +12,10 @@
 end
 
 plugin :tmp_restart
+
+before_fork do
+  require 'puma_worker_killer'
+
+  rolling_restart_interval = (ENV['PUMA_WORKER_RESTART_INTERVAL'] || (6 * 3600)).to_i
```
walf443: It's may be better if PUMA_WORKER_RESTART_INTERVAL is nil, disable puma worker killer.

walf443:
It may prevent  growing up memory usage without restarting servers for long time.
In my experience, it's helpful for some server administrators.

I'm agree to introduce puma-worker-killer.
but config/puma.rb  should be more adjustable.
(it maybe hard to just configure ENV settings.

notbakaneko:
I don't think this should be included as default, and instead should implemented in the fork being used if desired. It's too easy for misconfigured worker killers to have a negative performance impact or mask the actual issue to be resolved.

If the service is going to being run for the long term, it would be better to figure out where the memory is going and deal with it appropriately according to your circumstances.

ashfurrow:
Thanks for the pull request @peterkeen. It sounds like this is a controversial change – it would probably need to be opt-in, and documented. Would you be available to make those changes, and update your branch to resolve merge conflicts?

peterkeen:
I'll just close it. If people want it they can easily add it themselves to their deploy.

PR #1936, opened 2017-04-16
Fix #1670 - Update OStatus2 gem

### LINKED ISSUE #1670: 500 on /api/v1/search
I manually installed on an arm64 (aarch64) machine at bd834add567fc0a0f3960385a7f9abe4324ff212. 

I get a HTTP 500 response when I use the search, the logs says
``` NoMethodError: undefined method `n=' for #<OpenSSL::PKey::RSA:0x00007f900e0910>
Did you mean?  n```

Plus a stacktrace
```
app/services/follow_remote_account_service.rb:42:in `call'
app/services/account_search_service.rb:21:in `call'
app/services/search_service.rb:15:in `call'
app/controllers/api/v1/search_controller.rb:7:in `index'
app/controllers/concerns/localized.rb:24:in `block in set_locale'
app/controllers/concerns/localized.rb:23:in `set_locale'
```

I'm not a Ruby person, but how I see it what is actually crashing is https://github.com/tootsuite/ostatus2/blob/master/lib/ostatus2/magic_key.rb#L8

Now I'm wondering what I broke and why a `OpenSSL::PKey::RSA` should have a `n` field/method at all, because I don't see anything about that in the documentation. 

Any idea what is broken?

* * * *

Gargron:


ineffyble:
Updating branch to re-run Codeclimate, since it errored.

mjankowski (approved):

PR #1937, opened 2017-04-16
Updates update instructions.

ashfurrow:
Sort of goes with https://github.com/tootsuite/mastodon/pull/1925 . Basically it makes our instructions to admins running on Docker explicitly instruct them to checkout a tag version. I used `git merge RELEASE_TAG` because, if they've modified the code we don't want them to lose those modifications (which is critical for Docker users, since the default `docker-compose.yml` specifies non-persistent DB volumes). I'm happy to amend based on feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,11 +132,12 @@ Running any of these tasks via docker-compose would look like this:
 
 This approach makes updating to the latest version a real breeze.
 
-1. `git pull` to download updates from the repository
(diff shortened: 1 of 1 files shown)


mjankowski (changes requested):
Want to move this to docs repo? https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Production-guide.md

ashfurrow:
Yeah, good idea.

PR #1938, opened 2017-04-16
Fix #1813 - Alleviate extra requests when processing mentions

### LINKED ISSUE #1813: Mastodon Causing Load Spikes on GNU Social Instances
Summary:
GNU Social admins have been experiencing heavy load originating from Mastodon instances.

Description:
For the past week, there has been a massive influx of new users on the platform and some have decided to host new instances of Mastodon. The problem with the federation of users (and possibly followbots) is that there have been cases where load averages on GS instances like GNU/Smug, FreezePeach and Sealion have noticeably jumped since the attention Mastodon received from technology news sites and InfoSec professionals, and has forced some GS admins to redirect spammy requests elsewhere. I believe it's down to Mastodon constantly asking for timeline data and I know there have been efforts to mitigate the problem such as making use of a cache like Varnish. 

STR:
None, although admins of affected instances are free to offer reproducible examples.

Actual Results:
GNU Social instances have been bombarded with enough requests to go down (unintentional DDoS).

Expected Results:
Mastodon and GNU Social should be able to play nicely without one negatively affecting the other.

* * * *

- [x] I have searched/browsed the repo's other issues to ensure this is not a duplicate.

Gargron:
The `<link rel="mentioned" />` tag refers to accounts by href. So we were
matching the DB by the `url` attribute, and falling back to HTTP look-up.
However, GS and Mastodon use profile URLs as URIs, too, and the match
for that was missing. This could potentially alleviate some extra network
requests

### DIFF
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -161,13 +161,7 @@ class ProcessFeedService < BaseService
       xml.xpath('./xmlns:link[@rel="mentioned"]', xmlns: TagManager::XMLNS).each do |link|
         next if [TagManager::TYPES[:group], TagManager::TYPES[:collection]].include? link['ostatus:object-type']
 
-        url = Addressable::URI.parse(link['href'])
-
-        mentioned_account = if TagManager.instance.web_domain?(url.host)
-                              Account.find_local(url.path.gsub('/users/', ''))
-                            else
-                              Account.find_by(url: link['href']) || FetchRemoteAccountService.new.call(link['href'])
-                            end
+        mentioned_account = account_from_href(link['href'])
 
         next if mentioned_account.nil? || processed_account_ids.include?(mentioned_account.id)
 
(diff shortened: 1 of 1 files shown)


app/services/process_feed_service.rb
```diff
@@ -178,6 +172,16 @@ def mentions_from_xml(parent, xml)
       end
     end
 
+    def account_from_href(href)
+      url = Addressable::URI.parse(href)
+
+      if TagManager.instance.web_domain?(url.host)
+        Account.find_local(url.path.gsub('/users/', ''))
```
ineffyble: Is this one of those cases where Rubocop would tell us to use .delete instead of .gsub?

ineffyble (approved):
Makes sense to me

app/services/process_feed_service.rb
Gargron: It didn't highlight this. Besides, I just moved this code from one place to another, didn't introduce a new issue.

PR #1943, opened 2017-04-16
Disable CodeClimate duplication checks

Gargron:


ineffyble (approved):
Yes, yes, for the love of god, yes.

PR #1948, opened 2017-04-16
Use RSYNC for synced folder in vagrant on windows

Naouak:
On Windows, symbolic link are not working with nfs mount in VirtualBox. Symbolic links are required for browserify.

Using RSYNC on windows permit the creation of symbolic links.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -107,7 +107,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.hostsupdater.remove_on_suspend = false
   end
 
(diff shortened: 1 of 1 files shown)


beatrix-bitrot:
is still necessary/beneficial after the move to webpack?

Naouak:
I don't know, I didn't try to setup a vagrant box since the move to webpack.

wxcafe:
I'm pretty sure this should be closed now? Last activity apart from @ineffyble mentionning it in #4961 is from over 3 months ago and as mentionned we switched to webpack even then

PR #1962, opened 2017-04-16
Disable submit button when no text.

### LINKED ISSUE #1961: 422 when trying to submit an empty status
To reproduce:
* Don't enter any status text
* Click `Toot` (or whatever)

Expected behavior:
* The button should probably just be disabled if there's no text

* * * *

zautumnz:
Fixes #1961

### DIFF
--- a/app/assets/javascripts/components/actions/compose.jsx
+++ b/app/assets/javascripts/components/actions/compose.jsx
@@ -73,9 +73,13 @@ export function mentionCompose(account, router) {
 
 export function submitCompose() {
   return function (dispatch, getState) {
+    const status = emojione.shortnameToUnicode(getState().getIn(['compose', 'text'], ''));
+    if (!status || !status.length) {
+      return;
+    }
     dispatch(submitComposeRequest());
     api(getState).post('/api/v1/statuses', {
-      status: emojione.shortnameToUnicode(getState().getIn(['compose', 'text'], '')),
+      status,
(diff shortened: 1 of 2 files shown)


Gargron:
Actually, the reason this isn't already the case is that we want to greet users with a blue, ready-to-go "toot" button, rather than a grayed out one.

I suggest that perhaps you could achieve a similar effect just by returning early from the submitCompose action if the text is empty

zacanger:
Ah, understood. Changed to return early from `submitCompose` if no text, and dispatch a `success` so the button and textarea aren't disabled.

app/assets/javascripts/components/actions/compose.jsx
```diff
@@ -74,8 +74,13 @@ export function mentionCompose(account, router) {
 export function submitCompose() {
   return function (dispatch, getState) {
     dispatch(submitComposeRequest());
+    const status = emojione.shortnameToUnicode(getState().getIn(['compose', 'text'], ''));
+    if (!status || !status.length) {
```
Gargron: It can be simpler. Just do this before the dispatch of the request above, then you don't have to dispatch an extra success

Gargron (changes requested):

app/assets/javascripts/components/actions/compose.jsx
zacanger: Fix, thanks.

Gargron (approved):

PR #1965, opened 2017-04-16
Modify Japanese about following and followers

metafalz:
This Japanese translation of following and followers was opposite. I would like to fix it.


metaphallus:
Sorry, I misunderstood and close it.

PR #1970, opened 2017-04-17
Improve streaming API server performance with cluster

geta6:
I added the code from line 19 to line 33 of `index.js`. (other changes are indentation)

- It automatically restarts when node dies
- By using multiple clusters, CPU resources can be used efficiently

In order to make the default behavior the same as before, the number of clusters is set to `1` as the default value.

(This is a summary of the commit code used in pawoo.net)


Gargron (approved):
Very interesting!

I have been load-balancing traffic between 4 different streaming API servers on 4 different machines on mastodon.social. Your PR allows the same effect to be achieved on a single machine without much effort. Very nice!

PR #1974, opened 2017-04-17
Add message box explaining about NSFW

geta6:
(This PR is a proposal)

I added a message box explaining about **NSFW**.
This is mainly for cultural areas where the word **NSFW** is not pervasive. (for example, Japan)

As for the sentences to explain, I am not so familiar with other than Japanese, so I want to leave it to someone...

7223200b2d362e9d35e965b2704de0ad

(This is a summary of the commit code used in pawoo.net)


app/assets/javascripts/components/features/compose/containers/sensitive_guide_container.jsx
```diff
@@ -0,0 +1,40 @@
+import { connect } from 'react-redux';
+import TextIconButton from '../components/text_icon_button';
+import { changeComposeSensitivity } from '../../../actions/compose';
+import { Motion, spring } from 'react-motion';
+import { injectIntl, defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  message: { id: 'compose_form.sensitive_message', defaultMessage: 'Please be sure to mark NSFW if the image you are trying to post is erotic content' }
```
ashfurrow: I think it might make sense to broaden the language here, maybe something like:

> Please be sure to mark the post as NSFW if the image is not safe for work environments.

(note the period at the end too.)

I think being vague might help instance admins, but maybe this is text they'd like to customize.

yiskah: "Not safe for work" is a phrase that has come to be a bit of an idiom. When it comes to cross-cultural explanation it may not be clear what that means. A phrasing such as "If the image contains content others may deem inappropriate." or "If an image contains content others may wish not to see"

geta6:
Fixed conflicts and translation.

@ashfurrow @yiskah Thank you for message translation and explanation!

ashfurrow:
Neat, looks sharp!

Gargron:
If this is something many people want, perhaps it should be above the compose form like other "warnings"? I've recently added a `<WarningContainer />` that is responsible for showing the correct warning when needed.

I'm not sure if people want this, though. It sounds like something you'd only want to see once..?

Cassolotl:
Now that the button is not called NSFW (it's an eye symbol that changes to a crossed/slashed eye when clicked, so show that images will *not* be visible), is this unnecessary? There is no text, so it's down to cultural differences and instance rules/codes of conduct as to whether images are hidden, right?

PR #1986, opened 2017-04-17
Allow connecting to Redis via Unix socket

### LINKED ISSUE #1325: Allow access to Redis via Unix Socket
On some shared hosting platforms (e.g. [Uberspace](https://uberspace.de)), Redis is only available via a Unix socket, not a TCP port.

It would be awesome if the Mastodon config in `.env.production` supported specifying a Unix socket for Redis as an alternative to the current TCP access via host/port.

* * * *

yeah:
Allows admins to specify `REDIS_SOCKET` in their environment as an alternative to `REDIS_HOST` and `REDIS_PORT`, in order to connect to Redis via Unix sockets. Connections via TCP continue to work as well, obviously.

This is a prerequisite for running Mastodon on some shared hosting providers such as Uberspace which many users have requested.

This would fix #1325.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -2,6 +2,7 @@
 REDIS_HOST=redis
 REDIS_PORT=6379
 # REDIS_DB=0
+# REDIS_SOCKET=/tmp/redis.sock
 DB_HOST=db
 DB_USER=postgres
 DB_NAME=postgres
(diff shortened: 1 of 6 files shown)


jcalvert:
Just a general observation that the redis configuration here could use some DRYing out.

yeah:
I agree. I wanted to keep this commit focused on the issue at hand though. A DRY-up before/after this commit would surely be good.

PR #1989, opened 2017-04-17
Fix #1897 - Return reblogged: false on unreblog (was wrongly named va…

### LINKED ISSUE #1897: Unreblog still returns "reblogged": true
I noticed this while testing my client in development today, and confirmed the behaviour on mastodon.social.

Reproduction is pretty straightforward: boost a post that isn't boosted, the icon will show that it has been boosted. Tap to unboost. If you're watching the network flow, you'll see the response from `/api/vi/statuses/:id/unreblog` will contain `"reblogged": true`. The icon remains as if boosted.

Tap it again and you'll get a HTTP 404 error.

* * * *

Gargron:
…riable)

### DIFF
--- a/app/controllers/api/v1/statuses_controller.rb
+++ b/app/controllers/api/v1/statuses_controller.rb
@@ -77,9 +77,9 @@ class Api::V1::StatusesController < ApiController
   end
 
   def unreblog
-    reblog         = Status.where(account_id: current_user.account, reblog_of_id: params[:id]).first!
-    @status        = reblog.reblog
-    @reblogged_map = { @status.id => false }
+    reblog       = Status.where(account_id: current_user.account, reblog_of_id: params[:id]).first!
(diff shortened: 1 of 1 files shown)


mjankowski (approved):

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -77,9 +77,9 @@ def reblog
   end
 
   def unreblog
-    reblog         = Status.where(account_id: current_user.account, reblog_of_id: params[:id]).first!
-    @status        = reblog.reblog
-    @reblogged_map = { @status.id => false }
+    reblog       = Status.where(account_id: current_user.account, reblog_of_id: params[:id]).first!
```
krainboltgreene: You can also do `Status.find_by!(account: current_user.account, reblog_of_id: params[:id])`

PR #2015, opened 2017-04-17
Store current locale for sidekiq

alpaca-tc:
Get stored locale from redis.

related: https://github.com/mperham/sidekiq/blob/7385564bcd2fbded5199b4f549fe924fd816c384/lib/sidekiq/middleware/i18n.rb

```
before
> Sidekiq.client_middleware
<Sidekiq::Middleware::Chain:0x007fa010576ba8 @entries=[
  #<Sidekiq::Middleware::Entry:0x007fa010558ec8 @klass=SidekiqUniqueJobs::Client::Middleware, @args=[]>
]>

after
> Sidekiq.client_middleware
<Sidekiq::Middleware::Chain:0x007fa010576ba8 @entries=[
  #<Sidekiq::Middleware::Entry:0x007fa010558ec8 @klass=SidekiqUniqueJobs::Client::Middleware, @args=[]>,
  #<Sidekiq::Middleware::Entry:0x007fa01114b528 @klass=Sidekiq::Middleware::I18n::Client, @args=[]>
]>
```

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,3 +1,5 @@
+require 'sidekiq/middleware/i18n'
+
 host = ENV.fetch('REDIS_HOST') { 'localhost' }
 port = ENV.fetch('REDIS_PORT') { 6379 }
 password = ENV.fetch('REDIS_PASSWORD') { false }


Gargron:
This isn't needed at all, because mailers use their own with_locale blocks and there isn't anything else in the background that uses i18n, unless I'm missing something?

alpaca-tc:
Ooops that's right, I misunderstood 👍 
Thank you for reviewing pr 😄

Issue #2012, opened 2017-04-17
OStatus URIs and Mastodon handles


ThibG:
Hi,
It seems to me that Mastodon handles (@user@domain) are mostly undefined.
Indeed, to the best of my knowledge, they are based on an optional feature of OStatus: finding remote feeds and API endpoints using webfinger. However, this mechanism is both optional and unidirectional: as far as I know, even if the feeds and the API endpoints are discoverable through a webfinger query for acct:user@domain, they do not have to refer to “user” or “domain” in any way.

Worse, OStatus not being based on those handles, interactions typically never mention them even if they exist, and mention authors' URIs (which are not restricted to a specific format by OStatus) instead.

From my understanding of the code, Mastodon handles this situation in a few different ways (sometimes differently depending on the part of the code used):
- If given author information from an atom element (e.g.: salmon API request), it builds a @user@domain handle by using the author's name field as “user” and assume the author's URI to be an URL, using its domain part for “domain”
- If it has to process an URI matching an already-known account (thus with a known handle), it uses that
- Otherwise, if the URI is not known, it assumes it is an URL to a web page with a pointer to an atom feed, from which it uses the author's email field if present, or the author's name as the “user” and the domain part of the *feed*'s URL.

This seems ill-defined to me, especially since there seem to be three (well, one of them is my fault) different ways to build a handle for an unknown user (author's email ; author's name + URI host part ; author's name + feed URL part), and at least two to find known users (URI, or author's name + URI host part) depending on the part of the code.

ThibG:
To me, it makes sense to:
1. Require remote users to be discoverable via webfinger (this is already the case as far as I can tell)
2. Require the feeds' author's “email” field to map back to the webfinger-discoverable handle (this does not break federation with existing Mastodon instances)
  This may break federation with non-Mastodon OStatus servers, but I have no data on this.
3. Unknown account discovery should be opportunistic and not impose other requirements.
  (In particular, we should not require the URI to be an URL, nor require that URL domain to match the
  account's domain)

Given an URI, we should directly return matching users (this is already the case *for most parts of the code*).

For unknown account discovery, the goal is to find a handle (that then has to be queried via webfinger) given an URI or an “author” element from an atom element. It should behave predictably and reliably discover existing Mastodon instances, even when WEB_DOMAIN and LOCAL_DOMAIN are different.
Following the previous requirements, I propose the following:
- If we have an “author” element with an “email” element, use it as the handle to discover
[...]

PR #2017, opened 2017-04-17
Fix username of reply message when using `WEB_DOMAIN` env.

3846masa:
On v1.2, Mastodon detects account's local-domain from `./xmlns:author/xmlns:uri`.
e.g. )
- https://github.com/tootsuite/mastodon/blob/v1.2/app/services/process_interaction_service.rb#L16-L17
- https://github.com/tootsuite/mastodon/blob/v1.2/app/services/process_feed_service.rb#L271-L277

`./xmlns:author/xmlns:uri` will be created from `RoutingHelper` methods.
c.f.)
- https://github.com/tootsuite/mastodon/blob/v1.2/app/lib/atom_serializer.rb#L19-L23
- https://github.com/tootsuite/mastodon/blob/v1.2/app/lib/tag_manager.rb#L83-L84

Because of that, Mastodon use web-domain as local-domain.
So, username of reply message is mistaken.
(user@local.example.com (correct) -> user@web.example.com (wrong))

We should replace domain name in `./xmlns:author/xmlns:uri` when using `WEB_DOMAIN` env.

-----

NOTE:

Mastodon may use `./xmlns:author/xmlns:email` for detecting local-domain (e.g. when fetching user stream atom). It works correctly on v1.2.
e.g.)
- https://github.com/tootsuite/mastodon/blob/v1.2/app/services/fetch_remote_account_service.rb#L22-L29

### DIFF
--- a/app/lib/atom_serializer.rb
+++ b/app/lib/atom_serializer.rb
@@ -16,7 +16,7 @@ class AtomSerializer
   def author(account)
     author = Ox::Element.new('author')
 
-    uri = TagManager.instance.uri_for(account).gsub(Rails.configuration.x.web_domain, Rails.configuration.x.local_domain)
+    uri = TagManager.instance.uri_for(account)
 
     append_element(author, 'id', uri)
     append_element(author, 'activity:object-type', TagManager::TYPES[:person])
--- a/app/lib/atom_serializer.rb
+++ b/app/lib/atom_serializer.rb
@@ -16,7 +16,7 @@ class AtomSerializer
   def author(account)
     author = Ox::Element.new('author')
 
-    uri = TagManager.instance.uri_for(account)
+    uri = TagManager.instance.uri_for(account).gsub(Rails.configuration.x.web_domain, Rails.configuration.x.local_domain)
 
     append_element(author, 'id', uri)
     append_element(author, 'activity:object-type', TagManager::TYPES[:person])
--- a/app/lib/tag_manager.rb
+++ b/app/lib/tag_manager.rb
@@ -81,7 +81,7 @@ class TagManager
 
     case target.object_type
     when :person
-      account_url(target)
+      account_url(target).gsub(Rails.configuration.x.web_domain, Rails.configuration.x.local_domain)
     when :note, :comment, :activity
       unique_tag(target.created_at, target.id, 'Status')
     else


app/lib/atom_serializer.rb
```diff
@@ -16,7 +16,7 @@ def render(element)
   def author(account)
     author = Ox::Element.new('author')
 
-    uri = TagManager.instance.uri_for(account)
+    uri = TagManager.instance.uri_for(account).gsub(Rails.configuration.x.web_domain, Rails.configuration.x.local_domain)
```
nightpool: this seems like a very bad place to fix this. Why not change uri_for?

Gargron: Indeed, this is the wrong place

Gargron (changes requested):

Gargron:
>Mastodon may use ./xmlns:author/xmlns:email for detecting local-domain (e.g. when fetching user stream atom). It works correctly on v1.2.

I think Mastodon is the only software which uses the `<email />` tag for this, sadly.

app/lib/atom_serializer.rb
3846masa: I agree.
However, I don't know where does mastodon use `'TagManager.instance.uri_for`.
So, I can't guess where will changing `TagManager.instance.uri_for` affect.

Should I move that to `uri_for`?
If `TagManager.instance.uri_for` will be changed, I insert `gsub` in [here](https://github.com/tootsuite/mastodon/blob/v1.2/app/lib/tag_manager.rb#L83-L84).

3846masa:
I agree with @nightpool and @Gargron .
I move that to `uri_for`. Please check it.

ThibG:
Yes, this is what I was debugging yesterday. Unfortunately, changing the URI in those requests does not seem like a correct solution to me, as the users' URI *is* the web part (this is so that mentioned links of unknown users will still go to their web profile — and because that's how Mastodon currently discovers unknown remote mentioned accounts).

More generally, I think we should establish clear rules for mapping account handles (user@domain) with URIs, and use the same rules everywhere instead of doing slightly different things in different places (e.g. salmon requests processing, mentions processing, and fetching a remote user's profile), see #2012 .

PR #2052, opened 2017-04-18
Add meta tags for OGP to following & follow page

ykzts:
### Screenshot (on `/users/ykzts/followers`)

Screenshot on http://localhost:5000/users/ykzts/followers

### DIFF
--- a/app/views/accounts/_og.html.haml
+++ b/app/views/accounts/_og.html.haml
@@ -1,5 +1,4 @@
 %meta{ property: 'og:site_name', content: site_title }/
-%meta{ property: 'og:type', content: defined?(type) ? type : 'website' }/
 %meta{ property: 'og:title', content: [yield(:page_title).strip.presence, site_title].compact.join(' - ') }/
 %meta{ property: 'og:description', content: account.note }/
 %meta{ property: 'og:image', content: full_asset_url(account.avatar.url(:original)) }/
--- a/app/views/accounts/_og.html.haml
+++ b/app/views/accounts/_og.html.haml
@@ -1,0 +1,8 @@
+%meta{ property: 'og:site_name', content: site_title }/
+%meta{ property: 'og:type', content: defined?(type) ? type : 'website' }/
+%meta{ property: 'og:title', content: [yield(:page_title).strip.presence, site_title].compact.join(' - ') }/
+%meta{ property: 'og:description', content: account.note }/
+%meta{ property: 'og:image', content: full_asset_url(account.avatar.url(:original)) }/
+%meta{ property: 'og:image:width', content: '120' }/
+%meta{ property: 'og:image:height', content: '120' }/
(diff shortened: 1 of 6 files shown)


usagi-f:
There are also OGP haml files in the following dir.

```
app/views/stream_entries/_og_description.html.haml
app/views/stream_entries/_og_image.html.haml
```

Why don’t you put them in the same dir ?

```
// example

app/views/
  └ogp/
    ├_og_description.html.haml
    ├_og_image.html.haml
    └_og_account.html.haml
```

ykzts:
@usagi-f `app/views/accounts/_og.html.haml` is only called from accounts view. I do not think I need it.

usagi-f:
@ykzts 
okay, Whatever you want to do.

app/views/accounts/_og.html.haml
```diff
@@ -0,0 +1,8 @@
+%meta{ property: 'og:site_name', content: site_title }/
+%meta{ property: 'og:type', content: defined?(type) ? type : 'website' }/
```
mjankowski: That `type` seems sort of magical here ... where is that coming from?

app/views/accounts/followers.html.haml
```diff
@@ -1,6 +1,9 @@
 - content_for :page_title do
   = t('accounts.people_who_follow', name: display_name(@account))
 
+- content_for :header_tags do
+  = render partial: 'og', locals: { account: @account }
```
mjankowski: Can you use the more compact style here? -- `render 'og', account: @account`

app/views/accounts/_og.html.haml
mjankowski: Ah, I see that it gets set on on `accounts/show` but not the other two.

I might just leave this one line (type) out of the partial, and then include it in accounts/show as "profile" and leave it out of the other two views, since "website" is the default value for those.

ykzts: @mjankowski Yes! That is understanding.

app/views/accounts/followers.html.haml
ykzts: @mjankowski I changing it! Looks good.

ykzts:
@mjankowski Thank you for review! I fixed it.

app/views/accounts/_og.html.haml
mjankowski: It looks like you added this back into the show view, but did not remove it from here...?

ykzts: Oops... sorry. Please see 05b1b5c.

mjankowski (approved):
This looks good now, and I see you've noticed the moved files referenced here: https://github.com/tootsuite/mastodon/pull/2153

Whichever PR is merged first, the other will likely have to update.

ykzts:
@mjankowski revert accounts/{following,followers} in 85e459f. Maybe not conflict.

PR #2053, opened 2017-04-18
Adding JSON format for /about/more

woganmay:
Adding a JSON representation for the Extended Information view, including only the information that's public by default.


app/views/about/more.json.rabl
```diff
@@ -0,0 +1,10 @@
+node(:uri)                  { site_hostname }
+node(:title)                { Setting.site_title }
```
mjankowski: I think there's a `site_title` method available for this, and instance presented has lots of the others

mjankowski (changes requested):

app/views/about/more.json.rabl
woganmay: I was just going from a default elsewhere. Happy to change this if there's a faster way to do it.

woganmay: `/app/views/api/v1/instances/show.rabl` uses Setting. for the 3 main bits of info too, which what I was referencing here - I thought that was the right way to do it.

mjankowski: They are ultimately both calling the same things ... the request was mostly to match what we do in the html version of about/more.

As a side note here, I assumed you are doing this for api consumption...? If that's the case, than we might want to either a) add a `link` describing where to find this to the html version of about/more, b) think about whether `about/more` is actually the best url for this endpoint (as opposed to something in the official api), c) maybe this should just expand the existing instance api endpoint?

woganmay: I was actually thinking the same, with regards to (c). Since the api is all about returning machine-readable data, it would make sense to keep all the machine-readable bits in one place there.

I'll submit a PR for that quickly, and reference this one.

woganmay:
Ok @mjankowski I think I finally managed to pacify the machine.

This PR now just adds the instance_presenter to the instances controller, and returns the extended information for instances#show

app/views/api/v1/instances/show.rabl
```diff
@@ -1,6 +1,12 @@
 object false
 
-node(:uri)         { site_hostname }
-node(:title)       { Setting.site_title }
-node(:description) { Setting.site_description }
-node(:email)       { Setting.site_contact_email }
+node(:uri)                  { site_hostname }
+node(:title)                { Setting.site_title }
```
mjankowski: There's a `site_title` helper we could use here.

app/views/api/v1/instances/show.rabl
```diff
@@ -1,6 +1,12 @@
...
+node(:description)          { Setting.site_description }
```
mjankowski: For this one and the next few lines, I believe InstancePresenter knows about all of these values, and any that it doesn't could be added.

app/controllers/api/v1/instances_controller.rb
```diff
@@ -1,7 +1,12 @@
 # frozen_string_literal: true
 
 class Api::V1::InstancesController < ApiController
+  before_action :set_instance_presenter, only: [:show]
   respond_to :json
 
   def show; end
+
+  def set_instance_presenter
+    @instance_presenter = InstancePresenter.new
```
mjankowski: I might just call this `@instance` here, so that from the show view perspective we have a whole bunch of `{ @instance.whatever }` lines, which reads nicely from a "show me the instance in this view" perspective.

ykzts:
resolved with #4074.

PR #2061, opened 2017-04-18
Make S3_ENABLED default to false for Pull Request apps on Heroku

zunda:
This change makes it easier to handle Review Apps while requiring
an edit on S3_ENABLED when the app is being deployed through
Heroku Button.

When a Review App is created on Heroku, config vars are taken
from app.json. With S3_ENABLED defaulting to true, the Review app
can not be built because other S3 config vars unset would make
asset compliation fail.

### DIFF
--- a/app.json
+++ b/app.json
@@ -37,7 +37,7 @@
     },
     "S3_ENABLED": {
       "description": "Should Mastodon use Amazon S3 for storage? This is highly recommended, as Heroku does not have persistent file storage (files will be lost).",
-      "value": "true",
+      "value": "false",
       "required": false
     },
     "S3_BUCKET": {


ashfurrow:
Thanks for the pull request! From what I understand, Heroku deploys require an S3 bucket to work (because a Dyno's disk is not persisted). Going to request a review from someone who knows more about this.

Gargron:
I also do not understand this. Heroku does not work without S3, as it doesn't have a persistent file system.

zunda:
Thanks for looking into this. This PR is basically to make it possible to create a [Review App](https://devcenter.heroku.com/articles/github-integration-review-apps) which also makes use of `app.json`.

A Review App based upon the current `app.json` fails to build when pre-compiling assets because S3 is enabled and no credentials are given. During creation of a Review App, there is no chance to change config vars and credentials for S3 are left blank.

With this PR merged, Review App will be able to be created. On the other hand, an app created through Heroku Button would need `S3_ENABLED` edited, in addition to the credentials and other details of the S3 bucket. In my opinion, needing to edit one more config var for Heroku Button app is not too much of a burden but I would appreciate for your thoughts!

ashfurrow:
It sounds like we have two options:

1. **Don't merge**. Effectively prioritizes Heroku users by requiring Review App admins to use S3.
1. **Merge**. Effectively prioritizes Review App by letting users there not use S3 by default, while possibly losing user data on Heroku (since the data being "persisted" would be deleted on the next daily dyno reboot).

Does that sound like it covers all the pros/cons? Is there another option? Can Review App use a substitute `app.json` file?

nightpool:
@zunda where are you trying to create the Review App? Can you just push the app.json changes to your own fork? Is this supposed to be part of the Mastodon PR process? If so, is there interest in this from the developers, and who would be paying for it?

zunda:
tl;dr - I am retracting this PR. 

@nightpool - I wrote this PR when I had a trouble in creating a PR from *my own repository* where I should have been able to also overwrite `app.json`.

@ashfurrow - another reason why I wrote this PR was a friend of mine having a trouble in creating a Heroku app through the Heroku Button. With this PR merged, an app would have been created but lacking support for S3 would have hit the friend of mine after 24 hours with a loss of media files. It would be better to fail fast and let the user of Heroku Button to either disable S3 or provide required information.

So, my conclusion here is that for Review App, the developer should overwrite the `app.json` file. For Heroku Button app, s/he should look into the config vars more carefully. Now I'm convinced that this PR should NOT be merged. Sorry for the noise!

ashfurrow:
Cool, thanks for following up @zunda!

PR #2062, opened 2017-04-18
Translate onboarding dialog to japanese

walf443:


app/assets/javascripts/components/locales/ja.jsx
```diff
@@ -76,6 +76,24 @@ const ja = {
   "notifications.column_settings.show": "カラムに表示",
   "notifications.column_settings.sound": "通知音を再生",
   "notifications.settings": "カラム設定",
+  "onboarding.next": "次へ",
+  "onboarding.page_one.welcome": "Mastodonへようこそ！",
+  "onboarding.page_one.federation": "Mastodonは誰でも参加できるSNSです。",
+  "onboarding.page_one.handle": "あなたは今数あるMastodonインスタンスの1つである{domain}にいます。あなたのフルハンドルは{handle}です。",
```
usagi-f: 「フルハンドル」という言い回しはあまり馴染みがないです。
「ハンドル」もしくは「ハンドルネーム」が良いのではないでしょうか。

---

"フルハンドル" is not good.
"ハンドル" or "ハンドルネーム" is better.

Gargron: Should I wait with merging before this is cleared up?

walf443: "Full handle" should mean as global name on mastodon instances, I think. I agree to use "ハンドルネーム" more natural for japanese users. but that is not include that meanings. 

@usagi-f Do you have better idea for this?

walf443: @Gargron It's ok to merge this for the moment. And fix this point after the time.

Gargron (approved):

app/assets/javascripts/components/locales/ja.jsx
usagi-f: @walf443 okay, i think so too

PR #2064, opened 2017-04-18
Change language detection method to user account locale setting.

walf443:
related: #2091 

whatlanguage gem cannot handle to detect some langauge.
https://github.com/peterc/whatlanguage/blob/master/lib/whatlanguage.rb#L10

It maybe good to write patch whatlanguage to handling CJK languages. But it take some time for writing it.

https://instances.mastodon.xyz/list There are already many instances whoes almost people toot on japanese.

This PR detect language by user account's locale setting.
This change maybe not  acceptable for user that toot in English on sometimes and toot in Japanese sometime.

How about it?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,6 @@ gem 'sprockets-rails', :require => 'sprockets/railtie'
 gem 'statsd-instrument'
 gem 'twitter-text'
 gem 'tzinfo-data'
-gem 'whatlanguage'
 
 gem 'react-rails'
 gem 'browserify-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -455,7 +455,6 @@ GEM
     websocket-driver (0.6.5)
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.2)
-    whatlanguage (1.0.6)
     xpath (2.0.0)
       nokogiri (~> 1.3)
 
@@ -542,7 +541,6 @@ DEPENDENCIES
   tzinfo-data
   uglifier (>= 1.3.0)
   webmock
-  whatlanguage
 
 RUBY VERSION
    ruby 2.4.1p111
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -19,7 +19,7 @@ class PostStatusService < BaseService
                                       sensitive: options[:sensitive],
                                       spoiler_text: options[:spoiler_text] || '',
                                       visibility: options[:visibility],
-                                      language: account.user.locale || 'en',
+                                      language: account&.user&.locale || 'en',
                                       application: options[:application])
 
     attach_media(status, media)
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -19,7 +19,7 @@ class PostStatusService < BaseService
                                       sensitive: options[:sensitive],
                                       spoiler_text: options[:spoiler_text] || '',
                                       visibility: options[:visibility],
-                                      language: detect_language(text),
+                                      language: account.user.locale,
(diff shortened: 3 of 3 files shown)


mjankowski:
What if we used:

- a successful whatlanguage result if it has one
- if no WL result, use the user locale
- if user hasn't set one, use the instance default locale

yookoala:
In long run, its best if user can preview the langauge detection and, perhaps, manually override that when detected wrongly.

walf443:
@yookoala Thanks your feedback.  I'm agree that  language detection may return wrong result. So user should be select it.
But it's should not be bother to toot something.

I'd  like to improve language detaction on japanese. But while whatlanguage gem does not have support japanese, it always fail and detected as wrong langauge.

To implemen that, 

 - [ ] create API that detect language by text.
 - [ ] implement UI to call previous API and show detected language.
 - [ ] save text with laugage.

should I make this topic to issue?

walf443:
@mjankowski whatlanguage return wronge language that is not supported language. For example, I tried some japanese word, it detect as 'es'.

Do you mean?
  - [ ] Check whether whatlanguage gem supported lanuage or not.
  - [ ] If supported, use it.
  - [ ] if it does not supported, use the user locale

mjankowski:
> create API that detect language by text.

Yes, I think it would be good to preserve the `detect_language` method that this PR removed ... but instead of calling whatlanguage directly there, have that method call `LanguageDetector.new(string).language_iso` (or similar).

The implementation of that class would be the logic you just described, and should have it's own unit tests with supplied options.  I think some high level logic like:

Is the `I18n.default_locale` contained in the supported whatlanguage list?

If yes, try to use what language to match.

If it was not in the list, or whatlanguage did not return a result, then check if user has a locale set.

If they do, use that; if not, use the default locale on the server.

d6rkaiz:
whatlanguage gem detect sample.

for emoji:
```
$ bundle exec rails c
Creating scope :cache_ids. Overwriting existing method Notification.cache_ids.
Loading production environment (Rails 5.0.2)
irb(main):001:0> WhatLanguage.new(:all).language_iso('🙆')
=> :fa
irb(main):002:0> WhatLanguage.new(:all).language_iso('👍')
=> nil
irb(main):003:0> WhatLanguage.new(:all).language_iso('🙇')
=> nil
irb(main):004:0> WhatLanguage.new(:all).language_iso('😄')
=> :ru
irb(main):005:0> WhatLanguage.new(:all).language_iso('🌞')
=> :fa
irb(main):006:0> WhatLanguage.new(:all).language_iso('🐘')
=> :ru
```

for japanese (日本語)
```
$ bundle exec rails c
Creating scope :cache_ids. Overwriting existing method Notification.cache_ids.
Loading production environment (Rails 5.0.2)
irb(main):001:0> WhatLanguage.new(:all).language_iso('日本語')
=> :ru
irb(main):002:0> WhatLanguage.new(:all).language_iso('こんにちは、mastodon!')
=> nil
irb(main):003:0> WhatLanguage.new(:all).language_iso('明日は🌧です。')
=> :es
```

very very poor.

nightpool:
what language is only supposed to be used for larger texts–definitely not
things in the 100-300 character range.

walf443:
@nightpool if so, most mastodon toot does not over 100 - 300 character length, I think.

Gargron:
We need to find a replacement a more modern/more complete replacement for WhatLanguage. If that cannot detect a language, fallback to user's selected locale. Otherwise, default_locale.

However, this is not urgent, because the language data is not used for any filtering yet. It's just preparation for future features.

Gargron:
Closing in favour of #2099

PR #2073, opened 2017-04-18
Add "Mastodon" to user agent on External HTTP request

ik11235:
I want to add "Mastodon" to user agent on External HTTP request.

Because I think that the current user agent is unintelligible.
```
http.rb/2.2.1
```

So I want to change user agent to this.
```
http.rb/2.2.1 (Mastodon; +http://mastodon.dev/)
```

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class FetchLinkCardService < BaseService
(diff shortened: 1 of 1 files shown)


ik11235:
The reason I want this change is as follows.
Currently, Mastodon is not a user agent named himself.
But, some sites return a special page for OGP to Bot.

I think Mastodon needs to choose the following.
- include "Mastodon"
- do nothing
- Use another BOT's user agent

I think that I should use the site name and "Mastodon".
(The administrator knows the site by looking at the IP address.)

---
To make the discussion simple, I did not include it in this pull request.
 - Include Mastodon version in user agent
 - Changing HTTP requests elsewhere

Gargron (approved):

PR #2075, opened 2017-04-18
Add bundle install jobs option

Peranikov:
The --jobs option (-j for short) installs gems in parallel.

see also:

http://bundler.io/v1.5/whats_new.html
http://bundler.io/v1.14/man/bundle-install.1.html#OPTIONS


Gargron:
This is highly dependent on number of CPU cores, can we dictate the "4" setting for all? I don't think so

Peranikov:
You're right. I'll try `BUNDLE_JOBS` and `--build-arg`.
Thank you.

PR #2000, opened 2017-04-17
Create oc.yml

Quent-in:
Translated in General Occitan

PR #2077, opened 2017-04-18
Add Occitan (oc) translation

yookoala:
* Merged and squash @Quent-in works.
* Fixed YAML formats.
* Remove obsoleted translations.
* Add correct javascript and ruby references.

----

Resolves #2000 
Resolves #2001
Resolves #2002
Resolves #2004


mjankowski (approved):
Thanks for cleaning that all up!

Gargron (approved):

PR #2078, opened 2017-04-18
Fix #1642, fix #1912 - Dictate content-type file extension

### LINKED ISSUE #1642: not everyone can view photos
I'm seeing some inconsistencies in the ability to view posted photos in the last few days.

Some can't see mine, but I can see mine, my instance can see mine (according to the little feedback I've had but that might have been coincidence). I can see some others' photos, but a few I can't. Using Amaroq, I can't get to the photos at all, but at my computer, if I go all the way to the user's homepage (rather than having it come up in the right column) so far I can see them. I've spoken to one other user who noticed the problem, too.

* **

### LINKED ISSUE #1912: Photo cannot be viewed
https://social.tchncs.de/@lrshdl/810702
the preview image is fine, when clicked, it shows a bunch of text. The image doesn't show at all on dash (only a link)

* * * *

Gargron:
Previous change (#1718) did not modify how original file was saved on upload

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -21,6 +21,8 @@ class Account < ApplicationRecord
   validates_attachment_content_type :header, content_type: IMAGE_MIME_TYPES
   validates_attachment_size :header, less_than: 2.megabytes
 
+  before_post_process :set_file_extensions
+
   # Local user profile validations
   validates :display_name, length: { maximum: 30 }, if: 'local?'
   validates :note, length: { maximum: 160 }, if: 'local?'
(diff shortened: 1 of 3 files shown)


mjankowski (commented):
Is the thing this ultimately solves that we wind up with filenames w/out extensions (which is an issue for serving content for some web servers and browsers) ... but where we do know the mime type, and this adds them back?

It's sort of weird that paperclip doesn't have a more built-in way for doing that, but maybe not.

Gargron:
Paperclip doesn't change original filenames *at all* by default, yeah.

Gargron:
Paperclip checks the content type with the `file` utility (which is how it does validation). This adds the correct extension from that.

mjankowski (approved):

PR #2079, opened 2017-04-18
Add rate limits for logins and sign-ups by IP (5 in 5 minutes)

Gargron:
Should be enough for normal attempts

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -16,6 +16,11 @@ class Rack::Attack
     req.ip if req.path == '/auth' && req.post?
   end
 
+  # Rate limit forgotten passwords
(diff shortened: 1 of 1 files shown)


mjankowski (approved):
This looks good.

Is there anything at all under `/auth/` that we DONT want to rate limit? If not, you could consolidate those into one big rule.

Gargron:
Good point about also rate-limiting forgotten password requests - but probably not good idea to unite all of those under the same "key" in rack-attack because a user might reasonably use 3 attempts to sign up, 2 attempts to login, and then it's already at the edge of the rate limit.

mjankowski:
> probably not good idea to unite all of those under the same "key" in rack-attack because a user might reasonably use 3 attempts to sign up, 2 attempts to login, and then it's already at the edge of the rate limit

good point, agreed.

PR #2086, opened 2017-04-18
app: Align the breakpoint width for border layout to the one for columns

akihikodaki:
Commit 45776b55b0d97d6a6b8b202d3076f19f1d055573 gives 1024px for the
breakpoint width for the layout change of the columns, and but the one for
border layout had a different value (1024px).

The fact that the same layout without the border works means the border is
just a waste of the screen area. This change aligns the breakpoint width
for border layout to the one for columns in order to remove the border also
in such cases.

See also https://github.com/tootsuite/mastodon/commit/3807b0b171d588ccccfc6210c823e5ce87b9b90f#commitcomment-21794176. @ticky, who introduced the previous value, has pointed out the possibility that having borders could be preferred on phablet. I thought it is fine without the border when I resized the browser on desktop to test. However, I do not have any actual phablet or similar device, so I would like to ask opinions.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1158,7 +1158,8 @@ a.status__content__spoiler-link {
   position: relative;
 }
 
-@media screen and (min-width: 360px) {
+// min-width corresponds to signle-column + nav layout breakpoint
+@media screen and (min-width: 1024px) {
   .columns-area {
     padding: 10px;
   }
@@ -1211,7 +1212,8 @@ a.status__content__spoiler-link {
   overflow: hidden;
 }
 
-@media screen and (min-width: 360px) {
+// min-width corresponds to signle-column + nav layout breakpoint
+@media screen and (min-width: 1024px) {
   .tabs-bar {
     margin: 10px;
     margin-bottom: 0;


akihikodaki:
9de679407ef03d1238d0e4dc02c9eccb503653c0 adds comments.

ticky:
> The fact that the same layout without the border works means the border is just a waste of the screen area.

I wouldn’t call that true, it was a design decision which was made.

The design actually looks kind of awkward in my opinion with no padding, but given the constraints of a 320px wide display, it’s a tradeoff. To my eye, it looked okay on iPhone 6 size displays and greater, so it was left for those!

IMO, this is a heavy-handed change, and should be carefully considered as there will be people who expect the current layout.

akihikodaki:
> The design actually looks kind of awkward in my opinion with no padding, but given the constraints of a 320px wide display, it’s a tradeoff.

Could you tell me how it is awkward? I think it is not much different from other mobile applications and fine, but I would like to hear your opinion when I make this change.

> IMO, this is a heavy-handed change, and should be carefully considered as there will be people who expect the current layout.

I agree. This change would be a breaking one if it removes a favorable and intended feature.

For your information, here is what I see on my phone. (5.2", 1080x1920)
with border
without border

akihikodaki:
02b81935252e5ce967d347f4a3493f42c657a34d: rebased to 2fba94b36eac65a368ec10f4c8c5808baa0179e7.

Gargron:
We have a new breakpoint so this is no longer needed, I think. Thank you.

PR #2088, opened 2017-04-18
Disable toot button when above 500 characters

tsuwatch:
We can easily send invalid requests many times.
So, not to send request to server from frontend.
Any thoughts?

### DIFF
--- a/app/assets/javascripts/components/features/compose/components/compose_form.jsx
+++ b/app/assets/javascripts/components/features/compose/components/compose_form.jsx
@@ -198,7 +198,7 @@ const ComposeForm = React.createClass({
 
           <div style={{ display: 'flex' }}>
             <div style={{ paddingTop: '10px', marginRight: '16px', lineHeight: '36px' }}><CharacterCounter max={500} text={[this.props.spoiler_text, this.props.text].join('')} /></div>
-            <div style={{ paddingTop: '10px' }}><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || this.props.text.length > 500} /></div>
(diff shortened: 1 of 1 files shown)


Gargron (approved):

app/assets/javascripts/components/features/compose/components/compose_form.jsx
```diff
@@ -198,7 +198,7 @@ const ComposeForm = React.createClass({
 
           <div style={{ display: 'flex' }}>
             <div style={{ paddingTop: '10px', marginRight: '16px', lineHeight: '36px' }}><CharacterCounter max={500} text={[this.props.spoiler_text, this.props.text].join('')} /></div>
-            <div style={{ paddingTop: '10px' }}><Button text={publishText} onClick={this.handleSubmit} disabled={disabled} /></div>
+            <div style={{ paddingTop: '10px' }}><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || this.props.text.length > 500} /></div>
```
Gargron: Try this `this.props.text.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, "_").length` instead - otherwise there's wrong counts for emoji and other multibyte characters.

Gargron (changes requested):

app/assets/javascripts/components/features/compose/components/compose_form.jsx
tsuwatch: It makes sense, modified!

Gargron (approved):

PR #2105, opened 2017-04-18
Fonts cleanup

### LINKED ISSUE #2040: Assets are HUGE.
I was wondering... is there a reason on why the assets folder (compiled) is 103M, and why the js files are 9MB?

screenshot from 2017-04-17 23-28-36

screenshot from 2017-04-17 23-29-26

Really, WTF?!

* * * *

- [X] This bug happens on a [v1.2.0](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

mjankowski:
Removes all the fonts which were commented out in their font-face declarations. Please advise if these were used in some way outside those declarations. Reduces font storage size from about ~50MB to ~12MB.

This does not totally resolve https://github.com/tootsuite/mastodon/issues/2040 but is related to that.


Gargron:
Whoops, I already did that in #2103 D:

mjankowski:
Great!

PR #2111, opened 2017-04-18
Followers-only post federation

Gargron:
- [x] Use PuSH to send `:private` toots to subscribers
- [x] Filter subscribers by only domains where author has known (=authorized) followers
- [x] Settings UI to list followers grouped by domain, with a button to remove any

New settings page:

image

If your account is not locked:

image

"Remove followers" applies soft-blocks to all followers from selected domains. Soft-block means "block, then immediately unblock"

When posting to followers-only with unlocked account:

image

"Private" has been renamed to "followers-only"

Default post privacy settings have been restyled slightly:

image


app/workers/soft_block_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+class SoftBlockWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'pull'
+
+  def perform(account_id, target_account_id)
+    account        = Account.find(account_id)
+    target_account = Account.find(target_account_id)
+
+    BlockService.new.call(account, target_account)
```
mjankowski: The block followed by unblock here has the effect of:

- Remove any follows between the accounts, in both directions
- Create a block from account to target
- Clear the timeline and notifications of the account from anything from the target
- Salmon the block from the account to the target
- ...
- Remove the block
- Send the unblock salmon

This does not prevent accounts on the target domains from immediately refollowing the person, unless they also change their account to private, right?

I just want to make sure I understand the general flow here.

app/controllers/settings/followers_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Settings::FollowersController < ApplicationController
+  layout 'admin'
+
+  before_action :authenticate_user!
+
+  def show
```
mjankowski: Should the `@domains` here be paginated, maybe ordered by most followers? I'm not sure what's typical for how many instances a typical account has following them.

If so, maybe rename this action to `index`

app/controllers/settings/followers_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Settings::FollowersController < ApplicationController
```
mjankowski: I could see this being `Settings::FollowerDomainsController` with a single `index` action for the list, and then something like `Settings::FollowerDomains#update` with the array of domain values instead of `purge`.

Related - how do you feel about a "Block" button next to each domain that would just do them one at a time (and redirect right back to list) -vs- the checkboxes to select a bunch and then submit at once?

app/controllers/settings/followers_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  layout 'admin'
+
+  before_action :authenticate_user!
+
+  def show
+    @account = current_account
+    @domains = current_account.followers.reorder(nil).group('accounts.domain').select('accounts.domain, count(accounts.*) as accounts_from_domain')
```
krainboltgreene: Switch to `COUNT("accounts"."id")`

krainboltgreene (approved):

yiskah:
Question: Is this a toggle-able setting? Having it enabled by default and hidden in preferences will cause the same problems as last time.  I assume this time users will be notified that if they wish they can go into settings and enable this after reviewing who follows them?

Gargron:
This is a hard problem, because it's a one-off event... I wouldn't want to code in special cases for such things so early on in the lifetime of the project... That being said, it's not like old private toots would retroactively travel anywhere. A word-of-mouth announcement might be fine? Alternatively, and this should've been a thing since the start probably, show a warning in web UI if you selected private toot setting but your account isn't locked.

Also I remember everyone wants to rename "private" to "followers-only".

yiskah:
Word of mouth is absolutely not the way to go. The problem last time was that users were suddenly having this big change and didn't know to go auditing their followers because everything is suddenly different. Implementing a warning is a good idea.

dunn:
Are any of the proposed mitigations from https://marrus-sh.github.io/mastodon-info/everything-you-need-to-know-about-privacy-v1.3-020170427.html planned for future releases?

app/workers/soft_block_worker.rb
beatrix-bitrot: was this confirmed?

PR #2127, opened 2017-04-19
Replace User.count -> User.confirmed.count

alpaca-tc:
Everyday `bundle exec rake mastodon:daily` removes users.
IMO, `User.count` is not correct amount of users.

### DIFF
--- a/app/presenters/instance_presenter.rb
+++ b/app/presenters/instance_presenter.rb
@@ -15,7 +15,7 @@ class InstancePresenter
   end
 
   def user_count
-    Rails.cache.fetch('user_count') { User.count }
+    Rails.cache.fetch('user_count') { User.confirmed.count }
   end
 
   def status_count


app/presenters/instance_presenter.rb
```diff
@@ -15,7 +15,7 @@ def contact_account
   end
 
   def user_count
-    Rails.cache.fetch('user_count') { User.count }
+    Rails.cache.fetch('user_count') { User.confirmed.count }
```
mjankowski: Putting aside whether this is a good idea to change or not, this specific way of changing it is going to lead to mixed results ... because the cached result is set elsewhere and is going to still be the total user count. Only in cases where there is a cache miss is this code going to be run.

Gargron: Huh, where else is this cache set? I thought only here?

mjankowski: I was mistaken, not sure what I was thinking of

mjankowski (approved):

PR #2135, opened 2017-04-19
Remove upgrade from docker build

pierreozoux:
As explained [here](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run) it is not necessary to upgrade the underlying distribution, but better to trust the base underlying alpine ruby docker image.
Also use the `--no-cache` option as a shorter form - see [reference](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
     libxml2-dev \
     libxslt-dev \
     build-base" \
- && apk -U upgrade && apk add \
+ && apk --no-cache add \
     $BUILD_DEPS \
     nodejs@edge \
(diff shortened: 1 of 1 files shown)


Wonderfall:
Hi, my opinion : it's Alpine 3.4, I'm not sure if it's a good thing since it's not as maintained as Debian. I think this upgrade is needed to patch vulnerabilities, like the ones in OpenSSL.

Like I said, it's just my opinion. I'll keep the upgrade on my local Dockerfile.

pierreozoux:
Indeed the last push was 2 months ago...
Let's wait the response here: https://github.com/gliderlabs/docker-alpine/issues/270

Gargron:
I will close this for now.

PR #2138, opened 2017-04-19
Fix html escape characters in the URL

abcang:
Fixed the following issue.

* https://github.com/tootsuite/mastodon/issues/826
* https://github.com/tootsuite/mastodon/issues/678

Run encode and link simultaneously to prevent join between angle bracket and URL.

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -54,7 +54,7 @@ class Formatter
     chars = html.to_s.to_char_a
     html_attrs = {
       target: '_blank',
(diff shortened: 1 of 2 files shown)


Gargron (changes requested):
Could you please add a test case for when local_status contains html (xss attack) to make sure it gets escaped correctly? 

If this works, great job! Thank you!!

abcang:
Ok, I will add one.

Gargron (approved):

PR #2147, opened 2017-04-19
Fix possibility of unrightful webfinger redirect

Gargron:


mjankowski (changes requested):
This feels security related and like there's missing context. My guess from the changes is there's some scenario where the URI value we start with doesn't match the webfinger lookup...?

Can you give a little background on what this is checking, and add a spec around whatever correction we want in place here?

PR #2154, opened 2017-04-19
Don't toot request with only blank characters

tsuwatch:
toot with only blank characters is invalid, so don't request to a server also in this case.

### DIFF
--- a/app/assets/javascripts/components/features/compose/components/compose_form.jsx
+++ b/app/assets/javascripts/components/features/compose/components/compose_form.jsx
@@ -174,7 +174,7 @@ class ComposeForm extends React.PureComponent {
 
           <div className='compose-form__publish'>
             <div className='character-counter__wrapper'><CharacterCounter max={500} text={text} /></div>
-            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || text.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, "_").length > 500 || text.trim().length === 0} block /></div>
+            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || text.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, "_").length > 500 || (text.length !==0 && text.trim().length === 0)} block /></div>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


ashfurrow (approved):
Looks good! You've tested locally?

app/assets/javascripts/components/features/compose/components/compose_form.jsx
```diff
@@ -174,7 +174,7 @@ class ComposeForm extends React.PureComponent {
 
           <div className='compose-form__publish'>
             <div className='character-counter__wrapper'><CharacterCounter max={500} text={text} /></div>
-            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || text.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, "_").length > 500} block /></div>
+            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || text.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, "_").length > 500 || text.trim().length === 0} block /></div>
```
Gargron: Please do not disable the toot button for empty statuses. A new user must be greeted with a blue toot button. If you wish to prevent server requests, stop them in actions/statuses.jsx

Gargron (changes requested):

app/assets/javascripts/components/features/compose/components/compose_form.jsx
ashfurrow: Great point!

tsuwatch: It makes sense.
However, I think it makes confusing that the button with text is enabled, but do not request.
How about `(text.length !==0 && text.trim().length === 0)` ?

tsuwatch:
Yes, I tested local environment.

app/assets/javascripts/components/features/compose/components/compose_form.jsx
Gargron: Yes, that would be okay.

tsuwatch: Done, thanks!

Gargron (approved):

PR #2162, opened 2017-04-19
Extract error messages to locale file

alpaca-tc:
localize validation errors


mjankowski (approved):

Gargron (approved):

alpaca-tc:
:yellow_heart:

Gargron:
I didn't notice this, but the new locale files are not in the config/locale folder... So they probably don't work.

PR #2165, opened 2017-04-19
Move force_ssl check to production config

mjankowski:
The force_ssl method from controllers does not add all of the options that the
sitewide configuration in a config block does. For example, HSTS enforcement is
not added by the controller method, but is added by this style.

I see that the exact opposite of this change was done here - https://github.com/tootsuite/mastodon/commit/8459acd1234e9efcbf106e3eaa00ca871717d386 - but I'm not sure why.

Opening this PR to either sort that out, or close if still a concern.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,8 +5,6 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  force_ssl if: "Rails.env.production? && ENV['LOCAL_HTTPS'] == 'true'"
-
   include Localized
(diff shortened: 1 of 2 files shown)


mjankowski:
My guess is there might have been an issue with websockets and production and ssl...?

Gargron (approved):
I hope this will be fine

mjankowski:
I did a little more reading on this and through the backlog ... I think if requests to `/streaming/` things are being done *through* the rails app, this might cause a problem for people ... but I also don't know how you'd set that up to happen. I suspect people with streaming enabled have some routing at the web server level before it even hits the rails app which is sending those requests to the node app instead.

PR #2167, opened 2017-04-19
Updated note about 'none' SMTP authentication method in .env.production.sample

evilny0:
#1597 added 'none' to SMTP authentication. However, I spent some time not understanding why it was still not working. I had to comment SMTP_LOGIN & SMTP_PASSWORD.

Hopefully this will help others not losing time in case it happens for them as well.


Gargron (approved):

mig5:
This is actually a regression, the proper fix is https://github.com/tootsuite/mastodon/pull/2180 (but commenting out the setting altogether works too)

PR #2168, opened 2017-04-19
Add static gif for upload testing

saper:
This is reproduces https://github.com/tootsuite/mastodon/issues/2108


Gargron:
This is an additional test that should be merged right after #2171, which will make these tests pass. :+1:

saper:
Magic. They pass :)

Gargron (approved):

Issue #1789, opened 2017-04-14
Display version number on Admin page


thegibsonmainframe:
Please add version numbers somewhere in admin pages..

* * * *

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

wxcafe:
It would be nice to add it somewhere user-facing too, like in this text 
<img width="307" alt="screen shot 2017-04-15 at 01 44 22" src="https://cloud.githubusercontent.com/assets/3009327/25058744/2b16a7e4-217d-11e7-93ec-0da9d787d867.png">

ericblade:
Wouldn't it be probably most useful on the login screen?

thegibsonmainframe:
Probably so.

---- On Fri, 14 Apr 2017 22:42:42 -0700 notifications@github.com wrote ----

Wouldn't it be probably most useful on the login screen?

wxcafe:
I'd say both, cause logged-in users never see the login screen

SuliacLEGUILLOU:
I would like to try to work on this o/
I was thinking about some parsing on `git status` during asset compilation so dev could have information on the branch they are looking at.
Does anyone saw some other way around this issue ?

wxcafe:
`git status` and `git log` if it's not on master, yeah, sounds good

ashfurrow:
I think `git describe --tags` might strike the balance between useful and concise. It outputs something like `v1.2-80-gadeda7bf`, where `v1.2` is the latest tag, `80` is the number of commits done since that tag, and `gadeda7bf` is the specific commit git is on.

ericblade:
wow, never knew about git describe --tags.  That could be really handy to me in other things, too.

You could use environment variables, such as here https://github.com/moby/moby/issues/14191#issuecomment-242718433 .. 

Although I think you'd also need to tweak docker-compose as well to get that solution to work, and I'm just not familiar enough with Docker and Compose to advise beyond that. 

I did just try adding git to the container and running the git describe inside it, but the .git directory doesn't make it to the container (this is not surprising)

wxcafe:
you could also just add it to a file when you docker-compose and when you build the assets

SuliacLEGUILLOU:
@ashfurrow I just tested : My vagrant container has access to git and there for can sumon `git describe`
but I don't have any docker environment set up to test on the production env :/

ashfurrow:
[...]

PR #2169, opened 2017-04-19
Adds git to dependencies

ashfurrow:
As discussed [in this issue](https://github.com/tootsuite/mastodon/issues/1789#issuecomment-295304981), we would like to have the output from `git describe --tags` displayed in the UI of the Rails app as a sort of version number. To do so, `git` needs to be installed in the Docker images. This PR adds that.


ashfurrow:
Closing after discussion in #1789.

PR #2171, opened 2017-04-19
Fix #2108 - Fix gif uploads

### LINKED ISSUE #2108: Static gif upload fails
When trying to upload any non-animated gif, the upload fails with a 500

Tested on v1.2

```
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] Command :: file -b --mime '/tmp/d60cc6a6fba8b5fe7c9d59f6c5fdef0a20170418-14-1c8sm2r.gif'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] Command :: identify -format %n '/tmp/d60cc6a6fba8b5fe7c9d59f6c5fdef0a20170418-14-g31j0f.gif'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] Command :: identify -format %n '/tmp/d60cc6a6fba8b5fe7c9d59f6c5fdef0a20170418-14-g31j0f.gif'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] Command :: file -b --mime '/tmp/d60cc6a6fba8b5fe7c9d59f6c5fdef0a20170418-14-12znes6.gif'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] method=POST path=/api/v1/media format=html controller=Api::V1::MediaController action=create status=500 error='IOError: closed stream' duration=106.30 view=0.00 db=5.52
[795947d7-3a61-4fc6-8ee8-3c7f0055a531]   
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] IOError (closed stream):
[795947d7-3a61-4fc6-8ee8-3c7f0055a531]   
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] app/controllers/api/v1/media_controller.rb:13:in `create'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] app/controllers/concerns/localized.rb:24:in `block in set_locale'
[795947d7-3a61-4fc6-8ee8-3c7f0055a531] app/controllers/concerns/localized.rb:23:in `set_locale'
```

Heres the image i've been testing with
hit-the-pinata

* * * *

Gargron:
Add specs for media attachment gifv conversion


lib/paperclip/gif_transcoder.rb
```diff
@@ -7,7 +7,11 @@ class GifTranscoder < Paperclip::Processor
     def make
       num_frames = identify('-format %n :file', file: file.path).to_i
 
-      return file unless options[:style] == :original && num_frames > 1
+      unless options[:style] == :original && num_frames > 1
+        tmp_file = Paperclip::TempfileFactory.new.generate(attachment.instance.file_file_name)
+        tmp_file << file.read
+        return tmp_file
```
saper: why is that needed? I was checking this code a moment ago and I don't get it

app/models/media_attachment.rb
```diff
@@ -106,13 +106,18 @@ def set_shortcode
...
+    file.instance_write :file_name, [basename, extension].delete_if(&:empty?).join('.')
+  end
+
+  def appropriate_extension
+    mime_type = MIME::Types[file.content_type]
+
+    extensions_for_mime_type = mime_type.empty? ? [] : mime_type.first.extensions
+    original_extension       = Paperclip::Interpolations.extension(file, :original)
+
+    extensions_for_mime_type.include?(original_extension) ? original_extension : extensions_for_mime_type.first
```
saper: Why is this necessary?

lib/paperclip/gif_transcoder.rb
Gargron: The `IOError: closed` exception comes from Paperclip closing the file too early. After investigating this (thoughtbot/paperclip#875) the solution is to create an extra temp file to make sure it's open (I know, eugh)

app/models/media_attachment.rb
Gargron: Because the bahaviour of `content_type_extension` interpolator method from Paperclip is to return the format dictated by the style definition, rather than the content type of the actual file, first. In this case, the final style is mp4, so that method was renaming the file to actuallyagif.mp4, which made paperclip's content spoofing check fail.

saper: So maybe the style definition is wrong here? I thought the intention was here if the number of frames is 1 we leave the file alone?

Gargron: Basically we have to fight paperclip here a bit because what we're doing is more complicated than the library is meant to handle. Here is what we *want* to do: if it's an animated gif, convert it to video. If not, leave as is. But at the same time, we want to modify the *original* files, because the whole point is conserving space. However, paperclip doesn't handle modifications of the original file very well.

What you see here is the result of a lot of troubleshooting with `binding.pry` and reading paperclip's source code. It's not pretty code, and it's not very logical, but it works.

mjankowski (approved):
I have not done an exhaustive look through on this, but can confirm it works locally and specs pass (once ffmpeg installed).

app/models/media_attachment.rb
saper: Last question I promise... could the num_frames check be done earlier? For example in file_styles ? This way we could avoid doing anything to a static gif? My reading of Paperclip::Attachment is that unless style says something has to be done the processors will not be even called...

Gargron: Iirc the `identify` method is part of `Paperclip::Processor`, so harder to call outside of it.

PR #2172, opened 2017-04-19
Implement new unified rules for user account URIs

### LINKED ISSUE #2032: OStatus user URI verification/disambiguation
Currently, Mastodon will happily accept multiple remote accounts with the same account URI, which is ambiguous when interacting with remote instances: users are mentioned by their URI, and salmon requests are verified against the known public-key of a user based on its URI.

I might be wrong, but it seems OStatus does not require URIs to be “authenticated” in any way, so multiple remote users could use the same URI with no way to chose which one is entitled to use it. It could be used on a first come, first served basis, but then, different instances might have different “users” associated to the same URI, leading to “incompatible” views of the same toots, and it would block out salmon requests of one of the users.

This could have privacy implications, but I'm not completely sure which. I guess private messages should be fine, as users are locally identified by their webfinger handle, and not by their URI.

To be honest, I have a hard time believing there is such a flaw in OStatus, so I could be entirely wrong, and I'd be happy to be proven so.

I can only see one workaround, that is, only accepting the URIs that are either:
- URLs linking back to an atom feed with either:
  - an author/email tag corresponding to the initial webfinger query
  - if no author/email tag, an author/name tag corresponding to the username part of the webfinger acct, hosted at a domain corresponding to the domain part of the webfinger acct (required for quitter.no)
- webfinger “acct:” URIs that correspond to the initial query

ClearlyClaire:
This pull request is an attempt to fix #2032 and #2012.
The first commit specifically addresses #2032 by only accepting remote account URIs (the ones provided by the atom feeds) if they are either:
- An acct:user@domain URI which matches the webfinger acct:user@domain
- An URL that points to either an atom feed or a webpage referencing to one, and that atom feed maps back to the original user@domain
  - If the atom feed as an author/email node, it is said to map to its value
  - Otherwise, if it as an author/uri node and an author/name node, those are used to build the value it maps to
Note that it may perform additional requests, as it will attempt to fetch the atom feed accessible from the URI (some optimization should be possible here, as the atom feed should be the same that was used to discover the URI in the first place).

The following commits address #2012 by making various code sites use the functions created in the first commit instead of using their own logic.

The last commit adjusts the spec with the need to fetch a new URL: the web profile linking to the atom feed.

I broke the service convention you were using so far as it seemed the most convenient way to implement these changes, but any suggestion on how to cleanly do otherwise would be appreciated.

There are also two things that I think are still missing from this pull requests: additional rspec tests to validate the corner cases, and a way to remove users sharing the same URI as the one being successfully resolved/added: indeed, the whole purpose of this change is to ensure URIs are unique, but existing instances might already have colliding URIs for different user@name handles. Worse, in case of a migration (as mentioned in https://github.com/tootsuite/mastodon/pull/1267#issuecomment-294677714), one URI might refer to a different user@domain handle than it was referring to in the past. For this reason, other accounts sharing the same URI should be removed. It might also be advisable to translate references to the old account to the new one instead of deleting them, but I'm not completely sure about that.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -181,6 +181,10 @@ class Account < ApplicationRecord
     "acct:#{local_username_and_domain}"
   end
 
+  def needs_webfinger_update?
+    last_webfingered_at.nil? || last_webfingered_at <= 1.day.ago
+  end
+
   def subscribed?
     !subscription_expires_at.blank?
   end
--- a/app/services/fetch_remote_account_service.rb
+++ b/app/services/fetch_remote_account_service.rb
@@ -18,21 +18,12 @@ class FetchRemoteAccountService < BaseService
   def process_atom(url, body)
     xml = Nokogiri::XML(body)
     xml.encoding = 'utf-8'
+    follow_remote_account_service = FollowRemoteAccountService.new
+    uri = follow_remote_account_service.acct_uri_from_atom(xml)
 
-    email = xml.at_xpath('//xmlns:author/xmlns:email').try(:content)
-    if email.nil?
-      url_parts = Addressable::URI.parse(url).normalize
-      username  = xml.at_xpath('//xmlns:author/xmlns:name').try(:content)
-      domain    = url_parts.host
-    else
-      username, domain = email.split('@')
-    end
+    Rails.logger.debug "Going to webfinger #{uri}"
 
-    return nil if username.nil? || domain.nil?
-
-    Rails.logger.debug "Going to webfinger #{username}@#{domain}"
-
-    account = FollowRemoteAccountService.new.call("#{username}@#{domain}")
+    account = follow_remote_account_service.call(uri)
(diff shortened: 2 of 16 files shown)


ThibG:
I have rebased on top of latest master, but I have seen you're making changes to the salmon handling code to resolve part of this issue, I will read that properly later on.

ThibG:
Commit https://github.com/tootsuite/mastodon/commit/bafd22ecf487774c252a271d668716b0e1c84c6c covers a large part of this pull request, so I'm closing it for now, but some things are still missing:
- Ensuring uniqueness of actor URI (by mapping it back to an acct URI for instance)
- Support for acct URI (I've not seen anybody use that, but that is something I want to push forward as it is as usable as a profile URL while being exactly what software like Mastodon actually wants as a primary unifier)
- Tests (especially for the remote interaction part)

PR #2174, opened 2017-04-19
Use local_time gem for timezone replacement

mjankowski:
For the 'time ago' value of stream entry simple statuses and the timestamp value
of detailed statuses - output their UTC value and use the browser timezone to
convert locally.

Resolves https://github.com/tootsuite/mastodon/issues/1990


Gemfile.lock
```diff
@@ -224,6 +231,8 @@ GEM
       letter_opener (~> 1.0)
       railties (>= 3.2)
     link_header (0.0.8)
+    local_time (1.0.3)
+      coffee-rails
```
Gargron: Uh, we're reintroducing coffee...

app/views/stream_entries/_detailed_status.html.haml
```diff
@@ -32,7 +32,7 @@
   %div.detailed-status__meta
     %data.dt-published{ value: status.created_at.to_time.iso8601 }
     = link_to TagManager.instance.url_for(status), class: 'detailed-status__datetime u-url u-uid', target: stream_link_target, rel: 'noopener' do
-      %span= l(status.created_at)
+      %span= local_time(status.created_at, format: :default)
```
Gargron: So, what does this mean for i18n? Does it still use the YML strings?

Gemfile.lock
mjankowski: There are some libs that use jquery and dont need coffee that do "time ago" and some that do "replace utc time with local tz time" ... but this was the best I could find that did both, and it pulls in coffee-rails.

If "lets not have coffeescript anywhere" is a higher level goal, I can revisit and either use two separate libs, or look for another that does both with jquery-only.

app/views/stream_entries/_detailed_status.html.haml
mjankowski: Yeah, it uses `time.formats.default`, in this case, and both the ruby method and the JS support the bulk of the options available for localizing time formats.

Here's sample output:

    <span><time datetime="2017-04-16T16:38:48Z" data-local="time" data-format="%b %d, %Y, -- %H:%M">Apr 16, 2017, -- 16:38</time></span>

In this case the record was created at 16:38 UTC time, so that is what populates the datetime attribute and also what is in the value ... but then the JS converts this to local time (attached image)...

screen shot 2017-04-20 at 9 47 25 pm

The formatting is preserved on the JS conversion.

I believe that the datetime formatting use here in the detailed status is the same as before.

app/views/stream_entries/_simple_status.html.haml
```diff
@@ -1,7 +1,7 @@
 .status.light
   .status__header
     .status__meta
-      = link_to time_ago_in_words(status.created_at), TagManager.instance.url_for(status), class: 'status__relative-time u-url u-uid', title: l(status.created_at), target: stream_link_target, rel: 'noopener'
+      = link_to local_time_ago(status.created_at), TagManager.instance.url_for(status), class: 'status__relative-time u-url u-uid', title: l(status.created_at), target: stream_link_target, rel: 'noopener'
```
mjankowski: It might be worth noting that the format used in the stream entries one here is a slight format change.

Currently everything is like "13m", "39m", "3h", "10h", "1d", "2d", ... and I think never does more than the days ago count of n in a string like "nd".

The change here would behave differently, but has options: https://github.com/basecamp/local_time/blob/master/README.md#time-ago-helper

Gargron (approved):

PR #2180, opened 2017-04-20
Regression fix for unauthenticated SMTP

mig5:
A small regression with the fix to send unauthenticated SMTP introduced in https://github.com/tootsuite/mastodon/pull/1597. 

As I suspected (and my original patch fixed), you have to set SMTP_LOGIN and SMTP_PASSWORD to nil, even if blank in the envfile, otherwise brain damage in Ruby will still attempt auth even if SMTP_AUTH_METHOD is none

Do this with .presence so the value is set if it was provided, otherwise nil if blank.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -98,12 +98,10 @@ Rails.application.configure do
   config.action_mailer.smtp_settings = {
     :port                 => ENV['SMTP_PORT'],
     :address              => ENV['SMTP_SERVER'],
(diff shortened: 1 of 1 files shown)


mig5:
Note: I haven't attempted a valid SMTP auth with this fix, so testing is appreciated from someone who uses mailgun or whatever to make sure it's not a regression to fix a regression.

Gargron (approved):

PR #2181, opened 2017-04-20
Adds version to about/more and API

### LINKED ISSUE #1789: Display version number on Admin page

ashfurrow:
This adds a new `lib/version.rb` with a new `VERSION` constant string. This string is returned in `/api/v1/instances` as well as displayed on the about/more page.

(Edit: new screenshot.)

screen shot 2017-04-19 at 10 33 18 pm

screen shot 2017-04-19 at 8 27 23 pm

I'm still pretty new to this, so please feel free to point out anything that could be improved.

Fixes #1789.


csu:
It might make more sense to add this to the existing "Get current instance information" API endpoint (https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#instances) than to create a new one just for the version.

ashfurrow:
Makes sense to me, I'll update 👍

badosu:
Wouldn't it be sensible to use `lib/mastodon/version.rb`?

ashfurrow:
I mean, yeah that would make sense too. There's no current `mastodon` directory so this would be the only file in it. I wanted to keep it near the top of the directory structure because it is fairly important. But I'm open to moving it.

badosu:
I thought of this thinking of gem and namespace conventions, this is just a nitpick.

ashfurrow:
Gotcha. I'm happy to move it wherever, if that's more idiomatic.

app/views/about/_version.html.haml
```diff
@@ -0,0 +1,4 @@
+.panel
+  .panel-header= t 'about.version'
+  .panel-body
+    %strong= Mastodon::VERSION
```
mjankowski: Minor point, but as a general rule I try to avoid using constants directly in views ... can you wrap this up in InstancePresenter (view it to see how other values are already doing this), and then pass that in to the partial (see how `contact` is already doing that)?

mjankowski (changes requested):
I agree w/ moving it to `lib/mastodon/version.rb`.

I had one other comment about using the constant in the views, otherwise looks fine.

app/views/about/_version.html.haml
ashfurrow: Makes sense, will do 👍

ashfurrow:
Okay, thanks @mjankowski and @badosu for the feedback, I've applied those changes, let me know how it looks 👍

mjankowski (approved):

mjankowski:
This looks fine to me - I will defer to @Gargron on final approval here, because he's the one who will have to remember to change that number when he tags releases.

ashfurrow:
Makes sense!

Gargron (approved):

PR #2182, opened 2017-04-20
Fix #2120 - Use Status#as_tag_timeline on public hashtag page

### LINKED ISSUE #2120: Private and direct statuses with hashtags are exposed on standalone tags page
You can see examples [here](https://mastodon.social/tags/anteater) and [here](https://mastodon.social/tags/aardvark), and it should be easily reproducible by posting a private or direct toot with a hashtag in it, and then going to the corresponding standalone page for said tag while logged out.

These toots don't show up in the search in the web app, so I don't think it's by design.

Of course, users ordinarily should have no reason to use hashtags in toots at those privacy levels, but sometimes you do for meta reasons anyway (discussing site changes in a direct message thread is how I discovered it).

* * * *

Gargron:


mjankowski (approved):
The change itself looks fine.

Ideally we'd also do one of:

- Add a feature spec which shows that only public statuses are on the tags show page
- Add controller spec verifying that `Status.as_tag_timeline` is used here

PR #2185, opened 2017-04-20
Removes timestamp from URLs

### LINKED ISSUE #2123: Remove cache-busters from media URLs
The media URLs in the interface have a cache-buster on them (the current time in a query string).

`https://social.wxcafe.net/system/media_attachments/files/000/025/874/original/c5cfc535c7c45c95.png?1492570810`  
specifically the **`?1492570810`**

This isn't needed in the normal case, and harms native browser caching (the image must be redownloaded every time the cache-buster is regenerated). You only need to apply cachebusters if the previous load is stale or was bad.

It also harms usability with right click -> Copy link to image, as the extra numbers are confusing.

Proposal: Don't put cache-busters on media URLs.

* * * *

ashfurrow:
Documented [here](https://github.com/thoughtbot/paperclip/blob/7edb35a2a9a80c9598dfde235c7e593c023fc914/lib/paperclip/attachment.rb#L59).

Fixes #2123.


Gargron (approved):

Kjwon15:
This removes timestamp only when S3 enabled. Is there any reason to remain timestamp on other condition?

PR #2192, opened 2017-04-20
Extended length statuses without disrupting timelines (Monologuing)

KScl:
See [this monologue post](https://is.a.qute.dog/@KS/2272), and [this reply to it](https://is.a.qute.dog/@KS/2539), which demonstrates the feature on a live instance.  Multiple people have been telling me to submit a PR for this, though I have some concerns about code quality (mainly, I'm not sure I'm doing everything 100% right).

This PR adds support for extended length statuses, 'monologues', to be posted in a manner not unlike [TwitLonger](http://www.twitlonger.com), except the full status is displayed on the detailed status page, rather than being stored off site.  Enabling 'monologue mode' only requires clicking a single button (added to the right of the Content Warning button), and the server will automatically truncate the post to fit within the 500 character limit, and inserting the link to the detailed status page at the end to view the full content.  The truncated post is displayed on timelines, as context for other posts, or basically anywhere else that *isn't* the post's detailed status page.  As the truncated text is just stored as the status's "text", and the full text is only stored and served locally, this does not affect federation in any way.

As an added bonus, 'monologue' statuses can use Markdown formatting to emphasize text, organize text into multiple sections, etc..  This adds the gem [Redcarpet](https://github.com/vmg/redcarpet) as a dependency.

There is one minor change to UI; the character counter was moved below the Toot button, as the addition of a fifth option to toggle caused the Toot button to get pushed off of the column.  I'm not sure it's the best solution to the problem, but it's *a* solution to the problem, which beats just leaving it there.

As a side note, I haven't written tests for the new behavior and I *probably should* do that at some point in the future.


ericblade:
I'm just a tiny instance admin, but I like this idea.  I'd rather see it auto-trigger, and possibly even around 250 characters or so (i believe there was a discussion about having links count as fewer characters also, this would go well with that change, if it has happened or will happen), rather than have the user decide to toggle it on or not . . imo, the current UI format for the main web app, isn't well configured for 500 character posts .. fonts are too large, columns are too thin...

Gargron:
I commend your effort, and it is a nice implementation. However, we don't really want to force people to leave their app/client to view something if we can help it, so I don't think this will be part of the main repository.

PR #2197, opened 2017-04-20
ActivityPub: Add basic, read-only support for Outboxes, Notes, and Create/Announce Activities

evanminto:
As part of [ActivityPub](https://www.w3.org/TR/activitypub/) integration (see [related ticket](https://github.com/tootsuite/mastodon/issues/1557)), this branch adds [Activity Streams 2.0](https://www.w3.org/TR/activitystreams-core/) outboxes, accessible via URLs published in users' AS2 account representations. I tried to keep the changes fairly limited, but in order to create a functional outbox endpoint I needed to create a few related endpoints as well, for notes and activities. Those endpoints have basic support for reblogging and replies, which was *maybe* overkill, so I'm happy to take them out if we want to keep things simpler for now. There are fairly extensive tests of all of the endpoints.

Here's how I mapped from Mastodon-land to AP-land:
* Mastodon's statuses, which represent a user posting a message on the site, have two representations, in keeping with ActivityPub standards: a Note object representing the actual document that the user created, and a Create object representing the action of creating it.
* Why the distinction? Reblogging shows a clear case for it, since a reblog is represented as a single Announce activity, with the reblogged note as its object property.
* For now, **the outbox only shows public statuses**, which means it's basically like the public timeline, but for a single user. This seemed like the safest bet, but it may be more appropriate to include unlisted statuses as well (since this is what the profile page does). **Unlisted statuses are still accessible directly via their AP URL, but private and direct statuses are NOT.**

Potential areas to change:
* Should we use Announce for reblogs? AP lacks a clear reblogging function AFAICT. If so, should we announce the original note, or its creation activity? I went with the latter.
* Should we include unlisted statuses in the outbox?

Also my commit history is a *little* messy so I'd recommend squashing if you're not doing that already.


app/controllers/accounts_controller.rb
```diff
@@ -15,7 +15,9 @@ def show
         render xml: AtomSerializer.render(AtomSerializer.new.feed(@account, @entries.to_a))
       end
 
-      format.activitystreams2
+      format.activitystreams2 do
+        headers['Access-Control-Allow-Origin'] = '*'
```
Gargron: You should check `config/application.rb` for CORS settings, and move all of the CORS stuff there.

app/controllers/api/activitypub/outbox_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    @statuses = Status.as_outbox_timeline(@account).paginate_by_max_id(limit_param(DEFAULT_STATUSES_LIMIT), params[:max_id], params[:since_id])
+    @statuses = cache_collection(@statuses)
+
+    set_maps(@statuses)
+
+    # Since the statuses are in reverse chronological order, last is the lowest ID.
+    @next_path = api_activitypub_outbox_url(max_id: @statuses.last.id) if @statuses.size == limit_param(DEFAULT_STATUSES_LIMIT)
+
+    unless @statuses.empty?
```
Gargron: I feel like you could reuse logic from other API methods almost vebatim here, not sure why this complicated block

app/helpers/activitystreams2_builder_helper.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Activitystreams2BuilderHelper
+  # Gets a usable name for an account, using display name or username.
+  def account_name(account)
+    account.display_name.empty? ? account.username : account.display_name
```
Gargron: As I've learned, this should be written as `account.display_name.presence || account.username`

app/views/activitypub/types/announce.activitystreams2.rabl
```diff
@@ -0,0 +1,3 @@
+extends 'activitypub/intransient.activitystreams2.rabl'
+
+node(:type) { 'Announce' }
```
Gargron: ActivityPub has no verb for "share"?

Reblogs should definitely be "share [original note]", I don't feel like "announce [creation of original note]" is semantic at all

app/views/api/activitypub/outbox/show.activitystreams2.rabl
```diff
@@ -0,0 +1,23 @@
...
+  @statuses.map { |status| api_activitypub_status_url(status) }
+end
+
+node(:totalItems) { @statuses.count }
+node(:next)       { @next_path } if @next_path
+node(:prev)       { @prev_path } if @prev_path
+
+node(:name)       { |account| t('activitypub.outbox.name', account_name: account_name(account)) }
+node(:summary)    { |account| t('activitypub.outbox.summary', account_name: account_name(account)) }
+node(:updated) do |account|
```
Gargron: Local statuses are always strictly chronological so it's enough to get the "first" time without loops and max.

config/locales/en.yml
```diff
@@ -39,6 +39,15 @@ en:
     posts: Posts
     remote_follow: Remote follow
     unfollow: Unfollow
+  activitypub:
```
Gargron: If this is only ever going to be seen by other servers, is i18n necessary here?

Gargron (changes requested):
👍 

I feel like something could be said about the naming of the templates, but it's a minor detail that can be changed later.

config/locales/en.yml
evanminto: This is an area I don't know enough about. I did this just to be sure, but I'll run through the AP documentation again to check if there's a need for internationalization.

app/controllers/accounts_controller.rb
evanminto: I've been trying to do that, but it doesn't seem to be working. I may just leave it out of this PR (it's not strictly necessary right now).

app/views/activitypub/types/announce.activitystreams2.rabl
evanminto: It's a little odd, but I did some more reading and it looks like Announce is the canonical AP way to do this: https://github.com/w3c/activitypub/issues/186

That said, it's not a bad idea to at least change the name of the activity to say "share."

config/locales/en.yml
evanminto: For reference, here's the AS2 section on natural language values: https://www.w3.org/TR/activitystreams-core/#naturalLanguageValues

It looks like we can either just do English or provide all of the language options that Mastodon supports in the same JSON object. I'd err on the side of the former for now, but whether to add full language support later is your call.

evanminto:
I realized when I took another look at this that my implementation of pagination isn't quite right for the AS2 spec. I'm taking another look at it right now.

Gargron (approved):

PR #2213, opened 2017-04-20
Stricter whitelist rules

Fiaxhs:
I was browsing an instance where you need a whitelisted email domain, however I figured out you could circumvent the protection by using the whitelisted domain as a subdomain of your email address.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -108,5 +108,15 @@ RSpec.describe User, type: :model do
       user = User.new(email: 'foo@mastodon.space.userdomain.com', account: account, password: password)
       expect(user.valid?).to be_falsey
     end
+
+    it 'should not allow a user to be created with a specific blacklisted subdomain even if the top domain is whitelisted' do
+      old_blacklist = Rails.configuration.x.email_blacklist
+      Rails.configuration.x.email_domains_blacklist = 'blacklisted.mastodon.space'
+
+      user = User.new(email: 'foo@blacklisted.mastodon.space', account: account, password: password)
+      expect(user.valid?).to be_falsey
+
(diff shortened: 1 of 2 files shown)


mjankowski (commented):
Could you add a spec showing that blacklist doesn't have the same issue?

Also ... instead of the regex approach here, what if we just extracted the hostname portion of the email and did a (case insensitive?) check that it was in the list?

Fiaxhs:
spec for blacklist: on it.
We could drop the regexp instance yes, but only for whitelisting then.

Fiaxhs:
I tried to ditch the whitelist regexp, but it created some important issue in my opinion: you can't use it as wildcard to whitelist subdomains.

The instance I found this issue on is a big institution using `institution.tld` as top domain for emails and every service within this instituion as a specific subdomain: `user@service1.institution.tld`, `user@service2.institution.tld` etc.

By removing the regexp, the admin has the tedious task to list all subdomain of whitelisted subdomains (and add them if they're ever created).

I added a spec for specific subdomain blacklist + top domain whitelist.

Anyway, tell me what's the desired approach :)

spec/models/user_spec.rb
```diff
@@ -87,5 +103,20 @@
       user = User.new(email: 'foo@mastodon.space', account: account, password: password)
       expect(user.valid?).to be_truthy
     end
+
+    it 'should not allow a smart user to be created unless they are whitelisted' do
```
Gargron: Instead of "smart user" could you just describe what is being tested? This will become unclear in the future.

Gargron (changes requested):

Gargron (approved):

PR #2218, opened 2017-04-20
Fix can toot unicode spaces

magurotabetai:
I found fast_blank cannot validate unicode spaces.
For instance "\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200A\u202F\u205F\u3000" can be posted without validation actually.
So I replaced blank_as? to blank? for fixing bug.
[Reference URL](https://github.com/SamSaffron/fast_blank#compatibility-note "Reference URL")


Gargron:
What are the performance implications of *always* using `blank_as?` instead of `blank?`? Besides, I am not sure toots with unicode spaces are a huge issue.

magurotabetai:
@Gargron According to [benchmark](https://github.com/SamSaffron/fast_blank#how-fast-is-fast "benchmark")([code](https://github.com/SamSaffron/fast_blank/blob/master/benchmark)), `blank_as?` is as fast as `blank?` .
So it doesn't affect performance.

Gargron (approved):

PR #2226, opened 2017-04-20
Redirect to short_account_path when accessing /web/accounts/:id without signed in

tsuwatch:
Would like to be redirected.
I think it would be more convenient.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -14,7 +14,10 @@ class HomeController < ApplicationController
   private
 
   def authenticate_user!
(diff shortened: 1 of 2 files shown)


mjankowski (changes requested):
Can you describe this use case a little bit more, and why this makes sense?

If someone was trying to go to the signed-in-app page for a certain user, doesn't it make sense to bring them to a place they can sign in from, and not back to the public version of that user's page?

If they wanted the public page ... they'd have gone there?

tsuwatch:
It's exactly as you said...
I was thinking only about the convenience of unregistered users.
Users have a public page, but unregistered users can't know it from this URL.
But, I also think that this method is not a good way.

Thank you for reviewing.

PR #2230, opened 2017-04-20
Fix Chinese (China) translation availability

orzFly:
zh-CN is not referenced in the language list so it wouldn't be available to users, ever. This pull request fixes it.

@Artoria2e5 You just forgot... 😉


Artoria2e5:
`><`

Artoria2e5 (approved):

app/assets/javascripts/components/containers/mastodon.jsx
```diff
@@ -89,6 +90,7 @@ addLocaleData([
   ...uk,
   ...zh,
```
Artoria2e5: Is this `zh` doing anything here, though?

Artoria2e5: I mean, it's like nobody is gonna make a `zh` locale and have a label for that...

Gargron (approved):

PR #2044, opened 2017-04-17
Add simplified Chinese js translation

Artoria2e5:
This single-file commit, heavily based on the zh-hk version as of
ec9999c, fixes the missing js locale. Intentional, significant
departures from current zh-cn yml localizations have been marked with
rationales given.

PR #2242, opened 2017-04-20
Initial overhaul at zh-CN YML translations

Artoria2e5:
This commit provides a heavy proofreading of zh-CN translations, and
mainly draws from the zh-HK source when a translation is found too
incoherent. Translating directly from en is used when zh-HK translation
is not available.

This commit also completes the tweet-toot (choo-choo, actually) for
zh-cn introduced in #2044. Some minor copyediting, as well as
additional en translations, has been fed back into zh-hk text.

### DIFF
--- a/config/locales/zh-CN.yml
+++ b/config/locales/zh-CN.yml
@@ -1,50 +1,177 @@
 ---
 zh-CN:
   about:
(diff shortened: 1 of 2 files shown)


Artoria2e5:
… shouldn't have tapped that.

Gargron (approved):

PR #2245, opened 2017-04-21
Admin UI for confirming users

ashfurrow:
This PR exposes accounts' confirms statuses in the admin list page, and adds a button to the edit page to confirm them. I'm still pretty new at Rails, so please let me know if there are more idiomatic approaches to some of my decisions.

screen shot 2017-04-20 at 8 05 35 pm
screen shot 2017-04-20 at 8 05 41 pm

Fixes https://github.com/tootsuite/mastodon/issues/1297

### DIFF
--- a/app/controllers/admin/confirms_controller.rb
+++ b/app/controllers/admin/confirms_controller.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Admin
+  class ConfirmsController < BaseController
+    before_action :set_account
+
+    def create
+      @account.user.update(confirmed_at: Time.now.utc)
+      redirect_to admin_accounts_path
+    end
+
+    private
+
+    def set_account
+      @account = Account.find(params[:account_id])
+    end
+  end
+end
--- a/app/views/admin/accounts/show.html.haml
+++ b/app/views/admin/accounts/show.html.haml
@@ -77,6 +77,9 @@
   - else
     = link_to t('admin.accounts.silence'), admin_account_silence_path(@account.id), method: :post, class: 'button'
 
+  - unless @account.user.confirmed?
+    = link_to t('admin.accounts.confirm'), admin_account_confirm_path(@account.id), method: :post, class: 'button'
+
   - if @account.suspended?
     = link_to t('admin.accounts.undo_suspension'), admin_account_suspension_path(@account.id), method: :delete, class: 'button'
   - else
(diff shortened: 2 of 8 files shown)


app/controllers/admin/confirms_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Admin
+  class ConfirmsController < BaseController
```
mjankowski: This is a little nitpicky, but you asked for rails feedback...

I would probably put this action either at:

- `Admin::UsersController#update` - with a route like /admin/users/123 - and make sure the only thing which can be updated is the confirmed_at; or
- `Admin::Confirmations#create` with a route like /admin/users/123/confirmation - and have that find that user and confirm it.

app/controllers/admin/confirms_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    before_action :set_account
+
+    def create
+      @account.user.update(confirmed_at: Time.now.utc)
```
mjankowski: I believe devise provides a convenience method like `user.confirm` which does essentially what you are doing, with slightly more logic around it; so I might use that.

app/controllers/admin/confirms_controller.rb
ashfurrow: Okay, I _think_ I've got the second option done. Let me know what you think 👍

config/routes.rb
```diff
@@ -87,6 +87,7 @@
       resource :reset, only: [:create]
       resource :silence, only: [:create, :destroy]
       resource :suspension, only: [:create, :destroy]
+      post '/confirmation', to:'confirmations#create'
```
ykzts: `resource :confirmation, to: [:create]` ?

ashfurrow: Cool, thanks 👍

Gargron (approved):

app/views/admin/accounts/show.html.haml
```diff
@@ -77,6 +77,9 @@
   - else
     = link_to t('admin.accounts.silence'), admin_account_silence_path(@account.id), method: :post, class: 'button'
 
+  - unless @account.user.confirmed?
```
nightpool: This will break for non-local accounts, leading to a 500 error.

ashfurrow: Good catch, I'll submit a fix. Thanks!

Issue #1900, opened 2017-04-16
Clear/hide specific notification


karelz:
Clearing all notifications is awesome, sometimes you want to keep a few / clear (or hide) those you already reacted to (e.g. you already replied).

Maybe special icon per notification?
Another option is checkbox next to each notification, but it feels heavy-weight.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

magicalskye:
this is important for me too, when i get replies that cause me bad anxiety i can't really look at it without feeling bad, but it doesn't justify silencing the whole person, so i avoid looking at mastodon altogether because that thing is still there

Cassolotl:
I've seen folks post things like "I've got an awful reply in my mentions - can you reply to me with cat pictures or whatever to bump it down the list?" So I think there's a clear need for this feature.

Cassolotl:
For those who specifically find unpleasant mentions a problem, this issue might be worth thumbs-upping and subscribing to and commenting on:

Hide a specific status #1669

trwnh:
This is available in the API for a while, just not in the web app

mrgan:
I agree that the ability to hide/"archive" a reply—or a single status, as in #1669—would be great.

I used to use a Twitter client for a company account, to handle support requests etc. Since this was a multi-user app, and the point was to "handle" tweets and then file them away as closed, the app had an "archive" feature. My daily flow was to read through what came in, answer those that needed an answer, and archive the rest—much like email.

Occasionally, abusive or needlessly argumentative posts would come in. With those that truly crossed the line, muting or blocking the account was an option. But with those that were merely angry, redundant, or e.g. unfunny, archiving took it off my mind. There's an empowering feeling in getting yet another boneheaded reply and simply hiding it, without necessarily ending all communication with that person forever (which block/mute would do).

jose-dvm:
I would love to see this implemented, now I don't have the better experience when I'm involved in different conversations.

Does someone knows if this is already done for mobile app?

I think a swipe gesture triggering a confirmation dialog would be OK, I'm not sure about accesibility though.

Issue #686, opened 2017-03-26
option to hide toots that you have already replied to in notifications


hoodiek:
i don't quite know how to put it simpler than the title, but i would like to state that it isnt hard to implement on the code side. a simple toggle to hide any posts that have a reply-child in the thread means that you can double check you replied to everything during a busy period, or essentially keep the column smaller to feel less overwhelmed, in case having many notifications is difficult for you

hoodiek:
another user mentioned they would prefer the ability to manually dismiss individual notifications in the notifs column

LogicalDash:
Third option: replying to a notification moves it to the bottom of the column.

hoodiek:
the third option does not make any sense to me, LogicalDash: it just clutters your notifications by reordering them in a unintuitive order, whilst not making it clear you're done with them. i would not advise this option be taken at all.

PR #2251, opened 2017-04-21
Added API for single notification dismissal

expenses:
Adds a `/api/v1/notifications/dismiss` POST route that takes an `id` parameter and removes the notification with that id.

This is a start to solving #686 and #1900.

### DIFF
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -30,6 +30,10 @@ class Api::V1::NotificationsController < ApiController
     render_empty
   end
 
+  def dismiss
+    Notification.where(account: current_account, id: params[:id]).delete_all
+  end
+
   private
 
   def exclude_types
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/notifications_controller.rb
```diff
@@ -30,6 +30,11 @@ def clear
     render_empty
   end
 
+  def dismiss
+    Notification.where(account: current_account, id: params[:id]).delete_all
```
Gargron: Maybe `Notification.find_by!(account: current_account, id: params[:id]).destroy!` would be better here (throw 404 when bad ID, throw error if something wrong with deleting it)

Gargron (changes requested):

app/controllers/api/v1/notifications_controller.rb
expenses: And done!

Gargron (approved):

Gargron:
There is actually no reason why this shouldn't be `DELETE /api/v1/notifications/:id`, how did I not catch that...

PR #2255, opened 2017-04-21
Fix #2195 - Set locale to error pages

### LINKED ISSUE #2195: 404 error is not localized
[`config/locales/*.yml`](https://github.com/tootsuite/mastodon/blob/v1.2.2/config/locales/en.yml#L205) has `errors.404`:

```yaml
  errors:
    '404': The page you were looking for doesn't exist.
    '410': The page you were looking for doesn't exist anymore.
    '422':
      content: Security verification failed. Are you blocking cookies?
      title: Security verification failed
```

and there is a translation for it, e.g. [ja.yml](https://github.com/tootsuite/mastodon/blob/v1.2.2/config/locales/ja.yml#L205):

```yaml
  errors:
    '404': お探しのページは見つかりませんでした。
    '410': お探しのページはもう存在しません。
    '422':
      content: セキュリティ認証に失敗しました。Cookieをブロックしていませんか？
      title: セキュリティ認証に失敗
```
However, I haven't see this translation anywhere. Accessing to non-existent URLs always show an error message in English.

* * * *

takp:
Fix #2195 

(e.g.) errors/404 page with correct locale
image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -51,23 +51,29 @@ class ApplicationController < ActionController::Base
   protected
 
   def not_found
-    respond_to do |format|
-      format.any  { head 404 }
-      format.html { render 'errors/404', layout: 'error', status: 404 }
+    set_locale do
+      respond_to do |format|
+        format.any  { head 404 }
+        format.html { render 'errors/404', layout: 'error', status: 404 }
+      end
     end
   end
 
   def gone
-    respond_to do |format|
-      format.any  { head 410 }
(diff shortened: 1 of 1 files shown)


mjankowski:
Good changes, thanks!

These methods are basically all the same -- What do you think about wrapping them up into something like:

```
def respond_with_error(code)
 # ...
render "errors/#{code}", layout: :error, status: code
end
```

and then have each of the other methods call that one with the appropriate code?

Gargron (approved):

PR #2256, opened 2017-04-21
rubocop_todo.yml

ReyADayer:
create rubocop_todo.yml

```
bundle exec rubocop --auto-gen-config
```


mjankowski:
I understand what this is and how you made it and why someone would use it ... but I'm not sure this makes sense to insert now...?

If someone pushes changes which have violations, codeclimate flags that, and the person can fix them. If they push changes which expose what were already pre-existing violations, they can choose to fix them as part of PR, or we can decide to ignore them.

If someone wanted to fix rubocop stuff specifically, they can do that one PR at a time (I'd focus on just one improvement at a time, don't try to fix the whole thing in one massive commit/PR), and we can fold in over time.

What's the rationale here? -- do you have a followup plan to work on fixing some of the issues? If so ... maybe keep the `_todo` in your branch and use it to make progress and generate smaller PRs?

PR #2257, opened 2017-04-21
Align the back buttons to the right

173210:
The back buttons on the column headers are aligned to the right, so enforce the alignment also for the simple back buttons.

This change is preferred in the sense that the the back button will be far from the follow/unfollow button in the status page.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1158,7 +1158,8 @@ a.status__content__spoiler-link {
   position: relative;
 }
 
-@media screen and (min-width: 360px) {
+// min-width corresponds to signle-column + nav layout breakpoint
+@media screen and (min-width: 1024px) {
   .columns-area {
     padding: 10px;
   }
@@ -1211,7 +1212,8 @@ a.status__content__spoiler-link {
   overflow: hidden;
 }
 
-@media screen and (min-width: 360px) {
+// min-width corresponds to signle-column + nav layout breakpoint
+@media screen and (min-width: 1024px) {
   .tabs-bar {
     margin: 10px;
     margin-bottom: 0;


walf443:
Can you add two screenshots before change and after change? It's more easy to imagine changes.

akihikodaki:
Before:
Before

After
After

yiskah:
Would this only occur if your locale is set to a language which writes right-to-left? It's more intuitive to speakers of left-to-right languages to have "back" be on the left.

akihikodaki:
> Would this only occur if your locale is set to a language which writes right-to-left?

No, my locale is Japanese, which is written left-to-right.

> It's more intuitive to speakers of left-to-right languages to have "back" be on the left.

Actually it is, and `<` icon looks suggesting that. However, the header could also have a text, which is placed at the left side in the current implementation. Putting back button at the left side conflicts with that.

akihikodaki:
Rebased to cf845fed3824d3e3587ce9b2ad752c2b3f0a2a76.

ashfurrow:
I'm not sure I agree with this change – it looks very strange to me to have the back button on the right side, even if it is that way on the Federated timeline etc. Maybe that ought to be rethought instead? Up for discussion of course.

akihikodaki:
Here I summarize important factors for design decisions of column headers:

Column headers may have a button and an optional title. The button could be _back_ or _settings_ button. The width of column is 330px.

Each of them is trivial, but it is difficult to consider of all of them. Here are my concerns:

Let decide the back button to put on the left side and the title on the right side. Should we put the settings button on the left side? Probably it is not desirable, but is it preferred to alter the position of the title depending on the button?

Another answer is to put the title at the middle. I think it is common for smartphone/PDA applications, but 330px may not be sufficient. (Consider the combination of _Federated timeline_ and _< Back_.)

Learning from smartphone/PDA applications, we may have just _<_ icon, without text, for the back button, which has enabled them to put the title at the middle. But, is that really OK?

ashfurrow:
Those are valid questions, thanks for thinking this through. I think we need input from someone with a clearer picture of Mastodon's vision.

akihikodaki:
Refer to a previous comment for headers which only have back buttons.
https://github.com/tootsuite/mastodon/pull/2257#issuecomment-296355207

_Federated timeline_ and a back button:
_Federated timeline_ and a back button

_Notifications_ and buttons:
_Notifications_ and  buttons

akihikodaki:
38f63692bbb2e847d418ac147167643bf1ef9b0b: rebased to 2fba94b36eac65a368ec10f4c8c5808baa0179e7.

beatrix-bitrot:
I have no issues with the code here, but I'm not sure whether or not users will appreciate this change, so I've solicited some feedback here (though not sure anyone will answer!) https://glitch.social/@bea/289316

Gargron:
Design has changed somewhat since. Closing as outdated.

PR #2271, opened 2017-04-21
Hide some components rather than unmounting

akihikodaki:
Hide some components rather than unmounting them to allow to show again quickly and keep the view state such as the scrolled offset.

### DIFF
--- a/app/assets/javascripts/components/components/status_list.jsx
+++ b/app/assets/javascripts/components/components/status_list.jsx
@@ -60,7 +60,7 @@ class StatusList extends React.PureComponent {
   }
 
   render () {
-    const { statusIds, onScrollToBottom, trackScroll, isLoading, isUnread, hasMore, prepend, emptyMessage } = this.props;
+    const { statusIds, onScrollToBottom, scrollKey, shouldUpdateScroll, isLoading, isUnread, hasMore, prepend, emptyMessage } = this.props;
 
     let loadMore       = '';
     let scrollableArea = '';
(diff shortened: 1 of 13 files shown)


Gargron:
I'm trying this out. The route switching is definitely less laggy and the scroll positions are preserved when needed. But there is a visual glitch where you can see all IconButtons for a split second on top of the new route. Do you notice that?

akihikodaki:
I confirmed the bug. I will fix it.

akihikodaki:
7a282ff7327eac22ae84cc2624d4c5f0c3db553f fixed the issue by minimalizing CSS transitions.

akihikodaki:
6a33030eca14a76315335def9059b80624a4db37 makes `Container` `PureComponent`.

akihikodaki:
0db468fefd596165ebc45e48b0c9172429b0f256 fixes window height of the desktop layout.

ineffyble (dismissed):
Are you able to make changes to fix the eslint issues raised by the Travis build?

173210:
64156729a0246596aba9967a027f805c9071b7fd fixed the ESLint issue.

akihikodaki:
64156729a0246596aba9967a027f805c9071b7fd fixed the ESLint issue.

akihikodaki:
6d3cd715e4ceaf8b914663c651ece36e86081dfa is rebased to 72c984e1057306d1e4df49871b9fb658fd7cbcc6.

Gargron (approved):

nolanlawson:
Quick question: did anyone check if there's a memory cost to this PR? It seems [hiding rather than unmounting would keep more memory around](https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/death_unmounting_indepth.html), which may have an impact on low-memory devices. If not, I'm happy to look into this and test.

akihikodaki:
It should cost memory, but I'm not really concerned about that since the memory usage is still less than 50MB.

nolanlawson:
How are you calculating memory? I just checked mastodon.social in Chrome on Windows (64bit) and in Chrome's Task Manager I see 70MB memory being taken up by the Mastodon tab, although a Memory snapshot in DevTools shows only 21MB (note that Chrome's DevTools Memory analysis only shows JavaScript heap, not native heap). In Firefox I'm seeing two processes taking up about 220MB total and in Edge I'm seeing 106MB using the Windows Task Manager.

akihikodaki:
I have estimated with Firefox developer tools. It is insufficient if it records only JavaScript heap. Thanks for pointing out that.

akihikodaki:
I couldn't have found the increase of memory usage with Chromium task manager, but it is odd; this change should have cost memory anyway.

PR #2275, opened 2017-04-21
Add Japanese translation for display name and bio counter

ykzts:
Japanese translation for #2219 .

### DIFF
--- a/config/locales/simple_form.ja.yml
+++ b/config/locales/simple_form.ja.yml
@@ -4,7 +4,6 @@ ja:
     hints:
       defaults:
         avatar: 2MBまでのPNGやGIF、JPGが利用可能です。120x120pxまで縮小されます。
-        display_name: 名前は30文字まで設定することができます。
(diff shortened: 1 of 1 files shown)


Gargron (approved):

config/locales/simple_form.ja.yml
```diff
@@ -5,9 +5,10 @@ ja:
       defaults:
         avatar: 2MBまでのPNGやGIF、JPGが利用可能です。120x120pxまで縮小されます。
         display_name: 名前は30文字まで設定することができます。
```
Gargron: This line should probably be removed?

Gargron (changes requested):

config/locales/simple_form.ja.yml
ykzts: @Gargron oops... sorry. fix in 2243911

Gargron (approved):

PR #2278, opened 2017-04-21
Improve bio compatibility

Gargron:
Move HTML-formatted bio from `<poco:note />` to `<summary type="html" />`

### DIFF
--- a/app/lib/atom_serializer.rb
+++ b/app/lib/atom_serializer.rb
@@ -2,6 +2,7 @@
 
 class AtomSerializer
   include RoutingHelper
+  include ActionView::Helpers::SanitizeHelper
 
   class << self
     def render(element)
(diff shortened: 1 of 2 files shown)


app/lib/atom_serializer.rb
```diff
@@ -21,13 +22,13 @@ def author(account)
...
     append_element(author, 'email', account.local? ? account.local_username_and_domain : account.acct)
-    append_element(author, 'summary', account.note)
+    append_element(author, 'summary', Formatter.instance.simplified_format(account).to_str, type: :html) if account.note?
     append_element(author, 'link', nil, rel: :alternate, type: 'text/html', href: TagManager.instance.url_for(account))
     append_element(author, 'link', nil, rel: :avatar, type: account.avatar_content_type, 'media:width': 120, 'media:height': 120, href: full_asset_url(account.avatar.url(:original)))
     append_element(author, 'link', nil, rel: :header, type: account.header_content_type, 'media:width': 700, 'media:height': 335, href: full_asset_url(account.header.url(:original)))
     append_element(author, 'poco:preferredUsername', account.username)
     append_element(author, 'poco:displayName', account.display_name) if account.display_name?
-    append_element(author, 'poco:note', Formatter.instance.simplified_format(account).to_str) if account.note?
+    append_element(author, 'poco:note', account.local? ? account.note : strip_tags(account.note)) if account.note?
```
mjankowski: You could probably skip the ternary and just do `strip_tags(account.note)`, even on local accounts.

If markup wound up there somehow, we'd want to strip, and it won't hurt where there is none.

mjankowski (approved):

app/lib/atom_serializer.rb
Gargron: I figured strip_tags must cost some performance.

PR #2283, opened 2017-04-21
email black/whitelist match on domain name end only

cquest:
a whitelist on gouv.fr accepts emails like me@gouv.fr.mydomain.com where is logically shouldn't
same for blacklists...


mjankowski (changes requested):
Can you add a spec to the coverage for this showing which cases previously had issues but work now?

ykzts:
duplicated? https://github.com/tootsuite/mastodon/pull/2213

cquest:
fixed by https://github.com/tootsuite/mastodon/commit/7177e37b999d0a8b4e0382c193bcb973443a756f

PR #2287, opened 2017-04-21
Add option to remove uploaded avatar/header image

### LINKED ISSUE #1315: Add option to remove uploaded avatar/header image
Add an option to remove your uploaded avatar/header image around here or somewhere better:

removeah

Also, maybe add an administrative option somewhere to remove a user's avatar/header if it's offensive/illegal/instance rule breaking?

happycoloredbanana:
Fixes #1315

### DIFF
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -24,7 +24,7 @@ class Settings::ProfilesController < ApplicationController
   private
 
   def account_params
-    params.require(:account).permit(:display_name, :note, :avatar, :header, :locked)
+    params.require(:account).permit(:display_name, :note, :avatar, :delete_avatar, :header, :delete_header, :locked)
   end
 
   def set_account
--- a/app/models/concerns/account_avatar.rb
+++ b/app/models/concerns/account_avatar.rb
@@ -18,6 +18,8 @@ module AccountAvatar
     has_attached_file :avatar, styles: ->(f) { avatar_styles(f) }, convert_options: { all: '-quality 80 -strip' }
     validates_attachment_content_type :avatar, content_type: IMAGE_MIME_TYPES
     validates_attachment_size :avatar, less_than: 2.megabytes
+    attr_accessor :delete_avatar
+    before_validation :destroy_avatar_if_requested
 
     def avatar_original_url
       avatar.url(:original)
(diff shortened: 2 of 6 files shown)


expenses:
What does the UI for this look like?

happycoloredbanana:
edit_profile_-_mastodon

ashfurrow:
Cool!

expenses:
I feel like it would be better to have a `delete` or `X` button that clears the avatar/header field and does the same thing instead.

happycoloredbanana:
@expenses I'd like to understand it better. When you say "clears the avatar/header field" do you mean that the file previously selected with the "Browse" button goes back to the "No file selected" state? Or do you mean that we need a button that sends POST that deletes the current avatar?

In the first case, what will be the visual indication that the user ever clicked the "Delete" button? Because "No file selected" is not enough.

A third possibility could be when rendering this form to somehow show whether avatar/header currently exists. Maybe show the current images.

joyeusenoelle:
Opinion: The form *should* display the current header/avatar, perhaps separately from this PR. Currently there's no feedback after the user clicks "Save Changes", and users have to visit their own profile pages to see if an updated header/avatar has taken.

With that implemented, this change could then be a **Remove** or **Delete** checkbox adjacent to the current image that would remove that image and reset to the default.

I suggest a checkbox rather than a button for UX consistency; that action shouldn't be the only one on the page that's effected without clicking the "Save Changes" button.

beatrix-bitrot:
After thinking about it I agree that the current avatar and header should be displayed, but I think that's ok to go in another issue/PR, and the same with changing the way this feature works after the images are added.

Once the merge conflicts with this are resolved I think I'd be fine with it and would approve, though I would like another reviewer to look as well and would tag someone in for that.

happycoloredbanana:
Solved merge conflicts.

I think the ideal solution (and not for this PR) for this would be:
- displaying thumbnails of the avatar and the header on the "edit profile" form
- having "remove" check boxes for the avatar and the header
- some javascript to uncheck the boxes whenever a file is selected, and vice versa - checking the "remove" box must clean the "selected file" field
- probably, some visual hint about the avatar/header is about to be removed - e.g., the avatar's thumbnails turning to the default once the "remove" checkbox is on


A question: do you guys know whether there's another web form in the Mastodon's settings that uses javascript? Just wanted to see how it's done in the current codebase as an example. (Of course, not talking about the React stuff, just plain web forms like we are having here.)

beatrix-bitrot (approved):
looks good to me (:

PR #2288, opened 2017-04-21

modeverv:
I've removed avatar files by mistake and I want to recover.


modeverv:
fix problem and I will try again...

PR #2290, opened 2017-04-21

modeverv:
I've removed avatar files by mistake and I want to recover.

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -33,6 +33,32 @@ namespace :mastodon do
     end
   end
 
+  desc 'Refetch missing avatar'
+  task refetch_avatar: :environment do
+    require 'fileutils'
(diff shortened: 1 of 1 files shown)


modeverv:
try again..

lib/tasks/mastodon.rake
```diff
@@ -33,6 +33,34 @@ namespace :mastodon do
     end
   end
 
+  desc 'Refetch missing avatar'
+  task refetch_avatar: :environment do
```
Gargron: This entire task should be

```ruby
Account.remote.find_each do |account|
  account.avatar = URI.parse(account.avatar_remote_url) unless account.avatar.exists? || account.avatar_remote_url.blank?
  account.header = URI.parse(account.header_remote_url) unless account.header.exists? || account.header_remote_url.blank?
end
```

PR #2292, opened 2017-04-21

modeverv:
I've removed avator files by mistake and I want to recover.

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -33,56 +33,19 @@ namespace :mastodon do
     end
   end
 
-  module AvatarHeader
-    require 'fileutils'
-
-    def get_loop(field, method)
-      Account.remote.find_each do |account|
-        send(method, field, account)
-      end
-    end
-
-    def re_get(field, account)
-      return unless account.send(field + '_file_name').present? && account.send(field + '_remote_url').present?
-      return if File.exist?(account.send(field).path)
-      kind = 'refetch'
-      puts "#{kind} #{field} of #{account.username}"
-      mkdir_p(File.dirname(account.send(field).path))
-      open(account.send(field).path, 'wb') do |out|
-        open(account.send(field + '_remote_url')) do |data|
-          out.write(data.read)
-        end
-      end
-    rescue => ex
-      _raise_(field, account, ex, kind)
-    end
-
-    def re_save(field, account)
-      return unless account.send(field).exists? && account.send(field + '_remote_url').present?
-      kind = 'refresh'
-      puts "#{kind} #{field} of #{account.username}"
-      account.send(field + '=', URI.parse(account.send(field + '_remote_url')))
-      account.save
-    rescue => ex
-      _raise_(field, account, ex, kind)
-    end
-
-    def _raise_(field, account, ex, kind)
-      # need to think about "404"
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -33,6 +33,31 @@ namespace :mastodon do
     end
   end
 
+  task refetch_avatar: :environment do
+    require 'fileutils'
+    all = Account.all
+    all.each do |account|
+      next unless account.avatar_remote_url.present?
```
expenses: What about `next if account.avatar-remote_url.blank?`?

modeverv: Thanks. I will fix code.

modeverv:
I fixed code please re-review!

modeverv:
Sorry for poor github skill...
Gargron say to me 

> +  desc 'Refetch missing avatar'
> +  task refetch_avatar: :environment do
>     This entire task should be
>     Account.remote.find_each do |account|
>      account.avatar = URI.parse(account.avatar_remote_url) unless account.avatar.exists? || 
>     account.avatar_remote_url.blank?
>       account.header = URI.parse(account.header_remote_url) unless account.header.exists? || 
>     account.header_remote_url.blank?
>     end

so I will re-think my code.

modeverv:
I've not think about "header"(lack of code reading..)

modeverv:
I fixed code,please review!!

modeverv:
I now think we need 2 tasks.
1.task for refetch missing avatar and header
2.task for refresh all avatar and header

bradurani:
:+1: As an admin, I have needed this functionality after deleting my S3 bucket. I think this code could be DRYed up a bit though

vaygr:
Yeah, user avatars' refreshing from remote instances would be quite handy.

lib/tasks/mastodon.rake
```diff
@@ -33,6 +33,58 @@ namespace :mastodon do
     end
   end
 
+  module AvatarHeader
```
Gargron: I'm afraid this is really not idiomatic Ruby code and it is doing *a lot* of things not the right way, it will also not work with S3 or any other cloud storage backends due to trying to write to local files. In the example I have given you, I demonstrated how to do it using Paperclip in a really simple way

Gargron (changes requested):

lib/tasks/mastodon.rake
modeverv: - not idiomatic 
-> I want isolate model and rake.
- not work with S3
-> I don't use cloud storage
-> fix to your simple way

modeverv:
I've change code. please re-review.

lib/tasks/mastodon.rake
```diff
@@ -36,7 +36,7 @@ namespace :mastodon do
   desc 'Refetch missing avatar and header'
   task refetch_avatar_header: :environment do
     Account.remote.find_each do |account|
```
carlosjs23: The batch size for find_each is a default of 1000, it would be a good idea to limit the batch size as it is too high and causes performance problems according to my tests. i.e.
`Account.remote.find_each(batch_size: 8) do ...`

carlosjs23 (commented):
@Gargron what do you think about this?

PR #2294, opened 2017-04-22
Allowed the 'Search...' string of the emoji picker to be localized (#2263)

### LINKED ISSUE #2263: Be able to translate Emojione
This will probably not be possible, because I can't find any translations at https://github.com/Ranks/emojione, but I think it's necessary that at least the category descriptions, the initial hover text (insert emoji) and 'Search...' of Emojione can be translated. The shortcodes are also in English, but this will probably break a lot of stuff, so that is probably out of the question.

expenses:
Fixes #2263 by adding in a localization string for the `searchPlaceholder` of the [emojione-picker](https://github.com/tommoor/emojione-picker/blob/master/src/picker.jsx#L23) class


Gargron (approved):

expenses:
@Gargron I've also had a go at allowing the emoji categories to be translated as well (having trouble with `injectIntl`ing into objects). Should I make a different PR for that?

expenses:
Nevermind, created #2302.

PR #2297, opened 2017-04-22
[WIP] Html lang on statuses

mjankowski:
Changes to:

- Add html `lang` attribute in stream entries around status
- Update the language detector to ignore urls in strings when detecting language ... should fallback to account or instance default in those scenarios.

### DIFF
--- a/app/lib/language_detector.rb
+++ b/app/lib/language_detector.rb
@@ -9,10 +9,18 @@ class LanguageDetector
   end
 
   def to_iso_s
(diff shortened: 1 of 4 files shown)


Gargron (approved):

Gargron:
Oh whoops this was named WIP but it looked complete to me?

mjankowski:
It was fine to merge ... I was considering asking about more edge cases in language detection, and adding a spec to show that statuses with URLs fall back to a supplied account language (not just instance default), but that's fine - I can do those as followup.

Issue #2209, opened 2017-04-20
Content warning and normal text locations and placeholders are confusing to users


Natouille:
As a user, and as seen and heard by other users, the behavior of the content warning field is really confusing.
**When a user adds a CW, a field pops above the main field, and the user expects the content of this field to be the content that will be hidden.**

And it is logical : think of the human mind. 

- First you warn _("I'm gonna show you / tell you something scary / illegal / bad / nonsense / whatever")_ 
- and then you say / show it.

So it should be the same on the Mastodon UI : 

- first you type your warning _("beware")_ in the main field
- then you click the CW button
- the CW field appears **(best if BELOW the main field)** and you add your sensitive content

feth:
A precision.
The placeholder «Avertissement» (french version) may confuse users. I'd suggest a longer but imperative «Saisissez ici des mots-clefs d'avertissement» (Here, input the warning keywords).

Cassolotl:
I usually do it the other way round. I type something, then I'm like "hey, I should CW this" and I click the CW box, which appears and is empty. So it works for the way my mind works. I hope this option sticks around - but if I'm in the minority that's fine and I'm sure I'd get used to it if it changed!

Natouille:
@feth on my instance, it's written in English (blame the admin ? 😄 )

feth:
![image](https://cloud.githubusercontent.com/assets/554384/25229915/810436d6-25d2-11e7-9660-bdd6cbaa848a.png)

![maxresdefault](https://cloud.githubusercontent.com/assets/554384/25229952/a70ce936-25d2-11e7-92d2-9c41cfae7b28.jpg)

Natouille:
![mastodon_-_parti_pirate_-_2017-04-20_14 08 07](https://cloud.githubusercontent.com/assets/2006947/25229989/cb97f2aa-25d2-11e7-8030-52d624106d86.png)

You were saying ? 🤣

wxcafe:
We didn't find a good translation for Content Warning that was widely understood and well-known in french yet. Please feel free to do a PR on the translation file if you do. That being said, I don't think the Content Warning *text* should be what you had written in the toot box before, cause the new text field specifically says "Content Warning" and is smaller.

averageunicorn:
For me the way it is now is more intuitive (toot goes where it always goes, CW is a separate thing), but I expected the CW box to appear _below_ the toot box, not above. Took me a few seconds to find it the first time.

[...]

Issue #1807, opened 2017-04-14
Accessible CW Dropdown


jpdevries:
The CW dropdown seems to only be accessible to sighted mouse users. Suggesting, somehow, we add:
 - [ ] keyboard support
 - [ ] ARIA for accessibility tree and screen reader support

<details>
<summary>Screenshot</summary>
<img src="http://j4p.us/0Z1R0N2a0H1y/Screen%20Shot%202017-04-14%20at%209.55.20%20PM.png" alt="Screenshot showing the CW dropdown when expanded along with the HTML markup, which consists of inaccessible divs with no ARIA" />
</details>
<br>

I'm not sure precisely the best pattern to use for this. Perhaps along the lines of `aria-controls`, `aria-expanded`, usage of `role`, and then adding all the keyboard / arrow keys support one would expect in a `<select>`, because that is basically what we are mimicking?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

wavebeem:
Maybe this should be in a different ticket, but I also noticed it's quite difficult to tell if the CW button is actually *on* or not. I made a mockup of what the button could look like when it's on.

clear:
![image](https://cloud.githubusercontent.com/assets/176815/25061304/46760044-2168-11e7-9574-751d6e9d736a.png)

unclear:
![screen shot 2017-04-14 at 23 13 41](https://cloud.githubusercontent.com/assets/176815/25061313/7096f27a-2168-11e7-93bc-6f8f86616448.png)

Happy to break this out into its own ticket to potentially get this done faster.

LJWatson:
The "CW" button seems to work with the keyboard. It's based on a ```<button>``` element, so the browser provides the expected keyboard interaction based on the click handler.

It does have problems though:
* There is nothing to indicate to screen reader users that activating the button will cause content to appear;
* The content that appears is inserted into the DOM *before the button, making it difficult for screen reader users to locate it;
* The `aria-label` attribute provides a completely different label for the button, making it difficult for screen reader users to recognise the "CW" button when it's referenced by other people/help files etc.

Suggest using a basic disclosure pattern, inserting the revealed content into the DOM `immediately` after the button, and changing the label.
```
<button aria-label="Content Warning (CW)" aria-expanded="true" aria-controls="this">CW</button>
...
<div id="this">
// Where the revealed content goes
</div>
```

[...]

PR #2303, opened 2017-04-22
Add aria-controls and aria-expanded to CW button

stephenburgess8:
~Content warning field appearing above toot reflows the column and is far away from the CW button which triggers the reveal of the field. Move content warning field below the toot box to connect it more directly to the trigger button through proximity.~

~img width="295" alt="cw-below-toot-box" src="https://cloud.githubusercontent.com/assets/6003351/25301091/2dd8ef2a-26e4-11e7-8955-cd3b27386a3e.png"~

#2209
#1807

~NOTE: I removed 'SensitiveButtonContainer' because it was not the button that was triggering the CW input field. I could not identify any way in which that button was being used so I removed the button entirely.~

Update:

Do not remove SensitiveButton, do not move CW field below toot box. What remains of original case:

* Add aria expanded to text icon button, making the CW button show as expanded.
* Add the aria-controls information as well to help JAWS screen reader users jump to the CW input field.

### DIFF
--- a/app/assets/javascripts/components/features/compose/components/compose_form.jsx
+++ b/app/assets/javascripts/components/features/compose/components/compose_form.jsx
@@ -146,7 +146,7 @@ class ComposeForm extends React.PureComponent {
       <div className='compose-form'>
         <Collapsable isVisible={this.props.spoiler} fullHeight={50}>
           <div className="spoiler-input">
-            <input placeholder={intl.formatMessage(messages.spoiler_placeholder)} value={this.props.spoiler_text} onChange={this.handleChangeSpoilerText} onKeyDown={this.handleKeyDown} type="text" className="spoiler-input__input" />
+            <input placeholder={intl.formatMessage(messages.spoiler_placeholder)} value={this.props.spoiler_text} onChange={this.handleChangeSpoilerText} onKeyDown={this.handleKeyDown} type="text" className="spoiler-input__input"  id='cw-spoiler-input'/>
           </div>
         </Collapsable>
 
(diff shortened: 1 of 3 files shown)


yiskah (commented):
I am against this. The warning will appear above the post. It is more intuitive UI to have the text appear where it will appear when posted.

stephenburgess8:
@yiskah Thanks for the feedback. In the issue I listed 2209, moving the input field below the post was proposed. This was implementing that change. However if ultimately that change is not preferred by everyone, we could keep it at the top but still include the accessibility improvements. I'll leave that up to the consensus.

Gargron:
I agree with Shel here, it's more intuitive at the top. But feel free to keep the accessibility improvements :+1:

yiskah (changes requested):
Hi, I love your accessibility fixes. If you just remove the part where you move the content warning box then this is good to merge

STJrInuyasha:
> NOTE: I removed 'SensitiveButtonContainer' because it was not the button that was triggering the CW input field. I could not identify any way in which that button was being used so I removed the button entirely.

The SensitiveButtonContainer is the button that shows up when you add media, to mark it NSFW. Removing it is, well, not a very good idea.

stephenburgess8:
Started fresh, blew away the old changes in favor of a single commit. Updated the description. Remaining changes are:

* Add aria expanded to text icon button
* Add the aria-controls attribute to CW button

Thank for help getting through this one. I'm enjoying getting to know the code base.

Gargron (approved):

PR #2305, opened 2017-04-22
Translate "sensitive" as "要注意" for Japanese

akihikodaki:
The previous translation for "sensitive", "不適切" means "impertinent". This change suggests "刺激的", meaning inciteful, instead.

This is potentially controversial. Other translators' opinions will be appreciated.


lindwurm (changes requested):
I think "sensitive" is not always "刺激的", I'd like to propose "要注意".

app/assets/javascripts/components/locales/ja.jsx
```diff
@@ -123,7 +123,7 @@ const ja = {
   "status.replyAll": "全員に返信",
   "status.report": "通報",
   "status.sensitive_toggle": "クリックして表示",
-  "status.sensitive_warning": "不適切なコンテンツ",
+  "status.sensitive_warning": "要注意",
```
lindwurm: I suggest "要注意コンテンツ"

akihikodaki:
4a5d10a6de2c4f6c6d77e0735c998230381c3d5a adopts _要注意_ as a translation for _sensitive_.
Please tell us if you think _不適切_, _刺激的_, or any other word is more appropriate.

app/assets/javascripts/components/locales/ja.jsx
lindwurm: I suggest "要注意コンテンツ".

lindwurm (changes requested):

lindwurm:
thanks @akihikodaki , you have to change here too https://github.com/tootsuite/mastodon/blob/master/config/locales/ja.yml#L291

app/assets/javascripts/components/locales/ja.jsx
akihikodaki: _要注意コンテンツ_ sounds a bit strange. I wonder also whether _注意を要するコンテンツ_ is too verbose or too long to fit in the UI.

akihikodaki:
f43beb2126192b294b3cd88c6304bd09c7e2728b has _要注意コンテンツ_ as a translation, and translates also `config/locales/ja.yml`.

cannorin:
"閲覧注意" is a natural translation of "NSFW". since most Japanese people do not distinguish between "NSFW" and "sensitive content", the same translation is preferable.

"要注意" sounds more like "suspicious" or "marked"; for example, "要注意人物" means "person of interest".

cannorin:
fyi "コンテンツ" is not a precise translation of "content"; "コンテンツ" often means "image" or "video" in the context of Japanese microblogging. please consider replacing it with "内容".

akihikodaki:
> "閲覧注意" is a natural translation of "NSFW".

That makes sense, but CW has the same translation. Possibly we could also change the translation for CW? I think _警告_ is better for CW.

> fyi "コンテンツ" is not a precise translation of "content"; "コンテンツ" often means "image" or "video" itself in the context of Japanese microblogging. please consider replacing it with "内容".

If I recall correctly, NSFW is for image or video.

yumetodo:
Who on the earth can understand that ``閲覧注意`` is placeholder to warn for all other users?

In some case(except firefox), it is not clear that the string is actually a placeholder
(watch fig.1, fig.2, fig.3. Only Firefox has possibility that user can understand).

fig1

fig.1 example of Microsoft Edge.

fig2

fig.2 example of Google Chrome

image

fig.3 example of Mozilla Firefox 

So, we need to clarify that in text message(should not depend on css).

In my opinion, ``閲覧注意`` should be replaced with some text that many users can understand whether it is placeholder, like ``警告文``.

akihikodaki:
I don't think it is so ambiguous that users couldn't notice that it is an input area. `警告文` is too verbose then.
Also, it is probably should be done in the original text at first, not in the translation.

yumetodo:
>it is probably should be done in the original text at first, not in the translation.

I agree. ``Content warning`` should also be replaced, I think.

akihikodaki:
@yumetodo You may open another issue then. Could you make also clarify your opinion for the translation? Is _警告_ is OK as a translation for _content warning_?

cannorin:
> CW has the same translation.

I think it is good to share the same term among those translations (e.g. "閲覧注意", "閲覧に注意が必要な内容を含んでいます", "閲覧注意に設定する") .

> If I recall correctly, NSFW is for image or video.

"NSFW" indicates that *the content of* an e-mail message, a forum post, a video, a photo, or anything, is not suitable to be seen at the workplace.

what I want to say is that the word "コンテンツ" lost its original meaning (e.g. "コンテンツを配信" ).

cannorin:
> CW has the same translation.

I think it is good to share the same term among those translations (e.g. "注意書きを入力", "閲覧に注意が必要な内容を含んでいます", "閲覧注意に設定する") .

> If I recall correctly, NSFW is for image or video.

"NSFW" indicates that *the content of* an e-mail message, a forum post, a video, a photo, or anything, is not suitable to be seen at the workplace.

what I want to say is that the word "コンテンツ" lost its original meaning (e.g. "コンテンツを配信" ).

akihikodaki:
> I think it is good to share the same term among those translations (e.g. "注意書きを入力", "閲覧に注意が必要な内容を含んでいます", "閲覧注意に設定する") .

I mean, translations for different texts may have better to have different translations.

> "NSFW" indicates that the content of an e-mail message, a forum post, a video, a photo, or anything, is not suitable to be seen at the workplace.

Again, I suppose NSFW in Mastodon is for picture or video, and not for the message. That is why I consider _コンテンツ_ is reasonable.

yumetodo:
> Is 警告 is OK as a translation for content warning?

``警告`` is a best translation, I think.

>"NSFW" indicates that the content of an e-mail message, a forum post, a video, a photo, or anything, is not suitable to be seen at the workplace.

>what I want to say is that the word "コンテンツ" lost its original meaning (e.g. "コンテンツを配信" ).

NSFW is only for picture or video.

MaverickTse:
None of the proposed wording seems to fit that field IMO.
The purpose of that placeholder should be made the user to aware that what they write in there will goes public. The content is not necessarily "sensitive". 
Hence, the placeholder should be:
English: "(Public Content)"
日本語：（公開的な内容）

yumetodo:
@MaverickTse As pointed out in #2305 (comment) by @akihikodaki, this problem should not resolve at translate text. => #2310.

cannorin:
> NSFW in Mastodon is for picture or video

> NSFW is only for picture or video.

in Mastodon, yes. 
I wanted to say that "sensitive content" in Mastodon and "NSFW" in general is the same thing for Japanese people, and "閲覧注意" is a natural translation for the general "NSFW".
sorry for confusing you.

I understand why you prefer using "コンテンツ" for "sensitive content". but, at least, using "コンテンツ警告" as a translation of "content warning" seems wrong.

akihikodaki:
2fee6fa579c8517e572d56e76bd2ad9e2474d1b0: _CW_ -> _警告_, _NSFW_ -> _閲覧注意_

akihikodaki:
985f48f98bfdad69b246f36c22f4891d0dc8b649: applied a change suggested in https://github.com/tootsuite/mastodon/commit/2fee6fa579c8517e572d56e76bd2ad9e2474d1b0#commitcomment-21867050.

Gargron:
Posts on reddit can be marked "nsfw". If reddit has a Japanese localization, can I ask how "nsfw" is translated there? Perhaps we could use that.

Content warning could also be "spoiler warning". I don't know if those are perhaps more common concepts in Japanese.

Gargron:
I think that from an icon perspective, "NSFW" and "CW" should stay verbatim because they are so short. Kind of like "R18" is understandable even though those are not Hiragana/Katakana/Kanji characters, they should be perceived more like icons and not like translatable text.

yumetodo:
>I think that from an icon perspective, "NSFW" and "CW" should stay verbatim because they are so short. Kind of like "R18" is understandable even though those are not Hiragana/Katakana/Kanji characters, they should be perceived more like icons and not like translatable text.

Wait, wait, wait ,wait.
Nobody are discussing about replacing icon ``CW``/``NSFW``, I think.

image

this image is exmaple of https://github.com/tootsuite/mastodon/commit/985f48f98bfdad69b246f36c22f4891d0dc8b649

akihikodaki:
> Posts on reddit can be marked "nsfw". If reddit has a Japanese localization, can I ask how "nsfw" is translated there? Perhaps we could use that.

Unfortunately Reddit doesn't have translation for NSFW.

> Content warning could also be "spoiler warning". I don't know if those are perhaps more common concepts in Japanese.

Anyway, whether content warning should be "spoiler warning" or not, CW should have a translation distinct from NSFW to describe the meaning of warning. I suggest _警告_ for that.

> I think that from an icon perspective, "NSFW" and "CW" should stay verbatim because they are so short. Kind of like "R18" is understandable even though those are not Hiragana/Katakana/Kanji characters, they should be perceived more like icons and not like translatable text.

That makes sense.

Gargron (approved):

PR #2306, opened 2017-04-22
Add link for domain in land strip

ymmtmdk:
I want to go to "https://mastodon.social" from "https://mastodon.social/@Gargron".
Don't you?

### DIFF
--- a/app/views/shared/_landing_strip.html.haml
+++ b/app/views/shared/_landing_strip.html.haml
@@ -2,5 +2,5 @@
   = t('landing_strip_html',
     name: content_tag(:span, display_name(account), class: :emojify),
     domain: site_hostname,
-    root_path: root_path,
+    link_to_root_path: link_to(content_tag(:strong, site_hostname), root_path),
     sign_up_path: new_user_registration_path)
--- a/app/views/shared/_landing_strip.html.haml
+++ b/app/views/shared/_landing_strip.html.haml
@@ -2,4 +2,5 @@
   = t('landing_strip_html',
     name: content_tag(:span, display_name(account), class: :emojify),
     domain: site_hostname,
+    root_path: root_path,
     sign_up_path: new_user_registration_path)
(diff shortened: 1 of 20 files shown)


walf443:
`domain` variable is not url, but domain for example 'example.com'. so you should covert it to url.

ymmtmdk:
Oops. I will fix it (if I can).
Thank you.

ymmtmdk:
Fixed the problem. Could you review this please?

app/views/shared/_landing_strip.html.haml
```diff
@@ -2,4 +2,5 @@
   = t('landing_strip_html',
     name: content_tag(:span, display_name(account), class: :emojify),
     domain: site_hostname,
+    root_path: root_path,
```
mjankowski: Since this value: `<a href="%{root_path}"><strong>%{domain}</strong></a>` is the exact same in every translation, would it be possible to prepare this from the partial and send it a value that's already an html link (and do the right thing re: html_safe) so that the partial is something like:

    link_to_root_path: link_to(...)

and then the i18n like:

    landing_strip_html: '.... %{link_to_root_path} ...'

?

ymmtmdk: Sorry if I may misunderstand your intention.
I think your intetion is:

`link_to_root_path: link_to(content_tag(:strong, site_hostname), root_path),`

`landing_strip_html: <strong>%{name}</strong> is a user on %{link_to_root_path}. You can follow them or interact with them if you have an account anywhere in the fediverse. If you don't, you can <a href="%{sign_up_path}">sign up here</a>.`

Right?

mjankowski: Yeah, something like that -- essentially, try to avoid repeating the same large block of code within the translations themselves.

ymmtmdk: Fixed it. Thank you for your help!

mjankowski (approved):
Thanks for cleaning up, looks good.

Issue #2188, opened 2017-04-20
Specify redis DB in settings to allow sharing of redis with other processes


dasoran:
I try to run two mastodon instances on one postgres and one redis.
When I open timeline page, it seems good.
A short time later, the other instances post come this instances timeline.
When I reload the page, timeline is collect again.

I find that there is not config of redis-db(e.g.db0, db1,...) in 'streaming/index.js'.
I think there is other problems to use redis db settings.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

dasoran:
developed by #2869

PR #2307, opened 2017-04-22
Specify redis DB in settings to allow sharing of redis with other pro…

dasoran:
This implement is what I said on issue #2188.


app/workers/push_update_worker.rb
```diff
@@ -8,7 +8,9 @@ def perform(account_id, status_id)
     status  = Status.find(status_id)
     message = InlineRenderer.render(status, account, 'api/v1/statuses/show')
 
-    Redis.current.publish("timeline:#{account.id}", Oj.dump(event: :update, payload: message, queued_at: (Time.now.to_f * 1000.0).to_i))
+    prefix = ENV.fetch('REDIS_PUBSUB_PREFIX') { '' }
```
ashfurrow: Since this code is repeated a few times, what do you think about putting it either in a helper function or as a frozen constant somewhere?

app/services/fan_out_on_write_service.rb
```diff
@@ -57,16 +57,20 @@ def render_anonymous_payload(status)
   def deliver_to_hashtags(status)
     Rails.logger.debug "Delivering status #{status.id} to hashtags"
 
+    prefix = ENV.fetch('REDIS_PUBSUB_PREFIX') { '' }
+
     status.tags.pluck(:name).each do |hashtag|
-      Redis.current.publish("timeline:hashtag:#{hashtag}", @payload)
-      Redis.current.publish("timeline:hashtag:#{hashtag}:local", @payload) if status.local?
+      Redis.current.publish("#{prefix}timeline:hashtag:#{hashtag}", @payload)
+      Redis.current.publish("#{prefix}timeline:hashtag:#{hashtag}:local", @payload) if status.local?
```
bradurani: Should we wrap `Redis.current` with a wrapper class if we add prefixes? It seems like there's enough repeated logic here to warrant a separate class specifically for doing Redis operations rather than calling `Redis.current` directly everywhere

dasoran:
I'm sorry I dirty commit logs. I didn't find I have to make working branch.

I try to create class to proceed redis pub/sub.

ashfurrow:
Nice! Looks a lot cleaner!

dasoran:
I made a mistake in branch operation, so I close it

PR #2309, opened 2017-04-22
Localize with i18n for Devise::FailureApp

alpaca-tc:
Currently I18n.locale is changed inside the controller. Therefore, I18n.locale in
Devise::FailureApp always uses the default locale because it is called from outside of the controller.

side effect:

- Controller
  - If the controller doesn't include Localized, I18n.locale remains from previous request.
- Worker
  - No side effects because Redis does not store current locale.
- Mailer
  - No side effects because it uses `I18n.with_locale` every time.

### DIFF
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -4,25 +4,16 @@ module Localized
   extend ActiveSupport::Concern
 
   included do
-    around_action :set_locale
+    before_action :set_locale
   end
 
   private
 
   def set_locale
-    locale = default_locale
-
-    if user_signed_in?
-      begin
-        locale = current_user.try(:locale) || default_locale
-      rescue I18n::InvalidLocale
-        locale = default_locale
-      end
-    end
-
-    I18n.with_locale(locale) do
-      yield
-    end
+    I18n.locale = default_locale
+    I18n.locale = current_user.locale if user_signed_in?
+  rescue I18n::InvalidLocale
+    I18n.locale = default_locale
   end
 
   def default_locale
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -115,8 +115,7 @@ class ApplicationController < ActionController::Base
   end
 
   def respond_with_error(code)
-    set_locale do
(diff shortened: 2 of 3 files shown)


mjankowski (commented):
Can you explain more thoroughly what the failure is, and why the change in the controller fixes something failing outside of it?

Gargron:
I don't understand exactly what this is fixing. Also, I think all Devise controllers inherit from ApplicationController (somehow), since localizations for login and registration controllers worked.

alpaca-tc:
This PR fixes I18n.locale for rake middlewares. Mastodon uses Devise that depends on Warden. Those middleware can be found in `rake middleware`.

In the case of authentication failed, warden calls `throw(:warden)`. At the time Middleware of Warden delegates request to failure_app to generate error response after catching it. Unfortunately, I18n.locale is already reset then because I18n.with_locale is enabled in only the controller.

Related code:

- https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/activatable.rb#L8
- https://github.com/hassox/warden/blob/master/lib/warden/manager.rb#L136
- https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb#L95

alpaca-tc:
Change DEFAULT_LOCALE and try to sign in with unconfirmed email :smile:

Gargron (approved):

mjankowski:
Thanks for the explanation.

It would be good to see a failing spec that is fixed by this change, so that future refactors don't accidentally revert this (it's pretty subtle!)

app/controllers/concerns/localized.rb
```diff
@@ -4,25 +4,16 @@ module Localized
...
-        locale = current_user.try(:locale) || default_locale
-      rescue I18n::InvalidLocale
-        locale = default_locale
-      end
-    end
-
-    I18n.with_locale(locale) do
-      yield
-    end
+    I18n.locale = default_locale
```
alpaca-tc: It seems that `user_signed_in?` sometimes `throw(:warden)`, so I set default_locale at the first.

alpaca-tc:
I added spec 😄  💛

mjankowski (approved):

PR #2322, opened 2017-04-22
Alternative Paperclip config for OpenStack Swift

thurloat:
This has been in production at mastodon.club for about a week. Working like a charm.

It has only been tested with Cloud-A's swift deployment, though there's nothing about it that is not OpenStack Swift API compliant -- it works with any other tools that work with swift.

Example configuration file
```
SWIFT_ENABLED=true
SWIFT_USERNAME=<openstack_username>
SWIFT_TENANT=<openstack_tenant_id>
SWIFT_PASSWORD=<openstack_password>
SWIFT_AUTH_URL=<openstack_keystone_url>
SWIFT_CONTAINER=<swift_container_name>
SWIFT_OBJECT_URL=<object_server_url> # shortcuts the service catalog lookup & enables CDN where available
```


ykzts:
@tackeyy Can I use OpenStack Swift when `S3_SIGNATURE_VERSION=s3`?

Gargron:
I'm really really not sure how to go about this. There are *tons* of cloud storage providers, do we want to include all of their drivers and configurations for everybody?

Or would this be better as a documented way to modify code if you need it?

thurloat:
I figured it was worth sharing upstream because it's not provider specific, rather implementing the generic swift API spec, and is the only other supported paperclip backend (fog).

The config setup will work whether you're using ovh, rackspace, cloud-a, vexxhost, private-cloud-X, self-hosted swift, etc.

Gargron:
Hi, I'm sorry this has been ignored for a long time. As I have heard more people ask whether Mastodon works with OpenSwift, I have changed my mind about whether this is worth including upstream.

If you could resolve the conflicts, I will merge this.

Gargron (approved):

ykzts:
@thurloat I resolved conflicts. Please can you check this difference?

mayaeh:
Hello.
ObjectStrage EndPoint of my service does not include containers url and it was necessary to change it as follows.
https://github.com/mayaeh/mastodon/commit/1bda3f02e94964307d31bc3f4849f95c98ea943e

Also, I think that it would be more kind to add definitions to sample settings.
https://github.com/mayaeh/mastodon/commit/03dced9053c9ff155970da38b9d211609360c07a

thurloat:
@ykzts LGTM, thanks for fixing it up.

@mayaeh the object URL doesn't necessarily include the container name, especially if you're using a CDN (like I am) it could just be https://static.mastodon.social, depending on how you have it configured. It would be up to the deployer of an individual mastodon instance to know what the fully rendered path of their objects will be to use fog.

I like your idea of including the sample settings, and perhaps to have extra guidance it would be wise to include a note about object URLs frequently being the swift url + container name.

PR #2333, opened 2017-04-22
column & account menu Width Fix

YaQ00:
Account dropdown menu 140px => 25vw, 50vw, 90vw (responsive)

Width => drawer 20% (max-width 400px), column 80% / 3 (max-width 800px)

Width 2560px over => height 100vh, margin 0

Narrow column, smallest menu => Wide & modern column!


treby:
@YaQ00 Maybe you should provide screenshot and why this patch is reasonable :)

PR #2337, opened 2017-04-23
OEmbed support for PreviewCard

Gargron:
Oh boy, this is quite a whammy!

* I had to make an improved version of `OEmbed::ProviderDiscovery` because the original was using *regex* to get *link tags from HTML* and it broke if they were in the wrong order or there was a, say, `rel="alternate"` in the wrong place
* Expand `PreviewCard` schema to save the variety of attributes you can get from OpenGraph and OEmbed
* Use sanitize gem instead of Rails sanitize helper
* Try to get OEmbed from a URL, if that's no good, try OpenGraph
* Fix our own OEmbed API not working with new type of prettier URLs
* `<Card />` component can now display links, photos and videos:

image

image

image

image


expenses:
Great! Does it Oembed linked toots?

ineffyble:
Code Climate is not happy - do we care?

app/lib/provider_discovery.rb
```diff
@@ -0,0 +1,38 @@
...
+      raise OEmbed::NotFound, url if res.code != 200 || res.mime_type != 'text/html'
+
+      html = Nokogiri::HTML(res.to_s)
+
+      if format.nil? || format == :json
+        provider_endpoint = html.at_xpath('//link[@type="application/json+oembed"]')['href']
+        format ||= :json
+      end
+
+      if format.nil? || format == :xml
```
nightpool: The first conditional here will never get called, since if format is nil, we set it to :json on line 17.

nightpool: I'm not sure this series of conditionals is doing what you expect it to do. Are you trying to do something like this? currently it uses JSON, or XML, or JSON if no format is specified.

```ruby
def provider_endpoint(format)
  case format
  when :json
    html.at_xpath('//link[@type="application/json+oembed"]')['href']
  when :xml
    html.at_xpath('//link[@type="application/json+oembed"]')['href']
  else
    return provider_endpoint(:json) || provider_endpoint(:xml)
  end
end
```

Gargron:
For the time being, "rich" OEmbed types get ignored, because stuff like tweets tries to include `<script>` tags and looks generally shitty without them

app/lib/provider_discovery.rb
Gargron: Fixed

Gargron:
CodeClimate is not happy on this one but the formatting it *wants* me to do is unreadable so I say, ignore it.

app/controllers/api/oembed_controller.rb
```diff
@@ -14,8 +14,9 @@ def show
   def stream_entry_from_url(url)
     params = Rails.application.routes.recognize_path(url)
 
-    raise ActiveRecord::RecordNotFound unless params[:controller] == 'stream_entries' && params[:action] == 'show'
+    raise ActiveRecord::RecordNotFound unless %w(stream_entries statuses).include?(params[:controller]) && params[:action] == 'show'
```
mjankowski: Maybe extract that conditional into a method?

app/controllers/api/oembed_controller.rb
```diff
@@ -14,8 +14,9 @@ def show
...
 
-    StreamEntry.find(params[:id])
+    return StreamEntry.find(params[:id]) if params[:controller] == 'stream_entries'
```
mjankowski: These two lines could also be a separate method...

```ruby
def stream_entry
  if params[:controller] == 'stream_entries'
    StreamEntry.find(params[:id])
  else
    Status.find(params[:id]).stream_entry
  end
end
```

app/lib/provider_discovery.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class ProviderDiscovery < OEmbed::ProviderDiscovery
+  USER_AGENT = "#{HTTP::Request::USER_AGENT} (Mastodon/#{Mastodon::VERSION}; +http://#{Rails.configuration.x.local_domain}/)"
```
mjankowski: Is this string already assembled somewhere else?

Gargron: It would be great if we could have some sort of `HTTP` factory because headers and timeout settings have to be set from quite a few places, but I'm not sure how to do that idiomatically. `HttpFactory` doesn't sound like Ruby...

app/controllers/api/oembed_controller.rb
```diff
@@ -14,8 +14,20 @@ def show
   def stream_entry_from_url(url)
     params = Rails.application.routes.recognize_path(url)
```
mjankowski: I don't think this change needs to hold back merging this ... but it might be good to use a better local variable name here, since this is happening inside of a controller, and `params` is already a method (and used up in the `show` action...)

mjankowski (approved):

jeroenpraat:
Is there a list of currently working websites? I tried e.g. Soundcloud and Twitter, but that didn't yet work. Just a normal link preview.

PR #2349, opened 2017-04-23
Report processing improvements

ineffyble:
I suspect some of this will need refactoring, I'm not fully up to date with our best practices etc, but it all works.

Improvements for the report index and report view. Hopefully more to come.

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -11,4 +11,12 @@ class Report < ApplicationRecord
   def statuses
     Status.where(id: status_ids)
   end
+
+  def media_attachments
+    media_attachments = []
+    statuses.each do |s|
+      media_attachments.concat s.media_attachments
+    end
+    media_attachments
+  end
 end
--- a/app/views/admin/reports/index.html.haml
+++ b/app/views/admin/reports/index.html.haml
@@ -18,7 +18,7 @@
         %th= t('admin.reports.target')
         %th= t('admin.reports.reported_by')
         %th= t('admin.reports.comment.label')
-        %th= t('admin.reports.contents')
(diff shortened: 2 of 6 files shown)


app/models/report.rb
```diff
@@ -11,4 +11,12 @@ class Report < ApplicationRecord
   def statuses
     Status.where(id: status_ids)
   end
+
+  def media_attachments
```
mjankowski: We might be able to turn `statuses` into a `has_many`, and then this could just be `has_many :media_attachments, through: :statuses`, instead of doing the looping this way.

If we can't do that, this can probably be shortened to something like `statuses.collect(&:media_attachments).flatten`, or similar.

app/views/admin/reports/index.html.haml
```diff
@@ -13,20 +13,32 @@
   %table.table
     %thead
       %tr
-        %th
+        -# %th
```
mjankowski: Was this leftover from the work?

ineffyble:
It was leftover from half completed work of Gargron's, and currently shows
as a checkbox with no label or use. Commenting it out for now seemed
prudent.

app/views/admin/reports/index.html.haml
ineffyble: It was leftover from half completed work of Gargron's, and currently shows
as a checkbox with no label or use. Commenting it out for now seemed
prudent.

app/models/report.rb
ineffyble: Yeah, I knew there was a better way I wasn't quite thinking of.

Gargron (approved):

PR #2361, opened 2017-04-23
Filter on allowed user language preferences

mjankowski:
Related to https://github.com/tootsuite/mastodon/issues/691

This adds an `allowed_languages` array column to `User` which is meant to hold the iso codes for languages that user wants to see in their feed.

The prefs screen gets a list like this:

screen shot 2017-04-28 at 11 29 44 am

Once the options are selected, there's a new filter that gets added to the timelines.

The Home timeline is not changed at all ... it makes sense to me that if you have chosen to follow someone you see all of their statuses regardless of which language they are posting in. The local and federated timelines are both filtered using language, so that you only see statuses that have a language which in the list you have selected.

Feedback I'm looking for:

- Are there other places we need to filter this beyond just those few places in the timeline creation?
- I don't think there are any federation concerns here ... everything will still be sent and received as before, this is purely a filter on what a person sees in their timeline, not on what their instance receives. If account X on instance B creates only `es` statuses, and is followed by account Y on instance A, instance A will continue to receive all their statuses. Account Y will see those statuses in their home timeline (and in others, if they have `es` turned on), but if no other users on instance A are allowing `es` into their feed, they just won't see those. If they turned `es` back on and refreshed their feed, they would immediately see what had previously been hidden.
- Performance concerns?

Also, this is not directly related to adding the filter, but will be relevant to how it actually works --
 statuses that get posted which are only a URL with no surrounding text will currently get created as either the locale selected by the person who created them, or as the instance default wherever they came from. This means that on the filtering side, users will only see URL-only statuses created by people who have selected a locale which is a language they allow through. This is *probably* ok, but I could be persuaded in either direction on it. We might also want to reconsider whether setting a locale at all on URL-only statuses makes sense? ... or we might want to just strip all URLs out of statuses when we choose their language?

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -326,3 +326,10 @@ code {
     flex: 0 0 auto;
   }
 }
+
+.user_allowed_languages {
+  li {
+    float: left;
+    width: 50%;
+  }
+}
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -25,7 +25,8 @@ class Settings::PreferencesController < ApplicationController
 
   def user_params
     params.require(:user).permit(
-      :locale
+      :locale,
+      allowed_languages: []
     )
   end
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -82,6 +82,10 @@ class Account < ApplicationRecord
     prefix: true,
     allow_nil: true
 
+  delegate :allowed_languages,
+    to: :user,
+    prefix: false
+
   def follow!(other_account)
     active_relationships.where(target_account: other_account).first_or_create!(target_account: other_account)
   end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -82,9 +82,7 @@ class Account < ApplicationRecord
     prefix: true,
     allow_nil: true
 
-  delegate :allowed_languages,
-    to: :user,
-    prefix: false
+  delegate :allowed_languages, to: :user, prefix: false, allow_nil: true
 
   def follow!(other_account)
     active_relationships.where(target_account: other_account).first_or_create!(target_account: other_account)
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -119,6 +119,10 @@ class Status < ApplicationRecord
   end
 
   class << self
+    def in_allowed_languages(account)
+      where(language: account.allowed_languages)
+    end
+
     def as_home_timeline(account)
       where(account: [account] + account.following)
     end
@@ -198,6 +202,7 @@ class Status < ApplicationRecord
 
     def filter_timeline_for_account(query, account)
       query = query.not_excluded_by_account(account)
+      query = query.in_allowed_languages(account.user)
       query.merge(account_silencing_filter(account))
     end
 
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -202,7 +202,7 @@ class Status < ApplicationRecord
 
     def filter_timeline_for_account(query, account)
       query = query.not_excluded_by_account(account)
-      query = query.in_allowed_languages(account.user)
+      query = query.in_allowed_languages(account) if account.allowed_languages.present?
       query.merge(account_silencing_filter(account))
     end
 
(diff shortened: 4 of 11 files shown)


ineffyble:
UI wise, I had a look at Google Search.

They give you a little summary, with a link:
image

Which expands out into a four column layout of tickboxes:
image

Something similar could work well.

ineffyble:
Would URL-only statuses be the same as media-only statuses? I feel like those ideally wouldn't have a locale.

mjankowski:
> Would URL-only statuses be the same as media-only statuses? I feel like those ideally wouldn't have a locale.

Yeah, because I think when you upload media all that's happening is the status is being populated with the URL of your media once it has one. So from the perspective of "what language is this?", those look the same (they are both just a URL).

Right now, there's logic where if you post *JUST* a URL, it will skip the language-guess entirely, and just default to the instance locale. If you posted a URL followed by a bunch of text, it would attempt to guess based on the whole thing (including the URL), which might be sort of weird. I could see a change like "remove all URL strings from text before we guess the language" being a good improvement, although sort of outside the scope of this filtering.

mjankowski:
I've updated the PR description to reflect current state of things, removed the WIP label, updated screenshot, etc.

Gargron (approved):

seascape:
Wonderful stuff.

I realize this complicates implementation but I just thought I'd note that personally, I'd like the ability to also filter the home timeline. I do not follow back many people who follow me because they primarily (but not exclusively) post in languages I don't understand. It would be nice to be able to follow them without making my home timeline significantly less comprehensible.

So there is a use case imo, at least for some users.

thisisTQ:
Mhh... I usually toot in two languages. I am not sure what  language detection will make of that...

wimvanderbauwhede:
Is there any data on the accuracy of the language detection? Ideally, % false positives and negatives? 
Also, could there be an "All" for allowed, if the user doesn't want to use this feature?

Cassolotl:
@wimvanderbauwhede I'm a bit late here but this might be helpful: https://github.com/tootsuite/mastodon/issues/2704#issuecomment-298899120

wimvanderbauwhede:
Thanks a lot for that, exactly what I was looking for!

MightyPork:
Is some special setup needed to get this working? Kinrar just flipped mastodon.xyz to 1.4.1 and the filters seem to do nothing at all

Gargron:
@MightyPork Remote toots carry their own detected language from origin, so you might see incorrectly detected toots from servers that haven't upgraded?

MightyPork:
Ah, so it will really work only after the majority updates, gotcha! (Also didn't you want to take a break? =P)

Vanege:
> The Home timeline is not changed at all ... it makes sense to me that if you have chosen to follow someone you see all of their statuses regardless of which language they are posting in.

That does not make sense to me. Why would everyone want to see messages in a language that one can not understand, even among the people one follows? This is a huge problem for me, since I follow a lot of people who toot in different languages. I don't want to force multilingual people to multiply their number of accounts, one for each language in each thematic-instance. Also, they would also have to switch account each time they switch language, which is impractical.

If I follow someone, that does not mean I can/want to read 100% of the things this person say, but more than 0%. If I speak a minority language (Esperanto in my case) and want to use this language, I may receive only 5% interesting toots in my Home timeline since Esperanto-users are almost all multilingual (the same can be said about Breton-French, Catalan-Spanish ...). 5% of interesting toots are immensly better than 0% by not following people. But yet, I would prefer if the majority of the messages in my Home timeline was interesting to me, without having to exclude people.

Cassolotl:
(I follow some people who toot multilingually.)

So like, when you follow someone, not everything they say is going to be of interest. I follow someone who is a server admin, and I enjoy their selfies and whatever, but if they start talking about databases I can't understand what they're saying so I just scroll past it all once I've skimmed it to make sure it's not interesting. If they started posting *only* server admin stuff I would probably just unfollow them.

It's the same with people posting in multiple languages. If they started posting only in German I would also unfollow them.

But on the odd occasions that I skim something in another language and it looks important, I put it into a translation thing. I'd like to be able to keep doing that.

So if the language filter *can* be applied to the home timeline, I would like it to be optional.

PR #2378, opened 2017-04-24
Add an RTL stylesheet for Arabic i18n

Gargron:
cc @BoFFire

image

The Arabic translation has a lot of missing strings, but at least it is now possible to display it properly right-to-left.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,4 +8,9 @@ module ApplicationHelper
   def show_landing_strip?
     !user_signed_in? && !single_user_mode?
   end
+
+  def add_rtl_body_class(other_classes)
+    other_classes = "#{other_classes} rtl" if [:ar].include?(I18n.locale)
+    other_classes
+  end
 end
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -8,4 +8,9 @@ def active_nav_class(path)
   def show_landing_strip?
     !user_signed_in? && !single_user_mode?
   end
+
+  def add_rtl_body_class(other_classes)
+    other_classes = "#{other_classes} rtl" if [:ar].include?(I18n.locale)
```
ykzts: https://github.com/tootsuite/mastodon/pull/2378/files#diff-2bb925571600311d122bd27feec93485R134 doesn't use?

Gargron: Oops, good catch, that was my first idea and it didn't work well

BoFFire:
Cc @Gargron the RTL stylesheet is perfect :). We will push the missing strings tonight GMT+1. Thank you for the update.

PR #2381, opened 2017-04-24
Account domain blocks

Gargron:
- [x] Model
- [x] Filter from public timeline queries
- [x] Filter from notifications
- [x] Filter from home feed
- [x] Filter from streaming API
- [x] Filter from "context" (conversation threads)
- [x] API method to block/unblock domain
- [x] API method to list blocked domains

Probably better done in another PR:

- [ ] Web UI dropdown option to block domain on profile
- [ ] Web UI list of blocked domains


app/models/account.rb
```diff
@@ -107,6 +107,10 @@ def blocking?(other_account)
     blocking.include?(other_account)
   end
 
+  def domain_blocking?(other_account)
+    AccountDomainBlock.where(account: self, domain: other_account.domain).exists?
```
mjankowski: Account could probably use a has_many for :account_domains_blocks, which this method could then use.

app/models/status.rb
```diff
@@ -188,6 +188,8 @@ def permitted_for(target_account, account)
 
     def filter_timeline(query, account)
       blocked = Rails.cache.fetch("exclude_account_ids_for:#{account.id}") { Block.where(account: account).pluck(:target_account_id) + Block.where(target_account: account).pluck(:account_id) + Mute.where(account: account).pluck(:target_account_id) }
+      domains = Rails.cache.fetch("exclude_domains_for:#{account_id}") { AccountDomainBlock.where(account: account).pluck(:domain) }
```
mjankowski: Instead of a local variable here, how about a (Rails cached) method like `Account#blocked_domains`? which the next line could call directly

db/schema.rb
```diff
@@ -202,6 +210,14 @@
     t.datetime "image_updated_at"
     t.datetime "created_at",                      null: false
     t.datetime "updated_at",                      null: false
+    t.integer  "type",               default: 0,  null: false
```
mjankowski: This schema change is likely from another PR.

spec/fabricators/account_domain_block_fabricator.rb
```diff
@@ -0,0 +1,4 @@
+Fabricator(:account_domain_block) do
+  account_id 1
```
mjankowski: ONE!

wxcafe:
Closing #501 since this is the same feature and that old PR is not up to date anymore (by a long shot)

Gargron:
I can't imagine very well how the API for domain blocks should actually look.

```
GET /api/v1/domain_blocks - returns list of { id: domainBlockId, domain: blockedDomainString }
POST /api/v1/domain_blocks - creates domain block for "domain" param
```

How to unblock?! Use a DELETE route using the ID of the domain block? Use domains instead of IDs?

app/models/status.rb
```diff
@@ -284,14 +284,21 @@ def carried_over_reply_to_account_id
   end
 
   def find_statuses_from_tree_path(ids, account)
-    statuses = Status.where(id: ids).to_a
+    statuses = Status.where(id: ids).includes(:account).to_a
+
+    # FIXME: n+1 bonanza
```
beatrix-bitrot: Are you gonna fix the N+1 bonanza?

app/controllers/api/v1/domain_blocks_controller.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class Api::V1::DomainBlocksController < ApiController
+  before_action -> { doorkeeper_authorize! :follow }
+  before_action :require_user!
+
+  respond_to :json
+
+  def show
```
beatrix-bitrot: curious why this is show and not index since it's giving a collection rather than an individual object

config/routes.rb
```diff
@@ -150,7 +150,8 @@
       resources :favourites, only: [:index]
       resources :reports,    only: [:index, :create]
 
-      resource :instance, only: [:show]
+      resource :instance,      only: [:show]
+      resource :domain_blocks, only: [:show, :create, :destroy]
```
beatrix-bitrot: again, wondering why show and not index

spec/models/account_domain_block_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe AccountDomainBlock, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
beatrix-bitrot: indeed! add some tests or delete it

beatrix-bitrot (changes requested):
overall looks good, only very minor stuff

app/models/status.rb
Gargron: No it's been present since day 1, it's a fix for another day with less time pressure

app/controllers/api/v1/domain_blocks_controller.rb
Gargron: Because it's a "resource" route and not "resources" (so that we can send DELETE /api/v1/domain_blocks)

beatrix-bitrot (approved):
concerns addressed

Issue #2375, opened 2017-04-23
Using WEB_DOMAIN wrong URL is returned in Webfinger for atom, salmon


kballard:
I set up an instance whose public domain is eridi.us and whose WEB_DOMAIN is social.eridi.us. Using webfinger.net to look up kevin@eridi.us, I get the following:

```json
{
  "subject": "acct:kevin@eridi.us",
  "aliases": [
    "https://social.eridi.us/@kevin",
    "https://social.eridi.us/users/kevin"
  ],
  "links": [
    {
      "rel": "http://webfinger.net/rel/profile-page",
      "type": "text/html",
      "href": "https://social.eridi.us/@kevin"
    },
    {
      "rel": "http://schemas.google.com/g/2010#updates-from",
      "type": "application/atom+xml",
      "href": "https://eridi.us/users/kevin.atom"
    },
    {
      "rel": "salmon",
      "href": "https://eridi.us/api/salmon/1"
    },
    {
      "rel": "magic-public-key",
      "href": "data:application/magic-public-key,RSA.wek_FwrfGUhUxrmdbksDqGW-nMZM21h3kVCw5dYhE5_ln5nGTIOxySEneAJWDTMhRwsOm_TBSKfUJfW0kdt3vgwcGhPpzYjo4gYKZG54zMZ91mjc5uOmQXSmRsss17SOBQu_SlM01Bq3oar06ln94pVGd3t-NfS6DbFUgrJIkH7tL1jupGNUVEP_UvLBoegFCHusUmi6VDEjd_IpNR3OPnHCiHAjUe-Fz7j4m_POQc16Om_pHJ_5LfvNKszgQLOPGuLXqJaZfj6yZ8VhnSp_ERmiyUg1EAlAK3I16KrXT9gyMjidV8nI-R8VyHoElv0IFs4lZIPFoth7RMbeSjapmw==.AQAB"
    },
    {
      "rel": "http://ostatus.org/schema/1.0/subscribe"
    }
  ]
}
```

Note how both the Atom URL and the Salmon URL are using eridi.us instead of social.eridi.us.

* * * *

This bug happens on current `master` (ef5937da1ff2d6caca244439dd9b9b9ed85fb278).

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

dariusk:
I'm seeing something similar to this but not quite? I just set up an instance on a recent master (1801a36414ee79bd51aad23fc5ff6a90049e7270) like so:

[...]

PR #2386, opened 2017-04-24
Add additional documentation and warnings to the WEB_DOMAIN setting.

ClearlyClaire:
This feature is largely undocumented, and quite a number of users have
shot them in the feet already despite the warning (e.g. #2375). Added a bit of documentation
and expanded the warning until we have a mechanism for dealing with conflicting
user URIs.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -13,12 +13,8 @@ LOCAL_DOMAIN=example.com
 LOCAL_HTTPS=true
 
 # Use this only if you need to run mastodon on a different domain than the one used for federation.
-# When used, LOCAL_DOMAIN should serve a /.well-known/host-meta directing webfinger requests to WEB_DOMAIN.
-# Do *not* change LOCAL_DOMAIN once WEB_DOMAIN is set or if WEB_DOMAIN is set to a previous LOCAL_DOMAIN value,
-# as that would create *different accounts* (@user@LOCAL_DOMAIN and @user@WEB_DOMAIN) on remote server
-# with conflicting identifiers (https://WEB_DOMAIN/users/user), and *will* break federation for
-# previously-defined users!
-# Do *not* use this unless you know *exactly* what you are doing.
+# You can read more about this option on https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Serving_a_different_domain.md
+# DO *NOT* USE THIS UNLESS YOU KNOW *EXACTLY* WHAT YOU ARE DOING.
 # WEB_DOMAIN=mastodon.example.com
 
 # Application secrets
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -13,7 +13,12 @@ LOCAL_DOMAIN=example.com
 LOCAL_HTTPS=true
 
 # Use this only if you need to run mastodon on a different domain than the one used for federation.
-# Do not use this unless you know exactly what you are doing.
+# When used, LOCAL_DOMAIN should serve a /.well-known/host-meta directing webfinger requests to WEB_DOMAIN.
+# Do *not* change LOCAL_DOMAIN once WEB_DOMAIN is set or if WEB_DOMAIN is set to a previous LOCAL_DOMAIN value,
+# as that would create *different accounts* (@user@LOCAL_DOMAIN and @user@WEB_DOMAIN) on remote server
+# with conflicting identifiers (https://WEB_DOMAIN/users/user), and *will* break federation for
+# previously-defined users!
+# Do *not* use this unless you know *exactly* what you are doing.
 # WEB_DOMAIN=mastodon.example.com
 
 # Application secrets


dariusk:
This looks great, thank you.

kballard:
This looks good, but I don't suppose there's a URL you could include that would have more info on what .well-known/host-meta is supposed to actually do?

mjankowski:
What about putting a thorough explanation of this into the docs repo, and linking to it (with a strong warning like you have) from this env file?

ThibG:
Yes, that would definitely be the way to go. I don't have the time to write such documentation right now, though.

kballard:
It looks right now like the simplest way to handle this is to add something like

```nginx
location /.well-known/host-meta {
    return 301 https://webdomain.example.com$request_uri;
}
```

to your nginx config.

dariusk:
I will make a PR on the docs repo today and I'll include Apache and nginx
configurations

On Apr 24, 2017 12:41 PM, "Kevin Ballard" <notifications@github.com> wrote:

It looks right now like the simplest way to handle this is to add something
like

location /.well-known/host-meta {
    return 301 https://webdomain.example.com$request_uri;
}

to your nginx config.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/tootsuite/mastodon/pull/2386#issuecomment-296800636>,
or mute the thread

ThibG:
@dariusk yes, see https://github.com/tootsuite/documentation/pull/149

dariusk:
@thibg great, I'll take a look!

ThibG:
I can update it with a doc as soon as https://github.com/tootsuite/documentation/pull/149 gets merged

ThibG:
Updated the comments to point to the recently-merged documentation.

Gargron (approved):

yukorabbit:
Hi, Is there any other way than modifying Ngnix to setup redirect to WEB_DOMAIN?
I don't have permission to modify Ngnix, and tried other methods.

**1. Modified .htaccess**
I added following code on the .htaccess in webroot of LOCAL_DOMAIN.
```
Redirect permanent /well-known/*$ https://social.example.com/.well-known/$1
order deny,allow
```
**2. Hosting "host-meta" file in ".well-know" folder on LOCAL_DOMAIN**
I downloaded host-meta from social.example.com and uploaded to example.com
image

**3. Setup Forwarding Rules on CloudFlare** 
[This page explains how it works](https://support.cloudflare.com/hc/en-us/articles/218411427#redirects). Basically, it's dynamic domain forwarding.
image

Actually, none of those method doesn't work for me for the moment.
I'm not specialized in programming, so probably I'm doing wrong.

ThibG:
Hi @yukorabbit. Your proposed setups all seem fine to me in theory (I'm not fluent in Apache rewriting rules and stuff, so I don't know if they are incorrect, but the idea seems fine).
Could you publish the actual domain name you are using for your `LOCAL_DOMAIN` so I could try to see what's wrong?

yukorabbit:
@ThibG Sure :D Thank you for your time!
My LOCAL_DOMAIN is @yukorabb.it and WEB_DOMAIN is social.yukorabb.it.

I have following issues.
**1. my account (yukorabbit@yukorabb.it) cannot be followed by other users from other instances**
- My test account on the other instance followed my account. I saw the number of my main account's followers increased.
- When I see from my main account, nothing happens. 

**2. I can't mention users outside my instance**
When I mention my test account, I get "500 Internal Server Error".

ThibG:
@yukorabbit not completely sure why you can't mention your user, I seem to have no issue in that regard (you should check you indeed got a mention from me, though). That being said, I have found at least one somewhat recently-introduced bug with `WEB_DOMAIN` that I will attempt to fix today.

yukorabbit:
@ThibG I checked my account, but I didn't receive any message. (and that's what's happened  when I mention myself from my test account on the other instance.)

I'm going to look into my server. I'll let you know what  is actually wrong.

yukorabbit:
@ThibG
I checked the system log and access log. I don't know how to do with the result, so I just going to write what I discovered

**LOCAL_DOMAIN**
 /.well-known/webfinger?resource=acct:yukorabbit@yukorabb.it was accessed when I mentioned @yuko@yukorabb.it with my test account. I got the following log.
```
"GET /.well-known/webfinger?resource=acct:yukorabbit@yukorabb.it HTTP/1.1" 404 476 "-" "http.rb/2.2.2"
```
When I mentioned the test account with my @yuko@yukorabb.it account, there's no access to ```/.well-known/webfinger?resource=acct:yukorabbit@yukorabb.it```

**WEB_DOMAIN**
I searched errors in syslog. I don't find any errors oddly. Probably I'm looking into a wrong place.
I use Ubuntu 16.04 and Docker. 
The Sever is Scaleway, where has restrict SMTP port policy.
They block outgoing for port 25, 587, 465 by default, until you send a selfie with your passport :(
So I use mailgun for SMTP with 2525 port. I wonder if those errors are related to blocked port.

ThibG:
@yukorabbit I was talking about Mastodon logs, not HTTP logs. You should check mastodon-web and mastodon-sidekiq

yukorabbit:
@ThibG Ah, I'm sorry! 😣
Here's the error logs that I got.
The beginnings (with timestamp from Oct 11th to 12th) have a lot of errors because I installed Mastodon. 

[Sidekiq.log](https://github.com/tootsuite/mastodon/files/1382593/Sidekiq.log)
[mastodon_web.log](https://github.com/tootsuite/mastodon/files/1382594/mastodon_web.log)

ThibG:
@yukorabbit you indeed have an issue with your setup unrelated to `WEB_DOMAIN`: `/mastodon/public/system/accounts` is not writeable by your puma (mastodon-web) process, which causes an error when processing accounts, and thus any communication. You should check your setup.

yukorabbit:
I've got a clean install from [Masto.host](https://Masto.host) and now it works!
Thank you very much for your time to look into my issue, @ThibG.

Lastly, I found a mistake on [the previous post](https://github.com/tootsuite/mastodon/pull/2386#issuecomment-336368131) about Cloud Flare settings.

Here's the fixed one.
I hope This would help who can't setup Nginx :D

**Forwarding with Cloudflare Page Rules**
If the URL matches: ```*web_domain.com/.well-known*```
Then the settings are: ```Forwarding URL```, ```301 - Permanent Redirect``` to ```https://local_domain.com/.well-known$2```

Issue #2129, opened 2017-04-19
log-in in single user mode


xsteadfastx:
im running my instance in single user mode. i know there is a url for loggging in but i think it would be better if there is a link on the default profile.

thanks for this amazing software and everything it gets going right now.... its exciting times...

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

mjankowski:
> i know there is a url for loggging in but i think it would be better if there is a link on the default profile.

...can you clarify your request here?

archaeme:
I think he means that it would be nice to have a convenient link to log in when single-user mode is set.

xsteadfastx:
On Wed, 19. Apr 04:30, Francis wrote:
> I think he means that it would be nice to have a convenient link to log in when
> single-user mode is set.

Exactly. I was logged in when i set my instance to single user mode...
tried to login on a different computer and had to find out the right url
first.

ashfurrow:
Now that https://github.com/tootsuite/mastodon/issues/2129 has been merged, this issue should be resolved. It'll be included in the next release of Mastodon. Thanks again @xsteadfastx for opening!

xsteadfastx:
❤️❤️❤️

Issue #917, opened 2017-04-05
No link to connect on single user mode


Korko:
Hi,

When you are disconnected on a single user mode instance, you get directly to the user profile but there's no link to connect. So you have to manually remember to go to /auth/sign_in to log in.

Thanks

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Exagone313:
Or just /about

ashfurrow:
Now that https://github.com/tootsuite/mastodon/issues/2129 has been merged, this issue should be resolved. It'll be included in the next release of Mastodon. Thanks again @Korko for opening!

PR #2393, opened 2017-04-24
Login link for single user instances

KScl:
Adds a log in link to the footer on single user instances, if not currently logged in. Addresses #917, #2129, possibly others


Gargron (approved):

Gargron:
Hmm this is going to take some tests fixing because of the `single_user_mode?` helper

Gargron (approved):

PR #2396, opened 2017-04-24
Create a script for easy updates on docker-compose

halest:
Suggesting to add an easy-upgrade script for docker-compose-installations.
Stashes local changes, pulls from git, restores changes, recreates images and containers and runs migrations.


docker-compose-mastodon-update.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+git stash
+git pull
```
ashfurrow: I'm wary of using `git pull` here because we'd like to encourage admins to run on tagged versions of mastodon, instead of whatever happens to be on master at the time they upgrade. What do you think of changing this to ask for a tagged version instead?

docker-compose-mastodon-update.sh
```diff
@@ -0,0 +1,8 @@
...
+git stash pop
+docker-compose build
+docker-compose up -d
+docker-compose run --rm web rails assets:precompile
+docker-compose run --rm web rails db:migrate
```
ashfurrow: I've always wondered: is it advisable to run `rails db:migrate` while the app is running? Or should we bring down the server first to let the migrations happen atomically?

halest: Thank you for commenting. the db:migrate task is taken like this straight drom the [docker update guide](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Docker-Guide.md) so I am assuming it safe. I have run this script for the past month via cron and did not have a problem with it.

docker-compose-mastodon-update.sh
halest: As commented below I have run it like this for the past month via cron and did not have a problem. But you are correct. Best practice would indeed be to fetch and checkout, but you won't be able to run it unattended.

ashfurrow: I see. Hmm. I'm not sure we'd want to merge this as-is. We recently [updated the upgrade instructions](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Docker-Guide.md#updating) to explicit tell admins to check a tag rather than the latest. Always running the latest is bold! But I'm not sure we should recommend that generally for admins.

jeanphilippeD: Would something along these lines work for unattended?
https://julienrenaux.fr/2013/10/04/how-to-automatically-checkout-the-latest-tag-of-a-git-repository/

halest: That is very interesting and in my tests seemed to work. Thank you for that info, I can really use that in a lot of my scripts. I was wondering if the docs had changed because when I wrote this I'm pretty sure it just said git pull. I'm gonna put this in. I'll leave my personal version as is because I like running latest though. Seems I'm a bold person ;)

docker-compose-mastodon-update.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+git stash
+git fetch --tags
+latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
+git checkout $latestTag
```
ashfurrow: This looks a lot better! One more thing: can you add some documentation in this script, or the documentation repo, that explains what this script is, what it does, how to use it, and the limitations? (i.e.: if someone has made commits on their fork, as I have, a `checkout` is going to remove them.) Looking great!

halest: Certainly. Documentation added.

Gargron:
Would prefer not to keep this code in the repository; different updates also sometimes require different commands, there is no silver bullet solution.

PR #2404, opened 2017-04-24
add Mathjax support in your toots to enjoy pretty mathematical formulae in toots

mokjpn:
Using 'mathjax-rails' gem, every mastodon toots can contain beautiful mathematical formulae inside.
This PR has some of small codes to include mathjax-rails gem and apply that only when you write #math hashtag in your toots. I used setInterval() javascript to make new #math toots rendered periodically. 

Example screenshot:
https://m.okadajp.org/system/media_attachments/files/000/000/065/original/ba49c476fce42601.png?1492705908

One of Japanese mastodon community enjoyed this feature so much. They even innovated many new smileys using mathematical signs and fonts. The ideas of this PR were given from @EzoeRyou@friends.nico , @h_okumura@mstdn.jp , and other many community members.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,8 @@ gem 'react-rails'
 gem 'browserify-rails'
 gem 'autoprefixer-rails'
 
+gem 'mathjax-rails'
+
 group :development, :test do
   gem 'rspec-rails'
   gem 'pry-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -232,6 +232,8 @@ GEM
       nokogiri (>= 1.5.9)
     mail (2.6.4)
       mime-types (>= 1.16, < 4)
+    mathjax-rails (2.6.1)
+      railties (>= 3.0)
     method_source (0.8.2)
     microformats2 (2.1.0)
       activesupport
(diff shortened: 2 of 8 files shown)


jemus42:
Love the idea, but slightly concerned about performance, especially on instances (or just long threads) with a bunch of math on them.  

Maybe this could be expanded to KaTeX, which can use server-side processing, which then could be a setting for instance admins to opt-in to?

badosu:
Mathjax is [shutting down it's CDN](https://www.mathjax.org/cdn-shutting-down/).

It would be good to check if this is covered before merging, if this PR gets approved. So far it seems that the last update to this gem was made on Mar/2016.

mokjpn:
I agree this implementation might affect performance on client due to setInterval().
I found another implementation that hooks MutationObserver. This would be nice but I do not tried on my instance yet.

http://qiita.com/DUxCA/items/766415eeb7b1cb330543

Gargron:
I am not sure if the main repository should include support for MathJax..? It's cool, but very specialized.

mokjpn:
I proposed this, because MathJax support is relatively easy to implement and allows users wide range of  text representation. But it have not attracted many developers in weeks. I would like to close this.

nikolay:
@Gargron Except that Math is the roots of each and every department of science.

PR #2405, opened 2017-04-24
Persian translation

mabkenar:
All strings described in the [translation documentation](https://github.com/tootsuite/documentation/blob/master/Contributing-to-Mastodon/Translating.md) are translated into Persian (fa, fa_IR, فارسی).

There is one intentional exception: the file "doorkeeper.fa.yml" is kept like the original English version. This is because it covers strings that are shown to the admin of the instance, and many of the words used in that file have no agreed-upon Persian translation. Translating them would indeed make it more difficult for admins to manage the instance. Of course in the future, when a consensus about those terms in Persian is reached, that file will also be translated.


Gargron:
Please add `:fa` to this check here: https://github.com/tootsuite/mastodon/blob/master/app/helpers/application_helper.rb#L13 So the site would be displayed as RTL

Gargron (approved):

config/locales/fa.yml
```diff
@@ -0,0 +1,324 @@
+---
+fa:
+  about:
+    about_mastodon: ماستدون (Mastodon) یک شبکهٔ اجتماعی <em>آزاد و کدباز</em> است. یک جایگزین <em>غیرمتمرکز</em> برای شبکه‌های تجاری، که نمی‌گذارد ارتباط‌های شما را یک شرکت در انحصار خود بگیرد. یک سرور مورد اعتماد را انتخاب کنید&mdash;هر سروری که باشد، همچنان می‌توانید با سرورهای دیگر ارتباط داشته باشید. هر کسی می‌تواند سرور ماستدون خود را راه بیندازد و در <em>شبکهٔ اجتماعی</em> سهیم شود.
+    about_this: دربارهٔ این سرور
+    apps: برنامه‌ها
+    business_email: 'ایمیل کاری:'
+    closed_registrations: امکان ثبت نام روی این سرور هم‌اینک فعال نیست.
+    contact: تماس
+    description_headline: %{domain} چیست؟
```
Gargron: Syntax error, needs to be surrounded with "

Gargron (changes requested):

mabkenar:
Hmmm... I don't know how to fix the failing CI test. It seems that the number of active locales it expects is one less than what currently exists (because of this new `fa` locale). But I can't find where else is the list of locales is defined that makes this mismatch. If you give me some hints, I would appreciate and I may be able to fix it. Or you could go ahead and fix it yourself, and that would also be fine with me.

jeanphilippeD:
Hello,
Looking at the test failure, it seems you also need to update 'config.i18n.available_locales' in this file with the new locale:
https://github.com/tootsuite/mastodon/blob/master/config/application.rb

Gargron (approved):

mkody:
Well, for me something went wrong with that merge. :grin: 
image

config/locales/simple_form.fa.yml
```diff
@@ -0,0 +1,51 @@
+---
+en:
```
mkody: I guess you forgot to change that to `fa:`

PR #2414, opened 2017-04-24
Add rspec to further specify FollowRemoteAccountService

ClearlyClaire:
This pull request is a follow-up to #2147 to further specify the intended behavior of FollowRemoteAccountService and make sure it prevents hijacking remote accounts using webfinger, while still following legitimate redirections.

To give a bit more background, webfinger queries for an “acct:” URI are not required to return a response with the same “acct:” URI as subject. This allows some form of redirection (user a@example.com is now called b@example.com, or has moved to a@example.org).

However, until #2147 was merged, Mastodon would take the returned subject “acct:” URI as the actual ID of the remote account, even if it didn't match the requested “acct:” URI, making it possible for attackers to create or overwrite a remote account for a domain they do not control.

### DIFF
--- a/spec/fixtures/requests/redirected.host-meta.txt
+++ b/spec/fixtures/requests/redirected.host-meta.txt
@@ -1,0 +1,15 @@
+HTTP/1.1 200 OK
+Server: nginx/1.6.2
+Date: Sun, 20 Mar 2016 11:11:00 GMT
+Content-Type: application/xrd+xml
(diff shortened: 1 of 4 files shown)


ashfurrow (commented):
Wow, nice work! I don't fully understand the protocols you're working with but your description of the problem this PR fixes is :100:

Gargron (approved):

PR #2416, opened 2017-04-24
WIP+RFC: Reduce color count and add more variables

Naouak:
The aim of this pull request is to reduce the number of colors used throughout the default css of mastodon (near a hundred colors!).

This is a work in progress but I'm opening the pull request now for comments.

- [x] Moving all colors used to variable.scss
- [x] Reduce color count
- [ ] Rename colors according to their usage(e.g.: `$link-color` or `$link-color-hover`)

I propose also that we change the name of `$color1`to `$color8` to relevant names while customisation is still not heavy on current mastodon instances.

### DIFF
--- a/app/assets/stylesheets/about.scss
+++ b/app/assets/stylesheets/about.scss
@@ -2,7 +2,7 @@
   .wrapper {
     max-width: 600px;
     margin: 0 auto;
-    color: $color3;
+    color: $about-color;
     padding-top: 50px;
     padding-bottom: 50px;
 
@@ -15,7 +15,7 @@
     font: 46px/52px 'Roboto', sans-serif;
     font-weight: 600;
     margin-bottom: 20px;
-    color: $color4;
+    color: $about-title1-color;
     padding: 20px 0;
 
     img {
@@ -32,7 +32,7 @@
     line-height: 28px;
     font-weight: 400;
     margin-bottom: 20px;
-    color: $color5;
+    color: $about-title2-color;
   }
 
   h3 {
@@ -41,7 +41,7 @@
     line-height: 28px;
     font-weight: 400;
     margin-bottom: 20px;
-    color: $lighten20_color3;
+    color: $about-title3-color;
   }
 
   ul, ol {
@@ -67,7 +67,7 @@
     margin-bottom: 12px;
 
     a {
-      color: $color4;
+      color: $about-link-color;
       text-decoration: underline;
     }
   }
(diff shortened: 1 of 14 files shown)


mjankowski:
It looks like you've renamed a bunch of things that USE those initial color vars, but not the color vars themselves?

I have some related WIP on this - https://github.com/tootsuite/mastodon/compare/master...mjankowski:colors-variable-names - which was starting to go in similar direction.

stephenburgess8:
This is a great idea. I have some thoughts for taking or leaving:

One challenge is that this is a sort of API for customizing the look and feel. Since that's the case, it is better to have more general names. Similar to @mjankowski's names. But maybe, instead of `$theme-base-color: $custom-base-color`, just `$base-color: $theme-base-color`

My personal preference is to follow a naming convention with a reverse order from english. What I mean is instead of `$thirty-darken-gray` trying `$background-darken-30`.

Tying variable names to specific numbers or values means that if you ultimately change the _value_ you have to change the _variable name_ as well. For people customizing using the variable names, this could result in some cognitive dissonance when they change the name `base-color-darken-30` to instead be `darken($base-color, 35%)`. So, as much as possible, I think it's a good idea to go with something along the lines of `$base-color-darkest`, `$base-color-dark`, `$base-color-medium-dark`, `$base-color-medium`, etc. These values are intended to be changed, so giving them variable names that aren't tied to specific values could help make that clear.

Overall I think reducing the number of variations on colors is a great idea. Also, need to watch out for colors that are being set directly. I know for sure there's some `#000` and `#fff`s out there.

Naouak:
@stephenburgess8 Names are temporary. We need to reduce number of colors before naming all the variables. Once we have a nice subset of colors, we should be able to easily name them according to their usage.

Naouak:
@mjankowski Nice. I was planning to do that first but some colors are used for several non related things. That's why I prefer waiting for the end of cleanup/color count reducing to start naming things.

Kozlika:
As shown in the following screenshots, this refactoring of the CSS variable names has no impact on the way pages are rendered.

Before:
css-variables-before

After:
css-variables-after

mjankowski:
Closing after merge of https://github.com/tootsuite/mastodon/pull/2912

Naouak:
@mjankowski Your merge request did only a small portion of what this merge aims to do. You should not have closed this one.

mjankowski:
I didn't realize there was more coming -- apologies!

Please rebase + reopen when there's something to review.

Naouak:
I will. 

It is just a really long job (already did several hours on this one) and there is still the dreadful components.scss to refactor.

If anyone want to provide any help on this one, I will gladly accept every help available as it is a long process.

mjankowski:
It seems like you're trying to do a lot of stuff in one big PR ... it might be easier to work through by doing one discrete thing at a time and doing a series of PRs for the whole effort.  There will be less risk of getting out of sync with master ... and if we're not sure about the exact direction here, it will reduce the risk that you waste time doing something that won't be merged.

PR #2417, opened 2017-04-24
Make /api/v1/accounts API to not require authentication

### LINKED ISSUE #2273: "/timelines/public" endpoint leaking public account entity
Hi folks!

## Issue

I planned to work on a federated user search, so that you can find people based on the stuff they've written in their bio (to connect with the folks you are interested in and which are like-minded). For that, I'd need the `account` entity of a given user on a given instance.

I checked the accounts endpoint, and performed a test on my public account:

`https://soshl.one/api/v1/accounts/1`
```
{
	"error": "The access token is invalid"
}
```
I almost gave up when I realized that the endpoint is protected by authorization.

However, I checked the public timeline of my instance, and found the `account` entity in question:

`https://soshl.one/api/v1/timelines/public`
```
{
		"id": 1626,
		"created_at": "2017-04-21T15:59:01.427Z",
		"in_reply_to_id": null,
		"in_reply_to_account_id": null,
		"sensitive": false,
		"spoiler_text": "",
		"visibility": "public",
		"application": null,
		"account": {
			"id": 1,
			"username": "MrGilbert",
			"acct": "MrGilbert",
			"display_name": "Mr. Gilbert",
			"locked": false,
			"created_at": "2017-04-19T12:47:48.387Z",
			"followers_count": 6,
			"following_count": 32,
			"statuses_count": 37,
			"note": "¯\\_( ･ө･ )_/¯ || Unity 3D ||  .NET || Evolutionary Humanism || :flag_de: :flag_eu:  :flag_um:",
			"url": "https://soshl.one/@MrGilbert",
			"avatar": "https://soshl.one/system/accounts/avatars/000/000/001/original/e689cc23e167205f.jpg?1492626140",
			"avatar_static": "https://soshl.one/system/accounts/avatars/000/000/001/original/e689cc23e167205f.jpg?1492626140",
			"header": "https://soshl.one/headers/original/missing.png",
			"header_static": "https://soshl.one/headers/original/missing.png"
		},
		"media_attachments": [],
		"mentions": [],
		"tags": [],
		"uri": "tag:soshl.one,2017-04-21:objectId=1626:objectType=Status",
		"content": "<p>ok, and back to normal. :)</p>",
		"url": "https://soshl.one/@MrGilbert/1626",
		"reblogs_count": 0,
		"favourites_count": 0,
		"reblog": null
	},
```

So, by monitoring all instances, I could build up a quite sufficient database after 24 hours. Which would be awesome for my use-case, but I'm not sure if that's intended.

## Suggestion
It would be good to unify the endpoints at this point. For that, I've the following suggestion

- Open access to  `/api/v1/accounts/:id` route on **public**(!) profiles.
- _(Optional)_ Disallow exposing the `account` entity to non-authenticated authorities if the account is public - however, we would then need to figure out which data is actually necessary to be put into the `account` entity on the `/timelines/public` endpoint.

Any thoughts on this one? Or isn't it an issue at all, but on purpose? I would be fine with that, too.

### Disclaimer
The tests where performed on my own account, and for verification purpose also on mastodon.social

* * * *

happycoloredbanana:
Fixes #2273

- opened ```show```, ```search```, ```statuses```, ```followers```, ```following``` methods in the same way as it's done for the ```/api/v1/statuses/``` API (i.e., will not require a token, but if a token is provided, it will be used for authorization)
- moved common code to the base class
- improved tests to check all three cases: valid token, no token, bogus token

Large diffs for ```accounts_controller_spec.rb``` are due to indentation


app/controllers/api/v1/accounts_controller.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class Api::V1::AccountsController < ApiController
-  before_action -> { doorkeeper_authorize! :read }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute, :update_credentials]
+  before_action -> { doorkeeper_authorize_if_got_token! :read }, only: [:show, :following, :followers, :statuses, :search]
+  before_action -> { doorkeeper_authorize! :read }, except: [:show, :following, :followers, :statuses, :search, :follow, :unfollow, :block, :unblock, :mute, :unmute, :update_credentials]
```
expenses: If you've got that many things in the `except:` clause, isn't it easier to use `only:` instead?

happycoloredbanana: I agree that ```:only``` would look nicer.
I have used ```:except``` for the case when a new action is added and whoever does this forgets to add this new action to one of ```doorkeeper_authorize``` lines, so it falls automatically in the one with ```:except``` and at least some authorization is done.

Gargron (approved):

Gargron:
One of the reasons I am hesitant to merge this PR is because I'm not sure how much of the API *should* be opened up. I mean, it only takes an anonymous API request to /apps to get an OAuth token, which can be used as an app-only token (without user context). What are we gaining and losing from opening it up to not require even that?

happycoloredbanana:
From the API user's point of view: easier to use it in a curl one-liners and simple scripts without obligation to remember the client_id/client_secret and use them the next time the script runs.

From the admin's view: if everybody who wants to scrape an instance for some statistics will register an app, the app table in the database will fill up. And if the lazy ones will register a new app every time they run their script, the table will fill up even faster.

I think the second point is more important. (The first one we can solve by giving people libraries/examples).

PR #2418, opened 2017-04-24
Username above toot box UI changes

ashfurrow:
This PR makes the username above the new toot box into a link, and prepends the username with an `@`. This makes the username behave and appear consistently with the rest of the app.

Before, not a link:

screen shot 2017-04-25 at 1 11 35 am

After, a link:

screen shot 2017-04-25 at 1 11 20 am

It's a small thing, but always tripped me up :joy:


expenses:
Why not also show the display name as well? like `@username - Display name`?

Gargron (approved):

PR #2419, opened 2017-04-25
Post status idempotency - prevent duplicates from network failures

### LINKED ISSUE #2402: Idempotency header for new status creation
Implement a mechanism by which a client can supply a random string in the header of the POST /api/v1/statuses request. When status is created successfully on the server-side, cache its ID in relation to that string. If a client experiences network failure, it can try again with the same header. The server would then return the already existing status instead of creating duplicates. When a client experiences success, it can change the random string for the next status.

See <https://stripe.com/blog/idempotency>

Gargron:
Fix #2402 - Add Idempotency-Key header to PostStatusService that prevents duplicates. Web UI regenerates UUID for that header every time the compose form is changed or successfully submitted 

Also, fix Farsi i18n overwriting the English one

### DIFF
--- a/config/locales/simple_form.fa.yml
+++ b/config/locales/simple_form.fa.yml
@@ -1,5 +1,5 @@
 ---
-en:
+fa:
   simple_form:
     hints:
       defaults:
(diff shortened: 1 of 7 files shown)


config/locales/simple_form.fa.yml
```diff
@@ -1,5 +1,5 @@
 ---
-en:
+fa:
```
ashfurrow: Tricky, thanks for the fix 👍

ineffyble (approved):
Builds have passed, so this shouldn't break anything. I haven't tested this myself, but the concept is good, and I can't see any obvious code errors.

PR #2420, opened 2017-04-25
Fixed mention regex to support unicode

marrus-sh:
I haven't been able to test this but it should be pretty straightforward; right now because the regex uses `\w` things like `niñ@s` will create an `@s` mention. This changes it to stop mentions after any unicode word character.


Gargron:
If you could write out the test case in words ("if x then it should y") I can add the tests for you. I don't feel comfortable merging this blindly.

Artoria2e5:
Hmm, let me guess. If an `@` sign is preceded by anything that can form a word -- a latin character like `ñ`, an ASCII alphabet like `a`, a Han Character like `一`, then no at-mentions should be extracted. Current `\w` only handles the ASCII case while Ruby's `[:word:]` provides a more Unicode-oriented solution that handles the other two examples.

app/models/account.rb
```diff
@@ -3,7 +3,7 @@
 class Account < ApplicationRecord
   include Targetable
 
-  MENTION_RE = /(?:^|[^\/\w])@([a-z0-9_]+(?:@[a-z0-9\.\-]+[a-z0-9]+)?)/i
+  MENTION_RE = /(?:^|[^\/[:word:]])@([a-z0-9_]+(?:@[a-z0-9\.\-]+[a-z0-9]+)?)/i
```
Gargron: It would be nice if you added a simple test case for old behaviour vs new behaviour

Gargron (changes requested):

Gargron (approved):

PR #2422, opened 2017-04-25
[WIP] Establish trademark rules for the Mastodon project

Gargron:
The purpose of this PR is to codify the allowed use of "Mastodon" in domain names running the Mastodon software, in desktop/mobile/web apps/libraries that interface with the Mastodon software, redistributions of the Mastodon software with and without modifications (e.g. Docker images, Ansible scripts, etc). While preventing unrelated software from using the Mastodon brand.

Copyright and merchandise rules involving artwork (except for the logo), specifically by Dopatwo and Cutewitchirl, should also be codified. Those should stay a possession of the main Mastodon body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Mastodon
+Mastodon™
 ========
 
 [![Build Status](http://img.shields.io/travis/tootsuite/mastodon.svg)][travis]
@@ -7,7 +7,7 @@ Mastodon
 [travis]: https://travis-ci.org/tootsuite/mastodon
 [code_climate]: https://codeclimate.com/github/tootsuite/mastodon
 
-Mastodon is a free, open-source social network server. A decentralized solution to commercial platforms, it avoids the risks of a single company monopolizing your communication. Anyone can run Mastodon and participate in the social network seamlessly.
+Mastodon™ is a free, open-source social network server. A decentralized solution to commercial platforms, it avoids the risks of a single company monopolizing your communication. Anyone can run Mastodon™ and participate in the social network seamlessly.
 
 An alternative implementation of the GNU social project. Based on [ActivityStreams](https://en.wikipedia.org/wiki/Activity_Streams_(format)), [Webfinger](https://en.wikipedia.org/wiki/WebFinger), [PubsubHubbub](https://en.wikipedia.org/wiki/PubSubHubbub) and [Salmon](https://en.wikipedia.org/wiki/Salmon_(protocol)).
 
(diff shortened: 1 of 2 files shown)


ineffyble:
Is the little TM necessary in the README? I can't think of many open source projects that use it, and I know some of them have trademarks.

cwebber:
Trademark laws and copyright laws are two very different things, probably should be addressed separately.  That said...

> Copyright and merchandise rules involving artwork (except for the logo), specifically by Dopatwo and Cutewitchirl, should also be codified. Those should stay a possession of the main Mastodon body.
 
Codifying is good, but what does that mean?  There's a lot of artwork in Mastodon, including theming, etc.  I recommend personally choosing a libre license for your artwork generally.  If you need to protect the name from misuse, that's the domain of trademarks, not copyrights.

So, keep your artwork free, but as for trademark policy, if you're going down that path, you might want to look at how some other organizations handle their trademarks, [such as GNOME](https://www.gnome.org/foundation/legal-and-trademarks/), including their [usage guidelines](https://wiki.gnome.org/FoundationBoard/Resources/LicensingGuidelines).

Gargron:
I wanna use these for inspiration:

- [Nextcloud trademark policy](https://nextcloud.com/trademarks/)
- [EmojiOne artwork license](https://www.emojione.com/developers/free-license)

Trademark =/= copyright, and brand =/= artwork, but both are legal policy things which is why I am conflating the two in one PR.

cwebber:
EmojiOne's license is non-free, so fyi that'll prevent Mastodon from being packaged in something like Debian without people tearing out its branding.

Gargron:
Preferably I'd need a lawyer for this, I think. If you know anyone, feel free to e-mail me (though questionable whether I can afford a lawyer). Artwork =/= logo by the way. Only talking about the illustrations and re-selling them for profit.

mjankowski:
I know this is WIP and I expect more to come here ... but I'd expect more of this to go on the official project website, and less in the source code repo.

I think Rails itself does a decent job of balancing this -- there's no TM references in the GH repo (there is a license, of course), but then the project site has a TM page - http://rubyonrails.org/trademarks/ (the nextcloud one is also good).

My opinion here might vary a bit on who actual holds the TM and in what jurisdictions, and what your goals are to those ends.

cwebber:
> Artwork =/= logo by the way. Only talking about the illustrations and re-selling them for profit.

Ok, that's useful clarity... I was interpreting "artwork" as being pretty broad, clearly broader than you meant it!

ZiiX:
The ™ goes away once the trademark is granted in the US. You use the ™ to cite you have applied and have legally put a flag on the mounted while stating when it was first used in commerce.  Then, the ® is used once legally approved and granted. The ® doesn't have to be used everywhere but has to be prominent and publicly associated. Using the ™ now, though obnoxious looking, is simply how it works and it's professional.

yiskah:
I've heard a lot of contradicting things about the use of :tm: being necessary. If we can avoid it it would be ideal. It's very goofy and comes off as a joke. I don't think it looks professional at all given that many places do not do the same (see: firefox)

Would it be sufficient to only have it appear at the bottom of the page where the license is linked to?

Gargron:
Gonna redo this one from scratch I think.

PR #2426, opened 2017-04-25
attach_media should be in transaction of creation status

walf443:
if UPDATE "media_attachments" transaction commit delayed, cache_collection cache status row without media_attachments.

that caused followings:

pawoo_ _

This item has a media_attachment. But it's cached by cache_collection. So, timeline json return media_attachments to blank.

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -14,15 +14,17 @@ class PostStatusService < BaseService
   # @return [Status]
   def call(account, text, in_reply_to = nil, options = {})
     media  = validate_media!(options[:media_ids])
-    status = account.statuses.create!(text: text,
-                                      thread: in_reply_to,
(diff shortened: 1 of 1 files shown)


Gargron:
Let me confirm, this line: https://github.com/pixiv/mastodon/blob/69e1b1b2c047776cf89f3b6406982000bbd5e4af/app/services/post_status_service.rb#L32 was executed *before* `attach_media` finished?

walf443:
No. 
For example, /api/v1/timelines/public called by another user while updating media_attachments. it cache new statuses row without media_attachments.

Gargron:
Ah, I see! I didn't consider the public timelines access. Okay... That makes sense. I wonder if there are other related issues when statuses are created in other places e.g. ProcessFeedService...

Gargron (approved):

PR #2437, opened 2017-04-25
Create a script for easy backups on docker-compose installations

halest:
Suggesting to add an easy-backup script for docker-compose-installations.
Finds db-mountpoints, creates structure on configurable target and backs up db and folder structure

### DIFF
--- a/docker-compose-mastodon-backup.sh
+++ b/docker-compose-mastodon-backup.sh
@@ -20,10 +20,14 @@ mkdir -p "$target/mastodon/data"
 
 echo "Backing up Database..."
 volume=$(docker inspect --format '{{ range .Mounts }}{{ if eq .Destination "/var/lib/postgresql/data" }}{{ .Source }}{{ end }}{{ end }}' $dbContainerName)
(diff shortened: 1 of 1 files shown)


expenses:
Could this be a rake task instead?

halest:
@expenses I don't know, maybe? It's working pretty well as is and I know pretty much nothing about rake so I I'm the wrong person to ask.

thurloat:
I'm all for easier backup methods, and helping users out with it -- but I think this method is a little too bash-y.

I think a rake task would work better / be more familiar, and allow users to be able to leverage their `.env.production` files instead of having to maintain edits to a shell script.

Gargron:
Something [like this](https://backup.github.io/backup/v4/) would be more appropriate I think

Gargron:
We want to perhaps add a rake method in the future, but not bash scripts. Thank you for your effort. You may distribute this as a gist or tutorial or a repository with useful bash scripts for Mastodon, but it's not for the main repository.

PR #2441, opened 2017-04-25
Launch onboarding tutorial on-demand

### LINKED ISSUE #2019: There is now way to access Getting Started tips after they close
Getting Started tips are very useful for new users, but there is no way to bring them up after first seeing them after registering. I think there should be an entry in "getting started" menu for them.

* * * *

ashfurrow:
This has been an oft-requested feature from users on my instance, and we've had a few open issues. I didn't get any feedback on where the button should be but users have suggested in the righthand nav column links, so that's where I put it.

**Important**: I also believe this [fixes a bug](https://github.com/ashfurrow/mastodon/commit/6e4324df36c1e7ab110fe8045243cd1f2c8b882c) preventing the onboarding from working, so if we don't want to merge this as-is I can create a PR just for that fix.

screen shot 2017-04-25 at 4 14 05 pm

Fixes https://github.com/tootsuite/mastodon/issues/2019


expenses:
I like it! You'll have to fix the merge conflict tho ^.

ashfurrow:
I've merged the changes from https://github.com/tootsuite/mastodon/pull/1309 but there is a problem: with the new headings _and_ the added link the bottom furry elephant friend is now scrolling:

screen shot 2017-04-26 at 9 27 09 am

I was already a bit worried about the placement here before the merged changes, now I'm really concerned. Maybe it's not a big deal, but I imagine many users have 13" screens like I do, and now have a scrollable bottom region. I'm okay with that but wanted to flag it as a possible blocker. Let me know 👍

ashfurrow:
I'm getting many failed tests, like `undefined method `deep_symbolize_keys' for "Òc":String`. Any ideas? I'll try to take a closer look tomorrow.

ashfurrow:
🍏 

Is this something we want to review as-is or should we consider moving the link somewhere else?

yiskah:
We need to stop putting things in that menu. I don't know how large your screen is but on my 1080p screen the menu looks like this 
long menu and on my locked alt account i also have the follow requests button so it looks like this: 
locked long menu
My work computer is smaller and you can't even read the one sentence of text under the menu let alone see the elephant.

Is there a better place we can put the button? I don't think it needs to be ever-present. Somewhere easy to find but easily forgotten when you don't need it anymore

Wonderfall:
With the many changes from the webpack branch, you'll have to edit this line : 

```
  render() {
    const { intl, me, dispatch } = this.props;
```

ashfurrow:
Yeah yeah, I'm just going to close since this is a relatively easy PR once we have the design figured out in #2019.

Wonderfall:
It's too bad, I liked this feature. 😞

yiskah:
It might still happen just not with this specific code

Wonderfall:
I know, and I still use this code for my own instance. I look forward to have this feature implemented.

PR #2444, opened 2017-04-25
Enable to choose OpenStack object storage

hakoai:
Since the object storage I use is not S3 compatible, I made it enable to use Openstack Object Storage instead of S3.


ykzts:
dup #2322

hakoai:
Oops, sorry.

Issue #1985, opened 2017-04-17
Return media attachment dimensions


siuying:
MediaAttachment currently do not return dimension of the image/video. This prevent the view from properly sized the placeholder. 

If the api optionally return media dimension, we can make the client render the preview at proper size:

<img width="239" alt="Tweetbot for Mac" src="https://cloud.githubusercontent.com/assets/8608/25084935/fd9dab02-2391-11e7-8fe0-bff59962640e.png">

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

ashfurrow:
Did some brief research into this, Paperclip has a companion gem for storing the asset dimensions in a database column: https://github.com/teeparham/paperclip-meta Beyond a basic db migration, it looks like it would require "rebuilding thumbnails", and I'm not sure what's involved in that.

siuying:
The dimension field can be optional, then we don't need to worry about rebuilding thumbnails -- a rebuild task can be added later if someone really want all media have dimensions.

ashfurrow:
Tried for a while to get this, but it's beyond my abilities at the moment. If someone else could take a look, that'd be :100:

siuying:
I'm looking into it!

siuying:
Tried to add paperclip-meta, but it [conflicts with paperclip-av-transcoder](https://github.com/teeparham/paperclip-meta/issues/39). I don't really want to monkey patch it, so I use custom metadata format.

1. Add a field ``file_meta`` to MediaAttachment
2. For images, find the meta data using Paperclip::Geometry
3. For video, paperclip-av-transcoder will set the meta data

Note:

file_meta is json type, which require newer postgresql (9.4)

ashfurrow:
Nice work @siuying !

PR #2448, opened 2017-04-25
Add media dimensions

siuying:
See #1985

### DIFF
--- a/spec/models/media_attachment_spec.rb
+++ b/spec/models/media_attachment_spec.rb
@@ -11,6 +11,13 @@ RSpec.describe MediaAttachment, type: :model do
     it 'converts original file to mp4' do
       expect(media.file_content_type).to eq 'video/mp4'
     end
+
+    it 'sets meta' do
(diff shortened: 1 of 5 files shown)


Gargron (approved):

spec/models/media_attachment_spec.rb
```diff
@@ -23,5 +23,11 @@
     it 'leaves original file as-is' do
       expect(media.file_content_type).to eq 'image/gif'
     end
+
+    it 'sets dimensions' do
+      expect(media.meta[:width]).to eq 600
```
Gargron: I think this needs to be `media.file.meta`

Gargron (changes requested):

spec/models/media_attachment_spec.rb
siuying: Fixed, sorry have not setup the test env, not got it sorted.

Gargron (approved):

PR #2449, opened 2017-04-25
Convert to "container-based" travis build

Artoria2e5:
Since all ppa sources used are currently whitelisted by Travis, why don't we convert to a "sudo-less" structure that promises a shorter build start time?


Artoria2e5:
How should I verify it's booting faster though...

Gargron (approved):

PR #2471, opened 2017-04-26
Update Japanese translation for activitypub and authorized followers

anon5r:
- Add Japanese translation for Activitypub
- Add Japanese translation for Authorized followers
... and some new translation items.

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -40,14 +40,25 @@ ja:
     posts: 投稿
     remote_follow: リモートフォロー
     unfollow: フォロー解除
+  activitypub:
+    activity:
+      announce:
+        name: "%{account_name} さんがアクティビティをシェアしました"
+      create:
+        name: "%{account_name} さんがノートを作成しました"
+    outbox:
+      name: "%{account_name}の送信トレイ"
+      summary: "%{account_name} さんからのアクティビティコレクション"
   admin:
     accounts:
       are_you_sure: 本当に実行しますか?
+      confirm: 確認
+      confirmed: 確認済み
       display_name: 表示名
       domain: ドメイン
       edit: 編集
       email: E-mail
-      feed_url: Feed URL
+      feed_url: フィードURL
       followers: フォロワー数
       follows: フォロー数
       location:
@@ -136,6 +147,7 @@ ja:
       report: レポート#%{id}
       reported_account: 報告対象アカウント
       reported_by: 報告者
+      report_contents: 内容
       resolved: 解決済み
       silence_account: アカウントをサイレンス
       status: ステータス
@@ -204,6 +216,7 @@ ja:
       x_months: "%{count}月"
       x_seconds: "%{count}秒"
   errors:
+    '403': このページを表示する権限がありません
     '404': お探しのページは見つかりませんでした。
     '410': お探しのページはもう存在しません。
     '422':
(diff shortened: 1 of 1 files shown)


config/locales/ja.yml
```diff
@@ -284,8 +310,11 @@ ja:
     show_more: もっと見る
     visibilities:
       private: 非公開 - フォロワーだけに公開
```
mecab: I suggest removing `- フォロワーだけに公開` because it could be verbose when displayed with `private_long`. (and I think it is expected to be - see #2111.).

`public`, `unlisted` should also be updated.

config/locales/ja.yml
```diff
@@ -284,8 +310,11 @@ ja:
...
+      private_long: フォロワーにのみ表示されます
       public: 公開 - 公開タイムラインに投稿する
+      public_long: 誰でも見ることができます
```
mecab: if `- 公開タイムラインに投稿する` will be removed, put them here could help understanding. I mean
`誰でも見ることができ、かつ公開タイムラインに投稿されます。

mecab (commented):
Cool!

I put some suggestions related to #2111.

config/locales/ja.yml
```diff
@@ -40,14 +40,25 @@ ja:
...
     remote_follow: リモートフォロー
     unfollow: フォロー解除
+  activitypub:
+    activity:
+      announce:
+        name: "%{account_name} さんがアクティビティをシェアしました"
+      create:
+        name: "%{account_name} さんがノートを作成しました"
+    outbox:
+      name: "%{account_name}の送信トレイ"
```
ykzts: Don't need space after placeholder?

config/locales/ja.yml
```diff
@@ -215,6 +228,18 @@ ja:
     follows: フォロー
     mutes: ミュート
     storage: メディア
+  followers:
+    domain: ドメイン
+    explanation_html: あなたの投稿のプライバシーを確保したい場合、誰があなたをフォローしているのかを把握している必要があります。 <strong>プライベート投稿は、あなたをフォロワーのすべてのインスタンスに配信されます</strong>。 もしフォロワーのインスタンス管理者、またはソフトウェアがあなたのプライバシーを尊重してくれるか信頼できるかをいま一度確認し、必要に応じてフォロワーを削除することも検討してください。
```
ykzts: I think that "もう一度" is more general than "いま一度".

config/locales/ja.yml
mecab: and "さん" might be needed, IMO

config/locales/ja.yml
anon5r: Thank you for your review.
I think too. I will fix to remove  description after dash from `public`, `unlisted`, and `private`.

config/locales/ja.yml
```diff
@@ -274,6 +299,7 @@ ja:
     back: 戻る
     edit_profile: プロフィールを編集
     export: データのエクスポート
+    followers: フォロー承認
```
unarist: Since original text is "Authorized followers", literal translation would be "承認済みのフォロワー", and "フォロー承認" sounds like "Approve follow requests".

However, this page is used to remove follower instances, not individual followers. so we may want to translate it like "信頼済みのインスタンス".

config/locales/ja.yml
```diff
@@ -204,6 +216,7 @@ ja:
       x_months: "%{count}月"
       x_seconds: "%{count}秒"
   errors:
+    '403': このページを表示する権限がありません
```
mecab: missing `。`

Gargron (approved):

anon5r:
@mecab @ykzts @unarist Thank you for your reviews and more better traslations!

PR #2477, opened 2017-04-26
fix onboarding modal

Wonderfall:
I don't know if I'm doing this right, but this fixed the onboarding modal for me (domain is actually not defined for PageSix).

PR #2482, opened 2017-04-26
Fix undefined variable in OnboardingModal

ykzts:
The same problem to #2477.


ykzts:
```javascript
fetch('/api/web/settings', { method: 'put', credentials: 'include', body: JSON.stringify(Object.assign({},JSON.parse(document.getElementById('initial-state').text).settings,{onboarded:false})) }).then(console.log.bind(console)).catch(console.error.bind(console))
```

When you run the above in the Developer Tools Console, the modal opens again.

Gargron (approved):

PR #2483, opened 2017-04-26
Update German translation

benediktg:
especially add some missing translation strings and remove those that are not used anymore


benediktg:
I will open a rebased Pull Request

PR #2495, opened 2017-04-26
Add scheduler to Sidekiq to remove needs of cron on the system

mtparet:
### Description ###
Add gem sidekiq-scheduler https://github.com/moove-it/sidekiq-scheduler

Add a worker which is triggered each day.

Move some code from rake tasks to a service which could be called from the worker and from the rake tasks. (What do you think about this?)

Solve https://github.com/tootsuite/mastodon/issues/2473


app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+  def self.log_start_task(task)
+    log "Starting #{task} at #{Time.now.utc}"
+  end
+
+  def self.log_end_task(task)
+    log "Completing #{task} at #{Time.now.utc}"
+  end
+
+  def self.log(message)
+    puts message
```
mtparet: I keep the puts if we use it from the rake task and we want the message displayed directly to the console

mjankowski:
Is having to use cron really a problem?

kaniini:
@mjankowski it is an annoyance if you're trying to run on a serverless hosting (especially in this case, Scalingo), as most serverless hostings do not provide scheduled tasks.  right now a lot of serverless instances are completely broken because the docs did not even mention the crons being necessary on the deployment guides.

mtparet:
Even if is not a problem, it is a nice improvement:

- remove OS interaction needs
- when modifying scheduling tasks no human intervention is needed from mastodon instance maintainers
- monitoring of scheduled tasks is integrated in sidekiq view
- monitoring of the run of scheduled tasks is available through sidekiq

aral:
This issue, as I see it, could fundamentally break federation in Mastodon. Here’s why (and someone please correct me if I’ve missed anything):

1. The Heroku and Scalingo easy installation scripts result in what are essentially broken Mastodon installations at the moment and will result in PuSH subscriptions expiring unless people manually run the rake task to refresh them (which no one will)

2. As PuSH subscriptions expire, federation breaks.

I don’t know if we know / have a way of knowing how many installations of Mastodon are at Heroku/Scalingo but this would help us gauge the extent of the problem. Not to be alarmist but – unless I’m missing something – this has the potential to fundamentally break federation in Mastodon unless it is handled swiftly.

PS. In the meanwhile, on Scalingo, at least, here are instructions on manually running the rake task: https://mastodon.ar.al/@aral/39573

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      def clear_all
+        Redis.current.keys('feed:*').each { |key| Redis.current.del(key) }
+      end
+    end
+  end
+
+  class Users
+    class << self
+      def clear
```
expenses: `Users.clear` reads like it's wiping the db of users. Perhaps something like `cleanup` would be better?

orzFly:
+1 for this. Now there are lots of deployments in the wild don't have the cron jobs set up. 

@aral Not only serverless deployments. Not everyone would find the Cron Jobs part in the installation documentation.

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+class MastodonTasksService
```
mjankowski: This is sort of a weird name, because it's not really the same thing as the rest of the stuff in services.

I don't have a great idea for a better name/location ... maybe just `lib/maintenance.rb` or something boring?

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+  def self.daily
+    log_start_task('feeds::clear')
```
mjankowski: We can probably use less logging here?

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+    log_end_task('users::clear')
+
+    log_start_task('push::refresh')
+    Push.refresh
+    log_end_task('push::refresh')
+  end
+
+  class Media
+    class << self
+      def clear
```
mjankowski: Could these all just be class methods on the existing AR classes being used here, instead of inside this module?

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+        MediaAttachment.where(status_id: nil).where('created_at < ?', 1.day.ago).find_each(&:destroy)
+      end
+
+      def remove_silenced
+        MediaAttachment.where(account: Account.silenced).find_each(&:destroy)
+      end
+
+      def remove_remote
+        MediaAttachment.where.not(remote_url: '').where('created_at < ?', 1.week.ago).find_each do |media|
```
mjankowski: If you do move these to class methods, there are probably some existing scopes you can re-use to make this easier to read.

app/services/mastodon_tasks_service.rb
```diff
@@ -0,0 +1,101 @@
...
+      def clear
+        # Users that never confirmed e-mail never signed in, means they
+        # only have a user record and an avatar record, with no files uploaded
+        User.where('confirmed_at is NULL AND confirmation_sent_at <= ?', 2.days.ago).find_in_batches do |batch|
+          Account.where(id: batch.map(&:account_id)).delete_all
+          User.where(id: batch.map(&:id)).delete_all
+        end
+      end
+
+      def admins
```
mjankowski: This could probably just stay in the rake task ... isn't really part of "maintenance" ?

app/workers/daily_worker.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DailyWorker
```
mjankowski: The name here should probably be about what it does, not how often it is run?

app/services/mastodon_tasks_service.rb
mjankowski: I agree that `clear` on user sounds scary...

mtparet: I agree

app/services/mastodon_tasks_service.rb
mtparet: In my mind what is in `lib/` should not interacts directly with the application business logic and the best place I found was `services/` even if other stuff, here, are not the same.

About the name, `MaintenanceService`  ?

app/services/mastodon_tasks_service.rb
mtparet: yes indeed, I kept what we were doing in the rake task but I could change.

app/services/mastodon_tasks_service.rb
mtparet: If we consider that these methods are simple we could indeed move these to the AR classes.

app/services/mastodon_tasks_service.rb
mtparet: I'll take a look, thanks.

app/services/mastodon_tasks_service.rb
mtparet: Indeed, I move it here to keep all in one place but this one does not make sense to be shared.

app/workers/daily_worker.rb
mtparet: Yes indeed we could call it `MaintenanceWorker`.

Gargron:
#2799 introduces sidekiq-scheduler and moves refreshing of PuSH subscriptions to it instead of cron. That's the most important thing as @aral noted. There are other things but I am not sure if all of them should be "mandatory" for everybody:

- Removing unattached media (obviously mandatory, who needs tempfiles?)
- Unsubscribing from remote users with no local followers (some people might not wanna do that?)
- Removing accounts of unconfirmed users - frees up usernames/e-mails (sounds like it should be mandatory, but historically it's a recent addition)

aral:
@Gargron Thank you so much for #2799. Regarding your notes, I don’t immediately see how #1 and #3 would cause any trouble for anyone if they were mandatory. Regarding #2 (“Unsubscribing from remote users with no local followers”), I’d definitely _not_ want that as mandatory as that’s an important part of the process of discovering new people for me as someone who runs an instance of one just for myself :)

kaniini:
it would be nice to make them controlled by Sidekiq but selectable in the configuration panel so you can decide if you want them or not.

mjankowski:
I'd vote for:

- Make removal of unattached media an automated thing triggered by scheduler
- Keep remote account cleanup and local account/user cleanup as optional tasks performed via rake

kaniini:
Pretty much what @mjankowski said.  Make media cleanup automated, and remote account cleanup manually triggered.

PR #2502, opened 2017-04-26
use Twitter::Extractor for creating links

masarakki:
I found a problem in creating link tag.

Conditions:

- Status contains mentions
- Status contains url with `=@username` (in mentions)

For example,

```
@masarakki

http://example.com/?x=@masarakki
```

In this case,
At first, `encode_and_link_urls` convert url to `<a href="http://example.com/?x=@masarakki" rel="nofollow ...">http://example.com/?x=@masarakki</a>`. 
Then, because of `=@username` **matches to Account::ACCOUNT_RE**, `@masarakki` will be convert to link by `link_mentions`. 
At last, it will be `<a href="http://example.com/?x=<a href="/...">@masarakki</a> rel="nofollow ...">`

Sample: https://friends.nico/@masarakki/2717988

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -13,11 +13,10 @@ class Formatter
     return reformat(status.content) unless status.local?
 
     html = status.text
-    html = encode_and_link_urls(html)
+    html = encode_and_link_urls(html, status.mentions)
+
     html = simple_format(html, {}, sanitize: false)
     html = html.delete("\n")
-    html = link_mentions(html, status.mentions)
-    html = link_hashtags(html)
 
     html.html_safe # rubocop:disable Rails/OutputSafety
   end
@@ -37,8 +36,6 @@ class Formatter
     html = encode_and_link_urls(account.note)
     html = simple_format(html, {}, sanitize: false)
     html = html.delete("\n")
-    html = link_accounts(html)
-    html = link_hashtags(html)
 
     html.html_safe # rubocop:disable Rails/OutputSafety
   end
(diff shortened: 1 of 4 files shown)


Gargron:
cc @abcang

masarakki:
I think simple replacement is dangerous, it is better to use `placeholder` like replacement.

for example, 

html, links = encode_and_link_urls("@masakki Hello, world!,  http://example.com/?x=@masarakki")
# html: "@masarakki Hello, world!, \1", links: ['<a href="http://example.com/?x=@masarakki">http://example.com/?x=@masarakki</a>']
html, links = link_mentions(html, link)
# html: "\2 Hello, world!, \1", links: ['<a href="http://example.com/?x=@masarakki>...</a>' '<a href="/users/1">@masarakki</a>']
....

abcang:
I think that it is safe to replace it after extracting the target like twitter-text.

app/lib/formatter.rb
```diff
@@ -53,51 +50,66 @@ def encode(html)
...
+    end
+  end
 
-    chars = html.to_s.to_char_a
-    html_attrs = {
-      target: '_blank',
-      rel: 'nofollow noopener',
-    }
-    result = ''
+  def rewite(text, entities)
```
Gargron: Is this meant to be called rewrite?

Gargron (changes requested):

app/lib/formatter.rb
masarakki: oh sorry

Gargron (approved):

PR #2511, opened 2017-04-27
Auth sign out

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/2491


Gargron:
Do we want to make suspended users signing out easier? Signing out means they get to see/use the sign up page again. It's not hard to reset your cookies to get the same result, but I feel like maybe we shouldn't make it easier?

mjankowski:
You could have a shared public computer with a suspended user and the cookies wind up in a state where that browser cannot access that instance until cookies are reset or expire, which maybe not everyone knows how to do?

I see your point re: not wanting to make it easier for someone in an abuse/spam cycle to create new accounts more easily, though...

Gargron (approved):

PR #2512, opened 2017-04-27
Catch error when server decryption fails on 2FA

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/2488


app/controllers/auth/sessions_controller.rb
```diff
@@ -51,6 +51,9 @@ def two_factor_enabled?
   def valid_otp_attempt?(user)
     user.validate_and_consume_otp!(user_params[:otp_attempt]) ||
       user.invalidate_otp_backup_code!(user_params[:otp_attempt])
+  rescue OpenSSL::Cipher::CipherError => error
+    Rails.logger.debug "OTP Validation failed: #{error}"
```
mtparet: If this is a server error, should we log it with the `error` level rather than the `debug` level ?

mjankowski: Yes, good catch. Will update.

Gargron (approved):

PR #2518, opened 2017-04-27
Show boosted user's avatar

kawax:
boosted-avatar

margin-right: 5px
boosted


app/assets/javascripts/components/components/status.jsx
```diff
@@ -53,6 +53,7 @@ class Status extends React.PureComponent {
         <div className='status__wrapper'>
           <div className='status__prepend'>
             <div className='status__prepend-icon-wrapper'><i className='fa fa-fw fa-retweet status__prepend-icon' /></div>
+            <Avatar src={status.getIn(['account', 'avatar'])} staticSrc={status.getIn(['account', 'avatar_static'])} size={20} style={{ display: 'inline-block', verticalAlign: 'middle' }} />
```
ykzts: Styles should be written in [`components.scss`](https://github.com/tootsuite/mastodon/blob/0dd71af5fab7d2bef59a6e9d467b2742476aee00/app/assets/stylesheets/components.scss). I think it would be better to have a screenshot.

app/assets/stylesheets/components.scss
```diff
@@ -554,6 +554,12 @@ a.status__content__spoiler-link {
   .status__display-name strong {
     color: lighten($color1, 26%);
   }
+
+  .status__avatar-boost {
+    display: inline-block;
+    vertical-align: middle;
+    margin-right: 2px;
```
expenses: The margin in the screenshot looks a little small for me. Maybe change this to something more like `5px`?

expenses:
Also, what about applying this to notifications as well? For example:

notifications

ashfurrow:
Nice work @kawax! I think the idea @expenses has of applying to notifications makes sense, what do you think?

kawax:
Are there still conflicts?
I can not do any more.

kawax:
fixed

kawax:
rebase

app/assets/stylesheets/components.scss
```diff
@@ -903,6 +909,12 @@ a.status__content__spoiler-link {
   .fa {
     color: $color4;
   }
+
+  .notification__avatar {
+    display: inline-block;
+    vertical-align: middle;
+    margin-right: 5px;
+  }
```
expenses: This is pretty nitpicky, but seeing as both `.status__avatar-boost` and `.notification__avatar` have the exact same styles, do you think you could combine them?

app/assets/javascripts/components/features/notifications/components/notification.jsx
```diff
@@ -51,7 +54,9 @@ class Notification extends React.PureComponent {
           <div className='notification__favourite-icon-wrapper'>
             <i className='fa fa-fw fa-retweet' />
           </div>
-
+          <div className='account__avatar-boost'>
+            <Avatar src={notification.getIn(['account', 'avatar'])} staticSrc={notification.getIn(['account', 'avatar_static'])} size={20} />
+          </div>
```
stephenburgess8: Since this is the same code in three different places, it seems like a good candidate for making a generic component and being DRY.

Is it possible to set the size in CSS rather than as a prop?

Gargron:
Not a fan of inlining an image in that text blurb. Have you considered an alternative design where a smaller avatar is overlayed in the corner of the big avatar? I have to specify that I don't know whether I would merge *that* for sure... But I'd be curious to see how that looks.

kawax:
overlay

Gargron (approved):

PR #2519, opened 2017-04-27
Update onboard federation description

Wonderfall:
`em` doesn't have a style. So I also took the opportunity to change a bit the federation description, tell me if you don't like, but we should remove those `em` tags or add a style.

### DIFF
--- a/app/assets/javascripts/components/features/ui/components/onboarding_modal.jsx
+++ b/app/assets/javascripts/components/features/ui/components/onboarding_modal.jsx
@@ -25,7 +25,7 @@ const PageOne = ({ acct, domain }) => (
 
     <div>
       <h1><FormattedMessage id='onboarding.page_one.welcome' defaultMessage='Welcome to Mastodon!' /></h1>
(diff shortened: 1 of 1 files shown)


yiskah (changes requested):
Hi, thanks for looking at this. 

Parens should be avoided when explaining technical subjects. It breaks up the sentence and makes it more difficult to follow. This was discussed earlier on the last pull. It's better to have two sentences that are simple than one long run-on sentence.

"Known as Fediverse" is grammatically incorrect in English and is also inaccurate since The Fediverse as a colloquial term is usually associated more with GNU Social (see fediverse.org). I also wouldn't prioritize teaching this word to new users as it's not used anywhere in the UI.

`<em>` tags are the WCAG standard way to emphasize and italicize text so that it is accessible to screen readers. If they are not working properly the solution is not to change the sentence just to avoid using `<em>` tags. We should look into why that is and find a solution to make the `<em>` tags work.

I would suggest closing this pull request. Thank you for your work but I think this weakens the effectiveness of the modal.

yiskah:
Also thank you @ashfurrow for notifying me of this

Wonderfall:
IMHO it looks better with parens. But that's a debate I don't want to have...
At least remove the `em` tags. I know why they are used, but their style is not defined.

PR #2520, opened 2017-04-27
Use PostgreSQL inheritance for blocks and mutes

akihikodaki:


Gargron:
Hi! Could you provide a really short rationale for this? I'm assuming it's backwards-compatible?

I've never had to deal with Postgres table inheritance before.

akihikodaki:
Sure. `blocks` and `mutes` has the same scheme, and streaming performs the same query for each of them.

```SQL
SELECT target_account_id FROM blocks WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)}) UNION SELECT target_account_id FROM mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)})
```

This is redundant, and it could be obstacle for further optimizations. The tables could merge the table by adding a new column named `block` to distinguish `block` and `mutes`.

```diff
+    create_table "block_mutes", force: :casecade do |t|
+      t.integer  "account_id",        null: false
+      t.integer  "target_account_id", null: false
+      t.boolean  "block",             null: false
+      t.datetime "created_at",        null: false
+      t.datetime "updated_at",        null: false
+    end
```

That removes the redundant query, and allows further optimizations, such as `GROUP BY`.

```SQL
SELECT target_account_id FROM block_mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)})
```

```SQL
SELECT account_id FROM block_mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)}) GROUP BY account_id
```

However, this makes the queries which distinguish blocks and mutes slower. Partitioning is a solution for such cases. PostgreSQL provides the feature as inheritance and `CHECK`.

```
ALTER TABLE blocks ADD CONSTRAINT check_blocks_on_block CHECK(block = TRUE), INHERIT block_mutes
```

That allows to query only one of them with almost zero overheads.

Gargron:
Do you think it would make sense to change the db schema dump format from ruby to SQL? Or is the representation in the schema right now correct?

akihikodaki:
I think the current representation is fine. I have no reason to reject it.

Gargron:
Okay. I am finalizing release v1.3 right now. There is not enough time to let this brew in master before that, so I will withhold merging until v1.3 is released. :+1:

akihikodaki:
41989716773ee43e104feab04bf4372f1ec1f4fc: rebased to 439b2dceda8191659a61fe556136813f0078a240

Gargron (approved):

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170425202925) do
+ActiveRecord::Schema.define(version: 20180428000000) do
```
ykzts: `2018`? Probably it will break.

PR #2470, opened 2017-04-26
Red favicon for dev

LindseyB:
This is super low priority but I found it helpful in my own work so I don't confuse local dev and production environments to have a red favicon on the dev environment. It's something that I have found super helpful, though my feelings won't be hurt if you decide not to merge. 😄 

screen shot 2017-04-26 at 12 44 02 am

PR #2524, opened 2017-04-27
Added '(dev)' to the title in development mode

expenses:
@LindseyB's PR that added a red favicon to development mode is great (#2470), but it's not especially clear, so I though I might expand it by adding `(Dev)` to the title of the pages as well.

I considered using a translatable string instead for this, but thought that it might not be necessary. I'm ok to change it if it is though.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,6 +20,6 @@ module ApplicationHelper
   end
 
   def title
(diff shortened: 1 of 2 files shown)


expenses:
@ashfurrow Oops, don't know how I missed that! Fixed!

ashfurrow (approved):

PR #2525, opened 2017-04-27
Improve shared status verification

Gargron:
Instead of parsing shared status contents verbatim, make roundtrip to purported original URL. Confirm that the "original" URL is from the same domain as the author it claims to be from.

### DIFF
--- a/app/services/fetch_remote_status_service.rb
+++ b/app/services/fetch_remote_status_service.rb
@@ -39,9 +39,17 @@ class FetchRemoteStatusService < BaseService
 
     Rails.logger.debug "Going to webfinger #{username}@#{domain}"
 
-    return FollowRemoteAccountService.new.call("#{username}@#{domain}")
+    account = FollowRemoteAccountService.new.call("#{username}@#{domain}")
+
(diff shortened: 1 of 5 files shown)


app/services/fetch_remote_status_service.rb
```diff
@@ -39,9 +39,17 @@ def extract_author(url, xml)
 
     Rails.logger.debug "Going to webfinger #{username}@#{domain}"
 
-    return FollowRemoteAccountService.new.call("#{username}@#{domain}")
+    account = FollowRemoteAccountService.new.call("#{username}@#{domain}")
+
+    return nil if confirmed_domain?(domain, account)
```
ashfurrow: This doesn't quite read right to me, return `nil` if the domain is confirmed? If it's correct, a clarifying comment could be helpful 👍

ashfurrow (approved):
Makes sense to me 👍

mjankowski (commented):
It would be good to have a spec here showing that an invalid share gets replaced by the original, or rejected if there is no original.

PR #2534, opened 2017-04-27
Fix Doorkeeper error

ykzts:
revert https://github.com/tootsuite/mastodon/commit/88725d6ce85115ea3b0652007db5d40a1c069be3#diff-e621685ac1c2689eaedaedf334c6123bR39

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -18,6 +18,12 @@ class HomeController < ApplicationController
   end
 
   def find_or_create_access_token
-    Doorkeeper::AccessToken.find_or_create_for(Doorkeeper::Application.where(superapp: true).first, current_user.id, 'read write follow', Doorkeeper.configuration.access_token_expires_in, Doorkeeper.configuration.refresh_token_enabled?)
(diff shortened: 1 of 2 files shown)


Gargron:
I am surprised these two things are related, but most importantly, that commit was a fix for a specific issue. Each load of the home controller created a new access token instead of reusing the existing one. That configuration option is *supposed* to be the answer to that. Why and what does it break? We need to find an alternative fix.

Gargron:
Maybe `Doorkeeper::OAuth::Scopes.from_string('read write follow')` instead of `'read write follow'` in https://github.com/tootsuite/mastodon/blob/master/app/controllers/home_controller.rb#L21

ykzts:
@Gargron 

> Maybe Doorkeeper::OAuth::Scopes.from_string('read write follow') instead of 'read write follow' in https://github.com/tootsuite/mastodon/blob/master/app/controllers/home_controller.rb#L21

oh! Great! The problem is resolved.

ykzts:
Please re-review.

Gargron (approved):

PR #2538, opened 2017-04-27
Update Japanese translation (and English translation file) for JavaScript

unarist:
en.jsx:

- Remove duplicate/unused string
- Add `compose_form.lock_disclaimer` (#2111)

ja.jsx:

- Add missing translation
  - I translated "locked (account)" to  "非公開(アカウント)" according to existent translation now, but I think this translation is not correct. Twitter's Protected account hides user timeline, but Mastodon's locked account does not. It only means 1) following restriction 2) "follower-only" default IIRC, so it shouldn't be called "非公開" I think.
  - I left "Bon Appetoot!" as it is because I don't know good translation for it (and it isn't English).


app/assets/javascripts/components/locales/en.jsx
```diff
@@ -25,7 +25,6 @@ const en = {
   "account.unfollow": "Unfollow",
   "account.unmute": "Unmute @{name}",
   "boost_modal.combo": "You can press {combo} to skip this next time",
-  "column_back_button.label": "Back",
```
ykzts: why delete?

app/assets/javascripts/components/locales/en.jsx
```diff
@@ -94,7 +94,6 @@ const en = {
   "navigation_bar.public_timeline": "Federated timeline",
   "notification.favourite": "{name} favourited your status",
   "notification.follow": "{name} followed you",
-  "notification.mention": "{name} mentioned you",
```
ykzts: why delete?

app/assets/javascripts/components/locales/en.jsx
unarist: It is duplicate of L37.

ykzts: ok. Duplicate with line 37

app/assets/javascripts/components/locales/en.jsx
ykzts: If you do not use it anymore, if you do not need it, delete it in other languages.

unarist: Other `notification.*` strings are used in `app/assets/javascripts/components/features/notifications/components/notification.jsx`, but `notification.mention` is not used. Also I can't found use case in `app/assets/javascripts/**.jsx`.

However, I also cannot found use case in the tree at the time this line introduced (#1867), so it may be used in other places. How do you think @Gargron?

Gargron: notification.mention is used in actions/notifications.jsx for desktop notifications

unarist: new IntlMessageFormat(intlMessages[`notification.${notification.type}`], intlLocale)

Ah, I see. I will revert this line. Thanks!

app/assets/javascripts/components/locales/ja.jsx
```diff
@@ -109,6 +114,7 @@ const ja = {
   "onboarding.page_six.github": "MastodonはOSSです。バグ報告や機能要望あるいは貢献を{github}から行なえます。",
   "onboarding.page_six.apps_available": "iOS、Androidあるいは他のプラットフォームで使える{apps}があります。",
   "onboarding.page_six.various_app": "様々なモバイルアプリ",
+  "onboarding.page_six.appetoot": "Bon Appetoot!",
```
Gargron: This can be translated as "good luck!" or "enjoy!"

unarist: Well, I wanted to keep essence of "toot", but...hmm.

ykzts: I think that "ボナペトゥート!" is good.

Gargron (approved):

Issue #2545, opened 2017-04-27
Can't get remote followed; ActionView::Template::Error (path=users/<user>/remote_follow)


z411:
Been happening on master since yesterday (2016-04-26), I tried updating today but problem persists.
It happens when attempting to remote_follow an account on my instance.

```
Apr 27 17:52:29 kate bundle[8739]: [b396c76e-72b5-4213-8dcb-adff462ded2b] method=GET path=/users/z411/remote_follow format=html controller=RemoteFollowController acti
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/kaminari-core-1.0.1/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/doorkeeper-4.2.5/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/devise-4.2.1/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/pghero-1.6.4/app/views"
Apr 27 17:52:29 kate bundle[8739]: ' duration=6.23 view=0.00 db=0.57
Apr 27 17:52:29 kate bundle[8739]: [b396c76e-72b5-4213-8dcb-adff462ded2b]
Apr 27 17:52:29 kate bundle[8739]: [b396c76e-72b5-4213-8dcb-adff462ded2b] ActionView::Template::Error (Missing partial authorize_follow/_card with {:locale=>[:en], :f
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/kaminari-core-1.0.1/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/doorkeeper-4.2.5/app/views"
Apr 27 17:52:29 kate bundle[8739]: * "/home/mastodon/live/vendor/bundle/ruby/2.4.0/gems/devise-4.2.1/app/views"
[...]

PR #2546, opened 2017-04-27
Fix partial path in admin/reports

pfigel:
In #2505, the authorize_follow views were renamed to authorize_follows. This change was not applied in the show view of admin/reports, which causes a 500 when reports are viewed.

### DIFF
--- a/app/views/admin/reports/show.html.haml
+++ b/app/views/admin/reports/show.html.haml
@@ -4,11 +4,11 @@
 .report-accounts
   .report-accounts__item
     %strong= t('admin.reports.reported_account')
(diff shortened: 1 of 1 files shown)


mjankowski (approved):
Thanks for the fix -- can you add a spec which hits this view and would have exposed the failure?

patf:
This should fix #2545.

patf:
I'll do a follow-up PR to add spec coverage at some point this weekend.

mjankowski:
Related: https://github.com/tootsuite/mastodon/pull/2547

PR #2553, opened 2017-04-27
piwik support ( opensource self hosted alternative to GA )

gled-rs:
As I saw quite a bunch of instances using google analytics, here is a PR to add Piwik support to Mastodon ( self hosted / opensource clone of GA ).

It is in production for us on mastodon.host, anonymisation of the data being done by Piwik quite well.

if there's interest let me know as there's TODO:
- a env variable to enable to feature
- a few variable to customize to which piwik setup and which site id.


app/views/layouts/application.html.haml
```diff
@@ -28,3 +27,8 @@
 
   %body{ class: add_rtl_body_class(body_classes) }
     = content_for?(:content) ? yield(:content) : yield
+    = javascript_include_tag '/piwik'
+    %noscript 
+      %p 
+        %img(src="https://analytics.remote-shell.net/piwik.php?idsite=2&rec=1" style="border:0;" alt="")
```
expenses: Why are you using a tracking pixel from [https://analytics.remote-shell.net](analytics.remote-shell.net)?

gled-rs: second item of the TODO and in the commit logs ( urls are hardcoded right now ).

As I see some interest, working on dynamic for those parameters !

expenses:
While I think the default setting to not track users is a good one, supporting piwik is a good idea because it gives admins an easy alternative to GA.

gled-rs:
Thanks @expenses , I modified the PR so that now PIWIK_DOMAIN is dynamic and not anymore hardcoded. Indeed, I started that as I saw some other instances relying on GA, and to start digging into the code. 

Now I need to learn how to access PIWIK_ENABLED from application.html.haml and find a way to disable the _paq.push from status_containers.jsx if PIWIK_ENABLED != true ( thinking about a boolean in initial_state.json.rabl )

ykzts:
I do not think there is any reason why I have to include this upstream.

gled-rs:
@ykzts so you do not see also any reason not not to merge it ? ;) 

Actually, it just makes it easier for admins to setup piwik ( opensource and self hosted ) rather that hacking  google analytics all together in mastodon, and I rather have my data on piwik on a server I control than sending that to GA. It's a incentive for those instances, and it's disabled by default.

ykzts:
Isn't it contradictory to Mastodon's policy?

> Ethical design: no ads, no tracking

from: https://github.com/tootsuite/mastodon/blob/v1.3.1/config/locales/en.yml#L18

expenses:
@ykzts Policies like that can only really apply to Mastodon.social, because there is nothing stopping other instances from using tracking.

I get that people don't want to be tracked on mastodon instances, and for most instances, they won't be. But for the instances that _do_ implement tracking, being able to use Piwik easily instead of resorting to GA is a better solution.

ykzts:
@expenses In that case, I think you should hide the "Ethical design" ([code](https://github.com/tootsuite/mastodon/blob/v1.3.1/app/views/about/show.html.haml#L68-L70)) when `PIWIK_ENABLED` is `true`. It will fool the user.

gled-rs:
@ykzts concern noted.
Just also want to point you to this ;)

https://github.com/tootsuite/mastodon/blob/master/app/views/about/terms.en.html.haml#L7-L50

.env.production.sample
```diff
@@ -90,3 +90,8 @@ SMTP_FROM_ADDRESS=notifications@example.com
 # Cluster number setting for streaming API server.
 # If you comment out following line, cluster number will be `numOfCpuCores - 1`.
 STREAMING_CLUSTER_NUM=1
+
+# Piwik ( Optional )
+# PIWIK_ENABLED=true
```
nightpool: This can probably be condensed down to two variables, no? if there is a piwik domain, then enable piwik, otherwise don't.

app/controllers/home_controller.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
 class HomeController < ApplicationController
-  before_action :authenticate_user!
+  protect_from_forgery except: :piwik
+  skip_before_action :store_current_location, only: :piwik
+  # rubocop sees that as a hash ?!
+  before_action :authenticate_user!, :except => [:piwik] # rubocop:disable Style/HashSyntax
```
nightpool: this could alternatively be `%i(piwik)`, which is a little cleaner.

.env.production.sample
gled-rs: That is correct. I wanted to adopt the same style as above for S3, and it was also useful not to put the piwik domain in the initial_state json, but I have no feelings toward it ;)

I'll change it tomorrow, need to go soon now :/

app/controllers/home_controller.rb
gled-rs: Thanks for the tip !

I'm quite new or rusty in RoR/ruby pick one ;)

Correction pushed tomorrow !

config/piwik.yml
```diff
@@ -0,0 +1,33 @@
+# Configuration:
+# 
+# disabled
+#   false if tracking tag should be shown
+# use_async
+#   Set to true if you want to use asynchronous tracking
+# url
+#   The url of your piwik instance (e.g. localhost/piwik/
+# id_site
```
nightpool: why are these two things here and in `ENV`? they should be either one place or another but not both, right?

gled-rs: me forgetting to clean the mess left by piwik gem ;)

By the way, would you prefer using the piwik gem or the simple solution implemented at the moment ?

expenses:
@gled-rs It's also worth noting that the Private Policy [is just copy-pasted at the moment](https://mastodon.social/@Gargron/3482305).

gled-rs:
I know @expenses , that is going to be one of the next subjects, putting the terms in a setting like the instance description ;)

mjankowski (changes requested):
That's a lot of code to add for something which is optional, and maybe not used in that many instances?

Doing any sort of tracking or analytics strikes me as something best left to each instance, rather than mandated by the core app?

gled-rs:
72 lines of code is a lot ? ;) 

Anyway, there's around 5-6% of the instances out there today with some sort of GA deployed, just looked at the scrapping results.

Gargron:
I'm inclined to say that if you want to add tracking, you can modify the code downstream, but it doesn't need to be in the main repository.

PR #2554, opened 2017-04-27
No ratelimit on localhost

gled-rs:
For those of us who runs some bots/automated tools on the same server as mastodon, it can be helpful to whitelist localhost from the ratelimit.


ashfurrow (approved):
Cool, this makes sense to me. In the future it might be cool to have an admin0-configurable whitelist, but in the meantime this sounds like a good idea.

PR #2571, opened 2017-04-28
Update version to 1.3.1

ashfurrow:
I'll send a follow-up that automates this process.

### DIFF
--- a/lib/mastodon/version.rb
+++ b/lib/mastodon/version.rb
@@ -13,7 +13,7 @@ module Mastodon
     end
 
     def patch
-      0
+      1
     end
 
     def pre


ashfurrow:
Hmm, took a look at automating this but it's not toooooo easy after all. Maybe a "release checklist" doc would be better?

mjankowski (approved):
This is fine as a change, but the tag is already out, so oh well.

There are a few more fixes in now, and maybe .2 goes out soon and we can tag correctly.

Agreed that checklist or automate will help. There are many tools that do this for non packages and ruby gems, we can probably borrow.

ashfurrow:
Awesome, I've opened https://github.com/tootsuite/mastodon/issues/2581

PR #2577, opened 2017-04-28
Use the new Ruby 1.9 hash syntax in Gemfile

y-temp4:


mjankowski (approved):
I like how you call it "new". Solid

PR #2584, opened 2017-04-28
Add option to disable two factor auth in admin accounts panel.

### LINKED ISSUE #2578: Ability to disable 2fa from the accounts page.
After the last upgrade, one of my users lost the ability to log in to my instance due to issues with 2fa. It would be nice to be able to disable it from the admin UI, the same as resetting the password of a user.

Otherwise we are forced to disable it via the DB which while a simple procedure involves a bit more work than just hitting a button.

As an example query: `UPDATE users SET otp_required_for_login=false WHERE email='ex@mple.com';`

* * * *

cattebune:
Closes #2578, adds an option to disable 2FA in the accounts panel. UI change displayed below.

This will require updated translations, if anyone can help translate that would be awesome.

UI Change

### DIFF
--- a/app/controllers/admin/two_factor_authentications_controller.rb
+++ b/app/controllers/admin/two_factor_authentications_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Admin
+  class TwoFactorAuthenticationsController < BaseController
+    before_action :set_account
+
+    def destroy
+      @account.user.otp_required_for_login = false
+      @account.user.otp_backup_codes.clear
+      @account.user.save!
+      redirect_to admin_accounts_path
+    end
+
+    private
+
+    def set_account
+      @account = Account.find(params[:account_id])
+    end
+  end
+end
--- a/app/controllers/admin/two_factor_authentications_controller.rb
+++ b/app/controllers/admin/two_factor_authentications_controller.rb
@@ -2,19 +2,17 @@
 
 module Admin
   class TwoFactorAuthenticationsController < BaseController
-    before_action :set_account
+    before_action :set_user
 
     def destroy
-      @account.user.otp_required_for_login = false
-      @account.user.otp_backup_codes.clear
-      @account.user.save!
+      @user.disable_two_factor!
       redirect_to admin_accounts_path
     end
 
     private
 
-    def set_account
-      @account = Account.find(params[:account_id])
+    def set_user
+      @user = User.find(params[:user_id])
     end
   end
 end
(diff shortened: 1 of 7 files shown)


kayleej:
Whoops, looks like I left in some trailing newlines. I'll update and squash

wxcafe (approved):
lgtm

thurloat:
should the `otp_backup_codes` not be emptied out as well? 

Once a user gets to log in without their 2FA code, I think we should completely invalidate existing 2FA data for that user, since it's no longer a valid form of identification.

wxcafe:
we could, but I don't think it matters too much? how could these codes be used if 2fa is disabled?

thurloat:
yeah I suppose that's right. they'll get regenerated again when the user re-enables 2FA on their account.

kayleej:
To be honest, it totally slipped my mind to purge those. Probably should clear the backup codes, just in case. I'll go ahead and update the pr after lunch 😄

kayleej:
Updated, now runs `@account.user.otp_backup_codes.clear` before saving.
Functionality is still fine on my end, passes tests apart from the one failure brought in by 01c206326f51867d56d9b07bc89ab60049e75aec

app/controllers/admin/two_factor_authentications_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Admin
+  class TwoFactorAuthenticationsController < BaseController
+    before_action :set_account
+
+    def destroy
+      @account.user.otp_required_for_login = false
```
mjankowski: Can you do at least some and maybe all of:

- Roll all of this up into a method like `User#disable_two_factor!` or something (which would set required to false, clear the backups, and attempt save!)
- Add spec coverage for this controller action which shows that it does what it says
- Add model coverage for that method on User, if you create one like that

app/controllers/admin/two_factor_authentications_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      @account.user.otp_backup_codes.clear
+      @account.user.save!
+      redirect_to admin_accounts_path
+    end
+
+    private
+
+    def set_account
+      @account = Account.find(params[:account_id])
```
mjankowski: This could probably hit an `admin/users/:id/two_factor_authentication#destroy` route, since it's only doing things to the user record and not the account record.

mjankowski (dismissed):

kayleej:
I think that's correct, @mjankowski can you take another look for us and let me know if I should update anything?

Gargron (approved):

PR #2586, opened 2017-04-28
Fix broken spec for about/links view partial

mjankowski:


mjankowski:
I didn't see this other PR before doing this - https://github.com/tootsuite/mastodon/pull/2585 - but they are essentially the same thing.

I'd prefer to not use instance variables in partials, so if we merge the other PR to fix the spec, we should do a subsequent refactor to clean that up.

Gargron (approved):

ykzts:
@mjankowski Thank you for follow up of my inadequate work!

PR #2594, opened 2017-04-28
Add tests to the Feed model

JoelQ:
This adds a test for the `Feed#get` method. The method queries data from
Redis and then does a lot of tranformations on it. While testing, I
realized that some of the transformations were redundant so I removed
them.

### DIFF
--- a/spec/models/feed_spec.rb
+++ b/spec/models/feed_spec.rb
@@ -2,6 +2,20 @@ require 'rails_helper'
 
 RSpec.describe Feed, type: :model do
   describe '#get' do
-    pending
+    it "gets statuses with ids in the range, maintining the order from Redis" do
+      account = Fabricate(:account)
+      Fabricate(:status, account: account, id: 1)
+      Fabricate(:status, account: account, id: 2)
+      Fabricate(:status, account: account, id: 3)
+      Fabricate(:status, account: account, id: 10)
+      redis = double(zrevrangebyscore: [["val2", 2.0], ["val1", 1.0], ["val3", 3.0], ["deleted", 4.0]])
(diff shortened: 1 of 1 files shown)


app/models/feed.rb
```diff
@@ -10,9 +10,7 @@ def get(limit, max_id = nil, since_id = nil)
     max_id     = '+inf' if max_id.blank?
     since_id   = '-inf' if since_id.blank?
     unhydrated = redis.zrevrangebyscore(key, "(#{max_id}", "(#{since_id}", limit: [0, limit], with_scores: true).map(&:last).map(&:to_i)
-    status_map = Status.where(id: unhydrated).cache_ids.map { |s| [s.id, s] }.to_h
-
-    unhydrated.map { |id| status_map[id] }.compact
+    Status.where(id: unhydrated).cache_ids
```
mjankowski: It looks like previously part of what was being collected in the map was a `[s.id, s]` pair, which presumably is a status id and a full status record.

Does the change here to just return the `cache_ids` result (which I think is going to be an array of [:id, :updated_at]` for each status?) change anything w/ how this is ultimately used?

I don't know enough about how `cache_ids` and `cache_collection` and redis are intended to interact here to have a solid opinion.

Gargron (changes requested):

app/models/feed.rb
JoelQ: Ah! I was wondering why this was necessary. Thanks for the explanation. I've updated the test to reflect the requirement of maintaining the order from Redis. They now fail with my "refactor" but pass with the original implementation 👍

Gargron (approved):

PR #2596, opened 2017-04-28
Column Flexible Width, and Dropdown Menu Fix

YaQ00:
| before | after |
|----|-----|
|  honeyview_1-1  |  honeyview_1-2   |
| honeyview_2-1 | honeyview_2-2 |
| honeyview_4-1 | honeyview_4-2 |
| honeyview_5-1 | honeyview_5-2 |

Account dropdown menu 140px => 25vw, 50vw, 90vw (responsive)

Width => drawer 20% (max-width 400px), column 80% / 3 (max-width 800px)

Width 2560px over => height 100vh

Narrow column, smallest menu => Wide & modern column!


mjankowski (dismissed):
Can you add a better description of what's changing here, and move your screenshots onto this PR?

app/assets/stylesheets/components.scss
```diff
@@ -1217,10 +1240,9 @@ a.status__content__spoiler-link {
   }
 
   .column, .drawer {
-    width: 350px;
     border-radius: 4px;
-    height: 90vh;
-    margin-top: 5vh;
+    height: 100vh;
+    margin: 0;
```
stephenburgess8: Before only `margin-top` was applied but now we're applying margin to all sides. If you don't need to apply margin to all sides, you should avoid doing so.

stephenburgess8 (changes requested):
Overall I like the fact that you're making this responsive and allowing the columns to grow. If you could outline what all the changes you're making are for in a list, remove or comment the `!important`, avoid the use of the universal selector, and add some screenshots that would be helpful. Best.

app/assets/stylesheets/components.scss
```diff
@@ -995,6 +995,22 @@ a.status__content__spoiler-link {
...
+    width: 50vw;
+    * {
+      width: 50vw;
+    }
+  }
+  @media screen and (max-width: 500px) {
+    width: 90vw;
+    * {
+      width: 90vw;
+    }
```
YaQ00: Thank you for your proposal. I take up all suggest.

YaQ00:
take up all suggest.

stephenburgess8:
Hi - I like the flexible column width idea.

I don't like the idea that the dropdowns will always be so wide. The current dropdowns are small at first, then expand with the content to fill up the column. Unless there text in the dropdowns is long, the width of the dropdown is still small. There is no need to make the dropdown so wide unless the content is also wide. Keep in mind that I will not be the one approving the pull request.

The new master branch now has columns that are centered on the page. It may help to see screenshots with the new style.

YaQ00:
Thx for the reply, stephenburgess8.

But you forget about the long text such as CJK area. As a CJK user, I thought it was very hard to see.
I also think that even long texts will be in good condition if the change is made. It will have good touch device condition too.

Some people do not need the ID of the account menu, but I think that redundancy should not be eliminated.

What are your thoughts on my proposal? Tell me your idea.
d

YaQ00:
How do you think? @Gargron @mjankowski @ykzts

beatrix-bitrot (approved):
as far as I can tell all reviewers' suggestions have been implemented

there are no merge conflicts

and this is an important usability fix

approving

stephenburgess8:
Hi - 
I experienced a bug when my screen was less than 1024. The rest of the columns fall out of viewport.

This PR makes a stylistic change from going from fixed column widths to one with variable column widths.

Currently, the dropdowns expand only when there's additional content in them. In your version here, they are always as wide as the column. I like the way it is currently handled.

I hope I was helpful responding to your PR. Best -S

Gargron:
I am afraid we are not looking to make the columns flexible width. The general look and feel of the web UI is already set in stone. Thank you for your effort though!

STPKITT:
> I am afraid we are not looking to make the columns flexible width. The general look and feel of the web UI is already set in stone

Are you seriously? This is ridicolous! There's so much space on many types of displays, yet Mastodon users have to scroll down every 3 or 4 toots (on a Full HD display with default zoom settings). Pretty much unusable the way it is currently.

PR #2599, opened 2017-04-28
Set correct attachment type for rejected media

pfigel:
In #2110, a new attachment type "unknown" was introduced for attachments that were rejected due to a domain being blocked using `reject_media`. However, the "type" field was never set to "unknown" because a default value of "0" (image) is set for that column, causing the `type.blank?` expression to always equal false.

This version uses type_changed? instead, causing the type to be set to "unknown" unless a type has been explicitly set. This introduces a small change in behaviour causing the type to be set to unknown before paperclip calls `before_post_process`. Presumably this behaviour is more appropriate than the current one because the attachment type has not been determined by that point.

Included are new tests for `ProcessFeedService` and `UpdateRemoteProfileService` which now check that remote media is downloaded for non-blocked domains and is rejected for others.

### DIFF
--- a/app/models/media_attachment.rb
+++ b/app/models/media_attachment.rb
@@ -96,7 +96,7 @@ class MediaAttachment < ApplicationRecord
   private
 
   def set_shortcode
-    self.type = :unknown if file.blank? && type.blank?
+    self.type = :unknown if file.blank? && !type_changed?
 
     return unless local?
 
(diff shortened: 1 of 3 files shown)


Gargron:
>This introduces a small change in behaviour causing the type to be set to unknown before paperclip calls before_post_process

I believe I have tests for this, but this doesn't break gif->video conversion, does it?

patf:
Did some quick manual tests with a png, gif and video (both locally and with a federated DM). The type field was set correctly for all cases and the gif was converted to/displayed as a video.

Gargron (approved):

PR #2600, opened 2017-04-28
Extract user tracking into concern

mjankowski:


ashfurrow (approved):
Nice! Love the improved spec coverage 👍

app/controllers/concerns/user_tracking_concern.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  included do
+    before_action :set_user_activity, if: %i(user_signed_in? user_needs_sign_in_update?)
+  end
+
+  private
+
+  def set_user_activity
+    # Mark as signed-in today
+    current_user.update_tracked_fields!(request)
```
mjankowski: I could not get the "update when prior value is nil" spec to pass without switching to `update_tracked_fields!` (forces a save) being used here. I'm not sure if that reveals a flaw in the prior implementation, or if there's an interaction with Devise that I don't understand.

It seems harmless to switch to the `!` method, since we do want it to save here, though.

ashfurrow: Anyone we should request a review from that might be able to help?

PR #2603, opened 2017-04-28
Allow to set the Terms and conditions through admin interface.

gled-rs:
Simple setting added to allow instance admins to easily change their terms.

### DIFF
--- a/app/views/about/terms.en.html.haml
+++ b/app/views/about/terms.en.html.haml
@@ -2,7 +2,7 @@
   #{site_hostname} Terms of Service and Privacy Policy
 
 .wrapper
-  - if instance_presenter.site_terms.blank?
+  - if @instance_presenter.site_terms.blank?
     %h2 Privacy Policy
   
     %h3#collect What information do we collect?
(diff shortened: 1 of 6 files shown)


ashfurrow (approved):
Nice! For anyone else reviewing, it's helpful to ignore whitespace. You can do that at the following: https://github.com/tootsuite/mastodon/pull/2603/files?w=1

app/views/about/terms.en.html.haml
```diff
@@ -2,75 +2,78 @@
...
-
-  %p If we decide to change our privacy policy, we will post those changes on this page.
-
-  %p This document is CC-BY-SA. It was last updated May 31, 2013.
-
-  %p
-    Originally adapted from the
-    = succeed '.' do
-      = link_to 'Discourse privacy policy', 'https://github.com/discourse/discourse'
+  - if @instance_presenter.site_terms.blank?
```
Gargron: The issue here is that you are modifying `terms.en.html.haml` only (emphasis on "en"). Something must be shuffled around to make it `terms.html.haml` (locale-agnostic) and move the default text somewhere else. Either make the default value of the `site_terms` setting the English version of this in config/settings.yml, or put the whole text into a YML locale file (haml markup has to become html in both cases, of course)

Gargron (changes requested):

Issue #2601, opened 2017-04-28
Video player play button off center


mkosler:
The play button is off center for videos. 

![vid_player_bug](https://cloud.githubusercontent.com/assets/1159095/25549048/3d9db98a-2c35-11e7-84ba-35b82ed6c400.PNG)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Gargron:
@stephenburgess8 could this be related to #2338?

PR #2608, opened 2017-04-29
fix(video): Position of play icon

stephenburgess8:
Before:
screen shot 2017-04-28 at 11 41 42 pm

After:
screen shot 2017-04-28 at 11 41 49 pm

#2601

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -3133,6 +3133,7 @@ button.icon-button.active i.fa-retweet {
   border-radius: 100px;
   color: rgba(255, 255, 255, 0.8);
   font-size: 36px;
+  left: 50%;
   padding: 5px;
   position: absolute;
   top: 50%;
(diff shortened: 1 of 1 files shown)


ashfurrow (approved):

Gargron:
In the screenshot, the overlay icon seems to be 0px/0px away from border, but should be 4px/4px - is this a browser issue or another small regression?

stephenburgess8:
Good eye. This was one feature I missed in testing (video previews) . I think there may need to be some fixes to video playing in a post (not expanded). As well.

stephenburgess8:
@Gargron Two things added here -

1) I had an expand icon regression where it was positioned top rather than bottom.

2) I was noticing that I was letting a letterboxing effect with videos that were widescreen / landscape. AFAICT this was a preexisting visual bug. In the final commit here I changed `transform: translateY(-50%)` to `transform: translateY(-35%)` on the `status__video-player-video` class. If there's another reason not to do this we can just toss that commit, but I think it looks better. Here's two examples from different instances.

Before:
screen shot 2017-04-29 at 9 06 26 am

After:
screen shot 2017-04-29 at 9 05 31 am

Before:
screen shot 2017-04-29 at 9 07 05 am

After:
screen shot 2017-04-29 at 9 07 33 am

Gargron (approved):

PR #2609, opened 2017-04-29
Do not encode JSON object payloads to JSON strings (breaking change)

akihikodaki:
This is a breaking change for streaming API.

### DIFF
--- a/app/javascript/mastodon/containers/mastodon.js
+++ b/app/javascript/mastodon/containers/mastodon.js
@@ -237,13 +237,13 @@ class Mastodon extends React.Component {
       received (data) {
         switch(data.event) {
         case 'update':
-          store.dispatch(updateTimeline('home', JSON.parse(data.payload)));
+          store.dispatch(updateTimeline('home', data.payload));
           break;
         case 'delete':
           store.dispatch(deleteFromTimelines(data.payload));
           break;
         case 'notification':
-          store.dispatch(updateNotifications(JSON.parse(data.payload), getMessagesForLocale(locale), locale));
+          store.dispatch(updateNotifications(data.payload, getMessagesForLocale(locale), locale));
           break;
         }
       },
--- a/app/javascript/mastodon/features/community_timeline/index.js
+++ b/app/javascript/mastodon/features/community_timeline/index.js
@@ -54,7 +54,7 @@ class CommunityTimeline extends React.PureComponent {
       received (data) {
         switch(data.event) {
         case 'update':
-          dispatch(updateTimeline('community', JSON.parse(data.payload)));
+          dispatch(updateTimeline('community', data.payload));
           break;
         case 'delete':
           dispatch(deleteFromTimelines(data.payload));
--- a/app/javascript/mastodon/features/hashtag_timeline/index.js
+++ b/app/javascript/mastodon/features/hashtag_timeline/index.js
@@ -28,7 +28,7 @@ class HashtagTimeline extends React.PureComponent {
       received (data) {
         switch(data.event) {
         case 'update':
-          dispatch(updateTimeline('tag', JSON.parse(data.payload)));
+          dispatch(updateTimeline('tag', data.payload));
           break;
         case 'delete':
           dispatch(deleteFromTimelines(data.payload));
--- a/app/javascript/mastodon/features/public_timeline/index.js
+++ b/app/javascript/mastodon/features/public_timeline/index.js
@@ -54,7 +54,7 @@ class PublicTimeline extends React.PureComponent {
       received (data) {
         switch(data.event) {
         case 'update':
-          dispatch(updateTimeline('public', JSON.parse(data.payload)));
+          dispatch(updateTimeline('public', data.payload));
           break;
         case 'delete':
           dispatch(deleteFromTimelines(data.payload));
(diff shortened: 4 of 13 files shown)


ashfurrow:
Hi there @akihikodaki – thanks for the pull request! Does this resolve any GitHub issues? What are the implications of merging the PR? Any additional details would be helpful – thanks again!

akihikodaki:
No, there is no related issue.

The current implementation gives JSON something like this for `update` and `notification` event:
```JSON
{"payload":"{\"id\":1,\"created_at\":\"2000-01-01T00:00:00.000Z\",\"in_reply_to_id\":null,\"in_reply_to_account_id\":null,\"sensitive\":false,\"spoiler_text\":\"\",\"visibility\":\"public\",\"application\":{\"name\":\"Web\",\"website\":null},\"account\":{\"id\":1,\"username\":\"admin\",\"acct\":\"admin\",\"display_name\":\"\",\"locked\":false,\"created_at\":\"2000-01-01T00:00:00.000Z\",\"followers_count\":0,\"following_count\":0,\"statuses_count\":0,\"note\":\"\",\"url\":\"https://cb6e6126.ngrok.io/@admin\",\"avatar\":\"https://cb6e6126.ngrok.io/headers/original/missing.png\",\"avatar_static\":\"https://cb6e6126.ngrok.io/headers/original/missing.png\",\"header\":\"https://cb6e6126.ngrok.io/headers/original/missing.png\",\"header_static\":\"https://cb6e6126.ngrok.io/headers/original/missing.png\"},\"media_attachments\":[],\"mentions\":[],\"tags\":[],\"uri\":\"tag:cb6e6126.ngrok.io,2000-01-01:objectId=1:objectType=Status\",\"content\":\"<p>hello, world</p>\",\"url\":\"https://cb6e6126.ngrok.io/@admin/1\",\"reblogs_count\":0,\"favourites_count\":0,\"reblog\":null,\"favourited\":false,\"reblogged\":false}","event":"update"}
```
Note that `payload` is a JSON object encoded as a JSON string, which is redundant and inconsistent with `delete` event, which simply gives a JSON number without encoding as a JSON string.
```JSON
{"payload":1,"event":"delete"}
```
This change fixes those problems to ease the client side, but the existing clients will be broken.

ashfurrow:
I see, okay. We'd definitely need to document this change and communicate out to app developers. Not sure the best way to do that, anyone have suggestions?

Is there any way to make this backwards-compatible with older clients maybe?

akihikodaki:
> Is there any way to make this backwards-compatible with older clients maybe?

I have another suggestion to have a new convention for a streaming API. It may be possible to distinguish the old and new clients by testing whether they are conforming to the convention.

It is to change endpoints for WebSocket. The current implementation uses `stream` query to know which resource the client wants (e.g. `/api/v1/streaming?stream=user`), but I think it is more reasonable to have the parameter in the path (e.g. `/api/v1/strereaming/user`), considering the semantics of query and path. Also note that he event stream API also uses URLs like `/api/v1/strereaming/user`.

If we decide to change the endpoints, I will suggest to have more changes. I have already came up with the idea of having `Authorization` header instead of `access_token` query and the one of returning 404 status instead of silently closing WebSocket if the URL is wrong.

Though I am a bit reluctant to make such a big change to the existing implementation since I have already made [a new implementation in Go](https://github.com/kagucho/mastodon/tree/gostreaming), I will do that if you think it is desired.

Gargron:
@akihikodaki I'm extremely interested in your Go implementation! JavaScript can be a rightful mess to debug, and Go binaries are really nice and clean.

That being said it is not possible to set headers on a WebSocket connection from a browser. This is why the access token is set via a query string instead - I had no other choice.

akihikodaki:
> I'm extremely interested in your Go implementation! JavaScript can be a rightful mess to debug, and Go binaries are really nice and clean.

Do you think there is possibility to drop the current Node.js implementation? If so, I would make a pull request to replace it with the implementation in Go, and then updates it to have the new API.

> That being said it is not possible to set headers on a WebSocket connection from a browser. This is why the access token is set via a query string instead - I had no other choice.

Thank you for clarifying. I have had a look at the RFC of WebSocket, but I have not seen the Web API.

akihikodaki:
> the one of returning 404 status instead of silently closing WebSocket if the URL is wrong.

I have forgot it doesn't matter since it is irrelevant to endpoints.

beatrix-bitrot:
I would really appreciate this, especially for debugging. What do we need to do to get it merged? @Gargron

Gargron:
@beatrix-bitrot It's a breaking change. It will break all existing clients that use this. The benefit to this is kinda small really. Decoding JSON once or twice is not a performance bottleneck imo.

beatrix-bitrot:
@Gargron yeah that makes sense. I guess I just wonder like... should it stay this way forever then? or can it be scheduled to be included in some future release far in advance so all client authors will know about it?

I'm in favour of the latter. If it will stay this  way forever though (which would be a bit silly IMO), then should this PR be closed? Best not to let these things linger without resolution.

akihikodaki:
Recently I'm convinced with an idea to have a new API (`/api/v1/strereaming/user`) rather than breaking the existing one (`/api/v1/streaming?stream=user`). If it is preferred, I will close this pull request and prepare another.

beatrix-bitrot:
hm, before making a new PR maybe open an issue and describe your plan, then we can get feedback from more people including app developers before beginning development. I think once you open such an issue you can close this PR and refer to it in the issue.

akihikodaki:
Opened #3049.

Issue #1742, opened 2017-04-14
CLI utility to check coverage of a locale


yookoala:
It is currently quite hard to keep track on the translations. Strings spread across source code of Ruby and Javascript. It is hard for a translator to tell if a language translation is completed.

I propose we can add some CLI utility that will compare a given locale with the English (presume the English translation file is complete coverage). Then translator can have a reference how complete their work is.

I am willing to write such tool. But what language should I use? Ruby or Nodejs?

Also, where should I put such script?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mjankowski:
Running `i18n-tasks health` gets some of this.

Gargron:
`i18n-tasks health` is for this, although it only deals with Ruby localizations. The react-intl story is more complex. There are packages that are supposed to make managing translations easier, but they do not work with Browserify, or at least I could not get them to work. So it's connected to #1712

thor-the-norseman:
The guides should mention the tools that are available for this. I'm having a hell of a time keeping the translation up to date. Things are not only being added, but they are also being changed. I'm able to catch all the _additions_ so far by using a YAML merge tool I found, but I'm missing out on any _changes_ to existing strings, which can be a huge problem. Does the Ruby tool you mention deal with changes? I'm thinking it can't really be solved unless a diff can be performed against different commits of the *en.yml and en.jsx files to check for that.

thor-the-norseman:
As for JSX I'm temporarily converting the files to JSON and using a JSON merge tool on them to catch new strings from the English version. However, this will, as with the YAML merge, overwrite changes to existing strings. Some kind of 3-way merge is needed, where Git commit IDs can be specified or such, that runs on all the files and produces a set of proposed files I can finalize. It should take the changes to the English version and apply them to the localized version so that a mixed file is produced that I can finalize through editing.

thor-the-norseman:
The YAML is probably easy to parse. The JSX files are a problem. Why not use YAML? I mean, I'm sure there are 934895834 YAML parsers that could run in a browser at this point.

thor-the-norseman:
(JSON is another possibility, but the syntax is strict, and no comments are allowed.)

yookoala:
[browserify](http://browserify.org/), and hence [browserify_rails](https://github.com/browserify-rails/browserify-rails) does not have ability to support bundling files other than js.

[...]

PR #2465, opened 2017-04-26
Upgrade react-rails to v2.x

ykzts:
Sorry! There are many file changes. But, I just added `import React from 'react';`.

Also, never use jQuery in React Components. I deleted `//= require jquery`.

### with jQuery

```shell
$ du -sk public/assets/application-*.js*
2892	public/assets/application-acd2b42d93ab83e0fae21a51767e530d17ed4e0c2d9d0c3bd839cae162d48a67.js
496	public/assets/application-acd2b42d93ab83e0fae21a51767e530d17ed4e0c2d9d0c3bd839cae162d48a67.js.gz
```

### without jQuery

```shell
$ du -sk public/assets/application-*.js*
2776	public/assets/application-3a7e5f14a72c345ca3c138fefd79c813612c94afb5140d753428d0bb32225968.js
456	public/assets/application-3a7e5f14a72c345ca3c138fefd79c813612c94afb5140d753428d0bb32225968.js.gz
```

PR #2617, opened 2017-04-29
Replace sprockets/browserify with Webpack

### LINKED ISSUE #1712: Migrate from Browserify to Webpack

Gargron:
Fix #1712, #1742

What this toolchain upgrade provides:

- Live reloads during development
- Friendlier asset compilation output (e.g. progress, final bundle size)
- Code splitting, async chunk loading, smarter minifying
- Automated management of JS translations

Untested potential benefits:

- Smaller bundle size?

TODO:

- [ ] No more best_in_place for admin site settings - needs replacement
- [ ] No more local_time JS - needs integration or replacement
- [x] Make custom.scss work again

Development implications:

- Run `./bin/webpack-dev-server` at the same time as `rails s`
- Run `yarn run manage:translations` to get a list of all missing/redundant translations in the web UI


ykzts:
@Gargron Great change!

But, `application.js` is depends on jQuery. see #2465

Gargron:
@ykzts Unfortunately, jQuery is still a dependency because of jquery-ujs which makes some non-GET links work. There is rails-ujs which is jQuery-free but it didn't work with webpack. jQuery is also used in public.js. On the upside, webpack splits the code into vendor.js and application.js/public.js bundles, so dependencies are not loaded twice.

ykzts:
@Gargron Uh... As far as I see, `jquery-ujs` is not used in the part where React is used. Even if you delete it, `application.js` will run with the exception of the notification.

Gargron:
The web UI has one logout button which needs jquery-ujs

ykzts:
Oh... I see. Certainly. I'm sorry.

Gargron:
I don't understand why view specs are failing with this cryptic error. I changed nothing about them...

config/webpack/production.js
```diff
@@ -0,0 +1,46 @@
+// Note: You must restart bin/webpack-dev-server for changes to take effect
+
+/* eslint global-require: 0 */
+
+const webpack = require('webpack')
+const merge = require('webpack-merge')
+const CompressionPlugin = require('compression-webpack-plugin')
+const BabiliPlugin = require('babili-webpack-plugin')
```
nolanlawson: I'm not sure it's worthwhile to use both Babili and Uglify; AFAIK Uglify still produces smaller code than Babili, although the downside is that it only works on ES5, so e.g. if we wanted a pure ES6 build then that's the point where Babili would be very useful.

config/webpack/production.js
```diff
@@ -0,0 +1,46 @@
...
+        drop_debugger: true,
+        dead_code: true,
+        pure_getters: true,
+        negate_iife: true,
+        conditionals: true,
+        loops: true,
+        cascade: true,
+        keep_fargs: false,
+        warnings: true
+      },
```
nolanlawson: Just setting `compress` to `true` should be fine, unless there' is something in particular you wanted to tweak here?

config/webpack/production.js
```diff
@@ -0,0 +1,46 @@
...
+
+      mangle: false,
```
nolanlawson: I don't often see `mangle` set to `false` FWIW, would this break something to turn on?

config/webpack/production.js
```diff
@@ -0,0 +1,46 @@
...
+      },
+
+      sourceMap: false
+    }),
+    new CompressionPlugin({
+      asset: '[path].gz[query]',
+      algorithm: 'gzip',
+      test: /\.(js|css|svg|eot|ttf|woff|woff2)$/
+    })
+  ]
```
nolanlawson: It'd be a good idea to also use the `DefinePlugin` to ensure that `NODE_ENV` is set to `production` which cuts out a lot of development-only code in React as well as other libraries ([React docs on the subject](https://facebook.github.io/react/docs/optimizing-performance.html)). I also opened up a PR for doing the equivalent thing in Browserify FWIW: https://github.com/tootsuite/mastodon/pull/2635

nolanlawson (approved):

config/webpack/production.js
Gargron: Was implementing suggestions from this comment: https://github.com/tootsuite/mastodon/issues/1712#issuecomment-296016346

config/webpack/production.js
Gargron: rails/webpacker sets NODE_ENV=production in bin/webpack

config/webpack/production.js
Gargron: [This article](https://blog.craftlab.hu/how-to-do-proper-tree-shaking-in-webpack-2-e27852af8b21) claimed that babili was preferable to uglify because it could understand the new syntaxes better.

nolanlawson: Yeah so that doesn't really apply to tree-shaking; tree-shaking is about bundling ES modules, which is handled by Webpack 2 before the rest of the ES6-to-ES5 transpilation. You can test it, but I'm fairly certain Uglify alone (with mangle+compress) will still beat Babili, assuming you're outputting ES5 code. (In fact it looks like [Babili's own benchmarks](https://github.com/babel/babili#benchmarks) still show Uglify winning, although I know they're making progress.)

config/webpack/production.js
nolanlawson: I dunno, most every JS project I know of just uses uglify with `mangle` and `compress` set to `true`. You can tweak the compress settings in rare cases, but in my mind "mangle/compress" (aka `uglifyjs -mc`) is basically Uglify's "default" setting. :)

config/webpack/production.js
nolanlawson: Huh that's odd, React's docs still say to add DefinePlugin, but maybe it's no longer necessary for Webpack 2? Just checked the output bundles after `asset:precompile` and I'm not seeing any `NODE_ENV` references, so LGTM :+1:

nolanlawson:
FWIW, just tested the bundle size, and before this PR `application.js` is 2.97 MB, afterwards `application.js` is 563 kB and `vendor.js` is 1.57 MB, so combined they're 2.14 MB which is a huge reduction. Very nice! :tada:

config/webpack/production.js
Gargron: Afaik it would make trying to debug production errors way harder..?

Gargron:
Travis keeps breaking because node-sass needs some binaries that apparently don't get cached by Travis?

Gargron:
Here are upgrade notes:

* Ensure you have Node 6.x and not less
* Install bundle/yarn dependencies
* If it complains about node-sass, `npm rebuild node-sass --force`
* Make sure `CDN_HOST` contains protocol part, if you are using it
* Precompile, restart

kyzh:
I'm not sure if you are happy for people to provide feedback here.
Feel free to tell me off if you don't think it i appropriate.

This branch is live on glitch.social (thanks Bea).
I've been trying for a while and there is noticeable improvement.
It definitely helps with recent issues I had on mobile  #2648

config/webpack/production.js
nolanlawson: A good solution for that is sourcemaps. :)

Gemfile
```diff
@@ -58,18 +55,18 @@ gem 'sprockets-rails', require: 'sprockets/railtie'
 gem 'statsd-instrument'
 gem 'twitter-text'
 gem 'tzinfo-data'
+gem 'webpacker', github: 'rails/webpacker'
```
mjankowski: Using the `github:` style here produces a security warning on bundle install -- using `https://....` will fix it.

But also, I think this is on rubygems now: https://rubygems.org/gems/webpacker/versions/1.2

app/javascript/styles/basics.scss
```diff
@@ -1,6 +1,6 @@
 body {
   font-family: 'Roboto', sans-serif;
-  background: $color1 image-url('background-photo.jpg');
+  background: $color1 url('../images/background-photo.jpg');
```
mjankowski: It seems weird that the stylesheets move into `app/javascript/...` -- is that normal for webpack?

app/javascript/styles/fonts/montserrat.scss
```diff
@@ -0,0 +1,11 @@
+@font-face {
```
mjankowski: Same question here ... it seems weird that the fonts move into `app/javascript/...` -- is that normal?

config/initializers/assets.rb
```diff
@@ -8,6 +8,6 @@
 
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-Rails.application.config.assets.precompile += %w(application_public.js custom.css)
+# Rails.application.config.assets.precompile += %w(application_public.js custom.css)
```
mjankowski: is this left commented out on purpose, or should be a delete?

mjankowski:
Should we add a `Procfile.dev` (for people who use foreman in dev) or `bin/server` or some other way to include the running of the `webpack-dev-server` for people to get all in one process alongside the rails server and workers?

config/initializers/assets.rb
Gargron: It's commented out by default so I left it commented out instead of deleting

app/javascript/styles/fonts/montserrat.scss
Gargron: It's not normal but if I leave anything in assets it will get picked up by sprockets, I think, and then it would get processed twice.

Gemfile
Gargron: Okay. Just did what the webpacker readme recommended. Will gladly pin the version.

mjankowski: Yeah ... I think this was only possible once Rails 5.1 came out, and before that you DID need to pull from GH repo

config/webpack/production.js
nightpool: if mastodon published sourcmaps in prod I would be *all about* that. I personally like having mangle off otherwise though.

app/javascript/styles/fonts/montserrat.scss
mjankowski: I guess I would have assumed that the JS moves to app/javascript but anything which really is still just a static asset either a) stays in app/assets (and is made available to the JS, where needed? I think webpack does this...?), or b) moves to like - `app/fonts` or something?

The thing I'm asking about here isn't the move away from sprockets and to webpack, but about the naming convention of having fonts and images in an `app/javascript` dir...

Gargron: It isn't too weird really, as some React practices involve bundling CSS modules directly with components, e.g. CSS and JS side by side, with any relevant images included. We don't use that but it's still all close together.

mjankowski (approved):
I'm approving this on the assumption that we continue to find edge cases and work out best practices once the bulk of it is in.

nolanlawson:
On second analysis, it looks like this didn't shrink the bundle size much. :confused:

```
796K	application-6d6d82d4e4776ad9b07e.js
8.0K	public-39f8a254bfe861a34311.js
2.0M	vendor-5e5fbf294d06423aa79e.js
```

Looks like we're still at around 2.8MB of JavaScript total. I've run [webpack-bundle-analyzer](https://github.com/th0r/webpack-bundle-analyzer) on the latest master branch (4d22d03fabc5f249fd2a93037916b980e4d3f99b); here's [a live view of the data](http://bl.ocks.org/nolanlawson/raw/96df40358b82dec05e07b5e53a2740df/). A few things jump out at me:

- emojione takes up a ton, hopefully https://github.com/Ranks/emojione/pull/489 will get merged soon
- all the various language JSON files take up a lot inside of Mastodon itself
- `Buffer` is unnecessary; it's required by an unused API in `http-link-header`
- Lodash takes up a bit, worth looking into `lodash-webpack-plugin` and `babel-plugin-lodash`

Mostly I don't see any easy wins here; just need to assiduously go through dependencies and try to remove what's not needed, replace where possible, etc. Code-splitting would also be good. Keep in mind that Sean Larkin from the Webpack team recommends [<300KB (without gzip)](https://twitter.com/TheLarkInn/status/795782215288254464) for each chunk, whereas Addy Osmani on the Chrome team recommends [<100KB to hit <3 seconds on 3G](https://github.com/webpack/webpack/issues/3216).

Dockerfile
```diff
@@ -32,14 +30,14 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
     imagemagick@edge \
     ca-certificates \
  && npm install -g npm@3 && npm install -g yarn \
- && bundle install --deployment --without test development \
- && yarn --ignore-optional --pure-lockfile \
- && yarn cache clean \
```
hinaloe: Why remove `yarn cache clean` task? for https://github.com/rails/webpacker/issues/405 ?

PR #2627, opened 2017-04-29
More coverage yes more even more

mjankowski:


app/controllers/admin/confirmations_controller.rb
```diff
@@ -2,17 +2,15 @@
...
+      account_user.confirm
       redirect_to admin_accounts_path
     end
 
     private
 
-    def set_account
-      @account = Account.find(params[:account_id])
+    def account_user
+      Account.find(params[:account_id]).user || raise(ActiveRecord::RecordNotFound)
```
mjankowski: This fixes a scenario where a `nil` user (account w/out user) would raise when trying to confirm.

app/controllers/admin/domain_blocks_controller.rb
```diff
@@ -27,7 +27,7 @@ def show
 
     def destroy
       @domain_block = DomainBlock.find(params[:id])
-      UnblockDomainService.new.call(@domain_block, resource_params[:retroactive])
+      UnblockDomainService.new.call(@domain_block, retroactive_unblock?)
```
mjankowski: This was previously passing "1" or "0" instead of a boolean.

Gargron (approved):

Issue #1758, opened 2017-04-14
Rubocop config not matching Code Climate?


ineffyble:
I've seen several reports that errors being detected by Code Climate aren't showing up when people run Rubocop locally - does the Rubocop YML need to be updated?

Gargron:
Code Climate should be using the same rubocop config file. But Code Climate also has other checkers.. code duplication, something with JS, etc. Maybe code duplication checks should be disabled, they haven't be terribly useful.

expenses:
I emailed codeclimate support about this and got a nice reply:

> Hi there,
>
> Thanks for reaching out and I'm sorry for the trouble you're having with those version differences for rubocop.
>
> It's definitely our goal to always stay up to date with the most recent versions that are available for our engines.
>
> I don't have an ETA on when we'll update to these versions, but I can assure you it's something that's on our roadmap. If you'd like I'd be more than happy to reach out once these newer versions are available -- in fact, I've gone ahead and flagged your email to remind me to do so once support is available.
>
> That said I'm sorry for the limitation this is causing in the meantime. Thanks for the nudge on this, and I'll be in touch as soon as I can with updates on that.

PR #2628, opened 2017-04-29
Downgrade rubocop 0.48.1 => 0.46.0

yhirano55:
### Problems

- [Code climate use rubocop version `0.46.0`](https://docs.codeclimate.com/v1.0/docs/rubocop), but this project use `0.48.1`. 
- Different versions, different warnings, so I think we should use the same version.

### Resolved

- Downgraded rubocop version to `0.46.0` from `0.48.1`.
- Add `vendor/**/*` to Excludes of Allcops in `.rubocop.yml`.
- Fix several styles:
    - Added `frozen_string_literal`.
    - Changed percent literal delimiter from `[]` to `()`.
    - Modified Alignment.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -88,3 +88,4 @@ AllCops:
   - 'Rakefile'
   - 'node_modules/**/*'
   - 'Vagrantfile'
+  - 'vendor/**/*'
--- a/Capfile
+++ b/Capfile
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'capistrano/setup'
 require 'capistrano/deploy'
 require 'capistrano/scm/git'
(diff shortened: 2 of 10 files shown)


expenses:
Also see #1758. I personally think that we should keep it as-is and wait for code climate to update their version.

yhirano55:
Thanks for your reviewing.

It's problem that we may not pass checking on code climate because the rubocop version is different (e.g. case percent literal delimiter, when 0.48.1 we should use `[]` as default, when 0.46.0 we should use `()` as default)  Of Course, we can define rule. But I personally think that it's so troublesome (While comparing differences in version...)

I think easy to upgrade rubocop version to `>= 0.46.0`. So I think we had better to downgrade temporally rubocop version to `0.46.0` until code climate have upgraded to `>= 0.46.0`.

Gargron (approved):

PR #2631, opened 2017-04-30
Fix "Could not find actionview-5.0.2"

shinriyo:
```
==> default: /home/vagrant/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bundler-1.14.6/lib/bundler/spec_set.rb:87:in `block in materialize'
==> default: :
==> default: Could not find actionview-5.0.2 in any of the sources
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,7 @@ gem 'whatlanguage'
 gem 'react-rails'
 gem 'browserify-rails'
 gem 'autoprefixer-rails'
+gem 'actionview', '>= 5.0.2'
 
 group :development, :test do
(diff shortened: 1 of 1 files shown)


mjankowski:
can you clarify what this solves? ... that gem should be pulled in via dependencies of other gems...

mjankowski (changes requested):
can you clarify what this solves? ... that gem should be pulled in via dependencies of other gems...

Gargron:
I don't see how this could be necessary given that actionview is part of rails.

PR #2635, opened 2017-04-30
Use envify + NODE_ENV=production for smaller JS bundle

nolanlawson:
This uses [envify](https://github.com/hughsk/envify) as a global Browserify transform to replace usages of `process.env.NODE_ENV` with `"production"` when `RAILS_ENV` is also set to `"production"`, which allows Uglify to remove non-essential code. It also makes it so `react-addons-perf` is omitted in production mode, since it's not worthwhile in production (it just instruments `performance.mark`/`performance.measure`s in React's APIs).

Before this PR, `application.js` is 2.96 MB (2963246 bytes to be exact, 514 kB gzipped). After, it's 2.95 MB (2952387 bytes to be exact, 512 kB gzipped).

### DIFF
--- a/app/assets/javascripts/components.js
+++ b/app/assets/javascripts/components.js
@@ -3,7 +3,11 @@
 
 window.React    = require('react');
 window.ReactDOM = require('react-dom');
-window.Perf     = require('react-addons-perf');
+
+if (process.env.NODE_ENV !== 'production') {
+  // perf tracking (aka marks/measures) not needed in production mode
+  window.Perf = require('react-addons-perf');
+}
 
 if (!window.Intl) {
   require('intl');
--- a/config/application.rb
+++ b/config/application.rb
@@ -74,7 +74,7 @@ module Mastodon
     config.middleware.use Rack::Deflater
 
     # babel config can be found in .babelrc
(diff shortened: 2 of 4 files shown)


nolanlawson:
Just double-checked, and this also correctly uses `--NODE_ENV development` when running in dev mode (e.g. `rails s`).

Wonderfall:
Shouldn't `envify` be also added to `Dependecies` in `package.json`?

nolanlawson:
Yes, I added it to the `devDependencies`.

Wonderfall:
Yes but it fails when NODE_ENV is set to production.

nolanlawson:
What do you mean "it fails"? In this PR I'm not even setting `NODE_ENV`; I'm just explicitly passing it to `envify` which achieves the same thing.

nolanlawson:
Ah wait, I forgot to update `yarn.lock`. Will fix.

nolanlawson:
Note that https://github.com/tootsuite/mastodon/pull/2617 would make this PR obsolete.

Gargron:
As foretold, obsolete through #2617 :P

PR #2641, opened 2017-04-30
Minor changes in email notifications and Greek Translation

saggel:
I submit some minor editing on the english confirmation instructions -
mostly as a test before I submit my Greek translation!


app/views/user_mailer/password_change.en.html.erb
```diff
@@ -1,3 +1,3 @@
 <p>Hello <%= @resource.email %>!</p>
 
-<p>We're contacting you to notify you that your password on Mastodon has been changed.</p>
+<p>We're contacting you to notify you that your password on <%= @instance %> has been changed.</p>
```
mjankowski: Can you put this change in a separate PR?

config/application.rb
```diff
@@ -30,6 +30,7 @@ class Application < Rails::Application
       :ar,
       :bg,
       :de,
+	  :el,
```
mjankowski: alignment?

config/locales/doorkeeper.el.yml
```diff
@@ -0,0 +1,113 @@
+---
+el:
```
mjankowski: did anything in this file change from en?

config/locales/el.yml
```diff
@@ -0,0 +1,341 @@
+---
+el:
```
mjankowski: did anything in this file change from en?

mjankowski (changes requested):

config/locales/el.yml
saggel: The Greek code is el - did I do something wrong?

config/locales/doorkeeper.el.yml
saggel: The Greek code is el - did I do something wrong?

config/application.rb
saggel: I can fix this on the next pull where I'll complete the translation!

config/locales/el.yml
mjankowski: It looks like nothing change in the file...

The `en` changed to `el`, which is correct ... but it looks like nothing was translated?

app/views/user_mailer/password_change.en.html.erb
saggel: If you tell me how :-)

saggel:
If you think I should just do the pull again right from the beginning please do let me know! It was my first attempt!

config/locales/el.yml
saggel: Yes that's true - not done yet, but I thought it needs to be on the folder for the locale to work - the translation is done on 85% and I wanted to try it on an instance before completing it!

mjankowski:
Congrats on your first PR and thanks for the help! ... 

- The formatting changes to the mailers look good
- I'm not sure the change which adds `@instance` to the password change mailer will work, because I don't think that variable is in place in that mailer...
- I think the translations will just fall back to the `en` versions if some are missing

I think my preference would be to do one PR for the mailer changes, and another for all of the added `el` locale additions (including el mailers).

If you're working purely through the GH website and not a local copy, I can help create the mailer PR separately, and then we can update this one to only include the el changes.

saggel:
Awesome - let me know what I need to do - I use the github desktop app - but happy to do whatever is easier or faster for you!

yookoala:
Your works is based on an older version of Mastodon. Can you please rebase your work on the latest master branch? Or at least merge with the latest Mastodon's master?

Gargron:
Closing for inactivity

PR #2644, opened 2017-04-30
Updating Indonesian locale

virtualpain:
update and fix id.jsx, id.yml, doorkeeper.id.yml strings
add activerecord.id.yml


config/locales/activerecord.id.yml
```diff
@@ -0,0 +1,12 @@
+en:
```
mjankowski: should be `id` ?

mjankowski (changes requested):

config/locales/activerecord.id.yml
virtualpain: my bad, fixed in 6e8dac8

mjankowski (approved):

PR #2664, opened 2017-04-30
Why my polish translate is not change?

xor256:
I have trouble with polish translations. Some working and ... some no.
http://bankfotek.pl/view/2074460


kayleej:
Looks like this should be an issue instead of a PR, no?

Gargron:
This PR looks accidental. Closing.

Issue #1635, opened 2017-04-12
Column filters do not search content warning text


mkosler:
When I add a regex filter on a column, it does not search the content warning text.

Here is what it looks like when I am NOT filtering the word "hippo":

![hippo_filter_off](https://cloud.githubusercontent.com/assets/1159095/24973662/3216d8a6-1f85-11e7-9ad1-fc242f599ea0.PNG)

Here is what it looks like when I am filtering the word "hippo":

![hippo_filter_on](https://cloud.githubusercontent.com/assets/1159095/24973611/0c5134cc-1f85-11e7-8191-8dbf7d397719.PNG)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mkosler:
This has been brought up again today after people started posting a meme that used the content warning part as the main part of the post. You couldn't avoid it on your timeline.

![hemingway](https://cloud.githubusercontent.com/assets/1159095/25295889/38169dfe-26ab-11e7-9ecf-a0019e7aab24.PNG)

rtucker:
I have confirmed this issue is still present in 1.3.2.

Steps to reproduce in Vagrant environment:

1) Toot a message with a word in the CW that doesn't appear in the body
2) Attempt to filter out the CW word
3) See the toot not disappear
4) Attempt to filter out a word in the non-CW part
5) See the toot disappear properly

![screenshot_2017-04-30_16-11-40](https://cloud.githubusercontent.com/assets/97437/25567616/da98d0cc-2dbf-11e7-8158-e5acda26266f.png)

![screenshot_2017-04-30_16-12-17](https://cloud.githubusercontent.com/assets/97437/25567625/018eedf6-2dc0-11e7-9d78-2bff5a1c98e9.png)

PR #2665, opened 2017-04-30
ui: check spoiler_text against regex filter (#1635)

rtucker:
When filtering by regex, also check the spoiler_text if present.  This fixes issue #1635.

### DIFF
--- a/app/javascript/mastodon/features/ui/containers/status_list_container.js
+++ b/app/javascript/mastodon/features/ui/containers/status_list_container.js
@@ -26,10 +26,7 @@ const makeGetStatusIds = () => createSelector([
     try {
       if (showStatus) {
         const regex = new RegExp(columnSettings.getIn(['regex', 'body']).trim(), 'i');
-        showStatus = !regex.test(statusForId.get('reblog') ? statuses.getIn([statusForId.get('reblog'), 'unescaped_content']) : statusForId.get('unescaped_content'));
-        if (showStatus && statusForId.get('spoiler_text').length > 0) {
-          showStatus = !regex.test(statusForId.get('reblog') ? statuses.getIn([statusForId.get('reblog'), 'unescaped_spoiler_content']) : statusForId.get('unescaped_spoiler_content'));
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/features/ui/containers/status_list_container.jsx
```diff
@@ -27,6 +27,9 @@ const makeGetStatusIds = () => createSelector([
       if (showStatus) {
         const regex = new RegExp(columnSettings.getIn(['regex', 'body']).trim(), 'i');
         showStatus = !regex.test(statusForId.get('reblog') ? statuses.getIn([statusForId.get('reblog'), 'unescaped_content']) : statusForId.get('unescaped_content'));
+        if (showStatus && statusForId.get('spoiler_text').length > 0) {
+          showStatus = !regex.test(statusForId.get('reblog') ? statuses.getIn([statusForId.get('reblog'), 'unescaped_spoiler_content']) : statusForId.get('unescaped_spoiler_content'));
```
nightpool: considering that we're now checking against four possible content locations, maybe it's better to just append and then test them all in the same string?

something like:

```js
let content = statusForId.get('unescaped_content') || statusForId.get('reblog.unescaped_content');
content +=  || "";
content += statusForId.get('unescaped_spoiler_content') || statusForId.get('reblog.unescaped_content');
content += statusForId.get('unescaped_spoiler_content') || statusForId.get('unescaped_spoiler_content');

showStatus = !regex.test(content);
```

Gargron: I was gonna suggest what @nightpool said, except just concatenate all of the searchable strings when creating `unescaped_content` in the reducer. In fact, I suggest renaming that field to "search_index" so it's more clear what it is used for

rtucker: ok, updated this in a3e9c95

Gargron (approved):

Issue #1010, opened 2017-04-05
CW should include the image, too


riking:
Follow-up to #239 and #460:

Statuses with a content warning should hide & show the image along with the text.

- [x] I looked through the JS code for a quick fix and did not find one

The CW is handled in StatusContent, while click-to-view images are handled in MediaGallery, and the two components are separated from each other: https://github.com/tootsuite/mastodon/blob/07b166a/app/assets/javascripts/components/components/status.jsx#L100-L102

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nevillepark:
See also https://github.com/tootsuite/mastodon/issues/715

Kimik0:
Agreed. Content warnings should apply to all content, text, images, and @ usernames.

As it is now, images need to be marked separately, and @ usernames are shown even when they're part of the hidden content (they should only show if they're at the very start to indicate a reply).

I understand there may be technical or historical reasons for the way things work now, but from a user POV it's just counter-intuitive.

blinry:
I think this is really unexpected behaviour at the moment. If I want to post a picture of a spider, naturally I'd add a CW "spider", and expect the image to be hidden.

lifning:
Yep. I've seen several people in the timelines now who clearly intended to hide an image with it, not realizing that it's a separate tag on the image upload itself, and so the thing they wanted to put a warning on is plainly visible, and some innocuous, incidental text got hidden instead.

blinry:
So, do we have a consensus that in a CW-labelled toot, all other contents (like images and mentions) should be hidden? :) Would be willing to whip together a PR.

Cassolotl:
@blinry I've seen someone on Mastodon saying that there is a case for keeping it the way it is: it's handy for hiding the URL of an image if you want to show the image and the comment but not the URL. Sort of a way to tidy it up.

However, I've also seen a post that says you can upload an image *and then delete the image URL from the toot* and the image stays attached to the toot. So that would cancel it out!

\*tests it\*

Yep, the image stays attached to the toot even if you remove the image URL from the toot. So yes, aside from that I've never seen an argument against putting the attached images behind the CW.

riking:
@Cassolotl that breaks federation, though - the image doesn't show up when the post is pulled.

Also, that usecase is a stupid trick to get around an oddity of the software. That's not really something worth preserving.

lifning:
[...]

PR #2666, opened 2017-04-30
status: if spoiler_text is present, treat media as sensitive (#1010)

rtucker:
In the event that there is a CW, go ahead and treat any attached media as sensitive.

This addresses UI issue #1010

### DIFF
--- a/app/controllers/api/v1/statuses_controller.rb
+++ b/app/controllers/api/v1/statuses_controller.rb
@@ -61,7 +61,7 @@ class Api::V1::StatusesController < ApiController
                                          status_params[:status],
                                          status_params[:in_reply_to_id].blank? ? nil : Status.find(status_params[:in_reply_to_id]),
                                          media_ids: status_params[:media_ids],
-                                         sensitive: status_params[:sensitive],
+                                         sensitive: status_params[:spoiler_text].blank? ? status_params[:sensitive] : true,
                                          spoiler_text: status_params[:spoiler_text],
                                          visibility: status_params[:visibility],
                                          application: doorkeeper_token.application,
--- a/app/assets/javascripts/components/components/status.jsx
+++ b/app/assets/javascripts/components/components/status.jsx
@@ -61,13 +61,15 @@ class Status extends React.PureComponent {
       );
     }
 
+    const sensitive = status.get('sensitive') || status.get('spoiler_text').length > 0;
+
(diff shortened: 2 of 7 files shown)


yiskah (approved):
This appears to be a needed non-breaking change. I don't see any syntax errors.

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -61,7 +61,7 @@ def create
                                          status_params[:status],
                                          status_params[:in_reply_to_id].blank? ? nil : Status.find(status_params[:in_reply_to_id]),
                                          media_ids: status_params[:media_ids],
-                                         sensitive: status_params[:sensitive],
+                                         sensitive: status_params[:spoiler_text].blank? ? status_params[:sensitive] : true,
```
Gargron: Wouldn't it be easier to just have this line, without any others? Here you're already setting sensitive=true when spoiler_text=present. No reason to keep checking spoiler_text beyond this point.

Gargron (changes requested):

app/controllers/api/v1/statuses_controller.rb
beatrix-bitrot: mmm, the additional checks i guess cause this to apply retroactively as well as going forward though.

could be nice. if it doesn't do any harm let it go that way?

nightpool:
I'm going to repost my objection from #1010:

> I believe I was the person @cassolotl mentioned, and yes, I believe that
keeping the affordance for hiding text content separate from the affordance
for hiding image content is a good thing. I think that there are multiple
use cases served by keeping them separate that unifying them would not.

> A good example is a longer discussion behind a cut, with images as an
attention grabber or hook. Another example is linking and citing sources
for an aesthetic post—they often take up multiple lines, so having them
behind a show more is preferable. Another one is the workaround already
mentioned.

I think that a much better way forward would have the JS *default* to having the media marked as sensitive if there is spoiler text present (both when you add the media to a post with existing spoiler text and when you add the spoiler text to existing media)

rtucker:
I've been running this patch for a few days now, and upon further consideration, I think I agree with @nightpool -- I've seen a number of cases where the implied-NSFW was not appropriate.  I am going to close this pull request in favor of moving the implied-NSFW to the UI side of things.

Issue #699, opened 2017-03-30
Hashtags in content warnings don't work


nevillepark:
Currently, hashtags in the content warning field just behave like plain text. [Here's an example.](https://mastodon.social/@nev/1372994) 

This is inconvenient if, for example, a user wants to hide their #introductions post behind a content warning, but wants other users to be able to find other #introductions posts by clicking on the hashtag. It also prevents the effective use of hashtags for things some people may want to filter out and others may want to follow, like #food or #discourse. 

![screenshot from 2017-03-30 16-01-55](https://cloud.githubusercontent.com/assets/16054979/24523780/3b7dcb34-1562-11e7-8082-5ca3ddbee3f5.png)

ProgVal:
If a hashtag is in the content of a toot, it appears in the search even if the toot has a content warning.
So I don't think it's an issue.

nevillepark:
Sorry, this is what I mean: https://mastodon.social/@nev/1529407

If the hashtag is *in* the content warning, it isn't clickable and it doesn't appear in search results.

ProgVal:
Indeed, I misunderstood.

However, since content warnings are not displayed on non-mastodon instances, allowing tags there would create a gap between mastodon and the rest

skddc:
> However, since content warnings are not displayed on non-mastodon instances, allowing tags there would create a gap between mastodon and the rest

Maybe that's a good thing. If it works and is being used and liked by users, then it could make adding it to the underlying protocols and getting it adopted by other implementations a bit easier. If it's already outside of specs, then we might as well improve the feature based on user feedback and needs. No?

nevillepark:
Hmm. I wonder if there is a way to maintain the CW functionality on Mastodon, but degrade gracefully for non-Mastodon instances, in the Atom feed, etc. E. g. ```CW: {cw field} | {status}```.

Gargron:
We have attempted to make content warnings contain rich text, including mentions, links and hashtags. However, this meant that content warnings could not be copied into new toots automatically any longer. It is simpler to think of them as a plain text label on top of the actual content.

sparr:
@Gargron "this meant that content warnings could not be copied into new toots automatically any longer"

why not?

sparr:
@ProgVal why aren't content warnings displayed on non-mastodon instances? Can they not just be included at the top of the post?

melissaboiko:
[...]

PR #2683, opened 2017-05-01
Handle hashtags in spoiler_texts (partial fix for #699)

rtucker:
These commits add partial support for hashtags in the spoiler_text (aka content warning header).

The first commit covers parsing of spoiler_texts for hashtags, allowing them to be found by hashtag search.  The second commit creates links for hashtags in spoiler_texts on the server-generated views.

An attempt was made to link spoiler_text hashtags from the main web UI, but JavaScript is not my forte.  I believe `app/assets/javascripts/components/components/status_content.jsx` is where that change would need to be made, however.


app/services/process_hashtags_service.rb
```diff
@@ -2,7 +2,9 @@
 
 class ProcessHashtagsService < BaseService
   def call(status, tags = [])
-    tags = status.text.scan(Tag::HASHTAG_RE).map(&:first) if status.local?
+    text = status.text
+    text << " " + status.spoiler_text unless status.spoiler_text.blank?
```
Gargron: This might fail due to immutable strings..?

rtucker: Plausible.  Is also relatively ugly due to rusty Ruby skills.  d6f97e0 may be better?

app/services/process_hashtags_service.rb
```diff
@@ -2,8 +2,7 @@
 
 class ProcessHashtagsService < BaseService
   def call(status, tags = [])
-    text = status.text
-    text << " " + status.spoiler_text unless status.spoiler_text.blank?
+    text = [status.text, status.spoiler_text].reject(&:empty?).join(' ')
```
beatrix-bitrot: this seems nicer yes

beatrix-bitrot (approved):
only issue from the last review was corrected so marking this approved

beatrix-bitrot (dismissed):
ack, i spoke too soon, this has conflicts that need to be resolved!

rtucker (commented):
Whoops!  I've fixed up the conflict in 9130073 but github does not seem to be recognizing it as such.  Hrmmph.

rtucker:
I have no idea how to indicate that the requested changes have been made, but they have been (by resolving the conflict and updating the branch).

Gargron (approved):

Issue #2678, opened 2017-05-01
Typo on Edit Profile


Cassolotl:
Another minor thing. :) On /settings/profile:

![screen shot 2017-05-01 at 10 53 52](https://cloud.githubusercontent.com/assets/6607512/25577231/83f8be60-2e5c-11e7-8164-62140a9b4ba2.png)

When the number of characters remaining is 1, it should say "character" instead of "characters".

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

stephenburgess8:
Ideally, we would dynamically set the word character to change from plural to singular when the count reaches one (or negative one?). That doesn't seem straightforward, and may become odd to translate into other languages since languages will likely treat pluralization different in this case.  Embedding dynamic content in sentences makes it hard to translate, though with Mastodon's open source translating going on that may not be as significant of a problem. 

Here's the line where that text is set in English: https://github.com/tootsuite/mastodon/blob/master/config/locales/simple_form.en.yml#L10

What about changing it to `Characters left (1)`? 

https://ux.stackexchange.com/questions/81814/should-i-use-singular-or-just-default-plural-in-labeling-the-followers

Cassolotl:
Yeah, something like "Characters remaining: number" would be fine!

akihikodaki:
Closing with #2685.

PR #2685, opened 2017-05-01
feat(character-count): Do not embed dynamic value in sentence

stephenburgess8:
Restructure label to not embed dynamic value in sentence. This allows for grammatical correctness whether the number is one or not one.

#2678

screen shot 2017-05-01 at 10 03 37 am


stephenburgess8:
Feel free to close if you would prefer this not be changed.

config/locales/simple_form.en.yml
```diff
@@ -4,10 +4,10 @@ en:
     hints:
       defaults:
         avatar: PNG, GIF or JPG. At most 2MB. Will be downscaled to 120x120px
-        display_name: '%{counter} characters left'
+        display_name: 'Characters left: %{counter}'
```
Gargron: You know you can do something like:

```yml
display_name:
  one: 1 character left
  other: "%{count} characters left"
```

You just have to rename counter to count.

Gargron (changes requested):

stephenburgess8:
Thanks for your note on this. https://webtranslateit.com/docs/file_formats/yaml/

I've tested this view in all the languages that I've changed here. I've noticed two things which looks like either cruft or something is wrong:

For Brazilian Portuguese, the profile edit view isn't being translated. (this is happening before the changes I made)
The Turkish `devise.tr.yml` file uses these translations, but these aren't translated in that part of the code AFAICT. That should possibly be deleted from the Turkish translation file. Should I open an issue for either of those? I don't use them, I just noticed it when testing.

Gargron (approved):

PR #2695, opened 2017-05-01
grammar fixes

drewgates:
“such as” should be used instead of “like” under imports->preface.

Reworded imports->preface and recovery_instructions for better flow.


yiskah (approved):
These phrasings are more clear and easy to follow. I approve

mjankowski (approved):

PR #2697, opened 2017-05-01
annotate models

yhirano55:
[*proposal*] [ctran/annotate_models](https://github.com/ctran/annotate_models) can automatically annotate every time we run `db:migrate`. I think that annotations about schema info help us to implement. How do you think about it?


Gargron (approved):

mjankowski:
I'd rather not include this ... its a lot of noise in every file for something that can already been in schema.rb

Gargron:
@mjankowski I keep forgetting the schema sometimes and having to look into schema.rb, so this seemed convenient to me. Also I believe I have seen this type of thing in the Discourse codebase.

PR #2715, opened 2017-05-02
Fix subscription expiration condition

ykzts:
@Gargron I was mistaken in #2712. I fixed it so please check it again.

### DIFF
--- a/app/views/admin/pubsubhubbub/index.html.haml
+++ b/app/views/admin/pubsubhubbub/index.html.haml
@@ -11,7 +11,6 @@
       %th= t('admin.pubsubhubbub.last_delivery')
   %tbody
     - @subscriptions.each do |subscription|
-      - expired = Time.now.utc > subscription.expires_at
       %tr
         %td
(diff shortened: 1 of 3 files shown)


app/views/admin/pubsubhubbub/index.html.haml
```diff
@@ -11,7 +11,7 @@
       %th= t('admin.pubsubhubbub.last_delivery')
   %tbody
     - @subscriptions.each do |subscription|
-      - expired = Time.now.utc < subscription.expires_at
+      - expired = Time.now.utc > subscription.expires_at
```
mjankowski: Can you put this into a method on subscription, instead of setting a variable in a view?

ykzts: Certainly. That is beautiful.

ykzts: OK. And, add spec 4784f0e.

Gargron (approved):

PR #2721, opened 2017-05-02
Creation of activerecord for OC

Quent-in:


config/locales/activerecord.oc.yml
```diff
@@ -0,0 +1,12 @@
+en:
```
Gargron: Must be oc

Gargron (changes requested):

Gargron (approved):

PR #2734, opened 2017-05-03
Likely fix #2458, fix #2031 - handle out-of-order deletes for statuses

### LINKED ISSUE #2031: Direct messages are undeletable
How 2 reproduce:
1. Direct message a user on another instance
2. Delete the direct message on your instance
3. The direct message will not be deleted on the other instance

I haven't checked, but it also might not delete locally for other users too due to some conversations I've had.
This seems like it miiight be something that should be high priority.

* * * *

### LINKED ISSUE #2458: Deleting toots across instances
People post dumb/incorrect stuff accidentally sometimes. That's where deleting toots comes in handy.

My (possibly incorrect/outdated) understanding is that toot deletion does not federate across instances.

This can result in potential privacy problems, not to mention various social problems.

If a "CREATE TOOT" message can be federated across instances, I see no reason why a "DELETE TOOT" message can't be federated across instances too.

Now, this obviously has the potential, if improperly implemented, to result in censorship attacks. That's just something to think about when implementing this.

Ideally, Mastodon users would have private keys and all messages would be signed... but that's probably asking too much at this point.

**EDIT:** [I'm told](https://quitter.se/notice/11453539) that this is an OStatus issue, and if that's the case, this problem is big enough on its own that it may be worth considering an OStatus 2.0 effort at some point. Such an effort could be used to address various other problems with OStatus as well.

* * * *

Gargron:
If a delete arrives before the original status, cache that information
for 6h, and if the original status arrives in that window, ignore it


app/services/process_feed_service.rb
```diff
@@ -82,7 +87,13 @@ def notify_about_reblog!(status)
     def delete_status
       Rails.logger.debug "Deleting remote status #{id}"
       status = Status.find_by(uri: id)
-      RemoveStatusService.new.call(status) unless status.nil?
+
+      if status.nil?
+        redis.setex("delete_upon_arrival:#{id}", 6 * 3_600, id)
```
Taekiro: Typo : 6 * 3600 (no underscore)

ykzts: It is not typo. Integer including "`_`" is valid Ruby syntax.

Taekiro: my bad, sorry.

app/services/process_feed_service.rb
```diff
@@ -273,5 +284,9 @@ def thread(xml = @xml)
     def account?(xml = @xml)
       !xml.at_xpath('./xmlns:author', xmlns: TagManager::XMLNS).nil?
     end
+
```
mjankowski: whitespace here!

mjankowski (changes requested):
We should probably have at least one spec that shows that an incoming status which has a delete key present does get ignored...

Gargron:
Added as requested

mjankowski (approved):

PR #2739, opened 2017-05-03
Fix PuSH to instances where WEB_DOMAIN != LOCAL_DOMAIN

### LINKED ISSUE #2672: v1.3+ does not send toots to instances using a separate WEB_DOMAIN
EDIT: #2111 could be causing this. See comment https://github.com/tootsuite/mastodon/issues/2672#issuecomment-298348460.

Original issue:

- - -

Toots from a number of remote instances stopped appearing on my instance. Notifications remain working. Toots originating from my instance reach other instances just fine.

My instance is on v1.3.2, running on a separate WEB_DOMAIN. Logs show nothing particularly suspicious.

The problem started a couple of days ago, around Apr. 29. The problem appears to be instance-wide, not limited to individual users / toots, unlike #2613 . Manually refreshing PuSH, and re-following users did not fix the problem. Instances from which toots are missing (that I know of) range from v1.2.2 to v1.3.2.

Could someone please shed some light on why and how this is happening? Thank you!

* * * *

ClearlyClaire:
This is an attempt to fix #2672 and hasn't been tested yet.
This is done in two ways:
- Push public toots without attempting to verify the callback URL against the followers
- Add host part of followers' remote_url to the list of allowed domains. This makes assumptions not in OStatus, but should work with other Mastodon instances
Rspec tests should be added, and the way the followers_domains is built should probably be optimized.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -160,7 +160,7 @@ class Account < ApplicationRecord
 
   class << self
     def domains
-      reorder(nil).pluck('distinct accounts.domain')
+      reorder(nil).pluck('distinct accounts.domain') + where.not(web_domain: nil).reorder(nil).pluck('distinct accounts.web_domain')
     end
 
     def triadic_closures(account, limit: 5, offset: 0)
@@ -255,5 +255,7 @@ class Account < ApplicationRecord
     return if local?
 
     self.domain = TagManager.instance.normalize_domain(domain)
+    remote_domain = Addressable::URI.parse(remote_url).normalized_host
+    self.web_domain = remote_domain unless domain == remote_domain
   end
 end
--- a/app/workers/pubsubhubbub/distribution_worker.rb
+++ b/app/workers/pubsubhubbub/distribution_worker.rb
@@ -45,6 +45,6 @@ class Pubsubhubbub::DistributionWorker
   end
 
   def allowed_to_receive?(callback_url)
-    @domains.include?(Addressable::URI.parse(callback_url).host)
+    @domains.include?(Addressable::URI.parse(callback_url).normalized_host)
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'aws-sdk', '~> 2.9'
 gem 'paperclip', '~> 5.1'
 gem 'paperclip-av-transcoder', '~> 0.6'
 
+gem 'idn-ruby', require: 'idn'
 gem 'addressable', '~> 2.5'
 gem 'bootsnap'
 gem 'browser'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,6 +198,7 @@ GEM
       parser (>= 2.2.3.0)
       rainbow (~> 2.2)
       terminal-table (>= 1.5.1)
+    idn-ruby (0.1.0)
     jmespath (1.3.1)
     json (2.1.0)
     kaminari (1.0.1)
@@ -510,6 +511,7 @@ DEPENDENCIES
   http_accept_language (~> 2.1)
   httplog (~> 0.99)
   i18n-tasks (~> 0.9)
+  idn-ruby
   kaminari (~> 1.0)
   letter_opener (~> 1.4)
   letter_opener_web (~> 1.3)
--- a/app/lib/tag_manager.rb
+++ b/app/lib/tag_manager.rb
@@ -70,7 +70,7 @@ class TagManager
 
     uri = Addressable::URI.new
     uri.host = domain.gsub(/[\/]/, '')
-    uri.normalize.host
+    uri.normalized_host
   end
 
   def same_acct?(canonical, needle)
--- a/app/services/concerns/author_extractor.rb
+++ b/app/services/concerns/author_extractor.rb
@@ -14,7 +14,7 @@ module AuthorExtractor
 
       return nil if username.blank? || uri.blank?
 
-      domain = Addressable::URI.parse(uri).normalize.host
+      domain = Addressable::URI.parse(uri).normalized_host
       acct   = "#{username}@#{domain}"
     end
 
--- a/app/services/fetch_remote_status_service.rb
+++ b/app/services/fetch_remote_status_service.rb
@@ -24,7 +24,7 @@ class FetchRemoteStatusService < BaseService
     xml.encoding = 'utf-8'
 
     account = author_from_xml(xml.at_xpath('/xmlns:entry', xmlns: TagManager::XMLNS))
-    domain  = Addressable::URI.parse(url).normalize.host
+    domain  = Addressable::URI.parse(url).normalized_host
 
     return nil unless !account.nil? && confirmed_domain?(domain, account)
 
@@ -36,6 +36,6 @@ class FetchRemoteStatusService < BaseService
   end
 
   def confirmed_domain?(domain, account)
-    account.domain.nil? || domain.casecmp(account.domain).zero? || domain.casecmp(Addressable::URI.parse(account.remote_url).normalize.host).zero?
+    account.domain.nil? || domain.casecmp(account.domain).zero? || domain.casecmp(Addressable::URI.parse(account.remote_url).normalized_host).zero?
   end
 end
--- a/app/workers/pubsubhubbub/delivery_worker.rb
+++ b/app/workers/pubsubhubbub/delivery_worker.rb
@@ -43,7 +43,7 @@ class Pubsubhubbub::DeliveryWorker
   end
 
   def host
-    Addressable::URI.parse(subscription.callback_url).normalize.host
+    Addressable::URI.parse(subscription.callback_url).normalized_host
   end
 
   def headers
--- a/db/migrate/20170625195017_add_web_domain_to_accounts.rb
+++ b/db/migrate/20170625195017_add_web_domain_to_accounts.rb
@@ -1,0 +1,9 @@
+class AddWebDomainToAccounts < ActiveRecord::Migration[5.0]
+  def up
+    add_column :accounts, :web_domain, :string, null: true, default: nil
+  end
+
+  def down
+    remove_column :accounts, :web_domain
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170624134742) do
+ActiveRecord::Schema.define(version: 20170625195017) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -56,6 +56,7 @@ ActiveRecord::Schema.define(version: 20170624134742) do
     t.integer "followers_count", default: 0, null: false
     t.integer "following_count", default: 0, null: false
     t.datetime "last_webfingered_at"
+    t.string   "web_domain"
     t.index "(((setweight(to_tsvector('simple'::regconfig, (display_name)::text), 'A'::\"char\") || setweight(to_tsvector('simple'::regconfig, (username)::text), 'B'::\"char\")) || setweight(to_tsvector('simple'::regconfig, (COALESCE(domain, ''::character varying))::text), 'C'::\"char\")))", name: "search_index", using: :gin
     t.index "lower((username)::text), lower((domain)::text)", name: "index_accounts_on_username_and_domain_lower"
     t.index ["uri"], name: "index_accounts_on_uri"


TecyT:
Is modifying `Account#followers_domains` a good idea? I don't find it used elsewhere, but people would likely assume it to mean the `Account#domain`s of `Account#followers`. Maybe make something separate like `Account#followers_subscription_domains`?

Also it would be nice to have a test case to make sure this kind of regression won't happen anymore.

ThibG:
I'm open about using a new Account#followers_subscription_domains or similar, but you're not sure those will actually be used for subscription (except for Mastodon instances). Moreover, I can't really think of any use for Account#followers_domains as the domain part of the acct: URI doesn't mean much anyway (since actor URI, the feeds and API endpoints are what matters and can be hosted on a different domain entirely, as its the case with the issue at end). Either way, this should be documented.

As for the tests, I agree, but I haven't had the time to write them, and I'm on my way to my dayjob now.

TecyT:
> but you're not sure those will actually be used for subscription

Yes, but what matters is that Mastodon code use them for checking subscription validity. It is for making the code easier to understand.

> I can't really think of any use for Account#followers_domains as the domain part of the acct: URI doesn't mean much anyway

True. I can't think of any either. So we can rename `followers_domains`, instead of keeping it and making something new. No reason for continued usage of an underspecific name, after all.

> Either way, this should be documented.

I agree. It should be noted that an assumption is made here.

app/workers/pubsubhubbub/distribution_worker.rb
```diff
@@ -12,10 +12,10 @@ def perform(stream_entry_id)
 
     account = stream_entry.account
     payload = AtomSerializer.render(AtomSerializer.new.feed(account, [stream_entry]))
-    domains = account.followers_domains
+    domains = account.followers_subscription_domains if stream_entry.status&.hidden?
 
     Subscription.where(account: account).active.select('id, callback_url').find_each do |subscription|
-      next unless domains.include?(Addressable::URI.parse(subscription.callback_url).host)
+      next unless !stream_entry.status&.hidden? || domains.include?(Addressable::URI.parse(subscription.callback_url).host)
```
Gargron: This becomes hard to read because of "unless not" so maybe "next if hidden and not included"

app/models/account.rb
```diff
@@ -185,8 +185,12 @@ def subscribed?
     !subscription_expires_at.blank?
   end
 
-  def followers_domains
-    followers.reorder(nil).pluck('distinct accounts.domain')
+  # Domains potentially used for PuSH callback URLs of followers' instances
+  def followers_subscription_domains
+    domains = followers.reorder(nil).pluck('distinct accounts.domain')
+    remote_urls = followers.reorder(nil).pluck('distinct accounts.remote_url')
```
Gargron: I'm a bit concerned because this will be a huge array, every remote follower has a unique remote_url, distinct is kinda useless here

app/workers/pubsubhubbub/distribution_worker.rb
Gargron: If WEB_DOMAIN and LOCAL_DOMAIN must be related by a root domain (i.e. one of them is a subdomain) then maybe that can be the check instead?

Gargron (changes requested):

app/workers/pubsubhubbub/distribution_worker.rb
ThibG: Yes, it's a bit unreadable, I will change that. As for WEB_DOMAIN being a subdomain of LOCAL_DOMAIN, I'm uneasy about that, that is not an assumption we have made so far, and not one GNU Social has made either. I think multiple GNU Social users have acct: URIs pointing to different domains entirely (I can think of MMN-o, but some of those can be considered aliases). It would make the check easier, though.

app/models/account.rb
ThibG: Ah, yeah, I just copied it, but the “distinct” is useless indeed. I'm not sure how to optimize that. We could just .uniq the array, but I'm not sure it will be very useful. We could also add a “remote_url_domain” field to Account, to parse the URL only when refreshing the remote account details, and not at each private/direct message, but I'm not sure how much better that is.
One part of the change was also to compute this array only when pushing private/direct messages, and not public/unlisted ones.

ThibG: I'll push a slightly optimized version of this, which should probably be ok for about a thousand followers across a few hundreds remote domains, but not much more.

ThibG:
I have updated my PR with a few optimizations and fixes, but it will probably not scale well to users with thousands of followers. The best way to handle this would probably be to store the parsed-and-normalized domain alongside the URL in a field of the “accounts” table, but that would probably make for a fairly long-running database transition script…

ThibG:
The last commit introduces a web_domain field to the accounts table as a cache for the normalized domain part of remote_url. This comes with a migration script that I fear will be quite long-running.

db/migrate/20170503172351_add_web_domain_to_accounts.rb
```diff
@@ -0,0 +1,16 @@
+class AddWebDomainToAccounts < ActiveRecord::Migration[5.0]
+  def up
+    add_column :accounts, :web_domain, :string, null: true, default: nil
+
+    ActiveRecord::Base.transaction do
```
Gargron: You are absolutely right that this migration script is impossible to run. It will take hours if not days, and migrations place a lock on the entire schema meaning it would be hours of downtime. I am almost veering towards your initial approach of just iterating over everything.

app/workers/pubsubhubbub/distribution_worker.rb
```diff
@@ -12,10 +12,10 @@ def perform(stream_entry_id)
 
     account = stream_entry.account
     payload = AtomSerializer.render(AtomSerializer.new.feed(account, [stream_entry]))
-    domains = account.followers_domains
+    domains = account.followers_subscription_domains if stream_entry.status&.hidden?
 
     Subscription.where(account: account).active.select('id, callback_url').find_each do |subscription|
-      next unless domains.include?(Addressable::URI.parse(subscription.callback_url).host)
+      next if stream_entry.status&.hidden? && !domains.include?(Addressable::URI.parse(subscription.callback_url).normalize.host)
```
Gargron: There is a second approach. Instead of checking whether the destination is allowed during the loop here, do it in Pubsubhubbub::DeliveryWorker instead.

db/migrate/20170503172351_add_web_domain_to_accounts.rb
Gargron: Yeah, sorry about making you do the migration bit. See my comment on the other file.

Gargron (changes requested):

app/workers/pubsubhubbub/distribution_worker.rb
TecyT: It should be made a hashtable, or at least .uniq as @ThibG commented on the last review (though he said he is 'not sure it will be very useful'). An O(n^2) operation is still O(n^2) on threads: a few threads are nothing compared to 10k^2 if the person has a lot of followers.

Gargron: You're distributing to the same number of subscriptions either way, the difference is between checking a huge array for a value that needs to be transformed first (high memory use), or checking just one thing in the responsible job (low memory use)

Also, lots of small jobs are overall better for the system than a few long-running ones, because small jobs return consumed resources like connections faster, so jobs of different varities can run more or less at the same time, while one long-running job may block a sidekiq worker for a long time.

TecyT: I see what you mean. Not passing the table to the worker, but leveraging the DB index (essentially a hashtable), thus having both O(n) and benefits of small jobs. Much more clever than mine!

db/migrate/20170503172351_add_web_domain_to_accounts.rb
ThibG: I don't think that would ever take days, but hours is entirely possible, and that's not acceptable for a migration script. For the record, on my really small instance (only ~2k remote accounts known, but also low-end ARM hardware), the migration script took about 2 minutes to run.

app/workers/pubsubhubbub/distribution_worker.rb
ThibG: In my last proposal (well, before the last commit which switches to a database field), I am already using “.uniq” to avoid .normalize()-ing URLs from the same domain, as it's the most expensive part of the computation. This does however require the use of a large array at some point, indeed.

Checking it in Pubsubhubbub::DeliveryWorker would involve doing the whole thing for each PuSH subscriber (so, roughly multiply the total running time by the number of remote instances), and avoiding the big array means you will do multiple a .normalize() call for *each* follower, even if multiple ones happen to use the same domain…

ThibG: I'm afraid I'm out of ideas to make this efficient. I guess we could add the database field right now, without doing the full migration, but with everything to compute the web_domain whenever the account details are updated. Then, we can fetch distinct non-null web_domain from the followers, and do the big array thing for the remaining remote followers.

TecyT: @ThibG is right. Forgot that PuSH subscriptions are anonymous (again) for the first time. The DB index cannot be relied on. I believe filling the database gradually is a good idea.

If adding a column is expensive what about creating a new table? Is it a concievable situation for a user to have multiple 'authoized' domains?

ThibG:
Updated with my last suggestion (using both the web_domain value from the database as well as computing a set of domains on-the-fly for remote followers that don't have the value set). I haven't tested it yet.

app/workers/pubsubhubbub/distribution_worker.rb
ThibG: Yes, it is conceivable as it is only a guess. For instance, “authorized” domains include the domain part of the acct: URI too, and we might also allow the domain part of the PuSH or salmon endpoints…

ThibG:
Fixed a mistake in an earlier commit (which used `Addressable::UR`I's domain method instead of `host`). Now running on my instance, haven't seen any issue so far, but obviously haven't seen any improvement either, since the *remote* instances would have to run the fix.
In the meantime, I'm monitoring how fast the new `web_domain` field gets updated for known remote users.

ThibG:
I'm investigating a few ways to significantly reduce the time we spend parsing and normalizing URIs:
* Using `uri.normalized_host` instead of `uri.normalize.host`, avoiding to normalize the other components and creating a new object
* Using the “idn-ruby” gem which implements IDNA natively instead of using the default pure Ruby implementation

app/workers/pubsubhubbub/distribution_worker.rb
Gargron: Sorry, but I insist on the approach I described. Please go back to the approach before migrations/db changes, and move the check to the delivery worker.

ThibG: I'm not sure what you want me to do here. We would not be able to share the domains array anymore, and would need to re-compute it instead. Or iterate over all the followers without deduplicating as soon as possible, which will ease the RAM requirements, but make each job would run much longer.

ThibG: Do you want me to use something like:
```ruby
  def is_domain_subscribed?(domain)
    return true if followers.where(domain: domain).count.positive?
    return true if followers.where(web_domain: domain).count.positive?
    followers.reorder(nil).where('accounts.domain IS NOT NULL AND accounts.web_domain IS NULL')
             .select('accounts.remote_url').find_each do |url|
      return true if domain == Addressable::URI.parse(url).normalized_host
    end
    false
  end
```

As I said earlier, I fear this is going to be much slower than what I came up with earlier (at least without the `web_domain` field in the database). But you're right that it should consume less memory.

Gargron:
#2954 changes the distribution to work for everybody for public statuses again. I don't have a good solution to tracking "web_domain" since it's not something the protocol announces in any way. For all I know, the Atom feed, Salmon endpoint and PuSH hub of a user could all be on different domains. :(

ThibG:
Le 9 mai 2017 23:53:25 GMT+02:00, Eugen Rochko <notifications@github.com> a écrit :
>#2954 changes the distribution to work for everybody for public
>statuses again.

Good! That will already be a much better situation, although it would be good if follower-only toots worked too.

> I don't have a good solution to tracking "web_domain"
>since it's not something the protocol announces in any way. For all I
>know, the Atom feed, Salmon endpoint and PuSH hub of a user could all
>be on different domains. :(

Yes, indeed. This notion of private toots is a departure from OStatus anyway, which does not impose any kind of relation on the different URLs. My proposal was for it to work with *Mastodon* instances where LOCAL_DOMAIN ≠ WEB_DOMAIN.

ThibG:
I have rebased my branch on latest master and completely revamped it: first, there is a bunch of fixes and performance improvements, and the last commit is the most invasive one and introduces the “web_domain” field to the database.

Unlike in my previous approaches, this is meant to be set only when the domain part of the `remote_url` and the `domain` of an `Account` differ, to reduce the need for deduplication.

Moreover, the accounts are not updated by a migration script, and the update mechanism relies on the normal remote account mechanism instead, thus having almost no impact on performances, at the cost of undelivered toots when all the following conditions are met:
- The toot is private (followers-only)
- The remote follower is older than the migration and did not get updated
- The remote follower is on an instance where `WEB_DOMAIN` ≠ `LOCAL_DOMAIN`

ThibG:
Rebased on master and now running on my instance. Have seen no issue so far.

ThibG:
@Gargron are you ok with this approach or do you want me to change anything? Having to constantly rebase the migration script is a bit of pain!

ThibG:
Let me bump this PR again now that 1.4.1 is released! As earlier, I have been running with this changes (regularly rebasing them on master) without any issue I am aware of.

Gargron:
@ThibG Yeah, sorry about the delays on this one - I'm always hesitant to change db schema. Especially when in this case it's so specific to WEB_DOMAIN which is a Mastodon only thing and not even recommended. However, this led to me thinking - why don't we make more use of Webfinger aliases? Webfinger provides a canonical acct, and an array of aliases. When using WEB_DOMAIN, we could list that among the aliases. On the other end, perhaps accounts should have an aliases string array column, which would keep track of those. We still have the "that's a lot of select + parsing out domain out of acct or URL" problem but at least it's not hardcoded to web_domain

ThibG:
Yes, I can understand why you would be hesitant to change db schema for this. Webfinger aliases would contain entire URIs that would probably have to be parsed each time, which will probably be way too slow (my current solution is to save the domain itself in web_domain, at remote account creation/update, so that the lookup is fast). As for including it in the aliases, the profile URL (thus including the WEB_DOMAIN) is already part of the advertised aliases.

Gargron:
I have a new idea. I'm going to implement http signatures (there is a draft rfc for it). PuSH subscription requests will be signed on behalf of the user who triggered the subscription. We don't really need a specific user but once we have a user we can store the domain in the subscriptions table, and then filter by it.

ThibG:
That seems sensible to me as long as it is optional (since that's not part of OStatus).

PR #2742, opened 2017-05-03
changed to run foreman with vagrant

abcang:
Changed to run foreman with vagrant.
Also changed so that database setting can be set with environment variable.


Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: bundle exec puma -C config/puma.rb
-stream: yarn run start
-webpack: ./bin/webpack-dev-server
+web: PORT=3000 bundle exec puma -C config/puma.rb
```
mjankowski: I think this already honored `PORT` if set, and I'd rather not hardcode it to something...

config/database.yml
```diff
@@ -3,6 +3,10 @@ default: &default
   pool: <%= ENV["DB_POOL"] || ENV['MAX_THREADS'] || 5 %>
   timeout: 5000
   encoding: unicode
+  username: <%= ENV['DB_USER'] || 'postgres' %>
```
mjankowski: could vagrant set `DATABASE_URL` instead, which rails will already respect?

mjankowski (changes requested):

mjankowski (approved):

mjankowski:
If there are relevant changes here, can you make them to docs too? - https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Vagrant-guide.md

abcang:
Yes, I will do it later!

mjankowski (changes requested):
I'm not a regular vagrant user, but I pulled your branch and attempted locally.

After `vagrant up` from the branch, I get...

```
magicvault:mastodon mjankowski$ vagrant ssh -c "cd /vagrant && foreman start"
bash: foreman: command not found
Connection to 127.0.0.1 closed.
```

Is there a missing step somewhere in here, or did I miss something should have done?

abcang:
I changed it to install foreman at provisioning.
So, you probably need to run provisioning again.

```bash
$ vagrant provision
```

mjankowski (approved):

config/initializers/ostatus.rb
```diff
@@ -12,7 +12,7 @@
   config.x.use_s3       = ENV['S3_ENABLED'] == 'true'
 
   config.action_mailer.default_url_options = { host: web_host, protocol: https ? 'https://' : 'http://', trailing_slash: false }
-  config.x.streaming_api_base_url          = 'http://localhost:4000'
+  config.x.streaming_api_base_url          = "http://localhost:#{port.to_i + 100}"
```
Gargron: Can you tell me why you can't use port 4000 here?

Gargron (changes requested):

config/initializers/ostatus.rb
abcang: With foreman, $PORT is set at run time. Also, $PORT of the second and subsequent processes is the value obtained by adding 100 to the base port.
For example, when the base port is 3000, rails starts with port 3000 and node's server starts with port 3100.

```text:Procfile
web: bundle exec puma -C config/puma.rb             # $PORT is 3000
stream: yarn run start                              # $PORT is 3100
webpack: ./bin/webpack-dev-server --host 0.0.0.0    # $PORT is 3200
```
So streaming server can not be started with 4000.
Of course, it can be started with port 4000 if the port number is hard-coat.

```
stream: PORT=4000 yarn run start
```

Gargron: Aha, I see. I think it might be better to hardcode PORT=4000 in this case because everyone's already used to it running on that port, and changing defaults will have lots of people's installations breaking.

Gargron (approved):

PR #2745, opened 2017-05-03
Update of simple_forum_oc

Quent-in:


Gargron (approved):

config/locales/simple_form.oc.yml
```diff
@@ -12,6 +12,8 @@ oc:
         data: Fichièr CSV exportat d’una autra instància Mastodon
       sessions:
         otp: Picatz lo còdi d’autentificacion en dos temps (Two factor code) de vòstre mobil o utilizatz un de vòstres còdis de recuperacion.
+  user:
```
ykzts: broken indent.

Gargron: Off the top of my head I don't even remember which level allowed_languages is supposed to go on. Should't it be just under labels.defaults?

Gargron (changes requested):

config/locales/simple_form.oc.yml
ykzts: This label is hint. The position is no problem. But indentation is broken.

Quent-in: Okay, what should I do?

ykzts: @Quent-in 5f9cb4888298528afc4866d5dc72c43877a41305 is the difference that makes similar changes in Japanese. please refer.

ykzts (approved):

Gargron (approved):

PR #2747, opened 2017-05-03
Switch docker-compose to version 3.

valentin2105:
It allow possibility to Deploy Mastodon in a Swarm cluster directly from the compose file.

I will write a small doc to explain How to do this.


Gargron:
> `extends`: This option has been removed for version: "3.x" Compose files.

https://docs.docker.com/compose/compose-file/compose-versioning/#version-2x-to-3x

valentin2105:
@Gargron I missed this, I deleted every extends occurences.

Gargron (approved):

Wonderfall:
Why not the latest version 3.2?

Gargron:
@Wonderfall This is about the version of the config file, not of the docker-compose binary - unless the config file has minor versions?

Wonderfall:
The config file has minor versions since version 2. It's also possible `version: 3` refers to the latest 3.x, but I'm not sure.

valentin2105:
Updated in 3.2, (tested). But requiere a docker-compose --upgrade via Pip.

PR #2750, opened 2017-05-03
Replace background image with smaller svg pattern

mjankowski:
The main improvement here is replacing a ~264kb background photo which is basically a dark image with a small pattern in it, with a less than 1k svg file which produces a similar effect.

I'm not particularly attached to THIS svg file, but for page load purposes on slow connections it makes sense to me to use a smaller background image. I'll look for different options if the idea is sound but the particular choice is not.

screen shot 2017-05-03 at 9 23 01 am


ykzts (approved):

mjankowski:
This has been rebased after https://github.com/tootsuite/mastodon/commit/b85dec2b976e0bea9f7a3ede832f3c8e16c8e2ef and updated to use the base ui background color instead of the dark grey it previously had.
screen shot 2017-05-08 at 10 18 57 am

nolanlawson:
I'm a bit fond of the old image; could we just compress it with something like mozjpeg or jpeg-archive? I tried using jpeg-archive with various quality settings and got fairly good results, although the hue seems more purplish than blue now. But on the bright side, it can get as low as 32K (low quality) or 84K (medium quality): https://gist.github.com/nolanlawson/45c3b991bcf394256226d49c9359873b

(The medium one in particular looks good to me; I think low may be too far)

mjankowski:
> I'm a bit fond of the old image; could we just compress it with something like mozjpeg or jpeg-archive?

Short of a larger replacement, that would be great.  It used to be ~800k, and I got it down to where it is not (~260k) with some compression. I agree that your "medium" one is probably good enough, while also being a good further size reduction.

Gargron:
I intend to remove the background image altogether so I think this won't be necessary.

PR #2756, opened 2017-05-03
Fallback locales

alpaca-tc:
because of a lack of localization 👍


Gargron:
I had the impression this was default anyway?

alpaca-tc:
Sorry, I overlooked that. 🙏

PR #2758, opened 2017-05-03
Application prefs section

muffinista:
This adds a page to the preferences section to manage credentials for user-created applications.

The goal here is to make it easier for users to manage scripts that require API access. This could be clients, scripts that do some sort of data management, bots, etc. While it's obviously possible to get credentials for a script right now, it's a bit challenging and involved. With this code, you can create an app in the preferences section, and get a token right there. You can also regenerate your token if needed.

These changes also address an issue that exists in the code right now, which is that there is no way to edit the name or website of an app right now.


config/initializers/doorkeeper.rb
```diff
@@ -45,7 +45,7 @@
   # Optional parameter :confirmation => true (default false) if you want to enforce ownership of
   # a registered application
   # Note: you must also run the rails g doorkeeper:application_owner generator to provide the necessary support
-  # enable_application_owner :confirmation => true
+  enable_application_owner
```
beatrix-bitrot: does this have any effects on applications that exist already before these changes are applied?

muffinista: I don't believe so, and I'm running it in production with no reported issues.

beatrix-bitrot: that was my only concern really. there are some merge conflicts so i'm going to mark this as requesting changes

beatrix-bitrot (changes requested):

muffinista:
OK, conflicts should be resolved, thanks!

beatrix-bitrot (approved):

beatrix-bitrot (dismissed):
the CI checks are failing ):

muffinista:
So sorry about that! I'm travelling and made a typo using the github inline editor. CI is passing now.

beatrix-bitrot:
No worries. Also sorry to ask but could you add a screenshot or two? I'm trying to take a little more care with my reviews and so I'd either want to run your changes on one of my servers before approving or at least see how it looks in some screenshots.

muffinista:
No problem at all. First, the code is running on [botsin.space](https://botsin.space) if you want to create an account.


Here's the app list:

screenshot 2017-05-10 20 41 16

Here's the Create New App page:

screenshot 2017-05-10 20 42 02

The app page, with tokens:
screenshot 2017-05-10 20 42 57

akihikodaki (changes requested):
It may be good to have a notice to tell users to keep tokens _secret_ to the application page.
Mastodon has a variety of users and some of them are not familiar with them.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      expect(assigns(:applications)).to include(app)
+      expect(assigns(:applications)).to_not include(other_app)
+    end
+  end
+
+  
+  describe 'GET #show' do
+    it 'returns http success' do
+      get :show, params: { id: app.id }
+      expect(response).to have_http_status(:success)
```
akihikodaki: Spec that it acutally shows the application. (testing `assigns(:application)` is enough.)

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  describe 'GET #new' do
+    it 'works' do
+      get :new
+      expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe 'POST #create' do
+    describe 'success' do
```
akihikodaki: * Test if it actually creates an application.
* Use `context` instead of `describe`.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+      end
+
+      it 'renders form again' do
+        expect(response).to render_template(:new)
+      end
+    end
+  end
+  
+  describe 'PUT #update' do
+    describe 'success' do
```
akihikodaki: * Test if it actually updates an application.
* Use `context` instead of `describe`.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        expect(response).to have_http_status(:success)
+      end
+
+      it 'renders form again' do
+        expect(response).to render_template(:new)
+      end
+    end
+  end
+  
+  describe 'PUT #update' do
```
akihikodaki: It is not `PUT` but `PATCH`.

nightpool:
Is there a reason the redirect URI is a textarea instead of just an input? the styling looks odd.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  describe 'regenerate' do
+    let(:token) { user.token_for_app(app) }
+    before do
+      expect(token).to_not be_nil
+      put :regenerate, params: { application_id: app.id }
+    end
+
+    it 'should create new token' do
+      expect( user.token_for_app(app) ).to_not eql(token)
```
akihikodaki: Do not add spaces in parentheses.

spec/models/user_spec.rb
```diff
@@ -270,4 +270,25 @@ def fabricate
...
+      expect(user.token_for_app(app)).to be_a(Doorkeeper::AccessToken)
+    end
+
+    it 'persists a token' do
+      t = user.token_for_app(app)
+      expect(user.token_for_app(app)).to eql(t)
+    end
+
+    it 'is nil if user does not own app' do
+      app.owner = nil
```
akihikodaki: Use `update!` instead of a combination of an assignment and `save!`.

config/initializers/doorkeeper.rb
akihikodaki: I'm not sure even why `owner` is required. Could you explain the reason?

akihikodaki (changes requested):
For reuqests of additional tests; [pending examples](https://relishapp.com/rspec/rspec-core/v/3-6/docs/pending-and-skipped-examples) would be sufficient to merge this change, but actual examples will be appreciated.

Thank you for your contribution. This feature is essential.

config/initializers/doorkeeper.rb
muffinista: Yes, it sets the application owner to the current user if that data is available. Without that, we can't associate apps with owners.

akihikodaki: I mean I doubt associating apps with owners is necessary. Isn't tracking access tokens with resource owners enough?

muffinista: Maybe I'm missing something. The whole point of this PR is to associate applications with their creator so they can be managed and displayed in the settings section of that user. While tokens have an associated user, its very difficult to use that link to actually do any management of the app without some brittle hacks.

akihikodaki: I see. It looks good to me to add `owner`.

muffinista:
Thanks for the feedback! I think I've addressed most of the concerns raised here. I've updated the specs, added a note about not sharing API keys, and made a few other changes. I didn't change the textarea for the redirect URL for while I agree with @nightpool that the textarea looks a little weird here (especially with the default redirect), it makes it a lot easier to see URLs that are longer. I would be fine with changing it to a normal input though.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  
+  describe 'GET #show' do
+    it 'returns http success' do
+      get :show, params: { id: app.id }
+      expect(response).to have_http_status(:success)
+      expect(assigns[:application]).to eql(app)
+    end
+
+    it 'returns 404 if you dont own app' do
+      app.owner = nil
```
akihikodaki: Use `update!` here, too.

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  let(:user) { Fabricate(:user) }
+  let(:app) { Fabricate(:application, owner: user) }
+  
+  before do
+    sign_in user, scope: :user
+  end
+
+  describe 'GET #index' do
+    let(:other_app) { Fabricate(:application) }
+    before {
```
akihikodaki: I think `let!` is what you expect, isn't it?

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+    before do
+      post :destroy, params: { id: app.id }
+    end
+
+    it 'redirects back to applications page' do
+      expect(response).to redirect_to(settings_applications_path)
+    end
+
+    it 'removes the app' do
+      expect(Doorkeeper::Application.find_by(id: app.id)).to be nil
```
akihikodaki: This is trivial, but `be nil` is used here while the other part uses `be_nil`. Anyway we do not have standard about that as far as I know, so it does not blocking merging.

akihikodaki (changes requested):
I request some minor changes. This change will be ready to merge after they are addressed.

akihikodaki (approved):

muffinista:
@beatrix-bitrot @akihikodaki pinging to check if/when this might be merged, and if anything else is needed to help with that? thanks!

beatrix-bitrot:
@Gargron please merge, request changes, or close this~

db/migrate/20170427011934_re_add_owner_to_application.rb
```diff
@@ -0,0 +1,7 @@
+class ReAddOwnerToApplication < ActiveRecord::Migration[5.0]
+  def change
+    add_column :oauth_applications, :owner_id, :integer, null: true
+    add_column :oauth_applications, :owner_type, :string, null: true
```
Gargron: Wait, does this have to be polymorphic? We're never gonna attach the application to anything other than `User`. If this could be only owner_id with a foreign key ensuring integrity with the users table, that'd be far better.

Gargron (changes requested):

db/migrate/20170427011934_re_add_owner_to_application.rb
muffinista: Adding the foreign key is no problem.

I'm reluctant to remove the polymorphic association because it's part of `Doorkeeper` and the underlying code expects it, and removing it could have some weird side-effects. At the very least, removing `owner_type` breaks a bunch of specs I wrote which rely on [the Application fabricator](https://github.com/muffinista/mastodon/blob/application-prefs-section/spec/fabricators/application_fabricator.rb) and while I can probably figure out a way around that issue, it seems like it would introduce a lot of potential fragility.

Gargron: Alright, that will do for now, since owner_type is essentially redundant.

Gargron (approved):

db/migrate/20170427011934_re_add_owner_to_application.rb
muffinista: Thank you!

PR #2759, opened 2017-05-03
Additional specs for URI handling

ClearlyClaire:
Add more rspec tests for remote account interaction/discovery, following commit https://github.com/tootsuite/mastodon/commit/bafd22ecf487774c252a271d668716b0e1c84c6c

### DIFF
--- a/spec/services/process_interaction_service_spec.rb
+++ b/spec/services/process_interaction_service_spec.rb
@@ -3,6 +3,7 @@ require 'rails_helper'
 RSpec.describe ProcessInteractionService do
   let(:receiver) { Fabricate(:user, email: 'alice@example.com', account: Fabricate(:account, username: 'alice')).account }
   let(:sender)   { Fabricate(:user, email: 'bob@example.com', account: Fabricate(:account, username: 'bob')).account }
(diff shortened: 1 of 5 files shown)


spec/services/process_interaction_service_spec.rb
```diff
@@ -31,6 +32,34 @@
...
+    <uri>https://webdomain.com/users/carol</uri>
+  </author>
+
+  <id>someIdHere</id>
+  <activity:verb>http://activitystrea.ms/schema/1.0/request-friend</activity:verb>
+</entry>
+XML
+
+      envelope = OStatus2::Salmon.new.pack(payload, remote_sender.keypair)
+      subject.call(envelope, receiver)
```
Gargron: I think you forgot to stub a bunch of requests in here, possibly. Travis is failing

Gargron (changes requested):

spec/services/process_interaction_service_spec.rb
ThibG: I just backported the tests I wrote for my previous pull request, but the discovery rules you used are a bit different. I'll just add the email tag to the request, as that is something we both decided to handle the same way.

Gargron (approved):

PR #2761, opened 2017-05-03
i18n: updated Polish translation

mkljczk:
#807 
forgot about this (most important) file


Gargron (changes requested):
These files have moved and are now plain JSON, not JavaScript - no comments, no scripting, etc

m4sk1n:
@Gargron what do I need to change? where's scripting in this files?

PR #2767, opened 2017-05-03
WIP - Remove in place editor

mjankowski:
This removes the best_in_place gem, which no longer works post-webpack-changes.

So far I have done this stuff:

- Change admin/settings#index to list relevant settings, instead of in-place editing
- Link to edit for each one
- An edit view to update settings

screen shot 2017-05-03 at 4 32 24 pm
screen shot 2017-05-03 at 4 32 49 pm
screen shot 2017-05-03 at 4 33 01 pm

This is good, because it's better than not working.

This feels incomplete, because the actual edit pages are a little empty. It might be nicer to have one big form on the admin settings page, similar to user prefs page, but that's a bigger change here and I was shooting for "make it not broken" as a first pass.

This build will likely break because there is a lot of missing i18n now ... but I didn't want to go fixing those until we were sure of the approach here.

Let me know if this is "good enough" for now, or if you'd prefer to go the other route (full page form) right away...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,6 @@ gem 'hamlit-rails'
 gem 'pg'
 gem 'pghero'
 gem 'dotenv-rails'
-gem 'best_in_place', '~> 3.0.1'
 
 gem 'aws-sdk', '>= 2.0'
 gem 'paperclip', '~> 5.1'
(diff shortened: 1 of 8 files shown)


mjankowski:
Closing this in favor of other WIP which uses full page form (to follow)

Issue #2619, opened 2017-04-29
Empty home timeline after period of inactivity


robcee:
Didn't log in for a week. Came back and my Home timeline was empty. Also seen on Tusky.

Whatever mechanism caches a person's Followed toots doesn't seem to work if not actively logging in.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ashfurrow:
Interesting. Thanks for opening an issue to let us know – could you go to the "Extended Instance Info" page and let us know what version number is in the bottom right corner (if it's there at all, it might not be)?

adamipc:
I had this same issue, and possibly related, my home timeline doesn't load most new toots unless I hit shift+reload in my browser.

wolfteeth:
I just bumped into this issue on an instance I hadn't logged into for a month or so. It's still pretty bewildering, and probably discouraging for users who haven't logged in in a while. The obvious conclusion to come to is "no one has posted anything since I've been gone" which leads to the not-necessarily-truthful impression that their friends are all inactive.

At the very least it seems like their timeline should collect and hold onto a certain amount of recent posts to avoid this. OR the timeline should clearly communicate that it is regenerating, and then actually regenerate -- not just start following again at the moment the user logged in.

PR #2776, opened 2017-05-03
When streaming API is disconnected, poll home/notifications

Gargron:
- Display slightly different empty home timeline message if user is following others (see #2619)
- Cull notifications to 20 items when over 40 get added in real-time, just like happens in other columns
- Run manage:translations

image


app/javascript/mastodon/features/home_timeline/index.js
```diff
@@ -12,18 +12,31 @@ const messages = defineMessages({
...
 
   render () {
-    const { intl, hasUnread } = this.props;
+    const { intl, hasUnread, hasFollows } = this.props;
+
+    let emptyMessage = <FormattedMessage id='empty_column.home' defaultMessage="You aren't following anyone yet. Visit {public} or use search to get started and meet other users." values={{ public: <Link to='/timelines/public'><FormattedMessage id='empty_column.home.public_timeline' defaultMessage='the public timeline' /></Link> }} />;
+
+    if (hasFollows) {
+      emptyMessage = <FormattedMessage id='empty_column.home.inactivity' defaultMessage="Your home feed is empty. If you have been inactive for a while, it will be regenerated for you soon." />
+    }
```
ykzts: I think it would be better to create methods like `renderEmptyMessage` & `renderEmptyMessageHasFollows`.

```javascript
const emptyMessage = hasFollows ? this.renderEmptyMessage() : this.renderEmptyMessageHasFollows();
```

app/javascript/mastodon/locales/ar.json
```diff
@@ -149,24 +151,14 @@
...
-  "notification.follow": "{name} يتبعك",
-  "notification.favourite": "{name} أعجب بمنشورك",
-  "notification.reblog": "{name} قام بترقية تبويقك",
-  "notification.mention": "{name} ذكرك",
-  "notifications.column_settings.alert": "إشعارات سطح المكتب",
-  "notifications.column_settings.show": "إعرِضها في عمود",
-  "notifications.column_settings.follow": "متابعُون جُدُد :",
-  "notifications.column_settings.favourite": "المُفَضَّلة :",
-  "notifications.column_settings.mention": "الإشارات :",
-  "notifications.column_settings.reblog": "الترقيّات:",
```
ykzts: note: duplicated with https://github.com/tootsuite/mastodon/pull/2776/files#diff-f2c27f858abca0ea3fa97f9400fca112L84

app/javascript/mastodon/features/home_timeline/index.js
Gargron: Unless that is moved out of the class I don't see how it would improve performance, since `renderEmptyMessage()` would also execute React.createElement live?

app/javascript/mastodon/locales/ar.json
Gargron: This is all a result of `yarn run manage:translations`, which sorts JSON locale files alphabetically, etc.

app/javascript/mastodon/features/home_timeline/index.js
ykzts: Yes. But, https://github.com/tootsuite/mastodon/pull/2776/files#diff-c520240100d5372b6dfb9299ff13e7deR24 is always running.

Gargron: Ah! I see what you mean. This could also be an if/else then. Fair enough

app/javascript/mastodon/locales/ar.json
ykzts: Yes. This is a comment for when other people saw it.

nightpool:
Is it common practice to add the english translations to the specific locale files? Can't we use a default message instead?

app/javascript/mastodon/features/home_timeline/index.js
ykzts: I agree.

```javascript
let emptyMessage;

if (hasFollws) {
  emptyMessage = ...;
} else {
  emptyMessage = ...;
}
```

It is also good to write like the above.

szbalint:
Do I understand it right that the fallback would happen silently from the streaming api to polling?

If so, would it make sense to configuration-gate this? Otherwise it'd be quite hard to notice if the streaming api is unexpectedly down as opposed to running on a hosting environment where it's not supported.

Gargron:
@szbalint You would see WS connection failing in console logs. What kind of configuration-gate are you thinking of?

szbalint:
@Gargron I mean for this feature from a maintainability/ops perspective:
- an instance either runs the WS streaming service in which case it's better if that fails visibly when the service is down, as falling back to notifications would mask any issues until someone checks the console
- or an instance can't/won't run WS based streaming in which case there shouldn't be connection attempts to WS/defined endpoint in the configuration file.

It looks to me that those two usecases aren't overlapping so it'd make sense for me if instance admins can decide in the `.env.production` config or somewhere else whether they run the streaming service or not and if not that's when they'd get polling as a fallback.

ykzts (approved):

PR #2780, opened 2017-05-04
Revert "Clean up redis configuration. Allow using REDIS_URL to set advanced (#2732)"

akihikodaki:
REDIS_URL was intended to be a single exhaustive configuration for Redis, but it was inappropriate because of incompatibility of Unix socket URLs.

This reverts commit c997091166ba1801eea3a587c913b020b9b84ce4.

The specification of redis scheme does not specify about transporting protocol.
https://www.iana.org/assignments/uri-schemes/prov/redis

Both of redis-rb 3.2 and Node.js redis 2.6.5, which are dependents, assumes the transporting protocol is TCP by default, but they provide different syntax to change the configuration.

redis-rb 3.2 has "unix" scheme to configure Unix socket. The scheme behaves same with redis scheme except the default transporation is Unix socket.
https://github.com/redis/redis-rb/blob/v3.2.2/lib/redis/client.rb#L393

Node.js redis 2.6.5, on the other hand, regards the URL as the path to a Unix socket if the URL lacks the scheme and succeeding two slashes.
https://github.com/NodeRedis/node_redis/blob/v.2.6.5/lib/createClient.js#L29

Because of the difference, using URL is insufficient to add an intended feature, Unix socket support. It is even problematic for environments which have already used the former configuration variables and a user-defined configuration variable for Unix socket support.

This commit reverts the change and maintains the old configuration until we make another change considering the concern.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -1,7 +1,7 @@
 # Service dependencies
-# You may set REDIS_URL instead for more advanced options
 REDIS_HOST=redis
 REDIS_PORT=6379
+# REDIS_DB=0
 # You may set DATABASE_URL instead for more advanced options
 DB_HOST=db
 DB_USER=postgres
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -1,0 +1,10 @@
+development:
+  adapter: redis
+  url: redis://localhost:6379/1
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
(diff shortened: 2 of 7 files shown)


akihikodaki:
4c455c3af7a238cf419cb7c9235765505175f2b6: rebased to 8c9116dc98ab2328d00c1713d71f4b6792c387f5

Gargron:
Can i fix the issue in node.js without reverting ago the improvements from that PR?

akihikodaki:
@Gargron So you mean to align the syntax to redis-rb and to parse URIs in Node.js by your own? It is totally fine for me.

Gargron:
@akihikodaki Is #3183 a sufficient fix?

akihikodaki:
@Gargron It works for me. Thank you!

PR #2789, opened 2017-05-04
Replace best_in_place editor on admin settings page

mjankowski:
This page is broken right now because `best_in_place` does not work.

These changes are not perfect (we should have a form object or something better here), but they do make it not-broken, and remove that gem.
screen shot 2017-05-04 at 11 14 09 am


Gargron (approved):

Gargron:
Run `i18n-tasks remove-unused` to make Travis pass

PR #2793, opened 2017-05-04
More narrowly target the admin site settings area with input styles

mjankowski:
A previous change applied styles more broadly than intended. This uses a more
narrow selector to target only the inputs on the relevant page.


mjankowski:
Dupe of https://github.com/tootsuite/mastodon/pull/2796

PR #2794, opened 2017-05-04
Fix language selector on preferences page

### LINKED ISSUE #1558: Wrong default language in settings
My instance is configured like this: first, browser language auto-detection, next ``fr``, but in my ``/settings/preferences``, language is set to english (so, basically, if I'm changing anything and press ``Save``, my interface use english).
My browser is using Fr locales.

* * * *

AA4ch1:
The user's language is detected automatically.
But the language isn't set in the language selector by default.
So when non-English users change other settings on the Preferences page (e.g. Post privacy)
and save changes, the language setting is changed to English unintentionally.

This commit fixes this issue by set detected language to the selector by default.
And maybe fix #1558.

### DIFF
--- a/app/views/settings/preferences/show.html.haml
+++ b/app/views/settings/preferences/show.html.haml
@@ -5,7 +5,12 @@
   = render 'shared/error_messages', object: current_user
 
   .fields-group
-    = f.input :locale, collection: I18n.available_locales, wrapper: :with_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }
+    = f.input :locale,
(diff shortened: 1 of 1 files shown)


mjankowski:
Oh, good catch and good change.

Is adding a blank option here to allow users to leave nothing selected better ... or is default to their i18n'd language (like your change) better?

I guess if they are on a server where the default is ok for them, it's basically the same either way, and doesn't matter if it's saved as their locale or just that they are getting it set anyway.

mjankowski (approved):

PR #2797, opened 2017-05-04
Add Thai Language

jirayudech:


app/views/user_mailer/confirmation_instructions.th.html.erb
```diff
@@ -0,0 +1,12 @@
+<p>ยินดีต้อนรับ <%= @resource.email %> !</p>
+
+<p>คุณได้สร้างแอ็คเคาท์ <%= @instance %>.</p>
+
+<p>To confirm your inscription, กรุณาคลิ๊กที่ลิงค์ข้างล่างนี้ : <br>
```
Gargron: This is untranslated? Is that on purpose?

Gargron:
You can remove the `admin.settings.click_to_edit` key

Issue #2613, opened 2017-04-29
Some public posts from followed remote users missing


wolfteeth:
I just noticed this tonight while bouncing back and forth between two instances (mastodon.social and my self-run instance).

From my instance, a few public toots from people I follow over on mastodon.social were missing on my local instance, both in my home timeline and in their local (federated) profile. Other public toots from the same users as well as others (on mastodon.social and elsewhere) federated fine and appeared in my home timeline.

I don't know enough about Sidekiq to figure out if these are jobs that died or what I can/should change to revive them.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

mjankowski:
Are the missing statuses from a time before you were following them?

If not, are they all in one big block of time, or are they spread out?

jbbdude:
I'm seeing the same. To answer those questions, no (plus others on the instance were) and... mostly bunched together or since a certain time, but there are a few exceptions mostly based on toots or @ mentions.

Lana-chan:
had this happen to me yesterday, there was a ~30 minute window in which all posts from someone i follow just did not come to our instance. after a while they started showing up again, but only the posts after that window are now in our instance. posts from other people in the same instance still came through. the posts were all public.

wolfteeth:
I have been following the person in question since I started my instance, about two weeks. I just checked out their profiles both federated and remote, and it looks like I lost about four of their public toots made in the last couple days. Two of their toots did come through yesterday, but I'm missing some before and some after.

I'm a little worried that we're losing posts for other people now that I noticed this. As a small instance we're pretty dependent on federation.

wolfteeth:
Okay, even weirder: their public toots still aren't coming through - on ANY instance I follow them from. The same toots are missing on all instances. Are unlisted public posts no longer being sent to subscribers with v1.3.x? Many of this person's public toots are unlisted.

BrunoSpy:
I have the same issue, and I think it's a really big issue.
Just compare toots visible from mastodon.social : 
https://mastodon.social/web/accounts/60680
and the toots posted : 
https://metal.odon.space/@BrunoSpy
[...]

PR #2799, opened 2017-05-04
More robust PuSH subscription refreshes

### LINKED ISSUE #2473: feature request: make better use of sidekiq instead of having cronjobs
hi,

why does mastodon need a cronjob at all?  rails apps have an event loop which could be used to kick scheduled tasks off to sidekiq for processing.  this would make deployment on serverless hosting such as heroku/scalingo/aws lambda much simpler.

### LINKED ISSUE #2702: Instance URL is not correctly substituted in Mastodon's user agent
Looking through my nginx log, I found this:
"http.rb/2.2.2 (Mastodon/1.3.2; +http://{}/)"

I believe the "{}" there should actually be the instance URL.

* * * *

- [*] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [*] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
- Fix #2473 - Use sidekiq scheduler to refresh PuSH subscriptions instead of cron
- Fix an issue where / in domain would raise exception in TagManager#normalize_domain
- PuSH subscriptions refresh done in a round-robin way to avoid hammering a single
server's hub in sequence. Correct handling of failures/retries through Sidekiq (see
also #2613). Optimize Account#with_followers scope. Also, since subscriptions
are now delegated to Sidekiq jobs, an uncaught exception will not stop the entire
refreshing operation halfway through
- Fix #2702 - Correct user agent header on outgoing http requests


Procfile.dev
```diff
@@ -1,3 +1,4 @@
 web: PORT=3000 bundle exec puma -C config/puma.rb
+sidekiq: bundle exec sidekiq -q default -q push -q pull -q mailers
```
nightpool: will this need a corresponding announcement and docs update? Can we have an obvious admin error message if people aren't running the all of the queues?

app/helpers/http_helper.rb
```diff
@@ -1,13 +1,17 @@
...
-    HTTP.headers(user_agent: USER_AGENT)
+    HTTP.headers(user_agent: user_agent)
         .timeout(:per_operation, timeout)
         .follow
   end
+
+  private
+
+  def user_agent
+    "#{HTTP::Request::USER_AGENT} (Mastodon/#{Mastodon::Version}; +http://#{Rails.configuration.x.local_domain}/)"
```
nightpool: probably better to have `@@user_agent || = "#{HTTP::Request::USER_AGENT}...` etc. so we don't introduce any possible performance regressions. (generating a new string every time instead of once)

app/services/subscribe_service.rb
```diff
@@ -5,15 +5,21 @@ def call(account)
     account.secret = SecureRandom.hex
 
     subscription = account.subscription(api_subscription_url(account.id))
-    response = subscription.subscribe
+    response     = subscription.subscribe
 
-    unless response.successful?
+    if response.code > 299 && response.code < 500 && response.code != 429
```
mjankowski: Maybe wrap this up in a well named method and remove the comments? - `response_failed?(response.code)` or something...

app/services/subscribe_service.rb
```diff
@@ -5,15 +5,21 @@ def call(account)
...
+      # An error in the 4xx range (except for 429, which is rate limiting)
+      # means we're not allowed to subscribe. Fail and move on
       account.secret = ''
-      Rails.logger.debug "PuSH subscription request for #{account.acct} failed: #{response.message}"
+      account.save!
+    elsif response.code > 199 && response.code < 300
```
mjankowski: Same here - lose comments, use method with intention revealing name.

app/workers/scheduler/subscriptions_scheduler.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+require 'sidekiq-scheduler'
+
+class Scheduler::SubscriptionsScheduler
+  include Sidekiq::Worker
+
+  def perform
+    Account.expiring(1.day.from_now).partitioned.pluck(:id) do |id|
```
mjankowski: Maybe...


- It looks like `expiring` is only ever used for this purpose ... maybe add a method like `Account.expiring_soon` which just knows about `{ 1.day.from_now }`
- Move this whole query to a method in this class like `expiring_accounts.pluck(:id)` ...

Procfile.dev
Gargron: This'll need a docs update with recommendation to run `foreman start` instead of just `rails s`, though you may not always need sidekiq during dev so I think mostly it's fine.

mjankowski (changes requested):
The `SubscribeService` is basically untested before this PR. Can we add some spec coverage as part of making these changes?

Procfile.dev
nightpool: ah, sorry, this is my mistake. I misread this line and thought you were adding a queue here.

mjankowski: Is it possible to just do `bundle exec sidekiq` here and have it detect the queues from the newly added config file ... or do we need to specify the queues here?

If we CAN rely on just the config file, we can probably also remove the list of `-q` options from the main `Procfile`

mjankowski (approved):
There's probably some follow-up refactoring to do here around consolidating the pubsub error codes response handling (which is duped a bit now), but this is good enough to get in now I think.

I left one question about sidekiq queue specification.

Procfile.dev
Gargron: Yes that should be possible

PR #2803, opened 2017-05-05
Install foreman to use it on development

takahashim:
I'd like to execute `bundle install && bundle exec foreman` on development environment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
   gem 'bullet'
+  gem 'foreman'
   gem 'letter_opener'
   gem 'letter_opener_web'
   gem 'rubocop', '0.46.0', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -151,6 +151,8 @@ GEM
       i18n (~> 0.5)
     fast_blank (1.0.0)
     ffi (1.9.18)
+    foreman (0.84.0)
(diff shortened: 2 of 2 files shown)


AA4ch1:
Have you checked #2742?
I think this is duplicated with that.

takahashim:
Thanks for your comment.
I saw #2742 and [Vagrant-guide.md]( https://github.com/tootsuite/documentation/blob/be7b16d4f8a3d2c0a6bdfd79c1d07c8b8d25bcfd/Running-Mastodon/Vagrant-guide.md). But I don't use Vagrant, and force to execute `gem install foreman` globally is not a good idea, I guess.

AA4ch1:
Oh, I'm sorry, I jumped into the wrong conclusion.
I agree with your suggestion.

mjankowski:
I'd prefer we not bundle things which are optional (I'm a foreman user), and let people opt in or out to whatever local dev setup makes sense.

Its fine to expect developers to understand how to install foreman or forego or the heroku cli, or to not want to use foreman/procfile at all.

mjankowski (changes requested):
To clarify here - I think adding docs or something to explain how you might use (and install) foreman locally would be useful. I think adding the gem version of foreman (which is only one of several foreman tools, and which itself is only one of many ways to run the app) isn't required.

Gargron:
The thing is - foreman is a global tool. Which is why I don't think bundling it is necessary.

PR #2806, opened 2017-05-05
Minor fix in Travis config with bundler call

yookoala:
* bundler_args is not functional if using custom install script in `.travis.yml`. Directly attach the argument to the install script.
* Parallel to similar scripts in the same travis config. More resiliant to changes in configuration.
* Travis + rvm seems to ignore `.bundle/config` and hence use rvm global vendor directory by default. Adding `--path` will fix this and hence make `cache.bundler = true` really functional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,12 +34,10 @@ rvm:
 services:
   - redis-server
 
-bundler_args: --without development production --retry=3 --jobs=16
-
 install:
   - nvm install
   - npm install -g yarn
-  - bundle install
+  - bundle install --without development production --retry=3 --jobs=16
   - yarn install
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ services:
 install:
   - nvm install
   - npm install -g yarn
-  - bundle install --without development production --retry=3 --jobs=16
+  - bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
   - yarn install
 
 before_script:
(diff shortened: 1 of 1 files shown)


yookoala:
Bundler cache efficiency:
* Without cache
  * bundle install: 91.39s
* With cache
  * download cache archive: 2.44s
  * restore bundler cache: 5.51s
  * bundle install: 4.23s
  * store bundle: 1 - 20s

So saved around 1 minute per platform. And will produce around 300MB of cache per branch. Probably best to disable bundler cache.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 cache:
-  bundler: true
+  bundler: false
```
mjankowski: Can you explain this change more...? This seems counter to what we'd want

.travis.yml
```diff
@@ -34,12 +34,10 @@ rvm:
 services:
   - redis-server
 
-bundler_args: --without development production --retry=3 --jobs=16
-
 install:
   - nvm install
   - npm install -g yarn
-  - bundle install
+  - bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
```
mjankowski: The bundler caching docs claim that when `cache.bundler` is turned on in Travis config, that it will use `vendor/bundle` by default - https://docs.travis-ci.com/user/caching/#Bundler - does that not happen?

yookoala: No. Not happening. The original documentation expect you to use the default install script. You may check the Cache in travis. It is empty.

.travis.yml
yookoala: I found that the yarn cache is disabled in f5bf5ebb (`cache.yarn: false`). I suppose the logic would be to save disk space (?) from inefficient cache and maybe I should follow (?).

I can revert this commit and bundler cache will be functional.

Gargron (approved):

PR #2808, opened 2017-05-05
Streamline javascript translation by improving translationRunner

yookoala:
Allow translator to use nodejs script for translation detection and stuff.

To check the language file(s) you're interested in, you may use this routine:
```
yarn install
yarn build:development
yarn manage:translations [language code]
```

If you do not pass in any language code in the last command, you'd be checking all the available languages.

If you are not sure, you may use this command:
```
yarn manage:translations -- --help
```

Improved #1742


package.json
```diff
@@ -3,6 +3,7 @@
   "license": "AGPL-3.0",
   "scripts": {
     "postversion": "git push --tags",
+    "build:babel": "babel app/javascript --out-dir ./build",
```
Gargron: Surely this should be "build" running a webpack build? I'm not sure babel alone will handle everything given there are lots of requires.

Gargron (changes requested):

package.json
yookoala: This task for me is mainly for generating the translation strings source (so I can run `yarn manage:translations` with no error). I think it is not necessary for actually generating the webpack bundle.

Open for suggestion how to name / modify it.

yookoala: Rewritten as tasks "build:production" and "build:development".

Gargron (approved):

PR #2809, opened 2017-05-05
add docker-compose.override to gitignore

haoyayoi:
For easy to update, and protect default files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ redis
 *.swp
 /public/packs
 /node_modules
+
+# Ignore Docker option files
+docker-compose.override.yml


ykzts:
I think that `.git/info/exclude` should be used.

csu:
@ykzts The `docker-compose.override.yml` file is common practice for overriding configs with Docker Compose though:

> By default, Compose reads two files, a docker-compose.yml and an optional docker-compose.override.yml file. By convention, the docker-compose.yml contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services.
https://docs.docker.com/compose/extends/#understanding-multiple-compose-files

If this file is going to be widely used (e.g. the override file could be the recommended way to set persisted volumes for Postgres/Redis), it might make sense to put it in `.gitignore`.

Gargron (approved):

PR #2821, opened 2017-05-05
Misc spec coverage improvements

mjankowski:
- Fixes warning around `raise_error` without args
- Adds  coverage for `AccountFilter`, various methods on `Subscription`, and `NotificationMailer`
- Small refactors for clarity in those classes

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 # == Schema Information
 #
 # Table name: subscriptions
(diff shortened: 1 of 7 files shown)


ykzts (approved):

app/models/subscription.rb
```diff
@@ -15,18 +16,25 @@
...
 
 class Subscription < ApplicationRecord
-  MIN_EXPIRATION = 3600 * 24 * 7
-  MAX_EXPIRATION = 3600 * 24 * 30
+  SECONDS_IN_HOUR = 3600
+  HOURS_IN_DAY = 24
+  DAYS_IN_WEEK = 7
+  DAYS_IN_MONTH = 30
+
+  MIN_EXPIRATION = SECONDS_IN_HOUR * HOURS_IN_DAY * DAYS_IN_WEEK
```
Gargron: Okay, I am actually unsure if this is more readable, it seems like less. Perhaps this could be `7.days.in_seconds` or whatever the method for that is.

Gargron (changes requested):

Gargron (approved):

PR #2823, opened 2017-05-05
Handling failed http response

alpaca-tc:
Raise readable error for failed request.

Current error isn't clearly.

```
A RuntimeError occurred in accounts#follow:

  Could not parse document
  app/services/follow_remote_account_service.rb:78:in `get_feed'
```

### DIFF
--- a/app/services/follow_remote_account_service.rb
+++ b/app/services/follow_remote_account_service.rb
@@ -75,6 +75,7 @@ class FollowRemoteAccountService < BaseService
 
   def get_feed(url)
     response = http_client(write: 20, connect: 20, read: 50).get(Addressable::URI.parse(url).normalize)
+    raise Goldfinger::Error, "Feed attempt failed for #{url}: HTTP #{response.code}" unless response.code == 200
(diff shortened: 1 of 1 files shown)


Gargron (approved):

app/services/follow_remote_account_service.rb
```diff
@@ -75,6 +75,7 @@ def account_needs_webfinger_update?(account)
 
   def get_feed(url)
     response = http_client(write: 20, connect: 20, read: 50).get(Addressable::URI.parse(url).normalize)
+    raise "Feed attempt failed for #{url}: HTTP #{response.code}" unless response.code == 200
```
Gargron: I think this should be raised as a Goldfinger::Error because we already rescue_from those in ApiController

Gargron (changes requested):

alpaca-tc:
I fixed 😸

ykzts (approved):

Gargron (approved):

PR #2830, opened 2017-05-06
Remove legacy decorators, use lodash.debounce

nolanlawson:
Size savings of this PR:

| File | Bytes before | Bytes after | Improvement |
| --- | --- | --- | --- |
| application.js | 756293 | 753756 | 0.34% |
| vendor.js | 2089022 | 2062878 | 1.25% |
 
This also removes `babel-plugin-transform-decorators-legacy` which was causing a lot of problems for me as I tried to tinker with Babel settings. In general it feels risky to rely on legacy decorator functionality, especially given that [decorators are still at stage-2 in TC39](https://tc39.github.io/proposal-decorators/).

This removes all use of decorators (which was just the `@debounce` annotation from `react-decoration`) by replacing it with a vanilla `lodash.debounce`. Note that `import { debounce } from 'lodash'` will not import all of `lodash`, because we're using `babel-plugin-lodash`.


ykzts (approved):
I think it's a good change.

PR #2834, opened 2017-05-06
Fix Scheduler::SubscriptionsScheduler

Gargron:
Looks like a missing `.each` has slipped by. Added a test for it. Also Sidekiq has its own logger, use that instead of the Rails one. And change Sidekiq production log level to warn instead of info for less noise.

### DIFF
--- a/app/workers/distribution_worker.rb
+++ b/app/workers/distribution_worker.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-class DistributionWorker < ApplicationWorker
+class DistributionWorker
   include Sidekiq::Worker
 
   def perform(status_id)
     FanOutOnWriteService.new.call(Status.find(status_id))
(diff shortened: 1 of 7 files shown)


app/workers/distribution_worker.rb
```diff
@@ -1,11 +1,11 @@
...
 
-class DistributionWorker < ApplicationWorker
+class DistributionWorker
   include Sidekiq::Worker
 
   def perform(status_id)
     FanOutOnWriteService.new.call(Status.find(status_id))
   rescue ActiveRecord::RecordNotFound
-    info("Couldn't find the status")
+    true
```
Gargron: Just as a comment for future reference: we don't really care if this status is not found. For example, user deletes status right after sending it out. It might already be gone before this job gets executed. That's okay.

mjankowski (approved):

PR #2836, opened 2017-05-06
Optimize MuteService and AfterBlockService

alpaca-tc:
- Find target statuses by id from home timeline
- Call `zrem` with multiple keys

```
Example for account with 13_000 statuses.

puts Benchmark.measure {
  MuteService.new.call(account, target_account)
}

**before**

`0.600000   0.060000   0.660000 (  2.150312)`

**after**

`0.010000   0.000000   0.010000 (  0.034723)`
```

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -76,6 +76,14 @@ class FeedManager
     end
   end
 
+  def clear_from_timeline(account, target_account)
+    timeline_key = key(:home, account.id)
+    timeline_status_ids = redis.zrange(timeline_key, 0, -1)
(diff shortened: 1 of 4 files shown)


Gargron (approved):

alpaca-tc:
Oh, I should also optimize `app/services/after_block_service.rb`. 🤔

alpaca-tc:
It seems that `FeedManager#unmerge_from_timeline` is similar to `#clear_from_timeline`.  Should I remove `#clear_from_timeline` and use `FeedManager#unmerge_from_timeline` ?

Gargron (approved):

PR #2667, opened 2017-04-30
Update mastodon.jsx

diomed:
adds Croatian [I forgot before]

adds Occitan as well

PR #2845, opened 2017-05-06
locale: add Occitan locale-data for js

yookoala:
Translation works by @Quent-in.

Resolves the Occitan part of #2667. Also pave road for other similar cases.

### DIFF
--- a/app/javascript/mastodon/containers/mastodon.js
+++ b/app/javascript/mastodon/containers/mastodon.js
@@ -60,6 +60,7 @@ import it from 'react-intl/locale-data/it';
 import ja from 'react-intl/locale-data/ja';
 import nl from 'react-intl/locale-data/nl';
 import no from 'react-intl/locale-data/no';
+import oc from '../locales/locale-data/oc';
 import pt from 'react-intl/locale-data/pt';
 import ru from 'react-intl/locale-data/ru';
 import uk from 'react-intl/locale-data/uk';
@@ -94,6 +95,7 @@ addLocaleData([
   ...ja,
   ...nl,
   ...no,
+  ...oc,
   ...pt,
   ...ru,
   ...uk,
--- a/app/javascript/mastodon/locales/locale-data/README.md
+++ b/app/javascript/mastodon/locales/locale-data/README.md
@@ -1,0 +1,129 @@
+# Custom Locale Data
+
+This folder is used to store custom locale data. These custom locale data are
+not yet provided by [Unicode Common Locale Data Repository](http://cldr.unicode.org/development/new-cldr-developers)
+and hence not provided in [react-intl/locale-data/*](https://github.com/yahoo/react-intl).
+
+The locale data should support [Locale Data APIs](https://github.com/yahoo/react-intl/wiki/API#locale-data-apis)
+of the react-intl library.
+
+It is recommended to start your custom locale data from this sample English
+locale data:
+
+```javascript
+/*eslint eqeqeq: "off"*/
+/*eslint no-nested-ternary: "off"*/
+
+export default [
+  {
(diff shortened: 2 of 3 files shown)


yookoala:
What are those "`'jsx-a11y/*' was not found`" errors in [codeclimate report](https://codeclimate.com/github/tootsuite/mastodon/pull/2845)? How do I fix them?

yookoala:
Code Climate report would be resolved if #2850 is merged.

yookoala:
A little explanation is probably needed.

The library [react-intl](https://github.com/yahoo/react-intl) provides stock locale data for some languages with `react-intl/locale-data/*`. If you traced their library source code, the whole `react-intl/locale-data` directory is generated by a [build script](https://github.com/yahoo/react-intl/blob/master/scripts/build-data.js) that use [formatjs-extract-cldr-data](https://github.com/yahoo/formatjs-extract-cldr-data) to extract data in [Unicode Common Locale Data Repository](http://cldr.unicode.org/).

Since the repository does not include every language in the earth (for obvious reason), we need an alternative way to support those languages.

The output of `react-intl/locale-data/*` should fulfill react-intl's [Locale Data API](https://github.com/yahoo/react-intl/wiki/API#locale-data-apis). It is possible to manually craft your own locale data.

I have extracted and prettified the code of the English locale data in react-intl as an example for implementation (see the [README.md](https://github.com/yookoala/mastodon/blob/e14af3f0bf9a2da6bb29e55af6ce395eac922814/app/javascript/mastodon/locales/locale-data/README.md) in this PR). I have then invited @Quent-in to [translate into Occitan](https://github.com/tootsuite/mastodon/pull/2667#issuecomment-299648588). Hence this PR for Occitan and other language of similar situation.

Gargron (approved):

yookoala:
@Quent-in: Upon re-reading your translation, I think the locale data could be wrongly done.

Please take a look and the [pluralRuleFunction here](https://github.com/tootsuite/mastodon/blob/e4b84c7ba5bc4fa92f97aa8853d07931e460f87a/app/javascript/mastodon/locales/locale-data/oc.js#L6,L13). I think the usage of the pluralRuleFunction is to provide the correct key to search for plural translation.

If the field to translate is "year" and the plural function returns "one", then the program will look for "one" in the translations that structure like this:
```
future: {
     one: "dins {0} an",
     other: "dins {0} ans"
},
```

The purpose is to find the correct plural form for a given noun. If it were English, a singular item ("one") should have a string without "s", while a plural item ("other") should have a string with "s". In Chinese, there is no separated plural form so the function always returns "other".

Now. Back to your case.

Your plural function seems to return locale strings instead of "one" or "other". But your translations are still using "one", "other" and key. It could run into problem.

My advice:
1. If your language has no plural form, the plural function should just return "other".
2. If your language has 1 plural form, the plural function should return "one" or "other. No need to translate.
3. If your language has more than 1 plural form, you should have a different key for each plural form cases (preferable in English noun like "one", "two", "three", "four"). And you should have the exact same set of keys in all your translations below.

yookoala:
If you're not sure about git, you can give me an update version of the JS here. I can commit them for you :-)

Quent-in:
Hello!
So it would be about changing line 12 : **return a ? 1 == t && 11 != r ? "un" : 2 == t && 12 != r ? "dos" : 3 == t && 13 != r ? "pauc" : "autre" : 1 == e && l ? "un" : "autre"**
?
Occitan works like English for the plurial, mainly adding an S at the end of the word.
Thanks again for your advice!

yookoala:
@Quent-in Yes. If the plural rule is that simple, I have done it in #2957. Please take a look. Thanks.

PR #2852, opened 2017-05-06
Switch uglify to compress:true, mangle:true

nolanlawson:
This has a fairly enormous impact on the JS bundle size:

| file | bytes before | bytes after | improvement |
| --- | --- | ---- | --- |
| application.js | 740K | 564K | 23.98% |
| vendor.js | 2.0M | 1.5M | 25.46% |
| application.js.gz | 112K | 100K | 10.8% |
| vendor.js.gz | 380K | 320K | 16.19% |

Both `mangle:true` and `compress:true` are safe options, although it's true they make debugging more difficult (we can always enable source maps if we really care about this; there's no cost to shipping source maps in production because it only activates when the Dev Tools are open). 

I also tinkered with using Babili in conjunction with Uglify, but I found that Uglify alone produced the same results as Babili+Uglify, so it doesn't seem worth it to use both.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -12,22 +12,8 @@ module.exports = merge(sharedConfig, {
 
   plugins: [
     new webpack.optimize.UglifyJsPlugin({
-      compress: {
-        unused: true,
-        evaluate: true,
-        booleans: true,
-        drop_debugger: true,
-        dead_code: true,
-        pure_getters: true,
-        negate_iife: true,
-        conditionals: true,
-        loops: true,
-        cascade: true,
-        keep_fargs: false,
-        warnings: true
-      },
-
-      mangle: false,
+      compress: true,
+      mangle: true,
 
       output: {
         comments: false


beatrix-bitrot:
this is awesome, also i'd love source maps since i am always debugging in prod

nolanlawson:
Yep just need to make sure the source maps are separate files (otherwise you're shipping down a gigantic comment string which gets ignored), but it can definitely be done.

Gargron (approved):

Wonderfall:
I may be wrong but does this mean the webserver (nginx) does not compress anymore? Because I prefer to let the webserver compress on the fly so I can benefit from Brotli, which is a better algorithm than gzip.

nolanlawson:
This has no impact on Nginx. This impacts the size of the JavaScript text
files, but gzip/brotli is a separate concern.

-----
Nolan Lawson
http://nolanlawson.com

PR #2863, opened 2017-05-07
Add asynchronous emojione-picker (code-splitting)

nolanlawson:
This adds Webpack-based code-splitting for the `emojione-picker`, so that the code only needs to be loaded when the user actually clicks the dropdown button.

Since it's easier to explain with a video, here's [a gfycat to show it in action](https://gfycat.com/AcidicPrestigiousBoto). Note that the third JS file is only downloaded when the smiley face is clicked. 

### Bundle size improvements

This has a big impact on the size of the JavaScript that's downloaded and executed on page load. Here's a [webpack-bundle-analyzer view](http://bl.ocks.org/nolanlawson/raw/8e6d858bf3406a5cc74c89f3f0ca44ec/):

screenshot 2017-05-06 17 12 17

The yellow is `vendor.js`, the blue is `application.js`, and the purple is the new code-split chunk. So this PR removes a huge amount of code that doesn't need to run until the user clicks the little smiley-face button.

In particular, this PR reduces the size of `vendor.js` from 1.5M to 936K (or gzipped: 331K to 237K). The new chunk file is 548K (gzipped: 9.37K).

### Loading state, error state

Since the async chunk may take awhile to download, I added a simple "pulse" animation to show the user that something is happening. It's not super prominent, but it's something at least:

naivewetbee

If the request fails entirely then nothing happens; I put that as a TODO. My hunch is we should show the user an error, but I wasn't sure how to do that. Happy to get feedback! :smiley:

### DIFF
--- a/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
+++ b/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import Dropdown, { DropdownTrigger, DropdownContent } from 'react-simple-dropdown';
-import EmojiPicker from 'emojione-picker';
 import PropTypes from 'prop-types';
 import { defineMessages, injectIntl } from 'react-intl';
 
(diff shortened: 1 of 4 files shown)


app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
```diff
@@ -54,6 +61,24 @@ class EmojiPickerDropdown extends React.PureComponent {
     this.props.onPickEmoji(data);
   }
 
+  onShowDropdown () {
+    this.setState({active: true});
+    if (!EmojiPicker) {
+      this.setState({loading: true});
+      import('emojione-picker').then(TheEmojiPicker => {
```
nolanlawson: The [dynamic import proposal](https://github.com/tc39/proposal-dynamic-import) is at stage-3 in TC39 so I feel pretty confident betting on this. [The Webpack docs also recommend it](https://webpack.js.org/guides/code-splitting-async/) over the legacy `require.ensure()`.

Gargron (approved):

PR #2864, opened 2017-05-07
Add <ostatus:conversation /> tag to Atom input/output, thread muting

### LINKED ISSUE #59: Mute replies to specific status from landing in mentions

### LINKED ISSUE #325: Converations break on federation with GNU Social.
Mastodon somehow prohibits GS servers from setting the correct conversation id on third-party servers.

### LINKED ISSUE #500: self-untag
unsubscribe from thread in a form that is transparent to users, by marking a thread and all posts as "untagged" or some such mark. this would probably federate as a post that requests users untag you, and would on mastodon, at least, make anyone who replies to the post automatically not add their @ to the post, to encourage allowing people to peacefully leave conversations without incident

Gargron:
Only uses ref attribute (not href) because href would be the alternate link that's always included also.

Creates new conversation for every non-reply status. Carries over conversation for every reply. Keeps remote URIs verbatim, generates local URIs on the fly like the rest of them.

Fix #325 
Fix #500 
Fix #59 

Update:

Adds "muted" boolean property to status JSON.

- `POST /api/v1/statuses/:id/mute`
- `POST /api/v1/statuses/:id/unmute`

In the web UI: Displays "Disable notifications" vs "Re-enable notifications" for statuses in the notifications column.


Gargron:
This almost works, for some reason `Conversation.new` in before_create on the Status is not actually saved

app/models/status.rb
```diff
@@ -242,17 +249,31 @@ def account_silencing_filter(account)
...
+  end
+
+  def set_conversation
+    self.reply = !(in_reply_to_id.nil? && thread.nil?) unless reply
+
+    if reply? && !thread.nil?
+      self.in_reply_to_account_id = (thread.account_id == account_id && thread.reply? ? thread.in_reply_to_account_id : thread.account_id)
+      self.conversation_id        = thread.conversation_id
+    else
+      self.conversation = Conversation.new
```
nightpool: don't you want .create here?

Procfile.dev
```diff
@@ -1,4 +1,4 @@
 web: PORT=3000 bundle exec puma -C config/puma.rb
-sidekiq: bundle exec sidekiq
+sidekiq: PORT=3000 bundle exec sidekiq
```
mjankowski: This seems unrelated to the other work here?

app/models/status.rb
```diff
@@ -242,17 +249,31 @@ def account_silencing_filter(account)
...
     text&.strip!
     spoiler_text&.strip!
+  end
 
-    self.reply                  = !(in_reply_to_id.nil? && thread.nil?) unless reply
-    self.reblog                 = reblog.reblog if reblog? && reblog.reblog?
-    self.in_reply_to_account_id = (thread.account_id == account_id && thread.reply? ? thread.in_reply_to_account_id : thread.account_id) if reply? && !thread.nil?
-    self.visibility             = (account.locked? ? :private : :public) if visibility.nil?
+  def set_reblog
+    self.reblog = reblog.reblog if reblog? && reblog.reblog?
```
mjankowski: This is enjoyable.

app/models/status.rb
```diff
@@ -242,17 +249,31 @@ def account_silencing_filter(account)
...
-  private
+  def set_visibility
+    self.visibility = (account.locked? ? :private : :public) if visibility.nil?
+  end
+
+  def set_conversation
+    self.reply = !(in_reply_to_id.nil? && thread.nil?) unless reply
+
+    if reply? && !thread.nil?
+      self.in_reply_to_account_id = (thread.account_id == account_id && thread.reply? ? thread.in_reply_to_account_id : thread.account_id)
```
mjankowski: I might extract either the entire thing being set here to something like `find_reply_to_account_id`, or at least the conditional part to like `reply_to_thread_account?` or something else more intention revealing.

app/services/process_interaction_service.rb
```diff
@@ -123,7 +123,9 @@ def add_post!(body, account)
   end
 
   def status(xml)
-    Status.find(TagManager.instance.unique_tag_to_local_id(activity_id(xml), 'Status'))
+    uri = activity_id(xml)
+    return nil unless TagManager.instance.local_id?(uri)
+    Status.find(TagManager.instance.unique_tag_to_local_id(uri, 'Status'))
```
mjankowski: Can you explain that these changes are doing?

Gargron: This is kind of an edge case but unique_tag_to_local_id doesn't actually confirm that the URI belongs to this instance, only local_id? does that. Lots of important endpoints check local_id? properly, but not this one for favourites. So without checking this, a favourite could be created for a completely unrelated status just by incidentally having the same numeric ID part (this happened and was very confusing for the involved parties)

Procfile.dev
Gargron: I just noticed the dev environment was broken without this.

app/models/status.rb
Gargron: If I read this line out loud I can understand it pretty quickly. "If author of replied to post is me, and the replied to post is a reply, carry over replied-to account from the replied to post, otherwise it's the author of the replied to post"

nightpool: I think expanding this to an inline if statement would be easier to parse

```ruby
self.in_reply_to_account_id =  if thread.account_id == account_id && thread.reply?
    thread.in_reply_to_account_id
else
    thread.account_id 
end
```

app/services/process_interaction_service.rb
nightpool: Is there a reason that you would call unique_tag_to_local_id *without* a local URI? Then shouldn't we just guard that method appropriately to avoid further footguns?

app/services/process_feed_service.rb
```diff
@@ -132,6 +132,8 @@ def status_from_xml(entry)
 
       return [nil, false] if account.suspended?
 
+      conversation = find_or_create_conversation(entry)
```
mjankowski: It looks like this local variable is only ever used once as an arg to `conversation: ...` ... maybe we don't need the local var, and the assignment could just be from the method?

Gargron:
Yep, it's all part of the same conversation as long as it's the same reply chain.

nightpool:
some thoughts on the API side:

It feels weird to me that in some parts of the code it's called muting a status, and in some parts it's called muting a conversation. That feels like it could be very confusing from an api dev perspective. Are you hiding notifications for the status, or for all things in that same conversation? What happens when you mute a non-root status? does it only mute descendants, or siblings statuses as well?

It's possible naming the endpoint something like mute_conversation would help here. The restful way to do it would be to expose the conversation id to clients, and then the endpoint could be /conversations/<id>/mute. That might have weird implications though (can I GET a conversation? POST it?) so I get why it's probably not the first choice

Gargron:
@nightpool "mute status" will prevent any notification (mention, reblog, fav) from being created when related to any status in the conversation. A conversation starts with a non-reply status.

Does the concept of "conversation" matter to the api users? It's an implementation detail in the backend

nightpool:
@Gargron Yes, I think that the concept of a conversation is inherent to how this feature works, and exposing it to the API any other way would be surprising.

Gargron:
Oh for god's sake, I seem to have broken tests, and also messed up log output during tests. -_-

@nightpool I don't know anymore.

mal0ki:
@Gargron you need to stop breaking things in the middle of the night. ;)

app/models/status.rb
```diff
@@ -242,17 +253,39 @@ def account_silencing_filter(account)
...
+  def set_reblog
+    self.reblog = reblog.reblog if reblog? && reblog.reblog?
   end
 
-  private
+  def set_visibility
+    self.visibility = (account.locked? ? :private : :public) if visibility.nil?
+  end
+
+  def set_conversation
```
nightpool: because this is a before_create hook, muting will fail silently on statuses created before this change was pushed.

nightpool (changes requested):
Needs some UI/UX work. If I have a chain of self-replies A -> B -> C, and I click "mute notifications" on C, I wouldn't expect that to mute notifications for A and B, which is what it does. I think changing it to "Mute conversation" or "Mute notifications for conversation" would fix this.

Currently, muting fails silently for statuses created before whenever this code is released. An asynchronous one-time sidekiq backfill that populated conversation_ids would solve this problem, although we would have to figure out how it gets run. Failing that, adding a "`populate_conversation_id_if_not_present`" call to mute and unmute would probably be the second best choice.

my concerns about [API confusion](https://github.com/tootsuite/mastodon/pull/2864#issuecomment-300226440) still stand. We now expose conversation_id to the clients, it's just a matter of hooking up an endpoint for it now.

app/models/status.rb
nightpool: Also, it seems exceedingly odd that we're populating our own conversation ids for statuses we get from federation. Shouldn't we be using the conversation uri that they give us?

app/models/status.rb
```diff
@@ -242,17 +253,39 @@ def account_silencing_filter(account)
...
+    self.reply = !(in_reply_to_id.nil? && thread.nil?) unless reply
+
+    if reply? && !thread.nil?
+      self.in_reply_to_account_id = carried_over_reply_to_account_id
+      self.conversation_id        = thread.conversation_id
+    else
+      create_conversation!
```
nightpool: I think that in the situation where Server A sends us a status with a conversation_id, but no in_reply_to id, we'll overwrite the conversation id with a new one? 

This whole hook should probably be `if: :local?`

app/models/status.rb
```diff
@@ -141,6 +143,11 @@ def non_sensitive_with_media?
     !sensitive? && media_attachments.any?
   end
 
+  before_validation :prepare_contents
+  before_create     :set_reblog
+  before_create     :set_visibility
+  before_create     :set_conversation
```
nightpool: `if: :local?`, maybe? ones from remote servers get their own conversation id.

nightpool:
Meta note: the amount of changes to locale files made this somewhat hard to review. Is there a way we can quarantine those to a separate pull request in the future? And then also we can send that around to people doing translation and get input from them at the same time?

spec/models/conversation_mute_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ConversationMute, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
mjankowski: Probably delete this file or add something here?

config/environments/production.rb
```diff
@@ -19,11 +19,6 @@
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  ActiveSupport::Logger.new(STDOUT).tap do |logger|
```
mjankowski: is this related?

app/models/status.rb
Gargron: No, I've split up a before_validation method into smaller methods, there is no change in the logic.

app/javascript/mastodon/actions/notifications.js
```diff
@@ -17,6 +18,10 @@ export const NOTIFICATIONS_EXPAND_FAIL    = 'NOTIFICATIONS_EXPAND_FAIL';
 export const NOTIFICATIONS_CLEAR      = 'NOTIFICATIONS_CLEAR';
 export const NOTIFICATIONS_SCROLL_TOP = 'NOTIFICATIONS_SCROLL_TOP';
 
+const messages = defineMessages({
+  mention: { id: 'notification.mention', defaultMessage: '{name} mentioned you' },
```
Gargron: It's a fix for another issue.

app/models/status.rb
Gargron: Correct. That's fine. It's a new feature for new statuses, it's very hard/impossible to backfill anyway.

app/models/status.rb
Gargron: You're right in that conversation should not be overwritten when it's explicitly set from ProcessFeedService

app/views/api/v1/statuses/show.rabl
```diff
@@ -2,12 +2,14 @@ object @status
...
+node(:favourited, if: proc { !current_account.nil? }) { |status| defined?(@favourites_map) ? @favourites_map[status.id]         : current_account.favourited?(status) }
+node(:reblogged,  if: proc { !current_account.nil? }) { |status| defined?(@reblogs_map)    ? @reblogs_map[status.id]            : current_account.reblogged?(status) }
+node(:muted,      if: proc { !current_account.nil? }) { |status| defined?(@mutes_map)      ? @mutes_map[status.conversation_id] : current_account.muting_conversation?(status.conversation) }
 
 child reblog: :reblog do
   extends 'api/v1/statuses/_show'
 
   node(:favourited, if: proc { !current_account.nil? }) { |status| defined?(@favourites_map) ? @favourites_map[status.id] : current_account.favourited?(status) }
   node(:reblogged,  if: proc { !current_account.nil? }) { |status| defined?(@reblogs_map)    ? @reblogs_map[status.id]    : current_account.reblogged?(status) }
+  node(:muted,      if: proc { !current_account.nil? }) { false }
```
Gargron: We can fix this later if it's problematic, but generally you only care about mutes on statuses inside notifications, which are never nested

yiskah:
@nightpool in regards to silencing posts earlier in a conversation:

From a UX perspective, you're right that silencing the entire conversation_id rather than just that branch may not be intuitive. On the flipside, it would be a pain to have to find the "root post" and silence that one specifically, if I'm getting flooded with notifications as part of harassment.

Consider this: I have been mass-tagged in a direct post. this results in my notifications being flooded with replies and it's basically DDoSing my web interface unable to get through to the root post. By the time I log on there could be hundreds already. I want to be able to just from one of those posts silence the entire conversation and all of its branches, not just that branch, because loading the thread all the back to the root post could be quite difficult. This is not just a hypothetical scenario as this has happened to me; except without the ability to silence the thread. I just had to wait it out.

nightpool:
@yiskah yeah, definitely. to be more clear: I think that its *good* that we silence the entire conversation_id. I just think that the current copy ("Mute notifications" on the post dropdown) doesn't communicate this well and can lead to very confusing situations.

Gargron:
@nightpool Counter-point: On facebook you can also "disable notification for this post" which mutes all comment notifications for some root post.

nightpool:
Facebook doesn't have post threads though? I'm missing the connection (maybe because I don't use facebook)

this is the equivalent birdsite feature:

image

app/models/status.rb
nightpool: The change in the logic is that there's a new set_conversation, that's what I'm saying should be `if: :local?`

PR #2792, opened 2017-05-04
Specify redis DB in settings to allow sharing of redis with other processes

dasoran:
This implement is what I said on issue #2188.
I made a mistake in branch operation and fix merge conflict, so I recreate this (old PR #2307)

PR #2869, opened 2017-05-07
namespace redis usage

beatrix-bitrot:
These commits add the redis-namespace gem and change a few key points where redis is used to prefix all keys and channels with whatever is in the REDIS_NAMESPACE environment variable. This allows using the same redis server for multiple mastodon instances without keyspace conflicts.

addresses issue #2188 and is similar to PR #2792


config/initializers/redis.rb
```diff
@@ -9,14 +9,16 @@
   ENV['REDIS_URL'] = "redis://#{password.blank? ? '' : ":#{password}@"}#{host}:#{port}/#{db}"
 end
 
-Redis.current = Redis.new(
+namespace = ENV.fetch('REDIS_NAMESPACE') { '' }
```
vahnj: If someone wants to not set this, does an empty namespace actually work?

vahnj:
This kind of looks like it requires you to purge subscriptions on doing the upgrade, and I'm worried that an empty namespace may break but I don't know redis namespaces well enough to know if that's actually a problem or not!

config/initializers/redis.rb
beatrix-bitrot: i don't see why not because all it is is a string that's prepended to keys when doing redis stuff, so that should be a no-op

however for the sake of full disclosure i did not test that because i'm a monster who still doesn't have a dev environment and switching to an empty namespace would have resulted in everyone seeing empty timelines on my instance i used for testing

i'm not sure i understand the concern about purging subscriptions. can you tell me more?

vahnj: I don't have a 100% grasp on how subscriptions work but considering this is on the streaming service that means that when the upgrade happens and you change to a new REDIS_NAMESPACE that means all your previous subscriptions are now invalid since the id has the namespace prepended to it

So anyway I think that means you should be regenerating those subscriptions when you restart if a new REDIS_NAMESPACE is detected bc otherwise I think the streaming service will think it has valid subscriptions to <id> instead of <namespace>:<id>

beatrix-bitrot: ah shit trash will there be a : in there even with '' namespace? if so that's a bug and i will fix it. my goal was for these changes to do nothing at all if people don't have a namespace set

your timelines would fill up again but they would be empty immediately after the update because as you say, the streaming service would be looking in a new place

the rails side will add the stuff to the new place so things will return to normal as new toots arrived, but having everything disappear would be jarring and bad i think

beatrix-bitrot: confirmed! without setting a namespace all keys have just : prepended which is not the desired outcome

i will add a lil check to see if we have a namespace set and if not,  avoid using the namespacing features entirely

good catch! <3

streaming/index.js
```diff
@@ -87,13 +87,21 @@ if (cluster.isMaster) {
   const pgPool = new pg.Pool(Object.assign(pgConfigs[env], dbUrlToConfig(process.env.DATABASE_URL)))
   const server = http.createServer(app)
   const wss    = new WebSocket.Server({ server })
-
-  const redisClient = redis.createClient({
+  const redis_namespace = process.env.REDIS_NAMESPACE || null
```
Gargron: Could these be made camelCase like all other vars in this file?

Gargron (dismissed):
:+1:

streaming/index.js
beatrix-bitrot: Yes!

Gargron (approved):

PR #2873, opened 2017-05-07
Add source maps for dev/production

nolanlawson:
As requested by @beatrix-bitrot, this adds separate source maps for development and production. 

In development mode this uses `cheap-module-eval-source-map` which supposedly is one of [the faster options](https://webpack.js.org/guides/development/#source-maps) and I can also confirm works in both Firefox and Chrome.

In production mode it creates standard `*.js.map` files, so this should be costless in production since it'll only affect users who open up their Dev Tools. I can confirm that these production source maps work too, although it probably increases the precompile time since Uglify has to do more work.


Gargron (approved):

beatrix-bitrot:
just writing to say thanks @nolanlawson can't wait to give it a try!

PR #2882, opened 2017-05-07
raise-when-oembed-not-found

masarakki:
`Addressable::URI.parse(nil)` returns `nil` ...

### DIFF
--- a/app/lib/provider_discovery.rb
+++ b/app/lib/provider_discovery.rb
@@ -22,6 +22,7 @@ class ProviderDiscovery < OEmbed::ProviderDiscovery
         format ||= :xml if provider_endpoint
       end
 
+      raise OEmbed::NotFound, url if provider_endpoint.nil?
       begin
         provider_endpoint = Addressable::URI.parse(provider_endpoint)
         provider_endpoint.query = nil


ykzts (approved):

app/lib/provider_discovery.rb
```diff
@@ -22,6 +22,7 @@ def discover_provider(url, options = {})
         format ||= :xml if provider_endpoint
       end
 
+      raise OEmbed::NotFound, url unless provider_endpoint
```
ykzts: I think that it is better to be like `raise OEmbed::NotFound, 'URL is nil' if provider_endpoint.nil?`.

`raise` can only returns `String`

ykzts (changes requested):

app/lib/provider_discovery.rb
masarakki: `url` is a given string and it is guaranteed returns status code `200`.

ykzts: Ah... I mistook `url` and `provider_endpoint`.

ykzts: Is there a time when `provider_endpoint` returns `''`? Considering that, `if provider_endpoint.present?` might be better.

masarakki: `Addressable::URI.parse` returns `nil` only when given uri is `nil`.  https://github.com/sporkmonger/addressable/blob/master/lib/addressable/uri.rb#L87

In other case, it always returns instance of it.

so there is other way like this:
```ruby
begin
   provider_endpoint = Addressable::URI.parse(provider_endpoint)
   raise OEmbed::NotFound, url unless provider_endpoint
   provider_endpoint.query = nil
   provider_endpoint = provider_endpoint.to_s
rescue Addressable::URI::InvalidURIError
   raise OEmbed::NotFound, url
end
```

masarakki: I think using `if` or `unless` is problem about coding styles,  which better?

ykzts: There are many `if val.nil?` In this repository. I think that the same way is better

masarakki: おk

ykzts (approved):

PR #2884, opened 2017-05-07
l10n: Improved RTL character count

seefood:
* Drop mentions from text before counting the text, JS function.

* Same for Ruby function, plus updates to spec by @JoelQ 

Also attn @mabkenar @yookoala @BoFFire 

Thanks!

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -51,7 +51,9 @@ module StreamEntriesHelper
     rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
 
     if rtl_characters.present?
-      total_size = text.strip.size.to_f
+      # Remove mentions before counting characters to decide RTL ratio
+      justtext = text.gsub(/@[0-9A-Za-z_@]+/, '')
+      total_size = justtext.strip.size.to_f
       rtl_size(rtl_characters.to_a) / total_size > 0.3
     else
       false
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -52,7 +52,7 @@ module StreamEntriesHelper
 
     if rtl_characters.present?
       # Remove mentions before counting characters to decide RTL ratio
-      justtext = text.gsub(/@[0-9A-Za-z_@]+/, '')
+      justtext = text.gsub(Account::MENTION_RE, '')
       total_size = justtext.strip.size.to_f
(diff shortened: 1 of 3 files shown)


mabkenar:
This is great! I really appreciate your work on this. 
Could you also remove URLs before counting RTL/LTR chars? Based on my tests, at the moment links are counted twice (I guess the entire HTML string like <a href="http://domain.tld">domain.tld</a> is counted). It would be nice if URLs are completely removed.

seefood:
Naturally that would be the next logical step, but that RegEx is _a bit_ more complicated :-)
Once I get this baby step into the code, I'll research the right way to surgically remove URLs, Latin-only hashtags too maybe? what about stock symbols? and emoji? :)

mabkenar:
Anything that is language independent by nature, but is **represented** with Latin characters should be removed IMO. Therefore, I think emojis and symbols and image upload URLs should be removed, but Latin hashtags should not.

app/helpers/stream_entries_helper.rb
```diff
@@ -51,7 +51,9 @@ def rtl?(text)
     rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
 
     if rtl_characters.present?
-      total_size = text.strip.size.to_f
+      # Remove mentions before counting characters to decide RTL ratio
+      justtext = text.gsub(/@[0-9A-Za-z_@]+/, '')
```
Gargron: Probably `Account::MENTION_RE`?

Gargron (changes requested):

app/helpers/stream_entries_helper.rb
seefood: sounds like a plan... Wish you had a ready-made RE for URLs :)

seefood:
I don't have a test env handy, so I'm closing this for now. I'll reopen once I had a change to test it (Or if one of you can please test)

mabkenar:
@seefood I absolutely think this PR should make progress. While I cannot set up a dev environment for you myself, I can gladly help by setting up a paid Mastodon instance using http://masto.host/ (which makes it possible to have it based on any git repo instead of the official Mastodon repo). If you agree, I can do it for you and give you the credentials.

seefood:
I have one set up at home, I just get to it rarely (too busy between work and family, I rarely turn on my home desktop).

But today is saturday, and I'm at the keyboard, I hope to test it some and fire it off again...

seefood:
OK! I have improved and tested the code, I'll go make a PR.

PR #2890, opened 2017-05-07
Add final newline to locale files

ykzts:
In some editors, a line feed is inserted on the last line. Unnecessary differences will appear when editing the locale file directly.

Therefore change the script and add line breaks to the locale file.


ykzts:
Please see https://github.com/tootsuite/mastodon/pull/2890/files?w=1 .

Gargron (approved):

PR #2895, opened 2017-05-07
Fix yaml parse error uk.yml

yhirano55:
```
$ bundle exec i18n-tasks health
i18n-tasks: yaml parse error: ({}): did not find expected key while parsing a block mapping at line 3 column 3
```


ykzts:
duplicated with #2896. I am sorry but I will close it.

PR #2896, opened 2017-05-07
Fix Ukrainian malformed locale submitted in #2893, i18n-tasks normalize

Gargron:


ykzts (approved):

beatrix-bitrot (approved):
looks good, let's give it a try
rather the non-uk stuff be in a separate commit but whatever

PR #2897, opened 2017-05-08
Improve allowed language handling

mjankowski:
Resolves https://github.com/tootsuite/mastodon/issues/2766

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -47,6 +47,7 @@ class User < ApplicationRecord
 
   validates :locale, inclusion: I18n.available_locales.map(&:to_s), unless: 'locale.nil?'
   validates :email, email: true
+  validate :verify_allowed_languages
 
   scope :recent,    -> { order('id desc') }
   scope :admins,    -> { where(admin: true) }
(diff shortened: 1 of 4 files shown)


ykzts (approved):

Gargron:
Can we eventually turn it around from allowed languages to filtered languages? I feel like it will be more natural for people to mute languages rather than deciding which ones to see a priori. Language detection will never be perfect and we'll never have *all* languages as options in the UI

mjankowski:
Yeah I could see it either way. What I've been picturing is:

- Users default to having an empty array of allowed languages ... but that doesn't mean "none", that means "allow all"
- We only mark something as having a language when the detector has a "reliable" detection (so like, super short strings that it can't guess just won't be marked)
- If people only speak one or two languages, and they select those as the allowed, we will not filter their HOME timeline at all, they'll get everything from everyone they follow
- In local and public they would see a combination of things which were reliably detected as one of the languages they selected, and anything which could not be detected at all, but they would not see things which were reliably detected as a language not on their list.

I guess there's an edge case there which you highlighted ... which is that if someone spoke a language that did not have an i18n entry (and lets say they were using `en` as a default in the UI) ... they would not be able to opt IN to their language, because it would not be in the list of selectable languages ... but it very well could be being marked by CLD, and they wouldn't be able to filter out languages they don't speak.

For our purposes now, these changes do fix known issues ... but I'll give that a little more thought re: the filtered languages approach.

PR #2901, opened 2017-05-08
Remove rails_12factor

alpaca-tc:
Currently `public_file_server` can not be disabled because `rails_12factor` is installed. 😕 

---

**from [rails_12factor](https://github.com/heroku/rails_12factor)**

> Rails 5
> 
> We worked with the Rails core team to make Rails 5 work on twelve-factor platforms out of the box.
> 
>## New Rails 5 Apps
> 
> If you are starting a new application with Rails 5, you do not need this gem.
> 
> ## Migrating to Rails 5
> 
> You can remove this gem after making sure the following sections are added in your production.rb file:
>
> ``` 
> config/environments/production.rb
> 
> # Disable serving static files from the `/public` folder by default since
> # Apache or NGINX already handles this.
> config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
> 
> if ENV["RAILS_LOG_TO_STDOUT"].present?
>   logger           = ActiveSupport::Logger.new(STDOUT)
>   logger.formatter = config.log_formatter
>   config.logger = ActiveSupport::TaggedLogging.new(logger)
> end
> Make sure to add both the RAILS_SERVE_STATIC_FILES and RAILS_LOG_TO_STDOUT ENV vars and set them to true. (This is done for you on Heroku)
> ```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,6 +99,5 @@ end
 
 group :production do
   gem 'lograge'
-  gem 'rails_12factor'
   gem 'redis-rails'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -325,11 +325,6 @@ GEM
       railties (~> 5.0)
     rails-settings-cached (0.6.5)
       rails (>= 4.2.0)
-    rails_12factor (0.0.3)
-      rails_serve_static_assets
-      rails_stdout_logging
-    rails_serve_static_assets (0.0.5)
-    rails_stdout_logging (0.0.5)
     railties (5.0.2)
       actionpack (= 5.0.2)
       activesupport (= 5.0.2)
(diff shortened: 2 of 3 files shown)


beatrix-bitrot:
After this change, if RAILS_LOG_TO_STDOUT is not set, where does the log go (in production)?

alpaca-tc:
After merged this PR, if RAILS_LOG_TO_STDOUT is not set, write log to `STDOUT` and `log/production.log`.
hmm, should I fix something to migrate? 😕

Gargron:
As long as we keep writing logs to STDOUT *only*, it's fine. I am against writing to logfiles, because both Docker and SystemD - our main deployment strategies use stdout of the running process to log.

alpaca-tc:
Indeed! I fixes to keep writing logs to STDOUT only 😄  👍

Gargron (approved):

PR #2907, opened 2017-05-08
Change label arrangement vertically

abcang:
I changed the layout vertically because the label was broken in the middle.
This problem also occurs in other languages such as Chinese.

The selection and file placement are not changed from landscape orientation.

### /settings/preferences
Before:
before

After:
after

### /settings/import
Before:
image

After:
image

### DIFF
--- a/app/javascript/styles/admin.scss
+++ b/app/javascript/styles/admin.scss
@@ -121,12 +121,18 @@
 
     .label_input {
       label.select {
+        display: inline;
         width: 50%;
       }
 
       select {
         width: 50%;
         float: right;
+      }
+
+      label.file {
+        display: inline-block;
+        width: 100px;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


beatrix-bitrot:
which forms does this affect? if it affects forms other than the one in the screenshots you already posted can you show some screenshots of them too?

abcang:
There is impact on the three pages in this change.

### Profile page

before


after


### Import page

before
image

after
image

### New domain block page

before


after

abcang:
I updated the code, comments and images

Gargron:
Since this PR was created, I added the special style for the language filtering section, which makes it look like this:

image

Could this PR be closed?

abcang:
Although the problem of line feed has not been solved, I thought it would be better to close this pull request and create a special CSS.

PR #2909, opened 2017-05-08
Fix broken features list in RTL languages

### LINKED ISSUE #2905: RTL on Mastodon translated /about page
Hi,
I would like to report a minor but not urgent little display problem on the /about instance page.
Note that strings are translated but the "ticked boxes" are still LTR instead of being RTL.

Arabic :
pic1
Hebrew :
pic2
Farsi :
pic3
On LTR languages seems there is no problem and is displayed "normally" :
pic4
pic5

What it *should* be like ?
pic6

Thank you for you attention.

ykzts:
Fixes #2905

### DIFF
--- a/app/javascript/styles/rtl.scss
+++ b/app/javascript/styles/rtl.scss
@@ -112,6 +112,16 @@ body.rtl {
     margin-right: 6px;
   }
 
+  .fa-ul {
+    margin-left: 0;
+    margin-left: 2.14285714em;
+  }
+
+  .fa-li {
+    left: auto;
+    right: -2.14285714em;
+  }
+
   @media screen and (min-width: 1025px) {
     .column,
     .drawer {


BoFFire:
That's correct :) Thank you @ykzts

app/javascript/styles/rtl.scss
```diff
@@ -112,6 +112,16 @@ body.rtl {
     margin-right: 6px;
   }
 
+  .fa-ul {
+    margin-left: 0;
+    margin-left: 2.14285714em;
```
seefood: 2.14286132em might look a touch better.

Can you add the same for ar and he?

Thanks!

seefood (changes requested):

app/javascript/styles/rtl.scss
ykzts: This number is based on in [font-awesome.css](https://github.com/FortAwesome/Font-Awesome/blob/v4.7.0/css/font-awesome.css#L46). It is not something that can be changed by myself.

mjankowski (approved):

app/javascript/styles/rtl.scss
seefood: ahhhh... so fa is font-awesome, not something specific to the Farsi language. OK :-)

PR #2913, opened 2017-05-08
IE11 compatible

nullpointer2017:
ie11 compatible


app/javascript/mastodon/actions/store.js
```diff
@@ -1,11 +1,11 @@
 import Immutable from 'immutable';
 
 export const STORE_HYDRATE = 'STORE_HYDRATE';
-
+const NumberIsNaN = (number) => typeof number === 'number' && isNaN(number);
 const convertState = rawState =>
   Immutable.fromJS(rawState, (k, v) =>
     Immutable.Iterable.isIndexed(v) ? v.toList() : v.toMap().mapKeys(x =>
-      Number.isNaN(x * 1) ? x : x * 1));
+      NumberIsNaN(x * 1) ? x : x * 1));
```
ykzts: `babel-polyfill` is included [`Number.isNaN`](https://github.com/zloirock/core-js/blob/v2.4.1/modules/es6.number.is-nan.js).

app/javascript/packs/application.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import 'font-awesome/css/font-awesome.css';
 import '../styles/application.scss';
-
+require("babel-polyfill");
```
ykzts: You should not `require` anything not included in package.json.

ykzts: `babel-polyfill` also "`require`" unnecessary things. You should use "useBuiltIns: true" of `babel-preset-env`.

ykzts (changes requested):

ykzts:
In my opinion, I think that there is no merit to support the environment released three years ago like IE 11.

Gargron:
I agree with @ykzts

Issue #2899, opened 2017-05-08
Rightmost column doesn't have fixed height on Safari 10.1 os x 10.12.4


kirabug:
Hallo!

When using Mastodon 1.3? on https://toot.cafe/ the rightmost column scrolls past the height of the window, even though the first 3 columns do not. See attached screenshot. 

When using Mastodon 1.2.2 on https://wandering.shop all 3 columns have fixed height.

Possible regression issue?
<img width="1440" alt="screen shot 2017-05-07 at 5 59 58 pm" src="https://cloud.githubusercontent.com/assets/28521009/25788527/d02f7d62-3377-11e7-898d-40141482999e.png">

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

stephenburgess8:
@akihikodaki This seems to have been caused by #2271 specifically applying `height: 100%;` to `.mastodon-column-container`. I tested removing this style in Firefox, Chrome, and Safari and the column scrolling and height seems to be behaving as expected. Is there some reason that this style is added that's not clear from the discussion in #2271? Any reason not to remove this style, as removing it seems to fix this bug and not cause any other bug obvious immediately?

akihikodaki:
@stephenburgess8 Open a timeline with the single column layout in Firefox (make the window small), and you will see two scroll bars on the right side.

ykzts:
@kirabug Can you solve this problem in #3062?

kirabug:
I am only a reporter of issues, not a fixer. Don't have my own instance. Thanks for taking care of it!

ttrace:
following change solved this problem.
and sorry, I' working on 1.3.3 and can't same asset for new master repository.

```
diff --git a/app/assets/stylesheets/containers.scss b/app/assets/stylesheets/containers.scss
index 6f339f9..1ba0fc5 100644
--- a/app/assets/stylesheets/containers.scss
+++ b/app/assets/stylesheets/containers.scss
@@ -11,7 +11,7 @@
 
 .mastodon-column-container {
   display: flex;
-  height: 100%;
+  height: 100vh;
   width: 100%;
```

Gargron:
I think this shouldn't be an issue in master anymore.

ttrace:
[...]

PR #2916, opened 2017-05-08
Safari bugs - Page not loading, Column Height

stephenburgess8:
Tested in Firefox, Chrome, and Safari on Mac. #2899 

- Fix bug of rightmost column to extend beyond the bottom of the frame in Safari.
- Adjust padding on the column so scrolling looks like other columns.
- Use dynamic classnames rather than dynamically added styles for the mastodon column.
- Do NOT use inline styles.
- Fix column making whole page scroll by using display: none instead of visibility: hidden.

Screenshots:

Before and After in Chrome. The difference: See how the rightmost column extends below the other columns and there is no scrollbar? Now, in the after picture there IS a scrollbar and it extends the same height as the other columns
screen shot 2017-05-13 at 12 55 03 pmscreen shot 2017-05-13 at 12 59 51 pm

Before and After Safari:
screen shot 2017-05-13 at 1 11 26 pmscreen shot 2017-05-13 at 1 14 37 pm


ykzts (approved):

stephenburgess8:
@akihikodaki Any reason why these styles would be needed that is not immediately obvious? Thanks.

akihikodaki:
@stephenburgess8 I commented to the issue. https://github.com/tootsuite/mastodon/issues/2899#issuecomment-299890927

akihikodaki (changes requested):
As I mentioned at https://github.com/tootsuite/mastodon/issues/2899#issuecomment-299890927.

akihikodaki (changes requested):
Still failing; open a timeline in multi-column layout and then make the screen smaller.

akihikodaki:
Thank you for working on this. I do not have Mac, so that helps much.

stephenburgess8:
I have been having to recompile with webpack in order to get my CSS changes to show up. After I make this change and recompile, this change shows up and works for me. Can you confirm that `overflow-y` is compiled as `auto` and not `scroll` on the `scrollable` class? That said, I would like to spend a little time with this and do some more cross-browser testing before accepting it. `scrollable` class is used in many places.

akihikodaki:
Ah, sorry. Actually webpack was failing as you pointed out. However I now see it goes wrong on Chrome with the single-column layout.

stephenburgess8:
I can return to this a little later today.

app/javascript/styles/containers.scss
```diff
@@ -9,11 +9,29 @@
...
+  }
+}
+
 .mastodon-column-container {
   display: flex;
   flex: 1 1 auto;
 }
 
+.mastodon-column-container--hidden {
+  display: none;
```
akihikodaki: This breaks scroll (and possibly other states) preservation.

akihikodaki (changes requested):
This breaks scroll (and possibly other states) preservation.
Also, lease rebase rather than creating a merge commit instead.

akihikodaki:
Perhaps it is too difficult to keep everything correct on every supported browsers.
The current design of the UI is inconsistent in terms that only the scroll bar is tied to the window, leaving the column independent of the window in the single-column layout. In the multiple-column layout, each column has the scroll bar and they are completely independent of the window.
We may need issues or pull requests to fix the inconsistency.

app/javascript/styles/containers.scss
akihikodaki: (reproducible on `Mozilla/5.0 (X11; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0`, with the single-column layout)

stephenburgess8:
I'm sorry, I hadn't registered that you were doing visibility:hidden in order to save the scroll position.

When you get a chance, if you could please check this out one more time. I have made it so that scroll position is saved. I have tested in the following browsers in both desktop view and smaller mobile view:

Safari 9
Firefox 53.0.2 
Chrome 58.0.3029.110

NOT IE11 (have not been able to get my VM to work with localhost:3000)

There is a small visual bug in Safari in Mobile view but it is still usable and it's a much better Safari experience overall now. Should have a slightly improved UI in Chrome and Firefox.

akihikodaki:
The composer died on `Mozilla/5.0 (X11; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0`, with the mobile view. :/

akihikodaki:
By the way, Mastodon does not work on IE. Test with Edge for Windows.

stephenburgess8:
The wrapper I added around mobile column must be disrupting the compose. However, I think it's getting closer. I think it's worth it to provide support for more browsers, as that makes the software accessible to more people. Thanks for your help on this.

akihikodaki:
b94f0bd5db54fdb7487f145865bbea435d01fdbc looks good for me (`Mozilla/5.0 (X11; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0` and `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36`)

stephenburgess8:
Sorry, display none was not needed. Rechecked in Safari, Chrome, Firefox

ykzts (approved):

ykzts:
@stephenburgess8 Can you solve this problem in #3062?

Gargron:
I think this is no longer a problem.

PR #2919, opened 2017-05-08
Misc gem bumps

mjankowski:
Update to [Haml 5](https://github.com/haml/haml/blob/master/CHANGELOG.md#500) in prep for Rails 5.1 (includes perf improvements), and bring some other things along as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,7 +62,7 @@ gem 'webpacker', '~>1.2'
 group :development, :test do
   gem 'fabrication'
   gem 'fuubar'
-  gem 'i18n-tasks', '~> 0.9.6'
+  gem 'i18n-tasks'
   gem 'pry-rails'
   gem 'rspec-rails'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -89,7 +89,7 @@ group :development do
   gem 'rubocop', require: false
   gem 'brakeman', '~> 3.6.0', require: false
   gem 'bundler-audit', require: false
-  gem 'scss_lint', '0.42.2', require: false
+  gem 'scss_lint', require: false
   gem 'haml_lint', require: false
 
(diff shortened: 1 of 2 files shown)


Gargron:
Wait, but we were using hamlit. I'm pretty sure

mjankowski:
Oh ... I totally misunderstood what that was doing.  Let me review...

mjankowski:
Here's a summary...

- We are in fact using `hamlit` (via `hamlit-rails`) ... I didn't know that prior to you saying as much
- The reason I had bumped `haml` version here is I noticed it was out of date, and it didn't occur to me that we were not rendering haml with the `haml` gem
- The reason `haml` gem was in the repo at all is that `haml_lint` was added here https://github.com/tootsuite/mastodon/commit/48594b18e6285d87ee2f58daa08d356b571c8aa6 - and it pulls in `haml` as a dependency

We probably don't want both `hamlit` AND `haml` gems pulled in...

I've updated this PR to remove both the explicit `haml` gem line that I had just added, and to also remove the `haml_lint` gem (not totally related ... but I'm generally skeptical of keeping tons of linting stuff in the repo itself ... we can rely on local installs for people who care, and CI or external service like codeclimate/hound/etc if we want to catch during PR process...)

Gargron (approved):

yhirano55:
Oh...different versions between local and external services (CodeClimate) will occur each different violations. This PR includes risk that many developers may confuse, and may reduce their productivity...😢 

Certainly, it is not good to increase unnecessary dependencies. But `haml-lint` was `require: false` and be grouped on `development` on Gemfile. I think the possibility of causing problems is low (At least, after dealing with problems actually, I thought that deleting was not late).

If many developers seem confused like I said, please consider reverting.

PR #2928, opened 2017-05-08
Services specs for subscribe and unsubscribe

mjankowski:


app/services/subscribe_service.rb
```diff
@@ -8,28 +8,28 @@ def call(account)
...
       raise "Subscription attempt failed for #{account.acct} (#{account.hub_url}): HTTP #{response.code}"
     end
   end
 
   private
 
+  # Any response in the 3xx or 4xx range, except for 429 (rate limit)
   def response_failed_permanently?(response)
-    response.code > 299 && response.code < 500 && response.code != 429
+    (response.status.redirect? || response.status.client_error?) && !response.status.too_many_requests?
```
Gargron: Perhaps it's just me but HTTP number codes tell me a lot more than their word names. I know what the 4xx range is, but client_error requires some extra thinking.

mjankowski: I don't feel super strongly about this particular case, but I try to avoid [magic numbers](http://wiki.c2.com/?MagicNumber) and in this case, I think the well named methods here make it easy enough to understand the failure condition we're checking for, without needing to understand the full implementation details of what the  external service might be doing.

Gargron (approved):

PR #2935, opened 2017-05-09
Fix snackbar notification stacking order

sossii:
I fixed this problem ↓

### DIFF
--- a/app/javascript/mastodon/selectors/index.js
+++ b/app/javascript/mastodon/selectors/index.js
@@ -56,7 +56,10 @@ export const getAlerts = createSelector([getAlertsBase], (base) => {
       message: item.get('message'),
       title: item.get('title'),
       key: item.get('key'),
-      dismissAfter: 5000
+      dismissAfter: 5000,
+      barStyle: {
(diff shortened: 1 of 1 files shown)


app/javascript/styles/components.scss
```diff
@@ -3343,3 +3343,7 @@ button.icon-button.active i.fa-retweet {
   transform: translate(-50%, -50%);
 }
 /* End Video Player */
+
+.notification-bar {
```
ykzts: `.notification-bar` is the class name of [`react-notification`](https://github.com/pburtchaell/react-notification). It may be overwritten.

I think it would be better to add `activeBarStyle` to [`./app/javascript/mastodon/selectors/index.js` Line 59](https://github.com/tootsuite/mastodon/blob/ef226a6f223ad3bdf662efb0ed75694b3f0bfb56/app/javascript/mastodon/selectors/index.js#L59)

sossii: Thanks for your advice!
I fixed it so.

ykzts (approved):

PR #2938, opened 2017-05-09
Add AGPL-3+ copyright notice in source files

ghost:
To state the license of Mastodon in a way that is legally clear,
and avoid risk that the code will get disconnected from its license
would you please install these changes?  Without a license notice
in each file, it is easy for the code to be used in a way that
leaves users in ignorance of how they are allowed to use it.

Also, when using the GNU AGPL (or the GNU GPL), there is a decision
you need to make and state: is it version 3 only, or version 3-or-later?
The place you state the decision is in the per-file license notice.
The patch here uses version 3-or-later, but you could change it if
that's not your choice.

Loic Dachary <loic@gnu.org> & Richard M. Stallman <rms@gnu.org>

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -1,4 +1,21 @@
 # frozen_string_literal: true
+#
+# Mastodon, a GNU Social-compatible microblogging server
+# Copyright (C) 2016-2017 Eugen Rochko & al (see the AUTHORS file)
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
 
 class AboutController < ApplicationController
   before_action :set_body_classes
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,3 @@
+Eugen Rochko <eugen@zeonfederated.com>
+Shel R <Yiskah.Raphen@gmail.com>
+Effy Elden <github@effy.is> <git@effy.is>
--- a/.mailmap
+++ b/.mailmap
@@ -1,3 +1,5 @@
 Eugen Rochko <eugen@zeonfederated.com>
 Shel R <Yiskah.Raphen@gmail.com>
 Effy Elden <github@effy.is> <git@effy.is>
+Isabelle Knott <isabelle@blackle-mori.com>
+Kibigo <marrus-sh@users.noreply.github.com>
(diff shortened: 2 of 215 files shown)


app/controllers/about_controller.rb
```diff
@@ -1,3 +1,20 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
 # frozen_string_literal: true
```
ykzts: This magic comment must be on line 1.

ykzts (changes requested):

dachary:
@ykzts oops :-) Fixed and repushed, thanks for the hint !

akihikodaki:
Nice to see this change.

> Also, when using the GNU AGPL (or the GNU GPL), there is a decision
you need to make and state: is it version 3 only, or version 3-or-later?

Contributors does not give copyrights to an entity but they own copyrights for their code in this project, right? How could we make a consensus?

ykzts (approved):
I have only confirmed that the syntax of Ruby is not broken. I have not decided whether to merge this patch or not.

dachary:
> Contributors does not give copyrights to an entity but they own copyrights for their code in this project, right? How could we make a consensus?

First, the people who contributed a lot of code should discuss it and reach an agreement.

Loic Dachary &lt;loic@gnu.org> & Richard M. Stallman &lt;rms@gnu.org>

zonuexe:
@dachary 
> First, the people who contributed a lot of code should discuss it and reach an agreement.

Can we use GitHub's issue to form consensus?  Also, do we need to gather consent from all contributors?

There is https://github.com/mpv-player/mpv/issues/2033 in the precedent of an attempt to form an agreement using GitHub.

dachary:
> Can we use GitHub's issue to form consensus? 

Possibly, yes.

> Also, do we need to gather consent from all contributors?

The consent of all copyright holders is necessary. Being the author of a commit does not automatically make you a copyright holder though. If the people who contributed a lot of code want this change it will be worth reaching out to all of them and I'm willing to help with that.

@Gargron what do you think ?

zonuexe:
I think it is important to have the CLA(Contributor License Agreement) agree on new contributors.

For example, [CLA-assistant](https://cla-assistant.io/) is integrated with GitHub's PR, and you can save the consent of all contributors.
The CLA of my contribution for other project in the past has been saved as follows: https://cla-assistant.io/flycheck/flycheck?pullRequest=1045

TrashMacNugget:
> Also, do we need to gather consent from all contributors?

Not necessarily; right now, the only license information in the project is the LICENSE file with the AGPLv3 text. The AGPLv3 says:

> If the Program does not specify a version number of the GNU Affero General Public License, you may choose any version ever published by the Free Software Foundation.

Since nowhere in this repository is a version specified, it is completely legal to use Mastodon under a newer version of the AGPL if/when one comes out, and completely legal to say that the project is AGPLv3 or any later version.

Gargron:
We have added an AUTHORS file and a copyright notice into the README separately, so I think I'm going to close this PR. As much as it would be cool to have commits authored by RMS in the history, I don't like the idea of putting so much duplicate text into all the files. It's a big burden, any time you want to add a new file, you have to add that text. I have looked over Discourse, which is a GPL2 project, and they also don't have a license header in every Ruby file, so I think it's an okay choice.

Sorry it took a year to make a decision on this. I wasn't sure until now.

PR #2943, opened 2017-05-09
add content-type in pubsubhubhub request header

haoyayoi:
fix to https://github.com/tootsuite/mastodon/issues/2923

### DIFF
--- a/app/workers/pubsubhubbub/delivery_worker.rb
+++ b/app/workers/pubsubhubbub/delivery_worker.rb
@@ -18,7 +18,7 @@ class Pubsubhubbub::DeliveryWorker
     return if DomainBlock.blocked?(host)
 
     headers['User-Agent']      = 'Mastodon/PubSubHubbub'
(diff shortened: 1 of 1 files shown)


app/workers/pubsubhubbub/delivery_worker.rb
```diff
@@ -18,6 +18,7 @@ def perform(subscription_id, payload)
     return if DomainBlock.blocked?(host)
 
     headers['User-Agent']      = 'Mastodon/PubSubHubbub'
+    headers['Content-Type']    = 'application/x-www-form-urlencoded'
```
mjankowski: Is this actually correct...?

This is supposed to inform a subscriber about what content we're sending them, so I think something like `application/atom+xml` would be expected?

mjankowski (dismissed):

app/workers/pubsubhubbub/delivery_worker.rb
Gargron: :+1:

Gargron (approved):

PR #2947, opened 2017-05-09
Fixed bug that timeline can not be displayed by InvalidURIError

abcang:
When InvalidURIError occurs, 500 error occurs.
In this case the user's timeline is not displayed.

I modified to not convert invalid URL.
A URL containing a backslash in its host name is one of invalid URLs.

### DIFF
--- a/spec/lib/formatter_spec.rb
+++ b/spec/lib/formatter_spec.rb
@@ -123,6 +123,13 @@ RSpec.describe Formatter do
         expect(subject).to match '<p>&lt;img src=&quot;javascript:alert(&apos;XSS&apos;);&quot;&gt;</p>'
       end
     end
+
+    context 'contains invalid URL' do
+      let(:local_text) { 'http://www\.google\.com' }
+      it 'has valid url' do
+        expect(subject).to eq '<p>http://www\.google\.com</p>'
+      end
+    end
   end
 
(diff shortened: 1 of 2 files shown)


spec/lib/formatter_spec.rb
```diff
@@ -123,6 +123,13 @@
         expect(subject).to match '<p>&lt;img src=&quot;javascript:alert(&apos;XSS&apos;);&quot;&gt;</p>'
       end
     end
+
+    context 'contains invalid URL' do
+      let(:local_text) { 'http://www\.google\.com' }
+      it 'has valid url' do
+        expect(subject).to match '<p>http://www\.google\.com</p>'
```
akihikodaki: Could you tell me why you used `match` instead of `eq`?

akihikodaki (changes requested):
I found a suspicious thing. Please review. Thanks.

spec/lib/formatter_spec.rb
abcang: There is no deep reason.
I will change it to `eq`.

Gargron (approved):

akihikodaki (changes requested):
Could be squashed?

akihikodaki (approved):
Looks good to me!

PR #2948, opened 2017-05-09
Use PostgreSQL inheritance for blocks and mutes

akihikodaki:
This commit is expected to have the feature brought with commit 5135d609b774b177d3d3894b176a822d86b73d3c, but includes the following additional changes:

* Fix the version of the schema
* Change schema format to `:sql` to preserve inheritance
* Disable ID column when migrating
* Support reverting migration
* Drop `parallel:prepare`, which is not required and fails with `structure.sql`, from `.travis.yml`


mjankowski:
I understand what this is doing ... but it brings a lot of complexity for seemingly not that much in return.

If the primary concern is that these models are very similar to each other and we're duplicating a lot of things ... why don't we add a `muted?` boolean column to the `blocks` table, and drop `mutes` table entirely? (or vice versa)

If there's another concern or improvement here, can you elaborate?

akihikodaki:
> why don't we add a muted? boolean column to the blocks table, and drop mutes table entirely? (or vice versa)

For performance; see https://www.postgresql.org/docs/9.6/static/ddl-partitioning.html.

> If there's another concern or improvement here, can you elaborate?

I'm considering an improvement in streaming API server. The current implementation queries as many times as the number of listening connections, but it could be just one query to allow PostgreSQL optimizations.
```SQL
SELECT account_id FROM block_mutes WHERE account_id = ANY($1) AND target_account_id = ANY($2) GROUP BY account_id
```
https://github.com/kagucho/mastodon/commit/7c4588650463e9715ab4dcfca82f3e65bac65103#diff-a5bbad982dece712d3213154fc6e0b06R122

akihikodaki:
> If the primary concern is that these models are very similar to each other and we're duplicating a lot of things ... why don't we add a muted? boolean column to the blocks table, and drop mutes table entirely? (or vice versa)

If you are concerning the duplicate of models, we can make operations against `block_mutes`. They would autlmatically be reflected to `block` and `mute` table. Do you think that is less complicated?

akihikodaki:
e5759277232dc84cdb2099ee5373960bf51017ee: now `Block` and `Mute` have the interface same with the existing one.

mjankowski (changes requested):
I think I'm a 👎  on this change ... seems too much complexity for what it gets us.

If there are specific performance issues with that query, can you benchmark those and we can think about other ways to improve?

akihikodaki:
@mjankowski What do you think it is complicated? It has the same interface in ActiveRecord. Possibly I am missing something, so please tell me if you have an idea for that.

The problem I want to solve is this query I had shown the above:
```SQL
SELECT account_id FROM block_mutes WHERE account_id = ANY($1) AND target_account_id = ANY($2) GROUP BY account_id
```
This query should have to be divided in two queries if we do not have partitions:
```SQL
SELECT account_id FROM blocks WHERE account_id = ANY($1) AND target_account_id = ANY($2) GROUP BY account_id
```

```SQL
SELECT account_id FROM mutes WHERE account_id = ANY($1) AND target_account_id = ANY($2) GROUP BY account_id
```

We would have redundant operations in the next query; possibly some `account_id` may be already known it is muting or blocking, but the query does not care. Partitioning is a less complicated solution for that, in my opinion.

saper:
couldn't `blocks` and `mutes` be just views on the new table?

PR #2957, opened 2017-05-10
Fix Occitan's locale data plural function

yookoala:
* Fix the pluralRuleFunction of the locale data.
* Added more descriptions about how to write it.


yookoala:
@Quent-in Just updated the plural rule function here.

Quent-in:
It looks perfect to me!
Many thanks for your work on that subject!

Gargron (approved):

PR #2967, opened 2017-05-10
Optimize PrecomputeFeedService 30x

alpaca-tc:
Currently slow queries are occurred in `PrecomputeFeedService`.
Honestly, I can't explain this problem and solution in detail. I'm not familiar with PostgreSQL.

- This problem often occurred a day.
- I confirmed amount of account's followers. Account has 10 ~ 1000> followers.
- It seems that the problem is caused by the query plan.
  - I don't know why, but this PR way I solved it.

I want to ask to for your advise about this PR. 😕  🙏 

---

**before**

`EXPLAIN ANALYZE SELECT "statuses".* FROM "statuses" WHERE "statuses"."account_id" IN (1, 2, 3...) ORDER BY id desc LIMIT 100;`

```
Limit  (cost=0.43..2113.78 rows=100 width=316) (actual time=978.410..12230.796 rows=100 loops=1)
 ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.43..16140231.67 rows=763730 width=316) (actual time=978.409..12230.770 rows=100 loops=1)
  Filter: (account_id = ANY ('{1, 2, 3...}'::integer[]))
       Rows Removed by Filter: 1701437
Planning time: 2.141 ms
Execution time: 12230.841 ms
(6 rows)
```

**after**

Add `account_id desc` to `ORDER BY` section.

`EXPLAIN ANALYZE SELECT "statuses".* FROM "statuses" WHERE "statuses"."account_id" IN (1, 2, 3...) ORDER BY id desc, accont_id desc LIMIT 100;`

```
Limit  (cost=1273363.52..1273363.77 rows=100 width=316) (actual time=406.579..406.605 rows=100 loops=1)
 ->  Sort  (cost=1273363.52..1275272.85 rows=763730 width=316) (actual time=406.579..406.597 rows=100 loops=1)
       Sort Key: id DESC, account_id DESC
       Sort Method: top-N heapsort  Memory: 65kB
       ->  Index Scan using index_statuses_on_account_id on statuses  (cost=0.43..1244174.31 rows=763730 width=316) (actual time=0.058..388.025 rows=177439 loops=1)
        Index Cond: (account_id = ANY ('{1, 2, 3...}'::integer[]))
Planning time: 2.140 ms
Execution time: 406.672 ms
(8 rows)
```

### DIFF
--- a/app/services/precompute_feed_service.rb
+++ b/app/services/precompute_feed_service.rb
@@ -6,7 +6,8 @@ class PrecomputeFeedService < BaseService
   # @param [Account] account
   def call(_, account)
     redis.pipelined do
-      Status.as_home_timeline(account).limit(FeedManager::MAX_ITEMS / 4).each do |status|
+      # NOTE: Added `id desc, account_id desc` to `ORDER BY` section to optimize query.
+      Status.as_home_timeline(account).order(account_id: :desc).limit(FeedManager::MAX_ITEMS / 4).each do |status|
         next if status.direct_visibility? || FeedManager.instance.filter?(:home, status, account.id)
         redis.zadd(FeedManager.instance.key(:home, account.id), status.id, status.reblog? ? status.reblog_of_id : status.id)
       end


Gargron (approved):

beatrix-bitrot:
this is awesome

S-YOU:
May be this is the best with heapsort, and you have mixed index, which used less memory, stop when limit is reached. 

https://madusudanan.com/blog/all-you-need-to-know-about-sorting-in-postgres/#Heapsort

And first one removed 1701437 rows after it loaded into memory. (Rows Removed by Filter: 1701437)

alpaca-tc:
Sorry, unfortunately it seems that this PR contains the side effect.

- (too slow-> fast) If account's followers don't toots much: sorting records are faster than primary key scan. This is nice for me.
- side effect - (fast -> slow) If account has many followers and followers frequently toots: sorting records are slower than primary key scan. 😕 

I will see a slow log of pawoo.net for a while. 🙏

Gargron:
If it's any consolation, this query should be fairly rare and only done in background processing.

alpaca-tc:
@Gargron It seems that CPU usage has been optimized.

924b536afd02fe5d94c274d2d526c7f1_720

PR #2968, opened 2017-05-10
Enhance account filter

alpaca-tc:
I enhance AccountFilter to search by username, display_name, email and ip.
In the daily management work of mastodon, the search form is needed for me.

I have not changed the current behavior, but maybe this PR is a bit complicated.
If you do not like this, please reject. it's okay 😄  👍

### DIFF
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -23,7 +23,11 @@ module Admin
         :by_domain,
         :silenced,
         :recent,
-        :suspended
+        :suspended,
+        :username,
+        :display_name,
+        :email,
+        :ip
       )
     end
   end
--- a/app/views/admin/accounts/index.html.haml
+++ b/app/views/admin/accounts/index.html.haml
@@ -20,6 +20,21 @@
       %li= filter_link_to t('admin.accounts.order.alphabetic'), recent: nil
       %li= filter_link_to t('admin.accounts.order.most_recent'), recent: '1'
 
+= form_tag(admin_accounts_url, { method: 'GET', class: 'simple_form' }) do
+  .fields-group
+    - Admin::FilterHelper::ACCOUNT_FILTERS.each do |key|
+      - if params[key].present?
+        = hidden_field_tag key, params[key]
+
+    - %i(username display_name email ip).each do |key|
+      .input.string.optional
+        .label_input
+          %label.string.optional= t("admin.accounts.#{key}")
+          = text_field_tag key, params[key], class: 'string optional'
+
+    .actions
+      %button.btn= t('admin.accounts.search')
+
 %table.table
   %thead
     %tr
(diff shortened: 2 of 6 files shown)


Oscardono:
Rebeca usen name and displsy name its the same

Gargron:
Travis failure due to "unused" keys: `activemodel.attributes.form/account_filter.keyword` etc. Seems like a weird place to put strings - are you not using simple_form? It could be a better fit here.

app/controllers/admin/accounts_controller.rb
```diff
@@ -12,19 +15,19 @@ def show
...
         :remote,
         :by_domain,
         :silenced,
         :recent,
-        :suspended
+        :suspended,
+        :search_type,
+        :keyword
       )
+    rescue ActionController::ParameterMissing
```
mjankowski: You can probably use `fetch` with a `{}` default, instead of the rescue.

app/controllers/admin/accounts_controller.rb
```diff
@@ -2,8 +2,11 @@
 
 module Admin
   class AccountsController < BaseController
+    helper_method :account_filter_params
+
     def index
-      @accounts = filtered_accounts.page(params[:page])
+      @account_filter = Form::AccountFilter.new(account_filter_params)
```
mjankowski: Why the rename to `Form::...` ?

app/views/admin/accounts/index.html.haml
```diff
@@ -5,20 +5,33 @@
...
+      %li= filter_link_to t('admin.accounts.order.most_recent'), form_account_filter: account_filter_params.merge(recent: '1')
+
+= simple_form_for @account_filter, url: admin_accounts_path, method: :get do |f|
+  = render 'shared/error_messages', object: f.object
+
+  .fields-group
+    - account_filter_params.each do |key, value|
+      = f.hidden_field key, value: value
+    - search_types = f.object.class::SEARCH_TYPES.map { |key| [f.object.class.human_attribute_name("search_type.#{key}"), key] }
+    = f.input :search_type, collection: search_types, wrapper: :with_label
```
mjankowski: Could this just be one text field that searches across all of these things?

mjankowski (dismissed):

app/controllers/admin/accounts_controller.rb
alpaca-tc: well... I thought that the module included `ActiveModel::Model` should be name `Form::...` in this project.

app/views/admin/accounts/index.html.haml
alpaca-tc: Oh, if replaced each search types with simple attributes, I can implement form simply. I refactor them, please wait a while.

alpaca-tc:
@mjankowski I refactor this PR 😄  👍

Gargron (approved):

PR #2970, opened 2017-05-10
Fix regressions from #2683

Gargron:
- Properly format spoiler text HTML, while keeping old logic for blankness intact
- Process hashtags and mentions in spoiler text
- Format spoiler text for Atom
- Change "show more" toggle into a button instead of anchor
- Fix style regression on dropdowns for detailed statuses

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -630,7 +630,7 @@
       right: initial;
     }
 
-    &:after {
+    &::after {
       bottom: initial;
       margin-left: 7px;
       margin-top: -7px;
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -474,15 +474,18 @@
   }
 }
 
-a.status__content__spoiler-link {
+.status__content__spoiler-link {
   display: inline-block;
   border-radius: 2px;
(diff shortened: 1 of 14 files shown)


app/javascript/styles/components.scss
```diff
@@ -474,15 +474,18 @@
   }
 }
 
-a.status__content__spoiler-link {
```
beatrix-bitrot: does this refer to links within the spoiler text or the 'link' used to show/hide the status content?

if the latter, should this still be called blabla-link or should it be blabla-button since you changed it from a link to a button?

beatrix-bitrot (commented):
please check app/javascript/mastodon/actions/notifications.js

and see if you need to unescape or otherwise process the spoiler text when it's used in desktop notifications. the status contents are, so probably the spoiler text now should be as well since it now contains html.

also I love that you are always refactoring and fixing style issues but it makes reviewing harder/more time consuming, so I'd love it even MORE if that happened in a separate changeset (whenever possible ofc)

app/javascript/styles/components.scss
Gargron: It looks like a link

ykzts (approved):

PR #2981, opened 2017-05-11
Introduce common JavaScript file

akihikodaki:
This is used by `/settings/two_factor_authentication`.

### DIFF
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -13,10 +13,7 @@ function main() {
   const Mastodon = require('mastodon/containers/mastodon').default;
   const React = require('react');
   const ReactDOM = require('react-dom');
-  const Rails = require('rails-ujs');
   window.Perf = require('react-addons-perf');
-
-  Rails.start();
 
   require.context('../images/', true);
   require.context('../../assets/stylesheets/', false, /custom.*\.scss$/);
--- a/app/javascript/packs/common.js
+++ b/app/javascript/packs/common.js
@@ -1,0 +1,2 @@
+import { start } from 'rails-ujs';
+start();
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -3,11 +3,8 @@ import { length } from 'stringz';
 import { default as dateFormat } from 'date-fns/format';
 import distanceInWordsStrict from 'date-fns/distance_in_words_strict';
 import { delegate } from 'rails-ujs';
-import Rails from 'rails-ujs';
 
 require.context('../images/', true);
-
-Rails.start();
 
 const parseFormat = (format) => format.replace(/%(\w)/g, (_, modifier) => {
   switch (modifier) {
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -18,9 +18,8 @@
         = ' - '
       = title
 
-    = stylesheet_pack_tag 'vendor', media: 'all'
     = stylesheet_pack_tag 'application', media: 'all'
-    = javascript_pack_tag 'vendor', integrity: true, crossorigin: 'anonymous'
+    = javascript_pack_tag 'common', integrity: true, crossorigin: 'anonymous'
     = csrf_meta_tags
 
     = yield :header_tags
--- a/config/webpack/shared.js
+++ b/config/webpack/shared.js
@@ -40,8 +40,8 @@ module.exports = {
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
     new ManifestPlugin({ fileName: paths.manifest, publicPath, writeToFileEmit: true }),
     new webpack.optimize.CommonsChunkPlugin({
-      name: 'vendor',
-      minChunks: ({ resource }) => /node_modules/.test(resource)
+      name: 'common',
+      minChunks: 2
     })
   ],
 


ykzts (commented):
It is also explicitly written in [public.js](https://github.com/tootsuite/mastodon/blob/9cd3a6836b952bda507734b6a476dfaa9b4640fb/app/javascript/packs/public.js#L10). I do not think that this change is necessary.

ykzts:
Also in the context of Rails "vendor" refers to external libraries. You should not write common code in "`vendor.js`".

akihikodaki:
> It is also explicitly written in public.js. I do not think that this change is necessary.

Actually it is. `two_factor_authentication` does not use `public` layout, and other pages using `application` layout could use. That is why `Rails.start` should be in `vendor.js`. I will rebase to a commit after 72698bc3b49925a2b2955f32e5a562c1eecd729b and add a change to remove `Rails.start` from `public.js` as well.

> Also in the context of Rails "vendor" refers to external libraries. You should not write common code in "vendor.js".

Currently `vendor.js` is a file to contain common code extracted by `CommonsChunkPlugin`, without carrying the meaning of vendor libraries. I agree in the point that it should be renamed, but the problem is not introduced by this change and should be done in another change.

ykzts:
"`vendor.js`" is only for files under node_modules. Is not this an external library?

akihikodaki:
> "`vendor.js`" is only for files under node_modules. Is not this an external library?

Yes! I missed the line. I was wrong. 
I think we have better a chunk as a set of common modules, rather than one as a set of external libraries since it is imported by `application` layout, common for all pages. I will add a change to introduce `common.js` or something kind of, then add this change.

app/javascript/packs/vendor.js
```diff
@@ -0,0 +1,2 @@
+import {start} from 'rails-ujs';
```
ykzts: change to `import { start } from 'rails-ujs';`. add space.

ykzts (changes requested):

Gargron:
vendor.js is automatically generated from node_modules in the webpack config. I'm up for switching that to a common chunk.

akihikodaki:
@Gargron I've already made a change (commit 14d706182ae0c274dfb4ab89c7c61aa99c93abfa) to make a common chunk. I will add the change to this pull request since the change of rails-ujs depends on the file name.

akihikodaki:
14d706182ae0c274dfb4ab89c7c61aa99c93abfa and 4fa6ed019580b97a86281cee46c9ae41e3499eae introduces a common JavaScript file. 4fa6ed019580b97a86281cee46c9ae41e3499eae also includes a style fix for https://github.com/tootsuite/mastodon/pull/2981#discussion_r115905176.

akihikodaki:
fc7afaadf1f55c57220a99bca763643026b1d863 and ed82dcc86a617fc26218150e2a43711b8698cc16: rebased to a822f7a05ac4337d45cfd930bd0d779eaf07ad57.

nolanlawson:
I took a look at the change in JS bundle sizes. The `0-` and `1-` bundles were unchanged, so besides those:

Before:

```
576K	application-93debc98294691b19d85.js
8.0K	public-413a17ceb408901cf70e.js
800K	vendor-66bd28ca0514a0a122da.js
```

After:

```
1.1M	application-e6c988830cbc6dfa0818.js
316K	common-a72f9a3ad30faf5cfbf0.js
16K	public-e45723b1a0af00ed5219.js
```

So if the goal is to move more JS into the landing page (`common.js`) where it can be cached before users log in, then this is successful. The separate vendor (`node_modules`) bundle may not have been buying us much, unless the `vendor` bundle was changing less frequently than the `application.js` bundle, which seems unlikely to me.

nolanlawson:
CSS for those interested:

Before:

```
104K	application-04a4ba1c05de9b7bc2ab.css
32K	vendor-04a4ba1c05de9b7bc2ab.css
```

After:

```
136K	application-56249d4193940135bb90.css
```

akihikodaki:
@ykzts ping?

ykzts:
I can not find a reasonable reason to require this change. Why do you write in the description whether you need to make such a change?

akihikodaki:
@ykzts I had a description in the commit message. Sorry if I should have wrote one also in pull request comments.
> vendor chunk is a set of modules provided by external vendors, but now we can have a chunk as a set of modules shared by multiple entry points, which could be more efficent than having vendor chunk.

ykzts:
This will do unnecessary processing when another entry file that does not need Rails.start is added, do you think that the problem will not occur?

akihikodaki:
@ykzts Yes, but in the case probably we also need to reconsider the configuration of `CommonsChunkPlugin`; the form of `common.js` or `vendor.js` is highly dependent on other scripts in nature, whether we have `Rails.start` or not.

nolanlawson:
Perf-wise I'd agree it's roughly the same. It appears currently public.js+vendor.js is loaded at `/about`, then application.js is added when you get to `/web/getting-started`, whereas with this PR, it would be public.js+common.js, then application.js. So actually my previous statement about loading more JS upfront is wrong; the existing `vendor.js` system loads more upfront.

Depends on whether we think it's more valuable to preload some extra JS upfront, or have a lightweight landing page.

akihikodaki:
> Depends on whether we think it's more valuable to preload some extra JS upfront, or have a lightweight landing page.

I have no idea of situations preloading makes sense. It may speed up loading of the home layout, but users should wait longer anyway. Also, (as far as I know) no one claims it takes too long to show the home layout.
Reducing transfers always benefits for servers.

akihikodaki:
In my opinion, we have the following two reasonable options:
* Have `common.js` instead of `vendor.js` if we want to minimize transfers. However we should reconsider the configuration when we have another entry point.
* Have an additional entry point to call `Rails.start`, and remove `vendor.js`. We could simply create another entry point if we want, but entry points get bigger.

Gargron (approved):

ykzts (approved):
There is no problem with the code so we will leave the merge decision to @Gargron

nolanlawson (approved):
> I have no idea of situations preloading makes sense

@akihikodaki There's a tradeoff between preloading and on-demand loading, but in this case I'd say I like your PR because a lightweight landing page is a good thing; it very quickly tells someone "this is what Mastodon is" and then the additional JavaScript is only loaded later once the user is invested enough to want to signup/login.

> Also, (as far as I know) no one claims it takes too long to show the home layout.

I have been claiming this. :wink: Especially if you're on a mobile device or a slow connection, excessive resources can really slow down a page load. E.g. here's [a WebPageTest trace](https://www.webpagetest.org/result/170514_KS_P3Q/) for https://mastodon.social throttled to 3G; you can see it takes 4.3 seconds to render and 9.4 seconds to load, which is a lot considering that [half of users will abandon a page if it takes longer than 3 seconds to load](https://www.doubleclickbygoogle.com/articles/mobile-speed-matters/).

According to the waterfall view of that trace, the JS is render-blocking, so anything to reduce that initial JS payload is a good thing. So from that perspective I'm +1 on this PR. :smile:

nolanlawson:
BTW to be clear I meant "roughly the same" in terms of `/web`; in terms of the landing page this PR is definitely faster because it ships less JS.

akihikodaki:
> > Also, (as far as I know) no one claims it takes too long to show the home layout.
>
> I have been claiming this. 😉 Especially if you're on a mobile device or a slow connection, excessive resources can really slow down a page load. E.g. here's [a WebPageTest trace](https://www.webpagetest.org/result/170514_KS_P3Q/) for https://mastodon.social throttled to 3G; you can see it takes 4.3 seconds to render and 9.4 seconds to load, which is a lot considering that [half of users will abandon a page if it takes longer than 3 seconds to load](https://www.doubleclickbygoogle.com/articles/mobile-speed-matters/).

It is surprising. I was not aware of that, probably because I do not have a mobile connection and my Japanese friends have 4G connections.
I believe that could be improved by adopting more on-demand strategy. Even with preloading, you need to load both of `vendor.js` and `application.js`. On the other hand, if we adopt on-demand strategy in the `home` layout with [code splitting feature of Webpack](https://webpack.js.org/guides/code-splitting-async/), we could reduce the amount of the transfer, as we could for the public layout in this pull request.

PR #2982, opened 2017-05-11
Add specs for current Status#permitted? implementation

jackjennings:
This backfills specs for the current implementation of the `Status#permitted?` method. No changes have been made to the implementation, and all logic paths should now be covered.

There's a lot of knowledge about how accounts work embedded in this code, and I would like to submit a pull request that extract this authorization. At this point that would be to large of a change for a PR, but I'd like to begin at least by finishing the code coverage for this section.

### DIFF
--- a/spec/models/status_spec.rb
+++ b/spec/models/status_spec.rb
@@ -120,7 +120,63 @@ RSpec.describe Status, type: :model do
   end
 
   describe '#permitted?' do
-    pending
+    it 'returns true when direct and account is viewer' do
+      subject.visibility = :direct
+      expect(subject.permitted?(subject.account)).to be true
+    end
+
+    it 'returns true when direct and viewer is mentioned' do
+      subject.visibility = :direct
+      subject.mentions = [Fabricate(:mention, account: alice)]
+
+      expect(subject.permitted?(alice)).to be true
+    end
+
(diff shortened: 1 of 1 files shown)


spec/models/status_spec.rb
```diff
@@ -120,7 +120,57 @@
...
 
   describe '#permitted?' do
-    pending
+    it 'returns true when direct and account is viewer' do
+      status = Fabricate(:status, visibility: :direct)
+      expect(status.permitted?(status.account)).to be true
+    end
+
+    it 'returns true when direct and viewer is mentioned' do
+      viewer = Fabricate(:account)
```
ykzts: Can you use [`alice`](https://github.com/tootsuite/mastodon/pull/2982/files#diff-c479d72b97713a99cab1561b1a8ee212L4)?

spec/models/status_spec.rb
```diff
@@ -120,7 +120,57 @@
...
+      status = Fabricate(:status, visibility: :direct, mentions: [Fabricate(:mention, account: viewer)])
```
ykzts: Using `subject` may be beautiful.

ykzts (changes requested):

jackjennings:
Updated per comments.

ykzts (approved):

PR #2985, opened 2017-05-11
Add dynamic polyfills for older browsers

### LINKED ISSUE #2941: Blank page with Javascript errors attempting to login - happens at 'getting started' url
Hi all. Forgive newbie/user post.

This bug just started happening. Mastodon was working perfectly for me as of a week or so ago.

Now I get a blank page on the getting_started screen on several (3) instances, including social.mastodon.com.au.

Web Developer addon console in Firefox shows 4 errors:

mutating the [[Prototype]] of an object will cause your code to run very slowly; instead create the object with the correct initial [[Prototype]] value using Object.create vendor-4e62c70d855e9bcf7b20.js:1:1870785

TypeError: value.keywords.includes is not a function vendor-4e62c70d855e9bcf7b20.js:1:1059968

HTTP "Content-Type" of "audio/mpeg" is not supported. Load of media resource https://social.mastodon.com.au/sounds/boop.mp3 failed. getting-started

TypeError: node is null

---

I am afraid I am a just a user and don't know what version they are running. I can ask them if you need.

I looked in the github issues here and there seemed to be some Closed issues with the same or similar result (not exactly the same) but this is currently happening to me on 3 different instances.
It may just be they all happen to be running the same pull.

If you want me to try on other sites for testing please specify which sites.

Using:
Firefox 41.0.1 on Linux. Addons disabled produces no change.

* * * *

- [* ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nolanlawson:
This adds the polyfills necessary to get Mastodon to work in (some) older browsers, including those browsers without Symbol, Object.assign, Array.prototype.includes, or Number.isNaN. Fixes #2941.

To avoid unnecessarily bloating `application.js`/`vendor.js`, I also implemented a dynamic polyfill import. So on newer browsers, the polyfills aren't even loaded, meaning those browsers don't pay the extra transfer, parse, execution, or memory costs.

Before:

```
492K	0-94ac7dada86b50e71f8e.js
576K	application-f64c16b425070625b6b4.js
8.0K	public-97ac551546a2c3ef4baa.js
868K	vendor-84ee95ea049554078ba2.js
```

After:

```
492K	0-619231f40c485acc9083.js
88K	1-a614545cd8905c0661bb.js
576K	application-90837bbe7f6684a0316b.js
8.0K	public-413a17ceb408901cf70e.js
800K	vendor-0d1c800d2177f0e22b47.js
```

Note that the `vendor.js` has shrunk by 68K and there is a new chunk that weights 88K, containing polyfills for Intl, Symbol, Object.assign, Array.prototype.includes, and Number.isNaN.

I tested in iOS 9, and this fix works:

screenshot 2017-05-10 23 42 22

I tried testing IE11 but honestly there is still some more work left to do to get IE11 to work. Some of the styles are messed up, and a Promise polyfill (which doesn't work with `import()` since it itself is Promise-based) is also needed. I'm not sure how much effort we really want to put into IE11 support.

### DIFF
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -1,0 +1,32 @@
+require('font-awesome/css/font-awesome.css');
+require('../styles/application.scss');
+
+function onDomContentLoaded(callback) {
+  if (document.readyState !== 'loading') {
+    callback();
+  } else {
+    document.addEventListener('DOMContentLoaded', callback);
+  }
+}
+
+function main() {
+  const Mastodon = require('mastodon/containers/mastodon').default;
+  const React = require('react');
+  const ReactDOM = require('react-dom');
+  const Rails = require('rails-ujs');
+  window.Perf = require('react-addons-perf');
+
+  Rails.start();
+
+  require.context('../images/', true);
+  require.context('../../assets/stylesheets/', false, /custom.*\.scss$/);
+
+  onDomContentLoaded(() => {
+    const mountNode = document.getElementById('mastodon');
+    const props = JSON.parse(mountNode.getAttribute('data-props'));
+
+    ReactDOM.render(<Mastodon {...props} />, mountNode);
+  });
+}
+
+export default main
--- a/app/javascript/mastodon/polyfills.js
+++ b/app/javascript/mastodon/polyfills.js
@@ -1,0 +1,18 @@
+import 'intl';
+import 'intl/locale-data/jsonp/en.js';
+import 'es6-symbol/implement';
+import includes from 'array-includes';
+import assign from 'object-assign';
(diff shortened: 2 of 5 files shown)


nolanlawson:
I also checked that none of Firefox, Edge, or Chrome are loading the extra polyfills file. So this should be a win for perf. :smiley:

ykzts (approved):
good PR!

stephenburgess8:
@nolanlawson This causes the entire app to not load in Safari.

`!window.Intl` is true, which means `window.Intl` is false in Safari Version 9.1.2 (11601.7.7) (not the latest)

Issue #2922, opened 2017-05-08
Regression : some colors aren't customizable anymore


Naouak:
Introduced by #2912

Some colors are not customizable anymore.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

mjankowski:
Can you elaborate more on what you were doing before and what doesn't work now?

Naouak:
https://github.com/tootsuite/mastodon/blob/master/app/javascript/styles/variables.scss

````
$base-shadow-color: $black;
$base-overlay-background: $black;
$base-border-color: $white;
$simple-background-color: $white;
$primary-text-color: $white;
$valid-value-color: $success-green;
$error-value-color: $error-red;
````
None of those can be changed without modifying variables.scss when it was possible with custom.scss before.

**Comment by yookoala** [comment]
Other than "custom*.scss", I think we should specifically have a `variables.scss` in `app/assets/stylesheets` to override variables. Here is the reason:

1. CSS rules comes later is always more powerful than sooner ones. Therefore `custom*.scss` should be imported / required after `application.scss`.
2. SCSS variables are inherits from **declarations before uses**. That means if you want to override a variable in `application.scss`, you should override before importing it.
3. You can use SASS / SCSS [`!default`](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#variable_defaults_) syntax to have the default values declared after the overridden values.

So if I'm not mistaken, you cannot override scss variables in files after `applications.scss`. And you cannot override the result css rules in files before `applications.scss`.

So practically, the code in [pack/application.js](https://github.com/tootsuite/mastodon/blob/af6a84da147f4230807b37c64bc09760e6ab5055/app/javascript/packs/application.js#L6,L18) should in look something like this:
```javascript
require.context('../../assets/stylesheets/', false, /^variables.*\.scss/); // variable overrided here
import '../styles/application.scss';

//...
//...

require.context('../../assets/stylesheets/', false, /^custom.*\.scss$/);
```

And [javascript/styles/variables.scss](https://github.com/tootsuite/mastodon/blob/af6a84da147f4230807b37c64bc09760e6ab5055/app/javascript/styles/variables.scss#L15,L22) should looks like this:
```scss
// Variables for defaults in UI
$base-shadow-color: $black !default;
[...]

PR #2987, opened 2017-05-11
[WIP] Allow SCSS variables to be overridden

yookoala:
* Allow SCSS variables to be overridden with
  `app/javascript/pack/variables*.scss`

* Allow these SCSS variables to be overriden:
  * `$base-shadow-color`
  * `$base-overlay-background`
  * `$base-border-color`
  * `$simple-background-color`
  * `$primary-text-color`
  * `$valid-value-color
  * `$error-value-color`

An approach to solve #2922.

### DIFF
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -1,3 +1,7 @@
+// allow override variables here
+require.context('../../assets/stylesheets/', false, /variables.*\.scss$/);
+
+// import default stylesheet with variables
 require('font-awesome/css/font-awesome.css');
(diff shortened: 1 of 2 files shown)


mjankowski (approved):

esetomo:
I could not override it.

app/javascript/pack/variables.scss is not used.
app/javascript/packs/variables.scss is packed to variables.css. But, It is not loaded.
app/assets/stylesheets/variables.scss is packed in application.css. But, Overrides are not active. Because 
each require() are converted to CSS separately.

PR #2993, opened 2017-05-11
Fix #1426 - Trim long usernames in public follower/following lists

### LINKED ISSUE #1426: Trim long usernames (visually)

### LINKED ISSUE #2221: OpenSSL - SSLError : certificate verify failed
Hello,

Since the 18 April, around 19h30 (07:30 pm) my instance isn't federating. I tried to restart mastodon, apache, even the server, and nothing work.
At first, I thought it was a proxy error, because I had error in my apache errors file about proxy. Now it seems to work, there are no more errors, but it's still not federating.

On sidekiq, there are a lot of failed and "retries" jobs. It all says :
`OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=error: certificate verify failed`

My `openssl version` is `OpenSSL 1.1.0f-dev  xx XXX xxxx`.

I renewed my certificates, but this doesn't solve the problem...

I have updated to v1.2 yesterday, and now I'm on v1.2.2.

I'm lost, what could I do to solve this ?

* * * *

Gargron:
Fix #2221 - Catch OpenSSL exceptions when loading remote avatars/headers/attachments
Don't strip "rel" attribute from `<a>` tags when sanitizing (microformats)

### DIFF
--- a/app/models/concerns/account_avatar.rb
+++ b/app/models/concerns/account_avatar.rb
@@ -34,7 +34,7 @@ module AccountAvatar
 
       self.avatar              = URI.parse(parsed_url.to_s)
       self[:avatar_remote_url] = url
-    rescue OpenURI::HTTPError => e
+    rescue OpenURI::HTTPError, OpenSSL::SSL::SSLError, Paperclip::Errors::NotIdentifiedByImageMagickError => e
       Rails.logger.debug "Error fetching remote avatar: #{e}"
     end
   end
--- a/app/javascript/styles/accounts.scss
+++ b/app/javascript/styles/accounts.scss
@@ -302,6 +302,8 @@
         display: block;
         color: $ui-base-color;
         text-decoration: none;
+        text-overflow: ellipsis;
+        overflow: hidden;
 
         &:hover {
           .display_name {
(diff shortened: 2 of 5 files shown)


app/models/concerns/account_avatar.rb
```diff
@@ -34,7 +34,7 @@ def avatar_remote_url=(url)
 
       self.avatar              = URI.parse(parsed_url.to_s)
       self[:avatar_remote_url] = url
-    rescue OpenURI::HTTPError => e
+    rescue OpenURI::HTTPError, OpenSSL::SSL::SSLError, Paperclip::Errors::NotIdentifiedByImageMagickError => e
```
mjankowski: seems unrelated?

Gargron: This PR fixes two issues.

mjankowski: I'd put those into separate commits.

I'll approve this PR, can you can decide to split it up or not.

mjankowski (approved):

PR #2996, opened 2017-05-11
Fix set cache expires on InstancePresenter

ryosuked:
If you do not specify an expiration date and use the Rails.cache, the value is permanently preserved. Specify an expiration date, /about/more is updated.


Gargron:
This is generally set in rails cache setup but might've been broken in #2732 ?

rsk837033:
I withdraw once. Because it may be a problem of my own environment.

PR #2998, opened 2017-05-12
Use babel-preset-env on Streaming Server

ykzts:
Change the streaming server to use babel-preset-env as well as asset compilation. This shortens the load time at first boot.


nolanlawson:
The lodash-plugin won't buy us much since its main purpose is to reduce the overall bundle size, but that doesn't make a big difference, so overall this is a great PR. :+1:

Also worth noting: we should really move from `babel-node` to just running `babel` then `node`, since `babel-node` is not meant for production: https://babeljs.io/docs/usage/cli/#babel-node

nolanlawson (approved):

ykzts:
Ah, certainly the Streaming Server does not use lodash. Babel-plugin-lodash is unnecessary.

Gargron (approved):

ykzts:
Remove `babel-plugin-lodash` in `./streaming/.babelrc`.

PR #3002, opened 2017-05-12
Rebuild node-sass

zunda:
Got an error in asset precompile on Heroku deploying second time:

```
Module build failed: Error: ENOENT: no such file or directory, scandir '/tmp/.../node_modules/node-sass/vendor'
```

This could be avoided with rebuilding the module in the `postinstall` hook: https://github.com/rails/webpacker#troubleshooting . `node-sass` module seems to have some files missing when it is built in a directory for the second time:


ykzts (approved):

S-YOU:
I don't think everyone have to rebuild that huge node-sass every time.

Gargron (approved):

PR #3007, opened 2017-05-12
Remove unnecessary message in video player

ykzts:
Remove unnecessary messages added in #1879. It is duplicated with other keys, causing the correct message not to be displayed.

### DIFF
--- a/app/javascript/mastodon/components/video_player.js
+++ b/app/javascript/mastodon/components/video_player.js
@@ -9,7 +9,6 @@ const messages = defineMessages({
   toggle_sound: { id: 'video_player.toggle_sound', defaultMessage: 'Toggle sound' },
   toggle_visible: { id: 'video_player.toggle_visible', defaultMessage: 'Toggle visibility' },
   expand_video: { id: 'video_player.expand', defaultMessage: 'Expand video' },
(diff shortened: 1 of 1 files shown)


mjankowski (approved):

app/javascript/mastodon/components/video_player.js
```diff
@@ -9,7 +9,6 @@ const messages = defineMessages({
   toggle_sound: { id: 'video_player.toggle_sound', defaultMessage: 'Toggle sound' },
   toggle_visible: { id: 'video_player.toggle_visible', defaultMessage: 'Toggle visibility' },
   expand_video: { id: 'video_player.expand', defaultMessage: 'Expand video' },
-  expand_video: { id: 'video_player.video_error', defaultMessage: 'Video could not be played' }
```
Gargron: This is clearly a wrong key, but I think the video error message is displayed somewhere?

ykzts: It is use at [line 166](https://github.com/tootsuite/mastodon/blob/2a8e034472d8cc761370b9be499138f035114b25/app/javascript/mastodon/components/video_player.js#L166), but this does not use defineMessages.

PR #3015, opened 2017-05-12
Increase color contrast on privacy warning

yiskah:
The current text contrast on the privacy warning is a WCAG violation. I didn't notice this because my instance has a custom theme which is better. On default theme I am barely able to read the text with my impaired vision. This patch brings the contrast to Normal Text WCAG AA compliance, and Large Text WCAG AAA compliance.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -226,7 +226,7 @@
 }
 
 .compose-form__warning {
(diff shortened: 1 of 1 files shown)


yiskah:
I can't assign ppl anymore but @beatrix-bitrot :V review plz?

Gargron (approved):

PR #3016, opened 2017-05-12
Add conversation model, <ostatus:conversation />

Gargron:
Cleaner/more focused rebase of #2864

Resolves #2864

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -21,6 +21,7 @@
 #  favourites_count       :integer          default(0), not null
 #  reblogs_count          :integer          default(0), not null
 #  language               :string           default("en"), not null
+#  conversation_id        :integer
 #
 
 class Status < ApplicationRecord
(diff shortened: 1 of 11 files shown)


app/models/status.rb
```diff
@@ -141,6 +143,11 @@ def non_sensitive_with_media?
     !sensitive? && media_attachments.any?
   end
 
+  before_validation :prepare_contents
+  before_create     :set_reblog
+  before_create     :set_visibility
+  before_create     :set_conversation
```
nightpool: a more rails-like way to do this is to have this hook as either `if: :local?` or `unless: :conversation_id?` instead of checking it multiple times in the method.

db/schema.rb
```diff
@@ -62,6 +62,19 @@
     t.index ["account_id", "target_account_id"], name: "index_blocks_on_account_id_and_target_account_id", unique: true, using: :btree
   end
 
+  create_table "conversation_mutes", force: :cascade do |t|
```
mjankowski: I think this snuck into this PR from other schema changes?

db/migrate/20170506235850_create_conversations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateConversations < ActiveRecord::Migration[5.0]
+  def change
+    create_table :conversations, id: :bigserial do |t|
```
mjankowski: I know this came up on other PR, but - do we care at all about backfilling data? Or is the general approach here that we can just not care about it, and only things which exist after this gets deployed will have full conversation support?

mjankowski (approved):
I had a few questions, but this seems ok.

app/models/status.rb
Gargron: Thread stuff in there is independent of local/remote

db/migrate/20170506235850_create_conversations.rb
Gargron: Yes

PR #3017, opened 2017-05-12
Feature conversations muting

### LINKED ISSUE #59: Mute replies to specific status from landing in mentions

### LINKED ISSUE #500: self-untag

Gargron:
Fix #500 
Fix #59 

Adds "muted" boolean property to status JSON.

- `POST /api/v1/statuses/:id/mute`
- `POST /api/v1/statuses/:id/unmute`

In the web UI: Displays "Disable notifications" vs "Re-enable notifications" for statuses in the notifications column.

Depends on #3016


app/javascript/mastodon/locales/ar.json
```diff
@@ -142,6 +142,7 @@
   "status.load_more": "حمّل المزيد",
   "status.media_hidden": "الصورة مستترة",
   "status.mention": "أذكُر @{name}",
+  "status.mute_conversation": "Disable notifications",
```
mjankowski: Will all these `locales/*.json` (other than `en`) not just fallback to `en` if they don't exist?

app/models/account.rb
```diff
@@ -130,6 +130,10 @@ def mute!(other_account)
     mute_relationships.find_or_create_by!(target_account: other_account)
   end
 
+  def mute_conversation!(conversation)
+    ConversationMute.find_or_create_by!(account: self, conversation: conversation)
```
mjankowski: Since this is inside `Account`, could this just be `conversation_mutes.where(conversation: conversation).first_or_create!` (or similar) to avoid the use of `self` here?

app/models/conversation.rb
```diff
@@ -10,7 +10,7 @@
 #
 
 class Conversation < ApplicationRecord
-  validates :uri, uniqueness: true
+  validates :uri, uniqueness: true, unless: 'uri.nil?'
```
mjankowski: How about `if: :uri?` to avoid eval'ing a string?

app/services/notify_service.rb
```diff
@@ -43,10 +43,19 @@ def blocked?
...
     blocked ||= (@recipient.user.settings.interactions['must_be_follower']  && !@notification.from_account.following?(@recipient)) # Options
     blocked ||= (@recipient.user.settings.interactions['must_be_following'] && !@recipient.following?(@notification.from_account)) # Options
+    blocked ||= conversation_muted?
     blocked ||= send("blocked_#{@notification.type}?")                                                                             # Type-dependent filters
     blocked
   end
 
+  def conversation_muted?
+    if @notification.target_status
+      !@notification.target_status.conversation_id.nil? && ConversationMute.where(account: @recipient, conversation_id: @notification.target_status.conversation_id).exists?
```
mjankowski: Might put this into something like `@recipient.muting_conversation?(@notification.target_status.conversation)`, and have it handle nils itself...

spec/models/conversation_mute_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe ConversationMute, type: :model do
```
mjankowski: Either put something in here, or delete the file.

app/services/notify_service.rb
```diff
@@ -43,10 +43,19 @@ def blocked?
     blocked ||= (@notification.from_account.silenced? && !@recipient.following?(@notification.from_account))                       # Hellban
     blocked ||= (@recipient.user.settings.interactions['must_be_follower']  && !@notification.from_account.following?(@recipient)) # Options
     blocked ||= (@recipient.user.settings.interactions['must_be_following'] && !@recipient.following?(@notification.from_account)) # Options
+    blocked ||= conversation_muted?
```
mjankowski: spec coverage?

app/models/status.rb
```diff
@@ -181,6 +181,10 @@ def reblogs_map(status_ids, account_id)
       select('reblog_of_id').where(reblog_of_id: status_ids).where(account_id: account_id).map { |s| [s.reblog_of_id, true] }.to_h
     end
 
+    def mutes_map(conversation_ids, account_id)
```
mjankowski: spec coverage?

app/javascript/mastodon/locales/ar.json
Gargron: Autogenerated by `yarn run manage:translations`

nightpool:
To reiterate my comments from #2864:

Currently, muting fails silently for statuses created before whenever this code is released, with no indication at all that this is the case. This is incredibly hostile to users, and will basically make it look like we're shipping a broken feature.

An asynchronous one-time sidekiq backfill that populated conversation_ids would solve this problem, although we would have to figure out how it gets run. Failing that, adding a "populate_conversation_id_if_not_present" call to mute and unmute would probably be the second best choice.

There's some smaller UI/UX polish that's probably needed here as well. If I have a chain of self-replies A -> B -> C, and I click "mute notifications" on C, I wouldn't expect that to mute notifications for A and B, which is what it does. I think changing it to "Mute conversation" or "Mute notifications for conversation" would fix this.

Gargron:
@nightpool It's actually impossible to backfill this retroactively. I suggest that it shall fail with a 422 for statuses that don't have a conversation, how about that?

nightpool:
Why would it be impossible to backfill for local statuses?

Even if it was impossible to backfill, that technical detail shouldn't
compromise the user experience—we could fall back to blacklisting on "in
reply to" or something similar

Gargron:
I'm sorry but I will not do backfilling. First of all, it's not worth it, most conversations end within one day. Secondly, I will not be combing through 6M records just on mastodon.social alone to fill in this data. Microblogging is fast paced and ephermal, these old records will become almost completely irrelevant in just a few days.

Gargron:
@mjankowski All fixed, I think

ykzts (approved):

PR #3018, opened 2017-05-12
Fix color contrast some more in privacy warning

yiskah:
Latest master appears to have changed the `<strong>` to this unreadable grey color. 

capture

So this PR makes it look like this:

capture 2

If you want that to be white then it should be pure white. If someone would rather revert this to that strong dark blue color it was before that would be good too.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -236,7 +236,7 @@
   font-weight: 400;
 
   strong {
-    color: darken($primary-text-color, 33%);
+    color: $primary-text-color;
     font-weight: 500;
   }
 
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -236,7 +236,7 @@
   font-weight: 400;
 
   strong {
-    color: darken($ui-secondary-color, 65%)
+    color: darken($ui-secondary-color, 65%);
     font-weight: 500;
   }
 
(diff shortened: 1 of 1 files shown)


Gargron:
It was never meant to be white, I am not sure what change did this

Gargron:
Just make it `darken($ui-secondary-color, 65%)`

yiskah:
It's done.

Wonderfall:
I was about to make the same PR, glad I found this one before. 
I use `darken($ui-highlight-color, 8%)` on my instance and it renders very fine.

Gargron (approved):

app/javascript/styles/components.scss
```diff
@@ -236,7 +236,7 @@
   font-weight: 400;
 
   strong {
-    color: darken($primary-text-color, 33%);
+    color: darken($ui-secondary-color, 65%)
```
Gargron: Forgotten `;`

Gargron (changes requested):

app/javascript/styles/components.scss
yiskah: woops! that's what i get for copy-pasting

yiskah:
OK! +;

Wonderfall:
At least can you provide a screen on how it renders now?

This is how my proposition renders (but it's custom CSS) :

Gargron (approved):

PR #3020, opened 2017-05-12
Revert HTML CW changes

beatrix-bitrot:
set spoiler_text back to just be text, the way it was before. clients will have to link hashtags themselves if they want!

partly reverts changes from #2970 
completely reverts changes from #2683

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -36,6 +36,8 @@ class StatusContent extends React.PureComponent {
         link.setAttribute('title', mention.get('acct'));
       } else if (link.textContent[0] === '#' || (link.previousSibling && link.previousSibling.textContent && link.previousSibling.textContent[link.previousSibling.textContent.length - 1] === '#')) {
         link.addEventListener('click', this.onHashtagClick.bind(this, link.text), false);
+      } else if (media) {
+        link.innerHTML = '<i class="fa fa-fw fa-photo"></i>';
       } else {
         link.setAttribute('target', '_blank');
         link.setAttribute('rel', 'noopener');
@@ -68,11 +70,11 @@ class StatusContent extends React.PureComponent {
     const [ startX, startY ] = this.startXY;
     const [ deltaX, deltaY ] = [Math.abs(e.clientX - startX), Math.abs(e.clientY - startY)];
 
-    if (e.target.localName === 'button' || e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
+    if (e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
       return;
     }
 
-    if (deltaX + deltaY < 5 && e.button === 0 && this.props.onClick) {
+    if (deltaX + deltaY < 5 && e.button === 0) {
       this.props.onClick();
     }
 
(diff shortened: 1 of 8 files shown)


app/javascript/mastodon/components/status_content.js
```diff
@@ -68,11 +70,11 @@ class StatusContent extends React.PureComponent {
     const [ startX, startY ] = this.startXY;
     const [ deltaX, deltaY ] = [Math.abs(e.clientX - startX), Math.abs(e.clientY - startY)];
 
-    if (e.target.localName === 'button' || e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
+    if (e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
```
Gargron: We wanna keep this and 77

app/javascript/mastodon/components/status_content.js
```diff
@@ -36,6 +36,8 @@ class StatusContent extends React.PureComponent {
         link.setAttribute('title', mention.get('acct'));
       } else if (link.textContent[0] === '#' || (link.previousSibling && link.previousSibling.textContent && link.previousSibling.textContent[link.previousSibling.textContent.length - 1] === '#')) {
         link.addEventListener('click', this.onHashtagClick.bind(this, link.text), false);
+      } else if (media) {
```
Gargron: This should stay gone

app/javascript/mastodon/components/status_content.js
```diff
@@ -116,19 +118,14 @@ class StatusContent extends React.PureComponent {
       }
 
       return (
-        <div
-          ref={this.setRef}
-          className='status__content'
-          onMouseDown={this.handleMouseDown}
-          onMouseUp={this.handleMouseUp}
-        >
+        <div className='status__content' onMouseDown={this.handleMouseDown} onMouseUp={this.handleMouseUp}>
```
Gargron: All of below can stay as they were

app/lib/atom_serializer.rb
```diff
@@ -336,8 +336,8 @@ def add_namespaces(parent)
   end
 
   def serialize_status_attributes(entry, status)
-    append_element(entry, 'summary', Formatter.instance.format(status.proper, :spoiler_text, false).to_str, 'xml:lang': status.language, type: 'html') if status.spoiler_text?
     append_element(entry, 'content', Formatter.instance.format(status).to_str, type: 'html', 'xml:lang': status.language)
+    append_element(entry, 'summary', status.spoiler_text, 'xml:lang': status.language) if status.spoiler_text?
```
Gargron: Doesn't really matter but I liked summary be before content. Otherwise correct

Gargron (changes requested):

app/javascript/mastodon/components/status_content.js
beatrix-bitrot: are you sure? cause we introduced a regression where clicking the CW button would also open the detailed status and after this changeset it is fixed, which I suspect is related to this  particular change though i didn't test it in isolation

Gargron: I couldn't reproduce that behaviour, it shouldn't depend on button vs a anyway, but button is more a11y-friendly

beatrix-bitrot: agreed ok, will do so

Gargron (approved):

PR #3041, opened 2017-05-13
Catch import promise in application.js

stephenburgess8:
Hi -  #2985 changed how polyfills were being imported. This did not work for me in Safari 9 until I made the following changes:

- ~Call `main()` after the import completes.~
- ~Since none of the code in the import was "run", import the intl polyfill.~
- Catch promise error to help determine my error: need to run `yarn update`
- Make import promise catch a noop with commented out console.log for a debugging hint.

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,7 +4,7 @@ if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
   import('../mastodon/polyfills').then(main).catch(e => {
-    console.log(e)
+    // console.log(e); // no-op
   });
 } else {
   main();
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,7 +4,7 @@ if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
   import('../mastodon/polyfills').then(main).catch(e => {
-    // console.log(e); // no-op
+    console.log(e); // eslint-disable-line no-console
   });
 } else {
   main();
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -4,7 +4,7 @@ if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
   import('../mastodon/polyfills').then(main).catch(e => {
-    console.log(e); // eslint-disable-line no-console
+    console.error(e) // eslint-disable-line no-console
   });
 } else {
   main();
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -3,7 +3,9 @@ import main from '../mastodon/main';
 if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
-  import('../mastodon/polyfills').then(main);
+  import('../mastodon/polyfills').then(main).catch(e => {
+    console.log(e)
+  });
 } else {
   main();
 }


nolanlawson:
Hi, I saw your comment in https://github.com/tootsuite/mastodon/pull/2985#issuecomment-301258378. What do you mean "it doesn't work"? I tested in iOS Safari 9, and it worked just fine…

Also with this PR we would download the entire Intl polyfill for all browsers, which essentially negates the benefit of dynamic loading.

app/javascript/packs/application.js
```diff
@@ -1,9 +1,11 @@
 import main from '../mastodon/main';
 
 if (!window.Intl || !Object.assign || !Number.isNaN ||
-    !window.Symbol || !Array.prototype.includes) {
+  !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
-  import('../mastodon/polyfills').then(main);
+  require('intl');
+  require('intl/locale-data/jsonp/en.js');
+  import('../mastodon/polyfills').then(main());
```
nolanlawson: This is not doing what you might think it's doing. This will execute the `main()` function immediately rather than waiting for the promise to resolve (I wrote [a blog post about this](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html) in case it helps :smile:).

stephenburgess8: I'm confused, because without explicitly calling `main()`, main _never gets called _ in Safari 9. `import()` is a cool function, is there something about Safari 9 which does not support it? Mastodon was loading in Safari 9 before this change. I will read through the blog post again, thanks.

nolanlawson:
What do you see in the Safari console? And yes, the blog post explains
what's going on. :) The main() function likely doesn't get called because
the promise rejected with an error. You can add a .catch() if Safari's
console doesn't show the error.

-----
Nolan Lawson
http://nolanlawson.com

stephenburgess8:
Catching the error and throwing it produces this:
```
Error: Cannot find module "is-nan"
webpackMissingModule
eval code
eval
(anonymous function) — 1-d714deb3e19990d029af.js:14
__webpack_require__ — vendor.js:55
__webpack_require__
promiseReactionJob
```

stephenburgess8:
How is this for catching the error promise?

nolanlawson:
I'm AFK right now but that looks like a Webpack/yarn error. Did you run
yarn install?

-----
Nolan Lawson
http://nolanlawson.com

stephenburgess8:
The article says to catch errors from promises but this promise didn't have that. I changed this pull request to now be just catching the error from the promise and throwing it. Does this need anything more than just a console log?

nolanlawson:
lol yeah I didn't follow my own advice. :) Yes, adding a .catch() with a
console.log would be helpful for debugging.

-----
Nolan Lawson
http://nolanlawson.com

app/javascript/packs/application.js
```diff
@@ -3,7 +3,9 @@ import main from '../mastodon/main';
 if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
-  import('../mastodon/polyfills').then(main);
+  import('../mastodon/polyfills').then(main).catch(e => {
+    // console.log(e); // no-op
+  });
```
nolanlawson: Doing a no-op just means it will swallow the error...

stephenburgess8: However console logs are not allowed in Mastodon - it fails the build. I can't think of another way to get around that and still have this catch be useful. I think it's sort of a hint for debugging. Can you think of another way to get around the fact that Mastodon does not allow console logging?

stephenburgess8: In other words, if this had been there when I was debugging, it would have helped me out.

nolanlawson:
You can probably put in an eslint comment to disable it

-----
Nolan Lawson
http://nolanlawson.com

nightpool:
I'm not sure I see the value of this PR.

stephenburgess8:
It is small, but promises such as this should be caught as a matter of JavaScript best practices.

> Don't forget to .catch
> Remember, we want to take advantage of Promise's single flow of data and exceptions; and, ensure we don't let any errors get swallowed or dropped on the floor.

http://www.datchley.name/promise-patterns-anti-patterns/

nolanlawson:
The value is that not every browser shows rejected promises in the console,
so it can be very difficult to debug.

-----
Nolan Lawson
http://nolanlawson.com

nolanlawson (approved):

nightpool:
@nolanlawson that sounds like a problem that should be solved systemically in whatever promise library we're using?

nolanlawson:
We use native promises because it's better performance-wise, especially since some polyfills are quite large. E.g. I know bluebird logs rejected promises but it's also huge.

nightpool:
Regardless, that's not an argument to special case this one invocation of a promise. If we want to log rejected promises, we should do that everywhere, not one at a time.

nolanlawson:
If there is a special case for this one scenario it's that if the promise rejects then the main() function doesn't execute, meaning that Mastodon itself doesn't even load. I think it might be the case that, once Mastodon loads, promise rejections are shown as UI toasts but I would need to check.

ykzts (approved):

nightpool:
Okay but we should at LEAST be using console.error

app/javascript/packs/application.js
```diff
@@ -3,7 +3,9 @@ import main from '../mastodon/main';
 if (!window.Intl || !Object.assign || !Number.isNaN ||
     !window.Symbol || !Array.prototype.includes) {
   // load polyfills dynamically
-  import('../mastodon/polyfills').then(main);
+  import('../mastodon/polyfills').then(main).catch(e => {
+    console.log(e); // eslint-disable-line no-console
```
ykzts: use `console.error`

ykzts (changes requested):

stephenburgess8:
Thanks for your help on this. I looked into it further. Alerts are used via middleware to send alerts for errors like this. However this promise statement occurs before redux comes into the picture and before the middleware is applied.

ykzts (approved):

PR #3042, opened 2017-05-13
Add reserved usernames

hika-p:
I suggest reserving these words "system, official, owner, master, register, staff, support, admin, etc.".
At least, I added them as reserved usernames in developing another service.
Please let me know your opinions on this.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -46,6 +46,9 @@ class Account < ApplicationRecord
   include Attachmentable
   include Targetable
 
+  word_data = ReservedUsernameList.all
+  VALID_RESERVED_WORDS = word_data.map{ |word| word.word }
+
   # Local users
   has_one :user, inverse_of: :account
 
@@ -54,7 +57,7 @@ class Account < ApplicationRecord
 
   # Local user validations
   with_options if: 'local?' do
-    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }
+    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, exclusion: {in: VALID_RESERVED_WORDS}
     validates :display_name, length: { maximum: 30 }
     validates :note, length: { maximum: 160 }
   end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,15 @@
+require 'json'
+
 Doorkeeper::Application.create!(name: 'Web', superapp: true, redirect_uri: Doorkeeper.configuration.native_redirect_uri, scopes: 'read write follow')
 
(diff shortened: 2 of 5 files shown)


app/models/account.rb
```diff
@@ -54,7 +57,7 @@ class Account < ApplicationRecord
 
   # Local user validations
   with_options if: 'local?' do
-    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }
+    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, exclusion: {in: VALID_RESERVED_WORDS}
```
nightpool: This will blow up for users that already have accounts with these names, like support@mastodon.social, right? Ideally this would *only* be a check that affects the creation of users.

Also, I don't think this prevents someone from creating "Admin" or "HELP".

db/seeds.rb
```diff
@@ -1,7 +1,15 @@
...
 if Rails.env.development?
   domain = ENV['LOCAL_DOMAIN'] || Rails.configuration.x.local_domain
   admin  = Account.where(username: 'admin').first_or_create!(username: 'admin')
   User.where(email: "admin@#{domain}").first_or_initialize(email: "admin@#{domain}", password: 'mastodonadmin', password_confirmation: 'mastodonadmin', confirmed_at: Time.now.utc, admin: true, account: admin).save!
 end
+
+reserved_words = JSON.parse(File.read('db/reserved_username.json'))
+
+reserved_words.each do |word|
+  ReservedUsernameList.create(word: word)
```
nightpool: Putting this in seeds is confusing, because it won't update the database if words are added or removed from the json file. I think this probably shouldn't be in the database—instead, it should probably be loaded from the filesystem when the app loads.

nightpool (changes requested):

app/models/account.rb
hikaru-light: Thank you for your prompt review.

Yes, I guessed that this validation can prohibit users from creating accounts with special name.
It should be called before creation and response other name forms. Please let me clarify my understanding.

db/seeds.rb
hikaru-light: Thank you for your nice advice.

I try another way that I refer to your opinion. I want to implement this somehow.
Please give me some time.

mal0ki:
How is the progress of this going? 
I'm I think I'll add it to the 1.5 roadmap, if that is okay?

hikaru-light:
@mal0ki Thank you for your concern. I tried to use a gem to implement it because I received an advice.
If you have to add this soon, please do it.

hikaru-light:
Sorry, I have a grammar mistake.
I finished to revise it.

Gargron:
#3566 implemented this, closing (but thank you for putting the effort in!)

PR #3045, opened 2017-05-13
tiny js fix for 'show more' button in chrome

beatrix-bitrot:
the button would cause the toot to open as well as showing the hidden text which was incredibly annoying

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -68,7 +68,7 @@ class StatusContent extends React.PureComponent {
     const [ startX, startY ] = this.startXY;
     const [ deltaX, deltaY ] = [Math.abs(e.clientX - startX), Math.abs(e.clientY - startY)];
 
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/status_content.js
```diff
@@ -68,7 +68,7 @@ class StatusContent extends React.PureComponent {
     const [ startX, startY ] = this.startXY;
     const [ deltaX, deltaY ] = [Math.abs(e.clientX - startX), Math.abs(e.clientY - startY)];
 
-    if (e.target.localName === 'button' || e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
+    if (e.target.localName === 'button' || e.target.localName === 'span' || e.target.localName === 'a' || (e.target.parentNode && e.target.parentNode.localName === 'a')) {
```
Gargron: Why does this fix it though?

beatrix-bitrot: because in chrome e.target is the span element and in firefox (and maybe other browsers?) e.target is the button

why that is the case? i have no clue

Gargron (approved):

Issue #2588, opened 2017-04-28
Setting own profile to RTL scrambles others' non-RTL posts view


mabkenar:
This is very related to #2350 but slightly different: it concerns direction settings for the **skeleton** of the page. This issue is introduced in release 1.3.1

### What I see
If I set my profile setting to Persian (an RTL language), then this is how I see a normal [LTR post](https://social.tchncs.de/@jaltek/1409432):

![selection_182](https://cloud.githubusercontent.com/assets/3006332/25534773/3564a9e0-2c35-11e7-8c34-365450bb0767.png)
[looking messy]

### What others see
Another user with profile set to English/German/etc (or me with a Private browser window that does not know about my profile preferences) sees the page like this:

![selection_183](https://cloud.githubusercontent.com/assets/3006332/25534764/2f68276a-2c35-11e7-94c1-d519a9043911.png)
[looking fine]

Problems with the first view:

- user avatar of the main post is shown at left, while username is at right.
- user avatar for the reply post is shown at right and is blocking the time stamp ("an hour ago") of the reply post

### What should happen
Display of posts should not depend of my profile preferences. It should depend on the **post** language. For RTL posts, the avatar and username should appear at right, and the time stamp should appear at left. Moreover, margins should be set accordingly so that the avatar does not cover any text.

As long as this issue exists ~~the human race will never be free~~ setting the profile to an RTL language will breaks **rendering of all non-RTL posts**.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

mabkenar:
I would really like to solve (or help to solve) this issue because this is breaking my usage of Mastodon. Based on a [recent commit](https://github.com/tootsuite/mastodon/pull/3682/files) by @Gargron, I see that these two files are related to this bug:

https://github.com/tootsuite/mastodon/blob/master/app/views/stream_entries/_detailed_status.html.haml
https://github.com/tootsuite/mastodon/blob/master/app/views/stream_entries/_simple_status.html.haml

And this is what each of the two files render:
![3564a9e0-2c35-11e7-8c34-365450bb0767](https://user-images.githubusercontent.com/3006332/27168302-5e4a6b0c-51a5-11e7-939e-529086a88276.png)

[...]

PR #3047, opened 2017-05-13
l10n: Improve RTL detection (Stopgap until language detection improves)

### LINKED ISSUE #2350: RTL support for post contents in profile pages and individual posts view

selection_155

selection_157

* * * *

seefood:
Not too happy with the extra regex lines in JS (makes the compose box quite slower once the first RTL char is typed), but it decides directonality way better in the toot-deck UI

It seems this fixes #2350, however since even more toots are now classified as RTL than before, bug #2588 is happens even more often, which is unfortunate...

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -48,21 +48,24 @@ module StreamEntriesHelper
   end
 
   def rtl?(text)
-    rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
+    rtl_charcount = text.gsub(/[^\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]/m, '').strip.size.to_f
 
-    if rtl_characters.present?
-      total_size = text.strip.size.to_f
-      rtl_size(rtl_characters.to_a) / total_size > 0.3
+    if rtl_charcount.positive?
+      # Remove mentions before counting characters to decide RTL ratio
+      justtext = text.gsub(Account::MENTION_RE, '')
+      # Remove Email addresses
+      justtext = justtext.gsub(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '')
+      # Naiive catcher for URLs
+      justtext = justtext.gsub(/[-A-Za-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-A-Za-z0-9@:%_\+.~#?&\/=]*)/, '')
+
+      total_size = justtext.strip.size.to_f
+      rtl_charcount / total_size > 0.3
     else
       false
     end
   end
 
   private
-
-  def rtl_size(characters)
-    characters.reduce(0) { |acc, elem| acc + elem.size }.to_f
-  end
 
   def embedded_view?
     params[:controller] == EMBEDDED_CONTROLLER && params[:action] == EMBEDDED_ACTION
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -54,7 +54,7 @@ module StreamEntriesHelper
       # Remove mentions before counting characters to decide RTL ratio
       justtext = text.gsub(Account::MENTION_RE, '')
       # Remove Email addresses
-      justtext = justtext.gsub(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '')
+      # justtext = justtext.gsub(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '')
       # Naiive catcher for URLs
       justtext = justtext.gsub(/[-A-Za-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-A-Za-z0-9@:%_\+.~#?&\/=]*)/, '')
 
--- a/app/javascript/mastodon/rtl.js
+++ b/app/javascript/mastodon/rtl.js
@@ -26,9 +26,7 @@ export function isRtl(text) {
   // Remove mentions before counting characters to decide RTL ratio
   text = text.replace(/@[0-9A-Za-z_@]+/g, '');
   // Remove Email addresses
-  text = text.replace(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '');
-  // Naiive catcher for HTML tags
-  text = text.replace(/<[A-Za-z\/]+[^>]*>/g, '');
+  // text = text.replace(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '');
   // Naiive catcher for URLs
   text = text.replace(/[-A-Za-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-A-Za-z0-9@:%_\+.~#?&//=]*)/g, '');
 
--- a/app/javascript/mastodon/rtl.js
+++ b/app/javascript/mastodon/rtl.js
@@ -23,5 +23,14 @@ export function isRtl(text) {
     return false;
   }
 
+  // Remove mentions before counting characters to decide RTL ratio
+  text = text.replace(/@[0-9A-Za-z_@]+/g, '');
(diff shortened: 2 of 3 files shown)


seefood:
Reviews requested: @YAIR @mabkenar

app/helpers/stream_entries_helper.rb
```diff
@@ -48,22 +48,26 @@ def microformats_h_class(status, is_predecessor, is_successor, include_threads)
   end
 
   def rtl?(text)
-    rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
```
seefood: This was matching only the first word, not all the chars. Totally useless.

app/helpers/stream_entries_helper.rb
```diff
@@ -48,22 +48,26 @@ def microformats_h_class(status, is_predecessor, is_successor, include_threads)
...
+      total_size = justtext.strip.size.to_f
+      rtl_charcount / total_size > 0.3
     else
       false
     end
   end
 
   private
 
-  def rtl_size(characters)
```
seefood: no longer used in my code.

ira-lb:
@JoelQ This overrides original code by you, could you weigh in?

mabkenar:
Thank you for the updated PR. Unfortunately I cannot review the code since I have no prior experience with coding tootsuite and I don't have resources to make a dev environment to test.

seefood:
I just hope someone reviews this before v1.4 is in feature freeze...

seefood:
@Gargron what can I do to promote this PR? Thanks!

app/javascript/mastodon/rtl.js
```diff
@@ -23,5 +23,14 @@ export function isRtl(text) {
     return false;
   }
 
+  // Remove mentions before counting characters to decide RTL ratio
+  text = text.replace(/@[0-9A-Za-z_@]+/g, '');
```
Gargron: Instead of all of this it's probably better to just rely on what the language detected in the toot is? And then RTL if it's [arabic, farsi, hebrew, etc]

Then you wouldn't need to duplicate the logic in both ruby and JS

app/helpers/stream_entries_helper.rb
```diff
@@ -48,22 +48,25 @@ def microformats_h_class(status, is_predecessor, is_successor, include_threads)
...
-    if rtl_characters.present?
-      total_size = text.strip.size.to_f
-      rtl_size(rtl_characters.to_a) / total_size > 0.3
+    if rtl_charcount.positive?
+      # Remove mentions before counting characters to decide RTL ratio
+      justtext = text.gsub(Account::MENTION_RE, '')
+      # Remove Email addresses
+      justtext = justtext.gsub(/^([a-z0-9_\.-]+)@([\da-z\.-]+)\.([a-z\.]{2,6})$/, '')
+      # Naiive catcher for URLs
+      justtext = justtext.gsub(/[-A-Za-z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-A-Za-z0-9@:%_\+.~#?&\/=]*)/, '')
```
Gargron: Check https://github.com/tootsuite/mastodon/blob/master/app/lib/language_detector.rb#L32 for how URLs are stripped out when detecting language. Actually, the language detector should probably strip Account::MENTION_RE too. I don't think regexing e-mail addresses out is worth it, e-mail regex is ridiculously complex when done correctly.

Also, I think Ruby could also just rely on detected language instead of counting utf8 chars.

app/javascript/mastodon/rtl.js
Gargron: Let me correct myself: When displaying toots from the server, rely on language.

When composing a new toot, we still need char-based RTL detection. Stripping out mentions and URLs is good, e-mails not worth it (as I said above), HTML tags unnecessary because they will not be in the compose form.

Gargron (changes requested):

app/javascript/mastodon/rtl.js
seefood: Well, this detects it in RT while you type at the compose window. I don't know the language detection algo you mentioned, because These two are the RTL detection functions I have found. I only extended the existing code, didn't invent new design.

Also, I have no idea how the language detection work (I'll go read it) but right now my tests show it's far from perfect. I found it masked legit toots from the public timelne, I was testing only English and Hebrew, and marked those two languages in, but maybe half the toots were displayed, so it would be a step back from this algo.

app/helpers/stream_entries_helper.rb
seefood: I fear detection of the language may be more computationally expensive than three gsub() calls, but I don't intend to dig deeper into that 3rd party code. If it ran once on writing to the DB and written for future reads it would make some sense, but as it is, we have to detect directionality on the way out to the browser (or in the browser when talking JS), and I patched the code that does it today to do it better, that's all. maybe in the stretch till v1.5 I'll have time to test or improve the language detection mechanism.

Gargron: Language detection runs once, in any case, when toot is created. These helper methods are used on every render.

seefood: just to make a point, I added a language detection spec, here's what I got:

>   1) LanguageDetector to_iso_s Detects Hebrew language
>     Failure/Error: expect(result).to eq :he
>       expected: :he
>          got: :da
>     (compared using ==)

Yes, my Hebrew+mentions+hashtag is danish. Sorry, I can't rely on this mechainsm for directonality. if it gets fixed or improved, it can be retrofitted later for this task, but for now please merge this into 1.4.

seefood:
OK, here's the situation.
I have a working patch here that improves the problem, however @Gargron claims it's just enhancing the wrong solution. I'll try to break it down:

1. The JS code in `app/javascript/mastodon/rtl.js` is for real-time screen updates of the edit box and that's fine. I think there's no argument there.
2. The Ruby code in `app/helpers/stream_entries_helper.rb` is determining the directionality on the way to the client, instead of on the way into the database. I agree that's the wrong part of the chain to make that test. I'm all for calling it before adding it to the DB, and adding a directionality field in the table.
3. I'm not for deducing the directionality from the language because my tests show it's not working well. in fact it identifies Hebrew tweets and Danish and other languages, so that's kind of useless.

So far our options are:

1. Merge this PR as-is, into 1.4
2. Move it to the earlier part of the chain and add a directionality column to the DB.
3. Gargron's idea: get rid of the `rtl?` function and use the language (which is not working for me).
4. Leave this alone, I'll maintain this patch on the side for anyone who wants it.

I invite anyone with an interest in RTL support (@mabkenar @nightpool @eladhen @elad661 @ashfurrow @msappir @JoelQ, others?)  to weigh in on where to go next please.

seefood:
Oh yeah, Also option 5: I separate this to a JS patch and a Ruby patch and we argue them separately.

elad661:
I don't think that using the language detection is a good idea. It's easy to demonstrate that it simply doesn't work, even if you do add extra filtering for hashtags, mentions, and URLs. The solution that exists now is not "wrong", it is used by twitter, Google, and Facebook quite successfully. It's a tried and tested method that simply works. Ever since twitter implemented it I have not seen even a single complaint from Hebrew/Arabic speakers about the state of RTL support on twitter.

So I think that merging this PR is the correct idea.

Gargron:
Could you ignore my language detection comment and just do the other changes I requested? Like:

1. Strip out URLs the way language_detector.rb does it, for consistency
2. Don't bother stripping out e-mail addresses - you are unlikely to do it correctly with a simple regex and it's a rare enough case that it's unlikely to affect directionality of the text
3. In the JS version, also don't handle e-mails, and don't try to regex HTML, in fact you need to convert the text to plaintext before doing all the other regexes because HTML tags may be in the middle of strings you're looking for. See how status contents are converted to plaintext for desktop notifications in actions/notifications.js and for search_index in reducers/statuses.js. In fact if you ensure that the search_index is what's fed into the rtl function, you don't need to think about HTML at all

seefood:
2. easy, I'll just leave them remarked in case anyone (like my home instance) wants to re-enable them.
> you are unlikely to do it correctly with a simple regex
Well, I didn't write that one myself (though I have tested it and used it elsewhere in the past) and it's one of the more ubiquitous ones used on the web.

1+3, those are beyond my programming level, sorry. I don't feel confident enough to do this myself correctly. I can try it at the end of my work day (Yes, Israelis work on Sundays) but I prefer someone more familiar with Ruby and JS to fix these in this time frame.

PR #3052, opened 2017-05-14
Add ServiceWorker / AppCache with offline-plugin (work in progress)

nolanlawson:
Fixes https://github.com/tootsuite/mastodon/issues/2917.

What this adds:

1. ServiceWorker (Chrome/Firefox) and AppCache (Safari/Edge) support using `offline-plugin`
2. `offline-plugin` generates a SW and AppCache to precache all static Webpack assets
3. Custom ServiceWorker code handles requests that should be cached on-the-fly

How this should work:

1. In Chrome and Firefox, the app should "work" offline – i.e. refreshing the page while offline works, and resources like JS, CSS, and emoji only need to be fetched from the network once
2. Whenever `sw.js` is changed due to a `precompile`, the new SW will take over once the user navigates away and back
2. In Safari and Edge, the main HTML page will not work offline due to limitations (see below)
3. No data syncing or caching yet, so if you are fully offline then you won't see anything in your timeline

The main limitation I ran into is that the HTML pages at `/web` and `/settings` contain a CSRF token embedded as a meta tag, meaning that when the user signs out, we ought to clear that token. So my strategy is to listen for `/auth/sign_out` requests inside the ServiceWorker and then remove those HTML pages from the cache when the user signs out. Since AppCache doesn't support this behavior, those files just aren't cached at all for AppCache.

TODOs:

- [x] Test in Chrome, Edge, Firefox, and Safari on an actual instance (I may create a throwaway instance for this; it's extremely hard to test without a proper production setup)
- [x] Verify that the CSRF token continues to work correctly in a variety of cases (e.g. user manually refreshing, user signing out and back in, user closing the browser and then the token expires on the server and then they open the browser again)

TODOs that IMO can wait for a future PR:

- [ ] Add a mechanism so users can force an update when the `sw.js` changes (e.g. a toast saying "A new version is available, click here to reload")
- [ ] Cache timelines/notifications in IndexedDB
- [ ] Precache emoji without making multiple requests
- [ ] Add sourcemap to `sw.js`
- [ ] Add push notifications with FCM key defined in `.env`


nolanlawson:
This PR is now live at https://malfunctioning.technology (my new dev instance). So far Chrome, Safari, and Edge seem fine but Firefox Mobile is not working in offline mode. I also need to fix a bug that apparently will affect Chrome 59 (I'm getting a warning about it).

nolanlawson:
More issues to solve after testing:

- [ ] avoid caching 40x/50x errors
- [ ] fix whatever's not working in Firefox
- [ ] fix redirect issue that will break in Chrome 59

ramlmn:
* [ ] Use `cacheName` option of offline-plugin in webpack.production.js (to prevent future issues).
* [ ] Use semver or any other method to add 'version' in package.json so that it can be used in both sw-entry.js, and in webpack.production.js as [version](https://github.com/NekR/offline-plugin/blob/master/docs/options.md#version-string--plugin-offlineplugin--void) option instead of manually updating everytime.
* [ ] Exclude caching of `.gz` files in offline-plugin.

app/javascript/mastodon/service_worker/entry.js
```diff
@@ -0,0 +1,49 @@
...
+  cacheFirstAndUpdateAfter,
+  networkFirst,
+  deleteAllFromCacheMatching,
+  onRequest
+} from './helpers';
+
+const SIGN_OUT_REGEX = /^\/auth\/sign_out/;
+const EMOJI_REGEX = /^\/emoji\//;
+const WEB_REGEX = /^\/web\//;
+const ABOUT_REGEX = /^\/about/;
```
ramlmn: Shouldn't this be `/^\/about\//`?

ramlmn:
> more intelligent redirecting, don't cache all `/web`, `/about`, `/settings` pages individually.

@nolanlawson 

For `/web`, as the same response is served for all the pages (can be considered as app shell), you might just cache a single page and respond with the same for all requests under `/web` (and also the same for `/`). When we have network, we should always update this in cache.

For `/about` and `/settings`, you might have to cache them individually (or not let user change settings).

And `/auth/sign_in` should be redirected to `/web/timelines/home` assuming the user is logged in.

nolanlawson:
> Use cacheName option of offline-plugin in webpack.production.js (to prevent future issues)

I tested this out; it merely adds a prefix to the cacheName. I'm not sure what issue it would prevent, because collisions seem impossible?

> Use semver or any other method to add version key in package.json so that it can be used in both sw-entry.js, and in webpack.production.js as version option

I thought of this, but since `service_worker/entry.js` just caches on-the-fly it doesn't need to update the whole cache every time the Mastodon version changes. Also I don't have a system in place yet to delete old caches when the version is updated. So I mostly put that in place as a safeguard; the idea is that if we end up shipping a ServiceWorker that has a bug in it, then and only then will we need to increment the version.

app/javascript/mastodon/service_worker/entry.js
nolanlawson: There is also an `about` page without a `/` at the end.

ramlmn:
> I tested this out; it merely adds a prefix to the cacheName. I'm not sure what issue it would prevent, because collisions seem impossible?

Well, adding `cacheName` and `version` would just make it easy to walk through all the caches and remove old ones.

``` js
caches.keys().then(cacheNames => {
  cacheNames.map(cacheName => {
    if (cacheName !== `${cacheName}-v${version}`) {
      caches.delete(cacheName);
    }
  });
});
```

nolanlawson:
offline-plugin does this automatically, but yeah it couldn't hurt

-----
Nolan Lawson
http://nolanlawson.com

app/javascript/mastodon/service_worker/helpers.js
```diff
@@ -0,0 +1,167 @@
+// Various caching strategies to make entry.js easier to read. Partially
+// inspired by https://github.com/GoogleChrome/sw-toolbox
+
+const VERSION = '1.0.0'; // if necessary, we can increment this
+const CACHE_KEY = `mastodon-sw-v${VERSION}`;
+
+// taken from https://github.com/GoogleChrome/sw-toolbox/blob/master/lib/options.js
+const SUCCESS_RESPONSES = /^0|([123]\d\d)|(40[14567])|410$/
```
NekR: I'm not sure here, how 40* are success responses or other in 300/100 range?

NekR:
> offline-plugin does this automatically, but yeah it couldn't hurt

Yes, webpack automatically adds the version (passed as `version` option, default is the build date) to the cache name. So it looks like this: `webpack-offline:v248`. If you add a cache name option, it will be `webpack-offline:custom-cache-name:v248`.

In that name  `webpack-offline:custom-cache-name:` (`webpack-offline:` in case of no custom cache name) is a prefix and `v248` is a version.
On each SW update `offline-plugin` use the prefix from the **new** SW and removes old caches which has the same prefix. So, if you are going to use dynamic cache name then you will be leaving behind all previous caches on each update.

You actually should never change your cache name. You probably may want to add custom **static** cache name if you are thinking that someday you may have 2+ SWs generated by the `offline-plugin` on the same domain (that's the reason why custom cache names option has been added).

nolanlawson:
Yeah in this PR I am using a custom solution inside of `entry` to maintain a separate cache (basically it's I'm using `offline-plugin` for Webpack assets and `sw-toolbox`-esque custom implementation for everything else). It's fairly easy to keep the two caches separate since `offline-plugin`'s cache name isn't likely to collide with anything. I also doubt anyone will ever serve two `offline-plugin`-using sites from the same origin, although it can't hurt to add the cache name anyway. We wouldn't use a dynamic cache name; we'd probably just call it `mastodon` or something.

app/javascript/mastodon/service_worker/helpers.js
nolanlawson: I copypasta'd this from the sw-toolbox project. I checked a few different HTTP responses and it seems to correctly detect error ranges vs success ranges.

app/javascript/mastodon/main.js
```diff
@@ -27,6 +27,12 @@ function main() {
   // import customization styles
   require.context('../../assets/stylesheets/', false, /custom.*\.scss$/);
 
+  if (process.env.NODE_ENV === 'production') {
+    // avoid offline in dev mode because it's harder to debug
+    const OfflinePluginRuntime = require('offline-plugin/runtime');
+    OfflinePluginRuntime.install();
```
ramlmn: Install SW after page load (seems to affect the actual page load event).
``` js
window.addEventListener('load', _ => {
  const OfflinePluginRuntime = require('offline-plugin/runtime');
  OfflinePluginRuntime.install();
});

ramlmn:
Can we cache only `woff2` fonts, because all browsers that supports SW also support `woff2`?

app/javascript/mastodon/main.js
NekR: Do you mean ServiceWorker blocks loading of other resources? Are you on http2?

nolanlawson:
I'm keeping my list of remaining TODOs in [this comment](https://github.com/tootsuite/mastodon/pull/3052#issuecomment-301362450). Once those are all fixed this PR should be ready for final review.

app/javascript/mastodon/main.js
nolanlawson: Unless `offline-plugin` is running a lot of code on the main thread, there's no reason I can see to delay it. Even if that's the case, I think `requestIdleCallback()` would be a better choice.

And yeah @NekR most Mastodon instances should be on http2.

ramlmn: > Do you mean ServiceWorker blocks loading of other resources?

@NekR Don't know how it would interact with the main thread in the wild (grr.. grr... previous versions of Chrome). So, it should be safe to use it after the load event.

> Even if that's the case, I think `requestIdleCallback()` would be a better choice.

@nolanlawson `requestIdleCallback()` can be used to do like less important stuff (error reporting, analytics and stuff). In some cases it might not even be called (like on a slow mobile device).

There is still we can add from [https://youtu.be/CPP9ew4Co0M?t=12m2s](https://youtu.be/CPP9ew4Co0M?t=12m2s) 😜

nolanlawson:
I'm abandoning this PR for now. Issues:

- caching HTML is difficult due to embedded dynamic content (CSRF token)
- URLs that resolve to same content are tricky to manage
- not clear that this offers a huge perf win compared to standard long cache times + `cache-control:immutable`
- ServiceWorker already added for Push Notifications, need to merge
- this doesn't add true offline support anyway since toots aren't stored in IndexedDB, and while offline we are flooded with toast error messages

Abandoning for now, maybe true offline support can be picked up later.

sorin-davidoi:
Don't you think it is still worth to do for fonts + bundles as a start? I could look into adding/improving all the things you mention later on (e.g. #4459).

PR #3055, opened 2017-05-14
Hide signup path unless the instance allows new registration.

treby:
When an instance does not allow register account ( `open_registrations = false` ),

Before

After


ykzts:
What happens to languages ​​other than en and ja with this change?

treby:
@ykzts Thank you for your comment!

Actually, I've tried to add other language's keys by `i18n-tasks add-missing`, then there is so big diff (although not pr's related keys are added). So I wondered if is there any guidelines for new i18n key?

Of course, I'm happy to add other languages key by en translation, or just split existing `landing_strip_html` value by my hand (but since I'm not good at many languages, I may break existing translation...).

What can I do to improve this pr?

app/views/shared/_landing_strip.html.haml
```diff
@@ -1,6 +1,6 @@
 .landing-strip
   = t('landing_strip_html',
     name: content_tag(:span, display_name(account), class: :emojify),
-    domain: site_hostname,
-    link_to_root_path: link_to(content_tag(:strong, site_hostname), root_path),
-    sign_up_path: new_user_registration_path)
+    link_to_root_path: link_to(content_tag(:strong, site_hostname), root_path))
+  - if open_registrations?
```
ykzts: It would be better to use [`InstancePresenter`](https://github.com/tootsuite/mastodon/blob/a588358f41f36b2df1dcc4bb2c3df6c54ed52f88/app/presenters/instance_presenter.rb) like [`about/_links.html.haml`](https://github.com/tootsuite/mastodon/blob/a588358f41f36b2df1dcc4bb2c3df6c54ed52f88/app/views/about/_links.html.haml).

ykzts:
I think that it is okay to temporarily make it into English.

ykzts:
Of course you should divide within the range you can separate.

treby:
Okay I'll try it.

app/views/shared/_landing_strip.html.haml
treby: hmm...
To make this pr, at first, I tried to use `InstancePresenter` to get the configuration (and DRY).
but IMHO, it's not a very good idea to provide the presenter instance only for `_landing_strip.html.haml` .

Now, `_landing_strip.html.haml` is called from

- `app/views/accounts/show.html.haml` ( `app/controllers/accounts_controller.rb` )
- `app/views/stream_entries/show.html.haml` (`app/controllers/stream_entries_controller.rb` )

There two controllers are completely difference (unlike `about_controller`), so I think it's hard for developers to know how controller's `InstanceRepresenter` instance variable is used.

Of course, I'll add both controllers by your request, but my opinion is above.

treby: This is an patch image for your request (that I'm thinking).

```diff
diff --git a/app/controllers/accounts_controller.rb b/app/controllers/accounts_controller.rb
index 8eda9633..319955cf 100644
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -6,6 +6,7 @@ class AccountsController < ApplicationController
   def show
     respond_to do |format|
       format.html do
+        @instance_presenter = InstancePresenter.new
         @statuses = @account.statuses.permitted_for(@account, current_account).order('id desc').paginate_by_max_id(20, params[:max_id], params[:since_id])
         @statuses = cache_collection(@statuses, Status)
       end
diff --git a/app/controllers/stream_entries_controller.rb b/app/controllers/stream_entries_controller.rb
index a9ee7350..64277566 100644
--- a/app/controllers/stream_entries_controller.rb
+++ b/app/controllers/stream_entries_controller.rb
@@ -13,6 +13,7 @@ class StreamEntriesController < ApplicationController
       format.html do
         return gone if @stream_entry.activity.nil?
 
+        @instance_presenter = InstancePresenter.new
         if @stream_entry.activity_type == 'Status'
           @ancestors   = @stream_entry.activity.reply? ? cache_collection(@stream_entry.activity.ancestors(current_account), Status) : []
           @descendants = cache_collection(@stream_entry.activity.descendants(current_account), Status)
diff --git a/app/helpers/application_helper.rb b/app/helpers/application_helper.rb
index 8f1cd8fc..92ffac33 100644
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,10 +9,6 @@ module ApplicationHelper
     !user_signed_in? && !single_user_mode?
   end
 
-  def open_registrations?
-    Setting.open_registrations
-  end
-
   def add_rtl_body_class(other_classes)
     other_classes = "#{other_classes} rtl" if [:ar, :fa, :he].include?(I18n.locale)
     other_classes
diff --git a/app/views/accounts/show.html.haml b/app/views/accounts/show.html.haml
index a1904910..e21827da 100644
--- a/app/views/accounts/show.html.haml
+++ b/app/views/accounts/show.html.haml
@@ -9,7 +9,7 @@
   = render 'og', account: @account, url: short_account_url(@account, only_path: false)
 
 - if show_landing_strip?
-  = render partial: 'shared/landing_strip', locals: { account: @account }
+  = render partial: 'shared/landing_strip', locals: { account: @account, instance_presenter: @instance_presenter }
 
 .h-feed
   %data.p-name{ value: "#{@account.username} on #{site_hostname}" }/
diff --git a/app/views/shared/_landing_strip.html.haml b/app/views/shared/_landing_strip.html.haml
index 3cc61a2c..77b2678a 100644
--- a/app/views/shared/_landing_strip.html.haml
+++ b/app/views/shared/_landing_strip.html.haml
@@ -2,5 +2,5 @@
   = t('landing_strip_html',
     name: content_tag(:span, display_name(account), class: :emojify),
     link_to_root_path: link_to(content_tag(:strong, site_hostname), root_path))
-  - if open_registrations?
+  - if instance_presenter&.open_registrations
     = t('landing_strip_signup_html', sign_up_path: new_user_registration_path)
diff --git a/app/views/stream_entries/show.html.haml b/app/views/stream_entries/show.html.haml
index d01e82af..8ba16b78 100644
--- a/app/views/stream_entries/show.html.haml
+++ b/app/views/stream_entries/show.html.haml
@@ -13,7 +13,7 @@
   %meta{ property: 'twitter:card', content: 'summary' }/
 
 - if show_landing_strip?
-  = render partial: 'shared/landing_strip', locals: { account: @stream_entry.account }
+  = render partial: 'shared/landing_strip', locals: { account: @stream_entry.account, instance_presenter: @instance_presenter }
 
 .activity-stream.activity-stream-headless.h-entry
   = render partial: "stream_entries/#{@type}", locals: { @type.to_sym => @stream_entry.activity, include_threads: true }
```

treby:
@ykzts Added all language's translation and replied your review

app/views/shared/_landing_strip.html.haml
ykzts: Well surely it is.

ykzts (approved):

PR #3072, opened 2017-05-15
Fix change of status callbacks not setting in_reply_to_account_id and

Gargron:
possibly others when expected. Add some tests for it

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -72,7 +72,7 @@ class Status < ApplicationRecord
   cache_associated :account, :application, :media_attachments, :tags, :stream_entry, mentions: :account, reblog: [:account, :application, :stream_entry, :tags, :media_attachments, mentions: :account], thread: :account
 
   def reply?
-    !in_reply_to_id.nil? || super
+    !in_reply_to_id.nil? || attributes['reply']
   end
 
   def local?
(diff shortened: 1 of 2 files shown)


app/models/status.rb
```diff
@@ -72,7 +72,7 @@ class Status < ApplicationRecord
   cache_associated :account, :application, :media_attachments, :tags, :stream_entry, mentions: :account, reblog: [:account, :application, :stream_entry, :tags, :media_attachments, mentions: :account], thread: :account
 
   def reply?
-    !in_reply_to_id.nil? || super
+    !in_reply_to_id.nil? || attributes['reply']
```
Gargron: Technically unrelated but this is what the "super" did and it's more clear this way and possibly faster

spec/services/process_feed_service_spec.rb
```diff
@@ -31,6 +31,16 @@
         expect(Status.find_by(uri: 'tag:kickass.zone,2016-10-10:objectId=2:objectType=Status')).to_not be_nil
       end
 
+      it 'marks replies as replies' do
+        status = Status.find_by(uri: 'tag:kickass.zone,2016-10-10:objectId=2:objectType=Status')
+        expect(status.reply?).to be true
+      end
+
+      it 'sets accound being replied to when possible' do
```
beatrix-bitrot: typo accound -> account

beatrix-bitrot (approved):
looks good, let's try it out!

PR #3079, opened 2017-05-16
Improve Account#triadic_closures

alpaca-tc:
@Gargron Thank you for replying the other day 😄 

I improved `Account#triadic_closures` a bit.
Currently, we suggest accounts with mixed `Account#triadic_closures`. This is nice feature for us 👍 

- Reject blocked and muted accounts
- Add offset option

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -259,24 +259,30 @@ class Account < ApplicationRecord
       nil
     end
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

Gargron:
If you added an API/UI for suggestions, I would be interested in that 👍

PR #3080, opened 2017-05-16
Make faster ProcessFeedService

harukasan:
In our slow queries I found: 
```
SELECT "accounts".* FROM "accounts" WHERE "accounts"."uri" = '...' LIMIT 1;
```

and the query plan is:
```
paon=> EXPLAIN ANALYZE SELECT "accounts".* FROM "accounts" WHERE "accounts"."uri" = '...' LIMIT 1;
 Limit  (cost=0.00..7710.39 rows=1 width=899) (actual time=202.203..202.203 rows=0 loops=1)
   ->  Seq Scan on accounts  (cost=0.00..46262.32 rows=6 width=899) (actual time=202.201..202.201 rows=0 loops=1)
         Filter: ((uri)::text = '...'::text)
         Rows Removed by Filter: 297178
 Planning time: 1.129 ms
 Execution time: 202.238 ms
(6 rows)
```

It is due to `"accounts"."uri"` have no index.

### DIFF
--- a/db/migrate/20170516072309_add_index_accounts_on_uri.rb
+++ b/db/migrate/20170516072309_add_index_accounts_on_uri.rb
@@ -1,0 +1,6 @@
+class AddIndexAccountsOnUri < ActiveRecord::Migration[5.0]
+  def change
+    add_index :accounts, :uri
+  end
+end
+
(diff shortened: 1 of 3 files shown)


db/migrate/20170516072309_add_index_accounts_on_uri.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexAccountsOnUri < ActiveRecord::Migration[5.0]
+  def change
+    add_index :accounts, :uri
+  end
+end
+
```
alpaca-tc: Please remove blank line

Gargron (approved):

PR #3094, opened 2017-05-17
Use REDIS_DB in streaming

h-izumi:
According to c997091166ba1801eea3a587c913b020b9b84ce4, REDIS_DB is still used in web/sidekiq (`config/initializers/redis.rb`).
But seems to be not used in streaming.


beatrix-bitrot (dismissed):
Good catch!

All connection parameters other than password have defaults here. Can you default the db to 0 here please?

Thanks

h-izumi:
@beatrix-bitrot Thank you for review. I fix it!

Gargron (approved):

PR #3096, opened 2017-05-17
Use DB connection params in .env even on dev/test environment

unarist:
Currently, to avoid unix-domain sockets for DB connection or connect to the DB on another host for development environment, we have to edit `streaming/index.js` and so on. This patch allows those customization for dev/test environment with `.env` flie, like `.env.production` for production environment.

Note that I have left DB_NAME because it should be vary between development and test.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,5 +1,9 @@
 default: &default
   adapter: postgresql
+  username: <%= ENV['DB_USER'] || '' %>
+  password: <%= ENV['DB_PASS'] || '' %>
+  host: <%= ENV['DB_HOST'] || '' %>
+  port: <%= ENV['DB_PORT'] || '' %>
   pool: <%= ENV["DB_POOL"] || ENV['MAX_THREADS'] || 5 %>
   timeout: 5000
   encoding: unicode
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -68,8 +68,11 @@ if (cluster.isMaster) {
   // Cluster worker
   const pgConfigs = {
     development: {
+      user:     process.env.DB_USER || null,
+      password: process.env.DB_PASS || null,
       database: 'mastodon_development',
-      host:     '/var/run/postgresql',
+      host:     process.env.DB_HOST || '/var/run/postgresql',
+      port:     process.env.DB_PORT || null,
       max:      10
     },
 


Gargron:
You can just use DATABASE_URL/REDIS_URL to customize everything in detail to avoid getting an env variable for every setting

unarist:
Oh, I haven't noticed that `streaming/index.js` also understand `DATABASE_URL`.

However, `DATABASE_URL` seems to only effects in current environment. Running `db:setup` on development environment also creates for test environment, so it fails.

```console
$ echo $DATABASE_URL
//postgres@db/

$ rails runner 'puts ActiveRecord::Base.configurations'
{"default"=>{"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding"=>"unicode"}, "development"=>{"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding"=>"unicode", "database"=>"mastodon_development", "username"=>"postgres", "host"=>"db"}, "test"=>{"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding"=>"unicode", "database"=>"mastodon_test"}, "production"=>{"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding"=>"unicode", "database"=>"mastodon", "username"=>"mastodon", "password"=>nil, "host"=>"localhost", "port"=>5432, "prepared_statements"=>true}}

$ rails db:setup
Created database 'mastodon_development'
could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
Couldn't create database for {"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding"=>"unicode", "database"=>"mastodon_test"}
rails aborted!
PG::ConnectionBad: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
```

masarakki:
`rake db:create` or `db:migrate` changes both development and test databases.
When using `DATABASE_URL`, only development database was connected via `DATABASE_URL`,
but test database was connected via `database.yml` and it failed.

```
masaki@ganesa:mastodon/ (master) $ docker run -p 5432:5432 -d --restart=always postgres:alpi
ne
f2cb8bf92224a70e402d8909cfcd6d308b62c0519abfca7b64c3c9de2b964870
masaki@ganesa:mastodon/ (master) $ cat .env                                   
DATABASE_URL=postgres://postgres@localhost:5432/mastodon_development
masaki@ganesa:mastodon/ (master) $ rake db:create                              
Created database 'mastodon_development'
could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
Couldn't create database for {"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "encoding
"=>"unicode", "database"=>"mastodon_test"}
rake aborted!
PG::ConnectionBad: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

I think using `DATABASE_URL` is only for production, so in develpoment, it is better to use  ENV['DB_*'] in `defaults` of `database.yml`.

abcang:
There is related issue. https://github.com/rails/rails/issues/28827

unarist:
I'm closing this since similar fix has been introduced at #3909.

PR #3098, opened 2017-05-17
After remote follow worker specs

mjankowski:
Adds coverage, refactors those classes.

A question before this is merged:

- In the case of both workers, we skip the processing if the found account is nil, which makes sense
- In the case of the after remote follow worker, we process for locked accounts and skip processing for unlocked accounts
- In the case of the after remote follow request worker, we process for unlocked and skip for locked

I believe that I preserved this behavior in the spec coverage and in the refactor, but I just want to confirm that that is correct, because I did not do a thorough deep dive through the code to wrap my head around all the places this is used.

### DIFF
--- a/app/workers/after_remote_follow_request_worker.rb
+++ b/app/workers/after_remote_follow_request_worker.rb
@@ -5,15 +5,27 @@ class AfterRemoteFollowRequestWorker
 
   sidekiq_options queue: 'pull', retry: 5
 
+  attr_reader :follow_request
+
   def perform(follow_request_id)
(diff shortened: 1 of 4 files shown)


Gargron:
The purpose of these workers is that we may not have the latest information about a remote account at time of follow. So we want to check, if the account is *actually* still locked/unlocked, and if it's different to what we expected, send the relevant action instead.

mjankowski:
Cool, given that, I think this is probably fine.

Gargron (approved):

PR #3101, opened 2017-05-17
Fix character counter not updating for bio

Gargron:


config/locales/simple_form.en.yml
```diff
@@ -4,14 +4,14 @@ en:
     hints:
       defaults:
         avatar: PNG, GIF or JPG. At most 2MB. Will be downscaled to 120x120px
-        display_name: 
-            one: '1 character left'
-            other: '%{count} characters left'
+        display_name:
+            one: '<span class="name-counter">1</span> character left'
```
Gargron: This stuff is not ideal at all, but I didn't introduce it, so this is not the place to complain about it. But perhaps in the future the profile editing page should become a React component because I really dislike mixing static forms with jQuery-esque dynamic parts when we've got all of the React stuff already set up.

ykzts (approved):

PR #3103, opened 2017-05-17
Fix #942 - Add LDAP support

### LINKED ISSUE #942: [Feature Request] LDAP integration
Having LDAP integration would allow organizations with existing ldap infrastructure to automatically create a mastodon profile for its users. 

Gitlab has good LDAP integration and could be used as an example. https://docs.gitlab.com/ce/administration/auth/ldap.html

* * * *

Gargron:
This adds an optional authentication against an LDAP server. New configuration options:

```bash
# LDAP_ENABLED=true
# LDAP_HOST=localhost
# LDAP_PORT=389
# LDAP_BASE="dc=example,dc=org"
# LDAP_BIND_DN="cn=admin,dc=example,dc=org"
# LDAP_BIND_PASSWORD="admin"
# LDAP_EMAIL_FIELD=mail
# LDAP_USERNAME_FIELD=cn
```

This changes the user model slightly. "External" users (not created through Mastodon itself) do not need to have an internally saved password. Nor do they need e-mail validation, since the e-mail comes from the LDAP server. To make that possible, I had to remove the Validatable devise module which wasn't flexible enough, and re-do the validations myself in the User class.

Potential things to look out for:

- If you don't disable normal, internal registration, somebody signs up, and an LDAP user signs in who has the same username, an error will be thrown because the username is already taken
- Similarly, with the same e-mail, the LDAP user will get access to the already existing user account
- Probably should disable e-mail confirmation and password reset stuff for users without password??? And account settings too??
- What happens if you change the e-mail in LDAP? A new user record will attempt to be created, but fail because the username is already taken (but this is undesired anyway, you'd wanna keep the old user record!)
- Possibly more stuff..?? Comments welcome


jidea:
where is the SSL/TLS stuff ?

Gargron:
@jidea I'm pretty sure it's handled by the net/ldap gem, does it need anything else?

Gargron:
@jidea Ah I see, there is an extra connection param, `ldap.encryption = :simple_tls` when you want to have SSL. Yes I should add that.

jcgruenhage:
I'd rather have TLS optional, since a lot of LDAP servers aren't exposed to the public anyway. Mine for example is just in a docker container, and you can only reach it from other containers on the same host. Encrypting that traffic would east CPU cycles ^^

jcgruenhage:
And I'll comment on the other stuff you've mentioned above once I'm at a proper keyboard again :D

jcgruenhage:
> If you don't disable normal, internal registration, somebody signs up, and an LDAP user signs in who has the same username, an error will be thrown because the username is already taken.

This is something that can't be circumvented, unless you check that there is no user in the LDAP directory, but that would possibly block names that will never be used so just dealing with it the "come first serve first" way seems the way to go. (Or make it configurable)


> Similarly, with the same e-mail, the LDAP user will get access to the already existing user account
Probably should disable e-mail confirmation and password reset stuff for users without password??? And account settings too??

Email confirmation and password management should be handled by LDAP, yes, but how will an LDAP user get access to an account with the same name? That should be prevented by the internal/external type of users.


> What happens if you change the e-mail in LDAP? A new user record will attempt to be created, but fail because the username is already taken (but this is undesired anyway, you'd wanna keep the old user record!)

Can you change your mail address in mastodon? If so, just have it so that if some user tries to log in with his new email after a change, and the dn of that user equals the dn of the original user (maybe store that in the db), then the email gets changed and the user is successfully logged in.

gloaec (commented):
This works, but it disables the devise "validatable" mechanism. I think i would be more interesting to implement an oauth strategy so it can leave some space for some other authentication methods.

Gargron:
Closed in favour of #4032

PR #3109, opened 2017-05-18
i18l: Hebrew translation updates

seefood:


config/locales/he.yml
```diff
@@ -257,7 +257,7 @@ he:
       muting: רשימת השתקות
     upload: יבוא
   landing_strip_html: <strong>%{name}</strong> על השרת <strong>%{domain}</strong>. ניתן לעקוב אחריהם או לדבר איתם בעזרת חשבון על כל שרת קהילה ברחבי הפדרציה. ליצירת חשבון, ניתן<a href="%{sign_up_path}">להרשם כאן</a>.
-  landing_strip_signup_html: If you don't, you can <a href="%{sign_up_path}">sign up here</a>.
+  landing_strip_signup_html: אחרת, ניתן <a href="%{sign_up_path}">להרשם כאן</a>.
```
Gargron: Remove this from the one above

Gargron (changes requested):

config/locales/he.yml
seefood: Thanks for noticing. I use meld to pop out changed keys, I haven't noticed the change in value.

went over again, found a few more changes, pushed a fix.

Gargron (approved):

config/locales/simple_form.he.yml
```diff
@@ -5,9 +5,13 @@ he:
       defaults:
         avatar: 'PNG, GIF או JPG. מקסימום 2MB. גודל התמונה יוקטן ל-120x120px'
         display_name: 'מקסימום 30 תווים'
```
Gargron: Syntax error, can't be value + children at the same time, either one or the other

Gargron (changes requested):

config/locales/simple_form.he.yml
seefood: True.... one of those keys that became a node. pushing a fix.

Gargron (approved):

PR #2338, opened 2017-04-23
Replace inline styles with CSS classes (BEM)

stephenburgess8:
#584 #2028 

Mastodon uses a combination of styles-in-javascript and regular css classes, mostly following the BEM naming convention.

Mostly remove javascript styles to CSS with the notable exceptions of when properties are being dynamically set in the Javascript. The goal is a pure refactor, no style changes. The biggest benefit of this would be to facilitate theme setting.

Follow a BEM naming convention. Rule of thumb: alphabetize css attributes.

Testing notes: I have compared much of the UI and have not been able to identify any regressions. However due to the large number of lines changed here, regressions seem likely. I intend to do more regression testing if the general concept is accepted by the core maintainers.

PR #3115, opened 2017-05-18
Replace inline styles with stylesheet

ykzts:
follow up #2338

### DIFF
--- a/app/javascript/mastodon/components/autosuggest_textarea.js
+++ b/app/javascript/mastodon/components/autosuggest_textarea.js
@@ -79,6 +79,7 @@ class AutosuggestTextarea extends ImmutablePureComponent {
     }
 
     // auto-resize textarea
+    e.target.style.height = 'auto';
     e.target.style.height = `${e.target.scrollHeight}px`;
 
     this.props.onChange(e);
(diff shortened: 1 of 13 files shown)


app/javascript/mastodon/components/autosuggest_textarea.js
```diff
@@ -79,6 +79,7 @@ class AutosuggestTextarea extends ImmutablePureComponent {
     }
 
     // auto-resize textarea
+    e.target.style.height = 'auto';
```
Gargron: Since this isn't CSS, this is instantly overriden by the next line, I think.

ykzts: Surprisingly, without this line the height of the text area grows but it will not shrink.

Gargron: Interesting. In the future I would like to consider some form of declarative (state+props=>render) way of handling textarea height, instead of using these side effects and direct DOM manipulation.

ykzts:
fix for codeclimate.

Gargron (approved):

PR #3118, opened 2017-05-18
test: try to add atom serializer spec.

treby:
I tried to add pending specs.

This patch is experimental. I read spec file (only this one now), and it seems there are better ways. I wanna work on writing specs. So, I just want to know writing spec direction :)
Tell me your thinkings.

Thanks.

### DIFF
--- a/spec/lib/atom_serializer_spec.rb
+++ b/spec/lib/atom_serializer_spec.rb
@@ -190,8 +190,24 @@ RSpec.describe AtomSerializer do
   end
 
   describe '#follow_request_salmon' do
-    it 'returns dumpable XML'
-    it 'triggers follow request when processed'
+    let(:follow_request) { Fabricate(:follow_request, account: author, target_account: receiver) }
+    let(:atom_serializer) { described_class.new }
+    subject { atom_serializer.follow_request_salmon(follow_request) }
+
+    it { is_expected.to be_instance_of Ox::Element }
+
+    describe 'with .render' do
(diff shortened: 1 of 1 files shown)


spec/lib/atom_serializer_spec.rb
```diff
@@ -190,8 +190,24 @@
...
+
+    it { is_expected.to be_instance_of Ox::Element }
+
+    describe 'with .render' do
+      let(:rendered) { described_class.render(subject) }
+      it 'returns dumpable XML' do
+        expect(rendered).to be_a String
+      end
+
+      it 'triggers follow request when processed' do
```
treby: [IMHO] This spec is not for `AtomSeriealizer` ,  but `ProcessInteractionService`

spec/lib/atom_serializer_spec.rb
```diff
@@ -190,8 +190,24 @@
   end
 
   describe '#follow_request_salmon' do
-    it 'returns dumpable XML'
-    it 'triggers follow request when processed'
+    let(:follow_request) { Fabricate(:follow_request, account: author, target_account: receiver) }
+    let(:atom_serializer) { described_class.new }
+    subject { atom_serializer.follow_request_salmon(follow_request) }
+
+    it { is_expected.to be_instance_of Ox::Element }
```
treby: Actually, most of `AtomSerializer`'s method returns `Ox::Element` . I think we should test what included by the `Ox::Element`.

PR #3121, opened 2017-05-18
Update Rails to version 5.1.1

mjankowski:
Barebones upgrade - does not do much in the way of actually using new features. It should solve all relevant deprecations, though.


app/helpers/admin/filter_helper.rb
```diff
@@ -18,14 +18,18 @@ def table_link_to(icon, text, path, options = {})
...
 
   def filter_link_class(new_url)
-    filtered_url_for(params) == new_url ? 'selected' : ''
+    filtered_url_for(controller_request_params) == new_url ? 'selected' : ''
   end
 
-  def filtered_url_for(params)
-    url_for filter_params(params)
+  def controller_request_params
+    params.permit!
```
mjankowski: This is a bad solution to this problem ... but it did get these specs passing.

This whole thing probably needs a rewrite, maybe separately and before this PR merged.

app/helpers/application_helper.rb
```diff
@@ -20,7 +20,7 @@ def add_rtl_body_class(other_classes)
 
   def favicon_path
     env_suffix = Rails.env.production? ? '' : '-dev'
-    asset_path "favicon#{env_suffix}.ico"
+    asset_path "favicon#{env_suffix}.ico", skip_pipeline: true
```
mjankowski: Since this will always just be in `public` maybe we don't even need to use asset_path here?

app/lib/application_extension.rb
```diff
@@ -4,6 +4,6 @@ module ApplicationExtension
   extend ActiveSupport::Concern
 
   included do
-    validates :website, url: true, unless: 'website.blank?'
+    validates :website, url: true, if: :website?
```
mjankowski: Explore using `allow_nil` insetad?

app/models/notification.rb
```diff
@@ -48,10 +48,6 @@ class Notification < ApplicationRecord
 
   cache_associated :from_account, status: STATUS_INCLUDES, mention: [status: STATUS_INCLUDES], favourite: [:account, status: STATUS_INCLUDES], follow: :account
 
-  def activity(eager_loaded = true)
```
mjankowski: The method signature changed here, and there is no longer an `eager_loaded` argument passed in.

Removing this did not break any specs ... but my guess is there was just no coverage. Worth investigating how this was used (probably avoid queries?) and make sure it's not a regression.

app/models/user.rb
```diff
@@ -45,7 +45,7 @@ class User < ApplicationRecord
   belongs_to :account, inverse_of: :user, required: true
   accepts_nested_attributes_for :account
 
-  validates :locale, inclusion: I18n.available_locales.map(&:to_s), unless: 'locale.nil?'
+  validates :locale, inclusion: I18n.available_locales.map(&:to_s), if: :locale?
```
mjankowski: `allow_nil` ?

app/views/layouts/embedded.html.haml
```diff
@@ -2,7 +2,7 @@
 %html{ lang: I18n.locale }
   %head
     %meta{ charset: 'utf-8' }/
-    = stylesheet_link_tag 'application', media: 'all'
+    = stylesheet_pack_tag 'application', media: 'all'
```
mjankowski: Could be separate PR unrelated to rails 5.1

app/models/notification.rb
Gargron: Right so when you're doing rails preloading using `includes(:model)` but on a polymorphic association, when you're using the polymorphic association method, it loads the record fresh, ignoring the preload. So I overloaded that method to use the preloaded records. However, sometimes they are not preloaded, which is when I'd set the param to false. I don't remember why though.

app/views/layouts/embedded.html.haml
Gargron: Oh, this has been broken since webpack? Oops. Good catch.

app/models/notification.rb
mjankowski: Got it ... I'll do a quick check to see if we can maintain that behavior.

app/views/layouts/embedded.html.haml
mjankowski: 5.1 kicks up a warning about things which aren't owned by the asset pipeline ... even if they are pointing at things which exist via webpack.

beatrix-bitrot:
nice work!

mjankowski:
A bunch of stuff which was in here originally was pulled out and merged via other PRs - https://github.com/tootsuite/mastodon/pull/3137 - https://github.com/tootsuite/mastodon/pull/3129 - https://github.com/tootsuite/mastodon/pull/3125 - https://github.com/tootsuite/mastodon/pull/3124 - https://github.com/tootsuite/mastodon/pull/3123 - thanks @Gargron for doing those so quickly.

This one has been rebased after those were all added, and I think now reflects close to a minimum of what's required for 5.1 upgrade (but still does not go deep into USING many new features).

I think I will take a quick look at the polymorphic eager loading thing, and then this might be ready to go (or more review).

mjankowski:
This is probably ready now.

Gargron (approved):

PR #3122, opened 2017-05-18
feat(babel): Strip prop types

sorin-davidoi:
No need to ship our `propTypes`. Unfortunately this does not strip them from imported modules and does not remove the two modules from the bundle.

Before:
```bash
> ls -l --block-size=KB public/packs | grep 'application.*.js'
-rw-r--r-- 1 sorin sorin 1015kB May 19 01:22 application-e2333eb5abd97b680820.js
-rw-r--r-- 1 sorin sorin  241kB May 19 01:22 application-e2333eb5abd97b680820.js.gz
-rw-r--r-- 1 sorin sorin 6357kB May 19 01:22 application-e2333eb5abd97b680820.js.map
```

After:
```bash
> ls -l --block-size=KB public/packs | grep 'application.*.js'
-rw-r--r-- 1 sorin sorin 1007kB May 19 01:26 application-4cc778a2892b9ed4f620.js
-rw-r--r-- 1 sorin sorin  240kB May 19 01:26 application-4cc778a2892b9ed4f620.js.gz
-rw-r--r-- 1 sorin sorin 6305kB May 19 01:26 application-4cc778a2892b9ed4f620.js.map
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -33,13 +33,16 @@
     "production": {
       "plugins": [
         "lodash",
-        ["transform-react-remove-prop-types", {
-          "mode": "remove",
-          "removeImport": true,
-          "additionalLibraries": [
-            "react-immutable-proptypes"
-          ]
-        }],
+        [
+          "transform-react-remove-prop-types",
+          {
+            "mode": "remove",
+            "removeImport": true,
+            "additionalLibraries": [
+              "react-immutable-proptypes"
+            ]
+          }
+        ],
         [
           "transform-runtime",
           {
--- a/.babelrc
+++ b/.babelrc
@@ -33,6 +33,13 @@
     "production": {
       "plugins": [
         "lodash",
+        ["transform-react-remove-prop-types", {
+          "mode": "remove",
+          "removeImport": true,
+          "additionalLibraries": [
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -33,6 +33,13 @@
     "production": {
       "plugins": [
         "lodash",
+        ["transform-react-remove-prop-types", {
+          "mode": "remove",
+          "removeImport": true,
+          "additionalLibraries": [
+            "react-immutable-proptypes"
+          ]
+        }],
```
ykzts: replace indent format

ykzts (approved):

package.json
```diff
@@ -104,6 +104,7 @@
   "devDependencies": {
     "@kadira/storybook": "^2.35.3",
     "babel-eslint": "^7.2.2",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
```
ykzts: move to `depedencies`. `rails assets: precompile` runs `yarn install` with `NODE_ENV=production`.

ykzts (changes requested):

nolanlawson (approved):

ykzts (approved):

PR #3128, opened 2017-05-19
Define instance method outside #included

alpaca-tc:
```
# no good
module A
  extend ActiveSupport::Concern

  included do
    def hello_world; end
  end
end

# good
module A
  extend ActiveSupport::Concern

  included do
    ...
  end

  def hello_world; end
end
```

### DIFF
--- a/app/models/concerns/streamable.rb
+++ b/app/models/concerns/streamable.rb
@@ -6,36 +6,38 @@ module Streamable
   included do
     has_one :stream_entry, as: :activity
 
-    def title
-      super
-    end
-
-    def content
-      title
-    end
-
-    def target
-      super
-    end
-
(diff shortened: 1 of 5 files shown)


app/models/concerns/streamable.rb
```diff
@@ -6,36 +6,38 @@ module Streamable
...
-      super
+    after_create do
+      account.stream_entries.create!(activity: self, hidden: hidden?) if needs_stream_entry?
     end
+  end
 
-    def content
-      title
-    end
+  def title
```
alpaca-tc: IMO, `#title` , `#target` and `#thread` are not needed because it calls only `super` 🤔

Gargron (approved):

PR #3135, opened 2017-05-19
Make FeedInsertWorker to perform with followers in batches.

harukasan:
This PR is a feature request for performance.

FeedInsertWorker and PushUpdateWorker spawn for each followers when the toot is post.
It causes request N+1 queries by each toots.

This PR reduces the number of queries by processing followers in batches.
By default, the feature is disabled, and enabled by setting `FAN_OUT_BATCH_JOB_SIZE` to 2 or more.

In Pawoo, the number of Sidekiq jobs became 1/3 and CPU utilization decreased in the both RDB and Application (`FAN_OUT_BATCH_JOB_SIZE=4`).

How do you like it?

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -23,8 +23,8 @@ class FeedManager
 
   def push(timeline_type, accounts, status)
     accounts = Array.wrap(accounts)
-    accounts = accounts.reject do |account|
-      filter?(timeline_type, status, account.id) || !insert_and_check(timeline_type, status, account)
+    accounts = accounts.select do |account|
+      insert_and_check(timeline_type, status, account)
     end
 
     PushUpdateWorker.perform_async(accounts.map(&:id), status.id) unless accounts.empty?
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -21,20 +21,29 @@ class FeedManager
     end
   end
 
-  def push(timeline_type, account, status)
-    timeline_key = key(timeline_type, account.id)
+  def push(timeline_type, accounts, status)
+    accounts = Array.wrap(accounts)
+    accounts = accounts.reject do |account|
+      filter?(timeline_type, status, account.id) || !insert_and_check(timeline_type, status, account)
+    end
+
+    PushUpdateWorker.perform_async(accounts.map(&:id), status.id) unless accounts.empty?
+  end
+
+  def insert_and_check(type, status, account)
+    timeline_key = key(type, account.id)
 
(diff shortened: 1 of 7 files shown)


Gargron:
Originally there were no FeedInsertWorkers and everything happened in DistributeWorker synchronously. The reason why we changed it the other way around was that splitting up into smaller and faster workers made more efficient use of the connection pool (one job didn't use up a connection for a long time), this also meant that independent tasks weren't delayed needlessly. For example, rendering JSON for the streaming API takes a long time (~600ms), but it's comparatively less important than just adding the status ID into the recipient's feed (which allowed the recipient to see an updated home timeline on refresh even before streaming API is processed).

This has also been proven effective as a measure against the large backlog apocalypse on mastodon.social during the beginning of April - the queues clear way quicker.

cc @krainboltgreene

harukasan:
@Gargron
The purpose of this PR is reduce to queries and jobs caused by many followers of a toot.
The FeedInsertWorker and DistributeWorker is still separated.

harukasan:
FYI, our CPU metrics before/after apply the code

DB:
screen shot 2017-05-18 at 22 23 42

Application:
screen shot 2017-05-19 at 21 05 01

You can see that the bursts has decreased.

beatrix-bitrot:
the benchmark results are pretty impressive...

akihikodaki (approved):

Gargron:
>the benchmark results are pretty impressive...

But has throughput increased or decreased? Less CPU usage can also mean less work is being done. We should optimize for more throughput, not idler CPU.

harukasan:
I agree that the throughput is important. Because this PR aims to reduce SQL queries and Sidekiq's CPU instead of increasing the pushing latency, so it makes trade-off.
Huge concurrency causes overall latency.

I think `FAN_OUT_BATCH_JOB_SIZE` is not important for almost all, except for large instance like Pawoo.

harukasan:
So we need the benchmark that "sidekiq job overhead" vs. "pushing a feed to redis".
If we can stop using Sidekiq (or can optimize Sidekiq), it is better.

nightpool:
I think that throughput is not going to be meaningfully decreased by putting these in batches of 3 or 4, and moreover, larger sites are going to have to make a different tradeoff between throughput and performance no matter what. 

However, I do think queue starvation is a real problem, and we should be careful that making these jobs longer doesn't make it harder for other jobs to happen. Maybe we can resolve these problems by adding a new sidekiq queue for feed fanout?

either way, this needs conflicts resolved if we can consider a merge

PR #3136, opened 2017-05-19
 Introduce methods for bulk operations to FeedManager

akihikodaki:
# Description
This commit substitutes `filter?` of `FeedManager` with `filter_subscribers` and `filter_mentions`, which filter relations for ease of use and better performance.

It also removes `FeedInsertWorker`, whose main role was to call `filter?` for each insertions.

It removes `filter?` call from `NotifyService`.

`filter?` was a method to determine a status to be filtered or not. However, callers of `filter?` actually deal with relations and it is favorable to filter relations rather than requiring to call a method for each records in the relations. That could also benefit for performance, by reducing the linear query number to a constant number.

The main role of `FeedInsertWorker` was to call `filter?` for a status, and jobs for it was queued by `FanOutOnWriteService` for each statuses. Now `filter_subscribers` filters multiple statuses rather than a status, so remove `FeedInsertWorker` and move the feature to `FanOutOnWriteService`. In the case the synchronous filtering impacts performance, we need to
introduce another worker.

This commit removes `filter?` call from `NotifyService`. Now the callers filter mentions, which are in relations, by themselves.

# Result
The following is statements and explanations generated respectively for `#filter_subscribers does not filter followers of the account who reblogged` and `#filter_mentions filters accounts not following the author, who is silenced`.
```
> EXPLAIN SELECT "accounts".* FROM "accounts" WHERE "accounts"."id" = 1 AND ("accounts"."id" NOT IN (SELECT "blocks"."account_id" FROM "blocks" WHERE "blocks"."target_account_id" = 11741)) AND ("accounts"."id" NOT IN (SELECT "mutes"."account_id" FROM "mutes" WHERE "mutes"."target_account_id" IN (11742, 11741))) AND ("accounts"."id" NOT IN (SELECT "blocks"."target_account_id" FROM "blocks" WHERE "blocks"."account_id" = 11741)) AND ("accounts"."id" NOT IN (SELECT "account_domain_blocks"."account_id" FROM "account_domain_blocks" WHERE "account_domain_blocks"."domain" IS NULL));
                                                         QUERY PLAN                                                          
-----------------------------------------------------------------------------------------------------------------------------
 Index Scan using accounts_pkey on accounts  (cost=49.18..57.46 rows=1 width=1162)
   Index Cond: (id = 1)
   Filter: ((NOT (hashed SubPlan 1)) AND (NOT (hashed SubPlan 2)) AND (NOT (hashed SubPlan 3)) AND (NOT (hashed SubPlan 4)))
   SubPlan 1
     ->  Seq Scan on blocks  (cost=0.00..1.11 rows=1 width=4)
           Filter: (target_account_id = 11741)
   SubPlan 2
     ->  Seq Scan on mutes  (cost=0.00..27.50 rows=14 width=4)
           Filter: (target_account_id = ANY ('{11742,11741}'::integer[]))
   SubPlan 3
     ->  Seq Scan on blocks  (cost=0.00..1.11 rows=1 width=4)
           Filter: (account_id = 11741)
   SubPlan 4
     ->  Seq Scan on account_domain_blocks  (cost=0.00..19.40 rows=5 width=4)
           Filter: (domain IS NULL)
(15 rows)
```
```
> EXPLAIN SELECT 1 AS one FROM "mentions" INNER JOIN "accounts" ON "accounts"."id" = "mentions"."account_id" LEFT OUTER JOIN "follows" ON "follows"."account_id" = "accounts"."id" LEFT OUTER JOIN "accounts" "followings_accounts" ON "followings_accounts"."id" = "follows"."target_account_id" WHERE "mentions"."status_id" = 1 AND ("mentions"."account_id" != 2) AND ("mentions"."account_id" NOT IN (SELECT "blocks"."account_id" FROM "blocks" WHERE "blocks"."target_account_id" IN (11975, 11976))) AND "followings_accounts"."id" = 3;
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=1.12..25.45 rows=1 width=0)
   ->  Nested Loop  (cost=1.12..17.17 rows=1 width=4)
         ->  Nested Loop  (cost=0.00..15.48 rows=1 width=12)
               ->  Seq Scan on follows  (cost=0.00..7.20 rows=1 width=8)
                     Filter: (target_account_id = 3)
               ->  Index Only Scan using accounts_pkey on accounts  (cost=0.00..8.27 rows=1 width=4)
                     Index Cond: (id = follows.account_id)
         ->  Index Only Scan using index_mentions_on_account_id_and_status_id on mentions  (cost=1.12..1.68 rows=1 width=4)
               Index Cond: ((account_id = accounts.id) AND (status_id = 1))
               Filter: ((account_id <> 2) AND (NOT (hashed SubPlan 1)))
               SubPlan 1
                 ->  Seq Scan on blocks  (cost=0.00..1.11 rows=2 width=4)
                       Filter: (target_account_id = ANY ('{11975,11976}'::integer[]))
   ->  Index Only Scan using accounts_pkey on accounts followings_accounts  (cost=0.00..8.27 rows=1 width=4)
         Index Cond: (id = 3)
(15 rows)
```

# Discussion
It is fine for my small instance to filter synchronously in `FanOutOnWriteService`, but could it be problematic for other instances?

Now `filter_subscriptions` and `filter_mentions` deals with relations. Should they be ActiveRecord scopes?


Gargron:
Please mind that the old methods were easy to read, understand, and modify. This approach might be highly optimized but it will be much harder to adjust and a "here be dragons" territory for new developers (right now even I have trouble comprehending this code)

akihikodaki:
ActiveRecord scoping may help in the case. Now it has many joins, where, etc. However, it is not necessary to be in one method. They could be divided into several scopes.
Does it sound better?

akihikodaki:
c1d0568cf52399f305a7947d5b7646f9ecba3f13 has the following changes:
* Added scopes to `AccountInteractions`, and removed several `joins` and `where` from `filter_subscribers` and `filter_mentions`. Now it looks more similar to the current implementation.
* Updated specifications accordingly.
* Updated comments for better understanding.
* Written in standard SQL.
* Rebased to de475cf8d32744330f8029f13c539237a6567029.

akihikodaki:
Updated the above queries and `EXPLAIN` just in case anyone interested in this change.

akihikodaki:
c52bc3a455baa3973e2513007814fb7bb30f519d introduced `push_bulk` method and the whole pushing operation has been optimized.
In addition to the past changes, it removes `PushUpdateWorker` and introduced concurrent operations with Redis pipeline. In the point that it removes `PushUpdateWorker`, it is a solution similar to but more aggressive than #3135.

clworld:
What I think is:
- Process at once is good for database. (join / solve N+1 problem)
- Process at once is good for redis. (pipeline)
- Process at once is sometime relly bad for Sidekiq queue. (described below)

Big difference of task size of sidekiq jobs are relly bad for sidekiq queue.

Why?:
- Sidekiq does not have task scheduler. (It doesn't run slow task at low priority.)
- Sidekiq can't do preemptive multitask. (It doesn't pause execution of slow task when short task arrived.)

What happen?:
- All sidekiq threads are consumed by slow task.
- Sidekiq queue will be clogged.

So, size of each sidekiq tasks must be nearly same size. (size = execution time)
This is why FAN_OUT_BATCH_JOB_SIZE value which @harukasan recommends in #3135 is relly small. I think.
(It process only 4 accounts at once and PushUpdateWorker are still separated.)
(#3135 does also run PushUpdateWorker in batches.. It may be bad because PushUpdateWorker is relly slow.)

Changing job queue to which better than Sidekiq may solve this limitation, but I don't know that.

clworld:
I also wonder that do many things in redis.pipelined block is safe or not.
Because redis does not initialized with connection pool and operations to redis are single threaded. As I know.
I don't know redis.pipelined blocks operations from other thread or not.
If redis.pipelined does not block opetations from other thread, It's OK.

akihikodaki:
This new implementation does not even queue Sidekiq job, it pushes __synchronously__. I decided to do so because I thought each push is still small enough (according to my intuition.) Maybe that is not right depending on situations, so it may need to add a worker to push asynchronously.

Adding a worker to push asynchronously raises the question @clworld suggested. However I do not think that is not a serious problem because the worker would be frequently queued.
Some pushes (for example, status from an account followed by many others) may be heavy, but I'm expecting such deviance could be ignored when there are so many small queues. (This is based on my intuition as well. I cannot be sure what would happen in the real world till this change is deployed.)

> I also wonder that do many things in redis.pipelined block is safe or not.
> Because redis does not initialized with connection pool and operations to redis are single threaded. As I know.
> I don't know redis.pipelined blocks operations from other thread or not.
> If redis.pipelined does not block opetations from other thread, It's OK.

I think I need some benchmarks.

akihikodaki:
I had wrote a benchmark and here is the result. Note that this benchmark is quite a naive.
```
Rehearsal -------------------------------------------------------
get 100000 times      0.020000   0.010000   0.030000 (  0.020987)
zrevrank 1000 times   0.020000   0.000000   0.020000 (  0.020147)
zadd 1000 times       0.010000   0.000000   0.010000 (  0.014780)
publish 1000 times    0.010000   0.000000   0.010000 (  0.014181)
---------------------------------------------- total: 0.070000sec

                          user     system      total        real
get 100000 times      0.010000   0.010000   0.020000 (  0.015533)
zrevrank 1000 times   0.010000   0.000000   0.010000 (  0.013284)
zadd 1000 times       0.010000   0.000000   0.010000 (  0.015716)
publish 1000 times    0.020000   0.000000   0.020000 (  0.015273)
```
The number of `get` is large while the others are not because the number of `get` could be large depending on the number of followers while the number of other queries are limited because they are made only for subscribing accounts. In this case, it assumes the account has 100,000 followers and 1,000 of them are currently online.

`0.070000sec` is not bad, I think.

The benchmark is available at https://gist.github.com/akihikodaki/9abaacb98bdd1c1500f46ba16fbdfd73.

akihikodaki:
Edited the above comment and the benchmark; it was seriously wrong because it just `get` 1,000 times. I updated the result when it does `get` 10,000 times.

akihikodaki:
a8e47bdfa267c82f985615f7c6d338a7351543d1 uses `mget` instead of `get`. The benchmark has been also updated as well.

```
Rehearsal -------------------------------------------------------
mget 10000            0.070000   0.000000   0.070000 (  0.076775)
zrevrank 1000 times   0.010000   0.010000   0.020000 (  0.019425)
zadd 1000 times       0.020000   0.000000   0.020000 (  0.019957)
publish 1000 times    0.020000   0.000000   0.020000 (  0.013492)
---------------------------------------------- total: 0.130000sec

                          user     system      total        real
mget 10000            0.050000   0.000000   0.050000 (  0.059199)
zrevrank 1000 times   0.010000   0.000000   0.010000 (  0.012918)
zadd 1000 times       0.010000   0.010000   0.020000 (  0.013731)
publish 1000 times    0.010000   0.000000   0.010000 (  0.015680)
```

clworld:
Redis only benchmark is not fair I think.
There is ActiveRecord SQL execution and rabl rendering by InlineRenderer.
It seems rabl rendering of app/views/api/v1/statuses/show.rabl cause SQL execution. relly?

akihikodaki:
Queries for PostgreSQL are important, but I'm not really concerned about the performance because they had to run the following line or sort of in synchronous way. Such a query had to run before creating jobs.
```Ruby
status.account.followers.where(domain: nil).joins(:user).where('users.current_sign_in_at > ?', 14.days.ago).select(:id).reorder(nil)
```

akihikodaki:
e41e1197ac7d8cc29e659cbb21f623e450e306fb revived `PushUpdateWorker` because `InlineRenderer` takes long.
I used a new benchmark which tests the real `FeedManager.instance.push_bulk`.
https://github.com/kagucho/mastodon/tree/benchmark

```
   (3.1ms)  SELECT COUNT(*) FROM "accounts"
10063
  Account Load (12.1ms)  SELECT  "accounts".* FROM "accounts" LIMIT $1  [["LIMIT", 1000]]
       user     system      total        real
push_bulk   (3.4ms)  SELECT "accounts"."id" FROM "accounts"
  0.460000   0.040000   0.500000 (  0.504183)
```
(`10063` means it had pushed to 10,063 accounts. 1,000 of them were subscribing.)

clworld:
0.5 sec is not fatal, but does not acceptable I think.
Even if ProcessingWorker in my instance takes 1.3 sec sometime...

akihikodaki:
I wonder even the current implementation could deal with it in less than 0.5 seconds. I need to make a benchmark to confirm that, but that would be quite a difficult. It also depends on various configuration.

akihikodaki:
> Even if ProcessingWorker in my instance takes 1.3 sec sometime...

In case you concern if it may take longer in other instances, I assure that the performance the new implementation does not depend so much on the configuration or other contexts as the current one does. The method mostly consists of Redis operations whose number depends only on the number of pushes to be processed. The performance of Redis operations are not greatly affected by the number of keys (users.)

PR #3138, opened 2017-05-19
Unify the method of extracting tags

abcang:
The tag extracting method is different between tag link generation and tag generation.
For example, in the case of the tag `#テスト・テスト`, it was saved in the database as `#テスト`.
So you may not be able to search tags with tag links.

This problem can be solved by unifying the method of extracting tags.
I changed to use Mastodon's regular expression.

### DIFF
--- a/app/lib/extractor.rb
+++ b/app/lib/extractor.rb
@@ -30,4 +30,30 @@ module Extractor
     end
     possible_entries
   end
+
+  def extract_hashtags_with_indices(text, _options = {})
+    return [] unless text =~ /#/
+
+    tags = []
+    text.scan(Tag::HASHTAG_RE) do |hash_text, _|
+      match_data = $LAST_MATCH_INFO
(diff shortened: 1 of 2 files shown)


app/models/tag.rb
```diff
@@ -12,8 +12,6 @@
 class Tag < ApplicationRecord
   has_and_belongs_to_many :statuses
 
-  HASHTAG_RE = /(?:^|[^\/\)\w])#([[:word:]_]*[[:alpha:]_][[:word:]_]*)/i
```
Gargron: Does twitter-text actually support utf8 hashtags? I feel uncomfortable dropping our own rules here

abcang: I'm sorry. I will try to change the tag link generation process to use the regular expression of the mastodon.

abcang:
Updated code and description. Thank you.

Gargron (approved):

PR #3148, opened 2017-05-19
CAS + SAML authentication feature

gloaec:
This is supposed to an non-intrusive approach of CAS authentication (using an oauth provider). I think i made it suitable production settings, but i'm not so sure about that since I am no expert with Docker. I added 2 more translatable strings and provided the translation for both french and english. Let me know if I did something wrong or if you have suggestions for improvements.

Cheers !


app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,31 @@
+class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+
+  def self.provides_callback_for(provider)
+    _provider = provider.to_s.chomp "_oauth2"
+    class_eval %Q{
```
Gargron: Why this, can we do it any other way?

gloaec: Yes, we could indeed do it without the scary `class_eval`. It's just that the callback/process is pretty much the same for any provider you would decide to add (or not) in the future.

Gargron:
I like this, but it'll have to wait for after v1.4, since now we want to stabilize for a release instead of adding new features.

gloaec:
No problem, I understand. I guess the Ministry of Education can use the fork until the next stable release. And btw, your approach for LDAP authentication is, on second thought, better than implementing an oauth provider. With oauth, I will be difficult to allow log in through the current form. Since it would be a middleware, it would have come with its own login form (which is kind of unlogical according to the LDAP paradigm). So yeah, it's definitely better to implement a devise strategy, as you did ;)
Anyway, thanks for taking the time to review.

.env.production.sample
```diff
@@ -52,6 +52,28 @@ SMTP_FROM_ADDRESS=notifications@example.com
...
+# CAS_DISABLE_SSL_VERIFICATION=false
+# CAS_UID_KEY='user'
+# CAS_NAME_KEY='name'
+# CAS_EMAIL_KEY='email'
+# CAS_NICKNAME_KEY='nickname'
+# CAS_FIRST_NAME_KEY='firstname'
+# CAS_LAST_NAME_KEY='lastname'
+# CAS_LOCATION_KEY='location'
+# CAS_IMAGE_KEY='image'
+# CAS_PHONE_KEY='phone'
```
nightpool: This should probably be a rails initializer file instead of environment options, since it's pretty niche.

app/controllers/auth/omniauth_callbacks_controller.rb
nightpool: Can't we just do this with class_def and not try to use string interpolation?

.env.production.sample
Gargron: I believe ENV variables is absolutely fine. However, not all ENV variables need to be in the sample .env file. It is easier to configure ENV variables than to create initializers without code conflicts.

nightpool: I'm fine with having these as env variables, it just seems like with adding new authentication strategies, we're going to get an explosion of configuration variables, and we should think about some way of splitting them up into more meaningful chunks. (maybe have a .env.cas.sample file? or a snippets folder with example env files?)

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        sign_in_and_redirect @user, event: :authentication
+        set_flash_message(:notice, :success, kind: provider_id.capitalize) if is_navigational_format?
+      else
+        session["devise.#{provider}_data"] = env['omniauth.auth']
+        redirect_to new_user_registration_url
+      end
+    end
+  end
+
+  [:ldap, :cas].each do |provider|
```
immae: As said in https://github.com/tootsuite/mastodon/pull/4032#issuecomment-312914005 LDAP doesn't suit in this process

app/models/concerns/omniauthable.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    def find_or_create_user_for_oauth(auth)
+      # Get the existing user by email if the provider gives us a verified email.
+      # If no verified email was provided we assign a temporary email and ask the
+      # user to verify it on the next step via UsersController.finish_signup
+      email_is_verified = auth.info.email && (auth.info.verified || auth.info.verified_email)
+      email = auth.info.email if email_is_verified
+      uid = auth.uid
+      user = Account.find_by(username: uid).try(:user) if uid
```
immae: I think you're trusting too much the provider here:
if the provider is controlled by the user (in case of openID) for instance, it may have any uid it wishes, and thus login as anyone by just pretending to be that user.

Facebook allows you to change your login (?), so it could happen too even in Oauth.

More globally, you shouldn't trust informations coming from an external provider: either the user is created with the Oauth provider and we know it - because he's linked to that Identity, or he cannot magically link to an already existing user.

Note: With such a precaution in mind, do we really need an Identity model? I don't see a case where you could allow safely to have more than one identity belonging to a user

immae: (with that note, the provider and uid could very well simply belong to the user as well)

gloaec: I understand your point. In my opinion, it is the responsibility of the admins to trust or not a specific provider. And the uid can be anything you decide from the provider response :

    cas_options[:uid_field] = 'user'

It's not really a problem with CAS or SAML (for the usecases i can think of) since the user profile cannot be managed by the user himself. 
But for public providers like Facebook, I think you're right, it can be a problem. I like the fact to be able to get the same identity from different providers. But this requires an `Identity` model and the `find_by_uid` mechanism, I cannot think of any other way to do it :/ I'm opened to suggestions.
Another approach would be to implement different callbacks later on, when you decide to implement Facebook oauth or else, with double verification "uid" + "email" ? (cannot be done for CAS since the "email" is not always present)
@Gargron ? @nightpool ? What do you guys think ?

gloaec: So, from what I read, OmniAuth already advocate an approach to reconcile two identities :

1. An account shall be created for each provider the user logs in with
2. Each account has a secret token, which is visible in the private user profile
3. In order to merge 2 accounts, the user must log in with one provider, and then manually associate another account by filling this secret token from the other profile.

So, there's still a need for an `Identity` model in order to conciliate 2 accounts, but there's a lot of work to do on profile management. It's not the easiest approach, but it is the Omniauth recommandations.
Would that be suitable for you ?

immae: That would be a way to make sure that we don't have the security flaw stated above: reconciling two identities is not automatic and you need to prove that you owe both before that.

immae: And it could definitely benefit to ldap (and other means) too, so I'm definitely :+1:
But it's a lot of work...

Gargron: The way I understand it, and have done in other projects:

There is only one user+account, normally it is created with email/password, but when using OmniAuth, those are not important. So when you create an user+account with OmniAuth, you want to record OmniAuth-specific data (ldap dn, or facebook user ID, or whatever else). when you are already signed in, and use OmniAuth *again*, it shouldn't create a new account, it should connecting the new provider with the same user+account. So then for example, I am Gargron@mastodon.social, and I can login using GitHub OAuth and Google OAuth. This is usually how it's done. That's what the Identity (or Provider, or IdentityProvider) model should be for.

Many OmniAuth apps also allow you to disconnect a provider again. That's possibly with an IdentityModel but not if the information is stored on the User model.

gloaec: @Gargron, you'd suggest that there should be an automatic identity reconciliation mechanism based on email ? Only if email is verified by both providers I guess ?

If so, I would be much easier indeed. We can just forget about manual reconciliation for now. We'll worry about it when we actually implement some oauth strategy that cannot give us the user's email.

immae: @gloaec: I think what @Gargron meant was that you can only add an identity to an existing user if you add it while being already logged in as that user, which is actually a good way to avoid the flaw I talked about, but requires to add a new form "add new identity" available when you're logged in.

gloaec: @immae ok i see. I'm not familiar with such mechanism, but it doesn't seem to be too painful to implement. So now the question is :

The most common use scenario for CAS or SAML authentication is for organizations that already have an LDAP directory somewhere and just want to "plug" Mastodon to their existing SSO system. To do so, they will most of the time disable Mastodon database registration. Furthermore, they will use CAS or SAML, not both. So, is there really a need for reconciliation feature at this point ? Couldn't we just open an new "enhancement" issue for now and take care of that when we'll actually implement some oauth strategy that requires this feature (Facebook, Twitter, Google, Github...) ?

immae: @gloaec:
Sure, the only thing I’d like is that you remove that security flaw: in case of creating an identity, don’t link it to an already-existing user. In that case, fail (plugging an identity to an existing user should not happen here anyway)

Other than that, it’s a very fine PR, and I’ll rebase my LDAP one to use the identity model in preparation for what @Gargron talked about (multiple identiifcation ways for a given user) and is actually another thing.

gloaec:
I removed the email/uid reconciliation for now as agreed. So if this is ok, I'll rebase my [feature-saml](https://github.com/eole/mastodon/tree/feature-saml) branch on this one and make another pull-request ? Or else I can also merge everything here ?

immae:
@gloaec : Thanks!
you can simply push force to the same branch as initially and it will update the pull request

Some opinion about your solution though. it's not a requirement, you can decide what you want to do: As a user, if I subscribe the "normal" way to mastodon, I try "immae", and it get refused because someone already has this name. Ok, I find something else.
In your solution, you try immae_1 etc until you find something that doesn't exists yet. Given that it's a public name (you'll tell everyone "I'm immae_1@mastodon.social"), it's quite a hard choice to make in place of the user, no?
I know it shouldn't happen when the administrator chose to integrate CAS / whatever, since usually you either not have it or you have only that, and collision should not happen.
But I still find this choice strange.

Up to you to decide what you do of this thoughts, other than that it's a great job

gloaec:
@immae, I don't know how to deal with username collision for now. I guess it would involve some extra steps so the user can choose the username he wants. Unfortunately, I don't have much time to work on this anymore. I leave it as it is for now, since it shouldn't happen using CAS or SAML providers.

As discussed, I added the SAML authentication to this PR, since it uses the same classes and methods. Thanks for taking the time to review, for your pertinents remarks, and let me know if there's anything more I can do in order to see this PR merged.

Cheers !

immae (approved):
Not fond of the way of handling usernames for new users, but can be dealt with later, the rest is very good

immae:
@gloaec I added a review, but I'm not an active developper of mastodon, you'll have to find another person to accept your changes

NB: there are new conflicts

jcgruenhage:
A little thought on what to do if the username is taken already: Maybe just show some error message and don't log the user in at all? If this happens, it is probably not intentional, and something the admin of the instance want's to know about, so throwing an error (and sending him an email) probably makes sense.

evaryont:
I'd love to deploy Mastodon, but this (or the LDAP support, either one) is blocking for me.

In the situation where the administrator has allowed registration through Mastodon and has SAML (or even LDAP) support enabled,and there is a username conflict, I would expect an error to be thrown and nothing else happens. I, nor my users I expect, to have their username changed.

It's be up to the admin to fix the situation. As separate PR could add merging/migrating accounts from one type or another.

Gargron:
I have merged PAM authentication (#5303). PAM is supposed to handle integrations with LDAP and CAS. Do you think that is enough and this PR is no longer necessary?

gloaec:
@Gargron This PR is necessary and people are dying to see this feature released. PAM is again another authentication method that is just not comparable to CAS nor SAML. PAM/CAS authentication exists, but it's way less generic that the actual CAS approach. CAS and SAML provide their own authentication proxy (login form / logout action / service provider / ticket validation methods ). It is the most common/wanted SSO system today. Please don't give up on this.

The french ministry of education is giving me 2 more days to fix whatever is wrong with this PR. Just tell me what to do ?

Gargron (approved):

Gargron:
Okay, I'm approving this. The conflicts just need to be resolved. And it would be nice to make the extra dependencies optional in bundler, like in #6415. I realize you're out of time and that's okay, I'll pick it up I think.

Gargron:
Merged via #6425

gloaec:
Thank you, your commitment to the community is very appreciated.

PR #3150, opened 2017-05-19
Extract authorization policy for viewing statuses

jackjennings:
This pull request begins the process of centralizing the code for the authorization of status-related actions. This builds on the work done in #2982.

There are currently a few methods in `Status` that handle authorizing single statuses (`permitted?`) and filtering collections (`filter_from_context?`, `permitted_for`). Each of these appear to do the same thing, but are slightly different. Over time, they are bound to deviate and introduce unexpected privacy-related bugs. In addition, it is problematic that the `Status` class has a deep understanding of the implementation of `Account`.

Ideally, these methods are extracted into a collaborating object that performs authorization with the knowledge of the `Status` and `Account` interface.

I have added the [pundit](https://github.com/elabs/pundit) gem as a dependency, which encodes a simple, object-oriented pattern for authorization. This PR could be resubmitted without this dependency—though the pundit library is only ~300 LoC and stable.

A sample policy would be defined with this interface:

```ruby
class StatusPolicy
  attr_reader :account, :status

  def initialize(account, status)
    @account = account
    @status = status
  end

  def show?
    # return boolean indicated if status can be shown
  end
end
```

I have wrapped the pundit functionality in a new `Authorization` concern and exposes a public API of two methods: `authorize` and `authorize_with`. Both of these methods raise the (existing) `Mastodon::NotPermittedError` when the authorization check fails.

```ruby
# Authorizes against current_account
authorize @status, :show?

# Authorizes against supplied account:
authorize_with @account, @status, :show?
```

This also begins the process of standardizing this error as the sole error raised for authorization-related exceptions (rather than leaking the internal implementation of Pundit).

Outstanding questions:
* Should Services perform authorization? My preference would be to move all authorization checks into the controller level, as to not bloat the purpose of the Service. This is not how a few Services are currently implemented, but required me to add the `authorize_with` method in order to accommodate leaving this authorization check in place.
* Should the new `Authorization` mixin be applied to the `ApplicationController`? I have included it only in controllers and services that perform status authorizations at this point, as a conservative starting point.
* Several tests dictates that a `404` status code should be raised for certain un-permitted actions. I'm unsure if this is by design or an inconsistency that should be fixed. In several places I have re-raised a `ActiveRecord::RecordNotFound` exception in order to preserve the functionality under test.


jackjennings:
I'm submitting this for review to gauge interest in this solution, and to the points mentioned above. If this is something that could be merged, I would want to add a few controller tests where appropriate and document the new authorization concern.

Gargron:
I am in favour of this change but let's wait until after v1.4, such changes are prone to backfire in unexpected ways.

jackjennings:
Sounds good. I would be happy to revise to the above points in the OP if a reviewer can comment for acceptance post-1.4. My intention was to keep this change limited to both the scope of replacing one method and preserving existing functionality, but I can modulate if this should include slightly more or slightly less.

Regardless, I will backfire some specs for the controllers that this touches for better certainty that there are no regressions.

Gargron (approved):

PR #3152, opened 2017-05-19
Fix #2680 - Run processes in Docker as non-root user

### LINKED ISSUE #2680: Mastodon workers running as root in Docker containers
The current Dockerfile and docker-compose setup makes the Mastodon workers run as root in containers.

This is ill-advised [1] and unnecessary [2]. Running as root reduced the effort required to escape from the container to the host of the machine.

[1] - http://blog.dscpl.com.au/2015/12/don-run-as-root-inside-of-docker.html
[2] - https://docs.docker.com/engine/security/security/

* * * *

Gargron:
This might create issues for existing users where files in volumes are owned by the root user and cannot be used by the non-root one.

cc @Wonderfall

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,10 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
  && update-ca-certificates \
  && rm -rf /tmp/* /var/cache/apk/*
 
+RUN adduser --disabled-login -gecos 'Mastodon' mastodon
+
+USER mastodon
+
 COPY Gemfile Gemfile.lock package.json yarn.lock /mastodon/
 
 RUN bundle install --deployment --without test development \


Wonderfall:
Like I said on Mastodon, I'm not a fan, but the decision is yours. :)
In my opinion, UID/GID should be able to be easily configured. So you can add build-time variables (`ARG` with Docker), like I'm doing actually. But it can be even better with an entrypoint script IMO.

Gargron:
Feel free to submit a better PR

Wonderfall:
In case #3159 is not merged, the usage of adduser here is incorrect (we're using the busybox version). The correct version would be : 

```
adduser -g 'Mastodon' -h /mastodon -D -H mastodon
```

And something else in the approach of #3159 :
```
FROM ...

ARG UID=100

...

RUN adduser -g 'Mastodon' -h /mastodon -u $UID -D -H mastodon

PR #3156, opened 2017-05-19
Enable webm tests in API media controller

jackjennings:
These tests were pending, but it was unclear to me why. The git blame suggests this might have been a mistake, as the enabled tests pass for me.

### DIFF
--- a/spec/controllers/api/v1/media_controller_spec.rb
+++ b/spec/controllers/api/v1/media_controller_spec.rb
@@ -60,19 +60,19 @@ RSpec.describe Api::V1::MediaController, type: :controller do
         post :create, params: { file: fixture_file_upload('files/attachment.webm', 'video/webm') }
       end
 
-      xit 'returns http success' do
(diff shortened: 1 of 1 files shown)


Gargron:
They were disabled because they take super long comparatively, because webm files are large

jackjennings:
Should a shorter webm clip be used as a fixture?

jackjennings:
Or `MediaAttachment` could be stubbed.

mjankowski:
We should stub out all the paperclip/imagemagick interactions, so that `identify` and `convert` are not actually executed in specs. Should be sufficient to speed up.

jackjennings:
Sounds potentially dangerous to do globally—maybe some kind of opt-in per spec? I'm not familiar enough with RSpec to know what the options might be or if there is a standard way to do so.

Gargron:
Tests run for a long time already, don't want to make them run even longer. Sorry.

PR #3158, opened 2017-05-19
feat: Use CSS contain to avoid computations

sorin-davidoi:
We use the [`contain`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) to avoid repaints. [Browser support](http://caniuse.com/#search=contain) is limited to Chrome and Opera (hopefully Edge at some point).

To see what this does, enable Paint Flashing in the Chrome Dev Tools and perform the following on the latest `master`:
- Toggle the Column settings for the Home column - observe that it triggers repaints of the whole page
- Wait for the timer that fetches notifications to fire - two observations:
  - it triggers repaint of elements outside the notification column
  - if there are no notifications, it swaps the "empty view" with the "list view" for a moment (while `isLoading` is true)

This PR limits repaints to the column that triggered them and avoids the fast swapping of the view in the notification column (instead of displaying the "list view" if `isLoading`, we keep the same view that is currently displayed).

### DIFF
--- a/app/javascript/mastodon/features/notifications/index.js
+++ b/app/javascript/mastodon/features/notifications/index.js
@@ -100,7 +100,9 @@ class Notifications extends React.PureComponent {
       unread = <div className='notifications__unread-indicator' />;
     }
 
-    if (isLoading || notifications.size > 0) {
+    if (isLoading && this.scrollableArea) {
+      scrollableArea = this.scrollableArea;
+    } else if (notifications.size > 0) {
       scrollableArea = (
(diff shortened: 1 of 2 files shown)


ykzts (approved):

nolanlawson:
FWIW we (Edge) don't have any [immediate plans to implement](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/csscontainment/) but Firefox has already started [a partial implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1150081#c17). In any case it's harmless for browsers that don't support it. :smiley:

nolanlawson (approved):

PR #3159, opened 2017-05-20
Fix #2680 - Run processes in Docker as non-root user (alternative)

### LINKED ISSUE #2680: Mastodon workers running as root in Docker containers

Wonderfall:
Here's an alternative to https://github.com/tootsuite/mastodon/pull/3152.
First, #3152 won't work because it's not using correctly busybox adduser (`--disabled-login` doesn't exist, for example). And another problem is that with #3152, we'll always create a user with the same IDs (`1000`).

This PR was made with flexibility in mind : `UID` and `GID` should be easily set if needed (this can be done using environment variables, defaults are 991). It's also better not to create the user during the build process.

**Suggestion** : do you want to update `docker-compose.yml` or `.env.production`  to show how to configure these environment variables? This is not necessary, but I think it's better for visibility.

That being said, I couldn't make it more "transparent" than https://github.com/tootsuite/mastodon/pull/3152 : this means actual administrators coming from a previous release will have to `chown UID:GID /path/to/public/system`. Otherwise, even with optimisations, this will take too much (and useless) time before the process is run. No problem for new installations though, but this PR will require a release note.

### DIFF
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon
+find /mastodon -path /mastodon/public/system -prune -o -not -user mastodon -not -group mastodon -print0 | xargs -0 chown -f mastodon:mastodon
+su-exec mastodon:mastodon /sbin/tini -- "$@"
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,8 @@ FROM ruby:2.4.1-alpine
 LABEL maintainer="https://github.com/tootsuite/mastodon" \
       description="A GNU Social-compatible microblogging server"
 
-ENV RAILS_ENV=production \
-    NODE_ENV=production
+ENV UID=991 GID=991 \
+    RAILS_ENV=production NODE_ENV=production
 
 EXPOSE 3000 4000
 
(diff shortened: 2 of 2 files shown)


Gargron (approved):

docker_entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon
+find /mastodon -path /mastodon/public/system -prune -o -not -user mastodon -not -group mastodon -print0 | xargs -0 chown -f mastodon:mastodon
```
maur1th: This might have a huge performance impact and is pretty much in contradiction with Docker's philosophy of "start fast". We can already see the effects: https://github.com/tootsuite/mastodon/issues/3194

Seems like a bad tradeoff only to be able to set UID/GID.

If someone really wants to use a specific UID/GID, maybe use build args with defaults instead. So they can rebuild the image with whatever UID/GID they want without hampering the container boot time.

ykzts: > This might have a huge performance impact and is pretty much in contradiction with Docker's philosophy of "start fast". We can already see the effects: #3194
> 
> Seems like a bad tradeoff only to be able to set UID/GID.
> 
> If someone really wants to use a specific UID/GID, maybe use build args with defaults instead. So they can rebuild the image with whatever UID/GID they want without hampering the container boot time.

e

Gargron: @Wonderfall I just had this idea: How about checking permissions on *one* file, and only running the update if they are wrong? There's like 0 chance that permissions will be mixed, so if one file's good, the others are too. This will make startup as fast as before in most cases.

PR #3167, opened 2017-05-20
Use joins for account properties

akihikodaki:


Gargron:
Is this a performance improvement?

akihikodaki:
Yes. And it also reduces code complexity (removal of `.map { |a| [a.id, a] }.to_h`, etc)

Gargron (approved):

PR #3168, opened 2017-05-20
Performance improvements

sorin-davidoi:
- Invert the conditions in `status_list` `didComponentUpdate` - since computing `scrollTop` is the most expensive operation, we avoid 2 forced reflows and save `~40ms` (in the development build) on the first render
screen shot 2017-05-20 at 11 30 35

- Reduce node count in the status component by not rendering the dropdown items if the dropdown is not expanded (reduces number of nodes by `~10%`, from 1126 to 1006 - measured on `/web/getting-started` with no notifications and initial number of toots loaded). Also reduces the number of listeners, since each dropdown item has a `onClick` event.
- Same thing for `privacy_dropdown` - node count 1006 -> 986
- Same thing for `column_collapsable` - node count 986 -> 754 (the settings were being rendered even if they were collapsed)
- Cherry-pick imports from `react-motion` (reduced `application.js` by `3kB`)

This adds up to `~33%` less elements in the `DOM`.


Gargron (approved):

app/javascript/mastodon/components/dropdown_menu.js
```diff
@@ -43,6 +44,10 @@ class DropdownMenu extends React.PureComponent {
     this.dropdown.hide();
   }
 
+  handleShow = () => this.setState({ expanded: true })
+
+  handleHide = () => this.setState({ expanded: false })
+
```
nolanlawson: That's neat; I didn't know you could skip the explicit `.bind()` by using arrow functions here, but apparently [you can](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&code=class%20Foo%20%7B%0A%20stringify%20%3D%20()%20%3D%3E%20this.toString()%0A%7D). Very cool!

app/javascript/mastodon/components/collapsable.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Motion, spring } from 'react-motion';
+import Motion from 'react-motion/lib/Motion';
+import spring from 'react-motion/lib/spring';
```
nolanlawson: This doesn't appear to be part of [the public API per the docs](https://github.com/chenglou/react-motion) but if it reduces the bundle size, I'm in. :) It may need to be reverted if they release a breaking change, though.

spec/javascript/components/dropdown_menu.test.jsx
```diff
@@ -38,7 +38,40 @@ describe('<DropdownMenu />', () => {
...
+    wrapper.find(DropdownTrigger).first().simulate('click');
+    expect(wrapper.find(DropdownMenu).first().state('expanded')).to.be.equal(true);
+
+    wrapper.find('span').first().simulate('click');
+    expect(wrapper.find(DropdownMenu).first().state('expanded')).to.be.equal(false);
+  })*/
+
+  it('renders list elements for each props.items if expanded', () => {
+    const wrapper = mount(<DropdownMenu icon={icon} items={items} size={size} />);
+    wrapper.find(DropdownTrigger).first().simulate('click');
```
nolanlawson: Nice job adding tests!

sorin-davidoi: Had to, since I broke the existing ones 😄

PR #3170, opened 2017-05-20
Add webpack-bundle-analyzer for production builds

nolanlawson:
I've found `webpack-bundle-analyzer` to be an incredibly useful tool to analyze the bundle size and figure out what's adding bloat and which modules are in which chunk. This PR hooks it into the production builds, so it generates `report.html` and `stats.json`.

The `report.html` looks like this, and can even be served from `example.com/packs/report.html`:

screenshot 2017-05-20 07 13 44

The `stats.json` is a more generic output and can be used in tools like [webpack-analyse](https://webpack.github.io/analyse) to see things like the module dependency graph:

screenshot 2017-05-20 07 13 36

I was worried this might significantly slow down the precompile step, so I tested it (using `RAILS_ENV=production time bundle exec rails assets:precompile`) and surprisingly I found that (on my Macbook Air anyway) it took 122.30s before this PR and 104.68s after.  I doubt this PR actually makes precompile consistently faster (since that makes no sense), but in any case it doesn't seem to make it slower.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -6,6 +6,7 @@ const webpack = require('webpack');
 const merge = require('webpack-merge');
 const CompressionPlugin = require('compression-webpack-plugin');
 const sharedConfig = require('./shared.js');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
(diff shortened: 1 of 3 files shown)


nolanlawson:
rebased

Gargron (approved):

PR #3177, opened 2017-05-20
Catalan translation update

ghost:
Added config/locales/activerecord.ca.yml


config/locales/activerecord.ca.yml
```diff
@@ -0,0 +1,12 @@
+en:
```
Gargron: Wrong key, must be ca

Gargron (changes requested):

Gargron (approved):

PR #3182, opened 2017-05-20
Some Docker improvements

Wonderfall:
- improve docker_entrypoint.sh
- serve static files with puma by default (so it won't break for people using Docker Hub)
- sort packages list
- use virtual package for build deps
- show how to assign UID/GID

### DIFF
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -1,4 +1,14 @@
 #!/bin/sh
+
+### 1. Adds local user (UID and GID are provided from environment variables).
+### 2. Updates permissions, except for ./public/system (should be chown on previous installations).
+### 3. Executes the command as that user.
+
+echo "Creating mastodon user (UID : ${UID} and GID : ${GID})..."
 addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon
+
+echo "Updating permissions..."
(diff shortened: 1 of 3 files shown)


docker_entrypoint.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/sh
+
+### 1. Adds local user (UID and GID are provided from environment variables).
+### 2. Updates permissions, except for ./public/system (should be chown on previous installations).
```
Gargron: Is there any reason why you can't chown that here, too?

Wonderfall: It would take way too long even with my optimised command. I can try with GNU parallel but this seems a bit hacky IMO... 🤔

Wonderfall:
Hmm please wait a bit before merging, I got an error with `redis-rack` while downloading.
That said it's not due to these changes (I'm 99% it's working but I prefer to test before).

docker_entrypoint.sh
Gargron: Okay, makes sense.

Gargron:
@Wonderfall Okay. I think this is the last bit before 1.4 though, so the sooner this is ready the sooner v1.4rc1 can be tagged

Wonderfall:
It can be merged now, it works.

Gargron (approved):

PR #3186, opened 2017-05-20
Open the media content in a modal window

pfm-eyesightjp:
Open the media content(image/movie) in a modal window instead of in a new window.

modal

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -92,16 +92,14 @@ delegate(document, '.u-video', 'click', ({ target }) => {
     video.controls  = true;
   } else if (target.getAttribute('src')) {
     video.src = target.getAttribute('src');
-  } else if (target.getElementsByTagName('video')) {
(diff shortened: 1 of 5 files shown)


beatrix-bitrot:
Hmmm, I thought it already opens in a modal (on click)

When does media open in a new window?

pfm-eyesightjp:
Hello.

Timeline in /web/getting-started is modal window.

Timeline in /@account is not modal window.
-> Change to modal window.

beatrix-bitrot:
Ah thank you! That makes sense.

Gargron:
This has become pretty outdated. Basically, I prefer not to duplicate JS logic between React and non-React. Recently I moved the images rendering logic from being in both Haml and React to just being in React, mounted from a static element. The correct way to make modals work on static pages would be to find a way to mount the ModalContainer as well, and for them to use the same store so the eventing works. But this is complicated and not critical.

Issue #2900, opened 2017-05-08
List view keeps adding more DOM nodes the more you scroll


nolanlawson:
As you scroll down the main list of toots, every time you reach the bottom it adds more DOM nodes, on the order of around ~500 it seems. (For comparison, Gmail has ~3000 total elements on first load on desktop.)

This quickly exhausts the available memory. You really notice it on mobile devices, where scrolling to the bottom only 5 or 6 times may cause the webapp to become sluggish or crash.

![screenshot 2017-05-07 20 03 59](https://cloud.githubusercontent.com/assets/283842/25788767/5c544dc6-3360-11e7-8d04-6ee1aec3e79d.png)

Probably the best solution would be a virtual list, i.e. a list that only renders those elements that are currently visible and then swaps new ones in and out as necessary. It seems the [InfiniteLoader component](https://github.com/bvaughn/react-virtualized/blob/master/docs/creatingAnInfiniteLoadingList.md) in `react-virtualized` would be a good candidate, especially since it's already being used by `emojione-picker` and so it's one less dependency.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Concerns:

* Dynamic row height
* Connecting clickable ColumnHeader to the underlying List for scrollToTop animations

nolanlawson:
It will also break Ctrl-F but I feel that's a worthy tradeoff.

-----
Nolan Lawson
http://nolanlawson.com
https://github.com/nolanlawson

nolanlawson:
Another option is to simply clear out each list item's content when it's no longer in view, using a debounced scroll listener. I.e. you wait for the element to be invisible, then you replace it with an empty DOM node with the exact same width/height. It's not as good as a virtual list, but it reduces the total number of DOM nodes at least. Unfortunately this will probably mess up React's virtual DOM diffing so I'm not sure exactly how to achieve it.

fwenzel:
As for breaking DOM diffing, I suppose when it's out of view, one could flip a boolean on the list item component state and have it rerender itself as an empty box (and vice versa when it returns). This way React would be fully aware of what's happening.

nolanlawson:
[...]

PR #3191, opened 2017-05-20
Lazy load toots using IntersectionObserver

sorin-davidoi:
An attempt at fixing #2900 using  [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) instead of scroll listeners and libraries.

Workflow:
 - Render all the toots and save their height
 - Use `IntersectionObserver` to compute which toots are in the viewport. Toots that are outside get replaced with a `div` containing the author and status content (in order for browser search to continue working). The `div` has the original height, maintaining the list scrolling.
 - Render toots that are visible in the scrolling container plus a few more before and after. This avoids displaying the empty `div` (which has `opacity: 0`).

In case swapping between the actual toot and the placeholder `div` is not fast enough, I have  added a fade-in effect to make the loading appear smoother. You may see this on mobile or if you are scrolling fast.

[Browser support](http://caniuse.com/#feat=intersectionobserver): Chrome and Edge 15. Firefox and WebKit are currently implementing it - the [official WICG polyfill](https://github.com/WICG/IntersectionObserver/tree/gh-pages/polyfill) is used meanwhile.

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -26,6 +26,12 @@ class StatusList extends ImmutablePureComponent {
     trackScroll: true,
   };
 
+  state = {
+    isIntersecting: { },
+  }
+
+  statusRefQueue = []
+
   handleScroll = (e) => {
     const { scrollTop, scrollHeight, clientHeight } = e.target;
     const offset = scrollHeight - scrollTop - clientHeight;
@@ -42,6 +48,7 @@ class StatusList extends ImmutablePureComponent {
 
   componentDidMount () {
     this.attachScrollListener();
+    this.attachIntersectionObserver();
   }
 
   componentDidUpdate (prevProps) {
@@ -52,6 +59,39 @@ class StatusList extends ImmutablePureComponent {
 
   componentWillUnmount () {
     this.detachScrollListener();
+    this.detachIntersectionObserver();
+  }
+
+  attachIntersectionObserver () {
+    const onIntersection = (entries) => {
+      const isIntersecting = { };
+
+      entries.forEach(entry => {
+        const statusId = entry.target.getAttribute('data-id');
+
+        isIntersecting[statusId] = entry.isIntersecting;
+      });
+
+      this.setState(state => ({
+        isIntersecting: Object.assign({ }, state.isIntersecting, isIntersecting),
+      }));
+    };
+
+    const options = {
+      root: this.node,
+      rootMargin: '100% 0px',
+    };
+
+    this.intersectionObserver = new IntersectionObserver(onIntersection, options);
+
+    if (this.statusRefQueue.length) {
+      this.statusRefQueue.forEach(node => this.intersectionObserver.observe(node));
+      this.statusRefQueue = [];
+    }
+  }
+
+  detachIntersectionObserver () {
+    this.intersectionObserver.disconnect();
   }
 
   attachScrollListener () {
@@ -66,6 +106,15 @@ class StatusList extends ImmutablePureComponent {
     this.node = c;
   }
 
+  handleStatusRef = (node) => {
+    if (node && this.intersectionObserver) {
+      const statusId = node.getAttribute('data-id');
+      this.intersectionObserver.observe(node);
+    } else {
+      this.statusRefQueue.push(node);
+    }
+  }
+
   handleLoadMore = (e) => {
     e.preventDefault();
     this.props.onScrollToBottom();
@@ -73,10 +122,11 @@ class StatusList extends ImmutablePureComponent {
 
   render () {
     const { statusIds, onScrollToBottom, scrollKey, shouldUpdateScroll, isLoading, isUnread, hasMore, prepend, emptyMessage } = this.props;
+    const { isIntersecting } = this.state;
 
-    let loadMore       = '';
-    let scrollableArea = '';
-    let unread         = '';
+    let loadMore       = null;
+    let scrollableArea = null;
+    let unread         = null;
 
     if (!isLoading && statusIds.size > 0 && hasMore) {
       loadMore = <LoadMore onClick={this.handleLoadMore} />;
@@ -95,7 +145,7 @@ class StatusList extends ImmutablePureComponent {
             {prepend}
 
             {statusIds.map((statusId) => {
-              return <StatusContainer key={statusId} id={statusId} />;
+              return <StatusContainer key={statusId} id={statusId} isIntersecting={isIntersecting[statusId]} onRef={this.handleStatusRef} />;
             })}
 
             {loadMore}
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -64,16 +64,16 @@ class StatusList extends ImmutablePureComponent {
 
   attachIntersectionObserver () {
     const onIntersection = (entries) => {
-      const isIntersecting = { };
-
       this.setState(state => {
+        const isIntersecting = { };
+
         entries.forEach(entry => {
           const statusId = entry.target.getAttribute('data-id');
 
           state.isIntersecting[0][statusId] = entry.isIntersecting;
         });
 
-        return [state.isIntersecting[0]];
+        return { isIntersecting: [state.isIntersecting[0]] };
       });
     };
 
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -79,7 +79,7 @@ class StatusList extends ImmutablePureComponent {
 
     const options = {
       root: this.node,
-      rootMargin: '100% 0px',
+      rootMargin: '300% 0px',
     };
 
     this.intersectionObserver = new IntersectionObserver(onIntersection, options);
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -27,7 +27,7 @@ class StatusList extends ImmutablePureComponent {
   };
 
   state = {
-    isIntersecting: { },
+    isIntersecting: [{ }],
   }
 
   statusRefQueue = []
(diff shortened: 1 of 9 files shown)


sorin-davidoi:
We could also make use of the `IntersectionObserver` to implement `onScrollToTop`, `onScrollToBottom` and `onScroll` (using sentinels, as described [here](https://developers.google.com/web/updates/2016/04/intersectionobserver)).

sorin-davidoi:
The performance bottleneck seams to be merging `isIntersecting`:

```js
this.setState(state => ({
  isIntersecting: Object.assign({ }, state.isIntersecting, isIntersecting),
}));
```

ykzts (approved):

nolanlawson:
Can't you just call

```js
this.setState({isIntersecting: isIntersecting})
```

? My understanding is that React's `setState` does the merging automatically for you.

app/javascript/mastodon/components/status_list.js
```diff
@@ -53,6 +60,39 @@ class StatusList extends ImmutablePureComponent {
...
+
+      this.setState(state => ({
+        isIntersecting: Object.assign({ }, state.isIntersecting, isIntersecting),
+      }));
+    };
+
+    const options = {
+      root: this.node,
+      rootMargin: '100% 0px',
+    };
```
nolanlawson: I think we may be able to get better performance by using something like `500%` instead of `100%`. As is, this re-renders components almost constantly as you scroll, leading to very jerky scrolling because of the bottleneck you describe. Whereas if we allowed for 5x the available viewport height, then that should lead to less jumpy scrolling because it won't start un-rendering until you've scrolled 5x of the way down. I admit I still need to test this theory on a mobile device, though.

nolanlawson (changes requested):
This is a really impressive start, but after testing on both Chrome and Firefox on desktop I feel the scrolling is too jerky for it to be an acceptable tradeoff. As you say, the `setState()` is the bottleneck, and unfortunately according to my testing it seems to be related to React's rendering process rather than something we can directly change. I think maybe this can be recovered by increasing how far outside of the DOM node the user has to scroll before we re-render, or perhaps batching our renders in `requestIdleCallback()` or something like that, but I need to experiment a bit more, especially on a mobile browser.

app/javascript/mastodon/components/status_list.js
nolanlawson: Another option is to debounce rather than re-rendering when every individual element leaves or enters the viewport.

nolanlawson:
This is tricky. I tried `500%`, but it didn't change much because once you scroll down 5x the height, you still run into the UI-blocking. I also tried `debounce()` (https://github.com/nolanlawson/mastodon/commit/3f324c75eb0929e08cafa67f9b64d844343a503c), and this also didn't help much, because then we just end up with _really_ blocked scrolling due to all UI updates being queued up into one huge update.

I had a few thoughts while tinkering with this:

1. `requestIdleCallback()` may help because it tells us how many milliseconds are remaining and thus we can avoiding doing excessive `setState()` work inside of a single frame
2. Regardless, the `setState()` updates should really be prioritized such that invisible-to-visible renders are given higher priority than visible-to-invisible renders (since one affects short-term UX whereas the other is just to free up memory over the long term). Otherwise you can get in a situation where the UI is blocked for a long period of time _and_ content is not visible to the user, e.g. right after scrolling to the bottom of the screen.

sorin-davidoi:
Might have found a way around merging the objects. It is a sort of a hack, but we mutate `state.isIntersecting` and store it as an array in order to trick React into rendering.

nolanlawson:
I don't believe it's actually the merging of the objects itself that's the perf bottleneck here, but if it is then maybe a bloom filter would be a good replacement data structure, since it gives you fast lookups and merges at the expense of potential false positives (which in our case would just mean rendering occasional offscreen toots… no big deal). [bloomfilter](https://www.npmjs.com/package/bloomfilter) looks good and is only 1.5k after Uglify.

It seems to me though based on looking at Chrome's Timeline that most of the costs are in React itself, i.e. the costs of just updating the components and rendering them from an invisible state to a visible state. I could be wrong though.

sorin-davidoi:
I timed the actual merging and it sometimes takes up to `20ms`. But you are right, the render of the actual toots is the main problem. I've added some logic to `status` which will make them wait for a certain amount of time before rendering the empty div, prioritising invisible-to-visible.

app/javascript/mastodon/components/status_list.js
sorin-davidoi: > Whereas if we allowed for 5x the available viewport height, then that should lead to less jumpy scrolling because it won't start un-rendering until you've scrolled 5x of the way down.

Don't think this is how it works. With 5x it does the following:

```
Not rendered - next to render when you scroll up
Rendered
Rendered
Rendered
Rendered
Rendered
---------- <- Viewport start
Rendered
---------- <- Viewport end
Rendered
Rendered
Rendered
Rendered
Rendered
Not rendered - next to render when you scroll down
```

The same amount of rendering happens when you scroll, it is just that the chance of viewing the fade in animation when you scroll fast is smaller.

nolanlawson:
visible-to-invisible could also be put in a `requestIdleCallback()` (with polyfill) rather than just delaying. The benefit there is that it won't drop frames, but will just piecemeal start removing unnecessary DOM nodes.

sorin-davidoi:
You're right, `requestIdleCallback` should be ideal for this. Will give it a try tomorrow.

app/javascript/packs/application.js
```diff
@@ -1,7 +1,9 @@
 import main from '../mastodon/main';
 
 if (!window.Intl || !Object.assign || !Number.isNaN ||
-    !window.Symbol || !Array.prototype.includes) {
+    !window.Symbol || !Array.prototype.includes ||
+    !window.IntersectionObserver ||
+    !window.requestIdleCallback) {
```
nolanlawson: I love both `IntersectionObserver` and `requestIdleCallback`, but I'm not sure if I'm comfortable with `polyfills.js` becoming a "not Chrome" tax. In current master, the latest version of Chrome/Edge/Firefox/Safari don't need to load the polyfill; it's only e.g. iOS 9 that needs it. Whereas `IntersectionObserver` is only in Chrome and Edge, and `requestIdleCallback` is only in Chrome and Firefox.

I would prefer to create a `polyfills2.js` or something, so that we can differentiate polyfills that are 1) are in the latest version of each major browser vs 2) in select versions of Chrome/Firefox/Edge.

app/javascript/packs/application.js
```diff
@@ -1,9 +1,27 @@
...
+
+if (needsBasePolyfills) {
+  Promise.all([
+    import('../mastodon/base_polyfills'),
+    import('../mastodon/extra_polyfills'),
+  ]).then(main).catch(e => {
+    console.error(e); // eslint-disable-line no-console
+  });
+} else if (needsExtraPolyfills) {
+  import('../mastodon/extra_polyfills').then(main).catch(e => {
```
nolanlawson: Thank you, this is great. Might be good to add a comment explaining that we're doing this in order not to penalize modern versions of Edge, Firefox, and Safari.

app/javascript/styles/components.scss
```diff
@@ -554,6 +554,14 @@
...
   cursor: default;
 
+  @keyframes fade {
+    0% { opacity: 0; }
+    100% { opacity: 1; }
+  }
+
+  opacity: 1;
+  animation: fade 0.5s linear;
+
```
nolanlawson: The new fade animations are a nice touch

nolanlawson:
I'm now running this PR on https://malfunctioning.technology (along with https://github.com/tootsuite/mastodon/pull/3052); looking forward to testing this live in a few different browsers/devices! Looking at the code itself though, this seems great to me and hopefully will be a huge perf win. :)

nolanlawson:
Unfortunately it looks like it has some bugs on mobile – on both Chrome and Firefox for Android it stops being able to switch views when switching between Home/Notifications/Local/Federated… In the console I see the error `TypeError: r.props.onRef is not a function`. I can repro on desktop by changing the viewport width to a mobile size.

sorin-davidoi:
Fixed it. Forgot that a `Status` can wrap another `Status`.

Gargron:
@nolanlawson Can you confirm that this is ready to be merged now?

ykzts (approved):

nolanlawson:
Awesome @sorin-davidoi, I've got the new version running on https://malfunctioning.technology. Some things I've noticed:

- `document.querySelectorAll('*').length` rarely surpasses 1000 elements, which is a huge improvement over the 4000+ we used to have
- the fade animation is a bit distracting IMO; maybe reducing it from 0.5s to 0.3s would be okay? (This is purely aesthetic tho)
- On some mobile bowsers (e.g. Chrome on a Nexus 5 or Firefox on a Nexus 5X) the loading when you hit the bottom can take quite some time; it would be nice to have a loading indicator. Sometimes I'd get to the bottom and it would say "load more" and I'd be confused for a moment about whether it was loading or not.
- Edge on desktop seems to preserve the momentum as you scroll, meaning that when you hit the bottom it will scroll really fast once the toots load in. TBF though this may be a bug in Edge, and it's also not a deal-breaker.

So I've tested in mobile Safari on iOS, Android Firefox/Chrome on a Nexus 5 and Nexus 5X, and desktop Firefox/Chrome/Edge and overall while I have some minor nitpicks I think this PR is excellent and we should merge it. :+1: It's important to have a timeline that actually works no matter how far you scroll, and this PR solves it in a really elegant way without even needing a virtual list. Awesome work.

nolanlawson (approved):

nolanlawson:
The other thing I'd say is that for a future PR, maybe we should consider removing "Load more" entirely and just proactively loading more toots, so that the illusion of infinite-scrolling is never broken. At that point we could also maybe just get rid of the fade animation entirely, assuming we can optimistically load toots fast enough that the toot is already ready by the time the user scrolls.

Gargron:
>maybe we should consider removing "Load more" entirely and just proactively loading more toots

We used to not have that button, so some people for whom scrolling didn't work for some reason were stuck not being able to scroll back further.

jessebeach:
Please test this out with an assistive tech agent, like a screen reader (VoiceOver) to determine how it handles the removing and adding of content.
My guess is that will handle it well.

jessebeach:
Or is someone can clue me into when this merge is released to a public instance (@nolanlawson), I'll be glad to test it out.

sorin-davidoi:
@jessebeach mastodon.social is running this now (https://toot.cafe/web/statuses/610726).

jessebeach:
@sorin-davidoi oh neat! I also just tested on malfunctioning.technology. Seems to be working well.

PR #3195, opened 2017-05-21
Convert all color specifications to variables.

esetomo:
One step for theme support.

Purpose:

- Do not think it is easy to change the color scheme. (It looks easy with #2912 but it is not yet.)
- In order to aim to reduce the number of colors used.(I think that about 80 colors are not necessary.)


mjankowski (changes requested):
I don't really understand how this is an improvement ... these variable names are essentially just hard-coding the lighten/darken/etc.

In any scenario where you would override one of these, you would be making the name inaccurate.

esetomo:
By extracting the colors used, we can count variations.
By knowing the number of colors used, it is possible to discuss about too many colors in use.
```
$ grep ui-base-color app/javascript/styles/variables.scss | grep a100 | wc -l
23
```
23? Really?

I think that at most five levels are enough.
- $ui-base-color-lightest
- $ui-base-color-lighter
- $ui-base-color
- $ui-base-color-darker
- $ui-base-color-darkest
(There are not the best because there name are not by purpose.)

But I can not do this at once.
It is necessary to take the steps.

esetomo:
The problem when changing color scheme is that the brightness overflows and different designation will be the same color.

In the extreme example, if "$ui-base-color: #ffffff" is used, all lighten up to 40% will be the same color.

Therefore, I will add a safe color range to the comment.

mjankowski:
Ok, I'll stand by for another PR. Just to add a little clarification:

> By knowing the number of colors used, it is possible to discuss about too many colors in use.

I'm not sure that "too many colors" is all by itself necessarily a problem. If there are many colors which are legitimately needed and support the UI, that's fine. If there are many colors but some are dupes of each other or so close to each other that we don't really need two, and that sort of thing - we should fix that.

What I was trying to suggest was that having a variable like `$base-shadow-color--l100-a040` is not really an improvement, because the variable name is just encoding what had previously been done with sass where it is being used, and that if someone wants to change that, re-assigning that variable to some different value is going to be very confusing.

Here's a better example -- we use `lighten($ui-base-color, 10%)` a bunch of times, and it's always used in the stylesheets. If we moved that `lighten` method call into the variables scss file, and gave it a name like `$ui-base-color-lighter` (like you suggested), and then replaced all the `lighten` method calls in the scss files with that variable instead ... I think that would be an improvement (not perfect!, but an improvement), which we be both a) a good step on the path towards better theme/override support, b) also a good change in and of itself, unrelated to theming.

I just don't want to introduce "temporary" stuff which is not an improvement by itself - I'd rather make small changes which are both improvements on their own, and also steps towards larger change.

Issue #499, opened 2017-01-19
add ability to change the size of columns


Moth-Tolias:
context: as it is mastodon is just a centimetre too big to fit on my screen. if i could make them just a tad smaller, they would fit.

alternatively: being able to "minimise" the tooting column as with tweetdeck would also solve this problem for me, but i imagine there are others who would benefit from customisable column sizes

nevillepark:
It's a centimetre too small on mine. Would adjusting the CSS so that (above a certain screen width) each column takes up 25% of the screen solve the first problem?

elad661:
Right now the column width seem to be hardcoded as `width: 330px`.
To work better on wider screens, this could be changed to have both `min-width: 330px` and `width: 25vw`, so that it takes 25% of the screen and doesn't get too narrow if the screen is not wide enough (the min-width can be lower than 330 if needed I guess, but probably not much narrower than that because otherwise it'd be too narrow to use).

I don't think allowing the user to change column sizes individually is a good solution. In my opinion the best solution is to use media queries and the css "vw" unit, to provide *everyone* with a default that works and doesn't require the user to fiddle with column widths.

I guess automatically "collapsing" the "tooting" column with a media query if the screen is too narrow is a good solution for narrow screens (having the columns narrower than 330px will make them not very usable, I think).

flashcode:
On my screen, the 4 columns use ~60% of total width, it would be very nice to have something like 25% for each column and then use full width, so I don't spend my time to scroll columns which can contain some long posts.

Gargron:
See #3207

gsuberland:
#3207 does not currently resolve this at all. I recommend that this issue is re-opened.

Fishrock123:
I'd also like this to be reopened.

PR #3207, opened 2017-05-21
Allow mounting arbitrary columns

### LINKED ISSUE #351: Ability to add more panels/tabs
I would like to be able to add more tabs like you can on tweetdeck

Gargron:
Fix #351 

image
image
image

### DIFF
--- a/app/javascript/mastodon/actions/columns.js
+++ b/app/javascript/mastodon/actions/columns.js
@@ -12,7 +12,7 @@ export function addColumn(id, params) {
       params,
     });
 
-    saveSettings();
+    dispatch(saveSettings());
   };
 };
 
@@ -23,7 +23,7 @@ export function removeColumn(uuid) {
       uuid,
     });
 
-    saveSettings();
+    dispatch(saveSettings());
   };
 };
 
@@ -35,6 +35,6 @@ export function moveColumn(uuid, direction) {
       direction,
     });
 
-    saveSettings();
+    dispatch(saveSettings());
   };
 };
--- a/app/javascript/mastodon/actions/columns.js
+++ b/app/javascript/mastodon/actions/columns.js
@@ -2,6 +2,7 @@ import { saveSettings } from './settings';
 
 export const COLUMN_ADD    = 'COLUMN_ADD';
 export const COLUMN_REMOVE = 'COLUMN_REMOVE';
+export const COLUMN_MOVE   = 'COLUMN_MOVE';
 
 export function addColumn(id, params) {
   return dispatch => {
@@ -20,6 +21,18 @@ export function removeColumn(uuid) {
     dispatch({
       type: COLUMN_REMOVE,
       uuid,
+    });
+
+    saveSettings();
+  };
+};
+
+export function moveColumn(uuid, direction) {
+  return dispatch => {
+    dispatch({
+      type: COLUMN_MOVE,
+      uuid,
+      direction,
     });
 
     saveSettings();
--- a/app/javascript/mastodon/actions/columns.js
+++ b/app/javascript/mastodon/actions/columns.js
@@ -1,0 +1,27 @@
+import { saveSettings } from './settings';
+
+export const COLUMN_ADD    = 'COLUMN_ADD';
+export const COLUMN_REMOVE = 'COLUMN_REMOVE';
+
+export function addColumn(id, params) {
+  return dispatch => {
+    dispatch({
+      type: COLUMN_ADD,
+      id,
+      params,
+    });
+
+    saveSettings();
+  };
+};
+
+export function removeColumn(uuid) {
+  return dispatch => {
+    dispatch({
+      type: COLUMN_REMOVE,
+      uuid,
+    });
+
+    saveSettings();
+  };
+};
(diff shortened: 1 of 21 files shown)


Gargron:
Pain points: We still need to have a context column somewhere to open all the routes in when you click links. Also, it's hard to insert controls for "pinning" and "unpinning" columns into the headers (by hard I mean a lot of annoying code needs to be written or rewritten)

beatrix-bitrot:
hi! I have a couple requests

1. I had some issues with a column (federated timeline) that I created and pinned failing to update. it seemed ok after a refresh. maybe see if you can reproduce?
2. I think in the * menu any columns not currently being displayed should be listed, and any currently displayed don't need to be, so maybe they shouldn't? If I hide the notifications column it should be easy for me to get it back using the * menu.

Thanks!

beatrix-bitrot (approved):
nice

DJSundog:
I'm kind of disappointed this functionality is not exposed via the mobile web interface. Adding functionality that is desktop-only slights the significant portion of the world that has a mobile device but no computer. 

I'd request taking a look at adding this new functionality for all web-based layouts so the divide doesn't get any wider than it already is.

Gargron:
@DJSundog What exactly is the point of having multiple columns on a small screen?

Cassolotl:
I'm with @DJSundog - I think it'd be cool to have your custom columns show on mobile too, and be able to swipe between them. It seems to me that if we can have 6 buttons at the top on mobile view, why not be able to have more that we chose ourselves? :)

DJSundog:
@Gargron the same as having them on desktop, honestly. I'd like to be able to flip back and forth between feeds outside of the standard four. Having a feed for an active tag I'm interested in, a feed for a conversation I'm participating in, etc. would significantly increase the utility of the web client, regardless of the form factor of the device the web client is being viewed on.

joyeusenoelle:
Adding my vote to having this available on the mobile interface. It's not about multiple columns, it's about custom columns. The mobile view kicks in at a VERY wide aperture - so wide that I get it on my desktop just because I prefer to have my browser 2/3-screen instead of full screen. I'm frustrated to hear that this functionality might not be extended to every web user.

chosafine:
@Gargron for phones in landscape, having multiple columns (Tweetbot on iOS for iPhone Plus devices is a good example) can serve useful. Having one very wide column ruins usability in landscape on mobile.

nightpool:
I think having one column on mobile is probably net good, but like Noelle
said, it's not about multiple columns—it's about having custom columns

sydneyfalk:
Also, cheaper tablets are a thing. Low res but wider display means one huge wide column every time.

It might be useful to have a ticky that just 'turns off' automatic one-column layout, or at least lower the bar for enforcing it? Or maybe a ticky to force one-column, and a mutually exclusive other ticky that forces multi-column?

DJSundog:
For reference, this is two separate instances of the web app running split screen on my 10" Samsung tablet:

screenshot_20170611-122541

There are definitely devices in use that could benefit from being able to toggle multi-column display.

Gargron:
What's a good break point for turning into single-column layout? Currently it's 1024px. 680px?

DJSundog:
Due to HiDPI, this device shows 1024x768 landscape and 768x1024 portrait, so maybe 768px? (Not that my device is the important one, just as a data point)

sydneyfalk:
A lower threshold would help people using laptops or working with smaller screens (netbooks and such), I suppose.

But if a toggle's not off the table: Perhaps a three-state toggle?

        C O L U M N S
One------Auto------Multi

They can force one column (some people might like that on desktop, visual focus issues can be a thing), they can force multicolumn, they can let the auto-select happen.

Zazcallabah:
@Gargron You closed several issues relating to column size, referring to this one. But nothing here mentions column size?

faissaloo:
So how do I get the column to stretch the width of the window? @Gargron

Strit:
I think a way to manage column sizes could be done fine with percentages.

100% is the entire screen width of the window. Simply divide that by the number of active columns, so 4 as default, which would give each column 25% width of the window.
HTML websites have been doing similar things for decades at this point.

fbruetting:
The current width is way too narrow. Can’t we drag them as wide as we like to, or something like that?

fbruetting:
The current width is way too narrow. Can’t we drag them as wide as we like to, or something like that?

This here image is way more optimal to me (it’s from Corebird):

gsuberland:
I have to agree with others. I went searching for a way to get mastadon's columns to be bigger (they look ludicrously cramped on a 4K display) and found #499 - a closed issue that referenced this issue as being a fix. But it appears that this PR does nothing to actually fix that issue.

For reference, this is what it looks like on a 4K screen:

image

I understand that we could use Tampermonkey or similar tools to implement this, but that is neither portable nor reliable, and runs rather antithetical to the fact that this feature was purported to have been implemented.

fbruetting:
Yeah, when my mother complains, I also tell her to RTFM and hack it accordingly.


// But I don't have that scaling issue on Fedora.

Hoeloe:
I'm also using a 4K screen and did a simple change through Firefox's CSS editor to use percentages instead of hardcoded widths, just to see how that would look. 
image

Obviously, this works well, and wasn't a difficult change, so it would be nice if something like this were default, because the interface is extremely awkward currently.

PR #3213, opened 2017-05-22
fix-cache-returns-nil

masarakki:
In my instance, sometimes `user.settings.interactions` returns `nil` .
I understand that this code never returns `nil`, but something happen...

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -35,7 +35,6 @@ class Setting < RailsSettings::Base
           default_settings[key]
         end
       end
-
       val
     end
 
--- a/app/lib/settings/extend.rb
+++ b/app/lib/settings/extend.rb
@@ -5,7 +5,7 @@ module Settings
     extend ActiveSupport::Concern
 
     def settings
-      ScopedSettings.for_thing(self)
+      @settings ||= ScopedSettings.new(self)
     end
   end
 end
--- a/app/lib/settings/scoped_settings.rb
+++ b/app/lib/settings/scoped_settings.rb
@@ -1,14 +1,69 @@
 # frozen_string_literal: true
 
 module Settings
-  class ScopedSettings < ::Setting
-    def self.for_thing(object)
+  class ScopedSettings
+    def initialize(object)
       @object = object
-      self
     end
 
-    def self.thing_scoped
-      unscoped.where(thing_type: @object.class.base_class.to_s, thing_id: @object.id)
+    # rubocop:disable Style/MethodMissing
+    def method_missing(method, *args)
+      method_name = method.to_s
+      # set a value for a variable
+      if method_name[-1] == '='
+        var_name = method_name.sub('=', '')
+        value = args.first
+        self[var_name] = value
+      else
+        # retrieve a value
+        self[method_name]
+      end
+    end
+    # rubocop:enable Style/MethodMissing
+
+    def respond_to?(_method_name, _include_private = false)
+      true
+    end
+
+    def all_as_records
+      vars = thing_scoped
+      records = vars.map { |r| [r.var, r] }.to_h
+
+      Setting.send(:default_settings).each do |key, default_value|
+        next if records.key?(key) || default_value.is_a?(Hash)
+        records[key] = Setting.new(var: key, value: default_value)
+      end
+
+      records
+    end
+
+    def []=(key, value)
+      key = key.to_s
+      record = thing_scoped.find_by(var: key) || thing_scoped.new(var: key)
+      record.value = value
+      record.save!
+
+      Rails.cache.write(Setting.cache_key(key, @object), value)
+      value
+    end
+
+    def [](key)
+      Rails.cache.fetch(Setting.cache_key(key, @object)) do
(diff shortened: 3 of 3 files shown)


nightpool:
Looks like this changes a bunch of existing behavior—see the test failures. Can you outline exactly what the issue is and why it's a problem?

If you say "that this code never returns nil", then you should probably investigate *why* it's going on not just masking the symptoms.

Gargron:
Does `user.settings.interactions` for the same user sometimes return nil, or is it more of an issue of only some users missing defaults?

masarakki:
this happens both case user has some settings / no settings ,

I found what happened on rails console,
```
irb(main):048:0> Rails.cache.exist?(Setting.cache_key('interactions', user))
=> true
irb(main):049:0> Rails.cache.fetch(Setting.cache_key('interactions', user))
=> nil
# ..few minutes after
irb(main):050:0> Rails.cache.fetch(Setting.cache_key('interactions', user))
=> {"must_be_follower"=>false, "must_be_following"=>false}
```

Gargron:
So the problem is the old result is cached and the cache is not discarded when setting is updated? Or am I still misunderstanding what happens and why?

masarakki:
Essentially, this problem is `db_val` of 'interactions' or `defuallt_values['interactions']` never returns `nil`, but (it seems that) redis stored `nil`

app/models/setting.rb
```diff
@@ -35,7 +36,7 @@ def [](key)
           default_settings[key]
         end
       end
-
+      val = default_settings[key] unless Rails.cache.exist?(key)
```
Gargron: Should this not be `if val.nil?`

masarakki:
```
rails c                                [19:13:37]
Loading development environment (Rails 5.0.2)
[1] pry(main)> Settings::ScopedSettings.instance_variable_get(:@object)
=> nil
[2] pry(main)> User.first.settings
  User Load (0.6ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> Settings::ScopedSettings(id: integer, var: string, value: text, thing_type: string, thing_id: integer, created_at: datetime, updated_at: datetime)
[3] pry(main)> Settings::ScopedSettings.instance_variable_get(:@object)
=> #<User id: 1, email:`...>
```

I think it is the factor of its problem.
I fix to [using instance of `ScopeSettings`](https://github.com/dwango/mastodon/pull/68), and I'm testing in my instance.

alpaca-tc:
I noticed that current `Settings::ScopedSettings` is not thread-safe. I might be wrong, but is this the cause? 🤔 
I am going to create new PR to fix it.

Gargron:
@alpaca-tc Perhaps we should add monitor/synchronize to it? I don't know how to reproduce this issue for testing, so I am just guessing...

alpaca-tc:
@Gargron I am just guessing, too. I will try to reproduce it. 😕

alpaca-tc:
@Gargron just finished, we should fix this problem as soon as possible.
I'll try to create pull-request tomorrow 😄 

```
# evaluate script on rails console or rails runner
ActiveRecord::Base.logger = Logger.new('/dev/null')
raise 'please enable redis store' unless Rails.cache.is_a?(ActiveSupport::Cache::RedisStore)

def update_settings(user)
  new_value = user.settings['test'].to_i + 1
  user.settings['test'] = new_value # UPDATE "settings" SET "value" = $1, "updated_at" = $2 WHERE "settings"."id" = $3
end

begin
  threads = []

  # This thread doesn't update settings.
  # If other thread updates record or cache, this thread raise error.
  threads << Thread.new do
    user = User.find(1)
    previous_value = user.settings['test']

    50.times { user.settings; sleep(0.1) } # Overwrite Settings::ScopedSettings's @object to check thread-safe

    current_value = user.settings['test']

    unless previous_value == current_value
      raise "Ooops!! settings is updated by other thread. prevous:#{previous_value}, current:#{current_value}" 
    end
  end

  # Unfortunately, this thread updates settings of first user
  threads << Thread.new { 50.times { update_settings(User.find(2)) } }

  threads.each(&:join)
rescue => error
  puts error
end
```

alpaca-tc:
Ah, it seems that this PR fixed the problem, does not? @masarakki 

I hope to merge after our reviewing. how about it? @Gargron

Gargron (approved):

Gargron:
Yes I didn't notice the refactor was already in the PR. Still, should I wait for you to confirm this fixes the issue, @alpaca-tc ?

masarakki:
After this patch,  `Setting` is (maybe) used only for global settings.
So I think it's not so dangerous.

I suggest to avoid using `rails-settings`.

Gargron:
I am adding a Redis lock gem in #4258. One of the examples in the gem documentation is using the lock for fetching cache: <https://github.com/marioizquierdo/mario-redis-lock/blob/master/EXAMPLE_DOG_PILE_EFFECT.md> Perhaps this would be useful for this code? I think it's still a bit buggy, I remember sometimes seeing in my sidekiq logs that some local user's settings which should never be nil return nil and the job fails

PR #3216, opened 2017-05-22
Re-initialize Rails.cache after setting config.cache_store.

clworld:
In 1.4rc2, Rails.cache is missinitialized and not using Redis.
Because config/initializers/redis.rb is executed after Redis.cache is initialized.
Re-initialize Rails.cache or Setting config.cache_store in application.rb or environments is required.
This PR add re-initialize code in config/initializers/redis.rb.

http://stackoverflow.com/questions/5810289/setting-the-cache-store-in-an-initializer/38619281#38619281

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ gem 'statsd-instrument', '~> 2.1'
 gem 'twitter-text', '~> 1.14'
 gem 'tzinfo-data', '~> 1.2017'
 gem 'webpacker', '~> 1.2'
+gem 'redis-rails', '~> 5.0'
 
 group :development, :test do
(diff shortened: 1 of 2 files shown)


clworld:
Some regressions from https://github.com/tootsuite/mastodon/pull/2732/files remained.
1. cache_params[:namespace] isn't set if REDIS_NAMESPACE isn't set.
2. Non-standard redis-store creation.
3. Redis-store can't be used in development.

clworld:
I made alternative fix for this problem. https://github.com/tootsuite/mastodon/pull/3219

clworld:
Probrem is fixed by #3219.

PR #3233, opened 2017-05-22
Similarly to #2426, put creation of remote statuses in a transaction,

Gargron:
so that public timeline/caching would not encounter incomplete data

cc @walf443


walf443:
it looks good to me

ykzts (approved):

Issue #1039, opened 2017-04-06
Web push notifications


stuartlangridge:
Currently Mastodon provides the streaming API and mobile apps that want to use push notifications need to store the user's token on their server and listen to the streaming API and [generate push notifications themselves](https://github.com/tootsuite/mastodon/blob/master/docs/Using-the-API/Push-notifications.md). Modern browsers have now implemented [Web Push](https://developers.google.com/web/fundamentals/engage-and-retain/push-notifications/), which means that the Mastodon web app itself could send these push notifications to web browsers without having to use the company-specific APIs such as FCN. This would remove one of the major barriers to people using the Mastodon web app on their phones -- at the moment (based on a [very informal poll](https://mastodon.social/@sil/1686347)) the major piece of resistance to that is that the web doesn't do push notifications. As a data point, the [Other Place has started doing push notifications](https://lite.twitter.com/) to its web app as of today.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

gregjw:
+1

sorin-davidoi:
In progress: #3243.

sorin-davidoi:
@Gargron This is closed by #3243.

Issue #1681, opened 2017-04-13
OAuth access tokens - never removed, never expire


happycoloredbanana:
There are several (potential) issues with the tokens:

1. Access tokens are never removed from the database (table ```oauth_access_tokens```). This is how it's done in Doorkeeper (doorkeeper-gem/doorkeeper#536). The table keeps growing indefinitely.

1. Every time Mastodon's web GUI is loaded, a new token is issued. Concern: will it be possible to fill the ```oauth_access_tokens``` table up by simply refreshing the Mastodon page repeatedly? Can it be used as a DOS attack?

1. Access tokens never expire, and each user ends up with a bunch of them (see the above). Probably, can be bad security-wise.

I'm not a security expert, just asking.

* * * *

- [ x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

tobypinder:
Presumably matching for existing valid tokens using `first_or_initialize`/`first_or_create` would help with this. I haven't seen the masto implementation but might have fixed something similar in different apps using Doorkeeper.

sungo:
In terms of token expiry, the culprit is here: https://github.com/tootsuite/mastodon/blob/master/config/initializers/doorkeeper.rb#L26 where the initializer sets access tokens to never expire.

A better config here is to set access_tokens_expires_in to some not-nil value and then set use_refresh_token (https://github.com/tootsuite/mastodon/blob/master/config/initializers/doorkeeper.rb#L42)

That'll allow tokens to expire but also allow apps to refresh their tokens using the regular oauth flow without needing the user to login again.

sungo:
Given that the commit doesn't resolve the non-expiring oauth token issue, should I take that to mean that Mastodon has no interest right now in oauth tokens that expire and can be refreshed?

kaniini:
hey @Gargron could you elaborate on whether Mastodon considers the lack of expiring tokens to be a security issue?  do you know of any apps which explicitly depend on the lack of expiring tokens?

Gargron:
All apps depend on that. All tokens can be revoked by the app that has them, and by the user from the "authorized applications" page. Infinitely growing table of apps is a concern, but a small one, given how tiny that data is compared to everything else. A suggestion I have voiced previously was to discard apps with no actively used tokens within the last year.

kaniini:
What is your opinion on pruning unused tokens automatically?

Gargron:
[...]

PR #3243, opened 2017-05-23
Web Push Notifications

### LINKED ISSUE #1122: Subscribe for mobile notifications from accounts to drive engagement
A big part of what brings me back to twitter every morning are all the notifications on my phone from twitter accounts I've subscribed to. I read them like the morning paper.

If Mastodon had this, I'd be less inclined to visit the bird site and more inclined to stay just on Mastodon.

* * * *

sorin-davidoi:
This PR introduces the ability to send [Web Push Notifications](https://developers.google.com/web/fundamentals/engage-and-retain/push-notifications/web-push-protocol). The basic idea is to save a push subscription to the backend, which can then be used to wake up the Service Worker and display a notification when something happened, even if Mastodon is not opened. Together with #3052 this can make Mastodon pretty much indistinguishable from a native application (at least on Android).

Features to implement:
- [x] Register Service Worker
- [x] Generate VAPID keys in the backend
- [x] Pass the public VAPID key to the frontend
- [x] Register push subscription to the backend
- [x] Delete invalid push subscriptions
- [x] Send push notifications
  - [x] Mentions
  - [x] New followers
    - [x] Follow
    - [x] Follow request
  - [x] Favorites
  - [x] Boosts
- [x] Localize messages
- [x] Handle notification interactions
- [x] Integrate in the UI - one should be able to control the settings for each individual subscription (e.g. for what type of events it gets triggered). This is not a per account preference, as I assume people want different settings for each device (e.g. only push for mentions on mobile, but push all notifications on desktop).
- [x] Handle backend changing the `applicationServerKey`
- [x] Handle RTL
- [x] Display attachment
- [x] Unregister push subscription when logging out - can not figure out how to associate a subscription with a session, such that when the user logs out the subscription is destroyed
- [x] Set VAPID keys via configuration
- [x] Move logic out of `SettingsController` - I could not figure out how to create a new endpoint for this, so I've hijacked this controller for adding and updating subscriptions.
- [x] Add actions to the notification
  - [x] Favourite / boost status

[Browser support](http://caniuse.com/#search=push): Chrome and Firefox. Cannot be polyfilled.

Closes #1122, #1039.


nolanlawson:
I was just about to ask how you managed to get Chrome to do push notifications without a GCM/FCM ID, but it looks like they've [removed that requirement](https://web-push-book.gauntface.com/chapter-06/01-non-standards-browsers/), which is going to make our lives a lot easier. Hooray! :tada:

sorin-davidoi:
I think this is ready for a first review. Updated the description with stuff that still needs to be done. Tested on Firefox and Chrome on Linux (KDE Plasma) and on Android.

photo_2017-05-27_00-31-24
image
photo_2017-05-27_00-30-57
photo_2017-05-27_00-31-16

app/javascript/mastodon/web_push.js
```diff
@@ -0,0 +1,91 @@
...
+
+      return subscribe(registration).then(sendSubscriptionToBackend);
+    })
+    .then(subscription => store.dispatch(setSubscription(subscription)))
+    .catch(error => {
+      store.dispatch(clearSubscription());
+
+      try {
+        getRegistration()
+          .then(getPushSubscription())
```
akihikodaki: I suppose you mean `getPushSubscription` instead of `getPushSubscription()`?

db/schema.rb
```diff
@@ -358,6 +358,16 @@
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true, using: :btree
   end
 
+  create_table "web_push_subscriptions", force: :cascade do |t|
+    t.integer  "account_id"
```
akihikodaki: Why not index?

app/controllers/api/web/push_notification_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
akihikodaki: Spec?

app/controllers/api/web/settings_controller.rb
```diff
@@ -1,20 +1,49 @@
 # frozen_string_literal: true
 
 class Api::Web::SettingsController < ApiController
+  include ActionView::Helpers::TranslationHelper
+
   respond_to :json
 
   before_action :require_user!
 
   def update
```
akihikodaki: Spec this, too.

app/controllers/api/web/push_notification_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    web_notification.data = params[:data]
+    web_notification.save!
+
+    render_empty
+  end
+
+  private
+
+  def web_notification
+    @_web_push_notification ||= WebPushNotifications.where(user: current_user).first_or_initialize(user: current_user)
```
akihikodaki: Where `WebPushNotifications` comes from?

app/controllers/api/web/settings_controller.rb
```diff
@@ -1,20 +1,49 @@
 # frozen_string_literal: true
 
 class Api::Web::SettingsController < ApiController
+  include ActionView::Helpers::TranslationHelper
```
akihikodaki: What is it for?

akihikodaki (changes requested):
I have some questions and found something suspicious, so please review them. I'm looking forward this feature.

app/javascript/mastodon/web_push.js
sorin-davidoi: Yes.

app/controllers/api/web/settings_controller.rb
```diff
@@ -1,20 +1,49 @@
...
+    web_subscription = WebPushSubscription.new(
+      endpoint: params[:data][:web_push_subscription][:endpoint],
+      key_p256dh: params[:data][:web_push_subscription][:keys][:p256dh],
+      key_auth: params[:data][:web_push_subscription][:keys][:auth]
+    )
+
+    current_account.web_push_subscriptions << web_subscription
+    current_account.save!
+
+    render json: web_subscription.as_payload, status: 200
```
sorin-davidoi: Don't think so.

db/schema.rb
sorin-davidoi: This was generated by the following command, so I cannot answer that 😄  How should it look like?
```
rails generate model WebPushSubscription account_id:integer endpoint:string:unique key_p256dh:string key_auth:string
```

app/controllers/api/web/push_notification_controller.rb
sorin-davidoi: Ruby magic 😄 ? Guess it is the database model?

sorin-davidoi:
Will try to see tomorrow how Rails tests work and maybe add some specs.

db/schema.rb
akihikodaki: Please look at other tables. Here is an example:
```Ruby
t.index ["account_id"], name: "index_reports_on_account_id", using: :btree
```

app/controllers/api/web/push_notification_controller.rb
akihikodaki: Ruby can sometimes do magic, but it should be in `schema.rb` anyway if it is a database model.

app/controllers/api/web/settings_controller.rb
```diff
@@ -1,20 +1,47 @@
 # frozen_string_literal: true
 
-class Api::Web::SettingsController < ApiController
```
akihikodaki: This makes the spec fail.

akihikodaki (changes requested):
I found an error with CI.

sorin-davidoi:
If [this spec](https://www.chromestatus.com/feature/5743740178137088) gets traction we might even be able to reply from the notification.

Gargron:
33 files change - I'm already scared :D Is this ready for review?

sorin-davidoi:
I still haven't figured out how to attach the subscription to the session, so that when the user logs out the subscription gets deleted. Also need to pass the VAPID keys via config files / environment (right now they are generated when the server starts).

sorin-davidoi:
@akihikodaki  I explored a bit our options for clearing the subscriptions. We have the following in `doorkeeper.rb`:
```ruby
  # Reuse access token for the same resource owner within an application (disabled by default)
  # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/383
  reuse_access_token
```

What I understand from this is that we do not keep track of each user session (we use one token for all sessions), so there is no way to associate a push subscription with the session that created it. What is the reason for this? What would be the implications of issuing a new token for each login and deleting the token when the user logs out (attaching the push subscription to this token and deleting them together).

Another approach would be to manually delete the push subscription from the front-end when the user logs out, but that would be a huge security / privacy issue.

Don't really know how to proceed with this.

Gargron:
>What is the reason for this? 

The web UI uses the API like any other app would use the API, so it needs an access token. Reusing of the access token was implemented in response to #1681

akihikodaki:
> Another approach would be to manually delete the push subscription from the front-end when the user logs out, but that would create a huge security / privacy issue.

Could you tell me about your security concern?

sorin-davidoi:
1. Call `/auth/sign_out` -> success
2. Call endpoint deleting the current subscription -> fail

Result: the user is logged out, but the browser keeps getting push notifications for that account. Also someone can mess with the code and skip the request to delete the subscription.

Signing out and deleting the subscription should be an atomic operation, handled by the back-end.

Gargron:
http://www.jonathanleighton.com/articles/2013/revocable-sessions-with-devise/

akihikodaki:
@sorin-davidoi 
> the user is logged out, but the browser keeps getting push notifications for that account. Also someone can mess with the code and skip the request to delete the subscription.

Even now we have `access_token`, which remains valid after signing out.

spec/controllers/api/web/push_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        }
+      }
+    }
+  end
+
+  describe 'POST #create' do
+    before do
+      sign_in(user)
+
+      ActiveSession.deactivate warden.raw_session['auth_id']
```
sorin-davidoi: The tests fail because these two lines need to be executed after each call to `sign_in`. Tried putting the following into `spec_helper.rb` but didn't help. Any pointers?

```ruby
Devise.setup do |config|
  Warden::Manager.after_set_user except: :fetch do |user, auth, opts|
    ActiveSession.deactivate warden.raw_session['auth_id']
    warden.raw_session['auth_id'] = user.activate_session
  end
end
```

Gargron:
I suggest extracting Devise modifications into a separate PR to make reviewing easier

spec/controllers/api/web/push_subscriptions_controller_spec.rb
akihikodaki: I think it is because `Devise::Test::ControllerHelpers` does not support configuration by `Devise.setup`.
https://github.com/plataformatec/devise/blob/9fe7040db90be1acb499d2534b077266dbf3b209/lib/devise/test/controller_helpers.rb#L63

`Devise::Test::ControllerHelpers` is included in `rails_helper`. You may hook `sign_in` there. (it's is a bit hacky, but it cannot be helped, I think.)
https://github.com/tootsuite/mastodon/blob/fd66f7cdc0931a67a2e24d887fb79f44fc42e2bc/spec/rails_helper.rb#L26

akihikodaki (dismissed):

sorin-davidoi:
Will continue work once https://github.com/tootsuite/mastodon/pull/3616 is merged.

config/routes.rb
```diff
@@ -206,6 +206,12 @@
 
     namespace :web do
       resource :settings, only: [:update]
+      resource :push_subscriptions, only: [:create] do
+        # Why doesn't PUT /api/web/push_subscriptions/:id work?
```
sorin-davidoi: @Gargron Do you have some insights into this? If I call `axios.put('/api/web/push_subscriptions/2'` I get this in the logs:

```
Started PUT "/api/web/push_subscriptions/2" for 127.0.0.1 at 2017-06-24 16:34:57 +0200
Processing by ApplicationController#raise_not_found as HTML
```

`axios.put('/api/web/push_subscriptions')` works fine though.

Cannot seem to figure it out.

Gargron: "resource" means there's no index, show with no ID (e.g. there is always one item)

"resources" is what gives you both index and show with ID

sorin-davidoi:
Added actions when you are mentioned:

sorin-davidoi:
I think this is ready for review. The service worker is configured **not** to cache anything. The `offline-plugin` configuration is copied from #3052 and **is only active during in the production build**. All the lint errors are logging statements that I've left in to aid testing / debugging.

sorin-davidoi:
Any brave volunteers who would like to deploy this to a test instance? @nolanlawson , @beatrix-bitrot, heard you might fancy push notifications on mobile devices :smile:

nightpool:
Does this only work on mobile devices? Or can I use it on desktop browsers as well?

sorin-davidoi:
On desktop as well.

sorin-davidoi:
On @nolanlawson's todo list for #3052:
> update tootsuite/mastodon to specify that sw.js should be served with max-age:0 (otherwise the browser may cache the sw.js for as much as 24 hours)

I suppose we would need to do this here as well. I guess this needs to go in the documentation (the example nginx configuration)? Should also be specified in the release notes I think (together with generating the VAPID keys).

sorin-davidoi:
Unfortunately, this won't work in Samsung Internet. They support Service Worker and Push Notifications, but still rely on GCM for authentication.

nolanlawson:
Hmm seems the only fix for Samsung Internet would be to ask admins to sign up for GCM/FCM and put their credentials in a config somewhere. The other option is to wait for Samsung to fix it, which I hope they would since no other browser requires this (Chrome used to but dropped it, and Edge won't require any such thing).

I'd vote to just wait for Samsung to fix it.

sorin-davidoi:
Just tried the beta from the Play Store and still no support.

nolanlawson:
Asking admins to supply their own FCM/GCM feels very un-Mastodon to me (centralized proprietary service), so maybe let's just hope Samsung fixes it.

Can we detect Samsung Internet (using something other than UA sniffing) and treat it like a browser that just doesn't support push?

sorin-davidoi:
We don't detect it explicitly, but the setup is conservative. Will only be enabled in the UI if everything works: https://github.com/tootsuite/mastodon/pull/3243/files#diff-d5cfaccaa25810da832cba018b1257e3R113

nightpool:
I'm not too worried about Samsung Internet—do we have substantial evidence
that people are using it? It looks like usage is about 10% that of Chrome
on Android, and half of UC browser.

A much larger holdout here is Safari for iOS, but there's not much we can
do about that seeing as they haven't even started implementing service
workers yet.

sorin-davidoi:
Speaking of UC browser, I tried that one too. Mastodon does not seem to load at all? And there seems to be no way to debug it :man_shrugging:

nightpool:
aaaaa I want this landed already~~~ the new popups are so nice.

sorin-davidoi:
Blocked by #4058.

nightpool:
@sorin-davidoi why is that blocking? isn't that just a cosmetic error with the sessions screen?

It feels pretty unrelated to this pull request for me

sorin-davidoi:
Push subscriptions are attached to sessions (which are supposed to last until you log in). Since the sessions are created so often, you will get the "Subscription registered" multiple times a day. More important, if you enable push notifications, and a new subscriptions is created because of this problem, it will default to being turned off again.

On glitch.social, I get ~6 new sessions a day, and I have to remember to enable push subscriptions each time.

m4sk1n:
@sorin-davidoi, this is a bit annoying, but works good…

sorin-davidoi:
Yes, it works, but I would argue that having to enable them multiple times during the day is not a good experience. People will be confused as to why their preferences are not saved.

sorin-davidoi:
@beatrix-bitrot I've modified this to work with #4091.

beatrix-bitrot:
this currently needs a rebase. 

special attention should be paid to home controller and the refactoring of JS includes that was done very recently.

after that is done and CI passes I recommend this work be merged

nightpool:
codeclimate is complaining about the console.log statements—do you think we're good to take them out?

sorin-davidoi:
Despite all the testing, I have a feeling that this may cause problems, which will be hard to diagnose without these logs. Would it be okay to use `eslint-ignore-line` to silence them for now?

beatrix-bitrot (approved):
I, Beatrix Bitrot, being of sound mind and of my own volition APPROVE THESE CHANGES

nightpool:
@sorin-davidoi it's hard to see how user-side logging is really going to help in this case, (it's hard enough to get users to *tell* us when something goes wrong, much less get them to have the dev console open at the right time) but if you think it's important to keep it, I defer to your judgement

beatrix-bitrot:
ah, about the console log statements? i'm indifferent

for most users i don't think they be any use but on bleeding edge instances they may come in handy

maybe leave them in master for a bit and remove before the next tagged release? idk

another option is to leave them in here, then make another PR that removes them so that if anyone needs to enable them for testing all they need to do is revert the commit that removes them

idk

app/controllers/api/web/push_subscriptions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    active_session = current_session
+
+    unless active_session.web_push_subscription.nil?
+      active_session.web_push_subscription.destroy!
+      active_session.web_push_subscription = nil
+      active_session.save!
+    end
+
+    web_subscription = ::Web::PushSubscription.new(
+      endpoint: params[:data][:endpoint],
```
akihikodaki: Why not `params.require`?

app/controllers/api/web/push_subscriptions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def create
+    active_session = current_session
+
+    unless active_session.web_push_subscription.nil?
+      active_session.web_push_subscription.destroy!
+      active_session.web_push_subscription = nil
+      active_session.save!
+    end
+
+    web_subscription = ::Web::PushSubscription.new(
```
akihikodaki: Use `create!`.

app/controllers/api/web/push_subscriptions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      endpoint: params[:data][:endpoint],
+      key_p256dh: params[:data][:keys][:p256dh],
+      key_auth: params[:data][:keys][:auth]
+    )
+
+    web_subscription.save!
+
+    active_session.web_push_subscription = web_subscription
```
akihikodaki: Use `update!`.

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -0,0 +1,88 @@
...
+
+const handleNotificationClick = (event) => {
+  const reactToNotificationClick = new Promise((resolve, reject) => {
+    if (event.action) {
+      const action = event.notification.data.actions.find(({ action }) => action === event.action);
+
+      if (action.todo === 'expand') {
+        return expandNotification(event.notification);
+      } else if (action.todo === 'request') {
+        return makeRequest(event.notification, action)
```
akihikodaki: The following statement would be enough. A promise as resolved value will be evaluated and expanded.
```JavaScript
resolve(makeRequest(event.notification, action)
  .then(() => removeActionFromNotification(event.notification, action)));
```

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -0,0 +1,88 @@
...
+          .then(() => removeActionFromNotification(event.notification, action))
+          .then(resolve)
+          .catch(reject);
+      }
+
+      return null;
```
akihikodaki: Returned values does not really matter for promises.

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const nextNotification = cloneNotification(notification);
+
+  nextNotification.actions = actions;
+
+  return self.registration.showNotification(nextNotification.title, nextNotification);
+};
+
+const handleNotificationClick = (event) => {
+  const reactToNotificationClick = new Promise((resolve, reject) => {
```
akihikodaki: * Resolve or reject always.
* Do you even have to wrap in a new promise?

app/javascript/mastodon/web_push_subscription.js
```diff
@@ -0,0 +1,137 @@
...
+          console.log('Fetching subscription and unsubscribing...');
+          getRegistration()
+            .then(getPushSubscription)
+            .then(unsubscribe);
+        } catch (e) {
+          console.error('Something went wrong yet again:', error);
+        }
+      });
+  } else {
+    // eslint-disable-next-line no-console
```
akihikodaki: I think you are aware of, but here I note you should be consistent, whether you use `eslint-disable-next-line` or not. If you use, apply for all other `console`. If you do not, remove this.

db/migrate/20170625175513_create_web_push_subscriptions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateWebPushSubscriptions < ActiveRecord::Migration[5.1]
+  def change
+    create_table :web_push_subscriptions do |t|
```
akihikodaki: I guess the following columns cannot not be null, can they?

app/services/notify_service.rb
```diff
@@ -61,6 +61,27 @@ def create_notification
...
+        session.web_push_subscription.push(@notification)
+      rescue Webpush::InvalidSubscription, Webpush::ExpiredSubscription
+        # Subscription expiration is not currently implemented in any browser
+        session.web_push_subscription.destroy!
+        session.web_push_subscription = nil
+        session.save!
+      rescue Webpush::PayloadTooLarge, Webpush::TooManyRequests => e
+        Rails.logger.error(e)
+      rescue Webpush::Error => e
+        # Failing to send push notifications should not result in Internal Server Error
```
akihikodaki: > Failing to send push notifications should not result in Internal Server Error

Yes, but that is controllers' business. If it is just to eliminate Internal Server Error, rescue in controllers.

spec/controllers/api/web/push_subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Api::Web::PushSubscriptionsController do
+  render_views
+
+  let!(:user) { Fabricate(:user) }
```
akihikodaki: I think `let` is fine for these cases.

spec/fabricators/web_push_subscription_fabricator.rb
```diff
@@ -0,0 +1,6 @@
+Fabricator(:web_push_subscription) do
+  account_id 1
+  endpoint   "MyString"
```
akihikodaki: Use Faker gem.

akihikodaki (changes requested):
I have added many comments, but they are trivial and the logic itself looks fine. I hope they will be addressed soon and the change will be merged.

app/controllers/api/web/push_subscriptions_controller.rb
sorin-davidoi: Hm, does not seem to work, as this is called from an endpoint.

app/controllers/api/web/push_subscriptions_controller.rb
sorin-davidoi: Same comment as `create!`.

app/javascript/mastodon/web_push_subscription.js
sorin-davidoi: Will remove the other logs.

app/services/notify_service.rb
sorin-davidoi: Not really sure where to move the rescues then (this is called from a ton of places, all of them services - should each of them duplicate the rescues?).

app/controllers/api/web/push_subscriptions_controller.rb
nightpool: what's the problem with calling create from an endpoint?

sorin-davidoi: Ah, think I misunderstood the comment. I took it as to change the name of the method from `create` to `create!`. Will update it after I fix the conflicts :roll_eyes:

app/services/notify_service.rb
nightpool: Are we sending notifications synchronously? that seems really bad.

sorin-davidoi: To be honest I don't really know - don't have a strong grasp on the whole sidekiq thing.

config/webpack/production.js
```diff
@@ -39,5 +42,16 @@ module.exports = merge(sharedConfig, {
...
     }),
+    new OfflinePlugin({
+      publicPath: publicPath, // sw.js must be served from the root to avoid scope issues
+      caches: { }, // do not cache things, we only use it for push notifications for now
+      ServiceWorker: {
+        entry: path.join(__dirname, '../../app/javascript/mastodon/service_worker/entry.js'),
+        cacheName: 'mastodon',
+        output: '../sw.js',
+        publicPath: '/sw.js',
+        minify: false,
```
akihikodaki: Why don't you minify?

spec/fabricators/web_push_subscription_fabricator.rb
```diff
@@ -0,0 +1,6 @@
+Fabricator(:web_push_subscription) do
+  account_id 1
```
akihikodaki: `Web::PushSubscription` does not have `account_id` column.

akihikodaki (dismissed):
All problems I have requested are addressed, but I found a few other suspicious things.

Gargron (approved):

sorin-davidoi:
For things that were left unaddressed here see #4200.

PR #3245, opened 2017-05-23
don't notify me when my toot is faved by someone i muted

beatrix-bitrot:


Gargron:
I thought muting was for not seeing someone's content but for still letting them interact with you

beatrix-bitrot:
They still can, you just don't be notified when they fav your toots.

If you mute someone, you probably don't wanna be notified about them unless it conveys you some useful information (like they boosted your toot, now more people see it etc)

hoodiek:
agree with @beatrix-bitrot  on all counts as someone who uses mutes maybe more than the average user

nightpool:
just to be clear, `muting?` doesn't inherit from instance-wide silences, right? that would be a really bad surprise/regression for a bunch of people.

@Gargron in broad strokes: muting means that YOU shouldn't see THEM. blocking means that THEY shouldn't see YOU. there are nuances and exceptions to this, but that's the general gist.

kaniini:
What @nightpool says: the interaction could very likely be passive-aggressive (reminding the person muting them that they exist).  So it makes sense that all notifications from that user *except* the ones that mean the muted user is redistributing your content to others (boosts and possibly subtoots).

(Why subtoots?  Well, a less than pleasant server seems to enjoy using subtoots to coordinate targeted harassment, but I don't want to get into that one here.)

kaniini:
To further clarify why subtoots explicitly: it is nice to have an early warning that you're about to have a bunch of shit flung your way, so you can be prepared for it.

beatrix-bitrot:
Worry not! Other stuff is unchanged. Only favourites are affected by this.

kaniini:
Sure.  I was just trying to help define what the safety boundaries in a social media setting should be :)

ykzts (approved):
I don't think it's bad.

PR #3258, opened 2017-05-23
Cover HomeController more

akihikodaki:


spec/controllers/home_controller_spec.rb
```diff
@@ -4,9 +4,60 @@
...
+    context 'when signed in' do
+      let(:user) { Fabricate(:user) }
+      subject do
+        sign_in(user)
+        get :index
+      end
+
+      it 'assigns @body_classes' do
+        subject
+        expect(assigns(:body_classes)).to eq 'app-body'
```
mjankowski: These things which are just doing assertions against `assigns` are not super useful, and these methods were removed from rails a while ago (they are here only via the rails-controller-testing gem now), see rationale here: https://github.com/rails/rails/issues/18950

I think my preference would be that we eventually remove that gem, and improve either feature specs or request specs as is relevant in each case.

Gargron: I have been merging these (not having heard of the rationale in that issue) so at this point it would inconsistent to reject this one. I suppose it's not horrible if we merge it and then refactor it to be better tests later? Coverage is coverage, even if it covers a bit too much detail.

mjankowski: Yeah, that's fine -- just wanted to make note of it.

Similarly - I think there's a bunch of stuff in controller specs (a lot of which I've personally added) which would be better placed in either request or feature specs.

I've been prioritizing hitting C0 coverage first, and then ongoing refactors can come after that.

I have no objection to continuing to merge these things.

spec/controllers/home_controller_spec.rb
```diff
@@ -4,9 +4,60 @@
...
+      end
+
+      it 'assigns @token' do
+        Doorkeeper.configure { access_token_expires_in 42 }
+        app = Doorkeeper::Application.create!(name: 'Web', superapp: true, redirect_uri: Doorkeeper.configuration.native_redirect_uri)
```
Gargron: I think this field might be redirect_uris? Tests are failing

Gargron (changes requested):

spec/controllers/home_controller_spec.rb
akihikodaki: > I think my preference would be that we eventually remove that gem, and improve either feature specs or request specs as is relevant in each case.

In my opinion, most things to be "improved" is in the implementation, not the spec.

> But it's worth asking yourself why your controller is so complicated that
sharing it without a view really makes sense. To me it would indicate that
you should be extracting logic into models or concerns.

https://github.com/rails/rails/issues/18950#issuecomment-114613479

If we do such extractions, it would make sense to remove `assigns`, etc. It is, however, out of scope for the series of pull requests. Using the gem is still legitimate for the situation.

By the way, if you are for extracting logic to concerns, why not open an issue and clarify the direction? I think it is better to have such important decisions earlier, whichever direction we go.

> Coverage is coverage, even if it covers a bit too much detail.

I believe that is not too much. The problem is where it should be tested; the interface between models/concerns and controllers, or one between controllers and views.

Gargron (approved):

PR #3269, opened 2017-05-23
Use node instead of babel-node for streaming

nolanlawson:
The `babel-node` docs specify it is [not meant for production use](https://babeljs.io/docs/usage/cli/#babel-node), and the fix is pretty straightforward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build:development": "NODE_ENV=development yarn webpack -- --config config/webpack/development.js",
     "build:production": "NODE_ENV=production yarn webpack -- --config config/webpack/production.js",
     "manage:translations": "node ./config/webpack/translationRunner.js",
-    "start": "babel-node ./streaming/index.js",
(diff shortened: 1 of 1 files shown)


Gargron:
The way I understood babel-node docs, it's not recommended because the code is compiled before running, which is slower than code that doesn't need to be compiled. However, is it actually slower during runtime? Or is it just the compilation delay (as I understood it), in which case this change is changing nothing?

nolanlawson:
In terms of startup cost it's the same, but the main difference is that as
they say they keep an in-memory cache as they run the code, so basically
you have all [this overhead](
https://github.com/babel/babel/blob/master/packages/babel-cli/src/_babel-node.js)
while you're running babel-node. Also more than anything I would just like
to take their word for it that it's not meant for production. :)

-----
Nolan Lawson
http://nolanlawson.com

Gargron (approved):

PR #3278, opened 2017-05-24
Execute PushUpdateWorker only for accounts who uses StreamingAPI just now. 

clworld:
Executing PushUpdateWorker for accounts who aren't using StreamingAPI may waste resources.
This PR add 3 changes to avoid this.
1. Let streaming/index.js maintain redis keys which indicates who uses StreamingAPI now.
2. Don't call PushUpdateWorker when the accounts aren't using StreamingAPI now.
3. (optional) Add StreamingAPI endpoints which output notification only. (for application like tusky-api)

I add StreamingAPI endpoints because all accounts are recognized as using StreamingAPI even if they only need notifications when all accounts uses applications like tusky-api.
Notifications are proceed by NotifyService and I don't touch that. So, notifications are streamed always.
I don't know notifications stream should send delete event or not. (Sent by RemoveStatusService)

PushUpdateWorker does JSON rendering and send JSON clob to Redis. So, reducing PushUpdateWorker calls are effective I think.

I run this on my instance and seems work correctly, but I don't know how effective it is. because I don't have Mastodon instance with lots of users.


Gargron (approved):

Gargron:
(Waiting until after v1.4)

clworld:
Is it waiting 1.5 ?

Gargron (approved):

PR #3284, opened 2017-05-24
Multiple lines of Content Warning area

pfm-eyesightjp:
Added change the Content Warning area from Input tag to Textarea tag.
The purpose is to enrich the user's toot expression more.

cw

### DIFF
--- a/app/javascript/mastodon/components/collapsable.js
+++ b/app/javascript/mastodon/components/collapsable.js
@@ -3,10 +3,10 @@ import Motion from 'react-motion/lib/Motion';
 import spring from 'react-motion/lib/spring';
 import PropTypes from 'prop-types';
 
-const Collapsable = ({ fullHeight, isVisible, children }) => (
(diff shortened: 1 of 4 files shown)


pfm-eyesightjp:
Travis CI has a problem?
My repository failed the first time and passed the second time.
I have not fixed anything...
https://travis-ci.org/pfm-eyesightjp/mastodon

akihikodaki (changes requested):
CW is different from the body by nature and has several different features. For example, hashtag is not functioning in CW.
Is there any reason to change _only_ the size of input box, rather than to align the behavior of CW to the body completely or to keep as it is?

pfm-eyesightjp:
As it stands, CW can only be entered on one line.
The user can not write much description about the body including the spoiler.

The greater embraces the less.

Gargron:
We would like to not encourage people to write long content warnings. There mustn't be confusion about the equality of content warning and content - they are not equal, content warning has no formatting features. No mentions, hashtags, or links. So I think it should stay a single line input for clarity.

PR #3289, opened 2017-05-25
Go to root (home) after login in single user mode

knu:
In single user mode, visitors are redirected to the single user's profile page.  So, if you are the owner without a session, you start from that page, click the login button and authenticate yourself expecting you'll soon get started with the home page, but in reality you'll get redirected back to where you started from -- your own profile page.

This fixes the behavior by redirecting you home after login if you have started from your own profile page.

### DIFF
--- a/app/controllers/auth/sessions_controller.rb
+++ b/app/controllers/auth/sessions_controller.rb
@@ -35,10 +35,10 @@ class Auth::SessionsController < Devise::SessionsController
     params.require(:user).permit(:email, :password, :otp_attempt)
   end
 
-  def after_sign_in_path_for(_resource)
+  def after_sign_in_path_for(resource)
     last_url = stored_location_for(:user)
 
-    if [about_path].include?(last_url)
+    if home_paths(resource).include?(last_url)
       root_path
     else
       last_url || root_path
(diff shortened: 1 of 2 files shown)


app/controllers/auth/sessions_controller.rb
```diff
@@ -35,10 +35,13 @@ def user_params
...
 
-  def after_sign_in_path_for(_resource)
+  def after_sign_in_path_for(resource)
     last_url = stored_location_for(:user)
 
-    if [about_path].include?(last_url)
+    if [about_path].include?(last_url) ||
+       (single_user_mode? &&
+        resource.is_a?(User) &&
+        last_url == short_account_path(username: resource.account))
```
ykzts: can you use private method for conditions?

ykzts (changes requested):

app/controllers/auth/sessions_controller.rb
knu: OK.  Do you want the whole condition set to be factored out or just the part after `||` which I added?

ykzts: I think that the whole is desirable.

knu:
@ykzts It didn't seem very clean to me to extract the whole condition set which actually included two variables (`last_url`<sup>[*](#fn1)</sup> and `resource`), so I made a method to take `resource` and return a list of paths from where user should be redirected to `root_path`.

<a name="fn1"></a>*: `stored_location_for()` is a destructive method you shouldn't call many times, so you have to pass and reuse it.

Gargron (approved):

ykzts (approved):

knu:
Thanks!

PR #3291, opened 2017-05-25
Name the Webpack chunks

nolanlawson:
There's now a [magic comment](https://webpack.js.org/guides/code-splitting-async/#chunk-names) we can add to name asynchronous Webpack chunks. This just makes it easier to debug; no more chunks with names like `0-*.js`, `1-*.js`, etc.

Before:

screenshot 2017-05-24 18 21 34

After:

screenshot 2017-05-24 18 20 04


ykzts (approved):
good!

PR #3295, opened 2017-05-25
[nanobox] Adjustments for Nanobox development

danhunsaker:
Because Nanobox doesn't run data components in the same container as the code, there are a few tweaks that need to be made in the configuration to get WebPack to work properly in development mode.

The same differences lead to needing to use `DATABASE_URL` by default in the `.env` file for Rails to work correctly.

Limitations of our `.env` loader for Node.js mean the `.env` file needs to be compiled everywhere in order to work, so we compile it in development, now, too. Also, all the `.env.production` tweaks have been consolidated into a single command.

Finally, since Nanobox actually creates the database when it sets up the database server, using the existence of the database alone to determine whether to migrate or setup is insufficient. So we add a condition to `rake db:migrate:setup` to check whether any migrations have run - if the database doesn't exist yet, `db:setup` will be called; if it does, but no migrations have been run, `db:migrate` and `db:seed` are called instead (the same basic idea as what `db:setup` does, but it skips `db:create`, which will only cause problems with an existing DB); otherwise, only `db:migrate` is called.

None of these changes should affect development, and all are designed not to interfere with existing behaviors in other environments.

### DIFF
--- a/config/webpack/configuration.js
+++ b/config/webpack/configuration.js
@@ -12,7 +12,7 @@ const devServer = safeLoad(readFileSync(join(configPath, 'development.server.yml
 
 // Compute public path based on environment and CDN_HOST in production
 const ifHasCDN = env.CDN_HOST !== undefined && env.NODE_ENV === 'production';
-const devServerUrl = `http://${devServer.host}:${devServer.port}/${paths.entry}/`;
+const devServerUrl = `http://${env.LOCAL_DOMAIN || devServer.host}:${devServer.port}/${paths.entry}/`;
 const publicUrl = ifHasCDN ? `${env.CDN_HOST}/${paths.entry}/` : `/${paths.entry}/`;
 const publicPath = env.NODE_ENV !== 'production' ? devServerUrl : publicUrl;
 
--- a/.env.nanobox
+++ b/.env.nanobox
@@ -11,6 +11,8 @@ DB_NAME=gonano
 DB_PASS=$DATA_DB_PASS
 DB_PORT=5432
 
+DATABASE_URL=postgresql://$DATA_DB_USER:$DATA_DB_PASS@$DATA_DB_HOST/gonano
+
 # Federation
 # Note: Changing LOCAL_DOMAIN or LOCAL_HTTPS at a later time will cause unwanted side effects.
 # LOCAL_DOMAIN should *NOT* contain the protocol part of the domain e.g https://example.com.
--- a/boxfile.yml
+++ b/boxfile.yml
@@ -31,7 +31,7 @@ run.config:
     - yarn.lock
 
   extra_steps:
-    - cp .env.nanobox .env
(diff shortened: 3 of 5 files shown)


config/webpack/configuration.js
```diff
@@ -12,7 +12,7 @@ const devServer = safeLoad(readFileSync(join(configPath, 'development.server.yml
 
 // Compute public path based on environment and CDN_HOST in production
 const ifHasCDN = env.CDN_HOST !== undefined && env.NODE_ENV === 'production';
-const devServerUrl = `http://${devServer.host}:${devServer.port}/${paths.entry}/`;
+const devServerUrl = `http://${env.LOCAL_DOMAIN || devServer.host}:${devServer.port}/${paths.entry}/`;
```
akihikodaki: Could we create `development.server.yml` for Nanobox as we do for `.env`, or override `NODE_ENV` at runtime?
I do not think it is good to have an extra environment variable for a specific environment while we have a configuration file for the same purpose.

akihikodaki (changes requested):
Thank you for your contribution.
I have a question about your design decision. Could you answer for it?

config/webpack/configuration.js
danhunsaker: All the changes made in this PR are to facilitate running a development/testing environment across multiple servers/containers, rather than entirely on `localhost` directly. No new options were introduced, only adapted to use cases they weren't being used in before. I'll address your specific points one at a time, in order:

- **`development.server.yml`:** Since this is a file that is checked into Git, we'd have to replace it every time the code is updated upstream, and Git would complain about it every time, too. It's extra overhead that we don't have with `.env`.

- **`NODE_ENV` (and `RAILS_ENV`):** I considered adding another environment (`nanobox-dev` or some such) but decided that would quickly get out of hand with all sorts of additional environments being desired all over the map (including, in this case, at least the addition of `nanobox-test`). This feels like opening the door to some things we really don't want to have, not to mention needing code changes in a huge number of places throughout the codebase to account for the alternative environments everywhere they're consulted. I'd just set these to `production` if WebPack would continue working properly that way.

- **Extra environment variable:** This is actually not a new env var at all - it's the same one used in production to determine the domain used anywhere Mastodon needs to know it. It gives us an easy mechanism in development to differentiate between a standard single-server dev environment, and a more production-like multi-container dev environment. Single-server environments simply continue not setting it, while multi-server setups have the option of setting it (just as they would in production, albeit to a development value) to work properly in that setup.

I'm open to alternative approaches, but this is the approach with the least impact on other parts of the codebase, among the ones I could think of.

akihikodaki: > this is the approach with the least impact on other parts of the codebase, among the ones I could think of.

Thank you for clarifying. I agree to the conclusion.

akihikodaki (approved):

PR #3297, opened 2017-05-25
Add improved performance marks for development mode

nolanlawson:
All this does is add performance marks in development mode, to make it easier to debug JavaScript runtime performance. On production builds, it adds nothing.

I added
 * [react-addons-perf](https://facebook.github.io/react/docs/perf.html): React performance marks, i.e. tells us how long it takes each component to render (this was already in the build but wasn't doing anything because we needed to call `start()`)
 * [mark-loader](https://github.com/statianzo/mark-loader): Webpack performance marks, i.e. tells us how long it takes each module to load
  * a custom measure for the `main()` function using [marky](https://github.com/nolanlawson/marky)

This is what it looks like in the Chrome Dev Tools. Here's page load:

screenshot 2017-05-24 22 25 48

And here's scrolling:

screenshot 2017-05-24 22 42 42

The idea is that this gives us better tools to analyze runtime performance, because otherwise the React/Webpack stack traces are pretty opaque.

### DIFF
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -1,3 +1,5 @@
+const perf = require('./performance');
+
 // allow override variables here
 require.context('../../assets/stylesheets/', false, /variables.*\.scss$/);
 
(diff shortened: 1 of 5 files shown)


ykzts (approved):

ykzts:
@Gargron I neglected to confirm with Firefox. Sorry.

noplanman:
@Gargron I'm getting this exact error with the latest docker image (v.2.5.2 at this moment).

```
DOMException [SyntaxError: "An invalid or illegal string was specified"
code: 12
nsresult: 0x8053000c
```

Is there any fix for this?

nightpool:
this is a very old PR. whatever problem you're having, it's almost
certainly unrelated

Issue #251, opened 2016-11-24
Blocked users' toots still appear on the profiles of users who retoot them


Skirmisher:
Reworded: The posts on the profile page of a user are not filtered for blocked users, so retoots of them still appear. (Yes, I'm going to keep using that word.) This affects the home-interface user view and dedicated profile page view. (As a side note, those feel like different things, but it's hard to refer to them separately--what's the deal there?)

Thanks to @neuroghast@mastodon.social for pointing this out!

grimedango:
Seems like the filtering needs to be applied to those views, yes. This would be great to have implemented (if there isnt anything in particular stopping it from being implemented) cuz its a problem that still plagues other platforms.

hoodiek:
i just confirmed this issue still exists

ineffyble:
There has been some commits since the 16th in this area. Could someone verify if this issue is solved?

Gargron:
Nope, we're still not filtering on that API method atm

wxcafe:
this is still a thing.

Cassolotl:
I just saw and accidentally BOOSTED a toot by someone I block, in someone else's profile. This should absolutely NOT be possible!

nynhex:
Is this still reproducible on master?  I can bootstrap and try to reproduce and submit a PR.

Cassolotl:
@nynhex It's still happening on mastodon.social!

trwnh:
Ins't this the difference between "blocking" and "muting" someone?

Block = "I don't want you to see anything from me."
Mute = "I don't want to see anything from you."

Cassolotl:
@trwnh When you block someone, it *also* mutes them.

trwnh:
@Cassolotl so then, do toots from *muted* users appear when boosted? Because you can block and mute someone separately, and I have some people blocked but not muted. AFAIK I've always had to explicitly do both actions.

Cassolotl:
@trwnh I don't know, but that's okay because this particular issue/bug report is for blocked people, and I shouldn't have to mute AND block to stop seeing someone's toots anywhere.

Cassolotl:
Instance mod of vulpine.club said I could pass [this](https://vulpine.club/@mxsparks/99750775765519966) along:

>if i'm blocked by someone, it seems like i shouldn't be able to see their toots when other people boost them, except that I can

There are a whole bunch of replies from people who have given up on blocks working like they are supposed to. :S

nightpool:
[...]

PR #3300, opened 2017-05-25
Don't stream toots from users who have blocked the recipient user

unarist:
This filter was already applied on the `/api/v1/timelines/public` API, but not yet for the Streaming API.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -245,7 +245,7 @@ if (cluster.isMaster) {
           }
 
           const unpackedPayload  = JSON.parse(payload);
-          const targetAccountIds = [unpackedPayload.account.id].concat(unpackedPayload.mentions.map(item => item.id)).concat(unpackedPayload.reblog ? [unpackedPayload.reblog.account.id] : []);
+          const targetAccountIds = [unpackedPayload.account.id].concat(unpackedPayload.mentions.map(item => item.id));
           const accountDomain    = unpackedPayload.account.acct.split('@')[1];
 
           const queries = [
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -249,7 +249,7 @@ if (cluster.isMaster) {
           const accountDomain    = unpackedPayload.account.acct.split('@')[1];
 
           const queries = [
-            client.query(`SELECT 1 FROM blocks WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)}) UNION SELECT 1 FROM mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 1)})`, [req.accountId].concat(targetAccountIds)),
+            client.query(`SELECT 1 FROM blocks WHERE (account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 2)})) OR (account_id = $2 AND target_account_id = $1) UNION SELECT 1 FROM mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 2)})`, [req.accountId, unpackedPayload.account.id].concat(targetAccountIds)),
           ];
 
           if (accountDomain) {


clworld:
FYI: I checked filtering rule of each API about blocking/muting.
1. Streaming API for public, local, hashtag timeline
  - Filters stattuses receiver (blocks or mutes) sender, mentioned account or boosted account. (streaming/index.js)
2. REST API for public, local, hashtag timeline
  - Filters statuses receiver (blocks or mutes) sender or sender blocks receiver. (Status#not_excluded_by_account, Account#excluded_from_timeline_account_ids)
  - Lack of rule about mentioned account and boosted account may performance reason I think. (It needs join 2 additional tables)
3. Home timeline (Streaming API and REST API)
  - Filters statuses receiver mutes sender or boosted account. (FeedManager#filter_from_home?)
  - Filters statuses receiver blocks mentioned account or boosted account. (FeedManager#filter_from_home?)
  - Lack of rule blocked by sender and blocks sender is not a problem. I guess.

This PR fixes difference between 1 and 2 except about mentioned/boosted accout.

unarist:
Thanks, I've made a table.

```
[x] filtered
[ ] not filtered
[-] never appears regardless of relationships

1. Public timelines (Streaming)
+----------------------------------------------------+
|            | sender | booster | author | mentioned |
+------------+--------+---------+--------+-----------+
| blocks     |    x   |    x    |    x   |     x     |
+------------+--------+---------+--------+-----------+
| mutes      |    x   |    x    |    x   |     x     |
+------------+--------+---------+--------+-----------+
| blocked by |        |         |        |           |
+------------+--------+---------+--------+-----------+

2. Public timelines (REST)
+----------------------------------------------------+
|            | sender | booster | author | mentioned |
+------------+--------+---------+--------+-----------+
| blocks     |    x   |    -    |    -   |           |
+------------+--------+---------+--------+-----------+
| mutes      |    x   |    -    |    -   |           |
+------------+--------+---------+--------+-----------+
| blocked by |    x   |    -    |    -   |           |
+------------+--------+---------+--------+-----------+

3. Home timeline (Streaming/REST)
+----------------------------------------------------+
|            | sender | booster | author | mentioned |
+------------+--------+---------+--------+-----------+
| blocks     |    -   |    -    |    x   |     x     |
+------------+--------+---------+--------+-----------+
| mutes      |    x   |    x    |    x   |           |
+------------+--------+---------+--------+-----------+
| blocked by |    -   |    -    |    x   |           |
+------------+--------+---------+--------+-----------+

4. User timeline (REST)
(sender/booster means seeing their sent/boosted statuses on their user page)
+----------------------------------------------------+
|            | sender | booster | author | mentioned |
+------------+--------+---------+--------+-----------+
| blocks     |        |         |        |           |
+------------+--------+---------+--------+-----------+
| mutes      |        |         |        |           |
+------------+--------+---------+--------+-----------+
| blocked by |    x   |    x    |        |           |
+------------+--------+---------+--------+-----------+
```

Then I think:

- If receiver is blocked by author/booster, the status shouldn't be visible even on Streaming API.
(This PR intended to fix this, but I forgot about original author of boosted statuses.)
- Those boosted status also shouldn't be visible on someone's user timeline (#251)
- Filtering by mentioned users of the status should be unified especially on both APIs of public timelines. However, I don't know those statuses should be filtered or not.

Maybe I should open a discussion on elsewhere...

unarist:
> (This PR intended to fix this, but I forgot about original author of boosted statuses.)

Probably, I don't have to handle boosts because public timelines doesn't include boosted statuses, and existent `.concat(unpackedPayload.reblog ? [unpackedPayload.reblog.account.id] : []` part may also be removable.

Gargron:
>Probably, I don't have to handle boosts because public timelines doesn't include boosted statuses

Indeed, boosts aren't even sent to the streaming API (e.g. https://github.com/tootsuite/mastodon/blob/master/app/services/fan_out_on_write_service.rb#L17)

akihikodaki (approved):

Issue #3272, opened 2017-05-24
Error when following other local users (1.4rc3)


Jehops:
Hello,

Clicking the link to follow another local user results in the following console output and the web interface displays "Internal Server Error".  This is a new instance running 1.4rc3.  It doesn't display immediately, but eventually I will see that I am actually following the user.

```
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] method=POST path=/api/v1/accounts/3/follow format=html controller=Api::V1::AccountsController action=follow status=500 error='ArgumentError: wrong number of arguments (given 1, expected 0)' duration=102.19 view=0.00 db=
80.90
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd]
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] ArgumentError (wrong number of arguments (given 1, expected 0)):
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd]
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/models/notification.rb:52:in `activity'
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/models/notification.rb:99:in `set_from_account'
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/services/notify_service.rb:7:in `call'
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/services/follow_service.rb:41:in `direct_follow'
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/services/follow_service.rb:18:in `call'
[a226ab07-fc4e-46e8-9ba5-8adbae3cc1bd] app/controllers/api/v1/accounts_controller.rb:72:in `follow'
```

Jehops:
Some background.  I am working on packaging mastodon on FreeBSD.  When testing the package all the dependent gem versions conformed to the Gemfile, but some were ahead of what is specified in the Gemfile.lock.  When I downgraded lots of them this problem went away.

Gargron:
Is this fixed by #3305?

Jehops:
It seems to be.  When I revert back to depending on rails 5.1 and pull from commit b89ab7e, the error no longer appears.  Shall I close the issue?

PR #3305, opened 2017-05-25
Remove arguments in association reader

ykzts:
ref https://github.com/rails/rails/commit/09cac8c67afdc4b2a1c6ae07931ddc082629b277


ykzts:
It may be the cause of #3272.

Gargron (approved):

PR #3314, opened 2017-05-25
Fix #2922 - Change how customizations are saved

### LINKED ISSUE #2922: Regression : some colors aren't customizable anymore

Gargron:
- Files are now present in the git repository but .gitignore'd
- All loaded through Sass, so variable re-definitions work as expected

You will need to move your code out of `app/assets/stylesheets/custom.scss` into `app/javascript/styles/customizations.scss` and/or `app/javascript/styles/custom_variables.scss`, your code also no longer needs to explicitly re-import application SCSS, as it simply precedes yours

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ yarn-debug.log
 # Ignore Docker option files
 docker-compose.override.yml
 
+app/javascript/styles/custom_variables.scss
+app/javascript/styles/customizations.scss
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -1,8 +1,5 @@
 const perf = require('./performance');
 
-// allow override variables here
-require.context('../../assets/stylesheets/', false, /variables.*\.scss$/);
-
 // import default stylesheet with variables
 require('font-awesome/css/font-awesome.css');
 require('../styles/application.scss');
@@ -22,9 +19,6 @@ function main() {
   const ReactDOM = require('react-dom');
 
   require.context('../images/', true);
-
-  // import customization styles
-  require.context('../../assets/stylesheets/', false, /custom.*\.scss$/);
 
   onDomContentLoaded(() => {
     const mountNode = document.getElementById('mastodon');
--- a/app/javascript/styles/application.scss
+++ b/app/javascript/styles/application.scss
@@ -1,5 +1,6 @@
 @import 'mixins';
 @import 'variables';
+@import 'custom_variables';
 @import 'fonts/roboto';
 @import 'fonts/roboto-mono';
 @import 'fonts/montserrat';
@@ -19,3 +20,4 @@
 @import 'tables';
 @import 'admin';
 @import 'rtl';
+@import 'customizations';
--- a/app/javascript/styles/custom_variables.scss
+++ b/app/javascript/styles/custom_variables.scss
@@ -1,0 +1,1 @@
+// This file allows you to change variable definitions
--- a/app/javascript/styles/customizations.scss
+++ b/app/javascript/styles/customizations.scss
@@ -1,0 +1,1 @@
+// Your own custom SCSS here


nightpool:
for the record, gitignoring files that are already in the repository has no effect—gitignore only changes what files are **added** to the repository. files that are in the index and have changes will shown as modified no matter what is in gitignore

Gargron:
@nightpool Well, so much for all the people on Mastodon unilaterally recommending me to just "add it and then gitignore" -_-;

nightpool:
from a admin perspective, this works fine for us (cybre.space). We might want to ask some of the more total-conversion instances though (tagging @walfie @MasterDalK)

nightpool:
also the changing of the variable names is going to be very confusing to admins, so if we include a new file as part of this, we should also include @nolanlawson's map of new to old variables:

```scss 
$ui-base-color: $color1;
$ui-secondary-color: $color2;
$ui-primary-color: $color3;
$ui-highlight-color: $color4;
$base-border-color: $color5;
$simple-background-color: $color5;
$primary-text-color: $color5;
$error-value-color: $color6;
$valid-value-color: $color7;
$base-shadow-color: $color8;
$base-overlay-background: $color8;
```

esetomo:
custom_variables.scss must be import before variables.scss.

Gargron:
>what was the problem with the old (custom.scss) way? I would consider this way "less elegant" then custom.scss

It was done via sprockets, using a helper that checked if an asset exists. There is no equivalent for current setup

Gargron:
>custom_variables.scss must be import before variables.scss.

Are you.. sure? I have tested this and overriding variables definitely works this way.

esetomo:
"default!" of scss defines variable only when it is not defined,  it is not an overwrite permission setting.

walfie:
> We might want to ask some of the more total-conversion instances though (tagging ＠walfie ＠MasterDalK)

our instance (kirakiratter) doesn't actually use custom.scss, so this change wouldn't really affect us. 

we have our own weird css customization setup (which I don't think works with v1.4, so I'll need to change it, but that's a separate instance-specific issue)

esetomo:
There is no problem at present.
The problem occurs in the following cases.

default.scss:
$foo-color: 'default color' !default;
$foo-color-alt: lighten($foo-color, 10%) !default;

custom.scss:
$foo-font: 'custom color'

When import order default.scss -> custom.scss, the default value will be used for calculating $foo-color-alt and it will not be a custom value.

akihikodaki:
`.gitignore` does not work for git version 2.13.0.

akihikodaki (changes requested):
I think those SCSS should be tracked because they should be open-sourced with respect for AGPL.

To avoid the application of AGPL, they should be in an _aggregate_:
> A compilation of a covered work with other separate and independent works, which are not by their nature extensions of the covered work, and which are not combined with it such as to form a larger program, in or on a volume of a storage or distribution medium, is called an "aggregate" if the compilation and its resulting copyright are not used to limit the access or legal rights of the compilation's users beyond what the individual works permit. Inclusion of a covered work in an aggregate does not cause this License to apply to the other parts of the aggregate.

The compiled output is an _extension of the covered work_, and is _combined with it such as to form a larger program_ and thus is not an _aggregate_.

We have the following options to solve the issue.
* Drop `.gitignore` and mandate to release the SCSSes.
* Add an exception for SCSS customization, as like [linking exception](https://www.gnu.org/licenses/gpl-faq.html.en#GPLIncompatibleLibs).

I think it is fine to mandate to open-source, but we should ask other administrators for their opinions.

Chronister:
instance admins maintaining custom.scss additions can manually force-add it to the git repository. not sure that's a huge issue.

However, looking ahead, perhaps the problem can be solved another way. If the instance itself is aware of the contents of the scss file a la the /about/more page contents, then it can make that file available at a public address. Then, regardless of whether instance's fork of the code contains the custom.scss itself, it's still publicly available. This is well beyond the scope of this PR, but I think it's a direction worth exploring (see [my discourse post on the topic](https://discourse.joinmastodon.org/t/making-customization-easier/229)

akihikodaki:
> instance admins maintaining custom.scss additions in their own forks of mastodon can manually force-add it to their forked git repository.

That is right. But does `.gitignore` make sense if almost all admins force-add their SCSSes?

Naouak:
The way customization is handled in 1.4 is a huge regression as it's no longer possible to not load application.scss.

Some instances do a full rewrite of the css.

nightpool:
@akihikodaki @Chronister again, these files are already in the repository, so gitignore has no effect.

nightpool:
Also, @akihikodaki I think that your claim that because the files are under gitignore, they're not covered under the AGPL is kind of ludicrous. The AGPL doesn't care about your version control system.

akihikodaki:
@nightpool 
> The AGPL doesn't care about your version control system.

No, of course. I mean they _are_ covered by AGPL because they combined with original files in a way just not concatenating. So I believe `.gitignore` should be dropped.

jpfox:
Is this commit present in 1.4rc4 ?

Gargron:
@jpfox No, it's not.

I am going to try something different

PR #3315, opened 2017-05-25
Fix Firefox issue with performance marks

nolanlawson:
Fixes this issue:

    DOMException [SyntaxError: "An invalid or illegal string was specified"
    code: 12

For those curious about the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1331135


Gargron (approved):

Gargron:
Resolves #3309

PR #3322, opened 2017-05-26
Fix "Cannot read property" on missing status

unarist:
I've found this issue when I open UserA's status which replies to UserB whom I blocked/muted.

Properties I've removed in here were added with lazy loading using IntersectionObserver (8e4d1cba), but those statuses are not need to be tracked anyway because it will be rendered as only empty div.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -90,7 +90,7 @@ class Status extends ImmutablePureComponent {
     const { isHidden } = this.state;
 
     if (status === null) {
-      return <div ref={this.handleRef} data-id={status.get('id')} />;
+      return null;
     }
 
     if (isIntersecting === false && isHidden) {


ykzts (approved):

ykzts:
cc @sorin-davidoi

sorin-davidoi:
Shouldn't `null` be returned then to avoid rendering an empty div?

sorin-davidoi:
Or better yet, avoid mounting the component at all by filtering out the `null` statuses in `status_list`?

unarist:
> Shouldn't null be returned then to avoid rendering an empty div?

I've reverted to before that commit, but that seems good.

> Or better yet, avoid mounting the component at all by filtering out the `null` statuses in `status_list`?

Since `status_list` has only `status_id` now, it's needed to get each statuses from the state first.

ykzts (approved):

Gargron (approved):

PR #3329, opened 2017-05-26
Cover Admin::DomainBlocksController more

akihikodaki:
Also domain_block fabricator now sets unique domains


spec/controllers/admin/domain_blocks_controller_spec.rb
```diff
@@ -8,17 +8,23 @@
   end
 
   describe 'GET #index' do
-    it 'returns http success' do
-      get :index
+    it 'renders domain blocks' do
+      42.times { Fabricate(:domain_block) }
```
mjankowski: Instead of creating so many records, can you modify the limit value constant to force a page 2?

mjankowski (dismissed):

akihikodaki:
821ca0e85551ab8d03fe92d2f31f2eae1cb22d3d limits entries per page.

Gargron (approved):

spec/controllers/admin/domain_blocks_controller_spec.rb
```diff
@@ -8,17 +8,23 @@
...
 
   describe 'GET #index' do
-    it 'returns http success' do
-      get :index
+    it 'renders domain blocks' do
+      2.times { Fabricate(:domain_block) }
+
+      get :index, params: { page: 2 }
 
+      assigned = assigns(:domain_blocks).per(1)
```
mjankowski: Putting this `per` in the spec seems weird ... b/c the next line about `count` eq 1 is now just checking the result of something done right here, not done in the controller.

You could set `DomainBlock.paginates_per = 1` in an `around` here (and then set back after spec), which seems better?

spec/controllers/admin/domain_blocks_controller_spec.rb
```diff
@@ -33,13 +39,25 @@
...
 
   describe 'POST #create' do
-    it 'blocks the domain' do
+    it 'blocks the domain when succeeded to save' do
       allow(DomainBlockWorker).to receive(:perform_async).and_return(true)
+
       post :create, params: { domain_block: { domain: 'example.com', severity: 'silence' } }
 
       expect(DomainBlockWorker).to have_received(:perform_async)
+      expect(flash[:notice]).to eq 'Domain block is now being processed'
```
mjankowski: Maybe use `eq I18n.t(...)` instead of hard code?

akihikodaki: I don't think so because:
* The English text is definitive and the test case should be reviewed as well when the text has been changed.
* A test case does not have to be generic.

akihikodaki:
8cf75eed586be6382431be24a408001f3851ee9c applied https://github.com/tootsuite/mastodon/pull/3329#discussion_r118766156.

spec/controllers/admin/domain_blocks_controller_spec.rb
akihikodaki: Please tell me if you have counter arguments since the decision to use `I18n.t` or not affects other similar test cases.

mjankowski: I think my first preference would be to not even check the contents of `flash` in controller specs ... I think they are sort of an implementation detail, and if we care about the words showing up on the page, we should use feature specs.

Until we get to that point, leaving it in here is fine, but I'd prefer to check that it matches `I18n.t()` lookup instead of the string directly ... since that's what we really care that the controller is doing ... finding whatever that value is and putting it in the flash.

akihikodaki: I'm convinced. I'll make a change accordingly.

akihikodaki:
1b9af2b978f84b80371a0013f5c830834df1c56a applied https://github.com/tootsuite/mastodon/pull/3329#discussion_r119355325.

PR #3334, opened 2017-05-26
Introduce react-textarea-autosize instead of using style.height side …

Gargron:
…effects


ykzts (approved):
Looks like a good library

PR #3335, opened 2017-05-26
Updated Polish translation

mkljczk:
to będzie squashed…


Gargron (approved):

Gargron:
Key `admin.are_you_sure` is unused - wrong nesting?

m4sk1n:
Sorry for that, @Gargron…

PR #3336, opened 2017-05-26
Avoid comparing domains when looking for an exact match of a local account

### LINKED ISSUE #3064: account search API ignores specified instance when instance is local
The account search API (`/api/v1/accounts/search`) seems sane for cases where (1) the query isn't in the form `user@instance` and (2) the instance in the query is remote.  But in case (3), when the instance in the query is the local instance, the instance specification is completely ignored.

This leads to bad results like querying for `bar@mastodon.social` (an account that exists) and getting a list that doesn't include `bar@mastodon.social`, and includes many accounts on remote instances.  (In this specific case, that account is included when the query is `bar`, but not included when it's `bar@mastodon.social`.)

I think the API should treat case (3) more like case (2):

- never show users on instances that do not match the query;
- if there is an exact match for the `user` portion of the query, it should be first in the list

* * * *

happycoloredbanana:
This is supposed to fix #3064.

When looking for an exact account match, if the search string is in a `user@domain` form, check whether the domain is local before searching the accounts table.

### DIFF
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -59,7 +59,13 @@ class AccountSearchService < BaseService
   end
 
   def exact_match
-    @_exact_match ||= Account.find_remote(query_username, query_domain)
+    @_exact_match ||= begin
+      if domain_is_local?
+        Account.find_local(query_username)
+      else
+        Account.find_remote(query_username, query_domain)
+      end
+    end
   end
 
   def search_results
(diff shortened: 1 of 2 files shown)


Gargron:
>This is supposed to fix #3064.

Does it work?

happycoloredbanana:
Yes. This is how I tested it:

Created a one-letter local account "i" and many remote accounts with the same letter in `display_name`. Tried `/api/v1/accounts/search?q=i@mastodon.dev&limit=3` with curl and saw that only remote accounts are returned, although there's an exact match for the local account.
After the fix, it returns the exact match first and than - remote accounts.

Gargron:
Perhaps that could be a unit test with those exact actions?

happycoloredbanana:
Added a test.
Cannot figure out how Travis failures are even related.

Gargron (approved):

PR #3341, opened 2017-05-27
Introduce RecentOrderable concern for models

akihikodaki:
`ID`concern has scopes dealing with IDs, namely `recent`, which were common for some models, and `paginate_by_max_id`, which were provided `Paginable` concern and included `recent` scope.

This pull request also includes `ID` to `Subscription`.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -6,12 +6,12 @@ class AccountsController < ApplicationController
   def show
     respond_to do |format|
       format.html do
-        @statuses = @account.statuses.permitted_for(@account, current_account).paginate_by_max_id(20, params[:max_id], params[:since_id])
+        @statuses = @account.statuses.permitted_for(@account, current_account).paginate_by_recent(20, params[:max_id], params[:since_id])
         @statuses = cache_collection(@statuses, Status)
       end
 
       format.atom do
-        @entries = @account.stream_entries.where(hidden: false).with_includes.paginate_by_max_id(20, params[:max_id], params[:since_id])
+        @entries = @account.stream_entries.where(hidden: false).with_includes.paginate_by_recent(20, params[:max_id], params[:since_id])
         render xml: AtomSerializer.render(AtomSerializer.new.feed(@account, @entries.to_a))
       end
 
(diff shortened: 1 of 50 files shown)


app/models/concerns/id.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
-module Paginable
```
Gargron: Why rename this? Old name was very descriptive

akihikodaki: Because `recent` is not always used with pagination. `ID` is a concern for any models with incremental IDs.
I'm considering a better naming...

akihikodaki: After some thinking, I am convinced with an idea that `ID` is a valid name conforming the convention to name concerns after related properties.
Namely, `AccountAvatar`, `AccountHeader`, and `AccountInteractions` are representing properties those concerns deal with. `ID` concern also deals with IDs.

mjankowski (dismissed):
I think `ID` is not a great name for a concern/module here ... that's so generic as to not carry much meaning at all.

I also wouldn't combine the "methods which paginatable things need" module with the "order by id desc" module.

I think I would probably:

- Leave paginable alone
- Call the new module something like `OrderByRecent` or `RecentOrderable` or something more intention-revealing

I agree with the changes which replaces the explicit `order(id: :desc)` with the use of a scope or class method instead.

akihikodaki:
I think `RecentOrderable` is good also for `paginate_by_max_id`, isn't it?

mjankowski:
I'm not sure that "be able to paginate this model" and "be able to order this model by id desc" are necessarily the same feature or necessarily coupled to each other ... even if they happen to be now.

akihikodaki:
> I'm not sure that "be able to paginate this model" and "be able to order this model by id desc" are necessarily the same feature or necessarily coupled to each other... 

That is true. However, the name of scope is `paginate_by_max_id`. What about including `paginate_by_max_id` in `RecentOrderable` rather than including `Paginable` in `RecentOrderable`?

akihikodaki:
04936d73c2747cdae59ec53a901041c65a33db4e introduces `RecentOrderable`.

akihikodaki:
Rebased to d567a382e30b6d6f2b32dc249380e24b8d5f2be7.

PR #3351, opened 2017-05-27
Improve default language decision and spec

akihikodaki:


app/controllers/concerns/localized.rb
```diff
@@ -17,12 +17,12 @@ def set_locale
...
   def default_locale
-    ENV.fetch('DEFAULT_LOCALE') do
-      user_supplied_locale || I18n.default_locale
-    end
-  end
+    env = ENV['DEFAULT_LOCALE']
 
-  def user_supplied_locale
-    http_accept_language.language_region_compatible_from(I18n.available_locales)
+    http_accept_language.preferred_language_from([env]) ||
```
mjankowski: This is a lot of conditionals happening in one method...

What about breaking this out to something like...

`request_locale || env_locale || I18n.default_locale`

The impl of `request_locale` would then be like `preferred_locale || compatible_locale` and those methods would then check both the env and available.

mjankowski (changes requested):
The spec coverage improvements are solid.  One comment about the impl change.

akihikodaki:
062332ae3c30910377ce0e5d7c8b0283dd149e7a applied a change request at https://github.com/tootsuite/mastodon/pull/3351#discussion_r119094822. Sorry for late submission.

app/controllers/concerns/localized.rb
```diff
@@ -17,12 +17,24 @@ def set_locale
...
   end
 
-  def user_supplied_locale
-    http_accept_language.language_region_compatible_from(I18n.available_locales)
+  def env_locale
+    ENV['DEFAULT_LOCALE']
+  end
+
+  def request_locale
+    preferred_locale || http_accept_language
```
mjankowski: Should this be `preferred_locale || compatible_locale` ?

spec/controllers/concerns/localized_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+  before do
+    routes.draw { get 'success' => 'anonymous#success' }
+  end
+
+  shared_examples 'default locale' do
+    around do |example|
+      default_locale = ENV['DEFAULT_LOCALE']
```
mjankowski: We could probably use climate_control gem here as well

mjankowski (dismissed):

akihikodaki:
ba00df7dbfcc8b77cb8272e653a68b5377a2c70d applied https://github.com/tootsuite/mastodon/pull/3351#pullrequestreview-42896370.

Gargron (approved):

PR #3353, opened 2017-05-27
Cover Auth::RegistrationsController more

akihikodaki:


ykzts (approved):

spec/controllers/auth/registrations_controller_spec.rb
```diff
@@ -3,6 +3,45 @@
 RSpec.describe Auth::RegistrationsController, type: :controller do
   render_views
 
+  shared_examples 'checks for enabled registrations' do |path|
+    it 'redirects if it is in single user mode while it is open for registration' do
+      Fabricate(:account)
+      Setting.open_registrations = true
```
Gargron: Because `Setting` is global, changing it might affect other tests that run after, randomly. Storing the original value and restoring might be necessary

Gargron (dismissed):

spec/controllers/auth/registrations_controller_spec.rb
akihikodaki: I agree with that. However there are already similar codes and I followed them. I'll also fix them later.

akihikodaki:
ee60b5127956b1897581b47f4c6b27e189fd953e adds cleaning hooks.

ykzts (approved):

PR #3357, opened 2017-05-27
Fix order of paginated accounts in FollowerDomainsController and spec

akihikodaki:
Unordered pagination could result in unexpected behavior.

### DIFF
--- a/spec/controllers/settings/follower_domains_controller_spec.rb
+++ b/spec/controllers/settings/follower_domains_controller_spec.rb
@@ -5,15 +5,41 @@ describe Settings::FollowerDomainsController do
 
   let(:user) { Fabricate(:user) }
 
-  before do
-    sign_in user, scope: :user
+  shared_examples 'authenticate user' do
+    it 'redirects when not signed in' do
+      is_expected.to redirect_to '/auth/sign_in'
+    end
   end
 
   describe 'GET #show' do
+    subject { get :show, params: { page: 2 } }
+
+    it 'assigns @account' do
+      sign_in user, scope: :user
+      subject
+      expect(assigns(:account)).to eq user.account
+    end
+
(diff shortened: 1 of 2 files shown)


spec/controllers/settings/follower_domains_controller_spec.rb
```diff
@@ -5,32 +5,81 @@
...
+    subject { get :show, params: { page: 2 } }
+
+    it 'assigns @account' do
+      sign_in user, scope: :user
+      subject
+      expect(assigns(:account)).to eq user.account
+    end
+
+    it 'assigns @domains' do
+      followers = 11.times.map do |index|
```
mjankowski: Override pagination threshold instead of creating so many records?

mjankowski (dismissed):

spec/controllers/settings/follower_domains_controller_spec.rb
akihikodaki: It is not good for testing the query... I'm still considering other means, but that is possible as the last resort.

mjankowski: Can we move the huge string of methods in the controller into some scopes and class methods instead? That might make it easier to have a more thorough spec for the query improvement (in a model spec) while keeping the controller spec more focused, and the controller code more readable.

akihikodaki: I'm :+1: for the idea, but I'd like to focus on the fix of the order and spec of the current implementation in this pull request.

akihikodaki:
0b21068efe529eb195a9ac925a4e0cca28932438 overrides `per` as suggested.

unarist:
@mjankowski Have you checked update after your review?

Gargron (approved):

PR #3360, opened 2017-05-27
Improve Settings::ImportsController and spec

akihikodaki:


app/controllers/settings/imports_controller.rb
```diff
@@ -11,24 +11,27 @@ def show
   end
 
   def create
-    @import = Import.new(import_params)
+    if params[:import].nil?
```
mjankowski: Can this whole conditional go in a `before_action` check?

app/controllers/settings/imports_controller.rb
```diff
@@ -11,24 +11,27 @@ def show
...
+      @import = Import.new
+      render :show
+      return
+    end
+
+    @import = Import.new(params[:import].permit(:data, :type))
```
mjankowski: Why move this out of the `import_params` method?

app/controllers/settings/imports_controller.rb
```diff
@@ -11,24 +11,27 @@ def show
...
     @import.account = @account
 
-    if @import.save
-      ImportWorker.perform_async(@import.id)
-      redirect_to settings_import_path, notice: I18n.t('imports.success')
-    else
+    unless @import.save
       render :show
+      return
```
mjankowski: Try to avoid explicit returns.

app/controllers/settings/imports_controller.rb
```diff
@@ -11,24 +11,27 @@ def show
...
     end
+
+    ImportWorker.perform_async(@import.id)
```
mjankowski: This change of pulling these out of the `if @import.save` and down here instead doesn't feel like an improvement..?

mjankowski (changes requested):

app/controllers/settings/imports_controller.rb
akihikodaki: I don't know why it should be done in `before_action`. It is a role of `create` to `render :show`.

app/controllers/settings/imports_controller.rb
akihikodaki: Because the combination of the conditional and the statement is atomic.

app/controllers/settings/imports_controller.rb
akihikodaki: Yes, but I think it is the best solution for this case. Do you have any other idea?

app/controllers/settings/imports_controller.rb
akihikodaki: How could I deal with exception handling then?

mjankowski (changes requested):
I think you can make all of the added specs here pass with a change to one line in the controller:

```
  def import_params
    params.fetch(:import, {}).permit(:data, :type)
  end
```

PR #3365, opened 2017-05-27
Fix IntersectionObserver isIntersecting in Edge

nolanlawson:
It turns out Edge's implementation of IntersectionObserver is [missing the `isIntersecting` value](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/). Luckily we can calculate the same value from `intersectionRatio` and it works in both Chrome and Edge (as well as Firefox/Safari, where a polyfill is used).

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -70,7 +70,9 @@ class StatusList extends ImmutablePureComponent {
         entries.forEach(entry => {
           const statusId = entry.target.getAttribute('data-id');
 
-          state.isIntersecting[statusId] = entry.isIntersecting;
+          // Edge 15 doesn't support isIntersecting, but we can infer it from intersectionRatio
+          // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/
+          state.isIntersecting[statusId] = entry.intersectionRatio > 0;
         });
 
         // isIntersecting is a map of DOM data-id's to booleans (true for


sorin-davidoi:
This was my first approach, but I ran into some issues on Chrome. If I recall correctly, scrolling up and down really fast would result in some toots coming into view but still being "collapsed". Don't have my laptop to test right now. On the other hand this was quite early in the implementation, so it may be possible that the issue was something else.

ykzts (approved):

nolanlawson:
@sorin-davidoi Hm, I'm testing in Chrome 58 on a Mac and I cannot reproduce the issue you describe. I tried both two-finger scrolling and sidebar scrolling, scrolling as fast as I could. Did you see this on a mobile device? Or maybe it was just because of an early implementation?

sorin-davidoi:
It was on desktop on Linux. Will take a look in a few hours.

nolanlawson:
I have a Lubuntu box. I'll give it a shot. What kind of scrolling method? (keyboard/sidebar/mousewheel/etc.)

sorin-davidoi:
Mousewheel.

nolanlawson:
OK, I tested on Lubuntu using Chrome 58 on https://malfunctioning.technology and I cannot reproduce with a mousewheel or sidebar scrolling, no matter how fast I go. I'm scrolling the Federated Timeline since there are a lot of toots. 😃

sorin-davidoi:
Cool 👍, sorry for the trouble.

nolanlawson:
No prob, cross-browser is hard; I am paranoid and like to test everything thoroughly too! :)

nolanlawson:
Just tested in desktop Firefox and Safari as well, and this works. I can tell it works because, as I scroll, the number of DOM elements (`document.querySelectorAll('*').length`) stay about the same or sometimes even go down. :)

Issue #2383, opened 2017-04-24
Make confirmation modals optional


expenses:
I think that the confirmation modals added in #2279 should be turned on by default, but have settings in the options to turn the on and off.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Seems to have been solved by PR but PR didn't auto-close the issue.

PR #3368, opened 2017-05-27
Add preference setting for delete toot modal

jumbosushi:
PR for [#2383](https://github.com/tootsuite/mastodon/issues/2383)

* Set delete_modal preference to true by default
* Does not show confirmation modal if delete_modal is false
* Add ja translation for preference setting page

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -35,6 +35,7 @@ class Settings::PreferencesController < ApplicationController
     params.require(:user).permit(
       :setting_default_privacy,
       :setting_boost_modal,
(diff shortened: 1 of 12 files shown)


jumbosushi:
Couple questions:
- Should deleting toot allow Shift key option like boost to temporary disable modal?
- This PR only covers English and Japanese translations. Does it need more translations?

akihikodaki (approved):

akihikodaki:
> Should deleting toot allow Shift key option like boost to temporary disable modal?

That sounds too dangerous for me.

jumbosushi:
Okay sounds good 😀

Gargron (approved):

PR #3370, opened 2017-05-27
Improve streaming API cluster logging

Gargron:
- Clean up clustering code
- Log process errors as well as process exit

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -206,7 +206,7 @@ const startWorker = (workerId) => {
   };
 
   const errorMiddleware = (err, req, res, next) => {
-    log.error(req.requestId, err);
+    log.error(req.requestId, `${err}`);
     res.writeHead(err.statusCode || 500, { 'Content-Type': 'application/json' });
     res.end(JSON.stringify({ error: err.statusCode ? `${err}` : 'An unexpected error occurred' }));
   };
@@ -398,7 +398,7 @@ const startWorker = (workerId) => {
   });
 
   const onExit = () => {
-    log.error(`Worker ${workerId} exiting, bye bye`);
+    log.info(`Worker ${workerId} exiting, bye bye`);
     server.close();
   };
 
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -240,7 +240,7 @@ const startWorker = (workerId) => {
           const targetAccountIds = [unpackedPayload.account.id].concat(unpackedPayload.mentions.map(item => item.id)).concat(unpackedPayload.reblog ? [unpackedPayload.reblog.account.id] : []);
           const accountDomain    = unpackedPayload.account.acct.split('@')[1];
 
-          if (req.filteredLanguages.indexOf(unpackedPayload.language) !== -1) {
+          if (req.filteredLanguages instanceof Array && req.filteredLanguages.indexOf(unpackedPayload.language) !== -1) {
             log.silly(req.requestId, `Message ${unpackedPayload.id} filtered by language (${unpackedPayload.language})`);
             return;
           }
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -245,6 +245,7 @@ const startWorker = (workerId) => {
 
           if (Array.isArray(req.filteredLanguages) && req.filteredLanguages.includes(unpackedPayload.language)) {
             log.silly(req.requestId, `Message ${unpackedPayload.id} filtered by language (${unpackedPayload.language})`);
+            done();
             return;
           }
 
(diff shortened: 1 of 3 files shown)


Gargron:
cc @geta6

streaming/index.js
```diff
@@ -249,7 +240,7 @@ if (cluster.isMaster) {
           const targetAccountIds = [unpackedPayload.account.id].concat(unpackedPayload.mentions.map(item => item.id)).concat(unpackedPayload.reblog ? [unpackedPayload.reblog.account.id] : []);
           const accountDomain    = unpackedPayload.account.acct.split('@')[1];
 
-          if (req.filteredLanguages.indexOf(unpackedPayload.language) !== -1) {
+          if (req.filteredLanguages instanceof Array && req.filteredLanguages.indexOf(unpackedPayload.language) !== -1) {
```
ykzts: I think that "`Array.isArray(req.filteredLanguages) && req.filteredLanguages.includes(unpackedPayload.language)`" is better.

- [`Array.isArray` (ES 5.1)](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)
- [`Array.prototype.includes` (ES 2016)](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

Both are implemented in Node.js

streaming/index.js
```diff
@@ -16,6 +16,8 @@ dotenv.config({
   path: env === 'production' ? '.env.production' : '.env',
 });
 
+log.level = process.env.LOG_LEVEL || 'verbose';
```
Gargron: Ensure that log level is set at start of executing, rather than only when http/ws server starts listening

streaming/index.js
```diff
@@ -215,9 +209,9 @@ if (cluster.isMaster) {
   };
 
   const errorMiddleware = (err, req, res, next) => {
-    log.error(req.requestId, err);
+    log.error(req.requestId, err.toString());
```
Gargron: This *should* only print the error message, not the whole stack. And it does so on my dev environment. However, in production, it still dumps the entire stack trace into the logs...

streaming/index.js
```diff
@@ -249,8 +243,9 @@ if (cluster.isMaster) {
           const targetAccountIds = [unpackedPayload.account.id].concat(unpackedPayload.mentions.map(item => item.id)).concat(unpackedPayload.reblog ? [unpackedPayload.reblog.account.id] : []);
           const accountDomain    = unpackedPayload.account.acct.split('@')[1];
 
-          if (req.filteredLanguages.indexOf(unpackedPayload.language) !== -1) {
+          if (Array.isArray(req.filteredLanguages) && req.filteredLanguages.includes(unpackedPayload.language)) {
             log.silly(req.requestId, `Message ${unpackedPayload.id} filtered by language (${unpackedPayload.language})`);
+            done();
```
Gargron: This was the reason for streaming API hanging up. I didn't realize we were inside pgPool.connect here, didn't release connection, so the pool eventually ran out of connections

streaming/index.js
```diff
@@ -271,6 +266,7 @@ if (cluster.isMaster) {
 
             transmit();
           }).catch(err => {
+            done();
```
Gargron: This might have contributed to problems too, although I doubt this was the main culprit, now

streaming/index.js
```diff
@@ -309,26 +305,13 @@ if (cluster.isMaster) {
...
-
-      ws.ping();
-    }, 15000);
-
-    return (event, payload) => {
-      if (ws.readyState !== ws.OPEN) {
-        log.error(req.requestId, 'Tried writing to closed socket');
-        return;
-      }
+  const streamToWs = (req, ws) => (event, payload) => {
```
Gargron: Simplified, see below

streaming/index.js
```diff
@@ -372,6 +355,12 @@ if (cluster.isMaster) {
     const token    = location.query.access_token;
     const req      = { requestId: uuid.v4() };
 
+    ws.isAlive = true;
```
Gargron: According to example: <https://github.com/websockets/ws#how-to-detect-and-close-broken-connections>

Gargron:
I believe I have identified the main problem. However, I also changed other code. Perhaps I should extract my fix to a smaller PR?

ykzts:
It seems that there is no problem with the code. I also check the operation in the development environment.

ykzts (approved):
looks good to me

PR #3371, opened 2017-05-27
Simplify isIntersecting in status_list.js

nolanlawson:
This has the same functionality as before, but removes some unnecessary object allocations and adds a comment explaining what's going on.

Essentially the point of the array was to force the component to resolve `shouldComponentUpdate()` to true whenever the `isIntersecting` map changed (because a new array was created). But we can do this more directly by just incrementing an integer, and then there's no need for an extra array.


sorin-davidoi:
A much more elegant approach 👍

ykzts (approved):

PR #3332, opened 2017-05-26
Fix #2922 - Load stylesheet from "custom.css" entrypoint when present

Gargron:
This is pretty much the same way it worked as before, albeit with having to create `app/javascript/packs/custom.js` with `require('../styles/custom.scss')` (or whatever you want really), which will be a blank slate for you to import whatever you want

PR #3373, opened 2017-05-28
Resolve custom application stylesheet with Webpack

akihikodaki:
This implementation is a bit smaller and still has the following benefits:

* No need of `app/javascript/packs/custom.js`
For custom stylesheet, it typically has only `require('../styles/custom.scss')` and is redundant.

* No need to extract vendor stylesheet to another asset
Extracting vendor stylesheet could be forgotten by developers who do not use custom stylesheet.

> This could be fixed later, I am craving the sweet release of release right now

https://github.com/tootsuite/mastodon/pull/3332#event-1099915973

It is fine for me to leave this change for releases later than 1.4.1.


app/javascript/mastodon/main.js
```diff
@@ -2,7 +2,7 @@ const perf = require('./performance');
 
 // import default stylesheet with variables
 require('font-awesome/css/font-awesome.css');
-require('../styles/application.scss');
+require('application-style');
```
nolanlawson: I would maybe name this `mastodon-application-style` so that it's clear this isn't a package on npm. :)

app/views/layouts/application.html.haml
```diff
@@ -18,8 +18,7 @@
         = ' - '
       = title
 
-    = stylesheet_pack_tag 'common', media: 'all'
-    = stylesheet_pack_tag stylesheet_for_layout, media: 'all'
```
nolanlawson: The perf benefit of the `common` CSS file is definitely minimal, but why remove it?

nolanlawson (changes requested):
The `alias` system definitely looks more elegant than the previous system, and I like that this maintains legacy `custom.scss` behavior to an extent (i.e. that you must import `application` from within `custom`), but I'm not clear why we need to remove the `common` stylesheet?

app/views/layouts/application.html.haml
akihikodaki: Because `extract-text-webpack-plugin` bundles all stylesheets required by the entry point, just as did Webpack before #3332.

akihikodaki:
dbb3a9e7004b5fc837e91b166819113e6881b278 renames `application-style` alias to `mastodon-application-style`.

config/webpack/shared.js
```diff
@@ -53,18 +57,16 @@ module.exports = {
           // be loaded together
           return false;
         }
-
-        if (module.resource && /node_modules\/font-awesome/.test(module.resource)) {
-          // extract vendor css into common module
-          return true;
-        }
```
nolanlawson: Ah I see, this was the line that was causing a `common.css` bundle.

app/views/layouts/application.html.haml
nolanlawson: OK, it took me a bit to understand, but I see now that there is no `public.css` vs` application.css` – it's all the same CSS. So it really is pointless to have a `common.css` because all entry points require exactly the same CSS.

nolanlawson (approved):
This looks great to me – more elegant than the previous approach. The only remaining problem I see is that this will break any users who upgraded from the `custom.scss` system pre-1.4.1 to the system for 1.4.1. It may annoy people who are tired of all the change. But it is certainly simpler than the existing system.

Gargron (approved):

akihikodaki:
> It may annoy people who are tired of all the change.

Yes, we should not do breaking change too often. If this is going to be merged, we should be open for other improvements till the next release so that we will not have to break it again in later releases. (Maybe we can ask on Discourse or GitHub issue?)

I have an idea to allow to customize public stylesheet as well. I'd like to discuss about that if we have Discourse topic or GitHub issue.

nolanlawson:
I think this PR should be reviewed by folks like @beatrix-bitrot and @Chronister. Personally I have stopped using `custom.scss` because I found it easier to just modify `variables.scss` directly and then rebase. There has been a lot of churn in how `custom.scss` works, and while the proposed system is better than the old one, it is yet another change. So we should see how folks relying on `custom.scss` feel about it.

Chronister:
I have to rebase every time I apply updates anyway due to changes in other files; git workflow is not really a concern for me.

I'm principally concerned with minimizing the number of steps to actually get a working custom style after I update. To update to 1.4, I had to do the following:
 - Create a new file called custom.js to require my custom.scss
 - Move my custom.scss to the new stylesheet directory
 - Change all uses of "image-url('blah')" to "url('../rel/blah')" (which took me some trial and error to figure out, which was kinda painful under the new asset build times)

I was also unsure whether I still had to import the application.scss because of discussion during the release process about removing it, which added to my trial-and-error time to actually getting everything working.

I don't particularly care how the system works as long as no functionality is lost (I need to be able to change the color variables, and I'd like to have the option of discarding the base site style if I want to and going fully custom). The number of times that the system changes should ideally be minimized, because I don't really want to spend my update time trying to figure out where my custom stylesheet should go. It should also be documented, ideally more visibly and descriptively than it was in 1.4.

This should also be taken with the reservation that the customization system I actually want would ideally be done as much as possible at runtime, [as my suggestion on the discourse describes](https://discourse.joinmastodon.org/t/making-customization-easier/229)

akihikodaki:
> I'm principally concerned with minimizing the number of steps to actually get a working custom style after I update.

What do you think about removing `custom.js`? That is the breaking change in this pull request. Is it acceptable or not?

> This should also be taken with the reservation that the customization system I actually want would ideally be done as much as possible at runtime, [as my suggestion on the discourse describes](https://discourse.joinmastodon.org/t/making-customization-easier/229).

There is no additional obstacle to implement the feature in this pull request as far as I know.

Chronister:
@akihikodaki I haven't done anything with it besides the bare require statement necessary to get the custom.scss working again. If it's just to be removed I don't really mind right now, though I'll admit I'm a little curious what the original decision to do it that way was intending.

PR #3375, opened 2017-05-28
Settings::ProfilesController renders :show if parameters are missing

akihikodaki:


app/controllers/settings/profiles_controller.rb
```diff
@@ -14,20 +14,24 @@ class Settings::ProfilesController < ApplicationController
   def show; end
 
   def update
-    if @account.update(account_params)
-      redirect_to settings_profile_path, notice: I18n.t('generic.changes_saved_msg')
-    else
+    if params[:account].nil?
```
mjankowski: Move to `before_action`

app/controllers/settings/profiles_controller.rb
```diff
@@ -14,20 +14,24 @@ class Settings::ProfilesController < ApplicationController
...
       render :show
+      return
     end
-  end
 
-  private
+    account_params = params[:account].permit(:display_name, :note, :avatar, :header, :locked)
```
mjankowski: Use a method instead of local var

mjankowski (changes requested):

akihikodaki:
4f6bd93e84f82eac8ec2a3c768e40464f475bae3 is expected to solve problems pointed out at https://github.com/tootsuite/mastodon/pull/3375#pullrequestreview-40944109.

app/controllers/settings/profiles_controller.rb
```diff
@@ -14,17 +14,19 @@ class Settings::ProfilesController < ApplicationController
...
+    else
+      redirect_to settings_profile_path, notice: I18n.t('generic.changes_saved_msg')
     end
   end
 
   private
 
   def account_params
-    params.require(:account).permit(:display_name, :note, :avatar, :header, :locked)
+    params[:account]&.permit(:display_name, :note, :avatar, :header, :locked)
```
mjankowski: What are you trying to do here?

Is the concern that `params[:account]` can be nil, and you want to handle that case and the case where it has normal values via strong params?

How about something like: `params.fetch(:account, {}).permit(:display_name, ...)`

akihikodaki: Yes, I would like it to be `nil`.

Gargron:
I do not understand what this PR is fixing. For example, `params.require(:some_key)` can be found throughout all Rails apps...

akihikodaki:
If `account_params` is invalid, `account.update` fails and renders `:show`. However, it returns 400 when missing. That is inconsistent.

mjankowski:
Can you add a failing spec that shows what you are fixing?

akihikodaki:
@mjankowski It is great to see you again. I'll add a spec.

akihikodaki:
5aa126d6339c8507ef80f610a3da692b44c05df2 adds spec.

spec/controllers/settings/profiles_controller_spec.rb
```diff
@@ -23,5 +23,10 @@
       expect(account.reload.display_name).to eq 'New name'
       expect(response).to redirect_to(settings_profile_path)
     end
+
+    it 'renders :show if parameter is missing' do
+      put :update
+      expect(response).to render_template :show
```
mjankowski: What's the thinking behind having this be the response?

I would think that either of these approaches...

- Return a 400 as an error because the params are not valid when it's missing, or;
- Consider the lack of params to be a request for an empty update, which will "succeed", so do nothing and then redirect like any other success

...would make more sense than forcing the show to render ... since you'd have to purposefully do something wrong to create this situation?

akihikodaki: > Return a 400 as an error because the params are not valid when it's missing, or;

Here I'm considering the inconsistency with validation failure in `account.update` is problematic. (I would not like to spec that.) I think it is more convincing to return 400 for validation failure as well if it returns 400 for missing parameter, isn't it?

mjankowski: I think it's fine if they are different, because they are different things.

When the params are missing, it's very similar to a missing route or some other request which can't be fulfilled, so I'd expect a 4xx response (and rails catches the params error and does this, I think?) An exception to this would be if there are multiple ways the route could be valid and different params combination were optional or valid depending on the presence of other ones ... I'd expect something more graceful there.

When the params/route is valid - but the data included in the update is invalid - I'd expect a 200 response with a rendered page that includes a description of the data validation errors.

akihikodaki: > When the params/route is valid - but the data included in the update is invalid - I'd expect a 200 response with a rendered page that includes a description of the data validation errors.

Then why shouldn't it return 400 for invalid parameters while parameter missing does? I decided to return 200 for parameter missing because it does so for invalid parameters, which should have 400 if we respect RFC.

PR #3377, opened 2017-05-28
Improve Settings:PreferencesController and spec

akihikodaki:


mjankowski (changes requested):
Same feedback here re: controller changes as the other few had...

PR #3381, opened 2017-05-28
Lazy load routes and some components

sorin-davidoi:
Reduces `application.js` by ~30% (from `648kB` to `455kB`). There are still some things to address like error handling and perhaps show the loading indicator while fetching the bundles.

http://bl.ocks.org/sorin-davidoi/raw/bcce9c6bb4ae822ea8f26f4a5f6eae60/

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -8,18 +8,11 @@ import ModalContainer from './containers/modal_container';
 import { connect } from 'react-redux';
 import { isMobile } from '../../is_mobile';
 import { debounce } from 'lodash';
+import Immutable from 'immutable';
 import { uploadCompose } from '../../actions/compose';
 import { refreshTimeline } from '../../actions/timelines';
 import { refreshNotifications } from '../../actions/notifications';
 import UploadArea from './components/upload_area';
-
-const staticColumns = { };
-
-const lazyLoadStaticColumns = () => {
-  import(/* webpackChunkName: "features/home_timeline" */ '../home_timeline').then(Component => staticColumns.HomeTimeline = Component.default);
-  import(/* webpackChunkName: "features/compose" */ '../compose').then(Component => staticColumns.Compose = Component.default);
-  import(/* webpackChunkName: "features/notifications" */ '../notifications').then(Component => staticColumns.Notifications = Component.default);
-};
 
 const noOp = () => false;
 
@@ -32,11 +25,16 @@ class UI extends React.PureComponent {
 
   state = {
     width: window.innerWidth,
+    columns: new Immutable.Map(),
     draggingOver: false,
   };
 
   handleResize = () => {
     this.setState({ width: window.innerWidth });
+
+    if (!isMobile(window.innerWidth)) {
+      this.lazyLoadStaticColumns();
+    }
   }
 
   handleDragEnter = (e) => {
@@ -95,6 +93,15 @@ class UI extends React.PureComponent {
     this.setState({ draggingOver: false });
   }
 
+  lazyLoadStaticColumns = () => {
+    import(/* webpackChunkName: "features/home_timeline" */ '../home_timeline')
+      .then(Component => this.setState(state => ({ columns: state.columns.set('HomeTimeline', Component.default) })));
+    import(/* webpackChunkName: "features/compose" */ '../compose')
+      .then(Component => this.setState(state => ({ columns: state.columns.set('Compose', Component.default) })));
+    import(/* webpackChunkName: "features/notifications" */ '../notifications')
+      .then(Component => this.setState(state => ({ columns: state.columns.set('Notifications', Component.default) })));
+  };
+
   componentWillMount () {
     window.addEventListener('resize', this.handleResize, { passive: true });
     document.addEventListener('dragenter', this.handleDragEnter, false);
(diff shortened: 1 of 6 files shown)


app/javascript/mastodon/features/ui/index.js
```diff
@@ -128,11 +133,18 @@ class UI extends React.PureComponent {
         </ColumnsArea>
       );
     } else {
+      // TODO: Better check & error handling
```
akihikodaki: Why not just initialize `staticColumns` with `null` and test it is `null`? I guess you did so in favor of `const`, but `const` does not really make difference in this case because it just eliminates a possibility (reassignment) to get broken while there are many other similar possibilities.

akihikodaki (changes requested):

app/javascript/mastodon/features/ui/index.js
```diff
@@ -128,11 +133,18 @@ class UI extends React.PureComponent {
...
+      if (Object.keys(staticColumns).length !== 3) {
```
nolanlawson: why 3?

app/javascript/mastodon/features/ui/index.js
```diff
@@ -16,6 +13,14 @@ import { refreshTimeline } from '../../actions/timelines';
...
 import UploadArea from './components/upload_area';
 
+const staticColumns = { };
+
+const lazyLoadStaticColumns = () => {
+  import(/* webpackChunkName: "features/home_timeline" */ '../home_timeline').then(Component => staticColumns.HomeTimeline = Component.default);
+  import(/* webpackChunkName: "features/compose" */ '../compose').then(Component => staticColumns.Compose = Component.default);
+  import(/* webpackChunkName: "features/notifications" */ '../notifications').then(Component => staticColumns.Notifications = Component.default);
+};
+
```
nolanlawson: These `webpackChunkName` comments get repeated enough that I wonder if it wouldn't make sense to just create an `async_routes.js` or something that contains all the `import()s` in one place.

nolanlawson (commented):
This is really neat. I need to test this, but some quick thoughts:
- errors could be communicated via generic toasts; I believe this is what the emoji dropdown already does if it fails to fetch the async chunk
- loading state could be communicated via the blue bar up top (if it's not already)
- I'm a bit worried about continuing our reliance on ReactRouter v2. React already complains quite a bit in the console about it, and apparently it won't even work with React 16 once that's released. Any chance we could at least upgrade to v3? Or should that be a separate concern?

nolanlawson:
Also thanks for the WebpackBundleAnalyzer view, didn't realize the onboarding modal was so big! Nice fix to not require it for every user every time.

nolanlawson:
Hm, any idea why certain common modules aren't being put in the common chunk? Looking at the `features` chunks in the bundle analyzer, it seems like modules like `status.js` and `status_action_bar.js` are being repeated in each chunk?

app/javascript/mastodon/features/ui/index.js
sorin-davidoi: Because if I do that I am basically waiting for all three columns to be fetched until I display them. This way as soon as a column is fetched it is displayed. Also, if fetching just one of them fails, the user is still able to interact with the other two.

app/javascript/mastodon/features/ui/index.js
sorin-davidoi: Because there are three columns. Refactored to be cleared and not rely on a magic constant.

sorin-davidoi:
Could it be because of the dynamic imports (I'm thinking that Webpack puts stuff in the common chunk for static imports, with the assumption that if you use dynamic imports you don't want this magic by default)? Just a hunch, a quick search did not reveal much.

sorin-davidoi:
The emoji dropdown has its own `loading` state which gets set to `false` if the fetch fails. I introduced new actions for fetching bundles, which will show the loading bar and an error if the fetch fails.

The only thing is that it is very tedious to work with, since we cannot abstract `import` statements. I would like to write something like:

```js
fetchBundle('features/compose').then(Component => COMPONENTS.Compose = Component.default);
```

with `fetchBundle` handing the actions and fail state.

But since `import` [requires some static information](https://webpack.js.org/guides/code-splitting-async/#dynamic-import-import-) (basically the path), we have to do something like:

```js
store.dispatch(fetchBundleRequest());
import('features/compose')
  .then(Component => {
    COMPONENTS.Compose = Component.default;
    store.dispatch(fetchBundleSuccess());
  })
  .catch(err => store.dispatch(fetchBundleFail(err)));
```

I've added the loading bar / toasts to the emoji picker as well.

akihikodaki (approved):

sorin-davidoi:
I will also add "dummy" empty columns while the bundles are being fetched. Currently the right-most column is being push to the right as the bundles come in

Gargron:
Is this an upgrade of react-router? It kinda looks similiar to what I've seen from latest version of react-router, so maybe it wouldn't be hard to upgrade?

Gargron:
Also slightly concerned if this is going to make #3207 harder

sorin-davidoi:
@Gargron What if we keep this PR for "some components" and tackle the routes in another PR (after #3207 is merged)? Might also be worth discussion, as @nolanlawson mentions, if we should try to upgrade react-router before lazy-loading the routes.

nolanlawson:
"some components" sounds reasonable to me; e.g. the onboarding model seems especially good to extract.

As for `react-router`, yeah, I sorta feel like any routing changes should wait until that upgrade is complete. I'm not sure if we want to go with v3 or v4, though; I coulda sworn I read somewhere that v3 is still under active development.

sorin-davidoi:
Alright, closing this and opening another PR for "some components".

sorin-davidoi:
Taking [another shot at this](https://github.com/tootsuite/mastodon/pull/3879).

PR #3392, opened 2017-05-28
Fix video having black border on top due to regression from #2608

Gargron:
The combination of object-fit, relative position 50% from top and translating it back upwards 50% is what allows us to crop the video properly, so it needs to be +50%-50%

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -3395,7 +3395,7 @@ button.icon-button.active i.fa-retweet {
   object-fit: cover;
   position: relative;
   top: 50%;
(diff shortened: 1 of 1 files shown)


ykzts (approved):

nolanlawson (approved):
Tested in desktop Chrome, Edge, Firefox, and Safari, and I can reproduce the problem in all four as well as confirm that this PR fixes it.

Incidentally Edge doesn't support object-fit yet but it gracefully degrades (by having black bars on the left and right) and this PR is still an improvement because there's no more black bar on top.

PR #3397, opened 2017-05-28
Re-add stream end log for WebSocket

clworld:
This PR re-add Ending stream log for WebSocket which is removed by #3370.


nolanlawson (approved):
Seems fine to me unless there's any reason _not_ to log.

PR #3400, opened 2017-05-28
Tell puma and sidekiq daemons to write pid files

Jehops:
Pid files are useful for the rc.d scripts used to start/stop the daemons in the
FreeBSD port/package.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,6 +1,8 @@
 threads_count = ENV.fetch('MAX_THREADS') { 5 }.to_i
 threads threads_count, threads_count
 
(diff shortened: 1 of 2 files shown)


akihikodaki (changes requested):
I do not think it should be done in Mastodon.
You can write a shell script or something to emit PID file, which is more flexible and suitable for such features specific for environments.

Gargron:
Agree with Aki. The default way to run Mastodon is the 12-factor app approach, aka it runs in foreground, and it's the init system's job to manage it (be it Docker or Systemd). You can customize your deployment.

Jehops:
👍  I just patch for the package.

PR #3404, opened 2017-05-29
Use .dotextension to get .jpeg

saper:
Reported: https://discourse.joinmastodon.org/t/api-media-post-bug-on-jpg-jpe-extension/257/3

not tested at all yet

### DIFF
--- a/app/models/concerns/attachmentable.rb
+++ b/app/models/concerns/attachmentable.rb
@@ -13,9 +13,9 @@ module Attachmentable
     self.class.attachment_definitions.each_key do |attachment_name|
       attachment = send(attachment_name)
       next if attachment.blank?
-      extension = Paperclip::Interpolations.content_type_extension(attachment, :original)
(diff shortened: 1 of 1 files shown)


akihikodaki:
Could you tell me what version you used to reproduce the bug?

Gargron:
This might not be a fix. If dotextension uses the provided extension from the original filename, and merely checks if against content type, then it may allow empty extensions, which the original code is meant to fix. My question is: is "jpe" not a valid extension? And if so, why is it being returned by MIME::Types?

mal0ki:
.jpe seems to be used as a JPEG file extension 
https://en.wikipedia.org/wiki/JPEG

> Filename extension	.jpg, .jpeg, .jpe .jif, .jfif, .jfi

saper:
The problems is OP is sometimes getting .jpe, sometimes .jpeg - do I understand this correctly @cquest ?

Gargron:
It sounds like .jpe is a valid extension and this fix undoes another fix, so I am closing this PR.

PR #3406, opened 2017-05-29
Add IntersectionObserverWrapper to cut down on re-renders

nolanlawson:
This PR adds a new utility, `IntersectionObserverWrapper.js`, and changes how the `status_list.js` and `status.js` handle props/state in order to cut down on re-renders.

The basic idea is to have each `Status` responsible for its own `isIntersecting` state, rather than keeping a map of IDs to `isIntersecting` in the `StatusList`. The advantage here is that, if one `Status` changes its intersection state, the whole `StatusList` doesn't need to re-render. So e.g. we no longer need `intersectionCount`.

This cuts down on wasteful re-renders. I tested this by opening the home page and then scrolling down the local timeline twice, then doing `ReactPerf.printWasted()`. Before:

screenshot 2017-05-28 20 14 16

After:

screenshot 2017-05-28 20 15 13

(Note the reduction in Inclusive wasted time and Render count.)

I tested in Chrome, Edge, Firefox, and Safari (all on desktop) and confirmed this is working, insofar as I didn't see anything breaking and the DOM node count stays fairly consistent or decreases as you scroll.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -32,16 +32,16 @@ class Status extends ImmutablePureComponent {
     onOpenMedia: PropTypes.func,
     onOpenVideo: PropTypes.func,
     onBlock: PropTypes.func,
-    onRef: PropTypes.func,
-    isIntersecting: PropTypes.bool,
     me: PropTypes.number,
     boostModal: PropTypes.bool,
     autoPlayGif: PropTypes.bool,
     muted: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


nolanlawson:
This is live on https://malfunctioning.technology as well for those who want to test it out. :smiley:

/cc @sorin-davidoi

sorin-davidoi (approved):
Nice work! 👍

ykzts (approved):
good!

PR #3411, opened 2017-05-29
[RFC] Return 401 for an authentication error on WebSockets

unarist:
When the streaming API was connected with missing/invalid token, the connection will be closed just after **connected**. So each attempts will call reconnected handler (`TIMELINE_CONNECT`, `NOTIFICATIONS_REFRESH_REQUEST` and togge timeline polling) even if the connection will be closed soon.

This patch solves it by aborting handshake with HTTP 401, like Server-sent Events version does.

However, most of authentication errors will not be resolved by plain retrying, so we may not want to retry again and again. Returning 401 will make harder to distinguish an authentication error and others due to security restriction in the WebSockets API. Sending error information as WebSocket message would better for this, but this makes front-end scripts bit complicated probably.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -95,7 +95,6 @@ const startWorker = (workerId) => {
   const app    = express();
   const pgPool = new pg.Pool(Object.assign(pgConfigs[env], dbUrlToConfig(process.env.DATABASE_URL)));
   const server = http.createServer(app);
-  const wss    = new WebSocket.Server({ server });
   const redisNamespace = process.env.REDIS_NAMESPACE || null;
 
   const redisParams = {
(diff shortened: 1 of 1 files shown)


akihikodaki (approved):

streaming/index.js
```diff
@@ -352,44 +367,37 @@ const startWorker = (workerId) => {
...
   });
 
+  const wss    = new WebSocket.Server({ server, verifyClient: wsVerifyClient });
+
   wss.on('connection', ws => {
-    const location = url.parse(ws.upgradeReq.url, true);
-    const token    = location.query.access_token;
-    const req      = { requestId: uuid.v4() };
+    const location  = url.parse(ws.upgradeReq.url, true);
+    const req       = { requestId: uuid.v4(), accountId: ws.upgradeReq.accountId };
```
Gargron: I think this req is missing `filteredLanguages`

unarist: Ah, that's right. I've fixed it.

Gargron (approved):

PR #3421, opened 2017-05-29
Extract polyfill loading into single module

nolanlawson:
This is just code cleanup. It moves all the polyfill loading into a single module so that the same code can be used both from `public.js` and `application.js`.

I tested this on Chrome, Firefox, Safari, and iOS 9.

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -3,6 +3,7 @@ import { getLocale } from 'mastodon/locales';
 import { length } from 'stringz';
 import IntlRelativeFormat from 'intl-relativeformat';
 import { delegate } from 'rails-ujs';
+import loadPolyfills from '../mastodon/load_polyfills';
 
 require.context('../images/', true);
 
@@ -85,12 +86,6 @@ function main() {
   });
 }
 
-if (!window.Intl) {
-  import(/* webpackChunkName: "base_polyfills" */ 'mastodon/base_polyfills').then(() => {
-    main();
-  }).catch(error => {
-    console.log(error); // eslint-disable-line no-console
-  });
-} else {
-  main();
-}
+loadPolyfills().then(main).catch(error => {
+  console.log(error); // eslint-disable-line no-console
+});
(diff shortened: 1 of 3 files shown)


app/javascript/packs/public.js
```diff
@@ -3,6 +3,7 @@ import { getLocale } from 'mastodon/locales';
 import { length } from 'stringz';
 import IntlRelativeFormat from 'intl-relativeformat';
 import { delegate } from 'rails-ujs';
+import loadPolyfills from '../mastodon/load_polyfills';
```
ykzts: import loadPolyfills from 'mastodon/loadPolyfills';

ykzts (commented):
good.

app/javascript/packs/public.js
nolanlawson: `mastodon` isn't an npm package (well actually [it is](https://www.npmjs.com/package/mastodon) 😅 ) so I don't think that's what we want…

ykzts: The name mastodon is alias. It means real "app/javascript/mastodon".

ykzts: https://github.com/tootsuite/mastodon/pull/3421/files#diff-675fc760edfd7daf1b7d552fbd04fd99L2 other case

nolanlawson: Thanks, might as well!

nolanlawson: Actually TBH I find this a bit annoying because my editor (WebStorm) no longer allows me to ctrl-click to jump to the declaration for the imported module. I also can't find where the `mastodon` alias is declared. I don't think removing three characters is worth it.

ykzts: I agree. It may be good to rewrite other parts as well. However, irregularities are not good.

nolanlawson: Agreed, I'd rather it were just consistent everywhere. I'd be happy to see the alias removed assuming it's not too controversial. 🙂

ykzts: I created #3433.

ykzts (approved):

PR #3423, opened 2017-05-29
Remove react-imageloader

nolanlawson:
`react-imageloader` is causing a warning in the console:

    Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

Furthermore, its last version was released 2 years ago and [appears to be inactive](https://github.com/hzdg/react-imageloader/graphs/contributors). And we're not really using it for much; an `` tag works just fine instead.

Here's a screenshot before this PR (note the transparent white-and-gray checkerboard background):

screenshot 2017-05-29 11 29 06

And after:

screenshot 2017-05-29 11 29 32

(The point is that they look exactly the same. :smiley:)

### DIFF
--- a/app/javascript/mastodon/features/ui/components/image_loader.js
+++ b/app/javascript/mastodon/features/ui/components/image_loader.js
@@ -2,8 +2,6 @@ import React from 'react';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 import PropTypes from 'prop-types';
 
-import notFoundPng from '../../../../images/mastodon-not-found.png';
-
 class ImageLoader extends ImmutablePureComponent {
 
   static propTypes = {
--- a/app/javascript/mastodon/features/ui/components/image_loader.js
+++ b/app/javascript/mastodon/features/ui/components/image_loader.js
@@ -12,9 +12,8 @@ class ImageLoader extends React.PureComponent {
     error: false,
   }
 
-  constructor(props) {
-    super(props);
-    this.loadImage(props.src);
+  componentWillMount() {
+    this.loadImage(this.props.src);
   }
 
   componentWillReceiveProps(props) {
--- a/app/javascript/mastodon/features/ui/components/image_loader.js
+++ b/app/javascript/mastodon/features/ui/components/image_loader.js
@@ -1,0 +1,49 @@
+import React from 'react';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+import PropTypes from 'prop-types';
+
+import notFoundPng from '../../../../images/mastodon-not-found.png';
+
+class ImageLoader extends ImmutablePureComponent {
+
+  static propTypes = {
+    src: PropTypes.string.isRequired,
+  }
+
+  state = {
+    loading: true,
+    error: false,
+  }
+
+  constructor(props) {
+    super(props);
+    this.loadImage(props.src);
+  }
+
+  componentWillReceiveProps(props) {
+    this.loadImage(props.src);
+  }
+
+  loadImage(src) {
+    const image = new Image();
+    image.onerror = () => this.setState({loading: false, error: true});
+    image.onload = () => this.setState({loading: false, error: false});
+    image.src = src;
+    this.lastSrc = src;
+    this.setState({loading: true});
+  }
+
+  render() {
+    const { src } = this.props;
+    const { loading, error } = this.state;
+
+    // TODO: handle image error state
+
+    const imageClass = `image-loader__img ${loading ? 'image-loader__img-loading' : ''}`;
+
+    return <img className={imageClass} src={src} />; // eslint-disable-line jsx-a11y/img-has-alt
+  }
+
+}
+
+export default ImageLoader;
--- a/app/javascript/mastodon/features/ui/components/image_loader.js
+++ b/app/javascript/mastodon/features/ui/components/image_loader.js
@@ -1,8 +1,7 @@
 import React from 'react';
-import ImmutablePureComponent from 'react-immutable-pure-component';
 import PropTypes from 'prop-types';
 
-class ImageLoader extends ImmutablePureComponent {
+class ImageLoader extends React.PureComponent {
 
   static propTypes = {
     src: PropTypes.string.isRequired,
--- a/app/javascript/mastodon/features/ui/components/media_modal.js
+++ b/app/javascript/mastodon/features/ui/components/media_modal.js
@@ -3,7 +3,6 @@ import LoadingIndicator from '../../../components/loading_indicator';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import PropTypes from 'prop-types';
 import ExtendedVideoPlayer from '../../../components/extended_video_player';
-import ImageLoader from 'react-imageloader';
 import { defineMessages, injectIntl } from 'react-intl';
 import IconButton from '../../../components/icon_button';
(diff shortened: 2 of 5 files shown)


app/javascript/styles/components.scss
```diff
@@ -1137,15 +1137,6 @@
...
 
-.transparent-background,
-.imageloader {
-  background: url('../images/void.png');
-}
-
-.imageloader {
-  display: block;
-}
-
```
nolanlawson: `.transparent-background` is not used anywhere else.

app/javascript/mastodon/features/ui/components/media_modal.js
```diff
@@ -73,7 +72,7 @@ class MediaModal extends ImmutablePureComponent {
     }
 
     if (attachment.get('type') === 'image') {
-      content = <ImageLoader src={url} imgProps={{ style: { display: 'block' } }} />;
+      content = <img src={url} />;
```
ykzts: add alt='' or role='presentation'

ykzts (changes requested):
Does it make sense not to offer the same function as ImageLoader?

app/javascript/mastodon/features/ui/components/media_modal.js
nolanlawson: The purpose of both [`alt=''`](https://www.w3.org/WAI/tutorials/images/decorative/) and [`role='presentation'`](http://asurkov.blogspot.com/2012/02/aria-hidden-and-rolepresentation.html) (as I understand it) is to just hide the `img` from ATs. This particular `img` is the modal dialog for a user-uploaded image, and to me it really ought to have an `alt` supplied by the user, but unfortunately Mastodon doesn't provide any way to do this currently.

In any case the existing behavior of `ImageLoader` is to have neither `alt` nor `role` so the behavior is unchanged.

nolanlawson:
> Does it make sense not to offer the same function as ImageLoader?

We weren't using any of the features of `ImageLoader`; we were just showing an image. That's why I found it okay to replace with a regular ``. I don't see much need to do anything fancy here. :smiley:

app/javascript/mastodon/features/ui/components/media_modal.js
ykzts: Hm. Then you can disable ESLint settings at the line level?

nolanlawson: Done! :)

ykzts: :ok_man:

ykzts (approved):

Gargron:
> I don't see much need to do anything fancy here. 

The difference between img and ImageLoader was when you change the src of the img it shows the old one first before the updated one loads. This didn't look nice when clicking on different images to open them in the modal. Depending on how the surrounding DOM is rendered nowadays, I am not sure if this is still relevant, but worth checking.

nolanlawson:
OK, I will work on an improved PR.

nolanlawson:
OK, I've added a new `image_loader.js` that actually improves on the old one. Here's a demo before and after, throttled to 3G to demonstrate.

Before: https://gfycat.com/ExemplaryDeadLaughingthrush

After: https://gfycat.com/InbornWeepyDwarfmongoose

I think this can be improved even further:

1. For the first image, show the thumbnail instead of nothing while the image loads
2. When an image fails to load, show something like "image failed to load, please try again" or maybe a custom "broken image" image so we don't need to translate it.

But IMO those can wait for a future PR. 😃

app/javascript/mastodon/features/ui/components/image_loader.js
```diff
@@ -0,0 +1,47 @@
...
+
+  state = {
+    loading: true,
+    error: false,
+  }
+
+  constructor(props) {
+    super(props);
+    this.loadImage(props.src);
+  }
```
ykzts: Why is not it a componentDidMount?

app/javascript/mastodon/features/ui/components/image_loader.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+import PropTypes from 'prop-types';
+
+class ImageLoader extends ImmutablePureComponent {
```
ykzts: React.PureComponent

ykzts (changes requested):

app/javascript/mastodon/features/ui/components/image_loader.js
nolanlawson: It doesn't need access to the DOM used by the component. So this way it starts loading the image a bit earlier.

app/javascript/mastodon/features/ui/components/image_loader.js
nolanlawson: We've been using ImmutablePureComponent throughout the codebase. I'm not 100% sure on the difference, but I like ImmutablePureComponent because it has nice settings for specifying which props/states to diff.

app/javascript/mastodon/features/ui/components/image_loader.js
ykzts: In that case you should use componentWillMount. setState in constructor is not very good.

app/javascript/mastodon/features/ui/components/image_loader.js
ykzts: ImmutablePureComponent is for Immutable.js. Immutable.js is not used in this component, so I think it is unnecessary.

nolanlawson: OK I read https://facebook.github.io/react/docs/react-api.html#react.purecomponent and so I think I understand. It's fine to change.

app/javascript/mastodon/features/ui/components/image_loader.js
nolanlawson: I'm not familiar with React conventions. OK!

ykzts (approved):

sorin-davidoi:
Awesome! That empty modal annoys me greatly, thank you for taking time and fixing it!

sorin-davidoi:
I really like the way [`iron-image`](https://www.webcomponents.org/element/PolymerElements/iron-image) handles the loading state. They suggest resizing the image to `1%` of the initial size and saving it in the database encoded in `base64`. This field gets sent to the front-end, which displays it via `data:image` until the image is fetched, having a really nice blur effect. You can see it [here](https://shop.polymer-project.org/) on the main images (you may need to throttle the network to observe the initial image. We might want to look into this in the future, I think it is a nice user experience (although it might put the back-end under some stress by resizing each image).

Gargron:
@sorin-davidoi For the record, @DagAgren was the first to propose the base64 method.

Though displaying the stretched out thumbnails first is also a good first step! Given that the "meta" property includes width + height, it should be even possible to stretch them to the right dimensions before the real one loads.

DagAgren:
What I've been working on on is even more extreme: It's a custom image format that fits in a 20-30 character string, and unpacks to a nicely blurred image without using SVG tricks for blurring.

It's still under development though, and will take a while before I have the time to finish it. But it creates far less overhead than base64 mini thumbnails.

nolanlawson:
I like the idea of blurring the low-res image using CSS, especially given we already have the low-res image in the list view, so it could be reused for the modal.

PR #3425, opened 2017-05-29
Move status reblog authorization into policy

jackjennings:
This PR moves the authorization check for "reblogging" into the `StatusPolicy` class. Currently, this authorization check is only being used in the `ReblogService`.

The specs for `StatusPolicy` have been updated to cover the new `reblog?` predicate method.


jackjennings:
CI seems to be failing before the specs are run.

ykzts (approved):

PR #3426, opened 2017-05-29
Fix webpack building on Windows

Naouak:
You should never check paths using manually constructed paths on Node. Use `path.join` or `path.sep` to build paths.


ykzts (approved):
good catch!

nightpool:
have you actually tested this on windows? Windows always accepts both as path seperators

Naouak:
I have tested on Windows. It may accept both separators but it would not change anything if you check two path together.

config/webpack/shared.js
```diff
@@ -47,14 +47,15 @@ module.exports = {
     new webpack.optimize.CommonsChunkPlugin({
       name: 'common',
       minChunks: (module, count) => {
-        if (module.resource && /node_modules\/react-intl/.test(module.resource)) {
+        const reactIntlPathRegexp = new RegExp(`node_modules\\${sep}react-intl`);
```
ykzts: It seems that it will not move if there is `\\`.

ykzts (changes requested):

config/webpack/shared.js
Naouak: the \\ is only to add a slash to escape the separator. Without it, the regexp fails.

ykzts: Will not `new RegExp` escape it?

ykzts: Oh, I'm sorry, it was my misunderstanding.

ykzts (approved):

config/webpack/shared.js
Naouak: I thought too but you need an actual escaping in the regexp string to avoid confusions with a regexp class(\d \w...).

PR #3453, opened 2017-05-30
Add status destroy authorization to policy

jackjennings:
The PR adds a `destroy?` authorization check for statuses. Only admin users or the account that is connected to a status is authorized to destroy.

I've also (in a separate commit) added an explicit `unreblog?` check that is an alias for `destroy?`.

I have also added a spec for `ProcessInteractionService` for incoming delete "slaps". I'd appreciate a review on the spec code here.

### DIFF
--- a/app/controllers/admin/reported_statuses_controller.rb
+++ b/app/controllers/admin/reported_statuses_controller.rb
@@ -2,6 +2,8 @@
 
 module Admin
   class ReportedStatusesController < BaseController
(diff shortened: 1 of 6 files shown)


ykzts (approved):

jackjennings:
Sorry—updated these commits to test that the `RemoveWorker` actually receives the correct arguments in the spec.

ykzts (approved):

PR #3457, opened 2017-05-30
Type toots using speech recognition

sorin-davidoi:
[Browser support](http://caniuse.com/#feat=speech-recognition): Chrome (Firefox implemented it but it is under a flag, Edge is currently implementing it).

peek 2017-05-30 22-52

Stuff to address:
 - [ ] The API may return no results. We should perhaps display a notification informing the user.
 - [ ] Lazy-loading - it is pretty small but can be done in the "lazy load some components" PR discussed in #3381.

### DIFF
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -13,6 +13,7 @@ import Collapsable from '../../../components/collapsable';
 import SpoilerButtonContainer from '../containers/spoiler_button_container';
 import PrivacyDropdownContainer from '../containers/privacy_dropdown_container';
 import SensitiveButtonContainer from '../containers/sensitive_button_container';
+import VoiceRecorder from './voice_recorder';
 import EmojiPickerDropdown from './emoji_picker_dropdown';
 import UploadFormContainer from '../containers/upload_form_container';
 import TextIconButton from './text_icon_button';
@@ -184,6 +185,7 @@ class ComposeForm extends ImmutablePureComponent {
         <div className='compose-form__buttons-wrapper'>
           <div className='compose-form__buttons'>
             <UploadButtonContainer />
+            <VoiceRecorder onChange={this.handleChange} />
             <PrivacyDropdownContainer />
             <SensitiveButtonContainer />
             <SpoilerButtonContainer />
--- a/app/javascript/mastodon/features/compose/components/voice_recorder.js
+++ b/app/javascript/mastodon/features/compose/components/voice_recorder.js
@@ -1,0 +1,84 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { defineMessages, injectIntl } from 'react-intl';
+import IconButton from '../../../components/icon_button';
+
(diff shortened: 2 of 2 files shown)


Gargron:
My question is - how did you fit all of those icons into the compose bar? Did you make the compose column wider?

sorin-davidoi:
Yes, it is a bit wider to accommodate for the extra button. I think 4 is the upper limit on the number of buttons that can fit here (having an extra one would cause it to wrap on mobile).

unarist:
> I think 4 is the upper limit on the number of buttons that can fit here

We already have 4 buttons: upload, visibility, CW and NSFW. The recording button will be 5th button...

nolanlawson:
This is cute, but I have to ask if we really need this. It adds an extra icon, which adds additional noise to the UI. Plus, on the only browser+device combo where users would conceivably use this feature ([Chrome on Android](http://caniuse.com/#feat=speech-recognition)), the built-in Android keyboard already has a microphone icon that the user can press to input speech.

danhunsaker:
> on the only browser+device combo where users would conceivably use this feature

Not accurate. There are many reasons to use speech recognition instead of typing, and mobile devices are only one of them. Accessibility is far more of an afterthought in most software than it really ought to be.

Also, not all keyboards, even on Android devices, actually include a speech input icon/mode.

The extra icon is definitely something that causes real, concrete change, and that needs reviewing. The feature itself, however, is an improvement of the UI/UX that empowers more users to get involved.

nolanlawson:
> Accessibility is far more of an afterthought in most software than it really ought to be.

If we want to improve a11y, I'd say priority number 1 is adding alt tags to images. 🙃

You're right though that this could be an a11y feature, but I'd be interested in asking differently-abled folks (e.g. people with RSI) if this truly does empower them. It's possible to go overboard on a11y features by adding things like contrast settings, font size settings, zoom settings, and other stuff that's better handled by the browser/OS itself. (Then there's stuff like "disable auto-playing GIFs" that falls on the other side of the spectrum.)

In any case yeah this icon could be enabled via the settings (disabled by default) in which case it's a user choice and doesn't negatively impact the UI for folks who don't want it.

sorin-davidoi:
> We already have 4 buttons: upload, visibility, CW and NSFW.
I don't have NSFW either on toot.cafe nor did I see it on the current master.

Regarding the "utility" of this feature, I agree that it would better be handled by the operating system (that being said, on Linux at least I would have to go though some hoops to get this behavior). I implemented it mostly to play with the new API.  Glad to see it sparked a conversation.

unarist:
NSFW button will be shown when you upload an image.

gif

sorin-davidoi:
Right 😄 Then I guess it needs a new place if we decide to ship this (maybe next to the emoji-picker?).

Cassolotl:
Rolling CW and NSFW together (as in https://github.com/tootsuite/mastodon/issues/1010) would give a little more space! Not that I have an agenda there or anything...

sorin-davidoi:
Closing due to lack of consensus.

PR #3471, opened 2017-05-31
Refactor api/v1/apps controller

mjankowski:


ykzts (approved):

ykzts:
randomly fails...

mjankowski:
There's an intermittent failure there I see sometimes locally which comes up in the feature spec. I'm guessing it's a JS loading issue, but haven't tracked it down yet.

ykzts:
Probably it will be fixed with #3479.

PR #3490, opened 2017-06-01
Introduce StatusThreadingConcern

mjankowski:
Improves spec coverage for the methods used within ancestors and descendants methods.


app/lib/status_filter.rb
```diff
@@ -0,0 +1,50 @@
...
+class StatusFilter
+  attr_reader :status, :account
+
+  def initialize(status, account)
+    @status = status
+    @account = account
+  end
+
+  def filtered?
+    unless account.nil?
```
mjankowski: could this just be another method in the list?

app/models/concerns/status_threading_concern.rb
```diff
@@ -0,0 +1,89 @@
...
+        JOIN statuses ON statuses.id = search_tree.in_reply_to_id
+        WHERE NOT statuses.id = ANY(path)
+      )
+      SELECT id
+      FROM search_tree
+      ORDER BY path DESC
+    SQL
+  end
+
+  def descendant_ids
```
mjankowski: should this be cached too?

app/models/concerns/status_threading_concern.rb
```diff
@@ -0,0 +1,89 @@
...
+        JOIN statuses ON statuses.in_reply_to_id = search_tree.id
+        WHERE NOT statuses.id = ANY(path)
+      )
+      SELECT id
+      FROM search_tree
+      ORDER BY path
+    SQL
+  end
+
+  def find_statuses_from_tree_path(ids, account)
```
mjankowski: it would be nice if this method could be simpler

unarist:
How about having a test for `status.account.blocking?(viewer)` from StatusPolicy?

spec/models/concerns/status_threading_concern_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    it 'returns conversation history' do
+      expect(reply3.ancestors).to include(status, reply1, reply2)
+    end
+
+    it 'does not return conversation history user is not allowed to see' do
+      reply1.update(visibility: :private)
+      status.update(visibility: :direct)
+
+      expect(reply3.ancestors(viewer)).to_not include(reply1, status)
```
unarist: How about having a test for `status.account.blocking?(viewer)` from StatusPolicy?

mjankowski: I was trying to not re-test things which would have been already tested well in status policy spec ... if there's a case which is particularly concerning or not well spec'd there though, I can look at adding more coverage. Will check that out and report back.

app/models/status.rb
```diff
@@ -115,16 +116,6 @@ def hidden?
     private_visibility? || direct_visibility?
   end
 
-  def ancestors(account = nil)
-    ids = Rails.cache.fetch("ancestors:#{id}") { (Status.find_by_sql(['WITH RECURSIVE search_tree(id, in_reply_to_id, path) AS (SELECT id, in_reply_to_id, ARRAY[id] FROM statuses WHERE id = ? UNION ALL SELECT statuses.id, statuses.in_reply_to_id, path || statuses.id FROM search_tree JOIN statuses ON statuses.id = search_tree.in_reply_to_id WHERE NOT statuses.id = ANY(path)) SELECT id FROM search_tree ORDER BY path DESC', id]) - [self]).pluck(:id) }
-    find_statuses_from_tree_path(ids, account)
-  end
-
-  def descendants(account = nil)
-    ids = (Status.find_by_sql(['WITH RECURSIVE search_tree(id, path) AS (SELECT id, ARRAY[id] FROM statuses WHERE id = ? UNION ALL SELECT statuses.id, path || statuses.id FROM search_tree JOIN statuses ON statuses.in_reply_to_id = search_tree.id WHERE NOT statuses.id = ANY(path)) SELECT id FROM search_tree ORDER BY path', id]) - [self]).pluck(:id)
```
mjankowski: @Gargron - prior to my refactor here, this method (`descendants`) is not using the rails cache.

Is that purposeful and we don't want to cache those ... or should I add in the cache for that method while I'm working in here?

Gargron: You can't cache those. Ancestors never change but if you add a child then you'd need to invalidate the cache.

Gargron (approved):

spec/lib/status_filter_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+require 'rails_helper'
+
+describe StatusFilter do
+  describe '#filtered?' do
+    let(:status) { Fabricate(:status) }
+
+    context 'without an account' do
+      subject { described_class.new(status, nil) }
+
+      it { is_expected.not_to be_filtered }
```
Gargron: Yeah so this was a really bad assumption. It should be filtered if the status is private, or direct, or the author is silenced.

PR #3492, opened 2017-06-01
Fix #3194 - Docker: Use UID/GID as build-args

### LINKED ISSUE #3194: Permission update on docker entrypoint takes a long time

* * * *

maur1th:
To keep both the flexibility introduced by https://github.com/tootsuite/mastodon/pull/3159 and preserve the start time of the containers (see issue https://github.com/tootsuite/mastodon/issues/3194 for instance), add user and set its permissions at build time instead of run time.

Flexibility is preserved by using build arguments for UID and GID, so specific needs can be addressed by rebuilding the image with the appropriate `--build-arg` arguments.

Cc @Wonderfall


Wonderfall:
I thought the matter has already been discussed, harcoding things like this in the Dockerfile is not a good practice. The start time of the container shouldn't be affected, it's because of one single storage driver.

The use of UID/GID as build arguments has already been mentioned in this previous PR : https://github.com/tootsuite/mastodon/pull/3152. Unfortunately, it does not preserve the same flexibility. For example, what if the image is being built elsewhere (Docker Hub)?

tlemarchand:
While I understand the need for flexibility, I don't see the point of setting this parameter at runtime.
You can have a fixed UID/GID inside your image, and map it to another user on the host ( https://success.docker.com/KBase/Introduction_to_User_Namespaces_in_Docker_Engine ).
This way you have all the flexibility you need, without having a workaround inside the entrypoint.

Wonderfall:
All 3 options were proposed : setting permissions at build time, at run time, and using user namespaces (please read our previous discussions!). The use of these namespaces shouldn't be required, and it could too complicated for the final user. While I understand the need when the image is using root processes, I don't think it's the good way to handle permissions.

If you don't see the point of setting this parameter at runtime, the, tell us what arguments you have.

cc @xataz

maur1th:
Someone would need these namespaces only if he already wanted to be able to use specific UID/GID.

However I understand creating a user with hardcoded UID/GID isn't great for everyone so I'll just use my own Dockerfile and entrypoint on top of those ones.

Wonderfall:
And yet no one told here why setting permissions at run time is a bad thing. If I understand well, you pointed out the performance impact. But there should be none. It was discussed all over https://github.com/tootsuite/mastodon/issues/3194 and the responsible is somehow OverlayFS. So I think we should open an issue at the competent repository rather than doing nothing and telling people to change their storage driver (which is not a good solution either). Clearly the performance impact seems to come from OverlayFS.

maur1th:
To be fair I still think it's bad practice to `chown` your whole code base at each startup to set your IDs.

I think it over complicates things only to serve a very specific need. So I don't think it should be handled in a community image but rather the persons who have that need might want to build their own image (either from scratch or on top of the community one). 

But that's a matter of personal appreciation so I don't see the point in arguing about it.

Wonderfall:
I know many community images which are letting the user set its own IDs. I have in mind, for example, `emby/embyserver`. Instead I think the bad practice would be to use root or to use `chown` in the Dockerfile. Running `chown` at startup doesn't over complicate things, it's the opposite, so it makes very easy to change the IDs. 

Instead we should create our own image to match our needs? But a "very specific" need? Well, what I see is an important need and that's why the image should take care of that.  We got rid of root, which is a good thing. Now you don't want to run all your containers with the same IDs. In a classic installation, we're creating a new `mastodon` user with its own IDs, we're not reusing one.  And this is a good practice.

So yeah like you said, that's a matter of personal appreciation, we don't share the same vision. I'm for a simple Docker image which handles permissions without the need to rebuild.

xataz:
Hi,

I agree with @Wonderfall. 
Set UID/GID when starting the container provides a better modularity, it's so easy, but less performance (with overlayFS/aufs).
Set UID/GID when build the image provides a better performance, a modularity, but is less easy.
Use user namespace with remap provides a better security, user abstraction, but it's hard.

In fact it's a question of compromise, security/better modularity/easy, security/modularity/better performance, or better security/performance/hard. 

PS : Sorry for my bad english ^^

PR #3493, opened 2017-06-01
Change to returns nil when Account.find_local is given blank username (regression from #3485)

ykzts:
`HomeController#index` returns "404 Not Found" when no value is specified for `Setting.site_contact_username`.

### DIFF
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -30,7 +30,7 @@ module AccountFinderConcern
     end
 
     def account
-      scoped_accounts.take
+      scoped_accounts.take unless username.blank?
     end
 
     private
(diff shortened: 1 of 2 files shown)


mjankowski:
Related: https://github.com/tootsuite/mastodon/pull/3476

ykzts:
@mjankowski That is the behavior of `find_remote!`. It is different from `find_remote`.

mjankowski:
Yes, it is.

mjankowski:
Sorry about the brevity last night - I understand that change is different, but that change is what prompted me to do the large refactor that introduced this unspecified regression, which is why I linked it.

What do you think about this alternate approach to resolving this? https://github.com/tootsuite/mastodon/pull/3500

ykzts:
@mjankowski Your solution looks better than my solution. Good Code!

PR #3506, opened 2017-06-01
Reduce container size with clean yarn

hinaloe:
Erase unnecessary files with cleaning yarn cache (located on `/usr/local/share/.cache/yarn/v1`) and [`yarn clean`](https://yarnpkg.com/lang/en/docs/cli/clean/), container size will become very small (in my env, 1.46GB to 1.06GB)


Gargron:
Does assets:precompile still work after this?

hinaloe:
Yes, work fine!

Gargron (approved):

PR #3508, opened 2017-06-01
Docker: Smaller image, remove build dependencies

maur1th:
I noticed build-dependencies weren't removed from the the image although a virtual package was created (`apk add -t build-dependencies`). Not sure why and I couldn't find mention of that in the PR where it was introduced (https://github.com/tootsuite/mastodon/pull/3182). So I "fixed" it (`apk del --no-cache build-dependencies`) and split the `apk add RUN` instruction to better leverage build caching following this change.

Image size goes down from 1.41GB to 1.22GB and no more build tools are included in production containers.

I also removed `rm -rf /tmp/* /var/cache/apk/*` because those folders are empty (well, almost: 3 4KB files in /tmp/) at this point.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,18 +9,8 @@ ENV UID=991 GID=991 \
 
 EXPOSE 3000 4000
 
-WORKDIR /mastodon
-
 RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
- && apk -U upgrade \
- && apk add -t build-dependencies \
-    build-base \
-    libxml2-dev \
-    libxslt-dev \
-    postgresql-dev \
-    protobuf-dev \
-    python \
- && apk add \
+ && apk add --no-cache \
     ca-certificates \
     ffmpeg \
     file \
@@ -34,14 +24,22 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
     protobuf \
     su-exec \
     tini \
- && npm install -g npm@3 && npm install -g yarn \
  && update-ca-certificates \
- && rm -rf /tmp/* /var/cache/apk/*
+ && npm install -g npm@3 && npm install -g yarn
 
 COPY Gemfile Gemfile.lock package.json yarn.lock /mastodon/
+WORKDIR /mastodon
 
-RUN bundle install --deployment --without test development \
- && yarn --ignore-optional --pure-lockfile
+RUN apk add --no-cache -t build-dependencies \
+    build-base \
+    libxml2-dev \
+    libxslt-dev \
+    postgresql-dev \
+    protobuf-dev \
+    python \
+ && bundle install --clean --no-cache --deployment --without test development \
+ && yarn --ignore-optional --pure-lockfile \
+ && apk del --no-cache build-dependencies
 
 COPY . /mastodon
 


Wonderfall:
It was the case before 1.4 and the size was around 700MB.
https://github.com/tootsuite/mastodon/blob/1.3-stable/Dockerfile#L39

@Gargron made himself the decision to prefer caching over a reduced footprint. I personally prefer a reduced footprint but the decision is up to him.

Gargron:
At the time of writing, assets:precompile executes yarn install as well (needlessly so, but it's a gem issue), so with yarn cache clean things would be downloaded twice, I think. Of course, there is another point that maybe assets:precompile should be a step in the build process instead of something you execute yourself on a volume. Yet another point that you could split the docker image into two, one for the node.js based streaming API, and one for the ruby based main application (web + sidekiq), that would also reduce final image size (but you'd have two different images which would likely still amount to the same total size)

BastienM:
If I may, you got a choice to make concerning how the image is used.

For development purposes it will be much nicer to have a good caching mechanism but for production, a smaller footprint would be the best approach.

One solution would be using two set of tags : `v1.4` and `v1.4-dev`, each one having a dedicated Dockerfile with optimizations.
For instance, Node's official Docker image uses `v1.0` and `v1.0-slim` to reduce the footprint in production environment without sacrificing functionalities for the developers.

Wonderfall:
Yep we could split images, that would use the same total size, but we can base the straming image on the official node.js image (based on Alpine Linux) so we can always have the latest version (and we can control it, if necessary). I did that before so if you're okay with that, I can make a PR, and include `assets:precompile` in the web image.

Gargron:
@Wonderfall What will you name the dockerfiles?

@BastienM How can I separate them like this unless I maintain a different git branch which I'd rather not do?

Wonderfall:
Very simple, you can have 2 Dockerfiles, one would be `Dockerfile.streaming`, the other one would be `Dockerfile.web`. It's not a problem since we can use a custom Dockerfile name with `docker build` (using the `-f` parameter if my memory is correct), same goes for docker compose with `dockerfile:`.

BastienM:
@Gargron Exactly like @Wonderfall said. Plus Docker Hub does indeed support a lot of various option for the automatic builds.

Wonderfall:
So shall we separate split up the Dockerfile? I just need to be sure that's what we want, I can work on it eventually.

BastienM:
That would be for the best.
Another good point with this is that you will not have to rebuild the whole image each time the nodejs code is updated but not the ruby's part. And vice-versa.

Wonderfall:
Okay. I'm working on it.

Wonderfall:
I'm afraid it will be more difficult than I thought : 
- If assets are compiled during the build, Puma will have to serve them since a volume cannot be mounted anymore (otherwise assets will be erased).
- Compiling assets requires the secret key so I'm not sure how to deal with it yet.
- The ruby based image will still need node.js at some point, for compiling assets. It would become very tricky to get it working and then remove all unnecessary packages.
- While both images have a smaller size (430MB for the node image), having both will result in the same disk occupation as having a single image.

Gargron:
@Wonderfall 

>Compiling assets requires the secret key so I'm not sure how to deal with it yet.

You can probably define a default SECRET_KEY_BASE in the ENV command of the Dockerfile. Secret is not actually needed, it's just the Devise initializer complaining because the whole environment is loaded during precompile.

> It would be interesting to move npm install -g npm@3 && npm install -g yarn there too.

I think Node.js 6.x comes with NPM >= 3 so the first part is redundant. Also, I heard that Alpine might have a yarn package.

>If assets are compiled during the build, Puma will have to serve them since a volume cannot be mounted anymore (otherwise assets will be erased).

Yes, this is the main problem I can see. If you still define a volume, *after* the compile during build, I think the volume will have the data as a default. But I don't know what would happen on subsequent builds (i.e. upgrades). I think the build won't touch the volume anymore.

Wonderfall:
Yes I managed to get it working with SECRET_KEY_BASE, thanks.

> Also, I heard that Alpine might have a yarn package.

Unfortunately `yarn` is not available in Alpine 3.4 (we shouldn't use packages from `edge` too often, in fact we should remove them once the official image is based on Alpine 3.6), and I think it's better to keep it that way so we can always grab the latest version. Same for `npm`.

> Yes, this is the main problem I can see. If you still define a volume, after the compile during build, I think the volume will have the data as a default. But I don't know what would happen on subsequent builds (i.e. upgrades). I think the build won't touch the volume anymore.

There won't be any data if we mount something. In fact is it really a drawback? nginx won't serve the files directly but I'm not sure it will have a huge impact on performances.

Wonderfall:
https://paste.targaryen.house/?da25d85dbac23947#Sixegz5jmw5VXFsb/b49S3qNNIeTgM/anvO0nVkfEts=
Size : 1.34GB before, 1.21GB after. This reduces  footprint while keeping the use of cached layers.

Since assets are compiled during the build, would it be acceptable to launch the DB migration task in the entrypoint? Doing so would make the image very easy to update. It shouldn't take too much time.

Gargron:
This might not be relevant anymore? Closing since it's also inactive

PR #3510, opened 2017-06-02
Add index to media_attachments.account_id

mjankowski:


Gargron:
Is the schema update so large because of 5.1 changes?

mjankowski:
Yes, they (finally!) got rid of the effort to align spacing, so while this commit diff is sort of absurd, all future schema diffs will be cleaner.

Gargron (approved):

PR #3524, opened 2017-06-02
Gem version bumps

mjankowski:
Gets bootsnap to 1.0, and restores rake to latest because the upgraded `annotate` allows that now.


Gemfile
```diff
@@ -19,7 +19,7 @@ gem 'paperclip', '~> 5.1'
 gem 'paperclip-av-transcoder', '~> 0.6'
 
 gem 'addressable', '~> 2.5'
-gem 'bootsnap', '~> 0.3'
+gem 'bootsnap'
```
Gargron: Shouldn't we pin it to ~> 1.0 now?

mjankowski: I'd prefer to leave them unspecified except where it matters. I wouldn't have merged this, for example - https://github.com/tootsuite/mastodon/commit/c78e8c01a37c342977b35396007dce7f58052b20 - I don't think the situation that this prevents (someone running `bundle update` and not bothering to think about it's results) actually happens ... and it certainly wouldn't be merged if someone tried to do that.

That said, because bootsnap is very new and we have no idea if it's stable between versions or how it's own dependencies might grow ... I'd be fine locking to `~> 1.0` if you feel strongly that it should be.

Gargron (approved):

PR #3525, opened 2017-06-03
Fix isIntersecting in Chrome

nolanlawson:
Another way to fix #3502. If `entry.isIntersecting` is a boolean, then use it. Otherwise fall back to `entry.intersectionRect.height > 0`.

I have tested this on https://malfunctioning.technology (where it is live) and it seems to work in Edge 15, Chrome 58, and Firefox 54 on Windows 10. If anyone can still reproduce the bug on any browser please let me know. :smile:

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -88,9 +88,11 @@ class Status extends ImmutablePureComponent {
   }
 
   handleIntersection = (entry) => {
-    // Edge 15 doesn't support isIntersecting, but we can infer it from intersectionRatio
+    // Edge 15 doesn't support isIntersecting, but we can infer it
     // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/
-    const isIntersecting = entry.intersectionRatio > 0;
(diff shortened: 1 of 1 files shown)


nolanlawson:
BTW I have no idea if the bug will still be present in Edge or not (maybe there are still odd cases where the `intersectionHeight`  is 0 and `isIntersecting` should be false, but at the very least we won't break for Chrome or (presumably) browsers that use the polyfill such as Firefox or Safari.

It could also be that Chrome's implementation just differs from Edge, and Edge never reports `intersectionHeight === 0` unless it's not intersecting.

unarist (approved):
looks good! I've tested on Chrome 58 + Windows 10.

ykzts (approved):

ykzts:
The specification seems to have failed due to d567a382e30b6d6f2b32dc249380e24b8d5f2be7.

midzer:
@nolanlawson Is this a general fix to support all major browsers for any isIntersecting implementation?

nolanlawson:
@midzer there are three major implementations (Blink, EdgeHTML, the polyfill) and I tested all three. 😊

PR #3533, opened 2017-06-03
rspec testing made faster by test-queue gem

takayamaki:
I had made tests faster by test-queue gem.

On 3 cores VM, tests have faster about 1 min than before change.

Of course, standard rspec is executable too.

```
$ time bundle exec rspec-queue --format progress spec
[DEPRECATION] `have_enqueued_job` is deprecated.  Please use `have_enqueued_sidekiq_job` instead.

【omitted results because too long】

real	0m43.490s
user	1m37.480s
sys	0m16.280s
$ time bundle exec rspec --format progress spec
[DEPRECATION] `have_enqueued_job` is deprecated.  Please use `have_enqueued_sidekiq_job` instead.

【omitted results because too long】

real	1m39.779s
user	1m14.632s
sys	0m17.600s
$
```


ykzts:
`parallel_tests` that do similar things are already in use, but what are the merits of introducing `test-queue` newly?

takayamaki:
sorry, it's my very basically mistake. 

I could not find `parallel:create` and `parallel:prepare` in `rake -T`. 
But it was happening because I forgot to add `RAILS_ENV = test`.

I close this PR.

Issue #2068, opened 2017-04-18
How to get STREAMING_API_BASE_URL from API?


kmycode:
I'm API user.
Streaming API and others may be different.

for example, mstdn.jp api is
streaming api: https://streaming.mstdn.jp/api/v1/streaming
other apis: https://mstdn.jp/api/v1
https://mstdn.jp/@nullkal/308338

I would like to get STREAMING_API_BASE_URL automatically.
But, there seems no api to get in https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md

How to get it from API?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

tateisu:
I have same issue.
How to know ( STREAMING_API_BASE_URL in .env.production ) via API ?

Gargron:
This is pretty unfortunate, but the `STREAMING_API_BASE_URL` was intended for deveopment environments where you wouldn't have nginx and the streaming API could run on a random port... So we have no mechanism of communicating if this URL is non-standard. I suppose this is something that will need to be in `GET /api/v1/instance`

clworld:
Currently, mstdn.jp does 301 redirect to https://streaming.mstdn.jp/api/v1/streaming/ when we access https://mstdn.jp/api/v1/streaming/.
So, Apps using Streaming API must deal with 301 redirect  for the time being.

Gargron:
It is now returned from /api/v1/instance

PR #3556, opened 2017-06-04
Add streaming_api_base_url to /api/v1/instance

ikuradon:
Added streaming_api_base_url to /api/v1/instance for resolving streaming url from some applications.


ykzts:
related with #2068

app/views/api/v1/instances/show.rabl
```diff
@@ -5,3 +5,4 @@ node(:title)       { Setting.site_title }
 node(:description) { Setting.site_description }
 node(:email)       { Setting.site_contact_email }
 node(:version)     { Mastodon::Version.to_s }
```
ykzts:
```ruby
# ...
node(:email)                  { Setting.site_contact_email }
node(:version)                { Mastodon::Version.to_s }
node(:streaming_api_base_url) { Rails.configuration.x.streaming_api_base_url }
```

adjust the space.

app/views/api/v1/instances/show.rabl
```diff
@@ -1,7 +1,8 @@
...
-node(:title)       { Setting.site_title }
-node(:description) { Setting.site_description }
-node(:email)       { Setting.site_contact_email }
-node(:version)     { Mastodon::Version.to_s }
+node(:uri)                    { site_hostname }
+node(:title)                  { Setting.site_title }
+node(:description)            { Setting.site_description }
+node(:email)                  { Setting.site_contact_email }
+node(:version)                { Mastodon::Version.to_s }
+node(:streaming_api_base_url) { Rails.configuration.x.streaming_api_base_url }
```
Gargron: Is this naming convention/structure okay with everybody? Or should we perhaps put this into a "links" or "urls" object?

clworld:
I don't think there is much need to add more URLs in future. (Only few URLs are required for make APIs working)
But wrapping URLs with some block is better design I think.

ikuradon:
Should we use like this?
```
node(:version) { Mastodon::Version.to_s }
node :urls do
    { :streaming_api_base => Rails.configuration.x.streaming_api_base_url }
end
```

Gargron:
@ikuradon Yes, that is good. I think "base" is not necessary

Gargron (approved):

PR #3562, opened 2017-06-04
Fix #3550 - Add all missing foreign keys

### LINKED ISSUE #3550: Foreign keys for account_id references?
Quite some time ago, before Mastodon refreshed infos from webfinger, I instructed another instance's admin to manually remove information on my remote account (that is, delete rows of the “accounts” database with domain = 'mydomain'). We are still unable to communicate now, and I suspect it is actually because of the instructions I have given this admin at the time (I am able to communicate with other instances just fine).

I tried to reproduce this issue and couldn't completely. However, in the process, I manually removed a remote account (which got re-created just fine afterwards), without removing the toots referencing it, leading to runtime errors.

It appears to me that `statuses (account_id)`, `follows (account_id)`, `follows (target_account_id)` etc. should all have a `foreign key` constraint on `accounts (id)`. Is there any reason why that is not the case?

Gargron:
The rule on media attachments table is nullify instead of cascade, because presumably we would want to keep the records so that we have a reference to the files that need to be deleted, otherwise those files would be orphaned.

In fact, this is kind of a concern for accounts table as well, since it keeps references to avatars and headers, which is why you should use `Account#destroy` instead of deleting it directly from the database, but I digress.

Another concern is that the database *will* complain and the migration will fail if there are violations of these constraints, which will require manual intervention from the person doing the upgrade. Suggestions of how to handle this are welcome.

### DIFF
--- a/db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
+++ b/db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
@@ -1,6 +1,8 @@
 class AddForeignKeysForAccounts < ActiveRecord::Migration[5.1]
   def change
     add_foreign_key :statuses, :accounts, on_delete: :cascade
+    add_foreign_key :statuses, :accounts, column: :in_reply_to_account_id, on_delete: :nullify
+    add_foreign_key :statuses, :statuses, column: :in_reply_to_id, on_delete: :nullify
     add_foreign_key :account_domain_blocks, :accounts, on_delete: :cascade
     add_foreign_key :conversation_mutes, :accounts, on_delete: :cascade
     add_foreign_key :conversation_mutes, :conversations, on_delete: :cascade
@@ -16,12 +18,15 @@ class AddForeignKeysForAccounts < ActiveRecord::Migration[5.1]
     add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
     add_foreign_key :imports, :accounts, on_delete: :cascade
     add_foreign_key :media_attachments, :accounts, on_delete: :nullify
+    add_foreign_key :media_attachments, :statuses, on_delete: :nullify
     add_foreign_key :mentions, :accounts, on_delete: :cascade
     add_foreign_key :mentions, :statuses, on_delete: :cascade
     add_foreign_key :notifications, :accounts, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, column: :from_account_id, on_delete: :cascade
     add_foreign_key :preview_cards, :statuses, on_delete: :cascade
     add_foreign_key :reports, :accounts, on_delete: :cascade
     add_foreign_key :reports, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :reports, :accounts, column: :action_taken_by_account_id, on_delete: :nullify
     add_foreign_key :statuses_tags, :statuses, on_delete: :cascade
     add_foreign_key :statuses_tags, :tags, on_delete: :cascade
     add_foreign_key :stream_entries, :accounts, on_delete: :cascade
--- a/db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
+++ b/db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
@@ -1,0 +1,36 @@
+class AddForeignKeysForAccounts < ActiveRecord::Migration[5.1]
+  def change
+    add_foreign_key :statuses, :accounts, on_delete: :cascade
+    add_foreign_key :account_domain_blocks, :accounts, on_delete: :cascade
+    add_foreign_key :conversation_mutes, :accounts, on_delete: :cascade
+    add_foreign_key :conversation_mutes, :conversations, on_delete: :cascade
+    add_foreign_key :favourites, :accounts, on_delete: :cascade
+    add_foreign_key :favourites, :statuses, on_delete: :cascade
+    add_foreign_key :blocks, :accounts, on_delete: :cascade
+    add_foreign_key :blocks, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follows, :accounts, on_delete: :cascade
+    add_foreign_key :follows, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :imports, :accounts, on_delete: :cascade
+    add_foreign_key :media_attachments, :accounts, on_delete: :nullify
+    add_foreign_key :mentions, :accounts, on_delete: :cascade
+    add_foreign_key :mentions, :statuses, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, on_delete: :cascade
+    add_foreign_key :preview_cards, :statuses, on_delete: :cascade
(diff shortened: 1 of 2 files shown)


db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,36 @@
+class AddForeignKeysForAccounts < ActiveRecord::Migration[5.1]
+  def change
+    add_foreign_key :statuses, :accounts, on_delete: :cascade
```
ThibG: There is still a missing foreign key for `in_reply_to_account_id`. Maybe the action for this one should be `nullify`.

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,36 @@
...
+    add_foreign_key :mutes, :accounts, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :imports, :accounts, on_delete: :cascade
+    add_foreign_key :media_attachments, :accounts, on_delete: :nullify
+    add_foreign_key :mentions, :accounts, on_delete: :cascade
+    add_foreign_key :mentions, :statuses, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, on_delete: :cascade
+    add_foreign_key :preview_cards, :statuses, on_delete: :cascade
+    add_foreign_key :reports, :accounts, on_delete: :cascade
+    add_foreign_key :reports, :accounts, column: :target_account_id, on_delete: :cascade
```
ThibG: I think it is still missing a foreign key for `action_taken_by_account_id`.

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,36 @@
...
+    add_foreign_key :follow_requests, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follows, :accounts, on_delete: :cascade
+    add_foreign_key :follows, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :imports, :accounts, on_delete: :cascade
+    add_foreign_key :media_attachments, :accounts, on_delete: :nullify
+    add_foreign_key :mentions, :accounts, on_delete: :cascade
+    add_foreign_key :mentions, :statuses, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, on_delete: :cascade
```
ThibG: I think `from_account_id` is still missing.

ClearlyClaire (changes requested):
I think some foreign keys are still missing. I only focused on `:accounts` for now.

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,36 @@
...
+    add_foreign_key :blocks, :accounts, on_delete: :cascade
+    add_foreign_key :blocks, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follows, :accounts, on_delete: :cascade
+    add_foreign_key :follows, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :imports, :accounts, on_delete: :cascade
+    add_foreign_key :media_attachments, :accounts, on_delete: :nullify
```
ThibG: I think `media_attachments` is still missing a foreign key to `statuses`.

ClearlyClaire (changes requested):
`media_attachments` is also missing a foreign key on `statuses`

ClearlyClaire (approved):
Looks good to me

Gargron:
@ThibG The only problem is that this migration might fail if the db contains records that violate these constraints. I don't know how to solve that. Doing data checks inside the migration will likely be taking too long, and given that migrations put a lock on the whole db, that's not acceptable

ThibG:
Indeed, but if this case arises, it means the database is in an inconsistent state, and that the admin should take action anyway. So I don't see the migration failing as much of an issue, as long as the failure is easily understandable and clear instructions are given (e.g., a command to remove database entries with unmet constraints).

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,41 @@
...
+    add_foreign_key :mentions, :statuses, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, on_delete: :cascade
+    add_foreign_key :notifications, :accounts, column: :from_account_id, on_delete: :cascade
+    add_foreign_key :preview_cards, :statuses, on_delete: :cascade
+    add_foreign_key :reports, :accounts, on_delete: :cascade
+    add_foreign_key :reports, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :reports, :accounts, column: :action_taken_by_account_id, on_delete: :nullify
+    add_foreign_key :statuses_tags, :statuses, on_delete: :cascade
+    add_foreign_key :statuses_tags, :tags, on_delete: :cascade
+    add_foreign_key :stream_entries, :accounts, on_delete: :cascade
```
akihikodaki: Stream entries are bound to statuses, aren't they?

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
```diff
@@ -0,0 +1,41 @@
...
+    add_foreign_key :blocks, :accounts, on_delete: :cascade
+    add_foreign_key :blocks, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, on_delete: :cascade
+    add_foreign_key :follow_requests, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :follows, :accounts, on_delete: :cascade
+    add_foreign_key :follows, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, on_delete: :cascade
+    add_foreign_key :mutes, :accounts, column: :target_account_id, on_delete: :cascade
+    add_foreign_key :imports, :accounts, on_delete: :cascade
+    add_foreign_key :media_attachments, :accounts, on_delete: :nullify
```
akihikodaki: Is there any particular intention by keeping media attachments alive when accounts are deleted, while the deletion cascades for statuses?

akihikodaki (changes requested):

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
Gargron: Polymorphic associations cannot be constrained via postgres foreign keys, unfortunately.

db/migrate/20170604144747_add_foreign_keys_for_accounts.rb
Gargron: >The rule on media attachments table is nullify instead of cascade, because presumably we would want to keep the records so that we have a reference to the files that need to be deleted, otherwise those files would be orphaned.

akihikodaki (approved):

clworld:
FYI: This is SQL I used to make db:migrate success in my environment. https://pastebin.com/GpbtGmbU
Other admins require additional SQLs for other tables maybe.

Gargron:
@clworld Perhaps I will add a rake task for this.

PR #3563, opened 2017-06-04
Fix #3063 - Add dynamic app manifest

### LINKED ISSUE #3063: App name and description in manifest.json should be unique per instance
Once Mastodon has a Service Worker (e.g. via #3052), Chrome and other browsers will pop up an "add to homescreen" banner such as the following:

out

You can also invoke this manually using "add to homescreen" on Chrome, Opera, Firefox, or Samsung Internet for Android and get the same effect.

The name and description come from `manifest.json`, which today is a static file, so it just says

```json
{
	"name": "Mastodon",
	"description": "Mastodon is a free, open-source social network server."
}
```

Ideally this should be unique per instance, so that when you install various instances to your homescreen, each one has a separate name. The description may also matter in cases where PWAs get ingested into the app store (which is currently the plan for Edge PWAs on Windows).

* * * *

Gargron:


app/views/manifests/show.json.rabl
```diff
@@ -0,0 +1,9 @@
+object false
+
+node(:name)        { Setting.site_title }
+node(:description) { strip_tags(Setting.site_description.presence || I18n.t('about.about_mastodon')) }
+node(:icons)       { [{ src: '/android-chrome-192x192.png', sizes: '192x192', type: 'image/png' }] }
```
nolanlawson: Is it possible to add a 512x512 icon as well? This is nice because it's shown on the "splash screen" when the app is launched from the home screen. If not, no biggie.

app/views/manifests/show.json.rabl
```diff
@@ -0,0 +1,9 @@
+object false
+
+node(:name)        { Setting.site_title }
```
nolanlawson: There should be a `short_name` as well as a `name`; the `short_name` is used for icons on the homescreen (whereas the `name` is used in the splash screen). In our case these can just be the same probably.

app/views/manifests/show.json.rabl
```diff
@@ -0,0 +1,9 @@
...
+node(:description) { strip_tags(Setting.site_description.presence || I18n.t('about.about_mastodon')) }
+node(:icons)       { [{ src: '/android-chrome-192x192.png', sizes: '192x192', type: 'image/png' }] }
+node(:theme_color) { '#282c37' }
```
nolanlawson: A `background_color` would be nice as well, since it's used as the background in the splash screen and thus helps improve "perceived performance." It seems for us the best value would be `$ui-secondary-color`.

app/views/manifests/show.json.rabl
```diff
@@ -0,0 +1,9 @@
...
+node(:display)     { 'standalone' }
```
nolanlawson: In the future it'd be nice if this were customizable on a per-user basis. Some toot.cafe users have said they would like to keep the URL bar when launched in "app" mode

nolanlawson (changes requested):
The only must-have I think is the `short_name` which AFAIK is required for Chrome to consider it a PWA worthy of an install banner (that plus a Service Worker). Otherwise great work, love it!

Gargron:
@nolanlawson Please re-review

nolanlawson (approved):
Looks great!

PR #3564, opened 2017-06-04
Add regex filters on the community timeline and the public timeline.

fvh-P:
Add regex filters on the community timeline and the public timeline. It is the same function of the filter on the home timeline.
I show an example of filters in the following image.
capture

I also added locale settings, which is required for this implementation. However, I do not really understand except Japanese and English. In the other languages, my additional sentences may contain some mistakes. (Actually, these additions are the copies from the settings of the home column.)


app/javascript/mastodon/locales/zh-TW.json
```diff
@@ -29,6 +29,9 @@
   "column_back_button.label": "上一頁",
   "column_subheading.navigation": "瀏覽",
   "column_subheading.settings": "設定",
+  "community.column_settings.advanced": "進階",
```
Gargron: Since all of these are the same, could you just rename the original key to something more generic than `home.column_settings.advanced` and then use that everywhere?

app/javascript/mastodon/features/public_timeline/components/setting_text.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
Gargron: I don't see any changes to this file so why not import the existing component instead of copypasting?

app/javascript/mastodon/features/public_timeline/components/column_settings.js
```diff
@@ -0,0 +1,39 @@
...
+import ColumnCollapsable from '../../../components/column_collapsable';
+import SettingToggle from '../../notifications/components/setting_toggle';
+import SettingText from './setting_text';
+
+const messages = defineMessages({
+  filter_regex: { id: 'public.column_settings.filter_regex', defaultMessage: 'Filter out by regular expressions' },
+  settings: { id: 'public.settings', defaultMessage: 'Column settings' },
+});
+
+class ColumnSettings extends React.PureComponent {
```
Gargron: Since this component is identical for both "public" and "community" columns, I suggest only creating it in one of those, and importing from the other.

Gargron (changes requested):
Code duplication issues

app/javascript/mastodon/locales/zh-TW.json
fvh-P: Some of these have been already changed according to the type of the timelines. For example, please see pt.json(Portuguese) and pt-BR.json(Portuguese in Brazil), It is set as following.
```"home.settings": "Parâmetros da listagem Home"```
and I set community and public as following.
```"community.settings": "Parâmetros da listagem Local"```
```"public.settings": "Parâmetros da listagem Global"```
Therefore, I made the original keys for such individual changes in each languages.

app/javascript/mastodon/features/public_timeline/components/setting_text.js
fvh-P: Oh, sorry.
I thought about the possibility of individual change, but this function must be same in other timelines!!
I will move ```app/javascript/mastodon/features/home_timeline/components/setting_text.js``` to a new file ```app/javascript/mastodon/components/setting_text.js``` and import this from all the requiring files.
Is it OK?

app/javascript/mastodon/features/public_timeline/components/column_settings.js
fvh-P: Yea, I agree with you.
But, I'm afraid that I'm not a React specialist, and I have no idea to change the behavior of ```defineMessages()``` according to the type of timeline.
Well, how should I modify this component?

app/javascript/mastodon/features/public_timeline/components/setting_text.js
Gargron: Yep, that's fine

app/javascript/mastodon/features/public_timeline/components/column_settings.js
Gargron: I would rather not specialize the strings. If some languages specialized them earlier, it's better to fix them than to have lots of near-duplicates.

fvh-P:
Thank you for @Gargron 's advising. I modified some codes to solve duplication. However, my additional commitment caused many conflicts... Would anyone check these?

app/javascript/mastodon/features/public_timeline/components/column_settings.js
fvh-P: Thank you for advising. I modified some coes to solve duplication.

-- I moved one of three ```setting_text.js``` files and deleted other two files.
-- I deleted ```public_timeline/components/colomn_settings.js``` and modified ```public_timeline/index.js``` importing ```community_timeline/components/colomn_settings.js```
-- I modified the locale settings. (deleted ```community.~~~``` and ```public.~~~```, and changed ```home.~~~``` into ```column_settings.~~~``` to use in these three column settings.)

However, my additional commitment caused many conflicts... Would you check these?

Gargron:
I think one way to solve the conflicts would be to `git checkout HEAD app/javascript/mastodon/locales` which will reset the state of the locale files, then merge master into your branch, then run `yarn run manage:translations` again. I believe that should do it.

Gargron (approved):

fvh-P:
I'm very sorry. :(
Even though I had tried several ways to solve the conflicts, they did not seem to be solved.
Hence, I changed two ```column_settings.js``` to use ```home.~~~``` in the locale files.

Gargron:
That's actually fine too. The name of the key may be not so descriptive anymore but it's the path of least change = less conflicts. That's okay.

fvh-P:
Thanks a lot to you for your reviews and advice with kindness!!!

PR #3566, opened 2017-06-04
Added support for configurable reserved usernames (fix of #1382)

Gargron:
Resolves #1382 

Couldn't modify the PR's original branch so I had no choice but to open a new PR.

### DIFF
--- a/app/validators/unreserved_validator.rb
+++ b/app/validators/unreserved_validator.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class UnreservedValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.nil?
+    record.errors.add(attribute, I18n.t('accounts.reserved_username')) if reserved_username?(value)
+  end
+
+  private
+
+  def reserved_username?(value)
+    return false unless Setting.reserved_usernames
+    Setting.reserved_usernames.include?(value.downcase)
+  end
+end
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -381,6 +381,12 @@ RSpec.describe Account, type: :model do
       expect(account_2).to model_have_error_on_field(:username)
     end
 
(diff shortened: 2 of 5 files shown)


app/validators/unreserved_validator.rb
```diff
@@ -0,0 +1,15 @@
...
+  def validate_each(record, attribute, value)
+    return if value.nil?
+    record.errors.add(attribute, I18n.t('accounts.reserved_username')) if reserved_username?(value)
+  end
+
+  private
+
+  def reserved_username?(value)
+    return false unless Setting.reserved_usernames
+    Setting.reserved_usernames.include?(value.downcase)
```
mjankowski: Could this be one line of `Setting.reserved_usernames.to_a.include?(...)` ... I'm assuming this returns `nil` and not `[]` when it's not set?

spec/models/account_spec.rb
```diff
@@ -381,6 +381,12 @@
       expect(account_2).to model_have_error_on_field(:username)
     end
 
+    it 'is invalid if the username is reserved' do
+      account = Fabricate.build(:account, username: 'support')
+      account.valid?
```
mjankowski: Maybe `expect(account).not_to be_valid?` here? - to both trigger calling of valid to populate errors, but also to assert that the the `valid?` call result is false?

app/validators/unreserved_validator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class UnreservedValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.nil?
```
mjankowski: Could we put the call to this validator on a separate line in `Account` like:

```
validates :username, unreserved: true, if: :username?
```

so that we only bother calling this validator when a value is there?

mjankowski (approved):
This is probably fine as-is ... left a few comments around style that you can take or leave at your discretion.

PR #3575, opened 2017-06-05
Don't follow account if it's already followed

reneklacan:
Closes https://github.com/tootsuite/mastodon/issues/3102


Gargron (approved):

reneklacan:
@Gargron Failure on the CI doesn't seem to be related to my change, spec is passing both on 2.3.4 and on my machine with 2.4.1... Is it maybe just bit flaky spec?

Gargron:
Yep, never mind that.

PR #3579, opened 2017-06-05
Redirect to streaming_api_base_url

zunda:
Alternative solution from #3556 for #2068.

When Rails receives a request to streaming API, it most likely means that there is another host which is configured to respond to it. This is to redirect clients to that host if `STREAMING_API_BASE_URL` is set as another host.

### DIFF
--- a/app/controllers/api/v1/streaming_controller.rb
+++ b/app/controllers/api/v1/streaming_controller.rb
@@ -7,7 +7,7 @@ class Api::V1::StreamingController < ApiController
     if Rails.configuration.x.streaming_api_base_url != request.host
       uri = URI.parse(request.url)
       uri.host = URI.parse(Rails.configuration.x.streaming_api_base_url).host
-      redirect_to uri.to_s, :status => 301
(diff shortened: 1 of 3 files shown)


Gargron (approved):

app/controllers/api/v1/streaming_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Api::V1::StreamingController < ApiController
+  respond_to :json
+
+  def index
+    if Rails.configuration.x.streaming_api_base_url != request.host
+      uri = URI.parse(request.url)
+      uri.host = URI.parse(Rails.configuration.x.streaming_api_base_url).host
+      redirect_to uri.to_s, :status => 301
```
Gargron: >Use the new Ruby 1.9 hash syntax

Gargron (changes requested):
Code style

Gargron (approved):

PR #3580, opened 2017-06-05
Fix randomly fail (regression from #3560)

ykzts:
@akihikodaki typo?


akihikodaki (approved):
Absolutely it's typo.

PR #3592, opened 2017-06-05
Fix db:seed - only run some validations when the field was changed

Gargron:
I'm not 100% sure how to test this. But I believe something like this is good practice, I have seen this pattern in the GitLab codebase. However, correct me if I'm wrong on something.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -52,13 +52,17 @@ class Account < ApplicationRecord
   has_one :user, inverse_of: :account
 
   validates :username, presence: true
-  validates :username, uniqueness: { scope: :domain, case_sensitive: true }, unless: :local?
+
+  # Remote user validations
+  with_options unless: :local? do
+    validates :username, uniqueness: { scope: :domain, case_sensitive: true }, if: :username_changed?
+  end
 
   # Local user validations
   with_options if: :local? do
-    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, unreserved: true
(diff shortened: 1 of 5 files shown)


mjankowski:
The benefit here is some smallish time savings by avoiding validations which aren't going to be relevant for certain fields? I guess that's fine to do.

If you wanted to add specs for those validation skips - you could make records with invalid values for these skipped fields, and then have the spec expect them to be valid, and/or expect there to be no errors on a field which we know to actually be invalid - but which did not change as part of the valid/save step.

mjankowski (commented):
The benefit here is some smallish time savings by avoiding validations which aren't going to be relevant for certain fields? I guess that's fine to do.

If you wanted to add specs for those validation skips - you could make records with invalid values for these skipped fields, and then have the spec expect them to be valid, and/or expect there to be no errors on a field which we know to actually be invalid - but which did not change as part of the valid/save step.

Gargron:
I added tests. Please review/approve!

joyeusenoelle (approved):
This all looks valid to me!

ykzts (approved):

mjankowski (approved):

clworld:
I'm getting 'Note is too long' error in RemoteProfileUpdateWorker.
It seems "with_options if: :local?" is ignored.
This comment says nested conditions are overrided: https://github.com/tootsuite/mastodon/pull/3612#issuecomment-307101972
With reverting my instance to yesterdays version, this exception does not happen.

long_note

Gargron:
@clworld Thank you for reporting this so soon. I will try to submit a fix immediately.

Gargron:
@clworld #3644

PR #3595, opened 2017-06-05
Use preview image in <ImageLoader /> to provide immediate visual feedback

Gargron:
(before the full-size image is loaded)


nolanlawson:
The blue is a nice touch too. This looks great; I can't test it ATM but codewise it LGTM.

ykzts (approved):
I checked it with the browsers (Google Chrome, Firefox, Safari).

PR #3598, opened 2017-06-05
Introduce RelationshipMap class to encapsulate relationship API logic

mjankowski:
I have a code history question which I'll post as a review here prior to this being merge-ready.

### DIFF
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -5,6 +5,7 @@ class Api::V1::AccountsController < Api::BaseController
   before_action -> { doorkeeper_authorize! :follow }, only: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
   before_action :require_user!, except: [:show]
   before_action :set_account
+  before_action :set_relationship_map, except: [:show, :block]
 
   respond_to :json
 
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/accounts_controller.rb
```diff
@@ -5,51 +5,39 @@ class Api::V1::AccountsController < ApiController
...
   def follow
     FollowService.new.call(current_user.account, @account.acct)
-    set_relationship
     render :relationship
   end
 
   def block
     BlockService.new.call(current_user.account, @account)
-
-    @following       = { @account.id => false }
```
mjankowski: This commit describes what I think the history here was - https://github.com/tootsuite/mastodon/pull/3598/commits/7ce3d9afeb90cb50873dfbedb5d60a14932b2a0a - would be good to confirm.

Gargron: These explicit values are a cheat. Because we know exactly what the relationship we're gonna get after the above service call, we can just hardcode the value instead of doing a roundtrip to the database.

mjankowski: Alright - so it sounds like we should preserve the hardcoding here, even though we're back to using service instead of the worker?

mjankowski:
Updated to restore the dont-hit-the-DB approach for blocks, and added some coverage for the new mapper class.

app/lib/blocked_relationship_map.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class BlockedRelationshipMap
```
Gargron: Perhaps the other `RelationshipMap` would simply allow you to override some of the methods with a static value?

mjankowski:
This is awkward and I can't think of an obvious way to make it better ... closing PR for now, will keep branch around and improve if I think of something.

Issue #3388, opened 2017-05-28
raise_record_not_found_exception


Jehops:
I have started noticing these exceptions in the background workers log.  It's possible, though I'm not certain, they started after upgrading to 1.4rc5.

```
2017-05-28T10:53:32.639Z 85437 TID-9n3irk WARN: ActiveRecord::RecordNotFound: Couldn't find Status with 'id'=122
2017-05-28T10:53:32.639Z 85437 TID-9n3irk WARN: /usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/relation/finder_methods.rb:353:in `raise_record_not_found_exception!'
/usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/relation/finder_methods.rb:479:in `find_one'
/usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/relation/finder_methods.rb:458:in `find_with_ids'
/usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/relation/finder_methods.rb:66:in `find'
/usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/querying.rb:3:in `find'
/usr/local/lib/ruby/gems/2.3/gems/activerecord-5.0.3/lib/active_record/core.rb:151:in `find'
/usr/local/www/mastodon/app/workers/thread_resolve_worker.rb:9:in `perform'
/usr/local/lib/ruby/gems/2.3/gems/sidekiq-5.0.0/lib/sidekiq/processor.rb:204:in `execute_job'
/usr/local/lib/ruby/gems/2.3/gems/sidekiq-5.0.0/lib/sidekiq/processor.rb:175:in `block (2 levels) in process'
/usr/local/lib/ruby/gems/2.3/gems/sidekiq-5.0.0/lib/sidekiq/middleware/chain.rb:128:in `block in invoke'
/usr/local/lib/ruby/gems/2.3/gems/sidekiq-unique-jobs-5.0.8/lib/sidekiq_unique_jobs/server/middleware.rb:17:in `call'
/usr/local/lib/ruby/gems/2.3/gems/sidekiq-5.0.0/lib/sidekiq/middleware/chain.rb:130:in `block in invoke'
[...]

PR #3599, opened 2017-06-05
Try fixing ThreadResolveWorker calls

ClearlyClaire:
This is an attempt at fixing #3388.

From my understanding of ActiveRecord, a transaction is commited as soon as
the exit of the outmost ActiveRecord.transaction block. However, inner
transaction blocks will exit without the transaction being commited.

In this case, ThreadResolveWorker were fired *within* a transaction block,
so moving the call out of it should do the trick. However, this is somewhat
fragile, as this whole codepath could be called within yet another transaction.

### DIFF
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -72,6 +72,12 @@ class ProcessFeedService < BaseService
         status.save!
       end
 
+      if thread?(@xml)
+        Rails.logger.debug "Trying to attach #{status.id} (#{id(@xml)}) to #{thread(@xml).first}"
+        status.thread = find_or_resolve_status(status, *thread(@xml))
+        status.save!
+      end
+
       notify_about_mentions!(status) unless status.reblog?
       notify_about_reblog!(status) if status.reblog? && status.reblog.account.local?
 
@@ -153,11 +159,6 @@ class ProcessFeedService < BaseService
         visibility: visibility_scope(entry),
         conversation: find_or_create_conversation(entry)
       )
-
-      if thread?(entry)
-        Rails.logger.debug "Trying to attach #{status.id} (#{id(entry)}) to #{thread(entry).first}"
-        status.thread = find_or_resolve_status(status, *thread(entry))
-      end
 
       mentions_from_xml(status, entry)
       hashtags_from_xml(status, entry)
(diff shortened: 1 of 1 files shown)


Gargron:
Oh, I see - the problem was that ThreadResolveWorker starts working before the transaction is committed, therefore it can't find the parent status it's supposed to act on. Good catch. Not sure if there's a downside to moving the in_reply_to_id setting out of the transaction. I mean, there is - bad result will get returned and cached from some of the public timeline API calls.

Ideally we would only move the async call out of the transaction, not the initial value if it's available...

ThibG:
@Gargron hm, I don't think I'm changing the behaviour here. Only downside I see is that I'm making an extra `save!` call… maybe guarding it with `status.changed?` would make sense.

app/services/process_feed_service.rb
```diff
@@ -72,6 +72,12 @@ def create_status
         status.save!
       end
 
+      if thread?(@xml)
+        Rails.logger.debug "Trying to attach #{status.id} (#{id(@xml)}) to #{thread(@xml).first}"
+        status.thread = find_or_resolve_status(status, *thread(@xml))
+        status.save!
```
ThibG: Maybe `status.save! if status.changed?` would make sense.

Gargron:
@ThibG While we're inside the transaction block, calls to /api/v1/timelines/public will *not* return the status, because it doesn't exist yet. But it starts existing as soon as the transaction block commits. So committing it with an unset in_reply_to_id value will give clients an "early" version of the status, with an unset value. The caching layer will save it (though indeed, it should invalidate itself on next save! call)

ThibG:
@Gargron ah I see. And even with the `changed?` that would make one avoidable `save!` call. The best thing to do is probably to only move the async call, yeah…

Gargron (approved):

app/services/process_feed_service.rb
```diff
@@ -69,9 +69,16 @@ def create_status
           end
         end
 
+        status.thread = find_status(thread(@xml).first) if thread?(xml)
```
Gargron: Need `@xml` instead of `xml` here (Travis failure)

Gargron (changes requested):

app/services/process_feed_service.rb
ThibG: Oops, sorry you're right. Should be ok now!

Gargron (approved):

PR #3606, opened 2017-06-06
Fixes #3605 by returning account from database in case of race condition

### LINKED ISSUE #3605: Race condition when creating remote account information
Monitoring my sidekiq logs, I found a failure that prevented a toot from being processed and thus broke a thread: `ActiveRecord::RecordNotUnique` was raised because of a duplicate remote account was created.

```
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/follow_remote_account_service.rb:64:in `call'
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/concerns/author_extractor.rb:21:in `author_from_xml'
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/fetch_remote_account_service.rb:24:in `process_atom'
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/fetch_remote_account_service.rb:15:in `call'
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/process_feed_service.rb:217:in `account_from_href'
juin 06 03:10:23 neiro bundle[14946]: /home/mastodon/live/app/services/process_feed_service.rb:200:in `block in mentions_from_xml'
```

Looking at the database, the record for the remote user has indeed been created mere seconds before the error. This is unsurprising as `FollowRemoteAccountService` does a few HTTP requests between checking for the existence of a remote user in the database and saving it.

I'm not sure what the most appropriate fix would be, but rescuing `FollowRemoteAccountService#call` with a lookup on `ActiveRecord::RecordNotUnique` should do the trick.

* * * *

ClearlyClaire:
A better solution would probably be to ensure that there are no concurrent `FollowRemoteAccountService` calls in the first place, but this should do the trick.

I haven't properly tested it, as it is a bit of pain to reproduce.

### DIFF
--- a/app/services/follow_remote_account_service.rb
+++ b/app/services/follow_remote_account_service.rb
@@ -61,8 +61,13 @@ class FollowRemoteAccountService < BaseService
     account.uri     = get_account_uri(xml)
     account.hub_url = hubs.first.attribute('href').value
 
-    account.save!
-    get_profile(body, account)
+    begin
+      account.save!
+      get_profile(body, account)
+    rescue ActiveRecord::RecordNotUnique
+      # The account has been added by another worker!
+      return Account.find_remote(confirmed_username, confirmed_domain)
+    end
 
     account
   end


Gargron:
Maybe the cleaner solution would be to replace "new" with "create" a few lines above? Given that by that point, Webfinger has returned an answer, so we know it's *an* account, but the tradeoff would be creating accounts that don't have a salmon URL or hub URL etc. That might be problematic. Another approach might be locks via redis, though I have not looked into that. This solution might be the best one in the short run though.

cc @mjankowski @krainboltgreene

ThibG:
We could use `create` instead of `new`, and most of the fields' values are known at this point (including the salmon URL), but I have two issues with this:
1. While most fields' values are known at this point, `account.uri` is not, and that's perhaps the most important field with the username and domain
2. If we go that route, we need to be careful to destroy the account if an error occurs when fetching the feed

Gargron:
I am tempted to accept your PR as-is as there has been no input from anyone else and it seems to do the job.

Gargron (approved):

PR #3610, opened 2017-06-06
[Account] Skip unreserved validation for existing model.

treby:
by #3566, implemented reserved username. 
But it seems to enable validation even for `persisted` account. Is this intentional?

I noticed the issue by running `rails db:seed` , I may have to fix `db:seed` script, not validation :)

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -56,7 +56,8 @@ class Account < ApplicationRecord
 
   # Local user validations
   with_options if: :local? do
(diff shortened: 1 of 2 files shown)


Gargron:
Duplicate of #3592, but you have tests that I didn't put into mine yet. What do we do?

treby:
I didn't know the PR :P
I think #3592 seems good. So I'll add only test for it if you want.

treby:
created https://github.com/tootsuite/mastodon/pull/3612 , then I'll close this.

PR #3611, opened 2017-06-06
Fix tag search order and not to use tsvector

### LINKED ISSUE #3591: Hashtag search does not find tag itself - only similar tags
If you enter a hashtag at the search bar, only similar hashtags appear as suggestions, but the tag itself is not displayed. :( 

Examle: Search for "#linux": "#linux-gaming" and "#linux-kernel" are displayed as search results and if you click them, you'll get suitable posts - but "#linux" is not displayed, so you cannot find posts tagged with just "#linux". 

bildschirmfoto von 2017-06-05 18-02-39
bildschirmfoto von 2017-06-05 18-02-29

* * * *

unarist:
* Sort results by the name (fixes #3591)
* Switch search method to simple `LIKE` matching instead of tsvector/tsquery

Previously we used scores from ts_rank_cd() to sort results, but it didn't work because the function returns same score for all results. It's not for calculate similarity of single words. Sometimes this bug even push out exact matching tag from results.

Additionally, PostgreSQL supports prefix searching with standard btree index. Using it offers simpler code, but also less index size and some speed.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -21,22 +21,9 @@ class Tag < ApplicationRecord
   end
 
   class << self
-    def search_for(terms, limit = 5)
-      terms      = Arel.sql(connection.quote(terms.gsub(/['?\\:]/, ' ')))
-      textsearch = 'to_tsvector(\'simple\', tags.name)'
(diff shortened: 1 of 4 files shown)


Gargron (approved):

Gargron:
There's just two style issues. Please use single-quotes when there is no string interpolation.

PR #3612, opened 2017-06-06
Add account specs

treby:
Try to add specs for https://github.com/tootsuite/mastodon/pull/3592 .

I know my poor English, so correct these adequately, please.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -52,18 +52,12 @@ class Account < ApplicationRecord
   has_one :user, inverse_of: :account
 
   validates :username, presence: true
-
-  # Remote user validations
-  with_options unless: :local? do
-    validates :username, uniqueness: { scope: :domain, case_sensitive: true }, if: :username_changed?
-  end
+  validates :username, uniqueness: { scope: :domain, case_sensitive: true }, if: :username_changed?
 
   # Local user validations
-  with_options if: :local? do
-    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, unreserved: true, if: :username_changed?
-    validates :display_name, length: { maximum: 30 }, if: :display_name_changed?
-    validates :note, length: { maximum: 160 }, if: :note_changed?
-  end
+  validates :username, format: { with: /\A[a-z0-9_]+\z/i }, length: { maximum: 30 }, unreserved: true, if: -> { local? && username_changed? }
+  validates :display_name, length: { maximum: 30 }, if: -> { local? && display_name_changed? }
+  validates :note, length: { maximum: 160 }, if: -> { local? && note_changed? }
 
   # Timelines
   has_many :stream_entries, inverse_of: :account, dependent: :destroy
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -362,42 +362,99 @@ RSpec.describe Account, type: :model do
   end
 
   describe 'validations' do
-    it 'has a valid fabricator' do
-      account = Fabricate.build(:account)
-      account.valid?
-      expect(account).to be_valid
+    subject { account }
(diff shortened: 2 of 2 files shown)


app/models/account.rb
```diff
@@ -52,18 +52,12 @@ class Account < ApplicationRecord
...
-  end
+  validates :username, uniqueness: { scope: :domain, case_sensitive: true }, if: :username_changed?
 
   # Local user validations
-  with_options if: :local? do
-    validates :username, format: { with: /\A[a-z0-9_]+\z/i }, uniqueness: { scope: :domain, case_sensitive: false }, length: { maximum: 30 }, unreserved: true, if: :username_changed?
-    validates :display_name, length: { maximum: 30 }, if: :display_name_changed?
-    validates :note, length: { maximum: 160 }, if: :note_changed?
-  end
+  validates :username, format: { with: /\A[a-z0-9_]+\z/i }, length: { maximum: 30 }, unreserved: true, if: -> { local? && username_changed? }
```
mjankowski: I think the `with_options` usage here was more clear and this change is awkward.

spec/models/account_spec.rb
```diff
@@ -362,42 +362,99 @@
...
+      it_behaves_like 'not enabled for existing record'
+    end
+
+    context 'local? => false' do
+      let(:domain) { 'mastodon.social' }
+
+      it_behaves_like 'check username validation'
+      it_behaves_like 'not allowed duplicate username'
+
+      describe 'not insterested in username format.' do
```
mjankowski: Typo 'insterested'

spec/models/account_spec.rb
```diff
@@ -362,42 +362,99 @@
...
-        account.valid?
-        expect(account).to model_have_error_on_field(:username)
+    describe 'reserved username' do
+      let(:username) { 'support' }
+
+      it { is_expected.not_to be_valid }
+      it_behaves_like 'not enabled for existing record'
+    end
+
+    context 'local? => false' do
```
mjankowski: I would just call this `context 'A remote account'` or something, instead of the `local?` impl detail.

spec/models/account_spec.rb
```diff
@@ -362,42 +362,99 @@
...
+    subject { account }
+    let(:account) { Fabricate.build(:account, username: username, display_name: display_name, note: note, domain: domain) }
+    let(:domain) { nil } # domain = nil means the account is local.
+    let(:username) { 'Gargron' }
+    let(:display_name) { 'Eugen Rochko' }
+    let(:note) { 'Mastodon is a GNU Social-compatible microblogging server.' }
+
+    it { is_expected.to be_valid }
+
+    shared_examples 'check username validation' do
```
mjankowski: Thoughts on using a gem like shoulda-matchers or something instead of these?

mjankowski (changes requested):

app/models/account.rb
treby: I thought so, too.
But in fact, the options insides block will override `with_options`'s one.

```ruby
with_options if: :local? do
  validates :username, length: { maximum: 30 }, if: :username_changed?
end
```

We expect (and I expected) this will evaluate

```ruby
validates :username, length: { maximum: 30 }, if: -> { local? && username_changed? }
```
but actually

```ruby
validates :username, length: { maximum: 30 }, if: :username_changed?
```

the `local?` condition is not affected...
You can check this by reverting first commit (and some tests will fail).

spec/models/account_spec.rb
treby: Unfortunately, I'm not familiar with the gem. But I wanna try to use.
Could you tell me some sample codes you think about this?

Gargron:
I have added the relevant tests in #3592 I think those are enough

treby:
I'm afraid it causes some bugs.

> Each nesting level will merge inherited defaults in addition to their own.

https://github.com/rails/rails/blob/v5.1.1/activesupport/lib/active_support/core_ext/object/with_options.rb#L51-L63

PR #3614, opened 2017-06-06
Add rake task to prepare database for foreign keys introduced by #3562

Gargron:
Careful review is required. Is this actually the fastest way to do this?

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -212,7 +212,7 @@ namespace :mastodon do
       ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN users u ON u.id = oag.resource_owner_id WHERE oauth_access_grants.id = oag.id AND oag.resource_owner_id IS NOT NULL AND u.id IS NULL')
       ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN oauth_applications a ON a.id = oag.application_id WHERE oauth_access_grants.id = oag.id AND oag.application_id IS NOT NULL AND a.id IS NULL')
       ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN users u ON u.id = oat.resource_owner_id WHERE oauth_access_tokens.id = oat.id AND oat.resource_owner_id IS NOT NULL AND u.id IS NULL')
-      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN oauth_applications a ON a.id = oat.application_id WHERE oauth_access_tokens.id = oat.id AND oag.application_id IS NOT NULL AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN oauth_applications a ON a.id = oat.application_id WHERE oauth_access_tokens.id = oat.id AND oat.application_id IS NOT NULL AND a.id IS NULL')
 
       # All the nullifies:
       ActiveRecord::Base.connection.execute('UPDATE statuses SET in_reply_to_id = NULL FROM statuses s LEFT JOIN statuses rs ON rs.id = s.in_reply_to_id WHERE statuses.id = s.id AND s.in_reply_to_id IS NOT NULL AND rs.id IS NULL')
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -177,5 +177,49 @@ namespace :mastodon do
 
       Rails.logger.debug 'Done!'
     end
+
+    desc 'Ensure referencial integrity'
+    task prepare_for_foreign_keys: :environment do
+      # All the deletes:
+      ActiveRecord::Base.connection.execute('DELETE FROM statuses USING statuses s LEFT JOIN accounts a ON a.id = s.account_id WHERE statuses.id = s.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM account_domain_blocks USING account_domain_blocks adb LEFT JOIN accounts a ON a.id = adb.account_id WHERE account_domain_blocks.id = adb.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM conversation_mutes USING conversation_mutes cm LEFT JOIN accounts a ON a.id = cm.account_id WHERE conversation_mutes.id = cm.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM conversation_mutes USING conversation_mutes cm LEFT JOIN conversations c ON c.id = cm.conversation_id WHERE conversation_mutes.id = cm.id AND c.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM favourites USING favourites f LEFT JOIN accounts a ON a.id = f.account_id WHERE favourites.id = f.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM favourites USING favourites f LEFT JOIN statuses s ON s.id = f.status_id WHERE favourites.id = f.id AND s.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM blocks USING blocks b LEFT JOIN accounts a ON a.id = b.account_id WHERE blocks.id = b.id AND a.id IS NULL')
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -177,5 +177,49 @@ namespace :mastodon do
...
+      ActiveRecord::Base.connection.execute('DELETE FROM preview_cards USING preview_cards pc LEFT JOIN statuses s ON s.id = pc.status_id WHERE preview_cards.id = pc.id AND s.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM reports USING reports r LEFT JOIN accounts a ON a.id = r.account_id WHERE reports.id = r.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM reports USING reports r LEFT JOIN accounts a ON a.id = r.target_account_id WHERE reports.id = r.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM statuses_tags USING statuses_tags st LEFT JOIN statuses s ON s.id = st.status_id WHERE statuses_tags.tag_id = st.tag_id AND statuses_tags.status_id = st.status_id AND s.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM statuses_tags USING statuses_tags st LEFT JOIN tags t ON t.id = st.status_id WHERE statuses_tags.tag_id = st.tag_id AND statuses_tags.status_id = st.status_id AND t.id IS NULL')
```
clworld: 'LEFT JOIN tags t ON t.id = st.status_id' -> 'LEFT JOIN tags t ON t.id = st.tag_id' is correct

lib/tasks/mastodon.rake
```diff
@@ -177,5 +177,49 @@ namespace :mastodon do
...
+      ActiveRecord::Base.connection.execute('DELETE FROM web_settings USING web_settings ws LEFT JOIN users u ON u.id = ws.user_id WHERE web_settings.id = ws.id AND u.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN users u ON u.id = oag.resource_owner_id WHERE oauth_access_grants.id = oag.id AND u.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN oauth_applications a ON a.id = oag.application_id WHERE oauth_access_grants.id = oag.id AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN users u ON u.id = oat.resource_owner_id WHERE oauth_access_tokens.id = oat.id AND u.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN oauth_applications a ON a.id = oat.application_id WHERE oauth_access_tokens.id = oat.id AND a.id IS NULL')
```
clworld: I have many records of oauth_access_tokens with application_id IS NULL and this SQL remove them, but It seems my mistake….(I deleted oauth_applications record for 'Web' by mistake.)

clworld (commented):
I run all SQLs in my environment that is already migrated.(All SQLs must be no-op.)
2 SQLs updated/deleted non-zero records.
1 SQL is wrong, and 1 SQL is correct I think.

lib/tasks/mastodon.rake
Gargron: I have run this on a local copy of the mastodon.social database, and it deleted 2149 records. Which sounds like a big number. I'm not sure what's going on there...

Gargron:
I have run this on a local copy of production data (millions of records). It doesn't take too long and most queries don't delete much (because constraint violations are not that common, the most violations are related to the statuses table)

I have a bit of second thoughts about adding foreign keys to the oauth_* tables, since Doorkeeper didn't add them so maybe they knew better. Maybe there is a reason for oauth_access_tokens to have a NULL in application_id

unarist:
I like `NOT EXISTS` than `LEFT JOIN ... WHERE a.id IS NULL` for readability. Also it may makes bit faster:

```
DELETE FROM statuses USING statuses s LEFT JOIN accounts a ON a.id = s.account_id WHERE statuses.id = s.id AND a.id IS NULL;
 
Delete on statuses  (cost=2.61..4.63 rows=1 width=18)
 +-Nested Loop  (cost=2.61..4.63 rows=1 width=18)
    +-Hash Anti Join  (cost=2.47..4.26 rows=1 width=20)  Hash Cond: (s.account_id = a.id)
    |  +-Seq Scan on statuses s  (cost=0.00..1.57 rows=57 width=18)
    |  +-Hash  (cost=2.21..2.21 rows=21 width=10)
    |     +-Seq Scan on accounts a  (cost=0.00..2.21 rows=21 width=10)
    +-Index Scan using statuses_pkey on statuses  (cost=0.14..0.37 rows=1 width=14)  Index Cond: (id = s.id)
 
DELETE FROM statuses WHERE NOT EXISTS (SELECT 1 FROM accounts a WHERE a.id = statuses.account_id);
 
Delete on statuses  (cost=2.47..4.26 rows=1 width=12)
 +-Hash Anti Join  (cost=2.47..4.26 rows=1 width=12)  Hash Cond: (statuses.account_id = a.id)
    +-Seq Scan on statuses  (cost=0.00..1.57 rows=57 width=10)
    +-Hash  (cost=2.21..2.21 rows=21 width=10)
       +-Seq Scan on accounts a  (cost=0.00..2.21 rows=21 width=10)
```

lib/tasks/mastodon.rake
Gargron: Hmm actually I think I need to amend those queries with a "NOT NULL" as well, because there are reasonable cases where resource_owner_id is NULL, such as "client credentials" flow. That might explain why these ones deleted over 2k records from my dummy database.

Gargron:
@unarist Okay, I finally understand. I've read that the Postgres query plan for "NOT EXISTS" and "LEFT JOIN IS NULL" is identical. However in the LEFT JOIN case it has to match the row from the "DELETE" to a row from the "USING" query, which creates the index scan and the nested loop. I am assuming that in the "NOT EXISTS" case, the data is taken from the "DELETE" query directly. If that's the case, the latter really is slightly more performant.

beatrix-bitrot:
mmmmmmm
`
bundle exec rake mastodon:maintenance:prepare_for_foreign_keys
rake aborted!
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "oag"
LINE 1: ...tion_id WHERE oauth_access_tokens.id = oat.id AND oag.applic...
                                                             ^
: DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN oauth_applications a ON a.id = oat.application_id WHERE oauth_access_tokens.id = oat.id AND oag.application_id IS NOT NULL AND a.id IS NULL
`
any idea what's up here?

lib/tasks/mastodon.rake
```diff
@@ -209,10 +209,10 @@ namespace :mastodon do
...
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN users u ON u.id = oag.resource_owner_id WHERE oauth_access_grants.id = oag.id AND oag.resource_owner_id IS NOT NULL AND u.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_grants USING oauth_access_grants oag LEFT JOIN oauth_applications a ON a.id = oag.application_id WHERE oauth_access_grants.id = oag.id AND oag.application_id IS NOT NULL AND a.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN users u ON u.id = oat.resource_owner_id WHERE oauth_access_tokens.id = oat.id AND oat.resource_owner_id IS NOT NULL AND u.id IS NULL')
+      ActiveRecord::Base.connection.execute('DELETE FROM oauth_access_tokens USING oauth_access_tokens oat LEFT JOIN oauth_applications a ON a.id = oat.application_id WHERE oauth_access_tokens.id = oat.id AND oag.application_id IS NOT NULL AND a.id IS NULL')
```
unarist: > oag.application_id IS NOT NULL

should be `oat.application_id IS NOT NULL`

clworld (commented):
I checked:
- All tables in SQLs joined properly.
- There is no difference between SQLs and foreign keys.
- All SQLs updates/deletes 0 records for my environment that is already migrated.

It seems OK. Looks good for me.
Getting additional reviews by someone is better I think.

beatrix-bitrot (approved):
looks good to me!

PR #3615, opened 2017-06-06
Add missing word in pt and pt-BR (regression from #3564)

ykzts:
https://translate.google.com/#pt/en/Par%C3%A2metros%20da%20listagem%20Home

I feel I can not erase it.

cc @fvh-P

### DIFF
--- a/app/javascript/mastodon/locales/pt-BR.json
+++ b/app/javascript/mastodon/locales/pt-BR.json
@@ -79,7 +79,7 @@
   "home.column_settings.filter_regex": "Filtrar com uma expressão regular",
   "home.column_settings.show_reblogs": "Mostrar as partilhas",
   "home.column_settings.show_replies": "Mostrar as respostas",
-  "home.settings": "Parâmetros da listagem",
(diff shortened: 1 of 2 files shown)


ykzts:
cc @hugogameiro ? (#1966)

hugogameiro:
@ykzts is your question about how to translate "home" or something else?

ykzts:
@hugogameiro In #3564 the word "Home" has gone. Is this OK?

hugogameiro:
@ykzts I removed it because I didn't find a proper word to translate it. Where I could not remove it, I just didn't translate it, in there it made sense to remove it because you would only see that text when you were already in the home column/feed.
But what you suggest here is ok and probably is better for future proof.

hugogameiro (approved):

ykzts:
@hugogameiro thank you!

Gargron:
That key is now used for settings on other timelines too. So it's good that the in-string specification is gone.

ykzts:
@Gargron Hmm, I got it. I close this PR.

PR #3616, opened 2017-06-06
Revocable sessions

sorin-davidoi:
As discussed in #3243, this introduces revocable sessions as explained [here](http://www.jonathanleighton.com/articles/2013/revocable-sessions-with-devise/).

Left to do:
 - [x] Expose the maximum number of sessions as a configuration option (in `config/environments`?)
 - [x] Make tests pass (see [this discussion](https://github.com/tootsuite/mastodon/pull/3243#discussion_r119736357))

### DIFF
--- a/app/models/session_activation.rb
+++ b/app/models/session_activation.rb
@@ -11,7 +11,7 @@
 #
 
 class SessionActivation < ApplicationRecord
-  LIMIT = 20
+  LIMIT = Rails.configuration.x.max_session_activations
 
   def self.active?(id)
     id && where(session_id: id).exists?
--- a/app/models/session_activation.rb
+++ b/app/models/session_activation.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+# == Schema Information
+#
+# Table name: session_activations
+#
+#  id         :integer          not null, primary key
+#  user_id    :integer          not null
+#  session_id :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class SessionActivation < ApplicationRecord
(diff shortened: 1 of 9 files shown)


sorin-davidoi:
Where should the maximum number of session be set? Was thinking of `config/settings.yml`, but it seams that all those options are configurable from the Web UI.

Gargron:
@sorin-davidoi Probably a config/initializers/ file. What maximum number of sessions do you mean though? Why should there be a maximum?

sorin-davidoi:
With this modification, each login creates a new entry in the `session_activations` table. If the user frequently switches devices (or uses "private browsing" or just logs in frequently), I imagine this could lead to some space problems. So we place a limit on the number of active sessions (if `LIMIT` is 5 and your current login creates the sixth session, the first one gets deleted).

Gargron:
@sorin-davidoi I wonder if it would be more appropriate to store sessions in Redis rather than Postgres.

sorin-davidoi:
Not familiar at all with Redis, but I guess that would make #3243 harder? We need to link the subscription to the session that created it, to make sure that they get deleted together.

sorin-davidoi:
Added the configuration option.

sorin-davidoi:
Anything else I can do here to help with this?

Gargron:
It'd be cool to have some input from @mjankowski or @krainboltgreene on this, looks okay to me but perhaps they had to implement something similar for different applications.

Gargron (approved):

sorin-davidoi:
Rebase, moved the migration date to today.

PR #3623, opened 2017-06-07
Allow "class" attribute on the "a" tag in sanitization

unarist:
We use `<a ... class="u-url mention">` for mention links in profiles and statuses, but those classes will be removed if the content are from other Mastodon instances. This may confuses on styling with CSS.

This patch excludes `class` attribute on the `a` tag in sanitization.

Edit: I've found this also has (small) effect even on default CSS.

image

### DIFF
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -8,7 +8,7 @@ class Sanitize
       elements: %w(p br span a),
 
       attributes: {
-        'a'    => %w(href rel),
+        'a'    => %w(href rel class),
         'span' => %w(class),
       },
 


Gargron (approved):

clworld:
https://github.com/tootsuite/mastodon/issues/3790

unarist:
@clworld Oh... but similar appearance can be achieved with `class` attribute on `span` element, which available before this patch and needed for URL styling.

```
<p>lewd <a href="http://malware.ru/"><span class="status__content__spoiler-link">Show more</span></a></p>
```

So we should filter any class attributes.

PR #3624, opened 2017-06-07
Fix Code Climate failed (regression from #3622)

ykzts:
fix https://codeclimate.com/github/tootsuite/mastodon/pull/3622


clworld (approved):
OK.

PR #3635, opened 2017-06-08
Add Japanese translations

ykzts:
follow up #3261 and #3634

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -39,6 +39,7 @@ ja:
     people_who_follow: "%{name} さんをフォロー中のアカウント"
     posts: 投稿
     remote_follow: リモートフォロー
+    reserved_username: このユーザー名は予約されています。
     unfollow: フォロー解除
   activitypub:
     activity:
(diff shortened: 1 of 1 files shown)


config/locales/ja.yml
```diff
@@ -39,6 +39,7 @@ ja:
     people_who_follow: "%{name} さんをフォロー中のアカウント"
     posts: 投稿
     remote_follow: リモートフォロー
+    reserved_username: このユーザー名は予約されています。
```
clworld: It follows up #3566 also.

config/locales/ja.yml
```diff
@@ -226,6 +231,7 @@ ja:
       content: セキュリティ認証に失敗しました。Cookieをブロックしていませんか？
       title: セキュリティ認証に失敗
     '429': リクエストの制限に達しました。
+    noscript: Mastodonを利用する場合はJavaScriptを有効にしてください。
```
clworld: I wondered which is better '利用' or '使用', but '利用' is good enough.

clworld (approved):
Looks good to me.

config/locales/ja.yml
unarist: '利用' is good but it might sound like saying about Mastodon as a service, not only our web client. So it may be better with '使用' '使う', or rephrase it to something like:

- JavaScript を有効にしてください
- このページを表示するには JavaScript が必要です

but this is not a huge problem I think.

(Mentioning something like "Instead, you can use Mastodon from 3rd-party apps" may also be good, but this will be no longer translation issue)

ykzts: Certainly "use" has a slight incongruity. However, I do not want to change it significantly from the original phrase.

The difference between "使用" and "利用" is difficult. However, since it is provided as a web service, I thought that "利用" is more appropriate.

PR #3639, opened 2017-06-08
Support multiple trusted proxy ip addresses

nullkal:
It is often required to specify multiple trusted proxy IP addresses in the environment puma and nginx run in independent servers.

You can specify multiple trusted proxy IP addresses with this PR: `TRUSTED_PROXY_IP=10.2.2.4 10.2.2.5`

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -38,7 +38,7 @@ Rails.application.configure do
   config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Allow to specify public IP of reverse proxy if it's needed
-  config.action_dispatch.trusted_proxies = [IPAddr.new(ENV['TRUSTED_PROXY_IP'])] unless ENV['TRUSTED_PROXY_IP'].blank?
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -38,7 +38,7 @@
   config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Allow to specify public IP of reverse proxy if it's needed
-  config.action_dispatch.trusted_proxies = [IPAddr.new(ENV['TRUSTED_PROXY_IP'])] unless ENV['TRUSTED_PROXY_IP'].blank?
+  config.action_dispatch.trusted_proxies = ENV['TRUSTED_PROXY_IP'].split().map {|item| IPAddr.new(item) } unless ENV['TRUSTED_PROXY_IP'].blank?
```
Gargron: `split()` is the same as `split`

Spacing between `{ |`

Gargron (changes requested):

Gargron (approved):

PR #3640, opened 2017-06-08
Add explit admin actions to (re)subscribe/unsubscribe remote accounts

Gargron:
and re-download avatar/header

Also I've run `i18n-tasks normalize` and apparently it hasn't been run by anyone submitting locales for a while so that's a huge update -_-


Gargron:
Also: 
image

ykzts:
reports?

ykzts (approved):

Gargron:
Yeah!

PR #3642, opened 2017-06-08
Do not spawn a RemoteProfileUpdateWorker if we are going to synchrono…

ClearlyClaire:
…usly update the profile anyway

### DIFF
--- a/app/services/concerns/author_extractor.rb
+++ b/app/services/concerns/author_extractor.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module AuthorExtractor
-  def author_from_xml(xml)
+  def author_from_xml(xml, update_profile = true)
     return nil if xml.nil?
 
     # Try <email> for acct
@@ -18,6 +18,6 @@ module AuthorExtractor
       acct   = "#{username}@#{domain}"
     end
 
-    FollowRemoteAccountService.new.call(acct)
+    FollowRemoteAccountService.new.call(acct, update_profile)
   end
 end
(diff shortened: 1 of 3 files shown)


akihikodaki (approved):

app/services/concerns/author_extractor.rb
```diff
@@ -18,6 +18,6 @@ def author_from_xml(xml)
       acct   = "#{username}@#{domain}"
     end
 
-    FollowRemoteAccountService.new.call(acct)
+    FollowRemoteAccountService.new.call(acct, update_profile: update_profile)
```
Gargron: I might be wrong on the syntax rules because I've never used them like this but isn't that the hash syntax?

ThibG: I'm not sure what you are talking about here… it's just the syntax for named parameters?

Gargron: The callee syntax is indeed the syntax of named parameters, but *also* hash syntax, because named parameters are a shortcut for using one hash argument <https://robots.thoughtbot.com/ruby-2-keyword-arguments>

However, your method definitions seem to be wrong. `update_profile = true` is a default value, not named keyword argument. That would be `update_profile: true`

ThibG: Oh, indeed, good catch! I'll fix that.

ThibG: This has been fixed. Anything else preventing a merge?

Gargron (approved):

PR #3643, opened 2017-06-08
Swipe gestures

### LINKED ISSUE #2889: Swipeable views
Would it be feasible to integrate something like [react-swipeable-views](https://github.com/oliviertassinari/react-swipeable-views) into the Mastodon columns? As mobile devices seem to be getting bigger and taller, it is increasingly becoming difficult to get to any tabs at the top of the screen. I'd like to see something like this implemented or perhaps a possible discussion about moving the tab bar to the bottom of the screen.

* * * *

sorin-davidoi:
Swipe between columns and media. Closes #2889.

### DIFF
--- a/app/javascript/mastodon/features/ui/components/columns_area.js
+++ b/app/javascript/mastodon/features/ui/components/columns_area.js
@@ -2,12 +2,14 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import ImmutablePureComponent from 'react-immutable-pure-component';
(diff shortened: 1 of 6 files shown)


sorin-davidoi:
There is still an issue with the tab bar. If you tap on a link, it receives the `hover` state, but when you swipe away the `hover` state remains on the link that was tapped.

image

sorin-davidoi:
I think this is ready for review. Could not find a good fix for the issue above, so I've disabled the background color for the mobile UI (where swiping takes place).

Gargron (approved):

sorin-davidoi:
Fixed conflicts caused by upgrading to React Router 4.

PR #3644, opened 2017-06-08
Fix regression from #3592 - validation condition nesting

Gargron:


clworld (approved):
Applyed this on my staging server.
Sidekiq Jobs which is failed by 'Note is too long' are proceed successfully.

And it looks good to me.

mjankowski (approved):
I'm surprised this works this way! But - looks fine.

PR #3647, opened 2017-06-08
[WIP] Close modal by pressing navigating back

### LINKED ISSUE #3060: Android back button does not work on media overlay
STR:
* open mastodon UI on mobile
* switch to one of the timelines
* open the media attachment on a toot
* press Android device `<back>` button

Expected:
* media overlay closes and I am back on the timeline underneath

Actual:
* media overlay stays open
* timeline *underneath* goes back one step (to the previously visited column/page)

--> I suspect this can be fixed with a `pushState()` for the overlay state, so that the "back" behavior takes us to the timeline again.

Dismissing the overlay in any other way would then have to `popState()` too of course.

* * * *

sorin-davidoi:
Closes #3060.

### DIFF
--- a/app/javascript/mastodon/actions/modal.js
+++ b/app/javascript/mastodon/actions/modal.js
@@ -9,8 +9,9 @@ export function openModal(type, props) {
   };
 };
 
-export function closeModal() {
+export function closeModal({ skipGoingBack = false } = { }) {
   return {
     type: MODAL_CLOSE,
+    skipGoingBack,
   };
 };
--- a/app/javascript/mastodon/containers/mastodon.js
+++ b/app/javascript/mastodon/containers/mastodon.js
@@ -48,13 +48,13 @@ import { getLocale } from '../locales';
 const { localeData, messages } = getLocale();
 addLocaleData(localeData);
 
-const store = configureStore();
-const initialState = JSON.parse(document.getElementById('initial-state').textContent);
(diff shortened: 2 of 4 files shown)


sorin-davidoi:
While this work, I don't think it is a good solution (feels like a hack). The best way to go seems to be for the modals (and their content) to have their own routes.

penduin:
...Is the verdict that this needs to be done in a specific way?  Is this still in the works?  I would love to use the main web client on mobile, but my brain is incompatible with the modal behaviour; hitting the back button when an image is being shown keeps the image up while navigation happens underneath.  Super confusing.  And I can't seem to train myself to hunt for a subtle "X" near the top of the display.

I'm happy to dig in and try to figure it out myself, but I'm not likely to come up with the "correct" or "clean" solution.  :^)  I very much want back-to-close-modals though.

Gargron:
Hey, this doesn't look too bad to me. And it would be a very good addition, even I hit back on modals on mobile even though I should know better. Didn't see this until now

yuntan:
Hi folks, is there any progress on the topic? I think this feature is essential to improve mobile experience of mastodon web. cc @Gargron @sorin-davidoi

sorin-davidoi:
Not from my end.

Gargron:
I can't reopen this PR because the branch has been deleted.

PR #3652, opened 2017-06-09
Check ready state of document in public pages

ykzts:
`main()` will wait for polyfills loading. Therefore, there are times when it is executed after loading the HTML document.

Check at the value of [`document.readyState`](https://html.spec.whatwg.org/multipage/dom.html#current-document-readiness) to see if the loading is completed.

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -10,7 +10,7 @@ require.context('../images/', true);
 const { localeData } = getLocale();
 localeData.forEach(IntlRelativeFormat.__addLocaleData);
 
-function main() {
+function loaded() {
   const locale = document.documentElement.lang;
   const dateTimeFormat = new Intl.DateTimeFormat(locale, {
     year: 'numeric',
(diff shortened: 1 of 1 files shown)


app/javascript/packs/public.js
```diff
@@ -21,23 +21,29 @@ function main() {
...
-      content.textContent = relativeFormat.format(datetime);;
-    });
+  [].forEach.call(document.querySelectorAll('time.time-ago'), (content) => {
+    const datetime = new Date(content.getAttribute('datetime'));
+    content.textContent = relativeFormat.format(datetime);;
   });
+}
+
+function main() {
+  if (document.readyState === 'complete') {
```
yumetodo: Which is better `document.readyState === 'complete'` or `document.readyState !== '"loading'`?

ykzts: Oh, I'm sorry. I haven't seen this comment. I  just made similar changes. c538e82

app/javascript/packs/public.js
```diff
@@ -39,7 +39,7 @@ function loaded() {
 }
 
 function main() {
-  if (document.readyState === 'complete') {
+  if (['interactive', 'complete'].includes(document.readyState)) {
```
yumetodo: Why don't you write `document.readyState !== 'loading'` instead of calling `Array.prototype.includes()`?

ykzts: Because want to express state explicitly.

clworld (approved):
Looks good to me.

PR #3654, opened 2017-06-09
adjust the size of 'column-header__back-button'

fvh-P:
In Japanese environment, the back button on the community timeline and the public timeline is overlapping with column title ```ローカルタイムライン```.
ltl-large-button

I modified the padding size of this button smaller. (```right: 15px, left: 15px``` => ```right: 5px, left: 0px```)
ltl-smallbutton

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1527,7 +1527,7 @@
   cursor: pointer;
   flex: 0 0 auto;
   font-size: 16px;
-  padding: 15px;
+  padding: 0 5px 0 0;
   z-index: 3;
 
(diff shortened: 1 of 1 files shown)


clworld:
I found this happens not only in Japanese language environment. for example: Català.
I'll test this on other languages.

Gargron (approved):

clworld:
Hmm... I'm trying to test this...
It seems other style is changed.
profile
local

clworld:
This 'padding:' at line 1530 is '.column-back-button'.
'padding:' for '.column-header__back-button' is at line 1546.

PR #3657, opened 2017-06-09
revive-faster_assets

masarakki:
Are the dependencies enough?


config/deploy.rb
```diff
@@ -12,3 +12,6 @@
 
 append :linked_files, '.env.production', 'public/robots.txt'
 append :linked_dirs, 'vendor/bundle', 'node_modules', 'public/system'
+
+set :assets_prefix, 'packs'
```
Gargron: Will this work? Because `assets:precompile` executes both Sprockets and Webpack, there are still parts that use Sprockets, such as the Sidekiq and PGHero web views, so those things remain under "assets" I believe, while Webpack stuff goes into "packs".

Gargron:
Not sure if this works correctly with Webpacker and got no response, so closing.

PR #3669, opened 2017-06-09
Fix #3658 - Update RTL styles

### LINKED ISSUE #3658: 1.4.2: Panel title broken when using RTL language
style_broken_rtl
Left panel and center panel has same title. Title on left panel covered by other UI elements.
This happens in 1.4.2 and not happens in 1.4.1.

* * * *

Gargron:


clworld:
Yeah. It's fixed.
fixed

I found additional bug …
bad_move
(panel move left when clicking '>', panel move right when clicking '<')

Gargron:
@clworld I don't know how to fix that. It would require for the settings component to be aware of being in an RTL environment, and to switch order of the buttons based on that...

clworld (approved):
It's property fixes issue #3658. I checked.
Looks good to me for now.

PR #3677, opened 2017-06-09
Upgrade React Router

sorin-davidoi:
React router is the only dependency that is one major version behind (apart from `emojione`, regarding which a decision is still to be made). Upgrading it allows us to tackle lazy-loading without concerns regarding making upgrading more difficult down the road.

Changes:
 - [x] Replace `Router` with `BrowserRouter` (no need to create separate history)
 - [x] Move the actual routers to `UI` inside a `Switch` (since we cannot use `component` and `children` at the same time)
 - [x] The route components now get passed a `match` instead of `params`. I also could not find a nice way to specify how to pass the children to route component, so I created a small wrapper for `Router` which handles this (extracts `params` from `match` and passed the children).
 - [x] Remove usages fo `react-router-scroll`
 - [x] Calls like `this.context.router.push` become `this.context.router.history.push`
 - [x] As `react-router-scroll` is not compatible with `react-router-web`, we need to find another library / reimplement what functionality we lost by removing it

### DIFF
--- a/app/javascript/mastodon/components/column_back_button.js
+++ b/app/javascript/mastodon/components/column_back_button.js
@@ -9,8 +9,8 @@ class ColumnBackButton extends React.PureComponent {
   };
 
   handleClick = () => {
-    if (window.history && window.history.length === 1) this.context.router.push('/');
-    else this.context.router.goBack();
+    if (window.history && window.history.length === 1) this.context.router.history.push('/');
+    else this.context.router.history.goBack();
   }
 
   render () {
--- a/app/javascript/mastodon/components/column_back_button_slim.js
+++ b/app/javascript/mastodon/components/column_back_button_slim.js
@@ -9,8 +9,8 @@ class ColumnBackButtonSlim extends React.PureComponent {
   };
 
   handleClick = () => {
-    if (window.history && window.history.length === 1) this.context.router.push('/');
-    else this.context.router.goBack();
+    if (window.history && window.history.length === 1) this.context.router.history.push('/');
+    else this.context.router.history.goBack();
   }
 
   render () {
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -45,8 +45,8 @@ class ColumnHeader extends React.PureComponent {
   }
 
   handleBackClick = () => {
-    if (window.history && window.history.length === 1) this.context.router.push('/');
-    else this.context.router.goBack();
+    if (window.history && window.history.length === 1) this.context.router.history.push('/');
+    else this.context.router.history.goBack();
   }
 
   handleTransitionEnd = () => {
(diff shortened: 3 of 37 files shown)


sorin-davidoi:
To be honest, at a quick glance I am not sure what `react-router-scroll` was used for. The prop `shouldUpdateScroll` was always `undefined`, and `useScroll` don't think it served any purpose (since there is no scrolling on the `body`, only in columns). Might be wrong, would appreciate anyone taking a look at this.

Gargron:
@sorin-davidoi react-router-scroll saves the scrolling position inside ScrollContainer between route navigations, e.g. from public timeline -> status -> account, and if you click back twice you get the public timeline on the scrolling position where you were.

It's been touched by other contributors since the last time I personally put it in, so maybe it got broken, but the functionality is pretty important.

sorin-davidoi:
There seems to be no good alternative to `react-router-scroll` ([discussion about supporting `react-router-dom`](https://github.com/taion/react-router-scroll/issues/52)). I've installed the [fork](https://github.com/ytase/react-router-scroll) mentioned in the issue and it works fine:tm:, but it has not been merged yet.

The [documentation on scroll restoration](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md) hints to a solution similar to what the fork implements, but provides no implementation.

sorin-davidoi:
@nolanlawson, maybe you deploy this on malfunctioning.technology when you have some time? I think this required a bit of battle testing, and I don't have an instance to test this more extensively myself.

clworld:
I testing this branch on my staging environment and found 2 glitches.
1. When unpinning Home or Notifications, unpinned columns does not appear in getting-started panel.
(also Local & Federated timeline does not there?)
master(80c13bf0ef):
unpin_master
master(80c13bf0ef) + sorin-davidoi:upgrade-react-router
unpin_router
2.When opening Local or Federated timeline, these timeline does not have Back button and +pin button. 
master(80c13bf0ef):
new_panel_master
master(80c13bf0ef) + sorin-davidoi:upgrade-react-router
new_panel_router
Others are OK as I see.

sorin-davidoi:
@clworld Thanks for taking a look at this! Will try to fix them in the next few days.

sorin-davidoi:
@clworld Managed to fix them.

clworld (approved):
I tested this branch on my staging environment.
All glitches I reported are fixed.
It seems OK. Looks good to me.

nolanlawson:
@sorin-davidoi Can you fix the conflicts? Then I'd be happy to deploy to malfunctioning.technology. :)

sorin-davidoi:
@nolanlawson Sure! :smile:

sorin-davidoi:
@nolanlawson Just rebased against current master.

nolanlawson:
This is now live on malfunctioning.technology (with my SW PR as well (https://github.com/nolanlawson/mastodon/tree/sw), so you may need to reload).

sorin-davidoi:
Did a bit more testing on malfunctioning.technology and could not find any glitches.

PR #3679, opened 2017-06-09
Docker: smaller footprint, fix #3436

### LINKED ISSUE #3436: Proposal: Integrate assets into docker image
v1.4.1 introduces an issue with assets precompiling making it complicated to build webpack assets on certain instances. Unfortunately, currently the only way to workaround this issue is to build on another machine assets and copy the files to the packs directory.

Wouldn't it be easier to directly integrates packs into the docker image (instead of a volume) and build the webpack assets when building the docker image? 

It would permit building your assets on another machine and then publish your image through a custom repository (or directly through docker default image repo). It would also make installation a bit easier for new instance owners.

Wonderfall:
- Integrate assets into Docker image (fixes #3436)
- Remove build dependencies (virtual package)
- Remove npm and yarn cache
- Run database migrations task at startup (optional)

Some elements were taken from #3508 so thanks to @maur1th.
New image size is ~~818MB~~ 810MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,15 @@
 FROM ruby:2.4.1-alpine
 
-LABEL maintainer="https://github.com/tootsuite/mastodon" \
-      description="A GNU Social-compatible microblogging server"
-
 ENV UID=991 GID=991 \
+    RUN_DB_MIGRATIONS=false \
     RAILS_SERVE_STATIC_FILES=true \
-    RAILS_ENV=production NODE_ENV=production
-
-EXPOSE 3000 4000
+    RAILS_ENV=production \
+    NODE_ENV=production
 
 WORKDIR /mastodon
 
 RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
- && apk -U upgrade \
- && apk add -t build-dependencies \
-    build-base \
-    libxml2-dev \
-    libxslt-dev \
-    postgresql-dev \
-    protobuf-dev \
-    python \
- && apk add \
+ && apk -U add --no-cache \
     ca-certificates \
     ffmpeg \
     file \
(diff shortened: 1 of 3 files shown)


Wonderfall:
Maybe running the database migrations isn't such a great idea since we run multiple images...
Also it doesn't work yet with assets integrated, there's something wrong with URLs.

Wonderfall:
Now everything is fine except that the current nginx configuration file doesn't work. I had to remove the root directive and make 2 location blocks : `/` (proxy) and `/system` (local). Drawbacks : gzip/brotli doesn't work anymore so there is no compression... And I have no idea how to add headers for caching (with the location block suggested in the docs it gives 404 errors, so I had to remove this one). So any help would be very much appreciated on this issue with nginx.

Or I thought about an alternative : what if precompiled assets were moved to the volume when the container starts? Assets could be moved to a temporary location during the build so it can be possible.

Wonderfall:
Sorry for the monologue. The image is now fully working in production.

I'm not happy with [this change]( https://github.com/Wonderfall/mastodon-targaryen/blob/12945f876fb790f69a551792e915170381f73c45/Dockerfile#L43-L45) but this was the only way. Otherwise `uws` wouldn't build and it's the price to remove build dependencies from the image and use the Docker caching feature at the same time. It's not perfect, but this is a necessary redundancy.

Also, I made an environment variable (`RUN_DB_MIGRATIONS`) so the container can run `rake db:migrate` whenever it starts. It is disabled by default so we can enable this on one container. I'm not sure if you'll appreciate that but since we don't run `rake assets:precompile`, I thought it would be okay to find a way to make the update even easier.

[This change](https://github.com/Wonderfall/mastodon-targaryen/blob/03f6a97c73a498795ebf986781d2e888cea4da93/docker_entrypoint.sh#L29-L33) is tricky but it allows nginx to serve directly and it requires no modifications to the docker-compose. So the update to this Dockerfile requires no actions.

As a conclusion, this PR seems to be a good compromise between automation and caching, while drastically reducing the image size. I would completely understand if these changes are not welcome in the official repository, and feel free to tell me if something's wrong.

PR #3682, opened 2017-06-10
Improve RTL detection

Gargron:
- Use plaintext
- Strip out URLs
- Strip out mentions
- Strip out hashtags
- Strip out whitespace from "overall" count
- Consistent between JS and Ruby

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -47,11 +47,16 @@ module StreamEntriesHelper
     end
   end
 
+  def rtl_status?(status)
+    status.local? ? rtl?(status.text) : rtl?(strip_tags(status.text))
+  end
+
   def rtl?(text)
+    text = simplified_text(text)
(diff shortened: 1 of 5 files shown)


Gargron:
I have to mention that there were previous PRs #3047 and #2884 which didn't quite use the right things but had the right ideas.

beatrix-bitrot (approved):
LGTM

mabkenar:
Wow, it feels great to see this! Does this solve any of the two RTL issues #2350 and #2588? I can't test this now as glitch seems to be a bit behind.

mabkenar:
Glitch is up-to-date again, and yes! This commit does solve #2350. Thank you @Gargron @beatrix-bitrot and @seefood for all your great work.

Now about #2588 with [this user](https://glitch.social/@test887) page:

### This is what you see (good)


### This is what I see (bad)

It's because I have set my language to Persian in Mastodon preferences. For more details, see #2588. It contains a "what should happen" section where I explained the desired situation.

PR #3684, opened 2017-06-10
make status' timestamp absolutely

cobodo:
Add setting parameter `absolute_time` to show status' timestamp absolutely. By default, timestamps are shown relatively.


clworld:
It works in my environment, and I think some users like this.
I don't know/decide that this is good for Mastodon.

Things I wonder are:
- Is this date format OK for all languages?
- Why seconds added? (I'd like to see seconds in my timeline but currently Mastodon does not for toots older than 1 minutes.)
- Also app/javascript/mastodon/features/ui/components/boost_modal.js uses RelativeTimestamp.
- Changes for config/database.yml isn't needed.
- Absolute time is already shown in title-text. => I found it's difficult to make title-text shown on smartphone (no mouse-over).
- Absolute time shown in text and title-text. It's duplicated.

cobodo:
@clworld Thank you for reviewing my pull-req.

    Is this date format OK for all languages?
    Why seconds added? (I'd like to see seconds in my timeline but currently Mastodon does not for toots older than 1 minutes.)

I guess almost of Japanese are OK, but I'm not sure that people in many other countries are OK?

Seconds added by reason of "I'd like to see".
I think current Mastodon's behavior is "relative and only 1 phrase representation", so seconds are eliminated for toots older than 1 minutes. This absolute representation is short enough within same day, there's no reason to eliminate seconds.

    Also app/javascript/mastodon/features/ui/components/boost_modal.js uses RelativeTimestamp.

I didn't notice it. I'll also try to apply for it later.

    Changes for config/database.yml isn't needed.

It's my mistake. I'm sorry. Added revert commit.

    Absolute time is already shown in title-text. => I found it's difficult to make title-text shown on smartphone (no mouse-over).
    Absolute time shown in text and title-text. It's duplicated.

I like to overlook toots with absolute time, but mouse-over can action only one element. Of course, it's difficult to show on smartphones as you said.
I didn't mind text and title-text duplication when I first pull-req'ed. However, people using some text-reader software possibly feel strange.
Possible ways are:

1. simply eliminate title-text
2. fill title-text by relative representation

If 2, setting item label requires rethinking.

Gargron:
Absolute time is displayed in tooltip, I don't think we need this change, but thank you for taking the time to contribute.

Issue #3447, opened 2017-05-30
Autoload more toot of someone's toot list sometime fails on 1.4.1. (not on 1.3.3)


clworld:
On 1.4.1 (or master):
1. Open Mastodon and move to someone's toot list (for example: https://mastodon.social/web/accounts/1 )
2. Reload page.
3. Move to getting-started
4. Move to page opened at 1.
5. Scroll down toot list.
6. Only toots already loaded is shown and does not load more toots. (bug)

On 1.3.3, It load more toots properly.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

unarist:
Ah, that's my mistake. I have changed to use `next` field even on the first time request at https://github.com/tootsuite/mastodon/commit/cf4fe6cab86384971d4b7bd634f48b0a462efd48#diff-2e3225be2a2cdeb60c18038df0562779L154, but the function is also used for "Load newer statuses" operation. The `next` field will be empty if no newer statuses exist, and this prevents from loading more statuses.

I will try to fix this.

PR #3686, opened 2017-06-10
Fix #3665 - Refactor timelines reducer

### LINKED ISSUE #3665: multi-column support bug: clicking a hashtag opens in all columns
To reproduce:

Search multiple hashtags and pin them as columns.

Click on a hashtag. It opens in all of the pinned columns (except for Home and Notifications).

I'm not sure what the expected behavior should be. I think I'd like a clicked hashtag to open in a new column, but it shouldn't overwrite my existing pinned-hashtag columns.

Gargron:
Fix #3665

*Before:*

```
timelines
  home
  public
  community
  tag
  accounts_timelines
    3
    6
  accounts_media_timelines
    1
    3
    4
```

*After:*

```
timelines
  home
  public
  community
  hashtag:test
  hashtag:foobar
  account:3
  account:6
  account:1:media
  account:3:media
  account:4:media
```

___

Previously they all had different action creators and action signatures, e.g. `FETCH_ACCOUNT_TIMELINE_SUCCESS` specifically, etc. They also had special treatment in the reducer, with slightly different functions to handle each action. Also, there could be only one tag timeline, loading a new hashtag replaced the previous one.

Now everything is using the same action creators (with convenience curries) and the same action signatures e.g. `TIMELINE_REFRESH_SUCCESS`, and different hashtags can co-exist in the same world.


ykzts (approved):
The code looks good.

ykzts:
@beatrix-bitrot I will ask you to check with the production:smile:

unarist:
I found some warnings and errors on Chrome console:

```
./app/javascript/mastodon/features/report/index.js
88:24-46 "export 'refreshAccountTimeline' was not found in '../../actions/accounts'
./app/javascript/mastodon/features/report/index.js
93:26-48 "export 'refreshAccountTimeline' was not found in '../../actions/accounts'
```

Also this patch brings back account timeline issue #3447. Second fetching (with since_id) may not return next link, then it prevents load more.

Gargron:
@unarist `expandTimeline` uses `max_id: ids.last()`, ignoring the `next` path. This is because timelines can be cropped at around 40 items, and if you were using the `next` header, scrolling down again would bring you further and further into the past, skipping past the cropped items. Plus, as far as timelines are concerned `max_id` uses the primary ID

unarist:
@Gargron `next` path isn't used actually, but existence of it is important, because `handleScrollToBottom()` checks it as `hasMore`.

```
  handleScrollToBottom = () => {
    if (!this.props.isLoading && this.props.hasMore) {
      this.props.dispatch(expandAccountTimeline(Number(this.props.params.accountId)));
    }
  }
```

Gargron:
@unarist You are right. Perhaps `hasMore` should always be true. That's how it used to be at the very start...

clworld:
@unarist @Gargron 
I deployed this on my production. (my staging environment does not have enough number of toots…)
This PR reproduces issue https://github.com/tootsuite/mastodon/issues/3447.

clworld:
I caught some errors in Console.
timeline

unarist:
actions/timelines.js:

```js
export function updateTimeline(timeline, status) {
    const references = status.reblog ? getState().get('statuses').filter((item, itemId) => (itemId === status.reblog.id || item.get('reblog') === status.reblog.id)).map((_, itemId) => itemId) : [];
```

`getState().get('statuses')...` has `includes()` method, but `[]` has `include()` method...slightly different name.

And in `updateTimeline()` in reducers/timelines.js...
```
if (status.getIn(['reblog', 'id'], null) !== null) newIds = newIds.filterNot(item => references.include(item));
```

It only works for an array, not for Immutable.Map.

Gargron:
@unarist @clworld I have fixed the includes typo.

I am surprised if #3447 re-appears again, because expandTimeline does not use next? Let me double check

Gargron:
Oh wait! The problem is *only* because hasMore *blocks* the loading of more stuff, right?

clworld:
@Gargron 
I deployed fix and I checked r.include error does not happen.
unarist seems already go to sleep 1.5 hour ago.. (5:55 a.m. now in Japan)

unarist:
@Gargron

> hasMore blocks the loading of more stuff,

Right. `account/:id/statuses` with `since_id` may returns response without `next` link header, then it overwrites `timelines.['accounts:xxx'].next` with null. This means making `hasMore` to be false, and blocks the loading. We shouldn't overwrite that property on the loading *newer* statuses.

Well, maybe `account/:id/statuses` with `since_id` option should returns next link header, since we don't know there are more records below since_id.

Gargron:
@unarist Ah, I was thinking of a slightly different scenario. Same bug currently occurs in notification column too:

1. Scroll down to load more posts, or wait until >40 posts are loaded
2. If you are scrolled to top and a new post arrives, all posts after the 40th will be cut off
3. Scroll down again to the bottom to load more posts.

Expected: It will load the posts after the last visible one, without gaps  
Reality: It will load posts after the last invisible one, because that's where the "next" pointed already

So I think that the usage of "next" in the timelines specifically should be removed, and hasMore should always be true.

unarist:
> hasMore should always be true.

We shouldn't make hasMore *always* true, because it allows the user loading more statuses infinitely (#2066).

Also timelines may have "end" especially on remote account timeline.  So I've changed `accounts/:id:/statuses` API to return `next` link only if needed (#3311), to reduce redundant requests. but If you think handling this is too complicated, how about to set next to true on TIMELINE_REFRESH and keep current behavior on TIMELINE_EXPAND? This allows loading more at least once even on too short timeline, but it's not a huge problem.

Well, maybe we should store boolean value for `next` state instead of link header value itself to prevent misuse...

Gargron:
@unarist Okay, I think my last commit should take care of the issue. I think I had just accidentally removed the fix that was already there, i.e. only set "next" if previous "next" wasn't set in refreshTimeline (that's the since_id case)

I also found that there was a typo in expandNotifications, which had been there for a long time, and fixed that too.

app/javascript/mastodon/actions/notifications.js
```diff
@@ -124,10 +124,9 @@ export function refreshNotificationsFail(error, skipLoading) {
 
 export function expandNotifications() {
   return (dispatch, getState) => {
-    const url    = getState().getIn(['notifications', 'next'], null);
     const lastId = getState().getIn(['notifications', 'items']).last();
```
unarist: Oh, it's whole entity, not id...

unarist (changes requested):

clworld:
@unarist Yes. notifications stop doing load more...

clworld:
I found glitch reported by beatrix on Discord exists.
It happens when:
1. open Mastodon with home timeline  (scroll position is top of timeline)
2. new toot arrived.
3. Title indicates there is unread toots but scroll position is top of timeline already.
4. After scrolling (scroll down and move to top), unread indicator works good.

Gargron:
@clworld @beatrix-bitrot @unarist Okay, fixed!

unarist (approved):

PR #3688, opened 2017-06-10
Make sure email is case insensitive on all places

### LINKED ISSUE #3325: Login email field is case sensitive
This issue made me think there was a more serious bug with password changing but it turns out it's just a UX oversight.

When logging in, email field is case sensitive, resulting in a failed login even if the credentials match. I always type my email with the first character uppercase since it's my name and that's just muscle memory. It seems like the registration process converted my registered email to lowercase (perfectly acceptable, pretty standard) but the login process did not do this and it's doing a case sensitive match so after a few password change attempts thinking the database was not updating my password I realised this letter case was causing the issue.

I'd classify this as a bug because pretty much every other service I've ever used handles emails/usernames without caring about case and it caused quite a bit of confusion while trying to log in! I can imagine this should be a super simple fix.

* * * *

reneklacan:
When case insensitivity is enabled via devise's `config.case_insensitive_keys` then `.find_for_authentication` method needs to be used instead of `.find_by` because second mentioned returns `nil` when valid email with different cases is passed.

More info https://github.com/plataformatec/devise/wiki/How-To:-Use-case-insensitive-emails

Closes #3325

### DIFF
--- a/app/controllers/auth/sessions_controller.rb
+++ b/app/controllers/auth/sessions_controller.rb
@@ -27,7 +27,7 @@ class Auth::SessionsController < Devise::SessionsController
     if session[:otp_user_id]
       User.find(session[:otp_user_id])
     elsif user_params[:email]
-      User.find_by(email: user_params[:email])
+      User.find_for_authentication(email: user_params[:email])
     end
   end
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

saper:
Well, but email address *is* case sensitive.

reneklacan:
@saper yes, by standard recipient part is case sensitive but it is very common practice to make it case insensitive to avoid confusion and enhance security (all major email providers have case insensitive emails - eg gmail, yahoo, etc)

PR #3693, opened 2017-06-11
Add "Search mode" on the regex filter (fix #3653)

fvh-P:
The regex filter is now working only as a ruling out filter.
This enhancement makes the regex filter also as a searching filter.
26958655-d2af6f4a-4d06-11e7-97e1-97a0a989deec
The "Search mode" switch is applied, only matched toots will be shown.

I think the sentence ```Filter out by regular expressions``` is no longer appropriate due to this change, so I made a change.


yaymukund:
I like this!

I didn't realize it was inverting the search until I read this thread.

Maybe the "Search" option could instead be called "invert"? That's how I've heard people refer to it, but this might be programmer bias

fvh-P:
Yes, yes. This enhancement is just inverting the result of regex filter.
I named this mode "Search" because I thought this mode is available for provisional (maybe partial) search function. However, I also feel this name is a little complicating.
Give me your opinions.

Cassolotl:
Speaking as a non-technical user, "invert" would be *extremely* confusing! "Search" is a familiar and sense-making name, so I'd be happy with that. :)

Does this change mean you can't block *and* search at the same time?

yaymukund:
I agree "invert" sounds complicated. My confusion was with two things:

1. The default behavior is to invert the search
2. "Filter" versus "Search mode" sounds ambiguous / similar.

How about something like:

mastodon-ui-suggestion

(whipped this up in a jiffy, so please ignore ui quirks heh)

Anyway, i'm excited to have this functionality as I wanna search for things all the time 👍

Cassolotl:
Eek, this continues to be very confusing! :S I'm sorry, I don't know what to suggest, though I will have a think about it.

The best I can think of is, when the toggle is off, the text says "exclude toots with these words" - and when the toggle is on, the text says "show only toots with these words".

PR #3695, opened 2017-06-11

akihikodaki:

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -25,7 +25,7 @@ module Mastodon
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
     # config.time_zone = 'Central Time (US & Canada)'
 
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -60,8 +60,6 @@ class Application < Rails::Application
       :'zh-TW',
     ]
 
-    config.i18n.default_locale = :en
```
Gargron: No, I'm afraid you cannot do this. This option "default_locale" is not the same as pre-choosing a locale to be displayed. It controls the fallback direction for missing translations. When a translation is missing, it gets the string from the english one instead. This makes sense because the software is developed in english so that locale always contains all the keys. But if another is chosen here, things will break and error out because keys will be genuinely missing without a fallback.

akihikodaki:
I submitted a wrong patch set. I'll close this because I'm out and cannot submit the correct one.

PR #3696, opened 2017-06-11

akihikodaki:

This is succeeding #3695.


akihikodaki:
I have seen https://github.com/tootsuite/mastodon/pull/3695#discussion_r121277335. That applies also for this change set.
I'll close this pull request. Sorry for bothering.

PR #3697, opened 2017-06-11
Made UI more sexy/Merge with mate.

ghost:
I often complained, that my userstyle is a little better, than the standard Mastodon (S)CSS. I know, that my code looks like a little mess, but this is my first pull request. Just let me know in the comments, what I should change, so we can merge this thing.


Gargron:
Screenshot please

koyuawsmbrtn:
@Gargron I have nothing to test/deploy on

koyuawsmbrtn:
screenshot 2017-06-11 at 13 20 09

It should look something like this, when everything is right

ggtea:
I don't like Circlized Icon.. (I know Google+ use that)

koyuawsmbrtn:
@ggtea Ok, I'll try to edit it

app/javascript/styles/components.scss
```diff
@@ -90,6 +90,21 @@
   }
 }
 
+.column-collapsable__button,
+.column-icon-clear {
+  background-color: $classic-base-color !important;
+  border-top-right-radius: 5px;
+  -moz-border-top-right-radius: 5px;
```
clworld: It seems Mastodon does not use vendor prefixes for border-radius.
So, we can remove -moz-, -webkit- from border radius.
And Mastodon use border-radius rather than border-top-right-radus and so on. See: https://github.com/tootsuite/mastodon/blob/master/app/javascript/styles/accounts.scss#L255

app/javascript/styles/variables.scss
```diff
@@ -26,3 +26,9 @@ $ui-base-color: $classic-base-color !default;             // Darkest
 $ui-primary-color: $classic-primary-color !default;       // Lighter
 $ui-secondary-color: $classic-secondary-color !default;   // Lightest
 $ui-highlight-color: $classic-highlight-color !default;   // Vibrant
+
+//Values for the new Mastodon UI
+$user-simple-form: #1d2026;
+$column-button-hover: #3b434e;
+$column-link: #404657;
+$tab-delete: #2e3340;
```
clworld: I don't know adding variables are right.

clworld (commented):
I don't know changing button/header color is good or not.

app/javascript/styles/components.scss
unarist: Hint: We are using autoprefixer.

clworld:
I wonder that many "!important" in there scss are relly needed?

Gargron:
I'm gonna say sooner rather than later that we *rarely* accept visual changes to existing UI. I do not really see a benefit in making the column headers the same background color and making the avatars round. One design is not strictly better than the other. So this will probably not be merged, sorry.

m4sk1n:
Maybe instead of making changes directly to Mastodon interface, there should be good per-instance settings manager that would allow for example to use circled icons?

ggtea:
I think, rounded-corner UI is better than squared-corner UI.
Also Twitter uses rounded-corner UI.
no-round
↓
round

koyuawsmbrtn:
@m4sk1n even better: i would like to see a custom css feature. i don't mind if this pull request will be closed to be honest.

koyuawsmbrtn:
it won't build for some reason. i'll close this now.

PR #3699, opened 2017-06-11
Fix #3582 - Update OStatus2 gem

### LINKED ISSUE #3582: error 500 on follow salmons from hubzilla
Having a bit of difficulty finding the issue here. I'm sending a follow salmon, sent as application/magic-env+xml and Mastodon seems to consistently return error 500. I've been told the error in the Mastodon logs was nil content return from the salmon controller (and therefore un-parseable) but those notes with the precise message aren't available to me at the moment. 

It is difficult to include XML code in markdown without mangling so the example salmon as sent can be found at https://macgirvin.com/cloud/mike/slap.txt

The same construct works correctly in GNU-Social and other statusnet implementations but has consistently failed on Mastodon. I'm specifically testing against techn.ical.ist and not certain what version is in use there.

* * * *

Gargron:


ykzts (approved):
I checked https://github.com/tootsuite/ostatus2/commits/master .

PR #3700, opened 2017-06-11
Use debounce for dispatch scrollTopNotification and expandNotifications

unarist:
|Before|After|
|-|-|
|gif|gif2|

### DIFF
--- a/app/javascript/mastodon/features/notifications/index.js
+++ b/app/javascript/mastodon/features/notifications/index.js
@@ -13,6 +13,7 @@ import ColumnSettingsContainer from './containers/column_settings_container';
 import { createSelector } from 'reselect';
 import Immutable from 'immutable';
 import LoadMore from '../../components/load_more';
+import { debounce } from 'lodash';
 
 const messages = defineMessages({
   title: { id: 'column.notifications', defaultMessage: 'Notifications' },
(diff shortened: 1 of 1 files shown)


Gargron:
Perhaps this is a case for debounce? I don't like storing state outside of React (it is against the React philosophy, only acceptable for corner cases)

unarist:
Ah, I found similar fixes using debounce in our code. I will rewrite with that.

Gargron (approved):

app/javascript/mastodon/features/notifications/index.js
```diff
@@ -72,7 +77,7 @@ class Notifications extends React.PureComponent {
 
   handleLoadMore = (e) => {
     e.preventDefault();
-    this.props.dispatch(expandNotifications());
+    dispatchExpandNotifications();
```
Gargron: Should be `this.dispatchExpandNotifications()`

Gargron (changes requested):

app/javascript/mastodon/features/notifications/index.js
unarist: Oops, I've rebased and fixed it.

Gargron (approved):

PR #3672, opened 2017-06-09
Fix removal of status sending the original status to mentioned users instead of delete Salmon

Gargron:

PR #3704, opened 2017-06-12
Fix regression from #3672 - Do not use pipeline around zscore

Gargron:


unarist (approved):
I've confirmed this processes jobs which stopped by error.

PR #3714, opened 2017-06-12
Pipeline unpush in RemoveStatusService

akihikodaki:
This pull request reintroduces a feature removed by #3704.

### DIFF
--- a/app/services/remove_status_service.rb
+++ b/app/services/remove_status_service.rb
@@ -29,14 +29,11 @@ class RemoveStatusService < BaseService
 
   private
 
-  def remove_from_self
-    unpush(:home, @account, @status)
-  end
+  def remove_from_feeds
+    receiver_ids = @account.followers.local.pluck(:id)
+    receiver_ids << @account.id if @account.local?
 
-  def remove_from_followers
-    @account.followers.local.find_each do |follower|
-      unpush(:home, follower, @status)
-    end
+    unpush_bulk(:home, receiver_ids, @status)
   end
 
   def remove_from_affected
@@ -101,14 +98,30 @@ class RemoveStatusService < BaseService
     end
   end
 
-  def unpush(type, receiver, status)
-    if status.reblog? && !redis.zscore(FeedManager.instance.key(type, receiver.id), status.reblog_of_id).nil?
-      redis.zadd(FeedManager.instance.key(type, receiver.id), status.reblog_of_id, status.reblog_of_id)
-    else
-      redis.zremrangebyscore(FeedManager.instance.key(type, receiver.id), status.id, status.id)
-    end
+  def unpush_bulk(type, receiver_ids, status)
+    return if receiver_ids.empty?
 
-    Redis.current.publish("timeline:#{receiver.id}", @payload)
+    if status.reblog?
+      scores = redis.pipelined do
+        receiver_ids.each do |receiver_id|
+          redis.zscore(FeedManager.instance.key(type, receiver_id), status.reblog_of_id)
+        end
+      end
+
+      redis.pipelined do
+        receiver_ids.each_with_index do |receiver_id, index|
+          redis.zadd(FeedManager.instance.key(type, receiver_id), status.reblog_of_id, status.reblog_of_id) if scores[index].present?
+          redis.publish("timeline:#{receiver_id}", @payload)
+        end
+      end
+    else
+      redis.pipelined do
+        receiver_ids.each do |receiver_id|
+          redis.zremrangebyscore(FeedManager.instance.key(type, receiver_id), status.id, status.id)
+          redis.publish("timeline:#{receiver_id}", @payload)
+        end
+      end
+    end
   end
 
   def remove_from_hashtags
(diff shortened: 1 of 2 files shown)


app/services/remove_status_service.rb
```diff
@@ -57,14 +53,37 @@ def remove_reblogs
     end
   end
 
-  def unpush(type, receiver, status)
-    if status.reblog? && !redis.zscore(FeedManager.instance.key(type, receiver.id), status.reblog_of_id).nil?
-      redis.zadd(FeedManager.instance.key(type, receiver.id), status.reblog_of_id, status.reblog_of_id)
+  def unpush_bulk(type, receiver_ids, status)
```
Gargron: This method needs a bit of changing, I think. It's hard to tell what is happening, and instead of one loop, it loops 3 times over the same array in the worst case, and 2 times in the best case.

It could help to split the method into two, one for when the status is a reblog, and one for when it's not. When it's not a reblog, the method is really simple: just loop over receiver_ids in a pipeline, zremrangebyscore and publish at the same time. When it's a reblog, that's the more complicated case, and I accept its solution, except I would prefer not to use lambdas in this new method.

spec/rails_helper.rb
```diff
@@ -13,6 +13,7 @@
 
 ActiveRecord::Migration.maintain_test_schema!
 WebMock.disable_net_connect!
+Redis.current = Redis::Namespace.new("mastodon_test#{ENV['TEST_ENV_NUMBER']}", redis: Redis.current)
```
Gargron: This is 👍

spec/rails_helper.rb
```diff
@@ -32,6 +33,11 @@
     https = ENV['LOCAL_HTTPS'] == 'true'
     Capybara.app_host = "http#{https ? 's' : ''}://#{ENV.fetch('LOCAL_DOMAIN')}"
   end
+
+  config.after :each do
+    keys = Redis.current.keys
+    Redis.current.del(keys) if keys.any?
```
Gargron: This is 👍 💯 😂

Gargron (changes requested):

app/services/remove_status_service.rb
akihikodaki: I merged the lambdas and the last `redis.piplined` block as you suggested.
However, I'm still not convinced the idea to divide the method to two. They would be `unpush_reblog_bulk` and `unpush_original_bulk` or something sort of. It is a clearly mistake to call `unpush_reblog_bulk` without checking `status.reblog?` and therefore `if status.reblog?` and the content of those two methods should be in a method.

app/services/remove_status_service.rb
```diff
@@ -57,14 +53,34 @@ def remove_reblogs
...
+      end
+
+      redis.pipelined do
+        scores.each_with_index do |score, index|
+          next if score.nil?
+          redis.zadd(FeedManager.instance.key(type, receiver_ids[index]), status.reblog_of_id, status.reblog_of_id)
+        end
+
+        receiver_ids.each do |receiver_id|
+          redis.publish("timeline:#{receiver_id}", @payload)
```
unarist: How about merging `scores.each_with_index do |score, index|` and `receiver_ids.each do |receiver_id|`, like this?

```rb
scores.each_with_index do |score, index|
  redis.zadd(...) unless score.nil?
  redis.publish(...)
end
```

akihikodaki: That is not possible because the responses are not available in the pipeline. That is also the cause of the regression fixed by #3704.
See https://redis.io/topics/pipelining. It also suggests to write a Lua embedded script, but I think that is too much.

unarist: I said about L67-L70 and L72-L74 which are already in same pipeline.

akihikodaki: Oh, that is my fault. I was careless.
It is conforming to the behavior of the current implementation.
I think the behavior is to delete statuses which are not in the feed in Redis, but on the client. The number of statuses on the feed in Redis is limited to 40 and old statuses will be discarded, but they won't be discarded on the client.

app/services/remove_status_service.rb
Gargron: I was thinking like:

```ruby
def unpush_bulk(type, receiver_ids, status)
  if status.reblog?
    unpush_bulk_reblog(...)
  else
    unpush_bulk_status(...)
  end
end
```

akihikodaki: In the case `unpush_bulk_reblog` must always between `if status.reblog?` and `else` and `unpush_bulk_status` must always between `else` and `end`. I do not think such division makes sense.

app/services/remove_status_service.rb
unarist: `scores` also includes responses for discarded items because `Redis.pipelined` returns array of response of each calls. So the counts of items should be same between `scores` and `receiver_ids`.

akihikodaki: Now I finally got what you mean. Sorry for bothering. I'll make a change accordingly.

akihikodaki:
73fe51d5a0d2dfabb1a37ea8f2e1729a84d33c86 applies https://github.com/tootsuite/mastodon/pull/3714#pullrequestreview-43442667

Gargron:
I have refactored RemoveStatusService since this was submitted. Do you think this PR is still necessary?

akihikodaki:
I do not see your refactoring. The latest commit for `app/services/remove_status_service.rb` is commit 605e2a417c0757046671265e35a431a5ae97c0c4.

Issue #3711, opened 2017-06-12
The unread indicator appears on weird position


unarist:
![image](https://user-images.githubusercontent.com/705555/27028144-dbaf7e7e-4f9e-11e7-97a1-f0d4b1e04426.png)

It should be appears from the header by half, and it shouldn't move according to scroll.

![image](https://user-images.githubusercontent.com/705555/27028486-ddee1212-4f9f-11e7-9361-bce6507b1a62.png)

Reproduce steps:

1. Open the local timeline, and **scroll down**
2. Back to the Getting started
3. Wait to arrive new toots on the local timeline (or toot yourself)
4. Open the local timeline again by clicking "Local timeline" on Getting started

I think it's caused by having `.status-list__unread-indicator` in inside of `.scrollable`. That element was in outside of `.scrollable` previously.

Probably it solved by having it in each containers which uses StatusList instead of inside of StatusList, but it doesn't seems clear solution... Also I'm thinking that the indicator may not be needed.

EDIT: This seems to be happen only on Chrome.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [master] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

unarist:
Solved by #3720

PR #3720, opened 2017-06-12
Unread indicator was invisible behind column header, adjusted

Gargron:
image

### DIFF
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -57,6 +57,10 @@ class ColumnHeader extends React.PureComponent {
     const { title, icon, active, children, pinned, onPin, multiColumn, showBackButton } = this.props;
     const { collapsed, animating } = this.state;
 
+    const wrapperClassName = classNames('column-header__wrapper', {
+      'active': active,
+    });
+
     const buttonClassName = classNames('column-header', {
       'active': active,
     });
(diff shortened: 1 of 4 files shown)


unarist:
It may be invisible, or may be visible on weird position (#3711).

And the indicator seems to move according to scroll. I've tested on Chrome 59 on Windows.

Gargron:
@unarist Thank you for reporting. It behaves correctly on Firefox. I don't currently have Chrome to experiment with this. If you inspect the element, do you have any idea why it would behave that way instead of sticking to the .column which is position: relative? It should be position: absolute relative to .column. But what you are saying sounds like it's relative to some child element of .column like .scrollable.

unarist:
@Gargron I've also confirmed that it works on Firefox...

> sounds like it's relative to some child element of .column like .scrollable.

I think so too.

Sadly, I haven't found CSS solution yet and I thought the only way to fix this is move the indicator to outside of the scrollable container.

Gargron:
That sounds reasonable, I will try to do that

Gargron:
@unarist I have tried

unarist (approved):
Great!

unarist:
Oh wait, could you add `flex: 0 0 auto` to `.column-header__wrapper` and remove from `.column-collapsable`? (see #3313)

unarist (approved):

PR #3721, opened 2017-06-12
Fix #2619 - When redis feed is empty, fall back to database

### LINKED ISSUE #2619: Empty home timeline after period of inactivity

Gargron:
Home feeds are normally stored in redis. But removed for users inactive for > 14 days.

When user logs back in, a background worker starts to regenerate the feed. However, before that job completes, the user is greeted with a blank home timeline.

This change will fallback to database queries if RegenerationWorker has been queued up but hasn't finished running.

### DIFF
--- a/app/controllers/concerns/user_tracking_concern.rb
+++ b/app/controllers/concerns/user_tracking_concern.rb
@@ -17,7 +17,7 @@ module UserTrackingConcern
     current_user.update_tracked_fields!(request)
 
     # Regenerate feed if needed
-    RegenerationWorker.perform_async(current_user.account_id) if user_needs_feed_update?
+    regenerate_feed! if user_needs_feed_update?
   end
 
   def user_needs_sign_in_update?
@@ -26,5 +26,10 @@ module UserTrackingConcern
 
   def user_needs_feed_update?
     current_user.last_sign_in_at < REGENERATE_FEED_DAYS.days.ago
+  end
+
+  def regenerate_feed!
+    Redis.current.setnx("account:#{current_user.account_id}:regeneration", 3_600 * 24, true)
+    RegenerationWorker.perform_async(current_user.account_id)
   end
 end
--- a/app/controllers/concerns/user_tracking_concern.rb
+++ b/app/controllers/concerns/user_tracking_concern.rb
@@ -29,7 +29,7 @@ module UserTrackingConcern
   end
 
   def regenerate_feed!
-    Redis.current.setnx("account:#{current_user.account_id}:regeneration", 3_600 * 24, true)
+    Redis.current.setnx("account:#{current_user.account_id}:regeneration", true) == 1 && Redis.current.expire("account:#{current_user.account_id}:regeneration", 3_600 * 24)
     RegenerationWorker.perform_async(current_user.account_id)
   end
 end
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -7,7 +7,7 @@ class Feed
   end
 
   def get(limit, max_id = nil, since_id = nil)
-    if redis.zcard(key).zero?
+    if redis.exists("account:#{@account.id}:regeneration")
       from_database(limit, max_id, since_id)
     else
       from_redis(limit, max_id, since_id)
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -7,15 +7,28 @@ class Feed
   end
 
   def get(limit, max_id = nil, since_id = nil)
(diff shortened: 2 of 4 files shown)


app/models/feed.rb
```diff
@@ -7,15 +7,28 @@ def initialize(type, account)
...
     since_id   = '-inf' if since_id.blank?
     unhydrated = redis.zrevrangebyscore(key, "(#{max_id}", "(#{since_id}", limit: [0, limit], with_scores: true).map(&:last).map(&:to_i)
     status_map = Status.where(id: unhydrated).cache_ids.map { |s| [s.id, s] }.to_h
-
     unhydrated.map { |id| status_map[id] }.compact
   end
 
-  private
+  def from_database(limit, max_id, since_id)
+    results = Status.as_home_timeline(@account).paginate_by_max_id(limit, max_id, since_id).to_a
```
akihikodaki: You can use `reject` instead of `to_a` and `reject!`.

akihikodaki (changes requested):
I request a small minor change; please see the review.
By the way, I do not think the code to regenerate feeds are spread out across modules (`UserTrackingConcern`, `Feed`, and `PrecomputeFeedService`), but that can be resolved later.

clworld:
Isn't it good also fallback to database when user reached end of feed(400)?
In public/local timeline user can read history infinitely, but not in home timeline. User can read only 400 recent toots.

app/models/feed.rb
clworld: It seems Web UI will stop paging when reject! below filters all of statuses in page.
But it's OK because it's used only when Feed is not created.

Gargron: @akihikodaki I wanted to do it this way to avoid copying the array (reject! mutates existing array afaik)

@clworld Yes, that's a problem. Maybe limit should be doubled? Or do second request if the result of reject! is empty.. But even then there is no guarantee new items won't be rejected.

I am reminded that @akihikodaki wrote a version of the home query that includes the rules from FeedManager#filter?. Perhaps this is the place to dig that up and use it...

akihikodaki: > I wanted to do it this way to avoid copying the array (reject! mutates existing array afaik)

`to_a` copies an internal array, so the number of copy is same. The below is the long explanation.

Here is the implementation of `to_a`:
```Ruby
# Converts relation objects to Array.
def to_a
    records.dup
end
```
https://github.com/rails/rails/blob/v5.1.1/activerecord/lib/active_record/relation.rb#L249

`reject` is defined in `Enumerable`. `reject` itself copies the source, but there is no additional copy. The source is the very array copied in `to_a`.
> The class must provide a method `each`
https://docs.ruby-lang.org/en/2.4.0/Enumerable.html

```Ruby
    delegate :to_xml, :encode_with, :length, :each, :uniq, :to_ary, :join,
         :[], :&, :|, :+, :-, :sample, :reverse, :compact, :in_groups, :in_groups_of,
         :to_sentence, :to_formatted_s, :as_json,
         :shuffle, :split, :index, to: :records
```
https://github.com/rails/rails/blob/v5.1.1/activerecord/lib/active_record/relation/delegation.rb

In summary, both of `to_a` and `reject` are copying the same array once.

Gargron: @akihikodaki Done!

clworld (approved):
Looks good to me.

app/controllers/concerns/user_tracking_concern.rb
```diff
@@ -27,4 +27,9 @@ def user_needs_sign_in_update?
   def user_needs_feed_update?
     current_user.last_sign_in_at < REGENERATE_FEED_DAYS.days.ago
   end
+
+  def regenerate_feed!
+    Redis.current.setnx("account:#{current_user.account_id}:regeneration", true) == 1 && Redis.current.expire("account:#{current_user.account_id}:regeneration", 3_600 * 24)
```
clworld: SETNX + EXPIRE can be performed by one redis command. https://redis.io/commands/set
Redis-rb seems support them. https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L735

clworld: Redis.current.set("account:#{current_user.account_id}:regeneration", true, nx: true, ex: 3_600 * 24)

akihikodaki (approved):

PR #3725, opened 2017-06-12
Add links to project homepage

Gargron:
- Link Mastodon in "Powered by Mastodon" to joinmastodon.org

Where else should we put the link? Should I replace all "Source code" links leading to GitHub to lead to joinmastodon.org instead?


unarist:
Many instances use "Source code" links to link to their own forked repository I think.

I think "Powered by Mastodon" is appropriate for that, but if you want more, how about add link to `about.about_mastodon` text?

image

Gargron:
@unarist I do not want to put extra links into that text, to not spread the reader's attention into too many different directions.

I suppose the PR is ready as-is then.

ykzts (approved):

clworld:
I think changing about.other_instances link is neutral choice.
I already changed about.other_instances link to https://instances.mastodon.xyz/list currently.

PR #3728, opened 2017-06-13
Account deletion

### LINKED ISSUE #109: Delete account function?
Where is that function or when comes it?

### LINKED ISSUE #1175: Delete / Remove Account As Administrator
I saw pre-existing tickets regarding the ability for users to delete their accounts and I understand that is currently under review and possibly will be implemented. However from an Administrator's stand point, one who has access to the backend of the site as well as the psql database, does anyone know of a way to remove an account entirely? I will admit I'm very new to psql and with the help of Google I'm able to at least navigate my way through the database, but would like some assistance on removing an account. This particular use case is the user who signed up on my instance was simply testing out the UI and they left a Toot requesting their account be deleted.

* * * *

Gargron:
- [x] UI for deleting account
- [x] Handle avatar/header removal from remote profile
- [x] Create alternative to SuspendAccountService that *also* removes User record
- [x] Add tests

Fix #1175
Fix #109

### DIFF
--- a/app/controllers/settings/deletes_controller.rb
+++ b/app/controllers/settings/deletes_controller.rb
@@ -11,7 +11,7 @@ class Settings::DeletesController < ApplicationController
 
   def destroy
     if current_user.valid_password?(delete_params[:password])
-      Admin::SuspensionWorker.perform_async(current_user.account_id)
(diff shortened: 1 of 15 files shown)


beatrix-bitrot (approved):
LGTM

i want to test it

PR #3730, opened 2017-06-13
Use Class and Property Decorators

ykzts:
ref https://tc39.github.io/proposal-decorators/

### before

```javascript
class C extends React.Component {
  /* ... */
}

export default connect(makeMapStateToProps)(injectIntl(C));
```

### after

```javascript
@connect(makeMapStateToProps)
@injectIntl
export C extends React.Component {
  /* ... */
}
```


geta6:
Looks good to me.
`transform-decorators-legacy` is reliable transform plugin.

sorin-davidoi:
Just as a sidenote, decorators are not currently in the JavaScript standard. It is likely that they will be, but right now they are not. I would personally incline to keep the codebase as close as possible to the standard (on the other hand, dynamic imports are also not in the standard and yet we use them).

ykzts:
resolved conflicts.

Gargron (approved):

Gargron:
@sorin-davidoi We use various transpilations *anyway*, this cleans up boilerplate code massively, so I am for it!

PR #3735, opened 2017-06-14
Batched remove status service

Gargron:
Builds on top of #3728

- `BatchedRemoveStatusService` batches various operations as much as possible
- `SuspendAccountService` uses `BatchedRemoveStatusService`
- `Pubsubhubbub::DistributionWorker` can now accept arrays of stream entry IDs and unite them in a single PuSH payload (or two payloads, if both hidden and not hidden stream entries need to be delivered)

### DIFF
--- a/app/services/batched_remove_status_service.rb
+++ b/app/services/batched_remove_status_service.rb
@@ -36,8 +36,8 @@ class BatchedRemoveStatusService < BaseService
       batch_salmon_slaps(status) if status.local?
     end
 
-    Pubsubhubbub::DistributionWorker.push_bulk(@stream_entry_batches)
-    NotificationWorker.push_bulk(@salmon_batches)
+    Pubsubhubbub::DistributionWorker.push_bulk(@stream_entry_batches) { |batch| batch }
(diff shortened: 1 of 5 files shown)


app/services/batched_remove_status_service.rb
```diff
@@ -0,0 +1,114 @@
...
+    NotificationWorker.push_bulk(@salmon_batches)
+  end
+
+  private
+
+  def batch_push_updates(statuses)
+    stream_entry_ids = statuses.map { |s| s.stream_entry.id }
+
+    stream_entry_ids.each_slice(69) do |batch_of_stream_entry_ids|
+      @stream_entry_batches << [batch_of_stream_entry_ids]
```
unarist: - batch_feed_removal: **queues** removal
- batch_push_updates: **execute** removal in the method

These methods have different usage, so should have different name I think.

Also I'm thinking about making this method to only return sliced array , and queue it in `call()` method:

```
@stream_entry_batches << batch_feed_removal(status)

beatrix-bitrot (approved):
some very long/complex lines in this changeset but ok

PR #3743, opened 2017-06-14
Save settings when they are changed

### LINKED ISSUE #3738: Regex filters no longer saving after page reload
For the last several days, I've noticed on mastodon.social that regex filters on the Home and Local timelines don't save properly between page loads; a previous version of my filters reappears when I reload the page or open m.s in a new browser. Any new filters I set up perform correctly until I reload the page, but disappear on reload.

I can place this change to after the UK election, since the filters I set up to moderate that content are what appear when I load the page.

@beatrix-bitrot confirmed that this happens on glitch.social as well, which is running master.

* * * *

Gargron:
Fix #3738


beatrix-bitrot (approved):
LGTM

will test on glitch.social

PR #3745, opened 2017-06-14
Fix account delete form not accepting password

Gargron:
And update suspended account before removing content for quicker feedback to end-users


beatrix-bitrot (approved):
lgtm, will test asap

PR #3746, opened 2017-06-14
Enable localization fallback to English on Rails side

unarist:
Of course, we should have translation for all languages, but if we forgot that...

image

...It's should be avoided.


Gargron:
I think it's on by default in production and off in development, because in development you should notice missing translations. Is that the case?

unarist:
Oh, sorry. I understand.

PR #3750, opened 2017-06-14
Fix regression from #3748 - properly set defaultColumns

Gargron:


beatrix-bitrot (approved):
will test

unarist:
Oh, thanks for fix it.

PR #3754, opened 2017-06-15

akihikodaki:

This is succeeding #3696.

### DIFF
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -17,11 +17,7 @@ module Localized
   end
 
   def default_locale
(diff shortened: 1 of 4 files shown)


Gargron:
Imagine that someone goes on a Finnish instance, with DEFAULT_LOCALE=fi and they change the language to German, because they don't speak Finnish.

Fallback to Finnish would be useless, while fallback to English would (probably) be understood.

I think the fallback should not be customized.

akihikodaki:
Isn't it true also for Web? Actually I'm worrying about Chinese users of Pawoo, whose default locale is Japanese.

PR #3755, opened 2017-06-15
changed defaults

LottieVixen:
These changes are to reflect the documentation in production-guide.md

reason? I believe setting up a default instance should be possible with minimal changes to config files.


KellerFuchs (approved):
Yay!
Thanks for trying to make running instances easier.

akihikodaki (approved):

Gargron:
Okay, hold up. The previous defaults aren't random either. They facilitate the docker-compose deployment method. Should we really change it around? Documentation would need to be updated too. The way I see it, if Docker is the "easier" way of deploying, we're making it easier with fitting default configuration. When someone wants to deploy standalone, they have the technical prowess to know to adjust these values in the config.

beatrix-bitrot:
@Gargron since the docker one is more automated can it have its own .env file?

akihikodaki:
So we could have `.env.production.docker.sample` and `.env.production.standalone.sample`, or something sort of?

nightpool:
yeah i'm gonna throw my opinion behind gargron here. the "default" way to deploy mastodon should be using docker.

beatrix-bitrot:
in that case the docker stuff could stay in the file as is and like @akihikodaki said we could add a new file that matches what's in the documentation perhaps?

nightpool (changes requested):
see consensus above

PR #3756, opened 2017-06-15
Set default From address in config

akihikodaki:
The old implementation sets default `From` address in mailers. It sets the address from `SMTP_FROM_ADDRESS`, or `notifications@localhost`. The behavior is occasionally undesired results.

In production environment, `notifications@localhost` is likely to be incorrect.
In testing environment, the email address should not be varied by a environment variable.

After appling this change,

In production environment, it will throw an exception when launching Mastodon.
In testing environment, the address will be fixed with `notifications@localhost`.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < ActionMailer::Base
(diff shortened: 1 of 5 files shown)


mjankowski (approved):

aral:
Workaround to install updates following this merge:

`scalingo env-set SMTP_FROM_ADDRESS="<some email address>"`

PR #3758, opened 2017-06-15
Change NSFW button to Hide Image icon

coilysiren:
WIP

Closes https://github.com/tootsuite/mastodon/issues/2696

Current status:

image

Still to do:

- get a font awesome style icon
- DRY the button class ???

PR #3759, opened 2017-06-15
Replace TextIconButton for SensitiveButton to IconButton

### LINKED ISSUE #2696: Change text of the "NSFW" feature to "Hide Image" or similar
The label "NSFW" is not suitable for the feature's use case, for several reasons

- In practice, I have never hidden an image because it was "Not Safe For Work", I hide the image because it contained sensitive content. I believe this aligns with other people's use of this feature
- What is / is not safe for work depends on differing cultural contexts. Is me wearing a tank top NSFW somewhere? Mastodon is quite good at not requiring users to think outside their default culture context, but the NSFW button is the exception to that
- The word NSFW is, in general, used to hide erotic content. This is made explicit in an existing PR, even https://github.com/tootsuite/mastodon/pull/1974. The thing about that though, is that for sex workers erotic content is not NSFW, **erotic content is their work**. This goes beyond a mere semantics argument, its actually a tactic used to silence sex workers (the overwhelming majority of whom are women, which makes this issue functionally misogynist)
- ([I'm told that](https://toot.cat/@polymerwitch/456191)) the NSFW button does not have a translation option, so any resolution to this issue should also add this button to the translation interface

And just to be totally clear, my suggestion is that "NSFW" be changed to something akin to

* sensitive image
* hide image
* hide
* CW imagery

* * * *

- [N/A] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ykzts:
Resolve #2696.

Alternate of #3758.

### demo

demo movie

### DIFF
--- a/app/javascript/mastodon/features/compose/containers/sensitive_button_container.js
+++ b/app/javascript/mastodon/features/compose/containers/sensitive_button_container.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import IconButton from '../../../components/icon_button';
 import { changeComposeSensitivity } from '../../../actions/compose';
 import Motion from 'react-motion/lib/Motion';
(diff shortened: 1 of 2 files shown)


masarakki:
Good

Gargron:
I just wish font-awesome icons actually lined up horizontally :(

stephenburgess8:
I know line-height is used to adjust the positioning of the other buttons. Maybe it's possible to add a classname [e.g. `sensitive-icon-button`] and add line height here to drop the icon down by a couple pixels, as long as it doesn't push the container down.

ykzts:
middle

ykzts:
@Gargron 
> I just wish font-awesome icons actually lined up horizontally :(

I added line-height. It seems that it is horizontally.

beatrix-bitrot (approved):

beatrix-bitrot:
@Gargron please merge this or request changes

ginsterbusch:
hm .. using Font Awesome with the fa-fw class didnt work?

PR #3779, opened 2017-06-15
Collapsible toots

progval:
This PR is a rebase of #658 on the latest master.

However, I could not test it locally, but I'm sending the PR anyway in case you can test it yourself.

Thanks


ProgVal:
Meh, nevermind, both Codeclimate and Travis complain about this patch and I have no idea how to fix it

PR #3791, opened 2017-06-16
Position dropdown menu from left

ghost:
Base dropdown menu position on the left side of the page instead of the right so people can customize the width of columns in CSS without breaking them.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -217,14 +217,15 @@
 .dropdown--active::after {
   content: "";
   display: block;
-  position: absolute;
+  position: relative;
   width: 0;
   height: 0;
   border-style: solid;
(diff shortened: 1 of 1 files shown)


app/javascript/styles/components.scss
```diff
@@ -1203,7 +1200,7 @@
   }
 
   &.dropdown__right {
-    right: 0;
+    left: 160px;
```
unarist: `.dropdown__right` should be positioned from right side, because contents may be wider. For example, Occitan language:

image

unarist: @rainyday Could you fix this issue?

Gargron:
Dropdowns were reworked so now the positioning doesn't rely on CSS hacks and is a lot more accurate. Thanks though!

PR #3792, opened 2017-06-16
moved save button

LottieVixen:
As requested per [issue 3770](https://github.com/tootsuite/mastodon/issues/3770)


app/views/settings/preferences/show.html.haml
```diff
@@ -44,6 +47,3 @@
 
   .fields-group
     = f.input :setting_auto_play_gif, as: :boolean, wrapper: :with_label
-
-  .actions
-    = f.button :button, t('generic.save_changes'), type: :submit
```
ykzts: I think that it would be better to leave this. I think that it is better to have it both above and below.

beatrix-bitrot: makes sense

I don't think it hurts to have it in both places

@foxsan48 please add it back at the bottom~

app/views/settings/preferences/show.html.haml
```diff
@@ -44,6 +47,6 @@
 
   .fields-group
     = f.input :setting_auto_play_gif, as: :boolean, wrapper: :with_label
-
+
```
ykzts: remove unnecessary indent.

ykzts (approved):

beatrix-bitrot:
@Gargron please merge or close this if you won't merge it

PR #3793, opened 2017-06-16
Enable filtering toots with visibility

yamazaki-sensei:
I am a member of developing Naumanni, a kind of mastodon client.

This PR is to filter toots from status API with `visibility` parameter, in order to implement talking on mastodon DM.

Thanks.

### DIFF
--- a/app/controllers/api/v1/accounts/statuses_controller.rb
+++ b/app/controllers/api/v1/accounts/statuses_controller.rb
@@ -60,7 +60,9 @@ class Api::V1::Accounts::StatusesController < Api::BaseController
   end
 
   def filter_with_visibility_scope(visibility)
-    Status.where(visibility: Status.visibilities[visibility])
+    permitted_account_statuses.tap do |statuses|
+      statuses.merge!(Status.where(visibility: Status.visibilities[visibility]))
+    end
   end
 
   def pagination_params(core_params)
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/accounts/statuses_controller_spec.rb
```diff
@@ -35,4 +35,21 @@
       expect(response).to have_http_status(:success)
     end
   end
+
+  describe 'GET #index with valid visibility' do
+    it 'retures http success' do
```
pe-ter-: not `retures` but `returns`..?

almichest: Right. It must be fixed.

app/controllers/api/v1/accounts/statuses_controller.rb
```diff
@@ -58,6 +59,10 @@ def no_replies_scope
     Status.without_replies
   end
 
+  def filter_with_visibility_scope(visibility)
+    Status.where(visibility: Status.visibilities[visibility])
```
Gargron: You can't do this because it would likely conflict with Status#permitted_for and possibly leak statuses you aren't authorized to view.

Gargron (changes requested):

pe-ter-:
Hey @Gargron, now I took over this PR and fixed you pointed out.
What do you think?

app/controllers/api/v1/accounts/statuses_controller.rb
```diff
@@ -31,6 +31,7 @@ def account_statuses
     default_statuses.tap do |statuses|
       statuses.merge!(only_media_scope) if params[:only_media]
       statuses.merge!(no_replies_scope) if params[:exclude_replies]
+      statuses.merge!(filter_with_visibility_scope(params[:visibility])) if params[:visibility]
```
Gargron: When anonymous viewer requests this, the timeline's default filter is "WHERE visibility IN (public, unlisted)", doesn't this essentially replace that, allowing access to "WHERE visibility = private"? This is the main reason I haven't merged this PR, and I don't think I should. This messes with access controls.

Issue #3621, opened 2017-06-07
simple HTML tag not respected (with content type="html" in atom)


Shnoulle:
GnuSocial with mardown plugin allow to use markdown in notice. Then using `:100::100::100:` sended as `<code>:100::100::100:</code>` in atom (`content type="html"`).

But mastodon replace `:100:`by 💯 . 

Link of the original notice : `https://ostatus.shnoulle.net/api/statuses/show/708914.atom` , showing in mastodon like `https://framapiaf.org/users/framasky/updates/78376` or screenshot on 1.4.1 version

![capture du 2017-06-07 11-21-44](https://user-images.githubusercontent.com/1439428/26871534-d0f4b19a-4b73-11e7-9d43-a0075bdbe01b.png)

Same issue with `<em>Emphase</em>` and `<strong>Fort</strong>` : https://ostatus.shnoulle.net/api/statuses/show/708954.atom

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate (but unsure for the good word for searc)
- [x] This bug happens on a 1.4.1

Gargron:
1. em, strong and code are stripped out by sanitization. only p, br, span and a are allowed
2. the emoji conversion happens in JS through the emojione library

nolanlawson:
Yeah this seems like a security issue more than anything (e.g. against XSS). I'm not sure there's much value in supporting unsanitized HTML. The emoji conversion is also for the benefit of Mastodon users who use the emoji picker, which inserts e.g. `:100:`.

Shnoulle:
What ! XSS ! with simple HTML tag ! You kidding me ? You already allowing span why not strong, pre, em and p.

And about `:100:` : maybe it's cool for mastodon user but : if i want to send `:100:` how i do it ?

The atom content must rules.

PR #3810, opened 2017-06-16
Whitelist allowed classes for federated statuses

### LINKED ISSUE #3790: Whitelist classes for HTML elements in statuses
It's currently possible for a malicious OStatus server to send statuses that spoof parts of Mastodon's UI without violating the constraints of `sanitize_config.rb`:

```
<p>lewd <a class="status__content__spoiler-link" href="http://malware.ru/">Show more</a></p>
```

* * * *

nightpool:
Allowed classes are currently:

 - Any microformats class (h/p/u/dt/e-*)
 - the classes `mention`, `hashtag`, `ellipses` and `invisible`.

this last one is somewhat suspect, but Mastodon currently uses it to render truncated links.

Before this pull request gets merged, I would like more input on which classes should be allowed and which should be restricted. For example, I think that fa-spin is fine to be allowed, but I feel weird whitelisting it explicitly. It would be nice to allow some amount of simple formatting (opening the door for things like #853 and #3621 to happen on a federation level), but I don't know what classes we already have that could do similar things.

resolves #3790

### DIFF
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -39,7 +39,7 @@ class Sanitize
       },
 
       transformers: [
-        CLASS_WHITELIST_TRANSFORMER
+        CLASS_WHITELIST_TRANSFORMER,
       ]
     )
 
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -4,6 +4,21 @@ class Sanitize
   module Config
     HTTP_PROTOCOLS ||= ['http', 'https', :relative].freeze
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

app/lib/sanitize_config.rb
```diff
@@ -21,7 +36,11 @@ module Config
 
       protocols: {
         'a' => { 'href' => HTTP_PROTOCOLS },
-      }
+      },
+
+      transformers: [
+        CLASS_WHITELIST_TRANSFORMER
```
ykzts: Add comma (`,`).

PR #3820, opened 2017-06-17
l10n update for account deletion 

Quent-in:
+ some other changes.

### DIFF
--- a/config/locales/oc.yml
+++ b/config/locales/oc.yml
@@ -352,13 +352,13 @@ oc:
       subject: "%{name} a mes vòstre estatut en favorit"
     follow:
       body: "%{name} vos sèc ara !"
-      subject: "%{name} es a vos sègre ara"
+      subject: "%{name} vos sèc ara"
     follow_request:
-      body: "%{name} a demandar a vos sègre"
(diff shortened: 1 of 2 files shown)


config/locales/oc.yml
```diff
@@ -342,13 +352,13 @@ oc:
...
+      subject: "%{name} vos sèc ara"
     follow_request:
-      body: "%{name} a demandar a vos sègre"
+      body: "%{name} a demandat a vos sègre"
       subject: 'Demanda d’abonament : %{name}'
     mention:
-      body: 'Sètz estat mencionat per %{name} dins :'
-      subject: Sètz estat mencionat per %{name}
+      body: '%{name} vos a mencionat dins :'
+      subject: %{name} vos a mencionat
```
unarist: This translation must be quoted because the string starts with percent character.

unarist (changes requested):
Travis is failing due to the invalid YAML syntax.

config/locales/oc.yml
Quent-in: It should be fixed now :)

unarist (approved):

PR #3828, opened 2017-06-17
Fix streaming server. Redis connection subscribe for each channel.

deflis:
I fix streaming server issue.

* Redis connection subscribe all channel now.
* This version is Redis connection subscribe for each connected channel.


clworld (approved):
This PR may reduce traffic of Redis server when running multiple/clustered Streming API server.
Looks good to me.
I tested this on my staging environment. It seems OK.

Issue #3776, opened 2017-06-15
Timeline jiggles/jitters when lots of incoming toots are arriving


nolanlawson:
Mastodon v1.4.3. Steps to repro:

- Scroll down on a busy timeline
- Stop scrolling
- Wait for new toots to arrive
- Watch the timeline jitter

I can repro in Chrome, Edge, and Firefox, on both macOS and Windows.

![poorshamefulargali](https://user-images.githubusercontent.com/283842/27197122-f4c30da2-51c1-11e7-835d-75196a0865ee.gif) ![composedrashgelding](https://user-images.githubusercontent.com/283842/27197123-f4c348f8-51c1-11e7-8c5f-d810e1b19d65.gif) ![spryjaggedjanenschia](https://user-images.githubusercontent.com/283842/27197124-f4d99cfc-51c1-11e7-95e1-872e78cef613.gif)

I think I can tackle this one; my hunch is that we can just check the scrollTop before writing it again and not write it if it's within ~5px of the previous position. My only worry is that that might lead to layout thrashing, but this can be tested.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nolanlawson:
This is now fixed for Chrome (#3832) but needs a redesign to be fixed for all browsers.

nolanlawson:
should be fixed by #3972

PR #3832, opened 2017-06-17
Fix jittery scrolling for Chromium browsers (#3776)

nolanlawson:
This is a partial fix for #3776 (jiggling/jittery scrolling). This PR only fixes the issue for Chrome/Chromium. Here's why:

The main issue is that `scrollTop` doesn't provide subpixel precision – only integers. So if we set the `scrollTop` to an integer then it will always cause jiggling, because it may be off by a subpixel.

Unfortunately there doesn't seem to be any way to get a subpixel precision for the scroll. So the next best thing is to avoid re-setting an integer `scrollTop` when the `scrollTop` is already that exact same integer. This turns out to work for Chrome because they have [an intervention](https://www.chromestatus.com/feature/5700102471548928) to avoid moving the scroll position when elements are added to the top of a list. So the `scrollTop` is already correct; there's no reason to re-set it.

So by checking the `scrollTop` before setting it, we remove the jitter from Chrome. Unfortunately this PR does not fix Edge, Firefox, or Safari. The only way to fix the jitter in those browser (I believe) would be to avoid adding the elements entirely, e.g. similar to how Twitter does it with the "show more tweets" button.

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -6,6 +6,7 @@ import StatusContainer from '../containers/status_container';
 import LoadMore from './load_more';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 import IntersectionObserverWrapper from '../features/ui/util/intersection_observer_wrapper';
+import { debounce } from 'lodash';
 
 class StatusList extends ImmutablePureComponent {
 
@@ -29,7 +30,7 @@ class StatusList extends ImmutablePureComponent {
 
   intersectionObserverWrapper = new IntersectionObserverWrapper();
 
-  handleScroll = (e) => {
+  handleScroll = debounce((e) => {
     const { scrollTop, scrollHeight, clientHeight } = e.target;
     const offset = scrollHeight - scrollTop - clientHeight;
     this._oldScrollPosition = scrollHeight - scrollTop;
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/status_list.js
```diff
@@ -41,16 +42,26 @@ class StatusList extends ImmutablePureComponent {
     } else if (this.props.onScroll) {
       this.props.onScroll();
     }
-  }
+  }, 200, {
+    trailing: true,
+  });
```
nolanlawson: I can't see any reason not to add a debounce; we don't need precision beyond ~200 milliseconds to update the user's scroll position while they're scrolling. The `trailing` is to ensure that the scroll position truly gets updated even when the user stops scrolling.

ykzts (approved):

Gargron (approved):

app/javascript/mastodon/components/status_list.js
unarist: This prevents TIMELINE_SCROLL_TOP and loading more processing until they stops scroll. For example, below (trivial?) issues will happens:

- TIMELINE_SCROLL_TOP: The unread indicator won't appears when new toot received while scrolling.
- loading more: Fetch request will be delayed about 200ms.

nolanlawson: Looks like TIMELINE_SCROLL_TOP is [already debounced elsewhere](https://github.com/tootsuite/mastodon/blob/37dbfa4cd7f97cfbab68d1bfd774aabf5ca2e0af/app/javascript/mastodon/features/ui/containers/status_list_container.js#L53-L64)?

As for delaying fetch requests by 200ms, that doesn't seem like a big deal to me; I was more worried about continuously setting `scrollTop`while undebounced and thus invoking layout continuously.

PR #3842, opened 2017-06-18
Filter direct statuses in Status.as_home_timeline

akihikodaki:
The classes using `Status.as_home_timeline`, namely `Feed` and `PrecomputeFeedService` are expected to filter direct statuses as `FanOutWriteService` does, but their filtering were incomplete or missing.

This commit solves the problem by filtering direct statuses in `as_home_timeline` as the other similar methods such as `as_public_timeline` does.

Labelling _priority - high_; if my understanding is correct, the addressed bug had unexpectedly caused to make direct statuses visible for followers in the following certain condition:
* They were offline for 24 hours (which forces to regenerate feed) AND
* They reached to `api/v1/timelines/home` before the regenerations finishes.

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -131,7 +131,15 @@ class Status < ApplicationRecord
     end
 
     def as_home_timeline(account)
-      where(account: [account] + account.following)
+      # 'references' is a workaround for the following issue:
+      # Inconsistent results with #or in ActiveRecord::Relation with respect to documentation Issue #24055 rails/rails
+      # https://github.com/rails/rails/issues/24055
+      references(:mentions)
+        .where.not(visibility: :direct)
+        .or(where(mentions: { account: account }))
+        .where(follows: { account_id: account })
+        .or(references(:mentions, :follows).where(account: account))
+        .left_outer_joins(account: :followers).left_outer_joins(:mentions).group(:id)
     end
 
     def as_public_timeline(account = nil, local_only = false)
--- a/app/services/precompute_feed_service.rb
+++ b/app/services/precompute_feed_service.rb
@@ -23,11 +23,7 @@ class PrecomputeFeedService < BaseService
   end
 
   def process_status(status)
-    add_status_to_feed(status) unless skip_status?(status)
-  end
-
-  def skip_status?(status)
-    status.direct_visibility? || status_filtered?(status)
+    add_status_to_feed(status) unless status_filtered?(status)
   end
 
   def add_status_to_feed(status)
--- a/spec/models/status_spec.rb
+++ b/spec/models/status_spec.rb
@@ -181,15 +181,18 @@ RSpec.describe Status, type: :model do
   end
 
   describe '.as_home_timeline' do
+    let(:account) { Fabricate(:account) }
+    let(:followed) { Fabricate(:account) }
+    let(:not_followed) { Fabricate(:account) }
+
     before do
-      account = Fabricate(:account)
-      followed = Fabricate(:account)
-      not_followed = Fabricate(:account)
       Fabricate(:follow, account: account, target_account: followed)
 
-      @self_status = Fabricate(:status, account: account)
-      @followed_status = Fabricate(:status, account: followed)
-      @not_followed_status = Fabricate(:status, account: not_followed)
+      @self_status = Fabricate(:status, account: account, visibility: :public)
+      @self_direct_status = Fabricate(:status, account: account, visibility: :direct)
+      @followed_status = Fabricate(:status, account: followed, visibility: :public)
+      @followed_direct_status = Fabricate(:status, account: followed, visibility: :direct)
+      @not_followed_status = Fabricate(:status, account: not_followed, visibility: :public)
 
       @results = Status.as_home_timeline(account)
     end
@@ -198,8 +201,21 @@ RSpec.describe Status, type: :model do
       expect(@results).to include(@self_status)
     end
 
+    it 'includes direct statuses from self' do
+      expect(@results).to include(@self_direct_status)
+    end
+
     it 'includes statuses from followed' do
       expect(@results).to include(@followed_status)
+    end
+
+    it 'includes direct statuses mentioning recipient from followed' do
+      Fabricate(:mention, account: account, status: @followed_direct_status)
+      expect(@results).to include(@followed_direct_status)
+    end
+
+    it 'does not include direct statuses not mentioning recipient from followed' do
+      expect(@results).not_to include(@followed_direct_status)
     end
 
     it 'does not include statuses from non-followed' do


akihikodaki:
@alpaca-tc Could you test the following query? I'm not really confident with the performance of the query. This query change is related to #2967.
```SQL
> EXPLAIN ANALYZE SELECT "statuses".* FROM "statuses" LEFT OUTER JOIN "mentions" ON "mentions"."status_id" = "statuses"."id" WHERE ((("statuses"."visibility" != 3) OR "mentions"."account_id" = 1) AND "statuses"."account_id" IN (SELECT "accounts"."id" FROM "accounts" INNER JOIN "follows" ON "accounts"."id" = "follows"."target_account_id" WHERE "follows"."account_id" = 1 ORDER BY follows.id desc) OR "statuses"."account_id" = 1) GROUP BY "statuses"."id" ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                          QUERY PLAN                                                                          
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=405.33..482.46 rows=100 width=438) (actual time=22.214..22.573 rows=100 loops=1)
   ->  Group  (cost=405.33..26572.85 rows=33926 width=438) (actual time=22.214..22.558 rows=100 loops=1)
         ->  Nested Loop Left Join  (cost=405.33..26488.03 rows=33926 width=438) (actual time=22.212..22.467 rows=102 loops=1)
               Filter: ((((statuses.visibility <> 3) OR (mentions.account_id = 1)) AND (hashed SubPlan 1)) OR (statuses.account_id = 1))
               Rows Removed by Filter: 33
               ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.00..6060.13 rows=66519 width=438) (actual time=0.08..0.054 rows=131 loops=1)
               ->  Index Scan using index_mentions_on_status_id on mentions  (cost=0.00..0.28 rows=1 width=12) (actual time=0.001..0.001 rows=0 loops=131)
                     Index Cond: (status_id = statuses.id)
               SubPlan 1
                 ->  Nested Loop  (cost=0.00..405.06 rows=106 width=8) (actual time=6.248..22.129 rows=128 loops=1)
                       ->  Index Scan Backward using follows_pkey on follows  (cost=0.00..19.33 rows=106 width=8) (actual time=6.239..6.304 rows=128 loops=1)
                             Filter: (account_id = 1)
                             Rows Removed by Filter: 107
                       ->  Index Only Scan using accounts_pkey on accounts  (cost=0.00..3.63 rows=1 width=4) (actual time=0.123..0.123 rows=1 loops=128)
                             Index Cond: (id = follows.target_account_id)
                             Heap Fetches: 117
 Total runtime: 23.099 ms
(17 rows)
```

Gargron (approved):

akihikodaki:
c5b40167085cfec2939940f645ed87af6e846a30 updated the query and edited https://github.com/tootsuite/mastodon/pull/3842#issuecomment-309264604 as well.

alpaca-tc:
@akihikodaki My apologies for the late reply. here you are.

https://pawoo.net/web/accounts/1

```
Limit  (cost=1.72..874401.38 rows=100 width=332) (actual time=70.674..9965.812 rows=100 loops=1)
  ->  Group  (cost=1.72..113645725.49 rows=12997 width=332) (actual time=70.672..9965.745 rows=100 loops=1)
        Group Key: statuses.id
        ->  Nested Loop Left Join  (cost=1.72..113645693.00 rows=12997 width=332) (actual time=70.670..9965.443 rows=105 loops=1)
              Filter: ((((statuses.visibility <> 3) OR (mentions.account_id = 1)) AND (follows.account_id = 1)) OR (statuses.account_id = 1))
              Rows Removed by Filter: 6189510
              ->  Nested Loop Left Join  (cost=0.87..12380809.91 rows=21446578 width=336) (actual time=0.015..301.481 rows=78800 loops=1)
                    ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.44..1522700.58 rows=21446578 width=332) (actual time=0.009..66.121 rows=76118 loops=1)
                    ->  Index Scan using index_mentions_on_status_id on mentions  (cost=0.43..0.49 rows=2 width=12) (actual time=0.002..0.002 rows=0 loops=76118)
                          Index Cond: (status_id = statuses.id)
              ->  Nested Loop Left Join  (cost=0.85..4.60 rows=6 width=8) (actual time=0.008..0.099 rows=79 loops=78800)
                    ->  Index Only Scan using accounts_pkey on accounts  (cost=0.42..0.44 rows=1 width=4) (actual time=0.003..0.003 rows=1 loops=78800)
                          Heap Fetches: 57714
                    ->  Index Scan using index_follows_on_target_account_id on follows  (cost=0.43..3.03 rows=113 width=8) (actual time=0.004..0.064 rows=78 loops=78800)
                          Index Cond: (target_account_id = accounts.id)
Planning time: 0.644 ms
Execution time: 9965.895 ms
```

---

https://pawoo.net/web/accounts/34

```
Limit  (cost=1.72..874377.10 rows=100 width=332) (actual time=7.373..458.562 rows=100 loops=1)
  ->  Group  (cost=1.72..113651313.56 rows=12998 width=332) (actual time=7.373..458.507 rows=100 loops=1)
        Group Key: statuses.id
        ->  Nested Loop Left Join  (cost=1.72..113651281.06 rows=12998 width=332) (actual time=7.370..458.260 rows=100 loops=1)
              Filter: ((((statuses.visibility <> 3) OR (mentions.account_id = 34)) AND (follows.account_id = 34)) OR (statuses.account_id = 34))
              Rows Removed by Filter: 243110
              ->  Nested Loop Left Join  (cost=0.87..12381398.85 rows=21447637 width=336) (actual time=0.015..19.479 rows=4738 loops=1)
                    ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.44..1522772.91 rows=21447637 width=332) (actual time=0.009..4.247 rows=4592 loops=1)
                    ->  Index Scan using index_mentions_on_status_id on mentions  (cost=0.43..0.49 rows=2 width=12) (actual time=0.002..0.002 rows=0 loops=4592)
                          Index Cond: (status_id = statuses.id)
              ->  Nested Loop Left Join  (cost=0.85..4.60 rows=6 width=8) (actual time=0.009..0.076 rows=51 loops=4738)
                    ->  Index Only Scan using accounts_pkey on accounts  (cost=0.42..0.44 rows=1 width=4) (actual time=0.003..0.004 rows=1 loops=4738)
                          Heap Fetches: 4751
                    ->  Index Scan using index_follows_on_target_account_id on follows  (cost=0.43..3.03 rows=113 width=8) (actual time=0.004..0.049 rows=51 loops=4738)
                          Index Cond: (target_account_id = accounts.id)
Planning time: 0.660 ms
Execution time: 458.655 ms
```

nullkal:
I updated mstdn.jp to mastodon 1.4.4 and I saw long running queries I suspect this change is concerned.

image

alpaca-tc:
I just try to optimize the query. Please give me some advice if you have ideas or suggestions :)

akihikodaki:
The sequence of joining with `follows`, filtering with `follows`, joining with `mentions` if required, and then filtering `mentions` could be faster.
To achieve that, we need to tell the planner to join `follows` eariler than `mentions`, but I don't know how to do that.

alpaca-tc:
@akihikodaki How about the following query? This is still work in progress.

```
WITH following_account_ids AS (
  SELECT target_account_id
  FROM follows
  WHERE account_id = 34
), 
mentioned_status_ids AS (
  SELECT "mentions"."status_id" FROM mentions WHERE "mentions"."account_id" = 34
)

SELECT "statuses".*
FROM "statuses"
WHERE "statuses"."account_id" = 34 /* own statuses */
   OR ("statuses"."visibility" IN (0, 1, 2) AND "statuses"."account_id" IN (SELECT * FROM following_account_ids)) /* followings */
   OR ("statuses"."visibility" = 3          AND "statuses"."id" IN (SELECT * FROM mentioned_status_ids)) /* DM statuses */
ORDER BY "statuses"."id" DESC
LIMIT 200

account = Account.find(XXX)

following_account_ids = Follow.select(:target_account_id).where(account: account)
mentioned_status_ids = Mention.select(:status_id).where(account: account)

own_statuses = Status.where(account: account)
following_statuses = Status.where(visibility: [:public, :unlisted, :private]).where(account_id: following_account_ids)
direct_statuses = Status.where(visibility: [:direct]).where(id: mentioned_status_ids)

own_statuses.or(following_statuses).or(direct_statuses).limit(200)
```

alpaca-tc:
@akihikodaki IMO, to separate query is the best way. Could you review the following?

```
    def as_home_timeline(account, limit)
      following_account_statuses = where(account_id: account.following.ids)

      account_statuses = where(account: account)
      following_statuses = where(visibility: [:public, :unlisted, :private]).merge(following_account_statuses).reorder(nil).order(created_at: :desc)
      mentioned_statuses = Mention.joins(:status).where(account: account).merge(following_account_statuses)

      ids = account_statuses.limit(limit).ids + following_statuses.limit(limit).ids + mentioned_statuses.limit(limit).pluck(:status_id)
      where(id: ids).limit(limit)
    end
```

akihikodaki:
If we could use a raw query, `join_collapse_limit` would be the simplest and best solution. However, personally I'm not favor of a raw query and wonder if it can still be represented in one query. Subquery could be a solution.

akihikodaki:
By the way, you should be careful for
* the order and
* the limit

in case you merge multiple records. The order could be resolved with simple merge sort. The limit could be enhanced with `where('statuses.id > ?', the_records_we_already_have.last.id) if the_records_we_already_have.size >= limit` or something kind of.

Gargron:
I would be careful with limits, since PrecomputeFeedService wants to iterate over many more items (e.g. restore ~200 items in the home feed) than the live query. Perhaps for that purpose the methods should be split, or use an argument to decide whether to apply limits.

akihikodaki:
I'm adding such an argument to as-timeline methods since it is usable for this case and for Redis. Currently we have `Feed` model, but I think it should be `as_home_timeline` and `from_database` and `from_redis` should respectively be `as_home_timeline_from_database` and `as_home_timeline_from_redis`. All of them would have the argument for consistency.

Gargron:
mastodon.social is also currently having multiple long running queries (>650,000ms!) so I will probably have to tag a 1.4.5 when this is fixed before announcing the new release (rather than announcing 1.4.4, which I still haven't done publicly from my account/Mastodon account)

nightpool:
@Gargron is there no timeout on those queries/requests/workers? that seems like a more important thing to get fixed O.O

Gargron:
@nightpool There is a 90 sec timeout on request workers, however timeouts are a tricky business, because by interrupting the execution of code in arbitrary places you might end up with weird states (although transactions should take care of the worst cases)

Background workers have no timeouts by definition.

There is an option of "statement timeouts" in Rails for the database connection, however the setting is session-wide, which doesn't work with pgBouncer in transaction mode. The statement timeout can be set on pgBouncer level or Postgres level manually, however I do not know if that's a good idea, what value to take, and in any case it doesn't help most admins since it's outside of Mastodon itself.

PR #3844, opened 2017-06-18
Improve ActivityPub representations

### LINKED ISSUE #3438: ActivityPub: outbox pages contain unordered sets of activities
According to [the spec](https://www.w3.org/TR/activitystreams-core/#collections), `items` is a set, while `orderedItems` is an (ordered) list. This is due to the JSON-LD serialization of AS2: JSON-LD defaults to unordered sets. When using `list`, some clients may not be able to show the Activities in the right order. To fix this, the OrderedCollectionPages should contain `orderedItems` instead of `items`.

* * * *

### LINKED ISSUE #3439: ActivityPub: Outbox loads all posts into the cache, sometimes errors out
When loading the outbox of a user, all the public posts of that person are loaded into the cache, but eventually only the first and last post are used to determine the `max_id` and `since_id`, and the count (to determine `totalItems`). This results in the request taking a lot of time and loading all posts into memory (I've seen 8 seconds loading time and even 500's for a user with 5000 posts)

Example: Getting @Gargron's outbox at https://mastodon.social/api/activitypub/users/1/outbox just returns a 500.)

* * * *

### LINKED ISSUE #3458: ActivityPub: Boosts of federated posts are wrong
When an `Announce` activity in someone's outbox points to a `Create` activity from another instance, the `id` values still use the internal server. This means that external servers that use ActivityPub to federate with Mastodon will see duplicate Activities and Posts. The `id` value should point to the single source for the activity, so towards the source server, not the current instance's local representation of the post.

* * * *

Gargron:
- [x] Actor representation
- [x] Outbox representation
- [x] Followers collection representation
- [x] Following collection representation
- [x] Note representation

Note representation:

- [x] Account
- [x] Original vs reblog
- [x] Audience choice
- [x] Media attachments
- [x] Content
- [x] Content warning
- [x] URL
- [x] In reply to
- [x] Published
- [x] ~~~Sensitive content~~~ [Missing spec](https://github.com/swicg/general/issues/7)
- [x] ~~~Hashtags~~~ [Unclear spec](https://github.com/swicg/general/issues/6)
- [x] Mentions

Close #3438
Close #3439
Close #3458


evanminto (changes requested):
This is really thorough! Just two areas where I'm unsure:
* Did you intentionally remove paging from the Outbox? It looks like now the Outbox only shows the first page of results. This is acceptable according to the spec, just double-checking.
* If I'm reading this right, the fragment identifier `#create` doesn't match [the spec](https://www.w3.org/TR/activitypub/#obj-id) — the Create ID needs to either be publicly dereferencable or `null`. I suppose we could do `null` in this case, if we expect it to be basically a transient wrapper around the Note. That prevents other servers from attempting to dereference it and getting an unexpected value.

Still trying to get my VM up and running again so I can see this API actually running.

Gargron:
@evanminto 

>Did you intentionally remove paging from the Outbox?

Yes, since nothing is gonna use the outbox at first I opted not to introduce that yet.

>If I'm reading this right, the fragment identifier #create doesn't match the spec 

<https://octodon.social/users/cwebber/updates/143857>

>List the Create activities wrapping the objects.
>If you don't have Create activities as separate objects, you could do a hack with a "fragment identifier" like so:
>    {"id": "https://mastodon.social/users/Gargron/updates/3243062#create",
>     "type": "Create",
>     "object": {"type": "Note", "id": https://mastodon.social/users/Gargron/updates/3243062"}}

evanminto:
Re: Outbox, sounds good.

Re: `#create`, I'm... not sure I agree with Chris on that, though they're the spec co-author and I'm not. I want to double-check with them about it.

nightpool:
@evanminto I talked to chris about this, and I think the understanding we came to is that using fragments **does** count as publicly dereferenceable, because you can make the request and then use the object it returns. (using an algorithm to select the correct fragment identifier—json-ld is somewhat unclear on this, so we were considering including a guideline in activitypub on how to dereference fragment identifiers)

Gargron:
I have addressed the concern by turning the create URI to "/create" instead of "#create"

I have also added representation of hashtags and mentions, although hashtag representation is incomplete (spec is unclear). Also, sensitivity of the content still cannot be encoded in AP.

nightpool:
For the record—I think that `#create` is still the right way to go about this. this was the conversation I had with @cwebber about it:

> I think the best thing to do is to fetch the object at the uri, then search through the returned as2 object for the sub-object with the fragment.  Similar to how you'd "jump" to the right section in html based on the fragment.

and the original issue discussion:

https://github.com/w3c/activitypub/issues/224

Gargron:
@nightpool I had a later conversation with Chris about this as well. The problem is that "create" is above the "note" in hierarchy, so it wouldn't work that way - there is no reason for note JSON to return the create activity.

nightpool:
The other option is something like 

```json
{
    "id": "https://mastodon.social/users/Gargron/updates/3243062",
    "type": "Create",
    "object":  {
        "type": "Note",
        "id": null
    }
}
```

which is also supported by the spec (the two objects share the same identifier)

nightpool:
@Gargron This is why I think it needs to be hashed out a little more in the spec itself, but I would say the algorithm would be something like: 

1. dereference the URI into a JSON object
2. find the sub-object with an "id" equal to the URI you are trying to dereference

this would work in both cases?

evanminto (commented):
I like the new `/create` route for Create activities, but I still think the activities should be accessible at that URL. Right now the id doesn't point to a real resource. Not sure if you were planning on adding that.

That's my only concern.

Gargron:
@evanminto Okay, I rewrote all this stuff with serializers, which I think is a lot cleaner now.

However, the difference between Create/Announce and Note is really blurry still. For example! A local status can be a reblog, aka a wrapper around another status. So what to do when its URI is directly accessed? It's technically among the notes URIs, but it would need to be an Announce (or web redirect to URI of original?) It is the most frustrating thing. I suggest we leave this as-is and then adjust in future PRs.

I would also like to involve @cwebber in writing down an audience matrix for all of our visibility levels. There's cc, to, bcc, bto; and there's public, unlisted, followers+mentions-only and mentions-only (direct), and I have no good idea which values to fill in where.

Gargron:
The activity/object thing has been resolved.

cwebber:
> I would also like to involve @cwebber in writing down an audience matrix for all of our visibility levels. There's cc, to, bcc, bto

The difference between `to` and `cc` is IMO mostly presentational; `to` is the primary audience while `cc` is secondary.  You have a lot of flexibility to interpret that as you like since delivery will happen the same way.  `bcc` and `bto`... imo there's no reason to have `bto` because since these are both removed, there's nothing presentational about it...!  It can only be the origin server that plans to do anything different between primary/secondary for those blind address slots.

> and there's public, unlisted, followers+mentions-only and mentions-only (direct), and I have no good idea which values to fill in where.

Public post should be like:

```
{"to": ["https://www.w3.org/ns/activitystreams#Public"],
 "cc": ["https://social.example/alyssa/followers/"], ...}
```

(again, you can swap out to/cc as you wish, it barely matters, just my suggestion)

Followers-Only vs Unlisted?  I'm not sure what the difference is, it seems to me that they're the same:

```
{"to": ["https://social.example/alyssa/followers/"], ...}
```

I guess the difference is that one of them really is public but just doesn't get inserted in the Public timeline?  I'm not sure how to differentiate these though.

followers+mentions-only:

```
{"to": ["https://social.example/alyssa/followers/"]
 "cc": ["https://chatty.example/ben/"],
 "tag": [{"type": "Mention", "href": "https://chatty.example/ben/", "name": "@ben"}],
  ...}
```

mentions-only:

```
{"to": ["https://chatty.example/ben/"],
 "tag": [{"type": "Mention", "href": "https://chatty.example/ben/", "name": "@ben"}],
  ...}
```

Again, you have a lot of freedom on where to put to/cc.

Gargron:
@cwebber Implemented.

@evanminto Pls re-review.

puckipedia:
My two cents on the ActivityPub side: It seems the outbox has no `id` value, so my server refuses to load it in. The activities and objects though, it has no issue with it, no changes to either side of the connection.

.. however, the announces use a local `object`, and as such I can't use ActivityPub when getting stuff from a Mastodon server.

Gargron:
>however, the announces use a local object, and as such I can't use ActivityPub when getting stuff from a Mastodon server.

@puckipedia Please explain. It would return the object with a remote URI and all of its remote data as we know it, so it should be fine - the most important part being the ID. So what's wrong?

puckipedia:
I tested it out with a local server, and searched for https://mastodon.social/users/Gargron/updates/1 and then boosted it - this results in a proper status being loaded in the database, but the activitypub representation shows the actor being my instance's `admin`.

I am pretty sure this is because the object it's serializing is actually the boost, and, not the original status that was boosted. (`local?` seems to return true for the object that gets serialized into an `Announce`)

Gargron:
@puckipedia mastodon.social isn't running this PR! it's running 1.4.7. there is no live instance which uses this PR since it's not merged

puckipedia:
Okay, I found a more simple way: if I post a status on the admin account (which gets the url `https://example.com/users/admin/statuses/4`), and then boost it (even from the same account), the `object` contained in the `Announce` posts to `http://mastodon.puckipedia.com:3000/users/admin/statuses/5`.

While checking that, I realised I think it's best to add the audience (so `to` and `cc`) to the Activity too, as that may differ from the containing object

Gargron:
@puckipedia Fixed!

puckipedia:
Yay, that works! Kroeg now seems to have no issue loading the ActivityPub data!

evanminto (approved):
Nothing jumps out at me, so it looks good to me. I really like the serializers for the JSON-LD!

PR #3846, opened 2017-06-18
Unified styling of avatars

KScl:
Avatars now have one unified style across the entire application; the border size can be adjusted with $ui-avatar-border-size for those that prefer rounded avatars (or sharp corners).

This also removes the unnecessary motion effect given to avatars when viewing a user's profile from the web interface, as it clashed with everything else.

### DIFF
--- a/app/javascript/mastodon/features/account/components/header.js
+++ b/app/javascript/mastodon/features/account/components/header.js
@@ -5,9 +5,7 @@ import emojify from '../../../emoji';
 import escapeTextContentForBrowser from 'escape-html';
 import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
 import IconButton from '../../../components/icon_button';
-import Motion from 'react-motion/lib/Motion';
-import spring from 'react-motion/lib/spring';
-import { connect } from 'react-redux';
+import Avatar from '../../../components/avatar';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 
 const messages = defineMessages({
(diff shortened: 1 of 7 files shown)


Gargron:
I am sorry but I like the round->square animation and I believe that there is no need for consistency in shapes when the context of the displayed content is different. In timelines the avatars are informational, the space is to be used effectively. But when the avatar is large and centered over an area that displays the header image, a round shape is more appropriate. And a nice hover effect for expanding the corners. Design is sometimes subjective and I understand some people don't agree but I generally do not accept subjective design changes, otherwise a lot of the PRs would be changing this stuff back and forth.

Xkeeper0:
i, too, enjoy pointless inconsistency and distracting animations

STJrInuyasha:
This is more of an attempt to settle this argument once and for all. Consistent display of avatars is important for the people that actually draw and design their avatars, so they can adapt their design as necessary so no visual information is lost or hidden. It's impossible to do that if avatars look completely different based on where you're viewing them from.

And with it being a one-variable change to set rounded avatars in place of square ones, there wouldn't be further arguments over round vs square as everyone would get what they wanted one way or the other.

But as long as the inconsistency stays, I don't see the argument ever reaching a conclusion from either side.

unarist:
FYI: It can be disabled by CSS (i.e. customs.scss or Stylish):

```
.account__header__avatar { border-radius: 8% !important; }
```

zerojay:
The inconsistent look of the avatars is something that I find rather odd. Square with slightly rounded edges in the timeline, squared with larger rounded edges when looking at someone's profile (after clicking on their name) and then circular when looking at their page directly in a web browser. The inconsistency is somewhat jarring and means I don't know what parts of the profile image are going to be seen at any given time.

Gargron:
There are actually two styles:

- Square with rounded corners (timelines, profile avatar on hover)
- Round (profile avatar before hover)

The public profiles don't have the hover animation because they don't have anything dynamic at all because they're rendered as plain HTML. So there are a lot of dynamic parts missing and the templates are duplicates across React and Haml. But that's a different issue.

It's also useless to argue about predicting how the avatar is gonna be displayed. The web UI is just one application. Various native apps will follow different platform-specific design patterns, or make their own choices. Halcyon, for example, is a 1:1 replicate of the web Twitter UI. Your avatars will be displayed by software that isn't Mastodon, like GNU social, Friendica, Pleroma, Hubzilla, etc. You can't predict how it's going to be displayed. So that is a moot point. From an aesthetic point of view, as I outlined above, it makes sense to display a round image when it's overlaid on top of a header image, also when it's overall larger.

I thank @STJrInuyasha for the effort but I will close this PR.

Xkeeper0:
The "Well, we can't control how it shows up elsewhere" argument is so pointless it's almost laughable. This isn't a pull request to fix *other* applications -- it's to fix *this* one.

Also, there *are* actually three styles:

* Timeline view (~8% border radius)
* Profile view in a column, hovering (~33%)
* Profile view w/o hover (50%)

You can see the differences here: https://jsfiddle.net/phx80hyf/1/

PR #3850, opened 2017-06-19
Add Japanese translation "Action taken by"

noraworld:
Translated "Action taken by" into Japanese.

I don't think this is the best translation. If you find out better translation, I will change the present one.


clworld (approved):
I think it's good enough.

PR #3851, opened 2017-06-19
Unobserve status on unmount

alpaca-tc:
Fixes the following warning

```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Status component.
```

ezgif-3-ab639baddd

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -88,6 +88,16 @@ class Status extends ImmutablePureComponent {
     );
   }
 
+  componentWillUnmount () {
+    if (!this.props.intersectionObserverWrapper) {
+      // TODO: enable IntersectionObserver optimization for notification statuses.
+      // These are managed in notifications/index.js rather than status_list.js
+      return;
+    }
+
+    this.props.intersectionObserverWrapper.unobserve(this.props.id, this.node);
+  }
+
   handleIntersection = (entry) => {
     // Edge 15 doesn't support isIntersecting, but we can infer it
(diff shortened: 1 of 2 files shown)


Gargron (approved):

Gargron:
This seems to have created a blank space regression. I understand very little of this code so requesting @nolanlawson or @sorin-davidoi to take a look if there's something off here...

sorin-davidoi:
@Gargron Do you mean that the toot is rendered as empty? I cannot seem to reproduce.

Gargron:
@sorin-davidoi report by @beatrix-bitrot:

>there was a gap like that, but smaller
>directly above the post by wulf
>i scrolled down
>and when i scrolled up the gap was in a different place
>and larger?

Gargron:
@clworld was able to reproduce, and it wasn't present with this PR reverted, so the issue is in this patch somewhere. I just don't know what could have such an effect.

sorin-davidoi:
Found it, working on a fix.

sorin-davidoi:
@Gargron The bug not related to this PR. The `/api/v1/timelines/home` returns duplicate toots (check ids of the last request):

image

Gargron:
It looks like the screenshot shows websocket payloads? Which I suppose could have duplicates if toots arrive from different timelines or some such...

sorin-davidoi:
No, sorry, just was that in focus. It shows the second-to-last (`/home?max_id=11&limit=10`).

nolanlawson:
Oops, good catch :+1:

PR #3852, opened 2017-06-19
setting-for-account-deletable

masarakki:


Gargron:
I am not against this but I'd like to learn more about the context where this came up

app/controllers/settings/deletes_controller.rb
```diff
@@ -21,6 +22,10 @@ def destroy
 
   private
 
+  def check_enabled_deletion
+    redirect_to root_path if !Setting.open_deletion
```
ykzts: `unless Setting.open_deletion`

masarakki:
Because if account deletion is enabled, it's probable that someone say "DELETE ACCOUNT PLEASE" to others and it's make wold violent.

EzoeRyou:
Context I've heard from masarakki:

If the user can easily delete his account, there will be a lot of online harassment which goes something like:

"Yo! You worthless sucker!  Delete your account!"

Many people can't stand this kind of harassment and just follow the demand and delete his account. Then, they later want to revive the account, but it's technically impossible in current implementation.

If account deletion feature doesn't exist, this kind of harassment and subsequent unfortunate account deletion doesn't happen.
 
In an ideal world, harassment should be punished and prevented. But we can't censor all the toots before it is posted. So harassment happens and and there always some victims who's mind is weak enough to follow the harasser's demand.

So, some instances(like friends.nico) prefer the old way. If you want to delete your account, contact the admin and asked for it.

EzoeRyou:
masarakki also said we should have a account suspend feature which hide the account and all toots(rather and complete deletion), harassment victims who follow the demand can later revive the account.

app/views/auth/registrations/edit.html.haml
```diff
@@ -14,5 +14,6 @@
 
 %hr/
```
Gargron: Include this hr in the if

Gargron (changes requested):

masarakki:
updated.

Gargron (approved):

mistermantas:
No, no, no, this is stupid, because:

1. this makes switching instances harder since you can't just delete your old account
2. this goes against the want for privacy—cleaning everything off your account can be useful in many circumstances
3. and if someone deleted their account because of harassment, they can recreate it.
4. plus, this is just *another* thing to look out for when searching instances...

(If you're stupid enough to ignore the warnings though, and do delete your account unwillingly, that's your own problem, I think.) Maybe adding a 7-14 day restore period is a better idea?

Gargron:
@mistermantas Because currently account deletions are instant *and* bothersome to restore manually, I have accepted this. *Also* there are cases where you genuinely don't want to enable account deletions, for example if you are sourcing the accounts from a different software, e. g. LDAP or some other tight integration.

PR #3853, opened 2017-06-19
Do not call setState from unmounted component

alpaca-tc:
Stop an executing task by `scheduleIdleTask` if the component already unmounted.

Fixes the following warning.

```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Status component.
```


ykzts (approved):

sorin-davidoi:
Would be nice if `<Status />` cancels the async tasks that it spawned when it unmounts, but this works as well :+1:.

PR #3729, opened 2017-06-13
Upgrade Webpacker to version 2.0

ykzts:
- [CHANGELOG](https://github.com/rails/webpacker/blob/v2.0.0/CHANGELOG.md#20---2017-05-24) ([diff](https://github.com/rails/webpacker/compare/v1.2.0...v2.0.0))

Webpacker v2.0 has been released 2017-05-24. A major change has been made to the configuration files.

PR #3854, opened 2017-06-19
Re-add disableHostCheck (regression #3729)

ykzts:
ref #2790


unarist (approved):
I've checked this patch makes it work. Good!

Note: This affects which hosts can load resources from the dev server. Without this option, `127.0.0.1` and `localhost` with any port is allowed, but other hosts like `mastodon.dev` or `192.168.0.2` are not allowed.

PR #3855, opened 2017-06-19
Clicking on the CW text should expand the status

### LINKED ISSUE #3826: Clicking content warning should expand toot
It seems weird that you have to open the content warning and click on the text to expand the toot, because inside the expanded toot you have to click again to show the content.

unarist:
fixes #3826 

Note that this doesn't covers  `button > span > span`, `a > span > span` and other case. Mastodon won't generate these structure, but we don't know for other OStatus servers. So I think below code may be needed:

```
let element = e.target;
do {
  if (element.localName === 'button' || element.localName === 'a') {
    return;
  }
  element = e.parentNode;
} while(element !== undefined && element !== this.node);
```

but it doesn't look efficient. So I only did quick fix on this patch.


beatrix-bitrot (approved):
looks ok

i'd like to test this on glitch.social once it's in master

beatrix-bitrot:
works well :3

PR #3863, opened 2017-06-20
Don't attach IntersectionObserver for wrapped statuses

unarist:
This fixes a bug that sometimes boosted statuses being hidden on scrolling (since #3851).

Previously, we've attached IntersectionObserver twice for boosted statuses: wrapper Status and wrapped Status. This will call intersection handler twice, so this may results race condition...probably.

cc @nolanlawson


Gargron (approved):

nolanlawson:
Nice catch, this looks correct to me. :+1:

PR #3867, opened 2017-06-20
Fix RTL detection on Ruby side

### LINKED ISSUE #3803: Implementation of rtl? method is incorrect
[1] [rtl?](https://github.com/tootsuite/mastodon/blob/master/app/helpers/stream_entries_helper.rb#L54) method has the following line:

```rb
 rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
```

The vertical lines in `[ ]` are superfluous.
Because of it, vertical lines are counted as a RTL-caracter; so, the toot `a|b`, for example,  will be treated as RTL.

[2] `match` method in the above line will find only the first RTL-word in a toot.
So amount of RTL-characters in a toot will not be evaluated correctly.

unarist:
This fixes #3803, i.e. below bugs:

* pipe characters being counted as RTL character
* only first word being checked

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -53,11 +53,11 @@ module StreamEntriesHelper
 
   def rtl?(text)
     text = simplified_text(text)
-    rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
+    rtl_words = text.scan(/[\p{Hebrew}\p{Arabic}\p{Syriac}\p{Thaana}\p{Nko}]+/m)
 
-    if rtl_characters.present?
+    if rtl_words.present?
       total_size = text.size.to_f
-      rtl_size(rtl_characters.to_a) / total_size > 0.3
+      rtl_size(rtl_words) / total_size > 0.3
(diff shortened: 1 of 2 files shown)


app/helpers/stream_entries_helper.rb
```diff
@@ -53,11 +53,11 @@ def rtl_status?(status)
 
   def rtl?(text)
     text = simplified_text(text)
-    rtl_characters = /[\p{Hebrew}|\p{Arabic}|\p{Syriac}|\p{Thaana}|\p{Nko}]+/m.match(text)
```
clworld: Oh...:sweat_smile:

clworld (approved):
Looks good to me.

app/helpers/stream_entries_helper.rb
Gargron: @clworld What?

Gargron: I mean, is everything okay with this PR?

clworld: Yes.  Everything is okay. I said 'Oh...😅' because old code has '|' in '[]' in Regexp.

PR #3870, opened 2017-06-20
Fix conversations (fixes #3869)

### LINKED ISSUE #3869: Conversation tracking broken across instances
Conversation tracking is broken across instances (replies between users of an instance will be tied to the same `Conversation`, but any reply from a remote user will be assigned a new `Conversation`).

There are two reasons for this:
- First, there is code to track conversation based on explicit conversation URIs, but no conversation is created for a remote URI (`find_or_create_conversation` never *creates* a conversation)
- Second, replies processed from remote instances are `create`ed without information on their parent toots (`thread`), so the validation code will *always* create a *new* conversation

* * * *

ClearlyClaire:
Fix conversations handling by actually creating a `Conversation` object corresponding to unknown remote conversation URIs, and try to keep the same conversation across replies in case of missing `ostatus:conversation` URI. The second part is actually a bit fragile as it will fail for asynchronously fetched parent toots.

### DIFF
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -174,7 +174,7 @@ class ProcessFeedService < BaseService
         return Conversation.find_by(id: local_id)
       end
 
-      Conversation.find_by(uri: uri)
+      Conversation.find_by(uri: uri) || Conversation.create!(uri: uri)
     end
 
     def find_status(uri)
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -66,8 +66,6 @@ class ProcessFeedService < BaseService
           status.reblog = original_status.reblog? ? original_status.reblog : original_status
         end
 
-        status.thread = find_status(thread(@xml).first) if thread?(@xml)
-
         status.save!
       end
 
@@ -155,7 +153,8 @@ class ProcessFeedService < BaseService
         reply: thread?(entry),
         language: content_language(entry),
         visibility: visibility_scope(entry),
-        conversation: find_or_create_conversation(entry)
+        conversation: find_or_create_conversation(entry),
+        thread: thread?(entry) ? find_status(thread(entry).first) : nil
       )
 
       mentions_from_xml(status, entry)


app/services/process_feed_service.rb
```diff
@@ -155,7 +153,8 @@ def status_from_xml(entry)
         reply: thread?(entry),
         language: content_language(entry),
         visibility: visibility_scope(entry),
-        conversation: find_or_create_conversation(entry)
+        conversation: find_or_create_conversation(entry),
+        thread: thread?(entry) ? find_status(thread(entry).first) : nil
```
Gargron: No, do not do this please. First of all, new conversation wouldn't be created in validation callback if conversation is already set (line above). Secondly, the reason this is set in the other place is because of the transaction block and what find_status does.

Gargron (changes requested):

app/services/process_feed_service.rb
ThibG: If the remote software does not set “ostatus:conversations”, the result of `find_or_create_conversation` will be `nil` and a new `Conversation` will be created, that's why I added the second line: the validation will use the upstream status' if it exists. As for `find_status`, why wouldn't it be safe to use? It was already in the transaction block. The thing that should not be in the transaction block is `ThreadResolveWorker.perform_async`, which is outside the block.

app/services/process_feed_service.rb
```diff
@@ -174,7 +173,9 @@ def find_or_create_conversation(xml)
         return Conversation.find_by(id: local_id)
       end
 
-      Conversation.find_by(uri: uri)
+      conversation = Conversation.find_by(uri: uri)
+      return conversation unless conversation.nil?
+      Conversation.create!(uri: uri)
```
nightpool: i'm not going to comment on the correctness of this code, but the ruby style for this is Conversation.find_by || Conversation.create!

ThibG: Right, better this way. Done!

app/services/process_feed_service.rb
Gargron: Alright, I must have misremembered the conversation from your last PR about thread resolving. Seems fine then.

Gargron (approved):

PR #3879, opened 2017-06-20
Lazy load components

sorin-davidoi:
Roadmap:
- [x] Display progress bar while fetching bundles
- [x] Routes
   - [x] Lazy load ([react-router guide](https://reacttraining.com/react-router/web/guides/code-splitting))
   - [x] Loading state
      - [x] Empty column
      - [x] Move styles to SCSS
   - [x] Replay request (if bundle could not be fetched)
- [x] Columns:
  - [x] Compose (+ reducer)
  - [x] Home
  - [x] Notifications (+ reducer)
  - [x] Public timeline
  - [x] Community timeline
  - [x] Hashtag timeline
  - [x] Loading state - empty column
- [x] Modals
  - [x] Components:
    - [x] Media
    - [x] Onboarding
    - [x] Video
    - [x] Boost
    - [x] Confirm
    - [x] Report
  - [x] Replay request (if bundle could not be fetched)
  - [x] Move styles to SCSS
- [x] Do not show the waiting state if the bundle is fetched fast enough (`~200ms`)
- [x] Optimize bundles - perhaps tweak `minChunks`?
- [x] Measure size improvement

Bundles: https://gist.github.com/sorin-davidoi/07def6eb30df015fbfc8ced9fa36babd


sorin-davidoi:
When a modal bundle could not be fetched, the following is displayed, allowing users to retry the download
image

Same thing with route bundles
image

sorin-davidoi:
This reduces the size of `application.js` by 27% (673kB -> 485kb), while creating ~25 extra bundles. The progress bar is being used to show the download progress of each bundle. While the bundle is fetched, it displays a loading state in accordance with the content. If the fetch fails, an error component is shown, which provides a "refresh" button, allowing the user to replay the download.

nolanlawson:
Nice work! Based on the Webpack Bundle Analyzer output, though, it does appear that there are lots of common modules that are repeated throughout the async chunks – `status.js`, `video_player.js`, etc. I suppose WebpackCommonsChunkPlugin doesn't include those if they're included in async chunks? It seems useful to put those in `common.js`, given that there will probably be lots of users who load at least two of home/localTL/federatedTL/accountTL/hashtagTL/etc.

nightpool:
I'm not sure I see the practical benefit of this—won't it just increase latency on mobile due to the extra round trip requests?

Is there a way to measure whether the decreased bundle size will actually reflect real world usage?

sorin-davidoi:
It should decrease time to first paint (less JavaScript to parse, which is the main bottleneck on mobile), as well as memory usage. Don't think we are currently measuring those?

sorin-davidoi:
As a side note, it would be nice to set up Lighthouse integration, to measure these metrics for each PR.

sorin-davidoi:
Unscientific findings (running Lighthouse against current master and this branch):
 - Performance score: +3 (51 -> 54)
 - Render blocking scripts: -200ms (2070ms -> 1820ms)
 - First paint: -140ms (7060ms -> 6950ms)
 - First interactive: -600ms (7570ms -> 6950ms)

I would prefer to take the average of a few tests but I don't have the time to do it (is there a tool for it?).

nolanlawson:
> won't it just increase latency on poor connections due to the extra round trip requests?

That can be fixed by adding `<link rel=preload>`s for each of the chunks that we expect are very likely to get loaded by the user. I think this would be a smart addition; @sorin-davidoi you can add it to `application.html.haml` like so:

```haml
%link{:rel => "preload", :as => "script", :src => (asset_pack_path 'application.js')}/
```

Also for your Lighthouse score, what page did you test against? There are a few I think we want to prioritize:

* `/web/getting-started` – main home
* `/web/timelines/home` – mobile PWA home (defined in `manifest.json`)

I also imagine there are people who bookmark local/federated/home timelines, so those would be worthwhile to preload as well.

Overall, I believe it is worthwhile to only load those routes that the user explicitly requested, since it limits overhead of JS parse/exec (and also memory). But avoiding extra network roundtrips is important, and also I'd like to see some deduplication of the shared modules across async chunks.

sorin-davidoi:
I tested the local timeline (mobile UI). `<link rel=preload>` sounds nice, will try to add them! :+1:

sorin-davidoi:
I think we could also lazy-load some reducers (see https://stackoverflow.com/questions/32968016/how-to-dynamically-load-reducers-for-code-splitting-in-a-redux-application), but that is a can of worms I don't want to open for now.

beatrix-bitrot:
Is this ready for review?

sorin-davidoi:
Just a few final touches to do.

sorin-davidoi:
Nah, this is it for this PR. @beatrix-bitrot

sorin-davidoi:
I'd say this is ready for review. Sometimes there is a crash when loading the HomeTimeline (hard to reproduce) which I need to look into, but otherwise, it looks alright.

sorin-davidoi:
I'd say it is now ready for review.

beatrix-bitrot:
nice, because it's a big PR and not my area of expertise I will test this branch on glitch.social to make my review

app/javascript/mastodon/features/ui/index.js
```diff
@@ -13,69 +11,41 @@ import { debounce } from 'lodash';
...
+} from './util/async-components';
+
+const Notifications = () => AsyncNotifications().then(component => {
+  store.dispatch(refreshNotifications());
+  return component;
+});
+
+// Dummy import, to make sure that <Status /> ends up in the application bundle.
+// Without this it ends up in ~8 very commonly used bundles.
+import '../../components/status';
```
nolanlawson: nice fix

app/views/layouts/application.html.haml
```diff
@@ -20,6 +20,23 @@
...
+    = javascript_pack_tag 'reducers/search', integrity: true, crossorigin: 'anonymous', rel: 'preload'
+
+    = javascript_pack_tag 'features/home_timeline', integrity: true, crossorigin: 'anonymous', rel: 'preload'
+
+    = javascript_pack_tag 'features/notifications', integrity: true, crossorigin: 'anonymous', rel: 'preload'
+    = javascript_pack_tag 'reducers/notifications', integrity: true, crossorigin: 'anonymous', rel: 'preload'
+
+    = javascript_pack_tag 'features/community_timeline', integrity: true, crossorigin: 'anonymous', rel: 'preload'
+
+    = javascript_pack_tag 'features/public_timeline', integrity: true, crossorigin: 'anonymous', rel: 'preload'
```
nolanlawson: These should have `as="script"` as well; it's explained what the browser does with this extra information [here](https://developers.google.com/web/updates/2016/03/link-rel-preload).

nolanlawson (changes requested):
Codewise this looks great to me; I just have a few small suggestions. This is also deployed at malfunctioniong.technology (with service worker as well) for those who want to take a look.

nolanlawson:
In terms of bundle sizes this also looks great. I rebuilt and put a Webpack Bundle Analyzer view [here](http://bl.ocks.org/nolanlawson/raw/c8884382fb17c4cb5e733b8190e900c9/) for those who want to see. (Protip: make a Gist with an index.html and bl.ocks.org can host it for you. :smiley:)

Gargron:
@nolanlawson Waiting on you to check the changes.

nolanlawson (approved):
This looks great to me, and is live on malfunctioning.technology to test. (You may need to hard-refresh due to ServiceWorker.)

We may want to warn admins when we ship this that they will probably want to enable HTTP/2 since we are now preloading 13 JS files, which will be slow on HTTP/1 due to the 6-connections-per-origin limit. We may also want to tweak the preloading to limit it to only those components that are guaranteed to be loaded (e.g. "getting started" will only be loaded for first-time users; maybe we can detect that somehow with cookies?).

But those are minor concerns; overall this looks awesome to me and I'm :+1: to merge. I especially love the new `async-components.js` so we can keep all of those components in one place. Great work!

nolanlawson:
BTW http2 is recommended [in the tootsuite/documentation nginx config](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Production-guide.md#nginx) so hopefully most admins already have it.

unarist:
Component which is loaded before also seem to be lazy loaded.

image

I think it can be avoided by caching result of `import()`. (sample: https://github.com/unarist/mastodon/commit/4cf1fde8a2834cc42a2daa46da4d4d3270fef1ba)

image

Pros: avoid flashing of loading view, slightly reduces total time
Cons: single script execution will be longer, i.e. blocks browser longer


How about this?

sorin-davidoi:
Sounds like a good idea :+1:

nolanlawson:
Yup sounds like a good tradeoff!

nolanlawson:
@Gargron I propose we merge this PR along with https://github.com/unarist/mastodon/commit/4cf1fde8a2834cc42a2daa46da4d4d3270fef1ba; it should be a big win for the initial bundle size and I also really like how it cleans up the async route logic. :)

Gargron:
@sorin-davidoi Please guide me through where the "loading" and "error" views are defined and how to get to see them without a slow connection so I can make sure I like how they look visually.

sorin-davidoi:
@Gargron Open the Network tab in the Chrome development tools and tick "Disable cache".

For columns:
 - Loading state - Load a timeline, throttle the connection to GPRS, click on a toot.
 - Error state - Load a timeline, tick the "Offline" checkbox, click on a toot. After the error state is shown you can uncheck "Offline" and click retry, the normal page should appear.

For modals:
 - Loading state - throttle the connection to GPRS and open a media modal
 - Error state - tick "Offline" and open a media modal - same behavior with retry

PR #3883, opened 2017-06-21
Don't attach IntersectionObserver for wrapped statuses (again)

unarist:
(This patch has been merged as bugfix at #3863 and reverted, but still valuable as improvement)

Previously, we've attached IntersectionObserver twice for boosted statuses: wrapper Status and wrapped Status. but wrapped Status don't need to manage intersection and visibility by itself, because it's a part of wrapper Status.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -154,7 +154,10 @@ class Status extends ImmutablePureComponent {
   render () {
     let media = null;
     let statusAvatar;
-    const { status, account, ...other } = this.props;
+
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
I'm really sorry i reverted your fix last night. I was trying to revert the original pr through github ui, but it wouldn't let me because the same files have been modified since. So i tried to revert everything that touched those files. That didn't help so in the end i reverted just by modifying the code again. Thank you for resubmitting and sorry for the double trouble

PR #3887, opened 2017-06-21
Optimize Status.as_home_timeline

akihikodaki:
This solves a performance regression introduced by commit bab5a18232a163b0c3c6a245f7f95d50d7022b36 and discussed at #3842. The below is a new query. Note that the time does not help because the database is small and it has cached by previous `EXPLAIN ANALYZE` queries.

```
> EXPLAIN ANALYZE SELECT  "statuses".* FROM "statuses" LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id" LEFT OUTER JOIN "follows" ON "follows"."target_account_id" = "accounts"."id" LEFT OUTER JOIN "accounts" "followers_accounts" ON "followers_accounts"."id" = "follows"."account_id" WHERE ((("statuses"."visibility" != 3) OR "statuses"."id" IN (SELECT  "mentions"."status_id" FROM "mentions" WHERE "mentions"."account_id" = 1 LIMIT 100)) AND "follows"."account_id" = 1 OR "statuses"."account_id" = 1) GROUP BY "statuses"."id" ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                                     QUERY PLAN                                                                                     
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=141.41..475.33 rows=100 width=438) (actual time=0.092..1.530 rows=100 loops=1)
   ->  Group  (cost=141.41..144664.59 rows=43281 width=438) (actual time=0.091..1.513 rows=100 loops=1)
         ->  Nested Loop Left Join  (cost=141.41..144556.39 rows=43281 width=438) (actual time=0.089..1.413 rows=100 loops=1)
               Filter: ((((statuses.visibility <> 3) OR (hashed SubPlan 1)) AND (follows.account_id = 1)) OR (statuses.account_id = 1))
               Rows Removed by Filter: 26
               ->  Nested Loop Left Join  (cost=0.00..28181.10 rows=70956 width=442) (actual time=0.053..0.536 rows=118 loops=1)
                     ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.00..6452.75 rows=70956 width=438) (actual time=0.024..0.084 rows=118 loops=1)
                     ->  Index Only Scan using accounts_pkey on accounts  (cost=0.00..0.30 rows=1 width=4) (actual time=0.003..0.003 rows=1 loops=118)
                           Index Cond: (id = statuses.account_id)
                           Heap Fetches: 118
[...]

PR #3892, opened 2017-06-21
Fix regression from #3842

Gargron:
Simplify the query by omitting all direct statuses. Private statuses
are allowed because they are from accounts we are following (so
by definition)

Resolves #3887 (alternative)

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -131,15 +131,7 @@ class Status < ApplicationRecord
     end
 
     def as_home_timeline(account)
-      # 'references' is a workaround for the following issue:
-      # Inconsistent results with #or in ActiveRecord::Relation with respect to documentation Issue #24055 rails/rails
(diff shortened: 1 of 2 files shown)


nightpool:
Do we think this is related to the 500 errors on mastodon.social? I thought this code path was only for users that didn't have a redis timeline already, so it doesn't explain the problems established/active users are having

Gargron:
@nightpool The long running queries that do happen take up resources that other requests cannot get.

Gargron:
@nightpool I am currently deploying this PR to mastodon.social to see if that resolves the problem.

nightpool:
@Gargron +1 just wanted to make sure we were addressing the right issue

nightpool (approved):
this seems preferable to #3887—it's subtly incorrect, but not in a way that's super noticeable, and doesn't affect UX

PR #3895, opened 2017-06-22
Add index statuses on account_id and id

akihikodaki:
A comment in this change says:
> Statuses queried by `account_id` are often sorted by `id`. Querying statuses of an account to show them in his status page is one of the most significant examples.
> Add this index to improve the performance in such cases.

The following query is similar to those queried when responding with pages of accounts.

Before:
```
> EXPLAIN ANALYZE SELECT  "statuses".* FROM "statuses" WHERE "statuses"."account_id" = 1 AND "statuses"."visibility" IN (0, 1) ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                   QUERY PLAN                                                                    
-------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.29..441.07 rows=100 width=432) (actual time=0.043..3.608 rows=100 loops=1)
   ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.29..5179.41 rows=1175 width=432) (actual time=0.041..3.582 rows=100 loops=1)
         Filter: ((visibility = ANY ('{0,1}'::integer[])) AND (account_id = 1))
         Rows Removed by Filter: 4407
 Planning time: 1.007 ms
 Execution time: 3.674 ms
(6 rows)
```

After:
```
> EXPLAIN ANALYZE SELECT  "statuses".* FROM "statuses" WHERE "statuses"."account_id" = 1 AND "statuses"."visibility" IN (0, 1) ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                            QUERY PLAN                                                                
             
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.41..314.84 rows=100 width=432) (actual time=0.375..0.695 rows=100 loops=1)
   ->  Index Scan Backward using index_statuses_on_account_id_id on statuses  (cost=0.41..3694.90 rows=1175 width=432) (actual time=0.373..0.669 rows=100 loops=1)
         Index Cond: (account_id = 1)
         Filter: (visibility = ANY ('{0,1}'::integer[]))
         Rows Removed by Filter: 2
 Planning time: 0.361 ms
 Execution time: 0.791 ms
(7 rows)
```

Look at _Filter_ and _Rows Removed by Filter_. The execution time may not be accurate because I have not vacummed the database and not rebooted the server between those two tests.

### DIFF
--- a/db/migrate/20170610000000_add_statuses_index_on_account_id_id.rb
+++ b/db/migrate/20170610000000_add_statuses_index_on_account_id_id.rb
@@ -1,0 +1,13 @@
+class AddStatusesIndexOnAccountIdId < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    # Statuses queried by account_id are often sorted by id. Querying statuses
+    # of an account to show them in his status page is one of the most
+    # significant examples.
+    # Add this index to improve the performance in such cases.
+    add_index 'statuses', ['account_id', 'id'], algorithm: :concurrently, name: 'index_statuses_on_account_id_id'
+
+    remove_index 'statuses', algorithm: :concurrently, column: 'account_id', name: 'index_statuses_on_account_id'
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170609145826) do
+ActiveRecord::Schema.define(version: 20170610000000) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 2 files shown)


nightpool (dismissed):
LGTM although I'm worried about the performance cost of constructing this index during a migration, especially with the amount of statuses the bigger instances have.

Can you change this so it creates the index concurrently? You'll also have to use `disable_ddl_transaction!` because postgres index can't be created during transactions.

In this case you won't want to remove the original index until after the concurrent one is created, so maybe it should wait for a followup release.

akihikodaki:
301c2857b4bf4fff32a8302b77151d52be0e5c99 runs migration in the concurrent way.

> In this case you won't want to remove the original index until after the concurrent one is created, so maybe it should wait for a followup release.

If I recall correctly, queries in a migration runs in a transaction, so it would wait for the previous query finishes.

nightpool:
@akihikodaki right, but concurrently created queries *can't* run in a transaction. That's what the disable_ddl_transaction does.

akihikodaki:
@nightpool Ah, I forgot that I did `disable_ddl_transaction!`. But still queries in a migration would be queued serially in a connection and `CREATE INDEX CONCURRENTLY` would block in the connection, I think.
`CREATE INDEX CONCURRENTLY` does not lock the database, but does not it block the connection?

nightpool:
@akihikodaki uh, I'm not actually 100% sure. I thought that it did, but I might be wrong.

nightpool:
yeah, nevermind that was just me being stupid. Of course it waits for the index to finish building before returning.

akihikodaki:
That is fine. Could you approve or tell me other concerns if any?

nightpool:
I dismissed my review, and I think the PR is good, but I'm a very new contributor and this is pretty sensitive code (could break upgrades entirely for people) so I'm not comfortable approving it without someone who has more experience with the upgrade process and handling admin concerns looking at it.

beatrix-bitrot (approved):
read the discussion, seems ok

even the largest instances have only ~10M statuses these IDs are ints AFAIK so it should be fine i think

PR #3897, opened 2017-06-22
fixed vagrantfile

LottieVixen:
added rvm install ruby


ykzts (approved):

Gargron (approved):

abcang:
Originally, ruby is installed with the following command, but seems to fail due to bug of rvm.

 ```
curl -sSL https://get.rvm.io | bash -s stable --ruby=$RUBY_VERSION
```

https://github.com/rvm/rvm/issues/4068

PR #3908, opened 2017-06-23
Experimenting with Preact

sorin-davidoi:
I've opened this pull request to **discuss** the tradeoffs that we would get by switching to [Preact](https://preactjs.com/).

To my surprise, only minor changes were needed to make the codebase compatible (we are using `preact-compat`, but I still did not expect the transition to go so smoothly). These are:
 - When mapping over an Immutable list, we need to call `.toArray()` before we render the items
 - The `ref` callback is [not guaranteed to be called after the element is mounted](https://github.com/developit/preact/issues/477#issuecomment-269718643) - this seems to be the case for React as well, so we might be relying on undocumented behavior right now

The main advantage is that we reduce the size of `application.js` by 18% (from 672kB to 552 KB). React is no longer the biggest dependency (Immutable takes its place). If #3879 gets merged, `application.js` will be 387kb.

The only glitch I could find is that when toggling the Emoji picker, the styles of the textarea flash in a weird way for a second or so.

What do you think? If we manage to solve the issue(s), should we switch? Should we also investigate [Inferno](https://www.infernojs.org/)? Should we wait for React 16?

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -168,7 +168,7 @@ class MediaGallery extends React.PureComponent {
       );
     } else {
       const size = media.take(4).size;
-      children = media.take(4).map((attachment, i) => <Item key={attachment.get('id')} onClick={this.handleClick} attachment={attachment} autoPlayGif={this.props.autoPlayGif} index={i} size={size} />);
+      children = media.take(4).map((attachment, i) => <Item key={attachment.get('id')} onClick={this.handleClick} attachment={attachment} autoPlayGif={this.props.autoPlayGif} index={i} size={size} />).toArray();
     }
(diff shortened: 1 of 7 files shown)


nolanlawson:
I'm in favor of Preact both for 1) performance, and 2) not having to worry about the Facebook patents clause. However it may be prudent to wait for React v16 since I know they've been working on some huge perf improvements, both for bundle size (Rollup) and runtime perf (Fiber).

If there are any glitches we should definitely fix that, and I also think it may make sense to only enable Preact in production, since the dev tools for React are probably better than Preact's. Also I've discussed this with @Gargron before, and last time we talked he was leaning more towards sticking with React.

sorin-davidoi:
Will wait and see how React Fiber turns out.

PR #3909, opened 2017-06-23
more free pgconfig by .env

amazedkoumei:
It fixes websocket error on dev-env caused by wrong DB_USER.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,6 +7,10 @@ default: &default
 development:
   <<: *default
   database: mastodon_development
+  username: <%= ENV['DB_USER'] %>
+  password: <%= ENV['DB_PASS'] %>
+  host: <%= ENV['DB_HOST'] %>
+  port: <%= ENV['DB_PORT'] %>
 
 # Warning: The database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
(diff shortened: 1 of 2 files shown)


Gargron (changes requested):
👍 on the env variables, I know I said no to this in the past but honestly, I see no harm in this. But the default values are wrong. For example, my default postgres on Ubuntu 16.04 uses a socket instead of localhost/port, and uses my logged-in user instead of "mastodon". I think the pg driver knows this, so it's important to not overwrite those default values with our own ones.

amazedkoumei:
Thank you for your review.
I have fixed wrong default values, and understand your thoughts.
> it's important to not overwrite those default values with our own ones.

amazedkoumei:
I edit database.yml to behave as sama as streaming server. 
I confirmed that pg gem behaves as in accordance with https://www.postgresql.org/docs/9.2/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS .

and remove "for streaming-server" from this pull request's title.

Gargron (approved):

unarist:
Hmm, should we have `TEST_DB_*` variables separate from `DB_*` variables?

We had used same values for both environments previously, so using same variable is reasonable I think. or fallback to `DB_*` variables would be useful.

Gargron:
@unarist Yes, I think you are right.

PR #3912, opened 2017-06-23
Upgrade to Webpack 3 with module concatenation

nolanlawson:
This reduces the size of `application.js` from 670196 bytes to 655404 bytes (14.8kB saved). The other bundles are roughly the same size as before.

You can read more about how module concatenation (aka scope hoisting) works [here](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b). Essentially this will just take all of our modules inside of `application.js` and squash them into a single function scope instead of having multiple `function(){}` wrappers around each one. It de-penalizes us for having multiple small modules.

Interestingly this actually increases the gzip size of `application.js` slightly (171181 to 173646, 2.46kB) but I think the tradeoff is worth it given that the ungzipped size is what the browser has to parse, and less JS to parse is always good.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -12,6 +12,7 @@ module.exports = merge(sharedConfig, {
   stats: 'normal',
 
   plugins: [
+    new webpack.optimize.ModuleConcatenationPlugin(),
     new webpack.optimize.UglifyJsPlugin({
       sourceMap: true,
       mangle: true,
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "tiny-queue": "^0.2.1",
     "uuid": "^3.0.1",
     "uws": "^0.14.5",
-    "webpack": "^2.5.1",
+    "webpack": "^3.0.0",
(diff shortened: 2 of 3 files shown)


sorin-davidoi (approved):
Nice! Love it that we have all our dependencies up-to-date and can take advantage of new features almost right away!

Gargron (approved):

sorin-davidoi:
Appears to break webpack-dev-server: https://github.com/webpack/webpack-dev-server/issues/947

nolanlawson:
Hm should we roll back or can we just fork webpack-dev-server until they fix it? From the discussion, sounds like it's a fairly easy fix on their end.

sorin-davidoi:
I'd don't mind fixing it locally but other people might run into trouble. There is a PR open to fix it https://github.com/webpack/webpack-dev-server/pull/946, hope they merge it soon.

nolanlawson:
In that case we can just change our dependency in package.json to `"webpack-dev-server": "lencioni/webpack-dev-server#patch-1"` while we wait for a fix.

PR #3913, opened 2017-06-23
Fix errorMiddleware to prevent "TypeError: res.writeHead is not a function"

ikuradon:
error handling middleware require fourth arguments.
reference: http://expressjs.com/en/guide/error-handling.html

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -242,7 +242,7 @@ const startWorker = (workerId) => {
     accountFromRequest(req, next);
   };
 
-  const errorMiddleware = (err, req, res, next) => {
+  const errorMiddleware = (err, req, res, next) => {  // eslint-disable-line no-unused-vars
     log.error(req.requestId, err.toString());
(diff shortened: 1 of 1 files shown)


Gargron:
I don't understand how this would fix res. Also, the next variable is not used.

ikuradon:
Looks like if express.js handle fourth arguments, works as "error handling middleware" and supply err argument.
I don't know how to suppress unused variable, but this argument is necessary.

streaming/index.js
```diff
@@ -242,7 +242,7 @@ const startWorker = (workerId) => {
     accountFromRequest(req, next);
   };
 
-  const errorMiddleware = (err, req, res) => {
+  const errorMiddleware = (err, req, res, next) => {
```
Gargron: Add `// eslint-disable-line no-unused-vars` to this line

Gargron (changes requested):

Gargron (approved):

PR #3914, opened 2017-06-23
Apply babel to react-intl to remove prop-types

nolanlawson:
The `prop-types` package isn't actually being excluded from our build because it's used by `react-intl`. This fixes that, by applying Babel to `react-intl`.

The size of `application.js` is reduced from 670196 bytes to 655841 bytes (14.4kB removed).

### DIFF
--- a/config/webpack/loaders/babel.js
+++ b/config/webpack/loaders/babel.js
@@ -1,6 +1,7 @@
 module.exports = {
   test: /\.js$/,
-  exclude: /node_modules/,
+  // include react-intl because transform-react-remove-prop-types needs to apply to it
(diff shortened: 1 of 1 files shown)


nolanlawson:
This is live at malfunctioning.technology for those who want to test it. I don't observe anything broken by this.

Gargron (approved):

sorin-davidoi:
I'm getting a bunch of these when building for production:
```
WARNING in ./tmp/packs/locale_zh-TW.js
8:44-54 "export 'default' (imported as 'localeData') was not found in 'react-intl/locale-data/zh.js'
```

And this when I am logged out. Although (to me at least) seems to work fine.
image

Issue #3255, opened 2017-05-23
Endpoint for all direct messages to and from a given user, or all messages in general


DagAgren:
I'm working on a client, and I'd love to add a chat client like view for DMs. This would be greatly simplified if there was an endpoint where I could request all direct messages from and to a specific user.

Perhaps there could also be options to either limit it to just direct messages or all statuses referring to that user, so that you could mix in public messages too if you preferred.

This could perhaps be either a couple of options for `/api/v1/accounts/:id/statuses`, such as "include_me", "to_me_only" and "direct_only", or a separate endpoint like `/api/v1/accounts/:id/conversation`.

* * * *

- [x] I searched or browsed the repo's other issues to ensure this is not a duplicate (as far as I can tell).

Gargron:
Thank you for opening an issue for this 👍

PR #3920, opened 2017-06-24
Include direct statuses in Status.as_home_timeline

akihikodaki:
The feature was implemented in commit bab5a18232a163b0c3c6a245f7f95d50d7022b36 for consistency with `FanOutWriteService` and removed in commit 0190aac240fd804180a56b5fe174526a9e4c3f6d due to performance problem.
This restores it with an optimized query.

The below is the actual query and the query plan made for our instance, o.kagucho.net. Note that the instance is so small and would result in different query plan for bigger instances. For example, SubPlan 2 uses `SeqScan`, but it is inefficient and not likely to be used for a large table.
```
> EXPLAIN ANALYZE SELECT "statuses".* FROM "statuses" WHERE ((("statuses"."visibility" != 3) OR "statuses"."id" IN (SELECT "mentions"."status_id" FROM "mentions" WHERE "mentions"."account_id" = 1)) AND "statuses"."account_id" IN (SELECT "follows"."target_account_id" FROM "follows" WHERE "follows"."account_id" = 1) OR "statuses"."account_id" = 1) ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                     QUERY PLAN                                                                     
----------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=113.31..133.27 rows=100 width=432) (actual time=0.215..0.449 rows=100 loops=1)
   ->  Index Scan Backward using statuses_pkey on statuses  (cost=113.31..5559.42 rows=27288 width=432) (actual time=0.214..0.420 rows=100 loops=1)
         Filter: ((((visibility <> 3) OR (hashed SubPlan 1)) AND (hashed SubPlan 2)) OR (account_id = 1))
         Rows Removed by Filter: 30
         SubPlan 1
           ->  Bitmap Heap Scan on mentions  (cost=4.87..107.69 rows=75 width=8) (never executed)
                 Recheck Cond: (account_id = 1)
                 ->  Bitmap Index Scan on index_mentions_on_account_id_and_status_id  (cost=0.00..4.85 rows=75 width=0) (never executed)
                       Index Cond: (account_id = 1)
         SubPlan 2
           ->  Seq Scan on follows  (cost=0.00..4.83 rows=127 width=4) (actual time=0.015..0.099 rows=127 loops=1)
                 Filter: (account_id = 1)
                 Rows Removed by Filter: 99
 Planning time: 0.498 ms
 Execution time: 0.558 ms
(15 rows)
```

Requesting reviews from @Gargron and @abcang, expecting testing the query with their databases.


unarist (dismissed):
I found missing test case, which this implementation will fails...

```
it 'includes statuses mentiong recipient from non-followed' do
  Fabricate(:mention, account: account, status: @not_followed_status)
  expect(@results).to include(@not_followed_status)
end
```

(It's only filter expression problem, and it won't so affects query plan I think.)

app/models/concerns/status_timeline.rb
```diff
@@ -0,0 +1,78 @@
...
+      mention_status_ids = Mention.select(:status_id).where(account: account)
+      if limit.present?
+        mention_status_ids = mention_status_ids.where('status_ids < ?', max_id) if max_id.present?
+        mention_status_ids = mention_status_ids.where('status_ids > ?', since_id) if since_id.present?
+        mention_status_ids = mention_status_ids.limit(limit)
+      end
+
+      statuses = where.not(visibility: :direct)
+        .or(where(id: mention_status_ids))
+        .where(account_id: Follow.select(:target_account_id).where(account: account))
```
unarist: Should two lines above be interchanged?

abcang:
@akihikodaki It is the execution result in Pawoo.

```
> EXPLAIN ANALYZE SELECT "statuses".* FROM "statuses" WHERE ((("statuses"."visibility" != 3) OR "statuses"."id" IN (SELECT "mentions"."status_id" FROM "mentions" WHERE "mentions"."account_id" = 1)) AND "statuses"."account_id" IN (SELECT "follows"."target_account_id" FROM "follows" WHERE "follows"."account_id" = 1) OR "statuses"."account_id" = 1) ORDER BY "statuses"."id" DESC LIMIT 100;
                                                                                   QUERY PLAN

-------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------
 Limit  (cost=812.85..828.93 rows=100 width=333) (actual time=5.663..166.868 rows=100 loops=1)
   ->  Index Scan Backward using statuses_pkey on statuses  (cost=812.85..1790087.25 rows=11123922 width=333) (actual time=5.662..166.778 rows=100 loops=1)
         Filter: ((((visibility <> 3) OR (hashed SubPlan 1)) AND (hashed SubPlan 2)) OR (account_id = 1))
         Rows Removed by Filter: 138467
         SubPlan 1
           ->  Index Only Scan using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..714.83 rows=244 width=8) (actual time=0.045..0.522 r
ows=46 loops=1)
                 Index Cond: (account_id = 1)
                 Heap Fetches: 46
         SubPlan 2
           ->  Index Only Scan using index_follows_on_account_id_and_target_account_id on follows  (cost=0.43..96.78 rows=75 width=4) (actual time=0.074..0.5
54 rows=57 loops=1)
                 Index Cond: (account_id = 1)
                 Heap Fetches: 55
 Planning time: 1.392 ms
 Execution time: 167.039 ms
(14 rows)
```

unarist:
The query performance would depend on:

* Record count of skipped statuses
The query checks condition for every records on `statuses`. If followed users haven't tooted for a long time, the query takes a while to check so many statuses probably.
* Record count from subplans
The query hashes both subplan result. It will increase matching performance, but it needs to fetch all matched ids. The cost for this may reduced by adding single column indexes for `mentions.account_id` and `follows.account_id`.

So the result from @abcang looks a light case. For example, I'm interested in how this query works with ["some users" on Pawoo](https://github.com/tootsuite/mastodon/pull/3887#issuecomment-310166681).

abcang:
@unarist It took about 15 seconds for some accounts.

Gargron:
For what it's worth I am probably going to remove direct statuses from home timelines in FanOutOnWriteService for consistency. Most people don't expect to see them in the home feed anyway.

akihikodaki:
@Gargron Then what kind of an interface could we have to show direct messages _without_ mentions?

Gargron:
@akihikodaki #3255 is one potential interface, although I am not sure how to implement that, at the moment. (Perhaps it could query "conversations" by participants)

akihikodaki:
@Gargron That is a good suggestion. However I'm worrying about the user interface.

akihikodaki:
#4514 is a good alternative of the UI.

PR #3927, opened 2017-06-24
Add notification's time to timeline

pfm-eyesightjp:
timeline
timeline2


app/javascript/mastodon/components/status.js
```diff
@@ -186,6 +186,8 @@ export default class Status extends ImmutablePureComponent {
           <div className='status__prepend'>
             <div className='status__prepend-icon-wrapper'><i className='fa fa-fw fa-retweet status__prepend-icon' /></div>
             <FormattedMessage id='status.reblogged_by' defaultMessage='{name} boosted' values={{ name: <a onClick={this.handleAccountClick} data-id={status.getIn(['account', 'id'])} href={status.getIn(['account', 'url'])} className='status__display-name muted'><strong dangerouslySetInnerHTML={displayNameHTML} /></a> }} />
+            <span className='separator'>-</span>
```
akihikodaki: I think it needs `aria-hidden`.

akihikodaki (changes requested):
Thank you for your contribution. I like this feature. I've added a comment.

akihikodaki (approved):

pfm-eyesightjp:
Thank you for your review.

This PR is based on @matz 's [opinion](https://mstdn.jp/@matz/22956868).
I agree with his opinion. It is certainly necessary for us.

matz:
Thank you for taking time to consider my opinion.

nightpool (changes requested):
Sorry for leaving this PR to linger for so long.

Consensus among the developers was that we **should** expose this information, but right now that area is already somewhat cluttered, and we think adding a title attribute with the relative created_at timestamp is the correct way to proceed here. That way users can still see the information, but it's more of a power-user thing and not as overwhelming, visually, for the casual user. Using a title attribute means users on desktop can see it on hover, and users on mobile could see it on a long-press.

Unfortunately, this was from an informal discussion, and none of us remembered to update this issue with our decision. That really sucks, and I'm sorry that this PR has needlessly been in "limbo" for so long.

pfm-eyesightjp:
thanks for reply.

> We think adding a title attribute with the relative created_at timestamp is the correct way to proceed here.

Where are you supposed to add title attribute?
I want your advice.

roipoussiere:
Note: fix #6210

roipoussiere:
Any news here? I find this PR really useful! It's very disturbing to see only the date of the original post...

Issue #2347, opened 2017-04-23
Web client stayed logged in when using multiple tabs


ineffyble:
**Steps to replicate:**

- Log in to the web client
- Open a new tab
- Log out in the second tab
- Log in to another account in the second tab
- Switch to first tab
- Toot

**Expected:** Error message, or first tab client logs out

**Actual:** Toot is successfully published from the first account

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ashfurrow:
Yikes!

Gargron:
This shouldn't be weird really? Web client uses an OAuth access token. While your page is open, the access token stays the same so you can keep using the API. If you reload the page you won't have access to the access token anymore.

daprice:
This might be worse than previously thought. On iOS Safari I can do the following:

1. Start with a tab logged in to mastodon
2. log out
3. close the tab
4. leave the browser and come back to it later
5. reopen the closed tab from history by holding down the "+" button
6. press the back button to get back to the timeline

…and be able to toot/favorite/boost without having to enter any credentials. This has to do with Safari saving the page state along with browsing history, and I believe Firefox/Chrome do something similar.

As far as I can tell, this means that in order to prevent someone else from gaining complete access to their account, a user on a shared device would have to either (a) reset the browser (i.e. clear history, cookies, and any saved window/tab state) when finished using Mastodon, or (b) only ever use Mastodon with private browsing.

Gargron:
So uh.. I guess we could store the access token in a cookie then? Instead of having it in the HTML.

daprice:
Regardless of where the access token is stored, shouldn't it be invalidated on the server side when the user logs out, per https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Manual_Session_Expiration ?

Issue #588, opened 2017-02-01
Function to log out on all other instances


noiob:
After setting 2FA I noticed that there's no way to tell where I'm logged in from and activating 2FA doesn't end sessions that weren't started using 2FA. I suggest adding a button to log the user out of any other session so people can be sure their account is safe after changing their password/ activating 2FA.

ashfurrow:
Seems related to https://github.com/tootsuite/mastodon/issues/2347 as well.

noiob:
To some extent, this is solved by the 'Authorized Apps' settings, but not for web logins

ashfurrow:
Agreed. Do you think that #2347 is different enough that you'd like to leave this issue open? It's up to you!

noiob:
Yeah, definitely. #2347 is a fairly big security problem while this is just a feature that'd help people keep their accounts safe.

PR #3929, opened 2017-06-24
Add overview of active sessions

Gargron:
image

cc @sorin-davidoi

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -91,8 +91,13 @@ class User < ApplicationRecord
     settings.auto_play_gif
   end
 
-  def activate_session(description, ip)
-    session_activations.activate(SecureRandom.hex, description, ip).session_id
+  def activate_session(request)
+    detection = Browser.new(request.user_agent)
+
+    session_activations.activate(session_id: SecureRandom.hex,
+                                 browser: detection.name,
+                                 platform: detection.platform.id,
+                                 ip: request.ip).session_id
   end
 
   def exclusive_session(id)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -91,8 +91,8 @@ class User < ApplicationRecord
     settings.auto_play_gif
   end
 
-  def activate_session
-    session_activations.activate(SecureRandom.hex).session_id
+  def activate_session(description, ip)
+    session_activations.activate(SecureRandom.hex, description, ip).session_id
   end
 
   def exclusive_session(id)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -92,11 +92,8 @@ class User < ApplicationRecord
   end
 
   def activate_session(request)
-    detection = Browser.new(request.user_agent)
-
     session_activations.activate(session_id: SecureRandom.hex,
-                                 browser: detection.name,
-                                 platform: detection.platform.id,
+                                 user_agent: request.user_agent,
                                  ip: request.ip).session_id
   end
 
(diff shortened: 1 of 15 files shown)


Gargron:
Two sessions-related things remain (but outside of this PR, probably):

- #588 revoke session (but should this actually be a thing? it should ask for a password to confirm probably?)
- bind the access token to the session activation to resolve #2347

nightpool:
revoke session would be nice. re confirmation: I don't know of any current
"log out session" implementation that requires you to re-authenticate.
(google, tumblr, etc don't iirc)

Gargron:
@nightpool I mean, logging out *other* sessions than the one you're in right now. Is it safe without extra authentication?

nightpool:
@gagron right, sorry, that's what I meant too. I would assume it was safe,
because it's just revoking access—you still need to reauthenticate to
change the password or do anything that could lock someone out.

I checked three implementations, Tumblr, Github, and Google, and only
github requires you to reauthenticate before revoking a remote session.

akihikodaki:
If the revocation does not revoke the session which requested the revocation, the revocation could be used to perform DoS attack targeting particular users.
Assume Malory has aquired the session of Alice, but does not have the password nor any other means to activate new sessions. She wants to take over the account of Alice. In the case, Malory would continuosly revoke other sessions to prevent Alice from signing in. Malory can still perform various operations with the account of Alice.

Here, I suggest two solutions for the problem:
* Require authentication before revocation
* Revoke ALL sessions, including the session which requested the revocation

Requiring authentication effectively prevents attackers with sessions from taking over the account.
Revoking all sessions is also possible if an activation of a new session always will require authentication.

app/models/user.rb
```diff
@@ -91,8 +91,13 @@ def setting_auto_play_gif
     settings.auto_play_gif
   end
 
-  def activate_session
-    session_activations.activate(SecureRandom.hex).session_id
+  def activate_session(request)
+    detection = Browser.new(request.user_agent)
+
+    session_activations.activate(session_id: SecureRandom.hex,
```
akihikodaki: Why storing `name` for `browser` and `id` for `platform`?

app/views/auth/registrations/_sessions.html.haml
```diff
@@ -0,0 +1,16 @@
...
+%table.table
+  %thead
+    %tr
+      %th= t 'sessions.browser'
+      %th= t 'sessions.ip'
+      %th= t 'sessions.activity'
+  %tbody
+    - @sessions.each do |session|
+      %tr
+        %td= t 'sessions.description', browser: session.browser, platform: session.platform.capitalize
```
akihikodaki: * `session.browser` is not localized (e.g. "Generic Browser")
* `session.platform` is derived from ID and may not be human reable.

akihikodaki (changes requested):
I have some questions about browser detection so adds some comments.

I also think it is better to keep information about previous sessions. For example, an user may notice there are suspicious operations, but the attacker may have already signed out. In such a case the user could not review sessions.

That also allows to remove `current_sign_in_ip` and `last_sign_in_ip` from `User`.
EDIT: `current_sign_in_at` and `last_sign_in_at` could be removed as well.

app/models/user.rb
Gargron: @akihikodaki platform.name returns an english string like "Generic Linux", while id returns something like "linux", "windows", so it's easier to localize. I thought that browser.name would not contain english strings since most browsers have trademark names like firefox/internet explorer which do not need to be translated, but if as you say it also returns "Generic Browser", then I need to switch that to id too

akihikodaki (approved):

akihikodaki:
This change is ready to merge. I suggested to keep previous sessions and remove session information from `accounts`, but that could be implemented later as enhancements.

PR #3935, opened 2017-06-24
Fix #3910 - Require OTP authentication to disable 2FA

### LINKED ISSUE #3910: Require re-authentication to disable 2FA
Currently, if a user is logged in they can disable 2FA without having to re-authenticate. 

It's considered best practice to force the user to re-authenticate before they can disable a security feature (with either the password, an backup access code, or by using 2FA; the choice is usability v. security).

* * * *

Gargron:
image


akihikodaki (dismissed):
I think it is too dangerous. The user cannot have a recovery code when the user found he lost his recovery code. That is problematic for Mastodon, which does not provide any other methods to bypass 2FA.
We could allow users to recover their account with their email addresses when they lost the 2FA seeds and the recovery code. Enabling 2FA should not prevent from account recovery with email and we could provide another option to disable email recovery if necessary.

Gargron:
@akihikodaki There is always the method of contacting the admin. Consider this scenario:

1. You have 2FA enabled
2. You login on a public PC, but forget to logout (or login on your laptop, go away, somebody sits down on it)
3. The stranger generates new recovery codes (which invalidate the old ones), writes them down
4. The stranger can now login to the 2FA account if he also knows the password

I'm not sure if any other websites allow you to generate backup codes *after* generating them once during 2FA activation.

akihikodaki:
@Gargron I mean an easy method to recover the account without contacting the admin should be implemented _before_ applying this change, or users could be reluctant to use 2FA. Applying this change _later_ is fine.

Gargron:
@akihikodaki I restored that function.

beatrix-bitrot (approved):
added tests, addressed reviewer concerns, seems ok

PR #3937, opened 2017-06-25
Fix react-intl/locale-data import issue on production build (regression from #3914)

unarist:
Webpack seems to fail to import locale-data if those files has been proceed by babel (since #3914).

```
WARNING in ./tmp/packs/locale_ar.js
8:44-54 "export 'default' (imported as 'localeData') was not found in 'react-intl/locale-data/ar.js'

WARNING in ./tmp/packs/locale_bg.js
8:44-54 "export 'default' (imported as 'localeData') was not found in 'react-intl/locale-data/bg.js'

WARNING in ./tmp/packs/locale_ca.js
8:44-54 "export 'default' (imported as 'localeData') was not found in 'react-intl/locale-data/ca.js'

WARNING in ./tmp/packs/locale_de.js
8:44-54 "export 'default' (imported as 'localeData') was not found in 'react-intl/locale-data/de.js'

...
```

and this also breaks applying our translation.

Note that this won't be a problem on English locale, because react-intl includes it as default and works fine without manually added locale-data. Also this issue seems to only occurs on production build (`yarn run build:production`), but I'm not sure about reason.

### DIFF
--- a/config/webpack/loaders/babel.js
+++ b/config/webpack/loaders/babel.js
@@ -1,7 +1,10 @@
 module.exports = {
   test: /\.js$/,
   // include react-intl because transform-react-remove-prop-types needs to apply to it
-  exclude: /node_modules[\/\\](?!react-intl)/,
+  exclude: {
+    test: /node_modules/,
+    exclude: /react-intl[\/\\](?!locale-data)/,
+  },
   loader: 'babel-loader',
   options: {
     forceEnv: process.env.NODE_ENV || 'development',


Gargron (approved):

nolanlawson:
Oops sorry, that's my bad. Prop types are [only removed in production](https://github.com/tootsuite/mastodon/blob/a978b88997169782ac35f416bf88d6afd60edd1e/.babelrc#L37-L46) which is why you only see this bug in production. Thanks for the fix!

nolanlawson:
BTW it appears some prop-types are still appearing in production (see http://mastodon.social/packs/report.html) but maybe it's just not worth worrying about, especially if it's going to break stuff to remove it.

2017-06-30 09_26_25-microsoft edge

unarist:
@nolanlawson BTW, is removing prop-types in this way common manner on node packages? Applying babel to (already compiled) node_modules seems weird...

nolanlawson:
I agree it's weird, but apparently there's no other way to remove prop-types. I looked around for something else (e.g. a Webpack plugin) and didn't find one. :confused:

PR #3940, opened 2017-06-25
Bind web UI access tokens to sessions

### LINKED ISSUE #1681: OAuth access tokens - never removed, never expire

### LINKED ISSUE #2347: Web client stayed logged in when using multiple tabs

Gargron:
PR based on #3929

- Fix #2347 - Bind web UI access tokens to sessions

  When you logout, session also destroys the access token, so it's no longer
  valid. If access token is destroyed some other way, the session is also
  destroyed, requiring a re-login.

- Fix #1681 - Add scheduler to remove revoked access tokens and grants


beatrix-bitrot (approved):
lgtm~

app/models/session_activation.rb
```diff
@@ -53,4 +59,22 @@ def exclusive(id)
...
+  private
+
+  def assign_user_agent
+    self.user_agent = '' if user_agent.nil?
+  end
+
+  def assign_access_token
+    superapp = Doorkeeper::Application.find_by(superapp: true)
+
+    return if superapp.nil?
```
unarist: Since this check hasn't existed previously, it will breaks instances which don't have "Web" application. For example, Docker-Guide says you should run docker-compose run --rm web rake db:migrate, but db:migrate doesn't call db:seed which contains "Web" app generation.

Anyway, super app must be existent, so we shouldn't ignore non existence of it I think.

Gargron: This must have existed previously. The old code was in `HomeController#find_or_create_access_token`, which also called `Doorkeeper::Application.where(superapp: true).first` - how could those instances have worked then? Unless nil was allowed for application_id on oauth_access_tokens, which I didn't check.

unarist: We're allowing nil for application_id: https://github.com/tootsuite/mastodon/blob/v1.4.6/db/schema.rb#L193

unarist: Looks like our `db:migrate` does also `db:setup` now, but it was introduced at v1.4rc4...
https://github.com/tootsuite/mastodon/commit/256e3adc1d9508423aab8fcfb13745c9f85ff948#diff-1bf337030f4e56fa36281d23cf9e2082

unarist:
BTW, don't you force users logged out when access_token is missing?

PR #3946, opened 2017-06-25
WiP: Mark undelivered mentions

### LINKED ISSUE #3945: Visual indication when toots not delivered to mentionned accounts
Because of remote instances downtime, connectivity issues, etc., toots might not get delivered to remote instances.

There could be some visual indication on undelivered mentions for outgoing toots. UI-wise, it could be a different color for the mention link, or a symbol prepended or appended to the mention link.

This should be pretty easy to add, and I can do that if people are interested.

* * * *

ClearlyClaire:
This is a work-in-progress attempt to fix #3945.

It does correctly (I hope) mark outgoing mentions as undelivered by default, and mark them delivered whenever it can. It also outputs an “undelivered” class for those mentions' links in the formatted status HTML.

It lacks:
- Some CSS to actually have a visual indicator
- Maybe something else than CSS for accessibility
- Last but not least, some mechanism to refresh the mentions' status

EDIT: Also, I'm not too happy with all the mentions-specific stuff in more general notification code

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -21,8 +21,8 @@ class Formatter
 
     return reformat(raw_content) unless status.local?
 
-    linkable_accounts = status.mentions.map(&:account)
-    linkable_accounts << status.account
+    linkable_accounts = status.mentions.map { |item| [item.account, item.delivered] }
+    linkable_accounts << [status.account, true]
 
     html = raw_content
     html = "RT @#{prepend_reblog} #{html}" if prepend_reblog
@@ -113,8 +113,8 @@ class Formatter
 
     return link_to_account(acct) unless linkable_accounts
 
-    account = linkable_accounts.find { |item| TagManager.instance.same_acct?(item.acct, acct) }
-    account ? mention_html(account) : "@#{acct}"
+    account = linkable_accounts.find { |(item, _)| TagManager.instance.same_acct?(item.acct, acct) }
+    account ? mention_html(account[0], account[1]) : "@#{acct}"
   end
 
   def link_to_account(acct)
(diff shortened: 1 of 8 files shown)


Gargron:
The retrying is good, but I'm not sure this information should be exposed to all viewers. It only concerns the author. Also it only makes sense for local statuses.

ThibG:
I'm not sure there is value in *hiding* the info from other users than the author. As for it only making sense for local statuses: yes, and those statuses are the only ones that will have mentions with `delivered` as `false`.

Gargron:
Imagine I am looking at someone else's toot with mentions, and I see all this information about "a is notified", "b is not notified". What can I do with this information? Nothing, it is useless, but it adds to the perceived complexity of the UX. Even to the author, this information is only tangentially useful, if a remote server is down, you can't *do* anything about it.

Anyway, if you wanna extract the retrying fix to another PR i will merge it quickly. The other stuff will take some talking.

ThibG:
Split that into its own PR (#3960). As for this PR, it's intended as a small visual indicator (maybe a different color, or link decoration) to mark mentions as undelivered. Not being able to do anything about it isn't completely true: you might know how to contact the user through a different medium. I think an “undelivered” indication is helpful, it has been requested, and it is implemented in some way in most protocols where delivery can fail.

Gargron:
I think that this is a level of detail that people don't need. Admins should care about network failures, users should not. Even if this was offered in a special, separate view, I think the usefulness of the data would be near 0 because these aren't "read receipts" like in IM apps.

ThibG:
Well, I disagree, direct messages especially, are expected to be delivered, and will be assumed to be if there is no indication of the contrary. E-Mail usually don't have “read receipts” either and still have notification mechanisms for undelivered mails, for instance.

Anyway, this PR was really outdated and I don't have a proper idea how to do it properly for now, both in terms of UX and code. So I'm fine with the PR being closed. The issue it attempted to fix should remain open, though.

PR #3949, opened 2017-06-25
Fix #1624 - Send e-mail notifications to admins about new reports

### LINKED ISSUE #1624: Feature request - notifications for new reported policy violations
Could we add a notification of some sort for admin users when a toot or user has been reported for a policy violation? I apologize if I'm missing where this is controlled if it's present - I couldn't find it anywhere.

* * * *

Gargron:


nightpool:
this needs the manage:translations stuff still, right?

My only concern is that because these emails all have the same subject, they may get threaded by some mail clients (like gmail), which could make it easy to miss reports. Not sure how to trade that off with the possible privacy concerns of including the name of the reporter/reported in the subject line though. Maybe include the report ID? I think my ideal subject line would be something like "`New report for cybre.space (#101)`"

Gargron:
>manage:translations

That's for the React front-end, which plays no role here

Gargron:
Adjusted subject

nightpool:
@Gargron I just know that most times when we add a new translation key, we mirror that content to the rest of the translation files. Was that only for front-end translations?

nightpool (approved):
code LGTM

Gargron:
That's what i18n-tasks is for tho

PR #3954, opened 2017-06-26
Add .gitattributes file to avoid unwanted CRLF

danhunsaker:
When Windows checks out files, it defaults to changing line endings to CRLF. If these files are then copied to a Linux system to be run, and the endings aren't changed at some point in that process, things break. This file forces git to use LF for all text files on all systems (except the request testing specfiles) to prevent issues everywhere.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,14 @@
+*                             text=auto eol=lf
+*.eot                         -text
+*.gif                         -text
+*.gz                          -text
+*.ico                         -text
+*.jpg                         -text
+*.mp3                         -text
+*.ogg                         -text
+*.png                         -text
+*.ttf                         -text
+*.webm                        -text
+*.woff                        -text
+*.woff2                       -text
+spec/fixtures/requests/**     -text !eol


unarist (changes requested):
This patch breaks test fixtures. HTTP response includes CRLF, so you shouldn't normalize those files.

Only `.gitattributes` should appears as [a changed file](https://github.com/tootsuite/mastodon/pull/3954/files).

danhunsaker:
Ah. I had wondered about that. Fixing now.

danhunsaker:
Commit replaced. I tested the updated `.gitattributes` file to ensure it won't cause git to normalize those files again, and it was successful.

(Thought about forcing them all to CRLF, but decided against that; leaving them untouched across systems, as though they were binary files, allows more flexibility in writing tests...)

Gargron (approved):

.gitattributes
```diff
@@ -0,0 +1,14 @@
...
+*.ico                         -text
+*.jpg                         -text
+*.mp3                         -text
+*.ogg                         -text
+*.png                         -text
+*.ttf                         -text
+*.webm                        -text
+*.woff                        -text
+*.woff2                       -text
+spec/fixtures/requests/**     -text !eol
```
unarist: `!eol` seems to be not needed, but it's ok.

unarist (approved):

.gitattributes
danhunsaker: Yeah, spare bit of paranoia. I don't trust git as much as I should...

PR #3956, opened 2017-06-26
i18n: Updated Polish translation

mkljczk:


Gargron (approved):

Gargron (changes requested):
config/locales/pl.yml): could not find expected ':' while scanning a simple key at line 407 column 5

Gargron (approved):

PR #3962, opened 2017-06-26
Moving things around in the UI

MCMic:
Hello,

First of all please note the code in the PR is more of a proof of concept than clean code ready to be included.
It was just easier for me to try and code what I had in mind to see if it behaves well.

I identified the following problems with the UI:
- Unused space in the first column below the compose form
- The buttons in the header of the first column act on the last one
- In these buttons some change page instead (settings, logout)
- To access some common features you need to go through the «getting started» page, like to access favorites, or notifications (if they are un-pinned, of course)

So what I did was the following:
- Move the header button par to the last column instead of the first one
- Make it behave like the tabs bar from mobile UI (mobile UI feels way better than desktop UI IMO)
- Remove from it links to getting started, settings, logout
- Move the bottom part of the getting started page below the compose form, to be able to access settings and logout, and have the nice art there at all times

Attached is a screenshot of the result
mastodon_myui

What do you think?

### DIFF
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -1,8 +1,17 @@
 import React from 'react';
+import NavLink from 'react-router-dom/NavLink';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { FormattedMessage } from 'react-intl';
+import { connect } from 'react-redux';
+import ImmutablePropTypes from 'react-immutable-proptypes';
 
+const mapStateToProps = state => ({
+  me: state.getIn(['accounts', state.getIn(['meta', 'me'])]),
+  columns: state.getIn(['settings', 'columns']),
+});
+
+@connect(mapStateToProps)
 export default class ColumnHeader extends React.PureComponent {
 
   static contextTypes = {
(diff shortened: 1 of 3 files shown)


Gargron:
>Unused space in the first column below the compose form

This space is consumed as soon as:

- You reply to something
- You type in a long text
- You attach media

>The buttons in the header of the first column act on the last one

Yes, but *everything* acts on the last one.

I can understand the idea behind making the navigation bar act like a tab bar when it looks like it's updating tabs, however I think that moving the main navigation bar is unintuitive. Besides, it would mess with people's already existing muscle memory unnecessarily.

Thank you for caring enough to submit a PR. You shouldn't feel bad about it getting rejected because we almost never accept visual changes.

MCMic:
Hey, thanks for answering.

Yes, space is consumed if you are redacting something but that is only temporary. And my modifications does not get in the way of this.

I don’t understand what you call «main navigation bar».
In my UI the tabs are all on top of the column you can see them in, it makes sense.

If you never accept visual changes you should really look into having some kind of theme system because I’m already not updating because of my local changes I’d like to avoid having to maintain a fork.

IMO mastodon should care about having a consistent UI and something newcomers can easily understand. I was completely lost when I joined mastodon.

PR #3963, opened 2017-06-26
Responsive images in media gallery

sorin-davidoi:
This PR makes use of the `srcset` and `sizes` attributes of the `img` element to display responsive images. My pain-point is that in the mobile view, the preview image is too small and looks pixelated if the viewport width is over `400px`.

[Browser support](https://caniuse.com/#feat=srcset): Firefox, Chrome, Safari, Edge.

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -85,14 +85,24 @@ class Item extends React.PureComponent {
     let thumbnail = '';
 
     if (attachment.get('type') === 'image') {
+      const previewUrl = attachment.get('preview_url');
+      const previewWidth = attachment.getIn(['meta', 'small', 'width']);
+
+      const originalUrl = attachment.get('url');
+      const originalWidth = attachment.getIn(['meta', 'original', 'width']);
+
+      const srcSet = `${originalUrl} ${originalWidth}w, ${previewUrl} ${previewWidth}w`;
+      const sizes = `(min-width: 1025px) ${320 * (width / 100)}px, ${width}vw`;
+
       thumbnail = (
-        <a // eslint-disable-line jsx-a11y/anchor-has-content
+        <a
           className='media-gallery__item-thumbnail'
-          href={attachment.get('remote_url') || attachment.get('url')}
+          href={originalUrl}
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/components/media_gallery.js
```diff
@@ -85,14 +85,24 @@ class Item extends React.PureComponent {
     let thumbnail = '';
 
     if (attachment.get('type') === 'image') {
+      const previewUrl = attachment.get('preview_url');
+      const previewWidth = attachment.getIn(['meta', 'small', 'width']);
+
+      const originalUrl = attachment.get('remote_url') || attachment.get('url');
```
Gargron: You probably *don't* want to load remote_url just like this - it could be HTTP (while from HTTPS), and it would reveal the user's IP address to a remote server. So we've kinda decided against hotlinking.

sorin-davidoi: Isn't this what we are currently doing?

```js
attachment.get('remote_url') || attachment.get('url')
```

sorin-davidoi: Ah, we use it in `href`, not `src`.

sorin-davidoi: Ah, we use it in `href`, not `src`. `attachment.get('url')` then, I guess?

app/javascript/mastodon/components/media_gallery.js
```diff
@@ -85,14 +85,24 @@ class Item extends React.PureComponent {
...
+
+      const srcSet = `${originalUrl} ${originalWidth}w, ${previewUrl} ${previewWidth}w`;
+      const sizes = `(min-width: 1025px) ${320 * (width / 100)}px, ${width}vw`;
+
       thumbnail = (
-        <a // eslint-disable-line jsx-a11y/anchor-has-content
+        <a
           className='media-gallery__item-thumbnail'
-          href={attachment.get('remote_url') || attachment.get('url')}
+          href={originalUrl}
```
ykzts: `href={attachment.get('remote_url') || originalUrl}`?

ykzts (approved):

PR #3965, opened 2017-06-26
Turn report screen into a modal

Gargron:
image

(Account selected just for demonstration purposes)


beatrix-bitrot (approved):
seems like quite an improvement to me

Issue #3407, opened 2017-05-29
Glitchy-looking scrolling on iOS Safari


nolanlawson:
Not sure if this is a regression that was introduced in 1.4.1 or if it always existed, but scrolling can look kind of jumpy/jerky/glitchy on iOS Safari (10.3.1): https://gfycat.com/TiredAmazingBordercollie

I can't reproduce this issue on mobile Chrome or Firefox on an Android device. :confused:

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

daprice:
I've experienced this in MacOS Safari as well. The jumpy scrolling in @nolanlawson's recording is something I've noticed since I started using Mastodon several months ago (I attributed it to posts being added to the timeline during scrolling), but as of 1.4 or 1.4.1 it's worse and sometimes it starts jumping back and forth rapidly after the scroll finishes, and I have to reload the page to stop it.

daprice:
Not sure how helpful this is, but I opened the web inspector while it was happening and it looks like the first and last visible toots’ `<div>`s are constantly having a style tag added and removed multiple times a second, causing a feedback loop where they come in and out of the viewable area.

I uploaded a recording including the web inspector (warning: may cause seizures) [here](https://gfycat.com/EntireSingleArcherfish). Note that I wasn’t scrolling during this; it started after I scrolled down and more toots came into view. It’s difficult to replicate as it doesn’t do this after every scroll, but it seems like there are certain toots that always trigger it when they come into view (possibly ones containing images?).

nolanlawson:
hopefully fixed by #3972

PR #3972, opened 2017-06-27
Fix timeline jiggles/jitters

### LINKED ISSUE #3648: Timeline keeps repeatedly jumping up and down
On queer.party (running 1.4.1) I often have a problem when scrolling on the timeline where the view will start repeatedly jumping up and down. I suspect it's something to do with images being dynamically added and removed based on some threshold which is causing a dynamic equilibrium to occur.

See https://www.youtube.com/watch?v=PArRfqp0o4k for a video capture of the behavior. At the beginning of the video it's already taking place; then starting at 0:02 I'm scrolling normally, and then the behavior reappears at about 0:12. While I'm scrolling manually the behavior doesn't occur but it seems to be when the scrolling ends at a particular edge condition that it happens.

I am running Safari 10.1.2 (build 12603.3.3) on macOS Sierra.

* * * *

sorin-davidoi:
Hopefully, the final nail in the coffin since this closes #3648 , #3776, #3407.

[About `clientHeight`](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight):
>
> Note: This property will round the value to an integer. If you need a fractional value, use element.getBoundingClientRect().

This means that the height of the "rendered" toots and the height of the "unrendered" toots is not the same, so switching between them makes the scroll position jump.

image

image

A bit concerned about the performance of `getBoundingClientRect`. At first glance, it does not seem to be an issue.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -124,7 +124,7 @@ export default class Status extends ImmutablePureComponent {
 
   saveHeight = () => {
     if (this.node && this.node.children.length !== 0) {
-      this.height = this.node.clientHeight;
+      this.height = this.node.getBoundingClientRect().height;
     }
   }
 


Gargron (approved):

nolanlawson:
`getBoundingClientRect()` may force layout, which can be expensive. Another way to do this would be to use the `onIntersection` event, which passes in the rect object via the entry.

Regardless though you're right that we get more accurate height from the bounding rect; good call!

nolanlawson:
Hm just occurred to me that `onIntersection` may not work on non-Chrome browsers because currently only Chrome emits an event for all elements regardless of their intersection status (https://github.com/WICG/IntersectionObserver/issues/222). So perhaps this solution is the most performant one for now.

nolanlawson:
Actually wait, this may be doable. I'll open a PR.

nolanlawson:
Unfortunately it looks like observer entries won't work because the bounding rect that Chrome reports uses rounded integers, whereas `getBoundingClientRect()` doesn't. ☹️ Edge, Firefox, and the polyfill don't have this issue, so it's apparently a bug in Chrome. I'll go ahead and file it on them.

nolanlawson:
OTOH we could go with this solution and just accept jittering in Chrome. Arguably since Chrome has the scroll jank intervention, we could still get a jank-free solution across all browsers. But this needs more investigation.

nolanlawson:
Filed an issue on Chrome, hopefully they will fix soon: https://bugs.chromium.org/p/chromium/issues/detail?id=737228

sorin-davidoi:
Thanks for taking the time to report all these issues.

PR #3979, opened 2017-06-28
Put images behind CWs

marrus-sh:
This commit simply puts images inside a post's `<StatusContent>` so that it will be hidden behind CWs.

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -19,6 +19,7 @@ export default class StatusContent extends React.PureComponent {
     onExpandedToggle: PropTypes.func,
     onHeightUpdate: PropTypes.func,
     onClick: PropTypes.func,
+    children: PropTypes.element,
   };
 
   state = {
@@ -107,7 +108,7 @@ export default class StatusContent extends React.PureComponent {
   }
 
   render () {
-    const { status } = this.props;
+    const { status, children } = this.props;
 
     const hidden = this.props.onExpandedToggle ? !this.props.expanded : this.state.hidden;
 
@@ -144,7 +145,13 @@ export default class StatusContent extends React.PureComponent {
 
           {mentionsPlaceholder}
 
-          <div className={`status__content__text ${!hidden ? 'status__content__text--visible' : ''}`} style={directionStyle} dangerouslySetInnerHTML={content} />
+          <div className={`status__content__spoiler ${!hidden ? 'status__content__spoiler--visible' : ''}`}>
+
+            <div style={directionStyle} dangerouslySetInnerHTML={content} />
+
+            {children}
+
+          </div>
         </div>
       );
     } else if (this.props.onClick) {
@@ -155,8 +162,10 @@ export default class StatusContent extends React.PureComponent {
           style={directionStyle}
           onMouseDown={this.handleMouseDown}
           onMouseUp={this.handleMouseUp}
-          dangerouslySetInnerHTML={content}
-        />
+        >
+          <div dangerouslySetInnerHTML={content} />
+          {children}
+        </div>
       );
     } else {
       return (
@@ -165,7 +174,10 @@ export default class StatusContent extends React.PureComponent {
           className='status__content'
           style={directionStyle}
           dangerouslySetInnerHTML={content}
-        />
+        >
+          <div dangerouslySetInnerHTML={content} />
+          {children}
+        </div>
       );
     }
   }
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -19,6 +19,7 @@ export default class StatusContent extends React.PureComponent {
     onExpandedToggle: PropTypes.func,
     onHeightUpdate: PropTypes.func,
     onClick: PropTypes.func,
+    mediaIcon: PropTypes.string,
     children: PropTypes.element,
   };
 
@@ -108,7 +109,7 @@ export default class StatusContent extends React.PureComponent {
   }
 
   render () {
-    const { status, children } = this.props;
+    const { status, children, mediaIcon } = this.props;
 
     const hidden = this.props.onExpandedToggle ? !this.props.expanded : this.state.hidden;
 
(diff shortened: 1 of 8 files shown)


nightpool:
@akihikodaki why is this marked as a breaking change? I don't think it breaks any technical aspect of the software, and it certainly wouldn't have an impact on upgrades, or federation.

app/javascript/mastodon/components/status_content.js
```diff
@@ -144,7 +145,13 @@ export default class StatusContent extends React.PureComponent {
 
           {mentionsPlaceholder}
 
-          <div className={`status__content__text ${!hidden ? 'status__content__text--visible' : ''}`} style={directionStyle} dangerouslySetInnerHTML={content} />
+          <div className={`status__content__spoiler ${!hidden ? 'status__content__spoiler--visible' : ''}`}>
+
+            <div style={directionStyle} dangerouslySetInnerHTML={content} />
+
+            {children}
+
```
nightpool: remove extra newlines

marrus-sh:
to be clear this feature is already deployed on `dev.glitch.social`

nightpool:
I'm -1 on this for a couple of reasons:

 - it provides no indication to the user that there's any sort of media behind the content warning, so it has the potential to be very surprising for users when they open the content warning, expecting only textual content. For example, some people may be okay with textual discussions of specific content, but hate 

 - previous testing in this space (#2666) and my own experience indicates that there are valid usecases that this doesn't provide enough flexibility for. For example, some users use the hidden text to provide extended descriptions or source links for an image, while not cluttering up the main feed. In this case, hiding the image is counterproductive. Another example is using an image as a introduction or a call-to-action for a longer discussion (I've used this when I post mastodon statistics, so people can see the graphs without having to read my rambling analysis)

akihikodaki:
@nightpool I meant it breaks users' assumption which already exists. That could be technical or something different; such distinctions could be done with other tags such as _api_ and _deployment_.

akihikodaki (changes requested):
Apart from preference for this change, it has a problem regarding interoperability of other Mastodon instances.
For example, a user of an instance with this change may assume images would be hidden and post without NSFW. However, images are not hidden for users on other instances without this change.
You need to change the federation code as well to avoid that.

nightpool:
@akihikodaki While I agree that's an important thing to capture, it's a pretty substantially different thing then what that label has been used for in the past, or what it's trying to convey. If we tagged every pull request with that label that changed users expectations, we would be tagging the majority  of pull requests

I don't think the federation code is a blocker here, since there's no real reason this change can't go to all of the mastodon instances, and GNU social instances don't have the concept of sensitive media to worry about anyway. the small period where some instances have upgraded and others haven't isn't that big of a concern in my mind.

unarist:
Related discussion: #1010

Cassolotl:
>This commit simply puts images inside a post's <StatusContent> so that it will be hidden behind CWs.

I don't know what this means! Could we see screenshots of that this looks like in action?

Gargron:
The StatusContent component is reused in various places where it's not expected to contain more than text. That's one more problem with this PR.

As an alternative, you could extract the hidden/unhidden state upwards from StatusContent to the Status, and have it override state in MediaGallery and VideoPlayer. However, I am not convinced that is how this *should* work, and aki is right, this change would be breaking people's expectations.

Cassolotl:
@nightpool 

>it provides no indication to the user that there's any sort of media behind the content warning, so it has the potential to be very surprising for users when they open the content warning, expecting only textual content.

That's not true - there is a little picture image!

screen_shot_2017-06-28_at_14_10_56

>This change leaves the "sensitive" toggle in a no-mans-land of unclear usage, where it doesn't quite "fit in" with the rest of the product in any meaningful way. This is a more minor point, but preferably any pull request that resolved this issue would also take the sensitive toggle into account.

I would probably remove the Hide Media button completely, and change the CW button to the eye/crossed-eye symbol. I think that would be very clear for folks speaking all kinds of languages, and less confusing than the current situation.

Cassolotl:
>For example, a user of an instance with this change may assume images would be hidden and post without NSFW. However, images are not hidden for users on other instances without this change.

@akihikodaki How about if posts with CW selected are automatically tagged #NSFW?

unarist:
> That's not true - there is a little picture image!

Really? I cannot reproduce that behavior on this branch.

Previously we had replaced media urls with an icon like your screenshot, but that feature was removed at 1.4. Also you seems to be applied some user styles, so maybe those customization causes that behavior? (in default style, timestamp should appears on the right of username)

Cassolotl:
@unarist I don't have any userstyles applied in my browser, so it might be something specific to dev.glitch.social. Would it work for making it sufficiently clear there was an image under the CW? If so, could it be added to this PR?

nightpool:
@cassolotl the icon there is an unrelated dev glitch social feature, as far
as I know, and its not included in this pull request currently.

marrus-sh:
@Cassolotl The image icon in the top-right is a feature which was added with collapsed toots and isn't in this PR. (Mastodon puts its timestamp in the top-right and so it wouldn't fit 😉) However, we at GlitchSoc have (as of maybe an hour ago) made a few changes to make toots which contain images more apparent, and I'll hopefully add those along with other improvements to this PR sometime tomorrow. Stay tuned 📻

marrus-sh:
Now updated with patches and improvements. It's now obvious when toots have images inside.

image
image

akihikodaki:
@Cassolotl 
```
How about if posts with CW selected are automatically tagged #NSFW and marked sensitive, so that they would be hidden on other instances?
```
Sounds good for me. :+1:

marrus-sh:
@akihikodaki 

> A user of an instance with this change may assume images would be hidden and post without NSFW. However, images are not hidden for users on other instances without this change.

This is already something new users assume, and is already the way some frontends (eg., Amaroq) act. The purpose of making this PR in *upstream* is to make this assumption the standard.

“instances without this change” = instances running old software. I'm not saying that they're won't be a transition period, but I think this is a moot point in the long-run.

@Cassolotl

> How about if posts with CW selected are automatically tagged #NSFW and marked sensitive, so that they would be hidden on other instances?

I think the decision as to whether or not to unify CWs and sensitive content is beyond the scope of this PR, though.

akihikodaki:
@marrus-sh 
> “instances without this change” = instances running old software. I'm not saying that there won't be a transition period, but I think this is a moot point in the long-run.

(disclaimer: the below is just my opinion anyway. the decision to merge this or not is irrelevant.)
Here I clarify my opinion about this problem; the transition period matters even if it is few weeks. Though developers may understand why there are transition periods, it does nothing to do with users. Some users are newcomers and may not know well about Mastodon. Some may be just ignorant about recent changes. If you are kind of those users, how could you be convinced with the situation where you are told to put NSFW for your image while it is hidden for you?
My answer is no. Actually I'm really ignorant for various changes of Web services I use, and I do not think I have to be punished because of that.

wxcafe (approved):
LGTM.
I think this should be merged ASAP too.

Issue #3649, opened 2017-06-09
Provide account note in a plain text via API as opposite of update_credentials


unarist:
`/api/v1/accounts/update_credentials` only accepts notes in a plain text, but there is no way to obtain account note in a plain text via API. So apps which want to edit profile in the app have to strip tags.

I know that providing contents in a plain text is hard for remote accounts, but it should be able to do for local accounts, especially their own accounts.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Resolved with #4075

PR #3988, opened 2017-06-28
#3649 Added option return note in plain text (enoded HTML) in verify_credentials API.

debanshuk:
In `simplified_format` function (which is used to format HTML note) there is following condition at the beginning:

`return reformat(account.note) unless account.local?`

I have kept the same condition in new `simplified_plain_text_format` function too.

Issue link: https://github.com/tootsuite/mastodon/issues/3649

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -52,6 +52,11 @@ class Formatter
     html.html_safe # rubocop:disable Rails/OutputSafety
   end
 
+  def simplified_plain_text_format(account)
+    raise Mastodon::ValidationError, 'Local account needed to get note in plain text. Got remote account.' unless account.local?
+    encode(account.note)
+  end
+
   def sanitize(html, config)
     Sanitize.fragment(html, config)
   end
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -6,6 +6,7 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
 
   def show
     @account = current_account
+    @plain_text_note = query_params[:note_format] == 'plain'
     render 'api/v1/accounts/show'
   end
 
@@ -19,5 +20,9 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
 
   def account_params
     params.permit(:display_name, :note, :avatar, :header)
+  end
+
+  def query_params
+    params.permit(:note_format)
   end
 end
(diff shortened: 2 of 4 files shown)


akihikodaki (changes requested):
I'm not really for this feature. It prohibits future extensions allowed in the federation protocol and causes inconsistency with other APIs.
Applications can still convert HTML into plain text by themselves, which is also compatible for remote accounts. I understand `verify_credentials` does not require to support remote accounts, but I wonder it should have an option for such a limited usage.

Apart from my preference for the feature, I have added some comments, so I'd like you to review them.

db/schema.rb
```diff
@@ -155,6 +155,7 @@
     t.datetime "updated_at", null: false
     t.index ["account_id", "status_id"], name: "index_mentions_on_account_id_and_status_id", unique: true
     t.index ["status_id"], name: "index_mentions_on_status_id"
+    t.index ["status_id"], name: "mentions_status_id_index"
```
debanshuk: @akihikodaki, I don't remember doing any schema change. Not sure how it got it. I'll remove this.

debanshuk:
@akihikodaki, I have done changes according to your comments.

app/lib/formatter.rb
```diff
@@ -52,6 +52,11 @@ def simplified_format(account)
     html.html_safe # rubocop:disable Rails/OutputSafety
   end
 
+  def simplified_plain_text_format(account)
+    raise Mastodon::ValidationError, 'Got remote account. Excepting local.' unless account.local?
```
akihikodaki: Nice. It would be better to tell that `plain` option caused the error.

akihikodaki (dismissed):

nightpool:
Can you lay out the usecase for this? I think allowing clients to discard the formatting users have done on their bios is a pretty bad behavior to encourage.

debanshuk:
@nightpool, What should be the right way according to your to solve this problem?

nightpool:
I think the approach in https://github.com/tootsuite/mastodon/pull/4075 is better, but not perfect.

The key distinction here isn't that the account is local—it's that it's an account you can edit, so you need to be able to figure out the specific text used for editing, which is distinct from the text used for display.

My concern with #4075 is that it overloads the existing verify_credentials route, but I think it's much cleaner to add a new property to accounts that you can edit, one that shows the exact note text, to a new route.

debanshuk:
@nightpool, If I am understanding correctly, you are asking to add a new attribute (something like 'note_plain' or 'note_raw') in the response of verify_creadentials, which can be used for editing and the current 'note' attribute will be used for display purpose only. 

That actually does sound better to me too. 

I can do changes according to this logic.

ykzts:
Resolved with #4075

PR #3995, opened 2017-06-29
Delay unmount of some components

unarist:
This brings following benefits:

* Column switching would be bit smoothly. Swipe gesture works well with this.
* Preserve scroll position even if switch column by menu or swipe, not back button.
* In private mode on Safari, react-router-scroll won't work due to unavailability of sessionStorage. This implementation works.

Similar patch (#2271) has been implemented at 1.3.x, but reverted due to the performance and style issues. So I've introduced some improvements.

* Prevent update of status list while the column is hidden.
* No additional container element. This will reduces style regression.
* After some time (currently 5min.), hidden component will be unmounted.

Concerns:

* Performance: I don't know whether no performance issue for all platforms, though this patch and 1.4.x brings many performance improvements.
* Columns: home/local/federated timelines are implemented.
* Unmounting grace period: 5min, 3min, forever, etc...

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -37,6 +37,7 @@ export default class Status extends ImmutablePureComponent {
     autoPlayGif: PropTypes.bool,
     muted: PropTypes.bool,
     intersectionObserverWrapper: PropTypes.object,
+    visible: PropTypes.bool,
   };
 
   state = {
@@ -59,8 +60,25 @@ export default class Status extends ImmutablePureComponent {
 
   updateOnStates = ['isExpanded']
 
+  componentWillReceiveProps (nextProps) {
+    if (this.props.visible !== nextProps.visible) {
+      this.setState((prevState, props) => {
+        if (prevState.isIntersecting && !props.visible) {
+          scheduleIdleTask(this.hideIfNotIntersecting);
+        }
+        return {
+          isHidden: props.visible ? !prevState.isIntersecting : prevState.isHidden,
+        };
+      });
+    }
+  }
+
   shouldComponentUpdate (nextProps, nextState) {
-    if (!nextState.isIntersecting && nextState.isHidden) {
+    if (!nextProps.visible && nextState.isHidden) {
+      return this.props.visible || !this.state.isHidden;
+    } else if (nextProps.visible && this.state.isHidden !== nextState.isHidden) {
+      return true;
+    } else if (!nextState.isIntersecting && nextState.isHidden) {
       // It's only if we're not intersecting (i.e. offscreen) and isHidden is true
       // that either "isIntersecting" or "isHidden" matter, and then they're
       // the only things that matter.
@@ -94,6 +112,10 @@ export default class Status extends ImmutablePureComponent {
   }
 
   handleIntersection = (entry) => {
+    if (!this.props.visible) {
+      return;
+    }
+
     // Edge 15 doesn't support isIntersecting, but we can infer it
     // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/
     // https://github.com/WICG/IntersectionObserver/issues/211
@@ -119,7 +141,7 @@ export default class Status extends ImmutablePureComponent {
     // and if so, set our isHidden to true to trigger an unrender. The point of
     // this is to save DOM nodes and avoid using up too much memory.
     // See: https://github.com/tootsuite/mastodon/issues/2900
-    this.setState((prevState) => ({ isHidden: !prevState.isIntersecting }));
+    this.setState((prevState, props) => ({ isHidden: !props.visible || !prevState.isIntersecting }));
   }
 
   saveHeight = () => {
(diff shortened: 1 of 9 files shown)


sorin-davidoi:
I think this is a great idea, especially now with swipeable columns. I've always found that when changing columns, we spend a considerable amount of time unredering the current column before rendering the next one. Delaying the unredering should indeed make the transition smoother. My major concern in the number of nodes in the DOM. Now, since we "unrender" toots it should not be of grave concern.

Some issues I found:
 - The content of the hidden columns is not unrendered - the "rendered" toots are still in to DOM. I guess we need to manually handle this, since IntersectionObserver does not seem to fire.
 - I think a grace period of minutes is too long - would prefer something like 10 seconds.

unarist:
> The content of the hidden columns is not unrendered - the "rendered" toots are still in to DOM. I guess we need to manually handle this, since IntersectionObserver does not seem to fire.

At the first time, I didn't touched around IntersectionObserver. However, this makes status appearing bit later because it takes following steps:

1. Render status elements
2. IntersectionObserver catches intersection, and sets status "rendered" state
3. Render status contents

Then I've disabled IntersectionObserver around column hiding. To handle this appropriately, probably we want to 1) make "unrender" state just after column hidden and 2) make "render" state **just before** column appears... i.e. force "rendered" state even before real intersection.

But keeping "rendered" state will reduce remounting cost when the column will visible. So we may want to keep "rendered" state in short time, then make "unrendered" state...uh, it's probably too complicated.

BTW, current implementation won't re-enabled IntersectionObserver, so I will try to fix.

> I think a grace period of minutes is too long - would prefer something like 10 seconds.

I've intended following use cases.

* as a grace period of column change in short time or accidentally one
(in this case, I don't think making "unrender" state is needed, because it only exists 10 seconds)
* as a cache of previous timeline
(e.g. Home -> Compose -> Home, Home -> Account -> Home, etc.)
* as a cache of frequently used columns

I think 10 seconds is enough for a grace period, but not for other cases.

app/javascript/mastodon/components/status.js
```diff
@@ -119,7 +141,7 @@ export default class Status extends ImmutablePureComponent {
     // and if so, set our isHidden to true to trigger an unrender. The point of
     // this is to save DOM nodes and avoid using up too much memory.
     // See: https://github.com/tootsuite/mastodon/issues/2900
-    this.setState((prevState) => ({ isHidden: !prevState.isIntersecting }));
+    this.setState((prevState, props) => ({ isHidden: !props.visible || !prevState.isIntersecting }));
```
unarist: I've changed to make "unrendered" state while the status list is hidden, but you can try to disable it (previous behavior) by comment out `!props.visible ||`.

And this implementation probably not be efficient...

sorin-davidoi:
Caching all these columns (without unrendering) is going to hurt performance on mobile devices a lot. It is indeed fast on desktop to return to a cached column, I'm just not sure if it is a tradeoff worth making.

nolanlawson:
I tend to agree with @sorin-davidoi that I'm not sure the memory tradeoffs are worth the perf tradeoffs (especially for mobile). It would also be nice to see a timeline before/after or performance.measure before/after so we can know exactly what the benefit is.

Would it make sense to first look into why unmounting/mounting is so expensive in the first place? Last time I looked at `ReactPerf.printWasted()` there was still quite a bit of wasted time that could be optimized using `shouldComponentUpdate()`.

Another option is to use the technique Twitter Lite is using for a similar problem ([blog post](https://medium.com/@paularmstrong/twitter-lite-and-high-performance-react-progressive-web-apps-at-scale-d28a00e780a3), skip down to "Defer Rendering When Mounting & Unmounting Many Components"). Note that what they're doing is allowing the "switch" to be instant (in that the navigation bar renders instantly), but the column component itself renders a bit later. This improves perceived performance without having to use extra memory for a cache.

sorin-davidoi:
The trick mentioned in the article (with two calls to `requestAnimationFrame`) could be applied to the `Bundle` component from #3879, since it handles the loading, error and "full" state of all the columns (and some other components).

unarist:
> If we unrender them, there is no much difference, since we have to recreate 99% of the nodes in the column anyway.

I think so too. So memory consumption is a problem, this way is not a good solution I think. 

Actually, my main concern is mobile view since it only has single column and I change column frequently (to compose toots, or see other timelines). I also thought enable this behavior only on mobile views. So if it doesn't improve user experience on mobile devices, I would close this because too complication for less benefits...and it's probably true.

I'm frustrated by 

* huge lags on column changing (approx. 1 second)

nolanlawson:
I definitely agree it could be faster on mobile. (I use it on mobile too.) I'll profile and see if we can improve anything without needing to use extra memory. :)

app/javascript/styles/components.scss
```diff
@@ -1266,6 +1266,10 @@
   }
 }
 
+.column--hidden {
```
akihikodaki: c.f. https://github.com/tootsuite/mastodon/pull/2916#discussion_r116370464

akihikodaki:
Regarding stylesheets, #2916, related with #2271, would be informative, just in case you miss it.

unarist:
@akihikodaki Thanks for the information. I remember that bug existed, but I didn't read that PR actually.

Also I read your comment on the PR. While implementing this, I thought the scroll position may should be saved/restored for each components by ourselves, like `react-router-scroll` does.

Actually, I think this PR may be too complicated for only solving scroll position issue. So I've created separate issue for it (#4005).

unarist:
OK, I'm going to close this, and consider separately about scroll position.

PR #4000, opened 2017-06-29
Fix webpack-dev-server on Windows

nolanlawson:
Using 0.0.0.0 to bind to 127.0.0.1/localhost is unfortunately one of those things that just doesn't work on Windows. This fixes that, so that you can run webpack-dev-server on Windows, e.g. in Bash on Ubuntu on Windows.

### DIFF
--- a/bin/webpack-dev-server
+++ b/bin/webpack-dev-server
@@ -23,7 +23,7 @@ end
 begin
   dev_server = YAML.load_file(CONFIG_FILE)["development"]["dev_server"]
 
-  DEV_SERVER_HOST = "http#{"s" if args('--https') || dev_server["https"]}://#{args('--host') || dev_server["host"]}:#{args('--port') || dev_server["port"]}"
+  DEV_SERVER_HOST = "http#{"s" if args('--https') || dev_server["https"]}://#{dev_server["host"]}:#{args('--port') || dev_server["port"]}"
 
 rescue Errno::ENOENT, NoMethodError
   puts "Webpack dev_server configuration not found in #{CONFIG_FILE}."
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,4 @@
 web: PORT=3000 bundle exec puma -C config/puma.rb
 sidekiq: PORT=3000 bundle exec sidekiq
 stream: PORT=4000 yarn run start
-webpack: ./bin/webpack-dev-server --host 127.0.0.1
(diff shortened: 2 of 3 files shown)


kepstin:
This is a behaviour change on Linux - listening to 0.0.0.0 is the "wildcard" address, which means that the server is listening on all network interfaces, and can be accessed from other machines. With this change it will only be accessible from localhost - which might be a good idea anyways, for a development environment.

nightpool:
I prefer binding on 0.0.0.0, because it means I can test stuff from my phone and on other devices.

nolanlawson:
@nightpool Don't you have to change the config anyway to get that to work? If you're running a dev server on 192.168.0.x and it tries to hit 0.0.0.0:8080 from 192.168.1.x:3000 in the browser, it doesn't know to use 192.168.1.x right?

I agree though that the server can run on 0.0.0.0; it shouldn't make a difference since the problem is in _accessing_ 0.0.0.0 rather than serving from 0.0.0.0. I'll test and update the PR.

nolanlawson:
OK, this ^ should fix it so that `webpack-dev-server` _serves_ from 0.0.0.0 but is _accessed_ from 127.0.0.1. If you want it to be accessed from another URL (e.g. for testing on a phone), you'd just change the `host` in `webpacker.yml` like normal.

I can confirm that this fixes Mastodon's dev server on Windows, and I can also access Webpack from `<local network IP>:8080`.

nightpool (approved):
LGTM

bin/webpack-dev-server
```diff
@@ -23,7 +23,7 @@ end
 begin
   dev_server = YAML.load_file(CONFIG_FILE)["development"]["dev_server"]
 
-  DEV_SERVER_HOST = "http#{"s" if args('--https') || dev_server["https"]}://#{args('--host') || dev_server["host"]}:#{args('--port') || dev_server["port"]}"
+  DEV_SERVER_HOST = "http#{"s" if args('--https') || dev_server["https"]}://#{dev_server["host"]}:#{args('--port') || dev_server["port"]}"
```
nolanlawson: This variable is used to figure out how the webpage should access Webpack, and thus should use `dev_server["host"]`. But `--host 0.0.0.0` is still passed in to the `webpack` command below (via `Procfile.dev`) and so Webpack serves from 0.0.0.0.

PR #4001, opened 2017-06-29
Avoid using getBoundingClientRect to calculate height

nolanlawson:
(This PR is written on top of #3879 to avoid bitrotting myself when that gets merged. So unfortunately if you want to see a good diff you should check out https://github.com/nolanlawson/mastodon/commit/6057d38b10f40ef8cfed95c4782757e565a6ce5a.)

The goal of this PR is to avoid calling `getBoundingClientRect()` because it forces layout and thus can be expensive. In principle there's no need to call gBCR in order to get the height of list items, because that's already calculated within IntersectionObserver, which provides it automatically via [IntersectionObserverEntry.boundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/boundingClientRect).

Unfortunately it's not quite so easy, because [there's a bug in Chrome](https://crbug.com/737228) that causes this rect to be calculated incorrectly. Luckily the calculation is so inaccurate that we can easily detect it by comparing the two rects, and then fall back to manual `getBoundingClientRect()` calls only for Chrome.

So for browsers with built-in IntersectionObserver and without the bug (e.g. Edge 15+, Firefox 55+), this is a win because we never call gBCR. And for Chrome, it'll be fixed automatically when they fix the bug. As for browsers using the IntersectionObserver polyfill (e.g. Firefox <55, Safari), they are [already calling gBCR](https://github.com/WICG/IntersectionObserver/blob/b8da39911d14e3347f98ba44914378fc19a8786f/polyfill/intersection-observer.js#L622) and so this fix merely ensure they aren't calling it twice.

To show the perf difference, I ran the same scenario in Edge with and without this PR. The scenario was to just click on "local timeline" from the "getting started" page (using a dev build).

2017-06-29 12_05_35-celadon 06-29-2017 10-53-53 etl - windows performance analyzer

In Windows Performance Analyzer, I'm listing the DOM APIs from most to least expensive. If you ignore all the performance.mark/measure/clearMarks calls (which don't appear at all in production), you can see that 72ms of gBCR calls are completely eliminated. (Note this was done on an i5 SurfaceBook.)

/cc @sorin-davidoi

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -17,6 +17,7 @@ import { MediaGallery, VideoPlayer } from '../features/ui/util/async-components'
 // We use the component (and not the container) since we do not want
 // to use the progress bar to show download progress
 import Bundle from '../features/ui/components/bundle';
+import getRectFromEntry from '../features/ui/util/get_rect_from_entry';
 
 export default class Status extends ImmutablePureComponent {
 
@@ -101,6 +102,11 @@ export default class Status extends ImmutablePureComponent {
   }
 
   handleIntersection = (entry) => {
+    if (this.node && this.node.children.length !== 0) {
+      // save the height of the fully-rendered element
+      this.height = getRectFromEntry(entry).height;
+    }
+
     // Edge 15 doesn't support isIntersecting, but we can infer it
     // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/
     // https://github.com/WICG/IntersectionObserver/issues/211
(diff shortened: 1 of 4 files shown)


nolanlawson:
I should also note that this PR simplifies the code a lot, because instead of needing to re-calculate height whenever the status changes (e.g. when media is loaded), we can just check the bounding rect inside of `handleIntersection()`, at the moment we actually need the rect.

nolanlawson:
Heh well it seems not many folks noticed, because in that Chrome bug they said the bug has apparently been present since Chrome 51. 😬

sorin-davidoi:
In moments like these, I'm amazed the web platform even works at all. Out of all the APIs, we have only a handful of browser-specific workarounds (two in the same API).

ykzts (approved):

Gargron:
I merged the PR this was based on but unfortunately I don't think GitHub recognizes this sort of thing so I am not sure what you need to do to make this PR appear cleanly.

nolanlawson:
It needed a rebase, but it's ready now.

BTW it turns out that [the Chrome bug](https://crbug.com/737228) only occurs when the browser or OS is zoomed, so this PR ought to make Chrome faster as well in most cases. I did a quick perf check in Chrome without any zooming and saw that the `getBoundingClientRect()` and invalidations were eliminated. :smiley:

Gargron (approved):

PR #4012, opened 2017-06-30
Fix that AdminMailer does not send

ykzts:


akihikodaki (approved):
I'm approving, but here I note something not blocking merge:
It is odd that `AdminMailer`, which has nothing to do with stream entries, to have `StreamEntriesHelper`. It may be better to move `display_name` to somewhere else.

Issue #3866, opened 2017-06-20
unobserve() of IntersectionObserver breaks Edge support


unarist:
We've started to use `IntersectionObserver#unobserve()` at #3851, but this causes TypeMismatchError on Edge and prevents further script execution (e.g. back to Getting Started from Local timeline).

![image](https://user-images.githubusercontent.com/705555/27327208-d85a55a0-55e8-11e7-8b7d-7f72d41c193b.png)

I found some similar reports (`unobserve()` causes TypeMismatchError) by googling, so I think it would be a Edge's bug.

Microsoft Edge 40.15063.0.0
Microsoft EdgeHTML 15.15063

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [master] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

beatrix-bitrot:
@nolanlawson any thoughts on this?

sorin-davidoi:
I'll take a look.

sorin-davidoi:
I'd say we revert #3851. Unobserving the statuses, when they are unmounted, does not make sense, since they get unmounted when StatusList is unmounted, which destroys it's `IntersectionObserver`.

nolanlawson:
Thanks for the fix, and if you get a chance, please file a bug on Edge. :) We're working right now on making our IntersectionObserver implementation more compatible with Chrome's (and soon, Firefox's).

nolanlawson:
Also that particular fix was just fixing a React warning, so I would say we ignore the warning. Calling setState after unmount is due to how the requestIdleCallback wrapper is written and results in a no-op. So it makes the most sense just to ignore the warning IMO.

unarist:
I found actual cause.

```js
const observer = new IntersectionObserver(()=>{});
observer.unobserve(document.body);
```

This code calls `unobserve()` for the element which is not observed yet. It throws `TypeMismatchError` on Edge, and doesn't throw on Chrome and Firefox.

And we unobserve same target twice on column unmounting:

- Unmounting StatusList calls `disconnect()` (= unobserve all targets)
- Unmounting Status calls `unobserve()`

...so `TypeMismatchError` will be thrown on Edge.

Also I think this issue may occurs on rapid scrolling like this:

```js
const observer = new IntersectionObserver(()=>{});

// good
observer.observe(document.body);
observer.unobserve(document.body);
[...]

PR #4013, opened 2017-06-30
Unobserve status on unmount

unarist:
This fixes a warning on status unmounting (e.g. deletion) which mentioned in #3851.

This also resets IntersectionObserverWrapper on disconnect to avoid future `unobserve()` calls which has bug in Edge (#3866).


sorin-davidoi (approved):

nolanlawson (approved):
LGTM! And thanks for filing an issue on Edge. :)

PR #4014, opened 2017-06-30
Fix broken style in media gallery (regression from #3963)

ykzts:
| before | after |
|-|-|
|  |  |


sorin-davidoi:
Ups should have tested with different aspect ratios. Sorry about that.

Gargron (approved):

PR #4016, opened 2017-06-30
Allow to go home timeline back infinitely

akihikodaki:
(~~note: `xit` in 92e95fbec148dd6035aef533b200a878ab077242 could be removed with #4015~~ resolved)

__NOTE: the following description is out of date.__ See https://github.com/tootsuite/mastodon/pull/4016#issuecomment-323997016.

The basic idea is to limit ranges of status id.

The old query to regenerate feed retrieved statuses till it finds the particular number of statuses. It was problematic because:
* Not sustainable; for example, the burden could be twice or more as time passes and statuses are accumulated. (consider about new instances with many statuses such as mstdn.jp for example)
* Vulnerable for DoS attacks; it is easy to trigger the query even though attackers had to wait for 14 days.

Limiting range of ID effectively resolves the issue. (4f0ab569c1923ea7526929fb68d7815ae24103cc and 0c9313449b34b9d648e8c28814895a6caa5578f4)

With ID range limitation, it is safe to relax the limitation of numbers. c78b26eb3730c069bd74b163b6fee7f975a0310b allows to get statuses more than 400 by falling back to database when necessary.

With fallback to database, Redis does no longer have to hold many statuses and not have to be kept up-to-date for a long. 92e95fbec148dd6035aef533b200a878ab077242 changes the accordingly.

As the result, this change would have the following pros and cons.

# Pros
* Infinite status retrieval
* Less load for Redis
* Much less load for PostgreSQL when the old query took too long
* Better response in situations the old query took too long

# Cons
* Unable to get old statuses
* Less cached when getting the 40-100th newest statuses
* More load due to infinite status retrieval

### DIFF
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -10,7 +10,13 @@ class Feed
     if redis.exists("account:#{@account.id}:regeneration")
       from_database(limit, max_id, since_id)
     else
-      from_redis(limit, max_id, since_id)
+      statuses = from_redis(limit, max_id, since_id)
+
+      statuses += from_database(limit - statuses.size,
+                                statuses.last&.id || max_id,
+                                since_id)
+
+      statuses
     end
   end
 
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -10,12 +10,8 @@ class Feed
     if redis.exists("account:#{@account.id}:regeneration")
       from_database(limit, max_id, since_id)
     else
-      statuses = from_redis(limit, max_id, since_id)
-
-      statuses += from_database(limit - statuses.size,
-                                statuses.last&.id || max_id,
-                                since_id)
-
+      statuses  = from_redis(limit, max_id, since_id)
+      statuses += from_database(limit - statuses.size, statuses.last&.id || max_id, since_id) if statuses.size < limit
       statuses
     end
   end
--- a/spec/models/feed_spec.rb
+++ b/spec/models/feed_spec.rb
@@ -14,8 +14,19 @@ RSpec.describe Feed, type: :model do
       feed = Feed.new(:home, account)
       results = feed.get(3)
 
-      expect(results.map(&:id)).to eq [3, 2]
+      expect(results.map(&:id)).to eq [3, 2, 1]
       expect(results.first.attributes.keys).to eq %w(id updated_at)
+    end
+
+    it 'fall backs to database if Redis could not fill feed' do
+      account = Fabricate(:account)
+      statuses = 2.times.map { Fabricate(:status, account: account) }
+      Redis.current.zadd(FeedManager.instance.key(:home, account.id), statuses[1].id, statuses[1].id)
+
+      feed = Feed.new(:home, account)
+      results = feed.get(2, nil, 0)
+
+      expect(results.pluck(:id)).to eq statuses.pluck(:id).reverse
     end
   end
 end
--- a/spec/services/fan_out_on_write_service_spec.rb
+++ b/spec/services/fan_out_on_write_service_spec.rb
@@ -23,8 +23,7 @@ RSpec.describe FanOutOnWriteService do
   end
 
   it 'delivers status to local followers' do
-    pending 'some sort of problem in test environment causes this to sometimes fail'
-    expect(Feed.new(:home, follower).get(10).map(&:id)).to include status.id
+    expect(Feed.new(:home, follower).get(10, nil, 0).map(&:id)).to include status.id
   end
 
   it 'delivers status to hashtag' do


beatrix-bitrot:
overall, I think I like it. I have a couple comments and one question

1. I would appreciate more comments in the refactored get() method, since I think it's now a bit hard to read and probably can't be made much more readable without affecting performance.
2. The word 'continuously' is spelled incorrectly as 'continuosly' throughout the code
3. After this change, how can a user view very old statuses?

akihikodaki:
@beatrix-bitrot
1. I added comments which I thought helpful to understand.
2. Renamed the misspelled method.
3. You may give `max_id` to `/api/v1/timelines/home`. For the Web application, you can just scroll down. It will automatically call the endpoint with `max_id`.

Technowix:
I don't understand why this wouldn't be able to fetch old statuses? Isn't the "main" change a lazy population of the redis "cache" through the loading?

akihikodaki:
Because of the ID range limitation, which is actually the _main_ change. The new code will abandon to retrieve statuses for the requested number after it search statuses in a certain time range.

The length of range would be about 14 days for the latest statuses.
https://github.com/tootsuite/mastodon/pull/4016/files#diff-f132b598cc3b81683c143120cbf99edbR11

For older statuses, the length would be about a day.
https://github.com/tootsuite/mastodon/pull/4016/files#diff-4ddfad17f5a9eff0624179ccbd0ac36fR17

Note that you can still retrieve older statuses by specifying `max_id` or `since_id`. For example, if you found a very old status and you want to retrieve statuses in home timeline before it, specify `max_id` to its ID. You'll get statuses made in the day when the status was created.

The reason why this is the main change is that it significantly reduces the load of queries, or it is even impossible to maintain those queries without this change. Without this change, servers have to look through _all_ the statuses which have been ever created under some conditions.

akihikodaki:
The problem which would be addressed by commit b0ac481b06b099793786db76eda5e3e1c0817cf9 is impacting Pawoo instance.
The following is a case which took one minute.
```
> EXPLAIN ANALYZE SELECT  "statuses".* FROM "statuses" WHERE "statuses"."account_id" IN (?, ?, ?, ?, ?, ?, ?) AND "statuses"."visibility" IN (0, 1, 2) ORDER BY "statuses"."id" DESC LIMIT 100;

QUERY PLAN
--
Limit  (cost=61624.65..61624.90 rows=100 width=334) (actual time=60.674..60.727 rows=100 loops=1)
->  Sort  (cost=61624.65..61665.58 rows=16372 width=334) (actual time=60.673..60.693 rows=100 loops=1)
Sort Key: id DESC
Sort Method: quicksort  Memory: 76kB
->  Bitmap Heap Scan on statuses  (cost=467.51..60998.92 rows=16372 width=334) (actual time=7.341..60.468 rows=105 loops=1)
Recheck Cond: (account_id = ANY ('{?,?,?,?,?,?,?}'::integer[]))
Filter: (visibility = ANY ('{0,1,2}'::integer[]))
Heap Blocks: exact=105
->  Bitmap Index Scan on index_statuses_on_account_id_id  (cost=0.00..463.42 rows=16464 width=0) (actual time=7.299..7.299 rows=105 loops=1)
Index Cond: (account_id = ANY ('{?,?,?,?,?,?,?}'::integer[]))
Planning time: 14.853 ms
Execution time: 60.832 ms
```

Though the load caused by the issue is trivial for the server, users say that keeps browsers waiting too long.

akihikodaki:
9e7a9b0ec1d9408e7a6cefcfc700b3fa75b77235: this pull request has been kept opened for a long, so I dropped commits and left one commit which is not likely to regress UX in any terms.

app/models/feed.rb
```diff
@@ -10,7 +10,13 @@ def get(limit, max_id = nil, since_id = nil)
     if redis.exists("account:#{@account.id}:regeneration")
       from_database(limit, max_id, since_id)
     else
-      from_redis(limit, max_id, since_id)
+      statuses = from_redis(limit, max_id, since_id)
+
+      statuses += from_database(limit - statuses.size,
```
abcang: If `limit - statuses.size` is 0, it is better not to execute `from_database`

abcang (dismissed):

app/models/feed.rb
akihikodaki: I think it is trivial so have not added `if` statement (and spec)

abcang: This query will be executed even if `limit-statuses.size` is 0. I think this is a useless query execution.

akihikodaki: I mean the cost of the execution can be ignored.

ThibG:
This should be rebased on master, which adds a specialized `home_feed`.
There is also an issue with `from_database` potentially returning 0 statuses even in case were limit > 0 and such statuses exist in the database: that would happen if the first `limit` statuses returned from the SQL query were later filtered out (e.g., they are between one of your follower and someone you don't follow).

akihikodaki:
I'm a bit reluctant to make more changes because it is left for a long time and I'm busy in these days. But I still think this change is reasonable and will make required changes to get it merged if there is any chance. @Gargron, what is your opinion?

app/models/feed.rb
Gargron: There is no need to perform a LIMIT 0 query. It may not be slow in postgres itself, but sending the query to postgres (and taking up a connection from the pool with e.g. pgBouncer) is still a cost that can be avoided.

Gargron (approved):

ThibG:
As stated earlier, there is also an issue with `from_database` potentially returning 0 statuses, which will prevent paginating further.
I have a pretty straightforward fix for that, but it could be very costly in terms of the number of SQL queries in some situations… basically, a loop in the same style as `FeedManager#populate_feed`

Gargron:
>As stated earlier, there is also an issue with from_database potentially returning 0 statuses, which will prevent paginating further.

You have a point, I forgot about that. I do *not* want to do loops with queries in a live endpoint like this. I guess the PR isn't that useful after all. Sorry.

PR #4018, opened 2017-06-30
Defer status rendering

sorin-davidoi:
This is based on #3879 , so for a true diff see https://github.com/sorin-davidoi/mastodon/commit/a6e242daba97afdf2c1b103300680e88a0b92967

Addressing concerns raised by @unarist and discussed together with @nolan in #3995 related to the fact that switching columns takes a long time. This PR delays the rendering of some components inside `<Status />`. I took care to avoid changes in height. We improve the following metrics: time it takes to mount `StatusList` and `Status`.

Before:
image
Mounting StatusList: 100ms - 150ms
Mounting Status: ~6ms

After:
image
Mounting StatusList: ~49ms
Mounting Status: ~2.8ms

The small rendering "chunks" in the second screenshots are the defered components.

Still not fast enough to avoid skipping frames but better.


unarist:
It would makes changing column faster, but... users may see following state:

image

This looks weird. If we going to defer status rendering, whole contents should be deferred I think.

sorin-davidoi:
Hm, can you see that state during real usage? I could only see it if I wrapped the `this.setState({ showDeferedComponents: true });` inside two `requestAnimationFrame`.

unarist:
I've seen that state on iOS9 Safari and Edge.

gif

sorin-davidoi:
Ah. I guess the question is: do we want the users to read the content while we load the rest or hide the content until everything is ready? I would personally prefer to see the content as soon as possible, as long as the page doesn't jump up and down while loading the other stuff (as most ads do).

Taking a glance at my timeline, the text of the toots makes up the majority of the space (a.k.a. most toots are not one word), which would not make the view look as weird.

Which of the following would you prefer to be rendered later?
 - [ ] Display name
 - [ ] Action bar
 - [ ] Timestamp
 - [ ] Avatar

We cannot load the content later, as it is the one that determines the height (we could hide it though).

sorin-davidoi:
I guess these are subjective design decisions so pinging @Gargron here.

unarist:
Ah yes, that would be subjective. Probably, display name is important for me because I read it first when I read status text.

sorin-davidoi:
Yes, you are right. But I think the timestamp and the actions could be shown later.

beatrix-bitrot:
Is this ready for testing? If so I will bring it onto glitch.social when we finish testing the push notifications.

Gargron:
>I guess these are subjective design decisions so pinging @Gargron here.

If anything is going to be rendered deferred like that I recommend this look (not necessarily this library, but the look): https://github.com/danilowoz/react-content-loader

sorin-davidoi:
Took a quick look at `react-content-loader`. It is hard to configure to avoid height changes and it is quite heavy - might as well render the whole thing if we replace the timestamp and actions with it. No saying that these problems are insurmountable, but I don't have time right now to look into it.

I'd suggest taking a look at how this looks with real content, I would say it does not negatively impact the experience (maybe a short fade-in animation would help).

nolanlawson:
This is a really tough tradeoff, but I'd tend to side with @unarist that lazy-loading the author/actionbar/timestamp may look kind of awkward from a UX point of view. I wonder if it may even reduce the "perceived performance," i.e. users perceive it as being slower, since the content is being drawn in a scatter-shot way.

What I'm wondering is: why is it so expensive in the first place to render the author/timestamp/actionbar? We're talking about two pieces of text and four icons here. My hunch is that things like ReactMotion and ReactIntl are adding overhead, so there may be a way to speed those things up. (Or delay loading them by e.g. loading _just_ an icon, and then loading the icon wrapped in ReactMotion later. I dunno.)

Does Preact or React 16 improve the rendering at all?

Gargron:
React Fiber should improve rendering...

sorin-davidoi:
Alright, will see if we can piggy-back on React for this 👍

PR #4019, opened 2017-06-30
Faster emojify() algorithm, avoid regex replace

nolanlawson:
After doing some perf analyses on a real device (Nexus 5X), I saw that, when switching from the "getting started" column to the timeline column, the `replace()` function inside of `emoji.js` -> `shortnameToImage()` is very slow (50ms).

This is using a regex replacement using `emojione.regShortNames`, and the regex itself is fairly simple:

screenshot 2017-06-29 23 23 59

(Removed all but two shortnames from the regex to make it easier to read.)

Essentially we just need to replace `:strings_like_this:` and avoid `<tagsLikeThis>`. So we can write our own state machine that's faster than regex replacement.

So I wrote it, and can confirm that the speed is much improved:

out2

I also added some unit tests, and confirmed that the tests pass before my PR and after my PR (so my version has the same behavior as the regex version).

`unicodeToImage()` -> `replace()` is still slow, but that can wait for another PR.

/cc @sorin-davidoi @unarist

### DIFF
--- a/app/javascript/mastodon/emoji.js
+++ b/app/javascript/mastodon/emoji.js
@@ -36,7 +36,7 @@ const shortnameToImage = str => {
       if (shortname in emojione.emojioneList) {
         const unicode = emojione.emojioneList[shortname].unicode[emojione.emojioneList[shortname].unicode.length - 1];
         const alt = emojione.convert(unicode.toUpperCase());
-        const replacement = `<img draggable="false" class="emojione" alt="${alt}" title="${shortname}" src="/emoji/${unicode}.svg" />`
+        const replacement = `<img draggable="false" class="emojione" alt="${alt}" title="${shortname}" src="/emoji/${unicode}.svg" />`;
         str = str.substring(0, i) + replacement + str.substring(shortnameEndIndex + 1);
       } else {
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/emoji.js
```diff
@@ -19,16 +19,41 @@ const unicodeToImage = str => {
...
+  let insideShortname = false;
+  let shortnameEndIndex = -1;
+  while (i--) {
+    const char = str.charAt(i);
+    if (insideShortname && char === ':') {
+      const shortname = str.substring(i, shortnameEndIndex + 1);
+      if (shortname in emojione.emojioneList) {
+        const unicode = emojione.emojioneList[shortname].unicode[emojione.emojioneList[shortname].unicode.length - 1];
+        const alt = emojione.convert(unicode.toUpperCase());
+        const replacement = `<img draggable="false" class="emojione" alt="${alt}" title="${shortname}" src="/emoji/${unicode}.svg" />`
```
ykzts: add semi.

nolanlawson: :+1:

Gargron (approved):

PR #4021, opened 2017-06-30
Togglable filter links

noraworld:
Togglable filter links on `/admin/accounts`.

### Before:
toggle_filter_before

### After:
toggle_filter_after

### DIFF
--- a/app/helpers/admin/filter_helper.rb
+++ b/app/helpers/admin/filter_helper.rb
@@ -6,13 +6,19 @@ module Admin::FilterHelper
 
   FILTERS = ACCOUNT_FILTERS + REPORT_FILTERS
 
-  def filter_link_to(text, more_params)
-    new_url = filtered_url_for(more_params)
-    link_to text, new_url, class: filter_link_class(new_url)
+  def filter_link_to(text, link_to_params, link_class_params = link_to_params)
+    new_url = filtered_url_for(link_to_params)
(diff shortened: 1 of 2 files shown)


unarist:
Should it be togglable? It's enough to use same behavior as "Location" filter for "Moderation" filter, I think.

Also making only "Moderation" filter togglable seems inconsistent.

noraworld:
As for "Moderation", suppose we want to turn on only "Suspended" filter when "Silenced" is on and "Suspended" is off. Simply we click "Suspended" link, then both "Silenced" and "Suspended" are on. This behavior is correct because this case matches when there is (are) user(s) who is (are) silenced and suspended. However, "Silenced" is not turned off even though we click "Silenced" link. So we have to click "All" link, and then click "Suspended" link. I think this behavior is inconvenient and not intuitive.

Also as for "Location", "Local" / "Remote" filter remains on even though we click the filter that is on, not togglable, on the latest of the master branch. I made also "Location" filter togglable to keep consistency.

unarist:
Oh, sorry. I forgot about silenced && suspended is meaningful, and it's my misunderstood for toggle.

beatrix-bitrot:
Hmm, I like it. Do you think it's ok in the end @unarist ?

unarist (approved):
I'm thinking implementation can be improved maybe, but this feature looks good.

PR #4023, opened 2017-07-01
[nanobox] Add Automated Backups

danhunsaker:
This PR adds automatic backups to Nanobox instances. The database, Redis, and user files are backed up every day at 03:00 (server time) to the data warehouse component which comes with every Nanobox app. Old backups are automatically cleared out, but the number of backups that are left untouched can be configured by setting the `BACKUP_COUNT` environment variable to any integer value greater than 0 (the default is 1).

Also updated `.env.nanobox` to reflect the current `.env.production.sample`.

### DIFF
--- a/boxfile.yml
+++ b/boxfile.yml
@@ -153,8 +153,59 @@ worker.sidekiq:
 data.db:
   image: nanobox/postgresql:9.5
 
+  cron:
+    - id: backup
+      schedule: '0 3 * * *'
+      command: |
+        PGPASSWORD=${DATA_POSTGRES_PASS} pg_dump -U ${DATA_POSTGRES_USER} -w -Fc -O gonano |
+        gzip |
+        curl -k -H "X-AUTH-TOKEN: ${WAREHOUSE_DATA_HOARDER_TOKEN}" https://${WAREHOUSE_DATA_HOARDER_HOST}:7410/blobs/backup-${HOSTNAME}-$(date -u +%Y-%m-%d.%H-%M-%S).sql.gz --data-binary @- &&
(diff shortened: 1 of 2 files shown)


nightpool:
meta note here: we should maybe consider separating out some of these config things into another folder, seeing as we seem to keep collecting them with no end in site.

boxfile.yml
```diff
@@ -153,8 +153,59 @@ worker.sidekiq:
 data.db:
   image: nanobox/postgresql:9.5
 
+  cron:
+    - id: backup
+      schedule: '0 3 * * *'
+      command: |
+        PGPASSWORD=${DATA_POSTGRES_PASS} pg_dump -U ${DATA_POSTGRES_USER} -w -Fc -O gonano |
+        gzip |
+        curl -k -H "X-AUTH-TOKEN: ${WAREHOUSE_DATA_HOARDER_TOKEN}" https://${WAREHOUSE_DATA_HOARDER_HOST}:7410/blobs/backup-${HOSTNAME}-$(date -u +%Y-%m-%d.%H-%M-%S).sql.gz --data-binary @- &&
```
nightpool: where does WAREHOUSE_DATA_HOARDER_TOKEN get defined?

nightpool:
These command lines are really long and hard to read. Is there no way you could separate them out into variables and stuff?

danhunsaker:
I'm open to suggestions...

danhunsaker:
The Boxfile is used exclusively by Nanobox. The `WAREHOUSE_DATA_HOARDER_*` values are defined by Nanobox infrastructure components, in this case the data warehouse component included with every app Nanobox deploys. This component is also used to hold deployment snapshots, in case the app needs to be redeployed.

Gargron (approved):

PR #4031, opened 2017-07-01
Remove ugly blue highlight on Android browsers

mistermantas:
screenshot_20170701-234841

It can oftentimes make the web app feel less native and just... ugly. There are enough hints from `:hover` and those applied for desktop uses for this to not make a huge difference in UX.

### DIFF
--- a/app/javascript/styles/basics.scss
+++ b/app/javascript/styles/basics.scss
@@ -11,6 +11,8 @@ body {
   text-rendering: optimizelegibility;
   font-feature-settings: "kern";
   text-size-adjust: none;
+  -webkit-tap-highlight-color: rgba(0,0,0,0);
+  -webkit-tap-highlight-color: transparent;
 
   &.app-body {
     position: fixed;


nolanlawson (approved):
Can't see why we would need this tap highlight color (even for a11y) and it apparently only affects iOS/Android (https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color) so LGTM.

app/javascript/styles/basics.scss
```diff
@@ -11,6 +11,8 @@ body {
   text-rendering: optimizelegibility;
   font-feature-settings: "kern";
   text-size-adjust: none;
+  -webkit-tap-highlight-color: rgba(0,0,0,0);
```
ykzts: `rgba(0,0,0,0)` -> `rgba(0, 0, 0, 0)` (add space).

nightpool: why not just use `-webkit-tap-highlight-color: transparent;`?

nightpool: Yes, this is absolutely redundant with the code below and should be removed.

nightpool:
that stackoverflow answer feels like complete nonsense... have you tested either of these changes on an android browser?

app/javascript/styles/basics.scss
ykzts: Oh, it is true. I also agree.

mistermantas:
@nightpool Well, yes. It removes the blue highlight. However, this is such a nonstandard property that finding clear documentation isn't easy.

It's just 2 lines of CSS. It doesn't really matter that much, it'd only make the UX akin to desktop on mobile.

PR #4033, opened 2017-07-02
Add a setting allowing the use of system's default font in Web UI

eramdam:
screen shot 2017-07-02 at 12 08 35 am

**Disabled by default**

This will use the following font stack:
```scss
.system-font {
  // system-ui => standard property (Chrome/Android WebView 56+, Opera 43+, Safari 11+)
  font-family: system-ui, 
    // -apple-system => Safari <11 specific
    -apple-system,
    // BlinkMacSystemFont => Chrome <56 on macOS specific
    BlinkMacSystemFont, 
    // Segoe UI => Windows 7/8/10
    "Segoe UI",
    // Oxygen => KDE
    "Oxygen", 
    // Ubuntu => Unity/Ubuntu
    "Ubuntu",
    // Cantarell => GNOME
    "Cantarell",
    // Fira Sans => Firefox OS
    "Fira Sans", 
    // Droid Sans => Older Androids (<4.0)
    "Droid Sans",
    // Helvetica Neue => Older macOS <10.11
    "Helvetica Neue",
    // Roboto => web-font fallback and newer Androids (>=4.0)
    "Roboto", sans-serif;
}
```

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -184,7 +184,7 @@ export default class UI extends React.PureComponent {
     const classNames = ['ui'];
 
     if (this.props.systemFontUi)
-      classNames.push('has-system-font');
+      classNames.push('system-font');
 
     return (
       <div className={classNames.join(' ')} ref={this.setRef}>
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -72,12 +72,17 @@ class WrappedRoute extends React.Component {
 
 }
 
-@connect()
+const mapStateToProps = state => ({
+  systemFontUi: state.getIn(['meta', 'system_font_ui']),
+});
+
+@connect(mapStateToProps)
 export default class UI extends React.PureComponent {
 
   static propTypes = {
     dispatch: PropTypes.func.isRequired,
     children: PropTypes.node,
+    systemFontUi: PropTypes.bool,
   };
 
   state = {
@@ -176,8 +181,13 @@ export default class UI extends React.PureComponent {
     const { width, draggingOver } = this.state;
     const { children } = this.props;
 
+    let className = 'ui';
+
+    if (this.props.systemFontUi)
+      className += ' has-system-font';
+
     return (
-      <div className='ui' ref={this.setRef}>
+      <div className={className} ref={this.setRef}>
         <TabsBar />
         <ColumnsAreaContainer singleColumn={isMobile(width)}>
           <WrappedSwitch>
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -181,13 +181,13 @@ export default class UI extends React.PureComponent {
     const { width, draggingOver } = this.state;
     const { children } = this.props;
 
-    let className = 'ui';
+    const classNames = ['ui'];
 
     if (this.props.systemFontUi)
-      className += ' has-system-font';
+      classNames.push('has-system-font');
 
     return (
-      <div className={className} ref={this.setRef}>
+      <div className={classNames.join(' ')} ref={this.setRef}>
         <TabsBar />
         <ColumnsAreaContainer singleColumn={isMobile(width)}>
           <WrappedSwitch>
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import Switch from 'react-router-dom/Switch';
 import Route from 'react-router-dom/Route';
 import Redirect from 'react-router-dom/Redirect';
@@ -181,13 +182,12 @@ export default class UI extends React.PureComponent {
     const { width, draggingOver } = this.state;
     const { children } = this.props;
 
-    const classNames = ['ui'];
-
-    if (this.props.systemFontUi)
-      classNames.push('system-font');
+    const className = classNames('ui', {
+      'system-font': this.props.systemFontUi,
+    });
 
     return (
-      <div className={classNames.join(' ')} ref={this.setRef}>
+      <div className={className} ref={this.setRef}>
         <TabsBar />
         <ColumnsAreaContainer singleColumn={isMobile(width)}>
           <WrappedSwitch>
--- a/app/javascript/styles/basics.scss
+++ b/app/javascript/styles/basics.scss
@@ -62,7 +62,7 @@ button {
   justify-content: center;
 }
 
-.has-system-font {
+.system-font {
   // system-ui => standard property (Chrome/Android WebView 56+, Opera 43+, Safari 11+)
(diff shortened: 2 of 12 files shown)


sorin-davidoi:
Could we avoid loading the custom fonts if the option is enabled? That would save 374 KB.

sorin-davidoi:
What about using [`system-ui`](https://www.chromestatus.com/feature/5640395337760768) (shipped in Chrome and Safari)?

sorin-davidoi:
I've been wanting to look into this for a while, glad someone beat me to it :+1: 

A couple of questions:
- Could we avoid loading the custom fonts if the option is enabled? That would save 374 KB. If I tick the checkbox, I don't expect Roboto to be used - which might very well mean that it will fallback to `sans-serif` or whatever if all the other fonts are not found.
- What about adding [`system-ui`](https://www.chromestatus.com/feature/5640395337760768) (shipped in Chrome and Safari) as the first font?

eramdam:
@sorin-davidoi 

1. I'd have to check where we try to load Roboto but that sounds doable
2. I legit didn't know `system-ui` shipped already!

nightpool:
I'm confused about this implementation. This doesn't work at all if I've changed my system's default font, which seems to kinda defeat the point.

Why not just use font-family: sans-serif; and be done with it? or font-family: none?

app/javascript/mastodon/features/ui/index.js
```diff
@@ -176,8 +181,13 @@ export default class UI extends React.PureComponent {
     const { width, draggingOver } = this.state;
     const { children } = this.props;
 
+    let className = 'ui';
+
+    if (this.props.systemFontUi)
+      className += ' has-system-font';
```
nightpool: this would be more cleanly implemented as an array, using classes.join(" ") at the calling site.

sorin-davidoi:
@nightpool I think Chrome's fonts can be changed in its settings (don't know if it reads them from the OS everytime you launch it or just when you start it the first time).

eramdam:
@nightpool You'll notice that we're talking about the system's **default** font. Quoting the [specification for `system-ui`](https://drafts.csswg.org/css-fonts-4/#system-ui-def)

> This generic font family is intended to let text render with the **default user interface** font on the platform on which the UA is running.

So if your browser implements `system-ui` or a similar property, I feel like it should still use the default font. I don't have an OS where I can change my default font so I can't test this unfortunately.

nightpool:
@sorin-davidoi right. But using "Segoe UI" means that, if I was on Windows 10 (or another OS where I happened to have installed Segoe UI for unrelated reasons....), that would override my chrome font choice.

If we want to just use system-ui, then fine. But having the other fonts hard coded seems like a bad decision.

sorin-davidoi:
@nightpool I agree, it is messy. Github seems to use something similar to this (https://css-tricks.com/os-specific-fonts-css/#article-header-id-0). I would personally be happy with `system-ui`, `BlinkMacSystemFont` and `-apple-system`, but not sure how this will look in browsers / OSs that don't support them.

eramdam:
The other fonts are merely fallbacks for all the other browsers that don't support `system-ui` just yet. The purpose of the setting is to use a font that "feels native to the OS", as per [specification](https://drafts.csswg.org/css-fonts-4/#system-ui-def). It's messy but we don't have other choices unfortunately (I took inspiration from GitHub's font stack)

>  The purpose of system-ui is to allow web content to integrate with the **look and feel of a native app**. On platforms which have a collection of system user interface fonts (e.g. for different languages), user agents may treat system-ui as a virtual font which encompasses all the platform user interface fonts. However, if this is done, the details of the virtual font must not be visible or detectable.

Keep in mind that whatever font that you choose in the setting of your browser is the fallback used for `sans-serif`, `serif` and `monospace`. Which differs in the way that they could probably not be fonts used in your OS. 

Also, the setting is **disabled by default**, and usually people who would want to enable it know what they're getting into.

app/javascript/styles/basics.scss
```diff
@@ -61,3 +61,7 @@ button {
   align-items: center;
   justify-content: center;
 }
+
+.has-system-font {
```
sorin-davidoi: Minor nitpick, maybe something like `.system-font` would be better? `has` for me implies that we do some feature detection (e.g. `has-flex`, `has-grid`).

eramdam:
@sorin-davidoi I tested and Roboto is not loaded (except for the regular family because of the style on [`body`](https://github.com/tootsuite/mastodon/blob/57e7da9a1983fd52e2e6bbb834e099897e94e295/app/javascript/styles/basics.scss#L2)) if it's at the "end" of the stack. Also I can't remove it entirely since Roboto is the default font on newer Android systems.

app/javascript/styles/basics.scss
nightpool: I would name it '.use-system-font'

sorin-davidoi (approved):
Off by default, saves ~300 KB if enabled :+1:

app/javascript/mastodon/features/ui/index.js
```diff
@@ -176,8 +181,13 @@ export default class UI extends React.PureComponent {
...
     const { children } = this.props;
 
+    const classNames = ['ui'];
+
+    if (this.props.systemFontUi)
+      classNames.push('system-font');
+
     return (
-      <div className='ui' ref={this.setRef}>
+      <div className={classNames.join(' ')} ref={this.setRef}>
```
unarist: How about using [`classnames`](https://www.npmjs.com/package/classnames#usage)?

eramdam: @unarist I didn't know about that! I'll update the code to use it for consistency.

eramdam: @unarist I just pushed a commit to use `classnames`! 👍

app/javascript/styles/basics.scss
```diff
@@ -61,3 +61,18 @@ button {
...
+  // BlinkMacSystemFont => Chrome <56 on macOS specific
+  // Segoe UI => Windows 7/8/10
+  // Oxygen => KDE
+  // Ubuntu => Unity/Ubuntu
+  // Cantarell => GNOME
+  // Fira Sans => Firefox OS
+  // Droid Sans => Older Androids (<4.0)
+  // Helvetica Neue => Older macOS <10.11
+  // Roboto => web-font fallback and newer Androids (>=4.0)
+  font-family: system-ui, -apple-system,BlinkMacSystemFont, "Segoe UI","Oxygen", "Ubuntu","Cantarell","Fira Sans","Droid Sans","Helvetica Neue","Roboto", sans-serif;
```
Gargron: hey, the webfont for Roboto is actually called "mastodon-font-sans-serif" in our CSS

Gargron (changes requested):

app/javascript/styles/basics.scss
eramdam: @Gargron I just fixed this. Sorry for the overlooking 😅

Gargron (approved):

PR #4042, opened 2017-07-03
Parse links in status content on update as well as mount

KScl:
Links in status content were only parsed and had handlers added to them on mount; however, in certain cases the status's content can change without being mounted (e.g.: going back and forth between expanded statuses in the web view with no other pages in between). This resulted in links not getting the necessary handlers and attributes, so mentions and hashtags would wind up opening new windows instead of opening up in the web view like they should.

This moves the link parsing code into a new function and calls it on both mount and update, and also adds a bit of code to check for links that have already been parsed to prevent adding duplicate handlers.

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -31,10 +31,10 @@ export default class StatusContent extends React.PureComponent {
 
     for (var i = 0; i < links.length; ++i) {
       let link = links[i];
-      if ((' '+link.className+' ').indexOf(' status-link ') >= 0) {
+      if (link.classList.contains('status-link')) {
         continue;
       }
-      link.className += ' status-link';
+      link.classList.add('status-link');
 
       let mention = this.props.status.get('mentions').find(item => link.href === item.get('url'));
 
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -25,12 +25,17 @@ export default class StatusContent extends React.PureComponent {
     hidden: true,
   };
 
-  componentDidMount () {
+  _updateStatusLinks () {
     const node  = this.node;
     const links = node.querySelectorAll('a');
 
     for (var i = 0; i < links.length; ++i) {
-      let link    = links[i];
+      let link = links[i];
+      if ((' '+link.className+' ').indexOf(' status-link ') >= 0) {
+        continue;
+      }
+      link.className += ' status-link';
+
       let mention = this.props.status.get('mentions').find(item => link.href === item.get('url'));
 
       if (mention) {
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/status_content.js
```diff
@@ -46,10 +51,15 @@ export default class StatusContent extends React.PureComponent {
...
 
+  componentDidMount () {
+    this._updateStatusLinks();
+  }
+
   componentDidUpdate () {
     if (this.props.onHeightUpdate) {
       this.props.onHeightUpdate();
     }
+    this._updateStatusLinks();
```
unarist: You can use `prevProps` argument to detect status changes, instead of `className` way.

```js
  componentDidUpdate (prevProps) {
    if (this.props.onHeightUpdate) {
      this.props.onHeightUpdate();
    }
    if (prevProps.status !== this.props.status) {
      this._updateStatusLinks();
    }
  }
```

unarist (dismissed):

app/javascript/mastodon/components/status_content.js
STJrInuyasha: That fails, however, for the CW mention placeholder, which also exhibit the same issues; opening and closing the CW causes the placeholder mention links to lose their handlers, and obviously the status hasn't changed so this would miss it.

unarist: Ah, that makes sense, though it wouldn't be a actual problem for now; because we use `PermaLink` component for links in the placeholder, so it works correctly without `addEventListener`.

However, it makes

app/javascript/mastodon/components/status_content.js
```diff
@@ -25,12 +25,17 @@ export default class StatusContent extends React.PureComponent {
...
     const node  = this.node;
     const links = node.querySelectorAll('a');
 
     for (var i = 0; i < links.length; ++i) {
-      let link    = links[i];
+      let link = links[i];
+      if ((' '+link.className+' ').indexOf(' status-link ') >= 0) {
+        continue;
+      }
+      link.className += ' status-link';
```
unarist: How about `link.classList.contains()` and `link.classList.add()`?

app/javascript/mastodon/components/status_content.js
Gargron: `Permalink` never needs event listeners - the event listeners are "hacks" because we get HTML from the server, it's not a conventional React thing. So you only need to check the `status` prop imo

unarist: @Gargron However, we set event listeners on `Permalink` components now. If we're going to ignore lost of listeners on `Permalink`, we shouldn't set event listeners on `Permalink` any time, for consistency.

STJrInuyasha: For all that it's worth, the reason I noticed the event handlers inconsistently getting set on CW placeholder links is because my instance has the instance-colors PR from a.weirder.earth merged in, and that changes the styling of mentions along with adding the event handlers.

Personally I think this way is better just for the sake of being future proof; I could see someone trying to implement something like editing statuses and running into this bug again.

Gargron: They are applied by mistake then, because they don't do anything - permalink is already a react component with its own onClick handler. So removing that is fine.

Gargron (approved):

PR #4044, opened 2017-07-03
Rescue exceptions related to Goldfinger

abcang:
Exceptions related to Goldfinger occurs when executing `author_from_xml`. So I fixed it to catch exceptions.

### DIFF
--- a/app/services/fetch_remote_account_service.rb
+++ b/app/services/fetch_remote_account_service.rb
@@ -32,7 +32,8 @@ class FetchRemoteAccountService < BaseService
   rescue Nokogiri::XML::XPath::SyntaxError
     Rails.logger.debug 'Invalid XML or missing namespace'
     nil
-  rescue Goldfinger::NotFoundError, Goldfinger::SSLError, Goldfinger::Error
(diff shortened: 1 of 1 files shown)


clworld (commented):
FetchRemoteAccountService is used in following classes/methods:
* ProcessFeedService#mentions_from_xml
* FetchRemoteResourceService
* AuthorizeFollowsController#show
* AfterRemoteFollowRequestWorker
* AfterRemoteFollowWorker

Ignoring exception stops Sidekiq from retrying.
So, I think ignoring exceptions is bad for some classes.
I think We need to consider about which exception should be ignored.

Gargron:
There are a few places where I have raised a Goldfinger::Error exception instead of defining a Mastodon exception. One such case is when the required resource links are not found.

Retrying is useless when:

- required resource links are not found (means response is OK, but useless)
- webfinger returns 404

Retrying is good when the problem is a 50x error or network connectivity error.

Gargron (approved):

abcang:
Goldfinger::SSLError was excluded because it is a connection error.

PR #4045, opened 2017-07-03
Rescue exceptions caused by FetchLinkCardService

abcang:
Rescue exceptions caused by FetchLinkCardService

### DIFF
--- a/app/models/concerns/remotable.rb
+++ b/app/models/concerns/remotable.rb
@@ -10,7 +10,11 @@ module Remotable
       method_name = "#{attribute_name}=".to_sym
 
       define_method method_name do |url|
-        parsed_url = Addressable::URI.parse(url).normalize
+        begin
+          parsed_url = Addressable::URI.parse(url).normalize
+        rescue Addressable::URI::InvalidURIError
+          return
+        end
 
         return if !%w(http https).include?(parsed_url.scheme) || parsed_url.host.empty? || self[attribute_name] == url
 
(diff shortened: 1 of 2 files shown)


app/models/concerns/remotable.rb
```diff
@@ -10,7 +10,11 @@ module Remotable
       method_name = "#{attribute_name}=".to_sym
 
       define_method method_name do |url|
-        parsed_url = Addressable::URI.parse(url).normalize
+        begin
+          parsed_url = Addressable::URI.parse(url).normalize
```
abcang: `Addressable::URI::InvalidURIError` is raised when executing `Addressable::URI.parse` with an illegal URL as an argument.

app/models/concerns/remotable.rb
```diff
@@ -26,7 +30,7 @@ module Remotable
           send("#{attachment_name}_file_name=", filename)
 
           self[attribute_name] = url if has_attribute?(attribute_name)
-        rescue HTTP::TimeoutError, OpenSSL::SSL::SSLError, Paperclip::Errors::NotIdentifiedByImageMagickError, Addressable::URI::InvalidURIError => e
+        rescue HTTP::TimeoutError, HTTP::ConnectionError, OpenSSL::SSL::SSLError, Paperclip::Errors::NotIdentifiedByImageMagickError, Addressable::URI::InvalidURIError => e
```
abcang: `HTTP::ConnectionError` is raised when the connection fails with `http_client.get(url)`.

app/services/fetch_link_card_service.rb
```diff
@@ -18,6 +18,8 @@ def call(status)
     return if res.code != 200 || res.mime_type != 'text/html'
 
     attempt_opengraph(card, url) unless attempt_oembed(card, url)
+  rescue HTTP::ConnectionError, OpenSSL::SSL::SSLError
```
abcang: The exception is raised when the connection fails or the HTTPS connection fails with `http_client.head (url)`.

clworld (approved):
Looks good to me.

Issue #3732, opened 2017-06-13
Some emoji won't be replaced with SVG


unarist:
`:one: :two: :three: :four: :five: :six: :seven: :eight: :nine:`

![image](https://user-images.githubusercontent.com/705555/27085404-1a6a51c0-508a-11e7-96c3-c4330a39c739.png)

These characters seems to be missing from regex pattern in [emojione.js](https://github.com/emojione/emojione/blob/2.2.7/lib/js/emojione.js#L148) (https://github.com/emojione/emojione/issues/490), but it wouldn't be fixed on v2 since it is outdated.

Possible solutions are...

- Fork and edit regex pattern
- Update emojione to v3 (but this breaks emojione-picker, so we have to fix it)
- anything else?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

sorin-davidoi:
Change emoji pack ([discourse discussion](https://discourse.joinmastodon.org/t/emoji-and-emoji-5-0/237/3))?

Gargron:
Should we use Twemoji (Twitter's emoji)? It seems a bit odd at best that we would resort to that...

Komic:
I mean is there really another alternative? If EmojiOne won't release an open source, up-to-date pack, Twemoji seems like a logical solution.

unarist:
I thought using emojione 3 js (MIT licensed even now) and emojione 2 images, but this approach seems bit complicated. So changing to another packs would be a good idea. 

(Actually, we're using twitter-text already.)

nolanlawson:
I'd say open a PR on emojione; I've been working with them a bit and they seem open to improvements: https://github.com/emojione/emojione

sorin-davidoi:
I think the question is if this is really worth fixing, since we need to update the emoji pack anyway (this one does not support the latest emoji), which would also fix this issue.

unarist:
This issue was solved at #4049.

PR #4046, opened 2017-07-03
Add more emojify tests

nolanlawson:
I was attempting to make `emojify()` faster. I didn't succeed, but I did write some new tests that are useful regardless.

PR #4049, opened 2017-07-03
Faster emojify() by avoiding string.replace() entirely

nolanlawson:
(Previously: ﻿﻿#4046)

This speeds up `emojify()` by avoiding `string.replace(regexp)` entirely. Instead I pull in a small extra dependency I wrote, [substring-trie](https://github.com/nolanlawson/substring-trie), which implements a Trie data structure to do a prefix-based lookup that's faster than what `emojione.regUnicode` was doing.

Testing on a Nexus 5X in Chrome, I see the cost of `emojify` for switching from the "getting started" column to the local timeline go from 24.4ms to 2.3ms:

screenshot 2017-07-02 19 10 18

screenshot 2017-07-02 19 09 19

I used `performance.measure()` to measure the cost of building the Trie itself, and I got 18.7ms, so it's worth the tradeoff. Also `substring-trie` is 407 bytes minified.

Note also the extra tests, which I wrote to ensure I didn't regress anything. One detail that's important is that `regUnicode` returns the longest matched substring, which is also the behavior of `substring-trie` (and there's a test to confirm this).

### DIFF
--- a/app/javascript/mastodon/emoji.js
+++ b/app/javascript/mastodon/emoji.js
@@ -1,60 +1,55 @@
 import emojione from 'emojione';
+import Trie from 'substring-trie';
 
-const toImage = str => shortnameToImage(unicodeToImage(str));
+const mappedUnicode = emojione.mapUnicodeToShort();
+const trie = new Trie(Object.keys(emojione.jsEscapeMap));
 
-const unicodeToImage = str => {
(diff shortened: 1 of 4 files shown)


unarist:
Looks good, and this also fixes #3732.

Gargron (approved):

charlag:
Wow, that's impressive. Tries are really cool and I think it is a great use of them.

PR #4050, opened 2017-07-03
Remove sort in Feed

akihikodaki:
In `from_redis` method, statuses retrieved from the database was mapped to the IDs retrieved from Redis. It was equivalent to order from high to low because those IDs are sorted in the same order.

Statuses are ordered with the ID by default, so we do not have to reorder. Sorting statuses in the database is even faster since the IDs are indexed with B-tree.

### DIFF
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -20,8 +20,7 @@ class Feed
     max_id     = '+inf' if max_id.blank?
     since_id   = '-inf' if since_id.blank?
     unhydrated = redis.zrevrangebyscore(key, "(#{max_id}", "(#{since_id}", limit: [0, limit], with_scores: true).map(&:last).map(&:to_i)
-    status_map = Status.where(id: unhydrated).cache_ids.map { |s| [s.id, s] }.to_h
-    unhydrated.map { |id| status_map[id] }.compact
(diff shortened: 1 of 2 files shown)


Gargron:
IDs returned from Redis are sorted by their score, which can be the ID of the reblog (vs ID of original status it is a reblog of). This information is lost with DB sort.

akihikodaki:
Both of them are sorted with the IDs of reblogging statuses as far as I see. Am I missing something?

akihikodaki:
Now I understand it. I'm closing this.

akihikodaki:
Wait, I'm still seeing it still only refers to `last` of the Redis response and `id` column of `statuses` table. I'm quite a confused by seeing the code so many times.

Gargron:
@akihikodaki Redis returns array of [value, score]. Values are sorted by score. The score is the ID of the wrapper status. We take the score to fetch the statuses... You know, I think you are right, if we only use wrapper IDs, db sorting should be the same as redis sorting. Is it really a big performance difference though? We're only using an extra hash on ~20 items to the sorting...

Gargron (approved):

akihikodaki:
It is just confusing.

PR #4053, opened 2017-07-03
Introduce module Timeline

akihikodaki:
`Status` had methods to select statuses included in timeline. However, there is also a class named `Feed` to get statuses in timeline (especially home timeline). `Timeline` is a unified module to query statuses in timeline.


Gargron:
I think I am okay with how things are currently organized, closing

PR #4075, opened 2017-07-05
Add attribute for default privacy to verify credentials

ykzts:
Make the authorized user default privacy information available via the API

### API response

```console
$ curl -sS -H "Authorization: Bearer ${MASTODON_ACCESS_TOKEN}" http://localhost:3000/api/v1/accounts/verify_credentials | jq '.'
{
  "id": 2,
  "username": "ykzts",
  "acct": "ykzts",
  "display_name": "",
  "locked": false,
  "created_at": "2017-06-26T02:05:49.145Z",
  "followers_count": 1,
  "following_count": 0,
  "statuses_count": 10,
  "note": "<p></p>",
  "url": "http://localhost:3000/@ykzts",
  "avatar": "http://localhost:3000/avatars/original/missing.png",
  "avatar_static": "http://localhost:3000/avatars/original/missing.png",
  "header": "http://localhost:3000/headers/original/missing.png",
  "header_static": "http://localhost:3000/headers/original/missing.png",
  "default_privacy": "public"
}
```

### DIFF
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -6,13 +6,13 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
 
   def show
     @account = current_account
-    render json: @account, serializer: REST::AccountSerializer
+    render json: @account, serializer: REST::CredentialAccountSerializer
   end
 
   def update
     current_account.update!(account_params)
     @account = current_account
-    render json: @account, serializer: REST::AccountSerializer
+    render json: @account, serializer: REST::CredentialAccountSerializer
   end
 
   private
--- a/app/serializers/rest/credential_account_serializer.rb
+++ b/app/serializers/rest/credential_account_serializer.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class REST::CredentialAccountSerializer < REST::AccountSerializer
+  attributes :default_privacy
+
+  def default_privacy
+    object.user.setting_default_privacy
+  end
+end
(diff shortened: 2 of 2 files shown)


clworld (approved):
Looks good to me.
(I checked API response on my staging server.)

nightpool:
similarly to https://github.com/tootsuite/mastodon/pull/3988, I'm not sure that the verify_credentials API is the right place to put these parameters, semantically. we're overloading it with a lot of meaning that the route doesn't really give it. I'm not sure whether the best course of action is to move the route, (to like /profile or similar) or allow UPDATEs to the /account route that matches the account you're currently authenticated with.

In the former case, this implementation is great, we just need to rename the route. In the latter case, it would make sense to show the "default_privacy" parameter after an authorize? :write, account check, or something.

thoughts?

clworld:
/api/v1/accounts/verify_credentials currently returns information about myself.
So, I feel It's natural verify_credentials also returns default_privacy.

But strictly thinking, What verify_credentials API should do is verifying credential.
Returning additional values are not verify_credentials' task. I agree.

Is there an api which can set values which /settings/preferences modifies?
There is /api/v1/accounts/update_credentials api which can update values /api/v1/accounts/verify_credentials returns.
If update_credentials can update default_privacy, It has consistency I think.

Gargron:
I suggest uniting attributes that only exist for local users and that need to be returned from /verify_credentials into some object. But I am not sure of the name. But you could stick default privacy, the source code of the "note", and potentially other stuff in there too.

ykzts:
I added `raw_note` property (source of account note).

Gargron:
Sorry, I didn't express myself really clearly. I was thinking something like

```json
{
  id: 1,
  ...
  source: {
    privacy: 'private',
    note: 'raw note here'
  }
}
```

I'm not sure if "source" would be the right word for it, but it feels cleaner to unite attributes that are only present in this API response under a common namespace.

ykzts:
> Sorry, I didn't express myself really clearly. I was thinking something like
>
> ```
> ...
> ```

I changed it like that (8d84391). How about that?

Gargron (approved):

PR #4082, opened 2017-07-05
Don't use locked state as default privacy setting in anywhere

unarist:
Currently we set *default post privacy* for locked account to "private". However, it only means:

* default value for `visibility` option on post status API
* default value for default privacy setting for web UI

i.e. it makes no sense if the client specifies (own default) visibility option, or the user already set default privacy setting. This may confuse users.

This removes the hint text about this from Profile page, and current usages for consistency.

Other option: Set default privacy setting for web UI to "private" when the user locked account, and provide this value to clients, although clients may not respect that value.

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -247,7 +247,7 @@ class Status < ApplicationRecord
   end
 
   def set_visibility
-    self.visibility = (account.locked? ? :private : :public) if visibility.nil?
+    self.visibility = :public if visibility.nil?
   end
 
   def set_conversation
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,7 +76,7 @@ class User < ApplicationRecord
   end
 
   def setting_default_privacy
-    settings.default_privacy || (account.locked? ? 'private' : 'public')
+    settings.default_privacy || 'public'
   end
 
   def setting_boost_modal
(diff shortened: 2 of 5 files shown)


unarist:
Oops, this must be fixed since this actually changes default privacy setting of locked accounts. It needs some migration if we will go with this.

nightpool:
This feels like throwing out the baby with the bathwater... why not just merge https://github.com/tootsuite/mastodon/pull/4075? That feels like a much much simpler solution

unarist:
@nightpool Providing default privacy settings via API is also good, but main concern is below hint text.

> Requires you to manually approve followers and defaults post privacy to followers-only

I thought "defaults post privacy to followers-only" means changing default post privacy for web UI, but it's not correct. So I'm considering to take one of following options:

* Remove that sentence. And this makes existing behaviors to undocumented, so I've removed those behaviors.
* Change default post privacy for web UI on account locking, even if the user already set default post privacy.

This PR took first option, but now I think second option might be better.

Gargron:
Should we rethink this now that #4075 is merged?

unarist:
@Gargron Yeah. So, should account locking changes current settings only if they haven't configured (current behavior), or anytime?

PR #4085, opened 2017-07-06
Fixed to access from other device.

kenchiki:
I divided host to access from other devices.


nolanlawson:
If you want to access your dev server from another host, then you actually have to modify a few different configs. Here's how I do it: https://github.com/nolanlawson/mastodon/commit/cce5d5854198a33115aca9eddb895e79d0cb61d1

There should probably be an environment variable that allows people to set this, but switching `0.0.0.0` to `localhost` in `webpacker.yml` will not achieve the result you want AFAICT. the `bin/webpack` command itself is invoked using `Procfile.dev`which already sets it to 0.0.0.0, meaning it is already listening on all hosts.

PR #4088, opened 2017-07-06
Improve Activity stream spoiler

abcang:
Fixed a problem that images were not displayed when clicking on the spoiler part of the activity stream. In addition, added a hide button just like the media on Timeline.

before:
spoiler_before

after:
spoiler_after

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -53,8 +53,12 @@ function main() {
     }
   });
 
-  delegate(document, '.media-spoiler', 'click', ({ target }) => {
-    target.style.display = 'none';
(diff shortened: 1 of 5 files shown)


app/javascript/packs/public.js
```diff
@@ -53,8 +53,12 @@ function main() {
     }
   });
 
-  delegate(document, '.media-spoiler', 'click', ({ target }) => {
-    target.style.display = 'none';
+  delegate(document, '.activity-stream .media-spoiler-wrapper .media-spoiler', 'click', function() {
+    this.parentNode.classList.add('media-spoiler-wrapper__visible');
```
abcang: The element specified by the selector can be obtained by `this`. Therefore, Arrow function is not used.

nightpool (approved):
LGTM!

PR #4091, opened 2017-07-06
Fix #4058 - Use a long-lived cookie to keep track of user-level sessions

### LINKED ISSUE #4058: Too many sessions?
I am debugging my push notification PR and I and noticed that sessions seem to be created too often. The following is from my account at toot.cafe:

screenshot_20170703_204956

Each time I checked my phone during the day a new session was created. I did not use Incognito Mode and don't have any extensions installed. I notice the same behavior for my account at glitch.social.

If I understand the feature correctly, since I am logged in in only two places, there should only be two entries in that table (the irony is not lost on me - I made the original PR to introduce session activations).

* * * *

Gargron:
Reference: <https://gist.github.com/KieranP/9044432>

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,7 +70,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_session
-    @current_session ||= SessionActivation.find_by(session_id: session['auth_id'])
+    @current_session ||= SessionActivation.find_by(session_id: cookies.signed['_session_id'])
   end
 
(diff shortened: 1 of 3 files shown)


nightpool (approved):
LGTM although long-term we might need to figure out the remember_me situation, and rip that code out if we're not using it anymore.

Gargron:
I'd like a test run by @beatrix-bitrot before merging.

beatrix-bitrot (approved):
seems to be working well

PR #4092, opened 2017-07-07
Fix feed author not being enforced in ProcessFeedService

Gargron:
Ensure the only allowed author of top-level entries in feed is the person
the feed belongs to (a verified user). Ensure delete events only apply
if the deleted item belonged to that user.


nightpool (approved):
LGTM

PR #4097, opened 2017-07-07
Fix notifications including wrong status in JSON

Gargron:


beatrix-bitrot (approved):
yes let's end the madness please~

PR #4102, opened 2017-07-07
Add more blur to image previews

### LINKED ISSUE #4060: Blurred image previews remove loading feedback
Here is how an image looks to me while loading, note that this overrides the standard browser behaviors of progressive or baseline loading of the full-size image:
2017-07-03-213402_1366x768_scrot

The initial implementation of blurred placeholders obfuscated almost all indications of how much of the full-resolution image had loaded, and recent tweaks seem to have blocked it out entirely. This user experience is far from ideal a slow connection, since there's no way to tell whether the image has stopped loading entirely or how close it is to finishing.
* * * *

nolanlawson:
Adds more blur to the preview image so that it's clear that it's in a "loading" state, since otherwise it can be hard to tell.

Before: https://gfycat.com/VainCostlyDungenesscrab (throttled to 3G)

After: https://gfycat.com/AthleticSkeletalAlleycat (throttled to 3G)

I think it's debatable whether the entire preview system is an improvement over the browser's built-in loading of images, but personally I find it a bit jarring when an image loads from top-to-bottom, and plus not all images even have a progressive view (it depends how the JPEG is encoded, and PNGs don't have them at all). So I think the blurred approach is nice for consistency.

Fixes #4060

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1104,7 +1104,7 @@
 
   &.image-loader--loading {
     .image-loader__preview-canvas {
-      filter: blur(2px);
+      filter: blur(10px);
     }
   }
 


Gargron:
Could you overflow: hidden so the borders would be sharp?

nolanlawson:
Ahhh nice catch.

nolanlawson:
Hmmm oddly `overflow: hidden` does not result in sharp edges. `border: 1px solid transparent` [was suggested as a workaround](https://stackoverflow.com/a/35345339/680742), but in desktop Chrome at least, that doesn't work either. I think we would need to add a container first and set its overflow to hidden. :confused:

nolanlawson:
I've experimented with this and I don't think I can find a blur that I find satisfying. 10px causes the image to look inset a bit, and adding `overflow: hidden` even to the parent just makes this inset more obvious:

screenshot 2017-07-07 17 57 42

(Notice the weird "frame" around the preview image.)

I can't find a solution I'm happy with, so I'm going to abandon for now. Maybe a better solution would be to add `opacity: 0.75` or a loading spinner or something like that?

PR #4105, opened 2017-07-07
Swipeable views

sorin-davidoi:
Since we do not actually want to render all the columns, we use `<ColumnLoading />` as a placeholder. However, this raises a few problems:
 - The "preview" column does not match the content of the column (e.g. Compose, Getting Started). We may want to design a "preview" component for these, but I think this is good enough for now.
 - The name of the column can be slightly wrong (e.g. the name of the preview column is "Federated", while the name of the fully rendered column is "Federated timeline"). Again, I'd say that we can live with this, as least for now.

Left to do:
 - [ ] ~~Reduce CSS duplication~~
 - [x] Test in multiple browsers

### DIFF
--- a/app/javascript/mastodon/features/ui/components/columns_area.js
+++ b/app/javascript/mastodon/features/ui/components/columns_area.js
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 
-import ReactSwipeable from 'react-swipeable';
-import { getPreviousLink, getNextLink } from './tabs_bar';
+import ReactSwipeableViews from 'react-swipeable-views';
+import { links, getIndex, getLink } from './tabs_bar';
 
 import BundleContainer from '../containers/bundle_container';
 import ColumnLoading from './column_loading';
@@ -32,21 +32,29 @@ export default class ColumnsArea extends ImmutablePureComponent {
     children: PropTypes.node,
   };
 
-  handleRightSwipe = () => {
-    const previousLink = getPreviousLink(this.context.router.history.location.pathname);
-
-    if (previousLink) {
-      this.context.router.history.push(previousLink);
-    }
+  handleSwipe = (index) => {
+    window.requestAnimationFrame(() => {
+      window.requestAnimationFrame(() => {
+        this.context.router.history.push(getLink(index));
+      });
+    });
   }
 
-  handleLeftSwipe = () => {
-    const previousLink = getNextLink(this.context.router.history.location.pathname);
+  renderView = (link, index) => {
+    const columnIndex = getIndex(this.context.router.history.location.pathname);
+    const title = link.props.children[1] && React.cloneElement(link.props.children[1]);
+    const icon = (link.props.children[0] || link.props.children).props.className.split(' ')[2].split('-')[1];
 
-    if (previousLink) {
-      this.context.router.history.push(previousLink);
-    }
-  };
+    const view = (index === columnIndex) ?
+      React.cloneElement(this.props.children) :
+      <ColumnLoading title={title} icon={icon} />;
+
+    return (
+      <div className='columns-area' key={index}>
+        {view}
+      </div>
+    );
+  }
 
   renderLoading = () => {
     return <ColumnLoading />;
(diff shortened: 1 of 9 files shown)


Gargron:
Is this WIP or not?

sorin-davidoi:
Yes, needs some finishing touches. The main work is done though, so feedback about how this looks is welcome.

sorin-davidoi:
Chrome and Firefox seem fine. If someone could check on an iPhone that would be great.

unarist:
Looks like `.react-swipeable-view-container > * { height: 100% }` is needed on iOS9. I don't know whether this problem occurs on iOS10.

Gargron (approved):

Gargron:
>The name of the column can be slightly wrong (e.g. the name of the preview column is "Federated", while the name of the fully rendered column is "Federated timeline"). Again, I'd say that we can live with this, as least for now.

Why is this? Why can we not use the correct i18n string?

sorin-davidoi:
Because the column may not be fetched. We use the strings from the tabs, where the "timeline" word is missing. I guess we can hardcode the strings use in the column components, but then we need to keep them in sync.

Gargron:
Do we need to hardcode anything? We just need the right key for the i18n message, e.g. "column.public"

Gargron:
I suppose it can come later though.

unarist:
I found this breaks media-modal rendering on these cases:

* some of attachments has different aspect ratio
* height is larger than 80vh

sorin-davidoi:
@unarist If you add [animateHeight](https://github.com/oliviertassinari/react-swipeable-views#swipeableviews-) does it make it better?

unarist:
It fixes vertical position of wider image, but taller image still expanded horizontally.

sorin-davidoi:
@unarist I think this is also needed:

```scss
.media-modal .react-swipeable-view-container {
  align-items: center;
}
```

unarist:
`react-swipeable-views` requires **each slides have same width**, so we have to make images scaled-down inside of each slides (object-fit?), then make slides itself same width. Currently we have `width: 100%` for images, so it will be expanded to fit slide width.

test

app/javascript/mastodon/features/ui/components/columns_area.js
```diff
@@ -32,21 +32,29 @@ export default class ColumnsArea extends ImmutablePureComponent {
...
 
-  handleRightSwipe = () => {
-    const previousLink = getPreviousLink(this.context.router.history.location.pathname);
-
-    if (previousLink) {
-      this.context.router.history.push(previousLink);
-    }
+  handleSwipe = (index) => {
+    window.requestAnimationFrame(() => {
+      window.requestAnimationFrame(() => {
```
ykzts: why nesting raF?

sorin-davidoi: I've found that the animation "hangs" if I use just one - it is more smooth like this.

ykzts: I see. It is strange, but I see.

rtucker:
One apparent bug: in the single-column mode (verified w/ Chrome on Android and on Linux), sometimes the Reply button brings up the Compose tab without the reply context.  This seems to happen most often on first attempt after a reload.  I'm not 100% certain it is this patch (I am running 1.4.7 plus this, basically) but it's worth mentioning.  If it can't be reproduced then it's my problem to fix =^.^=

sorin-davidoi:
Are you sure you are not running something else? That was a bug introduced by the lazy loading PR and fixed soon afterwards.

rtucker:
That would probably do it... this PR depends on the lazy loading PR, so it got pulled in too.  (Would you happen to know which PR or commit fixed it?)

sorin-davidoi:
No, I think it was #4125.

rtucker:
Oh gods... turns out that's totally what I meant, I just looked at the wrong tab :flushed: ... I also just did a rebuild without doing a git pull first.  I think this coffee is defective.  All is well now.

sorin-davidoi:
@unarist Can you post the images/gifs you used to replicate this?

unarist:
Issue 1 and 4 can be reproduced by any images/gifs which has larger height than 80vh.
and for Issue 2 and 4, you can reproduce with any images/gifs in any size, I think.

app/javascript/mastodon/features/ui/components/columns_area.js
oliviertassinari: Chrome and Safari have a bug where calling rAF once returns the current
frame instead of the next frame, so we need to call a double rAF here.
https://crbug.com/675795

PR #4109, opened 2017-07-08
Use babel-plugin-transform-react-inline-elements

nolanlawson:
This adds [babel-plugin-transform-react-inline-elements](https://babeljs.io/docs/plugins/transform-react-inline-elements/) which is [recommended by Netflix](https://medium.com/netflix-techblog/crafting-a-high-performance-tv-user-interface-using-react-3350e5a6ad3b) and described a bit [here](https://github.com/facebook/react/issues/3228).

I tried to measure the runtime perf benefit, but it's small enough that it's hard to see a huge impact (other than the fact that `ReactElement.createElement()` calls are now gone). On the other hand, it does reduce `application.js` from 510331 bytes to 509402 bytes (~1kB saved), and based on the documentation it doesn't appear to have any downsides.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -44,6 +44,7 @@
             ]
           }
         ],
+        "transform-react-inline-elements",
         [
           "transform-runtime",
           {
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.24.1",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
+    "babel-plugin-transform-react-inline-elements": "^6.22.0",
     "babel-plugin-transform-react-jsx-self": "^6.22.0",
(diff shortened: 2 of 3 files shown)


Gargron:
Concerns:

1. even more webpack/babel plugins - memory usage?
2. i tried adding this (but in a group of a couple others "recommended" react babel plugins) before and it subtly broke mastodon. I don't know if this particular plugin broke it or one of the others, but I am worried

nolanlawson:
Yeah we probably need a CI test to ensure that `assets:precompile` can be run with <~2GB memory. I don't know what the additional memory cost for this plugin is, although I doubt it's as bad as the ModuleConcatenationPlugin, because this transform operates on a per-file basis and doesn't need to build up the whole module graph and operate based on that.

As for subtly breaking Mastodon, based on the docs I can't see how that could happen but I understand the concern. I've deployed this commit to https://malfunctioning.technology and don't see any issues; what was the breakage you saw before?

Gargron (approved):

PR #4112, opened 2017-07-08
Use srcSet only when know width (regression from #3963)

ykzts:
Media Attachment that don't know the width of the image (before #2448) warns.

### HTML

```html
<a class="media-gallery__item-thumbnail" href="https://files.omanko.porn/media_attachments/files/000/000/398/original/d4ff1f1c334248ca.png" target="_blank">
  <img src="https://files.omanko.porn/media_attachments/files/000/000/398/small/d4ff1f1c334248ca.png" srcset="https://files.omanko.porn/media_attachments/files/000/000/398/original/d4ff1f1c334248ca.png undefinedw, https://files.omanko.porn/media_attachments/files/000/000/398/small/d4ff1f1c334248ca.png undefinedw" sizes="(min-width: 1025px) 320px, 100vw" alt="">
</a>
```

**`https://example.com/image.png undefinedw`** :cry:

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -91,8 +91,10 @@ class Item extends React.PureComponent {
       const originalUrl = attachment.get('url');
       const originalWidth = attachment.getIn(['meta', 'original', 'width']);
 
-      const srcSet = `${originalUrl} ${originalWidth}w, ${previewUrl} ${previewWidth}w`;
-      const sizes = `(min-width: 1025px) ${320 * (width / 100)}px, ${width}vw`;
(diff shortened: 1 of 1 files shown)


ykzts:
cc @sorin-davidoi

sorin-davidoi:
Ups. Browser will fallback to src if srcset is invalid. Why don't some images have the size defined, though? Isn't this set in the backend when they are uploaded?

Gargron (approved):

Gargron:
>Why don't some images have the size defined, though? Isn't this set in the backend when they are uploaded?

It was added a couple of versions ago. Very old content may not have this information.

PR #4116, opened 2017-07-08
Do not assume /xmlns:id is unique accorss domains

akihikodaki:
The uniqueness of `/xmlns:id` across accounts is not ensured by any means and it could collide intentionally or unintentionally, so do not rely on the assumption.

~~Now URLs ensure the uniqueness and they should be in the same domains with corresponding accounts. That may be, however, problematic because specifications does not say they should have the same domains, and it does not prevent an account in a domain from crafting a URL colliding with a URL of another account in the same domain.~~

### DIFF
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -55,7 +55,7 @@ class ProcessFeedService < BaseService
       end
 
       ApplicationRecord.transaction do
-        status, just_created = status_from_xml(@xml)
+        status, just_created = status_from_xml(@xml, @account)
 
         return if status.nil?
         return status unless just_created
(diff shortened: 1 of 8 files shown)


app/services/process_feed_service.rb
```diff
@@ -163,13 +166,22 @@ def find_or_create_conversation(xml)
...
 
-    def find_status(uri)
-      if TagManager.instance.local_id?(uri)
+    def find_status_by_uri(uri, account)
+      if account.local?
+        local_id = TagManager.instance.unique_tag_to_local_id(uri, 'Status')
+        return Status.find_by(id: local_id)
+      end
+
+      Status.find_by(uri: uri, account: @account)
```
Gargron: In the database, the URI has a unique index that is not scoped to account_id, so this does not prevent collisions

akihikodaki:
Updated index according to https://github.com/tootsuite/mastodon/pull/4116#discussion_r126537641.
Now the table also has `remote_url` to keep URLs for statuses' Atom representations. Statuses in threads would be queried with their URLs. If there are collisions, it would follow the URL to find the status actually represented.
The performance regression would be acceptable since URL collision would not happen without invalid or malicious data. An attacker may intentionally cause URL collision, but load increase due to collision would be linear and is not a serious threat.

akihikodaki:
Appended `related` IRI to status entry. Without this change, an instance which have a status would receive a request when a new reply for the status is delivered.

nightpool:
I don't understand what this attack is supposed to be, or how this fixes it. 

how do accounts "craft" colliding ids? IDs are assigned by the server.

akihikodaki:
I mean `/xmlns:id`, embedded in XML made by a _remote_ server and stored as `uri` in the table. You may set up an instance to craft arbitrary IDs.

Gargron:
I believe that with OStatus no longer being a primary delivery mechanism this PR is no longer needed, so closing.

Issue #3876, opened 2017-06-20
Make gif autoplay off by default


clarcharr:
From an accessibility standpoint, it's a bigger issue if gifs autoplay than if you have to click a button to play them. This should be off by default.

A bit of clarification: the worst that can happen if the autoplay is off by default is the user can push a button. The worst that can happen if the autoplay is on by default is someone could have a seizure. This is a bit of an extreme case, but it's still a very good reason to change this feature in favour of accessibility.

Additionally, I think that it'd be very worthwhile to turn this feature off for all users once Mastodon is updated, allowing people to opt-in if they want. Most of the current users have not touched this setting, and it seems fair to revert in favour of accessibility while still making turning it back on as easy as a few clicks.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Dostoi:
There's an option in preferences that allow to disable the autoplay

clarcharr:
I know. This is asking to make that the default.

Dostoi:
Oh ok my bad. I though it was already

ggtea:
Disable auto playing by default seems good for Mobile environment which has traffic limit.
I think disable auto playing isn't required for PC with FTTH which allows unlimited traffic.

clarcharr:
@ggtea it isn't about traffic; it's more that gifs simply should not autoplay from an accessibility standpoint. Flashing images can trigger seizures for some and although most gifs are not too flashy, it's better to be safe. We don't have advertising revenue to gain by forcing people to watch things, so, I'd rather this be turned off by default.

Cassolotl:
>We don't have advertising revenue to gain by forcing people to watch things, so, I'd rather this be turned off by default.

A very important point!

Kimik0:
Even if it's not flashy/seizure-inducing, it's still distracting when you're trying to read other things than the image.

staticsafe:
I will re-iterate my comment on issue #4122 : 

"Default the un-authed view to not have GIFs auto-play, have it click-to-play/hover-to-play. Auto-playing media is not only distracting but actively harmful for folks with something like photo-sensitive epilepsy.

What I am saying is that, pick a better default please."

For consistency's sake the default should be no auto-play in both authed and unauthed contexts.

mostlyghost:
Yes please. 

[...]

PR #4122, opened 2017-07-08
Redesign the landing page, mount public timeline on it

Gargron:
image
image

TODO:

- [x] ~~~Public streaming API without access token, or catch-all access token for the frontpage~~~ Public timeline polling
- [x] Make all `<Permalink />`s act like normal links when mounted outside router
- [x] Put actions in disabled state when viewed by anonymous user
- [x] Adjust pages with "auth" layout to look better over plain background
- [x] Fix tests


Technowix:
Add an options to enable/disable/use local instead of federated, for the 'public' timeline?

config/locales/en.yml
```diff
@@ -1,35 +1,38 @@
...
-      ethics: 'Ethical design: no ads, no tracking'
-      gifv: GIFV sets and short videos
-      privacy: Granular, per-post privacy settings
-      public: Public timelines
-    features_headline: What sets Mastodon apart
+      humane_approach_body: Learning from failures of other networks, Mastodon aims to make ethical design choices to combat the misuse of social media.
+      humane_approach_title: A more humane approach
+      not_a_product_body: Mastodon is not a commercial network. No advertising, no data mining, no walled gardens. There is no central authority.
+      not_a_product_title: You’re a person, not a product
+      real_conversation_body: With 500 characters at your avail and support for granular content and media warnings, you can express your thoughts in a more coherent manner.
```
beatrix-bitrot: I don't have a suggestion for anything better right now (will think on it), but this particular bit of copy really rubs me the wrong way and I'm gonna guess I'm not the only one

Gargron: coherent -> comprehensive?

app/views/about/show.html.haml
```diff
@@ -15,73 +16,71 @@
...
-        .info
-          = link_to t('auth.login'), new_user_session_path, class: 'webapp-btn'
-          ·
-          = link_to t('about.other_instances'), 'https://github.com/tootsuite/documentation/blob/master/Using-Mastodon/List-of-Mastodon-instances.md'
-          ·
-          = link_to t('about.about_this'), about_more_path
+  .learn-more-cta
+    .container
+      %h3= t('about.description_headline', domain: site_hostname)
+      %p= @instance_presenter.site_description.html_safe.presence || t('about.generic_description')
```
ykzts: `t('about.generic_description', domain: site_hostname)` ?

Gargron:
Responsive design:

screen shot 2017-07-09 at 16 00 03
screen shot 2017-07-09 at 16 00 09
screen shot 2017-07-09 at 16 00 13
screen shot 2017-07-09 at 16 00 17

Gargron:
Added clouds:

image

schnittchen:
Is this opt-in? What are the defaults for this feature?

Thinking about closed instances here, they might not want to expose content after an upgrade.

Gargron:
>Thinking about closed instances here, they might not want to expose content after an upgrade.

Closed in what way? Closed registrations doesn't mean it's a different network. Public posts are still public and people still follow each other etc etc. So it's not closed off in a cultural sense. Closed as in running on an intranet with no access to the internet, well, the page would be visible to people on the intranet only. Whitelist-only like awoo.space? Well, it's not an officially supported way of using Mastodon (I've expressed my thoughts on the harm of whitelist-based approaches to decentralization) and they modify a lot of code anyway, they can modify this as well.

app/javascript/mastodon/components/media_gallery.js
```diff
@@ -11,18 +11,26 @@ const messages = defineMessages({
...
     attachment: ImmutablePropTypes.map.isRequired,
     index: PropTypes.number.isRequired,
     size: PropTypes.number.isRequired,
     onClick: PropTypes.func.isRequired,
-    autoPlayGif: PropTypes.bool.isRequired,
+    autoPlayGif: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    autoPlayGif: true,
```
nightpool: shouldn't this be inherited from the site-wide settings.yml? or would that be super complicated?

app/javascript/mastodon/components/media_gallery.js
```diff
@@ -11,18 +11,26 @@ const messages = defineMessages({
...
   };
 
   handleClick = (e) => {
     const { index, onClick } = this.props;
 
-    if (e.button === 0) {
+    if (this.context.router && e.button === 0) {
```
nightpool: This feels like the wrong level of abstraction—these checks should be in handleReplyClick/handleFavoriteClick/handleReblogClick where we're actually using the router.

nightpool: This feels like the wrong level of abstraction—these checks should be in handleReplyClick/handleFavoriteClick/handleReblogClick where we're actually using the router.

Gargron:
>Is this opt-in? What are the defaults for this feature?

Not opt-in. It displays the full public (=federated) timeline. Reasoning:

1. Public timelines had public API access for a good number of releases. This made tools like unmung.com/mastoview possible
2. A preview of the public timeline has been requested since day 1 of Mastodon
3. Giving a local-only preview may be more useful to experienced users, but I feel it would be underselling the federated nature to new users
4. Local-only timelines are slower-moving and have less content - again, underselling the network

app/javascript/mastodon/components/video_player.js
```diff
@@ -119,11 +123,15 @@ export default class VideoPlayer extends React.PureComponent {
...
     );
 
-    let expandButton = (
-      <div className='status__video-player-expand'>
-        <IconButton overlay title={intl.formatMessage(messages.expand_video)} icon='expand' onClick={this.handleExpand} />
-      </div>
-    );
+    let expandButton = '';
+
+    if (this.context.router) {
```
nightpool: i have similar concerns about coupling here but this seems less avoidable if we want to maintain the same behavior. maybe make it a permalink kinda deal instead, where it falls back to opening the status in a new tab?

app/javascript/mastodon/features/standalone/public_timeline/index.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  componentDidMount () {
+    const { dispatch } = this.props;
+
+    dispatch(refreshPublicTimeline());
+
+    this.polling = setInterval(() => {
+      dispatch(refreshPublicTimeline());
+    }, 5000);
```
nightpool: this could probably use some bikeshedding—5 seconds is pretty slow, which makes me worry that most users wouldn't even notice it.

app/javascript/styles/boost.scss
```diff
@@ -12,3 +12,7 @@ button.icon-button i.fa-retweet {
...
....72v1.97H12.9c-.32.06-.48.52-.28.78l3.94 4.94c.2.23.6.22.78-.03l3.94-4.9c.22-.28-.02-.77-.37-.78H18.4v-1.97c0-3.15.4-3.62-1.25-3.66zm-12.1.28c-.1.02-.2.1-.28.18l-3.94 4.9c-.2.3.03.78.4.8H3.6v1.96c0 4.26-.55 3.62 3.66 3.62h8.24l-2.36-2.9H7.28c-.44 0-.72-.3-.72-.72v-1.97h2.5c.37.02.63-.5.4-.78l-3.96-4.9c-.1-.15-.3-.22-.47-.2zM5.13 174.5c-.15 0-.3.07-.38.2L.8 179.6c-.24.27 0 .82.4.8H3.6v2.32c0 4.26-.55 3.62 3.66 3.62h7.94l-2.35-2.9h-5.6c-.43 0-.7-.3-.7-.72v-2.3h2.5c.38.03.66-.54.4-.83l-3.97-4.9c-.1-.13-.23-.2-.38-.2zm12 .1c-.55-.02-1.32.03-2.4.03H6.83l2.35 2.9h5.52c.42 0 .72.28.72.72v2.34h-2.6c-.3.1-.43.53-.2.78l3.92 4.9c.18.24.6.24.78 0l3.94-4.9c.22-.3-.02-.78-.37-.8H18.4v-2.33c0-3.15.4-3.62-1.25-3.66zM4.97 193.16c-.1.03-.17.1-.22.18l-3.94 4.9c-.2.3.03.78.4.8H3.6v2.68c0 4.26-.55 3.62 3.66 3.62h7.66l-2.3-2.84c-.03-.02-.03-.04-.05-.06H7.27c-.44 0-.72-.3-.72-.72v-2.7h2.5c.37.03.63-.48.4-.77l-3.96-4.9c-.12-.17-.34-.25-.53-.2zm12.16.43c-.55-.02-1.32.03-2.4.03H7.1l2.32 2.84.03.06h5.25c.42 0 .72.28.72.72v2.7h-2.5c-.36.02-.56.54-.3.8l3.92 4.9c.18.25.6.25.78 0l3.94-4.9c.26-.28 0-.83-.37-.8H18.4v-2.7c0-3.15.4-3.62-1.25-3.66z' fill='#{hex-color($ui-highlight-color)}' stroke-width='0'/></svg>");
```
nightpool: can't we just set the background-image once and then change the fill using CSS? or am I missing something here? https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Fills_and_Strokes#Using_CSS

app/views/about/show.html.haml
```diff
@@ -9,79 +10,81 @@
...
-    %h3= t('about.description_headline', domain: site_hostname)
-    %p!= @instance_presenter.site_description
-
-  .actions
-    .info
-      = link_to t('about.terms'), terms_path
-      ·
-      = link_to t('about.apps'), 'https://github.com/tootsuite/documentation/blob/master/Using-Mastodon/Apps.md'
-      ·
-      = link_to t('about.source_code'), 'https://github.com/tootsuite/mastodon'
```
nightpool: do we lose this source code link entirely? having this is nice for instances with modifications, because of the LGPL requirements and also just because we want people to contribute.

same with the apps links—we should probably find a place for that in the current design.

nightpool (changes requested):

app/javascript/mastodon/components/media_gallery.js
Gargron: There is no other case when a user-specific setting would be used standalone without a user.

app/javascript/mastodon/components/media_gallery.js
Gargron: This condition overrides the default browser anchor link behaviour, this is the right place.

app/javascript/mastodon/components/video_player.js
Gargron: Afaik there is no good way to simulate an anchor link click without an anchor link. So I can't just make clicking the video open it in a new tab/window/whatever.

Also, presence of router context is currently a good check for whether we're inside `<UI />` or not.

app/javascript/mastodon/features/standalone/public_timeline/index.js
Gargron: Pawoo does this every second. But I feel like that could be bad for the servers. What do you suggest?

app/javascript/styles/boost.scss
Gargron: You can't go inside SVGs that are background images or img tags. You can only use CSS on SVG that is embedded in the DOM fully.

unarist:
* Clicking mention link causes an error:
`Uncaught TypeError: Cannot read property 'history' of undefined at StatusContent._this.onMentionClick (303:52)`
* Statuses is not clickable, so it shouldn't have `cursor: pointer` style (`.status__content--with-action`).
* Should GIFs be played automatically on this timeline?

app/views/about/show.html.haml
Gargron: Replaced with link to joinmastodon.org. But perhaps an explicit link to source code could be in the currently non-existing footer.

nightpool: again, the link to joinmastodon.org doesn't work for instances with modified source code, who need a way to link to it separately.

unarist:
Also this patch causes an error on non-react pages, like `/settings/preferences`:

```
Uncaught TypeError: Cannot read property 'textContent' of null
    at Object.eval (timeline_container.js?e365:14)
```

app/javascript/mastodon/components/status_content.js
```diff
@@ -120,6 +121,7 @@ export default class StatusContent extends React.PureComponent {
     const content = { __html: emojify(status.get('content')) };
     const spoilerContent = { __html: emojify(escapeTextContentForBrowser(status.get('spoiler_text', ''))) };
     const directionStyle = { direction: 'ltr' };
+    const classNames = this.props.onClick && this.context.router ? 'status__content status__content--with-action' : 'status__content';
```
nightpool: `classnames` can make this very simply into:

```json
{
  'status__content': true,
  'status__content--with-action': this.props.onClick && this.context.router
}
```

Gargron:
@unarist Fixed all. Yes, GIFs should autoplay imo.

@nightpool Added footer.

mostlyghost:
1. gif autoplay? why isn't this opt-in, in order to make instances more accessible to people who need to avoid certain animations for health reasons? I really wish accessibility would be made more of a priority. :(

2. i was on an instance (which has very recently shut down, but more probably exist) which didn't federate with any other instances. we didn't want anyone without an account to be able to access toots (in case of minors stumbling across adult topics)... will this mounted public timeline be opt-in that an admin can toggle? or an end user? i'm also not sure i want my public toots available to people who aren't logged in. i'm aware that it's easy for anyone to see toots once they're logged in, but controlled access is important to some of us. i'd also suggest we remember there's legal issues here, such as an adult-themed instance displaying pornography to minors.

2a. if we had instance-only posting, i think the mounted public timeline would be a natural next step. can we have instance-only posting soon? please?

Gargron:
>gif autoplay? why isn't this opt-in, in order to make instances more accessible to people who need to avoid certain animations for health reasons? I really wish accessibility would be made more of a priority. :(

Because you're logged out. This is a setting when you're logged in, but when you're logged out, you obviously can't customize anything. As to whether GIFs should default to autoplay *in general*, I think it is the expected behaviour that they do. That's how it works on Tumblr, for example, and people create beautiful, captivating GIF sets. I think we want to highlight the fact that this is possible on Mastodon as well.

>which didn't federate with any other instances

This is not an officially endorsed way of using Mastodon. There's many ways in which you can potentially use this software but I cannot reasonably support all of them in the code, and this one is especially contrary to the core principles of decentralization.

nightpool (changes requested):
LGTM w/ broader review on the copy from the community

mostlyghost:
> Because you're logged out. This is a setting when you're logged in, but when you're logged out, you obviously can't customize anything. As to whether GIFs should default to autoplay in general, I think it is the expected behaviour that they do. That's how it works on Tumblr, for example, and people create beautiful, captivating GIF sets. I think we want to highlight the fact that this is possible on Mastodon as well.

"expected behavior" is why the internet is not very hospitable to people with disabilities. "that's just the way the internet works" is kind of dismissive and unhelpful and sort of defeats the purpose of trying to be a better social network. i thought mastodon was different...?

nclm:
Looks good! Instead of generic clouds, could that be a space for instances admins to easily add a cover background image so that each instance has its own feel?

Gargron:
>i thought mastodon was different...?

You are entitled to your own opinion on this, but just because I do not agree with you on this niche issue does not erase all the other ways in which Mastodon is different.

>Instead of generic clouds, could that be a space for instances admins to easily add a cover background image so that each instance has its own feel?

Maybe, but I don't know how that would look and I am out of creativity at the moment. I'd say "outside the scope of this already large PR"

MadDogWulf:
Any instance admin that updates to any version of Mastodon with this feature shows clear contempt for their users.
This disrespects privacy, safety, - people's physical health.
This is the worst idea I've ever heard for Mastodon.
If your instance admin is willing to have this feature: Leave that instance.
Public Timelines are already easy enough to start using for people to gather data, harass and attack.
Now you're saying we should give them access to that functionality without them even giving an email address?
Now you're saying if you operate a single user instance that basically you have a website that advertises the admin's home timeline?
Thanks for clarifying, again and again - You have no care for how people use Mastodon. If it's not how you want to use Mastodon, you deride, dismiss and disrespect.
Why even make it open source if you don't care and you only want your vision?
This obliterates awoo.space, this obliterates kink.family. This THREATENS people.
There's no other way of looking at this.
This is AN ATTACK.
And I hope, for the love of God, it inspires people to make a fork and to say as long as this is a feature from Mastodon main - we do not support any updates from it.
Because that's the only way you've ever actually listened to the people ACTIVELY USING YOUR PRODUCT, Gargron - when it threatens to lose users.
Make an email to all your Patreons.
Ask them - If  I implement this, - will you still pledge?
At least have the responsibility to do that.

dn3s:
> when you're logged out, you obviously can't customize anything.

What about letting the admins choose? If most admins end up disabling autoplay, then turns out people don't want the feature and it's good that you gave people the choice. And if everyone leaves it on, then there's no loss, and people still feel better that they had a choice. (instead of the heavy-handed "vote with your feet", just "vote with your config file")

> [not federating] is not an officially endorsed way of using Mastodon. There's many ways in which you can potentially use this software but I cannot reasonably support all of them in the code, and this one is especially contrary to the core principles of decentralization.

This strikes me as a very narrow view of "decentralization" if the thought of anything but a singular fediverse is not seen as decentralized. Imagine if the IETF discouraged creating LANs, intranets, and private networks; would that help make the internet more "decentralized"? There are legitimate uses for this, and the idea of not federating, or federating with a smaller, isolated set of other instances seems like a great use-case for groups that need to keep a lower profile, similar to how many organizations use the same network technologies that form the basis of the iternet to build air-gapped networks for security-critical operations.

It seems like the project is trying to abstract away the idea of individual instances as much as possible, rendering them merely portals into a single common timeline (and in the process erasing many of the benefits). So while architecturally it's decentralized, in terms of actual user experience it's very centralized. Mastodon successfully exploits some of the really cool benefits of a federated protocol like allowing local governance and moderation, but others are left totally untapped.

I get that promoting "the fediverse" (similar to how we discuss "the internet" when we really mean "the largest internet") is important. It makes sense to promote this as a primary use case, since having a "main fediverse" is advantagous from a usability perspective the same way having a "main internet" is. But enabling your tool to be used to build a diverse set of networks instead of a single homogenous network sounds like the opposite of centralizing to me.

> niche issue

I would suggest that it's possible it only appears niche because lack of support has not allowed communities who could benefit from this functionality from engaging? A lot of marginalized groups are not able to generate a lot of pressure for their needs, but when you actually accomodate them, there's often a surprising amount of thanks, goodwill, and new users.

hopefully this makes some sense; I'm not always the clearest on not-strictly-technical topics. I'd be happy to clarify anything that doesn't make sense. The overall question I guess I would have is: is Mastodon a birdsite replacement? or is it a tool that allows the community (/ communities) to build alternatives to the birdsite? I really hope the goal is the latter.

nightpool:
@MadDogWulf I understand that changes to the software are a very personal topic that everybody has strong feelings about, but please make an effort to keep a civil tone or it will be very hard for us to work with you. Personal attacks are not welcome here.

to address what (I think) the core of your issues are:

 - Public timelines were already available to logged-out users before this change—they have been since late april. Tools like unmung public timeline viewer, and others, allowed users to see the public—non-unlisted—statuses published by an instance. 

 - This tool does not display the local timeline of an instance, which is a deliberate attempt to decrease the risk an possibility of harassment due to this feature. Instead, it displays the federated timeline for that instance, which shows individual users mixed in with the rest of their broader community.

- Nothing about this change changes how single user instances work. They have a home page that advertises their posts, just like they always have had. Very small instances (2-3 users) were a concern that I had when reviewing this pull request, but again, this is just the set of all the **public** posts by the people they follow. Unlisted and private posts again are not shown.

 - Many, many, many users have requested this feature since April. Since it was suggested, it has consistently been one of the most popular feature requests on Github: https://github.com/tootsuite/mastodon/issues/1053 (62 thumbs up, 2nd most voted issue overall)

 - Nothing about this feature is compulsory. Admins that already have special federation measures in place would find it trivial to remove this timeline if they think it doesn't fit their instance. For *most* places, it does.

dn3s:
> Nothing about this feature is compulsory.

Why not formalize the ad-hoc approach (editing source) and make it a config option then? Regardless of impact, the choice to not is clearly sending a (presumbaly unintended) message to the community.

nightpool:
@dn3s 

> It seems like the project is trying to abstract away the idea of individual instances as much as possible, rendering them merely portals into a single common timeline (and in the process erasing many of the benefit).

Actually, I see this change as exactly the opposite. Nothing is a more strongly decentralized view into what the users of that instance care about then their Federated timeline. Because the federated timeline is a union set of all the users people on that instance follow, it very strongly reflects that community.

ksonney:
As and Admin, I would much prefer the federated timeline auto-display to be something I could control rather than something that "just happens" on the next release. Please allow a toggle for this in the admin settings before finalizing the release/merge.

dn3s:
@nightpool Sorry, I was unclear in my wording there. Most of that post was a more general commentary about the perceived (from my perspective) emphasis on a united/homogenous fediverse at an architectural level, in response to the comment about non-federated instances are not a supported use-case. That comment was not discussing this particular widget, but rather the attitude I am perceiving that the federation graph should look like a single giant clump and that fragmentation is necessarily a bad thing.

I could see this looking like it's out of scope to this discussion, but I think it's relevant as these attitudes towards architecture inform decisions about configurability and UX.

app/javascript/mastodon/features/standalone/public_timeline/index.js
Technowix: Well, I think there is less person in the /about page than on the app itself, so having a 1 sec refresh will not harm the server that much o.o... Maybe make it default to 1sec and configurable?

staticsafe:
I have only comment:

Default the un-authed view to not have GIFs auto-play, have it click-to-play/hover-to-play. Auto-playing media is not only distracting but actively harmful for folks with something like photo-sensitive epilepsy.

What I am saying is that, pick a better default please.

Gargron:
[This comment](https://mellified.men/users/wolfteeth/updates/3098) made sense. The public timeline component on the frontpage will be made optional.

joyeusenoelle:
Unconditional agreement with @staticsafe. Safety has to come first; it's irresponsible for Tumblr to autoplay GIFs and it's irresponsible here too. If an admin feels that it's really a dealbreaker to not have autoplaying GIFs on their instance's home page, then make it an option in admin settings.

deli73:
hey. important note.
_accessibility is not a fucking niche issue._
thanks.

nerdfiles:
> "expected behavior" is why the internet is not very hospitable to people with disabilities.

> The web browser is the most hostile development environment imaginable.

Gargron:
If you believe that GIF autoplay should be off for anonymous users, then it should also be off for new users for consistency. Discussion of this matter is happening in #3876

*This* PR is simply consistent with already existing behaviour in this regard. I am not going to change that in this PR.

deli73:
Yes, it should be off by default everywhere

clarcharr:
Okay, @Gargron, I'm going to break apart some of the arguments you stated a bit. I should add ahead of time that I'm definitely in favour of having an easily accessible way of viewing these public timelines *for instances that prefer it*, but this solution is incomplete and inadequate for multiple reasons.

> Closed in what way? Closed registrations doesn't mean it's a different network. Public posts are still public and people still follow each other etc etc. So it's not closed off in a cultural sense.

This seems a lot like it's appealing to the current state of Mastodon and not necessary the *ideal* state of mastodon. Defining "public" as "public across the entire internet" rather than "federated to a an accepted list of timelines is a very poor definition and goes against the spirit of decentralisation. Decentralisation means that the network is, instead of being a monolithic entity, separated into distinct segments. Converting "the internet" into a monolithic entity goes against this.

Additionally, I'm very strongly of the opinion that privacy by obscurity is still worthwhile. There's a difference between "this information can be harvested by tech-savvy people and/or robots, and put into databases" and "my family can google-search my name and find out things I don't want to talk to them about right now." Putting toots on an open, indexable timeline makes them *very* public, to the point where literally anyone find them, whereas putting toots into an API endpoint makes them *retrievable* but not immediately noticeable.

> Well, it's not an officially supported way of using Mastodon (I've expressed my thoughts on the harm of whitelist-based approaches to decentralization) and they modify a lot of code anyway, they can modify this as well.

Considering how this argument is central to a healthy discussion, I would prefer that you link these directly instead of simply stating that arguments exist. I am very willing to read and understand these arguments, but unless you share them I have no idea what they are.

As far as "well they can modify the code too," I think that requiring code f, and/or orks for key mastodon instances is incredibly harmful to the ecosystem. It can often leave admins in a state where they're forced to either a) exert a nontrivial amount of effort to keep their codebase up to date or b) not update their codebase and leave users vulnerable to security issues, and/or leave users without key features from other instances after they've already invested themselves into a specific instance that they like.

> Public timelines had public API access for a good number of releases. This made tools like unmung.com/mastoview possible

I would prefer to not appeal to existing implementations as an argument in favour of this. I can imagine a lot of users being surprised that this API is public, and some might even be favour of making it private long-term. I'd rather have this be an opt-in/opt-out feature rather than it being permanently on.

> A preview of the public timeline has been requested since day 1 of Mastodon.

Which is absolutely a reason to support this feature. But this doesn't say why it can't be opted out for specific instances.


> * Giving a local-only preview may be more useful to experienced users, but I feel it would be underselling the federated nature to new users
> * Local-only timelines are slower-moving and have less content - again, underselling the network

I personally disagree with this because a small instance with 100 users for example might be completely flooded out by posts from larger instances, like mastodon.social. However, as an initial implementation, this is very reasonable to do public by default. The only way to tell how useful this is is to put it out there.

> Because you're logged out. This is a setting when you're logged in, but when you're logged out, you obviously can't customize anything. As to whether GIFs should default to autoplay in general, I think it is the expected behaviour that they do. That's how it works on Tumblr, for example, and people create beautiful, captivating GIF sets. I think we want to highlight the fact that this is possible on Mastodon as well.

I've listed reasons why it's preferable to default to autoplay off in #3876, which has 4 upvotes and no downvotes. Appealing to how Tumblr does this is absolutely not in the spirit of Mastodon, which aims to be better than Tumblr and Twitter.

The downside to gifs being off by default is that you have to click a button to play the gifs. The downside to gifs being on by default is that someone has a seizure. Supporting accessibility in this case is the only logical option imo.

> This is not an officially endorsed way of using Mastodon. There's many ways in which you can potentially use this software but I cannot reasonably support all of them in the code, and this one is especially contrary to the core principles of decentralization.

See my arguments above.

> You are entitled to your own opinion on this, but just because I do not agree with you on this niche issue does not erase all the other ways in which Mastodon is different.

Except that this isn't just one user; a lot of disabled people use Mastodon and it's incredibly important to accomodate them. Labelling accessibility as a "niche issue" is insulting and ableist. The fact that Mastodon already has proper content warnings is a sign that it is very much more accessible than other networks, so, please do not push back accessibility as an important issue.

clarcharr:
Also adding that comments happened as I was typing my post and that the timeline will be disablable? I think.

As far as the GIF autoplay stuff happens, I think that the discussion on disabling GIFs by default in #3876 should happen before this feature is merged. It's so trivial to implement that it should be brought to attention before such a big change as this is put out there.

Gargron:
>As far as the GIF autoplay stuff happens, I think that the discussion on disabling GIFs by default in #3876 should happen before this feature is merged. It's so trivial to implement that it should be brought to attention before such a big change as this is put out there.

@clarcharr I disagree. I won't fault you or anyone else for not knowing our development workflow but master is not a stable branch, it represents the state of development. Things continually improve and large features are often split among small, easily digestable PRs. This PR is a large one, and should not be made any larger. What you *really* should mean is finishing the discussion before a new *release* is out, which I agree on.

clarcharr:
@Gargron oh, I didn't know that that was the case! We're on the same page then. Is there a particular milestone on GitHub that we could add that issue to to make sure it gets addressed?

nerdfiles:
Configurable controlling of playback seems important in that a fair interpretation of https://www.w3.org/WAI/WCAG20/quickref/#time-limits suggests that polling cycle duration on timelines would conflict with the length of the GIF. Naturally, the GIF would complete its iteration sometime after a timeline page update, for instance. Therefore, the user was not given a configurable way to control the amount of time they have to read the GIF content.

Gargron:
>that polling cycle duration on timelines would conflict with the length of the GIF

Polling does not reset the state of existing components, unless I am misunderstanding something about your comment.

beatrix-bitrot:
"The public timeline component on the frontpage will be made optional."

Thank~

ProgVal:
> Additionally, I'm very strongly of the opinion that privacy by obscurity is still worthwhile. There's a difference between "this information can be harvested by tech-savvy people and/or robots, and put into databases" and "my family can google-search my name and find out things I don't want to talk to them about right now." Putting toots on an open, indexable timeline makes them very public, to the point where literally anyone find them, whereas putting toots into an API endpoint makes them retrievable but not immediately noticeable.

@clarcharr Only public toots would be displayed in this timeline, and public toots are already indexable from the user profile.
Also, the timeline in this new landing page may be unindexable because it is dynamically loaded using Javascript.

duck57:
@clarcharr RE: local v. federated timelines on small instances, the flip side is on an instance that only has one or two active users: the local timeline would be the same two people posting over and over but the federated timeline also pulls in toots from the people they follow to break things up a bit.  I agree that it should be a configurable option, since which one presents a better representation of the instance is a decision that each individual admin needs to make.

clworld:
FYI: GNU Social has public FTL. i.e. https://freezepeach.xyz/main/all .
So, all toots federated to GNU Social are displayed on public page (indexing of GNU Social LTL/FTL is disallowed by robots.txt).

MCMic:
This feature is something I wanted :-)
That said, I have to agree that it will look weird for personal instances like mine. In this case I would like to be able to display the local timeline instead, which makes more sense IMO for a small instance.

Also, not completely related but I put it there, for a closed instance the public home page is kind of useless right now (it will be a bit better with this PR).
What I would love to be able to do is to put /about/more as public home page as it contains more interesting stuff and important links (terms, …).
It’s just missing a log-in link, and a timeline at the bottom like this PR does.

nerdfiles:
> Oh, and about GIF autoplay, people wanting GIF autoplay off everywhere should configure their browser, they can’t expect all websites to implement this on their own.

If the reason users wanted GIF autoplay off was not due to conflict with basic features of the product itself (streaming content that updates automatically), that would make sense — but in this case, the higher order question is the very intentional crafting of the UI and assumptions around UX. Users can and should configure their browsers, at their leisure and convenience. Forcing them to configure their browser because *other users* content has not be planned sufficiently within a community creates unmanageable social entropy. The motivation is that users should fee comfortable posting their content without the worry of introducing problems to other users — as content. So distilling the animation is just like normalizing typesettings. The same justifications hold. You wouldn't tell users "we're not going to set our fonts on our product because you can configure that in your browser." It is true you spend some non-negligible amount of time on typesetting, etc., when principally the user can configure it themselves. So why do that, and not offer *configurable* animated GIFs? If there were a *configurable-animated-GIF-lib* that you could install through `npm` or `homebrew` you'd use that if it were easy to implement, or made implementation easier. What's the principle to determine "when we have the justification to tell users to configure their browser"? — because such a principle is not entertained or authorized consistently. So it makes sense to product one's brand, in terms of GIF config, in just the same way they protect their brand by normalizing/typesetting fonts.

ProgVal:
@Gargron Thanks!

app/views/about/show.html.haml
```diff
@@ -9,79 +10,62 @@
...
+          Mastodon
 
-  %p!= t('about.about_mastodon')
+      %ul.nav
+        %li= link_to t('auth.login'), new_user_session_path, class: 'webapp-btn'
+        %li= link_to t('about.about_this'), about_more_path
+        %li= link_to t('about.other_instances'), 'https://joinmastodon.org/'
 
-  .screenshot-with-signup
-    .mascot= image_tag asset_pack_path('fluffy-elephant-friend.png')
```
ggtea: Do we no longer need a mascot? I still want mascot be there.

Cassolotl:
This new front page thing is looking AWESOME and I love it. I am excited to see it happen across the Mastodon network!

I think that gifs should be paused by default. Re: "people wanting GIF autoplay off everywhere should configure their browser, they can’t expect all websites to implement this on their own" - I am a reasonably techy frequent internet user on a very common OS using a very common browser (Mac, Chrome) and I have yet to find a way to auto-pause gifs in all cases all over the internet. Sometimes an extension will work on *some* gifs for a little while, but it never lasts (with the exception of x-kit). I have to close tabs as soon as I see animated gifs, pretty much. So yeah, telling us to each deal with the issue on our own machines, this is not an acceptable thing to say to someone! It's like saying, "oh you have a wheelchair? We shouldn't build ramps, you should carry a ramp around with you to lay over stairs when you need to enter a building."

This is an accessibility issue, and it's important, and thankfully it's pretty clear when an image is a paused gif. I do agree that folks should be able to see how awesome a site looks with all the gifs giffing happily away, but accessibility and not causing overload and not causing seizures and not throwing porn gifs in people's faces on porny instances etc. is really much more important! This whole front page shift is going to make such a huge difference to people joining up, I really do think it will help attract more users and such, but like... disabled people are potential users too. We're not an inconvenience, we're people who access sites just like anyone else and honestly seeing people talk like that about us is shocking and demeaning and humiliating.

Gargron:
GIFs default to non-animated now. But I added play-on-hover like for avatars. That's what Discord does and I am satisfied with that approach.

Cassolotl:
@gargron That sounds great. :)

Gargron:
Elephant friend is back:

image

nightpool (approved):
changes LGTM

MightyPork:
won't the URL collide with the trunk when it's too long? Otherwise it's :ok_hand:

config/locales/en.yml
```diff
@@ -1,35 +1,38 @@
...
-      chronology: Timelines are chronological
-      ethics: 'Ethical design: no ads, no tracking'
-      gifv: GIFV sets and short videos
-      privacy: Granular, per-post privacy settings
-      public: Public timelines
-    features_headline: What sets Mastodon apart
+      humane_approach_body: Learning from failures of other networks, Mastodon aims to make ethical design choices to combat the misuse of social media.
+      humane_approach_title: A more humane approach
+      not_a_product_body: Mastodon is not a commercial network. No advertising, no data mining, no walled gardens. There is no central authority.
+      not_a_product_title: You’re a person, not a product
```
mal0ki: "You are a person, not a product"
We haven't used informal writing in the rest of the text, so it would be bad to start now.

clworld: I think this sentence "You're a person, not a product" is referred to article like this: "You are Facebook's product, not customer" http://www.wired.co.uk/article/doug-rushkoff-hello-etsy

Gargron: Mal is talking about the difference between "you're" and "you are"

nightpool: calling "you're" informal feels faintly ridiculous to me—I checked three dictionaries, and none of them mark it as informal. Also, both Facebook and Twitter use it in official communications, and searching for it on the front page of the New York Times returns a bunch of results, and none for "you are".

Additionally, I think the context here is important: this is header copy, so it needs to be terse and readable. "You are a person, not a product" is very awkward to read aloud, takes longer to parse and understand, and generally would trip up a reader instead of draw them in—the opposite of what a good header should do.

nightpool: from a broader perspective, I think categorizing the tone of our copy as "informal" or "formal" is a false dichotomy. we definitely don't want to start sounding like over-processed corporate speak—we're trying to create a website users can trust, and that requires a careful balance of neutral tone and personable language.

mal0ki: Good points, all of it. And yeah, you got a point @nightpool, I'm just a bit broken from my thesis writing. It just felt like it was the only place we hyphenated, and thus it felt out if place. I should have been more clear in my comment. And I should do my full review before commenting (and add the comments together). There's definitely a balance to be had here.

PR #4128, opened 2017-07-09
fix(components/media_modal): Aspect ratio

sorin-davidoi:
A bit over my head with the CSS here. This should fix the issues mentioned by @unarist in #2889.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1281,6 +1281,10 @@
   }
 }
 
+.media-modal .react-swipeable-view-container {
+  align-items: center;
+}
+
 .react-swipeable-view-container > * {
   height: 100%;
 }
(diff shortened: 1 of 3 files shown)


app/javascript/styles/components.scss
```diff
@@ -1281,6 +1281,10 @@
   }
 }
 
+.media-modal .react-swipeable-view-container {
+  align-items: center;
```
unarist: Is not this line superseded by `animateHeight` option?

sorin-davidoi: Right, but it's superseded by `opacity: 0` (did not know how else to position the canvas behind the image).

unarist:
I'm worried about vertical moving on switching between different aspect ratios, but I don't have ideas yet...

<details>
<summary>GIF</summary>

test
</details>

Gargron (approved):

Gargron:
@unarist That gif looks pretty cool to me? Except that it seems a part of the old image is visible on the very left?

sorin-davidoi:
@unarist  We could just use the old package (`react-swipeable`), it's small enough.

unarist:
Maybe no one cares about it except me :)

PR #4129, opened 2017-07-10
Remote following success page

KScl:
Instead of immediately redirecting to web upon doing a remote follow, show an interstitial success page with links to web (the same redirect as before), and a link back to the remote user's profile, while also saying it's safe to close the page.

Personally, I keep my instance open in a pinned tab all the time but I prefer using the remote follow button to follow people over copying and pasting URLs around. Having remote follows automatically redirect to web was rather annoying since I would already have it open somewhere else.

Preview of how it looks:
rfconfirm

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -218,8 +218,8 @@ en:
   authorize_follow:
     error: Unfortunately, there was an error looking up the remote account
     follow: Follow
-    following_html: 'Success! You (<strong>%{self}</strong>) are now following:'
-    follow_request_html: 'You (<strong>%{self}</strong>) have sent a follow request to:'
+    following: 'Success! You are now following:'
+    follow_request: 'You have sent a follow request to:'
     post_follow:
       web: Go to web
       return: Return to the user's profile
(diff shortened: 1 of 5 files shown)


spec/controllers/authorize_follows_controller_spec.rb
```diff
@@ -94,7 +94,7 @@
       end
 
       it 'follows account when found' do
-        target_account = double(id: '123')
+        target_account = Account.new
```
Gargron: You can use `Fabricate(:account)`, please see other tests for examples.

config/locales/en.yml
```diff
@@ -218,6 +218,12 @@ en:
   authorize_follow:
     error: Unfortunately, there was an error looking up the remote account
     follow: Follow
+    following_html: 'Success! You (<strong>%{self}</strong>) are now following:'
```
Gargron: It is probably not necessary to specify self in these two strings because the user just clicked from a page that already did that.

STJrInuyasha: I was trying to keep the two pages looking similar in style, but now that I think about it, you're probably right.

Gargron (approved):

PR #4135, opened 2017-07-10
Drawer tab according to column

ykzts:
Add a link to the notifications to the drawer when there is no notification in columns.


beatrix-bitrot:
I like it!

Gargron:
I think it would be a bit inconsistent if the public/local icons are always on but only notifications is conditional.

ykzts:
@Gargron Certainly. I changed it.

### add all colums



### remove all columns

beatrix-bitrot:
the home column can also be un-pinned. will it show there if it is?

ykzts:
@beatrix-bitrot I add :house:

Gargron (approved):

beatrix-bitrot:
thanks @ykzts !!

Issue #2760, opened 2017-05-03
Compose tab does not work on Jolla Sailfish browser


bochecha:
Since last week, I can't open the compose tab on my Jolla.

(I think that corresponds to when the mstdn.fr admins upgraded to 1.3.0, but I'm not entirely sure. In any case, mstdn.fr is running Mastodon 1.3.2 now and the problem still exists)

The symptoms is that all the tabs work just fine, but when clicking on the compose tab, I get the contents of the last tab I displayed, as if the button did nothing.

The Jolla browser is based on Firefox 38, so I downloaded Firefox 38.8.0 ESR on my desktop and managed to reproduce the exact same issue by turning on the Responsive Design View and setting the viewport to 540x960, the size of my Jolla.

There, all the tabs worked just fine... except the compose tab.

When I click on the compose tab, I can see this in the Firefox console:

```
TypeError: value.keywords.includes is not a function application-de58e14cf4efdb9107448763bdfb1ffa4bbe12799d96dfe0af6d192f63434d94.js:82:3149
```

It would be really great if Mastodon worked on the Jolla browser! (of course, it would be great if Jolla updated their browser to a less ancient version of Firefox...)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

joyeusenoelle:
Another user is reporting this as well: https://octodon.social/users/cedesguin/updates/96782

It looks like this may be an issue with the emojione-picker. I'll open an issue with that repo.

objectifnul:
Also, I'd like to mention that webkit-based browsers on Jolla/SailfishOS (Webpirate and Webcat) don't work with any Mastodon instance.

joyeusenoelle:
If Webkit browsers on Sailfish don't work with Mastodon at all, that's probably a separate problem; please open a new issue with any information about it you can offer.

objectifnul:
Update 1.4.x makes the mobile web interface now completely non functional with SailfishOS :-(

Gargron:
In our quest to reduce bundle size, we have reduced the number of polyfills in favour of expecting modern browser features to already exist. Perhaps the Jolla browser fell victim to this.

cc @nolanlawson @sorin-davidoi

sorin-davidoi:
Latest master works find in development.
[...]

PR #4145, opened 2017-07-10
Update es5-ext to avoid CSP violation

unarist:
Since es5-ext used `new Function("...")`, it caused CSP violation unless "unsafe-eval" included. So this patch updates it to the version which fixes it.

Note that this package is used in polyfills (e.g. Symbol), so loaded only if needed. I've encountered this issue on iOS9.

cf. medikoo/es5-ext@d3864493


Gargron (approved):

sorin-davidoi:
This might also improve things for #2760.

PR #4146, opened 2017-07-10
HTTP signatures

Gargron:
Spec: <https://tools.ietf.org/html/draft-cavage-http-signatures-06>

Access control on Atom resource URLs can be implemented with this spec.

TODO:

- [x] Class to sign requests (`Request`)
- [x] Parsing and verifying signatures in controllers (`SignatureVerification`)
- [x] Replace direct usage of HTTP with new Request class
- [x] Test positive cases
- [x] Test cases when request has been tampered with (signature does not match)

Potential uses in the future:

- Fix private toot delivery to instances where LOCAL_DOMAIN != WEB_DOMAIN, by authenticating PuSH subscriptions and storing the signed account's "domain" along the callback_url (https://github.com/tootsuite/mastodon/pull/2739#issuecomment-313990715)
- General use in ActivityPub when fetching entities


ClearlyClaire (changes requested):
Worried about how signed `date` HTTP headers are handled, and about handling of multiple headers with the same fields (although it seems like Mastodon-originating requests won't have multiple occurrences of the same header field). Seems ok otherwise.

app/controllers/concerns/signature_verification.rb
```diff
@@ -0,0 +1,77 @@
...
+  private
+
+  def build_signed_string(signed_headers)
+    signed_headers = 'date' if signed_headers.blank?
+
+    signed_headers.split(' ').map do |signed_header|
+      if signed_header == Request::REQUEST_TARGET
+        "#{Request::REQUEST_TARGET}: #{request.method.downcase} #{request.path}"
+      elsif signed_header == 'date'
+        "date: #{Time.now.utc.httpdate}"
```
Gargron: How else are you supposed to prevent replay attacks? You need to check that a request that says it comes at a specific time did come at the specific time. This is an honest question - the spec itself is extremely blurry about this, though it does mandate the "date" header for some kind of verification.

app/lib/request.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    "keyId=\"#{key_id}\",algorithm=\"#{algorithm}\",headers=\"#{signed_headers}\",signature=\"#{signature}\""
+  end
+
+  def signed_string
+    @headers.map { |key, value| "#{key.downcase}: #{value}" }.join("\n")
+  end
+
+  def signed_headers
+    @headers.keys.join(' ').downcase
```
Gargron: I mean, by default we only have user-agent, date and host headers. Places that make use of `add_headers` only add headers that are required by e.g. spec. So I don't think there is an issue here.

app/workers/pubsubhubbub/confirmation_worker.rb
```diff
@@ -60,9 +60,7 @@ def confirm_callback
   end
 
   def callback_get_with_params
-    HTTP.headers(user_agent: 'Mastodon/PubSubHubbub')
-        .timeout(:per_operation, write: 20, connect: 20, read: 50)
-        .get(subscription.callback_url, params: callback_params)
+    Request.new(:get, subscription.callback_url, params: callback_params).perform
```
Gargron: I am unifying them in one place, which I think is the primary benefit of this. Also, letting a request stall for 90 seconds wasn't good behaviour, I think. We'd rather fail and retry later.

app/controllers/concerns/signature_verification.rb
ThibG: The `Date` header could indeed be used to avoid replay attacks, but it is completely out of this draft's scope (the `Date` header is actually a SHOULD, not a MUST). It is very unlikely that the request's `Date` header matches the time it is processed on the server down to the second.

If you are worried about replay attacks, you could verify that the difference between the time of the `Date` header and the time it is processed does not exceed some reasonable bound. I'm not sure what a reasonable bound would be, as many servers are slightly out of date and do not run ntp.

app/lib/request.rb
ThibG: I see no use for stuff like user-agent to be signed, but I have no real opposition to it either, so I'm fine with this.

app/workers/pubsubhubbub/confirmation_worker.rb
ThibG: Right.

app/controllers/concerns/signature_verification.rb
Gargron: So signing the date header makes sense to make every signature different, but verifying its value is not important, is that right? We could give it like a minute drift window, I think that's what TOTP does.

ThibG: Well, I guess the reason they recommend signing `Date` is to do that kind of verification, but it's a separate step.

Gargron:
@ThibG Addressed date issue and documented draft. Don't think duplicate headers should occur in our use case.

ThibG:
@Gargron I'm still worried a 30s window is much too tight: I routinely see time discrepancies of the order of minutes between Mastodon instances. Furthermore, I think at least in the case of PuSH subscriptions, replay attacks are a non-issue. Can't really say for ActivityPub, though.

Regarding multiple headers for the same field, it will not be an issue between Mastodon instances, but this means you don't actually verify signatures according to the spec, and any software that would sign multiple headers of the same name for whatever reason won't correctly interoperate with Mastodon.

Gargron:
@ThibG I am worried that if someone gets their hands on such requests in real time, they could replay them and the server would genuinely respond to them, within that time window. If we want to use this for access control, that's risky. That being said, SSL should protect the requests from being caught by a 3rd party, so I don't know if the risk is real.

ThibG:
@Gargron well, to be honest, there are things I don't get with the spec itself: why doesn't it cover signing the request data, but only headers? That doesn't really protect you from MITM attacks that would just change the request data while keeping the request headers intact.

Regarding replay attacks, as you said, TLS should protect the requests from being caught by a 3rd party. And if not, you have worse problems (namely, the aforementioned one, or tampering with the key exchange in the first place, etc.)

Gargron:
>why doesn't it cover signing the request data, but only headers?

@ThibG It delegates that to the Digest header from a different spec

ThibG:
@Gargron thanks, I missed that, that makes sense! So, assuming you follow RFC 3230 and sign the Digest header, replay attacks are just replay attacks. How much of an issue they are depends on the kind of messages you're dealing with, and I would argue they are not much of an issue for things like PuSH subscriptions. I think I would make the validity window an optional parameter, make it default to something like a few minutes, and specifically log requests that fail validating because of this.

Gargron:
>I think I would make the validity window an optional parameter, make it default to something like a few minutes, and specifically log requests that fail validating because of this.

@ThibG I'd like to leave it as-is. Nothing right now actually makes use of the value yet, and even if verification fails, all we get is the status quo, it's not like the request gets denied. We'll have time to figure this out in future PRs.

ThibG:
@Gargron I would still argue it should be an optional parameter, and such failures should be logged. As for my other gripe with this PR, the handling of multiple headers sharing the same field-name… it seems to me it would be quite complicated to get it right with RoR, which joins such values with “, ” (which is sound with regards to the HTTP specs, but is a problem for the HTTP Signature one). That should probably be explained somewhere in this module.

app/services/subscribe_service.rb
```diff
@@ -2,34 +2,54 @@
...
+  end
+
+  def subscription_params
+    {
+      'hub.topic': @account.remote_url,
+      'hub.mode': 'subscribe',
+      'hub.callback': api_subscription_url(@account.id),
+      'hub.verify': 'async',
+      'hub.secret': @account.secret,
+      'hub.lease_seconds': 30.days.seconds,
```
Gargron: This is actually a change, I think, because through a bug it was 7 days previously. I think we might want to keep it to 7 days instead of 30, because while 30 is more efficient, 7 means less "black hole" time if somebody messes up and loses their DB with all subscription data.

Gargron:
@ThibG 

>it seems to me it would be quite complicated to get it right with RoR, which joins such values with “, ”

From the spec:

>If there are multiple instances of the same header field, all header field values associated with the header field MUST be concatenated, separated by a ASCII comma and an ASCII space `, `, and used in the order in which they will appear in the transmitted HTTP message

So it works out of the box?

ThibG:
@Gargron ah yes, sorry, indeed, it should work out of the box then! So the only change I'm suggesting now is the time window as parameter, and some logging when it doesn't match the window.

Gargron:
@ThibG Fixed!

ClearlyClaire (approved):

PR #4147, opened 2017-07-10
Rewrite Immutablejs import statements using destructuring

sorin-davidoi:
This is the first step towards tree-shaking ImmutableJS - we only currently use about half of it. The module itself does not currently support it, but there are a few efforts to change it. This has not effect to the functionality or bundle size, it just prepares the code for the future (and makes it easier for me to continue trying to tree shake it).


sorin-davidoi:
I've used the naming idiom Immutable-${structureName} to avoid conflicts with the built-in data structures such as Map and Set.

Gargron (approved):

PR #4152, opened 2017-07-11
Require stylesheets in common.js

akihikodaki:
Require stylesheets in `common.js` because stylesheets are shared by the entry points.

### DIFF
--- a/app/javascript/mastodon/main.js
+++ b/app/javascript/mastodon/main.js
@@ -1,9 +1,5 @@
 const perf = require('./performance');
 
-// import default stylesheet with variables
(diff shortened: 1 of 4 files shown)


Gargron (approved):

nightpool:
Question: does this increase the size of the assets we serve for static pages? Or am I missing something?

nightpool:
Looks like this breaks the embed endpoint

akihikodaki:
The only change in the output is that the file which was named `application.css` is named `common.css`.
73a83c8a9c164aebad4f66f26c4666a6979d5349 fixes `embed`.

PR #4156, opened 2017-07-11
Fix #2848 - Rake task to redownload avatars/headers

### LINKED ISSUE #2848: Force re-fetch remote profile picture / header
Hello, I'm in a problem, I was forced to silence & un-cache media for a moment on my instance, removing all avatar/header/pics.
but now I un-silenced everybody since the problem is gone.
The problem is that profile picture / header isn't here, and isn't fetched until they change it since nobody unfollowed them while I was silencing thoses.
If someone can help me force fetch thoses, thanks (or even, re-fetch everyone we federate with... )

* * * *

Gargron:
Can be filtered by a specific domain
Resolves #2292

### DIFF
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -22,8 +22,8 @@ module Admin
     end
 
     def redownload
-      @account.avatar = @account.avatar_remote_url
-      @account.header = @account.header_remote_url
+      @account.reset_avatar!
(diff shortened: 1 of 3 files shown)


ykzts (approved):

malicioustoker:
What command would we run via Docker to accomplish this task?

Gargron:
`docker-compose run --rm web rails mastodon:media:redownload_avatars`

PR #4157, opened 2017-07-11
Revert "Wrap methods of ProcessFeedService::ProcessEntry in classes"

nightpool:
Reverts tootsuite/mastodon#4151

Caused two regressions:

 - in conversation_to_persistent, local statuses return just the conversation, where other statuses return an array of [Conversation, was_created]. this causes a type error for processing local statuses.

 - sometimes, `FetchRemoteStatusService` returns nil. This code does not properly handle that case, leading to a NoMethodError when we try to check whether it's a reblog or not.

see attached screenshot from glitch.social:

https://images.discordapp.net/attachments/231908446830723072/334351534559330305/unknown.png

To be clear: I think this is a good refactor, and I'm happy to see the code getting more OO. But unfortunately it's currently breaking lots of things, so it needs more thorough review before getting merged.

### DIFF
--- a/app/services/process_feed_service.rb
+++ b/app/services/process_feed_service.rb
@@ -22,285 +22,268 @@ class ProcessFeedService < BaseService
   class ProcessEntry
     def call(xml, account)
       @account = account
-      @fetched = Activity.new(xml)
+      @xml     = xml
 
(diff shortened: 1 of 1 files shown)


nightpool:
uh oh, Github reverted everything since then. I will fix.

Gargron (approved):

PR #4162, opened 2017-07-11
Fix boolean columns sometimes having a null value

Gargron:


ykzts:
It is likely that you need to execute `Status.where(sensitive: nil).update_all(sensitive: false)` before `db:migrate`.

Gargron:
@ykzts `change_column_null :domain_blocks, :reject_media, false, true`

<http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null>

>The method accepts an optional fourth argument to replace existing NULLs with some other value. Use that one when enabling the constraint if needed, since otherwise those rows would not be valid.

Actually, I am not sure if the fourth argument is the *value* that nulls will be replaced with, or if it decides whether the table-default for this column will be used.

ykzts:
I see.

ykzts (dismissed):

Gargron:
Looks like it's the former, not the latter: <https://apidock.com/rails/v4.2.7/ActiveRecord/ConnectionAdapters/PostgreSQL/SchemaStatements/change_column_null>

So I need to update all "true" to "false"

ykzts (approved):

PR #4164, opened 2017-07-12
Introduce Ostatus name space

akihikodaki:
This is a change same with 425acecfdb15093a265b191120fb2d4e4c4135c4, except that it has the following changes:

* Revert irrelevant change in `find_or_create_conversation`
* Fix error handling for `RemoteActivity`


nightpool:
Is it possible you could add regression tests for the two errors we uncovered after deploying this? that would be amazing

akihikodaki:
I have added a test for reblog failure. There is no additional spec for `find_or_create_conversation` bug since the problematic logic is just removed.

akihikodaki:
We have tested this patch for 24+ hours on o.kagucho.net and not experienced errors yet except that ones caused by remote servers which are down.

app/services/process_feed_service.rb
```diff
@@ -22,268 +22,285 @@ def process_entries(xml, account)
...
-      when :post, :share
-        return create_status
-      when :delete
-        return delete_status
-      end
+      @fetched = Activity.new(xml)
+
+      return unless [:activity, :note, :comment].include?(@fetched.type)
+
+      klass = case @fetched.verb
```
Gargron: I feel like this could be a private method that is just the switch statement. The indented assignment always looks out of place.

app/services/process_feed_service.rb
```diff
@@ -22,268 +22,285 @@ def process_entries(xml, account)
...
       nil
     end
 
-    private
-
-    def create_status
-      if redis.exists("delete_upon_arrival:#{@account.id}:#{id}")
-        Rails.logger.debug "Delete for status #{id} was queued, ignoring"
-        return
+    class Activity
```
Gargron: I wouldn't be against extracting all these inner classes into their own files, for example under app/services/ostatus/activities - your patch is a fantastic improvement in code organization, but the outer class has also become much bigger because of it, and harder to read.

Gargron (changes requested):
👍

akihikodaki:
cfc61829740b87f886809e884c34391af9e02959 applied changes requested in https://github.com/tootsuite/mastodon/pull/4164#pullrequestreview-50224566.

Gargron (approved):

app/services/process_feed_service.rb
```diff
@@ -16,274 +16,14 @@ def update_author(body, account)
...
-      status, just_created = nil
-
-      Rails.logger.debug "Creating remote status #{id}"
-
-      if verb == :share
-        original_status = shared_status_from_xml(@xml.at_xpath('.//activity:object', activity: TagManager::AS_XMLNS))
-        return nil if original_status.nil?
-      end
-
-      ApplicationRecord.transaction do
```
Gargron: @akihikodaki It seems like the transaction is gone completely. I think I am experiencing glitches due to it (incomplete statuses in cache).

PR #4176, opened 2017-07-12
Fix #1010 - When spoiler text is set, enforce sensitivity too

### LINKED ISSUE #1010: CW should include the image, too

Gargron:
There is a thing that can be done in web UI about connecting the visible/hidden state between the "show more" and the media spoiler, but this resolves the crux of the issue.

Resolves #3979

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -124,6 +124,7 @@ class Status < ApplicationRecord
   before_validation :set_reblog
   before_validation :set_visibility
   before_validation :set_conversation
+  before_validation :set_sensitivity
 
   class << self
     def not_in_filtered_languages(account)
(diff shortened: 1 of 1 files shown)


beatrix-bitrot (approved):
very well then! code's simple enough

i think the 'thing that can be done' should be done

but it doesn't need to be done in this changeset imo

PR #4177, opened 2017-07-12
Fix threaded replies to non-federated statuses not being filtered from the timeline

nightpool:
bug discovered by @vahnj. If someone you follow replies to a status your instance doesn't know about, it gets filtered. but if they continue replying in that thread, further replies don't get filtered. This adds logic to cover that edge case.

This also refactors the feed manager code because it's super out of date, and doens't use the account interaction maps. (lots of Follow.where(target_id: blah, account_id: blah))

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -106,9 +106,9 @@ GEM
       rack (>= 1.0.0)
       rack-test (>= 0.5.4)
       xpath (~> 2.0)
-    charlock_holmes (0.7.3)
     case_transform (0.2)
       activesupport
+    charlock_holmes (0.7.3)
     chunky_png (1.3.8)
     cld3 (3.1.3)
       ffi (>= 1.1.0, < 1.10.0)
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -11,11 +11,11 @@ class FeedManager
     "feed:#{type}:#{id}"
   end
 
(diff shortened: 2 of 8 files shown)


nightpool:
codeclimate won't let me dismiss false positives, but in order:

 - empty line at beginning of method is for docs (not sure if they're useful now that the code is more readable, but they're there)
 - = operator spacing is to align with the +=/||= on the next line

akihikodaki:
rubocop says line 120, around which there is no `+=`, has an extra space.

nightpool:
@akihikodaki fixes

Gargron:
See 6fd865c0004efbf11ee87c06fea9f48af567fabe

You're reverting deliberate optimizations - this code is a performance hot spot and we want it to handle as little information as possible, thus passing it an integer and not a full account, and using batched checks for mutes, blocks etc.

If I'm reading correctly the actual fix is the line `return true if status.in_reply_to_account.nil? || status.in_reply_to_id.nil?`, which in old code would be `return true if status.reply? && (status.in_reply_to_account.nil? || status.in_reply_to_id.nil?)`

nightpool:
@Gargron Since that code was written, the performance of muting?, blocking?, and following? have all been changed, and now the two alternatives create identical SQL queries, but one is much more readable and maintainable.

I'm not sure how passing an integer vs. a full account would change the performance in this case at all, seeing as everywhere we call it we are explicitly taking an id from a full activerecord object.

PR #4178, opened 2017-07-13
[WIP] Move emojify logic to the backend

sorin-davidoi:
Tasks:
 - [x] Replace Unicode characters
 - [x] Replace shortcodes
 - [ ] Find all endpoints that have to returned emojified text
 - [ ] Port Nolan's tests to Ruby
 - [ ] Enable native code optimizations


Gargron:
>Replace Unicode characters

Wait, wait, don't do this. We want to return unicode, because the html is consumed both by browsers but also by various apps. Don't make them have to parse img tags.

sorin-davidoi:
Should it be another field then?

Gargron:
I just thinking the emoji pack, be it emojione, twemoji, native, or whatever, should be the client's responsibility - in our case web UI. The server should return unicode.

Gargron:
All that I *want* from this PR is convert \:shortcodes\: into unicode - it doesn't have to be emojione-specific, if there is some other more general emoji index gem...

sorin-davidoi:
> The server should return unicode.
That makes sense, I just thought that since we are returning HTML for non-local statuses we could also add `img` tags. Will close this and hopefully open another one down the road.

PR #4182, opened 2017-07-13
Add object-fit polyfill for Edge

unarist:
[Edge 16 supports `object-fit`](http://caniuse.com/#search=object-fit), but it's still in preview version.

Also there is [another polyfill for 'srcset' attribute](https://github.com/scottjehl/picturefill), but I omited that since it's not huge issue like this I think.

cc @nolanlawson


Gargron (approved):

Gargron:
Please fix conflict!

unarist:
done.

nolanlawson:
Edge 16 will ship in the Fall Creators Update (in a few months), and as-is I didn't find the lack of object-fit to be a big deal when using Edge (the images were still visible, just a bit smaller). Just FWIW. :smiley:

PR #4187, opened 2017-07-13
fix(components/media_modal): Style issues

sorin-davidoi:
Fixes most of the issue described by @unarist [here](https://github.com/tootsuite/mastodon/pull/4105#issuecomment-314850475).

My CSS skills fail me at the following issue:
> object-fit won't shrink element itself, so they can't close modal on clicking those margin. It confuses users especially on single image which doesn't have prev/next buttons.

If anyone has any input on fixing this one let me know and I'll include it here.

### DIFF
--- a/app/javascript/mastodon/components/extended_video_player.js
+++ b/app/javascript/mastodon/components/extended_video_player.js
@@ -5,6 +5,8 @@ export default class ExtendedVideoPlayer extends React.PureComponent {
 
   static propTypes = {
     src: PropTypes.string.isRequired,
+    width: PropTypes.number,
+    height: PropTypes.number,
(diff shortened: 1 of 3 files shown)


Gargron (approved):

lynlynlynx:
To deal with non shrinking modal on single image, I use [this way.](https://github.com/lynlynlynx/mastodon/compare/sanin-master...lynlynlynx:fix-imageloader-aspect-chrome)

But it is difficult to shrink modal on multiple images for me.  
I have no idea to...  
- Get width of reduced image in each one
- Change swipeable-view-container width every time image is switched

@sorin-davidoi

PR #4189, opened 2017-07-13
Fix #4149, fix #1199 - Store emojis as unicode

### LINKED ISSUE #1199: Emoji Auto-complete + UTF8 representation
captura de tela de 2017-04-07 18-21-29

This screenshot was taken from quitter.is, the other side of the federation. Maybe Mastodon handles emoji/unicode in a nonstandard way?

### LINKED ISSUE #4149: Convert emoji shortcodes to UTF-8 in client for char-count + rendering purposes
Separated out from #1199 and #717 as a separate (but still important) issue. The codes should still be replaced on the server too, but there should be client support as well.

Gargron:
- Use unicode when selecting emoji through picker
- Convert shortcodes to unicode when storing text input server-side
- Do not convert shortcodes in JS anymore

Reference: <https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/10647>


app/javascript/mastodon/emoji.js
```diff
@@ -11,31 +11,13 @@ function emojify(str) {
   // The goal is to be the same as an emojione.regShortNames/regUnicode replacement, but faster.
```
sorin-davidoi: The description is outdated.

sorin-davidoi:
I'm wondering, shouldn't we make the emoji picker insert unicode instead of shortcodes? I'm looking into tree shaking EmojiOne and if we take out the shortcode substitution in the compose form (which basically pull in the entire list of shortcodes), we can shrink it by **~90%** (from `288 KB` to `32 KB`).

Gargron:
>I'm wondering, shouldn't we make the emoji picker insert unicode instead of shortcodes?

@sorin-davidoi I did that in this PR. So the emoji picker inserts unicode, but if you type out the shortcode, it also gets converted to unicode server-side.

sorin-davidoi:
Ah, missed that! Awesome, will follow this with the tree shaking!

beatrix-bitrot (approved):

unarist:
After this change, remote contains and also existing local contents which contains `:shortname` form won't be emoji?

PR #4190, opened 2017-07-14
Fix subsequent replies to unresolved status not being filtered from home

Gargron:
Resolves #4177 - smaller changeset


nightpool (approved):

nightpool (changes requested):
actually, can you copy over the spec I wrote too? that code is very well spec'd and i'd hate to regress there

nightpool (approved):
LGTM

PR #4191, opened 2017-07-14
Fix #3904 - Adjustable time period for mastodon:media:remove_remote via NUM_DAYS

### LINKED ISSUE #3904: Make "remove remote media attachments" time period adjustable
There is a rake command to drop remote media attachments after a week: `rails run media:remove_remote` ([code](https://github.com/tootsuite/mastodon/blob/75c6513c678148e9a6adfcda57cfd5c8a16c804a/lib/tasks/mastodon.rake#L57-L64)).

This is a great feature to keep the asset directory from growing indefinitely, but a week seems awfully short to be hardcoded.

Let's make this adjustable, either in `.env.production` or otherwise?

* * * *

Gargron:


ykzts (approved):

Angristan:
Is this valid ?

`RAILS_ENV=production NUM_DAYS=30 bundle exec rails mastodon:media:remove_remote`

Gargron:
Yeah

fwenzel:
Thank you!

PR #4202, opened 2017-07-14
🎄🔨 Force tree shake emojione

sorin-davidoi:
Based on #4189, will rebase as soon as it is merged.

We use [babel-plugin-preval](https://github.com/kentcdodds/babel-plugin-preval) to cherry-pick what we need from `emojione` and avoid shipping the entire library to the browsers. The entire library is still imported by `emoji-picker`, but it gets included in that bundle, which is only downloaded when the emoji drop-down is expanded.

Effect on the `common` bundle:

|             | Before  |   After   | Difference |
| --------- |:----------:|:----------:|:-------------:|
| Parsed | 749 KB | 493 KB | :small_red_triangle_down: -34% |
| Gziped | 182 KB | 142 KB | :small_red_triangle_down: -22% |

[Bundle view](http://bl.ocks.org/sorin-davidoi/raw/6db0b31fe7384713c759b7b53e3062e7/) - notice how small `emojione_light` is compared to `emojione`.

cc: @nolanlawson

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -22,7 +22,8 @@
       {
         "messagesDir": "./build/messages"
       }
-    ]
+    ],
+    "preval"
   ],
   "env": {
     "development": {
--- a/app/javascript/mastodon/emoji.js
+++ b/app/javascript/mastodon/emoji.js
@@ -1,8 +1,7 @@
-import emojione from 'emojione';
+import { unicodeToFilename } from './emojione_light';
 import Trie from 'substring-trie';
 
-const mappedUnicode = emojione.mapUnicodeToShort();
-const trie = new Trie(Object.keys(emojione.jsEscapeMap));
+const trie = new Trie(Object.keys(unicodeToFilename));
 
 function emojify(str) {
   // This walks through the string from start to end, ignoring any tags (<p>, <br>, etc.)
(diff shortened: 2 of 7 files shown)


Gargron:
Now that the original PR is merged, can you try rebasing on top of master?

sorin-davidoi:
Rebased. Also added a preload script for `emojione_picker`, since it is bigger now.

Gargron (approved):

sorin-davidoi:
Thank you @kentcdodds for this library, saves us quite a few bytes!

unarist:
Oh, this removed `title` attribute? I think `:emoji:` syntax is still useful for manual input, and `title` attribute helps knowing its shortname.

sorin-davidoi:
Yes, but displaying that (on hover, so only on desktop) means pulling in the entire array, which is **big**. I would argue that it is not worth it - you can always see the shortcode in the emoji picker.

kentcdodds:
So glad that preval helped! Good work here!

milesj:
Just chiming in here. Have you look at something like [emoji-database](https://github.com/milesj/emoji-database) over emojione?

sorin-davidoi:
We haven't. The thing is that we already use `emojione-picker`, which brings in `emojione`, so we would rather use it since it will be fetched and bundled anyway. There has been some talk about switching to another emoji set, if that happens we will definitely look into it!

milesj:
No worries, makes sense! If you guys do use it, and have any feedback, feel free to send it my way!

nolanlawson:
This is awesome. Another cool thing about this PR is that it will take less time to evaluate the JS, because if you look at emojione's source you'll see it does a lot of work to build up regexes and object maps.

It's a bit of a shame we no longer have the title, but potentially a low cost way to bring it back would be to pre-eval a trie structure. Or as @sorin-davidoi says, we can just accept that it's gone because perhaps it's not really worth it.

Gargron:
Just FYI people have since expressed interest in the title attribute, I think there is a high demand for it. It doesn't only teach people how to invoke the emojis without access to a unicode keyboard/utility, but it helps recognize emoji that are hard to understand visually.

kaniini:
the only thing i have interest in is reverting this merge.  it makes the style of the emojis inconsistent, and worse, breaks terribly on linux if you do not have a font that correctly supports emojis.

sorin-davidoi:
This PR does not affect how emojis are rendered. The only visual change it brings is that it removes the title attribute, which we will try to bring back due to the feedback received.

Issue #3166, opened 2017-05-20
Boost and favourite count are out of sync


sorin-davidoi:
When removing the favourite and boost status of a toot, the numbers in the `status_action_bar` get out of sync. The `unfavourite` and `unreblog` endpoints return the old values, probably since the work is async.
![](http://g.recordit.co/G408nT58Dw.gif)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

alexgleason:
This issue seems to be fixed on the frontend in v3.1.2, although POST /api/v1/statuses/:id/unfollow still returns the old favourites_count

xmflsct:
Agree, using API endpoint to perform action is still unsync. Have tried `statuses/:id/favourite` and `statuses/:id/unfavourite`. Any update on this?

vmstan:
This issue was filed against a now unsupported version of Mastodon. In an effort to clean up stale bugs, if this can be reliably reproduced on an updated version please open a new issue with updated logs/troubleshooting and we will take a look. Thank you.

PR #4090, opened 2017-07-06
Refactor JSON templates to be generated with ActiveModelSerializers instead of Rabl

Gargron:
Why:

- Code looks cleaner, feels like it will be cleaner than Rabl when implementing ActivityPub serializers
- I believe AMS might be more performant than Rabl

I have not yet managed to completely remove Rabl as a dependency. "Initial state" and OEmbed responses, as well as the old ActivityPub stuff that will be replaced still uses Rabl.

PR #4204, opened 2017-07-14
Fix response of unreblog/unfavourite APIs

unarist:
Regression from #4090 

Both APIs process asynchronously, so reblogged/favourited fields in the response should be set to `false` manually.


sorin-davidoi:
Does this fix #3166?

Gargron (approved):

Gargron:
@sorin-davidoi No, we don't touch the counters there. It could be solved client-side, I think.

unarist:
Actually, it is similar issue (i.e. those counts in the response is old value), but it exists from before dropping rabl.

PR #4205, opened 2017-07-14
Fix #2672 - Connect signed PuSH subscription requests to instance domain

### LINKED ISSUE #2672: v1.3+ does not send toots to instances using a separate WEB_DOMAIN

Gargron:
Resolves #2739

### DIFF
--- a/app/workers/pubsubhubbub/distribution_worker.rb
+++ b/app/workers/pubsubhubbub/distribution_worker.rb
@@ -35,16 +35,16 @@ class Pubsubhubbub::DistributionWorker
     @payload = AtomSerializer.render(AtomSerializer.new.feed(@account, stream_entries))
     @domains = @account.followers.domains
 
-    Pubsubhubbub::DeliveryWorker.push_bulk(@subscriptions.reject { |s| !allowed_to_receive?(s.callback_url) }) do |subscription|
+    Pubsubhubbub::DeliveryWorker.push_bulk(@subscriptions.reject { |s| !allowed_to_receive?(s.callback_url, s.domain) }) do |subscription|
       [subscription.id, @payload]
     end
   end
 
   def active_subscriptions
-    Subscription.where(account: @account).active.select('id, callback_url')
+    Subscription.where(account: @account).active.select('id, callback_url, domain')
   end
 
-  def allowed_to_receive?(callback_url)
-    @domains.include?(Addressable::URI.parse(callback_url).host)
+  def allowed_to_receive?(callback_url, domain)
+    (!domain.nil? && @domains.include?(domain)) || @domains.include?(Addressable::URI.parse(callback_url).host)
   end
 end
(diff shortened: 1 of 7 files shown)


app/workers/pubsubhubbub/distribution_worker.rb
```diff
@@ -35,16 +35,16 @@ def distribute_hidden!(stream_entries)
...
   def active_subscriptions
-    Subscription.where(account: @account).active.select('id, callback_url')
+    Subscription.where(account: @account).active.select('id, callback_url, domain')
   end
 
-  def allowed_to_receive?(callback_url)
-    @domains.include?(Addressable::URI.parse(callback_url).host)
+  def allowed_to_receive?(callback_url, domain)
+    (!domain.nil? && @domains.include?(domain)) || @domains.include?(Addressable::URI.parse(callback_url).host)
   end
```
ThibG: This is not new to this PR, but I think you actually want `.normalized_host` instead of `.host`.

Gargron:
Bonus round: at the cost of backward compatibility, we could require that the domain field be filled, or not send private toots to it - that way we'd get rid of the non-Mastodon instances getting private toots issue. But that's up for debate since it'd exclude outdated Mastodon instances too

ClearlyClaire (approved):
Apart from one corner case that is not new to this PR, looks good to me.

app/workers/pubsubhubbub/distribution_worker.rb
Gargron: If I remember correctly, callback_url is normalized before save.

Gargron:
Ah, I didn't adjust the tests.

Also, the method signature having so many params does look bad, but I dunno how to improve it.

app/workers/pubsubhubbub/distribution_worker.rb
ThibG: Ah, seems to be the case, indeed.

ThibG:
@Gargron I'm not completely against enforcing the subscription requests to be signed for private toot delivery, but I think this should be done in a later release, exactly for the reasons you mentioned.

PR #4211, opened 2017-07-15
Fix column swiping

unarist:
This fixes broken behavior and enable animation only on swiping.

* `onTransitionEnd` called before `onChangeIndex` if `animateTransitions` is false. In a result, `location.pathname` won't be updated on swiping.
* Transition depends `animateTransiton` value which is set before transition, i.e. we need to maintain `true` as much as possible to animate swiping. 

This also removes animation from transition caused by location changing.

<details>
<summary>Before</summary>

before
</details>

<details>
<summary>After</summary>

after
</details>

cc @sorin-davidoi

### DIFF
--- a/app/javascript/mastodon/features/ui/components/columns_area.js
+++ b/app/javascript/mastodon/features/ui/components/columns_area.js
@@ -33,8 +33,22 @@ export default class ColumnsArea extends ImmutablePureComponent {
     children: PropTypes.node,
   };
 
+  state = {
+    shouldAnimate: false,
+  }
+
+  componentWillReceiveProps() {
+    this.setState({ shouldAnimate: false });
+  }
+
+  componentDidMount() {
+    this.lastIndex = getIndex(this.context.router.history.location.pathname);
+    this.setState({ shouldAnimate: true });
+  }
+
   componentDidUpdate() {
     this.lastIndex = getIndex(this.context.router.history.location.pathname);
+    this.setState({ shouldAnimate: true });
   }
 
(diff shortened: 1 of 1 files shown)


Gargron (approved):

app/javascript/mastodon/features/ui/components/columns_area.js
```diff
@@ -74,9 +88,10 @@ export default class ColumnsArea extends ImmutablePureComponent {
 
   render () {
     const { columns, children, singleColumn } = this.props;
+    const { shouldAnimate } = this.state;
 
     const columnIndex = getIndex(this.context.router.history.location.pathname);
-    const shouldAnimate = Math.abs(this.lastIndex - columnIndex) === 1;
```
sorin-davidoi: Shouldn't we animate if we move to the previous / next tab? Might help with discoverability of the feature and - subjective - I think it looks nice.

sorin-davidoi: I also feel that it improves perceived performance.

unarist: When we use animation for location change based transition, `columnIndex` is already new one before transition, whereas it's old one on swiping based transition. So destination column will be rendered before animated transition.

I don't feel it improves perceived performance so much, on my (slow) device which takes some freeze time anyway. Also transition to rendering looks not bad, but different from behavior on swiping.

Well...I wonder that if we make same behavior as swiping, i.e. render destination while/after transition.

PR #4212, opened 2017-07-15
Optimize uri normalization

ClearlyClaire:
Minor optimizations out of my previous “fix-push” PR request (now obsolete) that heavily used URI normalization.
While it provided a ×5 speed-up in that situation, there should be no single task in Mastodon making such heavy use of URI normalization, so the benefits are not that obvious.
Nevertheless, URI normalization happens frequently in Mastodon, so this should contribute to reducing Mastodon's CPU usage overall.


Gargron (approved):

Gargron:
Note: I had to install `libidn11-dev` on my Ubuntu 16.04 to get ruby-idn to install.

beatrix-bitrot:
Yep same, needs to go in update guide/notes!

PR #4215, opened 2017-07-15
Add public key declaration to actors

Gargron:
Reference: https://github.com/w3c/activitypub/issues/203#issuecomment-297553229


Gargron:
cc @cwebber @puckipedia

puckipedia:
LGTM, seems to correspond with what I implemented (except that I link to a key instead of embedding it). 👍

puckipedia:
.. except, why not embed the publicKey if it isn't the root object?

Gargron:
@puckipedia Verbosity mostly

puckipedia:
so I would suggest at least having a `publicKey` pointer, else the acthor object you'd find embedded in an activity isn't the same as the object the id is pointing to...

Gargron:
@puckipedia Okay, publicKey now always included

PR #4216, opened 2017-07-15
Add ActivityPub inbox

Gargron:
- I expect the inbox to receive one Activity at a time
- I intend to handle only Create, Announce and Delete activities
- If the actor of the activity and the signer of the request are not the same, abort

TODO:

- [x] Inbox controller
- [x] Create activity
- [x] Handle mentions
- [x] Handle media attachments
- [x] Handle sensitivity
- [x] Handle hashtags
- [x] Handle audience targeting
- [x] Resolving threads
- [x] Handle status language
- [x] Handle conversations
- [x] Delete activity
- [x] Handle out of order deletes
- [x] Handle out of order undos
- [x] Announce activity
- [x] Like activity
- [x] Undo like activity
- [x] Follow activity
- [x] Undo follow activity
- [x] Block activity
- [x] Undo block activity
- [x] Update profile activity
- [x] Resolving new status by URI
- [x] Resolving new account by URI
- [x] Resolving ActivityPub account by username@domain


akihikodaki:
How about using https://github.com/ruby-rdf/json-ld?

Gargron:
Full json-ld looks very complicated and actual context expanding seems like a performance issue, we have a rather limited use case with AP so I'm okay with just doing it manually for now

puckipedia:
It seems you're expecting the inbox to get an OrderedCollection? Unlike WebSub, ActivityPub just POSTs the Create/Update/Delete/other Activity directly into the inbox:

> An HTTP POST request (with authorization of the submitting user) is then made to to the inbox, with the Activity as the body of the request.
(https://www.w3.org/TR/activitypub/#delivery)

akihikodaki:
> Full json-ld looks very complicated and actual context expanding seems like a performance issue, we have a rather limited use case with AP so I'm okay with just doing it manually for now

Then why don't you just drop `supported_context?`. That just prevents future extensions.

Gargron:
>It seems you're expecting the inbox to get an OrderedCollection

@puckipedia No, I expect to use this parser class for everything though

>Then why don't you just drop supported_context?. That just prevents future extensions.

I think we need to make sure it at least pretends to be ActivityPub JSON. I will change the method to check equality if context is a string, and membership if it's an array.

akihikodaki:
> I think we need to make sure it at least pretends to be ActivityPub JSON. I will change the method to check equality if context is a string, and membership if it's an array.

I still do not see why you still choose to implement that by yourself.

nightpool:
not sure where I stand on full context expansion but I don't think we should be too worried about it in the short term

Gargron:
@cwebber said ActivityPub itself (core) is guaranteed to always been in the correct form (no expanding needed)

akihikodaki:
I'm against because 1. once incomplete implementation is released, that could be persistent for a long time, and 2. it is easy to implement JSON-LD anyway thanks to the gem.

Gargron:
@akihikodaki I don't know what you want me to do with that gem. It does this: <https://github.com/ruby-rdf/json-ld#expand-a-document> which makes the JSON even harder to work with, and which is not required for the ActivityPub standard. It does not do anything special with ActivityStreams type inheritance (Collection/OrderedCollection etc), it does not take over resolving of URIs, I don't see any helpers for working with keys that can contain a string|array|hash. In short, I don't see how it would help us any more than `Oj.load` already does.

akihikodaki:
It is just an easy way to understand `@context`.

akihikodaki:
I was wrong; `@context` of JSON-LD is easy to parse and json-ld gem is unresaonably difficult to use.
In short, the specification requires `@context` to be `https://www.w3.org/ns/activitystreams`, `http://www.w3.org/ns/activitystreams`, an object which has `@vocab` property with one of those URIs, or an array which contains one of them.
json-ld gem provides an API to parse `JSON::LD::Context`, but it needs to have a block to fabricate remote documents for the purpose and that looks difficult.

app/services/activitypub/fetch_remote_account_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    self_reference                       = webfinger.link('self')
+
+    return false if self_reference&.href != @uri
+
+    @username = confirmed_username
+    @domain   = confirmed_domain
+
+    true
+  rescue Goldfinger::Error
+    false
```
beatrix-bitrot: should log here?!

app/lib/activitypub/activity.rb
```diff
@@ -0,0 +1,94 @@
...
+      when 'Like'
+        ActivityPub::Activity::Like
+      when 'Block'
+        ActivityPub::Activity::Block
+      when 'Update'
+        ActivityPub::Activity::Update
+      when 'Undo'
+        ActivityPub::Activity::Undo
+      end
+    end
```
krainboltgreene: Reduced to:

``` ruby
def klass
  ActivityPub::Activity.const_get(@json['type'])
end
```

nightpool:
should be const_get(, inherit=false) for security

app/lib/activitypub/activity.rb
Gargron: Hmm. Given that we're handling json from another server, I think I'll stick to explicitly handling types!

Gargron:
Hey, I just finished writing tests for all of this. You can now review it! Please.

@akihikodaki 
@ThibG 
@nightpool 
@krainboltgreene 
& whoever else wishes to (I don't want to be the only person who understands this code!)

PR #4220, opened 2017-07-16
Fix style settings for images in media modal

lynlynlynx:
Fix issues described by @sorin-davidoi on #4187 

### Shrink too wide single-image media modal.

This makes it easy to close single image.

### Centering images in media modal.

Set margin:auto to center left sided images in multiple-image media modal.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -2969,6 +2969,7 @@ button.icon-button.active i.fa-retweet {
     max-height: 80vh;
     width: auto;
     height: auto;
+    margin: auto;
   }
 
   .extended-video-player,
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1143,8 +1143,8 @@
     top: 0;
     left: 0;
     right: 0;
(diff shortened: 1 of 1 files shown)


ykzts (approved):

sorin-davidoi:
@lynlynlynx This breaks the vertical centering of the videos. Though I like the swipe animation, I wonder if it is worth all these CSS problems were are having.
image

lynlynlynx:
@sorin-davidoi Really? Vertical centering multiple videos is applied on my PCs and phones :thinking:
<details>
<summary>Like this</summary>
screenshot
</details>

But, single-video modal is not shrieked yet, I know.. This is because video size is set 80vw*80vh (for making small gifs easier to watch?).

### About swipe animations.
I know swipe animation is convenient for phones and it is used in most of apps. But fix all problems under current swipe style is not worthy, I think. I have no idea to fix all of them...

If we continue to use swipe animation for image modal, I think we have to change style like>  
Quit fixed left and right buttons, and hover them on images.  
or
Use lightbox (or similar one) plugin.

PR #4221, opened 2017-07-16
fix(emojis): Handle multicode emojis (e.g. country flags)

### LINKED ISSUE #4217: Emoji picker can't insert multi-point emoji after PR #4202 and #4189
Testing on master, the emoji picker is unable to insert composite emoji, such as any of the flags, the family emojis (apart from the first one, which uses only one point), or the skin modifiers (but they didn't work before either).

In addition, the emoji inside the textarea render using the system font, which makes them look ugly and unreadable if you're not on a system with native emoji support.

Also, the emoji tooltips are gone, which was a really useful feature for deciphering hard-to-understand emojis in the timelines.

screenshot_20170716_014206

sorin-davidoi:
Closes #4217.

### DIFF
--- a/app/javascript/mastodon/reducers/compose.js
+++ b/app/javascript/mastodon/reducers/compose.js
@@ -118,7 +118,7 @@ const insertSuggestion = (state, position, token, completion) => {
 };
 
 const insertEmoji = (state, position, emojiData) => {
-  const emoji = String.fromCodePoint(parseInt(emojiData.unicode, 16));
+  const emoji = emojiData.unicode.split('-').map(code => String.fromCodePoint(parseInt(code, 16))).join('');
 
   return state.withMutations(map => {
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/reducers/compose.js
```diff
@@ -118,7 +118,7 @@ const insertSuggestion = (state, position, token, completion) => {
 };
 
 const insertEmoji = (state, position, emojiData) => {
-  const emoji = String.fromCodePoint(parseInt(emojiData.unicode, 16));
+  const emoji = emojiData.unicode.split('-').map(code => String.fromCodePoint(parseInt(code, 16))).join('')
```
Gargron: Missing semicolon

Gargron (changes requested):

Gargron (approved):

PR #4225, opened 2017-07-16
Check table existence on prepare_for_foreign_keys

unarist:
Some tables were not existing at v1.3.3, so check existence to avoid UndefinedTable errors.


Gargron (approved):

unarist:
Oh, thanks!

PR #4226, opened 2017-07-16
fix(push-notifications): Refactor how Sidekiq jobs are handled

sorin-davidoi:
We now send the contact information to the subscription service - I have a suspicion some push requests were refused because this was missing.

Each push subscription now has its own worker - instead of putting N HTTP requests inside a worker (retries did not work properly).

All errors are allowed to bubble up so that they show up in Sidekiq.

cc: @beatrix-bitrot, @nightpool

### DIFF
--- a/app/services/notify_service.rb
+++ b/app/services/notify_service.rb
@@ -65,7 +65,11 @@ class NotifyService < BaseService
   end
 
   def send_push_notifications
-    WebPushNotificationWorker.perform_async(@recipient.id, @notification.id)
+    sessions_with_subscriptions_ids = @recipient.user.session_activations.where.not(web_push_subscription: nil).pluck(:id)
+
+    WebPushNotificationWorker.push_bulk(sessions_with_subscriptions_ids) do |session_activation_id|
+      [session_activation_id, @notification.id]
+    end
   end
 
   def send_email
--- a/app/workers/web_push_notification_worker.rb
+++ b/app/workers/web_push_notification_worker.rb
@@ -5,22 +5,18 @@ class WebPushNotificationWorker
 
   sidekiq_options backtrace: true
 
-  def perform(recipient_id, notification_id)
-    recipient = Account.find(recipient_id)
+  def perform(session_activation_id, notification_id)
+    session_activation = SessionActivation.find(session_activation_id)
     notification = Notification.find(notification_id)
 
-    sessions_with_subscriptions = recipient.user.session_activations.where.not(web_push_subscription: nil)
+    begin
+      session_activation.web_push_subscription.push(notification)
+    rescue Webpush::InvalidSubscription, Webpush::ExpiredSubscription => e
+      # Subscription expiration is not currently implemented in any browser
+      session_activation.web_push_subscription.destroy!
+      session_activation.update!(web_push_subscription: nil)
 
-    sessions_with_subscriptions.each do |session|
-      begin
-        session.web_push_subscription.push(notification)
-      rescue Webpush::InvalidSubscription, Webpush::ExpiredSubscription
-        # Subscription expiration is not currently implemented in any browser
-        session.web_push_subscription.destroy!
-        session.update!(web_push_subscription: nil)
-      rescue Webpush::PayloadTooLarge => e
-        Rails.logger.error(e)
-      end
+      raise e
     end
   end
 end
--- a/app/models/web/push_subscription.rb
+++ b/app/models/web/push_subscription.rb
@@ -12,6 +12,9 @@
 #  updated_at :datetime         not null
 #
 
+require 'webpush'
+require_relative '../../models/setting'
+
 class Web::PushSubscription < ApplicationRecord
   include RoutingHelper
   include StreamEntriesHelper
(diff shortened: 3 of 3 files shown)


app/services/notify_service.rb
```diff
@@ -65,7 +65,11 @@ def create_notification
   end
 
   def send_push_notifications
-    WebPushNotificationWorker.perform_async(@recipient.id, @notification.id)
+    sessions_with_subscriptions = @recipient.user.session_activations.reject { |session_activation| session_activation.web_push_subscription.nil? }
```
akihikodaki: Use `where`, which will be handled by the database.
```Ruby
@recipient.user.session_activations.where(web_push_subscription: nil)
```

app/services/notify_service.rb
```diff
@@ -65,7 +65,11 @@ def create_notification
...
+
+    sessions_with_subscriptions.each do |session_activation|
```
akihikodaki: Use `push_bulk` instead of `perform_async`.

app/workers/web_push_notification_worker.rb
```diff
@@ -5,23 +5,18 @@ class WebPushNotificationWorker
...
-        session.web_push_subscription.push(notification)
-      rescue Webpush::InvalidSubscription, Webpush::ExpiredSubscription
-        # Subscription expiration is not currently implemented in any browser
-        session.web_push_subscription.destroy!
-        session.web_push_subscription = nil
-        session.save!
-      rescue Webpush::PayloadTooLarge => e
-        Rails.logger.error(e)
-      end
+      throw e
```
akihikodaki: Do you mean `raise`?

akihikodaki (changes requested):
Added some Ruby and Rails related comments.

app/workers/web_push_notification_worker.rb
Gargron: It is possible to `throw`/`catch` in Ruby but it's a different way of doing control flow... Not sure what is intended in this case. (<https://stackoverflow.com/questions/51021/what-is-the-difference-between-raising-exceptions-vs-throwing-exceptions-in-ruby>)

akihikodaki: I thought it is meant to raise the error again.

Gargron: You're right, it definitely should be a `raise`.

sorin-davidoi: Yes, it is meant to raise the error again 😄

sorin-davidoi:
Fixed the issues, however, it seems that it takes slightly longer for the push notifications to arrive. The documentation for `push_bulk` states:
> In practice this method is only useful if you are pushing tens of thousands of jobs or more, or if you need to ensure that a batch doesn't complete prematurely.

We are pushing a maximum of 10 jobs at a time. Does it make sense to use it?

akihikodaki:
Keeping using `push_bulk` makes easier to code and refactor (you do not have to consider the number of jobs to be queued.)

app/services/notify_service.rb
```diff
@@ -65,7 +65,12 @@ def create_notification
   end
 
   def send_push_notifications
-    WebPushNotificationWorker.perform_async(@recipient.id, @notification.id)
+    sessions_with_subscriptions = @recipient.user.session_activations.where.not(web_push_subscription: nil)
+
+    Sidekiq::Client.push_bulk(
```
akihikodaki: We use `sidekiq-bulk` gem.

app/services/notify_service.rb
```diff
@@ -65,7 +65,12 @@ def create_notification
...
+      'class' => WebPushNotificationWorker,
+      'args' => sessions_with_subscriptions.map { |session_activation| [session_activation.id, @notification.id] }
```
akihikodaki: `pluck`, and then `map`.

akihikodaki (changes requested):

app/services/notify_service.rb
sorin-davidoi: Not sure what you mean, I basically copied the pattern from [here](https://github.com/tootsuite/mastodon/blob/master/app/services/fan_out_on_write_service.rb#L40).

akihikodaki: Oh, that implementation is actually not really good. You can do something like this:
```Ruby
sessions_with_subscriptions.pluck(:id).map { |session_activation_id| [session_activation_id, @notification.id] }
```

akihikodaki: That implementation was fine. It uses `select(:id)`. And it gives a block for `push_bulk`, which would be better than calling `map`.

app/services/notify_service.rb
```diff
@@ -65,7 +65,9 @@ def create_notification
   end
 
   def send_push_notifications
-    WebPushNotificationWorker.perform_async(@recipient.id, @notification.id)
+    sessions_with_subscriptions = @recipient.user.session_activations.where.not(web_push_subscription: nil)
+
+    WebPushNotificationWorker.push_bulk(sessions_with_subscriptions.pluck(:id).map { |session_activation_id| [session_activation_id, @notification.id] })
```
akihikodaki: Just in case you missed: you can give a block to `push_bulk` of sidekiq-bulk.
https://github.com/tootsuite/mastodon/blob/master/app/services/fan_out_on_write_service.rb#L40

akihikodaki (dismissed):

Gargron (approved):

PR #4228, opened 2017-07-16
feat(push-notifications): Open link in current tab if possible

sorin-davidoi:
Does not break functionality if the website is added to the home screen.

cc: @beatrix-bitrot

### DIFF
--- a/app/javascript/mastodon/service_worker/entry.js
+++ b/app/javascript/mastodon/service_worker/entry.js
@@ -1,1 +1,10 @@
 import './web_push_notifications';
+
+// Cause a new version of a registered Service Worker to replace an existing one
+// that is already installed, and replace the currently active worker on open pages.
+self.addEventListener('install', function(event) {
+  event.waitUntil(self.skipWaiting());
+});
+self.addEventListener('activate', function(event) {
+  event.waitUntil(self.clients.claim());
+});
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -50,6 +50,24 @@ const makeRequest = (notification, action) =>
     credentials: 'include',
   });
 
+const openUrl = url =>
+  self.clients.matchAll({ type: 'window' }).then(clientList => {
+    if (clientList.length !== 0 && 'navigate' in clientList[0]) { // Chrome 42-48 does not support navigate
+      const webClients = clientList
+        .filter(client => /\/web\//.test(client.url))
+        .sort(client => client !== 'visible');
+
+      const visibleClient = clientList.find(client => client.visibilityState === 'visible');
+      const focusedClient = clientList.find(client => client.focused);
+
+      const client = webClients[0] || visibleClient || focusedClient || clientList[0];
+
+      return client.navigate(url).then(client => client.focus());
+    } else {
+      return self.clients.openWindow(url);
+    }
+  });
+
 const removeActionFromNotification = (notification, action) => {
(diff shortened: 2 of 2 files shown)


ykzts (approved):

beatrix-bitrot:
not working for me on desktop chrome but i will test now on mobile...

sorin-davidoi:
@beatrix-bitrot Refresh,  close all Mastodon tabs and then open a new one. Service workers do not take control of the clients immediately.

sorin-davidoi:
Is working for me (on glitch.social).

beatrix-bitrot:
Working on mobile. Won't be able to test desktop for a while

beatrix-bitrot:
Actually an issue I have now... It doesn't open the thing the notification is about... Just... Some arbitrary one from before.... But it's the same one every time?!

sorin-davidoi:
Cannot reproduce :cry:

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -75,7 +87,7 @@ const handleNotificationClick = (event) => {
       }
     } else {
       event.notification.close();
-      resolve(self.clients.openWindow(event.notification.data.url));
+      resolve(openUrl(event.notification.data.url));
```
Gargron: I don't know how handleNotificationClick is attached exactly, but I remember reading that React events are synthetic and are reused between calls. So perhaps the thing @beatrix-bitrot is experiencing is event.notification.data.url referring to something from an earlier call. You could try making a copy of the data you're interested in outside of the Promise.

sorin-davidoi:
This is code that runs in the Service Worker, where you don't have access to the DOM - no React, no DOM APIs. Will try to reproduce this again later today.

sorin-davidoi:
Still not able to reproduce. I've changed a bit the logic that determines which client is focused (e.g. if one tab is on `/web` and another on `/settings`, open the link in `/web`). I've also added some code to skip waiting when a new service worker is installed - this is perfectly fine since we do not have to deal with caches and assets - although Nolan will probably change this in his PR.

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -50,6 +50,24 @@ const makeRequest = (notification, action) =>
     credentials: 'include',
   });
 
+const openUrl = url =>
+  self.clients.matchAll({ type: 'window' }).then(clientList => {
+    if (clientList.length !== 0 && 'navigate' in clientList[0]) { // Chrome 42-48 does not support navigate
+      const webClients = clientList
+        .filter(client => /\/web\//.test(client.url))
+        .sort(client => client !== 'visible');
```
sorin-davidoi: If there are multiple webClients, if one is visible it should be the first.

app/javascript/mastodon/service_worker/entry.js
```diff
@@ -1 +1,10 @@
 import './web_push_notifications';
+
+// Cause a new version of a registered Service Worker to replace an existing one
+// that is already installed, and replace the currently active worker on open pages.
+self.addEventListener('install', function(event) {
```
sorin-davidoi: @nolan Am I right in assuming that this is safe to do since we do not cache and there is no communication between the clients and the Service Worker?

nolanlawson: Yes `skipWaiting` should be fine in this case. Since we're not caching, there's no situation I can think of where the "frontend" would get out of sync with the "backend" (service worker).

PR #4231, opened 2017-07-17
Require any modules after loading polyfill in entry points

akihikodaki:
`app/javascript/mastodon/main.js` delayed the execution of modules, but other entry points didn't. That leads to failure in executing modules, which requires those polyfills.

Strictly enforce the rule to require any modules after loading polyfill in entry points.


Gargron (approved):

ykzts:
resolve conflicts.

PR #4250, opened 2017-07-18
Fix broken CSS reload due to webpack upgrade

MightyPork:
Rolled back webpack upgrade to `^3.2.0` from c45a75ad3419f15d639138ac34c320323367c870 that caused this JS error after a SCSS file was changed on a dev server, making CSS testing very painful.

"extract-text-webpack-plugin" warns it has a wrong peer dependency, but it does not seem to be an issue. 

Both are reverted to their state before c45a75ad3419f15d639138ac34c320323367c870 which I found to be the culprit using `git bisect` on @Gargron's request after whining about the bug ;)

screenshot_20170718_170144


ykzts (approved):

ykzts:
I think better downgrade. I will investigate the details later.

PR #4253, opened 2017-07-18
feat(emoji): Add back title attribute

sorin-davidoi:
By popular demand, this adds back the title attribute for the rendered emojis (which was removed in #4202).

`emojione_light` increases by 100%: 32KB -> 64KB (still an improvement compared to the 288KB `emojione` has).

### DIFF
--- a/app/javascript/mastodon/emoji.js
+++ b/app/javascript/mastodon/emoji.js
@@ -1,7 +1,7 @@
-import { unicodeToFilename } from './emojione_light';
+import { unicodeMapping } from './emojione_light';
 import Trie from 'substring-trie';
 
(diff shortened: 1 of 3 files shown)


sorin-davidoi:
The failing build does not seem to be related to this PR.

Gargron (approved):

Gargron:
It's a JS test failing because it was not adjusted:

    1) does unicode
    2) does multiple unicode

sorin-davidoi:
Yes, you're right. Saw some unrelated errors and did not scroll further.

nolanlawson:
Seems worth the cost. Discoverability of shortcodes is nice! :smiley:

PR #4258, opened 2017-07-18
Refactor ResolveRemoteAccountService

Gargron:
This class was always a mess. Unfortunately, I cannot completely get rid of the complexity. But hopefully it will be easier to add OStatus vs ActivityPub handling in here.

Also, I replaced the catch for RecordNotUnique exception with a transaction block. I am not 100% sure that will do the same thing, but it might work. If not, I can try adding a redis lock.

### DIFF
--- a/app/services/resolve_remote_account_service.rb
+++ b/app/services/resolve_remote_account_service.rb
@@ -11,97 +11,146 @@ class ResolveRemoteAccountService < BaseService
   # @param [String] uri User URI in the form of username@domain
   # @return [Account]
   def call(uri, update_profile = true, redirected = nil)
-    username, domain = uri.split('@')
+    @username, @domain = uri.split('@')
 
-    return Account.find_local(username) if TagManager.instance.local_domain?(domain)
+    return Account.find_local(@username) if TagManager.instance.local_domain?(@domain)
 
-    account = Account.find_remote(username, domain)
-    return account unless account_needs_webfinger_update?(account)
+    @account = Account.find_remote(@username, @domain)
+
+    return @account unless webfinger_update_due?
 
     Rails.logger.debug "Looking up webfinger for #{uri}"
 
-    data = Goldfinger.finger("acct:#{uri}")
+    @webfinger = Goldfinger.finger("acct:#{uri}")
 
-    raise Goldfinger::Error, 'Missing resource links' if data.link('http://schemas.google.com/g/2010#updates-from').nil? || data.link('salmon').nil? || data.link('http://webfinger.net/rel/profile-page').nil? || data.link('magic-public-key').nil?
+    raise Goldfinger::Error, 'Missing resource links' if links_missing?
 
-    # Disallow account hijacking
-    confirmed_username, confirmed_domain = data.subject.gsub(/\Aacct:/, '').split('@')
+    confirmed_username, confirmed_domain = @webfinger.subject.gsub(/\Aacct:/, '').split('@')
 
-    unless confirmed_username.casecmp(username).zero? && confirmed_domain.casecmp(domain).zero?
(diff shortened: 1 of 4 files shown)


ThibG:
I am pretty sure the transaction block *won't* have the same effect. Looks good to me otherwise.

Gargron:
@ThibG Alright, I'll give redis locks a go!

Gargron:
@ThibG I'm not sure how to test this, but I added a lock around the critical part.

app/services/resolve_remote_account_service.rb
```diff
@@ -11,97 +11,153 @@ class ResolveRemoteAccountService < BaseService
...
+
+    return @account unless webfinger_update_due?
 
     Rails.logger.debug "Looking up webfinger for #{uri}"
 
-    data = Goldfinger.finger("acct:#{uri}")
+    @webfinger = Goldfinger.finger("acct:#{uri}")
 
-    raise Goldfinger::Error, 'Missing resource links' if data.link('http://schemas.google.com/g/2010#updates-from').nil? || data.link('salmon').nil? || data.link('http://webfinger.net/rel/profile-page').nil? || data.link('magic-public-key').nil?
+    raise Goldfinger::Error, 'Missing resource links' if links_missing?
```
Gargron: The only reason this was made to raise this error is so we could display feedback to the user. But I am seeing sidekiq jobs failing *and retrying* with that error, which should not happen, because there's no point in retrying from this error. I haven't changed it because I didn't want to change functionality in the same PR, but this needs to be changed.

app/services/resolve_remote_account_service.rb
```diff
@@ -11,97 +11,153 @@ class ResolveRemoteAccountService < BaseService
...
+
+    @hub_url = hubs.first['href']
+  end
+
+  def atom_body
+    return @atom_body if defined?(@atom_body)
+
+    response = Request.new(:get, atom_url).perform
+
+    raise Goldfinger::Error, "Feed attempt failed for #{atom_url}: HTTP #{response.code}" unless response.code == 200
```
Gargron: This on the other hand, *should* be a retryable error.

app/services/resolve_remote_account_service.rb
```diff
@@ -11,97 +11,153 @@ class ResolveRemoteAccountService < BaseService
...
+
+    @canonical_uri = author_uri.content
+  end
+
+  def hub_url
+    return @hub_url if defined?(@hub_url)
+
+    hubs = atom.xpath('//xmlns:link[@rel="hub"]')
+
+    raise Goldfinger::Error, 'No PubSubHubbub hubs found' if hubs.empty? || hubs.first['href'].nil?
```
Gargron: This should not get retried.

app/services/resolve_remote_account_service.rb
```diff
@@ -11,97 +11,153 @@ class ResolveRemoteAccountService < BaseService
...
+    confirmed_username, confirmed_domain = @webfinger.subject.gsub(/\Aacct:/, '').split('@')
 
-    unless confirmed_username.casecmp(username).zero? && confirmed_domain.casecmp(domain).zero?
+    if confirmed_username.casecmp(@username).zero? && confirmed_domain.casecmp(@domain).zero?
+      @username = confirmed_username
+      @domain   = confirmed_domain
+    else
       return call("#{confirmed_username}@#{confirmed_domain}", update_profile, true) if redirected.nil?
-      raise Goldfinger::Error, 'Requested and returned acct URI do not match'
+      raise Goldfinger::Error, 'Requested and returned acct URIs do not match'
```
Gargron: This should not get retried.

beatrix-bitrot (approved):

PR #4271, opened 2017-07-19
fix(push_subscription): Guard against malformed favourite notification

sorin-davidoi:
~~Would be nice to filter out "unpushable" notifications and these ones outside of Sidekiq, to avoid creating useless jobs - will try to follow up with that.~~

Jobs are created only for pushable notifications.

### DIFF
--- a/app/models/web/push_subscription.rb
+++ b/app/models/web/push_subscription.rb
@@ -26,8 +26,6 @@ class Web::PushSubscription < ApplicationRecord
   before_create :send_welcome_notification
 
   def push(notification)
-    return unless pushable? notification
-
     name = display_name notification.from_account
     title = title_str(name, notification)
     body = body_str notification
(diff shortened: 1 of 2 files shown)


app/models/web/push_subscription.rb
```diff
@@ -28,6 +28,10 @@ class Web::PushSubscription < ApplicationRecord
   def push(notification)
     return unless pushable? notification
 
+    # HACK: Can be caused by quickly unfavouriting a status, since creating
+    # a favourite and creating a notification are not wrapped in a transaction.
+    return if notification.type.equal?(:favourite) && notification.target_status.nil?
```
Gargron: `return if notification.activity.nil?`

Gargron (changes requested):

Gargron (approved):

PR #4275, opened 2017-07-19
Fix webfinger retries

Gargron:
The code flow is hard to understand here so I would really like some thorough reviews on this. I want to stop `ResolveRemoteAccountService` from raising exceptions which there is no reason to retry (such as incompatible webfinger links, or missing user on server). On the other hand, I want to put all retryable errors into their own class instead of using `RuntimeError` - `Mastodon::UnexpectedResponseError`.

### DIFF
--- a/app/services/resolve_remote_account_service.rb
+++ b/app/services/resolve_remote_account_service.rb
@@ -50,6 +50,9 @@ class ResolveRemoteAccountService < BaseService
     end
 
     @account
+  rescue Goldfinger::Error => e
+    Rails.logger.debug "Webfinger query for #{uri} unsuccessful: #{e}"
+    nil
   end
 
   private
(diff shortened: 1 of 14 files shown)


app/services/resolve_remote_account_service.rb
```diff
@@ -123,27 +129,22 @@ def canonical_uri
       author_uri = owner.at_xpath('./xmlns:uri') unless owner.nil?
     end
 
-    raise Goldfinger::Error, 'Author URI could not be found' if author_uri.nil?
-
-    @canonical_uri = author_uri.content
+    @canonical_uri = author_uri.nil? ? nil : author_uri.content
```
ykzts: `author_url&.content`

Gargron: If author_uri cannot be nil in the else-clause

ykzts: hmm. I see

ykzts (approved):

PR #4281, opened 2017-07-20
Add callback_url/acct information for Sidekiq PuSH workers Exception.

clworld:
Some exception does not report hostname in exception message.
So, hostname does not shown in Sidekiq error console.
You need check database tables to know which host returns error.
With this PR, you can easily see which server is in trouble...
sidekiq_fix3

### DIFF
--- a/app/lib/exceptions.rb
+++ b/app/lib/exceptions.rb
@@ -8,11 +8,11 @@ module Mastodon
 
   class UnexpectedResponseError < Error
     def initialize(response = nil)
-      @response = response
-    end
-
-    def to_s
-      "#{@response.uri} returned code #{@response.code}"
+      if response.respond_to? :url
+        super("#{response.uri} returned code #{response.code}")
+      else
+        super
+      end
     end
   end
 end
--- a/app/lib/exceptions.rb
+++ b/app/lib/exceptions.rb
@@ -8,7 +8,7 @@ module Mastodon
 
   class UnexpectedResponseError < Error
     def initialize(response = nil)
-      if response.respond_to? :url
+      if response.respond_to? :uri
         super("#{response.uri} returned code #{response.code}")
       else
         super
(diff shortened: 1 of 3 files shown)


beatrix-bitrot (approved):
YES PLEASE

beatrix-bitrot:
ok maybe fix the broken tests d:

still... i really want this

app/lib/exceptions.rb
```diff
@@ -8,11 +8,11 @@ class RaceConditionError < Error; end
 
   class UnexpectedResponseError < Error
     def initialize(response = nil)
-      @response = response
-    end
-
-    def to_s
-      "#{@response.uri} returned code #{@response.code}"
+      if response.respond_to? :url
+        super("#{response.uri} returned code #{response.code}")
```
nightpool: url vs uri?

clworld: Oh. mistake…

nightpool:
What is the difference between "response.uri" and "acct"? One gives more meaningful information, in my opinion.

nightpool:
This feels like it's redundant now that we use MastodonUnexpectedResponse.

clworld:
I'm confused about RSpec executes Sidekiq worker directly and exceptions thrown by workers returns to caller...
I choose way that don't change Exception class. but this way assumes all StandardError are reproducible..
Another way is wrap errors by some Exception (Mastodon::WorkerError). It breaks rspec which checks which exception is raised.

clworld:
@nightpool 
responce.uri is what UnexpectedResponseError show on .message.
I don't change UnexpectedResponseError's output. I changed UnexpectedResponseError to make it reproducible.

What I changed in this PR is add acct/callback_url to Exceptions which Sidekiq workers raises.
callback_url is url to post OStatus payload.
acct is Mastodon account name to subscribe. For example: "clworld@md.ggtea.org" is my account's acct.

nightpool:
okay, that makes sense. LGTM

unarist (commented):
Looks good, but I'm not sure about `raise e.class "..."` is proper way or not.

clworld:
I send alternative implementation: https://github.com/tootsuite/mastodon/pull/4283
which don't use 'raise e.class "..."' thing. but wrap all exceptions.

Gargron (approved):

PR #4282, opened 2017-07-20
feat(compose): More space on mobile devices

sorin-davidoi:
Writing toots on mobile devices is a sub par experience due to the lack of space in the compose column - when the virtual keyboard covers the screen you don't see what you are typing and have to scroll to tap any of the buttons. This PR hides the tabs and search bar if the compose area is focused.

I would prefer to hide the navigation bar (avatar, username, edit profile) as well, but I don't really know where to place the "close" button then - it would be too close to the emoji picker. `react-motion` is not used because that could mean we have to connect more components to the state (plus we need media queries). We could probably improve the performance of the animation, but I don't really know how - changing `margin-top` is pretty computationally expensive since it triggers re-layout.

This also raises the problem of auto focusing the `textarea`, which implies that switching to the compose column automatically triggers the hiding of the two bars.

# Before

# After

# Animation


sorin-davidoi:
New animation - with a compact navigation bar - might be a bit too much?

Gargron:
@sorin-davidoi not too much but definitely too slowly

sorin-davidoi:
The recording distorted it a bit, it is faster. But we can definitely tweak it.

sorin-davidoi:
Tested on:
 - [x] Chrome, Firefox, Epiphany (Safari-like) on Linux
 - [x] Chrome on Android
 - [x] Edge 15

If someone could see how this looks on iOS that would be great.

Gargron (approved):

ThisIsMissEm:
I think this PR has introduced a regression (without knowing it), whereby there is now no close button on compose.

PR #4284, opened 2017-07-20
Use the same emoji data on the frontend and backend

Gargron:


app/lib/emoji.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'singleton'
+
+class Emoji
+  include Singleton
+
+  def initialize
+    data = Oj.load(File.open(File.join(Rails.root, 'node_modules', 'emojione', 'emoji.json')))
```
beatrix-bitrot: must we load it from node modules...?

Gargron: I'm not sure what else we can do. There is a problem with this, of course - now you must install node dependencies even on machines that would only run ruby workers.

MightyPork: this may be a heretic idea, but what about we simply copy the json from the package somewhere else in the codebase and update it manually or with some script from time to time?

beatrix-bitrot (approved):
in order for it to really be THE SAME emoji used in both front and back end it kinda has to be this way (taking the emoji in ruby from node modules) since that's where the emoji for the front end come from

be nice to revisit this maybe but for now i think it's fine

app/lib/emoji.rb
beatrix-bitrot: fuck @Gargron that is a bad one some instances may have those decoupled...

beatrix-bitrot: in that case i think the annoying solution is best and using piggo's heretic idea may be the way to go

IDEALLY the separate, versioned emoji json would be the one built into the front end rather than the one from node modules

is that tough?!

ykzts (approved):

PR #4286, opened 2017-07-21
Update Goldfinger gem to 2.0

Gargron:
- No masking of HTTP::Error and OpenSSL::SSL::SSLError
- No longer accepts non-HTTPS WebFinger endpoints


beatrix-bitrot (approved):
fix conflicts and you're good

nightpool:
This will need a mention in the release notes when we get there

PR #4295, opened 2017-07-21
fix(dropdown_menu): Open as modal on mobile

sorin-davidoi:
The dropdown has a few problems on mobile:
 - It is quite small, actions hard to tap
 - It is aligned to the right, which works great for the fixed columns but here looks weird (PR: #3791 takes a shot at this)
 - Can open outside the viewport, forcing the user to scroll
 - Actions cannot be taped if for the last status in the list (dropdown opens outside the viewport, but you cannot scroll down)

This keeps the same API but opens the dropdown as a modal in the mobile view (which it more in line with how native apps approach this problem).

# Before

# After

### DIFF
--- a/app/javascript/mastodon/features/ui/util/async-components.js
+++ b/app/javascript/mastodon/features/ui/util/async-components.js
@@ -102,10 +102,6 @@ export function ReportModal () {
   return import(/* webpackChunkName: "modals/report_modal" */'../components/report_modal');
 }
 
-export function ActionsModal () {
-  return import(/* webpackChunkName: "modals/actions_modal" */'../components/actions_modal');
-}
-
 export function MediaGallery () {
   return import(/* webpackChunkName: "status/MediaGallery" */'../../../components/media_gallery');
 }
--- a/app/javascript/mastodon/features/ui/util/async-components.js
+++ b/app/javascript/mastodon/features/ui/util/async-components.js
@@ -102,6 +102,10 @@ export function ReportModal () {
   return import(/* webpackChunkName: "modals/report_modal" */'../components/report_modal');
 }
 
+export function ActionsModal () {
+  return import(/* webpackChunkName: "modals/actions_modal" */'../components/actions_modal');
+}
+
 export function MediaGallery () {
   return import(/* webpackChunkName: "status/MediaGallery" */'../../../components/media_gallery');
 }
--- a/app/javascript/mastodon/components/dropdown_menu.js
+++ b/app/javascript/mastodon/components/dropdown_menu.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import ImmutablePropTypes from 'react-immutable-proptypes';
 import Dropdown, { DropdownTrigger, DropdownContent } from 'react-simple-dropdown';
 import PropTypes from 'prop-types';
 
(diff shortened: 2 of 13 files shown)


unarist:
Looks good on mobile device (though I doesn't think dropdown is so bad).

However, I also often use "mobile view" on PC, as like a sidebar. So I'm worried about it may requires more mouse movement. Also can we have a hover effect?

sorin-davidoi:
Doesn't it require mouse movement now though? Since the dropdown does not open next to the expand button.

Will try to add an hover effect, good point.

unarist:
New modal will open on vertically middle on window, even if I clicked expand button on the status on the bottom. So it may requires to move cursor about half of the display height.

image image

Well...I thought that is "long", but it may not I think now.

sorin-davidoi:
Ah. Ideally, this would not be based on `max-width` but on whether the dropdown was triggered by a touch event. Will see if this is even possible.

sorin-davidoi:
@unarist I've updated to only open the modal when the user touches the screen (which means it also appears on tablets which is great!).

sorin-davidoi:
Added the status preview - confirmation that you didn't fat-fingered the wrong toot:
image

unarist:
Since a status can be so long, so you should add `overflow-y: auto` and also `max-height` (or `flex-shrink: 1` to `<ul>`).

image

Gargron:
If you're gonne include a status preview I suggest making the modal the same width as the boost modal which also includes a preview (and generally making it consistent with that). I'm also not sure if centered text looks good here. Why not a left-aligned menu like the actual dropdown?

sorin-davidoi:
Updated:

image

sorin-davidoi:
Anything else I should change?

unarist:
While testing on iPhone 5s (320x568, or 320x460 w/o top/bottom bars), sometimes I felt that it's hard to close modal due to lack of touch area outside the modal.

image

Some thoughts:

* Add max-height or/and max-width to whole modal on portrait viewport
(also we can decrease height of menu items a bit?)
* Add close button

Note that this issue applies for other modals w/ status.

sorin-davidoi:
@unarist Went with max-height / max-width and reduced padding (in some cases there are already enough buttons in that list).

sorin-davidoi:
Updated the privacy dropdown

app/javascript/mastodon/features/ui/util/async-components.js
```diff
@@ -102,6 +102,10 @@ export function ReportModal () {
   return import(/* webpackChunkName: "modals/report_modal" */'../components/report_modal');
 }
 
+export function ActionsModal () {
+  return import(/* webpackChunkName: "modals/actions_modal" */'../components/actions_modal');
```
unarist: Should we have this as separate chunk? Since it's only 2KB in plain and 945 bytes in gzipped, I think it's better that include it in application pack to avoid loading delay and showing empty modal, though this may also apply to other modals.

sorin-davidoi: Good point :+1:

unarist (approved):
Looks good!

(I found swipe on modal causes scroll on status_list, and maybe I want to add hover style for actions_modal, but it can be another PR I think)

sorin-davidoi:
Since the actions modal appears only for touch devices, I don't think hover is needed?

unarist:
Ah yes, it's may not be "hover". I thought that I want to see what option touching now (before touchend).

PR #4297, opened 2017-07-22
[WIP] fix(status_list): Hide column header when scrolling down

sorin-davidoi:
Collapses the header to make more space for the actual content. This is very poorly implemented right now, opened the PR to see if there is interest in this feature.

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -33,6 +33,17 @@ export default class StatusList extends ImmutablePureComponent {
   handleScroll = debounce(() => {
     const { scrollTop, scrollHeight, clientHeight } = this.node;
     const offset = scrollHeight - scrollTop - clientHeight;
+
+    if (this._oldScrollPosition) {
+      const isScrollingDown = this._oldScrollPosition > scrollHeight - scrollTop;
+
+      if (scrollTop >= 250 || (!isScrollingDown)) {
+        window.requestAnimationFrame(() => {
+          document.querySelector('div[aria-hidden="false"][data-swipeable="true"]').classList.toggle('collapsed', isScrollingDown);
+        });
+      }
+    }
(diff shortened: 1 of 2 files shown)


DJSundog:
I'm generally not a fan of UI chrome that hides as a side effect of another UI interaction, but I think lots of folks like this, so my only request would be yet another toggle in the app settings allowing the end user to disable this functionality should it be implemented.

Gargron:
I am against this change, the header gives a quick scroll-to-top link as well as the context of what you are scrolling through.

sorin-davidoi:
Context is given by the tabs, but fair point - the scroll-to-top link disappears.

rysiekpl:
Scroll to top could be placed somewhere else, screen real estate is very valuable on mobile.

sorin-davidoi:
Perhaps I should have specified that this would only be enabled for specific media queries (`max-width: 600px; max-height: 800px` or whatever makes sense).

tomfhowe:
What about just shrinking it? The font could get smaller, the padding could decrease? On most tabs it feels like a big waste of space... Also I had no idea it was a back to top link until now... maybe an actual back to top button that appears when you start scrolling up could be more obvious?

tomfhowe:
The thing with scrolling back to the top is like I always used the tab buttons at the top to do that

sorin-davidoi:
I'll leave it open for visibility, perhaps it will lead to a better solution.

beatrix-bitrot:
Yeah if scroll to top is still possible even when the column head is hidden then I think it's ok?

akihikodaki:
I think `*` (Getting Started button) could be replaced with the icon according to the content if the content is not listed in the navigation.
Users can still go to Getting Started column by clicking "Back" button shown on the column head. (Though that is less intuitive for the single-column layout, that is a different topic.)

PR #4308, opened 2017-07-23
feat(tabs_bar): Placed at bottom (optional)

amithm7:
* Tabs bar at bottom is more accessible. Can be changed in Preferences.
* ~~Removing margin makes icons little bigger for fingers.~~
* ~~Removing padding in columns is more aesthetic for mobile~~

**EDIT** : Restored `360px` break-point, which puts back margin and padding as it was before.


amithm7:
Screens:
screenshot_20170723-110327
screenshot_20170723-110433

sorin-davidoi:
I personally like how this looks. A few remarks:
 - I feel that some people will want the old behavior - perhaps this should be an opt-out?
 - Though I am not an iPhone user, I feel compelled to point at the "bottom navigation" section of [this article](https://popmotion.io/blog/20170710-mobile-web-is-awful-and-were-all-to-blame/)
 - The column sometimes has a margin - looks inconsistent

image

amithm7:
@sorin-davidoi 
* Hmm, i don't have an iPhone. I wouldn't know. So, this is a fail?
* Column's height get's reduced by 20px, when width > 360px. Hence, it seems like a border. I didn't feel the need to remove it when screen is wide enough.

MightyPork:
fwiw dev.glitch.social has, as of yesterday, this implemented as a config option via our Settings Modal (App Settings), if anyone wanted to try it live.

sorin-davidoi:
> Hmm, i don't have an iPhone. I wouldn't know. So, this is a fail?

No, would just like some input from someone who can test this on an iPhone.

> Column's height get's reduced by 20px, when width > 360px. Hence, it seems like a border. I didn't feel the need to remove it when screen is wide enough.

I just feel that this looks a bit off (having the margin in portrait mode but not in horizontal mode):

ndarville:
There’s something about it that looks ... asymmetric? Maybe I’m just crazy.

I think the number of buttons is a little overwhelming, and some vertical borders between them would probably improve on that. :)

It would also help clarify their hitboxes.

sturmen:
As someone with a 5.7" screen on my smartphone (LG V20), I appreciate bottom navigation very much. The iPhone invisible touch target is a real concern, and I suggest we either leave it on top for them (user agent filtering?) Or perhaps make this an opt-in feature.

michcioperz:
I think it's important to have an option to keep it on top of the screen – for me it's natural to look for navigation there and not on the bottom.

ndarville:
The screenshot with both a top and bottom bar makes me think there has to be an alternative to two bars taking up a third of the screen. The top bar taking up all that space seems like such a waste, and viewing media is probably going to be a little tricky as a result.

I don’t if moving down the Notifications bell icon to the left of the bottom bar and the filter to the right is the solution—another option is of course to merge the bottom bar with the top bar—but there’s plenty of space to go around in the bottom and due to the fluid hitboxes, they buttons appear to take up a lot of space.

sorin-davidoi:
@ndarville Checkout #4297 for a proposal on hiding the top bar when scrolling.

joyeusenoelle:
I use an iPhone, and can confirm that a bottom-placed navigation in the browser would be unusable. In addition, I keep my desktop web browser at a width just under the bizarrely-wide 1024px mobile breakpoint, so having the navigation bar at the bottom would conflict with the behavior of every other menu.

This should definitely be optional; it breaks Mastodon usability in too many cases to be mandatory.

Gargron:
@joyeusenoelle What should the breakpoint be instead of 1024px?

sorin-davidoi:
I'm pretty sure the iPhone problem can be solved if anyone wants to tackle that.

ndarville:
Maybe we can set up some tasks so we’re all on the same page about what needs an up- or downvote?

* Top vs bottom bar
* Merge bars (`if device_height < max`)
* Add button borders for visuals and to show hitboxes
* `max-width` for buttons
* `max-width` for buttons to keep them closer together with smaller hitboxes
* Symmetric button placement with alignment
* Something something desktop breakpoints

Another one to consider is to offset the large hitbox issue by scaling the button icons, assuming they’re SVG.

I’ll let a maintainer decide what should be decided on and how, but I just wanted to recapitulate everything that’s going on already.

sorin-davidoi:
Shouldn't this be just for top vs bottom bar? We could open an issue for the others.

joyeusenoelle:
@Gargron The last time we talked about this, I think we settled on 768px.

clarcharr:
I personally don't like this because it forces me to move my hands all the way across the screen if I want to interact with both bars.

Perhaps there's a way to combine them both to make things easier?

ndarville:
One potential issue with placing the bar at the top is that people with large-ish phones `>4"` might find it strenuous to reach the top, especially if you’re doing a tootstorm.

Works fine for desktop, less so for large mobile form factors.

I’m a diehard iPhone 5S fan, so maybe people who use larger phones with one hand can speak to whether a top bar would become a problem for them.

clarcharr:
IMO making it harder to reach the top is a good thing, because then I won't accidentally press the buttons. And I swipe most of the time to get between tabs anyway.

ndarville:
OTOH, placing it at the top on at least iOS might carry the risk for users of hitting the status bar and trigger the scroll-to-top behaviour that drives me absolute insane when I hit the "Back to Timeline" button in Tweetbot, miss, and reset my tweet position.

Pros and cons, but important to list them all, small phones, big phones, desktop etc.

sorin-davidoi:
I've opened a pull request to this branch to make it optional. I think this should please everyone (iPhone user, people with larger screen, people who use the mobile UI on desktop...).

sorin-davidoi:
@Gargron On / off by default?

amithm7:
Should i remove the aesthetic changes i made with padding and margin removal for width > 360px?

sorin-davidoi:
@amithm7 You still have a merge conflict.

Gargron:
I think people are used to tabs at the top, so it would be bad to change that. Sorry for taking long to close this.

PR #4320, opened 2017-07-23
fix(columns_area): Manually set tabs style when swiping

### LINKED ISSUE #4319: Update tab bar before column loading
When I change column by swiping, it looks like 1. swipe and dummy column appears 2. column appears 3. tab bar is updated. I think it would be nice if we update tab bar when the swipe animation completes.

hoge

cc @sorin-davidoi

sorin-davidoi:
Closes #4319.

Don't really like this solution, feels hacky.

cc: @unarist


unarist:
It works smoothly. Looks good!

...except it's very hacky and non-react way.

sorin-davidoi:
Yes, feels very weird to work directly with the DOM. Can't think of any other solution that would not involve some refactoring.

ykzts (approved):

PR #4321, opened 2017-07-23
feat: Enable push subscription for mobile devices by default

sorin-davidoi:
Mobile devices do not support regular notifications. So we enable push notifications by default for these devices. I wish there was a better way to detect this, but it is impossible to reliably feature-detect this in the browsers (e.g. Chrome exposes the Notification API but throws errors when invoked).

### DIFF
--- a/app/controllers/api/web/push_subscriptions_controller.rb
+++ b/app/controllers/api/web/push_subscriptions_controller.rb
@@ -6,8 +6,8 @@ class Api::Web::PushSubscriptionsController < Api::BaseController
   before_action :require_user!
 
   def create
-    params.require(:data).require(:endpoint)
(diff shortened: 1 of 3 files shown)


app/controllers/api/web/push_subscriptions_controller.rb
```diff
@@ -16,10 +16,23 @@ def create
       active_session.update!(web_push_subscription: nil)
     end
 
+    # Mobile devices do not support regular notifications, so we enable push notifications by default
+    alerts_enabled = active_session.detection.device.mobile? || active_session.detection.device.tablet?
```
sorin-davidoi: This is basically Chrome and Firefox on Android.

Gargron (approved):

PR #4322, opened 2017-07-23
Redesign extended information page

Gargron:
When contact information is not set:

image

When contact information is set:

image

Mobile:

image

Textual changes:

- Contact info placeholder is shown even when contact info is not set
- "What is this instance" text uses the same default value as frontpage
- When extended site description is not set, displays headline "A good place for rules" and body "The extended description has not been set up yet"


app/javascript/styles/about.scss
```diff
@@ -120,18 +120,55 @@
...
-  margin: 20px 0;
-  display: flex;
-  justify-content: space-between;
-  border-top: 1px solid lighten($ui-base-color, 10%);
-  border-bottom: 1px solid lighten($ui-base-color, 10%);
-  padding-right: 14px;
+  background: darken($ui-base-color, 4%);
+  padding: 40px 0;
+
+  .panel {
```
nightpool: mostly just a comment, but the heavily nested scss style is much harder to read and maintain then the BEM style we use in the rest of the app. it makes it harder to override the content on the instance-owner side as well.

app/javascript/styles/about.scss
```diff
@@ -120,18 +120,55 @@
...
+    border-radius: 4px 4px 0 0;
+    right: 0;
+    bottom: -40px;
+
+    .panel-header {
+      font-family: 'mastodon-font-display', sans-serif;
+      font-size: 14px;
+      line-height: 24px;
+      font-weight: 500;
+      color: lighten($ui-base-color, 26%);
```
nightpool: we use this lighten(ui-base, 26%) a ton of places in this code and in the other about page code, it should be pulled out into a scss variable.

nightpool (changes requested):
left a few code style comments. otherwise lgtm

app/javascript/styles/about.scss
nightpool: Not sure how much you want to fix here, I'm not asking for you to rewrite the whole about.scss file. but moving some of the smaller comments to it would be nice.

Gargron: Well this is reusing classes that were already in here. I'm not adding anything new.

app/javascript/styles/about.scss
Gargron: Maybe, but that's a refactoring task.

nightpool: Right, but this PR is adding new uses of it.

app/javascript/styles/about.scss
nightpool: okay

nightpool (approved):

Issue #2480, opened 2017-04-26
Redunant "Reject Media" checkbox when setting up a Suspend domain block


fpiesche:
When adding a new domain block and setting the severity to Suspend, the "Reject Media" checkbox remains visible, active and (by default) unticked. This is a bit counterintuitive and has been remedied with a note on the checkbox saying that it's irrelevant to Suspensions.

Better behaviour would be to either automatically tick and disable the tickbox when the new block is set to Suspend, or to hide it altogether as it is implied by the Suspend setting.
 
* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on [tagged release](https://github.com/tootsuite/mastodon/releases) 1.2.2 and not on `master`.

PR #4325, opened 2017-07-23
Allow domain blocks that only reject media without silencing or suspending

### LINKED ISSUE #3489: Add option for rejecting media without silencing
This was probably brought up at some point in #1847, but I'd like to open a separate issue for this specific feature. If an instance from a different country is hosting media that violates my country's laws, but they aren't violating my instance's conduct guidelines, I don't want to silence them, I just want to keep their media files off my server.

* * * *

duxovni:
Fixes #3489, also #2480

### DIFF
--- a/app/models/domain_block.rb
+++ b/app/models/domain_block.rb
@@ -12,11 +12,12 @@
 #
 
 class DomainBlock < ApplicationRecord
-  enum severity: [:silence, :suspend]
+  enum severity: [:silence, :suspend, :media_only]
 
   attr_accessor :retroactive
 
   validates :domain, presence: true, uniqueness: true
+  validates with DomainBlockValidator
 
   has_many :accounts, foreign_key: :domain, primary_key: :domain
   delegate :count, to: :accounts, prefix: true
--- a/app/models/domain_block.rb
+++ b/app/models/domain_block.rb
@@ -12,7 +12,7 @@
 #
 
 class DomainBlock < ApplicationRecord
-  enum severity: [:silence, :suspend, :media_only]
+  enum severity: [:silence, :suspend, :noop]
 
   attr_accessor :retroactive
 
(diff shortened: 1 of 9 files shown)


app/validators/domain_block_validator.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class DomainBlockValidator < ActiveModel::Validator
```
nightpool: seems like this could be much more easily handled by adding a reject_media? method that is defined as (media_only? || @reject_media?)

app/models/domain_block.rb
```diff
@@ -12,11 +12,12 @@
 #
 
 class DomainBlock < ApplicationRecord
-  enum severity: [:silence, :suspend]
+  enum severity: [:silence, :suspend, :media_only]
```
Gargron: I suggest naming it "transparent" or "noop" and having it do absolutely nothing, thus not having to worry about functionality overload with the reject_media boolean.

Gargron (changes requested):

app/models/domain_block.rb
sdukhovni: That's reasonable. Should the menu option still be called "Reject media only" in the frontend?

Gargron: No, I think it should be called accordingly (Transparent or No-op). Solves you a lot of trouble, no more need for the validation, even hiding the boolean with JS could be omitted

sdukhovni: So we're ok with admins creating domain blocks that are complete no-ops? I guess it doesn't hurt anything, but it seems a little silly...

sdukhovni: Completed the renaming, and removed the validation.

Gargron (approved):

PR #4328, opened 2017-07-24
Redirect user profile page to short URL when HTML format (regression from #3844)

ykzts:
### v1.4.7

```console
$ curl -H 'Accept: text/html' -sSI http://localhost:3000/users/user1 | grep Location
Location: http://localhost:3000/@user1
```

### master (`4b4ea1f`)

```console
$ curl -H 'Accept: text/html' -sSI http://localhost:3000/users/user1 | grep Location
```

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -26,7 +26,7 @@ Rails.application.routes.draw do
     confirmations:      'auth/confirmations',
   }
 
-  get '/users/:username', to: redirect('/@%{username}'), constraints: lambda { |req| req.format.nil? }
+  get '/users/:username', to: redirect('/@%{username}'), constraints: lambda { |req| req.format.nil? || req.format.html? }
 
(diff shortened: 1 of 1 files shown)


beatrix-bitrot:
introduced in 8c45cd0e3683b528b65f416681c8272d5650f32d but why?

nightpool:
the actual content of that commit conflicts with the commit message, which
was "do not redirect /users/:username route if format is empty"—that should
be *unless* format is empty, according to the logic introduced.

beatrix-bitrot:
so... missing !

?

ykzts:
[`ActionDispatch#format`](https://github.com/rails/rails/blob/v5.1.2/actionpack/lib/action_dispatch/http/mime_negotiation.rb#L50-L58) returns `Mime[:html]` when it can not be returned. It returns `nil` only if `request.format = nil` was explicitly specified.

Gargron (approved):

PR #4332, opened 2017-07-24
Center mastodon logo on mobile landing page

lynlynlynx:
Fix left shifted mastodon logo on mobile landing page.  
Settings for popping out the logo to the left on PC landing page caused this issue.

----

BTW, you set new logo's opacity as "0.8" intentionally? @Gargron  
I think this elephant's complexion looks bad (too transparent)..  
I wonder if it feels like this, because new logo on joinmastodon.org is not transparent :thinking:  

logo

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -504,6 +504,7 @@
       .brand {
         a {
           padding-left: 0;
+          padding-right: 0;
           color: $white;
 
           &:hover img {
@@ -513,7 +514,7 @@
 
         img {
           height: 32px;
-          margin-right: 10px;
+          margin-right: -20px;
           position: relative;
           top: 4px;
           left: -10px;


Gargron:
>Fix left shifted mastodon logo on mobile landing page

I have also submitted a fix for that in #4322: https://github.com/tootsuite/mastodon/pull/4322/files#diff-17a4be811dec69f1782f8ee4cc54ee3bR790

>BTW, you set new logo's opacity as "0.8" intentionally? 

I thought the hover effect looked better. This page doesn't have an illustration behind the logo so it's not the same as joinmastodon.org.

lynlynlynx:
Oh, sorry I don't notice your pull request.
I close this one.

### About the logo

I understood.
Hover effect certainly becomes easy to realize the logo is selected, but I think mastodon logo's color #3088d4 is more important. Every SNS has fixed color scheme these days. I propose exchange hover and unhover opacity.
@Gargron

PR #4333, opened 2017-07-24
disable to blocking admin account

masarakki:
I think admin should be able to see any public toots for management.
it is most easy and simple way to implement this.

### DIFF
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -18,6 +18,10 @@ class Block < ApplicationRecord
 
   validates :account_id, uniqueness: { scope: :target_account_id }
 
+  validate do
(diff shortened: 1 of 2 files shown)


nightpool:
Hi, after some discussion, I think we're not interested in this pull request.

There are lots of reasons why users may choose to block an account, and admins have many tools available to them (reports, database access, public web pages) to view both public and private statuses stored on their server.

If you need to get in contact with a user who has you blocked, you can always use email. Failing that, if a users isn't interested in communication from you, then there's very little chance for resolving the situation without proceeding to action.

Thank you for taking the time to put this together though! It prompted a very interesting conversation, and caused me to think about blocks in a new way then I would have otherwise.

PR #4341, opened 2017-07-24
Fix oddities with public profile CSS

Gargron:


app/views/accounts/show.html.haml
```diff
@@ -23,9 +23,9 @@
     .accounts-grid
       = render 'nothing_here'
   - else
-    .activity-stream
+    .activity-stream.with-header
       = render partial: 'stream_entries/status', collection: @statuses, as: :status
-
-  .pagination
-    - if @statuses.size == 20
+
```
ykzts: delete unnecessary indent.

ykzts (changes requested):

ykzts (approved):

PR #4344, opened 2017-07-24
fix(compose_from): Do not autofocus on mobile

sorin-davidoi:
I am afraid users will be confused by the animation if it is triggered as a response to navigating to the compose tab. It should be triggered as a response to **typing** (when they focus the textarea), not to switching tabs.


sorin-davidoi:
Implementation shamelessly stolen from @MightyPork, who implemented this for dev.glitch.social.

app/javascript/mastodon/features/compose/components/compose_form.js
```diff
@@ -179,7 +180,7 @@ export default class ComposeForm extends ImmutablePureComponent {
             onSuggestionsClearRequested={this.onSuggestionsClearRequested}
             onSuggestionSelected={this.onSuggestionSelected}
             onPaste={onPaste}
-            autoFocus={!showSearch}
+            autoFocus={!showSearch && !isMobile(window.innerWidth)}
```
Gargron: Side effects .-. Oh well

Gargron (approved):

PR #4355, opened 2017-07-25
Send short account URL to remote instance (regression from #3844)

### LINKED ISSUE #4347: [master] Inconsistent canonical user URL
I'm not sure how best to describe this, but see [this thread](https://glitch.social/@bea/1249257) for an idea of what's going on.

For whatever reason there's a difference in what URL constitutes a mention between my server (which runs 1.4.7) and glitch.social (which runs master), causing mentions from one or the other side to not work properly in web, opening up new tabs instead of in the right column. This has only happened to servers that have upgraded to master, and I can't seem to find a 100% consistent cause.

* * * *

ykzts:
Resolve #4347 

```diff
--- /tmp/before.json	2017-07-25 17:12:28.000000000 +0900
+++ /tmp/after.json	2017-07-25 17:12:19.000000000 +0900
@@ -8,7 +8,7 @@
     {
       "rel": "http://webfinger.net/rel/profile-page",
       "type": "text/html",
-      "href": "http://localhost:3000/users/ykzts"
+      "href": "http://localhost:3000/@ykzts"
     },
     {
       "rel": "http://schemas.google.com/g/2010#updates-from",
@@ -18,7 +18,7 @@
     {
       "rel": "self",
       "type": "application/activity+json",
-      "href": "http://localhost:3000/users/ykzts"
+      "href": "http://localhost:3000/users/ykzts.json"
     },
     {
       "rel": "salmon",
```

### DIFF
--- a/app/views/well_known/webfinger/show.json.rabl
+++ b/app/views/well_known/webfinger/show.json.rabl
@@ -8,9 +8,9 @@ end
 
 node(:links) do
   [
-    { rel: 'http://webfinger.net/rel/profile-page', type: 'text/html', href: account_url(@account) },
+    { rel: 'http://webfinger.net/rel/profile-page', type: 'text/html', href: short_account_url(@account) },
     { rel: 'http://schemas.google.com/g/2010#updates-from', type: 'application/atom+xml', href: account_url(@account, format: 'atom') },
-    { rel: 'self', type: 'application/activity+json', href: account_url(@account) },
+    { rel: 'self', type: 'application/activity+json', href: account_url(@account, format: 'json') },
     { rel: 'salmon', href: api_salmon_url(@account.id) },
     { rel: 'magic-public-key', href: "data:application/magic-public-key,#{@magic_key}" },
     { rel: 'http://ostatus.org/schema/1.0/subscribe', template: "#{authorize_follow_url}?acct={uri}" },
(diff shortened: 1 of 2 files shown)


app/views/well_known/webfinger/show.json.rabl
```diff
@@ -8,9 +8,9 @@ end
 
 node(:links) do
   [
-    { rel: 'http://webfinger.net/rel/profile-page', type: 'text/html', href: account_url(@account) },
+    { rel: 'http://webfinger.net/rel/profile-page', type: 'text/html', href: short_account_url(@account) },
     { rel: 'http://schemas.google.com/g/2010#updates-from', type: 'application/atom+xml', href: account_url(@account, format: 'atom') },
-    { rel: 'self', type: 'application/activity+json', href: account_url(@account) },
+    { rel: 'self', type: 'application/activity+json', href: account_url(@account, format: 'json') },
```
Gargron: The ActivityPub self link needs no format because the format is supposed to depend on request header.

Gargron (changes requested):

app/views/well_known/webfinger/show.json.rabl
ykzts: okay. i fixed 316c8b5

Gargron (approved):

PR #4357, opened 2017-07-25
fix #4356 : missing sw.js in Docker environment

### LINKED ISSUE #4356: The file sw.js is not found
I upgraded Mastodon to `v1.5.0rc1` (tagged release), but `sw.js` is not found.

https://mastodon.noraworld.jp/sw.js

I found `sw.js` on other instances, which version is `v1.5.0rc1`. I heard this is a specific problem with a deployment on Docker, which means it does not occur on non-Docker deployment. My instance is deployed with Docker.

* * * *

skoji:
This patch should fix #4356 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ public/system
 public/assets
 public/packs
 public/packs-test
-public/sw.js
 .env
 .env.production
 node_modules/
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -48,7 +48,7 @@ module.exports = merge(sharedConfig, {
       ServiceWorker: {
         entry: path.join(__dirname, '../../app/javascript/mastodon/service_worker/entry.js'),
         cacheName: 'mastodon',
-        output: '../sw.js',
+        output: '../assets/sw.js',
         publicPath: '/sw.js',
         minify: true,
       },
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,0 +1,1 @@
+assets/sw.js


skoji:
Yes, mounting directories that does not change to a docker container is not normal.

To avoid this, I've considered other fixes (e.g. generate sw.js in `public/assets` directory and add `public/sw.js` symbolic link that points to `public/assets/sw.js`), but this kind of fix (or workaround) “exposes”  a fix of docker-specific problem to non-docker environment.  I think it is not desirable.

skoji:
I've implemented symbolic-link version.
https://github.com/skoji/mastodon/commit/58c748c8c62cb48734022a8c677831257512e5ab
If this is preferable,  I'll replace this PR with this commit.

Gargron:
This might be fine..? I think if people kept the host-side volume mappings of the default (e.g. public/assets, public/system pointing to their in-docker counterparts) this change shouldn't break anything. Otherwise they'd have to move system and assets folders into a common folder. Want some input on this @Wonderfall

Wonderfall:
Mounting the entire public folder will break a few things from Mastodon if the person is using the image from Docker Hub directly or is not using a volume not corresponding to the repository of Mastodon.

I prefer the other version.

Gargron:
I like this. But I can't test if it works as expected since I don't use Docker in production. @Wonderfall: this okay? @ikuradon: would this likewise solve the problem in your opinion?

I might prefer it over #4357 for simplicity (less changes)

Wonderfall:
It's okay for me, it's simple and should do the trick.

nightpool:
I'm a fan of this one, although it might cause problems for non-docker users who already have an sw.js, if they don't clear their assets before cloning the repository?

ikuradon:
Looks good to me, but i don't know that symlinks work with every environments.

skoji:
@nightpool 

>  it might cause problems for non-docker users who already have an sw.js

Yes. But it is not a problem for released version users.
v1.5 Release candidate or `master` users may already have an `sw.js` and released version (<= v1.4.7) users do not have one.

skoji:
I'm running this version on [my single user instance](https://sandbox.skoji.jp) and it is working fine.

Gargron (approved):

PR #4358, opened 2017-07-25
fix #4356 : sw.js not found on Docker.

### LINKED ISSUE #4356: The file sw.js is not found

ikuradon:
Another solution from #4357 for #4356 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ public/system
 public/assets
 public/packs
 public/packs-test
-public/sw.js
 .env
 .env.production
 node_modules/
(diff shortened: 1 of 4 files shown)


Gargron:
@sorin-davidoi Will this work?

sorin-davidoi:
Not sure, will have to test when I get home.

sorin-davidoi:
This serves `sw.js` from the `assets` directory. It needs to be served from root.

ikuradon:
sw.js is need to serve from root path, but current Docker configuration Can't persist /public/ folder.
So, I wrote simple controller that only load /assets/sw.js and serve with /sw.js path.
How do you think?

Gargron:
Closed in favour of  #4358

PR #4360, opened 2017-07-25
Add index favourites on account_id and id

akihikodaki:


Gargron:
Is this necessary? Indices cost a lot of disk space (and write-perf). I am kinda worried by how many indices we already have:

a21bbc5e98

akihikodaki:
I do not think the regression in the write performance is problematic, but its disk usage may be for your instance. I'll consider whether it can replace `index_favourites_on_account_id_and_status_id`.

akihikodaki:
`index_favourites_on_account_id_and_status_id` is used by `ProcessInteractionService`, and the service is called frequently.

So the disk usage of the new index would matter. Can you show the size of `index_favourites_on_account_id_and_status_id`? That should have a similar size.

Gargron:
index_favourites_on_account_id_and_status_id | index | 151 MB
-- | -- | --

akihikodaki:
I think it is worth to cost 150MB.
I have also tested the query made by `api/v1/favourites` for accounts in Pawoo, and found some have thousands favourites and takes few seconds to query. It is worth to index.

beatrix-bitrot:
I can't put reactions from mobile?!

Anyway I agree with Aki on this one.

Gargron (approved):

PR #4363, opened 2017-07-25
Allow longer link text on landing page (reduce padding for Mobile)

clworld:
Fixes string length problem on #4362.
iphone5_logout
iphone6_logon

Appearance on iPhone5(logon) isn't fixed. but behave same as English.
iphone5_logon
iphone5_logon_en

I use flexbox to get same appearance with shorter strings. 
iphone5_logout_en

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -721,6 +721,17 @@
         text-align: center;
       }
 
+      .nav {
+        display: flex;
+        flex-flow: row wrap;
+        justify-content: space-around;
+
+      }
+
+      .links a {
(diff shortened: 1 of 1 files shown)


Gargron (approved):

app/javascript/styles/about.scss
```diff
@@ -721,6 +721,17 @@
         text-align: center;
       }
 
+      .nav {
+        display: flex;
+        flex-flow: row wrap;
+        justify-content: space-around;
+
```
clworld: I had mistake leaving empty line...

PR #4365, opened 2017-07-25
feat(status_action_bar): Web share button

sorin-davidoi:
This introduces an extra button for sharing toots using the operating systems' sharing implementation. The button only appears for browsers that support this and only for public toots. Uses the [Web Share API](https://wicg.github.io/web-share/) ([short video](https://youtu.be/lhUzYxCvWew)).

Browsers that support this: none. Next version of Chrome (61) will turn this on by default, until then you can test by enabling the Experimental Web Platform features in Chrome on Android - `chrome://flags/#enable-experimental-web-platform-features`.

image

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -12,6 +12,7 @@ const messages = defineMessages({
   mute: { id: 'account.mute', defaultMessage: 'Mute @{name}' },
   block: { id: 'account.block', defaultMessage: 'Block @{name}' },
   reply: { id: 'status.reply', defaultMessage: 'Reply' },
+  share: { id: 'status.share', defaultMessage: 'Share' },
   replyAll: { id: 'status.replyAll', defaultMessage: 'Reply to thread' },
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/status_action_bar.js
```diff
@@ -57,6 +58,13 @@ export default class StatusActionBar extends ImmutablePureComponent {
     this.props.onReply(this.props.status, this.context.router.history);
   }
 
+  handleShareClick = () => {
+    navigator.share({
+      text: this.props.status.get('search_index'),
```
Gargron: Not sure about the usage of search_index for this purpose. It's both content warning and proper text, yes.. But it's purpose is to be used in client-side search. We might make it do fancier indexing stuff later and then this usage will be out of date.

sorin-davidoi: Wanted to avoid the stripping of HTML tags inside a click handler. Will update it.

Gargron: Ah... Well, fair enough I suppose. Okay. Use it then.

Gargron (approved):

PR #4368, opened 2017-07-25
Added external-link icons to joinmastodon navbar buttons

MightyPork:
The "Instance list" button can deceive users into going to an external site unknowingly, and as joinmastodon has almost the same CSS, one is easily fooled.

This PR adds external-link icons there to make the distinction obvious.

screenshot_20170726_000133

screenshot_20170726_000143


beatrix-bitrot:
oh, this is nice!

Gargron:
I hate FA... Can you see how the icon's rectangle is like 1px above baseline? Why are FA icons not aligned...

Gargron (approved):

PR #4376, opened 2017-07-26
Update Japanese translations

mayaeh:
Update Japanese translations for remote profile, settings toggle button and web share button.

But this translation contain my free translation.
Please consider this translation.


app/javascript/mastodon/locales/ja.json
```diff
@@ -1,7 +1,7 @@
 {
   "account.block": "ブロック",
   "account.block_domain": "{domain}全体を非表示",
-  "account.disclaimer_full": "Information below may reflect the user's profile incompletely.",
+  "account.disclaimer_full": "以下の情報は不正確な可能性があります。",
```
unarist: * A literal translation of "incomplete" would be 不完全, but maybe 不正確 is better as inaccuracy or incorrectness of toot/follow/follower count.
* Japanese has プロフィール as translation of "profile", but it sounds like name/bio/avater/banner/etc. and not include counts or toots. 情報 (information) will cover all of it. (though I don't know does this message mention those profile or toots)

unarist (approved):
I've left some my thoughts, but looks good!

PR #4380, opened 2017-07-26
fix(web_push_notification): Do not hard reload tab

sorin-davidoi:
`client.navigate` caused a hard reload, we instead post a message instructing the client to navigate to the right URL.

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -49,6 +49,10 @@ const mapStateToProps = state => ({
 @connect(mapStateToProps)
 export default class UI extends React.PureComponent {
 
+  static contextTypes = {
+    router: PropTypes.object.isRequired,
+  }
+
   static propTypes = {
     dispatch: PropTypes.func.isRequired,
     children: PropTypes.node,
@@ -123,6 +127,20 @@ export default class UI extends React.PureComponent {
     this.setState({ draggingOver: false });
   }
 
+  handleServiceWorkerPostMessage = ({ data }) => {
+    if (data.type === 'navigate') {
+      const url = new URL(data.url);
+
+      if (url.host === location.host && url.pathname.startsWith('/web/')) {
+        this.context.router.history.push(url.pathname.slice('/web/'.length - 1));
+      } else {
+        location.href = data.url;
+      }
+    } else {
+      console.warn('Unknown message type:', data.type); // eslint-disable-line no-console
+    }
+  }
+
   componentWillMount () {
     window.addEventListener('resize', this.handleResize, { passive: true });
     document.addEventListener('dragenter', this.handleDragEnter, false);
@@ -130,6 +148,7 @@ export default class UI extends React.PureComponent {
     document.addEventListener('drop', this.handleDrop, false);
     document.addEventListener('dragleave', this.handleDragLeave, false);
     document.addEventListener('dragend', this.handleDragEnd, false);
+    navigator.serviceWorker.addEventListener('message', this.handleServiceWorkerPostMessage);
 
     this.props.dispatch(refreshHomeTimeline());
     this.props.dispatch(refreshNotifications());
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -129,13 +129,7 @@ export default class UI extends React.PureComponent {
 
   handleServiceWorkerPostMessage = ({ data }) => {
     if (data.type === 'navigate') {
-      const url = new URL(data.url);
-
-      if (url.host === location.host && url.pathname.startsWith('/web/')) {
-        this.context.router.history.push(url.pathname.slice('/web/'.length - 1));
-      } else {
-        location.href = data.url;
-      }
+      this.context.router.history.push(data.path);
     } else {
       console.warn('Unknown message type:', data.type); // eslint-disable-line no-console
     }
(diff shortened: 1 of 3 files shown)


app/javascript/mastodon/features/ui/index.js
```diff
@@ -123,13 +127,28 @@ export default class UI extends React.PureComponent {
     this.setState({ draggingOver: false });
   }
 
+  handleServiceWorkerPostMessage = ({ data }) => {
+    if (data.type === 'navigate') {
+      const url = new URL(data.url);
+
+      if (url.host === location.host && url.pathname.startsWith('/web/')) {
+        this.context.router.history.push(url.pathname.slice('/web/'.length - 1));
```
nightpool: should we ever postMessage a request to go somewhere that's not the local domain? This feels like a security hole slash unnecessary complexity.

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -57,12 +57,16 @@ const openUrl = url =>
         .filter(client => /\/web\//.test(client.url))
         .sort(client => client !== 'visible');
 
-      const visibleClient = clientList.find(client => client.visibilityState === 'visible');
-      const focusedClient = clientList.find(client => client.focused);
+      if (webClients.length !== 0) {
+        return webClients[0].focus().then(client => client.postMessage({ type: 'navigate', url }));
+      } else {
+        const visibleClient = clientList.find(client => client.visibilityState === 'visible');
+        const focusedClient = clientList.find(client => client.focused);
```
nightpool: not sure if this is the desired behavior. why fall back to navigating something that's not the webClient? if I have a profile page open, won't this hijack it? is that desired?

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -57,12 +57,16 @@ const openUrl = url =>
         .filter(client => /\/web\//.test(client.url))
         .sort(client => client !== 'visible');
```
nightpool: should we include focused as a discriminator here like we do below?

app/workers/web_push_notification_worker.rb
```diff
@@ -9,7 +9,7 @@ def perform(session_activation_id, notification_id)
     session_activation = SessionActivation.find(session_activation_id)
     notification = Notification.find(notification_id)
 
-    return if session_activation.nil? || notification.nil?
+    return if session_activation.nil? || session_activation.web_push_subscription.nil? || notification.nil?
```
nightpool: why not session_activation&.web_push_subscription.nil?

app/javascript/mastodon/features/ui/index.js
```diff
@@ -123,13 +127,28 @@ export default class UI extends React.PureComponent {
...
+  }
+
   componentWillMount () {
     window.addEventListener('resize', this.handleResize, { passive: true });
     document.addEventListener('dragenter', this.handleDragEnter, false);
     document.addEventListener('dragover', this.handleDragOver, false);
     document.addEventListener('drop', this.handleDrop, false);
     document.addEventListener('dragleave', this.handleDragLeave, false);
     document.addEventListener('dragend', this.handleDragEnd, false);
+    navigator.serviceWorker.addEventListener('message', this.handleServiceWorkerPostMessage);
```
nightpool: should this guard against serviceWorkers not being supported?

nightpool (changes requested):
some subset of the above probably needs to be thought about

app/javascript/mastodon/features/ui/index.js
sorin-davidoi: Don't think we do, but I can get rid of the else branch just in case.

app/javascript/mastodon/service_worker/web_push_notifications.js
sorin-davidoi: If you add the website to the home-screen and go to the settings, the notification should open the "app", not Chrome with a new tab.

app/javascript/mastodon/service_worker/web_push_notifications.js
sorin-davidoi: Sure, we can have that as the first criteria.

app/workers/web_push_notification_worker.rb
sorin-davidoi: Cause I lack Ruby skills :)

app/javascript/mastodon/features/ui/index.js
sorin-davidoi: Right, forgot about that :+1:. Will move this to `web_push_subscription`.

app/workers/web_push_notification_worker.rb
Gargron: hey the "find" above will throw an exception so these guards are unnecessary. the only necessary guard is whether web_push_subscription is nil

app/workers/web_push_notification_worker.rb
```diff
@@ -9,7 +9,7 @@ def perform(session_activation_id, notification_id)
     session_activation = SessionActivation.find(session_activation_id)
     notification = Notification.find(notification_id)
 
-    return if session_activation.nil? || session_activation.web_push_subscription.nil? || notification.nil?
+    return if session_activation.nil? || notification.nil?
```
beatrix-bitrot: if i understand garg's feedback correctly this should just be

`return if session_activation.web_push_subscription.nil?`

beatrix-bitrot (dismissed):
just one little thing it seems

app/workers/web_push_notification_worker.rb
sorin-davidoi: Yes, but I took it out since it is out of the scope of the PR and gargron addresses it [here](https://github.com/tootsuite/mastodon/pull/4384/files#diff-9cfb6ca773d14b42531acfa66c3278c1R12).

nightpool (approved):
LGTM!

app/javascript/mastodon/features/ui/index.js
```diff
@@ -131,6 +143,10 @@ export default class UI extends React.PureComponent {
     document.addEventListener('dragleave', this.handleDragLeave, false);
     document.addEventListener('dragend', this.handleDragEnd, false);
 
+    if ('serviceWorker' in  navigator) {
```
nightpool: (nit) double space

PR #4381, opened 2017-07-26
Fix infinite scroll fluidity

Gargron:


nightpool (approved):
LGTM code-wise. hard for me to test for 'fluidity' but this change is probably as good as any

PR #4382, opened 2017-07-26
Fix #4026 - Accept backup codes for disabling 2FA

### LINKED ISSUE #4026: Ability to deactivate 2FA with recovery code
Use case: You have lost your phone or lose the account on your authenticator app and can no longer get your 2FA TOTP codes. Login using one of the backup recovery codes. When going to the Two-Factor section, you wish to disable 2FA and re-enable it in the authenticator app or a different phone, but it asks for the 2FA code to disable.

A security consideration is that anyone who has access to one's logged in browser can regenerated backup recovery codes and then disable 2FA. However, for the legitimate loss of 2FA device (or account in app), the user is forced to use recovery codes forever (having to regenerate them before using them all up).

Just as in login, a user should be able to use a recovery code to disable/reset 2FA. 

* * * *

Gargron:


nightpool (approved):

nightpool:
there's some small duplication here between this and SessionsController now, but i can also see the argument for not encapsulating this behavior into User.

PR #4384, opened 2017-07-26
Multiple Sidekiq retry adjustments

Gargron:
- Disable dead job queue for all jobs, because the dead job queue is never
  used by anyone except to clear it
- Make Salmon delivery worker a bit better at retrying, at first make it
  behave like the PuSH delivery worker, but after 3 tries go back to
  standard Sidekiq scheduling
- If a subscription job fails all the way through, unsubscribe from
  that user to prevent furter subscription attempts
- Fix WebPushNotificationWorker guarding against impossible conditions,
  make it guard against a real failure condition

### DIFF
--- a/app/workers/notification_worker.rb
+++ b/app/workers/notification_worker.rb
@@ -3,7 +3,15 @@
 class NotificationWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'push', retry: 5
+  sidekiq_options queue: 'push', retry: 10, dead: false
+
+  sidekiq_retry_in do |count|
+    if count < 3
+      5 * (count + 1)
+    else
+      (count ** 4) + 15 + (rand(30) * (count + 1))
+    end
+  end
 
   def perform(xml, source_account_id, target_account_id)
     SendInteractionService.new.call(xml, Account.find(source_account_id), Account.find(target_account_id))
--- a/app/workers/pubsubhubbub/subscribe_worker.rb
+++ b/app/workers/pubsubhubbub/subscribe_worker.rb
@@ -3,7 +3,7 @@
 class Pubsubhubbub::SubscribeWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'push', retry: 10, unique: :until_executed
+  sidekiq_options queue: 'push', retry: 10, unique: :until_executed, dead: false
 
   sidekiq_retry_in do |count|
     case count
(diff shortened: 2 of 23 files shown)


app/workers/web_push_notification_worker.rb
```diff
@@ -3,13 +3,13 @@
...
 
-  sidekiq_options backtrace: true
+  sidekiq_options backtrace: true, dead: false
 
   def perform(session_activation_id, notification_id)
     session_activation = SessionActivation.find(session_activation_id)
     notification = Notification.find(notification_id)
 
-    return if session_activation.nil? || notification.nil?
+    return if session_activation.web_push_subscription.nil?
```
sorin-davidoi: Can't the notification be `nil` though? Favourite -> job gets scheduled -> unfavourite -> job is executed.

Gargron: `Notification.find` will raise `ActiveRecord::RecordNotFound`

You may check something like `notification.activity.nil?` instead because that could be nil accidentally.

nightpool:
what's the benefits to getting rid of the dead job queue? I find it useful to understand what types of jobs are failing and why, even if they've passed out of retries. Maybe this is an argument that I should have better error analytics though....

subscribeworker should definitely not go into dead jobs though.

app/workers/pubsubhubbub/subscribe_worker.rb
```diff
@@ -18,6 +18,12 @@ class Pubsubhubbub::SubscribeWorker
     end
   end
 
+  sidekiq_retries_exhausted do |msg, e|
+    account = Account.find(msg['args'].first)
+    logger.error "PuSH subscription attempts for #{account.acct} exhausted. Unsubscribing"
+    ::UnsubscribeService.new.call(account)
```
nightpool: 🙌

app/workers/pubsubhubbub/subscribe_worker.rb
```diff
@@ -18,6 +18,12 @@ class Pubsubhubbub::SubscribeWorker
     end
   end
 
+  sidekiq_retries_exhausted do |msg, e|
```
ykzts: `|msg, e|` -> `|msg, _e|` (for RuboCop)

app/workers/notification_worker.rb
```diff
@@ -3,7 +3,15 @@
...
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'push', retry: 5
+  sidekiq_options queue: 'push', retry: 10, dead: false
+
+  sidekiq_retry_in do |count|
+    if count < 3
+      5 * (count + 1)
+    else
+      (count ** 4) + 15 + (rand(30) * (count + 1))
```
ykzts: `count ** 4` -> `count**4` (for RuboCop)

ref https://github.com/bbatsov/rubocop/issues/2972

ykzts: I think it is better to have around space.

Gargron:
So today I am a bit concerned about this PR because I (again) stuck too many different changes into a single PR. From the reactions it seems like there's some good stuff in there. I think I'll split it up into separate PRs.

PR #4387, opened 2017-07-26
Fix multipoint shortcode bug

### LINKED ISSUE #4386: Multi-point emoji shortcodes are broken again
This is the same issue we earlier had in front-end, except now it's in backend.

Only the first character of the emoji is inserted after parsing a shortcode.

Try: `:couple_ww:` `:flag_cz:`, inserts just one woman and the C symbol

Tested on upstream master, just pulled.

MightyPork:
Fixes #4386. I don't know ruby, but have googled up a fix and it worked, so :man_shrugging: 

edit: Removed snark.
but this really should've been caught by rspec, at least thats what I thought it's for

### DIFF
--- a/app/lib/emoji.rb
+++ b/app/lib/emoji.rb
@@ -32,7 +32,7 @@ class Emoji
 
   def codepoint_to_unicode(codepoint)
     if codepoint.include?('-')
-      codepoint.split('-').map(&:hex).pack('U')
+      codepoint.split('-').map(&:hex).pack('U*')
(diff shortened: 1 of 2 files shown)


Gargron:
>but this really should've been caught by rspec, at least thats what I thought it's for

Feel free to add a test case in this PR.

MightyPork:
I've never done this before, but this could be it

Gargron (approved):

PR #4388, opened 2017-07-26
feat(web_push_notifications): Group notifications

sorin-davidoi:
image

Tradeoffs:
 - Due to the complexity of dealing with locales in the Service Worker, we are sending the message in the payload (`%{count} notifications`) and replacing `%{count}` when the grouping occurs. This is suboptimal, to say the least, but a proper solution is much harder to implement.
 - No actions for the grouped notifications.
 - No ability to customize `MAX_NOTIFICATIONS` - tablet users might want to increase it since they have more space in the notification drawer.
 - If you receive `MAX_NOTIFICATIONS + 1` notifications, this will show the first `MAX_NOTIFICATIONS` individually and group when the last one arrives.
 - Notification body is a concatenation of the original notification titles (most recent on top) - should it be the body instead of the title? Things to consider: line breaks in the message, not enough space in the body...

Related to #4200.

### DIFF
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -1,3 +1,45 @@
+const MAX_NOTIFICATIONS = 5;
+const GROUP_TAG = 'tag';
+
+// Avoid loading intl-messageformat and dealing with locales in the ServiceWorker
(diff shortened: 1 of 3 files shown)


beatrix-bitrot (approved):
LGTM (mostly i want to test this on glitch.social)

ykzts (approved):
I think there is no problem with the code.

nightpool:
:shipit:

PR #4396, opened 2017-07-27
fix(push_subscription): Use star icon for favourite action

sorin-davidoi:
image
image

Would like to change to boost icon to something that is not circle-shaped (Android makes the icons gray, so the boost icon becomes a gray circle - not recognizable), but couldn't find an emoji for it.

### DIFF
--- a/app/models/web/push_subscription.rb
+++ b/app/models/web/push_subscription.rb
@@ -117,7 +117,7 @@ class Web::PushSubscription < ApplicationRecord
       when :mention then [
         {
           title: translate('push_notifications.mention.action_favourite'),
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
So I just noticed that these icons are just arbitrary images. We could export the same icons we use in the web UI as images! Instead of using graphical emoji. Unless there is a good reason to use those?

sorin-davidoi:
No good reason, I just went for the simplest solution at the time.

Gargron:
Okay, I will do that and submit a PR.

PR #4398, opened 2017-07-27
Update intersection-observer to version 0.4.0

ykzts:
ref https://github.com/WICG/IntersectionObserver/pull/224

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -107,17 +107,12 @@ export default class Status extends ImmutablePureComponent {
       this.height = getRectFromEntry(entry).height;
     }
 
-    // Edge 15 doesn't support isIntersecting, but we can infer it
-    // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/
-    // https://github.com/WICG/IntersectionObserver/issues/211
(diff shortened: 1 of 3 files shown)


nolanlawson:
FWIW we don't really need https://github.com/WICG/IntersectionObserver/pull/224 because we already work around the Edge issue, but upgrading won't hurt. :)

Gargron (approved):

nightpool:
We should probably just revert this given that we want to get 1.5 out tomorrow

ykzts:
@nightpool Chromium 57 on Mageia 5.1 will not download [`extra_polyfills.js`](https://github.com/tootsuite/mastodon/blob/v1.5.0rc3/app/javascript/mastodon/extra_polyfills.js).

virtualbox_mageia_01_08_2017_00_12_01

Probably other cause (The problem itself is reproduced).

Gargron:
Could it be the isIntersecting vs entry.isIntersecting change? Can you reproduce before/after with PR?

ykzts:
@Gargron Oh, nice! The problem seems to be there. Please wait a moment. I will submit a fix Pull Request.

ykzts:
@Gargron OK. I created PR (#4469).

PR #4404, opened 2017-07-27
fix french translation with non-breaking spaces

smdgjmigop:
Hi,
I did this, I no nothing in web stuff, I hope non breakable spaces dont break any web thing

### DIFF
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -5,7 +5,7 @@ fr:
     about_this: À propos
     closed_registrations: Les inscriptions sont actuellement fermées sur cette instance. Cependant, vous pouvez trouver une autre instance sur laquelle vous créer un compte et à partir de laquelle vous pourrez accéder au même réseau.
     contact: Contact
-    description_headline: Qu’est-ce que %{domain} ?
+    description_headline: Qu’est-ce que %{domain} ?
     domain_count_after: autres instances
     domain_count_before: Connectés à
     features:
@@ -27,12 +27,12 @@ fr:
     status_count_before: Ayant publié
     user_count_after: utilisateur⋅ice⋅s
     user_count_before: Abrite
-    what_is_mastodon: Qu’est-ce que Mastodon ?
+    what_is_mastodon: Qu’est-ce que Mastodon ?
   accounts:
     follow: Suivre
     followers: Abonné⋅es
     following: Abonnements
-    nothing_here: Rien à voir ici !
+    nothing_here: Rien à voir ici !
     people_followed_by: Personnes suivies par %{name}
     people_who_follow: Personnes qui suivent %{name}
     posts: Statuts
@@ -41,7 +41,7 @@ fr:
     unfollow: Ne plus suivre
   admin:
     accounts:
-      are_you_sure: Êtes-vous certain⋅e ?
+      are_you_sure: Êtes-vous certain⋅e ?
       confirm: Confirmer
       confirmed: Confirmé
       disable_two_factor_authentication: Désactiver l’authentification à deux facteurs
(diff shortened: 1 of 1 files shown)


ykzts:
cc @eramdam @Sylvhem @aaribaud (recent editors for fr.yml)

nightpool:
I think this is correct, although I'm wondering if there's a better way to do it technically. &nbsp; isn't exactly right, because these translations might not all be rendered into html, but maybe there's a yml way to represent these?

should the .json files also be updated?

aaribaud:
IIUC, this replaces narrow no-break spaces (Unicode e2 80 af) with "full-size" no-break spaces (c2 a0).

From a typographical / composition usage standpoint, this change should not be done: so-called "double punctuation" (question mark, colon, semicolon, exclamation mark) are supposed to be preceded by an "espace fine" (a narrow no-break space) in French typography, and that's what the current codebase does.

@smdgjmigop : from your commit/PR description, I _think_ you did witness cases where the space was actually broken and the punctuation rejected to the next line, and this is what prompted you to change what you saw as breaking spaces into (full-size) no-break spaces. Am I correct? If so, please describe how you witnessed it as precisely as possible (which OS, which browser or app, with versions ideally).

I'm asking this because the root cause could be in Mastodon code _or_ in a browser's (or app's) rendering engine failing to properly handle the narrow no-break spaces; being able to reproduce the issue may help directing it to the right project.

smdgjmigop:
@aaribaud yes, you're right.  And my emacs white-space mode fooled me by showing the same glyphs for regular spaces and narrow non breaking spaces. Using hexl-mode I actually see the narrow non-break spaces (btw, thank you, I didn't knew those). I am glad my pull request wasn't blindly accepted :-)

I witnessed this on Arch Linux with firefox and chromium, and on ubuntu (17.04) with firefox.

aaribaud:
@smdgjmigop Ok, I could act as if I were a Unicode guru, but the truth is, I just hexdumped your diff, noted the differing whitespace encodings, googled the hex byte sequences and discovered the Unicode narrow no-break space only 30 minutes before you. :innocent:

I can try and reproduce on Chromium or Firefox on my Xubuntu. What's the easiest/fastest way to cause text with a narrow no-break whitespace appear in my browser?

smdgjmigop:
@aaribaud You can check my web designer skills : 
http://simon.archipoff.emi.u-bordeaux1.fr/narrow_space.html
:-p

Sylvhem:
> so-called "double punctuation" (question mark, colon, semicolon, exclamation mark) are supposed to be preceded by an "espace fine" (a narrow no-break space) in French typography, and that's what the current codebase does

It does because I previously inserted narrow no-break space everywhere in this file ^^ (see #4107). But apparently, I missed some of them. Thanks @smdgjmigop for noticing!

smdgjmigop:
@Sylvhem you're welcome, I made a mess in your work :-p
how do I retract my pull request ? can I do that ?
I'll do the bug report for the rendering problem. Those non-break spaces aren't in the final html document displayed by the web browser.

smdgjmigop:
I made a bug report, see : https://github.com/tootsuite/mastodon/issues/4434 @aaribaud @Sylvhem @nightpool

aaribaud:
@smdgjmigop :smile: That's enough at least to conclude that FF respects the no-break part of the narrow space but, incidentally, messes up computing the minimum required width: the colon in your page is never rejected to 2nd line (which is good), but is overflown and hidden on the right when reducing the FF width to a minimum. I'll run other tests.

But in fact that's beside the topic here, i.e. the PR itself, and should be an issue... In the right project, to be determined. I suggest we discuss it under https://github.com/smdgjmigop/mastodon/commit/a1748b25f340dcf8847eae95e7e39e08aa335242 (possibly placing an ad hoc branch on the commit and removing it from your master).

@ykzts or @nightpool, as per https://github.com/tootsuite/mastodon/pull/4404#issuecomment-318709358, @smdgjmigop agrees that this PR should be rejected.

nightpool:
consensus is that this pull request should be closed, but can someone take the YML problem and work on it? Maybe a bug that we need to report to https://github.com/svenfuchs/i18n or https://github.com/svenfuchs/rails-i18n?

PR #4369, opened 2017-07-25
Improve accessibility

sorin-davidoi:
- [x] Use full text for tab links (Getting started had no text)
- [x] Use semantic markup (`nav`)
- [x] Label for settings toggle button

PR #4405, opened 2017-07-27
Improve accessibility (part 3)

sorin-davidoi:
Follow-up to #4369 and #4377. Mostly fixes errors discovered by WAVE Web Accessibility Tool.

cc: @yiskah

### DIFF
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -135,7 +135,7 @@ export default class ColumnHeader extends React.PureComponent {
 
     return (
       <div className={wrapperClassName}>
-        <h1 tabIndex={focusable && '0'} className={buttonClassName} aria-label={title} onClick={this.handleTitleClick}>
+        <h1 tabIndex={focusable && '0'} role='button' className={buttonClassName} aria-label={title} onClick={this.handleTitleClick}>
           <i className={`fa fa-fw fa-${icon} column-header__icon`} />
           {title}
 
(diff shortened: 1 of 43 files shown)


sorin-davidoi:
I think the warnings are wrong? `input` is inside `label` (which is what WAVE Web Accessibility Tool suggests).

unarist:
Looks like it was fixed last month? https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/v6.0.0

app/javascript/mastodon/components/column_header.js
```diff
@@ -133,15 +135,15 @@ export default class ColumnHeader extends React.PureComponent {
 
     return (
       <div className={wrapperClassName}>
-        <div role='heading' tabIndex={focusable && '0'} className={buttonClassName} aria-label={title} onClick={this.handleTitleClick}>
+        <h1 role='heading' tabIndex={focusable && '0'} className={buttonClassName} aria-label={title} onClick={this.handleTitleClick}>
```
unarist: It's heading element now, so `role='heading'` can be removable.

Gargron:
>Form controls using a label to identify them must be programmatically associated with the control using htmlFor

Hwhat?

sorin-davidoi:
Should I just disable the errors?

I've added the `inert` attribute to the siblings of `modal-root` when it is opened. I'm deliberately not including [the polyfill](https://github.com/WICG/inert#notes-on-the-polyfill) because I didn't test it's performance (relies on mutation observers).

Gargron (approved):

yiskah:
<3 you're the best sorin

PR #4410, opened 2017-07-28
Update webpack and related dependencies (without extract-text-webpack-plugin)

ykzts:
only minor and patch releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,9 +112,9 @@
     "tiny-queue": "^0.2.1",
     "uuid": "^3.1.0",
     "uws": "^8.14.0",
-    "webpack": "^3.0.0",
-    "webpack-bundle-analyzer": "^2.8.2",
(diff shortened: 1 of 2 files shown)


nightpool (dismissed):

Gargron:
To make sure: this does not re-introduce the live reloading bug, does it?

nightpool:
withdrawing my review b/c I didn't know we had downgraded webpack previously https://github.com/tootsuite/mastodon/pull/4250

ykzts:
Yes, that bug is caused by extract-text-webpack-plugin v3.0.0.

nightpool (commented):
.

nightpool:
(someone else feel free to review, I just don't feel comfortable doing so)

Gargron (approved):

PR #4411, opened 2017-07-28
Remove hash from chunk filename when dev env

ykzts:
Change the processing of `output.filename` and `output.chunkFilename` the same.


Gargron:
What does this accomplish?

ykzts:
Many files will not be created in the public / packs directory during development

Gargron (approved):

PR #4413, opened 2017-07-28
Fix broken PropTypes (regression from #4408)

ykzts:

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -41,8 +41,8 @@ export default class Status extends ImmutablePureComponent {
     autoPlayGif: PropTypes.bool,
     muted: PropTypes.bool,
     intersectionObserverWrapper: PropTypes.object,
-    index: PropTypes.oneOf(PropTypes.string, PropTypes.number),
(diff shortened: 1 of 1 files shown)


ykzts:
cc @sorin-davidoi

app/javascript/mastodon/components/status.js
```diff
@@ -41,8 +41,8 @@ export default class Status extends ImmutablePureComponent {
     autoPlayGif: PropTypes.bool,
     muted: PropTypes.bool,
     intersectionObserverWrapper: PropTypes.object,
-    index: PropTypes.oneOf(PropTypes.string, PropTypes.number),
-    listLength: PropTypes.oneOf(PropTypes.string, PropTypes.number),
+    index: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
ykzts: ref https://facebook.github.io/react/docs/typechecking-with-proptypes.html

unarist:
Actually, should it be "string or number" instead of "number"?

sorin-davidoi:
We are not doing any computations with it, and it gets converted to a string anyway (HTML attribute). Wanted to avoid manually casting it like `<Status index={String(i)} />`.

unarist (approved):

PR #4420, opened 2017-07-28
Upgrade to React 16

sorin-davidoi:
Works fine for the most part, some issues that need to be addressed (see [release notes](https://github.com/facebook/react/issues/10294)):
 - `react-addons-perf` does not work anymore
 - `react-router-scroll` imports `PropTypes` from the `react` package - @ykzts made a [PR](https://github.com/ytase/react-router-scroll/pull/1/files) to fix this - meanwhile we use the fork
 - Sometimes `columnId` is `undefined`, leading to crashed in `react-router-scroll` - gave it a default value
 - Detailed statuses, accounts fail to mount - again something with `react-router-scroll`
-  Tests fail
 - [The bundles] look weird... - looks like `react-dom` is loaded twice (I guess 15 and 16)

Temporarily modified the `postinstall` script to manually build `react-router-scroll` - might not work on Window.


sorin-davidoi:
I think everything works after bumping the React version used by `react-router-scroll`.

PR #4424, opened 2017-07-28
Fix current session not being displayed in sessions list

Gargron:


app/views/auth/registrations/_sessions.html.haml
```diff
@@ -19,10 +19,10 @@
         %td
           %samp= session.ip
         %td
-          - if request.session['auth_id'] == session.session_id
+          - if request.cookies.signed['_session_id'] == session.session_id
```
unarist: Cannot we use `ApplicationController#current_session` or something for this, instead accessing cookie directly?

app/views/auth/registrations/_sessions.html.haml
```diff
@@ -19,10 +19,10 @@
         %td
           %samp= session.ip
         %td
-          - if request.session['auth_id'] == session.session_id
+          - if current_session.id == session.session_id
```
unarist: not working.

for example, `current_session.id` is 34, but `session.session_id` is `816a3488dfacc7912615de4977ccdd2a`.

unarist (changes requested):

unarist (approved):

PR #4427, opened 2017-07-28
Count all URLs in text as 23 characters flat, do not count domain part of usernames

Gargron:
TODO:

- [x] Backend validation adjusted
- [x] Tests
- [x] Web UI counter adjusted

### DIFF
--- a/app/javascript/mastodon/features/compose/util/counter.js
+++ b/app/javascript/mastodon/features/compose/util/counter.js
@@ -1,0 +1,7 @@
+const urlPlaceholder = 'xxxxxxxxxxxxxxxxxxxxxxx';
+
+export function countableText(inputText) {
+  return inputText
+    .replace(/https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g, urlPlaceholder)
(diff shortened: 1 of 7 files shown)


app/javascript/mastodon/features/compose/util/counter.js
```diff
@@ -0,0 +1,7 @@
+const urlPlaceholder = 'xxxxxxxxxxxxxxxxxxxxxxx';
+
+export function countableText(inputText) {
+  return inputText
+    .replace(/https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g, urlPlaceholder)
```
nightpool: why 256 instead of {2,}?

Gargron: I did copy-paste the regex. If you have a more authoritative regex, be my guest

krainboltgreene (approved):

app/javascript/mastodon/features/compose/util/counter.js
ProgVal: The 256 part is not a big issue, as all domain names are [limited to 255 characters in practice](https://stackoverflow.com/a/28918017/539465).
However, the regexp for the TLD is `[a-z]{2,6}`, which wrong because TLDs are not limited to 6 characters. eg. `.network` (used by mastodon.network)

kensanata: Fixed by #4463.

PR #4429, opened 2017-07-28
fix(tabs_bar): Allow animation to end before navigating

sorin-davidoi:
Another hacky solution, similar to #4320.

cc: @unarist @nolanlawson

### DIFF
--- a/app/javascript/mastodon/features/ui/components/tabs_bar.js
+++ b/app/javascript/mastodon/features/ui/components/tabs_bar.js
@@ -43,8 +43,9 @@ export default class TabsBar extends React.Component {
     e.persist();
 
     requestAnimationFrame(() => {
-      const currentTab = this.node.querySelector('.tabs-bar__link.active');
-      const nextTab = e.nativeEvent.path.find(node => node.classList.contains('tabs-bar__link'));
+      const tabs = Array(...this.node.querySelectorAll('.tabs-bar__link'));
+      const currentTab = tabs.find(tab => tab.classList.contains('active'));
+      const nextTab = tabs.find(tab => tab.contains(e.target));
       const { props: { to } } = links[Array(...this.node.childNodes).indexOf(nextTab)];
 
       currentTab.classList.remove('active');
--- a/app/javascript/mastodon/features/ui/components/tabs_bar.js
+++ b/app/javascript/mastodon/features/ui/components/tabs_bar.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import NavLink from 'react-router-dom/NavLink';
 import { FormattedMessage, injectIntl } from 'react-intl';
+import { debounce } from 'lodash';
 
 export const links = [
   <NavLink className='tabs-bar__link primary' to='/statuses/new' data-preview-title-id='tabs_bar.compose' data-preview-icon='pencil' ><i className='fa fa-fw fa-pencil' /><FormattedMessage id='tabs_bar.compose' defaultMessage='Compose' /></NavLink>,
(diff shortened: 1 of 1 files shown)


nolanlawson:
This solution looks great to me! Is there really not a "React-ish" solution though, like what Twitter describes [here](https://medium.com/@paularmstrong/twitter-lite-and-high-performance-react-progressive-web-apps-at-scale-d28a00e780a3) under "Defer Rendering When Mounting & Unmounting Many Components"? I kinda figured we could just use their same approach of deferring the component.

nolanlawson (approved):

app/javascript/mastodon/features/ui/components/tabs_bar.js
```diff
@@ -25,16 +26,46 @@ export function getLink (index) {
...
+      const currentTab = this.node.querySelector('.tabs-bar__link.active');
+      const nextTab = e.nativeEvent.path.find(node => node.classList.contains('tabs-bar__link'));
+      const { props: { to } } = links[Array(...this.node.childNodes).indexOf(nextTab)];
+
+      currentTab.classList.remove('active');
+
+      const listener = debounce(() => {
+        nextTab.removeEventListener('transitionend', listener);
+        this.context.router.history.push(to);
+      }, 50);
```
nolanlawson: What's with the debounce? `transitionend` should only be called once, exactly when the transition has ended.

app/javascript/mastodon/features/ui/components/tabs_bar.js
```diff
@@ -25,16 +26,46 @@ export function getLink (index) {
...
+    this.node = ref;
+  }
+
+  handleClick = (e) => {
+    e.preventDefault();
+    e.persist();
+
+    requestAnimationFrame(() => {
+      const currentTab = this.node.querySelector('.tabs-bar__link.active');
+      const nextTab = e.nativeEvent.path.find(node => node.classList.contains('tabs-bar__link'));
```
unarist: `path` property is not exist in Firefox and Safari.

unarist (changes requested):

app/javascript/mastodon/features/ui/components/tabs_bar.js
sorin-davidoi: It has `transition: all`, and that class changes like 4 properties, all firing events.

app/javascript/mastodon/features/ui/components/tabs_bar.js
sorin-davidoi: Today I learned :+1:

unarist (approved):

sorin-davidoi:
@nolan Will give it a shot but last time I checked the biggest bottleneck was not the rendering but the unmounting of the all the component on the page.

unarist:
> Another problem is the animation. We would have to delay the mounting of the component not by animation frame, but by the duration of the transition (200ms).

I've missed this point, and yes, this change looks more laggy on the PC...

sorin-davidoi:
We could wrap it inside isUserTouching, like the actions modal.

unarist:
Although I don't think that feeling is not depending on whether it's PC or mobile device, but device performance or something.

sorin-davidoi:
Yes, but I would say that whether the device is touch or not is strongly correlated to the performance of the device. Don't know what to say, let's leave it as it is for now and hear more feedback?

unarist:
I found a odd behavior.

1. Open home column
2. Touch home tab, and just after, touch notification tab
3. It will show notification column in a moment, then back to home column

<details>
<summary>demo gif (w/ event timing log)</summary>

test
</details>

PR #4430, opened 2017-07-28
Improve modal styles for mobile

unarist:
* Add max-width / max-height to make cancelling by touch backdrop easier
* report: Make p tag in toots inline for `text-overflow: ellipsis`, and reduce margins
* boost: Fix action-bar layout, and remove `overflow-x: scroll`
* actions: Add `:active` style for touching feedback

Update: Added more fixes for landscape view.

* report: Move report comment into action-bar (screenshot without this change is [available here](https://user-images.githubusercontent.com/705555/28729054-031582de-7406-11e7-98eb-63e65bd4433e.png))
**Currently, this change will also applied to PC layout.**
* actions: Make scrollable if needed

cc @sorin-davidoi 

|Before|After|
|-------|------|
|image|image|
|image|image|
|image|image|

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -3435,19 +3435,6 @@ button.icon-button.active i.fa-retweet {
   }
 }
 
-.actions-modal {
-  .status {
-    background: $white;
-    border-bottom-color: $ui-secondary-color;
-    padding-top: 10px;
-    padding-bottom: 10px;
-  }
-}
-
-.actions-modal__actions {
-  overflow-y: auto;
-}
-
 .boost-modal__container {
   overflow-y: auto;
   padding: 10px;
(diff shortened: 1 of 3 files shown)


sorin-davidoi:
Sometimes the report button is not visible (you have to scroll to see it)?

image

unarist:
Ah, that's right. Also I cannot see some items on actions modal...

image

unarist:
@sorin-davidoi How about this?

image

image

and for actions-modal...well, two scroll bars.

image

sorin-davidoi:
I only get the scrollbar if I take out `flex-shrink: 0` from `.actions-modal ul`:
image

unarist:
I've updated to toggle scrollbar around 430px (I'm not sure this value, but iPhone 5s portrait doesn't need scrollbar I think.)

sorin-davidoi (approved):

app/javascript/styles/components.scss
```diff
@@ -3507,17 +3504,22 @@ button.icon-button.active i.fa-retweet {
...
-  max-height: 80vh;
-  max-width: 80vw;
-
   ul {
-    overflow-y: auto;
     flex-shrink: 0;
 
+    @media screen and (max-height: 430px) {
+      flex-shrink: 1;
+      overflow-y: auto;
```
unarist: I thought this style can be set always (outside of media query), but it doesn't show scrollbar on portrait -> landscape changing with that way. So I will set this only if needed.

unarist:
@Gargron This improves usability, but also fixed some elements being inaccessible in landscape view. However I have changed some layout, so I'd like to hear your impressions.

Gargron:
I do not like the layout change for the report modal, but otherwise this is fine. I'm sorry for not reviewing this sooner. If we can avoid the report modal change then this is good to merge.

Gargron:
This was so long this might no longer be relevant? Closing, if it's still relevant, please reopen (and fix conflicts).

PR #4435, opened 2017-07-28
Improve accessibility (part 6)

sorin-davidoi:
Follow-up to #4369, #4377, #4405, #4408, #4417 .

Allows the content of the toots to be read accesibility software, some minor improvements.

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -146,7 +146,7 @@ export default class StatusContent extends React.PureComponent {
       }
 
       return (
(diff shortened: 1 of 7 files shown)


app/javascript/mastodon/components/status_content.js
```diff
@@ -162,6 +162,8 @@ export default class StatusContent extends React.PureComponent {
       return (
         <div
           ref={this.setRef}
+          tabIndex='0'
+          aria-label={status.get('search_index')}
```
sorin-davidoi: I know this shouldn't be used, but it already contains the perfect information (CW + content), and I would rather avoid additional encodings/decodings.

Gargron (approved):

PR #4436, opened 2017-07-29
fix tabs_bar.js

MitarashiDango:
BugFix (Navigation bar buttons does not work.)


unarist (approved):
Regression from #4429

PR #4439, opened 2017-07-29
feat: Cache status height to avoid expensive renders

sorin-davidoi:
Scenario:
 - Scroll down the federated timeline (say that there are 200 toots, but only 30 are fully rendered)
 - You find an interesting toot and click on it, being taken to the details status
 - You click back to go to the federated timeline
 - Although the scroll position is restored, we render all the 200 toots instead of just the 30 required - this can take a few seconds

The reason that we render all 200 toots is that their rendered height was lost when they were unmounted (we have to fully render them again to save the original height).

This pull request saves the height of the toots in the store, so that we only fully render the 30 toots. This "cached height" is invalidated if the window is resized.

### DIFF
--- a/app/javascript/mastodon/reducers/statuses.js
+++ b/app/javascript/mastodon/reducers/statuses.js
@@ -39,6 +39,8 @@ import emojify from '../emoji';
 import { Map as ImmutableMap, fromJS } from 'immutable';
 import escapeTextContentForBrowser from 'escape-html';
 
+const domParser = new DOMParser();
+
 const normalizeStatus = (state, status) => {
   if (!status) {
     return state;
@@ -53,7 +55,7 @@ const normalizeStatus = (state, status) => {
   }
 
   const searchContent = [status.spoiler_text, status.content].join(' ').replace(/<br \/>/g, '\n').replace(/<\/p><p>/g, '\n\n');
-  normalStatus.search_index = new DOMParser().parseFromString(searchContent, 'text/html').documentElement.textContent;
+  normalStatus.search_index = domParser.parseFromString(searchContent, 'text/html').documentElement.textContent;
   normalStatus.contentHtml = emojify(normalStatus.content);
   normalStatus.spoilerHtml = emojify(escapeTextContentForBrowser(normalStatus.spoiler_text || ''));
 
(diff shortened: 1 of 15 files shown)


Gargron (approved):

sorin-davidoi:
Would be nice if this could be deployed on a test instance to make sure nothing breaks.

beatrix-bitrot:
@sorin-davidoi aye aye

Gargron:
Has this now been confirmed to not break stuff?

sorin-davidoi:
Not yet. If this will receive extra testing I've also moved the formatting logic into the reducers (no need to call `escapeTextContentForBrowser` and `emojify` on the same input at each render).

sorin-davidoi:
This is what I'm talking about:

image

We render toots only to immediately destroy them. This makes the experience laggy and if the device runs out of the memory the browser may actually kill the tab.

app/javascript/styles/components.scss
```diff
@@ -1578,7 +1578,6 @@
   overflow-y: scroll;
   overflow-x: hidden;
   flex: 1 1 auto;
-  backface-visibility: hidden;
```
sorin-davidoi: Creates redundant layers that the browser has to keep in the GPU memory.

app/javascript/mastodon/reducers/statuses.js
```diff
@@ -33,7 +35,11 @@ import {
   FAVOURITED_STATUSES_EXPAND_SUCCESS,
 } from '../actions/favourites';
 import { SEARCH_FETCH_SUCCESS } from '../actions/search';
+import emojify from '../emoji';
 import { Map as ImmutableMap, fromJS } from 'immutable';
+import escapeTextContentForBrowser from 'escape-html';
+
+const domParser = new DOMParser();
```
ykzts: good catch!

sorin-davidoi:
Think this is enough for this PR. @beatrix-bitrot maybe you can take this for a test pretty please?

nightpool:
What's the status of the PR? Can we merge it or do we still want to test it first?

sorin-davidoi:
I'd still want some testing first.

clworld:
@sorin-davidoi I deployed this on my staging server and does some test.
It seems work without glitch. It's OK as I see.

PR #4444, opened 2017-07-29
feat(tabs_bar): Avoid optimization for non-touch devices

sorin-davidoi:
This also fixes the issue pointed out by @unarist in #4429.


unarist (approved):
thanks!

PR #4445, opened 2017-07-29
add-libpq-dev

masarakki:


Gargron (approved):

zunda:
Just for future reference, could someone teach me what systems needed `libpq-dev` added to `Aptfile`? At least, [Heroku has the package installed on the stacks (base systems)](https://devcenter.heroku.com/articles/stack-packages). The change only adds 0.4 MB to the resulting slug on Heroku so there is no need to remove or revert this.

PR #4448, opened 2017-07-30
Remove outline from focused toot

ykzts:
| before | after |
|-|-|
|  |  |

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -462,6 +462,10 @@
   overflow: hidden;
   white-space: pre-wrap;
 
+  &:focus {
+    outline: 0;
+  }
+
   .emojione {
     width: 18px;
     height: 18px;
@@ -554,6 +558,10 @@
 
   opacity: 1;
   animation: fade 150ms linear;
+
+  &:focus {
+    outline: 0;
+  }
 
   &.status-direct {
     background: lighten($ui-base-color, 8%);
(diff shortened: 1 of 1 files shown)


unarist:
Probably it's useful while navigation by keyboard, even it's useless while using mouse.

Facebook and Twitter shows outline when I use keyboard, and it doesn't show when I use mouse. I think this behavior is better.

sorin-davidoi:
Isn't this bad practice regarding accessibility?

Gargron:
I suggest replacing the outline with a highlighted background (slightly less bright than the DM background, to avoid confusion - and slightly brighter for DMs)

ykzts:
@sorin-davidoi @Gargron I changed styles.I feel like I have a better style....

### focus toot



### focus only toot content

Gargron (approved):

beatrix-bitrot:
@ykzts @Gargron i request that this change be reverted before the release of 1.5.0

the reduction in contrast makes the text significantly more difficult to read

yiskah:
Sorin is right, you need to have a way to mark that an object has focus. The outline is standard. If you must use something else you *must* do some bug-testing first to make sure you aren't making things less accessible. I agree with bea that it would be best to just revert to using the outline; that or test the contrast ratios this brings for every possible text, color, etc. that post-objects (including favs in notifs) can appear with

nightpool:
additionally, this change makes it very easy to confuse private messages with ones that are focused.

ykzts:
@beatrix-bitrot @yiskah @nightpool 

> additionally, this change makes it very easy to confuse private messages with ones that are focused.

I agree. I actually used it and had a sense of incompatibility.

Please check #4474

Issue #787, opened 2017-04-03
Memory leak?


chris-martin:
Chrome reports the Mastodon webapp uses about 75 MB of memory when it starts, 200 MB after it's been running for a while, and - after I left it open overnight - 780 MB.

wblondel:
I think that's because toots from the feed never get unloaded

Gargron:
Correct. There is no user-land garbage collection on the Redux store so to speak. It's been bugging me for a while but I'm just running on the assumption that people occasionally close the page.

Maverynthia:
I tend to leave stuff open sooo... not often I close a page.

saper:
Wouldn't assume that either...

lutoma:
I use mastodon in http://rambox.pro/, so it's pretty much open 24/7 for me.

Geobert:
350Mo after 4hours and making the whole stuff very laggy. Typing a message is sluggish… not good user experience :-/

clarfonthey:
Gonna bump this issue again since I get significant slowdowns if I don't refresh the app every once in a while. I think that the optimisation to cap the number of messages left in the feed after a while would be a good start, since I believe you can still reload more if you go down far enough? I end up with ~3k messages to load if I leave the app open overnight and then if I click that it lags so hard I can't easily type up a new post.

I think that capping at 1k messages would be a reasonable improvement, with the option to reload older posts as you read further.

This is probably a decently-sized architectural redesign for the frontend but it's definitely needed to avoid atrocious lag.

phocks:
Have noticed it happening in Firefox too. Don't seem to have the same issue with Elk web interface.

PR #4459, opened 2017-07-31
feat(redux): Persist store

sorin-davidoi:
This does not really make sense on its own, but after #3052 is ready, this will enable the usage of the website offline: #3052 will cache the assets and this will cache the timelines.

Roadmap:
- [x] Lazy-load
- [ ] Fix hydration (do not overwrite the previous state)
- [x] Choose a better storage strategy - `localForage`?
- [ ] Set an upper bound on the size of the cache
- [x] Set a flag for statuses loaded from cache and show an indication that the information is outdated
- [x] Only enable for browsers with Service Worker support
- [ ] Handle errors - wipe the caches

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -234,7 +234,7 @@ export default class Status extends ImmutablePureComponent {
     }
 
     return (
-      <article aria-posinset={index} aria-setsize={listLength} className={`status ${this.props.muted ? 'muted' : ''} status-${status.get('visibility')}`} data-id={status.get('id')} tabIndex={wrapped ? null : '0'}  ref={this.handleRef}>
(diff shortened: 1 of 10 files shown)


ykzts:
I think it is better to use `localForage`.

sorin-davidoi:
I agree. Have to look into but, but it would be nice to only load the IndexedDB driver - not need for the others since this is only for browsers with Service Worker support.

nolanlawson:
At this point I think it may make sense to directly use IndexedDB – localForage was written for a time when WebSQL still reasonably needed to be supported due to Safari/Android. Another alternative is [Dexie](http://dexie.org/).

Performance-wise I'd be a bit concerned about stuffing the entire redux store into IndexedDB as a single document – can we break it up, or can we confirm whether or not the object is prohibitively large to insert as one single object?

sorin-davidoi:
> At this point I think it may make sense to directly use IndexedDB

👍 

> Performance-wise I'd be a bit concerned about stuffing the entire redux store into IndexedDB as a single document

I don't think that is the case. Have to check, but I think that it creates a document for each of the four store states.

Gargron:
If you manage to implement this with the upper bound on cache size, you will also close #787

sorin-davidoi:
The cache size I'm referring to is the size of the serialized store in IndexDB. This won't trim the in-memory store while the page is open - another approach is needed for that.

PR #4468, opened 2017-07-31
Modify "about.scss" for mail link in /about/more page

ghost:
The default html/css link color applies the email string in /about/more page.
For fix it, this pull request modifies "/app/javascript/styles/about.scss" 

[Before]

[After]

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -150,10 +150,14 @@
       a,
       span {
         font-weight: 400;
+      }
+
+      span {
         color: lighten($ui-base-color, 34%);
       }
 
       a {
+        color: $primary-text-color;
         text-decoration: none;
       }
     }


app/javascript/styles/about.scss
```diff
@@ -150,10 +150,14 @@
       a,
       span {
         font-weight: 400;
+      }
+
+      span {
         color: lighten($ui-base-color, 34%);
```
ykzts: 🤔

lynlynlynx:
I guess that this occurred by cache.  
This is because, in before image, `text-decoration: none;` is not applied as text color is not too.

Gargron:
I cannot reproduce the "before" image...

asria-jp:
As lynlynlynx said, cache can be considered a cause.
I’m sorry for taking up your time.

PR #4471, opened 2017-07-31
fix(web_push_notifications): Send message with welcome notifications

sorin-davidoi:
Since the format message was not sent with the first notification, grouping would fail. This includes the message in the payload for the welcome notification, but also reads it from the last notification (to avoid reading the misconfigured one).

### DIFF
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -9,7 +9,7 @@ const notify = options =>
     if (notifications.length === MAX_NOTIFICATIONS) {
       // Reached the maximum number of notifications, proceed with grouping
       const group = {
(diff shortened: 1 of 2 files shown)


sorin-davidoi:
@beatrix-bitrot This was why grouping would fail on Android :disappointed:

Gargron (approved):

Komic:
@sorin-davidoi Would this require resubscribing, or should it fix already existing subscriptions too?

sorin-davidoi:
It should fix them as soon as the next one comes.

Komic:
Great, thanks. :)

sorin-davidoi:
I think when the first one comes it will be grouped together with the previous notifications - even if those were dismissed. This is a bug in the Android implementation, but this should work around that.

PR #4472, opened 2017-07-31
Do not emojify ™, ® and © - no more gray on gray

MightyPork:
Exclude ™, ® and © from emojifications which makes them unreadable.

Old:
screenshot_20170731_215141

New:
screenshot_20170731_215154

To my knowledge those symbols are supported universally, even on windows.

### DIFF
--- a/app/javascript/mastodon/emoji.js
+++ b/app/javascript/mastodon/emoji.js
@@ -3,6 +3,8 @@ import Trie from 'substring-trie';
 
 const trie = new Trie(Object.keys(unicodeMapping));
 
+const excluded = ['™', '©', '®'];
+
 function emojify(str) {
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/emoji.js
```diff
@@ -19,7 +21,7 @@ function emojify(str) {
       insideTag = true;
     } else if (!insideTag && (match = trie.search(str.substring(i)))) {
       const unicodeStr = match;
-      if (unicodeStr in unicodeMapping) {
+      if (unicodeStr in unicodeMapping && excluded.indexOf(unicodeStr) === -1) {
```
sorin-davidoi: We use ES6, so you can write `excluded.includes(unicodeStr)` (more readable).

Gargron (approved):

PR #4473, opened 2017-07-31
Add "Mute conversation" button to statuses everywhere

MightyPork:
Removed the withDismiss attribute from StatusActionBar so the option to mute a conversation is accessible everywhere without having to hunt for a notification.

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -42,7 +42,6 @@ export default class StatusActionBar extends ImmutablePureComponent {
     onReport: PropTypes.func,
     onMuteConversation: PropTypes.func,
     me: PropTypes.number,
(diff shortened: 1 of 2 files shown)


beatrix-bitrot:
for the record:

i requested this!

i've mentioned it several times in other places (including on mastodon of course) and others always expressed interest in it as well, so i asked piggo to submit this PR here rather than at the fork.

nightpool:
my problem with this is that the copy doesn't make it clear it *only* will hide a conversation from your notifications—it doesn't do anything about, say, seeing it in your home timeline

beatrix-bitrot:
oh well that's legit AF, i thought it did that also. sigh

beatrix-bitrot:
mmm maybe i will pull this to glitch.social and modify the back end then we can re-submit

PR #4474, opened 2017-07-31
Re-add outline

ykzts:
ref #4448

I actually tried it and felt that the style with DM was confusing. This Pull Request rewinds the previous change, making Google Chrome and Firefox look the same.


Gargron (approved):

ykzts:
@Gargron Respect the original of the web browser c428829.

To unify all other parts with a blue outline has many changes and it is serious. I withdraw the original Pull Request.

PR #4475, opened 2017-08-01
Add libidn11 to Aptfile

ykzts:
Fix builds that fail on Heroku. (and sorting dependencies)

reported by https://mastodon.zunda.ninja/@zundan/137510


zunda:
I confirmed that this patch solves the problem on Heroku with the heroku-16 stack.

Gargron (approved):

PR #4476, opened 2017-08-01
Change logo URL for Heroku and Scalingo

ykzts:

### DIFF
--- a/app.json
+++ b/app.json
@@ -2,7 +2,7 @@
   "name": "Mastodon",
   "description": "A GNU Social-compatible microblogging server",
   "repository": "https://github.com/tootsuite/mastodon",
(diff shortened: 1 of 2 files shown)


app.json
```diff
@@ -2,7 +2,7 @@
   "name": "Mastodon",
   "description": "A GNU Social-compatible microblogging server",
   "repository": "https://github.com/tootsuite/mastodon",
-  "logo": "https://github.com/tootsuite/mastodon/raw/master/app/javascript/images/logo.svg",
```
ykzts: The raw format of GitHub returns `Content-type: text/plain`. SVG will not be rendered unless `Content-type` is `image/svg+xml` when imported with `img` element.

I thought it would be better to use an avatar image with a permanently URL.

Gargron (approved):

PR #4481, opened 2017-08-01
Update Japanese Translation (authorize_follow)

ghost:
This pull request updates Japanese translation file. (ja.yml)

> "リモートアカウントにエラーが発生しました"

Above text means "there was an error in the remote account".
Yet, this meaning is not correct. The message is supposed to contain following meaning.

> "Das Profil konnte nicht geladen werden" (de.yml)
> "there was an error looking up the remote account" (en.yml)

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -234,7 +234,7 @@ ja:
     reset_password: パスワードを再発行
     set_new_password: 新しいパスワード
   authorize_follow:
-    error: 残念ながら、リモートアカウントにエラーが発生しました。
+    error: リモートアカウント情報の取得中にエラーが発生しました。
     follow: フォロー
     follow_request: 'あなたは以下のアカウントにフォローリクエストを送信しました:'
     following: '成功! あなたは現在以下のアカウントをフォローしています:'
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -234,7 +234,7 @@ ja:
     reset_password: パスワードを再発行
     set_new_password: 新しいパスワード
   authorize_follow:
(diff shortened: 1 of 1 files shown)


nightpool:
needs review from another japanese translator/user.

clworld (changes requested):
This translation seems better than old one.
but new one dropped 'Unfortunately,'/'残念ながら、' when compared with en.yml. https://github.com/tootsuite/mastodon/blob/master/config/locales/en.yml#L238
Is there some reasons to remove 'Unfortunately,' ?

asria-jp:
Thank you for reviews. The reasons why I removed 'Unfortunately'('残念ながら') are following. 

1. Some language files(Such as de.yml and bg.yml)  don't have a word which has same meaning as 'Unfortunately'('残念ながら').
  So, 'Unfortunately'('残念ながら') is not necessarily mandatory .

2. In Japanese software, system messages which have 'Unfortunately'('残念ながら') are not so common.

For the above reasons, I think it may be a good idea to remove 'Unfortunately'('残念ながら').

clworld:
@asria-jp 
Situations when this message displayed is:
1. request remote_follow on remote instance
2. redirected to local instance which you have account
3. this message is displayed when local instance fail to fetch remote instance's account information.

I think '3.' is suporsed to be success always. So, something unlucky is happened when it failed.
This is what is unfortunate.

Are you confirmed about this situation about removing 'Unfortunately, '?
I think message '残念ながら' is not common too. It's not natural.
Should we prefer naturality to translate or preserve original?

asria-jp:
@clworld
I've decided to preserve original word in this PR.
Thank you for your kindness.

clworld (approved):
Looks good to me!

PR #4484, opened 2017-08-01
Fix column-back-button style for some browsers (regression #4457)

unarist:
Use `text-align: unset` instead of `text-align: start` which Edge doesn't support for now.

image image

Also remove default margin on Safari.

image image

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1590,8 +1590,9 @@
   flex: 0 0 auto;
   font-size: 16px;
   border: 0;
-  text-align: start;
+  text-align: unset;
   padding: 15px;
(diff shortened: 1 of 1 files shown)


clworld (approved):
I deployed it on my staging server. and it seems OK.
'Back' button position is correct on Edge.
And properly specified margin: 0 for .column-back-button.(I don't have iOS device now to real test it. but It seems OK.)

sorin-davidoi:
I meant to write `text-align: left`... sorry :disappointed:

unarist:
Note: This shouldn't be `text-align: left` because it aligns text left even on RTL languages. `text-align: start` (= default value) should handle correctly.

PR #4488, opened 2017-08-01
feat(yarn): Upgrade eslint-plugin-jsx-a11y

sorin-davidoi:
Want to enable some more rules but we need to upgrade it first.


ykzts:
Unfortunately we have to wait for [codeclimate/codeclimate-eslint](https://github.com/codeclimate/codeclimate-eslint) update:cry:

PR #4489, opened 2017-08-01
Dutch strings: typo

ghost:


config/locales/nl.yml
```diff
@@ -220,7 +220,7 @@ nl:
   applications:
     invalid_url: De opgegeven URL is ongeldig
   auth:
-    agreement_html: Wanneer je op registeren klikt ga je akkoord met <a href="%{rules_path}">onze gebruikersvoorwaarden</a> en <a href="%{terms_path}">ons privacybeleid</a>.
```
ykzts: added from https://github.com/tootsuite/mastodon/commit/f52ce92f2bc989154a69d20ab12cad927f94bc24#diff-9affb47e33db13b0717e5ffc5b2ef587R223 (author @jeroenpraat)

http://www.linguee.com/english-dutch/search?source=dutch&query=registreren

ykzts:
cc @Fjoerfoks

config/locales/nl.yml
jeroenpraat: @ykzts What is the issue?

I only corrected a previous typo. 'registeren' to 'registreren'

ykzts: This is a note.

jeroenpraat: @ykzts Why are you asking another Dutch speaker for me fixing a simple typo?

jeroenpraat: @ykzts I am a native Dutch speaker and I translated 99% of Mastodon. Actually I fixed my own typo. So hopefully you understand I find this very strange.

ykzts: I'm checking `*.nl.yml` and `nl.json` now. I saw you translating the majority of parts. I do not think this is a problem, but I also want other people's opinions.

jeroenpraat: 'Another people's opinion' for a (self-made) typo? 

'registeren' is not a word. 'Registreren' is. It means in English 'to register' (an account). I asked @Gargron to come here, cause I find this not very productive.

nightpool:
@jeroenpraat Hi Jeroen. Recently, we're trying to hold our translations to a higher quality by requiring at least one fluent speaker of the language to double-check them before we merge them into the repo. While I understand this seems like a difficult change, this is a much less strict standard then we hold any other type of contribution to. If you can get any other Dutch speaker with a github account to take a look at this pull request and leave a comment indicating that they're happy with it, that's good enough for us.

We're just trying to prevent typos like this from showing up in the first place!

jeroenpraat:
@nightpool This will not work, maybe for languages with a lot of Mastodon users, but not Dutch or any other language with a low user count on Mastodon. This will frustrate or even hold up translations. Now I see there is someone cc't that is not even using Mastodon. The only solution to achieve what you want is to use a 3rd party translation service.

I also have problems about the fact this is not communicated with translators at all. Or maybe we don't even a good list with translators, but that is because we don't use a separate service for translations.

Until this is solved, I won't add my translations to Github anymore, but only directly to my instance @ todon.nl. When Mastodon will use a 3rd party service, I will be glad to add it there.

nightpool:
@jeroenpraat I'm going to look into setting up weblate this weekend.

But if there are at least two users for any given language, then I don't see the problem with having one of them checking the other's work. I don't see why this would be more of a problem on github vs on a third party service like weblate?

jeroenpraat:
@nightpool That will only work if there are at least two (Dutch) users on Weblate. If there is only one, you have to trust that person or ignore the translation completely. If I am over there still the only one, the Mastodon devs can choose to trust me or choose not to update the Dutch translation. But don't cc some random Github user for a simple typo fix (I made myself).

jeroenpraat:
@nightpool Also something like Weblate or Transifex is more accessible for non-developers.

nightpool:
It wasn't a random github user, it was a user who had previously contributed Dutch translations to the repo.

Anyway, yes. this argument is stupid to have about a typo fix.

nightpool (approved):

jeroenpraat:
@nightpool Thanks.

PR #4492, opened 2017-08-02
fr.json update

ghost:


app/javascript/styles/basics.scss
```diff
@@ -1,6 +1,6 @@
 body {
   font-family: 'mastodon-font-sans-serif', sans-serif;
-  background: $ui-base-color;
+  background: $ui-base-color url('../images/background-photo.png');
```
ykzts: Is it the intended change?

Komic: nope; one sec

Komic: @ykzts there you go, sorry about that.

ykzts (approved):

PR #4494, opened 2017-08-02
Use GNU libiconv in Nokogiri

ykzts:
System default libiconv of Alpine Linux only supports some charset (e.g. UTF-8). Therefore, the preview card of the page which is not UTF-8 will be broken in the Docker environment.

Solve using GNU libiconv.

before | after
-|-
before screenshot | after screenshot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,9 @@ ENV UID=991 GID=991 \
     RAILS_SERVE_STATIC_FILES=true \
     RAILS_ENV=production NODE_ENV=production
 
+ARG LIBICONV_VERSION=1.15
+ARG LIBICONV_DOWNLOAD_SHA256=ccf536620a45458d26ba83887a983b96827001e92a13847b45e4925cc8913178
+
 EXPOSE 3000 4000
 
 WORKDIR /mastodon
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,8 +20,6 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
     build-base \
     icu-dev \
     libidn-dev \
-    libxml2-dev \
-    libxslt-dev \
```
ykzts: `libxml2` and `libxslt` will be downloaded during Nokogiri installation.

Wonderfall:
I also use GNU Libiconv in some of my Docker images for that reason. So +1 for this.
Isn't `libtool --finish /usr/local/lib` needed? Also it would be better if `$LIBICONV_VERSION` was set with a build-time variable (`ARG`) instead of an environment variable.

ykzts:
@Wonderfall Thank you for review!

Wonderfall:
The `libtool` package must be added as well. 😉 
Also `--prefix /usr/local` can be added as a parameter to `configure`.

I tried myself (https://github.com/Wonderfall/dockerfiles/blob/master/mastodon/Dockerfile.targaryen#L60-L65), this change is working.

+1 for merging!

ykzts:
@Wonderfall Okay! Thx😚

Gargron (approved):

PR #4500, opened 2017-08-02
make number of comparison in emojify() fewer

hcmiya:


app/javascript/mastodon/emoji.js
```diff
@@ -1,36 +1,37 @@
 import { unicodeMapping } from './emojione_light';
 import Trie from 'substring-trie';
 
-const trie = new Trie(Object.keys(unicodeMapping));
+const excluded = ['®', '©', '™'];
+excluded.forEach(c => {
+  if (unicodeMapping[c] !== undefined) {
+    delete unicodeMapping[c];
```
sorin-davidoi: Remove the strings on the server (`emojione_light.js`) - no need to ship them if they are not used.

sorin-davidoi (changes requested):

app/javascript/mastodon/emoji.js
hcmiya: I moved it to emojione_light.js

sorin-davidoi (approved):

sorin-davidoi:
My feedback was addressed and tests pass, so I guess this is OK? Thanks, @hcmiya, for taking the time and addressing this.

beatrix-bitrot (approved):
Feedback was addressed and looks ok to Sorin so it is ok by me

akihikodaki:
I have measured the performance but I cannot see any improvement. I inserted the following code into `app/javascript/mastodon/features/emoji/emoji.js`.

```JavaScript
function veryOldEmojify(str) {
  // This walks through the string from start to end, ignoring any tags (<p>, <br>, etc.)
  // and replacing valid unicode strings
  // that _aren't_ within tags with an <img> version.
  // The goal is to be the same as an emojione.regUnicode replacement, but faster.
  let i = -1;
  let insideTag = false;
  let match;
  while (++i < str.length) {
    const char = str.charAt(i);
    if (insideTag && char === '>') {
      insideTag = false;
    } else if (char === '<') {
      insideTag = true;
    } else if (!insideTag && (match = trie.search(str.substring(i)))) {
      const unicodeStr = match;
      if (unicodeStr in unicodeMapping && excluded.indexOf(unicodeStr) === -1) {
        const [filename, shortCode] = unicodeMapping[unicodeStr];
        const alt      = unicodeStr;
        const replacement =  `<img draggable="false" class="emojione" alt="${alt}" title=":${shortCode}:" src="/emoji/${filename}.svg" />`;
        str = str.substring(0, i) + replacement + str.substring(i + unicodeStr.length);
        i += (replacement.length - unicodeStr.length); // jump ahead the length we've added to the string
      }
     }
   }
  return str;
}

const oldEmojify = str => {
  let rtn = '';
  for (;;) {
    let match, i = 0;
    while (i < str.length && str[i] !== '<' && !(match = trie.search(str.slice(i)))) {
      i += str.codePointAt(i) < 65536 ? 1 : 2;
    }
    if (i === str.length)
      break;
    else if (str[i] === '<') {
      let tagend = str.indexOf('>', i + 1) + 1;
      if (!tagend)
        break;
      rtn += str.slice(0, tagend);
      str = str.slice(tagend);
    } else {
      const [filename, shortCode] = unicodeMapping[match];
      rtn += str.slice(0, i) + `<img draggable="false" class="emojione" alt="${match}" title=":${shortCode}:" src="/emoji/${filename}.svg" />`;
      str = str.slice(i + match.length);
     }
   }
  return rtn + str;
};

requestIdleCallback(() => {
  const example = 'いろはにほへとちりぬるをわかよたれそつねならむうゐのおくやまけふこえてあさきゆめみしゑひもせすん　あめつちほしそらやまかはみねたにくもきりむろこけひといぬうへすゑゆわさるおふせよえのえをなれゐて　あめふれはゐせきをこゆるみつわけてやすくもろひとおりたちうゑしむらなへそのいねよまほにさかえぬ　ひふみよいむなやこともちろらねしきるゆゐつわぬそをたはくめかうおえにさりへてのますあせゑほれーけー　とりなくこゑすゆめさませみよあけわたるひんかしをそらいろはえておきつへにほふねむれゐぬもやのうち';
  performance.mark('emoji-start');
  for (let count = 0; count < 256; count++) {
    oldEmojify(example);
    //veryOldEmojify(example);
  }
  performance.mark('emoji-end');
  performance.measure('emoji', 'emoji-start', 'emoji-end');
  console.log(performance.getEntriesByName('emoji'));
});
```

very old (before this change):
```
48.05000000000018
34.73999999999978
29.829999999999927
29.519999999999527
```

old (after this change):
```
48.529999999999745
30.940000000000055
37.725000000000364
33.91000000000031
```

PR #4509, opened 2017-08-03
i18n: Update Polish translation

mkljczk:


m4sk1n:
CC @KrzysiekJ

config/locales/pl.yml
```diff
@@ -228,6 +228,7 @@ pl:
     delete_account_html: Jeżeli chcesz usunąć konto, <a href="%{path}">przejdź tutaj</a>. Otrzymasz prośbę o potwierdzenie.
     didnt_get_confirmation: Nie otrzymałeś instrukcji weryfikacji?
     forgot_password: Nie pamiętasz hasła?
+    invalid_reset_password_token: Link resetowania hasła jest nieważny lub nieprawidłowy. Spróbuj ponownie.
```
KrzysiekJ: Maybe „Link do resetowania hasła”? Besides that, if link is either invalid or expired, then trying again seems to be pointless (this applies also to the English translation).

m4sk1n: Yup, it should be something like “Try to get another one” or tip like “Have you already used it?”

m4sk1n:
@KrzysiekJ can you review (and approve) it now?

config/locales/pl.yml
KrzysiekJ: What about the small issue with missing “do”? Actually, since the real issue is not with the link (an URL may be even copy-pasted from an email without a link), but with a token, I would make it „Token resetowania hasła jest nieważny lub nieprawidłowy” — if we’re not obligated to translate from English…

PR #4511, opened 2017-08-03
Add protocol handler. Handle follow intents

### LINKED ISSUE #442: Intent/share URLs
A really good simple step for letting third-party apps/sites/etc share to Mastodon would be implementing an Intent URL (e.g. https://dev.twitter.com/web/intents ).

At the simplest, I can imagine just opening the /web of the mastodon instance, with a parameter that pre-fills the toot text?

### LINKED ISSUE #2291: URL scheme for remote follow, share buttons
Decentralized nature of Mastodon causes rather awkward flow for the "Remote Follow" operation. Similarly, there's no way to make an easy "Share" button.

The root of the problem is that websites don't know which Mastodon instance they should use to perform the operations for their visitors. 

I think this problem can be solved quite well with custom URL schemes. It's possible to register a website (a Mastodon instance) to handle links with a custom URL scheme ([`registerProtocolHandler`](https://developers.google.com/web/updates/2011/06/Registering-a-custom-protocol-handler)). This registration is per user/per browser, so it's compatible with the decentralized system and doesn't rely on any central infrastructure.

There could be a URL scheme for Mastodon commands, such as "follow" or "toot". Users could allow their preferred instance or app to handle this URL scheme. Even users who use multiple instances could assign the URL scheme to a "chooser" app that lets them redirect the request to any of their instances.

The URL could be something like this:

    web+toot:follow/@handle@instance

and then the Remote Follow button could be something like this:

```html
<a href="web+toot:follow/@handle@instance">Remote Follow</a>
```

(actually bit more complex to have an HTTP URL fallback, and act more like button, especially in screen readers, but you get the idea)

This way a click on the URL could instantly redirect user to their preferred instance, without asking for the handle and the instance. Similarly, URLs like `web+toot:share?text=Hello` could be used to implement "Share" buttons that work as seamlessly as on centralized social networks.

Gargron:
Example: <https://jsfiddle.net/pvyoof6L/3/> (to work properly, run this PR on localhost, and open web UI once prior to running the fiddle)

TODO:

- [x] Follow intent
- [x] Share (toot) intent

image

image

- Fix #442
- Fix #2291


nightpool:
cc @sandhawke, https://github.com/swicg/general/issues/5

nightpool:
what's the reasoning behind using web+mastodon?

if we're using authorize_follow_path can we make it web+ostatus maybe? is remote follow an ostatus thing?

Gargron:
mastodon logo + "follow me on mastodon" + web+mastodon://

As long as somebody doesn't come up with a *standard*, I'm not going to claim any namespaces I have no authority on. This way we won't step on any future standards toes.

sandhawke:
So, if someone makes a fork of Mastodon do they  have to call it Mastodon in the protocol to be interoperable?

Gargron:
>So, if someone makes a fork of Mastodon do they have to call it Mastodon in the protocol to be interoperable?

Yes.

I don't know what you want me to do here. We don't have a standard, the linked swicg issue is not conclusive. You want me to name it ostatus or x-ostatus, but I doubt we'll switch everybody's buttons once Mastodon starts supporting ActivityPub, and then it'll be an ostatus protocol link going to an ActivityPub thing. I'd support a generic naming like "social" but then if someone wanted to invent a standard for that later they'd have to deal with backcompatibility.

nightpool:
But if we switch to activitypub, the API/buttons will have to change *anyway*, because they should be passing around uris instead of webfinger accounts?

Gargron:
No, in our case they should still pass around webfinger accounts. I've already said before - without webfinger you can't have humans interacting with ActivityPub accounts, because nobody is gonna remember the URLs, but webfinger accounts are easy to remember and type and they are also iconic. tl;dr we will never abandon webfinger. ActivityPub URLs matter only within ActivityPub itself, e.g. when those APIs talk to each other, not when interfacing with people.

nightpool:
Ah, I must have missed this detail. I understand why webfinger is good for the user side, but are you saying it would be required to interoperate with mastodon?

Gargron:
Yes, it's required to interoperate with Mastodon. There's just no point having accounts in the DB that you can't write a mention of. This might change in the future.

app/controllers/intents_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      when 'follow'
+        redirect_to authorize_follow_path(acct: uri.query_values['uri'].gsub(/\Aacct:/, ''))
+      when 'share'
+        redirect_to share_path(text: uri.query_values['text'])
+      else
+        not_found
+      end
+    else
+      not_found
+    end
```
nightpool:
```
  end
end

not_found
```
?

Gargron:
I am looking for UX feedback.

nightpool:
I understand the motivation, but that last commit should be a separate PR.

tcitworld:
You could first have a check with `isRegisteredProtocolHandler()` and then maybe use `unregisterProtocolHandler()` when the account is disabled ?

nightpool:
UX feedback:

 - It needs some kind of modal flow for adding the handler: Right now it's incredibly easy to miss the tiny handler register if you're not looking for it.

image

 - it's really jarring that the compose form doesn't close or show a "done!" screen when you're done with it—same with the follow form not closing after you've confirmed the follow. (it should have a "close" button, at least? this might be a thing we can just progressively add if we have the permissions for it, since normal tabs can't close themselves but iirc floating windows can)

 - I think the "follow me" button can pretty obviously be served by the instance that the user is on, although that's only a convenience. the "share" button is maybe less obvious—joinmastodon would probably be a fine place for it? 

 - one thing I would really like to see is progressively replacing "Remote Follow" buttons with these new "Follow" buttons if the user has a protocol handler registered. That would be a great first in-the-wild usage for these. It should still have a small "Or follow using your full handle..." link below for the full remote follow flow, just in case someone wants to use a second account or something (I don't think current implementations of registerProtocolHandler expose non-default handlers in a useful way, unfortunately...) 

   this is technically somewhat difficult, since contra @tcitworld there isn't actually a isRegisteredProtocolHandler API function that I can find, but adding the URI for an intent like "noop" or "web+mastodon://available" to an img element, and listening for onLoad or onError from that *should* give you the information you need to replace the "Remote Follow" button with a intent-enabled button

Gargron:
>It needs some kind of modal flow for adding the handler: Right now it's incredibly easy to miss the tiny handler register if you're not looking for it.

In firefox you get an entire top bar asking for permission.

I am actually a bit concerned about this UX because when you open Mastodon for the very first time, you get this:

- Onboarding modal
- Desktop notifications permission popup
- "You can now receive push notifications!" push notification (actually not sure if this is only after the previous point)
- And now "would you like to register protocol handler" too

nightpool:
yes I agree, we should stagger these out incrementally as users get notifications/take actions/spend some time on the page

(yeah 3 is dependent on 2 and probably can be removed—it was mostly left in to aid debugging)

ykzts (approved):

matyapiro31:
Q: How to toot "+" with this url?

Gargron:
@matyapiro31 You'll need to use urlencode

Cassolotl:
Is there any documentation on how to use this, especially with Wordpress? I am looking at the front page of the [documentation](https://github.com/tootsuite/documentation) and I have no idea which link I should follow to find what I need!

PR #4512, opened 2017-08-03
Move <Button /> styles to buttons.scss

sorin-davidoi:
There was a brief discussion about breaking up `components.scss`. Decided to start with one of the smallest components: `<Button />`.

I've decided to place the stylesheet next to the component. Currently, this goes against how the file structure is organized. I would argue that it is better to have them close together (maybe even go the next step and create a `button` directory which would contain the component, its stylesheet and its tests).


sorin-davidoi:
No really sure what to do with some style overwrites. The point of `button.scss` it to contain **all** the button's styles. Should the following go here or stay in `components.scss`? I would rather keep it in `button.scss` - this way if you update its styles you don't have to search in the whole project to see if someone, somewhere overwrite something.

```scss
.boost-modal__action-bar,
.confirmation-modal__action-bar,
.report-modal__action-bar {
  .button {
    flex: 0 0 auto;
  }
}
```

akihikodaki (changes requested):
This breaks custom stylesheets, doesn't it? Breaking the fat stylesheet (and hopefully introducing CSS module) would be really nice, but we should consider about that before implementing.

Issue #4510, opened 2017-08-03
opening a status in the detail pane should focus that status


nightpool:
Because the actual focus of the UI interaction moves (when someone clicks on something, we expect them to be looking at the detailed status pane) the :focus styling should move as well, right?

mjankowski:
Still relevant on 4.3? And if so, can you confirm which interface, and/or add screenshots...

PR #4517, opened 2017-08-04
Scroll columns area to right when children property is changed

akihikodaki:
The feature to pin column could hide the rightmost column, which is specified with `children` property of `ColumnsArea`.
The user is likely to see the column when the property changed, so scroll the area in such cases.


sorin-davidoi:
Maybe we should also focus the children (#4510)?

akihikodaki:
That makes sense, but I want to address a particular problem in this pull request: a column is not visible and users consider it is malfunctioning. The issue would need another pull request.

Gargron (approved):

PR #4524, opened 2017-08-05
Refactor Web::PushSubscription, remove welcome message

Gargron:
One caveat: `dir` is currently always `ltr` because `body` is inside DataSerializer and `dir` is outside.

Resolves #4519


ykzts:
@sorin-davidoi How about this? It looks good to me.

nightpool:
LGTM modulo conflicts and figuring out something about the dir.

can we just move that code around in the service worker? i'm assuming it doesn't *need* to be where it is...

beatrix-bitrot:
@nightpool can you elaborate on moving the code around the service worker?

ykzts (approved):

nightpool:
@beatrix-bitrot 

> dir is currently always ltr because body is inside DataSerializer and dir is outside

Why not just include `dir` in the data payload and not the outer payload?

Gargron:
Need @sorin-davidoi's input on that :(

sorin-davidoi:
Sorry for neglecting this for a while. If I understand correctly, we could include dir in the data payload, but then we need to copy it outside - see https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/service_worker/web_push_notifications.js#L46. Looks good to me but I cannot test this at the moment.

Gargron:
Okay. Would appreciate @beatrix-bitrot to test whether push notifs still work correctly after this PR.

PR #4526, opened 2017-08-05
Refactor Avatar and AvatarOverlay to have 'account' as prop instead of src and staticSrc

MightyPork:
- Avatar and AvatarOverlay now receive `account` prop instead of `src` and `staticSrc`. This avoids potential errors and removes needlessly verbose code elsewhere. I changed both to feel uniform and also because it's useful for adding other features to Avatar based on the account in the future
- Also removed animate attribute from compose navigation bar, which should
have never been there (gif keeps nonstop flashing in user's face)
- Added test for avatar overlay.

### DIFF
--- a/app/javascript/mastodon/components/account.js
+++ b/app/javascript/mastodon/components/account.js
@@ -70,7 +70,7 @@ export default class Account extends ImmutablePureComponent {
       <div className='account'>
         <div className='account__wrapper'>
           <Permalink key={account.get('id')} className='account__display-name' href={account.get('url')} to={`/accounts/${account.get('id')}`}>
-            <div className='account__avatar-wrapper'><Avatar src={account.get('avatar')} staticSrc={account.get('avatar_static')} size={36} /></div>
+            <div className='account__avatar-wrapper'><Avatar account={account} size={36} /></div>
             <DisplayName account={account} />
           </Permalink>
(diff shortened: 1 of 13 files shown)


Gargron:
>(gif keeps nonstop flashing in user's face)

I don't understand why this is a problem - the navigation bar displays *your own* avatar, if you're disturbed by the flashing, why do you set your avatar to a flashing picture?

MightyPork:
@Gargron I don't, but many people have animated avatars (eg. @jk's spinning coyote) that must be mightily distracting when it moves there all the time, instead of on hover as usual. (I noticed this when testing animated avatars on localhost)

beatrix-bitrot:
since they normally only animate on hover i think it is not so surprising that people would set avatars that they don't  want to see animate all the time forever

jk:
Spinning coyote??

MightyPork:
@jk sorry I didn't realize there's a user with the handle on github, it's nothing about you xD

sorin-davidoi:
Don't have time to test it right now but code-wise looks good :+1:

ykzts (approved):

beatrix-bitrot:
since 1.5.1 is out now, @Gargron or @ykzts please either merge or close this~

PR #4531, opened 2017-08-06
fr.yml update

ghost:
Translation for https://github.com/tootsuite/mastodon/pull/4523

PR #4533, opened 2017-08-06
Fix #4531

ghost:
wrong character, sorry about that~


ykzts (approved):
https://github.com/tootsuite/mastodon/pull/4532#issuecomment-320505018

PR #4541, opened 2017-08-07
Do not scroll the columns area due to redirection

akihikodaki:
Commit 9d1f8b9d6af1d384a4dd68bc6353a2fde5735b33 scrolls the columns area when the route changes since the user is likely to want to see the ightmost column in such cases.

However, redirection is automatic and does not indicate users' intension. Do not scroll the columns area due to one.

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -1,12 +1,12 @@
 import React from 'react';
 import classNames from 'classnames';
-import Redirect from 'react-router-dom/Redirect';
 import NotificationsContainer from './containers/notifications_container';
 import PropTypes from 'prop-types';
 import LoadingBarContainer from './containers/loading_bar_container';
 import TabsBar from './components/tabs_bar';
 import ModalContainer from './containers/modal_container';
 import { connect } from 'react-redux';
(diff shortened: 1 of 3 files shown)


app/javascript/mastodon/features/ui/index.js
```diff
@@ -7,6 +7,7 @@ import LoadingBarContainer from './containers/loading_bar_container';
 import TabsBar from './components/tabs_bar';
 import ModalContainer from './containers/modal_container';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
```
ykzts:
```diff
-import { withRouter } from 'react-router';
+import { withRouter } from 'react-router-dom';
```

add `-dom`.

akihikodaki:
2b8d6d6358b84a0aee3b88889b34ceba73034b44 applies https://github.com/tootsuite/mastodon/pull/4541#discussion_r131578916. It also merges an existing import (`Redirect`).

unarist:
@Gargron This bug happens every time I access `http://some.instance.com/` (then redirected to `/web/getting-started`) and really bugs me. Could you take a look this?

Gargron (approved):

PR #4544, opened 2017-08-07
[Docker] Add multicore support to "make" and "bundler"

ThomasLeister:
These changes add multicore support for ```make``` and ```bundler``` so ```docker-compose build``` time is decreased on multicore machines.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,7 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
  && rm libiconv.tar.gz \
  && cd /tmp/src/libiconv-$LIBICONV_VERSION \
  && ./configure --prefix=/usr/local \
- && make \
+ && make -j$(getconf _NPROCESSORS_ONLN)\
  && make install \
  && libtool --finish /usr/local/lib \
  && cd /mastodon \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -48,15 +48,15 @@ RUN echo "@edge https://nl.alpinelinux.org/alpine/edge/main" >> /etc/apk/reposit
...
+ && make -j$(getconf _NPROCESSORS_ONLN)\
  && make install \
  && libtool --finish /usr/local/lib \
  && cd /mastodon \
  && rm -rf /tmp/* /var/cache/apk/*
 
 COPY Gemfile Gemfile.lock package.json yarn.lock /mastodon/
 
-RUN bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-include=/usr/local/include \
+RUN bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-include=/usr/local/include --jobs $(getconf _NPROCESSORS_ONLN)\
```
ykzts: This is an option used in build for Nokogiri. Is it okay to add it here?

ThomasLeister: I'm not a Ruby / Bundler expert, but according to this article: https://robots.thoughtbot.com/parallel-gem-installing-using-bundler the ```--jobs``` option can be applied to every bundler config command. Worked for me.

ykzts: Yes, but that line is not `bundle install`.

ThomasLeister: https://bundler.io/v1.5/whats_new.html

> The --jobs option (-j for short) installs gems in parallel. For example, bundle install -j4 will use 4 workers. We've seen speedups of 40-60% on fresh bundle installs. To always install in parallel, run bundle config --global jobs 4 or set BUNDLE_JOBS

Okay, let me better add the -j option to the "bundle install" command.

ykzts (approved):
cool!

PR #4546, opened 2017-08-07
Added new Catalan strings

ghost:


ykzts (changes requested):
It seems that changes other than language files are also included. Can not remove these?

PR #4547, opened 2017-08-07
Set false to animated options for thumbnail processor

### LINKED ISSUE #3199: Transparent GIF avatars are broken
If you have a transparent GIF avatar, Mastodon automatically assumes you want it to crop the first frame of all transparency for some reason. Somehow, it manages to ruin the entire GIF along the way.

Reproduce:
1. Make a GIF avatar that has transparency and does not completely fill available space (specifically on the first frame)
2. Watch as Mastodon mangles it beyond recognition depending on how much space there was

* * * *

ykzts:
Resolve #3199

Fix the aspect ratio of animated GIF whose background is transparent.

### DIFF
--- a/app/models/concerns/account_avatar.rb
+++ b/app/models/concerns/account_avatar.rb
@@ -8,7 +8,7 @@ module AccountAvatar
   class_methods do
     def avatar_styles(file)
       styles = { original: '120x120#' }
-      styles[:static] = { format: 'png' } if file.content_type == 'image/gif'
+      styles[:static] = { animated: false } if file.content_type == 'image/gif'
       styles
     end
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

clworld:
I experienced glitch that static version of animated icon is not shown.
It seems filename of static thumbnail is changed but not regenerated.
Filename is changed from system/accounts/avatars/X/X/X/static/XXX.png to system/accounts/avatars/X/X/X/static/XXX.gif, but XXX.gif is not exists.
Is it needed regenerating whole animated icon's static thumbnails?

Gargron:
Hmm, you are right. I forgot that paperclip just assumes the format specified in the style is the extension. I don't want to force people to regenerate everything...

ykzts:
I forgot that this also applies to past things... Requiring regeneration would be a hassle for instance administrators.

ykzts:
I think that it is better to revert. #4550 

 I will consider a better way.

PR #4548, opened 2017-08-07
unify short description styling

TheInventrix:
Applies the same style class to the About description on both the landing page
and the about/more page.

This allows more consistent and simpler customization of instances.

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -484,7 +484,7 @@
     padding: 0;
   }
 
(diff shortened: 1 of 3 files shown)


Gargron (approved):

ykzts:
It looks broken from my eyes. Is this what you intended?

TheInventrix:
Dammit, I must have missed some part of my changes when setting it up for a PR, I'll take a look and see if I can find what it was.

PR #4560, opened 2017-08-08
Add missing @ to the onboarding modal

MightyPork:
before
screenshot_20170808_215547

after
screenshot_20170808_215612

Without the @, it's kind of confusing because the rest of the tutorial uses the handle with @. Also you need the @ to mention someone, so if a newbie learns their handle is without the @, it could cause some confusion (esp. bc it looks like an e-mail while it's not)


mitchellurgero:
Solid work.

Gargron (approved):

PR #4566, opened 2017-08-09
ActivityPub delivery

Gargron:
TODO:

- [x] Delivery worker
- [x] FavouriteService
- [x] UnfavouriteService
- [x] ReblogService
- [x] PostStatusService
- [x] RemoveStatusService
- [x] BatchedRemoveStatusService
- [x] FollowService
- [x] UnfollowService
- [x] BlockService
- [x] UnblockService
- [x] ProcessMentionsService
- [x] AuthorizeFollowService
- [x] RejectFollowService
- [x] Distribution worker
- [x] Update account

### DIFF
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -10,8 +10,9 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
   end
 
   def update
-    current_account.update!(account_params)
     @account = current_account
+    @account.update!(account_params)
+    ActivityPub::UpdateDistributionWorker.perform_async(@account.id)
     render json: @account, serializer: REST::CredentialAccountSerializer
   end
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -172,7 +172,7 @@ class Account < ApplicationRecord
     end
 
     def inboxes
-      reorder(nil).where(protocol: :activitypub).pluck('distinct coalesce(nullif(accounts.shared_inbox_url, \'\'), accounts.inbox_url)')
+      reorder(nil).where(protocol: :activitypub).pluck("distinct coalesce(nullif(accounts.shared_inbox_url, ''), accounts.inbox_url)")
     end
 
     def triadic_closures(account, limit: 5, offset: 0)
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -171,6 +171,10 @@ class Account < ApplicationRecord
       reorder(nil).pluck('distinct accounts.domain')
     end
 
+    def inboxes
+      reorder(nil).where(protocol: :activitypub).pluck('distinct coalesce(nullif(accounts.shared_inbox_url, \'\'), accounts.inbox_url)')
+    end
+
     def triadic_closures(account, limit: 5, offset: 0)
       sql = <<-SQL.squish
         WITH first_degree AS (
--- a/app/services/batched_remove_status_service.rb
+++ b/app/services/batched_remove_status_service.rb
@@ -15,9 +15,11 @@ class BatchedRemoveStatusService < BaseService
     @mentions = statuses.map { |s| [s.id, s.mentions.includes(:account).to_a] }.to_h
     @tags     = statuses.map { |s| [s.id, s.tags.pluck(:name)] }.to_h
 
-    @stream_entry_batches = []
-    @salmon_batches       = []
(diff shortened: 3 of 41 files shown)


app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -10,8 +10,9 @@ def show
   end
 
   def update
-    current_account.update!(account_params)
     @account = current_account
+    @account.update!(account_params)
+    ActivityPub::UpdateDistributionWorker.perform_async(@account.id)
```
nightpool: this feels like it should either be in a model callback or a service

app/models/account.rb
```diff
@@ -171,6 +171,10 @@ def domains
       reorder(nil).pluck('distinct accounts.domain')
     end
 
+    def inboxes
+      reorder(nil).where(protocol: :activitypub).pluck('distinct coalesce(nullif(accounts.shared_inbox_url, \'\'), accounts.inbox_url)')
```
nightpool: minor, but seems like using double quotes here would make it so you don't have to escape the sql single quotes?

nightpool: do we need an index on this?

app/services/batched_remove_status_service.rb
```diff
@@ -50,6 +57,22 @@ def batch_stream_entries(statuses)
     end
   end
 
+  def batch_activity_json(account, statuses)
+    account.followers.inboxes.each do |inbox_url|
```
nightpool: Delete isn't an addressed activity, and publicInbox is only for activities that are addressed to the Public collection. this feels like it should maybe be addressed by https://github.com/w3c/activitypub/issues/242 although that hasn't been spec'd yet.

app/services/remove_status_service.rb
```diff
@@ -38,15 +40,48 @@ def remove_from_followers
     end
   end
 
-  def remove_from_mentioned(stream_entry)
-    salmon_xml       = stream_entry_to_xml(stream_entry)
-    target_accounts  = @mentions.map(&:account).reject(&:local?).uniq(&:domain)
+  def remove_from_remote_mentioned
```
nightpool: remote_mentioned should either be renamed to remote_reblogged_or_mentioned or this should be split into two methods.

app/services/remove_status_service.rb
```diff
@@ -38,15 +40,48 @@ def remove_from_followers
...
+    # People who got mentioned in the status, or who
+    # reblogged it from someone else might not follow
+    # the author and wouldn't normally receive the
+    # delete notification - so here, we explicitly
+    # send it to them
+
+    target_accounts = (@mentions.map(&:account).reject(&:local?) + @reblogs.map(&:account).reject(&:local?)).uniq(&:id)
+
```
nightpool: If I have a status because someone I follow reblogged it, but I didn't reblog it myself, I wouldn't receive a delete. 

ActivityPub does not specify how to federate a Delete activity. tagging @cwebber because that feels like something the group should address.

app/workers/activitypub/delivery_worker.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class ActivityPub::DeliveryWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'push', retry: 5, dead: false
+
+  HEADERS = { 'Content-Type' => 'application/activity+json' }.freeze
```
nightpool: ActivityPub does not specify the content type for Server-to-Server interactions, but https://w3c.github.io/activitypub/#client-to-server-interactions implies that the *primary* content type of ActivityPub is `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`, and that all clients need to use that content type and that's the only one that's a `must` in the spec (`activity+json` is a `should`)

nightpool (changes requested):

nightpool:
Looks good! some specific stuff i'm not super sure about, and I didn't read the tests *super* throughly but the rest of it looks great

app/controllers/api/v1/accounts/credentials_controller.rb
Gargron: - Can't be in a callback, callbacks execute in-transaction (queued job might get old data)
- Can't be a service because while this controller uses `update!`, the preferences controller uses `if update` (different usage)

With only two usages, it's easier to just have this one extra line in both places.

app/services/batched_remove_status_service.rb
Gargron: sharedInbox does not have any limitations.

PR #4568, opened 2017-08-09
Support to response of ActivityPub that returns image URL

ykzts:
As far as I confirm, I can not fetch the avatar and header of the remote instance account from Mastodon 1.4.7 to 1.5.1.


Gargron:
I think that the old syntax was wrong. When a property can be either a string/array of strings, or object/array of objects, it means the string must be the URI that can be resolved to the object - not the URL to the file, as I initially thought.

ykzts:
Yes it is. But considering compatibility, I think that it is better to supports for string URL.

ykzts:
I think it will be solved in #4589.

PR #4570, opened 2017-08-09
fix text position of NSFW in Safari

pfm-eyesightjp:
`button` + `display: flex;` not work in Safari (Mac/iPhone).

cw

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -2315,19 +2315,11 @@ button.icon-button.active i.fa-retweet {
 }
 
 .media-spoiler {
-  align-items: center;
   background: $base-overlay-background;
   color: $primary-text-color;
-  cursor: pointer;
-  display: flex;
-  flex-direction: column;
   border: 0;
   width: 100%;
   height: 100%;
-  justify-content: center;
-  position: relative;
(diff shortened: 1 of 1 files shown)


ykzts (approved):

unarist (approved):
@Gargron don't we merge this?

Gargron:
@unarist I don't understand how it's supposed to center anything after flexbox is removed? 🤔

unarist:
Oh, I didn't care about that.

When I copied all computed styles to inline style using `Object.assign(elem.style,getComputedStyle(elem))` and changed to the div element, vertical centering disappears. So, looks like special handling on the button tag...

ykzts:
@Gargron 

> I don't understand how it's supposed to center anything after flexbox is removed? 

Even if you do not have `display: flex`, the `button` element style will be used.

Gargron:
Now it looks like this in Firefox:

bd7715b6e1

nullkal:
I confirmed it is displayed correctly in Firefox 55.0.3 on macOS 10.12.6 (Sierra).

pfm-eyesightjp:
oh, sorry...
This problem seems to be occurring when posting video file(mp4, etc.).

PR #4571, opened 2017-08-10
Handle ActivityPub follows correctly

Gargron:
ActivityPub follows are follow-requests. Always require an Accept.
If account is not locked, auto-accept.

### DIFF
--- a/app/lib/activitypub/activity/follow.rb
+++ b/app/lib/activitypub/activity/follow.rb
@@ -12,7 +12,7 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
       NotifyService.new.call(target_account, follow_request)
     else
       AuthorizeFollowService.new.call(@account, target_account)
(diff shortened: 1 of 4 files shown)


app/lib/activitypub/activity/follow.rb
```diff
@@ -6,7 +6,13 @@ def perform
 
     return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id'])
 
-    follow = @account.follow!(target_account)
-    NotifyService.new.call(target_account, follow)
+    follow_request = FollowRequest.create!(account: @account, target_account: target_account)
+
+    if target_account.locked?
```
nightpool: does activitypub have a concept of a locked account? I'm not sure I follow the flow of this logic.

Gargron: No but we do.

If we get a Follow and we are locked, we create a follow request and notify the user about the follow request. If we are not locked, we auto-accept the follow and notify the user about the follow.

nightpool (approved):

PR #4579, opened 2017-08-11
Don't wipe compose box content on mention or reply

MightyPork:
This is an effort to preserve text and settings of the compose box when the reply or mention button is pressed. Handles are prepended if missing.

Canceling a reply also no longer wipes the compose box.

Currently, the content is wiped, possibly destroying what text user might have had there before, perhaps preparing to reply. 

One situation to illustrate the problem - Alice makes a typo and deletes her post, then re-sends it with corrections. Meanwhile, Bob was typing a reply, but the old post vanished, and the new one appeared. Bob clicks reply on the corrected post, but his compose box is wiped in the process, deleting his reply.

If there are still issues to be addressed, let me know.
If this has no chance of being merged, also please tell me. I'll re-submit it to the glitch-soc fork.


app/javascript/mastodon/reducers/compose.js
```diff
@@ -141,6 +141,23 @@ const privacyPreference = (a, b) => {
   }
 };
 
+function addMentions(state, mentionsStr) {
```
nightpool: There's no reason to take all of "state" here if you're only using `'text'`

MightyPork:
I don't mean to sound impatient, but is there still something wrong?

app/javascript/mastodon/reducers/compose.js
```diff
@@ -141,6 +141,26 @@ const privacyPreference = (a, b) => {
...
+/**
+ * Add missing mentions to a status draft from the compose box
+ *
+ * @param text - the status draft
+ * @param mentionsStr - space-separated user handles to mention
+ * @return updated text
+ */
+function addMentions(text, mentionsStr) {
+  let missing = [];
+  for (const m of mentionsStr.split(' ')) {
```
akihikodaki: Use `Array.prototype.filter`.

akihikodaki (changes requested):
Sorry for keeping you waiting. I have found a point which can be improved. I hope your change would land soon.

Gargron:
Resolved by #8893

PR #4582, opened 2017-08-11
Add configuration to disable private status federation over PuSH

Gargron:
Follow-up to #4566

Federation of private statuses over Pubsubhubbub was enabled in #2111, with the implementation of ActivityPub with native audience targeting there is no more need to use the less reliable protocol which is implemented by platforms that do not support status privacy.

This would be a **breaking change** in that Mastodon instances that do not upgrade to ActivityPub would stop receiving private statuses. This *might* warrant an overall semver major version bump, once this configuration option is toggled.

### DIFF
--- a/app/services/process_mentions_service.rb
+++ b/app/services/process_mentions_service.rb
@@ -39,7 +39,7 @@ class ProcessMentionsService < BaseService
 
     if mentioned_account.local?
       NotifyService.new.call(mentioned_account, mention)
-    elsif mentioned_account.ostatus?
+    elsif mentioned_account.ostatus? && (Rails.configuration.x.use_ostatus_privacy || !status.stream_entry.hidden?)
       NotificationWorker.perform_async(stream_entry_to_xml(status.stream_entry), status.account_id, mentioned_account.id)
     elsif mentioned_account.activitypub?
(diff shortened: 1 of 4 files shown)


mkljczk (approved):

ykzts (approved):

nightpool:
I talked about this a little in dev chat, but I think the right thing to do here is create a ACTIVITYPUB_PRIMARY variable that gates the things we consider breaking changes. that gives us the ability to start the RC process now, and follow it up with some good low-hanging fruit, like improving the density of federation using forwarding, adding more metadata, etc.

And then a couple weeks out we can make a 2.0 release that basically just flips the switch. That gives us lots of time to improve the user experience, work out any bugs that people find, make sure we're 100% confident and we're also not breaking ostatus compat, stuff like that

The big benefit of this approach is that gets this stuff out into the world and gets people to start using it as soon as possible, which I think is really really important.

beatrix-bitrot:
I really like nightpool's ACTIVITYPUB_PRIMARY suggestion. Can we please do
this?

akihikodaki (approved):

nightpool (approved):
I'm +1 on these changes as is, but we need to figure out a staged rollout story for admins.

PR #4585, opened 2017-08-11
Add serializing/unserializing of "locked" actor attribute

Gargron:


nightpool:
what's the rationale for using `_:locked`? the JSON-LD standard only ever uses `_:` for the creation of `@id`s in the presence of cyclic data, never as keys.

Gargron:
@nightpool According to @puckipedia `_:locked` and `locked` are equivalent because it's not a registered property of any kind. For me personally it's better to keep track of which properties are not part of ActivityPub.

nightpool:
mmm. okay. I think I can see how this works. I'm down for whatever, but it seems more explicit to me to use something like `https://joinmastodon.org/schema/locked` as your key to avoid conflicts/be more JSON-LD

app/lib/activitypub/case_transform.rb
```diff
@@ -0,0 +1,24 @@
...
+      @camel_lower_cache ||= {}
+    end
+
+    def camel_lower(value)
+      case value
+      when Array then value.map { |item| camel_lower(item) }
+      when Hash then value.deep_transform_keys! { |key| camel_lower(key) }
+      when Symbol then camel_lower(value.to_s).to_sym
+      when String
+        camel_lower_cache[value] ||= if value.start_with?('_:')
```
nightpool: this should probably be general enough to handle other compact IRIs, like `foaf:name` (the first example from the JSON-LD spec, chosen at random)

nightpool (changes requested):
I'm fine with punting on namespacing. LGTM w/ camel_lower generalized just a *tiny* bit.

app/lib/activitypub/case_transform.rb
nightpool: maybe something like `.split(':').map {|i| i.underscore.camelize(:lower)}.join(':')`?

nightpool (approved):
Nevermind, looks like underscore.camelize(:lower) already does the right thing and I was confused.

PR #4586, opened 2017-08-12
Add alternate links to ActivityPub resources from HTML/HEAD variants

Gargron:


app/controllers/concerns/account_controller_concern.rb
```diff
@@ -41,6 +42,13 @@ def atom_account_url_link
     ]
   end
 
+  def actor_url_link
+    [
+      ActivityPub::TagManager.instance.uri_for(@account),
+      [%w(rel alternate), %w(type application/activity+json)],
```
nightpool: should be `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` throughout.

nightpool (changes requested):
LGTM w/ `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` instead of `application/activity+json`

app/controllers/concerns/account_controller_concern.rb
Gargron: See: <https://www.w3.org/TR/activitystreams-core/#media-type>

>This specification registers the application/activity+json MIME Media Type specifically for identifying documents conforming to the Activity Streams 2.0 format. 
>
> Note that while the Activity Streams 2.0 format uses JSON-LD conventions, there are a number of constraints and additional requirements for Activity Streams 2.0 implementations that justify the use of a specific media type.
>
>Because Activity Streams 2.0 can be considered a restricted profile of JSON-LD, Implementations SHOULD consider the `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` media type as being equivalent to `application/activity+json`.

nightpool:
Okay, we should maybe ask someone else, but activitypub specifically says
that you MUST respond to the ld+json profile one, and only MAY respond to
the other one.

Personally, I would also prefer to use activity+json, I just want to make
sure we're respecting the spec

PR #4589, opened 2017-08-12
Hook up URL-based resource look-up to ActivityPub

Gargron:


app/services/fetch_remote_account_service.rb
```diff
@@ -5,14 +5,19 @@ class FetchRemoteAccountService < BaseService
...
-      atom_url = url
-      body     = prefetched_body
+      resource_url = url
+      body         = prefetched_body
+      protocol     = :ostatus
     end
 
-    return nil if atom_url.nil?
-    process_atom(atom_url, body)
+    case protocol
```
ykzts: `case protocol ` -> `case protocol` (remove trailing whitespace)

app/services/fetch_remote_status_service.rb
```diff
@@ -5,14 +5,19 @@ class FetchRemoteStatusService < BaseService
...
-      atom_url = url
-      body     = prefetched_body
+      resource_url = url
+      body         = prefetched_body
+      protocol     = :ostatus
     end
 
-    return nil if atom_url.nil?
-    process_atom(atom_url, body)
+    case protocol
```
ykzts: `case protocol ` -> `case protocol` (remove trailing whitespace)

app/services/fetch_atom_service.rb
```diff
@@ -4,18 +4,10 @@ class FetchAtomService < BaseService
...
-
-    response = Request.new(:get, url).perform if response.code == 405
-
-    Rails.logger.debug "Remote status GET request returned code #{response.code}"
-
-    return nil if response.code != 200
-    return [url, fetch(url)] if response.mime_type == 'application/atom+xml'
-    return process_headers(url, response) if response['Link'].present?
-    process_html(fetch(url))
+    perform_request
```
Gargron: So what is happening here:

|Before|After|
|------|-----|
|First, HEAD request to inspect headers, check mime type, if it fits - do GET request, if not, check Link header, GET from it, if there is no Link header, GET the page and try to find a link in the HTML|Do a GET request, check mime type - if it fits, fine, if not, check headers, if no headers, check HTML. If something is in headers or HTML, do another GET to that|

In the best case that means HEAD + GET have become a single GET, in the worst case, HEAD + GET + GET has become GET + GET. Only in case of 404 it has become GET instead of HEAD, which miiiiight be a heavier request, but unless the body is actually read, maybe it doesn't matter.

ykzts (approved):

app/services/fetch_atom_service.rb
```diff
@@ -26,27 +18,55 @@ def call(url)
 
   private
 
-  def process_html(body)
-    Rails.logger.debug 'Processing HTML'
+  def perform_request
+    @response = Request.new(:get, @url).perform
```
nightpool: doesn't this need to have `Accept: activity+json, application/ld+json; profile="whatever"` to actually get an activitypub resource?

Gargron: You're actually right, thanks.

PR #4592, opened 2017-08-14
Set correct content-type for ActivityPub JSON

Gargron:


nightpool (approved):
looks like we probably can't do anything about the profile.

PR #4599, opened 2017-08-14
Add support for searching AP users

ykzts:


app/services/fetch_remote_resource_service.rb
```diff
@@ -34,6 +34,25 @@ def body
...
+    case xml_root
+    when 'feed'
+      'Person'
+    when 'entry'
+      'Note'
+    end
+  end
+
+  def json_data
+    @_json_data ||= Oj.load(body, mode: :strict)
```
Gargron:
```ruby
include JsonLdHelper

@_json_data ||= body_to_json(body)
```

ykzts: thank you for review! ef768ec

Gargron (approved):

PR #4601, opened 2017-08-14
Fix ActivityPub follow interaction and add more specs

unarist:
Accept activity will be sent from followee to follower with Follow object. We need to use Follow object's `actor` property to get follower, instead of `object` property (=followee).


Gargron (approved):

unarist:
Note: If you have tried to follow before this patch, you would have pending follow requests. These requests will not be removed automatically and also you can't cancel request from WebUI or API. In this case, you need to handle those requests from Rails console or delete from database directly.

PR #4604, opened 2017-08-15
Fix require_user! behavior when not logged in

abcang:
If you run require_user! when you are not logged in, set_user_activity raises an exception.

```
> undefined method `update_tracked_fields!' for nil:NilClass

app/controllers/concerns/user_tracking_concern.rb, line 17
----------------------------------------------------------

``` ruby
   12   
   13     private
   14   
   15     def set_user_activity
   16       # Mark as signed-in today
>  17       current_user.update_tracked_fields!(request)
   18   
   19       # Regenerate feed if needed
   20       regenerate_feed! if user_needs_feed_update?
   21     end
   22   
```

Therefore, I changed to execute set_user_activity only when current_user is not nil.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -62,10 +62,11 @@ class Api::BaseController < ApplicationController
   end
 
   def require_user!
-    current_resource_owner
-    set_user_activity
-  rescue ActiveRecord::RecordNotFound
-    render json: { error: 'This method requires an authenticated user' }, status: 422
+    if current_user
+      set_user_activity
+    else
+      render json: { error: 'This method requires an authenticated user' }, status: 422
+    end
   end
 
   def render_empty


Gargron (approved):

nightpool:
Why the behavior change to using `current_user` over `current_resource_owner` here?

abcang:
In `current_resource_owner`, if doorkeeper_token is falsy, it returns nil without raising an exception. So, `set_user_activity` may be executed when you are not logged in. `current_user` in `Api::BaseController` is executing `current_resource_owner` and returns nil if not logged in.

However, I noticed that there is a possibility of executing `current_user` provided by devise. This may be bad.

abcang:
It seems that the behavior to verify users authenticated with doorkeeper or devise is correct. In `Api::Web::SettingsController`, doorkeeper is not used, but `require_user!` is executed. So if you change to the following behavior, it will not work with the API controller that does not use doorkeeper.

```rb
  def current_resource_owner
    @current_user ||= User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
    nil
  end

  def current_user
    current_resource_owner || super
  end

  def require_user!
    if current_resource_owner
      set_user_activity
    else
      render json: { error: 'This method requires an authenticated user' }, status: 422
    end
  end
```

PR #4608, opened 2017-08-16
Redesign public profiles

Gargron:
Functional changes:

- Public profiles now display only public/unlisted toots, regardless of access control
- By default, replies are filtered out
- "With replies" option to include them again
- "Media" option to display only toots with media attachments
- "Posts" renamed to "toots" on public profiles

Visual changes:

image

Mobile view:

screen shot 2017-08-16 at 16 43 12
screen shot 2017-08-16 at 16 43 18

Followers/following cards:

image

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -7,8 +7,14 @@ class AccountsController < ApplicationController
   def show
     respond_to do |format|
       format.html do
-        @statuses = @account.statuses.permitted_for(@account, current_account).paginate_by_max_id(20, params[:max_id], params[:since_id])
+        if current_account && @account.blocking?(current_account)
+          @statuses = []
+          return
+        end
+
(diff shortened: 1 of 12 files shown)


m4sk1n:
How does it look on mobile devices?

PaulFree14:
yes! I really like to have the "toots" and "toots with replies".

mabkenar:
Now that you are at it, would you please make sure #2588 is taken care of? Thanks for the great work you do!

ykzts (approved):
looks good! I like this.

ZiiX:
There have been requests to allow a user to set their default 'landing' view of others. For example, only show their toots (without boosts OR replies) as well as other wanted an option to default to land with their media toots only. SO that their mastodon profile is essentially all their forward-facing content.
This looks amazing though. I'm more of a fan of a centered background + profile but this is growing on me.

PR #4609, opened 2017-08-16
l10n Update OC - ActivityPub (#4600)

Quent-in:
Update: new string + more translations for time in words


ykzts (approved):

ykzts:
Only @Quent-in translates this locale.

PR #4622, opened 2017-08-17
Show what protocol is used for accounts in admin/accounts#index

nullkal:
What this PR looks like: 

This PR adds "protocol" colomn to account list for admin. In addition to it, I added a helper to show protocol names like "OStatus", or "ActivityPub", not "ostatus", nor "activitypub".

### DIFF
--- a/app/helpers/account_helper.rb
+++ b/app/helpers/account_helper.rb
@@ -1,0 +1,12 @@
+module AccountHelper
+  def protocol_for_display(protocol)
+    case protocol
+    when 'activitypub'
+      'ActivityPub'
+    when 'ostatus'
+      'OStatus'
+    else
+      protocol
+    end
+  end
+end
--- a/app/helpers/account_helper.rb
+++ b/app/helpers/account_helper.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module AccountHelper
   def protocol_for_display(protocol)
     case protocol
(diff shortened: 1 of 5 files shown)


ykzts (approved):
good😄

app/helpers/account_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module AccountHelper
+  def protocol_for_display(protocol)
+    case protocol
+    when 'activitypub'
+      'ActivityPub'
+    when 'ostatus'
+      'OStatus'
```
nightpool: shouldn't this just be a matter of using the correct inflection?

app/views/admin/accounts/_account.html.haml
```diff
@@ -4,6 +4,9 @@
   %td.domain
     - unless account.local?
       = link_to account.domain, admin_accounts_path(by_domain: account.domain)
+  %td.protocol
+    - unless account.local?
+      = protocol_for_display(account.protocol)
```
nightpool: should just be acccount.protocol.camel_case

nullkal: How about "OStatus"?  It's written as ostatus in code, but we want it to be transformed to OStatus. Replacing all occurence of `ostatus` to `o_status` is nonsense and impossible.

nightpool: @nullkal we already have the inflections defined for this....

nightpool: https://github.com/tootsuite/mastodon/blob/master/config/initializers/inflections.rb

PR #4624, opened 2017-08-17
use existing inflections instead of custom helper

nightpool:
in #4622, we added an entirely new helper for something rails already does for us.

image


Gargron (dismissed):

nullkal:
`ActiveSupport::Inflector#humanize` translates `ostatus` into `Ostatus` while `OStatus` is more appropriate. This is the reason why I created an new helper method instead of using it.

nightpool:
one sec, this doesn't work because the helper isn't included, for some reason.

Gargron:
@nullkal We define some custom inflections in config/initializers/inflections.rb - I would assume the humanize method would use those definitions..?

nullkal:
I understood. Thanks!

Ref: https://github.com/tootsuite/mastodon/blob/master/config/initializers/inflections.rb#L16-L17

nightpool:
fixed.

Gargron:
CI build is broken due to some unrelated thing, mind submitting an empty commit to bump it?

nightpool:
@Gargron that's the /push build. I think travis got confused because I accidentally pushed this to tootsuite/mastodon instead of nightpool/mastodon. the /pr build is correct.

PR #4630, opened 2017-08-18
Support more variations of ActivityPub keyId in signature

### LINKED ISSUE #4606: ActivityPub: Key signature verifying only works when key ID is a fragment of the user
... `account_from_key_id` just runs the account resolver on the key's ID, which means that it assumes the key ID is a fragment of the user ID. https://github.com/tootsuite/mastodon/blob/fdea173237cfcd3a6b36f6ebccb0cb1a21cf9294/app/controllers/concerns/signature_verification.rb#L97-L103

This seems like a bug, because the outgoing requests seem to do the right thing:
https://github.com/tootsuite/mastodon/blob/fdea173237cfcd3a6b36f6ebccb0cb1a21cf9294/app/lib/request.rb#L72-L79

* * * *

Gargron:
Fix #4606

- Tries to avoid performing HTTP request if the keyId is an actor URI
- Likewise if the URI is a fragment URI on top of actor URI
- Resolves public key, returns owner if the owner links back to the key

What this does **not** do: We do not differentiate between keys used for the signature. In theory an account can have multiple public keys, and the keyId could refer to any one of them, and *that* key should be used to verify the signature. However, we don't have the storage structure for this - we cache one public key in the accounts table (design legacy from OStatus). So it is assumed that even if keyId links directly to a key, it is still the first/primary key of the account.

### DIFF
--- a/app/controllers/concerns/signature_verification.rb
+++ b/app/controllers/concerns/signature_verification.rb
@@ -98,7 +98,9 @@ module SignatureVerification
     if key_id.start_with?('acct:')
       ResolveRemoteAccountService.new.call(key_id.gsub(/\Aacct:/, ''))
     elsif !ActivityPub::TagManager.instance.local_uri?(key_id)
-      ActivityPub::FetchRemoteAccountService.new.call(key_id)
+      account   = ActivityPub::TagManager.instance.uri_to_resource(key_id, Account)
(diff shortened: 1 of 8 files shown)


Gargron:
cc @puckipedia

PR #4632, opened 2017-08-18
Improve ActivityPub/OStatus compatibility

Gargron:
*Note: OStatus URIs are invalid for ActivityPub. But we have them for
as long as we want to keep old OStatus-sourced content and as long as
we remain OStatus-compatible.*

- In Announce handling, if object URI is not a URL, fallback to object URL
- Do not use specialized ThreadResolveWorker, rely on generalized handling (see #4589)
- When serializing notes, if parent's URI is not a URL, use parent's URL


Gargron:
cc @unarist

unarist:
looks good, but is it OK to have `tag:...` as AP object id, in point of ActivityPub conformance?

Gargron:
@unarist It's not okay, but I don't see us having a choice. I said on IRC:

> this is probably the simplest way for me to deal with the OStatus remnants  
> i.e. okay, we're gonna publish invalid "id" attributes, which from a pure-AP implementation's perspective is junk, but at least it should work from Mastodon to Mastodon this way, and such junk "id"s will naturally disappear over time as OStatus becomes less and less used

Gargron:
We are facing a difficult and unique problem of upgrading a network protocol. Mastodon was not designed to be multi-protocol. Other multi-protocol software like Friendica/Hubzilla works in a slightly different way which is not fit for our UX: While Friendica users may *subscribe* to both Diaspora and OStatus users and interact with their posts, their posts are not *native* to each other, e.g. you cannot *actually* share a Diaspora post into OStatus; instead, Friendica creates a copy/wrapper of the contents, authored by whoever shared it, rather than original author (at least, to my understanding and in my experience).

This means that consumers of such shares cannot verify the integrity of the shared message, the proxy/bridge could be completely arbitrary. In Mastodon, on the other hand, we put a lot of value on verifying a message belongs to who it says it belongs to. We try to avoid malformed messages, etc. So this avenue is closed for us.

unarist (approved):

PR #4634, opened 2017-08-18
Always add nsfw tag when sensitive toot

### LINKED ISSUE #4610: Retoot/boosted status updates have strange tags.
Here is a status that is not tagged: [web](https://kitty.town/@KevinCarson1/1099102), [JSON](https://kitty.town/api/v1/statuses/1099102). Here is the same status, on a different server: [JSON](https://octodon.social/api/v1/statuses/5488082) – but now it is tagged NSFW! When you look at it in the stream of the user doing the boosting, you don't see this tag:

image

Now, from the point of view of a client looking at the toots on octodon.social, does this count as a toot tagged NSFW even though you can't see it in the web UI and even though the original toot is not tagged NSFW?

I think this is a bug and that the NSFW tag looks like random data.

I guess the alternative is that remove toots with CW are automatically tagged NSFW on remote servers and this just needs to be [better documented](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#status)?

* * * *

ykzts:
Close #4610

Mastodon implicitly adds an NSFW tag to a sensitive toot. This is only for Federates case (e90fcb46e3e5bc4144d8777c2267bec638e327a0). I think that similar behaviors are better for local instances as well.

### DIFF
--- a/app/lib/ostatus/atom_serializer.rb
+++ b/app/lib/ostatus/atom_serializer.rb
@@ -358,8 +358,6 @@ class OStatus::AtomSerializer
       append_element(entry, 'category', nil, term: tag.name)
     end
 
-    append_element(entry, 'category', nil, term: 'nsfw') if status.sensitive?
-
     status.media_attachments.each do |media|
       append_element(entry, 'link', nil, rel: :enclosure, type: media.file_content_type, length: media.file_file_size, href: full_asset_url(media.file.url(:original, false)))
     end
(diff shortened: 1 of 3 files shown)


akihikodaki (approved):
This is not a blocking matter, but why just remove the old spec and don't add a new spec? A pending example is sufficient to keep the missing spec.

Gargron:
The "nsfw" category has been a way to serialize "sensitive" boolean into Atom. It was *not* a good idea, it is a hack that mixes user-space data and system-space data, however @lambadalambda's Pleroma was already doing something with `#nsfw` hashtags so it was picked for maximum compatibility.

This behaviour has no reason to persist into ActivityPub, and has no reason to exist locally.

ykzts:
@Gargron Hm. I see. It's certainly reasonable.

PR #4639, opened 2017-08-19
Handle duplicate ActivityPub activities

Gargron:
Only perform side-effects when record processed for the first time

### DIFF
--- a/app/lib/activitypub/activity/follow.rb
+++ b/app/lib/activitypub/activity/follow.rb
@@ -4,7 +4,13 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
   def perform
     target_account = account_from_uri(object_uri)
 
-    return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.following?(target_account) || @account.requested?(target_account)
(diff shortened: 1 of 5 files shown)


app/lib/activitypub/activity/follow.rb
```diff
@@ -4,7 +4,7 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
   def perform
     target_account = account_from_uri(object_uri)
 
-    return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id'])
+    return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.following?(target_account) || @account.requested?(target_account)
```
unarist: Ignoring it seems good if we already have a "FollowRequest", but I'm not sure if we already have a "Follow".

When we receive Follow activity, the sender would be in "Requesting follow" phase and waits Accept / Reject activity. Ignoring request may keep them in requesting phase. Shouldn't we send Accept activity if we already accepted following?

Gargron: Hmm, okay. I agree

unarist (approved):

PR #4641, opened 2017-08-19
Improve about ja translation standalone.public_title

takayamaki:


clworld:
Yes, standalone.public_title isn't translated literally.
We translated "A look inside..." as 「連合タイムライン」("Federated timeline") because what shown on /about page(FTL) isn't inside and literal translation 「中を覗いてみる…」("Peeping inside...") feels not very good.
but 「連合タイムライン」("Federated timeline") causes problem when showing local timeline on /about page...
It's a difficult problem to translate it.
Is 「今こんな話をしています」("We are talking about like this.") best translation?

takayamaki:
Yes, this was very difficult translation.

There was such a proposal in the our (some peoples of imastodon.net) discussion.
 - 「ちょっと覗いてみましょう…」(literal translation)
 - 「今こんな話題が流れています」
 - 「今こんな話をしています」
 - 「中の様子」
and more...

I felt that our literal translation 「ちょっと覗いてみましょう…」 is not good.

Finally, we (imastodon.net) decide to use 「今こんな話をしています」.

app/javascript/mastodon/locales/ja.json
```diff
@@ -159,7 +159,7 @@
   "report.target": "{target} を通報する",
   "search.placeholder": "検索",
   "search_results.total": "{count, number}件の結果",
-  "standalone.public_title": "連合タイムライン",
+  "standalone.public_title": "今こんな話をしています",
```
clworld: I was thinking about this translation 1 day.
but I can't find translation better than this.
and this translation is better than current one.
(new translation does not depend on what timeline(LTL/FTL) is shown)
So, I approve this.

clworld (approved):
good enough.

ykzts (approved):
I agree too.

PR #4642, opened 2017-08-19
Make PreviewCard records reuseable between statuses

Gargron:
**Upgrade path**: Moves all previous preview cards into `deprecated_preview_cards` table, leaving the deletion for a rake task. Use `rake mastodon:maintenance:remove_deprecated_preview_cards` to clear it out and finally drop it.

**Benefits**:

- Same URL is not re-fetched again once we have a preview, previously it would have been refetched and stored for every toot the URL appears in individually
- Previews where the thumbnail is wider than it is tall are displayed in a slightly different layout (#4648)


app/services/fetch_link_card_service.rb
```diff
@@ -4,29 +4,40 @@ class FetchLinkCardService < BaseService
...
-    url  = url.to_s
-    card = PreviewCard.where(status: status).first_or_initialize(status: status, url: url)
-    res  = Request.new(:head, url).perform
+    @url  = @url.to_s
+    @card = PreviewCard.find_by(url: @url)
 
-    return if res.code != 200 || res.mime_type != 'text/html'
-
-    attempt_opengraph(card, url) unless attempt_oembed(card, url)
+    process_url if @card.nil?
```
nightpool: Feels like there's a race condition here where we try and make two preview cards at the same time and error on the slower one.

Gargron: I didn't put any uniqueness constraints on the url so it's not a big deal.

nightpool: yeah but there's a uniqueness validation

Gargron: I didn't mean to keep that in.

But I think I had a change of heart today morning. The `PreviewCard.find_by(url: @url)` would take an increasing amount of time unless an index is created. So I think I am going to add an index, and it may be unique.

Gargron:
@nightpool Changed!

@ykzts @unarist @abcang @akihikodaki I would like to ask if you think it would be problematic to truncate the preview_cards table or if it would be acceptable. Here is my reasoning:

- Preview cards on old statuses are rarely ever accessed, I think
- Preview cards load asynchronously, so people would probably not even notice if some are missing
- By truncating the table, we are getting rid of a lot of duplicates very quickly (example from nightpool: 132922 count(distinct url) vs 251582 total)
- If we had to copy over the old status_id values from preview_cards table into the join table, the migration would take up a long time (and lock up the db), which feels like a waste for something so secondary as link previews
- If I made it into a rake data migration task, it would still take a long while and make the upgrade for the release very annoying for a lot of people

So I think it makes sense.

ykzts:
looks good to me.I am concerned that even if the page is updated, the Preview Card remains old (`PreviewCard#refetch`?).

Gargron:
@ykzts I think that OEmbed/OpenGraph contents are usually immutable (e.g. tweets, youtube videos, articles); even facebook caches OpenGraph data, though they have a developer page for clearing that cache. So I don't think this is a problem.

ykzts:
@Gargron Certainly, but may be better to be able to refetch from the admin UI (no need to do with this PR).

ykzts (approved):

unarist:
I think truncating should also remove downloaded images, was it done in this?

app/services/fetch_link_card_service.rb
```diff
@@ -4,29 +4,45 @@ class FetchLinkCardService < BaseService
...
-    return if url.nil?
+    return if @url.nil? || @status.preview_cards.any?
 
-    url  = url.to_s
-    card = PreviewCard.where(status: status).first_or_initialize(status: status, url: url)
-    res  = Request.new(:head, url).perform
+    @url = @url.to_s
 
-    return if res.code != 200 || res.mime_type != 'text/html'
+    RedisLock.acquire(lock_options) do |lock|
```
nightpool: why not just use a postgres transaction?

akihikodaki: Maybe we could have a comment before the line?

akihikodaki:
Creating another table could be a solution, and I think it is good enough though it is a really simple idea.
Let's have a migration file to rename the existing table to `old_preview_cards` and another migration file to destroy the old files and table.

Gargron:
@unarist @akihikodaki I have changed the migration:

- Disabled transaction around TruncatePreviewCards migration
- Iterate over all preview cards, delete files (but do not bother updating records in db, to be faster)
- After that, truncate the table
- Then, in the next migration, remove old column and add new index

If I understand everything correctly, first migration should have no impact on ability of users to use the site during the migration. I'm not even sure if disabling the transaction is necessary, since users are not writing to that table directly anyway, it happens in background workers which are allowed to fail.

akihikodaki:
`TruncatePreviewCards` _must_ have a transaction, or `TRUNCATE` will delete records which are added after destroying files.
If you choose the strategy, the two migrations should be unified and the unified one should have a transaction wrapping `TRUNCATE` and column modifications, and the transaction should make sure that files created after the first destruction again.

By the way, I am not really for the strategy. Removed files would respond to the user agent with 404 during the migration. And, this is my intuition anyway, but the assumption that PreviewCard is allowed to fail sounds dangerous.
I suggest to have the renaming strategy which I described or another safer strategy.

Gargron:
@akihikodaki After renaming, do we just leave the old files around? Or when/how do we clean them up?

akihikodaki:
Stop Mastodon just before renaming, and restart it immediately once renaming has been done. The duration would be the only downtime.
Destroy the old files and table when convenient. That would not prevent Mastodon from keep running.

Gargron:
@akihikodaki I have updated (see new PR description)

akihikodaki (approved):

PR #4645, opened 2017-08-20
Refactoring streaming connections

abcang:
I extracted the processing of the streaming connection in the component and moved it to the action.


abcang:
In the current implementation, similar processing is done in each component, so I shared the processing. Also, I think that API requests should be executed by action, not by component.

Gargron (approved):

PR #4653, opened 2017-08-21
Do not try to re-subscribe to unsubscribed accounts

Gargron:


unarist:
Okay, looks good.

Now we need to subscribe manually to recover subscription (e.g. in case of remote server had been down for half a month). I think it is reasonable, but we can consider admin tools for this at some time. For example: filter by subscription status, automatic unsubscribing logs, etc.

app/models/account.rb
```diff
@@ -134,11 +134,11 @@ def subscribed?
   end
 
   def keypair
-    OpenSSL::PKey::RSA.new(private_key || public_key)
+    @keypair ||= OpenSSL::PKey::RSA.new(private_key || public_key)
   end
 
   def subscription(webhook_url)
-    OStatus2::Subscription.new(remote_url, secret: secret, lease_seconds: 30.days.seconds, webhook: webhook_url, hub: hub_url)
+    @subscription ||= OStatus2::Subscription.new(remote_url, secret: secret, webhook: webhook_url, hub: hub_url)
```
unarist: Is `lease_seconds` no longer needed?

Gargron: I removed it because it would confuse readers. The subscribe/unsubscribe methods of `OStatus2::Subscription` are no longer used anywhere. SubscribeService defines its own lease_seconds.

unarist (approved):

PR #4654, opened 2017-08-21
Periodically remove expired PuSH subscribers

Gargron:


app/models/subscription.rb
```diff
@@ -26,6 +26,7 @@ class Subscription < ApplicationRecord
 
   scope :confirmed, -> { where(confirmed: true) }
   scope :future_expiration, -> { where(arel_table[:expires_at].gt(Time.now.utc)) }
+  scope :expired, -> { where(arel_table[:expires_at].lt(Time.now.utc)) }
```
nightpool: doing it immediately seems maybe bad depending on when people set their renewal times to be? Something like expires_at.lt(Time.now.utc - 7.days) or 14 days is probably Good Enough and allows for a little bit of leniency.

Gargron: @nightpool This matters not. Renewing is the same as subscribing from scratch. The secret is fresh, etc. There is virtually no reason why you'd want to keep a particular record beyond its expiration date. Plus the removal worker runs only once a day.

nightpool (approved):
okay, LGTM then

PR #4657, opened 2017-08-21
Add GitHub Url library

lnanase:
Mastodon Github repository URL replace to dynamic library file.
All other instance need to do is change library file.

(Similar pull request #4643. Sorry, @zunda .)

Thank you.

### DIFF
--- a/lib/mastodon/github.rb
+++ b/lib/mastodon/github.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Mastodon
+  module Github
+    module_function
+
+    def host
+      'https://github.com'
+    end
+
+    def author
+      'tootsuite'
+    end
+
+    def repository
+      'mastodon'
+    end
+
+    def link_title
+      [author, repository].join('/')
+    end
+
+    def to_s
+      [host, author, repository].join('/')
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


zunda:
This looks better and more consistent than #4643 for me! I wonder if it is possible to insert a shim to represent tag/branch/commit hash in `Mastodon::Github` so that instances following the master branch can specify tag/branch/commit hash when appropriate.

nightpool:
-1 to tying this to github, tentative -1 to changing the onboarding modal since it's talking about mastodon the project and not just the source code/any instance customizations, +1 for everything else.

zunda:
Ah gothca. It is too consistent to change URL on onboarding modal.

lib/mastodon/github.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def repository
+      'mastodon'
+    end
+
+    def link_title
+      [author, repository].join('/')
+    end
+
+    def to_s
```
akihikodaki: Should be named `link_url` or something like that. There is no actual reason that the link title is not `to_s` and the URL is not `link_url`. Name them in a consistent rule then.

akihikodaki (changes requested):
It is a good idea to extract brand-related information from the code base, but this pull request actually spreads the information that Mastodon is hosted on GitHub over the code base. Please be aware that others, and the upstream in future may use other hosting services.

Gargron:
I chose #4643 because it's less brand attached and more general (sorry)

akihikodaki:
#4643 does not include JavaScript side while this pull request does. The addition of JavaScript side would be appreciated.
If you are willing to implement that, I suggests to have a YAML shared by Rails and webpack.
See `config/webpacker.yml` to load a YAML in webpack. [webpack's DefinePlugin](https://webpack.js.org/plugins/define-plugin/) would allow to expose it to the bundled scripts.
For Rails, you may let `Mastodon::Version` extend with `RailsSettings::Default` class.

PR #4658, opened 2017-08-22
authorize follow requests after unlocking account

masarakki:
1. @A is *locked* account.
2. @B want to follow @A and send follow-request.
3. @A changes his profile to *unlocked*.
4. But record of FollowRequest exists, so @B can't follow @A.
5. Because @A is unlocked, so @A can't see tab of *follow requests*.

I think all follow requests should be accepted automatically when account unlocked.

### DIFF
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -14,7 +14,7 @@ class Settings::ProfilesController < ApplicationController
   def show; end
 
   def update
-    if @account.update(account_params)
+    if UpdateAccountService.new.call(@account, account_params)
       ActivityPub::UpdateDistributionWorker.perform_async(@account.id)
       redirect_to settings_profile_path, notice: I18n.t('generic.changes_saved_msg')
     else
--- a/app/services/update_account_service.rb
+++ b/app/services/update_account_service.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class UpdateAccountService < BaseService
+  def call(account, params, raise_error: false)
+    was_locked = account.locked
+    update_method = raise_error ? :update! : :update
+    account.send(update_method, params).tap do |ret|
+      next unless ret
+      authorize_all_follow_requests(account) if was_locked && !account.locked
+    end
+  end
+
+  private
+
+  def authorize_all_follow_requests(account)
+    follow_requests = FollowRequest.where(target_account: account)
+    AuthorizeFollowWorker.push_bulk(follow_requests) do |req|
+      [req.account_id, req.target_account_id]
+    end
+  end
+end
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -12,7 +12,7 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
 
   def update
     @account = current_account
-    @account.update!(account_params)
+    UpdateAccountService.new.call(@account, account_params, raise_error: true)
     ActivityPub::UpdateDistributionWorker.perform_async(@account.id)
     render json: @account, serializer: REST::CredentialAccountSerializer
   end
--- a/app/workers/authorize_follow_worker.rb
+++ b/app/workers/authorize_follow_worker.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AuthorizeFollowWorker
+  include Sidekiq::Worker
+
+  def perform(source_account_id, target_account_id)
+    source_account = Account.find(source_account_id)
+    target_account = Account.find(target_account_id)
+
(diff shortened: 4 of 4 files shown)


app/services/update_account_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class UpdateAccountService < BaseService
+  def call(account, params)
+    before_locked = account.locked
+    account.update(params).tap do |ret|
+      return unless ret
+      AuthorizeAllFollowsWorker.perform_async(account.id) if before_locked && !account.locked
+      # authorize_all_follow_requests(account) if before_locked && !account.locked
```
masarakki: I can't decide using worker or not.

app/services/update_account_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class UpdateAccountService < BaseService
+  def call(account, params)
+    before_locked = account.locked
+    account.update(params).tap do |ret|
+      next unless ret
+      AuthorizeAllFollowsWorker.perform_async(account.id) if before_locked && !account.locked
+      # authorize_all_follow_requests(account) if before_locked && !account.locked
+    end
```
masarakki: I can't decide using worker or not.

akihikodaki: I think it is fine to do synchronously because almost all operations in `AuthorizeFollowService` are done by workers.

app/workers/authorize_all_follows_worker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AuthorizeAllFollowsWorker
+  include Sidekiq::Worker
+
+  def perform(account_id)
+    account = Account.find(account_id)
+    FollowRequest.where(target_account: account).each  do |req|
+      AuthorizeFollowWorker.perform_async(req.account_id, req.target_account_id)
```
akihikodaki: Having one query instead of lots of `Account.find` in `AuthorizeFollowWorker` would be better. The latency is not really important for this operation, and the overall load of workers would take precedence.
```Ruby
account = Account.find(account_id)
FollowRequest.where(target_account: account).includes(:account)
```
…where `.includes(:account)` is optional. I think it is fine to drop it and leave for future optimization when necessary.

akihikodaki (changes requested):

app/controllers/settings/profiles_controller.rb
```diff
@@ -14,7 +14,7 @@ class Settings::ProfilesController < ApplicationController
   def show; end
 
   def update
-    if @account.update(account_params)
+    if UpdateAccountService.new.call(@account, account_params)
```
Gargron: You should check if a similiar change is needed in app/controllers/api/v1/accounts/credentials_controller.rb (update via API)

app/services/update_account_service.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class UpdateAccountService < BaseService
+  def call(account, params)
+    before_locked = account.locked
```
Gargron: Minor nitpick: would call this `was_locked`

Then the conditional reads "if was locked and is not locked" which sounds better

Gargron (changes requested):

app/controllers/settings/profiles_controller.rb
masarakki: I fixed, but I think using `raise_error` option to detect `:update` or `update!` is not good.
Do you have any idea?

app/services/update_account_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class UpdateAccountService < BaseService
+  def call(account, params, raise_error: false)
+    was_locked = account.locked
+    update_method = raise_error ? :update! : :update
+    account.send(update_method, params).tap do |ret|
+      next unless ret
+      authorize_all_follow_requests(account) if was_locked && !account.locked
```
akihikodaki: Why not `authorize_all_follow_requests(account) if !ret && (was_locked && !account.locked)`?

app/services/update_account_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+  end
+
+  private
+
+  def authorize_all_follow_requests(account)
+    FollowRequest.where(target_account: account).each  do |req|
+      AuthorizeFollowWorker.perform_async(req.account_id, req.target_account_id)
```
akihikodaki: Mastodon uses `sidekiq-bulk`.

akihikodaki (changes requested):

app/services/update_account_service.rb
masarakki: I want to explicitly state that "following codes are only for succeeded situation" .
As another reason, it makes easy to add other code like this.

Gargron (approved):

akihikodaki (approved):

PR #4662, opened 2017-08-22
Fix bugs which OStatus accounts may detected as ActivityPub ready

unarist:
This patch fixes fallback handling on FetchAtomService, which is called from FetchRemoteAccountService

* Skip activity+json link if that activity is Person without inbox
* If unsupported activity was detected and all other URLs failed, retry with ActivityPub-less Accept header

then use FetchRemoteAccountService from ActivityPub::Activity::Create.

### DIFF
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -6,6 +6,8 @@ class ActivityPub::ProcessAccountService < BaseService
   # Should be called with confirmed valid JSON
   # and WebFinger-resolved username and domain
   def call(username, domain, json)
+    return unless json['inbox'].present?
+
     @json     = json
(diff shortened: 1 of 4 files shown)


app/services/activitypub/process_account_service.rb
```diff
@@ -6,6 +6,8 @@ class ActivityPub::ProcessAccountService < BaseService
   # Should be called with confirmed valid JSON
   # and WebFinger-resolved username and domain
   def call(username, domain, json)
+    return unless json['inbox'].present?
```
unarist: Unintended protocol changing will affects feed receiving, so I added this check here to ensure validity.

Maybe raise exception would be more useful for debugging.

Gargron (approved):

PR #4663, opened 2017-08-22
Allow columns to be wider on large screen.

Aldarone:
This line make the columns take full width on large screen but still allow horizontal scrollbar on small screens or when there is to much of them.

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1364,6 +1364,7 @@
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
+  flex-grow: 1;
 
   > .scrollable {
     background: $ui-base-color;


Aldarone:
(Screenshots incoming)

wxcafe:
imo this feature needs a toggle in the settings -- I personally definitely don't like the wide columns.

ElkoHRH (approved):

nightpool:
1. this doesn't affect the left-hand drawer, should it?
2. On cybre.space we allow the columns to resize between 400 and 330 px. This is a pretty good compromise between usability (long text lines are hard to read) and flexibility. Would be interested to hear @Gargron's thoughts on this.

nightpool (changes requested):
probably needs at least *some* max-width, although we can bikeshed the specific px value

clarcharr:
> On cybre.space we allow the columns to resize between 400 and 330 px. This is a pretty good compromise between usability (long text lines are hard to read) and flexibility. Would be interested to hear @Gargron's thoughts on this.

TBQH I would very much like to see these changes made to cybre.space backported to upstream, gated by features as necessary.

badosu:
Even if the wider columns are not accepted please just center the application so it's easier to read on large displays.

lostinlight:
Small change that makes a big difference :)

On a 1440 wide screen now:
screen preview without change

I am manually setting flex of .drawer & .columns to '0 1 25%' instead of fixed px width. Would be great to see Aldarone's PR merged. 
screen preview with change
 
Centering the wrapper in case wider columns aren't accepted, would *really* be helpful! 
`.ui {align-items: center}`

Gargron:
>imo this feature needs a toggle in the settings -- I personally definitely don't like the wide columns.

Agree with this. Will accept a variant of this behind a preference.

wxcafe:
There's one in glitchsoc, maybe we could just merge the code

trwnh:
For legibility, it's pretty widely accepted that lines of text should be 45-75ch wide. I understand cases where perhaps a user will want to override this, but at that point they may as well start setting their own custom CSS. I personally use 75ch as a width, on a 2560px wide monitor.

Alternatively, zooming in works wonders for making the columns wider, if you also don't mind having larger text.

PR #4664, opened 2017-08-22
Fix up the applications area

Gargron:
- Section it into "Development" area
- Improve UI of application form, index, and details


Gargron:
cc @muffinista

muffinista (approved):
👍

PR #4672, opened 2017-08-23
Fetch reblogs as Announce activity instead of Note object

unarist:
We use `/users/:user/statuses/:id` format URL as ActivityPub alternate link in the Atom feed (#4623). This endpoint represents Note object of the status, but it responds with wrong Note object for reblogs since reblogs itself can only be represented as Announce activity,

After this patch, we only use `/users/:user/statuses/:id/activity` for reblogs.

- [x] Process Create / Announce activity in FetchRemoteStatusService
- [x] Change url/uri for the reblog to `/users/.../statuses/.../activity`
- [x] Redirect to (local/remote) original status on non-activity endpoint for reblogs

### DIFF
--- a/app/controllers/statuses_controller.rb
+++ b/app/controllers/statuses_controller.rb
@@ -9,6 +9,7 @@ class StatusesController < ApplicationController
   before_action :set_status
   before_action :set_link_headers
   before_action :check_account_suspension
+  before_action :redirect_to_original, only: [:show]
 
   def show
     respond_to do |format|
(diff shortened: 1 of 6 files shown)


unarist:
Still I'm not sure about side effect of using activity URLs for id(uri) / url of reblogs, especially on OStatus/ActivityPub bridging...though it's probably ok.

akihikodaki:
7b943ce8c5dbbff52d67b83080c6ef331c3b7d47: I do not think it still makes sense to have representations for reblogs as objects since Activity Streams 2.0 does not have such a object type. Neither do I have an idea when the HTML representation is useful.
Why not just remove those representations and return `302 Found` for the original status?

unarist:
@Gargron This seems working. Could you review this?

Gargron (approved):

PR #4675, opened 2017-08-23
Pinned statuses

Gargron:
image

- "Pin on profile" / "Unpin from profile"
- "Pinned toot" (1) on public profile
- GET /api/v1/accounts/:id/statuses?pinned=true
- POST /api/v1/statuses/:id/pin
- POST /api/v1/statuses/:id/unpin
- Only non-reblogs, self-authored, public/unlisted statuses can be pinned


ProgVal:
Why not rebloged toots? (I have always been curious why Twitter added this restriction)

ProgVal:
Also, why not followers-only toots? Most private accounts I have seen on Twitter have a pinned tweet.

Gargron:
Can always open it up later but for now restricting it to public-only makes displaying logic much simpler.

Also a reblog is not a "real" status, it has no content on its own, so it's pointless to pin it.

ProgVal:
> Also a reblog is not a "real" status, it has no content on its own, so it's pointless to pin it.

I meant pinning someone else's toot.

Several times on Twitter I have been disappointed that wasn't possible.

Gargron:
Oh. That's just... A bit prone with abuse. I mean, sure you can already reblog public posts, you can screenshot them, etc, but I feel like if someone made my post permanently displayed on *their* profile, it would make me feel very uncomfortable.

Again, if there's demand, this could be opened up later.

beatrix-bitrot:
is this only for the public profile pages?

Gargron:
@beatrix-bitrot I didn't code anything to federate this yet, so I don't see a point in inconsistently showing this in the web UI but for local users only.

beatrix-bitrot:
that makes a lot of sense, thanks

if it's not TOO terrible to federate it and have it show in the web UI as well, i think you should tho

Gargron:
>25 files changed

I haven't even run `yarn manage:translations` yet

No federation in this PR, and I currently don't have a plan on how to do it, anyway.

beatrix-bitrot:
yeah, i didn't mean necessarily in this PR just, if you can figure out how to do it it would be very cool

jomo:
@ProgVal sorry this is slightly off topic; on Twitter you can actually pin your own retweets using the status ID of your retweet (requires some request forgery, though).

azbulutlu:
this will be very useful for those of us who wants to use our profiles to feature our art.

config/routes.rb
```diff
@@ -210,8 +214,14 @@
         resource :search, only: :show, controller: :search
         resources :relationships, only: :index
       end
+
       resources :accounts, only: [:show] do
-        resources :statuses, only: :index, controller: 'accounts/statuses'
+        resources :statuses, only: :index, controller: 'accounts/statuses' do
+          collection do
+            get :pinned
```
ykzts: Is not this used?

app/controllers/accounts_controller.rb
```diff
@@ -7,14 +7,17 @@ class AccountsController < ApplicationController
...
+
         if current_account && @account.blocking?(current_account)
           @statuses = []
           return
         end
 
-        @statuses = filtered_statuses.paginate_by_max_id(20, params[:max_id], params[:since_id])
-        @statuses = cache_collection(@statuses, Status)
-        @next_url = next_url unless @statuses.empty?
+        @pinned_statuses = cache_collection(@account.pinned_statuses.limit(1), Status) unless media_requested?
```
ykzts: In what sort order will this be returned?

Only the first pinned post was displayed.

Gargron: The same default_scope as all other statuses - id desc.

Yes, the public profile only displays the most recent pin. The account statuses API returns all of them as a collection, though.

ykzts (approved):

PR #4690, opened 2017-08-25
Allow multiple pinned statuses to be shown and make them be ordered b…

nullkal:
…y pinned date

### DIFF
--- a/db/migrate/20170824103029_add_timestamps_to_status_pins.rb
+++ b/db/migrate/20170824103029_add_timestamps_to_status_pins.rb
@@ -1,0 +1,5 @@
+class AddTimestampsToStatusPins < ActiveRecord::Migration[5.1]
+  def change
+    add_timestamps :status_pins, null: true
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20170824103029_add_timestamps_to_status_pins.rb
```diff
@@ -0,0 +1,5 @@
+class AddTimestampsToStatusPins < ActiveRecord::Migration[5.1]
+  def change
+    add_timestamps :status_pins, null: true
```
akihikodaki: The feature is not included in any release yet and we do not have to maintain the compatibility. What about truncate the table and set `null: false`?

akihikodaki (changes requested):

akihikodaki (approved):

Gargron:
Is it necessary to order by pin date? Why not chronologically like everything else?

nullkal:
If pinned posts are sorted by post date, visitors who decides if the content is updated by checking the top pinned post may not be able to notice the user additionally pinned an older post. So, I think pinned posts should be sorted by pinned date.

PR #4694, opened 2017-08-25
Add descriptions of scopes in label

nullkal:
Another solution of #4691. I think the name of the scope (e.g. `read`, `write`, or `follow`) should be displayed because `/oauth/authorize` takes scope as a parameter.


nullkal:
The same change has been made in #4691, so I close this PR.

PR #4695, opened 2017-08-25
i18n Updated strings (#4675 - pinned toot)

Quent-in:
+ a line I accidentally deleted


ykzts (dismissed):

config/locales/oc.yml
```diff
@@ -602,6 +607,7 @@ oc:
     disable: Desactivar
     enable: Activar
     enabled: L’autentificacion en dos temps es activada
+    enabled_success: Autentificacion en dos temps ben activada
```
ykzts: It may be better to replace the `enabled` message with the message of `enabled_success`.

The message of `enabled_success` is displayed when processing succeeds.

Quent-in: Thanks for this piece of advice! I've just made the change.

ykzts (approved):

PR #4697, opened 2017-08-25
Generalized the infinite scrollable list

abcang:
ScrollableList and IntersectionObserverArticle are generalized components of infinite scroll function of status list. This component uses IntersectionObserver and can be used not only for the status list, but also for notification and so on.


Gargron (approved):

Gargron:
I am not yet sure if this started with this PR, but I got a report of this happening: <https://niu.moe/@Guresuke/2913618>

>But on every time, notification column goes to top of itself when any message boxes clicked

I can confirm same is happening to me. Clicking on any notifications scrolls the column to top.

abcang:
I'm sorry. I also confirmed that the same thing will happen. And it fixed it in # 4792.

PR #4702, opened 2017-08-26
Add _:inReplyToAtomUri to ActivityPub

Gargron:


app/serializers/activitypub/note_serializer.rb
```diff
@@ -9,6 +9,7 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
   has_many :virtual_tags, key: :tag
 
   attribute :atom_uri, key: '_:atomUri', if: :local?
+  attribute :in_reply_to_atom_uri, key: '_:inReplyToAtomUri', if: :reply_to_local?
```
unarist: why this limits to local? it also useful for replies to remote OStatus status.

(or, match url column with inReplyTo, not only uri column)

unarist (approved):

PR #4704, opened 2017-08-26
Use _:atomUri in Delete activities as fallback

Gargron:


unarist:
in this context, atomUri means "atom uri of 'object'", not "atom uri of this object" as we used in Note object. maybe other name like "objectAtomUri"?

unarist:
...or use atomUri inside of the Tombstone object?

Gargron:
@unarist Done

unarist (approved):

PR #4709, opened 2017-08-26
Forward ActivityPub creates that reply to local statuses

Gargron:
- When we receive a payload that is a reply to a local status, we forward that payload to followers of the replied-to status author, but only if the reply is either public or unlisted
- When a local status is created and is a reply to another local status, we distribute that reply to followers of the replied-to status author, but only if the reply is either public or unlisted
- Skip forwarding of payloads that don't have a signature

### DIFF
--- a/app/lib/activitypub/activity/delete.rb
+++ b/app/lib/activitypub/activity/delete.rb
@@ -16,6 +16,8 @@ class ActivityPub::Activity::Delete < ActivityPub::Activity
   private
 
   def forward_for_reblogs(status)
+    return if @json['signature'].blank?
+
     ActivityPub::RawDistributionWorker.push_bulk(status.reblogs.includes(:account).references(:account).merge(Account.local).pluck(:account_id)) do |account_id|
       [payload, account_id]
     end
--- a/app/workers/activitypub/reply_distribution_worker.rb
+++ b/app/workers/activitypub/reply_distribution_worker.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class ActivityPub::ReplyDistributionWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'push'
+
+  def perform(status_id)
+    @status  = Status.find(status_id)
+    @account = @status.thread.account
+
+    return if skip_distribution?
+
+    ActivityPub::DeliveryWorker.push_bulk(inboxes) do |inbox_url|
+      [signed_payload, @account.id, inbox_url]
+    end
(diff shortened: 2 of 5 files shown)


app/workers/activitypub/reply_distribution_worker.rb
```diff
@@ -0,0 +1,42 @@
...
+    @status  = Status.find(status_id)
+    @account = @status.thread.account
+
+    return if skip_distribution?
+
+    ActivityPub::DeliveryWorker.push_bulk(inboxes) do |inbox_url|
+      [signed_payload, @account.id, inbox_url]
+    end
+  rescue ActiveRecord::RecordNotFound
+    true
```
beatrix-bitrot: when does this happen? if the status was deleted before we can distribute it? is it worth logging something here maybe?

Gargron: Consistent with all other workers. Not worth logging.

unarist (changes requested):

app/lib/activitypub/activity/delete.rb
```diff
@@ -16,6 +16,8 @@ def perform
   private
 
   def forward_for_reblogs(status)
+    return if @json['signature'].blank?
```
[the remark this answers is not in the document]
Gargron: I think I already did..? Current tests pass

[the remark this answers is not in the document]
unarist: Oops, sorry.

app/workers/activitypub/reply_distribution_worker.rb
```diff
@@ -0,0 +1,42 @@
...
+  sidekiq_options queue: 'push'
+
+  def perform(status_id)
+    @status  = Status.find(status_id)
+    @account = @status.thread.account
+
+    return if skip_distribution?
+
+    ActivityPub::DeliveryWorker.push_bulk(inboxes) do |inbox_url|
+      [signed_payload, @account.id, inbox_url]
```
unarist: When alice replies bob, json includes alice as `actor` and bob as `account_username`, then signed by bob's key. In this case, receiver instance treats the status is authored by bob instead of alice.

unarist (changes requested):

app/workers/activitypub/reply_distribution_worker.rb
Gargron: You're right, I messed it up. We need status.thread.account's followers, but we need status.account's signature.

unarist (approved):

PR #4714, opened 2017-08-27
Update FR locales

eramdam:
- [x] Add locales for the new profile view
- [x] Add locales for the new pin feature
- [x] Fix missing locales in `doorkeeper.fr.yml` and clean un-used ones
- [x] Fix a few typos in the locales


ykzts:
cc @aaribaud @Aldarone @Angristan @BoFFire @bzg @cgxxx @cyweo @Eychics @goofy-bz @Jnsll @kklleemm @Komic @nclm @ProgVal @SansPseudoFix @Sylvhem @tcitworld @trebmuh

app/javascript/mastodon/locales/fr.json
```diff
@@ -43,7 +43,7 @@
   "column_header.unpin": "Retirer",
   "column_subheading.navigation": "Navigation",
   "column_subheading.settings": "Paramètres",
-  "compose_form.lock_disclaimer": "Votre compte n’est pas {locked}. Tout le monde peut vous suivre et voir vos pouets restreints.",
+  "compose_form.lock_disclaimer": "Votre compte n’est pas {locked}. Tout le monde peut vous suivre et voir vos statuts privés.",
```
Aldarone: Why did you replace « pouet » with « statuts » ? 😭

config/locales/fr.yml
```diff
@@ -404,16 +426,23 @@ fr:
...
     preferences: Préférences
     settings: Réglages
     two_factor_authentication: Identification à deux facteurs
+    your_apps: Vos applications
   statuses:
     open_in_web: Ouvrir sur le web
     over_character_limit: limite de caractères dépassée de %{max} caractères
+    pin_errors:
+      ownership: Vous ne pouvez pas épingler un statut ne vous appartenant pas
+      private: Non-public toot cannot be pinned
```
Aldarone: You forgot one here 😉

Aldarone (changes requested):

tcitworld (changes requested):
I don't approve the change from `pouets` to `statuts`. People are quite used to `pouets` right now.

config/locales/fr.yml
```diff
@@ -212,7 +226,13 @@ fr:
     signature: Notifications de Mastodon depuis %{instance}
     view: 'Voir :'
   applications:
+    created: Application créée avec succès
+    destroyed: Application supprimée avec succès
     invalid_url: L’URL fournie est invalide
+    regenerate_token: Regénérer le jeton d'accès
+    token_regenerated: Jeton d'accès regénéré avec succès
+    warning: Soyez prudent⋅e avec ces données. Ne les partagez pas!
```
SansPseudoFix: Typo: in french, there is a space before (and after) `!`

app/javascript/mastodon/locales/fr.json
```diff
@@ -180,7 +180,7 @@
   "status.show_less": "Replier",
   "status.show_more": "Déplier",
   "status.unmute_conversation": "Ne plus masquer la conversation",
-  "status.unpin": "Unpin from profile",
+  "status.unpin": "Désépingler du profil",
```
SansPseudoFix: I understand why, but I don't think `désépingler` is a good idea. Why not "relâcher" like Firefox does?

config/locales/fr.yml
aaribaud: More precisely, in French the exclamation mark requires a thin non breakable space ('espace fine') before and a plain (breakable) space after (as do interrogation point, colon, semicolon and some others).

Komic:
I personally don't really like how "_pouet_" sounds in French but IMO it should be kept.

app/javascript/mastodon/locales/fr.json
Komic: I think "_Relâcher_" is too vague. "_Désépingler_" is a valid word _and_ more precise so why not~

trebmuh:
I agree with others' comments.
Apart from those details, good job and thanks @eramdam !

app/javascript/mastodon/locales/fr.json
aaribaud: I'm not too fond of "désépingler", which sounds like it was made up for the occasion -- although it does exist in French as a construction based on "épingler", its meaning is "remove pins from sth" rather than "free sth from the the pins which held it".

How about "Retirer"? Its common meaning include "remove a publication from display".

bzg:
> How about "Retirer"? Its common meaning include "remove a publication from display".

A strong +1 for "retirer".

Komic:
sure, whatever

eramdam:
@Aldarone @Komic @tcitworld The "pouet" was removed from basically everywhere else in the locales but one string, I can go and put it back everywhere else, but, like, I'm not sure when and how it was removed in the first place.

Aldarone:
@eramdam It seems that in the english strings there is a mix of « post » and « status » which saddens me a bit more. I won't object sticking close to the original strings though.

app/javascript/mastodon/locales/fr.json
eramdam: I went with "désépingler" because this is how it was translated in French in Twitter's UI 
image

As much as the word seems "weird" I agree with @Komic that it's more precise anyway, and will feel familiar instantly to those who used Twitter before coming to Mastodon since it's basically the same concept/feature.

eramdam:
@Aldarone yeah, it doesn't help. I'd say I can remove the commit on this specific PR and we can do/discuss the change (or not) later on, because it's quite a big chunk of strings to update.

config/locales/fr.yml
```diff
@@ -404,16 +426,23 @@ fr:
...
     two_factor_authentication: Identification à deux facteurs
+    your_apps: Vos applications
   statuses:
     open_in_web: Ouvrir sur le web
     over_character_limit: limite de caractères dépassée de %{max} caractères
+    pin_errors:
+      ownership: Vous ne pouvez pas épingler un statut ne vous appartenant pas
+      private: Non-public toot cannot be pinned
+      private: Les statuts non-publics ne peuvent pas être épinglés
+      reblog: Un partage ne peut être épinglé
```
kklleemm: What about "ne peut pas" ? (just like line 444 above)

app/javascript/mastodon/locales/fr.json
aaribaud: I'm not too fond of a translation choice that would be summed up as "let's use twitter-specific unusual French rather than plain simple French". In translation as in many other domains, simplicity is important.

eramdam:
@tcitworld @Aldarone @kklleemm I fixed the things you pointed out.

Now about the whole `pouet` situation: 
1. The FR locales already had a mix of `statuts` and `pouets`, I barely removed the last `pouet` (outside of the "send" button in the composer) for the sake of uniformity
2. The EN locales also have a mix of `toot`, `posts` and `status`, and I'm not sure what's the logic to follow is supposed to be
3. I put back the `pouet` that I removed with this PR
4. I'm more than happy to discuss a way to improve this whole thing w/ everybody involved, but I think it deserves its own PR. I think we had to find kind of a "logic" to follow, to decide when to use `statut` or when to use `pouets`, because some parts of the UI would be weird with `pouet` and others might be completely fine with it.

Gargron:
>The EN locales also have a mix of toot, posts and status, and I'm not sure what's the logic to follow is supposed to be

I wish I could answer that, but it's mostly just a chronological progression of the project!

Gargron:
Is this good to merge?

kklleemm (approved):

Gargron (approved):

PR #4718, opened 2017-08-27
Append new replies from timelines or notifications in web UI

Gargron:


nightpool:
Looks like this is only for statuses that are directly replying to other statuses? Feels weird that you can end up in an inconsistent state where if you go *further* down the reply chain, more statuses show up.

unarist:
Also this change will bring back an issue mentioned in #3331, probably.

Gargron:
Okay, I need to rework this. Likely too ambitious of a topic to get done before next release.

PR #4722, opened 2017-08-27
Add japanese translations for shorten display of large numbers

lynlynlynx:
Japanese translations for #4711 
Since there is no four digit break units option in Rails, I set same as English units 😭

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -352,6 +352,17 @@ ja:
     reblog:
       body: 'あなたのトゥートが %{name} さんにブーストされました:'
       subject: あなたのトゥートが %{name} さんにブーストされました
+  number:
+    human:
+      decimal_units:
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
Shouldn't this fallback to the one in en.yml? This doesn't differ from the english one, does it?

ykzts:
https://github.com/svenfuchs/rails-i18n/blob/cebc19038c0402e97bc6d47aabab44c1045428c2/rails/locale/ja.yml#L165-L173

By default `rails-i18n` will use the above. These are not common in Japanese.

Gargron:
Ah, so the problem is that we have rails-i18n gem. I was assuming all missing keys like that would fallback to en.yml. Hmm. I wonder if rails-i18n has exclusion settings...

PR #4723, opened 2017-08-28
fix toot button text be hidden when img pasted

matyapiro31:
This fix "toot!" button broken style in Japanese.

before:
```
____________
|  トゥー...|
‾‾‾‾‾‾‾‾‾‾‾‾ 
```
after:
```
__________
|トゥート!|
‾‾‾‾‾‾‾‾‾‾ 
```

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -78,7 +78,7 @@
 
   &.button--block {
     display: block;
-    width: 100%;
+    width: auto;
   }
 }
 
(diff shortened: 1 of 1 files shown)


lynlynlynx:
Japanese translation is not the longest, overflow ellipsis options are needed for Bulgarian and Ukrainian.
Since setting `width: auto;` dissolves overflow and text-overflow options, we cannot use it.

I think it is impossible to support all languages in the toot button space of current style.
Ukrainian can't be displayed all with no-padding buttons anyway. 

Please be sure to respect for all languages :ukraine: 
ws000032
With this commit.

ws000033
No padding button.

Gargron:
Sorry, but I agree with lynx.

PR #4732, opened 2017-08-29
Guarantee Subscription service first account has proper URL details

### LINKED ISSUE #4731: Deleted First User Breaks Subscriptions
Hello!  I discovered a major (yet small) bug breaking remote federation where my first account on the instance has been suspended (such as by deleting themselves), thus breaking subscription requests to other Mastodon servers.

I've fixed this locally by changing the line of code in the SubscribeService for the "some_local_account" method to always point at my account instead of at the first one.  (Should be line 45.)

Probably a better fix would be to establish a link to the first admin account, or to filter the accounts by undeleted ones, or whatever else you deem is a good solution.

* * * *

- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).  **I am on 1.5.1**

koboldunderlord:
Resolve #4731

Subscription Service potentially could break if the first user suspended
themselves, creating a situation where the urls that populate throughout
subscription service's PuSH request would cause the remote API to throw 503 errors.

Guaranteeing that the first account picked is not suspended prevents this problem.

### DIFF
--- a/app/services/subscribe_service.rb
+++ b/app/services/subscribe_service.rb
@@ -42,7 +42,7 @@ class SubscribeService < BaseService
   end
 
   def some_local_account
-    @some_local_account ||= Account.local.where(suspended:false).first
+    @some_local_account ||= Account.local.where(suspended: false).first
   end
 
   # Any response in the 3xx or 4xx range, except for 429 (rate limit)
--- a/app/services/subscribe_service.rb
+++ b/app/services/subscribe_service.rb
@@ -42,7 +42,7 @@ class SubscribeService < BaseService
   end
 
   def some_local_account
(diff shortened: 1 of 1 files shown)


app/services/subscribe_service.rb
```diff
@@ -42,7 +42,7 @@ def subscription_params
   end
 
   def some_local_account
-    @some_local_account ||= Account.local.first
+    @some_local_account ||= Account.local.where(suspended:false).first
```
ykzts: `Account.local.where(suspended: false).first` (add space)

Gargron (approved):

PR #4733, opened 2017-08-29
Add special oEmbed handling for some providers

unarist:
- Supports: Twitter, Facebook, Qvitter
- New card type `post`: shows author_name and allows longer description

I've stored `@Gargron` to `title` field, but it rendered in `.status-card__author` to show `Eugen` in bolder font instead. It seems bad way, and maybe `title` field is not appropriate too.

image

### DIFF
--- a/app/javascript/mastodon/features/status/components/card.js
+++ b/app/javascript/mastodon/features/status/components/card.js
@@ -17,6 +17,9 @@ const getHostname = url => {
   return parser.hostname;
 };
 
+const getProviderName =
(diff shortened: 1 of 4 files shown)


beatrix-bitrot:
:/ this has conflicts now

Gargron:
I believe that the OpenGraph based tweet previews aren't actually that bad. And I would prefer to avoid special treatment based on provider_name. You can experiment with enabling "rich" OEmbed instead

PR #4739, opened 2017-08-29
Avoid sending some ActivityPub payloads if the receiver will get them…

Gargron:
… through distribution


unarist:
Well, the problem is that delete later depends the *instance* receives only one Create payload. For example:

- followed by multiple users in same remote instance: Delivery to A@foo and B@foo
(although this will be solved by sharedInbox)
- mention to follower's instance: Delivery to A@foo and mention to B@foo
- mention to multiple users in same remote instance: Mention to A@foo and B@foo

In any of those cases, instance foo will receive multiple Create activity. Then below are what happens in receiver instance:

- ~~Create Delete~~: OK
- ~~Delete Create~~: OK
- ~~Create Create Delete~~: OK
- ~~Create Delete~~ Create: NG, delete later only works if status is not exists
- ~~Delete Create~~ Create: NG, delete later only works for one Create activity

In OStatus, multiple creation after delete works since we don't delete redis entry, but Create Delete Create case would happen in OStatus too.

unarist (approved):

PR #4742, opened 2017-08-30
Update status embeds

Gargron:
- Use statuses controller for embeds instead of stream entries controller
- Prefer /@:username/:id/embed URL for embeds
- Use /@:username as author_url in OEmbed
- Add follow link to embeds which opens web intent in new window
- Use redis cache in development
- Cache entire embed

image

### DIFF
--- a/app/controllers/api/oembed_controller.rb
+++ b/app/controllers/api/oembed_controller.rb
@@ -4,14 +4,14 @@ class Api::OEmbedController < Api::BaseController
   respond_to :json
 
   def show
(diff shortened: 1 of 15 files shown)


Gargron:
To check it out:

```html
<!DOCTYPE html>
<style type="text/css">
  body { background: #eee; }
</style>
<body>
<iframe src="http://localhost:3000/@admin/1/embed" frameborder=0 scrolling=no width=400 height=600></iframe>
</body>
```

ykzts (approved):
good:+1:

PR #4748, opened 2017-08-30
Embed modal

Gargron:
image

- Does not depend on the status being from Mastodon, only supporting OEmbed
- Loads preview of embed inside a dynamically created iframe

Unsolved issue: Height of the toot iframe. This seems unsolvable without including a script tag along with the iframe.

### DIFF
--- a/app/controllers/api/web/embeds_controller.rb
+++ b/app/controllers/api/web/embeds_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class Api::Web::EmbedsController < Api::BaseController
+  respond_to :json
+
+  before_action :require_user!
+
+  def create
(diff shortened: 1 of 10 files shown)


ykzts:
https://github.com/tootsuite/mastodon/blob/2db9ccaf3eeada3106e88e08163495ae8e741574/config/application.rb#L74

I think CORS of `/@:username/:id` is necessary.

Gargron:
@ykzts Okay, I have created a proxy through Rails to make that look-up. A bit concerned with the performance implication, but at least it *will* work for remote pages without CORS.

I could not make progress on the iframe heights, however. Maybe they'll stay broken for the time being.

ykzts (approved):

PR #4753, opened 2017-08-31
Remove identity context from output of LinkedDataSignature

Gargron:


unarist (approved):
OK, I have confirmed below cases.

* A json generated by LinkedDataSignature#sign! can be verified by LinkedDataSignature#verify_account! 
* Reply forwarding works correctly (i.e. payload which is actor != sender)

PR #4754, opened 2017-08-31
Do not rely on activity arriving exactly once after delete arrived

Gargron:


unarist:
Well, how about adding redis entry regardless status existence, to solve Create Delete Create case?

Gargron:
@unarist Done

unarist (approved):

PR #4757, opened 2017-08-31
 Adjust padding on the public profile page

cygnan:
Fixed an issue where padding-bottom was set to 0 and some links were hard to tap on some Android phones.

<details>
<summary>Screenshots</summary>

Before | After
--- | ---
before | after

</details>

### DIFF
--- a/app/javascript/styles/basics.scss
+++ b/app/javascript/styles/basics.scss
@@ -6,7 +6,7 @@ body {
   line-height: 18px;
   font-weight: 400;
   color: $primary-text-color;
-  padding-bottom: 40px;
+  padding-bottom: 20px;
   text-rendering: optimizelegibility;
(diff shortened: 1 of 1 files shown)


lynlynlynx:
40px is too wide for narrow devices. There is a meaning to reduce padding.

In the first place, 40px padding is too wide for any devices?
I think it is better to reduce padding to 20px as initial setting :thinking:

cygnan:
Ah, 40px padding is certainly a bit too wide. As far as I confirmed, the links are easy to tap on mobile devices even if padding-bottom is set to 20px.

Updated. How's this?

lynlynlynx (approved):
Yeah, it's nice! Thanks for adopting my opinion.

Gargron (approved):

PR #4761, opened 2017-09-01
Add ActivityPub handler for Delete->Actor activities

Gargron:


app/lib/activitypub/activity/delete.rb
```diff
@@ -13,8 +27,6 @@ def perform
     end
   end
 
-  private
```
unarist: Why removed `private`?

ykzts: @unarist https://github.com/tootsuite/mastodon/pull/4761/files#diff-b0a39931113e4e6d547ed221a1121a4cR12

unarist: @ykzts Oh, thanks.

unarist (approved):

PR #4763, opened 2017-09-01
Deduplicate with local status on Create activity

unarist:
This fixes below bug:

1. Bob replies to Alice in remote instance
2. Alice's instance forwards it to followers, including Bob (#4709)
3. **Bob's instance creates that status again** since it doesn't deduplicate with local status

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -25,7 +25,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   private
 
   def find_existing_status
-    status   = Status.find_by(uri: object_uri)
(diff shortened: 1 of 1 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -25,7 +25,7 @@ def perform
   private
 
   def find_existing_status
-    status   = Status.find_by(uri: object_uri)
+    status   = status_from_uri(object_uri)
     status ||= Status.find_by(uri: @object['_:atomUri']) if @object['_:atomUri'].present?
```
Gargron: Should this also be status_from_uri? Since atomUri could be local too, maybe..? 🤔

Gargron (approved):

PR #4766, opened 2017-09-01
Instantly upgrade account to ActivityPub if we receive ActivityPub payload

Gargron:


app/controllers/activitypub/inboxes_controller.rb
```diff
@@ -26,8 +26,12 @@ def body
...
 
   def upgrade_account
-    return unless signed_request_account.subscribed?
-    Pubsubhubbub::UnsubscribeWorker.perform_async(signed_request_account.id)
+    if signed_request_account.ostatus?
+      signed_request_account.update(last_webfingered_at: nil)
+      ResolveRemoteAccountWorker.perform_async(signed_request_account.acct)
+    end
+
+    Pubsubhubbub::UnsubscribeWorker.perform_async(signed_request_account.id) if signed_request_account.subscribed?
```
unarist: We shouldn't unsubscribe unless the account is recognized as ActivityPub, since PuSH subscription is the only way to distribute to them until that time.

app/workers/resolve_remote_account_worker.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ResolveRemoteAccountWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'pull'
```
unarist: Maybe this worker should be `unique`?

unarist (changes requested):

app/controllers/activitypub/inboxes_controller.rb
Gargron: But we've already received an AP payload from the account here. That means we don't need a PuSH subscription anymore. It's 99.9% guaranteed that ResolveRemoteAccountWorker will upgrade the account to ActivityPub from here on, I just don't think we should do it synchronously because it involves a couple doing a couple HTTP requests.

app/workers/resolve_remote_account_worker.rb
Gargron: Good point.

app/controllers/activitypub/inboxes_controller.rb
unarist: I was thinking delaying unsubscribe to next AP payload.

```
if ostatus?
  resolve async
else
  unsubscribe async if subscribed
end
```

Benefit of this would be small in normal case but the cost of this would also be small. So I'd like to prefer safer way, but it's ok if you don't want.

unarist: Actually, this may be a problem if they toots rarely, because we need to send updates via both of OS/AP until next their payload.

unarist (approved):

PR #4767, opened 2017-09-01
Define missing JSON-LD properties

Gargron:
Using _: property names is discouraged, as in the future,
canonicalization may throw an error when encountering that instead
of discarding it silently like it does now.

We are defining some ActivityStreams properties which we expect
to land in ActivityStreams eventually, to ensure that future versions
of Mastodon will remain compatible with this even once that happens.
Those would be `locked`, `sensitive` and `Hashtag`

We are defining a custom context inline for some properties which we
do not expect to land in any other context. `atomUri`, `inReplyToAtomUri`
and `conversation` are part of the custom defined OStatus context.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -26,7 +26,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
 
   def find_existing_status
     status   = status_from_uri(object_uri)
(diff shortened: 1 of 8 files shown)


unarist (approved):
Looks good.

unarist:
Note: This PR changes OStatus related property names, so some resolved issues like below may happen again. If you encounter those issues, make sure that remote instance also merged this PR.

- Remote status may appears twice
- Status deletion may not delete federated status
- other small issues

PR #4553, opened 2017-08-08
Use system's default font on non web UI pages

ykzts:
follow up #4033

PR #4774, opened 2017-09-02
Fix profile page when use system's font (regression from #4553)

ykzts:
ref https://gist.github.com/lindwurm/634ca7ea422d62f811103a185e63f845#file-docker_logs_mastodon_web_1_-tail-500-L211


unarist (approved):
👍

PR #4776, opened 2017-09-02
Raise an error for remote url in StatusFinder

unarist:
Previous implementation had allowed remote url with status id which also exists on local.

Then that bug leads /api/web/embed to return wrong embed url.

### DIFF
--- a/spec/controllers/api/oembed_controller_spec.rb
+++ b/spec/controllers/api/oembed_controller_spec.rb
@@ -8,6 +8,7 @@ RSpec.describe Api::OEmbedController, type: :controller do
 
   describe 'GET #show' do
     before do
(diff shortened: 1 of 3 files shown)


Gargron (approved):

spec/controllers/api/oembed_controller_spec.rb
```diff
@@ -8,6 +8,7 @@
 
   describe 'GET #show' do
     before do
+      request.host = Rails.configuration.x.local_domain
```
unarist: Default host is "test.host" while our local_domain is "cb6e6126.ngrok.io". Maybe we should set these configuration globally, but I added quick fix for now.

PR #4781, opened 2017-09-02
Make "unfollow" undo pending outgoing follow request too

### LINKED ISSUE #785: Can't cancel a pending private-account follow request
My mastodon instance where having SSL issues so others instances can't talk with me.
I made a follow request to a protected account.
I repaired the SSL issue.
I can't cancel the follow request.

I think this may be the same if the person doesn't accept or refuse the request ?

I tried using the API (Mastodon.py) but an unfollow respond with:
```In the server logs:
method=POST path=/api/v1/accounts/3/unfollow format=*/* controller=Api::V1::AccountsController action=unfollow status=500 error='NoMethodError: undefined method `account' for nil:NilClass' duration=22.79 view=0.00 db=4.39
[xxx] NoMethodError (undefined method `account' for nil:NilClass):
[xxx] app/services/unfollow_service.rb:16:in `build_xml'
[xxx] app/services/unfollow_service.rb:9:in `call'
[xxx] app/controllers/api/v1/accounts_controller.rb:89:in `u
```

And a follow:
```
In [21]: mastodon.account_follow(3)
Out[21]: {'error': 'Validation failed: Account has already been taken'}
```

The user exists in my instance:
```
In [22]: mastodon.account(3)
Out[22]:
{'acct': 'xxx@mastodon.social',
 'avatar': 'xxx',
 'created_at': '2017-04-02T21:28:41.124Z',
 'display_name': 'xxx',
 'followers_count': 0,
 'following_count': 0,
 'header': 'xxx',
 'id': 3,
 'locked': True,
 'note': 'xxx',
 'statuses_count': 3,
 'url': 'https://mastodon.social/@xxx',
 'username': 'xxx'}
```

Gargron:
- When local user unfollows other user, delete both the follow, and follow_request (if exists)
- When unfollowed user is remote, send "undo follow"/"unfollow" even for cancelled follow_request
- When processing ActivityPub "Undo->Follow", perform unfollow if remote user was following, and remove incoming follow request if the user was not following but had a pending follow request
- In OStatus processing:
  - for incoming follow, clear out any pending follow request from the follower if there was one before
  - ignore double follow requests
  - for incoming unfollow, clear out any pending follow request from the ex-follower

After this, unfollowing should clear out pending follow requests on both ends. A successful follow should also clear out pending follow requests.

Fix #785 

Screenshot of web UI change:

image

### DIFF
--- a/app/javascript/mastodon/containers/account_container.js
+++ b/app/javascript/mastodon/containers/account_container.js
@@ -32,7 +32,7 @@ const makeMapStateToProps = () => {
 const mapDispatchToProps = (dispatch, { intl }) => ({
 
   onFollow (account) {
-    if (account.getIn(['relationship', 'following'])) {
+    if (account.getIn(['relationship', 'following']) || account.getIn(['relationship', 'requested'])) {
       if (this.unfollowModal) {
         dispatch(openModal('CONFIRM', {
           message: <FormattedMessage id='confirmations.unfollow.message' defaultMessage='Are you sure you want to unfollow {name}?' values={{ name: <strong>@{account.get('acct')}</strong> }} />,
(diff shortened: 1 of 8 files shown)


ineffyble (approved):
Looks good to me and makes sense

Issue #4755, opened 2017-08-31
More gender neutral language in German translations


eribloodlust:
Currently, every user is called 'ein Benutzer' and the phrasing shows a clear assumption of a male user. Example:
"______ ist ein Benutzer auf witches.town. Du kannst ihm folgen oder mit ihm interagieren."

This thread below explains the need for more gender neutral language in the German translation of Mastodon, as well as some limitations and ideas from an accessibility and trans-inclusion perspective.
`https://discourse.joinmastodon.org/t/de-making-the-german-translation-more-gender-neutral-while-also-accessible/468`

My thoughts:

- Considering that people are not inputting their gender to Mastodon, having one gender neutral declination for all users seems like the best option.

- Because not every user is an 'er' or 'sie', the form used should be inclusive of non-binary people. (So, not 'er/sie')

- Ideally, this should be accessible to screen readers. It would be best to avoid special characters.

I have two proposals!

1. I propose to use either the ending -er_in for the suffix, and to use 'ein_e' for the article. For example 'ein_e Benutzer_in'. When it comes to the pronoun, I propose to use 'xier'. This option is admittedly not screen reader friendly but it's the most common way of referencing people in a gender neutral way in places that I am familiar with

    "______ ist ein_e Benutzer_in auf witches.town. Du kannst xiem folgen oder mit xiem interagieren."

2. The second proposal is to use the ending -lon, and to use the declination of 'ein' which corresponds with the neuter grammatical gender. For example 'ein Benutzlon'. Again I propose to use 'xier' as the pronoun, or another gender neutral pronoun if people are interested. 

    "______ ist ein Benutzlon auf witches.town. Du kannst xiem folgen oder mit xiem interagieren."

I hope that this issue is resolved. From many people I hear that changing just to a feminine standard and not masculine would be a huge change. But since gender neutral language is so important, and so rare in Germany, it would be great to set high goals and for Mastodon to be a place with progressive and inclusive use of language.

Finally, I apologize if the way I'm talking about German grammar is hard to understand or incorrect, it's not my first language!

* * * *

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

averageunicorn:
All of this! I'm the person who started the discourse thread and then got distracted, and I'm super happy that someone is picking this up.

[...]

PR #4789, opened 2017-09-03
Make german translation more gender neutral #4755

ghost:
As requested in issue #4755 a more gender neutral form of the german translation.

I preferred the "Profil" form over several genderspeak forms, because: 

- it doesn't hinder the flow while reading
- it doesn't confuse screen readers (for visually impaired)
- it does even cover the situation where the entity behind a profile is a just a script


Gargron (approved):

config/locales/simple_form.de.yml
```diff
@@ -33,10 +33,10 @@ de:
         setting_default_privacy: Beitragsprivatspäre
         severity: Gewichtung
         type: Importtyp
-        username: Nutzername
+        username: Profilname
       interactions:
         must_be_follower: Benachrichtigungen von Nicht-Folgern blockieren
```
averageunicorn: "Nicht-Folgern" should be changed to "Nicht-Folgenden"

averageunicorn:
Hi, thanks for this (not for calling the forms generally preferred by nonbinary people "genderspeak", though). One thing that I noticed that still has to be changed:

in simple_form.de.yml must_be_follower: "Nicht-Folgern" to "Nicht-Folgenden"

andydrop:
@averageunicorn 
The patch just covered the appearences of *utzer in mastodon. I will do a more broad approach within the next week, but that will need a quiet moment, that i don't have right now.

The focus will remain on screen reader usability and neutral forms combined.

Sorry for the "genderspeak" word. I used it in lack of another single word to subsume more than a dozen different approaches of gender neutral language that use syntax to specify diversity.

I would like to get more hints to other "words to change" via @DrOps@mastodon.social for the broader patch

averageunicorn:
@andydrop 
Great, thanks for exaplaining. If anything turns out to be awkward to rephrase, -lon was another form that was suggested for its screenreader compatibility. And the neopronouns (such as xier or nin) will work as well.

andydrop:
@averageunicorn 
If i absolutely had to use one of these forms, I personally like the Binnen-I (NutzerIn) and am appealed by the -lon (because it reminds me of Zini the Wuslon, for whom, I think, no gender was stated ever)

But generally i admire the "Marina Weisband" style of gender neutral writing.

PR #4796, opened 2017-09-03
Fix some ActivityPub JSON bugs

Gargron:
- Fix assumption that `url` is always a string. Handle it if it's an
  array of strings, array of objects, object, or string, both for
  accounts and for objects
- `sharedInbox` is actually supposed to be under `endpoints`, handle
  both cases and adjust the serializer

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -33,7 +33,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   def status_params
     {
       uri: @object['id'],
-      url: @object['url'] || @object['id'],
+      url: object_url || @object['id'],
       account: @account,
       text: text_from_content || '',
       language: language_from_content,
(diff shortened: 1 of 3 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -147,6 +147,16 @@ def language_from_content
     @object['contentMap'].keys.first
   end
 
+  def object_url
+    return if @object['url'].blank?
+
+    value = first_of_value(@object['url'])
```
unarist: Is first entry what we want to store? I don't know actual example, but we'd expect URL which can be resolved to activity json or HTML with link to those, I think.

Gargron: It's probably going to be fine. I don't see a use case for somebody to put a lot of different links into `url`. We could do a filter by `mimeType` but the array could be an array of strings or array of objects, and the objects could be missing `mimeType`, so I don't want to introduce the explosion of complexity here...

unarist (approved):

PR #4799, opened 2017-09-04
Fix #4794 - Fake instant follow in API response when account is believed unlocked

### LINKED ISSUE #4794: ActivityPub follow accept not forwarded to Web UI?
When issuing a follow request to a remote ActivityPub-enabled user, the Web UI displays a hourglass icon for the pending request, but does not update with the follow accept: a manual refresh of the whole Web UI is needed for it to be taken into account.

* * * *

Gargron:
See https://github.com/tootsuite/mastodon/issues/4794#issuecomment-326830374

### DIFF
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -14,6 +14,16 @@ class Api::V1::AccountsController < Api::BaseController
 
   def follow
     FollowService.new.call(current_user.account, @account.acct)
+
+    unless @account.locked?
+      relationships = AccountRelationshipsPresenter.new(
+        [@account.id],
+        current_user.account_id,
+        following_map: { @account.id => true },
+        requested_map: { @account.id => false }
+      )
+    end
+
     render json: @account, serializer: REST::RelationshipSerializer, relationships: relationships
   end
 
--- a/app/presenters/account_relationships_presenter.rb
+++ b/app/presenters/account_relationships_presenter.rb
@@ -4,12 +4,12 @@ class AccountRelationshipsPresenter
   attr_reader :following, :followed_by, :blocking,
               :muting, :requested, :domain_blocking
 
-  def initialize(account_ids, current_account_id)
-    @following       = Account.following_map(account_ids, current_account_id)
-    @followed_by     = Account.followed_by_map(account_ids, current_account_id)
-    @blocking        = Account.blocking_map(account_ids, current_account_id)
-    @muting          = Account.muting_map(account_ids, current_account_id)
-    @requested       = Account.requested_map(account_ids, current_account_id)
-    @domain_blocking = Account.domain_blocking_map(account_ids, current_account_id)
+  def initialize(account_ids, current_account_id, options = {})
+    @following       = Account.following_map(account_ids, current_account_id).merge(options[:following_map] || {})
+    @followed_by     = Account.followed_by_map(account_ids, current_account_id).merge(options[:followed_by_map] || {})
+    @blocking        = Account.blocking_map(account_ids, current_account_id).merge(options[:blocking_map] || {})
+    @muting          = Account.muting_map(account_ids, current_account_id).merge(options[:muting_map] || {})
+    @requested       = Account.requested_map(account_ids, current_account_id).merge(options[:requested_map] || {})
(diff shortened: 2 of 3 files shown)


tateisu:
I think this PR is not good.
This PR lacks consistency within Follow API and Relationship API , and overwrites accurate information with lie.
If a user reload the relationship, he may find the relation that was in the follow state back to requested state.
The opposite will be more easier for users to understand.

The real problem is lack of event to notify that follow operation is completed.

Gargron:
>overwrites accurate information with lie.

Overwatch computes players' and projectiles' positions on the player's client *before* the canonical data arrives back from the game server and then adjusts, because waiting for that data to arrive before displaying anything would be bad UX.

Here it is the same thing. ActivityPub introduces a change: now all accounts are essentially "locked", because their server must respond with an Accept first (it can be automatic though). But we don't want a UX degradation. We want accounts that auto-accept to look exactly like unlocked accounts looked before ActivityPub. We can either fake this in the client, or on the server. Usually I would do "lag compensation" on the client, but then every client out there would have to update to introduce the same change, essentially ActivityPub would introduce a breaking change into our own REST API. So I opt into changing our REST API to do the faking, so it works like before. Auto-accept takes a couple seconds maybe, before the sidekiq job is processed on our end, then on their end, then on our end again. So unless somebody is refreshing the relationship state every second, by the next time it is loaded, it will already be what we expect it to be for real.

tateisu:
Many games will recover if the predicted future is not appropriate.
This PR does not include such recovery, it relies on manual reloading.

Gargron:
@tateisu Also, I forgot, but this behaviour is consistent with what was happening in OStatus:

In OStatus we don't have the "accept" requirement, but we did have locked accounts. But locked state of remote account could change without us realizing. So there were cases like:

- Follow remote account, believe it's unlocked, so API creates follow
- Follow remote account that's believed to be locked, so API creates follow request

In both cases we needed to check whether our assumption was true, so there was AfterFollowWorker which fetched the remote profile to see the "locked" value. If it was what we expected, there would be no change. But if it was locked when we expected unlocked, our Follow would be erased in the db and replaced with FollowRequest, and the other way around too.

So technically the API responses were the same as this, with the same side effects, without the faking. However, without the faking the consequences were worse. In cases of unlocked->locked updates, we temporarily allowed access to the feed of a user without approval.

With faking, even though the user may be misguided into believing the follow was successful, access control behaves correctly.

nightpool:
Can someone explain in clearer details the UX degradation here? If I remember correctly, when we discussed this at the time we decided it wasn't a big issue.

ThibG:
So, currently, when following someone using AP (anyone), the following relation will be displayed as pending acceptance, until the server gets an Accept, process it, *and the Web UI is reloaded*. This is different than how it behaved in OStatus, where the follow request was assumed to be accepted (in case on non-locked account).

This change marks the follow request as accepted on the client side, which will match previous UX. However, it will not update the displayed relation if a Reject is received, until the Web UI is reloaded, which is, in my opinion, highly preferable.

Gargron:
@nightpool UX degradation:

Before:

- Profile (no lock icon). User expectation: click follow, "success", shows up as "you are following this user"
- Profile (lock icon). User expectation: click follow, hourglass, waiting for approval

After:

Only the latter case remains, with and without lock icon.

Consequence:

People coming from other networks like Twitter will be confused/disappointed that they seemingly cannot follow anyone. Even though people without the lock icon are auto-accept and the follow *will* happen just 1-2sec after the click.

Gargron:
This PR is not a change of behaviour, it is a *fix* which restores *old behaviour* and keeps the API *consistent*. I regret using the word "fake" in the title because it is *technically* accurate but apparently carries enough deontological baggage that it creates all this opposition.

nightpool:
Yes, to be clear, I'm not taking issue with the faking in this case, I'm just wondering if it's possible for us to do things slightly better then previously. I particularly liked the idea of putting it into the user websocket stream or a small new notification ("You followed [blank]!") 

don't see any reason not to approve this PR though.

ThibG:
Ok, this PR makes the behavior of AP accounts consistent with OStatus ones, but I think the current “new” behavior is actually (marginally) better (not a UX designer though), and it would be much better if it updated asynchronously and accurately.

nightpool (approved):

PR #4801, opened 2017-09-04
Non-Serial ("Snowflake") IDs

### LINKED ISSUE #1059: Integrate Snowflake IDs for statuses

aschmitz:
# Overview

This is a rather involved change, with the goal of moving IDs in all Mastodon tables away from the existing serial IDs to less-predictable IDs. While #1059 is open for this as the general idea of using "Snowflake" IDs (to borrow Twitter's term), it's not 100% clear what the different motivations for that issue are. My motivation is purely to hide the total number of entries in each table. (In particular, for single-user instances, this makes it significantly more difficult to determine the number of followers-only statuses, followed/associated accounts, etc.) However, this also implements most of the requests from #1059, with the exception of a "unique status IDs across the entire federation", which cannot be done in 64 bits (the limit of the integer type used by Postgres, and what we have to stick with unless we want even more massive changes).

The IDs generated by this scheme are guaranteed to be unique on a given database (as you'd expect), and are also well-ordered to within 1ms. (That is, multiple rows created within a millisecond may appear with out-of-order IDs, but any rows created more than a millisecond apart will have IDs with the ordering you would expect: the first one has a lower ID. In fact, within a given millisecond, most IDs will be in order, but there is a chance that some entries will appear "before" the first group, if many rows are inserted within the same millisecond.) This is done on the database side of the equation, to avoid depending on the time synchronization of (possibly multiple) Rails app servers. (Database servers, if separated from the app server, will tend to be a single server or at the very least have well-synchronized clocks if multiple masters are somehow involved.)

# Probably-Breaking Change

The API now returns strings for IDs rather than IDs. There is a good chance that this will break some API clients, although more testing is probably required. (To my surprise, Tusky handled this change with no trouble.) The rationale for this is that JavaScript cannot parse JSON with numbers with more than 53 bits precisely, and we intend to use 64 bits (or at the very least, currently 57 bits). While we could go the Twitter route and provide both a semantically-correct integer and a JavaScript-usable string, it appears as though serving only the string is likely to lead to fewer issues in the long term. I'm open to exposing IDs as both strings and integers if that decision appears more prudent, however. (Ruby can output this with no problem.) More rationale for this decision is in the commit message for 86bf2c3f545275a6545bb53222cb3b552ba76c54.

# Internal Changes

* A number of database columns that handle IDs have been adjusted to be `bigint`s from `int`s. This may be a long migration for some larger sites, although it ensures that all IDs are consistently using the non-serial generator.
* A post-migration hook has been added to ensure that all `id` columns use the new ID generator. (This is done as a Rake task to ensure that it is not accidentally skipped, and to keep the burden light for creators of new tables.)
* The Redis feeds for users now store both a score and value of a status ID as the sorted set entry, leaving reblog deduplication to a separate sorted set. This code has new tests to ensure it works as expected, and is documented in 9cbfe0a10259371e840aad08fee8222340af7f3f. Note that existing feeds will temporarily appear with reblogged statuses shown but not as reblogs. This was believed to be preferable to rebuilding all feeds, although code for doing so could be implemented. (As new updates are provided to feeds, they will be handled appropriately.)
* Finally, several locations that modify user feeds have been restructured to move all feed-changing logic into FeedManager. See the commit message in 9cbfe0a10259371e840aad08fee8222340af7f3f for more information.

I know that this is a rather intense set of patches, so I'm happy to work with the maintainers to get it to a mergeable state. Please feel free to make any comments about changes that should be made and I can try to make them. (Alternatively, this branch is open to edits from maintainers, if there are minor changes.)

Closes #1059.


nightpool:
If you want to conceal the velocity of all of your followed accounts (not sure what the threat model for this is but I'll take it as given) why not just create a bot that makes private messages to itself at random intervals? Or makes (X - statuses in last hour) private posts every hour?

It seems like there are vastly simpler ways to solve this problem

aschmitz:
There are obviously other ways to conceal that information to an extent. Covering the magnitude is more difficult by doing that: to provide good concealment, I'd still end up with very large numbers sooner or later, which would eventually lead to the rest of these changes being necessary (at the very least, the 4-byte ints to 8-byte bigint changes, although the full range of status IDs in the UI is an obvious thing to want to support anyway). So, I'd rather try to get these fixes in sooner rather than later.

nightpool:
This seems like a huge change for no perceivable benefit to >90% of instances.

Also if StreamEntry id's are still sequential then you're leaking that information ANYWAY.

aschmitz:
I'll agree that it's a large change, although I think there are reasonable justifications behind it. In particular, the large changes, namely the lingering 4-byte IDs in some tables and ints-as-doubles in Redis and JavaScript issues, both feel like technical debt that will be bumped up against sooner or later. I'd argue that it's much easier to plan to fix them now - and thoroughly test before rolling out any changes - than in an "oh whoops, now what?"  situation for a large instance later.

I'm definitely recommending carefully considering and testing what might go on with these changes, and I'm happy to help to the extent that I can. I made this PR not to get it merged immediately, but to try to see what can be done about some of these things in the near-to-medium term.

(In this PR, *all* IDs (including StreamEntry IDs) are converted to non-serial IDs, to avoid any problems like that, but yes, that's one of the reasons that just incrementing some counters on a random/timed basis is a risky solution. 😉)

Gargron:
@nightpool A couple comments:

- ActivityPub makes no use of StreamEntry and we recently changed embed URLs to be based on status, not StreamEntry. StreamEntry is a legacy from the earliest days and it's on its way out - one day we'll be able to get rid of it
- If this PR does what I think it does, the real benefit is not unguessability of database size (it's a small perk though I don't really see a threat model here), it's that we'll finally be able to backload old statuses without messing up chronology. This is what I'd been wishing for since I opened that Snowflake IDs issue

Most id columns have already been bigserial/bigint for future-proofness, but there's a good point here about the fact that if we keep putting them into JSON integers, one day, on any instance, it's going to cause trouble with integer overflow. So that change needs to happen one way or another.

Now, about the timing. This is a quite big PR so it will take a while for me to read through. We're also in the middle of the 1.6 release candidate phase, and I don't think such a big set of changes should go into 1.6 at this point, it might break something in subtle ways and delay our release even longer. However as we plan to bump to 2.0 soon after, this might be a good fit for that. Stringifying IDs in the API can be one of the breaking changes allowed by a major version bump.

aschmitz:
Being able to properly backfill statuses without messing up ordering is also a good point: the importer may have to twiddle a few of the lower bits at times to avoid duplicate IDs, but that can be done easily enough (particularly in batch backfill jobs).

I'd agree that this makes sense as more of a major version bump feature, particularly with the stringified IDs in the API: I have no particular need to get this in to 1.6, so if 2.0 is not far off, that seems like a good target.

(I expect that there will be unrelated database changes between now and then, and I can rebase this branch if/as necessary.)

db/migrate/20170903041257_ids_to_bigints.rb
```diff
@@ -0,0 +1,18 @@
+class IdsToBigints < ActiveRecord::Migration[5.1]
+  def change
```
akihikodaki: It would be better if it is reversible (currently it is not because of the regex.)

akihikodaki (dismissed):
Since 1.6.1 is out, it is better to merge as soon as possible to earn time enough to review (unless 1.6.2 is planned)

I made some comments, though they are not blocking merging.

db/migrate/20170903041257_ids_to_bigints.rb
aschmitz: Given that we currently have a mix of `bigint` and `integer` types, the only way to do this is to list them all explicitly. If that's desirable, I can do it. (Mostly, I did it this way so I could move the migration around to catch the latest columns before this gets merged. Now that it's actually getting merged, explicitly naming all the `integer` _id columns is fine, if lengthy.)

aschmitz:
Thanks for the review! I've made changes to address those comments. I do have one more fix for the JavaScript UI where I missed a numeric cast, so please don't merge this until I get that sorted out this evening.

Gargron:
What is the benefit of using timestamp based IDs for accounts and literally all other tables? Can we limit that part to statuses only? Most of the tables aren't even exposed to the outside

aschmitz:
In the abstract, this gives everything the same benefits you get from
snowflake IDs for statuses: easier backfills, harder to guess counts /
frequencies, ability to distribute ID generation, de facto support for more
than 32 bits of entries, etc.

In practice, I would say that it's relatively little trouble now, in
exchange for never having to worry about wanting those features for a table
in the future. (If, say, we do this only for tables that have IDs exposed
to users, then we have to track that and not later expose any other
existing tables.) It also helps for consistency ("should this relationship
column be an int or a bigint?"). So I'd also do it to reduce the cognitive
load of development.

Gargron:
Oh but I don't mean the bigint upgrade. In fact, bigint is default for primary IDs starting with Rails 5.1 (but Mastodon tables have been created before that). So that upgrade is good. But we only need chronological sorting for statuses. I'd like to minimize the surface area of this change.

akihikodaki:
By the way, if other tables do not get snowflake IDs, should they abandon string representations in JSON, or should they be string at the same time status IDs become?
In my opinion, they should be stringified as well for consistency and future changes.

Gargron:
Yes, all primary IDs should be bigint and all bigints should be stringified in API, I do not argue with that, I am in favour of that.

aschmitz:
Ah, I see what you're saying.

I think my answer is that if we're supporting 64-bit IDs, then they should
be supported throughout the app, not just internally and then in the
JavaScript/apps when a large instance hits a lower limit (depending on what
some external services/apps use internally, probably 32 or 53 bits).
Obviously this won't happen soon in practice for most tables, but I think
I'd still like to see the IDs be strings in the JSON for everything for
that reason.

In that case, the footprint of this change is already basically the same as
if we do snowflake IDs for everything: I'd at least want them for statuses
and accounts, at which point it's just a table setting difference for the
other tables.

I think the benefits of going with that consistently now (consistency being
a benefit in my mind, but also the ability to transparently switch to
multiple servers generating IDs if someone needs that feature for a large
instance, or backfill if for some reason we want to do that in order in the
future) outweigh the downsides (slightly longer IDs for things nobody sees?
a slightly more computationally intensive ID function on less-used
tables?), but I'm open to alternative opinions.

akihikodaki:
I do not think IDs of other tables really matters (both of snowflake and old good serial are fine), but here I leave some arguments:
Snowflake IDs is assuming the objects are ordered chronologically, but there is no reason to adopt the order in other tables; maybe they will have different arrangements. We cannot really foresee that, and therefore introducing snowflake IDs does not mean to broaden the possibility of feature extensions.

aschmitz:
"Normal" serial numbers are of course ordered chronologically as well, and it's not clear why you'd benefit from any other order for IDs that must be unique but determinable in advance. That's probably the nature of unforeseen future features, though.

I do still think that using snowflake IDs for everything has benefits I've outlined above, but given that the change can be done in a relatively simple post-migration hook that shouldn't conflict with anything else, I'd be fine adding that hook for my own instance, and won't fight over it here. (As long as we're agreed on bigint IDs and stringified IDs in JSON, that shouldn't run into any issues.)

If you think it's best to restrict those changes to a few tables, let me know, and I can rework the code to do so. A configurable flag might be nicer, but I suspect not of interest(?).

(Separately, I've found and fixed what I believe to be the last two Number/string issues in the JS UI in a05351ec7a8561872d5e2ecba92cd0ae9bbe7dc1.)

aschmitz:
One last tradeoff to consider: in the initial PR, I had said "Note that existing feeds will temporarily appear with reblogged statuses shown but not as reblogs. This was believed to be preferable to rebuilding all feeds, although code for doing so could be implemented." I believe we can have a migration to "fix" this (that is, not change whether old reblogs are displayed as reblogs): it will have to visit all feeds in Redis, but it would not have to hit the database beyond getting a list of user accounts. If this is desirable, let me know, and I can write it up.

I believe once the snowflakiness of non-Status tables and this are resolved, this should be ready to merge, although I'm always open to additional review.

app/controllers/api/v1/accounts/relationships_controller.rb
```diff
@@ -7,7 +7,10 @@ class Api::V1::Accounts::RelationshipsController < Api::BaseController
   respond_to :json
 
   def index
-    @accounts = Account.where(id: account_ids).select('id')
+    accounts = Account.where(id: account_ids).select('id')
+    # .where doesn't guarantee that our results are in the same order
+    # we requested them, so return the "right" order to the requestor.
+    @accounts = accounts.index_by(&:id).values_at(*account_ids)
```
krainboltgreene: We should be ordering in SQL as much as possible.

krainboltgreene:
I would love to see the frontend int -> string diff as a standalone PR.

app/controllers/api/v1/accounts/relationships_controller.rb
aschmitz: In general, I agree. However, the API for this function allows users to supply an arbitrarily-ordered set of IDs, and the tests imply they're expected to be returned in the same order (although I can't find any docs saying one way or the other). Doing this ordering in SQL would require synthesizing a CASE to sort by, which seems unlikely to be cleaner or faster.

In practice, this should be roughly O(n) (to the extent that you believe hashes are really O(1) insertion/lookup), with very small values of n. (Also in practice, the endpoint currently doesn't actually guarantee you get entries back in the same order you requested them, so we could just toss this change.)

If you think we should generate the CASE and do it in SQL, I can write that code, it just felt unnecessary here.

aschmitz:
I can make a separate PR for stringified IDs in JSON if you'd like, though it would be a strict dependency for this PR, as this PR generates IDs above the 53-bit level at which JS Numbers can fully represent ints.

(Alternatively, you can get the same by cherry-picking 86bf2c3f545275a6545bb53222cb3b552ba76c54 and a05351ec7a8561872d5e2ecba92cd0ae9bbe7dc1.)

Gargron:
To expand upon @krainboltgreene's point, the best way to get PRs merged quickly in this repository is to keep them as small as possible. Since we have a lot of people depending on our software, our priority is keeping it stable and not breaking too much at a time.

That explains e.g. why I want snowflake IDs in one table only at first, rather than all of them. One table would be easier to fix if something goes wrong later.

I don't necessarily want to put a new contributor through having to split this into multiple PRs like the stringifying of JSON ids, though having that as a separate PR would be fantastic and would get merged instantly. Maybe we should do it the cherry-pick way..?

Another reason smaller PRs is good is that it's easier to find a breaking point with git bisect later, since PR commits are squashed into one master commit. Also accountability.

aschmitz:
Sure, that's fine (and particularly if you squash on merge, will make blame/bisect a bit easier). I just wanted to clarify that while they can be evaluated in parallel if desired, the IDs->strings has to be merged before this, or there will be some problems with the UI.

I've submitted #5019 for the number->string conversion. Given that it uses the same commits and so won't conflict, I'm inclined to leave this branch as-is, without taking those commits out (in particular, so this branch can be tested manually without having to merge the two), but I can do that if someone wants.

I'll also make another commit pulling out the "Snowflake IDs for all tables" and replacing it with a migration to just apply it to the `statuses` table, since that seems to be the conclusion (for now?).

app/controllers/api/v1/accounts/relationships_controller.rb
nightpool: My understanding is that our code doesn't depend on this anywhere, and it was just ease of use for the tests. I can ask app developers though.

aschmitz:
I believe everything we've talked about is in this PR now:

* b9299a0f046f3710e2fcebcba5e8c60ef8b85b26 fixes the status streaming issue identified in #5019, along with one additional place in FeedManager that needed to get fixed.
* 1b9f813fd51cbfd440840db9849b7415bd6f84c3 (and 4287cbedab0a50f97efd85b9f4042356fc5a7434 because I forgot to test correctly) change to only implementing timestamp/snowflake IDs for the `statuses` table.
* dae1d1109b04c8d9ab035cb2bef5ad4078c9ad2f provides a relatively fast migration to ensure that reblogs still appear as reblogs after the transition in how they are stored. (Note that reversing this migration does nothing, as supporting that would require a database query for every status for every user's feed, which seems less valuable than the alternative of allowing one extra reblog of any status on a user's feed.)

As discussed, this shouldn't be merged until #5019 is, and I can fix any merge conflicts with that (or other interim changes) in this branch. If there are other concerns with the code in here, though, let me know and I can try to address them.

app/lib/feed_manager.rb
```diff
@@ -131,4 +155,58 @@ def filter_from_mentions?(status, receiver_id)
...
+      status_rank = redis.zrevrank(timeline_key, status.id)
+      return false if status_rank.nil?
+
+      # 2. Remove the reblogged status from the `:reblogs` zset.
+      redis.zrem(reblog_key, status.reblog_of_id)
+
+      # 3. Add the reblogged status to the feed using the reblogging
+      # status' ID as its score, and the reblogged status' ID as its
+      # value.
+      redis.zadd(timeline_key, status.id, status.reblog_of_id)
```
clworld: I wonder why this line is needed.
On new redis timeline, it always score == member.
Thus, original status id is not vanished when reblog is added.
So, re-inserting original status id is not needed. I think.

akihikodaki: Updating the score will prevent from reinserting the reblogged status after the reblog has been deleted.

clworld: To prevent re-inserting, reblog_key should be used instead of timeline_key.

akihikodaki: `reblog_key` is used to track existing reblogs, so you'll see removed reblogs if you do so.

clworld: really? Feed#from_redis is reading only from timeline_key as I see and not reading from reblog_key.
and reblog_key only stores reblog data for recent REBLOG_FALLOFF items in timeline_key (FeedManager#trim).

akihikodaki: That is bad. It means trimming would cause reblogs of old statuses to be deleted.

aschmitz: I can offer two things that will hopefully explain. The first, a long description of how these different zsets work, as I wrote it up while making the changes:

FeedManager maintains two sets of data in Redis for each feed. These feeds are keyed by `FeedManager#key`, here just `[key]`. The two sets of data we manage are:

1. `[key]`: a zset of statuses that show up in the feed
2. `[key]:reblogs`: a zset of statuses that are reblogs in the feed

When trimming the feed, we trim the feed to the last `MAX_ITEMS` statuses, and then trim the `:reblogs` zset to only include items with scores that indicate they would be in the last `REBLOG_FALLOFF` statuses.

When we add a status to a feed, we normally just add it to the feed zset (keyed by and with a value of the status ID). If the status is a reblog, instead we do the following:

1. If the status it's reblogging is in the last `REBLOG_FALLOFF` statuses on the feed, we ignore the reblog.
2. If the status it's reblogging is already reblogged in the `:reblogs` zset, we ignore the reblog, because this implies that the reblog is already in the last `REBLOG_FALLOFF` messages.
3. If we haven't given up yet, we add the reblog to the feed as usual, and also add it to the `:reblogs` zset with the value of the reblogged status' ID, and a score of the reblogging status' ID*.

*: Note that this is similar to the old behavior, where step 2 was skipped, and step 3 simply added a value with the score of the reblogging status' ID, and a value of the reblogged status's ID. We can't necessarily rely on this to be usable, because status IDs are 8-byte integers, and may not be fully representable in an IEEE double-precision float, as used by Redis for scores.

When we remove a status from a feed, we are unfortunately in the same position we had been before: because we may discard some legitimate reblogs, we cannot tell if removing the reblogged status would incorrectly remove it from our feed. Instead, if a removed status is a reblog, we do the following:

1. If the reblogging status is not in the feed, stop.
2. Remove the reblogged status from the `:reblogs` zset.
3. Add the reblogged status to the feed using the reblogging status' ID as its score, and the reblogged status' ID as its value.
4. Remove the reblogging status from the feed.

aschmitz: The second thing I'll say is more handwavey but may provide a better intuition: the reblog_key zset stores data that are used to keep a status from being reblogged more than once in the general case. It is not expected to be read from in the case of a feed being read.

In the current (pre-this-PR) state of affairs, a reblog is stored with a score of the reblogging status, and a value of the reblogged status. When a reblog is deleted, that entry is removed, and one with both the score and value set to the reblogged status is added (this is in order to to avoid that status randomly disappearing if the status should have already appeared in the feed, but got a reblog attached to it first somehow. (This would happen most in the case where a user views a reblogged status and then follows the link and follows the user: their old posts should still show up in the timeline regardless of whether or not they have been reblogged.)

In this changed value, by default, we store the reblogging status' ID as both the score and value of the reblogging status. The score is stored this way to allow sorting, and yet not run into precision issues with Redis' scores being stored as double. However, as in the case now, when we remove a reblog, we can't tell if the status should still be shown or not, so we continue to err on the side of showing it. In this case, we store with a score of the old reblogging status' ID (to keep it in the same sort order), and a value of the reblogged status' ID. That's what this line is doing.

(In the future, I'd like to store a list of reasons a reblogged status should be shown to users (reblogs by which users, largely), and only re-insert a status if it has been reblogged by other users. However, that's an entirely different discussion for a different pull request.)

akihikodaki: I have came up a different concern in this discussion so please let me describe it here though it is a bit off-topic:
You say:
> When trimming the feed, we trim the feed to the last MAX_ITEMS statuses, and then trim the :reblogs zset to only include items with scores that indicate they would be in the last REBLOG_FALLOFF statuses.

And the code would actually work as you describe. However, the behavior sounds bad when considering `from_redis`. When a user reblogs an old status:
> we add the reblog to the feed as usual

So the reblogged, old status would get a small score. Such a status would be trimmed so quickly, which effectively makes the status invisible from `from_redis`, and `/api/v1/timelines/*`, a controller using the method.
I think such statuses should be persistent on the feed, and they are in the current implementation.

akihikodaki: Don't mind. I got the idea. Sorry for confusing.

clworld: Thanks, I misunderstood this PR can solve the problem about https://github.com/tootsuite/mastodon/issues/2817#issuecomment-299501175 perfectly.
It reduces problem but not perfectly solve problem. So, 2. and 3.  are still needed. OK.

app/javascript/mastodon/components/account.js
```diff
@@ -21,7 +21,7 @@ export default class Account extends ImmutablePureComponent {
 
   static propTypes = {
     account: ImmutablePropTypes.map.isRequired,
-    me: PropTypes.number.isRequired,
+    me: PropTypes.string.isRequired,
```
Gargron: Were these forgotten in the other PR? Or could you rebase on top of current master so we get a cleaner diff?

app/javascript/mastodon/actions/store.js
```diff
@@ -5,8 +5,7 @@ export const STORE_HYDRATE_LAZY = 'STORE_HYDRATE_LAZY';
 
 const convertState = rawState =>
   fromJS(rawState, (k, v) =>
-    Iterable.isIndexed(v) ? v.toList() : v.toMap().mapKeys(x =>
-      Number.isNaN(x * 1) ? x : x * 1));
+    Iterable.isIndexed(v) ? v.toList() : v.toMap());
```
Gargron: This is just default fromJS(rawState) behaviour, so custom function could be removed

    const convertState = rawState => fromJS(rawState)

lib/tasks/db.rake
```diff
@@ -16,4 +16,135 @@ namespace :db do
       end
     end
   end
+
+  # Before we load the schema, define the timestamp_id function.
+  # Idiomatically, we might do this in a migration, but then it
+  # wouldn't end up in schema.rb, so we'd need to figure out a way to
+  # get it in before doing db:setup as well. This is simpler, and
+  # ensures it's always in place.
+  Rake::Task['db:schema:load'].enhance ['db:define_timestamp_id']
```
Gargron: The other solution is changing schema format from rb to sql

Gargron:
Hm I feel like the int->bigint change could also be split into a small PR which would get merged quickly. Then after rebase reviewing this will be much simpler.

At last I wanna bring up that backfilling older statuses is important and this is still missing a clear strategy for doing that. Can the SQL function take an optional argument, which it would use as timestamp? Then in, say, `app/lib/activitypub/activity/create.rb`, we would do a `SELECT timestamp_id('accounts', 'TIMESTAMP_HERE')` first before inserting the status with that id. Of course if there's a *better* way to do it, I'm all ears, it's just the only way I can see happening here.

app/javascript/mastodon/actions/store.js
aschmitz: Unfortunately this is from a commit in #5019, so I'm not sure how much sense it would make to revise it here. I can do that, or you can just make that change in master. (Or I could submit another PR for it, if you really want.)

app/javascript/mastodon/components/account.js
aschmitz: These were indeed in #5019. I'll rebase.

lib/tasks/db.rake
aschmitz: Personally I'd prefer to keep schema.rb and deal with this rather than shift to the less-used SQL format, but I don't care strongly.

aschmitz:
I'm wondering if it would be more useful to reimplement the `timestamp_id` function in Ruby in the backfill cases: even getting a value from running the function on the database side, we'd have to potentially retry a number of times to insert a status in case the ID collides with an existing ID (with extremely low odds, but it's possible). This would run the risk of the functions diverging if one changes but the other doesn't, though. I'm not opposed to adding an optional timestamp argument, I'm just not sure if it gains anything useful.

(What I might suggest instead is to generate a timestamp ID with a random lower two bytes, try to insert it, and then keep repeating that while incrementing the timestamp by a millisecond each time the insert fails. This would be harder to do in SQL, but easy enough in Ruby. Unfortunately, Postgres' `INSERT [...] ON CONFLICT` doesn't let us just say "increment the ID until it's not a conflict".)

aschmitz:
I created #5039 for the int -> bigint change, and merged in the latest changes including those from #5019 (rather than rebase + force push, as this will end up squashed anyway, though I could do a "clean" rebase if necessary). The git/GitHub diff looks more reasonable now, however.

aschmitz:
Merged master back in after #5039. Outstanding comments as far as I know:

* [ ] Do we just drop the re-ordering in RelationshipsController? It's not super relevant here, but did seem to be more problematic with larger IDs for some reason. In general, arrays are ordered, so I'd expect things to be returned in the order requested, but if this isn't already expected by API clients, it may be better to fix the test rather than change the output. https://github.com/tootsuite/mastodon/pull/4801#discussion_r139867507
* [ ] Do we switch the stored schema to SQL from Ruby? I'm fine with it as-is, but I don't think(?) it's a hard change to make if desired. https://github.com/tootsuite/mastodon/pull/4801#discussion_r140225774
* [ ] Do we make the `timestamp_id` SQL function accept a timestamp? This wouldn't be hard, but I'm not sure how useful it would be. https://github.com/tootsuite/mastodon/pull/4801#issuecomment-331144277

I can try to make any relevant changes this evening if someone wants to make the call on any of those.

db/migrate/20170920024819_status_ids_to_timestamp_ids.rb
```diff
@@ -0,0 +1,32 @@
+class StatusIdsToTimestampIds < ActiveRecord::Migration[5.1]
+  def up
+    # Prepare the function we will use to generate IDs.
+    Rake::Task['db:define_timestamp_id'].execute
+
+    # Set up the statuses.id column to use our timestamp-based IDs.
+    ActiveRecord::Base.connection.execute(
+      "ALTER TABLE statuses
```
akihikodaki: Use heredoc (probably `<<~`).

db/migrate/20170920032311_fix_reblogs_in_feeds.rb
```diff
@@ -0,0 +1,64 @@
...
+      timeline_key = fm.key(:home, account.id)
+      reblog_key = fm.key(:home, account.id, 'reblogs')
+      redis.zrange(timeline_key, 0, -1, with_scores: true).each do |entry|
+        if (entry[0] != entry[1])
+          # The score and value don't match, so this is a reblog.
+          # (note that we're transitioning from IDs < 53 bits so we
+          # don't have to worry about the loss of precision)
+
+          reblogged_id = entry[0]
+          reblogging_id = entry[1]
```
akihikodaki: You can: `reblogged_id, reblogging_id = entry`.

db/migrate/20170920032311_fix_reblogs_in_feeds.rb
```diff
@@ -0,0 +1,64 @@
...
+      # float coersion because in this reblog tracking zset, we only
+      # need the rebloggging status' ID to be able to stop tracking
+      # entries after they have gotten too far down the feed, which
+      # does not require an exact value.
+
+      # So, first, we iterate over the user's feed to find any reblogs.
+      timeline_key = fm.key(:home, account.id)
+      reblog_key = fm.key(:home, account.id, 'reblogs')
+      redis.zrange(timeline_key, 0, -1, with_scores: true).each do |entry|
+        if (entry[0] != entry[1])
```
akihikodaki: * Do not use parentheses for `if`.
* We usually use `next` in such cases.
* Why don't you use `lazy.reject` while you do in `FeedManager`? Let's keep them consistent. (Personally `lazy.reject` are too redundant for those cases.

spec/lib/feed_manager_spec.rb
```diff
@@ -150,5 +154,112 @@
...
+          FeedManager.instance.push('type', account, Fabricate(:status))
+        end
+
+        expect(FeedManager.instance.push('type', account, reblog)).to be true
+      end
+
+      it 'does not save a new reblog of a recently-reblogged status' do
+        account = Fabricate(:account)
+        reblogged = Fabricate(:status)
+        reblog1 = Fabricate(:status, reblog: reblogged)
```
akihikodaki: What about `reblogs = 2.times.map Fabricate(:status, reblog: reblogged)` as a less redundant alternative?

akihikodaki (changes requested):
The logic looks good but I have some comments related to Ruby style.

db/migrate/20170920032311_fix_reblogs_in_feeds.rb
aschmitz: Will do.

db/migrate/20170920032311_fix_reblogs_in_feeds.rb
aschmitz: I think `FeedManager` is using `lazy.reject` because it came from some code I was modifying. I'll make it not, because I agree that it's not useful here. Fixing the other issues.

spec/lib/feed_manager_spec.rb
aschmitz: Eh, sure. (I'm planning to wrap the Fabricate call in an explicit block, though. Let me know if I should absolutely avoid doing so.)

db/migrate/20170920024819_status_ids_to_timestamp_ids.rb
aschmitz: OK.

lib/tasks/db.rake
```diff
@@ -16,4 +16,143 @@ namespace :db do
...
+  # table using timestamp IDs.
+  Rake::Task['db:schema:load'].enhance do
+    Rake::Task['db:ensure_id_sequences_exist'].invoke
+  end
+
+  task :define_timestamp_id do
+    conn = ActiveRecord::Base.connection
+
+    # Make sure we don't already have a `timestamp_id` function.
+    unless conn.execute(
```
akihikodaki: You should do:
```Ruby
unless conn.execute(<<~SQL)
  SELECT …;
SQL
```

akihikodaki (changes requested):

lib/tasks/db.rake
aschmitz: TIL.

Gargron:
Not to push too hard but can we get an example backfilling code, maybe as separate PR based on this branch? Because like, unless remote statuses get the correct sorting there is not enough benefit to justify such a huge change.

aschmitz:
Sure. Are there multiple places the backfilling is being done these days, or is it all in one place already?

Gargron:
- grep for "Status.create" in app/lib/activitypub and app/lib/ostatus

akihikodaki (changes requested):
I meant to apply the previous comment for all heredocs.

aschmitz:
D'oh. Yes, just a moment. (I also have code for @Gargron's request, but I'll do the heredoc fixes first.) I also want to fix `rake db:setup` with `RAILS_ENV=development`, where Rake appears to be trying to do both the `test` and `development` databases separately and skips declaring the `timestamp_id` function in one of them.

aschmitz:
I'm going to claim Rails is behaving badly in doing internal, private things to handle loading the schema for `test` and `development` simultaneously in `development`, but 07baf2f should fix that either way.

Code Climate gives new reports from Rubocop for define_timestamp_id being too long (most of it being comments, and almost all of it being multiline SQL queries) and Brakeman claiming that there may be SQL injection (because it contains the string `table_name`? Or because it inlines `SecureRandom.hex(16)`?). As far as I'm concerned, both are ignorable so I can add the appropriate directives, but let me know if you think they should be addressed some other way.

I'll submit a WIP PR to generically handle backdating timestamp IDs shortly, once I move stuff around into the new library.

akihikodaki (approved):
Thank you for the contribution of a month.

PR #4805, opened 2017-09-04
Change Finnish wording of "free" on landing page.

roikale:
"Ilmainen" means "gratis", but Mastodon is free as in freedom, libre – "vapaa".
https://fi.wikipedia.org/wiki/Vapaa_ohjelmisto


ykzts:
cc @JantsoP

Gargron (approved):

PR #4808, opened 2017-09-04
Adjust status embeds

lynlynlynx:
### Permit width under 400px

<details>
<summary>Before and after images</summary>
screenshot1
</details>

### Change style of follow button to support narrow width

<details>
<summary>Before and after images</summary>
screenshot2
</details>

## Scrollable status embeds

Since scripts of getting height of iframe is not implemented yet, I remove `scrolling: 'no'` option.
(Scrolling option was obsoleted in HTML5)

<details>
<summary>Before and after images</summary>
screenshot3
</details>

## Fix overflow of embed-modal in narrow devices

<details>
<summary>Before and after images</summary>
screenshot4
</details>

## And some minor adjustments of styles

- **Since frameborder option was obsoleted in HTML5, I change it to `style="border: none;"` .**
- Rearrange style selector according to content order.
- Organize directory level.

### DIFF
--- a/app/serializers/oembed_serializer.rb
+++ b/app/serializers/oembed_serializer.rb
@@ -40,8 +40,7 @@ class OEmbedSerializer < ActiveModel::Serializer
     attributes = {
       src: embed_short_account_status_url(object.account, object),
       class: 'mastodon-embed',
-      frameborder: '0',
-      scrolling: 'no',
+      style: 'width: 400px; max-width: 100%; border: none;',
       width: width,
       height: height,
     }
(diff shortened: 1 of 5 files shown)


app/serializers/oembed_serializer.rb
```diff
@@ -40,8 +40,7 @@ def html
     attributes = {
       src: embed_short_account_status_url(object.account, object),
       class: 'mastodon-embed',
-      frameborder: '0',
-      scrolling: 'no',
+      style: 'width: 400px; max-width: 100%; border: none;',
       width: width,
```
Gargron: Style should use the width from the variable

Gargron (changes requested):

app/serializers/oembed_serializer.rb
lynlynlynx: I misunderstood width setting cannot be overwritten by max-width in style settings. I fixed it. Thanks!

Gargron (approved):

PR #4813, opened 2017-09-05
Add Smartphone screen favourite back button and adjust styles

voidsatisfaction:
At first, there was no back button on favourite column.

screen shot 2017-09-05 at 5 47 26 pm

So, I made back button for smart phone screen. but It was not in good shape.

screen shot 2017-09-05 at 5 46 10 pm

Therefore I fixed this style issue as well.

screen shot 2017-09-05 at 5 45 23 pm

**Other Column looks okay as well**

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1616,11 +1616,15 @@
   cursor: pointer;
   flex: 0 0 auto;
   font-size: 16px;
-  padding-right: 15px;
+  padding: 0 5px 0 0;
   z-index: 3;
 
   &:hover {
     text-decoration: underline;
+  }
+
+  &:last-child {
+    padding: 0 15px 0 0;
   }
 }
 
@@ -2176,7 +2180,7 @@ button.icon-button.active i.fa-retweet {
   color: $ui-primary-color;
   cursor: pointer;
   font-size: 16px;
-  padding-right: 15px;
+  padding: 0 15px;
 
   &:hover {
     color: lighten($ui-primary-color, 7%);
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1616,7 +1616,7 @@
   cursor: pointer;
   flex: 0 0 auto;
   font-size: 16px;
-  padding: 0 5px 0 0;
+  padding-right: 15px;
   z-index: 3;
 
   &:hover {
@@ -2168,9 +2168,6 @@ button.icon-button.active i.fa-retweet {
   height: 100%;
   display: flex;
   height: 48px;
-  @media screen and (max-width: 1024px) {
-    margin-right: 10px;
-  }
 }
 
 .column-header__button {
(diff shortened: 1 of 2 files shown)


app/javascript/styles/components.scss
```diff
@@ -2168,6 +2168,9 @@ button.icon-button.active i.fa-retweet {
   height: 100%;
   display: flex;
   height: 48px;
+  @media screen and (max-width: 1024px) {
+    margin-right: 10px;
```
unarist: This margin would needed even for PC if column setting doesn't exists. How about removing left padding from `.column-header__button` instead of media queries on here?

voidsatisfaction: Thank you for your advice! I will try that.

app/javascript/styles/components.scss
```diff
@@ -2176,7 +2176,7 @@ button.icon-button.active i.fa-retweet {
   color: $ui-primary-color;
   cursor: pointer;
   font-size: 16px;
-  padding: 0 15px;
+  padding-right: 15px;
```
lynlynlynx: - `padding: 0 15px;` setting is needed not to reduce column setting buttons.
- Using padding is required to overwrite user agent stylesheet.

app/javascript/styles/components.scss
```diff
@@ -1616,7 +1616,7 @@
   cursor: pointer;
   flex: 0 0 auto;
   font-size: 16px;
-  padding: 0 5px 0 0;
+  padding-right: 15px;
```
lynlynlynx: - Using `padding` is required to overwrite user agent stylesheet.

`padding: 0 5px 0 0;` and `padding: 0 15px 0 0;` for width under 1024px ?
I can't think of any smart ways.. :thinking:

lynlynlynx (changes requested):
This settings make setting buttons smaller, and have overwriting problems.
Please check comments in the source code.

app/javascript/styles/components.scss
unarist: Sorry, lynlynlynx is right. Reducing padding means reducing hit area, which shouldn't happen.

app/javascript/styles/components.scss
unarist: or, use `:last-child` instead of media queries?

```
padding: 0 5px 0 0;

&:last-child {
  padding: 0 15px 0 0;
}
```

lynlynlynx: Since there are discussions on columns width, I think using last-child is suitable. Thanks!

voidsatisfaction: Thank you for nice suggestion!

app/javascript/styles/components.scss
voidsatisfaction: I just put `padding: 0 15px`

lynlynlynx (approved):
Okay! This works properly now!

Gargron (approved):

PR #4814, opened 2017-09-05
Do not execute the job with the same arguments as the retry job

abcang:
Since `sidekiq-unique-jobs` does not check the retry queue, new jobs are pushed even though there is a unique job in the retry queue. So, Mastodon runs `Pubsubhubbub::SubscribeWorker` many times. In Pawoo, there were only 5000 unique jobs out of 60,000 `Pubsubhubbub::SubscribeWorker` retry jobs.
In this PR, check the retry queue when queuing a job to Sidekiq. At this time, confirmation of uniqueness is done by using the digest key of `sidekiq-unique-jobs`.

### DIFF
--- a/app/workers/pubsubhubbub/subscribe_worker.rb
+++ b/app/workers/pubsubhubbub/subscribe_worker.rb
@@ -3,7 +3,7 @@
 class Pubsubhubbub::SubscribeWorker
   include Sidekiq::Worker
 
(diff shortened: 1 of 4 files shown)


Gargron (approved):

Gargron:
Thank you for this, very appreciated :+1:

Issue #4795, opened 2017-09-03
No easy access to own pinned toots in Web UI


ThibG:
Pinned toots are not easily accessible from within the Web UI (to be unpinned for instance), it requires manually finding the pinned toots instead. Unless I have missed something, the easiest way to do so currently is to copy the toots' URL from the public profile to the Web UI search bar, not very user-friendly.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

voidsatisfaction:
It is solved by #4817. Thank you for your suggestion!

PR #4817, opened 2017-09-06
Add Pinned toot column

voidsatisfaction:
I just added Pinned toot section on web(react side)

It is easy for a mastodon user to access their own pinned toots.

related #4795


voidsatisfaction:
## At Navigator, there is Pinned toots

screen shot 2017-09-06 at 6 32 10 pm

## Inside of Pinned toots

screen shot 2017-09-06 at 6 32 22 pm

app/controllers/api/v1/accounts/statuses_pins_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Api::V1::Accounts::StatusesPinsController < Api::BaseController
```
Gargron: There is already an API for getting an account's pinned toots... No need to add a new API.

voidsatisfaction: Thank you for reply! I will fix it.

app/javascript/mastodon/actions/pin_statuses.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const PINNED_STATUSES_FETCH_REQUEST = 'PINNED_STATUSES_FETCH_REQUEST';
+export const PINNED_STATUSES_FETCH_SUCCESS = 'PINNED_STATUSES_FETCH_SUCCESS';
+export const PINNED_STATUSES_FETCH_FAIL = 'PINNED_STATUSES_FETCH_FAIL';
+
+export function fetchPinnedStatuses() {
+  return (dispatch, getState) => {
+    dispatch(fetchPinnedStatusesRequest());
+
+    api(getState).get('/api/v1/accounts/1/statuses', { params: { pinned: true } }).then(response => {
```
abcang: `'/api/v1/accounts/1/statuses'`

I think that it is necessary to access with the logged in user.

abcang (changes requested):

app/javascript/mastodon/actions/pin_statuses.js
voidsatisfaction: Thanks! I totally forgot that

abcang (approved):

PR #4819, opened 2017-09-06
fix text position of NSFW for video file

pfm-eyesightjp:
fix https://github.com/tootsuite/mastodon/pull/4570#issuecomment-327195309

30066447-34e8dbba-9258-11e7-8ee5-6ba30346dfc0

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -2343,6 +2343,22 @@ button.icon-button.active i.fa-retweet {
   height: 100%;
 }
 
+.media-spoiler__video {
+  align-items: center;
+  background: $base-overlay-background;
+  color: $primary-text-color;
+  cursor: pointer;
+  display: flex;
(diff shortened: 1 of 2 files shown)


Gargron (approved):

app/javascript/styles/components.scss
```diff
@@ -2343,6 +2343,22 @@ button.icon-button.active i.fa-retweet {
   height: 100%;
 }
 
+.media-spoiler__video {
+  align-items: center;
+  background: $base-overlay-background;
+  color: $primary-text-color;
+  cursor: pointer;
+  display: flex;
```
unarist: Wait, does this work on Safari...?  (c.f. #4570)

nullkal:
It seems the problem is the NSFW text for images are wrapped in <button> whereas the one for videos are wrapped in <div role="button">. So I think it is better we make them consistent.

PR #4821, opened 2017-09-06
Make timeline scroll position fixed when scrollpos != top (fix regression of #4697) 

clworld:
ScrollableList treats children as Array but children is Immutable.List.
So, scroll position fixation logic in ScrollableList does not work properly.
This PR fixes it and make scroll position fixed when new things arrived && scrollpos != top.

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -5,6 +5,7 @@ import IntersectionObserverArticle from './intersection_observer_article';
 import LoadMore from './load_more';
 import IntersectionObserverWrapper from '../features/ui/util/intersection_observer_wrapper';
 import { throttle } from 'lodash';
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/scrollable_list.js
```diff
@@ -95,7 +96,12 @@ export default class ScrollableList extends PureComponent {
 
   getFirstChildKey (props) {
     const { children } = props;
-    const firstChild = Array.isArray(children) ? children[0] : children;
+    let firstChild = children;
+    if (children instanceof ImmutableList) {
+      firstChild = children.get(0);
+    } else if (Array.isArray(children)) {
```
clworld: It seems that ScrollableList is always got children as Immutable.List. So, this code perhaps isn't needed.

abcang (approved):
Thank you for fixing it.

PR #4822, opened 2017-09-06
i18n: update Persian translation

mabkenar:


config/locales/doorkeeper.fa.yml
```diff
@@ -3,8 +3,10 @@ fa:
   activerecord:
     attributes:
       doorkeeper/application:
-        name: Name
+        name: Application name
```
ykzts: work in progress?

mabkenar: I intended it to be final. I just copied whatever I found in the corresponding English file. I saw a couple of words in `config/locales/doorkeeper.en.yml` had changed and I reflected those in the Persian translation.

Gargron (approved):

config/locales/doorkeeper.fa.yml
ykzts: i see.

mabkenar: Oh you mean why aren't they translated? That was a conscious decision from the beginning to not translate that file for the moment because (a) it is mostly admin-related and (b) there is still no agreed-upon Persian translation for most of those terms and a translation would only confuse admins.

PR #4832, opened 2017-09-06
Use casecmp() instead of casecmp?() for now

Jehops:
casecmp?() is only available in ruby 2.4.0.  Users running earlier ruby versions
would see errors, e.g., running
RAILS_ENV=production rails mastodon:maintenance:remove_deprecated_preview_cards.

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -47,7 +47,7 @@ namespace :mastodon do
     confirm = STDIN.gets.chomp
     puts
 
-    if confirm.casecmp('y')
+    if confirm.casecmp('y').zero?
       password = SecureRandom.hex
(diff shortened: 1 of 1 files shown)


ykzts (dismissed):

lib/tasks/mastodon.rake
```diff
@@ -47,7 +47,7 @@ namespace :mastodon do
     confirm = STDIN.gets.chomp
     puts
 
-    if confirm.casecmp?('y')
+    if confirm.casecmp('y')
```
Gargron: Correct me if I'm wrong but casecmp returns -1, 0 or 1. You should append `zero?` method for proper comparison

ykzts: Certainly. Need `zero?`

ykzts (approved):

PR #4840, opened 2017-09-07
Fetch statuses/following/followers numbers from ActivityPub collections

Gargron:


unarist:
What happens for those numbers when we receive new status / follow / followed?

unarist:
Hmm, when we receive new status, follow them or followed by them, those numbers will be incremented from fetched value, but it will synced next profile fetching again. I see.

unarist (approved):

nightpool:
I don't think this PR works.... we haven't been able to reproduce the correct behavior in any 1.6 instance. see https://cybre.space/@chr/3529727

Gargron:
@nightpool There's very aggressive caching on this, and it can only happen after the account is recognized as ActivityPub (so it's behind webfinger caching as well), and it's rather intransparent to the end-user. I tested it with a fresh account during development and it worked.

PR #4841, opened 2017-09-07
Fix language filter codes

Gargron:
CLD3 returns BCP-47 language identifier, filter settings expect
identifiers in the ISO 639-1 format. Convert between formats,
and exclude duplicate languages from filter choices (zh-CN->zh)


app/helpers/settings_helper.rb
```diff
@@ -30,6 +30,7 @@ module SettingsHelper
     th: 'ภาษาไทย',
     tr: 'Türkçe',
     uk: 'Українська',
+    zh: '简体中文',
```
ykzts: `中文`?

ykzts (approved):

PR #4846, opened 2017-09-07
Add follow suggestions

Gargron:
- GET /api/v1/accounts/suggestions returns triadic closures for authorized account
- Web UI column that displays the results from that API

Currently, the column is not linked from anywhere, but is available under /web/suggestions


beatrix-bitrot:
@Gargron can you write a little bit about this feature? Like... 

- what problem it solves
- how it works 
  - from a user point of view
  - from a technical point of view

it would make it much easier to review and get feedback from existing users/admins I think

Gargron:
@beatrix-bitrot Were you around when Neo4J was a dependency in Mastodon? This re-implemented the follow suggestions feature from back then but using only Postgres. It's not shown anywhere yet though, exactly because I'm not sure how to best present this to the end-user. So it's just a route you can open to get a generic accounts list.

beatrix-bitrot:
@Gargron i don't remember any such thing, but i also wonder not only about how to present the suggestions to the user but who will be suggested, on what basis, and whether they have any choice in the matter

Gargron:
[Triadic closures](https://dzone.com/articles/triadic-closures-are-the-new-black)

Gargron:
Check this e21a3fe0cd91dfeae76bce4d58c7611e78b60fbf

Gargron:
Some comments from the time of removal: https://github.com/nclm/mastodon/pull/3#commitcomment-19964258

beatrix-bitrot:
i see

i'd still appreciate it if you could write some stuff here in the comments in plain language about that so others can understand without reading commit history and articles about graph theory

Gargron:
The triadic closures method has been added in 017350e0ead4715ef4180eb7b1f6fecb67fbf554:

https://github.com/tootsuite/mastodon/blob/master/app/models/account.rb#L183-L209

First it selects the accounts the target user is following (set A). Then it selects accounts *those* accounts are following (set B), excluding set A and excluding users the target user is blocking/muting and self. Set B is grouped and ordered by overlap, e.g. if a lot of people from set A are following a person in set B, that person is higher up in the results.

beatrix-bitrot:
for me to feel ok with this feature (for my users and friends) I think I would want two things

1. users should opt-in to being suggested to others
2. users should be able to permanently dismiss suggestions

there may be more, but that is what is on my mind right now

joyeusenoelle:
+1 to bea's thoughts and an added thought of my own: the algorithm **must** not be able to recommend locked users unless they explicitly opt in.

Chronister:
+1 to the above. Locked accounts should never be recommended. I'm also cautious about adopting a purely popularity-based recommendation approach, as a small instance's idea of a "popular" user could be very different from a large instance, and popularity among a particular group doesn't necessarily signify willingness to be recommended to strangers.

Mastodon as a platform has marketed itself as approaching social media in a much more consent-driven way than centralized platforms and I think this feature has to be careful not to crib too much from them if it means possibly detracting from that approach.

vahnj:
I don't agree with generic recommendations nor tying it to other features- please separate it out from being private and make it a separate setting

TheInventrix:
This is a decent start at the feature, but I have a few thoughts on it.

bea's remark that it should be opt-in is, I think, very important, although I'm not entirely sure how to draw attention to the opt-in choice. Placing it in the same part of the settings as your user name, bio, and default post distribution setting seems the most obvious choice.

As someone with a "locked" account who does so simply to filter/view new followers, however, I disagree that such accounts should be excluded from the suggestions feature entirely. One of the benefits of Mastodon's approach to account and post privacy is that I can have my account set to approve followers _without_ having to completely lock away my account from the rest of the world; hardcoding locked accounts out of the suggestion feature is imo a step backwards.

For a more complicated suggestion: I'm not familiar enough with the back-end to know how doable this is, but what about instead of the weighting in the list being by account following overlap, basing it on the number of recent interactions of the people you follow with the accounts in question?

e.g. Account A is followed by 10 people you follow, but has been mentioned only twice by one of those people in the past week. Account B is followed by 5 people you follow, but has been mentioned fifteen times by those people in the past week. And Account C is followed by 2 people you follow, but hasn't been mentioned by any of them at all in the past week.

The ordering in the suggestions list would thus be: Account B, Account A, Account C

I imagine this is MUCH harder to implement and so if it was put in at all, it would be significantly further down the road, so I'm mostly mentioning it as food for thought on the feature. It has the benefits of avoiding the snowball effect of people who have more followers getting more followers, as well as boosting the suggestion of people who your friends actually _interact_ with and thus who you're more likely to be interested in.

jennifermf:
I would prefer this be opt-in. Actually, I'd prefer this not be implemented at all. One of the annoyances of other social networks is the need to promote a handful of Thought Leaders (prolific posters, or very popular posters with an almost cultlike following), which drowns out the quieter voices-- those tend to be marginalized people. 

I'm all for finding ways for people to find/follow people who want to be found and followed. I'd just like to see it in a less invasive way that's opt-in.

wxl:
I think I normally tend to give you the benefit of the doubt Eugen but this is not something I'm a fan of. The social network should do the socializing. Without suggestions and actually without knowing a lot of people, I managed to pool together a very pleasant timeline, based on the same general idea, though mostly I went by quality of content rather than just a mere relationships. I have IRL friends whose friends I have little interest in. If anything, this might be a little hidden feature, but certainly not something I want popping up all the time. The only thing worse than people spying on you for exploitative purposes is feeling kind you're being spied on.

lifning:
+1 to wxl's and Jennifer's sentiments. other social networks try to tell me what i might want, and 10 times out of 10 it's extremely uncomfortable. i don't think Mastodon should resemble that... and i don't think adding a snowball effect to encourage personality cults and "celebrities" would be healthy for the more equitable vibe the current community seems to have (at least in my neck of the fedi-woods)

Twitter, for instance, is a different site with a different focus and motives to manipulate user behavior in a way that doesn't necessarily benefit the user (encourage addiction, for one). it's quite possibly an antipattern to mimic their feature set Just Because.

Aldarone:
@Gargron > You should have included a screenshot of it.

From looking at the code it's pretty clear that the suggestions are far from being « shoved in our face at all time » or « popping up all the time » but in a new column, with a link to it from the « Getting Started » one.

I'm pretty hyped by this feature but I think that it should at least exclude protected accounts.

Perhaps if you want to add some configuration toggle my take on this is : 

- opt-out for public accounts
- opt-in for private account
- put the « hide from follow suggestion » checkbox right below the « protect account » one.

But it may be too hard of a job since the local instance would need to constantly refresh every suggested profiles to see if they're still hidden or not ?

Cassolotl:
There are people here saying "I don't think we need this because I managed to find a perfectly good bunch of people to follow on my own." I wanted to briefly comment that we're not going to hear from anyone else because the people who were not savvy enough to find good people to follow on their own *left Mastodon* and are not invested enough in it to comment on Github pull requests.

I started hand-holding new Mastodon members I referred. If I invited someone to Mastodon and they said "alright cool I'll give it a go" and then I left them to it, *they left Mastodon*. I now talk them through finding new followers and people to follow. I might recommend a few people based on how the new person tends to post and retweet on Twitter, and I will probably talk them through writing an `#introductions` post full of hashtags, at the very least. (I don't recommend that bridge thing because it's broken.) If they continue to ask questions I answer them or I might send them a link to an FAQ post and a tutorial post that I wrote on my blog, and that helps too. People I do this with tend to stick around, and people I ignore tend to leave after posting one or two times. Those who tell me why they leave usually say "I couldn't really find anyone to follow." There have been lots of people who've asked me Mastodon how-do-I questions via not-Mastodon (like Telegram, Twitter, etc) several days after I stopped holding their hands.

So yeah, something like this will be a big comfort to new people who want to join in but don't really know how and don't have a helpful human to hold their hand.

~

I agree that it should be opt-in in settings.

I don't think it should exclude locked accounts - I think locked accounts should be able to opt-in like everyone else, and if it's opt-*in* and not opt-*out* then locked accounts will be excluded automatically like everyone else.

I agree that follower count isn't necessarily a good indicator of who's good to follow, but I also think tendency to reply and be replied to isn't necessarily an indicator either, so I think including a number of factors is a good idea.

Gimmikku:
I think I shall throw my own unwarranted two cents in here.

I won't go through the same things that have already been repeated on here, but I will say that the problem that @Cassolotl has described is an issue which is a "cultural" problem with the Mastodon userbase. Implementing a technical solution to what is a cultural or behavioural phenomenon in the community is, in my opinion, a bad idea. IMO, software shouldn't be designed around "correcting" behaviour like this. I'm aware that major sites like Twitter, Facebook and Tumblr all cater content to increase user engagement, but the point of that engagement is, at the end of the day, for purposes that are very different from Mastodon's. Without such a specific purpose in mind, I'm not sure how this approach (technical solutions to encourage or discourage certain kinds of behaviour) would find its limits. It's not the burden of development to suggest how Mastodon users should engage with eachother. Such a method is the behaviour of websites which have an vested interest in doing so; Mastodon, as far as I'm aware, doesn't.

Gargron:
I am closing this because I don't have the energy to have another long-winded argument à la new landing page design. But a lot of the commenters here have clearly not read what this PR is and does (for example, the OP clearly states "this column is not linked from anywhere"). And by the way, when I asked what feature was preventing most people from switching to Mastodon - I got like 20 responses about content discovery being hard.

joyeusenoelle:
I don't think there needs to be a long-winded argument about this. Most of the comments so far boil down to:

1. I don't want to see these recommendations unless I ask for them
2. I don't want to be recommended unless I ask to be.
3. I want to be able to dismiss a *particular* recommendation permanently.
 plus one more I'd add:
4. I, an instance owner, want to be able to disable this wholesale. 

1 is addressed directly in the PR: a user won't see these recommendations without going to /web/suggestions. Suggestions don't pop up unwanted and the column can be dismissed at any time.

2 is easily addressed by making the feature opt-in in settings, but that runs into a bootstrapping problem: if everyone is out of the system to start with, nobody can be suggested. This can be solved with a modal that pops up the first time someone loads in on an instance that's implemented this feature:

    "We've added a new system that allows users to discover other users to follow. Would you like to be discoverable? (y/n) You can change this at any time in Settings."

3 just needs an additional column in the user table listing which users should be excluded from the suggestions. (Bonus points if the list can be reviewed and modified in Settings.)

4 avoids complaints from instances that advertise "no algorithms". It's perfectly reasonable to have this be a default-on, disable-if-desired feature, though.

The "cultural vs. technological" argument is, I think, a non-starter; if the *user problem* is "content discovery is hard" and the *cultural behavior* is "we don't actively want to help new users discover content" then the *responsibility of the developer* who wants to retain new users is to work on a technological solution to the *user problem* - especially since the developer, **by design**, doesn't have influence over the overall culture or specific microcultures (i.e. individual instances)!

If we can address 2, 3, and 4, I think we're good to go. (1 is already addressed!)

Gargron:
>3 just needs an additional column in the user table listing which users should be excluded from the suggestions. (Bonus points if the list can be reviewed and modified in Settings.)

This would only work for local users. Such preferences are between the user and the instance, not between instances. Even if that wasn't a problem, it would still complicate things. Do you opt out of the final results, or as a node in the graph? But most importantly... follow information and profiles themselves have always been public. The whole "algorithm" is equal to you scouting the following pages of the people you've already followed, it's just more convenient.

I'm also not sure why exactly appearing in recommendations is a problem for some. You appear in username autocompletes and search. If your account is locked, appearing in either of those things doesn't affect you in any tangible way..? If you want to avoid everything social and everything network in a social network, that's very confusing to me.

Also, if you are already an avid Mastodon user, you're kind of disqualified from this question. You've obviously already found people you like seeing content from, you're already "in". So yes, *you* don't need any follow recommendations. Instead you should put yourself in the shoes of a brand new user, who comes in, doesn't know anybody, maybe one person, has an empty home timeline. And if you believe that Mastodon should not be getting any new users who stay, well, you're entitled to your opinion but you are clearly dooming the project and dooming the project is not a direction I want to go in.

SoniEx2:
Let's think about this from a technical point of view:

1. USE THE API.

This can be done trivially with an external tool, using the API. So use it. Don't bother with opt-in, private accounts, etc, just use the API. This solves everything and doesn't add deliberate restrictions which only waste CPU time.

If you can do something with the API, then doing it built-in should be equivalent, not deliberately restricted.

joyeusenoelle:
> Such preferences are between the user and the instance, not between instances.

I think you've misunderstood the request, or I have. Here's the user story:

1. User A visits their suggestions and sees a person B whom they *know* they don't want to follow.
2. A clicks an "x" next to B's username in the suggestions list.
3. Mastodon adds B's user ID to a column ('dontsuggest') in A's user listing.
4. The next time A visits their suggestions, Mastodon checks A's user listing for the dontsuggest cell, finds B's user ID, and excludes B from A's (and only A's) suggestion results.

If the algorithm is suggesting B to A, then the local instance already knows about B, and so can use its local user ID in dontsuggest. Whether B is a local or remote user is almost irrelevant, and A's request to not see B in their suggestions doesn't affect anyone else.

> Also, if you are already an avid Mastodon user, you're kind of disqualified from this question. 

Not really. Your *intended* use case is for brand-new users looking for people to follow, but you've been in software development long enough to know that the intended use case isn't the *only* use case. Personally, I'd enjoy having a way to find new people to follow without having to manually troll my friends' conversations or rely on my friends boosting their friends' toots. I don't *need* the feature, but I would like to have it and I'd use it.

> if you believe that Mastodon should not be getting any new users who stay

I am genuinely struggling to figure out how on earth you read my comment to even come close to implying that.

All of that said:

> USE THE API.

If this can be trivially done by a third-party program then I agree that there's no reason at all to artificially limit the Mastodon core. So I'm down to two features I'd like to see:

1. Allow me to remove a user from the list of suggestions presented to me, as described above.
2. Allow instance owners to disable the algorithm on a per-instance basis.

Gargron:
@joyeusenoelle My reply wasn't to you but to everyone in this thread and the part you quoted first refers to point 2, not 3 (okay maybe I quoted the wrong thing in mine but the fact remains opting-out is not a thing done by adding a database field)

joyeusenoelle:
Okay, I understand.

SoniEx2:
PS: If anyone could make an external service using the existing API, please let me know, because I wanna use it.

PR #4853, opened 2017-09-08
Add script to make embedded iframes autosize

### LINKED ISSUE #1065: Embed posts in external websites
Would be nice to have twitter-like embedding of posts :slightly_smiling_face:

See this: https://mastodon.social/users/prashere/updates/619788

Gargron:
Fix #1065 

How it works:

Embed code includes script (embed.js). Embed.js iterates over all iframes with class "mastodon-embed" on the page. For each, it collects them in an array and assigns them IDs (position in array). It then attaches an on load callback to the iframe: when the iframe loads, embed.js sends a message (postMessage) to the iframe, asking it for height, and giving it its ID. JavaScript inside the iframe listens for these messages and answers them by sending a message to the parent window with the height and the ID it received. Embed.js listens for those messages and sets the height on the iframe with the given ID.

This way the height on the iframe is set correctly even if there are multiple iframes on the page. Each receives its own correct height.

### DIFF
--- a/app/javascript/mastodon/features/ui/components/embed_modal.js
+++ b/app/javascript/mastodon/features/ui/components/embed_modal.js
@@ -33,7 +33,8 @@ export default class EmbedModal extends ImmutablePureComponent {
       iframeDocument.close();
 
       iframeDocument.body.style.margin = 0;
-      this.iframe.height = iframeDocument.body.scrollHeight + 'px';
+      this.iframe.width  = iframeDocument.body.scrollWidth;
+      this.iframe.height = iframeDocument.body.scrollHeight;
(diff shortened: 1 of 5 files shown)


lynlynlynx:
I think `scrolling: 'no'` and `overflow: hidden` options should not be set for environments where loading scripts is not permitted(e.g. free blogs), or when the script is not loaded for some reason.
I tested this pull request and autosize works without these two options.

And one more..
Is there any reason to set `frameborder: 0` and `border: 0` options together? Embed codes get too long 😅

Gargron:
@lynlynlynx Okay now?

PR #4854, opened 2017-09-08
Handle stream_entry URL correctly in ActivityPub

unarist:
In before, the method uses stream_entry id as status id, so replied status was wrongly selected.

This PR uses StatusFinder which was introduced with `Api::Web::EmbedsController`.

### DIFF
--- a/app/lib/activitypub/tag_manager.rb
+++ b/app/lib/activitypub/tag_manager.rb
@@ -96,12 +96,14 @@ class ActivityPub::TagManager
       when 'Account'
         klass.find_local(uri_to_local_id(uri, :username))
       else
-        klass.find_by(id: uri_to_local_id(uri))
+        StatusFinder.new(uri).status
       end
     elsif ::TagManager.instance.local_id?(uri)
       klass.find_by(id: ::TagManager.instance.unique_tag_to_local_id(uri, klass.to_s))
     else
       klass.find_by(uri: uri.split('#').first)
     end
+  rescue ActiveRecord::RecordNotFound
+    nil
   end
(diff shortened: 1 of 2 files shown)


app/lib/activitypub/tag_manager.rb
```diff
@@ -96,12 +96,14 @@ def uri_to_resource(uri, klass)
       when 'Account'
         klass.find_local(uri_to_local_id(uri, :username))
       else
-        klass.find_by(id: uri_to_local_id(uri))
+        StatusFinder.new(uri).status
```
nightpool: i'm not sure this is the correct behavior here. At the very least, it should be explicitly checking for `when 'Status'` before using StatusFinder

Gargron: @nightpool To be fair we don't parse any URIs which are not Account or Status.

app/lib/activitypub/tag_manager.rb
```diff
@@ -85,7 +85,7 @@ def local_uri?(uri)
     !host.nil? && (::TagManager.instance.local_domain?(host) || ::TagManager.instance.web_domain?(host))
   end
 
-  def uri_to_local_id(uri, param = :id)
+  def uri_to_local_id(uri, param)
```
Gargron: Why did you change this?

unarist: Probably I thought `:id` is Status specific, and it shouldn't be defaulted. but now it looks generic term. I've removed this change.

Gargron (approved):

PR #4859, opened 2017-09-08
Temporarily pause timeline if there's been recent mouse movement

### LINKED ISSUE #1004: Temporarily pause timeline updates if there's been recent mouse movement
One of my biggest annoyances with TweetDeck is that things will often move out from under my mouse in between when I see it and when I've moved my pointer to interact with it. So, I end up targeting messages other than what I intended.

What I'd like to do is add some code that temporarily pauses changes to the timeline if there's been recent mouse movement over it. Not sure how long "temporarily" means, but my intent would be that things don't move while I'm in the middle of an intentional interaction.

Submitting this issue, but I hope to actually submit a PR myself once I've gotten deeper into the Mastodon stack.

Gargron:
Fix #1004

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -27,6 +27,10 @@ export default class ScrollableList extends PureComponent {
     trackScroll: true,
   };
 
+  state = {
+    lastMouseMove: null,
+  };
+
   intersectionObserverWrapper = new IntersectionObserverWrapper();
 
   handleScroll = throttle(() => {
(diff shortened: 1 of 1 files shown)


ykzts (approved):

lmorchard:
Oh hey, I filed that issue! Thank you 🎉

unarist:
I'm not sure, but I encountered weird behavior after this update.

* Open home, Scroll down, Click someone's avater, then account timeline appears and *jumps to scroll position in home`
* When page loaded (or first column loading), the timeline is already bit scrolled down

Gargron:
Could the old scroll position variable have been leaking between re-mounted components this entire time? Maybe that became more apparent when ScrollableList was abstracted and re-used in multiple places...

PR #4865, opened 2017-09-09
Disable mouse-based pause from #4859

Gargron:
It wasn't working ideally and introduced some annoying false positivies

Sorry @lmorchard, looks like not in 1.6 :cry:


unarist (approved):

lmorchard:
S'ok, just happy to see progress since I've totally not gotten around to this myself 😀

PR #4868, opened 2017-09-09
Set fallback address when empty notification address

ykzts:
Resolve https://github.com/tootsuite/mastodon/pull/3756#issuecomment-328296157


aral:
Thank you :)

Gargron (approved):

nightpool:
As a followup to this, is there something we can do to avoid requiring a
base secret when doing asset compilation?

Issue #4870, opened 2017-09-09
new user experience


nightpool:
We've been talking a lot (in different issues and on the network) about the experience of brand new users joining Mastodon, so I made this issue to summarize some of the current problems, and also the current solutions we have.

My hope is that a more holistic approach to this problem can give us a better understanding of what things are likely to help or not, and generate better ideas then just tackling individual parts in isolation.

On Mastodon, acquiring a new user requires four discrete steps.

1. **Finding an instance**

   Or, more specifically, finding an instance that fits the individual user, has a local community they'd like, etc. Currently both https://joinmastodon.com and https://instances.social attempt to solve this problem, but they have a lot of steps for users to take, don't have a lot of ways to differentiate between different instances, and in general just leave people more confused then before.

2. **Signing up for an account**

   This is classical onboarding. The current onboarding clickthrough is a good start, but it would be great if it was more interactive, and explained more parts of the software as you used it. For example, it would be great if it walked you through making your "first toot"—which could also serve as a way of helping others find you and interact with you, if it's appropriately tagged.

3. **Content discovery**

   Once we get users *on the site* how do they find enough people to follow that they're going to experience the software fully and enjoy the experience? For example, tumblr asks you to select 5 or so "categories" of things you're interested in and auto-populates your timeline with interesting posts from those tags. Facebook does contact voodoo magic to find your exact clones and suggests those. Something inbetween those would probably be the best for us.

4. \~\~**Retention**~~

   Once (if?) users have had a good experience, how do we help them remember to continue to use the site, instead of just forgetting about it? Notifications emails definitely help here. We could probably also do a better job about suggesting mobile apps, and maybe doing digest emails if you're not very active. 

anyway I don't think every suggestion here has to solve all 4 things at once, but it seems nice to get all of the current thinking about this stuff on the same page.

nightpool:
Two ideas I have currently:

 - a stumbleupon or web ring like site for finding instances. Some sort of frame-like interface where users are *actually on the front page* of instances, but with a little banner that says like "This is one instance out of many! This instance is run by BLANK, and you can read more about it or register below. Or click this button to view another random instance!"

   the benefits to this is that it reduces the friction between *finding* a server and *registering* for it (steps 1+2) and it gives confused users an easy-out (just sign up!) while not obscuring the process for people who want more choice (the stumbleupon interface is pretty good at this iirc)

[...]

PR #4871, opened 2017-09-09
Default follows for new users

Gargron:
When a new user confirms their e-mail, bootstrap their home timeline
by automatically following a set of accounts. By default, all local
admin accounts (that are unlocked). Can be customized by new admin
setting (comma-separated usernames, local and unlocked only)

Idea from #4870

### DIFF
--- a/app/services/bootstrap_timeline_service.rb
+++ b/app/services/bootstrap_timeline_service.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class BootstrapTimelineService < BaseService
+  def call(source_account)
+    bootstrap_timeline_accounts.each do |target_account|
+      FollowService.new.call(source_account, target_account)
+    end
+  end
+
+  private
+
+  def bootstrap_timeline_accounts
+    return @bootstrap_timeline_accounts if defined?(@bootstrap_timeline_accounts)
+
+    @bootstrap_timeline_accounts = bootstrap_timeline_accounts_usernames.empty? ? admin_accounts : local_unlocked_accounts(bootstrap_timeline_accounts_usernames)
+  end
+
+  def bootstrap_timeline_accounts_usernames
+    @bootstrap_timeline_accounts_usernames ||= (Setting.bootstrap_timeline_accounts || '').split(',').map { |str| str.strip.gsub(/\A@/, '') }.reject(&:blank?)
+  end
+
+  def admin_accounts
+    User.admins
+        .includes(:account)
+        .where(accounts: { locked: false })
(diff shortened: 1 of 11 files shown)


app/services/bootstrap_timeline_service.rb
```diff
@@ -0,0 +1,34 @@
...
+  private
+
+  def bootstrap_timeline_accounts
+    return @bootstrap_timeline_accounts if defined?(@bootstrap_timeline_accounts)
+
+    @bootstrap_timeline_accounts = bootstrap_timeline_accounts_usernames.empty? ? admin_accounts : local_unlocked_accounts(bootstrap_timeline_accounts_usernames)
+  end
+
+  def bootstrap_timeline_accounts_usernames
+    @bootstrap_timeline_accounts_usernames ||= (Setting.bootstrap_timeline_accounts || '').split(',').map(&:strip).reject(&:blank?)
```
nightpool: people are going to put @s in here and get confused.

config/locales/en.yml
```diff
@@ -167,6 +167,9 @@ en:
       unresolved: Unresolved
       view: View
     settings:
+      bootstrap_timeline_accounts:
+        desc_html: Separate multiple usernames by comma. Only local and unlocked accounts will work. Default when empty is all local admins.
```
beatrix-bitrot: maybe include an example or specify whether @ are required before usernames?

nightpool:
Code LGTM but I want to test this and make sure there's no "gap" where you'll see an empty timeline.

nightpool:
jesus github has some unintuitive buttons.

app/services/bootstrap_timeline_service.rb
Gargron: Riiiight but the "contact username" has been a setting since forever and also doesn't require a @ so I don't think this is a problem..?

beatrix-bitrot: ok but it's really easy to just leave a short message about it so maybe do that?

nightpool (approved):
LGTM with followup if it becomes a pain point.

app/services/bootstrap_timeline_service.rb
Gargron: But it's so prettily one-line right now 😢 I really don't think admins would be confused?

hakaba-hitoyo:
> Can be customized by new admin setting (comma-separated usernames, local and unlocked only)

I become more happy if some _remote_ users are listed in the automatic following list. Do you plan it in your future work?

PR #4873, opened 2017-09-09
Hide modal loading screen for media/video modals

Gargron:


ykzts (approved):

unarist:
Hmm, only media/video modals? For example, MediaLoading also looks different from confirm/boost modals, so the loading screen confuses me...

Also, maybe we don't need to have a separate chunk for it, like actions modal? https://github.com/tootsuite/mastodon/pull/4295#discussion_r129925002

PR #4897, opened 2017-09-11
Support OpenGraph video embeds

### LINKED ISSUE #135: Better OpenGraph support
Is there any good reason why we [disallow OpenGraphs from pages with no title](https://github.com/glitch-soc/mastodon/blob/1618b68bfa740ed655ac45d7d5f4f46fed6c8c62/app/services/fetch_link_card_service.rb#L98)? This seems kinda silly to me.

Also, note that a page with `<meta property="og:title" content="" />` (for example, an untitled blog post) will *not* use the page `<title>` instead, since empty strings are truthy in Ruby.

Finally, there's no reason not to at least use `site_name` in place of OEmbed's `provider_name` afaict.

Gargron:
It's not really OpenGraph, it's twitter:player property, but it's not OEmbed so that fits. For example, this allows Twitch clips to be displayed as embeds.

Also, fixes glitch-soc/mastodon#135 by making sure empty og:title falls back to page title properly.

**Note:** It seems like embedded Twitch clips autoplay. But cards are only displayed in detailed toot view, so this might be desired?

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class FetchLinkCardService < BaseService
+  include ActionView::Helpers::TagHelper
+
   URL_PATTERN = %r{https?://\S+}
 
   def call(status)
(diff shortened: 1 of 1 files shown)


nightpool:
It looks like twitter doesn't show the embed by default (even in the expanded view). They show a preview of the video card using the normal og:title, og:image, etc, with a prominent "play" button.

image

How does oembed work? does it generally autoplay?

we can special case twitch here by using `autoplay=false` but that's probably not a good idea.

Gargron:
I think this is fine to do. twitter:player is not a common property, and this *might* be desired behaviour. We can always fix later.

unarist (approved):

haywirez:
I think the play button should be *way* more prominent and contrasted to the background in the default styles - I didn't even realize the play option showing up when testing. 

Have a look, the current situation:
image

vs. something akin to the following CSS rules: 

```
{
  color: #ffffff;
  font-size: 64px;  
  mix-blend-mode: difference;
}
```

image

haywirez:
Also, I believe instance owners and users should be able to set whether the play button is gating the loading of the iframe by default, or have the option to set a blacklist/whitelist. This would greatly favor the profileration of decentralized content, artists hosting their own audio/video players for example.

Issue #4895, opened 2017-09-11
Toots sometimes missing from the home TL


ThibG:
My home TL is occasionally missing toots from people I follow. This seems to only affect replies (addressed to me or to someone I follow). It also seems to only affect users federating through ActivityPub.

My current hunch is that:
1. The toot is created immediately upon reception, with the `reply` field set to `true`, but without a `in_reply_to` id (cf. app/lib/activitypub/activity/create.rb:33)
2. Distribution and thread resolving are scheduled asynchronously
3. Distribution gets called before the thread could be resolved, thus resulting in a filtered toot, as the `reply` field is set but not the actual thread (app/lib/feed_manager.rb:96)

EDIT: confirmed that at app/lib/activitypub/activity/create.rb:17, `status.reply?` is true, but `replied_to_status` is nil. No idea why yet.

ThibG:
Ok, finally got it! My reasoning above is true, except in the first step, the `in_reply_to` field should have been set. But it is not set when the reply is to a status of the local instance, and that local instance has `LOCAL_DOMAIN` ≠ `WEB_DOMAIN`.

The reason for that is that `StatusFinder` will check that the URL is local, where local means “starts with `LOCAL_DOMAIN`”. In fact, it seems to me that for all intents and purposes in the current code, `TagManager.local_url?` is only used to check at Web Interface and API endpoints, and should thus check against `WEB_DOMAIN` and not `LOCAL_DOMAIN`.

PR #4904, opened 2017-09-12
Fix ActivityPub handling of replies with WEB_DOMAIN (#4895)

ClearlyClaire:
For all intents and purposes, `local_url?` is used to check if an URL refers
to the Web UI or the various API endpoints of the local instances. Those things
reside on `WEB_DOMAIN` and not `LOCAL_DOMAIN`.

### DIFF
--- a/app/lib/tag_manager.rb
+++ b/app/lib/tag_manager.rb
@@ -87,7 +87,7 @@ class TagManager
   def local_url?(url)
     uri    = Addressable::URI.parse(url).normalize
     domain = uri.host + (uri.port ? ":#{uri.port}" : '')
-    TagManager.instance.local_domain?(domain)
+    TagManager.instance.web_domain?(domain)
   end
 
   def uri_for(target)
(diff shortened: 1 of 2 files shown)


ThibG:
Tests fail because the spec is about `LOCAL_DOMAIN` and not `WEB_DOMAIN`. Not sure if the spec should be changed, the function changed to test against *both* `LOCAL_DOMAIN` and `WEB_DOMAIN`, or yet something else.

As said earlier, in practice, `local_url` is exclusively used for things that should be checked against `WEB_DOMAIN`.

ThibG:
Because all current usages of this function only make sense with WEB_DOMAIN. But I'm fine with both solutions, as I can't see any adverse effect of accepting LOCAL_DOMAIN here.

ThibG:
Actually, I can think of a (crazy) situation where that would matter! Let's have two domains, A and B, associated with two instances, one of which has WEB_DOMAIN=A and LOCAL_DOMAIN=B while the other has WEB_DOMAIN=B and LOCAL_DOMAIN=A.

I'm not sure there are people crazy enough to do this, but I think there are some instance admins that could prove me wrong :]

ThibG:
Decided that the correct behavior for `local_url?` would be to use web_domain exclusively, as all URLs handled by Mastodon are by definition on `WEB_DOMAIN`.

Gargron (approved):

PR #4907, opened 2017-09-12
[WiP] Whenever a remote keypair changes, unfollow them and re-subscribe to …

ClearlyClaire:
…them

In Mastodon (it could be different for other OStatus or AP-enabled software),
a keypair change is indicative of whole user (or instance) data loss. In this
situation, the “new” user might be different, and almost certainly has an empty
followers list. In this case, Mastodon instances will disagree on follower
lists, leading to unreliable delivery and “shadow followers”, that is users
believed by a remote instance to be followers, without the affected user
knowing.

Drawbacks of this change are:
1. If an user legitimately changes public key for some reason without losing
   data (not possible in Mastodon at the moment), they will have their remote
   followers unsubscribed/re-subscribed needlessly.
2. Depending of the number of remote followers, this may generate quite some
   traffic.
3. If the user change is an attempt at usurpation, the remote followers will
   unknowingly follow the usurper. Note that this is *not* a change of
   behavior, Mastodon already behaves like that, although delivery might be
   unreliable, and the usurper would not have known the former user's
   followers.

### DIFF
--- a/app/workers/refollow_worker.rb
+++ b/app/workers/refollow_worker.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ResubscribeWorker
+class RefollowWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: 'pull', retry: false
--- a/app/workers/refollow_worker.rb
+++ b/app/workers/refollow_worker.rb
@@ -8,7 +8,7 @@ class RefollowWorker
   def perform(target_account_id)
     target_account = Account.find(target_account_id)
 
-    target_account.followers.where(domain: nil).each do |follower|
+    target_account.followers.where(domain: nil).find_each do |follower|
       # Locally unfollow remote account
       follower.unfollow!(target_account)
 
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -17,7 +17,9 @@ class ActivityPub::ProcessAccountService < BaseService
 
     create_account  if @account.nil?
     upgrade_account if @account.ostatus?
+    old_public_key = @account.public_key
     update_account
+    ResubscribeWorker.perform_async(@account.id) if old_public_key != @account.public_key
 
     @account
   rescue Oj::ParseError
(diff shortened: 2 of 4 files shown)


Gargron:
I suggest a terminology change - subscriptions have a WebSub/PubSubHubbub connotation, which is not used in ActivityPub. So maybe not ResubscribeService, but RefollowService or something like that.

ThibG:
Done.

A possible improvement would be to fetch the user's Followers collection (if accessible), and only re-issue follow requests for the local followers that aren't in the remote Followers collection. This would avoid the first drawback I listed imho.

app/workers/refollow_worker.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class RefollowWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'pull', retry: false
+
+  def perform(target_account_id)
+    target_account = Account.find(target_account_id)
+
+    target_account.followers.where(domain: nil).each do |follower|
```
nightpool: should be in batches

Gargron (approved):

Gargron:
>A possible improvement would be to fetch the user's Followers collection (if accessible)

Hardly an improvement given how much HTTP requests you'd have to do to process it all, so I think it's better without.

ThibG:
I've just discovered two issues with my code :(

1. The code is needlessly triggered by the discovery of a new account (key goes from `nil` to an actual value).
2. The processes might be called from a transaction (when processing an incoming toot), in which case there may be a race condition between the worker and the account update. I'm not sure how to fix this.

Gargron:
@ThibG The worker doesn't need to know any account data to re-follow, tbh. Seems like the only problem is when the worker is called for new accounts, which is the wrong behaviour in the first place.

ThibG:
@Gargron Hm… API endpoints might have changed, especially if we are in a instance brought down / back up scenario. Wouldn't that be an issue to re-follow the remote user?

Gargron:
@ThibG I *highly* doubt inbox URL would change. It would be a super rare case. Not worth worrying about I think.

ThibG:
@Gargron hm, ok, I was kinda thinking about an instance being wiped and users re-registered in a different order, leading to different numeric identifiers for different users

Gargron:
@ThibG ActivityPub URIs are not numeric! It's /users/:username/inbox etc

ThibG:
Oh! Nice! But it would still fail for a previously-OStatus-only instance. Maybe we should enable this behavior only for ActivityPub-enabled accounts?

Gargron:
Oh. Yeah. Just remove the OStatus bit.

PR #2268, opened 2017-04-21
Add preview card to public statuses

mikoim:
This PR makes preview card displayable in public statuses. It needs to brush up design because I just copied a style from React implement.

PR #4911, opened 2017-09-12
Redesign video player

Gargron:
Inline:

image

Detailed status:

image

Modal:

image

Public profile:

image

Public toot:

image

Also, images and gifv now also reuse the React components even on public pages:

image

Likewise for link preview cards, which resolves #2268


unarist:
- thumbnail -> expand -> fullscreen: works
- thumbnail -> fullscreen: empty fullscreen...?

Gargron:
@unarist Cannot confirm, works fine both ways (Firefox 55, Linux)

unarist:
Hmm, I've confirmed working on Edge and Firefox, but the issue occurs on Chrome 61 on Windows.

Also the player always fill the screen on Edge and Firefox, but it will be in original video size at a maximum on Chrome.

Gargron:
@unarist Okay, should be fixed now

Gargron:
I have also added Aria labels to the buttons on the player

unarist:
Sizing seems still prevent by max-width/height on `.media-modal video`:

https://github.com/tootsuite/mastodon/blob/283a5ec1a4b4c771e36e5ad50f6e7c425fcfc4f7/app/javascript/styles/components.scss#L3063-L3073

and empty fullscreen issue is still exists on Chrome. I also investigating this, but no hints yet...

Gargron:
@unarist Can I merge this? 😅

app/javascript/styles/components.scss
```diff
@@ -3798,6 +3806,182 @@ button.icon-button.active i.fa-retweet {
...
 }
 
+.video-player {
+  overflow: hidden;
+  position: relative;
+  background: $base-shadow-color;
+
+  &:-webkit-full-screen {
+    width: 100%;
+    height: 100%;
```
unarist: `!important` is needed to override inline styles. Firefox has those styles in UA stylesheet.

also this can be merged into `&.fullscreen`?

unarist (changes requested):

app/javascript/styles/components.scss
unarist: and add `margin: 0` to override `margin-top` on `.status .video-player`.

Gargron:
@unarist Done!

unarist (approved):

unarist:
BTW, currently I can't hide controls on fullscreen? Hiding it a little while after user stops mouse movement would helps I think.

Although, it's not a blocking issue, so fixing in subsequent PRs.

PR #4913, opened 2017-09-13
Uploads for admin site settings

Gargron:
First use case:

Uploading thumbnail for instance. Thumbnail used as OpenGraph image on /about and /about/more pages, as well as new thumbnail property on GET /api/v1/instance API

cc @DagAgren


app/controllers/about_controller.rb
```diff
@@ -35,4 +36,8 @@ def initial_state_params
       token: current_session&.token,
     }
   end
+
+  def set_thumbnail
+    @thumbnail = Rails.cache.fetch('site_uploads/thumbnail') { SiteUpload.find_by(var: 'thumbnail') }
```
nightpool: this should probably just use `@instance_presenter` instead of duplicating it.

nightpool (changes requested):

nightpool (approved):
LGTM!

PR #4926, opened 2017-09-13
fix share intent.

MitarashiDango:
Bug fix (Privacy settings modal does not work when touch operation.)


MitarashiDango:
Sorry, I made a mistake in correction, fixed for pull request.

Gargron (approved):

PR #4930, opened 2017-09-13
Fix race condition when receiving an ActivityPub Create multiple times

### LINKED ISSUE #4905: Remote status from followers with mentions to local users processed twice
Status with mentions to a local user are processed twice, leading to noisy logs, with stuff like “ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_statuses_on_uri"”.

Haven't investigated in-depth yet.

ClearlyClaire:
Fixes #4905. The issue is that a same Create activity can be delivered multiple times to the same instance, and will be processed in concurrent tasks. The time from status processing to commit in the database is relatively long, which makes things worse and a race condition more likely to occur.

This proposed looks up the toot locally if an ActiveRecord exception is raised by the transaction. The reasoning behind the `raise e unless status` pattern is that we want to catch issues arising from the race condition, and not other issues, as I fear `ActiveRecord::RecordInvalid` might be too broad and catch unrelated exceptions.

(Or of course we may also use locks, but I have no experience with these in RoR)

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -4,31 +4,30 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   def perform
     return if delete_arrived_first?(object_uri) || unsupported_object_type?
 
-    status = find_existing_status
-
-    return status unless status.nil?
-
-    begin
-      ApplicationRecord.transaction do
-        status = Status.create!(status_params)
-
-        process_tags(status)
-        process_attachments(status)
+    RedisLock.acquire(lock_options) do |lock|
+      if lock.acquired?
+        @status = find_existing_status
+        process_status if @status.nil?
       end
-    rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid => e
-      status = find_existing_status
-      raise e unless status
-      return status
     end
 
-    resolve_thread(status)
-    distribute(status)
-    forward_for_reply if status.public_visibility? || status.unlisted_visibility?
-
-    status
+    @status
   end
 
   private
+
+  def process_status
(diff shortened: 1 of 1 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -4,10 +4,17 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   def perform
     return if delete_arrived_first?(object_uri) || unsupported_object_type?
 
-    status = find_existing_status
+    RedisLock.acquire(lock_options) do |lock|
+      status = find_existing_status
```
Gargron: Gotta do something like `if lock.acquired?`

Gargron: I think `status` must be first defined outside of the block, otherwise it's scoped to within the block only. Same thing in `process_status` below.

Gargron (changes requested):

app/lib/activitypub/activity/create.rb
```diff
@@ -4,10 +4,19 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
...
 
-    return status unless status.nil?
+    @status
+  end
+
+  private
 
+  def process_status
     ApplicationRecord.transaction do
       status = Status.create!(status_params)
```
Gargron: Scoping issue here as well

Gargron (changes requested):

ThibG:
Sorry for the multiple issues, I planned to test it right away, but I'm having issues with my ISP at the moment. I might not be able to work on it much more today.

Gargron (approved):

Issue #4837, opened 2017-09-07
Don't normalize URIs in Unicode NFKC


unarist:
Mastodon normalizes URIs in the status by using Addressable, to NFKC.

https://github.com/tootsuite/mastodon/blob/85c7c42098b5a5b10033eb8de79f6cb5dce8d462/app/lib/formatter.rb#L102-L109

Well, what if the linked website has unnormalized URI? The link will be broken if we normalize it.

For example, we can't link this article in the toots (and it can be posted to Twitter and Facebook):

http://katsu2000x.hatenablog.com/entry/2017/03/17/android%E3%81%A7%E9%80%9A%E7%9F%A5%E3%81%8C%E6%9D%A5%E3%81%AA%E3%81%84%E4%BA%8B%E8%B1%A1%E3%82%92%E5%9B%9E%E9%81%BF%E3%81%99%E3%82%8B%EF%BC%88IPv6%E7%B7%A8%EF%BC%89_

Also I don't know why we need to normalize non user-input URIs, like [`hub_topic`](https://github.com/tootsuite/mastodon/blob/7d853b514a5b6d6405d2019e414b237e1e123571/app/controllers/api/push_controller.rb#L55).

Looks like [RFC3987](https://www.ietf.org/rfc/rfc3987.txt) mentioned this problem in sec. 5.3.2.2.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
>Also I don't know why we need to normalize non user-input URIs, like hub_topic.

Because our HTTP library is not able to perform requests on unicode URLs - at least I believe that was the original reasoningfor normalizing all URIs.

unarist:
Hmm...

I also found an upstream issue, but no progress 😕 
https://github.com/sporkmonger/addressable/issues/8

ThibG:
Should be fixed by #6134

PR #4941, opened 2017-09-14
Enable to recognize most kinds of characters as URL paths (Mainly for CJK)

fvh-P:
Related #4837 

Currently, only ASCII (and Cyrillic) characters and several symbols are recognized as URL paths and automatically linked. The behavior is by design of [`twitter-text`](https://github.com/twitter/twitter-text/tree/master/rb) gem. However, [`FetchLinkCardService`](https://github.com/tootsuite/mastodon/blob/1618b68bfa740ed655ac45d7d5f4f46fed6c8c62/app/services/fetch_link_card_service.rb#L4) recognizes non-ASCII paths as a part of the URL.

linkcard
In the above example, `twitter-text` does not recognize "アイドルマスター" as a URL path, but FetchLinkCardService recognizes it and the LinkCard was generated correctly.

This PR solves inconsistency between them. In particular, **this enables to recognize most kinds of characters (of course including Chinese, Japanese, Korean) as URL paths.** The reasons for this change are as follows.

 1. Nowadays, many Japanese websites are using Japanese characters (Hiragana, Katakana and Kanji) in their URL, and I think many Korean or Chinese websites are using their own characters.
2. Usually, **percent-encoding** is used for URLs containing non-ASCII string, but we can not understand the encoded URL. (Can you understand what the following URL refers to? https://ja.wikipedia.org/wiki/%E6%97%A5%E6%9C%AC)
3. Dividing URLs and normal sentences by white-space is not so strange even in the CJK area. 　Unfortunately `twitter-text` does not think so :( Read [Here](https://github.com/twitter/twitter-text/tree/master/rb#urls). I think the design that URLs are not correctly recognized as URLs is a larger problem than `twitter-text` said.

After this change, the URL contains Japanese characters is shown like the following image.

The problem that FetchLinkCardService recognizes multiple URLs separated by double-byte space as a single URL has also been resolved.

Also, this modification passes all existing and additional test cases.

If you can read Japanese, please also read [here](https://github.com/imas/mastodon/issues/57).
日本語が読める方は[こちら](https://github.com/imas/mastodon/issues/57)もお読みください。

- [x] Enable to recognize most kinds of characters as URL paths
- [x] Solve the problem in FetchLinkCardService
- [x] Add some test cases

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
 class FetchLinkCardService < BaseService
-  include ActionView::Helpers::TagHelper
-
-  URL_PATTERN = %r{https?://\S+}
+  URL_PATTERN = %r{
+    (                                                                                                 #   $1 URL
+      (https?:\/\/)?                                                                                  #   $2 Protocol (optional)
+      (#{Twitter::Regex[:valid_domain]})                                                              #   $3 Domain(s)
+      (?::(#{Twitter::Regex[:valid_port_number]}))?                                                   #   $4 Port number (optional)
+      (/#{Twitter::Regex[:valid_url_path]}*)?                                                         #   $5 URL Path and anchor
+      (\?#{Twitter::Regex[:valid_url_query_chars]}*#{Twitter::Regex[:valid_url_query_ending_chars]})? #   $6 Query String
+    )
+  }iox
 
   def call(status)
     @status = status
@@ -42,7 +48,7 @@ class FetchLinkCardService < BaseService
 
   def parse_urls
     if @status.local?
-      urls = @status.text.match(URL_PATTERN).to_a.map { |uri| Addressable::URI.parse(uri).normalize }
+      urls = @status.text.scan(URL_PATTERN).map { |array| Addressable::URI.parse(array[0]).normalize }
     else
       html  = Nokogiri::HTML(@status.text)
       links = html.css('a')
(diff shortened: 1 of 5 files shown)


config/initializers/twitter_regex.rb
```diff
@@ -0,0 +1,43 @@
+module Twitter
+  class Regex
+    require 'yaml'
```
Gargron: Wait, why are we requiring yaml here?

fvh-P: oh, just forgetting to erase it after several tries in another way.

Gargron (approved):
Looks good to me!

app/services/fetch_link_card_service.rb
```diff
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
 class FetchLinkCardService < BaseService
-  include ActionView::Helpers::TagHelper
-
-  URL_PATTERN = %r{https?://\S+}
+  URL_PATTERN = %r{
```
unarist: URL_PATTERN is also used in StatusLengthValidator, to treat all URLs as 23 characters length. Same regex is also used in WebUI side to calculate and display character count.

https://github.com/tootsuite/mastodon/blob/bdcc9e2ceb3a359e43775de9b2afd5b1e64eef4a/app/javascript/mastodon/features/compose/util/counter.js#L5

So if you want to change it, you also should change WebUI side too. WebUI allowed some status but server returns an error, will confuses user.

PR #4949, opened 2017-09-14
Use OrderedCollectionPage in followers/following list

nullkal:
Currently, `/users/:username/followers` and `/users/:username/following` with the request header `Accept: application/activity+json` (In other words, when you are trying to fetch them in an Activity Streams 2 object's form) return an `OrderedCollection` while they return a part of followers/following collection and it seems they should treat pagination properly through `OrderedCollectionPage`, so I'm trying to implement this modification. However, there are few available examples of `OrderedCollectionPage` and I'm not sure my understanding is correct. So, I want to hear the other people's opinion.

If we can fetch followers/following collection via Activity Streams 2, we can fetch all of the relationships of a remote account in the standard way, and show them in the web UI. I think it's important because the current behavior, in which we can see only the relationships that are concerned with a local account for remote account, is preventing the Mastodon users from remote following the remote accounts which are not in the federated timeline. 

c.f. https://mstdn.jp/@nullkal/40976989

### DIFF
--- a/app/controllers/follower_accounts_controller.rb
+++ b/app/controllers/follower_accounts_controller.rb
@@ -17,12 +17,29 @@ class FollowerAccountsController < ApplicationController
 
   private
 
+  def page_url(page)
+    account_followers_url(@account, page: page) unless page.nil?
+  end
+
   def collection_presenter
-    ActivityPub::CollectionPresenter.new(
-      id: account_followers_url(@account),
+    page = ActivityPub::CollectionPresenter.new(
+      id: account_followers_url(@account, page: params.fetch(:page, 1)),
       type: :ordered,
       size: @account.followers_count,
-      items: @follows.map { |f| ActivityPub::TagManager.instance.uri_for(f.account) }
+      items: @follows.map { |f| ActivityPub::TagManager.instance.uri_for(f.account) },
+      part_of: account_followers_url(@account),
+      next: page_url(@follows.next_page),
+      prev: page_url(@follows.prev_page)
     )
+    if params[:page].present?
+      page
+    else
+      ActivityPub::CollectionPresenter.new(
+        id: account_followers_url(@account),
+        type: :ordered,
+        size: @account.followers_count,
+        first: page
+      )
+    end
   end
 end
--- a/app/controllers/following_accounts_controller.rb
+++ b/app/controllers/following_accounts_controller.rb
@@ -17,12 +17,29 @@ class FollowingAccountsController < ApplicationController
 
   private
 
+  def page_url(page)
+    account_following_index_url(@account, page: page) unless page.nil?
+  end
+
   def collection_presenter
-    ActivityPub::CollectionPresenter.new(
-      id: account_following_index_url(@account),
+    page = ActivityPub::CollectionPresenter.new(
+      id: account_following_index_url(@account, page: params.fetch(:page, 1)),
       type: :ordered,
       size: @account.following_count,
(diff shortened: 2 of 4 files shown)


Gargron:
Summoning @cwebber and @evanminto, I am also not really sure how paginated collections are supposed to work 😅

akihikodaki (approved):
I approve this change. However I have some feedback which is not blocking:

1. I do not think `ActivityPub::CollectionPresenter` makes sense for the usage. Other serializers parses objects and directly converts into ActivityStreams representation. It is possible also for the collection by utilizing properties provided by Kaminari.
Note that such a serializer would be more dependent on the underlying object: we would need a more specialized serializer, as serializers for `Undo` activity are: `UndoAnnounceSerializer`, `UndoBlockSerializer`, `UndoFollowSerializer`, and `UndoLikeSerializer`. We would have `OrderedFollowerAccountCollectionPage` and `OrderedFollowerAccountCollection` or something sort of.

2. Why doesn't it have `startIndex`?

evanminto:
Wasn't able to totally get what the JSON looks like from glancing at this PR, but [ActivityStreams 2.0's paging model is specified here](https://www.w3.org/TR/activitystreams-core/#collections). There's a really good diagram in there. I'll try to take a look at this in more detail soon, but for now I can try to help explain the paging model.

Basically, when it's paged the Collection doesn't have any items itself. Instead, it has links to the beginning and end nodes of a linked list of CollectionPages: "first" and "last," respectively. It may also have a "current" link that points to the page with the most up-to-date items (I guess because the collection could be in any order, though I'm less sure about that property). Each CollectionPage, then, acts as a sort of mini-Collection with a subset of the total items, and a "next" and/or "prev" link to its adjacent pages. It should also have a "partOf" link back to the parent Collection. To my knowledge the model doesn't allow an exhaustive list of pages or a page count, though the top-level Collection may include a "totalItems" property referring to the items in all pages combined.

@cwebber, please correct me if I got any of that wrong.

akihikodaki:
This pull request is different from what @evanminto in the following terms:
* `first` property is an actual object rather than a link.
* It does not have `last` property.
* It does not have `current` property.

I think aligning the implementation to the suggestion makes sense; turning `first` property into a link and adding `last` and `current` property.

nullkal:
I refered to EXAMPLE 20 in the Activity Streams 2.0 specification (https://www.w3.org/TR/activitystreams-core/) to implement it tentatively, and it has the `first` property as an actual object and does not have neitiher `last` nor `curent` property.

akihikodaki:
Thanks for pointing out the specification. Still I think there could be use cases of `last` and `current`, and `first` is not necessary to be an actual object in those cases.

akihikodaki:
I should have noted that those differences are not blocking anyway; this pull request is ready to merge in my opinion.

nullkal:
I think the current way of pagenation (?page=n) has a problem that followers/following collection would have duplicates or slip-outs when some other user follows/unfollows the user when the user fetching all of the followers/following collection at the same time. It would be better switching to some other method like `?max_id=n`, which we currently use in statuses pagination.

akihikodaki:
Is this still WIP?

nullkal:
It is okay to merge, but I've never written any test about it. We'll have to write it.

akihikodaki:
Pending examples are enough (many contributors do not write even those xD)

evanminto:
Oh yeah, in JSON-LD pretty much everything can be either an object literal or a URL pointing to the object. I should have specified that when I said "link" before.

PR #4951, opened 2017-09-14
Add scheduled worker to purge old user IPs

### LINKED ISSUE #4886: Mastodon should properly follow its log retention policy
In mastodon's default terms of service, it says that the server operators will make a good-faith effort to 

- Retain server logs containing the IP address of all requests to this server no more than 90 days.
- Retain the IP addresses associated with registered users and their posts no more than 5 years.

However, Mastodon is not configured to follow these policies and doesn't provide guidance for admins on how to implement them. In particular

- The documentation for setting up Mastodon should include a section on configuring logrotate to delete old server logs.
- There should be a task that runs regularly to delete sessions that have been inactive for 5 years or more, and clear the `last_sign_in_ip` field generated by Devise's trackable. (I realize this is a weird corner case that only comes up when users log in very infrequently, but I still think this is important as a matter of principle. Also, my instance has a customized terms of service with shorter log retention times, so this feature would actually be useful for us.)

* * * *

duxovni:
Partially fixes #4886. We still need to update the deployment guides with info on setting up logrotate to purge old server logs.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -24,3 +24,6 @@
   subscriptions_cleanup_scheduler:
     cron: '2 2 * * 0'
     class: Scheduler::SubscriptionsCleanupScheduler
+  ip_cleanup_scheduler:
+    cron: '0 4 * * *'
+    class: Scheduler::IpCleanupScheduler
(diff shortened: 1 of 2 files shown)


config/sidekiq.yml
```diff
@@ -24,3 +24,6 @@
   subscriptions_cleanup_scheduler:
     cron: '2 2 * * 0'
     class: Scheduler::SubscriptionsCleanupScheduler
+  ip_cleanup_scheduler:
+    cron: '0 4 * * *'
```
Gargron: Does it need to run every night? Feels like once a week should be enough.

Gargron (approved):

PR #4954, opened 2017-09-14
When web UI URL used while logged out, redirect to static page

Gargron:


ykzts (approved):

nightpool:
We should make sure to only do this when the status is local, otherwise we
open ourselves up to being used as a trusted proxy to a malicious website.

Gargron:
@nightpool If people didn't use the web URLs, this wouldn't be needed. But they do, and they use it to refer to local *as well as* remote toots. Restricting redirect to local-only would make this fix essentially useless.

PR #4955, opened 2017-09-14
When accessing uncached media attachment, redownload it

Gargron:
- When local cache of media attachment is missing, API returns URL of media-proxy instead of static file. When this media-proxy is accessed, the cache is redownloaded from original URL (a lock ensures this only happens once). This is also not done for local or already downloaded attachments. Then the request redirects to the new static URL. The created_at of the media attachment is updated to prevent it to be swiped by mastodon:media:remove_remote again.
- No longer set attachments type to "unknown" during mastodon:media:remove_remote, this means apps can access the media-proxy transparently from image view/video player/etc without having to know the attachment had been uncached

### DIFF
--- a/app/controllers/media_proxy_controller.rb
+++ b/app/controllers/media_proxy_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class MediaProxyController < ApplicationController
+  include RoutingHelper
+
+  def show
+    RedisLock.acquire(lock_options) do |lock|
+      if lock.acquired?
+        @media_attachment = MediaAttachment.remote.find(params[:id])
+        redownload! if @media_attachment.needs_redownload?
+      end
+    end
+
+    redirect_to full_asset_url(@media_attachment.file.url(version))
+  end
+
+  private
+
+  def redownload!
+    @media_attachment.file_remote_url = @media_attachment.remote_url
+    @media_attachment.touch(:created_at)
+    @media_attachment.save!
+  end
+
+  def version
+    if request.path.ends_with?('/small')
+      :small
+    else
+      :original
+    end
+  end
+
+  def lock_options
+    { redis: Redis.current, key: "media_download:#{params[:id]}" }
+  end
+end
--- a/app/controllers/media_proxy_controller.rb
+++ b/app/controllers/media_proxy_controller.rb
@@ -7,7 +7,7 @@ class MediaProxyController < ApplicationController
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
         @media_attachment = MediaAttachment.remote.find(params[:id])
-        redownload! if @media_attachment.needs_redownload?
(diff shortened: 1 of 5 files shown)


app/controllers/media_proxy_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class MediaProxyController < ApplicationController
+  include RoutingHelper
+
+  def show
+    RedisLock.acquire(lock_options) do |lock|
+      if lock.acquired?
+        @media_attachment = MediaAttachment.remote.find(params[:id])
+        redownload! if @media_attachment.needs_redownload?
```
Gargron: Uh, need to check existence of domain block with reject_media here, I think

beatrix-bitrot (approved):
hhhhhrnnnnnnngghh!

Komic:
should pretty much close https://github.com/tootsuite/mastodon/issues/4491? I can't test right now

Gargron:
@Komic Oh yeah! I think so!

Komic:
Hmmm, just updated, still seeing the same behaviours as described in https://github.com/tootsuite/mastodon/issues/4491, not sure how to proceed. `media-gallery__item` divs are empty, not sure how to check if stuff is happening in the background.

Angristan:
I'm not really sure how this works, but would that mean that if I access a toot old enough so its attachement has been deleted by `mastodon:media:remove_remote`, it should be re-cached and I should see it again?
If that's the case, then it does not work. Example:
screenshot_17-09-2017_23-33-03

Gargron:
This works only if it's actually accessed. Which is not the case in the screenshot because it's type=unknown so it's displayed as a link instead of being loaded.

nightpool:
@Gargron so we need to redownload all of the missing MediaAttachments and clear them again for this to work?

Komic:
I don't understand when this is supposed to trigger then. If you try to access it, you'll just load it directly from the remote instance in a new tab or whatever?

Gargron:
API-wise if you get the URL from the media_attachment's url or preview_url attribute and visit it (by e.g. loading it from a `<video>` or ``) that's when the re-download will happen.

@nightpool I guess. I don't expect anyone to do that. It'll start working for you for the ones that get cleared in the future. Some instances have held back from clearing them (like mastodon.social) so for those it's gonna be more immediately useful.

Komic:
I see. Would a rake task to allow old, already removed media to benefit from this be possible?

Komic:
(please poke me if you'd prefer discussing this in a separate issue, I'll open one~)

Komic:
(bump~)

Gargron:
@Komic Oh, forgot to respond. No, it would be impractical as it would require you to redownload and re-process absolutely everything

Komic:
alrighty, thanks!

PR #4957, opened 2017-09-15
Introduce specialized serializers for collections

akihikodaki:
`CollectionSerializer` was abstract, and controllers which use it had concrete logics by themselves. This commit creates a specialized serializers and removes the logics from controllers.


Gargron:
I don't think that that part of the code needed a refactor, sorry

PR #4958, opened 2017-09-15
Bump ruby version to 2.4.2

zunda:
https://www.ruby-lang.org/en/news/2017/09/14/ruby-2-4-2-released/

`Gemfile.lock` is also updated.

TODO
- [x] Update `Dockerfile` with Alpine release of ruby-2.4.2


yukimochi:
I tried this PR with ruby 2.4.2 in Fedora 26 (Ruby was build by source) and Docker ruby:2.4.2-alpine3.6 (self-builded by official Dockerfile) .
I encounted error when webpush created, and webpush doesn't work.
Job:"WebPushNotificationWorker"
Error:"Webpush::ResponseError: host: fcm.googleapis.com, #<Net::HTTPInternalServerError 500 Internal Server Error readbody=true> body: <HTML> <HEAD> <TITLE>Internal Server Error</TITLE> </HEAD> <BODY BGCOLOR=..."

Gargron:
>500 Internal Server Error 

This is fcm.googleapis.com error. Unrelated to Ruby version

yukimochi:
I know. But rollback to Ruby v2.4.1 and revert this, WebPush works. I couldn't determine the reason why occur error.

zunda:
Is this something I expect to see when WebPush is enabled?

screenshot_2017-09-15_07-49-51

If so, I could confirm that Chrome showed this both on [ruby 2.4.2](https://github.com/zunda/mastodon/commit/8f5430ef5afdc3ae27ae8d4c1657e9f3dd2b314b) and rolling back to [ruby 2.4.1](https://github.com/zunda/mastodon/commit/305372729a149d4e8a30a086afc7c96ccfac5682) on the instance running on Heroku, unfortunately.

Might there be something I maybe able to reproduce the problem?

sorin-davidoi:
My guess is that `jwt` is to blame: https://github.com/zaru/webpush/issues/42

yukimochi:
In my test environment, use jwt 1.5.6 instead jwt 2.0.0 (rewrite Gemfile.lock), WebPush works well. The problem may this.

patf:
I've been running 2.4.2 in production for about a day now. No issues reported or observed, using jwt 1.5.6.

Gargron (approved):

Issue #2280, opened 2017-04-21
Higher-contrast accessibility option


citizenk:
As noted on [this toot](https://computerfairi.es/users/maunzikation/updates/2884), Mastodon's default style is not accessible to all. There are technical work-arounds (userstyles and such), but they may imply the help from other people or a certain technical knowledge to be useful.

An option to a higher-contrast, higher-visible interface would be nice, especially if offered from the get-go through the welcome screens.

Reference: https://computerfairi.es/users/maunzikation/updates/2884

* * * *

- [ * ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

wxcafe:
This should definitely be added in when there's a theme engine available (see #2028)

expenses:
This could be added with #1587 too.

nevillepark:
Reviving this thread to note that in the default theme, the text "Information below may reflect the user's profile incompletely. View full profile" is still too low-contrast.

![screenshot from 2017-11-16 12-53-43](https://user-images.githubusercontent.com/16054979/32907246-60a070d0-cacd-11e7-8da3-1e23ceff89c9.png)

Gargron:
High contrast theme has been added recently in #7213

Issue #2028, opened 2017-04-17
Theme Engine


mikebabb:
Would it be possible to add colour theming options to the user preferences menu?

(I'm raising this issue as requested by @ashfurrow in this thread: https://mastodon.technology/@ashfurrow/293424)

ashfurrow:
I'd be interested in helping with this, either spec'ing out or implementing.

wxcafe:
Let's make this issue the discussion about having a theme system in mastodon, I believe this would be great if possibly a bit complicated

pahakalle:
Would be nice to have a simple way to change lets say login page color and what not, but also a place to add custom css in the admin settings.

wxcafe:
There is a way to add custom CSS, but it's not in the UI: https://github.com/tootsuite/mastodon/pull/1368

NelsonTriesCoding:
There is this cool chrome-extension called Stylus / StylishThemes that allow people to apply custom CSS to websites (like Twitter's web Client, Facebook, Stack Overflow, GitHub...) for example take a look at that : https://github.com/StylishThemes/GitHub-Dark 

Is that what you where talking about @wxcafe ?

wxcafe:
No. I'm talking about a way for an admin to set a default theme for an instance. See the PR referenced in my previous comment

NelsonTriesCoding:
Oh, I see, my bad, should have opened it before assuming things.

wxcafe:
No problem, don't worry

mjankowski:
Sort of related to this, and a bunch of the other theme-related requests ... it would be a good first step towards any sort of theming improvement for someone to give a once-over to the scss files and replace the collection of `$colorX` (1-8) variables with more meaningful names ... because it will be sort of maddening to produce themes (assuming we just mean "a new option for a color scheme", and not a larger-scale html/css replacement) based on just those variable names.

ashfurrow:
I can take a look at that unless someone beats me to it.

Kimau:
It would be super handy as a starting point to make sure more elements have css classes so custom style sheets are easier to write.

I had to go pretty broad for example for a recent style hack to work because buttons, name ect... dont have css class. Only the content box. See example below.

Example: Compact Toots Style
---------------
`
.status > div {
    max-height: 0;
    font-size: 0.8rem;
    transition: max-height 200ms;
    overflow: hidden;
}

.status div.status__content {
    max-height: 60px;
    font-size: 0.8rem;
[...]

PR #4959, opened 2017-09-15
Add support for multiple themes

chr-1x:
Many non-flagship instances wish to differentiate themselves by providing customized themes for mastodon. However, not all users may necessarily appreciate the changes these themes make (and simply not using that instance may not be an option if they are deeply connected in the community). Furthermore, flagship may wish to someday provide alternate themes (such as a light theme) for accessiblity or preference reasons. This PR provides baseline support for this capability:

 - Add `config/themes.yml` where instance administrators can list out a series of theme names and paths
 - Refactor the webpack pipeline to package each listed theme separately
 - Change the application layout to use the package corresponding to the user's selected theme
 - Provide a ruby interface to the theme list in `app/lib/themes.rb`
 - Add a <select> field on the user preferences page, populated with the themes the instance provides


Chronister:
Closing until I fix the Nil default preference issue that's causing the travis build to fail

app/javascript/packs/common.js
```diff
@@ -2,7 +2,6 @@ import { start } from 'rails-ujs';
 
 // import default stylesheet with variables
```
nightpool: this comment is now out of date.

lynlynlynx:
LGTM 👍 
I tested this pull request, and this works really nice.
At long last, we can now add high contrast theme as secondary default theme!

This pull request is related to #2028 and #2280 .

unarist (changes requested):
* Since themes may be deleted in the future, then we should fallback to default theme. Currently it causes an error page.
* I think each theme should have "display name". It allows admins to change display name without breaking their current preference.

unarist:
and personally, I'd like to place that dropdown to next to language selection, like Kirakiratter does:

image

nightpool:
Currently on cybre.space having win95 enabled breaks the admin page:

image

not sure whether this should be considered a bug in this PR or not.

app/views/layouts/application.html.haml
```diff
@@ -19,6 +19,7 @@
       = title
 
     = stylesheet_pack_tag 'common', media: 'all'
+    = stylesheet_pack_tag (current_account&.user&.setting_theme || Setting.default_settings['theme']), media: 'all'
```
nightpool: law of demeter violation. this should probably be a method in ApplicationController instead

app/lib/themes.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'singleton'
+require 'yaml'
+
+class Themes
+  include Singleton
+
+  def initialize
+    @conf = YAML.load_file(Rails.root.join('config', 'themes.yml'))
```
nightpool: can't remember if this needs to be frozen or not in modern rails

app/models/user.rb
```diff
@@ -110,6 +110,10 @@ def setting_noindex
     settings.noindex
   end
 
+  def setting_theme
+    settings.theme
```
nightpool: alternatively could do fallback here but that seems less good.

app/views/layouts/application.html.haml
nightpool: law of demeter violation. also, this is just a little bit too complex of code. this should probably be a method in ApplicationController or something instead. you could handle fallback there too.

config/locales/simple_form.en.yml
```diff
@@ -44,6 +44,7 @@ en:
         setting_noindex: Opt-out of search engine indexing
         setting_system_font_ui: Use system's default font
         setting_unfollow_modal: Show confirmation dialog before unfollowing someone
+        setting_theme: Theme
```
nightpool: Might need to be a bit more verbose.

nightpool (changes requested):

app/views/layouts/application.html.haml
lynlynlynx: Please be sure to add theme dependent `stylesheet_pack_tag` to embedded.html.haml as well.

lynlynlynx (changes requested):

Chronister:
As of 8780db68f7f93ca576984554066bdfe01c2ae778, we now:
 - check current user's theme in ApplicationController
 - check there if the current user's theme is a known theme and return the configured default if not
 - look up theme names in the internationalization strings (falling back to the raw theme name if unavailable), so these can both be different from what's in themes.yml and localized to other languages
 - show a hint explaining the scope and effect of the theme setting
 - show the theme setting at the top of the preferences, above language (negotiable -- it may make more sense to visually group it with accessibility or display settings, or move these out into their own tab entirely)
 - use current_theme in embedded.html.haml

app/controllers/application_controller.rb
```diff
@@ -77,6 +78,13 @@ def current_session
     @current_session ||= SessionActivation.find_by(session_id: cookies.signed['_session_id'])
   end
 
+  def current_theme
+    if Themes.instance.names.include? current_account&.user&.setting_theme
+      return current_account&.user&.setting_theme
```
akihikodaki: I do not think it needs `&.` operator here. If they gets `nil`, it is a programming error and should fail rather than returning `nil`.

lynlynlynx (dismissed):

Chronister:
oops, i should look into why codeclimate says things before making changes. looks like 'and' and '&&' have different precedence.

app/views/layouts/embedded.html.haml
```diff
@@ -5,6 +5,7 @@
     %meta{ name: 'robots', content: 'noindex' }/
 
     = stylesheet_pack_tag 'common', media: 'all'
+    = stylesheet_pack_tag current_theme, media: 'all'
```
unarist: In most cases, embeds will be seen by non-logged in users. So I think it should be fixed to default theme, and it shouldn't be changed by user preferences.

unarist (changes requested):

app/controllers/application_controller.rb
```diff
@@ -77,6 +78,13 @@ def current_session
     @current_session ||= SessionActivation.find_by(session_id: cookies.signed['_session_id'])
   end
 
+  def current_theme
+    if (current_account && (Themes.instance.names.include? current_account.user.setting_theme))
+      return current_account&.user&.setting_theme
```
nightpool: should just be current_user, right?

nightpool: I *think* this should cover most cases and make more sense?

```ruby
return Setting.default_settings['theme'] unless Themes.instance.names.include? current_user&.setting_theme
return current_user.setting_theme
```

nightpool (dismissed):
LGTM modulo changing embedded.html.haml to use only the default theme and fixing the current_user issue

unarist (approved):

app/controllers/application_controller.rb
```diff
@@ -77,6 +78,11 @@ def current_session
     @current_session ||= SessionActivation.find_by(session_id: cookies.signed['_session_id'])
   end
 
+  def current_theme
+    return Setting.default_settings['theme'] unless Themes.instance.names.include? current_user&.setting_theme
```
akihikodaki: I think `return current_user.setting_theme if Themes.instance.names.include? current_user&.setting_theme` is easier to understand.

akihikodaki (approved):
Approving but have a non-blocking comment.

PR #4966, opened 2017-09-16
Fix filterable_languages method of SettingsHelper

akihikodaki:


db/schema.rb
```diff
@@ -328,7 +328,6 @@
     t.string "language"
     t.bigint "conversation_id"
     t.boolean "local"
-    t.index ["account_id", "id"], name: "index_statuses_on_account_id_id"
```
ykzts: why delete?

akihikodaki:
d64ec0a3047a76c0b6c11cd597833554e334550f: Fix https://github.com/tootsuite/mastodon/pull/4966#discussion_r139281006

Gargron (approved):

PR #4971, opened 2017-09-16
Add a new ProfileChange notification type (fixes #4108)

### LINKED ISSUE #4108: Feature request: new notification type when a user you interact with changes they profile picture
Hello,

A good part of the population, including me, primarily make usage of its visual memory to remember things. In the context of mastodon this apply to "recognizing other users based on their profile picture". The problem is when a user changes its profile picture, if the user isn't known enough by someone with a visual memory, confusion ensue and the user it not recognized anymore, awkward situations could ensues.

To mitigate this problem, I'm suggesting the following solution: send a notification when a user "with which I interact" changes it's profil pictures.

"with which I interact" could be defined in several ways:

- one on my following
- one on my following or one on my followers
- one on my following or one of the people with who I've been discussing with recently

The first one is obviously the easiest and less "will generate annoyed users" solution.

Of course, like any other notification this one can be disabled. I have no opinion on the opt-in/opt-out when it's introduced.

I'm opening this ticket following way more feedback than I've expected on that idea: https://social.wxcafe.net/@bram/1623515 (in proportions to my regular interactions).

Kind regards,

* * * *

ClearlyClaire:
This Pull Request introduces a new notification type for when people you follow change their name or profile picture.

There are still a few things to sort out:
- This is the first notification for transient information, whereas all previous notifications are backed by long-lived activity records that are still meaningful after every notification is discarded. For this reason, I have limited it to one profile change by followed user, but there is no automatic cleanup when all notifications are discarded. This could be improved.
- The UI only shows the new name/profile picture. This is far from ideal, but should be easy to change by someone familiar with Mastodon's web UI (I need help with that!). Ideally, I would like old and new avatar/name couples side by side with an horizontal arrow from the old one to the new one.

### DIFF
--- a/app/javascript/mastodon/reducers/settings.js
+++ b/app/javascript/mastodon/reducers/settings.js
@@ -26,6 +26,7 @@ const initialState = ImmutableMap({
   notifications: ImmutableMap({
     alerts: ImmutableMap({
       follow: true,
+      profile_change: true,
       favourite: true,
       reblog: true,
       mention: true,
@@ -33,6 +34,7 @@ const initialState = ImmutableMap({
 
     shows: ImmutableMap({
       follow: true,
+      profile_change: true,
       favourite: true,
       reblog: true,
       mention: true,
@@ -40,6 +42,7 @@ const initialState = ImmutableMap({
 
     sounds: ImmutableMap({
       follow: true,
+      profile_change: true,
       favourite: true,
       reblog: true,
       mention: true,
--- a/app/services/update_account_service.rb
+++ b/app/services/update_account_service.rb
@@ -1,13 +1,24 @@
 # frozen_string_literal: true
 
 class UpdateAccountService < BaseService
+  include ProfileChangeNotifier
+
   def call(account, params, raise_error: false)
     was_locked = account.locked
     update_method = raise_error ? :update! : :update
-    account.send(update_method, params).tap do |ret|
+
+    if params['display_name'] != account.display_name || !params['avatar'].nil?
+      prepare_profile_change(account)
+    end
+
+    res = account.send(update_method, params).tap do |ret|
       next unless ret
       authorize_all_follow_requests(account) if was_locked && !account.locked
     end
+
+    notify_profile_change
+
+    res
   end
 
   private
--- a/app/controllers/api/web/push_subscriptions_controller.rb
+++ b/app/controllers/api/web/push_subscriptions_controller.rb
@@ -23,6 +23,7 @@ class Api::Web::PushSubscriptionsController < Api::BaseController
       alerts: {
         follow: alerts_enabled,
         favourite: alerts_enabled,
+        profile_change: alerts_enabled,
         reblog: alerts_enabled,
         mention: alerts_enabled,
       },
--- a/app/javascript/mastodon/actions/notifications.js
+++ b/app/javascript/mastodon/actions/notifications.js
@@ -46,6 +46,9 @@ export function updateNotifications(notification, intlMessages, intlLocale) {
       notification,
       account: notification.account,
       status: notification.status,
+      display_name: notification.display_name,
+      avatar: notification.avatar,
+      avatar_static: notification.avatar_static,
       meta: playSound ? { sound: 'boop' } : undefined,
     });
 
--- a/app/javascript/mastodon/components/profile_change.js
+++ b/app/javascript/mastodon/components/profile_change.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import ImmutablePropTypes from 'react-immutable-proptypes';
+import PropTypes from 'prop-types';
+import Avatar from './avatar';
+import DisplayName from './display_name';
+import Permalink from './permalink';
+import { injectIntl } from 'react-intl';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+
+@injectIntl
+export default class ProfileChange extends ImmutablePureComponent {
+
+  static propTypes = {
+    account: ImmutablePropTypes.map.isRequired,
+    profileChange: ImmutablePropTypes.map.isRequired,
+    intl: PropTypes.object.isRequired,
+    hidden: PropTypes.bool,
+  };
+
+  render () {
+    const { account, profileChange, hidden } = this.props;
+
+    if (!account) {
+      return <div />;
+    }
+
+    if (hidden) {
+      return (
+        <div>
+          {account.get('display_name')}
+          {account.get('username')}
+        </div>
+      );
+    }
+
+    // TODO: better UI and RTL support
+    return (
+      <div className='account'>
+        <div className='account__wrapper'>
+          <Permalink key={account.get('id')+'-old'} className='account__display-name' href={account.get('url')} to={`/accounts/${account.get('id')}`}>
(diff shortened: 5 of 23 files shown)


app/javascript/mastodon/reducers/settings.js
```diff
@@ -21,20 +21,23 @@ const initialState = ImmutableMap({
   notifications: ImmutableMap({
     alerts: ImmutableMap({
       follow: true,
+      profile_change: true,
```
sorin-davidoi: I think you also need to add the key [here](https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/reducers/push_notifications.js#L7).

ThibG: Thanks, I'll do so when I manage to get “normal” notifications working like I want first.

ThibG:
I've been fighting with the front-end codebase, and if I understand correctly, what I'm currently missing is:
- Include previous avatar URLs and previous `display_name` in the serialization (notification_serializer.rb)
- Get the react.js controllers/containers/whatever to use them. Unfortunately, I really don't get that part, and I could use some help with this.

app/javascript/mastodon/reducers/settings.js
sorin-davidoi: Sure thing. On the same note, [this](https://github.com/tootsuite/mastodon/blob/master/app/models/web/push_subscription.rb#L27) may also need to be changed, since `data['alerts'][notification.type.to_s]` may not exist.

ThibG:
This almost works, but not quite: I can't figure out how to copy the avatar attachment from an `Account` to `ProfileChange`. Indeed, in this PR's current state, the *new* attachment ends up being saved.

We may be able to save the `ProfileChange` record before updating the `Account` one, but it is rather awkward, especially wrt. error/validation handling and factorization across the different ways of updating a profile (local profile, remote OStatus, remote ActivityPub).

app/services/update_account_service.rb
```diff
@@ -18,4 +23,8 @@ def authorize_all_follow_requests(account)
       [req.account_id, req.target_account_id]
     end
   end
+
+  def notify_followers(account, previous_avatar, previous_display_name)
```
nightpool: unused?

ThibG: Indeed! Fixed!

ThibG:
Cleaned up a bit, added i18n strings for English and French, and tried in “production” (my single-user instance). Everything seems to work fine, although Desktop notifications and Web Push notifications are not nearly as useful since they only show the new account profile.

The code could certainly be refactored in a nicer way, but I'm not completely sure how.

The `ProfileChange` React.js containers and components could also be improved by people actually knowing CSS and React.js, but it's functional.

ThibG:
Confirmed to work with both OStatus and ActivityPub!

Outstanding issues:
- It's not really readable when “long” usernames are involved
- Duplicate `ProfileChange` records and notifications might be created (race conditions? Need to investigate further)
- `ProfileChange` records are created even if there are no followers to notify (also includes newly-discovered accounts)

app/javascript/mastodon/reducers/settings.js
ThibG: Wouldnt that return nil, thus a false-y value, which is the default anyway?

nightpool:
So, just to be clear, this doesn't actually notify someone when the profile gets changed—only when their local server notices it?

ThibG:
@nightpool yes, but that's equivalent when dealing with ActivityPub (not with OStatus though), as Mastodon pushes profile changes to ActivityPub followers.

nightpool:
Ah, okay. but we'll only notify once per person? So if I change my display name twice it'll only notify for the first one? Or am I misunderstanding?

ThibG:
@nightpool only one notification will be saved per followed account, yeah. Followers will see only the last change.

app/javascript/mastodon/reducers/settings.js
ThibG: Assuming ActiveRecord handles `:json` fields as `Hash` objects, this should be fine: https://ruby-doc.org/core-2.2.0/Hash.html#method-i-5B-5D

Also didn't run into any error while trying it out.

akihikodaki (changes requested):
Honestly I am not in favor of this change. Notifications can be noisy and the number should be minimized.
I'd like to ask @Gargron whether it is acceptable or not. I'll review the details if so.

nightpool:
I think I'm *tentatively* in favor of this change, as long as it's only
limited to display names and avatars.

It might make more sense to put it in the home feed though, as basically a
synthetic status? Not sure. I don't know how much I like putting things
that *don't directly relate to you* in your notifications

ThibG:
@akihikodaki I guess it could be noisy, but profile changes are typically limited. I know I can't really generalize from my experience, but following ~70 people, I hardly even had a profile change that was not explicitly to try out that feature. It could also be disabled by default (not what my current feature request does).

@nightpool yes, it is limited to display names and avatar changes, I have no plan to extend it to any other kind of profile change. I can see how it could make sense to put it in the home feed, and I'm not opposed to that (not strongly in favor though). It would probably require a lot more work, however.

Psycojoker:
@akihikodaki @nightpool hello, you might want to read the original rational behind this idea https://github.com/tootsuite/mastodon/issues/4108

The short version is that for a non negligible group of users the avatar and/or the displayed name is the way to recognize other users in opposition to the username. Therefor, an unexpected change of avatar can results in a lot of confusion because the user can't be recognized anymore.

The group of users include people who have visual memories or dyslexic users has seen from the (enthusiastic) feedback gathered from sharing this idea, a real need has been expressed for that.

I think that adding a new kind of notification, that can be disabled or is disabled from the start, to make Mastodon a more accessible experience in general to its users is totally worth it and can totally make a difference :)

kyzh:
I concur with @psycojoker.

As a left to right reader, the profile picture is the main way to identify people.
Display name are a constant change for some accounts and actual handle are not often visible in full.
I'm personally always confused when people change their avatar.
Acknowledging this by having a setting, even hidden in the menu and off by default would be great way to improve my experience.

akihikodaki:
ThibG:
> I guess it could be noisy, but profile changes are typically limited.

kyzh:
> Display name are a constant change for some accounts and actual handle are not often visible in full.

I'm actually concerning accounts with such a constant change of their avatars or display names. Even such accounts are minor, it will make the notification ineffective for users with many followings and is not sustainable.

ThibG:
> It could also be disabled by default (not what my current feature request does).

If it is going to be merged, it should be enabled by default, or people just don't notice. This feature is helpful especially for inactive users, and enabling it by default is important to let such users take advantage of it.

ThibG:
Rebased on latest master.

@akihikodaki even if a user constantly changes their profile, there will be at most one notification for them (note, this is not completely true, as the client does not get notified of deleted notifications, similarly to what happens if someone repeatedly favorite/unfavorite a status, the notification will be sent out repeatedly even though it is stored only once).

I don't see many other options besides the Home Timeline thing, which has the advantage of not cluttering the notification column, but would require some very significant refactoring.

nightpool:
@Psycojoker i'm very well aware of the motivation. I just wanted to make sure I understood the implementation

PR #4972, opened 2017-09-16
Merge twitter-text

akihikodaki:
twitter-text was an external dependency, but big intrusive patches gets
added to adopt to Mastodon.

By merging them into one, code will be easier to read. Integration tests
are also possible.


Gargron:
I disagree with this. This will make future maintenance and project management harder. If we need to modify twitter-text behaviour I suggest we make a fork of twitter-text and use via bundler. Sorry.

PR #4977, opened 2017-09-17
Remove local_only scope in Status

akihikodaki:
# Question
Why does it have `local` and `local_only` scope at the same time?


Gargron:
>Why does it have local and local_only scope at the same time?

Seems like I forgot to remove it when I added "local" boolean to the table.

akihikodaki:
Then this is fine to merge.

Gargron (approved):

PR #4988, opened 2017-09-17
Custom emoji

Gargron:
- In OStatus: `<link rel="emoji" name="coolcat" href="http://..." />`
- In ActivityPub: `{ type: "Emoji", name: ":coolcat:", href: "http://..." }`
- In REST API: Status object includes `emojis` array (`shortcode`, `url`)
- Domain blocks with reject media stop emojis
- Emoji file up to 50KB
- Web UI handles custom emojis
- Static pages render custom emojis as `` tags

Side effects:

- Undo #4500 optimization, as I needed to modify it to restore
  shortcode handling in emojify()
- Formatter#plaintext should now make sure stripped out line-breaks
  and paragraphs are replaced with newlines


akihikodaki:
Can you share the reason why you don't use HTML in the federation? HTML is one of the most common standard.

Gargron:
>Can you share the reason why you don't use HTML in the federation? HTML is one of the most common standards

For the same reason that we don't put EmojiOne `` tags into status body. It's a visual decoration that is highly dependent on the environment where it's displayed. Moreover, we do not want to allow arbitrary image tags in federated statuses. Therefore we need the metadata "this is an emoji". We would have to, client-side or server-side, match up the image tag with that metadata, and transform the HTML to point to the locally cached emoji image file. Doing so via text transform seems much easier to me.

akihikodaki:
>  It's a visual decoration that is highly dependent on the environment where it's displayed.

So you mean they could be represented in a way different from `img` and its `alt` attribute are intended?

> Moreover, we do not want to allow arbitrary image tags in federated statuses. Therefore we need the metadata "this is an emoji".

1. I do not think a claim as "emoji" is not sufficient to prohibit other kind of images. They should be validated with objective criteria such as the file size as you pointed out, but also the way it is displayed (inline), the physical size (scaled to the font size), and so on.

2. Why not a custom tag? Using a tag is a obvious way to markup in HTML.

> transform the HTML to point to the locally cached emoji image file.

That is nothing to do with it, in my opinion. The URL is described with HTML attribute, which Mastodon actually deals with for validations.

lambadalambda:
This is a re-implementation of the same scheme we're already using in Pleroma. (https://gitgud.io/lambadalambda/pleroma/compare/c17c8ce36d35db03a007a264cc2e507afa9b803e...7c82b8219734102ff24d9dc24226c08351e608cc) 

Reasons for not putting the emoji in HTML:
- shortcodes are expressive already, so fall back to just text is ok.
- Implementations tend to aggressively strip new tags 
- Implementations all strip img tags

The custom emoji are supposed to be completely optional and should be safely ignored by non-supporting implementations. Just keeping a shortcode + meta information is the least invasive way to add them, I think.

nightpool:
The idea that *every* implementation that wants to support custom emoji now has to parse shortcodes in exactly the same way makes me very nervous, personally. It's much easier to use an HTTP parser to rewrite/de-duplicate embedded image urls then it is to figure out if dashes count as part of shortcodes, or periods do, or other colons do. @lambadalambda Which way does Plemora fall on the ipv6 problem, for example?

lambadalambda:
what do you mean 'parse the same way'? You get an array of shortcode -> image pairs for every used emoji and do a straight replace in the frontend.

nightpool:
Currently mastodon parses IPv6 addresses (like ` 2001:910:100::1`) in a way that prevents them from also being parsed as emoji shortcodes. The current implementation of this disallows any chained emoji—so :check::check::check: wouldn't get parsed either. This is the kind of stuff I mean

Gargron:
This can always be changed later, no? The status quo is plaintext short codes that don't display as anything special. In that regard we're not making anything worse, and adding a progressive enhancement. If our regex proves troublesome, we can adjust it. None of this is critical

app/javascript/mastodon/emoji.js
```diff
@@ -3,28 +3,48 @@ import Trie from 'substring-trie';
...
+      shortnameStartIndex = i;
+    } else if (!insideTag && (match = trie.search(str.substring(i)))) {
+      const unicodeStr = match;
+      if (unicodeStr in unicodeMapping) {
+        const [filename, shortCode] = unicodeMapping[unicodeStr];
+        const alt      = unicodeStr;
+        const replacement =  `<img draggable="false" class="emojione" alt="${alt}" title=":${shortCode}:" src="/emoji/${filename}.svg" />`;
+        str = str.substring(0, i) + replacement + str.substring(i + unicodeStr.length);
+        i += (replacement.length - unicodeStr.length); // jump ahead the length we've added to the string
+      }
```
krainboltgreene: We should turn this into a package on npm with tests and type annotations.

akihikodaki:
> * shortcodes are expressive already, so fall back to just text is ok.

That is true. The compatibility with old implementations would be perfect.
But what about new implementations? Any implementations should assume shortcode replacement happens if a name of the media matches even though names are not intended to be parsed and do not have to be validated in nature. That is not good.

* Implementations tend to aggressively strip new tags
* Implementations all strip img tags

Then what about `is` attribute?

> This can always be changed later, no?

* I tend to be critic this kind of change because I have seen such cases make difficult to keep compatibility (but often I have to.) Of course we may end discussions and go with the implementation if it takes too long, but I think it is not time to hurry yet.
* "can be changed later" is "will never changed". Even I would be reluctant to change once it is implemented. Though I agree that "none of this is critical" and I am fine to go with the spec, I just want to make sure that we have covered obvious alternatives before this change lands.

Gargron:
>"can be changed later" is "will never changed"

@akihikodaki This is very true. "There is nothing more permanent than a temporary solution". But: We are relatively far away from next release (probably), so if this gets merged into master, there will still be enough time to submit changes. The core of the PR is not being disputed at all, the local statuses will still have shortcodes in their raw body in the database, the only difference is whether to convert them to img in Formatter or not. It'll be nice to see this PR in action on glitch.social or some other instance once this is merged.

app/models/custom_emoji.rb
```diff
@@ -0,0 +1,38 @@
...
+class CustomEmoji < ApplicationRecord
+  SHORTCODE_RE_FRAGMENT = '[a-zA-Z0-9_]{2,}'
+
+  SCAN_RE = /(?<=[^[:alnum:]:]|\n|^)
+    :(#{SHORTCODE_RE_FRAGMENT}):
+    (?=[^[:alnum:]:]|$)/x
+
+  has_attached_file :image
+
+  validates_attachment :image, content_type: { content_type: 'image/png' }, presence: true, size: { in: 0..50.kilobytes }
```
ykzts: Does not support SVG?

ykzts:
screenshot

In public UI.

app/models/custom_emoji.rb
Gargron: I believe SVG might be a security hazard, PNG is a good start

ykzts: Okay. I agree.

Gargron:
@ykzts Is this :coolcat::coolcat: (no space) or is the problem multiple emojis?

ykzts:
@Gargron No. This problem seems to occur when an emoticon is the first character.

### failed case

```plain
[15] pry(main)> Status.find(223).text
=> ":coolcat: test"
[16] pry(main)> Formatter.instance.format(Status.find(223), custom_emojify: true)
=> "<p>:coolcat: test<img draggable=\"false\" class=\"emojione\" alt=\":coolcat:\" title=\":coolcat:\" src=\"http://localhost:3000/system/custom_emojis/images/000/000/001/original/emojo.png?1505776220\" /> test</p>"
```

### success case

```plain
[17] pry(main)> Status.find(222).text
=> "test :coolcat:"
[18] pry(main)> Formatter.instance.format(Status.find(222), custom_emojify: true)
=> "<p>test <img draggable=\"false\" class=\"emojione\" alt=\":coolcat:\" title=\":coolcat:\" src=\"http://localhost:3000/system/custom_emojis/images/000/000/001/original/emojo.png?1505776220\" /></p>"
```

Gargron:
@ykzts In good tradition of computer science, it was an off-by-one error! Fixed!

Gargron:
The Code Climate issue will have to stay - no way around it.

ykzts:
`ppa.launchpad.net` is down...?

ykzts (approved):

Gargron:
Tests are completing locally, so overruling Travis

clworld:
Some lacks about custom emoji (memo)
* Custom emojis can't be used in Display name, Bio.
* Custom emoji federation about updating image and deleting image is not supported.

Gargron:
@clworld I think there's probably no way to make them work in display name and bio. We rely on status metadata to carry their information... Accounts don't have that...

Issue #4888, opened 2017-09-11
about/more in 1.6.0 causes long-running query


asonix:
Counting the statuses on this instance should be fast, since there's only ~3,000, but requests time out because of it.

The first two queries in the attached image are the long-running count.

![image](https://user-images.githubusercontent.com/5304643/30255389-33deb9be-9669-11e7-915a-3ff6a329ca14.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
It sounds like they got "stuck" and you could kill them or restart Mastodon. Sometimes this happens even to requests which should be highly optimized.

asonix:
I have restarted mastodon-web multiple times, and this issue persists.

Observations:
- This didn't seem to happen with previous versions of mastodon
- This doesn't happen all the time, but it happens enough to notice.

KindlyFire:
I have the same problem (or similar) here (maly.io) three times in the last two days. Sidekiq fails to execute tasks because the 15 threads are blocked, waiting for SQL queries to complete. Killing all long-running queries fixes the problem and Sidekiq starts handling tasks again.

Have this problem on 1.6.0 and 1.6.1.

KindlyFire:
I do not know if it is related to `/about/more` though.

unstabler:
:'( 
![screenshot_20171025_123904](https://user-images.githubusercontent.com/964412/31979453-cc355cf0-b981-11e7-93b9-9dea7b553d9d.png)

ykzts:
Resolve from #4996.

PR #4996, opened 2017-09-18
Use Account.local.sum(statuses_count) instead of Status.local.count

akihikodaki:
It is faster.

### DIFF
--- a/app/presenters/instance_presenter.rb
+++ b/app/presenters/instance_presenter.rb
@@ -21,7 +21,7 @@ class InstancePresenter
   end
 
   def status_count
-    Rails.cache.fetch('local_status_count') { Status.local.count }
+    Rails.cache.fetch('local_status_count') { Account.local.sum(:statuses_count) }
   end
 
   def domain_count


ykzts:
Could you paste benchmark?

akihikodaki:
Sure.

Before:
```
> EXPLAIN ANALYZE SELECT COUNT(*) FROM statuses WHERE statuses.local OR statuses.uri IS NULL;
QUERY PLAN
Aggregate  (cost=2097206.31..2097206.32 rows=1 width=0) (actual time=211399.367..211399.367 rows=1 loops=1)
  ->  Seq Scan on statuses  (cost=0.00..2079733.00 rows=6989323 width=0) (actual time=0.015..209753.089 rows=7333522 loops=1)
        Filter: (local OR (uri IS NULL))
        Rows Removed by Filter: 30663909
Planning time: 1.157 ms
Execution time: 211401.419 ms
```

After:
```
> EXPLAIN ANALYZE SELECT SUM(statuses_count) FROM accounts WHERE domain IS NULL;
QUERY PLAN
Aggregate  (cost=59544.74..59544.75 rows=1 width=4) (actual time=5376.827..5376.828 rows=1 loops=1)
  ->  Seq Scan on accounts  (cost=0.00..58832.52 rows=284886 width=4) (actual time=0.526..5255.530 rows=285078 loops=1)
        Filter: (domain IS NULL)
        Rows Removed by Filter: 176648
Planning time: 1.995 ms
Execution time: 5376.909 ms
```

So it is approximately 40 times faster.

ykzts (approved):
good!

akihikodaki:
This removes the first two queries shown in #4888.

PR #4998, opened 2017-09-18
Fix incomplete account records being read

Gargron:
Problem: create_account saves record without public_key, url, and various endpoints, which breaks tasks that rely on those attributes being always present (the record may also appear to end users extremely empty looking)

Ideal solution: Transaction block around create_account and update_account (which fills this data). Impossible because update_account has to fetch some fields from HTTP requests (public key may not be embedded in payload, for example; also avatar, header), so transaction would be held up for up to multiple seconds, therefore decreasing throughput of other processes

Solution:

- Put account processing into redis lock
- Do not save until record is complete

Consequence: Account will not "exist" until it is complete, therefore the code path should take any usage of the account to the redis lock, which will wait until the first processing completes, at which point the record will exist and be complete.

### DIFF
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -8,11 +8,11 @@ class ActivityPub::ProcessAccountService < BaseService
   def call(username, domain, json)
     return if json['inbox'].blank?
 
-    @json           = json
-    @uri            = @json['id']
-    @username       = username
-    @domain         = domain
-    @collections    = {}
+    @json        = json
+    @uri         = @json['id']
+    @username    = username
+    @domain      = domain
+    @collections = {}
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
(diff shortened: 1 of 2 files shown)


app/services/activitypub/process_account_service.rb
```diff
@@ -8,18 +8,25 @@ class ActivityPub::ProcessAccountService < BaseService
...
-    old_public_key = @account&.public_key
-    create_account  if @account.nil?
-    upgrade_account if @account.ostatus?
-    update_account
-    RefollowWorker.perform_async(@account.id) if !old_public_key.nil? && old_public_key != @account.public_key
+    @json           = json
+    @uri            = @json['id']
+    @username       = username
+    @domain         = domain
+    @collections    = {}
```
akihikodaki: It has redundant spaces.

app/services/activitypub/process_account_service.rb
```diff
@@ -8,18 +8,25 @@ class ActivityPub::ProcessAccountService < BaseService
...
+
+    RedisLock.acquire(lock_options) do |lock|
+      if lock.acquired?
+        @account        = Account.find_by(uri: @uri)
+        @old_public_key = @account&.public_key
+        @old_protocol   = @account&.protocol
+
+        create_account  if @account.nil?
```
akihikodaki: I do not think it should have two spaces either.

akihikodaki (changes requested):

akihikodaki (approved):

PR #5001, opened 2017-09-18
Add will-change to improve scrolling perf

nolanlawson:
This improves the performance of scrolling on mobile browsers, especially mobile Chrome. It adds `will-change: transform`, which provides a hint to the browser to move the scrollable element to the GPU instead of repainting on the main thread.

There is some discussion [on StackOverflow](https://stackoverflow.com/a/41026886/680742). More obviously you can see the difference using the Chrome Dev Tools. Before, the "Scrolling performance issues" feature will warn about "repaints on scroll":

2017-09-18 12_57_35-developer tools - https___toot cafe_web_statuses_1690410

Afterwards, it doesn't:

2017-09-18 12_57_24-developer tools - https___toot cafe_web_statuses_1690410

You can also see the difference in how the FPS is affecting during scroll. Before:

2017-09-18 12_57_04-developer tools - https___toot cafe_web_statuses_1690410

After:

2017-09-18 12_58_50-developer tools - https___toot cafe_web_statuses_1690410

You can also test with your own eyes. 😃 I've pushed the change to https://malfunctioning.technology, (on top of v1.6.1), and if you open it in mobile Chrome (I used Chromium 63 on a Nexus 5), you'll see scrolling the home timeline is way faster than v1.6.1.

### Browsers

I tested in mobile Firefox on Android and didn't see a difference with or without this change. Edge will also not be affected, because Edge doesn't support `will-change`. I didn't have time to test Safari on an iPhone, so somebody should do that first to make sure it's okay.

### TODOs

As for what's causing the scrolling issues in the first place, it seems to me that various `touchstart`/`touchmove` listeners are being added by `react-swipeable-views`, `react-dom`, `is_mobile.js`, and `react-simple-dropdown`. It may be a good idea to audit these and set `passive: true` where we can. (In any situation where we're not calling `event.preventDefault()`, we can set `passive: true`; see [passive event listeners](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners))

BTW this would explain why touch scrolling is affected (mobile), but not mousewheel scrolling (desktop). Our `wheel` listener is passive, so mousewheel should be fine. (Scrolling is horribly complicated and even this `will-change` thing is new to me; I wrote [a blog post on scrolling](https://blogs.windows.com/msedgedev/2017/03/08/scrolling-on-the-web/) in case it's helpful. 😃)

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -1587,6 +1587,7 @@
   overflow-x: hidden;
   flex: 1 1 auto;
   -webkit-overflow-scrolling: touch;
+  will-change: transform; // improves perf in mobile Chrome
 
   &.optionally-scrollable {
     overflow-y: auto;


Gargron (approved):

Gargron:
Congrats on PR 5001!

nolanlawson:
FWIW Safari is reportedly about the same: https://witches.town/@viv/5263101

unarist:
Now scrolling feels smoother than 1.5.1. Great!

PR #5002, opened 2017-09-19
Admin interface for listing, adding and removing custom emojis

Gargron:


config/locales/en.yml
```diff
@@ -108,6 +108,18 @@ en:
       unsubscribe: Unsubscribe
       username: Username
       web: Web
+    custom_emojis:
+      created_msg: Emoji successfully created!
+      delete: Delete
+      destroyed_msg: Emojo successfully destroyed!
```
lynlynlynx: Typo 😭

Gargron: I did it on purpose as a nod to a popular joke on Mastodon. It's a rare occuring string, should be fine

lynlynlynx: Ah, convinced 👍

app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Admin
+  class CustomEmojisController < BaseController
+    def index
+      @custom_emojis = CustomEmoji.all
```
ykzts: Do you not paginate?

Gargron: It makes no sense to me to have more emojis than 40 or 50 at most, so no pagination required

That being said I realized this should filter by local ones only (for now)

ykzts: okay! and

> local ones only

Certainly.

ykzts (approved):

akihikodaki (approved):
Emojos everywhere.

app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Admin
+  class CustomEmojisController < BaseController
+    def index
+      @custom_emojis = CustomEmoji.where(uri: nil)
```
ykzts: Ah...

`uri: nil` 🙅‍♂️ 
`domain: nil` 🙆

PR #5004, opened 2017-09-19
Define emoji context for ActivityPub

Gargron:


ykzts (approved):

akihikodaki (approved):

Gargron:
I am asking in IRC whether it should be `http://joinmastodon.org/ns/emoji` or `http://joinmastodon.org/ns/emoji#`

akihikodaki:
Following existing examples, it should have the hash.
https://github.com/w3c/activitystreams/commit/7e7a5bcc4412c9f0e0cbf01c1466608149cfc5d3

cwebber:
The hash is good because it means fragment-ville, which means you can't necessarily retrieve it, which is usually what you want.  (Exceptions for things like schema.org where you *can* always retrieve the object so they do `{"schema": "http://schema.org/"}` and then you can retrieve `schema:Person` at [http://schema.org/Person](http://schema.org/Person).)

If you're going to make a namespace, why not go all out and make a non-Emoji-specific one like `{"toot": "http://joinmastodon.org/ns/#"}`?  Then you can do `toot:Emoji` and next time you have something very Mastodon specific like, I dunno, `toot:FortuneCookie`, you wouldn't have to define a new namespace?

Gargron:
@cwebber Okay, but what is better? If a different platform wants to implement emojis, but not "fortune cookies", wouldn't it be better if they only needed to include emoji-specific context, rather than the "general mastodon" context?

Gargron:
```
[04:40:44] <cwebber> Gargron: I think you're conflating contexts and vocabularies
[04:41:16] <cwebber> it's a confusing thing so I don't blame you, especially since you can often retrieve a json-ld context from the vocabulary URL
[04:43:11] <cwebber> Gargron: vocabulary namespace is "where the terms go" as in terms of "this specific uri really means this one specific thing".  so "http://joinmastodon.org/ns/#Emoji" in this example would mean Emoji as a specific *concept*
[04:43:23] <cwebber> whereas you can have multiple contexts map to the same term
```

cwebber:
I said this on IRC but I'll copy here:

```
<cwebber> Gargron: I think you're conflating contexts and vocabularies
<cwebber> it's a confusing thing so I don't blame you, especially since you                      
          can often retrieve a json-ld context from the vocabulary URL                           
<cwebber> Gargron: vocabulary namespace is "where the terms go" as in terms of                   
          "this specific uri really means this one specific thing".  so                          
          "http://joinmastodon.org/ns/#Emoji" in this example would mean Emoji                   
          as a specific *concept*
<cwebber> whereas you can have multiple contexts map to the same term
```

PR #5007, opened 2017-09-19
Add japanese translations for custom emoji.

mayaeh:
Add japanese translations for Custom emoji.

"alphanumeric" is translated "英数字" in Japanese.
But I adopted "半角英数字" meaning "Half size alphanumeric characters".

Because there are 2 bytes of alphanumeric characters in Japanese characters.
e.g. ＡＢＣ，１２３

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -111,7 +111,7 @@ ja:
     custom_emojis:
       created_msg: 絵文字の作成に成功しました
       delete: 削除
-      destroyed_msg: 絵文字の削除に成功しました
+      destroyed_msg: Emojoは消滅しました！
       emoji: 絵文字
       image_hint: 50KBまでのPNG画像を利用できます
       new:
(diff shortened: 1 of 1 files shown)


lynlynlynx:
Please modify these two below,
- `image_hint` should not contain spaces around 50KB and PNG same as icon and header sections.
- `shortcode_hint` should be "2文字", no space, same as username and bio sections.

----

Personally I think about `destroyed_msg`

I think using popular joke "Emojo" as it is is more humorous.
"Emojoは消滅しました！" :thinking:
Hmm, difficult, needs more opinions...

mayaeh:
Thank you.
Emojo ... 🤔

mayaeh:
I read https://github.com/tootsuite/mastodon/pull/5002#discussion_r139579169
( thanks for lynlynlynx )

I adopted it because it felt interesting about Emojo.
But I think many Japanese people Won't know it's Joke.

I'm worrying about restoring destroyed_msg.

mayaeh:
Restore destroyed_msg.

lynlynlynx:
I discussed with Japanese on Mastodon, and finally I agree with translating usually.

Gargron (approved):

PR #5014, opened 2017-09-19
Retry ActivityPub delivery a few more times

ClearlyClaire:
Increase the number of attempts to deliver toots through ActivityPub.

If I understand correctly how sidekiq works (https://github.com/mperham/sidekiq/blob/bd93f8430b2d1a7e320b7185aee114ef2bdabae5/lib/sidekiq/job_retry.rb#L203), Mastodon's current number of retries (up to 5 times) means delivery will be attempted for about a cumulative 10 minutes (7~14min), which is a pretty small amount of time (for perspective, on my small server, upgrading my instance may take up to one hour).

The proposed change would bring that cumulative time to around 1h20~1h40.

The downside is that this increases the chance of toots being displayed out-of-order, as they are saved to the timeline in the order of processing.

### DIFF
--- a/app/workers/activitypub/delivery_worker.rb
+++ b/app/workers/activitypub/delivery_worker.rb
@@ -3,7 +3,7 @@
 class ActivityPub::DeliveryWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'push', retry: 5, dead: false
+  sidekiq_options queue: 'push', retry: 8, dead: false
 
   HEADERS = { 'Content-Type' => 'application/activity+json' }.freeze
 


unarist:
Alternative option: increase each retrying delays.

https://github.com/tootsuite/mastodon/blob/7b8ecde32d732e830e628a2d713c1bd7ab69b9cf/app/workers/pubsubhubbub/subscribe_worker.rb#L8-L19

Delivery issues often lasts for a minutes or few hours. In that case, increasing delays reduces redundant tries and other jobs can work instead.

nightpool:
Maybe doing something like 
https://github.com/tootsuite/mastodon/blob/c8580eb806512fcc8ca76303e7d837f77c2cb413/app/workers/pubsubhubbub/subscribe_worker.rb#L8-L19

may work?

ThibG:
Yes it would, but I think a few tries within minutes are still a good idea, as you want the toots to be delivered in a timely fashion if possible (as opposed to PuSH subscriptions where you have some leeway).

Gargron (approved):

PR #5016, opened 2017-09-19
redo refactoring for emojify() with customEmojis

hcmiya:


Gargron (approved):

Gargron:
I am sorry, but this has led to a regression. The following text does not turn the unicode emoji into an img:

>Tired: ⓢⓞⓤⓟ
>Wired: 💩

Prior to this PR it works alright.

PR #5017, opened 2017-09-20
Fix custom emojis with non-1:1 aspect ratio being stretched

Gargron:


ykzts:
screenshot on timeline

good

ykzts (approved):

PR #5018, opened 2017-09-20
Make dropdowns render into portal, expand animation

Gargron:
image

Benefits:

- Dropdown no longer gets cut off by column, since it's on an independent layer above everything else
- Positioning of the dropdown is better bound to the dropdown icon now, regardless of layout
- Sweet little animation for when dropdown opens

Caveats:

@sorin-davidoi I temporarily removed the actions modal from the dropdown because it wasn't obvious to me how to make it work, could you help me plug it back in?


Gargron:
Also, if I also replace the "privacy dropdown" and the "emoji dropdown", then I think that's all dropdowns done and react-simple-dropdown dependency could be removed.

app/javascript/mastodon/components/dropdown_menu.js
```diff
@@ -1,53 +1,50 @@
 import React from 'react';
 import ImmutablePropTypes from 'react-immutable-proptypes';
```
ykzts: remove this line.

app/javascript/mastodon/components/dropdown_menu.js
```diff
@@ -56,90 +53,109 @@ export default class DropdownMenu extends React.PureComponent {
...
+          title={ariaLabel}
+          active={expanded}
+          disabled={disabled}
+          size={size}
+          ref={this.setTargetRef}
+          onClick={this.handleClick}
+        />
+
+        <Overlay show={expanded} placement='bottom' target={this.findTarget}>
+          <DropdownMenu items={items} onClose={this.handleClose} />
```
ykzts: ![screenshot on Dev Console](https://user-images.githubusercontent.com/12539/30624068-3f3dde68-9df6-11e7-85e2-bebeda6143f2.png]

app/javascript/mastodon/components/dropdown_menu.js
```diff
@@ -1,53 +1,50 @@
...
-    size: PropTypes.number.isRequired,
-    direction: PropTypes.string,
-    status: ImmutablePropTypes.map,
-    ariaLabel: PropTypes.string,
-    disabled: PropTypes.bool,
+    onClose: PropTypes.func.isRequired,
+    style: PropTypes.object.isRequired,
+    placement: PropTypes.string.isRequired,
+    arrowOffsetLeft: PropTypes.number.isRequired,
+    arrowOffsetTop: PropTypes.number.isRequired,
```
ykzts: Is `.isRequired` unnecessary?

ref https://github.com/tootsuite/mastodon/pull/5018/files#r139862589

app/javascript/mastodon/components/dropdown_menu.js
```diff
@@ -56,90 +52,109 @@ export default class DropdownMenu extends React.PureComponent {
...
-      <Dropdown ref={this.setRef} active={isUserTouching ? false : expanded} onShow={this.handleShow} onHide={this.handleHide}>
-        <DropdownTrigger className='icon-button' style={iconStyle} role='button' aria-expanded={expanded} onKeyDown={this.handleToggle} tabIndex='0' aria-label={ariaLabel}>
-          <i className={iconClassname} aria-hidden />
-        </DropdownTrigger>
+export default class Dropdown extends React.PureComponent {
 
-        {dropdownContent}
-      </Dropdown>
+  static propTypes = {
+    icon: PropTypes.string.isRequired,
```
sorin-davidoi:
```js
isUserTouching: PropTypes.func,
isModalOpen: PropTypes.bool.isRequired,
onModalOpen: PropTypes.func,
onModalClose: PropTypes.func,;
```

lynlynlynx:
Now, menus in statuses columns have `dropdown__left` class to avoid overflowing to the right edge of the browser. (Unfortunately, we can't avoid in some languages even if `dropdown__left` is set.)
But, with this commit, menus in statuses columns have `bottom` class, same as other columns.
<details>
<summary>This causes (before and after images)
</summary>
on master <-   -> with this commit
screenshot
</details>

I think menus in statuses columns should be dropped to left.
It seems this commit have drop-left menu option? (Sorry, I'm not sure how this works.)

Gargron:
@sorin-davidoi I have re-intergrated actions modal, all good :ok_hand: 

@lynlynlynx The dropdowns are now centered ("bottom"), which simplifies RTL handling. Also one of the main points of this change is that the modal can clip over the container borders, so left/right alignment should no longer be necessary: 

image

PR #5019, opened 2017-09-20
Change IDs to strings rather than numbers in API JSON output

aschmitz:
As discussed in #4801, this is being broken out into a separate PR to break up separate changes into separate PRs, and also to make it easier to identify immediate causes for changes post-squashing when merged.

From that PR:

# Probably-Breaking Change

The API now returns strings for IDs rather than IDs. There is a good chance that this will break some API clients, although more testing is probably required. (To my surprise, Tusky handled this change with no trouble.) The rationale for this is that JavaScript cannot parse JSON with numbers with more than 53 bits precisely, and we intend to use 64 bits (or at the very least, currently 57 bits). While we could go the Twitter route and provide both a semantically-correct integer and a JavaScript-usable string, it appears as though serving only the string is likely to lead to fewer issues in the long term. I'm open to exposing IDs as both strings and integers if that decision appears more prudent, however. (Ruby can output this with no problem.) More rationale for this decision is in the commit message for 86bf2c3.

### DIFF
--- a/app/controllers/api/v1/accounts/relationships_controller.rb
+++ b/app/controllers/api/v1/accounts/relationships_controller.rb
@@ -7,7 +7,10 @@ class Api::V1::Accounts::RelationshipsController < Api::BaseController
   respond_to :json
 
   def index
-    @accounts = Account.where(id: account_ids).select('id')
+    accounts = Account.where(id: account_ids).select('id')
+    # .where doesn't guarantee that our results are in the same order
+    # we requested them, so return the "right" order to the requestor.
+    @accounts = accounts.index_by(&:id).values_at(*account_ids)
     render json: @accounts, each_serializer: REST::RelationshipSerializer, relationships: relationships
   end
 
--- a/app/controllers/api/v1/accounts/relationships_controller.rb
+++ b/app/controllers/api/v1/accounts/relationships_controller.rb
@@ -7,10 +7,7 @@ class Api::V1::Accounts::RelationshipsController < Api::BaseController
   respond_to :json
 
   def index
-    accounts = Account.where(id: account_ids).select('id')
-    # .where doesn't guarantee that our results are in the same order
-    # we requested them, so return the "right" order to the requestor.
-    @accounts = accounts.index_by(&:id).values_at(*account_ids)
+    @accounts = Account.where(id: account_ids).select('id')
     render json: @accounts, each_serializer: REST::RelationshipSerializer, relationships: relationships
   end
 
--- a/app/javascript/mastodon/actions/store.js
+++ b/app/javascript/mastodon/actions/store.js
@@ -5,8 +5,7 @@ export const STORE_HYDRATE_LAZY = 'STORE_HYDRATE_LAZY';
 
 const convertState = rawState =>
   fromJS(rawState, (k, v) =>
-    Iterable.isIndexed(v) ? v.toList() : v.toMap().mapKeys(x =>
-      Number.isNaN(x * 1) ? x : x * 1));
+    Iterable.isIndexed(v) ? v.toList() : v.toMap());
 
 export function hydrateStore(rawState) {
   const state = convertState(rawState);
--- a/app/javascript/mastodon/components/account.js
+++ b/app/javascript/mastodon/components/account.js
@@ -21,7 +21,7 @@ export default class Account extends ImmutablePureComponent {
 
   static propTypes = {
     account: ImmutablePropTypes.map.isRequired,
-    me: PropTypes.number.isRequired,
+    me: PropTypes.string.isRequired,
     onFollow: PropTypes.func.isRequired,
     onBlock: PropTypes.func.isRequired,
     onMute: PropTypes.func.isRequired,
--- a/app/javascript/mastodon/components/autosuggest_textarea.js
+++ b/app/javascript/mastodon/components/autosuggest_textarea.js
@@ -128,7 +128,7 @@ export default class AutosuggestTextarea extends ImmutablePureComponent {
   }
 
   onSuggestionClick = (e) => {
-    const suggestion = Number(e.currentTarget.getAttribute('data-index'));
+    const suggestion = e.currentTarget.getAttribute('data-index');
     e.preventDefault();
     this.props.onSuggestionSelected(this.state.tokenStart, this.state.lastToken, suggestion);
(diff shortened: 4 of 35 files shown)


Gargron (approved):

nightpool:
Do we have to worry about JSON serialized to Redis or similar? what about the streaming API?

nightpool:
Redis says this about scores:

>**Range of integer scores that can be expressed precisely**
>
>Redis sorted sets use a double 64-bit floating point number to represent the score. In all the architectures we support, this is represented as an IEEE 754 floating point number, that is able to represent precisely integer numbers between -(2^53) and +(2^53) included. In more practical terms, all the integers between -9007199254740992 and 9007199254740992 are perfectly representable. Larger integers, or fractions, are internally represented in exponential form, so it is possible that you get only an approximation of the decimal number, or of the very big integer, that you set as score.

aschmitz:
I'm not aware of JSON that gets *stored* in Redis: some JSON gets sent to Redis pubsub, but that doesn't get stored (and any messages would switch when the application server code gets updated). A quick case-insensitive grep through for "redis" appears to confirm this, but if you know of somewhere where JSON is *stored* in Redis, that may need to be addressed.

The streaming API uses that Redis pubsub to get objects serialized by InlineRenderer, which itself uses REST::StatusSerializer or REST::NotificationSerializer, which are both updated here.

Gargron:
I think there are some places where raw JSON objects are sent into pubsub, without serializers, such as for status deletes, notification deletes.. Where it just sticks the ID of the object into an event. Those might need a to_s as well.

nightpool:
https://github.com/tootsuite/mastodon/blob/master/app/lib/feed_manager.rb#L27-L35 

seems like an issue given the score stuff above, and we may need to choose a way to pad it to make the lexicographic order correct.

https://github.com/tootsuite/mastodon/blob/master/streaming/index.js#L262-L267

this also seems like an issue

aschmitz:
The double-precision floats as Redis scores are probably best addressed in #4801, but the short version is that that PR does use IDs as scores, but also moves to storing the IDs as values.

Redis can support multiple entries with the same score, and sorts them lexicographically, which in almost all cases means in ID order (with the exception of cases where we are *just* rolling over to an extra digit in the decimal representation and still coming out to the same value as a double-precision float, which will happen once relatively soon, and then practically never again).

Previously, the Redis sorted sets used the potential for different scores and IDs to track whether an entry was a reblog or not, and to deduplicate reblogs. #4801 introduces a separate sorted set to track reblogs, so we can get away with having the same score and value for the feed sorted set. (This also makes it somewhat easier to track all recent reblogs of a status that a user might see, potentially allowing the UI to list all of them, and avoiding the issue of an un-reblog resulting in a non-followed user's status showing up in the feed, but those are potential future improvements and not super relevant here.)

nightpool:
Ah, I missed that there were redis changes that didn't get split out. Sounds good. Aside from the streaming API code I linked, the other place we use ids heavily that I could think of being a problem is sidekiq job arguments.

nightpool:
(which are stored in redis as JSON values afaik)

aschmitz:
Yeah, that streaming inadvertent cast is likely to be an issue in a few cases. I think we can get away with changing line 267 to:

~~~javascript
const encodedPayload = typeof payload === 'number' ? JSON.stringify(message) : JSON.stringify(payload);
~~~

(Or `'"' + message + '"'` if you prefer.) Does that seem like a viable strategy, or should I go through and change the values that are sent to Redis via publishing?

nightpool:
I'm not sure that helps. message is something like `{event: "delete", payload: 231268371283, queued_at: 199292301}`. The payload is supposed to be just an ID, and we'd be passing an object with a nested payload. We'd need to change anywhere we consumed the streaming API to accommodate that and it seems worse to do it there then to just fix the publish calls, right?

aschmitz:
Oops, neither of those are going to be practical, I'll have to go through and change any publishes of IDs to strings.

nightpool:
Okay i think LGTM if we fix the redis publishes. We should get this out in the wild soon enough to find any bugs and make sure the API developers have enough time to upgrade.

aschmitz:
591a9af356faf2d5c7e66e3ec715502796c875cd should fix the Redis issues in *this* branch, although I believe I'll need to make a different set of changes in #4801, because I moved some of the status deletion code around.

I believe deletes are the only cases where IDs are the only data passed, and I double-checked with `egrep -Ri "redis.*publish"`, but if you're aware of other cases or want to double-check, you probably know that code better than I do.

nightpool:
Yeah, I double checked all the Redis calls and these seem to be the only ones (excepting the use of scores) that could be a problem.

nightpool (approved):

kvvzr:
This PR affects the native client app.
Almost native apps almost are using strict languages, they can not convert from Long to String type and will crash.

Gargron:
Correct, this is a breaking change and will be released with Mastodon 2.0

kvvzr:
I'm sorry. I was misunderstanding.
When parsing old API response, it just manually converts IDs to String type, so it is not difficult... :bow:

clworld:
Some thought:
I feel 'breaking change released with Mastodon 2.0' is not a far future thing.
Pawoo already 1.6 and other huge instance may get 1.6 soon.
So, almost nothing is preventing from releasing 2.0. I think.
Time remains for 3rd party developer to deal with new api response structure is not huge, isn't it?

and FYI: PR note says "Tusky handled this change with no trouble.".
but author of SubwayTooter says
"Android API JSON perser parses Number String as Double and convert into Long when calling optLong(),
So, It has prescision problem like Javascript (works only on < 63bits)."
Thus, also Tusky may caught trouble with > 53bit IDs without changes.

aschmitz:
The note about optLong in Android is interesting, although that's a surprising behavior. It looks like Tusky actually stores IDs as strings anyway, though? (For example, see https://github.com/Vavassor/Tusky/blob/master/app/src/main/java/com/keylesspalace/tusky/entity/Status.java#L65 .) At any rate, Tusky is able to reply to, reblog, and delete statuses using the right, non-truncated ID, so I suspect the string type may be saving it from having to make any changes in this case.

clworld:
@aschmitz Oh, sorry. I misguessed Tusky's implementation. Excuse me.

PR #5020, opened 2017-09-20
Reduce wasted renders for intersection_observer_article.js

nolanlawson:
This reduces some unnecessary re-renders, as shown in this screenshot of `ReactPerf.wasted()`. (Scenario: load the local timeline in mobile mode, scroll to the bottom once.)

out

### DIFF
--- a/app/javascript/mastodon/components/intersection_observer_article.js
+++ b/app/javascript/mastodon/components/intersection_observer_article.js
@@ -21,6 +21,14 @@ export default class IntersectionObserverArticle extends ImmutablePureComponent 
     isHidden: false, // set to true in requestIdleCallback to trigger un-render
   }
 
+  updateOnProps = [
+    'id',
+    'index',
+    'listLength',
+    'saveHeightKey',
+    'cachedHeight',
+  ]
+
   shouldComponentUpdate (nextProps, nextState) {
     if (!nextState.isIntersecting && nextState.isHidden) {
       // It's only if we're not intersecting (i.e. offscreen) and isHidden is true


app/javascript/mastodon/components/intersection_observer_article.js
```diff
@@ -21,6 +21,14 @@ export default class IntersectionObserverArticle extends ImmutablePureComponent
...
   }
 
+  updateOnProps = [
+    'id',
+    'index',
+    'listLength',
+    'saveHeightKey',
+    'cachedHeight',
+  ]
+
```
nolanlawson: This change should be safe, because these are the props that are no longer compared:

- `intersectionObserverWrapper` - no sense in diffing this; it doesn't change anyway
- `onHeightChange` - listener, no sense diffing it
- `children` - this is already covered by `listLength`. There's no case I can think of where a toot would change but the `listLength` wouldn't change. Toots are added or subtracted from the list, but not replaced.

nolanlawson:
On second thought, I realize I am confusing `children` and `listLength`. I need to investigate more to understand what this code is doing, and how to safely reduce re-renders.

nolanlawson:
I think the issue here is actually `intersection_observer_article_container.js`. The problem is that we are spending time re-rendering it, because its descendant (`intersection_observer_article.js`) is the only one with both the props/state to know when to avoid re-renders in [its `shouldComponentUpdate`](https://github.com/tootsuite/mastodon/blob/b35406b700e643c50ac9726e5f5e1650604e0879/app/javascript/mastodon/components/intersection_observer_article.js#L24-L37).

So ideally we should just move the logic from the parent to the child and then the wasted re-renders would go away.

PR #5024, opened 2017-09-20
Measures to time out when waiting for startup of an instance running on Passenger.

HOPMOON:
Our instance is running on Nginx-Passenger. Nginx-Passenger stops unused instances and starts when needed. The instance normally starts in about 5 seconds. However, we have a lot of instances running on one server, so it takes more than 10 seconds to start up when the server is busy. At that time, if the external server sends Toote, a timeout error will occur.
HTTP :: TimeoutError: Delivery failed for https: //***.m.to/inbox: Read timed out after 10 seconds on https: //***.m.to/inbox
Please approve this amendment to avoid this error.

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -81,7 +81,7 @@ class Request
   end
 
   def timeout
-    { write: 10, connect: 10, read: 10 }
+    { write: 10, connect: 60, read: 10 }
   end
 
   def http_client


clworld:
Increasing timeout may causes performance issue of Sidekiq processing. Because it may occupies Sidekiq worker for only waiting 60 sec.

Sidekiq will retry request after some seconds and it may be success (because target instance went up by 1st failed request). So, it isn't a problem. I think.

Current 10 secs timeout is set by "HTTP signatures" https://github.com/tootsuite/mastodon/pull/4146 . In this PR, @gargron says "We'd rather fail and retry later." at this comment https://github.com/tootsuite/mastodon/pull/4146#discussion_r127252597 .

So, relying on Sidekiq retry is good enough. I think.

Gargron:
Yes, what @clworld said. If we raise the timeout to 60+10+10 seconds, that's going to be a single Sidekiq job occupying a thread for 80 seconds in the worst case.

PR #5027, opened 2017-09-20
Disable private status federation over OStatus

Gargron:


ykzts (approved):

Komic:
bye felicia

Gargron:
@Komic Who is felicia?

Komic:
http://www.urbandictionary.com/define.php?term=bye%20felicia just saying good riddance, nice job!~

Issue #1011, opened 2017-04-05
LOCAL_HTTPS does not work with custom domain and Heroku


pixeldesu:
So, I tried fiddling around with Heroku, custom domains and HTTPS for quite a while now, and I'm currently stuck at the last little thing that makes this a bit of a pain for my custom instance.

The first scenario is that `LOCAL_HTTPS` is causing an endless redirect loop. I've read about this on here and how to get around it, but I found no way setting this with the instance that can be automatically deployed to Heroku.

The other way I tried getting this to work is disabling `LOCAL_HTTPS` but having HTTPS and redirects from my DNS setup and Heroku themselves. This works fine for most of the stuff, I can even properly federate, the only downside of this is that the websockets connection is not over WSS (so, to see new stuff I have to manually reload).

Sounds really stupid, but I've been sitting around with this for a few hours now, and I'm really close to solving this. So, if anyone got stuff on Heroku working this way, I'd be glad on any tips on how you did this!

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

maxguru:
I am using nginx proxy setup within a docker container (https://hub.docker.com/r/jwilder/nginx-proxy/).  It already sends the 'X-Forwarded-Proto' header (see https://github.com/jwilder/nginx-proxy#replacing-default-proxy-settings).  Maybe it needs to be 'X-Forwarded-Protocol'?

Spunkie:
I'm running into the same issues. I followed the heroku setup guide and set cloudflare to `flexible` ssl but setting `LOCAL_HTTPS` to `true` causes redirect loops.

@pixeldesu Also would issue https://github.com/tootsuite/mastodon/issues/1126 possibly be caused by the websocket issues you mentioned?

pixeldesu:
@Spunkie I gave up on hosting on Heroku, as the files are not kept updated to reflect what is needed for Heroku to properly run it.

You need another dyno (or spawn inside a child process of the Rails app) for the Streaming websocket, I added this with just appending

```
streaming: npm run start
```

to the Procfile, which...adds another dyno, but was the easiest way I wanted to work with. 

After that you need to fix the way the `streaming/index.js` is pulling the Redis credentials. You can change the 4 lines trying to get the Redis credentials from the environment into

```
const redisClient = redis.createClient(process.env.REDIS_URL)
```
and connecting to Redis should work fine.

[...]

PR #5033, opened 2017-09-21
fix : mixed content of custom emoji

syui:
## Problem

My instance, since the URL of custom emoji is `http://`, there was a mixed content problem.

This commit is to fix it.

`img src` : `http://xxx/emoji.png` -> `//xxx/emoji.png`

I am not using AWS S3.

## Find and fix mixed content

https://developers.google.com/web/fundamentals/security/prevent-mixed-content/fixing-mixed-content

### Error message

```js
Mixed Content: The page at 'https://${host}/${user}' was loaded over HTTPS, but requested an insecure video 'http://${host}/system/custom_emojis/images/000/000/xxx/original/coolcat.png'. This content should also be served over HTTPS.
```

### Page where the problem occurred

public user page.

url : `https://${host}/@user`, example : https://mstdn.syui.cf/@syui

### Before

```html
<img draggable="false" class="emojione" alt=":coolcat:" title=":coolcat:" src="http://${host}/system/custom_emojis/images/000/000/xxx/original/coolcat.png?xxxxxxx">
```

###  After

`img src` : `http://` -> `//`

```html
<img draggable="false" class="emojione" alt=":coolcat:" title=":coolcat:" src="//${host}/system/custom_emojis/images/000/000/xxx/original/coolcat.png?xxxxxxx">
```

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -101,6 +101,9 @@ class Formatter
         emoji     = emoji_map[shortcode]
 
         if emoji
+          if emoji.match(/^http:\/\/.*/) then
+            emoji.sub!('http://','//')
+          end
           replacement = "<img draggable=\"false\" class=\"emojione\" alt=\":#{shortcode}:\" title=\":#{shortcode}:\" src=\"#{emoji}\" />"
           before_html = shortname_start_index.positive? ? html[0..shortname_start_index - 1] : ''
           html        = before_html + replacement + html[i + 1..-1]


Gargron:
Are you sure you have set LOCAL_HTTPS=true?

The custom emoji path is not generated differently than avatars, headers, media attachments, etc. This PR should not be necessary.

syui:
@Gargron 
In my environment did not work `LOCAL_HTTPS = true`.

- heroku + cloudflare

It may be similar to this symptom. #1011

Gargron:
@syui `CDN_HOST` has the format protocol://host so are you sure you have `https://` in there?

syui:
@Gargron `CDN_HOST` was not set.
Would you let me know what your thoughts are on this matter.

syui:
@Gargron I confirmed that it works well when I set `CDN_HOST`.
Close PR, Thank you so, so much!

PR #5034, opened 2017-09-21
fix : mobile style of /about/more

syui:
`/about/more` : site setting (description)

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -124,6 +124,7 @@
     box-sizing: border-box;
     max-width: 800px;
     margin: 0 auto;
+    word-wrap: break-word;
   }
 
   .header-wrapper {


app/javascript/styles/about.scss
```diff
@@ -124,6 +124,8 @@
     box-sizing: border-box;
     max-width: 800px;
     margin: 0 auto;
+    word-wrap: break-word;
+    overflow-wrap: break-word;
```
lynlynlynx: Until Edge supports `overflow-wrap` , I think setting only `word-wrap` is fine, same as other selectors have only `word-wrap` (except `overflow-wrap` for `.compact-header .h1`, wow it's my mistake, I noticed just now 😱 )

lynlynlynx (approved):

PR #5036, opened 2017-09-21
Eliminate re-renders for intersection_observer_article.js

nolanlawson:
Revised version of https://github.com/tootsuite/mastodon/pull/5020. Based on my testing, this should entirely eliminate wasted renders for this component.

Before:

screenshot 2017-09-21 10 28 55

After:

screenshot 2017-09-21 10 29 32

(To test, I loaded the local timeline in mobile mode and then scrolled to the bottom once.)

I did this by simplifying the logic in `shouldComponentUpdate`. This appears to catch all cases where attributes may change, although if I missed something let me know. I tested various cases and didn't see any where React failed to render a necessary update.

### DIFF
--- a/app/javascript/mastodon/components/intersection_observer_article.js
+++ b/app/javascript/mastodon/components/intersection_observer_article.js
@@ -1,10 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import ImmutablePureComponent from 'react-immutable-pure-component';
 import scheduleIdleTask from '../features/ui/util/schedule_idle_task';
 import getRectFromEntry from '../features/ui/util/get_rect_from_entry';
+import { is } from 'immutable';
 
-export default class IntersectionObserverArticle extends ImmutablePureComponent {
+// Diff these props in the "rendered" state
+const updateOnPropsForRendered = ['id', 'index', 'listLength'];
+// Diff these props in the "unrendered" state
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/intersection_observer_article.js
```diff
@@ -1,10 +1,15 @@
...
-import ImmutablePureComponent from 'react-immutable-pure-component';
 import scheduleIdleTask from '../features/ui/util/schedule_idle_task';
 import getRectFromEntry from '../features/ui/util/get_rect_from_entry';
+import { is } from 'immutable';
 
-export default class IntersectionObserverArticle extends ImmutablePureComponent {
+// Diff these props in the "rendered" state
+const updateOnPropsForRendered = ['id', 'index', 'listLength'];
+// Diff these props in the "unrendered" state
+const updateOnPropsForUnrendered = ['id', 'index', 'listLength', 'cachedHeight'];
```
nolanlawson: `children` seems to never change. `saveHeightKey` doesn't matter because it's not rendered; it's just used in the `handleIntersection` function. `onHeightChange` is a function. `intersectionObserverWrapper` never changes. I believe the props listed above are the only ones that matter for diffing.

nolanlawson:
This PR is live at https://malfunctioning.technology in case you want to test. 😊

Gargron (approved):

PR #5038, opened 2017-09-22
Document REDIS_NAMESPACE

nolanlawson:
I noticed that this feature (https://github.com/tootsuite/mastodon/pull/2869) didn't seem to be documented anywhere. Seems `.env.production.sample` would be a good place for it?

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -1,5 +1,6 @@
 # Service dependencies
 # You may set REDIS_URL instead for more advanced options
+# You may also set REDIS_NAMESPACE to share Redis between multiple Mastodon servers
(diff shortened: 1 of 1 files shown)


Gargron (approved):

clworld:
I wonder that is this feature works perfectly...
I believe that problems reported are fixed in nowadays..
but I sometimes heard that users trying use namespace / db with Redis got trouble...

Gargron:
@clworld the db option indeed does not work well, because redis pub/sub does not use db. However, namespace should be alright.

nightpool:
we should remove the DB option, tbh

PR #5039, opened 2017-09-22
Change ID columns to bigints

aschmitz:
Breaking this out of #4801, this changes all ID columns to bigints. Because it changes data types, this is a migration that may take large installations a little while to run.

The default ID column type in recent versions of Rails is bigint, but this fixes up ID columns (and columns which reference IDs) that were generated before that Rails change.


Gargron (approved):

db/schema.rb
```diff
@@ -10,20 +10,20 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170917153509) do
+ActiveRecord::Schema.define(version: 20170918125918) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "account_domain_blocks", id: :serial, force: :cascade do |t|
```
nightpool: The removal of id: serial seems incorrect here? Is this a consequence of changing to bigints or is it a schema change left over from the other PR?

aschmitz: It's confusing, but no, this is actually "correct". Rails had written out `id: :serial` chunks for these tables to specify that a 32-bit `serial` column was to be used, rather than the 64-bit `bigserial` that is now the default. Given that these tables have now switched to 64-bit columns, it no longer specifies the type, as it will default to `bigserial`.

(And just to make sure, I didn't edit `db/schema.rb` by hand, this file was the result of a `rake db:drop && rake db:create && rake db:migrate`.)

nightpool:
LGTM

db/schema.rb
nightpool: great, just wanted to make sure that there weren't migrations here on accident because the old one wasn't dropped.

yukimochi:
Why this migrate include deprecated_preview_cards?
This table is not used anymore and some instances erased this table.

aschmitz:
The migration included all tables with ID columns, and that table hasn't officially been deleted yet. If it's actually unused, I suspect a PR to remove it would be in order, but I haven't investigated in any detail.

yukimochi:
I reconfirmed migrations, my points was misunderstanding. Sorry.

Gargron:
@aschmitz Actually that table has been excluded from schema.rb!

Weird thing after I ran these migrations on two machines, same result on both. schema.rb changes like this:

```diff
-    t.bigint "tag_id", null: false
+    t.integer "tag_id", null: false
```

Gargron:
Oh man, this migration is a whole can of worms. I was a bit quick to approve...

It has run for about 10 minutes (with downtime) on glitch.social, which has approx. 2218986 statuses. Extrapolating, this means about 90 minutes downtime for mastodon.social, 3 hours for mstdn.jp... etc.

If I had added strong_migrations to the repository earlier, it would've rung alarm bells on this migration

On one hand it is clear that long-term, bigint > int. And the earlier this migration runs, the quicker, because tables grow over time. However, 3+ hours of downtime is inacceptable... I'm not sure what to do to fix this. Ideas?

ProgVal:
> I'm not sure what to do to fix this. Ideas?

Would this work? https://github.com/ankane/strong_migrations#renaming-or-changing-the-type-of-a-column

Gargron:
@ProgVal Consider that this is already merged into master and some admins have already run it...

Not only that, but a 6-phase migration split between db:migrate and code updates is very very hard to accomplish...

But would it help to split the migration file into different files for each table? Because every migration file runs in its own transaction, having all change_column's in one means a really big transaction... We might be able to minimize deadlock this way?

Another idea is to remove the updates to the statuses table from this migration. The changed columns are account_id and in_reply_to_account_id, and honestly, they will not become a problem until the accounts table actually becomes bigger than what int can store, and no instance is anywhere close to that moment.

patf:
GitLab has a number of migration helpers which might come in handy for this, see https://github.com/gitlabhq/gitlabhq/blob/2b4b76495dea4f037fbd958222fcaabdaf66bd67/lib/gitlab/database/migration_helpers.rb for details. It's implemented with triggers and even seems to handle foreign keys.

I'm not entirely sure what the best approach would be here, given that this was already merged and deployed on some instances. Changing existing migrations is gross, but getting rid of the changes to the status table might be the least worst option here. I *think* this would not cause any issues for those who already ran this migration and now already have bigints. For the rest, the change to bigint could be performed in a separate migration using something like the GitLab helpers - whenever it becomes necessary. Those who already have bigints would do some unnecessary work, but I suppose that's acceptable.

aschmitz:
Hey everyone, sorry for the trouble. Somehow I'd missed most of these notifications yesterday, so I'm a bit behind on this.

I think borrowing some of GitLab's migration helpers would probably be useful, but there may be a way to do this without making too much trouble. The plan as it is in my head at the moment is to:

* Create new (unused) columns for the ID columns that exist, and add triggers to keep them up to date. (Shouldn't block for long)
* Copy existing data into those new columns. (Shouldn't lock at all.)
* Add indices for those new columns in the background. (I believe Postgres allows this without locking: though that means the migration will take longer than if it were locked, this seems preferable.)
* Swap the column names such that the new columns are the "real" columns. (Locks everything, but should be fast.)
* Drop the old columns. (Assuming this can be done without massive amounts of time locked, anyway?)

I'd particularly like to avoid a situation where some users have keys in one int/bigint state, and some have them in another. (I'd also like to keep account IDs as bigints, as I already have >32-bit IDs, but if that seems like the most expedient way of fixing this, we can revisit it later.)

I'll investigate how feasible doing the above is, and try to play with some deliberately-gigantic tables, but that will probably take me a bit to get useful results. In the meantime, I suppose the general advice is that large installations should hold off on this migration for the time being.

patf:
The GitLab helpers do roughly what you described, with the addition of handling existing foreign keys, so there's probably some room for code re-use.

> I'd particularly like to avoid a situation where some users have keys in one int/bigint state, and some have them in another. (I'd also like to keep account IDs as bigints, as I already have >32-bit IDs, but if that seems like the most expedient way of fixing this, we can revisit it later.)

Removing the slow status column changes from the existing migration and adding a new one using the trigger/gitlab approach *should* meet all of these requirements, while remaining compatible with already-migrated instances (who'd just do some copying from one column of the same type to another).

Issue #4961, opened 2017-09-15
Webpack doesn't reload automatically in Vagrant/VirtualBox


aschmitz:
(Opening this as an issue rather than a PR to see if there's a reason for the existing behavior. If not, feel free to cherry-pick https://github.com/aschmitz/mastodon/commit/f866b860351e871b41b62ce8faff312c9b56c418 or I can open a PR.)

In Vagrant on macOS and Linux (I'm unsure about Windows), Vagrant + VirtualBox uses NFS to share the shared `/vagrant` folder. The way this is configured does not support `inotify`, so webpack doesn't notice that its JavaScript files have changed.

This can be "fixed" by setting a `poll` key inside the `watchOptions` object on the `devServer` config for webpack, with the frequency with which webpack should check for updates. On my development computer, a frequency of 1000 (once a second) appears to keep CPU usage low and still feels vaguely responsive, but a higher number (2000-3000) may be desirable if many developers are using slower computers or hard drives rather than SSDs. The delay in updating is dominated by the recompilation time regardless.

As I said, I'm not sure if there's a reason for avoiding this behavior, so I wanted to solicit opinions before actually opening a PR, but the fix is simple enough if someone wants to just go ahead and make it or cherry-pick it.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

wxcafe:
pretty sure you can open a PR with that, yeah.

bryszard:
Shouldn't this one be closed? I can see a merged PR here.

PR #5040, opened 2017-09-22
Configure webpack to poll for changes in development

aschmitz:
Vagrant on Linux/macOS hosts shared files via NFS, which doesn't support inotify-based watching of files. This tweak makes webpack check for changes every second, and rebuild if necessary. This removes the need to restart Foreman every time a frontend file changes. Note that rebuilding is still a relatively lengthy process.

The polling frequency can be changed to taste.

### DIFF
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -4,6 +4,17 @@ const merge = require('webpack-merge');
 const sharedConfig = require('./shared.js');
 const { settings, output } = require('./configuration.js');
 
+const watchOptions = {
+  ignored: /node_modules/,
+};
+
+if (process.env.VAGRANT) {
+  // If we are in Vagrant, we can't rely on inotify to update us with changed
+  // files, so we must poll instead. Here, we poll every second to see if
+  // anything has changed.
+  watchOptions.poll = 1000;
+}
+
 module.exports = merge(sharedConfig, {
   devtool: 'cheap-module-eval-source-map',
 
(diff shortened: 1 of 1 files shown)


aschmitz:
See also #4961 as the related issue.

nightpool:
Is there a way to configure this so that it only happens on NFS based systems? Or, at least, does this override or complement the inotify behavior?

aschmitz:
I hadn't actually investigated that, but it makes sense that some
developers would run this outside of Vagrant. Digging in a bit, webpack
delegates this to watchpack, which passes it off to chokidar.

It looks like setting `poll` does indeed force polling in at least one case
(macOS), so it's probably polite to avoid that where not necessary.

The config file for webpack is fully JavaScript, so we could attempt some
sort of detection there before setting the `poll` property, although I'm
not sure how to detect NFS off the top of my head. Alternatively, it looks
like setting the `CHOKIDAR_USEPOLLING` and `CHOKIDAR_INTERVAL` environment
variables would trigger the same behavior, and there may be a way of having
that happen only in Vagrant.

I'll think about how to best detect that and update this. Suggestions are
obviously welcome too.

Gargron (changes requested):
Make this controlled by an env variable, and set this env variable from the Vagrant box/definition

aschmitz:
Would it be sufficient to reuse the `VAGRANT` environment variable that's set in `.env.vagrant` , or should we add another variable to make it clearer for non-Vagrant users who want the same functionality?

Gargron:
Sure, use `VAGRANT`. Wasn't aware it exists 😅

Gargron (approved):

PR #5045, opened 2017-09-22
Do not filter statuses with unknown languages

akihikodaki:


spec/models/status_spec.rb
```diff
@@ -173,6 +173,22 @@
...
+  describe '.not_in_filtered_languages' do
+    it 'does not include statuses in filtered languages' do
+      user = Fabricate(:user, filtered_languages: ['en'])
+      status = Fabricate(:status, language: 'en')
+
+      expect(Status.not_in_filtered_languages(user.account)).not_to include status
+    end
+
+    it 'includes status with unknown language' do
+      user = Fabricate(:user)
```
unarist: This should have filtered_languages, because `language in ...` condition will be omitted unless the user has filtered_languages.

akihikodaki:
d144f6d55b800269aba615e12be1116b7ed25741: Fix https://github.com/tootsuite/mastodon/pull/5045#pullrequestreview-64632337

unarist (approved):

PR #5046, opened 2017-09-22
Switch from EmojiOne to Twemoji, different emoji picker

Gargron:
image

- [x] Remove redundant PNG graphics
- [x] Replace SVGs with Twemoji ones
- [x] Adjust unicode->filename conversion
- [x] Replace emojione-picker with emoji-mart
- [x] Make emoji-mart use local spritesheets instead of CDN

Thoughts:

- Server-side is still using the same lib/assets/emoji.json file for shortcode->unicode conversion (might be outdated now?)
- emojify() still uses the emojiList from old emojione package, which might be outdated, and also duplicates some data with the emoji index that emoji-mart brings to the table


beatrix-bitrot (approved):
wow this changeset is github hurting juice

nolanlawson:
I did a basic performance analysis of this commit. The total JS bundle size (for all bundles) dropped slightly from 2287516 bytes (2.29MB) to 2234053 bytes (2.23). CSS increased slightly from 183680 bytes (184kB) to 183998 bytes (184kB).

The new emoji picker is the main source of the JS savings:

screenshot 2017-09-23 09 32 39

The report shows `emoji-mart.js` takes up slightly less space than `emoji.js` and `emojione.json`:

out

As for this:

> emojify() still uses the emojiList from old emojione package, which might be outdated, and also duplicates some data with the emoji index that emoji-mart brings to the table

This is not such a huge deal IMO, because as you can see above, `emojione_light.js` (which is small) is in the `common` bundle, whereas `emoji-mart` is only loaded on demand. I agree that having out-of-date mappings is not ideal, but perf-wise this seems fine to me.

Also apparently `emoji-mart` is distributed as one big JS file (https://github.com/missive/emoji-mart/issues/86), meaning potentially we couldn't split it up even if we wanted to. (Although perhaps we could do some `preval` trickery to achieve this.)

Another perf win: `emoji-mart` sends the emoji themselves as one big spritesheet (/emoji/sheet.png, 2.9MB) as opposed to sending each emoji as a separate PNG. This will be a huge win for instances on HTTP/1 but should also help those on HTTP/2 because the compression of headers should be better. (Interestingly though the total size of the old PNGs was actually smaller: 1.97 MB total. I checked with `zopflipng-bin` though and couldn't shrink `sheet.png` any further.)

I haven't looked into other perf considerations for this PR (runtime perf, whether `emoji-mart.js` itself can be shrunk further) but overall, this LGTM. :smiley:

nolanlawson:
OTOH it looks like with recent commits we have removed the on-demand loading of the emoji picker? That's a bit of a shame, because it means `emoji-mart` is now in the common bundle:

screenshot 2017-09-23 10 09 44


(This is based on commit 1e02ba111ae38ab758135b5b2b46f34c672ca02e.)

Gargron:
@nolanlawson Because I had to import emojiIndex from emoji-mart in places other than emoji_picker_dropdown.js it refused to become a chunk, so I removed the chunking.

nolanlawson:
OK so if we can get `emoji-mart` to become modular, then we can bring back the separate chunk? In that case I might take some time to work on a PR to `emoji-mart`.

Gargron:
@nolanlawson I've opened a number of issues over on the emoji-mart repo. Tried to do some of it myself, but couldn't get the dev environment set up so I could see what I was doing. Also couldn't get my flavour of babel to compile (writing React classes without the `funcName = () => {}` and `static defaultProps = {}` syntax is a nightmare)

nolanlawson:
OK if I have time I'll try to look into those things. The repo seems [fairly active](https://github.com/missive/emoji-mart/graphs/contributors) so it may be good to concentrate development efforts over there, and at worst we can fork and point our package.json to the GitHub branch.

nolanlawson:
OK FWIW I looked into this, and as a test I wrote [a module](https://gist.github.com/nolanlawson/040031ab05276e84f02cc277902f28e8) that takes `require('emoji-mart').emojiIndex` and extracts only the code needed for that functionality. Unfortunately it wouldn't help us much:

screenshot 2017-09-23 18 26 06

`emojiIndex` alone is ~683kB whereas `emoji-mart` total is ~899kB. So there's no win from force tree-shaking (since the ~683kB is duplicated).

However if `emoji-mart` ever does become natively tree-shakeable, the async picker module would save us 300kB. So that's good to know. :smiley:

matyapiro31:
I don't understand this change. 
Why you change emoji fonts?
Do you hate emojione?
I love its round country flags.
It destroys the style of toot.

nightpool:
@matyapiro31 Emojione is no longer being updated and is going closed source. It won't get any of the new Unicode 10 emoji, which is the main reason for this change. see also https://github.com/tootsuite/mastodon/issues/4452

PR #5052, opened 2017-09-23
Add custom emojis to the emoji picker

Gargron:
image

### DIFF
--- a/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
+++ b/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
@@ -4,6 +4,8 @@ import { defineMessages, injectIntl } from 'react-intl';
 import { EmojiPicker as EmojiPickerAsync } from '../../ui/util/async-components';
 import { Overlay } from 'react-overlays';
 import classNames from 'classnames';
+import ImmutablePropTypes from 'react-immutable-proptypes';
+import { buildCustomEmojis } from '../../../emoji';
 
 const messages = defineMessages({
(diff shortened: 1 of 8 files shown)


app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
```diff
@@ -212,6 +218,27 @@ class EmojiPickerMenu extends React.PureComponent {
...
+        name,
+        short_names: [name],
+        text: '',
+        emoticons: [],
+        keywords: [name],
+        imageUrl: url,
+      });
+    });
+
+    return emojis;
```
ykzts:
```javascript
return this.props.custom_emojis.map(emoji => {
  const shortcode = emoji.get('shortcode');
  const url = emoji.get('url');
  const name = shortcode.replace(':', '');

  return {
    name,
    short_names: [name],
    text: '',
    emoticons: [],
    keywords: [name],
    imageUrl: url,
  };
});
```

How about using Array.prototype.map as above?

ykzts:
I think that `background-repeact: no-repeat` is necessary.

app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
Gargron: I could just use `.toJS()` on the Immutable.List, probably, likely without any performance losses. But it's not a big deal...

ykzts: i agree

ykzts (approved):

PR #5053, opened 2017-09-23
Add emoji autosuggest

### LINKED ISSUE #5044: Shortcode parsing inconsistency between custom and standard emoji
Standard shortcodes must be surrounded by whitespace, whereas custom emoji do not.
They use two different parsers, one server-side and one client-side.

This discrepancy will be confusing to users. I think it should be fixed to either work without spaces on both sides (preferred), or at least require them consistently.

example by wryk @ witches.town
screenshot_20170922_175749

Gargron:
image

image

Some credit goes to glitch-soc/mastodon#149

Fixes #5044 by removing shortcode processing from the server-side. The autosuggest inserts unicode straight away. However: you can no longer put shortcodes into display names or bios and expect them to turn into unicode. Shortcodes that are submitted will remain shortcodes.


app/javascript/mastodon/components/autosuggest_textarea.js
```diff
@@ -18,11 +20,11 @@ const textAtCursorMatchesToken = (str, caretPosition) => {
     word = str.slice(left, right + caretPosition);
   }
 
-  if (!word || word.trim().length < 2 || word[0] !== '@') {
+  if (!word || word.trim().length < 2 || ['@', ':'].indexOf(word[0]) === -1) {
```
ykzts: `!['@', ':'].includes(word[0])`

ykzts:
without custom emojis?

app/javascript/mastodon/components/autosuggest_textarea.js
Gargron: I'm sorry but I'm really used to the old way... It's not really important is it? We don't have a code style for includes vs indexOf

Gargron:
@ykzts Custom emojis are loaded in #5052. This PR does use them when they are loaded.

app/javascript/mastodon/components/autosuggest_textarea.js
ykzts: includes is short and simple. However, it is not important.

ykzts:
@Gargron 

> Custom emojis are loaded in #5052. This PR does use them when they are loaded.

Okay!

ykzts (approved):

app/javascript/mastodon/emoji.js
```diff
@@ -47,3 +47,43 @@ const emojify = (str, customEmojis = {}) => {
 };
 
 export default emojify;
+
+export const toCodePoint = (unicodeSurrogates, sep = '-') => {
```
hcmiya: This function can be written by one liner
```
(unicodeSurrogates, sep = '-') => unicodeSurrogates.match(/./ug).map(c => c.codePointAt(0).toString(16)).join(sep);
```

nclm:
Does it starts autocompleting after any space followed by a colon, or does it wait for at least a first letter? Github does the former and it makes it very annoying to write in French (and probably other languages with the same typographic rules) because all the colons are preceded by a space, so it keeps offering emoji suggestions each time there’s a colon in a sentence. The latter option would be more sensible.

Gargron:
@nclm It does not show up anything unless you typed at least one letter after :

MightyPork:
Could it be two? (that was the threshold in my version)
because then you type :C for a frowny face and get the suggestions - and from one letter it won't suggest anything good anyway
Also, only 3? I think there's space for at least 5 if you don't want a scrollbar…

PR #5054, opened 2017-09-23
In detail status view, display attachment uncropped if there's only one

Gargron:
A bit of a bummer: no way to make the media spoiler element the same height as the image/video it hides (since image/video is not in DOM when media spoiler is displayed)

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -5,6 +5,7 @@ import IconButton from './icon_button';
 import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
 import { isIOS } from '../is_mobile';
 import classNames from 'classnames';
+import sizeMe from 'react-sizeme';
 
 const messages = defineMessages({
   toggle_visible: { id: 'media_gallery.toggle_visible', defaultMessage: 'Toggle visibility' },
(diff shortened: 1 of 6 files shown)


Gargron:
@akihikodaki Let me know if you have ideas about the media spoiler height. I imagine this PR would be very desired by Pawoo since it's great for artworks.

akihikodaki:
Actually the media spoiler doesn't have to represent the height of the content. I think even one line is enough.

akihikodaki:
I'm testing this on the local environment, and it looks really nice with drawings. Thank you for the improvement.

nightpool:
We could cache the size of the image in the media attachment record. I
think paperclip might have an option for that?

abcang:
I think that it is possible to calculate the height by using the aspect in meta of media_attachments.

Gargron:
@abcang The react component is not aware of its own width, though... So even if I hack that in by checking the DOM, it won't be very react-like. There are ways to make it happen using higher-order components like autosize (new dependency), but...

If the height of the media spoiler does not matter, then this is done.

abcang:
Hmm, it seems difficult to calculate the height of the media spoiler. I think that fixing the height of the media spoiler is not a problem.

akihikodaki:
Maybe we could make up that with stretching animation or kind of?

Gargron:
@akihikodaki @abcang I used react-sizeme to make the media spoiler the right height. I also made it so the "standalone" mode is only enabled *if* there is enough information (meta->small->aspect)

I also enabled this mode on public detailed status pages.

abcang:
I think it's good!

PR #5061, opened 2017-09-23
Thread toot notification mails by conversation

ClearlyClaire:
This change threads e-mail notifications by conversation when possible.

I initially wanted to use the (single-quoted) Mastodon-generated conversation identifiers as the local-part of the message identifier, but I decided to just take the index of the conversation in the database: indeed, we have no control over external conversations URI and thus I don't know how we could derive unique and valid e-mail addresses from those.


Gargron (approved):
Seems cool

Gargron (approved):

PR #5063, opened 2017-09-23
Change mobile layout breakpoint from 1024px to 630px

Gargron:
630px = Drawer + 1 Column


nolanlawson:
Makes sense to me. I've definitely seen some weird layouts on desktop when resizing a window (e.g. to be half the width of a laptop screen). :+1:

nolanlawson (approved):

PR #5067, opened 2017-09-24
Generate 500.html with assets:precompile, remove loading from Google Fonts

Gargron:
Previously, 500.html was a static file and therefore it could not load our own fonts because for that it needs to load and parse the webpack manifest file. It needs to be a static file because it needs to be displayed when Rails is completely dysfunctional.

This moves the template to Rails, thus allowing it to load webpack assets, but generates a static version during assets:precompile rake task.

Please mind that there is no use in submitting translations for the `errors.500` string, as assets:precompile generates a single version.


nightpool:
translations could be based on the default locale, I guess?

public/500.html
```diff
@@ -1,43 +1,21 @@
 <!DOCTYPE html>
```
ykzts: `git rm public/500.html`

?

Gargron:
@nightpool The localization is not a change in this case. It remains as it was (previously hardcoded as English)

ykzts (approved):

PR #5072, opened 2017-09-24
WIP: emoji for account

akihikodaki:
The current implementation is domain-oriented. Obviously the choice has various pros and cons compared to alternatives. Here I point out a disadvantage of the domain-oriented implementation and a possible improvement by changing it account-oriented.

The disadvantage I'd like to describe here is the proprietary of emojis by domains. More specifically, it can be:
1. a limitation of ability to create emojis for normal users (those who are not admins, don't know how to launch an instance, cannot ask admins to add an emoji)
2. a limitation to migrate to other instances
3. a means to force users to migrate instances with emojis

Importantly, 2 and 3 can effectively construct some kind of norms that _users should be in a particular instance so that we can use common emojis_. For example, if an emoji named `:catnip:` become quite a popular as if it is an Internet meme, people interested in it would move to the page. They may even say "why don't you come here to use the cool emoji?" That is against decentralization.

This pull request proposes an approach similar to status delivery to decentralize emojos.

# Goals
* Any accounts can create emojos
* Authors can remove, but cannot update emojis
* Emojis can be identified by id and account (e.g. `:catnip@manekineko@cats.org:`)
* People can use others' emojis, depending on the visibility
* People can own create emoji collections and can refer emojis in them in short form (e.g. `:catnip:`)

# Progress
- [X] Emoji creation
- [X] Emoji federation
- [ ] Emoji favourite (in progress)
- [ ] Emoji removal
- [ ] Emoji report
- [ ] Lint
- [ ] Test

# Question
* I'm implementing "favourite emojis" collection as similar to status favourites. Should it be visible for other accounts? Should a creation of a favourite emoji be notified? In the current implementation, favourites are visible only for the favouring accounts.
* Should it keep the feature to bind emoji to instances? The implementation should be similar to favourite emojis, but emojis belong to instances.


tateisu:
how about API to listing emojis, for picker or autosuggest?

akihikodaki:
There are APIs to be implemented yet.

The list of emojis which the Web UI shows include:
1. Emojis the user favourited
2. Emojis created by the user

1 is necessary to make custom emojis scalable; not matter how many emojis there are, users can pick ones easily by referring to the list.
2 is a primary method to share emojis the user created. Others can pick emojis in statuses and "favourite".

The index of 1 is at `/api/v1/emoji_favourites`. Users can also edit the list via `/api/v1/emojis/favourite` and `/api/v1/emojis/unfavourite`. They works just as endpoints for status favourites do.

APIs for 2 are not yet implemented. The Web UI retrieves emojis from `initial-state`.
I am planning to implement `/api/v1/accounts/:account_id/emojis`. That would provide same functionality for third-party apps.

akihikodaki:
@Gargron gave some opinions at https://mastodon.social/@Gargron/19389045:
* It is too complex
* Server flavour is not bad
* Adding an admin UI for remote emojis with a simple "copy" button to make a remote emoji available to your users

Considering them, I'll make some changes for the design. First, I'll open another pull request to implement "copy" button in the admin UI, and this pull request would deal with minimal extensions.
The large portion of this pull request is for implementing importing feature of emojis. Putting them into another pull request reveal the complexity of the account-oriented features and allows to plan how to reduce it.

Implementing "copy" button would need a way to prevent name collision different from `:catnip@manekineko@cats.org:` format because it is a feature to import names from multiple namespaces to one namespace. If different instances provide different emojis with same names, they collide when using "copy" button.
I'm going to solve the issue by implement shortcode renaming feature. `:catnip@manekineko@cats.org:` format will be abandoned in favor of it.

akihikodaki:
I'll also extract emoji bindings to statuses. The feature makes emojis of old statuses persistent even after updating emojis or replacing emojis with the same name. That prevents the future depletion of the namespace.

akihikodaki:
I'll close this issue. I'm no longer interested in the feature because the emoji feature landed with 2.0.0 and there is no way to implement but adding a different protocol.

PR #5078, opened 2017-09-24
Add strong_migrations gem to warn when creating unsafe migrations

Gargron:


ykzts (approved):
https://github.com/ankane/strong_migrations looks good

PR #5080, opened 2017-09-24
Fix emojify() by generating a mapping to existing Twemoji files

Gargron:
Before:  🏳️‍🌈 went to a broken image because EmojiOne data pointed to `1f3f3-1f308.svg` while the new file is at `1f3f3-fe0f-200d-1f308.svg`

Now: Pride flag is fixed:

image

A new rake task emojis:generate downloads a full list of valid
unicode sequences from unicode.org and checks it against existing
Twemoji files, finally generating a map from each sequence to the
existing file (e.g. when there's multiple ways an emoji can be
expressed). The map is dumped into app/javascript/mastodon/emoji_map.json

That file is loaded by emojione_light.js (now a misnomer) which
decorates it further with shortcodes taken from emoji-mart's index.


Gargron:
@nolanlawson You wrote the tests that are failing, so I seek your advice. A part of these is just renaming of shortcodes and filenames, but it also seems to treat that exclamation point weirdly, and i don't know what's the expected behaviour here:

```
  25 passing (119ms)

  2 failing

  1) emojify does unicode:

      AssertionError: expected '<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":woman-woman-boy-boy:" src="/emoji/1f469-200d-1f469-200d-1f466-200d-1f466.svg" />' to equal '<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":family_wwbb:" src="/emoji/1f469-1f469-1f466-1f466.svg" />'

      + expected - actual

      -<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":woman-woman-boy-boy:" src="/emoji/1f469-200d-1f469-200d-1f466-200d-1f466.svg" />

      +<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":family_wwbb:" src="/emoji/1f469-1f469-1f466-1f466.svg" />

      

      at Context.<anonymous> (spec/javascript/components/emojify.test.js:24:94)

  2) emojify does multiple unicode:

      AssertionError: expected '❗ <img draggable="false" class="emojione" alt="#️⃣" title=":hash:" src="/emoji/0023-fe0f-20e3.svg" />' to equal '<img draggable="false" class="emojione" alt="❗" title=":exclamation:" src="/emoji/2757.svg" /> <img draggable="false" class="emojione" alt="#️⃣" title=":hash:" src="/emoji/23-20e3.svg" />'

      + expected - actual

      -❗ <img draggable="false" class="emojione" alt="#️⃣" title=":hash:" src="/emoji/0023-fe0f-20e3.svg" />

      +<img draggable="false" class="emojione" alt="❗" title=":exclamation:" src="/emoji/2757.svg" /> <img draggable="false" class="emojione" alt="#️⃣" title=":hash:" src="/emoji/23-20e3.svg" />

      

      at Context.<anonymous> (spec/javascript/components/emojify.test.js:34:48)
```

nolanlawson:
Tests are failing in Travis: https://travis-ci.org/tootsuite/mastodon/jobs/279214982

screenshot 2017-09-24 09 03 32

Looks like the tests just need to be updated because of the `-200d-` joiners not being used.

clworld:
Applied it on mdx.ggtea.org. I see some glitch.
What I post.
```
♀ ♂
🙅🏾 🙅🏾‍♂️
🙆🏾 🙆🏾‍♂️
🙅 🙅‍♂️
🙆 🙆‍♂️
````
What I get.
mdx_webui
What I get on md.ggtea.org(master) .
md_webui

It seems it can handle skin-tone only or gender only variant, but it can't handle when both added.
Additionally, It converts '♀', '♂' as emoji. (It is same behavior as Twitter does, so, it's correct when using twemoji.)

File: app/javascript/mastodon/emojione_light.js [hunk 1]
```diff
@@ -1,13 +1,40 @@
 // @preval
```
nolanlawson: I don't believe it's valuable to keep using `@preval`. I tested with and without it, and because the emoji data is duplicated, `common.js` is 125kB smaller if we remove `@preval` (1463880 bytes vs 1338993 bytes).

Also I tested on Chrome on a Nexus 5, and the time to run this code client-side is only 35 milliseconds. So it's not a good tradeoff IMO.

Gargron:
No matter what I do with this PR, *something* is always broken. I don't know what to do. This is too arcane. Maybe we'll go back to EmojiOne v2.2 because at least their index corresponds with their filenames.

nolanlawson:
Here's with and without `@preval` (note the big `emojione_light.js` is gone:

out

nolanlawson:
I think `@preval` makes sense, but only if we're using it 100% of the time for the `common` chunk and then fully-loading the picker in an async chunk.

clworld:
¿🤔?

Gargron:
Update: I managed to fix it with brute force

spec/javascript/components/emojify.test.js
```diff
@@ -22,10 +22,10 @@ describe('emojify', () => {
...
   it('does unicode', () => {
     expect(emojify('\uD83D\uDC69\u200D\uD83D\uDC69\u200D\uD83D\uDC66\u200D\uD83D\uDC66')).to.equal(
-      '<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":family_wwbb:" src="/emoji/1f469-1f469-1f466-1f466.svg" />');
-    expect(emojify('\uD83D\uDC68\uD83D\uDC69\uD83D\uDC67\uD83D\uDC67')).to.equal(
-      '<img draggable="false" class="emojione" alt="👨👩👧👧" title=":family_mwgg:" src="/emoji/1f468-1f469-1f467-1f467.svg" />');
-    expect(emojify('\uD83D\uDC69\uD83D\uDC69\uD83D\uDC66')).to.equal('<img draggable="false" class="emojione" alt="👩👩👦" title=":family_wwb:" src="/emoji/1f469-1f469-1f466.svg" />');
+      '<img draggable="false" class="emojione" alt="👩‍👩‍👦‍👦" title=":woman-woman-boy-boy:" src="/emoji/1f469-200d-1f469-200d-1f466-200d-1f466.svg" />');
+    expect(emojify('👨‍👩‍👧‍👧')).to.equal(
+      '<img draggable="false" class="emojione" alt="👨‍👩‍👧‍👧" title=":man-woman-girl-girl:" src="/emoji/1f468-200d-1f469-200d-1f467-200d-1f467.svg" />');
+    expect(emojify('👩‍👩‍👦')).to.equal('<img draggable="false" class="emojione" alt="👩‍👩‍👦" title=":woman-woman-boy:" src="/emoji/1f469-200d-1f469-200d-1f466.svg" />');
```
nolanlawson: This is much more semantic TBH

nolanlawson (approved):
Except for the fact that we don't need `@preval` anymore, this looks great to me. :+1:

We can add `@preval` back in once we 1) re-enable on-demand picker loading, and 2) modify other uses of emoji data to use `@preval`. I'm happy to start working on that if you'd like.

Gargron:
@nolanlawson Maybe I am confused about what preval does, but this file performs some operations on a rather large object. I thought preval evaluates this operation during build, so at the end you only get the final object as export. Is that not true?

sorin-davidoi:
It's true, but we load the bigger file anyway, so it only duplicates data. If we no longer load the entire file (like it was with emojione), then it makes sense to use it.

Gargron:
@sorin-davidoi emoji_map.json is not loaded from anywhere else - or what do you mean?

sorin-davidoi:
Ah, sorry, I was looking at the bundles posted by Nolan, but those are outdated since you created `emoji_map.json`. In that case, I guess it makes sense to keep using it.

nolanlawson:
Ah yes you're right @gargron; since `emoji_map.json` is not loaded anywhere else, `preval` still makes sense.

That said, I would still like to trim and code-split `emoji-mart` somehow, but it may not be possible/useful until they support ES modules/tree-shaking. Our usage of `emojiIndex.search` also makes this difficult.

PR #5084, opened 2017-09-24
Split up int->bigint migration per table, do not change statuses

Gargron:
- Group change_column's by table. Each migration file is a transaction/lock, smaller locks = better
- Do not modify statuses table. It is the biggest table.


db/migrate/20170918125937_ids_to_bigints20.rb
```diff
@@ -0,0 +1,11 @@
+class IdsToBigints20 < ActiveRecord::Migration[5.1]
```
Gargron: The question is, will the stream_entries table survive until the int limit is exhausted?

akihikodaki (changes requested):
The change should have been expected to keep consistency between tables; leaving foreign key references in `statuses` as it is just cause more confusing.
If altering `statuses` table is not practical, why don't you just abandon the change?

Gargron:
@akihikodaki 

I believe that upgrading to bigint sooner rather than later is good (and easier). This is why statuses table had bigint id for months already. But the statuses table is so large it is not practical to touch it. The fields that it would upgrade are:

- account_id
- in_reply_to_account_id
- application_id

Neither accounts nor applications are fast-growing tables. So I believe it won't be soon that those columns will approach integer overflow.

They will need to be upgraded eventually, but it needs special care. One of the suggested procedures is creating new columns, writing to both, making old column read-only, then removing it, over many migrations. However, as you can see that is very complicated. Since unfortunately the original PR was already merged, I feel like I need to at least fix it asap to not bring down instances for 3 hours.

akihikodaki:
I agree the opinion that `statuses` should not have to be upgraded now. But I think it applies also for `accounts`.

Gargron:
@akihikodaki So I'll remove all accounts stuff from the migrations then? cc @aschmitz

akihikodaki:
Yes, I meant that.

akihikodaki:
Requesting a review from @nullkal; his instance, mstdn.jp has the largest number of statuses (even larger than Pawoo.)

aschmitz:
This is certainly one way to do something similar to the original migration. I have a mostly-complete change that instead uses the GitLab concurrent migrations code to replace `IdsToBigints`, I'm just working out a dangling issue with index names, if you can hang on a bit longer.

aschmitz:
Well, that took longer than expected. I went down a long path of trying to make `id` column (not `foo_id` column) changes concurrent, but that looks like a lot more effort than it's worth, and the primary concern appeared to be about `foo_id`-type column changes. Anyway, the PR is at #5088 if someone wants to take a look.

Gargron:
I guess I am closing this in favour of #5088..? I still think that splitting the migrations might be helpful, in case something goes wrong halfway through, you can continue from where it failed rather than restarting from the very beginning (I think)

@aschmitz @akihikodaki

aschmitz:
Recovery in the case of an aborted concurrent column change will be
interesting, but should be doable. It's not in a transaction, so in theory
it should be resumable if we build that in (which is probably a good idea,
given that it can take a while).

Issue #4989, opened 2017-09-17
Add a way to hide someone from timelines, but not from notifications


ProgVal:
One may want to hide someone else from their timelines (eg. because they write too many tweets/toots), but still see notifications in case they want to talk.

This was possible in Mastodon using the "mute" feature, before v1.5.

Could a similar feature be added again?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nightpool:
see also https://github.com/glitch-soc/mastodon/pull/100

Gargron:
>This was possible in Mastodon using the "mute" feature, before v1.5.

Yeah, this used to be "mute". And then people were like "mute should be like *that* instead", and I told them, but surely this is useful functionality worth keeping, and they were like "nope, want it to work this way", so now it works this way, and the circle is complete.

ProgVal:
@Gargron Yeah, I saw that discussion :/

egypturnash:
I was wandering to the Issue Tracker to see if there was any discussion of whether or not someone mentioning you and someone you muted should or should not show up in your mentions. (Currently it does.) Clearly there is a lot of difference of opinion as to What Mute Should Do. And honestly I can see times when all of these possibilities are the Right Thing.

Possibly 'mute' could use a submenu that offers "mute their toots but show when they mention me/mute toots and mentions/mute all toots from or involving this person". There is probably a better way to express these three options. Maybe separate checkboxes for muting "their toots", "them mentioning me", and "other people mentioning them". This may, however, be the start of a slippery slope that leads to something like DeviantArt's watchlist management page. (see attached)

![screen shot 2017-09-17 at 8 10 41 pm](https://user-images.githubusercontent.com/403223/30528100-7d6cb0b0-9be4-11e7-8d81-fefb0b0bd3cf.png)

Cassolotl:
How about...

* **Sneaky block** - you don't see them at all, no notifications or anything, but they can see all your stuff. They have no way of knowing that you've sneaky-blocked them.

* **Obvious block** - you don't see them at all, no notifications or anything, *and they can't see you either, even if they look really hard.* It is obvious that they have been obvious-blocked because they plug in your username to look at your profile and they get a big error.

[...]

PR #5087, opened 2017-09-25
Optional notification muting

ekiru:
This PR allows users to choose between the current mute behavior for notifications (which mutes posts in timelines, context around other statuses in the detailed view, and mutes notifications) and the old (pre-1.5?) behavior (which is the same except for allowing notifications from the muted user) for each muted user. It should address #4989. It's just a cherry-pick of the implementation of this feature [over in the glitch-soc fork](https://glitch-soc.github.io/docs/features/optional-notification-muting/) (the link contains a description of the UI and behavior and a few screenshots).

The mute modal, pictured below, has a checkbox added to it (with a default label of "Hide notifications from this user?") which defaults to being checked (to match the current behavior). When viewing a list of accounts (such as for muted users, follower/following lists, etc.), there are also now two buttons/icons for muted users instead of just the one: one to unmute and one to toggle whether notifications from the user are muted.

The new mute modal

The API changes are fairly minimal:

1. A new `/api/v1/mutes/details` endpoint that returns an array of objects describing the `Mute`s instead of just an array of accounts; the `/api/v1/mutes` endpoint is left unchanged for backward-compatibility's sake and returns all mutes, regardless of whether they mute notifications or not.
2. The `/api/v1/accounts/:id/mute` endpoint accepts an optional `notifications` parameter, which defaults to true and determines whether the mute will mute notifications or not.
3. When the API returns a `Relationship`, if the target account is muted by the user, it returns a JSON object with a boolean property `notifications` rather than just `true` for the `muting` property.

If this gets merged, I'll make a PR to the docs repo to document these changes.


app/controllers/api/v1/mutes_controller.rb
```diff
@@ -8,10 +8,15 @@ class Api::V1::MutesController < Api::BaseController
   respond_to :json
 
   def index
-    @accounts = load_accounts
+    @data = @accounts = load_accounts
     render json: @accounts, each_serializer: REST::AccountSerializer
   end
 
+  def details
```
Gargron: How about using `Api::V2::MutesController` instead? e.g. `GET /api/v2/mutes`?

app/serializers/rest/mute_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class REST::MuteSerializer < ActiveModel::Serializer
+  include RoutingHelper
+
+  attributes :id, :account, :target_account, :created_at, :hide_notifications
+
+  def account
```
Gargron: Kinda of questionable whether this is necessary given that you can only ever look at your own mutes?

app/javascript/mastodon/features/ui/components/mute_modal.js
```diff
@@ -0,0 +1,103 @@
...
+            <FormattedMessage
+              id='confirmations.mute.message'
+              defaultMessage='Are you sure you want to mute {name}?'
+              values={{ name: <strong>@{account.get('acct')}</strong> }}
+            />
+          </p>
+          <p>
+            <label htmlFor='mute-modal__hide-notifications-checkbox'>
+              <FormattedMessage id='mute_modal.hide_notifications' defaultMessage='Hide notifications from this user?' />
+              <input id='mute-modal__hide-notifications-checkbox' type='checkbox' checked={notifications} onChange={this.toggleNotifications} />
```
Gargron: You could use the Toggle component for this

ekiru: Would you prefer that? `<input type=checkbox />` felt more appropriate to me, but I don't really have any explicit reason for that, and I suppose it is a little inconsistent with the rest of the UI.

app/serializers/rest/mute_serializer.rb
ekiru: I removed the account from the serializer.

app/controllers/api/v1/mutes_controller.rb
ekiru: Ah, good idea. I moved it to there in 2d003baddafbad335dc2cbf73c3d8db813d62229 and d24bfaae7f15f7548f949b6abb48c3c31d2c0a86

app/javascript/mastodon/features/ui/components/mute_modal.js
Gargron: Yes, i would prefer consistency. The line height looks a bit too small anyway and this might fix that.

ekiru:
Switched to a Toggle but I need to tweak the styling a bit to get it lined up properly; at present it looks like...

screen shot 2017-09-25 at 23 18 28

ekiru:
Now it's still using Toggle but looks a lot better than the screenshot in my comment before this one:

screen shot 2017-09-25 at 23 57 44

Cassolotl:
This is kind of a minor comment, but could there be more space between the two rows of text?

beatrix-bitrot:
The question now posed by @Gargron is this: is this the ideal way to structure such an endpoint from an API perspective? Once it's added to v2 other new endpoints will be expected to behave consistently with this one etc, so there is an issue of establishing a good precedent/pattern here.

I have invited reviews from @nightpool @unarist @ykzts so that we may have more perspectives on this issue.

app/javascript/mastodon/components/account.js
```diff
@@ -70,7 +80,18 @@ export default class Account extends ImmutablePureComponent {
       } else if (blocking) {
         buttons = <IconButton active icon='unlock-alt' title={intl.formatMessage(messages.unblock, { name: account.get('username') })} onClick={this.handleBlock} />;
       } else if (muting) {
-        buttons = <IconButton active icon='volume-up' title={intl.formatMessage(messages.unmute, { name: account.get('username') })} onClick={this.handleMute} />;
+        let hidingNotificationsButton;
+        if (muting.get('notifications')) {
+          hidingNotificationsButton = <IconButton active icon='bell' title={intl.formatMessage(messages.unmute_notifications, { name: account.get('username')  })} onClick={this.handleUnmuteNotifications} />;
```
akihikodaki: > `{ name: account.get('username')  }`

The last two spaces are redundant.

spec/controllers/api/v2/mutes_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    let(:mutes) { JSON.parse(response.body) }
+
+    it 'returns http success' do
+      expect(response).to have_http_status(:success)
+    end
+
+    it 'returns one mute' do
+      expect(mutes.size).to be(1)
```
akihikodaki: Omit parentheses around `be` arguments; keep the style consistent in your changes.

spec/models/concerns/account_interactions_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+
+describe AccountInteractions do
+  describe 'muting an account' do
+    before do
+      @me = Fabricate(:account, username: 'Me')
```
akihikodaki: Use `let` in new specs.

akihikodaki (changes requested):
I have made comments but they are kind of trivial. I think they should be easy to fix.

beatrix-bitrot:
thanks @akihikodaki for your review

@ekiru please address comments and resolve conflicts~

Gargron:
So my main concern with the new APIs is that until now our API has been more or less consistent in which entities it returns. There's the big three: Account, Status, Relationship. And all methods operated on Account models instead of the intermediary specific relationship models (I believe that the primary ID of a block does not matter, for example, what matters is account->target_account) which remained hidden. The v2 API exposes those, though only for Mute models. This means for it to become consistent, we'll need to also expose Block and Follow etc etc models. Which makes me uncomfortable.

There is a small precedent for doing it differently. The verify_credentials API method returns your own Account, but it also contains an additional property of your saved defaults.

Another alternative is modifying the Relationship entity such that instead of boolean `true` values it would return option hashes. For example, where previously it would've been

```js
{ muted: true }
```

now would be 

```js
{ muted: { includes_notifications: false } }
```

This would be a breaking API change too, though I suspect it wouldn't be a big deal because a non-empty hash is a truthy value in many languages.

ekiru:
I believe I've made all the changes requested by @akihikodaki. I've also rebased off tootsuite/master and fixed the merge conflicts.

I'll make the API changes once I'm more certain what is desired there.

app/javascript/mastodon/reducers/mutes.js
```diff
@@ -0,0 +1,29 @@
...
+export default function mutes(state = initialState, action) {
+  switch (action.type) {
+  case MUTES_INIT_MODAL:
+    return state.withMutations((state) => {
+      state.setIn(['new', 'isSubmitting'], false);
+      state.setIn(['new', 'account'], action.account);
+      state.setIn(['new', 'notifications'], true);
+    });
+  case MUTES_TOGGLE_HIDE_NOTIFICATIONS:
+    return state.setIn(['new', 'notifications'], !state.getIn(['new', 'notifications']));
```
akihikodaki: Please use `update` method

app/models/concerns/account_interactions.rb
```diff
@@ -70,8 +74,14 @@ def block!(other_account)
...
   end
 
-  def mute!(other_account)
-    mute_relationships.find_or_create_by!(target_account: other_account)
+  def mute!(other_account, notifications: nil)
+    notifications = true if notifications.nil?
+    mute = mute_relationships.create_with(hide_notifications: notifications).find_or_create_by!(target_account: other_account)
+    # When toggling a mute between hiding and allowing notifications, the mute will already exist, so the find_or_create_by! call will return the existing Mute without updating the hide_notifications attribute. Therefore, we check that hide_notifications? is what we want and set it if it isn't.
+    if mute.hide_notifications? != notifications
+      mute.hide_notifications = notifications
```
akihikodaki: Use `update!` method.

spec/controllers/api/v2/mutes_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    Fabricate(:mute, account: user.account, hide_notifications: false)
+    allow(controller).to receive(:doorkeeper_token) { token }
+  end
+
+  describe 'GET #index' do
+    before do
+      get :index, params: { limit: 1 }
+    end
+
+    let(:mutes) { JSON.parse(response.body) }
```
akihikodaki: Use `body_as_json` method.

db/migrate/20170716191202_add_hide_notifications_to_mute.rb
```diff
@@ -0,0 +1,5 @@
+class AddHideNotificationsToMute < ActiveRecord::Migration[5.1]
+  def change
+    add_column :mutes, :hide_notifications, :boolean, default: false, null: false
```
akihikodaki: You should set `default: true` to let all existing mutes to hide notifications.

akihikodaki (dismissed):
I found some problems I missed in the previous review.

db/migrate/20170716191202_add_hide_notifications_to_mute.rb
Gargron: Actually, this should be caught by strong_migrations.. Does that gem not work? Adding a not-null column is a blocking operation. We recently added some helper methods from gitlab to use when an operation like this must be performed

Gargron:
@ekiru Yeah, I would much prefer this approach to exposing standalone mute objects. v2 sounds good. I feel like "mute off" should just be "false", not an object, then if there *is* a mute, that's an object with extra options that can be true or false. This reminds me of some Rails options. What the actual properties are called I don't mind, if it's called hide_notifications in the database I don't see a point in renaming it in the API.

aschmitz:
Rebased on master, made all of @akihikodaki's suggested changes, as well as @Gargron's non-blocking column addition (although per https://github.com/tootsuite/mastodon/pull/5088#issuecomment-331930812 it wasn't likely to be a big issue either way, better safe than sorry I suppose).

@Gargron: I'm not 100% sure what you're talking about in terms of "mute off" being false: there are a few cases here:

* POST to `/api/v1/accounts/:account_id/mute`
  * With no body: creates or updates a mute for `account_id`, with `hide_notifications` set to true.
  * With a JSON object of `{"notifications":value}`: creates or updates a mute for `account_id`, with `hide_notifications` set to `value`.
* POST to `/api/v1/accounts/:account_id/unmute`
  * With no body: removes a mute for `account_id`. (This feels non-RESTful to me, but whatever, it's the existing API.)
* GET to `/api/v1/mutes`: Returns an array of Account objects which are muted.
* GET to `/api/v2/mutes`: Returns an array of Mute objects, each containing:
  * `id` (an ID)
  * `target_account` (an Account that's muted)
  * `created_at` (a timestamp)
  * `hide_notifications` (whether or not notifications are hidden)

I believe this appears to reflect what you'd discussed above. If there's something else you're thinking of, let me know. (The other potential reading is that you're recommending a new virtual property being added to Account in `mutes` responses, but that seems suboptimal, as it's effectively modifying the type that's being returned. We can talk about that if necessary, though.)

aschmitz:
It has been pointed out that perhaps @Gargron meant that the whole `/api/v2/mutes` should be discarded? That's apparently an acceptable solution to @ekiru, so I can effect that if so desired.

The relationships endpoint has the following behavior for the `muting` value:

* If not muting, the value is `false`.
* If muting and hiding notifications, the value is `{"notifications": true}`
* If muting and *not* hiding notifications, the value is `{"notifications": false}`

Which I believe is probably what @Gargron was referring to above.

nightpool:
Hows this going? Would love to get this merged post-2.0. Let me know if I can help in any way

Gargron (approved):

PR #5043, opened 2017-09-22
Fix the migration error when deprecated_preview_cards has been deleted

nullkal:
Fixes #5042

PR #5088, opened 2017-09-25
Make IdsToBigints (mostly!) non-blocking

aschmitz:
This pulls in GitLab's MigrationHelpers, which include code to make column changes in ways that Postgres can do without locking. In general, this involves creating a new column, adding an index and any foreign keys as appropriate, adding a trigger to keep it populated alongside the old column, and then progressively copying data over to the new column, before removing the old column and replacing it with the new one.

A few changes to GitLab's MigrationHelpers were necessary:

* Some changes were made to remove dependencies on other GitLab code.
* We explicitly wait for index creation before forging ahead on column replacements.
* We use different temporary column names, to avoid running into index name length limits.
* We rename the generated indices back to what they "should" be after replacing columns.
* We rename the generated foreign keys to use the new column names when we had to create them. (This allows the migration to be rolled back without incident.)

# Big Scary Warning

There are two things here that may trip up large instances:

1. The change for tables' "id" columns is not concurrent. In particular, the stream_entries table may be big, and does not concurrently migrate its id column. (On the other hand, x_id type columns are all concurrent.)
2. This migration will take a long time to run, *but it should not lock tables during that time* (with the exception of the "id" columns as described above). That means this should probably be run in `screen` or some other session that can be run for a long time. Notably, the migration will take *longer* than it would without these changes, but the website will still be responsive during that time.

These changes were tested on a relatively large statuses table (256k entries), and the service remained responsive during the migration. Migrations both forward and backward were tested.

Regrettably, the order of some columns changes due to this migration, so there are a lot of useless diff lines. I could exclude them from the change, but they'll come up the next time someone runs `rake db:create && rake db:migrate`. Similarly, some foreign keys get new names from the migration.


aschmitz:
Note that this replaces the old migration, because it has the same effect and can be used interchangeably with it: if you already ran that migration, there's no need to run the new one. (We could do the same for IdsToBigints2 , but it appears that change is unlikely to lock tables for very long? If it is, that's an easy change now that the rest of this is done.)

aschmitz:
The Travis failure appears to just be because of some unrelated flapping tests, but I can't re-queue it myself (without making a commit, anyway). The version on this branch of my fork (up to date with master) succeeds: https://travis-ci.org/aschmitz/mastodon/builds/279377510

akihikodaki:
I wonder who is willing to use this change. If there are few people who will use this change, maybe they should review and test by themselves. I doubt any reviewer can test and say it is really reliable.

nullkal:
I think it's OK as long as the tests for the helper functions are also pulled.

But, I'm worried that it is too large-scale method for almost all other instances in which the downtime caused by IdsToBigints is acceptable.

aschmitz:
I've started work on pulling in the GitLab MigrationHelper tests, but unfortunately they're rather dependent on parts of GitLab's testing infrastructure, so it's going to take a while to get everything working. In particular, getting ActiveRecord::Migration to be populated with a connection is a bit annoying. I'll also probably need to fully rewrite some tests that are tied to specific GitLab models.

Gargron:
You say that the migration will be non-blocking but very long running, but it still includes changes to `id`s, which cannot be optimized - they are mixed together. So this long running migration will occasionally deadlock.

I believe it must be split. Make a blocking but minimal migration (`id`s only) for next release. Then in the release *after that*, add the long-running purely non-blocking one. I believe that is probably the least disruptive way, I am just very very sorry we got these migrations into master before realizing the weight and discussing these implications.

aschmitz:
Are you suggesting that as separate *releases*, or simply separate migrations? If it's a matter of wanting to get all of the heavy locking bits out of the way first (which makes sense, as it's easier to schedule downtime then), we can re-order `INCLUDED_COLUMNS` to put them first without any trouble. If you meant doing it in actually-separate releases, I would be concerned about leaving the database in a state where IDs could be larger than the columns that reference them, but if we hold off on #4801 until then, it might not be a significant problem.

On a separate matter, can someone with access to a large instance get rough numbers for the number of rows in each table? I can try to set up a local test with realistic numbers for migrations and see how big of an impact the `id` columns will have. (The main reason they aren't included in this migration is the difficulty in swapping them in as primary keys. I have code to build a unique index for them, and reuse it as the new primary key index, but I still would need to copy the sequence value and test the heck out of that before I was comfortable with it.)

akihikodaki:
Here you can see number of statuses and users:
Mastodon instances
https://instances.social/list/advanced

mstdn.jp has the largest number of statuses. @nullkal

aschmitz:
Yep. However, I'm mostly interested in the tables that don't have
easily-derived counts: largely stream_entries, but also any other table
with an id column that needs changing and might be large.

akihikodaki:
For your information, counting takes much time, so you may refer to `pg_stats`. The catalog may also reveal some other interesting numbers.
https://www.postgresql.org/docs/9.6/static/view-pg-stats.html

aschmitz:
Ah, yes, that's why I was asking for rough numbers. Hopefully nobody ran
off to run `SELECT COUNT (*)` queries. :)

nullkal:
This is the (approximate) record numbers of all tables in mstdn.jp. I hope this information help.
```
mastodon=# SELECT schemaname,relname,n_live_tup
  FROM pg_stat_user_tables
  ORDER BY n_live_tup DESC;
 schemaname |        relname         | n_live_tup
------------+------------------------+------------
 public     | statuses               |   33597681
 public     | conversations          |   18729250
 public     | notifications          |   12705005
 public     | media_attachments      |    2254240
 public     | oauth_access_tokens    |    1911506
 public     | stream_entries         |     597439
 public     | favourites             |     507040
 public     | mentions               |     405300
 public     | statuses_tags          |     377983
 public     | preview_cards_statuses |     320344
 public     | accounts               |     287677
 public     | preview_cards          |     193562
 public     | users                  |     148902
 public     | subscriptions          |     148865
 public     | blocks                 |     138923
 public     | session_activations    |      72603
 public     | follow_requests        |      19857
 public     | follows                |      18868
 public     | tags                   |      16052
 public     | oauth_access_grants    |      11427
 public     | web_push_subscriptions |       9101
 public     | oauth_applications     |       7666
 public     | mutes                  |       6247
 public     | settings               |       4074
 public     | web_settings           |       1068
 public     | status_pins            |        205
 public     | conversation_mutes     |        201
 public     | reports                |         71
 public     | account_domain_blocks  |         16
 public     | imports                |          3
 public     | domain_blocks          |          1
 public     | schema_migrations      |          1
 public     | ar_internal_metadata   |          0
 public     | site_uploads           |          0
(34 rows)
```

Gargron:
>Are you suggesting that as separate releases, or simply separate migrations?

I suggest separate releases. I think it's important to not stretch the admins' good will, therefore the upgrade process should never be too long or too difficult. Spreading out over multiple releases means you get maybe 10min of downtime during the id upgrade, and then weeks later no downtime but 3 hours of the rest of the migration.

aschmitz:
Oof, that notifications table is going to be rough. I'll see if I can get
code together to migrate `id` columns in the background this evening. In
theory, the remaining steps are following up why the serial number is
destroyed and getting it recreated correctly, then a lot of testing.

akihikodaki:
I'm not for separated releases. Instances whose migration time matters would take different strategies, and imposing one strategy by separating releases would not make sense or make it even harder for admins to arrange their migrations.
The default strategy is fine to focus on small instances, whose migration time is trivial.

Gargron:
Ideas:

- Split the blocking and non-blocking migrations into separate files
- Print out some kinda messages like "Please bring your instance down now!" and "You can start your instance back up, now" using `say` (http://edgeguides.rubyonrails.org/active_record_migrations.html#changing-the-output-of-running-migrations)

aschmitz:
I don't think we can do this with any significant blocking, looking at the numbers from @nullkal.

Currently, the blocking sections are "just" `id` columns. The `notifications` table has an `id` column that needs to be adjusted, and it has over 12 million rows. That will take quite a long time even on its own. It also looks like we'll need to handle columns with foreign key constraints from other tables: the `oauth_access_tokens` table has an `id` column with a couple of those, and its ~2 million rows would also take a while to naïvely convert. I believe there's a way to handle this by expanding the GitLab migration helpers, though, and I'm looking into that.

Gargron:
@aschmitz 

>I don't think we can do this with any significant blocking

Sorry, did you mean to say "without any significant blocking"? Because the rest of the comment reads very pessimistic.

aschmitz:
Mm, more I meant that we should attempt to do this without blocking (alternatively "I don't think we can block during this migration"). The rest is pessimistic toward the idea of blocking for `id` column migrations, but I'm hopeful that I can make those work concurrently as well. It's just a bit tricky.

Gargron:
@aschmitz Alright. I wouldn't put too much faith into GitLab's helpers though. There are no magic gains to be had here, all they do is the only no-downtime process of changing columns: create a new column, write to both columns, make old column read-only, delete old column. And this is very hard to apply to primary IDs.

I suggest splitting the migration into two files. The first would convert the IDs, which would require downtime (probably). The second would upgrade everything else where downtime can be avoided. It's important to put the downtime first, because the non-blocking migrations will take longer.

aschmitz:
I've added two commits to perform non-blocking\* changes to all of the columns we need to change in this migration. This code is probably best considered a preview, and should be reviewed and further tested before it is merged (or anyone runs it, probably).

I'm leaving IdsToBigints2 alone, as the affected table is only 16k rows even in the most-used instance, for which any related downtime should measure (much) less than 5 seconds.

I understand we're in a bit of a rush to get some solution merged in. In an ideal world, I would do the following things:

1. More thoroughly test this with large amounts of data in at least the `statuses` and `notifications` tables.
2. Support breaking in the middle of the transaction and resuming: this is somewhat difficult because some necessary changes *cannot* happen in transactions (lest they block the table), but have no good way to recover from. For example, we must remove foreign keys pointing at account IDs and replace them with foreign keys pointing to the new account ID column, then swap the columns. If this procedure is interrupted in the middle, automatic recovery may be difficult. A "middle ground" solution may be to detect and recover from termination that happens while copying data, which is where the vast majority of time will be spent. We could print messages to the console regarding when it is "safe" to terminate the migration. Even this, though, is likely to take several hours to write and test sufficiently.
3. Port over GitLab's tests for their existing functionality.
4. Provide automated tests for new functionality.
5. More generically handle foreign keys (not just for primary keys), and remove expectations that primary keys are called `id`.

I can do any or all of those if desired. However, my guess is that a few additional eyes on d5d396a6eb768caade6b5b4e370b64fa0c0bea98 and 120885886320f4e4d330246b15e011c7a40b406b plus more thorough testing with sample data to ensure that things remain responsive as expected may be the most expedient solution to the current problem (for which I apologize: I had expected the migration would take a few minutes to run on large instances, not hours).

\*: These migrations are now as non-blocking as they can possibly be: a few transactions are started and nearly-immediately closed, to allow swapping columns. These don't appear to depend on the amount of data in the database, and take small fractions of a second each.

aschmitz:
(With that said, I won't be performing further testing tonight, as that seems best to do when rested. I will also have more limited time for development over the next two days.)

akihikodaki:
Thanks, the last two commits look good, but I have not tested.

aschmitz:
My plans for today include:

1. Making the MigrationHelpers able to "resume" a column rename, and integrating that into this migration. (Although because there's no way to tell "where we left off" without significant overhead, this means simply scrapping the most recent temporary column and restarting with that one.)
2. Sorting by Postgres' row count estimates to frontload the large table migrations. This has the effect of making it easier for administrators to see if/that the migration will take a long time, and let them abort it safely to retry later.
3. Adding some warning and progress text to the migration. I intend to put a short warning about the length of the migration and a (10-second?) pause before starting the work, then provide rough progress estimates using the row count estimates from earlier, if I can keep that from being too leaky in the MigrationHelper.
4. Testing the migration against a database loaded with a decently large sample dataset in four combinations:
   * To make sure that the app is responsive during the migration, and to make sure that interrupting and resuming works as intended.
   * Forwards and backwards

I should be able to get all of that done today. I will look at trying to pull in GitLab's tests for the code, but I'm not all that thrilled with them: they make sure that the code is calling the right functions, but don't test particularly well for correctness or to ensure that tables aren't locked extensively. (This isn't to complain about GitLab, those tests would be particularly difficult to do generically, more to indicate that I'm not sure how useful they'll be.)

If I get all of that done and a gist for filling a test database so someone else can replicate the tests, does that seem like a reasonable place to leave this PR, or are there are additional things we should do? (I'd be fine trying to get the GitLab tests in, but perhaps as a separate PR given that this appears to be blocking 2.0.)

aschmitz:
The migration now:

* Warns the user that the migration is very long and waits 10 seconds before proceeding.
* Can be interrupted while copying data. Re-running the migration skips any completed columns and restarts any that were partially completed.
* Provides status while copying data, including ETA and percentage (on a per-column basis, although not an overall number), as well as a note on whether or not it is safe to interrupt the migration (it indicates that interruption is safe if at least 10 seconds of data copying remain).

I tested it against a Postgres database loaded with fake data using this script: https://gist.github.com/aschmitz/13f7dc5f2964bac6ff7b6cf8c215fe05 . It creates 10% of the rows in @nullkal's estimated counts. Running in Vagrant talking with a copy of Postgres on an SSD over the network, it took approximately 12 minutes to complete, and remained responsive during that time.

I also tested the down migration (and found an issue with the earlier version, where the down migration ended up with no practical effect), as well as interrupting and resuming both the up and down migrations. The site remained responsive during all such changes, with minor pauses (possibly up to two seconds?) when changing the defaults of newly-added columns.

I'm inclined to call this good to go, as I'm not aware of any additional practical tests to run. If there are any additional concerns or comments on the code, I'd be happy to address them.

db/migrate/20170918125918_ids_to_bigints.rb
```diff
@@ -68,21 +68,52 @@ class IdsToBigints < ActiveRecord::Migration[5.1]
   ]
   INCLUDED_COLUMNS << [:deprecated_preview_cards, :id] if table_exists?(:deprecated_preview_cards)
```
Gargron: deprecated_preview_cards is *dead*, even if it exists. No need to update it. It's not used in the code. For anything.

Gargron (approved):
I'm gonna approve to signify that this is good work and I'm going to merge this but you really should forget about deprecated_preview_cards and not waste any possible time on it. I mean, it's in the name that the table is deprecated.

db/migrate/20170918125918_ids_to_bigints.rb
aschmitz: Sure, I'd be fine removing this line; I just added it in to keep up parity with #5043. (Is there a reason that table hasn't just been dropped yet?)

Gargron: It has been dropped! It's not in the schema. See rake task mastodon:maintenance:remove_deprecated_preview_cards

aschmitz: Ah, I suppose my confusion has come from it being created in the migrations and then not removed by them. (For consistency's sake, I often do `rake db:create && rake db:migrate` to make sure I'm not getting something weird from my own schema.rb.) I do see it being ignored by SchemaDumper, though.

ThibG:
I cannot comment this change itself, but I'm slightly worried about all those changes in `db/schema.rb` without a new migration script. I guess it's my own fault for running my prod server on master, but I already ran that migration script, meaning I now have a different schema. Also, are all the foreign key names present in `db/schema.rb` generated in a deterministic way?

aschmitz:
@ThibG I understand the concern. In practice I'm not expecting the differences to matter, but the fact that the foreign keys will have different names is an interesting one: if future migrations try to modify them *by name* (rather than by finding which foreign keys actually apply programmatically), they may have problems for users who used the old migration and not this one. Is it worth adding another migration to fix up any old foreign key names to the new ones? (The other differences should only be around column ordering in the database, which as I understand it has no practical effect, so I'm not particularly concerned about it.)

The generated foreign keys themselves are indeed deterministic but not the same as Rails' because Rails' generation isn't technically able to be referenced: https://github.com/tootsuite/mastodon/pull/5088/files#diff-3f490a2a7efe19b63d16461f015131e8R260

0xa:
As a late post-mortem on whether or not the migration should be split:
migrating mastodon.social took a long time, first because of an I/O load issue that crashed Docker, then because of lacking free space for the database. During this I've had to run db:migrate a few times, and a lot of time could have been saved by having more atomic migrations. The longest steps ran on 19M rows and took around one hour each. (and had to be done again on every db:migrate run)

aschmitz:
Hm, that's surprising. You were on the latest version of this PR (or coming
from master/a tagged release)? The I/O and disk usage I'd say aren't all
too unexpected that they'd be issues, but the latest MigrationHelper
shouldn't have re-migrated a column once it was fully migrated to the new
type. (Although if it kept failing in the middle of the longest one, it
would keep restarting that until it finished, which I guess could lead to
the type of behavior you were seeing.)

0xa:
master at the time, which is the rc1 i think.

PR #5091, opened 2017-09-25
Fix the 500.html rendering error in production environment

nullkal:
After #5067 is merged, I got an error `ActionView::Template::Error: Can't find common.css in /Users/nullkal/devel/mastodon/public/packs/manifest.json. Is webpack still compiling?` in the production environment (edge.mstdn.jp).

It seems the reason of it is that `manifest.json` is not reloaded after the webpack's asset compilation. So, I added the code to make it be reloaded manually.


nullkal:
Duplicate of #5090

PR #5095, opened 2017-09-25
Put Emoji in attachment

akihikodaki:
The definition of `tag` says:
> The key difference between `attachment` and `tag` is that the former implies association by inclusion, while the latter implies associated by reference.

ActivityVocabulary, section 4
https://www.w3.org/TR/activitystreams-vocabulary/#dfn-tag

Though emoji behaves like other tags already implemented in terms that they are inlined, it is irrelevant from the definition. When strictly following the specification, `attachment` is suitable for Emoji.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -61,8 +61,6 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
         process_hashtag tag, status
       when 'Mention'
         process_mention tag, status
-      when 'Emoji'
-        process_emoji tag, status
       end
     end
   end
@@ -81,14 +79,14 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     account.mentions.create(status: status)
   end
 
-  def process_emoji(tag, _status)
-    shortcode = tag['name'].delete(':')
+  def process_emoji(attachment, _status)
+    shortcode = attachment['name'].delete(':')
     emoji     = CustomEmoji.find_by(shortcode: shortcode, domain: @account.domain)
 
     return if !emoji.nil? || skip_download?
 
     emoji = CustomEmoji.new(domain: @account.domain, shortcode: shortcode)
-    emoji.image_remote_url = tag['href']
+    emoji.image_remote_url = attachment['url']
     emoji.save
   end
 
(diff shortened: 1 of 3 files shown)


Gargron:
I am not convinced:

The Emoji references the shortcode in the text, similarly to how mentions reference people and hashtags reference hashtags.

Attachments are resources meant to be created/stored with the message, like e-mail attachments. But emojis are not meant to be re-created for every post, but rather, reference an existing resource.

Therefore, emojis are "tags"

akihikodaki:
> The Emoji references the shortcode in the text, similarly to how mentions reference people and hashtags reference hashtags.

The specification says nothing about that.

> But emojis are not meant to be re-created for every post, but rather, reference an existing resource.

The description is convincing. Even though it does not have `id`, the combination of `shortcode` and `domain` identify an unique resource.
But, why doesn't it have `id` if Emoji is persistent, unique resource then?

Gargron:
@akihikodaki Link objects do not have `id`... Even though Mention refers to persistent Person, it does not have an `id`, using `href` to refer to the Person instead. Similarly, `Emoji` `href` references the emoji. Except in our case there is no object describing the emoji, it's just a simple file, therefore it's a file URL and not resource URI.

akihikodaki:
`href` is not `id` and `id` should actually be retrieved by dereferencing `href` in case of `Mention`.
The spec doesn't prohibit `href` to be an alias. (It does for `id`; `id` is included in referenced objects and putting an alias to `id` of an object effectively makes the object different.)

nightpool:
i'm tentatively in support of attachment here but also interested in other opinions. definitely seems like a borderline case.

akihikodaki:
It's not released yet. `master` is to break compatibility.

akihikodaki:
Apart from the legitimacy of `href` as the sole identifier, the current implementation reinventing the identification by using shortcode and the domain which delivered the emoji, and it is causing confusion.
It is because it does not follow Activity Streams way while it works for the purpose perfectly.
What is your opinion about using Activity Streams infrastructure to identify objects and make them "shared objects", as the spec of `tag` says?

cwebber:
I'm thinking about how I'd render attachments in MediaGoblin for instance, and I would definitely not expect emoji to be there... they don't seem like attachment "files".  `tag` is also a bit weird but less weird; we also have Mentions that go there, which are also things that tend to be embedded in the document in a sense.

akihikodaki:
I guess the majority of us do not have strong preference for `attachment` or prefer `tag`. I'll close this issue.

PR #5096, opened 2017-09-25
Add missing validations in ActivityPub::Activity::Create

akihikodaki:


Gargron (approved):

app/lib/activitypub/activity/create.rb
```diff
@@ -20,8 +20,11 @@ def process_status
     ApplicationRecord.transaction do
       @status = Status.create!(status_params)
 
-      process_tags(@status)
-      process_attachments(@status)
+      begin
+        process_tags(@status)
+        process_attachments(@status)
+      rescue ActiveModel::ValidationError
```
Gargron: Wait, I am against this. Which validation errors do you want to suppress? You could for example use `save` instead of `save!` and `create` instead of `create!` etc to get the same result without a blank error suppression!

Gargron (changes requested):

akihikodaki:
4f06c1f9b28ce440ff3210b1ad7dedad677b7a32: apply a change requested in https://github.com/tootsuite/mastodon/pull/5096#pullrequestreview-64932341, and also rescue `Addressable::InvalidURIError`.

Gargron (approved):

PR #5100, opened 2017-09-25
Fix #5059 - Stop processing payload if it's from local account

### LINKED ISSUE #5059: Deleted statuses get repopulated
Occasionally after deleting a status, users are seeing it reappear in their feeds:

image

Current theory is that this happens due to the signature-based federation stuff. something like:

1. post 1 from server A -> B
2. simultaneously:
    a. delete 1 federated from A -> B
    b. reply/fav/boost of 1 from B -> A
3. B's reply/fav/boost contains 1 embedded and signed, so we re-create 1 using the remote status content

The HTML is escaped because all "local" statuses have their HTML escaped, but we're creating it using the federated content, which is already HTML-formatted.

I think the root cause here is that we should never attempt to find_or_create local statuses when resolving them, only find them.

Gargron:


nightpool:
looks like a couple of process_collection_service specs need to be fixed to have remote users.

nightpool (approved):

PR #5102, opened 2017-09-25
Remove web UI warnings about OStatus privacy

Gargron:
- OStatus no longer used for private/direct posts

### DIFF
--- a/app/javascript/mastodon/features/compose/containers/warning_container.js
+++ b/app/javascript/mastodon/features/compose/containers/warning_container.js
@@ -1,51 +1,23 @@
 import React from 'react';
-import ImmutablePropTypes from 'react-immutable-proptypes';
 import { connect } from 'react-redux';
 import Warning from '../components/warning';
-import { createSelector } from 'reselect';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
-import { OrderedSet } from 'immutable';
 
-const getMentionedUsernames = createSelector(state => state.getIn(['compose', 'text']), text => text.match(/(?:^|[^\/\w])@([a-z0-9_]+@[a-z0-9\.\-]+)/ig));
-
-const getMentionedDomains = createSelector(getMentionedUsernames, mentionedUsernamesWithDomains => {
-  return OrderedSet(mentionedUsernamesWithDomains !== null ? mentionedUsernamesWithDomains.map(item => item.split('@')[2]) : []);
+const mapStateToProps = state => ({
+  needsLockWarning: state.getIn(['compose', 'privacy']) === 'private' && !state.getIn(['accounts', state.getIn(['meta', 'me']), 'locked']),
 });
 
-const mapStateToProps = state => {
-  const mentionedUsernames = getMentionedUsernames(state);
-  const mentionedUsernamesWithDomains = getMentionedDomains(state);
-
-  return {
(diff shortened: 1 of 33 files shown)


app/javascript/mastodon/features/compose/containers/warning_container.js
```diff
@@ -1,51 +0,0 @@
...
-  return {
-    needsLeakWarning: (state.getIn(['compose', 'privacy']) === 'private' || state.getIn(['compose', 'privacy']) === 'direct') && mentionedUsernames !== null,
-    mentionedDomains: mentionedUsernamesWithDomains,
-    needsLockWarning: state.getIn(['compose', 'privacy']) === 'private' && !state.getIn(['accounts', state.getIn(['meta', 'me']), 'locked']),
-  };
-};
-
-const WarningWrapper = ({ needsLeakWarning, needsLockWarning, mentionedDomains }) => {
-  if (needsLockWarning) {
-    return <Warning message={<FormattedMessage id='compose_form.lock_disclaimer' defaultMessage='Your account is not {locked}. Anyone can follow you to view your follower-only posts.' values={{ locked: <a href='/settings/profile'><FormattedMessage id='compose_form.lock_disclaimer.lock' defaultMessage='locked' /></a> }} />} />;
```
nightpool: this warning should still stay right?

`Your account is not {locked}. Anyone can follow you to view your follower-only posts`

app/javascript/mastodon/features/compose/containers/warning_container.js
```diff
@@ -1,51 +0,0 @@
...
-  } else if (needsLeakWarning) {
-    return (
-      <Warning
-        message={<FormattedMessage
-          id='compose_form.privacy_disclaimer'
-          defaultMessage='Your private status will be delivered to mentioned users on {domains}. Do you trust {domainsCount, plural, one {that server} other {those servers}}? Post privacy only works on Mastodon instances. If {domains} {domainsCount, plural, one {is not a Mastodon instance} other {are not Mastodon instances}}, there will be no indication that your post is private, and it may be boosted or otherwise made visible to unintended recipients.'
```
nightpool: I think there should still maybe be *some* type of warning about trusting remote servers? Not sure.

Definitely feels like there should be **some** copy in the main UI about this.

nightpool (dismissed):

app/javascript/mastodon/features/compose/containers/warning_container.js
Gargron: We only put it in because of GNU social. While it's technically true, it's just as true when you're writing an e-mail or messaging someone on XMPP. The remote server has to be trusted. In fact, logging in to a website is trusting a remote server with your data. I don't think the difference with Mastodon is now such that a special warning is warranted.

app/javascript/mastodon/features/compose/containers/warning_container.js
Gargron: Yeah but isn't it kinda obvious? Why do we have this warning?

akihikodaki: It is nice to have a first-time warning. The current warning should be left as is till then. There is no reason to change it now.

beatrix-bitrot (approved):
in AP posts are addressed to recipients, therefore an implementation would have to be malicious rather than incompetent to violate privacy stuff

SO

I think I agree with garg's reasoning for now

PR #5109, opened 2017-09-26
Implement EmailBlackList

utam0k:
I implemented admin settings screen can manage email blacklist.
I think that it is not good to manage EmailBlackList with environment variables.
BlacklistedEmailDomain is managed by the DB.

image


app/validators/blacklisted_email_validator.rb
```diff
@@ -13,6 +13,7 @@ def blocked_email?(value)
 
   def on_blacklist?(value)
     return false if Rails.configuration.x.email_domains_blacklist.blank?
+    return true if BlacklistedEmailDomain.block?(value)
```
akihikodaki: You put the check _after_ `return false if Rails.configuration.x.email_domains_blacklist.blank?`. It means it does not check the black list on the database will not be checked if the environment variable is empty. I guess that is not what you expect.

spec/fabricators/blacklisted_email_domain_fabricator.rb
```diff
@@ -0,0 +1,4 @@
+Fabricator(:blacklisted_email_domain) do
+  domain "example.com"
```
akihikodaki: You may use `Faker::Internet.domain_name`.

akihikodaki (changes requested):
Thank you for your pull request. That is what we need. However, I have found some problems in your patch, so please review them. Also I have some comments:

I don't think we should have `note` in this pull request. Neither domain blocks nor `EMAIL_DOMAIN_BLACKLIST` has the attribute. Adding `note` attribute is an irrelevant change and should be done in a different pull request, reasonably adding the attribute to domain blocks as well.

I also think the environment variable should be removed and provide a Rake task to import the blacklist from the environment variable. That is not a requirement and we can do that in a later pull request anyway.

utam0k:
@akihikodaki 
I would like to implement these when I have time.
> I don't think we should have note in this pull request. Neither domain blocks nor EMAIL_DOMAIN_BLACKLIST has the attribute. Adding note attribute is an irrelevant change and should be done in a different pull request, reasonably adding the attribute to domain blocks as well.
> 
> I also think the environment variable should be removed and provide a Rake task to import the blacklist from the environment variable. That is not a requirement and we can do that in a later pull request anyway.

spec/fabricators/blacklisted_email_domain_fabricator.rb
```diff
@@ -0,0 +1,3 @@
+Fabricator(:blacklisted_email_domain) do
+  domain { sequence(:domain) { |i| "#{i}#{Faker::Internet.domain_name}" } }
```
akihikodaki: `sequence` :+1:

spec/controllers/admin/blacklisted_email_domains_controller_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
```
akihikodaki: I have missed that specs lack `frozen_string_literal` comment. Add them also to specs (though the rule is not enforced well for existing specs…)

akihikodaki (changes requested):

spec/controllers/admin/blacklisted_email_domains_controller_spec.rb
utam0k: I also overlooked it.
Thank you for pointing out.
I fixed it earlier.

db/schema.rb
```diff
@@ -337,6 +358,7 @@
     t.integer "reblogs_count", default: 0, null: false
     t.string "language"
     t.bigint "conversation_id"
+    t.json "enquete"
```
ykzts: unrelated changed?

db/schema.rb
```diff
@@ -413,12 +435,17 @@
     t.boolean "otp_required_for_login", default: false, null: false
     t.datetime "last_emailed_at"
     t.string "otp_backup_codes", array: true
+    t.string "provider"
+    t.string "uid"
+    t.boolean "hide_oauth", default: false
```
ykzts: unrelated changed?

db/schema.rb
```diff
@@ -413,12 +435,17 @@
...
     t.string "filtered_languages", default: [], null: false, array: true
     t.bigint "account_id", null: false
     t.index ["account_id"], name: "index_users_on_account_id"
     t.index ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["filtered_languages"], name: "index_users_on_filtered_languages", using: :gin
+    t.index ["provider", "uid", "hide_oauth"], name: "index_users_on_provider_and_uid_and_hide_oauth"
+    t.index ["provider", "uid"], name: "index_users_on_provider_and_uid", unique: true
```
ykzts: unrelated changed?

config/locales/en.yml
```diff
@@ -152,6 +152,16 @@ en:
...
+    blacklisted_email_domains:
+      add_new: Add new
+      created_msg: Email domain black successfully created
+      destroyed_msg: Email domain black successfully deleted
+      domain: Domain
+      new:
+        create: Create block
+        title: New Email domain black
+      delete: Delete
+      title: BlackListed Email Domains
```
ykzts: Arrange the keys in alphabetical order.

nightpool:
The current copy—"blacklisted email domains"—doesn't explain what this feature is or what it's for. Maybe this should go under a "Registration" section instead, with some explanatory text?

Also blacklist is a compound word, not two words mangled by programming convention—it should be spelled as "Blacklist" or "blacklist" everywhere, as appropriate.

akihikodaki:
`EmailDomainBlock` would be better to align to `DomainBlock`. Though `EMAIL_DOMAIN_BLACKLIST` is fine for environment variables, now it sounds odd, considering it is a model.

akihikodaki:
Please change internal identifiers as well (e.g. name of model)

Gargron (approved):

akihikodaki (approved):

PR #5112, opened 2017-09-26
Fix #5104 - GET /api/v1/apps/verify_credentials to confirm app works

### LINKED ISSUE #5104: API to get client-app information
Strictly, client-app registration is not permanent. 
It may revoked , or all of server data may be lost.

But currently API users  can't verify the client registration is still alive or not.

Specially in case of oAuth2 browser authentication, 
the app just open URL in external browser.
app itself can't handle error about client-app registration.

then my app always re-register client app information.
It waste server's database rows.

if we have API to verify a client registration is still alive or not, we can reuse client information.
- the API request should requires client secret.
- the API response should not contains client secret.

* * * *

Gargron:


app/controllers/api/v1/apps/credentials_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V1::Apps::CredentialsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read }
+
+  respond_to :json
+
+  def show
+    render json: doorkeeper_token.application, serializer: REST::StatusSerializer::ApplicationSerializer
```
nightpool: maybe we need to move this out of StatusSerializer if it's now shared?

Gargron: If there's gonna be a 3rd place where this is used, then I'll agree. At 2, not really worth it.

nightpool (approved):

PR #5114, opened 2017-09-26
Validate id of ActivityPub representations

akihikodaki:


app/services/activitypub/fetch_remote_account_service.rb
```diff
@@ -5,8 +5,13 @@ class ActivityPub::FetchRemoteAccountService < BaseService
 
   # Should be called when uri has already been checked for locality
   # Does a WebFinger roundtrip on each call
-  def call(uri, prefetched_json = nil)
-    @json = body_to_json(prefetched_json) || fetch_resource(uri)
+  def call(uri, resource_option)
```
Gargron: I don't like the overloading of the 2nd argument here, maybe `options = {}` (or named arguments) would be better

Gargron: I think that's my main comment, because of how this is used in calls it's hard to understand what is happening.

Gargron (dismissed):

akihikodaki:
754b51c45d442fe974044a2a3c00da520a851d54: applied changes requested at https://github.com/tootsuite/mastodon/pull/5114#pullrequestreview-65240543, and also reviewed integration with other validations.
Additionally, `ActivityPub::FetchRemoteStatusService` no longer parses activities. The behavior is incompatible with this change in nature.

app/services/activitypub/fetch_remote_status_service.rb
```diff
@@ -4,36 +4,33 @@ class ActivityPub::FetchRemoteStatusService < BaseService
...
-    if %w(Note Article).include? @json['type']
-      {
-        'type'   => 'Create',
-        'actor'  => first_of_value(@json['attributedTo']),
-        'object' => @json,
-      }
-    else
-      @json
-    end
+    { 'type' => 'Create', 'actor' => actor_id, 'object' => @json }
```
Gargron: Why did you change this? What if the link leads to an Announce or Move? Then this would wrap it inside Create and pass on to handle like a status without type validation..?

akihikodaki: `id` should not lead to an Activity if it is conforming to ActivityPub… I'm reviewing the implementation to figure out when it could receive an Activity.

unarist: When we receive post / share activity via OStatus, we fetch AP version and use it if available.

https://github.com/tootsuite/mastodon/blob/c3f9c74719463be942b570c0a2e662775d33631c/app/lib/ostatus/activity/creation.rb#L12-L15

Then, FetchRemoteStatusService may be called with URI of an Announce activity. I think it's only case.

akihikodaki: Thank you for your quick answer. I'll amend the patch accordingly.

akihikodaki: In commit bfa808e0143f16ac6ce5dd1ff8bdc81e15148dd9, `OStatus::Activity::Creation` no longer delegates to ActivityPub because the provided ActivityPub representations are not signed while OStatus representations are.
If we still want the delegation, we need to sign activities in activity method of statuses controller, and to check the signature.

app/lib/ostatus/activity/creation.rb
```diff
@@ -67,7 +62,7 @@ def process_status
   end
 
   def perform_via_activitypub
-    [find_status(activitypub_uri) || ActivityPub::FetchRemoteStatusService.new.call(activitypub_uri), false]
+    [find_status(activitypub_uri) || ActivityPub::FetchRemoteStatusService.new.call(activitypub_uri, id: false), false]
```
unarist: Why don't you remove this method entirely?

Note: This method also does deduplication when OS payload handled just after AP payload, but it's not needed now because we use AP id for OS id for new posts...probably.

akihikodaki:
85b429d6fb0329d0c7a9129a249fa32e0e6b9384 applies a change suggested at https://github.com/tootsuite/mastodon/pull/5114#discussion_r141838621.

Gargron (approved):
I think/hope this is fine, though the code is still hard to read. The tests are almost unchanged and they pass, so this should work.

PR #5118, opened 2017-09-26
Replace self-rolled statsd instrumention with localshred/nsa

Gargron:
Improvements over old solution:

- controller stats
- db stats
- rails cache hit stats
- sidekiq job processing stats


beatrix-bitrot (approved):
mm good. i should try statsd 🤔

PR #5120, opened 2017-09-27
Update React to version v15.6.2

ykzts:
**Goodbye PATENTS**

- [React v15.6.2 - React Blog](https://facebook.github.io/react/blog/2017/09/25/react-v15.6.2.html)

> ### All Packages
> 
> - Switch from BSD + Patents to MIT license


Gargron:
#5119 Oops XD

But hey, yours works (for now) whereas mine is stuck waiting on deps to update

ykzts:
In order to upgrade to React v16, we have to wait for some libraries to be updated. But since v15.6.2 has no major changes other than license changes, it can be updated immediately. 😄

Gargron:
We're 16 now!

PR #5099, opened 2017-09-25
New error page graphic. Other error page improvements

Gargron:
oops

- 500.html generated with admin-set default locale if set
- Error page `<title>` includes Mastodon site title
- 500 title changed to "This page is not correct" (ref: <https://www.youtube.com/watch?v=2VCAP_seh1A>)
- 500 content appended with "on our end" to make clear it's not user's fault

PR #5127, opened 2017-09-28
Use nullDB during asset compilation

danhunsaker:
This change allows Nanobox users to continue deploying Mastodon instances, and also speeds up the asset compilation phase for other users who might want it. It adds the ActiveRecord nullDB driver, and lets users switch to it by setting an environment variable (`DB_ADAPTER=nulldriver`), which can be done as part of the `rake assets:precompile` command. For Nanobox users, it relieves the need for an actual database connection during the `compile` step (part of the deploy process which prepares the code for upload to the server). For all users, it speeds up asset compilation by converting all DB calls to no-ops, which is considerably faster than talking to an actual Postgres DB during a phase which doesn't actually need DB access to begin with (even though Rails loads itself entirely for the process anyway).

Blocks tootsuite/documentation#403

### DIFF
--- a/boxfile.yml
+++ b/boxfile.yml
@@ -44,7 +44,7 @@ run.config:
 
 deploy.config:
   extra_steps:
-    - NODE_ENV=production bundle exec rake assets:precompile
+    - NODE_ENV=production DB_ADAPTER=nulldb bundle exec rake assets:precompile
   transform:
     - "sed 's/LOCAL_HTTPS=.*/LOCAL_HTTPS=true/i' /app/.env.nanobox | envsubst > /app/.env.production"
     - |-
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
 
 before_script:
   - bundle exec rake parallel:create parallel:load_schema parallel:prepare
-  - bundle exec rails assets:precompile
+  - bundle exec rails assets:precompile DB_ADAPTER=nulldb
   - ln -s /usr/bin/x86_64-linux-gnu-g++-6 "$HOME/g++"
 
 script:
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ gem 'paperclip', '~> 5.1'
 gem 'paperclip-av-transcoder', '~> 0.6'
 
 gem 'active_model_serializers', '~> 0.10'
+gem 'activerecord-nulldb-adapter'
 gem 'addressable', '~> 2.5'
 gem 'bootsnap'
 gem 'browser'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -39,6 +39,8 @@ GEM
       activemodel (= 5.1.4)
       activesupport (= 5.1.4)
       arel (~> 8.0)
+    activerecord-nulldb-adapter (0.3.7)
+      activerecord (>= 2.0.0)
     activesupport (5.1.4)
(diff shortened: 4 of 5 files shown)


nullkal:
After #5067, `assets:precompile` requires the true database connection.

nullkal (changes requested):
After #5067, `assets:precompile` requires the *true* database connection.

This is because the generation of 500 error page use the instance name in the database after that.

danhunsaker:
OK, fair enough. I'll adjust the PR to also re-run the `bundle exec rake assets:generate_static_pages` task after the DB is accessible. That's the only portion that needs DB access, yes?

nullkal (approved):

nightpool:
Wait, which part of 5067 requires the database? I compile my assets locally
and upload them—am I going to run into issues?

Gargron:
@nightpool #5099 adds `Setting.site_title` call to the `<title>` of the error page layout.

I thought people would like that. I believe that maybe it should be removed now though.

danhunsaker:
Not sure why Travis builds are failing, here... Thoughts?

Gargron:
Okay i removed the need for db in master

boxfile.yml
```diff
@@ -56,6 +56,7 @@ deploy.config:
         erb /app/nanobox/nginx-web.conf.erb > /app/nanobox/nginx-web.conf
         erb /app/nanobox/nginx-stream.conf.erb > /app/nanobox/nginx-stream.conf
     - touch /app/log/production.log
+    - bundle exec rake assets:generate_static_pages
```
nightpool: this can be removed.

danhunsaker: Commit removed.

nightpool:
Looks like Travis is still failing, although I'm not 100% sure why. Do those commands work locally on your computer?

My current working theories are either 1) the DB_ADAPTER environment variable isn't falling back to postgres correctly or b) the nulldb adapter itself is interfering with the schema load process.

danhunsaker:
I'll double check that in a bit. Away from the keyboard for the moment.

akihikodaki:
For anyone interested in the test failure: you can reproduce it with the following command:
```shell
RAILS_ENV=test ./bin/rails db:abort_if_pending_migrations
```

akihikodaki (changes requested):
I figured out what is wrong with CI and how to fix.
You should prevent from requiring the adapter in `config/application.rb` if `DB_ADAPTER` is not nulldb. Otherwise it hooks `ActiveRecord::Schema` and disrupts db tasks.

akihikodaki:
I had an experiment to remove `activercord/railties` (and optionally Redis) dependency from `assets:precompile`. Now I prefer the approach in the experiment because:
* It showed it still calls `Setting.site_title` via `title` method. Hopefully the approach eliminates such bugs.
* It was not so difficult.
* No additional dependency (it actually reduces dependencies required to build assets.)

danhunsaker:
Removing the DB dependency during asset precompile is what I originally set out to do, but I'm not as familiar with Rails as I'd like. How does one go about removing dependencies from a Rake task?

danhunsaker:
@akihikodaki I think your approach of removing railties is superior to this one.

Between that, and since builds are working again (if a bit on the slow side, still) now that the DB calls have been removed from the asset compilation phase, I'm going to close this one.

PR #5128, opened 2017-09-28
Flush body when POST requests

abcang:
When sending a POST request to another instance by DeliveryWorker etc., body is not read. So, even though the request has been completed, the connection will last for a while. This PR solves this problem by reading the body by calling the flush method.

Related: https://github.com/httprb/http/pull/125

### DIFF
--- a/app/services/send_interaction_service.rb
+++ b/app/services/send_interaction_service.rb
@@ -12,7 +12,7 @@ class SendInteractionService < BaseService
 
     return if !target_account.ostatus? || block_notification?
 
-    delivery = build_request.perform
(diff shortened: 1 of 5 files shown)


Gargron:
Question: If we don't read the body at all, is there any reason to, well, ready the body before closing socket? Is there a way to just close socket even quicker?

Gargron (approved):

abcang:
I'm sorry, I did not investigate enough. Since I found the close method earlier, I think that this is probably usable.

https://github.com/httprb/http/blob/v2.2.2/lib/http/client.rb#L87

Issue #5048, opened 2017-09-22
Profile of remote accounts not refreshed until direct interaction


ThibG:
It is very common to get toots from remote users without following those users (boosts, replies, mentions)… in that case, the account is resolved and the profile populated when that user's first toot is retrieved by the instance.

However, only very specific interactions may then cause those profiles to update: getting a salmon slap (deprecated with ActivityPub) from them, following them, getting a follow request from them, or manually fetching one of their toots.

In particular, that means the profile of people noone on your instance follow or directly interact with but are frequently boosted from people you follow won't ever update. As a striking example, here is how `@mastodon@mastodon.social` looks like to me: 
![oldmasto](https://user-images.githubusercontent.com/384364/30759871-3211abe6-9fd8-11e7-99e8-af0fe937edd8.png)

ThibG:
Fixed by d37a56c07cdc475f9a0f5a87d65d1e258c48644b and 7bea1530f4b396ae384502b3fcbf8d34f22005e1

PR #5129, opened 2017-09-28
If HTTP signature is wrong and webfinger cache is stale, retry with resolve

Gargron:
If the signature could not be verified and the webfinger of the account was last retrieved longer than the cache period, try re-resolving the account and then attempting to verify the signature again

### DIFF
--- a/app/controllers/concerns/signature_verification.rb
+++ b/app/controllers/concerns/signature_verification.rb
@@ -44,6 +44,15 @@ module SignatureVerification
     if account.keypair.public_key.verify(OpenSSL::Digest::SHA256.new, signature, compare_signed_string)
       @signed_request_account = account
       @signed_request_account
+    elsif account.possibly_stale?
+      account = account.refresh!
+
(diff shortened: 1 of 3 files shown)


ClearlyClaire (changes requested):
While it doesn't look *wrong* to me, I fail to see what it solves. Indeed, `account_from_key_id` already updates remote data (provided old enough `last_webfingered_at`) by calling `ResolveRemoteAccount`.

On a side note, I have noticed the following files include `SignatureVerification` without seemingly using it:
- `app/controllers/accounts_controller.rb`
- `app/controllers/stream_entries_controller.rb`

Gargron:
@ThibG Uh no, account_from_key_id only calls ResolveRemoteAccountService for OStatus keyId format which is not used by Mastodon anymore.

ThibG:
@Gargron ah, right, and the ActivityPub codepath indeed tries to get the local account info first… hm I think I'd prefer calling `ResolveRemoteAccountService` unconditionally (see #5048), but I'm fine with your proposed change, then!

ClearlyClaire (approved):

PR #5131, opened 2017-09-28
After repeated delivery failures, give up on inboxes

Gargron:
**Note:** This can be probably written in a much better way, ideas on how to structure the code are welcome

What makes us give up on an inbox:

- 7 days of successive DeliveryWorker *exhaustions* (i.e. after all natural Sidekiq retries for a job have been exhausted)

What returns an inbox back to normal:

- A successful delivery cancels it out
- An incoming delivery from account of the inbox cancels it out

### DIFF
--- a/app/controllers/activitypub/inboxes_controller.rb
+++ b/app/controllers/activitypub/inboxes_controller.rb
@@ -32,6 +32,7 @@ class ActivityPub::InboxesController < Api::BaseController
     end
 
     Pubsubhubbub::UnsubscribeWorker.perform_async(signed_request_account.id) if signed_request_account.subscribed?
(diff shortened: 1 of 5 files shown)


app/lib/request.rb
```diff
@@ -85,6 +85,6 @@ def timeout
   end
 
   def http_client
-    HTTP.timeout(:per_operation, timeout).follow
+    HTTP.timeout(:per_operation, timeout).follow(max_hops: 2)
```
unarist: Is this change related to this topic?

beatrix-bitrot (approved):
nice!

beatrix-bitrot:
not a blocking thing but this would be nice to have as an environment variable and easy too

PR #5135, opened 2017-09-28
Improve performance of modal and swipe animations

nolanlawson:
This PR improves the framerate of the modal animations, both when opening the modal and swiping between items in the `react-swipeable-views`. It's also a design change, but I believe the performance improvements outweigh the design changes.

It's a bit hard to explain with words, so I made [a small website](https://nolanlawson.s3.amazonaws.com/mastodon/index.html) showing the before-and-after for 1) mobile Chromium 63 on Nexus 5, 2) mobile Firefox Nightly 57 on Nexus 5, and 3) Safari 10 on an iPod Touch 6th generation.

The performance difference is most obvious with Firefox and least obvious with Chrome. Also note that these videos aren't a perfect representation because the recording itself adds some overhead. But you should still be able to see the framerate difference.

I'm going to comment inline to explain the changes, but some good background reading is [High Performance Animations](https://www.html5rocks.com/en/tutorials/speed/high-performance-animations/) which basically explains that the only values that can be animated cheaply are `opacity` and `transform`.

### DIFF
--- a/app/javascript/mastodon/features/ui/components/media_modal.js
+++ b/app/javascript/mastodon/features/ui/components/media_modal.js
@@ -84,14 +84,17 @@ export default class MediaModal extends ImmutablePureComponent {
       return null;
     }).toArray();
 
+    const containerStyle = {
+      alignItems: 'center', // center vertically
+    };
+
     return (
       <div className='modal-root__modal media-modal'>
         {leftNav}
 
         <div className='media-modal__content'>
           <IconButton className='media-modal__close' title={intl.formatMessage(messages.close)} icon='times' onClick={onClose} size={16} />
-
-          <ReactSwipeableViews onChangeIndex={this.handleSwipe} index={index} animateHeight>
+          <ReactSwipeableViews containerStyle={containerStyle} onChangeIndex={this.handleSwipe} index={index}>
             {content}
           </ReactSwipeableViews>
         </div>
--- a/app/javascript/mastodon/features/ui/components/modal_root.js
+++ b/app/javascript/mastodon/features/ui/components/modal_root.js
@@ -1,7 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import TransitionMotion from 'react-motion/lib/TransitionMotion';
-import spring from 'react-motion/lib/spring';
 import BundleContainer from '../containers/bundle_container';
 import BundleModalError from './bundle_modal_error';
 import ModalLoading from './modal_loading';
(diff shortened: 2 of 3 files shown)


app/javascript/mastodon/features/ui/components/media_modal.js
```diff
@@ -84,14 +84,17 @@ export default class MediaModal extends ImmutablePureComponent {
...
+
     return (
       <div className='modal-root__modal media-modal'>
         {leftNav}
 
         <div className='media-modal__content'>
           <IconButton className='media-modal__close' title={intl.formatMessage(messages.close)} icon='times' onClick={onClose} size={16} />
-
-          <ReactSwipeableViews onChangeIndex={this.handleSwipe} index={index} animateHeight>
+          <ReactSwipeableViews containerStyle={containerStyle} onChangeIndex={this.handleSwipe} index={index}>
```
nolanlawson: This is the major design change in this PR. `animateHeight` is disabled because `height` cannot be animated cheaply. It may be kinda-high-framerate on powerful desktop machines, but on mobile devices it looks pretty bad.

Because the height is not changing, I add `align-items: center` to the flexbox container so that the items are vertically centered. The design tradeoff is that there is now a letterboxing on the top and bottom of some images. I actually think this is an improvement, though, because it looks better when you're tapping-and-dragging to swipe left and right (IMO).

app/javascript/mastodon/features/ui/components/modal_root.js
```diff
@@ -94,38 +95,31 @@ export default class ModalRoot extends React.PureComponent {
...
+                (<BundleContainer fetchComponent={MODAL_COMPONENTS[type]} loading={this.renderLoading(type)} error={this.renderError} renderDelay={200}>
+                  {(SpecificComponent) => <SpecificComponent {...props} onClose={onClose} />}
+                </BundleContainer>) :
+              null
+            }
           </div>
-        }
-      </TransitionMotion>
+        </div>
+      </div>
```
nolanlawson: The problem with `react-motion` is that it animates by updating the `opacity` in `rAFs`. This means that JavaScript needs to run on every frame, which is a waste given that this should be a simple fade-in/fade-out animation. So I replaced it with a simple CSS transition defined below in `components.scss`.

The animation is not only simpler but also much more performant. `react-motion` might be useful for fancy animations, but for the simple stuff it is overkill.

nolanlawson:
BTW you can compare the before and after in your own browser using https://malfunctioning.technology (before) and https://horribly.malfunctioning.technology (after).

app/javascript/mastodon/features/ui/components/modal_root.js
nolanlawson: Note though that there is occasionally some minor flickering only in Chrome and only when the modal does a "fade in." I believe this is due to the image loader and its `<canvas>`+`` logic, and I would like to address this in another PR. In any case, I think Safari and Firefox are a huge improvement and Chrome is still a slight improvement.

sorin-davidoi (approved):
Great improvement indeed in Firefox, thank you for this! 🚀

PR #5137, opened 2017-09-28
Add image descriptions to title attribute to view on mouse hover/long-press.

ghost:
A small change to add a ``title`` attribute alongside the alt description of images so that they're also visible on mouse-over or by long-pressing the preview on mobile devices.

screenshot 105
screenshot_20170928-142801


app/javascript/mastodon/features/ui/components/media_modal.js
```diff
@@ -76,9 +76,9 @@ export default class MediaModal extends ImmutablePureComponent {
       const height = image.getIn(['meta', 'original', 'height']) || null;
 
       if (image.get('type') === 'image') {
-        return <ImageLoader previewSrc={image.get('preview_url')} src={image.get('url')} width={width} height={height} alt={image.get('description')} key={image.get('preview_url')} />;
+        return <ImageLoader previewSrc={image.get('preview_url')} src={image.get('url')} width={width} height={height} alt={image.get('description')} title={image.get('description')} key={image.get('preview_url')} />;
```
Gargron: This is useless, I am not sure if the ImageLoader component expects a title prop.

app/javascript/mastodon/features/ui/components/media_modal.js
```diff
@@ -76,9 +76,9 @@ export default class MediaModal extends ImmutablePureComponent {
...
       } else if (image.get('type') === 'gifv') {
-        return <ExtendedVideoPlayer src={image.get('url')} muted controls={false} width={width} height={height} key={image.get('preview_url')} alt={image.get('description')} />;
+        return <ExtendedVideoPlayer src={image.get('url')} muted controls={false} width={width} height={height} key={image.get('preview_url')} alt={image.get('description')} title={image.get('description')} />;
```
Gargron: This is likewise useless, the video player definitely does not expect a title attribute and even if it did, a title tooltip over a video is pretty weird.

Gargron (changes requested):
I think that title tooltips are maybe not the best way to display this information, especially considering they cannot be equally applied to images *and* videos

I also don't know if screen readers are going to read alt and title twice on the images, which would not be desired.

Gargron (approved):

beatrix-bitrot:
for what it's worth, when i asked i was told screen readers will read both alt and title

PR #5141, opened 2017-09-29
Option that requests admin's approval on registration

Nanamachi:
related issue: #146 

Add option that requests approval to admin on registration.
This may be, as haniawni mentioned in above issue, beneficial for semi-closed community like club or small society.

- Add new option 'Restricted Registration' on Site Settings.
- If this option is true, 
  - email that notifies a newbie has been arrived and requests approvals will be sent to all admins.
  - admins can approve them via admin/accounts.
  - hint that says "Your account will be activated after admin's approval" will be displayed on registration page.
- else,
  - newbies are automatically approved. This approval is not cancelled if the option turn on after registration.
- New column 'Approved' added on admin/account table.
- By default, this option is set false.


app/controllers/admin/settings_controller.rb
```diff
@@ -15,12 +15,14 @@ class SettingsController < BaseController
       timeline_preview
       bootstrap_timeline_accounts
       thumbnail
+      need_approval
```
nightpool: `require_approval` might make more sense as a name

app/models/user.rb
```diff
@@ -143,6 +149,36 @@ def web_push_subscription(session)
...
+
+  def request_approval
+    if Setting.need_approval
+      User.admins.each do |admin|
+        UserMailer.new_user_waiting_for_approval(admin, self, 'hogehoge').deliver_later
+      end
+    else
+      approve
+    end
+  end
```
nightpool: we would generally put code like this into a RequestApprovalService and the RegistrationsController instead of using a callback.

app/models/user.rb
```diff
@@ -31,6 +31,9 @@
 #  last_emailed_at           :datetime
 #  otp_backup_codes          :string           is an Array
 #  filtered_languages        :string           default([]), not null, is an Array
+#  approval_sent_at          :datetime
```
nightpool: Seems unused?

db/migrate/20170929051000_add_approved_to_user.rb
```diff
@@ -0,0 +1,7 @@
+class AddApprovedToUser < ActiveRecord::Migration[5.1]
+  def change 
+    add_column :users, :approval_sent_at, :datetime
```
nightpool: Seems unused?

nightpool (dismissed):
hey, great to see this pull request! This looks like a great feature and I know there's definitely interest in it. Added some comments about ways you can better match the style of Mastodon's codebase.

nightpool:
[Looks like this broke some tests as well:](https://travis-ci.org/tootsuite/mastodon/builds/281114655?utm_source=github_status&utm_medium=notification)

image

Nanamachi:
Thanks for your review, nightpool!
I've fixed along with your advice. How is it?

nightpool:
Wow, sorry! I totally didn't see your response, my apologies.

app/services/request_approval_service.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RequestApprovalService < BaseService
+  def call(user)
+    return true if user.id.nil?
+
+    if Setting.require_approval
+      User.admins.each do |admin|
```
Gargron: If you look at the mailer for reports, usually the recipients (admins) are queried from within the mailer, rather than the other way around like here.

app/services/request_approval_service.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RequestApprovalService < BaseService
+  def call(user)
+    return true if user.id.nil?
```
Gargron: Is the return value used for anything? I don't think so. (i.e. should simply return)

app/services/request_approval_service.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    if Setting.require_approval
+      User.admins.each do |admin|
+        UserMailer.new_user_waiting_for_approval(admin, user).deliver_later
+      end
+    else
+      user.approve
```
Gargron: Since this is a service object, I suggest doing the exclamation mark stuff so it raises an error, rather than relying on return values. So i.e. rename method to `approve!`, use `update!` from within

Gargron (changes requested):
👌 minor changes needed

Gargron:
I feel like with the introduction of limited use invites this PR is no longer a necessary addition. Do you agree?

PR #5143, opened 2017-09-29
Remove react-sizeme

nolanlawson:
The issue with `react-sizeme` is that it works by listening to scroll events and checking `getStyle()` on the element, which forces layout:

screenshot 2017-09-28 23 47 34

Also this happens for every single media in the timeline, even including those that aren't `standalone` and thus don't even need to know the `width`.

As it turns out, though, I think we can just omit setting the image `height` and it will have the same effect, because browsers will adjust the height of the image automatically. I checked in desktop Firefox, Safari, Chrome, and Edge, and it appears standalone images are being correctly uncropped.

Also this PR seems to fix a bug I noticed in GIFs ([mp4 in this case](https://malfunctioning.technology/@nolan/124950)) getting cut off:

screenshot 2017-09-28 23 40 09
screenshot 2017-09-28 23 40 36

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -207,19 +207,42 @@ export default class MediaGallery extends React.PureComponent {
     this.props.onOpenMedia(this.props.media, index);
   }
 
+  handleRef = (node) => {
+    if (node && this.isStandaloneEligible()) {
+      // offsetWidth triggers a layout, so only calculate when we need to
+      this.setState({
+        width: node.offsetWidth,
+      });
+    }
+  }
+
+  isStandaloneEligible() {
+    const { media, standalone } = this.props;
+    return standalone && media.size === 1 && media.getIn([0, 'meta', 'small', 'aspect']);
+  }
+
   render () {
-    const { media, intl, sensitive, height, standalone } = this.props;
+    const { media, intl, sensitive, height } = this.props;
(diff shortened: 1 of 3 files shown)


Gargron:
The height is required for the media spoiler, not the image

nolanlawson:
My bad. OK I will look into fixing this.

nolanlawson:
OK, I fixed the aspect ratio in "sensitive" mode. I checked various combinations of images and MP4s, with and without sensitive mode, and everything appears to work the same as before. The only difference is we're only calculating `offsetWidth` when we absolutely need to. This also still fixes the issue with MP4s getting cut off.

I've deployed this to https://malfunctioning.technology in case you want to check.

Gargron:
@nolanlawson I had a wild idea that if the browser treats canvas the same as img (in resizing it fluidly), we could create a canvas with the width/height from metadata to make sure the media spoiler stretches over it.

nolanlawson:
Hmmm unfortunately I don't believe it does; I think images have a "natural" sizing that come from when they're decoded, but canvas doesn't. There may be some trick you can do with background-image and background-sizing though.

It'd be nice if we could avoid the `offsetWidth` calculation entirely, but as-is it only happens for the detail view, so at least it doesn't interfere with timeline scrolling.

Gargron (approved):

nolanlawson:
Yay! :smiley: :tada: #perfmatters

Issue #5107, opened 2017-09-26
Custom emoji for long-run


akihikodaki:
The implementation of custom emoji is optimized to provide additional server flavour quickly. However, it has several problems when considering to run the feature for a long time:
~~1. limiting choices of instances by emojis~~
1. Sources of copied emojis are unknown
2. limited namespace
3. limiting ability to create emojis for those who are not admins, don't know how to launch an instance, cannot ask admins to add an emoji

~~1 can effectively construct some kind of norms that _users should be in a particular instance so that we can use common emojis_. For example, if an emoji named `:catnip:` become quite a popular as if it is an Internet meme, people interested in it would move to the page. They may even say "why don't you come here to use the cool emoji?" That is against decentralization.~~ Solved with #5231.

For details of 1, see #5287.

2 is caused by the broad namespace and the write-once architecture. The emojis are shared in an instance (some instances are quite a large even now), and the emoji cannot be removed, replaced, or renamed without breaking the statuses already made.

3 matters when Mastodon accepts users who are far from Web development community. Such users can be easily found in some large instances like mstdn.jp or Pawoo.

Here I suggest solutions corresponding to those challenges.

# ~~Remote emoji fetching feature (breaking, #5145)~~
~~It was suggested by @Gargron at https://mastodon.social/@Gargron/19389045. Though he mentioned it as "copy", I consider it should be implemented by tracking the source of emojis in order to cache or block emojis, and to potential copyright dispute.
This would be realized with a breaking change of federation and database:~~
* ~~ActivityPub's `Emoji` will refer to an `Image` as icon, where the `Image` has an `id` to track the source.~~
* ~~OStatus's `href` of `link[rel=emoji]` would now point an `entry` which has an `id` and `link[rel=enclosure]`.~~
* ~~The database will have `shortcodes` table which references `custom_emojis`.~~
* ~~`custom_emojis` will have `uri` column.~~

# Scoped shortcode (breaking)
shortcode is a currently permanent, domain-wide identifier. However, it is seen by users and connecting the internal implementation and the UI without any reason.
By limiting the scope of shortcode to statuses, statuses would keep consistent representation of emojis.
This would also involve a breaking change on the federation and the database:
* ActivityPub's `Emoji` would be put in `attachment` where it is currently put in `tag`.
* (no change for OStatus)
* `shortcodes` table will belong to `statuses`.

# Emoji creation for accounts (#5072)
This does not involve any federation change.

# `icon` attribution
[...]

PR #5145, opened 2017-09-29
Implement emoji import feature

akihikodaki:
shortcode as identifier of custom emojis is a barrier when implementing emojis shared by multiple instances because it is not global identifier.
This is a part of a change suggested at #5107. ~~It is not tested well yet.~~

### DIFF
--- a/app/controllers/admin/custom_emojis_controller.rb
+++ b/app/controllers/admin/custom_emojis_controller.rb
@@ -4,17 +4,48 @@ module Admin
   class CustomEmojisController < BaseController
     def index
       @custom_emojis = CustomEmoji.local
+      @status = Status.new
     end
 
-    def new
+    def import_form
+      @custom_emoji = CustomEmoji.new
+      render_import_form_with_status_url import_form_params.require(:url)
+    rescue ActionController::ParameterMissing
+      redirect_to action: :index, flash: { error: I18n.t('admin.custom_emojis.status_unspecified_msg') }
+    end
+
+    def upload_form
       @custom_emoji = CustomEmoji.new
       @custom_emoji.build_custom_emoji_icon
     end
 
-    def create
+    def import
+      super_custom_emoji = CustomEmoji.find(import_params.require(:super_id))
+    rescue ActionController::ParameterMissing
+      @custom_emoji = CustomEmoji.new
+      flash.now[:error] = I18n.t('admin.custom_emojis.emoji_unspecified_msg')
+      render_import_form_with_status_id params.require(:status).require(:id)
+    rescue ActiveRecord::RecordNotFound
+      @custom_emoji = CustomEmoji.new
+      flash.now[:error] = I18n.t('admin.custom_emojis.emoji_not_found_msg')
+      render_import_form_with_status_id params.require(:status).require(:id)
+    else
       @custom_emoji = CustomEmoji.new(
-        custom_emoji_icon: CustomEmojiIcon.new(image: resource_params.dig(:custom_emoji_icon, :image)),
-        shortcode: resource_params[:shortcode]
+        custom_emoji_icon: super_custom_emoji.custom_emoji_icon,
+        shortcode: import_params[:shortcode] || super_custom_emoji.shortcode
+      )
+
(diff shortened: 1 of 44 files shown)


akihikodaki:
I tested the changes and added fixes specs with commit 88491e56a500d20adeee837d1ca436d53017047e. Now it is ready to review.

akihikodaki:
Implemented emoji import feature with commit aca06f341ca28ac187fc73d108e8caaf166ba93e.

akihikodaki:
01a406c892247738936b6c02eab2751d0cbcd2b6, 71b65493abb59609defc74952461f8472a5c21be: irrelevant but later specs are dependent on them. A commiter may cherry-pick them earlier.
9533624bfb1d9b1d51990a5f751449e85e0141b5: spec emoji import feature

I'm requesting a review from @Gargron, who implemented custom emoji and suggested emoji copy feature.

beatrix-bitrot:
:/

I think it's bit large and complex for implementing the emoji copy feature. Can we not just copy the record for a custom emoji but change the domain so it is local? That should allow it to be used by local users and achieve what we need right?

akihikodaki:
I decided to track the origin:
* to cache
* to avoid or to make it easy to resolve potential copyright disputes
* to block

akihikodaki:
I have reviewed the change and I think the complexity is not excessive. Though the number of lines: 1210 addition looks quite a huge, but the large portion is spec (even including specs for existing code to make sure it does not break.) The line number of this pull request is less than 500 lines. When compared to #5231, it is increase of < 200 lines.


So it is whether the three features I mentioned worth 200 lines. I think it is, and probably those for #5239 think so too.

Gargron:
Comments:

- I don't like the CustomEmoji <-> CustomEmojiIcon separation in the database. Too complicated!
- Admin UI changes not needed in my opinion
- The ActivityPub emoji serializer seems to be missing the "id" attribute

If you really insist on a more object-oriented emoji representation, I suggest this:

```js
{
  @context: { ... },
  id: 'https://uri/of/emoji',
  name: 'shortcode here',
  updated: 'so we know if we need to re-cache',
  icon: {
    type: 'Image',
    mediaType: 'image/png',
    url: 'http://url/to/file.png'
  }
}
```

And honestly this would keep changes minimal, I think, because on the processing side we'd only need to change taking the "href" attribute to taking the icon->"url" attribute, and additionally checking "updated"

I can do this today, I think.

akihikodaki:
@Gargron Thank you for your suggestion. I appreciate the simplicity.
However `shortcode` should be changable to avoid collision when copying. That's why I did so.

Apart from that, if you think it is fine and go with the your idea, you can make even simpler by directly putting the URL to image to `icon` property. I hope we can do similar keeping shortcode renaming...

akihikodaki:
Now I have time to reply other comments so here are my answers:
* It is nice if we could avoid CustomEmoji <-> CustomEmojiIcon separation, but as far as I know, that is impossible in nature. Consider the case of hashtags; they are so similar to emojis in the feature and Mastodon actually has two tables for them (`tags` table and `statues_tags` join table)
* Admin UI changes are required because it was implemented before #5231. I can drop the change and move features #5231 misses into other pull requests if any.
* That is because `id` is not necessary to implement the feature.

akihikodaki:
> The ActivityPub emoji serializer seems to be missing the "id" attribute

I also note that it should not have the attribute because it is a Link.

Gargron:
Well, a Link cannot have an icon, so... I just turned into an Object in my other PR.

akihikodaki:
I got it.

akihikodaki:
I wonder whether I can simply put it to `href`. It means there could be different Links which refers to the same object. (`shortcode` differs so the Link is not exactly same, but `href` is same.) Anyway the spec does not prohibit so it may not matter. Also, I have to reconsider ActivityPub representation if #5231 gets merged.

akihikodaki:
I'll close this issue. I'm no longer interested in the feature because the emoji feature landed with 2.0.0 and there is no way to implement but adding a different protocol.

PR #5146, opened 2017-09-29
Append confirmation link as plain text

tribela:
Some mail application is malfunctioning with links.
link as plaintext is useful when this situation:
screenshot from 2017-09-29 20-05-41
(I was not able to click that link, So I used "view original" feature to complete email confirmation)

### DIFF
--- a/app/views/user_mailer/confirmation_instructions.en.html.erb
+++ b/app/views/user_mailer/confirmation_instructions.en.html.erb
@@ -5,6 +5,9 @@
 <p>To confirm your inscription, please click on the following link : <br>
 <%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %>
 
+<p>If above link not worked, Copy and paste this URL</p>
+<span><%= confirmation_url(@resource, confirmation_token: @token) %></span>
+
 <p>Please also check out our <%= link_to 'terms and conditions', terms_url %>.</p>
 
 <p>Sincerely,<p>
--- a/app/views/user_mailer/confirmation_instructions.en.html.erb
+++ b/app/views/user_mailer/confirmation_instructions.en.html.erb
@@ -2,8 +2,8 @@
 
 <p>You just created an account on <%= @instance %>.</p>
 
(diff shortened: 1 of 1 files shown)


Gargron:
There should be a plaintext version of the e-mail for this, the template ends with text.erb instead of html.haml. I don't think we should put plaintext links into html templates.

nightpool:
@Gargron It's very common to put plain-text confirmation links into emails in case they get blocked by the client. Since getting a squeaky-clean email setup is very, very hard, this seems useful.

app/views/user_mailer/confirmation_instructions.en.html.erb
```diff
@@ -5,8 +5,11 @@
 <p>To confirm your inscription, please click on the following link : <br>
 <%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %>
 
+<p>If above link not worked, Copy and paste this URL</p>
```
Gargron: > If the above link did not work, copy and paste this URL into your address bar:

Gargron (changes requested):

Gargron (approved):

Kjwon15:
OMG I changed incorrectly and it merged...
I'm very sorry for this. :cry:

It should be
```html
<p>To confirm your inscription, please click on the following link : <br>
<%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>

<p>If the above link did not work, copy and paste this URL into your address bar: <br>
<span><%= confirmation_url(@resource, confirmation_token: @token) %></span>
```
not
```html
<p>If the above link did not work, copy and paste this URL into your address bar: <br>
<%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>

<p>If above link not worked, Copy and paste this URL</p>
<span><%= confirmation_url(@resource, confirmation_token: @token) %></span>
```

Gargron:
@Kjwon15 Oh, I don't know how I missed it... Tired eyes. Please submit a fix asap...

PR #5150, opened 2017-09-29
Force no cache on web UI page, to force full reload in pinned tabs

Gargron:
e.g. when Firefox is opened and Mastodon is in a pinned tab,
previously the page would be loaded with stale state that didn't
include some updates the user had already seen, and without any
triggers to force the timelines to refresh. Now the page just
gets reloaded to start with

This might fix the same issue in Chrome for Android as well


nolanlawson:
Is this what TweetDeck does, just out of curiosity?

Gargron:
@nolanlawson I thought it worked but further testing revealed that it did not

PR #5152, opened 2017-09-30
Improve IntersectionObserverArticle perf

nolanlawson:
This should improve performance of IntersectionObserverArticle, especially in Chrome. The main idea is to avoid calling `getBoundingClientRect()` unless we are in a `requestIdleCallback`. (Recall that we have to do this for Chrome because of a bug in its IntersectionObserverEntry objects.)

We only need to call gBCR to calculate the height, which is used for unrendered, and this is a low-priority task, so it should be `rIC`-able. This should avoid the situation where we're spending a lot of time in `onIntersection` doing both React rendering and `gBCR`.

I've deployed this to malfunctioning.technology; I'd like to thoroughly test across browsers and hopefully have @sorin-davidoi take a look before merging. :smiley:

Before:

screenshot 2017-09-29 18 39 33

(Notice all the purple rectangles with red corners… those are `gBCR` calls forcing reflows.)

After:

screenshot 2017-09-29 18 39 36

(`gBCR` calls are deferred until later and don't block rendering during scroll.)

### DIFF
--- a/app/javascript/mastodon/components/intersection_observer_article.js
+++ b/app/javascript/mastodon/components/intersection_observer_article.js
@@ -58,26 +58,31 @@ export default class IntersectionObserverArticle extends React.Component {
   }
 
   handleIntersection = (entry) => {
-    const { onHeightChange, saveHeightKey, id } = this.props;
+    this.entry = entry;
 
-    if (this.node && this.node.children.length !== 0) {
-      // save the height of the fully-rendered element
-      this.height = getRectFromEntry(entry).height;
+    scheduleIdleTask(this.calculateHeight);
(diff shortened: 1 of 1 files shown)


abcang (approved):
It is great! Thank you!

nolanlawson:
Tested this on 4 desktop browsers (Chrome/Edge/Firefox/Safari) and didn't see any new bugs. On mobile Chrome on a Nexus 5, scrolling is noticeably smoother. On mobile Firefox, it's about the same.

Currently running a before/after on https://horribly.malfunctioning.technology (before) and https://malfunctioning.technology (after).

PR #5153, opened 2017-09-30
Add not assigned background colors.

matyapiro31:
`.features` and `.extended-description` inherit `background` property from `body`. This lets css `filter()` function work incorrectly.
For example, an extension for colour blind people doesn't work on it.

### DIFF
--- a/app/javascript/styles/about.scss
+++ b/app/javascript/styles/about.scss
@@ -462,7 +462,7 @@
 
   .features {
     padding: 50px 0;
-    backgroud: $ui-base-color;
+    background: $ui-base-color;
 
     .container {
       display: flex;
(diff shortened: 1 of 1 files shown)


app/javascript/styles/about.scss
```diff
@@ -462,6 +462,7 @@
 
   .features {
     padding: 50px 0;
+    backgroud: $ui-base-color;
```
ykzts: typo?

`backgroud:` -> `background:`

matyapiro31:
typo fixed.

ykzts (approved):

lynlynlynx (changes requested):
`.features` and `.extended-description` background colors are NOT `$ui-base-color` .

lynlynlynx:
Are these settings really needed?
Background color in landing pages is set individually for `about-body` class of `body`, not a general background color of  `body`.

Gargron:
I agree with lynx on this. You can just put this into your custom SCSS when you need it.

PR #5156, opened 2017-09-30
Add new statistics to /api/v1/instance

TheKinrar:


nightpool:
I think exposing stats via the API without also exposing them on the /about/more page or somewhere else publicly visible is a bad idea, since it could be seen as misleading the user/admins about what data is publicly available.

failing that, an admin "stats" page with the warning "This data is public and accessible by any API client!" would be okay. I also really just want these stats for myself :D

TheKinrar:
This looks like a great idea, however I probably won't do it as I don't know Ruby in fact, and probably won't be able to do that as it should be done.

akihikodaki (changes requested):
* Just adding counts is not good. They are quite a heavy task and we need reasoning when adding them.
* Newcomers may not know that `first_user_created_at` exists, which may break their expectation.

nightpool:
Maybe this would be better to calculate as a sidekiq scheduled task instead?

PR #5158, opened 2017-09-30
Enable streaming on heroku

tribela:
I am using dokku(Heroku like personal PaaS) but I found mastodon is not running streaming server by default on heroku(https://github.com/tootsuite/mastodon/issues/1119).

This PR will make mastodon is able to stream on heroku by default.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,3 +1,5 @@
 https://github.com/heroku/heroku-buildpack-apt
 https://github.com/Scalingo/nodejs-buildpack
 https://github.com/Scalingo/ruby-buildpack
+https://github.com/ryandotsmith/nginx-buildpack
+https://github.com/danp/heroku-buildpack-runit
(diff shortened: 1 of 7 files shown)


ykzts:
cc @zunda

zunda:
I'm not sure if running the streaming server along with Puma because they scale differently. Personally, I run the streaming server as a separate Heroku app that shares Postgres and Redis from the primary app.

Anyways, I wanted to know how much memory overhead Nginx and runit have. I [tried](https://gist.github.com/zunda/7c1531827509c872d566ff0a21a46018) but the [Heroku Buildpack: NGINX](https://github.com/ryandotsmith/nginx-buildpack) doesn't seem to be compatible with the current stack:

```
$ git push heroku master
  :
remote: -----> nginx-buildpack app detected
remote: cp: cannot stat 'bin/nginx-heroku-16': No such file or directory
remote:  !     Push rejected, failed to compile nginx-buildpack app.
  :
```

akihikodaki:
@Kjwon15 Sorry for late reply, but I wonder if having a sperate app as @zunda suggested works for you. If it is, we can update the documentation to solve the issue you encountered. Otherwise, we can continue reviewing your change.

PR #5159, opened 2017-09-30
Toggle contain:strict on fullscreen

nolanlawson:
Previously we removed `contain: strict` because it interfered with video fullscreen (2bbf987a0a352a36ef0cc7f06fe366b60593e89c). This PR adds it back in, but removes it whenever the video player goes fullscreen.

I also refactored the fullscreen-handling code and moved it to its own module. I removed the `ms-` prefixed calls because Edge uses the WebKit-prefixed ones, and we don't support IE.

### Perf improvements

This is only a perf improvement in Chrome because it's [the only browser to support CSS containment](http://caniuse.com/#feat=css-containment). But you can see the difference:

out

### Browser support

I tested desktop Chrome, Edge, Firefox, and Safari. Everything works fine except for Safari, which is broken for video fullscreen both before and after this commit. So I'll open a separate issue for that.

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -6,6 +6,8 @@ import LoadMore from './load_more';
 import IntersectionObserverWrapper from '../features/ui/util/intersection_observer_wrapper';
 import { throttle } from 'lodash';
 import { List as ImmutableList } from 'immutable';
+import classNames from 'classnames';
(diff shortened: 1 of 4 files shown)


nolanlawson:
Fixed some lint issues

abcang (approved):

nightpool:
this seems to be broken again, or was never fixed for *detailed* statuses (works fine in a timeline)

PR #5161, opened 2017-09-30
Reorganize preferences page

Gargron:
image
image

- "Site theme" should *not* be at the top of preferences
- "Site theme" should not be an option if there is only one theme
- Separated "e-mail digest" from other e-mails, added explanation of what it is
- Segregated options into sections by topic - languages, publishing, notifications, other, web

### DIFF
--- a/app/javascript/styles/forms.scss
+++ b/app/javascript/styles/forms.scss
@@ -22,6 +22,16 @@ code {
     margin-top: 4px;
   }
 
+  h4 {
+    text-transform: uppercase;
+    font-size: 13px;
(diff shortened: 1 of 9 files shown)


ykzts (approved):

JeanGauthier:
According to your screenshot you cannot filter my language, nice of you :P

Gargron:
@JeanGauthier CLD3 can detect about 99 languages, Occitan is not one of them

JeanGauthier:
Weirdly I see it as a good news he he.

PR #5163, opened 2017-10-01
Put display/username on separate lines in inline statuses

Gargron:
image

- Clean up obsolete dropdown CSS
- Use new feature to disable preview completely in emoji-mart
- Improve some focus styles

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -138,8 +138,8 @@ export default class Status extends ImmutablePureComponent {
             {Component => <Component
               preview={video.get('preview_url')}
               src={video.get('url')}
-              width={239}
-              height={110}
+              width={287}
+              height={161}
               sensitive={status.get('sensitive')}
               onOpenVideo={this.handleOpenVideo}
             />}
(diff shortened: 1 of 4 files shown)


mathias-b:
Good idea, actually the username is often unreadable:
screen shot 2017-10-01 at 13 33 51

Technowix:
Would be great to increase infos here then, this empty space made my eyes feel bad.
But would also reduce the size of username :/
image

hrefhref:
That makes the timeline way less readable IMO.

rainbowmultiverse:
How about an expander thing to display the user name? That way you get the compactness of the current layout and an easy way to see user names without wasting a whole line all the time when it's not always useful information. Default expanded or closed could be a user setting.

Aldarone:
I'm really not a fan of it. Username is less important than content and with this loss of vertical space, there will be less visible content.

ProgVal:
How does it play with longer time strings? On my screen, "il y a x minutes" (= the translation for "x hours ago") is larger than the space left by the four buttons

sorin-davidoi:
There could also be five buttons (Chrome >= 61 on Android).

kyzh:
I would find it particulary helpful.
It would make finding people's name easier on mobile.

tcql:
I've been playing with a userstyle to test out some alignment changes, and I have to say I prefer having the content partially indented, otherwise the hierarchy feels really chaotic to my eyes. They way I've got it set on my end, the status content lines up ~ with the left edge of the mini-avatar that shows up if the post is a boost:

screen shot 2017-10-01 at 09 10 18

Of course, action buttons should line up with the content. Still not sold on the status prepend alignment or the name/username/time all stacked, but overall I enjoy the extra width for content, and having left indent makes it feel kind of like the user info is a heading and the content is paragraphs

Gargron:
This proved to be controversial after all, and I do not like the countersuggestions. Overall, keeping the old look is preferable both for familiarity and screen estate. I made another PR instead of this #5171

tinybarks:
A simple solution for desktop would be to just let people be able to hover their mouse over the name to view the username and instance.

Gargron:
@tinybarks The other PR adds a tooltip that does that.

PR #5166, opened 2017-10-01
Nuke redirected accounts from database

ClearlyClaire:
I sometimes have issue federating to other servers.
This is very likely caused by those servers having two accounts in their database matching a same URI: indeed, I had at some point `@thib@social.sitedethib.com` handled with the same URI as my current account (`@thib@sitedethib.com`). This means when resolving my account, those remote servers may find either of those two accounts, one of which has stale data (including public key), and one has not.

One very likely failure caused by this is processing incoming activities from my account: the remote server may resolve the old account, and fail to verify signature since only my new account has up-to-date info.

The proposed change, which is slightly… brutal, is to entirely delete a redirected account. As said, it's quite brutal, and will only fit certain scenarios. This is highly related to #2032 which is still exist with ActivityPub.

### DIFF
--- a/app/services/resolve_remote_account_service.rb
+++ b/app/services/resolve_remote_account_service.rb
@@ -31,7 +31,9 @@ class ResolveRemoteAccountService < BaseService
       @username = confirmed_username
       @domain   = confirmed_domain
     elsif redirected.nil?
-      return call("#{confirmed_username}@#{confirmed_domain}", update_profile, true)
+      redirected_account = call("#{confirmed_username}@#{confirmed_domain}", update_profile, true)
+      @account.destroy unless @account.nil? || redirected_account.nil?
+      return redirected_account
     else
       Rails.logger.debug 'Requested and returned acct URIs do not match'
       return


ThibG:
Why I think it's somewhat ok to do so:
- If an account sets a redirect to another one, that resolves properly, it is intentional and is extremely unlikely to be because of a temporary failure

Why it may not be ok:
- It may not mean that the underlying account has been wiped, only that its canonical name changed. In that case, the account should probably be “migrated” instead (statuses re-attributed, follows changed, etc.)

Why it is not enough:
- `ResolveRemoteAccountService` is actually seldom called at all, which means that account nuking could take ages to actually occur
- It fits my specific case because Mastodon redirects `@user@WEB_DOMAIN` to `@user@LOCAL_DOMAIN`, but other similar cases not covered by this behavior may occur

Other stuff:
- There are a lots of reasons why ActivityPub inbox processing may fail, none of which are logged
- Similarly, all those error cases happen in a sidekiq job, thus none of them are reported to the originating instance

ThibG:
Superseeded by #5177

PR #5170, opened 2017-10-01
Search popout

Gargron:
image


nightpool:
Can we put the "simple text" copy up st the top, since it's the most
common, have it say something like "search for...." and then below it
change "search format" to "Advanced search" or something?

nolanlawson (approved):

PR #5171, opened 2017-10-01
Use own, shorter relative timestamps

Gargron:
Alternative to #5163. The reason relative timestamps were so long was because react-intl does not support a shorter format. However, it's possible to do manually.

- When less than 10 seconds: "now"
- When less than 3 days: "30s", "2m", "1h", "2d"
- Otherwise: "09/24" (localized day-month)

Edit: Also make full username@domain show up in tooltip over the display name in statuses

### DIFF
--- a/app/javascript/mastodon/components/relative_timestamp.js
+++ b/app/javascript/mastodon/components/relative_timestamp.js
@@ -4,10 +4,10 @@ import PropTypes from 'prop-types';
 
 const messages = defineMessages({
   just_now: { id: 'relative_time.just_now', defaultMessage: 'now' },
-  seconds: { id: 'relative_time.seconds', defaultMessage: 's' },
-  minutes: { id: 'relative_time.minutes', defaultMessage: 'm' },
-  hours: { id: 'relative_time.hours', defaultMessage: 'h' },
-  days: { id: 'relative_time.days', defaultMessage: 'd' },
+  seconds: { id: 'relative_time.seconds', defaultMessage: '{number}s' },
+  minutes: { id: 'relative_time.minutes', defaultMessage: '{number}m' },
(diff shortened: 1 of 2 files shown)


nightpool:
can we do both of these, and have 

```
Display name       5m
username

status status status
```

Gargron:
@nightpool No

I think the other PR is too problematic and with this one instead, unjustified

nightpool:
I really like this PR, but I think we should still separate the username:

image


see how it's still nearly impossible to see the domain part?

Gargron:
@nightpool That "external" icon isn't canon. Also, that's why I added a tooltip.

nightpool:
the external icon isn't the major issue here.

beatrix-bitrot:
should we perhaps consult some users of non english/latin based languages before going ahead with this?

in some languages the timestamps are already very compact so it seems a little weird for everyone to end up with these short versions because the timestamp takes up so much space in english

unarist:
In Japanese, "30秒前" (30 seconds ago) can be shorter like "30秒" (30 seconds), and it is what Twitter does. However, in the same rule, "3日" can be read as either of "3 days" and "3rd day of the month". So I'd like to avoid it and back to "3日前", and also for other units for consistency. This problem won't occur on Twitter because they use absolute date always. Anyway, backing to old style can be done in the translation.

Also, I think not a few people want absolute format (even shorter than a day). So absolute date format would be acceptable for those users and it's used in the Twitter. However, this PR displays date in "09/30" format even in Japanese locale. I think "9月30日" format is popular and used in many websites, though I don't know how we can change, and it would be a matter of a preference...

gabrielelucci (approved):

beatrix-bitrot (approved):
s, m, d, and h can be changed in translations so my concern is addressed

PR #5172, opened 2017-10-01
When muting, clear web UI like for blocks

Gargron:
Previously, muting didn't mean removing from notifications. But now that it's a shadow version of a block, it should do the same client-side cleanup as a block.


beatrix-bitrot:
will this also apply to conversation muting and if not, can it? people have been asking for that for quite some time

nightpool:
this is just client-side cleanup

Gargron:
@beatrix-bitrot I guess since I could also fit that in, now that I realize the muting doesn't clear stuff server-side either and it probably should... Just tell me what exactly you expect to "clear" from a conversation mute?

akihikodaki (approved):

PR #5175, opened 2017-10-02
Code-split emoji-mart picker and data

nolanlawson:
This brings back async loading of the emoji picker, by:

1. Copying over the search logic from `emoji-mart`, and
2. Using `@preval` to build a lite version of the emoji data.

Here's the before and after:

out

`common.js` is reduced from 1.35 MB (1347732) to 924 kB (923900). A new chunk is added, `emoji_picker.js`, weighing 623 kB (623198).

**Pros**:

- 424 kB of JavaScript removed from initial page load.
- `emoji_data_light.js` gets updated whenever `emoji_mart` updates it, thanks to `@preval`.

**Cons**:

- If `emoji-mart` changes their search implementation, then the autocomplete search may become functionally different from the picker search.
- More JS overall is loaded if you use the picker (because you end up loading both `emoji_data_light.js` and `data.js` from `emoji-mart`).

Given that people probably only open the picker in a small minority of sessions, I think the tradeoff is worth it.

This is currently running on https://horribly.malfunctioning.technology for those who want to test it.

### DIFF
--- a/app/javascript/mastodon/emoji_data_light.js
+++ b/app/javascript/mastodon/emoji_data_light.js
@@ -1,0 +1,17 @@
+// @preval
+const data = require('emoji-mart/dist/data').default;
+const pick = require('lodash/pick');
+
+const condensedEmojis = {};
+Object.keys(data.emojis).forEach(key => {
+  condensedEmojis[key] = pick(data.emojis[key], ['short_names', 'unified', 'search']);
+});
+
+// JSON.parse/stringify is to emulate what @preval is doing and avoid any
+// inconsistent behavior in dev mode
+module.exports = JSON.parse(JSON.stringify({
+  emojis: condensedEmojis,
+  skins: data.skins,
+  categories: data.categories,
+  short_names: data.short_names,
+}));
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -1,6 +1,6 @@
 import api from '../api';
-import { emojiIndex } from 'emoji-mart';
 import { throttle } from 'lodash';
+import { search as emojiSearch } from '../emoji_index_light';
 
 import {
   updateTimeline,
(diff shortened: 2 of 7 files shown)


ykzts (approved):

app/javascript/mastodon/emoji_data_light.js
```diff
@@ -0,0 +1,17 @@
...
+Object.keys(data.emojis).forEach(key => {
+  condensedEmojis[key] = pick(data.emojis[key], ['short_names', 'unified', 'search']);
+});
+
+// JSON.parse/stringify is to emulate what @preval is doing and avoid any
+// inconsistent behavior in dev mode
+module.exports = JSON.parse(JSON.stringify({
+  emojis: condensedEmojis,
+  skins: data.skins,
+  categories: data.categories,
```
sorin-davidoi: Since this is only used in `emoji_index_light`, maybe we should use one-letter keys to save space?

sorin-davidoi: In `emojione_light` we used arrays to get rid of the keys altogether.

nolanlawson: Yeah we could maybe trim this even further. Using arrays instead of objects may be a good start. I'm interested though in the tradeoff between transfer cost and decompression cost – e.g. with arrays instead of objects, how much more expensive is it to decompress on the client side? Does it offset the gains in transfer/parse costs?

The search code could also be trimmed, because there are lots of emoji-mart options we're not using. I also believe there may be repeated information between `emojione_light.js` and `emoji_data_light.js`.

sorin-davidoi: > I'm interested though in the tradeoff between transfer cost and decompression cost – e.g. with arrays instead of objects, how much more expensive is it to decompress on the client side? Does it offset the gains in transfer/parse costs?

I'm curious, how would you measure that? On the surface, I would expect decompression to be faster since the payload is smaller and arrays are a somehow simpler data structure than hash-maps.

sorin-davidoi: | I also believe there may be repeated information between emojione_light.js and `emoji_data_light.js`.

How come `emojione_light` is still needed/used?

nolanlawson: Good question. I think it may have something to do with how the emoji codepoints are expressed in Twemoji? @Gargron?

PR #5177, opened 2017-10-02
[WiP] Fix uri handling (#5167)

### LINKED ISSUE #5167: acct: URI VS author URI
# Author URI

Both OStatus and ActivityPub uniquely identify users/authors using author URIs, which need to be publicly dereferencable and are encouraged to be HTTP(S) URLs. In Mastodon, that URI is the same for ActivityPub and OStatus: `https://<WEB_DOMAIN>/users/<Username>`.

That means, when two instances interact, an author (including in mentions) is identified solely by the author URI.

Every other OStatus or ActivityPub implementation I know use HTTP(S) URLs as author URIs.

In Mastodon's model, author URI uniqueness is *not* enforced (multiple `Account` objects can share the same `uri` field).

# acct: URI

While from ActivityPub or OStatus' point of view, authors are identified by an arbitrary URI, Mastodon identifies authors using an [acct: URI](https://tools.ietf.org/html/rfc7565), that is, a pair of username and domain that can be resolved using webfinger.

In Mastodon's model, the username + domain pair uniqueness *is* enforced.

# acct: URI VS Author URI

This means that when looking up a remote user, Mastodon will use the username + domain pair, which will map to an author URI. The converse is not true, however: there is no mechanism to go from an author URI to an acct: URI, even less so uniquely!

So, in general, there is a mapping from acct: URIs to author URIs, but not the other way around.

# Why the current implementation is flawed

Since the protocol only uses the author URI and not the acct: URI, Mastodon must find account information based on the author URI, which is not well-defined when there are multiple `Account` items sharing the same `uri` field.
There are multiple situations where Mastodon's current implementation causes different kind of issues:

- Multiple acct: URIs for a single author: it is somewhat common in the GNUStatus world to have different acct: URIs aliases with no canonical one for a single author URI. In this case, all the accounts will share the same profile info, including keypairs and endpoint URLs, but statuses, followers, likes etc. may be split between the two identical accounts.
- In my case, I migrated from a `user@domain` acct: URI to a `user@otherdomain` acct: URI while keeping the same author URI. The old account is dead and redirects to the new one, meaning that the profile info (including the keypair) is stale on every instance that knew it, which sometimes cause huge federation issues (when the stale account is selected).
- This last case could also be triggered intentionally by an adversary wanting to hamper federation by creating a dummy OStatus user with the victim's author URI, from a Webfinger provider they control.

# Proposed solutions

Having thought a bit about that issue, I think there are two possible ways forward:
- Ensure a mapping from author URI to acct: URI
- Accept that a single author might have multiple acct: URI

The first one is what I proposed in #2032 months ago, but I do not think it's the best way to do it, as it reduces compatibility by making hard requirements on top of OStatus or ActivityPub, that contradict with some existing usages (e.g., the GNUStatus situation described above).

The other solution involves splitting the username/domain pair from the `Account` model, and putting it in a new `Alias` table instead. Apart from the refactoring, this would also require thinking about how to present a “canonical” acct: URI (or alternatively presenting the list of known acct: URIs) to the end user.
This would not affect discovery.

ClearlyClaire:
This is a work-in-progress attempt to fix #5167.
Basically, it forces every author URI to map back to an “acct:” URI by the same rules used to discover an “acct:” URI for unknown users, both in OStatus and ActivityPub, then proceed to merge/delete duplicate accounts.

What is still missing:
- Do not temporarily create URI conflicts if none exist (the current patched code will create a new account with shared URI then proceed to delete the old one)
- ~~~Re-follow on account migration~~~
- ~~~Avoid (up to two per account update) needless webfinger calls when discovering ActivityPub authors~~~
- Provide a task to deduplicate every problematic set of users in the database
- Enforce author URI uniqueness at model/database level
- Make sure there aren't any race condition
- RSpec: ~~~remote account merging and~~~ re-following
- Actually test it

### DIFF
--- a/app/services/account_deduplication_service.rb
+++ b/app/services/account_deduplication_service.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+class AccountDeduplicationService < BaseService
+  def call(reference_account)
+    Account.where.not(id: reference_account.id).where(uri: reference_account.uri).each do |account|
+      if account.protocol == reference_account.protocol && account.public_key == reference_account.public_key
+        # The accounts are functionnaly the same, merge them
+        merge_accounts(reference_account, account)
+      end
+      account.destroy
+    end
+  end
+
+  private
+
+  def merge_accounts(reference_account, account)
+    ActiveRecord.transaction do
+      # Re-attribute statuses
+      account.statuses.find_each do |status|
+        status.account = reference_account
+        status.save
+      end
+
+      # Re-attribute favourites
+      account.favourites.find_each do |fav|
+        fav.account = reference_account
+        fav.save
+      end
+
+      # Re-attribute mentions
+      account.mentions.find_each do |mention|
+        mention.account = reference_account
+        mention.save
+      end
+
+      # Re-follow reference account
+      account.followers.where(domain: nil).find_each do |follower|
+        # Schedule re-follow
+        begin
+          FollowService.new.call(follower, reference_account)
+        rescue Mastodon::NotPermittedError, ActiveRecord::RecordNotFound, Mastodon::UnexpectedResponseError, HTTP::Error, OpenSSL::SSL::SSLError
+          next
+        end
+      end
+    end
+  end
+end
--- a/app/services/activitypub/fetch_remote_account_service.rb
+++ b/app/services/activitypub/fetch_remote_account_service.rb
@@ -5,7 +5,9 @@ class ActivityPub::FetchRemoteAccountService < BaseService
 
   # Should be called when uri has already been checked for locality
   # Does a WebFinger roundtrip on each call
-  def call(uri, id: true, prefetched_body: nil)
+  # If the uri was retrieved from a WebFinger query at "acct:#{webfinger_username}@#{webfinger_domain}",
+  # some WebFinger roundtrips might be avoided
+  def call(uri, id: true, prefetched_body: nil, webfinger_username: '', webfinger_domain: '')
     @json = if prefetched_body.nil?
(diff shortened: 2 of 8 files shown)


ThibG:
I'm not too sure about possible race conditions, but it should be otherwise fine. As illustrated by the RSpec, I think it handles gracefully every sensible case.

However, I don't think it can ensure `uri` uniqueness at all times: there might be weird edge cases where a new `Account` is temporarily created with an already-existing `uri` value. One (the only one?) such case could be the following: `acct:a@domain` serves `uri1`, `acct:b@domain` used to serve `uri2`, `acct:b@domain` changes to also serve `uri1` and becomes the canonical `acct:` URI for `uri1`. In this case, my code will select the account matching `b@domain`, change its data (including uri), and then merge `a@domain` into it, meaning there is temporarily two `Account` objects with the same `uri` value.

Gargron:
I think the "merge duplicate accounts" functionality is something we need, together with a fix for the accounts username/domain index (remove case-sensitive unique index, convert case-insensitive index to unique index)

Not sure about the other changes, I need a closer look, but it's all a bit outdated and merge-conflicty now. Protip: I may not look at PRs that have "WIP" in the title, until they don't have "WIP" in the title anymore... So here we are 5 months later

ThibG:
Well, this was one of many attempts to start a discussion about this messy duplicate/redundant IDs stuff going on, and the PR *really* needed some discussion before merging.

I'll finish the things I'm working on right now, and I guess I'll try to make sense of this issue again, now that you seem available to discuss it.

Issue #5050, opened 2017-09-22
Make og:image on Toot detail page optional


cooox:
When linking a Toot from another service, e.g. Twitter, which crawls the Opengraph meta tags, the user's avatar is displayed as the image. This might be disturbing, if one only wants to display a Card with the title and description.

Is there a way to toggle the og:image inclusion? Maybe a setting like the existing `setting_noindex`, e.g. `setting_disable_cardavatar`, could be introduced and handled like the option in 447b8bc44eb33ae6ffdfbcc7ae52fbe0bba93ece?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Maybe we should disable avatars-as-og:image completely? That being said, that is what Twitter does for tweets, too.

cooox:
I checked multiple solutions, but the fact that og:image is not optional according to http://ogp.me more or less resolves this issue. 

Besides, currently the Twitter card type `summary_large_image` is used, maybe `summary` is a better solution. https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary

That being said, Twitter asks to use the meta tag with `name` instead of `property`, https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started, so 
`= opengraph 'twitter:card', 'summary'` should really be `%meta{ name: 'twitter:card', content: 'summary' }` in the view: https://github.com/tootsuite/mastodon/blob/cdad7977fc94cd6a1a97841ed0f25e8504cb80d6/app/views/stream_entries/show.html.haml#L17

One thing that looked a lot more promising was using the oops.png-image in https://github.com/tootsuite/mastodon/blob/cdad7977fc94cd6a1a97841ed0f25e8504cb80d6/app/views/stream_entries/_og_image.html.haml#L21 meaning `= opengraph 'og:image', full_asset_url('/oops.png')`. 

Works better with the higher resolution - maybe there's a good Card-image that can be used with Mastodon?

nightpool:
I don't see any difference between summary card and summary_large_image
card on the page you linked?

cooox:
[...]

PR #5197, opened 2017-10-03
Switch to "summary" Twitter Card type (#5050)

cooox:
Switches from "summary_large_image" to "summary" Card type for Twitter in the post meta tags. See the discussion and comparison in issue #5050.
Also replaces opengraph with meta, since Twitter asks for name instead of property.

### DIFF
--- a/app/views/stream_entries/show.html.haml
+++ b/app/views/stream_entries/show.html.haml
@@ -6,6 +6,8 @@
   %link{ rel: 'alternate', type: 'application/json+oembed', href: api_oembed_url(url: account_stream_entry_url(@account, @stream_entry), format: 'json') }/
   %link{ rel: 'alternate', type: 'application/activity+json', href: ActivityPub::TagManager.instance.uri_for(@stream_entry.activity) }/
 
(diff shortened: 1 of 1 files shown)


cooox:
The test `spec/controllers/api/v1/accounts/relationships_controller_spec.rb:53` seems to have random hiccups. 
     
    1) Api::V1::Accounts::RelationshipsController GET #index provided multiple IDs returns JSON with  correct data                                                                                                                       
     Failure/Error: expect(json.first[:id]).to eq simon.id          
                                                                                  
       expected: 15258                                                       
            got: 15259                                                   
                                                                     
       (compared using ==)                                                   
     # ./spec/controllers/api/v1/accounts/relationships_controller_spec.rb:53:in `block (4 levels) in <top (required)>'```

Two users are returned in the JSON, but the first user in the JSON does not always have to be the one that was first specified in the test (e.g. `json.first` does not have to be `simon`, as `simon` could also be the second entry).

Gargron (changes requested):
In my opinion:

- summary_large_image when there are media attachments
- summary otherwise

I don't think name vs property matters since currently the card opt-in already works so I would avoid that change.

PR #5199, opened 2017-10-03
Add spec for emoji_index_light.js

nolanlawson:
This adds tests for the changes made in https://github.com/tootsuite/mastodon/pull/5175. The goal is to ensure that our "light" `emojiIndex.search()` implementation doesn't deviate too much from `emoji-mart`'s version.

### DIFF
--- a/spec/javascript/components/emoji_index.test.js
+++ b/spec/javascript/components/emoji_index.test.js
@@ -1,0 +1,81 @@
+import { expect } from 'chai';
+import { search } from '../../../app/javascript/mastodon/emoji_index_light';
+import { emojiIndex } from 'emoji-mart';
+import { pick } from 'lodash';
+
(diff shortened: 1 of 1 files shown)


spec/javascript/components/emoji_index.test.js
```diff
@@ -0,0 +1,81 @@
...
+import { pick } from 'lodash';
+
+const trimEmojis = emoji => pick(emoji, ['id', 'unified', 'native', 'custom']);
+
+// hack to fix https://github.com/chaijs/type-detect/issues/98
+// see: https://github.com/chaijs/type-detect/issues/98#issuecomment-325010785
+import jsdom from 'jsdom';
+global.window = new jsdom.JSDOM().window;
+global.document = window.document;
+global.HTMLElement = window.HTMLElement;
```
nolanlawson: This is required when we use `.deep.equal()` because of a bug in Chai.

Gargron (approved):

PR #5201, opened 2017-10-03
Compress emoji_data_light.js

nolanlawson:
This compresses `emoji_data_light` by converting objects to arrays, as suggested by @sorin-davidoi  (https://github.com/tootsuite/mastodon/pull/5175#discussion_r142101167). The size of `common.js` is reduced from 925 kB (925004) to 868 kB (867839), a savings of 57.2 kB.

I was worried there may be a tradeoff with the cost of decompression, so I [tested with perf marks](https://github.com/nolanlawson/mastodon/commit/ff8289150405ed886cc9f20f2ddea9f449c86fef) and measured the decompress time as 15ms on a Nexus 5 running Chrome and 4.5ms on the same device running Firefox. So the decompression costs are negligible.

Before and after:

out

It's very likely we could compress this further (e.g. by combining `emoji_data_compressed.js` and `emojione_light.js`, but this is a good step in that direction.

### DIFF
--- a/app/javascript/mastodon/emoji_data_compressed.js
+++ b/app/javascript/mastodon/emoji_data_compressed.js
@@ -1,0 +1,22 @@
+// @preval
+const data = require('emoji-mart/dist/data').default;
+const pick = require('lodash/pick');
+const values = require('lodash/values');
+
+const condensedEmojis = Object.keys(data.emojis).map(key => {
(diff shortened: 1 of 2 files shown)


nolanlawson:
This is live on https://malfunctioning.technology right now

sorin-davidoi (approved):
Awesome! 🚀

PR #5207, opened 2017-10-03
Use separate workers to process imports, retry failures

Gargron:


beatrix-bitrot (approved):
seems much better, nice job on the negative loc change too

PR #5211, opened 2017-10-04
WIP: Set backdated timestamp IDs on new old Statuses

aschmitz:
For any "old" status (where the creation time is before the update time, as both are already set by Rails by the time we get to around_save), we now generate an ID for the "old" (or potentially, future-dated) status, and attempt to use that to insert.

In the vast majority of cases, this should succeed. If it doesn't, we retry up to a thousand times, each time incrementing the ID by a random number between 0 and 99, to avoid too many collisions.

Note that this also necessitates setting `created_at` to nil (or leaving it unset), not Time.utc.now, when creating new statuses. (Although the only downside to setting a time is a small amount of extra work, nothing catastrophic.)

### DIFF
--- a/lib/mastodon/timestamp_ids.rb
+++ b/lib/mastodon/timestamp_ids.rb
@@ -2,6 +2,46 @@
 
 module Mastodon
   module TimestampIds
+    class Callbacks
+      def self.around_create(obj)
+        if obj.created_at == obj.updated_at
+          # If this is just a new status, pass it along.
+          yield
+        else
+          # Otherwise, we want to backdate this ID to have it sort
+          # nicely by ID.
+          obj.id = Mastodon::TimestampIds.id_at(obj.created_at)
+          tries = 0
+          begin
+            # This will almost never be necessary, but if for some
+            # reason we have a glut of statuses all inserted at the
+            # same time, we may need to increment the ID until we find
+            # one that's free. We increment by a random amount to avoid
+            # repeatedly querying the same set of IDs in such a case.
+            # (This mostly exists to avoid a DoS attack, and should
+            # rarely be useful under normal circumstances.)
+            obj.id = rand(100) if tries.positive?
+            yield
+          rescue ActiveRecord::RecordNotUnique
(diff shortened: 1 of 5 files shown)


aschmitz:
This is based on #4801, and appears to work in some cursory testing, but is marked a WIP because it needs tests.

Gargron:
Time to rebase! I usually do a branch switcharoo with cherry-pick on the latest commit to make things right.

clworld:
Things I afraid about toot backfillings.
* Faked publish time (maybe future): If toot with faked publish time (1 year later for example) arrived, that toot stay top of Home & Fedeteted TL (because it will get largest id).
* Delayed deliver: If remote instance's sidekiq push queue is slow enough, User will get remote instance's toot at N-hour before position (especially on REST API access.)
* Breaks differential fetch by REST API client: REST API access with since_id = last fetched max status_id will not get newly arrived toots. It's breaking change.

I think normal arriving remote status should be numbered by arriving time and reject futute publish time for backfillings(boosted original status, searched status, importing, etc).

Gargron:
Solutions:

- Add a flag/option to the processing classes; if true, use local timestamp instead of created_at. Set this flag to true in codepaths that come from ActivityPub::InboxController only (real-time deliveries)
- If a remote timestamp is in the future, override it with local timestamp

lib/mastodon/timestamp_ids.rb
```diff
@@ -2,6 +2,46 @@
...
+          tries = 0
+          begin
+            # This will almost never be necessary, but if for some
+            # reason we have a glut of statuses all inserted at the
+            # same time, we may need to increment the ID until we find
+            # one that's free. We increment by a random amount to avoid
+            # repeatedly querying the same set of IDs in such a case.
+            # (This mostly exists to avoid a DoS attack, and should
+            # rarely be useful under normal circumstances.)
+            obj.id = rand(100) if tries.positive?
```
nightpool: should be `+=`?

aschmitz:
Closing this in favor of #5260.

PR #5213, opened 2017-10-04
Check Webfinger-returned author URI even when not redirected

ClearlyClaire:
The whole point of verified_webfinger? is to check the WebFinger-discoverable
URI maps back to the known author URI. This was not actually verified if the
first Webfinger request was not a redirection.

I can't think of a scenario where that could be abused by an attacker, but it's worth fixing nonetheless.

### DIFF
--- a/app/services/activitypub/fetch_remote_account_service.rb
+++ b/app/services/activitypub/fetch_remote_account_service.rb
@@ -31,7 +31,7 @@ class ActivityPub::FetchRemoteAccountService < BaseService
     webfinger                            = Goldfinger.finger("acct:#{@username}@#{@domain}")
     confirmed_username, confirmed_domain = split_acct(webfinger.subject)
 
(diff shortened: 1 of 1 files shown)


Gargron:
>This was not actually verified

Well, it was, because that correct WebFinger response was from the same host as the URI, and it is *highly* unlikely (I would consider it impossible) that WebFinger is not under the control of the same entity as the profile.

Gargron (approved):

ThibG:
@Gargron yup, *highly* unlikely, close to impossible, that's why I said I can't see how an attacker could take advantage of that. Still, I guess this kind of mismatch could have happened with buggy implementations or messed up deployments. Also would happen with the legitimate but very silly setup (`acct:user@a` having author URI `https://b/users/user`, and `acct:user@b` having author URI `https://a/users/user`) I plan to do someday to catch all these kinds of edge cases.

PR #5214, opened 2017-10-04
Better Japanese translation

nzws:
"障がい者" is a soft expression.

### DIFF
--- a/app/javascript/mastodon/locales/ja.json
+++ b/app/javascript/mastodon/locales/ja.json
@@ -201,7 +201,7 @@
   "tabs_bar.notifications": "通知",
   "upload_area.title": "ドラッグ＆ドロップでアップロード",
   "upload_button.label": "メディアを追加",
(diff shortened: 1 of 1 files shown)


ykzts:
"障害" means "障礙". "礙" has the meaning of "hindering".

Since "礙" is not included in [list of Joyo kanji](http://www.bunka.go.jp/kokugo_nihongo/sisaku/joho/joho/kijun/naikaku/kanji/), in modern times "害" is substituted for it.

"害" generally means "hurt" or "disaster". However, in "障害" it has no special meaning and exists only for sound.

I don't think this change is meaningful.

zonuexe:
@yuzulabo 
Some screen reader implementations do not properly pronounce `障がい者`.
This change is not only nonsense, it may impair accessibility.  Please do not apply also to UI of other applications.

yuzulabo:
understood.
Thank you for telling me.

PR #5217, opened 2017-10-04
Fix regression in FetchRemoteResourceService

ClearlyClaire:
`FetchAtomService`'s interface changed in #5114 but `FetchRemoteResourceService` has not been updated for this change.

I only went through #5114 quickly so it is possible that this fix breaks what was intended by #5114.


akihikodaki (approved):

Gargron (changes requested):
(The tests fail properly)

ThibG:
I'm under the impression that those tests don't reflect the changes that were made in #5114, hm…

ThibG:
Changed the stubs in the test to match changes made in #5114, tests pass again

Gargron (approved):

PR #5224, opened 2017-10-04
Add atom renderer to hashtags

ClearlyClaire:
Render hashtags as an atom feed when accessing `https://WEB_DOMAIN/tags/hashtag.atom`, thus allowing people to subscribe to given hashtags.

I'm not too happy with the code duplication, but statuses do not necessarily have a `StreamingEntry` associated.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -13,6 +13,12 @@ class TagsController < ApplicationController
         @initial_state_json   = serializable_resource.to_json
       end
 
+      format.atom do
+        @statuses = Status.as_tag_timeline(@tag, current_account, params[:local]).paginate_by_max_id(20, params[:max_id])
+        @statuses = cache_collection(@statuses, Status)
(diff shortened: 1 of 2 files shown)


akihikodaki:
Couldn't you make an Atom bridge using an existing API?

Gargron:
Hmm.. I am not looking to expand the Atom endpoints anymore... At least not for the purpose of federation - we should focus on ActivityPub-only going forward. StreamEntry is a legacy model and hopefully we can get rid of it sometime.

ThibG:
It's not about federation, but rather giving off-the-shelf atom feed readers access to tags. That seemed like low-hanging fruit to me, but I understand not going this way if you want to get rid of the atom code.

Gargron:
Here is my objection: Our original Atom feeds are for federation, not end-user consumption. They are valid feeds, but the problem is, end-user feed readers are often very confused by all the extra tags and namespaces. Feed validators go nuts too. There's a lot of issues written about how IFTTT does something wrong with the Atom feeds. So if your target is end-users with feed readers, I think maybe you'll just want to add RSS specifically, instead? With the explicit goal of making content readable on a feed reader, rather than federation.

I'm going to close this.

PR #5225, opened 2017-10-05
Log delays in delivery of toots

zunda:
After Japan national day of load testing, aka televising of "Castle in the Sky" on which many internet users post "balse" to SNSs, I noticed Mastodon doesn't have metrics that indicate how it is communicating to the fedeverse.

This change is to add logs like below that will show how quickly Mastodon is delivering toots to remote instances as a good measure of how quickly Mastodon is working on queued tasks.

```
source=Pubsubhubbub::DeliveryWorker destination="https://example.com/api/subscriptions/12345" measure#delivery.delay=65sec count#delivered=1
```

```
source=ActivityPub::DeliveryWorker destination="https://example.com/inbox" measure#delivery.delay=66sec count#delivered=1
```

The additional metrics can be visualized with, e.g., Librato like below.

screen shot 2017-10-04 at 9 56 30 pm


Gargron:
Have you tried hooking up Mastodon to statsd? When configured, Mastodon sends render times, query times, cache hits, and sidekiq processing times to statsd, which can output them into a time series database. Perhaps you could make this be a statsd invocation too, instead of a log entry?

zunda:
Ok, I'll see how it runs on Heroku.

PR #5229, opened 2017-10-05
Compress and combine emoji data

nolanlawson:
This further compresses the emoji data, about as far as it can go (I believe). The two previous modules (`emoji_data_compressed.js` and `emojione_light.js`) are combined into a single module (`emoji_compressed.js`).

Before and after:

out

(Live versions: [before](https://bl.ocks.org/nolanlawson/raw/015cee018347bcbaa69c7d24b5969d0f/before.html) / [after](https://bl.ocks.org/nolanlawson/raw/015cee018347bcbaa69c7d24b5969d0f/after.html), and [here's what the JSON file looks like](https://bl.ocks.org/nolanlawson/raw/015cee018347bcbaa69c7d24b5969d0f/emoji_compressed.json))

The main insight here is that we can use arrays for everything to save space, and that the SVG filenames and "unified" names can be computed client-side. Unfortunately, as @Gargron discovered, I couldn't find a consistent rule for computing these. So my strategy was to write functions that could cover the majority of cases, and then for the edge cases, we just ship those in the JSON file.

For calculating `filename` there are 754 out of 3374 that I couldn't compute, and for `unified` there are 145 out of 1484 that I couldn't compute. I'm not sure if `unified` has to be kept exactly the same as it is in `emoji-mart` (sometimes it's inconsistent between lowercase and uppercase) but I kept it the same in case it matters.

**Pros**:

* Size of `common.js` reduced from 870kB to 794kB (76.3kB saved)
* Emojis moved into separate `features/emoji` directory for code cleanliness
* More comments to explain what's going on in the compression code
* Fixed a case of emojis without shortcodes showing `title=":null:"`

**Cons**:

* Gzipped size of `common.js` only reduced from 228kB to 216kB
* Code is arguably more complicated due to compression/uncompression
* There is a runtime overhead to the decompression – I calculated 95.5ms on Chrome on Nexus 5 and 77.4ms on Firefox on Nexus 5

**Future work**:

We can probably reduce the decompression runtime cost by running computations just-in-time. For instance, instead of a `unicodeMapping`, we could just have a function that takes a unicode string and returns the filename and title, on demand. This may also improve the memory cost.

However, I think we are reaching diminishing returns on shrinking the emoji size. Gzip takes care of most repeated code, and decompression costs will only increase the more we try to shrink the JSON file.


app/javascript/mastodon/emoji_data_compressed.js
```diff
@@ -1,22 +0,0 @@
-// @preval
```
nolanlawson: This file combined into `emoji_compressed.js`

app/javascript/mastodon/emoji_data_light.js
```diff
@@ -1,16 +0,0 @@
-const data = require('./emoji_data_compressed');
```
nolanlawson: this file combined into `emoji_compressed.js`

app/javascript/mastodon/emojione_light.js
```diff
@@ -1,38 +0,0 @@
-// @preval
```
nolanlawson: this file combined into `emoji_compressed.js`

app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
```diff
@@ -6,7 +6,7 @@ import { Overlay } from 'react-overlays';
 import classNames from 'classnames';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import detectPassiveEvents from 'detect-passive-events';
-import { buildCustomEmojis } from '../../../emoji';
+import { buildCustomEmojis } from '../../emoji/emoji';
```
nolanlawson: maybe we should rename this to `index.js`, not really sure

app/javascript/mastodon/features/emoji/emoji.js
```diff
@@ -35,8 +35,9 @@ const emojify = (str, customEmojis = {}) => {
       if (!rend) break;
       i = rend;
     } else { // matched to unicode emoji
-      const [filename, shortCode] = unicodeMapping[match];
-      replacement = `<img draggable="false" class="emojione" alt="${match}" title=":${shortCode}:" src="${assetHost}/emoji/${filename}.svg" />`;
+      const { filename, shortCode } = unicodeMapping[match];
+      const title = shortCode ? `:${shortCode}:` : '';
+      replacement = `<img draggable="false" class="emojione" alt="${match}" title="${title}" src="${assetHost}/emoji/${filename}.svg" />`;
```
nolanlawson: some emojis don't have a shortcode, this was showing as `null` before

app/javascript/mastodon/features/emoji/unicode_to_filename.js
```diff
@@ -0,0 +1,26 @@
+// taken from:
```
nolanlawson: I took this from twemoji but it doesn't seem to work 100% of the time. I really don't understand how twemoji builds the SVG files

app/javascript/mastodon/features/emoji/unicode_to_unified_name.js
```diff
@@ -0,0 +1,17 @@
+function padLeft(str, num) {
```
nolanlawson: This is my hunch as to how emoji-mart computes the `unified` name, but again it doesn't work 100% of the time due to inconsistencies between lowercase and uppercase… not sure why.

nolanlawson:
This is live on https://malfunctioning.technology, and I also added some new tests to try to ensure this doesn't break any emoji behavior.

nolanlawson:
Fixed conflict

Gargron (approved):
Well, if you say there are no regressions...

nolanlawson:
The tests say there are no regressions. :wink:

PR #5230, opened 2017-10-05
When processing custom emoji, ensure a non-animated version exists

Gargron:
Use the non-animated version in web UI, but return both in API

When I picked the PNG format for custom emojis, my intention was to not allow animations, otherwise I would've allowed GIF. We respect people who want to disable GIF autoplay and use hover-to-play instead, but this is not possible in this case.


beatrix-bitrot (approved):
appreciate it

would like a way for users to choose to see animated ones in the web UI but this is good regardless

Technowix:
So ehhh, no option for thoses who want to keep animated one?

Technowix:
We should be able to enable / disable animation, we where speaking of a "interface wide" option for this, where did it go?

PR #5238, opened 2017-10-06
Fix thinking_face emoji autocomplete

nolanlawson:
As discovered by @unarist, there's a bug in `emoji-mart`'s search where no results are returned when you type `:thinking_fac`:

> `search()` function split it by `-` `_` and treat as AND query. that `intersect()` function calculates intersection of all search result, by iterating items in one Set and check existence in other Set.
to use `Array.prototype.filter()`, it converts Set to Array using spread operator, but Babel compiles `[...x]` to `[].concat(x)`, and it results array which has `x` as an item.

We can fix this by avoiding `[...x]` in the `intersect()` function, which this PR accomplishes. I'll also open a PR on emoji-mart so that it can get fixed over there too.

### DIFF
--- a/app/javascript/mastodon/features/emoji/emoji_utils.js
+++ b/app/javascript/mastodon/features/emoji/emoji_utils.js
@@ -125,13 +125,16 @@ function getData(emoji) {
 }
 
 function intersect(a, b) {
-  let aSet = new Set(a);
-  let bSet = new Set(b);
-  let intersection = new Set(
-    [...aSet].filter(x => bSet.has(x))
-  );
-
-  return Array.from(intersection);
+  let set;
+  let list;
+  if (a.length < b.length) {
+    set = new Set(a);
+    list = b;
+  } else {
+    set = new Set(b);
+    list = a;
+  }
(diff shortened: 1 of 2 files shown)


unarist (approved):

nolanlawson:
Hm, looks like emoji-mart may have already fixed this in the latest release? The [intersect() implementation](https://github.com/missive/emoji-mart/blob/f8e824e1c5b3117b93141caedc84957e9c6f063d/src/utils/index.js#L136-L141) is changed.

unarist:
Yes, I've thought too before...but...oh.

They now [passes array of result array to uniq() directly](https://github.com/missive/emoji-mart/blob/f8e824e1c5b3117b93141caedc84957e9c6f063d/src/utils/emoji-index.js#L132-L134), not intersect() 😕

nolanlawson:
Related: https://github.com/missive/emoji-mart/issues/131

BTW this also applies to emoji with `-` in them. I think we should fix this in emoji-mart first and then port the changes over to our own code, since emoji-mart has already heavily changed their implementation.

PR #5243, opened 2017-10-06
Encode custom emojis as resolveable objects in ActivityPub

Gargron:
Within master this is a breaking change. Instead of encoding custom emojis as Link entities with href pointing to the static file, they are now encoded as Object entities with an icon property that points to the static file:

```js
{
  @context: { ... },
  id: 'https://uri/of/emoji',
  name: 'shortcode here',
  updated: 'so we know if we need to re-cache',
  icon: {
    type: 'Image',
    mediaType: 'image/png',
    url: 'http://url/to/file.png'
  }
}
```

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -86,6 +86,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def process_emoji(tag, _status)
+    return if skip_download?
     return if tag['name'].blank? || tag['icon'].blank? || tag['icon']['url'].blank?
 
     shortcode = tag['name'].delete(':')
@@ -94,7 +95,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     updated   = tag['updated']
     emoji     = CustomEmoji.find_by(shortcode: shortcode, domain: @account.domain)
 
-    return if skip_download? || (!emoji.nil? && emoji.updated_at >= updated && emoji.image_remote_url == image_url)
+    return unless emoji.nil? || emoji.updated_at >= updated
 
     emoji ||= CustomEmoji.new(domain: @account.domain, shortcode: shortcode, uri: uri)
     emoji.image_remote_url = image_url
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -86,15 +86,18 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def process_emoji(tag, _status)
-    return if tag['name'].blank? || tag['href'].blank?
+    return if tag['name'].blank? || tag['icon'].blank? || tag['icon']['url'].blank?
 
     shortcode = tag['name'].delete(':')
+    image_url = tag['icon']['url']
+    uri       = tag['id']
+    updated   = tag['updated']
     emoji     = CustomEmoji.find_by(shortcode: shortcode, domain: @account.domain)
 
-    return if !emoji.nil? || skip_download?
+    return if skip_download? || (!emoji.nil? && emoji.updated_at >= updated && emoji.image_remote_url == image_url)
 
-    emoji = CustomEmoji.new(domain: @account.domain, shortcode: shortcode)
-    emoji.image_remote_url = tag['href']
+    emoji ||= CustomEmoji.new(domain: @account.domain, shortcode: shortcode, uri: uri)
+    emoji.image_remote_url = image_url
     emoji.save
   end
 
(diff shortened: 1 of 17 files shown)


akihikodaki:
Sorry I was replying late, but it is not what was intended in #5145.
The form of the representation does not matter at all. The intension is:
* to keep the sole, unique identifier to the _Image_ object and
* to allow to edit shortcode to avoid collision when copying for scalability

I should have clarified that also in the first comment of the pull request. My apologies.

Gargron:
@akihikodaki If you want to copy an emoji but the shortcode is taken, you can just save the image and upload it like a new emoji. The "copy" function is just a convenience thing. I don't believe that shortcodes should be changeable at all, though!

akihikodaki:
> The "copy" function is just a convenience thing.

What do you want to resolve then? If you do not care caching, copyright and domain blocking, you do not need #5145 and probably this.

> I don't believe that shortcodes should be changeable at all, though!

Then how should we resolve the collision?

Gargron:
>If you do not care caching, copyright and domain blocking

Sorry, but domain blocking is already implemented? Emojis don't get downloaded with reject_media. Also, admins can "disable" an individual emoji from admin UI. Regarding caching, this PR starts storing remote file URL and serializes an "updated" attribute; if the file URL is different or if the "updated" timestamp is newer than when we last touched our cache, it does get updated!!

>Then how should we resolve the collision?

I am not sure collision of what you are talking about? Collision of local emojis when you want to copy a remote one to a local one?

akihikodaki:
> Sorry, but domain blocking is already implemented?

I mean to block the copied emojis depending on the source, just like boosts.

> I am not sure collision of what you are talking about? Collision of local emojis when you want to copy a remote one to a local one?

Yes, that is what I am concerning. The collision which happens when there is emojis with same shortcode.

Komic:
neat~
any upgrading guideline in particular?

cwebber:
The `updated` I assume will be a date?  Also I assume these Emoji objects will be fetchable from their id?

Seems fine as an object structure.  I'm still not sure how these are used... could you paste an example?

Gargron:
@cwebber Yes, updated is a datetime just like published is on a Note. And yes they are resolveable via id. They are used in-place of the Link ones.

akihikodaki:
`updated`? I think it is irrelevant from #5145. It is misleading to say this is an alternative of the pull request.

app/lib/activitypub/activity/create.rb
```diff
@@ -86,15 +86,18 @@ def process_mention(tag, status)
...
+    return if tag['name'].blank? || tag['icon'].blank? || tag['icon']['url'].blank?
 
     shortcode = tag['name'].delete(':')
+    image_url = tag['icon']['url']
+    uri       = tag['id']
+    updated   = tag['updated']
     emoji     = CustomEmoji.find_by(shortcode: shortcode, domain: @account.domain)
 
-    return if !emoji.nil? || skip_download?
+    return if skip_download? || (!emoji.nil? && emoji.updated_at >= updated && emoji.image_remote_url == image_url)
```
nightpool: this would probably be clearer as two return statements

nightpool: do we need the `emoji.image_remote_url == image_url` check? wouldn't the save just be a no-op?

app/serializers/activitypub/image_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def type
+    'Image'
+  end
+
+  def url
+    full_asset_url(object.url(:original))
+  end
+
+  def media_type
```
nightpool: is this new? is there a practical implication to this?

app/serializers/activitypub/note_serializer.rb
```diff
@@ -142,21 +142,6 @@ def name
...
-    end
-
-    def href
-      full_asset_url(object.image.url)
-    end
-
-    def name
-      ":#{object.shortcode}:"
-    end
+  class CustomEmojiSerializer < ActivityPub::EmojiSerializer
```
nightpool: what's the difference between this and the Emoji serializer? do we need both?

nightpool (commented):
looks good to me w/ small nits

app/serializers/activitypub/image_serializer.rb
Gargron: No, it's just nice

app/serializers/activitypub/note_serializer.rb
Gargron: AMS derives serializer name from class name, virtual_tags is array of various objects, some of which are instances of CustomEmoji, but what we need is called ActivityPub::EmojiSerializer. It's easier to do this than define an AMS method that changes how serializer names are derived.

nightpool: Does the `def object_type; emoji; end` thing mean this isn't relevant? having a hard time finding the AMS docs on this, sorry.....

nightpool (approved):

PR #5245, opened 2017-10-06
Fix reloading at status column

crakaC:
Reloading browser when status detail column is shown, causes the error below.


crakaC:
Sorry for many merge commits.
I'll rebase it and then send new pull request.

PR #5246, opened 2017-10-06
Fix overflowing in web UI

lynlynlynx:
- It is needed to re-set text-overflow and overflow in `display: block;` .
With this commit, all overflowing problems are solved! 🎉 

Screenshot ( I use "THISISATESTFORVERYLONGUSERNAME", 30 letters.)
ba26

- ~~Fix: dropdown menu modal overflows outside the edge of the browser.~~

~~Before and after screenshots~~
ba27
~~Of course, modal sizing works properly.~~
ws000107

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -731,6 +731,12 @@
   .status__display-name strong {
     color: $ui-base-lighter-color;
   }
+
+  > span {
+    display: block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
 }
 
 .status__action-bar {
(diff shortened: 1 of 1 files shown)


lynlynlynx:
Unfortunately, `max-width: max-content;` property is not yet supported in Edge.
I reverted fixing overflowing dropdown menu modal.

nightpool:
Can we use the max-content anyway? does is *break* on edge or is it just not supported? Can we use `@supports(max-width: max-content)` to gate this?

nightpool:
```
display: block;
overflow: hidden;
text-overflow: ellipsis;
```

seems like it's getting repeated a lot of places. Maybe it makes sense to make a mixin rule?

Gargron (approved):

lynlynlynx:
@nightpool thanks for your comments. 
max-width breaks styles on Edge, and it is not a fundamental way, so i look for another.
Repeating a lot, thats right, ill try to organize them with mixins before long

PR #5251, opened 2017-10-06
Japanese translation (relative time)

MitarashiDango:


unarist (approved):
Note for translators: See also https://github.com/tootsuite/mastodon/pull/5171#issuecomment-333527880 for about using "時間前" instead of "時間", and others.

ykzts (approved):
I agree.

PR #5256, opened 2017-10-07
Update emoji-mart to v2.1.1

nolanlawson:
This fixes the issue of searching for e.g. `"thinking_fac"` or `"woman-facepalmi"` (https://github.com/missive/emoji-mart/issues/131). I also re-copypasta'd our version of `emoji_mart_search_light.js` over.

Incidentally this reduces the size of `emoji_picker.js` from 612K to 608K thanks to their new ES module build.

PR #5258, opened 2017-10-07
Fix custom emoji in emoji picker, persist skin tone

Gargron:
- Fix regression from #5256
- Improve emoji picker loading animation
- Persist skin tone selection in web settings
- Re-order categories so "custom" is on top

### DIFF
--- a/app/javascript/styles/components.scss
+++ b/app/javascript/styles/components.scss
@@ -2653,19 +2653,36 @@ button.icon-button.active i.fa-retweet {
   flex-direction: column;
 }
 
-@keyframes pulse {
-  0% {
-    opacity: 1;
+@keyframes heartbeat {
+  from {
+    transform: scale(1);
+    transform-origin: center center;
+    animation-timing-function: ease-out;
   }
 
-  100% {
-    opacity: 0.5;
+  10% {
+    transform: scale(0.91);
+    animation-timing-function: ease-in;
+  }
+
+  17% {
+    transform: scale(0.98);
+    animation-timing-function: ease-out;
+  }
+
+  33% {
+    transform: scale(0.87);
+    animation-timing-function: ease-in;
+  }
+
+  45% {
+    transform: scale(1);
+    animation-timing-function: ease-out;
   }
 }
 
 .pulse-loading {
-  animation: pulse 1s ease-in-out infinite;
-  animation-direction: alternate;
+  animation: heartbeat 1.5s ease-in-out infinite both;
 }
 
 .emoji-picker-dropdown__menu {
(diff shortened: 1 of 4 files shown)


nolanlawson:
LGTM, but how did this regress exactly? We upgraded from emoji-mart 2.0.1 to 2.1.1, there shouldn't have been any breaking changes…

nolanlawson (approved):

app/javascript/styles/components.scss
```diff
@@ -2653,19 +2653,36 @@ button.icon-button.active i.fa-retweet {
...
 }
 
-@keyframes pulse {
-  0% {
-    opacity: 1;
+@keyframes heartbeat {
+  from {
+    transform: scale(1);
+    transform-origin: center center;
+    animation-timing-function: ease-out;
```
nolanlawson: This new animation is cute, but why do we need to change the `animation-timing-function` in each keyframe? Can't we just set `animation-timing-function: ease-in-out` once here?

nolanlawson: [This](https://github.com/tootsuite/mastodon/commit/5404ff3995b9215d6ac610ae6b00d351cfdb8d3a) is what I'm talking about; I just tested in Chrome and Firefox and it looks the same to me…

app/javascript/styles/components.scss
```diff
@@ -2653,19 +2653,36 @@ button.icon-button.active i.fa-retweet {
   flex-direction: column;
 }
 
-@keyframes pulse {
-  0% {
-    opacity: 1;
+@keyframes heartbeat {
+  from {
+    transform: scale(1);
+    transform-origin: center center;
```
nolanlawson: `transform-origin: center center;` should normally go inside of the `.pulse-loading {}` instead of here; this might technically work in some browsers but seems safer to me not to put it here.

app/javascript/styles/components.scss
Gargron: @nolanlawson I actually generated this with http://animista.net/play/attention/pulsate

Gargron:
@nolanlawson 

>LGTM, but how did this regress exactly? 

There were no custom emojis in the emoji picker at all! In the async loading code, there was the EmojiMart.emojiIndex.search line, which was *supposed* to add them, which did nothing because EmojiMart.emojiIndex was undefined. There is some complicated stuff going on there so I just opted to just pass the emojis into the Picker component instead.

nolanlawson:
> EmojiMart.emojiIndex was undefined

I think that was my bad; I forgot to export it from `emoji_picker.js`.

PR #5259, opened 2017-10-07
In REST API, when URL can be missing, cast it to nil

Gargron:
Reference: <https://mastodon.social/@WAHa_06x36/20836861>


ykzts (approved):

DagAgren:
Nice!

PR #5260, opened 2017-10-07
Set snowflake IDs for backdated statuses

Gargron:
- Rename Mastodon::TimestampIds into Mastodon::Snowflake for clarity
- Skip for statuses coming from inbox, aka delivered in real-time
- Skip for statuses that claim to be from the future

I reimplemented #5211 with the feedback and ideas from the discussion.

### DIFF
--- a/lib/mastodon/snowflake.rb
+++ b/lib/mastodon/snowflake.rb
@@ -1,7 +1,31 @@
 # frozen_string_literal: true
 
-module Mastodon::TimestampIds
+module Mastodon::Snowflake
   DEFAULT_REGEX = /timestamp_id\('(?<seq_prefix>\w+)'/
+
+  class Callbacks
+    def self.around_create(record)
+      now = Time.now.utc
+
+      if record.created_at.nil? || record.created_at >= now || record.created_at == record.updated_at
+        yield
+      else
+        record.id = Mastodon::Snowflake.id_at(record.created_at)
+        tries     = 0
+
+        begin
+          yield
+        rescue ActiveRecord::RecordNotUnique
+          raise if tries > 100
+
+          tries     += 1
+          record.id += rand(100)
+
+          retry
+        end
+      end
+    end
+  end
 
   class << self
     # Our ID will be composed of the following:
@@ -112,6 +136,13 @@ module Mastodon::TimestampIds
           $$ LANGUAGE plpgsql;
         SQL
       end
+    end
+
+    def id_at(timestamp)
+      id  = timestamp.to_i * 1000 + rand(1000)
+      id  = id << 16
+      id += rand(2**16)
+      id
     end
 
     private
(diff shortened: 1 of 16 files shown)


Gargron:
@aschmitz

lib/mastodon/snowflake.rb
```diff
@@ -1,8 +1,31 @@
...
+      if record.created_at.nil? || record.created_at >= now || record.created_at == record.updated_at
+        yield
+      else
+        record.id = Mastodon::Snowflake.id_at(record.created_at)
+        tries     = 0
+
+        begin
+          yield
+        rescue ActiveRecord::RecordNotUnique
+          raise if tries > 100
```
aschmitz: We need to increment `tries` in this block too.

lib/mastodon/snowflake.rb
```diff
@@ -114,6 +137,13 @@ def ensure_id_sequences_exist
       end
     end
 
+    def id_at(timestamp)
+      id  = timestamp.to_i * 1000
```
aschmitz: Note that this forces the millisecond part to zero. If we'll never get one from a source where this would be called, this is vaguely OK (though I'd prefer to add them in if they exist), but we might also want to just pick a random millisecond if it's zero? Lower likelihood of hitting a duplicate ID, not that it's all that likely in the first place.

aschmitz (commented):
Looks good to me with the exception of one issue and a minor comment.

I'm not sure if app/workers/activitypub/processing_worker.rb is the only place where we'd be backfilling old statuses (nothing in OStatus?), but I haven't actually spent much time on that side of the code, so I assume you know what's up there.

lib/mastodon/snowflake.rb
Gargron: I figure since we only ever call this from networked timestamps, usec would always be 0, therefore it can be skipped. I agree we could do a random millisecond part though. ...How exactly do I do that? `+rand(1000)`?

lib/mastodon/snowflake.rb
Gargron: Whoops

aschmitz: Yeah, I went back and checked: I missed it in my copy too. :disappointed:

lib/mastodon/snowflake.rb
aschmitz: Yep, I kind of figured that was why, and it doesn't bother me too much.

Just adding a `+ rand(1000)` should do it, indeed.

Gargron:
@aschmitz @beatrix-bitrot Updated yo

beatrix-bitrot (approved):
comments have been addressed!!!!!!!!

aschmitz:
Oops, yep, sorry, looks good!

PR #5261, opened 2017-10-07
Show buffering in video player

Gargron:
Lighter line behind the progress bar when buffered


nolanlawson (approved):
Tested it, looks great

PR #5264, opened 2017-10-07
Use ES module build of react-router-dom

nolanlawson:
This reduces the size of `common.js` from 797 kB to 791 kB by using the ES module builds of `react-router-dom` and `history`.

I tested, and there is no benefit to deep-linking `react-router-dom` because it exports a `"module"`, which appears to do tree-shaking correctly in Webpack. Plus, to deep-link the ES modules as opposed to the CJS modules, we would need to import e.g. `'react-router-dom/es/Route'` instead of `'react-router-dom/Route'`. So I use `import { Route } from 'react-router-dom'` which is terser and has the same effect.

OTOH there is a bug currently in `react-router-dom` where it double-imports `history`'s CJS and ES build (https://github.com/ReactTraining/react-router/issues/5576). So to work around that, we can just use `NormalModuleReplacementPlugin` for now.


sorin-davidoi (approved):
Glad this is no longer required 👍

nolanlawson:
Yeah apparently some libraries benefit from deep-linking whereas others don't, which I guess is just the state of the npm ecosystem. ¯\\_(ツ)\_/¯

PR #5265, opened 2017-10-08
Dynamically calculate card height for embeds instead of padding

Gargron:
The padding trick was hard-coded to a 16:9 ratio, but we can use width and height provided from OEmbed information and width of the card itself to calculate a new height

Unfortunately, imgur's embeds seem generally broken in terms of height, I can observe that on birdsite as well

### DIFF
--- a/app/javascript/mastodon/features/status/components/card.js
+++ b/app/javascript/mastodon/features/status/components/card.js
@@ -30,6 +30,10 @@ export default class Card extends React.PureComponent {
     maxDescription: 50,
   };
 
+  state = {
+    width: 0,
+  };
+
   renderLink () {
     const { card, maxDescription } = this.props;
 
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/features/status/components/card.js
```diff
@@ -75,14 +79,25 @@ export default class Card extends React.PureComponent {
     );
   }
 
+  setRef = c => {
+    if (c) {
+      this.setState({ width: c.offsetWidth });
```
nolanlawson: So unfortunately `offsetWidth` triggers layout… but this is only for video embeds right? Basically YouTube and whatnot?

Gargron: Yes. YouTube, Vimeo, Soundcloud, Imgur, Twitch Clip, etc

Gargron: Mind you, StatusCard is only ever mounted from DetailedStatus!

nolanlawson (approved):

PR #5266, opened 2017-10-08
Use object URI only in Announce, instead of embedding

### LINKED ISSUE #5178: Discrepancy in activitypub announce activity hashtag url if two instances are involved
If i get an announce activity (boost) from instance-1 with content with hashtags from instance-2 the tag urls of the activity points to instance-1 (the boosters instance).

I would expect the urls to point to instance-2 since it's a boost.

* * * *

Gargron:
Fix #5178


Gargron:
@akihikodaki @unarist

akihikodaki (approved):
It looks nice. You may also leave comments in `ActivitySerializer` why you made this change.

PR #5269, opened 2017-10-08
SCSS placement for Make the theme creation easier

YaQ00:
**Please review my poor English**

SCSS placement for addition of theme function.

This allows administrators to easily create themes (Without causing conflicts)!


nightpool:
What's the difference between this and the existing theme functionality?

YaQ00:
@nightpool 

Up to now:
The theme function did not exist.
Therefore, it was no problem to name "application.scss".
The administrator can customize "custom.scss" by reading it from "shared.js".

2.0:
With the addition of theme function, 
・"application.scss" is no longer a suitable name.
・custom.scss was abolished.
I am **very concerned** about these.
Also, the administrator has to create a theme directory and deal with conflicts. However, these SCSSs with relative paths can not be diverted.
I think that the file created by the administrators should be stored in a dedicated directory.
Besides, it is necessary for the administrators to make it easy to divert existing files.

This PR:
・By defining custom.scss in the "custom" directory, you can create common parts of any theme.
・You can easily imagine which SCSS the default theme uses, from the directory structure.
・By creating themes in line with the guidelines (README.md), administrators can easily share themes.
=>Administrators-friendly

lynlynlynx:
I agree with these two,
- giving directory structure to every themes
- changing "application.scss" name to "theme_default.scss"

----

> The administrator can customize "custom.scss" by reading it from "shared.js".

Special functions for custom.scss is already abolished after multiple theme support was implemented.
https://github.com/tootsuite/mastodon/pull/4959/files#diff-52e5ee150ad902a05f7767da436cb5eeL18
As we became more flexible to create themes, there is no meaning to force using custom.scss to create common parts of all of themes.

YaQ00:
@lynlynlynx 

> As we became more flexible to create themes, there is no meaning to force using custom.scss to create common parts of all of themes.

When you want to apply minimal customization to the default theme...
What kind of things do you think is appropriate as a structure that keeps code additions to a minimum?

In order to prompt the transition from the customization method so far, I think that it is appropriate that custom.scss is imported to theme_default.scss.

In this structure,
· Common code for all themes (styles/custom)
· Code of default theme **(administrator never touches)**
· Custom theme code

It can be roughly divided.

nightpool:
@YaQ00 
>When you want to apply minimal customization to the default theme...

Current themes are easily able to apply minimal customization to the existing theme. The themes PR was specifically written so that there is no loss of functionality or migration required—everything you can do with custom.scss, you can do by adding a "`custom: custom.scss`" theme and making it the default.

YaQ00:
@nightpool OK, I think that's right.

What do you think about parts other than "custom.scss"?

lynlynlynx:
@YaQ00 

> When you want to apply minimal customization to the default theme...
What kind of things do you think is appropriate as a structure that keeps code additions to a minimum?

Even more if you think admin should not touch the default mastodon theme, the default theme should remain the default theme, and any customization should be done on a new one. As nightpool says, it is possible to set this new one as default, and users can also select the mastodon theme.
If admin does not particularly care about modifying the default mastodon theme, they can create a diff file on their own, and just add it to theme_default.scss, it's easy.

> In order to prompt the transition from the customization method so far, I think that it is appropriate that custom.scss is imported to theme_default.scss.

Since it is the essence of the multiple theme function that we can customize themes very flexibly, it is not needed any intervening features especially for pepple who had been able to use custom.scss with an understanding of javascript "shared.js". Adding "custom.scss" just complicates the multiple theme function, and causes coufusion.

YaQ00:
@lynlynlynx @nightpool 
I Close the PR once and PR again only with layered and renamed.

PR #5275, opened 2017-10-08
Track frequently used emojis in web UI

### LINKED ISSUE #5144: Emojos in the new picker don't have tooltips
(I understand that's part of the new library we use, but it feels like a regression)

### LINKED ISSUE #5203: [emojos] emoji picker "frequently used" should update with used shortcodes
It would be nice if the two ways of entering emoji were connected and the "frequently used" section actually reflected what you frequently use, not just through the picker.

Gargron:
- Click in emoji picker as well as selection of autosuggest increase counter for that emoji
- Settings reducer gets some new events that track whether the settings are "dirty" (unsaved changes)
- Settings save now debounced by 5 seconds (saves API calls)
- Fix #5203
- Fix #5144

### DIFF
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -1,6 +1,7 @@
 import api from '../api';
 import { throttle } from 'lodash';
 import { search as emojiSearch } from '../features/emoji/emoji_mart_search_light';
+import { useEmoji } from './emojis';
 
 import {
   updateTimeline,
@@ -305,6 +306,8 @@ export function selectComposeSuggestion(position, token, suggestion) {
     if (typeof suggestion === 'object' && suggestion.id) {
       completion    = suggestion.native || suggestion.colons;
       startPosition = position - 1;
+
+      dispatch(useEmoji(suggestion));
     } else {
       completion    = getState().getIn(['accounts', suggestion, 'acct']);
       startPosition = position;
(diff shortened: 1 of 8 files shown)


nolanlawson (changes requested):
Everything LGTM except for a couple things:

1. Frequently used emoji don't seem to survive a refresh? At least when I test locally
2. I feel we should cut these off at 8 or 16 so that these don't grow without bound (see screenshot)

nolanlawson:
Seems they grow without bound:

screenshot 2017-10-08 11 43 30

Gargron:
@nolanlawson They grow to 4 lines, I took over that default from emoi-mart - do you think it should be two lines only?

Gargron:
Also, save is debounced by 5 seconds, so yeah if you immediately refresh they won't save, but I doubt that will happen in practice

nolanlawson:
I feel like 2 or maybe 3 lines is best but I'm not a heavy emoji user. Maybe good to ask others?

nolanlawson (approved):

MightyPork:
I was trying to test this, but after deleting a custom emoji that was in my recents, the picker does not open anymore.

screenshot_20171008_221440

screenshot_20171008_221424

MightyPork:
I still managed to confirm that tooltips work though :+1:

Gargron:
This is more of a general emoji-mart problem, it does need to be fixed, though separately.

MightyPork:
yeah just saying I can't test any further because my picker now always crashes. this is stored server-side? i hope i can find some "phpmyadmin" that works with postgre :S

MightyPork:
good, managed to find and delete the JSON object in web_settings using rubymine. Other than that, seems to work okay! Shortcodes from the suggestions bar are added correctly

PR #5279, opened 2017-10-08
add clustering for node.js streaming.

matyapiro31:
* add npm package cluster.
* add cluster on pacakge.json, and yarn.lock.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "babel-preset-env": "^1.6.0",
     "babel-preset-react": "^6.24.1",
     "classnames": "^2.2.5",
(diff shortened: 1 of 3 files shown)


Gargron:
But we already have clustering..?

matyapiro31:
clustering for service mastodon-streaming.I think it's not supported now, is it?

Gargron:
https://github.com/matyapiro31/mastodon/blob/74d31c21317011335fff4e2f0f8fb21b134c358e/streaming/index.js#L492

akihikodaki:
I think this pull request should be closed, isn't it?

PR #5282, opened 2017-10-09
Remove timestamps on any option

tribela:
Timestamps were not removed on non-S3 usages.

Related to #2185 and #2123


ashfurrow (approved):
Good catch 👍

PR #5292, opened 2017-10-09
Grouped web push notification should not have an icon

Gargron:
I'm frankly not sure what happens when there's no icon - maybe it displays no icon next to the notification, or maybe it displays the badge (ideally), but displaying the launcher icon next to the notification definitely does not feel native.

### DIFF
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -14,7 +14,6 @@ const notify = options =>
           .sort((n1, n2) => n1.timestamp < n2.timestamp)
           .map(notification => notification.title).join('\n'),
         badge: '/badge.png',
-        icon: '/android-chrome-192x192.png',
         tag: GROUP_TAG,
         data: {
           url: (new URL('/web/notifications', self.location)).href,


nolanlawson:
So I think this may need some investigation, because AIUI this icon will be used for more than just the notification grouping:

1. Icon shown on the home screen when doing "add to home screen"
2. Icon shown on the splash screen when launching from the home screen
3. Criteria for the [add to homescreen banner](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android) in Chrome, i.e. if we remove the `"icon"` member then this banner will no longer appear
4. Various scenarios for non-Chrome browsers (e.g. as of the next Edge release PWAs will be listed in the Windows Store, assuming they have an `"icon"`)

So I think we should probably do a thorough audit of what effect this actually has on various browsers (e.g. Samsung Internet, Opera, Firefox have different PWA implementations than Chrome even on the same version of Android). But given the issues above I would be inclined to just keep the `"icon"` for now even if it looks weird as a notification.

FWIW I believe this may also change in Android Oreo because they are doing "web APKs" now, which is a new PWA implementation where the app is a true separate APK as opposed to a homescreen bookmark.

Gargron:
@nolanlawson Wait, this removes the icon property from one type of Notification, there's no way this affects the splash screen or home screen or add to homescreen banner - that's based on manifest.json only

sorin-davidoi:
I would suggest setting the icon to the badge and updating the badge to have color (if I remember correctly Android makes it monochrome automatically).

Gargron:
@sorin-davidoi So what happens when the icon property on the notification is just missing? My naive guess was that it would use the badge instead, automatically?

sorin-davidoi:
I don't know, but I would venture to guess that it is browser dependent (not in the spec).

Gargron:
Alright, this is not very important

nolanlawson:
> Wait, this removes the icon property from one type of Notification, there's no way this affects the splash screen or home screen or add to homescreen banner - that's based on manifest.json only

My bad, I misunderstood. In any case seems like it needs testing because it'll vary from browser to browser

PR #5303, opened 2017-10-10
pam authentication

devkral:
This pull request allows to authenticate against pam.
If in the pam environment a email address is found, it is used.
For the first login a user account with the pam password is created if pam allows login.
Elsewise the account data is updated. Login with the email is possible but neither email nor password are updated this way.
Pam login disables resetting password, password length check (this is why I removed validatable; every password allowed by pam should work for pam authentication)

I think this solution would be more general than using ldap,...


devkral:
Should work now. Sry, I forgot to test.

beatrix-bitrot:
@devkral test suite is still failing pal!

devkral:
I will have look. I ran the test suite without problems. I think its the deliberately missing schema.db update (MR conflicts are pre-programmed and many people hate updates there)

devkral:
hiredis-rb build is failing. It is delaying the updates.

devkral:
it was the missing schema.rb entry.

devkral:
Ping?

akihikodaki:
I really like this feature, but I have not really conceived that we should add a column for this. The configuration is really helpful in certain uses, but such configuration should be still rare.

devkral:
I think about two-way logins (and registrations): you can either login or registrate the conventional way or with pam. If you don't have an extra column it opens security holes and complicates the implementation.
For example if you prefix names with pam_ you have to ensure, that no other user uses this name prefix. Also you have to strip the prefix from the displayed name (impersonation issues arise from this). In the db administration have to look for every name with and without prefix.
The extra column pam_name has the advantage that a fallback exists in case the username is given to somebody else.

Gargron:
Can someone explain what PAM is, where it is used, and what the importance of it is?

devkral:
Imagine PAM as local omniauth just much more powerful.
It is a standard for 20 years mostly used by organisations.

You can have complex rulesets via servicefiles with multiple authentication providers (e.g. fingerprint, local password, ldap, kerberos), environment setup and much more. Just look it up.
The best part is you don't have integrate these authentication methods in your software, pam does it.

It would be very useful for all kinds of organisations or people who want to be able to login with their normal user account, e.g. shared servers, local computer.

devkral:
rebased

akihikodaki:
> I think about two-way logins (and registrations): you can either login or register the conventional way or with pam. If you don't have an extra column it opens security holes and complicates the implementation.

I understand your concern and your solution would work perfectly to addres it. But my concern is not really about PAM configuration, but about other configurations not using PAM. The table (`accounts`) is big in some instances, and having an extra column would not be preferred.

devkral:
How big is the bloat? Can the databases optimize it out?
Because an extra table could cost much more space.

devkral:
Further alternatives: switch between both modis with a special email domain or a pam flag for the user.

akihikodaki:
Pawoo probably has the largest population (310,931 users). Adding an extra column is not so problematic for the scale. I admit that it is just not my preference. 😕

devkral:
What would be the preferred alternative? E.g. special email address like username@pam?

devkral:
Ignore the deleted comment, I did misread it. But still if you have a better proposal just say :).
But maybe you have implement it (I am not sure if I can).

devkral:
I found a better solution without pam_name: We use my fork:
https://github.com/devkral/devise_pam_authenticatable2
there a blank password could be used to differ between pam and non pam 
I just have to create a new release and release it as gem

devkral:
here is the result (needs some tests) (I will switch to this later):
https://github.com/devkral/mastodon/tree/pam_dev5

db/migrate/20171004112117_add_pam_name_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddPamNameToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :pam_name, :string, null: true, default: nil
```
Gargron: You have a unique validation on this column, but it's not safe against race conditions. I suggest a unique index as a db-level constraint.

config/initializers/devise.rb
```diff
@@ -30,10 +30,55 @@
...
 
+module Devise
+  mattr_accessor :pam_authentication
+  @@pam_authentication = false
+  mattr_accessor :pam_service
+  @@pam_service = 'rpam'
+  module Strategies
+    class PamUserAuthenticatable < Authenticatable
+      def valid?
+        Devise.pam_authentication && params[scope] && params[scope][:email] && !params[scope][:email].index('@')
```
Gargron: People could have e-mails as LDAP identifiers, couldn't they?

I am against having both registrations and PAM *at the same time*, if PAM is active all logins should go through it - at least that's the LDAP principle from other PRs if I remember correctly.

config/initializers/devise.rb
```diff
@@ -298,4 +343,9 @@
   # When using OmniAuth, Devise cannot automatically set OmniAuth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # authentification with pam possible
+  #config.pam_authentication = true
```
Gargron: This and the one below should probably become ENV variables so people don't have to edit source code to configure it.

config/locales/simple_form.en.yml
```diff
@@ -51,6 +51,7 @@ en:
         severity: Severity
         type: Import type
         username: Username
+        username_or_email: Username or Email
```
Gargron: I don't see which form uses this label but it's again my concern about mixing PAM-users and registration-users.

Gargron (changes requested):

devkral:
sry for overwriting. This is a new version having nearly nothing in common with the old version

config/locales/simple_form.en.yml
devkral: you have to uncomment config.pam_authentication = true in initializers.
This feature is optional.

devkral:
for a clean solution I added the remember_token column. With it a password is not necessary anymore

devkral:
I hope I understood the http_authenticatable method correctly. Can please somebody check it?

devkral:
now it should work. Sry didn't saw, that the password_required? rules are more complicated.

devkral:
When I see it correctly the session stuff introduced by me can be removed.

devkral:
sry still have a bug

devkral:
it seems as if the valid? function of strategies is sometimes not checked.
I use now a fallback and return nil if PAM is disabled.

devkral:
sry valid? is correct. The fork I used still messed up password and other attributes. After cleaning up, everything looks fine now.

config/locales/simple_form.en.yml
devkral: the login form uses it. It may wasn't there because of indentation issues.

devkral:
in my tests the PR doesn't harm anything anymore and allows logins with pam. Shall I refactor it?

devkral:
Some design details:
* I had to add a new column for the remember_token
* If the encrypted_password is set, the login blocked. This is the easiest way to block pam users from logging into database accounts
* - there is currently no way to prevent registrations for pam names except to turn off registrations

devkral:
a fix for the last problem can be found here:
https://github.com/devkral/mastodon/tree/pam_block_registration

It is not ready because it shows the wrong error

app/models/user.rb
```diff
@@ -84,6 +87,33 @@ class User < ApplicationRecord
...
+  end
+
+  def pam_get_name
+    return account.username if account.present?
+    super
+  end
+
+  def pam_setup(_attributes)
+    acc = Account.new(username: pam_get_name)
+    acc.save!(validate: false)
```
Gargron: I'm not entirely sure `validate: false` works on the `!` version of the save method, I have a vague memory of running into this. More importantly, why do you need to not validate the account? I'm pretty sure the username is the *only* required part, and you do give it here.

db/migrate/20180109143959_add_remember_token_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddRememberTokenToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :remember_token, :string, null: true
```
Gargron: I don't see what is using this column?

devkral: this allows remember_me for models without password. Sadly it was the cleanest solution possible.

config/locales/simple_form.en.yml
devkral: everything ok now? Or shall I change something?

Gargron (approved):
I am approving this but I really want a second opinion. I am not entirely happy with PAM configuration being hardcoded into an initializer, since all *other* settings are controllable via environment variables. Docker deployments would not be able to easily configure these. I am expecting future PRs to modify some parts of this functionality before the next release.

immae:
This is a great job, thanks @devkral

Gargron:
Turns out we need some kind of system level dependency for the rpam2 gem to install. We should make this optional as PAM users will be a minority...

PR #5304, opened 2017-10-10
Remove prop types from external libraries

nolanlawson:
This adds a second `babel-loader` which only applies `transform-react-remove-prop-types`. It removes 1.21 kB from `common.js` and 114 bytes from `application.js` (every little bit counts!).

You might recall that in https://github.com/tootsuite/mastodon/pull/3937 we ran into a regression because of Babel being applied to third-party libraries (`node_modules`). This PR ensures that when Babel runs on `node_modules`, it only runs the one transform.

This is live on https://malfunctioning.technology ; I tested and didn't see any errors due to removing the prop-types.

### DIFF
--- a/config/webpack/loaders/babel.js
+++ b/config/webpack/loaders/babel.js
@@ -4,11 +4,7 @@ const env = process.env.NODE_ENV || 'development';
 
 module.exports = {
   test: /\.js$/,
-  // include react-intl because transform-react-remove-prop-types needs to apply to it
-  exclude: {
-    test: /node_modules/,
-    exclude: /react-intl[\/\\](?!locale-data)/,
-  },
+  exclude: /node_modules/,
   loader: 'babel-loader',
   options: {
     forceEnv: env,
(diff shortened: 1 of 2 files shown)


config/webpack/loaders/babel.js
```diff
@@ -4,11 +4,7 @@ const env = process.env.NODE_ENV || 'development';
 
 module.exports = {
   test: /\.js$/,
-  // include react-intl because transform-react-remove-prop-types needs to apply to it
-  exclude: {
-    test: /node_modules/,
-    exclude: /react-intl[\/\\](?!locale-data)/,
```
nolanlawson: There's no more need to add special behavior for react-intl. The reason we were doing this before was to explicitly remove the prop-types from react-intl while not breaking on other babel transforms being applied to it. Since the new loader only runs one babel transform, we can just apply it to all `node_modules`, whereas this loader doesn't apply at all to `node_modules`.

Gargron (approved):

PR #5305, opened 2017-10-10
Fix #5293 - Pre-fill frequently used emojis to avoid bugs

### LINKED ISSUE #5293: New emoji picker doesn't work correctly when "frequently used emoji" doesn't exist
When "frequently used emoji" doesn't exist:

1. Clicking 2nd tab doesn't work.
2. 2nd tab doesn't be highlighted as it scrolls.

See this...
i

* * * *

Gargron:


nolanlawson:
Feels like a hack, but makes sense for now assuming we can't fix it in emoji-mart yet. Is the list based on anything?

nolanlawson (approved):

Gargron:
@nolanlawson I took the list from emoji-mart, but replaced some negative icons with things that are more popular on Masto, like thinking face, ok hand, yum, etc

PR #5311, opened 2017-10-11
Make the apple-touch-icon great again

uvexz:
It looks like this:


matyapiro31 (commented):
No need to add color for dots.

PR #5315, opened 2017-10-11
Avatar Emoji

utam0k:
AvatarEmoji is a custom emoji used Account's avatar.
It uses metadata of short_code, url, static_url like CustomEmoji.
shortcode is `@username`, url is ` avatar.original_url`, static_url is `avatar.static_url`.
When tooting with `:@username:`, it changes to avatar of that user.

Unlike CustomEmoji, AvatarEmoji uses existing resources.
Therefore, it is not necessary to add or register emojis.

It has already been introduced in friends.nico and actually used.
In this PullReq, not only friends.nico but also AvatarEmoji is available for all instances.
https://friends.nico/@utam0k/37459454
https://friends.nico/@utam0k/37101753
https://friends.nico/@utam0k/36949650
image


utam0k:
Currently it can be used in status text.
We would like to ask our opinion about AvatarEmoji.

Aldarone:
I'm a huge fan of this idea but the styling is not so good. Plus, I don't like that the nickname is removed.

I think riot.im does it really well : 
capture d ecran_2017-10-11_20-27-46

If you can make it this way I'd even be in favor of it becoming the default mention display !

Gargron:
I don't have time/energy to review this and include it in 2.0.0, just saying. But it might come in after that.

nullkal:
I think this feature works well with instances only for emoji registration purpose (e.g. https://moji.m.to/). We can reuse fetching/updating mechanisms of profile to get/update emojis.

Aldarone:
Is there any progress on that matter ? @utam0k what do you think about my design suggestion ?

utam0k:
@Aldarone 
Thank you for your comment.
I think this feature is kind of Emoji.
Your design is good, however I feel is it is not like Emoji.

utam0k:
@Gargron 
Hi Mr.Gargon.
Thank you for your release of 2.0.

Please tell me more your opinion.
Can I continue to develop more feature in Avatar Emoji?

Gargron:
If you've put a :-1: or :confused: on this PR, I'd like to hear in text why. Assuming that the code for this feature is clean, I don't see much harm in supporting it, especially if some instances in the wild already support it (friends.nico)

On the other hand, it's a bit strange, because these should count as mentions, shouldn't they? It's a bit unusual that you could just use someone's profile picture as an emoji without any permission or knowledge of the user.

I don't know so I would appreciate extra feedback.

renatolond:
I saw this in the federated timeline a couple of times coming from friends.nico and it's interesting, don't think I would use it, but I agree that maybe it should be a mention too. Seems to me you're @-ing the user, just using an emoji instead.

akihikodaki:
I expect this feature is useful, but I am still not for this change because:
1. it is much dependent on context,
2. some may not want others to use their avatar as emojis, and
3. it may lead people to create dummy accounts for emojis.

1\. Why do we have to introduce a new syntax and federation protocol just for avatar emojis? Are we going to add new syntax every time we add emojis from different kind source? A syntax introduced in suc a way will not have backward compatibility, and make the implementation complex.

2\. It is easy to allow everyone to use avatars of anyone else as emojis, but I suspect some people do not want that. Maybe we could make the option opt-in (NOT opt-out. Opt-out may not work for older versions of Mastodon and new federation implementations, and it is hard to notice the feature for users on such implementations.) However, it also raises a different question: who bothers to opt-in the feature?

3\. I have seen some people launched instances to create their own emojis, but then abandoned. Creating accounts is an easier option for such people to create emojis and yet polluting the network.

Gargron:
I see Aki's point and I am inclined to agree. Sorry @utam0k! :cry:

PR #5322, opened 2017-10-11
In thread view, only scroll first update, scroll to replied-to post

Gargron:
- Only perform the "scroll into view" on first update, not when e.g. additional replies are appended
- Instead of scrolling to the focused status, scroll to the replied-to status, but align that scroll to top

### DIFF
--- a/app/javascript/mastodon/features/status/index.js
+++ b/app/javascript/mastodon/features/status/index.js
@@ -78,6 +78,7 @@ export default class Status extends ImmutablePureComponent {
 
   componentWillReceiveProps (nextProps) {
     if (nextProps.params.statusId !== this.props.params.statusId && nextProps.params.statusId) {
+      this._scrolledIntoView = false;
       this.props.dispatch(fetchStatus(nextProps.params.statusId));
     }
   }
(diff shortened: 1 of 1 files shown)


nolanlawson:
This LGTM, one small issue I can see is

1. You load toot # 4 in a thread
2. Toot # 3 is scrolled to
3. Toots # 1 and # 2 are loaded later, and they're above the fold

So you might not realize that 1 and 2 exist unless you think to scroll up.

But then again it'd probably be too distracting if it kept scrolling to the top as new ancestor toots come in, so :+1: from me.

nolanlawson (approved):

unarist:
For nolan's concern, I think it would be nice if we scroll to toot # 3 with some offset.

Gargron:
>Toots # 1 and # 2 are loaded later, and they're above the fold

This does not happen, all ancestors are loaded at once

nolanlawson:
@Gargron That's odd, maybe we're talking about two different things. This is what I saw, in both Chrome and Firefox: https://gfycat.com/FriendlyFlawlessGermanwirehairedpointer

Gargron:
Your gif shows the behaviour that I intended :thinking:

nolanlawson:
Yeah so my point was that I was a bit surprised that after clicking on
4, it scrolled to 3 but didn't show 1 or 2 unless I scrolled up. But if
this is what you intended, I think it's fine. :)
-----
Nolan Lawson
https://nolanlawson.com


On Wed, Oct 11, 2017, at 11:18 AM, Eugen Rochko wrote:
> Your gif shows the behaviour that I intended 🤔


> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[1], or mute the thread[2].> 


Links:

  1. https://github.com/tootsuite/mastodon/pull/5322#issuecomment-335901585

PR #5325, opened 2017-10-11
Only preload JS in /web pages

nolanlawson:
Issue: preloading these JS files for public pages (e.g. `/about/more`, public toots) is a waste, because signed-out users will never need them. In this PR, I wrap the preloads in an `if user_signed_in?` check.

This is a bit hacky, because it might be cleaner if we just distinguished between public pages and non-public pages, but I wasn't sure how to do that.


nightpool (dismissed):
these calls should be moved to `app/views/home/index.html.haml`, the same place we load the application bundle. (`javascript_pack_tag.*application`)

nolanlawson:
Yes you're right, will fix

nolanlawson:
Fixed! Now `preload` is only used in `/web/*`, but not e.g. `/about/more`, `/settings`, or a public toot page.

ykzts (approved):

PR #5336, opened 2017-10-11
Make theme directory and Rename application.scss

YaQ00:
Up to now (~v1.6.1) :
The theme function did not exist.
Therefore, it was no problem to name "application.scss".

2.0:
With the addition of theme function, "application.scss" is no longer a suitable name.
I think that the file created by the administrators should be stored in a dedicated directory, But when diverting a file, the administrator can not put the @import master file in a folder other than the Styles folder. Because it's a relative path.
=> hard to customize

This PR:
・Renamed "application.scss" to "theme_default.scss".
・You can easily imagine which SCSS the default theme uses, from the directory structure. 
image

Example
image

---

Added on Oct 18,2017 

This PR is to provide clarity to the theme creators by clearly stratifying the SCSS.
(There is an error in the title, the rename has been rejected)

With this PR, the theme creator can create a standard that "the theme is right under the styles directory and the actual SCSS is hierarchical".
The Styles directory is concise and you can determine how many themes you have right now.


lynlynlynx (dismissed):
All that's left is to resolve conflicts, please 😉

lynlynlynx:
Codeclimate issues are not related to this pull request, so now all checks have passed.

You can check the previous discussion at #5269 .

lynlynlynx:
@YaQ00 resolve conflicts plz.

And one more.
Considering tootsuite/documentation#421 .
A default theme is not always the mastodon theme, so..
the mastodon theme's name should be "theme_mastodon.scss", and folder name should be "mastodon" .

nolanlawson:
Won't this break custom themes that are doing

```scss
@import 'application';
```

? It seems like everyone will have to change their custom themes to `@import 'theme_default';` instead…

lynlynlynx:
@nolanlawson 
I know, and I have to change my custom themes too, but multiple theme support is not officially released yet.
It is a fact that adding many themes makes "styles" directory complicated, and "application.scss" name is no longer a suitable name.
So, I think now is the last chance to change directory and name.

nightpool:
@lynlynlynx Custom theme support was released as part of Mastodon v1.2, the only changes that 2.0 makes is the ability to have *multiple* custom themes.

nightpool:
This is probably a good change from an architectural perspective, but i'm unsure if there's a reason to make it (with the associated breakage costs) under the current usage. Do you have a motivating use-case for the PR you can share, or code you can link that uses it?

nightpool:
"But when diverting a file, the administrator can not put the @import master file in a folder other than the Styles folder. Because it's a relative path"

why not just use `@import "../application.scss"`?

lynlynlynx:
> Custom theme support was released as part of Mastodon v1.2, the only changes that 2.0 makes is the ability to have *multiple* custom themes.

@nightpool That's right, but custom theme support using `custom.scss` is already **abolished** after _multiple_ theme support was implemented. So, in the first place, everyone who uses custom theme till v1.6.1 must change their themes on v2.0.

----

I know you two wonder we force administrators to change their themes, but I think the number of instances using _multiple_ theme support is very few, only a fraction of master tracking instances.
Hmm, I'm just a css adjuster, not a vigorous theme creator. My opinion is a bit biased?

nightpool:
@lynlynlynx Like I said on the last ticket—when upgrading to 2.0 there's no need to change themes, you just need to change the YML file.

Anyway, I think putting this stuff in sub-folders is basically good, but I don't know if I like the change from "application -> default". That might just be nickpicking though. Most of the time custom themes are *additive* to application styles, not *replacing* them, so "application" is a better term for it.

lynlynlynx:
@nightpool Thank you for telling me your opinion. Certainly, changing the name causes confusion in spite of nothing functionally changed. I come around to your opinion.

So finally, my opinion is below. Ah, font files folder should be directly under styles folder?
````
📁 styles
|-- 📁 fonts
|-- 📁 mastodon
|   `-- scss files
`-- application.scss
````
How do you think about it?
@YaQ00 @nolanlawson

YaQ00:
@nightpool 
Motivation of this PR:
When creating a theme, I could not place the imported file in the "theme" directory. It is difficult to see only imported files when placed in the Styles directory. "application.scss" is just one of themes. It is strange that it is named as if all themes are being controlled.
Although the opinion of the nightpool is wonderful, it is not so good that it is conscious as if the theme creator needs to use "application.scss". 
Please understand that the theme creator has already become confused.
As @lynlynlynx said earlier, I think that the theme creator is the last chance to make it easy to understand.

> the mastodon theme's name should be "theme_mastodon.scss", and folder name should be "mastodon" .

> Ah, font files folder should be directly under styles folder?

I agree with putting "mastodon" and "fonts" directly under.
However, I think that the name change of "application.scss" is an important part of this PR.

I am sorry for the delay in my opinion.
Let's discuss a little more. If application.scss is better, I will obey.

YaQ00:
When more people create themes, these will be too late. That is, it is a 2.0 release.

lynlynlynx:
@YaQ00 Thanks for adopting my opinion, but changing the name of application.scss is not realistic option. I can understand what you want to say, but I think it is too ideal. As I said before, changing the name causes confusion in spite of nothing functionally changed. I got another opinion from an administrator using custom themes, "Too big change forces too hard changing my sass code.".

This commit must be taken in before v2.0 release. We cant make too big change only in few days, so I still 
fully adopt nightpool's opinion about the name of application.scss.

YaQ00:
@lynlynlynx Thank you for your polite reply. I only made the directory hierarchy. Please review!

nightpool (approved):
thanks for the update! I'm glad we could all reach consensus on this issue.

my thoughts are similar to @lynlynlynx's

lynlynlynx (approved):
Thanks everyone 🎉 I'm grad to come to an agreement too, and approve this!

nolanlawson:
Could someone please summarize the current state of this PR? What changes will admins have to make to their themes if this gets merged?

nightpool:
Nolan: none. This PR only changes the backend organization of the scss
files so that there's a more clear space and example for putting custom
themes in their own folder

unarist:
@YaQ00 Could you append bottom lines into the summary of this PR?

Also we should have updated the title...this PR no longer renames application.scss...

YaQ00:
@unarist Sorry, I'll be careful.

YaQ00:
This PR is to provide clarity to the theme creators by clearly stratifying the SCSS.
(There is an error in the title, the rename has been rejected)

With this PR, the theme creators can make a standard of "theme is right under Styles and other themes are hierarchical".
The Styles directory is concise and you can determine how many themes you have right now.

Issue #4860, opened 2017-09-08
Web notifications should replace line breaks with spaces


noiob:
This toot
![grafik](https://user-images.githubusercontent.com/8197071/30229104-df66c34c-94e0-11e7-8036-93be565fbe8a.png)
yields the following notification
![grafik](https://user-images.githubusercontent.com/8197071/30229126-f54a37fc-94e0-11e7-867b-33912e34a595.png)
which is pretty unreadable. I get that stripping line breaks is needed, but they should be replaced with spaces.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

PR #5361, opened 2017-10-12
replace newlines in desktop notifs with spaces instead of just removing them

MightyPork:
fix desktop notifications removing spaces without any replacement, causing words to be joined
this bug was there for several months now, i finally got enough of it lol
screenshot_20171012_202118

added all possible newline types, seems to have no negative effects and better to have it more general just in case imo

### DIFF
--- a/app/javascript/mastodon/actions/notifications.js
+++ b/app/javascript/mastodon/actions/notifications.js
@@ -31,6 +31,7 @@ const fetchRelatedRelationships = (dispatch, notifications) => {
 
 const unescapeHTML = (html) => {
   const wrapper = document.createElement('div');
+  html = html.replace(/<br \/>|<br>|\n/, ' ');
   wrapper.innerHTML = html;
   return wrapper.textContent;
 };


Gargron:
I'm not sure if desktop notifications honour `\n` newlines or not, but if they do, wouldn't it be better to replace br with those

MightyPork:
I think it supports \n, but it doesn't seem to be very reliable - the w3 example page uses " \n"`, looks like for compatibility.

this has advantages:
- multiline toots won't show ridiculously huge notifications
- it's easy (textContent seems to discard newlines, so some workaround would be needed)

sorin-davidoi:
This also affects push notifications (#4860) - the same fix may be required, but in the backend (I think [this line](https://github.com/tootsuite/mastodon/blob/master/app/serializers/web/notification_serializer.rb#L16) is responsible). @MightyPork do you think you can address this as well?

MightyPork:
@sorin-davidoi I don't know ruby, so it'd be something horrible glued together from stackoverflow, better if we merge this first and you can add that later i think

sorin-davidoi:
> I don't know ruby, so it'd be something horrible glued together from stackoverflow

That would not be something bad, as that is how the push notifications were implemented 😄

Gargron (approved):

PR #5363, opened 2017-10-12
Retoot count increases without reason

### LINKED ISSUE #4916: Retoot count increases without reason
I recently made a toot with one account, and retooted it from another, and was surprised to find it indicated 2 retoots already.  I checked, but mine was the only retoot.  A few hours later, that same post has 3 retoots, but still only shows one when the details are checked.

Here's the post in question:
https://nfg.zone/web/statuses/203818

Running 1.6.0

* * * *

cacheflow:
-The store_uri method for Statuses was being called on after_create and causing reblogs to be incremented twice.
-This calls it when the transaction is finished by using after_create_commit.
-Fixes #4916.

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -134,7 +134,7 @@ class Status < ApplicationRecord
     CustomEmoji.from_text([spoiler_text, text].join(' '), account.domain)
   end
 
-  after_create :store_uri, if: :local?
+  after_create_commit :store_uri, if: :local?
 
   before_validation :prepare_contents, if: :local?
   before_validation :set_reblog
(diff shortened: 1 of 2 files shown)


brah:
wow thanks! @cacheflow great work bud!

brah:
looking small!

brah:
stay small @cacheflow !

krainboltgreene (approved):

krainboltgreene:
Great catch.

brah:
please reverse this. this exposes a very bad bug which my machine crash.

I'm using windows 95, v1.9.47 on a chrome browser to reproduce

PR #5384, opened 2017-10-13
Update French translation

yannicka:
A small update of the french translation.


app/javascript/mastodon/locales/fr.json
```diff
@@ -145,7 +145,7 @@
   "onboarding.page_six.apps_available": "De nombreuses {apps} sont disponibles pour iOS, Android et autres. Et maintenant… Bon appouétit !",
   "onboarding.page_six.github": "Mastodon est un logiciel libre, gratuit et open-source. Vous pouvez rapporter des bogues, suggérer des fonctionnalités, ou contribuer à son développement sur {github}.",
   "onboarding.page_six.guidelines": "règles de la communauté",
-  "onboarding.page_six.read_guidelines": "S’il vous plaît, n’oubliez pas de lire les {guidelines} !",
+  "onboarding.page_six.read_guidelines": "S’il vous plait, n’oubliez pas de lire les {guidelines} !",
```
trebmuh: Both are fine, "s'il vous plait" or "s'il vous plaît". No needs to change here, unless you're looking towards consistency across the whole translations. If it would be the case, please let us know.

config/locales/fr.yml
```diff
@@ -513,7 +513,7 @@ fr:
       public: Public
       public_long: Tout le monde peut voir vos statuts
       unlisted: Public sans être affiché sur le fil public
-      unlisted_long: Tout le monde peut voir vos statuts mais ils ne seront pas sur listés sur les fils publics
+      unlisted_long: Tout le monde peut voir vos statuts, mais ils ne seront pas sur listés sur les fils publics
```
trebmuh: It's fine with or without the comma.

trebmuh (commented):
Good catches. I'm not sure about the 2 notes I left in the commit. Is there any reason to change those which are already fine?

app/javascript/mastodon/locales/fr.json
yannicka: It's just to follow the *[Rectifications orthographiques du français en 1990](https://fr.wikipedia.org/wiki/Rectifications_orthographiques_du_fran%C3%A7ais_en_1990)* where the circumflex accent is removed on a lot of words (even if both forms are accepted).

config/locales/fr.yml
yannicka: This is to highlight the two proposals and mark the pause.

app/javascript/mastodon/locales/fr.json
trebmuh: The 1990 dictionary is "proposing" an new form for this (and other) word, not "imposing". Looking quickly at the FR translation here, there are other "plaît" but no "plait".

Hence, I'd be inclined to leave "plaît" as it is, with the circumflex accent.

config/locales/fr.yml
trebmuh: There is no need for that from my point of view since this sentence is not difficult to understand as a whole.

I'm fine with or without it, and would recommend "without".

yannicka:
@trebmuh, I've updated as you said.

trebmuh (approved):
👍

yannicka:
I also noted that "emoji" is not standardized between the accented and unaccented version :

There is two occurrences of the unaccented version:

- https://github.com/tootsuite/mastodon/blob/eb96aa86a42b14bcc16512ecdffb2cbfee5e66c4/app/javascript/mastodon/locales/fr.json#L72
- https://github.com/tootsuite/mastodon/blob/eb96aa86a42b14bcc16512ecdffb2cbfee5e66c4/config/locales/fr.yml#L128

The others have the accent.

Should I add the accent on these two occurrences?

Gargron (approved):

trebmuh:
@yannicka : I'd recommend to do it if that's improving consistency across the translations.

PR #5387, opened 2017-10-14
Fix #5351 - Redirection with page params after copy emoji

### LINKED ISSUE #5351: "Copy Emoji" dialog takes user back to first page
When in the "Custom Emoji" section of the admin panel, if you click on "copy emoji" it takes you back to the first page of the list. This makes it really annoying copying a bunch of emojis, since you have to find your place back each time.

* * * *

tkbky:
## What

As reported in #5351, on admin panel, when copy an emoji (non-local emoji) in the *Custom Emoji* section and it's on page > 1, it always redirect back to page 1 regardless of whether it's successful/fail. This change fixes that.

## Why

It's an inconvenience when trying to copy multiple emojis located on page > 1.

## Other possible issues

The same inconvenience happens on the other actions on `CustomEmojisController`, such as `#copy`, `#enable` & `#disable`

### DIFF
--- a/app/controllers/admin/custom_emojis_controller.rb
+++ b/app/controllers/admin/custom_emojis_controller.rb
@@ -31,10 +31,12 @@ module Admin
       emoji = CustomEmoji.new(domain: nil, shortcode: @custom_emoji.shortcode, image: @custom_emoji.image)
 
       if emoji.save
-        redirect_to admin_custom_emojis_path, notice: I18n.t('admin.custom_emojis.copied_msg')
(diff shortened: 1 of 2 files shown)


Gargron:
Have you tested if this works? Because I don't think the copy method receives a `:page` param, you would need to add it to the `link_to` - unless I am mistaken

tkbky:
@Gargron you're right. 😅 Embarrassingly, I miss it when I commit.

Gargron (approved):

PR #5390, opened 2017-10-14
Close connection when succeeded posting

abcang:
Since we do not need to read the body, close the connection without using `flush`.

Related: #5128

Also, to pass the test we need to use a webmock that is compatible with the new http.rb. (https://github.com/bblimke/webmock/pull/721)


Gargron (approved):
🎊

PR #5397, opened 2017-10-15
Fix #5082 - disable retweet link for followers only toot

### LINKED ISSUE #5082: Boost button is visible and clickable in fourth column on followers-only toots
I think that followers-only toots should not have a clickable boost symbol and a number next to it.

I know there's a padlock underneath it but still! If it can't be boosted, you shouldn't be able to click through to see a list of boosters. The boost symbol and the number imply that there's something worth clicking there and it's weird.

screen_shot_2017-09-24_at_18_16_02

(This is low priority, for sure!)

Edit: Pull request here: https://github.com/tootsuite/mastodon/pull/5397

* * * *

tkbky:
## What

Followers-only toots are not boostable/retweetable yet retweet icon and count is shown. This fixes that.

## Why

Make the toot a little less weird 😂 

## Screenshots

screencapture-localhost-3000-web-statuses-98828179277168451-1508028741617
screencapture-localhost-3000-users-admin-updates-5-1508028713673

### DIFF
--- a/app/views/stream_entries/_detailed_status.html.haml
+++ b/app/views/stream_entries/_detailed_status.html.haml
@@ -42,7 +42,6 @@
     - if status.direct_visibility?
       %span<
         = fa_icon('envelope')
-        %span= status.reblogs_count
     - elsif status.private_visibility?
       %span<
         = fa_icon('lock')
--- a/app/views/stream_entries/_detailed_status.html.haml
+++ b/app/views/stream_entries/_detailed_status.html.haml
@@ -39,9 +39,17 @@
       - else
         = link_to status.application.name, status.application.website, class: 'detailed-status__application', target: '_blank', rel: 'noopener'
       ·
-    %span<
-      = fa_icon('retweet')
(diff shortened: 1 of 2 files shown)


app/views/stream_entries/_detailed_status.html.haml
```diff
@@ -39,9 +39,17 @@
       - else
         = link_to status.application.name, status.application.website, class: 'detailed-status__application', target: '_blank', rel: 'noopener'
       ·
-    %span<
-      = fa_icon('retweet')
-      %span= status.reblogs_count
+    - if status.direct_visibility?
+      %span<
+        = fa_icon('envelope')
+        %span= status.reblogs_count
```
Gargron: This is not needed since direct messages cannot be reblogged?

Gargron: This is not needed since direct messages cannot be reblogged?

Gargron (changes requested):

Gargron (approved):

PR #5400, opened 2017-10-15
Add up-to-date korean translation on serverside

voidsatisfaction:
# Why?

Needs for serverside translation for Korean


Gargron:
@Kjwon15 is this ok?

Kjwon15:
Seems good!

Gargron (approved):

PR #5405, opened 2017-10-15
Make hot keys optional, off by default

nolanlawson:
Hotkeys are an awesome feature, but given that:

1. It's a power user feature; most people won't ever use it
2. It carries some perf overhead in terms of bundle size and extra components
3. It's not useful on mobile devices
4. It might confuse users who accidentally hit a key and don't understand why the UI changed

...I would argue that it should be an optional feature, off by default.

This PR adds a new option to the preferences:

screenshot 2017-10-15 12 11 34

This adds a body class called `use-hotkeys` that is used at the same time we load polyfills to conditionally load a chunk called `hotkeys.js`. (Hence it will delay the first load if you have the feature enabled, but after that the JS file should be cached.)

The perf impact is that `common.js` is reduced from 792 kb (791539) to 757 kB (757267) – 34.3 kB smaller. Also there is less work to create the hotkey/mousetrap components, but I'm not sure what the exact perf impact is there.

out

This is a functional change, so I'm happy to discuss whether or not this should be on by default (with a preload to avoid delaying first load), but I've laid out the case for why it should be off by default.

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -40,6 +40,7 @@ class Settings::PreferencesController < ApplicationController
       :setting_delete_modal,
       :setting_auto_play_gif,
       :setting_system_font_ui,
+      :setting_use_hotkeys,
       :setting_noindex,
       :setting_theme,
       notification_emails: %i(follow follow_request reblog favourite mention digest),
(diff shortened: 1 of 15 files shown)


nolanlawson:
This is live on https://horribly.malfunctioning.technology for the moment

Gargron:
I don't agree with this. I don't like adding settings for everything and I am not convinced that this warrants a setting as opposed to being always on.

nolanlawson:
Withdrawing for now, need to demonstrate runtime perf benefit to make it worth it

PR #5409, opened 2017-10-16
Ensure that feed renegeration restores non-zero items

### LINKED ISSUE #5398: Home feed regeneration yields empty timeline sometimes
Some people who came back after >14 days of inactivity have reported seeing an empty timeline.

I have tried reproducing whether regeneration works (and added more tests), but as far as I can see, yes, regeneration happens. So the cause must be that it yields empty results?

My guess is that because it does last-mile filtering, it's possible that all results from the database are filtered out. For example, if all returned results within the limit (that happens to be just 100 items) are:

- replies to people the user is not following
- mention people the user has blocked
- reblogs of people the user has blocked
- etc

It's not impossible that this happens within just 100 items. My ideas for solving this are:

- Increase the limit from 100 to more (400?). This would *reduce* the problem, but not guarantee that it never occurs
- Accept that regenerated feeds should look different to fresh feeds, e.g. by either completely excluding all replies, or not filtering them out
- Move last-mile filtering to the SQL query, making the query slower but more accurate in returning the 100 items
- Instead of fetching 100 items from all accounts the user follows in total, split the worker into one query per account, e.g. 100 from alice, 100 from bob, which would reduce likelihood of all items getting filtered out. Splitting it into multiple workers would parallelize the process, making is potentially faster, though it could also have a negative impact on Sidekiq throughput for users who follow a lot of people

Gargron:
Fix #5398

Ordering the home timeline query by account_id meant that the first
100 items belonged to a single account. There was also no reason to
reverse-iterate over the statuses. Assuming the user accesses the
feed halfway-through, it's better to have recent statuses already
available at the top. Therefore working from newer->older is ideal.

If the algorithm ends up filtering all items out during last-mile
filtering, repeat again a page further. The algorithm terminates
when either at least one item has been added, or if the database
query returns nothing (end of data reached)

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -100,11 +100,24 @@ class FeedManager
   end
 
   def populate_feed(account)
-    prepopulate_limit = FeedManager::MAX_ITEMS / 4
-    statuses = Status.as_home_timeline(account).order(account_id: :desc).limit(prepopulate_limit)
-    statuses.reverse_each do |status|
-      next if filter_from_home?(status, account)
-      add_to_feed(:home, account, status)
+    added  = 0
+    limit  = FeedManager::MAX_ITEMS / 2
+    max_id = nil
+
+    loop do
+      statuses = Status.as_home_timeline(account)
+                       .paginate_by_max_id(limit, max_id)
+
+      break if statuses.empty?
+
+      statuses.each do |status|
+        next if filter_from_home?(status, account)
+        added += 1 if add_to_feed(:home, account, status)
+      end
+
+      break unless added.zero?
+
+      max_id = statuses.last.id
     end
   end
 


Gargron:
Aside: Should I use the same strategy here?

https://github.com/tootsuite/mastodon/blob/894da3dcca781e27ce9c5130f1021526ac8a6887/app/models/feed.rb#L26-L30

To make sure that even on first load of inactive user, feed is not empty?

aschmitz:
I don't have any idea why this was grouped by `account_id` before, but this is clearly a better way of doing it. I don't have numbers for how many times you're likely to filter out *all* potential statuses, but making sure there's at least one seems fine.

In practice, this seems likely to be overkill, but it's in the right direction, so sure, I say go for it. :+1:

In terms of `Feed#from_database`, I'd be inclined to think leaving that alone is fine: it only gets called when the worker is in the process of rebuilding using exactly this code, so an extra loop over it (while usually not actually used?) would only be doing extra work that the worker is already doing. On the off chance that it's empty, the worker will fill in a status eventually, and not block the web worker while it does it. (If you *do* want to use the same code, I'd consider just calling `FeedManager.instance.populate_feed`, to avoid needlessly duplicating the code, but then I don't really see the point of a separate worker.)

akihikodaki:
Sorry for late review, but I think the reverse-iteration made difference in case of reblogs. When adding reblogs, `add_to_feed` checks if there is the reblogged status *before*. That would not work without reverse-iteration.

PR #5413, opened 2017-10-16
Refactor and simplify icon_button.js

nolanlawson:
This is just a minor refactoring, to use `classnames` instead of building an array, and to pull all the props out as `const`s to make it more readable.


ykzts (approved):
good! :smile:

Issue #5379, opened 2017-10-13
Toots disappear, leaving behind gaps in the column


monsterpit-daggertooth:
*Since 2.0.0rc1*:

Sometimes, toots will disappear from a column and leave gaps that grow over time.  Toots around the gap are often repeated.

On occasion, the entire UI will *also* disappear, leaving behind a ``TypeError: this.node is undefined`` in Firefox's console. 

There are more screenshots here:  https://monsterpit.net/@daggertooth/98796561391496220

![2017-10-13 11-37-01](https://user-images.githubusercontent.com/30352165/31556960-9e00c2c6-b00c-11e7-9db6-538b9e91ae26.gif)
![screenshot 155](https://user-images.githubusercontent.com/30352165/31557088-16c382ca-b00d-11e7-9cdc-f7ac64ab059f.png)

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

sorin-davidoi:
What version of Firefox are you using?

monsterpit-daggertooth:
@sorin-davidoi 
I'm on 56.0.1.

mkosler:
Happening on mastodon.social as well
![space-above-local-timeline](https://user-images.githubusercontent.com/1159095/31563908-98a1e46c-b026-11e7-8d09-2c4a3f0a4ef5.PNG)

sorin-davidoi:
Can you confirm that the hidden toots have `opacity: 0` in the dev-tools?
![image](https://user-images.githubusercontent.com/2109702/31565767-5ac3da46-b068-11e7-8eb0-79804987035c.png)

monsterpit-daggertooth:
@sorin-davidoi 

Yes:
![screenshot 158](https://user-images.githubusercontent.com/30352165/31566731-8e01686a-b031-11e7-953e-8eae0957291a.png)

sorin-davidoi:
It is to be expected that toots that are not visible on screen to have `opacity: 0`. It seems that in this case some toots that are visible still have `opacity: 0` set.

I personally use Firefox 58.0a1 and I don't notice this, but it does not seem to be a Firefox-only issue since @mkosler uses another browser (from the screenshot).  @mkosler, can you tell us the browser and which version of it are you using?

mkosler:
[...]

Issue #5377, opened 2017-10-13
Toots float at top of timeline


stefanhayden:
Randomly tweets will float at the top of my federated timeline. The "relative time" is always "now". This example happened on chrome 61.0.3163.100 / windows 10. I do not know how to reproduce atm.

![example](https://user-images.githubusercontent.com/87616/31555075-a2ba6e2e-b00d-11e7-996e-2270b7fda3cf.gif)

* * * *

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

unarist:
Yeah, I also experienced this issue recently. This might be related to toot dupe bug on the timeline, which also happens sometimes on v2.0.0rc.

unarist:
I got some information about my issue, and I think it's related to your one.

![image](https://user-images.githubusercontent.com/705555/31585826-4b6cb828-b203-11e7-9ea0-179f3aa8b9b5.png)

In this case, one status disappears, and same status appears many time below of it.

Statuses in the timeline should have same `aria-setsize` value and it will be updated when timeline being longer. However, bugged statuses have different values for it, and it won't be updated. Looks like those statuses are no longer controlled by React.

So floated statuses in your timeline were in same situation I think.

This might be related to React 16 which we've began to use from v2.0.0, but still I don't know the cause and reproducing steps...

unarist:
Oh, I forgot to paste some screenshot... In this time, those buggy elements didn't appear on React Dev Tools. Therefore I think React forgot about those elements.

PR #5417, opened 2017-10-16
Filter out duplicate IDs in timelines reducer

Gargron:
Possibly the cause of #5379, #5377

### DIFF
--- a/app/javascript/mastodon/reducers/timelines.js
+++ b/app/javascript/mastodon/reducers/timelines.js
@@ -31,10 +31,10 @@ const initialTimeline = ImmutableMap({
 });
 
 const normalizeTimeline = (state, timeline, statuses, next) => {
-  const ids       = ImmutableList(statuses.map(status => status.get('id')));
+  const oldIds    = state.getIn([timeline, 'items'], ImmutableList());
+  const ids       = ImmutableList(statuses.map(status => status.get('id'))).filter(newId => !oldIds.includes(newId));
   const wasLoaded = state.getIn([timeline, 'loaded']);
   const hadNext   = state.getIn([timeline, 'next']);
-  const oldIds    = state.getIn([timeline, 'items'], ImmutableList());
 
   return state.update(timeline, initialTimeline, map => map.withMutations(mMap => {
     mMap.set('loaded', true);
@@ -45,8 +45,8 @@ const normalizeTimeline = (state, timeline, statuses, next) => {
 };
 
 const appendNormalizedTimeline = (state, timeline, statuses, next) => {
-  const ids    = ImmutableList(statuses.map(status => status.get('id')));
   const oldIds = state.getIn([timeline, 'items'], ImmutableList());
+  const ids    = ImmutableList(statuses.map(status => status.get('id'))).filter(newId => !oldIds.includes(newId));
 
   return state.update(timeline, initialTimeline, map => map.withMutations(mMap => {
     mMap.set('isLoading', false);


unarist:
The mentioned issue can be reproduced by these steps:

1. To make reproducing easy, add `sleep 5` or something into timeline API. For example, [between line 8 and 9 of Api::V1::Timelines::PublicController#show](https://github.com/tootsuite/mastodon/blob/master/app/controllers/api/v1/timelines/public_controller.rb#L8-L9).
2. Open local timeline in the right most column in  WebUI
3. Open some status, then back.
3. While fetching local timeline w/ since_id is in progress, post some status.
4. Then you get an warning like this, and the buggy status appears on the timeline.

```
warning.js?19d7:33 Warning: Encountered two children with the same key, `.$98838703291534900`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

unarist:
The mentioned issue can be reproduced by these steps:

1. To make reproducing easy, add `sleep 5` or something into timeline API. For example, [between line 8 and 9 of Api::V1::Timelines::PublicController#show](https://github.com/tootsuite/mastodon/blob/master/app/controllers/api/v1/timelines/public_controller.rb#L8-L9).
2. Open local timeline in the right most column in  WebUI
3. Open some status, then back.
3. While fetching local timeline w/ since_id is in progress, post some status.
4. Then you get an warning like this, and the buggy status appears on the timeline.

```
warning.js?19d7:33 Warning: Encountered two children with the same key, `.$98838703291534900`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

and you can see it won't happen with this fixes.

app/javascript/mastodon/reducers/timelines.js
```diff
@@ -45,8 +45,8 @@ const normalizeTimeline = (state, timeline, statuses, next) => {
 };
 
 const appendNormalizedTimeline = (state, timeline, statuses, next) => {
-  const ids    = ImmutableList(statuses.map(status => status.get('id')));
   const oldIds = state.getIn([timeline, 'items'], ImmutableList());
+  const ids    = ImmutableList(statuses.map(status => status.get('id'))).filter(newId => !oldIds.includes(newId));
```
unarist: Doing it's on here also OK, but probably we can omit this for now because we append items to timeline only by "load more".

app/javascript/mastodon/reducers/timelines.js
```diff
@@ -31,10 +31,10 @@ const initialTimeline = ImmutableMap({
 });
 
 const normalizeTimeline = (state, timeline, statuses, next) => {
-  const ids       = ImmutableList(statuses.map(status => status.get('id')));
+  const oldIds    = state.getIn([timeline, 'items'], ImmutableList());
+  const ids       = ImmutableList(statuses.map(status => status.get('id'))).filter(newId => !oldIds.includes(newId));
```
unarist: I think most cases of this issue is when the status (A) arrives via streaming while timeline refreshing (Bs) due to column changing. So existing status A would be newer than fetched items. However, this implementation places fetched statuses (`ids`) before newer existing status A.

I think it's better if we can remove duplicated statuses from `oldIds` and prefer fetched `ids` (ordered correctly), but not must be I think.

unarist (approved):
Looks good, but I added some comments.

PR #5419, opened 2017-10-16
Keep references to all reblogs of a status on home feed

### LINKED ISSUE #4210: Toots on home column from users I don't follow
I'm seeing Toots on my home column that aren't from users I follow and they aren't boosted toots either.

* * * *

Gargron:
When inserting reblog: Add to set of reblogs of this status on
the feed, if original status was present in the feed, add it to
that set as well.

When removing a reblog: Remove it from that set. Take random
remaining item from the set. If one exists, re-insert it into feed,
otherwise do not re-insert anything.

Fix #4210

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -161,13 +161,19 @@ class FeedManager
   # either action is appropriate.
   def add_to_feed(timeline_type, account, status)
     timeline_key = key(timeline_type, account.id)
-    reblog_key = key(timeline_type, account.id, 'reblogs')
+    reblog_key   = key(timeline_type, account.id, 'reblogs')
 
     if status.reblog?
+      reblog_set_key = key(timeline_type, account.id, "reblogs:#{status.reblog_of_id}")
+
       # If the original status or a reblog of it is within
       # REBLOG_FALLOFF statuses from the top, do not re-insert it into
       # the feed
       rank = redis.zrevrank(timeline_key, status.reblog_of_id)
+
+      redis.sadd(reblog_set_key, status.reblog_of_id) unless rank.nil?
+      redis.sadd(reblog_set_key, status.id)
+
       return false if !rank.nil? && rank < FeedManager::REBLOG_FALLOFF
 
       reblog_rank = redis.zrevrank(reblog_key, status.reblog_of_id)
@@ -188,7 +194,7 @@ class FeedManager
   # do so if appropriate.
   def remove_from_feed(timeline_type, account, status)
     timeline_key = key(timeline_type, account.id)
-    reblog_key = key(timeline_type, account.id, 'reblogs')
+    reblog_key   = key(timeline_type, account.id, 'reblogs')
 
     if status.reblog?
       # 1. If the reblogging status is not in the feed, stop.
@@ -198,11 +204,15 @@ class FeedManager
       # 2. Remove the reblogged status from the `:reblogs` zset.
       redis.zrem(reblog_key, status.reblog_of_id)
 
-      # 3. Add the reblogged status to the feed.
-      # Note that we can't use old score in here
-      # and it must be an ID of corresponding status
-      # because we need to filter timeline by status ID.
-      redis.zadd(timeline_key, status.reblog_of_id, status.reblog_of_id)
+      # 3. Remove reblog from set of this status's reblogs, and
+      # re-insert another reblog or original into the feed if
+      # one remains in the set
+      reblog_set_key = key(timeline_type, account.id, "reblogs:#{status.reblog_of_id}")
+
+      redis.srem(reblog_set_key, status.id)
+      other_reblog = redis.srandmember(reblog_set_key)
+
+      redis.zadd(timeline_key, other_reblog, other_reblog) if other_reblog
 
       # 4. Remove the reblogging status from the feed (as normal)
     end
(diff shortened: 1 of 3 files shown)


app/workers/scheduler/feed_cleanup_scheduler.rb
```diff
@@ -6,16 +6,17 @@ class Scheduler::FeedCleanupScheduler
 
   def perform
     redis.pipelined do
-      inactive_users.pluck(:account_id).each do |account_id|
+      inactive_users.each do |account_id|
         redis.del(FeedManager.instance.key(:home, account_id))
+        redis.del(FeedManager.instance.key(:home, account_id, 'reblogs'))
```
aschmitz: Oops, missed that somehow. Good call.

app/workers/scheduler/feed_cleanup_scheduler.rb
```diff
@@ -6,16 +6,17 @@ class Scheduler::FeedCleanupScheduler
...
       end
     end
   end
 
   private
 
   def inactive_users
-    User.confirmed.inactive
+    @inactive_users ||= User.confirmed.inactive.pluck(:account_id)
```
aschmitz: Style nit that this should maybe become `inactive_user_ids`, but it's only defined in this tiny class, so whatever.

app/lib/feed_manager.rb
```diff
@@ -161,13 +161,19 @@ def filter_from_mentions?(status, receiver_id)
...
 
     if status.reblog?
+      reblog_set_key = key(timeline_type, account.id, "reblogs:#{status.reblog_of_id}")
+
       # If the original status or a reblog of it is within
       # REBLOG_FALLOFF statuses from the top, do not re-insert it into
       # the feed
       rank = redis.zrevrank(timeline_key, status.reblog_of_id)
+
+      redis.sadd(reblog_set_key, status.reblog_of_id) unless rank.nil?
```
aschmitz: Does this get us anything? I think it might be best to only put reblogs in the `reblogs` set: if the original status is deleted, the reblogs should go with it (and we should check for that case in `unpush`), but there's nothing obvious to be gained by tracking the fact that the original status was in the timeline at some point, is there?

app/lib/feed_manager.rb
```diff
@@ -161,13 +161,19 @@ def filter_from_mentions?(status, receiver_id)
...
+      redis.sadd(reblog_set_key, status.id)
```
aschmitz: If you're concerned about Redis usage, we can alternatively do this only if another reblog is already within `REBLOG_FALLOFF`. It requires a bit more logic (to insert the other reblogging status, but only the first time, when creating the set), but doesn't increase Redis usage in the possibly-most-common case of only seeing one reblog.

app/lib/feed_manager.rb
```diff
@@ -188,7 +194,7 @@ def add_to_feed(timeline_type, account, status)
   # do so if appropriate.
   def remove_from_feed(timeline_type, account, status)
     timeline_key = key(timeline_type, account.id)
-    reblog_key = key(timeline_type, account.id, 'reblogs')
+    reblog_key   = key(timeline_type, account.id, 'reblogs')
 
     if status.reblog?
```
aschmitz: We should handle the case where `status.reblog? == false` and unconditionally delete the reblog_set_key, I think? If a status is deleted, we can safely toss all reblogs of it. (I don't actually know what happens right now in the case where we only have a reblog in our feed and the original is deleted, but it might be worth looking into.)

aschmitz (changes requested):
A few comments and suggestions. I haven't gotten a chance to look at the test changes in detail.

app/lib/feed_manager.rb
Gargron: If the final *reblog* is deleted, but the original *was* originally in the feed, we want to re-insert the original - no?

app/lib/feed_manager.rb
Gargron: True

app/lib/feed_manager.rb
Gargron: Mhhh I don't know what you mean

app/workers/scheduler/feed_cleanup_scheduler.rb
Gargron: Yeah that's uh, not substantial.

app/lib/feed_manager.rb
aschmitz: We don't remove the original when we see the reblog (and in fact, we ignore the reblog), so no, we probably don't need to do anything to track the old status, and should move the other `sadd` to below the `return false`.

app/lib/feed_manager.rb
aschmitz: If we want to minimize the stuff we add to Redis, we could:

* Not add to reblog_set_key for the *first* reblog.
* Only add to reblog_set_key for the second-or-later reblog.

(I had thought that we'd have to add the original reblog to reblog_set_key, but I actually don't think we would: if it gets deleted, our first move would be to remove it from reblog_set_key anyway. If we do want to track all of the reblogs for a status that a user could see (which would be nice to expose in the UI, but not what this PR appears to be about), then we'd have to add to that set, but for the limited purposes of this PR, we could get away with not creating the extra set in cases where there was only one user-visible reblog for a given status.)

Gargron:
@aschmitz Can we merge this and you submit a PR with patches you suggested? I'm.. really tired.. 😩

aschmitz:
Sure.

PR #5423, opened 2017-10-16
Fix reduced motion breaking public galleries

nolanlawson:
I'm not sure if this is a very React-y/Redux-y way of solving this problem, but this fixes the issue of galleries being broken on public statuses.

Instead of using Redux's `connect()`, it adds a body class called `reduce-motion` (similar to `system-font`), which we can check once inside a custom Higher Order Component to decide whether to override the `Motion` styles or not.

Benefits:

* works on pages without Redux
* don't have to worry if the component will update whenever anything in the Redux store changes (i.e. no potential perf problems)
* opens possibility to add CSS styles for `reduce-motion` since it's a body class now

Downsides:

* Not very React-y/Redux-y
* We need our HoC to support everything `Motion` supports, although for now that appears to be just `style` and `defaultStyle` so we're okay

### DIFF
--- a/app/javascript/mastodon/features/ui/util/optional_motion.js
+++ b/app/javascript/mastodon/features/ui/util/optional_motion.js
@@ -1,34 +1,56 @@
 // Like react-motion's Motion, but checks to see if the user prefers
 // reduced motion and uses a cross-fade in those cases.
 
+import React from 'react';
 import Motion from 'react-motion/lib/Motion';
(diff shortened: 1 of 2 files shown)


Gargron (approved):

nolanlawson:
Just realized that from a code cleanliness perspective I probably should have put this into the `initial-state` JSON object. I'll make an issue to clean that up.

PR #5428, opened 2017-10-17
Clean up reblog tracking keys, related improvements

aschmitz:
This should make sure we don't leave keys laying around Redis that can't be accessed by anything. Added some tests to that effect, and cleaned up some behavior in weird cases (for example, if there were two reblogs of a status, the first of which was deleted, a third reblog would have been added to the feed: now it isn't, as the second reblog is already there).

This should probably land in the same tag as #5419 (and be deployed by anyone who has it), otherwise we risk leaving data in Redis that will never be removed.

### DIFF
--- a/app/workers/scheduler/feed_cleanup_scheduler.rb
+++ b/app/workers/scheduler/feed_cleanup_scheduler.rb
@@ -5,18 +5,33 @@ class Scheduler::FeedCleanupScheduler
   include Sidekiq::Worker
 
   def perform
+    reblogged_id_sets = {}
+    feedmanager = FeedManager.instance
+
     redis.pipelined do
-      inactive_users.each do |account_id|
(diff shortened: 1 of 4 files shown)


aschmitz:
(CC: @unarist, @Gargron)

app/workers/scheduler/feed_cleanup_scheduler.rb
```diff
@@ -5,18 +5,36 @@ class Scheduler::FeedCleanupScheduler
   include Sidekiq::Worker
 
   def perform
+    reblogged_id_sets = {}
+    feedmanager = FeedManager.instance
```
Gargron: FeedManager -> feed_manager

But since we're in a hurry I'll clean it up later, it's not substantial

Gargron (approved):

PR #5434, opened 2017-10-17
Fix scrolling of dropdown when offscreen

nolanlawson:
This is a very small fix, but it appears to allow the body to scroll for both the status dropdown and the emoji dropdown when they extend offscreen:

screenshot 2017-10-17 08 03 00

screenshot 2017-10-17 08 27 50

Here's [a video](https://gfycat.com/FortunateSlightCub) showing what it looks like in Firefox on macOS.

I tested this in Firefox, Edge, Chrome, and Safari, I tested modals and fullscreen videos, I tested mobile viewport sizes, and everything seems good to me. But I may have missed something, so I would appreciate a thorough review from someone else, since this changes a style on the entire `body` element and so may have some other effect I'm not aware of.

### DIFF
--- a/app/javascript/styles/basics.scss
+++ b/app/javascript/styles/basics.scss
@@ -30,7 +30,7 @@ body {
   }
 
   &.app-body {
-    position: fixed;
+    position: absolute;
     width: 100%;
     height: 100%;
     padding: 0;


Gargron (approved):

nolanlawson:
OK this is live on https://malfunctioning.technology, asking people to help test…

nolanlawson:
I think long-term we should probably do a redesign where the tooltip is either 1) contained to the column, so it scrolls with the column, or 2) automatically adjusts below/above/to the right of the target based on available space, or 3) both. For the time being though this seems like a stopgap solution.

I tested on Edge/Chrome/Firefox on Win10 and it seems same as Mac, mobile browsers appear unaffected since everything is modals instead of dropdowns.

PR #5437, opened 2017-10-17
When status is fetched instead of delivered, do not stream it

Gargron:


unarist (approved):

ThibG:
What was the rationale for this change?

Gargron:
A fetched status is an old status so it's weird if it suddenly appears in the streamed timelines (appears out if order)

ThibG:
Hm ok… it could be inserted in-order, I guess, though. And it shouldn't prevent notifications to fire off.

PR #5440, opened 2017-10-17
Dutch: A few strings I found last minute

ghost:


config/locales/simple_form.nl.yml
```diff
@@ -4,13 +4,17 @@ nl:
     hints:
       defaults:
         avatar: PNG, GIF of JPG. Maximaal 2MB. Wordt teruggeschaald naar 120x120px
+        digest: Wordt na een lange periode van inactiviteit verzonden, met een samenvatting van vermeldingen tijdens je afwezigheid.
         display_name: Maximaal 30 tekens
```
Gargron: Cannot have value + children, remove value

Gargron (changes requested):

config/locales/simple_form.nl.yml
jeroenpraat: I don't understand @Gargron? This comes directly from https://github.com/tootsuite/mastodon/blob/33ea042decc03c494c78a896934455ef79abc16b/config/locales/simple_form.en.yml#L7

The only difference is see the comma and the period.

Gargron: Umm no the comment is about line 8

```yml
key: This is incorrect
  child: foo
  another: bar
```

jeroenpraat: oooooh :-( Sorry, fixing it now

Gargron (approved):

PR #5443, opened 2017-10-18
Run `i18n-tasks checked-normalized` in Travis CI

ykzts:


JeanGauthier:
Maybe you could do your magic with yarn or something like that on i18n files before the final release ;)

Gargron (approved):

Gargron (changes requested):
Actually this would unfairly affect translators without a dev environment..

ykzts:
I think that this is useful for users who do not have a development environment to find wrong locale files on CI.

Gargron (approved):

PR #5449, opened 2017-10-18
Fix pt-BR translation strings related to advanced search.

renatolond:
Some last minute translations I've found after installing rc3 yesterday.


renatolond:
cc @contraexemplo

Gargron (approved):

PR #5451, opened 2017-10-18
Do not auto-insert text links from uploads. Insert ZWS if otherwise empty

Gargron:


Petitsurume:
@Gargron why are you making this change?

Gargron:
@Petitsurume 

- The auto-insert confuses users into believing the links are required for images to show up, which is false
- The links are ugly and distracting
- They are not needed for any purpose as the images are carried over by metadata (if an app does not display the presence of attachments even as a simple list, that's the app's fault)

unarist:
I found that we only include original URL (direct link) into OS/AP payload. So remote servers won't know text_url. If the origin server changes CDN URL, original URL will be dead but text_url lives.

Although this should be fixed by including text_url in those payloads probably.

nolanlawson:
@unarist It seems fairly uncommon that an instance will change their CDN? Also this will only affect boosted toots whose host instance has cleared non-local media, right?

I don't see this as a critical change for 2.0, and there are arguments for keeping the links (e.g. usability of tapping a link to view a full-sized image rather than right-click -> "view image" which can be hard on some browsers), but overall yeah I think it's a lot cleaner without them, and so seems like a sensible change.

unarist:
@nolanlawson I've seen instances which starts without image server, then switch to S3 or something. but yes, it's uncommon.

I think they can remove text_urls from their toot at composing even now, but I don't have objection.

BTW, even if we change to this way in WebUI, still any apps can insert text_url into toots. So replacing to fa-picture also seems good. If we provide text_url to fediverse, they can compare URLs in the toot with remote_url in media_attachment object, and replace it to fa-picture like we did before.

Gargron:
@unarist The text_url solution requires touching AP code and database schemes, which I cannot do right now in this time frame. But this solution is quick and simple.

unarist:
@Gargron Well, I thought it as an idea for future versions.

nightpool:
Currently, there are a couple of places in the web UI that we wouldn't show links for but need to:

 - when your instance is `media_block`ing users that you're following, you might want to know that they've posted a attachment of some kind, and click through to see it. this change will massively impact the current "tacit compromise" around Pawoo, which is that people can still follow artists but all media has to be fetched through Pawoo and isn't locally cached. After making this change, there will be no way to see media from these instances.

 - When the media attachment itself has been lost to time or deleted by your local instance, through tasks such as `rake mastodon:media:remove_remote`. This may not be a problem with the new deleted media attachments thing, but even so it's still a problem for federating to older instances.

 - When the media attachment exists but isn't in a format supported by your instance, such as audio attachment or a different video format. (Glitch is going to be supporting audio-as-audio attachments in the near future, for example)

PR #5461, opened 2017-10-19
Update model annotations to use BIGINT for IDs

danhunsaker:
All the migrations have been updated to use BIGINTs for ID fields in the DB, but ActiveRecord needs to be told to treat those values as BIGINT as well. This PR does that.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -3,7 +3,7 @@
 #
 # Table name: accounts
 #
-#  id                      :integer          not null, primary key
+#  id                      :bigint           not null, primary key
 #  username                :string           default(""), not null
 #  domain                  :string
 #  secret                  :string           default(""), not null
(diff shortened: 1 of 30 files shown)


danhunsaker:
Just a quick bump to get some feedback, at least.

Gargron (approved):

Gargron:
Unfortunately, since this is auto-generated, this will all get overwritten the next time we run db:migrate

danhunsaker:
That shouldn't be an issue, since the autogenerated annotations should then correctly be `bigint` instead of `integer`?

Gargron:
@danhunsaker Well, for whatever reason, the annotate gem picks them up as integer and not bigint. Otherwise this PR wouldn't have had anything to correct either.

danhunsaker:
I was a little confused about why the annotations hadn't been updated already in that case.

So we're looking at a bug in the annotations gem, then?

ykzts:
It is fixed by https://github.com/ctran/annotate_models/pull/515 . But it has not been released yet.

PR #5462, opened 2017-10-19
Use contenthash for ExtractTextWebpackPlugin

akihikodaki:
`[hash]` is not documented.

This bug was found by @abcang .

### DIFF
--- a/config/webpack/shared.js
+++ b/config/webpack/shared.js
@@ -55,7 +55,7 @@ module.exports = {
         resource.request = resource.request.replace(/^history/, 'history/es');
       }
     ),
-    new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
+    new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[contenthash].css' : '[name].css'),
     new ManifestPlugin({
       publicPath: output.publicPath,
       writeToFileEmit: true,


ykzts (approved):

abcang:
`hash` is not documented, but webpack is doing substitution.  I tried css generation three times with a file name including `hash`, `contenthash`, `chunkhash`, and each hash looks like this.

-| contenthash | hash | chunkhash
--|---|---|---
1 | 466b176bf3a2f237544d479450d1ed38 | d6575da4f473e1d4e068 | c115fc6a59e7e6a123dd 
2 | 592b689ab806a459df16c68ab02c61cd | 4babf0fe221113089680 | c115fc6a59e7e6a123dd
3 | 466b176bf3a2f237544d479450d1ed38 | a2b23121145bd8811e11 | c115fc6a59e7e6a123dd

Sometimes `contenthash` is consistent, `hash` is different every time, `chunkhash` is the same every time. `chunkhash` seems to be replaced by webpack like `hash`, and it is also used for JavaScript filename. So, I think `chunkhash` should be used.

abcang:
`contenthash` is replaced by ExtractTextWebpackPlugin, but` hash` and `chunkhash` seem to be replaced by Webpack. Currently Mastodon works correctly with any hash type, but in Pawoo Music it only worked with `chunkhash`.

I tried to generate css of Pawoo Music 3 times with file names including `hash`,` contenthash`, `chunkhash`. And each hash is as follows.

-| contenthash | hash | chunkhash
--|---|---|---
1 | 466b176bf3a2f237544d479450d1ed38 | d6575da4f473e1d4e068 | c115fc6a59e7e6a123dd 
2 | 592b689ab806a459df16c68ab02c61cd | 4babf0fe221113089680 | c115fc6a59e7e6a123dd
3 | 466b176bf3a2f237544d479450d1ed38 | a2b23121145bd8811e11 | c115fc6a59e7e6a123dd

Sometimes `contenthash` is consistent, `hash` is different every time, `chunkhash` is the same every time. So, if we want to change the type of the hash, it may be good to choose `chunkhash`.

akihikodaki:
The use of `chunkhash` is not documented in `extract-text-webpack-plugin`, so, in my opinion, it should not be used unless the upstream (NOT Pawoo Music or other modifications) has problems with the documented alternative `contenthash` in this case.

abcang (approved):

nightpool:
@akihikodaki @abcang Do we know which features may be causing this instability? I know @beatrix-bitrot was having the *opposite* problems occasionally—where the data would be different but `[hash]` wouldn't change. 

@beatrix-bitrot do you remember what the problem was here?

I really don't want to merge something broken under the argument "well it works for us" without understanding the actual limitations.

nolanlawson:
Sounds like we need to do more research on this. There appears to be a lot of confusion over the difference between these three: https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/153

nolanlawson:
Does `chunkhash` change when the contents _actually_ change? I think the biggest danger here is collisions, not inconsistent hashes. If folks are using `cache-control:immutable` and the same hashes are being generated for different content, then that could break things because old content could be loaded.

akihikodaki:
@nightpool
> Do we know which features may be causing this instability? I know bea was having the opposite problems occasionally over at glitch-soc—where the data would be different but [hash] wouldn't change.

Honestly no. Pawoo Music is experiencing inconsistent hash, but it is probably because of multiple CSS modules while the upstream has only one unified module with SCSS import. (@abcang found the problem not occurs on the upstream even if he added a module, so we need more investigation. Maybe the content or number of modules matters, or it does not happen on master. Note that Pawoo Music is still on 1.4.7.)

Here I'm suggesting to use `contenthash` just because it is a documented alternative. That makes sure it would work even in future, and we can report problems such as one glitch-soc is experiencing to the upstream of extract-text-webpack-plugin:

> I know bea was having the opposite problems occasionally over at glitch-soc—where the data would be different but `[hash]` wouldn't change.

abcang:
I'm sorry. `chunkhash` does not seem to update hash if it only changes css. So it should not be used as css filename.

nolanlawson:
I think I found a comment that explains what's going on: https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/153#issuecomment-225017219

The linked blog post says:

> Due to an issue in Webpack, this method of generating hashes still isn’t deterministic. To ensure hashes are generated based on the file contents, use [webpack-md5-hash](https://github.com/erm0l0v/webpack-md5-hash) plugin. Here is an example how to integrate it into your project: https://github.com/okonet/webpack-long-term-cache-demo/pull/3/files

So it sounds like maybe we should try `chunkhash` and this `webpack-md5-hash`? If not, maybe we need to do a diff between files to see what, if anything, is changing. (E.g. do we have a module somewhere that is exporting randomized content?)

akihikodaki:
I identified the problem Pawoo Music is experiencing, which still happened when using `contenthash`.)

tl; dr: the output of extract-text-webpack-plugin is not deterministic when having multiple entry points, so hashes could change by chance in nature. The problem glitch-soc experienced would not happen with `contenthash` so _we should merge this_.

1. `extract-text-webpack-plugin` loads modules in asynchronous way. (https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/8de6558e33487e7606e7cd7cb2adc2cccafef272/src/index.js#L144)
2. `extract-text-webpack-plugin` concatenates modules in the order of loading.
3. `extract-text-webpack-plugin` executes modules in the concatenated order.
4. `extract-text-webpack-plugin` concatenates output with the execution order.

Because of those steps, the order of contents in the output changes by chance. Therefore `contenthash`, which is the hash of the output, _should_ change the value as well, and actually it does. Pawoo Music and instances with similar configuration should avoid the problem by using one precompiled assets throughout their distributed system. (or they would have different files even if they managed to make the path same.)

Fortunately, the characteristics of `contenthash` would help resolve the problem glitch-soc experienced. As I noted above, it is a hash devoted to the output, and assures it always has different hashes when the output is different.

nolanlawson:
Thanks for clarifying!

PR #5474, opened 2017-10-19
Eliminate space around emoji

nullkal:
Example: https://edge.mstdn.jp/@nullkal/98856648358617597

**Before:**
image

**After:**
image

This PR eliminates space around emojis. We often use emojis like this, so I created this patch so that they are connected seamlessly.

### DIFF
--- a/app/javascript/mastodon/features/emoji/emoji.js
+++ b/app/javascript/mastodon/features/emoji/emoji.js
@@ -28,7 +28,7 @@ const emojify = (str, customEmojis = {}) => {
         // if you want additional emoji handler, add statements below which set replacement and return true.
         if (shortname in customEmojis) {
           const filename = allowAnimations ? customEmojis[shortname].url : customEmojis[shortname].static_url;
-          replacement = `<img draggable="false" class="emojione custom-emoji" alt="${shortname}" title="${shortname}" src="${filename}" />`;
+          replacement = `<img draggable="false" class="emojione" alt="${shortname}" title="${shortname}" src="${filename}" />`;
           return true;
         }
         return false;
--- a/app/javascript/mastodon/features/emoji/emoji.js
+++ b/app/javascript/mastodon/features/emoji/emoji.js
@@ -28,7 +28,7 @@ const emojify = (str, customEmojis = {}) => {
         // if you want additional emoji handler, add statements below which set replacement and return true.
         if (shortname in customEmojis) {
           const filename = allowAnimations ? customEmojis[shortname].url : customEmojis[shortname].static_url;
-          replacement = `<img draggable="false" class="emojione" alt="${shortname}" title="${shortname}" src="${filename}" />`;
+          replacement = `<img draggable="false" class="emojione custom-emoji" alt="${shortname}" title="${shortname}" src="${filename}" />`;
           return true;
         }
         return false;
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -520,8 +520,9 @@
   }
 
   .emojione {
-    width: 18px;
-    height: 18px;
+    width: 20px;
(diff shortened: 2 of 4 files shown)


lynlynlynx:
I know its interesting to make looooong fox, but for common use of emojis, top and bottom paddings are needed. 
ws000123
and padding-top for every status__content makes number of toots displayed fewer. Only for connecting emojos!!!😕


However I think, removing left and right paddings is not bad, because a space is automatically inserted.

nullkal:
I think this PR is important because [Slack](https://slack.com/) has the same behavior (There is no space around custom emojis) and people importing custom emojis from Slack would be disappointed to look space around emojis they added.

I made a change not to modify unicode emoji's behavior. Because of that, this PR affects only custom emojis now. We can add spaces by editing image for custom emojis, so I think there is little concern about custom emojis.
 
image
image

And in regard to padding-top, how about decreasing status__action-bar's margin-top from 10px to 5px. This modification sets off the increase of the height.

lynlynlynx:
From a viewpoint of design, I cannot tolerate uneven styles (sizes, margins) between custom emojis and normal emojis. I cannot tolerate making general design worse ONLY for connecting emojos.

General administrators don't add connenting emojos, adding full-size square emojis is even more general. Users can adjust horizontal gaps by themselves, but they cannot adjust vertical gaps.

> And in regard to padding-top, how about decreasing status__action-bar's margin-top from 10px to 5px. This modification sets off the increase of the height.

It's not bad, thanks.

nullkal:
FYI: Slack's screenshot is like this: 
image

Anyway, we need more people's opinions about this, because talk about 'General admins' without actually listen to them is meaningless.

nullkal:
When I took a screenshot of Slack, I noticed that Slack has *no* space around not only custom emojis, but also normal emojis. I think this behavior is better now, and reverted 6a5bdf0 and 54a8c60.

noraworld:
I like this because some users make a big emoji by concatenating them like looooong fox. Another approach I think is better is to emojify with no space, not to change the appearance. For example, `Fox:fox::fox` is converted to `Fox🦊🦊`.

This is also convenient for those who don't prefer a space between a word and an emoji. Japanese has no space unlike English, so I often glue a word and an emoji together, e.g. `かえる🐸`.

To emojify with no space is not easy because a shortcode is closed with two colons, so we discriminate between first colon and last colon. But I have implemented it before in other project. I'll make a pull request if this approach is favored and no one doesn't do.

I'd also like to ask your opinions.

Gargron (approved):

Gargron:
This PR introduced 5px extra space between status text and username which looks very odd. It really stands out when you switch between 2.0.0 and master frontends. Can that detail be fixed @nullkal @lynlynlynx ?

nullkal:
I added the space because emojis is cropped without it. Another solution is `overflow: visible`.

lynlynlynx:
This PR makes emojis as large as they overflow height of letters, so it is impossible to keep exactly the same as before.
I have two suggestions below, please give me your opinion @Gargron .

master | suggestion 1 <br> 2px extra space | suggestion 2 <br> 0px extra space and move emojis 2px  lower
--|--|--
ws000127 | ws000128 | ws000129

This PR reduces the margin between status text and action buttons from 10px to 5px. Should it be also restored? (I think 5px is enough, ummm)

nullkal:
Suggestion: add `overflow: visible` to `.status__content`, `.reply-indicator__content`: 



Is there any reasons to set `overflow: hidden` to it?

lynlynlynx:
`overflow: visible` revokes the margin between status text and username, it's not good.

ws000008

Gargron:
@lynlynlynx Both your suggestions (1, 2) look better than master to me. I think 2 is best. And yes I want the margin restored too...

PR #5479, opened 2017-10-20
Don't remove originals of boosted toots from timeline

### LINKED ISSUE #5097: Boosted toots jump to top of timeline
When I boost a toot far from the top of my home timeline, it immediately disappears and moves to the top of the timeline, making it hard to interact further with the toot (eg favouriting, replying).

* * * *

duxovni:
I believe this fixes #5097 . I'd like to hear from @Gargron what the original rationale was for including this line, though.

### DIFF
--- a/app/javascript/mastodon/reducers/timelines.js
+++ b/app/javascript/mastodon/reducers/timelines.js
@@ -70,7 +70,6 @@ const updateTimeline = (state, timeline, status, references) => {
   return state.update(timeline, initialTimeline, map => map.withMutations(mMap => {
     if (!top) mMap.set('unread', unread + 1);
     if (top && ids.size > 40) newIds = newIds.take(20);
-    if (status.getIn(['reblog', 'id'], null) !== null) newIds = newIds.filterNot(item => references.includes(item));
     mMap.set('items', newIds.unshift(status.get('id')));
   }));
 };
--- a/app/javascript/mastodon/reducers/timelines.js
+++ b/app/javascript/mastodon/reducers/timelines.js
@@ -55,7 +55,7 @@ const appendNormalizedTimeline = (state, timeline, statuses, next) => {
   }));
 };
 
-const updateTimeline = (state, timeline, status, references) => {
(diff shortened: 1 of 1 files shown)


Gargron:
Do people want to see duplicate original+reblog in their home?

nullkal:
Some people toots after boost when they want to talk something related to the boosted toot. 

So, I want to see the same toot whenever some followee boosted.

nolanlawson:
I find it annoying when the toot disappears after I boost it, yeah. I always have to remember to favorite it _before_ boosting, and as @nullkal says you can't reply after boosting without finding the toot all over again. I think maybe in this one case duplicates can be tolerated.

nolanlawson:
OTOH this PR appears to duplicate _all_ reblogs, not just the one reblogged by the user themselves?

sdukhovni:
I think that this PR is a good idea even if we're not dealing with toots the user boosted themselves. If I'm about to click the "favorite" button on a toot, and someone I'm following boosts that toot a second before I click, the toot will disappear from under my mouse and I'll click on some other random toot. If we don't want duplicates, then rather than removing the original we should just not show the boost.

sdukhovni:
But yup, as @nullkal points out, not showing the boost would make "LB" posts more confusing, so I think it's fine to just have duplicates.

sdukhovni:
Also, I believe that we still won't have duplicates in close proximity, because if a toot is already near the top of your timeline then boosts don't seem to get added to the timeline at all.

sdukhovni:
Just checking in, are there any objections to this PR that haven't been addressed?

charlag:
I agree, current behavior is counter-intuitive and I would rather see duplicates or see the original post.

akihikodaki (approved):
I'm for this change. Disappearing toots are too annoying.

akihikodaki:
> Do people want to see duplicate original+reblog in their home?

Reblogs made in the same session do not have to be inserted anyway.

PR #5485, opened 2017-10-20
Feature: Unlisted custom emojis

nullkal:
This PR enables instance admins to create a custom emoji which is invisible in picker.
I think instance admins often want to add the same emoji in different names for convenience (e.g. :saba:, :sabacan:, and :sabakan:), but showing all of them bloats the picker.

### DIFF
--- a/app/controllers/admin/custom_emojis_controller.rb
+++ b/app/controllers/admin/custom_emojis_controller.rb
@@ -22,6 +22,14 @@ module Admin
       end
     end
 
+    def update
+      if @custom_emoji.update(resource_params)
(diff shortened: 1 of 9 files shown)


nolanlawson:
This issue has occurred in toot.cafe as well – e.g. `:js` and `:javascript:`, `:node:` and `:nodejs:`.

nolanlawson:
Do these invisible emoji still show up in autocomplete? It seems yes, but just checking. Otherwise they may not be very discoverable.

Gargron (approved):

Issue #5426, opened 2017-10-16
Unify usage of initial page state


nolanlawson:
(Follow-up to https://github.com/tootsuite/mastodon/pull/5423)

We have lots of initial state in the web app that is unchanged after the page is loaded. Currently we're handling this a few different ways:

- separate JS files (e.g. `locale_*.js`)
- `initial-state` embedded JSON
- body classes (e.g. `system-font`, `reduce-motion`)
- Redux store (e.g. `auto_play_gif`)

I think the locales are fine as-is, and some body classes are unavoidable due to usage from CSS (e.g. `system-font)`, but in general I think we should unify this so that there's a single module that holds our initial static state, derived from the `initial-state` embedded JSON.

My reasoning:

- body classes are not ideal because you have to wait for the `DOMContentLoaded`/`ready` event, meaning you can't directly export from a module and have to create `getFoo()` functions instead (e.g. `optional_motion.js`)
- Redux store is not ideal because these values never change, and so you add overhead in Redux `connect()` HoCs because they update whenever anything in the store changes, which is wasted work. Plus you have to pass props everywhere when it's completely unnecessary to do so (e.g. `auto_play_gif`)
- We can't do inline scripts because of CSP on some instances, so inline JSON is next best thing

I also noticed that we parse our current `inline-state` three times with `JSON.parse()` when it only needs to be parsed once. So I'd like to clean this up and refactor it so it's handled in one place.

nightpool:
>We can't do inline scripts because of CSP on some instances, so inline JSON is next best thing

does nginx have a way to set CSP nonces?

nolanlawson:
It seems [possible but also very tricky](https://scotthelme.co.uk/csp-nonce-support-in-nginx/). Also inline JSON is really not a whole lot worse than inline JS (it's just a bit awkward), so I'm not bothered if we can't do inline scripts.

PR #5501, opened 2017-10-22
Refactor initial state: reduce_motion and auto_play_gif

nolanlawson:
This is the first step to fixing #5426.

I've moved all usage of the `initial-state` element into a single file: `initial_state.js`. I've also pointed all accesses of `auto_play_gif` and `reduce_motion` to it.

This allowed us to remove one unnecessary Redux `connect` component, as well as avoid passing `autoPlayGif` as a prop everywhere. It also avoids calling `JSON.parse()` on `initial-state` 3 times unnecessarily, which is what we were doing before. So this deletes a lot of code.

I've tested auto-playing GIFs in headers, media, and emoji, in the app and on public pages, and it seems to work the same as before. (Side note: avatars in timelines currently don't appear to respect `autoPlayGif:true`.)

Still left to do: `boostModal`, `deleteModal`, `me`, and `unfollowModal`, but I've saved those for later to keep this PR relatively small.

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -27,7 +27,6 @@
     = yield :header_tags
 
   - body_classes ||= @body_classes || ''
-  - body_classes += ' reduce-motion' if current_account&.user&.setting_reduce_motion
   - body_classes += ' system-font' if current_account&.user&.setting_system_font_ui
 
   %body{ class: add_rtl_body_class(body_classes) }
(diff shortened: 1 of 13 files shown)


app/views/layouts/application.html.haml
```diff
@@ -27,7 +27,6 @@
     = yield :header_tags
 
   - body_classes ||= @body_classes || ''
-  - body_classes += ' reduce-motion' if current_account&.user&.setting_reduce_motion
```
nolanlawson: We're not currently using this in the CSS, but we can add it back in later if need be

ykzts (approved):

PR #5502, opened 2017-10-22
Allow users to restrict incoming DMs to follows only

koboldunderlord:
Paul & I put together a changeset that introduces users the ability to prevent DMs from reaching them.

It works similarly to how other notification blocking mechanisms works.


vahnj:
Closing this as I accidentally pull-requested the wrong commit.

PR #5503, opened 2017-10-22
Allow users to restrict incoming DMs to follows only

koboldunderlord:
Paul & I put together a changeset that introduces users the ability to prevent DMs from reaching them.

It works similarly to how other notification blocking mechanisms works.


vahnj:
There are a couple things that need to change in this after reviewing:

- This should allow users to accept DMs from people they've submitted follow requests to
- There are style problems (Currently fixing)
- Tests are needed

PR #5506, opened 2017-10-23
Delay queuing to sidekiq when deleting account

abcang:
When deleting an account, a lot of jobs are added to the queue of sidekiq all at once, so I changed it to add it to the queue little by little. Sidekiq::Client.push_bulk is used directly because sidekiq-bulk does not support time specification.

### DIFF
--- a/app/services/batched_remove_status_service.rb
+++ b/app/services/batched_remove_status_service.rb
@@ -44,7 +44,7 @@ class BatchedRemoveStatusService < BaseService
       batch_salmon_slaps(status) if status.local?
     end
 
-    at = (statuses.size * batch_index).seconds.after.to_i
+    at = ((@stream_entry_batches.size + @salmon_batches.size + @activity_json_batches.size) * batch_index).seconds.after.to_i
     Sidekiq::Client.push_bulk('class' => Pubsubhubbub::RawDistributionWorker, 'args' => @stream_entry_batches, 'at' => at)
     Sidekiq::Client.push_bulk('class' => NotificationWorker, 'args' => @salmon_batches, 'at' => at)
     Sidekiq::Client.push_bulk('class' => ActivityPub::DeliveryWorker, 'args' => @activity_json_batches, 'at' => at)
--- a/app/services/batched_remove_status_service.rb
+++ b/app/services/batched_remove_status_service.rb
@@ -9,7 +9,8 @@ class BatchedRemoveStatusService < BaseService
   # Remove statuses from home feeds
   # Push delete events to streaming API for home feeds and public feeds
   # @param [Status] statuses A preferably batched array of statuses
-  def call(statuses)
+  # @param [Integer] batch_index
+  def call(statuses, batch_index = 1)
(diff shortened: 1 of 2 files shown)


ykzts (approved):

beatrix-bitrot:
THANK YOU!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

app/services/suspend_account_service.rb
```diff
@@ -17,8 +17,8 @@ def purge_user
   end
 
   def purge_content
-    @account.statuses.reorder(nil).find_in_batches do |statuses|
-      BatchedRemoveStatusService.new.call(statuses)
+    @account.statuses.reorder(nil).find_in_batches(batch_size: 100).with_index(1) do |statuses, batch_index|
```
nightpool: probably should be with_index(0), so the first push has no delay? otherwise it will take (# of followers) seconds before they even see the first statuses deleted. 

If it takes someone with 2000 followers 30 minutes before they see the *first* status deleted after trying to delete their account, that's probably a bad idea.

app/services/batched_remove_status_service.rb
```diff
@@ -9,7 +9,8 @@ class BatchedRemoveStatusService < BaseService
   # Remove statuses from home feeds
   # Push delete events to streaming API for home feeds and public feeds
   # @param [Status] statuses A preferably batched array of statuses
-  def call(statuses)
+  # @param [Integer] batch_index
+  def call(statuses, batch_index = 1)
```
nightpool: 0 here as well

nightpool:
ah, mind that. you'd only need 20 followers with distinct domains before activitypub_json_batches approaches size 2000.

nightpool:
I think this is a good idea, but it goes much too far in the wrong direction. To give some concrete examples, the average cybre.space user has a couple hundred followers. I, personally, have around 5.7k statuses, with 700 followers and around 200 distinct domains between them. This means that if I were to delete my account today, I would (conservatively, because this undercounts NotificationWorker) need approximately 300 hours, or 13 days, to delete my account fully. And my account is *small* compared to many (much smaller then the two that inspired #5542, for example)

abcang:
Yes, it takes a lot of time to complete deletion. The data on that instance is deleted first and the delivery to other instances is delayed. I think that it is important to prevent the delay of other delivery (especially new tout delivery) rather than preventing delays in delivering deletion. I think that the method like priority is the wisest, but it seems to be complicated ...

nightpool:
I *understand* how long deletion takes. along with Alda, I had a user on my instance delete their data a couple of days before witches.town did so. They were a larger account then even mine. There deletion took a long time especially because many of the jobs had to be retried due to octodon.social being down intermittently during the deletion period. Saying that it would take over **2 weeks** to process the deletion fully is just straight-up ridiculous though. there has to be a better middle ground

abcang:
Okay, I understand. Close this PR.

nightpool:
@abcang Okay. It might be useful to just add a constant scaling factor as well.... (1 second per 10 jobs, maybe?)

PR #5509, opened 2017-10-23
Fix copying emojos: redirect to the page you were on

barzamin:
Copying emojos in the admin panel used to bump you back to the first page of emojos; this (very small!) change fixes that, so you don't lose your place if you have lots of emojos!


ykzts (approved):

nolanlawson (approved):

HuskyAlvarez (commented):
Erin I told yew a natural liar

PR #5511, opened 2017-10-23
Add DM-only non-follow muting & accept notifications by follow request

koboldunderlord:
This is a continuation of #5503.

It fixes all the Ruby style issues and introduces two new tests, which both test the DM muting feature against no follows as well as tests it against follow requests.

Since a follow request from a recipient indicates an intent to follow, it doesn't make sense to eliminate notifications from that person a recipient is trying to follow.

### DIFF
--- a/app/services/notify_service.rb
+++ b/app/services/notify_service.rb
@@ -18,7 +18,7 @@ class NotifyService < BaseService
   private
 
   def visibility
-    unless @visibility.nil? && (@activity.is_a?(Follow) || @activity.is_a?(FollowRequest))
+    unless !@visibility.nil? || @activity.is_a?(Follow) || @activity.is_a?(FollowRequest)
       status = @activity.is_a?(Status) ? @activity : @activity.status
       @visibility = status.visibility
     end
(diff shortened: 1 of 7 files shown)


beatrix-bitrot:
LGTM but pinging @unarist to double check and approve

app/services/notify_service.rb
```diff
@@ -16,6 +17,14 @@ def call(recipient, activity)
 
   private
 
+  def visibility
+    unless !@visibility.nil? || @activity.is_a?(Follow) || @activity.is_a?(FollowRequest)
+      status = @activity.is_a?(Status) ? @activity : @activity.status
```
unarist: This also blocks favorite notification for DMs you sent, not only mention notification for received DMs. It seems unwanted behavior for this feature.

app/services/notify_service.rb
```diff
@@ -16,6 +17,14 @@ def call(recipient, activity)
 
   private
 
+  def visibility
+    unless !@visibility.nil? || @activity.is_a?(Follow) || @activity.is_a?(FollowRequest)
```
unarist: How about using `return @status_visibility if defined?(@status_visibility)`? Then you can cache `nil` too.

unarist (changes requested):

unarist:
Question: What if A is going to send DM to B when A isn't following B?

Currently, A can send and B will reply...but A can't see it (also B won't know that their post has been ignored). I don't think it's good design. So I think, for example,

* Show error if you're going to send DM to non-following user, or
* Allow replies to your DM even if you haven't followed them (same as Twitter)

As compared to Twitter, this implementation also lacks "DMs are allowed/disallowed" indication and sending it will fail silently. This may work better for spams, but normal users may also confuses.

app/services/notify_service.rb
vahnj: I don't agree that it's unwanted behavior - it stops someone from spamming the favorite button to flood your notifications.

vahnj:
I think these are valid criticisms since these are limitations of the current implementation of this but I also don't think it's something that should block a change like this.

- Settings don't currently federate at all so we'd have to build something that allows certain settings to federate to allow instances to identify whether a user is blocking DMs or not to display a message
- If the user *REALLY* wants to get back to the person blocking DMs, they can change their status to private

app/services/notify_service.rb
vahnj: sounds good, i'll fix that up

vahnj:
Double checking this I think turning on DM responses in a thread started by the muting user is good and something I'll look into

nightpool:
yeah I think the first problem @unarist brings up is probably the more critical one—although it's very similar to, eg, #5535 (where you can DM someone using OStatus and it won't go through at all).

The idea is that Person A, who blocks DMs from non-followers, initiates a DM with B, who doesn't follow them. This should at *least* show a warning or error for person A, who will never be able to see a reply B makes.

It might be better to add an exception so that, for example, person B can *reply* to DMs person A makes (like twitter), but I could see that going either way.

app/services/notify_service.rb
unarist: Hmm, okay. although `Block only Direct Messages from people you don't follow` sounded that it only blocks DM itself (i.e. mentions).

Gargron:
Last update on this was 12 days ago and the status is "changes requested". Any updates?

vahnj:
I haven't had the time/energy to put into fixing this.  Currently my idea is to move this behavior to be integrated with the thread muting feature, such that if you receive a DM from someone you don't follow, it mutes the thread and doesn't send the notification, but if you send a message first, the thread is not muted so you'll still see notifications from them

vahnj:
I'm going to close this request while I work on the requested changes

vahnj:
Leaving a note here for when I come back to work on this:

"Mute Direct Message conversations initiated by people you don't follow" seems like a good display text for the checkbox with the implementation I described

PR #5522, opened 2017-10-25
Add margin to account for Edge disappearing scrollbar

nolanlawson:
Because we use auto-hiding scrollbars in Edge (https://github.com/tootsuite/mastodon/pull/975), it's unfortunately a little difficult to click on the dates in a toot, because the scrollbar might obscure them:

before

(Scrollbars invisible on the left, scrollbars visible on the right.)

I propose adding a padding to account for the space taken up by the scrollbar. So afterwards it looks like this:

after

In Edge 16 (the latest release, Fall Creators Update) the scrollbars are 16px. On older versions  of Edge they're 12px. So on Edge <16 the padding will be slightly too high, but hopefully folks will just upgrade quickly to Edge 16.

This PR won't affect non-Edge browsers because of the `@supports` rule.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -620,6 +620,12 @@
   border-bottom: 1px solid lighten($ui-base-color, 8%);
   cursor: default;
 
+  @supports (-ms-overflow-style: -ms-autohiding-scrollbar) {
+    // Add margin to avoid Edge auto-hiding scrollbar appearing over content.
+    // On Edge 16 this is 16px and Edge <=15 it's 12px, so aim for 16px.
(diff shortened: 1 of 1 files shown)


unarist:
Hmm, background color and horizontal line won't be expanded to column width?

image

although it may need more CSS codes than current one...

nolanlawson:
Crap, I missed that. Maybe we should add the margin lower in the DOM hierarchy instead.

nolanlawson:
OK, fixed. This ensures that the background color for DMs extends all the way to the right:

out

And so does the horizontal line:

out2

All this PR does now is increase the existing padding for statuses by `16px` in Edge.

unarist (approved):

PR #5538, opened 2017-10-27
Optimize FixReblogsInFeeds migration

unarist:
We have changed how we store reblogs in the redis for bigint IDs. This process is done by 1) scan all entries in users feed, and 2) re-store reblogs by 3 write commands.

However, this operation is really slow for large instances. e.g. 1hrs on friends.nico (w/ 50k users). So I have tried below tweaks.

* It checked non-reblogs by `entry[0] == entry[1]`, but this condition won't work because `entry[0]` is String while `entry[1]` is Float. Changing `entry[0].to_i == entry[1]` seems work.
  -> about 4-20x faster (feed with less reblogs will be faster)
* Write operations can be batched by pipeline
  -> about 6x faster
* Wrap operation by Lua script and execute by EVALSHA command. This really reduces packets between Ruby and Redis.
  -> about 3x faster

I've taken Lua script way, though doing other optimizations may be enough.

```
1000 feeds x 400 statuses (25% are reblogs)

current impl:

      user     system      total        real
 25.870000  32.590000  58.460000 (100.097238)

compare fix:

      user     system      total        real
  7.100000   8.030000  15.130000 ( 25.818915)

compare fix + pipeline:

      user     system      total        real
  2.880000   0.090000   2.970000 (  4.355067)

with script:

      user     system      total        real
  0.220000   0.030000   0.250000 (  1.564265)
```

### DIFF
--- a/db/migrate/20170920032311_fix_reblogs_in_feeds.rb
+++ b/db/migrate/20170920032311_fix_reblogs_in_feeds.rb
@@ -3,48 +3,62 @@ class FixReblogsInFeeds < ActiveRecord::Migration[5.1]
     redis = Redis.current
     fm = FeedManager.instance
 
+    # Old scheme:
+    # Each user's feed zset had a series of score:value entries,
+    # where "regular" statuses had the same score and value (their
+    # ID). Reblogs had a score of the reblogging status' ID, and a
+    # value of the reblogged status' ID.
+
+    # New scheme:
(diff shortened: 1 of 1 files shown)


nullkal:
👍 

FixReblogsInFeeds was *really* slow in mstdn.jp, too. It took 3835 seconds in mstdn.jp.

Gargron (approved):

PR #5539, opened 2017-10-27
Resolve remote accounts when mentioned even if they are already known

ClearlyClaire:
If there is one place where the remote account information *needs* to be up-to-date (at least for the public key and protocol information), it's when delivering toots (especially direct messages).

Without this change, Mastodon mentioning an user never triggers account resolving (unless the remote user is not known at all), which is particularly annoying when dealing with OStatus → ActivityPub migration, as Direct Messages are disabled for OStatus.

This change triggers resolving (still capped to one resolving per account per day in `ResolveRemoteAccountService`) when mentioning remote users, which may increase traffic a bit.

### DIFF
--- a/app/services/process_mentions_service.rb
+++ b/app/services/process_mentions_service.rb
@@ -11,15 +11,10 @@ class ProcessMentionsService < BaseService
     return unless status.local?
 
     status.text.scan(Account::MENTION_RE).each do |match|
-      username, domain  = match.first.split('@')
-      mentioned_account = Account.find_remote(username, domain)
-
-      if mentioned_account.nil? && !domain.nil?
-        begin
-          mentioned_account = resolve_remote_account_service.call(match.first.to_s)
-        rescue Goldfinger::Error, HTTP::Error
-          mentioned_account = nil
-        end
+      begin
+        mentioned_account = resolve_remote_account_service.call(match.first.to_s)
+      rescue Goldfinger::Error, HTTP::Error
+        mentioned_account = nil
       end
 
       next if mentioned_account.nil?


ThibG:
Let's not merge this right now, I have something to figure out…

ThibG:
Fixed a blocking issue in #5543, so this pull request is effectively based on that last one.

nightpool:
Adding this traffic to *all* mentions is an incredibly large performance hit. I think it's more critical to fix https://github.com/tootsuite/mastodon/issues/5535 so we don't silently and unexpectedly break direct messages with ostatus users.

if as a stop-gap we want to scope this pull request to only re-fetch for direct/private messages mentioning ostatus users, that would be fine. I think an equally useful stopgap is to *always* deliver to mentioned users regardless of visibility.

nightpool (changes requested):
either scope this so it's only for ostatus users and only on private/DM messages, or open a new pull request to always deliver to explicitly mentioned accounts (modulo gargron's opinion)

ThibG:
It is going to be a performance hit, indeed, but it is still capped at max one resolve per remote account per day, so I do not expect it to have such a huge performance impact.

However, you're right, for the intended purpose (OStatus → ActivityPub switch), it can be restricted to private/direct messages to OStatus users. I'm going to change it accordingly.

nightpool:
Oh, I thought this would have been on the other side of the webfinger_update_due check. In that case I guess it's probably fine?

nightpool (approved):

ThibG:
Thanks! As for #5535 I have a very crude idea in mind: not processing the mention at all if we know we are not going to deliver it. It's not great design, but it's much better than the current situation: at least a user would know that *something*'s up is the mention is not processed (no link, full handle).

PR #5545, opened 2017-10-27
Update remote ActivityPub users when fetching their toots

ClearlyClaire:
On small instances, it is really frequent for remote users toots from remote non-followed users to be fetched (because of a followed user replying or boosting them), but in that case, in the ActivityPub codepath, their profile is *not* updated, which may sometime cause confusion, e.g. when a message references their avatar/name, or suggests reading their bios.

This commit changes this behavior by refreshing profile data (respecting the common 1 refresh per day per account cap) even if the user is known.

### DIFF
--- a/app/services/activitypub/fetch_remote_status_service.rb
+++ b/app/services/activitypub/fetch_remote_status_service.rb
@@ -16,7 +16,7 @@ class ActivityPub::FetchRemoteStatusService < BaseService
     return if actor_id.nil? || !trustworthy_attribution?(@json['id'], actor_id)
 
     actor = ActivityPub::TagManager.instance.uri_to_resource(actor_id, Account)
-    actor = ActivityPub::FetchRemoteAccountService.new.call(actor_id, id: true) if actor.nil?
+    actor = ActivityPub::FetchRemoteAccountService.new.call(actor_id, id: true) if actor.nil? || needs_update(actor)
 
     return if actor.suspended?
 
@@ -43,5 +43,9 @@ class ActivityPub::FetchRemoteStatusService < BaseService
 
   def expected_type?
     %w(Note Article).include? @json['type']
+  end
+
+  def needs_update(actor)
+    actor.possibly_stale?
   end
 end


app/services/activitypub/fetch_remote_status_service.rb
```diff
@@ -44,4 +44,9 @@ def supported_context?
   def expected_type?
     %w(Note Article).include? @json['type']
   end
+
+  def needs_update(actor)
+    # TODO: optimize by ruling out accounts that have local followers
```
ThibG: I think the full account object may actually be present in the fetched json, but I'm not too sure it's mandated by the spec. If it is, it could be considered prefetched, avoiding further queries, and we could completely bypass the daily limit this way.

app/services/activitypub/fetch_remote_status_service.rb
```diff
@@ -16,7 +16,7 @@ def call(uri, id: true, prefetched_body: nil)
     return if actor_id.nil? || !trustworthy_attribution?(@json['id'], actor_id)
 
     actor = ActivityPub::TagManager.instance.uri_to_resource(actor_id, Account)
-    actor = ActivityPub::FetchRemoteAccountService.new.call(actor_id, id: true) if actor.nil?
+    actor = ActivityPub::FetchRemoteAccountService.new.call(actor_id, id: true) if actor.nil? or needs_update(actor)
```
akihikodaki: Mastodon uses `||`

app/services/activitypub/fetch_remote_status_service.rb
akihikodaki: * It is not mandatory.
* You should _always_ fetch the account (or _dereference_ the ID in the language of the spec) to make sure the ID actually points to the account.

akihikodaki (changes requested):

app/services/activitypub/fetch_remote_status_service.rb
ThibG: In that case, the account is already known, and the json is already considered authoritative for the author's toot, so I'm tempted to just use that info instead of making a new request…
But I will just trigger an account refresh for now, it can be optimized later on.

ThibG:
Addressed your remarks. I kind of want to avoid doing unnecessary HTTP requests if possible (and also bypass the timer in that case), but it's ok for now.

akihikodaki (approved):
Actually it is inefficient since it does not even reuse the HTTP connection, but works™.

PR #5548, opened 2017-10-27
Avoid modifying emoji data inline

nolanlawson:
This `splice()` appears to cause some bugs in certain Webpack configurations (e.g. `--display-optimization-bailout`). There's really no need to modify the array in-place; it's safer to use `slice()`.

Note that this PR doesn't change anything in production; the output JS files are exactly the same size. (I checked.) This just ensures we don't shoot ourselves in the foot if Webpack decides to re-run the script a second time for debugging purposes (which appears to be what is happening here).


nolanlawson:
To be clear: the bug is that this error is thrown: https://github.com/tootsuite/mastodon/blob/bf0ee1a25c10105e096d677782d9c0ae3e36f5a5/app/javascript/mastodon/features/emoji/emoji_compressed.js#L69-L71

ykzts (approved):

PR #5549, opened 2017-10-28
Implement registration through secret-key

kyrias:
I'm pretty new to Ruby and Rails, but was interested in this feature so figured that was a good enough reason to dig into it.  I would greatly appreciate any comments!


kyrias:
For some more details, I was thinking about starting out simply, and then see about implementing a fuller version of the keys where you can specify a list of them, and probably with a way to have one-time-use secret keys.

kyrias:
Ping?

Gargron:
Some comments:

Comparing the value of the setting with strings throughout the code in multiple places is not okay, preferably this should be isolated in some single method. For example, maybe implement virtual methods on the Setting class like `Setting#registrations_open?` and `Setting#registrations_closed?`. This is because if anything will change in the future, it's easy to forget some random string referenced in multiple places, whereas a method, if it would be changed, would become an obvious syntax/runtime error.

I would also like to hear some feedback from admins if this is a good long-term method of registration limiting that they'd like to see in Mastodon. @nolanlawson @Aldarone @ykzts @nightpool @beatrix-bitrot

kyrias:
I implemented `Settings#registrations_open?` now, though I'm not sure if `Setting#registrations_closed?` would be useful, though it would be easy to implement anyway.  I'm currently passing the key query parameter to `Settings#registrations_open?` as direct argument since it can't access the params map directly.  Would it make sense to pass the params directly to it instead, and have the function get the proper index?

Appreciate the input!

nightpool:
@Gargron the fact alone that there's no UI to this change means that I don't think it's ready for master. how are admins supposed to find out how to register people? They have to make the URL query itself and know where to use it and all of that. A static secret key is also really easy to leak and doesn't provide much protection in the long run.

nightpool:
Sorry, I thought I left this comment last week when I first saw this pull request, not sure what happened to it.

kyrias:
Well, the description in the admin panel describes how to pass the secret key in the URL, but I could have it more explicit, possibly have it linked explicitly?

And yeah, that's a possible problem right now.  I was thinking that this way it'd be easy to have the secret key in rotation at least, when necessary.  If I could get this land I was planning on looking into adding a one-time key method as well, but that's more work which is why I started with just the static key.

jeroenpraat:
I like the idea of this. It sounds like an invitation code you can give to people. The admin can than choose between 1. a code that can be used once; 2. a code that can be used a number of times; 3. a code that can be used until the admin decides to block it.

Out of experience I know that people using invitation codes to sign up, have a bigger change to stay active.

nightpool:
@jeroenpraat that is not what this pull request implements.

jeroenpraat:
@nightpool Seems like almost the same to me. Only difference is that you have to use an URL for it, instead of entering a code.

Aldarone:
I like the idea but as @nightpool said, a static key with no UI is not gonna be of much use.

I'd be happy with the suggestion of @jeroenpraat but that would require a way to store the token and its associated parameters (remaining usage count before expiration and an optionnal expiration date ?)

akihikodaki:
Generally speaking, passphrase is not really good for security because they are not bound to any identity.
I suggest to use email for invitation. This method is more secure, and easier for invited guests.

PR #5552, opened 2017-10-28
Twidere mention workaround

ClearlyClaire:
This pull request works around a bug in both Twidere (https://github.com/TwidereProject/Twidere-Android/issues/1001) and Tootdon.

Tootdon and Twidere construct `@user@domain` handles from mentions in toots based
solely on the mention text and account URI's domain without performing any
webfinger call or retrieving account info from the Mastodon server.

As a result, when a remote user has `WEB_DOMAIN` ≠ `LOCAL_DOMAIN`, Twidere and
Tootdon will construct the mention as `@user@WEB_DOMAIN`. Now, this will usually
resolve to the correct account (since the recommended configuration is to have
`WEB_DOMAIN` perform webfinger redirections to `LOCAL_DOMAIN`) when processing
mentions, but won't do so when displaying them (as it does not go through the
whole account resolution at that time).

While this is essentially a workaround for a bug in other software, it's a simple workaround whereas fixing Twidere would be more involved. In addition, it also fixes some inconsistency in Mastodon: it is somewhat dubious for Mastodon to manage to resolve the account in one place (`ProcessMentionsService`) and fail in another (status serialization).

Lastly, this PR depends on #5539 but that could be changed.

### DIFF
--- a/app/services/process_mentions_service.rb
+++ b/app/services/process_mentions_service.rb
@@ -10,17 +10,20 @@ class ProcessMentionsService < BaseService
   def call(status)
     return unless status.local?
 
-    status.text.scan(Account::MENTION_RE).each do |match|
+    status.text = status.text.gsub(Account::MENTION_RE) do |match|
       begin
-        mentioned_account = resolve_remote_account_service.call(match.first.to_s)
+        mentioned_account = resolve_remote_account_service.call($1)
       rescue Goldfinger::Error, HTTP::Error
         mentioned_account = nil
       end
 
-      next if mentioned_account.nil? || (mentioned_account.ostatus? && status.stream_entry.hidden?)
+      next match if mentioned_account.nil? || (mentioned_account.ostatus? && status.stream_entry.hidden?)
 
       mentioned_account.mentions.where(status: status).first_or_create(status: status)
+      "@#{mentioned_account.acct}"
     end
+
+    status.save!
 
     status.mentions.includes(:account).each do |mention|
       create_notification(status, mention)
--- a/app/javascript/mastodon/features/compose/util/counter.js
+++ b/app/javascript/mastodon/features/compose/util/counter.js
@@ -5,5 +5,5 @@ const urlPlaceholder = 'xxxxxxxxxxxxxxxxxxxxxxx';
 export function countableText(inputText) {
   return inputText
     .replace(urlRegex, urlPlaceholder)
-    .replace(/(?:^|[^\/\w])@(([a-z0-9_]+)@[a-z0-9\.\-]+[a-z0-9]+)/ig, '@$2');
+    .replace(/(^|[^\/\w])@(([a-z0-9_]+)@[a-z0-9\.\-]+[a-z0-9]+)/ig, '$1@$3');
 };
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -44,7 +44,7 @@
 #
 
 class Account < ApplicationRecord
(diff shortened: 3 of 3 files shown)


app/services/process_mentions_service.rb
```diff
@@ -10,23 +10,21 @@ class ProcessMentionsService < BaseService
...
-
-      if mentioned_account.nil? && !domain.nil?
-        begin
-          mentioned_account = resolve_remote_account_service.call(match.first.to_s)
-        rescue Goldfinger::Error, HTTP::Error
-          mentioned_account = nil
-        end
+    status.text = status.text.gsub(Account::MENTION_RE) do |match|
+      begin
+        mentioned_account = resolve_remote_account_service.call($1)
```
ThibG: Codeclimate complains about $1, but I'm not sure there is a better way to do that in Ruby, at least I have not found one reading the documentation for gsub.

nightpool: what's wrong with `resolve_remote_account_service.call(match.first.to_s)`?

app/services/process_mentions_service.rb
```diff
@@ -10,23 +10,21 @@ class ProcessMentionsService < BaseService
...
+      rescue Goldfinger::Error, HTTP::Error
+        mentioned_account = nil
       end
 
-      next if mentioned_account.nil?
+      next match if mentioned_account.nil?
 
       mentioned_account.mentions.where(status: status).first_or_create(status: status)
+      mentioned_account.domain.nil? ? "@#{mentioned_account.username}" : "@#{mentioned_account.username}@#{mentioned_account.domain}"
```
nightpool: this line should just be "@#{mentioned_account.acct}"

app/services/process_mentions_service.rb
```diff
@@ -10,23 +10,21 @@ class ProcessMentionsService < BaseService
...
     end
 
+    status.save!
```
nightpool: at this point have we already delivered the message to subscribers? I don't know if we're allowed to re-write the message here then—we'd have to pull this code somewhere else.

app/services/process_mentions_service.rb
ThibG: match is the full matched string, not a list of groups :/

nightpool (dismissed):

nightpool:
I'm not sure merging this change is a good idea. While in theory processing webfinger aliases is a useful thing to be doing when @mentioning users, this really needs to be fixed upstream as well—there's no guarantee that users with web_domain != local_domain will have webfinger aliases correctly set up.

app/services/process_mentions_service.rb
ThibG: No we haven't, this is right before delivering it.

nightpool: How do we know that though? I don't know if this is the right thing to do from a code organization perspective. adding a transaction might create a nested transactions error, for example.

ThibG: I understand your concern, and I share it to some extent. However, the serialized status sent over both the WebUI and remote servers through OStatus and ActivityPub already depends on this function, as the serialization depends on the `mentioned_account.mentions` created by this function.

ThibG:
@nightpool fixed the issues you pointed out. I somewhat agree with you, although I also see it as a consistency fix for the following behavior: Mastodon resolves the account, records a “mention” relationship and notify the remote user, but the serialized toot does not contain the html-rendered mention, which breaks further interaction as the WebUI does not pick it up on reply

Gargron (approved):

ThibG:
Rebased on latest master, so it doesn't include the first commit anymore.

PR #5558, opened 2017-10-28
Avoid emojifying on invisible text

### LINKED ISSUE #5554: Emojis in the invisible part of the URL is visible
```
http://example.com/web/timelines/public」text🍊🍊🍊
```

image

* * * *

hcmiya:
Fixes #5554

### DIFF
--- a/app/javascript/mastodon/features/emoji/__tests__/emoji-test.js
+++ b/app/javascript/mastodon/features/emoji/__tests__/emoji-test.js
@@ -57,5 +57,21 @@ describe('emoji', () => {
     it('does an emoji whose filename is irregular', () => {
       expect(emojify('↙️')).toEqual('<img draggable="false" class="emojione" alt="↙️" title=":arrow_lower_left:" src="/emoji/2199.svg" />');
     });
+
+    it('avoid emojifying on invisible text', () => {
+      expect(emojify('<a href="http://example.com/test%F0%9F%98%84"><span class="invisible">http://</span><span class="ellipsis">example.com/te</span><span class="invisible">st😄</span></a>'))
+        .toEqual('<a href="http://example.com/test%F0%9F%98%84"><span class="invisible">http://</span><span class="ellipsis">example.com/te</span><span class="invisible">st😄</span></a>');
+      expect(emojify('<span class="invisible">:luigi:</span>', { ':luigi:': { static_url: 'luigi.exe' } }))
+        .toEqual('<span class="invisible">:luigi:</span>');
+    });
+
+    it('avoid emojifying on invisible text with nested tags', () => {
+      expect(emojify('<span class="invisible">😄<span class="foo">bar</span>😴</span>😇'))
+        .toEqual('<span class="invisible">😄<span class="foo">bar</span>😴</span><img draggable="false" class="emojione" alt="😇" title=":innocent:" src="/emoji/1f607.svg" />');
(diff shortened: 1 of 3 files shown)


app/javascript/mastodon/features/emoji/__tests__/emoji-test.js
```diff
@@ -57,5 +57,10 @@ describe('emoji', () => {
     it('does an emoji whose filename is irregular', () => {
       expect(emojify('↙️')).toEqual('<img draggable="false" class="emojione" alt="↙️" title=":arrow_lower_left:" src="/emoji/2199.svg" />');
     });
+
+    it('avoid emojifying on invisible text', () => {
+      expect(emojify('<a href="http://example.com/test%F0%9F%98%84"><span class="invisible">http://</span><span class="ellipsis">example.com/te</span><span class="invisible">st😄</span></a>'))
+        .not.toEqual('<a href="http://example.com/test%F0%9F%98%84"><span class="invisible">http://</span><span class="ellipsis">example.com/te</span><span class="invisible">st<img draggable="false" class="emojione" alt="😄" title=":smile:" src="/emoji/1f604.svg" /></span></a>');
```
nightpool: this should probably be `.toEqual('<a href="http://example.com/test%F0%9F%98%84"><span class="invisible">http://</span><span class="ellipsis">example.com/te</span><span class="invisible">st😄</span></a>')` instead of `.not.toEqual()`, to be more resilient to changes in the emojify code (reordering the attributes and then regressing would still cause this test to pass)

app/javascript/mastodon/features/emoji/emoji.js
```diff
@@ -36,6 +36,11 @@ const emojify = (str, customEmojis = {}) => {
     } else if (tag >= 0) { // <, &
       rend = str.indexOf('>;'[tag], i + 1) + 1;
       if (!rend) break;
+      if (str.slice(i, rend) === '<span class="invisible">') {
+        // avoid emojifying on invisible text
+        const span = '</span>';
```
nightpool: this wont work with nested spans correctly. not sure how much we care about that?

nightpool (changes requested):

nightpool:
first one is a requested change. not sure how much we care about the second

app/javascript/mastodon/features/emoji/emoji.js
hcmiya: Yes, I understand that It's safer to handle by whole anchor text. But with that method, the emoji domain will not be comverted...

hcmiya: alt way (visible parts also doesn't be converted)
```
if (str.slice(i, rend).startsWith('<a ')) {
  const a = '</a>';
  rend = str.indexOf(a, rend) + a.length;
}
```

Gargron (approved):

hcmiya:
I made another solution in #5567. Which do you think is better?

hcmiya:
I close this PR due to conflict with #5570.

nightpool:
What does this have to do with 5570? They're two different bugs.

akihikodaki:
Reopen this, please?

hcmiya:
I reopened from request and rewrited for handling nested tags.

Gargron (approved):

nightpool (approved):

PR #5560, opened 2017-10-29
Changing repository URL on about page

kuropen:
Changed to point here.


kuropen:
Sorry, it is mistake.

PR #5561, opened 2017-10-29
Change 2FA input to use "number" type

### LINKED ISSUE #5527: Two factor code input should use keypad on mobile devices
When entering the two-factor code on my phone, it's always a hassle to switch to the number part of the keyboard. Setting the input type / constraints right should allow the browser to show a number pad automatically.

According to [this stackoverflow answer](https://stackoverflow.com/questions/6178556/phone-numeric-keyboard-for-text-input#answer-31619311), setting `<input type="number" pattern="[0-9]*" inputmode="numeric">` should work on iOS and Android, but it's from 2015, so there might be a better way.

* * * *

nolanlawson:
Fixes #5527

Here's what mobile Chrome looks like:

chrome

And it also works in mobile Firefox:

firefox

It doesn't seem to work on iOS for some reason:

screenshot 2017 10 29 11 14 13

On some desktop browsers (Chrome, Firefox, Safari – see below) there is now a +/- toggle to increase or decrease the number, but I don't think it's a big deal, and it's nice to have the numeric keypad on mobile. In most cases people will just be copy-pasting this anyway.

screens

### DIFF
--- a/app/views/auth/sessions/two_factor.html.haml
+++ b/app/views/auth/sessions/two_factor.html.haml
@@ -2,7 +2,7 @@
   = t('auth.login')
 
 = simple_form_for(resource, as: resource_name, url: session_path(resource_name), method: :post) do |f|
-  = f.input :otp_attempt, type: :number, placeholder: t('simple_form.labels.defaults.otp_attempt'), input_html: { 'aria-label' => t('simple_form.labels.defaults.otp_attempt'), :autocomplete => 'off' }, required: true, autofocus: true, hint: t('simple_form.hints.sessions.otp')
+  = f.input :otp_attempt, type: :number, placeholder: t('simple_form.labels.defaults.otp_attempt'), input_html: { 'aria-label' => t('simple_form.labels.defaults.otp_attempt'), :autocomplete => 'off', type: 'number' }, required: true, autofocus: true, hint: t('simple_form.hints.sessions.otp')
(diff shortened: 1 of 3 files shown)


Gargron:
Can we remove the +/- toggle with CSS? -webkit-appearance?

lynlynlynx:
Recovery codes contain letters?

nolanlawson:
Glancing through the [HOTP](https://tools.ietf.org/html/rfc4226) and [TOTP](https://tools.ietf.org/html/rfc6238) specs it seems the value should always be a digit. At least I see it referred to as "digits" in the spec.

As for removing the +/-, if we can only do it on desktop Chrome/Safari then that's a bit unfortunate because Firefox won't get it. Edge doesn't have the +/- at all FWIW.

Given that 1) it causes ugliness on desktop, and 2) most people will be copy-pasting the value anyway, and 3) I'm not 100% sure looking at the spec that only numeric codes are allowed, I'm tempted to close this PR.

kyrias:
The HOTP spec does list using alphanumeric values as an extension.  Never seen it in practice though.

nightpool:
@lynlynlynx is correct—the generated 2fa recovery codes contain letters.

nolanlawson:
OK, this seems like more trouble than it's worth, and could break in the future. Closing for now.

PR #5505, opened 2017-10-23
Refactor initial state: unfollow_modal, boost_modal, delete_modal

nolanlawson:
Part 2 of #5501. I made a separate PR so it's a bit easier to digest. :smiley:

This moves `unfollow_modal`, `boost_modal`, and `delete_modal` to the `initial_state` store.

PR #5563, opened 2017-10-29
Refactor initial state: "me"

### LINKED ISSUE #5426: Unify usage of initial page state

nolanlawson:
The meta `me` property (i.e. the string ID of the current user) never changes without a page reload. So we can just grab it statically from the `initial_state` instead of passing it around everywhere in the props.

This is a follow-up to #5501 and #5505 and fixes #5426.

### DIFF
--- a/app/javascript/mastodon/actions/pin_statuses.js
+++ b/app/javascript/mastodon/actions/pin_statuses.js
@@ -4,12 +4,13 @@ export const PINNED_STATUSES_FETCH_REQUEST = 'PINNED_STATUSES_FETCH_REQUEST';
 export const PINNED_STATUSES_FETCH_SUCCESS = 'PINNED_STATUSES_FETCH_SUCCESS';
 export const PINNED_STATUSES_FETCH_FAIL = 'PINNED_STATUSES_FETCH_FAIL';
 
+import { me } from '../initial_state';
+
 export function fetchPinnedStatuses() {
   return (dispatch, getState) => {
     dispatch(fetchPinnedStatusesRequest());
 
-    const accountId = getState().getIn(['meta', 'me']);
-    api(getState).get(`/api/v1/accounts/${accountId}/statuses`, { params: { pinned: true } }).then(response => {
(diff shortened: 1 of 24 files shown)


ykzts (approved):

nightpool (commented):
app/javascript/mastodon/reducers/meta.js seems like it needs to be updated?

nolanlawson:
Yeah I could remove `me` from there as well. Honestly most of `meta` probably needs to go away, since it's almost all static and therefore doesn't need to be managed in Redux.

nolanlawson:
OK, one more LoC removed 😄

nightpool:
LGTM

nightpool (approved):

Issue #5553, opened 2017-10-28
Treating URLs as fixed length allows users to post super long status


unarist:
We have 500 characters length validation for statuses, but users can avoid this by putting any number of characters in the URL. Then this may damage clients or remote instances.

I think we should have hard limit for status length, which won't be affected by URL handling.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nightpool:
I mean, currently remote instances can specify arbitrarily long statuses? so this is already something we need to handle and I don't see why we should avoid it. Furthermore, URLs are truncated to their host portion during display, so I don't see this being a major problem for the front-end.

unarist:
Ah yes, remote instance can do this already.

>  URLs are truncated to their host portion during display, so I don't see this being a major problem for the front-end.

Actually it's in redux store and DOM element, even if it's not displayed. URL with one million characters makes WebUI significantly slow, for example.

Also, the original issue reported by @imncls@knzk.me was a URL with lot of emojis. I opened this issue as it's common problem on URL handling, but I've realized that "lot of emojis" was the point. We replace emojis to Twemoji in the front-end, so performance impact is bigger than normal characters.

Gargron:
What options do we have?

trwnh:
Realistically?

1) Have URLs count against the character limit (since most instances default to 500, this can be seen as "enough" for all but the longest URLs)
2) Optional URL shortening module (autoshort of every single URL is bad -- any implementation should be up to the user to use)
3) Do nothing, continue suffering from performance drops / etc related to long URLs

nightpool:
.... or just add a 25kb hard limit on local statuses the same way we do for
remote activities (I dont know what the actual number is)

Gargron:
@nightpool We have a 25kb limit for statuses? :thinking: That's news to me

nightpool:
Huh, nevermind, I swore the last time I was looking at the signature verification code it had somehting like that in there.....

vmstan:
URL are limited via #17908

PR #5567, opened 2017-10-30
don't display any descendants of .invisible

hcmiya:
This is an alternative to #5558 for resolving #5553, #5554.
#5558 will break when `span` is nested. Also, if images is inserted into the `.invisible` by another bug, It will be displayed.
Using `!important` rule is generally considered a bad practice. However, it's rare to apply other styles to hidden elements, so this is OK, probably...

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -246,11 +246,9 @@
   position: absolute;
 
   img, svg {
-    display: inline-block !important;
     margin: 0 !important;
     border: 0 !important;
     padding: 0 !important;
-    font-size: 0 !important;
     width: 0 !important;
     height: 0 !important;
   }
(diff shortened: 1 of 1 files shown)


nightpool:
* selector is really bad for performance. Can we use a more specific selector instead?

hcmiya:
Like img or svg?

kyrias:
While it's not optimal when there are more specific options, the fear of `*` is mostly a leftover from the IE6 era, and it's been shown to be pretty insignificant performance-wise for browsers newer than that,[¹](http://www.stevesouders.com/blog/2009/03/10/performance-impact-of-css-selectors/)[²](https://www.telerik.com/blogs/css-tip-star-selector-not-that-bad), and it's not really "really bad for performance" anymore.

nightpool:
@kyrias I'm more concerned about unwanted side-effects rather then performance.

Gargron (approved):

PR #5570, opened 2017-10-30
 Do not emojify short code preceeded or succeeded by [:space:]

### LINKED ISSUE #5528: Shortcode without breaking white-space gets emojified
https://edge.mstdn.jp/@fn_aki/98889258670037012
> :ahegao::ahegao:

This sequence is not emojified, which is an expected behavior. However,

https://edge.mstdn.jp/@fn_aki/98888356925691971
> :ahegao: :ahegao::ahegao:

Only the first `:ahegao:` should be emojified, but the latter two `:ahegao:` are actually emojified.

UPDATE: this behavior was found by @manohito.

* * * *

akihikodaki:
Resolves #5528. This involves a breaking change to replace `[^[:alnum:]]` with `[[:space:]]` in the regular expression used on the server side.

### DIFF
--- a/app/javascript/mastodon/features/emoji/__tests__/emoji-test.js
+++ b/app/javascript/mastodon/features/emoji/__tests__/emoji-test.js
@@ -45,6 +45,30 @@ describe('emoji', () => {
       expect(emojify('<p data-foo="\uD83D\uDC69\uD83D\uDC69\uD83D\uDC66"></p>')).toEqual('<p data-foo="\uD83D\uDC69\uD83D\uDC69\uD83D\uDC66"></p>');
     });
 
+    it('does custom emoji without surrounding characters', () => {
+      expect(emojify(':emojo:', { ':emojo:': { static_url: 'emojo.png' } })).toEqual('<img draggable="false" class="emojione" alt=":emojo:" title=":emojo:" src="emojo.png" />');
+    });
+
+    it('does custom emoji with surrounding HTML tags', () => {
+      expect(emojify('<p>:emojo:</p>', { ':emojo:': { static_url: 'emojo.png' } })).toEqual('<p><img draggable="false" class="emojione" alt=":emojo:" title=":emojo:" src="emojo.png" /></p>');
+    });
+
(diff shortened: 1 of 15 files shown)


akihikodaki:
Labeling work-in-progress; a test points out it does not handle the case when it is in an HTML tag.

noraworld:
The best specification I think is also to emojify consecutive short codes. For example, `:+1::+1:` should be emojified to `👍👍`, and `hello:blush:` should be replaced with `hello😊`.

These two examples are not emojified now because of no spaces between short code and letter / short code. But they should be emojified because these mistakes (that short code is not emojified) often happen I think.

akihikodaki:
@noraworld I heard they are not emojified because of IPv6 addresses. Of course, there should be options other than completely disabling emojifications without spaces, but we need to look back the previous discussions and show counterarguments to withdraw the current design. Unfortunately I do not know such discussions other Mastodon developers should have done before.

akihikodaki:
4766b6c1b2d24843f5c7417aee5b74790450d8a5: rebased to 2bcc81700c586a543e849e222c640df89da7fcb0, and now it deals with HTML tags well. Unlabelling _work in progress_.

noraworld:
@akihikodaki I'm sorry I have no idea why they are not emojified with IPv6 addresses. Could you tell me the reason in detail?

akihikodaki:
@noraworld I mean IPv6 addresses can (unexpectedly) be emojified. For example, an address may contain `:ca::ca:`. They should be shown as-is, but they can be interpreted as two Canada flags if `:ca:` is assigned to Canada flag.

noraworld:
@akihikodaki I see. According to your description, I suspect it's a bug, so should be fixed I think. But the specification that the consecutive short codes are emojified is also important for me. I believe we need to discuss this specification.

akihikodaki:
https://github.com/tootsuite/mastodon/pull/5570#event-1356873836: The server-side implementation is still incomplete.

akihikodaki:
https://github.com/tootsuite/mastodon/pull/5570#issuecomment-346790647: addressed

PR #5582, opened 2017-11-01
Remove babel-plugin-react-transform

ykzts:
This package is never used.


Gargron:
Are you sure it's not required for React JSX?

ykzts:
Yes, babel-plugin-transform-react-jsx (not babel-plugin-react-transform) is used for React JSX.

Gargron (approved):

PR #5599, opened 2017-11-05
Retry thread resolving

ClearlyClaire:
Thread resolving is one of the few tasks that isn't retried on failure.
One common cause for failure of this task is a well-connected user replying to
a toot from a little-connected user on a small instance: the small instance
will get many requests at once, and will often fail to answer requests within
the 10 seconds timeout used by Mastodon.

This changes makes the ThreadResolveWorker retry a few times, with a
rapidly-increasing time before retries and large random contribution in order
to spread the load over time.

### DIFF
--- a/app/workers/thread_resolve_worker.rb
+++ b/app/workers/thread_resolve_worker.rb
@@ -3,7 +3,11 @@
 class ThreadResolveWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'pull', retry: false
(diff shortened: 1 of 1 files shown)


nightpool (approved):
LGTM w/ codeclimate fixes

Gargron:
Yes, now that we have snowflake IDs, fetching threads even much later is okay. Good change.

ThibG:
Rebased and fixed style issue.

PR #5604, opened 2017-11-06
Unify file upload to using fog

ykzts:
Simplify paperclip config.

ref https://github.com/tootsuite/mastodon/pull/2322#issuecomment-297076488


nightpool:
What's fog, and why should we use it over the working config that's been
tested?

Gargron:
@nightpool fog is a backend-agnostic driver for paperclip. Early on paperclip had built-in configuration for s3, but they switched to using fog drivers, which there is a variety of for various backends like s3, openstack, google cloud, etc. But I haven't used fog myself yet.

Gargron (approved):

PR #5608, opened 2017-11-07
Define CacheRemovable#remove_blocking_cache

ysksn:
There were three too similar methods.

- `AccountDomainBlock#remove_blocking_cache`
- `Block#remove_blocking_cache`
- `Mute#remove_blocking_cache`

### DIFF
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -18,13 +18,9 @@ class Block < ApplicationRecord
 
   validates :account_id, uniqueness: { scope: :target_account_id }
 
-  after_create  :remove_blocking_cache
-  after_destroy :remove_blocking_cache
-
-  private
-
-  def remove_blocking_cache
-    Rails.cache.delete("exclude_account_ids_for:#{account_id}")
-    Rails.cache.delete("exclude_account_ids_for:#{target_account_id}")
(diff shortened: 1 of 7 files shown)


app/models/block.rb
```diff
@@ -12,19 +12,17 @@
...
 
-  after_create  :remove_blocking_cache
-  after_destroy :remove_blocking_cache
-
-  private
-
-  def remove_blocking_cache
-    Rails.cache.delete("exclude_account_ids_for:#{account_id}")
-    Rails.cache.delete("exclude_account_ids_for:#{target_account_id}")
+  REMOVE_BLOCKING_CACHE = -> do
```
Gargron: I don't understand why a constant proc is better than a method definition? I agree with abstracting to concern module, but I think the methods should remain methods.

Gargron (changes requested):

app/models/block.rb
ysksn: Using these constants instead of defining similar methods are not sophisticated enough, but better at least in terms of removing the redundant tests.

ysksn: agreed :+1:

PR #5611, opened 2017-11-07
Make fullscreen video in detailed status plays in fullscreen (Partly Fix #5160)

### LINKED ISSUE #5160: Fullscreen video on desktop Safari doesn't play fullscreen
In desktop Safari 11.0 on Sierra, when you click the fullscreen button, the video doesn't go fullscreen. Instead it seems to get stuck in the column.

Seen on 1df453aff6f46e1fddc665bffa3f68312f672956. Related: #5159.

out2

* * * *

nullkal:


app/javascript/mastodon/features/status/index.js
```diff
@@ -65,10 +67,19 @@ export default class Status extends ImmutablePureComponent {
     intl: PropTypes.object.isRequired,
   };
 
+  constructor (props) {
```
abcang: `context` should also be passed as an argument.

```js
constructor (props, context) {
  super(props, context);
}
```

Or, it is recommended to assign state without using constructor.

```js
export default class Status extends ImmutablePureComponent {
  static propTypes = {};

  state = {};

  componentWillMount () {}
}
```

app/javascript/mastodon/features/status/index.js
```diff
@@ -65,10 +67,19 @@ export default class Status extends ImmutablePureComponent {
...
+    super(props);
+    this.state = {};
```
abcang: I thought it was better to set it to the initial value of `fullscreen` of state.

Gargron (approved):

PR #5613, opened 2017-11-07
Hide disabled custom emojis from emoji picker and emoji auto suggestions.

kedamaDQ:
Make the same behavior as /api/v1/custom_emojis.


Gargron (approved):

kedamaDQ:
Thanks for merging!!

Issue #3515, opened 2017-06-02
Redirect to unexpected page


kanata2:
In some cases, when you click "Back" button rendered by "column_back_button.js", it redirect to unexpected URL such as another web page, your startup page (which you saw just before).
I think that it would be better to redirect to timeline or getting-started page instead of the previous page in those cases.

---
Steps to reproduce bug:

- click on a toot or user.
- copy URL (like ~/web/accounts/123456, ~/web/statuses/123456)
- open the new tab and copy this URL
- in the user page or toot page, click "Back" button

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
  - this is probably one case of this issue: https://github.com/tootsuite/mastodon/issues/2329

MightyPork:
Running into this bug daily.

I have a bookmarked layout with the local pinned and federated in the last column.

When I want to see the start page, I automatically go for the back button, but that closes mastodon and opens my g-mail or whatever I had open before. Quite frustrating tbh.

Gargron:
Duplicate of #946

Issue #946, opened 2017-04-05
Remain in app UI when clicking back button on a column


jerome-tgl:
1. I remote-followed someone in another instance (from [this URL](https://octodon.social/authorize_follow?acct=BrendanEich%40mastodon.social))
3. The page directsme to [this URL](https://octodon.social/web/accounts/2452), which shows the Mastodon main UI with the user's profile in the right-column
3. I click the `< back` button in this column

What happens: I'm redirected to the [previous URL](https://octodon.social/authorize_follow?acct=BrendanEich%40mastodon.social), as if I clicked the back button of my Web browser.

What I expected: that the user's profile would be close and the right-column would display the getting-started guide, or the local/public timeline.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

michaeljdeeb:
I wanted to start a conversation around solving this specifically for the remote follow use-case as #1522 is more complicated given that it would be a major privacy concern if your browser exposed your search history. Also, the remote follow use-case is more likely and can be jarring to new users.

This behavior does not occur on [community_timeline](https://github.com/tootsuite/mastodon/blob/master/app/assets/javascripts/components/features/community_timeline/index.jsx#L89) or [public_timeline](https://github.com/tootsuite/mastodon/blob/a0ed88a99b736cc786b048126580a7e73c68b07e/app/assets/javascripts/components/features/public_timeline/index.jsx#L89) components because they use [column_back_button_slim](https://github.com/tootsuite/mastodon/blob/f4045ba3d962105ae4a7c0ee785a83c678ca2f8a/app/assets/javascripts/components/components/column_back_button_slim.jsx#L27-L29) which always returns the user to `/`. You'll notice that if you click Local Timeline, then Federated Timeline, then back, you are returned to Getting Started instead of the expected Local Timeline.

The [account_timeline](https://github.com/tootsuite/mastodon/blob/master/app/assets/javascripts/components/features/account_timeline/index.jsx#L67) is different because it can be called directly with a URL or while clicking on a profile in one of your timelines so instead it uses [collumn_back_button](https://github.com/tootsuite/mastodon/blob/f4045ba3d962105ae4a7c0ee785a83c678ca2f8a/app/assets/javascripts/components/components/column_back_button.jsx#L18-L19). The section I've highlighted looks like it was coded with good intentions, but I'm not entirely sure that if statement is ever called.

[...]

PR #5616, opened 2017-11-07
Show confirmation dialog on leaving WebUI while composing

unarist:
Currently, Back button and Back hotkey can cause leaving from WebUI, as well as browser's back button. Users may hit those buttons accidentally, and their composing text will be lost.

So this prevents it by showing confirmation dialog from `onbeforeunload` event.

Thoughts:

* Custom message on `returnValue` is only used on Edge. So maybe we can put simple text or blank string.
* `onbeforeunload` can also mitigate #946 #3515, but I don't think it's good because 1) this may not work on mobile and some cases 2) I think it should back to Getting Started or simply ignore instead of unwanted dialog.
* Even this case, we may want to save and load composing text silently without dialog.

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -46,7 +46,7 @@ import { defineMessages, injectIntl } from 'react-intl';
 import '../../components/status';
 
 const messages = defineMessages({
-  beforeUnload: { id: 'ui.beforeunload', defaultMessage: 'Your composing text will be lost by leaving from Mastodon.' },
+  beforeUnload: { id: 'ui.beforeunload', defaultMessage: 'Your draft will be lost if you leave Mastodon.' },
 });
 
 const mapStateToProps = state => ({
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/features/ui/index.js
```diff
@@ -39,13 +39,19 @@ import {
...
 import { HotKeys } from 'react-hotkeys';
 import { me } from '../../initial_state';
+import { defineMessages, injectIntl } from 'react-intl';
 
 // Dummy import, to make sure that <Status /> ends up in the application bundle.
 // Without this it ends up in ~8 very commonly used bundles.
 import '../../components/status';
 
+const messages = defineMessages({
+  beforeUnload: { id: 'ui.beforeunload', defaultMessage: 'Your composing text will be lost by leaving from Mastodon.' },
```
Gargron: "Your draft will be lost if you leave Mastodon" sounds better

nightpool:
Isn't the bigger problem when you click "reply" or something else that clears the text area? shouldn't we show the warning then?

unarist:
@nightpool Ah yes, definitely. I think it can be a different PR because it would have different warning text with our confirmation modal.

nightpool:
Sounds good

Gargron (approved):

PR #5622, opened 2017-11-08
Add and Remove tests for FollowRequest

ysksn:
- make an alias for `#destroy!`
- remove the validation tests
- add tests for `#authorize!`

### DIFF
--- a/app/models/follow_request.rb
+++ b/app/models/follow_request.rb
@@ -27,7 +27,5 @@ class FollowRequest < ApplicationRecord
     destroy!
   end
 
-  def reject!
-    destroy!
-  end
+  alias reject! destroy!
 end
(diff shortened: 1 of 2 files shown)


app/models/follow_request.rb
```diff
@@ -27,7 +27,5 @@ def authorize!
     destroy!
   end
 
-  def reject!
-    destroy!
-  end
+  alias reject! destroy!
```
ykzts: `alias :reject!, :destory!` ?

ysksn: > [When aliasing a method, you can use either its name or a symbol:](https://docs.ruby-lang.org/en/2.4.0/syntax/miscellaneous_rdoc.html#label-alias)

I actually don't care either to use bareword(reject!) or symbol(:reject), but RuboCop asked me to use bareword.

Do you want me to use symbol?

Gargron (approved):

PR #5636, opened 2017-11-09
Selectable federate mode

Nanamachi:
Related Issue: #861 

As argued in above issue, local instance only post is desirable feature for smaller or closed community. This PR enables below:

- select whether to deliver or show the the post to user on other instances, like this pic.

- when federate flag is false:

  - neither PuSH nor ActivityPub work for non-reply toot
  - ActivityPub distributes mentions only for mentioned user (like direct message)
  - visitor who do not log in the instance cannot see the toots via any methods below: web, atom or api
  - reblogs will be also unfederated

- federate flag is true in default and modifiable with config/settings.yml
- individual user CANNOT save default settings --- this does NOT mean they cannot choose, but only cannot save.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -48,7 +48,11 @@ class AccountsController < ApplicationController
   end
 
   def default_statuses
-    @account.statuses.where(visibility: [:public, :unlisted])
+    if current_account.nil?
+      @account.statuses.where(visibility: [:public, :unlisted], federate: true)
+    else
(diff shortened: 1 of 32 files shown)


akihikodaki (changes requested):
Thank you for your contribution, but I'm against this change because of three reasons:
* Mastodon is really powerful and it is definitely possible to use it in closed community, but I believe other chatting tools such as Mattermost and RocketChat are better for such uses.
* Too many options. We already have tons of options.
* Enabling this feature by default can kill federation in future.

You may make additional changes to address those problems.

PR #5649, opened 2017-11-10
ActivityPub: Store the follow request ID in the follow request

puckipedia:
This should fix the issue of outgoing follow accepts having wrong `object` id, by just not serializing the `object` and storing the incoming `id` for the `Follow`:

    {
        "type": "Follow",
        "object": "https://mastodon.example.com/users/puckipedia",
        "actor": "https://kroeg.example.com/users/puckipedia",
        "id": "https://kroeg.example.com/users/puckipedia/status/abcdef"
    }

results in:

    {
        "type": "Accept",
        "actor": "https://mastodon.example.com/users/puckipedia",
        "object": "https://kroeg.example.com/users/puckipedia/status/abcdef",
        "id": "https://mastodon.example.com/users/puckipedia#accepts/follows/123"
    }

Tested by following a Mastodon user from a Kroeg user. Response seemed to be fine.

oh, and sorry @Wxcafe for pointing this at your instance before remembering this was an issue

### DIFF
--- a/app/lib/activitypub/activity/follow.rb
+++ b/app/lib/activitypub/activity/follow.rb
@@ -6,13 +6,13 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
 
     return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.requested?(target_account)
 
+    follow_request = FollowRequest.create!(account: @account, target_account: target_account, remote_id: @json['id'])
+
     # Fast-forward repeat follow requests
     if @account.following?(target_account)
-      AuthorizeFollowService.new.call(@account, target_account, skip_follow_request: true)
+      AuthorizeFollowService.new.call(@account, target_account)
       return
     end
-
-    follow_request = FollowRequest.create!(account: @account, target_account: target_account)
 
     if target_account.locked?
       NotifyService.new.call(target_account, follow_request)
--- a/app/serializers/activitypub/accept_follow_serializer.rb
+++ b/app/serializers/activitypub/accept_follow_serializer.rb
@@ -3,7 +3,8 @@
 class ActivityPub::AcceptFollowSerializer < ActiveModel::Serializer
   attributes :id, :type, :actor
 
-  has_one :object, serializer: ActivityPub::FollowSerializer
+  attribute :remote_id, key: :object
+  delegate :remote_id, to: :object
 
   def id
     [ActivityPub::TagManager.instance.uri_for(object.target_account), '#accepts/follows/', object.id].join
--- a/app/models/follow_request.rb
+++ b/app/models/follow_request.rb
@@ -8,6 +8,7 @@
 #  account_id        :bigint           not null
 #  id                :bigint           not null, primary key
 #  target_account_id :bigint           not null
+#  remote_id         :string
 #
 
 class FollowRequest < ApplicationRecord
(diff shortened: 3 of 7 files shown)


puckipedia:
The issue with the broken IDs was actually a separate issue, which has also been fixed here:
If you follow someone who is already followed, the code "fast-forwards" the follow, and directly authorizes a transient follow request. Which won't have an ID. Fixed this by putting these follow requests in the database as well, before accepting the follow.

app/lib/activitypub/activity/follow.rb
```diff
@@ -6,14 +6,14 @@ def perform
 
     return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.requested?(target_account)
 
+    follow_request = FollowRequest.create!(account: @account, target_account: target_account, remote_id: @json['id'])
```
akihikodaki: I do not think writing an object which lives just in an execution context to persistent database can be tolerated. What about passing `FollowRequest` object to `AuthorizeFollowService` _without_ making it persistent?

app/serializers/activitypub/accept_follow_serializer.rb
```diff
@@ -3,7 +3,8 @@
 class ActivityPub::AcceptFollowSerializer < ActiveModel::Serializer
   attributes :id, :type, :actor
 
-  has_one :object, serializer: ActivityPub::FollowSerializer
+  attribute :remote_id, key: :object
```
akihikodaki: It makes any `Follow` request made before this change fail, right? I do not think it is OK. We need a fallback.

akihikodaki (changes requested):
I think it is important that Mastodon conforms to the standard and really appreciate this change.
On the other hand, I also believe the backward compatibility also matters and should be maintained as long as it is possible without violating the standard.

Please also make the corresponding change to the activity handler. The handler is located at `app/lib/activitypub/activity/accept.rb`. It should also give you a hint to make the new implementation compatible with old Mastodons.

Gargron:
I was under the impression that #5772 was a refactor of this PR but apparently it's missing the database changes which results in #6216 and #6211 (am I right in assuming this changeset would fix those, @puckipedia?)

puckipedia:
right, so I hadn't looked at this PR for a while. Effectively, this is a fix for the /other/ side of the equation, and before this patch is merged I think there'd need to be a few more changes:

- Keep track of the Follow `id` inside of the follow request *and* normal follow table
- When receiving a Follow, if already accepted update the follow table to contain the new follow id, then reply with an accept
- When receiving an undo follow, either accept the follow ID or an object describing the follow (optional?)
- When accepting/rejecting a follow, send back the follow ID (including, possibly, part of the follow, this /should/ be filtered out)

this would be a slightly nicer fix than what is being done in this PR, imo. (I do think both cover most of the situations it'll encounter)

Gargron:
@puckipedia I can't do this under these time constraints (we need to release another patch asap). If I had a month to do it, like for a minor release, then that's fine, I would sit down and refactor all this stuff, but I don't have the headspace for it right now. Now, should we just revert #5772 for the time being?

puckipedia:
so far, I've been able to work around the issue. So I wouldn't say too critical for now, I think?

akihikodaki:
> I was under the impression that #5772 was a refactor of this PR

It is not. #5772 is a change for local follow requests. This pull request is for remote follow requests.

> but apparently it's missing the database changes which results in #6216 and #6211 (am I right in assuming this changeset would fix those, @puckipedia?)

I will investigate those issues and reply to each thread.

PR #5650, opened 2017-11-10
 Add tests for PreviewCard#save_with_optional_image!

ysksn:


spec/models/preview_card_spec.rb
```diff
@@ -1,5 +1,30 @@
...
+    context 'object is valid' do
+      it 'keeps image' do
+        preview_card.save_with_optional_image!
+        expect(preview_card.image.present?).to be true
+      end
+    end
+
+    context 'object is invalid' do
+      before do
+        c = 0
```
ysksn: Since `#save_with_optional_image!` calls `#save!` twice, I don't want to raise `ActiveRecord::RecordInvalid` in second time.

Gargron (approved):

PR #5655, opened 2017-11-10
Monkey-patch Paperclip to use GraphicsMagick instead of ImageMagick

ClearlyClaire:
GraphicsMagick is generally considered more secure and more efficient than ImageMagick, although I am not sure how much of that is true.
Unfortunately, Paperclip uses ImageMagick, and not GraphicsMagick.
Unless I am mistaken, there is no other way than monkey-patching Paperclip to use GraphicsMagick.

Marking it as work in progress as I am not too sure about the benefits, and I'm not too sure GraphicsMagick is an exact drop-in replacement to ImageMagick for our purposes. I will be running this patch and see if any issue arises.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN apk -U upgrade \
     file \
     git \
     icu-libs \
-    imagemagick \
+    graphicsmagick \
     libidn \
     libpq \
     nodejs \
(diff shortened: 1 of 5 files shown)


ThibG:
GraphicsMagick cannot be used as a drop-in replacement for ImageMagick in Paperclip, as it does not support the same options. For instance, when handling gifs, `convert` is called with `-layers "optimize"` which is not supported by GraphicsMagick.

wxcafe:
Okay, so should this be closed then?

ThibG:
Yeah indeed. It would be cool to replace ImageMagick with something safer and less memory-consuming, but this probably isn't the correct approach.

PR #5669, opened 2017-11-13
Add option to block direct messages from people you don't follow

### LINKED ISSUE #5326: Allow people to reject all DMs from people they don't follow
I have noticed that a lot of harassment on mastodon happens in DMs. This protects the harassers because most of the protections against harassment are social and a common tactic is to send harassing messages over DM where they can't be seen by most people and then pretend that the target of the harassment is being unreasonable when they publicly react because other people can't see the cause.

I think that a setting to allow accounts to not receive DMs from at least people that the account doesn't follow would go a long way to helping this sort of harassment. If it could be set up so you can have multiple levels like blocking all DMs, only allowing DMs from your followers or only allowing DMs from people you follow that would be better. I think that at least the last one is necessary to help prevent harassment.

* * * *

Gargron:
Fix #5326 

I heard some discussion about communicating to the end-user that the message could not be delivered when it's blocked like this, **however** there are no such indications when notifications are blocked using any other means in NotifyService, and they are generally impossible to implement consistently between local and remote accounts and therefore undesirable. Plus since a direct message is not generally *directed* as much as it *tags* one or more people who may or may not see it...

### DIFF
--- a/app/controllers/settings/notifications_controller.rb
+++ b/app/controllers/settings/notifications_controller.rb
@@ -26,7 +26,7 @@ class Settings::NotificationsController < ApplicationController
   def user_settings_params
     params.require(:user).permit(
       notification_emails: %i(follow follow_request reblog favourite mention digest),
-      interactions: %i(must_be_follower must_be_following)
+      interactions: %i(must_be_follower must_be_following must_be_following_dm)
     )
   end
 end
--- a/app/services/notify_service.rb
+++ b/app/services/notify_service.rb
@@ -53,8 +53,15 @@ class NotifyService < BaseService
     @notification.type == :mention && @notification.target_status.direct_visibility?
   end
 
+  def response_to_recipient?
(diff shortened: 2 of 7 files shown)


vahnj:
Looks like this is just a face-lift on the changes Paul and i made and doesn't address the criticism of the change - if a user initiates a direct message with someone they don't follow with the option on, there is no way to see responses from said person, unless I'm misreading your changes.

nightpool:
@Gargron I think it makes a lot of sense from a UX and product perspective to a do a Reject { Note } for DMs that get filtered due to this option, which would bring us to parity (IMO) with twitter's feature in this area.

nightpool (approved):
LGTM. Errors are probably not critical here

inmysocks:
I think that it is important to point out that letting an abuser know that their message hasn't gotten through causes more abuse because then they will try and find another channel to use for harassment.

PR #5670, opened 2017-11-13
[Nanobox] Apply Release Notes Changes

danhunsaker:
Apparently I missed some things in earlier commits/releases that needed to be applied to the Nanobox setup. All minor things, nothing that breaks anything, but still best to get them in place.

- Move cron jobs to their own component, so the Sidekiq component can be scaled up to multiple instances without causing issues with running the same cron job multiple times at once.
- Update cron jobs to the latest requirements
- Add new variables to `.env.nanobox`
- Update Nginx to use correct cache header directives

### DIFF
--- a/boxfile.yml
+++ b/boxfile.yml
@@ -42,6 +42,7 @@ run.config:
 
   fs_watch: true
 
+
 deploy.config:
   extra_steps:
     - NODE_ENV=production bundle exec rake assets:precompile
@@ -59,6 +60,7 @@ deploy.config:
   before_live:
     web.web:
       - bundle exec rake db:migrate:setup
+
 
 web.web:
   start:
@@ -78,6 +80,7 @@ web.web:
     data.storage:
       - public/system
 
+
 web.stream:
   start:
     nginx: nginx -c /app/nanobox/nginx-stream.conf
@@ -91,8 +94,13 @@ web.stream:
   writable_dirs:
     - tmp
 
+
 worker.sidekiq:
-  start: bundle exec sidekiq -c 5 -q default -q mailers -q pull -q push -L /app/log/sidekiq.log
+  start:
+    default: bundle exec sidekiq -c 5 -q default -L /app/log/sidekiq.log
+    mailers: bundle exec sidekiq -c 5 -q mailers -L /app/log/sidekiq.log
+    pull: bundle exec sidekiq -c 5 -q pull -L /app/log/sidekiq.log
+    push: bundle exec sidekiq -c 5 -q push -L /app/log/sidekiq.log
 
   writable_dirs:
     - tmp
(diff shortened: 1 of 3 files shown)


boxfile.yml
```diff
@@ -105,6 +113,20 @@ worker.sidekiq:
...
+
+  log_watch:
+    rake: 'log/production.log'
+
+  network_dirs:
+    data.storage:
+      - public/system
+
   cron:
     - id: generate_static_gifs
```
Gargron: Wait, hold on. Are these run as cron? They are not regular tasks.. They were upgrade one-offs...

danhunsaker: I didn't see any mention of _when_ to run them - or not - so they've been hiding in the background for a while, now. Will happily remove whichever ones I can! It's pretty clear you at least mean the task to generate a static image from a gif; does this apply to any of the others as well?

boxfile.yml
```diff
@@ -131,6 +144,10 @@ worker.sidekiq:
       schedule: '30 00 * * *'
       command: 'bundle exec rake mastodon:push:clear'
 
+    - id: update_remote_avatars
```
Gargron: Remove this one

boxfile.yml
```diff
@@ -105,20 +113,25 @@ worker.sidekiq:
...
+
+  log_watch:
+    rake: 'log/production.log'
+
+  network_dirs:
+    data.storage:
+      - public/system
+
+  cron:
     - id: update_counter_caches
```
Gargron: Remove this one

boxfile.yml
```diff
@@ -105,20 +113,25 @@ worker.sidekiq:
...
       schedule: '50 * * * *'
       command: 'bundle exec rake mastodon:maintenance:update_counter_caches'
 
-    # runs feeds:clear, media:clear, users:clear, and push:refresh
-    - id: do_daily_tasks
-      schedule: '00 00 * * *'
-      command: 'bundle exec rake mastodon:daily'
-
     - id: clear_silenced_media
```
Gargron: Very subjective on whether you want to run this or not. Do you want to delete stuff shadowbanned people upload? Part of the appeal of the "silence" feature is that it's a more or less invisible quarantine for the person who's in it.

Gargron (changes requested):

boxfile.yml
danhunsaker: Is there already a mechanism for relaying avatar changes through the Fediverse as hey happen, then? I'd like to show my users the current avatar of anyone they follow, and this _appears_ to be the only way to make it happen, so I'd want to comment it out rather than remove it entirely if there's no alternative already in place.

boxfile.yml
danhunsaker: This is done automatically as well, then? Or do the docs just not explain its purpose very well?

boxfile.yml
danhunsaker: I'll comment this one out and leave a note on when it might be desirable to enable it, then. :+1:

boxfile.yml
Gargron: Yes, obviously. This cron is troubleshooting.

boxfile.yml
Gargron: This CLI command has been used in the upgrade when counter caches were first introduced and needed to be filled.

boxfile.yml
danhunsaker: Where does it state that, exactly? This is the only place I've seen that mentioned...

boxfile.yml
danhunsaker: Ok, cool. Consider it gone.

boxfile.yml
danhunsaker: I don't see `update_remote_avatars` listed in that doc at all.

Gargron (approved):

PR #5679, opened 2017-11-13
Home from db after limit

ClearlyClaire:
This change allows to scroll back past the 400 toots limit of the Redis cache for the Home Timeline.
Indeed, some users like to read all the toots from their timeline, but may be away long enough for it to exceed the 400 toots limit.

This PR is tagged as work in progress as it may be greatly inefficient---although that only affects people actually scrolling past those 400 toots or waiting for the cache to be repopulated.

I guess it would be ideal to move the various filters to the SQL query in `Status#as_home_timeline`.

### DIFF
--- a/app/models/home_feed.rb
+++ b/app/models/home_feed.rb
@@ -8,11 +8,11 @@ class HomeFeed < Feed
   end
 
   def get(limit, max_id = nil, since_id = nil)
-    if redis.exists("account:#{@account.id}:regeneration")
-      from_database(limit, max_id, since_id)
-    else
-      super
+    unless redis.exists("account:#{@account.id}:regeneration")
+      statuses = super
+      return statuses unless statuses.empty?
     end
+    from_database(limit, max_id, since_id)
   end
 
   private
--- a/app/models/home_feed.rb
+++ b/app/models/home_feed.rb
@@ -18,8 +18,13 @@ class HomeFeed < Feed
   private
 
   def from_database(limit, max_id, since_id)
-    Status.as_home_timeline(@account)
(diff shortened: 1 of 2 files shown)


ThibG:
I currently have no idea what's going on with the failing tests, to be honest.

app/models/feed.rb
```diff
@@ -24,9 +24,14 @@ def from_redis(limit, max_id, since_id)
   end
 
   def from_database(limit, max_id, since_id)
-    Status.as_home_timeline(@account)
-          .paginate_by_max_id(limit, max_id, since_id)
-          .reject { |status| FeedManager.instance.filter?(:home, status, @account.id) }
+    for attempt in 0..20 do
```
ykzts: `for attempt in 0..20 do` -> `21.times do`

ThibG: Right, thanks! This commit is not needed, though, I just tried to avoid the test failures, but I still don't understand it. I will drop that commit.

ThibG:
Fixed the test (since it now passes). This is still sub-optimal as it may return less statuses than requested while performing potentially useless SQL queries, but it works.

ThibG:
Removed the “work in progress” label as I'm probably not going to work on the SQL side of things.
It works reasonably well for my single-user instance, not sure how bad it would be performance-wise in practice.

akihikodaki:
Duplicate of #4016?

ThibG:
@akihikodaki yes, indeed, gonna close this one, but you want to fix `from_database` not returning anything if the first database results are filtered

PR #5680, opened 2017-11-13
Display toot in timeline if it's from a followed account and mentions us

ClearlyClaire:
I am not sure whether that is just me, but I'm frequently confused when a public toot from someone appears in my notifications but not in my home timeline. This happens when the toot is a reply to someone not followed.

This pull request changes the behavior to always include in the home timeline toots from people we follow and mentioning us.


nightpool:
How often does this happen? I've never seen this behavior.

nightpool:
The feels really inconsistent and like it muddies the rules of the notifications tab/home tab in a way that makes it harder to understand why things ended up where.

ThibG:
Well, basically, *every* time that happens I'm wondering if something's broken. But maybe it's just me, so I'm not really pushing this change. I understand your concern.

Gargron:
I think this would be a confusing change, it also adds an extra query to a hotspot, so I will err on the side of keeping things the way they are. Thank you.

PR #5687, opened 2017-11-14
Add tests for StreamEntry

ysksn:
- `#object_type`
- `#verb`
- `#mentions`


spec/models/stream_entry_spec.rb
```diff
@@ -6,6 +6,88 @@
...
+        expect(stream_entry.object_type).to be :activity
+      end
+    end
+
+    context 'orphaned? and targeted? are false' do
+      let(:orphaned) { false }
+      let(:targeted) { false }
+
+      it 'calls status.object_type' do
+        expect(stream_entry).to receive_message_chain(:status, :object_type)
```
Gargron: I've been seeing these kinds of tests often recently and it worries me. They test the implementation rather than input->output, therefore any future refactoring will break the tests which is inconvenient.

Gargron (approved):

Issue #5693, opened 2017-11-14
Ability to change more settings in admin panel


Cassolotl:
I'm asking for this because basically, as Mastodon gets more popular and continues to be hard to install, people are going to be paying other people to host their instance for them more and more - and for that you don't need much technical ability, you just do everything admin-ish through admin settings. Being able to customise things like the character limit through admin settings instead of having to ask your host to do it is going to become more and more important.

Inspired by [this toot by a non-techy admin](https://queer.town/@alice/99002490323809366).

Signed, another non-techy wannabe admin.

* * * *

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Fiona:
My wife is the linked toot and she asked me to point out that she meant bios as well as toots

wxcafe:
So as mentioned by @Gargron in #5697, configurable character limits are not a planned feature. Allowing more stuff to be configured from the graphical admin panel would be great tho, so I’m renaming this issue

Gargron:
Please either add a checklist for things you think are lacking in the admin settings, or open separate issues for each setting. As-is, I do not know what this issue is asking for (character limit is not planned to be configurable)

Cassolotl:
See also: *Let the admins easily setup the character limits #12265*

Issue #4915, opened 2017-09-13
API to get server constants (toot length, ...)


tateisu:
It world be nice to make API to get constants thats is used in validation.

it provides some constants to need to validate profile,toots,media attachment.
- no need to new endpoint. it may be in instance-information API.
- no need to enumerate all the possibilities. it shows some build-time constants that used in official mastodon.

Wanted attributes from app-side validations:
- max-length of toot + spoiler_text (already defined in web-app)
- max-length of user display name (already defined in web-app)
- max-length of user note (already defined in web-app)
- max-number of media attachment (already defined in web-app)
- list of mime types that acceptable for media attachment. (already defined in web-app)
- max-length of attached media bytes count. (before re-encoding, it may depends on limitation in Web front server. maybe we need to make variable in .env.production ? )

tateisu:
As far as I know, mstdn.maud.io has modified these constants
Kanzaki-don also changes the upper limit of the number of bytes of attached media

lambadalambda:
I'd be very interested in that too. Now that Pleroma speaks the Mastodon api, it would be good if it was possible to tell client apps how much characters they can actually use.

wxcafe:
This is a really good idea yeah.

unarist:
I'm thinking `limits` object in `/api/v1/instance`, for example:

```rb
{
    uri: "example.com",
    title: "example",
    description: "<p>hoge</p>",
    email: "admin@example.com",
    version: "2.1.0",
    thumbnail: "https://example.com/system/site_uploads/files/000/000/001/original/mstdn_opengraph.png"
    urls: {
      streaming_api: "wss://example.com"
    },
    stats: {
      user_count: 1,
      status_count: 1,
      domain_count: 1
    },
    limits: {
      account_display_name_length: 30,
      account_note_length: 160,
      status_length: 500,
      # 4 for images, but 1 for video?
      status_attachment_count: 4,
      # less than 8.megabytes
      attachment_size: 8.megabytes - 1,
      attachment_mime_types: ['image/jpeg', 'image/png', 'image/gif', 'video_webm', 'video/mp4'],
      attachment_extensions: ['.jpg', '.jpeg', '.png', '.gif', '.webm', '.mp4', '.m4v']
    }
  }
```

nullkal:
How about `capability` instead of `limits`? Then, we can add information not only limitation, but also whether the instance implements features such as `pinned toots` by simply add `pinned_toots: true`.

[...]

PR #5697, opened 2017-11-14
Make character limit configurable

lambadalambda:
This adds a character limit configurable by setting and environment variable. It also adds the character limit to the initial state and instance endpoint, so clients can use it to set the correct limit.

Default is still 500 characters if no other value is set.

This relates to #5693 and #4915.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -136,4 +136,4 @@ STREAMING_CLUSTER_NUM=1
 # GID=1000
 
 # Maximum allowed character count
-# MAX_CHARS=500
+# MAX_TOOT_CHARS=500
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -134,3 +134,6 @@ STREAMING_CLUSTER_NUM=1
 # If you use Docker, you may want to assign UID/GID manually.
 # UID=1000
 # GID=1000
+
+# Maximum allowed character count
+# MAX_CHARS=500
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -20,7 +20,7 @@ import { length } from 'stringz';
 import { countableText } from '../util/counter';
 import initialState from '../../../initial_state';
 
-const maxChars = initialState.max_chars;
+const maxChars = initialState.max_toot_chars;
 
 const messages = defineMessages({
   placeholder: { id: 'compose_form.placeholder', defaultMessage: 'What is on your mind?' },
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -19,6 +19,7 @@ import ImmutablePureComponent from 'react-immutable-pure-component';
 import { length } from 'stringz';
 import { countableText } from '../util/counter';
 import initialState from '../../../initial_state';
+
 const maxChars = initialState.max_chars;
 
 const messages = defineMessages({
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -18,6 +18,8 @@ import { isMobile } from '../../../is_mobile';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 import { length } from 'stringz';
 import { countableText } from '../util/counter';
+import initialState from '../../../initial_state';
+const maxChars = initialState.max_chars;
 
 const messages = defineMessages({
   placeholder: { id: 'compose_form.placeholder', defaultMessage: 'What is on your mind?' },
@@ -201,8 +203,8 @@ export default class ComposeForm extends ImmutablePureComponent {
           </div>
 
           <div className='compose-form__publish'>
-            <div className='character-counter__wrapper'><CharacterCounter max={500} text={text} /></div>
-            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || this.props.is_uploading || length(text) > 500 || (text.length !== 0 && text.trim().length === 0)} block /></div>
+            <div className='character-counter__wrapper'><CharacterCounter max={maxChars} text={text} /></div>
+            <div className='compose-form__publish-button-wrapper'><Button text={publishText} onClick={this.handleSubmit} disabled={disabled || this.props.is_uploading || length(text) > maxChars || (text.length !== 0 && text.trim().length === 0)} block /></div>
           </div>
         </div>
       </div>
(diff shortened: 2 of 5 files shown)


.env.production.sample
```diff
@@ -134,3 +134,6 @@ STREAMING_CLUSTER_NUM=1
 # If you use Docker, you may want to assign UID/GID manually.
 # UID=1000
 # GID=1000
+
+# Maximum allowed character count
+# MAX_CHARS=500
```
ThibG: I'd suggest MAX_TOOT_CHARS or something similar, as one may want custom maximum length for other things (such as bios).

ClearlyClaire (commented):
Looks good to me, except for the minor config naming issue.

.env.production.sample
lambadalambda: Makes sense, i made that change.

.env.production.sample
```diff
@@ -134,3 +134,6 @@ STREAMING_CLUSTER_NUM=1
 # If you use Docker, you may want to assign UID/GID manually.
 # UID=1000
 # GID=1000
+
+# Maximum allowed character count
+# MAX_TOOT_CHARS=500
```
m4sk1n: afaik, toots are called statuses in code…

lambadalambda: You think it's worth to rename it again?

ThibG: I'm personally fine either ways. `MAX_STATUS_CHARS` would be more consistent with the source code, but the configuration option is intended for admins, not for developers.

ClearlyClaire (approved):

Gargron:
I have denied such patches in the past. If somebody wants to hack it in, they may do so, but I don't want to encourage it. I feel like the UX is designed, visually and behaviourally, around a certain number of characters, and deviations have a negative impact on it. For example, I want people to pick their Mastodon based on, for example, the policies, or the theme, but I want the base level of functionality to be the same so it doesn't go into the choice.

The same thing as here is with the file size limits. Some people want to make them bigger. Well, great, but if your users make use of those higher limits their attachments/avatars will not show up on other instances which have to cache that stuff. You can do that, but I'm not going to make that an Official Thing.

lambadalambda:
I can understand that, but many instances already change their character counts. Some do it for jokes (e.g. the 512 characters for a tech-oriented instance), some do it for real reasons (the 10.000 characters on an instance for writers).

If it's easily changeable in the main codebase or not isn't really important to me, but I'd like to have some API to get the character count of the instance, so apps can work correctly with modified servers.

SirCmpwn:
>I have denied such patches in the past. If somebody wants to hack it in, they may do so, but I don't want to encourage it. I feel like the UX is designed, visually and behaviourally, around a certain number of characters, and deviations have a negative impact on it. For example, I want people to pick their Mastodon based on, for example, the policies, or the theme, but I want the base level of functionality to be the same so it doesn't go into the choice.

:-1:

In my opinion, this isn't your place to decide - it's for the instances to decide. Don't make them hack it. And they *are* hacking it - I've seen a good handful of toots in my timeline that are rather long, and because Mastodon isn't intelligent about alternative character limits it can make for a very annoying UX to have to scroll past ridiculously long toots.

Mastodon should support arbitrary lengths (but default to 500 is totally fine, opinionated default settings are better than opinionated hardcoded values), and also have a configurable maximum length at which the rest of the toot will be hidden behind a "read more" button. We already have such a UX for content warnings, this isn't breaking new ground design-wise.

Mastodon is good because it gives users that freedom they don't have on Twitter. You should embrace that.

connyduck:
For App-Developers like me it is really important to have an API to get the character limit of the currently connected instance. Please give us an API so we can adapt our clients to the already modified mastodon-servers and provide our users the best experience.
Maybe the limit should not be configurable arbitrarly but only within a sensible limit like 140 -10 000.

Lertsenem:
As a compromise, how about hard coding that toot with more than 500 chars get automatically wrapped in a CW with an ellipsis?

That's already kind of an informal etiquette anyway, and leaves the web UI intact for the most part.

akihikodaki:
I wonder if there is any chance for this pull request to be merged. If so, we can have more discussions and improvements. If not, and this issue gets closed, others who like this change can decide to maintain it by themselves or abandon.

lambadalambda:
@Gargron Can you please either merge or close this?

wxcafe:
Up @Gargron

jennifermf:
can we revisit this? if we can have a billion custom emojo, can we please please please get a 500 character limit bio as the default?

the average mastodon user doesn't fit neatly into twitter-sized boxes. it's time to give people more space to describe who they are.

Shreewebs:
I am looking to increase TOOT character count to 600. It seems file .env.production.sample no longer has such options open.
Looking for a tweak!
Thank you

Grawl:
Just did this at my fork https://github.com/Grawl/mastodon/commit/aef40d7cfa2b27219bf6ef5b08bc804f6fdf7ee2

Grawl:
just did it in my fork https://github.com/Grawl/mastodon/commit/3fd85bcf412235e546fe4ac4dd418f2dc4ddd033

tyteen4a03:
Are there any scripts to work around this?

r3d-f0x:
> As a compromise, how about hard coding that toot with more than 500 chars get automatically wrapped in a CW with an ellipsis?

I would **not** want this done.  Would love to see long toots past the default 500 characters be limited to the first 500 up to the complete word and then expandable by the user, but using "CW" will devalue the use of CW.  I run a right leaning site an intend to use CW as a way of moderating toots that could be problematic without outright deleting them.  Doing anything that would get people into the habit of clicking to see CW without thinking about it would make it a lot more challenging to deal with problematic posts.

itbeard:
Hello! @Gargron maybe in 2022 it makes sense to reconsider your decisions about this feature?

rkingett:
I said this in another issue but I have no problem saying it here. Many Blind instance administrators want this, especially because it is easier to read one lengthy post than 60 posts broken up into a thread and Blind instance admins want to cultivate a space dedicated to their users needs.

JiFish:
I wanted to add another reason for this that I don't think has been mentioned yet: I'm using managed Mastodon hosting for my instance. In this situation, I can't reasonably expect my host to apply a hack for just me.

expertmanofficial:
I would like to wonder how do people comment here even through the issue is already closed?

tyteen4a03:
> I would like to wonder how do people comment here even through the issue is already closed?

Closed issues only mean the issue is considered closed. Maintainers can choose to lock an issue which is another thing.

yawnbox:
1. Please give communities freedom to choose how much their profiles can say, without hacks.
2. Please lower the bar for admins so the diversity of admins increases.

rkingett:
I can't tell if @Gargron  knows why people pick servers or not. People already pick servers based on a higher character limit and on the public timelines I see people asking for two things.

1. A high character limit. 2,000 and up.
2. An instance theme to go along with the high character limit.

As I said previously, the flagship instances can keep this at 500 if they want. Just make it easier for admins to cadre to their communities.

Simx72:
> People already pick servers based on a higher character limit and on the public timelines I see people asking for two things.

I do. Seeing if in 2026, this has changed.

Issue #5035, opened 2017-09-21
Language filtering on the streaming websocket


renatolond:
Hi!

I've just brought up my instance and I was checking the global timeline since there's not much happening on the local one yet. And I've noticed a strange behaviour:

The first screenshot is while it's streaming the content coming:
![image](https://user-images.githubusercontent.com/173791/30693276-1a90e7f2-9ecf-11e7-8a72-d1e6884c3a41.png)

However, if I refresh:
![image](https://user-images.githubusercontent.com/173791/30693286-2b7e7886-9ecf-11e7-87da-0041c8249229.png)

It seems to me that the streaming api is not filtering by language and then when it gets the info back from the regular api, it doesn't show toots detected with the filtered language.

I think this is unrelated to #3706, since the toots are filtered if I update the page (most of them, anyway).

I'm on 1.6.1.
* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

renatolond:
This seems worst in 2.0.0. It seems that now there's no difference when updating the page for the first time, so either the streaming problem was fixed and now I'm seeing only #3706 or the first load by the server is now also affected.

renatolond:
Okay, I've been testing around a bit, and I found out that there's quite a bit of statuses without the language tagged. I went to try and find out why and I noticed that specially for short texts that CLD3 detect as Japanese, if we strip html_tags, the detection gets reliable and probability quite high, otherwise, it's not the case.

For instance:
```
2.4.2 :063 > new_text
 => "<p>上手いし正論だし返す言葉もないです</p>"
2.4.2 :064 > CLD3::NNetLanguageIdentifier.new(1, 2048).find_language(new_text)
 => #<struct Struct::Result language=:ja, probability=0.5030454993247986, :reliable?=false, proportion=1.0>
2.4.2 :065 > nu_text = '上手いし正論だし返す言葉もないです'
 => "上手いし正論だし返す言葉もないです"
2.4.2 :066 > CLD3::NNetLanguageIdentifier.new(1, 2048).find_language(nu_text)
 => #<struct Struct::Result language=:ja, probability=1.0, :reliable?=true, proportion=1.0>
```

I wonder if html_stripping as part of simplify_text would improve detection. I use something like that in the crossposter using Loofah and Rails PermitScrubber.

PR #5701, opened 2017-11-14
Improve language filter

renatolond:
After investigating #5035 I've made some tests at masto.donte.com.br to find out what could improve the language filter and this are the results of my investigation.

I'm removing html tags as the first stage of simplify text, so that the final text sent to CLD3 only the original text or as close to that as possible.

Also, when receiving a text in ActivityPub, if the language doesn't come, I'm trying to detect it from the text that was received.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -173,7 +173,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def language_from_content
-    return nil unless language_map?
+    return LanguageDetector.instance.detect(text_from_content, @account) unless language_map?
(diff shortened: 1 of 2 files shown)


ykzts (approved):

ykzts:
Why Travis does not run?

renatolond:
Sorry, added one more commit to remove the custom_emoji before the detection.

renatolond:
Don't know about travis, I checked over there and it seems the PR is not appearing on the list 🤔 
I can close this one and open another to see if it solves it.

ykzts:
1. Close PR
2. Delete branch
3. Restore branch
4. Reopen PR

Can you try out?

PR #5721, opened 2017-11-16
Refactor ActivityPub::Activity.klass

ysksn:


app/lib/activitypub/activity.rb
```diff
@@ -23,28 +23,7 @@ def factory(json, account, options = {})
...
-      when 'Update'
-        ActivityPub::Activity::Update
-      when 'Undo'
-        ActivityPub::Activity::Undo
-      when 'Accept'
-        ActivityPub::Activity::Accept
-      when 'Reject'
-        ActivityPub::Activity::Reject
-      end
+      "ActivityPub::Activity::#{@json['type']}".constantize
```
Gargron: This feels unsafe to me.

PR #5724, opened 2017-11-16
Improve language filter

renatolond:
After investigating #5035 I've made some tests at masto.donte.com.br to find out what could improve the language filter and this are the results of my investigation.

I'm removing html tags as the first stage of simplify text, and I'm also removing custom_emoji codes after the replacements that were already there, so that the final text sent to CLD3 is as close to the original text as possible. 

Also, when receiving a text in ActivityPub, if the language doesn't come, I'm trying to detect it from the text that was received.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -173,7 +173,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def language_from_content
-    return nil unless language_map?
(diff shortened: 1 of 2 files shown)


ykzts:
Umm...

renatolond:
Let me try something, one sec

renatolond:
Yeah, my fault, it was on the masto-donte-com-br org, and travis did not had access to it. I'll remove the empty commit :)

ykzts (approved):

PR #5731, opened 2017-11-17
i18n: Update ja translation for moderator badge.

mayaeh:
Add Japanese translation for #5728 .

This is still English, but I thought this was good for the badge.

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -48,6 +48,7 @@ ja:
     reserved_username: このユーザー名は予約されています。
     roles:
       admin: Admin
+      moderator: Moderator
     unfollow: フォロー解除
   admin:
     account_moderation_notes:
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -48,7 +48,7 @@ ja:
     reserved_username: このユーザー名は予約されています。
     roles:
       admin: Admin
-      moderator: Moderator
+      moderator: Mod
     unfollow: フォロー解除
   admin:
     account_moderation_notes:


config/locales/ja.yml
```diff
@@ -48,6 +48,7 @@ ja:
     reserved_username: このユーザー名は予約されています。
     roles:
       admin: Admin
```
matt-auckland: These would be the correct translations
Admin: 管理者
Moderator: モデレーター

I'm not a native speaker but, I double checked with native speaker who hosts a Mastodon Instance

ncls7615: I agree
I think that this is better

mayaeh: I thought English notation was good for the badge.

lynlynlynx:
- If we leave them as in English, changing notation causes blurring. It's important.
- Neither "Moderator" nor "モデレーター" are familiar to Japanese people.
- For Japanese people, hierarchical relationship between "管理者" and "モデレーター" is unclear. 

Taking them into consideration, I think using "Mod" is better.

ncls7615:
I don't think that is good...

> Neither "Moderator" nor "モデレーター" are familiar to Japanese people.

I agree.
But.
Moderator? モデレーター? Mod?
These are the same in its essentials.

- Admin: 管理者
- Moderator: モデレーター

I think this translate is the best.

clworld:
admin
mod
admin_j
mod_j

clworld:
As appearance, I feel untranslated shorter things cooler.
I don't think 'モデレーター' is definitely better than others(except it's `translated`). モデレーター / Moderator / Mod are all unfamiliar to Japanese user.
I feel both are OK.

lynlynlynx:
Familiarity is not a key issue. but unfamiliarity obscures hierarchical relationship, what we want to express with these badges the most. I don't recommend using katakana loanword and kanji word for the same place, because clarifying relationship between the two is difficult. （馴染みが無いことが問題なのではなく，馴染みが無いことが我々が最も伝えるべきである上下関係の不明瞭さに繋がる．カタカナ外来語と漢字熟語間で関係を明瞭にすることは難しいため，二者を併用することを推奨しない）

If we use translated badges, I prefer to use
Admin/Mod: 管理者/準管理者, 管理者/副管理者, 管理者/管理者補佐.

No change: I think using Admin/Mod is the best.

wxcafe:
Could you maybe get this into the original PR? That way we could have a single place to review the feature and we could add all the translations there. Thanks :)

Gargron (approved):

mayaeh:
I'm sorry, I don't know what to do.
I didn't consider applying it to other languages.

PR #5740, opened 2017-11-17
Increase contrast of Admin badges

lynlynlynx:
This is to comply [WCAG 2.0 guidelines on contrast ratio](https://www.w3.org/TR/WCAG20/#visual-audio-contrast) . The minimum contrast ratio has established as 4.5.

Present contrast ratio of admin badges is 3.62.
With this commit, the ratio increases to 4.98.

|Before|After|
|-------|-----|
|ws000158|After|

### DIFF
--- a/app/javascript/styles/mastodon/accounts.scss
+++ b/app/javascript/styles/mastodon/accounts.scss
@@ -542,8 +542,8 @@
   }
 
   &.admin {
-    color: $error-red;
-    background-color: rgba($error-red, 0.1);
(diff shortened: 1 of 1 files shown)


nightpool:
@ykzts @Gargron FYI

nightpool:
I'm not sure I'm a huge fan of the way the color looks here, but we can hopefully find a happy medium that still hits the contrast guidelines

Gargron (approved):

PR #5741, opened 2017-11-17
Remove translations of default theme, except en.yml

lynlynlynx:
I think the translation of default theme "Mastodon" should fallback to en.yml. It's because, when we change default theme, we need to change names of default theme in ALL language files 😩 . "Mastodon" is the same translation in all languages.


wxcafe:
Seems sensible to me

Gargron (approved):

PR #5744, opened 2017-11-17
Make it possible to bind streaming service to BIND instead of 0.0.0.0

elegaanz:
The same as #1536, but without conflict and with a newline at the end of the file.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -460,7 +460,7 @@ const startWorker = (workerId) => {
     });
   }, 30000);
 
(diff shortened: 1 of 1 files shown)


nightpool:
I think the title of this pull request is misleading—the streaming service *already* binds to 0.0.0.0. If i'm not mistaken, this makes it possible to bind the streaming service to *127.0.0.1*

can you clarify what you mean by the title of this PR?

Gargron (approved):

PR #5746, opened 2017-11-18
Profile redirect notes

### LINKED ISSUE #177: Support account migration
A lot of people seem to be jumping on https://mastodon.social right now, even though the end goal is to have users separated out across multiple federated instances.  However, if people start putting up a lot of content and getting followers on the primary instance, this will be a disincentive to move providers.

I think this largely means that Mastodon needs to support two things:
* Account import across providers, which should be authenticated on both ends to prevent people bulk copying another person's account, and
* A "redirected account" field against user accounts, which indicates the full URL of the new user account

When users configure a redirect location against their account (whether explicitly on an account page, or implicitly set during a cross-provider account import), the instance on which the account is should _implicitly_ and automatically redirect followers.

That is, if I have the account @hq on the primary instance (which I do), and I set up the account @hach-que on another Mastodon service, the @hq should:
* Remain on the primary instance, and not be disabled in any way
* Show posts from @hach-que after the redirection is set up
* Disallow posting from the @hq account while ever the redirection is in place
* Existing followers of @hq should start seeing posts from @hach-que instead
* New followers of @hq should be allowed to follow the account (and internally, they are following @hq, but see posts from @hach-que)
* Followers should **not** actually have their lists updated to follow @hach-que - in the event of a mistaken redirect, removing the redirect should act exactly as if there never was a redirect in the first place

This is just my 2c on how I think this should work, but I'm interested to hear other people's thoughts.

Gargron:
See also #177 (partial)

image

image

- [x] Database structure to store information about where account moved
- [x] Serialize this information in REST API and ActivityPub
- [x] Deserialize this information from ActivityPub
- [x] Display note on public profiles
- [x] Display note on profiles in web UI

---

Please mind that this PR does not include the automatic followers unfollow/re-follow of new account, nor does it include a way to actually set this "move" information, yet. Therefore, it's only one step towards solving #177

### DIFF
--- a/app/javascript/mastodon/features/account_timeline/components/moved_note.js
+++ b/app/javascript/mastodon/features/account_timeline/components/moved_note.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ImmutablePropTypes from 'react-immutable-proptypes';
+import { FormattedMessage } from 'react-intl';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+import AvatarOverlay from '../../../components/avatar_overlay';
+import DisplayName from '../../../components/display_name';
+
+export default class MovedNote extends ImmutablePureComponent {
+
+  static contextTypes = {
+    router: PropTypes.object,
+  };
+
+  static propTypes = {
+    from: ImmutablePropTypes.map.isRequired,
+    to: ImmutablePropTypes.map.isRequired,
+  };
+
+  handleAccountClick = e => {
+    if (e.button === 0) {
+      e.preventDefault();
+      this.context.router.history.push(`/accounts/${this.props.to.get('id')}`);
+    }
+
+    e.stopPropagation();
+  }
+
+  render () {
+    const { from, to } = this.props;
+    const displayNameHtml = { __html: from.get('display_name_html') };
+
+    return (
+      <div className='account__moved-note'>
+        <div className='account__moved-note__message'>
+          <div className='account__moved-note__icon-wrapper'><i className='fa fa-fw fa-suitcase account__moved-note__icon' /></div>
+          <FormattedMessage id='account.moved_to' defaultMessage='{name} has moved to:' values={{ name: <strong dangerouslySetInnerHTML={displayNameHtml} /> }} />
+        </div>
+
+        <a href={to.get('url')} onClick={this.handleAccountClick} className='detailed-status__display-name'>
+          <div className='detailed-status__display-avatar'><AvatarOverlay account={to} friend={from} /></div>
+          <DisplayName account={to} />
+        </a>
+      </div>
+    );
+  }
+
+}
--- a/app/lib/activitypub/adapter.rb
+++ b/app/lib/activitypub/adapter.rb
@@ -9,6 +9,7 @@ class ActivityPub::Adapter < ActiveModelSerializers::Adapter::Base
       {
         'manuallyApprovesFollowers' => 'as:manuallyApprovesFollowers',
         'sensitive'                 => 'as:sensitive',
+        'movedTo'                   => 'as:movedTo',
         'Hashtag'                   => 'as:Hashtag',
         'ostatus'                   => 'http://ostatus.org#',
         'atomUri'                   => 'ostatus:atomUri',
--- a/app/javascript/mastodon/features/account/components/header.js
+++ b/app/javascript/mastodon/features/account/components/header.js
@@ -7,6 +7,7 @@ import Motion from '../../ui/util/optional_motion';
 import spring from 'react-motion/lib/spring';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 import { autoPlayGif, me } from '../../../initial_state';
+import classNames from 'classnames';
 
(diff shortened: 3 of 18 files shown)


app/lib/activitypub/adapter.rb
```diff
@@ -9,6 +9,7 @@ class ActivityPub::Adapter < ActiveModelSerializers::Adapter::Base
       {
         'manuallyApprovesFollowers' => 'as:manuallyApprovesFollowers',
         'sensitive'                 => 'as:sensitive',
+        'movedTo'                   => 'as:movedTo',
```
ThibG: Should you really use ActivityStreams' namespace for that? I couldn't find any documentation for “movedTo”

Gargron: I made it up since it doesn't exist, but probably should exist. @cwebber is okay with it

ThibG: Alright, that's fine with me.

ThibG:
I'm wondering if it would make sense to redirect webfinger queries, too.

Gargron:
@ThibG 

> Gargron - Yesterday at 9:56 PM
@nightpool i just implemented a webfinger redirect for accounts and getting second thoughts. maybe the only redirect should be the public profile page. because a webfinger redirect would brick any look-up of all old content... plus it's a rabbit hole, there's a ton of ActivityPub endpoints, it's unclear whether you should be able to redirect them and them still be considered valid, and there's no way to tell if the new server has the same ActivityPub URI structure, etc. and so that would also brick stuff. redirecting public profile page and adding a "redirects" to the REST API so apps can show a special note, maybe that's all
nightpool - Yesterday at 10:01 PM
hm.
I don't see how redirecting the webfinger will break lookup of old content?
Gargron - Yesterday at 10:04 PM
it's all very confusing but say I am looking up "Hello world" by Gargron@mastodon.social, but Gargron@mastodon.social redirects to gargron@gonext.gg. Let's say we modified our webfinger code to even follow redirects in the first place. Now it creates "Hello world" as created by gargron@gonext.gg. You favourite it. That activity is sent to gargron@gonext.gg... but that post does not exist there, so it blackholes
That's just one example
nightpool - Yesterday at 10:04 PM
How are you looking up that post, by URI?
Gargron - Yesterday at 10:05 PM
Webfinger code not supporting redirects in all versions is even bigger. It'll just mean the content will not be processed in the first place.
nightpool - Yesterday at 10:05 PM
It should still have mastodon.social/users/gargron as the post actor
yeah the redirect is maybe a bigger concern.....
Gargron - Yesterday at 10:05 PM
It doesn't matter how you look up the post, when it checks the author it does webfinger.
nightpool - Yesterday at 10:07 PM
I don't think that's true
FetchRemoteStatusService uses attributedTo.
Gargron - Yesterday at 10:10 PM
when mastodon looks up an activitypub actor for the first time, i.e. there's no existing db record, it performs a webfinger verification
which leads to the scenario i described
there's also the question of... what if the record exists, but then webfinger cache is stale so we webfinger again, and it redirects. wtf do then
nightpool - Yesterday at 10:11 PM
Sure, okay
Gargron - Yesterday at 10:12 PM
thinking about all the ways in which this sort of thing can fuck up things is making my brain hurt

ThibG:
@Gargron yeah, indeed, that's what I was thinking. Since the content is still there, it kind of makes sense to not stop it from federating. The advantage is that it would be completely transparent to whichever software, OStatus or ActivityPub, supporting “movedTo” or not. The downside is that it would stop the content of the old profile from federating anymore, although it's still there… it's probably not worth it, I guess.

ThibG:
Another thing I'm wondering about and you haven't brought up in this PR is make every follower of the old account automatically send a follow request to the new one (and potentially unfollow the old one) once it's discovered to have moved.

Gargron:
@ThibG No that's not at all included in this PR.

ThibG:
Yes, I know, I was simply wondering if doing that would make sense.

app/javascript/mastodon/features/account_timeline/components/moved_note.js
```diff
@@ -0,0 +1,48 @@
...
+
+  render () {
+    const { from, to } = this.props;
+    const displayNameHtml = { __html: from.get('display_name_html') };
+
+    return (
+      <div className='account__moved-note'>
+        <div className='account__moved-note__message'>
+          <div className='account__moved-note__icon-wrapper'><i className='fa fa-fw fa-suitcase account__moved-note__icon' /></div>
+          <FormattedMessage id='account.moved_to' defaultMessage='{name} has moved to:' values={{ name: <strong dangerouslySetInnerHTML={displayNameHtml} /> }} />
```
ThibG: Wouldn't it better if the name was a link to the new account?

ClearlyClaire (approved):
I'm not very familiar with React.js, but everything else looks fine to me.

app/javascript/mastodon/features/account_timeline/components/moved_note.js
Gargron: That's the old name

Cam-B:
One step further. Awesome!

PR #5753, opened 2017-11-19
Fix NoMethodError at ActivityPub::FetchRemoteStatusService

abcang:
`actor` may be nil, and rarely `NoMethodError` occurs.


app/services/activitypub/fetch_remote_status_service.rb
```diff
@@ -18,7 +18,7 @@ def call(uri, id: true, prefetched_body: nil)
     actor = ActivityPub::TagManager.instance.uri_to_resource(actor_id, Account)
     actor = ActivityPub::FetchRemoteAccountService.new.call(actor_id, id: true) if actor.nil? || needs_update(actor)
 
-    return if actor.suspended?
+    return if actor.nil? || actor.suspended?
```
ykzts: `actor&.suspended?`

abcang: In that case, it is not returned when actor is nil.

ykzts: Ah, I see.

ykzts (approved):

PR #5757, opened 2017-11-19
Add logging of admin actions

Gargron:
image

- [x] suspend account
- [x] silence account
- [x] unsuspend account
- [x] unsilence account
- [x] memorialize account
- [x] disable login for account
- [x] enable login for account
- [x] confirm e-mail for account
- [x] disable 2fa for account
- [x] reset password for account
- [x] promote account role
- [x] demote account role
- [x] disable emoji
- [x] enable emoji
- [x] dismiss report
- [x] create domain block
- [x] remove domain block
- [x] create e-mail domain blacklisting
- [x] remove e-mail domain blacklisting
- [x] create emoji
- [x] remove emoji
- [x] remove status
- [x] update status (e.g. sensitive boolean)

Notes:

- For updates, promotes and demotes, it stores changed attributes in the log entry
- For creates and destroys, it stores all of the record's attributes in the log entry


nightpool:
these should show up when viewing the page for the account that an action is taken on, probably?

joshuap:
> How to store info about record in the log after it's deleted?

We've solved this by copying the table, i.e. if the table is `emoji`, a duplicate `deleted_emoji` table would be created. The target in the `admin_action_logs` would then reference the `DeletedEmoji` model rather than `Emoji`. `DeletedEmoji` may contain a subset of attributes (e.g., only the ones needed to display the deleted record in logs).

> Store some extra data e.g. from which to which role user is promoted/demoted?

For this, maybe add a `metadata` json/hstore column to `admin_action_logs`? Logfmt comes to mind, which represents logs using simple key-value pairs. User promotion would look something like:

```
action=promote account=admin target=@alice role=admin from_role=user
```

Representing logs like that can make them easier to export to other systems, too.

Gargron:
@joshuap I've looked up how acts_as_auditable does it - when a record is destroyed, it puts all of the records's attributes into a hstore. Although, especially with CustomEmoji, that might be kind of pointless since the image file is not stored alongside and therefore no recovery or display of that data would actually be useful.

If I don't do anything at all, the only thing I could do then is display "admin deleted CustomEmoji#345" which is kind of useless. So I think the middle ground is maybe, as you said, storing only attributes necessary for the barest minimum useful display, such as "admin deleted :cool_emoji:"

There's also acts_as_paranoid and similar approaches. I think that's gonna be useful for the statuses table, but probably not worth it for custom_emojis..? Separately from this PR, I want to make sure status deletions and account suspensions become reversible (with n days retention e.g. data is really deleted after 30 or 90 days instead of instantly)

joshuap:
I'm not a huge fan of `default_scope` in *acts_as_paranoid* (or in general), but I've used/contributed to the [paranoia](https://github.com/rubysherpas/paranoia) gem which does the same thing. It also has an option to skip the `default_scope` in case you want to use a more explicit approach when hiding deleted records.

I'm curious, how large can the statuses table grow on an instance connected to most of the federated network?

m4sk1n:
@Gargron, can you merge my PR for this branch? :)

app/controllers/admin/reports_controller.rb
```diff
@@ -25,12 +25,15 @@ def update
     def process_report
       case params[:outcome].to_s
       when 'resolve'
-        @report.update(action_taken_by_current_attributes)
+        @report.update!(action_taken_by_current_attributes)
+        log_action :dismiss, @report
```
nightpool: should probably be `:resolve` to match the existing terminology.

Gargron: My concern is that that *any* action "resolves" a report. In this line, it's a resolve without any *side effects* like silence/suspend. It'd be weird if we had "resolves" logged for these but not for the silence/suspend versions. Therefore, I used a more clear terminology, "dismiss", i.e. "resolve without doing anything"

Angristan:
Wouldn't it be nice to have the possibility to make this page public or at least partially, allowing the users to have more transparency regarding admins/mods actions? 
Just giving an idea, it's already a nice feature, so thanks for your work.

nightpool (dismissed):
Changes LGTM, but they should probably also include a Site Settings log type, right? I couldn't find logging for that anywhere.

app/controllers/admin/reports_controller.rb
nightpool: hmm. why *aren't* we logging both? I.e., suspending a user could be:

```
log_action :resolve, @report
log_action :suspend, @user
```

otherwise you'll have no idea what report a given action was taken in response to

Gargron:
The commit history of this PR is an awful mess because I wasn't able to rebase due to PRs made against this PR. :(

nullkal (commented):
We need to run `i18n-tasks normalize`.

nullkal (approved):

PR #5760, opened 2017-11-20
Soft-delete statuses, really delete after 30 days

Gargron:
Reasons:

1. Accountability:
  - if a report ends in a suspension, and statuses are gone instantly, it's very hard to tell what was reported and whether the suspend was justified
  - if an abuser sends abuse but then deletes the evidence we can't know what it was
2. Restoration:
  - if an account was suspended in error, it should be possible to restore it
  - if a user has second thoughts about their account deletion, they should be able to login again within 30 days to restore their account, as is common practice

---

- [x] Soft-deleted statuses don't show up anywhere
- [x] Sidekiq scheduler properly deletes soft-deleted statuses after 30 days
- [x] Suspending/deleting an account no longer immediately removes statuses, uploads, followers, follows, notifications, blocks, etc, instead it only soft-deletes statuses and marks account as suspended
- [x] Display deleted statuses in admin reports
- [ ] Add a way to differentiate account *deletions* from account *suspensions*
- [ ] Properly clean up followers, follows, notifications, blocks etc of a deleted/suspended account after a data retention period
- [ ] Unsuspending should restore the account **but without statuses that were deleted manually**

### DIFF
--- a/app/models/concerns/paranoid.rb
+++ b/app/models/concerns/paranoid.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Paranoid
+  extend ActiveSupport::Concern
+
+  included do
+    scope :without_deleted, -> { where('deleted_at IS NULL') }
+    scope :with_deleted,    -> { unscoped.recent }
+    scope :only_deleted,    -> { with_deleted.where('deleted_at IS NOT NULL') }
+  end
+
+  def soft_destroy
+    touch(:deleted_at)
+  end
+
+  alias soft_destroy! soft_destroy
+
+  def destroyed?
+    deleted_at.present? || super
+  end
+end
--- a/db/migrate/20171120001007_add_deleted_at_to_statuses.rb
+++ b/db/migrate/20171120001007_add_deleted_at_to_statuses.rb
@@ -1,0 +1,9 @@
+class AddDeletedAtToStatuses < ActiveRecord::Migration[5.1]
+  def change
+    add_column :statuses, :deleted_at, :datetime, null: true, default: nil
+
+    commit_db_transaction
+
+    add_index :statuses, :deleted_at, algorithm: :concurrently
+  end
+end
--- a/app/javascript/styles/mastodon/admin.scss
+++ b/app/javascript/styles/mastodon/admin.scss
@@ -277,6 +277,13 @@
   display: flex;
   margin-bottom: 10px;
 
+  &.deleted {
+    .entry,
+    .batch-checkbox {
+      opacity: 0.6;
+    }
+  }
+
   .activity-stream {
     flex: 2 0 0;
     margin-right: 20px;
--- a/app/models/concerns/status_threading_concern.rb
+++ b/app/models/concerns/status_threading_concern.rb
@@ -34,7 +34,7 @@ module StatusThreadingConcern
         SELECT statuses.id, statuses.in_reply_to_id, path || statuses.id
         FROM search_tree
         JOIN statuses ON statuses.id = search_tree.in_reply_to_id
-        WHERE NOT statuses.id = ANY(path)
+        WHERE NOT statuses.id = ANY(path) AND statuses.deleted_at IS NULL
       )
       SELECT id
       FROM search_tree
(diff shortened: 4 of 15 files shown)


ThibG:
I understand the concern, but I'm not too sure about those changes to be honest. At the very least, this should be made very clear to the end-user.
Also, 30 days seems rather long for the intended purposes.

clarcharr:
I agree; there should be a clear note in the UI that the account will be held for X days, and an email notification should be sent out on the initial soft-deletion and the actual deletion.

Additionally, I think that the retention period should be smaller and configurable; some instances may prefer a week or even no retention.

Cassolotl:
If the posts aren't visible for those 30 days I don't see a problem.

akihikodaki:
Any status update?

Technowix:
I would love this, would allow to be more reactive on issues while preventing them to spread like fire.
I would like to add : 
- Make post owner (and those mentioned ?) aware that his post have been deleted / in the process of deletion
- Maybe the ability to add a "reason for suppression", would help users getting moderated understand why this post is in the process of deletion, would also permit users who want to delete their post to provide a reason for this ^-^'

app/models/concerns/paranoid.rb
```diff
@@ -0,0 +1,21 @@
...
+    scope :without_deleted, -> { where('deleted_at IS NULL') }
+    scope :with_deleted,    -> { unscoped.recent }
+    scope :only_deleted,    -> { with_deleted.where('deleted_at IS NOT NULL') }
+  end
+
+  def soft_destroy
+    touch(:deleted_at)
+  end
+
+  alias soft_destroy! soft_destroy
```
ThisIsMissEm: shouldn't this be `touch!` such that it throws an error?

ThisIsMissEm: Shouldn't the method be soft_destroy! always?

db/migrate/20171120001007_add_deleted_at_to_statuses.rb
```diff
@@ -0,0 +1,9 @@
+class AddDeletedAtToStatuses < ActiveRecord::Migration[5.1]
+  def change
+    add_column :statuses, :deleted_at, :datetime, null: true, default: nil
```
ThisIsMissEm: isn't this migration unsafe? I think I got a warning for this sort of thing

ThisIsMissEm:
@Technowix said:
> Make post owner (and those mentioned ?) aware that his post have been deleted / in the process of deletion
> Maybe the ability to add a "reason for suppression", would help users getting moderated understand why this post is in the process of deletion, would also permit users who want to delete their post to provide a reason for this ^-^'

I'd definitely be up for adding email notifications to moderator actions (obviously there are times where silent action of good, but default to notify).

Likewise, I'd like to change all actions to require a "note" which is stored with the Account or Report whenever an action is performed, such that there's visibility as to why.

However, I think these are a completely different issue to soft-deletes and deserve issues in their own rights.

ThisIsMissEm:
Created #7127 for what's described in https://github.com/tootsuite/mastodon/pull/5760#issuecomment-381243784

clarcharr:
IMHO, the endgame for this should be:

1. Retention time should be instance-configurable, with an option for no retention at all.
2. Users should be made aware of the retention time on registration ideally, but at minimum whenever they delete something. Maybe there could be an option for "don't show me again" for toots so that they don't become a bother.
3. Instances should have the option to disable retention entirely for toots/accounts without any reports.
4. Admins should have the option to delete retained data either through a cleanup script or a UI.

Obviously, this shouldn't all be a blocker for an initial version, but I'd say at minimum it should be disable-able entirely in its first iteration. I can think of many instances for which retaining data would be less desirable, and simply having a complete on/off switch is best. Ideally, there should be the option to delete all retained data when it's turned off, but I wouldn't call that entirely necessary.

ThisIsMissEm:
@Gargron does the closed & deleted status here mean that this is no longer being implemented?

Gargron:
@ThisIsMissEm This PR got hella outdated so regardless of whether the functionality will be added, it won't be via this PR.

Soft-deletes locally are rather easy and straight-forward. But federating them in a way that makes them restorable? It's a mess. No clue what approach to take. Federating a "suspended" state that doesn't remove any data but advises it to be hidden might be a much better approach, as in the issue linked above

PR #3271, opened 2017-05-24
Add option to mute boosts from a user (#648)

spinda:
This implements #648, adding an additional relationship allowing users to mute particular users' boosts without muting them entirely. I've found myself wanting this quite a bit, as some of the users I follow are a bit over-eager with the boost button.

A new menu item is added to user profiles:

user profile menu screenshot

A menu item is also added to boosted statuses, to mute boosts from the user doing the boosting:

status menu screenshot

This is my first time contributing to Mastodon, or even writing Ruby, so please let me know where I've messed up. 🌈

Issue #648, opened 2017-02-20
Turn off boosts from specific user in home timeline


wolfteeth:
Similar to Twitter's "Disable Retweets" feature. I'm following User X, but I only want to see their toots in my home timeline, not the toots they have boosted.

I should still be able to see the boosted toots by directly visiting User X's profile.

jwisser:
Definitely seconded. I'd also like a preference to disable boosts universally. I find a lot more value in only seeing what people take the time to talk about themselves. Boosts/retweets encourage a lot of noise because they're so low-effort.

Cassolotl:
@giromide says in #1026:

> Options in the account would thus be:
> Mute @SomeAccount
> Mute boosts via @SomeAccount
> Block @SomeAccount
> Report @SomeAccount

akihikodaki:
What about favourites? Are they still notified even boosts are turned off?

Cassolotl:
@akihikodaki This isn't about *notifications*, this is about the boosts themselves. We want to be able to follow someone but only see their original posts, but not see the posts they boost.

erisdev:
please yes to this, it's nice to be able to turn them off altogether but i'd love to see boosts from my friends who always boost things i like. i just wanna hide the boosts from folx who always boost fundraising and political shit

wxcafe:
Duplicate of #133

PR #5762, opened 2017-11-20
Allow hiding of reblogs from followed users

aschmitz:
This adds a new entry to the account menu to allow users to hide future reblogs from a user (and then if they've done that, to show future reblogs instead).

This does not remove or add historical reblogs from/to the user's timeline; it only affects new statuses.

The API for this operates by sending a "reblogs" key to the follow endpoint. If this is sent when starting a new follow, it will be respected from the beginning of the follow relationship (even if the follow request must be approved by the followee). If this is sent when a follow relationship already exists, it will simply update the existing follow relationship. As with the notification muting, this will now return an object (`{reblogs: [true|false]}`) or `false` for each follow relationship when requesting relationship information for an account. This should cause few issues due to an object being truthy in many languages, but some modifications may need to be made in pickier languages.

Database changes: adds a `show_reblogs` column (default `true`, non-nullable) to the `follows` and `follow_requests` tables. Because these are non-nullable, we use the existing MigrationHelpers to perform this change without locking those tables, although the tables are likely to be small anyway.

Tests included.

See also <https://github.com/glitch-soc/mastodon/pull/212>.


aschmitz:
(I think this closes #648, and replaces #3271.)

app/javascript/mastodon/reducers/accounts_counters.js
```diff
@@ -126,6 +126,7 @@ export default function accountsCounters(state = initialState, action) {
   case STATUS_FETCH_SUCCESS:
     return normalizeAccountFromStatus(state, action.status);
   case ACCOUNT_FOLLOW_SUCCESS:
+    if (action.alreadyFollowing) { return state; }
```
aschmitz: eslint hates this line. I can make it ternary if you prefer, but that seemed *less* clear, not more.

Gargron:
>(I think this closes #648, and replaces #3271.)

To have it automatically do this, write "fix #648, resolve #3271" in the body of your PR (or any commit message)

app/javascript/mastodon/reducers/accounts_counters.js
Gargron: You can make it a function call instead

app/lib/feed_manager.rb
```diff
@@ -165,7 +165,9 @@ def filter_from_home?(status, receiver_id)
       should_filter &&= status.account_id != status.in_reply_to_account_id                                               # and it's not a self-reply
       return should_filter
     elsif status.reblog?                                                                                                 # Filter out a reblog
-      should_filter   = Block.where(account_id: status.reblog.account_id, target_account_id: receiver_id).exists?        # or if the author of the reblogged status is blocking me
+      src_id = status.account_id
```
Gargron: Why is this a variable?

db/migrate/20171028221157_add_reblogs_to_follows.rb
```diff
@@ -0,0 +1,21 @@
+require Rails.root.join('lib', 'mastodon', 'migration_helpers')
+
+class AddReblogsToFollows < ActiveRecord::Migration[5.1]
+  include Mastodon::MigrationHelpers
+
+  safety_assured do
```
Gargron: This wrapping should not be necessary

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -13,9 +13,11 @@ def show
   end
 
   def follow
-    FollowService.new.call(current_user.account, @account.acct)
+    reblogs_arg = { reblogs: params[:reblogs] }
```
Gargron: This could be a private method

app/javascript/mastodon/reducers/accounts_counters.js
Gargron: If blocks within a switch are kind of confusing. A ternary probably *is* better, and if it's less readable I think you can multiline it.

app/models/concerns/account_interactions.rb
```diff
@@ -66,8 +74,12 @@ def follow_mapping(query, field)
     has_many :domain_blocks, class_name: 'AccountDomainBlock', dependent: :destroy
   end
 
-  def follow!(other_account)
-    active_relationships.find_or_create_by!(target_account: other_account)
+  def follow!(other_account, reblogs: nil)
+    reblogs = true if reblogs.nil?
```
Gargron: Can this not be part of the argument's default value..?

app/services/follow_service.rb
```diff
@@ -6,25 +6,38 @@ class FollowService < BaseService
   # Follow a remote user, notify remote user about the follow
   # @param [Account] source_account From which to follow
   # @param [String, Account] uri User URI to follow in the form of username@domain (or account record)
-  def call(source_account, uri)
+  # @param [true, false, nil] reblogs Whether or not to show reblogs, defaults to true
+  def call(source_account, uri, reblogs: nil)
+    reblogs = true if reblogs.nil?
```
Gargron: Same here, doesn't this mean the default value of the argument should be true and not nil?

app/lib/feed_manager.rb
Gargron: Why is this a variable? Seems like it's only used in one place.

Gargron (changes requested):

app/controllers/api/v1/accounts_controller.rb
aschmitz: In theory, yes? I already think it's not great because we're using it as an options argument as well as a return value (mostly for convenience and line size, because it's the same in both cases for the moment). I don't think it really has any purpose being a method, but I can make it one if you want.

app/lib/feed_manager.rb
aschmitz: Mostly because if I don't do this, it pushes the comments column out even farther, which seemed undesirable, but I can do whatever you'd like with it.

db/migrate/20171028221157_add_reblogs_to_follows.rb
aschmitz: Hm, right you are. I thought strong_migrations complained otherwise, guess not(? or at least not when I just tried it).

app/models/concerns/account_interactions.rb
aschmitz: In fact no, apparently. (Well, it could be, but we still need to catch the nil case: it may come in from params as nil.)

app/services/follow_service.rb
aschmitz: Same as the other case.

aschmitz:
Tests pass, minus the Code Climate issue about `class_methods` in AccountInteractions being too big of a block, which I'm inclined to ignore unless you want to massively refactor that. (Passing a proc to `follow_mapping` isn't a terrible idea, but it also won't save many lines, and I think the intent of that check is something other than what it's catching here.)

I think I've addressed the review comments, but let me know if there's something you think still needs changing. Thanks!

Gargron (approved):

tateisu:
how about faking follow status ?
can we call "hide-boost" request while follow status is faked?

PR #5763, opened 2017-11-20
Fixed duplicating URL of photo type of oEmbed

abcang:
The photo of oEmbed will change the URL. The URL obtained with OEmbed::Providers may already be saved. So, to avoid duplication of URLs, the preview card needs to be checked again.


Gargron (approved):

akihikodaki:
I do not think this is a correct solution. The problem is that the different kind of values, namely the URL of the document bound to the oEmbed representation, and the URL of the photo in the oEmbed, are saved as the same attribute. This makes PreviewCard caching fail.

PR #5735, opened 2017-11-17
Add “staff” badge to admin user pages

ykzts:

PR #5764, opened 2017-11-20
i18n: (zh-CN) Add translations for #5746 & #5735

ghost:


Gargron (approved):

config/locales/zh-CN.yml
```diff
@@ -243,6 +244,9 @@ zh-CN:
         open:
           desc_html: 允许任何人建立一个帐户
           title: 开放注册
+      show_staff_badge:
```
Artoria2e5: I guess it's more like admin+moderator for stuff here? Ah anyway we call both of them something 管 something. That's fine.

Artoria2e5 (commented):
👍

config/locales/zh-CN.yml
SerCom-KC: Yeah this does sound confusing if taking it seriously but I haven't got a better idea yet

PR #5775, opened 2017-11-21
Add embed columns to preview cards

akihikodaki:
This resolves the problem described with https://github.com/tootsuite/mastodon/pull/5763#issuecomment-345930146.

Note that the migration included in this change rebuilds the entire `preview_cards` table and takes long.
I'm not sure it is tolerated, but it is unlikely. If not, I will introduce additional changes.

### DIFF
--- a/app/javascript/mastodon/features/status/components/card.js
+++ b/app/javascript/mastodon/features/status/components/card.js
@@ -101,7 +101,7 @@ export default class Card extends React.PureComponent {
         onClick={this.handlePhotoClick}
         role='button'
         tabIndex='0'
-        src={card.get('url')}
+        src={card.get('embed_url')}
         alt={card.get('title')}
         width={card.get('width')}
         height={card.get('height')}
--- a/app/models/preview_card.rb
+++ b/app/models/preview_card.rb
@@ -21,6 +21,7 @@
 #  height             :integer          default(0), not null
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
+#  embed_url          :string           default(""), not null
 #
 
 class PreviewCard < ApplicationRecord
--- a/app/serializers/rest/preview_card_serializer.rb
+++ b/app/serializers/rest/preview_card_serializer.rb
@@ -6,7 +6,7 @@ class REST::PreviewCardSerializer < ActiveModel::Serializer
   attributes :url, :title, :description, :type,
              :author_name, :author_url, :provider_name,
              :provider_url, :html, :width, :height,
-             :image
+             :image, :embed_url
 
   def image
     object.image? ? full_asset_url(object.image.url(:original)) : nil
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -74,9 +74,6 @@ class FetchLinkCardService < BaseService
 
     return false unless response.respond_to?(:type)
 
-    # The photo will change the URL. So, to avoid duplication of URLs, PreviewCard needs to be checked again.
-    @card = PreviewCard.find_by(url: response.url) || @card if response.type == 'photo'
-
     @card.type          = response.type
     @card.title         = response.respond_to?(:title)         ? response.title         : ''
     @card.author_name   = response.respond_to?(:author_name)   ? response.author_name   : ''
@@ -90,9 +87,9 @@ class FetchLinkCardService < BaseService
     when 'link'
       @card.image = URI.parse(response.thumbnail_url) if response.respond_to?(:thumbnail_url)
     when 'photo'
-      @card.url    = response.url
-      @card.width  = response.width.presence  || 0
-      @card.height = response.height.presence || 0
+      @card.embed_url = response.url
+      @card.width     = response.width.presence  || 0
+      @card.height    = response.height.presence || 0
     when 'video'
       @card.width  = response.width.presence  || 0
       @card.height = response.height.presence || 0
--- a/db/migrate/20171130000000_add_embed_url_to_preview_cards.rb
+++ b/db/migrate/20171130000000_add_embed_url_to_preview_cards.rb
@@ -1,0 +1,18 @@
+require Rails.root.join('lib', 'mastodon', 'migration_helpers')
+
+class AddEmbedUrlToPreviewCards < ActiveRecord::Migration[5.1]
+  include Mastodon::MigrationHelpers
+
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_column_with_default :preview_cards, :embed_url, :string, default: '', allow_null: false
+    end
+  end
+
+  def down
+    execute "UPDATE preview_cards SET url=embed_url WHERE embed_url!=''"
+    remove_column :preview_cards, :embed_url
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171129172043) do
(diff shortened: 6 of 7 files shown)


akihikodaki:
codeclimate is complaining too many lines in a method, but this change just adds newlines and does not add statements. It is fine to ignore.

config/initializers/strong_migrations.rb
```diff
@@ -1,3 +1,3 @@
 # frozen_string_literal: true
 
-StrongMigrations.start_after = 20170924022025
+StrongMigrations.start_after = 20171120000000
```
abcang: 🤔

db/migrate/20171120000000_add_embed_columns_to_preview_cards.rb
```diff
@@ -0,0 +1,32 @@
...
+                         .where({ preview_cards: { type: :photo } })
+                         .reorder(nil)
+                         .group(:id)
+                         .pluck(:id)
+
+        PreviewCard.where(type: :photo).delete_all
+      end
+    end
+
+    add_column :preview_cards, :embed_url, :string, default: '', null: false
```
abcang: I think that we should use `Mastodon::MigrationHelpers#add_column_with_default` that can add columns without locking.

db/migrate/20171120000000_add_embed_columns_to_preview_cards.rb
```diff
@@ -0,0 +1,32 @@
...
+    rename_column :preview_cards, :html, :embed_html
```
abcang: What is the reason for changing column names? If you change the column name, saving the preview card will fail until restarting the process.

abcang (changes requested):

akihikodaki:
c692082600268e2bd12665dd5928e6a263900993 eliminates blocking when migrating with up direction.

akihikodaki:
deed5c1a79cadc9d1d8dac4c480f9eaa7e0e8492: rebased to e84fecb7e97851ed56f4d954e2d68128bb87da37.

db/migrate/20171121000000_prepare_embed_columns_in_preview_cards.rb
```diff
@@ -0,0 +1,31 @@
...
+  include Mastodon::MigrationHelpers
+
+  def up
+    safety_assured do
+      cleanup_concurrent_column_rename :preview_cards, :height, :embed_height
+      cleanup_concurrent_column_rename :preview_cards, :width, :embed_width
+      cleanup_concurrent_column_rename :preview_cards, :html, :embed_html
+    end
+
+    status_ids = Status.joins(:preview_cards)
```
abcang: There is no need to join Status. I think that this query is enough.

```
PreviewCard.where(type: 'photo').joins('JOIN preview_cards_statuses ON preview_cards_statuses.preview_card_id = preview_cards.id').pluck(:status_id)
```

```
SELECT "status_id" FROM "preview_cards" JOIN preview_cards_statuses ON preview_cards_statuses.preview_card_id = preview_cards.id WHERE "preview_cards"."type" = 1
```

db/migrate/20171120000000_add_embed_columns_to_preview_cards.rb
```diff
@@ -0,0 +1,28 @@
...
+          rename_column_concurrently :preview_cards, :html, :embed_html
+          rename_column_concurrently :preview_cards, :width, :embed_width
+          rename_column_concurrently :preview_cards, :height, :embed_height
+        end
+
+        dir.down do
+          cleanup_concurrent_column_rename :preview_cards, :height, :embed_height
+          cleanup_concurrent_column_rename :preview_cards, :width, :embed_width
+          cleanup_concurrent_column_rename :preview_cards, :html, :embed_html
+          execute 'UPDATE preview_cards SET url=embed_url WHERE embed_url IS NOT NULL'
```
abcang: Since embed_url is not null, it seems that all urls will be changed. I think that should be limited to photo.

db/migrate/20171121000000_prepare_embed_columns_in_preview_cards.rb
akihikodaki: You are right. However, I still prefer a query without any raw SQL code. It is not performance-critical and fine to have the little overhead.

abcang: Ok, I see.

db/migrate/20171124000000_prepare_embed_columns_in_preview_cards.rb
```diff
@@ -0,0 +1,25 @@
...
+    safety_assured do
+      cleanup_concurrent_column_rename :preview_cards, :height, :embed_height
+      cleanup_concurrent_column_rename :preview_cards, :width, :embed_width
+      cleanup_concurrent_column_rename :preview_cards, :html, :embed_html
+    end
+
+    status_ids = Status.joins(:preview_cards)
+                       .where({ preview_cards: { type: :photo } })
+                       .reorder(nil)
+                       .group(:id)
```
abcang: I think that it is better to use `distinct` instead of `group`.

```
Status.joins(:preview_cards).where({ preview_cards: { type: :photo } }).reorder(nil).distinct.pluck(:id)
```

```
SELECT DISTINCT "statuses"."id" FROM "statuses"
INNER JOIN "preview_cards_statuses" ON "preview_cards_statuses"."status_id" = "statuses"."id"
INNER JOIN "preview_cards" ON "preview_cards"."id" = "preview_cards_statuses"."preview_card_id"
WHERE "preview_cards"."type" = 1
```

abcang (approved):

Gargron:
I understand embed_url (I think) but why rename html, width and height?

akihikodaki:
Because all of them are for embedding while other attributes are not but kind of summary.

akihikodaki:
My explanation may be a little insufficient so here is a bit more:
PostgreSQL does not allow to have an object-oriented representation, so we need to utilize the identifiers for the purpose.
Here, `PreviewCard` ideally can have these accessors: `id`, `url`, `title`, `description`, `image`, `author`, `provider`, `created_at`, `updated_at`, and `embed` which is added with this pull request.
`author` and `provider` is an object with two attributes. `embed` is polymorphic, and should be an object whose super class has `width` and `height`.
We are obligated to clarify such an intention especially when the underlying system is not that sophisticated.

Gargron:
>We are obligated to clarify such an intention especially when the underlying system is not that sophisticated.

Okay, I understand your reasoning, but I still believe that adding a complicated long-running migration just to improve developer-visible column naming is unnecessary. Personally, embed_width and width makes no difference to me. You could just add a code comment instead.

db/migrate/20171123000000_add_embed_columns_to_preview_cards.rb
```diff
@@ -0,0 +1,30 @@
...
+class AddEmbedColumnsToPreviewCards < ActiveRecord::Migration[5.1]
+  include Mastodon::MigrationHelpers
+
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_column_with_default :preview_cards, :embed_url, :string, default: '', allow_null: false
+
+      rename_column_concurrently :preview_cards, :html, :embed_html
```
Gargron: I want to get rid of the column renames.

Gargron (changes requested):

akihikodaki:
86ac0b047abd0b0263b8ee19cd1bad90ab617182: removed column renames and rebased to 9302369aa5648bab42a2cd6078607d7cbaa92458.

Gargron (approved):

PR #5777, opened 2017-11-21
Allow to open a modal for embedded photo

akihikodaki:
screenshot-2017-11-21 mastodon dev


akihikodaki:
For anyone interested in testing this change: the screenshot shows a picture on pixiv, but pictures on pixiv no longer looks so nice because it reports height incorrectly and height is now used for the layout of the modal. Use other standard-conforming oEmbed providers for testing.

ykzts (approved):
nice!

PR #5780, opened 2017-11-22
Remove validation tests

ysksn:


akihikodaki (changes requested):
I agree most of the validations are obvious and specs for them seem redundant. (though I still prefer to have specs for them, that is not really a strong opinion.)

However, the behaviors of the validations for regular expressions are not necessarily obvious. I'm not for removing specs for them,

Gargron:
I think the specs should stay, because someone could change validations on the model and mess them up. I have messed up like that in 2.3.2 because we didn't have a good test spec for that particular validation. Closing.

PR #5783, opened 2017-11-22
Delegate #version_number and #source_url to Mastodon::Version

ysksn:
- Create an alias `Mastodon::Version.version_number` for `Mastodon::Version.to_s`
- Delegate `InstancePresenter#version_number` and `InstancePresenter#source_url` to `Mastodon::Version`

Even though `Mastodon::Version != Mastodon::Version.to_s`, I believe it's OK because It is always used like this so far:

```ruby
"#{Mastodon::Version}"   # => implicitly calling .to_s
```
https://github.com/tootsuite/mastodon/blob/master/app/views/about/show.html.haml#L71

### DIFF
--- a/app/presenters/instance_presenter.rb
+++ b/app/presenters/instance_presenter.rb
@@ -12,6 +12,12 @@ class InstancePresenter
     to: Setting
   )
 
+  delegate(
+    :version_number,
+    :source_url,
+    to: 'Mastodon::Version'
+  )
+
   def contact_account
     Account.find_local(Setting.site_contact_username)
   end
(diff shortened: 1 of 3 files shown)


akihikodaki (changes requested):
I'm not for aliasing `to_s` as `version_number` because the module has `to_a`, which also represents the version number, and even sounds closer to the idea the name suggests. It is a number, not digits or a decimal, right?

:+1: for `source_url`.

Gargron:
I agree with Aki, I don't see much point in changing this code, since this PR has been inactive since I'm closing.

PR #5789, opened 2017-11-23
status: preserve visibility attribute when reblogging (infoleak fix)

kaniini:
While investigating an infoleak problem specific to mastodon-hardened, I observed that mastodon always sets post visibility to public for reblogs.

There are no logical cases where a reblog should not match the visibility of it's parent.

This should fix *all* remaining visibility-related Mastodon OStatus infoleaks.

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -278,6 +278,7 @@ class Status < ApplicationRecord
 
   def set_visibility
     self.visibility = (account.locked? ? :private : :public) if visibility.nil?
+    self.visibility = reblog.visibility if reblog?
     self.sensitive  = false if sensitive.nil?
   end
 


Gargron (approved):
For clarification this only affects public<->unlisted as others are not allowed be reblogged at all. Still, good.

kaniini:
I suspect the code climate thing is complaining about the pre-existing use of ternary operator here.

unarist:
Currently we don't show reblogs on public timelines (local/federated), so it seems like :unlisted even if original post was posted as :public. So, why don't we set :unlisted always?

kaniini:
`:unlisted` always would solve the issue in mastodon directly, this approach solves the issue even in cases where there are additional visibility settings (such as mastodon-hardened's `:local` visibility setting).

Gargron:
Rebase on current master and the codeclimate issue should go away.

PR #5793, opened 2017-11-23
add index on stream_entries table

takayamaki:
This pull request is adding index for below statement .
`SELECT  "stream_entries".* FROM "stream_entries" WHERE "stream_entries"."activity_type" = ? AND "stream_entries"."account_id" = ? AND "stream_entries"."hidden" = ? ORDER BY "stream_entries"."id" DESC LIMIT ?`

This statement executed 67,112 times.

## before
```
                                                                               QUERY PLAN                                                                               
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..754.44 rows=20 width=48) (actual time=54.970..167.629 rows=20 loops=1)
   ->  Index Scan Backward using index_stream_entries_on_id on stream_entries  (cost=0.43..450939.73 rows=11961 width=48) (actual time=54.968..167.617 rows=20 loops=1)
         Filter: ((NOT hidden) AND ((activity_type)::text = 'Status'::text) AND (account_id = 11109))
         Rows Removed by Filter: 298624
 Planning time: 0.109 ms
 Execution time: 167.658 ms
(6 rows)
```
## after
```
                                                                   QUERY PLAN                                                                   
------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..167.31 rows=20 width=48) (actual time=1.718..5.579 rows=20 loops=1)
   ->  Index Scan Backward using test3 on stream_entries  (cost=0.43..103891.09 rows=12451 width=48) (actual time=1.717..5.572 rows=20 loops=1)
         Index Cond: ((account_id = 11109) AND ((activity_type)::text = 'Status'::text))
         Filter: (NOT hidden)
 Planning time: 0.126 ms
 Execution time: 5.603 ms
(6 rows)
```

### DIFF
--- a/db/migrate/20171129172043_add_index_on_stream_entries.rb
+++ b/db/migrate/20171129172043_add_index_on_stream_entries.rb
@@ -1,0 +1,7 @@
+class AddIndexOnStreamEntries < ActiveRecord::Migration[5.1]
+  def change
+    commit_db_transaction
+    add_index :stream_entries, [:account_id, :activity_type, :id], algorithm: :concurrently
+    remove_index :stream_entries, name: :index_stream_entries_on_account_id
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171125190735) do
+ActiveRecord::Schema.define(version: 20171129172043) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -431,7 +431,7 @@ ActiveRecord::Schema.define(version: 20171125190735) do
     t.datetime "updated_at", null: false
     t.boolean "hidden", default: false, null: false
     t.bigint "account_id"
-    t.index ["account_id"], name: "index_stream_entries_on_account_id"
+    t.index ["account_id", "activity_type", "id"], name: "index_stream_entries_on_account_id_and_activity_type_and_id"
     t.index ["activity_id", "activity_type"], name: "index_stream_entries_on_activity_id_and_activity_type"
   end
 


kaniini (approved):
this sped up timeline queries significantly on my instance

Gargron:
Isn't a query like this used only on the Atom feed and only when you're paginating it? How could this have improved timelines @kaniini?

kaniini:
Sorry, I meant when loading some specific account's timeline.

db/migrate/20171123163635_add_index_on_stream_entries.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexOnStreamEntries < ActiveRecord::Migration[5.1]
+  def change
+    commit_db_transaction
+    add_index :stream_entries, [:id, :account_id, :activity_type], algorithm: :concurrently
```
akihikodaki: I think you should put `id` at the last for sorting. By putting `account_id` at the very beginning you can also cover `index_stream_entries_on_account_id` and remove it.

akihikodaki (dismissed):

akihikodaki:
Requesting a review from @abcang; he has suggested a similar change for Pawoo, but it is also concerning the idea I suggested in https://github.com/tootsuite/mastodon/pull/5793#pullrequestreview-78996361.

Gargron:
>I meant when loading some specific account's timeline.

I still don't think this could have any effect on that? As far as I remember account timelines use `statuses` directly without querying through `stream_entries` first - only the OStatus Atom feed does that, and it's never accessed programmatically from Mastodon code.

abcang:
I also recognized that stream_entries is currently used only for atom feed. However, instances that do not correspond to ActivityPub are accessing atom feed, so I think that it will be effective for instances connected to such instances. Also, I think that it is more effective for instances that have many stream_entries like Pawoo.

takayamaki:
Thank you for review.
It is more faster.
```
mastodon=> explain analyze SELECT "stream_entries".* FROM "stream_entries" WHERE "stream_entries"."activity_type" = 'Status' AND "stream_entries"."account_id" = ***** AND "stream_entries"."hidden" = 'f' ORDER BY "stream_entries"."id" DESC L
IMIT 20;
                                                                  QUERY PLAN                                                                   
-----------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..78.54 rows=20 width=48) (actual time=0.037..0.078 rows=20 loops=1)
   ->  Index Scan Backward using test6 on stream_entries  (cost=0.43..50676.88 rows=12975 width=48) (actual time=0.035..0.067 rows=20 loops=1)
         Index Cond: ((account_id = 1) AND ((activity_type)::text = 'Status'::text))
         Filter: (NOT hidden)
 Planning time: 0.178 ms
 Execution time: 0.100 ms
(6 rows)

mastodon=>
```

takayamaki:
This statement was executed 45974 times in the last 6 days at imastodon.net.
If average execution time is 100 ms, it takes 12.77 minutes per day.

I think this is still necessary.

abcang (approved):

PR #5796, opened 2017-11-23
Update terminology (e-mail blacklist)

mkljczk:
imho looks better

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -217,13 +217,13 @@ en:
       undo: Undo
     email_domain_blocks:
       add_new: Add new
-      created_msg: Email domain block successfully created
+      created_msg: Successfully added e-mail domain to blacklist
       delete: Delete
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -217,13 +217,13 @@ en:
       undo: Undo
     email_domain_blocks:
       add_new: Add new
-      created_msg: Email domain block successfully created
+      created_msg: Successfully added e-mail domain to blacklist
       delete: Delete
-      destroyed_msg: Email domain block successfully deleted
+      destroyed_msg: Sccessfully deleted e-mail domain from blacklist
```
Gargron: Typo

Gargron (changes requested):

config/locales/en.yml
m4sk1n: oops…

Gargron (approved):

PR #5814, opened 2017-11-25
Add consumable invites

### LINKED ISSUE #977: Invite-Only mode for registrations
Currently, I can either open or close registration on my mastodon instance. When registration is open, it looks like I can blacklist certain domains from signing up, but can't otherwise limit registration. I'd like to be able to close registrations except for a whitelist of email addresses, to allow my family members to register without letting randos register.

* * * *

Gargron:
- By default configured to only be available to admins
- Admin setting to make it available to mods, normal users, or completely disable
- Generate 8-char readable code (link) that enables to pass through closed registrations
- Configurable optional limit by max uses or expiration time for each code
- Users can delete their invite codes
- Admin UI to view all codes, delete

Normal user UI:

image

Admin UI:

image

Admin settings UI:

image

- [x] Invites structure
- [x] Consuming invites to sign up even when registrations are closed
- [x] UI for creating invites

___

- Fix #977
- Resolve #5549


ykzts:
Related to #977?

app/models/user.rb
```diff
@@ -47,12 +48,14 @@ class User < ApplicationRecord
...
 
   belongs_to :account, inverse_of: :user, required: true
+  belongs_to :invite, counter_cache: :uses
   accepts_nested_attributes_for :account
 
   has_many :applications, class_name: 'Doorkeeper::Application', as: :owner
 
   validates :locale, inclusion: I18n.available_locales.map(&:to_s), if: :locale?
   validates_with BlacklistedEmailValidator, if: :email_changed?
+  validates_with InviteCodeValidator
```
nightpool: This should only validate when the record is being created, so records don't start erroring once the invite expires

app/models/invite.rb
```diff
@@ -0,0 +1,42 @@
...
+#
+# Table name: invites
+#
+#  id                         :integer          not null, primary key
+#  user_id                    :integer
+#  code                       :string           default(""), not null
+#  expires_at                 :datetime
+#  max_uses                   :integer
+#  uses                       :integer          default(0), not null
+#  default_follow_account_ids :integer          is an Array
```
nightpool: Unused?

nightpool (dismissed):

app/controllers/invites_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    else
+      @invites = Invite.where(user: current_user)
+      render :index
+    end
+  end
+
+  def destroy
+    @invite = Invite.where(user: current_user).find(params[:id])
+    authorize @invite, :destroy?
+    @invite.destroy!
```
nightpool: instead of destroying the record, this should probably just set it expires_at=0 or something so that it's still auditable.

otherwise we have the situation where someone makes 100 accounts and deletes the invites so they can't be traced back to them.

config/locales/en.yml
```diff
@@ -424,6 +429,25 @@ en:
...
     upload: Upload
   in_memoriam_html: In Memoriam.
+  invites:
+    delete: Delete
+    expires_in:
+      '1800': 30 minutes
+      '21600': 6 hours
+      '3600': 1 hour
+      '43200': 12 hours
+      '86400': 1 day
```
nightpool: shouldn't this just be done with an interval thing?

app/views/invites/_form.html.haml
```diff
@@ -0,0 +1,9 @@
+= simple_form_for(@invite) do |f|
+  = render 'shared/error_messages', object: @invite
+
+  .fields-group
+    = f.input :max_uses, wrapper: :with_label, collection: [1, 5, 10, 25, 50, 100], label_method: lambda { |num| I18n.t('invites.max_uses', count: num) }, prompt: I18n.t('invites.max_uses_prompt')
+    = f.input :expires_in, wrapper: :with_label, collection: [30.minutes, 1.hour, 6.hours, 12.hours, 1.day].map(&:to_i), label_method: lambda { |i| I18n.t("invites.expires_in.#{i}") }, prompt: I18n.t('invites.expires_in_prompt')
```
nightpool: 7 days is probably useful here. 6 hours is probably not super useful?

config/locales/en.yml
Gargron: No, we overwrote the rails strings to be super short for relative time. Unless you know a better way

ykzts (approved):

PR #5815, opened 2017-11-25
Do not filter the status collection after muting and blocking

### LINKED ISSUE #5469: Mutes and blocks hide statuses even from the targets' profiles in the same session
On the Web interface, a mute hides statuses from a particular user. It even hides from user timeline.
That is undesired in a situation that
* a user thinks muted person is too noisy for home timeline but
* the user is still interested in the person and want to see the statuses when it is convenient for him.

* * * *

akihikodaki:
Filtering the status collection wipes out even the profiles of muted and blocked accounts. However, the behavior is inconsistent with the server-side behavior.
This resolves #5469.

### DIFF
--- a/app/javascript/mastodon/reducers/statuses.js
+++ b/app/javascript/mastodon/reducers/statuses.js
@@ -23,10 +23,6 @@ import {
   TIMELINE_EXPAND_SUCCESS,
 } from '../actions/timelines';
 import {
-  ACCOUNT_BLOCK_SUCCESS,
-  ACCOUNT_MUTE_SUCCESS,
-} from '../actions/accounts';
-import {
   NOTIFICATIONS_UPDATE,
   NOTIFICATIONS_REFRESH_SUCCESS,
   NOTIFICATIONS_EXPAND_SUCCESS,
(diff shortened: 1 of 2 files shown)


Gargron (approved):

app/javascript/mastodon/reducers/statuses.js
```diff
@@ -139,9 +127,6 @@ export default function statuses(state = initialState, action) {
     return normalizeStatuses(state, action.statuses);
   case TIMELINE_DELETE:
     return deleteStatus(state, action.id, action.references);
-  case ACCOUNT_BLOCK_SUCCESS:
```
Gargron: You need to remove these from imports too (CodeClimate)

ykzts (dismissed):
https://github.com/tootsuite/mastodon/pull/5815/files#diff-cc35b0b54a87d744fce947f1209592a0R25

It is not used. Please delete it.

akihikodaki:
b448e6d6df91531f5bb56bb82757248d06df4e72 applies https://github.com/tootsuite/mastodon/pull/5815#pullrequestreview-79004399.

PR #5817, opened 2017-11-25
Change the disclaimer under the sign up form

Sylvhem:
- Change the disclaimer under the sign up form. This change should make the message more clear and coherent. See [this discussion on Mastodon](https://dev.glitch.social/@Sylvhem/99062862188384217).
- Update the French translation accordingly.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -333,7 +333,7 @@ en:
     warning: Be very careful with this data. Never share it with anyone!
     your_token: Your access token
   auth:
-    agreement_html: By signing up you agree to <a href="%{rules_path}">our terms of service</a> and <a href="%{terms_path}">privacy policy</a>.
+    agreement_html: By signing up you confirm to have read <a href="%{rules_path}">our extended informations page</a> and agree to <a href="%{terms_path}">our terms of service</a>.
     change_password: Security
     delete_account: Delete account
     delete_account_html: If you wish to delete your account, you can <a href="%{path}">proceed here</a>. You will be asked for confirmation.
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -333,7 +333,7 @@ en:
     warning: Be very careful with this data. Never share it with anyone!
     your_token: Your access token
   auth:
-    agreement_html: By signing up you confirm to have read <a href="%{rules_path}">our extended informations page</a> and agree to <a href="%{terms_path}">our terms of service</a>.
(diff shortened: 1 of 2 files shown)


nightpool (changes requested):
I agree with noelle, this should probably be "you agree to follow" and not "confirm to have read", at least for the instance rules. 

Also "confirm to have read" is not a very natural construction in English. If others agree, I think the best wording is maybe something like "you agree to follow the rules of the instance, and have read and understand the terms of service."

That's a little wordy though....

Sylvhem:
Why not "By signing up, you agree to follow the rules of the instance and our terms of service." then? It's less wordy.

akihikodaki:
> Why not "By signing up, you agree to follow the rules of the instance and our terms of service." then? It's less wordy.

I'm :+1: for @nightpool's opinion. Service providers and other users have to follow the terms of service, and you can know that by reading them. (e.g. _What is your data retention policy?_ section)
It is not just to restrict users' behaviors, but also to provide some assurances for them.

Sylvhem:
> Service providers and other users have to follow the terms of service, and you can know that by reading them.

But if you agree to the terms of service, it implies that you read them or that someone else read theme for you, don't you think? Or am I missing something?

akihikodaki:
> But if you agree to the terms of service, it implies that you read them or that someone else read theme for you, don't you think? Or am I missing something?

We need clear descriptions rather than ones which requires such an inference.

Sylvhem:
> We need clear descriptions rather than ones which requires such an inference.

I agree that our descriptions must be clear, but I think this inference is simple enough. Logically, you *can't* agree with a text if you didn't read it, or didn't have it read by someone else.
Also, don't you think that reducing the number of times we use terms such as "read" will make the overall experience nicer for people with visual impairments who can't read by themselves? (I really don't know about this, it's just a supposition.)
Anyway, it's not really a hill I want to die on. But since there doesn't seem to have a consensus on that, maybe we should ask other people what they think of this issue.

mattcoxonline:
I'm sat on the same side of the fence as @Sylvhem on this one; short and sweet is better, and more to the point.

> By signing up, you agree to follow the rules of the instance and our terms of service.

Or even flipping the two agreements:

> By signing up, you agree with our terms of service and to follow the rules of the instance.

I don't really see the need for this to attempt to be anything above it's status. Legalise is bullshit because it's too hard to read and understand for the average Joe and any agreements made, and the terms of service itself, should be in a language and terminology that is so simple everyone can understand.

nightpool (approved):

nightpool:
changes LGTM

PR #5820, opened 2017-11-26
Give each status a copyright

mike-burns:
This PR is incomplete but what do we think of the idea?

- Works of art have a copyright in most countries; this documents the currently-implied copyright.
- This could set the way for changing the copyright on a per-status basis, if we were to expose that in the UI.

Also, I hate the string duplication but wanted to get a discussion going before I figured out how to best fix that within this codebase. Same with the lack of tests.

---

Every work created by a person has a copyright of some type, in most
countries. Document the previously-implied copyright using the Creative
Commons license Attribution 4.0 International (CC BY 4.0).

https://creativecommons.org/licenses/by/4.0/

For ActivityPub/ActivityStreams, bring in the ccREL ontology, connecting
the work with a `license` property.

For OStatus, augment the Atom entry with a link to the license, as
documented by CC: https://wiki.creativecommons.org/wiki/Atom

Add a `license` value to the REST API for anything involving a status.

This does not expose anything user-facing; it simply seeks to document
the current state of things in a machine-readable way.


akihikodaki (changes requested):
The idea is great (one of my friends gives an explicit license for his posts and he would be grateful with the change), but I do not agree with the implementation.

First, your change will be applied to all the old posts when the server has the later versions. That means updating their instances will require to update the acknowledgement of _all_ users. That is nearly impossible for some instances.

Secondly, your change does not allow to change the license. Why doesn't it do so while it is allowed to change the instance policy?

At the end, I strongly recommend to have notices about the license somewhere on the UI and to provide an API to retrieve the license when posting. The lack of the feature gives doubts on the validity of the license claim, and even `cc:license` context may be considered as non-reliable information, which is my major concern.

akihikodaki:
For your information, the terms of Discourse, on which the terms of Mastodon is based, has section named _User Content License_, and says the contributions will be licensed under CC-BY-NC-SA 3.0. It is a nice way to tell users about their copyrights.

mike-burns:
Thank you for the feedback, @akihikodaki !

Here's how I'll move this forward:

- Add a `license` field to the `statuses` table.
- Provide a UI for selecting a license -- in the media/privacy/CW cluster seems right.
- Only send license information if one is explicitly selected for that status.
- Have a default license applied to the status, which can be saved for the user as a setting.
- Show license information on a status in the Web UI.

Seem right?

Some decisions that I don't feel qualified to make:

- This is my first time working with OStatus, ActivityPub, etc. Did I get all the parts of the codebase where we send the status around?
- What is the default license that new instances should ship with?  Is CC attribution-only correct? Public domain? None? Again this gets into the weird territory where all works do have a copyright so "none" (in most countries) is not actually a legally-correct option.
- Is it OK to have a drop-down pre-populated with all the CC licenses? Should it be free-form text?
- People may want to retroactively apply licenses to prior statuses -- should we have a UI for that?
- I am mixing the words "license" and "copyright" in here -- which is this, and is it OK to conflate them a little bit?

akihikodaki:
@mike-burns Thank you for your quick reply.

> Here's how I'll move this forward:

They look nice, but also tough to implement. I would highly appreciate if you made those changes. Here is some feedback for them.

> * Provide a UI for selecting a license -- in the media/privacy/CW cluster seems right.

I don't prefer to add a more control to the UI. The column is messy even now. It is still fine to expose an API for third-party applications to provide the same feature. Please note that this is just my opinion and there is no clear convention about new change for the column, so others may have different opinions.

> * Only send license information if one is explicitly selected for that status.

I want to make sure what a license "explicitly selected" would be. In my opinion:
* The statuses (i) made by users who have registered after their instances applied this change should have an explicitly selected license because such instances will have a default license and have told them about that.

* The statuses (i) made by users who have registered before their instances applied this change and (ii) made before they set their default license (iii) whose license have not selected via the compose UI will _not_ have an explicitly selected license.

Technically, that can be implemented by:
* adding (possibly nullable) `default_license` field to the instance configuration,
* adding nullable `default_license` field to `users` table, and
* adding nullable `license` field to `statuses` table.

The `license` field for statuses made before this change should be `NULL`.
Otherwise, the `license` field should be:
* the value chosen in the compose form if one is chosen
* or `default_license` field of the user if it is not `NULL`
* or `default_license` field of the instance

If the `license` field of a status is `NULL`, the status does not have its explicitly selected license. Otherwise, the value indicates its explicitly selected license.

> This is my first time working with OStatus, ActivityPub, etc. Did I get all the parts of the codebase where we send the status around?

It looks good to me. However, please note that you may add a change to interpret the license field of remote statuses and fill the corresponding column of the database if you are adding a change for the database.

> What is the default license that new instances should ship with? Is CC attribution-only correct? Public domain? None? Again this gets into the weird territory where all works do have a copyright so "none" (in most countries) is not actually a legally-correct option.

The following is just my preference and other reviewers may have different opinion, but:
* I prefer the latest CC-BY-SA.
* Public domain is not good. It may introduce a kind of stereotype that everybody can do anything for statuses on Mastodon. At least we need to show that we are taking account of copyrights.
* "None", is actually safe because it is the current "default" license.

Again, thank you for your considerations.

akihikodaki:
I have not answered all questions...

> Is it OK to have a drop-down pre-populated with all the CC licenses? Should it be free-form text?

Free-form text does not seem possible because of the specification of CC. I suggest to implement with pre-populated CC licenses as the initial version.

> People may want to retroactively apply licenses to prior statuses -- should we have a UI for that?

I do not think so because:
* updating federated statuses is tough and
* usually you cannot withdraw the license

* I am mixing the words "license" and "copyright" in here -- which is this, and is it OK to conflate them a little bit?

I understand those words are different so you do not have to. You may choose appropriate one. Here is my long description: (disclaimer: I am not a lawyer and I am Japanese.)
In my understanding, copyright is the right of the owner and license is statement describing the condition the owner does _not_ begin disputes. So you may say you give a license for each status and you will not sue others using your copyright while they are conforming to the license. Note that copyright is not necessarily "given"; but license is.

mike-burns:
I found a few hours so here's an update:

- Adds a `default_license` setting on the user, with a dropdown in the settings presenting all the CC licenses.
- The `default_license`is applied to new statuses, persisted on the `statuses` table.
- The status' license URL is sent via OStatus, ActivityPub, and the REST API.
- The REST API also sends the license name.
- Show the license in the Web UI (JS and HTML, with some CSS).
- If the default license is unset, ignore it.
- The the status' license is unset, ignore it.
- Some light testing.

Still on my to-do list:
- Instance defaults. Or can this be merged without that feature?

I need help:

- How should it look in the Web and JS UI? If it should be any fancier than this, I'd love assistance there.
- Are there tests for the serializers that I had missed?
- Are there JS tests? Integration/system/feature tests?
- Translations.

Is this a fine start?

mike-burns:
@trwnh I see you gave this a thumbs-down. Anything you want to discuss?

app/views/settings/preferences/show.html.haml
```diff
@@ -18,6 +18,8 @@
 
     = f.input :setting_default_sensitive, as: :boolean, wrapper: :with_label
 
+    = f.input :setting_default_license, collection: Status::LICENSE_URLS.keys, wrapper: :with_label
```
mike-burns: Perhaps `Status::LICENSE_URLS.keys` is not user-friendly. Maybe this whole input is confusing. Very open to opinions here.

trwnh:
@mike-burns Not particularly. I just disagree with implementing anything that reifies copyrights in general, especially in open software.

akihikodaki:
> Not particularly. I just disagree with implementing anything that reifies copyrights in general, especially in open software.

The copyright is based on laws and cannot be ignored by any means. If you disagree the idea of copyrights, you need to clarify the stance (that is why FSF made GPL.)

app/models/status.rb
```diff
@@ -31,6 +31,15 @@ class Status < ApplicationRecord
   include Cacheable
   include StatusThreadingConcern
 
+  LICENSE_URLS = {
+    'Attribution (CC BY)' => 'https://creativecommons.org/licenses/by/4.0/',
+    'Attribution-ShareAlike (CC BY-SA)' => 'https://creativecommons.org/licenses/by-sa/4.0/',
+    'Attribution-NoDerivs (CC BY-ND)' => 'https://creativecommons.org/licenses/by-nd/4.0/',
```
akihikodaki: The web page says "Attribution-NoDerivatives" (while it says "Attribution-NonCommercial-NoDerivs" for BY-NC-ND. How confusing!)

app/serializers/activitypub/note_serializer.rb
```diff
@@ -86,6 +86,10 @@ def local?
     object.account.local?
   end
 
+  def license
```
akihikodaki: Use `key` option of `attribute` method. See http://www.rubydoc.info/gems/active_model_serializers/0.10.6/ActiveModel/Serializer#attribute-class_method for details.

app/views/settings/preferences/show.html.haml
akihikodaki: It also prevents from localization. Creative Commons localizes license names on their Web sites. It is better to conform the prior instances and localize license names as they do.

akihikodaki (changes requested):
Because of localization of license names, I'd like to have changes to use identifiers which are different from the names for the database. URIs are totally fine in my opinion.

trwnh:
@akihikodaki:
> The copyright is based on laws and cannot be ignored by any means. If you disagree the idea of copyrights, you need to clarify the stance (that is why FSF made GPL.)

To clarify, I do not think that copyright status should be supported in general. It may be the law in several countries, but the existence of laws does not require accommodating them. I would oppose anything that made it easier to enforce that law. In this case, that would mean *not* providing an exposed copyright status in the UI. This is my belief, and I am expressing it here to provide some dissent to the idea, though this pull request may ultimately end up being accepted.

mike-burns:
Ah great points re localization. I'll change what we store in the DB and how that makes it to the user. I'll also make the other improvements you suggest.

Is this idea OK to merge without instance defaults? As-is it will preserve existing functionality. We can add instance defaults in a separate diff.

akihikodaki:
@trwnh
> I would oppose anything that made it easier to enforce that law.

That makes sense.

@mike-burns

> Ah great points re localization. I'll change what we store in the DB and how that makes it to the user. I'll also make the other improvements you suggest.

Nice!

> Is this idea OK to merge without instance defaults? As-is it will preserve existing functionality.

Yes.

> We can add instance defaults in a separate diff.

I agree.

mike-burns:
@akihikodaki updated:

- Store the URL in the DB.
- Use i18n to show the label.
- Populate the translations from creativecommons.org.

[I typically work on this branch in 1 hour bursts but today I had the time to spend the requisite number of hours remembering how simple_form, collections, and i18n interact.]

akihikodaki (changes requested):
It is almost done! The only thing left is to implement receivers for federation. See `app/lib/activitypub/activity/create.rb` for ActivityPub, `app/lib/ostatus/activity/creation.rb` for OStatus.

akihikodaki:
@Gargron This pull request has grown much. I would like to know if there is a chance to get this merged.

mike-burns:
Ah right good catch; added to the receiver side of things.

Much of the diff growth is in i18n.

mike-burns:
If you'd like me to re-write some git history just say the word -- though let's wait until we even know that we got it right and that we want it merged.

app/lib/ostatus/atom_serializer.rb
```diff
@@ -348,6 +348,10 @@ def add_namespaces(parent)
   def serialize_status_attributes(entry, status)
     append_element(entry, 'link', nil, rel: :alternate, type: 'application/activity+json', href: ActivityPub::TagManager.instance.uri_for(status)) if status.account.local?
 
+    if status.license_url
+      append_element(entry, 'link', rel: :license, type: 'application/rdf+xml', href: status.license_url)
```
akihikodaki: It should not say `type: 'application/rdf+xml'`.

I noticed that `LICENSE_URLS` are not for `application/rdf+xml`. For example, the RDF for CC BY is not `https://creativecommons.org/licenses/by/4.0/`, but `https://creativecommons.org/licenses/by/4.0/rdf`.
So either `LICENSE_URLS` or the `type` attribute must be changed. According to CC REL specification, removing the `type` attribute is better.

akihikodaki (changes requested):
The OStatus interpreter tests if the `type` attributes of links are `application/rdf+xml`, but it seems incorrect.

akihikodaki:
The Git history is fine. We do not have any rule to define how to organize commits as far as I know.

app/lib/activitypub/adapter.rb
```diff
@@ -17,6 +17,8 @@ class ActivityPub::Adapter < ActiveModelSerializers::Adapter::Base
         'conversation'              => 'ostatus:conversation',
         'toot'                      => 'http://joinmastodon.org/ns#',
         'Emoji'                     => 'toot:Emoji',
+        'cc'                        => 'https://creativecommons.org/ns#',
+        'license'                   => 'cc:license',
```
cwebber: The license term doesn't exist in cc: (it's in fact the one thing not in cc: in ccRel), it's in the xhtml namespace.  Since you probably only need one term from xhtml I'd simply do:

```
 'license'       => 'http://www.w3.org/1999/xhtml/vocab#license'
```

app/serializers/rest/status_serializer.rb
```diff
@@ -3,7 +3,8 @@
 class REST::StatusSerializer < ActiveModel::Serializer
   attributes :id, :created_at, :in_reply_to_id, :in_reply_to_account_id,
              :sensitive, :spoiler_text, :visibility, :language,
-             :uri, :content, :url, :reblogs_count, :favourites_count
+             :uri, :content, :url, :reblogs_count, :favourites_count,
+             :license, :license_url
```
Gargron: It feels like it would be a bit cleaner, API wise, to have a `license: { name: ..., url: ... }` object

Gargron (changes requested):
With a couple changes as requested by me, aki and cwebber I think this is a viable addition. Sorry it took like, 8 months to get to this, time flies and it feels like new year's was just yesterday.

nightpool:
Two questions on UX:

 - After a user selects a default license, how do they unset it?
 - Can the user put in a custom default license value? Do we think they should be able to?

db/migrate/20171221044254_add_license_to_statuses.rb
```diff
@@ -0,0 +1,5 @@
+class AddLicenseToStatuses < ActiveRecord::Migration[5.1]
+  def change
+    add_column :statuses, :license_url, :text, null: true
```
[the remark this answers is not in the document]
mike-burns: For Postgres, this is the canonical article: https://www.depesz.com/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/

I'm willing to do something else if we want to support other DBMSes more natively?

mike-burns:
Thanks for the great feedback everyone. I'll rebase and address your comments in a few days.

db/migrate/20171221044254_add_license_to_statuses.rb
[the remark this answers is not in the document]
Gargron: We don't plan on supporting any other DB type since we already use some postgres-specific syntax and column types. However, no other URL has a "text" type so far and I don't see a reason to do this one differently.

app/serializers/rest/status_serializer.rb
```diff
@@ -86,6 +87,13 @@ def pinnable?
       %w(public unlisted).include?(object.visibility)
   end
 
+  def license
+    I18n.t(
+      Status::LICENSE_URLS[object.license_url],
+      scope: 'simple_form.options.user.setting_default_license'
+    )
```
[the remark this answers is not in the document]
Gargron: Hmm. The only concern is, how do you deal with unrecognized licenses. As far as *recognized* licenses go, this seems fine. (Unless you want to push this concern to the app's i18n...)

db/migrate/20171221044254_add_license_to_statuses.rb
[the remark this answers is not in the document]
mike-burns: Cool -- I value consistency. I'll make the change.

app/serializers/rest/status_serializer.rb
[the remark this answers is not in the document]
nightpool: i would just imagine that we store/federate license name along with license url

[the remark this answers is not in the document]
nightpool: i guess that's more difficult though, and fucks up i18n

mike-burns:
@nightpool 

> After a user selects a default license, how do they unset it?

Currently not possible. Is that a feature we want?

One thing I keep coming back to in my mind is: every status has a license (in Europe and NA, at least); this is about exposing it and giving the user the option to change it.

But if you think it makes sense to, I can add a null value to the dropdown in `preferences/show.html.haml`.

> Can the user put in a custom default license value? Do we think they should be able to? does the additional nuance justify the additional UX complexity?

You raise some good points, especially technical ones inline. Regardless of how we limit the user on Mastodon there's also the rest of the fediverse to consider.

How should we federate the human-visible value of the license? Should we additionally send a default name (e.g. "BY-NC-SA") and leave it up to the client to render that as they please (we would continue to run it through i18n (e.g. "Attribution-NonCommercial-ShareAlike (CC BY-NC-SA)"))? Should we continue to send only the URL and show something like "©" with a link to their URL if we don't recognize it? And what of licenses that don't have a URL?

Simplest option: we only support CC licenses for now. Only send URL, have a fallback for URLs we don't recognize.

Happy to change this as you see fit. Let me know how you think we can make this work.

PR #5821, opened 2017-11-26
Use account.display_name for og:title single toot pages

Aldarone:
Because it seems nicer that way.


akihikodaki (approved):

Gargron:
Currently:

ss 2017-11-29 at 10 57 29

After this PR:

"Eugen
Eugen on mastodon.social"

Is that really nicer? 🤔

Aldarone:
I have this : 
capture d ecran_2017-11-29_23-00-45

Perhaps yours is caused by #5828 ?

ykzts:
It is necessary when the user does not set a name.

PR #5164, opened 2017-10-01
Implement hotkeys for web UI

Gargron:
- Fix #2102
- Fix #471
- Fix #376

Demo: <https://mastodon.social/@Gargron/20188468>

Hotkeys:

- `r` to reply
- `m` to mention author
- `f` to favourite
- `b` to boost
- `enter` to open status
- `p` to open author's profile
- `up` or `k` to move up in the list
- `down` or `j` to move down in the list
- `1`-`9` to focus a status in one of the columns
- `n` to focus the compose textarea
- `alt`+`n` to start a brand new toot
- `backspace` to navigate back
- `s` to focus search
- `esc` to un-focus compose textarea/search
- various "goto" combinations

PR #5823, opened 2017-11-26
Add Keyboard Shortcuts Legend

### LINKED ISSUE #5573: Add guide to hotkeys
(I think this has already been brought up but I can't find an issue so I'm opening one.) Currently the only way to learn what mastodon's new hotkeys are is to go to the release notes on github. There should be some in-UI guide to the hotkeys, possibly a modal that pops up when you press `?`.

* * * *

joshuap:
This adds a "Keyboard Shortcuts" legend (displayed in the rightmost column) which is toggled via a new "?" hotkey. When subsequently pressed from the Keyboard Shortcuts legend, "?" will navigate back to the previous location:

keyboard-shortcuts

I copied the hotkeys/legend directly from #5164; let me know if I'm missing any. I'm still getting familiar with the codebase.

Fixes #5573

### DIFF
--- a/app/javascript/mastodon/features/keyboard_shortcuts/index.js
+++ b/app/javascript/mastodon/features/keyboard_shortcuts/index.js
@@ -6,6 +6,8 @@ import ImmutablePureComponent from 'react-immutable-pure-component';
 
 const messages = defineMessages({
   heading: { id: 'keyboard_shortcuts.heading', defaultMessage: 'Keyboard Shortcuts' },
+  hotkey: { id: 'keyboard_shortcuts.hotkey', defaultMessage: 'Hotkey' },
+  description: { id: 'keyboard_shortcuts.description', defaultMessage: 'Description' },
   reply: { id: 'keyboard_shortcuts.reply', defaultMessage: 'to reply' },
   mention: { id: 'keyboard_shortcuts.mention', defaultMessage: 'to mention author' },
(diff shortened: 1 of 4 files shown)


app/javascript/mastodon/features/keyboard_shortcuts/index.js
```diff
@@ -0,0 +1,63 @@
...
+    multiColumn: PropTypes.bool,
+  };
+
+  render () {
+    const { intl } = this.props;
+
+    return (
+      <Column icon='question' heading={intl.formatMessage(messages.heading)} hideHeadingOnMobile>
+        <div className='keyboard-shortcuts scrollable optionally-scrollable'>
+          <table>
```
sorin-davidoi: I would add the header rows for accessibility reasons - we can give them `height: 0` if we decide to hide them for aesthetic reasons.

sorin-davidoi (changes requested):
Just a small nitpick - thanks for taking the time to fix this! 😄

app/javascript/mastodon/features/keyboard_shortcuts/index.js
joshuap: @sorin-davidoi take a look at 2ec97c12f4b140a33a25f5f8cf4f9a8ed12eef07 ? I used a different CSS technique which should work for accessibility (had trouble getting `height: 0;` to work with the table, but if you want to suggest some CSS I'm happy to revise again).

sorin-davidoi (approved):

PR #5838, opened 2017-11-28
Show "expired" in invite index

ykzts:
ref #5814

Show "Expired" in the invite index. This makes it easier to understand valid links.

### Screenshot

screenshot

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -432,7 +432,8 @@ en:
     upload: Upload
   in_memoriam_html: In Memoriam.
   invites:
-    delete: Delete
+    delete: Force expire
+    expired: Expired
     expires_in:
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -432,7 +432,8 @@ en:
     upload: Upload
   in_memoriam_html: In Memoriam.
   invites:
-    delete: Delete
+    delete: Force expire
```
Gargron: Maybe "deactivate"? "Force expire" is very technical

Gargron (changes requested):

config/locales/en.yml
ykzts: 👍 962e864

Gargron (approved):

PR #5848, opened 2017-11-29
Add semi-support for Video/Image objects in ActivityPub

Gargron:
PeerTube now talks ActivityPub: Chocobozzz/PeerTube#104

Video and Image objects will create corresponding text status records with manually crafted text contents (title + URL). Furthermore, the code now supports extracting the language from `nameMap` attribute as fallback of `contentMap` (name = title, since there is no title attribute)

The implications of this is that since PeerTube implements the right OpenGraph tags (e.g. `twitter:player`) such toots will actually embed the PeerTube video player as their preview card; it also means that replies to the "toot" will still be sent to the right object upstream and show up as comments on PeerTube.

Mind that at the time of writing PeerTube actually needs to fix their language serialization because they use a non-existing `language` attribute when the established way is `nameMap`/`contentMap`, and likewise the text/html URL is missing from their `url` array.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class ActivityPub::Activity::Create < ActivityPub::Activity
+  SUPPORTED_TYPES = %w(Article Note).freeze
+  CONVERTED_TYPES = %w(Image Video).freeze
+
   def perform
     return if delete_arrived_first?(object_uri) || unsupported_object_type?
 
@@ -165,7 +168,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def text_from_content
-    return Formatter.instance.linkify([text_from_name, object_url].join(' ')) if converted_object_type?
+    return Formatter.instance.linkify([text_from_name, object_url || @object['id']].join(' ')) if converted_object_type?
 
     if @object['content'].present?
       @object['content']
@@ -214,11 +217,11 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   end
 
   def supported_object_type?
-    %w(Article Note).include?(@object['type'])
+    SUPPORTED_TYPES.include?(@object['type'])
   end
 
   def converted_object_type?
-    %w(Video Image).include?(@object['type'])
+    CONVERTED_TYPES.include?(@object['type'])
   end
 
   def skip_download?
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -194,22 +194,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
 
   def object_url
     return if @object['url'].blank?
-
-    # The url attribute can be a string, an array of strings, or an array of objects.
-    # The objects could include a mimeType. Not-included mimeType means it's text/html.
-    # We need to prefer text/html. So if the attribute has the fullest information,
-    # we've got to select an item that either has no mimeType or text/html mimeType
-    value = if @object['url'].is_a?(Array) && !@object['url'].first.is_a?(String)
-              @object['url'].find { |url| url['mimeType'].blank? || url['mimeType'] == 'text/html' }
-            elsif @object['url'].is_a?(Array)
-              @object['url'].first
-            else
-              @object['url']
-            end
-
-    return value if value.nil? || value.is_a?(String)
-
-    value['href']
+    url_to_href(@object['url'], 'text/html')
   end
 
   def content_language_map?
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -41,7 +41,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
       url: object_url || @object['id'],
       account: @account,
       text: text_from_content || '',
-      language: language_from_content,
+      language: detected_language,
       spoiler_text: @object['summary'] || '',
       created_at: @options[:override_timestamps] ? nil : @object['published'],
       reply: @object['inReplyTo'].present?,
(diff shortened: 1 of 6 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -165,40 +165,77 @@ def in_reply_to_uri
...
-  def language_from_content
-    return LanguageDetector.instance.detect(text_from_content, @account) unless language_map?
-    @object['contentMap'].keys.first
+  def text_from_name
+    if @object['name'].present?
+      @object['name']
+    elsif name_language_map?
+      @object['nameMap'].values.first
+    end
+  end
```
nightpool: can be `@object['name] || @object['nameMap']&.values&.first` but it feels like we need a helper here instead, since every text value (name, content, preferredUsername, tags) can have this form and we have helpers for similar JSON-LD conventions (`equals_or_includes`, `value_or_id`)

app/lib/activitypub/activity/create.rb
```diff
@@ -165,40 +165,77 @@ def in_reply_to_uri
...
+
+  def detected_language
+    if content_language_map?
+      @object['contentMap'].keys.first
+    elsif name_language_map?
+      @object['nameMap'].keys.first
+    elsif supported_object_type?
```
nightpool: why not do this for converted object types?

app/lib/activitypub/activity/create.rb
```diff
@@ -165,40 +165,77 @@ def in_reply_to_uri
...
 
   def object_url
     return if @object['url'].blank?
 
-    value = first_of_value(@object['url'])
+    # The url attribute can be a string, an array of strings, or an array of objects.
+    # The objects could include a mimeType. Not-included mimeType means it's text/html.
+    # We need to prefer text/html. So if the attribute has the fullest information,
+    # we've got to select an item that either has no mimeType or text/html mimeType
+    value = if @object['url'].is_a?(Array) && !@object['url'].first.is_a?(String)
```
nightpool: again, probably needs a function in JsonLdHelpers

app/lib/formatter.rb
```diff
@@ -69,6 +64,14 @@ def format_spoiler(status)
     html.html_safe # rubocop:disable Rails/OutputSafety
   end
 
+  def linkify(text)
+    html = encode_and_link_urls(text)
```
nightpool: does encode_and_link_urls sanitize correctly here? Or am I missing where the HTML-safety happens?

app/lib/activitypub/activity/create.rb
Gargron: If a Video does not have a language attribute, what's the point of trying to detect language from its description? What would language even refer to if we're talking about a Video or an Image? for text posts it's clear. For media, it is not, because the thing itself is not the text, the text is merely a description, which may be missing, which may be in a different language.

app/lib/activitypub/activity/create.rb
Gargron: This only comes up twice so I disagree about the need to abstract it.

app/lib/activitypub/activity/create.rb
Gargron: This is a more specific version of first_or_value, I can't easily generalize it.

app/services/activitypub/fetch_remote_status_service.rb
```diff
@@ -42,7 +42,7 @@ def supported_context?
   end
 
   def expected_type?
-    %w(Note Article).include? @json['type']
+    %w(Note Article Video Image).include? @json['type']
```
nightpool: this should be centralized into a constant on Create.rb

app/lib/formatter.rb
Gargron: Yes, it does that. This function has been extracted verbatim from simplified_format.

app/lib/activitypub/activity/create.rb
```diff
@@ -165,40 +165,77 @@ def in_reply_to_uri
...
   def unsupported_media_type?(mime_type)
     mime_type.present? && !(MediaAttachment::IMAGE_MIME_TYPES + MediaAttachment::VIDEO_MIME_TYPES).include?(mime_type)
   end
 
+  def supported_object_type?
+    %w(Article Note).include?(@object['type'])
+  end
+
+  def converted_object_type?
+    %w(Video Image).include?(@object['type'])
```
nightpool: this should use a constant string array so fetch_remote_status_service can be in-sync

app/lib/activitypub/activity/create.rb
```diff
@@ -165,40 +165,77 @@ def in_reply_to_uri
   end
 
   def text_from_content
+    return Formatter.instance.linkify([text_from_name, object_url].join(' ')) if converted_object_type?
```
nightpool: should maybe fall back to id if object_url isn't included?

app/lib/activitypub/activity/create.rb
nightpool: It's not because of DRY reasons that I'm saying it should be abstracted, it's because it's a fundamental part of processing JSON-LD content that can be implemented/tested/used independently of any actual activity processing.

app/lib/activitypub/activity/create.rb
nightpool: There may be other converted object types, like if we wanted to implement base Activities, but I see your point.

Gargron: >There may be other converted object types

AP doesn't have a general language attribute. It can only provide languages for name/content. That means it can only be relevant for text posts where object == its text.

app/lib/activitypub/activity/create.rb
nightpool: Isn't this the way we should process any link value though?

app/lib/formatter.rb
nightpool: ✅

nightpool (commented):
SHOULD:

 - remove hard-coded types in fetch_remote_status_service
 - fall back to id if object['url'] isn't included

MAY: 

 - extract the complicated link-value logic into another function for readability

nightpool (approved):
LGTM!

Chocobozzz:
Hi,

The custom `language` in PeerTube is in fact the video's language. The language of the name/content could be different and is unknown in PeerTube.

We added the `text/html` url in the object -> https://github.com/Chocobozzz/PeerTube/commit/165cdc75bf1942ed687f78094c2bd366839a7c99#diff-852ba6d236e543da3d7c6d5128f06759R619

The last incompatibility is the signature algorithm, we need to implement `RsaSignature2017`. I asked the devs of the module we use [if there are any news](https://github.com/digitalbazaar/jsonld-signatures/pull/28) and I'm waiting for their response. If nothing move in the next weeks I'll implement it myself :)

Gargron:
@Chocobozzz ld-sigs are only relevant for forwarding which in case of peertube doesn't seem like a big deal to me. like, if i replied to a video of yours, you'd want to forward my reply to all people you know who have that video too; but they won't notice if you don't do that. the other use case is when you delete a video, you'd want people who have reblogged it to also forward the delete to their followers so it disappears from all the places it propagated.

drequivalent:
Can't follow chocobozzz@peertube.cpy from my 2.1.0 instance
screenshot from 2017-12-18 03-46-49
screenshot from 2017-12-18 03-46-58

Chocobozzz:
It's `chocobozzz@peertube.cpy.re:443` but it won't really work, it's still a WIP on PeerTube. I'm currently working on this.

drequivalent:
Understood.

PR #5854, opened 2017-11-29
Fix too many forwards

### LINKED ISSUE #5761: ActivityPub: deleting a post spams a whole lot of Undo's to followers
If someone deletes a post, I get a whole bunch of `Undo` messages from each user. it seems these are based on an unknown statistic:

- `mastodon.social`: 1 Undo
- `octodon.social`: 3 Undos
- `vulpine.club`: 2 Undos
- `icosahedron.website`: 17 Undos (!!!!!!)

Now I *should* dedupe these values. But! 17 Undo's is still like 16 too much. Seriously. I didn't even get the original Delete!

The post ID is `tag:mastodon.social,2017-06-09:objectId=8499030:objectType=Status`, btw
* * * *

Gargron:
Fix #5761

### DIFF
--- a/app/lib/activitypub/activity/delete.rb
+++ b/app/lib/activitypub/activity/delete.rb
@@ -30,8 +30,11 @@ class ActivityPub::Activity::Delete < ActivityPub::Activity
   def forward_for_reblogs(status)
     return if @json['signature'].blank?
 
-    ActivityPub::RawDistributionWorker.push_bulk(status.reblogs.includes(:account).references(:account).merge(Account.local).pluck(:account_id)) do |account_id|
-      [payload, account_id]
+    rebloggers_ids = status.reblogs.includes(:account).references(:account).merge(Account.local).pluck(:account_id)
+    inboxes        = Account.where(id: Follow.where(target_account_id: rebloggers_ids)).inboxes - [@account.preferred_inbox_url]
+
+    ActivityPub::DeliveryWorker.push_bulk(inboxes) do |inbox_url|
+      [payload, account_id, inbox_url]
     end
   end
 
--- a/app/lib/activitypub/activity/delete.rb
+++ b/app/lib/activitypub/activity/delete.rb
@@ -31,10 +31,10 @@ class ActivityPub::Activity::Delete < ActivityPub::Activity
     return if @json['signature'].blank?
 
     rebloggers_ids = status.reblogs.includes(:account).references(:account).merge(Account.local).pluck(:account_id)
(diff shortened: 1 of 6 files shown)


app/services/remove_status_service.rb
```diff
@@ -22,7 +23,12 @@ def call(status)
 
     @status.destroy!
 
-    return unless @account.local?
+    # There is no reason to send out Undo activities when the
+    # cause is that the original object has been removed, since
+    # original object being removed implicitly removes reblogs
+    # of it. The Delete activity of the original is forwarded
+    # separately.
+    return if !@account.local? || @options[:original_removed]
```
nightpool: still don't know if I agree with this—some implementations might treat the deletion of the original vs deletions of reblogs separately

(for example, this is how it works on tumblr so sites looking to implement tumblr-like behavior where reblogs can add addition comments might not like it)

spec/lib/activitypub/activity/delete_spec.rb
```diff
@@ -46,7 +46,7 @@
 
       it 'sends delete activity to followers of rebloggers' do
         # one for Delete original post, and one for Undo reblog (normal delivery)
```
nightpool: comment needs to be updated

nightpool:
still have reservations about skipping Undo -> Reblogs, but the deduplication stuff LGTM

app/services/remove_status_service.rb
Gargron: I mean if you think about it in terms of user actions, nobody actually chooses to un-reblog a post that's been deleted. What we've been doing is enforcing a certain implicit behaviour explicitly. If your software doesn't have referential integrity and you want to display reblogs-of-dead-objects, maybe that's fine?

Given that both delete-forwarding and undo-announces go out to the same recipients, removing this behaviour instantly halves the required traffic.

app/lib/activitypub/activity/delete.rb
```diff
@@ -31,10 +31,10 @@ def forward_for_reblogs(status)
     return if @json['signature'].blank?
 
     rebloggers_ids = status.reblogs.includes(:account).references(:account).merge(Account.local).pluck(:account_id)
-    inboxes        = Account.where(id: Follow.where(target_account_id: rebloggers_ids)).inboxes - [@account.preferred_inbox_url]
+    inboxes        = Account.where(id: ::Follow.where(target_account_id: rebloggers_ids).select(:account_id)).inboxes - [@account.preferred_inbox_url]
```
nightpool: oh, yikes, that's subtle and annoying.

nightpool (approved):

PR #5858, opened 2017-11-30
i18n: Update ja translations for KeyboardShortcuts.

mayaeh:
Update Japanese translations for #5835 .


Gargron (approved):

app/javascript/mastodon/locales/ja.json
```diff
@@ -102,23 +102,23 @@
...
+  "keyboard_shortcuts.column": "左からn番目のカラム内最新トゥートに移動",
+  "keyboard_shortcuts.compose": "トゥート入力欄に移動",
+  "keyboard_shortcuts.description": "説明",
+  "keyboard_shortcuts.down": "カラム内一つ下に移動",
+  "keyboard_shortcuts.enter": "トゥートの詳細を表示",
+  "keyboard_shortcuts.favourite": "お気に入り",
+  "keyboard_shortcuts.heading": "キーボードショートカット一覧",
+  "keyboard_shortcuts.hotkey": "ホットキー",
+  "keyboard_shortcuts.legend": "この一覧を表示",
+  "keyboard_shortcuts.mention": "メンション",
```
SerCom-KC: Just a reminder, I think in Japanese [`mention` is also translated as `返信`](https://github.com/mayaeh/mastodon/blob/d1a58bf9209a6a59617679c79f849bf47a122517/app/javascript/mastodon/locales/ja.json#L207) at the moment. I'm not a native speaker though.

mayaeh: Thank you.

I made separate translations, because mention and reply have slightly different functions.

I translated "Reply" as "返信".
"返信" means reply to something, and there is a target to return.
Reply function is a reply to TOOT, and it isn't displayed on the local TL etc.
Also reply function is displayed as a series of conversations.

And I translated "Mention" as "メンション".
Mention function is a reply to a person, and it is also displayed in the local TL etc.
I think that Mention function of SNS service is often used mainly when contact someone.

Also "Mention" is often translated as "言及" in Japanese.
"言及" is a hard expression and is not used much in everyday conversation.

I think that other SNS service also translates "Mention" as "メンション" and uses it.

So I translated "Mention" into "メンション".

SerCom-KC: I know, I'm just talking about unification here, in order to avoid confusions :)

PR #5859, opened 2017-12-01
Using double splat operator

ykzts:
`def method(options = {})` -> `def method(**options)`


Gargron:
Can you refresh my memory on what the benefit of this is?

ykzts:
By using the double splat operator, you can prevent mixing of `'stringkey' => 'value'` which is not suitable as the value of options.

```ruby
def test1(**options); end
  
test1(test: 1) # => nil

test1({ 'test' => 1}) # => wrong number of arguments (given 1, expected 0)
```

Gargron (approved):

PR #5863, opened 2017-12-01
Remove redudant import statement in components.scss

cormojs:


cormojs:
Oh I choosed wrong branch.

PR #5864, opened 2017-12-01
Remove redundant import statement in SCSS

cormojs:


lynlynlynx (approved):
I have tested removing this for a while, and it has been confirmed working properly.

ykzts (commented):
why chmod?

cormojs:
@ykzts I notice that my shared directory(smbd) has wrong permission setting. Now I fixed it.

PR #5870, opened 2017-12-02
Allow muting conversations from anywhere

ClearlyClaire:
Allow muting conversations from anywhere, not only from the notification column.

Indeed, while it's most useful in the notifications column (since muting a conversation only prevents notifications for this conversation from being created), it makes sense for this menu entry to appear everywhere.


ThibG:
Already fixed by #6099

PR #5879, opened 2017-12-03
Keep WebPush settings

abcang:
When logging out, the WebPush setting disappears. So, I propose a feature to save the WebPush setting to localStorage and restore it at login.


ykzts:
cc @sorin-davidoi

abcang:
I want to solve this problem because it seems like a bug that WebPush settings disappear when logging out. So, I think that another method that does not use localStorage is acceptable (for example, setting is saved for each account, not for each session). I will close this PR if there are other good solutions.

Also, in this PR, since the setting is saved with the key containing the account id in localStorage, the settings of others are not applied at login.

ykzts (approved):

PR #5881, opened 2017-12-04
fix steam.js param bug

LittleTheFu:


app/javascript/mastodon/stream.js
```diff
@@ -1,6 +1,6 @@
 import WebSocketClient from 'websocket.js';
 
-export function connectStream(path, pollingRefresh = null, callbacks = () => ({ onConnect() {}, onDisconnect() {}, onReceive() {} })) {
+export function connectStream(path, pollingRefresh = null, callbacks = (dispatch, getState) => ({ onConnect() {}, onDisconnect() {}, onReceive() {} })) {
```
ykzts: Neither `dispatch` nor ` getState` is used.

ykzts:
I do not think this is a bug. I close this Pull Request, but if you think it is a bug please reopen and explain what kind of bug it is.

PR #5885, opened 2017-12-05
Omit optional dependencies on JRuby

Gargron:
Probably not entirely working, but removes some things that are not needed on JRuby that don't have ports.

Reference: https://www.youtube.com/watch?v=Lja3HDcHNJA&feature=youtu.be&t=32m

cc @headius


akihikodaki:
For your information, cld3-ruby does not compile either because of a bug in JRuby. cld3-ruby should be removed to workaround or a fix, https://github.com/jruby/jruby/pull/4118 should be applied.

PR #5887, opened 2017-12-05
Break out nested relationship API keys

### LINKED ISSUE #5856: API response changes on #5762 broke many apps
quote from #5762:

> As with the notification muting, this will now return an object ({reblogs: [true|false]}) or false for each follow relationship when requesting relationship information for an account. This should cause few issues due to an object being truthy in many languages, but some modifications may need to be made in pickier languages.

Before:

```json
[{
  "id": "1",
  "following": true,
  "followed_by": true,
  "blocking": false,
  "muting": false,
  "requested": false,
  "domain_blocking": false
}]
```

After:

```json
[{
  "id": "1",
  "following": {
    "reblogs": true
  },
  "followed_by": true,
  "blocking": false,
  "muting": false,
  "requested": false,
  "domain_blocking": false
}]
```

Then, when I open followee's profile on iOS apps...

* Amaroq **crashed**
* Pawoo couldn't show any relationsips, and follow/unfollow button disappears
* iMast, Tootle, friends.nico show as "not following"

also my friends says Tusky and SubwayTooter are also broken. I think those **breaking change** shouldn't be happen on Mastodon 2.1 because it's not a major version up.

* * * *

- [-] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

aschmitz:
This closes #5856 by restoring the existing behavior of the `muting` and `following` keys (returning booleans rather than truthy or false). It adds `showing_reblogs` and `muting_notifications` keys:

* `showing_reblogs` returns true if:
  1. You've requested to follow the user, with reblogs shown, or
  2. You are following the user, with reblogs shown.
* `muting_notifications` returns true if you have muted the user and their notifications as well.

### DIFF
--- a/app/javascript/mastodon/components/account.js
+++ b/app/javascript/mastodon/components/account.js
@@ -81,7 +81,7 @@ export default class Account extends ImmutablePureComponent {
         buttons = <IconButton active icon='unlock-alt' title={intl.formatMessage(messages.unblock, { name: account.get('username') })} onClick={this.handleBlock} />;
       } else if (muting) {
         let hidingNotificationsButton;
-        if (muting.get('notifications')) {
+        if (account.getIn(['relationship', 'muting_notifications'])) {
           hidingNotificationsButton = <IconButton active icon='bell' title={intl.formatMessage(messages.unmute_notifications, { name: account.get('username') })} onClick={this.handleUnmuteNotifications} />;
         } else {
           hidingNotificationsButton = <IconButton active icon='bell-slash' title={intl.formatMessage(messages.mute_notifications, { name: account.get('username')  })} onClick={this.handleMuteNotifications} />;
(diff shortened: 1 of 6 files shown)


aschmitz:
Note that I'm doing this in RelationshipSerializer rather than adding additional methods to AccountInteractions and AccountRelationshipsPresenter to avoid double-querying Follow, FollowRequest, and Mute. We can push the methods down to either if you prefer, it just feels like that would be extra unnecessary code and queries, so I did it this way.

aschmitz:
Oh, I can't request reviews. CC @Gargron, @nightpool, and @unarist, then.

Gargron:
I guess this is the simplest fix, even if the resulting JSON is not very elegant. I looked into trying to make a V2 version but realized it would not simply be a matter of creating one API method, it would require to create V2 versions of each method that returns a relationship. That being said, please fix tests.

Gargron (approved):

PR #5888, opened 2017-12-05
Change belongs_to_required_by_default to true

ykzts:
In Rails v5.x, belongs_to defaulted to `required: true`. But in Mastodon `ActiveRecord::Base. Belongs_to_required_by_default` is `nil`. This is caused by one of the gems (https://github.com/rails/rails/issues/23589)

Resolved with the solution at https://github.com/rails/rails/issues/23589#issuecomment-305557647.


akihikodaki (approved):

ykzts:
resolve conflicts.

PR #5890, opened 2017-12-05
Fix rspec

ClearlyClaire:
Fix specs involving `AccountSearchService` since its interface has changed.

### DIFF
--- a/spec/services/account_search_service_spec.rb
+++ b/spec/services/account_search_service_spec.rb
@@ -72,7 +72,7 @@ describe AccountSearchService do
         describe 'and there is no account provided' do
           it 'uses search_for to find matches' do
             allow(Account).to receive(:search_for)
-            subject.call('two@example.com', 10, false, nil)
+            subject.call('two@example.com', 10, nil, resolve: false)
 
             expect(Account).to have_received(:search_for).with('two example.com', 10)
           end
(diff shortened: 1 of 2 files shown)


spec/services/account_search_service_spec.rb
```diff
@@ -72,7 +72,7 @@
         describe 'and there is no account provided' do
           it 'uses search_for to find matches' do
             allow(Account).to receive(:search_for)
-            subject.call('two@example.com', 10, false, nil)
+            subject.call('two@example.com', 10, nil, resolve: false)
```
ykzts: https://github.com/tootsuite/mastodon/blob/82884ac5c43092c0b0625f7d81c930bb023c9469/app/services/account_search_service.rb#L6

I feel that the thing before change is correct. Am I looking at the right place?

ThibG: @ykzts indeed, you're looking at master, not at feature-lists-ui (on top of which this PR applies): https://github.com/tootsuite/mastodon/blob/feature-lists-ui/app/services/account_search_service.rb#L6

ykzts: Oh sorry. I misunderstood that this is for master branch.

Gargron (approved):

PR #5894, opened 2017-12-06
Remove rabl dependency

Gargron:
One dependency less


Gargron:
For some reason the tests start failing after rabl is removed from Gemfile. That is the only line that breaks all the tests. It's currently unclear to me why that happens, because nothing should be using rabl...

Gargron:
I have tracked the problem down to this:

With rabl present, `Oj.dump(a: 1)` returns `{"a":1}`, with rabl removed it returns `{":a":1}` (the colon of the ruby symbol present in the string).

It is currently unclear WTF is going on

Gargron:
FOUND THE CULPRIT

https://github.com/nesquena/rabl/blob/11c51e36987bdb7fc63cccbc499312ff8564a02a/lib/rabl/configuration.rb#L22

ykzts (approved):

PR #5906, opened 2017-12-06
fix weblate for ja

nightpool:
Currently, yaml2po crashes on values that are just integers. See here for more info: https://github.com/translate/translate/issues/3619

this is a workaround until the relevant bug is closed.

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -448,7 +448,7 @@ ja:
     expires_in_prompt: 無期限
     generate: 作成
     max_uses:
-      one: 1
+      one: '1'
       other: "%{count}"
     max_uses_prompt: 無制限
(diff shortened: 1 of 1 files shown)


config/locales/ja.yml
```diff
@@ -448,7 +448,7 @@ ja:
     expires_in_prompt: 無期限
     generate: 作成
     max_uses:
-      one: 1
+      one: "1"
```
ykzts: `"1"` -> `'1'`

double quote to single quote.

ykzts (changes requested):

config/locales/ja.yml
nightpool: ahhh I had a single quote but I thought the double quote would be better for consistency. my b. orz

ykzts (approved):

mayaeh:
I'm sorry, I didn't notice this problem.
Thank you.

PR #5918, opened 2017-12-07
Weblate translations

mastodon-weblate:
There are some issues with how Weblate re-formats the language files, **do not merge**


nightpool (changes requested):
(placeholder until we fix simple_form.pt.yml)

PR #5922, opened 2017-12-08
Weblate translations

Gargron:


nightpool (approved):
There were more PT simple form changes but I think theyre lost now. LGTM

PR #5923, opened 2017-12-08
Rate limit by user instead of IP when API user is authenticated

### LINKED ISSUE #668: Rate limit by user instead of IP

Gargron:
Fix #668

Thanks Gitlab for reference: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/config/initializers/rack_attack_global.rb

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,6 +1,48 @@
 # frozen_string_literal: true
 
 class Rack::Attack
+  class Request
+    def authorization
+      get_header('HTTP_AUTHORIZATION')   ||
+      get_header('X-HTTP_AUTHORIZATION') ||
+      get_header('X_HTTP_AUTHORIZATION') ||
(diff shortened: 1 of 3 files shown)


config/initializers/rack_attack.rb
```diff
@@ -1,31 +1,65 @@
 # frozen_string_literal: true
 
 class Rack::Attack
+  class Request
+    def authorization
```
aschmitz: Is this duck typing to satisfy Doorkeeper::OAuth::Token.authenticate when we pass it `self`?

aschmitz (commented):
This looks good overall. The limit might be a *bit* low for some API clients, but it looks like that's already the existing limit, so I guess it's probably fine for now.

config/initializers/rack_attack.rb
Gargron: Yes, it is. No other way to do it afaik.

abcang (approved):

PR #5942, opened 2017-12-09
Change streaming API URL when remote development

ykzts:
currently always `ws://localhost:4000`.


unarist:
How about using STREAMING_API_BASE_URL as well as production? It allows us to use different port for each instances or reverse proxy, and it's useful for running multiple dev instances on the same host.

ykzts:
> How about using STREAMING_API_BASE_URL as well as production? It allows us to use different port for each instances or reverse proxy, and it's useful for running multiple dev instances on the same host.

Hm, sounds good. I tried. cb82c8aff0acdec90ecc2eb8e6bd2c4133701182

unarist (approved):
Also STREAMING_API_BASE_URL allows dev to use `wss:`, which is required for https instance to simulate federation. Great 👍

PR #5957, opened 2017-12-10
Add preview image to photo preview cards

akihikodaki:
The below is a screenshot of a preview of one media attachment (left) and one embedded photo (right). Note that the preview is only shown after a card is loaded.
screen shot 2017-12-10 at 16 11 14

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -9,7 +9,7 @@ import StatusContent from './status_content';
 import StatusActionBar from './status_action_bar';
 import { FormattedMessage } from 'react-intl';
 import ImmutablePureComponent from 'react-immutable-pure-component';
-import { MediaGallery, Video } from '../features/ui/util/async-components';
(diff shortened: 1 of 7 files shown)


akihikodaki:
I would like to include this change to v2.1 because it already has another change for `preview_cards` table. (#5775).
The migration task `mastodon:maintenance:migrate_photo_preview_cards` can deal with those two changes at once.
cc: @Gargron

Gargron:
After

>If card is simply included in status and we ensure FetchLinkCardService happens before or during distribution, we won't need a separate real-time delivery mechanism for link cards.

is solved, this will be a nice change. But I think it's not needed right now because

>Note that the preview is only shown after a card is loaded.

And that would be a slightly confusing UI "jump" for users.

akihikodaki:
@Gargron
> And that would be a slightly confusing UI "jump" for users.

I see. What about applying only the change for database?

Gargron:
>What about applying only the change for database?

Okay, that is fine!

PR #5961, opened 2017-12-10
Do not try to rescue ActiveRecord error in Mastodon::Snowflake::Callbacks

### LINKED ISSUE #5960: Status creation fails on id collision in transaction
#5260 introduces `id` override, and it is expected to try up to 100 times if `id` collides. However, it does not really work because it can be in transaction (e.g. [ActivityPub implementation](https://github.com/tootsuite/mastodon/blob/4c6b5dbe96b565d3db3cbf0912f3b9911bc3922a/app/lib/activitypub/activity/create.rb#L23)), and `id` collision immediately causes rollback.

I do not think it really matters, but it is a bug anyway so here is an issue.

* * * *

akihikodaki:
Callbacks in `Mastodon::Snowflake::Callbacks` may run in a transaction, and it cannot continue the execution if it caused an error in such a condition.

Resolves #5960.

### DIFF
--- a/lib/mastodon/snowflake.rb
+++ b/lib/mastodon/snowflake.rb
@@ -4,25 +4,11 @@ module Mastodon::Snowflake
   DEFAULT_REGEX = /timestamp_id\('(?<seq_prefix>\w+)'/
 
   class Callbacks
-    def self.around_create(record)
+    def self.before_create(record)
       now = Time.now.utc
 
-      if record.created_at.nil? || record.created_at >= now || record.created_at == record.updated_at
-        yield
-      else
-        record.id = Mastodon::Snowflake.id_at(record.created_at)
-        tries     = 0
-
-        begin
-          yield
-        rescue ActiveRecord::RecordNotUnique
-          raise if tries > 100
-
-          tries     += 1
-          record.id += rand(100)
-
-          retry
-        end
+      if record.created_at.present? && record.created_at < now && record.created_at != record.updated_at
+        record.id = Mastodon::Snowflake.id_at(record.created_at, record.class.table_name)
       end
     end
   end
@@ -61,11 +47,9 @@ module Mastodon::Snowflake
     # external users cannot derive the sequence base given the
     # timestamp and table name, which would allow them to
     # compute the table ID sequence number.
-    def define_timestamp_id
-      return if already_defined?
-
+    def define_timestamp_id(name)
       connection.execute(<<~SQL)
-        CREATE OR REPLACE FUNCTION timestamp_id(table_name text)
+        CREATE OR REPLACE FUNCTION #{name}(table_name text, timestamp_for_id timestamp with time zone)
         RETURNS bigint AS
         $$
           DECLARE
@@ -75,7 +59,7 @@ module Mastodon::Snowflake
           BEGIN
             time_part := (
               -- Get the time in milliseconds
-              ((date_part('epoch', now()) * 1000))::bigint
+              ((date_part('epoch', timestamp_for_id) * 1000))::bigint
               -- And shift it over two bytes
               << 16);
 
@@ -85,7 +69,7 @@ module Mastodon::Snowflake
               substr(
                 -- Of the MD5 hash of the data we documented
                 md5(table_name ||
-                  '#{SecureRandom.hex(16)}' ||
+                  '#{old_salt || SecureRandom.hex(16)}' ||
                   time_part::text
                 ),
                 1, 4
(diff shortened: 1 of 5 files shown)


akihikodaki:
35e533cf102425ace91dbf01a3b6133105578439 resolves codeclimate's requests, and also uses `timestamp_id`.

Gargron:
I need @aschmitz for this

lib/mastodon/snowflake.rb
```diff
@@ -138,10 +124,10 @@ def ensure_id_sequences_exist
       end
     end
 
-    def id_at(timestamp)
-      id  = timestamp.to_i * 1000 + rand(1000)
+    def id_at(table_name, timestamp)
+      id  = timestamp.to_i * 1000
       id  = id << 16
-      id += rand(2**16)
+      id |= ApplicationRecord.connection.select_value('SELECT timestamp_id($1)', nil, [[nil, table_name]]) & 65_535
```
aschmitz: Does this really get us anything over the `rand(2**16)`? The returned ID is almost certainly not for the millisecond we're calling `id_at` for (as it's almost certainly being called for a historical time, not the current time), which means it's roughly random. If it's called several times in the same millisecond, we can at least be guaranteed that the low bits won't be the same, but I don't know how much that matters.

lib/mastodon/snowflake.rb
```diff
@@ -4,25 +4,11 @@ module Mastodon::Snowflake
...
-        rescue ActiveRecord::RecordNotUnique
-          raise if tries > 100
-
-          tries     += 1
-          record.id += rand(100)
-
-          retry
-        end
+      if record.created_at.present? && record.created_at < now && record.created_at != record.updated_at
+        record.id = Mastodon::Snowflake.id_at(record.class.table_name, record.created_at)
```
aschmitz: I gather you're doing this because `ActivityPub::Activity::Create#process_status` operates in a transaction, and so the RecordNotUnique could fail the transaction anyway.

This will just fail straight away instead. Is there anything to be gained from doing this, rather than changing line 19 to something like this?

~~~ ruby
raise if (tries > 100) || (ActiveRecord::Base.connection.open_transactions > 0)
~~~

That has the same effect inside a transaction, but not outside a transaction. (I'm not actually sure how frequently this would be called outside a transaction, but it's worth considering, maybe?)

aschmitz: Either way, it might be good to handle this outside of the transaction (for example, in `ActivityPub::Activity::Create#process_status`), if we're going to be concerned about the potential for duplicate IDs.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: > if we're going to be concerned about the potential for duplicate IDs

I'm not so I just removed it. We are not concerning that when we use `timestamp_id` anyway.

akihikodaki: GitHub went mad and posted a same comment so many times. Sorry for messing.

akihikodaki:
bd8f2e8fb303f6c4b35214839a88d186d17bdc07 introduces a new function to avoid collision.

PR #5967, opened 2017-12-10
Support GNU Social share URLs

### LINKED ISSUE #4940: Support GNU Social share URLs
GNU Social supports a less automated share functionality than Mastodon (see #4511). One needs to enter an URL of an instance to be redirected to a post composing form (you can see how it works at https://publiccode.eu/), Although the URL is different than in Mastodon: `/notice/new?status_textarea=Lorem+ipsum`, it is trivial to redirect such queries.

* * * *

ilpianista:
Closes #4940

### DIFF
--- a/app/controllers/intents_controller.rb
+++ b/app/controllers/intents_controller.rb
@@ -8,6 +8,8 @@ class IntentsController < ApplicationController
       case uri.host
       when 'follow'
         return redirect_to authorize_follow_path(acct: uri.query_values['uri'].gsub(/\Aacct:/, ''))
+      when 'notice'
+        return redirect_to share_path(text: uri.query_values['status_textarea'])
       when 'share'
         return redirect_to share_path(text: uri.query_values['text'])
       end


akihikodaki (changes requested):
Please add a comment to indicate the intent type is used by GNU Social.

ilpianista:
@akihikodaki isn't a `git blame` enough?

akihikodaki:
@ilpianista No. `git blame` is for reviewing history; it will be harder to figure out the intention of the code when they are rewritten, renamed, or splitted into multiple files. That is because `git blame` is not a tool for such a use.
Old good comments are no longer used to record revision history, but they are still alive to clarify the intention of the code.

nightpool (commented):
This was not included originally—in the absence of spec in this area we opted not to include the existing verbiage and I'd like to understand why before going back on that decision

Gargron:
I don't think there's any precedent in our code for supporting gnu social-specific features. If there was a high demand for this I would understand, but I have never ever seen a "share to gnu social" link in the wild where you could input a Mastodon server, so I don't think this change is justified.

ilpianista:
Please close #4940 as well.

steelman:
@Gargron 
I've mentioned one public [site](http://publiccode.eu) in my report which enables sharing via GS link, so you can go and see it for yourself. Searching the web for "Share on GNU Social" returns some more [hits](https://duckduckgo.com/?q=%22Share+on+GNU+Social%22). Not many, I admit. In my humble opinion, however, such tiny details, which make software operate a little better in the existing environment, are important. They are for me.

Thank you for Mastodon anyway.

PR #5968, opened 2017-12-10
Improve error handling in streaming/index.js

barzamin:
On an unhandled worker exception, we should log the exception
and gracefully exit with nonzero status, instead of letting workers
silently fail and restart in an endless loop.

Note: we previously tried to handle the `'error'` signal.
That's not a signal Node fires; my patch traps `'uncaughtException'`,
which is what the code was _trying_ to do.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -520,12 +520,14 @@ const startWorker = (workerId) => {
 
   const onError = (err) => {
     log.error(err);
+    server.close();
+    process.exit(0);
   };
 
   process.on('SIGINT', onExit);
   process.on('SIGTERM', onExit);
   process.on('exit', onExit);
-  process.on('error', onError);
+  process.on('uncaughtException', onError);
 };
 
 throng({


streaming/index.js
```diff
@@ -520,12 +520,14 @@ const startWorker = (workerId) => {
 
   const onError = (err) => {
     log.error(err);
+    server.close();
+    process.exit(0);
```
yipdw: If there's an error, do we want to exit with code zero or something else?  Would this prevent supervisors from restarting errored processes (e.g. systemd in `Restart=on-failure` mode)?

barzamin: throng will trap process.exit and reboot the worker; this is why we were hitting an error loop before.

Gargron:
I think some errors are worth restarting the workers for... But if the error is EADDRINUSE that should 1) display message in the logs 2) exit for good. That's been the Nr. 1 reason for 100% CPU restart cycles.

barzamin:
@Gargron good point. I'll look at the `throng` docs and figure out how to do that.

Gargron:
@barzamin something like `err.code === 'EADDRINUSE'`

barzamin:
@Gargron the issue, though, is that `throng` will restart the worker no matter what exit code we return. Not sure if there's a way to bypass the restart?

Gargron (approved):

PR #5974, opened 2017-12-11
Revert "Add check for visibility.nil? even though it can't ever be, t…

akihikodaki:
…o check for race conditions"

This reverts commit 8232f76c482d3046055bd7bf224ef7835d0fa399.

The error is unexpected and have not been observed for a long time.

Pawoo and Pawoo Music saw this error only on May 24, 2017. The exception hardly makes sense.

### DIFF
--- a/app/lib/exceptions.rb
+++ b/app/lib/exceptions.rb
@@ -4,7 +4,6 @@ module Mastodon
   class Error < StandardError; end
   class NotPermittedError < Error; end
   class ValidationError < Error; end
-  class RaceConditionError < Error; end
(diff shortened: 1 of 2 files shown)


Gargron:
>The exception hardly makes sense.

I agree with *that*, the column cannot even be nil, *however* I have observed this happening in the past. It had little reason to happen then as it has little reason to happen now, what I'm trying to say, it *could* happen.

If we found an explanation for why it happened I'd be okay with removing this check, but until then, I want to err on the side of caution.

PR #5977, opened 2017-12-11
make it possible to stream public timelines without authorization

nullkal:
Public timelines can be fetched via `/api/v1/timelines/public` without authorization, but streaming them needs authorization token now. This PR make it consistent between normal fetching APIs and streaming APIs.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -239,7 +239,7 @@ const startWorker = (workerId) => {
 
   const wsVerifyClient = (info, cb) => {
     const location = url.parse(info.req.url, true);
-    const auth_required = !PUBLIC_STREAMS.some(stream => stream === location.query.stream);
+    const authRequired = !PUBLIC_STREAMS.some(stream => stream === location.query.stream);
 
     accountFromRequest(info.req, err => {
       if (!err) {
@@ -248,7 +248,7 @@ const startWorker = (workerId) => {
         log.error(info.req.requestId, err.toString());
         cb(false, 401, 'Unauthorized');
       }
-    }, auth_required);
+    }, authRequired);
   };
 
   const PUBLIC_ENDPOINTS = [
@@ -264,8 +264,8 @@ const startWorker = (workerId) => {
       return;
     }
 
-    const auth_required = PUBLIC_ENDPOINTS.some(endpoint => endpoint === req.path);
-    accountFromRequest(req, next, auth_required);
+    const authRequired = !PUBLIC_ENDPOINTS.some(endpoint => endpoint === req.path);
+    accountFromRequest(req, next, authRequired);
   };
 
   const errorMiddleware = (err, req, res, {}) => {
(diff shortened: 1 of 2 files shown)


streaming/index.js
```diff
@@ -227,6 +227,19 @@ const startWorker = (workerId) => {
   };
 
   const wsVerifyClient = (info, cb) => {
+    const PUBLIC_STREAMS = [
```
abcang: I think that it is better to define PUBLIC_STREAMS and PUBLIC_ENDPOINTS outside the function.

abcang:
If authentication is skipped, since `req.accountId` does not have a value, I think you need to change some methods such as streamFrom. Especially streamFrom will filter on accounts if `needsFiltering` is true, so we need to fix.

streaming/index.js
```diff
@@ -208,42 +208,65 @@ const startWorker = (workerId) => {
...
+  ];
+
   const authenticationMiddleware = (req, res, next) => {
     if (req.method === 'OPTIONS') {
       next();
       return;
     }
 
-    accountFromRequest(req, next);
+    const auth_required = PUBLIC_ENDPOINTS.some(endpoint => endpoint === req.path);
```
abcang: Maybe `!PUBLIC_ENDPOINTS.some(endpoint => endpoint === req.path)`. Also, it is better to change the variable name to camel case.

streaming/index.js
```diff
@@ -313,26 +337,30 @@ const startWorker = (workerId) => {
             return;
           }
 
-          const queries = [
-            client.query(`SELECT 1 FROM blocks WHERE (account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 2)})) OR (account_id = $2 AND target_account_id = $1) UNION SELECT 1 FROM mutes WHERE account_id = $1 AND target_account_id IN (${placeholders(targetAccountIds, 2)})`, [req.accountId, unpackedPayload.account.id].concat(targetAccountIds)),
-          ];
+          if (req.accountId != null) {
```
abcang: I think that simply does not have any problem. `accountId` can never be 0.

```
if (!req.accountId) {
```

And you should not need to change `.eslintrc.yml`.

abcang (changes requested):

abcang (approved):

streaming/index.js
```diff
@@ -383,16 +411,18 @@ const startWorker = (workerId) => {
 
   // Setup stream end for WebSockets
   const streamWsEnd = (req, ws, closeHandler = false) => (id, listener) => {
+    const accountId = req.accountId || '(anonymous user)';
```
Gargron: Can we use IP instead?

streaming/index.js
```diff
@@ -275,8 +297,10 @@ const startWorker = (workerId) => {
   };
 
   const streamFrom = (id, req, output, attachCloseHandler, needsFiltering = false, notificationOnly = false) => {
+    const accountId = req.accountId || '(anonymous user)';
```
Gargron: Can we use IP instead?

Gargron:
I think conceptually I am okay with this. Originally I was against it because keeping open connections is a burden and allowing anonymous connections means you would have to block by IP instead of simply revoking access tokens.

However if this becomes a thing than the /about page could get real-time updates instead of poll intervals.

Gargron (approved):

PR #5979, opened 2017-12-11
Change account_id non-nullable in lists

akihikodaki:


db/migrate/20171201000000_change_account_id_nonnullable_in_lists.rb
```diff
@@ -0,0 +1,7 @@
+require Rails.root.join('lib', 'mastodon', 'migration_helpers')
+
+class ChangeAccountIdNonnullableInLists < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :lists, :account_id, false
```
Gargron: There *shouldn't* be any rows with a null here, but if by some chance there are? Maybe use the safe method from MigrationHelpers? If there are no rows with a null it shouldn't take longer.

Gargron (approved):

PR #5980, opened 2017-12-11
 Fan out preview card on write

akihikodaki:
Blocked by #5961.

Popular social media, including Twitter embeds or adds graphs for contents identified by URIs in texts. To achieve this feature on Mastodon clients, preview cards should also be streamed.
This change adds the feature to publish preview cards via Redis. Application developers may subscribe them, and stream to their Mastodon clients.

Note that this pull request does not include changes for the streaming server or those for Web client; I am considering to break up changes to support streaming preview cards because this pull request was growing (about 700 lines). Even though the major part is devoted for the specs and the actual code is not that big (~200 lines), it makes hard to review.


Gargron:
This can't make it into 2.1

I also think this might be the wrong approach. If `card` is simply included in `status` and we ensure FetchLinkCardService happens *before or during* distribution, we won't need a separate real-time delivery mechanism for link cards.

akihikodaki:
> This can't make it into 2.1

I do not think it should be included in 2.1 either. It is too late to review such a big change.

> I also think this might be the wrong approach. If card is simply included in status and we ensure FetchLinkCardService happens before or during distribution, we won't need a separate real-time delivery mechanism for link cards.

I did not think that is possible because of the delivery performance requirement. In #3136, @clworld suggested, ideally, the latency should be less than 500ms. However, FetchLinkCardService makes multiple HTTP requests, which could take more than one second.
If you think we can relax the performance requirement, please tell me. I will accordingly implement timeout for FetchLinkCardService.

Gargron:
@akihikodaki Facebook seems to solve this problem by performing preview even before status is submitted on the client-side. That might be a possible solution.

akihikodaki:
@Gargron That's also what Google+ does and it is really a nice solution for local statuses. However we also have federation. How should we deal with it?

Gargron:
@akihikodaki First we really need to fix the 3 HTTP requests into 2, because it does two GET requests only because the stupid oembed library is coded that way. Secondly I think that longer processing time for remote statuses is more or less invisible, so adding more processing time could be acceptable.

akihikodaki:
@Gargron 

> First we really need to fix the 3 HTTP requests into 2
I meant there should be three requests for the page identified by the URI in the text, the OGP/oEmbed representation, and an image, respectively. So it means the current implementation may make requests 3 or more. :( I will fix it.

> Secondly I think that longer processing time for remote statuses is more or less invisible, so adding more processing time could be acceptable.

There are some cases that is not true. Consider a case that some posted a status saying "This page seems to have heavy load. https://slow.cat/." That would stop even the post. There could also be DoS attack.
We need to consider how long latency is acceptable and implement timeout anyway.

PR #5985, opened 2017-12-12
Improve public status page title

Gargron:
Before:

> Mastodon

After:

> Display name: "Text of the toot truncated to 50 characters..." - Mastodon


ykzts (approved):

BartG95:
In case of a CW, it should show the CW, and not the actual content in the title?

PR #5987, opened 2017-12-12
Change conditional to avoid nil into string error in sidekiq

renatolond:
When obtaining information about users with mastodon in a different subdomain, sidekiq was giving out a 'no implicit conversion of nil into String'.
This was happening for instance with the user at https://mastodon.sergal.org/users/Sir_Boops.

### DIFF
--- a/app/services/fetch_remote_status_service.rb
+++ b/app/services/fetch_remote_status_service.rb
@@ -40,6 +40,6 @@ class FetchRemoteStatusService < BaseService
   end
 
   def confirmed_domain?(domain, account)
-    account.domain.nil? || domain.casecmp(account.domain).zero? || domain.casecmp(Addressable::URI.parse(account.remote_url.blank? ? account.uri : account.remote_url).normalized_host).zero?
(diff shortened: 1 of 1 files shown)


app/services/fetch_remote_status_service.rb
```diff
@@ -40,6 +40,6 @@ def process_atom(url, prefetched_body:)
   end
 
   def confirmed_domain?(domain, account)
-    account.domain.nil? || domain.casecmp(account.domain).zero? || domain.casecmp(Addressable::URI.parse(account.remote_url || account.uri).normalized_host).zero?
+    account.domain.nil? || domain.casecmp(account.domain).zero? || domain.casecmp(Addressable::URI.parse(account.remote_url.blank? ? account.uri : account.remote_url).normalized_host).zero?
```
ykzts: `account.remote_url.presence || account.uri`

renatolond: Didn't know that one, nice! :)
Changed

ykzts (approved):

Gargron (approved):

PR #6003, opened 2017-12-13
Add filters to admin UI for custom emojis

Gargron:


unarist:
Should not we keep the filter after copying remote emoji?

unarist (approved):

PR #5937, opened 2017-12-09
Update Chinese (Simplified) translations

ghost:
### References for review on 51631aea3532abcc26babb77c3fa8f3087d69056 and d11b461413312a4b9393ef118fbe9f1464ed29d6

_Take email notification for new follow requests as an example_

**BEFORE**
Subject: `待审核的关注者：exampleuser@example.com`
Body:
```
Example User，

exampleuser@example.com 请求关注你

查看： https://example.com/web/follow_requests

---

来自 example.com 的 Mastodon 通知
更改电子邮件首选项：https://example.com/settings/preferences
```

**AFTER**
Subject: `来自 exampleuser@example.com 的关注请求`
Body:
```
Example User：

exampleuser@example.com 向你发送了关注请求！

点此链接查看详情： https://example.com/web/follow_requests

---

这是一封来自 example.com 的 Mastodon 电子邮件通知。
使用此链接更改你的电子邮件首选项：https://example.com/settings/preferences
```

-------------
cc @Artoria2e5 @leopku
**Please also review other commits as well.**

PR #5911, opened 2017-12-07
Improve Chinese (Simplified) translations

ghost:
### References for review on e74accb1b5e38800ea672def1abaf9bff5da04f0
**BEFORE**

**AFTER**

"Moderation"

"Administration"

----------
@Artoria2e5 please review

PR #6005, opened 2017-12-13
Update Chinese (Simplified) translations

ghost:
WIP until all related PRs are merged and v2.1.0 is ready.
@Artoria2e5 @leopku please review

### DIFF
--- a/config/locales/zh-CN.yml
+++ b/config/locales/zh-CN.yml
@@ -161,6 +161,7 @@ zh-CN:
         update_status: "%{name} 刷新了 %{target} 的嘟文"
       title: 运营日志
     custom_emojis:
+      by_domain: 域名 
       copied_msg: 成功将表情复制到本地
       copy: 复制
       copy_failed_msg: 无法将表情复制到本地
--- a/config/locales/zh-CN.yml
+++ b/config/locales/zh-CN.yml
@@ -116,6 +116,7 @@ zh-CN:
       roles:
         admin: 管理员
         moderator: 协管
+        staff: 工作人员
         user: 普通用户
       salmon_url: Salmon URL
       search: 搜索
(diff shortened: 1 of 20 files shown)


config/locales/zh-CN.yml
```diff
@@ -116,6 +116,7 @@ zh-CN:
       roles:
         admin: 管理员
         moderator: 协管
+        staff: 工作人员
```
SerCom-KC: Obviously I have to change translations of `staff` because of this. Any other ideas?

SerCom-KC:
Not essential suggestion: QQ uses `快捷键` to refer to short codes for custom emojis.
Also, full approve needed for #5991, #5937 and #5911

config/locales/zh-CN.yml
SerCom-KC: @Gargron Yep, I know, the original translation for `staff` is `管理员`, which is the same as `admin`. Don't think `工作人员` (staff) fits well here, so I'm asking for suggestions :P

Gargron: It's okay to use "admin" imo. Rather than separate between all/user/mod/admin this filter only separates between all and mod+admin, the results for mod+admin will not be numerous in most circumstances so it's okay to not subdivide them.

Gargron (approved):

Artoria2e5:
@SerCom-KC not a fan of the "shortcode" change -- `#(滑稽)` is not usually seen as a "hotkey", neither should `:new_moon_with_face:`. everything else seems ok.

SerCom-KC:
@Artoria2e5
`短名` sounds less obvious when talking about indicating what shortcodes can do... I think I'll keep that unchanged for now.
So you're okay with `监察员` and all other stuff?

PR #6013, opened 2017-12-13
 Fix title of /terms page

ghost:
`example.com Terms of Service and Privacy Policy` → `example.com Terms of Service`
#5817


Gargron:
But why though?

SerCom-KC:
Since the disclaimer on the sign up page has changed in #5817?
If this is not necessary you can close it :P

PR #6015, opened 2017-12-13
Excahnge the order of spoiler-input and unlocked warning

lynlynlynx:
Why does unlocked account warning interrupt between CW input and toot input area?
I think this order is more natural.

image


Gargron (approved):

lynlynlynx:
Ahhh where whitespace come from... thanks so much Gargron 👍

PR #6016, opened 2017-12-13
Fix overflowing emojis on some devices

lynlynlynx:
I find out some devices display images in paragraph a bit higher, I can't check all devices, so I set overflow as visible.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -604,7 +604,7 @@
   line-height: 20px;
   word-wrap: break-word;
   font-weight: 400;
-  overflow: hidden;
+  overflow: visible;
   white-space: pre-wrap;
 
   &:focus {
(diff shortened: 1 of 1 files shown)


app/javascript/styles/mastodon/components.scss
```diff
@@ -604,7 +604,7 @@
   line-height: 20px;
   word-wrap: break-word;
   font-weight: 400;
-  overflow: hidden;
+  overflow: visible;
```
Gargron: This is a bad idea because this line was guarding against zalgo text going outside a single toot's area.

Gargron (changes requested):

app/javascript/styles/mastodon/components.scss
lynlynlynx: I understand, but emojis have negative margin, it is difficult to display well in all unknown devices without space between status__info and status__content.
I update this pull request, add 2px space, like suggention 2 of https://github.com/tootsuite/mastodon/pull/5474#issuecomment-349231615 .

Gargron (approved):

PR #6023, opened 2017-12-14
Mark sensitive when receive from domain blocked instance

ykzts:
I want to give the NSFW flag to all the media received from instances where many users forgot to attach the NSFW flag (e.g. Pawoo).


nullkal:
It would be helpful if there is "Mark all images as NSFW" feature aside of "Reject media files". Maybe make a dropdown box like as "severities"?

SerCom-KC:
Consider modify `admin.domain_blocks.new.severity.desc_html` as well?

Like changing
`Use <strong>None</strong> if you just want to reject media files.`
to
`Use <strong>None</strong> if you just want to reject and/or sensitive media files.`

ykzts:
@nullkal sounds good.

yukimochi:
I think this feature (now implementation) is a little danger.

Mark sensitive feature makes a risk boosting NSFW picture for outor instance carelessly.
So, I think require the way to know " this picture is masked by admin (not by author)".

ykzts:
hm. I want time to think about specifications.

PR #6026, opened 2017-12-14
Prevent nested migrated accounts, or migrations to self

### LINKED ISSUE #6022: Account migration allow loop
Account migration setting allow redirect user himself.
(on 2.1.0 rc5)

* * * *

Gargron:
Fix #6022

### DIFF
--- a/app/controllers/settings/migrations_controller.rb
+++ b/app/controllers/settings/migrations_controller.rb
@@ -28,6 +28,7 @@ class Settings::MigrationsController < ApplicationController
   end
 
   def migration_account_changed?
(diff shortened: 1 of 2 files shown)


mkljczk (approved):

ykzts (approved):

nightpool:
So this doesn't provide any sort of error to the client, it just fails silent when serializing?? This should at *least* be a validation....

Gargron:
@nightpool I will *not* add any more localizeable strings at this stage, it's a nightmare. So yes.

PR #6029, opened 2017-12-14
“Two-factor code” → “Two-factor authentication code”

mkljczk:
It has more sense for me…


nightpool:
I don't think there's much chance for confusion here....

m4sk1n:
So maybe label it as “suggestion”?

m4sk1n:
Should I close this PR?

PR #6030, opened 2017-12-15
Do not hide statuses from silenced accounts from other silenced accounts

Gargron:


nightpool:
Personally I think this is a bad idea—I know the original conception of silenced was a sort of hellban, but right now it's very noticeable when you get silenced on a small instance, because the local timeline disappears. Adding things that make it *more* noticeable seems like the wrong direction.

Gargron:
@nightpool I think you misunderstand. This makes it *less* noticeable, not more. Originally silenced accounts could see their own posts and posts of other silenced accounts. But due to the implementation of StatusFilter that became bugged and silenced accounts could no longer see even *their own* posts in threads, which made the shadowban *very* obvious and annoying.

nightpool (approved):

nightpool (approved):
okay, right, that makes sense.

nightpool:
Do we know when this changed? I don't remember it happening last time I was poking the threading code....

PR #6036, opened 2017-12-15
 Do not show desktop notification settings if not supported

### LINKED ISSUE #4236: Desktop notifications settings visible even if they are not supported
We currently detect push notification support and only show the toggles if they are supported. We should do the same for regular desktop notifications.

The current behaviour is confusing - toggling the settings makes no difference on Chrome on Android and iOS (among others).

* * * *

akihikodaki:
This resolves #4236.

~~Note that the initial state of `push_notifications` reducer is also updated accordingly to reflect the result of the feature detection to the default values. Using the same result for the default values and the UI eliminates the potential inconsistency.~~

By the way, the state of `push_notifications` reducer should not have been used as the default preferences. The original code intended to send the notification preferences as `alerts` property of `data` object *only* if they are saved. Otherwise, the server should have done the feature detection and have set the default parameters depending on that. However, there is a bug that it *always* sends the notification preferences loaded from the `push_notifications` state. ~~I just exploited that to implement this.~~

### DIFF
--- a/app/javascript/mastodon/reducers/push_notifications.js
+++ b/app/javascript/mastodon/reducers/push_notifications.js
@@ -1,5 +1,5 @@
 import { STORE_HYDRATE } from '../actions/store';
-import { SET_BROWSER_SUPPORT, SET_SUBSCRIPTION, CLEAR_SUBSCRIPTION, ALERTS_CHANGE } from '../actions/push_notifications';
+import { SET_SUBSCRIPTION, CLEAR_SUBSCRIPTION, ALERTS_CHANGE } from '../actions/push_notifications';
 import Immutable from 'immutable';
 
 const initialState = Immutable.Map({
@@ -11,7 +11,6 @@ const initialState = Immutable.Map({
     mention: false,
   }),
   isSubscribed: false,
-  browserSupport: false,
 });
 
 export default function push_subscriptions(state = initialState, action) {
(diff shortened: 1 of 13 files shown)


app/javascript/mastodon/reducers/push_notifications.js
```diff
@@ -1,17 +1,17 @@
...
+import { supportsDesktopNotifications } from '../agent';
 
 const initialState = Immutable.Map({
   subscription: null,
   alerts: new Immutable.Map({
-    follow: false,
-    favourite: false,
-    reblog: false,
-    mention: false,
+    follow: !supportsDesktopNotifications(),
```
sorin-davidoi: Can't check right now, but I think that Firefox for Android supports regular notifications, so this will be disabled.

akihikodaki: I heard that they are enabled for mobile devices because they do not support regular notifications, so I thought they should be disabled in the case. Is that wrong?

akihikodaki: I could not find any comments saying so. Probably that's just my misconception. I'll fix it soon.

akihikodaki:
496c2ede3ef1d06b3157e86b2b026fdf28b0d7e7: Addressed https://github.com/tootsuite/mastodon/pull/6036#pullrequestreview-83846740.

PR #6037, opened 2017-12-15
Do not override the default push notification settings

akihikodaki:
This resolves a bug described in #6036.

### DIFF
--- a/app/javascript/mastodon/actions/push_notifications/registerer.js
+++ b/app/javascript/mastodon/actions/push_notifications/registerer.js
@@ -51,13 +51,6 @@ export function register () {
   return (dispatch, getState) => {
     dispatch(setBrowserSupport(supportsPushNotifications));
 
-    if (me && !pushNotificationsSetting.get(me)) {
-      const alerts = getState().getIn(['push_notifications', 'alerts']);
-      if (alerts) {
-        pushNotificationsSetting.set(me, { alerts: alerts });
-      }
-    }
-
     if (supportsPushNotifications) {
       if (!getApplicationServerKey()) {
         console.error('The VAPID public key is not set. You will not be able to receive Web Push Notifications.');


abcang:
Why is this change necessary? Because pushNotificationsSetting operates only localStorage, store of redux seems not to be related.

By the way, there is a bug in this implementation about being stored in localStorage with or without subscription. I think that it will be fixed by changing to this implementation.

```javascript
if (!pushNotificationsSetting.get(me)) {
  const alerts = store.getState().getIn(['push_notifications', 'alerts']);
  if (store.getState().getIn(['push_notifications', 'subscription']) && alerts) {
    pushNotificationsSetting.set(me, { alerts: alerts });
  }
}
```

Is there any reason you do not want to save settings to localStorage when the subscription exists?

akihikodaki:
@abcang
> Why is this change necessary? Because pushNotificationsSetting operates only localStorage, store of redux seems not to be related.

It is overriding the settings on the server side. The user agent loads the settings from the local storage, and sends it with the subscription. (see `sendSubscriptionToBackend`) That is the problem this change addresses.

> By the way, there is a bug in this implementation about being stored in localStorage with or without subscription. I think that it will be fixed by changing to this implementation.
> ```JavaScript
> if (!pushNotificationsSetting.get(me)) {
>   const alerts = store.getState().getIn(['push_notifications', 'alerts']);
>   if (store.getState().getIn(['push_notifications', 'subscription']) && alerts) {
>     pushNotificationsSetting.set(me, { alerts: alerts });
>   }
> }

`alerts` is *always* `Immutable.Map`.

> Is there any reason you do not want to save settings to localStorage when the subscription exists?

As I described the above, they will be mistakenly sent to the server and overrides the default preferences. And you do not even have to do so since the preferences included in the server response will eventually be saved. (see `register` function.)

abcang (approved):

akihikodaki:
472f57e125c3cfbc909fcd3a31ddf473dfdf51ea: rebased to 138512d20414de5782deca515b35d1d4f5406e1c.

PR #6039, opened 2017-12-15
Remove period from the version number

zunda:
2.1.0. -> 2.1.0


Gargron:
It's kinda too late now 😭

Gargron (approved):

PR #6047, opened 2017-12-16
Refactor web_push_subscription

akihikodaki:
`web_push_subscription` lived out of Flux architecture, but it should be in the framework because it uses Redux store. Now it is a normal action.
It also makes easy to locate the implementation by getting `register` method and `saveSettings` method together in a module.


ykzts:
cc @sorin-davidoi

sorin-davidoi:
Didn't test, but the changes look good 👍

ykzts (approved):

PR #6068, opened 2017-12-19
delete X-UA-Compatible

Otakan951:
Mastodon doesn't work on IE, but it has "X-UA-Compatible" content="IE=edge".

### DIFF
--- a/spec/fixtures/requests/idn.txt
+++ b/spec/fixtures/requests/idn.txt
@@ -11,13 +11,14 @@ Accept-Ranges: bytes
 <html>
 	<head>
 		<meta charset="UTF-8">
+		<meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1" />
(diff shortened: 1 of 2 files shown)


nightpool:
@nolanlawson thoughts?

akihikodaki (approved):

spec/fixtures/requests/idn.txt
```diff
@@ -11,7 +11,6 @@ Accept-Ranges: bytes
 <html>
 	<head>
 		<meta charset="UTF-8">
-		<meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1" />
```
ykzts: I think it is unnecessary.

yumetodo: @ykzts Do you mean this should be replaced to `<meta http-equiv="X-UA-Compatible" content="IE=edge" />`?(Google Chrome Frame is not supported)

ykzts: This file is used for testing only. It simulates the contents of an external web page and has nothing to do with Mastodon.

PR #6071, opened 2017-12-19
Add confirmation step for email changes

### LINKED ISSUE #3871: No verification email for changing email address
It seems that currently there's no verification email sent for changing email address.
I'd like to confirm why we don't have it.
I guess there could be 2 problems if we don't have verification for changing email address.
1. Mastodon could be an auto spam maker
2. It hinders logging in if users submit their email address with mistyping

As for the first problem, I assume following case.
1) Create an account and raise it to have much followers
2) Change notification settings to send email for every events
3) Change email address to someone's one whom the account owner wants to harm
4) Act something to send spam email

For this case, we also can see another problem. It is that notification email contains the link to see the post but he/she can't see the post if he/she doesn't have any accounts in the instance.

If the instance owner doesn't provide any public contact such as email for /about/more, basically there's nothing victims can do to stop spams.

As for the second problem, as you may know, people always do mistake.
If he/she change his/her email address with typo, they can't log in the instance forever.

* * * *

pfigel:
This adds a confirmation step for email changes of existing users. Like the initial account confirmation, a confirmation link is sent to the new address.

Additionally, a notification is sent to the existing address when the change is initiated. This message includes instruction to reset the password immediately or to contact the instance admin if the
change was not initiated by the account owner.

Fixes #3871


app/mailers/user_mailer.rb
```diff
@@ -13,7 +13,9 @@ def confirmation_instructions(user, token, **)
     return if @resource.disabled?
 
     I18n.with_locale(@resource.locale || I18n.default_locale) do
-      mail to: @resource.unconfirmed_email.blank? ? @resource.email : @resource.unconfirmed_email, subject: I18n.t('devise.mailer.confirmation_instructions.subject', instance: @instance)
+      mail to: @resource.unconfirmed_email.blank? ? @resource.email : @resource.unconfirmed_email,
+           subject: @resource.pending_reconfirmation? ? I18n.t('devise.mailer.reconfirmation_instructions.subject', instance: @instance) : I18n.t('devise.mailer.confirmation_instructions.subject', instance: @instance),
```
akihikodaki: You can write like this, which is shorter:
```Ruby
I18n.t(@resource.pending_reconfirmation? ? 'devise.mailer.reconfirmation_instructions.subject' : 'devise.mailer.confirmation_instructions.subject', instance: @instance)
```

app/views/user_mailer/email_changed.en.html.erb
```diff
@@ -0,0 +1,15 @@
+<p>Hello <%= @resource.email %>!</p>
+
+<% if @resource.try(:unconfirmed_email?) %>
```
akihikodaki: Please use `&.` (safe navigation operator)

akihikodaki (approved):
It would be nice to have this security improvement.

However, I clarify that this is *not enough*. Because a password and an OTP can be reset by email confirmation, the email address change must be protected by them as well.

That can be implemented in a later pull request anyway, so this comment is not blocking this pull request to be merged.

akihikodaki (changes requested):
Sorry for messing, but I meant to request some changes. The logic is fine but I have some suggestions: see the last review.

patf:
Thanks for the review! Code changes have been pushed.

> However, I clarify that this is not enough. Because a password and an OTP can be reset by email confirmation, the email address change must be protected by them as well.

I'm not sure I follow. Being able to reset both the password and the OTP by having access to the email would reduce the existing authentication logic to a single factor (email), I don't think that is the case? Perhaps I'm misunderstanding what you meant. 😁

That said, I've looked at a large number of implementations (including Facebook, Twitter, Google, GitHub and Amazon), and they all gate email changes behind only one factor (typically the password, GitHub allows either). I believe the assumed threat model is a hijacked session (via stolen cookies, unlocked devices, ...), so one factor sufficiently mitigates this.

akihikodaki (approved):
The changes I requested in the last review are all applied.

akihikodaki:
> Being able to reset both the password and the OTP by having access to the email would reduce the existing authentication logic to a single factor (email), I don't think that is the case? 

That is *fine* if the email account is secured with 2FA (and it is likely to be secured better using SMS and other means, which are not possible for Mastodon.)

app/models/user.rb
```diff
@@ -41,12 +41,15 @@ class User < ApplicationRecord
 
   ACTIVE_DURATION = 14.days
 
-  devise :registerable, :recoverable,
-         :rememberable, :trackable, :validatable, :confirmable,
-         :two_factor_authenticatable, :two_factor_backupable,
-         otp_secret_encryption_key: ENV['OTP_SECRET'],
+  devise :two_factor_authenticatable,
```
Gargron: Can you do this? Split devise definitions? That's not obvious to me. What if the last one takes precedence? Why is this change needed anyway?

Gargron (changes requested):

Gargron:
As a general comment: Please mind that we used to have something like this (we were using unconfirmed_email and stuff) and then we switched to not doing it because of weird bugs. It was a year ago though so I don't remember any details, so just... tread carefully.

app/models/user.rb
patf: Sorry, should have added a comment for that portion. Yes, that works. In fact it was necessary to get devise to send the notification immediately after the change was requested (and before it's confirmed). Perhaps that's the weird bug you're recalling? I remember checking the commit history and seeing that `config.send_email_changed_notification` was removed when `devise-two-factor` was added - likely because of this. IIRC it's a side-effect of not using `DatabaseAuthenticatable` directly, which is where the notification is implemented, and splitting the definition resolves it. (I don't seem to recall the exact root cause. 😞)

You can check [GitLab](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/models/user.rb) for a similar pattern.

Gargron: Alright, thanks for clearing it up!

Gargron (approved):

davidosborne97 (commented):
Stealing is wrong

PR #6075, opened 2017-12-20
Reduce the number of synchronous resolves when posting toots

ClearlyClaire:
In order to gracefully handle OStatus → ActivityPub migration, Mastodon resolves mentioned remote accounts.
This is only needed for OStatus accounts mentioned in private/direct toots, though.
This PR restricts those synchronous resolves to those cases.


ykzts (approved):
looks good.

PR #6077, opened 2017-12-21
Add OTP and password confirmation for 2FA settings

### LINKED ISSUE #3961: 2FA recovery codes shouldn’t be regenerable
With 2-factor authentication, there are obviously more ways to be unable to login (like if you’re on your front door and you forgot your keys) than without. That means more human mistakes, and more users complaints/issues. So implementing recovery codes has been a convenient way to provide secure login even if they lost their phone/keys/whatever.

Recovery codes serve as a list of single-use 2FA codes, that the user has to store securely when 2FA is activated, so, if or when the secret is lost or forgotten (broken phone, for example), they can login once, and deactivate 2FA, or regenerate the token. The whole point of recovery codes are that the list is small (but sufficient to prevent basic mistakes) so the attack surface is minimal, and that those codes are kept secret after generation.

Mastodon currently provides a way to regenerate the recovery codes on demand, without password confirmation. This is a security flaw, as anyone accessing an unlocked computer can regenerate, and store, a new list of recovery codes (thus making pretty useless the usage of 2FA). In this case, the attacker, given they know the password (because, you know, it’s [hunter2](https://mostsecure.pw) — just kidding, don’t use it —), they’re able to login **at any time** (which is pretty bad for TOTP), and do anything to the account.
Even without password access, a rogue recovery codes regeneration causes the user to lose possibility to login entirely if they lose the 2FA secret.

There is also an UX issue here: there is no way for a user to know if their codes are the latest generated, and misclicking the "regenerate" button can happen, immediately disabling old codes (and the user may not have the reflex to store/print those and revoke the obsolete ones).

So, basically, the problem is that a user can regenerate the recovery codes list. Adding password confirmation should very slightly increase security, but it’s not a serious workaround.

My suggestion would be to completely remove this regeneration feature.

* * * *

pfigel:
In order to make it harder for an attacker to persistently hijack an account given a stolen session, an OTP and password confirmation step is added for all two-factor related settings, including disabling two-factor auth and regenerating recovery codes.

Additionally, a notification email is sent to the user when two-factor auth is disabled or recovery codes are regenerated.

Fixes #3961


akihikodaki (changes requested):
This is not enough to mitigate threats.

Consider the case a attacker obtained the password and the current OTP, and he wants to keep the chance to gain the control of the account later, but not now. Such a scenario should be considered because OTP is the very mechanism to keep resource safe in such a case.

The attacker can regenerate the recovery token using the OTP and the password, and keep it secret. The victim do not know about that, but the attacker is able to take control of the account anytime (NOT one time!)

To mitigate such a use case, you should confirm he has:
* the current backup code OR
* ~~the control of the email account (which is the last resort of authentication)~~

patf:
This is a bit more than most mainstream 2FA implementations ask for to regenerate recovery codes, but I can't think of any regressions this introduces and it does add a bit of an additional defensive layer beyond the simple "stole an existing session" scenario.

I'm thinking of a slightly different approach. In order to regenerate the recovery codes, we should ask for:

- the user's password
- any valid 2FA token (OTP or a recovery code)

After regenerating the code, we should inform the user about this via email. (If the attacker decided to also change the email, the victim will be made aware of that as of #6071.)

Allowing the user to enter an OTP takes care of the "lost recovery codes" scenario, which will likely be common for users wishing to regenerate the codes in the first place. In a phishing scenario, the attacker would not be able to use the same OTP to both login *and* reset the recovery codes (due to `consumed_timestep`).

I'm refraining from adding an email "backdoor" as that would effectively reduce the security to one factor (the password) in the case of password reuse, which is something that 2FA is supposed to make less of a problem.

Does this sound okay to you?

akihikodaki:
@patf I'm convinced the email authentication is insufficient, but OTP confirmation is also insufficient.
First, let me clarify that the most important feature of 2FA is adding a *different* kind of factor. In this case, a factor which is safe even if it is temporarily compromised.

Does the email authentication have the factor? I thought so but not necessarily, especially when concerning social hacking.
Then what about OTP? That has a problem I raised in the last comment.

So we do not have any means to regenerate recovery code safely? Well, I admit that. __The mechanism of recovery code does not seem to work.__ Here I'm suggesting to remove the feature of recovery code and to add an alternative.

In the below, I discuss about an alternative for recovery code, though it is kind of off topic (and long).

Email confirmation for resetting the OTP secret. That is actually so simple and may seem to have the same problem with the email confirmation for recovery code generation, but it does not have.
In the contrary to the recovery code regeneration, the user would soon be aware of OTP secret reset, and does not have the problem I described in the last comment.
The one-time factor is also preserved. Even if the email account is compromised once, the user can recover the Mastodon account when he recovered the email account.
This alternative would work well when the email account is secured with Mastodon's 2FA or equivalent. Actually the email account providers tend to have a good 2FA implementation (and they can use SMS to reset 2FA!), so delegating the security feature to the email account would be nice.

patf:
If I understand correctly, your suggested alternative would still effectively reduce Mastodon's authentication mechanism to one factor - the email address. Consider the following scenario:

1. Attacker gains access to a user's email account
2. Attacker performs a password reset
3. Attacker uses the newly-set password to perform the first login step
4. Attacker performs a OTP secret reset
5. Attacker uses a TOTP derived from the new OTP secret to successfully log in
6. Attacker changes the email address to one under their control in case the email account is recovered, giving them persistent access

Regarding the fourth step, I'm not certain if your suggestion was to allow the secret reset similar to a password reset (i.e. outside of an existing session). If you didn't mean to allow for that, we'd be effectively locking out users who lose their OTP secret and don't happen to have an existing session available, no? This would not be an uncommon scenario - think: users who only use Mastodon on their phone who happen to lose their phone (which also serves as the TOTP device).

akihikodaki:
@patf Even in such a scenario, the user can know there was an attack. That is better than being stolen the regeneration key without knowing an attack happened.

> I'm not certain if your suggestion was to allow the secret reset similar to a password reset (i.e. outside of an existing session).

Yes, OTP reset should be allowed outside of a session.

>  think: users who only use Mastodon on their phone who happen to lose their phone (which also serves as the TOTP device).

The email account (or the phone) should still be protected by a password. Otherwise the condition I described ("the email account is secured with Mastodon's 2FA or equivalent") is violated and the account would be compromised. I'm certain a user who enable OTP would protect his phone with password or equivalent.

patf:
> Even in such a scenario, the user can know there was an attack. That is better than being stolen the regeneration key without knowing an attack happened.

This can be achieved without basically making the email address a SPOF for account compromises, using the approach I described in my first comment.

There's a separate discussion hidden in all this where we're really weighing the benefits of things like email-link- (à la Slack) or OAuth2-style authentication ("social logins"), where your account security is effectively inherited (though I'd have some concerns regarding centralization), but I feel that we should not add an email backdoor for users who go out of their way to enable a second factor. Recovery codes are as close to an industry standard as you can get in that space, and the amended regeneration procedure would cover more bases than just about all major 2FA implementations.

> think: users who only use Mastodon on their phone who happen to lose their phone (which also serves as the TOTP device).

To clarify: I mentioned this scenario in case your solution would now allow OTP secret reset outside of existing sessions, to demonstrate that this would effectively lock out the user permanently, not as part of the threat model.

akihikodaki:
> > Even in such a scenario, the user can know there was an attack. That is better than being stolen the regeneration key without knowing an attack happened.
>
> This can be achieved without basically making the email address a SPOF for account compromises, using the approach I described in my first comment.

What do you mean by "this"? To eliminate attacks which the user cannot be aware of? Or securing the account in the scenario you described before?

> I feel that we should not add an email backdoor for users who go out of their way to enable a second factor.

2FA and disabling the email backdoor are different topics. For example you may add "account protection token" feature, which disables the email backdoor and use the token instead.

> > think: users who only use Mastodon on their phone who happen to lose their phone (which also serves as the TOTP device).
>
> To clarify: I mentioned this scenario in case your solution would now allow OTP secret reset outside of existing sessions (which wasn't the case), to demonstrate that this would effectively lock out the user permanently, not as part of the threat model.

It was the case. I'm suggesting to allow OTP secret reset outside of existing sessions, and that can lock out the user. However the user can know such an attack happened.

patf:
> What do you mean by "this"? To eliminate attacks which the user cannot be aware of? Or securing the account in the scenario you described before?

Making the user aware of the attack. The user would receive either an "email changed" notification (if the attacker changed the email) or a "recovery code regenerated" notification.

> 2FA and disabling the email backdoor are different topics. For example you may add "account protection token" feature, which disables the email backdoor and use the token instead.

I don't think we can ignore it in this topic. Currently, users who have 2FA enabled do not have an email backdoor - access to the email inbox would not reveal the OTP. This would change with your approach.

> It was the case. I'm suggesting to allow OTP secret reset outside of existing sessions, and that can lock out the user. However the user can know such an attack happened.

Sorry, it was supposed to read "not allow OTP secret reset outside of existing sessions". 😁

akihikodaki:
@patf
> Making the user aware of the attack. The user would receive either an "email changed" notification (if the attacker changed the email) or a "recovery code regenerated" notification.

How do you notify? If the attacker is controlling the email account, email could be dismissed.

> I don't think we can ignore it in this topic. Currently, users who have 2FA enabled do not have an email backdoor - access to the email inbox would not reveal the OTP. This would change with your approach.

I don't think anyone is aware of the removal of email backdoor.

> Sorry, it was supposed to read "not allow OTP secret reset outside of existing sessions". 😁

I see.

patf:
> How do you notify? If the attacker is controlling the email account, emails could be dismissed.

This also applies to your approach, no?

> I don't think anyone is aware of the removal of email backdoor.

I don't think the question is whether users are aware of it. What's important is that this would *add* an attack vector that is currently mitigated for users with two-factor authentication enabled.

I don't want to base my whole argument on an appeal to authority, but we should not ignore the myriad of existing 2FA implementations that have a recovery code mechanism and don't offer a way to reset the *second* factor via email. It would be surprising to see security teams from that many companies come up with similar implementations if the email alternative truly would be a better trade-off.

akihikodaki:
@patf 
> > How do you notify? If the attacker is controlling the email account, emails could be dismissed.

> This also applies to your approach, no?

No. In my approach, any attacks cannot be done without making changes users will be aware. (A user will know his OTP is reset because he can no longer login with an OTP generated with the old OTP seed.)

> I don't think the question is whether users are aware of it. What's important is that this would *add* an attack vector that is currently mitigated for users with two-factor authentication enabled.

If it is necessary to eliminate a bigger attack vector (recovery code stealing without user awareness), it is fine.
However if users were aware of it, users may manage their credentials according to their old understanding, and make a security hole.

> I don't want to base my whole argument on an appeal to authority, but we should not ignore the myriad of existing 2FA implementations that have a recovery code mechanism and don't offer a way to reset the second factor via email. It would be surprising to see security teams from that many companies come up with similar implementations if the email alternative truly would be a better trade-off.

I would like to ask you for examples.

akihikodaki:
I have just tested to reset the password of Microsoft account, and it asked me to provide *two* of the followings:
* OTP
* Other communicative means (SMS, other email accounts)
* Recovery token

It also said it is possible to recover my account after 30 days even if I have just one.
Their system implies:
* If we are allowing to reset password and OTP with just a email account, we better add some duration.
* Recovery token is for recovering the account, but not only for recovering OTP.

patf:
> No. In my approach, any attacks cannot be done without making changes users will be aware. (A user will know his OTP is reset because he can no longer login with an OTP generated with the old OTP seed.)

Detection solely through the OTP secret rotation seems fairly unlikely. It is common for users to remain logged in for months at a time - for example, all of my current sessions were created more than two months ago. Even if the user happens to need to login after the compromise, the attacker could simply disable two-factor authentication. The lack of a OTP prompt is not an immediate red flag, and in some cases is even expected: some login implementations occasionally ask for the password to renew the session, but skip the second factor if the user has a valid "remember me" token.

> If it is necessary to eliminate a bigger attack vector (recovery code stealing without user awareness), it is fine.

I don't agree that that the possibility of stealing recovery codes (given the password **and** a valid OTP) is a bigger attack vector than simply having an email account compromised, especially when the solution does not in any way *guarantee* detection either.

> I would like to ask you for examples.

Sure. Some examples from popular services I use:

- Google: Existing recovery codes are accessible from a logged-in session after a password prompt. No notification after regeneration.
- Facebook: Same, except codes are always regenerated.
- Apple: Same as Facebook.
- GitHub: Same as Google, except that it accepts either the password or an OTP to access or regenerate the codes.
- Twitter: Same as Google, except there is only one code.
- Amazon: Does not have any recovery codes; allows multiple TOTP devices and phone-based 2FA instead.
- MailChimp: Recovery code can be obtained by toggling the two-factor setting. No separate password prompt. The OTP seed and recovery code is not rotated during this.
- SendGrid: Uses Authy's first-party integration, which comes with its own recovery mechanism (IIRC via phone or backup password).

akihikodaki:
Sorry, I missed your reply.

> Detection solely through the OTP secret rotation seems fairly unlikely. It is common for users to remain logged in for months at a time - for example, all of my current sessions were created more than two months ago. Even if the user happens to need to login after the compromise, the attacker could simply disable two-factor authentication. The lack of a OTP prompt is not an immediate red flag, and in some cases is even expected: some login implementations occasionally ask for the password to renew the session, but skip the second factor if the user has a valid "remember me" token.

Invalidating sessions with duration can be a solution.

> I don't agree that that the possibility of stealing recovery codes (given the password and a valid OTP) is a bigger attack vector than simply having an email account compromised, especially when the solution does not in any way guarantee detection either.

I am not convinced that password and a valid OTP are enough yet. It eliminates the one-time factor.

> Sure. Some examples from popular services I use:

Microsoft does not notify when generating recovery code either, but they requires other communicative means to use the recovery code. The notification for recovery code generation is not the only possible solution, but there could be alternatives.

patf:
I made some changes based on this discussion that should further improve the implementation. To summarize:

- We now require both the password and OTP for any two-factor related action
- Notification emails are sent both when codes are regenerated and when two-factor auth is disabled

This matches or exceeds what most two-factor/recovery code implementations do (for a comparison, see my previous comment) and is at the very least an improvement over what we have today.

patf:
Fixed some conflicts and changed the mailers to use the new email templates. Please re-review, thanks!

PR #6085, opened 2017-12-22
Add rake task to check and purge accounts that are missing in origin

Gargron:
Maybe this is useful? :thinking: 

image

image

The task checks every remote ActivityPub account by performing a HEAD request at the account's URI. We only check ActivityPub because OStatus account's can be a bit more finicky I imagine. The number of accounts that will be checked is pre-counted and displayed in the progress bar. The accounts are iterated in a "partitioned" way, i.e. the domains are rotated in such a way that no subsequent requests are made to the same domain and they are spread out as much as possible, to prevent overloading any particular domain.

Unfortunately the progress bar appears anew under every prompt. Maybe we should get rid of the prompts?


beatrix-bitrot (approved):
Looks nice!

PR #6088, opened 2017-12-22
Allow to retrieve thumbnails from media shortcodes

ilpianista:
If the 'thumb' parameter is present, then we redirect to the thumbnail
url of the media.

Like in `https://functional.cafe/media/EZ52XzdwWk_fosPB0cE/?thumb`

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -6,7 +6,11 @@ class MediaController < ApplicationController
   before_action :verify_permitted_status
 
   def show
-    redirect_to media_attachment.file.url(:original)
+    if params[:thumb].present?
+      redirect_to media_attachment.file.url(:small)
+    else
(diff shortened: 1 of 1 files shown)


akihikodaki:
What kind of use case are you concerning?

ilpianista:
@akihikodaki for example you get toot's text, but you don't want to execute another query to get the attachment info. You already have all the informations to generate a thumbnail of that media and a link to the original size.

akihikodaki:
@ilpianista Internally, `media_attachment.file.url(:small)` queries for a database, so the number of queries do not change. In terms of APIs, as far as I know, all APIs returns statuses with media attachments, including URLs to preview cards. Then, when do you get only texts but URLs to preview cards?

Gargron:
I see no reason for this feature, sorry. Thank you for taking the time to contribute anyway.

PR #6032, opened 2017-12-15
Update Ruby to version 2.4.3

ykzts:
- [Ruby 2.4.3 Released - Ruby Programming Language](https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-4-3-released/)

> Ruby 2.4.3 has been released.
> 
> This release includes some bug fixes and a security fix.
> 
> - [CVE-2017-17405: Command injection vulnerability in Net::FTP](https://www.ruby-lang.org/en/news/2017/12/14/net-ftp-command-injection-cve-2017-17405/)
> 
> There are also som bug fixes. See [commit logs](https://github.com/ruby/ruby/compare/v2_4_2...v2_4_3) for more details.

- [x] rbenv
- [x] rvm
- [x] Travis
- [x] Docker (https://github.com/docker-library/official-images/pull/3810)

PR #6097, opened 2017-12-25
Add support Ruby v2.5.0

ykzts:
and drop Ruby v2.3.x in CI.

- [x] gems ([`required_ruby_version`](https://gist.github.com/ykzts/55c3a5b896709c81149fc42d961a020a))
    - [x] https://github.com/akihikodaki/cld3-ruby/pull/15
    - [x] https://github.com/plataformatec/devise/issues/4719
- [x] rbenv
- [x] rvm
- [x] Travis
- [x] Docker https://github.com/docker-library/ruby/pull/178

Close #6032


Gemfile
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-ruby '>= 2.3.0', '< 2.5.0'
+ruby '>= 2.4.0', '< 2.6.0'
```
Gargron: In my opinion: No reason to drop support for 2.3.0. Some people may still use it and there's no *critical* reason to upgrade the Ruby runtime

ykzts: okay

Gemfile
```diff
@@ -28,7 +28,7 @@ gem 'browser'
 gem 'charlock_holmes', '~> 0.7.5'
 gem 'iso-639'
 gem 'cld3', '~> 3.2.0'
-gem 'devise', '~> 4.2'
+gem 'devise', git: 'https://github.com/plataformatec/devise.git', ref: '1009096172f2cbc86bcd54d053c89a09be67fb9f'
```
ykzts: https://github.com/plataformatec/devise/issues/4719

It has not been released yet.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.4.2-alpine3.6
+FROM ruby:2.5.0-alpine3.7
```
ykzts: Since Ruby 2.5, official images using Alpine Linux 3.6 are not provided.

ykzts:
Devise is not a stable version, but there is no problem.

beatrix-bitrot:
thank you @ykzts !!!!!!!!

ykzts:
[Devise v4.4.0](https://github.com/plataformatec/devise/releases/tag/v4.4.0) has been released!

.travis.yml
```diff
@@ -37,20 +37,20 @@ addons:
     - yarn
 
 rvm:
-  - 2.3.4
```
akihikodaki: I guess you forgot to revert this.

Gemfile.lock
```diff
@@ -110,7 +110,7 @@ GEM
       activesupport
     charlock_holmes (0.7.5)
     chunky_png (1.3.8)
-    cld3 (3.2.1)
+    cld3 (3.2.2)
```
akihikodaki: Thank you for updating. ❤️

akihikodaki (dismissed):

.travis.yml
ykzts: https://github.com/tootsuite/mastodon/pull/6032#discussion_r157369178

akihikodaki: In my opinion, the configuration should match with `Gemfile`. That is the reason I added the line in the commit which changed the version requirement in `Gemfile`.
I also wonder if the review https://github.com/tootsuite/mastodon/pull/6097#discussion_r158653842 meant to keep the CI configuration or not, so cc: @Gargron 

(For your information, I relaxed the version requirement for RHEL SCL users and CentOS SCL users. Now they have Ruby 2.4 and I use it. I guess the other latest operating systems which are able to run Mastodon have Ruby 2.4.)

akihikodaki:
Arch Linux shipped Ruby 2.5. It is a good chance to merge this soon even though Arch Linux users are able to manage to install older versions using rbenv or whatever.
https://www.archlinux.org/packages/extra/x86_64/ruby/

Gargron (approved):

.travis.yml
Gargron: I'm OK with no longer testing against 2.3 but also keeping it compatible with the Gemfile. I feel like the 2.3 tests are redundant and just make CI take longer, but at the same time, if someone is still on 2.3, I don't want to *force* them to upgrade sooner than they would otherwise. It's very unlikely that we will start using Ruby features or syntax that is not already available in 2.3

ProgVal: @Gargron Thanks! (I use Ruby 2.3.3)

akihikodaki: If you use Ruby 2.3.x should be more careful because nobody tests it on the version.
You may have to tests and fix some if it is broken even if you don't have to rewrite `Gemfile`.

akihikodaki: If you use Ruby 2.3.x should be more careful because nobody tests it on the version.
You may have to tests and fix some if it is broken even if you don't have to rewrite `Gemfile`.

PR #6101, opened 2017-12-25
Make host_meta/webfinger replies cacheable (fixes #6100)

### LINKED ISSUE #6100: Allow caching of webfinger replies
As the admin of a mostly-single-user instance running on a low-end ARM server, I am interested in lowering Ruby's workload. One way to do would be to perform caching on the nginx reverse-proxy serving my domain.

One (relatively minor) thing that could be efficiently cached is the webfinger queries: indeed, they contain mostly immutable data.

However, it is currently difficult to achieve because Mastodon sets session cookies, in addition to setting the `Cache-Control` header to `max-age=0, private, must-revalidate`. This is most probably unnecessary, and not doing so would enable caching of webfinger replies.

* * * *

ClearlyClaire:
Drop common code for handling users and sessions as webfinger queries
are very basic, public APIs.

Also explicitly mark results as cacheable with “expires_in”.

### DIFF
--- a/app/controllers/well_known/host_meta_controller.rb
+++ b/app/controllers/well_known/host_meta_controller.rb
@@ -4,6 +4,8 @@ module WellKnown
   class HostMetaController < ActionController::Base
     include RoutingHelper
 
+    before_action { response.headers['Vary'] = 'Accept' }
+
     def show
       @webfinger_template = "#{webfinger_url}?resource={uri}"
 
--- a/app/controllers/well_known/host_meta_controller.rb
+++ b/app/controllers/well_known/host_meta_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module WellKnown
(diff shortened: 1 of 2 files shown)


ThibG:
Making use of it with nginx requires a few changes, such as enabling caching and enabling “proxy_buffering”. Also, I'm not sure “Vary: Accept-Encoding” set by Rails is correct, I think it should be “Vary: Accept” or “Vary: Accept, Accept-Encoding”?

An even more interesting change would be to cache public toots for a few minutes, as a common pattern is a lot of remote servers fetching it at the same time (e.g., when some content is boosted). This would require more work, though, as the concerned controllers also handle interaction with signed-in users.

akihikodaki (approved):

mdbraber:
FYI - with the update to 2.5.0 caching web replies can cause issues because the template changed from authorize_follow to authorize_interaction, see #8497

PR #6104, opened 2017-12-26
Fix XML oEmbed support discovery

akihikodaki:
This allows to use oEmbed, an open protocol, instead of Twitter's proprietary player card, when fetching preview cards of SoundCloud.
(fyi, SoundCloud also provides oEmbed in JSON format, but it incorrectly reports `text/json+oembed`. Just ignore it and use `text/xml+oembed`.)

### DIFF
--- a/app/lib/provider_discovery.rb
+++ b/app/lib/provider_discovery.rb
@@ -29,7 +29,7 @@ class ProviderDiscovery < OEmbed::ProviderDiscovery
       end
 
       if format.nil? || format == :xml
-        provider_endpoint ||= html.at_xpath('//link[@type="application/xml+oembed"]')&.attribute('href')&.value
+        provider_endpoint ||= html.at_xpath('//link[@type="text/xml+oembed"]')&.attribute('href')&.value
         format ||= :xml if provider_endpoint
       end
 
--- a/spec/fixtures/requests/oembed_json_xml.html
+++ b/spec/fixtures/requests/oembed_json_xml.html
@@ -1,8 +1,14 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <!--
+      oEmbed
+      https://oembed.com/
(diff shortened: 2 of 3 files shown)


app/lib/provider_discovery.rb
```diff
@@ -29,7 +29,7 @@ def discover_provider(url, **options)
       end
 
       if format.nil? || format == :xml
-        provider_endpoint ||= html.at_xpath('//link[@type="application/xml+oembed"]')&.attribute('href')&.value
+        provider_endpoint ||= html.at_xpath('//link[@type="text/xml+oembed"]')&.attribute('href')&.value
```
ykzts: Why do you delete support of the correct format `application/xml+oembed`? 

e.g. `//link[@type = "text/xml+oembed" or @type = "application/xml+oembed"]`

akihikodaki: The specification says nothing about `application/xml+oembed`. `text/xml+oembed` is, on the other hand, mandatory and it is assured that looking for `text/xml+oembed` is enough.

> The type attribute must contain either `application/json+oembed` for JSON responses, or `text/xml+oembed` for XML.

https://oembed.com/

ykzts: Hm. I see.

ykzts (approved):

Gargron (approved):

nightpool:
does this break twitch embeds?

nightpool:
(I'm wondering how the application/xml+oembed got in there if it's not to spec)

akihikodaki:
> does this break twitch/youtube embeds?

No, as far as I know. YouTube provides `application/json+oembed` and `text/xml+oembed`.
Twitch, does not provide a link for the oEmbed representation though it has, so Mastodon fallbacks to Twitter's player card. The behavior is not affected by this change.

PR #6106, opened 2017-12-26
Use const instead of let for constant

akihikodaki:


app/javascript/mastodon/features/getting_started/index.js
```diff
@@ -48,7 +48,7 @@ export default class GettingStarted extends ImmutablePureComponent {
   render () {
     const { intl, myAccount, columns, multiColumn } = this.props;
 
-    let navItems = [];
+    const navItems = [];
```
ykzts: https://github.com/tootsuite/mastodon/pull/6106/files#diff-322ec4ce8678beb1db1e7aee74ed7a7aR71

ykzts (approved):

PR #6107, opened 2017-12-26
Introduce information menu

akihikodaki:
In the old implementation, keyboard shortcuts and other information are located in different places.

That is confusing in case you were looking for keyboard shortcuts and found user guide first. You may expect keyboard shortcuts are listed in the user guide, but they are actually in a different place.

After applying this change, users should always open information menu if they need some helpful or additional information.

screenshot-2017-12-28 mastodon dev 1

UPDATE: updated the screenshot after applying #6109. Scroll bars are no longer shown and the UI look better.


koyuawsmbrtn:
What about adding a tutorial button to re-visit the tutorial dialog in case someone missed it? Or is it already happening on the user-guide-button?

akihikodaki:
@koyuawsmbrtn That may be nice, but it is out of the scope of this pull request. You may open an issue for your suggestion.
In my opinion, the user guide is better for reviewing and the content of the tutorial should be included instead of replayed. If you have different opinions, open an issue and discuss about that in the issue.

akihikodaki:
Information menu button is always shown (as links to FAQ, the user guide, the list of apps were) because newcomers may miss the button if it is hidden in scrollable navigation menu.
screenshot-2017-12-28 mastodon dev

chriswmartin:
There is also this as a similar alternative UI https://github.com/glitch-soc/mastodon/pull/266.

akihikodaki:
The functionality this pull request implements and one glitch-soc has are same, but I have a different motivation to implement this feature; to keep the navigator(s) for the information items **always** shown for newcomers who have little knowledge about Mastodon's UI.

It is possible to add *"information subheading" (like "navigation subheading and "settings subheading") to Getting started column*, instead of *creating a new column*.
In the case, the navigation area of the getting started column will get taller and need more scrolling. I guess that's what glitch-soc developers wanted to avoid.
However, information subsection has another problem; a newcomer may have difficulty to find user guide and other helpful information. Many Mastodon users, including those who use glitch-soc instance know they can find items hidden in the navigation column, so the overflowing navigation area introduces little inconvenience. In contrast, __a newcomer may have greater difficulty__ if the information subsection is hidden, because __he does not know even such a subsection is present.__

In terms of that, this change is almost entirely for newcomers.

Gargron:
Solved by redesign #7676

PR #6109, opened 2017-12-26
Move the mastodon on Getting Started column to drawer column

akihikodaki:
Getting Started column obtained many links, and it became much taller. Because of its height, Getting Started column required long scrolling on devices with small screen, such as 4 inch phones and 10 inch laptops.

This change moves the mastodon which took large space on the column to drawer column. The drawer column has only the compose form and has more space.

Before (only his head is shown!):
screenshot-2017-12-26 mastodon dev

After (he is shown perfectly!):
screenshot-2017-12-26 mastodon dev 1


dolfinsbizou:
Wow great idea!

beatrix-bitrot (approved):
nice!

PR #6114, opened 2017-12-27
keep the same filters and page when doing custom emojo stuff (fixes #6112)

### LINKED ISSUE #6112: When clicking "copy" on the emojo list it doesn't always stay on the same page
Say I'm on `/admin/custom_emojis?by_domain=cybre.space&page=4&shortcode=`, so I've used the search feature. Then I'll click on the "Copy" link next to an emojo. The only information that link retains is the page number, so I'll get a completely different list of emojos afterwards. This should be an easy fix (add `by_domain` and `shortcode` fields to the "Copy" link).

* * * *

beatrix-bitrot:


app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -32,23 +33,26 @@ def update
 
       if @custom_emoji.update(resource_params)
         log_action :update, @custom_emoji
-        redirect_to admin_custom_emojis_path, notice: I18n.t('admin.custom_emojis.updated_msg')
+        flash[:notice] = I18n.t('admin.custom_emojis.updated_msg')
       else
-        redirect_to admin_custom_emojis_path, notice: I18n.t('admin.custom_emojis.update_failed_msg')
+        flash[:alert] =  I18n.t('admin.custom_emojis.update_failed_msg')
       end
+      redirect_to admin_custom_emojis_path(page: params[:page], **@filter_params)
```
nightpool: rule of three says `admin_custom_emojis_path(page: params[:page], **@filter_params)` can probably be a helper? then you can just pass it a "notice" keyword and don't have to refactor to use `flash`

nightpool (changes requested):

app/controllers/admin/custom_emojis_controller.rb
beatrix-bitrot: -_-;

i suppose i could have the helper pass whatever is given (because sometimes we should alert rather than notice) 

what do you reckon such a helper should be called?

the reason i didn't make one is that it didn't feel like it would be a very meaningful method and i like to avoid increasing indirection if i can

nightpool: `current_filter_path`?

nightpool:
http://guides.rubyonrails.org/action_controller_overview.html#default-url-options may be relevant

(although perhaps it will break/be added to non-relevant urls in which case the helper is more appropriate except it doesn't work in views..... :( )

beatrix-bitrot:
i will try it and see how it goes. was out all today

beatrix-bitrot:
the default URL options way does work with the important exception that it completely breaks the 'reset' button. i will see if i can get around that by explicitly nulling out the params for that one

beatrix-bitrot:
oop spoke too soon, it also affects all of the menu items in the sidebar 😅 

@nightpool can we let this one go as is for now?

nightpool (approved):

nightpool:
yeah, sure. let's merge it

PR #6115, opened 2017-12-27
Allow HTTP caching of json view of public statuses

ClearlyClaire:
Whenever a user replies or boosts a toot, it tends to cause multiple instances to access it roughly at the same time. This may lead to hundreds of requests that have to be handled by puma.

By allowing the front-end web server to cache public toots for a few minutes, the number of requests handled by puma can be greatly reduced.

This commit is a first step toward that goal, changing “Cache-control” headers as needed for json rendering of public toots (the HTML renderings are excluded as they may also contain private ancestors/replies).

Unfortunately, “Set-Cookie” is still set to maintain user sessions. I don't think this is needed for json rendering of public toots (since that's only going to be used for federation, anyway), but I'm not sure how to properly change it.

### DIFF
--- a/app/controllers/statuses_controller.rb
+++ b/app/controllers/statuses_controller.rb
@@ -28,7 +28,10 @@ class StatusesController < ApplicationController
                content_type: 'application/activity+json'
 
         # Allow HTTP caching for 3 minutes if the status is public
-        expires_in(3.minutes, public: true) unless @stream_entry.hidden?
+        unless @stream_entry.hidden?
+          request.session_options[:skip] = true
+          expires_in(3.minutes, public: true)
+        end
       end
     end
   end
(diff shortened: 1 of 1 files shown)


Gargron:
Afaik nothing changes in the ActivityPub JSON of a toot once it's published so I think the JSON could be cached in Rails. That way the only DB call would be `select id, updated_at from statuses` This is a good point. Being nginx-cacheable would be some value on top.

ThibG:
I don't know anything about caching in Rails, so I cannot really comment on that, although I guess it would make sense, and would allow for much longer caching on the Rails side of things, since invalidation would be possible.

However, considering how slow Rails is and how few workers I can run, I'm really in favor of having caching on nginx's side.

I think I have managed to properly disable session cookies for ActivityPub json rendering of public toots, thus allowing caching in nginx.

Gargron (approved):

PR #6125, opened 2017-12-28
Add more instance stats APIs

Gargron:
- GET /api/v1/instance/peers  
  Returns list of all domain names known to this instance
- GET /api/v1/instance/activity
  Returns counts of new local statuses, new users, and active users in weekly buckets

Both APIs can be turned off individually from admin settings. Private and direct statuses do not contribute to weekly status counts.

### DIFF
--- a/app/controllers/api/v1/instances/activity_controller.rb
+++ b/app/controllers/api/v1/instances/activity_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Api::V1::Instances::ActivityController < Api::BaseController
+  before_action :require_enabled_api!
+
   respond_to :json
 
   def show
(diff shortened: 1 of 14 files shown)


app/controllers/api/v1/instances/activity_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def activity
+    weeks = []
+
+    12.times do |i|
+      week    = i.weeks.ago.to_date
+      week_id = week.cweek
+
+      weeks << {
+        week: week.to_time.to_i.to_s,
```
aschmitz: This should probably either be `week_id` or a timestamp for the start of the week, not just a day in that week.

aschmitz (commented):
This looks fine to me modulo the date issue in ActivityController. It'd be nice if `Rails.cache` would easily return the TTL on cached data so we could set an actually-accurate expires header, but that doesn't seem too practical right now (and it's not super important anyway).

app/controllers/api/v1/instances/activity_controller.rb
Gargron: It is the timestamp of the start of the week

aschmitz:
:+1: LGTM.

ykzts (approved):

annando:
Which version will have this API endpoint? And is it accessible without login?

Gargron:
@annando 2.1.2 and yes like the /api/v1/instance endpoint it requires no authentication

PR #6134, opened 2017-12-30
Don't normalize URLs in toots

ClearlyClaire:
URL normalization is ill-defined and may cause certain links to break.

As normalization serves no purpose in rendering links in toots, this pull request avoids normalizing them.

Some more background is available in this thread: https://mastodon.gougere.fr/@bortzmeyer/99264235735779201

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -171,10 +171,10 @@ class Formatter
   end
 
   def link_to_url(entity)
-    normalized_url = Addressable::URI.parse(entity[:url]).normalize
-    html_attrs     = { target: '_blank', rel: 'nofollow noopener' }
+    url        = Addressable::URI.parse(entity[:url])
+    html_attrs = { target: '_blank', rel: 'nofollow noopener' }
 
-    Twitter::Autolink.send(:link_to_text, entity, link_html(entity[:url]), normalized_url, html_attrs)
+    Twitter::Autolink.send(:link_to_text, entity, link_html(entity[:url]), url, html_attrs)
   rescue Addressable::URI::InvalidURIError, IDN::Idna::IdnaError
     encode(entity[:url])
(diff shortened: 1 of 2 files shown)


ThibG:
Note that there are numerous other places where Mastodon makes use of `Addressable::URI`'s normalization method. This is probably ill-advised for the same reasons, but I didn't went through it to actually check.

bortzmeyer:
May be related to #4837

ThibG:
The tests don't pass because the specs expect the normalization to actually take place, but that can be amended if we agree to drop normalization altogether.

Unlike ActivityPub URIs, user-provided URLs are not IRIs so they should not be normalized.

unarist (commented):
Looks good if we can assume status body is HTML5 which allows using an IRI as an URL.

The only thing we should avoid is NFKC normalization in Unicode, so we don't need to avoid Punycode conversion and percent encoding, but I think it's OK too.

ThibG:
I don't think we can assume anything about user-provided URLs, especially not that they are IRI, so we definitely shouldn't do NFKC normalization. I guess we may do other kinds of normalization steps—although I don't see any reason why we should,—but it doesn't seem to be possible using `Addressable:URI` without also doing NFKC normalization.

Gargron:
@ThibG Is the problem localized to paths being normalized? Can we still normalize hosts? Because I would prefer to see some kind of xn-- when hovering the link instead of those unicode-lookalikes

ThibG:
@Gargron as far as I can tell, yes, we could still normalize the hosts. I don't really see the benefit of doing the normalization, though. Also, it appears Firefox at leasts displays the punycode-encoded domain when hovering regardless of whether the link had a punycode-encoded domain name or not.

Gargron (approved):

PR #6138, opened 2017-12-31
360 Panorama Viewer (front-end only)

### LINKED ISSUE #5867: 360/panorama photo viewer
Equirectangular 360 photos seem to be rendered as standard (large) images. There are several examples of content-sensitive handling for this type of image on the web. Facebook, for example, creates a continuously 2d-scrollable view-port onto the wrapped image, and allows an initial viewport to be specified on upload. Can we replicate such an interface?

- https://pannellum.org/ is one example of a WebGL-based viewer seemingly designed for webapp integration.
- Given local caching of images, maybe we can do an entirely local decision for when to swap in the image viewer, based on exif data?
- Is there an obvious way to add initial-viewport metadata to a toot with 360 media? Appending polar coordinates to the URL?

* * * *

c-probably-archived20260718:
This adds basic front-end support for images with [spherical XMP metadata](https://developers.google.com/streetview/spherical-metadata) and partially resolves https://github.com/tootsuite/mastodon/issues/5867, but because the back-end strips all image metadata when uploading a media attachment (and because I don't know Ruby so I have no idea how to change that), this does not actually work without replacing the image files in the database.

demo

Some code adapted from [exif-js](https://github.com/exif-js/exif-js) and from [marzipano](https://github.com/google/marzipano).

The user interface is very basic and supports mouse-dragging, scroll-zooming, and pinch-zooming in Safari.

[Potential] issues:
- No back-end support
- Images are downsized a lot when uploading. Because the images are blown up when projected onto a sphere, this is much more noticeable
- accessibility?


app/javascript/mastodon/features/ui/components/image_loader.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Motion from '../util/optional_motion';
+import spring from 'react-motion/lib/spring';
+import PanoramaViewer, { getPanoramaData } from './panorama_viewer';
```
sorin-davidoi: Haven't looked at the size, but could this be lazy-loaded? Only a small fraction of images will be panoramas, so it does not really make sense to always load it (see https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/features/ui/util/async-components.js).

cpsdqs: The part that checks for panoramas can't be lazy loaded, but the WebGL renderer could. Not sure if it'd be worth it though, the webpack chunks here are at least 30kB and the uncompressed renderer is just 9kB

cpsdqs: oops I forgot about gl-matrix, which is huge. I can't find anything where the webpack chunks are defined though?

sorin-davidoi: They are defined [here](https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/features/ui/util/async-components.js) - for an example how to use see [the emoji picker](https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js#L4).

Gargron:
This is really cool. But yeah I guess we need to figure out how to use ImageMagick's `-strip` without stripping out that spherical metadata.

cpsdqs:
Maybe the spherical metadata could be extracted before stripping and inserted back after?

akihikodaki:
XMP can contain any kind of data so it should be audited, but ImageMagick is not capable to do that.

> Maybe the spherical metadata could be extracted before stripping and inserted back after?

That is the only possible solution, I think.

Gargron:
>That is the only possible solution, I think.

Do we even have the tools to do that, or would it require adding another system-level dependency. Also, paperclip makes this very complicated :frowning:

akihikodaki:
> Do we even have the tools to do that, or would it require adding another system-level dependency.

Unfortunately there is no well-maintained Ruby implementation, so it would require system-level dependency. Exiv2, a command-line metadata editor, is available on Alpine Linux, Arch Linux, CentOS, Debian, Fedora, Homebrew, and RHEL.

> Also, paperclip makes this very complicated 😦

Not that much. Implementing a new processor by extending `Paperclip::Thumbnail` would be a good solution. See http://www.rubydoc.info/gems/paperclip/Paperclip/Thumbnail for the documentation.

cpsdqs:
can someone maybe add a “help wanted” tag?

akihikodaki:
#6317 deals with the help-wanted concern.

cpsdqs:
I just updated this, and commit [e7932c5](https://github.com/tootsuite/mastodon/pull/8292) is not part of this PR; I don’t know why that ended up there

neuhaus:
Looks like 99% of the work is already done here.. could we get this re-opened and merged?

Sasha-Sorokin:
This probably needs to be rebased @neuhaus. It's not very useful to reopen unless it is maintained, the PR may be recreated after all.

PR #6142, opened 2018-01-01
Set minimum height for mastodon on drawer

akihikodaki:
Someone exploited the lack of the minium height to remove the mastodon. That is not good, so here is a fix.

https://glitch.social/@kibi/99255525411055620

### DIFF
--- a/app/javascript/mastodon/features/compose/index.js
+++ b/app/javascript/mastodon/features/compose/index.js
@@ -12,6 +12,7 @@ import Motion from '../ui/util/optional_motion';
 import spring from 'react-motion/lib/spring';
 import SearchResultsContainer from './containers/search_results_container';
 import { changeComposing } from '../../actions/compose';
+import elephantUIPlane from '../../../images/elephant_ui_plane.svg';
 
 const messages = defineMessages({
   start: { id: 'getting_started.heading', defaultMessage: 'Getting started' },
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron:
I found that gif very funny... Are you sure it needs to be fixed? I mean, if the text box grows that large... Text box definitely has higher priority than decorative elephant.

akihikodaki:
@Gargron
> Text box definitely has higher priority than decorative elephant.

I agree. The elephant should not bother the text box. However, the scroll bar would be necessary soon if the text box is growing. I do not think additional 40px can bother a user much.
If it forces to show the scroll bar even if the text box is not grown, that would be too bothering; you are sure you do not need scrolling even if later. But that does not happen because any display which is capable to show Mastodon should be tall enough to contain the compose form and the elephant

beatrix-bitrot:
i like the way the elephant shrinks :3c

akihikodaki:
> i like the way the elephant shrinks :3c

Yes, it shrinks, but it just does not go away after applying this change.

Gargron:
> `min-height: 47px;`

Okay, it still shrinks. I misunderstood this change at first. If you adjust it for current master I will merge.

akihikodaki:
f0ba0bf6b64f9609305a39e9f671e95f8e19aac0: rebased to ac1093256c500f55a6578836c3364d4a8a67ee58.

nazco:
We at chaos.social observed a problem with the changes made by this merge request. We use a custom version of the mastodon image with the ccc's fairydust. In the past you could easily overwrite the css that was loading the mastodon image as background (and currently is) within a custom theme. But by now, while loading the "original" mastodon image with a hard coded image tag this is only possible by setting a `display: none;` for the image tag. This isn't a big issue but it needs additional code to disable the displaying of the image tage and the image is loaded regardless of this change. Maybe we could archive the same changes while using the old css background variant?

akihikodaki:
@nazco You may be interested in #6461.

My opinion is that it is fine to edit `app/javascript/mastodon/features/compose/index.js` to change the path or to replace `app/javascript/images/elephant_ui_plane.svg`. It is not worse than overriding CSS class, or, in other words, Overriding CSS class is as unstable as editing JavaScript.

PR #6143, opened 2018-01-01
Add a button to compose on another instance for intents

akihikodaki:
Though we have neat `web+mastodon` protocol for sharing, some Web sites use a link for intents on particular instances. That would work better for those who use such instances because it does not require registering the protocol handler, but that would not work at all for others.
This commit adds a button to "compose on another instance button" to fix the behavior which is an obstacle for federation.
screenshot-2018-1-1 mastodon dev

### DIFF
--- a/app/controllers/intents_controller.rb
+++ b/app/controllers/intents_controller.rb
@@ -9,7 +9,7 @@ class IntentsController < ApplicationController
       when 'follow'
         return redirect_to authorize_follow_path(acct: uri.query_values['uri'].gsub(/\Aacct:/, ''))
       when 'share'
-        return redirect_to share_path(text: uri.query_values['text'])
+        return redirect_to share_path(text: uri.query_values['text'], web: true)
       end
     end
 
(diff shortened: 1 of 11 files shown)


unarist:
Sounds good, though I'm worried if web+mastodon behavior confuses users.

What if they haven't been registered web+mastodon protocol on other instances (or any instances)? Also, Chrome doesn't have "Which site you want to open with?" dialog and users have to choose instance from context menu if they want to use other than default one. I'm not sure how many users know this...

Actually, this is not specific issue for this usage and I don't mean using it is bad idea.

This may be still useful even if called from web+mastodon, but how about to show this button only if the user visits share page directly?

akihikodaki:
> This may be still useful even if called from web+mastodon, but how about to show this button only if the user visits share page directly?

I am not sure what do you mean by "the user visits share page directly". Are you concerning the case that the user navigated to the share page without signing in?

unarist:
I mean when navigated, by direct link to the particular instance instead of web+mastodon://.

akihikodaki:
20de4596f62783bf241fdd042327ae043a3772b2 changes according to https://github.com/tootsuite/mastodon/pull/6143#issuecomment-366502825; it will not show the button when `web+mastodon` is already used.

PR #6146, opened 2018-01-01
Adding Serbian latin translations

stalker314314:
Serbian latin (sr-Latn) is generated automatically from Serbian (sr) translation. Also changed some wording in original (Serbian) translation.


ykzts:
I don't know the advantage of adding Serbian latin. Is not Serbian alone enough?

stalker314314:
I am not going to go into constitutional/political/regional/cultural reasons for having both scripts, you can read on [wiki more](https://en.wikipedia.org/wiki/Serbian_language#Writing_system), I guess. It is also defined in [RFC5646](https://tools.ietf.org/html/rfc5646) from [here](https://github.com/tootsuite/documentation/blob/master/Contributing-to-Mastodon/Translating.md#declaring-the-language) and it is in other [projects](https://l10n.gnome.org/teams/sr/). Beauty of it is that Cyrillic version can be automatically converted to latin (I can share script, if needed), and that's why all translators favor Cyrillic to be translated first. I think there is no reason to not have both, but if you don't want to - it's your call.

app/helpers/settings_helper.rb
```diff
@@ -30,6 +30,7 @@ module SettingsHelper
     ru: 'Русский',
     sk: 'Slovensky',
     sr: 'Српски',
+    'sr-Latn': 'Srpski (latinica)'
```
ykzts: add comma.

ykzts (changes requested):

app/helpers/settings_helper.rb
stalker314314: Oops, sorry, fixed (same commit, squashed)

ykzts (approved):

stalker314314:
Thanks (again!)

PR #6157, opened 2018-01-02
Allow boosting own private toots

ClearlyClaire:
Boosting your own private toots is sometimes requested, and I don't see any reason why that shouldn't be possible.
While I honestly don't know how important such a feature is, it did not seem too hard to implement, so here it is.

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -17,6 +17,8 @@ const messages = defineMessages({
   more: { id: 'status.more', defaultMessage: 'More' },
   replyAll: { id: 'status.replyAll', defaultMessage: 'Reply to thread' },
   reblog: { id: 'status.reblog', defaultMessage: 'Boost' },
+  reblog_private: { id: 'status.reblog_private', defaultMessage: 'Boost to original audience' },
+  cancel_reblog_private: { id: 'status.cancel_reblog_private', defaultMessage: 'Unboost' },
   cannot_reblog: { id: 'status.cannot_reblog', defaultMessage: 'This post cannot be boosted' },
   favourite: { id: 'status.favourite', defaultMessage: 'Favourite' },
   open: { id: 'status.open', defaultMessage: 'Expand this status' },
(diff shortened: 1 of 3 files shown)


ThibG:
I have successfully tested those changes, but the UI needs some more work: while it works on a desktop browser, it relies on hovering the lock icon to show the usual “boost” buttons. This design has a few very important limitations:
- It requires users to hover on the lock icon to discover the feature
- The whole hovering thing does not make sense on mobile

akihikodaki:
> I have successfully tested those changes, but the UI needs some more work: while it works on a desktop browser, it relies on hovering the lock icon to show the usual “boost” button. This design has a few very important limitations:

I think the visibility should be shown in a different place (bottom right, for example,) and, if such a UI change is planned, it should be done in the same release.

Sylvhem:
dev.glitch.social already separated the privacy indicator from the boost button (see glitch-soc/mastodon#55).
1. There is a privacy indicator in the top right corner of every toot.
2. When you can't boost a toot, the boost button is simply disabled.

public toot
private toot

May I suggest that we adopt a similar approach on this?

ThibG:
I guess #6230 means this PR is rejected?

jpdeplaix:
Awesome ! Any idea why Travis is failing here ? Any idea whether this PR can be accepted ?

ThibG:
Travis wasn't failing because of the code, but because of something wrong with Travis/ubuntu. I will rebase the patch and trigger a new CI pass.
No idea if this PR can be accepted, but it still needs a bit of UI work, as described in earlier posts.

Gargron:
What if boosting a private status was possible from the dropdown menu instead of the icon button that also works as privacy indicator? That way the change wouldn't mess with people's muscle memory. You could even add wording like "Boost to original audience" or something to better convey what's happening.

Sylvhem:
@Gargron I'm not really a big fan of the idea I must say. It would be really weird to have two ways of boosting toots. Also, if the only way to boost a private toot is in the dropbox menu, I'm afraid a lot of people will miss this feature: if the boost button is replaced by a padlock, they will just assume you can boost it.
Besides, the proposed solution will allow to know a toot privacy after posting it, which is especially useful if you want to differentiate between public and unlisted messages.
In my opinion, since they have different meanings, we should really make the privacy icon and the boost icon two different things.

Gargron (approved):

kit-ty-kate:
Is there any plan of merging glitchsoc's way of displaying visibility indicator into vanilla for a later version ? Otherwise I'm happy to see that at least allowed in some way. Thanks for merging this.

PR #6162, opened 2018-01-03
[!] Sanitize incoming classlist properly

puckipedia:
Previously (when the spinny text issue was fixed) the classlist was done by splitting by spaces. This is not enough. Any HTML whitespace (space, tab, newline, carriage return, and formfeed) are valid class separators. Together with some new features like custom emoji, you can definitely make some really impressive fake objects. Or just spinny text again. \*shrug\*

Alternatively, while testing out my code, I figured that the `return true` inside of the block actually returns from the function that sanitizes the class, effectively not sanitizing any class after the first one. Well done.

cc @nightpool

### DIFF
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -6,7 +6,7 @@ class Sanitize
 
     CLASS_WHITELIST_TRANSFORMER = lambda do |env|
       node = env[:node]
-      class_list = node['class']&.split(' ')
+      class_list = node['class']&.split(/[\t\n\f\r ]/)
 
       return unless class_list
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

app/lib/sanitize_config.rb
```diff
@@ -6,14 +6,14 @@ module Config
 
     CLASS_WHITELIST_TRANSFORMER = lambda do |env|
       node = env[:node]
-      class_list = node['class']&.split(' ')
+      class_list = node['class']&.split(/[\t\n\f\r ]/)
```
valerauko: wouldn't just `\s` work? according to docs, "/\s/ - A whitespace character: /[ \t\r\n\f\v]/"

PR #6171, opened 2018-01-03
Cache JSON of immutable ActivityPub representations

Gargron:
cc @ThibG

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -2,7 +2,8 @@
 
 class AccountsController < ApplicationController
   include AccountControllerConcern
-  include SignatureVerification
+
+  before_action :set_cache_headers
 
   def show
     respond_to do |format|
@@ -26,10 +27,11 @@ class AccountsController < ApplicationController
       end
 
       format.json do
-        render json: @account,
-               serializer: ActivityPub::ActorSerializer,
-               adapter: ActivityPub::Adapter,
-               content_type: 'application/activity+json'
+        skip_session!
+
+        render_cached_json(['activitypub', 'actor', @account.cache_key], content_type: 'application/activity+json') do
+          ActiveModelSerializers::SerializableResource.new(@account, serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter)
+        end
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -2,7 +2,8 @@
 
 class AccountsController < ApplicationController
   include AccountControllerConcern
-  include SignatureVerification
```
Gargron: (This wasn't used)

app/controllers/accounts_controller.rb
```diff
@@ -26,10 +27,11 @@ def show
...
       format.json do
-        render json: @account,
-               serializer: ActivityPub::ActorSerializer,
-               adapter: ActivityPub::Adapter,
-               content_type: 'application/activity+json'
+        skip_session!
+
+        render_cached_json(['activitypub', 'actor', @account.cache_key], content_type: 'application/activity+json') do
+          ActiveModelSerializers::SerializableResource.new(@account, serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter)
+        end
```
ThibG: The rest seems fine, but isn't this going to cause Mastodon to serve stale account info (display name, avatar, header image, etc.)?

ClearlyClaire (changes requested):

app/controllers/accounts_controller.rb
Gargron: @ThibG: No. `#cache_key` method returns a combination of model name, ID and updated_at

ThibG: Alright, it's fine with me, then!

ClearlyClaire (approved):

app/controllers/application_controller.rb
```diff
@@ -123,11 +123,23 @@ def respond_with_error(code)
...
   def render_cached_json(cache_key, **options)
+    options[:expires_in] ||= 3.minutes
+    cache_key              = cache_key.join(':') if cache_key.is_a?(Enumerable)
+    content_type           = options.delete(:content_type) || 'application/json'
+
     data = Rails.cache.fetch(cache_key, { raw: true }.merge(options)) do
       yield.to_json
     end
 
     expires_in options[:expires_in], public: true
```
ThibG: Sorry I didn't catch that earlier, but you're unconditionally marking it as public here, which may cause problems.
Indeed, I believe it is possible to get private toots rendered with “public: true” using this code. This shouldn't immediately cause issues as “Set-Cookies” headers would be set, preventing the response to be cached, but it's still worrying.

PR #6172, opened 2018-01-03
Send one Delete of Actor in ActivityPub when account is suspended

### LINKED ISSUE #5420: Account deletion causes excessive Sidekiq tasks
When a user is suspended or deletes their account, the sidekiq `push` queue is filled with Tombstones activity. If it was an account with many many toots, it can take a while before the instance can push other activities in the fediverse, effectively locking it out of the federation until every tombstone is processed.

I suggest the creation of a new sidekiq queue of lower priority than push to deal with the tombstones.

(And I'm not saying that because WitchesTown has currently 2 million jobs (1% of the total instance activity) waiting in the queue 😇

Gargron:
Fix #5420

The handler for this type of payload was introduced in e482595a5dfd69efacffeddbd48f48c5b92deaea which is >=1.6.0

Please review carefully, if I've made any implementation mistakes then self-deleted/suspended accounts will leave all their content intact on remote servers.

### DIFF
--- a/app/services/batched_remove_status_service.rb
+++ b/app/services/batched_remove_status_service.rb
@@ -17,9 +17,7 @@ class BatchedRemoveStatusService < BaseService
 
     @stream_entry_batches  = []
     @salmon_batches        = []
-    @activity_json_batches = []
     @json_payloads         = statuses.map { |s| [s.id, Oj.dump(event: :delete, payload: s.id.to_s)] }.to_h
-    @activity_json         = {}
     @activity_xml          = {}
 
     # Ensure that rendered XML reflects destroyed state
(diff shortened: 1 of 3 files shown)


ykzts (approved):

app/services/batched_remove_status_service.rb
```diff
@@ -57,22 +51,6 @@ def batch_stream_entries(account, statuses)
...
 
-  def batch_activity_json(account, statuses)
-    account.followers.inboxes.each do |inbox_url|
-      statuses.each do |status|
-        @activity_json_batches << [build_json(status), account.id, inbox_url]
-      end
-    end
-
-    statuses.each do |status|
-      other_recipients = (status.mentions + status.reblogs).map(&:account).reject(&:local?).select(&:activitypub?).uniq(&:id)
```
unarist: `other_recipients` is needed to handle non-followers, but RawDistributionWorker doesn't handle those recipients...?

unarist: Also we need to handle reblogs even for Delete -> Actor from remote instance. This has been handled in Delete -> Note before

https://github.com/tootsuite/mastodon/blob/7571c37c991c7bcaf83c524105f4a3b64b19670c/app/lib/activitypub/activity/delete.rb#L31-L40

but now we only relays Delete -> Actor.

PR #6179, opened 2018-01-04
Fix accessibility of column headers

ndarilek:
As a screen reader user new to Mastodon, I encountered the following issues with the column headers as designed:
 * Jumping between them was difficult. FOr instance, passing my home timeline to reach notification settings was difficult to impossible, especially considering infinite scrolling.
 * There doesn't appear to be any means for triggering the control via the keyboard. the `titleClick` handler only responds to mouse clicks.
 * I didn't even realize there was a Settings toggle until I made this change.

Thanks for using ARIA in your designs. It's a huge help. But adding a `button` role doesn't add keyboard handling and other button behavior. Also, because the role was on the heading container, it obscured the controls within the container itself. This fix resolve that. It also exposes the headings as headings rather than buttons, enabling skipping columns by using screen readers' heading navigation commands.

Since I myself am blind, if this fix requires additional visual styling, I'd like help applying that so it can be merged. I'd consider it an essential accessibility fix for my and other blind users' existence on the platform. Thanks!

### DIFF
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -23,7 +23,6 @@ export default class ColumnHeader extends React.PureComponent {
     icon: PropTypes.string.isRequired,
     active: PropTypes.bool,
     multiColumn: PropTypes.bool,
-    focusable: PropTypes.bool,
     showBackButton: PropTypes.bool,
     children: PropTypes.node,
     pinned: PropTypes.bool,
@@ -31,10 +30,6 @@ export default class ColumnHeader extends React.PureComponent {
     onMove: PropTypes.func,
     onClick: PropTypes.func,
   };
-
-  static defaultProps = {
-    focusable: true,
-  }
 
   state = {
     collapsed: true,
@@ -68,7 +63,7 @@ export default class ColumnHeader extends React.PureComponent {
   }
 
   render () {
-    const { title, icon, active, children, pinned, onPin, multiColumn, focusable, showBackButton, intl: { formatMessage } } = this.props;
+    const { title, icon, active, children, pinned, onPin, multiColumn, showBackButton, intl: { formatMessage } } = this.props;
     const { collapsed, animating } = this.state;
 
     const wrapperClassName = classNames('column-header__wrapper', {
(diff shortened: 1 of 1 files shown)


beatrix-bitrot:
I will test. Not really familiar enough with the code to offer a good review based on reading it alone.

marrus-sh:
I opened a new pull request which fixes the visual styling for this PR at #6199 since I don't have write access on tootsuite. It's just this commit plus one of my own.

ndarilek:
Thanks a bunch!

Gargron:
Wouldn't it be simpler to add keyboard handlers on top of onClick? I can think of a bunch of places throughout the UI where we use role='button' on a non-button element for a11y reasons, I don't want to introduce inconsistencies.

sorin-davidoi:
It's not just keyboard handlers, it's also about adding `tabIndex="0"` and so on. ARIA says that when a native tag exists for the function you want to implement, use it and don't try to copy it's behavior to other elements.

The fact that we are using `role="button"` is not ideal and we should move to using native buttons - I've addressed some of these in my series of PRs from July (e.g. https://github.com/tootsuite/mastodon/pull/4457/files#diff-a5845c6ed5c6da5d8e43329f72fa9e35L20), but seems like I've missed some or didn't know better then.

ndarilek:
Right, this, and generally when folks do fancy stuff with ARIA to 
override default behaviors, they shoot themselves in the foot. For 
instance, this custom widget grouped both the title button and the 
settings collapsable button under the same control with the button role. 
My screen reader presented both to me as a single button, with no 
indication that there were even settings attached until I fixed the 
issue and discovered the new control.

Plus, I think button keyboard handling may be platform-specific. I.e. on 
some platforms, both Space and Enter cause clicks. On some, only Enter 
dispatches a click. It isn't enough to implement keyboard handling. You 
also have to implement it based on the platform's conventions.

You may have inconsistencies as PRs like this land, but in the long run, 
`role` should be a tool of last resort, not commonly used in lieu of 
standard elements. Better to have inconsistencies but converge on a 
standard, than to have a bunch of non-standard widgets implementing 
platform-specific keyboard handling, which is going to make Mastodon 
accessibility a significantly larger challenge.

Gargron:
This was merged via #6199

PR #6180, opened 2018-01-04
Fix enforce HTTPS in production.

yukimochi:
When .env.production don't has 'LOCAL_HTTPS=true', cookies don't has secure attribute and so on ...

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -46,7 +46,7 @@ RSpec.configure do |config|
   config.include ActiveSupport::Testing::TimeHelpers
 
   config.before :each, type: :feature do
-    https = ENV['LOCAL_HTTPS'] == 'true'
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron (approved):

spec/rails_helper.rb
```diff
@@ -46,7 +46,7 @@ def sign_in(resource, _deprecated = nil, scope: nil)
   config.include ActiveSupport::Testing::TimeHelpers
 
   config.before :each, type: :feature do
-    https = ENV['LOCAL_HTTPS'] == 'true'
+    https = Rails.env.production? || ENV['LOCAL_HTTPS'] == 'true'
```
Gargron: I'm pretty sure this is always the "test" environment though, in spec/rails_helper.rb 🤔

PR #6182, opened 2018-01-04
fix overlooked LOCAL_HTTPS references

nightpool:
This PR fixes two overlooked references to LOCAL_HTTPS. In order, this fixes:

 - the last-resort fallback to forcing HTTPS if the web server (using a non-default configuration) manages to forward a HTTP request to the 

   note that this may break users (such as apache users) with misconfigured servers that don't forward x-forwarded-proto. This needs a mention in the release notes otherwise it will cause an infinite loop.

 - mastodon session cookies will now be marked 'https only' again, to prevent against MITM downgrade attacks.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,7 +30,7 @@ class ApplicationController < ActionController::Base
   private
 
   def https_enabled?
-    Rails.env.production? || ENV['LOCAL_HTTPS'] == 'true'
+    Rails.env.production?
   end
 
   def store_current_location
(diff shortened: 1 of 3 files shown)


Gargron (approved):

app/controllers/application_controller.rb
```diff
@@ -30,7 +30,7 @@ def raise_not_found
   private
 
   def https_enabled?
-    Rails.env.production? && ENV['LOCAL_HTTPS'] == 'true'
+    Rails.env.production?
```
ykzts: `Rails.env.production? || ENV['LOCAL_HTTPS'] == 'true'`

nightpool: breaks tests

PR #6183, opened 2018-01-04
Use disable_ddl_transaction! to prevent warnings on migration

unarist:
Migration is wrapped by transaction, so manual `commit_db_transaction` without transaction restarting causes "there is no transaction in progress" warnings. We should use `disable_ddl_transaction!` instead, if we can omit transaction completely.

cc @takayamaki


Gargron (approved):

takayamaki:
👀

PR #6189, opened 2018-01-05
Un-animate 500 error page.

### LINKED ISSUE #6060: Request: could the 500 page become non-blinking?
I'm sorry but I find the blinking 500 page (oops.gif) extremely annoying. Could it please be replaced by a non-blinking image? Thank you!

* * * *

clarfonthey:
Fixes #6060. The graphic should not be animated; this issue was [mentioned](https://github.com/tootsuite/mastodon/pull/5099#issuecomment-332009129) on the original PR and was ignored.

I kept the old GIF in the PR so that it's not forgotten, but I did make a new image that only has one frame. This image is also a GIF in case an admin wishes to copy it over the existing `oops.gif` without upgrading Mastodon.


clarcharr:
@Gargron Would you mind elaborating on your opinion regarding this PR? I'm not sure what the confused react means, and you were the one who made the original PR.

clarcharr:
@dolfinsbizou I agree! This is mostly reverting the previous PR until a better image is provided; I'm more than happy to use whatever people want, just, I don't think that the current GIF is good for multiple reasons.

In addition to the flashing, it just looks very aggressive and stressful. The goal of the error page design was to clarify to the user that the error is not their fault and to not worry, while this image doesn't really convey that.

I do like the aesthetic of the new version more than the mastodon being carried by the birds, though.

clarcharr:
@Gargron thoughts?

akihikodaki:
I am for this change. I think it is not only a matter of preference, but it is also about accessibility.

[Web Content Accessibility Guidelines (WCAG) 2.0](https://www.w3.org/TR/WCAG20/) ha a guideline about flashing:
> a [flash](https://www.w3.org/TR/WCAG20/#flash-def) or rapidly changing image sequence is below the threshold (i.e., content __passes__) if any of the following are true:
> 1. there are no more than three __general flashes__ and / or no more than three __red flashes__ within any one-second period; or
> 2. the combined area of flashes occurring concurrently occupies no more than a total of .006 steradians within any 10 degree visual field on the screen (25% of any 10 degree visual field on the screen) at typical viewing distance
>  where:
> * A __general flash__ is defined as a pair of opposing changes in [relative luminance](https://www.w3.org/TR/WCAG20/#relativeluminancedef) of 10% or more of the maximum relative luminance where the relative luminance of the darker image is below 0.80; and where "a pair of opposing changes" is an increase followed by a decrease, or a decrease followed by an increase, and
> * A __red flash__ is defined as any pair of opposing transitions involving a saturated red.

The blinking image violates the statement in some environments.

I understand the elephant is cute and I appreciate it, but the accessibility must be first-priority.

dolfinsbizou:
@akihikodaki Thank you for the link. Using this definition, can we say that the current animation has one or more general flashes? According to the WCAG 2.0, a general flash is a sudden change in luminance, and there is not that on the current animation (at least, I don't think so).
Even though the movement is quite "brutal", it's not a flash. However, it's a movement, and the WCAG mentions moving informations:
> **2.2.2 Pause, Stop, Hide:** For moving, blinking, scrolling, or auto-updating information, all of the following are true: (Level A) 
> * **Moving, blinking, scrolling:** For any moving, blinking or scrolling information that (1) starts automatically, (2) lasts more than five seconds, and (3) is presented in parallel with other content, there is a mechanism for the user to pause, stop, or hide it unless the movement, blinking, or scrolling is part of an activity where it is essential [...]

Well, can we consider the elephant friend as an "information"? Is it worth adding animation controls?

To be sure that the current animation doesn't have a flash, we should check if there is an effective change of luminance greater than 10% between two frames (it could occur on the elephant's computer screen, or maybe on the orange dashes above his head).

serin-delaunay:
The gif is might depict physical movements, but that doesn't mean that the animation does not contain flashes. Here are the four frames of the animation, which as far as I can tell runs at 10 frames per second (meaning that a single frame transition can happen three times within one second):

frame_0_delay-0 1s

frame_1_delay-0 1s

frame_2_delay-0 1s

frame_3_delay-0 1s

The last two frames have large areas of changing luminance, as visible in this blend:

54e52e39-be02-4d45-8cf5-4fe5a0fbedf7

However, even if the luminance change is below 10% or the area covered by the changes is less than 0.06 steradians, it should be noted that the flash thresholds above are defined by being able to cause seizures. It's already been established in #6060 that there are other ways for this animation to harm users; to accommodate autistic users and users with vestibular disorders, the thresholds may be lower.

>Well, can we consider the elephant friend as an "information"? Is it worth adding animation controls?

Mastodon already has animation controls for logged-in users, but this animation ignores them. However, I don't think extending that option to cover the error page is a good solution; it would still punish users who encounter an error page while on another instance.

clarcharr:
Not to mention that this is the 500 error page, which should work regaadless of user settings, considering how this page can show up when there's an error that prevents those settings from being loaded.

dolfinsbizou:
@serin-delaunay Thank your for your in depth explanations! I was indeed forgetting vestibulars disorders. Considering all this, it appears reasonable to try to look for another graphic (mildly animated or even static) to replace this one.

koyuawsmbrtn:
can a feature be added, that the error pages will be getting the users' setting for autoplaying gifs when they're logged in? i like this animation and as far as i know it's not enabled by default.

Cassolotl:
@koyuawsmbrtn That would only work on the instance you were logged in on. Instances you've not joined wouldn't know your setting, but they could be paused by default?

koyuawsmbrtn:
@Cassolotl that's exactly what I meant

clarcharr:
Going to just close this PR because it was ignored.

PR #6199, opened 2018-01-05
Fix column headers accessibility

marrus-sh:
Styling fixes for #6179 since I don't have push access; see there for discussion.


beatrix-bitrot (approved):

Gargron:
Clickable area has been reduced to the line-height of the "Home" label, which is quite noticeable as I rarely aim to click exactly in the middle of the header. (18px clickable, 30px unclickable)

PR #6200, opened 2018-01-05
Fix PuSH workers

ClearlyClaire:
PuSH subscription/unsubscription/confirmation appears to have been buggy for a while.
This patch uses `Rails.logger` instead of the undefined name `logger` where needed.

### DIFF
--- a/app/workers/pubsubhubbub/subscribe_worker.rb
+++ b/app/workers/pubsubhubbub/subscribe_worker.rb
@@ -20,7 +20,7 @@ class Pubsubhubbub::SubscribeWorker
 
   sidekiq_retries_exhausted do |msg, _e|
     account = Account.find(msg['args'].first)
-    logger.error "PuSH subscription attempts for #{account.acct} exhausted. Unsubscribing"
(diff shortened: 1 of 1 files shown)


Gargron:
Undefined? https://github.com/mperham/sidekiq/wiki/Logging

ThibG:
Weird. Anyway, I had that kind of stuff in my logs, and the proposed fix did fix it:

> déc. 30 22:24:47 neiro bundle[409]: 2017-12-30T21:24:47.116Z 409 TID--byr8sk Pubsubhubbub::SubscribeWorker JID-29c14c3512d559af2c7733ab WARN: NameError: undefined local variable or method `logger' for Pubsubhubbub::SubscribeWorker:Class

ThibG:
Oh, it is probably just the `sidekiq_retries_exhausted` block that is affected.

ThibG:
Revised the patch to change only the required line, using `Sidekiq.logger` instead of `logger`.

Gargron (approved):

PR #6203, opened 2018-01-06
Report weekly active users and weekly toots in /api/v1/instances

ghost:
In #6125, we get number of  weekly active users and weekly toots at /api/v1/instance/activity. Further more, weekly statistics also should be reported at /api/v1/instances.


Gargron:
I'm sorry but I don't think it's necessary. However, thank you for your contribution!

PR #6207, opened 2018-01-06
Allow HTTP caching of atom-rendered public toots (OStatus compatibility)

ClearlyClaire:
Enable the same kind of HTTP caching as for ActivityPub

### DIFF
--- a/app/controllers/stream_entries_controller.rb
+++ b/app/controllers/stream_entries_controller.rb
@@ -10,6 +10,7 @@ class StreamEntriesController < ApplicationController
   before_action :set_stream_entry
   before_action :set_link_headers
   before_action :check_account_suspension
+  before_action :set_cache_headers
 
   def show
     respond_to do |format|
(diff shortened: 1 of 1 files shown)


app/controllers/stream_entries_controller.rb
```diff
@@ -19,6 +20,8 @@ def show
       end
 
       format.atom do
+        skip_session! unless @stream_entry.hidden?
+        expires_in 3.minutes, public: true
```
ykzts: `public: !@stream_entry.hidden?, private: @stream_entry.hidden?`

If you do not do as above, does private toot caching on CDN?

ThibG: In practice, this shouldn't happen (because the session cookie is still set, preventing caching, and because a logged-in user requesting a private toot this way is extremely unlikely), but it is something I have overlooked, you're right.

ykzts (approved):

ThibG:
Ping @Gargron

ThibG:
@Gargron I don't know if you're avoiding this because it's OStatus, but handling OStatus requests faster will also free more CPU time for ActivityPub requests. In situations where hundreds of instances (including old OStatus-only Mastodon instances as well as other OStatus-compatible software), every speedup counts, including OStatus

Gargron (approved):

Gargron:
@ThibG Funnily enough yes, I was avoiding this because it's OStatus.

ThibG:
@Gargron thanks! I don't want to sound harsh, but I would appreciate if you voiced such concerns instead of silently avoiding pull requests. It's easier to discuss objections we know about…

PR #6214, opened 2018-01-07
Fix unintended cache

ykzts:
see also #6175


unarist (approved):

ThibG:
To be honest, I don't understand what this is fixing: passing a false-y value to “public:” already prevents caching, doesn't it? And you're defaulting to `true`, so I can't see how it can prevent unintended caches?

ClearlyClaire (approved):

PR #6218, opened 2018-01-07
Move Article from supported to converted types

puckipedia:
So this may sound like a weird PR, but:

- Nothing currently uses `Article` (I want to, though)
- `Article` is defined as a "any kind of multi-paragraph written work", and thus, doesn't fit well within the limits of Mastodon toots

So, why not, like Image/Video now does, just show the name of the post plus a link to it? This allows people on Mastodon to reply to Articles without getting them squashed into a tiny column.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class ActivityPub::Activity::Create < ActivityPub::Activity
-  SUPPORTED_TYPES = %w(Article Note).freeze
-  CONVERTED_TYPES = %w(Image Video).freeze
+  SUPPORTED_TYPES = %w(Note).freeze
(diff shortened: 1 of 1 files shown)


Gargron (approved):

nightpool:
how does this affect federation w/ Hubzilla? Don't they use Article?

Gargron:
Do they? Do they have titles for each post? It feels like they shouldn't be called articles unless they do. I really like the idea of allowing macroblog software to interact with Mastodon in the same way as PeerTube, but we have treated Article and Note equally so far so if Hubzilla used the other type for their posts then it's on us to not suddenly change it...

Gargron:
Worst case, we can make it so the title + link behaviour is used when title is present, falling back to the previous behaviour if the title is missing.

nightpool:
I'll double check and circle back around—it's a problem for another day.

gled-rs:
Plume uses Article...

PR #6219, opened 2018-01-08
Fix bad URL schemes being accepted

Gargron:
- Only http/https
- URL (displayed to the user) and URI must be on the same host
- For posts: URL and URI must be on the same host as actor's URI

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -255,10 +255,10 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   def invalid_origin?(url)
     return true if unsupported_uri_scheme?(url)
 
-    needle   = Addressable::URI.parse(url)
-    haystack = Addressable::URI.parse(@account.uri)
+    needle   = Addressable::URI.parse(url).host
+    haystack = Addressable::URI.parse(@account.uri).host
 
-    !haystack.host.casecmp(needle.host).zero?
+    !haystack.casecmp(needle).zero?
   end
 
   def reply_to_local?
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -5,7 +5,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   CONVERTED_TYPES = %w(Image Video).freeze
 
   def perform
-    return if delete_arrived_first?(object_uri) || unsupported_object_type?
+    return if delete_arrived_first?(object_uri) || unsupported_object_type? || invalid_origin?(@object['id'])
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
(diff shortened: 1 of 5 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -245,6 +252,15 @@ def skip_download?
     @skip_download ||= DomainBlock.find_by(domain: @account.domain)&.reject_media?
   end
 
+  def invalid_origin?(url)
+    return true if unsupported_uri_scheme?(url)
+
+    needle   = Addressable::URI.parse(url)
+    haystack = Addressable::URI.parse(@account.uri)
```
nightpool: this may be more clear if you do the \.host calls here, so it's:

```ruby
object_domain = Addressable::URI.parse(url).host
actor_domain  = Addressable::URI.parse(@account.uri).host
```

nightpool (commented):
do we think same-origin is a reasonable restriction to apply to urls as *well* as ids? i'm down to implement if so, but this may break the assumptions of, e.g., bridgy.fed.

nightpool (changes requested):
LGTM with the host logic cleaned up.

nightpool (approved):

snarfed:
hi all! apologies for being late to the party here. @nightpool you were right, this did break https://fed.brid.gy/ in some ways. mind elaborating on the motivation and the problems you were hoping to solve? was it spoofing? something else?

as background, bridgy fed translates between ActivityPub and IndieWeb (microformats in HTML + webmention). it tries to proxy as transparently as possible, so we'd like to continue using IndieWeb users' own domains and post URLs as AP actor and object `url`s.

we do wrap actors' `id`s, e.g. https://fed.brid.gy/snarfed.org instead of https://snarfed.org, in order to translate their microformatted HTML and serve it as AS2. we could also wrap and redirect actor and object `url`s, which this change seems to require, but then the user-visible actor and post links on mastodon would include fed.brid.gy. ([examples in the last two replies here.](https://mastodon.social/@swentel/100776214865184909)) that still works, but it's somewhat ugly and misleading.

more background in https://github.com/snarfed/bridgy-fed/issues/16#issuecomment-360934786 and below. cc @swentel. let us know what you think! we're open to ideas. thanks in advance!

nightpool:
yes, the concern is spoofing. happy to talk more about this on discord or IRC, but since there's no "linkback" check on the URL of a post or actor, I think it's pretty important that users can trust it not to be spoofed.

nightpool:
actor urls are the main concern here.

snarfed:
thanks for the explanation! makes sense. i'm definitely still curious though, if you all were building a proxy like bridgy fed, what you'd do for actor and object ids and urls. let me know if you have any thoughts!

(i haven't seen you on IRC in #mastodon yet, but i'm alo happy to talk more there.)

Gargron:
Use your own bridgy URLs for everything when requested with json type, redirect to real URL when requested with html type, that might work.

snarfed:
thanks! yeah, that's the baseline. i was hoping for something a bit more elegant and transparent that used the user's domain and URLs directly, but this is an ok compromise. thanks!

snarfed:
just so you all know, we did this, and it works fine. examples here: https://mastodon.social/@swentel/100776214865184909

it's awkward and a bit ugly, as expected, but no matter. vive la difference. thanks for your support!

PR #5772, opened 2017-11-21
Allow to dereference Follow object for ActivityPub

akihikodaki:
Resolves #5770.

Issue #6211, opened 2018-01-06
new undo follows have *no* ID


puckipedia:
```
{
    "id": "https://mastodon.social/users/phschn#follows/1482040/undo",
    "type": "Undo",
    "actor": "https://mastodon.social/users/phschn",
    "object": {
        "type": "Follow",
        "actor": "https://mastodon.social/users/phschn",
        "object": "https://puckipedia.com/"
    },
    "signature": {
        "type": "RsaSignature2017",
        "created": "2018-01-06T18:37:15Z",
        "creator": "https://mastodon.social/users/phschn#main-key",
        "signatureValue": "AuKk5Fsbm4wV6GP2etr0dZLRsyn2AMXXqy56fFo4Wd4OTrPpFurvljsvkeh2mGzGMweoPTP83C/oRC01T4D61EhEo1RAIYKYTE11NeL+Wl8gLPTC6+F5xCQzLCfF00TMBDNL2h2dEnbPsDXjeMrjy1kDSVVmpAFpAnVs3ng6200P5NgvJlpZcBOQbNS/M8OEeGDh/w5y9cmpSRJKDZr3CA6AVihR5FqCvpn9Z6MSCjzR2OKw5rU+Oie8CBkdGwVcJmHBbj68uR3ox6c+smAEcnVM+TVMCzw2GcDkcd0wI4bnUUOSJGjf7fL2P/QFKroANDAfqWe05fPKecZqA7dbzQ=="
    }
}
```
* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nightpool:
`https://mastodon.social/users/phschn#follows/1482040/undo`?

puckipedia:
I meant the object inside the Undo

nightpool:
Ah. Well.

akihikodaki:
This is intentional, but since here is a question, we may need a review if it is ActivityPub-conforming.

nightpool:
Note that I don't think this is STRICTLY incorrect, because nested objects can have null ids, indicating that they're anonymous parts of the parent context.

However, all ids have to be dereferenceable and https://mastodon.social/users/phschn#follows/1482040/undo definitely isn't.

nightpool:
(closed because #6221 was merged)

puckipedia:
[...]

PR #6221, opened 2018-01-08
Revert #5772

Gargron:
https://github.com/tootsuite/mastodon/pull/5649#issuecomment-355847223


spec/models/follow_request_spec.rb
```diff
@@ -1,38 +1,25 @@
...
+      expect(follow_request).to model_have_error_on_field(:account)
     end
 
-    it 'correctly passes show_reblogs when false' do
-      follow_request = Fabricate.create(:follow_request, show_reblogs: false)
-      follow_request.authorize!
-      target = follow_request.target_account
-      expect(follow_request.account.muting_reblogs?(target)).to be true
-    end
-  end
```
nightpool: these removals seem unrelated?

nightpool: these removals seem unrelated?

nightpool (changes requested):
LGTM with spec changes cleaned up (some show_reblog changes got caught in the cross-fire)

nightpool (approved):

akihikodaki:
This should be merged as soon as possible if #6211 is making Mastodon incompatible with Kroeg, and later we can investigate the behavior should be fixed.
Otherwise, #6224 should be merged instead of this.

PR #6224, opened 2018-01-08
Do not mark follow request dereferencable if it is from remote

### LINKED ISSUE #6216: Accepts of Follows use an even wronger ID than before
```
{
    "type": "Accept",
    "id": "https://glitch.social/users/bea#accepts/follows/21608",
    "actor": "https://glitch.social/users/bea",
    "object": {
        "id": "https://glitch.social/users/puckipedia/follows/21608",
        "type": "Follow",
        "actor": "https://puckipedia.com/",
        "object": "https://glitch.social/users/bea"
    }
}
```

@akihikodaki this is somehow even more wrong than before, and Kroeg really really doesn't like it, and it isn't according to spec.
* * * *

akihikodaki:
This resolves #6216.

### DIFF
--- a/app/serializers/activitypub/follow_serializer.rb
+++ b/app/serializers/activitypub/follow_serializer.rb
@@ -22,6 +22,6 @@ class ActivityPub::FollowSerializer < ActiveModel::Serializer
   end
 
   def dereferencable?
-    object.respond_to?(:object_type)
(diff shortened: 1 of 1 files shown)


akihikodaki:
Closes due to preference to #6221.

Issue #5818, opened 2017-11-25
Followers-only toots that your instance knows about already should be searchable using the URL by people who follow the author


Cassolotl:
Scenario:

1. Sandy has been following Paul for like... 3 weeks.
2. Yesterday Paul tooted something on the followers-only privacy setting, and Sandy saved the URL for later boosting or whatever.
3. Sandy pastes the URL into their Mastodon web UI so that they can boost the toot.

**Expected behaviour:** Sandy's instance "knows" Paul's followers-only toot and knows that Sandy follows Paul. Sandy's search should bring up the toot.

**Current behaviour:** Sandy's search yields 0 results, because the toot they are searching for is set to followers-only.

This might be a duplicate but I honestly don't know what to search for. :S It is inspired by [this thread](https://mastodon.starrevolution.org/@Laurelai/99066325170584371).

* * * *

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nightpool:
it is maybe possible to do this in the case where the status has already been populated, *if* the searched for URI is the same as the url or uri that was sent with the post (i.e. there's no redirection happening).

it is probably not currently feasible for us to fetch remote media with this, even though I could imagine a HTTP Signatures way of doing it.

puckipedia:
not parsing HTTP signatures on GETs for statuses is a thing that I've been missing due to my Kroeg instance not being as well populated and trying to retrieve entire post chains, but failing because there's posts that might be follower-only, but before I was a follower

trwnh:
this is supported now

Cassolotl:
Oh great, then I will close this. Thanks! :)

PR #6225, opened 2018-01-08
Allow retrieval of private statuses (single or in outbox) using HTTP signatures

puckipedia:
This defines a new concern that allows a controller to use either the normal cookie auth, or fall back to HTTP signatures. The outbox controller and statuses controller then use this concern to allow remote servers to get private statuses, by authenticating as a specific user. 
This is the server-to-server communication required for #5818, effectively

### DIFF
--- a/app/controllers/activitypub/outboxes_controller.rb
+++ b/app/controllers/activitypub/outboxes_controller.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class ActivityPub::OutboxesController < Api::BaseController
+  include SignatureVerification
+
   before_action :set_account
 
   def show
-    @statuses = @account.statuses.permitted_for(@account, current_account).paginate_by_max_id(20, params[:max_id], params[:since_id])
+    @statuses = @account.statuses.permitted_for(@account, signed_request_account).paginate_by_max_id(20, params[:max_id], params[:since_id])
     @statuses = cache_collection(@statuses, Status)
 
     render json: outbox_presenter, serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, content_type: 'application/activity+json'
--- a/app/controllers/concerns/signature_authentication.rb
+++ b/app/controllers/concerns/signature_authentication.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module SignatureAuthentication
+  extend ActiveSupport::Concern
+
+  include SignatureVerification
+
+  def current_account
+    super || signed_request_account
+  end
+end
--- a/app/controllers/statuses_controller.rb
+++ b/app/controllers/statuses_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class StatusesController < ApplicationController
+  include SignatureAuthentication
   include Authorization
 
   layout 'public'


Gargron:
I am for this but I'm a bit worried about what could go wrong 🤔 

@ThibG @beatrix-bitrot @nightpool @akihikodaki

ThibG:
I'm not completely sure I'm for this, but I don't think much could go wrong.
Caching is already disabled for private toots, so it won't leak. It's not more work for the server either, since it had to resolve the toot to figure out it was private.

akihikodaki:
I'm even not sure what `ActivityPub::OutboxesController` is currently used for. I am wondering if disabling the default authentication mechanism is possible.

puckipedia:
it's currently used for the /outbox endpoint, which shows the last 10(?) messages in AS2 format (and can be tricked into giving more), this would preferably be paginated at some point so other servers can iterate and take in older messages (e.g. follower-only)

akihikodaki:
I mean I wonder who currently use `/outbox` endpoint.

ThibG:
@akihikodaki well, it's part of the spec. Also, Mastodon currently uses it, but only to get the number of statuses when discovering a remote account…

Also, while this isn't a blocker for this pull request, I really wonder how useful it may be. In particular, how can a client (well, remote server) chose which account they should issue a request on the behalf of? When a given users inputs the toot's URI in the search box, it's fairly reasonable to do the request on their behalf (how did they get the URI, though?), but what about `ResolveThreadWorker` for instance?

ThibG:
Finally, I would also advocate removing the default local-user authentication mechanism for outbox as well as OStatus and ActivityPub rendering of toots if that makes code simpler, as a logged-in user typically won't access toots this way.

puckipedia:
Yeah, there's a bunch of work that is slightly unclear. I myself use the credentials of the user that receives the thread to resolve it, but that's because I lazy load it the moment someone views it...

akihikodaki:
@ThibG 

> well, it's part of the spec.

Mastodon's authentication mechanism is not part of the spec, so it should be possible to keep consistent with other ActivityPub endpoints, which implements only HTTP  signature as authentication mechanism.

> Also, Mastodon currently uses it, but only to get the number of statuses when discovering a remote account…

I see. Such an access should not be affected by authentication, so it is safe to implement HTTP signautre as authentication mechanism.

nightpool:
Removing local authentication would make the pull request much more complicated and make AP stuff harder to debug. :-1:

akihikodaki:
> Removing local authentication would make the pull request much more complicated

That's not true. See `inbox` as reference. It just needs to replace `@account` with `signed_request_account` and `include SignatureVerification`. -2, +4 lines. Much simpler.

> make AP stuff harder to debug.

That's true. But who cares while other AP endpoints accepts only HTTP signatures?

puckipedia:
I feel like removing local account support on status endpoints \*only\* if rendering JSON would be complicated, but I guess I'd be okay if the inbox'd be changed to HTTP signature only, I guess

nightpool:
yes I was talking about the status endpoints.

ClearlyClaire (approved):
I'm ok with those changes, although I think we can drop the usual authentication mechanism for everything but the `StatusesController` (actually, I'm not against dropping it even from there, as a typical logged-in user will read the toots from the Web UI anyway, and not from the “public” view)

puckipedia:
@nightpool @ThibG I changed it so that the outbox only allows HTTP signatures

ThibG:
LGTM

akihikodaki (approved):

Gargron (approved):

PR #6226, opened 2018-01-08
Add the author of a status to cc if reblogged

puckipedia:
This makes slightly more sense, and ensures that the author of a post is always referenced in the audience (which some servers might rely on). And the announce is POSTed to the author's inbox anyways.


Gargron (changes requested):

puckipedia:
Should be cleaned up now

Gargron (approved):

PR #6230, opened 2018-01-09
Skip ActivityPub Announces of non-public objects

Gargron:


nightpool (approved):
LGTM

ThibG:
Hm… ActivityPub does not restrict the `Announce` activity to objects with a public audience.
This change basically makes #6157 useless for Mastodon.

akihikodaki:
> This change basically makes #6157 useless for Mastodon.

We should have a change to accept boosts by the author as soon as possible though such frequent releases may be bothering server administrators.

ThibG:
@akihikodaki I think the only restriction that really makes sense is to not widen the audience of the original toot. E.g., it could be boosted by a follower of the original poster, and that boost would then be seen only by the intersection of the booster's audience and the original poster's audience.

I'm not sure how Mastodon handled that before the change discussed here, though.

akihikodaki:
@ThibG See #6236.

PR #6232, opened 2018-01-10
[DO NOT MERGE] Get Ruby 2.5 support

Exagone313:
I got Mastodon running on Ruby 2.5.0 by updating these two gems (cld3 and devise), but this support is untested. I'm running it on a single user instance on Arch Linux and I'd like to know if tests can pass (will also need to test on 2.5.0 as well).

I think it should also be the time to check if all gems could be updated.

Thoughts?


Exagone313:
Damn, there is already a PR for this.

PR #6235, opened 2018-01-10
Skip CSRF token check in ActivityPub controllers

pfigel:
Prior to #6223, the CSRF token check was skipped for all API controllers (which is what the ActivityPub controllers inherit from). #6223 changed the Api::BaseController to use `:null_session` instead, which
produces "Can't verify CSRF token authenticity." log lines whenever the CSRF token is invalid.

This introduces a separate BaseController class for ActivityPub which skips the CSRF check.

Note: I'm pretty sure there are a few other API controllers for which this is also the case (and where adding the `skip_before_action` would make sense). I'll look through the other controllers tomorrow, this should probably be a WIP for now.

### DIFF
--- a/app/controllers/activitypub/base_controller.rb
+++ b/app/controllers/activitypub/base_controller.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class ActivityPub::BaseController < Api::BaseController
+  skip_before_action :verify_authenticity_token
+end
--- a/app/controllers/activitypub/inboxes_controller.rb
+++ b/app/controllers/activitypub/inboxes_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ActivityPub::InboxesController < Api::BaseController
+class ActivityPub::InboxesController < ActivityPub::BaseController
   include SignatureVerification
 
   before_action :set_account
--- a/app/controllers/activitypub/outboxes_controller.rb
+++ b/app/controllers/activitypub/outboxes_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ActivityPub::OutboxesController < Api::BaseController
+class ActivityPub::OutboxesController < ActivityPub::BaseController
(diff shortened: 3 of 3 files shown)


Gargron:
Is something broken because of this??

patf:
Luckily the log is just informational. The `:null_session` just nils the session if the CSRF check fails, which should be fine for `/inbox`. The request still succeeds.

I'm not 100% certain if this affects other API endpoints. If there's anything that requires the session to be there *and* won't get a CSRF token from some consumers, then maybe?

Gargron:
Ah, okay, nah, nothing in the API needs a session. Had a bit of a panic attack because 2.1.3 just came out earlier today, I couldn't bear another thing breaking.

nightpool:
Does it produce the log lines when the csrf token is missing, or just when
it's invalid? You said invalid in your issue text, and if it's just that
then i don't see why this is a big deal. If it's logging too much, we
should just fix the logs right?

But if we really want to make this change we should disable sessions
entirely rather then deal with possibly of using unverified sessions by
accident

patf:
> Does it produce the log lines when the csrf token is missing, or just when it's invalid?

Both.

> But if we really want to make this change we should disable sessions entirely rather then deal with possibly of using unverified sessions by accident

I don't think this is a concern for the ActivityPub controllers (which is what the PR addresses so far). As for the other places where we might need to skip the check, is your concern that we'd accidentally do so for something that would be vulnerable to CSRF?

Not sure if we can fully disable sessions for API controllers. I see a couple of references to session in `PushSubscriptionsController`. `current_user` seems to be fine (using the doorkeeper token), unless I'm misreading that code.

patf:
I've given this some further thought and think suppressing the logs while keeping `:null_session` for all APIs makes more sense. I'll also add `protect_from_forgery with: :exception` to the one API controller that actually uses sessions so we're consistent with the other non-API controllers (and avoid bugs/side-effects of running that code with empty sessions).

I'll open a new PR for that.

PR #6236, opened 2018-01-10
Accept ActivityPub announce from the author of the original note

akihikodaki:


akihikodaki:
See https://github.com/tootsuite/mastodon/pull/6230#issuecomment-356566481

ThibG:
I would allow handling `Announce` from any user, even if they are not the original author, as long as the boosted toot isn't shown to people excluded from the original audience (e.g., B boosts A's private toots, this boost is then shown to the intersection of A and B's followers).

akihikodaki:
@ThibG That is difficult.

app/lib/activitypub/activity/announce.rb
```diff
@@ -35,6 +36,6 @@ def fetch_remote_original_status
   end
 
   def announceable?(status)
-    status.public_visibility? || status.unlisted_visibility?
+    status.account == @account || status.public_visibility? || status.unlisted_visibility?
```
ThibG: Every other part of the source code compares account identifiers instead of instantiated account objects.

ClearlyClaire (dismissed):

akihikodaki:
d918ee58ecda3665bce75ad94ad1236448dcc8d7 addresses https://github.com/tootsuite/mastodon/pull/6236#pullrequestreview-87834151.

ThibG:
@akihikodaki I don't think it has to be difficult, you may accept `Announce` without filtering, but only display the toot to the `Announce`'s audience if they are in the original toot's audience. It may be error-prone, though, so I understand you'd want to filter upfront…

After a quick look in the code, I think my suggestion could be implemented by editing `status_policy.rb` to check the requested permission on the target toot as well as the reblogged one.

Gargron (approved):

akihikodaki:
Sorry for a late reply, but I still don't think it is easy enough. The difficulty is in its semantics. For example, consider the case that:

1. someone posts a `private` status, and
2. a locked account reblogged the status

Then what would be the visibility of the `Announce` of the locked account? In my opinion, the user of the locked account would expect the visibility of the reblogs made with the account is limited to users authorized by him (and that's why he locks his account.) On the other hand, the author of the reblogged status obviously assumes the visibility of the status is limited to his followers.

Therefore, the visibility of the reblogged status must be the intersection. This is a bit difficult for the database and extremely difficult for ActivityPub. The visibility filtering of statuses has been a barrier for several database optimizations. I am also not sure how to represent the intersection in ActivityPub.

ThibG:
Handling incoming `Announce`s which audience conforms to Mastodon's model seems pretty easy, as I described (just display it to local users that are in the intersection of both audiences).

But you're right, I hadn't really considered federation. And I completely forgot that the announced object is inlined in `Announce`s, as Mastodon always fetch the original object manually. Outgoing `Announce` objects would thus need to be sent to the intersection of the audience, which I'm not completely sure is possible in general… incoming Announces should be fine as long as their audience conforms to Mastodon model. I'm not sure about the other cases, though (but it's a more general issue with Mastodon having its own model for audiences, which doesn't really match ActivityPub).

Issue #6242, opened 2018-01-11
[Suggestion] Follow request badge on web UI


Kjwon15:
Currently, _getting started_ column is look like this:

![screenshot from 2018-01-12 00-36-31](https://user-images.githubusercontent.com/5047683/34832661-de332030-f730-11e7-8c25-124a70ef1648.png)

When someone requests to follow me, It is notified by email. But it's more comfortable when notified on web.

My suggestion is show badge for this.
![image](https://user-images.githubusercontent.com/5047683/34832978-b0e6d634-f731-11e7-88b1-53280ea5a9c4.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #6244, opened 2018-01-11
Implement follow_requests badge

tribela:
image

Related: #6242 

Additional note: I'm very confused to use `getState`. Can someone review for [this](https://github.com/tootsuite/mastodon/pull/6244/files#diff-322ec4ce8678beb1db1e7aee74ed7a7aR37)?


Gargron:
Performing HTTP requests from inside `render` is wrong, passing the entire `getState` to a component is also wrong. Further resources: [Redux](https://redux.js.org/). I appreciate the thought but I would have to re-implement this myself.

PR #6245, opened 2018-01-11
Handle sessions that can't be translated

### LINKED ISSUE #6165: Missing translation for Otter Browser user agent
screenshot_20180103-070156
`browser` gem can detect Otter Browser, but there’s no string for this browser in en.yml file, so it can’t even use the default one… :)

tribela:
This commit fixes #6165.

### DIFF
--- a/app/views/auth/registrations/_sessions.html.haml
+++ b/app/views/auth/registrations/_sessions.html.haml
@@ -16,7 +16,7 @@
             %span{ title: session.user_agent }<
               = fa_icon "#{session_device_icon(session)} fw", 'aria-label' => session_device_icon(session)
               = ' '
-              = t 'sessions.description', browser: t("sessions.browsers.#{session.browser}", default: "#{session.browser}"), platform: t("sessions.platforms.#{session.platform}", default="#{session.platform}")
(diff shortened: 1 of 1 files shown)


app/views/auth/registrations/_sessions.html.haml
```diff
@@ -16,7 +16,7 @@
             %span{ title: session.user_agent }<
               = fa_icon "#{session_device_icon(session)} fw", 'aria-label' => session_device_icon(session)
               = ' '
-              = t 'sessions.description', browser: t("sessions.browsers.#{session.browser}"), platform: t("sessions.platforms.#{session.platform}")
+              = t 'sessions.description', browser: t("sessions.browsers.#{session.browser}", default: "#{session.browser}"), platform: t("sessions.platforms.#{session.platform}", default="#{session.platform}")
```
ykzts: `default="#{session.platform}"` -> `default: "#{session.platform}"`?

ykzts (changes requested):

app/views/auth/registrations/_sessions.html.haml
Kjwon15: Oh!

ykzts (approved):

PR #6251, opened 2018-01-13
Fix home regeneration

### LINKED ISSUE #5634: Returning users come back to empty timelines
I just talked to someone who came back to an empty Home after a long time away from Mastodon. It's understandable that you'd want to clear out timelines for inactive people to save space, but it's not a great experience coming back to an empty Home.

Maybe Mastodon could invoke the new user thing where they follow admin accounts when it detects someone with an existing account and an empty timeline. Or just regenerate the timeline with a friendly "welcome back" message and some useful information about major changes since the last version they used.

* * * *

Gargron:
Fix #5634

image

- There was a bug: The regeneration marker ("regeneration in progress!") was not unset by the RegenerationWorker as it was supposed to. Because of this, it persisted for the full 24h. While that marker is present, the home feed returned results from the DB directly, which due to filtering can be awfully incomplete and sometimes entirely empty.
- The `GET /api/v1/timelines/home` API will now return HTTP status code 206 (Partial Content) when a regeneration is in progress
- RegenerationWorker was moved up from "pull" queue to "default" queue because it needs to finish quickly. People are waiting.
- The web UI will now handle the 206 response code for the home feed: It will display a loading indicator and reload the home timeline API every 3 seconds until the status code is 200.


rainbowmultiverse:
Classic under construction GIF but with a mastodon in a hardhat.

Gargron:
@rainbowmultiverse It's probably going to be a sorting mail metaphor.

nightpool:
I don't know if this fixes the problem @sdukovni reported, where the redis key wasn't even apparently getting set.

Gargron:
@nightpool I was not able to replicate that, tbh. I have erased my home timeline, set my last_sign_in_at and current_sign_in_at values to the ones in the bug report, refreshed the web UI, and the regeneration key was set.

app/workers/regeneration_worker.rb
```diff
@@ -3,7 +3,7 @@
 class RegenerationWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'pull', backtrace: true, unique: :until_executed
```
aschmitz: No more backtraces?

app/javascript/mastodon/features/list_timeline/index.js
```diff
@@ -120,13 +120,17 @@ export default class ListTimeline extends React.PureComponent {
     if (typeof list === 'undefined') {
       return (
         <Column>
-          <LoadingIndicator />
+          <div className='scrollable'>
```
aschmitz: Is there a reason to not make LoadingIndicator and MissingIndicator scrollable on their own?

aschmitz (commented):
Looks fine, just highlighting a few things.

app/javascript/mastodon/actions/timelines.js
```diff
@@ -104,8 +105,12 @@ export function refreshTimeline(timelineId, path, params = {}) {
     dispatch(refreshTimelineRequest(timelineId, skipLoading));
 
     api(getState).get(path, { params }).then(response => {
-      const next = getLinks(response).refs.find(link => link.rel === 'next');
-      dispatch(refreshTimelineSuccess(timelineId, response.data, skipLoading, next ? next.uri : null));
+      if (response.status === 206) {
+        dispatch(refreshTimelineSuccess(timelineId, [], skipLoading, null, true));
```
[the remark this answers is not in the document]
Gargron: This is intended

app/javascript/mastodon/features/list_timeline/index.js
Gargron: No but I would abstract if there's >= 3 usages.

app/workers/regeneration_worker.rb
Gargron: This class does not throw exceptions, it's not necessary. If it ever does throw one, it'll still show up in the logs, just without a huge backtrace.

aschmitz (approved):

ykzts (approved):

jeroenpraat:
@Gargron is there a workaround to fix this for users who are currently (v2.1.3) affected?

voronoipotato:
Well Garg I had tried, and I'm glad I didn't put too much effort into this because I know jack diddly about redis. Good catch.

PR #6256, opened 2018-01-15
HTML e-mails (UserMailer)

Gargron:
**Edit:** I have fixed the inconsistencies between "e-mail" and "email" but have not updated the screenshots.

screen shot 2018-01-15 at 04 30 31
screen shot 2018-01-15 at 04 30 41
screen shot 2018-01-15 at 04 30 46
screen shot 2018-01-15 at 04 30 51
screen shot 2018-01-15 at 04 30 57


Pbtflakes:
My first impression is that the templating system is very good, but the HTML mailers decrease legibility for end users. I can read a few lines of unformatted text quickly and easily in the current messages, since I don't need to sift through the icons, buttons, etc., and then strain my eyes to read gray text on a gray background.

Personally I would advocate for sending these routine messages with only text/plain mimetype, but I hope that at least they won't be turned into small webpages.

nightpool:
@Pbtflakes Having HTML emails increases mastodon's credibility and perception a lot amongst non-technical users, and a well-designed email is *significantly* more likely to be noticed and understood by users then a plain text one. this is a very welcome change.

Gargron:
@Pbtflakes Plus the plain text e-mails are always there (multipart) so if you prefer them, you can always have them displayed in your e-mail client.

Pbtflakes:
@Gargron If we're talking "non-technical" users, I have yet to find a webmail interface that supports switching to plaintext display, so "always" is more "if you've already set up a local mail client."

@nightpool I fail to see how well-designed and plain text are mutually exclusive for email, or how "non-technical" users are supposedly unable to understand a plain text message which spells out exactly the same content with more legible formatting.

Whether they're technical or non-technical, people simply can't read gray text on gray backgrounds as well as black text on a white backgrounds, there needs to be more contrast.

tylersticka:
I'm really digging the design of these templates! I did have a couple of points of feedback, I apologize in advance if this is the wrong place for design notes.

Assuming my [Contrast](https://usecontrast.com/) app isn't fibbing (sometimes the hex values it picks are just a _tad_ off, so it's possible), the button text might be _just_ shy of [WCAG AA](http://accessible-colors.com/) contrast recommendations (it passes for AA+, so if it's over `18px` then please disregard), and the footer copy outright fails.

screen shot 2018-01-15 at 4 54 05 pm
screen shot 2018-01-15 at 4 54 18 pm

My only other design-related comment is that the design of the "new email address" template kinda cuts off the sentence, especially in comparison to the other two...

> The email address for your account has been changed to:
>
> ---
>
> foo@bar.com

It's a somewhat awkward place for a section break, IMO.

Other than those minor quibbles, I think these convey a lot more continuity with the typical Mastodon user interface. I like it! 👍

ykzts (approved):
nice work 👍

PR #6257, opened 2018-01-15
Surround mid-text display names with bdi tags

### LINKED ISSUE #1095: Need defensive measures against Unicode direction overrides
If you put U+202E RIGHT-TO-LEFT OVERRIDE in your display name, the effects spill out of the display name field and affect other parts of the UI, as shown here:

screen shot of the bug described above

There are probably a bunch of other closely-related bugs.

(Inspired by a similar bug in Twitter found by [@0xabad1dea](https://twitter.com/0xabad1dea).)

* * * *

Gargron:
Fix #1095

See https://github.com/tootsuite/mastodon/issues/1095#issuecomment-357559460

### DIFF
--- a/app/javascript/mastodon/components/__tests__/__snapshots__/display_name-test.js.snap
+++ b/app/javascript/mastodon/components/__tests__/__snapshots__/display_name-test.js.snap
@@ -4,14 +4,16 @@ exports[`<DisplayName /> renders display name + account name 1`] = `
 <span
   className="display-name"
 >
-  <strong
-    className="display-name__html"
-    dangerouslySetInnerHTML={
-      Object {
-        "__html": "<p>Foo</p>",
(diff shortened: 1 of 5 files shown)


wxcafe:
:/

Sylvhem:
*pat, pat* @wxcafe

joyeusenoelle:
Suggestion: Could this be a per-user toggle, so people bothered by the UI interference can turn it off for their displays without affecting anyone else?

lanodan:
> per-user toggle
Appart for keeping the “wxcafe glitch” what is the use case? I think this would add ~useless code.

nightpool:
I'm not against this in principle, but you'll have to find someone else to approve it :P

nightpool:
Push notifications are also affected iirc

PR #6258, opened 2018-01-15
Display unfollow button even if account moved

### LINKED ISSUE #6128: Impossible to unfollow someone who's "moved"
Here's one of my old accounts that I still follow from ephemeral.glitch.social, but I've set up that redirect notice:

screen shot 2017-12-29 at 12 04 53

It's not possible to unfollow them, because the follow/unfollow icon has gone.

I think the icon should be there when you follow them, so that it's possible to unfollow - but when you have unfollowed them it should disappear so you can't follow them again.

* * * *

Gargron:
Fix #6128

### DIFF
--- a/app/javascript/mastodon/components/account.js
+++ b/app/javascript/mastodon/components/account.js
@@ -93,7 +93,7 @@ export default class Account extends ImmutablePureComponent {
             {hidingNotificationsButton}
           </Fragment>
         );
-      } else if (!account.get('moved')) {
+      } else if (!account.get('moved') || following) {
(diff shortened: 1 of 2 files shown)


mkljczk (approved):

ykzts (approved):

MrH4nds:
has this been fixed? i had some trouble unfollowing recently, although eventually found a solution: https://docs.joinmastodon.org/user/contacts/#relationships 

...not sure why it this function is missing in the main interface, but i guess the separate menu works!

PR #6273, opened 2018-01-16
Replace tutorial modal with welcome e-mail

Gargron:
- [x] Remove tutorial modal
- [x] Add welcome e-mail

image
image


nightpool:
Lots of people don't check their email, I don't think this is a very good change.

Gargron:
>Lots of people don't check their email

Only one person ever told me they clicked through the tutorial. Everyone else immediately closes it. It's a waste of translation time and code.

An e-mail can be re-opened at any time, and it works even if somebody's using an app and not the web UI, or have closed it. It can serve as a reminder to get back on. And besides, they *just* checked their e-mail to even login in the first place!

app/views/layouts/mailer.text.erb
```diff
@@ -1,5 +1,5 @@
 <%= yield %>
 ---
 
-<%= t('application_mailer.signature', instance: site_hostname) %>
+<%= t 'about.hosted_on', domain: site_hostname %>
```
aschmitz: No parentheses?

aschmitz (approved):
Haven't actually tested the email, but the code changes look fine.

app/views/layouts/mailer.text.erb
Gargron: I'm not sure what you're asking... That's valid ruby syntax. Code style wise we don't really have a rule either way, but I figured in templates I wanna keep stuff as short as possible.

aschmitz: Yeah, it seems mixed in the rest of the codebase. I just called it out here because you removed them. :smile:

Gargron: That was more about me copypasting the line from the HTML template since I want the plaintext and HTML versions to use the same strings (I will have a separate PR cleaning up the remaining user_mailer text templates)

TheKinrar (approved):

yiskah:
This seems like a rushed decision that did not involve a lot of feedback solicitation from different demographics. How many people exactly did you ask about their experience? 10? 30? 500? There's a million users. Of the people you asked, who was it? How many worked in IT? Or engineering? How many were artists? How many don't know any code? I have a habit of welcoming new people a lot, and often the tips I give I get rebuffed with "yeah it said that in the welcome thing." My anecdotal experience conflicts with yours. Scientifically speaking we can't say either one of these is valid and the other not.

The benefit of the modal to someone who needs it is far greater than the annoyance of clicking "skip" once and never worrying about it again. 

Add a welcome email, sure, but why remove the modal? I don't want to toot my own horn but I had FOSS UX design groups asking me to do write-ups about the pedagogy of the modal because it was so well-received and highly regarded. I had devs from other federated social networks asking me to make one like Mastodon's for their software (which I didn't end up doing for reasons which aren't relevant.) Ever since we added the modal I haven't gotten constant questions about what the federated timeline is. It's overall positive. The purpose isn't just "retention" it's just overall improving the experience by having the app explain itself.

rainbowmultiverse:
I usually hate these popups, but I also usually figure out interfaces fast on account of being a huge nerd.

This seems like a good compromise: https://glitch.social/@kibi/99373973266856098

Maybe it could be made more prominent to new users. A lot of webapps seem to go that route. There's a little bouncy thing (motion sickness risk) or highlighting on the guide link.

aschmitz:
I should note that my approval of this PR was for its implementation of
what it intended to do, without really thinking about whether it was a good
idea to do or not.

I think there's a pretty good argument to be made that the onboarding modal
was useful and instructive, particularly with it being "in context" when
users were using it. I think kibi's suggestion for being able to bring it
back up is good as well (and perhaps putting that in the "getting started"
links is appropriate).

That said, I do think the new email looks more professional and is more
informative than the old ones. I do like the idea of adding some
screenshots to it (like the onboarding modal has), but I certainly don't
think scrapping it is the right option.

What about keeping the emails (and incrementally improving them in the near
future), and restoring the onboarding modal? For people who've read the
email thoroughly first, it at least points out where things are, and is
easily skipped if necessary. For people too excited to use Mastodon to read
their email, it gives them information at exactly the right time.

Issue #828, opened 2017-04-04
Direct messages should not increment one's post count


codl:
Given that direct messages are supposed to be private, I dont think anyone should be able to tell that one is sending DMs just by watching their post count increase

PR #6281, opened 2018-01-17
use-counter_cache-rails

masarakki:
Rails's `counter_cache: true` run updating count query every create/detele child item,
and it is a cause of very heavy or deadlock on purging statuses when account deletion/suspend.

This is an idea that store `statuses_count` in Redis instead of Rails's `counter_cache`,
create/detele status updates only redis value,
and write back to database every 5 minutes for `sum(:statuses_count)` in `/about` .


Gargron:
Okay, so the question I've got is... Why even bother updating/keeping a SQL column instead of just using Redis cache and falling back to `SELECT COUNT()`?

masarakki:
In `/about/more` page, it was `Status.local.count`, but  it changed to `Account.local.sum(:statuses_count)` because of this query was very slow. (#4996)
So I think it should be kept.

app/models/account.rb
```diff
@@ -79,6 +79,12 @@ class Account < ApplicationRecord
   has_many :mentions, inverse_of: :account, dependent: :destroy
   has_many :notifications, inverse_of: :account, dependent: :destroy
 
+  counter_cache :statuses
+  STATUSES_COUNT_UPDATED_CACHE_KEY = '_statuses_count_updated_ids'
```
akihikodaki: What is prefix `_` for?

app/workers/statuses_count_writeback_worker.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class StatusesCountWritebackWorker
+  include Sidekiq::Worker
+
+  def perform(account_id)
+    account = Account.find(account_id)
+    account.update(statuses_count: account.statuses_count(force: true))
```
abcang: If force is true, this query will be executed multiple times every 5 minutes.

```sql
SELECT COUNT(*) FROM "statuses" WHERE "statuses"."account_id" = $1
```

I am worried about this.

masarakki: I think redis-base counting may be diffarence from real numbers, 
so I want clear cache periodically.
I think it is best to clear cache before write to DB.
do you have any good idea?

app/workers/scheduler/statuses_count_writeback_scheduler.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+require 'sidekiq-scheduler'
+
+class Scheduler::StatusesCountWritebackScheduler
+  include Sidekiq::Worker
+
+  def perform
+    redis.sscan_each(Account::STATUSES_COUNT_UPDATED_CACHE_KEY) do |account_id|
+      StatusesCountWritebackWorker.perform_async(account_id)
+      redis.srem(Account::STATUSES_COUNT_UPDATED_CACHE_KEY, account_id)
```
akihikodaki: You may move swap this line with the previous line. Otherwise it would have a race condition like this:
1. The scheduler found cache key
2. The scheduler performs status count writeback
3. Another thread performs an update and sets cache key
4. The scheduler removes cache key

In this case, the update performed in step 3 will be ignored.

It is unlikely to be a problem because `StatusesCountWritebackWorker` is asynchronously executed and therefore it would run after the cache key removal in most cases. However I don't think you have to introduce such a danger.

app/workers/statuses_count_writeback_worker.rb
akihikodaki: Maybe simply remove `force`? Actually I don't know the reason why it should have `force`.

masarakki:
fixed.

why travis not finished??
i have no permission to see travis so i can't know why it stopped

Gargron:
To re-run travis, you can submit an empty commit (´git commit --allow-empty´)

I cannot fix this because the travis build doesn't exist at all, I can't click on it to rerun it.

Gargron:
This PR switches us from Rails built-in counter_cache to a gem written by @iguchi1124. I have two concerns about this:

1. Maintenance of that gem (it only has 8 stars and 2 contributors at the moment)
2. There exists an old issue I'd like to eventually resolve, #828. It would be possible to solve it by using counter_culture gem instead of counter_cache-rails, but counter_culture does not write to Rails.cache. Can we... integrate those two ideas?

iguchi1124:
@masarakki  @Gargron 
Thanks for pick up my gem!

If you like to use this, I'd like to add counter_cache-rails maintainers from this organization members or move repository into this organization.

masarakki:
It seems that travis doesn't restart even if I pushed rehashed commit.
what can i do?

masarakki:
https://github.com/iguchi1124/counter_cache-rails/pull/10

@Gargron , @iguchi1124 

This pull-request enable to increment/decrement total-statuses  on cache,
and it can replace WritebackWorker to simple api `after_cache_statuses_count` .

Gargron:
@masarakki Yes I like that idea... It would be nice to finally merge this PR, but please make sure CodeClimate passes, e.g. no space between -> in lambda: `->(arg) { ... }`

Gargron:
https://github.com/tootsuite/mastodon/pull/7302#issuecomment-392367360

PR #6282, opened 2018-01-17
Show Pinned toots in WebUI Profile page

### LINKED ISSUE #6117: Pinned toots do not show up in the multi-column UI.
If you pin a toot, it only shows up when you visit server.name/@username in a new tab. It does not show up when you bring up a user's profile in the main web UI by clicking on their username.

I've seen this on both my own instance (which is still on 2.0.0, I need to spend a day wearing my Admin Hat) and on mastodon.social.

* * * *

- [ x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

rinsuki:
image

close #6117.

**WARNING**: this pull-request is include minimum breaking change of API.
please check a74f81b commit message before merge


app/javascript/mastodon/locales/en.json
```diff
@@ -239,6 +239,7 @@
   "status.show_more": "Show more",
   "status.unmute_conversation": "Unmute conversation",
   "status.unpin": "Unpin from profile",
+  "status.pinned": "Pinned toot",
```
m4sk1n: Strings should be sorted alphabetically, run `yarn manage:translations`, you can also add Polish string for me – “Przypięty wpis”.

rinsuki: added :D

Gargron:
I'm afraid this is missing a crucial part which is making the pinned toots display on remote instances. Currently they consistently do not display on web UI and always display on public profile. After this PR local users would suddenly have pinned toots on web UI but remote users would not. I can't accept that.

Making pinned toots federate is a bit more involved than this. Not only do we need a Collection linked from the Actor, we also need Add/Remove (to/from Collection) activities to be sent out when you pin/unpin a toot. I don't expect you to implement this, I am planning to get around to it eventually.

m4sk1n:
Shouldn't it be closed because of #6610?

PR #6283, opened 2018-01-17
When must_be_following_dm is on, only notify if recipient dm'ed user

renatolond:
Currently, when must_be_following_dm is on, if a user sends a direct message replying to any status from the recipient, the recipient gets a notification. This should not be the case, as if the recipient posted something publicly this can be used to spam their notifications.

### DIFF
--- a/app/services/notify_service.rb
+++ b/app/services/notify_service.rb
@@ -53,13 +53,8 @@ class NotifyService < BaseService
     @notification.type == :mention && @notification.target_status.direct_visibility?
   end
 
-  def replied_to_status_is_direct_message?
-    replied_to_status = Status.find(@notification.target_status.in_reply_to_id)
-    replied_to_status.direct_visibility?
-  end
-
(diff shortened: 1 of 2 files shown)


app/services/notify_service.rb
```diff
@@ -53,8 +53,13 @@ def direct_message?
     @notification.type == :mention && @notification.target_status.direct_visibility?
   end
 
+  def replied_to_status_is_direct_message?
+    replied_to_status = Status.find(@notification.target_status.in_reply_to_id)
```
Gargron: This whole method is just:

`@notification.target_status.thread&.direct_visibility?`

Gargron (changes requested):

app/services/notify_service.rb
renatolond: Ah, I was looking for something like that and couldn't find :)
Changed it :)

Gargron (approved):

PR #6286, opened 2018-01-18
Replace drawer elephant graphic with a vector image

Gargron:
- [x] On compose drawer
- [x] On compose (standalone)
- [x] On modal


Gargron:
cc @akihikodaki

akihikodaki (approved):

PR #6302, opened 2018-01-19
Do not throw away statuses obtained via websocket when API request finishes

ClearlyClaire:
On my small instance, rendering the timelines in the initial API calls can take several seconds, during which the websocket code may forward new toots. Up until the proposed change, the web UI would, on API call completion, discard any toot received via websocket in the meantime.


Gargron:
This does not check for duplicates. If there are duplicate items in the list, it will bug out the rendered DOM in weird ways.

ThibG:
@Gargron `ids` is already filtered to avoid such duplicates?

Gargron (approved):

Gargron:
Right, I forgot I fixed that elsewhere.

PR #6305, opened 2018-01-20
Render LOCAL_DOMAIN as unicode in presentational views

### LINKED ISSUE #6269: does not render international domain names
in case LOCAL_DOMAIN is set as sequence of unicode characters, then web interface does not load, but shows an error, and there is an error in the software output:

```
method=GET path=/ format=html controller=HomeController action=index status=500 error='URI::InvalidURIError: URI must be ascii only "https://\u0569\u0578\u0582\u0569.\u0570\u0561\u0575/"' duration=71.49 view=0.00 db=27.31
```

However, if LOCAL_DOMAIN set as punycode, then everything works, but the account names are also punycoded, and there is non human friendly/readable punycode domain everywhere, instead of human readable IDN unicode.

* * * *

Gargron:
Fix #6269

### DIFF
--- a/app/helpers/instance_helper.rb
+++ b/app/helpers/instance_helper.rb
@@ -6,6 +6,6 @@ module InstanceHelper
   end
 
   def site_hostname
-    Rails.configuration.x.local_domain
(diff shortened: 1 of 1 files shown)


ykzts (approved):
:heart:

norayr:
yaaay!

PR #6309, opened 2018-01-20
Change mailer avatar url

ghost:
* Change avatar URL in mailer to full path

### DIFF
--- a/app/views/notification_mailer/_status.html.haml
+++ b/app/views/notification_mailer/_status.html.haml
@@ -19,7 +19,7 @@
                                 %tbody
                                   %tr
                                     %td{ align: 'left', width: 48 }
-                                      = image_tag url_to_image(status.account.avatar, host: "#{root_url}"), alt:''
+                                      = image_tag full_asset_url(status.account.avatar), alt:''
                                     %td{ align: 'left' }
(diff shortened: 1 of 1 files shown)


app/views/notification_mailer/_status.html.haml
```diff
@@ -19,7 +19,7 @@
                                 %tbody
                                   %tr
                                     %td{ align: 'left', width: 48 }
-                                      = image_tag status.account.avatar
+                                      = image_tag url_to_image(status.account.avatar, host: "#{root_url}"), alt:''
```
Gargron: Use `full_asset_url` method from RoutingHelper instead

Gargron (changes requested):

Gargron (approved):

app/views/notification_mailer/_status.html.haml
```diff
@@ -19,7 +19,7 @@
                                 %tbody
                                   %tr
                                     %td{ align: 'left', width: 48 }
-                                      = image_tag status.account.avatar
+                                      = image_tag full_asset_url(status.account.avatar), alt:''
```
Gargron: Oh damn, it should have been `status.account.avatar.url`, goddammit

PR #6315, opened 2018-01-21
Return 404 instead of 403 if record is not visible

akihikodaki:


Gargron:
What exactly does this fix? (You can DM me if it's not safe to announce)

akihikodaki:
It is not an urgent fix, but for consistency and eliminates potential information leakage.

Unauthorized access for some pages are rejected with 404 even before this change. The pages of the private statuses are such examples. This change makes the behavior consistent throughout this application.

Of course, it is possible to make it always return 403 for unauthorized access, but returning 404 effectively prevents an attacker from knowing the identifier for a private resource is valid or not. It is also a common behavior for many Web sites. So I decided to use 404.

akihikodaki:
I found Mastodon has already similar logics for statuses (`app/controllers/api/v1/statuses_controller.rb`), but it is not covering other objects. This change is the refactored version of the change to apply the logic to all objects.

PR #6318, opened 2018-01-21
fix filter search index being too strict when looking for line breaks

### LINKED ISSUE #6316: regex timeline filters don't handle newlines well
this is an oddity I noticed when trying to block the new 

V
E
R
T
I
C
A
L

meme.

`([\r\n]+[A-Z])+` should do the trick here, but it blocks only *some* instances of the meme. I haven't figured out what's the difference between those that are and aren't caught.

codl:
while local toots do have the XHTML-formatted `<br />`, remote toots seem to have the modern HTML `<br>`

this fixes #6316

### DIFF
--- a/app/javascript/mastodon/reducers/statuses.js
+++ b/app/javascript/mastodon/reducers/statuses.js
@@ -54,7 +54,7 @@ const normalizeStatus = (state, status) => {
     normalStatus.reblog = status.reblog.id;
   }
 
-  const searchContent = [status.spoiler_text, status.content].join('\n\n').replace(/<br \/>/g, '\n').replace(/<\/p><p>/g, '\n\n');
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/reducers/statuses.js
```diff
@@ -54,7 +54,7 @@ const normalizeStatus = (state, status) => {
     normalStatus.reblog = status.reblog.id;
   }
 
-  const searchContent = [status.spoiler_text, status.content].join('\n\n').replace(/<br \/>/g, '\n').replace(/<\/p><p>/g, '\n\n');
+  const searchContent = [status.spoiler_text, status.content].join('\n\n').replace(/<br ?\/?>/g, '\n').replace(/<\/p><p>/g, '\n\n');
```
ykzts: I think it's better to call `\s *` instead of ` \`.

akihikodaki (approved):

MightyPork:
good stuff, thanks

PR #6320, opened 2018-01-21
armenian translations initial commit

norayr:
Hello, web part is completely translated.
As mentioned at https://github.com/tootsuite/documentation/blob/master/Contributing-to-Mastodon/Translating.md#declaring-the-language
we've also edited ```app/helpers/settings_helper.rb``` and ```config/application.rb``` files, however we did not understand if we indeed need to introduce changes to the following files:
```
app/javascript/mastodon/locales/index.js
app/javascript/mastodon/containers/mastodon.js
```

Thank you.


ykzts (approved):

norayr:
thank you! will it work? don't we need to touch those two files i have mentioned?

Gargron:
We do need to edit them, but it can be done separately.

norayr:
so you can do that, thank you. (:

Issue #6331, opened 2018-01-22
Home timeline for returning users not fixed?


Gargron:
In #6251 I did a lot to fix #5634, but I got this report from a user of mastodon.social which runs 2.2.0rc1, which included the fix:

>It said the feed was empty, or something similar. It seems to have kind-of sorted itself out now, I'm seeing toots going forward, but I'm still not seeing anything going backwards past this morning
>
>Except! I can see the two new people I followed when I go backwards in my timeline. Everyone else is not showing up.

https://mastodon.social/@VirtualClint/99395232707153354

I'm kind of at the end of my wits because I have tested locally prior to the fix: Followed a bunch of accounts, made a bunch of posts from them, then removed the home timeline from redis, set my current_sign_in_at and last_sign_in_at values to long ago, and I got back a working home timeline that included everything it needed to. I don't know where and how it goes wrong.

**I would appreciate if another person tried debugging this.**

___

Some pointers:

- Home feeds are stored in redis
- They're pruned for ppl who didn't login for 2 weeks
- UserTrackingConcern is responsible for updating login times at most once a day and queueing up the RegenerationWorker if the 2 week absence is detected
- RegenerationWorker iterates over a SQL version of the home timeline and it's supposed to fill the redis timeline

jennifermf:
in case it's helpful for whoever tackles this, I am having a similar issue: It's only in my notifications column, and it says there are no notifications despite seeing toots mentioning me in my home timeline. Occasionally I will get a notification, but it always disappears on refresh. I always receive desktop/browser notifications. I also get a 500 error every time I load/reload the page. This is happening on toot.cat, after being logged out longer than 2 weeks.

ProgVal:
Did you try logging the status of `account:XXX:regeneration`, to see if it someone gets dropped before the regeneration actually happens?

Gargron:
@ProgVal That key does not determine whether regeneration *happens*, it's only a marker for when it's queued (`RegenerationWorker.perform_async`) and yes there are some unit tests that confirm both that it gets queued and that the marker is set, and finally unset when it's done.

Gargron:
@jennifermf That absolutely cannot be related to this issue. Nothing about notifications depends on your last login time. That sounds more like there's some bugged out notification among the recent ones that breaks the API request.

ProgVal:
@Gargron I meant that it may be removed because of an external factor that only happens on production.

Gargron:
[...]

PR #6333, opened 2018-01-23
Change PrecomputeFeedService method to brute force

Gargron:
See also: #6331. Yet another guess attempt by me.

On one hand I added the timecop gem and rewrote UserTrackingConcern tests to be more readable (in my opinion) and also made sure to check last_sign_in_at values as well, since it's *that* value that's checked for regeneration preconditions.

However, I did not find anything broken or odd.

In a second step I decided to try an alternative approach to home timeline regeneration. Instead of using a SQL query from `Status.as_home_timeline` together with big limits and pagination, I decided to "brute force" the process by simulating the user re-following every person, i.e. executing MergeWorker things synchronously. Synchronously because we need to know when it's "done" to unset the regeneration marker.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ group :test do
   gem 'simplecov', '~> 0.14', require: false
   gem 'webmock', '~> 3.0'
   gem 'parallel_tests', '~> 2.17'
+  gem 'timecop', '~> 0.9'
 end
 
 group :development do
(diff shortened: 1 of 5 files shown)


Gargron:
The real problem & fix is #6341

But I wonder if this is a better regeneration method or not cc @akihikodaki @unarist @abcang

akihikodaki:
I am not sure the second change could really be an improvement. It should be performant for accounts with less follows, but it should be much worse for accounts with many follows.
We may introduce a follow threshold to determine to use the single query method or the merge method, but I would like to know what you intend to improve by the change first.

Gargron:
Alright, I'm closing this since the fix was in another PR and this doesn't seem like an improvement...

PR #6335, opened 2018-01-23
Replace relative URLs in CSS only for Premailer

### LINKED ISSUE #6311: Precompiled assets are hostname dependent
It would be the best if hostname of the [full_asset_url](https://github.com/tootsuite/mastodon/blob/v2.2.0rc1/app/helpers/routing_helper.rb#L15) doesn't depend upon the hostname. What do you folks think referring to assets from the web UI using the relative URI, i.e., `/path/to/asset.js` when `ASSET_HOST` is not defined while using the absolute URI, i.e.,m `https://example.com/path/to/asset.js` for assets included in e-mails? I actually tried to come up with a pull request but found that I don't know  much about webpacker.

At Heroku, precompiled assets maybe used on different apps in a [Pipeline](https://devcenter.heroku.com/articles/pipelines). This makes it possible to test the app in a staging app before deploying to the production. For Mastodon, this stopped working when #6301 was merged. The precompiled assets started to refer to `LOCAL_DOMAIN` environment variable from the staging app (`zundan-mastodon-experiment.herokuapp.com`) where precompile happens and  different from the production app (`mastodon.zunda.ninja`):

54c433a8c53cabce

For now, I could work around this with [reverting](https://github.com/zunda/mastodon/commit/8d7d60353e723c8c186220cc7461a5aca96f2153) the change to `config/webpack/configuration.js` so that webpacker doesn't care about `LOCAL_DOMAIN`. However, I feel this is not the best solution.

Gargron:
Fix #6311

### DIFF
--- a/config/webpack/configuration.js
+++ b/config/webpack/configuration.js
@@ -27,7 +27,7 @@ function formatPublicPath(host = '', path = '') {
 
 const output = {
   path: resolve('public', settings.public_output_path),
-  publicPath: formatPublicPath(env.ASSET_HOST || env.WEB_DOMAIN || env.LOCAL_DOMAIN, settings.public_output_path),
+  publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path),
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


Gargron:
@zunda Please take a look

zunda:
Thanks! Merging this branch fixed the issue on my instance where assets are fetched from the `LOCAL_DOMAIN` for the precompile environment.

PR #6341, opened 2018-01-23
Fix home regeneration not being triggered when logging in

### LINKED ISSUE #6331: Home timeline for returning users not fixed?

Gargron:
Fix #6331

UserTrackingConcern is circumvented by SessionsController#create because it calls warden, which calls the User#update_tracked_fields! method directly. Move returning user logic to that method.

I.e. the home regeneration logic would work perfectly if you still have a valid cookie. If you have to go through the login form, the login form would update the current_sign_in_at values without triggering home regeneration, and subsequent page visits would not pick up the slack because they think the job is done.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -129,7 +129,7 @@ class User < ApplicationRecord
     new_user = !confirmed?
 
     super
-    update_statistics! if new_user
+    prepare_new_user! if new_user
   end
 
   def confirm!
@@ -137,7 +137,12 @@ class User < ApplicationRecord
 
     skip_confirmation!
     save!
-    update_statistics! if new_user
+    prepare_new_user! if new_user
+  end
+
+  def update_tracked_fields!(request)
+    super
+    prepare_returning_user!
   end
 
   def promote!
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -220,9 +225,23 @@ def sanitize_languages
...
 
-  def update_statistics!
+  def prepare_new_user!
     BootstrapTimelineWorker.perform_async(account_id)
     ActivityTracker.increment('activity:accounts:local')
     UserMailer.welcome(self).deliver_later
   end
+
+  def prepare_returning_user!
+    ActivityTracker.record('activity:logins', id)
```
renatolond: Couldn't you track more than once per day if the user logs out and then logs in again?

Gargron: @renatolond It's a weekly hyperloglog. I.e. adding the same id 2000 times won't count as more than 1 per week.

renatolond: Ah, okay, since the previous code was after the return I wasn't sure :)

renatolond (approved):

nightpool:
LGTM 👍

nightpool (approved):

PR #6343, opened 2018-01-24
Improve web api protect

abcang:
The API starting with `/api/web/` is assumed to be used from the Web UI. So, I think it would be better to verify the CSRF token. Also I fixed to return json when validation of CSRF token fails.


ykzts (approved):

Gargron:
I remember merging some changes to this code in the past that changed this behaviour, but I do not remember the specifics (unfortunately). Can anyone confirm this is not a step backwards?

app/javascript/mastodon/actions/push_notifications/registerer.js
```diff
@@ -36,7 +36,7 @@ const subscribe = (registration) =>
 const unsubscribe = ({ registration, subscription }) =>
   subscription ? subscription.unsubscribe().then(() => registration) : registration;
 
-const sendSubscriptionToBackend = (getState, subscription) => {
```
akihikodaki: Why removing `getState`?

abcang: The Web API can use sessions to identify users. So, I think that OAuth token is unnecessary.

akihikodaki: I see. I wonder if @Gargron is mentioning this change, or the forgery protection. This could be in a different pull request in such a case. It would also help when this change is to be reverted as Gargron suspects.

Anyway, it is only when my guess is true. @Gargron's confirmation is needed.

akihikodaki (approved):
It should be merged soon so that we have enough time to evaluate this change before the next release.

Gargron:
So the thing is, I think in a way we have a design error, because app developers would also want to use the Web Push API, not only our own web UI. The Web Push API can be used without browsers, I believe, by running your own autopush server. It's very close to the tusky-api thing I made except with essentially webhooks instead of a running streaming API. Anyway. So my point is that the web push endpoint is out of place there and it should ideally not require a CSRF token, only a good access token...

unarist:
@Gargron

#6223 has 1) enabled CSRF checking on all API endpoints and 2) made WebUI to set both of CSRF token and access_token on all API requests including push subscription. It used `protect_from_forgery with: :null_session` to allow usage of REST APIs without CSRF token. However, it will prints warnings for invalid CSRF tokens, so #6240 disabled it because we don't need to care it on REST APIs (and made push subscription API to raise exception for invalid tokens).

Now, this PR is going to

* Make all `/api/web` endpoints including push subscription API to return error description as JSON. CSRF checking is already working on that API, but it returns only HTTP 422 with empty body.
* Removes unused OAuth access_token from push subscription API calls. This can be done by omitting first argument (getState) from [`api()`](https://github.com/tootsuite/mastodon/blob/v2.3.3/app/javascript/mastodon/api.js#L22-L25) calls. Note that we still use our `api()` function instead of vanilla `axios` instance which we used before #6223. I think this can be done on #6223 too.

So I agree these changes 👍

PR #6344, opened 2018-01-24
Re-Remove translations of default theme

lynlynlynx:
See #5741 
Translations "mastodon" for the default theme are added via Weblate. ~Is there any way to avoid it?~ [issue filed]


akihikodaki (approved):

akihikodaki:
Maybe an issue should be filed.

lynlynlynx:
Thanks @akihikodaki . I filed a issue.

ghost (approved):
Seems valid to me and is indeed an annoyance.

lynlynlynx:
Please merge (or dismiss) this before major release, @Gargron .

nightpool:
I'm not sure i understand, what's the problem with having these translations?

config/locales/sr.yml
```diff
@@ -695,8 +695,6 @@ sr:
 
       <p>Originally adapted from the <a href="https://github.com/discourse/discourse">Discourse privacy policy</a>.</p>
     title: Услови коришћења и политика приватности инстанце %{instance}
-  themes:
-    default: Мастодонт
```
nightpool: this was added intentionally by a human https://github.com/tootsuite/mastodon/pull/6133/files

config/locales/sr-Latn.yml
```diff
@@ -695,8 +695,6 @@ sr-Latn:
 
       <p>Originally adapted from the <a href="https://github.com/discourse/discourse">Discourse privacy policy</a>.</p>
     title: Uslovi korišćenja i politika privatnosti instance %{instance}
-  themes:
-    default: Mastodont
```
nightpool: this was added intentionally by a human as far as I can tell

config/locales/ru.yml
```diff
@@ -667,8 +667,6 @@ ru:
       <p>This document is CC-BY-SA. It was last updated May 31, 2013.</p>
       <p>Originally adapted from the <a href="https://github.com/discourse/discourse">Discourse privacy policy</a>.</p>
     title: Условия обслуживания и политика конфиденциальности %{instance}
-  themes:
-    default: Mastodon
```
nightpool: this was added intentionally by a human https://github.com/tootsuite/mastodon/pull/6227

akihikodaki:
@nightpool #5741 may be informative for you.

It is notable that they are added by humans. It should be noted well if we keep these translations removed.

Gargron:
The problem is that Weblate can't read i18n configuration from i18n-tasks, where the list of translations to be ignored it stored. Also, I don't know where an equivalent function is in Weblate UI. So, people will always keep re-submitting these...

jeroenpraat:
@nightpool The problem with having these translations is because of themes.

If you can read [here](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Customizing.md#multiple-themes) you need to add something to config/locales/en.yml if you want a nice description, but when your default theme is not called 'Mastodon' you need to remove the respective translations from all locale files or else you need to manually remove those strings during every upgrading.

Gargron:
Until Weblate can learn to ignore certain keys, people will keep re-submitting these translations, and you'll have to keep re-removing them. So I think that this PR will only add churn to the code, not worth it.

PR #6358, opened 2018-01-25
Rescue when there's no extension in the remotable

renatolond:
Sometimes the remotable is pointing to a directory with no file extension. Maybe it should not be expecting to identify based on extensions to begin with, but since it's the case, it should be ready for it.

This caused errors on my sidekiq of the type:
```
WARN: TypeError: no implicit conversion of nil into String
WARN: /home/mastodon/live/app/models/concerns/remotable.rb:31:in `extname'
```

### DIFF
--- a/app/models/concerns/remotable.rb
+++ b/app/models/concerns/remotable.rb
@@ -29,7 +29,7 @@ module Remotable
           filename = matches.nil? ? parsed_url.path.split('/').last : matches[1]
           basename = SecureRandom.hex(8)
           begin
-            extname  = File.extname(filename)
+            extname = File.extname(filename)
           rescue TypeError
             return
           end
--- a/app/models/concerns/remotable.rb
+++ b/app/models/concerns/remotable.rb
@@ -28,10 +28,10 @@ module Remotable
           matches  = response.headers['content-disposition']&.match(/filename="([^"]*)"/)
           filename = matches.nil? ? parsed_url.path.split('/').last : matches[1]
           basename = SecureRandom.hex(8)
-          extname = unless filename.nil?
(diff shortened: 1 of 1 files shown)


app/models/concerns/remotable.rb
```diff
@@ -28,7 +28,11 @@ module Remotable
           matches  = response.headers['content-disposition']&.match(/filename="([^"]*)"/)
           filename = matches.nil? ? parsed_url.path.split('/').last : matches[1]
           basename = SecureRandom.hex(8)
-          extname  = File.extname(filename)
+          begin
+            extname = File.extname(filename)
+          rescue TypeError
+            return
+          end
```
ykzts:
```ruby
extname = File.extname(filename) unless filename.nil?
```

akihikodaki (dismissed):
The current behavior is not preferred as you say, but I do not think yours is either.
The file extension is optional on HTTP. Such lack of extension should be tolerated.

renatolond:
Hi! Sorry, I've been away on holidays and haven't had the time to pick this back up. Indeed, the extension is optional, but the way the remotable is written right now, I wasn't sure if ignoring the extension would be okay (I don't know the repercussions of the file name without the extname, is it expected up ahead?)

In any case, I'll try to pick this back up soon and make the requested changes.

renatolond:
Finally got around to make the changes @ykzts had requested :)
I actually changed a little more because I think the concat below would fail when extname was not defined.

Gargron (approved):

Gargron:
>Do not use unless with else. Rewrite these with the positive case first.

renatolond:
Inverted!

PR #6369, opened 2018-01-28
Isolate internal services from external networks in Docker configuration

akihikodaki:
The database and Redis do not need external connections, so isolate them and prevent unauthorized access.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,8 @@ services:
   db:
     restart: always
     image: postgres:9.6-alpine
+    networks:
+      - internal_network
 ### Uncomment to enable DB persistance
 #    volumes:
 #      - ./postgres:/var/lib/postgresql/data
@@ -11,6 +13,8 @@ services:
   redis:
     restart: always
     image: redis:4.0-alpine
+    networks:
+      - internal_network
 ### Uncomment to enable REDIS persistance
 #    volumes:
 #      - ./redis:/data
@@ -21,6 +25,9 @@ services:
     restart: always
     env_file: .env.production
     command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    networks:
+      - external_network
+      - internal_network
     ports:
       - "3000:3000"
     depends_on:
@@ -37,6 +44,9 @@ services:
     restart: always
     env_file: .env.production
     command: npm run start
+    networks:
+      - external_network
+      - internal_network
     ports:
       - "4000:4000"
     depends_on:
@@ -52,5 +62,13 @@ services:
     depends_on:
       - db
       - redis
+    networks:
+      - external_network
+      - internal_network
     volumes:
       - ./public/system:/mastodon/public/system
+
+networks:
+  external_network:
+  internal_network:
+    internal: true


akihikodaki:
I found this change is quite a useful when customizing the network configuration, so I added "deployment" tag.

Apart from this change, I was working on IPv6 support on Docker, hearing that mstdn.maud.io is planning to implement but has not achieved yet. We have to specify an external network with IPv6 enabled for the purpose. After applying this change, it will be possible by adding a line `COMPOSE_FILE=docker-compose.yml:docker-compose.network.yml` and creating a new file named `docker-compose.network.yml` with the following content:
```YAML
version: '3'
networks:
  external:
    external:
      name: mastodon
```

Gargron:
>The database and Redis do not need external connections, so isolate them and prevent unauthorized access.

Redis and Postgres containers do not expose their ports to the host even in current configuration, they are "linked" to the web/sidekiq/streaming containers, so if I understand correctly, this is not a *security* improvement.

How much better/easier does this make deployments? Will it mess with anyone's current deployments? I would like to hear from other admins before I approve this. E.g. @unarist @Technowix (I am not very sure who else uses Docker... Sorry...)

akihikodaki:
> > The database and Redis do not need external connections, so isolate them and prevent unauthorized access.
>
> Redis and Postgres containers do not expose their ports to the host even in current configuration, they are "linked" to the web/sidekiq/streaming containers, so if I understand correctly, this is not a security improvement.

They do. They are just not mapped to a loop back address, and are accessible via the IP address of each containers.

> How much better/easier does this make deployments? Will it mess with anyone's current deployments? I would like to hear from other admins before I approve this. E.g. @unarist @Technowix (I am not very sure who else uses Docker... Sorry...)

mstdn.maud.io, maintained by @lindwurm (and @unarist I guess?) is on Docker. I actually wrote the procedure to configure IPv6 I mentioned earlier because he is planning to serve it via IPv6 (though it is not in priority.) However, this change will not benefit the instance.

Indeed the gain of this change in terms of IPv6 configuration is that it makes additional modifications for `docker-compose.yml` unnecessary, but the instance already has some modifications. This change saves a few lines and that's it for the instance.

Gargron:
```
  external:
    external: true
  internal:
    internal: true
```

The naming of the networks is really confusing. It took a while for me to understand what I was reading. Is this a convention that is widely used within docker-compose projects?

akihikodaki:
> The naming of the networks is really confusing. It took a while for me to understand what I was reading. Is this a convention that is widely used within docker-compose projects?

No, I invented them. Should I add `_network` suffix or sort of?

Gargron:
Yes, that would be a good solution.

akihikodaki:
a4372c488ced7eda26017e32e7a879b4fedb05ed adds `_network` suffix.

Gargron (approved):

ThibG:
This seems to break DNS-based addressing of containers within the network, e.g., the “web” container cannot find the “db” container, etc.
see https://github.com/kata-containers/runtime/issues/175

PR #6376, opened 2018-01-29
Show the result of post in post intent page

akihikodaki:


akihikodaki:
046ffb0b41befada830422a397a4100e6cdfeb92
screenshot-2018-1-30 mastodon dev

app/javascript/mastodon/containers/compose_container.js
```diff
@@ -29,7 +30,9 @@ export default class TimelineContainer extends React.PureComponent {
     return (
       <IntlProvider locale={locale} messages={messages}>
         <Provider store={store}>
-          <Compose />
+          <BrowserRouter basename='/web' forceRefresh>
+            <Compose />
+          </BrowserRouter>
```
ThibG: What is this for?

app/javascript/mastodon/reducers/compose.js
```diff
@@ -54,6 +54,7 @@ const initialState = ImmutableMap({
   default_sensitive: false,
   resetFileKey: Math.floor((Math.random() * 0x10000)),
   idempotencyKey: null,
+  last: null,
```
ThibG: I'm not sure about having this information here, it seems to be only ever useful to the particular case of the web intent page… there ought to be a better way, although I'm not familiar enough with React to know better :/

app/javascript/mastodon/containers/compose_container.js
akihikodaki: For `Link` of React Router DOM. It does no routing, but it just navigates to the linked page.

app/javascript/mastodon/reducers/compose.js
akihikodaki: I don't think there is a better way for this particular case. There are several options such as to have different Redux store or to utilize callbacks and React state. However they bring too much complexity for just one variable.

PR #6381, opened 2018-01-30
Remove wave from list drawer

akihikodaki:


ykzts (approved):

Gargron:
What does this change do?

akihikodaki:
Before: 
screenshot-2018-2-2 mastodon dev 1
After: 
screenshot-2018-2-2 mastodon dev

The wave is not intended to be shown in the list modal, and it can be noisy when an account description is overlapping. So this is a _bug_.

PR #6382, opened 2018-01-30
Unify links container implementation in about pages

akihikodaki:
They were redundant, and also had a inconsistency; the button for "other instances" had an icon for the external link in "more" page, but it didn't in the other pages.

This unifies the implementation, and the external link icon is now shown in all the about pages.


ykzts (approved):
good :heart:

PR #6389, opened 2018-01-30
delete outer join in status model

deflis:
refactor status model

### before
`
SELECT  "statuses"."id", "statuses"."updated_at" FROM "statuses" LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id" WHERE "statuses"."visibility" = $1 AND (statuses.reblog_of_id IS NULL) AND (statuses.reply = FALSE OR statuses.in_reply_to_account_id = statuses.account_id) AND (1=1) AND "accounts"."silenced" = $2 ORDER BY "statuses"."id" DESC LIMIT $3
`

### after
`
 SELECT  "statuses"."id", "statuses"."updated_at" FROM "statuses" WHERE "statuses"."visibility" = $1 AND (statuses.reblog_of_id IS NULL) AND (statuses.reply = FALSE OR statuses.in_reply_to_account_id = statuses.account_id) AND (1=1) AND "statuses"."account_id" IN (SELECT "accounts"."id" FROM "accounts" WHERE "accounts"."silenced" = $2) ORDER BY "statuses"."id" DESC LIMIT $3
`

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -111,6 +111,7 @@ class Account < ApplicationRecord
   scope :expiring, ->(time) { remote.where.not(subscription_expires_at: nil).where('subscription_expires_at < ?', time) }
   scope :partitioned, -> { order('row_number() over (partition by domain)') }
   scope :silenced, -> { where(silenced: true) }
+  scope :not_silenced, -> { where(silenced: false) }
   scope :suspended, -> { where(suspended: true) }
   scope :recent, -> { reorder(id: :desc) }
   scope :alphabetic, -> { order(domain: :asc, username: :asc) }
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -69,10 +69,10 @@ class Status < ApplicationRecord
   scope :without_reblogs, -> { where('statuses.reblog_of_id IS NULL') }
   scope :with_public_visibility, -> { where(visibility: :public) }
   scope :tagged_with, ->(tag) { joins(:statuses_tags).where(statuses_tags: { tag_id: tag }) }
-  scope :excluding_silenced_accounts, -> { left_outer_joins(:account).where(accounts: { silenced: false }) }
-  scope :including_silenced_accounts, -> { left_outer_joins(:account).where(accounts: { silenced: true }) }
+  scope :excluding_silenced_accounts, -> { where(account_id: Account.select(:id).not_silenced) }
+  scope :including_silenced_accounts, -> { where(account_id: Account.select(:id).silenced) }
   scope :not_excluded_by_account, ->(account) { where.not(account_id: account.excluded_from_timeline_account_ids) }
-  scope :not_domain_blocked_by_account, ->(account) { account.excluded_from_timeline_domains.blank? ? left_outer_joins(:account) : left_outer_joins(:account).where('accounts.domain IS NULL OR accounts.domain NOT IN (?)', account.excluded_from_timeline_domains) }
+  scope :not_domain_blocked_by_account, ->(account) { where.not(account_id: Account.select(:id).where(domain: account.excluded_from_timeline_domains)) if account.excluded_from_timeline_domains.present? }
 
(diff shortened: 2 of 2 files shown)


takayamaki:
## before
```
mastodon-production=> explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses" LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id" WHERE "statuses"."visibility" = 0 AND (statuses.reblog_of_id IS NULL) AND (statuses.reply = FALSE OR statuses.in_reply_to_account_id = statuses.account_id) AND (1=1) AND "accounts"."silenced" = 'f' ORDER BY "statuses"."id" DESC LIMIT 40;
                                                                        QUERY PLAN                                                                         
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.72..360.64 rows=40 width=16) (actual time=0.025..0.302 rows=40 loops=1)
   ->  Nested Loop  (cost=0.72..24726273.39 rows=2747993 width=16) (actual time=0.024..0.284 rows=40 loops=1)
         ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.43..9921961.91 rows=3900370 width=24) (actual time=0.014..0.123 rows=41 loops=1)
               Filter: ((reblog_of_id IS NULL) AND ((NOT reply) OR (in_reply_to_account_id = account_id)) AND (visibility = 0))
               Rows Removed by Filter: 27
         ->  Index Scan using index_accounts_on_id on accounts  (cost=0.29..3.79 rows=1 width=8) (actual time=0.003..0.003 rows=1 loops=41)
               Filter: (NOT silenced)
               Rows Removed by Filter: 0
 Planning time: 0.325 ms
 Execution time: 0.338 ms
(11 rows)

mastodon-production=>
```

## after
```
mastodon-production=> explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses" WHERE "statuses"."visibility" = 0 AND (statuses.reblog_of_id IS NULL) AND (statuses.reply = FALSE OR statuses.in_reply_to_account_id = statuses.account_id) AND (1=1) AND "statuses"."account_id" IN (SELECT "accounts"."id" FROM "accounts" WHERE "accounts"."silenced" = 'f' ) ORDER BY "statuses"."id" DESC LIMIT 40;
                                                                        QUERY PLAN                                                                         
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.72..254.30 rows=40 width=16) (actual time=0.022..0.275 rows=40 loops=1)
   ->  Nested Loop Semi Join  (cost=0.72..24726273.39 rows=3900370 width=16) (actual time=0.022..0.261 rows=40 loops=1)
         ->  Index Scan Backward using statuses_pkey on statuses  (cost=0.43..9921961.91 rows=3900370 width=24) (actual time=0.013..0.119 rows=41 loops=1)
               Filter: ((reblog_of_id IS NULL) AND ((NOT reply) OR (in_reply_to_account_id = account_id)) AND (visibility = 0))
               Rows Removed by Filter: 31
         ->  Index Scan using index_accounts_on_id on accounts  (cost=0.29..3.79 rows=1 width=8) (actual time=0.002..0.002 rows=1 loops=41)
               Filter: (NOT silenced)
               Rows Removed by Filter: 0
 Planning time: 0.308 ms
 Execution time: 0.307 ms
(11 rows)

mastodon-production=>
```

Gargron:
Is this a significant difference or random fluctuation? 0.020 ms is not much.

deflis:
This is fix of bad code.
`OUTER JOIN` is bad idea. This is bad technique of such a case.
So I changed query to subquery using `IN` clause.
This fixes the code to be more readable and easier to maintain.

Gargron:
>OUTER JOIN is bad idea. This is bad technique of such a case.

Can you elaborate? The query planner seems to do the same thing in both cases, and i cannot find any conclusive search results for "OUTER JOIN is a bad idea in postgres". I am not against using IN instead of OUTER JOIN but I don't want to change code without a good reason.

akihikodaki:
> This fixes the code to be more readable and easier to maintain.

Not for me. I assume such relational algebra is more familiar for relational database users.

I think this pull request should be closed, and reopened later if any improvement in this change found.

PR #6409, opened 2018-02-02
Fix saving of oEmbed image

abcang:
From version 5.2 of paperclip, URI adapters are no longer loaded by default, so oEmbed's image saving failed.

https://github.com/thoughtbot/paperclip/pull/2435

It was fixed by assigning it to `image_remote_url` in the same way as `attempt_opengraph`.

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -91,13 +91,13 @@ class FetchLinkCardService < BaseService
 
     case @card.type
     when 'link'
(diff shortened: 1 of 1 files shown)


akihikodaki:
it does not use URI adapter as far as I know. See `Remotable` for details.

I guess you are concerning an error report of Pawoo. It is caused by its own extension and irrelevant for the upstream.

akihikodaki (approved):
I reviewed the change and understood what you meant. I approve this change.
~~I think this should be released as soon as possible (2.3.1).~~

PR #6410, opened 2018-02-02
Raise Mastodon::ValidationError when host for HTTP request is private

akihikodaki:


app/lib/request.rb
```diff
@@ -41,6 +42,12 @@ def headers
 
   private
 
+  def sanitize_url
+    if PrivateAddressCheck.resolves_to_private_address? @url.host
```
Gargron: `raise ... if ...` in one line

Gargron: And since the method is supposed to raise an exception, I would call it `sanitize_url!` with exclamation mark.

akihikodaki:
7f9df5f1c0a0542077b2d81f4701e65224d78bf0 addresses https://github.com/tootsuite/mastodon/pull/6410#pullrequestreview-93603797.

akihikodaki:
0302d969e501883d0088eb8d09c7380f1cc000ab fixes the spec.

unarist:
Is this restriction also applied to federation? I often test federation between instances on the local network, so I want to disable it on development environment.

akihikodaki:
23bef157ddba3a89cd54148af455a7458c72c804 disables the validation on development environment.

Gargron (approved):

saper:
What problem does it solve?

saper:
I think this change is incorrect. `getaddress()` returns only the first address given by the name resolution mechanism of the operating system. When opening the socket, Ruby takes all addresses give by the name resolution and iterates all of them to get the matching address family and to find out the address that finally responds (this bypasses the operating system logic to do so, which may breaks things, but this is another story).

Therefore it is trivial for the attacker to execute its attack despite this "protection".

Second, those things are policy settings and belong to the configuration settings (well documented). This setting is questionable as the default.

Setting up firewall may be in the end easier, safer and working better!

akihikodaki:
It gives the resolved address to Ruby's `TCPSocket` and therefore it is impossible to circumvent the protection in my opinion.

It is true that firewall may still be necessary and a corresponding documentation change is pending. https://github.com/tootsuite/documentation/pull/531

Gargron:
Follow-up: This change is not applied to the Goldfinger gem. We should change the Goldfinger gem so it can accept a HTTP instance as an argument.

PR #6415, opened 2018-02-02
Make PAM gem optional, allow configuration over environment

Gargron:
Follow-up to #5303:

- Only install PAM gem if PAM_ENABLED=true
- Allow configuring PAM over environment variables


devkral:
I am not sure if it is necessary. As I wrote, one of the use cases is:
local users can use their password for mastodon. Would be a cool local user setup.

Gargron:
>I am not sure if it is necessary

Bundler install will fail if PAM headers are not installed on the system. There is no reason to create this requirement when most systems do not use PAM currently and will not use it in the future.

Gargron:
The build fails because I forgot to make `devise :pam_authenticable` optional as well. Currently not sure how to accomplish that.

config/initializers/devise.rb
```diff
@@ -315,22 +315,13 @@ def valid?
...
-  # name of the pam service used for checking if an user can register
-  # pam "account" section is evaluated
-  # nil for allowing registration of pam names (not recommended)
-  config.pam_controlled_service = "rpam"
+  if ENV['PAM_ENABLED'] == 'true'
+    config.pam_authentication     = true
+    config.usernamefield          = nil
+    config.emailfield             = 'email'
+    config.check_at_sign          = true
+    config.pam_default_suffix     = ENV.fetch('PAM_DEFAULT_SUFFIX') { 'pam' }
```
devkral: It would be nice to have comments what the options can archive

devkral:
to your question how to make devise :pam_authenticable optional
test for Devise.pam_authentication. e.g. like this:
devise :pam_authenticable if Devise.pam_authentication

devkral:
or you can replace Device.pam_authentication by use_pam?

devkral:
Hope it helps you

Gargron:
@devkral Have I brought over the comments for the variables correctly? To be honest they were not written very clearly, at least for someone who has never worked with PAM before, I wasn't exactly sure what they meant.

ykzts (approved):

devkral:
@Gargon yes you have. Only one minor mistake: 
# Name of the pam service (pam "auth" section is evaluated)
should be:
# Name of the pam service (pam "auth" and "account" section is evaluated)

PR #6421, opened 2018-02-03
Upgrade Vagrant box to Xenial

northerner:
Upgrading the dev vagrant box to Xenial allows support for newer Redis versions, fixing this issue: https://github.com/tootsuite/mastodon/issues/6399

This update also also adds the libpam0g-dev dependency to the Vagrant box required by the PAM gem to build native extensions: https://github.com/tootsuite/mastodon/commit/04fef7b8886bb78f3473e143894a521ca578f1db

As of this commit the Vagrant box will build correctly with `vagrant up`, this lets the site run and all tests pass.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -79,7 +79,7 @@ VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
-  config.vm.box = "ubuntu/trusty64"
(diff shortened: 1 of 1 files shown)


Gargron:
>This update also also adds the libpam0g-dev dependency to the Vagrant box required by the PAM gem to build native extensions: 04fef7b

This is a recent change and will likely be fixed by #6415

Gargron (approved):

PR #6423, opened 2018-02-03
Full-text search for authorized statuses

### LINKED ISSUE #1152: search API call not returning any statuses, ever.
'statuses' of the content search API call (/api/v1/search) never seems to actually return any statuses.
Hashtags and accounts are returned without problems.

Steps to reproduce: https://mastodon.social/api/v1/search/?q=e

Since "e" is the most frequently used character, we should be getting pretty much all statuses here, shouldn't we?

* * * *

### LINKED ISSUE #4293: Full text search of my own posts
I don't usually realize that something is important when I'm in the fog of it.

Full text search of my own posts will help me find something I've posted about in the past.

Restricting the search to my own posts should limit abuse potential while still adding a huge amount of value.

Thoughts?

### LINKED ISSUE #5880: Enhancement: Full-text search scoped to users' past interactions
As a Mastodon user, I'd like to be able to find toots that I've run across in the past, using full-text search.

I know that full-text search across the fediverse is [intentionally unsupported](https://github.com/tootsuite/mastodon/issues/2637) for anti-harassment reasons. While that's a problem for *discovery* (of people and content) it's also a problem for *retrieval*. We can fix the latter separately!

I would like the instance server to maintain a search index of posts that the users have directly interacted with (posted, replied to, favorited, boosted) or been shown (posts that have ever ended up in their Home or Notifications view). This would not enable harassment.

Searches of this index would be scoped to the user—just because something is indexed on my behalf doesn't mean another user on the instance would necessarily be able to retrieve it. I think this still allows deduplication in the index, depending on the search technology.

(Also see: https://github.com/tootsuite/mastodon/issues/4293 for just searching own posts.)

* * * *

Gargron:
- Search API will return statuses that match the query
- Only for logged in users
- Only if you are author of the status,
- Or you were mentioned in it
- Or you favourited or reblogged it
- Configuration over `ES_ENABLED`, `ES_HOST`, `ES_PORT`, `ES_PREFIX`
- Run `rails chewy:deploy` to create & populate index

Fix #5880
Fix #4293
Fix #1152

image


LienRag:
Nice, thanks!
Note that it still doesn't allow search in all statuses.
I understand the privacy concern behind it, but it's still a problem.
Could there be a way for each mastodon member and/or instance to allow full text search either to everybody or to selected users/instances?
And to authorize/forbid that for specific toots (i mean, define a general policy and then be able to exclude specific toots from that general policy)?

Gargron:
We don't want to implement search for finding *new* stuff. We want to implement search for *things you've already seen*.

LienRag:
Yes, I get it.
But still I very often (and am certainly not alone in that) want to search for toots that I have already seen and can't find, either in the Home, Local or Federated Timelines. And usually (not always, so this addition is already a very interesting improvement of Mastodon) I am not the author of the toot, nor mentioned in it nor have I retooted it.
Main use case (but not the only one): someone toots a link to an article, I open the link and leave it to read it later, and then when I read it I want to make remarks about it, but can't find the original toot in order to reply to it.
I really don't think that I'm the only one which such a workflow...

LienRag:
Note that public toots are already indexed by Google and other search tools, so there's no real gain in privacy by making them unsearchable on Mastodon itself...

rainbowmultiverse:
Most of the abuse on Twitter this constraint is meant to resolve happens in a short window and directly from a retweet (usually a quote retweet these days) since it tends to resolve around the Current Hot Discourse. Every step needed to fling abuse at someone cuts it down by a wide margin.

Also, if it's that easy to find by Google search, you can find it that way.

gled-rs:
may we get an idea of why adding a dependancy to ES for that ? 

Pgsql includes a very powerful FTS engine, that would allow less dependancy with the project.

As I already offered on mastodon, I can provide some feedback and help for the implementation.

Gargron:
@gled-rs 

- Performance affecting critical functions of Mastodon: If ES is slow, it only affects search queries. If postgres is slowed down by a search query, it slows down everything
- Storage requirements separation: ES is optional, can be on a different disk, you can purge the entire index and it won't break anything; a growing postgres database is more difficult to partition this way
- The Chewy gem is really simple to use and has saved a lot of work in this PR. It's an ES-specific gem.
- There is a lot of documentation and resources written about using ES for search results, how to improve them, etc. Postgres FTS is simple in the most basic use (we use it for username autocompletes!), but going deeper is more complicated (tokenization, different languages stemming/stopwords, boosting)

I think making this whole feature optional should soften your concerns about the dependency choice, right?

gled-rs:
Actually it's quite the opposite.

But first, I welcome this PR as a good move in usability for the end user.

My reasoning is the following:
- a user will probably not understand why search is working on that instance, but not on this other small one, and will tend to stick with the one that provides him the most features.
- most often on smaller instances, if ES is deployed, ES will be deployed alongside the rest, so 1 server for everything. The performance argument is then void, as pgsql will be impacted by what ES is doing, but with less visibility ( why is there less I/O available ? I don't understand why my query takes so long there's nothing else running ? )
- if the point above is not true, then we are creating tiers of instances based on the financial capacity of the admin. Those with less funds will not be able to host a feature complete instance ( I suppose you saw the outcry about that on Mastodon, which was the trigger for the message above ), back to point 1.
- Regarding storage/disk space on pgsql, you can always shard/replicate/create different DBs for different purposes. I guess it's more a seasoned DBA subject than the purpose of this post though, but want that to also be said, as PG is amazingly easy to scale with a little bit of tuning/planning.

I can see your arguments about how easier it is to use ES, especially with the Chewy gem masking all the complexity of FTS, that being said, I can help to implement FTS in pgsql if you want, masking a bit of the complexity to you ?

Regarding now my testings on mastodon.host ( which has FTS as you probably know, ~ 50Gb DB ):
- Speed of querying the m.h DB with a simple two word query: Sphinx > ES ~= Pgsql.
- Disk space: Pgsql < Sphinx < ES.

Once you have your tsvector index, FTS is really fast on PG, and takes not that much space ( we are talking about an index size of less than 5gb for m.h, so way less than what other indexes cumulated are taking ).

PS: Do not get me wrong, ES is a fantastic tool too, but I think is way overkill for what you are trying to achieve there, and probably a strong barrier to entry for most.

nightpool:
@gled-rs the main issue we ran into is that account insertions or updates take very very long amounts of time on PG with full text indexes. whereas ES updates are done in the background, so they don't need to update their index synchronously. Is there a way to fix this and not tank the performance of table updates for our most frequently-updated table?

gled-rs:
It is a bit late here, probably the reason I am a bit confused ( and I'm on my phone ). I fail to see the link between account insertion and creating an index in the statuses table to search your own toot... Are you talking about implementing that also for account search ?Can you show me the DB schema you were using ? There is probably room for improvement if it was such a strain. Were you using triggers to index or doing it in your insert query ? Stored procedures were used ?I can provide a few metrics, we did not saw that issue on m.h when I activated our FTS ( at the moment our main DB issue is a way too frequent vacuum/analyze cycle on a few tables ). I would happily review and provide help to go further, as you see fit.PS: if you want to touch base on that on xmpp for an easier chat, don't hesitate, gled@mastodon.host On Feb 5, 2018 7:52 PM, nightpool <notifications@github.com> wrote:@gled-rs the main issue we ran into is that account insertions or updates take very very long amounts of time on PG with full text indexes. whereas ES updates are done in the background, so they don't need to update their index synchronously. Is there a way to fix this and not tank the performance of table updates for our most frequently-updated table?

gled-rs:
Should I post here the full how to activate FTS in Pgsql ?

It will be then trivial for anyone to replicate and do whatever they want with FTS on their server though, instead of a big patch with limits in different places...

Gargron:
@gled-rs Account search is already implemented (see username autocompletes for example) using FTS, you can look at the schema to see how it's done. The full-text index is touched on UPDATEs even when the indexed column is not changed.

ES offers stemming/stopwords for various languages (even though I am not sure how to take advantage over those dynamically depending on toot language yet), and it's something [Discord also picked for their own search](https://blog.discordapp.com/how-discord-indexes-billions-of-messages-e3d5e9be866f). Many of their reasons are not relevant to us, such as clustering (although a large instance like Pawoo or friends.nico may wish for this to be possible), but this one is also important:

>Engineers on the team had more experience working with Elasticsearch

Pre-1.0 Mastodon had a dependency on Neo4J for a short time. It was then completely removed. It's easy to undo this feature or replace it afterwards, since neither the DB schema nor the API is changed in any way.

Gargron:
We need to wait until toptal/chewy#630 makes it into a new release of Chewy before this is truly production-ready so admins don't get weird errors in rake tasks. For reference, the actual index creation/importing works currently, but index creation is supposed to be recorded in a different structure to allow incremental changes afterward and that's the part that currently throws errors.

gled-rs:
Got it.

I'll probably release then the pgsql FTS version on my fork without witholding a few necessary steps, to allow everyone to bypass the Es implementation, and also allow resource limited admins to still have the feature.

beatrix-bitrot (approved):

PR #6424, opened 2018-02-04
Add dependency in Dockerfile about pam.

yukimochi:
Required by the PAM gem to build native extensions: 04fef7b .


Gargron:
#6415

PR #6425, opened 2018-02-04
CAS + SAML authentication feature

Gargron:
Continuation of #3148 since I can't write to someone else's branch.

image


ykzts (approved):
I think the code is okay. For checking CAS and SAML, it is necessary to check with an active instance.

PR #6431, opened 2018-02-05
match hashtag regex in js client with server

### LINKED ISSUE #6414: hashtag warning appears for things that aren't hashtags
2018-02-02t22-44-51-0o0o

there's a discrepancy between what the server and client consider to be a hashtag. this makes the client display its "hashtags don't work in non-public toots" warning even if the status doesn't have any valid hashtags in it

some examples are `#1 dad`, `#!/bin/bash`

* * * *

northerner:
This fixes #6414

the slight mismatch in hashtag regex between js and ruby was causing
hashtag warning to be displayed for unlisted tweets when an invalid
hashtag was entered

exact version of ruby regex not possible in JS as POSIX bracket
expressions are not supported, this version approximates and doesn't
give same unicode support

Ruby:
[[:word:]_]*[[:alpha:]_][[:word:]_]*

JS:
\w*[a-zA-Z]\w*

### DIFF
--- a/app/javascript/mastodon/features/compose/containers/warning_container.js
+++ b/app/javascript/mastodon/features/compose/containers/warning_container.js
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 import { me } from '../../../initial_state';
 
-const APPROX_HASHTAG_RE = /(?:^|[^\/\)\w])#(\S+)/i;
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Artoria2e5:
For Unicode support (in 2020 browsers?), the ES2018 regex extension for Unicode properties can be used. `[\p{Alphabetic}\p{Mark}\p{Decimal_Number}\p{Connector_Punctuation}\p{Join_Control}]` can be used to replace `\w`, and `\p{Alphabetic}` to replace `[a-zA-Z]`.

PR #6446, opened 2018-02-08
Perform request spec

tooooooooomy:
Please check it.

Thanks!

### DIFF
--- a/spec/services/fetch_atom_service_spec.rb
+++ b/spec/services/fetch_atom_service_spec.rb
@@ -1,4 +1,24 @@
 require 'rails_helper'
 
 RSpec.describe FetchAtomService do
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
There is no immediate harm in these tests but you should know that unit tests should not try to look into private methods or internal variables. The unit test must confirm the public API of the class works, i.e. how other code will use it. The internal implementation of the class is subject to change (refactoring). The perfect refactoring is when you change the code but don't have to touch the tests. With tests like these, they would also have to be modified if the internal implementation changes in the future.

kazu9su:
@Gargron Thank you for pointing them out.
I understood your testing style so I will check the behavior of public methods from next time.

PR #6451, opened 2018-02-09
Interactive `rake mastodon:setup` task

Gargron:
Demo: <https://asciinema.org/a/KWYsfWQYxwupT4jp7YcmsYUge>

- Generates all secrets automatically
- Reasonable defaults for docker setups
- Reasonable defaults for non-docker setups
- Tests postgres and redis configurations
- Guides through Amazon S3, Wasabi or Minio setups for paperclip
- Tests SMTP configuration with a test e-mail
- Executes db:setup and assets:precompile in the end
- Option to create the admin user at the end


lib/tasks/mastodon.rake
```diff
@@ -4,6 +4,258 @@ require 'optparse'
...
+  task :setup do
+    prompt = TTY::Prompt.new
+    env    = {}
+
+    begin
+      prompt.say('Your instance is identified by its domain name. Changing it afterward will break things.')
+      env['LOCAL_DOMAIN'] = prompt.ask('Domain name:') do |q|
+        q.required true
+        q.modify :strip
+        q.validate /\A[a-z0-9\.\-]+\z/i
```
ProgVal: What about internationalized domain name?

lib/tasks/mastodon.rake
```diff
@@ -4,6 +4,258 @@ require 'optparse'
...
+        q.modify :strip
+      end
+
+      prompt.say "\n"
+      prompt.say 'This configuration will be written to .env.production'
+
+      if prompt.yes?('Save configuration?')
+        cmd = TTY::Command.new(printer: :quiet)
+
+        File.write(Rails.root.join('.env.production'), "# Generated with mastodon:setup on #{Time.now.utc}\n\n" + env.each_pair.map { |key, value| "#{key}=#{value}" }.join("\n") + "\n")
```
ProgVal: This does not include comments, and I think it puts keys in a random order. This would not make it easy to edit the file afterward.

ProgVal:
Idea: check connection to other services (postgresql, redis, postfix) right after questions are asked, and re-ask the questions if the connection fails.

lib/tasks/mastodon.rake
Gargron: Read the error message right below this line. This env variable needs ascii only, no unicode allowed

ProgVal: Oops, sorry.
Wouldn't it make sense to have a punycode encoder, though?

lib/tasks/mastodon.rake
Gargron: That doesn't really matter. Nobody canceled `.env.production.sample`. Order of the params does not matter at all for the machine.

ProgVal:
This looks great, thanks!

ykzts (approved):

PR #6459, opened 2018-02-10
Fix password recovery

ClearlyClaire:
Password recovery has been broken since 04fef7b8886bb78f3473e143894a521ca578f1db.
There are three causes for that:
1. Incorrect indentation
2. `use_pam?` instead of `!use_pam?` (the logic in registrations/edit.html.haml confirms this)
3. `current_user` is null, at least in my case

I fixed the first two issues, but not the third. So password recovery might still fail if `use_pam?` returns true I guess.

### DIFF
--- a/app/views/auth/passwords/edit.html.haml
+++ b/app/views/auth/passwords/edit.html.haml
@@ -1,18 +1,18 @@
 - content_for :page_title do
   = t('auth.set_new_password')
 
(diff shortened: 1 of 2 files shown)


ThibG:
Ok, trusting you on this here

Gargron (approved):

devkral:
thank you.

PR #6460, opened 2018-02-11
Account archive download

### LINKED ISSUE #201: Account takeout/archive download
An important feature of Twitter is that you can request to download an archive of all your tweets. This means if you delete your account (or delete much of its history) you can still retain a local copy. It also is useful for various kinds of local data-mining without overloading the server with API calls. And it gives you an important backup of your history in case the instance ever goes down.

I don't know if GNU Social or its associated protocols have any system for this yet. If not, I imagine it could be as simple as a ZIP file containing an atom feed of everything that user's posted. Ideally this would (optionally?) include media too (which Twitter doesn't do).

Twitter includes a simple static web page to help you browse your archive, but it's slow and spartan, and not a basic requirement of this feature, more a nice-to-have. If it's an atom feed, any atom feed viewer can read it, after all. So long as it's machine-readable, someone else can worry about this problem.

This is related to https://github.com/Gargron/mastodon/issues/177 and https://github.com/Gargron/mastodon/issues/109.

Gargron:
Fix #201 

- Create tar.gz
- Includes all media
- Statuses exported as ActivityPub JSON (can be potentially used in non-Mastodon software)
- ActivityPub JSON of the actor (profile data)
- Avatar, header
- PEM of the private key of your account (used for signing)

image

image


beatrix-bitrot:
👀

elomatreb:
A .zip export would be nice for Windows, if at all possible.

Gargron:
If you're not going to know how to open a tar archive, you're not going to make any sense out of the json file inside either. This export is for other programs and people who write them.

Sylvhem:
Just a thought, why not creating a tar.xz file instead of a tar.gz one? I thought LZMA has better compression ratio than deflate.

nightpool:
i would also still lean towards ZIP, if there were gems that made it as easy as these tarwriters.

i was just about to mention the actor, that's a good idea. it would also be great if the export could contain your following collection and liked posts, as well as maybe your followers and blocks/mutes for completeness?

akihikodaki (dismissed):
It should have archive expiration. Otherwise it would take much disk space.

UPDATE: I have forgot to mention that the persistent archive can also be a security problem, and that should be much a bigger problem. The recent rumor that Twitter keeps tweets after deletion has earned some reactions. But solving the issue is much easier anyway.

akihikodaki:
> Just a thought, why not creating a tar.xz file instead of a tar.gz one? I thought LZMA has better compression ratio than deflate.

That's true, but it is debatable whether it is worth XZ's computation cost.

Gargron:
>It should have archive expiration. Otherwise it would take much disk space

It already deletes any previous archives when creating a new one. Do you mean expiring the last archive too? At what interval?

akihikodaki:
> It already deletes any previous archives when creating a new one. Do you mean expiring the last archive too?

Yes.

> At what interval?

Well, from a week (same with regeneration rate limit) to a month.

PR #6469, opened 2018-02-13
Add LDAP strategy

### LINKED ISSUE #942: [Feature Request] LDAP integration

Gargron:
Fix #942 

- Unlike CAS and SAML, which redirect to another website, the LDAP strategy presents quite an ugly omniauth login form
- It seems like the uid configuration is ignored, as the uid that the omniauthable concern receives is the "cn=a,dc=etc" string, which is an invalid account username so login fails


Gargron:
Closed in favour of #6556

PR #6471, opened 2018-02-13
Click card to embed external content

### LINKED ISSUE #5173: Make external embed loaded on click
Right now, when you link a youtube video for example, it load immediately the widget when you click on the post, it's a huge privacy issue since Youtube now know where you are / etc.

I would suggest to add a little "preview" of the player, that can be click, and then it would load the Youtube widget with autostart or whatever you can imagine, but at least don't immediately sell the user to google's eyes 👀 

* * * *

Gargron:
Fix #5173 

image

image

- No external resource is loaded before user clicks play
- Play for photo embeds is replaced with zoom icon, opens image modal
- Back-end change: thumbnails now saved for videos
- Description text trim function has been improved to not cut off words in the middle

### DIFF
--- a/app/javascript/mastodon/features/status/components/card.js
+++ b/app/javascript/mastodon/features/status/components/card.js
@@ -20,6 +20,16 @@ const getHostname = url => {
   return parser.hostname;
 };
 
+const trim = (text, len) => {
+  const cut = text.indexOf(' ', len);
+
+  if (cut === -1) {
+    return text;
+  }
+
+  return text.substring(0, cut) + (text.length > len ? '…' : '');
+};
+
 export default class Card extends React.PureComponent {
 
   static propTypes = {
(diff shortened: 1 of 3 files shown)


akihikodaki:
The idea sounds good. But I think the expanding icon is a bit confusing and may give a different idea from the feature to the user: opening the linked Web site in a new tab.

What about having a small globe icon somewhere on the thumbnail? Some description may be provided with text aside the icon like "external resource" or similar.

Gargron:
@akihikodaki Do you use Discord? I think people will have the correct expectations about the play/external buttons.

akihikodaki:
@Gargron Yes, well, I checked `#project-dev` but I don't see replies saying it is confusing nor those saying it is fine. It is nice if we could hear others' opinions there.

akihikodaki (approved):

akihikodaki:
The behavior was not confusing, but I was confused what this pull request meant. This is intended to replace the embedded remote content with a simple link. It sounds good for me and I approved this change.

nightpool:
@Gargron does this mean youtube videos now take two clicks to start playing?

akihikodaki:
@nightpool You may just click the link instead of opening the status.

Well, this is just my opinion, but it is really nonsense to be unable to embed the preview card only in the expanded view. Users would just click the link instead of expanding the status. We would need different changes to improve the overall user experience.

Gargron:
@akihikodaki I already implemented autoplay in a different PR, so it's not two clicks to play, it's just one.

PR #6479, opened 2018-02-15
Do not hide NSFW media/CW'd text in OpenGraph tags

Gargron:
Reasoning: HTML title tag affects everyone. But OpenGraph only affects
when somebody is deliberately sharing the content, usually in an
environment where such content is expected. Hiding the content in
OpenGraph tags results in deceitful previews which inhibit the
shareability of the post.

Example: Somebody writes a clever post about politics but kindly
puts a "uspol" content warning on it. Mastodon users are thankful,
but sharing this post on another platform results in non-Mastodon
users believing the entire contents of the post is "uspol" and not
clicking through/reading and re-sharing.

### DIFF
--- a/app/views/stream_entries/_og_description.html.haml
+++ b/app/views/stream_entries/_og_description.html.haml
@@ -1,4 +1,1 @@
-- if activity.is_a?(Status) && activity.spoiler_text?
-  = opengraph 'og:description', activity.spoiler_text
-- else
-  = opengraph 'og:description', activity.content
+= opengraph 'og:description', [activity.spoiler_text, activity.text].reject(&:blank?).join("\n\n")
--- a/app/views/stream_entries/_og_image.html.haml
+++ b/app/views/stream_entries/_og_image.html.haml
@@ -1,4 +1,4 @@
-- if activity.is_a?(Status) && activity.non_sensitive_with_media?
+- if activity.is_a?(Status) && activity.media_attachments.any?
   - activity.media_attachments.each do |media|
     - if media.image?
       = opengraph 'og:image', full_asset_url(media.file.url(:original))


Aldarone:
Perhaps prepend Content Warning in the content tag?

If I boost (share something on my Timeline) I would not expect the CW and nsfw tags to be removed, sharing a link elsewhere is similar IMO.

I think non mastodon users should not be more exposed to triggering or nsfw content than mastodon users are.

Aldarone:
Perhaps prepend Content Warning in the content tag?

If I boost (share something on my Timeline) I would not expect the CW and nsfw tags to be removed, sharing a link elsewhere is similar IMO.

I think non mastodon users should not be more exposed to triggering or nsfw content than mastodon users are.

Gargron:
> Perhaps prepend Content Warning in the content tag?

Okay, CW will be prepended to the description with an empty line between.

nightpool:
I'm on board with this for CW'd text, but I'm not sure if it makes sense for NSFW media. Why provide a NSFW preview, and then send someone to a blacked-out page that requires they click to see the NSFW media? Seems like it should either be one or the other.

Gargron:
I checked behaviours of OpenGraph previews of some other platforms with semi-support for NSFW images:

- Twitter: Even if account is marked as sensitive, tweets with images still show the images in preview
- Tumblr: Even if post is tagged #nsfw the preview shows the image
- Danbooru (software) image boards: even if rating:explicit, the preview shows the image

On the other hand:

- Pixiv prepends [R-18] to og:title and shows no image in the preview (*But we can't use this strategy because on our platform hidden media is not always pornography, which R-18 is heavily associated with. It could be innocent stuff like food, or it could be gore in the extreme case*)
- deviantArt shows an image of a grey stop sign as the image in the preview

Most importantly I believe that: 

1. We want to capture migrating users from Tumblr and Twitter, rather than Pixiv or deviantArt, so the UX expectations that they bring from those platforms matter more, even if those platforms are not known for good native support for NSFW controls
2. My reasoning in the opening post about sharing contexts and the implicit opt-in to the content when sharing it outside of Mastodon still makes sense to me
3. The behaviour of OpenGraph tags does not diminish the support for CW/NSFW *in* Mastodon, since OpenGraph is for *outside* of Mastodon

app/views/stream_entries/_og_description.html.haml
```diff
@@ -1,4 +1 @@
-- if activity.is_a?(Status) && activity.spoiler_text?
-  = opengraph 'og:description', activity.spoiler_text
-- else
-  = opengraph 'og:description', activity.content
+= opengraph 'og:description', [activity.spoiler_text, activity.text].reject(&:blank?).join("\n\n")
```
nightpool: should maybe be `"CW: " + activity.spoiler_text`?

nightpool:
I'm okay with it if it feels like there's consensus. I think my concern about the uncensored image -> censored image -> uncensored image flow is less important then having good previews and making the static page consistent with the timeline. 

im still worried that people will expect the image to be hidden and then be (potentially badly) surprised when it isn't

app/views/stream_entries/_og_description.html.haml
Gargron: I know we have "CW" on a button in the UI but adding it inside of a user's post could be problematic because it's an English string.

nightpool:
(alternative that came up but seemed too confusing: a "don't hide sensitive media on permalink pages" option that affects both the static page and link preview behavior per-account)

nightpool (approved):

clarcharr:
Thought: perhaps Mastodon could emit custom meta tags that could be parsed in a similar way to OpenGraph to check for content warnings? This would mean that sites which specifically integrate with Mastodon's tags could show the content warnings on their own UIs before even showing the page title.

Essentially, `cw:subject` would be the text for the content warning and `cw:body` would be the remaining body of the toot. `og:description` and `description` could include the text with the CW at the top for systems which don't support this.

Gargron:
@clarcharr I like this idea a lot!

app/views/stream_entries/_og_description.html.haml
Sixthhokage1: (late comment I know, but...)
I tried this just now, and in addition to being a hardcoded English string it also ends up prepending a "CW: " line on top of non-CWed toots as well. But it would be a good idea to try and add this so that, say if I'm linking toots in Discord with people who don't use masto, they have the context that this line up top is a content warning without me having to say so or having to click through.

deli73:
alternate fix proposal: change the preview to make it clear that the content warning is a content warning and not the entire content of the post
e.g. prepending "content warning: " and appending "(click link to see full post)" or something like that.
fixes the confusion while retaining the usefulness of CWs between sites

Sixthhokage1:
Personally, I reverted the change on my instance and modified app/views/stream_entries/_og_description.html.haml to read
 ```
- if activity.is_a?(Status) && activity.spoiler_text?
  = opengraph 'og:description', "CW: " + activity.spoiler_text + "\n\nOpen link to view"
- else
= opengraph 'og:description', activity.content
```
In my own use-case, a Discord server I'm on has an NSFW channel that requires any explicit content to not have an explicit preview and be given a content warning. It's much easier for me to have that automated with the opengraph tags than to duplicate effort suppressing the preview with angle brackets and copying the content warning to Discord, especially on mobile.

Gargron:
Would everyone be appeased if I add a URL param to disable OG tags? So by default it would work like right now but if you have special requirements you can append something like ?no_preview to the URL to avoid an opengraph preview?

Fidgetcetera:
No, because what people are asking for is a preview which still shows the content warning, not no preview at all. 

Ideally that'd be default, since when using mastodon, we expect posts to be hidden behind their warnings, and when sharing them, we don't expect to suddenly need to take extra steps to hide things.

Putting "content warning: " before the warning is plenty enough for people to figure out they should click the link to get the whole post.

nightpool:
@Sixthhokage1 i tested both tumblr content marked sensitive and tumblr blogs marked sensitive when me and @Gargron were discussing this pull request. tumblr blogs marked sensitive are completely invisible unless you have an account which is why we only really considered posts marked sensitive for comparison

Sixthhokage1:
Anything considered explicit under e621 rating standards or that the user thinks would be particularly sensitive has embeds manually disabled by enclosing the link in angle brackets and a content warning given. Which is a pain in the ass on mobile

Xkeeper0:
Hey I'm gonna drop in here and be blunt for a while because I super disagree with this.

CWs are important. The ability to do useful CWs and tagging is one of the main features of Mastodon over other services, as it allows for people to make informed decisions of what they see. **Ignoring these by default is absurd.**

Take this scenario:

1. User A posts some sensitive media and tags it appropriately. Everyone on Mastodon sees "`sensitive [see more]`. This is good.
2. User B wants to share this, so they copy the link to the status from their timeline or the URL bar or whatever. (Nobody on the planet is going to manually add `?respect_content_warnings` to the URL.)
3. User B pastes their link into Discord, Telegram, or some other OpenGraph-supporting medium.
4. Anybody who is a party to this now gets a nice big preview of whatever it was User B decided to share, **without any ability to choose whether or not to see it**, completely and utterly defeating the purpose of CWing the content in the first place.

> We want to capture migrating users from Tumblr and Twitter, rather than Pixiv or deviantArt, so the UX expectations that they bring from those platforms matter more, even if those platforms are not known for good native support for NSFW controls

"Well, other services are bad" is **no excuse** to do what they do.

----

To give an example of how "well, having the CW in the text fixes it" is wrong:

Imagine that the CW is just "pet death" and the text is "XXX died last night :(" and the picture was of said pet. NONE of this is hidden now because it fits within the preview and image size.

image

Now replace that sad smiley face with something a lot worse and imagine how well that would go.

Revert this.

KScl:
Why are we caring about following other services' failures?

I just got linked something NSFW while I was at work, by my gf who thought that the CW/sensitive flags would stop it from being shown until I was somewhere with some privacy, as those flags would normally provide... except due to this change, that no longer happens.

This is as far as I can tell the _opposite_ of what users expect when they share content.

lordminx:
This is a terrible, terrible change.

CWs are there for a reason and one of the reasons why Mastodon took off so well. One of the core features of mastodon should not be thrown out by default since it's obviously not what users would expect.

"Ignore CWs when sharing links" is an Opt-In feature if I've ever seen one.

Gargron:
It seems like you have jumped into the thread without trying to understand the preceding discussion and the goals of this change.

Yes, I'm changing an existing behaviour, and since you were used to it, that means I am taking it away from you. I don't expect *you in particular* to be happy about that, but I have to do it to make Mastodon more inclusive to other groups of people. I think that this is a small change (a few OpenGraph tags!) with a big potential positive impact. The downside is that you, who were used to the old behaviour, will now have to:

>Anything considered explicit under e621 rating standards or that the user thinks would be particularly sensitive has embeds manually disabled by enclosing the link in angle brackets and a content warning given. 

Which you were already doing for non-Mastodon links anyway.

Fidgetcetera:
This isn't just "old users expect this and it's changed oh no"

Any user, even someone who's just joined Mastodon, who goes to share a link will expect it to not show what's behind the content warning, because that's the precedent set in the *rest of the software.*

It's like if you shared a Twitter link to a post with images, and the preview gave the person's profile image instead of previewing one of the post's images. It wouldn't make sense, because that's not the expected content of the post on the platform.

deli73:
goddammit blargron you can have both options at least make no-preview the _default_,

instead of: "ignores CWs unless you put `?respect_content_warnings` on the url"
we could do: "respects CWs unless you put `?preview_hidden_content` on the url" or something

if it's really so important to nsfw artists to have that option, then they'll be able to do it still, but it won't cause a massive issue for everyone else like this would.

-L

KScl:
I didn't just jump in, Eugen. I saw this PR at least a week ago, it just turns out that the quickest way for someone's opinion to change from "eh, whatever" to "vehement dislike" is to actively experience negative effects from said change.

Also, the issue isn't inherently with _our_ behavior; those of us talking here already know the change exists and know to hide previews when possible. (By the way, Telegram doesn't allow this, as far as I know.) The issue is with _every other user of the software_ who _doesn't_ know this change has taken place, who will proceed to share links as they always have and then find out hours later that they accidentally posted porn to their SO while they were working since content warnings suddenly weren't respected anymore by previews.

I was going to post more, but as I was writing both of the above two posts came in and basically said my other points anyway. I can see a benefit to being able to show content through the preview sometimes, but _by God it should not be the default_.

Xkeeper0:
Ah yes I just jumped in without considering anything, definitely not after talking to many people who went "what" "why" "this is terrible" "now I have to remove all my nsfw/sensitive stuff"

Meanwhile the only thing people gain is from seeing sensitive content in previews without -- *zounds!* -- having to click another link.

Why don't you just make CWs and media show up unhidden by default in the Mastodon interface? After all, people posted that stuff intending to share it.

Gargron:
>Why don't you just make CWs and media show up unhidden by default in the Mastodon interface? After all, people posted that stuff intending to share it.

What I'm about to say here is a digression from the topic and I don't expect anyone to continue *this* thread, so please see my previous post for the relevant stuff. But I want to say, this is nonsense. *Within* Mastodon, there is one environment, in which we control what the UX is like. *Outside* of Mastodon, there is competition, between links to different platforms. If the same content within a Twitter link is more likely to be clicked and shared than within a Mastodon link, then that's a loss for our platform, and specifically a loss for people who choose to post on Mastodon and *want* their stuff to be seen. And the odds are *already* against us because many platforms have special rules for displaying Twitter links which are not sourced from OpenGraph.

Cassolotl:
I just saw [this](https://scalie.club/@Dex/99704730846052046) in the public timeline, which summarises my view pretty well:

> i'll be honest, it has been kind of confusing when i've linked toots to non-masto using friends and the cw text was the only thing displayed in the description
> 
> my preferred solution would be for the external description to be "content warnings: (cw text)" and the external image to be the user's avatar

KScl:
> Okay, I'd just like to take a moment and say that it is very frustrating to hear such things 31 days after it's been proposed and discussed [...]

Look, I know it's frustrating to see complaints so long after the original proposal, but this is bound to happen at some point. Every one of us can't personally vet every pull request ever, and even if we could, actually experiencing the effects of the pull request may result in a complete change of opinion (as in my case).

> How do we make sure that both use cases are satisfied & are discoverable?

This is obviously not in the scope of this PR, but I personally think there should be a "Share" dialog available in the drop-down between "Expand this status" and "Embed", that contains both variants of the URL that can be copied just like in the Embed dialog.

deli73:
yeah. again, what's important is that the default is NOT showing the fucking CW'd content.
Alos this shouldn't need to be said but our top priority here should be making the site _safe_, not attracting more users. When the focus becomes "attracting more people", most of the people attracted end up being huge assholes (see: the techbro flood of late last year) and i for one would rather focus on keeping existing communities safe.
-L

Cassolotl:
>it is very frustrating to hear such things 31 days after it's been proposed and discussed, and especially after 8 days of release candidate testing, and after the final release was made?

I can understand this frustration, but the list of pull requests isn't frequented by the vast majority of users, because 95% of pull requests are complete gibberish to non-coders. Most average users find out about features after they are merged and their instance gets updated.

Most people who read pull requests pre-merge are opinionated, technically-minded, developery people who are also very established users of Mastodon. People who are used to using features that might not be totally perfect because they understand how they were made that way, and people who are likely to tolerate or not even notice that something is kinda UI-broken. If you want to avoid this happening in future, maybe some work needs to be done to make feature proposals and design proposals accessible to average users - and not just accessible, but their views as people who are used to using software *better than this* perhaps need to be actively sought or they won't know that we need them. Most people are just not that opinionated. Something something focus groups, maybe?

Gargron:
>Alos this shouldn't need to be said but our top priority here should be making the site safe, not attracting more users.

These two things are absolutely not mutually exclusive, the only reason why you are here in the first place is that at some point some part of the software attracted you as a new user. Let me quote the project's description from the README:

>The social focus of the project is a **viable** decentralized alternative to commercial social media silos

I put emphasis on viable. Viable as in "you can actually use it instead", i.e. it's competetive with the alternatives. Clearly we have network effects to fight too, but if the *software itself* inhibits your success, then people won't choose it. I understand if you think I made the wrong choice in this particular case, but if you think the reason I did it, the whole idea of making the platform more inclusive to more groups so it can become more widespread, is wrong, then we're not on the same page, and never were.

If you want to make a version of Mastodon that repels new users, you can, because it's free software. However I see no way that you can make *me* do it with *my* version of Mastodon.

One of my favourite stories is about one Mastodon user at a gamedev conference who was asked how they could be found online. The user said they were on Mastodon, and instead of stares, they got understanding nods. *That's* the utility of widespread use for Mastodon users.

___

As for the actual issue, here are the various ways forward:

| Way | Caveats |
|-----|---------|
|Make it the poster's user setting|You don't know what to expect when sharing the link|
|Make it the viewer's user setting|Impossible because Discord/Telegram/etc don't use your session to fetch preview|
|Make it an admin setting instance-wide|Well, you still don't know what to expect when sharing the link unless you start remembering the behaviours of different servers|
|Add URL param to control this|URL param not discoverable, so mostly the default will be used anyway|
|Add URL param with sharing screen|Sharing screen not super discoverable if people just copypaste the URL from permalink or address bar; Public pages don't even have a UI precedent for sharing controls|
|Change default to hide images|NSFW artists, sex workers and other adult content creators as well as people who just CW food pics or jokes are disadvantaged compared to if they chose Twitter or Tumblr to post|
|Leave as is|Accidentally show porn to SO in Discord/Telegram/etc while they are at work|

Pick your poison.

Cassolotl:
I don't understand the URL param ones, so even your meta isn't accessible to average users. But I think I understand most of them, so it's cool.

It sounds like it's possible to change the URL manually so that CWs are respected, but it's manual and not the default so it doesn't get used. Can some automation be added? We have a UI that asks for user preferences as a result of a bunch of people objecting to a change (mute: mute in replies y/n?) so maybe something like that could work?

Embed > If toot has content warning, do pop-up: Ignore content warning y/n? > If yes, give non-default link so that embedded toot has visible toot text and images, despite CW. If no, give default link that says "content warning: [content warning]" when posted elsewhere.

Gargron:
>Edit: But I do think that literally just adding the words "content warning: " to the start of an embed thingy so that people understand that USpol isn't the entirety of the post would solve 99% of problems.

I'm pretty sure most people's issue here is not with the text changes but with images. That being said, putting English words inside what's supposed to be the toot text is a bit Anglocentric. A lot of people who don't speak English use Mastodon. That is why I avoided that solution.

deli73:
> putting English words inside what's supposed to be the toot text is a bit Anglocentric
then make that text localizable based on the instance it's posted on????????

"Add URL param with sharing screen" + default being not showing the image is perfectly fine. I don't agree with considering people to be "disadvantaged" for having to take a bit of extra action to not hide content _that they chose to hide on the site_, Really not appreciating your prioritizing of easy exposure over people's fucking safety.
-L

Gargron:
>people's fucking safety.

Please explain how any of these changes affect people's *safety*. We're still talking about Discord previews in controlled environments with people who know each other, right? I want to follow your arguments and come to an agreement, but this sounds like you're raising emotional stakes just for the sake of it.

Cassolotl:
> I'm pretty sure most people's issue here is not with the text changes but with images.

Ohhh fair point.

> That being said, putting English words inside what's supposed to be the toot text is a bit anglocentric. A lot of people who don't speak English use Mastodon. That is why I avoided that solution.

I think even a triangle with an exclamation mark in it ⚠ (like, "⚠ USpol") would be better than just posting the sensitive content in full. That's an emoji, and they're intended to be clear and recognisable in any language, right?

Cassolotl:
>then make that text localizable based on the instance it's posted on

Or the detected language of the toot, since we have language detection per toot for the public timelines now!

deli73:
> Please explain how any of these changes affect people's safety

If people share a toot in an environment where people could be hurt by the content, and they _expect_ it to hide it, and it _doesn't_, it can and will cause harm.

multiple people here have mentioned this _exact_ thing happening to them, including people who _already knew about the change_. It's not intuitive, many people will forget if they haven't directly witnessed harm caused by it yet.
-L

Cassolotl:
>many people will forget if they haven't directly witnessed harm caused by it yet.

It's that privilege thing. It's like when men make profiles on OKcupid pretending to be women and they are like "wow women around me told me it was bad but I seriously had no idea." If you don't need CWs, it is very easy to dismiss how helpful they are for reducing harm.

ekiru:
> I'm pretty sure most people's issue here is not with the text changes but with images.

fwiw, I am more bothered by the text changes than the image change (mostly because of the patterns of what I am likely to share). It's also the case that other people have brought up their concerns specifically with the text changes earlier in this thread and proposed the same thing and you appear to have ignored those concerns (in the sense of not responding to them at all). This discourages other people from echoing their similar concerns because other people have already visibly expressed the concern without any effect and is why I had not posted about it before.

cosmicchipsocket:
This is only about the billionth time I've seen an open source project implement user-hostile changes, and then its developers get angry at the backlash. It's almost like there are flaws when your software development process is entirely tech-oriented. You say you want an ethical alternative to existing social media. With this kind of process you won't have an ethical alternative, or really any alternative at all. When you anger the users, they begin to look elsewhere. People are going to start finding alternatives to your alternative. Don't take your userbase for granted.

Actually listen to your users.

I'm thinking that a lot of open source projects, including this one, need to adopt a system whereby user-affecting changes have to actually be approved by the users first.

deli73:
^
_thank you_

Gargron:
>It's almost like there are flaws when your software development process is entirely tech-oriented.

This change was initiated *in the first place* from talking to people not represented on GitHub.

>When you anger the users, they begin to look elsewhere. People are going to start finding alternatives to your alternative.

It's called a fork. There are many, because I'm releasing my work as free, open source software. Please reflect on the entitlement seeping through your comment.

cbondurant:
On the topic of the ?respect_content_warnings idea. That could work if toots had a dedicated share button, kinda like how youtube has a dedicated share button that has a built-in for sharing the current timestamp, and either sharing or not sharing the current playlist. Which adds ?s=N to the share link.

The share toot button could contain the option to _disable_ content warnings (because its better to have disabling be the choice made with intent) as well as other options that might come up in the future. It could even go so far as "only disable text content warning" or "only image" if those features turn out to improve the user experience.

Even if this isn't the solution that is gone with, I do agree that fully disabling them completely causes just as many problems with the open graph display as there were before. It just moved the problem to something else.

cosmicchipsocket:
> This change was initiated in the first place from talking to people not represented on GitHub.

You need to consider not only the people who want the change, but also the people who might be impacted by it. Knowing the implications will lead to better decisions in the future.

> It's called a fork.

Forks are technologically oriented, which most users are not. Forks run the risk of fragmentation, and they can lead to confusion over behavior when different instances employ different behaviors, a point you had brought up yourself.

Cassolotl:
I agree, most people who are impacted by this change would not be able to just set up their own instance, let alone make modifications to suit their needs.

deutrino:
I feel like for CW'd posts, hiding the images and adding (perhaps with localized words based on instance or toot language) "⚠️ content warning: cw text here" is a workable enough compromise that could be implemented quickly.

The question then becomes how to do away with the behavior for users who want their CW'd posts to display a full preview outside the network. A couple observations:

  * a setting applicable to the user's entire profile is problematic because it breaks the expectations of *other* users who might be sharing from the permalink
  * a setting that any user sharing the link can set (whether by button or modifying the url, etc) could be problematic because it may break the expectations of the OP
  * it would seem from these opposing concerns that perhaps the OP should be the only person able to disable the CW in the share preview when sharing a CW'd toot, which then leads to thorny implementation questions, not only about how to implement the UI in mastodon itself, but how could eg mobile clients do this; all of which is worth considering but would likely take more thought and more work than the initial compromise. 

I'm fully on board with Mastodon being as rich a multimedia experience as possible and as interoperable as possible with things like previews, etc. I think it's good for the network. And I'm a heavy tumblr user who'd like to see lots of cool art, videos, even music shared via Mastodon. A fair number of the comments I post on github for Masto and my preferred mobile client Mastalab reflect this. I could absolutely see myself being frustrated with in-Masto CW vs out-of-Masto preview CW when trying to share a cool image post from my aesthetic alt. 

All that being said, it seems like having CWs propagate by default, and working to allow at least the OP to disable that behavior when they themselves share a toot, may be the best compromise.

deutrino:
BTW @Gargron, it's clear to me that you've put a good deal of thought into this both before and after it blew up - I mean, it does take effort to stick a formatted table in a github comment! - and so I wanted to make clear that most people really do appreciate your blood sweat & tears spent on this project. You've built a network that's home to literally hundreds of thousands of people and groups, many with disparate interests. It will be tough to balance things. Thanks for sticking with it.

cbondurant:
> a setting that any user sharing the link can set (whether by button or modifying the url, etc) could be problematic because it may break the expectations of the OP

I feel like in the case of sharing content, the expectations of the OP in posting isn't the consideration that needs to be made. In most cases specific instance of linking a toot does not directly affect the OP and the OP would not know the context in which that toot was shared, meaning if you gave them the only one with power to decide, the decision would mostly end up being done arbitrarily, or more likely always left at the default.

In the case of an option that the sharer can choose, but defaulting to the safe option of hidden, the onus is on the sharer to not make a bad decision and directly share CW'd content in a place it should stay CW'd. This would be no different than any other case of having to think through linking dangerous content, e.g. not being dumb enough to link porn in a sfw context. 

The important part of this way of doing it is it requires a direct and affirmative action to disable the CW, meaning there is very little chance of a dangerous accident. While it still technically allows for someone to intentionally link a CW'd message in a poor context without the CW, that's something they can already do using any other source on the internet.

PR #6482, opened 2018-02-16
Cache relationships in API

Gargron:
Relationship retrieval is expensive because it hits 5 different tables in 6 queries.

- Cache set when relationships are first retrieved
- Cache is invalidated when a model that's part of the relationship is created/destroyed
- Hit the relationships API for search results in web UI, now that it's cheaper

### DIFF
--- a/app/presenters/account_relationships_presenter.rb
+++ b/app/presenters/account_relationships_presenter.rb
@@ -57,12 +57,12 @@ class AccountRelationshipsPresenter
   def cache_uncached!
     @uncached_account_ids.each do |account_id|
       maps_for_account = {
-        following: following,
-        followed_by: followed_by,
-        blocking: blocking,
-        muting: muting,
-        requested: requested,
-        domain_blocking: domain_blocking,
+        following:       { account_id => following[account_id] },
+        followed_by:     { account_id => followed_by[account_id] },
+        blocking:        { account_id => blocking[account_id] },
+        muting:          { account_id => muting[account_id] },
+        requested:       { account_id => requested[account_id] },
+        domain_blocking: { account_id => domain_blocking[account_id] },
       }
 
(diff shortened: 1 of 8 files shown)


app/presenters/account_relationships_presenter.rb
```diff
@@ -5,11 +5,67 @@ class AccountRelationshipsPresenter
...
+        @uncached_account_ids << account_id
+      end
+    end
+
+    @cached
+  end
+
+  def cache_uncached!
+    @uncached_account_ids.each do |account_id|
+      maps_for_account = {
```
akihikodaki: It should cache the relationships of the account specified by `account_id` rather than saving the relationships of all the account specified by `account_ids` for memory efficiency.

app/presenters/account_relationships_presenter.rb
```diff
@@ -5,11 +5,67 @@ class AccountRelationshipsPresenter
...
+      blocking: {},
+      muting: {},
+      requested: {},
+      domain_blocking: {},
+    }
+
+    @uncached_account_ids = []
+
+    @account_ids.each do |account_id|
+      maps_for_account = Rails.cache.read("relationship:#{@current_account_id}:#{account_id}")
```
akihikodaki: `read_multi` may be used, may be not.

akihikodaki (changes requested):

app/presenters/account_relationships_presenter.rb
Gargron: Thanks for catching this!

app/presenters/account_relationships_presenter.rb
Gargron: I think with a limit on maximum requested relationships below 1000 there's almost no performance difference between read and read_multi, it would be a bit more bothersome to match back the values to the IDs..

akihikodaki (approved):

PR #6484, opened 2018-02-16
Push discovered status through streaming API within a time window

Gargron:
Time window of 6 hours.

Alice and Bob are on server Alpha. Bob doesn't have any followers but Alice. Alice has followers on server Charlie. Alice boosts Bob's status. Alice's followers will see Bob's status, but nobody else on server Charlie will see it, because it won't appear on the federated timeline. So Bob has less chances of getting followers from server Charlie.

This PR fixes this behaviour as long as Bob's post is not older than 6 hours.

### DIFF
--- a/app/lib/activitypub/activity.rb
+++ b/app/lib/activitypub/activity.rb
@@ -74,7 +74,7 @@ class ActivityPub::Activity
 
     # Only continue if the status is supposed to have
     # arrived in real-time
-    return unless @options[:override_timestamps]
+    return unless @options[:override_timestamps] || status.within_realtime_window?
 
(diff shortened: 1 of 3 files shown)


akihikodaki (approved):
Reblogs are certainly what are missing in federated timeline and I like this idea.

Well, I am a bit worrying that `Status` model has too much things about timelines, (`with_realtime_window?`) but that does not have to be addressed now.

akihikodaki:
By the way, should `within_realtime_window?` be named as such? It may be better to change the name to represent the feature (to determine whether it should be distributed now) rather than the implementation (to check the creation time.)

PR #6486, opened 2018-02-17
Redesign landing page (again)

Gargron:
Inspired by Pawoo

New admin setting: hero image (fallsback to instance thumbnail)

Responsive design demo:

Large:

screen shot 2018-02-18 at 00 00 23
screen shot 2018-02-18 at 00 00 30
screen shot 2018-02-18 at 00 00 38

Medium:

screen shot 2018-02-18 at 00 01 05
screen shot 2018-02-18 at 00 01 09

Small:

screen shot 2018-02-18 at 00 01 22
screen shot 2018-02-18 at 00 01 32
screen shot 2018-02-18 at 00 01 36

### DIFF
--- a/app/javascript/styles/mastodon/about.scss
+++ b/app/javascript/styles/mastodon/about.scss
@@ -1,3 +1,129 @@
+$fluid-breakpoint: 1180px;
+$column-breakpoint: 600px;
+$small-breakpoint: 960px;
+
+.container {
+  box-sizing: border-box;
+  max-width: 1140px;
+  margin: 0 auto;
+  position: relative;
+
+  @media screen and (max-width: $fluid-breakpoint) {
+    width: 100%;
+    padding: 0 10px;
+  }
+}
+
+.show-xs,
+.show-sm {
+  display: none;
+}
+
+.show-m {
+  display: block;
+}
+
+@media screen and (max-width: $small-breakpoint) {
+  .hide-sm {
+    display: none !important;
+  }
+
+  .show-sm {
+    display: block !important;
+  }
+}
+
+@media screen and (max-width: $column-breakpoint) {
+  .hide-xs {
+    display: none !important;
+  }
+
+  .show-xs {
+    display: block !important;
+  }
+}
+
+.row {
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0 -5px;
+
+  @for $i from 1 through 15 {
+    .column-#{$i} {
+      box-sizing: border-box;
+      min-height: 1px;
+      flex: 0 0 percentage($i / 15);
+      max-width: percentage($i / 15);
+      padding: 0 5px;
+
+      @media screen and (max-width: $small-breakpoint) {
+        &-sm {
+          box-sizing: border-box;
+          min-height: 1px;
+          flex: 0 0 percentage($i / 15);
+          max-width: percentage($i / 15);
+          padding: 0 5px;
+
+          @media screen and (max-width: $column-breakpoint) {
+            max-width: 100%;
+            flex: 0 0 100%;
+            margin-bottom: 10px;
+
+            &:last-child {
+              margin-bottom: 0;
+            }
+          }
+        }
+      }
+
+      @media screen and (max-width: $column-breakpoint) {
+        max-width: 100%;
+        flex: 0 0 100%;
+        margin-bottom: 10px;
+
+        &:last-child {
+          margin-bottom: 0;
+        }
+      }
+    }
+  }
+}
+
+.column-flex {
+  display: flex;
+  flex-direction: column;
+}
+
+.separator-or {
+  position: relative;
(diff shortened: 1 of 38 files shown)


charlag:
Looks cool!
What do you think about some kind of a spoiler for the mobile layout, for fields or for description or IDK? Is it worth it?

akihikodaki:
This is just my own opinion, but I don't like that 3-columns landing page much. The left side column is so empty and not balanced with the right side.
I don't know how Pawoo's landing page was designed, so this is just my inference, but 3-column layout may have been considered as essential to emphasize Pawoo banner at center. I guess, however, most instances do/will not have such a big hero image.

An alternative choice is to layout the page like this:
<table>
<tr><td>FORM</td><td>HERO</td><td rowspan="2">TIMELINE</td></tr>
<tr><td colspan="2">DESCRIPTION</td></tr>
</table>

noiob:
I think this is way too busy for a landing page. The current page is a lot more friendly. While I agree that it could use a rework, having some space around the UI elements really helps. On the three-column layout I don't know where to look first and while responsive design is cool, shifting the signup form around makes it even more confusing.

Let's also keep in mind that some instances choose to not display the timeline on the front page. The layout still has to look good without it.

Gargron:
>while responsive design is cool, shifting the signup form around makes it even more confusing.

Well, nobody resizes their window usually, so for someone with a big screen and someone with a small screen the form would always be in the same place (from their perspective)

noiob:
True, I was talking more about from a designer's perspective, it makes it harder to put emphasis on the main form. If that makes sense

RobertRence:
Some various thoughts:

1) By defaulting to "sign up" instead of "log in", you use more screen space. 
2) There is a significant amount of dead space on the left column of the primary layout.
3) It's not representative of the actual site. There is an example of the timeline stream on the right, but the site itself uses 4 columns. 

My thoughts would lean to:

Log in | Hero | About | Home to
----------------------- | ------------- | --------------------- | ------------------------
Demo On your mind | Demo Home | Demo Notification | Demo Getting started

But you do you.

koyuawsmbrtn:
this looks gorgeous...ship it, please 👍

app/javascript/styles/mastodon/about.scss
```diff
@@ -460,111 +529,282 @@
...
+      margin: 40px 0 !important;
+    }
+
+    &__action {
+      text-align: center;
+    }
+  }
+
+  .features-list {
+    margin-top: 20px;
```
brennovich: Don't we have variables for consistent spacing across pages?

app/javascript/styles/mastodon/about.scss
```diff
@@ -460,111 +529,282 @@
...
+      .visual {
+        flex: 0 0 auto;
+        display: flex;
+        align-items: center;
+        margin-left: 15px;
+
+        .fa {
+          display: block;
+          color: $ui-primary-color;
+          font-size: 48px;
```
brennovich: Also, don't we have variables for font sizes neither?

app/javascript/styles/mastodon/containers.scss
```diff
@@ -1,4 +1,4 @@
-.container {
+.container-alt {
```
brennovich: I don't get this changes 🤔 Is it to avoid colision with other usages of `.container` in other templates?

config/locales/ar.yml
```diff
@@ -14,7 +14,6 @@ ar:
       humane_approach_title: أسلوب يعيد الإعتبار للإنسان
       not_a_product_title: إنك إنسان و لست سلعة
       real_conversation_title: مبني لتحقيق تواصل حقيقي
-    find_another_instance: إبحث عن مثيل خادوم آخر
```
brennovich: Why do we need to remove `find_another_instance` translations? Isn't used by new template anymore, or it's not used since a long time?

brennovich (commented):
Just passing around as I'm new to mastodon and I'm using your instance 🤓

I saw you toot about this PR and I thought that I could start contributing with Mastodon <3

I made some silly questions, I'm without context and didn't even run the project yet, so feel free to disconsider this review.

Most of questions are to understand a bit more about the project, though.

Also, I'd suggest to squash all commits related with new landing page  before merge (I'd put in another commit removal of `find_another_instance` if it's just a boy scout change).

I hope you find any of this helpful.

config/locales/ar.yml
Gargron: Mostly because the Travis build will complain if there's unused translations.

To give a bit more context, I recently accidentally linked an instance with closed registrations to someone who hasn't heard of Mastodon. They believed Mastodon was closed for sign ups, despite the "Find another instance" button, because, I'm assuming, it's just too different to what they expect.

By renaming that button back to "Sign up" we might do better.

PR #6494, opened 2018-02-17
Implement text blocks

akihikodaki:
This is a response to:
Un carattere indiano fa crashare iPhone, Mac e iPad | MobileWorld
http://www.mobileworld.it/2018/02/14/carattere-indiano-crash-iphone-mac-ipad-144881/

Because such a problem is expected to occur also in the future, this kind of feature would be necessary.

~~This implementation is primitive and has rooms of improvements, but works.~~

UPDATE:

* Implemented various improvements. Now it supports logging and severity "silence", and is usable not only for security purpose, but also for moderation purpose. The silence feature is expected to be used to exclude aggressive words from the public timeline without interfering users' activity.

* The "reject" severity reports what word is prohibited in its error message. It is like HTTP 451 and to prevent arbitrary censorship on the fediverse.

* ~~Spec and edit feature is pending.~~

UPDATE:

* All features has been implemented.

### DIFF
--- a/app/controllers/admin/text_blocks_controller.rb
+++ b/app/controllers/admin/text_blocks_controller.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Admin
+  class TextBlocksController < BaseController
+    before_action :set_new_text_block, only: [:index, :create]
+    before_action :find_and_set_text_block, only: [:show, :update, :destroy]
+    before_action :authorize_text_block
+
+    def index
+      page = params[:page]
+      @text_block = TextBlock.new if page.nil?
+      @text_blocks = TextBlock.recent.page(page)
+    end
+
+    def show; end
+
+    def create
+      if @text_block.save
+        log_action :create, @text_block
+        redirect_to({ action: :index }, notice: I18n.t('admin.text_blocks.created'))
+      else
+        @text_blocks = TextBlock.page(0)
+        flash.now[:alert] = @text_block.errors.full_messages.first
+        render :index
+      end
+    end
+
+    def update
+      if @text_block.update text_block_params
+        log_action :update, @text_block
+        redirect_to({ action: :index }, notice: I18n.t('admin.text_blocks.updated'))
+      else
+        flash.now[:alert] = @text_block.errors.full_messages.first
+        render :show
+      end
+    end
+
+    def destroy
+      if @text_block.destroy
+        log_action :destroy, @text_block
+        flash[:notice] = I18n.t('admin.text_blocks.destroyed')
+      else
+        flash[:alert] = @text_block.errors.full_messages.first
+      end
+
+      redirect_to action: :index
+    end
+
+    private
+
+    def authorize_text_block
+      authorize @text_block
+    end
+
+    def set_new_text_block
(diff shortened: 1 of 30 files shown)


app/controllers/admin/text_blocks_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module Admin
+  class TextBlocksController < BaseController
+    def index
+      @page = params[:page]
+      @texts = Kaminari.paginate_array(Setting.blocked_texts).page(@page)
+    end
+
+    def create
+      text = params.require(:id)
```
abcang: I think `:text` rather than `:id` is suitable.

nightpool:
this seems like a bad idea. shouldn't app developers block text snippets
that have issues on their end, before rendering? why do we have to add this
added complexity to mastodon for one Apple bug?

akihikodaki:
> shouldn't app developers block text snippets that have issues on their end, before rendering?

No. I believe Mastodon's Web UI supports iOS.

> why do we have to add this added complexity to mastodon for one Apple bug?

Not only for one bug. Not only for one product. That kind of problems occurred many times in history of the Internet. I have not mentioned in the first comment, but it is obvious that there are much more examples if it comes to the necessity of text filtering: not only security, but moderation, legal, censorship, etc.

config/settings.yml
```diff
@@ -50,6 +50,7 @@ defaults: &defaults
   activity_api_enabled: true
   peers_api_enabled: true
   show_known_fediverse_at_about_page: true
+  blocked_texts: []
```
abcang: Probably not necessary anymore

Gargron:
The more I think about it the more I am entirely uncomfortable with the subject of this PR.

akihikodaki:
Indeed I would not use instances with arbitrary text blocks, especially if they are with "reject" severity. That's why here is this pull request. This change assures to tell users what texts are rejected:

> Prohibited text: %{text}

This change complements the room between the ideal situation and the reality; the instances may be forced to reject some texts because of security (e.g the recent iOS case), or legal. An instance owner can still show he has nothing to hide by using the feature even if he he cannot help certain texts.
In other words, _if he doesn't use the feature and still rejects some texts, he has something to hide deliberately._ That is what I want to show.

jeroenpraat:
Feel likes a censorship tool to me. When it's cause of legal issues, an instance owner can choose between:
1. suspend the user
2. moves its instance to another jurisdiction

akihikodaki:
> When it's cause of legal issues, an instance owner can choose between:

This pull request is not to help an instance owner to deal with legal issues but to tell an user how the owner is dealing with.
As I noted earlier, text blocks are implemented in the way that:
* not intrusive ("silence" severity) or
* acknowledgeable ("reject" severity)

The actions you suggested are extreme and may not be possible. Instead of requiring them to the owners, this allows users to know how the owners deal with problems and migrate to another instance if necessary.

donkeybonks:
Consider, a hypothetical `kids.social` might want to block the top 100 things that are unsafe to young kids: swearing, cursing, certain nouns. I don't think its evil to allow site admins to block text as long as it is transparent.

In the above example, even transparency could be an issue.

PR #6496, opened 2018-02-18
Set Docker permissions during the build process

### LINKED ISSUE #3194: Permission update on docker entrypoint takes a long time

Gargron:
Fix #3194

I'm not sure why @Wonderfall didn't open this PR...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,10 @@ FROM ruby:2.5.0-alpine3.7
 LABEL maintainer="https://github.com/tootsuite/mastodon" \
       description="A GNU Social-compatible microblogging server"
 
-ENV UID=991 GID=991 \
-    RAILS_SERVE_STATIC_FILES=true \
+ARG UID=991
+ARG GID=991
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -70,10 +72,11 @@ RUN bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-in
 
 COPY . /mastodon
```
moritzheiber: You can now [run chown while copying](https://docs.docker.com/engine/reference/builder/#copy), which is more efficient and saves you an additional run statement. I'd get rid of the `chown` statement in line 76 in favor of using `COPY --chown=${UID}:${GID}`?

moritzheiber:
Are we still using `docker-entrypoint.sh` after this PR? Can we get rid of it?

moritzheiber (commented):
Are we still using `docker-entrypoint.sh` after this PR? Can we get rid of it?

Dockerfile
Wonderfall: We could give it a try, indeed.

PR #6497, opened 2018-02-18
Fix prev/next links on public profile page

### LINKED ISSUE #4950: Link to "previous" toots missing when browsing publicly accessible profile page
Hello all,

this concerns v1.6.0 as it is currently deployed on hex.bz, I am not sure about other instances. I have attached the following screenshot to illustrate my issue:

screenshot from 2017-09-14 19-11-18

It is taken from https://hex.bz/@fce?max_id=2014079.

In this example I have already clicked the 'next' button more than once, so I would expect a link back to the previous page on the left-hand side, but I do not see it there. Please add this. :)

Greetings,
Fabian Schmidt

* * * *

valrus:
This fixes #4950 by adding a "prev" link on the public profile page after the "next" link is clicked.

I changed the logic a bit as well. Previously the "next" link would show up whenever there were any statuses on the current page, leading to a "There's nothing here" page if there were no further-back statuses to retrieve. Now it doesn't show up once you've hit the end. Both new behaviors are demonstrated in this screenshot:
pagination_fix

This is my first PR! 🎉  I'm happy to scale back my fix if the changes beyond what was strictly necessary to fix #4950 are deemed inappropriate here, and please let me know if this is a case that would benefit from a test! I like writing tests when possible but in my experience they're not usually that helpful for UI stuff like this.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -18,7 +18,8 @@ class AccountsController < ApplicationController
         @pinned_statuses = cache_collection(@account.pinned_statuses, Status) if show_pinned_statuses?
         @statuses        = filtered_statuses.paginate_by_max_id(20, params[:max_id], params[:since_id])
         @statuses        = cache_collection(@statuses, Status)
-        @next_url        = next_url unless @statuses.empty?
+        @next_url        = next_url if @statuses.last.id > filtered_statuses.last.id
+        @prev_url        = prev_url if @statuses.first.id < filtered_statuses.first.id
       end
 
       format.atom do
@@ -70,12 +71,20 @@ class AccountsController < ApplicationController
   end
 
   def next_url
+    pagination_url(max_id: @statuses.last.id)
+  end
+
+  def prev_url
+    pagination_url(since_id: @statuses.first.id)
+  end
+
+  def pagination_url(max_id: nil, since_id: nil)
     if media_requested?
-      short_account_media_url(@account, max_id: @statuses.last.id)
+      short_account_media_url(@account, max_id: max_id, since_id: since_id)
     elsif replies_requested?
-      short_account_with_replies_url(@account, max_id: @statuses.last.id)
+      short_account_with_replies_url(@account, max_id: max_id, since_id: since_id)
     else
-      short_account_url(@account, max_id: @statuses.last.id)
+      short_account_url(@account, max_id: max_id, since_id: since_id)
     end
   end
 
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AccountsController < ApplicationController
+  PAGE_SIZE = 20
+
   include AccountControllerConcern
 
   before_action :set_cache_headers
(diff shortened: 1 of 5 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -18,7 +18,8 @@ def show
         @pinned_statuses = cache_collection(@account.pinned_statuses, Status) if show_pinned_statuses?
         @statuses        = filtered_statuses.paginate_by_max_id(20, params[:max_id], params[:since_id])
         @statuses        = cache_collection(@statuses, Status)
-        @next_url        = next_url unless @statuses.empty?
+        @next_url        = next_url if @statuses.last.id > filtered_statuses.last.id
+        @prev_url        = prev_url if @statuses.first.id < filtered_statuses.first.id
```
ykzts:
```ruby
unless @statuses.empty
  @next_url = next_url if @statuses.last.id > filtered_statuses.last.id
  @prev_url  = prev_url if @statuses.first.id < filtered_statuses.first.id
end
```

ykzts (dismissed):

app/views/accounts/show.html.haml
```diff
@@ -39,6 +39,9 @@
 
       = render partial: 'stream_entries/status', collection: @statuses, as: :status
 
-  - if @statuses.size == 20
+  - if @next_url || @prev_url
     .pagination
-      = link_to safe_join([t('pagination.next'), fa_icon('chevron-right')], ' '), @next_url, class: 'next', rel: 'next'
+      - if @prev_url
+        = link_to safe_join([t('pagination.prev'), fa_icon('chevron-left')], ' '), @prev_url, class: 'prev', rel: 'previous'
```
ykzts: `safe_join([fa_icon('chevron-left'), t('pagination.prev')], ' ')`

app/controllers/accounts_controller.rb
```diff
@@ -70,12 +73,20 @@ def set_account
   end
 
   def next_url
+    pagination_url(max_id: @statuses.last.id)
+  end
+
+  def prev_url
+    pagination_url(since_id: @statuses.first.id)
```
Gargron: The reason there's no previous button is because since_id doesn't do what a previous button should do. since_id will display the most recent items potentially cut off by the ID, not the items *immediately before* the ID.

Cassolotl:
Prevous/next is a thing that causes confusion on some Tumblr layouts, and the solution is for those to be called older/newer instead. Can we do that here too?

valrus:
@Cassolotl: Absolutely, if and when I figure out how to address the _crippling flaw_ pointed out by @Gargron.

The solutions I've thought of so far seem inelegant but I do still intend to implement something here.

valrus:
Finally got around to a second pass at this, which works around the problem that `since_id` doesn't do what we want. I made a new pagination scope for this, with a comment explaining the purpose and distinction.

As well, I moved to "older/newer" per @Cassolotl's request. There were no localization strings for those so I added them just for English, but if an automatic translation would suffice I can add them for other languages too. (If there are any tools for adding localizations, please let me know.)

With the move to "older"/"newer" I thought it made more sense to have "older" on the left and "newer" on the right. All told, it now looks like this (with page size at 3 for tidiness, but it's still 20 in the PR):
masto_new_paging

Gargron (approved):

PR #6510, opened 2018-02-19
Add a possibility to skip chown.

### LINKED ISSUE #3194: Permission update on docker entrypoint takes a long time

pierreozoux:
fixes #3194 or at least workaround for bigger instalations


Gargron:
Resolved by #6514

PR #6515, opened 2018-02-20
Fix avatar and header issues by using custom geometry detector

Gargron:
Revert a part of #6508. The file passed to dynamic styles method was not actually a file, but an instance of Paperclip::Attachment, which broke all styles by always returning {} from the method.

One problem with GIF avatars was that Paperclip::GeometryDetector reported wrong dimensions for them, e.g. 120x120 GIF avatar would for some reason be detected as 120x53. By writing our own geometry parser, we can use FastImage (a pure Ruby gem), which also happens to be faster than
ImageMagick, to detect image dimensions, which are also correct.

Paperclip::LazyThumbnail will only invoke `convert` if current/target dimensions/format mismatch. Mind that this means `-strip` won't be applied to those files, but I think that's a minor side effect.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ gem 'omniauth', '~> 1.2'
 
 gem 'doorkeeper', '~> 4.2'
 gem 'fast_blank', '~> 1.0'
+gem 'fastimage'
 gem 'goldfinger', '~> 2.1'
 gem 'hiredis', '~> 0.6'
 gem 'redis-namespace', '~> 1.5'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -185,6 +185,7 @@ GEM
     faraday (0.14.0)
       multipart-post (>= 1.2, < 3)
     fast_blank (1.0.0)
+    fastimage (2.1.1)
     ffi (1.9.18)
     fog-core (1.45.0)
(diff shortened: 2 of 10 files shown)


akihikodaki:
I found it is a temporary and insufficient fix after some investigations.

The bug was not in resize operation, which this change avoids, but in the GIF optimizer. The output of the buggy GIF optimizer can vary not only by resize operations, but also any other unknown factors. It may work as fix in some environments, but may not work in others.

I have submitted a fix and a new version of ImageMagick with it will be available in one month or so, and its AppImage, which works just by downloading a binary, will be available via the CI as well.
https://github.com/ImageMagick/ImageMagick/pull/993

So I suggest to wait for a month or so and then instruct to use the AppImage. However, it may be not acceptable because of one of the following reasons:
* We need a fix as soon as possible.
* We don't want to mess the environments. (Obviously, the `PATH` environment variable must be updated to use the AppImage, the administrator should not forget to check the latest ImageMagick by himself, and he should revert the variable when the fix is available via a package manager.)

In such a case, the optimizer can be disabled by editing `lib/paperclip/thumbnail.rb` in Paperclip.

beatrix-bitrot (approved):
let's have a try~~

PR #6520, opened 2018-02-21
Focal points

Gargron:
**Why:** Different apps on different devices have different requirements for displaying media attachments, which means it's not good to crop server-side images to any particular ratio. However, users should be able to select which part of the uploaded image is most important to be shown in any crop. Therefore: **focal points**

- A focal point can be selected before posting status
- Focal points are part of  the `meta` property of media attachments in the API
- Web UI respects focal points when cropping
- ActivityPub federates focal points with `focalPoint` property on `Image`s


ykzts (approved):

jeroenpraat:
Note that in the UI it's called 'Crop' (already translated in various languages). This is a wrong term and confusing, cause people expect that the image can be cropped. IMO it should be 'Focal' (from Focal points).

Gargron:
I'm worried "focal" will not be understood at all. At least with "crop" you get a rough idea of what the button will do (something related to image dimensions)

jeroenpraat:
And 'Focus'? Cause Crop is really something different.

sundevour:
consider Highlight or Preserve as a more accurate one-word descriptor of what it does to part of the image?

trwnh:
'Focus' is fine and accurate (quite literal). 'Crop' may not be 100% accurate, but it is easier to understand semantically. I do think it is possible to confuse 'Crop' with a function to actually crop your image (and might even be appropriate if Mastodon ever adds a more feature-full image editor).

Cassolotl:
"Focus"  is good and clear, I think.

jeroenpraat:
@Gargron I would like that you rethink the wording here. Multiple people here prefer 'focus' and 'Highlight' is also an option. But I think Focus is better, cause it will be better translatable (actually in many languages it doesn't have to be translated at all).

Cassolotl:
I agree that "focal" is not clear - but "focus" is both a noun and a verb, and commonly used in English to mean what it's meant to mean here. "What's the focus of this picture? Which part am I meant to focus on?"

"Crop" doesn't make sense here because that describes a very specific widely-known function, and when the focal point thing does something different folks are going to get confused.

PR #6522, opened 2018-02-21
Show target status on audit log page

abcang:
Changed to show the target status on the audit log page.

image


ykzts (approved):

Gargron:
I'm sorry, I just think that's too massive? All log entries should preferably be close to the same height. I'm also not sure it's truly necessary. :pensive:

abcang:
I see. Close this issue.

PR #6528, opened 2018-02-21
Only store redirect location if not in JSON format

### LINKED ISSUE #6526: Manifest.json
Hello!

I’m not sure what the issue is, or how to fix it, but once someone logs into my instance, they are sent to this address https://noff.co/manifest.json instead of https://noff.co/web/getting-started.

If a user goes to the instance address /after/ logging in, then they are sent to the correct page. Any advise?

Gargron:
Fix #6526

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -35,7 +35,7 @@ class ApplicationController < ActionController::Base
   end
 
   def store_current_location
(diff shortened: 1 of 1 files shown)


malicioustoker:
I can confirm that this change fixes the issue!

OccultWarlock (approved):
This change fixed the issue!

Issue #6532, opened 2018-02-22
Force authentication strategy (e.g. SAML, CAS, LDAP)


ineffyble:
It would be useful if there was a way to force all users to register/login with a specified authentication method (e.g. SAML).

I imagine in many "enterprisey" contexts, all users will be using their SSO account. Giving them an option to login with a username/password they haven't set may confuse them (let alone if they try to "reset password").

I would see this as:

- New variable to set a forced/locked authentication strategy
- Either only present this button on the login form, or do away with it altogether and redirect straight away as if the user had pressed the button

* * * *

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #6540, opened 2018-02-22
New variable OAUTH_REDIRECT_AT_SIGN_IN + Ref #6538

gloaec:
I think that 99% of SSO users would rather be redirected to their IDP at login than having to click on an ugly button "SAML" or "CAS".

(+ fixed previous PR #6538 with more generic approach )


Gargron (approved):

Gargron:
Conflict?

gloaec:
@Gargron Ya, bad timing pulling from master before you merged, i'll fix tomorrow

ineffyble:
Fixes #6532 👍

PR #6543, opened 2018-02-23
Better grid layout for the landing page

lynlynlynx:
Better layout for the new landing page #6486 .
This uses grid layout instead of flexbox.

You can check how this works on [my instance's landing page](https://md.cryo.jp/about)

With the timeline view | Without the timeline view
--|--
screencapture-md-cryo-jp-about-1519388030642 | screencapture-md-cryo-jp-about-1519388078529
screencapture-md-cryo-jp-about-1519388039992 | screencapture-md-cryo-jp-about-1519388085403
screencapture-md-cryo-jp-about-1519388050737 | screencapture-md-cryo-jp-about-1519388097725

### DIFF
--- a/app/javascript/styles/mastodon/about.scss
+++ b/app/javascript/styles/mastodon/about.scss
@@ -22,7 +22,7 @@ $small-breakpoint: 960px;
     display: grid;
     grid-gap: 10px;
     grid-template-columns: 1fr 2fr;
-    grid-auto-columns: 1fr;
+    grid-auto-columns: 25%;
     grid-auto-rows: max-content;
 
     .column-0 {
@@ -53,7 +53,7 @@ $small-breakpoint: 960px;
   @media screen and (max-width: $small-breakpoint) {
 
     .grid {
-      grid-template-columns: 4fr 6fr;
+      grid-template-columns: 40% 60%;
 
       .column-0 {
         display: none;
--- a/app/javascript/styles/mastodon/about.scss
+++ b/app/javascript/styles/mastodon/about.scss
@@ -15,117 +15,172 @@ $small-breakpoint: 960px;
   }
 }
 
-.show-xs,
-.show-sm {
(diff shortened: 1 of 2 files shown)


Gargron:
When I opened your instance it looked like this: 

grafik

It looked correct before the timeline contents loaded, I think something inside the statuses is pushing that column wider.

Anyway, I'm not really sure why we would need to switch to CSS grids when it already worked fine. The change that improved the layout when timeline is disabled is valuable, but I'm not sure CSS grids are necessary..?

lynlynlynx:
Thanks for reporting the bug. I was forgetting Edge behave peculiarly. I'll fix that.

Why grid?: I'ts because, the new landing page has grid design. CSS grid layout can make(coding) grid design simplest and easiest. We can design easily and flexibly for with-timeline views and non-timeline views, on any number of columns.

Gargron:
@lynlynlynx Just FYI I am on Firefox, not Edge

lynlynlynx:
Wider column is fixed! I set the width as explicit percentage. Please check how this works again, thanks @Gargron .

Gargron:
@lynlynlynx Last remaining bug is this:

grafik

Flex-basis of 80px is too small. For example Pawoo has 12,731,979 statuses, that's a long number. I think that area could remain the same as before this PR?

lynlynlynx:
Thanks. Okay, I revert settings for information board as before.
Information board has overflowing bugs when it has large number, but I can't find the best way now... I'll pursue a better way and make new pull request.

lynlynlynx:
And, I set `align-items: flex-end;` to align the numbers of information board.
Please wait for updating commit,  I afk for few hours 🚶 💻

lynlynlynx:
Information board is completely fixed! Thanks for your cooperation!

Gargron (approved):

PR #6546, opened 2018-02-23
Specify AGPLv3+

TrashMacNugget:
Since the documentation doesn't specify you can use Mastodon as AGPLv3 or any later version.

> If the Program does not specify a version number of the GNU Affero General Public License, you may choose any version ever published by the Free Software Foundation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mastodon",
-  "license": "AGPL-3.0",
+  "license": "AGPL-3.0+",
   "engines": {
     "node": ">=6"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mastodon",
-  "license": "AGPL-3.0+",
+  "license": "AGPL-3.0-or-later",
   "engines": {
     "node": ">=6"
   },


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mastodon",
-  "license": "AGPL-3.0",
+  "license": "AGPL-3.0+",
```
ykzts: `AGPL-3.0-or-later` ([SPDX License List | Software Package Data Exchange (SPDX)](https://spdx.org/licenses/)

ykzts (changes requested):

Gargron:
But why is this needed?

TrashMacNugget:
@Gargron
Better license compatibility. If in the future there is some library using (A)GPLv4 that Mastodon wants to use, or some (A)GPLv4 project wants to use some Mastodon code, there would be no incompatibilities. Since it is not specified, that is currently the way it is now, but it is better to specify so there is no legal uncertainty.

akihikodaki:
An alternative option is to have a license notice to specify only version 3. (The license itself recommends to have one on all files, but I don't think it is mandatory. Writing the intension to `LICENSE` or `README.md` may also be enough.)
Such a notice should be helpful for future authors of softwares licensed with some later version.

TrashMacNugget:
@akihikodaki I think it would be much better to just leave it as-is, since the project already is licensed as AGPLv3 or (at your option) any later version.

ykzts (approved):

PR #6551, opened 2018-02-25
Visually distinguish threads in conversation

### LINKED ISSUE #279: Visually distinguish threads in single-toot reply view
Currently, replies to a toot are organized by thread, so one complete chain of replies is above the next-oldest reply to the original toot, and so on. It would be useful to have a visual indicator that defines and separates each thread, at least to the point where reply chains are walled off from each other.

Gargron:
Fix #279 

image

Code adopted from Objective-C written by @dagagren

Current limitations:

- No pretty turns, only vertical lines
- Temporary and small color palette
- If there is a lot of threads, the lines can fill up the entire horizontal space of the toot


trwnh:
Lines alone looks a bit confusing. Perhaps a small horizontal indent might help with visually parsing that kind of structure?

Issue #4890, opened 2017-09-11
Feature Request: notifications for all toots of particular users


wxl:
Birbsite has a notification option for "accounts you follow whose Tweets or live videos you don’t want to miss." [(link)][1] I certainly have some accounts I follow that aren't heavy users and I worry sometimes I lose their toots among the more chatty, so a similar feature would be great. 

[1]: [https://support.twitter.com/articles/20169887]

rainbowmultiverse:
This would need to be opt-in from the end of the person you're subscribing to. Some people don't want to be watched that closely, even if they aren't worried about stalkers and harassers.

wxl:
If someone's worried about stalkers and harassers, it would be
reasonable to assume they would have a locked account because they would
want to limit who is capable of following them. Assumedly, they wouldn't
mind that someone who they have accepted as a follower could follow them
more closely. If they don't have a locked account, they're certainly not
doing everything they can to protect themselves, even without the
feature requested. I'm not sure it's reasonable to make the opt-in a
blocker to this request to protect the people that aren't taking
advantage of the software. I do think the opt-in would be a great
additional feature, but not a blocker. Don't you think?

thiagomgd:
Also, if someone has an open profile, you can just use their rss feed and use it with ifttt to receive alerts via any other system. Or even, a stalker could create a fake profile and follow just that person.

So I agree with @wxl  on this.

Cassolotl:
>if someone has an open profile, you can just use their rss feed and use it with ifttt to receive alerts via any other system. Or even, a stalker could create a fake profile and follow just that person.

I don't think "someone with non-basic abilities can get around this feature, so there is no point having this feature" is a solid argument. That could be used to object to the ability to block people, for example - I've had to block someone who followed me from nine different accounts already. It's the same argument as "we can't guarantee that our toots are deleted from federated instances, therefore a 'delete account' feature is pointless".

>If someone's worried about stalkers and harassers, it would be reasonable to assume they would have a locked account because they would want to limit who is capable of following them.

I don't think someone should have to lock down and make private their entire account in order to minimise harassment. We already have tools and features in place to minimise harassment for people with public accounts. Just because we are posting publicly, doesn't mean we shouldn't make things difficult for harassers. It's like saying there's no point having "followers only" toot privacy for people who don't require follow-requests.

I think whether or not this feature should be opt-in and how it intersects with harassment and stalking should be a discussion to have with people who have experienced serious online harassment.

[...]

PR #6555, opened 2018-02-25
Make it possible to request notification about new posts

### LINKED ISSUE #6437: Feature Request: notification about all new toots
This is not what a regular user wants by default, but could be useful in a few cases. For example, one can create a secondary account with this feature enabled to follow the loved (and not too chatty) ones to be well-informed.

You can see this request as a variant of #4890 but it's more generic. On the other hand a bit hackish to maintain a secondary account, but it's not so rare even on Twitter IMO. This feature could be a good temporary solution until a very (or over-)sophisticated notification system will come to reality.

(By the way the official Twitter app also supports this.)

rpallai:
For this pretty straightforward feature we have to store Notification#type explicitly as it does not derive obviously from activity_type any more, because of both reblogs and posts has activity_type = Status.

Closes #6437


app/models/notification.rb
```diff
@@ -6,22 +6,25 @@
...
 #  activity_type   :string
+#  type            :string           not null
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #  account_id      :integer
 #  from_account_id :integer
 #
 
 class Notification < ApplicationRecord
+  self.inheritance_column = 'sti_type'
```
Gargron: nil?

app/models/notification.rb
```diff
@@ -37,19 +40,20 @@ class Notification < ApplicationRecord
...
-    where(activity_type: types)
+    types = TYPES - (exclude_types.map(&:to_sym) + [:follow_request])
+    where(type: types)
   }
 
   cache_associated :from_account, status: STATUS_INCLUDES, mention: [status: STATUS_INCLUDES], favourite: [:account, status: STATUS_INCLUDES], follow: :account
 
   def type
-    @type ||= TYPE_CLASS_MAP.invert[activity_type].to_sym
+    @type ||= self[:type].to_sym
```
Gargron: `super.to_sym` prooobably

app/views/notification_mailer/post.html.haml
```diff
@@ -0,0 +1,45 @@
...
+                    .col-6
+                      %table.column{ cellspacing: 0, cellpadding: 0 }
+                        %tbody
+                          %tr
+                            %td.column-cell.text-center.padded
+                              %table.hero-icon{ align: 'center', cellspacing: 0, cellpadding: 0 }
+                                %tbody
+                                  %tr
+                                    %td
+                                      = image_tag full_pack_url('icon_cached.png'), alt: ''
```
Gargron: That's the icon for a reblog

db/migrate/20180225105236_add_type_to_notifications.rb
```diff
@@ -0,0 +1,15 @@
+class AddTypeToNotifications < ActiveRecord::Migration[5.1]
+  def up
+    add_column :notifications, :type, :string
+    Notification.where(activity_type: 'Mention').update_all(type: 'mention')
```
Gargron: These are going to take a long time. I think there is an `in_batches` variant so it doesn't lock writes to the table as much

Gargron (changes requested):
Seems mostly fine

Gargron:
About the icon: Did you take it from Material Icons? That's where all other e-mail icons are from (except the reblog which I had to take from font-awesome), and the SVG variants are saved in the repository as well, in case the PNG ever needs to be regenerated.

rpallai:
Material Icons looks good, changing in progress..

app/lib/feed_manager.rb
```diff
@@ -30,6 +30,7 @@ def push_to_home(account, status)
     return false unless add_to_feed(:home, account.id, status)
     trim(:home, account.id)
     PushUpdateWorker.perform_async(account.id, status.id, "timeline:#{account.id}") if push_update_required?("timeline:#{account.id}")
+    NotifyService.new.call(account, status, :post) unless status.reblog? && status.reblog.account == account
```
Gargron: Wait, so you just duplicate the entire home feed in the notifications? I thought the idea was to opt-in to get notifications like that from *particular users*, like those who post rarely.

This won't work, because it would explode storage requirements (notifications table is a lot spacier than redis feeds). Also, in any case this feels like something that should happen in FanOutOnWriteService rather here.

Gargron: Okay, I confused #6437 and #4890. But even in this case, it's too much work to save home feed items as notifications, unless they need to be shown in the notifications column, which I don't think they do. Currently the code is written so you can send a push notification/e-mail from a Notification object but I think it would be easier to just add special cases for that. After all, a one-off e-mail/push notification does not require a corresponding notification record.

rpallai: I'll rework this patch into a new PR based on you suggestions as my goal is just to send one-off notifications about all new posts without keeping a history. This is the wrong approach for sure.

Anyway, this work became a good base for #4890, maybe I'll pick that too.

Thanks for the review, I'll be back. :)

PR #6557, opened 2018-02-25
Ensure the app does not even start if OTP_SECRET is not set

Gargron:


app/models/import.rb
```diff
@@ -26,7 +26,7 @@ class Import < ApplicationRecord
 
   validates :type, presence: true
 
-  has_attached_file :data, url: '/system/:hash.:extension', hash_secret: ENV['PAPERCLIP_SECRET']
```
ykzts: note:

> Remove PAPERCLIP_SECRET (it's not used by anything, actually)
> 
> Imports are for internal consumption and the url option isn't even
> used correctly, so we can remove the hash stuff from them

ykzts (approved):

PR #6559, opened 2018-02-25
Improve public account cards

### LINKED ISSUE #6467: Follow buttons on another user's list of following/followers as with tw*
Currently it's hard to populate your account's feed from the following/followers lists of another user by pressing "follow" buttons -- mainly because they don't exist. 

So currently you have to navigate to profile pages individually and click follow from there. 

Could there be a follow button for each box per user on these list pages? 

Thanks in advance! 

@xeb@mastodon.social

Gargron:
- Add follow/unfollow/remote follow buttons
- Format the bio properly
- Always show username@domain, even for local accounts

Fix #6467

### DIFF
--- a/app/controllers/follower_accounts_controller.rb
+++ b/app/controllers/follower_accounts_controller.rb
@@ -7,7 +7,9 @@ class FollowerAccountsController < ApplicationController
     @follows = Follow.where(target_account: @account).recent.page(params[:page]).per(FOLLOW_PER_PAGE).preload(:account)
 
     respond_to do |format|
-      format.html
+      format.html do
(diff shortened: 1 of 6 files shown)


Sylvhem:
Looks like a great improvement!
But are you sure you want to always show username@domain? Personally, I use this to see at the first glance which accounts are local and which ones are not.

Gargron:
@Sylvhem Yes, because I don't want to have to keep in mind which URL i'm viewing them at, I want to know at a glance who's from where.

Sylvhem:
@Gargron Seems logical.

ykzts (approved):

PR #6570, opened 2018-02-28
Federated reports

### LINKED ISSUE #2176: Federated Reports
Currently, the only solution for reporting someone on another instance is to either sign up for that instance and report them from there, or to @ the admin. This is clunky.
Reports should be federated across instances, but admins should be able to block specific instances or reporters. 
This could be done by having the instance send a UUID based on the user's full handle (as well as a salt, so bruteforcing is not possible) so they have a unique identifier, but bad admins cannot find users who are reporting bad behavior and harass them.
Thanks!

* * * *

Gargron:
Fix #2176

screen shot 2018-02-28 at 05 20 11

### DIFF
--- a/app/controllers/api/v1/reports_controller.rb
+++ b/app/controllers/api/v1/reports_controller.rb
@@ -13,13 +13,13 @@ class Api::V1::ReportsController < Api::BaseController
   end
 
   def create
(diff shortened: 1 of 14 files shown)


ykzts (approved):
looks gooooooood!

PR #6572, opened 2018-02-28
Add "Toots/Toots with replies/Media" tab below profile header

Gargron:
image


ykzts (approved):
:heart:

beatrix-bitrot:
good feature. the styling of the options leaves something to be desired though i think. looks a little weird with no dividers or tabs, and with them all mushed to the left.

PR #6580, opened 2018-03-01

akihikodaki:
This is same with #3754, but has an important fix regrading https://github.com/tootsuite/mastodon/pull/3754#issuecomment-333878071. (UPDATE: and my reply for the comment had a misconception, https://github.com/tootsuite/mastodon/pull/3754#issuecomment-333879192.)
`config.i18n.fallbacks` was intended to avoid such a problem, but it was misconfigured in the prior pull request. It is fixed in this pull request.

### DIFF
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -17,11 +17,7 @@ module Localized
   end
 
   def default_locale
-    request_locale || env_locale || I18n.default_locale
-  end
-
-  def env_locale
-    ENV['DEFAULT_LOCALE']
+    request_locale || I18n.default_locale
   end
 
   def request_locale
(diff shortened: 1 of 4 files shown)


Gargron:
I'm still not sure why you want to touch `I18n.default_locale` in any way? We set the user's locale when needed, and already have the logic for a fallback chain, I don't see a benefit in using `I18n.default_locale` over `ENV['DEFAULT_LOCALE']`

akihikodaki:
@Gargron Because it allows to set the default locale all over the application, including email, where `DEFAULT_LOCALE` is not used. That was intended in #3754. It would also eliminate duplicates of the same logic of `ENV['DEFAULT_LOCALE'] || I18n.default_locale`.

Gargron (approved):

PR #6583, opened 2018-03-01
Add license info to README

TrashMacNugget:


akihikodaki:
I would like to know your intension.

TrashMacNugget:
The AGPL recommends adding a notice letting you know what license it is under. This helps since the project is AGPLv3+, but if there is no notice it is not clear if it is AGPLv3+ or just AGPLv3.

akihikodaki (approved):
It sounds good idea to me.

PR #6590, opened 2018-03-02
Responsively enforce 16:9 ratio on all media thumbnails in web UI

Gargron:
Thumbnails in web UI are meant to be 16:9 and are calibrated for the desktop column width, but mobile devices vary in width so for some of the wider ones, like the iPhone, the thumbnails are all a bit too wide & short.

This uses the mostly already captured width state to calculate a responsive height.

Also change video player behaviour to "contain" rather than "cover" videos that don't fit the ratio, unlike images and GIFs, it's expected that a video is shown fully.

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -284,7 +284,7 @@ export default class MediaGallery extends React.PureComponent {
         style.height = width / this.props.media.getIn([0, 'meta', 'small', 'aspect']);
       }
     } else if (width) {
-      style.height = Math.floor(width / (16/9));
+      style.height = width / (16/9);
     } else {
       style.height = height;
     }
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -283,8 +283,9 @@ export default class MediaGallery extends React.PureComponent {
       if (width) {
         style.height = width / this.props.media.getIn([0, 'meta', 'small', 'aspect']);
       }
+    } else if (width) {
+      style.height = Math.floor(width / (16/9));
     } else {
-      // crop the image
       style.height = height;
     }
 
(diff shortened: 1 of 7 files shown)


app/javascript/mastodon/components/media_gallery.js
```diff
@@ -283,8 +283,9 @@ export default class MediaGallery extends React.PureComponent {
       if (width) {
         style.height = width / this.props.media.getIn([0, 'meta', 'small', 'aspect']);
       }
+    } else if (width) {
+      style.height = Math.floor(width / (16/9));
```
akihikodaki: Why `floor`?

app/javascript/mastodon/features/video/index.js
```diff
@@ -246,12 +254,23 @@ export default class Video extends React.PureComponent {
...
+    const { preview, src, inline, startTime, onOpenVideo, onCloseVideo, intl, alt, detailed } = this.props;
+    const { containerWidth, currentTime, duration, buffer, dragging, paused, fullscreen, hovered, muted, revealed } = this.state;
     const progress = (currentTime / duration) * 100;
+    const playerStyle = {};
+
+    let { width, height } = this.props;
+
+    if (inline && containerWidth) {
+      width       = containerWidth;
+      height      = Math.floor(containerWidth / (16/9));
```
akihikodaki: Redundant spaces.

akihikodaki (changes requested):

app/javascript/mastodon/components/media_gallery.js
Gargron: Do floating point pixels matter in height?

akihikodaki: I don't think it practically matters, even if they get rendered. But I don't think `floor` makes difference, either.

akihikodaki (approved):

PR #6593, opened 2018-03-02
When search enabled, display hint in search popout

Gargron:
Also, switch "getting started" icon in tabs bar (display in mobile layout) from asterisk to hamburger, to make it more obvious it acts as a menu.


akihikodaki (approved):

m4sk1n:
Is it just for me that the string isn't correct?

PR #6594, opened 2018-03-02
Move "compose" on mobile to floating action button

Gargron:
screen shot 2018-03-02 at 05 52 31

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -1878,7 +1878,7 @@
   font-size: 14px;
   font-weight: 500;
   border-bottom: 2px solid lighten($ui-base-color, 8%);
-  transition: all 200ms linear;
+  transition: all 50ms linear;
 
   .fa {
     font-weight: 400;
@@ -1895,7 +1895,6 @@
   &:active {
     @media screen and (min-width: 631px) {
       background: lighten($ui-base-color, 14%);
-      transition: all 100ms linear;
     }
   }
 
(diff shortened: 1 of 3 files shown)


akihikodaki (approved):
I am a bit surprised seeing you make this change after versioning RC.

Gargron:
@akihikodaki Until a git tag exists, the version is not immutable/finalized. Sometimes I bump it in anticipation...

app/javascript/styles/mastodon/components.scss
```diff
@@ -5017,3 +5016,27 @@ noscript {
...
+
+.floating-action-button {
+  position: fixed;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 3.9375rem;
+  height: 3.9375rem;
+  bottom: 1.3125rem;
+  right: 1.3125rem;
```
unarist: @Gargron Why did you use rem which we haven't used before?

zcdunn:
So where would the search field be with this change?

Gargron:
@zcdunn Search is still on the compose page, nothing changed about that.

app/javascript/styles/mastodon/components.scss
Gargron: I looked up how the floating action button was positioned in another Material UI application :thinking: I figured the button needs to be easily clickable so I shouldn't mess with these values.

zcdunn:
So you can't swipe between home and compose columns now? It seems pretty weird to have to hit a button with a pencil on it to search.

PR #6610, opened 2018-03-04
Federate pinned statuses over ActivityPub

Gargron:
- Send `Add`/`Remove` activities to followers on pin/unpin
- Pinned statuses are the "featured" collection (retrievable)
- When account cache expires normally, fire off async task to sync up pinned toots
- Display pinned toots in web UI


akihikodaki (changes requested):

Gargron:
@akihikodaki Addressed

akihikodaki (approved):

PR #6620, opened 2018-03-04
Replace "log in" and "log out" with "sign in" and "sign out"

akihikodaki:
A few messages used "sign in" where the others used "log in" and "log out".
This pull request resolves the inconsistency.

While "log in" and "log out" are more common, "sign in" and "sign out"
are chosen because they are consistent with "sign up". They are also used
in Devise, which Mastodon depends on much, and routes /auth/sign_in
and /auth/sign_out generated by the library.


Gargron:
I don't think this is worth the changes and resetting stats.

akihikodaki:
stats?

Gargron:
Yes you changed redis keys: `Redis.current.pfcount("activity:signins:#{week_id}").to_s`

akihikodaki:
I should have asked about "changes" as well. I want to ask about it because it would change how I solve the issue (updating messages saying "sign in", etc.)

Gargron:
I don't think 100 file changes churn is worth changing synonyms. "Sign in" and "Log in" and "Login" are interchangeable.

PR #3196, opened 2017-05-21
Implement hash tag completion

orekyuu:
This PR implements completion of hashtags.
The tooted hashtag is included in the candidate.
Candidates are stored localStorage.

PR #6621, opened 2018-03-04
Implement tag auto-completion by history

akihikodaki:
This is a functionality similar to one implemented in Pawoo:
https://github.com/pixiv/mastodon/commit/21a3c70f8083b1347d2b8420ed7001b78c2c9620

UPDATE: @unarist pointed out this is an alternative for #3196.

### DIFF
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -1,6 +1,7 @@
 import api from '../api';
 import { throttle } from 'lodash';
 import { search as emojiSearch } from '../features/emoji/emoji_mart_search_light';
+import { tagHistory } from '../settings';
 import { useEmoji } from './emojis';
 
 import {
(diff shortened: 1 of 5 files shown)


abcang:
I think that complementing by `/api/v1/search` is more convenient than complementing by history. Why is it compatible with Pawoo implementation?

akihikodaki:
> I think that complementing by /api/v1/search is more convenient than complementing by history. Why is it compatible with Pawoo implementation?

To be honest I do not always appreciate account suggestion much because it suggests lots of people but the number of people I often mention is quite a limited.

A tag will be tooted even more often than mentions for an account. Suggestions ordered by recent uses would be a better choice.

app/javascript/mastodon/actions/compose.js
```diff
@@ -322,6 +340,48 @@ export function selectComposeSuggestion(position, token, suggestion) {
...
+export function updateTagHistory(tags) {
+  return {
+    type: COMPOSE_TAG_HISTORY_UPDATE,
+    tags,
+  };
+}
+
+export function hydrateCompose() {
+  return (dispatch, getState) => {
+    const me = getState().getIn(['meta', 'me']);
```
abcang: `me` can be acquired by importing initial_state.

`import { me } from '../initial_state';`

abcang:
I understood. I agree broadly.

Complementing by using api can choose the correct tag. So, I thought it would be nice to implement it as another feature.

akihikodaki:
> Complementing by using api can choose the correct tag. So, I thought it would be nice to implement it as another feature.

What do you suggest to implement as another feature? To search tags on the instance, to use history, or to combine them?

abcang:
I believe it is best to combine tag search and history. But, I accepts this PR even without implements of tag search.

Gargron (approved):

PR #6623, opened 2018-03-04
Allow to customize hostname

akihikodaki:
While site_title setting was allowed to change to represent instance name,
domain was widely used instead as instance name. However, many instances
identifies themselves with names different from the domain.

This adds site_hostname setting which replaces the domain.


akihikodaki:
ac26665ef1d7e89bcd35f6ea859ddfc997efe2f5 resolves the prior CI failure.

Gargron:
In my opinion site_hostname is used in many places where a domain name is expected. I don't think we need to change anything about it. If someone wants to customize that much, they can just edit the code.

nynhex:
I don't get the use case either.  👎

akihikodaki:
Consider the case of Kirakiratter. Kirakiratter is named as Kirakiratter, not kirakiratter.com. Users would expect "email from Kirakiratter" instead of "email from kirakiratter.com". This pull request allows to change the display name like so.

This kind of naming is also common in email provider. Gmail uses `gmail.com` as its domain, but gmail.com is not the name of the provider.

nightpool:
shouldn't kirakiratter have that as their site title then?

akihikodaki:
`site_title` replaces Mastodon with the given text. `site_hostname` replaces the domain string with the given text.
The difference is clear in `title` of HTML. The current HTML title of Kirakiratter is:
> kirakiratter.com - Kirakiratter

I doubt the title is preferred because:
* It does not say it is Mastodon
* The word "Kirakiratter" is repeated twice.
* A browser also says its domain is `kirakiratter.com`. "kirakiratter.com" in title is redundant in this sense.

These problems can be solved by unsetting `site_title` and setting `site_hostname` to Kirakiratter.
> Kirakiratter - Mastodon

Gargron:
Kirakiratter is on something like v1.5 though and they have such heavy customizations they probably hardcoded Kirakiratter instead of Mastodon, rather than using site_title.

akihikodaki:
I'm not sure they do so, but the behavior is same even on upstream.

PR #6628, opened 2018-03-04
Add some UI for user-defined domain blocks

### LINKED ISSUE #5855: Domain blocks improvements on the user side
I've seen this feedback floating around on the fediverse recently and it's justified IMO:

While hiding a specific domain is fairly easy in the UI, there are no (or at least no easy way) to:

- know what domain you previously muted
- unmute said domains. Currently the only way to un-hide a domain is to go on a user's profile who is on the domain you've hidden and click `Unhide <domain>`

I feel like domain hiding is a very powerful feature for users but the lack of proper UI related to it seems confusing at best and dangerous at worst.

* * * *

ClearlyClaire:
This is a work-in-progress attempt to fix #5855.

~~~Currently, it only lists the domains, without letting the user actually unblock them.~~~

~~~Fixing that will require quite some bit of refactoring, as the Web UI currently ties block/unblock requests to a single remote account (which also causes some discrepancies when viewing the profile of multiple users on the same remote instance: only one account is updated in the Web UI).~~~

### DIFF
--- a/app/javascript/mastodon/features/domain_blocks/index.js
+++ b/app/javascript/mastodon/features/domain_blocks/index.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import ImmutablePropTypes from 'react-immutable-proptypes';
+import PropTypes from 'prop-types';
+import LoadingIndicator from '../../components/loading_indicator';
+import Column from '../ui/components/column';
+import ColumnBackButtonSlim from '../../components/column_back_button_slim';
+import DomainContainer from '../../containers/domain_container';
+import { fetchDomainBlocks, expandDomainBlocks } from '../../actions/domain_blocks';
+import { defineMessages, injectIntl } from 'react-intl';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+import { debounce } from 'lodash';
+import ScrollableList from '../../components/scrollable_list';
+
+const messages = defineMessages({
+  heading: { id: 'column.domain_blocks', defaultMessage: 'Hidden domains' },
+  unblockDomain: { id: 'account.unblock_domain', defaultMessage: 'Unhide {domain}' },
+});
+
+const mapStateToProps = state => ({
+  domains: state.getIn(['domain_lists', 'blocks', 'items']),
+});
+
+@connect(mapStateToProps)
+@injectIntl
+export default class Blocks extends ImmutablePureComponent {
+
+  static propTypes = {
+    params: PropTypes.object.isRequired,
+    dispatch: PropTypes.func.isRequired,
+    domains: ImmutablePropTypes.list,
+    intl: PropTypes.object.isRequired,
+  };
+
+  componentWillMount () {
+    this.props.dispatch(fetchDomainBlocks());
+  }
+
+  handleLoadMore = debounce(() => {
+    this.props.dispatch(expandDomainBlocks());
+  }, 300, { leading: true });
+
+  render () {
+    const { intl, domains } = this.props;
+
+    if (!domains) {
+      return (
+        <Column>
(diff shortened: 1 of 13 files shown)


app/javascript/mastodon/reducers/domain_lists.js
```diff
@@ -0,0 +1,23 @@
...
+import { Map as ImmutableMap, List as ImmutableList } from 'immutable';
+
+const initialState = ImmutableMap({
+  blocks: ImmutableMap(),
+});
+
+export default function userLists(state = initialState, action) {
+  switch(action.type) {
+  case DOMAIN_BLOCKS_FETCH_SUCCESS:
+    return state.setIn(['blocks', 'items'], ImmutableList(action.domains)).setIn(['blocks', 'next'], action.next);
```
akihikodaki: What about using OrderedSet instead of List?

app/javascript/mastodon/features/domain_blocks/index.js
```diff
@@ -0,0 +1,71 @@
...
+          <LoadingIndicator />
+        </Column>
+      );
+    }
+
+    return (
+      <Column icon='ban' heading={intl.formatMessage(messages.heading)}>
+        <ColumnBackButtonSlim />
+        <ScrollContainer scrollKey='domain_blocks'>
+          <div className='scrollable' onScroll={this.handleScroll}>
```
akihikodaki: Why not `ScrollableList`?

app/javascript/mastodon/reducers/relationships.js
```diff
@@ -23,6 +23,14 @@ const normalizeRelationships = (state, relationships) => {
   return state;
 };
 
+const setDomainBlocking = (state, accounts, blocking) => {
+  accounts.forEach(id => {
+    state = state.setIn([id, 'domain_blocking'], blocking);
```
akihikodaki: What about using `withMutations`?

akihikodaki (dismissed):

app/javascript/mastodon/reducers/domain_lists.js
ThibG: Hm… good question, I initially copied how `user_lists` worked, but `OrderedSet` seems fine. I'll try using it.

app/javascript/mastodon/features/domain_blocks/index.js
ThibG: Same answer, I copied the `blocks` feature, which uses the same pattern. I'll look up `ScrollableList`.

app/javascript/mastodon/reducers/relationships.js
ThibG: Hm, I don't know `withMutations` and I'm not sure how it would make things better?

akihikodaki: `withMutations` makes an immutable object mutable in a given callback, and returns the muted object. It improves performance by removing redundant immutable object construction. And you won't have to write something like `state = state.setIn(...);`. Just `state.setIn(...);` will be fine.

app/javascript/mastodon/features/domain_blocks/index.js
ThibG: Hm, `ScrollableList` implements a lot more functionality that we need. I'm not completely sure I want to switch to that.

akihikodaki: It is preferred since it has `throttle`. Even though `throttle` may not be that important, it is unlikely to have such improvements which `ScrollableList` will have in future.

ThibG: Ok. Updated to use `ScrollableList`.

Gargron (approved):

PR #6631, opened 2018-03-05
Add media timeline

ykzts:
### screenshots

Toots | Media
-|-
Screenshot on Toots | Screenshot on Media

### DIFF
--- a/app/javascript/mastodon/actions/streaming.js
+++ b/app/javascript/mastodon/actions/streaming.js
@@ -40,10 +40,9 @@ const refreshHomeTimelineAndNotification = (dispatch, done) => {
   dispatch(expandHomeTimeline({}, () => dispatch(expandNotifications({}, done))));
 };
 
-export const connectUserStream = () => connectTimelineStream('home', 'user', refreshHomeTimelineAndNotification);
-export const connectCommunityStream = () => connectTimelineStream('community', 'public:local');
-export const connectMediaStream = () => connectTimelineStream('community', 'public:local');
-export const connectPublicStream = () => connectTimelineStream('public', 'public');
-export const connectHashtagStream = (tag) => connectTimelineStream(`hashtag:${tag}`, `hashtag&tag=${tag}`);
-export const connectDirectStream = () => connectTimelineStream('direct', 'direct');
-export const connectListStream = (id) => connectTimelineStream(`list:${id}`, `list&list=${id}`);
+export const connectUserStream      = () => connectTimelineStream('home', 'user', refreshHomeTimelineAndNotification);
+export const connectCommunityStream = ({ onlyMedia } = {}) => connectTimelineStream(`community${onlyMedia ? ':media' : ''}`, `public:local${onlyMedia ? ':media' : ''}`);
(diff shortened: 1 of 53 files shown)


akihikodaki:
I am not a fan of this change because:
* we already have several timelines: home, local, and federated. Why add even more?
* the navigation bar has too many buttons. As it gets more items, it looks more messy and each buttons will be smaller and hard to tap on smartphone.
* "media" timeline is confusing by nature. Does it list media in home, local, or federated?

The problems listed above may be resolved by adopting the UI similar to #6572.

ykzts:
demo

🤔

akihikodaki:
Well, the setting column requires extra steps to use the feature. It would be troublesome especially if a user wants to see both of federated timeline and media timeline since it requires to click more to switch them. So I'm also :thinking: .

By the way, the UI introduced by #6572, which I suggested earlier, does not have such a problem. I suspect you have a reason you still don't adopt #6572. The "only media" option in column setting and the media timeline is similar in terms that those preferences will be persistent when they are pinned. Is that what you intend? Or do you have reasons you prefer the media timeline or the "only media" option or ones you do not like #6572 for this purpose?

jeroenpraat:
Actually I like it, but I agree with @akihikodaki about the fact you have to click twice. This can be resolved with a toggle icon on the timelines titlebar (see images for an example).

Off (current situation):
media-timeline-off

On:
media-timeline-on

akihikodaki:
I think it is better. It would be nicer if it is colored when it is on since a colored icon often carries semantic meaning to show it is enabled.

Gargron:
>By the way, the UI introduced by #6572, which I suggested earlier, does not have such a problem

I agree with Aki that the tabs might be a good fit for this.

ykzts:
@akihikodaki @jeroenpraat @Gargron I updated (See screenshot of description).

akihikodaki (approved):

jebuskrust:
Greetings,

Although this is pretty neato... 👍 
When the local / federated timelines are pinned, it doesnt filter the pinned timelines :(

trwnh:
I have to say this change is a bit confusing and looks weird...

- Tabs take up too much vertical space
- Tab titles are confusing ("Toots" includes media as well)
- Only present in local/federated, not in home column

It would be much better to specify "only media" as an available option in all columns.

jeroenpraat:
I agree with both previous comments. I also think these tabs take to much space. I still advise to do it like the screenshots I provided or hide it in the column settings.

@Gargron Probably better to revert this PR and delay it to a .1 release.

Gargron:
Okay, fair enough.

SerCom-KC:
Still looks buggy to me after #7575...

Steps to reproduce:
1. Pin local/federated timeline when the filter set to "Toots"
2. Click "Back" button on the rightmost column (so it goes back to "Getting started")
3. Click "Media" on the pinned local/federated column
4. Click "Back" button on the rightmost column again

After step 3, the media timeline is displayed on the rightmost column instead of the pinned one
and then after step 4 the browser navigates to the previous history page instead of goes back to "Getting started"

ykzts:
@renatolond #7581

PR #6642, opened 2018-03-05
Added Travis configuration for building Docker images on for the Hub

moritzheiber:
This PR implements building Docker images for either `master` or release tags on TravisCI to work around the issue of the Docker Hub using an outdated version of Docker.

Before merging this PR @Gargron needs to run the following command inside project directory:

```sh
$ travis encrypt DOCKER_HUB_PASSWORD=<their-docker-hub-password> --add
```
and check-in the resulting `.travis.yml` for this PR, otherwise the pushes to the Hub will fail (since they require a valid login).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,59 +1,64 @@
 language: ruby
+sudo: required
+services:
+  - docker
 cache:
   bundler: true
   yarn: true
   directories:
-    - node_modules
-    - public/assets
-    - public/packs-test
-    - tmp/cache/babel-loader
+  - node_modules
+  - public/assets
+  - public/packs-test
+  - tmp/cache/babel-loader
 dist: trusty
 sudo: false
 branches:
   only:
-    - master
-
+  - master
+  - /^v\d+\.\d+(\.\d+)?(\S*)?(-\S*)?$/
 notifications:
   email: false
-
 env:
   global:
-    - LOCAL_DOMAIN=cb6e6126.ngrok.io
-    - LOCAL_HTTPS=true
-    - RAILS_ENV=test
-    - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
-    - PARALLEL_TEST_PROCESSORS=2
-
+  - LOCAL_DOMAIN=cb6e6126.ngrok.io
+  - LOCAL_HTTPS=true
+  - RAILS_ENV=test
+  - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+  - PARALLEL_TEST_PROCESSORS=2
+  - DOCKER_HUB_USERNAME=gargron
+before_install:
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y install docker-ce
 addons:
   postgresql: 9.4
   apt:
     sources:
-      - trusty-media
-      - sourceline: deb https://dl.yarnpkg.com/debian/ stable main
-        key_url: https://dl.yarnpkg.com/debian/pubkey.gpg
+    - trusty-media
+    - sourceline: deb https://dl.yarnpkg.com/debian/ stable main
+      key_url: https://dl.yarnpkg.com/debian/pubkey.gpg
     packages:
-      - ffmpeg
-      - libicu-dev
-      - libprotobuf-dev
-      - protobuf-compiler
-      - yarn
-
+    - ffmpeg
+    - libicu-dev
+    - libprotobuf-dev
+    - protobuf-compiler
+    - yarn
 rvm:
-  - 2.4.2
-  - 2.5.0
-
+- 2.4.2
+- 2.5.0
 services:
-  - redis-server
-
+- redis-server
 install:
-  - nvm install
-  - bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
-  - yarn install
-
+- nvm install
+- bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
+- yarn install
 before_script:
-  - ./bin/rails parallel:create parallel:load_schema parallel:prepare assets:precompile
-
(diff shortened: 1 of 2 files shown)


nightpool:
shouldnt gargron run that command and push the commit to this PR BEFORE
merging it?

moritzheiber:
@nightpool True, I completely forgot about commit contributions on PRs. It's been a long day :disappointed: 
@Gargron You can do this before merging the PR, obviously, as @nightpool suggested.

.travis.yml
```diff
@@ -1,59 +1,64 @@
...
+  - public/packs-test
+  - tmp/cache/babel-loader
 dist: trusty
 sudo: false
 branches:
   only:
-    - master
-
+  - master
+  - /^v\d+\.\d+(\.\d+)?(\S*)?(-\S*)?$/
```
ineffyble: Can you clarify what this regex is?

ineffyble (approved):
LGTM

.travis.yml
moritzheiber: It matches branches (yes, for some reason Travis detects tags as branches) with release criteria, i.e. "v2.3.0rc1", or "v.2.3.4-test".

.travis.yml
```diff
@@ -1,59 +1,64 @@
...
-  - bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
-  - yarn install
-
+- nvm install
+- bundle install --path=vendor/bundle --without development production --retry=3 --jobs=16
+- yarn install
 before_script:
-  - ./bin/rails parallel:create parallel:load_schema parallel:prepare assets:precompile
-
+- "./bin/rails parallel:create parallel:load_schema parallel:prepare assets:precompile"
```
akihikodaki: Redundant quotes.

deploy_hub.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+TAG="${TRAVIS_TAG:-latest}"
```
akihikodaki: `latest` is confusing. It sounds like the latest "release", but it is actually not release. I suggest just to use `TRAVIS_BRANCH`; a user should know that something from Mastodon named `master` is unstable.

deploy_hub.sh
```diff
@@ -0,0 +1,17 @@
...
+  docker logout
+}
+
+trap _logout EXIT TERM
+
+docker login -u "${DOCKER_HUB_USERNAME}" -p "${DOCKER_HUB_PASSWORD}"
+docker build -t "${DOCKER_HUB_USERNAME}/mastodon:${TAG}" .
+docker tag "${DOCKER_HUB_USERNAME}/mastodon:${TAG}" "${DOCKER_HUB_USERNAME}/mastodon:latest"
+docker push "${DOCKER_HUB_USERNAME}/mastodon:${TAG}"
+docker push "${DOCKER_HUB_USERNAME}/mastodon:latest"
```
akihikodaki: I do not think the last line is necessary.

akihikodaki (changes requested):

akihikodaki:
I wonder how long it takes since it builds `libiconv`.

.travis.yml
```diff
@@ -1,59 +1,64 @@
...
-    - node_modules
-    - public/assets
-    - public/packs-test
-    - tmp/cache/babel-loader
+  - node_modules
+  - public/assets
+  - public/packs-test
+  - tmp/cache/babel-loader
 dist: trusty
 sudo: false
```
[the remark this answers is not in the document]
moritzheiber: Right, well spotted. It worked without having removed it, but this can be considered obsolete. I'll fix it :+1:

.travis.yml
moritzheiber: This went through Travis' own linter (which is why some parts of the file have changed) and I'd rather like to leave it that way.

deploy_hub.sh
moritzheiber: I'm just trying to replicate current behavior. Currently `latest` points at the tip of master (although it hasn't been built in a while due to the bug this PR is trying to solve), as far as I can tell, hence the same behavior in here.

deploy_hub.sh
moritzheiber: It is if `TRAVIS_TAG` is actually set. You would want a tagged release to become `latest` as well. Also, this is a no-op in the daemon if it detects that the digest has already been pushed.

moritzheiber:
@akihikodaki Are you okay with the explanation I've given to your requests above? Also, the overhead for compiling libiconv is roughly 30 seconds. Seeing as the majority of our Travis runs are spend on installing system dependencies and running RSpec tests right now I wouldn't consider it a large overhead.

(Also, this isn't done for PRs, just for `master` and release tags)

deploy_hub.sh
ineffyble: `latest` has a tradition of being bleeding edge in any case - https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375

akihikodaki: > I'm just trying to replicate current behavior. Currently `latest` points at the tip of master (although it hasn't been built in a while due to the bug this PR is trying to solve), as far as I can tell, hence the same behavior in here.

I do not think there is any strict rule yet. It is a good chance to consider about it well.

> If latest should be bleeding edge (I'd be in favor of it) or not isn't for this PR to decide.

It is if there is no rule yet.

> `latest` has a tradition of being bleeding edge in any case

Bleeding edge "release", maybe? Indeed many Docker images are just for releases and it is hard to find examples of CI.

akihikodaki (approved):

akihikodaki:
I have concerns about `latest` but I think it is not that important to block merging this.

akihikodaki:
Well, is an explanation why it is rejected provided? I'm just curious.

moritzheiber:
@akihikodaki Hey, sorry! I told Eugen to close it because it was essentially superseded by #6662. Using the "native" Docker Hub environment, along with what Eugen has now declared to be `edge` and `latest`, makes sense, and we don't need to rely on Travis for building our images.

I hope that clears it up. I should've posted a follow-up earlier.

akihikodaki:
@moritzheiber Thank you for your reply. I understood now.

PR #6645, opened 2018-03-05
Add bio fields

### LINKED ISSUE #121: No URL in profile
This means that I can't link out to my own site, and that I can't do [distributed verification](http://www.kevinmarks.com/distributed-verify.html) using Mastodon.

Bonus points for checking the rel='me' link back in Mastodon itself

### LINKED ISSUE #232: bio options
a toot-length bio section, the current length is Just Not Enough, 500 characters will always be sufficient. a readmore might be necessary for the column view, and that's fine.

on this topic, if you truly want to allow and support people, i would highly suggest a pronouns section, optional, of course, to ease that social layer of difficulty with accepting people's genders and pronouns. thank you for listening.

### LINKED ISSUE #3211: Add pronoun field alongside bio
It'd be really nice if we could have some sort of explicit pronoun field in profiles. It was suggested before alongside other bio feature requests but all the relevant issues appear to be closed. So, I'll open a dedicated issue for it.

What I was thinking:

* It would offer checkboxes for a few presets, plus a free text field limited to a small number of characters.
* The presets would likely be she/her/her, he/him/his, and they/them/their, and could be translated.
* It would be an optional field, and not checking anything is equivalent to "address by name only."
* Checking "other" would automatically count as checking "they" on languages that are not the user's language.
* Languages can interpret "they" as "address by name only" if they don't have a standard-accepted equivalent (for example, French).
* Languages can choose to interpret "he" and "she" as "they" if they don't have any gendered pronouns.
* If we choose to interpret not checking any pronouns as not showing anything, we should have a separate checkbox for "name only."

The language for this field would probably be something similar to "address by" rather than "pronouns."

If it's necessary to get users to buy into it, we could probably do what Facebook does and list it as "male/female/other" and then only show the checkboxes if someone selects other, picking he/she for male/female.

Basically what I had in mind was having checkboxes like:

* he
* she
* they
* other: __________

* * * *

Gargron:
A feature that allows adding structured data to profiles. Based on @marrus-sh's implementation (could not find a glitch-soc PR for it though), except the fields are not part of the text.

- Fix #3211
- Fix #232
- Fix #121

Screenshots:

image

image

image

___

- [x] Display on public page
- [x] Return from REST API
- [x] Edit
- [x] Federate

___

**How they are stored internally:** As a JSONB column on the accounts table, which contains an array of `{ "name": "", "value": "" }` objects. Why: Normally something like this would be implemented via a has_many association with separate database records for every field, but I do not want to add so many more eager loaded queries to everywhere, and the fields don't really need to be queried or updated standalone, so being nested in the account record seems okay.

**How they are federated:** Inside the actor's `attachment` tag they are represented as:

```json
{
  "type": "PropertyValue",
  "name": "",
  "value": ""
}
```

Where PropertyValue and value come from the schema.org context, while name comes from the vanilla ActivityStreams context. The value can contain HTML, while the name cannot.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_04_02_040909) do
+ActiveRecord::Schema.define(version: 2018_04_10_204633) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_stat_statements"
@@ -75,6 +75,7 @@ ActiveRecord::Schema.define(version: 2018_04_02_040909) do
     t.boolean "memorial", default: false, null: false
     t.bigint "moved_to_account_id"
     t.string "featured_collection_url"
+    t.jsonb "fields"
     t.index "(((setweight(to_tsvector('simple'::regconfig, (display_name)::text), 'A'::\"char\") || setweight(to_tsvector('simple'::regconfig, (username)::text), 'B'::\"char\")) || setweight(to_tsvector('simple'::regconfig, (COALESCE(domain, ''::character varying))::text), 'C'::\"char\")))", name: "search_index", using: :gin
     t.index "lower((username)::text), lower((domain)::text)", name: "index_accounts_on_username_and_domain_lower"
     t.index ["uri"], name: "index_accounts_on_uri"
--- a/spec/services/activitypub/process_account_service_spec.rb
+++ b/spec/services/activitypub/process_account_service_spec.rb
@@ -1,5 +1,26 @@
 require 'rails_helper'
 
 RSpec.describe ActivityPub::ProcessAccountService do
-  pending
+  subject { described_class.new }
+
+  context 'property values' do
+    let(:payload) do
+      {
+        id: 'https://foo',
+        type: 'Actor',
+        inbox: 'https://foo/inbox',
+        attachment: [
+          { type: 'PropertyValue', name: 'Pronouns', value: 'They/them' },
+          { type: 'PropertyValue', name: 'Occupation', value: 'Unit test' },
+        ],
+      }.with_indifferent_access
+    end
+
+    it 'parses out of attachment' do
+      account = subject.call('alice', 'example.com', payload)
+      expect(account.fields).to be_a Hash
+      expect(account.fields).to include('Pronouns' => 'They/them')
+      expect(account.fields).to include('Occupation' => 'Unit test')
+    end
+  end
 end
--- a/spec/services/activitypub/process_account_service_spec.rb
+++ b/spec/services/activitypub/process_account_service_spec.rb
@@ -18,9 +18,14 @@ RSpec.describe ActivityPub::ProcessAccountService do
 
     it 'parses out of attachment' do
       account = subject.call('alice', 'example.com', payload)
-      expect(account.fields).to be_a Hash
-      expect(account.fields).to include('Pronouns' => 'They/them')
-      expect(account.fields).to include('Occupation' => 'Unit test')
+      expect(account.fields).to be_a Array
+      expect(account.fields.size).to eq 2
+      expect(account.fields[0]).to be_a Account::Field
+      expect(account.fields[0].name).to eq 'Pronouns'
+      expect(account.fields[0].value).to eq 'They/them'
+      expect(account.fields[1]).to be_a Account::Field
+      expect(account.fields[1].name).to eq 'Occupation'
+      expect(account.fields[1].value).to eq 'Unit test'
     end
   end
 end
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -11,7 +11,9 @@ class Settings::ProfilesController < ApplicationController
   obfuscate_filename [:account, :avatar]
   obfuscate_filename [:account, :header]
 
-  def show; end
+  def show
+    @account.build_fields
+  end
 
   def update
     if UpdateAccountService.new.call(@account, account_params)
@@ -25,7 +27,7 @@ class Settings::ProfilesController < ApplicationController
   private
 
   def account_params
-    params.require(:account).permit(:display_name, :note, :avatar, :header, :locked)
+    params.require(:account).permit(:display_name, :note, :avatar, :header, :locked, fields_attributes: [:name, :value])
   end
 
   def set_account
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -63,4 +63,15 @@ module ApplicationHelper
   def opengraph(property, content)
     tag(:meta, content: content, property: property)
   end
+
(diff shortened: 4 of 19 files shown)


akihikodaki:
I am not for this change.

A writer of the bio fields may be excited and appreciate that it allows markup like that, but I do not think such a markup will actually benefit readers. A colon-separated text or even a normal statements would work well.

Gargron:
>A writer of the bio fields may be excited and appreciate that it allows markup like that, but

No, this change does not plan for special markup, rather, edit profile form should have +/- field inputs.

akihikodaki:
> No, this change does not plan for special markup, rather, edit profile form should have +/- field inputs.

I meant the table is the special markup.

Gargron:
@akihikodaki is normalizeAccount not run for the "me" account included with initialState in current master? i am trying to bring my "bio fields" PR up to date and running into this issue...

ThisIsMissEm:
I was thinking of doing a similar PR, but making the fields configurable by the admins (perhaps "default fields") and then also adding the ability for "locked" fields, which require an admin or certain application to edit.

Our use case is a default field that adds a "verified" status, and that being managed by an application, such that you can't manually add this field, and you could then know with certainty that we have actually verified the user.

This is important as we have users that may meet in person (potentially more likely than in other mastodon instances).

akihikodaki:
@Gargron 
> @akihikodaki is normalizeAccount not run for the "me" account included with initialState in current master? i am trying to bring my "bio fields" PR up to date and running into this issue...

I have just checked it and it looks fine for me. I added `console.log`:
```diff
diff --git a/app/javascript/mastodon/actions/importer/index.js b/app/javascript/mastodon/actions/importer/index.js
index 5b18cbc1d..907fe558d 100644
--- a/app/javascript/mastodon/actions/importer/index.js
+++ b/app/javascript/mastodon/actions/importer/index.js
@@ -45,6 +45,7 @@ export function importFetchedAccounts(accounts) {
   }
 
   accounts.forEach(processAccount);
+  console.log(normalAccounts);
   putAccounts(normalAccounts, !autoPlayGif);
 
   return importAccounts(normalAccounts);
```

And here is its output:
```JSON
[{"id":"1","username":"admin","acct":"admin","display_name":"ねこ","locked":false,"created_at":"2018-04-05T12:02:21.752Z","note":"<p></p>","url":"http://localhost:3000/@admin","avatar":"http://localhost:3000/avatars/original/missing.png","avatar_static":"http://localhost:3000/avatars/original/missing.png","header":"http://localhost:3000/headers/original/missing.png","header_static":"http://localhost:3000/headers/original/missing.png","followers_count":1,"following_count":0,"statuses_count":9,"display_name_html":"ねこ","note_emojified":"<p></p>"}]
```

Gargron:
@ThisIsMissEm I'm sorry but that won't work. A verified status like that should *not* federate, in my opinion, we should not teach users to trust anything like that that comes out of other servers. Besides, you don't even need something so complicated. Add a boolean column to the accounts table and put a checkmark on public profiles, somewhere where it can't be confused with any emoji.

ThisIsMissEm:
@Gargron oh, hm. I hadn't really thought about that as a valid option — seemed like it'd cause a compatibility issue and be a bit too much of a fork. I'm still getting bearings as to managing the forked codebase and making sure we can still update to upstream.

ThisIsMissEm:
@Gargron for that, is it possible to use react components on the `/admin` or `/settings` pages? I'd be happy to do the UI for you if that's why you're leaving it as an open todo.

spec/services/activitypub/process_account_service_spec.rb
```diff
@@ -1,5 +1,26 @@
...
+          { type: 'PropertyValue', name: 'Occupation', value: 'Unit test' },
+        ],
+      }.with_indifferent_access
+    end
+
+    it 'parses out of attachment' do
+      account = subject.call('alice', 'example.com', payload)
+      expect(account.fields).to be_a Hash
+      expect(account.fields).to include('name' => 'Pronouns', 'value' => 'They/them')
+      expect(account.fields).to include('name' => 'Occupation', 'value' => 'Unit test')
```
ThisIsMissEm: is it worth checking that `account.fields` doesn't have more than the payload declared?

Cassolotl:
These screenshots look awesome! :)

ykzts (approved):

kevinmarks:
Regarding verification, see the http://www.kevinmarks.com/distributed-verify.html approach. Add rel=me to these links, and check that they link back with a rel=me as well. Then you can tell that the same person can edit both pages.

kevinmarks:
Excellent, I'll give it a try when this gets deployed.

Wolf480pl:
Does it render to semantic HTML, eg. using a microformat like hCard ?

Wolf480pl:
>It's a th/td table.

I'd probably go for dt/dd

>Is there hCard markup for arbitrary information fields?

The whole idea of hCard is to tell the fields apart, eg. tag the email field as email, the nickname field as nickname, the birthday field as birthday, etc. Now that I think of it, it'd probably require some further UI to let the users optionally add hCard classes... sounds like a separate feature I guess?

kevinmarks:
Arbitrary properties are allowed in h-card, but that might not be ideal.
Worth putting u-url on url properties.

Defined properties are http://microformats.org/wiki/h-card#Properties

Gargron:
>The whole idea of hCard is to tell the fields apart, eg. tag the email field as email, the nickname field as nickname, the birthday field as birthday, etc

Well, I never said this was an hCard. The whole idea of *this* function is to let users define whatever they deem necessary on their profile. So while it could be something common like "homepage", it could also be "github" or "pronouns" or "avatar made by" or "favourite cake".

kevinmarks:
You could use the h-card ones as autocomplete (they're based on vcard,
which has had a lot of convergence research)

db/schema.rb
```diff
@@ -75,6 +75,7 @@
     t.boolean "memorial", default: false, null: false
     t.bigint "moved_to_account_id"
     t.string "featured_collection_url"
+    t.jsonb "fields"
```
saper: If `fields` is JSONB shouldn't be a Hash and not an Array? 

See report on https://discourse.joinmastodon.org/t/error-on-profile-edit-for-single-user-nomethoderror-undefined-method-for-hash/1563

nightpool:
@saper no, because profile fields are sorted and json objects are arbitrary
order

saper:
> @saper no, because profile fields are sorted and json objects are arbitrary order

Thanks. But it seems to me that JSONB storage will lose that ordering? Is this the correct serialization (or am I completely confusing something)?

nightpool:
@saper jsonb columns can be any json type—arrays, objects, or primitive values.

saper:
Ok, but when ActiveRecord fetches JSONB from PostgreSQL, what do we get?

nightpool:
@saper it will return an array for a json array, a hash for a json object, a corresponding rails value (string, number, boolean etc) for json primitive values.

PR #6654, opened 2018-03-06
Do not include build dependencies in final Docker image

akihikodaki:
This reduces the image size from 1.77GB to 1.56GB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,9 @@
-FROM ruby:2.5.0-alpine3.7
-
-LABEL maintainer="https://github.com/tootsuite/mastodon" \
-      description="Your self-hosted, globally interconnected microblogging community"
-
-ARG UID=991
-ARG GID=991
+FROM ruby:2.5.0-alpine3.7 as base
 
 ENV RAILS_SERVE_STATIC_FILES=true \
     RAILS_ENV=production NODE_ENV=production
 
-ARG LIBICONV_VERSION=1.15
-ARG LIBICONV_DOWNLOAD_SHA256=ccf536620a45458d26ba83887a983b96827001e92a13847b45e4925cc8913178
-
-EXPOSE 3000 4000
-
-WORKDIR /mastodon
-
 RUN apk -U upgrade \
- && apk add -t build-dependencies \
-    build-base \
-    icu-dev \
-    libidn-dev \
-    libressl \
-    libtool \
-    postgresql-dev \
-    protobuf-dev \
-    python \
  && apk add \
     ca-certificates \
     ffmpeg \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,6 +18,8 @@ EXPOSE 3000 4000
 
 WORKDIR /mastodon
 
+COPY Gemfile Gemfile.lock package.json yarn.lock .yarnclean /mastodon/
```
Gargron: This means you have to reinstall the entire OS inside the container every time any of the source files change.

Dockerfile
```diff
@@ -44,31 +46,27 @@ RUN apk -U upgrade \
...
-COPY Gemfile Gemfile.lock package.json yarn.lock .yarnclean /mastodon/
-
-RUN bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-include=/usr/local/include \
+ && bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-include=/usr/local/include \
  && bundle install -j$(getconf _NPROCESSORS_ONLN) --deployment --without test development \
- && yarn --pure-lockfile \
- && yarn cache clean
+ && /tmp/src/yarn/yarn-v$YARN_VERSION/bin/yarn --pure-lockfile \
+ && /tmp/src/yarn/yarn-v$YARN_VERSION/bin/yarn cache clean \
+ && apk del build-dependencies \
```
moritzheiber: This should probably include `--purge`?

moritzheiber (commented):
While I do like the overall intention, it's a conundrum. Currently the Docker images are used for both, independent source code compilation and deployment and "regular" deployments without touching the code inside ([the instructions are pretty explicit](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Docker-Guide.md#building-your-own-image)). Removing the build dependencies would inherently break said functionality and I would prefer a different solution (e.g. splitting up the images, separating concerns, build vs. deployment; using a base image with `libiconv` and `yarn` already installed).

akihikodaki:
9ce976a79fd9037a17adb42d79fed074f7366fb6 keeps cache functional (https://github.com/tootsuite/mastodon/pull/6654#pullrequestreview-101450377.)
It also fixes `assets:precompile (https://github.com/tootsuite/mastodon/pull/6654#pullrequestreview-101453557 I didn't know webpacker depends on yarn and it was unexpected that it does not work.)

moritzheiber:
@akihikodaki The problem persists. If I wanted to add additional gems (which, currently, depending on the environment we need to do for things like SAML or LDAP support) I couldn't do that with the final container image. Or am I reading the `Dockerfile` incorrectly?

akihikodaki:
> If I wanted to add additional gems (which, currently, depending on the environment we need to do for things like SAML or LDAP support) I couldn't do that with the final container image. Or am I reading the Dockerfile incorrectly?

That is, unfortunately, true even without this change. You must pass `PAM_ENABLED`, `LDAP_ENABLED`, `CAS_ENABLED`, and `SAML_ENABLED` *before* building the final image.

It is because RubyGems are installed in the container and this pull request has nothing to do about that.

akihikodaki:
09dd26be50d05ffc4635f1b3d2f3ebf4389d48f5: rebased to f5ee2d469bb2ff398571694a95a16c8e819153c9.

PR #6656, opened 2018-03-06
Do not use npm

akihikodaki:
Both of yarn and npm are used in Mastodon, but the combined usage requires
a redundant dependency and may lead to data inconsistency.

Considering that yarn has autoclean feature which npm does not have,
this change replaces all npm usage with yarn.

This change requires documentation update. Most notably, the following
command must be executed before assets precompilation if any system
dependency of node-sass has changed:

yarn install --force --pure-lockfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,6 @@ RUN apk -U upgrade \
     libidn \
     libpq \
     nodejs \
-    nodejs-npm \
     protobuf \
     su-exec \
     tini \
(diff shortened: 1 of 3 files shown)


Gargron (approved):

akihikodaki:
I found the `postinstall` script was a workaround for a bug already fixed:
https://github.com/yarnpkg/yarn/pull/3704

It has been a year since the fix has been applied and the problem should not occur if the system is properly updated. A documentation update I suggested earlier is not necessary.

PR #6666, opened 2018-03-07
Adjust privacy policy

Gargron:
- Describe in detail Mastodon's use of data
- Change data retention period of IP addresses from 5 years to 12 months

Sorry about localizations, but I think it would be bad to keep incorrect translations.

Also, I am not sure if:

- The old version must be kept somehow?
- This requires opt-in from everybody's users? (This is why "breaking" label)

Below is a render of the policy:

___

<blockquote>
<h2>Privacy Policy</h2>
<h3 id="collect">What information do we collect?</h3>

<ul>
<li><em>Basic account information</em>: If you register on this server, you may be asked to enter a username, an e-mail address and a password. You may also enter additional profile information such as a display name and biography, and upload a profile picture and header image. The username, display name, biography, profile picture and header image are always listed publicly.</li>
<li><em>Posts, following and other public information</em>: The list of people you follow is listed publicly, the same is true for your followers. When you submit a message, the date and time is stored as well as the application you submitted the message from. Messages may contain media attachments, such as pictures and videos. Public and unlisted posts are available publicly. When you feature a post on your profile, that is also publicly available information. Your posts are delivered to your followers, in some cases it means they are delivered to different servers and copies are stored there. When you delete posts, this is likewise delivered to your followers. The action of reblogging or favouriting another post is always public.</li>
<li><em>Direct and followers-only posts</em>: All posts are stored and processed on the server. Followers-only posts are delivered to your followers and users who are mentioned in them, and direct posts are delivered only to users mentioned in them. In some cases it means they are delivered to different servers and copies are stored there. We make a good faith effort to limit the access to those posts only to authorized persons, but other servers may fail to do so. Therefore it's important to review servers your followers belong to. You may toggle an option to approve and reject new followers manually in the settings. <em>Please keep in mind that the operators of the server and any receiving server may view such messages</em>, and that recipients may screenshot, copy or otherwise re-share them. <em>Do not share any dangerous information over Mastodon.</em></li>
<li><em>IPs and other metadata</em>: When you log in, we record the IP address you log in from, as well as the name of your browser application. All the logged in sessions are available for your review and revocation in the settings. The latest IP address used is stored for up to 12 months. We also may retain server logs which include the IP address of every request to our server.</li>
</ul>

<hr class="spacer" />

<h3 id="use">What do we use your information for?</h3>

<p>Any of the information we collect from you may be used in the following ways:</p>

<ul>
<li>To provide the core functionality of Mastodon. You can only interact with other people's content and post your own content when you are logged in. For example, you may follow other people to view their combined posts in your own personalized home timeline.</li>
<li>The email address you provide may be used to send you information, notifications about other people interacting with your content or sending you messages, and to respond to inquiries, and/or other requests or questions.</li>
</ul>

<hr class="spacer" />

<h3 id="protect">How do we protect your information?</h3>

<p>We implement a variety of security measures to maintain the safety of your personal information when you enter, submit, or access your personal information. Among other things, your browser session, as well as the traffic between your applications and the API, are secured with SSL, and your password is hashed using a strong one-way algorithm. You may enable two-factor authentication to further secure access to your account.</p>

<hr class="spacer" />

<h3 id="data-retention">What is our data retention policy?</h3>

<p>We will make a good faith effort to:</p>

<ul>
<li>Retain server logs containing the IP address of all requests to this server, in so far as such logs are kept, no more than 90 days.</li>
<li>Retain the IP addresses associated with registered users no more than 12 months.</li>
</ul>

<p>You can request and download an archive of your content, including your posts, media attachments, profile picture, and header image.</p>

<p>You may irreversibly delete your account at any time.</p>

<hr class="spacer"/>

<h3 id="cookies">Do we use cookies?</h3>

<p>Yes. Cookies are small files that a site or its service provider transfers to your computer's hard drive through your Web browser (if you allow). These cookies enable the site to recognize your browser and, if you have a registered account, associate it with your registered account.</p>

<p>We use cookies to understand and save your preferences for future visits.</p>

<hr class="spacer" />

<h3 id="disclose">Do we disclose any information to outside parties?</h3>

<p>We do not sell, trade, or otherwise transfer to outside parties your personally identifiable information. This does not include trusted third parties who assist us in operating our site, conducting our business, or servicing you, so long as those parties agree to keep this information confidential. We may also release your information when we believe release is appropriate to comply with the law, enforce our site policies, or protect ours or others rights, property, or safety.</p>

<p>Your public content may be downloaded by other servers in the network. Your public and followers-only posts are delivered to the servers where your followers reside, and direct messages are delivered to the servers of the recipients, in so far as those followers or recipients reside on a different server than this.</p>

<p>When you authorize an application to use your account, depending on the scope of permissions you approve, it may access your public profile information, your following list, your followers, your lists, all your posts, and your favourites. Applications can never access your e-mail address or password.</p>

<hr class="spacer" />

<h3 id="coppa">Children's Online Privacy Protection Act Compliance</h3>

<p>Our site, products and services are all directed to people who are at least 13 years old. If this server is in the USA, and you are under the age of 13, per the requirements of COPPA (<a href="https://en.wikipedia.org/wiki/Children%27s_Online_Privacy_Protection_Act">Children's Online Privacy Protection Act</a>) do not use this site.</p>

<hr class="spacer" />

<h3 id="changes">Changes to our Privacy Policy</h3>

<p>If we decide to change our privacy policy, we will post those changes on this page.</p>

<p>This document is CC-BY-SA. It was last updated March 7, 2018.</p>

<p>Originally adapted from the <a href="https://github.com/discourse/discourse">Discourse privacy policy</a>.</p>
</blockquote>


m4sk1n:
I guess that depends of the country, so what about for example making old instances not work after upgrade if owners won’t add one line line to config file, to state that they know about the change and they will do the things required by their local law?

config/locales/en.yml
```diff
@@ -657,70 +657,77 @@ en:
...
+        <li><em>Direct and followers-only posts</em>: All posts are stored and processed on the server. Followers-only posts are delivered to your followers and users who are mentioned in them, and direct posts are delivered only to users mentioned in them. In some cases it means they are delivered to different servers and copies are stored there. We make a good faith effort to limit the access to those posts only to authorized persons, but other servers may fail to do so. Therefore it's important to review servers your followers belong to. You may toggle an option to approve and reject new followers manually in the settings. Please keep in mind that the operators of the server and any receiving server may view such messages, and that recipients may screenshot, copy or otherwise re-share them.</li>
```
nightpool: "to review servers your followers belong to" is slightly odd wording, i'd say "to review **the** servers your followers belong to".

config/locales/en.yml
```diff
@@ -657,70 +657,77 @@ en:
...
+      <ul>
+        <li><em>Basic account information</em>: If you register on this server, you may be asked to enter a username, an e-mail address and a password. You may also enter additional profile information such as a display name and biography, and upload a profile picture and header image. The username, display name, biography, profile picture and header image are always listed publicly.</li>
+        <li><em>Posts, following and other public information</em>: The list of people you follow is listed publicly, the same is true for your followers. When you submit a message, the date and time is stored as well as the application you submitted the message from. Messages may contain media attachments, such as pictures and videos. Public and unlisted posts are available publicly. When you feature a post on your profile, that is also publicly available information. Your posts are delivered to your followers, in some cases it means they are delivered to different servers and copies are stored there. When you delete posts, this is likewise delivered to your followers. The action of reblogging or favouriting another post is always public.</li>
```
nightpool: The fact that we list here explicitly all the different types of data you can create makes me a little scared that we'll have to update the privacy policy every time we add a new feature, which would be bad. For example, where do reports fall under here? What if we made lists public? You might want to add a more general category, or something like "information we collect from you includes but is not limited to"

While the specificity is good and useful, because you removed the very general `We collect information from you when you register on our site and gather data when you participate in the forum by reading, writing, and evaluating the content shared here` it's possible someone could read this and conclude that recording who made a report is against our privacy policy, or that displaying a list will be, which would be bad.

config/locales/en.yml
```diff
@@ -657,70 +657,77 @@ en:
...
       <h3 id="use">What do we use your information for?</h3>
 
       <p>Any of the information we collect from you may be used in one of the following ways:</p>
 
       <ul>
-        <li>To personalize your experience &mdash; your information helps us to better respond to your individual needs.</li>
-        <li>To improve our site &mdash; we continually strive to improve our site offerings based on the information and feedback we receive from you.</li>
-        <li>To improve customer service &mdash; your information helps us to more effectively respond to your customer service requests and support needs.</li>
-        <li>To send periodic emails &mdash; The email address you provide may be used to send you information, notifications that you request about changes to topics or in response to your user name, respond to inquiries, and/or other requests or questions.</li>
+        <li><em>To personalize your experience</em> &mdash; you can only interact with other people's content and post your own content when you are logged in. Furthermore you may follow other people to view their combined posts in your own personalized home timeline.</li>
```
nightpool: "— For example, you can...."

config/locales/en.yml
nightpool: you might want to make sure to say that "authorized persons" here includes the moderator of the server.

config/locales/en.yml
```diff
@@ -657,70 +657,77 @@ en:
...
+      <p>We do not sell, trade, or otherwise transfer to outside parties your personally identifiable information. This does not include trusted third parties who assist us in operating our site, conducting our business, or servicing you, so long as those parties agree to keep this information confidential. We may also release your information when we believe release is appropriate to comply with the law, enforce our site policies, or protect ours or others rights, property, or safety.</p>
 
-      <h3 id="third-party">Third party links</h3>
+      <p>When you authorize an application to use your account, depending on the scope of permissions you approve, it may access your public profile information, your following list, your followers, your lists, all your posts, and your favourites. Applications can never access your e-mail address or password.</p>
```
nightpool: again, same issue here with ", including but not limited to"

nightpool (changes requested):

nightpool:
(IANAL, this is not legal advice, etc)

jeroenpraat:
I think it's still US centric, like the mention of COPPA. Also it's important to separate the retention of IP adresses in the web server logs (depends on the server config) and in the Mastodon database (done by the Mastodon software).

nightpool:
@jeroenpraat we're adapting a privacy policy (from Discourse) for US sites. If other jurisdictions have other requirements, we're not really qualified to asses or handle that, unfortunately.

jeroenpraat:
@nightpool If it's for US sites only, I think it's better to ask during setup if the instance is hosted (and such) in the US. Most instances are **not** hosted in the US, so a default like this makes no sense IMO.

akihikodaki:
> If it's for US sites only, I think it's better to ask during setup if the instance is hosted (and such) in the US. Most instances are not hosted in the US, so a default like this makes no sense IMO.

That is true, but nobody knows what should be the default, which is unfortunate, I guess.

config/locales/en.yml
duggulous: I think the idea that a privacy policy is something you set and forget is antiquated and problematic.  As convenient as it might be in the short term to just not bother with it, we developers have to come to grips with the fact that the privacy policy is something that grows with the site, just like test suites and documentation. Point of fact, the privacy policy IS documentation, and it needs to remain both up to date, and specific.  Developers _should_ make a habit of updating the privacy policy whenever the data collected changes.  (ideally, this could be at least partially automated)

"Includes but is not limited to" effectively expands the bounds on what could be collected to infinity, which kind of defeats the purpose of this document.

config/locales/en.yml
duggulous: "In some cases it means they are delivered to different servers ..."
Is it uncertain in which cases posts get delivered to other servers?
My understanding may be incorrect but I what I thought was:
 if those users' accounts are hosted on other Mastodon instances, then the posts they are eligible to read will be copied to the server for their instance, and will be subject to the privacy policy of that instance, which may be more permissive than this one.

config/locales/en.yml
```diff
@@ -658,69 +658,81 @@ en:
...
 
-      <p>We do not sell, trade, or otherwise transfer to outside parties your personally identifiable information. This does not include trusted third parties who assist us in operating our site, conducting our business, or servicing you, so long as those parties agree to keep this information confidential. We may also release your information when we believe release is appropriate to comply with the law, enforce our site policies, or protect ours or others rights, property, or safety. However, non-personally identifiable visitor information may be provided to other parties for marketing, advertising, or other uses.</p>
+      <p>We do not sell, trade, or otherwise transfer to outside parties your personally identifiable information. This does not include trusted third parties who assist us in operating our site, conducting our business, or servicing you, so long as those parties agree to keep this information confidential. We may also release your information when we believe release is appropriate to comply with the law, enforce our site policies, or protect ours or others rights, property, or safety.</p>
```
duggulous: Any trusted third parties should be explicitly listed, so that users can seek the corresponding privacy policies and understand who is accessing their data.

duggulous: "agree to keep the information confidential" is not a particularly strong protection, as there are many ways to misuse data that don't involve sharing it with others.   
Ideally, no no information would be transferred to third parties unless they were also:
bound to not use the information for any purposes outside of assisting us in operating the site, conducting our business and providing the service.  
Also, these guarantees should apply not just to PII, but to any data the user has not made public.

Gargron: We need to distribute such a privacy policy that would be valid for all installations. So it can't get too specific. I doubt that admins will go hunting down the, say, DigitalOcean privacy policy, or Mailgun privacy policy (please mind that admins *can* completely rewrite the privacy policy themselves from the admin interface, if they decide to include that information)

You also have to see this in perspective to the prior privacy policy. Is it overall an improvement, or is it not? I can't fix everything at once.

config/locales/en.yml
Gargron: I am inclined to agree that being explicit, and updating when we need to collect more, is what makes a privacy policy useful and trustworthy...

config/locales/en.yml
maskedmakrel: Without an enforcement provision, the policy is fairly meaningless.  You make it clear that the policy is merely a "good faith" effort in which the weakest link is the default for privacy.  What is that actual mechanism for safeguarding privacy from abuse?  I'll be honest and suggest that as a new user, I took a dim view of the privacy options offered and assume they are no better than any other service but that there was less crass commercialization of my own posts.  It's hard to imagine that even the metadata provided is necessary for the improvement of the user experience as users tend to provide concrete feedback on services and will let you know what they want changed.  As for "private communications", the service appears to provide no real protection at all.  Unless users were to communicate with some kind of trusted encryption, it appears that private communications are somewhat open.  The ACLU has a 45 page report on "Informational Privacy in the Digital Age" which might provide you with some background as to the complexity of the issues at hand- but no one read reports.

Cassolotl:
It says "we, us, our" in all the headings, except this one:

> What is your data retention policy?

It should be changed to "What is our data retention policy?"

config/locales/en.yml
```diff
@@ -669,70 +669,78 @@ en:
...
-        <li>To improve our site &mdash; we continually strive to improve our site offerings based on the information and feedback we receive from you.</li>
-        <li>To improve customer service &mdash; your information helps us to more effectively respond to your customer service requests and support needs.</li>
-        <li>To send periodic emails &mdash; The email address you provide may be used to send you information, notifications that you request about changes to topics or in response to your user name, respond to inquiries, and/or other requests or questions.</li>
+        <li>To provide the core functionality of Mastodon. You can only interact with other people's content and post your own content when you are logged in. For example, you may follow other people to view their combined posts in your own personalized home timeline.</li>
+        <li>The email address you provide may be used to send you information, notifications about other people interacting with your content or sending you messages, and to respond to inquiries, and/or other requests or questions.</li>
```
nightpool: I think this needs to have a list bullet about site moderation, or using your information (such as IP addresses) to determine a code of conduct violations or attempts to evade a moderation action.

nightpool (changes requested):

nightpool:
Other then that one bullet point, LGTM

Cassolotl:
>Please keep in mind that the operators of the server and any receiving server may view such messages, and that recipients may screenshot, copy or otherwise re-share them. Do not share any dangerous information over Mastodon.

I feel like "operator" is not very clear to the average user, who is used to terms like "admins" and "mods"! So maybe it could say "the admins and moderators of the server and any receiving server may view such messages"?

Gargron:
I disagree, I think "operator" is the most appropriate term, i.e. someone involved in operating the service: mod, admin, sysadmin, database administrator

nightpool:
Yes, there are people responsible for and that may have access to your database that are not moderators or administrators—for example, virtual hosting providers.

nightpool (approved):

PR #6669, opened 2018-03-07
Fix SAML name attribute statements

### LINKED ISSUE #6535: SAML attribute "full_name" should be used or become "first_name" & "last_name"
The SAML patch provides support for a "full_name" auth attribute by setting `SAML_ATTRIBUTES_STATEMENTS_FULL_NAME`.

However, this attribute is never used. `models/concerns/omniauthable.rb` sets `display_name` based on joining a `first_name` and `last_name` if provided.

The result of this is the display name is not able to be automatically set for users registering with SAML.

cc @moritzheiber

ineffyble:
Fixes #6535. Add additional first_name and last_name SAML attribute statement options, and modify Omniauthable concern to use full_name or first_name + last_name if not available.


ineffyble:
@gloaec Does this look right to you?

.env.production.sample
```diff
@@ -204,7 +204,9 @@ STREAMING_CLUSTER_NUM=1
 # SAML_SECURITY_ASSUME_EMAIL_IS_VERIFIED=true
 # SAML_ATTRIBUTES_STATEMENTS_UID="urn:oid:0.9.2342.19200300.100.1.1"
 # SAML_ATTRIBUTES_STATEMENTS_EMAIL="urn:oid:1.3.6.1.4.1.5923.1.1.1.6"
-# SAML_ATTRIBUTES_STATEMENTS_FULL_NAME="urn:oid:2.5.4.42"
+# SAML_ATTRIBUTES_STATEMENTS_FULL_NAME="urn:oid:2.16.840.1.113730.3.1.241"
```
Gargron: That looks like it could be a number sequence from LOST

Gargron (approved):

app/models/concerns/omniauthable.rb
```diff
@@ -58,13 +58,14 @@ def user_params_from_auth(auth)
       email_is_verified = auth.info.verified || auth.info.verified_email || assume_verified
       email             = auth.info.verified_email || auth.info.email
       email             = email_is_verified && !User.exists?(email: auth.info.email) && email
+      display_name      = auth.info.full_name || [auth.info.first_name, auth.info.last_name].join(' ')
```
gloaec: @ineffyble Thanks for the work ! Just one comment : If the full_name is not provided by SAML, the user willl be registered will fullname " " (space char), since `[nil, nil].join(' ') == ' '`. I fixed my previous suggestion by adding `strip` there https://github.com/tootsuite/mastodon/issues/6535#issuecomment-367932856

@Gargron Dont play lottery with those numbers ;)

Gargron: Do `.compact.join` then

PR #6672, opened 2018-03-07
Remove text requirement when media attached from statuses

Gargron:
- Do not auto-insert media URLs into status text in web UI
- Text is not required when media is attached to status
- If content warning is submitted with media and no text, content warning becomes text
- Do not turn empty text into `<p></p>` in formatter

### DIFF
--- a/app/javascript/mastodon/reducers/compose.js
+++ b/app/javascript/mastodon/reducers/compose.js
@@ -90,7 +90,6 @@ function appendMedia(state, media) {
     map.update('media_attachments', list => list.push(media));
     map.set('is_uploading', false);
     map.set('resetFileKey', Math.floor((Math.random() * 0x10000)));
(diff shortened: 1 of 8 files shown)


app/javascript/mastodon/reducers/compose.js
```diff
@@ -106,7 +105,6 @@ function removeMedia(state, mediaId) {
 
   return state.withMutations(map => {
     map.update('media_attachments', list => list.filterNot(item => item.get('id') === mediaId));
-    map.update('text', text => text.replace(media.get('text_url'), '').trim());
```
ykzts: https://github.com/tootsuite/mastodon/pull/6672/files#diff-3fb459a28b27d6839fb62d18bceecf8aL104 This line is unnecessary.

ykzts (approved):

PR #6684, opened 2018-03-07
fix validation error (media only status)

MitarashiDango:
Fixed an issue that caused a validation error.  (fix #6672)

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -53,7 +53,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
       visibility: visibility_from_audience,
       thread: replied_to_status,
       conversation: conversation_from_uri(@object['conversation']),
-      media_attachments: process_attachments,
+      media_attachments: process_attachments.take(4),
     }
   end
 
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -21,9 +21,11 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
 
   def process_status
     media_attachments = process_attachments
+    create_status_params = status_params
+    create_status_params['media_attachments'] = media_attachments
 
     ApplicationRecord.transaction do
-      @status = Status.create!(status_params)
+      @status = Status.create!(create_status_params)
 
       process_tags(@status)
       attach_media(@status, media_attachments)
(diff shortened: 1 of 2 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -21,9 +21,11 @@ def perform
 
   def process_status
     media_attachments = process_attachments
+    create_status_params = status_params
+    create_status_params['media_attachments'] = media_attachments
 
     ApplicationRecord.transaction do
-      @status = Status.create!(status_params)
+      @status = Status.create!(create_status_params)
```
ykzts:
```ruby
@status = Status.new(status_param)

process_tags(@status)
attach_media(@status, media_attachments)

@status.save!
```

🤔

cc @abcang

akihikodaki: That does not work, I think, because `attach_media` refers to the ID of the status.

ykzts: hm

Gargron: They can be reordered.. You can create media unattached first, and then pass to Status.create like in PostStatusService...

By the way, sorry I missed this..

abcang: I think that it is good to give media_attachments as a parameter of Status. And `attach_media` will not be used.

abcang:
Probably, `OStatus::Activity::Creation` needs a similar fix.

akihikodaki (changes requested):
Nice bug catch. Please see my review comment. Also, as @abcang said, don't forget to remove `attach_media`.

MitarashiDango:
Addressed

akihikodaki (approved):
It looks good to me.

app/lib/activitypub/activity/create.rb
```diff
@@ -54,6 +51,7 @@ def status_params
       visibility: visibility_from_audience,
       thread: replied_to_status,
       conversation: conversation_from_uri(@object['conversation']),
+      media_attachments: process_attachments,
```
abcang: In this case, it is not good to store the media in the transaction.

abcang (changes requested):

app/lib/activitypub/activity/create.rb
MitarashiDango: Save the return value of `process_status_params` to the `status_params` variable beforehand and pass it as an argument of `Status.create`?

abcang: It looks good!

abcang:
It may be better to keep the function of limiting the number of media.

MitarashiDango:
Addressed
- Fixed an issue where media was included in transaction.
- Restored the limit on the number of attached media.

abcang (approved):

Gargron (approved):

PR #6685, opened 2018-03-07
docs: Add AUTHORS file

espenronnevik:
GNU coding standards suggest that source code should include a file listing the software authors. Depending only on the git log to provide this information is not a good solution, nor does the license require it. 

The following commit adds a file named AUTHORS which includes a short statement listing Mastodon's GitHub URL and every author that has created a commit up to when the commit was created. The information is pulled from the git log with name and email, sorted in order of appearance.

I made the decision to include emails since a few commits do not include real names but handles. Although I have tried to avoid duplicates, there are probably a few where name or email have changed.

Review the list at your discretion.


Gargron:
Some of the contributors told me that they don't want to be found via their real name / e-mail. While it's simple to clone the repository and view these details, so they're not secret, Google doesn't clone repositories, so there's still some value in not making such an obvious document. I suggest using GitHub usernames instead of raw name / e-mail.

espenronnevik:
Fair comments. There is no true and correct way to make this file. It should, of course, be stressed that being listed as an author should be voluntary, nor that every contributor has to be listed. It is fairly common to not list people who perform a simple typo correction for instance. I leave the actual contents of the file up to you as the project owner.

I'll work on pulling the necessary data out of GitHub to address your suggestion, as I can see no way to map the git log data to GitHub. But if we are going that route, would changing the filename to AUTHORS.md (to match the rest of the textfiles) and then using usernames with a link to GitHub profile page be a good idea?

That would make the list entries look like this:
[Gargron](https://github.com/Gargron)

akihikodaki:
> Depending only on the git log to provide this information is not a good solution,

Why do you think it is not good?

> nor does the license require it.

It is not required to preserve `AUTHORS` either as far as I know. AGPL-3.0 just requires to preserve the copyright notices. So, at least in some countries, it may be required to preserve one if you wrote:

> Copyright (C) 2017-2018 Mastodon contributors, listed in AUTHORS file along with this program

…or sort of. Indeed, "see AUTHORS" notice is very common in copyright notices.

And here, a question arises: isn't stating "see Git history for the copyright holders" enough? Honestly, I have seen few times such a notice, and it may require a further investigation if it is valid.

nightpool:
I don't think there's any legal reason we need to add the Authors file, and because of that we might want a disclaimer that it may be out of date and is provided for informational purposes only (since I don't see us updating it any *more* often then once per release) but other then that I'm happy to see this get merged!

nightpool (approved):

jeroenpraat:
@espenronnevik Is the order related to the number of commits or the total size of the commits? If it's possible I prefer the latter. I also suggest to look into other repositories of tootsuite, especially documentation, or did you already did that?

espenronnevik:
@akihikodaki 
> Why do you think it is not good?

My initial thoughts were that it is not good because it is perfectly legal to use Mastodon without ever touching git, at least as far as I know. Someone could check out the source code from git, make a zip or tarball and distribute it without the git history. At least a clear list displaying that the software has many authors and here they are or at least a method to reach them would be good.

> AGPL-3.0 just requires to preserve the copyright notices

You bring up a very good point in that the AUTHORS file is usually tied to the copyright notice. It's easier to refer to a single file than to, say, 200 separate copyright notices. That said though, where **is** Mastadons copyright notice? Admittedly I have just been idly searching through the files, but the only reference to any copyright notice is in lib/mastodon/migration_helpers.rb which says "Copyright (c) 2011-2017 GitLab B.V." Surely there must be something I am missing here?

I am no expert on GPL or AGPL Licenses, but as far as I can understand them every file should have a notice stating the copyright.

The AGPL License says:

> each file should have at least the "copyright" line and a pointer to where the full notice is found.

Your example uses the Apache License, which as far as I know is a more permissive license than GPL. Again, I don't really know enough about them to be able to make a comparison.

@nightpool 
Updating it once per release should be more than enough. It could, of course, include the boilerplate "and others" at the end, just to cover all the bases.

espenronnevik:
@jeroenpraat The order seems to be according to the number of commits. This seemed to be the only sorting method I could get out of the GitHub API, at least easily.

jeroenpraat:
@espenronnevik OK, no problem. And the other repositories in the project?

espenronnevik:
@jeroenpraat Oh, I'm sorry. I'm not sure if I understood what you wanted me to look into?

jeroenpraat:
@espenronnevik Did you got the names and number of commits only from the tootsuite/mastodon repository, or also repositories listed under [tootsuite](https://github.com/tootsuite/), especially [tootsuite/documentation](https://github.com/tootsuite/documentation)?  

I think it would be a good idea to combine those.

Gargron:
@jeroenpraat Disagree, the repositories are different and may have different licenses etc.

jeroenpraat:
OK, well than it would be nice that at least the docu repo will also have an author file.

akihikodaki:
@espenronnevik

> My initial thoughts were that it is not good because it is perfectly legal to use Mastodon without ever touching git, at least as far as I know.

> You bring up a very good point in that the AUTHORS file is usually tied to the copyright notice. It's easier to refer to a single file than to, say, 200 separate copyright notices. That said though, where is Mastodons copyright notice?

So adding a copyright notice and referring to the author list is necessary, and AUTHORS file itself has nothing to do with it. If you are motivated by the legal reason, you need a different change.

> Someone could check out the source code from git, make a zip or tarball and distribute it without the git history.

By the way, I found that is what GitHub does when requesting a zip or tarball. AUTHORS file would prevent from removing the attributions with the feature by mistake.

> Your example uses the Apache License, which as far as I know is a more permissive license than GPL. Again, I don't really know enough about them to be able to make a comparison.

The permissive nature of the license is irrelevant from the topic. However, I agree that we need to be careful to compare such instances because both of the Apache license and AGPL are complicated.

akihikodaki (approved):
I approve this change, considering it is helpful to prevent from removing the author list when requesting an archive from GitHub.

I have a few comments about the implementation:
* What is the ordering of the list? I guess it depends on the numbers of their commits, but since the values vary by time, I suggest to use another ordering (alphabetical order or the first contribution date.) I do not want to make my name persistent near the top of the list even after I stopped to contribute to this project.
* `CONTRIBUTING.md` should be updated to tell new contributors to update `AUTHORS.md`.
* I prefer to include my name and email addresses as well. (This is just my personal preference and I am not suggesting to do so for all contributors.)

Gargron:
Could you document how this list was generated?

Sylvhem:
> That said though, where is Mastodons copyright notice? Admittedly I have just been idly searching through the files, but the only reference to any copyright notice is in lib/mastodon/migration_helpers.rb which says "Copyright (c) 2011-2017 GitLab B.V." Surely there must be something I am missing here?

@espenronnevik There is currently no copyright notices in Mastodon. #2938 is a pending pull request submitted last year that added both an AUTHOR file and copyright notices. Personally, I think it's a really good idea to introduce both of those things.

Like @nightpool pointed out, we may want to add something like "This document is provided for informational purposes only. Since it is only updated once per release, the version you are looking at may be currently out of date. To see the full list of contributors, consider looking at the [Git history](https://github.com/tootsuite/mastodon/graphs/contributors) instead." at the end of the file.

Also, this pull request should probably tagged with `legal`.

espenronnevik:
@Gargron The file was generated by making GET calls to https://api.github.com/repos/tootsuite/mastodon/contributors, which gives every contributor by order of commits, merging all 15 pages and then creating the list by pulling the "login" and "html_url" attribute from each element. I admit that the ordering leaves something to be desired, I would personally have preferred the chronological order of the original file, but the GitHub API does not sort by this order and I don't know of an easy way to match git log data (name/email) with GitHub users.

@Sylvhem I agree on the copyright notice. I was under the impression that the license required it to have any meaning at all, preferably on all files but one at the minimum. But that is another pull request and discussion I guess.

Issue #6605, opened 2018-03-03
Unable to build v2.3.0rc1 with docker


Moskito99:
I tried to build version 2.3.0rc1 with docker-engine 17.05.0-ce
I get the following error:
```
Step 16/19 : COPY --chown=mastodon:mastodon . /mastodon
ERROR: Service 'web' failed to build: Unknown flag: chown
```

After digging around, it seems the error has something to do with the docker-engine.

chown is not available in the docker-engine 17.05.0-ce, it is in 17.09 and up....
```
marcus@zeus:~/u4u.org/mastodon$ docker version
Client:
 Version:      17.05.0-ce
 API version:  1.29
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:54 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.05.0-ce
 API version:  1.29 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:54 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

This is the actual packet for Ubuntu 16.04 LTS
```
marcus@zeus:~/u4u.org/mastodon$ apt-cache policy docker-engine
docker-engine:
  Installiert:           17.05.0~ce-0~ubuntu-xenial
  Installationskandidat: 17.05.0~ce-0~ubuntu-xenial
  Versionstabelle:
 *** 17.05.0~ce-0~ubuntu-xenial 500
        500 https://apt.dockerproject.org/repo ubuntu-xenial/main amd64 Packages
        100 /var/lib/dpkg/status
```

Is it possible to update to 2.3.0rc1 without the chown command?

Regards,
Marcus

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

redtachyons:
You may have to upgrade docker version, chown flag is relatively new

Moskito99:
Hi,
yes I know!
But this is not how it should work.
Sorry, I chose a stable LTS distro to prevent problems with other apps /docker 
container, whatever...

Maybe the change of the dockerfile is a better solution.
https://stackoverflow.com/questions/44766665/how-do-i-docker-copy-as-non-root

COPY . /mastodon

And after this you can do the following:

[...]

PR #6698, opened 2018-03-08
Revert "Separate chown command in Dockerfile. Use tootsuite/mastodon …

akihikodaki:
…image (#6662)"

This reverts commit d7573fe584ea622e22f5a68f2bf120b7682c49f7.

The problem the commit addressed is occurred on an out-dated Docker configuration.

The report of the problem was made in #6605. The reporter has shown his Docker installation:
```
marcus@zeus:~/u4u.org/mastodon$ apt-cache policy docker-engine
docker-engine:
  Installiert:           17.05.0~ce-0~ubuntu-xenial
  Installationskandidat: 17.05.0~ce-0~ubuntu-xenial
  Versionstabelle:
 *** 17.05.0~ce-0~ubuntu-xenial 500
        500 https://apt.dockerproject.org/repo ubuntu-xenial/main amd64 Packages
        100 /var/lib/dpkg/status
``` 

The used edition is the Community Edition. The edition has two release channels, stable and edge. The latest stable version is 17.12.1, and 17.05.0- is an out date Community Edition version. Moreover, as suggested in https://github.com/tootsuite/mastodon/issues/6605#issuecomment-370557897, the repository configuration is also old.

Updating the repository configuration and then Docker will solve the problem. I have already confirmed the up-to-date configuration works. (See https://github.com/tootsuite/mastodon/issues/6495#issuecomment-370328359. Ubuntu 16.04, which the reporter of #6605 also used.)

We also do not have to worry about the version provided for Ubuntu 16.04 by its distributor, Canonical. The version is too old even to use Mastodon's compose file written in version 3 format.

Considering all the above, there is no need to support the old version. An administrator using an old version should be advised to update Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ruby:2.5.0-alpine3.7
 
 LABEL maintainer="https://github.com/tootsuite/mastodon" \
-      description="Your self-hosted, globally interconnected microblogging community"
+      description="A GNU Social-compatible microblogging server"
 
 ARG UID=991
 ARG GID=991
@@ -73,9 +73,7 @@ RUN addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodo
  && mkdir -p /mastodon/public/system /mastodon/public/assets /mastodon/public/packs \
  && chown -R mastodon:mastodon /mastodon/public
 
-COPY . /mastodon
-
-RUN chown -R mastodon:mastodon /mastodon
+COPY --chown=mastodon:mastodon . /mastodon
 
 VOLUME /mastodon/public/system /mastodon/public/assets /mastodon/public/packs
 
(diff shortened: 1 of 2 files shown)


nightpool:
There's no reasons to break people's installations needlessly, or require an upgrade when the separate RUN command is, for all intents and purposes, equivalent to the copy --chown instruction.

Also, your PR seems to revert some unrelated changes:
image

nightpool:
Furthermore, the version mentioned is what's used for building on the Docker Hub. If Docker's own infrastructure doesn't even support the command, how can we require others to use it in good faith?

akihikodaki:
> Also, your PR seems to revert some unrelated changes:

It was included in the commit.

> Furthermore, the version mentioned is what's used for building on the Docker Hub.

It seems to use Docker Enterprise Edition, and its latest version is still 17.06. I agree we should wait for a new Enterprise Edition release.
https://docs.docker.com/enterprise/

PR #6699, opened 2018-03-08
Simplify the test script

ghost:


Gargron:
>6699

Nice!

But I don't mind seeing at a glance which commands are going to be run instead of a wildcard...

ykzts:
Unfortunately if you use wildcards, the order of run will be changed. You should run Jest after lint.

Gargron:
Alright, I'm going to decline this PR, sorry.

imncls:
no worries~!

PR #6707, opened 2018-03-09
Change avatar size on form hints

ykzts:
Follow up #6651.


Gargron:
Should maybe wait until Weblate is merged, because Weblate will need resetting after this gets merged...

PR #6718, opened 2018-03-09
api: show if elasticsearch is enabled or not when requesting instance properties

kaniini:
This allows remote instances to know if a Mastodon instance has elasticsearch enabled or not, which can be useful for not federating with those instances.

### DIFF
--- a/app/serializers/rest/instance_serializer.rb
+++ b/app/serializers/rest/instance_serializer.rb
@@ -5,7 +5,7 @@ class REST::InstanceSerializer < ActiveModel::Serializer
 
   attributes :uri, :title, :description, :email,
              :version, :urls, :stats, :thumbnail,
-             :languages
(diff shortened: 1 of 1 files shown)


kaniini:
why did you close this?

Gargron:
This PR is not submitted in good faith. If your threat model is evil sysadmins, then it's not only trivial to add ElasticSearch (original commit took 78 lines of code, good chunk of which was autogenerated), it's even more trivially simple to make an API like this return the wrong value.

kaniini:
My threat model is not evil sysadmins, it's techbros (with all due respect, such as yourself) using elasticsearch to snoop through the messages of marginalized people.  I do not wish for the marginalized people I manage instances for to have their messages indexed in this way by a remote server.  As such, if there is a signal that an instance is doing this, I would like to make use of it.

Issue #6686, opened 2018-03-07
Responsive image being cropped when increasing the browser window


Technowix:
On latest commit https://github.com/tootsuite/mastodon/commit/4ca60c665e02103b8730748c9b82b52bcd0d8d1c
I'm using Firefox 58

When i open an image, it show me the picture well 
![image](https://user-images.githubusercontent.com/1571367/37102989-7cecd73e-2229-11e8-860f-c4e00ee8b432.png)

But when I increase the size of the browser windows, the image try to fit in the new space / is rightly being responsive, but got hidden behind an invisible layer 
![image](https://user-images.githubusercontent.com/1571367/37103156-db191714-2229-11e8-9186-179d75cdb514.png)

ykzts:
cc @yuntan

Technowix:
https://maly.io/@marsxyz/99641023470652550

yuntan:
Confirmed.

Gargron:
I suspect it has something to do with using document.clientHeight instead of CSS with vh. I would argue it's better to use a height smaller than 100vh to account for browser title bar on mobile to avoid setting the height in JS. My impression from using the RCs for a few days is that pictures that stretch the entire screen on desktop are very overwhelming (I have a large screen)

Technowix:
Same, maybe profit for the screen-space when the size of the windows became to be small, but put some limits.
In addition, when we reduce the size of the window, the picture also get cropped :
![image](https://user-images.githubusercontent.com/1571367/37184531-0f3aa0ea-233c-11e8-859c-2403409dc745.png)
![image](https://user-images.githubusercontent.com/1571367/37184538-1a837efe-233c-11e8-9e7f-8026b1f44571.png)

yuntan:
> I suspect it has something to do with using document.clientHeight instead of CSS with vh. I would argue it's better to use a height smaller than 100vh to account for browser title bar on mobile to avoid setting the height in JS.

OK, I'll try to use another method to avoid using JS.

> My impression from using the RCs for a few days is that pictures that stretch the entire screen on desktop are very overwhelming (I have a large screen)

In other respects, some users complain that the modal is more difficult to close than before when aspect ratio of screen and image is almost same (there is no margin). I think that putting margins at the sides of an image is wrong because screen width of mobile devices is narrow. Thus I'll put margins on top and bottom of image by `max-height: 80vh`.

PR #6723, opened 2018-03-10
Fix MediaModal height when browser window is resized

yuntan:
Fixed the bug reported in #6686.


akihikodaki (commented):
Is it really fine to listen resize event without debouncing? At least, you should show that the listener finishes in 16ms (1 / 60 second.)

yuntan:
I decided to avoid setting height by JS. I'll make another PR.

PR #6730, opened 2018-03-10
Adds port mapping for ElasticSearch docker container

ashfurrow:
Hello! Nice work on 2.3.0. I had some trouble getting ElasticSearch to work via Docker, and was only able to get Rails to connect to the container once adding the port mapping in this PR.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,8 @@ services:
 #      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
 #    networks:
 #      - internal_network
+#    ports:
+#      - "9200:9200"
 #### Uncomment to enable ES persistance
 ##    volumes:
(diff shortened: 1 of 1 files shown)


ykzts:
cc @akihikodaki

Gargron:
@ashfurrow Have you uncommented the "depends_on" item from other services? It's easy to miss. Ports should only matter for external access, which you absolutely do not want for ES (note how postgres and redis also don't have ports defined here). See also: https://github.com/tootsuite/mastodon/issues/6665#issuecomment-371759898

ashfurrow:
Oh interesting! I uncommented that as well but was still getting a "connection refused" error message. I'll try removing it later, once the `chew:deply` task has finished. Thanks for clarifying – sounds like this PR should be closed then.

PR #6732, opened 2018-03-10
Change search tooltip, sounds bad to me

mkljczk:


akihikodaki (changes requested):
I prefer the previous one, which is a canonical sentence.

UPDATE: note that I am not a native speaker of English.

Gargron:
>Simple search **that** returns

This sentence is wrong. I appreciate the concern but I think the current version is good enough.

PR #6735, opened 2018-03-11
Canonical link for media attachments

akihikodaki:
#6672 removed media link from the status text, but a user may still want to use the link. This utilizes ActivityStreams representation to include the canonical link.

### DIFF
--- a/app/serializers/activitypub/note_serializer.rb
+++ b/app/serializers/activitypub/note_serializer.rb
@@ -105,7 +105,26 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
     end
 
     def url
-      object.local? ? full_asset_url(object.file.url(:original, false)) : object.remote_url
+      if object.local?
+        if current_server.accepts_link?
+          [
+            {
+              type: 'Link',
+              href: full_asset_url(object.file.url(:original, false)),
+              rel: 'self',
+            },
+            {
+              type: 'Link',
+              href: media_proxy_url(object, :original),
+              rel: 'canonical',
+            },
+          ]
+        else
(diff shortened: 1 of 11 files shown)


app/serializers/activitypub/note_serializer.rb
```diff
@@ -105,7 +105,26 @@ def media_type
     end
 
     def url
-      object.local? ? full_asset_url(object.file.url(:original, false)) : object.remote_url
+      if object.local?
+        if current_server.accepts_link?
```
Gargron: This is essentially unworkable, We can't generate different JSON based on user-agent header. In the majority of cases we push the JSON somewhere, rather than it being fetched by someone.

However, it might be fine to add the two Link objects anyway? I remember that we had some code for getting the right href out of a list of links, I might be misremembering where it is applied though.

Anyway, either that works, or this PR cannot be merged. But on the upside, I don't really see the canonical links as being useful anymore?

Gargron (changes requested):

app/serializers/activitypub/note_serializer.rb
akihikodaki: Indeed it is wrong, but it can be fixed.

> But on the upside, I don't really see the canonical links as being useful anymore?

Well, I would like to know why media have canonical links first. I thought the canonical links are more reliable than the direct links because direct links could point to object storage, which are often ran by cloud providers and administrators have less control.

Gargron: >Well, I would like to know why media have canonical links

They were made to be put into text for non-empty statuses. You have a valid point about redirects, but that does put extra strain on Rails, so I don't think we'll be using that...

PR #6738, opened 2018-03-11
Change owner of public directiories in Docker entrypoint

akihikodaki:
Docker sets the owner of public directories to root if those directories
are missing, which prevents fresh installation.

This commit resolves the problem by:
* executing commands by administrator as root
* changing the onwer of /mastodon/public/system in the entry point

The chown operation in the entry point quickly finishes because it is
applied only to a directory.

Another difference from the old `docker_entrypoint.sh` is that it does not use `su-exec`, which removed by #6722. Instead, it simply `su` and `exec`. I confirmed it works as expected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     image: tootsuite/mastodon
     restart: always
     env_file: .env.production
-    command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    command: /mastodon/docker_entrypoint.sh bundle exec rails s -p 3000 -b '0.0.0.0'
     networks:
       - external_network
       - internal_network
@@ -55,7 +55,7 @@ services:
     image: tootsuite/mastodon
     restart: always
     env_file: .env.production
-    command: yarn start
+    command: /mastodon/docker_entrypoint.sh yarn start
     networks:
       - external_network
       - internal_network
@@ -70,7 +70,7 @@ services:
     image: tootsuite/mastodon
     restart: always
     env_file: .env.production
-    command: bundle exec sidekiq -q default -q mailers -q pull -q push
+    command: /mastodon/docker_entrypoint.sh bundle exec sidekiq -q default -q mailers -q pull -q push
     depends_on:
       - db
       - redis
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+chown mastodon:mastodon /mastodon/public/system
+exec su mastodon -c "exec /sbin/tini -- $*"
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,16 +60,8 @@ RUN bundle config build.nokogiri --with-iconv-lib=/usr/local/lib --with-iconv-in
  && yarn --pure-lockfile \
  && yarn cache clean
 
-RUN addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon \
- && mkdir -p /mastodon/public/system /mastodon/public/assets /mastodon/public/packs \
- && chown -R mastodon:mastodon /mastodon/public
+RUN addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon
 
 COPY . /mastodon
 
-RUN chown -R mastodon:mastodon /mastodon
-
 VOLUME /mastodon/public/system /mastodon/public/assets /mastodon/public/packs
-
-USER mastodon
-
-ENTRYPOINT ["/sbin/tini", "--"]


Gargron:
I don't see it. When the image is built, the folders that will become volumes are created and then chowned to mastodon:mastodon, so they should have correct permissions when they become volumes.

akihikodaki:
> When the image is built, the folders that will become volumes are created and then chowned to mastodon:mastodon

Well, I have no idea why it happens, but it is easy to reproduce. Run the following commands to reproduce the issue:

```
git clean -d -x -f public
docker-compose build
docker-compose run --rm web rake assets:precompile
```

akihikodaki:
A Docker contributor tells to "do the bind-mount, then `chown` inside container", so apparently it is an expected behavior.
https://github.com/moby/moby/issues/2259#issuecomment-26564115

Gargron:
I think it's better to maybe modify documentation to tell people to chown stuff than to go back to an entrypoint file and chowning on startup - that's what we wanted to avoid.

(By the way does the documentation have any info about SAFETY_ASSURED=1 which is now required for db:setup? That needs to be added too if it's not already there...)

akihikodaki:
> I think it's better to maybe modify documentation to tell people to chown stuff than to go back to an entrypoint file and chowning on startup - that's what we wanted to avoid.

We don't have to avoid that since it will be done in a moment. The old script was very, very slow because it chowned recursively, but my version chowns just one directory, without recursive operation. It is much faster than people typing `mkdir` and `chown`.

moritzheiber:
There are several things happening here, and I'm not sure whether we should cover all the angles:

1. We are mounting the host directories `public/assets`, `public/packs` and `public/system` inside the container. Docker assumes whatever content is inside a folder from the host takes precedence, including its permissions. If the folder doesn't exist before it creates it as root, i.e. the process running inside the container would no longer by able to access it. This only happens if you're using the `docker-compose.yml` outside of the project directory though.

2. If the folder exists its permissions are "transposed" (unless you are using UID/GID mapping) into the container, the folder will be owned by whatever UID/GID owns the folders on the host. Now that'll be less of an issue than having it owned by `root`, but still impractical.

My question here would be: Why aren't we using regular volumes instead of host mounts? From my understanding these mounted folders are supposed to be empty, right? If we're worried about persistence just mounting a named volume would be the easier solution.

akihikodaki:
> …If the folder doesn't exist before it creates it as root, i.e. the process running inside the container would no longer by able to access it. This only happens if you're using the docker-compose.yml outside of the project directory though.

I confirmed the problem with `docker-compose.yml` in the project directory.

> My question here would be: Why aren't we using regular volumes instead of host mounts? From my understanding these mounted folders are supposed to be empty, right? If we're worried about persistence just mounting a named volume would be the easier solution.

Because those directories should be accessible from as file server outside of the containers such as nginx.

moritzheiber:
Then let's have either the `public` directory as a volume or move the content that needs to accessed out of the `public` folder. Mounting host volumes just because of this is an anti-pattern, IMHO. With this `ENTRYPOINT` script we are just trying to work around it.

akihikodaki:
> Then let's have either the `public` directory as a volume or move the content that needs to accessed out of the `public` folder. Mounting host volumes just because of this is an anti-pattern, IMHO.

I am not sure why changing the path of the volume to `public` or out of `public` allows to avoid the anti-pattern. I don't see the difference.

moritzheiber:
The different is that the files necessary for nginx to be accessed are then living in a named volume, under control by the container, instead of the on host machine (which we need the mounts for, apparently).

Gargron:
I am not familiar enough with named volumes (I think this got added after I learned Docker...), I just don't want to put existing setups through any new hoops. As long as the solution does not require manual intervention to avoid data loss, I'm okay with it.

akihikodaki:
Let me clarify that I intend to include this in v2.3.1. I want to make the change minimal.

docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     image: tootsuite/mastodon
     restart: always
     env_file: .env.production
-    command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    command: /mastodon/docker_entrypoint.sh bundle exec rails s -p 3000 -b '0.0.0.0'
```
Gargron: See what was undone in be9bab171dc2b1fe43bc742decb71f64541ca347: the entrypoint does not need to be hardcoded in docker-compose.yml

Gargron (changes requested):

docker-compose.yml
akihikodaki: It does because we do not want to use the entrypoint for `docker-compose run`. It should only be used when running a server process.

Gargron: People who don't use docker-compose.override.yml method will get merge conflicts from this change

Also, strictly speaking the streaming API does not need any directory permissions, I think.

What if someone runs using Docker but not docker-compose? People only need to chown once per installation lifetime, so I think that's a simpler solution than re-adding an entrypoint in this way...

akihikodaki: > People who don't use docker-compose.override.yml method will get merge conflicts from this change

The documentation suggests only to modify the file for volumes, and such changes will not conflict with this change.

> Also, strictly speaking the streaming API does not need any directory permissions, I think.

The script is used to change the permission from `root` to `mastodon`.

> What if someone runs using Docker but not docker-compose? People only need to chown once per installation lifetime, so I think that's a simpler solution than re-adding an entrypoint in this way...

If there is any people who runs without docker-compose, they should be able to solve their own problems by themselves. They are not following the documentation anyway.

> People only need to chown once per installation lifetime, so I think that's a simpler solution than re-adding an entrypoint in this way...

Why not make it simple for people trying out new installation rather than for people editing much of `docker-compose.yml` or managing containers without docker-compose?

docker_entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+chown mastodon:mastodon /mastodon/public/system
```
Wonderfall: https://github.com/tootsuite/mastodon/issues/3194 is why this part got removed.
By the way, the old entrypoint was also targeting this single directory. So I'm a bit confused.

PR #6747, opened 2018-03-11
Use vh instead of clientHeight for media modals

### LINKED ISSUE #6686: Responsive image being cropped when increasing the browser window

Gargron:
Fix #6686 

cc @yuntan


yuntan:
ReactSwipeableViews overflows from screen.

screenrecord_2018-03-12-08-34-26

(tested environment: Xperia Z5 compact, screen size 720×1280, image size 500×1000)

PR #6752, opened 2018-03-12
Sort contributors from oldest to newest

akihikodaki:
This makes easy to update the list. A new contributor will just prepend his
identity to the list.

Here is a script to generate the list. If you prefer a different order you can edit this to generate a list in the order.
https://gist.github.com/akihikodaki/8b3c4fa8a238cf3442e48ce2f342e327


Gargron:
Isn't oldest to newest with append more obvious/natural?

akihikodaki:
I just used the order of timeline.

akihikodaki:
efbea70e8d89dfe932a93ddb2d6f3533dfc6ae79: from oldest to newest, as suggested at https://github.com/tootsuite/mastodon/pull/6752#issuecomment-372282670.

Gargron:
FYI, this isn't oldest to newest still, it's newest to oldest (the opposite). A new contributor can just append their name to the end of the list. I think the list we have currently is already oldest to newest so no reason for this PR...

Issue #4093, opened 2017-07-07
Endpoint for last read notification


DagAgren:
When using multiple clients (such as web and mobile), it would be nice to have a count of unread notifications that is synced between both.

There are many ways this could be implemented. A simple solution off the top of my head is an endpoint for "ID of last read notification" which you can POST and maybe GET, and and "unread": field on Notification objects that is set to `id > lastReadId` (that check could be done locally, of course, but that would require 2 GETs instead of one for each update).

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

PR #6756, opened 2018-03-12
Track unread notification count across sessions

akihikodaki:


Gargron:
Related: #4093

streaming/index.js
```diff
@@ -381,6 +381,23 @@ const startWorker = (workerId) => {
...
   };
 
+  const streamNotificationEnd = req => {
+    pgPool.connect((err, client, done) => {
+      if (err) {
+        log.error(err);
+        return;
+      }
+
+      client.query('UPDATE users SET last_read_notification_id = (SELECT id FROM notifications ORDER BY id DESC LIMIT 1) FROM oauth_access_tokens WHERE oauth_access_tokens.resource_owner_id = users.id AND oauth_access_tokens.token = $1 AND oauth_access_tokens.reading_notifications', [req.token], err => {
```
Gargron: I don't think performing UPDATEs through the streaming API is a good idea. First, the frequency with which the UPDATEs could be executed, and secondly because so far it was safe (and smart) to run the streaming API from PostgreSQL read-only replica, but with this change it wouldn't be possible.

At last: Streamed does not mean seen... E.g. delivered vs opened indicators in chat apps.

db/migrate/20180310000000_add_last_read_notification_id_to_users.rb
```diff
@@ -0,0 +1,25 @@
...
+  disable_ddl_transaction!
+
+  def change
+    add_column :users, :last_read_notification_id, :bigint
+
+    reversible do |dir|
+      dir.up do
+        safety_assured do
+          add_column_with_default :oauth_access_tokens, :reading_notifications, :boolean, allow_null: false, default: false
+          update_column_in_batches :users, :last_read_notification_id, 9e9
```
Gargron: Would it not be easier to use -1 as the "magic value"? I am actually not sure if the space requirements for these numbers are equal or not, but I have a distrust of "max integer"

config/routes.rb
```diff
@@ -266,6 +266,7 @@
         collection do
           post :clear
           post :dismiss
+          resource :unread, module: :notifications, only: [:show, :update]
```
Gargron: It might make more sense to call the API method "read"? "I have just *read* this notification ID" feels more natural.

Gargron (changes requested):
I am sorry for a very late review, but better late than never! :pray:

streaming/index.js
akihikodaki: Sorry for late reply. Do you have any alternative idea?

PR #6764, opened 2018-03-13
Change columns in notifications nonnullable

akihikodaki:


ClearlyClaire (approved):
This is only loosely related, but at some point, there was a possibility of race condition causing some notifications to not have an underlying activity. I think that it is still the case, and this PR won't help with it.

PR #6772, opened 2018-03-13
Change the title of spoiler button by state

akihikodaki:
The title will give the precise representation of the current state. It
would be helpful for blind people.

See also #6771.


Gargron (approved):

Gargron:
Conflicts tho

PR #6773, opened 2018-03-13
Add ?no_preview param to hide OpenGraph preview for status

Gargron:
See https://github.com/tootsuite/mastodon/pull/6479#issuecomment-372607251

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -29,6 +29,10 @@ module StreamEntriesHelper
     [prepend_str, account.note].join(' · ')
   end
 
+  def hide_opengraph_preview?
+    params.key?(:no_preview)
+  end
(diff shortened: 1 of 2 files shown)


app/helpers/stream_entries_helper.rb
```diff
@@ -29,6 +29,10 @@ def account_description(account)
     [prepend_str, account.note].join(' · ')
   end
 
+  def hide_opengraph_preview?
+    params.key?(:no_preview)
```
akihikodaki: Why don't you use `ActiveModel::Type::Boolean.new.cast`?

akihikodaki (changes requested):

akihikodaki:
In #6479, the usefulness of this feature is rejected and should not be merged till it is proven.

PR #6775, opened 2018-03-13
Enable following back from the Follow requests column

fvh-P:
Currently, when you approve or reject a follow request from any account, that account 
immediately disappears from the column. It is inconvenient to following back that account.

This PR makes the process —receiving and approving the follow requests, and following back— **seamless**, by leaving the account list displayed and install an account relationship button there.

relationship

### DIFF
--- a/app/javascript/mastodon/containers/account_relationship_button_container.js
+++ b/app/javascript/mastodon/containers/account_relationship_button_container.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
+import AccountRelationshipButton from '../components/account_relationship_button';
+import {
+  followAccount,
+  unfollowAccount,
+  blockAccount,
+  unblockAccount,
+  muteAccount,
+  unmuteAccount,
+} from '../actions/accounts';
+import { openModal } from '../actions/modal';
+import { initMuteModal } from '../actions/mutes';
+import { unfollowModal } from '../initial_state';
+import { makeGetAccount } from '../selectors';
+
+const messages = defineMessages({
+  unfollowConfirm: { id: 'confirmations.unfollow.confirm', defaultMessage: 'Unfollow' },
+});
+
+const makeMapStateToProps = () => {
+  const getAccount = makeGetAccount();
+
+  const mapStateToProps = (state, { id }) => ({
+    account: getAccount(state, id),
+  });
+
+  return mapStateToProps;
+};
+
+const mapDispatchToProps = (dispatch, { intl }) => ({
+
+  onFollow (account) {
+    if (account.getIn(['relationship', 'following']) || account.getIn(['relationship', 'requested'])) {
+      if (unfollowModal) {
+        dispatch(openModal('CONFIRM', {
+          message: <FormattedMessage id='confirmations.unfollow.message' defaultMessage='Are you sure you want to unfollow {name}?' values={{ name: <strong>@{account.get('acct')}</strong> }} />,
+          confirm: intl.formatMessage(messages.unfollowConfirm),
+          onConfirm: () => dispatch(unfollowAccount(account.get('id'))),
+        }));
+      } else {
+        dispatch(unfollowAccount(account.get('id')));
+      }
+    } else {
+      dispatch(followAccount(account.get('id')));
+    }
+  },
+
+  onBlock (account) {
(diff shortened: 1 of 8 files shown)


app/javascript/mastodon/features/follow_requests/index.js
```diff
@@ -59,9 +72,9 @@ export default class FollowRequests extends ImmutablePureComponent {
 
         <ScrollContainer scrollKey='follow_requests'>
           <div className='scrollable' onScroll={this.handleScroll}>
-            {accountIds.map(id =>
+            {this.state.accountIds ? this.state.accountIds.map(id =>
```
ykzts: `{this.state.accountIds ? this.state.accountIds.map(id => {}) : null` -> `{this.state.accountIds && this.state.accountIds.map(id => {})`

fvh-P: okay, thanks.

app/javascript/mastodon/features/follow_requests/components/account_authorize.js
```diff
@@ -23,10 +24,43 @@ export default class AccountAuthorize extends ImmutablePureComponent {
...
+      const disabled = requestState !== 'requested';
+      return (
+        <div className='account--panel'>
+          <IconButton className={`account--panel__button ${requestState === 'authorized' ? requestState : ''}`} title={intl.formatMessage(messages.authorize)} icon='check' onClick={this.onClickAuthorize} disabled={disabled} />
+          <IconButton className={`account--panel__button ${requestState === 'rejected' ? requestState : ''}`} title={intl.formatMessage(messages.reject)} icon='times' onClick={this.onClickReject} disabled={disabled} />
+        </div>
+      );
+    };
+
+    if (!account) {
```
akihikodaki: `account` is marked required but it seems for a case the property is missing. Which is correct?

app/javascript/mastodon/features/follow_requests/components/account_authorize.js
```diff
@@ -23,10 +24,43 @@ export default class AccountAuthorize extends ImmutablePureComponent {
...
+      return <div />;
```
akihikodaki: Why not `null`?

app/javascript/mastodon/containers/account_relationship_button_container.js
```diff
@@ -0,0 +1,72 @@
...
+import { makeGetAccount } from '../selectors';
+
+const messages = defineMessages({
+  unfollowConfirm: { id: 'confirmations.unfollow.confirm', defaultMessage: 'Unfollow' },
+});
+
+const makeMapStateToProps = () => {
+  const getAccount = makeGetAccount();
+
+  const mapStateToProps = (state, props) => ({
```
akihikodaki: The second argument is named `props` in `mapStateToProps`, but it is destructed in `mapDispatchToProps`. It is better to keep them consistent.

akihikodaki (changes requested):

app/javascript/mastodon/containers/account_relationship_button_container.js
fvh-P: Sorry, what should I do specifically?

app/javascript/mastodon/features/follow_requests/components/account_authorize.js
fvh-P: Oh, it's my carelessness. This part is redundant.

fvh-P: By the way, Account has the same code block for the case you pointed. I've not modified this, but is it better to remove...? 
https://github.com/tootsuite/mastodon/blob/3987bd18a4bfaee49ea9a3439fc51610aeca6c2c/app/javascript/mastodon/components/account.js#L58-L60

akihikodaki: You are right. But it is out of scope of this pull request. It should be handled in a different pull request.

app/javascript/mastodon/containers/account_relationship_button_container.js
akihikodaki: Replace `props` with `{ id }`. You will be able to refer to the ID just with `id` instead of `props.id`. That's what `mapDispatchToProps` does.

fvh-P:
Thank you for your reviews!
I modified some parts of PR. Please check again and point if any.

PR #6789, opened 2018-03-15
Enable updating additional account information from user preferences via rest api

### LINKED ISSUE #6553: Allow syncing of privacy/sensitive preferences across apps
This is an often requested feature for Tusky. It is already possible to retrieve the preferences via `GET /api/v1/accounts/verify_credentials`, but not to set them via the api. It would be super awesome if that feature could be added. Unfortunately I don't understand enough of Ruby to send a pull request myself.

* * * *

Tak:
Resolves #6553

### DIFF
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -13,6 +13,7 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
   def update
     @account = current_account
     UpdateAccountService.new.call(@account, account_params, raise_error: true)
+    UserSettingsDecorator.new(current_user).update(user_settings_params) if user_settings_params
     ActivityPub::UpdateDistributionWorker.perform_async(@account.id)
     render json: @account, serializer: REST::CredentialAccountSerializer
   end
@@ -21,5 +22,21 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
 
   def account_params
     params.permit(:display_name, :note, :avatar, :header, :locked)
+  end
+
+  def user_settings_params
+    return nil unless params.key?(:source)
+
+    source_params = params.require(:source).permit(
+      :privacy,
+      :sensitive
+    )
+    sensitive = source_params.key?(:sensitive) ? source_params[:sensitive] : object.user.setting_default_sensitive
+    return ActionController::Parameters.new(
+      user: {
+        setting_default_privacy: source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
+        setting_default_sensitive: sensitive ? '1' : '0', # database doesn't like boolean values for default_sensitive
+      }
+    ).require(:user).permit(:setting_default_privacy, :setting_default_sensitive).to_h
   end
 end
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -31,12 +31,9 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
       :privacy,
       :sensitive
     )
-    sensitive = source_params.key?(:sensitive) ? source_params[:sensitive] : object.user.setting_default_sensitive
-    ActionController::Parameters.new(
-      user: {
-        setting_default_privacy: source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
-        setting_default_sensitive: sensitive ? '1' : '0', # database doesn't like boolean values for default_sensitive
-      }
-    ).require(:user).permit(:setting_default_privacy, :setting_default_sensitive).to_h
+    {
+      'setting_default_privacy' => source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
+      'setting_default_sensitive' => source_params.key?(:sensitive) ? source_params[:sensitive] : object.user.setting_default_sensitive,
+    }
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -22,4 +23,20 @@ def update
   def account_params
     params.permit(:display_name, :note, :avatar, :header, :locked)
   end
+
+  def user_settings_params
+    return nil unless params.key?(:source)
+
+    source_params = params.require(:source).permit(
+      :privacy,
```
akihikodaki: Why renaming `setting_default_privacy` as `privacy`? Indeed `setting` is kind of duplicate and fine to remove, but `default` is not such a duplicate and should be preserved.

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -22,4 +23,20 @@ def update
...
+      :sensitive
+    )
+    sensitive = source_params.key?(:sensitive) ? source_params[:sensitive] : object.user.setting_default_sensitive
+    ActionController::Parameters.new(
+      user: {
+        setting_default_privacy: source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
+        setting_default_sensitive: sensitive ? '1' : '0', # database doesn't like boolean values for default_sensitive
+      }
+    ).require(:user).permit(:setting_default_privacy, :setting_default_sensitive).to_h
```
akihikodaki: Why not create a hash at first place if you eventually converts to hash?

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -22,4 +23,20 @@ def update
...
+
+    source_params = params.require(:source).permit(
+      :privacy,
+      :sensitive
+    )
+    sensitive = source_params.key?(:sensitive) ? source_params[:sensitive] : object.user.setting_default_sensitive
+    ActionController::Parameters.new(
+      user: {
+        setting_default_privacy: source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
+        setting_default_sensitive: sensitive ? '1' : '0', # database doesn't like boolean values for default_sensitive
```
akihikodaki: It's not database who doesn't like boolean value, but `UserSettingsDecorator` is. `UserSettingsDecorator` should be updated instead of adding a workaround.

spec/controllers/api/v1/accounts/credentials_controller_spec.rb
```diff
@@ -42,6 +46,8 @@
           expect(user.account.note).to eq("Hi!\n\nToot toot!")
           expect(user.account.avatar).to exist
           expect(user.account.header).to exist
+          expect(user.setting_default_privacy).to eq('unlisted')
```
akihikodaki: Thanks for spec. I appreciate it.

akihikodaki (changes requested):

app/controllers/api/v1/accounts/credentials_controller.rb
Tak: This is because that's the way it goes out in `verify_credentials` - the user of the api just modifies the values in the structure she receives and sends it back.

app/controllers/api/v1/accounts/credentials_controller.rb
Tak: Ok, I can do that. I wasn't sure about the additional effects of making a change like that.

app/controllers/api/v1/accounts/credentials_controller.rb
Tak: I tried just returning a hash that seemed identical to me, but the update failed when I did that, and I'm not sure why. I assumed there was some hidden side effect of doing it this way...

Tak: ...and there is - the key symbols are converted to strings. Updating...

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -22,4 +23,17 @@ def update
   def account_params
     params.permit(:display_name, :note, :avatar, :header, :locked)
   end
+
+  def user_settings_params
+    return nil unless params.key?(:source)
+
+    source_params = params.require(:source).permit(
```
akihikodaki: You don't need `permit` here.

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -22,4 +23,17 @@ def update
...
+      :privacy,
+      :sensitive
+    )
+    {
+      'setting_default_privacy' => source_params.key?(:privacy) ? source_params[:privacy] : object.user.setting_default_privacy,
```
akihikodaki: `fetch` method should be convenient for the purpose.

app/lib/user_settings_decorator.rb
```diff
@@ -8,7 +8,7 @@ def initialize(user)
   end
 
   def update(settings)
-    @settings = settings
+    @settings = decoerce_values(settings)
```
akihikodaki: Instead of adding decoercing operation, let's use `ActiveModel::Type::Boolean.new.cast` in `coerce_values`. It accepts both of boolean and numbers.

akihikodaki (changes requested):

Tak:
Is there any further action I need to take on this? Or it's just waiting for final review?

akihikodaki (approved):

PR #6325, opened 2018-01-22
Note if the user is already following the target when authorizing follow

akihikodaki:
screenshot-2018-1-22 follow tenshi - mastodon dev
UPDATE: Resolves #4536.

PR #6791, opened 2018-03-15
Add already_following message to the authorize_follows dialog

### LINKED ISSUE #4536: Please show messages or buttons on remote follow dialog even if already followed
On the master (#4523), remote follow dialog is like below when I try to follow whom I'm already following.

image

and I thought, "oh, there are neither buttons nor messages. Is this a bug? or am I wrong?".

How about having message like "You're already following this user", or Go to Web / Return to the profile buttons like success page?

DavidLibeau:
Resolve #4536 

_Tell me if I did something wrong!_

### DIFF
--- a/app/views/authorize_follows/show.html.haml
+++ b/app/views/authorize_follows/show.html.haml
@@ -3,6 +3,8 @@
 
 .form-container
   .follow-prompt
+    - if current_account.following?(@account)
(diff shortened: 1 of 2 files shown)


akihikodaki (commented):
Duplicate of #6325.

DavidLibeau:
#6325

PR #6799, opened 2018-03-17
[Nanobox] Stream backups to the warehouse

danhunsaker:
The `curl` docs are terrible. Use `-X POST -T` instead of `--data-binary`, to avoid loading entire backups into memory _before_ transferring to the warehouse, and just stream the data across as it comes in.


akihikodaki (commented):
Why `>&2`?

Gargron (approved):

danhunsaker:
@akihikodaki To ensure any output gets to the right part of the internal logging system.

PR #6802, opened 2018-03-17
Cache HTML page with Service Worker

akihikodaki:
This is the first step to make Mastodon work offline. It is also required
by Chromium to trigger Web Manifest automated install prompt.

### DIFF
--- a/app/javascript/mastodon/service_worker/entry.js
+++ b/app/javascript/mastodon/service_worker/entry.js
@@ -1,10 +1,30 @@
 import './web_push_notifications';
 
+function fetchRoot() {
+  return fetch('/', { credentials: 'include' });
+}
+
 // Cause a new version of a registered Service Worker to replace an existing one
 // that is already installed, and replace the currently active worker on open pages.
 self.addEventListener('install', function(event) {
-  event.waitUntil(self.skipWaiting());
+  const promises = Promise.all([caches.open('mastodon-web'), fetchRoot()]);
+  const asyncAdd = promises.then(([cache, root]) => cache.put('/', root));
+
+  event.waitUntil(asyncAdd);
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
> TypeError: ServiceWorker script at https://[redacted]/sw.js for scope https://[redacted]/ encountered an error during installation.

@akihikodaki Why? Did you test this?

akihikodaki:
@Gargron I have tested on Chromium. I'll test on Firefox tomorrow (it's at midnight in Japan.)

nolanlawson:
IIRC the reason I didn't implement HTML caching when I added ServiceWorker support was because the csrf token is embedded in the HTML, and it was unclear to me what would happen if we just cached that forever:

```html
<meta name="csrf-token" content="<token>" />
```

For the Firefox vs Chrome thing you may also need to add `redirect: follow`. I seem to recall Chrome and Firefox differing on this one: https://stackoverflow.com/a/40277730

Gargron:
I don't think caching that is different to, for example, keeping the page open for 2 days, etc... A token's a token. It's probably fine.

akihikodaki:
It does not matter for online environment because it fetches from the remote first even if cache exists.
It causes the problem you suspect for offline environment. It is still possible to recover by reloading the page.

The difference between Firefox and Chrome is interesting. I'll see if it caused the problem.

akihikodaki:
I have just tested but I had no such an error. My user agent is:
```
Mozilla/5.0 (X11; Linux x86_64; rv:59.0) Gecko/20100101 Firefox/59.0
```

I checked the code, and I don't think `redirect: follow` should matter because `/` should not redirect.

PR #6804, opened 2018-03-17
Do not show network error too often if offline

akihikodaki:


Gargron:
I don't understand why the new `passive` option is necessary or what it does, please explain

akihikodaki:
`passive` option is set when the fetch request is a side effect rather than the main effect. Namely, it is set for requests made by scrolling. The error message will be suppressed by setting the option, and a user can scroll page without being bothered by errors.

unarist:
Requests made by scrolling are due to scrollToBottom handlers? I think those are not so often and using debounce seems enough.

Also we may want to adjust debounce for any errors on fetching not only due to offline, if it bothers.

akihikodaki:
> Requests made by scrolling are due to scrollToBottom handlers? I think those are not so often and using debounce seems enough.

It is not debounced, but it is throttled. The duration is 150ms, which is quite a small for humans.

> Also we may want to adjust debounce for any errors on fetching not only due to offline, if it bothers.

The feature is already implemented.

Gargron:
There are conflicts because of your other PRs, please fix and I'll merge

PR #6805, opened 2018-03-17
Do not drop follow relationships on user-specific domain blocks

ClearlyClaire:
This behavior is extremely confusing, especially since the wording (hiding)
does not hint at this and that follow relations from the user to blocked accounts
are preserved.

### DIFF
--- a/spec/services/block_domain_from_account_service_spec.rb
+++ b/spec/services/block_domain_from_account_service_spec.rb
@@ -10,10 +10,4 @@ RSpec.describe BlockDomainFromAccountService, type: :service do
     subject.call(alice, 'evil.org')
     expect(alice.domain_blocking?('evil.org')).to be true
   end
-
-  it 'purge followers from blocked domain' do
-    wolf.follow!(alice)
-    subject.call(alice, 'evil.org')
-    expect(wolf.following?(alice)).to be false
-  end
 end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -126,6 +126,7 @@ class Account < ApplicationRecord
   scope :matches_username, ->(value) { where(arel_table[:username].matches("#{value}%")) }
   scope :matches_display_name, ->(value) { where(arel_table[:display_name].matches("#{value}%")) }
   scope :matches_domain, ->(value) { where(arel_table[:domain].matches("%#{value}%")) }
+  scope :not_domain_blocked_by_account, ->(account) { where('domain IS NULL OR domain NOT IN (?)', account.excluded_from_timeline_domains) if account.excluded_from_timeline_domains.present? }
 
   delegate :email,
            :unconfirmed_email,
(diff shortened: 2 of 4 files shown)


spec/services/block_domain_from_account_service_spec.rb
```diff
@@ -11,9 +11,9 @@
     expect(alice.domain_blocking?('evil.org')).to be true
   end
 
-  it 'purge followers from blocked domain' do
```
akihikodaki: No need of spec for behavior without code. Preserving such will keep spec growing.

akihikodaki (changes requested):

spec/services/block_domain_from_account_service_spec.rb
ThibG: Well, ok, but it confirms the behavior is intentional, as opposed to leading it unspecified

ThibG: Updated to match the requested changes, although I kinda disagree, as mentioned earlier.

akihikodaki: I'm pretty sure that anyone who found this change is undesired can figure out that follow relations from the blocking users are preserved. It is fine in the case.

akihikodaki (approved):

Gargron:
Okay. here's an issue: Dropping followers from the domain you're hiding was part of the feature, since that makes sure your toots don't go to that domain (directly). I recently got asked by someone to make that a thing, they didn't know it was already a thing. So while I can see there's confusion/incosistencies there, I feel like this is the wrong direction of fixing them? The alternatives would be: fixing the wording, or making the *active* relationships go away too...

ThibG:
Hm, ok, I can see that point. But as your comment highlights, this is currently very confusing, i.e., even if that person asked you that feature, they thought it wasn't already how it worked, due to the wording.

Regarding dropping existing relationships, I guess if we decide to actually drop them, we should drop both passive and active relationships, and probably not silently so.

I'm still leaning towards not dropping them and not changing the wording. But I can see value with both alternatives, so maybe we should add new options, to mimick user block and silences?

trwnh:
+1 for the user confusion: I assumed that muting an instance would *only* hide all statuses originating from there. I personally explicitly do NOT want to drop following OR follower relationships after muting an instance. The language specifically refers to "hiding" a domain, not blocking it. Muting an instance should be roughly equivalent to muting every single user on that instance.

Gargron:
>What we could do without changing the existing behavior of not sending toots to muted instances is check whether they are muted, without dropping the relationship (so that if the user stops muting an instance, their following relationships remain intact).

I can agree with that.

ThibG:
Updated the PR to not distribute to domain-blocked followers. I'm not sure how efficient that is, though.

Issue #6761, opened 2018-03-12
Attempting to connect to IPv6 address on IPv4-only host 


Jehops:
This is with 2.3.1rc2, but I believe the issue started earlier.

```
2018-03-12T13:21:45.094Z 86844 TID-a3oqxw ActivityPub::ProcessingWorker JID-d54802549aadb871df136517 INFO: start
2018-03-12T13:21:45.161Z 86844 TID-a3oqxw ActivityPub::ProcessingWorker JID-d54802549aadb871df136517 INFO: fail: 0.067 sec
[...]

PR #6813, opened 2018-03-18
request: in the event of failure, try other IPs (#6761)

rtucker:
In the case where a name has multiple A/AAAA records, we should
try subsequent records instead of immediately failing when we have a
failure on the first IP address.

This significantly improves delivery success when there are network
connectivity problems affecting only IPv4 or IPv6.

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -94,9 +94,16 @@ class Request
   class Socket < TCPSocket
     class << self
       def open(host, *args)
-        address = IPSocket.getaddress(host)
-        raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address)
-        super address, *args
+        outer_e = nil
+        Addrinfo::foreach(host, nil, nil, :SOCK_STREAM) do |address|
+          raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
+          begin
+            return super address.ip_address, *args
+          rescue => e
+            outer_e = e
+          end
+        end
+        raise outer_e if outer_e
       end
 
       alias new open
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -96,8 +96,8 @@ class Request
       def open(host, *args)
         outer_e = nil
         Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
-          raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
           begin
+            raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
             return super address.ip_address, *args
           rescue => e
             outer_e = e
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -95,7 +95,7 @@ class Request
     class << self
       def open(host, *args)
         outer_e = nil
-        Addrinfo::foreach(host, nil, nil, :SOCK_STREAM) do |address|
+        Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
           raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
           begin
             return super address.ip_address, *args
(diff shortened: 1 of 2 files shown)


rtucker:
The CI build is failing on:

`rspec ./spec/lib/request_spec.rb:63 # Request#perform with private host raises Mastodon::ValidationError`

This is due to the test case being tied to IPSocket.getaddress.  I don't currently have a good development environment for running the tests, so feel free to fix this up.

Also there ought to be a test case to test for the round-robin working as expected.

app/lib/request.rb
```diff
@@ -94,9 +94,16 @@ def http_client
   class Socket < TCPSocket
     class << self
       def open(host, *args)
-        address = IPSocket.getaddress(host)
-        raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address)
-        super address, *args
+        outer_e = nil
+        Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
+          raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
```
akihikodaki: Please move this line in the `begin...rescue` section. Such a case that one of addresses are invalid but the others are valid should have been tolerated in releases prior to v2.3.0.

akihikodaki (dismissed):

rtucker:
Currently stalled on:

```
/home/travis/build/tootsuite/mastodon/vendor/bundle/ruby/2.4.0/gems/http-3.0.0/lib/http/connection.rb:126:in `close': HTTP::Timeout::Null#closed? at /home/travis/.rvm/rubies/ruby-2.4.2/lib/ruby/2.4.0/forwardable.rb:157 forwarding to private method NilClass#closed?
                                                                                
  1) Request#perform with private host raises Mastodon::ValidationError
     Failure/Error: expect{ subject.perform }.to raise_error Mastodon::ValidationError
     
       expected Mastodon::ValidationError, got #<NoMethodError: undefined method `closed?' for nil:NilClass on http://example.com/> with backtrace:
         # /home/travis/.rvm/rubies/ruby-2.4.2/lib/ruby/2.4.0/forwardable.rb:227:in `closed?'
         # 
         #   Showing full backtrace because every line was filtered out.
         #   See docs for RSpec::Configuration#backtrace_exclusion_patterns and
         #   RSpec::Configuration#backtrace_inclusion_patterns for more information.
     # ./spec/lib/request_spec.rb:74:in `block (4 levels) in <top (required)>'
     # ./spec/lib/request_spec.rb:66:in `block (4 levels) in <top (required)>'
```

My Ruby/RSpec experience is lacking, so I'm going to take a break for the night.

spec/lib/request_spec.rb
```diff
@@ -48,6 +48,13 @@
         expect(a_request(:get, 'http://example.com')).to have_been_made.once
       end
 
+      it 'executes a HTTP request when the first address is private' do
+        allow(Addrinfo).to receive(:foreach).with('example.com', nil, nil, :SOCK_STREAM).and_return([
```
akihikodaki: See https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/yielding.

akihikodaki (changes requested):

app/lib/request.rb
```diff
@@ -94,9 +94,16 @@ def http_client
   class Socket < TCPSocket
```
Gargron: @akihikodaki Isn't there different classes for Socket and SSLSocket?

akihikodaki: `SSLSocket` uses `TCPSocket`, so patching only `TCPSocket` is enough.

Gargron: But this doesn't patch TCPSocket, only our own Socket class, no?

akihikodaki: `SSLSocket` takes a `TCPSocket` instance as an argument when initializes, and HTTP gem uses our patched `TCPSocket` to create an instance for the argument. That's how it works.

Gargron: Okay, thanks.

akihikodaki (approved):
I fixed the spec, and now it is ready to merge. I hope it will be in v2.3.2.

PR #6814, opened 2018-03-18
Introduce redux-persist

akihikodaki:
This makes some Redux state persistent and let it work offline.

Some caveats:
* redux-immutable was removed because it is incompatible with redux-persist. It was still possible to remove the dependency because Mastodon does not use features of Immutable like `set` functions for the global state.
* Persistent immutable objects are serialized with `toJS` and deserialized with `fromJS`. Therefore persistent objects are limited to particular types. As far as I see, the objects to be persistent match the condition. [redux-persist-transform-immutable](https://github.com/rt2zz/redux-persist-transform-immutable) is a more sophisticated alternative, but I decided not to use it because of https://github.com/rt2zz/redux-persist-transform-immutable/issues/33 and the simple serialization works for this case.
* We must change the version of persistent store when we change the structure of it after applying this change.

### DIFF
--- a/app/javascript/mastodon/actions/accounts.js
+++ b/app/javascript/mastodon/actions/accounts.js
@@ -68,7 +68,7 @@ export function fetchAccount(id) {
   return (dispatch, getState) => {
     dispatch(fetchRelationships([id]));
 
-    if (getState().getIn(['accounts', id], null) !== null) {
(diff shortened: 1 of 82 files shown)


Gargron:
Yes, garbage collection is the hardest part because of normalization :(

akihikodaki:
Closes due to preference of #6826.

PR #6817, opened 2018-03-18
If DEFAULT_LOCALE is set, enforce it instead of HTTP request locale

### LINKED ISSUE #6784: Cannot set the interface to Portuguese only
http://masto.pt is a Portuguese only instance. However, as far as we can see, there is no way to lock the interface to Portuguese. The screenshot below  show the problem: coming from France with a french browser, I see a french interface around the instance description which says it is for Portuguese speakers.

masto.pt screenshot

Actually, since mastodon is translated into about 37 languages and dialects, it means our lone admin is supposed to somehow understand them in case a problem with a foreign user arise, which is not practical on such a small instance.

* * * *

Gargron:
Fix #6784 

Allows instances to be more obvious about being regional.

### DIFF
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -17,7 +17,11 @@ module Localized
   end
 
   def default_locale
-    request_locale || I18n.default_locale
+    if ENV['DEFAULT_LOCALE'].present?
+      I18n.default_locale
+    else
+      request_locale || I18n.default_locale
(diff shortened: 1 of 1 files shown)


ykzts (approved):
:heart:

nightpool:
Oh cool, I wasn't expecting this to happen but this means we can remove some custom cybrespace hacks.......

PR #6818, opened 2018-03-18
Revert #6479, hide sensitive text/images from OpenGraph previews

Gargron:
See: #6479

Display summary of attachments in description, and mark up content warning if present, e.g.:

    Attached: 3 images · Content warning: Dota 2

When text is not supposed to be hidden, it looks more like:

    Attached: 3 images

    Here is the text of the toot

With #6817, multilinguagility should be assured...

Other changes:

Instead of "Alice on Mastodon", use "Alice (@alice@example.com)" in OpenGraph titles, since "Mastodon" would be value of the OpenGraph `site_name`

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -29,6 +29,35 @@ module StreamEntriesHelper
     [prepend_str, account.note].join(' · ')
   end
 
+  def media_summary(status)
+    attachments = { image: 0, video: 0 }
+
+    status.media_attachments.each do |media|
+      if media.video?
+        attachments[:video] += 1
+      else
+        attachments[:image] += 1
+      end
+    end
+
+    text = attachments.to_a.reject { |_, value| value.zero? }.map { |key, value| t("statuses.attached.#{key}", count: value) }.join(' · ')
+
+    return if text.blank?
+
+    t('statuses.attached.description', attached: text)
+  end
+
+  def status_text_summary(status)
+    return if status.spoiler_text.blank?
+    t('statuses.content_warning', warning: status.spoiler_text)
+  end
+
+  def status_description(status)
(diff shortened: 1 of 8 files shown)


Cassolotl:
This sounds good. :) Could we see some screenshots? Also, thank you!

Gargron:
>Could we see some screenshots?

Of what, Discord? It's difficult to get those until this patch is deployed on a live instance.

Cassolotl:
Well, I would love to see some screenshots when it gets tested on a live instance, if that happens before this is merged.

Gargron:
@Cassolotl The examples are right there in the OP. That's how it's gonna look.

Cassolotl:
I just struggle to imagine it, so it would help me to see how it's going to look, to know if I will find this easy to understand, etc.

Gargron:
image

nightpool:
these changes look great!

do you think we want to update the page title as well or no?

image

Gargron:
How do you want to update the page title?

nightpool:
@Gargron I was just thinking about how you changed the account page title and was wondering if there was a similar change the status page title that made sense (`"summary" - Admin (@admin@localhost:3000)`, or something)

After thinking about possible options for a while I don't think it makes much sense though, so withdrawn.

Cassolotl:
@Gargron Thank you, this is really good. :)

nightpool (approved):

nightpool:
@Cassolotl avatar

KScl:
:+1: Thank you. <3

rtucker:
Now running in production on vulpine.club.  Here's how it looks in Slack:

screenshot 2018-03-18 16 28 11

Source toot: https://vulpine.club/@rey/99706788560684438

Looks good to me ^^

PR #6820, opened 2018-03-18
Fix ActivityPub http->https redirect following in DeliveryWorker

Gargron:
> HTTP::StateError: Delivery failed for http://mastodon.hk/inbox: can't follow 301 Moved Permanently redirect on http://mastodon.hk/inbox

Some people might have a problem where they initially started using Mastodon as HTTP, and then upgraded to HTTPS. The redirect breaks delivery. But we can manually retry the HTTPS version if we encounter a HTTP inbox.

Other solutions:

- Refuse to work with HTTP URIs altogether
- Do nothing since for a few versions it's not possible to run Mastodon in production without HTTPS, so such misconfigured instances should eventually disappear?

### DIFF
--- a/app/workers/activitypub/delivery_worker.rb
+++ b/app/workers/activitypub/delivery_worker.rb
@@ -12,7 +12,16 @@ class ActivityPub::DeliveryWorker
     @source_account = Account.find(source_account_id)
     @inbox_url      = inbox_url
 
-    perform_request
+    begin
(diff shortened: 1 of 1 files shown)


Gargron:
I'm not sure I like this code :weary:

akihikodaki:
I'm for the option "Refuse to work with HTTP URIs altogether".

PR #6823, opened 2018-03-18
SMTP_OPENSSL_VERIFY_MODE should not be set to 'peer' unless there is a client certificate

progval:
Otherwise, this gives the following error:

```
OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)
```

PS: I don't know how to write Ruby code, please check the syntax is ok

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -256,10 +256,11 @@ namespace :mastodon do
             q.modify :strip
           end
 
-          env['SMTP_OPENSSL_VERIFY_MODE'] = prompt.ask('SMTP OpenSSL verify mode:') do |q|
-            q.required
-            q.default 'peer'
-            q.modify :strip
+          if prompt.yes?('Do you want to use a client certificate to authenticate to the server?', default: false)
+            env['SMTP_OPENSSL_VERIFY_MODE'] = 'peer'
+            prompt.ask('Path to the client certificate:') do |q|
+              q.required
+              q.modify :strip
           end
         end
 


lib/tasks/mastodon.rake
```diff
@@ -256,10 +256,11 @@ namespace :mastodon do
             q.modify :strip
           end
 
-          env['SMTP_OPENSSL_VERIFY_MODE'] = prompt.ask('SMTP OpenSSL verify mode:') do |q|
-            q.required
-            q.default 'peer'
-            q.modify :strip
+          if prompt.yes?('Do you want to use a client certificate to authenticate to the server?', default: false)
+            env['SMTP_OPENSSL_VERIFY_MODE'] = 'peer'
+            prompt.ask('Path to the client certificate:') do |q|
```
nightpool: This client certificate never gets put into the environment anywhere?

ProgVal: Oops, indeed. I'm going to fix it

ProgVal: Hmm, actually, Mastodon does not support client certificates. Should we nuke `SMTP_OPENSSL_VERIFY_MODE`?

akihikodaki: Didn't you notice this because you are going to use this feature? Or were you just reviewing the code?

Gargron: I also noticed that mastodon:setup-generated SMTP config failed for me until I removed `SMTP_OPENSSL_VERIFY_MODE` from it, when setting up a new instance.

ProgVal: @akihikodaki I noticed this because someone on IRC had that issue.

ineffyble:
@ProgVal Could you merge master from tootsuite into your branch, to fix the CI fails on this PR?

ProgVal:
@ineffyble Done.

lib/tasks/mastodon.rake
akihikodaki: Sorry for late answer:
> Hmm, actually, Mastodon does not support client certificates. Should we nuke `SMTP_OPENSSL_VERIFY_MODE`?

I think so.

PR #6826, opened 2018-03-19
Store objects to IndexedDB

akihikodaki:
This is an alternative of #6814.

The architecture which this code is expected to be part of has some concepts:
* A _store_ is a collection of objects.
* A _root store_ is a store of objects which don't have dependencies or are dependent on a few common dependencies. (`accounts` and `statuses`, implemented in this pull request, are _root stores_.)
* A _subordinate store_ is a store of objects which depend on objects in a _root store_.

Objects stored earlier in _root store_ will be evicted to maintain its size constant (1024 in the current implementation.) The dependent objects in other stores will be evicted as well when an object in _root store_ is evicted. Therefore the size of a _subordinate store_ is expected to depend on the size of _root stores_.

A _root store_ has auto-incremental number as primary key to note when it is fetched from the authority server. A _subordinate store_ doesn't, on the other hand.

Moreover, `importer` actions introduced in this pull request is expected to be a good place to prefetch media and cache them.

### DIFF
--- a/app/javascript/mastodon/actions/statuses.js
+++ b/app/javascript/mastodon/actions/statuses.js
@@ -1,7 +1,10 @@
 import api from '../api';
+import asyncDB from '../db/async';
+import { evictStatus } from '../db/modifier';
 
 import { deleteFromTimelines } from './timelines';
 import { fetchStatusCard } from './cards';
+import { importFetchedStatus, importFetchedStatuses, importAccount, importStatus } from './importer';
 
 export const STATUS_FETCH_REQUEST = 'STATUS_FETCH_REQUEST';
 export const STATUS_FETCH_SUCCESS = 'STATUS_FETCH_SUCCESS';
@@ -34,6 +37,48 @@ export function fetchStatusRequest(id, skipLoading) {
   };
 };
 
+function getFromDB(dispatch, getState, accountIndex, index, id) {
+  return new Promise((resolve, reject) => {
+    const request = index.get(id);
+
+    request.onerror = reject;
+
+    request.onsuccess = () => {
+      const promises = [];
+
+      if (!request.result) {
+        reject();
+        return;
+      }
+
+      dispatch(importStatus(request.result));
+
+      if (getState().getIn(['accounts', request.result.account], null) === null) {
+        promises.push(new Promise((accountResolve, accountReject) => {
+          const accountRequest = accountIndex.get(request.result.account);
+
+          accountRequest.onerror = accountReject;
+          accountRequest.onsuccess = () => {
+            if (!request.result) {
+              accountReject();
+              return;
+            }
+
+            dispatch(importAccount(accountRequest.result));
+            accountResolve();
+          };
+        }));
+      }
+
+      if (request.result.reblog && getState().getIn(['statuses', request.result.reblog], null) === null) {
+        promises.push(getFromDB(dispatch, getState, accountIndex, index, request.result.reblog));
+      }
+
+      resolve(Promise.all(promises));
+    };
+  });
+}
+
 export function fetchStatus(id) {
   return (dispatch, getState) => {
     const skipLoading = getState().getIn(['statuses', id], null) !== null;
@@ -47,18 +92,26 @@ export function fetchStatus(id) {
 
     dispatch(fetchStatusRequest(id, skipLoading));
 
-    api(getState).get(`/api/v1/statuses/${id}`).then(response => {
-      dispatch(fetchStatusSuccess(response.data, skipLoading));
-    }).catch(error => {
+    asyncDB.then(db => {
+      const transaction = db.transaction(['accounts', 'statuses'], 'read');
+      const accountIndex = transaction.objectStore('accounts').index('id');
+      const index = transaction.objectStore('statuses').index('id');
+
+      return getFromDB(dispatch, getState, accountIndex, index, id);
+    }).then(() => {
+      dispatch(fetchStatusSuccess(skipLoading));
+    }, () => api(getState).get(`/api/v1/statuses/${id}`).then(response => {
+      dispatch(importFetchedStatus(response.data));
+      dispatch(fetchStatusSuccess(skipLoading));
+    })).catch(error => {
       dispatch(fetchStatusFail(id, error, skipLoading));
     });
   };
 };
 
-export function fetchStatusSuccess(status, skipLoading) {
+export function fetchStatusSuccess(skipLoading) {
   return {
     type: STATUS_FETCH_SUCCESS,
-    status,
     skipLoading,
   };
 };
@@ -78,6 +131,7 @@ export function deleteStatus(id) {
     dispatch(deleteStatusRequest(id));
 
     api(getState).delete(`/api/v1/statuses/${id}`).then(() => {
+      evictStatus(id);
       dispatch(deleteStatusSuccess(id));
       dispatch(deleteFromTimelines(id));
     }).catch(error => {
@@ -113,6 +167,7 @@ export function fetchContext(id) {
     dispatch(fetchContextRequest(id));
 
     api(getState).get(`/api/v1/statuses/${id}/context`).then(response => {
+      dispatch(importFetchedStatuses(response.data.ancestors.concat(response.data.descendants)));
       dispatch(fetchContextSuccess(id, response.data.ancestors, response.data.descendants));
 
     }).catch(error => {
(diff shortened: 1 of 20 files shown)


akihikodaki:
Completed implementation for accounts and statuses. I'll wait till this gets merged so that the change won't be too big.

Gargron:
I have tested the PR and everything seems to still work. For some reason Firefox developer tools do not show me the contents of the accounts and statuses stores. I think that might be a Firefox issue... I do not have experience with IndexedDB so it's hard for me to say whether this is good. I like the idea of cache eviction, #787 is a very old issue.

It would be great if @nolanlawson or @sorin-davidoi could take a look.

akihikodaki:
I have an idea for cache eviction which should be combined with this change: to use `WeakMap`.

By replacing `ImmutableMap` used around reducers with `WeakMap`, an object will be evicted from Redux soon if it is no longer used. There is no need for reference counter or similar techniques because it uses the native garbage collector of JavaScript. It may be concerning that `WeakMap` will evict objects so soon, but IndexedDB is fast enough to recover the evicted objects.

However I know some cons of this solution as well. While `WeakMap` evicts each objects, many present actions fetches multiple objects at once. It means:
* the dispatchers of such actions should maintain reference for those objects so that they won't be evicted or,
* smaller actions to fetch individual objects from IndexedDB should be added.

We have to add tons of action dispatches whichever option you choose, even if it should be easier than implementing reference counters.

app/javascript/mastodon/actions/statuses.js
```diff
@@ -47,18 +87,26 @@ export function fetchStatus(id) {
...
 
-    api(getState).get(`/api/v1/statuses/${id}`).then(response => {
-      dispatch(fetchStatusSuccess(response.data, skipLoading));
-    }).catch(error => {
+    asyncDB.then(db => {
+      const transaction = db.transaction(['accounts', 'statuses'], 'read');
+      const accountIndex = transaction.objectStore('accounts').index('id');
+      const index = transaction.objectStore('statuses').index('id');
+
+      return getFromDB(dispatch, getState, accountIndex, index);
```
unarist: > 48: function getFromDB(dispatch, getState, accountIndex, index, id)
> 95: return getFromDB(dispatch, getState, accountIndex, index);

5th parameter is missing?

akihikodaki:
8051633c79609dfb7ca1ebebfd77c986d825f85e addresses the issue on Firefox and https://github.com/tootsuite/mastodon/pull/6826#pullrequestreview-105390950. The cause of the problem on Firefox was dead transaction. The lifetime of transaction is arbitrary as long as it matches the minimal requirement, so be careful when digging around this code.

app/javascript/mastodon/db/modifier.js
```diff
@@ -0,0 +1,86 @@
+import asyncDB from './async';
+
+const limit = 1024;
+
+function put(name, objects, callback) {
```
nolanlawson: Typically it's safer and less error-prone to call the callback from the transaction's`oncomplete` callback. It's guaranteed to be called when the transaction completes, so you don't have to do any extra bookkeeping yourself.

app/javascript/mastodon/db/modifier.js
```diff
@@ -0,0 +1,86 @@
...
+      function add() {
+        store.add(object).onsuccess = done;
+      }
+
+      index.getKey(object.id).onsuccess = retrieval => {
+        if (retrieval.target.result) {
+          store.delete(retrieval.target.result).onsuccess = add;
+        } else {
+          add();
+        }
```
nolanlawson: Instead of doing `getKey` to check for existence, then `delete()` if it exists, then call `add()` to re-insert it, you can just use the `put()` API which does the same thing. `put()` will overwrite any existing values. It also doesn't throw if the value already exists.

app/javascript/mastodon/db/modifier.js
```diff
@@ -0,0 +1,86 @@
...
+
+    function done() {
+      left--;
+
+      if (left <= 0) {
+        store.count().onsuccess = count => {
+          const excess = count.target.result - limit;
+
+          if (excess > 0) {
+            store.getAll(null, excess).onsuccess =
```
nolanlawson: Edge doesn't support `getAll`/`getAllKeys` yet. (Sorry 😅) There is a polyfill you can use: https://github.com/dumbmatter/IndexedDB-getAll-shim

nolanlawson: Also in this case it appears you can use `getAllKeys` rather than `getAll` because you are just returning the `id`s, right?

nolanlawson (commented):
I haven't had a chance to thoroughly review this yet, but I have some general questions:

1. What is the model for refreshing stale data? I.e. if a status is old, how does it get updated in the DB?

2. Is data ever cleared from the database, or does it just grow indefinitely?

3. I think a good performance boost may be found by adding an LRU cache on top of IDB. Most of the time, the user is fetching statuses that were just inserted into the database. Thus we can use something like https://github.com/sindresorhus/quick-lru to keep the top ~100 statuses cached in memory and avoid a round-trip to the database. Just a suggestion.

akihikodaki:
f455ec4a3e5c63f6adc3e27cbd35baed17516cae is for https://github.com/tootsuite/mastodon/pull/6826#discussion_r175821971.

app/javascript/mastodon/db/modifier.js
akihikodaki: Actually this `put` method was named from IndexedDB's `put`. However, it is different from `put` in terms that it always creates a new record. Thanks to that, a newly fetched record always has a new key. It allows to find out-dated records by sorting the key, corresponding to the fetched time.

app/javascript/mastodon/db/modifier.js
akihikodaki: The key is not `id`.

sorin-davidoi:
Don't really have any feedback on first sight but I can say that I am very excited to see this being picked up again! 😍

What is the plan of handling this in the UI? Will there be any indication that the statuses are served from cache (and thus may be out-of-date)?

akihikodaki:
3faeb01b6da2fad92b2bb6be0faba21f03806c96 disables IndexedDB when `getAll` is not available. I decided to disable it instead of adding polyfills or workaround because:
* It is rarely tested on Microsoft Edge, which is known to have the issue, and probably it will take long to notice that the browser has more problems if any. Just disabling the feature is the safest option.
* Microsoft Edge does not support Service Worker, and the most important goal of this change cannot be achieved: to let it work offline.

@nolanlawson 
> I have some general questions:

I have not answered for them so here are:

> 1. What is the model for refreshing stale data? I.e. if a status is old, how does it get updated in the DB?

> 2. Is data ever cleared from the database, or does it just grow indefinitely?

Stale objects in _root store_ (see the top comment) will have earlier keys. The importer actions will evict objects with such keys when it found there are too many records. Dependent objects in _subordinate store_ will be evicted at the same time.

> 3. I think a good performance boost may be found by adding an LRU cache on top of IDB. Most of the time, the user is fetching statuses that were just inserted into the database. Thus we can use something like https://github.com/sindresorhus/quick-lru to keep the top ~100 statuses cached in memory and avoid a round-trip to the database. Just a suggestion.

We already have cache: Redux store! But it is not LRU in sense that it does not garbage collect LRU objects.
I have suggested garbage collection with `WeakMap`. See https://github.com/tootsuite/mastodon/pull/6826#issuecomment-374631176. The major difference from LRU and `WeakMap` is that LRU can maintain objects for a bit after they lost all references if they are expected to be referenced shortly after. `WeakMap` will evict objects almost immediately after they lost references.
So which should we implement? I'm for `WeakMap` because it is easy to implement and I don' think the performance improvement of LRU is not that great as IndexedDB is fast enough.

akihikodaki:
@sorin-davidoi
> What is the plan of handling this in the UI? Will there be any indication that the statuses are served from cache (and thus may be out-of-date)?

I have no plan. The fetching actions actively skips the fetching operation if the object to be fetched is already in Redux store even without this change. If we want precise representation of volatile objects, such actions should be modified whether this pull request is merged or not.

Gargron (approved):

nolanlawson:
@akihikodaki Microsoft Edge will support Service Worker in the next release (should be coming out any day now). I understand not wanting to introduce uncertainty, but the `getAll`/`getAllKeys` polyfill is tiny. When I get a chance, I'll see if I can open a PR to dynamically apply the polyfill.

akihikodaki:
@nolanlawson That is nice. I hope we will get Mastodon work offline on latest Edge.

PR #6834, opened 2018-03-19
change defaults for redis, postgres

devkral:
* if REDIS_HOST is set to default, logins doesn't work. It isn't obvious and costed me 3 hours to figure out.
* Postgresql throws an error but it would be nicer if it just works.
* rename default database to mastodon as clashes are more unlikely (had that issue)


nightpool:
the defaults are appropriate for the Docker environment, which is what we
recommend most admins use

Gargron:
You can use rake mastodon:setup to guide you through this. The defaults in the sample have always been appropriate for Docker.

devkral:
ok thank you. Didn't know. A warning in the documentation would be nice.

Issue #6737, opened 2018-03-11
Replying to yourself and multiple users reverses the order of mentions.


trwnh:
When replying to multiple users, each new reply reverses the order of the mentions included. 
![image](https://user-images.githubusercontent.com/10606431/37250951-13554346-24cd-11e8-876f-219816573fd0.png)

In this example, I am replying to a toot that initially mentioned `@itsaffenmann` and `@umby`.
However, the reply to it loads mentions in reverse-order -- the last user mentioned in the previous toot becomes the primary recipient of the new toot. This is bad semantically because when replying to yourself, you are effectively continuing the chain -- and your message should still be targeted at the last person who spoke before you.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

trwnh:
Closing because #6836 was merged

PR #6836, opened 2018-03-19
Fixes/mentions reply order

ClearlyClaire:
When replying to a toot from the WebUI, the compose box is pre-filled with mentions pulled from the original toot. Unfortunately, they are sometimes reversed (see #6737).

This is because the WebUI relies on the order of the `mentions` fields of the status. This order is not enforced on the server side and will be in fact reversed.

This does not occur when replying to a toot from another Mastodon instance, since the mention list will be serialized twice (ActivityPub/OStatus, then the REST API), reversed each time.

This Pull Request fixes this by enforcing the serialized order to match the order the `Mention` records were inserted into the database.

Unfortunately, this means the behavior will be broken for toots from other instances that haven't been updated to this patch.


ThibG:
Woops

ThibG:
I guess the WebUI could parse the HTML-formatted message to add mentions in the order of the matching links… but I don't really want to go that way…

Gargron (approved):

PR #6838, opened 2018-03-19
Cancel outdated pending compose suggestions

ClearlyClaire:
This PR cancels outdated compose suggestions requests to prevent outdated results to show up instead of newer (usually more precise) ones, or when the current composed text does not call for a completion.

### DIFF
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -1,4 +1,5 @@
 import api from '../api';
+import { CancelToken } from 'axios';
 import { throttle } from 'lodash';
 import { search as emojiSearch } from '../features/emoji/emoji_mart_search_light';
 import { tagHistory } from '../settings';
@@ -10,6 +11,8 @@ import {
   refreshCommunityTimeline,
   refreshPublicTimeline,
 } from './timelines';
+
+let cancelFetchComposeSuggestionsAccounts;
 
 export const COMPOSE_CHANGE          = 'COMPOSE_CHANGE';
 export const COMPOSE_SUBMIT_REQUEST  = 'COMPOSE_SUBMIT_REQUEST';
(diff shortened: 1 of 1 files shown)


LienRag:
I don't get it, how do you decide what is obsolete or not?

ThibG:
@LienRag the most recently-issued request is the only one which results we are interested in. Any other request that might still be running can be canceled.

ThibG:
Hm, some unexpected behavior is still going on, and I'm not completely sure why: when typing something that could be a username (e.g. `@gaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`) then pressing backspace so it is quickly deleted, I seem to still get an answer for `@ga`. I guess the cancellation function might be created asynchronously and `clearComposeSuggestions` might be called too early?
Anyway, that's an edge case, the behavior is still way better for me.

app/javascript/mastodon/actions/compose.js
```diff
@@ -11,6 +12,9 @@ import {
   refreshPublicTimeline,
 } from './timelines';
 
+var CancelToken = axios.CancelToken;
```
Gargron: Use `let`

app/javascript/mastodon/actions/compose.js
```diff
@@ -257,13 +261,22 @@ export function undoUploadCompose(media_id) {
...
     type: COMPOSE_SUGGESTIONS_CLEAR,
   };
 };
 
 const fetchComposeSuggestionsAccounts = throttle((dispatch, getState, token) => {
+  if (_CancelFetchComposeSuggestionsAccounts) {
+    _CancelFetchComposeSuggestionsAccounts();
+  }
   api(getState).get('/api/v1/accounts/search', {
+    cancelToken: new CancelToken(function (cancel) {
```
Gargron: Use `cancel => {` for brevity

Gargron (changes requested):
This change is very welcome, but new js syntax is better

app/javascript/mastodon/actions/compose.js
```diff
@@ -11,6 +12,9 @@ import {
   refreshPublicTimeline,
 } from './timelines';
 
+let CancelToken = axios.CancelToken;
+let _CancelFetchComposeSuggestionsAccounts;
```
akihikodaki: What is this underscore?

ThibG: Just a sign that this is to remain private to this module. But this convention doesn't appear to be used widely/consistently within Mastodon, so I guess I can drop it (and also start with a lowercase).

akihikodaki: I'm for dropping this because the code knows it, even without such a symbol. The code will not allow to use the variable out of this module.
If i recall correctly, underscore as prefix is seen in some components of Mastodon, but that could make sense because the code doesn't know a variable is private, and allows to use it out of the component (if `ref` is there.)

Gargron (approved):

app/javascript/mastodon/actions/compose.js
```diff
@@ -1,4 +1,5 @@
 import api from '../api';
+import axios from 'axios';
```
Gargron: Oh I get it now

`import axios, { CancelToken } from 'axios';`

should work instead of `let CancelToken`

PR #6841, opened 2018-03-20
Fix OpenGraph meta tag's translation error

mayaeh:
After #6818 , "translation missing" is displayed in og: description of OGP meta tag on toot direct link  in untranslated language.
This PR will fix it.

before:

after:

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -40,16 +40,16 @@ module StreamEntriesHelper
       end
     end
 
-    text = attachments.to_a.reject { |_, value| value.zero? }.map { |key, value| t("statuses.attached.#{key}", count: value) }.join(' · ')
+    text = attachments.to_a.reject { |_, value| value.zero? }.map { |key, value| t("statuses.attached.#{key}", count: value, default: t("statuses.attached.#{key}", locale: :en)) }.join(' · ')
 
     return if text.blank?
 
-    t('statuses.attached.description', attached: text)
+    t('statuses.attached.description', attached: text, default: t('statuses.attached.description', locale: :en))
   end
 
   def status_text_summary(status)
     return if status.spoiler_text.blank?
-    t('statuses.content_warning', warning: status.spoiler_text)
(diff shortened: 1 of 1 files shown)


mayaeh:
Please wait for the merge yet.
I am looking at problems pointed out at Mastodon.

Gargron:
Odd. It should definitely just fallback to the `en` locale in production?

tateisu:
if fall back to non-en language, it may reproduct missing translation. I think fallback to en is mandatory.

mayaeh:
Sorry, I chose this approach because I could not find any other way.
I thought en would be better than choosing a language that might not have a translation.

Gargron:
This code:

https://github.com/tootsuite/mastodon/blob/master/config/application.rb#L78-L83

Should've made sure than when the default locale is not English, missing translations still fallback to English. Why is it not working then?

unarist:
https://github.com/tootsuite/mastodon/blob/a5c6c748e096f61d00bbd778a263e22117e1ae9f/config/application.rb#L79-L81

https://github.com/tootsuite/mastodon/blob/a5c6c748e096f61d00bbd778a263e22117e1ae9f/config/environments/production.rb#L57-L59

Well...it is overwritten.

MitarashiDango:
In my environment this phenomenon was reproduced.
Also, it was confirmed that `DEFAULT_LOCALE` of `.env.production` is affecting.
Is this the intended behavior?

Gargron:
>Well...it is overwritten.

@akihikodaki aaaaaaaaaaaaaaa

akihikodaki:
I'm really sorry. Recently I had two bugs, including this, which cannot be reproduced on my development environment, but I think I could find out them before it gets merged if I was careful enough.

I'm preparing a fix and submit shortly.

ineffyble:
@mayaeh This issue was addressed by #6843 - are you happy to close?

mayaeh:
Thank you for being taught.
I am also checking merging # 6843 in my environment.
Please let me close this after that running.

Issue #6837, opened 2018-03-19
Usernames are case-sensitive on local database, but not on remote domain (even with the same id)


trwnh:
Earlier today, Krita made an account on mastodon.art, and I followed `@krita@mastodon.art` from `mastodon.social`. The account followed me back. But later in my home column, I see a boost from `@Krita@mastodon.art`, and attempting to load it in the webapp seems to be treating it as a different account that is neither following me nor am I following it. However, loading the profile URLs `mastodon.art/@krita` or `mastodon.art/@Krita` both resolve to the same account. 

krita in webapp: 
![chrome_2018-03-19_14-38-25](https://user-images.githubusercontent.com/10606431/37618339-f87b1678-2b83-11e8-89da-d59bad4ab96f.png)

Krita in webapp
![chrome_2018-03-19_14-38-40](https://user-images.githubusercontent.com/10606431/37618367-0b49c448-2b84-11e8-9439-496a79892712.png)

krita in profile
![chrome_2018-03-19_14-37-49](https://user-images.githubusercontent.com/10606431/37618408-28c83fea-2b84-11e8-9526-27680c2dc7b7.png)

Krita in profile
![chrome_2018-03-19_14-37-55](https://user-images.githubusercontent.com/10606431/37618418-2f00864c-2b84-11e8-96b3-1008ae47e4bc.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

trwnh:
They also appear as two different accounts in search. 
![image](https://user-images.githubusercontent.com/10606431/37618594-b64c2fe8-2b84-11e8-8ecd-6c021d45cc79.png)

Gargron:
@angristan reported the same from mstdn.io. This should not be possible at all, the db constraint in the username column is case insensitive, so duplicates should not be allowed at all 

I wonder if these accounts have different activitypub URIs. I don't think those are case insensitive. But they are also canonical and never expected to change. Was the username of the krita account changed after creation, resulting in a different URI?

trwnh:
I'm not sure that they *should* have different URIs. I don't think it was changed -- what I'm most sure of is that the following occured:

1) Krita makes an account at `mastodon.art/@Krita`. They send this toot which shows up as such (lowercase): https://mastodon.art/@krita/99710831343843952
[...]

Issue #6667, opened 2018-03-07
Issues after changing WEB_DOMAIN to be the same as LOCAL_DOMAIN


aminb:
So I started out on a setup with different `LOCAL_DOMAIN` and `WEB_DOMAIN` in order to run mastodon on a subdomain while using top-level user ids.

I decided to try changing my setup by commenting out `WEB_DOMAIN` and effectively moving mastodon to top-level. However, after the change, my instance seems to have trouble communicating with other mastodon instances (GNU social instances seem to be just fine with the change).

Here are some of the errors I see in the Sidekiq panel:

**First**, this is the error I got when trying to follow `@a@functional.cafe` from `@b@nda.li`:

- Job: `ActivityPub::DeliveryWorker`
- Arguments:
```
[...]

PR #6842, opened 2018-03-20
Use username/domain to match existing accounts in ActivityPub

Gargron:
See also: #6837, #6667 

This makes ActivityPub processing behaviour consistent with ResolveAccountService, and username/domain ultimately authoritative over URIs. That's not *purist*, I admit, but at the same time: If the username/domain is how people address each other (and it is), why should something else differentiate accounts? So I think it's fine.

This PR does not resolve the potential presence of duplicates.


ykzts (approved):

cwebber:
Hm.  This could result in problems with interoperating with activitypub servers that aren't using webfinger.  It's also not to spec generally.  What happens when you interact with something that doesn't have a preferredUsername field?

PR #6843, opened 2018-03-20
Fix i18n fallback configuration conflicts with environment configurations

akihikodaki:
This fixes a bug reported at https://github.com/tootsuite/mastodon/pull/6841#issuecomment-374607268.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -55,8 +55,8 @@ Rails.application.configure do
   # config.action_mailer.raise_delivery_errors = false
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
-  # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
+  # English when a translation cannot be found).
+  config.i18n.fallbacks = [:en]
 
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -56,7 +56,7 @@
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
```
ineffyble: Is this comment still accurate? Seems like it should say something more like "make lookups for any locale fall back to 'en'"?

akihikodaki: No. Addressed in https://github.com/tootsuite/mastodon/pull/6843/commits/7820893aca79826b19c51da8dc834b13278db0b7. Thanks for pointing out this.

ineffyble:
To clarify how this works for my own benefit:
- Instance owner sets a default locale via environment variable
- This default locale is used for all the localised strings shown within Mastodon
- If a string is not available in this locale, Mastodon will instead use the English version of the string

In addition:
- If the default locale is not provided, or is unavailable or invalid, it is set to English

Does that all sound right?

akihikodaki:
Yes. And it has been the expected behavior, so it is a _bug_ if it doesn't work as you stated. (That's why this pull request is labelled _bug_ here.)

nightpool (approved):
i just touched this code recently for cybrespace-specific locale changes, so i can confirm this is the correct change to make. thanks aki!

nightpool:
(to be clear, the bug here is that we were setting `.fallbacks = [:en]` in application.rb, but it was getting overwritten later by the default `.fallbacks = true` in production.rb.

ineffyble (approved):

PR #6847, opened 2018-03-20
paperclip: allow S3 server-side encryption

rtucker:
S3 offers a server-side encryption feature, where they will encrypt data
at-rest and decrypt it on-the-fly before serving. This may provide an
additional layer of protection for our users' media assets.

It is disabled by default, but can be enabled by adding:

S3_SERVER_SIDE_ENCRYPTION=AES256

to .env.production.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -91,6 +91,7 @@ SMTP_FROM_ADDRESS=notifications@example.com
 # S3_REGION=
 # S3_PROTOCOL=http
 # S3_HOSTNAME=192.168.1.123:9000
+# S3_SERVER_SIDE_ENCRYPTION=AES256
 
 # S3 (Minio Config (optional) Please check Minio instance for details)
 # S3_ENABLED=true
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -29,6 +29,7 @@ if ENV['S3_ENABLED'] == 'true'
       'Cache-Control' => 'max-age=315576000',
     },
     s3_permissions: ENV.fetch('S3_PERMISSION') { 'public-read' },
+    s3_server_side_encryption: ENV.fetch('S3_SERVER_SIDE_ENCRYPTION') { false },
     s3_region: s3_region,
     s3_credentials: {
       bucket: ENV['S3_BUCKET'],


Gargron:
I don't see how it offers any meaningful protection... You are trusting S3 to do this, using their own keys, as I understand?

rtucker:
It's a little more nuanced: each object gets its own key, then those keys are encrypted with a master key, which they rotate regularly.  Or so they say.

Does this buy anything?  Probably.  Does it hurt?  Probably not.

Here's the docs:

https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

akihikodaki:
> Does this buy anything? Probably.

I doubt it. If someone want something securer, he should not upload media on the Internet.

codl:
This only gives the illusion of added security, which is worse than no added security, and I don't think it should be added

Please excuse my armchair security...ing, and please correct me if I'm wrong, but, here are the two threat models I can think of that this would allegedly protect against:

* Someone gets access to an S3 server, and reads your file.

  This **would not** prevent that attack. As far as I can tell, all the files that mastodon puts into paperclip are publically readable, and while the paths may not be public, they're also not encrypted. The attacker with access to your encrypted files also has access to your *unencrypted* file paths and can now just request the unencrypted files from S3 directly.

* Someone gets access to all the S3 servers where copies of your file are stored, and overwrites your file with a naughty file, also overwriting the checksum, wherever that is stored. 

  This **would** prevent that attack as the new file could not be decrypted by the original file's key. Is it a credible threat? idk, it's really far-fetched 🤷

akihikodaki:
> but, here are the two threat models I can think of that this would allegedly protect against:
> * Someone gets access to an S3 server, and reads your file.
>
>  This **would not** prevent that attack.

Even worse, such an attacker will focus on stealing data with encryption option because data interesting for him is likely to be encrypted. On the other hand, there is too much plain, trivial data on the server and he may not pay attention to it. So not enforcing encryption could be even secure in some case.

rtucker:
... and that's precisely why, if the option is available, I encrypt everything, without regard to whether or not it's particularly sensitive. You're going to have to sift through a *lot* of encrypted cat pictures to find anything juicy.

(There are options for using a customer-provided encryption key with S3... it would be cool to use that somehow, but I'm not sure it is possible or plausible in practice.)

nynhex:
Honestly, this could go either way.  It seems a bit overkill, but again it depends on your use case.  Also, I've done this (s3 encryption with paperclip) on another platform I wrote and while aws can decrypt in flight, there's a tad bit of overhead associated with it.  I've benchmarked it and it's like a 3-4ms lag vs serving from an unencrypted s3 bucket.  Note this was using KMS on s3, not customer provided keys, but the crypto cipher is the same so it's likely customer provided keys would have the same ripple effect.

Gargron:
I don't think this adds anything, sorry. I'm pretty sure the S3 feature is just AWS marketing. Gonna close.

PR #6858, opened 2018-03-21
Do not re-query mentions from serializers

Gargron:
Mentions were pre-fetched. There's no need to query n+1 times again just to sort them.

Fix performance regression from #6836


Gargron:
@ThibG

PR #6861, opened 2018-03-21
Encode . as - in username URLs (potential fix)

### LINKED ISSUE #6860: Usernames with dots in them break routes
See https://github.com/tootsuite/mastodon/pull/6844#issuecomment-374993289

This is because the dot is used as separator for format in Rails. Before you could easily pull up the ActivityPub JSON of a resource by adding `.json` to the URL, if we change the constraints of the routes, this will become impossible.

Gargron:
Just one possible solution... It does not look super pretty, though.

Fix #6860

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -221,7 +221,7 @@ class Account < ApplicationRecord
   end
 
   def to_param
-    username
+    username.gsub('.', '-')
   end
 
   def excluded_from_timeline_account_ids
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -13,7 +13,7 @@ module AccountFinderConcern
     end
 
     def find_local(username)
-      find_remote(username, nil)
+      find_remote(username.gsub('-', '.'), nil)
     end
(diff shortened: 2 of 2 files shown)


tigre-bleu:
We might have a problem if @john.doe and @john-doe are both existing accounts.

Gargron:
@tigre-bleu We don't allow dashes in usernames :P

tigre-bleu:
😄 Ok, well I think the issue requiring dots in usernames is also valid for dashes (As a reminder the goal is to support already existing LDAP user names).

IMHO any character valid as a the user part of an email adress shall be allowed as a mastodon username.

nickswift:
Have you considered encoding periods and hyphens and other such characters with ASCII codes, ie. `%2E` and `%2D` respectively? It doesn't make for the most attractive URLs, but if the goal is to observe LDAP usernames and the like, this seems like the most straightforward way to go about it.

tigre-bleu:
Another maybe easier to implement solution could be to authenticate using the email adress from the LDAP and to let the user choose a username with the legacy constraints.

PR #6862, opened 2018-03-22
Fix MENTION_RE to not match nil usernames

ClearlyClaire:
Fix broken regexp that matched nil mentions (just a single `@` sign)


Gargron (approved):
Oh, I misunderstood your earlier comment. Yeah, I think this was my mistake when extracting USERNAME_RE.

PR #6863, opened 2018-03-22
Restore username validation to disallow dots, for now

Gargron:
Usernames with dots in them do not work with routes, because the dot usually separates the desired page format (e.g. json). I don't want to mess with changing route constraints for this patch release.


tagnotfound:
Hey,

I am facing a similar issue with the username. Since my CAS Server is returning the usernames with a dot. 

Can we follow the approach of replacing dots with an underscore at https://github.com/tootsuite/mastodon/blob/ad94e1d50a1a13951d82aefcba97f581c54ca70a/app/models/concerns/omniauthable.rb#L67

PR #6872, opened 2018-03-22
Invites: Add '1 week' as expire option

ghost:
IMO a max. of 1 day is too short. Not everyone has the time and motivation to use an invite in a 24 hour period. 1 week as a max. is I think a good compromise between convenience and security.

Live testing it on Todon.nl. No issues so far.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -475,6 +475,7 @@ en:
       '3600': 1 hour
       '43200': 12 hours
       '86400': 1 day
+      '604800': 1 week
     expires_in_prompt: Never
     generate: Generate
     max_uses:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -474,8 +474,8 @@ en:
       '21600': 6 hours
       '3600': 1 hour
       '43200': 12 hours
-      '86400': 1 day
       '604800': 1 week
+      '86400': 1 day
     expires_in_prompt: Never
     generate: Generate
     max_uses:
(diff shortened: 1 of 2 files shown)


jeroenpraat:
I can't find why this give errors. Hopefully someone knows what's wrong.

To be sure I disable live testing on my instance.

config/locales/en.yml
```diff
@@ -475,6 +475,7 @@ en:
       '3600': 1 hour
       '43200': 12 hours
       '86400': 1 day
+      '604800': 1 week
```
ykzts: sort.write above line 477.

jeroenpraat: Thanks

nightpool (approved):

nynhex:
This was a really good call.  1 day is too short 1 week seems just right 🌮

akihikodaki (approved):

trwnh:
Optimally, shouldn't this allow for custom expiry? For example, 2 or 3 days? Or 2 weeks? Or 42 hours?

nightpool:
@trwnh no, that's too much complexity for no real gain

trwnh:
@nightpool I'd argue that {1 hour, 12 hours, 1 day, 1 week} is a really broad span of time -- if I ever had to give out an invite, for example, 1 day is too short and 1 week is too long, or 1 week is too short (but is the next level 1 month?), and so on. And it seems more complex to add every single arbitrary point along the way when it could simply be passed directly to the time-out function.

jeroenpraat:
Thanks!

PR #6875, opened 2018-03-23
Change “Toots with replies” to “Toots and replies”

### LINKED ISSUE #6870: Cosmetic: "Toots with replies" is not toots with replies
in english locales, the phrasing of the "Toots with replies" tab is misleading. on first encountering that phrase, i interpreted it as meaning "only those toots which have received replies" rather than "both this person's toots and this person's replies to other toots". suggested alternative would be "Toots and replies" (or maybe "Toots & replies").

toots with replies

at a cursory glance, this appears to be a potential issue in other languages as well ("Pouets avec réponses" vs. "Pouets et réponses"), though not universally ("投稿と返信" makes sense)

* * * *

mkljczk:
fixes #6870

### DIFF
--- a/app/javascript/mastodon/features/account_timeline/components/header.js
+++ b/app/javascript/mastodon/features/account_timeline/components/header.js
@@ -99,7 +99,7 @@ export default class Header extends ImmutablePureComponent {
         {!hideTabs && (
           <div className='account__section-headline'>
             <NavLink exact to={`/accounts/${account.get('id')}`}><FormattedMessage id='account.posts' defaultMessage='Toots' /></NavLink>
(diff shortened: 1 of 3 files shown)


akihikodaki (approved):

trwnh:
Would it break anything to change the id name as well?

trwnh:
@m4skin so why not do it? ;)

nightpool:
@trwnh because it matches the path "/with_replies", which is *would* break things to change

trwnh:
@nightpool That's fair, I'm just nitpicking for consistency

shmibs:
cool ^_^

maybe could also be a good idea to have this as an excuse, native speakers checking over the other (probably just western) languages this may have percolated into via translation?

PR #6876, opened 2018-03-23
Allow clients to fetch statuses made while they were offline

akihikodaki:
This is also a part of changes for offline environment.
Chrome on Android keeps Web pages for a long duration, and allows to open them immediately. (I guess other mobile platforms do similar.) However, Mastodon client does not fetch statuses made while it was offline and a user will be forced to reload the page if he wants to see them. This is to solve the problem.

This change also unified timeline refreshing actions and expanding actions and reduced redundancy. It is meaningful as refactoring, too.

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -1,10 +1,30 @@
+import { debounce } from 'lodash';
 import React from 'react';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import PropTypes from 'prop-types';
 import StatusContainer from '../containers/status_container';
 import ImmutablePureComponent from 'react-immutable-pure-component';
+import LoadMore from './load_more';
 import ScrollableList from './scrollable_list';
 import { FormattedMessage } from 'react-intl';
+
+class LoadGap extends ImmutablePureComponent {
+
+  static propTypes = {
+    disabled: PropTypes.bool,
+    maxId: PropTypes.string,
+    onClick: PropTypes.func.isRequired,
+  };
+
+  handleClick = () => {
+    this.props.onClick(this.props.maxId);
+  }
+
+  render () {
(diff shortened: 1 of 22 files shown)


app/javascript/mastodon/components/status_list.js
```diff
@@ -1,11 +1,31 @@
...
 import ImmutablePureComponent from 'react-immutable-pure-component';
+import LoadMore from './load_more';
 import ScrollableList from './scrollable_list';
 import { FormattedMessage } from 'react-intl';
 
+class LoadGap extends ImmutablePureComponent {
+
+  static propTypes = {
+    disabled: PropTypes.bool,
+    maxId: PropTypes.string,
```
Gargron: I think it's possible to use min_id now too?

akihikodaki: It decides that a gap is filled by looking for statuses duplicate in the fetched list and the local list. `min_id` is omitted to bring such duplicate statuses.

Gargron: Ah, I think we didn't expose it in the API and there's no equivalent yet. Still, I wonder if min_id is necessary for the gap to function properly?

Gargron (approved):

jebuskrust:
Greetings, 

I believe this introduces a bug where after 20 items in the home timeline feed causes an unexpected error to occur.

akihikodaki:
It is not reproducible for me, but I think there is such a bug, and it requires a more precise procedure to reproduce. It would be nice if you tell more details.

I tried to reproduce the bug on:
```
Mozilla/5.0 (X11; Linux x86_64; rv:59.0) Gecko/20100101 Firefox/59.0
```

I tried three different procedures but none of them reproduced the bug:
* Open home timeline, scroll down the timeline to load more than 20 items
* Open home timeline, and wait for new items more than 20 to be streamed
* Open home timeline, kill the streaming server, and wait for new items more than 20 to be polled

PR #6902, opened 2018-03-25
Cache media

akihikodaki:
Now accounts avatar is visible even on offline environment.


Gargron:
Should this be the task of a database, or should this be the task of the service worker? I believe file cache is the service worker's issue

akihikodaki:
> Should this be the task of a database, or should this be the task of the service worker? I believe file cache is the service worker's issue

None of them. It should be a task of normal JavaScript. Cache is exposed to window context for this kind of usage.

Gargron (approved):
Okay, I see. Pretty cool :+1:

j-a4:
I'm running into an issue in Chrome 65.0.3325.181 where I'm presented with an error messaging indicating that the website is down, but it's actually due to caching system exceeding the quota.

931e399f1e76f832
unknown

Force refreshing allows the page to load, but any normal navigation to the instance fails with that message. I was able to fix it by clearing the cache.

saper (commented):
Do we have to use `process.env` like this?

akihikodaki:
I also concerned about information leakage, but I just ignored the concern here because we already do same for the normal entrypoints with `EnvironmentPlugin`. If you are really worrying about that, it should be also modified as well.

saper:
Yes, the other use is also not good. What are we using here really - `process.env.CDN_HOST`, `process.env.NODE_ENV`, anything else?

akihikodaki:
That is all as far as I know.

Gargron:
@akihikodaki Could you please investigate the above "quota exceeded" crash? If it's unfixable, we would have to revert this functionality.

akihikodaki:
@Gargron The problem is fixable. I'm preparing a fix.

PR #6904, opened 2018-03-25
feat(ShowMore): Add classname to show more/show less button

stephenburgess8:
Hi - Based on this toot - https://mastodon.observer/web/statuses/99742882182857387

Some people may want to apply custom styles to the show more / show less buttons. However in the current scheme there's no simple way to differentiate between the two. This change is to add a classname to indicate which button is active to allow custom styling or future styling. Thanks.


ykzts (approved):

stephenburgess8:
Thanks!

PR #6915, opened 2018-03-26
Fix moved account handling in IndexedDB feature (regression from #6826)

unarist:
* **Fix stack overflow on importFetchedAccounts**

  image

  When the account has moved property, it should process destination
  account instead of source account itself.

* **Set account id instead of account object for moved property**

  This restores "foo has moved to" indication on account view, and
  fixes `reblog` index on `accounts` object store.

### DIFF
--- a/app/javascript/mastodon/actions/importer/index.js
+++ b/app/javascript/mastodon/actions/importer/index.js
@@ -39,7 +39,7 @@ export function importFetchedAccounts(accounts) {
     pushUnique(normalAccounts, normalizeAccount(account));
 
     if (account.moved) {
-      processAccount(account);
+      processAccount(account.moved);
     }
   }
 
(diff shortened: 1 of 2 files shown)


akihikodaki (approved):
Thanks. I hope it fixes timelines.

unarist:
I've added one more fix and now moved account handling will works fine.

btw, should we evict all caches related to the source account when the destination account has been evicted? At least, statuses from the source account won't be related to the destination account.

https://github.com/tootsuite/mastodon/blob/85a395fab6d7077a252bfe6f96673931ea3aa5ee/app/javascript/mastodon/db/modifier.js#L50-L61

akihikodaki:
> btw, should we evict all caches related to the source account when the destination account has been evicted? At least, statuses from the source account won't be related to the destination account.

`fetchAccount` expects that, so yes.

PR #6918, opened 2018-03-26
Update Polish string

maciej-trebacz:
The title 'Polszczyzna' for Polish locale was incorrect, it should read 'Polski'. This pull request fixes that.

### DIFF
--- a/app/helpers/settings_helper.rb
+++ b/app/helpers/settings_helper.rb
@@ -25,7 +25,7 @@ module SettingsHelper
     nl: 'Nederlands',
     no: 'Norsk',
     oc: 'Occitan',
-    pl: 'Polszczyzna',
+    pl: 'Polski',
     pt: 'Português',
     'pt-BR': 'Português do Brasil',
     ru: 'Русский',


ykzts:
cc @KrzysiekJ @m4sk1n @rysiekpl

m4sk1n:
#5488
https://capitalism.party/users/KrzysiekJ/updates/231

M4v3R:
@m4sk1n oops, missed that. Still, I think 'Polszczyzna' sounds... artificial. No software I know use that term, nor do people use it casually. I would suggest to reconsider that choice :).

M4v3R:
@m4sk1n One more thing I'd like to add about why I'm not a fan of 'Polszczyzna' - when I saw that I got an impression that it was a product of machine translation which often produces this kind of artificial sounding translations. And that put me off a bit, thinking that no big effort was put in to Polish translation. I see now that I was wrong, but that was the first impression I had so wanted to share that.

m4sk1n:
No software I know (besides Mastodon) uses “toot” term for posting… ;)

M4v3R:
True, and the same applies to Twitter which invented the term "tweet", and yet uses standard wording for everything else :)

zrzut ekranu 2018-03-26 o 12 38 00

rysiekpl (changes requested):
`Polszczyzna` is nicer IMVHO. Not many other services use it, but it's perfectly valid, and is a nice traditional way of referring to the Polish language.

nightpool:
If no other services use it—and i could not find a single one that does, among the many i use—is it going to put off new users from joining mastodon, if they get the same impression @M4v3R did?

> when I saw that I got an impression that it was a product of machine translation which often produces this kind of artificial sounding translations

This is very concerning to me.

KrzysiekJ:
Language names are listed, which are arguably nouns, and [„polszczyzna” is adequate for that purpose](http://sjp.pwn.pl/doroszewski/polszczyzna.html), according to the dictionary of Polish edited by Doroszewski. On the other hand, [the same dictionary doesn’t mention usage of „polski” standalone as a noun describing language](https://sjp.pwn.pl/doroszewski/polski;5476592.html) (there is obsolete „polskie” however).

PR #6922, opened 2018-03-26
Increased column size on 631px media queries

PudgyPoppins:
Without 25% width (currently 300px)
without 25 width currently 300px

With 25% width
with 25 width


Gargron:
There's no max width at all, is there? What if you have a very wide screen and only a few columns?

trwnh:
Not sure I personally like the wide look, either. But if one should be implemented, then the max width should be 45-75ch.

PR #6939, opened 2018-03-28
Fix cross origin error when caching response from CDN

akihikodaki:


app/javascript/mastodon/service_worker/entry.js
```diff
@@ -54,8 +54,9 @@ self.addEventListener('fetch', function(event) {
 
       if (cached === undefined) {
         const fetched = await fetch(event.request);
+        const control = fetched.headers.get('Cache-Control');
```
unarist: We can't read any headers for no-cors (opaque filtered) response.

https://fetch.spec.whatwg.org/#concept-filtered-response-opaque

> An opaque filtered response is a filtered response whose type is "opaque", url list is the empty list, status is 0, status message is the empty byte sequence, header list is empty, body is null, and trailer is empty.

unarist (changes requested):

akihikodaki:
I had a wrong assumption that `Cache-Control` is not filtered, but it is actually is. So: #6941.

PR #6941, opened 2018-03-28
Document CORS requirement for asset host

akihikodaki:
Required by #6902.


unarist (approved):
Background: We want to cache only successful responses, but we cannot get response code nor any headers from cross-origin resources without CORS headers. Otherwise, we need to choice "cache CDN resources even if the request failed" or "don't cache CDN resources".

Gargron (approved):

PR #6953, opened 2018-03-29
Fix ReferenceError when Cache API is missing (regression from #6902)

unarist:
```
Reference Error: Can't find variable: caches
```

Cache API is not supported on Safari / WebKit yet.

Since those caching is optional, this patch simply ignores it.

### DIFF
--- a/app/javascript/mastodon/storage/modifier.js
+++ b/app/javascript/mastodon/storage/modifier.js
@@ -4,7 +4,10 @@ import { autoPlayGif } from '../initial_state';
 const accountAssetKeys = ['avatar', 'avatar_static', 'header', 'header_static'];
 const avatarKey = autoPlayGif ? 'avatar' : 'avatar_static';
 const limit = 1024;
-const asyncCache = caches.open('mastodon-system');
+
+// ServiceWorker and Cache API is not available on iOS 11
+// https://webkit.org/status/#specification-service-workers
+const asyncCache = window.caches ? caches.open('mastodon-system') : Promise.reject();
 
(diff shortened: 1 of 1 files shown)


akihikodaki (changes requested):
* Particularly what version(s)?
* It would be nice if it has a comment.

akihikodaki:
It is included in Service Worker specification, so it would be enough to refer this page:
https://webkit.org/status/#specification-service-workers

unarist:
I've confirmed the issue on iOS11.2.6 and [latest automated build for Windows](https://build.webkit.org/builders/Apple%20Win%20Release%20%28Build%29/builds/8490), but hmm, "Supported In Preview"? Safari TP may have this API, but now I don't have a macOS to test that.

I will add those comment anyway.

Gargron (approved):

akihikodaki:
> I've confirmed the issue on iOS11.2.6 and latest automated build for Windows, but hmm, "Supported In Preview"? Safari TP may have this API, but now I don't have a macOS to test that.

It is important to note that it was still preview when you wrote the code.

unarist:
> It is important to note that it was still preview when you wrote the code.

Hmm, still I'm not sure your point. I've noted about why do we need this fix for now, but should I also mention preview builds?

akihikodaki:
> Hmm, still I'm not sure your point. I've noted about why do we need this fix for now, but should I also mention preview builds?

Sorry. No, I just missed your last line.

akihikodaki (approved):

Issue #3326, opened 2017-05-26
Add option to profile dropdown to directly send direct message


ghost:
When I have someone's profile open, the easiest way to send a direct message to that person seems to be to select "mention" from the hamburger menu below their profile and then to set the privacy to "direct message".

This seems to be a bit unintuitive; it took me a while to figure it out. As a new user I would expect to directly find a "direct message" link on a profile (assuming that we follow each other etc.).
Additionally, it is very easy to first write the message and then forget to set the privacy option properly.

Therefore, I would find it preferable if one could directly select "write direct message" and would then be dropped in the compose field with the right privacy option already selected.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

cphuntington97:
It would be very easy to add "Direct" underneath "Mention" on the profile menu.

Cassolotl:
We have this now! :)

![screen shot 2018-05-26 at 12 32 51](https://user-images.githubusercontent.com/6607512/40575668-f2657962-60e0-11e8-8440-c6fe674085e5.png)

PR #6956, opened 2018-03-29
Feature: Direct message from menu

ThisIsMissEm:
Two main commits here, the first implements the direct message from user menu, the second adds a warning about the audience of the direct message, as you wouldn't see this warning otherwise (as you're not manually selecting a privacy setting)

Implements #3326

sjut8yt8ja


app/javascript/mastodon/locales/en.json
```diff
@@ -2,6 +2,7 @@
   "account.block": "Block @{name}",
   "account.block_domain": "Hide everything from {domain}",
   "account.blocked": "Blocked",
+  "account.direct": "Direct Message @{name}",
```
Gargron: I would prefer "Direct message @{name}" (small m), precedents: "Edit profile" (small p)

ThisIsMissEm: oh! sure

Also, how does the localisation system work exactly? I don't know all the languages supported, so can't translate for them

app/javascript/mastodon/locales/en.json
```diff
@@ -56,6 +57,7 @@
   "column_header.unpin": "Unpin",
   "column_subheading.navigation": "Navigation",
   "column_subheading.settings": "Settings",
+  "compose_form.direct_message_warning": "This toot will only be visible to all the mentioned users.",
```
Gargron: I like this.

Due to some recent discussions, I think we might need to include a warning not to exchange dangerous information in DMs because DMs are stored on the server and can be read by admins. I'm not sure there's a consensus on this yet, but this would be the place to put it if we decide to.

app/javascript/mastodon/locales/en.json
Gargron: You just add the English string. Volunteers add translations using weblate.joinmastodon.org or other PRs. `yarn manage:translations` just minimizes code diff later because it makes sure all files are formatted and ordered the same.

app/javascript/mastodon/locales/en.json
ThisIsMissEm: We've had a few similar discussions on switter.at, although haven't come to a conclusion yet. I think it should just be assumed that admins potentially can read direct messages for moderation purposes, they're not private messages, they're direct messages.

The inspiration behind this warning was accidentally direct messaging an abuser when dm'ing with a abuse reporter, which caused a lot of small issues.

ThisIsMissEm: Also, the wording is a bit awkward here, but I can't think of a better way to word it. I want to make sure it's stating the visibility and that ALL the users mentioned will see it.

app/javascript/mastodon/locales/en.json
ThisIsMissEm: okay, cool; I'll rebase and re-run `yarn manage:translations`

Gargron: FWIW i wouldn't be worried about the casing on files other than en.json. It'll get translated eventually anyway. So you can save yourself that trouble imo.

app/javascript/mastodon/locales/en.json
Gargron: Right. Okay. I was gonna post a strawpoll about that to gather feedback, but mastodon.social went down due to Docker issues.

Re: awkward wording, yes, the "only" followed by "all the" is a bit odd. Is "all the" necessary? "This toot will only be visible to mentioned users" -> If I mention someone, they will see it.

Gargron (approved):

app/javascript/mastodon/locales/en.json
ThisIsMissEm: It's more, I expected it to only be the first mention, i.e., in:

```
@switter I've been having issues with @lolahunt, they're so aussie
```

That going to lolahunt & switter, when I only meant to send to switter is an issue. 

Where as if I did this, I'd expect it to go to both:
```
@switter @lolahunt really loving switter! thank you!
```

(lola, you're absolutely fine promise)

ThisIsMissEm: So I wanted a way to communicate that it's ANY mention in the Toot.

app/javascript/mastodon/locales/en.json
ThisIsMissEm: That's kinda why I did it as a separate commit.

app/javascript/mastodon/locales/en.json
Gargron: How about "The toot will be sent to all mentioned users"?

ThisIsMissEm: hm, yeah, but that doesn't indicate it's "private"

ThisIsMissEm:
After using this in production for a bit, there's a few issues I've noticed:

1) Using search to find a user and then clicking the profile direct message doesn't close search, it stays open, making the button look like it didn't do anything.
2) I think the message should extract and last all the usernames from the message: "This toot will be sent directly to @..., @..., and ..."

PR #6972, opened 2018-03-30
Better admin view of statuses

ClearlyClaire:
Some fixes in the admin view:
- Clearly display whether a toot is a boost or not (previously, boosts were misrepresented as original content from the booster, without the media attachments)
- Display toot's privacy

screenshot-2018-3-30 etat du compte - mastodon instance perso de thibg

This also changes how toots are displayed on the public/simple page.

screenshot-2018-3-30 t h i b g thib sitedethib com

### DIFF
--- a/app/javascript/styles/mastodon/stream_entries.scss
+++ b/app/javascript/styles/mastodon/stream_entries.scss
@@ -85,6 +85,13 @@
 
         .status__relative-time {
           color: $lighter-text-color;
+          float: left;
+        }
+
+        .status__visibility {
+          color: $lighter-text-color;
+          float: right;
+          padding-left: 4px;
         }
       }
     }
--- a/app/views/admin/statuses/index.html.haml
+++ b/app/views/admin/statuses/index.html.haml
@@ -36,7 +36,7 @@
         .batch-checkbox
           = f.check_box :status_ids, { multiple: true, include_hidden: false }, status.id
         .activity-stream.activity-stream-headless
-          .entry= render 'stream_entries/simple_status', status: status
+          .entry= render 'stream_entries/status', status: status
         .account-status__actions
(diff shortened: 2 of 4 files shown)


nightpool:
Can you include a screenshot of the new boost indication?

Sylvhem:
Have you tried to put the privacy icon to the right of the timestamp instead of the left? By doing so, all the icons will be aligned. I think it'll look even better like that.

ThibG:
@Sylvhem yes, why not!
Pushed a change to do just that.

screenshot-2018-4-2 t h i b g thib sitedethib com

screenshot-2018-4-2 t h i b g thib sitedethib com 1

Sylvhem:
@ThibG I was more thinking of the way toots are displayed on the admin panel (your first screenshoot).

ThibG:
Oh ok, sorry about that. I'll try to do that.

ThibG:
@Sylvhem updated the PR, this is the current result:

screenshot-2018-4-2 etat du compte - mastodon instance perso de thibg

Gargron:
Notes before review:

- There is another PR that touches reports awaiting review currently, so careful about conflicts
- I feel like there might be value in stopping trying to display toots in admin UI like real toots and instead just use simplified table views?

ThibG:
- As far as I know, there is no current PR conflicting this.
- Hm… maybe. Although I'm not too sure how that table should look like.

ThibG:
@Gargron ping? Admittedly, there are two changes here: a really simple fix to admin views, and an additional marker for toot privacy in all static views. The second change has value in both admin and non-admin views. I could split it in two PRs if you want.

Gargron:
I think this is obsolete now.

PR #6996, opened 2018-04-01
Do not default SMTP verify mode to "peer", default to "none"

Gargron:
Resolves #6823


ProgVal:
When `peer` is selected, a client certificate is required.

Gargron:
The setup utility is not omnipotent. If extra config is needed, just edit the .env file.

nightpool (approved):
gonna approve for now and we can tweak if it continues to be an issue

ProgVal:
LGTM

PR #7012, opened 2018-04-02
Add label for sensitive pictures

ClearlyClaire:
Add a “sensitive” label besides the eye icon to make sensitive media more obvious, refactor a bit to match glitch-soc's implementation.

### DIFF
--- a/app/views/stream_entries/_content_spoiler.html.haml
+++ b/app/views/stream_entries/_content_spoiler.html.haml
@@ -1,7 +1,0 @@
-.media-spoiler-wrapper{ class: sensitive == false && 'media-spoiler-wrapper__visible' }
-  .spoiler-button
-    .icon-button.overlayed
-      %i.fa.fa-fw.fa-eye
-  .media-spoiler
-    %span= t('stream_entries.sensitive_content')
-    %span= t('stream_entries.click_to_show')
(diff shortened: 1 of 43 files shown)


ThibG:
This is what it looks like in the Web UI:
screenshot-2018-4-3 mastodon instance perso de thibg 1

And in on public pages:
screenshot-2018-4-3 t h i b g thib sitedethib com

app/views/stream_entries/_content_spoiler.html.haml
```diff
@@ -1,7 +0,0 @@
-.media-spoiler-wrapper{ class: sensitive == false && 'media-spoiler-wrapper__visible' }
-  .spoiler-button
-    .icon-button.overlayed
-      %i.fa.fa-fw.fa-eye
-  .media-spoiler
-    %span= t('stream_entries.sensitive_content')
-    %span= t('stream_entries.click_to_show')
```
ykzts: NOTE: no longer used.

ykzts (approved):
👍

Gargron (changes requested):
I am not quite okay with the visual design of this, though I consider it useful. What if the "sensitive" label was positioned next to / instead of the "GIF" indicator and in the same style as that one?

ThibG:
I'm not sure about positioning the label next to / instead of the “GIF” indicator, both because it's a different thing entirely and makes more sense around the eye icon, and because it's technically more difficult to do so.
However, there is no reason for it to be styled differently, so I'll fix that.

Issue #6192, opened 2018-01-05
Too few color variables in style sheet theme


cormojs:
Hi, I'm working on a new mastodon theme and suffering from a problem of visual accessibility.
The problem is that user re-definable color variables defined in `variables.scss` are too few so we can only create very simple colored themes. e.g. I think that It would be very difficult if we try to make "white(=opposing to default)" color theme.

Does anybody have a good idea?

YaQ00:
I think that this problem can be solved by extending variables to some extent with mastodon. For example, it is strange that text color has ``$ui-base-color`` , is not it? 

I think this is a structural defect.
Text as text color, button as button color, background as independent color, I think that each should be separated.

franck-paul:
May be you should have a look to this (light) [theme](https://github.com/skiant/mastodon-light-theme) developed by Mathieu Dubois.

I used it as a base to do my own (Cafe) theme for my [own instance](https://cafe.des-blogueurs.org/). Some screenshots [here](https://open-time.net/post/2018/01/19/Cafe-des-blogueurs-213)

But I agree that the color variables are not enough detailed to do it properly.

Sylvhem:
This is a huge accessibility issue. As of today, there is a too few variables to make a truly accessible theme.
For example, `$ui-primary-color`, which is used both as a background and a text color, needs to be:
- light enough to be used for text on a `$ui-base-color` background;
- light enough to be used as a background for a dark text;
- dark enough to be used for text on a white background.

When you put all of this constraints together, for each variables, you can't have a theme that respect the contrast recommendations every time.
I don't know what the proper fix would be, but I guess we need to add more variables.

This issue should be tagged with "accessibility" and should probably be at least of medium priority.

Cassolotl:
Have the recent merged PRs resolved this sufficiently?

Issue #333, opened 2016-12-03
Disallow digit-only hashtags


marrus-sh:
On Twitter and other sites, hashtags which consist only of characters in the `[0-9]` ASCII range are not allowed, to allow the additional use of # as a number sign (see https://mastodon.social/users/marrus_sh/updates/348872 for an example). It is not *strictly* necessary that Mastodon follow this behaviour, as users could always use a fullwidth hash `＃` or the less ambiguous numero sign `№` as their number sign instead, but I figured raising the issue/discussion would be worthwhile.

TazeTSchnitzel:
What are the OStatus/GNU Social/etc. rules for this, if any?

Gargron:
There are no rules at all. For example, GS allows periods in hashtags, which imo is a terrible choice when people end their sentences with #hashtags. I mean, this shouldn't be too hard to fix with a slightly pickier regex... Like, hashtag can't consist of only digits. Not actually sure how to write that out right now, here is the regex in question:

https://github.com/Gargron/mastodon/blob/master/app/models/tag.rb#L6

marrus-sh:
Should be

```rb
/(?:^|[^\/\w])#([[:word:]_]*[[:word_except_numbers:]_][[:word:]_]*)/i
```

where `:word_except_numbers:` is all the word characters except for digits. Ensures that one non-digit character exists but allows any number of characters before or after

marrus-sh:
**Update:** The POSIX class you want is `[:alpha:]`

```rb
/(?:^|[^\/\w])#([[:word:]_]*[[:alpha:]_][[:word:]_]*)/i
```

will do it, with the minor exception that it also disallows hashtags consisting entirely of connector punctuation (which shouldn't be a problem, as connector punctuation wouldn't make sense in isolation anyway). You could also remove the underscore here if you want to prevent, ie `@_@` from showing up with the hashtag `@_` (as it now does).

Issue #123, opened 2016-11-02
Notifications


bkardell:
Seems fairly simple to add and this would make a pretty huge difference to my experience at least.

https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

Gargron:
- [ ] Replace mentions column with notifications that include mentions, favourites, reblogs and follows
- [ ] Use desktop notifications for new items

PR #7027, opened 2018-04-03
Add a light theme

### LINKED ISSUE #317: Add light theme to web UI
Right now Mastodon displays using a `dark` theme, which understandably a lot of people like (and it does a lot to visually distinguish Mastodon from, say, Twitter). However, it would be nice if there was a profile setting for both `light` and `high-contrast` display modes, configurable in user preferences. The rationale for these are as follows:

- `light`: Emoji, as well as some images (does Mastodon support image transparency?), are designed for rendering on a white background. Emoji like those in the following image often do not feature outlines and can be difficult to see on Mastodon currently. This may not bother some users, but this presents a readability concern for those who frequently make use of such emoji.

    

    <small>(The final four emoji were displayed as text despite being followed by U+FE0F in my input… might this be a separate issue? I haven't tested though and it's possible there was something wrong on my end.)</small>

- `high-contrast`: See Issue #126. This can be the same as `light` if `light` meets contrast requirements; on the other hand, having both light and dark high-contrast modes might be preferable. 

There are other potential solutions for these issues (adding a white text-shadow on hard-to-read emoji and improving contrast on the main website skin, perhaps), but I know some people are partial to the current look of the site and separate display modes seems like the best way to keep everyone happy.

Sylvhem:
- Add a light theme (fix #317).
- Update some locales accordingly.

Mastodon's landing page with a clear theme
Mastodon's detailed informations page with a clear theme
Mastodon's main interface with a clear theme
Mastodon's profile page with a clear theme


RX14:
My personaly opinion is that it'd look a lot better with more contrast, think far less grey, more pure or almost white behind text areas. Light themes also don't have to be completely light, they can have large dark areas in them too. Discord's light theme is my favourite example of this - I actually perfer discord's light theme over the dark even though i'm a heavy user of light themes elsewhere.

lynlynlynx:
I think this theme seems Classic theme, not a light theme. And, Is classic theme really needed for every instances? Every designers want add their own themes as mastodon's second default theme 🤷‍♀️. Mastodon's second default theme should be high contrasted (inevitably this will be light theme) for accessibility. This requires large scale refactoring to completely separate variables for letters and backgrounds. Related issue #6192 .

Ref.
My instances which default themes as [classic theme](https://md.cryo.jp/about) and [high contrast theme](https://mstdn.sanin.link/about) (but not enough high contrasted yet, I'm working on refactoring for better accessibility) .

Cassolotl:
> I think this theme seems Classic theme, not a light theme. And, Is classic theme really needed for every instances? Every designers want add their own themes as mastodon's second default theme 🤷‍♀️. 

Just trying to make sure I understand you here - are you saying that vanilla Mastodon should only have one theme by default, and that if admins want a light theme they should add it manually?

I agree that a high-contrast theme with dark text on light background would be awesome, but this is still pretty great and going to be a lot easier on my eyes if it get merged (except the new toot box, but I can live with that!).

lynlynlynx:
> Just trying to make sure I understand you here - are you saying that vanilla Mastodon should only have one theme by default, and that if admins want a light theme they should add it manually?

That's right I think so, but if @Gargron accept adding several themes, the name of this theme should not be "light" to reserve the name "light" for high contrast theme.

This request is finally accepted or not, we can discuss about that vanilla Mastodon should be vanilla or not, that the new toot box should be another color(for example), the name of this theme, and so on.

Cassolotl:
Ahh I see what you mean. Well, this *is* a light theme, and if there's a separate high-contrast theme it could be called "high-contrast"! But yes, details. ;)

Sylvhem:
Thank you a lot @RX14, @lynlynlynx and @Cassolotl for your feedback. I will try to answer your various concerns.

> My personal opinion is that it'd look a lot better with more contrast, think far less grey, more pure or almost white behind text areas

Sadly, since the current variable system isn't that great, choosing a lighter color as a base background color cause some unwanted part of the UI to become white. I tried it and it didn't worked that great for me.
Of course, if it's really needed, I can still rewritte more of the default code. But as a rule of thumb, I tried to do that as little as possible to prevent the theme from breaking with every change to the HTML structure. Apart from accessibility concerns, I avoided that as much as I could.

> I think this theme seems Classic theme, not a light theme. And, Is classic theme really needed for every instances?

I'm sorry, but I'm not sure to understand what you mean by "classic". What is a classic theme exactly?

> Mastodon's second default theme should be high contrasted (inevitably this will be light theme) for accessibility.

If you take a closer look at it, I think you'll find this theme pretty accessible. Sure, it's not perfect, but I make sure to have a good contrast when it was possible. I tested colors combinations using [WebAIM's Contrast Checker](https://webaim.org/resources/contrastchecker/) and I rewritten some parts of the code to display specific elements in a different hue than the one used by default.

> This requires large scale refactoring to completely separate variables for letters and backgrounds. Related issue #6192 .

I know well about this issue, and I plan to work on it in the future. But I though it was still a good think to work on a light theme in the meantime since it's a feature people ask for.

> (except the new toot box, but I can live with that!)

We could still change that! What is that you don't like with the toot box exactly?

> Ahh I see what you mean. Well, this is a light theme, and if there's a separate high-contrast theme it could be called "high-contrast"

I think I agree with you on this. However, I think it'll be even better if the two defaul themes (dark and light) were both high-contrast themes.

Cassolotl:
>> (except the new toot box, but I can live with that!)
>
>We could still change that! What is that you don't like with the toot box exactly?

Partly that the text typed into the box would be light text on a dark background, which is a thing I would like to escape from in a light theme. And partly because I think it stands out in a way that I don't think is very attractive or complementary, if that makes sense? It's *very* dark compared to everything else, and the button bar below it looks weird... It's hard to explain, and for some of it I'm not really sure why it looks off!

> I think it'll be even better if the two defaul themes (dark and light) were both high-contrast themes.

Totally agree! :)

RX14:
If the theme requires large scale refactoring to get a satisfactory light theme, we should do the large scale refactoring first.

Cassolotl:
I do agree, but it sounds like a very big job, and who knows when that'll get done, and it doesn't look like Sylv wants to do that instead, and so we might as well in the mean time enjoy Sylv's hard work, no?

trwnh:
IMO vanilla Mastodon's default template should stick to the Mastodon branding palette. Yes, currently the only theming options are to have an admin install a custom .scss file, or to apply a userstyle with something like Stylish and Stylus. I agree mostly with lynlynlynx: 

> Is classic theme really needed for every instances? Every designers want add their own themes as mastodon's second default theme 🤷‍♀️

Which is absolutely true -- _admins_ can install one of several .scss themes like https://github.com/skiant/mastodon-light-theme, which don't need "official" designation; likewise, _users_ can install a userstyle like https://userstyles.org/styles/153362/mfc-mastodon-flat-css (shameless self-promo).

The only real limitations of userstyles is that they can only be applied on browsers that support them (usually desktops only), but in return they offer many customizations for the user, who can choose exactly what they want and don't want. The limitation of .scss themes is that they require admin installation, but in return they can be switched more easily and are more powerful due to using Sass instead of vanilla CSS.

For any user customizability, I would be in favor of allowing users to set custom color variables -- in fact, my userstyle does exactly this, by grouping elements together into logical palettes and applying variable colors, which can be customized manually or via userstyles.org's settings before install. This might require some minor refactoring (#317, #6192) but more significantly, an interface somewhere in the webapp settings.

re: @Sylvhem's contribution specifically: I'm seconding what RX14 said -- it needs more contrast. For example, the "Clean Slate" palette of my userstyle uses `#ddd` for the page background, but keeps the columns at `#fff` to contrast strongly with the default text at `#123`. Bold/strong text is a full `#000` for maximal contrast. But to add a little flair, I made the column header backgrounds `#333`. In a future revision, I'm going to make the column headers match the elephant friend in the corner.

For reference:

Clean Slate

Sylvhem:
> IMO vanilla Mastodon's default template should stick to the Mastodon branding palette.

I agree with that. That's why is kept as much as I can in this theme.

> I'm seconding what RX14 said -- it needs more contrast.

I'll add some since it seems to be the main problem with my proposal. I would just like to point out, again, that this theme has probably all the contrast it needs accessibility-wise.

RX14:
Accessibility aside, it looks boring and bland and honestly ugly to me.

Sylvhem:
> Accessibility aside, it looks boring and bland and honestly ugly to me.

Thanks for this constructive comment @RX14! It'll really help me improve this theme further.

RX14:
I apologise for that comment, it was rude. 

What I meant to say that the theme looks bland because of the low contrast. I think making the theme have higher contrast will remove the overabundance of grey tones, and make the theme much less "boring".

RX14:
To hopefully redeem myself, I've messed around with inspect element and came up with what I think would be more towards what I prefer as a light theme:



It's nowhere near usable, and I generated my colour pallette by clicking around in the colour picker, but it more closely reflects the fact that I think the *actual* text elements should be pure black text on pure white background, and that being afraid of using dark colours (top bar highlight), even for a large part of the screen area (left column, darker than most colours on this current theme) should not be shied away from. Ignore the dark scrollbars, it's an artifact of my gtk theme and firefox.

RX14:
More screenshots, I've lightened the dark colours for the header, and lightened the border-bottom on each post:



It looks better, but perhaps I've gone too far and the colours are a little washed out. I still believe far better than the OP though. I think I did quite well for a backend developer :)

Sylvhem:
Based on your feedback, I made some changes to the theme.
- The theme is now lighter overall.
- It now only use various hues of the Mastodon palette's colors.
- The contrast have been improved in some parts of the interface.

Mastodon's landing page with a light theme
Mastodon's detailed informations page with a light theme
Mastodon's main interface with a light theme
Mastodon's profile page with a light theme

xvilo:
Sorry to interrupt, but I hope you don't think this is finished, right?

RX14:
I still don't like the grey on the column backgrounds, and there still isn't much colour.

Sylvhem:
> I still don't like the grey on the column backgrounds

I'll try something about that.

> and there still isn't much colour

Since this is meant to be an official theme, I don't plan to add more color to it. I want it to looks like a clear version of the base theme, not something brand new.

RX14:
> Since this is meant to be an official theme, I don't plan to add more color to it. I want it to looks like a clear version of the base theme, not something brand new.

If the result doesn't look the best we can make it, isn't that worse than having a theme which isn't 100% "up to brand". I'm sure you can make something decent from the existing colour pallete if you tried.

Sylvhem:
- The composition column is now darker while the other columns are now white.
- On the landing page, the text on login in button is now white.
- On previous version, the landing page's disabled boost buttons were white. They are now of the same color that the other disabled buttons.

Mastodon's landing page with a light theme.
Mastodon's main interface with a light theme.

RX14:
@Sylvhem I think that just applying a single dark colour to the column tops would do a world of difference here.

trwnh:
looks better! just a few concerns:
* inactive buttons maybe shouldn't be pure white -- there needs to be a distinction between active and inactive buttons
* black on blue is a bit hard to read
* popup menu being dark grey looks out of place

Sylvhem:
> I think that just applying a single dark colour to the column tops would do a world of difference here.

I'm sorry, but I'm not exactly sure what you mean here. Do you mean having a single dark color used for the text in both the column titles (Home, Local timeline…) and the icons?

> inactive buttons maybe shouldn't be pure white -- there needs to be a distinction between active and inactive buttons

Don't worry, they are lighter, not pure white.

> black on blue is a bit hard to read

That's what I thought too, but using white on blue wasn't conclusive either. Actually, the contrast ratio was even worst.

> popup menu being dark grey looks out of place

Based on your comment, I just changed it. Does it looks better like this?

Mastodon's main interface with a light theme.

RX14:
> Do you mean having a single dark color used for the text in both the column titles (Home, Local timeline…) and the icons?

I mean use a dark, rich colour for the background of the column titles, with white text on top. See my screenshots for reference.

> That's what I thought too, but using white on blue wasn't conclusive either. Actually, the contrast ratio was even worst.

I'd darken the highlight colour then (and use the same colour for the column title backgrounds) and use white text.

roipoussiere:
I really like @RX14's proposition, I think there is still too much grey on @Sylvhem theme.

Sylvhem:
I really don't know about that. I really want this theme to match the dark one as much as possible.
I really would want @Gargron's advice on this.

trwnh:
OK, I've been playing around a bit with the official Mastodon color palette -- how's this? Keeping in mind @RX14's suggestion for a darker column head, and using *only* the four official Mastodon colors + 50% brightness on the muted text:

image

Was this more of the palette you had in mind?

RX14:
@trwnh that looks really good actually.

Sylvhem:
Can someone review this?

Sylvhem:
Since this pull request is meant to solve a [high priority issue](https://github.com/tootsuite/mastodon/issues/317), it would really nice if it could be merged in time for Mastodon 2.4. So I would really want to have your opinion on the look and feel of the theme @Gargron.

ghost (approved):
Great work!

Aldarone (approved):

ykzts (approved):
:heart:

Cassolotl:
Is this the same light theme that Glitch recently added?

Sylvhem:
Yes, it is :).

Cassolotl:
@Sylvhem Awesome! Thanks. :) I just need to wait and find out if the low-contrast bits I've found happen on vanilla as well as on the Glitch theme, and if they do I'll make a new issue.

Sylvhem:
@Cassolotl  No problem, and don't hesitate to tag me in it if you open one.

steckerhalter:
@Sylvhem I'm running 2.4.0 but don't have the light theme as an option in the UI. Do I need to do something so it's available?

PR #7028, opened 2018-04-03
update gem, test pam authentication

devkral:
* update gem, enables test mode which doesn't require pam libraries and can spoof pam data
* tests for pam authentication


devkral:
I messed the travis bundler setup. Should be fixed now.

Gargron (approved):

PR #7000, opened 2018-04-02
Feature: Report improvements (#6967)

ThisIsMissEm:
There's a fair bit going on here; The first commit adds a section to each report giving assignment information and last updated at dates (general metadata about the report). The third adds notes, and whilst I was doing that I released that the UI had become confusing, so I've attempted to realign it.

screen shot 2018-04-02 at 7 24 56 am
screen shot 2018-04-02 at 7 25 16 am

Definitely wanting feedback on the adjusted UI, as well as the code/implementation, I based the report notes implementation off the back of account notes; Perhaps there's a better way to implement it.

PR #7032, opened 2018-04-03
Feature: Improve reports ui

ThisIsMissEm:
After playing with the improved reports UI from #7000, I still wasn't happy with it. I've since redesigned the form for adding report notes, and added the "ActionLog" history (in a rough form) so we can see all the actions taken in relation to a specific report.

Note: I wasn't sure on the right way to implement fetching the ActionLog, so went with a relation, but now I think a Finder method would be better, as it needs to pull from the Report, the statuses of the report and the reported user; Though I'm not sure on the best way to implement this.

Here's what the page now looks like:

localhost_3000_admin_reports_2 2

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -61,27 +61,28 @@ class Report < ApplicationRecord
   end
 
   def history
-    report_log = Admin::ActionLog.where(
-      target_type: 'Report',
-      target_id: id,
-      created_at: created_at..updated_at
-    ).unscope(:order)
+    time_range = created_at..updated_at
 
-    target_account_log = Admin::ActionLog.where(
-      target_type: 'Account',
-      target_id: target_account_id,
-      created_at: created_at..updated_at
-    ).unscope(:order)
+    sql = [
+      Admin::ActionLog.where(
+        target_type: 'Report',
+        target_id: id,
+        created_at: time_range
+      ).unscope(:order),
 
-    statuses_log = Admin::ActionLog.where(
-      target_type: 'Status',
-      target_id: status_ids,
-      created_at: created_at..updated_at
-    ).unscope(:order)
+      Admin::ActionLog.where(
+        target_type: 'Account',
+        target_id: target_account_id,
+        created_at: time_range
+      ).unscope(:order),
 
-    query = "((#{report_log.to_sql}) UNION ALL (#{target_account_log.to_sql}) UNION ALL (#{statuses_log.to_sql})) as admin_action_logs"
-    sql = Admin::ActionLog.connection.unprepared_statement(query)
+      Admin::ActionLog.where(
+        target_type: 'Status',
+        target_id: status_ids,
+        created_at: time_range
+      ).unscope(:order),
+    ].map { |query| "(#{query.to_sql})" }.join(' UNION ALL ')
 
-    Admin::ActionLog.from(sql)
+    Admin::ActionLog.from("(#{sql}) AS admin_action_logs")
   end
 end
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -61,13 +61,26 @@ class Report < ApplicationRecord
   end
 
   def history
-    report_log = Admin::ActionLog.where(target_type: 'Report', target_id: id).where(:created_at => created_at..updated_at).unscope(:order)
-    target_account_log = Admin::ActionLog.where(target_type: 'Account', target_id: target_account_id).where(:created_at => created_at..updated_at).unscope(:order)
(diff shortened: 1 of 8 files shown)


ThisIsMissEm:
I really want to try to get the additional ActionLog items that are relevant into the report, but I'm not sure how to do this. Open to suggestions.

nightpool:
it would be really good if the moderation log and action log displayed
interleaved with each other I think

ThisIsMissEm:
@nightpool oh, I agree, but I've absolutely no idea how to do that in rails. Not to mention that it's not the full action log for now, which I'm struggling to build.

Gargron:
I'm liking this a lot. :+1:

app/models/report.rb
```diff
@@ -39,4 +39,36 @@ def statuses
...
+    report_log = Admin::ActionLog.where(target_type: 'Report', target_id: id).unscope(:order)
+    target_account_log = Admin::ActionLog.where(target_type: 'Account', target_id: target_account_id).unscope(:order)
+    statuses_log = Admin::ActionLog.where(target_type: 'Status', target_id: status_ids).unscope(:order)
+
+    sql = Admin::ActionLog.connection.unprepared_statement {
+      "((#{report_log.to_sql}) UNION ALL (#{target_account_log.to_sql}) UNION ALL (#{statuses_log.to_sql})) as admin_action_logs"
+    }
+
+    Admin::ActionLog.from(sql)
+  end
```
ThisIsMissEm: This probably isn't ideal or perfect, so I'm open to suggestions here @nightpool @Gargron

app/models/report.rb
```diff
@@ -39,4 +39,49 @@ def statuses
...
+    ).unscope(:order)
+
+    statuses_log = Admin::ActionLog.where(
+      target_type: 'Status',
+      target_id: status_ids,
+      created_at: created_at..updated_at
+    ).unscope(:order)
+
+    query = "((#{report_log.to_sql}) UNION ALL (#{target_account_log.to_sql}) UNION ALL (#{statuses_log.to_sql})) as admin_action_logs"
+    sql = Admin::ActionLog.connection.unprepared_statement(query)
```
Gargron: It seems that this method takes a block argument, i.e. `unprepared_statement { query }`, but I am not totally sure why unprepared_statement is needed in here?

ThisIsMissEm:
@Gargron this is just what I found when looking up how to do union queries with rails; there's probably a better way to do this, though I'm a bit out of my depth with rails knowledge needed.

app/models/report.rb
Gargron: We may want to just use https://github.com/brianhempel/active_record_union to do this the clean way

Gargron:
Here is a fix for the UNION stuff that works for me:

```diff
diff --git a/app/models/report.rb b/app/models/report.rb
index e1bfa99..5b90c7b 100644
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -61,27 +61,28 @@ class Report < ApplicationRecord
   end
 
   def history
-    report_log = Admin::ActionLog.where(
-      target_type: 'Report',
-      target_id: id,
-      created_at: created_at..updated_at
-    ).unscope(:order)
-
-    target_account_log = Admin::ActionLog.where(
-      target_type: 'Account',
-      target_id: target_account_id,
-      created_at: created_at..updated_at
-    ).unscope(:order)
-
-    statuses_log = Admin::ActionLog.where(
-      target_type: 'Status',
-      target_id: status_ids,
-      created_at: created_at..updated_at
-    ).unscope(:order)
-
-    query = "((#{report_log.to_sql}) UNION ALL (#{target_account_log.to_sql}) UNION ALL (#{statuses_log.to_sql})) as admin_action_logs"
-    sql = Admin::ActionLog.connection.unprepared_statement(query)
-
-    Admin::ActionLog.from(sql)
+    time_range = created_at..updated_at
+
+    sql = [
+      Admin::ActionLog.where(
+        target_type: 'Report',
+        target_id: id,
+        created_at: time_range
+      ).unscope(:order),
+
+      Admin::ActionLog.where(
+        target_type: 'Account',
+        target_id: target_account_id,
+        created_at: time_range
+      ).unscope(:order),
+
+      Admin::ActionLog.where(
+        target_type: 'Status',
+        target_id: status_ids,
+        created_at: time_range
+      ).unscope(:order),
+    ].map { |query| "(#{query.to_sql})" }.join(' UNION ALL ')
+
+    Admin::ActionLog.from("(#{sql}) AS admin_action_logs")
   end
 end
```

ThisIsMissEm:
@Gargron this is now ready for review again, the 1 issue to fix from code climate isn't an issue, because we know what we're doing™

Gargron (approved):

PR #7043, opened 2018-04-04
Ensure SynchronizeFeaturedCollectionWorker is unique and clean up

### LINKED ISSUE #7041: Sidekiq full of duplicate SynchronizeFeaturedCollectionWorker jobs
https://github.com/tootsuite/mastodon/pull/7019 fixed the issues with defederation (Sidekiq full of ~170k jobs, most failing), but after applying the patch, Sidekiq constantly has ~10 busy jobs doing `SynchronizeFeaturedCollectionWorker` for the same account ID. When one completes, it seems to just start another one:

screenshot 2018-04-04 08 48 36

I checked the account in question, and it's a remote account that posts fairly frequently (once every few minutes) and has two pinned toots. Oddly when the account is viewed on toot.cafe, only one pinned toot is shown.

Right now the impact is just that Sidekiq is always uselessly busy, but I'm worried this could cause problems if the user in question starts posting _more_ frequently, or if this points to deeper issues, etc.

Mastodon 2.3.3, Ruby 2.4.2 w/ jemalloc, Ubuntu 16.10.

* * * *

Gargron:
Fix #7041 

I have changed @nolanlawson's fix for the underlying issue, because I believe that the only condition under which `@account` could be nil in that line is if the lock was not acquired. So returning early in the case should be a better fix. Since this is a rare and strange issue in the first place, I don't know if I'm right.

The sidekiq job thing should ensure multiple jobs with the same argument should not be allowed. Although it will not clear out already enqueued jobs.

### DIFF
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -23,10 +23,10 @@ class ActivityPub::ProcessAccountService < BaseService
         create_account if @account.nil?
         update_account
         process_tags
-      else
-        return
       end
     end
+
+    return if @account.nil?
 
     after_protocol_change! if protocol_changed?
     after_key_change! if key_changed?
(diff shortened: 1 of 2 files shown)


Gargron:
I believe CodeClimate is wrong on this one :P

PR #7048, opened 2018-04-05
Remove "nsfw" category for sensitive statuses in OStatus serializer

### LINKED ISSUE #7011: OStatus/Atom: Remove explicit "nsfw" <category> from CW posts
Hi,

Over at https://github.com/friendica/friendica I noticed that Mastodon posts sent over OStatus that have a Content Warning have a `<category name="nsfw">` set even though it doesn't appear as a hashtag in the text.

We recently improved the way we are showing users why a post is collapsed in Friendica, whether it is because of an author-set CW or a user-defined filter settings (for example, hide posts with the hashtag `#nsfw`). And I noticed that Content Warning posts were also triggering the `#nsfw` filter rule even though the tag didn't appear in the post body.

Semantically, CW and NSFW are serving two different purpose and as such, the `#nsfw` tag shouldn't be automatically tacked on CW posts. If understand correctly, Mastodon posts can be set as "sensitive" without having a CW set, and in this case I believe the `#nsfw` tag still is warranted.

* * * *

Gargron:
Fix #7011

### DIFF
--- a/app/lib/ostatus/atom_serializer.rb
+++ b/app/lib/ostatus/atom_serializer.rb
@@ -364,8 +364,6 @@ class OStatus::AtomSerializer
       append_element(entry, 'category', nil, term: tag.name)
     end
 
-    append_element(entry, 'category', nil, term: 'nsfw') if status.sensitive?
-
     status.media_attachments.each do |media|
       append_element(entry, 'link', nil, rel: :enclosure, type: media.file_content_type, length: media.file_file_size, href: full_asset_url(media.file.url(:original, false)))
     end
--- a/app/services/process_hashtags_service.rb
+++ b/app/services/process_hashtags_service.rb
@@ -7,7 +7,5 @@ class ProcessHashtagsService < BaseService
     tags.map { |str| str.mb_chars.downcase }.uniq(&:to_s).each do |tag|
       status.tags << Tag.where(name: tag).first_or_initialize(name: tag)
     end
-
-    status.update(sensitive: true) if tags.include?('nsfw')
   end
 end
--- a/spec/lib/ostatus/atom_serializer_spec.rb
+++ b/spec/lib/ostatus/atom_serializer_spec.rb
@@ -386,12 +386,6 @@ RSpec.describe OStatus::AtomSerializer do
         expect(entry.category[:term]).to eq 'tag'
       end
 
-      it 'appends category element for NSFW if status is sensitive' do
-        status = Fabricate(:status, sensitive: true)
-        entry = OStatus::AtomSerializer.new.entry(status.stream_entry)
-        expect(entry.category[:term]).to eq 'nsfw'
-      end
-
       it 'appends link elements for media attachments' do
         file = attachment_fixture('attachment.jpg')
         media_attachment = Fabricate(:media_attachment, file: file)


ykzts (approved):

ykzts:
https://github.com/tootsuite/mastodon/blob/0950900/spec/lib/ostatus/atom_serializer_spec.rb#L389-L393 :eyes:

Cassolotl:
Hello, sorry for the non-technical question, but: will this break whether #nsfw forces images to be hidden?

Gargron:
@Cassolotl As far as I am aware "#nsfw forces images to be hidden" was broken when we switched from OStatus to ActivityPub half a year or so ago. It might have still been working *towards GNU social and really-out-of-date nodes* (on the receiving end), but I don't remember any ActivityPub special handling for the "nsfw" hashtag (we already decided nsfw is a problematic term)

Cassolotl:
@Gargron See this tiny test thread: https://dev.glitch.social/@cassolotl/99806996137686905

app/lib/ostatus/atom_serializer.rb
```diff
@@ -364,8 +364,6 @@ def serialize_status_attributes(entry, status)
       append_element(entry, 'category', nil, term: tag.name)
     end
 
-    append_element(entry, 'category', nil, term: 'nsfw') if status.sensitive?
```
nightpool: I *think*, reading #7011 closely, the logic that's desired is `if status.sensitive? && status.with_media?`

I would prefer that instead of removing it altogether, we keep it like this for compatibility's sake. I agree that NSFW is not the term we *want* to be using, but harm reduction is more important then ideological purity, and it's the term that part of the fediverse has standardized on.

nightpool (changes requested):

nightpool:
from #7011:

> If understand correctly, Mastodon posts can be set as "sensitive" without having a CW set, and in this case I believe the #nsfw tag still is warranted.

Gargron:
The nsfw tag is never warranted. It was suggested by the Pleroma developer at the very start when Pleroma was a frontend for GNU social. GNU social itself has never treated "nsfw" as anything special. It's not a standard. Pleroma doesn't use GNU social anymore, they use ActivityPub, Mastodon doesn't use OStatus anymore, we use ActivityPub. ActivityPub does not need to have special "nsfw" tags to transmit any data. It serves zero purpose.

nightpool:
If no GNU Social frontends support #nsfw then i'm fine dropping it.

nightpool (approved):
1. the twitter crossposting thing is a bad hack—it only works at 2 degrees of remove, and it leads to the origin server and any remote servers that received the status to have different ideas of what the status actually is. this is bad. Instead, the crossposter should just handle this case itself, and create the status with `sensitive: true` if it contains the nsfw hashtag.

2. after a cursory code inspection, I *think* this will break pleroma federation for the default (non-mastodon) frontend, but because we use as:sensitive they have no-one but themselves to blame for that.

lambadalambda:
Bit late to the party, but both Qvitter and Pleroma on GS support this

nightpool:
by pleroma on gs you mean pleroma fe with a GS backend? didn't know that was supported. I checked a couple qvitter instances but couldn't see anything indicating they actually hid any content.

lambadalambda:
Gargron is talking about it two posts above :) I know both SPC and Smug do it in Qvitter, too, although they might only do it for logged-in users.

rogerbraun:
Incoming from OStatus also worked fine with the hack.
screenshot-2018-4-23 mastodon

lambadalambda:
Incoming from OStatus also worked fine with the hack.
screenshot-2018-4-23 mastodon

mal0ki:
@lambadalambda are you only talking about incoming from? I think this one mostly references outgoing to. Also see the referenced #7011 for further discussion. 

and let us know if there's something we're still missing here.

lambadalambda:
both directions worked fine

lambadalambda:
They use a qvitter plugin for it that does it.

Lionirdeadman:
I think this 'hack' should be kept for as long as Mastodon still uses OStatus as fallback.

Otherwise this harms communication with certain OStatus implementation.

It doesn't harm development of Mastodon and therefore there is no reason to remove it.

treyssatvincent:
You don't think <category term="NSFW"/> mean sensitive ?

Even if it's not a standard, it's usage (and widely used as is around the web).

mal0ki:
What I see here are two concerns of breaking, either on GNUSocial’s side or directly on Friendica, see issue #7011  . I understand why having everything tagged as NSFW is bad and I understand why it's useful. 

Either decision will break something, that’s why this PR is labeled as breaking. When Mastodon started going into ActivityPub it was also announced that eventually it will break with OStatus. Now, I do know that GNUSocial is working to implement ActivityPub through the summer. 
So the real concern here is, could we hold out with this hack for the next few months (we can set a delay until August maybe?), and in the meanwhile look for better alternatives, and offer GNUSocial some time to catch up with this development?

I do know there has been plenty of time to find alternatives to a bad hack, but sometimes we aren’t encouraged to fix it until we have a reason to, like it being removed in its entirety. 

I am open to hear _alternative solutions_ to this issue, that is something else than “don’t do it because it will break” (even if that’s also a valid response, which we’ve now heard). 

Question to all, what adjustments could we make not only on the Mastodon side, but on other sides, to make this more manageable?

lambadalambda:
As far as I can tell #7011 isn't about the sensitive->nsfw tag mapping at all, but about not forcing nsfw tags on CW. If CWs force the sensitive flag, having the extra tag would not be a problem for #7011.

mal0ki:
@lambadalambda This one should refer to what I mentioned:
https://github.com/tootsuite/mastodon/issues/7011#issuecomment-378041014

lambadalambda:
I think @MrPetovan is simply mistaken about where the problem lies. If you look at https://github.com/tootsuite/mastodon/blob/master/app/models/status.rb#L324, you'll see that ALL posts with a CW are also set to sensitive in Mastodon. There are no non-sensitive CW posts in Mastodon, which is what is being discussed in #7011. This change can only 'fix' things in Friendica if they don't parse the sensitive bit, in which case any kind of sensitivity signaling will fail for both them and GNU Social.

MrPetovan:
I don't mind CW sensitive posts, we are collapsing CW posts by default on Friendica, I just minded the arbitrary add of the `#nsfw` tag to sensitive posts, including CW ones. There's no "sensitive" bit in OStatus to parse, only the CW string and the slapped `#nsfw` tag that is now removed.

lambadalambda:
So now there's no signaling of any way to ostatus servers that something is sensitive, even though there was an existing way that worked for over a year now.

mal0ki:
@lambadalambda would they even do anything with it though? Except for the few exceptions who run a plugin?

Gargron:
>even though there was an existing way that worked for over a year now.

We used it when Mastodon<->Mastodon communication was OStatus, because there was no other way to communicate that bit of information which to us was crucial, and Pleroma which at the time was a GS frontend also benefitted from it. But now that Mastodon does not use OStatus, and Pleroma does not use OStatus, only users of that optional Qvitter plugin you mentioned remain who would still make any use of that information. Vanilla GNU social was not handling "sensitive" stuff in any special way for that year, was it?

I was happy to oblige @MrPetovan's request because encoding that information the same way a user-land hashtag would be encoded, using a hashtag with imperfect nomenclature, was not something I liked, and removing it now that almost nobody has a use for it, makes the code a step cleaner.

MrPetovan:
There's the CW string, still.

treyssatvincent:
@lambadalambda adding manually a #nsfw` tag will still work ? It keeps Mastodon a little further away from OStatus, but we know Mastodon aim to be fully ActivityPub.

The main problem for me here is adding a `#nsfw` hashtag add serialized/parseable/semantic information and because of this PR it will be ignored for Mastodon users.

It may make the code cleaner and closer to standard, but it's usage, NSFW meaning is known by almost everybody on Internet.

MrPetovan:
> It may make the code cleaner and closer to standard, but it's usage, NSFW meaning is known by almost everybody on Internet.

The problem doesn't lie with the hashtag itself, if it's added manually than I will do something about it. The problem is that it was automatically added for CW, and not all CW are NSFW.

mal0ki:
I want to append to this:
There's been a warning that Mastodon would break federation with OStatus eventually, and there's been a lot of time to get read for such a break. This is a minor fix, but a step towards that break.

treyssatvincent:
@MrPetovan I fully agree with you CW ≠ sensitive. My point is just, why removing this : 

``` ruby
status.update(sensitive: true) if tags.include?('nsfw')
```

It only affect Mastodon side and don't concern #7011

mal0ki:
That's true @MrPetovan

kaniini:
@MrPetovan What is Friendica's current behavior for this?  I'm willing to work on implementing the appropriate conversions in Pleroma, but I can't promise that other OStatus implementations will follow.

kaniini:
Personally I would prefer that Mastodon retain current OStatus behavior, formally define a version where OStatus will be dropped, and then drop OStatus in it's entirety when that release happens.  In other words, some sort of "OStatus flag day."

MrPetovan:
There are two different behaviors on Friendica:
- CW posts are collapsed by default, but a setting switch can disable this behavior.
- An optional content filter feature allows to collapse posts based on keywords and/or hashtags. The `nswf` keyword is the default.

Before this change, with the CW disabled and the content filter feature enabled, CW posts would still be collapsed because of the existence of a machine-added `#nsfw` tag in the posts.

After the change, these posts aren't collapsed if their author didn't actually add the `nsfw` string in their post, in a hashtag or not.

kaniini:
@MrPetovan How do you propose handling attachments that are sensitive then?

MrPetovan:
We don't have this concept in Friendica so we haven't looked into it. Personally, I ask remote contacts to manually add the `#nsfw` tag to such posts.

kaniini:
@MrPetovan when Friendica has an alternate proposal for NSFW media, we will implement that, but retaining NSFW media support is important to us (Pleroma).

MrPetovan:
I get you, but since Friendica never carried this information, there’s no automated way of doing it at the moment.

jeroenpraat:
@Gargron Am I right to assume that with this commit GNU social's sensitive/NSFW posts are displayed unfiltered on the federated timeline? If so, I want to suggest to drop all OStatus posts from the federated timeline. That also solves the annoyance that GNU Social posts can't be filtered on language.

treyssatvincent:
This make sense, revert the change on `ProcessHashtagsService` or drop the whole OStatus compatibility.
-------- Message d'origine --------

jeroenpraat:
@treyssatvincent That is not what I'm suggesting. I think it is a good thing to keep the OStatus compatibility as long as needed by the Mastodon community. So I support people connecting with OStatus users, but to keep it off the federated timeline as explained in my previous reply.

treyssatvincent:
@jeroenpraat removing OStatus from the federated timeline solve one of the problems raised here and could be a first step before dropping OStatus support.

mal0ki:
@jeroenpraat No, that should not be what happens. This is in the other direction. Not to Mastodon. As I interpreted it. If stuff is still tagged NSFW it will get hidden etc. (I believe I was reassured about this a few times). It's just arbitrarily adding nsfw to the sensitive statuses, which aren't necessarily nsfw (any CW). And it caused issues in Friendica who are filtering NSFW specifically, and then ended up filtering normal CWs too because of that flag. 

Does this clear it up?

lambadalambda:
@mal0ki #nsfw tagged posts from gnu social will show up unfiltered, it already happens on niu.moe  which tracks master.

mal0ki:
Thank you @treyssatvincent . For some reason I was led to believe the case was as I described it. It's what I had asked about before, and I had felt like I was receiving a confirmation on that it would still be handled accordingly.

mal0ki:
Okay, so this is more in line with what I wanted to say initially, so here goes: 
I want us to hide content from Ostatus when it shows up on our timelines, if they tagged it with #nsfw it should auto-hide on our side. I don't care if we tag on an artifical #nsfw or if we do it some other way. I need for us to have it done simply because we don't want untagged nudes, violence etc on the PTLs. I do not necessarily think that this would be a major problem if we did merge this PR, but I don't want it to be a problem.

If the merging of this PR changes this behaviour (which I had wrongly been led to believe wasn't the case when we discussed it last), I don't think we should do it. 

It's better to wait until we find a better way to deal with it. 

That said, I want to keep in touch with @kaniini  and @MrPetovan  about how they can handle it on the Ostatus side, so we can handle it differently too (I'll take responsibility for this myself). If you want to maybe discuss this further on one of your relevant git/lab/hub repo's i'd be happy to keep an eye on the issue and give creative input when I can. 

Does this sound like an okay compromise?

MrPetovan:
How Mastodon handles posts with the `#nsfw` tag is unrelated to my initial request, which was for Mastodon to stop slapping the tag on sensitive posts when transporting with OStatus.

This part can be freely reverted without any consequences for federation with Friendica or GNU Social:
https://github.com/tootsuite/mastodon/pull/7048/files#diff-ee3077b014d270244fe9d163d0fab100L11

PR #7052, opened 2018-04-05
Remove Puma pidfile before boot if container receives SIGTERM

ladyisatis:
Fixes an issue mentioned in https://github.com/docker/compose/issues/1393

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     image: tootsuite/mastodon
     restart: always
     env_file: .env.production
-    command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    command: bash -c "rm -f /mastodon/tmp/pids/server.pid; bundle exec rails s -p 3000 -b '0.0.0.0'"
     networks:
       - external_network
       - internal_network


Gargron:
Question: Is the PID file even needed inside Docker? Can it be disabled via some config? Would be much preferable to command like this

Reverite:
I tried looking, but nope, Puma doesn't have an ability to disable it, and several other projects have had to do something this "hacky" unfortunately.

Reverite:
On second thought, Docker supports mounting directory as a tmpfs, so something like this in the docker compose yml for the web container might work?

```
    tmpfs:
      - /mastodon/tmp/pids:uid=991,gid=991,mode=0644
```

moritzheiber:
I wouldn't want to hardcode the UID/GID though

Reverite:
After discussion on Discord, I tried this:

```
  tmpfs:
    - /mastodon/tmp/pids
```

Instance went down and wouldn't boot because of these errors:

```
/mastodon/vendor/bundle/ruby/2.4.4/gems/rack-2.0.3/lib/rack/server.rb:362:in 'initialize': Permission denied @ rb_sysopen - /mastodon/tmp/pids/server.pid (Errno::EACCES)
```

Gargron (approved):

iGEL:
@Gargron In our Rails app (not mastodon), I just set the Pid file to `/dev/null`, which seems to work fine.
`rails server -b 0.0.0.0 -P /dev/null`

PR #7053, opened 2018-04-05
Add a circuit breaker for ActivityPub deliveries

Gargron:
>A circuit breaker is an automatically operated electrical switch designed to protect an electrical circuit from damage caused by excess current from an overload or short circuit. **Its basic function is to interrupt current flow after a fault is detected.**

The purpose of `DeliveryFailureTracker`, and how it is integrated, is to stop *enqueueing* deliveries to dead endpoints (after 7 days of continuous failures). Items that are already enqueued are not affected, and short term failures are ignored.

When a host is experiencing issues, it can respond way too slowly, maxing out the 10 second timeout and failing. More and more deliveries to the same host will be enqueued, congesting the push queue, each job taking 10 seconds... In this case, what we need is:

- Do not stop *enqueueing* items, because the host will likely come back sometime later, and will need those items
- But we need to reduce the failure time from 10 seconds to 0 seconds

Stoplight is a circuit breaker gem. It wraps our delivery code. There are three states:

- green: deliveries go through transparently
- red: after 3 failures, deliveries fail instantly (delivery not actually executed)
- yellow: after cooldown period of 60 seconds, deliveries go through transparently again, and either succeed (back to green) or fail (back to red)

The threshold (3) and cooldown (60 seconds) could be adjusted. Perhaps retry scheduling could also be adjusted to better work with those parameters, for example, the first 3 failures could follow the backoff algorithm in use right now, after that, wait at least 60 seconds before scheduling the next retry.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ gem 'devise-two-factor', '~> 3.0'
 group :pam_authentication, optional: true do
   gem 'devise_pam_authenticatable2', '~> 9.0'
 end
+
 gem 'net-ldap', '~> 0.10'
 gem 'omniauth-cas', '~> 1.1'
 gem 'omniauth-saml', '~> 1.10'
@@ -79,6 +80,7 @@ gem 'sidekiq-bulk', '~>0.1.1'
 gem 'simple-navigation', '~> 4.0'
 gem 'simple_form', '~> 3.4'
 gem 'sprockets-rails', '~> 3.2', require: 'sprockets/railtie'
+gem 'stoplight', '~> 2.1.3'
 gem 'strong_migrations'
 gem 'tty-command'
 gem 'tty-prompt'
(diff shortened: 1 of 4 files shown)


davefp:
Solid patch. :shipit:

ThibG:
This means that out of those 16 retries, some of them actually aren't, though?
Wouldn't it be cleaner/simpler to just tweak the time between attempts?

ClearlyClaire (approved):
Looks good to me, if, as I understand from Stoplight's documentation, it immediately raises its own error in “red” states, instead of the “timeout of 0 seconds” thing hinted at in the description of your PR.

Gargron:
@ThibG That's what I meant. I'm just wondering now if maybe I should use the host only for stoplight identifier, so in the admin view of "known domains" i could then display the stoplight color easily. Because otherwise its `example.com` in one place and `https://example.com/inbox` in another...

ThibG:
Hm… I would be against that I think, as you could imagine a domain just providing webfinger handles to people, which would then use their own URL on different domains/servers. In this scenario, a single broken server would stop federation for everyone under that domain.

nightpool:
maybe compromise on `URI::parse(sharedInbox).domain`?

PR #7054, opened 2018-04-05
add minimal diagnostics to purge_removed_accounts

abochmann:
Informational messages for mastodon:maintenance:purge_removed_accounts in --force mode. Not suitable for logging thanks to the progress bar, but it's possible to record the output using script(1)

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -792,15 +792,23 @@ namespace :mastodon do
 
         begin
           code = Request.new(:head, account.uri).perform(&:code)
-        rescue StandardError
+        rescue StandardError => e
           # This could happen due to network timeout, DNS timeout, wrong SSL cert, etc,
           # which should probably not lead to perceiving the account as deleted, so
           # just skip till next time
+          progress_bar.pause
+          progress_bar.clear
+          print "ignored: #{account.acct}, #{e}\n"
+          progress_bar.resume
           next
         end
 
         if [404, 410].include?(code)
           if options[:force]
+            progress_bar.pause
+            progress_bar.clear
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -792,15 +792,23 @@ namespace :mastodon do
 
         begin
           code = Request.new(:head, account.uri).perform(&:code)
-        rescue StandardError
+        rescue StandardError => e
           # This could happen due to network timeout, DNS timeout, wrong SSL cert, etc,
           # which should probably not lead to perceiving the account as deleted, so
           # just skip till next time
+          progress_bar.pause
```
ThibG: `progress_bar.pause`, `progress_bar.clear` and `progress_bar.resume` could be moved out of the conditionals I guess, as every codepath calls them

ClearlyClaire (approved):
This could use some minor refactoring regarding `progress_bar.pause`, `progress_bar.clear` and `progress_bar.resume`, but looks good to me otherwise

lib/tasks/mastodon.rake
abochmann: The progress bar is slightly broken anyway - it tends to overflow its initial value towards the end, I assume because the instance learns about new accounts while the task is running.
I'll look at the pause/resume, but it may exceed my non-existing knowledge of the language.

ThibG: I just meant that those functions are currently called in the same order regardless of the branch that ends up being executed. Therefore, they could be moved outside of the conditional to make the code a bit cleaner.

Gargron:
I have reworked this into tootctl so this patch is no longer needed, but thank you.

PR #7057, opened 2018-04-06
When creating status, if no sensitive status is given, use default

renatolond:
Clients using the API that do not provide the sensitive flag are always posting with false sensitive option.

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -28,7 +28,7 @@ class PostStatusService < BaseService
       status = account.statuses.create!(text: text,
                                         media_attachments: media || [],
                                         thread: in_reply_to,
-                                        sensitive: options[:sensitive],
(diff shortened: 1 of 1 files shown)


app/services/post_status_service.rb
```diff
@@ -28,7 +28,7 @@ def call(account, text, in_reply_to = nil, **options)
       status = account.statuses.create!(text: text,
                                         media_attachments: media || [],
                                         thread: in_reply_to,
-                                        sensitive: options[:sensitive],
+                                        sensitive: (options[:sensitive].nil? ? account.user&.setting_default_privacy : options[:sensitive]),
```
Gargron: Why `setting_default_privacy`? That has the values "public", "unlisted", "private", etc...

renatolond: Sorry, distraction on my part, fixing

renatolond:
Updated :)

Gargron (approved):

PR #7058, opened 2018-04-06
Forward deletes on the same path as reply forwarding

Gargron:
Public reply to someone is forwarded to followers of the replied-to person. Delete of that reply is sent to the replied-to person, but is not forwarded to their followers. This fixes that.


ClearlyClaire (approved):
Looks good to me

PR #7059, opened 2018-04-06
Domain blocking as rake task

ghost:
This PR is a proposal to add domain blocking as a rake task.  The goal here is to automate the blocking of domains from a shared list.  It includes two tasks.  One to add a single domain to the blocklist and a second one to download a blocklist from a website and add all the domains to the blocklist.

Example
```
bundle exec rake domains:block['badnetwork.social','silence',false]
```

### DIFF
--- a/lib/tasks/domains.rake
+++ b/lib/tasks/domains.rake
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require 'pry'
 require 'net/https'
 require 'json'
 
--- a/lib/tasks/domains.rake
+++ b/lib/tasks/domains.rake
@@ -3,6 +3,12 @@ require 'pry'
 require 'net/https'
 require 'json'
 
+def check_severity(severity)
+  if ![:silence, :suspend, :noop].include?(severity.to_sym)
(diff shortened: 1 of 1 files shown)


Gargron:
>The goal here is to automate the blocking of domains from a shared list.

Big **no** from me to this. I implemented the domain blocks in the first place, so it's not that I am against them in principle, I think they are necessary. But automated shared lists? No. Absolutely not. There is no oversight over those lists. Remember when a couple trans activists were blocked by half of Twitter because the author of a popular autoblock tool had a personal vendetta against them? I have seen some of the domain block lists, and while many of the items on them are justified, I've also had multiple experiences of false positives being added to them without any real research or outreach from the author of the list.

Remember that a domain block is a destructive, last resort nuclear option. Everyone should be fully aware of the weight of the decision for each domain; the least you could do is confirm that it actually is what people say it is.

usbsnowcrash:
I understand your reasoning so I am closing.  I still feel like there is utility in being able to run admin commands from rake but since my main use case was to consumer someone elses list I am closing.

PR #7060, opened 2018-04-07
Make scroll bars wider

sts10:
I think the scroll bars on Mastodon's front-end could be a bit wider than 8px. I probably didn't execute this exactly right, but I _tried_ to bump it up to 12px in `basics.scss` (I admittedly did not spend a lot of time trying to figure out which scss file). 

Also, I couldn't figure out how to test this cuz I couldn't figure out how to install Ruby 2.5.1 on my machine :(

So in general this is not a great PR haha. But my request still stands -- wider scroll bars would be nice.

### DIFF
--- a/app/javascript/styles/mastodon/basics.scss
+++ b/app/javascript/styles/mastodon/basics.scss
@@ -125,3 +125,8 @@ button {
     outline: 0 !important;
   }
 }
+
+::-webkit-scrollbar {
+  width: 12px;
+  height: 12px;
+}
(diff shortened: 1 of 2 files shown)


Gargron (changes requested):

Gargron (approved):

sts10:
Psyched that this got merged! My first contribution to Mastodon! 

I'm assuming you've tested this on a few browsers and it looks good? I'm only worried cuz I have yet to get Mastodon to run locally (due to unrelated Ruby gem problems on my home machine), and I can't quite produce it via Chrome's Inspect Element.

nightpool:
yeah i think this broke scrollbars on chrome

sts10:
Ack, that's what I was worried about. 

Maybe we should just change the hard-coded `8px` to `12px`, either in basics.scss or reset.scss, rather than removing the specification altogether.

PR #7061, opened 2018-04-07
Free storage if it is exceeding disk quota

akihikodaki:


app/javascript/mastodon/actions/statuses.js
```diff
@@ -92,12 +92,17 @@ export function fetchStatus(id) {
...
-    asyncDB.then(db => {
+    openDB().then(db => {
       const transaction = db.transaction(['accounts', 'statuses'], 'read');
       const accountIndex = transaction.objectStore('accounts').index('id');
       const index = transaction.objectStore('statuses').index('id');
 
-      return getFromDB(dispatch, getState, accountIndex, index, id);
+      return getFromDB(dispatch, getState, accountIndex, index, id).then(() => {
+        db.close();
+      }, error => {
```
Gargron: Is the `then(callback, errorCallback)` syntax preferred over `then(callback).catch(callback)`?

Gargron (approved):
I don't really understand this code because IndexedDB and Service Workers are beyond my expertise, so I would appreciate more reviews by other people, but it looks fine to me (with the exception of some code style inconsistencies)

app/javascript/mastodon/storage/modifier.js
```diff
@@ -166,3 +179,27 @@ export function putStatuses(records) {
     .then(evictStatusesByRecords)
     .catch(printErrorIfAvailable);
 }
+
+export function freeStorage() {
+  return navigator.storage.estimate().then(({ quota, usage }) => {
+    if (usage + 8388608 < quota) {
```
nightpool: what does 8388608 represent here?

Gargron: I understood it as arbitrary padding to make sure the quota isn't even *close* to being exceeded, but I suppose for readability it would be better to extract this value into a constant with an appropriate name.

app/javascript/mastodon/actions/statuses.js
akihikodaki: No. It is a semantic pattern rather than a syntax. We see `then(onresolution).catch(onrejection)` even more often than `then(onresolution, onrejection)` because the former can catch errors in `onresolution` while the latter cannot.

In this case, however, the two callbacks do the same thing and catching an error of a callback with the other one does not make sense.

akihikodaki:
f93cec9f07df5ad33f0a5f67b38a10c75b9a3844 assigns the constant value to `storageMargin`. (https://github.com/tootsuite/mastodon/pull/7061#discussion_r179928460)

PR #7062, opened 2018-04-07
Feature: Improve remote report display

ThisIsMissEm:
Resolves https://github.com/tootsuite/mastodon/issues/6994

screen shot 2018-04-07 at 10 10 09 am

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -32,6 +32,14 @@ class Report < ApplicationRecord
     :flag
   end
 
+  def from_domain
+    Report.where(account: Account.by_domain(account.domain))
+  end
+
+  def targeting_domain
+    Report.where(target_account: Account.by_domain(account.domain))
+  end
+
   def statuses
     Status.where(id: status_ids).includes(:account, :media_attachments, :mentions)
   end
--- a/app/services/report_service.rb
+++ b/app/services/report_service.rb
@@ -34,7 +34,7 @@ class ReportService < BaseService
   def forward_to_origin!
     ActivityPub::DeliveryWorker.perform_async(
       payload,
-      some_local_account.id,
+      @source_account.representative&.id,
       @target_account.inbox_url
     )
   end
@@ -44,11 +44,7 @@ class ReportService < BaseService
       @report,
       serializer: ActivityPub::FlagSerializer,
       adapter: ActivityPub::Adapter,
-      account: some_local_account
+      account: @source_account.representative
     ).as_json)
-  end
-
-  def some_local_account
-    @some_local_account ||= Account.local.where(suspended: false).first
   end
 end
--- a/app/views/admin/reports/show.html.haml
+++ b/app/views/admin/reports/show.html.haml
@@ -62,9 +62,24 @@
     = render 'authorize_follows/card', account: @report.target_account, admin: true
     = render 'admin/reports/account_details', account: @report.target_account
   .report-accounts__item
-    %h3= t('admin.reports.reported_by')
-    = render 'authorize_follows/card', account: @report.account, admin: true
-    = render 'admin/reports/account_details', account: @report.account
+    - if @report.account.local?
+      %h3= t('admin.reports.reported_by')
+      = render 'authorize_follows/card', account: @report.account, admin: true
+      = render 'admin/reports/account_details', account: @report.account
+    - else
+      %h3= t('admin.reports.reported_by_remote_instance')
+      = render 'authorize_follows/card', account: @report.account, admin: true
+      .table-wrapper
+        %table.table
+          %tbody
+            %tr
+              %td{ colspan: 2 }= t('admin.reports.remote_instance_representative', domain: @report.account.domain)
+            %tr
+              %td= t('admin.reports.created_from_remote_instance')
+              %td= pluralize(@report.from_domain.count, t('admin.accounts.show.report'))
+            %tr
+              %td= t('admin.reports.targeting_remote_instance')
+              %td= pluralize(@report.targeting_domain.count, t('admin.accounts.show.report'))
 
 %h3= t('admin.reports.comment.label')
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -155,6 +155,13 @@ class Account < ApplicationRecord
     "#{username}@#{Rails.configuration.x.local_domain}"
   end
 
+  def representative
+    return nil unless local?
+
+    site_contact = Account.find_local(Setting.site_contact_username) if Setting.site_contact_username.present?
+    site_contact || User.admins.first&.account || User.first&.account
+  end
+
   def to_webfinger_s
     "acct:#{local_username_and_domain}"
   end
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -120,6 +120,7 @@ class Account < ApplicationRecord
   scope :recent, -> { reorder(id: :desc) }
   scope :alphabetic, -> { order(domain: :asc, username: :asc) }
   scope :by_domain_accounts, -> { group(:domain).select(:domain, 'COUNT(*) AS accounts_count').order('accounts_count desc') }
+  scope :by_domain, ->(value) { where(arel_table[:domain].lower.eq(value&.downcase)) }
   scope :matches_username, ->(value) { where(arel_table[:username].matches("#{value}%")) }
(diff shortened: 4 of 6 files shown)


ThibG:
Hmm, this is an improvement, but it still doesn't make clear that the person listed is not necessarily the person who has actually filed the report.

ThisIsMissEm:
@ThibG welcome to suggestions, at least now it's guaranteed to be an admin on the remote server, rather than any random first unsuspended user.

app/services/report_service.rb
```diff
@@ -44,11 +44,11 @@ def payload
...
       adapter: ActivityPub::Adapter,
-      account: some_local_account
+      account: local_admin_account
     ).as_json)
   end
 
-  def some_local_account
-    @some_local_account ||= Account.local.where(suspended: false).first
+  def local_admin_account
+    @local_admin_account ||= User.admins.first.account
```
Gargron: Idea: Why not use "contact account" when it's set? Also idea: Abstract it out into, for example, `Account.representative`... :thinking:

ThisIsMissEm: As in `Settings.site_contact_username` ? 

I didn't know it existed, but we certainly could use that.

ThibG:
@ThisIsMissEm maybe “Report made on remote instance” instead of “Report by remote account”, as the latter suggests that the listed account is the person who made the report?
In a similar fashion, “Reports assigned to this account” rather than “Reports created by this account” in the case of remote toots… but this may also be misleading for other reasons…

ThisIsMissEm:
@ThibG how about "reports created from this instance" / "reports about from this instance" and "Report made on remote instance (admin)"

i.e., the idea is that the user below is the person to contact on the remote instance about the report created on their instance.

ThibG:
“Reports created from this instance” / “Reports about this instance” and “Report made on a remote instance (admin)” sounds good to me. With the small caveat that with the “representative” user selection not quite addressed yet, remote reports might be split across a few different users…

ThibG:
@ThisIsMissEm yeah, I've seen this change and I'm in favor of it. But not everyone has applied this change yet, and the “representative” may change over time, hence the caveat that the “Reports created from this instance” / “Reports about this instance” might be misleading if that points to reports made by/about the representative account and they change over time.

ThisIsMissEm:
There doesn't seem to be a good way to lay this out..

screen shot 2018-04-08 at 3 14 04 pm
screen shot 2018-04-08 at 3 15 08 pm

It'd be perhaps better if the report stats showed next to the account, rather than beneath the account

app/models/report.rb
```diff
@@ -32,6 +32,14 @@ def object_type
     :flag
   end
 
+  def from_domain
+    Report.where(account: Account.by_domain(account.domain))
+  end
+
+  def targeting_domain
+    Report.where(target_account: Account.by_domain(account.domain))
+  end
```
ThibG: I am worried about the performances here, but I'm no Rails expert, so I'll let someone more knowledgeable comment.

app/views/admin/reports/show.html.haml
```diff
@@ -55,10 +55,20 @@
...
+      = render 'authorize_follows/card', account: @report.account, admin: true
+      .table-wrapper
+        %table.table
+          %tbody
+            %tr
+              %td= t('admin.reports.created_from_remote_instance')
+              %td= link_to pluralize(@report.from_domain.count, t('admin.accounts.show.report')), admin_reports_path(account_id: @report.account.id)
+            %tr
+              %td= t('admin.reports.targeting_remote_instance')
+              %td= link_to pluralize(@report.targeting_domain.count, t('admin.accounts.show.report')), admin_reports_path(target_account_id: @report.account.id)
```
ThibG: The links still point to reports attributed to / against the particular account and not reports against / reported by users on the instance.

app/services/report_service.rb
ThisIsMissEm: @Gargron I started doing this, and the I realised: there's absolutely no validation on the Setting.site_contact_username value — you can set it to pretty much anything.

Interestingly the about page doesn't show the correct user is site_contact_username is a remote user, despite this being accepted in settings.

ThisIsMissEm: So, we'd end up with probably this:

```ruby
  def representative
    return nil unless local?

    site_contact = Account.find_local(Setting.site_contact_username) if Setting.site_contact_username.present?
    site_contact || User.admins.first.account
  end
```

If we're to follow the lead from InstancePresenter, then:
```ruby
  def representative
    return nil unless local?

    site_contact = Account.find_local(Setting.site_contact_username)
    site_contact || User.admins.first.account
  end
```

Which relies on the fact that username can't be an empty string

app/models/report.rb
ThisIsMissEm: Yeah, it'd probably be better to include `domain` on the Report object, such that we could just do `Report.where(target_domain: domain).count` and `Report.where(domain: domain).count`

But this would require database changes, and be backwards incompatible.

app/views/admin/reports/show.html.haml
ThisIsMissEm: Good catch.. this would require changes to ReportFilter to support domains, so you'd end up with a url like `/admin/reports?domain=switter.at` or `/admin/reports?targeted_domain=switter.at`

I'm not sure if that's a good idea.

app/models/report.rb
Gargron: This becomes a `SELECT ... WHERE id IN (SELECT ... )` query which is fine, it's not like all accounts are loaded into Rails before being passed into the where clause

app/views/admin/reports/show.html.haml
ThisIsMissEm: @Gargron would there be any support for a "domains" model which was just the string of the domain + an ID which we could use for joins? It feels a little odd to use the domain in the filter URL — alternatively, I could do the hash of the domain, as prevent the raw domain being in the URL.

Gargron:
@ThisIsMissEm To be honest I would prefer if this PR contained the least possible amount of changes, aka just the display and the some_local_account change. I don't think there's a pressing need for new filtering functionality

ThisIsMissEm:
@Gargron I would agree — I can just remove the links here, as that'd probably make most sense.

ThisIsMissEm:
Having an issue running the tests, something about rPam2 missing, yet bundler says everything is installed correctly?

```
rspec spec/models/account_spec.rb
                                                                                                                                                                
An error occurred while loading ./spec/models/account_spec.rb.
Failure/Error: require File.expand_path('../../config/environment', __FILE__)

NameError:
  uninitialized constant Rpam2
# ./config/environments/test.rb:65:in `<top (required)>'
# ./config/environment.rb:5:in `<top (required)>'
# ./spec/rails_helper.rb:2:in `<top (required)>'
# ./spec/models/account_spec.rb:1:in `<top (required)>'
# ------------------
# --- Caused by: ---
# NameError:
#   uninitialized constant Rpam2
#   ./config/environments/test.rb:65:in `<top (required)>'
No examples found.                                                                                                                                              
 0/0 |================================================================= 100 ==================================================================>| Time: 00:00:00 
[rspec-sidekiq] WARNING! Sidekiq will *NOT* process jobs in this environment. See https://github.com/philostler/rspec-sidekiq/wiki/FAQ-&-Troubleshooting

Finished in 0.00099 seconds (files took 3.04 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
Coverage report generated for RSpec to /mastodon/coverage. 65 / 16042 LOC (0.41%) covered.
```

Gargron:
I think I want to make the PAM test suite optional because it's really annoying, but anyway to make it work currently you need to `bundle install --with pam_authentication`

ThisIsMissEm:
@Gargron okay, I managed to get the tests running, but I still have like, 30 failures locally. I don't think their related to anything I've done, a lot are about this [new rails 6 issue](https://github.com/rspec/rspec-rails/issues/1857) then there are issues with the URLs mismatching.

I use .dotenv locally whenever I switch directories to load up the correct .env configuration, so I'm guessing that's something to do with it.

ThisIsMissEm:
Here's the updated UI / Display: 
screen shot 2018-04-18 at 3 35 26 am

ThisIsMissEm:
Closing this as it now is obsolete / conflicts heavily due to #7188

PR #7067, opened 2018-04-07
Prevent admins and moderators eavesdropping in private and direct toots

### LINKED ISSUE #6986: Prevent moderators from accessing private/direct toots that haven't been flagged
There have been much discussion recently about admins/moderators being able to read private/direct toots from the admin panels.

There are many issues with this feature, including:
1. Moderators not knowing whether a toot is public/unlisted/private/direct (#6972)
2. Moderators/admins seeing private/direct toots without opting in (if they should be able to access them at all, this should be an extra, conscious step)
3. Moderators being able to access unflagged private/direct toot

This issue is about 3., which subsumes 2.

I think it is an issue that admins can access unflagged private/direct messages. Offending messages can be flagged by users even if they are private or direct, and even if the recipient is on a remote instance since last release.

This change has already been implemented on a few instances, by merging the following commit: https://github.com/Aldarone/mastodon/commit/2700d4744d12749c2ef3d3dd174c16fe330ccc4d

* * * *

Aldarone:
Fix #6986

### DIFF
--- a/app/controllers/admin/statuses_controller.rb
+++ b/app/controllers/admin/statuses_controller.rb
@@ -12,7 +12,7 @@ module Admin
     def index
       authorize :status, :index?
 
-      @statuses = @account.statuses
+      @statuses = @account.statuses.where(visibility: [:public, :unlisted])
 
       if params[:media]
         account_media_status_ids = @account.media_attachments.attached.reorder(nil).select(:status_id).distinct


nynhex:
@Aldarone thanks for this.  I've already pulled this into my fork and am testing.

ClearlyClaire (approved):
Looks good to me

rtucker:
As this is a privacy-critical fix, there probably ought to be test cases for it.

(I could tackle this at some point but am not in a position to do so right now.)

Gargron:
@rtucker This is a change in an admin controller and it's very very simple.

That being said, I feel like there needs to be a way to opt back into looking at DMs, otherwise it's an abuseable blind spot.

rtucker:
It's important enough that it's worth describing as an intentional behavior, on the chance that a future rewrite/refactoring breaks it by accident.

Also, I do believe this will still allow reported messages to be seen regardless of visibility, right? That should be good enough for legit abuse imho

PR #7072, opened 2018-04-08
Add search item to tab bar for mobile devices

ThisIsMissEm:
Presently we on Switter get a lot of confused people asking "how do I search", and it's always from people on mobile devices. On a desktop, it's easy, the search field is right there, but currently, on mobile, you first have to tap "compose" then select the search box.

This PR adds a "search" item to the tab bar for mobile devices, whilst reusing all the existing "search within compose" functionality. The `isSearchPage` is purely to prevent the animations that are normally present here.


Gargron (approved):

ThisIsMissEm:
@Gargron ha, I was just about to go back and fix that missing prop warning. Thanks for shipping it!

PR #7074, opened 2018-04-08
Feature: Allow staff to change user emails

ThisIsMissEm:
Presently on Switter.at, we get a lot of users accidentally signing up with the wrong email address, like, they use `@proton.com` instead of `@protonmail.com` which means they never can finish creating their account.

This PR implements the ability for a staff member to find a user and change their email address. The reason it does not issue an `email_changed` email is that the current email address is usually invalid and emailing it would decrease your deliverability stats / increase spam score.

The change is captured in the audit log, with details:

screen shot 2018-04-08 at 2 27 45 pm

The view changes are as follows:
- Account show:
screen shot 2018-04-08 at 2 28 43 pm

- New Change Email screen:
screen shot 2018-04-08 at 2 30 37 pm

### DIFF
--- a/app/models/admin/action_log.rb
+++ b/app/models/admin/action_log.rb
@@ -35,6 +35,11 @@ class Admin::ActionLog < ApplicationRecord
       self.recorded_changes = target.attributes
     when :update, :promote, :demote
       self.recorded_changes = target.previous_changes
+    when :change_email
+      self.recorded_changes = ActiveSupport::HashWithIndifferentAccess.new(
+        email: [target.email, nil],
+        unconfirmed_email: [nil, target.unconfirmed_email]
+      )
     end
   end
 end
--- a/app/controllers/admin/change_emails_controller.rb
+++ b/app/controllers/admin/change_emails_controller.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Admin
+  class ChangeEmailsController < BaseController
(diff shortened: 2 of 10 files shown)


app/models/admin/action_log.rb
```diff
@@ -35,6 +35,11 @@ def set_changes
       self.recorded_changes = target.attributes
     when :update, :promote, :demote
       self.recorded_changes = target.previous_changes
+    when :change_email
+      self.recorded_changes = ActiveSupport::HashWithIndifferentAccess.new(
+        email: [target.email, nil],
+        unconfirmed_email: [nil, target.unconfirmed_email]
+      )
```
ThisIsMissEm: This might be a bit contentious, but essentially in the ActionLog view, you want to see both the current email and what it's been changed to. This isn't captured with `target.previous_changes` as we're not updating the email directly, but rather requiring confirmation of the email update.

ThisIsMissEm:
On https://github.com/tootsuite/mastodon/pull/7074/files#diff-64be43f43a5e15792e0f9d355cecc7a3 — I've not written many tests from scratch before, so let me know if there's something I should do differently.

ThisIsMissEm:
Pinging @Gargron for review (perhaps it's an idea to get a "reviewers bot" setup?

Gargron (approved):

PR #7075, opened 2018-04-08
Allow conversation muting from any toot in the conversation, not only…

ClearlyClaire:
… our own


Gargron:
The muting only affects notifications. How do you explain to people that clicking it outside notifications will only hide it from notifications?

ThibG:
Good point, but it also stands for our own toots. I guess we just need better wording, something like “Mute notifications for this conversation” (but that's a bit long…)

Gargron:
>but it also stands for our own toots

Well, we would *always* receive notifications for our own toots, so it makes sense that you can mute them preemptively from anywhere.

PR #7077, opened 2018-04-08
Hide search from Compose on mobile devices

ThisIsMissEm:
We're (switter.at) presently seeing large numbers of users accidentally tooting what they're trying to search for. This PR hides the search form from the Compose view, now that we have a dedicated "search" tab on mobile.

### DIFF
--- a/app/javascript/mastodon/features/compose/index.js
+++ b/app/javascript/mastodon/features/compose/index.js
@@ -90,7 +90,7 @@ export default class Compose extends React.PureComponent {
       <div className='drawer'>
         {header}
 
-        <SearchContainer />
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Admittedly, once this PR is accepted, we could probably implement this nicer with React-Router, but I want to fly this as a straw-man first, to see if we're happy with the change in functionality.

The general goal of a follow-up pull-request would be to detangle the two (compose and search) and whilst on desktop we'd still have search above the compose box, on mobile, we'd hide it completely behind a search route)

Gargron (approved):

Gargron:
More space for the compose box is fine for me

PR #7089, opened 2018-04-09
Feature: Direct message from Statuses

ThisIsMissEm:
Implements the ability to send the author of a status a direct message from the menu on a Status.

Currently does not carry through the context to the compose form (i.e., it's just like direct messaging from the user profiles), I'm not sure if it should actually do a "direct reply" instead, the user intent is a little vague here.

Also changes the "mention" functionality to default back to "public" if it's on "direct" — this is potentially contentious.

screen shot 2018-04-09 at 3 15 12 pm
screen shot 2018-04-09 at 3 15 00 pm

### DIFF
--- a/app/javascript/mastodon/reducers/compose.js
+++ b/app/javascript/mastodon/reducers/compose.js
@@ -259,16 +259,18 @@ export default function compose(state = initialState, action) {
   case COMPOSE_UPLOAD_PROGRESS:
     return state.set('progress', Math.round((action.loaded / action.total) * 100));
   case COMPOSE_MENTION:
-    return state
-      .update('text', text => `${text}@${action.account.get('acct')} `)
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
+      map.set('focusDate', new Date());
+      map.set('idempotencyKey', uuid());
+    });
   case COMPOSE_DIRECT:
-    return state
-      .update('text', text => `@${action.account.get('acct')} `)
-      .set('privacy', 'direct')
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
+      map.set('privacy', 'direct');
+      map.set('focusDate', new Date());
+      map.set('idempotencyKey', uuid());
+    });
   case COMPOSE_SUGGESTIONS_CLEAR:
     return state.update('suggestions', ImmutableList(), list => list.clear()).set('suggestion_token', null);
   case COMPOSE_SUGGESTIONS_READY:
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -31,6 +31,7 @@ export default class Status extends ImmutablePureComponent {
     onFavourite: PropTypes.func,
     onReblog: PropTypes.func,
     onDelete: PropTypes.func,
+    onMention: PropTypes.func,
     onPin: PropTypes.func,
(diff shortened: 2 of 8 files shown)


app/javascript/mastodon/reducers/compose.js
```diff
@@ -259,10 +259,15 @@ export default function compose(state = initialState, action) {
...
-    return state
-      .update('text', text => `${text}@${action.account.get('acct')} `)
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
+      map.set('focusDate', new Date());
+      map.set('idempotencyKey', uuid());
+
+      if (map.get('privacy') === 'direct') {
```
Gargron: Don't do this. Sometimes I reply to multiple people and instead of typing their names, I would use the dropdown option to mention them. This does not mess with the type of message I'm composing. If it turned my DM into a public message without me noticing/expecting it that would be really bad.

Gargron (changes requested):

app/javascript/mastodon/reducers/compose.js
ThisIsMissEm: Hence the reason I said it might be contentious — wasn't sure if that behaviour was desired.

I'll remove this change, but keep the one for the correct insertion of mentions.

app/javascript/mastodon/reducers/compose.js
```diff
@@ -259,16 +259,18 @@ export default function compose(state = initialState, action) {
   case COMPOSE_UPLOAD_PROGRESS:
     return state.set('progress', Math.round((action.loaded / action.total) * 100));
   case COMPOSE_MENTION:
-    return state
-      .update('text', text => `${text}@${action.account.get('acct')} `)
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
```
ThisIsMissEm: This correctly spaces text in the compose area, previously you'd end up with:
`"Started writing something" -> "Started writing something@gargron "`

Now you'll always get: `"Started writing something @gargron "`

app/javascript/mastodon/reducers/compose.js
```diff
@@ -259,16 +259,18 @@ export default function compose(state = initialState, action) {
...
-      .update('text', text => `@${action.account.get('acct')} `)
-      .set('privacy', 'direct')
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
+      map.set('privacy', 'direct');
+      map.set('focusDate', new Date());
+      map.set('idempotencyKey', uuid());
+    });
```
ThisIsMissEm: Using `withMutations` to reduce changes to the `state` immutable map. See: https://facebook.github.io/immutable-js/docs/#/Map/withMutations

app/javascript/mastodon/reducers/compose.js
```diff
@@ -259,16 +259,18 @@ export default function compose(state = initialState, action) {
...
   case COMPOSE_DIRECT:
-    return state
-      .update('text', text => `@${action.account.get('acct')} `)
-      .set('privacy', 'direct')
-      .set('focusDate', new Date())
-      .set('idempotencyKey', uuid());
+    return state.withMutations(map => {
+      map.update('text', text => [text.trim(), `@${action.account.get('acct')} `].filter((str) => str.length !== 0).join(' '));
```
ThisIsMissEm: This now has the same behaviour as `compose mention`, however, I think we should just ask whether the user wants to replace the text if they're clicking on the menu for "direct message"

ThisIsMissEm:
@Gargron submitted changes 👍

Gargron (approved):

PR #7094, opened 2018-04-09
Implement finding accounts based on hashtags in their "note"

ThisIsMissEm:
I'm not saying this is perfect, but it'd greatly increase the ability to discover accounts you may be interested in following. Still shows the hashtags beneath.

From my reading xpath in postgresql is actually safe, although, it does require the postgresql server to be compiled with `configure --with-libxml` (on my mac this seems to be default) — we could avoid this if we had a note_raw column which stripped the html from the note, then it'd be a simple text search.

screen shot 2018-04-09 at 7 26 56 pm

(That account is using the bio I have on Switter.at)

Additionally, this works with multiple hashtags, and they're AND'ed. Other non-hashtag text is ignored.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -341,6 +341,10 @@ class Account < ApplicationRecord
       end
     end
 
+    def search_by_hashtags(hashtags)
+      where("note <> '' AND (array_remove(xpath('//a[contains(@class, \"hashtag\")]//text()', note::xml)::text[], '#')) @> ARRAY[?]", hashtags)
+    end
+
     private
 
     def generate_query_for_search(terms)
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -15,7 +15,11 @@ class AccountSearchService < BaseService
   private
 
   def search_service_results
-    return [] if query_blank_or_hashtag? || limit < 1
+    return [] if query_blank? || limit < 1
+
+    if query_hashtag?
+      return Account.search_by_hashtags(query_hashtags).limit(limit)
+    end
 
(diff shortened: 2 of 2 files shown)


renatolond:
I tested the query and got an error in my instance
```
ERROR:  could not parse XML document
DETAIL:  line 1: Extra content at the end of the document
```
One user has `<p>`, `</p>` in his description, something like: `<p>a</p><p>b</p>`

ThisIsMissEm:
@renatolond yeah, just encountered that too — I thought mastodon enforced the formatting of the text for the note to be a specific way. Will have to get a "note_text" field with all the HTML stripped by mastodon.

In which case it's probably better just adding a "hashtags" metadata attribute (I have another PR in the works for arbitrary instance based attributes for profiles

Gargron:
I think letting people search accounts by hashtags is a good compromise but I am doubtful about this implementation (its requirements and performance). It could be simpler and faster to just create an accounts-to-hashtags many-to-many join table and relationship

AAaaand I suppose it would have to be done really similarly to statuses in how this stuff federates too, so putting a `tag` array on ActivityPub accounts etc. Judging by many thumbs up on your PR this seems like it's worth doing so if you want I can take it over

ThisIsMissEm:
I might be a little time constrained as I have to do client work to pay some bills, but I’d be open to doing it if someone sponsored the feature.

Otherwise I’ll get to it when I get to it.

I think I am familiar enough with the Mastodon codebase to fully implement this.

ThisIsMissEm:
Closed and will reopen once I've written the many-to-many style code, which is a complete rewrite, and not just an update to this PR.

PR #7103, opened 2018-04-11
Remove use of separate (slower) step to chown /mastodon

danhunsaker:
The `COPY` command supports `chown`ing files automatically, so just have _it_ ensure everything belongs to the correct user.

That means we need to _create_ the user earlier, which is fine.

Also switch to the `mastodon` user earlier in the file so that `bundle` and `yarn` commands are run as that account instead of `root`. This avoids needing to fix permissions on anything generated rather than copied into place.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,10 @@ EXPOSE 3000 4000
 
 WORKDIR /mastodon
 
+RUN addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon \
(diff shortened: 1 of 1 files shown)


nightpool:
see also https://github.com/tootsuite/mastodon/pull/6514 and
https://github.com/tootsuite/mastodon/pull/6662

danhunsaker:
Ah. Old software prevents new software from being built intelligently. Well, fair enough.

Gargron:
Going to have to stick with the slow version because of backwards-compatibility :(

PR #7104, opened 2018-04-11
Redirect to account status page for page of status stream entry

akihikodaki:
Commit 519119f657cf97ec187008a28dba00c1125a9292 missed a change for
stream entry page. Instead of duplicating the change, redirect to account
status page. It would also help crawlers (of search engines, for example)
to understand a stream entry URL and its corresponding status URL points
to the same page.


Gargron (approved):

akihikodaki:
b8b3810390de5d3a0fd4e481d9f4e5498ba0c812 implements the spec.

PR #7109, opened 2018-04-11
Force convert to JPG for preview card thumbnails to avoid animations

### LINKED ISSUE #7093: GIFs in .status-card autoplay regardless of GIF autoplay settings
This is a difficult problem to solve technically but I want it to be tracked because it just occurred and made me nauseous.

* * * *

Gargron:
Fix #7093

### DIFF
--- a/app/models/preview_card.rb
+++ b/app/models/preview_card.rb
@@ -34,7 +34,7 @@ class PreviewCard < ApplicationRecord
 
   has_and_belongs_to_many :statuses
 
-  has_attached_file :image, styles: { original: { geometry: '400x400>', file_geometry_parser: FastGeometryParser } }, convert_options: { all: '-quality 80 -strip' }
+  has_attached_file :image, styles: { original: { geometry: '400x400>', format: 'jpg', file_geometry_parser: FastGeometryParser } }, convert_options: { all: '-quality 80 -strip' }
 
   include Attachmentable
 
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
This is a far simpler solution to this problem that what I'd thought of 😅

app/models/preview_card.rb
```diff
@@ -34,7 +34,7 @@ class PreviewCard < ApplicationRecord
 
   has_and_belongs_to_many :statuses
 
-  has_attached_file :image, styles: { original: { geometry: '400x400>', file_geometry_parser: FastGeometryParser } }, convert_options: { all: '-quality 80 -strip' }
+  has_attached_file :image, styles: { original: { geometry: '400x400>', format: 'jpg', file_geometry_parser: FastGeometryParser } }, convert_options: { all: '-quality 80 -strip' }
```
ThisIsMissEm: Would it make sense to do this similar to account avatars with a lambda proc based on the file?

```
has_attached_file :image, styles: ->(f) { image_styles(f) }, convert_options: { all: '-quality 80 -strip' }

def image_styles(file)
  styles = { original: { geometry: '400x400>', file_geometry_parser: FastGeometryParser } }
  styles.original.format = "png"  if file.content_type == 'image/gif'
  styles
end
```

ykzts (approved):

PR #7117, opened 2018-04-12
[WIP] Fix #7116: allow to specify docker user on run time

### LINKED ISSUE #7116: Docker Compose based setup does not support overriding user
The current docker file hard coded to have the running user as mastodon.mastodon (991.991). The only way to override that is to inject build arguments (`UID` and `GID`) when you're building the image. 

The issue arises after PR #3159, that attempts to fix issue #2680.

This approach is rather non-flexible. You are forced to decide the user id and group id provided by the container. This also make the installation process confusing.

Docker tools actually supports you to inject UID and GID when you run it.

The `docker run` command support a `--user` argument that can inject the running uid / gid into the container. The `docker-compose` config also support `user` with similar purpose. You can totally run the docker image of `web` with the user that actually owns the `public/*` folder. Then uploads would work naturally.

The beauty of this approach is you don't have to deal with the arbitrary `991` ownership at all.

The current `.env.production.sample` already has variable `UID` and `GID` supposedly for that purpose. But the current docker-compose file does not reference them. Thus you'd need to customize your `docker-compose` file yourself and there is no documentation for that.

* * * *

yookoala:
* added `user` key to `web` service of `docker-compose.yml` that reference the `UID` and `GID` in `.env.production` file (as shown in `.env.production.sample`). Allow site administrator to specify the UID and GID to own the public upload folders instead of using the hard-coded `mastodon` (991:991) user.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,6 +50,7 @@ services:
       - ./public/assets:/mastodon/public/assets
       - ./public/packs:/mastodon/public/packs
       - ./public/system:/mastodon/public/system
+      - ./tmp:/mastodon/tmp
 
   streaming:
     build: .
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,6 +36,7 @@ services:
     restart: always
     env_file: .env.production
     command: bundle exec rails s -p 3000 -b '0.0.0.0'
+    user: "${UID:-991}:${GID:-991}"
     networks:
       - external_network
       - internal_network


yookoala:
Note: The UID and GID is set to the "current user" in unix environment. This may be conflicting with existing mastodon installations (although it would be much more intuitive for new installations).

Can change to `WEB_UID` and `WEB_GID` if needed. Please advice.

ykzts:
UID and GID are no longer used.

PR #7125, opened 2018-04-13
Add color variables of texts for better accesibility

lynlynlynx:
Text colors and background colors are completely separated, and ALL parts of mastodon UI gets at least WCAG 2.0 contrast ratio A lavel.
I add some colors on my own judgment, so please do not hesitate to suggest to change them.

Related issue #6192


app/javascript/styles/mastodon/variables.scss
```diff
@@ -26,7 +31,7 @@ $ui-base-color: $classic-base-color !default;                  // Darkest
 $ui-base-lighter-color: lighten($ui-base-color, 26%) !default; // Lighter darkest
 $ui-primary-color: $classic-primary-color !default;            // Lighter
 $ui-secondary-color: $classic-secondary-color !default;        // Lightest
-$ui-highlight-color: $classic-highlight-color !default;        // Vibrant
+$ui-highlight-color: #2b5fd9;
```
Gargron: Why this?

lynlynlynx: This is to acquire contrast ratio between white texts and background as `$ui-highlight-color`,. Without this, the contrast ratio is 3.45 (at least 4.5 is required).
I set `$highlight-text-color` as `$classic-highlight-color` and apply this to all texts, loading bar, and so on, so changes are only background color of buttons with text.

ghost (approved):
Seems really great, thanks for your work!

Gargron:
The buttons colors and icon-buttons colors are different. Icon buttons are altogether purple now.

image

lynlynlynx:
Yes, icon buttons colors are also changed. This is why we must acquire contrast ratio for icon buttons same as texts, so I set new icon buttons colors as `$action-button-color`. But you don't like the new color? Please tell me which you like @Gargron .

Gargron:
>But you don't like the new color?

I don't like visual changes without a lot of discussion beforehand. The web UI has a set color palette that is connected to our logo, it's our "brand colors", and purple is not part of that. I'm okay with slightly adjusting shades to improve contrast, but not changing hues.

lynlynlynx:
Alright, I changed `$action-button-color`.
Current(master) action buttons color; #606984 , C27 M20 Y0 K48
New `$action-button-color`; #8d9ac2 , C27 M20 Y0 K24
How about this?

ws000013

Gargron (approved):

Gargron:
@lynlynlynx I merged this, I'm looking at it, and the buttons (not icon buttons) are still not fixed. I do not like that color at all. It clashes with the background in a disturbing way. And the datetimes/usernames look grey, they don't look like they blend with the background. Overall it started looking like a high-contrast theme. While that's good for people with poor vision (mine isn't 20/20 anymore either) it also takes away from the good looks.

To be honest I thought the main point of this PR was reorganizing variables for easier theming, with small adjustments in contrast. But this doesn't sit well with me. If you want, we can include this as a "high contrast" theme by default?

jebuskrust:
😢 My theme is not very happy with these changes. But should be able to figure it all out

lynlynlynx:
@Gargron  Okay, understood. This theme should be separated as "high contrast" theme for making the best use of  Mastodon's brand colors and that non-impaired people do not hurt their eyes. But please, please understand all items should not be blended each other for visually/color impaired people.  Anyway, I am very grateful that you accept to add high contrast theme as Mastodon's second default theme.

lynlynlynx:
I'm working on separating two themes. It will be done in this weekend. :running_man:

Gargron:
Thank you. By the way, I think you should not use rgba for text colors. I noticed text on white backgrounds was completely invisible because of this, also it doesn't blend with the color scheme, so rather than being a mix of background color + white light, it's a shade of grey.

lynlynlynx:
@Gargron Thanks for your advice. From the viewpoint of what you say, complying brand colors, and sharing codes with the default theme, I changed text colors of high contrast theme to lightened or darkened existing colors.

Issue #1254, opened 2017-04-08
Support Tor-hidden servers


Ekleog:
Some of the publicity given to Mastodon is thanks to the US government asking Twitter for the real identity of one of its users, I think. However, a question rapidly arises: is Mastodon safer than Twitter for this use case? Will an individual holding a server for a small group of people be more efficient than Twitter in suing a government in response, to protect its users?

Unfortunately, I don't think so, as size is required to make governments flex, would it be just to pay legal costs.

In order to better protect the users, a technical solution could be put in place to actually protect server admins from being "blackmailed" (I don't know the actual term for a government forcing one to deliver information) into giving away information about its users. Even the users using Tor is not enough, as private messages could be reached this way, or worse, malicious biometrics-based identification JS could be put on the server to identify the user despite Tor. (even if a special client could be developed to ban the development of such biometrics, the PM issue still exists)

I think this calls for running some mastodon servers as Tor hidden services.

The potential issues we identified are:
1. Not all Mastodon instances run a Tor client.
2. If relaying is put in place, banning a Tor server is much more complex (and it can change identity really easily to escape the ban)

Solutions could be:
1. Change the protocol so that designated Tor bridges relay messages from/to the Tor network. Each Mastodon server would be configured to use one (or more?) such bridge(s), to get the message updates from Tor. As I don't know the details of the protocol, I don't know whether this means the server would have to trust the Tor bridge relatively to the name of the source server, but I'm not sure it matters much.
2. Banning could be done either at a server level (trusting the name the Tor bridge sends the Mastodon instance), or at the network level (refusing any message incoming from Tor by not using any Tor bridge).

This would most likely require a protocol addition to support Tor bridges, but I'd think it's worth it, for privacy and security reasons.

What do you think about it?

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

transifex3:
any updates here?

Icekhaos:
With the new FOSTA bill, this should now be a requirement, as well as probably *very* thorough encryption of the service.

At least if we have distributed Tor nodes here, then we should be able to keep related things alive.

hcmiya:
I wrote a quick example of instructions of building a Tor instance.
https://gist.github.com/hcmiya/40f3810108c954b3a24017a78844e0b6

hcmiya:
I wrote a quick example of federating with Tor instances from clearnet.
[...]

PR #7134, opened 2018-04-14
HTTP proxy support for outgoing request, manage access to hidden service

hcmiya:
Proxy support enables to serve mastodon with making federation from restricted network. Although Docker is known to have a problem of outgoing IPv6 connection, we can expect that the problem will be improved by the use of HTTP proxy.
By combining Tor and Privoxy, Mastodon can federate with instances in Tor, e.g. http://nq5jmc5rsyo4fiph.onion/ (run by me), http://6cuelujo3prytx3k.onion/ (Pleroma)

Most people do not want to connect to Tor's site, so I made it possible to set by option. Unless set to true, Sidekiq will not accumulate errors related to Tor connection.

Using HTTP proxy is not solution for #1254. 1) It's necessary to disable HTTPS to serve Mastodon in Tor. 2) Only the part derived from mastodon uses HTTP proxy. If Mastodon's Gems communicates with the outside, it could leak information.

You can see the test of federation with tor instance at https://experimental.mastodon.home.js4.in/about


app/lib/request.rb
```diff
@@ -11,9 +11,10 @@ class Request
   def initialize(verb, url, **options)
     @verb    = verb
     @url     = Addressable::URI.parse(url).normalize
-    @options = options.merge(socket_class: Socket)
+    @options = options.merge(socket_class: Socket).merge(Rails.configuration.x.http_client_proxy)
     @headers = {}
 
+    raise Mastodon::HostValidationError, 'blocked access to darknet' if !Rails.configuration.x.access_to_darknet && /\.(onion|i2p)$/.match(@url.host)
```
Gargron: - Use capital B at the beginning of message
- Extract the if-clause to a method

app/lib/request.rb
```diff
@@ -127,21 +128,24 @@ def body_with_limit(limit = 1.megabyte)
...
-    class << self
-      def open(host, *args)
-        outer_e = nil
-        Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
-          begin
-            raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
-            return super address.ip_address, *args
-          rescue => e
-            outer_e = e
+    if Rails.configuration.x.http_client_proxy.empty?
```
Gargron: Seems like this conditional can be avoided by conditionally setting the `socket_class` option in initialize instead

config/initializers/http_client_proxy.rb
```diff
@@ -0,0 +1,26 @@
+Rails.application.configure do
+  config.x.http_client_proxy = {}
+  ['HTTP_PROXY', 'http_proxy'].each do |name|
+    next if ENV[name].blank?
+    proxy = URI.parse(ENV[name])
+    raise Mastodon::ValidationError, "Unsupported proxy type: #{proxy.scheme}" unless ["http", "https"].include? proxy.scheme
```
Gargron: This error is meant for application start, so no need to use Mastodon::ValidationError. You can just raise an anonymous exception.

config/initializers/http_client_proxy.rb
```diff
@@ -0,0 +1,26 @@
+Rails.application.configure do
+  config.x.http_client_proxy = {}
+  ['HTTP_PROXY', 'http_proxy'].each do |name|
```
Gargron: Why is lowercase variant needed?

.env.production.sample
```diff
@@ -214,3 +214,10 @@ STREAMING_CLUSTER_NUM=1
 # SAML_UID_ATTRIBUTE="urn:oid:0.9.2342.19200300.100.1.1"
 # SAML_ATTRIBUTES_STATEMENTS_VERIFIED=
 # SAML_ATTRIBUTES_STATEMENTS_VERIFIED_EMAIL=
+
+# Use HTTP proxy for outgoing request (optional)
+# http_proxy=http://gateway.local:8118
```
Gargron: Why lowercase name?

.env.production.sample
```diff
@@ -214,3 +214,10 @@ STREAMING_CLUSTER_NUM=1
...
+# Access control for darknet sites.
+# ALLOW_ACCESS_TO_DARKNET=true
```
Gargron: Is there any reason to say "darknet" instead of "tor"? Does this setup support more than just Tor? There are negative connotations with "darknet" so I would prefer another term used for these variables and method names.

Gargron (changes requested):

.env.production.sample
Gargron: From the first link:

>But http_proxy and HTTP_PROXY is treated specially under CGI environment.

It seems like neither should be used. Maybe `HTTP_PROXY_URL` instead, since the value is a URL?

hcmiya: That's the story behind how http_proxy got used. Since Mastodon does not use CGI, it should not be vulnerable. Also, since the setting of proxy by http_proxy is common, there are also many gems that use it (Try `grep -r http_proxy vendor/`). Therefore, I think that it makes sense to use http_proxy for setting.

.env.production.sample
hcmiya: "Darknet" includes not only Tor but also I2P.
I also feel that this name is not good. Should I change it to "SKIP_PRIVATE_ADDRESS_CHECK"?

Gargron: HIDDEN_SERVICE would be an okay alternative wherever DARKNET is used currently

hcmiya: it's nice.

MacLemon:
> 1) It's necessary to disable HTTPS to serve Mastodon in Tor.   

That is only valid when using non-EV certificates. (Which most instances will be using anyway, but there are legal and tried ways to get an .onion name as SAN in a certificate.)

I doubt it's a good idea to support v2 onions at all at this point in time. They're extremely weak (1024bit/RSA, SHA1 shortened to 80bits…) and I wouldn't use them anymore for *new* deployments, which is what we're talking about.
I suggest to support v3 onions only to force everyone who wants to run Mastodon over tor and have it federated, to go with a sane and secure version from the beginning.
v3 onions are supported in the current stable branch of tor.

v3 Onions are configured just like v2 onions:
``` shell
HiddenServiceDir /var/db/tor/<youronionv3servicename>/
HiddenServiceVersion 3
HiddenServicePort 80 127.0.0.1:80
```
The difference is that they're cryptographically secure unlike v2 services.

Since running Onion services has a lot of further implications, this should be documented well with best practices and many caveats so people who want to support tor to provide additional access/transport protections for their users don't expose internal-only services (like server dashboards, etc.) by accident.

app/lib/request.rb
nightpool: "Instance does not support hidden service connections" would maybe be a better message for admins?

Gargron: You can use `@url.tld` btw

.env.production.sample
majestrate: for the record in pleroma the feature is called "upstream proxy support" and has no references to "darknet" or "hidden service" since that is what the feature really is at its core.

.env.production.sample
Gargron: Okay, http_proxy is fine then

.env.production.sample
yegortimoshenko: "Hidden service" is a former name for that functionality. Tor has switched to calling those Onion services: https://www.torproject.org/docs/onion-services.html.en

config/initializers/http_client_proxy.rb
```diff
@@ -11,11 +11,16 @@
...
+  config.x.access_to_darknet = ENV['ALLOW_ACCESS_TO_DARKNET'] == 'true'
+  config.x.darknet_via_transparent_proxy = ENV['DARKNET_VIA_TRANSPARENT_PROXY'] == 'true'
 end
 
 module Goldfinger
   def self.finger(uri, opts = {})
-    opts = opts.merge(Rails.configuration.x.http_client_proxy)
+    to_darknet = /\.(onion|i2p)(:\d+)?$/.match(uri)
+    raise Mastodon::HostValidationError, 'blocked access to darknet' if !Rails.configuration.x.access_to_darknet && to_darknet
+    opts = opts.merge(Rails.configuration.x.http_client_proxy).merge(ssl: !to_darknet)
```
yegortimoshenko: * Onion services can force HTTPS: https://3g2upl4pq6kufc4m.onion/
* This seems to only allow either Tor or I2P to be used with the instance, not both at the same time.

config/initializers/http_client_proxy.rb
```diff
@@ -11,11 +11,16 @@
...
   end
+
+  config.x.access_to_darknet = ENV['ALLOW_ACCESS_TO_DARKNET'] == 'true'
+  config.x.darknet_via_transparent_proxy = ENV['DARKNET_VIA_TRANSPARENT_PROXY'] == 'true'
 end
 
 module Goldfinger
   def self.finger(uri, opts = {})
-    opts = opts.merge(Rails.configuration.x.http_client_proxy)
+    to_darknet = /\.(onion|i2p)(:\d+)?$/.match(uri)
```
yegortimoshenko: This doesn't seem to be specific to either I2P or Tor. Someone might want to federate with servers on Freenet, or on some mesh network or private network that requires a proxy for access.

It seems that there are two concerns: list of zones for which SSL verification is disabled, and list of zones which are accessed via a proxy.

app/lib/request.rb
```diff
@@ -100,6 +100,14 @@ def http_client
     @http_client ||= HTTP.timeout(:per_operation, timeout).follow(max_hops: 2)
   end
 
+  def use_proxy?
+    Rails.configuration.x.http_client_proxy.present?
+  end
+
+  def block_hidden_service?
+    !Rails.configuration.x.access_to_darknet && ['onion', 'i2p'].any?(@url.tld)
```
Gargron: Forgotten variable

config/initializers/http_client_proxy.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  config.x.access_to_hidden_service = ENV['ALLOW_ACCESS_TO_HIDDEN_SERVICE'] == 'true'
+  config.x.hidden_service_via_transparent_proxy = ENV['HIDDEN_SERVICE_VIA_TRANSPARENT_PROXY'] == 'true'
+end
+
+module Goldfinger
+  def self.finger(uri, opts = {})
+    to_hidden = /\.(onion|i2p)(:\d+)?$/.match(uri)
+    raise Mastodon::HostValidationError, 'Instance does not support hidden service connections' if !Rails.configuration.x.access_to_hidden_service && to_hidden
+    opts = opts.merge(Rails.configuration.x.http_client_proxy).merge(ssl: !to_hidden)
```
yegortimoshenko: Onion services can force HTTPS: https://3g2upl4pq6kufc4m.onion/

yegortimoshenko: Onion services can force HTTPS: https://3g2upl4pq6kufc4m.onion/

config/initializers/http_client_proxy.rb
```diff
@@ -0,0 +1,24 @@
...
+    config.x.http_client_proxy[:proxy] = { proxy_address: host, proxy_port: proxy.port, proxy_username: proxy.user, proxy_password: proxy.password }.compact
+  end
+
+  config.x.access_to_hidden_service = ENV['ALLOW_ACCESS_TO_HIDDEN_SERVICE'] == 'true'
+  config.x.hidden_service_via_transparent_proxy = ENV['HIDDEN_SERVICE_VIA_TRANSPARENT_PROXY'] == 'true'
+end
+
+module Goldfinger
+  def self.finger(uri, opts = {})
+    to_hidden = /\.(onion|i2p)(:\d+)?$/.match(uri)
```
yegortimoshenko: This isn't specific either to I2P or Tor: someone might want to federate with servers on Freenet, or on some mesh network or private network that requires a proxy for access.

It seems that there are two concerns: list of TLDs for which TLS is not enforced, and list of TLDs which are accessed via a proxy.

Gargron (approved):

.env.production.sample
```diff
@@ -214,3 +214,10 @@ STREAMING_CLUSTER_NUM=1
 # SAML_UID_ATTRIBUTE="urn:oid:0.9.2342.19200300.100.1.1"
 # SAML_ATTRIBUTES_STATEMENTS_VERIFIED=
 # SAML_ATTRIBUTES_STATEMENTS_VERIFIED_EMAIL=
+
+# Use HTTP proxy for outgoing request (optional)
+# http_proxy=http://gateway.local:8118
+# Access control for hidden service.
+# ALLOW_ACCESS_TO_HIDDEN_SERVICE=true
```
yegortimoshenko: "Hidden service" is a former name for that functionality. Tor has switched to calling those onion services: https://www.torproject.org/docs/onion-services.html.en

> Using Tor "rendezvous points," other Tor users can connect to these onion services, formerly known as hidden services, each without knowing the other's network identity.

Gargron (changes requested):

app/lib/request.rb
hcmiya: Just an omission, sorry

Gargron (approved):

albjeremias:
is there any guide or tutorial on this? :)

hcmiya:
I will write an article about it in #1254.

PR #7137, opened 2018-04-15
Moderation: Allow moderators to disable the ability for an account to post media

ThisIsMissEm:
The account is still allowed to upload media, however, the creation of the status will not be allowed. Disabling media at upload seems to have a less clear meaning to a user (i.e., it seems like something went wrong).

Unfortunately, this is a breaking change, as not all clients respect the error message sent back (ahem, amaroq)

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -45,6 +45,7 @@
 #  moved_to_account_id     :integer
 #  featured_collection_url :string
 #  fields                  :jsonb
+#  media_uploads_disabled  :boolean          default(FALSE), not null
 #
 
 class Account < ApplicationRecord
@@ -184,6 +185,14 @@ class Account < ApplicationRecord
       user&.disable! if local?
       update!(memorial: true)
     end
+  end
+
+  def disable_media_uploads!
+    update!(media_uploads_disabled: true)
+  end
+
+  def enable_media_uploads!
+    update!(media_uploads_disabled: false)
   end
 
   def keypair
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -2,9 +2,13 @@
 
 module Admin
   class AccountsController < BaseController
-    before_action :set_account, only: [:show, :subscribe, :unsubscribe, :redownload, :remove_avatar, :enable, :disable, :memorialize]
+    before_action :set_account, only: [
+      :show, :subscribe, :unsubscribe, :redownload, :remove_avatar,
+      :enable, :disable, :enable_media_uploads, :disable_media_uploads,
+      :memorialize
+    ]
     before_action :require_remote_account!, only: [:subscribe, :unsubscribe, :redownload]
-    before_action :require_local_account!, only: [:enable, :disable, :memorialize]
+    before_action :require_local_account!, only: [:enable, :disable, :enable_media_uploads, :disable_media_uploads, :memorialize]
 
     def index
       authorize :account, :index?
@@ -47,6 +51,20 @@ module Admin
       authorize @account.user, :disable?
       @account.user.disable!
       log_action :disable, @account.user
+      redirect_to admin_account_path(@account.id)
+    end
+
+    def disable_media_uploads
+      authorize @account, :disable_media_uploads?
+      @account.disable_media_uploads!
+      log_action :disable_media_uploads, @account.user
+      redirect_to admin_account_path(@account.id)
+    end
+
+    def enable_media_uploads
+      authorize @account, :enable_media_uploads?
+      @account.enable_media_uploads!
+      log_action :enable_media_uploads, @account.user
       redirect_to admin_account_path(@account.id)
     end
 
--- a/app/policies/account_policy.rb
+++ b/app/policies/account_policy.rb
@@ -25,6 +25,14 @@ class AccountPolicy < ApplicationPolicy
     staff?
   end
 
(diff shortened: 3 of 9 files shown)


ThisIsMissEm:
This probably needs tests — it's just really late here & I'm not sure how I should properly test this change.

ClearlyClaire (approved):
Looks good to me, but what about profile picture/header image?

ThisIsMissEm:
@ThibG the main goal here is to prevent image spam in the timeline — I see it being more used as a temporary measure to drive home that the admins/moderators can and will stop you posting images if you're being spammy.

For profile pictures, we have #6998, and for headers, we actually need a way for moderators to:
a) see the header image for an account
b) remove the header image for an account

I'll do up that PR now as we're wanting to clamp down on header image policy violations over at Switter.at — we'd previously been leaving it to later, as they're only visible once you click through to a users profile, whereas the avatar is passively visible (it can come up in the federated streams, search, etc.

app/models/account.rb
```diff
@@ -45,6 +45,7 @@
 #  moved_to_account_id     :integer
 #  featured_collection_url :string
 #  fields                  :jsonb
+#  media_uploads_disabled  :boolean          default(FALSE), not null
```
Gargron: Is there a good reason to put this on `accounts` which includes remote people, rather than `users` which stores `admin` and `moderator` booleans?

Also, I am concerned that a single boolean for this one type of permission is too narrow and will have to be changed with a lot of effort somewhere down the line. This is entering a flexible-role territory, the kinda things that usually get stored as either bitmasks, or separate records, or... well, sometimes separate booleans, but I think more care and discussion needs to be had before approving this

Gargron (changes requested):

ThisIsMissEm:
You make a fair point about users vs accounts, and I could definitely update it.

The urgency on this has decreased from our end, as we opted to flat out suspend a few accounts (we then received angry emails about this). 

I think in the interest of shipping it, a simple boolean column defaulting to false (i.e., it can’t ever be made to default to true, by configuration) means that if we do add some kind of bit mask in the future then we can migrate the data; the idea is that this is a temporary measure for admins to use.

The options that mastodon currently has for dealing with users are delete them entirely, or silence them locally (which doesn’t work to control spam in the fediverse)

This is where these temporary measures (block images, block public posts) could come in handy; you block them for a little while, they respond to the moderator email/dm, and if they agree to follow the rules, you unblock, if not you ban/suspend them.

It’s mostly about stopping the immediate problem vs longer term actions.

Gargron:
Okay, makes sense. Yes, I would prefer it on the user model. And perhaps you can define the one about public posts at the same time too? With a consistent naming scheme. I think that maybe starting with the same word is a good idea so you can see at a glance which attributes belong together, e.g. disabled, disabled_media, disabled_public, or whatever else.

ThisIsMissEm:
Okay! Sure!

Would it make sense to have a "Disable Federation" on a per-user basis? Or is this even possible? (perhaps it'd just mean that the federated toots become unlisted leaving the server for federation?)

I'd go with:
- disabled: user cannot login, or perform any action
- disabled_media_statuses: user cannot post statuses containing media
- disabled_public_statuses: user cannot post public statuses (allows them to DM or reply to unlisted)
- disabled_public_federation: user can post public, however, their statuses become unlisted on federation.

ThisIsMissEm:
Oh, and `disabled_public_federation` wouldn't generate an error for the user, it'd just be a silent change.

Gargron:
disable_statuses_media and disable_statuses_public makes sense, I would withhold the federation bit for now because it's very prone to bikeshedding what that should encompass.

nightpool:
Isn't the point with this though that you can now just check
disable_statuses_public for the user and that will work?

Gargron:
(In my opinion, disabling media *should* fail it at the upload stage, not tooting stage. If you go through the trouble of disallowing media uploads, you don't really want them to have the capacity to fill up your media storage (until the garbage collection sweep comes along))

ThisIsMissEm:
@Gargron I did try this, and user tested it, and the media failing at upload was less clear to the user. That said though, I'd like to propose changing how media is uploaded and make it instead at tooting stage rather than ahead of time. This would also allow us to work with File & Canvas APIs to provide image manipulation features purely in the client.

ThisIsMissEm:
I'll open a new PR once I've redone this.

Gargron:
>That said though, I'd like to propose changing how media is uploaded and make it instead at tooting stage rather than ahead of time

One of the advantages of uploading before tooting is that you don't have to wait forever for the toot to send. While the upload time in total is probably the same in both approaches, the perception is different because the user can still type during upload.

ThisIsMissEm:
@Gargron yeah, but it also means if you select the wrong image, it's immediately gone to the internet. Additionally, you can't send whilst the image is uploading, nor can you do any image manipulation client-side.

trwnh:
I don't think pre-uploading or uploading AOT is something that prevents client-side editing of images, but that depends on the workflow. Client apps are free to preprocess any images before uploading them, right? That would best be handled by a PR for the Mastodon Web UI client to implement editing as a feature, but as Gargron said, it would cause more noticeable wait times on slower connections -- alternatively, this can be avoided by doing the processing post-upload and server-side, which would be more computationally expensive.

ThisIsMissEm:
@trwnh yeah, but why upload an image we don't know is going to actually be tooted?

PR #7147, opened 2018-04-16
Make RubyGems for development and test optional

akihikodaki:
Commit 8e88a18316d45a459a31d67487bccc247592d187 makes --with flag
necessary when configuring test. Traditionally, however, Mastodon required
to have --with flag for production configuration. If we have --with flag
for one situation, we can make such a flag unnecessary for the other.

This commit makes --with flag optional for production environment. While
the behavior is contrast with the traditional behavior, I decided to do
so because:
* It is easy to implement as both of group pam_authentication and test will
be optional.
* The production environment should be prioritized than the test
environment.

You can install RubyGems with the following command after applying this
commit:
* For production (removing all --without flags)
bundle install
* For development (adding two --with flags)
bundle install --with development test
* For minimal test (adding one --with flag, removing one --with and one
--without flag)
bundle install --with test --without production

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ services:
 
 install:
   - nvm install
-  - bundle install --path=vendor/bundle --with pam_authentication --without development production --retry=3 --jobs=16
+  - bundle install --path=vendor/bundle --with test --without production --retry=3 --jobs=16
   - yarn install
 
 before_script:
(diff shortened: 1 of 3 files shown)


Gargron:
I do not agree with this approach. It's important to keep conventions in mind. The development-first, production-second approach is standard in Rails applications. In fact, by default, the "development" environment installs production dependencies too, but not the other way around, so it's a lot better as a default.

(However, I believe the PAM test suite should be optional...)

PR #7148, opened 2018-04-16
Paginate descendant statuses in public page

akihikodaki:
See also #7102.


Gargron:
I'd like a review from someone else regarding the UX that this introduces. They're cool changes but I wanna make sure threads don't become less convenient because of them.

Sylvhem:
Can we have some screenshoots of the result posted here?

Gargron:
I wonder if maybe some special exception needs to be made for self-replies. It seems like a self-reply chain should not be paginated, what do you think?

akihikodaki:
We can relax the threshold for self-replies, but the pagination is still desired for performance and security. I wonder what the threshold should exactly be. It should be sufficient for most use cases, including self-replies.

Gargron (approved):

nightpool:
4 is an extremely small limit. Most every reply chain has more then 4 statuses in them.

PR #7150, opened 2018-04-16
Introduce flat layout to contexts reducer

akihikodaki:
This allows to filter out replies in threads even if contexts of those replies are not fetched. (and hopefully makes easier to implement context pagination, which I'm working on.)

### DIFF
--- a/app/javascript/mastodon/actions/timelines.js
+++ b/app/javascript/mastodon/actions/timelines.js
@@ -13,21 +13,9 @@ export const TIMELINE_SCROLL_TOP = 'TIMELINE_SCROLL_TOP';
 
 export const TIMELINE_DISCONNECT = 'TIMELINE_DISCONNECT';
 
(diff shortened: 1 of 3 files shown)


Gargron:
@akihikodaki Speaking of flat contexts, we might get a speed boost out of querying conversations by conversation_id instead of a recursive tree algorithm. Have you thought about that?

akihikodaki:
@Gargron That's right, but still such a performance boost is trivial. We already have a sort of N+1 query to show each statuses of the tree. It will be just 2N instead of N+1 if we adopt the flat layout. It would be problematic if someone has complained about N+1, but in reality nobody cares about that. Twice of trivial queries is still trivial.
By the way, there is another option: N queries instead of 2N. That can be achieved by reusing the results queried in `Status` component. But I didn't make such a change because it does not make sense to add more code if the performance boost is little.

Gargron (approved):

PR #7151, opened 2018-04-16
Fix scrolling behavior

ClearlyClaire:
Update React.JS version to make use of new lifecycle methods, make scrolling position update more robust, should fix some “TL jump” issues associated with folding/unfolding CWs or resizing the window, but not those related to dynamic height of `MediaGallery`.


Gargron (approved):

ThisIsMissEm:
@ThibG Would it perhaps make sense to find the element at the top of the visible scroll region, and then scroll-to that element? See: https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementFromPoint

That is, the scroll height could change if statuses higher up were deleted; the goal is to find the element you're currently scrolled to, and keep that scrolled in view.

Gargron:
>'prevState' is defined but never used.

ThibG:
@Gargron eh, you want me to remove that argument? I'm not too familiar with Javascript, but just removing it although it's part of the API doesn't sit well with me…

Gargron:
Trust me that you can do it. JS callers don't care about your function definition. 0 or 100 arguments, all the same to them. It only matters which arguments you want to use inside your code.

package.json
```diff
@@ -83,8 +83,8 @@
     "prop-types": "^15.5.10",
     "punycode": "^2.1.0",
     "rails-ujs": "^5.1.2",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0",
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0",
```
ykzts: Should update yarn.lock.

PR #7160, opened 2018-04-16
Store URIs of follows, follow requests and blocks for ActivityPub

Gargron:
Same URI passed between follow request and follow, since they are the same thing in ActivityPub. Local URIs are generated during creation using UUIDs and are passed to serializers.

The URIs look like: `https://example.com/payloads/55e21949-a286-4668-8d23-74c7089bea13` They do not resolve to anything in this PR. Maybe it should be a fragment URI instead? :thinking: 

Resolve #5649

### DIFF
--- a/app/lib/activitypub/activity/block.rb
+++ b/app/lib/activitypub/activity/block.rb
@@ -7,6 +7,6 @@ class ActivityPub::Activity::Block < ActivityPub::Activity
     return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.blocking?(target_account)
 
     UnfollowService.new.call(target_account, @account) if target_account.following?(@account)
-    @account.block!(target_account)
+    @account.block!(target_account, uri: @json['id'])
   end
 end
--- a/app/lib/activitypub/activity/follow.rb
+++ b/app/lib/activitypub/activity/follow.rb
@@ -12,7 +12,7 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
       return
     end
 
-    follow_request = FollowRequest.create!(account: @account, target_account: target_account)
+    follow_request = FollowRequest.create!(account: @account, target_account: target_account, uri: @json['id'])
 
     if target_account.locked?
       NotifyService.new.call(target_account, follow_request)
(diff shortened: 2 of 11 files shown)


Gargron:
@ThibG The thing is, to make them resolveable, we suddenly need some index table that maps these uuids to the model tables, or worse yet, has to store the actual JSON permanently. That approach is good in some ways but it's also a looooooot of new storage overhead suddenly, I'm not sure I'm okay with that.

It's important to keep a perspective on what this PR is replacing and fixing.

ThibG:
@Gargron hm, why would the actual JSON need to be permanently stored?

Gargron:
I don't remember now, it's been 2 weeks. But at the very least, there would have to be a table tying the payloads UUIDs back to other records. So either a polymorphic association, or a JSONB column storing the output. At that point JSONB might be simpler to deal with because no on-the-fly rendering and no need to worry about missing DB-level integrity constraints. Buuuuuut that's also a can of worms in terms of not having that become a mess of different implementations (given that everything is on-the-fly right now).

The main complaint of implementers like @lambadalambda is that Mastodon does not pass the same URI *back* as it was given, which is the *primary* fix of this PR. The locally generated URIs go from being unresolveable because they're fragment URIs, to being unresolveable because it's a fake route.

ThibG:
Right, I understand why you've written this PR, I'm just wondering how much work would be needed to make the objects resolvable.

I think you can bypass the need for polymorphic associations and stuff simply by putting follow requests to their own namespace, e.g., `base_url/follow_requests/UUID`

ThibG:
To clarify, I'm fine with follow requests not being resolvable, at least for now. But I think this is something we should think about right now before this becomes a pain in the ass to change later if we end up needing them to be resolvable. That is why I would suggest the follow requests to be in their own namespace, so that we could use the usual routing mechanisms to find them and render them.

Gargron:
Yes, I can see that being a solution, I'm just a bit unsure about whether encoding the type as part of the URL is good design...

Gargron:
I thought about this again. There's very little value in ever fetching these via URIs, and i don't see mastodon or another software ever doing that. Therefore, I think it's OK to use these URIs as-is, and it will not be a problem if the URI structure has to be changed later. @ThibG

ClearlyClaire (approved):

PR #7171, opened 2018-04-17
Message content overflow

lanodan:
``20ch`` and ``30ch`` are arbitrary values.
I have a media-query version as a user-style lying somewhere that I could adapt.

This request was originally done for the Mastodon-FE fork of pleroma: https://git.pleroma.social/tyge/mastofe/merge_requests/1

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -644,6 +644,7 @@
   overflow: hidden;
   white-space: pre-wrap;
   padding-top: 2px;
+  overflow-y: auto;
 
   &:focus {
     outline: 0;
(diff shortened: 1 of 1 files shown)


Gargron:
ch:

>Represents the width, or more precisely the advance measure, of the glyph "0" (zero, the Unicode character U+0030) in the element's font.

max-height based on character width?

ex:

>Represents the x-height of the element's font. On fonts with the "x" letter, this is generally the height of lowercase letters in the font; 1ex ≈ 0.5em in many fonts.

app/javascript/styles/mastodon/components.scss
```diff
@@ -716,6 +717,14 @@
   }
 }
 
+.reply-indicator__content {
+  max-height: 9ex;
+}
+
+.status__content {
+  max-height: 17ex;
+}
```
ThibG: Wait, this would apply to *any* status?

ClearlyClaire (commented):
I'm not a fan of having so many scrollbars, but I can see the issue with the reply indicator…
(Although it's much better nowadays with proper placement of the dropdown menus, and the ability to scroll in the compose column itself)

PR #7180, opened 2018-04-18
Fix comparing id

abcang:
If the status of ID which is not snowflake is included, the order of the timeline is incorrect. I fixed the comparison of string id.


ykzts (approved):

Gargron (approved):

jebuskrust:
Had to revert this commit, going back and forth between timelines causes the first status to be duplicated on the feeds.

Gargron:
Ah so this is the culprit?

jebuskrust:
Yessir

PR #7185, opened 2018-04-18
Fix: Use "welches" instead of "dass" in translation

matthiasbeyer:
Reported by https://metalhead.club/@thomas/99881521526619858

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -428,7 +428,7 @@ de:
     archive_takeout:
       date: Datum
       download: Dein Archiv herunterladen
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
As I'm not a linguist, I would like to have some tumbs-up from another german. Just to be sure everything is correct here.

ykzts:
cc @jumoru (#7186) @koyuawsmbrtn (3eb1f52e5ab8817cad2e087c27122a8d4e1b3ef3)

ykzts (approved):

PR #7186, opened 2018-04-18
Fix: Use "exportierten" instead of "exportieren" in translation

jumoru:
Spotted when looking at https://metalhead.club/@thomas/99881521526619858


ykzts:
duplicate with #7185

koyuawsmbrtn:
I would re-open this pull request since this issue is a different one than #7185 (not a duplicate!)

ykzts:
Just with a different approach, we are solving the same problem. This is duplication.

Gargron (approved):

PR #7188, opened 2018-04-19
Improve report layout

### LINKED ISSUE #7128: Proposal: Merge display of report notes & actions
Thanks to the help of J at AssemblyFour, I now know how to achieve this 

```ruby
# rename def history to actions

def history
  @history ||= (notes.latest + actions).sort_by(&:created_at)
end
``` 

Then it’s just a matter of designing it to look good.

screen shot 2018-04-13 at 11 58 36 pm

* * * *

Gargron:
- Display reported statuses in table format, omitting redundant information and showcasing visibility level and sensitivity status
- Get rid of individual delete/nsfw buttons in favour of always using the batch method in the table header to reduce code complexity
- Unify reported account/reported by account into one table at the top
- Fix #7128 (Merge report notes and action log) (Close #7175)

Screenshots:

image
image

image

image
image


nightpool:
honestly having the bio in the report interface is pretty good for judging what's going on at a glance, but i can see if you want to omit it

nightpool:
without a clear boundary, the stats for each account blend together:

image

why aren't these horizontal

Gargron:
I think I made the top table worse. Should go back to having those in a separate table, and have the top table only have rows of the same height...

ThisIsMissEm:
@nightpool said:
> honestly having the bio in the report interface is pretty good for judging what's going on at a glance, but i can see if you want to omit it

I agree, the bio can be very useful here, is there perhaps a nice way we can display it?

Gargron:
>I'd also like to drop the action buttons at the top of this screen — in our experience they're rarely used, and are better to be used with a Note (either on the account or report)

I am surprised to hear that, I'm going to ask around. I'm not a big fan of how they look though, indeed

ThisIsMissEm:
> On 19 Apr 2018, at 6:42 pm, Eugen Rochko <notifications@github.com> wrote:
> 
> I'd also like to drop the action buttons at the top of this screen — in our experience they're rarely used, and are better to be used with a Note (either on the account or report)
> 
> I am surprised to hear that, I'm going to ask around. I'm not a big fan of how they look though, indeed
> 
Yeah, usually we need to look at the account, and see why we need to suspend it, so we end up using the buttons there. Also, given the destructive nature of suspend, we only rarely use that as a feature (e.g., only in cases of spam, under-age, or posting absolutely illegal content (child porn, violent nationalism, advocating violence, supporting terrorism, etc).

I think all actions should definitely require a note to be created for why you've taken the action you've taken. I'm thinking it'd be an idea to merge Report Notes with Account Notes, and just have a Moderation Notes table, then in the activity log, we can link to the appropriate action that generated the log.

Gargron:
WIP label removed, screenshots updated.

ykzts (approved):

PR #7195, opened 2018-04-20
Add extract_foreign_key_action to Mastodon::MigrationHelpers

### LINKED ISSUE #7119: extract_foreign_key_action not available in ActiveRecord 5.2.0
Some of the initial database migrations rely on a helper function that [calls](https://github.com/tootsuite/mastodon/blob/b21db9bbde3784556a5861e2cdec0c8c2184c72e/lib/mastodon/migration_helpers.rb#L614) [`extract_foreign_key_action`](https://apidock.com/rails/v4.2.1/ActiveRecord/ConnectionAdapters/PostgreSQL/SchemaStatements/extract_foreign_key_action). With the recent release of ActiveRecord 5.2.0, the `extract_foreign_key_action` method has been removed.

I think for the time being a quick pin down to ActiveRecord 5.1.6 would be appropriate, until the migration helper can be cleaned up for future versions of ActiveRecord

* * * *

- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this). (I am a user... so I'm worrying about this I guess? I am running master)

ykzts:
Resolve #7119

ref https://github.com/rails/rails/pull/28052

```console
$ DB_NAME=mastodon_development2 ./bin/rails db:create db:migrate
```

### DIFF
--- a/lib/mastodon/migration_helpers.rb
+++ b/lib/mastodon/migration_helpers.rb
@@ -985,6 +985,17 @@ into similar problems in the future (e.g. when new tables are created).
         BackgroundMigrationWorker.perform_in(delay_interval * index, job_class_name, [start_id, end_id])
       end
     end
+
+    private
+
+    # https://github.com/rails/rails/blob/v5.2.0/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L678-L684
+    def extract_foreign_key_action(specifier)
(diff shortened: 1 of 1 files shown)


ykzts:
cc @azah

Gargron (approved):

PR #4514, opened 2017-08-04
Direct messages column

KScl:
Adds a column that shows all direct messages you've ever sent or received. Streaming has been tested and appears to be working fine on the servers I have it running on.

Column screenshot

Currently, the column link is only on Getting Started because I thought the tabs bar was getting cramped *enough* as it is, but that's easy enough to change.

PR #7200, opened 2018-04-20
Fix #6157: boosting own private toots

mayaeh:
Fix that #6157 did not work.
Update Japanese translations for #4514, 6157, #7178 .


Gargron (approved):

Gargron:
Please update from master so the CI works!

mayaeh:
Understood. I'll try.

PR #7202, opened 2018-04-20
Fix the hot key (j, k) does not function correctly when there is a pinned toot in account timeline.

TakesxiSximada:
In the case where there is a toot pinned to the account timeline, the hot key (j, k) was not functioning properly because the index of the array of the toot is shifted by the number of pinned toot.

When computing the index, checked whether it was a pinned tote or not, and get index from the array of pinned tote  if this tote is pinned. If not, get index from array of normal toot and sum pinned toot count.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -115,11 +115,11 @@ export default class Status extends ImmutablePureComponent {
   }
 
   handleHotkeyMoveUp = e => {
-    this.props.onMoveUp(this.props.status.get('id'), e.target.getAttribute('featured'));
+    this.props.onMoveUp(this.props.status.get('id'), e.target.getAttribute('data-featured'));
   }
 
   handleHotkeyMoveDown = e => {
-    this.props.onMoveDown(this.props.status.get('id'), e.target.getAttribute('featured'));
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/components/status.js
```diff
@@ -233,7 +233,7 @@ export default class Status extends ImmutablePureComponent {
 
     return (
       <HotKeys handlers={handlers}>
-        <div className={classNames('status__wrapper', `status__wrapper-${status.get('visibility')}`, { focusable: !this.props.muted })} tabIndex={this.props.muted ? null : 0}>
+        <div className={classNames('status__wrapper', `status__wrapper-${status.get('visibility')}`, { focusable: !this.props.muted })} tabIndex={this.props.muted ? null : 0} featured={featured ? 'true' : null}>
```
ykzts: `featured={...}` -> `data-featured={...}`

TakesxiSximada: okay, fixed on https://github.com/tootsuite/mastodon/pull/7202/commits/8def64f529c7aef5e72e3a9b48a1644575ff7ce8

ykzts (approved):

PR #7207, opened 2018-04-20
Do not set emoji as inline-block (fixes #5743)

### LINKED ISSUE #5743: Custom emoji shortcodes should be displayed inline if images don’t load
When a user has images disabled in their browser, toots should display the full short code for any emoji in the post. Currently, emoji are entirely useless for those folks.

I can mouseover an emoji to see its code. Please please please fix it so that mouseover text is part of the body of a toot when images are disabled.

screen shot 2017-11-17 at 11 56 37 atm
screen shot 2017-11-17 at 11 58 24 atm

* * * *

ClearlyClaire:


ykzts (approved):

Gargron:
Does this have no downsides?

ThibG:
I wonder. I thought a bit more about it, and I think there might be issues with the size of a rendered status changing as the emoji gets loaded?

unarist:
I think that it was for using fonts to render emojis (like fontawesome) because non-replaced phrasing elements won't use some properties about dimensions. However, img element is a replaced element which uses those properties even without `display: inline-block`.

img element will be [treated as a non-replaced phrasing element if the image is not available](https://html.spec.whatwg.org/multipage/rendering.html#images-3), and ignores `width` property. This behavior is what we expect.

PR #7222, opened 2018-04-21
Ability to specify Redis passwd on mastodon:setup

### LINKED ISSUE #7221: Prompt for Redis password on mastodon:setup
If your redis db requires a password then using the `mastadon:setup` command won't work.

maxolasersquad:
Closes #7221

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -107,9 +107,16 @@ namespace :mastodon do
           q.convert :int
         end
 
+        env['REDIS_PASSWORD'] = prompt.ask('Redis password:') do |q|
+          q.required false
+          a.default nil
+          q.modify :strip
+        end
+
         redis_options = {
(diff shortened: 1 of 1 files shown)


lib/tasks/mastodon.rake
```diff
@@ -107,9 +107,14 @@ namespace :mastodon do
           q.convert :int
         end
 
+        env['REDIS_PASSWORD'] = prompt.ask('Redis password:') do |q|
+          q.required false
+          a.default nil
+          q.modify :strip
```
Gargron: No `end`? :thinking:

Gargron (changes requested):

lib/tasks/mastodon.rake
maxolasersquad: Fixed.

Gargron (approved):

PR #7223, opened 2018-04-21
Resize images before upload in web UI to reduce bandwidth

### LINKED ISSUE #7218: Downsize image in web UI before upload
Especially on mobile devices like iOS, where images may be too large for upload, it would be great to use the canvas API to automatically downsize images before they are passed to the API.

Gargron:
Fix #7218


ykzts (approved):

app/javascript/mastodon/actions/compose.js
```diff
@@ -174,6 +174,79 @@ export function submitComposeFail(error) {
...
+
+  for (let i = 0; i < rawLength; ++i) {
+    uInt8Array[i] = raw.charCodeAt(i);
+  }
+
+  return new Blob([uInt8Array], { type: contentType });
+};
+
+const resizeImage = (inputFile, callback) => {
+  if (inputFile.type.match(/image.*/)) {
```
unarist: Animated gifs will pass this test and it will lost animation.

unarist (changes requested):

app/javascript/mastodon/actions/compose.js
Gargron: But a GIF image larger than 1280px... is there any point in keeping it animated, since it will be a large file size..?

unarist: but if you do so *silently*. they won't be aware of losing animation before posting it.

I think most cases will be covered even if we exclude GIFs from this process.

unarist (approved):

PR #7231, opened 2018-04-22
Remove most behaviour disparities between blocks and mutes

### LINKED ISSUE #5713: If Alice is muted, I still see her posts when Bob boosts them.
I muted user Alice to avoid seeing their posts.
Bob (who I haven't muted) boosted one of Alice's posts.
Aside from blocking Alice (which I'd rather not do), how can I ensure I don't see her posts unless I specifically ask for them?

* * * *

- [?] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

### LINKED ISSUE #7230: Toots from blocked users still appear in the Home timeline when boosted
Some users seem to see toots from people they have blocked if they have been boosted by someone they follow.
For example, on Eldritch Café, one user apparently sometimes sees toots from someone they have blocked and muted when they are boosted by one of their followings. According to [this thread](https://eldritch.cafe/@Sylvhem/99903014689784844), other peoples seems to experience this behavior.
According to [this comment](https://pleroma.site/notice/64474), the problem may be coming from a badly formated SQL query. Applying https://github.com/kaniini/mastodon-hardened/commit/47d8ae216d9e24ca9a760459160149ae73675e04 may help.
I couldn't reproduce this bug myself.

* * * *

Gargron:
The only differences between block and mute should be:

- Mutes can optionally NOT affect notifications
- Mutes should not be visible to the muted

Fix #7230
Fix #5713

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -145,19 +145,20 @@ class FeedManager
     redis.exists("subscribed:#{timeline_id}")
   end
 
+  def blocks_or_mutes?(receiver_id, account_ids, context)
+    Block.where(account_id: receiver_id, target_account_id: account_ids).any? ||
+      (context == :home ? Mute.where(account_id: receiver_id, target_account_id: account_ids).any? : Mute.where(account_id: receiver_id, target_account_id: account_ids, hide_notifications: true).any?)
+  end
+
   def filter_from_home?(status, receiver_id)
     return false if receiver_id == status.account_id
     return true  if status.reply? && (status.in_reply_to_id.nil? || status.in_reply_to_account_id.nil?)
 
-    check_for_mutes = [status.account_id]
-    check_for_mutes.concat([status.reblog.account_id]) if status.reblog?
-
-    return true if Mute.where(account_id: receiver_id, target_account_id: check_for_mutes).any?
-
     check_for_blocks = status.mentions.pluck(:account_id)
+    check_for_blocks.concat([status.account_id])
     check_for_blocks.concat([status.reblog.account_id]) if status.reblog?
 
-    return true if Block.where(account_id: receiver_id, target_account_id: check_for_blocks).any?
+    return true if blocks_or_mutes?(receiver_id, check_for_blocks, :home)
 
     if status.reply? && !status.in_reply_to_account_id.nil?                                                                      # Filter out if it's a reply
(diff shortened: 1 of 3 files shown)


Cassolotl:
Ooo does this also fix #251 by any chance?

trwnh:
related comment: https://github.com/tootsuite/mastodon/issues/7230#issuecomment-383396227

>The only differences between block and mute should be:
>- Mutes can optionally NOT affect notifications
>- Mutes should not be visible to the muted

Are those really the only two differences?

Gargron:
After this PR yes, I believe so.

app/lib/feed_manager.rb
```diff
@@ -145,19 +145,20 @@ def push_update_required?(timeline_id)
     redis.exists("subscribed:#{timeline_id}")
   end
 
+  def blocks_or_mutes?(receiver_id, account_ids, context)
+    Block.where(account_id: receiver_id, target_account_id: account_ids).any? ||
+      (context == :home ? Mute.where(account_id: receiver_id, target_account_id: account_ids).any? : Mute.where(account_id: receiver_id, target_account_id: account_ids, hide_notifications: true).any?)
```
nightpool:
```
Block.where(account_id: receiver_id, target_account_id: account_ids) ||
Mute.where(account_id: receiver_id, target_account_id: account_ids, hide_notifications: context == :notifications)
```

?

Gargron: @nightpool No that will query `WHERE hide_notifications = FALSE` whereas the real query should be `WHERE hide_notifications = ANY` but that's not an indexed column so it's cheaper to omit the clause at all than do it with an `IN` or `OR`

Sylvhem:
I'm really sorry to not be that enthusiastic, but I'm not sure this change will make those functions more useful and more easy to understand :s.  Maybe a discussion is needed in order to understand what those two functions should be exactly?

unarist:
Maybe this won't be applied to the user timeline which also shows replies and boosts like home?

Gargron:
User timelines are not filtered, never have been.

nightpool (approved):
+1 but the current split of authority between NotifyService and FeedManager makes me nervous—it would be very easy for us to add another feed manager call (for example, on backfills) and accidentally skip the NotifyService logic all together.

should maybe be addressed in a followup

PR #7238, opened 2018-04-23
Introduce thread-loader

ykzts:
ref https://webpack.js.org/guides/build-performance/

### DIFF
--- a/config/webpack/loaders/sass.js
+++ b/config/webpack/loaders/sass.js
@@ -6,6 +6,7 @@ module.exports = {
   use: ExtractTextPlugin.extract({
     fallback: 'style-loader',
     use: [
+      'thread-loader',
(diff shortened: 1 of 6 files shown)


config/webpack/loaders/sass.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   use: ExtractTextPlugin.extract({
     fallback: 'style-loader',
     use: [
+      'thread-loader',
```
unarist: https://webpack.js.org/guides/build-performance/#sass
> node-sass has a bug which blocks threads from the Node.js threadpool. When using it with the thread-loader set workerParallelJobs: 2.

Will this not affect to our case?

Gargron:
Closing inactive PR :pray:

ykzts:
I forgot this Pull Request 😔 It may be reopened at a later date.

PR #7245, opened 2018-04-23
Fix fullscreen video player (fixes #7244)

### LINKED ISSUE #7244: Full-screen video player incorrectly scales videos
There are two issues with the full-screen video player incorrectly scaling videos:
- If the fullscreen button is pressed from an inline video, the video will keep the same size, even if it's not the native size nor fills the screen
- This doesn't happen from the modal, but the video still isn't scaled up to fill the screen, and stays in the top-left corner instead

* * * *

ClearlyClaire:


Gargron:
Does this have any side effects? :thinking:

ThibG:
Not to my knowledge. It only affects the fullscreen video player and seem to work well with any aspect ratio, and both with videos smaller and larger than screen size.

Gargron (approved):

Issue #7243, opened 2018-04-23
Bug: Suspended accounts show up identically to Locked/Private accounts


ThisIsMissEm:
We’ve currently an issue where a few users who have been suspended keep beibg reported because the Web UI shows suspended accounts identically to locked/private accounts.

I’d say via API these should return a 404, with status suspended in the JSON body, then in the web UI we can render them differently.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Gargron:
I thought they give a 410 error?

ThisIsMissEm:
Here's how they appear in the Web UI: 
<img width="336" alt="screen shot 2018-04-23 at 7 02 27 pm" src="https://user-images.githubusercontent.com/30827/39141735-e9cd49b4-4728-11e8-8399-d5f87f44f69e.png">

The direct page (i.e., `https://switter.at/@feds` gives an error)

ThisIsMissEm:
Looks like AccountSearchService didn't filter out suspended accounts, working on a patch for this now.

We should also have UI for showing suspended accounts, in case you click on a Username and find them suspended.

Gargron:
Ah, Account#search_for and advanced_search_for contain a suspended=false clause, but exact matches in AccountSearchService don't. Is that it?

Also wow, apparently the REST API does *not* return a 410 when an account is suspended! That completely slipped my mind.

PR #7246, opened 2018-04-23
Prevent suspended accounts from appearing in AccountSearchService

ThisIsMissEm:
Partial fix for #7243 — this prevents you from being able to search/find suspended accounts, but doesn't do anything if you directly go to that account in the web UI (e.g., open in web or from a toot that's cached in a timeline).

There should probably be a change to relationships/statuses to make those calls fail for suspended accounts.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -117,6 +117,7 @@ class Account < ApplicationRecord
   scope :partitioned, -> { order(Arel.sql('row_number() over (partition by domain)')) }
   scope :silenced, -> { where(silenced: true) }
   scope :suspended, -> { where(suspended: true) }
+  scope :without_suspended, -> { where(suspended: false) }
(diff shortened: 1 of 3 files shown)


app/models/account.rb
```diff
@@ -117,6 +117,7 @@ class Account < ApplicationRecord
   scope :partitioned, -> { order(Arel.sql('row_number() over (partition by domain)')) }
   scope :silenced, -> { where(silenced: true) }
   scope :suspended, -> { where(suspended: true) }
+  scope :without_suspended, -> { where.not(suspended: true) }
```
Gargron: `where(suspended: false)` :grin:

ThisIsMissEm: Checked, and fixed.

Gargron (approved):

PR #7247, opened 2018-04-23
WIP: Mark reports targeting the account as resolved when suspending

ThisIsMissEm:
For some reason, I can't get the tests to work correctly — I suspect it's a persistence issue, but I'm not sure.

### DIFF
--- a/app/services/suspend_account_service.rb
+++ b/app/services/suspend_account_service.rb
@@ -8,6 +8,7 @@ class SuspendAccountService < BaseService
     purge_user!
     purge_profile!
     purge_content!
+    resolve_reports!
     unsubscribe_push_subscribers!
   end
 
@@ -47,6 +48,10 @@ class SuspendAccountService < BaseService
     @account.avatar.destroy
     @account.header.destroy
     @account.save!
+  end
+
+  def resolve_reports!
+    @account.targeted_reports.in_batches.update_all(action_taken: true)
   end
 
   def unsubscribe_push_subscribers!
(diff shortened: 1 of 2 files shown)


Gargron:
Hmm. I was sure that when I did suspend/silence from a report, it resolved all reports about the same account already.

Gargron:
Ohhhh, this is for when you suspend *outside* of a report. Hm. Is it still WIP? Tests seem to have passed.

ThisIsMissEm:
@Gargron I think I just needed to remove the resolve lines here: https://github.com/tootsuite/mastodon/blob/master/app/controllers/admin/reports_controller.rb#L47-L60

Your prior solution only did it in certain cases. I'd guess on silencing, it shouldn't resolve all.

Perhaps it'd be nice to have a button on admin accounts for "resolve all reports targeting" and "resolve all reports from" — the use case for that is some malicious users try flooding with reports to mask their other bad behaviour.

app/services/suspend_account_service.rb
```diff
@@ -49,6 +50,10 @@ def purge_profile!
     @account.save!
   end
 
+  def resolve_reports!
+    @account.targeted_reports.in_batches.update_all(action_taken: true)
```
ThisIsMissEm: I think this should narrow the scope to only update the unresolved reports, rather than all reports. Also, it has an interesting side effect that we don't know who took the action. I think this is the original reason I left it as wip, @Gargron

ThisIsMissEm:
As a further aside, it'd be really good to get that `process_report` logic out of the controller and into a service, if possible.

Gargron:
Fixed by #9519

PR #7256, opened 2018-04-24
more sane phrasing in 🇸🇪 translation

stemid:
Noticed this phrase "Vem författade" when I first saw the default front page of mastodon and I don't think it's right.

Proper phrase I think would be "Som författat" but I also felt that the word "författa" when compared to "skapat" (created) felt rigid, old fashioned and so formal that it might be off putting. I felt that "Som skapat" fits better. It translates to "Who have created" or "Who have made".

### DIFF
--- a/config/locales/sv.yml
+++ b/config/locales/sv.yml
@@ -9,7 +9,7 @@ sv:
     contact_missing: Inte inställd
     contact_unavailable: N/A
     description_headline: Vad är %{domain}?
-    domain_count_after: annan instans
+    domain_count_after: andra instanser
     domain_count_before: Uppkopplad mot
     extended_description_html: |
       <h3>En bra plats för regler</h3>
--- a/config/locales/sv.yml
+++ b/config/locales/sv.yml
@@ -29,7 +29,7 @@ sv:
     other_instances: Instanslista
     source_code: Källkod
     status_count_after: statusar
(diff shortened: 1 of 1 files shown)


nightpool:
tagging @ErikXXon, who contributed that translation, and @shellkr, who most recently touched this file (on Weblate)

shellkr (approved):
While I do not agree "författat" is neither too formal nor wrong. I mean, it is the correct word for the translation. Created or "skapat" is a different. It do however work in this context. Perhaps "Som skrivit" would have been even better... In my mind they all work. I'll approve this change.

nightpool (approved):

stemid:
@shellkr @nightpool before you approve I just wanted to tack on another small update. Only just noticed this text was also wrong when the number of connected instances showed on my own instance.

"Connected to X other instances" is in swedish "Ansluten till X andra instanser". The current translation "annan instans" refers to just one instance.

shellkr:
@stemid Why don't you join the Swedish group on Weblate (https://weblate.joinmastodon.org)? It would be greatly appreciated if you did. ;)

stemid:
@shellkr absolutely but I think the translation is largely ok. It was just a few annoying issues that stood out because they were visible by default on new instances even for non-logged in users.

PR #7259, opened 2018-04-24
Add RSS feeds for end-users

### LINKED ISSUE #6178: Add RSS feed for tags
It would be wonderfull if I could parse a RSS feed for existing tags, same as for users.
Something like https://instance.mastodon/tags/rss.atom

* * * *

Gargron:
Got a lot of complaints about Atom feeds confusing IFTTT and whatever else, because the Atom feeds contain a lot of extra information for OStatus federation. RSS is meant for end-users instead.

Fix #6178

### DIFF
--- a/app/serializers/rss/account_serializer.rb
+++ b/app/serializers/rss/account_serializer.rb
@@ -1,0 +1,37 @@
+class RSS::AccountSerializer
+  include ActionView::Helpers::NumberHelper
+  include StreamEntriesHelper
+  include RoutingHelper
+
+  def render(account, statuses)
+    builder = RSSBuilder.new
+
+    builder.title("#{display_name(account)} (@#{account.local_username_and_domain})")
+           .description(account_description(account))
+           .link(TagManager.instance.url_for(account))
+           .logo(full_asset_url(asset_pack_path('logo.svg')))
+           .accent_color('2b90d9')
+
+    builder.image(full_asset_url(account.avatar.url(:original))) if account.avatar?
+    builder.cover(full_asset_url(account.header.url(:original))) if account.header?
+
+    statuses.each do |status|
+      builder.item do |item|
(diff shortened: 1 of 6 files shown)


app/serializers/rss/account_serializer.rb
```diff
@@ -0,0 +1,37 @@
+class RSS::AccountSerializer
```
ykzts: `# frozen_string_literal: true`

app/serializers/rss/tag_serializer.rb
```diff
@@ -0,0 +1,35 @@
+class RSS::TagSerializer
```
ykzts: `# frozen_string_literal: true`

app/serializers/rss/account_serializer.rb
```diff
@@ -0,0 +1,37 @@
...
+          item.enclosure(full_asset_url(media.file.url(:original, false)), media.file.content_type, length: media.file.size)
+        end
+      end
+    end
+
+    builder.to_xml
+  end
+
+  def self.render(account, statuses)
+    self.new.render(account, statuses)
```
ykzts: `self.new.render` -> `new.render`

app/serializers/rss/tag_serializer.rb
```diff
@@ -0,0 +1,35 @@
...
+          item.enclosure(full_asset_url(media.file.url(:original, false)), media.file.content_type, length: media.file.size)
+        end
+      end
+    end
+
+    builder.to_xml
+  end
+
+  def self.render(tag, statuses)
+    self.new.render(tag, statuses)
```
ykzts: `self.new.render` -> `new.render`

ykzts (changes requested):

ykzts (approved):

PR #5406, opened 2017-10-15
Swedish translation

ErikXXon:
I added the Swedish translation file for Matodon

PR #7263, opened 2018-04-25
Better phrasing in swedish translation

stemid:
Ok I'm back with some more nitpicks. :) I just started my own node in Swedish and as I'm using it I'm seeing these issues in the translation so excuse the multiple PRs. 

In the backend file config/locales/sv.yml having the word "enhet" there makes for a strange layout like this.

bloodraven-20180425 111030 1524647430

In the frontend file app/javascript/mastodon/locales/sv.json it's not so bad but I felt that "Justera status sekretess" should either be "Justera status-sekretess" or just "Justera sekretess" is enough in Swedish. Writing it with three words is not correct. Even "Justera status av sekretess" would be more correct.

And the getting_started.open_source_notice value in the same file is just that it sounds better not repeating på twice and also you're bug reporting via Github, at <the link>. So using via instead of på is more appropriate.


ykzts:
cc @ErikXXon (#5406) @shellkr ([Weblate](https://weblate.joinmastodon.org/changes/?lang=sv&project=mastodon))

ykzts (approved):

PR #7286, opened 2018-04-29
Special treatment for statuses that begin with a mention

Gargron:
**TL;DR**: This introduces the behaviour that requires you to prepend your status with a "." when you *do* want to avoid it.

Detect when a status begins with a mention, locally (regex) or via ActivityPub (HTML parsing). Set "in reply to account" information which is used in the feed manager filter (home), and adjust public timelines query to exclude such statuses.

- [ ] Unsolved case: When the first-mention thing happens in a self-reply chain.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -88,7 +88,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     return if tag['href'].blank?
 
     account = account_from_uri(tag['href'])
-    account = ::FetchRemoteAccountService.new.call(tag['href'], id: false) if account.nil?
+    account = ActivityPub::FetchRemoteAccountService.new.call(tag['href'], id: false) if account.nil?
     return if account.nil?
     account.mentions.create(status: status)
   end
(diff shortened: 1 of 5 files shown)


Technowix:
So uh, it's not possible to link public posts anymore?

mal0ki:
I'm raising some concerns regarding this PR

0xa:
My initial concern was about seeing mentions in my home and local timelines, public statuses that are sometimes clearly not meant to be sent to that many people (but not "requiring" a DM). Replies ("conversations") would currently get hidden, but not the initial mention.

I think the potential privacy implications are important and more likely to scare/discourage new users contacting another person they know, than the potential missing reply from the admin (which arguably isn't an issue on smaller instances, but maybe that could be solved in a different way)

As solutions, i can think of:
- showing mentions to only people the *reader* follows, treating mentions like conversations in the home timeline. Also on the public timeline If that's technically reasonable. (maybe a client-side filtering would be enough - it's really more about *how many* people will see it than *who could*)
- adding a light warning, when posting a public/unlisted mention that isn't a reply, that would explain that the effective privacy level won't be affected by the mention but will with replies.

Gargron:
>showing mentions only when mentioning people the reader follows, treating mentions like conversations in the home timeline. Also on the public timeline If that's technically reasonable.

@0xa This is exactly what this PR implements....

0xa:
Well then nice I'm all for it :relaxed: 
It was more a reply to @mal0ki / more talk on the issue

kit-ty-kate:
@Gargron has this PR been closed "for good" or is it going to be reimplemented ?

trwnh:
>Even if it's "expected behaviour" from Twitter, a lot of twitter users don't know that when you add an @ in the beginning it doesn't display for your feed. Which I think is bad. Also, we're not Twitter.

I don't think Twitter even does this anymore. Pretty sure they changed it in 2016. Implemwnting this PR would essentially be implementing pre-2016 Twitter behavior, which was unintuitive but made sense once explained. https://mashable.com/2016/05/24/twitter-mentions-explained/#9SYRcqzAqZqX

There's also perhaps some ambiguity that would be introduced and likely confusion about toots beginning with @ that are now suddenly invisible. I'm assuming this PR would be targeted at people trying to start a conversation and having that be pushed to the public timelines / etc., but I'm off the opinion that the optimal solution would be to suggest making that toot direct when it begins with a mention.

kit-ty-kate:
Can't we add a new "start a conversation" button or something like that to effectively start one and not create a "regular" toot ?
Are there technical problems with this solution ?

mal0ki:
@kit-ty-kate @trwnh Thank you for being patient with me taking some time updating on this issue. 

We decided to instead find a compromise, by adding a mention function to people's profiles instead, so it becomes hidden from the timelines when you go through that. Which isn't this PR. 

From a UX perspective, going to someone's profile to mention them, having that be treated as a thread/reply/conversation, would be the expected behaviour, as we see it. 

We're just going to approach the problem from a different angle. So the problem if seeing random "replies" will get handled, to some extent.

kit-ty-kate:
Awesome ! Thanks a lot @mal0ki <3

PR #7293, opened 2018-04-30
[i18n] Occitan update

Quenty31:
Almost complete, just missing the recent changes in the privacy policy.


Gargron:
>The following data requires normalization:
>config/locales/oc.yml
>Run `i18n-tasks normalize` to fix

Quenty31:
@Gargron yes but I don't know how to do so =/

PR #7295, opened 2018-04-30
Replace Nokogiri with Oga

Gargron:
> Oga is an XML/HTML parser written in Ruby. It provides an easy to use API for
parsing, modifying and querying documents (using XPath expressions). Oga does
not require system libraries such as libxml, making it easier and faster to
install on various platforms. To achieve better performance Oga uses a small,
native extension (C for MRI/Rubinius, Java for JRuby).
>
>-  Low memory footprint
>-  High performance (taking into account most work happens in Ruby)

Benchmarks needed.

Replacing host-meta/webfinger XML generation with Ox reduced time spent in view from 25ms to 0.5ms. (Although pretty much nobody uses host-meta or XML webfinger anymore)


spec/fixtures/requests/koi8-r.txt
```diff
@@ -2,7 +2,7 @@ HTTP/1.1 200 OK
 Server: nginx/1.11.10
 Date: Tue, 04 Jul 2017 16:43:39 GMT
 Content-Type: text/html
-Content-Length: 273
+Content-Length: 384
```
Gargron: @akihikodaki It seems that this value was wrong, because when I checked, the HTML loaded was cut-off abruptly. This caused an error in the koi8-r test suite that was not related to the encoding, but to an unclosed tag. (I think Nokogiri swallowed the error)

Gargron:
```ruby
status = FetchRemoteStatusService.new.call('https://mastodon.art/@Zaquiel/99960724912328628')

require 'benchmark/ips'

Benchmark.ips do |x|
  x.report('oga') { html = Oga.parse_html(status.text); html.css('a') }
  x.report('nokogiri') { html = Nokogiri::HTML(status.text); html.css('a') }
  x.compare!
end
```

Results:

```
Warming up --------------------------------------
                 oga   137.000  i/100ms
            nokogiri   490.000  i/100ms
Calculating -------------------------------------
                 oga      1.366k (± 5.3%) i/s -      6.850k in   5.030721s
            nokogiri      6.264k (±22.2%) i/s -     28.420k in   5.022102s

Comparison:
            nokogiri:     6264.4 i/s
                 oga:     1366.5 i/s - 4.58x  slower
```

Closing this PR :joy:

PR #7301, opened 2018-05-01
Slightly reduce RAM usage

Gargron:
- Add derailed_benchmarks gem for performance benchmarking
- Clean up requires throughout code (no performance change, but sidekiq plugins get required via the Gemfile, so manual require in each worker file has no effect)
- Benchmark showed high number of allocations from ruby-oembed gem. Reimplemented using own service class.
- Do not require TTY gems from Gemfile, require them from rake task. Reduced RAM of non-rake processes by a very small amount.

|Type|Before|After|
|------|-|----|
|At boot time|102.9414 - 108.4727 MiB|97.4727 - 103.6758 MiB|
|During request|87.0625 - 88.375 MiB|84.9531 - 85.1484 MiB|

___
Replaced preloaded JSON-LD contexts (stored as Ruby classes) with a method that hooks into the JSON-LD document loader and caches the response of the context using Rails cache.

Benefits:

- This reduces the boot time RAM by a further ~6 MiB
- Loads *any* JSON-LD context using our own code and timeouts and caches it for quick retrieval
- Technically the memory requirements are the same, but those ~6 MiB are in Redis and only retrieved when needed, and can be garbage-collected after
- Supports changing contexts better than hardcoding them

___

Other unfixed high-memory offenders:

- Chewy (toptal/chewy#644) (21.4531 MiB)
- Charlock Holmes (3.0195 MiB)


brennovich (approved):

nightpool (approved):
+r when tests pass, but we should make sure to keep an eye on the OEmbed stuff during the RC cycle in case this introduces regressions.

PR #7302, opened 2018-05-01
Use Redis-based counter cache for Account#statuses_count

### LINKED ISSUE #828: Direct messages should not increment one's post count

Gargron:
UPDATEs on the accounts table are quite slow. Creating a status increments the count, but the real problem is deleting statuses in batches. UPDATEs for every deleted status bring everything
to a halt. This way, the increments/decrements happen in Redis, and then eventually are written back into the database when the count cache expires and is re-fetched from the database.

Cleaner re-implementation of #6281

Fix #828

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,7 @@ gem 'webpush'
 
 gem 'json-ld', '~> 2.2'
 gem 'rdf-normalize', '~> 0.3'
+gem 'counter_cache-rails', '~> 0.4'
 
 group :development, :test do
   gem 'fabrication', '~> 2.20'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -133,6 +133,8 @@ GEM
     colorize (0.8.1)
     concurrent-ruby (1.0.5)
     connection_pool (2.2.1)
+    counter_cache-rails (0.4.0)
+      rails (>= 5.0)
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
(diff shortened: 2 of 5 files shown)


abcang:
If the statuses_count is referenced after the cache has expired, the count query will be executed. It takes about 4 seconds for users who have 10000 statuses to run count queries on Pawoo when there is no database cache.

```SQL
SELECT COUNT(*) FROM "statuses" WHERE "statuses"."account_id" = xxx AND ("statuses"."visibility" != 3)
```

```
Aggregate  (cost=9007.35..9007.36 rows=1 width=0) (actual time=3983.380..3983.380 rows=1 loops=1)
  ->  Index Only Scan using index_statuses_20180106 on statuses  (cost=0.57..8994.27 rows=5231 width=0) (actual time=0.025..3976.386 rows=12892 loops=1)
        Index Cond: (account_id = xxxx)
        Filter: (visibility <> 3)
        Rows Removed by Filter: 74
        Heap Fetches: 10278
Planning time: 0.063 ms
Execution time: 3983.406 ms
```

If this query is called more than once at a time, the database load will probably be very high. Therefore, I think that it is necessary to have a mechanism to ensure that the query is executing only once even if refered to the statuses_count when the cache is expired.

For example, if there is no cache, there is a method to insert the statuses_count existing in the database into redis first, then reflect the execution result of SQL in redis.

Gargron:
I am sorry for bothering the counter_cache-rails maintainer, but I now believe using Redis is the wrong approach, after all. The negative impact of losing Redis data and/or accidentally running a `COUNT` SQL query is massive. Fixing the underlying issue should go like this:

Adjust BatchedRemoveStatusService to *not* call `#destroy_all`. The only association that needs callbacks to be run is `media_attachments` to delete associated files from storage. Therefore, that association should be queried separately and used with `#destroy_all`, and statuses should use `#delete_all`. That way, counter cache on accounts table will not be touched repeatedly. Since the only use of BatchedRemoveStatusService is account removal, the accuracy of the count after the operation isn't even important.

PR #7306, opened 2018-05-02
Add missing tests for user.rb

shuheiktgw:
I just added some missing tests for `user.rb`!
~~Also, I refactored user.rb a bit by adding `new_user?` method.~~  I removed this commit!

Thank you! 👍


app/models/user.rb
```diff
@@ -173,10 +173,10 @@ def confirm
   end
 
   def confirm!
-    new_user = !confirmed?
-
     skip_confirmation!
     save!
+
+    new_user = !confirmed?
```
Gargron: Why does this change work, if skip already changes this value?

Gargron:
@ykzts FYI I think rspec-retry does not work very cleanly, e.g. foreign key violations in 2nd and 3rd retries because it doesn't reset the database properly? This might confuse some contributors (the original try error is the real error)

Gargron (approved):

ykzts:
dump.rdb?

shuheiktgw:
Whoops, my apology. I'll create a PR to remove it.

PR #7311, opened 2018-05-02
Keep notification when muting_notifications is true

abcang:
Fixed a bug that existing notifications disappear if muting_notifications is set to true and muted.

### DIFF
--- a/app/services/mute_service.rb
+++ b/app/services/mute_service.rb
@@ -10,5 +10,6 @@ class MuteService < BaseService
     else
       FeedManager.instance.clear_from_timeline(account, target_account)
     end
+    mute
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/mute_service.rb
```diff
@@ -3,9 +3,12 @@
...
-    FeedManager.instance.clear_from_timeline(account, target_account)
+
     mute = account.mute!(target_account, notifications: notifications)
-    BlockWorker.perform_async(account.id, target_account.id)
-    mute
+    if mute.hide_notifications?
+      BlockWorker.perform_async(account.id, target_account.id)
+    else
+      FeedManager.instance.clear_from_timeline(account, target_account)
+    end
```
Gargron: I think the method is meant to return the mute at the end

abcang: OK. fixed it.

Gargron (approved):

PR #7317, opened 2018-05-02
Improve PostStatusService performance

Gargron:
Offload creation of local notifications to a worker. Remove two redundant SQL queries from ProcessMentionsService, remove n+1 XML/JSON serialization via memoization


ClearlyClaire (approved):
Wow I didn't realize the ActivityPub and OStatus payloads didn't actually depend on the mentioned account!

Looks good to me!

Issue #7039, opened 2018-04-04
Accessibility: The currently selected item in the post visibility dropdown is not marked for screen readers.


MarcoZehe:
To reproduce: Compose a new toot, and click open the Visibility dropdown. Screen readers such as VoiceOver on the Mac announce this as a collapsed or expanded menu.

1. First problem: Keyboard focus does not go to the menu items once they appear.
2. Second problem, these menu items are just buttons. The currently selected visibility option is not distinguishable from the others via some ARIA attribute such as aria-selected="true" on the selected/active item (not focused, but currently functionally active), so I don't know what the visibility is set to currently. Especially when replying to a post, I now first have to make sure I always select Direct Message if the post I reply to was such an item in the first place.
3. Third problem, instead of being buttons, these should probably be role menuitem, but only if keyboard support for arrowing up and down through them is implemented as well.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

PR #7318, opened 2018-05-02
Dropdowns accessibility

ClearlyClaire:
This is an attempt to fix accessibility issues with the different dropdown menus, including #7039.
It is not quite complete yet as keyboard navigation is not implemented.
Also, I'm far from being a web accessibility expert, so I could really use some review/feedback here.

### DIFF
--- a/app/javascript/mastodon/components/dropdown_menu.js
+++ b/app/javascript/mastodon/components/dropdown_menu.js
@@ -156,9 +156,6 @@ export default class Dropdown extends React.PureComponent {
 
   handleKeyDown = e => {
     switch(e.key) {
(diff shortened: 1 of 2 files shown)


ThibG:
I think I'm done implementing this, but I would appreciate some reviewing/feedback here.

Gargron (approved):
Seems fine...

mal0ki:
@ThibG did this fix all the issues in #7039, if so I'll close that issue.

ThibG:
Hm maybe not, the generic dropdown menu items are still buttons, and not menuitems, maybe that should be changed.

PR #7326, opened 2018-05-03
Do not override timestamps for incoming toots

ClearlyClaire:
I feel like this has already been discussed, but I could not remember nor find the rationale behind overriding timestamps on incoming toots.
I think this causes more harm than good. Indeed, it is an unnecessary distinction between fetched and received toots. Furthermore, it occurs from time to time that a temporary failure or large sidekiq queue would cause messages to be delayed for some extended amount of time. In this situation, having a large difference between the displayed time and the actual time the toot was sent can be quite confusing.


Gargron:
cc @aschmitz

Gargron (approved):

PR #7327, opened 2018-05-03
Prevent timeline from moving when cursor is hovering over it (fixes #7278)

### LINKED ISSUE #7278: Prevent timeline from moving when cursor is hovering over it
See [this toot](https://icosahedron.website/@Xkeeper/99934217414178449):

> can we get a feature where the timeline won't move if you are currently or have recently moved your mouse over it
> 
> im real tired of going to click thing a and then thing b slides in under it

* * * *

ClearlyClaire:
To some extent, this prevents the timeline from moving when the cursor is hovering it (fixes #7278).
To some extent only, because the overall scroll logic still doesn't handle properly the deletion of posts below the scroll position.

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -35,6 +35,7 @@ export default class ScrollableList extends PureComponent {
 
   state = {
     fullscreen: null,
+    mouseOver: false,
   };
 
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
>'e' is defined but never used

PR #7337, opened 2018-05-03
Throttle media post

akihikodaki:
I noticed that the rate limit for the media is so low that it allows to fill 10 gigabytes in 20 minutes, effectively causing denial-of-service, so here is a fix. I appreciate opinions about the concrete numbers for the rate limit:

The previous rate limit allowed to post media so fast that it is possible
to fill up the disk space even before an administrator notices. The new
rate limit is configured so that it takes 24 hours to eat 10 gigabytes:
10 * 1024 / 8 / (24 * 60 / 30) = 27 (which is rounded to 30)

The period is set long so that it does not prevent from attaching several
media to one post, which would happen in a short period. For example,
if the period is 5 minutes, the rate limit would be:
10 * 1024 / 8 / (24 * 60 / 5) = 4

This long period allows to lift the limit up.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -53,6 +53,10 @@ class Rack::Attack
     req.ip if req.api_request?
   end
 
+  throttle('throttle_media', limit: 30, period: 30.minutes) do |req|
+    req.authenticated_user_id if req.post? && req.path.start_with('/api/v1/media')
+  end
+
   throttle('protected_paths', limit: 25, period: 5.minutes) do |req|
     req.ip if req.post? && req.path =~ PROTECTED_PATHS_REGEX
   end


Gargron:
Do you know why IP-based rate limits are so high? We had complaints from people on shared IPs (households, universities). Also I remember somebody was relying on the high limits so they could ignore messing with TRUSTED_PROXY configuration because they had a multi-server setup where the IP always looked like 127.0.0.1. I do not think that's reasonable, but shared IPs are a problem. Maybe this could be API-key dependent after all?

akihikodaki:
8093082fda654747ff2edc0799211240d573e418: now the limit is user-ID-based.

Gargron (approved):

PR #7338, opened 2018-05-03
Add a missing question mark in rack_attack.rb

akihikodaki:
Reported at https://github.com/tootsuite/mastodon/commit/b1d4471e36830aa7dc4425ff6c48035a95c9367a#r28841392.
~~\# question: why did it work?~~ I found I haven't `./bin/rails dev:cache`. I confirmed the rate limit is now in effect.


Gargron (approved):

Gargron:
>question: why did it work?

My guess: It's inside a lambda, so not executed at boot.

PR #7344, opened 2018-05-03
Do not re-position scroll when loading more (inserting items from below)

ClearlyClaire:
When a scrollable list is loading, let it scroll to top. Otherwise, having the cursor hover the list while it's initially loading will result in it scrolling to the bottom (and triggering another load event).

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -72,7 +72,7 @@ export default class ScrollableList extends PureComponent {
     const someItemInserted = React.Children.count(prevProps.children) > 0 &&
       React.Children.count(prevProps.children) < React.Children.count(this.props.children) &&
       this.getFirstChildKey(prevProps) !== this.getFirstChildKey(this.props);
-    if (someItemInserted && this.node.scrollTop > 0 || this.state.mouseOver) {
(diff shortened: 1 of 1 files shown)


Gargron (approved):

nightpool:
this is running on glitch right now and it's REALLY confusing behavior

nightpool:
i keep missing posts because i left my mouse over the home timeline and forgot about it

ThibG:
Scroll position + column header + “glow” should still indicate new posts.
I don't really find this behavior confusing, but would it be better if it only triggered when the cursor moved?

Gargron:
I am getting a vague memory that I tried that before, and I think I was annoyed because if you moved your mouse, and a new post appeared, it would obviously no longer scroll to top when even more posts came, so the timeline was constantly getting "stuck" - that was the *intention*, but it didn't feel right.

I'm gonna say you're probably going to have to take the whole thing back out.

charlag:
Will we have some alternative solution later? Because timeline moving under your cursor is kinda annoying too...

Gargron:
Timeline only moved when you're scrolled all the way to the top... so misclicks happen, but not as often as my mouse is hovering over the timeline. I'd rather have the former :P

PR #7346, opened 2018-05-04
Marginally improve convert/ffmpeg calls performance with posix-spawn

Gargron:
Terrapin (formerly Cocaine):

>You can potentially increase performance by installing the posix-spawn gem. This gem can keep your application's heap from being copied when forking command line processes. For applications with large heaps the gain can be significant. To include posix-spawn, simply add it to your Gemfile or, if you don't use bundler, install the gem.


nightpool:
.... how can we tell if paperclip uses this correctly?

Gargron:
@nightpool Paperclip uses terrapin to run the convert/identify/file commands. paperclip-av-transcoder is a bit less recently updated, but it uses the av gem, which uses cocaine to run its commands. Cocaine is the same gem as Terrapin, just an older version. Both versions have the quoted functionality: when posix-spawn is available, it is used.

cc @mike-burns

nightpool (approved):

ThibG:
Hm, glitch-soc have used that gem for a while, and I had an issue with it on 32b ARM so I locally disabled it. Can't remember what was the issue exactly, though…

PR #7348, opened 2018-05-04
4 profile fields max, store only 255 characters per name/value

### LINKED ISSUE #7303: Limit characters in bio fields
See #6645: The fields currently do not have character limits. They should do. In the absence of better ideas, just 255 character per name and per value. Also, put a hard limit on the number of fields. Basically, restrict how much data people can dump into that DB column.

Gargron:
Fix #7303

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -74,6 +74,7 @@ class Account < ApplicationRecord
   validates_with UnreservedUsernameValidator, if: -> { local? && will_save_change_to_username? }
   validates :display_name, length: { maximum: 30 }, if: -> { local? && will_save_change_to_display_name? }
   validates :note, length: { maximum: 160 }, if: -> { local? && will_save_change_to_note? }
+  validates :fields, length: { maximum: 4 }, if: -> { local? && will_save_change_to_fields? }
 
   # Timelines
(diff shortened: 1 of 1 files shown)


trwnh:
Hmm, 4 fields seems a bit... low? And 255 characters seems much higher than any intended use case I can think of. Not to advocate for strictly lower limits, but 64 seems closer to average than 255.

Gargron:
In this case I only care about having any limit at all. I don't want to put a character counter next to every field. We'll just visually trim the data. However, i doubt that more than 4 fields will look good in profiles, so that's what I chose.

koyuawsmbrtn:
Wouldn't it be better to limit this to initially three and then adding up to 3 more fields with a plus sign on the bottom? When the six fields are reached the button is disabled and when hovering over it, it should say, that the limit has reached. Would make a better user experience in my opinion.

nightpool (approved):

trwnh:
I understand that to be practical, the limit on "how many fields" should be low (to prevent very tall profiles, since it's currently implemented as a vertical table). I just think that the limit should be 5 at minimum -- even the release notes list 5 use cases despite the limit being 4. And fwiw, 5 is also a "rounder" number (though I'm personally biased towards odd numbers over even numbers due to better centering).

PR #7350, opened 2018-05-04
Add tests for admin custom emojis controller

shuheiktgw:
I added tests for admin/custom_emojis_controller and fixed small bug I found in the controller!
Thanks! 👍

### DIFF
--- a/spec/controllers/admin/custom_emojis_controller_spec.rb
+++ b/spec/controllers/admin/custom_emojis_controller_spec.rb
@@ -1,0 +1,115 @@
+require 'rails_helper'
+
+describe Admin::CustomEmojisController do
+  render_views
+
+  let(:user) { Fabricate(:user, admin: true) }
+
+  before do
+    sign_in user, scope: :user
+  end
+
+  describe 'GET #index' do
+    subject { get :index }
+
+    before do
+      Fabricate(:custom_emoji)
+    end
+
+    it 'renders index page' do
+      expect(subject).to have_http_status 200
+      expect(subject).to render_template :index
+    end
+  end
+
+  describe 'GET #new' do
+    subject { get :new }
+
+    it 'renders new page' do
+      expect(subject).to have_http_status 200
+      expect(subject).to render_template :new
+    end
+  end
+
+  describe 'POST #create' do
+    subject { post :create, params: { custom_emoji: params } }
+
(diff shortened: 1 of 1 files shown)


spec/controllers/admin/custom_emojis_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    let(:image) { fixture_file_upload(Rails.root.join('spec', 'fixtures', 'files', 'emojo.png'), 'image/png') }
+
+    before do
+      put :update, params: { id: custom_emoji.id, custom_emoji: params }
+    end
+
+    context 'when parameter is valid' do
+      let(:params) { { shortcode: :updated, image: image } }
+
+      it 'succeeds in updating custom emoji' do
```
ThibG: This should work, but I can't find anything making use of that functionality (changing the shortcode), so I'm not sure that's intended?

ClearlyClaire (dismissed):
Apart from the use of symbols rather than strings for shortcodes, and the other inline remark I have made, this looks good to me.

spec/controllers/admin/custom_emojis_controller_spec.rb
shuheiktgw: The original `shortcode` is `test` (set it [here](https://github.com/shuheiktgw/mastodon/blob/7808f82d22f76c05efbc8169eae5826deab671bc/spec/controllers/admin/custom_emojis_controller_spec.rb#L57)) and the test asserts that the shortcode is updated to `updated`! 👍

shuheiktgw:
@ThibG Thank you for your reviews! I fixed them! 👍

app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -56,7 +56,7 @@ def copy
       emoji.image = @custom_emoji.image
 
       if emoji.save
-        log_action :create, emoji
+        log_action :copy, emoji
```
Gargron: We [do not have translations for "copy"](https://github.com/tootsuite/mastodon/blob/master/config/locales/en.yml#L148-L176). Copying a custom emoji is essentially creating one. Please revert this

Gargron (changes requested):

Gargron (approved):

Issue #7204, opened 2018-04-20
collections/featured is fetched too often


ThibG:
I have yet to track exactly why, but collections/featured for a given user seems to be fetched several times an hour, which is at odds with the “refresh account info at most once a day” thing.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Hmm, okay, I want to know about this too :thinking:

ThibG:
One reason seems to be that this update (along with requests to outbox, followers and following) is triggered unconditionally by `ActivityPub::ProcessAccountService`, which is also called when receiving an `Update`. Those are costly requests that could probably be avoided.

I don't think that covers everything though since I rarely issue such `Update` activities… maybe my account is needlessly re-resolved because of `WEB_DOMAIN` ≠ `LOCAL_DOMAIN`?

ThibG:
One of the most obvious cause of load in Mastodon is remote instances fetching account data, typically causing requests to the following endpoints:
- json actor (`AccountsController`)
- webfinger (`WellKnown::WebfingerController`)
- outbox (`ActivityPub::OutboxesController`)
- following (`FollowingAccountsController`)
- followers (`FollowerAccountsController`)
- featured collection (`ActivityPub::CollectionsController`)

Everything except the json and webfinger requests is expensive. Now that's identified, let's look at where and why the expensive requests are made:

- outbox, following and followers are fetched by `ActivityPub::ProcessAccountService` in order to display an accurate toots/following/followers count
- the featured collection is fetched by `ActivityPub::FetchFeaturedCollectionService` to replicate pinned toots, it's triggered unconditionally by `ActivityPub::ProcessAccountService`

So, all those requests are made whenever an ActivityPub account is processed, that is, whenever account info is fetched and whenever a followed account is updated (`app/lib/activitypub/activity/update.rb`).

ThibG:
So, all of those expensive requests are made whenever an account is fetched/updated, but when does that happen? Unfortunately, it's a bit of a mess, but there are two rules of thumb:
- For followed accounts, whenever the account issues an `Update` (so when they change avatar, etc.)
- For accounts that are not followed, whenever the need to get their info arises (typically when one of their statuses is fetched) and `account.possibly_stale?` is true.

[...]

PR #7357, opened 2018-05-04
Do not render first page of following and followers collections unless explicitly asked to

ClearlyClaire:
Partial fix to #7204 making following and followers collections cheaper to render.

I haven't changed the outbox controller yet as it's using a different logic and an intermediate cache.

### DIFF
--- a/app/controllers/follower_accounts_controller.rb
+++ b/app/controllers/follower_accounts_controller.rb
@@ -4,11 +4,10 @@ class FollowerAccountsController < ApplicationController
   include AccountControllerConcern
 
   def index
-    @follows = Follow.where(target_account: @account).recent.page(params[:page]).per(FOLLOW_PER_PAGE).preload(:account)
-
     respond_to do |format|
       format.html do
-        @relationships = AccountRelationshipsPresenter.new(@follows.map(&:account_id), current_user.account_id) if user_signed_in?
+        follows
+        @relationships = AccountRelationshipsPresenter.new(follows.map(&:account_id), current_user.account_id) if user_signed_in?
       end
 
       format.json do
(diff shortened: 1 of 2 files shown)


Gargron:
Good thing I uploaded my #7356 now before we did the same thing twice. It seems like luckily we did different parts!

Gargron (approved):

app/controllers/follower_accounts_controller.rb
```diff
@@ -4,11 +4,9 @@ class FollowerAccountsController < ApplicationController
   include AccountControllerConcern
 
   def index
-    @follows = Follow.where(target_account: @account).recent.page(params[:page]).per(FOLLOW_PER_PAGE).preload(:account)
-
     respond_to do |format|
       format.html do
-        @relationships = AccountRelationshipsPresenter.new(@follows.map(&:account_id), current_user.account_id) if user_signed_in?
+        @relationships = AccountRelationshipsPresenter.new(follows.map(&:account_id), current_user.account_id) if user_signed_in?
```
Gargron: If `user_signed_in?` is false, `follows` is never called, so `@follows` is not assigned, but it is required for the HTML views.

```
ActionView::Template::Error:
       undefined method `map' for nil:NilClass
```

Gargron (changes requested):

app/controllers/follower_accounts_controller.rb
Gargron: Far easier solution is just to call follows once above this line

ThibG: Sorry about that. Fixed it.

Gargron (approved):

PR #7359, opened 2018-05-04
Do not check outbox/featured if account has local followers

Gargron:
See #7204: If account has local followers, we will receive Create/Delete and Add/Remove and update those things correctly.

I'm not sure about this PR though. Are there edge cases where someone might have followers before the correct data is fetched? Network issues could also lead to desync which would no longer get corrected...

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -164,6 +164,10 @@ class Account < ApplicationRecord
     subscription_expires_at.present?
   end
 
+  def locally_followed?
+    passive_relationships.any?
+  end
+
   def possibly_stale?
     last_webfingered_at.nil? || last_webfingered_at <= 1.day.ago
   end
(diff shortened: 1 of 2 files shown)


ykzts (approved):

ThibG:
This would be a nice change, but yeah, there might be corner cases:
1. Info might not be accurate at the moment the first person starts following this person (a possible fix for that would be to unconditionally fetch up-to-date info just after the first follow)
2. Long network failures may cause such data to be out-of-sync. Maybe conditionally forcefully-update, e.g. when `possibly_stale?` returns true?

unarist:
I think current Add/Remove handling misses about...

* statuses not in statuses table (e.g. posted before follow)
* very old statuses with OStatus ids
* Remove activity may be arrived before Add activity (c.f. delete_later in Delete activity)

Although those things can be fixed in Add/Remove handling, we can ignore some or all of those if we do synchronization sometimes.

PR #7371, opened 2018-05-05
Fallback to old on-the-fly URI for follows/blocks if no stored URI

### LINKED ISSUE #7367: Displayed 500 error at WebUI when unfollowing other instance's account at first
If I unfollow the account of another instance, displayed "500 Internal Server Error" at WebUI.
Also, although it seems to have failed in unfollow, it actually succeeds.
This will not come out when following  and unfollowing the next time.

environment: Equivalent to master 2ef9d6505 ( difference is [here](https://github.com/tootsuite/mastodon/compare/master...mayaeh:for-edge ) )

The output to the server log was as follows:

bundle[26483]: [ff518863-fd34-4f35-ba72-db7f0dd4cac6] method=POST path=/api/v1/accounts/3/unfollow format=html controller=Api::V1::AccountsController action=unfollow status=500 error='JSON::LD::JsonLdError::InvalidIdValue: invalid @id value: value of @id must be a string or hash if framing: nil' duration=150.61 view=0.00 db=11.86

* * * *

Gargron:
Fix #7367

### DIFF
--- a/app/serializers/activitypub/block_serializer.rb
+++ b/app/serializers/activitypub/block_serializer.rb
@@ -5,7 +5,7 @@ class ActivityPub::BlockSerializer < ActiveModel::Serializer
   attribute :virtual_object, key: :object
 
   def id
-    [ActivityPub::TagManager.instance.uri_for(object.account), '#blocks/', object.id].join
+    ActivityPub::TagManager.instance.uri_for(object) || [ActivityPub::TagManager.instance.uri_for(object.account), '#blocks/', object.id].join
   end
 
   def type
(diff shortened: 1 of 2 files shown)


nightpool:
this still wont use `payload` uris for local statuses?

```
     return target.uri if target.respond_to?(:local?) && !target.local?
```

nightpool:
oh, nevermind, I guess:

image

.... this seems Bad

nightpool:
Why not just `target.uri if target.respondto(:uri) && target.uri.present? || target.respondto(:local) && target.local.present?`

PR #7385, opened 2018-05-06
Add Corsican localization

alixrossi:


config/locales/activerecord.co.yml
```diff
@@ -0,0 +1,13 @@
+---
+en:
```
Gargron: co

palindromordnilap:
Okay, it seems I need to run `i18n-tasks normalize` on the files. Unfortunately I can't do that on the one computer I have access to right now. Oh well.

Gargron:
You don't have to close the PRs every time... Maybe someone will come along and run i18n-tasks normalize for you...

palindromordnilap:
I'm closing it because I'm gonna make a bunch of commits and pushes and I'm not gonna spam that PR with them until I'm done.

nightpool:
it's better to keep the PR open, so it only takes up 1 place in the list.

palindromordnilap:
I can actually reopen it without making a new one, I just didn't notice that the first time.

PR #7390, opened 2018-05-06
Add Corsican localization

Gargron:
Resolve #7385


nightpool (approved):

nightpool:
maybe needs to be rebased against master? 
image

PR #7395, opened 2018-05-07
Do not mark remote status sensitive even if spoiler text is present

akihikodaki:
Old statuses and statuses from Pawoo, which runs a modified version of
Mastodon, may not have been marked sensitive even if spoiler text is
present.

Such statuses are still not marked sensitive if they are local or
arrived before version upgrade. Marking recently fetched remote status
sensitive contradicts the behavior.

Considering what people expected when they authored such statuses, this
change removes the sensitivity enforcement.

Of course there is another option: perform migration to mark all statuses with spoiler texts sensitive.

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -172,7 +172,6 @@ class Status < ApplicationRecord
   before_validation :set_reblog
   before_validation :set_visibility
   before_validation :set_conversation
-  before_validation :set_sensitivity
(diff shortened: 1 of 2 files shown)


Gargron:
This seems like a minor issue, no? How often do people look at old statuses?

Gargron (approved):

PR #7410, opened 2018-05-08
Store the first supported Actor type in actor_type

ekiru:
This changes ProcessAccountService to set the actor_type to the first recognized Actor type in the JSON-LD document.

When an Actor has multiple types, this avoids storing it as the stringification of an Array and allows Actors with multiple types to be recognized as bots if they have Service or Application as the first standard ActivityStreams Actor type in the type property.

I've also updated the example in the specs for resolving accounts with multiple accounts to check that for an Actor with `["Person","vcard:individual"]` as its type, the `actor_type` attribute on the Account gets set to "Person".


nightpool (approved):
👍 looks great!

PR #7420, opened 2018-05-08
Reword the direct message warning

### LINKED ISSUE #7030: Reword direct message warning to set proper privacy expectations
The DM warning is misleading, saying that “This toot will only be visible to all the mentioned users.”.

Currently, any moderator plus anyone with read access to the database can read those toots.

The warning should be updated to reflect that.

Sylvhem:
- Reword the direct message warning to set proper privacy expectations (fix #7030).
- Update the French translation accordingly.

### DIFF
--- a/app/javascript/mastodon/features/compose/containers/warning_container.js
+++ b/app/javascript/mastodon/features/compose/containers/warning_container.js
@@ -21,7 +21,7 @@ const WarningWrapper = ({ needsLockWarning, hashtagWarning, directMessageWarning
     return <Warning message={<FormattedMessage id='compose_form.hashtag_warning' defaultMessage="This toot won't be listed under any hashtag as it is unlisted. Only public toots can be searched by hashtag." />} />;
   }
   if (directMessageWarning) {
-    return <Warning message={<FormattedMessage id='compose_form.direct_message_warning' defaultMessage='This toot will only be visible to all the mentioned users.' />} />;
+    return <Warning message={<FormattedMessage id='compose_form.direct_message_warning' defaultMessage='This toot will only be sent to all the mentioned users. However, the operators of your instance and any receiving instances may see this message.' />} />;
   }
 
   return null;
--- a/app/javascript/mastodon/locales/en.json
+++ b/app/javascript/mastodon/locales/en.json
@@ -59,7 +59,7 @@
   "column_header.unpin": "Unpin",
   "column_subheading.navigation": "Navigation",
   "column_subheading.settings": "Settings",
-  "compose_form.direct_message_warning": "This toot will only be sent to all the mentioned users. However, the operators of your instance and any receiving instances may see this message.",
+  "compose_form.direct_message_warning": "This toot will only be sent to the mentioned users. However, the operators of yours and any of the receiving instances may inspect this message.",
   "compose_form.hashtag_warning": "This toot won't be listed under any hashtag as it is unlisted. Only public toots can be searched by hashtag.",
   "compose_form.lock_disclaimer": "Your account is not {locked}. Anyone can follow you to view your follower-only posts.",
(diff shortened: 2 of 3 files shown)


ClearlyClaire (approved):

trwnh:
>Facebook and Twitter do not warn their users like this and it's an unfair advantage for them.

I'd rather we err on the side of transparency. FB and Twitter choose not to warn users because they don't care about privacy, and are willing to read private messages when wanted or requested.

Sylvhem:
> We talked about this before when the previous warning text came up, but yeah "will only be sent to all" doesn't sound correct.

Maybe "This toot will only be sent to the mentioned users." would be better?

Gargron:
Please use:

>This toot will only be sent to the mentioned users. However, the admin of yours and the receiving instance may ___ this message.

- see?
- view?
- inspect?

Not sure which is best. It needs to be indicated that it would take targeted action by the admin, rather than "all DMs wheezing by on the admin's screen"

trwnh:
"...However, instance admins may view this message if it is reported."

AFAIK, admins can only see a direct message if it's 1) reported, or 2) if they look through the database manually, correct? And there's another open issue for 3) clicking a button to temporarily view direct messages while investigating a report, right? The above language should cover case 1 and maybe case 3, though IDK what language to use to cover case 2.

Gargron:
What if instead of "may", it's just "able to"?

Sylvhem:
> However, the admin of yours and the receiving instance may ___ this message.

I choose to use "operator" instead of "admin" because we seemed to agreed in [this discussion](https://github.com/tootsuite/mastodon/issues/7030), that it should be highlighted that not only the admins but also the moderators and the sysadmins could access to people DM. Since "operator" is the term used on [the new privacy policy](https://github.com/tootsuite/mastodon/pull/6666), I think we should reused it here.

> Not sure which is best. It needs to be indicated that it would take targeted action by the admin, rather than "all DMs wheezing by on the admin's screen"

Inspect is probably the best choice then.

New proposal: "This toot will only be sent to the mentioned users. However, the operators of yours and any of the receiving instances may inspect this message.".

> The above language should cover case 1 and maybe case 3, though IDK what language to use to cover case 2.

The wording I suggest actually covers all three cases.

Gargron:
>New proposal: "This toot will only be sent to the mentioned users. However, the operators of yours and any of the receiving instances may inspect this message.".

:+1:

Gargron (approved):

ThisIsMissEm (commented):
I honestly think this message creates undue amounts of FUD; perhaps: “this message is not encrypted & is stored in plaintext.”

Only server admins with database access can see DMs unless they’re reported.

Also, I know I’m late to comment, have been away with medical issues, still not back to 100%

ThisIsMissEm:
You could use symmetric-encryption, a ruby gem to do exactly that. I also apply this gem to email addresses and oauth tokens, such that if the database is accessed by someone unauthorized, then it limits exposure.

This database level security doesn’t prevent anyone from using the rails console with production details in order to read it, so, it’s really protecting against accidental exposure.

Perhaps: “This toot is stored just the same as all other toots, however, it is only shown to mentioned recipients. Administrators may still access it – do you trust them?”

If you’re tooting a DM to a foreign server, similar to the “do you trust this foreign server?” Message.

Gargron:
>I also apply this gem to email addresses and oauth tokens,

But then you can't query by those values quickly? And you kinda have to. Be careful re: performance

ThisIsMissEm:
You can, you just encrypt the value, then search the database; it’s not using unique encryption keys per row, just per environment.

True, you couldn’t do partial matches anymore. Not sure how much of a problem this would be. I know we’ve occasionally used partial matches to find duplicate accounts, but usernames or display names tend to be more common.

PR #7423, opened 2018-05-09
Hide reblogs of accounts that have been muted

TheInventrix:
This adds a check to make sure boosts *of* accounts that have been muted are not shown in the home timeline.

Currently, if you have User A muted, any boosts of User A's posts by people you follow will show up in your home timeline, which is not the expected behavior for a mute function. This fixes that.


app/lib/feed_manager.rb
```diff
@@ -168,6 +168,7 @@ def filter_from_home?(status, receiver_id)
     elsif status.reblog?                                                                                                         # Filter out a reblog
       should_filter   = Follow.where(account_id: receiver_id, target_account_id: status.account_id, show_reblogs: false).exists? # if the reblogger's reblogs are suppressed
       should_filter ||= Block.where(account_id: status.reblog.account_id, target_account_id: receiver_id).exists?                # or if the author of the reblogged status is blocking me
+      should_filter ||= Mute.where(account_id: receiver_id, target_account_id: status.reblog.account_id).exists?                # or if I have the author of the reblogged status muted
```
Gargron: Line 161 already checks for this though?? How is this different? (Also Line 170 might no longer be necessary either)

Also I think we might have unit tests to confirm this behaviour, if they're missing an edge case it would be good to add it to confirm a fix actually does its job

Gargron (changes requested):

PR #7434, opened 2018-05-10
Add devise_pam_authenticatable2 to testgroup

abcang:
Since `devise_pam_authenticatable2` is used in test, it changed to belong to the test group. It is installed when `--without test` is not specified or when `--with pam_authentication` is specified. That is, `devise_pam_authenticatable2` will be installed if both are specified.

related: #7147


Tak:
You can't run `rspec` on a vanilla development install without this change

Gargron:
I have to admit that the `group` stuff is incredibly confusing and I have no idea if this change is going to break anything or what exactly its effects are. Group within a group?

PR #7441, opened 2018-05-10
Add tests for invites controller

shuheiktgw:
Added tests for invites controller and also small refactorings. 👍

### DIFF
--- a/app/controllers/invites_controller.rb
+++ b/app/controllers/invites_controller.rb
@@ -8,9 +8,9 @@ class InvitesController < ApplicationController
   before_action :authenticate_user!
 
   def index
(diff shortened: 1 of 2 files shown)


app/controllers/invites_controller.rb
```diff
@@ -8,9 +8,9 @@ class InvitesController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    authorize :invite, :create?
+    authorize :invite, :index?
```
shuheiktgw: Is this supposed to be `index?` instead of `create?`

Gargron (approved):

PR #7445, opened 2018-05-10
Add REST API for Web Push Notifications subscriptions

### LINKED ISSUE #7368: Generalised push notifications
I have been experimenting with the idea of relaying web push notifications to native mobile notifications for my iOS app. This would entail setting up my own notification endpoint that receives web pushes from Mastodon, decodes them, and pushes them to APNs. This would be a lot robuster and more lightweight than keeping a ton of connections open to the notification streams of every user.

This seems to be theoretically workable, but there are practical problems. The smaller is that there is no endpoint in the regular API for subscribing to pushes. It only exists in the web API.

The larger is that push subscriptions are tied to sessions, and to send out pushes, Mastodon iterates through the available sessions and pushes to their push endpoints, if any. And I do not think there is a reasonable way for an app using the regular APIs to have a session for this purpose?

So what would be needed to make this work are:

* Endpoints to subscribe, update and unsubscribe for push notifications in the regular API.
* Decoupling push subscriptions from sessions.

As an alternative, it might not be needed to change the current web push code for this, but instead add a second type of simplified push. This kind of implementation does not really need all the features of web pushes, it just needs a notification payload to be pushed an endpoint. Web pushes do this, but they also encrypt the payload, which may be redundant. Either solution would work fine, though.

Gargron:
- POST /api/v1/push/subscription
- PUT /api/v1/push/subscription
- DELETE /api/v1/push/subscription
- New OAuth scope: "push" (required for the above methods)

Fix #7368

___

Principally, this changes distribution logic from depending on SessionActivation, instead directly querying WebPushSubscriptions. This PR includes a data migration that fills missing new columns on that table.


abcang (approved):
I think that there is no problem with implementation.

DagAgren:
What happens if you ask for "push" scope on an old version?

PR #7452, opened 2018-05-11
Improvements to toots display in admin view

ClearlyClaire:
* Shows “Boosted from [link to admin view of original account]” for boosted toots and show proper media/CW/sensitive status

screenshot-2018-5-11 etat du compte - thib - mastodon instance perso de thibg

* Restore “show more” button from public views to make CW usage clearer (though content is initially unfolded)

screenshot-2018-5-11 etat du compte - thib - mastodon instance perso de thibg 1

### DIFF
--- a/app/helpers/admin/account_moderation_notes_helper.rb
+++ b/app/helpers/admin/account_moderation_notes_helper.rb
@@ -10,10 +10,16 @@ module Admin::AccountModerationNotesHelper
     end
   end
 
+  def admin_account_inline_link_to(account)
+    link_to admin_account_path(account.id), class: name_tag_classes(account, true) do
+      content_tag(:span, account.acct, class: 'username')
+    end
+  end
+
   private
 
-  def name_tag_classes(account)
-    classes = ['name-tag']
+  def name_tag_classes(account, inline = false)
+    classes = [inline ? 'inline-name-tag' : 'name-tag']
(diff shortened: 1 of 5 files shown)


Gargron:
I don't see the use in the show more button. If you insist, just prepend a "content warning: " localized string before it.

ThibG:
The use is of the show more button is simply to have a recognizable marker than the thing is a CW.
A slightly thicker font weight is not really recognizable. As for prepending with a localized “Content warning:” only partially solves that, because a toot starting with “Content warning: blalabla” isn't easily recognizable from a toot with a proper content warning.

Gargron:
@ThibG Okay how about making the bold text... $error-red or something. I do not want a button that does nothing there. I don't want any button there. Sorry

ThibG:
It does a thing, it hides and shows the toot's contents, exactly like on a public page (just that the toots are shown by default). True, this toggle functionality isn't extremely useful here, but I think there is value in having a consistent UI…

ThibG:
I have removed the button.


pr

Gargron (approved):
Nice

theotix:
wtf

ProgVal:
wat

arobase-che:
💕 "Make content warnings swag" 💕

gg

sireebob:
this is bullshit

PR #7457, opened 2018-05-12
Use self.navigator in code common for window and service worker

akihikodaki:
See https://github.com/tootsuite/mastodon/issues/7430#issuecomment-388370577.


akihikodaki:
Not working.

PR #7476, opened 2018-05-14
Improve index on statuses for api/v1/accounts account_id statuses

akihikodaki:
Queries with the combination of account_id, id, and visibility can be
categorized in three types:
1. Querying for public and unlisted to enumerate statuses visible to
anyone.
2. Querying for public, unlisted, and private to enumerate statuses
visible to follower.
3. Querying for direct to enumerate own direct statuses.

1 and 2 is covered by the index with condition 'visibility IN (0, 1, 2)'.
It would bring better performance in case that there are many direct
statuses.

The index with condition 'visibility = 3' is just for 3. It would be much
faster to query direct statuses thanks to this query.

The total size of those two indexes are expected to be smaller than the
deleted one because they are partial and does not have to cover all the
table.

In case of Pawoo, index with condition 'visibility IN (0, 1, 2)' costed 2.8 GB while the removed index costed 3.5GB. The index with condition 'visibility = 3' is not created as it has no benefit without direct column.

### DIFF
--- a/db/migrate/20180514130000_improve_index_on_statuses_for_api_v1_accounts_account_id_statuses.rb
+++ b/db/migrate/20180514130000_improve_index_on_statuses_for_api_v1_accounts_account_id_statuses.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class ImproveIndexOnStatusesForApiV1AccountsAccountIdStatuses < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :statuses, [:account_id, :id, :visibility], where: 'visibility IN (0, 1, 2)', algorithm: :concurrently
+    add_index :statuses, [:account_id, :id], where: 'visibility = 3', algorithm: :concurrently
+    remove_index :statuses, column: [:account_id, :id, :visibility, :updated_at], order: { id: :desc }, algorithm: :concurrently, name: :index_statuses_20180106
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_05_10_230049) do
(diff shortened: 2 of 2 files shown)


akihikodaki:
For your information, here is the predicted worst case query for the direct column on Pawoo, taking 2 seconds to execute without this improvement:
```
> EXPLAIN ANALYZE SELECT statuses.id FROM statuses LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id AND mentions.account_id = ? WHERE (mentions.account_id = ? OR statuses.account_id = ?) AND visibility = 3 LIMIT 20
Limit  (cost=1740.62..120965.30 rows=20 width=8) (actual time=1854.100..1856.143 rows=20 loops=1)
 ->  Hash Left Join  (cost=1740.62..2761791.95 rows=463 width=8) (actual time=1854.100..1856.132 rows=20 loops=1)
       Hash Cond: (statuses.id = mentions.status_id)
       Filter: ((mentions.account_id = ?) OR (statuses.account_id = ?))
       Rows Removed by Filter: 2963
       ->  Index Only Scan using index_statuses_20180106 on statuses  (cost=0.57..2453633.18 rows=314679 width=16) (actual time=0.531..1711.653 rows=2983 loops=1)
             Index Cond: (visibility = 3)
             Heap Fetches: 2122
       ->  Hash  (cost=1730.34..1730.34 rows=777 width=16) (actual time=141.014..141.014 rows=2021 loops=1)
             Buckets: 2048 (originally 1024)  Batches: 1 (originally 1)  Memory Usage: 111kB
             ->  Index Only Scan using index_mentions_on_account_id_and_status_id on mentions  (cost=0.56..1730.34 rows=777 width=16) (actual time=1.072..139.963 rows=2021 loops=1)
                   Index Cond: (account_id = ?)
                   Heap Fetches: 1509
Planning time: 0.315 ms
Execution time: 1856.189 ms
```

The result with this improvement is not available as I noted before though I believe the improvement would be huge.

Gargron (approved):

Issue #7479, opened 2018-05-14
Account URI not updated when using ActivityPub


wiktor-k:
Hello,
I noticed that during remote account update using WebFinger and ActivityPub the `uri` field is not updated to new value (fragment from `app/services/activitypub/process_account_service.rb`:

```rb
  def set_immediate_attributes!
    @account.inbox_url               = @json['inbox'] || ''
    @account.outbox_url              = @json['outbox'] || ''
    @account.shared_inbox_url        = (@json['endpoints'].is_a?(Hash) ? @json['endpoints']['sharedInbox'] : @json['sharedInbox']) || ''
    @account.followers_url           = @json['followers'] || ''
    @account.featured_collection_url = @json['featured'] || ''
    @account.url                     = url || @uri
    @account.display_name            = @json['name'] || ''
    @account.note                    = @json['summary'] || ''
    @account.locked                  = @json['manuallyApprovesFollowers'] || false
    @account.fields                  = property_values || {}
    @account.actor_type              = actor_type
  end
```

(missing `@account.uri = @uri`).

but when using OStatus it is updated (fragment from `app/services/resolve_account_service.rb`):
```rb
  def update_account
    @account.last_webfingered_at = Time.now.utc
    @account.protocol            = :ostatus
    @account.remote_url          = atom_url
    @account.salmon_url          = salmon_url
    @account.url                 = url
    @account.public_key          = public_key
    @account.uri                 = canonical_uri  # <---------- here
    @account.hub_url             = hub_url
    @account.save!
  end
```

This causes issues when one remote account changes it's primary URL.

I think it should be similar for both protocols but would like someone more aquatinted with Mastodon codebase to double check it.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

PR #7488, opened 2018-05-14
Fix account URI when updating ActivityPub account

wiktor-k:
Updates account `uri` field on each call to `update_account` instead of only once during `create_account` to mirror the same behavior in OStatus [`ResolveAccountService` class][0].

ActivityPub accounts are identified using `@username` and `@domain` pair instead of URI since #6842.

This fixes a bug when the account identified by `@username` and `@domain` changes its URI (#7479).

[0]:
https://github.com/tootsuite/mastodon/blob/03b69ebc450efc07246bd64204276941b7ede3fc/app/services/resolve_account_service.rb#L121

### DIFF
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -44,7 +44,6 @@ class ActivityPub::ProcessAccountService < BaseService
     @account.protocol    = :activitypub
     @account.username    = @username
     @account.domain      = @domain
-    @account.uri         = @uri
     @account.suspended   = true if auto_suspend?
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
No but you changed one line, explained what you were doing, and you are closing a bug report. I suppose the only way this PR could be better if you included the string "Fix #7479" anywhere in commit message or PR body so the bug would be automatically closed when this is merged.

wiktor-k:
Good idea, let me replace this commit in a minute...

PR #7493, opened 2018-05-15
Cache attachments on external host with service worker

akihikodaki:


unarist:
* Does it work with non-S3 configuration, e.g. Swift?
* Can we remove "The asset host must allow cross origin request from WEB_DOMAIN or LOCAL_DOMAIN" comments just before CDN_HOST in .env.production? (if this patch is good, though)

akihikodaki:
> Does it work with non-S3 configuration, e.g. Swift?

No.

> Can we remove "The asset host must allow cross origin request from WEB_DOMAIN or LOCAL_DOMAIN" comments just before CDN_HOST in .env.production? (if this patch is good, though)

No. We need CORS configuration on *both* hosts.

akihikodaki:
a7d07c169882ad57d65e70ada18d7e950272b713 supports Swift, too. I do not have Swift host so it is not tested. It should work as far as I see the source code of Paperclip.

By the way, Paperclip supports load balancing (see https://github.com/thoughtbot/paperclip/commit/2dcb1766c290902ca86d125ee74f5892d870e251), but it is not supported by this change. The load balancing feature is not documented and inflexible. I do not expect anybody uses it, but I note this here just in case.

akihikodaki:
0a954d3f2e639c10775752320266dcae91524d5b addresses https://github.com/tootsuite/mastodon/pull/7493#pullrequestreview-123281251.

abcang (approved):

abcang (approved):

config/webpack/production.js
```diff
@@ -6,8 +6,9 @@ const CompressionPlugin = require('compression-webpack-plugin');
 const sharedConfig = require('./shared.js');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 const OfflinePlugin = require('offline-plugin');
-const { env, publicPath } = require('./configuration.js');
+const { publicPath } = require('./configuration.js');
 const path = require('path');
+const { URL } = require('url');
```
unarist: WHATWG URL API is supported from nodejs 8.0 (as non-experimental), but our Dockerfile (and package.json) uses nodejs 6.x for now.

We need to use legacy URL API instead or upgrade those requirements.

unarist: -> #7665

PR #7500, opened 2018-05-15
Let navigator follow redirect instead that handling redirect in fetch

### LINKED ISSUE #7473: mastodon home screen app on iOS is not always working with service worker
When adding to home screen, the URL that is used to load the app is https://mastodon.social/web/timelines/home
When loading https://mastodon.social/web/timelines/home without being logged in, the service worker is returning a redirected response. As per fetch spec, this will fail the load.

Testing can be done on Mac/Safari or Mac/Chrome by doing the following:
  - Go to https://mastodon.social/ and log in
  - Make sure service worker is installed
  - Log out
  - Go to https://mastodon.social/web/timelines/home
It should go through the service worker and fail

mastodon service worker could be updated to check for the redirection in case of navigation loads.
Or the manifest.json could point to another URL that is not redirecting.

akihikodaki:
Resolves #7473 with #7498.

### DIFF
--- a/app/javascript/mastodon/service_worker/entry.js
+++ b/app/javascript/mastodon/service_worker/entry.js
@@ -10,7 +10,7 @@ function openWebCache() {
 }
 
 function fetchRoot() {
-  return fetch('/', { credentials: 'include' });
+  return fetch('/', { credentials: 'include', redirect: 'manual' });
 }
 
 const firefox = navigator.userAgent.match(/Firefox\/(\d+)/);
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
Sorry, conflicts :pray:

akihikodaki:
f8dd4524b60253e6d4052b1cbe5d52fbeae4ffd0: rebased to 17c1a62ec81e78f45c44882a5aba19ef8e7aeb44.

PR #7507, opened 2018-05-15
Update bot preference text

nightpool:
1. this change makes the wording more neutral—with the recent discourse around quote unquote russian bots, there's no reason to play into that and suggest that we need to be "warned", as if bots were a bad thing to have on your network.
   -  (a compromise here may be "flags this account as performing automated actions or behavior")
2. many accounts, like meetup, event, or brand accounts, do not represent persons but should not use the "bot" badge. see also colloquial use of "russian bots", as above, fabricated personas that don't represent people but aren't automated, something we should discourage calling a 'bot'. 
3. it gets to the heart of what we're trying to signal, automated interaction or posting.

see discussion on https://github.com/tootsuite/mastodon/pull/7391

### DIFF
--- a/config/locales/simple_form.en.yml
+++ b/config/locales/simple_form.en.yml
@@ -4,7 +4,7 @@ en:
     hints:
       defaults:
         avatar: PNG, GIF or JPG. At most 2MB. Will be downscaled to 400x400px
-        bot: Warns people that the account does not represent a person
+        bot: Indicates that this account exhibits automated behavior
         digest: Only sent after a long period of inactivity and only if you have received any personal messages in your absence
         display_name:
           one: <span class="name-counter">1</span> character left
--- a/config/locales/simple_form.en.yml
+++ b/config/locales/simple_form.en.yml
@@ -4,7 +4,7 @@ en:
     hints:
       defaults:
         avatar: PNG, GIF or JPG. At most 2MB. Will be downscaled to 400x400px
-        bot: Indicates that this account exhibits automated behavior
+        bot: This account mainly performs automated actions and might not be monitored
         digest: Only sent after a long period of inactivity and only if you have received any personal messages in your absence
         display_name:
           one: <span class="name-counter">1</span> character left


nightpool:
I'm liking "Indicates that this account performs automated actions or behavior" better but it's technically grammatically incorrect, can someone think of a rephrasing? i'm worried about "exhibits automated behavior" being too complex, especially for non-native speakers.

Kithop:
Trying to spitball other phrases...
"Indicates that this account posts automatically generated content"?
"This account's posted content and actions are controlled directly by software"?
"Performs autonomous activities without human intervention"?

nightpool:
it makes the text more neutral—"warns people" implies a moral judgement.

agirisan:
"Indicates that the account may be automated"?

Lana-chan:
i like how this was not the point being originally contested at all and we just moved the goalposts

Kithop:
It's another, valid thing that needs addressing and fixing, regardless.

Lana-chan:
ok, fair point and i agree with it, however i feel it is being used to ignore all other discussion that happened, and divert away from it

Gargron:
It's "bot" in discord, slack, IRC and other environments. It's a commonly used and understood term.

Gargron:
>it makes the text more neutral—"warns people" implies a moral judgement.

So the original intent of the whole thing is to *warn* people that the given account is not a person and will not understand interactions in a human way (e.g. Hey have you watched that episode of Westworld? What did you think of that?), but instead acts like, say, a newsbot or a markov chain bot (posts articles from an RSS feed, never replies), or as a tool bot (e.g. RemindMe, or one of the dice rollers, will reply only to special syntax)

Reword that as you will.

Gargron:
Also there is an outstanding issue for letting admins mark an account as a bot when the account holder fails to do so. There might be other features in the future that control bot interactions. Perhaps this checkbox should be moved out of "Edit profile" and into "Development" section.

nightpool:
brainstormed this some with gargron and we came up with:

> This account mainly performs automated actions and might not be monitored

thoughts?

Gargron (approved):

Issue #1326, opened 2017-04-09
Content warnings automatically added to replies


Kimik0:
When replying to a toot that has a content warning, a content warning is automatically added to the reply.

I understand that replies to sensitive content can be expected to also be sensitive, but it's rather confusing to have a content warning automatically added without user input or indication that one will be added.

I think this would be better handled instead at the receiver's UI end by not showing replies until the content warning is expanded. If this is considered to be a problem when boosts are taken into account, at least the UI should clearly indicate that a content warning will be added automatically

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

msappir:
Current behavior has a clear advantage though, in that by default it marks the entire thread in the same way, making it easy to continue ignoring a thread you're not interested in, or conversely following new responses to it if you are. Since CWs are intended for topics which not everyone wants to read about, this seems like a pretty good fit.
I also have to say I don't find the current behavior that confusing, but that's just me.

Gargron:
The CW is quite visibly pre-filled. It happens on UI level, not server level. You are free to just toggle the CW button to remove it.

lutoma:
While I agree with the mechanic of automatically marking replies with CWs, I think it could still be made clearer in the UI. When it happened to me the first time, I was also a bit confused until I realized what had happened.

MightyPork:
Many times I thought it copied the original toot and posted it instead of the reply, then promptly deleted it so as not to embarrass myself.

Now, I finally looked more closely and realized this is what's happening. What a bad feature!

ashfurrow:
@lutoma do you have any suggestions on ways to make the UI more clear?

ashfurrow:
It sounds like adding cw's automatically is part of the desired feature set, though I agree it could be made more clear. Anyone have any suggestions?

MightyPork:
I'm no designer, but it'd help if the tootbox more clearly indicated that CW is on

mock-up with a random CW from the FTL
![screenshot_20170508_201216](https://cloud.githubusercontent.com/assets/2041118/25818380/bd0f597e-342a-11e7-872e-b919fe94e888.png)

Artoria2e5:
+1 on reminding users that CW is on. 

The red frame appears a bit intrusive here… Bolding the text in the warning input may be a good idea. Or should we make TOOT button red and replace TOOT with BOOM?

thiagomgd:
[...]

PR #7508, opened 2018-05-15
 Put the CW field between the toot we are replying to and the toot field

ariasuni:
As explained in #1326, users may be confused by or forget about the CW field automatically field with the CW of the toot they are replying to; notably, the CW field may be very far away or no longer be visible when the toot is long.

The solution I propose is to put the CW field just above the toot being composed itself, so it’s always clear that there’s a CW (or not). It’s also more logical to put the CW nearer the toot than the toot being replied to.

Screenshots before/after:

screenshot_2018-05-16 mastodon devscreenshot_2018-05-16 mastodon dev after


ClearlyClaire (commented):
Why remove the Collapsable component/animation?

Gargron:
Spacing looks wrong.

ClearlyClaire (approved):

ghost (approved):

Gargron:
Yes, I do think the spacing needs change.

Gargron:
I also haven't seen how the animation looks now. I'm not so sure: spring-based animation is generally more interesting to look at than CSS transitions. I do not agree with the arguments about style in the markup, we use Motion throughout the application so it's more consistent for us to have that than CSS transitions.

ariasuni:
The problem I had with height of the element in the markup is that I couldn’t modify the size of the spacing under the CW easily with my custom CSS extension (I assume it’s the same for themes).

ariasuni:
Here’s how it looks now:

screenshot_2018-05-31 mastodon dev

ariasuni:
I also kept the CSS animation, since I didn’t notice any difference visible difference with the Spring one, and for the easier theming of margins.

PR #7513, opened 2018-05-16
French translation tweaks

Miaourt:
- Use more neutral terminology for readability.
- Maintains the use of "Personne" instead of "Utilisateur⋅ice" to emphasis on the "You're not the product".
- Some fixes.


trebmuh:
Nice catches IMHO.

ykzts (approved):

PR #7521, opened 2018-05-17
Improve payload format of Web Push API now that it's open

Gargron:
> Good lord what is happening in there

Previously the contents of the Web Push API payloads closely resembled the structure of JavaScript's [Notification](https://developer.mozilla.org/en-US/docs/Web/API/Notification). But now that the API is open to non-browser apps, and given that there is no required coupling between contents of the payload and a Notification object, here is how I changed the payload:

```json
{ 
  "access_token": "...",
  "preferred_locale": "en",
  "notification_id": "12345",
  "notification_type": "follow",
  "title": "So and so followed you",
  "body": "This is my bio",
  "icon": "https://example.com/avatar.png"
}
```

The title, body and icon attributes are included as a fallback so you can construct a minimal notification if you cannot perform a network request to the API to get more data.

The access_token is included because that's the only way to make API requests from the service worker in JS.

I have changed the service worker accordingly. One challenge was i18n, since previously the pre-localized string was part of the payload. I have solved this by pre-evaluating a special file with only notification-relevant strings. It's not too big and contains all we need. Now push notification i18n in the webapp is done properly rather than relying on the server to send strings.


nightpool:
it sounds like (from feedback in discord) that putting more information in a single notification payload makes more sense because mobile applications may not have the resources or access to make additional requests in the background. accordingly we should probably render the full notification resource inline rather then relying on the app to be able to make subsequent requests

the locale changes look good though

tateisu:
(copy from discord)
I believe that there are developers who do not like to exclude content to be displayed from the payload. Mobile applications are not always communicable.  ST is compromising on real-time performance.

sorry weird translation, "Mobile applications are not always communicable. " means network connectivity. if payload has content data that required to show notification, app can show it without network connection that is often limited by the OS.

When the background application requests network access, OS's Job Scheduler imposes restrictions on its execution interval. This may cause delays. Since my application may handle a large number of accounts, it is not so concerned with real-time nature. But other apps may target higher responsiveness.

---

I think that the access token itself is unnecessary. Even though the application server can assert security without receiving an access token at all, it is not good to include an access token in the payload.

I think that Gargron's PR includes an access token because it assumes that the application server requests details of the notification. I want you to doubt our common sense a bit. Do you even think that we will not misuse the access token?

---
In fact the fixed data required by the application server can be included in the endpoint URL. my app's endpoint url contains FCM's device ID, user full acct (for multi account), alerts flags, digested account account identifier that binded to server public key on app server.

---

The current WebPush payload contains the notification ID in the tag part, but is it possible to have it in the HTTP header too? This means that the application server knows the notification ID without decrypting it, but I think that it is almost safe because there is no public API that can get notifications

---
What can be a problem with those applications is the size constraint of the FCM or APN payload. The endpoint callback receives the encrypted content body and several HTTP headers necessary for decryption. All of them and additional information peculiar to the application must be stored in JSON and have a size of 3 k or less. Encoding the encrypted data with Base64 increases the content body to 133%.
( Z85 encoding can be used to decrease overhead to 125% )

About locale. Applications and OS have language specification, but I think that it is a good idea to have a "preferred_locale" hint. Instead of sending the entire notification message, you should send information such as the type of notification, ellipsized message body, and the sender of the notification so that the application side can construct the notification message.

tateisu:
When I send 1 chars mention, the endpoint callback receives 836 bytes request body.
When I send 500 chars mention, the endpoint callback receives 1356 bytes request body.
When I send 5000 chars mention (from Pleroma), the endpoint callback receives 5556 bytes request body.

We can't redirect over about 2k data to FCM, APN without decrypting and spoiling on app server.
It's not good for security. 
Please ellipsize menton's message content, then we have a choice to make app server redirects payload without decryption and client app only can decrypt it.

DagAgren:
For my client, I can do a request to get the real notification, but there are limitations. For instance, if the request gets delayed by more than 30 seconds, the default notification will be shown anyway.

At least on my side I can dig out the access token, so I do not need that, although it is convenient to have there.

I think I agree with @tateisu that it still makes sense to include the minimal data required to construct a notification without network access. That would mean the title, the body (possibly pre-shortened), any image, and a flag for sensitive material.

Once the user opens the notification, it would also be good to have the notification type and target id, to be able to predict where to navigate in the app.

nightpool (changes requested):
Sorry, reviewed this earlier but i was on mobile and didn't realize github wouldn't submit my comments

LGTM with some serviceworker changes

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -1,36 +1,32 @@
...
 
       return self.registration.showNotification(group.title, group);
-    } else if (notifications.length === 1 && notifications[0].tag === GROUP_TAG) {
-      // Already grouped, proceed with appending the notification to the group
-      const group = cloneNotification(notifications[0]);
+    } else if (notifications.length === 1 && notifications[0].tag === GROUP_TAG) { // Already grouped, proceed with appending the notification to the group
+      const group = { ...notifications[0] };
 
-      group.title = formatGroupTitle(group.data.message, group.data.count + 1);
+      group.title = formatMessage('notifications.group', options.data.preferred_locale, { count: notifications.length + 1 });
```
nightpool: notifications.length is always 1 here, it should be group.data.count

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -40,57 +36,81 @@ const notify = options =>
...
+
+  // Placeholder until more information can be loaded
+  event.waitUntil(notify({
+    title,
+    body,
+    icon,
+    tag: notification_id,
+    badge: '/badge.png',
+    url: '/web/notifications',
+  }));
```
nightpool: there's no `data: {preferred_locale}` here, so it'll cause an undefined error if you try to roll it up.

nightpool: i think it also needs a timestamp for the same reason, although we can just use Date.new()

nightpool (approved):

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -40,57 +36,82 @@ const notify = options =>
...
+  (new IntlMessageFormat(locales[locale][messageId], locale)).format(values);
 
-  for(var k in notification) {
-    clone[k] = notification[k];
-  }
+const handlePush = (event) => {
+  const { access_token, notification_id, preferred_locale, title, body, icon } = event.data.json();
+
+  // Placeholder until more information can be loaded
+  event.waitUntil(notify({
```
sorin-davidoi: @Gargron I don't think you can call `event.waitUntil` multiple times (or what is the behavior if you do). Something like this might be better suited:

```js
event.waitUntil(
    notify({ ... })
        .then(() => fetchFromApi(...))
        .then(notification => notificationToOptions(notification))
        .then(options => notify(options);
);
```

This way `event.waitUntil` is called only once.

nightpool: what is the reason we can't use async/await in service workers again?

nightpool: The waitUntil() method must be initially called within the event callback, but after that it can be called multiple times, until all the promises passed to it settle.

Gargron: I read the above sentence from the docs multiple times but was still confused by it.

nightpool: i think the intention is "waitUntil must be called at least once but can be called possibly more times"

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -40,57 +36,81 @@ const notify = options =>
...
+  return fetch(url, {
+    headers: {
+      'Authorization': `Bearer ${accessToken}`,
+      'Content-Type': 'application/json',
+    },
 
-  event.waitUntil(notify(options));
+    method: method,
+    credentials: 'include',
+  });
```
sorin-davidoi: `fetch` returns a stream and it throws only if there is a network error (e.g. and not if the endpoint return `400`).

```js
fetch(...)
   // Make sure to reject if the request did not succeed
   .then(res => res.ok ? res : throw new Error(res.status)
   // Convert the stream into JSON
   .then(res => res.json())
```

PR #7525, opened 2018-05-17
HTTP signatures spec no longer requires algorithms field

### LINKED ISSUE #7520: Update HTTP Signatures To Be Compliant With Draft 10
Draft 10 was released a day or two ago; going through the [diff](https://tools.ietf.org/rfcdiff?difftype=--hwdiff&url2=draft-cavage-http-signatures-10.txt) the `algorithm` parameter is now optional and implementations are expected to infer the correct algorithm based on the `keyId` metadata.

This means Mastodon may theoretically obtain HTTP signatures without an `algorithms` parameter. I'm not a ruby person so my limited understanding of the code [here](https://github.com/tootsuite/mastodon/blob/master/app/controllers/concerns/signature_verification.rb#L112) makes it seem like it is incompatible.

I had been talking to @nightpool about HTTP Signatures `algorithms` literally a few days ago, which is the only reason I noticed this.

For reference, the spec discussion I have about this is [here](https://github.com/w3c-dvcg/http-signatures/issues/29#issuecomment-389627529).

* * * *

Gargron:
Fix #7520

### DIFF
--- a/app/controllers/concerns/signature_verification.rb
+++ b/app/controllers/concerns/signature_verification.rb
@@ -107,9 +107,7 @@ module SignatureVerification
 
   def incompatible_signature?(signature_params)
     signature_params['keyId'].blank? ||
-      signature_params['signature'].blank? ||
-      signature_params['algorithm'].blank? ||
-      signature_params['algorithm'] != 'rsa-sha256'
+      signature_params['signature'].blank?
   end
 
   def account_from_key_id(key_id)


nightpool:
Okay, although this leaves us in a weird state where we just assume the signature is RSA-256, without verifying it. I think that's just a implementation niceness thing though, it doesn't seem like a security risk.

nightpool (approved):

msporny:
> Okay, although this leaves us in a weird state where we just assume the signature is RSA-256, without verifying it. I think that's just a implementation niceness thing though, it doesn't seem like a security risk.

This is not an issue until you start supporting other signature formats, like ed25519-based ones... once that happens, you will want to check the key metadata to make sure you're using a key designated for the purpose of HTTP Signature-based authentication. You're good for now, but just giving you a heads up on "The Right Way"(tm) to do this in the future.

PR #7526, opened 2018-05-17
Resolve unknown status from Add activity, skip Remove if unknown

### LINKED ISSUE #7518: Add/Remove activity causes error if the target status is not known
```
NoMethodError: undefined method `account_id' for nil:NilClass

/mastodon/app/lib/activitypub/activity/remove.rb:9:in `perform'
/mastodon/app/services/activitypub/process_collection_service.rb:43:in `process_item'
```

and same error for Add activity.

* * * *

Gargron:
Fix #7518

### DIFF
--- a/app/lib/activitypub/activity/add.rb
+++ b/app/lib/activitypub/activity/add.rb
@@ -4,9 +4,10 @@ class ActivityPub::Activity::Add < ActivityPub::Activity
   def perform
     return unless @json['target'].present? && value_or_id(@json['target']) == @account.featured_collection_url
 
-    status = status_from_uri(object_uri)
+    status   = status_from_uri(object_uri)
+    status ||= fetch_remote_original_status
 
-    return unless status.account_id == @account.id && !@account.pinned?(status)
+    return unless !status.nil? && status.account_id == @account.id && !@account.pinned?(status)
 
     StatusPin.create!(account: @account, status: status)
   end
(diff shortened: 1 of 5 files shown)


app/lib/activitypub/activity/add.rb
```diff
@@ -4,9 +4,10 @@ class ActivityPub::Activity::Add < ActivityPub::Activity
   def perform
     return unless @json['target'].present? && value_or_id(@json['target']) == @account.featured_collection_url
 
-    status = status_from_uri(object_uri)
+    status   = status_from_uri(object_uri)
+    status ||= fetch_remote_original_status
```
ThibG: It's… not defined here? How come the tests didn't catch that? Am I missing anything?

ClearlyClaire (changes requested):

app/lib/activitypub/activity/add.rb
Gargron: The tests assume the referenced status was already known. And that's kind of the more likely scenario because only receive Add/Remove if you follow, and if you follow, you get new posts too. I think this case is just of someone pinning an old, pre-follow post.

ThibG: I meant, I `fetch_remote_original_status` is not defined in this part of the code, afaict, I'm surprised the continuous integration tooling haven't caught that, but I understand the tests themselves did not cover this case at all.

Gargron: OH. Oops. I see what you mean. Somehow thought that was defined on the Activity superclass, because I remember that's how it's structured in the OStatus equivalent :weary:

Gargron: @ThibG It's not caught by the test because that line would only be executed if status_from_uri returned nil, which in the test, it does not

Gargron:
@ThibG Fixed and added test coverage

ClearlyClaire (approved):

PR #7528, opened 2018-05-17
exclude-other-silenced-accounts

masarakki:
Now, silenced users can show other silenced user's statuses, why is it necessary?

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -296,7 +296,8 @@ class Status < ApplicationRecord
 
     def account_silencing_filter(account)
       if account.silenced?
-        including_silenced_accounts
+        including_myself = left_outer_joins(:account).where(account_id: account.id).references(:accounts)
+        excluding_silenced_accounts.or(including_myself)
       else
         excluding_silenced_accounts
       end
--- a/spec/models/status_spec.rb
+++ b/spec/models/status_spec.rb
@@ -537,17 +537,6 @@ RSpec.describe Status, type: :model do
           expect(results).to include(es_status)
         end
       end
-
-      context 'where that account is silenced' do
-        it 'includes statuses from other accounts that are silenced' do
-          @account.update(silenced: true)
-          other_silenced_account = Fabricate(:account, silenced: true)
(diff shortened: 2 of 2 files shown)


nightpool:
>Now, silenced users can show other silenced user's statuses, why is it necessary?

what do you mean by this? isn't this change actually the opposite? i.e. it prevents silenced users from seeing other silenced user's statuses?

---

```ruby
 217/536 |======== 40 ========>                                |  ETA: 00:00:38 

  1) Status.as_public_timeline with an account passed in where that account is silenced includes statuses from other accounts that are silenced
     Failure/Error: expect(results).to include(other_status)
     
       expected #<ActiveRecord::Relation []> to include #<Status id: 100044383610076677, uri: "https://cb6e6126.ngrok.io/users/dejuan_schuppe188/...", text: ...onversation_id: 274, local: true, account_id: 312, application_id: nil, in_reply_to_account_id: nil>
       Diff:
       @@ -1,2 +1,2 @@
       -[#<Status id: 100044383610076677, uri: "https://cb6e6126.ngrok.io/users/dejuan_schuppe188/...", text: "Lorem ipsum dolor sit amet", created_at: "2018-05-17 11:22:26", updated_at: "2018-05-17 11:22:26", in_reply_to_id: nil, reblog_of_id: nil, url: nil, sensitive: false, visibility: "public", spoiler_text: "", reply: false, favourites_count: 0, reblogs_count: 0, language: nil, conversation_id: 274, local: true, account_id: 312, application_id: nil, in_reply_to_account_id: nil>]
       +[]
       
     # ./spec/models/status_spec.rb:548:in `block (5 levels) in <top (required)>'
```

you'll need to fix the failing test before merge

Gargron:
@nightpool No, the change is somewhat correct. Our current behaviour is that silenced users only see other silenced users in the public timelines. It's like a hellban. This PR removes that behaviour.

takayamaki:
Are there any other concerns?

Gargron (approved):

PR #7531, opened 2018-05-17
User agent for WebFinger

### LINKED ISSUE #6294: Webfinger requests do not contain hostname in user agent
As in title. they just say `http.rb/2.2.2` in the user agent. so if a webfinger fails, I will not know which instance tried to find a specific user.

* * * *

hcmiya:
closes #6294

### DIFF
--- a/lib/mastodon/version.rb
+++ b/lib/mastodon/version.rb
@@ -50,7 +50,7 @@ module Mastodon
     end
 
     def user_agent
-      @user_agent ||= "#{HTTP::Request::USER_AGENT} (Mastodon/#{Version}; +#{Rails.configuration.x.use_https ? 's' : ''}://#{Rails.configuration.x.web_domain}/)"
+      @user_agent ||= "#{HTTP::Request::USER_AGENT} (Mastodon/#{Version}; +http#{Rails.configuration.x.use_https ? 's' : ''}://#{Rails.configuration.x.web_domain}/)"
     end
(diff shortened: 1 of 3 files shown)


lib/mastodon/version.rb
```diff
@@ -48,5 +48,9 @@ def source_url
         source_base_url
       end
     end
+
+    def user_agent
+      @user_agent ||= "#{HTTP::Request::USER_AGENT} (Mastodon/#{Version}; +http#{Rails.configuration.x.use_https ? 's' : ''}://#{Rails.configuration.x.local_domain}/)"
```
ThibG: Why not use `root_url`? Also, this is going to link to e.g., https://sitedethib.com instead of https://social.sitedethib.com, (`WEB_DOMAIN` vs. `LOCAL_DOMAIN`) which isn't a huge deal I guess, but it still feels a bit weird.

ClearlyClaire (changes requested):

lib/mastodon/version.rb
hcmiya: `root_url` requires `RoutingHelper` but not loaded in this scope.
I will rewrite to use WEB_DOMAIN...

Gargron (approved):

ClearlyClaire (approved):

Gargron:
CircleCI not working here..?

PR #7532, opened 2018-05-17
Add preference to hide following/followers lists

### LINKED ISSUE #6901: Hide followers and follows to protect user's privacy
I wonder if user profiling can be done at low cost from the user's follows or followers list.

It should be possible to automate the build a social graph and list the interests of a user from this.
Has it be done already on twitter data for example?

I guess for example determining if the user is gay, what political side he is, etc... can be trivial to get.

* * * *
- [x ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
- Public pages
- ActivityPub collections (does not return pages but does give total)
- REST API (unless it's your own) (does not federate)

image

image

Fix #6901

### DIFF
--- a/app/controllers/api/v1/accounts/follower_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/follower_accounts_controller.rb
@@ -19,7 +19,7 @@ class Api::V1::Accounts::FollowerAccountsController < Api::BaseController
   end
 
   def load_accounts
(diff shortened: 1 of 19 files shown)


ykzts (approved):

trwnh:
does this not hide the follower/following counts as well?

PR #7541, opened 2018-05-18
Do not raise delivery failure on 4xx errors, increase stoplight threshold

Gargron:
Stoplight failure threshold from 3 to 10
Status code 429 will raise a failure/get retried


nightpool:
where is the 429 coming from?

Gargron:
>429 Too Many Requests
>The user has sent too many requests in a given amount of time ("rate limiting")

While I don't think Mastodon has rate limits on the inbox by default, it's possible to add that using Mastodon code and it's also quite likely to occur in other software implementations down the line.

nightpool:
ohh okay i get it now. 429 *should* trigger the stoplight error, but 404/409 *shouldn't*

nightpool (approved):

PR #7546, opened 2018-05-19
Improve direct message warning again

Gargron:
Ref: https://github.com/tootsuite/mastodon/pull/7420#pullrequestreview-120703069

image

"Learn more" link opens privacy policy in new tab. Relevant bit is at the top of the policy:

image


ProgVal:
Replace "dangerous" with "sensitive"?

PR #7555, opened 2018-05-19
[RU] Updated strings for 2.4.0

MasterGroosha:
Updated and added some strings for 2.4.0

### DIFF
--- a/config/locales/ru.yml
+++ b/config/locales/ru.yml
@@ -40,6 +40,7 @@ ru:
     following: Подписан(а)
     media: Медиаконтент
     moved_html: "%{name} переехал(а) на %{new_profile_link}:"
+    network_hidden: Эта информация недоступна
     nothing_here: Здесь ничего нет!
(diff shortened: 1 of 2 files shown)


config/locales/ru.yml
```diff
@@ -80,7 +81,7 @@ ru:
       domain: Домен
       edit: Изменить
       email: E-mail
-      email_status: Статус электронной почты
+      email_status: Статус e-mail
```
Gargron: Why this?

MasterGroosha: @Gargron it says literally the same ("e-mail" means "электронная почта"), is a common word (it's well known word even for those Russians who don't know English) and has the same length as English version so this string won't be cut if string is viewed on narrow screen.

If you don't agree with my point of view, I can revert this string to original state.

Gargron: I do speak Russian but I was wondering about the reasoning. To make it shorter makes sense.

Gargron (approved):

PR #7562, opened 2018-05-20
Use LoadingBar instead of blur to indicate that an image is loading (fixes #4060)

### LINKED ISSUE #4060: Blurred image previews remove loading feedback

ClearlyClaire:
It doesn't change how images are loaded (maybe using `srcset` would be better?), but removes the blur on the preview and uses a fake loading bar instead.

### DIFF
--- a/app/javascript/mastodon/features/ui/components/image_loader.js
+++ b/app/javascript/mastodon/features/ui/components/image_loader.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { LoadingBar } from 'react-redux-loading-bar';
 import ZoomableImage from './zoomable_image';
 
 export default class ImageLoader extends React.PureComponent {
@@ -23,6 +24,7 @@ export default class ImageLoader extends React.PureComponent {
   state = {
     loading: true,
     error: false,
+    width: null,
   }
 
   removers = [];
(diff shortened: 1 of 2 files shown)


ThibG:
I had trouble purposefully triggering a slow load without messing with the cache, so the thumbnail doesn't actually load in the following screencast, but here is what it looks like…

pr

(https://sitedethib.com/77280d9c9593045609df3eba2f412d92e32cd454.mp4)

ThibG:
Actually there are a few issues left:
- The 3px loading bar is not displayed over the image, but above it (not a huge deal, but it “eats”/move the image down by 3px)
- When the image is fully loaded, the loading bar is briefly displayed at the very top of the screen (this is because the canvas doesn't take 100% of the available space, but the `ZoomableImage` div does)

So it may need some more tweaks, and I may need help with that, the CSS here already involves a lot of stuff I'm not 100% used to.

Gargron (approved):

PR #7569, opened 2018-05-21
Unescape HTML entities in rich web push notifications in web UI

Gargron:


ykzts (approved):

app/javascript/mastodon/service_worker/web_push_notifications.js
```diff
@@ -1,5 +1,6 @@
 import IntlMessageFormat from 'intl-messageformat';
 import locales from './web_push_locales';
+import { unescape } from 'lodash';
```
sorin-davidoi: Does this get tree-shaken or does it pull in the entire package?

ykzts: https://github.com/tootsuite/mastodon/blob/5ea643b27908b14bd89ff068fc87e446e8cbcd32/.babelrc#L38

[`babel-plugin-lodash`](https://github.com/lodash/babel-plugin-lodash) is used.

PR #7570, opened 2018-05-21
Use #any? instead of #exists? when checking media attachments

Gargron:
If media_attachments are not loaded, SQL query is the same, but the #exists? method performs SQL query even if preloaded


ykzts (approved):
🙇

PR #7571, opened 2018-05-21
Use real container width in MediaGallery srcSet

### LINKED ISSUE #7568: Responsive thumbnail sizes incorrect unfit for public pages
Thumbnails correctly use `srcSet`, but there are multiple issues with the `sizes` attribute:
- The `(min-width: 1025px)` selector is arbitrary and does not seem to fit any existing logic
- `${320 * (width / 100)}px` assumes the image is going to be displayed in a 330px-wide column, while toots in public views are much wider, making the browser select the undersized thumbnail while the full-sized image would be much more fitting

* * * *

Gargron:
Fix #7568

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -20,6 +20,7 @@ class Item extends React.PureComponent {
     index: PropTypes.number.isRequired,
     size: PropTypes.number.isRequired,
     onClick: PropTypes.func.isRequired,
+    displayWidth: PropTypes.number,
   };
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


app/javascript/mastodon/components/media_gallery.js
```diff
@@ -263,9 +264,9 @@ export default class MediaGallery extends React.PureComponent {
       const size = media.take(4).size;
 
       if (this.isStandaloneEligible()) {
-        children = <Item standalone onClick={this.handleClick} attachment={media.get(0)} />;
+        children = <Item standalone onClick={this.handleClick} attachment={media.get(0)} displayWidth={width} />;
```
ThibG: It appears to work, but I'm not sure how, as `width` is undefined when it's rendered for the first time.

Gargron: Well, not sure what we can do there. Hardcode a fallback width? That's back to status quo with ~330px

ThibG: It's kind of redundant with the CSS, but we can always know which size it is going to be, from the context. That would be best. But apparently it works as is, so… :man_shrugging:

PR #7575, opened 2018-05-21
Save onlyMedia prop when pinning column

Gargron:
Ref: https://github.com/tootsuite/mastodon/pull/6631#issuecomment-390677846

Opening public timeline in multiple columns even with different param seems to lead to the previous one being disconnected somehow.


ykzts:
I think that the same correspondence is necessary also in the `community_timeline`.

ykzts (approved):

PR #7594, opened 2018-05-22
Remove pointer to light theme until it is more polished

Gargron:
Sorry @Sylvhem, there are some contrast issues with the light theme. 2.4 is already at least a week behind schedule so I need to hurry up, and rather than release an unfinished theme we'll just remove the reference here and put it back in in 2.4.1 later down the line.


ykzts (approved):

lynlynlynx:
Thanks for giving us a time to fix it. I have not checked the new theme at all yet :cold_sweat: .

PR #7598, opened 2018-05-23
Enable media timeline

ykzts:


Gargron (approved):

MightyPork:
I feel like this should be toggleable in the settings, or by the admin. It may be valuable for art-focused instances [[1]](https://elekk.xyz/@noelle/100152970949082387), but otherwise it makes the interface more cluttered without much gain (i don't see ever using it, and will hide it with a userstyle)

screenshot_20180605_174125

trwnh:
I would prefer making it an option for all columns, within the existing column settings. It's kind of weird that there isn't a media filter on the Home timeline, and I agree it makes the column headers look unaligned.

Gargron:
Media filter on home timeline is not so easily possible as on the public ones.

PR #7614, opened 2018-05-24
optimize direct timeline

tateisu:
The direct timeline is mix of DM from user and DM to user.
This PR separates query to pair of _from_me and _to_me, then merge after querying for performance.

Also this PR does:

- Fix filter condition. old implementation use filters for public timeline, but this TL is NOT public.
- Fix bug in rspec test.
- Optimize pagination of _to_me part that using mentions.status_id instead of statuses.id
- Prevent conversion of constant condition into bind parameter. this is required to use partial index when  prepared statement is enabled.

Temporary revoked:

- Add partial indexes.
- Add mentions.direct column that is used in partial index.

----
some information that is NOT included in this PR.

using union for merge 2 query on DB side
https://gist.github.com/tateisu/a7de7a8ff816e83ea2894e599499849d
- too hard to pagination. 

adding partial index for statuses
https://gist.github.com/tateisu/cc6bff006b898094262245491b631f2f
- this query still scan all DM and then filter it.


akihikodaki:
Thank you for your effort to address the issue.

Query and index optimization are for different purposes, and if you see improvements in both, probably you would need index optimization too. In case of your query, `_from_me` would take much if an account has few direct statuses without an index.

I tested an index after some discussion with you on Mastodon. In case of `@173210@o.kagucho.net`, though, an index greatly contributed to the performance and I do not see any room of improvement in the query plan.

Before applying an index:
```
mastodon=> explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
mastodon-> LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id AND mentions.account_id = 1    
mastodon-> LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
mastodon-> WHERE (mentions.account_id = 1 OR statuses.account_id = 1) AND "statuses"."visibility" = 3  
mastodon-> AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;
                                                                                       QUERY PLAN                                                     
                                  
------------------------------------------------------------------------------------------------------------------------------------------------------
----------------------------------
 Limit  (cost=18989.38..18989.38 rows=1 width=16) (actual time=6370.358..6370.369 rows=40 loops=1)
   ->  Sort  (cost=18989.38..18989.38 rows=1 width=16) (actual time=6363.888..6363.895 rows=40 loops=1)
         Sort Key: statuses.id DESC
         Sort Method: top-N heapsort  Memory: 26kB
         ->  Nested Loop  (cost=684.16..18989.37 rows=1 width=16) (actual time=357.594..6357.582 rows=123 loops=1)
               ->  Hash Left Join  (cost=683.87..18984.03 rows=1 width=24) (actual time=357.558..6353.799 rows=123 loops=1)
                     Hash Cond: (statuses.id = mentions.status_id)
                     Filter: ((mentions.account_id = 1) OR (statuses.account_id = 1))
                     Rows Removed by Filter: 7
                     ->  Index Only Scan using index_statuses_20180106 on statuses  (cost=0.42..18293.50 rows=145 width=24) (actual time=1.440..5994.2
96 rows=130 loops=1)
                           Index Cond: (visibility = 3)
                           Heap Fetches: 130
                     ->  Hash  (cost=680.05..680.05 rows=272 width=16) (actual time=356.077..356.077 rows=272 loops=1)
                           Buckets: 1024  Batches: 1  Memory Usage: 21kB
                           ->  Bitmap Heap Scan on mentions  (cost=10.53..680.05 rows=272 width=16) (actual time=13.166..355.843 rows=272 loops=1)
                                 Recheck Cond: (account_id = 1)
                                 Heap Blocks: exact=101
                                 ->  Bitmap Index Scan on index_mentions_on_account_id_and_status_id  (cost=0.00..10.46 rows=272 width=0) (actual time
=13.130..13.130 rows=272 loops=1)
                                       Index Cond: (account_id = 1)
               ->  Index Scan using accounts_pkey on accounts  (cost=0.29..5.32 rows=1 width=8) (actual time=0.029..0.029 rows=1 loops=123)
                     Filter: (NOT silenced)
 Planning time: 36.560 ms
 Execution time: 6373.108 ms
(24 rows)
```

Applying the index:
```
mastodon=> CREATE INDEX test ON statuses (account_id, id) WHERE visibility = 3;
CREATE INDEX
```

After applying the index and restarting PostgreSQL to invalidate caches on memory:
```
mastodon=> explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id AND mentions.account_id = 1    
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE (mentions.account_id = 1 OR statuses.account_id = 1) AND "statuses"."visibility" = 3  
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;
                                                                                    QUERY PLAN                                                        
                            
------------------------------------------------------------------------------------------------------------------------------------------------------
----------------------------
 Limit  (cost=1249.11..1249.12 rows=1 width=16) (actual time=763.299..763.306 rows=40 loops=1)
   ->  Sort  (cost=1249.11..1249.12 rows=1 width=16) (actual time=763.298..763.301 rows=40 loops=1)
         Sort Key: statuses.id DESC
         Sort Method: top-N heapsort  Memory: 26kB
         ->  Nested Loop  (cost=573.50..1249.10 rows=1 width=16) (actual time=603.562..763.148 rows=123 loops=1)
               ->  Hash Right Join  (cost=573.22..1243.77 rows=1 width=24) (actual time=586.983..688.652 rows=123 loops=1)
                     Hash Cond: (mentions.status_id = statuses.id)
                     Filter: ((mentions.account_id = 1) OR (statuses.account_id = 1))
                     Rows Removed by Filter: 7
                     ->  Bitmap Heap Scan on mentions  (cost=10.53..680.05 rows=272 width=16) (actual time=10.186..114.667 rows=272 loops=1)
                           Recheck Cond: (account_id = 1)
                           Heap Blocks: exact=101
                           ->  Bitmap Index Scan on index_mentions_on_account_id_and_status_id  (cost=0.00..10.46 rows=272 width=0) (actual time=10.15
6..10.156 rows=272 loops=1)
                                 Index Cond: (account_id = 1)
                     ->  Hash  (cost=560.87..560.87 rows=145 width=24) (actual time=573.690..573.690 rows=130 loops=1)
                           Buckets: 1024  Batches: 1  Memory Usage: 16kB
                           ->  Bitmap Heap Scan on statuses  (cost=8.83..560.87 rows=145 width=24) (actual time=3.466..573.501 rows=130 loops=1)
                                 Recheck Cond: (visibility = 3)
                                 Heap Blocks: exact=89
                                 ->  Bitmap Index Scan on test  (cost=0.00..8.79 rows=145 width=0) (actual time=0.055..0.055 rows=130 loops=1)
               ->  Index Scan using accounts_pkey on accounts  (cost=0.29..5.32 rows=1 width=8) (actual time=0.605..0.605 rows=1 loops=123)
                     Filter: (NOT silenced)
 Planning time: 393.201 ms
 Execution time: 763.777 ms
(25 rows)
```

Be aware that the account does not have that much direct statuses. It should use merge join instead of hash join for accounts with many direct statuses, but I'm not sure whether PostgreSQL planner is wise enough. The index should also be tested on accounts with many direct statuses. If not, this change would help PostgreSQL adopting a more optimal plan.

app/models/status.rb
```diff
@@ -196,6 +196,21 @@ def as_direct_timeline(account)
       apply_timeline_filters(query, account, false)
     end
 
+    def as_direct_timeline_from_me(account)
+      query = where("statuses.account_id = #{account.id}")
```
abcang: This is better

```ruby
query = where(account_id: account.id)`

File: app/models/status.rb [hunk 2]
```diff
@@ -196,6 +196,21 @@ def as_direct_timeline(account)
...
+              .where(visibility: [:direct])
+
+      apply_timeline_filters(query, account, false)
+    end
+
+    def as_direct_timeline_to_me(account)
+      query = joins("LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id AND mentions.account_id = #{account.id}")
+              .where("mentions.account_id = #{account.id}")
```
abcang: Duplicate conditions of `mentions.account_id`

app/models/status.rb
```diff
@@ -196,6 +196,21 @@ def as_direct_timeline(account)
...
 
+    def as_direct_timeline_from_me(account)
+      query = where("statuses.account_id = #{account.id}")
+              .where(visibility: [:direct])
+
+      apply_timeline_filters(query, account, false)
+    end
+
+    def as_direct_timeline_to_me(account)
+      query = joins("LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id AND mentions.account_id = #{account.id}")
```
abcang: This is better

```ruby
query = Status.joins(:mentions).where(Mention.arel_table[:account_id].eq(account.id))
```

app/controllers/api/v1/timelines/direct_controller.rb
```diff
@@ -19,19 +19,36 @@ def load_statuses
   end
 
   def cached_direct_statuses
-    cache_collection direct_statuses, Status
+    a1 = cache_collection direct_statuses_from_me, Status
+    a2 = cache_collection direct_statuses_to_me, Status
+    (a1 + a2)
+      .sort { |a, b| b.id <=> a.id }
+      .uniq(&:id)
+      .slice(0, limit_param(DEFAULT_STATUSES_LIMIT))
```
abcang: Using `take` is simple.

```ruby
.take(limit_param(DEFAULT_STATUSES_LIMIT))
```

app/controllers/api/v1/timelines/direct_controller.rb
```diff
@@ -19,19 +19,36 @@ def load_statuses
   end
 
   def cached_direct_statuses
-    cache_collection direct_statuses, Status
+    a1 = cache_collection direct_statuses_from_me, Status
+    a2 = cache_collection direct_statuses_to_me, Status
+    (a1 + a2)
+      .sort { |a, b| b.id <=> a.id }
```
abcang: It is better to use `sort_by`.

```ruby
.sort_by(&:id).reverse
```

abcang (commented):
`as_direct_timeline` is unnecessary and should be deleted.

tateisu:
@abcang,  all fixed.

@akihikodaki , 
Of course I agree that adding an index is necessary. However, it should be tested in several instances after this PR and then included in another PR.

I do not agree with "There is no room for improvement in the query plan." Merge Join is used in queries before applying this PR, but it Sorts the rows that is much larger than limit. There is no sort of mentions in the query after applying this PR. The number of unused rows is somewhat reduced.

It is not included in this PR, but if you devise more,  you can add the pagenation condition not only for statuses.id but also for mentions.status_id . You may be able to reduce the unused rows in scanning mentions.

pagenation to statuses.id
```
SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE mentions.account_id = 1  AND "statuses"."visibility" = 3  
and statuses.id < 99722268340454164
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;

Limit  (cost=0.99..1444.30 rows=1 width=16) (actual time=5.342..20.664 rows=40 loops=1)
  ->  Nested Loop  (cost=0.99..1444.30 rows=1 width=16) (actual time=5.332..20.362 rows=40 loops=1)
        ->  Nested Loop  (cost=0.70..1438.62 rows=1 width=24) (actual time=5.262..19.030 rows=40 loops=1)
              ->  Index Only Scan Backward using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..393.52 rows=161 width=8) (actual time=0.288..7.879 rows=724 loops=1)
                    Index Cond: (account_id = 1)
                    Heap Fetches: 629
              ->  Index Only Scan using statuses_dm on statuses  (cost=0.28..6.48 rows=1 width=24) (actual time=0.006..0.007 rows=0 loops=724)
                    Index Cond: ((id = mentions.status_id) AND (id < '99722268340454164'::bigint))
                    Heap Fetches: 24
        ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..5.67 rows=1 width=8) (actual time=0.014..0.018 rows=1 loops=40)
              Heap Fetches: 38
Planning time: 1.906 ms
Execution time: 21.050 ms
```

pagenation to statuses.id and mentions.status_id
```
SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE mentions.account_id = 1  AND "statuses"."visibility" = 3  
and statuses.id < 99722268340454164 and mentions.status_id < 99722268340454164
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;

Limit  (cost=0.99..1195.63 rows=1 width=16) (actual time=1.139..15.925 rows=40 loops=1)
  ->  Nested Loop  (cost=0.99..1195.63 rows=1 width=16) (actual time=1.130..15.627 rows=40 loops=1)
        ->  Nested Loop  (cost=0.70..1189.95 rows=1 width=24) (actual time=1.065..14.362 rows=40 loops=1)
              ->  Index Only Scan Backward using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..325.08 rows=134 width=8) (actual time=0.064..5.455 rows=564 loops=1)
                    Index Cond: ((account_id = 1) AND (status_id < '99722268340454164'::bigint))
                    Heap Fetches: 469
              ->  Index Only Scan using statuses_dm on statuses  (cost=0.28..6.44 rows=1 width=24) (actual time=0.007..0.007 rows=0 loops=564)
                    Index Cond: ((id = mentions.status_id) AND (id < '99722268340454164'::bigint))
                    Heap Fetches: 24
        ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..5.67 rows=1 width=8) (actual time=0.012..0.016 rows=1 loops=40)
              Heap Fetches: 38
Planning time: 1.478 ms
Execution time: 16.213 ms
```

Heap Fetches for mentions is decreased  from  629 to 469.

abcang (approved):

akihikodaki:
> I do not agree with "There is no room for improvement in the query plan." Merge Join is used in queries before applying this PR, but it Sorts the rows that is much larger than limit. There is no sort of mentions in the query after applying this PR. The number of unused rows is somewhat reduced.

I guess you are talking about: https://gist.github.com/tateisu/cc6bff006b898094262245491b631f2f
```
QUERY PLAN                                                                                         
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=399.99..3372.87 rows=1 width=16) (actual time=21.193..25.817 rows=40 loops=1)
   ->  Nested Loop  (cost=399.99..3372.87 rows=1 width=16) (actual time=21.183..25.517 rows=40 loops=1)
         ->  Merge Left Join  (cost=399.70..3367.86 rows=1 width=24) (actual time=21.129..24.350 rows=40 loops=1)
               Merge Cond: (statuses.id = mentions.status_id)
               Filter: ((mentions.account_id = 1) OR (statuses.account_id = 1))
               Rows Removed by Filter: 48
               ->  Index Only Scan Backward using statuses_dm on statuses  (cost=0.28..2962.95 rows=1869 width=24) (actual time=0.389..2.157 rows=88 loops=1)
                     Heap Fetches: 88
               ->  Sort  (cost=399.42..399.82 rows=161 width=16) (actual time=20.504..21.093 rows=148 loops=1)
                     Sort Key: mentions.status_id DESC
                     Sort Method: quicksort  Memory: 117kB
                     ->  Index Only Scan using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..393.52 rows=161 width=16) (actual time=0.055..13.303 rows=1470 loops=1)
                           Index Cond: (account_id = 1)
                           Heap Fetches: 1182
         ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..5.00 rows=1 width=8) (actual time=0.010..0.014 rows=1 loops=40)
               Heap Fetches: 39
 Planning time: 21.167 ms
 Execution time: 26.263 ms
(19 rows)
```
Here you have to pay attention to two parts, which have different reasons why they query so many statuses:
```
               ->  Index Only Scan Backward using statuses_dm on statuses  (cost=0.28..2962.95 rows=1869 width=24) (actual time=0.389..2.157 rows=88 loops=1)
                     Heap Fetches: 88
```

```
               ->  Sort  (cost=399.42..399.82 rows=161 width=16) (actual time=20.504..21.093 rows=148 loops=1)
                     Sort Key: mentions.status_id DESC
                     Sort Method: quicksort  Memory: 117kB
                     ->  Index Only Scan using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..393.52 rows=161 width=16) (actual time=0.055..13.303 rows=1470 loops=1)
                           Index Cond: (account_id = 1)
                           Heap Fetches: 1182
```

The former one has tons of rows because it does not have `Index Cond`. That is because you cannot really have `Index Cond` because of the order of multicolumn index:
```
create index statuses_dm on statuses(id,account_id,updated_at) where visibility=3;
```
The problem is solved in my index.

I don't know about the later one, so it is concerning also for me. Ideally it should not have extra `Sort` and instead it should sort using index `index_mentions_on_account_id_and_status_id`. My result does not have the problem, but the table size may matter. The query using union does not, either. I'd like to know whether the queries introduced by this pull request do or do not.

tateisu:
currently my instance has both of 
- statuses_dm on statuses (id, account_id, updated_at) WHERE visibility = 3;
- statuses_dm_account on statuses(account_id, id, updated_at) WHERE visibility = 3;
and still statuses_dm is used for query that before apply this PR.
The selection of the query plan and the index may change depending on the data situation.

After applying this PR, _from_me part will be like as:
```
explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE statuses.account_id = 1 AND "statuses"."visibility" = 3  
and "statuses"."id" < 99722268340454164
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;

 Limit  (cost=0.56..16.61 rows=1 width=16) (actual time=0.173..142.684 rows=40 loops=1)
   ->  Nested Loop  (cost=0.56..16.61 rows=1 width=16) (actual time=0.165..142.360 rows=40 loops=1)
         ->  Index Only Scan Backward using statuses_dm_account on statuses  (cost=0.28..8.29 rows=1 width=24) (actual time=0.111..140.563 rows=40 loops=1)
               Index Cond: ((account_id = 1) AND (id < '99722268340454164'::bigint))
               Heap Fetches: 31
         ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..8.31 rows=1 width=8) (actual time=0.020..0.025 rows=1 loops=40)
               Index Cond: (id = 1)
               Heap Fetches: 40
 Planning time: 0.819 ms
 Execution time: 143.018 ms
```

And _to_me part will be like as:
```
explain analyze SELECT "statuses"."id", "statuses"."updated_at" FROM "statuses"     
LEFT OUTER JOIN mentions ON statuses.id = mentions.status_id 
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE mentions.account_id = 1 AND "statuses"."visibility" = 3  
and "statuses"."id" < 99722268340454164
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;

 Limit  (cost=0.99..1444.35 rows=1 width=16) (actual time=5.217..21.009 rows=40 loops=1)
   ->  Nested Loop  (cost=0.99..1444.35 rows=1 width=16) (actual time=5.207..20.702 rows=40 loops=1)
         ->  Nested Loop  (cost=0.70..1438.62 rows=1 width=24) (actual time=5.115..19.277 rows=40 loops=1)
               ->  Index Only Scan Backward using index_mentions_on_account_id_and_status_id on mentions  (cost=0.43..393.52 rows=161 width=8) (actual time=0.099..8.310 rows=724 loops=1)
                     Index Cond: (account_id = 1)
                     Heap Fetches: 629
               ->  Index Only Scan using statuses_dm on statuses  (cost=0.28..6.48 rows=1 width=24) (actual time=0.006..0.006 rows=0 loops=724)
                     Index Cond: ((id = mentions.status_id) AND (id < '99722268340454164'::bigint))
                     Heap Fetches: 24
         ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..5.72 rows=1 width=8) (actual time=0.014..0.018 rows=1 loops=40)
               Heap Fetches: 38
 Planning time: 1.822 ms
 Execution time: 21.328 ms
```

The query plan becomes very simple and the query plan will be more stable.
This is the effect of this PR. 

(wasteful reading of mentions is still exists, but this will not improve unless we change the table structure.)

Gargron:
>(wasteful reading of mentions is still exists, but this will not improve unless we change the table structure.)

How do you imagine the table structure needs to be changed?

Gargron:
I wonder if there would be a benefit in storing mentioned account ids in an array column instead of separate table. (History: there was no "notifications" at the beginning, so instead there was a "mentions" column/API, so it made sense to have a separate table. Nowadays nothing queries the mentions table directly)

tateisu:
to avoid waste reading of mentions table/index

(1) add visibility column to mentions.

- alter table mentions add column visibility smallint not null default 0;
- update mentions set visibility=statuses.visibility from statuses where statuses.id = mentions.status_id;
- create index mentions_dm on mentions(account_id,status_id) where visibility=3;
- (when insert record to mentions, set visibility column)
# this migration is very slow... update all rows in mentions table.

or add direct column to mentions.

- alter table mentions add column direct boolean not null default FALSE;
- update mentions set direct=TRUE from statuses where statuses.id = mentions.status_id and statuses.visibility=3;
- create index mentions_direct on mentions(account_id,status_id) where direct;
- (when insert record to mentions, set direct column)
# this migration is faster, but supports only direct visibility

(2) change pagenation to mentions.status_id instead of statuses.id

(3) change _to_me query like as:

```
explain analyze SELECT "statuses"."id", "statuses"."updated_at" 
FROM mentions
LEFT OUTER JOIN statuses ON statuses.id = mentions.status_id 
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id"       
WHERE mentions.account_id = 1 AND mentions.direct and statuses.visibility=3
and "mentions"."status_id" < 99722268340454164
AND "accounts"."silenced" = false ORDER BY "statuses"."id" DESC LIMIT 40;
```

then waste reading will be eliminated.
```
Limit  (cost=0.84..22.32 rows=1 width=16) (actual time=0.274..3.341 rows=40 loops=1)
   ->  Nested Loop  (cost=0.84..22.32 rows=1 width=16) (actual time=0.262..3.050 rows=40 loops=1)
         ->  Nested Loop  (cost=0.55..16.60 rows=1 width=24) (actual time=0.162..1.652 rows=40 loops=1)
               ->  Index Only Scan Backward using mentions_direct on mentions  (cost=0.28..8.29 rows=1 width=8) (actual time=0.085..0.235 rows=40 loops=1)
                     Index Cond: ((account_id = 1) AND (status_id < '99722268340454164'::bigint))
                     Heap Fetches: 0
               ->  Index Only Scan using statuses_dm on statuses  (cost=0.28..8.29 rows=1 width=24) (actual time=0.017..0.020 rows=1 loops=40)
                     Index Cond: (id = mentions.status_id)
                     Heap Fetches: 24
         ->  Index Only Scan using accounts_not_silenced on accounts  (cost=0.29..5.71 rows=1 width=8) (actual time=0.016..0.020 rows=1 loops=40)
               Heap Fetches: 38
 Planning time: 3.212 ms
 Execution time: 3.665 ms
```

tateisu:
- add some partial index
- add mentions.direct column that is used for partial index.
- migration updates mentions.direct column for each DIRECT status.
- mentions.before_save callback is used to update mentions.direct column before INSERT/UPDATE
- Status.as_direct_timeline_to_me(account) is changed to use partial index of mentions.
- Api::V1::Timelines::DirectController.direct_statuses_to_me is changed to pagenate for mentions.status_id instead of statuses.id

I think that all my ideas were implemented. Please let us know if you have any comments.

tateisu:
after applied this RP, this query is used to get statuses that mentions to the user.

```
SELECT  "statuses"."id", "statuses"."updated_at" FROM "statuses" 
INNER JOIN mentions on statuses.id = mentions.status_id 
LEFT OUTER JOIN "accounts" ON "accounts"."id" = "statuses"."account_id" 
WHERE (mentions.direct and mentions.account_id= 1) AND "statuses"."visibility" = 3 AND 1=1 
AND "accounts"."silenced" = FALSE ORDER BY "statuses"."id" DESC, mentions.status_id desc LIMIT 20

Limit  (cost=0.55..30.51 rows=1 width=24) (actual time=0.041..0.079 rows=7 loops=1)
  ->  Nested Loop  (cost=0.55..30.51 rows=1 width=24) (actual time=0.041..0.078 rows=7 loops=1)
        ->  Nested Loop  (cost=0.27..28.67 rows=1 width=32) (actual time=0.027..0.054 rows=7 loops=1)
              ->  Index Only Scan Backward using index_mentiond_direct on mentions  (cost=0.13..8.19 rows=3 width=8) (actual time=0.021..0.028 rows=7 loops=1)
                    Index Cond: (account_id = 1)
                    Heap Fetches: 7
              ->  Index Scan using index_statuses_dm on statuses  (cost=0.14..6.82 rows=1 width=24) (actual time=0.003..0.003 rows=1 loops=7)
                    Index Cond: (id = mentions.status_id)
        ->  Index Only Scan using index_accounts_not_silenced on accounts  (cost=0.28..1.82 rows=1 width=8) (actual time=0.002..0.003 rows=1 loops=7)
              Heap Fetches: 7
Planning time: 5.516 ms
Execution time: 0.177 ms
```

tateisu:
I've check query log by customizeing postgresql's log_statement.

the constant parameter such as "where visibility=3" is changed to bind parameter for prepared statements by Rails's default behavior.

then PostgreSQL can't use partian index because that decides which index is used when statement is prepared, this timing postgres can't know about bind parameter.

to avoid this, set PREPARED_STATEMENTS=false in .env.production.
If your instance already use pgbouncer, maybe already it is set.

tateisu:
Now using "arel_table" to disable "bind parameter for constant condition" that is required for using partial index. This is work if even  PREPARED_STATEMENTS=false is NOT set (default).

tateisu:
Fixed all things suggested at Mastodon from @abcang .
Please review this PR.

tateisu:
commits rebased.

tateisu:
I think this is not my fault.
＞ci/circleci: test-ruby2.5 — Your tests failed on CircleCI 
＞ActionView::Template::Error:
＞        Webpacker can't find locale_en.js in /home/circleci/projects/mastodon/public/packs-test/manifest.json. Possible causes:

tateisu:
index size example:

statuses 2.71GB
index_statuses_dm_account 56KB
index_statuses_dm 48KB

mentions 183MB (after adding "direct" column )
index_mentions_direct 48KB

accounts 111MB
index_accounts_not_silenced	1.2MB

takayamaki:
I feel that This PR is too big and Premature optimization.

My PR #7633 is about x100 faster than before the commit.
Also, #7614 before adding some commit was the same.
I think that it is fast enough.

At first, we should merge without changing tables and release as rc.
Then if still too slow in production, we may consider adding partial index or column at that time.

tateisu:
Even if I divide the PR, I think that I keep the following items.
"Optimize pagination of _to_me part that using mentions.status_id instead of statuses.id"
It will make difference in case of benchmark with pagination.

tateisu:
About this case, I will wait for the judgment by  project collaborators.

tateisu:
I thought about it now.

This TL is rather close to home + notification TL than public TL.
There is no need to filter silence  in this TL? 
Regardless of the presence or absence of silence, I should be able to read my remarks and mentions for myself.

- account_silencing_filter: certainly unnecessary
- not_in_filtered_languages: unnecessary
- not_domain_blocked_by_account: unnecessary
- not_excluded_by_account: REQUIRED?

tateisu:
- change filter condition for direct timeline. old implementation use filters for public timeline, but direct timeline is not public.
- revoke introducing index_accounts_not_silenced because it is not used for direct timeline.
- fix rspec test : re-querying TL after append mentions fabric.

tateisu:
in commit https://github.com/tootsuite/mastodon/pull/7614/commits/58efdd84873f7c9879450fd0242c68ec05034be0 , I've temporary revoke additional column and partial index.
if you have try previous commits, please db:rollback x4 times BEFORE merge this commit.

Gargron (changes requested):

tateisu:
@akihikodaki this PR currently does not contains index, but if I prepare the next PR, statuses_dm_account and mentions_dm will be included. after applied this PR, statuses_dm is used in _to_me part ,that checks visibility=3 and get updated_at with index only scan, but index only scan is not mandatory.

In case of DM TL, I recognize that the instance requiring query optimization is the majority. Your case is rather rare.

tateisu:
@Gargron
Some other method in Status returns ActiveRecord::Relation.
But in case of direct timeline, I can't keep it because query result merging is works on application side.
Status.as_home_timeline in current commit returns array of Status or cache_ids object (selectable by method argument).

Is this acceptable?

tateisu:
Now Status.as_direct_timeline returns ActiveRecord::Relation if cache_ids method argument is false.
( with using more 1 DB query. but currently it used only from spec test. )

Gargron (approved):

Gargron:
StatusThreadingConcern also does not return relation, that is fine

PR #7618, opened 2018-05-25
Fix color of link anchors in /about/more

lynlynlynx:


Gargron:
They look fine to me in 2.4 - what kind of fix is this?

lynlynlynx:
When high contrast theme was separated, text colors are completely separated from UI colors, bnt I overlooked to change this anchors color. In default theme, they look fine, but it becomes a problem in the new mastodon-light theme.

Gargron (approved):

PR #7644, opened 2018-05-27
Improve counter caches on Status and Account

### LINKED ISSUE #828: Direct messages should not increment one's post count

Gargron:
Do not touch statuses_count on accounts table when mass-destroying statuses to reduce load when removing accounts, same for reblogs_count and favourites_count

Do not count statuses with direct visibility in statuses_count

Fix #828

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -41,12 +41,12 @@ class Status < ApplicationRecord
 
   belongs_to :application, class_name: 'Doorkeeper::Application', optional: true
 
-  belongs_to :account, inverse_of: :statuses, counter_cache: true
+  belongs_to :account, inverse_of: :statuses
   belongs_to :in_reply_to_account, foreign_key: 'in_reply_to_account_id', class_name: 'Account', optional: true
   belongs_to :conversation, optional: true
 
   belongs_to :thread, foreign_key: 'in_reply_to_id', class_name: 'Status', inverse_of: :replies, optional: true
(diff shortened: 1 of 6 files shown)


app/models/status.rb
```diff
@@ -167,6 +167,17 @@ def emojis
     @emojis ||= CustomEmoji.from_text([spoiler_text, text].join(' '), account.domain)
   end
 
+  def mark_for_mass_destruction!
+    @mark_for_mass_destruction = true
```
nightpool: should be @maked... do we have tests on this code?

app/models/status.rb
```diff
@@ -355,4 +366,40 @@ def update_statistics
...
+      Account.where(id: account_id).update_all('statuses_count = GREATEST(COALESCE(statuses_count, 0) - 1, 0)')
+    end
+
+    return unless reblog?
+
+    if association_cached?(:reblog)
+      reblog.update_attribute(:reblogs_count, [reblog.reblogs_count - 1, 0].max)
+    else
+      Status.where(id: reblog_of_id).update_all('reblogs_count = GREATEST(COALESCE(reblogs_count, 0) - 1, 0)')
+    end
```
nightpool: also worried about the non-association cached branches not getting tested. maybe add a `.reload` to the tests?

nightpool (changes requested):

nightpool (approved):

PR #7646, opened 2018-05-28
Do not write to IndexedDB if stored and updated objects are equal

### LINKED ISSUE #7615: Having a high-traffic feed open causes enormous disk I/O in Firefox
On mastosoc (2.4.0), with Firefox 60.0, I've noticed my hard disk light is on way too frequently when I've got something like the Federated Timeline open. By flipping `sysctl vm.block_dump=1` and watching `dmesg -w`, I can see it's coming from IndexedDB doing a bunch of sqlite and fsync calls every time a new post appears.

Aside from that, there's now an enormous (varies, but I've seen over 150MB) directory in

    ~/.mozilla/firefox/$profiledir/storage/default/https+++mastodon.social^firstPartyDomain=mastodon.social/cache/morgue/

This usage also shows up in the preferences window when I check how much data each site's storing, even though IIRC the browser default is supposed to be 5MB:

2018-05-25_01 29 54_1089x357

If I had to take an educated guess, what's happening here is all those posts are getting saved for offline mode and then getting evicted by the browser's offline storage limit, resulting in the browser generating a huge garbage pile and never cleaning up after itself.

Obviously Mozilla needs to fix their horrid implementation here and stop trashing my shiny new SSD, but I guess what I'm asking for in this issue is if there's any way it can be mitigated on this end?

* * * *

Gargron:
Fix #7615

### DIFF
--- a/app/javascript/mastodon/storage/modifier.js
+++ b/app/javascript/mastodon/storage/modifier.js
@@ -1,4 +1,5 @@
 import openDB from './db';
+import { isEqual } from 'lodash';
 
 const accountAssetKeys = ['avatar', 'avatar_static', 'header', 'header_static'];
 const storageMargin = 8388608;
@@ -38,11 +39,17 @@ function put(name, objects, onupdate, oncreate) {
         }
 
         if (retrieval.target.result) {
-          if (onupdate) {
-            onupdate(object, retrieval.target.result, putStore, deleteObject);
-          } else {
-            deleteObject();
-          }
+          putStore.get(retrieval.target.result).onsuccess = ({ target }) => {
+            if (isEqual(target.result, object)) {
+              return;
+            }
+
+            if (onupdate) {
+              onupdate(object, retrieval.target.result, putStore, deleteObject);
+            } else {
+              deleteObject();
+            }
+          };
         } else {
           if (oncreate) {
             oncreate(object, addObject);


akihikodaki (changes requested):
The behavior to always write IndexedDB is intentional. It updates ID and marks objects are latest.

Gargron:
> The behavior to always write IndexedDB is intentional. It updates ID and marks objects are latest.

That is not okay because it means disk I/O for every record that streams or is loaded into the webapp. Offline storage should not negatively impact online use. I think this is so that recently accessed records don't get evicted by storage quota? There must be another way.

akihikodaki:
> That is not okay because it means disk I/O for every record that streams or is loaded into the webapp. Offline storage should not negatively impact online use. I think this is so that recently accessed records don't get evicted by storage quota? There must be another way.

That is possible to delay the update if matters, but such disk I/O should be a few killobytes at most and I don't think it is an issue to be solved.

Gargron:
Idea: Keep accessed keys in a simple JS variable. When evicting, only exclude objects referenced from that variable. That way, objects are always persisted for the immediately next reload (no guarantees beyond that but I don't think that's required)

akihikodaki:
> Idea: Keep accessed keys in a simple JS variable. When evicting, only exclude objects referenced from that variable. That way, objects are always persisted for the immediately next reload (no guarantees beyond that but I don't think that's required)

That sounds OK. The variable could be Redux state (just check `statuses` and `accounts`.)

PR #7647, opened 2018-05-28
Record trending tags from ActivityPub, too

Gargron:
Ref: https://github.com/tootsuite/mastodon/pull/7638#issuecomment-392398256


ykzts (approved):
Good quick fix

PR #7649, opened 2018-05-28
optimize direct message 2: add mentions.direct column and some partial index

tateisu:
This is the separated part from https://github.com/tootsuite/mastodon/pull/7614 .

Please try this if you feel slowly when showing direct message timeline, 

- add partial index for statuses: ["account_id", "id", "updated_at"], name: "index_statuses_dm_account", where: "(visibility = 3)"
- add column mentions.direct  (boolean not null), true if related status is direct message
- add partial index for mentions: ["account_id", "status_id"], name: "index_mentions_direct", where: "direct"
- add where expression to direct timeline query that references mentions.direct column. 

The size of the partial index depends on the ratio of DM in statuses or mentions. The same thing can be said for index update overhead. The update overhead will not increase for status and mentions other than DM.

For additional columns, disk usage will be increased a few bytes per every rows in the mentions table.
Migration takes time depends on size of mentions because this repository does not allows nullable boolean column and then we have to update all rows in mentions table.

### DIFF
--- a/app/models/mention.rb
+++ b/app/models/mention.rb
@@ -8,6 +8,7 @@
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #  account_id :bigint(8)
+#  direct     :boolean          default(FALSE), not null
 #
 
 class Mention < ApplicationRecord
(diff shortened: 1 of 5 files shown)


tateisu:
The reason for creating the index is not only for improving the speed of general users but also for avoiding the worst case.

- if there is no index, direct timeline API may be very heavy in worst case.
- Attackers can create worst case easily.

Gargron:
Resolved by #8832

PR #7658, opened 2018-05-28
Deduplicate accounts and make unique username/domain index case-insensitive

### LINKED ISSUE #6667: Issues after changing WEB_DOMAIN to be the same as LOCAL_DOMAIN

### LINKED ISSUE #6837: Usernames are case-sensitive on local database, but not on remote domain (even with the same id)

### LINKED ISSUE #6937: Unique index on accounts username/domain is not case insensitive
For some reason (a mistake, oversight or confusion), there are two indices on the accounts table:

- lower(username), lower(domain): case-insensitive, *not* unique
- username, domain: case-*sensitive*, unique

Only the last one is unique! That's a waste of space, but it's also bad because it allows duplicate accounts to be created under race conditions or other edge cases (those that were discovered, were patched already, though)

We need to:

- Turn the case-insensitive index unique and remove the redundant one
- But before we can do that, we need to, in an automated manner, "fix" duplicate rows, by either merging, removing, or renaming the accounts

In the case of merging, that would require updating all associate records to a different account_id. In case of removing, need to decide which of the accounts to keep... This is further complicated by the bug (already fixed) that occurred in 2.3.2 and allowed, under some circumstances, multiple local accounts with the same username (different casing).

Very likely related issue: #3005

Gargron:
> **Previously on Mastodon**: Username/domain uniqueness constraint exists, but is case-*sensitive*. Case-insensitive index also exists, but is not unique. Duplicates caused primarily by admins changing casing of usernames in database, and that one time we had a missing validation in Ruby for one short release.

Fix #6937
Fix #6837
Fix #6667

1. Get all accounts with overlapping username/domain
2. For each account:
    1. Select most recently updated (used) one to be the reference account
    2. Compare others against it
        - If the public key is the same, it's definitely the same person, so they can be merged. Change attributions on owned associations to the reference account
        - Delete the other account
3. Re-create case-insensitive index with uniqueness constraint
4. Remove now useless case-sensitive unique index

> Damn, why do all those issues end in 7s. I just noticed.

### DIFF
--- a/db/migrate/20180528141303_fix_accounts_unique_index.rb
+++ b/db/migrate/20180528141303_fix_accounts_unique_index.rb
@@ -1,0 +1,84 @@
+class FixAccountsUniqueIndex < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    say ''
+    say 'WARNING: This migration may take a *long* time for large instances'
+    say 'It will *not* lock tables for any significant time, but it may run'
+    say 'for a very long time. We will pause for 10 seconds to allow you to'
+    say 'interrupt this migration if you are not ready.'
+    say ''
+    say 'This migration will irreversibly delete user accounts with duplicate'
+    say 'usernames. You may use the `rake mastodon:maintenance:find_duplicate_usernames`'
+    say 'task to manually deal with such accounts before running this migration.'
+
+    10.downto(1) do |i|
+      say "Continuing in #{i} second#{i == 1 ? '' : 's'}...", true
+      sleep 1
+    end
+
+    duplicates = Account.connection.select_all('SELECT string_agg(id::text, \',\') AS ids FROM accounts GROUP BY lower(username), lower(domain) HAVING count(*) > 1').to_hash
+
+    duplicates.each do |row|
+      deduplicate_account!(row['ids'].split(','))
+    end
+
+    remove_index :accounts, name: 'index_accounts_on_username_and_domain_lower' if index_name_exists?(:accounts, 'index_accounts_on_username_and_domain_lower')
+    safety_assured { execute 'CREATE UNIQUE INDEX CONCURRENTLY index_accounts_on_username_and_domain_lower ON accounts (lower(username), lower(domain))' }
+    remove_index :accounts, name: 'index_accounts_on_username_and_domain' if index_name_exists?(:accounts, 'index_accounts_on_username_and_domain')
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  private
+
+  def deduplicate_account!(account_ids)
+    accounts          = Account.where(id: account_ids).to_a
+    accounts          = account.first.local? ? accounts.sort_by(&:created_at) : accounts.sort_by(&:updated_at).reverse
+    reference_account = accounts.shift
+
+    accounts.each do |other_account|
+      if other_account.public_key == reference_account.public_key
+        # The accounts definitely point to the same resource, so
+        # it's safe to re-attribute content and relationships
+        merge_accounts!(reference_account, other_account)
+      elsif other_account.local?
+        # Since domain is in the GROUP BY clause, both accounts
+        # are always either going to be local or not local, so only
+        # one check is needed. Since we cannot support two users with
+        # the same username locally, one has to go. 😢
+        other_account.user.destroy
+      end
+
+      other_account.destroy
+    end
+  end
+
+  def merge_accounts!(main_account, duplicate_account)
+    [Status, Favourite, Mention, StatusPin, StreamEntry].each do |klass|
+      klass.where(account_id: duplicate_account.id).update_all(account_id: main_account.id)
+    end
+
+    # Since it's the same remote resource, the remote resource likely
+    # already believes we are following/blocking, so it's safe to
+    # re-attribute the relationships too. However, during the presence
+    # of the index bug users could have *also* followed the reference
+    # account already, therefore mass update will not work and we need
+    # to check for (and skip past) uniqueness errors
+    [Follow, FollowRequest, Block, Mute].each do |klass|
+      klass.where(account_id: duplicate_account.id).find_each do |record|
+        record.update(account_id: main_account.id)
+      rescue ActiveRecord::RecordNotUnique
+        next
+      end
+
+      klass.where(target_account_id: duplicate_account.id).find_each do |record|
+        record.update(target_account_id: main_account.id)
+      rescue ActiveRecord::RecordNotUnique
+        next
+      end
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_05_14_140000) do
+ActiveRecord::Schema.define(version: 2018_05_28_141303) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -77,10 +77,9 @@ ActiveRecord::Schema.define(version: 2018_05_14_140000) do
     t.jsonb "fields"
     t.string "actor_type"
     t.index "(((setweight(to_tsvector('simple'::regconfig, (display_name)::text), 'A'::\"char\") || setweight(to_tsvector('simple'::regconfig, (username)::text), 'B'::\"char\")) || setweight(to_tsvector('simple'::regconfig, (COALESCE(domain, ''::character varying))::text), 'C'::\"char\")))", name: "search_index", using: :gin
-    t.index "lower((username)::text), lower((domain)::text)", name: "index_accounts_on_username_and_domain_lower"
+    t.index "lower((username)::text), lower((domain)::text)", name: "index_accounts_on_username_and_domain_lower", unique: true
     t.index ["uri"], name: "index_accounts_on_uri"
     t.index ["url"], name: "index_accounts_on_url"
-    t.index ["username", "domain"], name: "index_accounts_on_username_and_domain", unique: true
   end
 
   create_table "admin_action_logs", force: :cascade do |t|


Gargron:
@ThibG @aschmitz Check this out. Is this alright?

aschmitz:
I'll defer to @marrus-sh on whether or not this makes sense in APub, but the code looks like it does what you describe.

It's not clear to me if/how this would fix #6667, though: that appears to be a difference in domain, not capitalization. (Unless I've missed something?)

aschmitz:
To repost my comment from Mastodon:

Technically it's a bit racey: If something comes in from the other account while you're in the process of merging, it could get missed. This is unlikely, but possible. I'm not sure if you can fix that, though: maybe `other_account.delete` before the `merge_accounts!` if statement? Any of the necessary cleanup from `.destroy` is probably handled in your reassigning bits.

Gargron:
>It's not clear to me if/how this would fix #6667, though: that appears to be a difference in domain, not capitalization. (Unless I've missed something?)

The index on `uri` is not unique, so that wasn't the problem (unlike my first comment in the issue suggests, I was wrong). When you change WEB_DOMAIN/LOCAL_DOMAIN setup, the `domain` doesn't actually change (since the point of WEB_DOMAIN is having a top-level domain while running from a subdomain), it's the URI that changes, because the URI has to be resolveable, it's based on WEB_DOMAIN. So the URI changed but domain remained the same, but because the code tried finding/creating based on URI first and domain second, it tried to make a duplicate and failed. This was fixed in #6842, actually, I dunno why I didn't close that issue then.

Gargron:
>Technically it's a bit racey: If something comes in from the other account while you're in the process of merging, it could get missed. This is unlikely, but possible. I'm not sure if you can fix that, though: maybe other_account.delete before the merge_accounts! if statement? Any of the necessary cleanup from .destroy is probably handled in your reassigning bits.

Reference account is the most recently updated one, it's highly unlikely that both accounts could be active at the same time, therefore any new content would arrive to the reference account, not the other accounts.

Foreign keys should take care of any orphaned associations.

Mostly duplicates should be rare, so any resulting oddities would likely not be noticed.

One thing that does give me pause is the treatment of local duplicates. Those were possible due to a short-lived bug in a release. My migration will just delete them. I don't see a lot of choice here, since automatically changing the usernames would both be difficult to implement and it would mess with federation. There is also a previously added rake task for finding such local duplicates to manually deal with, if one is inclined to do so.

aschmitz:
In theory could merging the local accounts work? Other Mastodon clients will eventually merge the two anyway, though I'm not sure about other servers. (I'm a little iffy on merging local accounts given that they *could* be completely different users, though.)

Gargron:
@aschmitz Two local accounts are *always definitely* different people, with different user record, password, e-mail etc.

Gargron:
>Other Mastodon clients will eventually merge the two anyway

If you mean other servers, since we perform the public_key comparison, if they're really two different accounts, they will not be merged either.

aschmitz:
Ah, okay, fair enough. Is there a convenient way to simply disable/rename one? The admin would have to handle notifying the users somehow (or we could fire off an email), but it might be better than deleting a whole user's information.

Gargron:
No, there's not - if it remains in the database, it will prevent the fixed index from being created.

aschmitz:
Not even renaming to, say, `username_1` etc.? (This may be bad because users can't rename themselves after the fact, but could potentially be preferable to actually deleting all of their data.)

Gargron:
Automatically changing the usernames would both be difficult to implement (now have to loop over to make sure username1 doesn't exist, username2, username3, etc) and it would mess with federation.

Since duplicate accounts are a mistake, and a rare one at that, it's okay and cleaner to just delete them.

ThibG:
I'm still not sold on having case-insensitive handling of usernames, it's yet another deviation from the spec.
Otherwise, this seems to do what you described.
If you go that route, you might also want to change `app/models/account.rb:72`.

Gargron:
>I'm still not sold on having case-insensitive handling of usernames, it's yet another deviation from the spec.

**_I am the spec_**

<details>
  <summary>Click to expand</summary>
  That's a Star Wars reference, I am not the spec. However, the spec is unusable without webfinger anyway.
</details>

nightpool:
@ThibG 

The fact that we require unique and immutable preferredUsernames at all is a deviation from the spec. (which only requires unique ids and doesn't talk about immutability at all). Given that, and the justifications we have for it ("textual @-mentions should map 1:1 with accounts"), i dont think case-sensitive verses case-insenstive is a meaningful change here.

db/migrate/20180528141303_fix_accounts_unique_index.rb
```diff
@@ -0,0 +1,68 @@
...
+      if other_account.public_key == reference_account.public_key
+        # The accounts definitely point to the same resource, so
+        # it's safe to re-attribute content and relationships
+        merge_accounts!(reference_account, other_account)
+      elsif other_account.local?
+        # Since domain is in the GROUP BY clause, both accounts
+        # are always either going to be local or not local, so only
+        # one check is needed. Since we cannot support two users with
+        # the same username locally, one has to go. 😢
+        other_account.user.destroy
```
nightpool: I think it's preferable to error out here and force admins to resolve this manually, to pick the account with the most statuses/followers/etc. would be very bad if this ended up destroying a legit account. (although unlikely given the raciness involved in getting this to happen in the first place)

db/migrate/20180528141303_fix_accounts_unique_index.rb
```diff
@@ -0,0 +1,68 @@
...
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  private
+
+  def deduplicate_account!(account_ids)
+    accounts          = Account.where(id: account_ids).order(updated_at: :desc)
+    reference_account = accounts.shift
+
+    accounts.each do |other_account|
```
nightpool: the each block here should be wrapped in a transaction I think?

nightpool (changes requested):

db/migrate/20180528141303_fix_accounts_unique_index.rb
Gargron: Is that really necessary? Even if it fails halfway through, that's still work done that won't need to be repeated on a future run.

db/migrate/20180528141303_fix_accounts_unique_index.rb
Gargron: Is anything stopping me from using TTY prompts here to give dynamic options? Except that it feels wrong for migrations to require interactive input.

trwnh:
>@Gargron:
>>Two local accounts are always definitely different people, with different user record, password, e-mail etc.
> 
>@ThibG:
>>I'm still not sold on having case-insensitive handling of usernames, it's yet another deviation from the spec.

One other thing to consider besides simple case-sensitivity: what if a hypothetical future ActivityPub implementation allows for changing usernames? In that situation, "case-insensitive" isn't enough either. Or how about another ActivityPub implementation that allows export/import of accounts to a completely different domain (and also may or may not allow for username changes)?

In both those cases, it's still the same Account, with the same public key, just with a different Webfinger pointer. Perhaps this is out-of-scope for this particular PR, but it's definitely a future extension to consider. This would also technically be the superset of the problem: when a pointer changes but the account is the same, for any reason.

ThibG has a point where two accounts are *different* and have the same name with different casing (e.g. claudia/Claudia @ mastodon.social, two different people). Over federation, a remote instance could follow both. On the local instance, any user can follow both. The case-insensitivity only matters when changing username on the same account (e.g. krita/Krita @ mastodon.art, same account).

I guess ultimately this PR is addressing one narrow case: it assumes overlaps ONLY occur on case-insensitive user/domain pairs. If we use public keys to deduplicate, then worst-case we would have to scan over all users the instance is aware of. Perhaps the more robust alternative would be to use public-key as an index in the database, so that duplicate public keys are detected immediately. Of course, that in turn assumes that public keys are unique for every single ActivityPub account...

db/migrate/20180528141303_fix_accounts_unique_index.rb
nightpool: I'm saying each iteration of the loop should be a new transaction, so we don't lose half the progress but an error doesn't leave an account in a half-deleted state.

nightpool:
@Jarred-Sumner we can't change the type of the column at this point. Plus, we don't want the behavior citext gives here—if two accounts share the same name with different casing, we want to fail fast rather then running into nasty compatibility issues down the line when software that assumes they should be distinct ends up "sharing" the same account on our end.

db/migrate/20180528141303_fix_accounts_unique_index.rb
nightpool: lots of people do unmonitored deploys, heroku deploys, etc, that would break here

nightpool:
Question—given what I said above, do ActivityPub processing queries/webfinger lookup queries **use** the new index? since the index is a function index on `lower(username), lower(domain)` we might not end up using it in any of our queries.

db/migrate/20180528141303_fix_accounts_unique_index.rb
```diff
@@ -0,0 +1,84 @@
...
+    say 'It will *not* lock tables for any significant time, but it may run'
+    say 'for a very long time. We will pause for 10 seconds to allow you to'
+    say 'interrupt this migration if you are not ready.'
+    say ''
+    say 'This migration will irreversibly delete user accounts with duplicate'
+    say 'usernames. You may use the `rake mastodon:maintenance:find_duplicate_usernames`'
+    say 'task to manually deal with such accounts before running this migration.'
+
+    10.downto(1) do |i|
+      say "Continuing in #{i} second#{i == 1 ? '' : 's'}...", true
```
nightpool: should be `:subitem` instead of true, so what the flag does is more obvious.

nightpool (approved):
i'd prefer if you change the boolean flag -> self-documenting symbol thing but it's not a huge deal

trwnh:
>Nothing I do here will change the fact that changing usernames will break things in Mastodon. So let's not even bring it up.

Just wondering how this PR could be expanded upon in the future. If there's a better place that would be on-topic, I can post it there instead, but my question was moreso about the deduplication itself rather than a user/domain change. In a future where Webfinger is used only for lookup and addressing (not for database), and where accounts can be fully migrated, it would probably be worth considering deduplicating in more cases than just case-sensitivity. Still, this PR is an important first step towards solving the larger issue. Right now, it's database admins making manual edits, and we assume that they only change the case.

db/migrate/20180528141303_fix_accounts_unique_index.rb
```diff
@@ -0,0 +1,84 @@
...
+    # Since it's the same remote resource, the remote resource likely
+    # already believes we are following/blocking, so it's safe to
+    # re-attribute the relationships too. However, during the presence
+    # of the index bug users could have *also* followed the reference
+    # account already, therefore mass update will not work and we need
+    # to check for (and skip past) uniqueness errors
+    [Follow, FollowRequest, Block, Mute].each do |klass|
+      klass.where(account_id: duplicate_account.id).find_each do |record|
+        record.update(account_id: main_account.id)
+      rescue ActiveRecord::RecordNotUnique
```
ykzts: `do`-`rescue` is not support on Ruby 2.4

PR #7664, opened 2018-05-29
Use URL polyfill

### LINKED ISSUE #7665: `TypeError: URL is not a constructor` on assets:precompile by docker-compose
Using docker-compose image (edge tag), the error below happens when running `docker-compose run --rm web bin/rails assets:precompile`

```
Starting mastodon_redis_1 ...
Starting mastodon_db_1 ... done
Webpacker is installed 🎉 🍰
Using /mastodon/config/webpacker.yml file for setting up webpack paths
Compiling…
Compilation failed:
/mastodon/config/webpack/production.js:32
  const { host } = new URL(process.env.SWIFT_OBJECT_URL);
                   ^
TypeError: URL is not a constructor
    at Object.<anonymous> (/mastodon/config/webpack/production.js:32:20)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at requireConfig (/mastodon/node_modules/webpack/bin/convert-argv.js:97:18)
    at /mastodon/node_modules/webpack/bin/convert-argv.js:104:17
```

I think this may be because `node` version is too old.

```
host:~# nsenter --target $(docker inspect --format '{{.State.Pid}}' mastodon_web_1) --mount --uts --ipc --net --pid
eb2a98733242:/# node --version
v6.10.3
eb2a98733242:/#
```

https://github.com/tootsuite/mastodon/blob/8bb74e50beb3e6602c183e59086ccebe87f96fb2/Dockerfile#L1

Current Dockerfile uses `ruby:2.4.4-alpine3.6` image, but [it has nodejs-6.10.3-r2](https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.6).
nodejs-8.* is provided by [alpine-3.7](https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.7) or later, so I think that the Dockerfile should be updated to use newer version of alpine image.

This problem might be introduced by <https://github.com/tootsuite/mastodon/commit/d95642f6d913a99fc44f0ac0695d53534afb7962#diff-2f0be8bdcc2ac49bb84f6cb05245e696R11>.

* * * *

ykzts:
Resolve #7665

> Added in: v7.0.0, v6.13.0

[The WHATWG URL API](https://nodejs.org/dist/latest-v6.x/docs/api/url.html#url_the_whatwg_url_api)

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -8,7 +8,7 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
 const OfflinePlugin = require('offline-plugin');
 const { publicPath } = require('./configuration.js');
 const path = require('path');
-const { URL } = require('url');
+const { URL } = require('whatwg-url');
 
 let compressionAlgorithm;
 try {
--- a/package.json
+++ b/package.json
@@ -121,7 +121,8 @@
     "webpack-bundle-analyzer": "^2.9.1",
     "webpack-manifest-plugin": "^1.2.1",
     "webpack-merge": "^4.1.1",
-    "websocket.js": "^0.1.12"
+    "websocket.js": "^0.1.12",
+    "whatwg-url": "^6.4.1"
   },
   "devDependencies": {
     "babel-eslint": "^8.2.1",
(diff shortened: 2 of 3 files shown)


unarist (approved):
The error described in the issue seems resolved and the host part of `SWIFT_OBJECT_URL` is embedded in sw.js correctly. 👍

PR #7671, opened 2018-05-29
Fix context building in the reducer

### LINKED ISSUE #7668: Unhandled Promise Rejection. Browser freezes after upgrading to 2.4.0
Many users have reported that their browsers hang after I upgrade to 2.4.0

The issue can be repeated (mostly) by clicking on a toot again after replying to it, affected browsers are:
Safari / Chrome / Firefox

Anyone has the same issue?

unarist:
regression from #7150

This fixes below bugs:

* addReply() had used native compare operator for string ids
  => descendants may appears at wrong position
* CONTEXT_FETCH_SUCCESS had added the focused status as the reply of the *first* status in ancestors, not last status.
  => descendants may also appears wrong position as well as correct position ([demo.gif](https://user-images.githubusercontent.com/705555/40662997-bb82b27a-6392-11e8-9591-03e2d03a63be.gif))
  </details>
* TIMELINE_UPDATE had added the status to replies of *itself* instead of replied status
  => browser will hangs if you open the status due to a circular reference (probably fixes #7668)

### DIFF
--- a/app/javascript/mastodon/reducers/contexts.js
+++ b/app/javascript/mastodon/reducers/contexts.js
@@ -5,6 +5,7 @@ import {
 import { CONTEXT_FETCH_SUCCESS } from '../actions/statuses';
 import { TIMELINE_DELETE, TIMELINE_UPDATE } from '../actions/timelines';
 import { Map as ImmutableMap, List as ImmutableList } from 'immutable';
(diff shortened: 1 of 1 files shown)


Gargron (approved):

ykzts:
image
image


why?

PR #7672, opened 2018-05-29
Reduce wasted work in RemoveStatusService due to inactive followers

Gargron:


nightpool (approved):

app/models/concerns/account_interactions.rb
```diff
@@ -4,6 +4,17 @@ module AccountInteractions
...
+    def followers_for_local_distribution
+      followers.local
+               .joins(:user)
+               .where('users.current_sign_in_at > ?', User::ACTIVE_DURATION.ago)
+    end
+
+    def lists_for_local_distribution
+      lists.joins(account: :user)
+           .where('users.current_sign_in_at > ?', User::ACTIVE_DURATION.ago)
+    end
```
nightpool: should not be in class methods

PR #7697, opened 2018-06-01
Use different algorithm for trending tags

Gargron:
Sort hashtags based on how different the currently observed usage frequency is to the expected (previously observed) usage frequency. Count uses by unique people instead of total volume within time frame. Threshold of 5.

Refs:

- <https://www.youtube.com/watch?v=duHxpSTmwW0>
- <http://blogs.ischool.berkeley.edu/i290-abdt-s12/files/2012/08/Kostas_Trends_Sept_13_2012.pdf>

### DIFF
--- a/app/models/trending_tags.rb
+++ b/app/models/trending_tags.rb
@@ -2,17 +2,16 @@
 
 class TrendingTags
   KEY                  = 'trending_tags'
-  HALF_LIFE            = 1.day.to_i
-  MAX_ITEMS            = 500
   EXPIRE_HISTORY_AFTER = 7.days.seconds
+  THRESHOLD            = 5
 
   class << self
     def record_use!(tag, account, at_time = Time.now.utc)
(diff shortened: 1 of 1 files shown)


Gargron:
I have to separately rework the getting started column again to adjust for feedback about trends taking up too much space (actually, I get both feedback about how it's too much, and about how it doesn't display enough trends) but this algorithm unlike the previous one will not display tags that aren't actually being used, and so it may deliver empty results, in which case hiding the widget makes sense.

jeroenpraat:
@Gargron Ok, that sounds good. Although I would advise to hide poor results (as far as it's possible to measure that) as well. Imagine a new user or a new instance owner on a small/new instance. They will get a bad impression from Mastodon and the rest of the fediverse, even much worse than looking to the public timelines, cause trending hashtags are of course a subset of all available public toots.

Issue #7702, opened 2018-06-01
Admin Option to Disable Trending Feature


lawremipsum:
Instance admins should be able to easily disable the trending hashtag feature.

* * * *

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nightpool:
why?

mal0ki:
Why it would be good to be allowed to turn off trending hashtags on a server:
* It's federated, so anyone outside of your server could affect what your people see (to the extent that you federate with them), by gaming the trending.
* It will affect discourse, and instance admins may not want that and would therefor want to turn it off.
* Even if it doesn't directly mess up the chronological timelines more than boosts already do, it will affect it in the way of "people talk about what people talk about and talk more about it"

All these reasons are valid reasons for an instance admin to not want to have this feature turned on on their instance.

lawremipsum:
- multiple admins want to
- it is of arguable value
- it alters the feel of a Mastodon instance and emphasizes/implements values that admins may disagree with; e.g., algorithmic trending topics is an anti-pattern that promotes things in a way that negatively influences communities

Laurelai:
A lot of admins do not want this feature as it contributes to the gamification of social media, something we left twitter to avoid. So turning this off should be an option. In fact it should be off by default if you absolutely have to add trending tags and frankly, you shouldn't.

jh4c:
Trending topics have never been well received on any network that's tried to implement them. There is no user-friendly way to do so, it's an inherently user-antagonistic system, it created echo chambers of opinion and offers nothing new. I thought the point of Mastodon was to leave out all the bullshit from other networks?

nightpool:
I think the following settings would be a good idea

Calculate trending topics based on:

 - All known posts
 - Local posts
 - Do not calculate trending topics

with the understanding that instances that don't calculate trending topics
are making it harder for new users to get started on Mastodon. (Because
most users are intimidated by the full local timeline)

Gargron:
>I thought the point of Mastodon was to leave out all the bullshit from other networks?

That's kind of a loaded way of arguing. The point of Mastodon is being a decentralized, standards-based social network with a focus on user experience and anti-abuse tools. What is and isn't "bullshit from other networks" is far from clear and has no direct relation to the mission statement. The feature request over at #271 has gained a lot of support so clearly it's a feature that some people consider useful.

[...]

PR #7711, opened 2018-06-02
Remove trending hashtags

Gargron:
https://github.com/tootsuite/mastodon/issues/7702#issuecomment-394100570


Thann:
I feel like this is reason enough to keep the index =[
screenshot from 2018-06-02 12-40-51

Thann:
I'd like to suggest optionally sorting the search results by "popularity" but I fear starting another shitstorm ^.^

Gargron:
I doubt that would be considered bad by anyone but since usage & trends are in redis, I can't use them to sort results coming from postgres without a lot of overhead. It would require changing some code.

koyuawsmbrtn:
delet dis

remram44:
Is the absence of spam from supporters of the feature on #7702 taken as non-support? 🙁 Removing the list from plain sight while other solutions are still available to get it (even for non-tech people!) seems like the worst of both worlds.

mal0ki:
> The usage stats for the search results remain.

Will it also order them according to this? Instead of alphabetical? 

It's a good feature, and probably the better part of the original implementation. I'm glad you considered keeping it as part of a compromise.

jebuskrust:
Strawpoll

PR #7719, opened 2018-06-04
Add tests for embeds controller

shuheiktgw:
Added tests for embeds_controller! 👍


spec/lib/status_finder_spec.rb
```diff
@@ -6,10 +6,11 @@
   include RoutingHelper
 
   describe '#status' do
+    subject { described_class.new(url) }
```
ykzts: good catch!

ykzts (approved):

PR #7722, opened 2018-06-04
Fix activity-stream-tabs colors and set text variables for text colors

lynlynlynx:
- An active activity-stream-tabs button doesn't have enough contrast.
- Fix color settings of texts for these are not applied text color variables.


lynlynlynx:
|Before|After|
|--|--|
|image|image|
|image|image|

The latter one looks almost same but text color settings are all united as text color variables.

Gargron:
Okay. By the way, I think the separators (horizontal borders etc) in light theme should not be so dark, they should be very subtly darker than the background, it looks jarring otherwise.

Gargron (approved):

PR #7733, opened 2018-06-04
Detect extname from Content-Type

ykzts:
### failed cases 

- `https://cdn.example.com/v1/media.png;compress=true`
- `https://embed.example.com/decorate.php?illust_id=68957437`

### DIFF
--- a/app/models/concerns/remotable.rb
+++ b/app/models/concerns/remotable.rb
@@ -24,14 +24,16 @@ module Remotable
           Request.new(:get, url).perform do |response|
             next if response.code != 200
 
-            matches  = response.headers['content-disposition']&.match(/filename="([^"]*)"/)
-            filename = matches.nil? ? parsed_url.path.split('/').last : matches[1]
+            content_type = parse_content_type(response.headers['content-type'])
+            extname      = detect_extname_from_content_type(content_type)
+
+            if extname.nil?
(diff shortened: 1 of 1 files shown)


Gargron (approved):

app/models/concerns/remotable.rb
```diff
@@ -24,14 +24,16 @@ def remotable_attachment(attachment_name, limit)
           Request.new(:get, url).perform do |response|
             next if response.code != 200
 
-            matches  = response.headers['content-disposition']&.match(/filename="([^"]*)"/)
-            filename = matches.nil? ? parsed_url.path.split('/').last : matches[1]
+            content_type = response.headers['content-type']&.split(/\s*;\s*/).first
```
Gargron: >  Do not chain ordinary method call after safe navigation operator.

app/models/concerns/remotable.rb
```diff
@@ -24,14 +24,16 @@ def remotable_attachment(attachment_name, limit)
...
+            extname      = MIME::Types[content_type].first&.extensions.first
```
Gargron: > Do not chain ordinary method call after safe navigation operator.

Gargron (changes requested):

app/models/concerns/remotable.rb
ykzts: 857407a 💫

Gargron (approved):

PR #7734, opened 2018-06-04
Fix context performance by partially reverting #7083

nightpool:
This was causing a major performance lag when requesting status context on cybrespace, and led to many timeouts. it's already been reverted by pawoo.

Do we know why this wans't causing problems on mastodon.social? It feels like it would have been really obvious....

### DIFF
--- a/app/models/concerns/status_threading_concern.rb
+++ b/app/models/concerns/status_threading_concern.rb
@@ -51,12 +51,12 @@ module StatusThreadingConcern
   end
 
   def descendant_statuses(limit, max_child_id, since_child_id, depth)
-    Status.find_by_sql([<<-SQL.squish, id: id, limit: limit, max_child_id: max_child_id, since_child_id: since_child_id, depth: depth])
+    descendants_with_self = Status.find_by_sql([<<-SQL.squish, id: id, limit: limit, max_child_id: max_child_id, since_child_id: since_child_id, depth: depth])
       WITH RECURSIVE search_tree(id, path)
       AS (
         SELECT id, ARRAY[id]
         FROM statuses
-        WHERE in_reply_to_id = :id AND COALESCE(id < :max_child_id, TRUE) AND COALESCE(id > :since_child_id, TRUE)
+        WHERE id = :id AND COALESCE(id < :max_child_id, TRUE) AND COALESCE(id > :since_child_id, TRUE)
         UNION ALL
         SELECT statuses.id, path || statuses.id
         FROM search_tree
@@ -68,6 +68,8 @@ module StatusThreadingConcern
       ORDER BY path
       LIMIT :limit
     SQL
+
+    return descendants_with_self - [self]
   end
 
   def find_statuses_from_tree_path(ids, account)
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
:thinking:

nightpool:
fixed tests

akihikodaki:
I  and @abcang still have no idea why this fixed the issue and am a bit worrying that other changes to those queries may have similar issues but have not just surfaced yet. It is possible, considering that it reproduced on particular instances and not on others.

Pawoo's database does not seem to have something notable except that it is big. The size should be irrelevant considering other large instances do not have the issue as far as we know.
It is concerning for me that PostgreSQL version of Pawoo is a bit old when thinking about this kind of difficulty of bug reproduction.

@nightpool Can you describe your PostgreSQL version, or something particular of your database if any?

nightpool:
We're using postgres 9.5

akihikodaki:
> postgres 9.5

Hmm, the version is suspicious. Thanks for the information.

app/models/concerns/status_threading_concern.rb
```diff
@@ -51,12 +51,16 @@ def descendant_ids(limit, max_child_id, since_child_id, depth)
...
+    depth += 1 if depth.present?
+    limit += 1 if limit.present?
+
+    descendants_with_self = Status.find_by_sql([<<-SQL.squish, id: id, limit: limit, max_child_id: max_child_id, since_child_id: since_child_id, depth: depth])
       WITH RECURSIVE search_tree(id, path)
       AS (
         SELECT id, ARRAY[id]
         FROM statuses
-        WHERE in_reply_to_id = :id AND COALESCE(id < :max_child_id, TRUE) AND COALESCE(id > :since_child_id, TRUE)
+        WHERE id = :id AND COALESCE(id < :max_child_id, TRUE) AND COALESCE(id > :since_child_id, TRUE)
```
ThibG: Hm, I'm afraid that doesn't make much sense.
This will always select either the root status (if `max_child_id` and `since_child_id` or undefined or define a large enough range to include the root status), or no status at all, while it was supposed to select a subset of the direct replies to the current status.
So if my understanding is correct, `max_child_id` and `since_child_id` have been broken ever since this commit.

On the other hand, I'm not sure how useful `max_child_id` and `since_child_id` are for actually paginating… (especially since it's not currently used in a meaningful way anywhere in Mastodon)

nightpool: not sure I understand this comment. we're still recursing on in_reply_to_id, we're just starting with the the root status instead of replies to the root status)

ThibG: Yes it is supposed to be recursive, but since_child_id and max_child_id are supposed to apply to the direct children of the root toot, but it will be applied to the root toot instead.

Meaning this part of the query will either select the root status or no status at all. And the query itself will either return the whole thing, or no status at all, since_child_id and max_child_id are not used in a meaningful way.

PR #7739, opened 2018-06-05
Sort Elasticsearch result in descending order of created_at

ushitora-anqou:
Present code shows the result of Elasticsearch in random order, but sorting the result toots by their creation time helps us find toot we want more easily.


Gargron:
I believe the order of the results is by relevance, no?

ushitora-anqou:
Oh really? I overlooked that, sorry.

 I'll close this PR.

Issue #7226, opened 2018-04-22
Vagrantfile: Clearer error when/if RVM fails to install


marrus-sh:
When attempting to use Mastodon via Vagrant for development, the following line failed:

```bash
gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
```

(This happens sometimes because `keys.gnupg.net` occasionally times out, or so I gather; when I ran the command manually it worked fine.)

This prevented RVM from installing and naturally meant that any of the following steps which required ruby didn't work.

However, there was no clear indication that this was the problem, making it somewhat difficult to debug. If RVM fails to install, the provisioning script should terminate immediately and notify the user to probably just try again.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Aside: Nobody I know who contributes to Mastodon regularly uses Vagrant, so nobody keeps it up to date or notices in time when it breaks. If anyone wants to become the person who maintains Vagrant, be my guest.

trwnh:
this didn't happen last time i tried the Vagrantfile, so i'm closing this as stale / can't reproduce. if there are any issues with the current Vagrantfile then please file those as new issues

PR #7740, opened 2018-06-05
Fix Vagrantfile

ushitora-anqou:
Present `Vagrantfile` has some problems:

- RVM very often fails to be installed (related to #7226 ).
  - Using rbenv instead works fine.
- Memory and CPUs are not sufficient to run Mastodon.

This PR fixes them.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -42,14 +42,14 @@ sudo apt-get install \
   libpam0g-dev \
   -y
 
-# Install rvm
-read RUBY_VERSION < .ruby-version
-gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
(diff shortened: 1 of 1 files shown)


Gargron:
I remember someone replaced rbenv with rvm because rbenv compiles from source and rvm uses binaries, which saves time. I don't remember who it was but should probably tag them into this for feedback.

Gargron:
Yep, #2346 / @ineffyble

ushitora-anqou:
When I tried `vagrant up` outside, I couldn't install RVM, so wrote this patch. But I tried again now at home, and I could install RVM correctly. It may depend on my network environment such as proxy. Also, I found that this patch doesn't use `.ruby-version`. This PR shouldn't be merged. I'll close it.

Thanks for your quick reply and sorry for my lightheaded acts.

PR #7755, opened 2018-06-07
Hide media timeline tabs behind column header toggle

Gargron:
image
image


Gargron:
I don't quite like how this looks. The alternative is to... hide the tabs again...

trwnh:
Why not simply make it a toggle like all the other toggles? That would at least be a consistent design pattern throughout the app.

image

PR #7758, opened 2018-06-08
Move footer of getting started to bottom

ykzts:
before | after
-|-
screenshot on Getting started | screenshot on Getting started


Gargron:
Idk, I quite like the column being shorter on desktop. It might have a positive effect in being less overwhelming to new users since it draws the attention to the other columns instead so it's like 3 instead of 4 to look at.

ykzts:
Okay! I see.

PR #7763, opened 2018-06-09
Add tests for intents_controller

shuheiktgw:
Added tests for intents_controller! 👍

### DIFF
--- a/app/controllers/intents_controller.rb
+++ b/app/controllers/intents_controller.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 class IntentsController < ApplicationController
-  def show
-    uri = Addressable::URI.parse(params[:uri])
+  before_action :check_uri
+  rescue_from Addressable::URI::InvalidURIError, with: :handle_invalid_uri
 
+  def show
     if uri.scheme == 'web+mastodon'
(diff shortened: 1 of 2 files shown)


app/controllers/intents_controller.rb
```diff
@@ -15,4 +17,18 @@ def show
 
     not_found
   end
+
+  private
+
+  def check_uri
+    not_found if uri.blank?
```
shuheiktgw: Currently there is a problem that when accessing `localhost:3000/intent` it raises a error, so 
 handle it

app/controllers/intents_controller.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 class IntentsController < ApplicationController
-  def show
-    uri = Addressable::URI.parse(params[:uri])
 
+  before_action :check_uri
+  rescue_from Addressable::URI::InvalidURIError, with: :handle_invalid_uri
```
shuheiktgw: Also there is another problem that if the uri is invalid, `Addressable::URI.parse` raises  `Addressable::URI::InvalidURIError`. So I added error handling for that error.

Gargron (approved):

PR #7765, opened 2018-06-09
Fix domain hiding logic

Gargron:
**Retire the block & unblock ("soft-blocking") behaviour from the code**
There is a lot of risk of out-of-order arrival of such messages. Mastodon understands follow rejection since e85287284611f7de431d9c51353125c265ebfe3f (v2.3.0), this can cut down on traffic and funky state desync between instances.

**Reject follow from other server when hiding a domain**
Instead of silently dropping passive follow relations locally, also send a follow rejection.

**Unify follower removal from "Authorized followers by domain" page with "domain blocks"**
Rejecting followers and follow requests now happens asynchronously and is a common behaviour between account domain blocks and specifically only removing followers from a certain domain.

I don't think that the "Purge" button on "Authorized followers" page warrants creating a domain block, though, since the user probably doesn't expect to hide all content from that domain with that action.

**Adjusted confirmation dialog for "Hide everything from example.com"**

> Are you really, really sure you want to block the entire {domain}? In most cases a few targeted blocks or mutes are sufficient and preferable. You will not see content from that domain in any public timelines or your notifications. Your followers from that domain will be removed.

___

State of "Hide everything from example.com" after this PR:

- Remove followers from example.com
- Remove follow requests from example.com
- Hide toots from example.com in local/federated/hashtag timelines
- Notifications / mentions from example.com are dropped
- Does *not* make *you* unfollow someone you've been following from example.com
- If you follow someone from example.com, you can still get notifications (like mentions) from them
- You cannot be followed by someone from example.com even if unlocked, as the server will not auto-approve their follow request


trwnh:
A few concerns:

- What are the potential side-effects of following someone that cannot follow you back? This sounds more like a block than a hide.
- How can a user differentiate between "I don't want to see things from this domain" (e.g. switter, botsinspace) versus "I don't want my statuses to reach this domain" (e.g. unsafe.space, thechadzone, freezepeach)?

Gargron:
>What are the potential side-effects of following someone that cannot follow you back? This sounds more like a block than a hide.

There's plenty of following without being followed back in the fediverse, so this doesn't stand out to me. It's not the nicest thing but seems to be within the scope of what people want from this entire feature.

>How can a user differentiate between "I don't want to see things from this domain" (e.g. switter, botsinspace, more of a mute) versus "I don't want my statuses to reach this domain" (e.g. unsafe.space, thechadzone, freezepeach, more of a block)?

I would add an extra toggle but I really don't want to stick that into 2.4.1. This patch has been trouble enough already, trending hashtags feel like they were months ago but it's only been like a few days or something.

PR #7767, opened 2018-06-09
Fix has_one association on Web::PushSubscription

Gargron:


nightpool (approved):

nightpool:
ref: https://github.com/tootsuite/mastodon/pull/7764

PR #7768, opened 2018-06-09
Fix background color of emoji-mart-bar in light theme

lynlynlynx:
|Before|After|
|--|--|
|image|image|

I cant find out why unselected emoji-mart-anchor-bar appears only in mastodon-light theme...

### DIFF
--- a/app/javascript/styles/mastodon-light/diff.scss
+++ b/app/javascript/styles/mastodon-light/diff.scss
@@ -47,6 +47,14 @@
   background: darken($ui-base-color, 6%);
 }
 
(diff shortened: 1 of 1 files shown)


lynlynlynx:
Thanks, finally I find out why anchor appears. It occurs due to the display scaling. I adjust anchors in #7772 .

> By the way, the search element is supposed to be very subtle compared to the background. I don't think a grey input background works well.

Thats right. I make emoji-mart-search input's background color same as in default theme.

image

Gargron (approved):

PR #7780, opened 2018-06-10
Dockerfile: run asset precompilation as a build step

kliu128:
This saves Docker users from having to run `rails assets:precompile` manually upon upgrade, by embedding compiled assets and packs in the image.

See glitch-soc/mastodon#536 for some prior discussion.

/assets and /packs should no longer be volumes, since they are embedded in the image now. This might require documentation updates as well.

The build step uses randomly-generated, placeholder `OTP_SECRET` and `SECRET_KEY_BASE` to run the compilation process. (Let me know if there's anything wrong with this approach; I'm not sure if these values are used for anything during compiling.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,6 @@ VOLUME /mastodon/public/system
 
 USER mastodon
 
-RUN OTP_SECRET=$(bundle exec rails secret) SECRET_KEY_BASE=$(bundle exec rails secret) bundle exec rails assets:precompile
+RUN OTP_SECRET=precompile_placeholder SECRET_KEY_BASE=precompile_placeholder bundle exec rails assets:precompile
 
 ENTRYPOINT ["/sbin/tini", "--"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,8 +77,10 @@ COPY . /mastodon
 
 RUN chown -R mastodon:mastodon /mastodon
 
-VOLUME /mastodon/public/system /mastodon/public/assets /mastodon/public/packs
+VOLUME /mastodon/public/system
 
 USER mastodon
+
+RUN OTP_SECRET=$(bundle exec rails secret) SECRET_KEY_BASE=$(bundle exec rails secret) bundle exec rails assets:precompile
 
 ENTRYPOINT ["/sbin/tini", "--"]


ykzts:
ref #3436

Reverite:
So `docker-compose.yml` will also have to be modified to remove mounting `/assets` and `/packs`, but it looks good to me. My command that I use for my build looks similar: `RUN SECRET_KEY_BASE=_ OTP_SECRET=_ bin/rails assets:precompile`

The only problem will be that Webpack seems to read `CDN_HOST` (per [this file](https://github.com/tootsuite/mastodon/blob/master/config/webpack/configuration.js#L37)).

I guess setting the asset host can be done by setting an HTML tag for what CDN_HOST usually is, i.e. `<meta name="cdnHost" content="assets.mastodon.dev">` and replacing where `process.env.CDN_HOST` is read from in any JS files can be replaced with `document.querySelector("meta[name='cdnHost']").getAttribute("content");`

Dockerfile
```diff
@@ -77,8 +77,10 @@ COPY . /mastodon
 
 RUN chown -R mastodon:mastodon /mastodon
 
-VOLUME /mastodon/public/system /mastodon/public/assets /mastodon/public/packs
+VOLUME /mastodon/public/system
 
 USER mastodon
 
+RUN OTP_SECRET=$(bundle exec rails secret) SECRET_KEY_BASE=$(bundle exec rails secret) bundle exec rails assets:precompile
```
Info-Screen: I would recommend to use a static secret here (e.g. `precompile`, `invalid` or just `_`) to decrease the build time.

Pneumaticat: Good point! Will do.

Gargron (approved):

hugogameiro:
Won't this cause a run `rails assets:precompile` every time you restart docker-compose even if you haven't upgraded and don't need assets to be precompiled?

Gargron:
@hugogameiro No, it's executed during `build`, that is, if you get your images from the Docker Hub, it's already done.

insom:
I was running with this commit and my CSS was pretty broken until I removed the two volumes from my `docker-compose.yml`, too. Just leaving this here in case it helps anyone.

angristan:
This is nice because faster updates, however I think Nginx/Apache are better at handling static ressources requests than Puma.

Info-Screen:
You should still be able to do that.

Shadowghost:
You can still use Apache or nginx as reverse proxy, but you need to revert [this commit](https://github.com/tootsuite/mastodon/commit/8c279b1648fc0a07b5c5d4a0e18526bcb51c7700) to make it work again. Should still work with assets pre compilation on build.

angristan:
That's true :smile:

PR #7781, opened 2018-06-10
little improved in getting started column

rinsuki:
- align height with other columns
- navigation items list is now scrollable if height of window is narrow

<details>
<summary>diff images</summary>

image
image

</details>


ykzts:
https://github.com/tootsuite/mastodon/pull/7758#issuecomment-395629748

PR #7789, opened 2018-06-12
Add profile options on compose form

chr-1x:
As of 2.4.1 the only way to access mutes, blocks, and domain block options is from a submenu when viewing your own profile

This PR adds an action button to the compose form to expose the same options in a more convenient and discoverable way. It provides essentially the same menu options without reintroducing clutter to the getting started menu and using an intuitive connection (your profile information is already shown in the compose form, so options related to it, such as your mutes and domain blocks, make sense to access from this button).

May need some UI work, the button is a little low contrast right now. Also should be considered whether the "edit profile" link under your username is now redundant considering the same option is available in the new menu.

image
image

### DIFF
--- a/app/javascript/mastodon/features/compose/components/action_bar.js
+++ b/app/javascript/mastodon/features/compose/components/action_bar.js
@@ -44,7 +44,7 @@ export default class ActionBar extends React.PureComponent {
     return (
       <div className='compose__action-bar'>
         <div className='compose__action-bar-dropdown'>
-          <DropdownMenuContainer items={menu} icon='bars' size={24} direction='right' />
+          <DropdownMenuContainer items={menu} icon='ellipsis-v' size={24} direction='right' />
         </div>
       </div>
     );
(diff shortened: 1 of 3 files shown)


Gargron:
About "Edit profile": Even if it's redundant, removing it would leave empty space that would look worse. Also, it's important enough that its discovery should be prioritized (people might not think of clicking on a dropdown icon). So yes, keeping it this way is good.

Does this play well with the "unfocus textarea" ("hide keyboard") icon that appears in that place on mobile when the textarea is focused?

Also since the hamburger icon is overused between getting started and here and also symbolizing lists, could you use the vertical ellipsis icon? (Three vertical dots)

chr-1x:
Latest commit makes it play nicely with the X button. 
pr7789a

Also now uses the ellipsis-v icon instead of hamburger button. I do want to point out that this is inconsistent with the icon used on profiles, so maybe it should be changed both places? 
image

Gargron (approved):

PR #7797, opened 2018-06-13
Hide status content when spoiler text is not empty (regression #7395)

### LINKED ISSUE #7788: Statuses with CW (Content Warning) from Pleroma are sometimes shown expanded by default.
From a Mastodon instance, if you see a status from (some?) Pleroma instances, with a CW, it is sometimes shown expanded by default even if you've had no previous interactions with the user, thread or message.

Probably easiest to reproduce by pasting a link to a specific status into the search box. I tried this on a newly created account on mastodon.social, loading e.g. this status by @kaniini about possibly prerendered footage in E3 https://pleroma.site/objects/dbccd46c-fa80-4cdc-92c2-5d4eff8a8590 (which showed as open by default), and one of my toots about rds databases on amazon: https://glitch.social/@kepstin/100188061069476651 (which showed as closed by default).

I have no idea why this is happening, but it seems to be an interop problem between the different implementations.

ykzts:
Resolve #7788

### DIFF
--- a/app/javascript/mastodon/actions/importer/normalizer.js
+++ b/app/javascript/mastodon/actions/importer/normalizer.js
@@ -50,13 +50,14 @@ export function normalizeStatus(status, normalOldStatus) {
     normalStatus.spoilerHtml = normalOldStatus.get('spoilerHtml');
     normalStatus.hidden = normalOldStatus.get('hidden');
   } else {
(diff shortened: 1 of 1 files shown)


kaniini:
that does not resolve #7788, the statuses in question had spoiler text.

ykzts:
Ah, sorry. I mistook commit message.

Gargron (approved):

Issue #6361, opened 2018-01-25
want to get "Opt-out of search engine"


osapon:
I would like information of "Opt-out of search engine indexing" to be included in API account entity and JSON of user information.
Even if I want to create an external search engine, I need to obtain from the user page whether it is opted out.
It is easy to use, if I can get it with toots or user information.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

koyuawsmbrtn:
well, such a feature actually exists

![image](https://user-images.githubusercontent.com/4673062/35460684-297312fc-02e5-11e8-98ad-7f39936ad221.png)

i actually don't know what you mean. if the feature is enabled a meta tag called `<meta content='noindex' name='robots'>` will be placed into the profile pages and status html  code. there's no need for an extra api.

osapon:
It is wasteful communication to acquire the user page again after getting toots.
If the account entity contained in toots contains information, it can be reflected immediately.

ThisIsMissEm:
Perhaps it'd be an idea to use the X-Robots-Tag header? https://developers.google.com/search/reference/robots_meta_tag

osapon:
Currently, only HTML has "no-robot" indication.
The account data obtained from the API has no "no-robot" attribute.

I may make application that uses toots and accounts, I can not judge whether or not to show that user to the search engine crawler from the acquired JSON.
After acquiring with JSON, accessing the HTML again only to acquire the "no-robot" attribute also loads the instance.

ThisIsMissEm:
Right, the X-Robots-Tag header on API responses would include either `noindex` or another robots option.

ThisIsMissEm:
See also #7164 — potentially there needs to be a way for a API client to specify it is a bot, without relying on user agent allow, perhaps: `X-Requested-With: Bot` or `X-Requested-With: Crawler` and the API would return different or no content based on that?

trwnh:
`noindex` is exposed via API for local accounts as of v4.0, although a more dedicated opt-out might come later for per-post searchability.

PR #7822, opened 2018-06-16
Add `noindex` flag to API repsonse

rinsuki:
ref: #6361
image

if not local user, return value is null.

### DIFF
--- a/app/serializers/rest/account_serializer.rb
+++ b/app/serializers/rest/account_serializer.rb
@@ -49,8 +49,8 @@ class REST::AccountSerializer < ActiveModel::Serializer
   end
 
   def noindex
-    user = object.user
-    user.settings['noindex'] if user
+    return true unless object.user
+    object.user.settings['noindex']
   end
 
   def moved_and_not_nested?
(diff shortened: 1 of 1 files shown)


Gargron:
The usefulness of this data when it's incomplete for remote users is arguable. I think for this to be done completely, the property must become part of the ActivityPub actor (and therefore move from settings to accounts table?). The property would also be a new addition to AP vocabulary and therefore warrants extended discussion.

rinsuki:
yes, this is incomplete for remote users. however, in some cases, it is really useful. for example, auto opt-out from search engine that crawl only local timeline.

app/serializers/rest/account_serializer.rb
```diff
@@ -48,6 +48,11 @@ def header_static
     full_asset_url(object.header_static_url)
   end
 
+  def noindex
+    user = object.user
+    user.settings['noindex'] if user
```
nightpool:
```ruby
return true unless object.user
object.user.settings['noindex']
```

rinsuki:
@nightpool for example, <http://mastodonsearch.jp/>,  <https://tootsearch.chotto.moe/>, etc crawls public timeline.

tateisu:
tootsearch reads FTL.

hakaba-hitoyo:
The `discoverable` flag (#11688) suits for this purpose?

Gargron:
> The `discoverable` flag (#11688) suits for this purpose?

Maybe. But `discoverable` is not the same value as "opt out of search engines". It would have slightly different meaning.

Issue #7849, opened 2018-06-19
RESULTS_LIMIT hardcoded


pooza:
In Api::V1::SearchController, the RESULTS_LIMIT constant is hard coded.
It should be specified from environmental variables etc.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #7850, opened 2018-06-19
Result limit hardcoded

pooza:
refs #7849


pooza:
Sorry, this is a mistake.

PR #7851, opened 2018-06-19
RESULT_LIMIT can be set from environment variables

pooza:


nightpool:
what is the motivation for this change? more then 5 results would be very awkward to navigate through with our current scrollbar. 

what is your use case for making this change?

pooza:
Even when using it from the Web client, I feel that five toots are not enough at the upper limit.
Also around me, several instance administrators have rewritten RESULTS_LIMIT.

Also, search API is not for web clients only, is it?

I think that it is not good that the upper limit of the number of searches that may change depending on the application is hard-coded.

nightpool:
>Also, search API is not for web clients only, is it?

this is not an argument for changing the search limit globally, it's an argument for allowing RESULTS_LIMIT to be customized by a API parameter. If you wish to contribute code to do so, it would be welcome. However, I don't think the solution you gave (an environment variable) is good enough, because it obscures the bigger problem by using a band-aid solution

PR #7858, opened 2018-06-19
Update sprockets gem

zunda:
https://blog.heroku.com/rails-asset-pipeline-vulnerability


nightpool:
>Note: The default value of this setting that ships with Rails in production.rb is false. By default, Rails apps running in production mode are not vulnerable to this exploit.

ykzts (approved):

Issue #7833, opened 2018-06-17
On firefox, corrupt content du to a network error "NS_ERROR_FILE_NO_DEVICE_SPACE"


Psycojoker:
Hello,

With @wxcafe, on [his instance](https://social.wxcafe.net) we both regularly encounter a weird bug on firefox where is refuse to load the page because of a "corrupt content error".

We haven't find any log on the server side and this screenshot is the best and only log I could find about the error message:

![2018-06-17-043911_1278x512_scrot](https://user-images.githubusercontent.com/41827/41512261-eef71798-7285-11e8-9982-4e3aa310c3df.png)

To roughly translate it:

> Impossible to load <url>. A worker service communicated an object « promise » to FetchEvent.respondWith() which has been rejected with the reason « NS_ERROR_FILE_NO_DEVICE_SPACE »

It doesn't seems to be an instance configuration problem because several instances are hosted on the same server (but I might be mistaking). 

I try doing some research but couldn't find something really useful (there is this boring ticket https://bugzilla.mozilla.org/show_bug.cgi?id=935947) and absolutely no description here (thx mozilla) https://developer.mozilla.org/en-US/docs/Mozilla/Errors

Could it be that Mastodon doesn't properly clean the cache or something like that related to web workers taking too much place?

I don't know if that makes any difference but I'm using 2 custom users lists/columns on this instance.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a 2.4.1 on social.wxcafe.net

PR #7863, opened 2018-06-20
Return collect response when fail put cache to storage

hinaloe:
When failing ServiceWorker Cache putting (ex: `Quota exceeded`), the response will not return to front process. So images or page cannot show.

Errors should be ignored.

(There is a possibility to solve #7833 )

Example problem:

image


Gargron (approved):

hinaloe:
Thanks!

PR #7875, opened 2018-06-22
TOR federation in docker-compose

ghost:
This PR adds an optional section to the docker-compose.yml that shows how to enable tor federation.  I hope that by adding this into the docker-compose it will encourage more instances to enable tor federation.  I have been running this on my [instance](https://banana.dog) for a while so I am pretty sure it is stable.  Thanks to [Sir Boops](https://mastodon.sergal.org/@Sir_Boops) who helped me set this up and runs it on on their instance as well.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -80,6 +80,24 @@ services:
     volumes:
       - ./public/packs:/mastodon/public/packs
       - ./public/system:/mastodon/public/system
+## Uncomment to enable federation with tor instances along with adding the following ENV variables
+## http_proxy=http://privoxy:8118
+## ALLOW_ACCESS_TO_HIDDEN_SERVICE=true
+## HIDDEN_SERVICE_VIA_TRANSPARENT_PROXY=true
+#  tor:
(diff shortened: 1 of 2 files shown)


Gargron (approved):

ykzts:
cc @hcmiya

hcmiya:
I am not familiar with Docker, but if it follows the procedure of composing an HTTP proxy with privoxy and tor, I think there is no problem.

docker-compose.yml
```diff
@@ -80,6 +80,24 @@ services:
     volumes:
       - ./public/packs:/mastodon/public/packs
       - ./public/system:/mastodon/public/system
+## Uncomment to enable federation with tor instances along with adding the following ENV variables
+## http_proxy=http://privoxy:8118
+## ALLOW_ACCESS_TO_HIDDEN_SERVICE=true
+## HIDDEN_SERVICE_VIA_TRANSPARENT_PROXY=true
```
hcmiya: if `http_proxy` is used, `HIDDEN_SERVICE_VIA_TRANSPARENT_PROXY` is not used so it should be removed.

PR #7912, opened 2018-06-29
Fix JS security issue in unescapeHTML()

MaciekBaron:
As far as I understand, the purpose of the `unescapeHTML()` function is to remove any HTML tags, and replace paragraphs and break line tags with new line characters.

The current implementation can be exploited using HTML entity characters.

`unescapeHTML('<script>evil()</script>')` will return `evil()`, however `unescapeHTML('&lt;script&gt;evil()&lt;/script&gt;')` will return `<script>evil()<\script>`. This is because in the process of setting the `innerHTML` and then requesting the textContent, those HTML entities get changed back into their original representation. This leaves the gate open for a potential XSS attack, as a developer might assume the function returns a safe string.

This fix prevents that from happening, by first setting the innerHTML once more to the textContent, and then requesting the textContent, thus removing any HTML that has been encoded using HTML entities.

### DIFF
--- a/app/javascript/mastodon/utils/html.js
+++ b/app/javascript/mastodon/utils/html.js
@@ -1,5 +1,6 @@
 export const unescapeHTML = (html) => {
   const wrapper = document.createElement('div');
   wrapper.innerHTML = html.replace(/<br\s*\/?>/g, '\n').replace(/<\/p><p>/g, '\n\n').replace(/<[^>]*>/g, '');
+  wrapper.innerHTML = wrapper.textContent;
   return wrapper.textContent;
 };


MaciekBaron:
`test-ruby2.5` seems to be failing due to something not related to this change (`Webpacker can't find locale_en.js…`)

Gargron:
Hold on, let's walk through this. We get *trusted* HTML from the server. It's been sanitized, it contains p, a, br and span tags, everything else is HTML-encoded e.g. as `&gt;`

The unescapeHTML function is needed to get rid of the p, a, br and span tags and display just the plain text inside the desktop and push notifications e.g. body of Notification object. Are we on the same page here?

MaciekBaron:
Right, I see, so this function would never be used for anything else?
And parsing HTML-encoded entities into their string representation is expected behaviour?

If so, we can close this pull request. I was just worried that at some point in the future someone might use this function for a different purpose, other than the Notification object.

Gargron:
Examining the usage of unescapeHTML in the source code, here are the places where it's used:

- actions/importer/normalizer.js for value_plain, which goes inside a title attribute
- actions/notifications.js, which goes inside a Notification body
- reducers/compose.js, which goes inside a textarea

MaciekBaron:
Thanks. I think the code is not exploitable at the moment, but could be in the future, if someone used that function for a new purpose. I guess in that case this is just a matter of expectations: if it's fine for the function to return HTML in it (or it is in fact expected to), then we can of course leave it as it is. Otherwise I think we should get rid of it, just in case.

Gargron:
You could add a comment above the function warning not to use it on untrusted HTML

MaciekBaron:
That sounds like a reasonable solution if we want to keep the tags in the content. In that case I'll close this PR.
Thanks a lot!

PR #7918, opened 2018-06-30
Re-add follow recommendations API

Gargron:
New REST API:

    GET /api/v1/suggestions

Was previously removed in 8efa081f210d72ed450c39ac4cde0fd84fb3d3fb due to the Neo4J dependency also being removed.

The algorithm uses interactions.

**What interactions?** If you engage with a person you're not following, it bumps their score on your hidden list. Between replies, favourites and reblogs, the weight of the score bump is different. The idea is that you may have lots of positive interactions with a person, but forget to follow them.

**What problem does this solve?** Help regular users improve their home feed by reminding them to follow people they have positive experiences with.

__

**Update:** I am removing triadic closures. I realized because of limited data for remote users (we only know which local accounts they are following) the results would always be skewed towards local accounts, so it's better to not do it at all.


PaulFree14:
I would like to see within the follow recommendations also a method being used that has it focus on breaking silos.

ykzts:
Does the Web UI implementation later?

albjeremias:
i want to disable this on my instance...

Gargron:
Please explain why you think that you want to disable this on your instance.

albjeremias:
Well i don't find this feature useful, I think this is the best explanation till now:
Being on mobile I haven't read code, but personally I'd need explicit opt out so I can say both  "don't show me any recommendations" and "don't recommend me".

(Besides locking the account. )

> Not really a fan of algorithmic recommendation systems, but I can see others might like them. Personally wouldn't wanna participate, though.
https://cybre.space/@minx/100298178840709626

can I ask you the same? :) why not have this option as opt-in/opt-out per instance?

Gargron:
@ykzts Yes, people will likely argue about the UI so I want to split them.

ykzts (approved):

elomatreb:
If we absolutely must have this feature, I am still very much missing any mention of being able to permanently dismiss users from the recommendations.

nevillepark:
I'm also not seeing an option to exclude oneself from the algorithm, or to enable/disable it on an instance-wide basis.

calmasis:
Hi! If you're implementing this, could you include an option for users to turn it off? I'm certain you don't have bad intentions, but I've seen so many spammy, data-mining sites using follow recommendations that they're irrevocably associated in my mind. I'd quite strongly prefer not to have to see follow recommendations, if that is an option. Thank you :)

lindner:
I highly suggest that this Follower Suggestion design be peer-reviewed.

Content recommendation systems have a huge effect on how communities grow and evolve. One thing I like about Mastodon is that user recommendations are done by humans, not machines. This go-slow approach is more resilient to algorithms (even the simple one presented here..)

Here's a recent paper that discusses how recommendation systems can reinforce bias.

[https://blog.acolyer.org/2018/05/24/algorithmic-glass-ceiling-in-social-networks-the-effects-of-reco...](https://blog.acolyer.org/2018/05/24/algorithmic-glass-ceiling-in-social-networks-the-effects-of-recommendation-on-social-diversity/)

So before this gets unleashed on the world please do your research and study the available literature. Get experts to review it and listen to the people that have been doing this for a long time. You can then avoid repeating the mistakes of the past.

(Originally published at: http://www.1500wordmtu.com/2018/i-highly-suggest-that-this-follower-suggestion)

inmysocks:
I would really rather not have this sort of compiled information easily accessible. I thought the point of mastodon was that it wouldn't do things like track the people who use it.

Also it seems custom made as a really useful tool to enable harassment. When you are feeling belligerent instead of having to remember who that random depressed person you tell to slit their wrists when you notice them you have a nice list to pick from.
Yes, everyone could just block the harassers but that moves the responsibility onto the abused party and it makes harassment easier. Things that hurt vulnerable people and help those trying to hurt them are probably a bad idea in social media.

joyeusenoelle:
Recommendation: in `account_interactions.rb`, modify `remove_potential_friendship()` as such:

    def remove_potential_friendship(local_account, other_account)
        PotentialFriendshipTracker.remove(local_account.id, other_account.id)
    end

Then add a second `remove_potential_friendship()` call to `mute!` and `block!` such that *both* users are removed from each other's potential friendships.

joyeusenoelle:
Additional recommendation: I don't see it in the code, and I'm not sure where this happens in the codebase, but un-favoriting and un-boosting a toot should have the effect of un-bumping the potential friendship by the same amount. This prevents manipulation of the system by repeatedly favoriting and unfavoriting or boosting and unboosting a status. (I'm not sure what the goal of such manipulation would be, but I'm also not a malicious user.)

PaulFree14:
>What problem does this solve?

the problem it solves seem to me rather fictional then really existing. (that's my personal understanding of it sofar)

if I often boost a persons toots why should I forget that I often boost that person?

If I forget, which is ok, I latest can remember that I boost a post from that person, when I boost it. And can then simply decide if I want to follow or not.

I only can see this feature being used at one point in future as in twitter, to promote specific ppl. and by that creating communities that are biased through the software.

PaulFree14:
Improofing discoverability for those who want it is great. 
Having it mandatory propably not. 

Having a look how it's done in other projects is great to get ideas and to not reproduce same misstakes, if there are any.

 As a positive example how it can be done I found hubzilla:

- plugin that user can turn on/off showing suggestions of whom to follow
- letting the user decide to not be suggested to others
- having a local searchable database where people can, if they want, be listed based on their interest, where they life...etc.
- it's everyones own decision if to be listed or not and what to be listed
- you can search within multible local database at the same time

joyeusenoelle:
Paul, if you remember 100% of the people whom you don't follow but with whom you've interacted, this feature isn't intended for you.

Since it *by design* can't suggest anyone you don't interact with, and suggests people proportionally to how much you interact with them, it can't be used to promote specific people.

And since it's only an API endpoint and not a UI element, it can't be mandatory.

It sounds like you're arguing against a feature that isn't actually the one under discussion.

PaulFree14:
I've been sharing concerns related to this feature. They don't need to be true in it's current state, that I have them. 
(thx for listening, I'm out)

Oreolek:
This is a naive and very harmful approach. 

Imagine this scenario: there is a brand with a custom instance. It wants to drive engagement, it hires a SMM manager. The manager makes a lottery: "reblog this toot for a chance to win!" Now every user who reblogged that would see the brand in their recommendations, every day, until they engage in several long conversations or reblog something else to clear the list.

Obscuring recommendation algorithms is a security measure, because there are people who actively exploit them. Open source algorithms would be exploited in no time using unethical tricks polished in other social networks.

joyeusenoelle:
If the user in question interacts so little that a single boost can keep that account in their recommendations for a long period of time, then yes, that might happen. But simply muting the brand - even temporarily - will remove them from your list of recommendations until the next time you interact with them. And since two favorites are worth a boost, it's easy to drive a single-boost interaction down the chart.

And again, in the current implementation, you only see recommended users if you explicitly ask for them, and I agree that that should remain the case if a UI element is added. (Make it a column that goes in the "utility" slot - the Getting Started/LTL/FTL/etc column - and can be summoned and dismissed as easily as those are.)

Laurelai:
This is an exploitable feature that will alter the behavior of users to act in more attention grabbing but toxic ways.

bunnybooboo:
For privacy reasons I'd prefer users had to OPT-IN to such an algorithm. As an additional ethical and debiasing measure I believe it's benefitial for an instance to also be required to OPT-IN. I'd prefer purposeful, selective, granularity regarding my personal acct used in any promotional activity.

In my opinion this is territory which has made other sites destructive. I recognise MANY users will want it. Perhaps even MOST users. Those blurry edge cases should be considered in the design contemplation though.

There is a myriad of reasons why both options (inclusion of this feature vs exclusion of this feature) might prove successful. Without a strategic scoring matrix in place of the pluses and minuses across a number of issues, a user like myself would likely only bring individual biased opinion. 

What I feel needs to be considered beyond an agile methodology is the likely OUTCOMES: What would be the likelihood of when people overuse such feature? What about if they managed to be able misuse or abuse it? Who would likely benefit from its addition? And why is this benefitial for the wider community? Who would experience a decline in their experience? How might this negative be countermeasured?

Those kind of questions are often missing from controversial feature sets like this.

hakaba-hitoyo:
> I only can see this feature being used at one point in future as in twitter, to promote specific ppl. and by that creating communities that are biased through the software.

> Imagine this scenario: there is a brand with a custom instance. It wants to drive engagement, it hires a SMM manager. The manager makes a lottery: "reblog this toot for a chance to win!" Now every user who reblogged that would see the brand in their recommendations, every day, ....

This is serious problem, but in our decentralized social network, we can anytime go away from the evil instance which abuses its recommendation.

https://pawoo.net provides its _who to follow_ panel since 2017 May, which is actually the list of the famous illustrators. Japanese or other Pawoo users rarely complain about this feature, but the #7918 seems better (more gentle or moderate) approach than Pawoo's.

I hope the official UI of the /api/v1/suggestions is provided very soon. In the future, the instances can provide their recommendation engine which is designed for their policy.

PR #7928, opened 2018-07-02
Fix 'Unknown addon provider in scalingo.json'

u1-liquid:
Deploy button is broken now. this fixes 'Unknown addon provider' error

### DIFF
--- a/scalingo.json
+++ b/scalingo.json
@@ -97,7 +97,7 @@
     "postdeploy": "bundle exec rails db:migrate && bundle exec rails db:seed"
   },
   "addons": [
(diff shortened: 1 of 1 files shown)


ykzts:
cc @johnsudaar

Soulou:
Hi there Scalingo's CTO here, apologies for this regression. This patch would definitely fix the deployment, we're also going to fix this regression to have the old version working too. That being said, I would encourage the merging here.

Gargron:
Great, thanks for clearing it up!

Gargron (approved):

PR #7929, opened 2018-07-02
Add more granular OAuth scopes

### LINKED ISSUE #7121: Proposal: Finer-grained oauth/api permissions
At present, the scopes for oAuth applications seem to be too high level to properly control account security. Current scopes are:

- read
- write
- follow

Issues from the way current scopes are setup:
- No way to deny an application access to certain information
> I can't authorize and say "no, you can't read my statuses" — this is similar to the security issue with Facebook permissions in regards to Cambridge Analytica, sometimes you just want to "log in as", and not actually give the app permission to see your data.
- Applications requesting follow permissions can read a users blocks and mutes
- Applications with Write permissions can update account settings (the permission request says "post on your behalf" (bug?)
- Applications with Read permissions can read direct messages
- Applications with Read permissions can clear users notifications (bug)

I'd propose breaking the permissions down into more granular scopes, giving users more control over their data.

Likewise, I'd like propose we change how we figure out the scope for the serializers, as to enable per-attribute level scopes. This would mean changing the `serialization_scope` from the default of `current_user` to the `doorkeeper_token`, and introducing a base class for all serializers, similar to the following:

```ruby
class Api::BaseController < ApplicationController
  serialization_scope :doorkeeper_token
end

class REST::BaseSerializer < ActiveModel::Serializer
  def current_user
    @current_user ||= User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
    @current_user = nil
  end

  def current_user?
    !current_user.nil?
  end

  def superapp?
    doorkeeper_token && doorkeeper_token.application.superapp? || false
  end

  def oauth_scope? *required_scopes
    doorkeeper_token && doorkeeper_token.includes_scope? *required_scopes || false
  end
end
```

This would require updating all the existing serializer's to make them inherit from this base, but then would allow things like excluding fields based on scopes. 

For instance, `Api::V1::Accounts::CredentialsController` we could then expose the users' email address if the right scope was requested:

```ruby
class REST::CredentialAccountSerializer < REST::AccountSerializer
  attribute :email, if: -> { oauth_scope? :read_email }

# .. snip...
end
```

Or on statuses:

```
class REST::StatusSerializer < ActiveModel::Serializer
  # .. snip ..

  attribute :favourited, if: -> { current_user? && oauth_scope? :read_favourites }
  attribute :reblogged, if: -> { current_user? && oauth_scope? :read_boosts }
  attribute :muted, if: -> { current_user? && oauth_scope? :read_mutes }

  # .. snip ..
end
```

At the end of the day, the current permissions

* * * *

- [x] Related issues:
  - https://github.com/tootsuite/mastodon/issues/6901

Gargron:
Fix #7121

* write
  * write:accounts
    * `PUT /api/v1/accounts/verify_credentials`
    * `POST /api/v1/statuses/:id/pin`
    * `POST /api/v1/statuses/:id/unpin`
  * write:blocks
    * `POST /api/v1/accounts/:id/block`
    * `POST /api/v1/accounts/:id/unblock`
    * `POST|DELETE /api/v1/domain_blocks`
  * write:favourites
    * `POST /api/v1/statuses/:id/favourite`
    * `POST /api/v1/statuses/:id/unfavourite`
  * write:filters
    * `POST /api/v1/filters`
    * `PUT|DELETE /api/v1/filters/:id`
  * write:follows
    * `POST /api/v1/accounts/:id/follow`
    * `POST /api/v1/accounts/:id/unfollow`
    * `POST /api/v1/follows`
    * `POST /api/v1/follow_requests/:id/authorize`
    * `POST /api/v1/follow_requests/:id/reject`
  * write:lists
    * `POST|DELETE /api/v1/lists/:id/accounts`
    * `POST /api/v1/lists`
    * `PUT|DELETE /api/v1/lists/:id`
  * write:media
    * `POST /api/v1/media`
    * `PUT /api/v1/media/:id`
  * write:mutes
    * `POST /api/v1/statuses/:id/mute`
    * `POST /api/v1/statuses/:id/unmute`
    * `POST /api/v1/accounts/:id/mute`
    * `POST /api/v1/accounts/:id/unmute`
  * write:notifications
    * `POST /api/v1/notifications/clear`
    * `POST /api/v1/notifications/:id/dismiss`
  * write:reports
    * `POST /api/v1/reports`
  * write:statuses
    * `POST /api/v1/statuses/:id/reblog`
    * `POST /api/v1/statuses/:id/unreblog`
    * `POST /api/v1/statuses`
    * `DELETE /api/v1/statuses/:id`
* read
  * read:accounts
    * `GET /api/v1/accounts/verify_credentials`
    * `GET /api/v1/accounts/:id/followers`
    * `GET /api/v1/accounts/:id/following`
    * `GET /api/v1/accounts/search`
    * `GET /api/v1/statuses/:id/favourited_by`
    * `GET /api/v1/statuses/:id/reblogged_by`
    * `GET /api/v1/accounts/:id`
  * read:blocks
    * `GET /api/v1/blocks`
    * `GET /api/v1/domain_blocks`
  * read:favourites
    * `GET /api/v1/favourites`
  * read:filters
    * `GET /api/v1/filters`
    * `GET /api/v1/filters/:id`
  * read:follows
    * `GET /api/v1/accounts/relationships`
    * `GET /api/v1/follow_requests`
  * read:lists
    * `GET /api/v1/accounts/:id/lists`
    * `GET /api/v1/lists/:id/accounts`
    * `GET /api/v1/lists`
    * `GET /api/v1/lists/:id`
  * read:mutes
    * `GET /api/v1/mutes`
  * read:notifications
    * `GET /api/v1/notifications`
    * `GET /api/v1/notifications/:id`
  * read:reports
    * `GET /api/v1/reports`
  * read:search
    * `GET /api/v1/search`
    * `GET /api/v2/search`
  * read:statuses
    * `GET /api/v1/accounts/:id/statuses`
    * `GET /api/v1/timelines/direct`
    * `GET /api/v1/timelines/home`
    * `GET /api/v1/timelines/list/:id`
    * `GET /api/v1/statuses/:id`
    * `GET /api/v1/statuses/:id/context`
    * `GET /api/v1/statuses/:id/card`
* follow (legacy)
  * read:blocks
  * read:follows
  * read:mutes
  * write:blocks
  * write:follows
  * write:mutes
* push

All UIs remain the same before, except the developer form for creating new apps, which I have adjusted:

image


ThisIsMissEm:
I agree on the scopes, but not the design. We should group scopes by “privacy” level. Read public profile, Read Public toots, Etc. it should be grouped by what people do. 

If I use an app, yes, I want to know they can do each thing, but I also what that high level overview.

A list like this isn’t useful to non technical users. We must make that simple for people to understand. What does each thing mean? Maybe displau it grouped by topic? 

```
Favourites:   (read y/n) (write y/n)
Toots:           (read y/n) (write y/n)
```

ykzts (approved):

ThisIsMissEm:
That is, it should be grouped by type, not access level; they’re different vertices.

ThisIsMissEm:
So, basically, you need to transpose this into a grid: field / read / write

For the top levels, they should just auto-check all below. Check all / read, abd it checks everything below it in read

ThisIsMissEm:
The current design is too developer focused; from a user perspective I want to know who and what can do what things.

Gargron:
@ThisIsMissEm The screenshot is from the development "create app" UI. The authorize page is untouched, and simply lists the human description of the requested scopes as before.

Gargron:
I'm hoping it's an incremental improvement and will allow for more improvements on top of it in the future. Most importantly there is still no way to get different statuses depending on scope, it's all or nothing.

ThisIsMissEm:
@gargron okay, so, I’d take things that are generic (read/write/follow) and put those in a separate section, to discourage their use; I think “push” could also be split up into finer grain items, but it isn’t critical for now.

Then the remainder can be groups by object like shown above, rather than a flat list. Currently the list looks overwhelming and jumbled, but by adding grouping we can make it much clearer for developers & help encourage them to make better choices

Gargron:
Maybe I could assign color codes to each scope, and the top-level ones could be dangerous red or something, additionally to re-grouping them. Personally I don't know if client apps should get the top-level or all of the specific ones. What if a new feature comes out, like filters? Then all apps would have to re-register and re-authenticate their users...

I'm probably not going to add anything more to this PR though, because 37 changed files is on the high end of what I expect other contributors to read and review.

ThisIsMissEm:
I think even for client apps people should know exactly what that application is going to be able to access on their behalf. It's not like we can create an access token that'll only work on the device — nothing stops apps from exfiltrating access tokens to remote servers to scrap data and collect information breaching user privacy.

app/controllers/api/v1/follows_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Api::V1::FollowsController < Api::BaseController
-  before_action -> { doorkeeper_authorize! :follow }
+  before_action -> { doorkeeper_authorize! :follow, :'write:follows', :'write:follows:create' }
```
valerauko: Is the `:'write:follows:create'` intentional? Is there a third tier too?

app/controllers/api/v1/lists_controller.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class Api::V1::ListsController < Api::BaseController
-  before_action -> { doorkeeper_authorize! :read },    only: [:index, :show]
-  before_action -> { doorkeeper_authorize! :write }, except: [:index, :show]
+  before_action -> { doorkeeper_authorize! :read, :'read:lists' },    only:  [:index, :show]
```
valerauko: Previously the colons were aligned too (there are a few others like this)

config/initializers/doorkeeper.rb
```diff
@@ -55,7 +55,32 @@
...
+                  :'write:favourites',
+                  :'write:filters',
+                  :'write:follows',
+                  :'write:lists',
+                  :'write:media',
+                  :'write:mutes',
+                  :'write:notifications',
+                  :'write:reports',
+                  :'write:statuses',
+                  :read,
```
valerauko: Does this need to be added to `optional_scopes` too?

spec/controllers/api/v1/domain_blocks_controller_spec.rb
```diff
@@ -4,14 +4,24 @@
...
   let(:user)  { Fabricate(:user, account: Fabricate(:account, username: 'alice')) }
-  let(:token) { Fabricate(:accessible_access_token, resource_owner_id: user.id, scopes: 'follow') }
+  let(:token) { Fabricate(:accessible_access_token, resource_owner_id: user.id, scopes: scopes) }
 
   before do
     user.account.block_domain!('example.com')
     allow(controller).to receive(:doorkeeper_token) { token }
   end
 
+  shared_examples 'forbidden for wrong scope' do |wrong_scope|
```
valerauko: I'd say this should be moved to some shared place once other controller specs use it too (blocks and accounts already have the same though)

PR #7932, opened 2018-07-02
Disable service worker media cache and IndexedDB storage

### LINKED ISSUE #7615: Having a high-traffic feed open causes enormous disk I/O in Firefox

### LINKED ISSUE #7893: PWA uses too much space in web browser
Hi

I just saw that Mastodon uses **769MB** of data in my profile, I think this is due to the PWA. This is way too much, and I don't need images or videos to be saved locally at all.
Is there anything that can be done?

Thanks!

* * * *

Gargron:
Fix #7893
Fix #7615

### DIFF
--- a/app/javascript/mastodon/actions/importer/index.js
+++ b/app/javascript/mastodon/actions/importer/index.js
@@ -1,5 +1,5 @@
-import { autoPlayGif } from '../../initial_state';
-import { putAccounts, putStatuses } from '../../storage/modifier';
+// import { autoPlayGif } from '../../initial_state';
+// import { putAccounts, putStatuses } from '../../storage/modifier';
 import { normalizeAccount, normalizeStatus } from './normalizer';
 
 export const ACCOUNT_IMPORT = 'ACCOUNT_IMPORT';
@@ -45,7 +45,7 @@ export function importFetchedAccounts(accounts) {
   }
 
   accounts.forEach(processAccount);
-  putAccounts(normalAccounts, !autoPlayGif);
+  //putAccounts(normalAccounts, !autoPlayGif);
 
   return importAccounts(normalAccounts);
 }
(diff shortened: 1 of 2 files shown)


Gargron:
Is this correct, @akihikodaki? I am sorry that I have to undo your work, there's just more downsides than upsides from this PWA stuff.

akihikodaki:
Yes, it seems fine. What about just disabling Web Cache (without disabling IndexedDB?), by the way? If you have reasons to disable IndexedDB, I would like you to note them here so we can fix revisit later.

Gargron:
IndexedDB seems unused for anything practical, since timeline contents are not saved, once you are offline, you can't really look at anything useful, unless you happen to have the profile or individual status page open. From one of the linked issues, it sounds like it's cause of high disk I/O, so I figured it's safer to disable it.

PR #7934, opened 2018-07-03
Allow the media modal to be closed by pressing back

MaciekBaron:
The purpose of this change is to be able to close the media modal using the back button (popular behaviour on Android phones).


Gargron (approved):

Gargron:
There is an issue when the modal is used outside of the /web interface, e.g. on a public page.

     TypeError: "this.context.router is undefined"
     TypeError: "this.unlistenHistory is not a function"

MaciekBaron:
I'll try fixing that today, will add a check if the router is injected
properly.

Issue #7642, opened 2018-05-27
Mastodon 2.4 stopped supporting Ruby 2.3


ProgVal:
Hi,

Mastodon 2.4's release notes state the following:

> If you are on Ruby 2.3.x and lower you will get errors when uploading images. Please upgrade to Ruby 2.5.1 or at least 2.4.x

As Ruby 2.3.0 is only two years old and is still the version in Debian 9 (stable) and Ubuntu 17.10 (supported until july 2018), could you consider restoring its support?

Thanks!

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Exagone313:
I should also add that the Gemfile contains `ruby '>= 2.3.0', '< 2.6.0'` and everything except uploading seems to work.

PR #7935, opened 2018-07-03
Restore support to ruby 2.3, add ruby 2.3 to circle ci

renatolond:
This replace calls of String#match? with rails Regex#match?
This follows the same idea used to keep Rails 5.2 compatible with Ruby 2.2.2 in https://github.com/rails/rails/pull/32973

Solves #7642

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,13 @@ jobs:
         environment: *ruby_environment
     <<: *install_ruby_dependencies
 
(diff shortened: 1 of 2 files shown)


Gargron:
I don't think adding more build time to all PRs is a good idea though. I think only 2 jobs can run at the same time with our CircleCI plan.

Gargron (approved):

PR #7936, opened 2018-07-03
Add white outline to black emojis (fix #5083)

### LINKED ISSUE #5083: Add a light background behind "dark text" emojis
As noted in #2475, Mastodon's EmojiOne 2 seems to assume a light background, making a bunch of emojis unreadable. This not only applies to nonsense emoji representations for ©️ ®️ ™️, but also messes up, among others, 🔜. Since we now have full control of what emoji images we are using, it makes sense to go ahead and put a light background behind those troublesome images.

Glyphs to edit: 🎵🎶➕➖➗✖️💲💱™️©️®️〰️➰🔚🔙🔛🔝🔜✔️☑️⚫🔳▪️◾◼️⬛♠️♣️🏴🎱🎮🌑🖋️✒️💣📸📷. Inherently black things (geometric and playing cards stuff) can be made as bright as 👤. Clock symbols and silhouettes appear OK.

(For some reason I believe 👁️‍🗨️ is not working properly.)
See also EmojiTwo/emojitwo#110.

* * * *

- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases), 1.6.1.

MaciekBaron:
This change adds a white outline to black emojis so they are visible while using the default theme. It also reduces their size slightly so that the outline does not leak out outside the container.

The list of black emojis can be easily updated by appending a string to a SASS variable.

image

### DIFF
--- a/app/javascript/styles/application.scss
+++ b/app/javascript/styles/application.scss
@@ -22,3 +22,4 @@
 @import 'mastodon/tables';
 @import 'mastodon/admin';
 @import 'mastodon/rtl';
+@import 'mastodon/accessibility';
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron:
Unfortunately, it doesn't seem to look any good on Firefox (deployed on mastodon.social): 

grafik

I would suggest a revert because this is visually worse than them just not showing up on a dark theme (fwiw discord and tweetdeck have the same problem and aren't doing anything to correct it either)

MaciekBaron:
I personally think it's better to have it look better on most browsers, than have it hardly visible on all browsers, but if the general consensus is to remove the outline, I'm happy with that.

trwnh:
Wouldn't it be better to apply a filter like `invert(100%)` instead of a white outline? As long as the icons are greyscale, this shouldn't mess up any colors. If it does, then you can apply a filter `hue-rotate(180deg)` as well.

PR #7952, opened 2018-07-05
Adjusts colors in the contrast theme, for better contrast

mareklach:


lynlynlynx (changes requested):
- Do NOT change fixed dependent colors, `classic-*-color`s. Changing any colors must be done by replacing non dependent colors, `$ui-*-color`s etc..
- Please teach me a reason why you choose these new colors. This is because these colors should be derived from mastodon's default color scheme.
- I test this pull request on my environment, and the appearance seems okay, but I cannot find out how the contrast issues are improved. So, please show me the improvement of contrast ratio by numerical value and the before and after images.

nightpool:
@lynlynlynx can you explain what you mean by "fixed dependent colors"? The difference between classic-* and ui-* and the restrictions on each are not at all obvious to me from the code in question.

lynlynlynx:
> can you explain what you mean by "fixed dependent colors"? The difference between classic-* and ui-* and the restrictions on each are not at all obvious to me from the code in question

From the viewpoint of codes, `classic-*-color`s do not have `!default` flags(= colors are fixed) in default mastodon theme, so their colors cannot be overwritten in customized themes.

From the viewpoint of backward compatibility, `classic-*-color`s have their origin in old(classic) colors used before [#2912's change for improving scss color variables](https://github.com/tootsuite/mastodon/pull/2912/files#diff-758519e06ffd726dc79fb8a4fdf522e8), so their colors should be fixed for backward compatibility for the former(non-multiple) custom theme users.

@nightpool

PR #7975, opened 2018-07-07
Add option to not consider word boundaries when processing keyword filtering

ClearlyClaire:


app/lib/feed_manager.rb
```diff
@@ -200,7 +200,16 @@ def phrase_filtered?(status, receiver_id, context)
     active_filters = Rails.cache.fetch("filters:#{receiver_id}") { CustomFilter.where(account_id: receiver_id).active_irreversible.to_a }.to_a
 
     active_filters.select! { |filter| filter.context.include?(context.to_s) && !filter.expired? }
-    active_filters.map! { |filter| Regexp.new("\\b#{Regexp.escape(filter.phrase)}\\b", true) }
+    active_filters.map! do |filter|
+      if filter.whole_word
+        sb = filter.phrase =~ /\A[[:word:]]/ ? '\b' : ''
```
Gargron: Whole word, but not if the phrase begins with a non-word character. Is this actually intuitive?

ThibG: If one end of the phrase is not part of a word, it doesn't create a word boundary.
E.g., if phrase is `" bar"`, the regexp `\b\ bar\b` won't match `“ bar”` unless a word precedes it, as in `“foo bar”`. I think it's most intuitive this way.

The downside is that a phrase like `"|foo"` will also match `"!|foo"`. Not sure it's much less intuitive though.

Gargron: One solution (and probably what I should be doing anyway) is strip the phrase so it doesn't have any whitespace around it. However people might want to filter @twitter.com and since that appears as something@twitter.com, the logic might actually work well for that case.

ThibG: `something@twitter.com` would work, but not `something @twitter.com` or `@twitter.com` and I think that would be confusing.

Gargron (approved):

app/lib/feed_manager.rb
```diff
@@ -200,7 +200,16 @@ def phrase_filtered?(status, receiver_id, context)
...
+        eb = filter.phrase =~ /[[:word:]]\Z/ ? '\b' : ''
```
tateisu: you should use \z instead of \Z

nightpool: why?

Gargron: Multiline matches, i think.

PR #7984, opened 2018-07-09
Music Player in the sidebar

lfuelling:
This adds a configurable music player to the "Getting Started" sidebar. Because the audio element is separated from the controls, the music doesn't stop when changing views or when leaving the browser on mobile.

Instance admins can change the visibility, name and the stream URL in the admin settings. By default, the Mastodon notification sound is used and the player is not visible.

It looks like this:

screenshot 2018-07-09 at 17 39 36

The admin interface looks like this (is using i18n):

screenshot 2018-07-09 at 11 57 56


MaciekBaron:
Thanks for your change lfuelling! If you can resolve the merge conflicts we can review it and merge it :)

lfuelling:
Done :)

lfuelling:
Oh great, new conflicts while waiting for review.

Gargron:
Hm, right. I'm sorry for not responding sooner but I... didn't know what to say. Your patch is *cool* and you've done a good job but this should really not be in Mastodon. There's no reason for it to be in Mastodon. It's not a music streaming app, admins don't run radio stations and people don't sign up to listen to radio.

Perhaps this is something that your community needs, and this PR is valuable as a way to cherry-pick those changes to people who want them. If we had a plugin system, this would be a plugin, unfortunately, plugin systems are not easy to design, especially with all the React and Webpack stuff going on.

Again, sorry for not responding sooner.

lfuelling:
Then it would be nice to just say that instead of leaving me in uncertainty. Because I had the same reason in my head this entire month.

Gargron:
I'm sorry, it's a real fault of mine, I have trouble saying "no". :frowning:

PR #7998, opened 2018-07-11
Add federation relay support

### LINKED ISSUE #7399: Federation relays
I'm not saying that we're planning to implement this, but it's an interesting idea worth a discussion. One glaring hole of Mastodon UX is bootstrapping a new server with content so it doesn't feel empty.

See: https://github.com/jaywink/social-relay

>Application to act as a relay for public posts using the Diaspora protocol. Keeps track of nodes and their subscription preferences, receives payloads and forwards the payloads to subscribers. The aim is to pass public posts around in an efficient way so any new node in the network can quickly subscribe to lots of public activity, without having to wait a long time to create social relationships.

A way to fix #1589 

The interesting, and perhaps negative aspect of this is that the implementation of such a relay server would be completely orthogonal/independent of ActivityPub. As far as the relay is concerned, it's just passing strings around and managing subscriptions. It's negative in my opinion because there's no standard to fall back on.

Gargron:
Fix #7399

**What?** Federation relays are intermediaries between Mastodon servers. A Mastodon server can subscribe to a relay and publish all public posts to it, receiving all public posts it receives from others in turn.

**Why?** New & smaller servers struggle with content. New users who sign up on such servers find themselves in a desolate place and don't stick around. It takes users following other users to fill up the federated timeline... Or so it was before federation relays. **Removes the need for follow-bots**

**Both participation in a relay as well as the choice of a relay are _completely optional_.**

image

TODO:

- [x] Backend
- [x] Admin settings

___

The source code for the relay server is: https://source.joinmastodon.org/mastodon/pub-relay

### DIFF
--- a/app/models/relay.rb
+++ b/app/models/relay.rb
@@ -12,9 +12,13 @@
 #
 
 class Relay < ApplicationRecord
-  validates :inbox_url, presence: true, if: :will_save_change_to_inbox_url?
+  PRESET_RELAY = 'https://relay.joinmastodon.org/inbox'
+
+  validates :inbox_url, presence: true, uniqueness: true, url: true, if: :will_save_change_to_inbox_url?
 
   scope :enabled, -> { where(enabled: true) }
+
+  before_destroy :ensure_disabled
 
   def enable!
     activity_id = ActivityPub::TagManager.instance.generate_uri_for(nil)
@@ -58,5 +62,10 @@ class Relay < ApplicationRecord
 
   def some_local_account
     @some_local_account ||= Account.local.find_by(suspended: false)
+  end
+
+  def ensure_disabled
+    return unless enabled?
+    disable!
   end
 end
--- a/app/models/relay.rb
+++ b/app/models/relay.rb
@@ -43,6 +43,7 @@ class Relay < ApplicationRecord
       '@context': ActivityPub::TagManager::CONTEXT,
       id: activity_id,
       type: 'Follow',
+      actor: ActivityPub::TagManager.instance.uri_for(some_local_account),
       object: ActivityPub::TagManager::COLLECTIONS[:public],
     }
   end
@@ -52,9 +53,11 @@ class Relay < ApplicationRecord
       '@context': ActivityPub::TagManager::CONTEXT,
       id: activity_id,
       type: 'Undo',
+      actor: ActivityPub::TagManager.instance.uri_for(some_local_account),
       object: {
         id: follow_activity_id,
         type: 'Follow',
+        actor: ActivityPub::TagManager.instance.uri_for(some_local_account),
         object: ActivityPub::TagManager::COLLECTIONS[:public],
       },
     }
--- a/app/models/relay.rb
+++ b/app/models/relay.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+# == Schema Information
+#
+# Table name: relays
+#
+#  id                 :bigint(8)        not null, primary key
+#  inbox_url          :string           default(""), not null
+#  enabled            :boolean          default(FALSE), not null
+#  follow_activity_id :string
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#
+
+class Relay < ApplicationRecord
+  validates :inbox_url, presence: true, if: :will_save_change_to_inbox_url?
+
+  scope :enabled, -> { where(enabled: true) }
+
+  def enable!
+    activity_id = ActivityPub::TagManager.instance.generate_uri_for(nil)
+    payload     = Oj.dump(follow_activity(activity_id))
+
+    ActivityPub::DeliveryWorker.perform_async(payload, some_local_account.id, inbox_url)
+    update(enabled: true, follow_activity_id: activity_id)
+  end
+
+  def disable!
+    activity_id = ActivityPub::TagManager.instance.generate_uri_for(nil)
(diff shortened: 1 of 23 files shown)


app/models/relay.rb
```diff
@@ -0,0 +1,62 @@
...
+    payload     = Oj.dump(unfollow_activity(activity_id))
+
+    ActivityPub::DeliveryWorker.perform_async(payload, some_local_account.id, inbox_url)
+    update(enabled: false, follow_activity_id: nil)
+  end
+
+  private
+
+  def follow_activity(activity_id)
+    {
```
puckipedia: the Follow activity does not contain the actor, it seems? It seems the actor is implied at the relay side based on who signed the message... I'd prefer just implicitly mentioning it in the Follow..

nightpool: +1

nightpool: I assume you meant "explicitly" in the last sentence?

aveao:
@nightpool I just realized that this is probably the wrong place to write this on (issue would be better), so I'll probably take it there, but I think that you missed the point of my message.

Gargron:
@aveao

- Using the relay is optional. If you want to keep the status quo of organic connections, you can. If the relay goes down, it's not catastrophic, you're just back at square one.
- You can choose any relay you want. Yes it's a centralized point, but that's the only way it can fulfill it's purpose. Nobody says the relays couldn't use p2p tech between each other, potentially...
- Although I am not decided on how I will design the admin UI for this yet, the code itself is built to support working with multiple relays at the same time, so you could have redundancy.

app/models/relay.rb
akihikodaki: `some_local_account` is also confusing. It is already present in `ReportService`, but we are going to have two activities (`Follow` for this case and `Flag` for `ReportService`). I think it is a good time to have account-independent `Service` if Mastodon is going to have more and more such activities.

akihikodaki:
You say:
```Ruby
Relay.create!(inbox_url: 'https://relay.joinmastodon.org/inbox').enable!
```
But is it a temporary means, or you are planning to keep doing that even though you may wrap it with Web UI? It is very strange that it activates relay while `Follow` activity is actually performed. `Follow` is for subscription, not for publishing.

There are two options to solve the problem:
1. Just use `toot:Relay` instead of `Follow`
2. Divide `relays` into `relay_subscribers` and `relay_publishers`, and perform `Accept` for `relay_subscribers` and `Follow` for `relay_publishers`. 

I'm for 2 because it is closer to the semantics of follow-bots. I accept them because it is not harmful for us, but when it comes the current implementation of relay, accepting a relay also means I send `Follow` activity to the instance, subscribing the statuses on the instance. I have to consider various things in this case: what kind of toots does it have? Is it reliable? Isn't it going to be a spamming instance?
The easiest solution for me would be accepting follow-bots and rejecting relays. I suspect it is also true for other people [not confirmed].

nightpool:
@akihikodaki i don't understand your suggestion. Are you saying relays should follow us as well as us following them?

nightpool:
I don't think most admins care about whether they're sending follows or accepts or whatever, this is purely a federation question.

akihikodaki:
@nightpool
I realized https://github.com/tootsuite/mastodon/pull/7998#issuecomment-404373165 is too long and hard to understand while reading :/ In short:
The current relay implementation is always bi-directional. But it is:
* a technically wrong use of `Follow` activity
* not an exact alternative of follow-bots, which is one-directioned

akihikodaki:
@nightpool

> We're delivering to a subset here, but I think it's still within the spirit of the MAY

It is not `sharedInbox`. It is `inbox`.

Even if it applies to this case, it is still MAY so we cannot assume that happens. For example, if someone want to relay with your instance and sends `Follow` activity, you have to reply it with `Accept` or `Reject`. But how do you determine that? You may know it is bi-directional but you should not assume that. If we want this kind of assumption, a new type like `toot:Relay` is more appropriate, as I suggested.

Gargron:
You can send to the relay without subscribing. I don't see anything that would prevent that. It's completely up to the implementation of the relay server how to handle that, and it's entirely separate from the Follow activity. Follow activity is only for subscribing. And indeed in my code the two are united, at the same time as you subscribe, we "enable" it on our end to send stuff to it too.

Gargron:
The relay only has one inbox, so inbox/sharedInbox doesn't make a difference there. The relay code will discard any message that does not contain the public collection in its primary audience, so accidentally sending a DM to the relay actor should not be problematic.

The relay code on its side uses some_local_account's inbox instead of sharedInbox, and i understand that could be a problem with non-Mastodon subscribers, but the relay code is very immature and provisional in general right now so I'll just fix that, it's not a big problem.

I plan to replace some_local_account with a system user in the future but see no reason to bind that to this PR.

adbelle (approved):
Nice.

ykzts (approved):

akihikodaki:
> And indeed in my code the two are united, at the same time as you subscribe, we "enable" it on our end to send stuff to it too.

I am worrying the case when `Follow` activity is rejected. If rejection causes disabling publishing, it is an unexpected side effect. If rejection does not disable publishing, the state of subscription and publishing will be inconsistent.

annando:
Is there some documentation how to get content from the relay and how to get the relay to distribute your content? Although Friendica doesn't support AP right now, I would like to use the relay when this is done.

PR #8010, opened 2018-07-13
Disable language detection for texts shorter than 140 characters

Gargron:
It's wildly inaccurate (and confident of it!) for short sentences:

image

### DIFF
--- a/app/lib/language_detector.rb
+++ b/app/lib/language_detector.rb
@@ -3,12 +3,16 @@
 class LanguageDetector
   include Singleton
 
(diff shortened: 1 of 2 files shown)


ykzts (commented):
nice!

ykzts (approved):

Kjwon15:
I agree for CLD3 is inaccurate in latin character set.
But many of Asian language that are written in non-latin characters are use completely different character set. And CLD3 is detecting it accurately (As seen below)
image

How about enable CLD3 when post contains non-latin characters?

PR #8015, opened 2018-07-13
Delay link crawling for previews

### LINKED ISSUE #4486: Mastodon can be used as a DDOS tool
Hi !

Today I found this tweet : https://twitter.com/mattiasgeniar/status/892446659245993984

I tried to post a link on my instance, Mastodon.cloud, and follow the link's web server logs -> 

400 instant requests. 

Imagine I flood 10 link, I think that go to generate more than 4k requests.. 
It's not great for link's web server..

Any idea how to mitigate this on futur releases ? 
Why Mastodon need to crawl the link ?

Thanks 
* * * *

Gargron:
Fix #4486

TODO:

- [x] Do not crawl links when receiving remote status
- [x] Crawl links when requested by end-user

### DIFF
--- a/app/controllers/api/v1/statuses_controller.rb
+++ b/app/controllers/api/v1/statuses_controller.rb
@@ -35,7 +35,8 @@ class Api::V1::StatusesController < Api::BaseController
   end
 
   def card
-    @card = @status.preview_cards.first
+    @card   = @status.preview_cards.first
+    @card ||= FetchLinkCardService.new.call(@status)
 
     if @card.nil?
       render_empty
--- a/app/lib/activitypub/activity.rb
+++ b/app/lib/activitypub/activity.rb
@@ -73,8 +73,6 @@ class ActivityPub::Activity
   end
 
   def distribute(status)
-    crawl_links(status)
-
     notify_about_reblog(status) if reblog_of_local_account?(status)
     notify_about_mentions(status)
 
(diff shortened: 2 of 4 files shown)


Gargron:
@akihikodaki I finally implemented this, however the on-demand crawl has an unfortunate worst-case scenario (valid link with no valid preview) which may be used to reduce performance.

nightpool:
i'm tentatively pro, but we do exchange one vulnerability for another here—by reducing the likelihood of a DOS attack, we increase the ability for remote users to collect data on when and how many mastodon users are seeing their statuses (especially private and direct statuses)

Also, tying up web workers to resolve foreign links is potentially pretty bad for performance.....

Gargron:
Then maybe I'll just do a perform_async with a random delay within a minute or something like that...

ThibG:
Not how many mastodon users, but how many mastodon instances.
We could use a mix of both approaches (spread out fetching in I don't know, about an hour, but still fetch on demand), or do things differently for private/direct toots.

joellimberg:
Some thoughts (background: I run a single-user instance just for myself, and I'm not using it too actively):

crawling on demand would theoretically allow for the host of the linked content to track when I'm actively using Mastodon, but I think it would be better, resource usage wise, to not fetch data I will probably never see.

Another privacy-preserving option would be to crawl links on-demand, **but** also pre-crawl a small random % of all links. This means that the link host cannot know if I was actually active, or it was a random crawl.

PR #7448, opened 2018-05-11
Update uws dependency to 10.148.0

ignisf:
This release fixes issues with open events, adds Node 10 support
and drops Node 4, 5, 6 and 7. This fixes running the streaming
server on Arch and other distros that ship node 10.

Changes: https://github.com/elementengineering/uWebSockets-bindings/compare/6152fb6787704a61584f920eb27ff5441d3c8c8f...689c15c59e5065b22415266f398ed29118dbe562

PR #8019, opened 2018-07-14
Upgrade Node.js to v8.x on Docker image

ykzts:
Node.js v6 is no longer supported in #7448. Therefore, streaming does not run in Docker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "tiny-queue": "^0.2.1",
     "uglifyjs-webpack-plugin": "^1.2.7",
     "uuid": "^3.1.0",
-    "uws": "^10.148.1",
+    "uws": "10.148.0",
     "webpack": "^4.16.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -116,7 +116,7 @@
     "tiny-queue": "^0.2.1",
     "uglifyjs-webpack-plugin": "^1.2.7",
     "uuid": "^3.1.0",
-    "uws": "^10.148.1",
+    "uws": "10.148.0",
```
ykzts: uws v10.148.1 does not support Alpine.

ykzts: The source code of uws v10.148.1 is not published on GitHub．

https://github.com/uNetworking/uWebSockets-bindings/blob/95e44dba351ed8e6d91aa559d4bb9bb539e1c681/nodejs/dist/package.json#L3

🤔

Gargron (approved):

PR #8021, opened 2018-07-14
Minor fix regarding domain blocks

ClearlyClaire:
~~~Minor refactoring of the backend for consistency, and a~~~ small fix avoiding a spurious error in the front-end.

### DIFF
--- a/app/javascript/mastodon/actions/domain_blocks.js
+++ b/app/javascript/mastodon/actions/domain_blocks.js
@@ -128,7 +128,7 @@ export function expandDomainBlocks() {
   return (dispatch, getState) => {
     const url = getState().getIn(['domain_lists', 'blocks', 'next']);
 
-    if (url === null) {
+    if (!url) {
       return;
     }
 


Gargron (approved):

ThibG:
Uh, I thought the backend change was sound, but apparently it doesn't work? I don't understand why, though.

Gargron:
The destroy route generated for "resources" is based on ID. That's why I used "resource" instead.

ThibG:
Ok, I think I got it now!

Gargron:
The URL *is* different and therefore this is a breaking API change FYI.

Instead of `DELETE /api/v1/domain_blocks` it's now `DELETE /api/v1/domain_blocks/:domain`

ThibG:
And the tests don't catch that?
Anyway, nevermind, I'll just drop the server-side changes

Le 14 juillet 2018 23:15:34 GMT+02:00, Eugen Rochko <notifications@github.com> a écrit :
>The URL *is* different and therefore this is a breaking API change FYI.
>
>Instead of `DELETE /api/v1/domain_blocks` it's now `DELETE
>/api/v1/domain_blocks/:domain`
>
>-- 
>https://github.com/tootsuite/mastodon/pull/8021#issuecomment-405050350

-- 

Gargron:
The tests don't test the URL string, they test the controller method invocation. In-path params and query string / form / body params are all inside the same variable in there.

PR #8039, opened 2018-07-16
Fix ActivityPub::UpdateDistributionWorker regression

Gargron:
Regression from #7998 let to profile updates not sending


smorimoto (approved):

ykzts (approved):
oops...

PR #8041, opened 2018-07-17
Update supported Node.js version

abcang:
Since the version of Node.js necessary for uws has increased, Mastodon also updates the necessary version. Also, by updating the version, `whatwg-url` is no longer needed.

Related: 
https://github.com/tootsuite/mastodon/pull/7448
https://github.com/tootsuite/mastodon/pull/7664


ykzts (approved):

ProgVal:
Is that update to uws required? This breaks supports of Nodejs 6, [which is used by some instances](https://github.com/tootsuite/mastodon/pull/7448#issuecomment-388500625)

PR #8073, opened 2018-07-25
Prevent scrolling main frame when navigating in image gallery with arrow keys

ClearlyClaire:
The image gallery modal allows to use the left and right keys to navigate several media.
However, this can also lead to horizontal scroll.

This PR changes that behavior by handling keydown events rather than keyup events, and preventing standard event handling/propagation in the case of left and right arrows.

I am not sure what changing from keyup to keydown may entail.

### DIFF
--- a/app/javascript/mastodon/features/ui/components/media_modal.js
+++ b/app/javascript/mastodon/features/ui/components/media_modal.js
@@ -54,19 +54,23 @@ export default class MediaModal extends ImmutablePureComponent {
     this.setState({ index: index % this.props.media.size });
   }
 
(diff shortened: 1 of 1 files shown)


nightpool:
A related bug I've noticed is that when you press the "back" button in a image frame on android chrome, the column *underneath* the image frame will navigate, and the image frame will remain up. Do you think you could address this issue as well?

Gargron (approved):

ThibG:
@nightpool hasn't this been fixed by #7934 already?

nightpool:
Ah, possibly, my bad.

PR #8076, opened 2018-07-26
Call address instead of exchange

abcang:
Since `Resolv::DNS::Resource::IN::A` has no exchange method, we call the address method instead.


nightpool:
can we add tests for this? is the blacklist tested in general?

Gargron (approved):

abcang:
It seems difficult to mock a dns request, but it seems to be possible to test by mocking `Resolv::DNS#getresources`. But will it be a meaningful test?

PR #8083, opened 2018-07-27
Resize images by area instead of fixed dimensions

Gargron:
To improve the way super tall or super ride images are treated, the
numbers remain the same, 1280x1280 and 400x400, but if an image
is less in one dimension than the other, the other can become larger

Thanks to @WAHa_06x36@mastodon.social for the tip


ykzts (approved):

ykzts:
:heart:

PR #8090, opened 2018-07-28
Fix incorrect context definition for the "featured" keyword

### LINKED ISSUE #8077: http://joinmastodon.org/ns#featured is improper type
currently, `http://joinmastodon.org/ns#featured` is defined in the context as `toot:featured`. However, this means that the featured thing isn't an `@id`, but an `http://www.w3.org/2001/XMLSchema#string`. The context should contain:
`"featured": {"@id": "toot:featured", "@type": "@id"}` instead.

* * * *

Gargron:
Fix #8077

### DIFF
--- a/app/lib/activitypub/adapter.rb
+++ b/app/lib/activitypub/adapter.rb
@@ -18,7 +18,7 @@ class ActivityPub::Adapter < ActiveModelSerializers::Adapter::Base
         'toot'                      => 'http://joinmastodon.org/ns#',
         'Emoji'                     => 'toot:Emoji',
         'focalPoint'                => { '@container' => '@list', '@id' => 'toot:focalPoint' },
-        'featured'                  => 'toot:featured',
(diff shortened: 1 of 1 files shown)


nightpool (commented):
do we also need to fix this for movedTo?

Gargron:
@nightpool I don't know - maybe - we pretend that movedTo is in the ActivityStreams namespace because it comes from a thread on their issue tracker and it should become an official property at some point.

nightpool:
yeah I'm working on getting sensitive and movedTo integrated but afaik
they're not currently in the context document?

puckipedia:
this should indeed be fixed for anything that points to another AS2 object, since those are an ID, and not a string (I would probably consider anything that is always an IRI, like the ostatus predicates, candidate for changing to `@id`? though this might be more preference on my side, seeing as i'm the only one using actual json-ld for activitypub lol)

PR #8093, opened 2018-07-28
Do not hide hashtag timeline in mobile view, make it full-window

Gargron:


nightpool:
screenshots? I don't have a dev environment open right now

ykzts (approved):

PR #8095, opened 2018-07-29
Do not scroll up when modal is opened

### LINKED ISSUE #8008: Modal regression with scrollable areas
After #7934, opening a modal makes the scrollable area in the not-pinned column jump to top. Once the modal is closed, it jumps back to the old position.

* * * *

ykzts:
Resolve #8008

### DIFF
--- a/app/javascript/mastodon/components/status_list.js
+++ b/app/javascript/mastodon/components/status_list.js
@@ -1,12 +1,12 @@
 import { debounce } from 'lodash';
 import React from 'react';
+import { FormattedMessage } from 'react-intl';
 import ImmutablePropTypes from 'react-immutable-proptypes';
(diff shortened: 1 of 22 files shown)


ThibG:
There are a few more uses of `ScrollContainer`, most notably the detailed status view and the account media timeline.

ykzts:
@ThibG Thank you 👍

The account media timeline does not open modal so it has not been fixed.

ThibG:
Right, however the modal can be opened from another column, and the account gallery will still scroll to top

ykzts:
> Right, however the modal can be opened from another column, and the account gallery will still scroll to top

Ah, that's right.

Gargron (approved):

ykzts:
dry

PR #8099, opened 2018-07-30
Fix no avatars on mobile devices with 'smaller' screens

todoneunl:
Although these 'smaller' screens are just normal configured LG G2's. It doesn't make sense to hide an avatar at all by the way.

### DIFF
--- a/app/javascript/styles/mastodon/containers.scss
+++ b/app/javascript/styles/mastodon/containers.scss
@@ -375,10 +375,6 @@
             border: 0;
             border-radius: 4px;
           }
-
-          @media screen and (max-width: 360px) {
-            display: none;
-          }
         }
       }
 


Gargron:
What do you mean "fix" if it's intentional?

The usernames can be quite long and they push the button out of the screen, it's better to hide the avatar, since the avatar is literally next to every status on the page.

jeroenpraat:
It's called a fix, cause IMO avatars should always be visible on top of a profile. I tested this and the avatar is displayed on the line above the username (so there's a linebreak after the avatar).

schermafdruk van 2018-07-30 18-41-40

Although I think this is even better (follow button on next line centered):

mockupprofile

jeroenpraat:
*one more thing: You say "since the avatar is literally next to every status on the page". 

That's true, but people don't see that before they start swiping. It's important people see the avatar in the first impression.

ThibG:
It's not even true, since the person's last toots may be boosts

jeroenpraat:
Closing this now, cause it's outdated. Although I still wish to see the avatar on top.

PR #8101, opened 2018-07-30
Convert MOV and WEBM to MP4, raise maximum limit to 40MB

### LINKED ISSUE #5533: Video uploads fail with error 422 on at least iOS 11
I didn't run into this issue (in fact I don't even have a modern iOS device!); @_ADifferentKind@mastodon.social did.

From what I can gather, the cause could be either or both of the following: iOS 11 changed default saved video format to H.265, and iOS since at least 2012 saves videos with the .mov extension.

* * * *

Gargron:
Separate size limits for images and video. Images remain at 8MB,
while videos can be up to 40MB.

Fix #5533

### DIFF
--- a/app/models/media_attachment.rb
+++ b/app/models/media_attachment.rb
@@ -25,10 +25,11 @@ class MediaAttachment < ApplicationRecord
   enum type: [:image, :gifv, :video, :unknown]
 
   IMAGE_FILE_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif'].freeze
-  VIDEO_FILE_EXTENSIONS = ['.webm', '.mp4', '.m4v'].freeze
+  VIDEO_FILE_EXTENSIONS = ['.webm', '.mp4', '.m4v', '.mov'].freeze
 
-  IMAGE_MIME_TYPES = ['image/jpeg', 'image/png', 'image/gif'].freeze
-  VIDEO_MIME_TYPES = ['video/webm', 'video/mp4'].freeze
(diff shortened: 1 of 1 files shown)


ykzts (approved):

rinsuki:
I think, dont need re-encode if .mov file contains H.264 stream

Gargron:
@rinsuki I don't think you can just plop a .mov file into `<video>` so re-encoding seems necessary no matter what. Hopefully ffmpeg knows to avoid unnecessary work...

rinsuki:
@Gargron if .mov file with HTML5 video compatible codecs(e.g. H.264 and AAC), only needs `ffmpeg -i in.mov -codec copy out.mp4`. this is only swapping file containers, and this is very fast 🚀

Gargron:
@rinsuki Hmm.... So we'd need to check the metadata before the appropriate processor/style is selected...

rinsuki:
> When you share HEIF or HEVC media using other methods, such as AirDrop, Messages, or email, it's shared in a more compatible format, such as JPEG or H.264.
https://support.apple.com/en-us/HT207022

🤔

renatolond:
I think this might still be limited to 8MB by the default nginx configuration suggested in the production guide: https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Production-guide.md#nginx-configuration

`client_max_body_size 8m;`

JBaczuk:
@redtachyons It was, but it looks like it's been changed to 80m: https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Production-guide.md

PR #8102, opened 2018-07-30
Enforce username format for remote users, too

### LINKED ISSUE #8058: Person masking their identity?
This looks well dodgy.

screen shot 2018-07-22 at 15 01 41

screen shot 2018-07-22 at 15 01 50

screen shot 2018-07-22 at 15 01 52

Clicking their avatar leads to: https://ap.uwu.st/gar.ajson

* * * *

Gargron:
Initially I thought there might be valid reasons for remote users to
have a different, unpredicted username format. However, I now realize
such a difference would be unusable and unexpected within Mastodon.

Fix #8058

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -68,6 +68,7 @@ class Account < ApplicationRecord
 
   # Remote user validations
   validates :username, uniqueness: { scope: :domain, case_sensitive: true }, if: -> { !local? && will_save_change_to_username? }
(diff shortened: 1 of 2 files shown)


ykzts (approved):

nightpool:
I'm really iffy on this one. the idea that no remote users can ever have - or even à in their name seems like a major miscalculation.

Gargron:
@nightpool Our regex won't let people mention them anyway, so that seems okay.

PR #8103, opened 2018-07-30
Use workspace instead of caching for built assets

ClearlyClaire:
Tests cannot run without built assets, and we want
exactly the matching assets. This is not a cache.


Gargron:
I don't know CircleCI enough to understand what difference this makes?

Gargron (approved):

Issue #7908, opened 2018-06-29
Split tables based on column hotness 


Gargron:
Because postgres internally copies the entire row when any column is updated (this is cleaned up during a vacuum), having tables with many columns that update often is more IO and faster table growth.

Things like counter caches should be moved out to separate tables and queried with a join.

Gargron:
Correct me if I'm wrong but partitioning is on the other axis. Partitioning is splitting the table by rows (which e.g. makes indices smaller and look-up time faster), while I am talking about splitting by columns to reduce the amount of data that postgres copies with each update.

Gargron:
One of the top-used queries on mastodon.social is

    UPDATE "accounts" SET "statuses_count" = ?, "updated_at" = ? WHERE "accounts"."id" = ?

So I suppose we should start with the accounts table.

1. Create special table
2. Make code write to both old columns and new table
3. Copy old data to new table
4. Remove old columns (will this lock up the table?)

PR #8104, opened 2018-07-31
Move status counters to separate table, count replies

Gargron:
See #7908 for separate table justification

Also, Rails 5.2 had a breaking change in `cache_key` behaviour, which removed the updated_at timestamp from the key! Therefore, the cache was always stale. `cache_version = false` restores old behaviour we were expecting...

The data from statuses table is copied in a migration using UPSERT mechanics introduced in Postgres 9.5 to avoid conflicts.

Instructions:

1. Run `SKIP_POST_DEPLOYMENT_MIGRATIONS=true rails db:migrate` first
2. Restart Mastodon
3. Run `rails db:migrate`

See also: Showing replies_count in web UI (#8181)

### DIFF
--- a/app/models/favourite.rb
+++ b/app/models/favourite.rb
@@ -32,20 +32,11 @@ class Favourite < ApplicationRecord
   private
 
   def increment_cache_counters
-    if association(:status).loaded?
-      status.update_attribute(:favourites_count, status.favourites_count + 1)
-    else
-      Status.where(id: status_id).update_all('favourites_count = COALESCE(favourites_count, 0) + 1')
-    end
+    status.increment_count!(:favourites_count)
   end
 
   def decrement_cache_counters
     return if association(:status).loaded? && (status.marked_for_destruction? || status.marked_for_mass_destruction?)
-
-    if association(:status).loaded?
-      status.update_attribute(:favourites_count, [status.favourites_count - 1, 0].max)
-    else
-      Status.where(id: status_id).update_all('favourites_count = GREATEST(COALESCE(favourites_count, 0) - 1, 0)')
-    end
+    status.decrement_count!(:favourites_count)
   end
 end
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -26,6 +26,8 @@
 #
 
 class Status < ApplicationRecord
+  self.cache_versioning = false
+
   include Paginable
   include Streamable
   include Cacheable
@@ -59,6 +61,7 @@ class Status < ApplicationRecord
 
   has_one :notification, as: :activity, dependent: :destroy
   has_one :stream_entry, as: :activity, inverse_of: :status
+  has_one :status_stat, inverse_of: :status
 
   validates :uri, uniqueness: true, presence: true, unless: :local?
   validates :text, presence: true, unless: -> { with_media? || reblog? }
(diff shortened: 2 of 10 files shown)


app/models/favourite.rb
```diff
@@ -33,19 +33,19 @@ class Favourite < ApplicationRecord
 
   def increment_cache_counters
     if association(:status).loaded?
-      status.update_attribute(:favourites_count, status.favourites_count + 1)
+      status.increment_count!(:favourites_count)
     else
-      Status.where(id: status_id).update_all('favourites_count = COALESCE(favourites_count, 0) + 1')
+      StatusStat.where(status_id: status_id).update_all('favourites_count = COALESCE(favourites_count, 0) + 1')
```
Gargron: Crap, just realized the record may not exist when this is executed. Hmm.

Gargron:
@unarist @abcang This is a big change (in terms of data, not size of PR), so could you review it? In terms of performance implications and design...

aschmitz:
I don't believe the `CopyStatusStats` migration will *lock* (or block) here, so that's fine as far as it goes. (Postgres' MVCC handling will keep old rows from being vacuumed while it's running, but that's not a big deal, it's not like it should be running for days or anything.)

I'd suggest adding a trigger to update these stats, though, along the lines of https://github.com/tootsuite/mastodon/blob/master/lib/mastodon/migration_helpers.rb#L698 . (Otherwise, any changes between this migration being started and the code being rolled out will be omitted.) Even with that trigger, the ON CONFLICT DO UPDATE will overwrite any values that change *during* the migration, although I'm not sure there's a much better way to do it. (You could take the old value if the overwriting value would be zero, which is probably more right more of the time, but there's no good way without making the migration significantly more complicated. A sketch on how to do that is below if you want it.)

Other than that race, this looks fine. I'm not sure if you really need timestamps on StatusStat, but other than doubling the amount of space used for status timestamps, it's not really hurting anything (and could possibly be useful, since the update timestamp on Status won't catch these changing any longer).

You'll need another migration to backfill reply counts, but I assume you know that.

If you really want to not lose updated counts during that migration, one option would be to:

1. Create the table with nullable counts and a default value of null.
2. Add triggers to update status rows when reblogs_ or favourites_count is updated. (They may need to *create* the row in the first place, but that's doable. Don't forget to set timestamps if you leave them in.)
3. Run essentially the same `INSERT`you have now, but only overwrite columns if they're null in the status_stats row you'd be overwriting. (Otherwise they changed during the migration.)
4. Run `change_column_default(:status_stats, :reblogs_count, 0)`, use MigrationHelpers' `update_column_in_batches(:status_stats, :reblogs_count, 0)`, and then `change_column_null(:status_stats, :reblogs_count, false)` (and the same for favourites_count).

(In theory, you'll want to do the same when filling in the replies_count, but *that* query will probably take a while, and you'll want the trigger to not just increment/decrement the count, but to calculate it correctly each time, because there's no "correct" value to reference.)

Gargron:
@aschmitz Thank you for the assessment! I feel more confident about this update now. As for counter accuracy, I think it is negligible (some super old posts have negative counters due to the first ever migration close to two years ago). For example, remote posts only count local favourites/reblogs anyway, so we're not losing correctness because we don't have any.

db/migrate/20180812162710_create_status_stats.rb
```diff
@@ -0,0 +1,12 @@
+class CreateStatusStats < ActiveRecord::Migration[5.2]
+  def change
+    create_table :status_stats do |t|
+      t.belongs_to :status, foreign_key: { on_delete: :cascade }, index: { unique: true }
```
abcang: I think that it is better to set `null: false`. Also, it seems that foreign_key is not reflected in `db/schema.rb`.

app/models/status.rb
```diff
@@ -26,6 +24,8 @@
 #
 
 class Status < ApplicationRecord
+  self.cache_versioning = false
```
abcang: It may be better to stop changing cache_versioning and stop using cache_key directly.

db/post_migrate/20180813113448_copy_status_stats_cleanup.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CopyStatusStatsCleanup < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    safety_assured do
+      remove_column :statuses, :reblogs_count, :integer, default: 0, null: false
+      remove_column :statuses, :favourites_count, :integer, default: 0, null: false
```
aschmitz: These will cause any existing uses of these columns to throw errors until the code is migrated: I might suggest waiting on this for a later release, so most people won't have old code running when these columns are removed.

aschmitz: Ah, right-o. Didn't realize that had changed. (Or that this was in `post_migrate`.) Carry on, then!

Gargron:
@abcang Fixed

PR #8127, opened 2018-08-05
Remove protocol handler registration

Gargron:
Browser UI too confusing / scares people, protocol URIs not used much

### DIFF
--- a/app/javascript/mastodon/containers/mastodon.js
+++ b/app/javascript/mastodon/containers/mastodon.js
@@ -38,13 +38,6 @@ export default class Mastodon extends React.PureComponent {
       window.setTimeout(() => Notification.requestPermission(), 60 * 1000);
     }
 
-    // Protocol handler
-    // Ask after 5 minutes
-    if (typeof navigator.registerProtocolHandler !== 'undefined') {
-      const handlerUrl = window.location.protocol + '//' + window.location.host + '/intent?uri=%s';
-      window.setTimeout(() => navigator.registerProtocolHandler('web+mastodon', handlerUrl, 'Mastodon'), 5 * 60 * 1000);
-    }
-
     store.dispatch(showOnboardingOnce());
   }
 


ThibG:
It is valuable, especially in a decentralized setting. How would you go about making a “share on Mastodon” button otherwise?

Tagadda:
Can we add a Modal or something to explain that thing ? Maybe in the "tutorial" ?

unarist:
I like this feature, but I also think current UX for this is bad. We should describe how it works before requesting registration or make it opt-in.

Gargron:
>How would you go about making a “share on Mastodon” button otherwise?

I suggest that share button codes be generated by a particular instance. Upon clicking, it should ask for your domain equivalent to the "remote follow" dialog and use the same mechanism to redirect to the user's instance.

>I like this feature, but I also think current UX for this is bad.

I agree that this would have been ideal if the browser UX was acceptable, but it is not. The prompting is annoying and outlandish, there are not enough JS controls for it, users don't know what it does.

>We should describe how it works before requesting registration or make it opt-in.

Given the large number of opt-in features in Mastodon that most people don't know about and don't use, I think it's wasted effort to do that.

nightpool (approved):
do we also need to update any docs?

rugk:
> How would you go about making a “share on Mastodon” button otherwise?

FYI actually there is such a project already, see https://aly-ve.github.io/Mastodon-share-button/.

And as this is inconvenient and actually a downgrade to the web+mastodon protocol, I also think of integrating this into my [browser extension](https://github.com/rugk/mastodon-simplified-federation), so the redirect happens totally automatic. :smile: (just as before with protocol registration) See https://github.com/Aly-ve/Mastodon-share-button/issues/13.
This requires the same code on many websites though, as the add-on needs something "stable" to parse, so:
> I suggest that share button codes be generated by a particular instance

…is something, I would really try to avoid.
At least some meta tag or so _could_ be standardized.

1br0wn:
Has anyone talked to W3C about standardising this? They did write ActivityPub after all :)

quentinsf:
Not having seen this issue, I suggested a protocol handler in a Toot and it got significant numbers of boosts and favourites.
https://mastodon.me.uk/@quentinsf/109539579506128582
I would certainly vote for its reintroduction...

Issue #7752, opened 2018-06-06
Restore normal functionality of CWs in admin interface


bugQ:
A fellow admin complained that the flashy changes made in [this PR](https://github.com/tootsuite/mastodon/pull/7452) made reading a report very difficult due to their sensory processing issues.  In addition, I noticed that this change removed the "show more" button because nobody was able to convince Gargron of the purpose of the button, which he viewed as annoying.  So, here I am to do some convincing.

The purpose of any content warning is threefold:
1. to warn content consumers that the content therein can be distressing to some people (often to the point of triggering anxiety attacks, especially in people with PTSD or other conditions),
2. to let the content consumer decide whether to engage with it (**if such a choice is possible**),
3. having decided to engage with it, to give the content consumer a chance to prepare.

If I had to guess, people who have not had experience with the need for these warnings generally assume only the second point.  Since it's the job of server admins to engage with _all_ user reports, to the uninitiated, it might seem that CWs are "useless".  However, the first and third points are still applicable regardless of the context, and the second can still be relevant in the case of servers with multiple admins, who may delegate work between each other for personal reasons.

I hope this makes it evident why the aforementioned PR was a disaster for many admins of varying psychologies and backgrounds, and I request that CWs once again function identically in all views.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ThibG:
I just want to make clear that #7452 did not remove the “show more” button, it was removed earlier and my initial plan with this PR was to restore that very button. This PR only fixed boosts being misrepresented as original toots and content warnings not standing out at all (although what stuck for that last part was a joke commit eh).

So I agree with this issue since that's what the PR was originally meant to be.

dn3s:
come on though, animated rainbow text? did you just learn a new css trick or something? i know it's kinda dishonest to pretend this project cares about accessibility but still.

dn3s:
@ThibG i was not aware that you weren't expecting that PR to be merged as-is. i wish you would have explained that more clearly as i wouldn't have responded so aggressively, but regardless i apologize.

PR #8129, opened 2018-08-05
Get rid of the Content Warning rainbows

ClearlyClaire:
In my opinion, the “correct” fix would be to simply have a “Show more” button just like everywhere else, as initially proposed in #7452, for clarity, consistency, and for the reasons enumerated in #7752.

In the meantime, this PR fixes the most obvious accessibility issues with this joke commit I did not really expect to get merged.

~~~Kept the animated rainbows otherwise because the content warnings don't really stand out, but we could also remove it entirely.~~~

### DIFF
--- a/app/javascript/styles/mastodon/tables.scss
+++ b/app/javascript/styles/mastodon/tables.scss
@@ -1,9 +1,3 @@
-@keyframes Swag {
-  0% { background-position: 0% 0%; }
-  50% { background-position: 100% 0%; }
-  100% { background-position: 200% 0%; }
-}
-
 .table {
   width: 100%;
   max-width: 100%;
@@ -193,22 +187,7 @@ a.table-action-link {
 
     strong {
       font-weight: 700;
-    }
-  }
-}
-
-.no-reduce-motion .batch-table {
-  .status__content {
-    padding-top: 0;
-
-    strong {
-      font-weight: 700;
-      background: linear-gradient(to right, orange , yellow, green, cyan, blue, violet,orange , yellow, green, cyan, blue, violet);
-      background-size: 200% 100%;
-      -webkit-background-clip: text;
-      background-clip: text;
-      color: transparent;
-      animation: Swag 2s linear 0s infinite;
+      color: $warning-red;
     }
   }
 }
--- a/app/javascript/styles/mastodon/tables.scss
+++ b/app/javascript/styles/mastodon/tables.scss
@@ -193,6 +193,16 @@ a.table-action-link {
 
     strong {
       font-weight: 700;
+    }
+  }
+}
+
+.no-reduce-motion .batch-table {
+  .status__content {
+    padding-top: 0;
+
+    strong {
+      font-weight: 700;
       background: linear-gradient(to right, orange , yellow, green, cyan, blue, violet,orange , yellow, green, cyan, blue, violet);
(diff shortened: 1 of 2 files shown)


nightpool:
I would prefer to remove it entirely. aside from accessibility issues, it
makes reading the actual content warning nearly impossible

ThibG:
@nightpool alright, but it doesn't sufficiently stand out to me without it. Would changing the color to `$error-red` be better?

screenshot_2018-08-05 account statuses - thib - dev instance

ThibG:
Or rather `$warning-red`, this seems more appropriate and has slightly higher contrast.

screenshot_2018-08-05 account statuses - thib - dev instance 1

smonff:
Isn't there a `$warning-color` already defined in the Mastodon UI rules? Just asking.

smonff:
I find it appropriate that the $error-red color has been  changed to the $warning-red. 

Then, in term of contrast between background and text, is this red light enough for being used on a dark background? Plus, if you are color blind, you eventually don't see the redness.

ThibG:
I would think that it is enough. As for colorblindness, you're right, you may not see this additional color hint.

I still think the best way to handle this would not be with color, but by adopting the same behavior than everywhere else, i.e., use a button and hide what's behind the CW, but that was rejected last time I proposed it.

ThibG:
*sigh* I can change it back to the default color, but if this color doesn't work, then we have wider issues, it's used in other places.
You want this text to be different than the toot's content because how would you understand that it's the Content Warning otherwise? (No, the “Content Warning” prefix isn't enough, a regular toot without CW could start this way too)

dn3s:
This is based on what i've heard from others talking about the issue on mastodon:

CWs still have use in an admin context. on instances with multiple admins, one can for example delegate reviewing a particular report to another admin if it's a subject that will be harmful to engage in for that particular admin. Even in the case of single-admin instances, admins can choose to review reports based on their content on their own terms instead of having them pop up in their face with no warning.

Obviously the benefits of this are limited as posts being reported were likely not made in good faith, but it's still helpful.

Remember, people with ptsd, photosensitive epilepsy, or a wide variety of mental health issues that benefit from CWs can make great admins, as they are often better equipped to understand the difficulties marginalized users face. We may as well make their experience a bit better.

Gargron:
Okay, can you try using html details/summary tags for this?

<details>
<summary>Like</summary>
<div>This</div>
</details>

I'm just not fond of duplicating JS over so many places nor do I really like the "show more" button styling to be honest even though I came up with it.

ThibG:
@Gargron I can give it a try, but it is apparently unsupported in Edge, so I'm not completely sure how it would look like there.

dn3s:
i'm a little confused by the scope of this discussion. why are we talking about implementation details of CWs here, when that's already been established elsewhere? why not implement CWs in this context as they are everywhere else, and if people don't like the current implementation, have that as a separate discussion? This introduces an unnecessary blocker to fixing the immediate issue.

Is there some reason I'm missing that makes this specific case an ideal place to start rolling out a new CW implementation, and some reason why rolling out that change is relevant to the issue this PR is attempting to fix? Or is there some reason why the current implementation is appropriate on the timeline interface but inappropriate on the admin interface?

Gargron (approved):

PR #8141, opened 2018-08-08
Update Rails

S-H-GAMELINKS:
Rails 5.2.1 released!

ref: https://weblog.rubyonrails.org/2018/8/7/Rails-5-2-1-has-been-released/


abcang:
Using RAW_POST_DATA seems to be one reason why the test fails.

https://github.com/rails/rails/blob/v5.2.1/actionpack/CHANGELOG.md

> Reset RAW_POST_DATA and CONTENT_LENGTH request environment between test requests in ActionController::TestCase subclasses.

Gargron (approved):

PR #8142, opened 2018-08-08
Update /terms and /about/more to use public layout

Gargron:
Consistent header and footer area between public profiles, public toots, terms page and extended about page.

Notes:

- Removes "connected to x instances" blurb. It misleads people, it's technically correct but not in the way that most people assume.

image

image


trwnh:
What's misleading about "connected to x instances", and how would one access that information if they wanted to know?

nightpool:
it's misleading because mastodon doesn't "connect" to instances, our
federation works on a user-to-user level, with individual subscriptions

I still think it's a good thing to show, even if it gives people the wrong
idea it does convey it's original purpose—each mastodon instance is only
one part of a healthy, functioning network

maybe we can replace it with a count of all known remote accounts?

ykzts (approved):

renatolond:
Arriving a bit late, but could it be replaced by "connected to X users" instead coming from the number of users followed? I don't know if we have that data easily without duplication. (Or even connected to X users in Y instances, again, coming from the number of users followed)

PR #8156, opened 2018-08-09
Update docker-compose to cope with build-time assets

insom:
#7780 means that asset compilation happens as a build step.

Having the assets and packs volumes defined in `docker-compose.yml` breaks this. For people who run under Docker Compose, I believe this will fix their CSS (which even running the asset recompilation separately did not).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,8 +46,6 @@ services:
       - redis
 #      - es
     volumes:
-      - ./public/assets:/mastodon/public/assets
-      - ./public/packs:/mastodon/public/packs
       - ./public/system:/mastodon/public/system
 
   streaming:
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Shadowghost:
Since apache reverse proxies require the assets to be accessible from the host, this PR breaks the reverse proxy. I checked and tested it with the latest master. Any ideas how to solve this without reverting this PR?

Shadowghost:
How to do that exactly when running the docker images behind an apache reverse proxy? Documentation should be updated if this goes live with the next release. I can understand the pressuring, it's annoying :)

nightpool:
can't we just have a volume that shares the built assets?

insom:
I was already passing these requests to Puma, so didn't notice, but I see that the [documentation reverse proxy alternatives](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Alternatives.md) has a line:

     ProxyPassMatch ^/(assets|avatars|emoji|headers|packs|sounds|system|.well-known/acme-challenge) !

which would also need to have `assets` and `packs` removed.

Looks similar in the [Nginx config, too](https://github.com/tootsuite/documentation/blob/f1c0dea8368ec5e97fa2aec5ef00d4b9cdd9b658/Running-Mastodon/Production-guide.md#nginx-configuration)

PR #8157, opened 2018-08-09
Ensure single user is a local user (fixes #8154)

### LINKED ISSUE #8154: Single user mode breaks when new User is created
I decided to change my user on my self hosted Mastodon instance. So I created a new user with the desired username, then unfollowed every on the old account and eventually deleted the old user. This causes my new account not to be the one with the lowest ID in the accounts table anymore. Instead, one of the accounts that i follow is in there, with the lowest ID. This effectively breaks the single user mode, as it now forwards to a non-existing user on my server.

It would be great to have an option in the admin area, or a rake task or something similar to correct this issue. I tried to update the ID in the database, but there are too many foreign key constraints on it that I did not find a way to solve this issue.

If there is already a way to resolve this, please let me know. Thanks.

* * * *

ClearlyClaire:


app/controllers/home_controller.rb
```diff
@@ -58,7 +58,7 @@ def default_redirect_path
     if request.path.start_with?('/web')
       new_user_session_path
     elsif single_user_mode?
-      short_account_path(Account.first)
+      short_account_path(Account.local.first)
```
Gargron: Let's make sure it's not suspended (deleted) either...

Gargron (changes requested):

unarist:
I'm bit surprised that the account is not `Account.find_local(Setting.site_contact_username)` one, btw.

ThibG:
@unarist I guess we could do that, but we would have to make sure `Setting.site_contact_username` points to a valid user and is set appropriately when setting up an instance…

Gargron (approved):

PR #8175, opened 2018-08-12
fix: change Identity's id column to a bigint

### LINKED ISSUE #8775: In `identities` table, foreign key to `users` table is the wrong data type?
I noticed that for the `identities` table, the foreign key to the `users` table is `user_id INTEGER`, while the `id` column in the `users` table is `id BIGSERIAL`...might this cause issues when (if) a user has id > max_integer? I know the chances are pretty slim, but I figured I would open this issue anyway.

image
image

* * * *

aschmitz:
This appears to be the last model created using a 5.0 migration,
where column types defaulted to `integer` rather than `bigint`.

This migration changes the column type to match that of all of the
other ID columns.

Fix #8775

### DIFF
--- a/app/models/identity.rb
+++ b/app/models/identity.rb
@@ -3,12 +3,12 @@
 #
 # Table name: identities
 #
-#  id         :integer          not null, primary key
 #  user_id    :integer
 #  provider   :string           default(""), not null
 #  uid        :string           default(""), not null
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
+#  id         :bigint(8)        not null, primary key
 #
 
 class Identity < ApplicationRecord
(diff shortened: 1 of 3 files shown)


nightpool:
running cleanup_concurrent_column_type_change in the same migration means that people will have to turn off their web workers while this migration is running, right?

Gargron (approved):

Gargron:
Could you reapply the migration on top of latest schema? I'm sorry for the trouble but we had like 3 PRs affecting the schema at the same time as this.

Gargron:
@aschmitz Because the PR's branch is not on this repository, I can't submit changes to it myself without a manual merge into master. I'd prefer to avoid that.

aschmitz:
I think the GitHub "allow maintainers to push to this branch" thing is
enabled, so you should(?) be able to push, even though it's under my
namespace, if you want. Otherwise, yeah, I should rebase this and
force-push it, but I haven't yet gotten a chance and I'm not sure if I will
soon. I'll try to when I get some time, though.

Gargron:
Do you know how I would do that? What would the push command look like? I never found it.

aschmitz:
I think
https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/
is the help page on it, but I have to imagine there's an easier way. (I
think you could set my repo as an upstream and pull the PR branch from your
repo, but I'm not at a computer to try it.)

PR #8181, opened 2018-08-12
 Display replies count in web UI 

Gargron:
Follow-up to #8104 

image

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -32,6 +32,16 @@ const messages = defineMessages({
   embed: { id: 'status.embed', defaultMessage: 'Embed' },
 });
 
+const obfuscatedCount = count => {
+  if (count < 0) {
+    return 0;
+  } else if (count <= 1) {
+    return count;
+  } else {
+    return '1+';
+  }
+};
+
 @injectIntl
 export default class StatusActionBar extends ImmutablePureComponent {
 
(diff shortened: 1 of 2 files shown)


renatolond:
I can understand this on public pages, but I really dislike this on the web UI. Do you think you could add an option to disable it?

Gargron:
The issue tracker is full of requests like "I wish I knew when posts have replies to them", this is addressing that and also an even more specific request from @ThisIsMissEm (switter.at had to implement reply counts on their own but this implementation is better, I think).

I think it's almost unnoticeable so I don't want to add a way to disable it, however, I might change my opinion after it's seen some real-world use if it indeed turns out to be annoying.

renatolond:
The same way there's a lot of "I wish I knew when posts have replies to them", there's also a lot of posts on Mastodon about how not having an easily accessible counter for likes/boosts/reply removes a bit of the reinforcing loop of interacting with a post because it has interactions already.

Personally I think this starts adding more of the classic social media experience of adding incentive for comments because the post already has replies.

I understand streamlining an implementation that already exists, but I think a toggle, even on instance level if nothing else, would be good for this option.

Gargron:
>not having an easily accessible counter for likes/boosts/reply

I am sorry but I have specifically seen such opinions about *likes and boosts* but not *replies*. Where's the harm in knowing a post leads to a sizeable discussion? Every GitHub issue and every forum thread has a post counter.

m4sk1n:
I love the way how not number-focused Mastodon is, so…

imbsky:
What you are saying is that you don't even need followers count.

renatolond:
> What you are saying is that you don't even need followers count.

I don't think there's need for a followers count under my profile pic, which is equivalent to this case imo.

> I am sorry but I have specifically seen such opinions about likes and boosts but not replies. Where's the harm in knowing a post leads to a sizeable discussion? Every GitHub issue and every forum thread has a post counter.

For me it suffers from the same issue as boosts/favs, it's just a different kind of interaction. I saw on the topic you posted asking about this someone suggesting showing a "there is/there isn't" replies, and for me that is the max of communication I'd like to see in the interface regarding replies, adding numbers for me suffers from "Hm, there's 20 replies in this one, I'll check it out"

trwnh:
I'm echoing general sentiments that it's nice for Mastodon to be less focused on numbers. I think most people can agree that faves/boosts can lead to a positive bias and cause people to value certain posts more than others without actually engaging with them. With that said, I'm not sure the same value judgement can be applied to replies -- i.e., having more replies is not necessarily good. Whereas likes and retweets are generally and unambiguously "good" to most people's psychologies, having more replies can be seen as a *negative* thing, especially in the case where there are more replies than likes/RTs (i.e. "getting ratio'd", as Twitter users would call it).

With that said: it should be considered how important and how significant a counter on replies would be. It can be established that there are both positive and negative value judgements that can be applied to it, but as @renatolond says, it may be enough to show the mere presence of replies. In the same way that a double-arrow is used for toots that *are* replies, there should be some indication that a toot *has* replies. It just doesn't necessarily need to be a number -- a simple dot or underline could do. Or perhaps in keeping with the pattern, use a triple-arrow? Or at least, putting a silhouette on the other end, or perhaps even repurposing the double-arrow icon to indicate a conversation in either direction.

image

PR #8185, opened 2018-08-13
Unuse ActiveRecord::Base#cache_key

abcang:
Change to give the instance key instead of cache_key to the key of the cache. Rails automatically processes the version information by specifying the instance as a key.


app/controllers/accounts_controller.rb
```diff
@@ -42,7 +42,7 @@ def show
       format.json do
         skip_session!
 
-        render_cached_json(['activitypub', 'actor', @account.cache_key], content_type: 'application/activity+json') do
+        render_cached_json(['activitypub', 'actor', @account], content_type: 'application/activity+json') do
```
abcang: If an array is given to the key of Rails.cache.fetch, the key will be like `activitypub/actor/accounts/1`.

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -17,7 +17,7 @@ class Api::V1::StatusesController < Api::BaseController
   CONTEXT_LIMIT = 4_096
 
   def show
-    cached  = Rails.cache.read(@status.cache_key)
+    cached  = Rails.cache.read(@status)
```
abcang: It is not related to this PR, but the cached account information may be old ...

Gargron:
Could you explain the significance of these changes? I do not understand

abcang:
When cache_versioning is true from Rails 5.2 cache_key no longer includes the value of updated_at, and version information such as the value of updated_at is now managed by ActiveSupport::Cache::Store. There is an advantage that the number of keys increases for each version, and wasteful caching is not created.
Version information can be given by using the version option of methods such as Rails.cache.write, but if an instance of ActiveRecord is given to the key of cache, the value of updated_at is automatically used as version.
ActiveRecord of Rails 5.2 cache_versioning is true by default. By setting cache_versioning to false it is possible to include the value of updated_at in cache_key as before, but it is better to keep cache_versioning to be true for cache efficiency.

Gargron:
@abcang How does this PR interact with the statuses/statuses_stats separation? Currently we fetch updated_at from statuses_stats if it's more recent. But if we won't use updated_at, then updates on StatusStat need to bump the Status' cache version

abcang:
When cache_collection is used, since updated_at of statuses_stats is read by Status.cache_ids, it works correctly. I think that it will not work properly if Status.cache_ids is not used. Also, since json which is used for ActivityPub does not use information of statuses_stats, it seems no problem.

Gargron (approved):

PR #8198, opened 2018-08-14
Fix null constraint violation in copy status stats migration

Gargron:


ClearlyClaire (approved):
I don't know what CircleCI is doing, but looks good to me.

PR #8200, opened 2018-08-14
Add "sign in to participate" message to public toot page

Gargron:
image

### DIFF
--- a/app/javascript/styles/mastodon/stream_entries.scss
+++ b/app/javascript/styles/mastodon/stream_entries.scss
@@ -36,7 +36,8 @@
 
     &:last-child {
       .detailed-status,
-      .status {
+      .status,
+      .load-more {
         border-bottom: 0;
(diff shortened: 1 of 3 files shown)


ThibG:
Hm… I'm not too sure about this one… you already have a “Sign in” button on the page (although it's on the top of the page), I'm not too sure this new button brings much value… unless it somehow redirects to that exact thread in the Web UI, which does not appear to be the case.

Gargron:
The value is in a call-to-action... "Sign in" is practical, but here we are saying "here is what you can do with this". For example, Instagram has a "Log in to like or comment" on every post.

ThibG:
Hum, ok… but if you're going to make such a link, I think it should redirect to the conversation once logged in.
(Sorry, I thought this would be easy to do, but after having a look at the auth controller, I know that my request is going to be a PITA to implement)

ykzts (approved):

PR #8202, opened 2018-08-14
Add remote interaction dialog for toots

Gargron:
Same mechanism as remote follow, but for toots. Buttons on public toot pages become clickable, open the modal dialog for remote interactions. Eventually opens the web UI with the status in dialog window.

Also:

* Fixes `.modal-button` dialog opening
* Changes `/authorize_follow?acct=` to `/authorize_interaction?uri=`
* Hides account/logout header from remote follow/interaction dialog
* Display reply counts (0, 1, 1+) next to reply button on public toots

image

- [x] Remote interaction dialog
- [x] Authorize remote interaction dialog


ykzts (approved):

rugk:
> Changes /authorize_follow?acct= to /authorize_interaction?uri=

Does this change affect _all_ these remote actions, i.e. also replaces the old remote follow?
I need to know this, for [my remote auto-follow add-on](https://github.com/rugk/mastodon-auto-follow).

Gargron:
@rugk The acct param still works so you don't need to worry.

rugk:
Okay, great. Thanks.

BTW as for the URL that is shown when you click on interaction buttons of a toot, i.e. e.g. this one: `https://mastodon.social/interact/100605415617223528`, I think that is quite bad for my use case. As I cannot just assume this is Mastodon. The URL is not unique enough – compared to the one having `remote_follow` inside of it.

rugk:
You say the old URI is not replaced. During my testing, I found out something else.

Accessing this URL, only shows a 404 error: https://mastodon.social/authorize_follow?acct=WelshPixie@mastodon.social

Only this works: https://mastodon.social/authorize_interaction?acct=WelshPixie@mastodon.social

As such, it's still not the same URL. :laughing:

Gargron:
Oh, right, I misread that.

When deciding to change it, I considered that it requires a webfinger lookup to get the URL template, and it's not cached, so changing the URL would not be a problem. So what's the best way forward? Can you adjust this in your extension, or do you want the old URL to redirect to the new one?

rugk:
I guess I could also just do a webfinger lookup. (Just don't know what to call there… :laughing:)

But if the old version could redirect to the new one, I would prefer that, of course.

eloquence:
This is a great feature, but it seems to also require providing instance information even *if you're logged in on the same instance where you're trying to interact with a toot.* So, for example, if I view https://mastodon.social/@RJanse59/100626309128964888 on mastodon.social and click the "Reply" button, it asks me to enter a username to interact from, even though I am logged in on mastodon.social. It even suggests I might want to create an account.

We probably can't assume that I want to reply through my mastodon.social account, but I would argue that we should at least make that a seamless default (i.e. indicate the user name you're logged in from, and offer the option to "Reply from a different account" or similar). Personally, I would argue that this should be done before the final 2.5.0 release, because it would make the whole feature more seamless and reduce the potential for confusion of users who end up in this interaction flow from their home instance.

PR #8203, opened 2018-08-14
Use Rails' counter caches through size, restore old spec

ClearlyClaire:
https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache

### DIFF
--- a/app/models/export.rb
+++ b/app/models/export.rb
@@ -29,11 +29,11 @@ class Export
   end
 
   def total_follows
-    account.following_count
+    account.following.size
   end
 
   def total_followers
-    account.followers_count
+    account.followers.size
   end
 
   def total_blocks
--- a/spec/models/export_spec.rb
+++ b/spec/models/export_spec.rb
@@ -48,17 +48,17 @@ describe Export do
   describe 'total_follows' do
     it 'returns the total number of the followed accounts' do
       target_accounts.each(&account.method(:follow!))
-      expect(Export.new(account.reload).total_follows).to eq 2
+      expect(Export.new(account).total_follows).to eq 2
     end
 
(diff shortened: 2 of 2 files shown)


ThibG:
This basically implements the change I requested for #8199. According to the documentation, when a counter cache is set up (which is our case), `size` uses that. Otherwise, if the records are loaded, it uses `length` and if that's not the case, it uses `count`.
It seems to re-read the cache counter as needed, producing the old results.

Gargron:
I don't really get why this is important. For example followers relation will never be loaded *entirely* so `Array#size` fallback is useless (would be incorrect). So the only fallback it gives you is `Relation#count`. I like calling the attribute explicitly instead because there is no situation where I would want that count to actually be executed live. The only other thing is that it implicitly reloads the attributes if the followers collection changed on the loaded object, but that's a scenario only valid for the unit test.

ThibG:
> I don't really get why this is important.

It's not extremely important, it just seems more natural to actually use cache counters the way they are intended.

> For example followers relation will never be loaded *entirely* so `Array#size fallback` is useless (would be incorrect).

Yup, we don't really care about this one.

> So the only fallback it gives you is `Relation#count`.

Not really a fallback but the whole point of this.

> I like calling the attribute explicitly instead because there is no situation where I would want that count to actually be executed live.

Fair enough, even though it will never do the query live, because you have a counter cache set up.

> The only other thing is that it implicitly reloads the attributes if the followers collection changed on the loaded object, but that's a scenario only valid for the unit test.

Is that really the case? That's the main reason for this change. You could also explicitly reload the attribute if the collection changed, instead of calling `#size` if you really want to avoid that.

Gargron:
I'm sorry, I don't think this is a good change.

PR #8211, opened 2018-08-15
Update reply counters only if the reply is public/unlisted

Gargron:


ThibG:
Hm, it would be a bit confusing to see that there is no reply when you can actually read one.
Also, what about public → unlisted downgrades? I think it makes more sense to just not count anything more private than unlisted?

Gargron:
Makes sense. Changed.

ClearlyClaire (approved):

PR #8219, opened 2018-08-16
Prevent actions log from crashing when displaying deleted status (fixes #8133)

### LINKED ISSUE #8133: Error when consulting audit log after a reported toot was deleted by the user
I reported a toot, then marked it as sensitive. Then, the user deleted their toot.
Now I can't consult audit log (error 500).

Here's some log from journalctl : 

```
method=GET path=/admin/action_logs format=html controller=Admin::ActionLogsController action=index status=500 error='ActionView::Template::Error: No route matches {:account_username=>nil, :action=>"show", :controller=>"statuses", :id=>#<Status id: nil, uri: nil, text: "", created_at: nil, updated_at: nil, in_reply_to_id: nil, reblog_of_id: nil, url: nil, sensitive: true, visibility: "public", spoiler_text: "", reply: false, favourites_count: 0, reblogs_count: 0, language: nil, conversation_id: nil, local: nil, account_id: nil, application_id: nil, in_reply_to_account_id: nil, local_only: nil>}, possible unmatched constraints: [:account_username, :id]' duration=27.52 view=0.00 db=5.96

ActionView::Template::Error (No route matches {:account_username=>nil, :action=>"show", :controller=>"statuses", :id=>#<Status id: nil, uri: nil, text: "", created_at: nil, updated_at: nil, in_reply_to_id: nil, reblog_of_id: nil, url: nil, sensitive: true, visibility: "public", spoiler_text: "", reply: false, favourites_count: 0, reblogs_count: 0, language: nil, conversation_id: nil, local: nil, account_id: nil, application_id: nil, in_reply_to_account_id: nil, local_only: nil>}, possible unmatched constraints: [:account_username, :id]):
    4:       = image_tag action_log.account.avatar.url(:original), alt: '', width: 40, height: 40, class: 'avatar'
    5:     .log-entry__content
    6:       .log-entry__title
    7:         = t("admin.action_logs.actions.#{action_log.action}_#{action_log.target_type.underscore}", name: content_tag(:span, action_log.account.username, class: 'username'), target: content_tag(:span, log_target(action_log), class: 'target')).html_safe
    8:       .log-entry__timestamp
    9:         %time= l action_log.created_at
  10:     .spacer
app/lib/tag_manager.rb:44:in `url_for'
app/helpers/admin/action_logs_helper.rb:37:in `log_target_from_history'
app/helpers/admin/action_logs_helper.rb:8:in `log_target'
app/views/admin/action_logs/_action_log.html.haml:7:in `_app_views_admin_action_logs__action_log_html_haml__3907127229702845673_69936307206560'
app/views/admin/action_logs/index.html.haml:4:in `_app_views_admin_action_logs_index_html_haml___220621279602622634_69936306619620'
```

As a workaround, I deleted the audit log entry and the report form the database.

* * * *

- This bug happens on master, commit : https://github.com/tootsuite/mastodon/commit/4da03a298a20e5007718764483544593556a5ed4

ClearlyClaire:
When a status is updated (not deleted or created) in the admin view, only its id and the actual modification are saved. This meant that if the status got deleted afterwards, the status was deleted afterward, the code was still trying to link to it.

EDIT: This is not perfect, and it can't be (because we haven't the required information in the database), but here is what it looks like:
screenshot_2018-08-16 audit log - dev instance

### DIFF
--- a/app/helpers/admin/action_logs_helper.rb
+++ b/app/helpers/admin/action_logs_helper.rb
@@ -34,7 +34,11 @@ module Admin::ActionLogsHelper
       link_to attributes['domain'], "https://#{attributes['domain']}"
     when 'Status'
       tmp_status = Status.new(attributes)
-      link_to tmp_status.account&.acct || "##{tmp_status.account_id}", TagManager.instance.url_for(tmp_status)
+      if tmp_status.account
+        link_to tmp_status.account&.acct || "##{tmp_status.account_id}", admin_account_path(tmp_status.account_id)
+      else
+        I18n.t('admin.action_logs.deleted_status')
+      end
     end
   end
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -184,6 +184,7 @@ en:
         unsuspend_account: "%{name} unsuspended %{target}'s account"
         update_custom_emoji: "%{name} updated emoji %{target}"
         update_status: "%{name} updated status by %{target}"
+      deleted_status: "(deleted status)"
       title: Audit log
     custom_emojis:
       by_domain: Domain


Gargron (approved):
Ah...

Issue #7710, opened 2018-06-02
uWebSockets and FreeBSD


nanaya:
I don't know how much you guys care about FreeBSD compatibility, but switching to `uws` caused streaming server to stop working because [it doesn't support FreeBSD](https://github.com/uNetworking/uWebSockets/wiki/Misc.-details).

```
/home/edho-mastodon/mastodon/node_modules/uws/uws.js:38
            throw new Error('Compilation of µWebSockets has failed and there is no pre-compiled binary ' +
            ^

Error: Compilation of µWebSockets has failed and there is no pre-compiled binary available for your system. Please install a supported C++11 compiler and reinstall the module 'uws'.
```

Thankfully it seems to be still mostly compatible with the original `ws` module so switching back to it and updating one deprecation fixes it (or at least seems to fix it, haven't done extensive test).

I imagine something like OS check to figure out which websocket module to use would allow using best available websocket module for each OSes.

The diff looks like something like this (ignore `fs` import which I'm using to listen to socket file instead of ip address).

```diff
diff --git a/streaming/index.js b/streaming/index.js
index 4eaf66865..60962fd87 100644
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -7,8 +7,9 @@ const redis = require('redis');
 const pg = require('pg');
 const log = require('npmlog');
 const url = require('url');
-const WebSocket = require('uws');
+const WebSocket = require('ws');
 const uuid = require('uuid');
+const fs = require('fs');

 const env = process.env.NODE_ENV || 'development';

@@ -503,8 +504,11 @@ const startWorker = (workerId) => {

   const wss = new WebSocket.Server({ server, verifyClient: wsVerifyClient });

-  wss.on('connection', ws => {
-    const req      = ws.upgradeReq;
+  wss.on('connection', (ws, req) => {
+    // stolen from https://github.com/herenow/wsProxy/pull/5
+    if (ws.upgradeReq === undefined) {
+        ws.upgradeReq = req;
+    }
     const location = url.parse(req.url, true);
     req.requestId  = uuid.v4();
     req.remoteAddress = ws._socket.remoteAddress;
```

ignisf:
Hey, are you sure it's due to lack of FreeBSD support and not because the current uws dep version does not support node 10? Could you check the build log?

nightpool:
closing as dupe of #6853/#7448 (tracking issue for node 10 support)

nanaya:
[...]

PR #8223, opened 2018-08-16
replace sabotaged uws with ws

devkral:
Recently the uws package was sabotaged by its maintainer and causes now sometimes runtime errors.
see:
https://www.npmjs.com/package/uws
and
https://www.reddit.com/r/node/comments/91kgte/uws_has_been_deprecated/

Sorry for replacing a technical better solution by an inferior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "tiny-queue": "^0.2.1",
     "uglifyjs-webpack-plugin": "^1.2.7",
     "uuid": "^3.1.0",
-    "uws": "10.148.0",
+    "ws": "6.0.0",
     "webpack": "^4.16.0",
     "webpack-bundle-analyzer": "^2.13.1",
     "webpack-cli": "^3.0.8",
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -7,7 +7,7 @@ const redis = require('redis');
 const pg = require('pg');
 const log = require('npmlog');
 const url = require('url');
-const WebSocket = require('uws');
+const WebSocket = require('ws');
 const uuid = require('uuid');
 
 const env = process.env.NODE_ENV || 'development';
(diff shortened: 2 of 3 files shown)


devkral:
"npm run start" fails under some unknown circumstances and is not repairable.
I helped somebody setting up a mastodon instance. Sry, I forgot to ask for the log.

Gargron:
Hm. So I've read up on it and the developer released an empty version and deprecated the project (apparently in protest of npm semver practices??). But older versions should still be okay, right? Can we remain with uws until there an issue? Also, someone might fork it. I don't know the exact benchmarks between these libraries but some Mastodon servers handle very big loads so I would prefer not to sacrifice any performance whenever possible.

nightpool:
npm packages are (to my understanding) immutable, so i don't think there's any problem with remaining on the current version.

Gargron:
In that case I will close this, but thank you for the submission.

nanaya:
My patch in #7710 seems to still work for replacing uws back to ws. Not sure about the performance though as I'm using it on my own single user instance so performance doesn't really matter.

ProgVal:
Ping.

uws no longer compiles properly on Debian Buster (because of [this bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=895672)). I had to restore an old uws binary from my backups to fix mastodon-streaming.

Also, uws' error handling is terrible, no matter the actual error, it prints: `Compilation of µWebSockets has failed and there is no pre-compiled binary available for your system. Please install a supported C++11 compiler and reinstall the module 'uws'.`

nightpool:
if you want uws replaced with ws, you'll have to show compelling evidence
that it doesn't introduce a performance regression. the bug you mentioned
seems to be an npm bug—the fact that it breaks uws is just coincidence, so
I don't think it provides evidence to remove the package.

ProgVal:
I'm adding this to @ddevault 's argument that this dependency is a pain for some instance admins. It's not the first time that I have to deal with its issues either.

I have no idea about the perfs, but it's wasting my time (and apparently some other people's time as well)

Gargron:
Soon https://github.com/tootsuite/ragequit will become available

PR #8231, opened 2018-08-17
Revert "Upgrade Doorkeeper to 4.4.1 (#8197)"

h3poteto:
This reverts commit 464daffdf9a37e9a773d224a162fad022890d463.

Now Doorkeeper v4.4.1 has a problem about `native_redirect_uri`, so some client applications can not login.
https://github.com/doorkeeper-gem/doorkeeper/issues/1130

The fix will be released at Doorkeeper v4.4.2. 
So please wait for upgrade Doorkeeper before release v4.4.2.


ThibG:
We probably shouldn't revert the db changes, though? (Or do that with an additional migration)

h3poteto:
Oh, I forgot about migration.
Is it safe to revert the source code by leaving migration as it is?

h3poteto:
I excepted migration files from this revert.

ThibG:
I would also avoid touching `db/migrate/20171201000000_change_account_id_nonnullable_in_lists.rb` but otherwise looks good to me.

Gargron:
Doorkeeper before 4.4.0 has a security issue. Perhaps upgrade to git master is necessary.

h3poteto:
@Gargron OK, I see. 
But now, master branch of Doorkeeper is a branch for 5.x, and Doorkeeper 5.x is RC2; not stable version.
There is `4.4.0` branch for 4.x, but the fix is not merged.
https://github.com/doorkeeper-gem/doorkeeper/pull/1136

So, now I can not upgrade Doorkeeper.

Gargron:
The fix was submitted only 21 hours ago, perhaps it will be merged soon. We are not in release candidate phase, so there is no rush, right? Let's wait.

Gargron (approved):

PR #8245, opened 2018-08-17
Implement Undo { Accept { Follow } } (fixes #8234)

### LINKED ISSUE #8234: Handle ActivityPub Undo Accept
How to force-unfollow remote followers isn't explicitly specified in the ActivityPub spec.
However, it makes sense to me that it could look like the following
1. `Undo` the `Accept` of the `Follow` request
2. `Reject` the `Follow` request

Currently, Mastodon should handle the `Reject`, but it would also make sense to handle the `Undo`.
The most logical way to handle the `Undo` would be to revert the `follow` relation to a `follow_request`.
This may also make sense if we end up implementing something like #8233.

Also, even if this is not useful to us right now, I would argue this should be implemented anyway, so that Mastodon could handle this activity correctly coming from other software or future Mastodon versions.

ClearlyClaire:
Receiving a Undo corresponding to a Follow from a local account to the sender is processed by reverting the Follow to a FollowRequest.

### DIFF
--- a/app/lib/activitypub/activity/undo.rb
+++ b/app/lib/activitypub/activity/undo.rb
@@ -5,6 +5,8 @@ class ActivityPub::Activity::Undo < ActivityPub::Activity
     case @object['type']
     when 'Announce'
       undo_announce
+    when 'Accept'
+      undo_accept
     when 'Follow'
       undo_follow
     when 'Like'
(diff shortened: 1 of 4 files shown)


ThibG:
I'm not sure what's going wrong :weary:

Gargron (approved):

PR #8278, opened 2018-08-18
Link to mobile apps page

### LINKED ISSUE #8269: re-include link to list of apps in the sidebar
No idea why it was removed. I think its a quite useful link and it only needs 4 letters.

https://github.com/tootsuite/documentation/blob/master/Using-Mastodon/Apps.md

screen shot 2018-08-18 at 12 00 32

Gargron:
Fix #8269

### DIFF
--- a/app/views/layouts/public.html.haml
+++ b/app/views/layouts/public.html.haml
@@ -45,6 +45,6 @@
             %h4= t 'footer.more'
             %ul
               %li= link_to t('about.source_code'), Mastodon::Version.source_url
-              %li= link_to 'joinmastodon.org', 'https://joinmastodon.org'
+              %li= link_to t('about.apps'), 'https://joinmastodon.org/apps'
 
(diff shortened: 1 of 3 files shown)


app/views/layouts/public.html.haml
```diff
@@ -45,6 +45,6 @@
             %h4= t 'footer.more'
             %ul
               %li= link_to t('about.source_code'), Mastodon::Version.source_url
-              %li= link_to 'joinmastodon.org', 'https://joinmastodon.org'
+              %li= link_to t('about.apps'), 'https://joinmastodon.org/apps'
```
ThibG: Not sure about this one

Gargron: "What is mastodon?" links to joinmastodon.org so I figured there's no need to have duplicates.

ThibG:
Oh yeah OK. LGTM then.

PR #8281, opened 2018-08-18
Do not use WHERE NOT IN in CopyStatusStats migration

### LINKED ISSUE #8275: CopyStatusStats Migration consumes a lot of time
I found a bug that CopyStatusStats Migration consumes a lot of time.
The cause commit is https://github.com/tootsuite/mastodon/pull/8256/commits/f1fd7b79a92d675b6e71fbe61402209f0090d4ba .

Please try to remove "where.not (id: StatusStat.select ('status_id')). "
I confirmed that it works well.

Gargron:
Fix #8275

As the batch operation progresses, the statuses_stats table grows,
and the WHERE NOT IN subquery becomes more expensive

### DIFF
--- a/db/migrate/20180812173710_copy_status_stats.rb
+++ b/db/migrate/20180812173710_copy_status_stats.rb
@@ -3,7 +3,7 @@ class CopyStatusStats < ActiveRecord::Migration[5.2]
 
   def up
     safety_assured do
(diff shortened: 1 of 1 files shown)


Gargron:
@ThibG I would say those are edge cases and accuracy isn't super important anyway for this kind of data.

ThibG:
Right. And I guess it was already the case with the previous version of the script. The faster the better then.

PR #8282, opened 2018-08-18
Make user active duration configurable via environment variable

Gargron:
For smaller instances with people who always come back, this value may be raised.

For bigger instances with more turnover, this value may be lowered to decrease server load.


Gargron:
Ah, maybe it should be an env variable

Gargron:
(Updated PR)

ykzts (approved):

PR #8292, opened 2018-08-19
Make dropdown animations respect their placement

c-probably-archived20260718:
(Kind of a follow-up to #6091)

Makes dropdowns (including the privacy dropdown) using top placement have the correct transform origin.
gif of dropdowns being opened with correct transform origins

Also makes the privacy dropdown button not look strange when using top placement.
screenshot of the privacy dropdown button with top placement correct rounded corners


Gargron (approved):

renatolond:
After this merge I get:
```
Warning: Failed prop type: The prop `placement` is marked as required in `PrivacyDropdownMenu`, but its value is `null`.
```

On my console when the page is opened. Should placement really be required? it doesn't seem like the case for the other component.

PR #8298, opened 2018-08-19
Don't filter own toots (fixes #8289)

### LINKED ISSUE #8289: [Feature] Don't filter my own posts (or option to show own posts?)
Say if I'm filtering the word "Trump", if I post saying "I'm filtering the word Trump, just so you know" then that post isn't visible to me.

I'd like to be able to see my own posts (which are clearly acceptable and not triggering to me, because I wrote them). If that's not a universal want, then I'd like to have the option. Maybe in amongst these settings?:

screen shot 2018-08-19 at 09 52 08

* * * *

ClearlyClaire:
fixes #8289

### DIFF
--- a/app/javascript/mastodon/selectors/index.js
+++ b/app/javascript/mastodon/selectors/index.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect';
 import { List as ImmutableList } from 'immutable';
+import { me } from '../initial_state';
 
 const getAccountBase         = (state, id) => state.getIn(['accounts', id], null);
 const getAccountCounters     = (state, id) => state.getIn(['accounts_counters', id], null);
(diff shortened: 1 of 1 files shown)


nightpool (commented):
does this also fix "drop instead of hide" filters or are those not broken?

app/javascript/mastodon/selectors/index.js
```diff
@@ -83,7 +84,7 @@ export const makeGetStatus = () => {
         statusReblog = null;
       }
 
-      const regex    = regexFromFilters(filters);
+      const regex    = (accountReblog || acountBase).get('id') !== me && regexFromFilters(filters);
```
Gargron: acount -> account

ThibG:
@nightpool they aren't broken, own toots are never filtered in the backend afaict
https://github.com/tootsuite/mastodon/blob/20fefdb7145b48449e9d6a824a9846dadfd6ac2f/app/lib/feed_manager.rb#L153-L154

nightpool:
sgtm!

Gargron (approved):

PR #8300, opened 2018-08-19
Load statuses from database instead of re-fetching them from search box

ClearlyClaire:
This is a “workaround” for #8214, making it less useful, and less likely to leak user identity if we end up implementing it.

### DIFF
--- a/app/services/resolve_url_service.rb
+++ b/app/services/resolve_url_service.rb
@@ -12,6 +12,10 @@ class ResolveURLService < BaseService
 
     return process_local_url if local_url?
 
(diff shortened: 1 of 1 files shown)


Gargron:
Problem: URI != URL, e.g. on mastodon, URI is /users/alice/statuses/123, while URL is /@alice/123

ThibG:
Ah, yeah, I expected this to be a corner case, but if it doesn't match in Mastodon itself…

Gargron:
So I assume this can be closed

PR #8305, opened 2018-08-20
Increase reach of Delete->Actor activities

### LINKED ISSUE #7316: Account / toot deletion should be broadcast to all known instances
Account and toot deletion should be broadcast to all known instances.
Indeed, account deletion is currently broadcast to followers, which is not nearly enough, as it is frequent for account information to be fetched by an instance without any follower.
Toot deletion mishandling is a little less frequent, but there can still be corner cases: e.g., unfollowing everyone that has boosted a toot from an un-followed user.

* * * *

Gargron:
Fix #7316

### DIFF
--- a/app/services/suspend_account_service.rb
+++ b/app/services/suspend_account_service.rb
@@ -23,9 +23,7 @@ class SuspendAccountService < BaseService
 
   def purge_content!
     if @account.local?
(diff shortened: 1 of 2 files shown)


ClearlyClaire (approved):
LGTM!

Issue #8177, opened 2018-08-12
Larger warning textbox


toadlyBroodle:
It would be nice to have the size of the warning input textbox increased to multilines for easier input of longer visible content. Running it in Firefox ATM only allows single line input textbox with no word wrapping.

Also, pressing *up_arrow* key while editing text in warning input textbox would improve UX. It's the little things ;) 

- [ x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #8308, opened 2018-08-20
change the spoiler input to a textarea

jacobherrington:
Refs #8177 

This PR lets the user see two lines (with word wrapping) while editing their spoiler message.

1e7daa41-9af0-4362-bb36-ef825db4bf52

This is my first contribution to Mastodon and I'm fairly new to React so feedback is appreciated.


ThibG:
If we do go this way, I would prefer the TextArea to default to one line of height and grow as needed (you may want to use `Textarea` from `react-textarea-autosize` for that).

jacobherrington:
@ThibG Fixing the trailing spaces now, will look into `react-textarea-autosize`. Thanks!

PR #8314, opened 2018-08-20
Add preference to filter registered bots from timelines

Tak:
Adds a user preference for filtering statuses bots from timelines and notifications, based on account `actor_type`

This affects
- local timeline
- federated timeline
- DM timeline
- tag timelines
- notifications

I explicitly didn't touch the home timeline, since if you're following a bot, or somebody you're following boosts a bot's status, that's probably "not bot" enough.

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -47,6 +47,7 @@ class Settings::PreferencesController < ApplicationController
       :setting_noindex,
       :setting_theme,
       :setting_hide_network,
+      :setting_filter_bots,
       notification_emails: %i(follow follow_request reblog favourite mention digest report),
       interactions: %i(must_be_follower must_be_following)
     )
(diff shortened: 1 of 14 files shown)


config/webpacker.yml
```diff
@@ -36,6 +36,7 @@ development:
 
 test:
   <<: *default
+  compile: true
```
Tak: This is apparently necessary to pass the 2.5 circleci config, since I touched `simple_form.en.yml`

app/controllers/settings/preferences_controller.rb
```diff
@@ -46,6 +46,7 @@ def user_settings_params
       :setting_noindex,
       :setting_theme,
       :setting_hide_network,
+      :setting_filter_bots,
```
Gargron: There are two approaches to the desired result:

- User setting that affects all queries
- Query param that affects queries like `local`, `only_media`, and web UI setting to store param state

Why did you choose the former?

Tak: I guess it felt like a similar setting to `chosen_languages`, so I followed a similar path.
I also thought that it would make sense to be able to synchronize this preference with third-party apps.

Tak: Was...that a request for action? Or just a question?

PR #8322, opened 2018-08-20
reorder simple form ul to list vertically (fixes #8236)

### LINKED ISSUE #8236: [UX] Put languages in filter list into a helpful order
Here's what they look like right now:

screen shot 2018-08-17 at 10 59 17

Having "select the ones you want to see" instead of "select the ones you don't want to see" is a great change, but it's now really hard to find the languages you do want to see! A-Z left-to-right in columns is really hard to read and impractical. Maybe A-Z top-to-bottom would be better?

* * * *

NecroTechno:
Modified the filter languages list so that they are listed alphabetically vertically rather than horizontally.

screenshot from 2018-08-20 20-31-45

### DIFF
--- a/app/javascript/styles/mastodon/forms.scss
+++ b/app/javascript/styles/mastodon/forms.scss
@@ -154,9 +154,10 @@ code {
       margin-bottom: 15px;
     }
 
-    li {
-      float: left;
-      width: 50%;
+    ul {
+      columns: 2;
+      -webkit-columns: 2;
+      -moz-columns: 2;
     }
   }
 
--- a/app/javascript/styles/mastodon/forms.scss
+++ b/app/javascript/styles/mastodon/forms.scss
@@ -156,8 +156,6 @@ code {
 
     ul {
       columns: 2;
-      -webkit-columns: 2;
-      -moz-columns: 2;
     }
   }
 


Cassolotl:
Are they sorted by their name in English rather than the visible name? Because Bahasa Indonesia comes after Magyar, and Suomi comes before Français! That doesn't seem right. (It's better than it was though.)

app/javascript/styles/mastodon/forms.scss
```diff
@@ -154,9 +154,10 @@ code {
       margin-bottom: 15px;
     }
 
-    li {
-      float: left;
-      width: 50%;
+    ul {
+      columns: 2;
+      -webkit-columns: 2;
```
Gargron: This and below is not necessary, we have autoprefixer generating these.

Gargron (changes requested):

NecroTechno:
@Cassolotl it's currently sorted by the language's [ISO 639-1 code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). For example, Bahasa Indonesia comes just before Italiano as their ISO 639-1 codes are "id" and "it" respectively.

I definitely understand the desire to order it purely alphabetically, but I'm conscious of both:
1) changes to the existing functionality of displaying the localised names of the languages and
2) ordering it alphabetically as to group languages with non-basic latin letters together.

Honestly, I'm not really sure how else to tackle that particular problem.

app/javascript/styles/mastodon/forms.scss
NecroTechno: Oops, my bad. I've updated this now.

Gargron (approved):

Cassolotl:
@NecroTechno Alright, thanks for the explanation. :) It's better than it was before, so I'm into it! Thanks for working on this.

trwnh:
Yeah, like I said in https://github.com/tootsuite/mastodon/issues/8236#issuecomment-413833552, "Suomi" is technically "Finnish", so that's why it comes before "Français" (and after "Farsi"). :)

PR #8329, opened 2018-08-20
Allow mods to disable login, improve message when login disabled

Gargron:
Previously, only admins could enable/disable user login, but it's more useful for mods instead of suspending.

Previously, disabled login led to login form message "Your account is not activated", which confused people because it's the message relating to e-mail confirmations. Now, when login is disabled you can login but see a "You don't have permission to view this page" message.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -7,6 +7,8 @@ class Api::BaseController < ApplicationController
   include RateLimitHeaders
 
   skip_before_action :store_current_location
+  skip_before_action :check_user_permissions
+
   protect_from_forgery with: :null_session
 
   rescue_from ActiveRecord::RecordInvalid, Mastodon::ValidationError do |e|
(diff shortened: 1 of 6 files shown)


nightpool:
it would be really useful if we could set a custom message to show users when we disable their login

Sylvhem:
> Now, when login is disabled you can login
I think that's also really confusing. Is there a reason we can't just display them a specific message like "You can't login" or "Login is disabled for this account"?

Gargron:
@Sylvhem I just think this feature is a lot more useful as a temporary lockout for moderation purposes and in that way destroying the user session does not seem necessary.

ashleyhull-versent:
If the account is being placed on hold... I'd be nice to have a custom message why?
"You're been placed in the fridge because of X... we will unlock your account tomorrow"

and consider giving mods the ability to disable for a set time - "disable account for 1 day" with message

Gargron:
Not planning to include any new data structures for messages or timeouts in this particular PR.

PR #8339, opened 2018-08-21
Revert to using Paperclip's filesystem storage, and fix dangling records in remove_remote

ClearlyClaire:


ThibG:
Actually, that may be too much of performance hit, ̀#exists?` checks for the actual existence of the file afaict.

ThibG:
Ok, for some reason, there are some media for which `#exists?` is false while `#present?` is true.
This shouldn't happen, I don't know why it did, but I updated the worker to destroy the attachment (and thus the record claiming the file exists) regardless of whether the actual file exists.

ykzts (approved):

Gargron (approved):

Gargron:
How is CircleCI still not finished...

ThibG:
@Gargron I re-pushed because it just refused to run last time I pushed.
It's maybe due to the fact that you are out-running your plan by about 500% :x

Gargron (approved):

gllmhyt:
Can you please release with this fix as soon as possible? My server media storage is getting out of hand and will soon have its drive fully loaded and be unusable: unable to run, unable to boot, requiring manual deletion of files where it should have been automatically deleted.

Issue #8342, opened 2018-08-21
Sidekiq queues are misused


ClearlyClaire:
Sidekiq queues are misused.
Indeed, unless queues are given weight, sidekiq pulls jobs from the queues in the order they are defined.
This means that, if you run sidekiq according to the systemd file defined in https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Production-guide.md, no “mailers” job will run before all “default” ones are processed, no “pull” will be processed before “default” or “mailers” etc.
I don't think they were defined to imply any priority, but it is what they currently do.
In addition, the systemd unit file referenced above do not agree with the order defined in `config/sidekiq.yml`.

I believe there is no reason to not have just three queues, as `critical`, `default`, `low`, with most of the stuff in `default`, and some maintenance jobs like `remote_remove` in low.

- Sidekiq documentation on queues and priority: https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues
- Some tips on using them: https://philsturgeon.uk/2016/11/16/tips-on-sidekiq-queues/

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
I'm against this for the following reasons:

When you scale beyond a single Sidekiq process, you usually split by queue. So I'll have a 5 threads process for mailers, two 45 threads processes for default, same for push, etc. It's very very useful because you can make sure that certain jobs aren't held up by other jobs, e.g. the push queue could be lagging behind but local users would be fine if the default queue is okay.

Secondly the reason I have not changed or renamed the queues so far is that it would require all admins to go into their systemd files and whatnot and change it around. It's a big headache but even more than that, historically people have glossed over changes like that, which results in a lot of troubleshooting issues on this bug tracker and broken servers.

Gargron:
TL;DR: The categorization of jobs is currently pretty good, I would not move push jobs into the default queue at all. Renaming them to be more neutral, i.e. default to critical, push to high, mailers to normal, pull to low, is a noble goal, but probably not possible. I don't think your migration plan is viable because I remember the reason 4 separate queues were needed in the first place, there's no way a release would survive with just 2.

Gargron:
[...]

PR #8344, opened 2018-08-21
Use only two sidekiq queues

ClearlyClaire:
Use only two sidekiq queues, as they implicitly define priority.
Every worker now use the `default` (high-priority) queue except for
the following workers using the `mailers` (low-priority) one:
- DigestMailerWorker (because it already was)
- BackupWorker (because it's a relatively low-priority resource-intensive task)
- LinkCrawlWorker (because it's only used for preview cards, which are not super useful)
- Maintenance::DestroyMediaWorker (because those are a lot of tasks that could disrupt normal operations)
- Maintenance::RedownloadAccountMediaWorker (same as above)
- Maintenance::UncacheMediaWorker (same as above)

It's the suggested first step in solving #8342

### DIFF
--- a/app/workers/activitypub/post_upgrade_worker.rb
+++ b/app/workers/activitypub/post_upgrade_worker.rb
@@ -3,7 +3,7 @@
 class ActivityPub::PostUpgradeWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: 'pull'
+  sidekiq_options queue: 'default'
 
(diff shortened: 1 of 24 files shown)


ThibG:
Closing because it's counter-productive in settings with one sidekiq process per queue (see https://github.com/tootsuite/mastodon/issues/8342#issuecomment-414673295)

PR #8353, opened 2018-08-21
Add confirmation step to account suspensions

Gargron:
image


ykzts (approved):

ashleyhull-versent:
I know it's written clearly but "suspend" kind of means "nuke from orbit"? 🗡

PR #8372, opened 2018-08-22
Improve federated ID validation

Gargron:


ThibG:
This looks ok to me, but I'm a bit uneasy with the mess that this code is becoming, so I would like if someone else also reviewed.

ykzts (approved):

Issue #7092, opened 2018-04-09
Add abbility to change an instance default theme from the administration panel


Sylvhem:
The only way to change an instance's default theme is to edit the `config/themes.yml` file to rename the chosen theme to "default". Sadly, it prevents admin without a sysadmin access to do so. Since there is more and more instances hosted on plateforms like [Masto.host](https://masto.host/) and [MaaStodon](https://maastodon.net/), it may start to be problem.
So, that's why I think there should be way to change an instance default theme from the administration panel.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ThisIsMissEm:
To add more info for any beginner wanting to work on this, you'll need to look at:
- app/views/layouts/application.html.haml
- app/controllers/application_controller.rb
- app/views/admin/settings/edit.html.haml
- app/controllers/admin/settings_controller.rb

And read up on the ruby gem `rails-settings-cached`, particularly, you'll just want `Setting.theme`.

Cassolotl:
Related to *Wanted: ability to import and export themes via the admin settings page #5614*, I think? Not sure if duplicate! But this one has helpful technical pointers so it should stay open. :)

JMendyk:
Hi, I would like to work on this feature. I've implemented draft version of this feature, but I have some questions and would like to ask maintainer/contributor for help and opinion.

Currently, default user's theme points to "default" but is named "Mastodon" in UI. Users that want to use *real* Mastodon theme won't see what they expect when selecting "Mastodon" (bad UX) if sysadmin changes path of default theme in `config/themes.yml`. Likewise, Mastodon doesn't provide users with the clear option to use instance's default theme.

**My target for this feature is to:**
- separate real "Mastodon" theme from concept of instance's default theme,
- let (non sys)admin set default UI theme from a set of existing themes,
- let the user select real "Mastodon" theme,
- let the user select instance's default theme.

**To achieve this, I proposes following steps:**
1. Rename "default" theme to "mastodon" in `config/themes.yml` and in locale files.
2. Treat theme's value "default" as intent to use instance's default theme 
*or* introduce special value for this intent (e.g. "instance_theme")
3. Add instance's default theme selection option to admin settings UI.
4. (translators' help required) Add translation of "Use instance's default theme" to locale files.

**My issue:**
[...]

PR #8381, opened 2018-08-23
Add ability to change an instance default theme from the administration panel (#7092)

JMendyk:
This feature enables administrators to change instance's default theme from web interface without need to modify `config/themes.yml`.

If this pull request gets accepted and makes it's way to release version, I will update documentation's [Running-Mastodon > Customizing](https://github.com/tootsuite/documentation/blob/master/Running-Mastodon/Customizing.md) section to mention this new feature.

### DIFF
--- a/app/lib/settings/scoped_settings.rb
+++ b/app/lib/settings/scoped_settings.rb
@@ -2,6 +2,8 @@
 
 module Settings
   class ScopedSettings
+    DEFAULTING_TO_UNSCOPED = %w().freeze
+
     def initialize(object)
       @object = object
     end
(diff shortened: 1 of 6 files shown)


app/lib/settings/scoped_settings.rb
```diff
@@ -50,15 +54,22 @@ def [](key)
       Rails.cache.fetch(Setting.cache_key(key, @object)) do
         db_val = thing_scoped.find_by(var: key.to_s)
         if db_val
-          default_value = Setting.default_settings[key]
+          default_value = ScopedSettings.default_settings[key]
```
Gargron: I am curious, what was Setting -> ScopedSettings change necessary?

Gargron (approved):

PR #8387, opened 2018-08-23
Give focused status a sensible aria-label for screen readers

### LINKED ISSUE #8192: Give the element that gets focused when navigating with j and k a sensible aria-label for screen readers.
Right now, if using a screen reader in non-virtual mode so one can use the J and K keys to navigate up and down, and 1 through 3 to quickly switch columns, the screen reader NVDA on Windows with Firefox will only speak "document x of y, section" when landing on a new item. On the Mac, because VoiceOver treats the container as something to grab all child text from, it will read something in the order it appears in in the DOM, but it's actually not the most pleasant user experience.

For efficiency when reading, I suggest to add an aria-label attribute to the div tha tgets focus with j and k, and add information in this order, constructed from the toot's data:

- Author screen name, followed by a comma and a space,
- Toot text with all punctuation left intact, followed by another comma and a space. Don't worry if punctuation is contained at the end, screen readers will handle this.
- The time stamp, like "1 minute ago", followed by another comma and a space,
- The username including instance.
- If this is a boost, another comma and space, and then text like "boosted by ", followed by whoever boosted the toot

Granted, the order could be argued about, but over the years, I found this to be the order of information most blind users I talked to) found most useful, based on similar clients for Twitter and other social media platforms.

For example, [this toot of mine](https://toot.cafe/@marcozehe/100542795612909474) would be read:

Marco Zehe, I got back from vacation today. Caught up on Bugmail and other stuff. I brought my build system up to speed. I was hyper-focused all morning. And now I feel a bit like a balloon that the air was let out of., 23h ago, @marcozehe@toot.cafe

* * * *

Gargron:
Fix #8192

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -18,12 +18,12 @@ import classNames from 'classnames';
 // to use the progress bar to show download progress
 import Bundle from '../features/ui/components/bundle';
 
-export const textForScreenReader = (intl, status, rebloggedByText = false) => {
+export const textForScreenReader = (intl, status, rebloggedByText = false, expanded = false) => {
   const displayName = status.getIn(['account', 'display_name']);
 
   const values = [
     displayName.length === 0 ? status.getIn(['account', 'acct']).split('@')[0] : displayName,
-    status.get('search_index'),
+    status.get('spoiler_text') && !expanded ? status.get('spoiler_text') : status.get('search_index').slice(status.get('spoiler_text').length),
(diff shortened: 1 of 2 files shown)


Gargron:
Wait https://github.com/tootsuite/mastodon/issues/8192#issuecomment-415449336:

>> I'm assuming it doesn't make sense to have the aria-label change based on the "show more"/"show less" button
>
> That is exactly what should happen. Only add the CW text if it is hidden and the full text instead when it is showing.

ykzts (approved):

PR #8394, opened 2018-08-23
Preserve hashtag casing in web UI hashtag history

### LINKED ISSUE #8241: [Accessibility] Screen readers like #HashtagsLikeThis, but the Mastodon UI only offers #hashtagslikethis
Screenshot:

screen shot 2018-08-17 at 11 43 06

* * * *

Gargron:
Fix #8241

### DIFF
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -130,7 +130,7 @@ export function submitCompose() {
         'Idempotency-Key': getState().getIn(['compose', 'idempotencyKey']),
       },
     }).then(function (response) {
-      dispatch(insertIntoTagHistory(response.data.tags));
+      dispatch(insertIntoTagHistory(response.data.tags, status));
       dispatch(submitComposeSuccess({ ...response.data }));
 
       // To make the app more responsive, immediately get the status into the columns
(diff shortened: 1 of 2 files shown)


ThibG:
Maybe retain only what's considered a hashtag by the backend, instead of things that look like hashtags?

Gargron:
@ThibG What's the difference? The backend does not preserve casing so working with backend-returned tags won't lead anywhere. At best, take them and scan the text for their occurence case-insensitively, then save whatever case was in the text, but I am pretty sure this works just as well.

ThibG:
@Gargron non-latin alphabetic characters iirc

ThibG:
I'm not sure how that would work in JS, but, can you build case-insensitive regexps out of the recognized tags, and use that to find the tags in the status' text?

app/javascript/mastodon/actions/compose.js
```diff
@@ -390,13 +391,13 @@ export function hydrateCompose() {
...
 
-function insertIntoTagHistory(tags) {
+function insertIntoTagHistory(text) {
   return (dispatch, getState) => {
     const state = getState();
     const oldHistory = state.getIn(['compose', 'tagHistory']);
     const me = state.getIn(['meta', 'me']);
-    const names = tags.map(({ name }) => name);
-    const intersectedOldHistory = oldHistory.filter(name => !names.includes(name));
+    const names = text.match(APPROX_HASHTAG_RE).map(name => name.replace(/^\s?#/, ''));
```
ThibG: Instead of `APPROX_HASHTAG_RE`, you'd use something like `new RegExp('(?:^|[^\\/\)\\w])(' + response.data.tags.map(escapeRegExp).join('|') + ')', 'i')`, I guess.

See 95c8232109f5e59273de48858ede7fbe7ce63d58 for why the regexps don't match between ruby and javascript

Gargron:
@ThibG Implemented.

app/javascript/mastodon/actions/compose.js
```diff
@@ -390,13 +390,13 @@ export function hydrateCompose() {
...
 
-function insertIntoTagHistory(tags) {
+function insertIntoTagHistory(recognizedTags, text) {
   return (dispatch, getState) => {
     const state = getState();
     const oldHistory = state.getIn(['compose', 'tagHistory']);
     const me = state.getIn(['meta', 'me']);
-    const names = tags.map(({ name }) => name);
-    const intersectedOldHistory = oldHistory.filter(name => !names.includes(name));
+    const names = recognizedTags.map(tag => text.match(new RegExp(`#${tag.name}`, 'i'))[0].slice(1));
```
ThibG: Oh yeah, right, the hashtags shouldn't need to be escaped, valid hashtags are only word characters.
I can think of one last corner case, that is, the hashtag being a substring of another string in the status, for instance a URL hash… but I guess that's a rare corner case…

ClearlyClaire (approved):

PR #8396, opened 2018-08-23
Only display web push notifications after API call (fixes #7902)

### LINKED ISSUE #7902: Duplicate/similar web notifications sent
image

Follow and mention notifications are being sent twice, with slightly different text:

Mentions:
- "x mentioned you"
- "you were mentioned by x"

Follows:
- "x followed you"
- "x is now following you"

* * * *

- [x] This bug happens on  [v2.4.2](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

ClearlyClaire:
Fixes duplicate notifications and weird notification-updating behaviors across browsers/operating systems at the cost of slightly more latency.

### DIFF
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -80,15 +80,7 @@ const handlePush = (event) => {
 
   // Placeholder until more information can be loaded
   event.waitUntil(
-    notify({
-      title,
-      body,
-      icon,
-      tag: notification_id,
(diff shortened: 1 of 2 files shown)


Gargron:
I remember I wanted to do it this way and someone stopped me. Gotta find what the reason was, it might be important here.

Gargron:
The discussion was in #7521 but I cannot find the relevant comments now, so I guess we'll just roll with it.

Gargron (approved):

PR #8411, opened 2018-08-24
Add improved CLI interface for removing remote media

### LINKED ISSUE #8187: media:remove_remote is too heavy
Since #7678 , if you run the media:remove_remote task in an environment that does not run for a long time, I think that there are cases where many jobs are added to Sidekiq and the instance may become malfunctioning.

In my instance I use Swift Object Storage to store media.

when execute the media:remove_remote task at last month, more than 30,000 jobs were added to Sidekiq at a time.

Then Object Storage returns a 503 error, most of the jobs was gone on Retry as is.

At that time I could not post images to Mastodon.

After that, it waited for all jobs moved to the Retry-tab over an hour, cleared all the retries with rails console.

The next image is a comparison with the number of ordinary jobs.

I think that there are other instances that do not routinely execute the media:remove_remote task.

Is it possible to insert wait processing according to the environment into this task, or make it possible to select the former format without using Sidekiq?

* * * *

Gargron:
./bin/tootctl media remove --days 7 --background

Make the old rake task point to it. Benefits: It lets you choose whether to process files sequentially, or in background.

Fix  #8187

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -512,14 +512,9 @@ namespace :mastodon do
 
     desc 'Remove cached remote media attachments that are older than NUM_DAYS. By default 7 (week)'
     task remove_remote: :environment do
-      time_ago = ENV.fetch('NUM_DAYS') { 7 }.to_i.days.ago
-      nb_media_attachments = 0
-
-      MediaAttachment.where.not(remote_url: '').where.not(file_file_name: nil).where('created_at < ?', time_ago).select(:id).reorder(nil).find_in_batches do |media_attachments|
-        nb_media_attachments += media_attachments.length
-        Maintenance::UncacheMediaWorker.push_bulk(media_attachments.map(&:id))
-      end
-      puts "Scheduled the deletion of #{nb_media_attachments} media attachments"
+      require_relative '../mastodon/media_cli'
+      cli = Mastodon::MediaCLI.new([], days: ENV['NUM_DAYS'] || 7)
+      cli.invoke(:remove)
     end
 
     desc 'Set unknown attachment type for remote-only attachments'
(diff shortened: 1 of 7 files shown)


ykzts (approved):

ThibG:
I'm not sure about adding yet another interface.
On the other hand, allowing people to run this task sequentially rather than using a queue is good, although the impact should be limited if said queue is really the lowest-priority one.

lib/tasks/mastodon.rake
```diff
@@ -512,14 +512,9 @@ namespace :mastodon do
...
-      time_ago = ENV.fetch('NUM_DAYS') { 7 }.to_i.days.ago
-      nb_media_attachments = 0
-
-      MediaAttachment.where.not(remote_url: '').where.not(file_file_name: nil).where('created_at < ?', time_ago).select(:id).reorder(nil).find_in_batches do |media_attachments|
-        nb_media_attachments += media_attachments.length
-        Maintenance::UncacheMediaWorker.push_bulk(media_attachments.map(&:id))
-      end
-      puts "Scheduled the deletion of #{nb_media_attachments} media attachments"
+      require_relative '../mastodon/media_cli'
+      cli = Mastodon::MediaCLI.new([], days: ENV['NUM_DAYS'] || 7)
```
mayaeh: I think that ".to_i" is requires for "['NUM_DAYS']".

However, I still feel that it is not working properly.

I executed next line.
<code>RAILS_ENV=production NUM_DAYS=60 bundle exec rake mastodon:media:remove_remote</code>

After that, when I executed next line,  multiple jobs were added to Sidekiq.
<code>RAILS_ENV=production bundle exec bin/tootctl media remove --days 60 --background</code>

Issue #3285, opened 2017-05-24
Toot only to the local timeline.


carlosjs23:
Currently there is the option "Unlisted" but as expected the toots are not published in either of the two timelines and the toots are only shown to my followers in their home column, and there are times when I only want to publish in the local timeline of my community.
* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Cassolotl:
This is a good idea! It's also already been suggested, so I recommend adding your +1/thumbs-up/comments to #861. :)

carlosjs23:
Oh thanks, I was looking for the issue and I could not find it.

Issue #2698, opened 2017-05-01
do-not-federate flag


hoodiek:
federation is a strength, but also in some cases, a downside. users should be able to post purely to their own instance, and choose not to federate something. this is for all level of posts, due to the many uses of such a tool, depending on instance type, post intent, etcetera. there are several posts about such a thing, and this solves all of them, and yet no individual post specifies it to be available for all posting levels, so i created this one.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

bortzmeyer:
Yes, this would be really useful (think of the NO_EXPORT attribute of BGP) for things like:

- local announcements '"we will upgrade, and be down one hour")
- messages intended to a specific community (animal rights demo, for a instance dedicated to animal rights)

gordonzola:
There is already a pull request for that, with comments from @Gargron : #1761

hipparkhos:
Something that would be nice in the future would be to implement some sort of "circles", i.e. I post a status, but only for followers which are in a liste of circles. In real life you have different circles of relationships. You might want to limit the exposure of some toots.

"do-not-federate" could be a first step : only followers on the local instance would see the status. That would be very useful both for admin messages, and for thematic instances.

alex73630:
Possible duplicate of this issue : https://github.com/tootsuite/mastodon/issues/861

wolfteeth:
+1. I was thinking about how useful this would be for announcements AND to have the old style of local-only private toots. Being able to turn off federation for some posts seems very valuable. It lets users avoid the concerns of private toot federation, instead of making private toots useless for anyone with followers running something besides Mastodon.

ashfurrow:
This looks like a duplicate of https://github.com/tootsuite/mastodon/issues/861 – I'm going to close this issue but please continue the discussion over there. Thanks again for the suggestion!

Issue #896, opened 2017-04-05
The ability to post ONLY to the local TL on the instance it originates from


WDC1:
[Issue text goes here].

The ability to post ONLY to the local TL on the instance it originates from

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

WDC1:
Actually looks like this is something that has been opened  already Although it would be a good feature for all users not just Admins:
https://github.com/tootsuite/mastodon/issues/861

PR #8427, opened 2018-08-25
[WIP] Instance only statuses

### LINKED ISSUE #861: Add a "Local timeline" privacy option

renatolond:
This is a WIP PR for instance-only statuses. This is largely based on the glitch-soc implementation of instance-only to make it easier to incorporate back into glitch-soc.

This assumes that clients are still unaware of local-only statuses and to avoid leaking replies outside, if the local_only flag is missing, replies to local_only posts are posted as local_only.

As in the glitch-soc implementation, local_only statuses can only be seen by users that are logged in, logged out users do not see the local_only statuses in hashtag timelines or in the "take a look inside" one. local_only cannot be fetched by the other instances, even with the link. The default option is to federate the toot

Fixes #861, #896, #1761, #2698, #3285

What is missing:
- [ ] Refactor "local_only" into something else to avoid confusion with "local", maybe "instance_only"
- [ ] Change icons into something more meaningful
- [X] Add customization in preferences to which federation status is preferred.
- [x] Add "local_only" indication in moderations screens
- [x] Fix react privacydropdown warning, probably something is duplicated since the federationdropdown component is copied over from the privacy one.
- [x] Run i18n, add pt-br translations

image
image
image
image
image
image


Cassolotl:
>With this it becomes "unless you want ALL of their content, in which case you better sign up on the same server as them".

I guess the difference then is I'm cynical in that I think most people will want to blast their content to as many followers as possible and I think people will be lazy and ignore it if it's default-off, and I'm also trusting in that I think most people will post local-only *only when they're posting content that is truly not relevant to other servers* - like, "wow our instance is slow today, huh?" or "I think [new feature] on glitch.social could be cool, what do you think?"

I've also seen people setting up instances for particular towns, cities, universities and colleges. So if someone was a member of one of those, they might want to post "hey, bake sale in the student union foyer at 2pm today!" or "does anyone know where I can get a wide-screen TV near Main Street?" or whatever, without bothering people who would not be anywhere near those places.

I don't think Mastodon is diminished by allowing people to be insular. I feel like if you're going to promote Mastodon as being many small, safe communities focused on particular interests or geographical locations then it absolutely makes sense to allow people to post to only their own instance. I think this is a good idea, and I like having it on dev.glitch.social even though I have only used it a couple of times.

But Mastodon is already pretty insular compared to similar centralised services so I can see why you're concerned and not wanting it to be more insular.

trwnh:
Would it be possible to instead treat this as a targeting issue than a federation issue? If the aim is to have a certain post visible to users of an instance, then could there be a Collection of all the users on the instance that could be cc'ed, in the same way that the `#public` is cc'd on public posts? It sounds like it would be better long-term to support this as a specialized case of arbitrary audiences, rather than make an entirely new feature for basically the same concept.

nightpool:
my main concern is that this adds yet another layer of complexity when
explaining the privacy settings to users. people already make serious
mistakes because they don't understand the existing privacy levels (like
set their account to locked and use a third party client that doesn't
respect default privacy settings).

adding yet another checkbox doubles the layer of complexity.

I think the usecase here would be better served by having an
"announcements" feature for admins that gets stickied to the top of local
user's feeds and doesn't federate. this would be analogous to "pinned
posts" on Discourse or other forum stickies.

nightpool:
my main concern is that this adds yet another layer of complexity when
explaining the privacy settings to users. people already make serious
mistakes because they don't understand the existing privacy levels (like
set their account to locked and use a third party client that doesn't
respect default privacy settings).

adding yet another checkbox doubles the level of complexity.

I think the usecase here would be better served by having an
"announcements" feature for admins that gets stickied to the top of local
user's feeds and doesn't federate. this would be analogous to "pinned
posts" on Discourse or other forum stickies.

Cassolotl:
> I think the usecase here would be better served by having an "announcements" feature for admins that gets stickied to the top of local user's feeds and doesn't federate. this would be analogous to "pinned posts" on Discourse or other forum stickies.

I do agree that this is a good idea, but it wouldn't help the non-admins who want an instance-only visibility setting for their posts, I think it serves a different purpose.

renatolond:
> A linked chain to federate, a broken chain to break the link from the rest of the fediverse - so the broken chain of not federating has negative connotations and is a little discouraging.

I think the link/link broken is good, but I would like even more something like this one: https://www.iconfinder.com/icons/175933/connections_graph_network_relations_structure_icon
And one with the relations broken for the local only, I think it would convey better and avoid associations with links, but there's nothing like that in fontawesome, that's why I went for the link ones.

Cassolotl:
>I think the link/link broken is good, but I would like even more something like this one: https://www.iconfinder.com/icons/175933/connections_graph_network_relations_structure_icon

Ohh that's a great one. Maybe in time fontawesome would have something like that. :)

renatolond:
Local timelines are not private, they have all the public posts by local users and there are tools to browse them from outside the instance (I think at least one mobile app also implemented a similar function recently)

Local-only/instance-only posts would only be viewable, according to this implementation and glitch-soc one, to logged in users.

PaperFixie:
I am a member of multiple instances, one of which is purely for communication with my local community geographically and I think it would be beneficial to have local only posts for those of us who use another instance for socializing across the fediverse, but use a different one for talking to our literal neighbors.

melissamcewen:
I think this is beneficial and would improve my experience with Mastodon immensely. Also recently learned about a hostile instance where the admin may be using the DB to look at followers-only tweets (where the toot in question is on another instance, followers only, and the tooter follows/is followed by someone on the hostile instance. Instance only would be a barrier to such behavior. Especially for meta toots that have to deal with instance moderation this would be really nice.

apiontek:
I am in favor of this feature.

> imagine mastodon.social gets this feature, and suddenly signing up somewhere else is not an option because you'd be missing out!

Why would someone make effort to post local-only content that I, a user at some other instance, need to see?

Imagine users on mastodon.social sending DMs to other users -- am I missing out? Imagine users I want to follow who have privacy settings requiring approval. Imagine posts that are locked.

> Your posts go to your followers.

If I have a private account and approve follow requests, that means everyone who wants to follow me doesn't get to see all my content.

That's a restriction based on personal privacy consideration, which is good and important, but it doesn't serve the same purpose as instance-only toots would, which I feel is also a good and important option to have.

If for some reason I do an instance-only toot that seems so great it should go beyond my instance, I can always re-toot it, or someone on the local can always copy/paste it.

Much like private toots that one might later decide to make public, no one ever gets things perfect on the first go. It'd be nice to have this option available for use.

a-dows:
I run a geographically-oriented instance, and this would provide massive utility to us.

"Instance-only + followers" toots would help me make administrative announcements, and give peace-of-mind to users so they can feel comfortable posting classifieds or other personally-identifiable content, and in general facilitate the growth of a class of instance that is popping up all around the world.

I understand that it adds complexity, but if the feature is opt-in on an administrative level, it would improve my users' Mastodon experience tremendously.

FlamingOntheNet:
I'm a member of a regionally-focused instance, and there are many in the instance that would like a local_only feature. Personally, some posts are either more appropriate to the instance because they are very regionally-specific (meet-up, organizing, and local advocacy toots) or are related to trust issues from lessons learned from birdsite. I like to keep most of my toots in the federated timeline and my followers, but there are times when I simply want to avoid it but want it to be readable by the entire instance.

renatolond:
> "Instance-only + followers"

In this case you mean that the post doesn't leave your instance, but only goes to all the followers on your instance?

a-dows:
> In this case you mean that the post doesn't leave your instance, but only goes to all the followers on your instance?
Yes, 100%.

a-dows:
If this feature is implemented, my followers who are not on my instance won't be bothered by locals-only content that doesn't apply to them—my general-interest toots will reach them, while my administrative and locals-only content won't. This is a better experience for both poster AND follower.

> With this it becomes "unless you want ALL of their content, in which case you better sign up on the same server as them".

My followers have _permission_ to see my toots; they do not have the _right_ to see absolutely everything I post. With local-only posting as an option, my federated followers will still get 100% of my general-interest content.

giromide:
The privacy levels on Mastodon are already a bit tricky, and I don't think this makes them any harder to understand. I see more benefit in the comfort of knowing one's posts won't leave their instance at all than detriment of more privacy confusion. Plainly explaining the feature is enough. Geographic instance users will love this.

t54r4n1:
Instance only meetups are fun and I want to plan more of them, as a mod of a geographic instance please let me have a tool to do this!

rtxanson:
I'm not on a region-specific instance, but follow a ton of people on a region-specific instance because I live in that region. I'd be in favor of "instance only + followers" if "followers" could also mean followers on other instances. Alternatively, would there be a way for me to filter a timeline to only see instance-specific tweets, meaning I could potentially make an account for the sole purpose of making sure I'm not missing important stuff?

I can see how there's a difference between a need for admins and mods of instances to communicate information that is only relevant to people on the instance (server updates, downtime, etc.), but what to do about regionally relevant information for people who aren't on a regional instance? I guess it might be up to posters to determine if their target audience also includes a selection of people who might need to see that content too.

lawremipsum:
I admin mspsocial.net (170 registered accounts). Here's my case for local-only status privacy:

* _It's intuitive_. We have a local timeline, but currently no way to post to it without also posting to the federated timeline. My users ask how to do this frequently, and I have to explain they can't.
* _It gives users choice._ More granular privacy controls are better, as a general rule, and add value for users who choose to take advantage of them.

The lack of local-only compels people to share posts more broadly than they would like to serve a network purpose ("feed more content to the federated timeline"). However,
* Local-only will make the federated timeline _better_ and _more useful_ in most circumstances. The federated timeline on large instances is almost useless because it has _too much_ content. Do users on other instances really need/want to wade through posts about parking in Poughkeepsie, or the level of bus service in Boston, or the weather in Walla-Walla? Why shouldn't the author of those statuses have the discretion to spare the world things that are not of general interest?

Mastodon has largely grown out of the need to programmatically generate content for the federated timeline. Some of it could be spared everyone's eyeballs and nobody will feel the loss. And I have no doubt that a lot of stuff that could be local only will still be posted under the default world-readable setting.
* Adding the option puts the choice of whether to publish that information more widely in the hands of the author, not the code. Nobody would be *required* to post local-only. It would simply be an option so a user could avoid cluttering feeds with information unlikely to be of interest.

I'm neutral about whether it should be implemented as local-only or local-and-friends-only, but I suspect one or the other has deeper implications for the way federation works, so if that's the case the simpler implementation would be better than none at all.

trwnh:
From a lot of the recent comments it's sounding like the reason that this is seen as a complex expansion is in fact *because* of the way the current privacy system is set up, and not merely incidental or adjacent to it. Right now, "instance only" looks like it will have to create multiple new options instead of just one, in order to cover what is essentially "instance-only public" as several people seem to want, and "instance-only-followers-only" as @a-dows describes, as well as "instance-only public + followers-only" as @rtxanson describes, and so on. There's no simple way to cover all of these expectations in a concise manner using single-state privacy-as-a-radio-button.

What's *really* being done on the backend should be more akin to a set of booleans, where each audience is either allowed or disallowed from receiving the content. Instead of having the privacy dropdown double in length every time a new privacy feature is added, it would be much better to add one toggle per feature/audience instead, or perhaps as a multi-selector field with a "Deliver to..." field (i.e. `to:`) and a "Visible to..." field (i.e. `cc:`). As it stands right now, there isn't a clear 1:1 mapping of what each privacy option is called semantically and what it's supposed to do technically; people get confused about e.g. what "unlisted" means in exact terms, and whether hashtags count as a public timeline, etc.

I still think that Mastodon should support fully-custom audiences in the future, but the groundwork for a privacy revamp should be done first or else it will get even more confusing. See also: #5686, where a "trusted followers" feature is requested to allow locking your account but still letting anyone receive your public posts. #7182 gets into this conundrum as well, although IMO it confuses lists with audiences.

a-dows:
> There's no simple way to cover all of these expectations in a concise manner using single-state privacy-as-a-radio-button.

That's true—but that's also not what's being proposed. There's already a privacy button (public, unlisted, followers only, direct), but this feature would implement an additional _scope_ button—local or federated.

I'm confident that the combination of **privacy + scope** covers all the mentioned possibilities. As for @rtxanson, I think creating a private alt account on their region's instance is the right idea, and doesn't require a new feature or change.

trwnh:
@a-dows 
>I'm confident that the combination of privacy + scope covers all the mentioned possibilities.

Given that privacy and scope are near-synonyms, I don't think that matters too much -- after all, this PR is only adding one more toggle button, it just happens to be separate from the dropdown menu (which in itself represents multiple hidden toggles).

annalee:
As an admin for a small, interest-focused instance (wandering.shop), I'd find local-only posting immensely helpful, and I believe my members would, as well. There are certainly other ways to give admins the ability to make announcements, but there are plenty of other use-cases here that may not be apparent to users of huge instances like .social.

Small instances--especially region- or topic-focused ones--often have a local "culture" that they interact with differently than they interact with the fediverse. We have a sense of our local instance as a neighborhood. And some conversations make sense as neighborhood-only.

I get the concern about people suddenly feeling like they're missing out on conversations from folks they're following. But right now we have a situation where folks feel compelled to either include unwanted parties in conversations, or else not hold those conversations on Mastodon. We're choosing folks' desire to see others' content over folks' desire to choose who can see their content.

To be a bit extreme about it, we're currently choosing "don't go out in public if you don't want people to talk to you" over "people who are out in public deserve privacy and personal space." Providing tools that empower users to set their own level of visibility is _immensely_ helpful to those of us who have to deal with harassment online. When your choice is no visibility or all the visibility, and you have to weigh how much harassment all the visibility is gonna get you, you often choose no visibility. You choose silence, or locked posts. (Or walled gardens like slack). Intermediate options help foster increased engagement, by allowing users to choose the audience that makes them feel safe enough to speak.

Letting people set clear boundaries doesn't strike me as confusing. I'd think most users would rather to be told up front that they're not invited to a particular discussion, rather than finding out after the fact that they've annoyed somebody by barging in where they weren't wanted.

saper:
This is about fundamental question whether instances are something like "groups" or whether they just provide decentralization and robustness against broken servers and malicious admins. I am still not sure that an instance creates a "community". 

When I came to the Internet I was amazed to be able to poke into some net life on other continents, learning a lot for whatever I would like to be doing locally. 

Local instance posts prevent me from running my own little instance (I do) and be able to meaningfully participate or just listen. I find it important to be able to read lots of totally unrelated stuff that is not addressed to be on a federated timeline. Just be the fact of being exposed to other languages, communities thousands miles away makes me a richer and better person. 

Even local admin announcements should be read outside - for example for the users having trouble accessing it, or for freaks like myself who want to tune into the general health state of the federated network. 

Gated communities full of local-only posts will become similar to closed fora that annoy you with their registration requirement. They will grow into isolated silos with time.

Don't be afraid - people not interested will not listen. There is enough stuff online. Why bother with people who are not part of the community of interest?

If you have something to hideinstead, I am afraid Mastodon is not the tool for you.

saper:
Thank you @renatolond - I think I made my point too short regarding "nothing to hide". In [a Mastodon conversation has been pointed out](https://tootplanet.space/@Heliodora/100633818243609584), that we already hide things, like DMs and follower-only posts. So this is not a fair argument from my side.

I think it is not good to tie community to the technical instance instead of a follower-followed graph. In the previous era this would mean, "to join our community you have to have aol.com email address".

Sure, there were walled communities like this, but is this what we want here?

I am observing some small, very tightly-knit instances with a very specific (and sometimes sensitive) stuff. I agree that local timeline is a pretty good tool (except you don't  sadly get replies...) for those. But even there one can have few external valuable contributors who should need not to be forced to register locally. And to avoid the "well, actually" it might be the best to lock the account and post mostly followers-only as many do in those instances.

And I think "We are going down" is that kind of announcement that is very useful when propagated to the rest of the network.

MxFraud:
As an admin of a small-ish instance (500 registered, 200 active in the last 3 days) currently using the glitch-soc fork with local only toots, I'd like to offer a perspective.

We federate with 1529 instances.
The peering is growing at the same rate as the number of user and the number of posts.
Here is the data: https://fediverse.network/tabletop.social/stats
The point is that local toot didn't isolate us from the fediverse.

There are a few communities around the boardgame/tabletop hobby.
There are dice.camp and boardgames.social to name two popular community running tootsuite mastodon (so not glitch-soc).
New people often create an account all of those community.
Some told me specifically that they just want to look into the local timeline.
They have a main and maybe 3 alts on all of those boardgame/tabletop hobby community.
That is to say that people are already creating account on multiple community to look at the local timeline on community with no option for local toot.

There are also people that only want local toot.
On my community there are people doing only and just local toot.
I have no visibility on DM as I have no access to the database, but I would assume that they DM people outside of my community.
I didn't check who they follow, but I would be surprise if they didn't follow people from outside the community too.
They also reply to otherwise public and federated toot, adding the restriction to local toots.
(I didn't question their motives, so I won't assume anything.)
That is to say that, providing more choices we enable people with different need (potentially having more different people interested in joining in the federation).

From my community's view, I can say that the usefulness of the "local toot" feature is not up for debate.
(I would like to stress that I understand and fully support that the decision to include this feature in software is up to tootsuite leadership.)

apiontek:
>Local instance posts prevent me from running my own little instance (I do) and be able to meaningfully participate or just listen. I find it important to be able to read lots of totally unrelated stuff that is not addressed to be on a federated timeline.

This doesn't make sense. Everyone generally wants to participate in the wider network. The default would still be be federating.

In terms of some toot you might not see -- you already don't know what you're missing, how would things change?

If you think someone's local-only toot should be disseminated to the whole network, how did you see it in the first place? How does that work with DMs, currently? If a private toot happens to get to you, what if you just reached out to someone and asked them to re-toot publicly?

It only makes sense if you have no faith in the existing network. If you think that *everyone* only federates because they have no other option.

Siloed fora already exist. Yet people here chose to participate in a federated network. They will continue to do so, because most people feel like you do.

>Sure, there were walled communities like this, but is this what we want here?

That's not what this is about. I don't see anyone asking for walled communities. I think if you fear a slippery slope you need to make an argument for how and why the entire fediverse would slip down that slope.

My counterexample: reddit allows locked subreddits. Yet the majority are still open, because most people see the value in that; and yet, communities that feel the need for that privacy have the option.

And again, this isn't even about locking down / disabling federation. It's just a toot scope option.

codesections:
I've been giving this a lot of thought and have come to the conclusion that this change should be made—but only **after** another change.  Specifically, I think Mastodon should *first* allow users to have other local timelines columns in the UI (and to pin those columns first, if they wish) and *then* allow local-only posts.  Here's my reasoning:

*  Right now, if we implemented local-only posts, we'd have them in a gray area: they're public—including to people outside the instance—in that anyone *can* go to the instance homepage and see them.  But they're also hard to access—most people don't bother browsing the local timeline of other instances because it's hard to do.
*  Having posts in that privacy gray area seems **really bad**.  If local-only posts are limited to people who have registered on the instance 95% of the time but, because they're publicly available, are seen by outsiders 5% of the time—well, then it gets really easy to forget about that 5% and make a mistake.
*  As others have noted, local-only posts would be helpful for posts that aren't private but that are of local-only interest (e.g., a local bake sale).
*  However, as others have also noticed, it's possible to be interested in the local goings-on of multiple communities (a student might want to follow local news in their hometown and their college town, for example.  If we expand the discussion to non-geography focused instances, the case for following multiple instances gets even stronger.)
*  If we allow local-only posts *and* let non-members pin the local timeline, we get the best of both worlds.  Someone who lives in Minneapolis could join an instance of their choice and still pin the local timeline of the mspsocial.net instance to follow local events.  People who post to the mspsocial.net local-only timeline would also not be confused about privacy—it would clearly be a post that doesn't go to all followers but that is easy to see for anyone who is interested in local events.

In sum, we *should* have local-only posts, but only *after* we have the ability to view and pin local timelines in the web UI.

jeroenpraat:
I share the concerns of @Gargron, but I have another solution: groups.

Groups, something that @Gargron once mentioned that it was something he wanted to work on. If you develop a groups feature, and you add an admin option to automatically add new users to a group (compare with the current auto-follow), you have ~ the same as local only.

As a group I see a special user that lives somewhere on an instance that people can follow, just like a normal user. Sending a toot to that special (group) user, also sends it to the group followers (as a sort of relay). Toots to groups can have the same kind of visibility levels as toots to normal users, but a group-user can override those privacy settings with a max. visibility level (all with the existing visibility levels). A group-user can also lock its account, to be able to approve group followers. If [this issue](https://github.com/tootsuite/mastodon/issues/8352) is resolved, an admin can also add new users to a locked group. A group profile looks (almost) the same as a normal user profile, but with a UI option to pin the column like hashtags.

renatolond:
> they're public—including to people outside the instance—in that anyone can go to the instance homepage and see them. But they're also hard to access—most people don't bother browsing the local timeline of other instances because it's hard to do.

This is not true, local-only posts are *only* available to logged in users. If you navigate to the local timeline of an instance with local-only posts without being logged in you only see their federated posts.

renatolond:
> I share the concerns of Gargron, but I have another solution: groups.
> Groups, something that Gargron once mentioned that it was something he wanted to work on. If you develop a groups feature, and you add an admin option to automatically add new users to a group (compare with the current auto-follow), you have ~ the same as local only.

This is very similar to what trwnh suggested in the beginning, and for me the answer is the same:

> I think it could be the case, but I don't think adding the cc'ed features for groups with meta-groups like local is such an easy change, and if we integrate this into mastodon and later change this into groups, I believe this could be changed in a migration to work with such a feature.

jeroenpraat:
@renatolond I appreciate your PR, but IMO it's really better to develop a group feature, cause everyone can benefit from it.

codesections:
> This is not true, local-only posts are only available to logged in users. If you navigate to the local timeline of an instance with local-only posts without being logged in you only see their federated posts.

Ok, true.  But they're visible to logged in users *from any federated instance* (unless I'm misunderstanding the implementation).  This means 1) that they're trivially accessible to anyone (so long as at least one federated instance has open registrations) and 2) that letting people pin the local timelines of other instances wouldn't represent a *change* in the privacy status of a post—local-only posts would still be visible to logged-in users of other instances.  It would just make opting in to viewing these posts easier for logged-in users of other instances.

renatolond:
No, sorry if I wasn't clear, local-only are only visible to user logged in the instance they were created. As in: if I create a local-only post on social.lond.com.br and you are not logged in that instance, you cannot see my local-only post, even with the privacy setting in "public"

codesections:
Oh, then I did misunderstand the implementation in the PR.  Thanks for clearing that up!

Now that I understand the implementation, I think that the current implementation should not be merged.  Consider a few scenarios:

*  A user on mspsocial.net enrolls in college somewhere else that has its own local instance.  They don't want to give up the mspsocial local timeline, but also want to get the "hey, bake sale in the student union foyer at 2pm today!" posts on the new local timeline.  Result: they are forced to create two accounts.
*  A user creates an account on fosstodon.org because they are interested in FOSS, but later discovers that their town has a geographic instance.  They want the geographic instance's local toots.  Result: they either move instances (to one that doesn't soot them as well, topically) or create two accounts.

I think there'd be lots of scenarios like this.  If local toots get used at all, it seems like they'd create significant pressure for people to create accounts on multiple instances.  I'd prefer to avoid that.

Thus, I'd oppose the current implementation, though I could support a tweaked implementation that avoids creating pressure for multiple accounts.

melissamcewen:
I think people are thinking about this too broadly. I don't see the bake sale thing being a use case. Why would someone local-only a post about a bake sale? You want as many people to come to your bake sale as possible. Unless they have safety concerns, they would likely make it fediverse-wide. 

When people make things have less of an audience, they have reasons, and we should respect that- not worry it gives us less content to consume.

apiontek:
I think local-only is distinct and important. Groups may also be a good idea, but it's not as related as people think. Groups are useful on Facebook and it makes sense to me that something like that might come to Mastodon.

But birdsite doesn't have groups, and so clubs & groups create separate accounts. Unless groups become a thing, I'm not sure why that wouldn't be done on Mastodon. If I go to a college and I want announcements from the college, from the student council, from the environmental club -- I'm going to follow the accounts for the college, the student council, and the environmental club.

But setting that aside, local-only is a community privacy issue. An instance community may want to have discussions that don't federate beyond the instance. It seems pretty clear-cut beneficial as a non-default option.

If you feel it will be misused for these other purposes, creating multi-account pressures, then maybe some other feature like "Groups" for those other purposes is needed, but it seems separate to me, and again, people already create targeted accounts in social media.

saper:
@apiontek but this is a valid concern - participating in multiple instances that use local only posts forces the user to create multiple accounts on the respective instances. This creates the problem that federation was created to solve in the first place.

apiontek:
@saper I disagree. I think you're over-thinking what local-only toots would be used for. Why would you want to be in on an instance's private communication? Do you think they don't already create separate fora for said private communication, now, because there isn't a local-only option? You're already missing out on that.

In other words, the absence of local-only option forces communication that could be done on-instance into entirely different platforms.

I do think there's a great case to be made for "Groups" in Mastodon. For example, I follow some people on the bsd instance but I didn't join because I don't need an account there. But it's also not a discussion forum. The people there don't only talk about BSD; the idea of a topic-based instance seems almost baroque to me. Isn't that what hashtags are for? If I want to see what people are saying about BSD or linux, I'm going to check out those hashtags or do a search, but I won't find anyone people on my instance don't follow.

Groups could solve that issue.

But local-only toots seems a completely separate issue to me. If the concern is people misusing local-only to approximate something like groups, that seems reasonable to me.

sireebob:
tl;dr: if you'd rather people not use it a certain way, warn them first, but let them do it.

people settle into instances partly based on the convenient access to desirable content that comes through their local timeline, even though everything that shows up on the local timeline is technically public. i know i spent my first months on mastodon not even looking at my home timeline, just local, because (not to speculate on the reasons), it ended up being overall more relevant and interesting.

is this in itself a problem? if so, it's already a problem, and a separate one to solve, possibly with the aid of some added ui/pane flexibility. (being able to include local timeline in a pane alongside list(s) or even home timeline content sounds nice, but that may not be sufficient and again, that's separate.)

i think it's reasonable not to *want* people to decide to use the local-only feature if we assume they're doing so solely because they can't imagine how outside users would find their posts relevant, or because - after spending a lot of time looking at the local timeline - they've decided they're satisfied enough to share toots solely with the locals.

still, i think it's important that people *can* post to local only at their total discretion, as they may have legitimate reasons (such as safety or trust concerns, or need-to-know posting) to do so. followers-only is not a realistic substitute for this, given the inherent lack of security or privacy *guarantees* in federation.

if you really don't want people to do it, maybe warn them of the implications the first time they switch to it and confirm it's what they want (or every time, with a "do not ask me again") - heck, add a sad or lonely mastodon graphic to the dialog if you want - but they ought to be able to do it. (i know such a graphic would be enough to dissuade me from using it unless i felt i had an important enough reason.)

regardless, i think people will be able to influence each other to make the right choices. see existing pattern: someone says "i wish i could boost this" and, if acceptable, person makes a less private re-post.

sireebob:
this also seems like a useful feature for that "Mastodon for the organization" concept highlighted at joinmastodon.org

dariusk:
@renatolond Testing this out on a small instance (https://friend.camp) and it looks like local-only posts still appear in the profile page of the account that posted them. For example, you can go to https://friend.camp/@darius and scroll to see some of my local only posts like this one:

image

You CANNOT link to the individual post and see it ( for example, [this post](https://friend.camp/@darius/100648101694102395) doesn't render unless you are logged in to my instance, but you can still see it on the profile.

dariusk:
@renatolond FYI I *think* this commit fixes the issue: https://github.com/masto-donte-com-br/mastodon/commit/b63529e7cc02b26ebf7340f8d8cf3a4e51297854

trwnh:
@jeroenpraat 
>If you develop a groups feature, and you add an admin option to automatically add new users to a group (compare with the current auto-follow), you have ~ the same as local only.
@renatolond 
>This is very similar to what trwnh suggested in the beginning, and for me the answer is the same: I think it could be the case, but I don't think adding [it] is such an easy change [...] I believe this could be changed [later].

Worth noting that a group feature and an audience feature are similar but with a key distinction: groups can be joined, audiences cannot. I suppose since there is a `Group` `Actor`, posting to a `Group` would require being able to post to arbitrary `Collection` of `Actor`s, some of which may very well be `Group`s. But it's still possible to make group posts a separate feature and user-flow unrelated to audience or privacy.

However, it's this bit that has me more concerned:

@renatolond 
> local-only are only visible to users logged in the instance they were created. As in: if I create a local-only post on social.lond.com.br and you are not logged in that instance, you cannot see my local-only post, even with the privacy setting in "public"

The requirement of being logged in to view what is otherwise a "public" post is very confusing. At this point, the post is not actually part of the public web; it is part of the deep web. It brings up the possibility of many sticky situations -- for example, how will you treat local-only posts that mention someone on another server? Can you create local-only replies to remote public posts? What happens if you send a direct post with local-only to a remote user?

Also, how does that affect loading posts over the API? Since they won't federate, I assume the ActivityPub addressing is irrelevant (since Mastodon does not use ActivityPub internally, only for federation). The same cannot be said for the API -- if I GET /api/v1/timelines/public?local=true, will these 'local-only' posts be returned?

kevinflo:
Just want to chime in using mspsocial.net (very awesome instance!) as an example of why I'm currently not a fan of the idea of instance-only statuses. I actually live in the twin cities, and was really happy to discover an instance dedicated to this region. Thing is, I already have an account at mastodon.social and have invested the time there to start constructing a social graph there and I've put links elsewhere on the web to let people know my mastodon.social account is where to follow me.

I absolutely want to see toots and content from mspsocial.net, but if local-only toots were a thing I think lots of people there would toot things local-only rightly thinking they're more relevant to local people etc. Thing is... I *am* local people. My account being on the mastodon.social instance and not the mspsocial.net instance is entirely the result of when I found out about mspsocial.net existing and not a reflection of whether or not I want to participate in that community. In other words, I want to be able to be every bit as much a member of that community from mastodon.social as I would be from mspsocial.net. I definitely do not want to create a whole new mastodon account just so I can fully participate. This is much in the same way as I would not want to create a whole separate email address just to be able to communicate with certain people via email.

egypturnash:
@kevinflo If we ever get account migration nailed down, then most of your concerns vanish.

Personally I like it. Not everything needs to be broadcast to the world. Memo to self: consider installing this patch on dragon.style.

(Or beg toot.planet for their implementation, which just adds a new entry to the existing privacy menu for “public, but only on the instance” with a very cute icon choice that matches their branding.)

saper:
@egypturnash I am not sure account migration would help. I, for example, participate in local online communities in many countries I have had opportunity to live or work. So I tend to be local to cities like Belgrade, or Kiev. I wouldn't want to move my social presence everytime I find something interesting in some new city. I want to use my account (like mastodon.social or my own instance) to access local conversations of locals in Serbia and the Ukraine. I don't want to be forced to migrate every time I find another interesting instance.

dariusk:
I'm just gonna put this out there: my users love the local-only feature so far. Like, they adore it. They are talking more than ever before and it's precisely because they know actors outside of [Friend Camp](https://friend.camp) (which is an invite-only server built on pre-existing webs of trust) can't see it.

I don't think local-only posts add much of anything to large generalist instances but to add to the chorus here: for Mastodon instances where it's effectively a small niche community with an added bonus of federation in case you want to talk to the outside world, it's so far really great.

I plan to continue updating this thread with further observations of how my small community engages with it because it's important to have data on how real users are using the feature rather than only throwing out hypotheticals.

dariusk:
Update from my instance! This will sound like it's off-topic at first but I bring it back to instance-only statuses at the end.

One of my users "soft blocked" (blocked and unblocked) about 30 people from their locked account, which has the effect of forcing the remote user to unfollow them. But the next day one of those remote users they soft blocked was still listed as following them. Now, my user's local UI on the web interface indicated that they were NOT being followed by that remote user. But when I checked the remote user's profile (at mastodon.social) my user's locked account was right there in the following list.

I don't know if this was a race condition or some sort of database syncing issue or what. I am unable to reproduce it personally, otherwise I'd file a bug against it.

**Why this has to do with instance-only posts**

Having implemented my own ActivityPub server, I can see these problems happening incredibly easily. Like I get an "Undo Follow" request, I send a confirmation that I undid the Follow, but maybe my server rebooted at just the wrong time and the unfollow never happened. And of course that's not saying anything about what a malicious actor can do (I doubt that mastodon.social was running malicious code so we can strike that off this particular occurrence).

These kinds of sync/race conditions are far, far, far less likely to happen between users on the same instance as one another. I am considering telling my users that they should ONLY use instance-only posts for any kind of information that they want reasonable (by social network standards) security around.

I think it's a disservice to not offer some kind of instance-only posting to users. It's a needed level of security.

(And like I could say "oh well if it's something that sensitive don't use Mastodon for it" but this user was simply making mundane complaints on a locked account that they didn't want a stranger to see just because that stranger had previously followed them at one point.)

renatolond:
@dariusk I didn't use up the same in the end, I did this: https://github.com/tootsuite/mastodon/pull/8427/commits/8894ccb84b194209af631cc706fbb93c9824068c

renatolond:
> how will you treat local-only posts that mention someone on another server? 
Right now that's quite in the open. The code for glitch-soc, if I'm not mistaken, still notifies remote users.

> Can you create local-only replies to remote public posts? 
In theory, yes. And I believe that if you don't mention the original user, it could be a local-only reply

> What happens if you send a direct post with local-only to a remote user? And so on.
Same as the mention, at the moment

> if I GET /api/v1/timelines/public?local=true, will these 'local-only' posts be returned?
No, unlogged access to the API won't return those posts

codesections:
Something recently crystallized for me, and I want to share it here in case it helps the discussion.

> I'm just gonna put this out there: my users love the local-only feature so far. Like, they adore it.

> Just want to chime in using mspsocial.net (very awesome instance!) as an example of why I'm currently not a fan of the idea of instance-only statuses. I actually live in the twin cities, and was really happy to discover an instance dedicated to this region. Thing is, I already have an account at mastodon.social

I agree with @dariusk that there are users who would *really like* this feature—in fact, I'd be one of them.  If we had local-only toots, I could see using the feature a lot.

I also agree with @kevinflo that there are users who would really be hurt by this change—people who would like to follow local events in an instance without migrating their account to that instance (or managing multiple accounts).  This could be people like @kevinflo who have an account on a generalist instance but want to follow local events for their region.  It could also be people who want to follow multiple local/special interest instances.  (Just the other day, I was talking to someone who was feeling torn between fosstodon.org and mastodon.art—and that's without any local-only posts.  If both instances had a few local-only posts, that effect would only get worse.)

So, when evaluating this issue, we have to make a trade off.  We either have to prioritize the interests of people who want to make local-only toots, or we need to prioritize the interests of people who want to read toots of local interest in other instances.  It's a tough trade off to make and strong arguments can be (and have been!) made on both sides.

In the end, I come down against this change because I think it would exacerbate one of the biggest problems that's stopping Mastodon from growing.  Right now, one of the biggest problems new users face is picking an instance (the tools are getting better, but it's still really hard).  A big part of our response has been to tell users "Yes, use the tools to pick an instance that looks good.  But it doesn't *really* matter: regardless of what instance you pick, you can still follow people from other instances."  That's already only half true—as people have pointed out, local timelines are great!—but this change would make it even less true.  This change would make it even more important to pick the right instance, which would put up a stronger barrier to joining.

That, in combination with the many people like @kevinflo who would be harmed by this change, strikes me as good enough reason to oppose this change.

sireebob:
seems like some people want:
1) a visibility option that works exactly like unlisted *except* it shows up in the local timeline. in this situation, the content could still be found and seen on home timelines.

while others want (subject of this issue):
2) toots that can't leave the instance and are only visible by other instance users.

we keep acting like people who want 1 really want 2. i think the desire for 2 is real and separate, and if given another option, those who really want 1 wouldn't use 2.

if neither of these issues has support, well... honestly, based on comments here, it sounds like the local timeline is inconsistent with the intentions of mastodon and should be removed. people would hate that, though (self included), so something else would need to replace it.

steckerhalter:
yeah well, it looks like @Gargron is against this anyway, so why bother to try to get it into mainline?
personally I'd like to be able to address only users of my server. there are many use cases for that. the argument that it fosters centralization dynamics is not very good IMO, rather it gives you more options, more freedom for the user and the community, which I think is good.
but we hit on the same issue here why Eugen doesn't want to have boosts with quotes. it would be a very useful feature to make for more interesting discussions. instead of just boosting stuff, people could engage more and add their POV. again more freedom. but Eugen says that it fosters abuse in this case. as someone pointed out you can harass people just as well without it. but you can't easily add your two cents to a toot without it.

dariusk:
Update: we just had an instance-wide movie night 
consisting entirely of instance only posts, consumed through our local timeline without cluttering up other instances or having to hashtag or cw so people could mute who don't want to see it. We like this.

dariusk:
For anyone who's merged this, [here's a commit](https://github.com/friendcamp/mastodon/commit/3da78528845d376ca4d007703d5da3bc0805aff6) that is a little hack using LocalStorage to "remember" your last manual federation/local setting in the web client.

Ideally the `default_federation` state (currently hard-coded to `true`) would get its own DB entry and setting in the settings page but I don't really know how to touch the serverside stuff without breaking anything. So this is my frontend-only hack.

JBaczuk:
> Update: we just had an instance-wide movie night
> consisting entirely of instance only posts, consumed through our local timeline without cluttering up other instances or having to hashtag or cw so people could mute who don't want to see it. We like this.

@dariusk Would love to know how to get it running (non-rails expert). This is a privacy feature in my opinion, and I think it would be valuable to document so others can utilize it on their own instance.

I tried bundle install, then restarting the services and now I'm getting:

`[24113e7e-07c8-4ca6-a2a1-67d3e36ff365] ActiveRecord::StatementInvalid (PG::UndefinedTable: ERROR:  relation "status_stats" does not exist`

Do I need to run something like `rake db:migrate` to update the db relations?

dariusk:
@jbaczuk Yes, you'll need to migrate since it adds database fields for local/federated. That should be it though!

JBaczuk:
> @JBaczuk Yes, you'll need to migrate since it adds database fields for local/federated. That should be it though!

Thanks! Ok, I think what got it working was: 
```
$ bundle install --deployment
$ RAILS_ENV=production bundle exec rake assets:precompile
$ RAILS_ENV=production bundle exec rake db:migrate
```

The use case for me is I'm trying to pitch to friends and family that mastodon is decentralized and private. If the standard approach is all or nothing, then that doesn't really afford privacy. Sometimes we want to just post to the group inside our instance.  Currently the only way to do that is not allow any followers who are on another instance.

dariusk:
@renatolond Thanks for your continued commits to this! I'm going to merge them soon since a default federation option is one of the things my users have been asking for.

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -33,6 +33,7 @@ def user_settings_params
       'setting_default_privacy' => source_params.fetch(:privacy, @account.user.setting_default_privacy),
       'setting_default_sensitive' => source_params.fetch(:sensitive, @account.user.setting_default_sensitive),
       'setting_default_language' => source_params.fetch(:language, @account.user.setting_default_language),
+      'setting_default_federation' => source_params.fetch(:sensitive, @account.user.setting_default_federation),
```
dariusk: This should fetch `:federation`, not `:sensitive`, I had to fix this in my branch.

config/locales/simple_form.en.yml
```diff
@@ -65,6 +65,7 @@ en:
         phrase: Keyword or phrase
         setting_auto_play_gif: Auto-play animated GIFs
         setting_boost_modal: Show confirmation dialog before boosting
+        setting_default_federation: Always allow my toots to reach other instances
```
dariusk: You might want to change this wording (and in the pt-BR translation too). To me this means my toots will ALWAYS reach other instances even if I set it to local-only. I suggest `Allow my toots to reach other instances by default`.

dariusk (changes requested):
Found one bug I had to fix, and one wording change suggestion.

app/controllers/api/v1/accounts/credentials_controller.rb
saper: Regardless of what I think about this change, this setting is strange and hard to understand.

Is there a possibility to have this feature turned off for the instance (hopefully by default)?

lawremipsum:
I want/need this feature, and my users want it too. I'd love to hear whether patching it into my instance is likely to play nicely with future upgrades on the main branch...

renatolond:
@lawremipsum I'm probably keeping this branch up-to-date since I'm testing it on two instances I admin. But have in mind that's not finished, and might still have issues :)

wbrawner:
Hey, just wanted to chime in on this for another perspective. I've set up a self-hosted mastodon instance for my family, and I'd love to be able to post things just to my family, but still also be able to post things to the rest of the fediverse occasionally, so to kind of emulate a private group if you will. I realize I could just tag only my family members and limit the toot's publicity but then of course they'd get notifications for every time I post, which isn't exactly a great user experience. I understand the concerns about centralization but I don't think centralization is _always_ a bad thing. As mentioned in the comments above, I think having the option to enable/disable local-only toots and leave it off by default is a sensible choice. Above all though, I think the important thing is to allow users to make that choice for themselves, in this case users being site admins. While writing this I've realized that I think private groups would suffice, so that a toot could be limited to only people within a certain group and not send notifications unless people are explicitly tagged in the toot. Ideally I just need _some_ easy way to communicate things directly to my family without it going to the rest of the internet.

config/locales/simple_form.en.yml
renatolond: Yeah, I ended up using the same wordind as the other `setting_default_*`, like the sensitive one that says "always mark media as sensitive". I do think your wording is better, I'll change it. Maybe the other one should be changed too :)

bgiesing:
I've talked to a lot of people looking to join Mastodon cause the Tumblr situation but they won't because Local-only isn't an option.

They want to post stuff that they would be too embarrassed to post if the wider public across the whole fediverse would be able to see it and use it against them but at the same time, they don't want to make the posts Private or Unlisted since nobody in their instance (which is geared toward the exact thing they would be embarrassed to post) could see it then.

I think the best solution for "local-only" posts is if it was "Local-only" for just the Public timelines. People in other instances could still see the post if they followed you or had the direct link. That way it's a mix of Public and Unlisted, Public for your local instance, Unlisted for everyone else.

renatolond:
Merged v2.7.0rc1 into this :)

kevinflo:
> They want to post stuff that they would be too embarrassed to post if the wider public across the whole fediverse would be able to see it and use it against them but at the same time, they don't want to make the posts Private or Unlisted since nobody in their instance (which is geared toward the exact thing they would be embarrassed to post) could see it then.

I think it's dangerous to bake into the application something that reinforces/teaches users that this is even technically possible. Any semblance of privacy or constrained visibility in a one-to-many communication system is misleading, and encourages risky behavior.

I think this is doubly important in an application like mastodon, which is a kind of living classroom for newer/younger users who may be digesting concepts like federation through experiencing them on a mastodon instance. To make it appear as though their communication could be restricted to their instance (and make them feel safe behaving as such) when any other user on that instance is a living API into it strikes me as a missed opportunity. Many may understand that private in a one-to-many system isn't really _private_. Many surely will not and will act accordingly. If that happens, the blame cannot be put on them for the application inaccurately/misleadingly/irresponsibly classifying & labeling public communication as private.

dariusk:
@kevinflo By that logic, would say direct messages are a user interface flaw because the people on the receiving end could simply copy/paste and repost?

I admit that my instance is a little different from the one you're quoting, in that it is an instance based on mutual trust rather than the detente of mutual embarassment over the same topic. In my case, posting local-only to our instance is about as secure as posting to private group chat, and everyone knows it, and I take the time as admin/moderator to onboard every single user with this knowledge.

I think this kind of feature makes perfect sense for an invite-only, web-of-trust based server like mine. What I might want to see with this is to make it a feature that admins can enable or disabled on the server (similar to how fulltext search is handled). Because it makes sense as a feature on some servers, and doesn't make sense as a feature on others. And like fulltext search, it's going to be a security/safety risk for some servers and not a security/safety risk for others.

dariusk:
@saper We interact very often with people on other instances. What ends up happening is we post far more often on average, in fact. I was looking at the stats the other day. Our active users post on average 44 times per week, whereas other instances seem to have 5 to 20 per week per active user. [Here's a spreadsheet with some data I collected.](https://docs.google.com/spreadsheets/d/17b753o8qlpQHmATvNIFkT61X1OTJLuQvkW41uiYb3pc/edit?usp=drivesdk) 

It seems that we interact with other instances like normal, and our local only interactions are _additive_.

codesections:
@dariusk that's very interesting data—thanks for sharing.  How did you go about collecting it?  I'd be interested in collecting a broader sample if there's a way to automate it.

(By the way, you have minor error in cell E14 that makes mstdn.jp look like quite the outlier)

dariusk:
@codesections Most instances publish this data at a url like

https://friend.camp/api/v1/instance/activity

I get the total users from the about page. 

And I fixed the error, thanks!

lawremipsum:
Writing again to say that I run an instance that federates well, has users that interact both within the community and with the wider fediverse, and will benefit from local only posts. I'll be merging this into my fork.

I reject the idea that access controls are a harmful concept, as has the Mastodon project, otherwise it wouldn't have DMs, unlisted toots, or locked accounts. Maybe 60 years ago someone argued that the concept was dangerous, yet Unix empowers users turn off the world-readable file permission even though a bad actor could still share the contents.

renatolond:
Merged 2.8.0 into this, thanks @dariusk for the reminder!

renatolond:
Merged 2.8.4 into this :)

PopeRigby:
Bump

renatolond:
Merged v3.1.1 into it :)

frainz-de:
I would still like this feature

blaine:
Just bumping this - I want to add that most of the concerns here seem to be around forcing people to be on a single instance, and breaking the federated model. I want to add two things that haven't been discussed as possibilities on this issue as far as I can tell:

1. A clear use-case for this feature beyond small niche networks is intra-company networks; if I wanted to set up a mastodon instance for my company, I would need to not federate, since clearly the goal would be to share private or commercially sensitive content. However, with local-only posts, it would be possible for employees to share some things publicly, but most things privately.

2. If people are concerned with that loss of federation means that multiple accounts need to be maintained to access multiple networks, set up the activitypub equivalent of email forwarding. There's nothing in federation model that prevents this (indeed, it was a core design goal when we created webfinger). Where forwarding doesn't work (e.g., if you want to create a "local" community that is composed of people who are distributed across a variety of mastodon instances), the "Groups" model that has been mentioned in the comments on this issue and works for virtually all chat applications would be a logical extension of this behaviour.

There are plenty of legal and commercial reasons to implement this, and I'm happy to discuss those if it's relevant or interesting. I for one firmly believe that the Fediverse can only be truly successful if it manages to dislodge the incumbents' control over online sociality. The only way that can happen is if it's both possible and supported for the fediverse to be used in commercial contexts, and I sincerely hope that's coming. I'd rather a world without capitalism, but honestly I don't think that's Mastodon's fight to win.

If there's a better way to implement this sort of site-specific functionality, possibly through some plugin infrastructure or similar, I'd be keen to explore that. I'm not up to date with the current state of mastodon so it's entirely possible that I've missed something, but I need this or similar functionality and would prefer to not diverge massively from mainline mastodon, precisely for the reason of continuing support for federation! It'd be easy to run a non-federating instance to achieve my goals, but the downside is immense.

GerryT11:
I am in favor of this feature, for these reasons here: https://github.com/tootsuite/mastodon/issues/861#issuecomment-726247196

I agree with @Gargron that it doesn't make sense for general-purpose instances (e.g. mastodon.social) and would be counterproductive there, but it would allow local communities (in my case a local sports club) that have internal/irrelevant stuff and public toots. Thus, why not having this feature (built-in or as add-on) and disabling it by default,  but the admin of an instance can switch it on?

renatolond:
I've merged the latest tag on this :)

grahhnt:
This feature would still be a nice feature to add to Mastodon

unitof:
For those who aren't aware: this feature has been implemented in (and is one of the few diverging features) of the https://github.com/hometown-fork/hometown/ fork of Mastodon, which runs https://weirder.earth, https://realsocial.life, and [others](https://github.com/hometown-fork/hometown/wiki/Hometown-servers).

PR #8429, opened 2018-08-25
Default puma to running on localhost only.

dracos:
I see https://github.com/tootsuite/mastodon/issues/1283 was closed but web and streaming, when installed using the production guide, both default to listening on 0.0.0.0 though the production guide sets them up to only be needed internally.

I also see https://github.com/tootsuite/mastodon/pull/1747/files which did similar but closed due to inactivity.

This fixes the web service, the streaming can be fixed in the production guide so I'll submit a PR there too.

You can still choose to listen on 0.0.0.0 by running puma with the `--bind` option.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -4,7 +4,8 @@ threads threads_count, threads_count
 if ENV['SOCKET'] then
   bind 'unix://' + ENV['SOCKET']
 else
-  port ENV.fetch('PORT') { 3000 }
+  port_num = ENV.fetch('PORT') { 3000 }
+  bind "tcp://127.0.0.1:#{port_num}"
 end
 
 environment ENV.fetch('RAILS_ENV') { 'development' }


Gargron:
If you bind to 127.0.0.1 inside Docker doesn't it become inaccessible outside Docker? I think that might have been the reason for this default, I am not sure.

dracos:
Docker already binds to 127.0.0.1: https://github.com/tootsuite/mastodon/blob/93537c55605eea78aac5ceeec7ca080b3c3befca/docker-compose.yml#L43

Gargron (approved):

zunda:
Unfortunately, this change broke Heroku apps resulting with

```
Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch
```

Other instances that are not on Docker or nginx may have similar issue that the app becomes inaccessible from outside the machine.

Might there be something `config/puma.rb` make use of to decide to which IP address it should bind? Otherwise, this should be configurable with an environment variable, imo.

Gargron:
For Heroku, the Procfile must be modified with `-b 0.0.0.0`, no?

dracos:
That sounds good to me; I didn't realise they'd be listening directly externally without anything inbetween.

dracos:
Have submitted #8442, thanks.

Gargron:
I feel like everything just got needlessly more complicated :confused: Can we do it so nobody needs to do string interpolation? Just have the 0.0.0.0 and port be separate params?

dracos:
puma does not take a `host` parameter (and I think `port` is only for backwards compatibility), only a `bind` parameter.

dracos:
Plus at https://github.com/tootsuite/mastodon/pull/1747#discussion_r114925682 you said you didn't want an envvar for `HOST` anyway :)

Gargron:
I just realized this is gonna break my config on a multi-machine setup. I don't mind adjusting, but this change might catch more admins by surprise. The thing is, you've gotta be using iptables or ufw anyway, and even if the 3000 or 4000 port is open to the internet, it's not like you're getting anything you wouldn't be getting through nginx.

Do you mind if I revert this after all and close the referenced issue?

dracos:
Not at all, sorry for causing this confusion.

PR #8437, opened 2018-08-25
Add CLI interface for importing custom emoji

### LINKED ISSUE #8435: API call or command for bulk updating emoji
I want to be able to bulk delete and bulk upload (or bulk update) emoji to my mastodon instance, specifically [Mutant Standard](https://mutant.tech). Every update makes changes and I don't want to go through 14,000 images manually every time there is an update.

* * * *

Gargron:
```                      
Usage:
  tootctl emoji import PATH

Options:
  [--prefix=PREFIX]                
  [--suffix=SUFFIX]                
  [--overwrite], [--no-overwrite]  
  [--unlisted], [--no-unlisted]    

Description:
  Imports custom emoji from a TAR archive specified by PATH.

  Existing emoji will be skipped unless the --overwrite option is provided, in 
  which case they will be overwritten.

  With the --prefix option, a prefix can be added to all generated shortcodes. 
  Likewise, the --suffix option controls the suffix of all shortcodes.

  With the --unlisted option, the processed emoji will not be visible in the 
  emoji picker (but still usable via other means)
```

Fix #8435

### DIFF
--- a/lib/mastodon/cli_helper.rb
+++ b/lib/mastodon/cli_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+dev_null = Logger.new('/dev/null')
+
+Rails.logger                 = dev_null
+ActiveRecord::Base.logger    = dev_null
+HttpLog.configuration.logger = dev_null
+Paperclip.options[:log]      = false
--- a/lib/mastodon/emoji_cli.rb
+++ b/lib/mastodon/emoji_cli.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'rubygems/package'
+require_relative '../../config/boot'
+require_relative '../../config/environment'
+require_relative 'cli_helper'
+
+# rubocop:disable Rails/Output
+
+module Mastodon
+  class EmojiCLI < Thor
+    option :prefix
+    option :suffix
+    option :overwrite, type: :boolean
+    option :unlisted, type: :boolean
+    desc 'import PATH', 'import emoji from a TAR archive at PATH'
+    long_desc <<-LONG_DESC
+      Imports custom emoji from a TAR archive specified by PATH.
+
+      Existing emoji will be skipped unless the --overwrite option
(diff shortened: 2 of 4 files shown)


connyduck:
Why only tar? Can't you make it work with zip and/or folders too? Otherwise awesome feature.

Gargron:
@connyduck Because tar is supported by the ruby stdlib so no extra dependencies are required. If you search "how to open zip files in ruby" it's a bit of a mess. It's not like there is an established format for emoji packages already so might as well pick the one easiest to work with, plus the backup service exports as tar so there's some precedent there.

lib/mastodon/cli_helper.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+dev_null = Logger.new('/dev/null')
+
+Rails.logger                 = dev_null
+ActiveRecord::Base.logger    = dev_null
+HttpLog.configuration.logger = dev_null
+Paperclip.options[:log]      = false
```
ThibG: Why disable logging entirely?

Gargron: Clean stdout, otherwise it's just a mess of ActiveRecord/Paperclip messages. This is ported over from the rake file!

ThibG: Hm, ok. This sounds like this could still be useful, though. Maybe make that configurable with a `--verbose` switch for instance?

Gargron: Thor doesn't seem to have a callback system so that would require very repetetive code in every method, right now it's just two but I still don't like that.

lib/mastodon/emoji_cli.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'rubygems/package'
```
ykzts: Is this line required?

ykzts (approved):

angristan:
Is there a way to export the local emojis from Mastodon?

Issue #593, opened 2017-02-03
media upload feature does not make it clear what's allowed


wchristian:
The errors you get from uploading a media file sadly come from the server, after upload, and are quite cryptic.

I think:

1. media uploads should be checked against simple filename checks *before* uploading
2. error messages should contain info on what's allowed
3. possibly also make a tooltip for the media button (and no, that doesn' obsolete 2, due to non-hoverable hardware)

PR #8443, opened 2018-08-25
Specify allowed media in upload button tooltip.

dracos:
Fixes part 3 of #593.


Gargron:
I guess someone might say tooltips are not accessible enough to be a real fix for that problem, but I don't know what else to do either :thinking:

Gargron (approved):

PR #8444, opened 2018-08-25
Revert "Default puma to running on localhost only."

Gargron:
Reverts tootsuite/mastodon#8429

>I just realized this is gonna break my config on a multi-machine setup. I don't mind adjusting, but this change might catch more admins by surprise. The thing is, you've gotta be using iptables or ufw anyway, and even if the 3000 or 4000 port is open to the internet, it's not like you're getting anything you wouldn't be getting through nginx.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -4,8 +4,7 @@ threads threads_count, threads_count
 if ENV['SOCKET'] then
   bind 'unix://' + ENV['SOCKET']
 else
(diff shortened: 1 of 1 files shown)


dracos:
Sure you've noticed, but just in case, I think this reverts too much.

Gargron:
That's a quirk of the GitHub UI. If you look at files changed, you'll see it only reverts the right thing.

PR #8445, opened 2018-08-25
Spread out crawling randomly to avoid DDoSing the link

### LINKED ISSUE #4486: Mastodon can be used as a DDOS tool

Gargron:
Fix #4486

### DIFF
--- a/app/lib/activitypub/activity.rb
+++ b/app/lib/activitypub/activity.rb
@@ -104,7 +104,7 @@ class ActivityPub::Activity
 
   def crawl_links(status)
     return if status.spoiler_text?
-    
+
     # Spread out crawling randomly to avoid DDoSing the link
     LinkCrawlWorker.perform_in(rand(1..59).seconds, status.id)
   end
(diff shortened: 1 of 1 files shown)


nightpool (approved):

angristan:
Interesting approach 🤔

Gargron:
It means that for some servers, their users won't see a link preview for a toot for a minute after it first appears in the system. Also, spreading out ~4,000 requests over 60 seconds might still overload websites, but it's a bit better than ~4,000 in the same second.

The other approach, #8015, would incidentally leak information to the website such as "someone from this server actually viewed a toot with a link to this website", and looking up remote sites in Puma threads is a big performance bottleneck.

Anyway, perhaps the requests should be spread out over even more time than just a minute, but I don't know what would be acceptable UX-wise.

angristan:
Yes, it seems to be the best and easiest workaround for now.

PR #8446, opened 2018-08-25
Fix CW issues in static view on Safari (Fixes #8354)

### LINKED ISSUE #8354: Content Warnings broken in static view on Safari
I was going to look into this myself but I haven't had time so I'm submitting an issue in case someone else has a chance to check it out.

Right now (since the static page redesign) in Safari (on both macOS and iOS), CWs are broken on the static pages for *toots only* (not profiles); for example, https://mastodon.social/@Gargron/100452801467744526

The console shows `Error: The elements you're trying to select don't exist.`, and scripting on the page seems broken.

This is a real problem for iOS users, as they can't open any shared toot with a CW (choosing a different browser is not an option). Firefox and Chrome on macOS both handle the toots fine.

* * * *

AstroPatty:
Fixes issue #8354 

The Rellax function throws an error when there are no parallax objects on the page (such as static view for a post). Chrome and Firefox don't seem to mind but Safari appears to stop script execution at the point of the error, preventing public.js from delegating "show more" behavior. Checking to ensure parallax components exist prior on the page to calling Rellax fixes the issue.

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -78,10 +78,10 @@ function main() {
         })
         .catch(error => console.error(error));
     }
-        
-    const parallaxComponents = document.querySelectorAll('.parallax')
+    
+    const parallaxComponents = document.querySelectorAll('.parallax');
     if (parallaxComponents.length > 0 ) {
-        new Rellax('.parallax', { speed: -1 });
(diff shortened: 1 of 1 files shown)


Gargron (approved):
I wouldn't have thought of it! Thanks. (I also don't have a Safari to test with)

PR #8447, opened 2018-08-25
Allow most kinds of characters in URL query (fixes #8408)

### LINKED ISSUE #8408: URL Parsing does not work with UTF-8 characters
URLs that contain UTF-8 characters are not correctly recognized.
As an example see https://mastodon.technology/@ashfurrow/100605546141133321 (or the image below in case the status is not accessible anymore)
screenshot 2018-08-24 16 14 09

In the example, the URL is cut off before the tick (✓).

JMendyk:
Similiary to #4941, regexes from [twitter-text](https://github.com/twitter/twitter-text/tree/master/rb) were weakened to allow most kinds of characters in URL query string. Includes tests with unicode character in query string to avoid potential regressions in future.

Fixes #8408


Gargron:
I'm worried this might break something :thinking:

Gargron:
@JMendyk It might affect how people use URLs in messages if it "eats up" characters that used to separate text from the URL... I don't know this affects non-ASCII languages especially, Japanese doesn't use spaces between words, do they use spaces around URLs?

JMendyk:
Well, I didn't know that Mastodon API returns toots with already formatted text, which makes my idea pointless as the text (including links) has already been altered in some way.

JMendyk:
I created an alternative approach, in which before calling TwitterText, source text is url escaped to remove unicode characters; then when processed by TwitterText, brought back to old form. I am aware that it certainly has issues/unexpected consequences but I decided to share it because current match-all regex is obviously no good of a solution and as a PoC/food for thought.

Gargron (approved):

PR #8457, opened 2018-08-26
Optionally expose public toots over API without auth

### LINKED ISSUE #7087: Make user statuses available when not authenticated
Currently, it seems that /api/v1/accounts/:id/statuses return 401 if you are not authenticated.

My client supports browsing timelines and statuses before logging in, but this breaks showing user pages.

It seems a bit strange that this endpoint requires authentication, since it is publicly visible through the web interface. Could it be changed to work when not authenticated?

(I think the same applies to /api/v1/accounts/:id but I haven't looked into it properly.)

* * * *

qguv:
Fixes #7087. Admins would have to opt-in.

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -22,6 +22,7 @@ module Admin
       min_invite_role
       activity_api_enabled
       peers_api_enabled
+      public_toots_api
       show_known_fediverse_at_about_page
       preview_sensitive_media
       custom_css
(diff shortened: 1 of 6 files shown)


DagAgren:
If it is opt-in, then is the choice exposed to the client somewhere so it can adjust its UI accordingly?

Gargron:
@DagAgren I am not sure about all of this, that's a bit premature

qguv:
This feature was implemented with #9573. Closing as this PR is no longer necessary. Cheers!

maxf:
Nice! Thanks :+1:

PR #8458, opened 2018-08-26
Fix low-hanging rubocop gripes

qguv:
Takes rubocop warnings from 124 down to 63. Remaining warnings might actually have some sense to them, but I want to let Garg make those calls.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,6 +62,9 @@ Metrics/ParameterLists:
 Metrics/PerceivedComplexity:
   Max: 20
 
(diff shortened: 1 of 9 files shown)


Gargron:
Not sure about the HTTP status codes (int is more readable to me than remembering what they are called) but everything else looks good.

qguv:
I don't mind the status codes either (plus they're language-independent), but rubocop complains. Perhaps those lints should just be turned off?

Gargron:
Feel free to turn that lint off.

Gargron (approved):

PR #8466, opened 2018-08-26
Add CLI task for rotating keys

Gargron:
```
Usage:
  tootctl accounts rotate [USERNAME]

Options:
  [--all], [--no-all]  

Description:
  Generate and broadcast new RSA keys as part of security maintenance.

  With the --all option, all local accounts will be subject to the rotation. 
  Otherwise, and by default, only a single account specified by the USERNAME 
  argument will be processed.
```

### DIFF
--- a/lib/mastodon/accounts_cli.rb
+++ b/lib/mastodon/accounts_cli.rb
@@ -21,11 +21,16 @@ module Mastodon
     def rotate(username = nil)
       if options[:all]
         processed = 0
+        delay     = 0
 
-        Account.local.find_each do |account|
-          rotate_keys_for_account(account)
-          processed += 1
-          say('.', :green, false)
+        Account.local.find_in_batches do |accounts|
+          accounts.each do |account|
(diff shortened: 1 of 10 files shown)


lib/mastodon/accounts_cli.rb
```diff
@@ -0,0 +1,50 @@
...
+      With the --all option, all local accounts will be subject
+      to the rotation. Otherwise, and by default, only a single
+      account specified by the USERNAME argument will be
+      processed.
+    LONG_DESC
+    def rotate(username = nil)
+      if options[:all]
+        processed = 0
+
+        Account.local.find_each do |account|
```
nightpool: For large instances, it's probably better to do this in batches, instead of all at once. Even for an instance like cybre.space, federating 1500 updates at once is a lot for everyone else to handle.

nightpool (changes requested):

nightpool (approved):
LGTM although we should make sure to test this on some medium-sized instances

PR #8473, opened 2018-08-26
Add CSS class to replies

MaciekBaron:
This change adds `status__wrapper-reply` and `status-reply` CSS classes to replies for additional styling needs.


nightpool (approved):

nightpool:
we might want an explicit !! so it's clear we're coercing this to a boolean

nightpool:
yes the concern was clarity, not correctness

ClearlyClaire (approved):
Looks good to me, but you may also want something similar for public pages?

nightpool:
that's out of scope for this PR imo

PR #8477, opened 2018-08-27
AWS Cloudformation deployment - Full end-to-end Mastodon Instance

aus-social:
Hey team...

This is my 0.1.0 release of my Cloudformation... there are a few things we need to improve but it's almost a completely hands off instance deployment.

### DIFF
--- a/aws/cloudformation.yml
+++ b/aws/cloudformation.yml
@@ -29,28 +29,28 @@ Parameters:
       Type: String
 
     Domain:
-      Description: 'example.com'
+      Description: 'social.example'
       Type: String
-      Default: 'darkarts.club'
+      Default: 'social.example'
 
     DomainEmail:
-      Description: 'admin@example.com - used to register certbot'
+      Description: 'admin@example.com - used to register certbot/admin account'
       Type: String
-      Default: 'admin@darkarts.club'
+      Default: 'admin@social.example'
 
     R53Zone:
       Description: AWS KeyName
       Type: 'AWS::Route53::HostedZone::Id'
 
     ReleaseVersion:
-      Description: 'v2.4.4'
+      Description: 'v2.5.0rc1'
       Type: String
-      Default: 'v2.4.4'
+      Default: 'v2.5.0rc1'
 
     EC2PrivateIP:
       Description: 'Private IP for the EC2 Primary Instance'
       Type: String
-      Default: '10.32.0.34'
+      Default: '10.xxx.xxx.xxx'
 
     MastodonLocalUserPassword:
       Description: 'mastodon local user password'
@@ -62,14 +62,19 @@ Parameters:
       Type: 'AWS::EC2::KeyPair::KeyName'
 
     ImageId:
-      Description: Official Ubuntu 18.04 AMI
+      Description: Official Ubuntu 18.04 LTS - Bionic
       Type: String
-      Default: ami-0118debc71f4a5be3
+      Default: ami-09b984525e8ba2206
+
+    SwapSize:
+      Description: GB of Swap space
+      Type: String
+      Default: 5
 
     MailgunLOGIN:
       Description: 'Mailgun username'
       Type: String
-      Default: 'notifications@mail.example.com'
+      Default: 'notifications@mail.social.example'
     MailgunPASSWORD:
       Description: 'Mailgun password'
       Type: String
@@ -176,7 +181,11 @@ Resources:
             Version: 2012-10-17
             Statement:
               - Effect: Allow
-                Action: 's3:*'
+                Action: 
+                  - 's3:ListBucket'
+                  - 's3:GetObject*'
+                  - 's3:PutObject*'
+                  - 's3:DeleteObject*'
                 Resource: 
                   - !Sub 'arn:aws:s3:::${MasotodonS3bucket}'
                   - !Sub 'arn:aws:s3:::${MasotodonS3bucket}/*'
@@ -350,13 +359,13 @@ Resources:
                 echo "$?" >> /var/tmp/bootstrap.txt
 
                 echo "Activate Swap" >> /var/tmp/bootstrap.txt
-                fallocate -l 5G /swapfile
+                fallocate -l ${SwapSize}G /swapfile
                 chmod 600 /swapfile
                 mkswap /swapfile
                 swapon /swapfile
                 echo '/swapfile       none    swap    sw      0       0' | tee -a /etc/fstab
 
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -417,11 +426,13 @@ Resources:
                   echo "$?" >> /var/tmp/bootstrap.txt
 
                   echo "db:setup" >> /var/tmp/bootstrap.txt
-                  docker-compose run --rm web rails db:setup -e SAFETY_ASSURED=1 -e RAILS_ENV=production # Create DB tables <-- BORKED?
+                  docker-compose run --rm web bundle exec rails db:setup -e SAFETY_ASSURED=1 -e RAILS_ENV=production 
+                  # Create DB tables (Destructive: Deletes existing tables)
                   echo "$?" >> /var/tmp/bootstrap.txt
 
                   echo "assets:precompile" >> /var/tmp/bootstrap.txt
-                  docker-compose run --rm web rails assets:precompile  # Compile assetts
+                  docker-compose run --rm web bundle exec rails assets:precompile  
+                  # Compile assetts (Notice: docker will complete this step using the build step)
                   echo "$?" >> /var/tmp/bootstrap.txt
 
                   echo "docker-compose up -d" >> /var/tmp/bootstrap.txt
@@ -430,7 +441,7 @@ Resources:
 
                 - { REDIS_ENDPOINT: !GetAtt ElastiCacheRedis.RedisEndpoint.Address, REDIS_PORT: !GetAtt ElastiCacheRedis.RedisEndpoint.Port, DB_ENDPOINT: !GetAtt RDSPostgreSQLInstance.Endpoint.Address, DB_PORT: !GetAtt RDSPostgreSQLInstance.Endpoint.Port, S3_ACCESSKEY: !Ref MasotodonS3bucketKey, S3_SECRETKEY: !GetAtt MasotodonS3bucketKey.SecretAccessKey }
               
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -466,7 +477,7 @@ Resources:
                 certbot certonly --standalone -d ${Domain} --agree-tos -m ${DomainEmail} --non-interactive
                 echo "$?" >> /var/tmp/bootstrap.txt
 
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -478,7 +489,7 @@ Resources:
                 ## RAILS_ENV=production
                 ## @daily cd /home/mastodon/live && /home/mastodon/.rbenv/shims/bundle exec rake mastodon:media:remove_remote
 
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -580,7 +591,7 @@ Resources:
 
                   error_page 500 501 502 503 504 /500.html;
                 }
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -629,7 +640,7 @@ Resources:
                     external_network:
                     internal_network:
                       internal: true
-              mode: "000777"
+              mode: "000755"
               owner: "ubuntu"
               group: "ubuntu"
 
@@ -659,7 +670,11 @@ Resources:
           PolicyDocument:
             Statement:
               - Effect: Allow
-                Action: 's3:*'
+                Action:
+                  - 's3:ListBucket'
+                  - 's3:GetObject*'
+                  - 's3:PutObject*'
+                  - 's3:DeleteObject*'
                 Resource: 
                   - !Sub 'arn:aws:s3:::${MasotodonS3bucket}'
                   - !Sub 'arn:aws:s3:::${MasotodonS3bucket}/*'
--- a/aws/cloudformation.yml
+++ b/aws/cloudformation.yml
@@ -52,6 +52,11 @@ Parameters:
       Type: String
       Default: '10.32.0.34'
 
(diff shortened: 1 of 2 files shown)


aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+  MasotodonS3buckerUser:
+    Type: 'AWS::IAM::User'
+    Properties:
+      Policies:
+        - PolicyName: S3
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action: 's3:*'
```
ojbr: We shouldn't need access to wildcard. I'd suggest limiting our scope to these:
- `ListBucket`
- `GetObject*`
- `PutObject*`
- `DeleteObject*`

Otherwise you're basically giving the IAM user the permission to run wild

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+
+                echo "Create mastodon group" >> /var/tmp/bootstrap.txt
+                addgroup -gid 991 mastodon
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "Create mastodon user" >> /var/tmp/bootstrap.txt
+                adduser --quiet --disabled-password --gid 991 --uid 991 mastodon --gecos GECOS
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "Grant mastodon sudo" >> /var/tmp/bootstrap.txt
```
ojbr: Er, why?

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                usermod -a -G sudo mastodon
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "change mastodon password" >> /var/tmp/bootstrap.txt
+                echo "mastodon:${MastodonLocalUserPassword}" | sudo chpasswd
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "Install Docker" >> /var/tmp/bootstrap.txt
```
ojbr: If we're doing a CF template, then it might be a good idea to avoid using Docker. The stack isn't particularly heavy; adding Docker to the equation is a little bit lazy

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                usermod -aG docker mastodon
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "install docker-compose" >> /var/tmp/bootstrap.txt
+                curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose
+                chmod +x /usr/local/bin/docker-compose
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "Activate Swap" >> /var/tmp/bootstrap.txt
+                fallocate -l 5G /swapfile
```
ojbr: A configurable swapfile size parameter would be nice. Depending on the scale of the box, this may or may not be wanted

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                chmod 600 /swapfile
+                mkswap /swapfile
+                swapon /swapfile
+                echo '/swapfile       none    swap    sw      0       0' | tee -a /etc/fstab
+
+              mode: "000777"
```
ojbr: 0755

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+
+                  echo "copy AWS docker-compose.yml" >> /var/tmp/bootstrap.txt
+                  rm -f docker-compose.yml
+                  cp /var/tmp/docker-compose.yml /home/mastodon/live/docker-compose.yml
+
+                  echo "create .env.production" >> /var/tmp/bootstrap.txt
+                  cat >> /home/mastodon/live/.env.production << EOF
+                  LOCAL_DOMAIN=${Domain}
+                  SINGLE_USER_MODE=false
+                  SECRET_KEY_BASE=fc6379db2af899b3c980c83fc725f06996e9c5994b8adf34ec3c346731b6964c7b850209e0edfbc594d26a95108b6fe344bd0c9d2deeee127b51184d75a90cce
```
ojbr: All these secrets need to be parameterized in the YAML, with sensible defaults to highlight the need to change them. We don't want to encourage secret re-use!

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                  echo "docker-compose build" >> /var/tmp/bootstrap.txt
+                  docker-compose build 
+                  echo "$?" >> /var/tmp/bootstrap.txt
+
+                  echo "db:setup" >> /var/tmp/bootstrap.txt
+                  docker-compose run --rm web rails db:setup -e SAFETY_ASSURED=1 -e RAILS_ENV=production # Create DB tables <-- BORKED?
+                  echo "$?" >> /var/tmp/bootstrap.txt
+
+                  echo "assets:precompile" >> /var/tmp/bootstrap.txt
+                  docker-compose run --rm web rails assets:precompile  # Compile assetts
```
ojbr: Bundler context:
`docker-compose run --rm web bundle exec rails assets:precompile`

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+
+                  cd /home/mastodon/live  # Just incase
+
+                  echo "docker-compose build" >> /var/tmp/bootstrap.txt
+                  docker-compose build 
+                  echo "$?" >> /var/tmp/bootstrap.txt
+
+                  echo "db:setup" >> /var/tmp/bootstrap.txt
+                  docker-compose run --rm web rails db:setup -e SAFETY_ASSURED=1 -e RAILS_ENV=production # Create DB tables <-- BORKED?
+                  echo "$?" >> /var/tmp/bootstrap.txt
```
ojbr: In a bundler context would be a good idea:
```
docker-compose run --rm web bundle exec rails db:setup -e SAFETY_ASSURED=1 -e RAILS_ENV=production
```

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                  docker-compose run --rm web rails assets:precompile  # Compile assetts
+                  echo "$?" >> /var/tmp/bootstrap.txt
+
+                  echo "docker-compose up -d" >> /var/tmp/bootstrap.txt
+                  docker-compose up -d
+                  echo "$?" >> /var/tmp/bootstrap.txt
+
+                - { REDIS_ENDPOINT: !GetAtt ElastiCacheRedis.RedisEndpoint.Address, REDIS_PORT: !GetAtt ElastiCacheRedis.RedisEndpoint.Port, DB_ENDPOINT: !GetAtt RDSPostgreSQLInstance.Endpoint.Address, DB_PORT: !GetAtt RDSPostgreSQLInstance.Endpoint.Port, S3_ACCESSKEY: !Ref MasotodonS3bucketKey, S3_SECRETKEY: !GetAtt MasotodonS3bucketKey.SecretAccessKey }
+              
+              mode: "000777"
```
ojbr: 0755

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+                apt-get update
+                apt-get --yes install python-certbot-nginx
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+                echo "certbot certonly standalone" >> /var/tmp/bootstrap.txt
+                systemctl stop nginx.service
+                certbot certonly --standalone -d ${Domain} --agree-tos -m ${DomainEmail} --non-interactive
+                echo "$?" >> /var/tmp/bootstrap.txt
+
+              mode: "000777"
```
ojbr: 0755

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+            Principal:
+              Service: ec2.amazonaws.com
+            Action: 'sts:AssumeRole'
+      Path: /
+      Policies:
+        - PolicyName: Oracle-ec2
+          PolicyDocument:
+            Statement:
+              - Effect: Allow
+                Action: 's3:*'
```
ojbr: Same again for this wildcard

Also on L668, don't give us more than we need D:

aws/readme.md
```diff
@@ -0,0 +1,51 @@
...
+* 1 x S3 Bucket for storage (with AWS IAM account)
+
+## Installation Guide
+1) Setup a new Route53 zone and forward the domain nameservers correctly.
+2) Register with mailgun.com and setup required DNS records for approval.
+3) Create the AWS CFN Stack with the required details.
+4) Register an account on your instance (log in and promote as admin using cli)
+
+## Troubleshooting
+I'm sending user-data/meta-data debug information into a file - `/var/tmp$ tail -f /var/tmp/bootstrap.txt`
```
ojbr: "Debug information is logged to..." - let's take this out of first person

aws/readme.md
```diff
@@ -0,0 +1,51 @@
+# AWS Cloudformation Stack - End to End deployment
+* 1x EC2 Server (Dockerised Mastodon Instance)
+* 1x RDS PostgreSQL (for Database)
+* 1 x ElasticCache Redis (for shared cache)
+* 1 x S3 Bucket for storage (with AWS IAM account)
```
ojbr: Can we document the created IAM Users/Roles?

ghost (changes requested):
Hey! Thanks for the CFT.
There are a few things I've highlighted which could be a bit of a security concern, and general good practice. Otherwise, it's a pretty good first CFT for Mastodon.

Perhaps we could drop Docker when running on EC2? Given the fact we're already doing a bunch of setup in user data scripts, it would make the overall result smaller and faster.
Cc @Gargron

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+
+    UniqueID:
+      Description: 'lower case'
+      Default: mastodon1
+      Type: String
+
+    Domain:
+      Description: 'example.com'
+      Type: String
+      Default: 'darkarts.club'
```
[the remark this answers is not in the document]
ashleyhull-versent: Thanks, this was just a typo.

aws/cloudformation.yml
```diff
@@ -0,0 +1,699 @@
...
+      Description: 'admin@example.com - used to register certbot'
+      Type: String
+      Default: 'admin@darkarts.club'
+
+    R53Zone:
+      Description: AWS KeyName
+      Type: 'AWS::Route53::HostedZone::Id'
+
+    ReleaseVersion:
+      Description: 'v2.4.4'
```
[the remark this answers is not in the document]
ashleyhull-versent: Thanks, this was just a typo again.

aws/cloudformation.yml
ashleyhull-versent: I'm not 100% sure what you mean? this is just following the production guide to match the GID/UID from the docker container.

ashleyhull-versent: Sudoer is not required for the installation.. but I figured it be nice for the admin to have for maintenance.

Let's drop it for a little extra security posture :)

aws/cloudformation.yml
ashleyhull-versent: 1) I don't think Docker is lazy, but I also think it's worth it for the upgrade (and fallback) path.

* git fetch
* git stash
* git checkout {newversion}
* git pop
* edit docker-compose.yml image to {newversion}
* docker-compose build
* db:migrate if required
* docker-compose up -d

That's a pretty clean upgrade path?.... I can be convinced to remove all of the docker for the "performance" boot.

2) ECS is not a good fit for my template... and while I think Mastodon could easily be turned into a ECS task.... it's not "microservice" enough atm.

ECS also would assume you're using an ALB in front of your ECS tasks and that would increase the cost.

aws/cloudformation.yml
ashleyhull-versent: A lot of there Mastodon instances are running on 1-2GB instances.. so 5GB was overkill..

but I've turned that into a parameter.

aws/cloudformation.yml
ashleyhull-versent: ok.. this was one of those chicken and egg things....

`docker-compose run --rm web rake secret`
`docker-compose run --rm web rake mastodon:webpush:generate_vapid_key`

requires the docker-build build has completed... but I was generating the .env.production in a previous step.

@Gargron Is it safe to change the secrets AFTER you've done the db:setup? I think they're used for MFA stuff?

aws/cloudformation.yml
ashleyhull-versent: I didn't know the difference.. If the bundle is the preferred method, we should consider updating the doco to match.

I've made these changes.

aws/cloudformation.yml
ashleyhull-versent: also.. `RAILS_ENV=production` might not be required... I'll check the dockerfile

aws/cloudformation.yml
ashleyhull-versent: fixed.

ashleyhull-versent:
1) This CFT has been created with the goal of allowing people to launch multiple Mastodon instances on the account.  All of the resources should have unique names - so somebody could launch multiple Mastodons on a single account using the same CFT without clashing.

2) this CFT does not use any KMS (at-rest) encryption on the RDS/ElasticCache (extra cost and extra work but it totally could)

3) I also think it's important to stress this isn't meant to be hands off after deployment.

There are plenty of customization/maintenance and operational work that will be required after it is deployed.

MaciekBaron:
Just a general comment, can we squash these commits so our git history is cleaner?

aus-social:
Sorry. I am bad at git. I'll wait to hear from the team regarding the progress.

aws/cloudformation.yml
ojbr: The front end doesn't need to incur downtime if you skip using Docker. But for a v1 CFT, Docker can probably stay

aws/cloudformation.yml
ojbr: I believe there is a secrets rotation tool in 2.5.0 with toot-ctl, but I could be mistaken as I haven't seen it documented yet

ojbr:
Thanks for the hardening fixes :+1: 
We just need to decide how we want to handle secrets in the .env.production, then I think this is good to go

ashleyhull-versent:
@Gargron could you consider making a `mastodon-aws` repo and moving these files in?

then we can work to PR my TODO list without spamming you.

MaciekBaron:
Anyone can create a fork and ask people to contribute to it, that way things can be resolved independently before updating this repo.

aus-social:
I'm going to make some additional changes and push this PR again soon.

PR #8506, opened 2018-08-28
Fix CLI interface for removing remote media

mayaeh:
I like #8411 's CLI interface very much.
Thank you!

But I think that it can't delete media files without "--background" option.
Because media_cli.rb's "m" is "MediaAttachment" which extracted "id" only at the time of extraction, I think that it does not include other values.

I think that this change will fix this.

### DIFF
--- a/app/workers/maintenance/uncache_media_worker.rb
+++ b/app/workers/maintenance/uncache_media_worker.rb
@@ -6,7 +6,7 @@ class Maintenance::UncacheMediaWorker
   sidekiq_options queue: 'pull'
 
   def perform(media_attachment_id)
-    media = MediaAttachment.find(media_attachment_id)
+    media = media_attachment_id.is_a?(MediaAttachment) ? media_attachment_id : MediaAttachment.find(media_attachment_id)
 
     return if media.file.blank?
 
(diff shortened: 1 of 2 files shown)


app/workers/maintenance/uncache_media_worker.rb
```diff
@@ -6,7 +6,7 @@ class Maintenance::UncacheMediaWorker
   sidekiq_options queue: 'pull'
 
   def perform(media_attachment_id)
-    media = media_attachment_id.is_a?(MediaAttachment) ? media_attachment_id : MediaAttachment.find(media_attachment_id)
+    media = MediaAttachment.find(media_attachment_id)
```
Gargron: It's very slow like this. I think it's better to `select` more columns if background option is false

mayaeh: Okay. Thank you for your advice.

mayaeh: @Gargron  Is it okay to delete "select()" like "MediaAttachment.where.not(remote_url: '').where.not(file_file_name: nil).where('created_at < ?', time_ago).reorder(nil).find_in_batches" ?

Gargron (approved):

PR #8507, opened 2018-08-28
Add animate custom emoji param to embed pages

renatolond:
Add a "animate_custom_emoji" param to embed pages to allow embedding toots with animated emoji.

### DIFF
--- a/app/controllers/statuses_controller.rb
+++ b/app/controllers/statuses_controller.rb
@@ -54,6 +54,7 @@ class StatusesController < ApplicationController
     skip_session!
     expires_in 180, public: true
     response.headers['X-Frame-Options'] = 'ALLOWALL'
+    @animate_custom_emoji = params[:animate_custom_emoji] == "1"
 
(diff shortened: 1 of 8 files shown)


app/controllers/statuses_controller.rb
```diff
@@ -54,6 +54,7 @@ def embed
     skip_session!
     expires_in 180, public: true
     response.headers['X-Frame-Options'] = 'ALLOWALL'
+    @animate_custom_emoji = params[:animate_custom_emoji] == "1"
```
Gargron: Maybe name the param "autoplay" and use Rails truthiness parser / boolean type cast used in other parts of the code.

Make it control not only custom emoji but avatars and media too

renatolond: I did it, but it seems even with autoPlayGif set to true, the gifs do not autoplay in the public pages :thinking:

Gargron (approved):

Issue #8503, opened 2018-08-28
[UX] "Follow" button too prevalent on a migrated account


aleksandar-todorovic:
**Note:** This only applies to those who are *not* logged in.

I have migrated from one account to another one. Now, when I visit my old account (while not being logged in), I see this:

![screenshot from 2018-08-28 14-03-32](https://user-images.githubusercontent.com/4890548/44721748-2de70d00-aacb-11e8-9649-cf317a038583.png)

As you can notice from the screenshot above, the "follow" button is not only visible, but a higher color contrast makes it the most prominent element on the page. The text below it seems almost unnoticeable.

Proposal: Something visually similar to how it's done while being logged in:

![screenshot from 2018-08-28 14-07-43](https://user-images.githubusercontent.com/4890548/44721960-d1382200-aacb-11e8-84da-21ff8c98a48b.png)

In my opinion, the second screenshot gives a much better impression that the account is migrated from. Even with `[old]` added to my name in the previous sentence, I think that the people who glance at the first screenshot will not figure out that the profile is indeed migrated from. Instead, they'll be lured by the follow button as the primary call-to-action.

**Note no. 2:** This is _not_ a request to remove the follow button completely, just making it less prominent than it currently is in this specific scenario.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

dracos:
Have had a few people follow me even after moving too, have submitted a PR that greys out the whole header to match what happens in the in-column view (including the follow button), as a suggestion.

PR #8520, opened 2018-08-29
Inactive header in main profile to match column.

dracos:
This PR adds an inactive class to the main profile page, greying it out in the same way as when a moved profile is viewed in-column (see second screenshot on #8503), including the follow button. Might count as fixing #8503.

screen shot 2018-08-29 at 19 57 42

### DIFF
--- a/app/javascript/styles/mastodon/containers.scss
+++ b/app/javascript/styles/mastodon/containers.scss
@@ -345,6 +345,23 @@
     margin-bottom: 10px;
     box-shadow: 0 0 15px rgba($base-shadow-color, 0.2);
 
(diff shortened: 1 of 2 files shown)


Gargron (approved):

ThibG:
Having a gray but clickable button is weird. I would remove it entirely.

dracos:
The issue did specifically say "This is not a request to remove the follow button completely, just making it less prominent" so I assumed that had been discussed/discounted somewhere. I personally wouldn't mind it being removed, but I didn't know if there were reasons why moved people might still want to allow following.

PR #8522, opened 2018-08-29
Show column scrollbars only when necessary

### LINKED ISSUE #8521: Show column scrollbars only when necessary
Instead of showing scrollbars all the time by setting

`overflow-y:scroll`

let the browser decide when to show and not show by setting overflow behaviour to auto.

* * * *

andi1984:
Instead of showing scrollbars all the time by setting overflow-y:scroll let the browser decide when to show and not show by setting overflow behaviour to auto.

Fixes #8521

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -2038,7 +2038,7 @@ a.account__display-name {
 }
 
 .scrollable {
-  overflow-y: scroll;
+  overflow-y: auto;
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
The original reason for always displaying the scrollbar was to avoid jumping (https://github.com/tootsuite/mastodon/issues/8521#issuecomment-417409571), so I must reject this PR, but thank you for the contribution.

PR #8526, opened 2018-08-30
Fix different counts for display name and bio

### LINKED ISSUE #4419: Server- and client-side bio and display name count endlines and emoji combinations inconsistently
Hi!

When using UTF-8 chars (multi-byte ones, I assume), there's inconsistency in the "characters left" indication. I'm using mastodon.xyz which is in 1.4.7, problem still exists in 2.5.0rc1
When I submit the edited changes, the value that comes back it's smaller than the one that it's calculated by the frontend.
* * * *

renatolond:
This follows the fix suggested by https://tech.mybuilder.com/managing-newlines-and-unicode-within-javascript-and-php/

Fixes #4419

Is there a way to test public.js? I know that there's some tests inside `app/javascript/mastodon/`, but could I also create one for this file?

It seems before this code was using stringz length, which considers `'🏳️‍🌈'.length` == 1. The problem is that the backend considers `"🏳️‍🌈".length` == 4, even if we use rails mb_chars thingy: `"🏳️‍🌈".mb_chars.length` == 4.

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -1,6 +1,7 @@
 import loadPolyfills from '../mastodon/load_polyfills';
 import ready from '../mastodon/ready';
 import { start } from '../mastodon/common';
+import punycode from 'punycode';
 
 start();
 
@@ -134,7 +135,7 @@ function main() {
     const name        = document.querySelector('.card .display-name strong');
 
     if (nameCounter) {
-      nameCounter.textContent = 30 - length(target.value);
+      nameCounter.textContent = 30 - punycode.ucs2.decode(target.value).length;
     }
 
     if (name) {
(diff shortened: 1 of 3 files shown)


dracos:
Doesn't this fix the issue the 'wrong' way round by saying 🏳️‍🌈 is 4 rather than 1 at both client/server? Wouldn't it be better to keep using `stringz` in the JS and fix the rails code to count it as 1, e.g. `"🏳️‍🌈".grapheme_clusters.length == 1`

Gargron:
>If ruby 2.5 support can be assumed

We have to deal with 2.3 unfortunately.

dracos:
I've submitted a PR at #8531 that maintains the newline fix from this PR (yay) and uses the grapheme regex from `stringz` on the server to do char counting in both display name/bio and the status field. I've tested this on my own instance and posted 500 🏳️‍🌈 successfully and 30 🏳️‍🌈 display name.

renatolond:
I thought about that but didn't go that way because it would be one more regex to maintain, but I think it's indeed a good direction, let's move the talk over to your PR :)

PR #8531, opened 2018-08-30
Use grapheme length regex serverside

dracos:
This uses the regex that `stringz` uses client-side on the server so that the character counting is the same in both places. This should let you post 500 🏳️‍🌈 as a status (previously 250), and up to the appropriate limits for display name and bio (previously e.g. 40 🏳️‍🌈 for bio rather than 160). Fixes the initial display on the edit profile page also.

This includes the newline fix from #8526.

(The regex in `stringz` comes from lodash: https://github.com/lodash/lodash/blob/48511837573d99138fc43b7e9f04eea7b01839fe/.internal/unicodeSize.js for a more nicely broken-down version)

The regex is pretty good but does not cover the full Unicode UAX29 grapheme cluster algorithm, for that you'd need something like https://github.com/orling/grapheme-splitter (and a Ruby equivalent).

### DIFF
--- a/app/validators/status_length_validator.rb
+++ b/app/validators/status_length_validator.rb
@@ -15,7 +15,7 @@ class StatusLengthValidator < ActiveModel::Validator
   end
 
   def countable_length(status)
-    total_text(status).mb_chars.grapheme_length
+    total_text(status).scan(Account::GRAPHEME_RE).length
   end
 
   def total_text(status)
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -596,10 +596,22 @@ RSpec.describe Account, type: :model do
         expect(account).to model_have_error_on_field(:display_name)
       end
 
+      it 'display name with newline and carriage returns gets modified to have only newline' do
+        account = Fabricate.build(:account, display_name: "My display name is\r\ncool")
+        account.valid?
+        expect(account.display_name).to eq "My display name is\ncool"
+      end
+
       it 'is invalid if the note is longer than 160 characters' do
         account = Fabricate.build(:account, note: Faker::Lorem.characters(161))
         account.valid?
         expect(account).to model_have_error_on_field(:note)
+      end
+
+      it 'note with newline and carriage return gets modified to have only newline' do
+        account = Fabricate.build(:account, note: "My note came from a weird browser\r\nBut I like it!")
+        account.valid?
+        expect(account.note).to eq "My note came from a weird browser\nBut I like it!"
       end
     end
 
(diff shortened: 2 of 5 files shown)


renatolond:
Does this grapheme_re changes often? As in: do we need to update this regex as part of updating the emoji packages?

spec/models/account_spec.rb
```diff
@@ -596,11 +596,23 @@
...
+        expect(account.display_name).to eq "My display name is\ncool"
+      end
+
       it 'is invalid if the note is longer than 160 characters' do
         account = Fabricate.build(:account, note: Faker::Lorem.characters(161))
         account.valid?
         expect(account).to model_have_error_on_field(:note)
       end
+
+      it 'Note with newline and carriage return returns correct size' do
```
renatolond: Sorry, I fixed the name in the display name test, but not in this one, could you maybe match it to the newline one?

app/validators/status_length_validator.rb
```diff
@@ -15,7 +15,7 @@ def too_long?(status)
   end
 
   def countable_length(status)
-    total_text(status).mb_chars.grapheme_length
+    total_text(status).scan(Account::GRAPHEME_RE).length
```
renatolond: Maybe we could add a spec to make sure that this works?

dracos: Have hopefully added one

Gargron:
Using a regex to count characters makes me feel weird :confused: This is because of the need to support 2.3, right? How about this: Use the proper 2.5 method, but polyfill it for older Ruby versions with the regex. It will be easier to remove then.

dracos:
> Using a regex to count characters makes me feel weird 😕

It's what you're already doing in the JavaScript :) Sure, I'm happy to do that and use the 2.5 method if available and polyfill to the regex if you like, but then there will again be inconsistencies between client and server counting e.g. this 4-char string:

```
# Ruby 2.5
irb(main):012:0> "\u0061\u0928\u093F\u4E9C\u{10083}".grapheme_clusters.length
=> 4

# JS in the client
stringz.length('\u0061\u0928\u093F\u4E9C\u{10083}')
5
```

You'd have to switch to something like grapheme-splitter in the client to do proper JS grapheme counting; the regex the JS uses does not deal with e.g. Hangul syllable sequences, all the non-Latin characters in https://unicode.org/cldr/utility/list-unicodeset.jsp?a=[:Grapheme_Cluster_Break=Extend:] or all the characters in https://unicode.org/cldr/utility/list-unicodeset.jsp?a=[:Grapheme_Cluster_Break=Spacing_Mark:]. But it is of course much better than naive string counting. And it's not *wrong* - if you copy that string into somewhere and delete it, you'll see it deletes as 5 characters, just two combine for display purposes. So I thought consistency of counting between client and server submission was more important (and easier to do!).

So let me know if you'd prefer perfect count-syncing between server and client (use regex on server as well as client, works fine for emoji, surrogate pairs, Latin combining characters), or server to use 2.5+ counting if available and then be out of sync with client counting for characters where they differ.

Issue #8485, opened 2018-08-27
Can't enable "federation relay"


33b5e5:
I'm the admin of raki.social. I upgraded to 2.5.0rc1 from the release tag. 

The upgrade went fine, but I haven't been able to get the new "federation relay" feature working. After adding the suggested relay (https://relay.joinmastodon.org/inbox) it shows up "Disabled":

![screen shot 2018-08-27 at 11 12 04 am](https://user-images.githubusercontent.com/2781306/44677163-1c7a0400-a9ea-11e8-82b2-6c8579c0b2a7.png)

If I click "Enable", it seems like it's trying to do something, but nothing really happens -- there's no error and it remains "Disabled".

I don't know how to troubleshoot this further. I'm not sure where to look for relevant logs. I tried grepping through syslog but didn't see anything related.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Marlinc:
I ran into the same issue

ThibG:
It shows as `disabled` if the relay hasn't replied back to your follow request. Which it probably didn't do yet because it's lagging behind.
The UI should probably show that it's pending, rather than disabled!

33b5e5:
Ah... okay @ThibG. I will wait for a while and see if it gets the reply. Thanks.

I agree the UI should be improved. Displaying "Pending" would be a great start.

Gargron:
I think it's a mistake on my part... As @ThibG mentions I forgot to change the UI from enabled/disabled to disabled/pending/enabled, and there might be a bug with actually recording the "accept" event because the relay seems to be sending them (and is not lagging behind significantly) and yet this issue exists.

carstenwawer:
Got the same problem. Should we wait for the next release?

noellabo:
Problems may occur when upper case is included in inbox_url.

It may be fixed by adding the following modification to pub-relay.
https://source.joinmastodon.org/mastodon/pub-relay/merge_requests/2

mdbraber:
@noellabo have you tried this change? I've just recompiled my binary on relay.mastodon.nl to test but I can't see any change.

Gargron:
Indeed, it seems that it hasn't fixed this problem.

[...]

PR #8542, opened 2018-08-31
Sign follow/unfollow request of federation relay

u1-liquid:
ActivityPub::DeliveryWorker.perform_async doesn't signing request payloads.
It should be signed by caller, obviously.

I just confirmed this works without relay server changes.
image

### DIFF
--- a/app/models/relay.rb
+++ b/app/models/relay.rb
@@ -26,7 +26,7 @@ class Relay < ApplicationRecord
 
   def enable!
     activity_id = ActivityPub::TagManager.instance.generate_uri_for(nil)
-    payload     = Oj.dump(follow_activity(activity_id))
+    payload     = build_json(follow_activity(activity_id))
 
     update!(state: :pending, follow_activity_id: activity_id)
     ActivityPub::DeliveryWorker.perform_async(payload, some_local_account.id, inbox_url)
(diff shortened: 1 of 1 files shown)


u1-liquid:
related #8485

Gargron:
That doesn't seem necessary as the message doesn't need to be forwarded anywhere. So this change doesn't really do anything..?

u1-liquid:
As reported in #8485 , Some of instances having trouble following relay server.  
Solution suggested by @mdbraber was actually works for my instance ( tsuki.network )
I thought that "https://w3id.org/security/v1" context prevents request being processed.

.... Maybe I'm wrong and it might be timing or mystery bug of relay server.

u1-liquid:
I suspect as per https://source.joinmastodon.org/mastodon/pub-relay/blob/master/src/inbox_handler.cr#L20 will requires json signing for all of received requests.
actually I'm curious why it gives 202 for unsigned request...

Gargron:
It's resolved: https://github.com/tootsuite/mastodon/issues/8485#issuecomment-417520855

u1-liquid:
It's strange, I could not enable without this when I tried it just a while ago ...
Anyway, I just confirmed that the relay status can be turned 'Enabled' with latest master repository.
Sorry for interrupting you!

PR #8547, opened 2018-09-01
Make detailed-status__wrapper actually wrap detailed status

trwnh:
.detailed-status__wrapper does not actually wrap the detailed status right now, as the class is applied to the scrollable div and not the focusable div.

### DIFF
--- a/app/javascript/mastodon/features/status/index.js
+++ b/app/javascript/mastodon/features/status/index.js
@@ -415,7 +415,7 @@ export default class Status extends ImmutablePureComponent {
         />
 
         <ScrollContainer scrollKey='thread' shouldUpdateScroll={shouldUpdateScroll}>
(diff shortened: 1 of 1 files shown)


trwnh:
Can this be reviewed before 2.6.0?

Gargron:
I'm sorry but could you please clarify the significance of this? Does something change in the style or JS behaviour?

trwnh:
@Gargron Right now, when attempting to apply styles to the div immediately wrapping a detailed status, there is no easy selector that can be used to target it. It doesn't matter so much for the default style, but it makes custom styling harder than it should be. It's incredibly overly-broad in scope:
image

It's also technically a bug, since the __wrapper class is immediately outside every other status class.

Gargron (approved):

PR #8555, opened 2018-09-01
Integrating `:one` and `:other` in ar.yml (regression from #8535)

### LINKED ISSUE #8554: mastodon.social broken in arabic
Hi,
mastodon.social is completely broken when I try to access to it with my browser user-agent language is in arabic. (all URLs).
In Farsi, it works. Please, someone to check those `<span` problem in translation.
Thank you :)

ykzts:
Resolve #8554

image

In Arabic, `zero:`, `one:`, `two:`, `few:`, `many:` and,`other:` are required when handling numerical values. In `one:` and `other:` alone, Rails returns an error.

In this pull request, we changed to use only `other:` as fallback. Since it is just before v2.5.0 is released, it only keeps ad hoc fixes.

### DIFF
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -30,22 +30,16 @@ ar:
     other_instances: خوادم أخرى
     privacy_policy: سياسة الخصوصية
     source_code: الشفرة المصدرية
-    status_count_after:
-      one: منشور
-      other: منشورات
+    status_count_after: منشورات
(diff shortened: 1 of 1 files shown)


ykzts:
In Arabic, `zero:`, `one:`, `two:`, `few:`, `many:` and,`other:` are required when handling numerical values.

Gargron (approved):

PR #8594, opened 2018-09-03
Add Czech language option and locale data

Aditoo17:


app/javascript/mastodon/locales/locale-data/Mastodon-cs-data.txt
```diff
@@ -0,0 +1,138 @@
+/*eslint eqeqeq: "off"*/
```
ykzts: [`locale-data/cs.js`](https://unpkg.com/react-intl@2.4.0/locale-data/cs.js) https://unpkg.com/react-intl@2.4.0/locale-data/cs.js

Gargron: Also the file is named `Mastodon-cs-data.txt`

Aditoo17: Yeah, was uploading that and forgot to rename it.

ykzts (approved):

Issue #8428, opened 2018-08-25
Add the ability to write instance's rules in several languages


Sylvhem:
At the moment, there is no practical way to display the content of the /about, /about/more and /terms pages in several languages. Sure, you can put the different versions of the same text inside `<div lang="x">` tags (that's what we do at Eldritch Café, see [/about/more](https://eldritch.cafe/about/more)) but it requires that the admin is familiar with HTML and it can makes really long pages as a result. Moreover, as for Mastodon 2.4.5, you can't even use this method on the /about page since the text must stay short (this is going to change in Mastodon 2.5).
I would suggest that Mastodon use a similar function that the one currently used on [Liberapay](https://liberapay.com/). The website allow you to write a description in various languages (see screenshot). We could do the same thing, using the browser's language to display the right text, with a fallback on the instance's default language in case the text hasen't been translated in one of favorite languages of the user.

![Screenshot of Liberapay's interface.](https://user-images.githubusercontent.com/5222316/44618876-7958ab80-a87e-11e8-8081-0c9dd2baea1e.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

ThomasLeister:
I'd like to have this feature, too! Currently there are no nice ways to present e.g. the privacy terms in multiple languages (if you want to use your own text).

PR #8608, opened 2018-09-04
[WIP] Add language selector on (anonymous) landing page

### LINKED ISSUE #5691: Language selector on the landing/login page
Forking the second point from #4265, as IMO focused issues are always more efficient than wishlists.

The current Mastodon landing page (when not logged in) is localized based on location/browser/system locale settings (#1804 I guess).

This seems to be problematic for users who may live in a country whose language they don't speak, see e.g. https://www.reddit.com/r/gamedev/comments/7cnwu5/new_mastodon_decentralized_twitterlike_instance/dpsh0f0/

Having an easy way to change the language from the landing page would be good for those users. Currently you need to register/login blindly, find the settings, and change the locale there.

* * * *

- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) - `master` not tested, but likely affected

JMendyk:
This PR aims to solve issues described in #5691:
> The current Mastodon landing page (when not logged in) is localized based on location/browser/system locale settings (...) This seems to be problematic for users who may live in a country whose language they don't speak

> I can especially imagine this on public computers, like libraries and internet café's.

## Checklist:
- [x] Add language selection button on landing page
- [x] Implement language selection modal
    When language is selected, "lang" cookie is set (only for this session) with selected language/locale and the page is reloaded.
- [x] Backend respects client's language choice, in order: 
      1. Inferred from location/browser/system locale
      2. "lang" cookie (if set)
      3. User language preference (if signed in)
- [x] Selected locale is stored in preferences during registration
- [x] Fix modal styling for mastodon-light

## Post-feedback checklist:
- [x] Remove flags (and twemoji)
- [ ] Use session cookie instead of a seperate one:
  - [ ] Wait for #8657
  - [ ] Add route for changing sessions' locale
  - [ ] Redirect to locale route on locale change
- [ ] Improve language button design

### Language select button
language_select_button

### Language select modal
[(old, with flags)](https://user-images.githubusercontent.com/9089004/45028304-d0137180-b044-11e8-9f4f-45ef39be2390.png)

### GIF
[(old, with flags)](https://user-images.githubusercontent.com/9089004/45029310-8c226b80-b048-11e8-9db4-b7c3b2ef6b49.gif)

If merged, resolves #5691.

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -7,6 +7,12 @@ class AboutController < ApplicationController
   def show
     serializable_resource = ActiveModelSerializers::SerializableResource.new(InitialStatePresenter.new(initial_state_params), serializer: InitialStateSerializer)
     @initial_state_json   = serializable_resource.to_json
+    @language_selector_props = {
+      locale: I18n.locale,
+      supportedLocales: I18n.available_locales.map do |locale|
+        { code: locale, name: SettingsHelper::HUMAN_LOCALES[locale] }
+      end,
+    }.to_json
   end
 
   def more
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -11,6 +11,7 @@ module Localized
 
   def set_locale
     I18n.locale = default_locale
+    I18n.locale = cookies[:lang]
     I18n.locale = current_user.locale if user_signed_in?
   rescue I18n::InvalidLocale
     I18n.locale = default_locale
--- a/app/javascript/mastodon/components/language_select.js
+++ b/app/javascript/mastodon/components/language_select.js
@@ -1,0 +1,31 @@
+import React from 'react';
+import { ImmutablePureComponent } from 'react-immutable-pure-component';
+import PropTypes from 'prop-types';
+
+import IconButton from './icon_button';
+
+export default class LanguageSelect extends ImmutablePureComponent {
+
+  static propTypes = {
+    locale: PropTypes.string.isRequired,
+    supportedLocales: PropTypes.arrayOf(PropTypes.object).isRequired,
+    onLocaleChange: PropTypes.func.isRequired,
+  };
+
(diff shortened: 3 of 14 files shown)


Sylvhem:
This is a really great feature, thank you for your work on this! However, I have two observations.
First, I would suggest to not use country flags in the language selector. Most languages are used in more than one country and most European languages have been been imposed to the rest of the world by colonialism and imperialism. For example, I don't think people from Burkina Faso don't want to be reminded of the French colonization every time they try to switch language.
In addition, I wanted to know if you though how this would interact with a possible implementation of #8428?

nightpool:
it's completely independent from 8428, since one is about localizing
strings and one is about selecting which localization to use.

I like this, but I think the language button is pretty hard to notice.
maybe we can make it display the name of the current language, as well as
being a little more button-like?

Gargron:
Thoughts:

- Do not use flags (lesson from joinmastodon.org language picker feedback)
- Use `session` instead of separate cookie

I agree that visually it's maybe not in the right place. The about page could use the new footer design, and the new footer design could include a language switch.

JMendyk:
Thanks for your opinions!

At some point I realised that flags while being a nice addition to find your _country's_ language easier will probably not work out. I decided to let them stay for discussion's sake.

Regarding #8428, I haven't seen this issue before. They don't really seem related (at least on implementation level) but I think that multi-language rules would be a nice addition to Mastodon.

I'll look into making language button more button-like with current locale's name to make it stand out more.

I initailly wanted to store locale in session but if I'm not wrong it requires creating new route which will be responsible for changing locale. I haven't found an already existing controller which would make sense to handle it nor did I want to create a new one. 

@Gargron I'll wait for your #8657 with storing locale in session.

With new footer design, button could be moved there. Or if new footer design is planned for not-too-far future, I could remove the button completely from about page and language selection functionality would be exposed with the arrival of the new footer.

PR #8619, opened 2018-09-05
config/boot is required first

abcang:
If tootctl is executed without going through `bundle exec`, loading `thor` will fail. By loading `config/boot` first, tootctl can be executed without going through `bundle exec`.


Gargron (approved):
Ah

PR #8627, opened 2018-09-05
feat(auth/session_controller): Send Clear-Site-Data when logging out

sorin-davidoi:
Will clear the browser's cache, cookies and storage.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data
https://w3c.github.io/webappsec-clear-site-data/


Gargron (approved):

ThibG:
I'm not too sure about this… some people may log out and log back in with different accounts on the same instance. Loosing cache and hashtag history might be a pain in this case

Gargron:
@ThibG It's probably negligble

PR #8642, opened 2018-09-07
tootctl accounts [add|del|cull]

Gargron:
```
Usage:
  tootctl accounts add USERNAME --email=EMAIL

Options:
  --email=EMAIL                    
  [--confirmed], [--no-confirmed]  
  [--role=ROLE]                    
                                   # Default: user
  [--reattach], [--no-reattach]    
  [--force]                        

Description:
  Create a new user account with a given USERNAME and an e-mail address provided 
  with --email.

  With the --confirmed option, the confirmation e-mail will be skipped and the 
  account will be active straight away.

  With the --role option one of "user", "admin" or "moderator" can be supplied. 
  Defaults to "user"

  With the --reattach option, the new user will be reattached to a given 
  existing username of an old account. If the old account is still in use by 
  someone else, you can supply the --force option to delete the old record and 
  reattach the username to the new account anyway.
```

```
Usage:
  tootctl accounts del USERNAME

Description:
  Remove a user account with a given USERNAME.
```

```
Usage:
  tootctl accounts cull

Options:
  [--dry-run], [--no-dry-run]  

Description:
  Query every single remote account in the database to determine if it still 
  exists on the origin server, and if it doesn't, remove it from the database.

  Accounts that have had confirmed activity within the last week are excluded 
  from the checks.

  If 10 or more accounts from the same domain cannot be queried due to a 
  connection error (such as missing DNS records) then the domain is considered 
  dead, and all other accounts from it are deleted without further querying.

  With the --dry-run option, no deletes will actually be carried out.
```

### DIFF
--- a/lib/mastodon/accounts_cli.rb
+++ b/lib/mastodon/accounts_cli.rb
@@ -43,6 +43,81 @@ module Mastodon
       end
     end
 
+    option :email, required: true
+    option :confirmed, type: :boolean
+    option :role, default: 'user'
+    option :reattach, type: :boolean
+    option :force, type: :boolean
+    desc 'add USERNAME', 'Create a new user'
+    long_desc <<-LONG_DESC
+      Create a new user account with a given USERNAME and an
+      e-mail address provided with --email.
+
+      With the --confirmed option, the confirmation e-mail will
+      be skipped and the account will be active straight away.
+
+      With the --role option one of  "user", "admin" or "moderator"
+      can be supplied. Defaults to "user"
+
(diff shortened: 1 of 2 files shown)


Gargron:
@ashleyhull-versent It generates a random password which is output to the CLI. The user is supposed to change the password once they login.

ashleyhull-versent:
That's perfect from an interactive point - I'm mostly thinking about my AWS CFN automation.

happy with that tho.

ykzts (approved):

PR #8657, opened 2018-09-09
Change locale detection to run once per session

### LINKED ISSUE #6462: Wrong language used on the two factor code screen
Both my main instance's default interface language and my browser's default language are set to French. I'm using two factors authentication to log in.
Normally, when logging to Mastodon, the screen used to enter my two factor code is in French.
However, sometimes it's in English, in German or in Chinese.
Mastodon's two factor code screen in English
Mastodon's two factor code screen in German
Mastodon's two factor code screen in Chinese

My instance, [Witches Town](https://witches.town/), is currently running on [Mastodon 2.2.0rc2](https://github.com/tootsuite/mastodon/releases/tag/v2.2.0rc2), but I noticed this behavior since quite some times now. Other releases are probably also affected.
My web browser is Mozilla Firefox ESR 52.6.0 (64 bits) and my OS is Debian GNU/Linux Stretch. My browser isn't configured to use German nor Chinese as fallback languages.
* * * *

### LINKED ISSUE #11157: When many UI errors are generated, the language switches.
### Expected behaviour

The language stays the same as normal.

### Actual behaviour

The language occasionally changes.

Cause a bunch of errors (say, 429 Throttled)
Screenshot 2019-06-22 at 2 02 02 PM

### Specifications

v2.9.2, Chrome OS, 75.0.3770.88

Gargron:
Fix #6462, Fix #11157

Prevent locale leak between requests by using `I18n.with_locale`. For logged out users, the locale is set once on first request, then it's a session variable. This also opens up a possibility for locale switching to be implemented.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -138,11 +138,7 @@ class ApplicationController < ActionController::Base
   def respond_with_error(code)
     respond_to do |format|
       format.any  { head code }
-
-      format.html do
-        set_locale
-        render "errors/#{code}", layout: 'error', status: code
-      end
+      format.html { render "errors/#{code}", layout: 'error', status: code }
     end
   end
 
--- a/app/controllers/concerns/localized.rb
+++ b/app/controllers/concerns/localized.rb
@@ -4,16 +4,19 @@ module Localized
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_locale
+    around_action :set_locale
   end
 
   private
 
   def set_locale
-    I18n.locale = default_locale
-    I18n.locale = current_user.locale if user_signed_in?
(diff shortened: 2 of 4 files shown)


nightpool:
this is mmmmaybe bad if people change their locale in the browser? it would
be pretty surprising to them to not have it change on mastodon as well, if
it was set erroneously in the first place.

JMendyk:
Maybe default locale could be stored along with current timestamp in `session[:locale]` and, if too old, invalidated and loaded again. This way we would limit calls to `http_accept_language` and make it respond to locale change soonish after it happened.

Gargron:
I don't think people generally update their browser language preferences... Maybe once.

ykzts (approved):

PR #8673, opened 2018-09-10
Highlight active tab in action bar 

rhoio:
bildschirmfoto 2018-09-10 um 23 48 05


rhoio:
Didn't know that there existed bars in the public view, so I put it where they looked best for me.

Changed to border-bottom and same pixel width for consistency.

ClearlyClaire (approved):

PR #8674, opened 2018-09-10
Group reports by target account

Gargron:
image

Accounts used in screenshot only for illustration purposes!


ashleyhull-versent:
Having the ability to "minimalise" or show/hide an account.. incase one account gets 1000+ reports and I'd want to hide it to find the other reports would be cool.

ThibG:
I like the idea of grouping reports by target account, but without the table labels, it is slightly less clear who is the reporter and who is the assigned admin. It also seems to take more space per report?

Gargron:
It's especially confusing in the screenshot because the admin account is both, but I think in practice seeing mods' names on the right will make it obvious.

It does take more space per report but I think it's balanced as to be clearly readable and skimmable.

ThibG:
I still think that having no label whatsoever for those columns is confusing.

Also, I'm also wondering whether the paginating makes this too confusing?
Like, if there are very old and very new reports targeting one person, and enough reports in between to make them appear on different pages, you'd think that person would have only been reported recently, unless you go back to the old pages.

Issue #5916, opened 2017-12-07
Replies to users outside of a list appear on the list


renatolond:
I'm on 2.1.0.rc1 and I've added some people to a list. When they answer to people I follow but are not in the list, the replies appear on the list. I expected that only replies to other members of the list would appear. Currently, I can only see the reply and not the original toot in the list, without clicking to expand the toot.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

chriswmartin:
fwiw this was expected behavior for me

saper:
But if one does expand the reply a whole conversation is visible?

cbayerlein:
Yes, in the single view of posts the behavior is as expected.

It's really just about the semantics in the lists. Personally, I find it irritating to find individual shreds of conversation without the appropriate context. The expectation from the user's perspective is that these posts are hidden. This also corresponds to the usual behavior with other, comparable social networks - e. g. the one with the bird :-)

saper:
It is also possible that people want to have those scraps on the lists, too. 

On Pleroma one can see replies on the timeline as well. I find it very useful and I manually expand interesting conversations. I can interact more because of this.

(Regarding the implementation, I am not sure it works like this - most probably they are extracted in one go, but I haven't checked yet).

cbayerlein:
I agree with @renatolond . In my list, I have some people who tend to use Mastodon for casual chatting with others. Then, I always get only half of the conversations, which is too less to understand it and too much to just fly over it. I could also add the usual conversation partners of the list members, but then the list gets even more filled up and even less easy to read.

So I think it depends on how you use lists. The best solution would be to provide an user option (best by list).

But for now, I'd love to implement that suggested behaviour on my own instance, to provide some "proof of concept"; For me it would also be a nice learning tutorial. But as I've written on discourse, I would need some help and guidance, especially in the beginning.

cbayerlein:
Just had a brief look into the code. It seems that the right starting point would be the FeedManager (app/lib/feed_manager.rb) : 

`def push_to_list(list, status)
    return false unless add_to_feed(:list, list.id, status)
    trim(:list, list.id)
    PushUpdateWorker.perform_async(list.account_id, status.id, "timeline:list:#{list.id}") if push_update_required?("timeline:list:#{list.id}")
    true
[...]

PR #8683, opened 2018-09-12
Exclude replies from list timelines

cbayerlein:
A small tweak in list behaviour: Now all posts that are replies are filtered out, so that now only "genuine" posts are displayed in the list.

Of course this is a bit sub-optimal in that it would be better to have a certain subset of these replies - namely those on posts whose authors are also list members. But filtering that requires a bit more complex logic.

One would have to find the post to which the reply relates. Then seek the author of this post and then a comparison, if they are a list member.

The best way would be to provide the user with a setting which gives them an option to choose whether the want the "old" or "new" behavior.

This is a first approach to solve issue #5916

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -40,6 +40,7 @@ class FeedManager
   end
 
   def push_to_list(list, status)
+    return false if status.reply?
     return false unless add_to_feed(:list, list.id, status)
     trim(:list, list.id)
     PushUpdateWorker.perform_async(list.account_id, status.id, "timeline:list:#{list.id}") if push_update_required?("timeline:list:#{list.id}")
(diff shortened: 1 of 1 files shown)


app/lib/feed_manager.rb
```diff
@@ -40,6 +40,7 @@ def unpush_from_home(account, status)
   end
 
   def push_to_list(list, status)
+    return false if status.reply?
```
Gargron: I suggest allowing self-replies.

`return false if status.reply? && status.in_reply_to_account_id != status.account_id`

Gargron (changes requested):

app/lib/feed_manager.rb
cbayerlein: Ok, I totally agreee :)

Gargron (approved):

PR #8691, opened 2018-09-13
Fix building assets in test environment

ClearlyClaire:
Webpacker (Ruby wrapper to webpack) uses RAILS_ENV-based environments while
the javascript configuration for webpack re-reads this configuration file using
the NODE_ENV environment variable. This means that when RAILS_ENV=test, running
“assets:precompile” compiled the production packs in “public/packs” while
webpacker expects them in “public/packs-test”. This causes Ruby to recompile
them on-the-fly, possibly leading to race conditions in parallel_tests.

This changes:
- Disables on-the-fly compilation in test environment
- Changes the javascript part to read the correct environment

### DIFF
--- a/config/webpack/configuration.js
+++ b/config/webpack/configuration.js
@@ -7,7 +7,7 @@ const { readFileSync } = require('fs');
 
 const configPath = resolve('config', 'webpacker.yml');
 const loadersDir = join(__dirname, 'loaders');
-const settings = safeLoad(readFileSync(configPath), 'utf8')[env.NODE_ENV];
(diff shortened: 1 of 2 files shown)


Gargron (approved):

ykzts (approved):
nice!

nightpool:
is there any reason we WOULDN'T want to always have RAILS_ENV == NODE_ENV?

ThibG:
@nightpool according to https://github.com/rails/webpacker#custom-rails-environments that's a common usage pattern. Other than that, we have a bunch of tests that check whether NODE_ENV is strictly equal to “development” or “production” but nothing else. So I imagine that having NODE_ENV = "test" might break things. It could also break things in dependencies I guess.

PR #8693, opened 2018-09-13
[WIP] notify users after admin actions

devkral:
Users should know if their posts were updated by an admin.
This patchset will lay the foundation by sending users messages if their posts are modified.

Maybe instance local messages from Admin can be integrated later.

### DIFF
--- a/app/controllers/admin/reported_statuses_controller.rb
+++ b/app/controllers/admin/reported_statuses_controller.rb
@@ -16,11 +16,11 @@ module Admin
     private
 
     def status_params
-      params.require(:status).permit(:sensitive)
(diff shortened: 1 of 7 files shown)


Gargron:
Implementing this is on my roadmap for 2.6 but this is probably not how I would go about it. I intend to add an e-mail template and a way to enter a custom message for the e-mail, but I do not intend to save that message in the database, nor to use the notifications API for it. ActionLog was not indended to be connected to the REST API or be user-facing, and fwiw ActionLog currently has design issues when it comes to storing changesets, it's not very elegant...

m4sk1n:
closed by #9519?

Gargron:
Indeed

PR #8702, opened 2018-09-15
enable counters for all status actions

lfuelling:
This PR adds counters for all status actions (fav, reblog, reply) and removes the obfuscation ("1+") for them.

Please ignore the theming in the screenshot below.
Counters for status actions

**Also available as a [`.patch` file](https://gist.github.com/lfuelling/6ba7f49aaca02581ede3e4d887683de8) (also modifies the detailed status).**

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -32,16 +32,6 @@ const messages = defineMessages({
   embed: { id: 'status.embed', defaultMessage: 'Embed' },
 });
 
-const obfuscatedCount = count => {
-  if (count < 0) {
-    return 0;
-  } else if (count <= 1) {
-    return count;
-  } else {
-    return '1+';
-  }
-};
-
 export default @injectIntl
 class StatusActionBar extends ImmutablePureComponent {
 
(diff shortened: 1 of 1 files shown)


Gargron:
There's reasons it is as it is. Thank you for the thought but I consider this patch unnecessary.

lfuelling:
It would be nice to know the reasons. 

Anyway, thanks for the reply.

trwnh:
@lfuelling Always-visible counters = subconscious prioritization of posts depending on the size of the response. There's an inherent value judgement in attaching a number to a post by default and with emphasis, as opposed to hidden in the metadata -- see: "getting ratio'd" on Twitter, where controversial posts receive more replies than retweets and likes and are subsequently mocked, leading to dogpiles where people try to raise "the ratio" as much as possible. Alternatively, a post that receives an overwhelming number of boosts and favs may be seen as more popular or more correct, due to psychological priming.

There's also the fact that interaction counts are calculated per-server rather than globally, so they are by definition inaccurate and not really worth promoting so heavily. The 0/1/1+ counters are there to indicate the _presence_ of replies and not necessarily how many of them there are; you can tell at a glance whether a question may have been answered, or if the post is the start of a thread or replies, or so on.

lfuelling:
Then why add the reply indicator? You could apply the same points to that and it looks terrible without text next to the other buttons. Following your arguing, it's unnecessary to know if there are replies until you are interested into a particular post. (This could also apply to the "this is a reply"/double arrow indicator, but at least it's useful and looks good.)

trwnh:
@lfuelling I'm personally in favor simplifying the 0/1/1+ notation to a notification-dot-like icon or a +, in a similar way to the double-arrow, or perhaps even repurposing the double-arrow into a general-purpose "this post is part of a thread" indicator. The 0/1/1+ is currently a compromise that allows people to know the difference between 1 or arbitrarily-many replies. The main concern is overloading the semantic meaning of the reply icon -- a double-arrow can indicate conversation or multiple participants (i.e. "reply all"); adding dots or pluses to the mix could make that icon too visually crowded.

In the case of questions and answers, it's obviously not a judgement on whether they are correct, just that they exist and that you should maybe read them instead of replying blindly and giving the same answer that several other people have given before you.

PR #8703, opened 2018-09-16
Redesign forms, verify link ownership with rel="me"

### LINKED ISSUE #121: No URL in profile

### LINKED ISSUE #3821: Per-instance verified users/checkmarks
A checkmark system solves a lot of hazards with identity theft and fake accounts. Adding decentralized, per-instance verification would be really nice.

* * * *

### LINKED ISSUE #7133: Redesign forms
The underline-only no-background design where placeholders are used over labels is not looking very good anymore. Especially when the inputs are filled it's hard to tell what is what without remembering. Also, the padding on the inputs that has to be there because of the background tint on focus makes the text not line up with other text, which is annoying.

### LINKED ISSUE #8276: Verified users
Mastodon should offer a way to support verified users. See [that toot of Wil Wheaton](https://mastodon.cloud/@wilw/100572583503071267): He uses his verified Facebook account to verify that a specific Mastodon user belongs to him. That's obviously a subpar solution.

Years ago app.net had a [html based validation for your website](https://3manfactory.co.uk/2013/05/01/app-net-introduces-domain-verification/) and domain validation (with text entries added to your dns). You created the code in your profile and added it to the website or dns server and you got a little checkmark in your profile and it also showed which domain was used to validate that account.

* * * *

Gargron:
**Summary:**

- Redesign forms to
    - Make better use of space horizontally
    - Use labels instead of placeholders
    - Add quick navigation to preferences page
    - Reorder site settings thematically
    - Improve how settings pages look on mobile
- Add link ownership verification via `rel="me"`

> **How does link verification work?** When a profile is updated, the local as well as remote servers (independently) check the links for the presence of a link-back with `rel="me"` attribute to the user's profile. If so, a timestamp of the check is saved.
>
>**Why is `rel="me"` required?** Because intent matters. It is much easier to sneak a plain link on someone's website. For example, in a WordPress comment, or through a Mastodon boost, or a Tumblr reblog. However, custom rel attributes are usually not allowed in those. On the other hand, profile links set by the owner of a page can and do contain this attribute.
>
>**What does this give us?** This is not a silver bullet. Nothing beats checking someone's government identification documents. However, that approach is closed off to federated platforms. It would not scale if done properly, it would be awful for privacy, and the most naive approach would be subject to trivial spoofing. So, this is the next best thing. If you have a personal website that people know is official, proving the ownership of that site is the same as proving your identity.
>
>**Why aren't you using DNS cryptography?** An HTML link is the lowest possible barrier to entry. Editing DNS and messing with public keys would exclude a lot of, if not most non-technical people, and as such I do not consider those approaches worth the implementation time right now.

**References:**

- Fix #7133
- Fix #121

To some extent:

- Fix #8276
- Fix  #3821

**Verified links:**

image
image

**Forms redesign:**

image
image
image
image
image
image
image
image
image
image
image
image


lfuelling:
A TXT record for alternative verification would be nice. For example I wouldn't like having to add a link to my profile on the index of my GitLab instance and stuff like that.

Verifying multiple profiles is already possible with the current method. There should also be another way to confirm multiple accounts using DNS. (comma separated links?)

Gargron:
The complexity with doing DNS verification comes from the fact that each link needs a special verification string, so the field must be saved before that string can be given to the user for copypasting, which means the actual verification must be done in a yet another step (+ wait for DNS propagation)

Don't forget that each server must verify on its own. I don't see how the user could have any control over *when* the checks happen, currently it's just tied into when the profile has been updated.

lfuelling:
Why are profile links not okay for DNS? 
I'd expect it to be harder to add a DNS record to one of my domains than injecting HTML into one of my pages.

sivy:
I think this is a pretty good solution! Not complicated UI-wise, and doesn't require the user's browser to do the validation work.

sivy:
@Gargron: how does the link verification get triggered?

sivy:
"ownership of this link was checked..." might be better as:

> Ownership of the linked site by this user was verified on [date]

with the negative case being:

> Ownership of the linked site by this user could not be verified as of [date]

Alternatively, no messaging on unverified links might also be better, so that no negative connotations are applied to links that are not intended or desired to be verified.

I'm still not sure what I think about using "verification" language in this feature, given the baggage it carries from birdsite, but it is terminology that users understand.

app/services/verify_link_service.rb
```diff
@@ -0,0 +1,29 @@
...
+  def perform_request!
+    @body = Request.new(:get, url).add_headers('Accept' => 'text/html').perform do |res|
+      res.code != 200 ? nil : res.body_with_limit
+    end
+  end
+
+  def link_back_present?
+    return false if @body.empty?
+
+    Nokogiri::HTML(@body).xpath('//link[@rel="me"]').any? { |link| link['href'] == @link_back }
```
measlytwerp: Shouldn't the XPath read `//link[@rel="me"]|//a[@rel="me"]`? The copy you've written mentions `<a rel="me"` but this won't match that.

Gargron: My bad, it should be `a` instead of `link`, not sure if `link` is worth checking at all, that would be a real rarity

lanodan: True, both ``link`` and ``a`` are valid for ``rel=me``.

ThibG: I initially thought it was a “link” attribute and not a “a” attribute. Can't find the spec (if any) right now, but if link's in it, you should restore it, even if it “would be a real rarity”.

Gargron:
@ThibG Okay, both anchor and link tags will be checked.

ykzts (approved):

ChameleonScales:
How long should we expect the ownership verification to take?

ChameleonScales:
I deleted my link, saved, re-added it and re-saved and the green highlighting appeared instantly.
So I guess it only works at the moment you add the link in your profile settings.

tasn:
I think it would be a good idea to add a 2-hop check?
What do I mean?
If a page has `rel=me` pointing to another page that's not a mastodon account, check if that page has `rel=me` to the mastodon account.

Use case?
Many websites such as Twitter and Github mark your homepage as `rel=me` and only let you add one such link to your profile. This means a Mastodon user looking to verify their Github page would need to change the e.g. Twitter URL to their mastodon account (instead of e.g. their home page) which is undesirable.
However, with the suggested change people can keep these links pointing to the user's website and have verification work.

I've encountered this issue with EteSync. Relevant links to see what I'm talking about:
https://www.etesync.com
https://mastodon.social/@EteSync
https://github.com/etesync
https://twitter.com/EteSyncHQ

Gargron:
@tasn I suggest opening a separate feature request for this!

tasn:
Done. #11391

Sorry for the noise.

PR #8708, opened 2018-09-17
add several missing alt tags

jacobherrington:
Noticed a few missing alt tags!


ykzts:
I think that mascot image does not need to read speech in general.

I think that it should be using `alt=""`.

jacobherrington:
@ykzts fair enough, changes made!

ykzts (approved):

Gargron (approved):

PR #8714, opened 2018-09-17
Unconditionally re-encode locally-uploaded images to strip metadata

ClearlyClaire:
This strips metadata on file upload by re-encoding the files, at the cost of image quality and processing resources.

### DIFF
--- a/lib/paperclip/lazy_thumbnail.rb
+++ b/lib/paperclip/lazy_thumbnail.rb
@@ -20,7 +20,7 @@ module Paperclip
     private
 
     def needs_convert?
-      needs_different_geometry? || needs_different_format?
+      needs_different_geometry? || needs_different_format? || needs_metadata_stripping?
     end
 
     def needs_different_geometry?
@@ -30,6 +30,10 @@ module Paperclip
 
     def needs_different_format?
       @format.present? && @current_format != @format
+    end
+
+    def needs_metadata_stripping?
+      @attachment.instance.respond_to?(:local?) && @attachment.instance.local?
     end
   end
 end


nightpool:
1) this means that
https://github.com/tootsuite/mastodon/blob/master/lib/paperclip/lazy_thumbnail.rb
is now dead code, and we should remove it.

2) I'm not convinced this PR is the best solution. is it possible to strip
metadata without reencoding the file? what are the benchmarks on each?

ThibG:
@nightpool 

1. Yup, that is dead code, but I kept it because 2.

2. I'm trying to find solutions that don't sacrifice processing time or image quality, but it's not possible out of the box with Paperclip/ImageMagick

ThibG:
Updated to keep `LazyThumbnail` as it has its own resizing logic. However, despite the name, it's not lazy anymore.

ThibG:
The only tool I have found to strip image metadata is `exiftool` (there is also `mat2`, which uses `exiftool` for images)… however, this is a perl dependency that is pretty uncomfortable to use…

elomatreb:
ImageMagicks in-place component can strip metadata without reencoding, `mogrify -strip image.jpg`

ThibG:
@elomatreb `mogrify` still re-encodes, even if it does so in-place

elomatreb:
Mh, I could have sworn it was smart enough to not touch the image if it doesn't need to. Nevermind then.

Gargron:
Keep in mind that lazy thumbnail is a result of #6515, which fixed all sorts of visual glitches with GIF avatars from remote users. We don't really care about EXIF metadata in remote images because presumably they already stripped it themselves when the image was turned into the dimensions we expect.

Reverting that fix means not only more computational overhead for ImageMagick stuff (which I consider acceptable because we already dealt with that pre-optimization, not the end of the world), but unfortunately it means the return of something like #3199 (I am not sure if all glitches had corresponding bug reports), which itself is an ImageMagick bug as far as I understand.

The fix in question became a problem specifically in #8083, when the lazy thumbnail processor was hooked up to media attachments. EXIF data in avatars and headers is unlikely to contain sensitive information, but media attachments may obviously come straight from a camera. This is my bad, I had not connected the dots between the different patches.

I suggest that we keep the behaviour for remote images (through some kinda option passed to the processor?) but unconditionally re-encode locally uploaded images.

ThibG:
@kitlith thanks, exiv2 looks pretty much like what I want! I'm slightly concerned about security, though, that's yet another binary we're going to feed user-provided images to (the same can be said about exiftool or imagemagick, though)

ThibG:
@Gargron yeah, that's one of the things I'm looking at, but so far I don't know how to process remote and local attachments differently

Gargron:
@ThibG The processor can call methods on the model, so maybe something like `attachment.instance.respond_to?(:remote?) && attachment.instance.remote?`

ThibG:
@Gargron updated to do just that!

Gargron (approved):

PR #8719, opened 2018-09-18
Fix performance regression in Account::Field#verifiable?

Gargron:
Regression from #8703

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -323,7 +323,7 @@ class Account < ApplicationRecord
     end
 
     def verifiable?
-      value.present? && (value.start_with?('http://') || value.start_with?('https://'))
+      value.present? && value.start_with?('http://', 'https://')
     end
 
     def mark_verified!
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -323,7 +323,7 @@ class Account < ApplicationRecord
     end
 
     def verifiable?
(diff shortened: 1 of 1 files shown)


malicioustoker:
Bump

ClearlyClaire (approved):
Minor improvement possible according to CodeClimate, but I'm otherwise ok with it.
I wonder if we need `verifiable?` to be used at all when serializing, though. A `null` `verified_at` would be fine, right?

ThibG:
@malicioustoker this is very unlikely: this is about an inefficiency in serializing any account, regardless of the timeline it appears in

Gargron:
@malicioustoker @ThibG I have experienced local timeline slowness on one instance I ran for someone, but that was pre-2.5.0. The PostgreSQL server used a HDD and not an SSD on that machine, that's all that jumps to mind. There might be some working memory related issue where the local timeline query doesn't fit in there and so is unusually slow.

Gargron:
I recommend playing around with this tool for your PG configuration: https://pgtune.leopard.in.ua/ Maybe it will help.

malicioustoker:
Oh wow, great, thanks! I’ll report back after playing around with it. I really appreciate it!

-- Reverend Glen

malicioustoker:
All the changes have been made, yet it made no difference with the long load time for the local timeline :-/

-- Reverend Glen

PR #8734, opened 2018-09-19
Switch from puma to falcon

ClearlyClaire:
Puma's worker-based model may not be too suitable for Mastodon, which frequently depends on external services. [Falcon](https://github.com/socketry/falcon/) seems to be a promising async server. Furthermore, it seems to handle Websocket, so we could eventually remove the uws-based node.js streaming server.

On the other hand, it does not seem to be widely used, and the documentation is a bit lacking.

I have not performed any benchmark yet.


Gargron:
I'm not sure if it's wise to trust a new and unpopular server for something as important as Mastodon. What if there's bugs and the original developer is unavailable? We'll have to become falcon developers ourselves? What if there's bugs that nobody noticed because nobody was using it at scale until now? This is a big leap of faith.

ThibG:
Yes, I understand your concerns, which is why I plan to test this for a while before removing the WiP tag.
You know perfectly well why I want to move away from puma, though.

nightpool:
>You know perfectly well why I want to move away from puma, though.

i'm not sure I follow this. can you be a little less cryptic?

nightpool:
I agree with gargron that this sounds like something that would be better for individual testing/in a fork until we're pretty confident that it's production-ready

ThibG:
There are a few places where we use `RedisLock`, I guess this could lead to deadlocks.
There are also a few places where we make HTTP queries without using our `Request` class, which limits the benefits of switching to an asynchronous web server.

ggtea:
I think that async-ness is one of the most desired thing on task-queue(Sidekiq) which is stalled often by slow response from remote servers.
Rack-servers(Puma) requires remote HTTP request (which takes long time) only in small case (e.g. explicitly search remote Account), as I know.
So, async efforts are small for Rack-servers. I think.

More, I wonder:
Is mixing ordinary REST and Web-Sockets(C10K) on same server safe? Isn't it going to make management cost bigger?
Is Ruby fiber based rack implementation performant like nodejs based servers when handling C10K?

ThibG:
@ggtea there are lots of cases where puma performs HTTP queries: URL-based searches, media proxy, inbox whenever the remote account is unknown or the signature is invalid…

Sidekiq jobs are indeed also an issue, but a somewhat of a less critical issue I think.

I don't know how performant falcon is in dealing with websockets, I haven't even tried porting the streaming server yet. As for safety, we could also run two different server processes, but having everything written in Ruby and not requiring a dead lib (uws).

ggtea:
@ThibG 
I confirmed Puma does really weird task...
URL-based searches: This is what I expected.
media proxy:
 It seems media should be re-downloaded before they get appeared in TL.
 Proxy url on TL causes clients send request to proxy in sync. it may be critical.
inbox: really weird. I know inbox should return correct result, so webfinger must be performed on Puma and it's unavoidable...

nightpool:
image

nightpool:
i actually kind of like this idea

ThibG:
I dislike this idea for several reasons:
- It forces the sending instance to manage a retry queue (which may be good practice in general, but is not currently that widespread, and is more work)
- This just moves slow requests from synchronous puma to synchronous sidekiq, which queue can fill up just as fast (this does have the advantage of not blocking puma itself, though, so even though federation jobs will be stalled, some interaction is still possible)

Gargron:
> I agree with gargron that this sounds like something that would be better for individual testing/in a fork until we're pretty confident that it's production-ready

Has anyone been running this in production? I am still a little torn because due to lower popularity, Falcon may not receive security reports / updates as quickly as Puma, but I am interested in terms of improving performance...

ThibG:
Oops, I have stopped running this, not because it didn't work, but because I hadn't the energy to continuously rebase those changes.
Also, this PR is still missing asynchronous ActiveStorage support, which means while most requests will have comparable or better performances, PostgreSQL-bound ones will halt more concurrent requests.

PR #8736, opened 2018-09-19
Support min_id-based pagination in REST API

### LINKED ISSUE #5819: Pagination controls aren't particularly useful for position-remembering clients.
I'm trying to update Amaroq to remember timeline position between restarts or view refreshes. As such, I was hoping to be able to use the `max_id` and `since_id` parameters to control which toots we get back from the API.

However, these parameters don't seem to to work quite the way I expected. They are effectively unusable for this "remember timeline position" use case and we would have to resort to a sequential spam of the timeline API to get enough toots to get us down to the remembered one.

Take a peek at the head of the timeline.
```
http 'https://toot.cafe/api/v1/timelines/home?limit=10' 'Authorization: Bearer foobar' | jq '.[] | .id'
"99067141923265252"
"99067078945904879"
"99067065159308178"
"99067009582024915"
"99066983051474906"
"99066982760931898"
"99066949753677356"
"99066901008261781"
"99066834960670058"
"99066422645335901"
```

Pretend like we're starting up and we remember having seen 99066949753677356 at the top of our timeline:
```
http 'https://toot.cafe/api/v1/timelines/home?limit=3&max_id=99066949753677356' 'Authorization: Bearer foobar' | jq '.[] | .id'
"99066901008261781"
"99066834960670058"
"99066422645335901"
```

**Issue 1** note that the max_Id isn't actually in the list of returned statuses. So if we wanted to show a view of all the things that the user was looking at before and we stored the id of the topmost toot, we would somehow still need to get one more toot from the timeline to restore the previous state.

Now lets say we want a page of toots since the topmost one we're currently seeing:
```
http 'https://toot.cafe/api/v1/timelines/home?limit=3&since_id=99066901008261781' 'Authorization: Bearer foobar' | jq '.[] | .id'
"99067141923265252"
"99067078945904879"
"99067065159308178"
```

**Issue 2**: instead of getting the next three messages after the one we asked for, we get three messages that are after the one we asked for, starting with the head of the timeline. So to get the actual "previous" page, we'd have to page through a bunch of requests.

I guess what would be nice is a new pair of parameters like

* `up_to_id=[id]` which returns a page of results with the first one having id `[id]`
* `after_id=[id]` which returns a page of results with the last one being the exact next toot after (newer than) `[id]`.

* * * *

Gargron:
Fix  #5819

- [x] Home
- [x] List
- [x] Account statuses
- [x] Public
- [x] Tag
- [x] Notifications
- ~~~Direct~~~

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -53,6 +53,10 @@ class Api::BaseController < ApplicationController
     [params[:limit].to_i.abs, default_limit * 2].min
   end
 
+  def params_slice(*keys)
+    params.slice(*keys).permit(*keys)
+  end
+
   def current_resource_owner
     @current_user ||= User.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
   end
--- a/app/controllers/api/v1/accounts/statuses_controller.rb
+++ b/app/controllers/api/v1/accounts/statuses_controller.rb
@@ -81,7 +81,7 @@ class Api::V1::Accounts::StatusesController < Api::BaseController
 
   def prev_path
     unless @statuses.empty?
-      api_v1_account_statuses_url pagination_params(since_id: pagination_since_id)
+      api_v1_account_statuses_url pagination_params(min_id: pagination_since_id)
     end
   end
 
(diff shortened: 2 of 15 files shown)


ykzts (approved):

DagAgren:
Could you let me know when this goes live so I can test against it?

Also, what release version includes it so I can set up my version gate.

PR #8746, opened 2018-09-21
Adjust beforeUnload conditions

### LINKED ISSUE #8701: "Leave Site" warning a little too sensitive
Sometimes, when I first navigate to my local Masto instance on the web, I want to scroll down. I hit the space bar, but this ends up putting a space into the "What is on your mind?" new-toot text entry field.

When I'm done scrolling the timeline with my scroll wheel and then close the tab or go to do something else, I get the "Leave site? Changes you made may not be saved." warning.

If it's a busy day or I'm particularly tired, I have to cancel and go check what the unsaved changes I made are, because the new-toot text entry field is still blank (although the memo/hint text is now gone, I'm also often scrolled over to the right to read my timeline), and I didn't think that I actually made a "change," so much as I hit space and the UI didn't scroll.

Even if I backspace everything in the text field and then try to exit, the warning persists, even though the "Toot" button is now disabled.

It would be nice if the dirty flag logic were to be updated so that 'blank' or otherwise meaningless toots didn't count as an actual unsaved change.

At the very least, deleting everything out of the toot box (and resetting it to the "original" state) should reset the dirty flag.

Something even better would be is: if the first keystroke into the page were a space, it scrolls instead of enters text into the toot box. I don't think many people are leading their toots with an empty space character anyway.

Thanks for the awesome project! I love using Mastodon.
* * * *

ClearlyClaire:
Fixes #8701

### DIFF
--- a/app/javascript/mastodon/features/ui/index.js
+++ b/app/javascript/mastodon/features/ui/index.js
@@ -59,7 +59,8 @@ const messages = defineMessages({
 
 const mapStateToProps = state => ({
   isComposing: state.getIn(['compose', 'is_composing']),
-  hasComposingText: state.getIn(['compose', 'text']) !== '',
+  hasComposingText: state.getIn(['compose', 'text']).trim().length !== 0,
+  hasMediaAttachments: state.getIn(['compose', 'media_attachments']).size > 0,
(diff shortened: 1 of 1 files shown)


ykzts (approved):

barbeque:
Thank you very much!

PR #8756, opened 2018-09-22
Adds follow action timestamp to notification

### LINKED ISSUE #8755: Show timestamp of a follow notification
When I view a follow notification, I do not see displayed a timestamp of the follow action. When I hover over a few elements to see if it's hidden behind a tooltip for cosmetic reasons, I cannot find it.

screen shot 2018-09-22 at 16 01 29

Looking at the rendered HTML in inspector, I see that that the timestamp is available in the `aria-label` on the notification.

https://github.com/tootsuite/mastodon/blob/05f90e3695c410526fc7fa79fc6a9bdaa717cb87/app/javascript/mastodon/features/notifications/components/notification.js#L83

So, the data is accessible to screen readers, but not to other users. That's a switch!

The timestamp should be added _somewhere_, probably as a title on a span surrounding the notification text.

* * * *

colindean:
Fixes #8755 

I ensured that this change passed `npm run test:lint` but I'm not able to set up the whole test suite presently.

### DIFF
--- a/app/javascript/mastodon/features/notifications/components/notification.js
+++ b/app/javascript/mastodon/features/notifications/components/notification.js
@@ -85,8 +85,9 @@ class Notification extends ImmutablePureComponent {
             <div className='notification__favourite-icon-wrapper'>
               <i className='fa fa-fw fa-user-plus' />
             </div>
-
-            <FormattedMessage id='notification.follow' defaultMessage='{name} followed you' values={{ name: link }} />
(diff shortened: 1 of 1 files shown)


colindean:
The test failure doesn't look like it came from my addition…

ykzts (approved):

Gargron (approved):

colindean:
🎉 Thanks, @Gargron and @ykzts!

PR #8758, opened 2018-09-22
[Bugfix] Fix a problem about generating an url of notifications

GenbuHase:
## Overview
It will be work well by adding base-url as a prefix.


nightpool:
not sure i quite follow. Even if we do fix the logic there, why doesn't it currently fall through to window.open on line 186?

GenbuHase:
@nightpool 
Using URL constructor, 1st argument must be absolute path without a baseUrl as 2nd argument.

Gargron (approved):

PR #8761, opened 2018-09-23
Disable the “TOOT” button if the toot text is empty

ClearlyClaire:
This doesn't change the criteria for accepting toots, it just makes
the button reflects them better.

### DIFF
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -156,7 +156,7 @@ class ComposeForm extends ImmutablePureComponent {
     const { intl, onPaste, showSearch, anyMedia } = this.props;
     const disabled = this.props.is_submitting;
     const text     = [this.props.spoiler_text, countableText(this.props.text)].join('');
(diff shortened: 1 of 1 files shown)


Gargron:
No, this was intentional... The button has to be blue and inviting as a default state so people do not mistake it for meaning they cannot post anything.

ThibG:
I can kind of understand that, but it's also confusing at times. Especially since the button appears clickable but *nothing* happens when it is clicked.
Note that Twitter, for instance, disables the button when the textbox is empty.

Other clients have another approach: the button is always enabled, but clicking on it when the conditions aren't met will highlight why the toot could not be sent.

Gargron:
I don't think it's at all confusing. Look at GitHub for example. The "comment" button is green even when the comment textarea is empty, even though you can't post an empty comment.

ThibG:
@Gargron but it's never ever disabled, and when you click with an empty text, it tells you what's wrong instead of doing nothing

Gargron:
I have yet to see someone for whom it's actually a problem though.

wonderboom:
I agree with this design decision but I suppose you could make ThibG's proposal an optional UI/UX feature, for the people being actually confused by this button ?

PR #8766, opened 2018-09-24
Replace "elephant_ui_plane.svg" asset with Custom mascot

shleeable:
This is my first pass of this feature patch... I'm new to Ruby.. I'll test and patch bugs tonight.

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -19,6 +19,7 @@ module Admin
       theme
       thumbnail
       hero
+      mascot
       min_invite_role
       activity_api_enabled
       peers_api_enabled
@@ -41,6 +42,7 @@ module Admin
     UPLOAD_SETTINGS = %w(
       thumbnail
       hero
+      mascot
     ).freeze
 
     def edit
--- a/app/presenters/instance_presenter.rb
+++ b/app/presenters/instance_presenter.rb
@@ -44,4 +44,8 @@ class InstancePresenter
   def hero
     @hero ||= Rails.cache.fetch('site_uploads/hero') { SiteUpload.find_by(var: 'hero') }
   end
+
+  def mascot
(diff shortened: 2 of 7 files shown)


ykzts (approved):

ashleyhull-versent:
There is a hardcoded path for the file in `mastodon/app/javascript/styles/mastodon/modal.scss`

but I don't understand how to name that dynamic yet.

MaciekBaron:
Maybe I'm a bit nitpicky but could we squash these commits into one?

ashleyhull-versent:
> Maybe I'm a bit nitpicky but could we squash these commits into one?

I need to get my head around git - I'd assume this would be squashed at merge? or can I squash them before hand?

MaciekBaron:
The merge would bring all your commits into master.

The easiest way would probably be running:

    git reset --soft HEAD~7
    git commit -m "Replace SVG asset with Custom mascot"

And then doing a force push to your branch (with `-f`). This why we have one, nicely labelled commit that explains your change in the log.

MaciekBaron (approved):

Gargron:
One thing that comes to mind is, if you're planning to make all mascots customizable, naming the setting "mascot" is not going to age well. On the other hand... There are really a lot of elephants in the UI. If only some can be configured, it's going to be an inconsistency, but if you're going to customize all of them, that's a big mess too. I'm not sure if that can of worms should be opened at all?

ashleyhull-versent:
@Gargron mascot was used because that's the common reference in the code - but I understand it could be confusing.. Maybe it's better to talk about goals - I want to change this dude every now and then. as mentioned in https://github.com/tootsuite/mastodon/issues/8525

image

Is your recommendation to edit the hardcoded paths/css directly and rebuild? this means I'm `git stash`ing a few extra files or making a few extra modifications to a few extra files every upgrade.

MayMeow:
is there any way to revert customized mascot back to default one?

PR #8484, opened 2018-08-27
Add bot icon to bot avatars and migrate to newer version of Font Awesome

MaciekBaron:
This is a prerequisite for implementing #8347.

This change updated the repo used to https://www.npmjs.com/package/@fortawesome/fontawesome-free which is more up to date.

PR #8776, opened 2018-09-25
Update icon name changed by the Font Awesome 5 (regression from #8484)

mayaeh:
In Font Awesome 5, several icon names have changed.
This will fix it.
However, this may not be enough.
Please refer to the following:
https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes


ykzts (approved):
👍

PR #8796, opened 2018-09-26
Fix some icon names changed by the Font Awesome 5

mayaeh:
related: #8484 #8776 #8792 .

I am searching instead of "fa-expand" and "fa-compress".
( app/javascript/mastodon/features/video/index.js )


mayaeh:
Instead of "fa-expand" it is "fa-expand-alt", but it seems to be only Pro version.
( https://fontawesome.com/icons/expand-alt?style=solid )

This is used for expand / compress video players.

Before fontawesome-free 5.2.0:



After fontawesome-free 5.2.0 (now):

mayaeh:
"fa-expand" and "fa-compress" 's icon is different from the FontAwesome 5 and earlier, but since there is no substitute, I will leave it as it is now.
Is that okay?

Gargron (approved):

PR #8798, opened 2018-09-26
Profile fields fixes

ClearlyClaire:
The first commit raises the name and value limit from 255 to 8092 for remote statuses to take into account HTML formatting.
The second commit avoids calling `verifiable?` in serializers in preparation for verifying remote account fields links.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -310,10 +310,11 @@ class Account < ApplicationRecord
     attributes :name, :value, :verified_at, :account, :errors
 
     def initialize(account, attributes)
+      char_limit = account.local? ? 255 : 8092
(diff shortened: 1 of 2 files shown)


ThibG:
I haven't found the energy to write the verification procedure for remote accounts yet, but there are things we have to be careful about when processing it: `invisible` and `ellipsis` classes may be used for phishing.

Gargron:
The situation is a bit of a mess. Maybe it's an argument in favour of reworking the system to use a separate, special and single field for a profile's URL. Then no more HTML woes between servers. It just sucks it'd have to fit somewhere else in the profile UI.

PR #8799, opened 2018-09-27
Migrate to font-awesome 5.0.

yukimochi:
Migrate font awesome classname from `fa` to `fas`.

※ Include fixing name change
`fa-external-link` -> `fa-external-link-alt`
`fa-thumb-tack` -> `fa-thumbtack`

### Why?
`fa` means `fas` in fa5, nothing needed to do it?
but `fa` was used in fa4 fonts classname.

some of fa4 fonts assign `fab` or `fab` classname in fa5.
Continued using `fa` may create confusion.

※ Admins should notice that mastodon now can use `fas` fa5 fonts only in /about and /about/more.

ex.)
`fa fa-github` in fa4 -> `fab fa-github` in fa5 ※ now can't use it in mastodon.
`fa fa-check-circle-o` in fa4 -> `far fa-check-circle` in fa5 ※ now can't use it in mastodon.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -55,7 +55,7 @@ module ApplicationHelper
 
   def fa_icon(icon, attributes = {})
     class_names = attributes[:class]&.split(' ') || []
-    class_names << 'fa'
+    class_names << 'fas'
     class_names += icon.split(' ').map { |cl| "fa-#{cl}" }
 
     content_tag(:i, nil, attributes.merge(class: class_names.join(' ')))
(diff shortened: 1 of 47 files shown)


ykzts:
cc @MaciekBaron

MaciekBaron:
So just to clarify the situation here, this change is just to avoid confusion. Font Awesome after 5.0 prefers to explicitly specify the style (`.fas` meaning **f**ont **a**wesome **s**olid). `.fa` still works though.

Gargron (approved):

PR #8807, opened 2018-09-27
Limit the number of people that can be followed from one account

### LINKED ISSUE #2311: Follow limit
Some users send a lot of follow requests by API.
Please add follow limit feature.
ref : https://support.twitter.com/articles/66885

- No limit until 2000-5000 follows.
- Deny if follows > followers*1.1

* * * *

Gargron:
Configurable hard limit of 7,500, and above that, configurable ratio of 1.1 * followers, controlled by:

- `MAX_FOLLOWS_THRESHOLD`
- `MAX_FOLLOWS_RATIO`

Fix #2311

**Why?** There is a type of spambot where the advertising is in display name and avatar, and not in the posts, and it tries to follow every account in the fediverse, because then it appears in people's notifications. There are other spambots, but there is no good reason for anyone to follow so many accounts, and following many accounts puts a lot of strain on the server.

**Why not a rate limit?** The "import follows" function would not be affected by the rate limit, therefore it wouldn't stop anything. In this patch, the "import follows" function cuts off at the limit.

### DIFF
--- a/app/models/follow.rb
+++ b/app/models/follow.rb
@@ -25,6 +25,7 @@ class Follow < ApplicationRecord
   has_one :notification, as: :activity, dependent: :destroy
 
   validates :account_id, uniqueness: { scope: :target_account_id }
+  validates_with FollowLimitValidator, on: :create
 
   scope :recent, -> { reorder(id: :desc) }
 
--- a/app/models/follow_request.rb
+++ b/app/models/follow_request.rb
@@ -22,6 +22,7 @@ class FollowRequest < ApplicationRecord
   has_one :notification, as: :activity, dependent: :destroy
 
   validates :account_id, uniqueness: { scope: :target_account_id }
+  validates_with FollowLimitValidator, on: :create
 
   def authorize!
     account.follow!(target_account, reblogs: show_reblogs, uri: uri)
--- a/app/validators/follow_limit_validator.rb
+++ b/app/validators/follow_limit_validator.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class FollowLimitValidator < ActiveModel::Validator
+  LIMIT = ENV.fetch('MAX_FOLLOWS_THRESHOLD', 7_500).to_i
+  RATIO = ENV.fetch('MAX_FOLLOWS_RATIO', 1.1).to_f
+
+  def validate(follow)
+    return if follow.account.nil? || !follow.account.local?
+    follow.errors.add(:base, I18n.t('users.follow_limit_reached', limit: self.class.limit_for_account(follow.account))) if limit_reached?(follow.account)
+  end
+
+  class << self
+    def limit_for_account(account)
+      if account.following_count < LIMIT
+        LIMIT
+      else
+        account.followers_count * RATIO
+      end
+    end
+  end
+
+  private
+
+  def limit_reached?(account)
+    account.following_count >= self.class.limit_for_account(account)
+  end
+end
--- a/app/workers/import_worker.rb
+++ b/app/workers/import_worker.rb
@@ -37,6 +37,8 @@ class ImportWorker
   end
 
   def import_rows
-    CSV.new(import_contents).reject(&:blank?)
+    rows = CSV.new(import_contents).reject(&:blank?)
+    rows = rows.take(FollowLimitValidator.limit_for_account(@import.account)) if @import.type == 'following'
+    rows
   end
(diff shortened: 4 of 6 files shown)


nightpool:
so just to be clear, the limit is max(7500, 1.1*followers)?

if that's the case, then I think "soft limit" would be a better term then "hard limit" (which would tend to imply the OPPOSITE behavior)

m4sk1n:
Still, somebody can just set up an instance without this limit…

ThibG:
I'm not fundamentally opposed to this, but:
- Importing will *silently* drop follows
- This doesn't seem like a very effective measure overall, since remote accounts can bypass this, 7500 is pretty high anyway, and at worst the spammer will just use multiple accounts

Gargron:
@ThibG We can't process infinitely large CSV files either, and processing them synchronously to show the user an error message does not seem wise. Excluding imports from this limit is a bad idea too, since it's the easiest way for spambots to operate.

For the record, the 7,500 number is taken from Instagram, which has that limit. Twitter has a 5,000 limit.

ThibG:
I guess I'm advocating for an asynchronous signaling of import issues.

ykzts (approved):

otrapersona:
👎

Angular-Angel:
I have run into this limit. Would it be possible to get some kind of a utility function to remove accounts that no longer exist from ones follows, or something? I don't actually feel like I follow that many people. Or at least, not that many people who are actually active. :/

Angular-Angel:
I talked to my admin and they said they can;t change it cause they don't run the code directly, they do so through masto.host, and there's no admin control for this short of editing the code. Would it be possible to add a control to the admin board or something? This is seriously crippling my use of the network, I can;t follow anyone new. :/

nightpool:
@angular-angel That's a bug in masto.host, not mastodon. We're not affiliated in any way with masto.host, so if something isn't configurable, that's their fault.

nightpool:
Try unfollowing some people who don't post regularly

Angular-Angel:
But you could still add it as an admin board feature to make it more accessible, right? Or is it already an admin board feature and masto.host broke it or something? :/

As for unfollowing people manually, I'm lazy and was hoping for an easier way. Besides, think of all the poor unfortunate souls that find themselves in my position! Something must be done, I tell you! :V

eishiya:
> Try unfollowing some people who don't post regularly

A big reason to follow a lot of people is to make it easy to notice when they *do* post, e.g. with the Bell option, or with Lists. It's the frequent posters who are easiest to see without following.

I understand the goal of this limit, but I wonder if perhaps the hard limit could be relaxed for *individual accounts*, so that the relatively few users who have a legitimate need for a higher limit can be given it or no limit at all (they're few enough that a manual overview should be no problem for mods), without setting the limit higher for the potential abusive users/bots.

I realise that the ratio is meant to be a way for individuals to have higher limits, but from what I've seen, people who follow a lot of people don't tend to be the sort of people who get a huge number of followers - they tend to be people who mainly *read* and maybe *reply*, and such posts don't tend to get one many followers. The ratio approach on Twitter always seemed to me like something that rewards spammer networks rather than real people, and is a lazy way of automating the limit. Mastodon has the benefit of human admins and mods who are able to understand *why* their users follow people.

Gargron:
7,500 people is a huge number though

eishiya:
It is! I'm certainly in no danger of ever hitting it. But I've seen (human) users hit Twitter's limit and a user on the instance I used expressed worries about hitting it on Mastodon too. With Mastodon's current smaller user base and with the local timeline on a well-chosen home instance being a repalcement for many follows, I think they're in less danger of hitting the limit, but apparently some people just... really like to follow a load of people and don't care about being able to keep up. I don't understand them, but I do think they're worth accommodating.

Angular-Angel:
Yeah, I hit that number. Now, a lot of those are old accounts, many on defunct instances - but I'd still appreciate a higher number,so I don't need to spend huge amounts of time digging through my follow list to remove them.

cfwelch:
I have been importing lists of journalists and scholars from specific disciplines to my account and hit this limit immediately. I thought these lists would be a great way to get started on Mastodon and to be exposed to more diverse kinds of information than on other platforms. I would appreciate some way for me to add more accounts. Maybe if there is a verification step to show that I'm not a bot or spammer. At the moment, I am just hoping to add a couple thousand more accounts.

PR #8829, opened 2018-09-29
Add tootctl settings registrations open

Gargron:
- `tootctl settings registrations open`
- `tootctl settings registrations close`


Gargron:
Not sure if maybe, alternatively, the "open" and "close" could be subcommands. More bothersome to code though.

ykzts (approved):

PR #8832, opened 2018-09-29
Add conversations API

### LINKED ISSUE #3255: Endpoint for all direct messages to and from a given user, or all messages in general

Gargron:
Fix #3255

Replacement for computationally-heavy and unsatisfying direct timelines API. The new API would list DM conversations, rather than individual DMs, which allows for UX that people are more used to.

- REST API: `GET /api/v1/conversations`
- Streaming API: New event `conversation` in the `direct` stream

___

- [x] Data structures
- [x] REST API
- [x] Streaming API
- [x] Tests
- [x] Basic support in Web UI
- [x] Polished design in Web UI
- [x] Keyboard navigation in Web UI

### DIFF
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -1,16 +1,55 @@
 # frozen_string_literal: true
 
 class Api::V1::ConversationsController < Api::BaseController
+  LIMIT = 20
+
   before_action -> { doorkeeper_authorize! :read, :'read:statuses' }
   before_action :require_user!
+  after_action :insert_pagination_headers
+
+  respond_to :json
 
   def index
-    render json: conversations, each_serializer: REST::ConversationSerializer
+    @conversations = paginated_conversations
+    render json: @conversations, each_serializer: REST::ConversationSerializer
   end
 
   private
 
-  def conversations
-    ConversationAccount.where(account: current_account).order(last_status_id: :desc)
+  def paginated_conversations
+    ConversationAccount.where(account: current_account)
+                       .paginate_by_id(limit_param(LIMIT), params_slice(:max_id, :since_id, :min_id))
+  end
+
+  def insert_pagination_headers
+    set_pagination_headers(next_path, prev_path)
+  end
+
+  def next_path
+    if records_continue?
+      api_v1_conversations_url pagination_params(max_id: pagination_max_id)
+    end
+  end
+
+  def prev_path
+    unless @conversations.empty?
+      api_v1_conversations_url pagination_params(min_id: pagination_since_id)
+    end
+  end
+
+  def pagination_max_id
+    @conversations.last.last_status_id
+  end
+
+  def pagination_since_id
+    @conversations.first.last_status_id
+  end
+
+  def records_continue?
+    @conversations.size == limit_param(LIMIT)
+  end
+
+  def pagination_params(core_params)
+    params.slice(:limit).permit(:limit).merge(core_params)
   end
 end
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -17,7 +17,7 @@ class Api::V1::ConversationsController < Api::BaseController
   private
 
   def paginated_conversations
-    ConversationAccount.where(account: current_account)
+    AccountConversation.where(account: current_account)
                        .paginate_by_id(limit_param(LIMIT), params_slice(:max_id, :since_id, :min_id))
   end
 
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class Api::V1::ConversationsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:statuses' }
+  before_action :require_user!
+
+  def index
+    render json: conversations, each_serializer: REST::ConversationSerializer
+  end
+
+  private
+
+  def conversations
+    ConversationAccount.where(account: current_account).order(last_status_id: :desc)
+  end
+end
--- a/app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
+++ b/app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ImmutablePropTypes from 'react-immutable-proptypes';
+import ImmutablePureComponent from 'react-immutable-pure-component';
+import ConversationContainer from '../containers/conversation_container';
+import ScrollableList from '../../../components/scrollable_list';
+import { debounce } from 'lodash';
+
+export default class ConversationsList extends ImmutablePureComponent {
+
+  static propTypes = {
+    conversationIds: ImmutablePropTypes.list.isRequired,
+    hasMore: PropTypes.bool,
+    isLoading: PropTypes.bool,
+    onLoadMore: PropTypes.func,
+    shouldUpdateScroll: PropTypes.func,
+  };
+
+  getCurrentIndex = id => this.props.conversationIds.indexOf(id)
+
+  handleMoveUp = (id) => {
+    const elementIndex = this.getCurrentIndex(id) - 1;
+    this._selectChild(elementIndex);
+  }
+
+  handleMoveDown = (id) => {
+    const elementIndex = this.getCurrentIndex(id) + 1;
+    this._selectChild(elementIndex);
+  }
+
+  _selectChild (index) {
+    const element = this.node.node.querySelector(`article:nth-of-type(${index + 1}) .focusable`);
+
+    if (element) {
+      element.focus();
+    }
+  }
+
+  setRef = c => {
+    this.node = c;
+  }
+
+  handleLoadOlder = debounce(() => {
+    const last = this.props.conversationIds.last();
+
+    if (last) {
+      this.props.onLoadMore(last);
+    }
+  }, 300, { leading: true })
+
+  render () {
+    const { conversationIds, onLoadMore, ...other } = this.props;
+
+    return (
+      <ScrollableList {...other} onLoadMore={onLoadMore && this.handleLoadOlder} scrollKey='direct' ref={this.setRef}>
+        {conversationIds.map(item => (
+          <ConversationContainer
+            key={item}
+            conversationId={item}
+            onMoveUp={this.handleMoveUp}
+            onMoveDown={this.handleMoveDown}
+          />
+        ))}
+      </ScrollableList>
+    );
+  }
+
+}
--- a/app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
+++ b/app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
@@ -18,12 +18,12 @@ export default class ConversationsList extends ImmutablePureComponent {
 
   getCurrentIndex = id => this.props.conversationIds.indexOf(id)
 
-  handleMoveUp = (id) => {
+  handleMoveUp = id => {
     const elementIndex = this.getCurrentIndex(id) - 1;
     this._selectChild(elementIndex);
   }
 
-  handleMoveDown = (id) => {
+  handleMoveDown = id => {
     const elementIndex = this.getCurrentIndex(id) + 1;
     this._selectChild(elementIndex);
   }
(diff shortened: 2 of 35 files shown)


ClearlyClaire (commented):
So, I'm not completely sure what conversations are supposed to be here. I gather that they are threads of direct messages and are associated with a given number of participants. Multiple conversations with the same set of participants seem possible. Is that on purpose? It seems like a toot can be part of multiple conversations (through adding/removing participants). Is that also on purpose?

Finally, I'm very confused at how participants are computed, there might be a bug, there? To me, participants would be the author of the toot + all mentioned users. It seems that in this case, it's all mentioned users and not the author of the toot? I think there might be a bug there.

(Haven't reviewed front-end or streaming API changes)

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def paginate_by_max_id(limit, max_id = nil, since_id = nil)
+      query = order(arel_table[:last_status_id].desc).limit(limit)
+      query = query.where(arel_table[:last_status_id].lt(max_id)) if max_id.present?
+      query = query.where(arel_table[:last_status_id].gt(since_id)) if since_id.present?
+      query
+    end
+
+    def add_status(recipient, status)
+      conversation = find_or_initialize_by(account: recipient, conversation_id: status.conversation_id, participant_account_ids: participants_from_status(status))
```
ThibG: It's somewhat unlikely to happen, but I'm worried about race conditions here.

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+        conversation.save
+      end
+
+      conversation
+    end
+
+    private
+
+    def participants_from_status(status)
+      (status.mentions.pluck(:account_id) - [status.account_id]).sort
```
ThibG: The toot's author isn't part of the conversation… ?

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def remove_status(recipient, status)
+      conversation = find_by(account: recipient, conversation_id: status.conversation_id, participant_account_ids: participants_from_status(status))
+
+      return if conversation.nil? || conversation.last_status_id != status.id
+
+      while (last_status_id = conversation.status_ids.pop)
+        last_status = Status.find(last_status_id)
+        break if last_status
+      end
```
ThibG: I find this loop slightly confusing. Maybe a comment should be added to make clear that it only removes “orphaned” items? Also, why not remove “orphaned” items regardless of whether the last toot was removed?

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def paginate_by_max_id(limit, max_id = nil, since_id = nil)
+      query = order(arel_table[:last_status_id].desc).limit(limit)
+      query = query.where(arel_table[:last_status_id].lt(max_id)) if max_id.present?
+      query = query.where(arel_table[:last_status_id].gt(since_id)) if since_id.present?
+      query
+    end
+
+    def add_status(recipient, status)
+      conversation = find_or_initialize_by(account: recipient, conversation_id: status.conversation_id, participant_account_ids: participants_from_status(recipient, status))
```
ThibG: (Re-stating my concern about race conditions as github marked my comment as obsolete)

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+#
+#  id                      :bigint(8)        not null, primary key
+#  account_id              :bigint(8)
+#  conversation_id         :bigint(8)
+#  participant_account_ids :bigint(8)        default([]), not null, is an Array
+#  status_ids              :bigint(8)        default([]), not null, is an Array
+#  last_status_id          :bigint(8)
+#
+
+class ConversationAccount < ApplicationRecord
```
nightpool: I think AccountConversations makes at least a *little* bit more sense here.

Gargron: You're right, that makes perfect sense, and I hate how much I'm gonna need to search and replace here oof

app/models/domain_block.rb
```diff
@@ -3,12 +3,13 @@
...
-#  updated_at   :datetime         not null
-#  severity     :integer          default("silence")
-#  reject_media :boolean          default(FALSE), not null
+#  id             :bigint(8)        not null, primary key
+#  domain         :string           default(""), not null
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  severity       :integer          default("silence")
+#  reject_media   :boolean          default(FALSE), not null
+#  reject_reports :boolean          default(FALSE), not null
```
nightpool: unrelated to this pull?

nightpool (commented):
Incomplete, sorry :(

app/models/conversation_account.rb
ThibG: Because it only removes toots from the array if the removed toot is the last one. This is to track the last status, which should always be the visible one.

app/models/conversation_account.rb
ThibG: `find_or_create`is not atomic: https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by
But here, it's even worse, we are using `find_or_initialize_by`, which means the database record is only created when we call `save` a few lines further.
There is a possibility multiple incoming toots for a same new conversations would be processed in parallel. This could for instance occur if sidekiq is down while someone sends us a string of DMs, which could then be processed at the same time when sidekiq goes up, resulting in two conversations for the same threads.

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+# == Schema Information
+#
+# Table name: conversation_accounts
+#
+#  id                      :bigint(8)        not null, primary key
+#  account_id              :bigint(8)
+#  conversation_id         :bigint(8)
+#  participant_account_ids :bigint(8)        default([]), not null, is an Array
+#  status_ids              :bigint(8)        default([]), not null, is an Array
```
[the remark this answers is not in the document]
ThibG: Not sure about `participant_account_ids`, but afaik, `status_ids` makes sense because it's ordered and only used to track the latest status in a thread more efficiently when deleting statuses.
(Although it probably doesn't handle toots received out-of-order that well, hm)

app/models/domain_block.rb
ashleyhull-versent: https://github.com/tootsuite/mastodon/commit/dced62557476ce42c6733674fe73fea9d671b663#diff-2ca536250498005a761e8b1427fe2515R12

app/controllers/api/v1/conversations_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  def pagination_since_id
+    @conversations.first.last_status_id
+  end
+
+  def records_continue?
+    @conversations.size == limit_param(LIMIT)
+  end
+
+  def pagination_params(core_params)
+    params.slice(:limit).permit(:limit).merge(core_params)
```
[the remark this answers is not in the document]
Gargron: Slice is required to avoid warnings in the log about unpermitted parameters being passed.

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+
+class ConversationAccount < ApplicationRecord
+  after_commit :push_to_streaming_api
+
+  belongs_to :account
+  belongs_to :conversation
+  belongs_to :last_status, class_name: 'Status'
+
+  def participant_account_ids=(arr)
+    self[:participant_account_ids] = arr.sort
```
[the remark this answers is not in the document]
Gargron: No, it doesn't

app/models/conversation_account.rb
Gargron: Indeed, we don't update the record if a deleted status is not the *last* status, therefore the array could contain IDs of statuses that no longer exist.

app/models/conversation_account.rb
```diff
@@ -0,0 +1,109 @@
...
+      ((status.mentions.pluck(:account_id) + [status.account_id]).uniq - [recipient.id]).sort
+    end
+  end
+
+  private
+
+  def push_to_streaming_api
+    return unless subscribed_to_timeline?
+
+    if destroyed?
```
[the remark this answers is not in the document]
Gargron: It's a stub. There probably *should* be a delete event for these. The problem is event naming. It would have to be called something ugly like "conversation_delete"

ThibG:
Do conversations with blocked users disappear from the list? It doesn't seem so by having a quick look at the code.

app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
```diff
@@ -0,0 +1,68 @@
...
+    const elementIndex = this.getCurrentIndex(id) - 1;
+    this._selectChild(elementIndex);
+  }
+
+  handleMoveDown = id => {
+    const elementIndex = this.getCurrentIndex(id) + 1;
+    this._selectChild(elementIndex);
+  }
+
+  _selectChild (index) {
```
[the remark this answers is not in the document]
Gargron: Because this method is called from other methods rather being an event handler directly, there are no issues with `this` binding that the other syntax is meant to solve.

app/javascript/mastodon/reducers/notifications.js
```diff
@@ -69,7 +69,7 @@ const expandNormalizedNotifications = (state, notifications, next) => {
     }
 
     if (!next) {
-      mutable.set('hasMore', true);
+      mutable.set('hasMore', false);
```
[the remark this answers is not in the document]
Gargron: It's a bug I found when copying the code to conversations reducer. I think it's relevant.

app/javascript/mastodon/features/direct_timeline/components/conversations_list.js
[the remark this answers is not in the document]
nightpool: yeah, the lambda form is probably better but we should be consistent throughout the file (and ideally throughout the project), lexical vs function target `this` is a hard distinction to remember in the moment.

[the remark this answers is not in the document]
Gargron: Event handlers use `=`, everything else uses the normal class method form.

ThibG:
I'm still worried about possible race conditions, but otherwise, this looks fine.
I will probably try that next week on my instance.
What about existing conversations (prior to the migration), though?

ThibG:
Hm, yeah, having an API to delete conversations instead of deleting them on block/mute would be nice (sorry about my conflicting earlier proposal). Shouldn't be hard to do either, although I'm not sure how the UI would look like.

nightpool:
we could add the rails lock_version column to avoid race conditions

nightpool:
what? lock_version definitely works accross ruby processes.....
https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html

Gargron:
>This locking mechanism will function inside a single Ruby process. To make it work across all web requests, the recommended approach is to add lock_version as a hidden field to your form.

Gargron:
Anyway, what I was saying, DM conversations usually don't have so many participants that an update happens within the same second to the point a race condition would actually occur. But if it does, I am ready to add a locking mechanism in the future.

I just want to merge this now

ThibG:
As I said, a likely possibility is getting two successive DMs from the same person. Depending on the load, they could very well be processed concurrently.

streaming/index.js
```diff
@@ -525,9 +526,11 @@ const startWorker = (workerId) => {
       ws.isAlive = true;
     });
 
+    let channel;
```
nightpool: Not sure I understand this change. Why make it mutable? We never fall through any of the switch cases, right?

nightpool (approved):

nightpool:
LGTM!

streaming/index.js
Gargron: A few of the switch cases needed to reuse the same assembled string (channel) in two places at once, but a switch clause is not a separate scope, so you can't redefine a const there. Therefore, it's easiest to define the channel up there.

PR #8852, opened 2018-10-02
Use process.env to allow configuring of socket permissions

gomasy:
Puma can be set with environment variables, but streaming cannot do that...

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -586,7 +586,7 @@ const startWorker = (workerId) => {
 
   if (process.env.SOCKET || process.env.PORT && isNaN(+process.env.PORT)) {
     server.listen(process.env.SOCKET || process.env.PORT, () => {
-      fs.chmodSync(server.address(), 0o666);
+      fs.chmodSync(server.address(), process.env.SOCKET_PERMISSION || 0o666);
       log.info(`Worker ${workerId} now listening on ${server.address()}`);
     });
   } else {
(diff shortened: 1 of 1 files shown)


Gargron:
>Puma can be set with environment variables, but streaming cannot do that...

I cannot find a Puma setting for `SOCKET_PERMISSION`, could you clarify?

Gomasy:
screenshot from 2018-10-03 14-27-33
https://github.com/puma/puma/blob/master/README.md

Puma can change permissions of the socket by adding the umask parameter.

streaming/index.js
```diff
@@ -586,7 +586,7 @@ const startWorker = (workerId) => {
 
   if (process.env.SOCKET || process.env.PORT && isNaN(+process.env.PORT)) {
     server.listen(process.env.SOCKET || process.env.PORT, () => {
-      fs.chmodSync(server.address(), 0o666);
+      fs.chmodSync(server.address(), process.env.SOCKET_PERMISSION || 0o666);
```
Gargron: SOCKET_PERMISSION will be a string, but 0o666 is not a string. Perhaps conversion is needed here?

Also maybe call it SOCKET_UMASK?

Gargron (changes requested):

Gomasy:
@Gargron It has changed.
It is calculated if exists, but if it does not exist 0o666 is set as usual.

PR #8855, opened 2018-10-02
add ffmpeg initializer

### LINKED ISSUE #8648: sandbox ffmpeg by default
ffmpeg is project with a [lot of security fixes](https://www.ffmpeg.org/security.html) shipped every version. Adding [firejail](https://firejail.wordpress.com/), other than adding SELinux or AppArmor support, is fast and portable across all Linux 3.x+ targets and provides some protection from malicious user input.

A [profile for ffmpeg](https://github.com/netblue30/firejail/blob/master/etc/ffmpeg.profile) that works with Mastodon out of the box ships with firejail, but could probably be refined (e.g. Mastodon doesn't need ffmpeg to have write access to the FS).

I've written [a guide on how to set this up manually](https://write.as/serenitylabs/using-firejail-to-sandbox-ffmpeg-in-mastodon).
* * * *

ghost:
fixes #8648 

this might be useful if multiple versions of ffmpeg are installed, or if additional sandboxing should be applied

### DIFF
--- a/config/initializers/ffmpeg.rb
+++ b/config/initializers/ffmpeg.rb
@@ -1,3 +1,3 @@
-if defined? ENV['FFMPEG_BINARY']
+if ENV['FFMPEG_BINARY'].present?
     FFMPEG.ffmpeg_binary = ENV['FFMPEG_BINARY']
 end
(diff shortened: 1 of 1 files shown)


config/initializers/ffmpeg.rb
```diff
@@ -0,0 +1,3 @@
+if defined? ENV['FFMPEG_BINARY']
```
Gargron: `if ENV['FFMPEG_BINARY'].present?`

Gargron (changes requested):

Gargron (approved):

PR #8861, opened 2018-10-03
Leave unknown language as nil if account is remote

tribela:
It is not reasonable to use Local instance's language setting for foreign user's toot.

### DIFF
--- a/app/lib/language_detector.rb
+++ b/app/lib/language_detector.rb
@@ -12,7 +12,6 @@ class LanguageDetector
   def detect(text, account)
     input_text = prepare_text(text)
     return if input_text.blank?
-    return detect_language_code(input_text, true) unless account.local?
 
     detect_language_code(input_text) || default_locale(account)
   end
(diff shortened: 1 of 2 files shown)


Kjwon15:
When language field is `nil`, Is it never shown to user who filtered some languages?

Gargron:
The language detector is super inaccurate for short messages. Also, you can't ignore when people set a posting language preference (which is hopefully more correct than the detector, although vandalism is also possible)

Kjwon15:
So.. It's better to set `nil` on foreign toot which not having language field?

ThibG:
@Kjwon15 When a status' language is set to `nil`, it is always shown.

@Gargron afaict this change doesn't ignore the posting language setting, it just drops the fallback to default user / instance setting on receiving remote toots for which a language wasn't defined and for which the detection has been deemed unreliable.

ClearlyClaire (approved):

Kjwon15:
Yes, It actually fixes a bug

PR #8868, opened 2018-10-03
Fix link verification for remote accounts

Gargron:
Resolve #8798

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -312,8 +312,8 @@ class Account < ApplicationRecord
     def initialize(account, attributes)
       @account     = account
       @attributes  = attributes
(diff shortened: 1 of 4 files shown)


ykzts (approved):

app/models/account.rb
```diff
@@ -334,6 +344,16 @@ def mark_verified!
...
       { name: @name, value: @value, verified_at: @verified_at }
     end
+
+    private
+
+    def string_limit
+      if account.local?
+        255
+      else
+        2047
```
ThibG: I'm worried that's still a bit small. With custom emoji and/or percent-encoded links, this number can rise pretty quickly.
But at least it should cover the issue of having one <255 chars plain link truncated when federating.

PR #8878, opened 2018-10-04
lint pass 2

aus-social:
Learning Ruby by fixing code quality alerts... as per discussion with @nightpool, some of these things are by choice.. please feel free to veto these or comment away


config/initializers/omniauth.rb
```diff
@@ -3,7 +3,7 @@
 end
 
 Devise.setup do |config|
-  # Devise omniauth strategies
+  #  Devise omniauth strategies
```
Gargron: This is strange

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -120,9 +120,9 @@ def call_create
 
       def call_update
         patch :update, params: {
-                id: app.id,
+          id: app.id,
                 doorkeeper_application: opts
```
Gargron: I don't think this is right

config/initializers/omniauth.rb
ashleyhull-versent: correct... I think this is a bashism..

shleeable (changes requested):

spec/controllers/settings/applications_controller_spec.rb
ashleyhull-versent: I missed the second line?

spec/controllers/settings/applications_controller_spec.rb
```diff
@@ -139,14 +139,14 @@ def call_update
     context 'failure' do
       before do
         patch :update, params: {
-                id: app.id,
+          id: app.id,
                 doorkeeper_application: {
```
Gargron: Spacing off here as well

aus-social: I re-alligned the whole block -  I think this is better now.

Gargron (approved):

PR #8879, opened 2018-10-04
lint pass 3

aus-social:
Learning Ruby by fixing code quality alerts... as per discussion with @nightpool, some of these things are by choice.. please feel free to veto these or comment away


Gargron:
It's highly unlikely that this has a positive impact on the performance of the test suite, and it takes away human-readability of the code a little bit, so I would pass on this.

PR #8880, opened 2018-10-04
Update node package.json

aus-social:
Lots of changes - increased risk, but node isn't used for much after build time. (I think)


ykzts:
Did you make sure to run?

PR #8893, opened 2018-10-05
Add a confirmation dialog when hitting reply and the compose box isn't empty

### LINKED ISSUE #878: Use confirmation dialog before erasing draft to reply to new status
This is the sort of thing that you presumably get trained out of pretty fast, but it's kind of annoying the first few times it happens.

If you start writing something in the "What is on your mind" edit box on the left, and then notice someone says something you want to respond to, and you hit the little 'Reply' arrow under their toot, your existing text is discarded with no undoo possible (as far as I can tell.)

ClearlyClaire:
Fixes #878

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -45,6 +45,7 @@ class Status extends ImmutablePureComponent {
   static propTypes = {
     status: ImmutablePropTypes.map,
     account: ImmutablePropTypes.map,
+    askReplyConfirmation: PropTypes.bool,
     onReply: PropTypes.func,
     onFavourite: PropTypes.func,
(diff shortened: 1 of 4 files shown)


app/javascript/mastodon/components/status.js
```diff
@@ -71,6 +72,7 @@ class Status extends ImmutablePureComponent {
     'account',
     'muted',
     'hidden',
+    'askReplyConfirmation',
```
ThibG: Thinking about it, I'm not too sure about the performance implications of this. Maybe the `mapDispatchToProps` thing should be rewritten instead to use `getState`.

kit-ty-kate:
Thanks!! <3

ykzts (approved):

PR #8899, opened 2018-10-06
Speed up production build

smorimoto:
The source map is unnecessary for production build.
It slows down the build.

Benchmark:

| Before | After |
| :-----: | :----: |
| 1m5.159s | 0m54.477s |

`Built on a VPS with 4 virtual cores and 8GB memory.`

And parallelization makes it a lot faster.

Benchmark:

| Before | After |
| :-----: | :----: |
| 1m5.159s | 0m43.487s |

`Built on a VPS with 4 virtual cores and 8GB memory.`

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -50,11 +50,11 @@ module.exports = merge(sharedConfig, {
     minimize: true,
     minimizer: [
       new UglifyJsPlugin({
+        cache: true,
+        parallel: true,
         sourceMap: true,
 
         uglifyOptions: {
-          mangle: true,
-
           compress: {
             warnings: false,
           },


imbsky:
`mangle` is also one cause of slow down build speed, but I really don't know if it's necessary.
Can I get your opinion?

ykzts:
> The source map is unnecessary for production build.

Mastodon intentionally outputs source map in production.

imbsky:
I see. I will fix it, and is `mangle` the same as that?

ykzts:
I think that `mangle` can be changed to` false`.

imbsky:
I pushed fixed commit.

ykzts (approved):

nightpool:
What is the bundle size before and after this commit? I feel like this change was made too hastily—mangle is a key compression tool, and it would be bad if this change increased the bundle download time for all users just to make compilation slightly faster for admins.

nightpool:
I can't find any difference in bundle sizes when I was testing this locally, but I suspect that means that the mangle option *wasn't* turned off (because it's `true` by default according to https://github.com/mishoo/UglifyJS2#minify-options)

imbsky:
Oh, that makes sense…

PR #8903, opened 2018-10-07
Add fallback for PostgreSQL without upsert in CopyStatusStats

### LINKED ISSUE #8590: Newest release breaks compatibility with PostgreSQL 9.4
We're on an older LTS of Ubuntu, which only gets 9.4 by default. Until v2.5.0 it worked without problems, but the migrations for the latest release include one with an upsert statement, which is only available in postgres 9.5 and newer:

https://github.com/tootsuite/mastodon/blob/master/db/migrate/20180812173710_copy_status_stats.rb#L12-L13

I looked into upgrading postgres on this server, but it would be too difficult with our current setup. Would be great if this migration could be refactored to also work on 9.4, because everything else in Mastodon seems to be compatible.

* * * *

Gargron:
Fix #8590

### DIFF
--- a/db/migrate/20180812173710_copy_status_stats.rb
+++ b/db/migrate/20180812173710_copy_status_stats.rb
@@ -3,20 +3,52 @@ class CopyStatusStats < ActiveRecord::Migration[5.2]
 
   def up
     safety_assured do
-      Status.unscoped.select('id').find_in_batches(batch_size: 5_000) do |statuses|
-        execute <<-SQL.squish
-          INSERT INTO status_stats (status_id, reblogs_count, favourites_count, created_at, updated_at)
-          SELECT id, reblogs_count, favourites_count, created_at, updated_at
-          FROM statuses
-          WHERE id IN (#{statuses.map(&:id).join(', ')})
-          ON CONFLICT (status_id) DO UPDATE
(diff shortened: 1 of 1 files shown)


db/migrate/20180812173710_copy_status_stats.rb
```diff
@@ -3,20 +3,52 @@ class CopyStatusStats < ActiveRecord::Migration[5.2]
...
 
   def down
     # Nothing
   end
+
+  private
+
+  def supports_upsert?
+    version = select_one("SELECT current_setting('server_version_num') AS v")['v'].to_i
+    version >= 90600
```
nightpool: someone said 9.5 i think?

db/migrate/20180812173710_copy_status_stats.rb
```diff
@@ -3,20 +3,52 @@ class CopyStatusStats < ActiveRecord::Migration[5.2]
...
+      SQL
+    end
+  end
+
+  def up_slow
+    say 'Upsert is not available in PostgreSQL below 9.6, falling back to slow import of counters'
+
+    # We cannot use bulk INSERT or overarching transactions here because of possible
+    # uniqueness violations that we need to skip over
+    Status.unscoped.select('id, reblogs_count, favourites_count, created_at, updated_at').find_each do |status|
```
nightpool: find_in_batches?

nightpool (approved):

Issue #6359, opened 2018-01-25
Join several hashtags in a search/column


renatolond:
I noticed that there's multiple tags in which people add which songs they are listening to (np, nowplaying, radiotoot, tootradio, pouetradio, etc). I'd like to follow those hashtags for idea on what to listen, but for now the only way I found is to open a column for each of them and them have toots that use multiple of those appearing on several columns. I would like to be able to follow all of them in a single column.
* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Letiteuf55:
Yeah, i need that too, i'm running out of space on my screen with all my # researches !

Granitosaurus:
This is a huge issue for me as well as my 13" laptops screenspace is rather limited.   

I also proposed another feature to address the same screenspace issue: #7050 
Combination of both of these features would greatly improve user experience and screen-space effiency.

Gargron:
Are we talking about OR, AND, or both?

renatolond:
I'm talking about OR. I would like to see posts that contain either np, nowPlaying, radioToot or tootRadio at the same column

nightpool:
i think or is the most useful feature considering how people use hashtags,
I wouldn't be worried about ANDs

woj-tek:
Could I suggest making operator optional (defaulting to OR for example, though for me AND would be more natural)?
For example I would be quite often interested in narrowing search more instead of expanding them.

roipoussiere:
Just to resume the discussion, the behavior could be:

- `#nowplaying #radiotoot` -> one of these hashtags
- `#nowplaying or #radiotoot` -> one of these hashtags
- `#cute and #cat` -> both of them

gdpelican:
I'm thinking to have a look at this over the next wee while.

Apologies for the extremely rough mockup with hasty copy, but I was thinking something along the lines of this:
- Perform a search, and add a new hashtag column to your desktop
- Open the settings for that hashtag column
- Add the option to specify additional hashtags to include in this column
- Add the option to specify whether this is an 'AND' or an 'OR' relationship

![captura de pantalla 2018-09-13 a la s 3 49 14 p m](https://user-images.githubusercontent.com/750477/45515020-a0fbaf00-b76c-11e8-8f33-24bee4dd481f.png)

Thoughts?

[...]

PR #8904, opened 2018-10-07
Allow joining several hashtags in a single column

gdpelican:
Ref #6359 

Watch [this video](https://www.useloom.com/share/175fa307e42a431690745f41d61f4271) for more details on intended functionality :)


ashleyhull-versent:
Excellent video explainer... does this limit to hashtags filtered from the home feed and/or local feed and/or federated feed? 

this feature is worth it on it's own, but I'd like to see hashtags being pulled from the fediverse to get access to toots in and out of my social circle. if possible, I'd be nice to be able to filter in or out either locally or from the firehose.

gdpelican:
I have to admit to still be wrapping my head around the federation features of Mastodon, but this PR always respects 'local' variable that's passed through on the backend.

So I believe the answer is 'If your normal hashtag column displays remote tweets, then the modified AND/OR/NOT column will as well', but there's nothing in here at the moment around going and fetching statuses from other instances, which would likely be out of scope for this particular PR.

ashleyhull-versent:
Totally understandable... thanks :)

ChameleonScales:
I've been requesting that feature. Thank you for this PR, I hope it will make its way in Mastodon.
Afaik the toots that show up are either in your instance or linked to your instance through following or boosting (if someone in your instance follows the user who tooted or boosted his toot)

Gargron:
This is pretty cool, I just watched the video. There are some code climate issues that need to be resolved. Double-check that the SQL code isn't actually vulnerable to SQL injection, and if it isn't, you can use brakeman to whitelist those lines. I would prefer ORM instead of raw SQL wherever possible for this reason, too.

gdpelican:
Alright, thanks for the feedback guys.

I've had a swing at styling the selects, and done a refactor so that the tags can be applied in all 'modes' at once (full power!)

image

image

I've also put those SQL injection warnings into the brakeman file, and changed the code up a little bit so it's a bit more obvious there's no injection to be had there (we now call `to_i` on the input before it goes in)

Gargron (approved):

Gargron:
Wait, how does the streaming part work?

app/javascript/mastodon/reducers/timelines.js
```diff
@@ -86,6 +87,10 @@ const deleteStatus = (state, id, accountId, references) => {
   return state;
 };
 
+const clearTimeline = (state, timeline) => {
+  return state.updateIn([timeline, 'items'], list => list.filter(() => false));
```
Gargron: `list => list.clear()`

app/models/status.rb
```diff
@@ -79,6 +79,17 @@ class Status < ApplicationRecord
   scope :including_silenced_accounts, -> { left_outer_joins(:account).where(accounts: { silenced: true }) }
   scope :not_excluded_by_account, ->(account) { where.not(account_id: account.excluded_from_timeline_account_ids) }
   scope :not_domain_blocked_by_account, ->(account) { account.excluded_from_timeline_domains.blank? ? left_outer_joins(:account) : left_outer_joins(:account).where('accounts.domain IS NULL OR accounts.domain NOT IN (?)', account.excluded_from_timeline_domains) }
+  scope :tagged_with_all, ->(tags) {
```
Gargron: I guess stylistically if you need a multiline lambda for a scope, the scope is better off being a class method

app/controllers/tags_controller.rb
```diff
@@ -16,14 +16,16 @@ def show
       end
 
       format.rss do
-        @statuses = Status.as_tag_timeline(@tag).limit(PAGE_SIZE)
+        @statuses = HashtagQueryService.new.call(@tag, params.slice(:any, :all, :none)).distinct.limit(PAGE_SIZE)
```
Gargron: If `distinct` is called on all uses of the service, why not put it inside the service?

Gargron (dismissed):
Sorry, reviewed it again and found some issues. Most importantly, the streaming API is untouched, so even with all the column params, new statuses will only appear from the main hashtag.

Changing the streaming API to support such filtering is not straightforward and I don't have advice ready. Each streaming API consumer subscribes to one redis pubsub channel (well, the channels are multiplexed actually). But the the thing is, you can't really subscribe to a union of two channels. They aren't published in the same server tick either, there might be any delay.

Perhaps your only option is to subscribe to all individual streaming API channels in the web UI, and then filter on ingress based on the tags included in the JSON.

app/controllers/tags_controller.rb
gdpelican: Ah, it was because rails 5's `or` method is verrrry picky about what kinds of clauses you can join via or (and distinct was blowing it up), but I believe we can pull it into the service by calling `distinct` on both the original query and the 'or' one.

app/models/status.rb
gdpelican: There's a subtle difference here in that in a class method, `self` refers to the class itself (ie, Status, which would map to Status.all), whereas within a scope `self` refers to the existing relation that the scope's being called on.

So a straight port to a class method doesn't quite work because instead of reducing the existing relation, we start reducing from all statuses

gdpelican:
Okay great, I've patched up a couple little things you mentioned and will have a poke into the streaming stuff over the next while.

An initial glance makes me think there are two possible solutions:
- Define a specific stream which includes the additional arguments... right now it's `hashtag?tag=cats` and maybe it could be `hashtag?tag=cats&any[]=dogs`. The `streamFrom` method on the streaming server is still a bit opaque to me, but I'd think this approach would be better if we can get it.

- Have the column also subscribe to everything in the 'any' field, and then discard clientside anything that comes through which violates the 'and' or 'not' fields

trwnh:
The screenshot you posted looks better, but I'm still wondering what would happen if you wanted to remove the "cats" hashtag while keeping all the other tags. Would you have to make an entirely new column? This would be inconvenient if you had many "additional" tags but only wanted to change the "primary".

gdpelican:
@trwnh Hm, I'm not really sure I'm following you here; there's not currently a function that I'm aware of to update the hashtag of a column (I suppose the workflow is 'make a new column, pin it, and unpin the old one'?), and I'm not sure it's within the scope of this PR to add it, since this is simply trying to add options to an existing column.

You could actually do some odd stuff with these settings like 'cats OR dogs WITHOUT cats', which would end up being a slightly-less-performant 'dogs' column, but I wouldn't really want to encourage that behaviour.

Gargron:
I mean, tagId is just a prop passed from column settings in the settings reducer, so it could be changed on the fly, I think

gdpelican:
Alright, I've updated this to do the second option here, which is

- When subscribing a hashtag column, subscribe to the 'main' tag, as well as all of the tags in the 'any' field
- When a status comes through one of those subscriptions, run it through an 'accept' function which will ignore it if it doesn't meet the criterion of either the 'all' or 'none' fields.

justinmayer:
Great work on this feature, James. In addition, I like the inclusion of the video to help introduce the use case and initial implementation. Bravo! 🚀

gdpelican:
Hey @Gargron , is there anything else I can do here to shore this one up?

Gargron (approved):

gdpelican:
Thanks for the support on this @Gargron; I had a great first PR experience with Mastodon. Feel free to ping me if there are any issues or enhancements that come up around this and I'll be happy to take a look.

Gargron:
It takes more than 4 seconds to load "fediplay OR np" so something about these queries isn't quite right.

Gargron:
Oh and it seems to reload the column every time you navigate anywhere. I should open an issue for this.

boddhisattva:
@gdpelican With reference to the issue [here](https://github.com/mastodon/mastodon/issues/9279) I was keen to watch [the video](https://www.useloom.com/share/175fa307e42a431690745f41d61f4271) that you are referring to in your post [here](https://github.com/mastodon/mastodon/pull/8904#issue-367512728). On clicking the video link, I get: `It may have been deleted or archived by its creator.` I was wondering if you could kindly upload the video again and may be to a place like google drive or so where it may be available for relatively longer if one needs to better understand the work done as part of this PR?

PR #8912, opened 2018-10-07
rubocop issues - Cleaning up

shleeable:
Starting to get more into the code and make actual changes... all of these make sense to me - but as with my other PRs, you might have personal choice to keep using different code styles.

### DIFF
--- a/app/lib/settings/scoped_settings.rb
+++ b/app/lib/settings/scoped_settings.rb
@@ -47,7 +47,6 @@ module Settings
       record.update!(value: value)
 
       Rails.cache.write(Setting.cache_key(key, @object), value)
-      value
     end
 
     def [](key)
--- a/app/lib/settings/scoped_settings.rb
+++ b/app/lib/settings/scoped_settings.rb
@@ -47,6 +47,7 @@ module Settings
       record.update!(value: value)
 
       Rails.cache.write(Setting.cache_key(key, @object), value)
+      value
     end
 
     def [](key)
--- a/app/models/concerns/account_interactions.rb
+++ b/app/models/concerns/account_interactions.rb
@@ -209,6 +209,8 @@ module AccountInteractions
          .where('users.current_sign_in_at > ?', User::ACTIVE_DURATION.ago)
   end
 
+  private
+
   def remove_potential_friendship(other_account, mutual = false)
     PotentialFriendshipTracker.remove(id, other_account.id)
     PotentialFriendshipTracker.remove(other_account.id, id) if mutual
--- a/app/models/concerns/account_interactions.rb
+++ b/app/models/concerns/account_interactions.rb
@@ -209,8 +209,6 @@ module AccountInteractions
          .where('users.current_sign_in_at > ?', User::ACTIVE_DURATION.ago)
   end
 
-  private
-
   def remove_potential_friendship(other_account, mutual = false)
     PotentialFriendshipTracker.remove(id, other_account.id)
     PotentialFriendshipTracker.remove(other_account.id, id) if mutual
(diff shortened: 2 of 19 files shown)


config/initializers/trusted_proxies.rb
```diff
@@ -4,7 +4,7 @@ def trusted_proxy?(ip)
       if Rails.application.config.action_dispatch.trusted_proxies.nil?
         super
       else
-        Rails.application.config.action_dispatch.trusted_proxies.any? { |proxy| proxy === ip }
+        Rails.application.config.action_dispatch.trusted_proxies.any? { |proxy| proxy.include?(ip) }
```
Gargron: Why? This is not right. `proxy` is IPAddr, not an array.

config/initializers/open_uri_redirection.rb
```diff
@@ -2,7 +2,7 @@
 
 module OpenURI
   def self.redirectable?(uri1, uri2) # :nodoc:
-    uri1.scheme.downcase == uri2.scheme.downcase ||
+    uri1.scheme.casecmp(uri2.scheme) ||
```
Gargron: This is wrong. Output of casecmp is -1, 0, 1, not boolean. Need to `casecmp(...).zero?` for correct behaviour

app/models/concerns/account_interactions.rb
```diff
@@ -209,8 +209,6 @@ def lists_for_local_distribution
          .where('users.current_sign_in_at > ?', User::ACTIVE_DURATION.ago)
   end
 
-  private
```
Gargron: Why?

app/lib/settings/scoped_settings.rb
```diff
@@ -47,7 +47,6 @@ def []=(key, value)
       record.update!(value: value)
 
       Rails.cache.write(Setting.cache_key(key, @object), value)
-      value
```
Gargron: `Rails.cache.write` does not have a documented return value. I would need some reasoning for leaving  this function without a return value.

Gargron (changes requested):

app/lib/settings/scoped_settings.rb
ashleyhull-versent: ok, the fact this value is written and used as the return was lost on me.. It seems to alert in rubocop as "Variable `value` used in void context".

https://rubocop.readthedocs.io/en/latest/cops_lint/#lintvoid

app/models/concerns/account_interactions.rb
ashleyhull-versent: ok. I was certain this is just a typo... why is the private just sitting there between these two defs?

I'll revert this change.

config/initializers/open_uri_redirection.rb
ashleyhull-versent: thanks for the catch... I actually meant to .zero it

config/initializers/trusted_proxies.rb
ashleyhull-versent: sorry. I miss-read this.

ashleyhull-versent:
@Gargron uncommited those mistakes.

config/initializers/open_uri_redirection.rb
```diff
@@ -2,7 +2,7 @@
 
 module OpenURI
   def self.redirectable?(uri1, uri2) # :nodoc:
-    uri1.scheme.downcase == uri2.scheme.downcase ||
+    uri1.scheme.casecmp(uri2.scheme).zero? ||
```
ashleyhull-versent: This looks correct to me - but please confirm.

Gargron (approved):

PR #8922, opened 2018-10-08
Fix error when pinning the DM column

mayaeh:
After #8832 , if you tried to pin  the DM column in WebUI, the screen will not be displayed.

I wrote a PR to fix this.

However, when I look at the change in # 8832, it seems to be deleting part of the part related to pinning.
( Reference: https://github.com/tootsuite/mastodon/pull/8832/files#diff-8c49e70a7c6be1febaec5e02cc9dd4f2R87 )

I am worried about whether it is okay to fix this or if it is better to disable pinning.

### DIFF
--- a/app/javascript/mastodon/features/direct_timeline/index.js
+++ b/app/javascript/mastodon/features/direct_timeline/index.js
@@ -5,7 +5,7 @@ import Column from '../../components/column';
 import ColumnHeader from '../../components/column_header';
 import { expandConversations } from '../../actions/conversations';
 import { addColumn, removeColumn, moveColumn } from '../../actions/columns';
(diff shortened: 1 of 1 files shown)


Gargron:
Do you mean `trackScroll={!pinned}`? It could probably be re-added somehow, it is needed to disable scroll restoration because pinned columns don't need scroll restoration (because they are not in router)

mayaeh:
Maybe I had misunderstood about whether to disable DM column pinning.

I saw the changed code and thought that it might be that you disabled to pin it for load countermeasures.

Gargron:
No, i did not intend to disable pinning!

mayaeh:
I understand. Thank you!

Gargron (approved):

PR #8940, opened 2018-10-09
Do not push DMs into the home feed

Gargron:
This is a UX-oriented change. With conversations API, DMs now feel a lot more like DMs. Having them in the home timeline has always been a little strange and unnecessary.

Furthermore, we need to improve support for restricted-audience posts, i.e. private posts where not all followers are allowed to see it, but only an explicit list of recipients. From protocol view, that looks the same as a DM, but from the UX view, *those* are the posts that would need to be in the home feed. So this PR is a clean up in preparation for that.

### DIFF
--- a/app/javascript/mastodon/reducers/conversations.js
+++ b/app/javascript/mastodon/reducers/conversations.js
@@ -1,5 +1,7 @@
 import { Map as ImmutableMap, List as ImmutableList } from 'immutable';
 import {
+  CONVERSATIONS_MOUNT,
+  CONVERSATIONS_UNMOUNT,
   CONVERSATIONS_FETCH_REQUEST,
   CONVERSATIONS_FETCH_SUCCESS,
   CONVERSATIONS_FETCH_FAIL,
@@ -11,6 +13,7 @@ const initialState = ImmutableMap({
   items: ImmutableList(),
   isLoading: false,
   hasMore: true,
+  mounted: false,
 });
 
 const conversationToMap = item => ImmutableMap({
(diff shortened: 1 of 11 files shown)


trwnh:
I'm not sure this makes 100% sense, because aren't those direct posts still technically replies to something? I like having my home feed show the people I'm following, including their direct posts to me. It might be fine if they were filtered out only when the DM column is pinned, like with the notification column. And of course, it'd be best as an option.

nightpool:
> like with the notification column

Are you saying that if I unpin the notifications column, notifications will show up in my home feed? Is this true?

trwnh:
@nightpool No, I meant that DMs are made to not show in the notification column when the DM column is unpinned. Sorry for any confusion

nightpool (approved):

app/javascript/mastodon/reducers/conversations.js
```diff
@@ -11,6 +13,7 @@ const initialState = ImmutableMap({
   items: ImmutableList(),
   isLoading: false,
   hasMore: true,
+  mounted: false,
```
ThibG: This is slightly confusing as `mounted` is used as a number below.

app/services/fan_out_on_write_service.rb
```diff
@@ -6,15 +6,12 @@ class FanOutOnWriteService < BaseService
   def call(status)
     raise Mastodon::RaceConditionError if status.visibility.nil?
 
-    deliver_to_self(status) if status.account.local?
-
     render_anonymous_payload(status)
 
     if status.direct_visibility?
-      deliver_to_mentioned_followers(status)
-      deliver_to_direct_timelines(status)
```
ThibG: This means the old API won't be able to access newer direct toots. I thought we could have some transition period before dropping that?

Issue #8067, opened 2018-07-24
ActivityPub Audience targeting confusion


waweic:
First of all: I am relatively new to ActivityPub, so this question may be absolute nonsense, but I would love to get things cleared up.

Looking at any outbox, any `"to"` field on any object + activity shows `"https://www.w3.org/ns/activitystreams#Public"`
Meanwhile, the real addressees (like the person someone answered to or the own followers) are in the `"cc"` field. This looks like strange behaviour to me. Surely, the toots should be visible to the public, but primarily to the own followers or the person one answered to. So this should be the other way round, in my opinion.

This way, I see another problem. Something like [this](https://gist.github.com/waweic/d7c079876a402a51103df9e06a036b15) (obviously taken from gargron) with its audience targeting looks like a public comment and not like a reply because the public is first and the original creator is just in `"cc"`.

Is this wanted behaviour? If not, could it be changed easily? 

For boosts:
[W3C/ActivityPub/186](https://github.com/w3c/activitypub/issues/186) (I just wanted to add this as a possible additional source of information)

nightpool:
you can read https://github.com/tootsuite/mastodon/pull/3844#issuecomment-314897285 for a detailed rationale of mastodon's audience targeting choices. I don't see an issue, the difference between to and cc is mostly arbitrary anyway.

waweic:
I got another thought. The ActivityPub Spec [says](https://www.w3.org/TR/activitystreams-vocabulary/#h-audiencetargeting):

> The Primary audience consists of those entities directly involved or owning the object. 
> The Secondary audience consists of the collection of entities sharing an interest in the object but who might not be directly involved (e.g."followers"). 

This way, the public would not even qualify for the CC field.
Maybe, the ["audience"](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-audience) field would be more adequate in this case, because it

> Identifies one or more entities that represent the total  population of entities for which the object can considered to be  relevant.

Which would apply to the public in most cases.

Issue #7394, opened 2018-05-07
Activitypub interop: private posts are not delivered in mastodon if receiver is not mentioned


git-marijus:
If sending a private message via the "to" or "bto" property set to a subset of followers and those followers are not mentioned, the messages will not be delivered in mastodon.
The mention should probably just trigger notifications but not be mandatory for delivery. At least i did not see anything related in the activitypub spec. 

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

nightpool:
i mentioned this when preparing the impl report but unfortunately didn't have bandwidth to fix it at the time, so it got dropped. https://github.com/tootsuite/mastodon/issues/5631#issuecomment-343041015

should be a pretty easy change if someone wants to take and fix, we can just change it so that mentions are a necessary but not sufficient condition.

nightpool:
yes, because Mastodon's UX has no other way of exposing an "intended
audience" to the user, so it would be very confusing (and a possible spam
vector) if you were notified for a post that apparently had nothing to do
with you.

considering the scope of mastodon as a microblogging application, adding
the additional complex UX for arbitrary audience values is probably not a
desired feature

git-marijus:
> so it would be very confusing (and a possible spam vector) if you were notified for a post that apparently had nothing to do with you

That makes it impossible to share a private message with a subset of followers where one of them should be mentioned. One would have to mention (notify) all of them.

Delivering to "to" and notify the one who is mentioned would be considered less spam i guess. Mastodon could still use mentions for private messages and add the mentions to "to". That would not change anything regarding UX for mastodon but would make interoperability with projects which have a more complex privacy model smoother.

nightpool:
```
Mastodon could still use mentions for private messages and add the mentions to "to". That would not change anything regarding UX for mastodon but would make interoperability with projects which have a more complex privacy model smoother.
```

This is our current behavior, as far as I know. Do you have an example otherwise?

nightpool:
Hmm, unless i'm misunderstanding your proposal..... Are you saying mastodon should do that for posts it *originates*? or posts it *receives*?

git-marijus:
For posts it receives. Sorry if the title and toplevel post are not clear...

nightpool:
[...]

PR #8950, opened 2018-10-10
Improve support for aspects/circles

Gargron:
Improve support for aspects (feature that is allowed to exist in ActivityPub, and is implemented in Hubzilla and possibly other implementations). Allowed recipients are explicitly listed in the audience fields.

**Before:** A message with explicitly listed recipients would be delivered as a DM notification to all those recipients, but only as long as those were also tagged.

**After:** Save silent mentions, i.e. those that appear in `to`/`cc` but not in `tag` for access control, do not create a notification for silent mentions, push those into home feeds instead (only if the author is followed by the recipients), and ideally mark the post as "private" instead of "direct" in the REST API.

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -88,7 +88,7 @@ class FeedManager
     end
 
     query.each do |status|
-      next if status.direct_visibility? || filter?(:home, status, into_account)
+      next if status.direct_visibility? || status.limited_visibility? || filter?(:home, status, into_account)
       add_to_feed(:home, into_account.id, status)
     end
 
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -156,12 +156,12 @@ class FeedManager
     return true  if status.reply? && (status.in_reply_to_id.nil? || status.in_reply_to_account_id.nil?)
     return true  if phrase_filtered?(status, receiver_id, :home)
 
-    check_for_blocks = status.mentions.active.pluck(:account_id)
+    check_for_blocks = status.active_mentions.pluck(:account_id)
     check_for_blocks.concat([status.account_id])
 
     if status.reblog?
       check_for_blocks.concat([status.reblog.account_id])
-      check_for_blocks.concat(status.reblog.mentions.active.pluck(:account_id))
+      check_for_blocks.concat(status.reblog.active_mentions.pluck(:account_id))
     end
 
     return true if blocks_or_mutes?(receiver_id, check_for_blocks, :home)
@@ -188,7 +188,7 @@ class FeedManager
     # This filter is called from NotifyService, but already after the sender of
     # the notification has been checked for mute/block. Therefore, it's not
     # necessary to check the author of the toot for mute/block again
-    check_for_blocks = status.mentions.active.pluck(:account_id)
+    check_for_blocks = status.active_mentions.pluck(:account_id)
     check_for_blocks.concat([status.in_reply_to_account]) if status.reply? && !status.in_reply_to_account_id.nil?
 
     should_filter   = blocks_or_mutes?(receiver_id, check_for_blocks, :mentions)                                                         # Filter if it's from someone I blocked, in reply to someone I blocked, or mentioning someone I blocked (or muted)
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -156,12 +156,12 @@ class FeedManager
     return true  if status.reply? && (status.in_reply_to_id.nil? || status.in_reply_to_account_id.nil?)
     return true  if phrase_filtered?(status, receiver_id, :home)
 
-    check_for_blocks = status.mentions.pluck(:account_id)
+    check_for_blocks = status.mentions.active.pluck(:account_id)
     check_for_blocks.concat([status.account_id])
 
     if status.reblog?
       check_for_blocks.concat([status.reblog.account_id])
-      check_for_blocks.concat(status.reblog.mentions.pluck(:account_id))
+      check_for_blocks.concat(status.reblog.mentions.active.pluck(:account_id))
     end
 
     return true if blocks_or_mutes?(receiver_id, check_for_blocks, :home)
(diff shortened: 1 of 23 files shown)


ThibG:
Having it show as private might lead the receiving person to think every one of the author's followers are allowed to see that information.

Also, what about mentioned accounts that are not part of the audience?

Gargron:
- The look of the UI does not affect internal behaviour
- These messages are closer to private than DM in terms of UX (i.e. less personal, not to be highlighted)
- Adding a new visibility level would be a big API change in regards to apps

Whatever happens internally, it is easier to mask as "private" in the API to get the desired UX.

trwnh:
> push those into home feeds instead (only if the author is followed by the recipients)

I think I understand the reasoning for this, but there is one concern: if a message is shared without a mention, as is possible in other systems, then how does a Mastodon user know to follow the sender in order to receive the message?

Typically in those other social models, a notification is generated for "x has started sharing with you". This lets you know that someone has made you part of an audience, and you can then choose to receive those posts. But because Mastodon uses a follow-first model instead of a share-first model, only follower notifications are currently generated.

My interpretation would be that there should be an "incoming share requests" section, similar to "follower requests". This "incoming" section would show either the messages themselves directly as a column, or it would show just the profiles listed with a follow button.

Gargron:
>Typically in those other social models, a notification is generated for "x has started sharing with you". This lets you know that someone has made you part of an audience, and you can then choose to receive those posts. But because Mastodon uses a follow-first model instead of a share-first model, only follower notifications are currently generated.

>My interpretation would be that there should be an "incoming share requests" section, similar to "follower requests". This "incoming" section would show either the messages themselves directly as a column, or it would show just the profiles listed with a follow button.

Not interested in that at all. It's just a venue for spam. Marketing account sends share requests to every single person in the fediverse etc etc. The idea behind aspects/circles is that you choose which of your followers you share something with, not that you just arbitrarily pick random people to share with. Non-followers do not need to be supported in Mastodon.

trwnh:
>Non-followers do not need to be supported in Mastodon.

OK, this definitely matches up with my expectation. But might there be a difference between `to` and `cc` interpretations? If `to` is direct targeting and `cc` is visibility targeting, then would that logic apply only to rejecting messages from non-followers with `cc: you`? Or would messages `to: you` missing a mention still show up (in your notifications and/or your direct messages)? Related: #7394, #8067

Gargron:
I do not think that there is any meaningful difference between `to` and `cc` when we're not talking about a magic value like the public collection, that's why process_audience clumps them together. The deciding factor is the presence of a `tag`. I believe that this will work alright.

app/lib/feed_manager.rb
```diff
@@ -156,12 +156,12 @@ def filter_from_home?(status, receiver_id)
     return true  if status.reply? && (status.in_reply_to_id.nil? || status.in_reply_to_account_id.nil?)
     return true  if phrase_filtered?(status, receiver_id, :home)
 
-    check_for_blocks = status.mentions.pluck(:account_id)
+    check_for_blocks = status.mentions.active.pluck(:account_id)
```
nightpool: `status.active_mentions`? (with a scope_procedure on the other side)

app/lib/feed_manager.rb
```diff
@@ -156,12 +156,12 @@ def filter_from_home?(status, receiver_id)
...
     check_for_blocks.concat([status.account_id])
 
     if status.reblog?
       check_for_blocks.concat([status.reblog.account_id])
-      check_for_blocks.concat(status.reblog.mentions.pluck(:account_id))
+      check_for_blocks.concat(status.reblog.mentions.active.pluck(:account_id))
```
nightpool: ibid, `status.reblog.active_mentions`

app/models/mention.rb
```diff
@@ -18,6 +19,9 @@ class Mention < ApplicationRecord
 
   validates :account, uniqueness: { scope: :status }
 
+  scope :active, -> { where(silent: false) }
+  scope :passive, -> { where(silent: true) }
```
nightpool: maybe just `active` and `silent`? to avoid introducing another new concept

app/serializers/activitypub/note_serializer.rb
```diff
@@ -68,7 +68,7 @@ def cc
   end
 
   def virtual_tags
-    object.mentions.to_a.sort_by(&:id) + object.tags + object.emojis
+    object.mentions.to_a.reject(&:silent?).sort_by(&:id) + object.tags + object.emojis
```
nightpool: `object.active_mentions.to_a`

app/serializers/rest/status_serializer.rb
```diff
@@ -88,7 +99,7 @@ def pinnable?
   end
 
   def ordered_mentions
-    object.mentions.to_a.sort_by(&:id)
+    object.mentions.to_a.reject(&:silent?).sort_by(&:id)
```
nightpool: `object.active_mentions`

app/serializers/rest/status_serializer.rb
```diff
@@ -36,6 +36,17 @@ def current_user?
     !current_user.nil?
   end
 
+  def visibility
+    # This visibility is masked behind "private"
+    # to avoid API changes because there are no
+    # UX differences
```
nightpool: I think it makes sense to add a `limited_audience` array to the API, as a 1) backwards compatible way of indicating the correct visibility and 2) a way for apps to display the list of users who have access to a post.

app/workers/activitypub/reply_distribution_worker.rb
```diff
@@ -21,7 +21,7 @@ def perform(status_id)
   private
 
   def skip_distribution?
-    @status.private_visibility? || @status.direct_visibility?
+    !(@status.public_visibility? || @status.unlisted_visibility?)
```
nightpool: On line 12: `return unless account.present? && distribute?`

here: `def distribute?; @status.public_visibility || @status.unlisted_visibility?`

nightpool (dismissed):

app/serializers/rest/status_serializer.rb
Gargron: I don't think apps should be able to display a list of users who have access to a post.

app/serializers/activitypub/note_serializer.rb
Gargron: I don't want to touch the eager loading code. Mentions is eager loaded, so it's easier to filter the array than mess with the query causing a n+1

nightpool: active_mentions could also be eager loaded, i think.

Gargron:
@ThibG Yes, I think so.

ykzts (approved):

PR #8951, opened 2018-10-10
Move network calls out of transaction in ActivityPub handler

Gargron:
Mention and emoji code may perform network calls, but does not need
to do that inside the database transaction. This may improve availability
of database connections when using pgBouncer in transaction mode.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -22,12 +22,16 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   private
 
   def process_status
-    status_params = process_status_params
+    @tags     = []
+    @mentions = []
+    @params   = {}
+
+    process_status_params
+    process_tags
 
     ApplicationRecord.transaction do
-      @status = Status.create!(status_params)
-
-      process_tags(@status)
+      @status = Status.create!(@params)
+      attach_tags(@status)
     end
 
     resolve_thread(@status)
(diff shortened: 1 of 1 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -42,62 +46,77 @@ def find_existing_status
...
-      sensitive: @object['sensitive'] || false,
-      visibility: visibility_from_audience,
-      thread: replied_to_status,
-      conversation: conversation_from_uri(@object['conversation']),
-      media_attachment_ids: process_attachments.take(4).map(&:id),
-    }
-  end
-
-  def process_tags(status)
+    @params = begin
```
nightpool: why `begin` here? shouldn't this just be `@params = {` ?

nightpool (approved):

nightpool:
approve modulo formatting

app/lib/activitypub/activity/create.rb
Gargron: Otherwise rubocop will require me to align the hash with the `=` which sucks. begin/end does nothing except improve the formatting.

nightpool: which rule was it? `Layout/AlignHash` definitely doesn't work that way. We should make sure to change the rubocop lint formatting when we run into a situation where it's behaving badly, so that we're consistent across PRs

Gargron: As far as I am aware we are required to:

```ruby
hash = {
           foo: bar,
           baz: boo,
         }
```

Which looks bad.

nightpool: My understanding is that the default Layout/AlignHash and [Layout/IndentHash](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/IndentHash) rule should pass with this code:

    hash = {
      foo: bar,
      baz: boo,
    }

PR #8957, opened 2018-10-11
Set Content-Security-Policy rules through RoR's config

ClearlyClaire:
This is ported from glitch-soc, where this seems to work.
It may have been insufficiently tested with some configurations, though.


Gargron (approved):

33b5e5:
FYI, this change causes an error with embeds from external instances:

`Content Security Policy: The page’s settings blocked the loading of a resource at https://toot-lab.reclaim.technology/embed.js (“script-src”).`

`Loading failed for the <script> with source “https://toot-lab.reclaim.technology/embed.js”.`

To reproduce, find a toot with media that's from an instance other than your own, then choose "embed" from the ... overflow menu. You can view the errors in the console. 

I'm not sure how much it matters though? I almost feel like blocking third party JS here is a feature. I can live without external formatting.

33b5e5:
This is probably much worse: the change also prevents media from loading from object storage.

`Content Security Policy: The page’s settings blocked the loading of a resource at https://s3.us-west-1.wasabisys.com/REDACTED/FILENAME.mp4 (“media-src”).`

I can think of multiple ways to resolve that, including by proxying the media, but it seems bad to have a default CSP that breaks external S3 storage.

nightpool:
media_src includes assets_host by default, so i'm not sure what the problem could be there. Maybe there's a difference in asset_host and paperclip configuration?

33b5e5:
@nightpool I'm not sure, but as far as I can tell, assets_host only references my instance's hostname and not the S3 hostname.

rugk:
So has this been tested with https://csp-evaluator.withgoogle.com/ or other tools?

ThibG:
@rugk yes. I invite you to try out yourself with one of the many 2.6.1 instances out there.
It could probably be made stricter thanks to CSP2 hash/nonces or CSP3 strict-dynamic, but this would involve more webpack trickery than I know of.

PR #8959, opened 2018-10-11
Improve signature verification safeguards

Gargron:
- Enforce a 12 hours time window on signed requests to prevent replayability of captured requests. **Low importance**, since HTTPS normally prevents MITM attacks that would allow capturing requests.
- Ensure that the list of signed headers in the signature header is downcased. A differently-cased list could allow the sender to side-step the special-case verification of the `Digest` header. **Low importance**, since HTTPS normally prevents MITM attacks that would allow tampering with the request.

### DIFF
--- a/spec/controllers/concerns/signature_verification_spec.rb
+++ b/spec/controllers/concerns/signature_verification_spec.rb
@@ -73,6 +73,30 @@ describe ApplicationController, type: :controller do
       end
     end
 
+    context 'with request older than a day' do
+      before do
+        get :success
+
+        fake_request = Request.new(:get, request.url)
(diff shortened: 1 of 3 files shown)


spec/controllers/concerns/signature_verification_spec.rb
```diff
@@ -73,6 +73,30 @@ def alternative_success
       end
     end
 
+    context 'with request older than a day' do
```
hellekin: `#matches_time_window?` uses `12.hours.ago`, spec mentions "older than a day", and the spec itself sets the header to `2.days.ago`.

PR #8967, opened 2018-10-12
CSP fixes

ClearlyClaire:


ThibG:
Oh, haven't seen that you already have fixed some of that.
Hm, you could either take this or merge the missing bits I guess.

config/initializers/content_security_policy.rb
```diff
@@ -2,17 +2,34 @@
...
-#
-#   # Specify URI for violation reports
-#   # p.report_uri "/csp-violation-report-endpoint"
-# end
+if Rails.env.production?
+  assets_host = Rails.configuration.action_controller.asset_host || "https://#{ENV['WEB_DOMAIN'] || ENV['LOCAL_DOMAIN']}"
+  data_hosts = [assets_host]
+
+  if ENV['S3_ENABLED'] == 'true'
+    attachments_host = ENV['S3_ALIAS_HOST'] || ENV['S3_CLOUDFRONT_HOST'] || ENV['S3_HOSTNAME'] || "s3-#{ENV['S3_REGION'] || 'us-east-1'}.amazonaws.com"
```
33b5e5: What is the trailing `'us-east-1'}.amazonaws.com"` for?

config/initializers/content_security_policy.rb
```diff
@@ -2,17 +2,34 @@
...
+    p.default_src     :none
+    p.frame_ancestors :none
+    p.script_src      :self, assets_host
+    p.font_src        :self, assets_host
+    p.img_src         :self, :https, :data, :blob
+    p.style_src       :self, :unsafe_inline, assets_host
+    p.media_src       :self, :data, *data_hosts
+    p.frame_src       :self, :https
+    p.worker_src      :self, assets_host
+    p.connect_src     :self, :blob, Rails.configuration.x.streaming_api_base_url, *data_hosts
```
33b5e5: Adding the `:blob` to connect_src resolves another error I was running into after previous CSP changes. The error was like:

`Content Security Policy: The page’s settings blocked the loading of a resource at blob:https://$my_instance/$redacted_url (“connect-src”).`

I was seeing this when trying to upload a JPG attachment. So 👍 on the `:blob` addition.

config/initializers/content_security_policy.rb
```diff
@@ -2,17 +2,34 @@
...
+    attachments_host = ENV['SWIFT_OBJECT_URL']
+  else
+    attachments_host = nil
+  end
+  data_hosts << attachments_host unless attachments_host.nil?
+
+  Rails.application.config.content_security_policy do |p|
+    p.base_uri        :none
+    p.default_src     :none
+    p.frame_ancestors :none
```
mayaeh: I think that frame_ancestors is necessary for the embed of toot.

On my server I set "frame-ancestors https:" and running the Mastodon.

ThibG: The controller responsible for handling embedded toots overwrites this!

mayaeh: I'm sorry, I did not know that.

config/initializers/content_security_policy.rb
Gargron: I do not believe that connect-src should affect image resizing. That should be img-src...

ThibG: But it does. I don't remember why exactly, it might be the `readAsDataURL` or `canvas.toBlob`.

config/initializers/content_security_policy.rb
Gargron: I have already merged a fix for this that does not rely on composing these URLs manually, by whitelisting https. Whitelisting https for images and videos is better than no CSP at all. At the same time, predicting which URLs it's gonna be is very complicated, because there's S3, Wasabi, Google Cloud, Swift, etc, and afaik no way to get that out of Paperclip without reading the environment variables again and essentially duplicating its logic. So, imo not worth it. Just whitelist https.

Gargron:
I think #8962 has already solved everything

ThibG:
Nope, image resizing is still broken, fails with `Content Security Policy: The page’s settings blocked the loading of a resource at blob:https://mastodev.sitedethib.com/591a57e7-1f70-43e9-bc59-600cb3c37d1c (“connect-src”).` in both Firefox and Chromium. Everything else should be fine I guess.

Gargron:
@ThibG In that case please re-submit that addition on top of master

Gargron (approved):

PR #8978, opened 2018-10-13
Unlink socket file when exists.

yukimochi:
When exists socket file which created in previous process, streaming raise error looping and use 100% cpu time.

This avoid above problem by unlink existing files which given socket path.

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -617,6 +617,9 @@ const startWorker = (workerId) => {
   }, 30000);
 
   if (process.env.SOCKET || process.env.PORT && isNaN(+process.env.PORT)) {
+    if (process.env.SOCKET && fs.existsSync(process.env.SOCKET)) {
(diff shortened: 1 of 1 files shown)


Gargron:
The same problem occurs when port is taken, the workers try to restart infinitely.

Does other software also delete the socket on startup? What if you accidentally start two at the same time, I feel like the socket should be a safeguard against that.

So maybe it's possible to check existence of socket / availability of port in the master process before clustering happens?

yukimochi:
I don't know ideal treat ...
But in puma, when existing file given socket path it delete it and start.

 - In puma
clustering happen -> `There is already a server bound to: socket/puma.sock (RuntimeError)` in secondary
(Previous process is still working well.)
 - In node
clustering happen -> `ERR! Error: bind EADDRINUSE socket/node.sock` in secondary
(Previous process is still working well.)

I think this changes makes no effect about clustering problem.

PR #8979, opened 2018-10-14
RTL: fix toot alignments in public pages, fixes #2350

### LINKED ISSUE #2350: RTL support for post contents in profile pages and individual posts view

mabkenar:
Since 2.5.0, RTL toots in public pages are broken. They are "direction; rtl", but "text-align: left".
I have tested this commit on RTL profiles (solves the issue indeed) and on LTR profiles (as far as I can see, no visible side effects whatsoever). 

Please also check that I'm not breaking something inadvertently (it seems to me that I'm not).

An [RTL profile](https://mastodon.social/@mansoor) before (left-justified :unamused:):
selection_595
and after (correct text alignment):
selection_594

An [LTR profile](https://social.tchncs.de/@milan) before:
milan_before
and after (no difference):
milan_after

### DIFF
--- a/app/javascript/styles/mastodon/stream_entries.scss
+++ b/app/javascript/styles/mastodon/stream_entries.scss
@@ -3,7 +3,6 @@
   border-radius: 4px;
   overflow: hidden;
   margin-bottom: 10px;
-  text-align: left;
 
   @media screen and (max-width: $no-gap-breakpoint) {
     margin-bottom: 0;


herotux:
:rose:

Gargron (approved):

BoFFire:
Thank you !

PR #8993, opened 2018-10-16
always allow DMs from staff

### LINKED ISSUE #8969: Allow moderators to always send direct messages to local users
Users can disable direct messages from people they don't follow. That is of course a good privacy feature, but moderators need to be able to send direct messages to local users. Of course moderators can decide not to use direct messages, but moderation in the open is mostly not very productive.

So please allow moderators to always send direct messages to local users.

* * * *

qguv:
For moderation purposes, staff (mods and admins) will still be able to DM users, even when they have restricted their DMs to followers. Closes #8969.

### DIFF
--- a/app/services/notify_service.rb
+++ b/app/services/notify_service.rb
@@ -59,9 +59,14 @@ class NotifyService < BaseService
     @notification.target_status.in_reply_to_account_id == @recipient.id && @notification.target_status.thread&.direct_visibility?
   end
 
+  def from_staff?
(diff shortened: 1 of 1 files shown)


Gargron (approved):

jeroenpraat:
Thanks! :smiley:

mayaeh:
@qguv @Gargron I think that this will not reach if the destination checks "Block notifications from non-followers" at the same time.

The same applies to "Block notifications from people you do not follow".

PR #8994, opened 2018-10-16
Fix some bad localization strings

### LINKED ISSUE #8887: /auth/edit broken when language set to Finnish
As a user, when mastodon's UI language is set to Finnish, security page in user preferences (/auth/edit) returns an error page.

All other languages work fine, and browser language has no effect. I have tested this on both firefox and chrome, on multiple servers running 2.5.0.

* * * *

-  [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

qguv:
Fixes #8887.


Gargron (approved):

Gargron:
I wish there was a way to prevent this from happening once and for all. Or at least detect it. But it's like, not viable to re-run the test suite in every locale.

PR #9010, opened 2018-10-19
Show suggested follows on search screen in mobile layout

Gargron:
Reminder: Suggestions were added in #7918 and are based on who you interact with who you do not follow. E.g. if you boost someone a lot from seeing other people's boosts of that person, it makes sense you might be interested in following the original source; or if you reply to someone a lot, maybe you'd want to follow them

Each suggestion can be dismissed


ykzts (approved):

jeroenpraat:
I probably missed something, but why only in the mobile layout and not in the desktop layout?

PR #9018, opened 2018-10-19
RTL: fix user stats in about page

mabkenar:
Before:
image

After:
image


mabkenar:
Would be nice if numbers are also converted to their Persian versions, e.g., 67 to ۶۷. Could you please give me a hint how can I do that? React should have good libraries for such things.

ykzts (approved):

PR #9020, opened 2018-10-19
Fix admin account avatar margin in about page

smorimoto:
| Before | After |
| :-----: | :----: |
|  |


Gargron:
There should be no left margin ("A" aligned with avatar and "C" below)

Gargron (approved):

PR #9021, opened 2018-10-19
RTL: fix preferences layout

mabkenar:
This was heavier compared to previous ones. Please test and make a quick visual inspection if the resulting layout is indeed like the second screenshot below.

Before:
image

After:
image


Gargron (approved):

BoFFire:
Nice ! Thank you !

PR #9033, opened 2018-10-21
Fix domain label position and color

kedamaDQ:
Fix domain label position which is hiding the border of the input form.

before:
ltr-before

after:
ltr-after

In the RTL, fix position and color mismatch of domain label.

before:
rtl-before

after:
rtl-after


ykzts (approved):

kedamaDQ:
Oh, this is fix for firefox users.

PR #9046, opened 2018-10-21
Do not move CWs to toot body when toot body is empty

### LINKED ISSUE #6825: Do not use content warning as body when body is empty
In Mastodon 2.3.0, CWs are not behaving as expected when the body of the toot is absent but an image is attached.

Input: I add a CW to a toot, attach an image, and leave the body of the toot blank.

Expected behavior: The image is NSFW-flagged, the CW is applied as a standard CW, and the body of the toot remains empty.

Observed behavior: The image is NSFW-flagged, the CW is removed, and the text of the CW is added as the body text of the toot.

Additional information: a toot with a CW but no body text and no image attached won't post at all, even though the Toot button is lit up, since nominally there is no content.

* * * *

ClearlyClaire:
Instead of leaving the toot body blank, it replaces it with a placeholder text such as 📹 or 🖼 according to the type of attached media.

This is a port of https://github.com/glitch-soc/mastodon/pull/788 and a way to fix #6825

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -21,7 +21,7 @@ class PostStatusService < BaseService
 
     media  = validate_media!(options[:media_ids])
     status = nil
-    text   = options.delete(:spoiler_text) if text.blank? && options[:spoiler_text].present?
+    text   = '.' if text.blank? && options[:spoiler_text].present?
 
     ApplicationRecord.transaction do
       status = account.statuses.create!(text: text,
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -21,7 +21,10 @@ class PostStatusService < BaseService
 
     media  = validate_media!(options[:media_ids])
     status = nil
-    text   = '.' if text.blank? && options[:spoiler_text].present?
+    if text.blank? && options[:spoiler_text].present?
+      text = '.'
+      text = media.find(&:video?) ? '📹' : '🖼' unless media.empty?
+    end
 
     ApplicationRecord.transaction do
       status = account.statuses.create!(text: text,


nightpool:
Placeholder text is a really bad idea, because you're putting words/symbols
into a user's mouth that they never meant to say. I would be much more in
favor of just falling back to '.', if we *really* think having the show
more button is meaningful here (I'm still not convinced that it is)

ThibG:
Tbh, I'm torn about the placeholders, but I implemented them after feedback from some glitch-soc users.
The whole “use a non-empty content even though the content was originally empty” was originally to accommodate for Mastodon's handling of CW. I proposed other alternatives but they did not seem to get any traction, so I just ported what I made for glitch-soc.

ThibG:
Crap I tried to delete duplicate comments from @trwnh, but github deleted *all* of them, including the original one?

Anyway, the comment read as
>> The whole “use a non-empty content even though the content was originally empty” was originally to accommodate for Mastodon's handling of CW.
>
> Wasn't it because older versions / OStatus couldn't support empty content? There's no technical limitation against empty content, and there's no technical limitation against empty content with CW (except for the current conversion behavior).

Mastodon used a placeholder “.” for this reason at some point in the past, but I was talking about the reason for using placeholders in *this PR*, and that reason is purely having the “hide/show” spoiler button do something that isn't confusing.

ThibG:
Crap I tried to delete duplicate comments from @trwnh, but github deleted *all* of them, including the original one?

Anyway, the comment read as
>> The whole “use a non-empty content even though the content was originally empty” was originally to accommodate for Mastodon's handling of CW.
>
> Wasn't it because older versions / OStatus couldn't support empty content? There's no technical limitation against empty content, and there's no technical limitation against empty content with CW (except for the current conversion behavior).

Mastodon used a placeholder “.” for this reason at some point in the past, but I was talking about the reason for using placeholders in *this PR*, and that reason is purely having the “hide/show” spoiler button do something that isn't confusing.

trwnh:
>that reason is purely having the “hide/show” spoiler button do something that isn't confusing.

Ahh, ok. So we can still look at all the possible outcomes:

State | Text only | Media only | Text and media | No text, no media
--- | --- | --- | ---
CW closed | "Show more" to expand text | "Show more 🖼" to display thumbnail | "Show more 🖼" to display content/attachment | ???
CW open | "Show less" to hide text | "Show less 🖼" to hide thumbnail | "Show less 🖼" to hide content/attachment | ???

i.e. in no-content cases where media exists, the media can be hidden. But in no-content cases where no media exists, it's possible to just disallow this, and convert incoming summary/nocontent messages to content, without losing semantic meaning. IMO, the real issue is with the media cases, because for those, summary and content are not really the same; not so much with text-only cases.

PR #9038, opened 2018-10-21
RTL: remove blank character inside bdi

mabkenar:
When RTL layout is active, an account with LTR display-name is rendered incorrectly. I'm talking about this:
bdi_before

Before:
image

After:
image

Or maybe I am getting it wrong, and _this_ is not the blank character I should remove?

PR #9052, opened 2018-10-22
Add styles to suffix of display name

ykzts:
Additional to #9038

### DIFF
--- a/app/javascript/mastodon/components/__tests__/__snapshots__/display_name-test.js.snap
+++ b/app/javascript/mastodon/components/__tests__/__snapshots__/display_name-test.js.snap
@@ -14,12 +14,15 @@ exports[`<DisplayName /> renders display name + account name 1`] = `
       }
     />
   </bdi>
-   
   <span
-    className="display-name__account"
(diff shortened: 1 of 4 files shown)


Gargron:
I just have to wonder if this is the best fix, or if I should just revert #9038. A space is supposed to be more natural there...

ykzts:
I think that space should be displayed even in rtl.

mabkenar:
Yes, a space should be there, but not inside bdi element. Outside of it, so that when page is RTL and display name is in latin, the space follows the RTL flow, instead of being trapped inside the bdi box.

Gargron:
But the space *wasn't* inside the bdi tag, it was after the tag (between bdi and span)

mabkenar:
> But the space _wasn't_ inside the bdi tag, it was after the tag (between bdi and span)

Oh I see the root for my confusion: In the web app, space is outside the bdi tag (correct). In public pages, it is inside the bdi tag (incorrect). This is what I see right now on my instance at v2.5.2.

PR #9059, opened 2018-10-22
Reduce server load caused by anonymous viewing.

### LINKED ISSUE #9035: Make public profile pages proxy-cacheable
### Pitch

Non-PWA toot pages currently have `Cache-Control: max-age=0, private, must-revalidate`, which means a caching proxy will never save a copy. Changing this to `max-age=0, public` will allow proxies to cache the pages and serve them if the site goes down or is otherwise unable to serve a response.

### Motivation

If the site goes offline or is too slow, a compliant caching proxy currently has no choice but to keep pounding the server. I'd like to see the pages be cacheable instead.

There is very little that changes on a public page between page loads. Other than a new toot being made (which can only happen when the site is up), the only things that change are the buttons (follow, log in, register, log out, etc.). Those could be handled by a script that runs after page load because logged-in users are guaranteed to have JavaScript support.

### Proof-of-concept

1. Set up a Mastodon instance behind Cloudflare with the Always Online setting enabled and a page rule enabling "origin cache control" (all available on the free plan).
2. Browse the Mastodon site for a while, then turn off the Mastodon web service. Continue trying to browse.
3. Most pages will return a generic Cloudflare "the site is down and there's nothing we can do about it" message. However, for some reason, user account pages (eg. https://mastodon.social/@Gargron) are cached in some sort of JSON format.

BenLubar:
Do not start a session if the current user is not logged in for public-facing pages.

Mark pages that don't care about sessions as publicly cacheable.

Keep the max age as 0 so proxies and browsers will still try to retrieve an updated version but can still fall back to the stale version if the site is down or too slow.

Fixes #9035.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -151,6 +151,11 @@ class ApplicationController < ActionController::Base
     response.headers['Vary'] = 'Accept'
   end
 
+  def mark_cacheable!
+    skip_session!
+    expires_in 0, public: true
+  end
+
   def skip_session!
     request.session_options[:skip] = true
   end
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -10,6 +10,8 @@ class AccountsController < ApplicationController
   def show
     respond_to do |format|
       format.html do
+        mark_cacheable! unless user_signed_in?
+
         @body_classes      = 'with-modals'
         @pinned_statuses   = []
         @endorsed_accounts = @account.endorsed_accounts.to_a.sample(4)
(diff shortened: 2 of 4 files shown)


Gargron:
@nightpool Is this safe?

nightpool:
I don't understand why we're coupling it to skip_session, that feels like
overloading the helper too much.

I'll have to think about safety a little bit more. at the very least, we
should make sure to add the logged in status to whatever cache key we're
sending.

app/controllers/application_controller.rb
```diff
@@ -152,5 +152,6 @@ def set_cache_headers
 
   def skip_session!
     request.session_options[:skip] = true
+    expires_in 0, public: true
```
Gargron: Could you create a new method that encompasses `skip_session!` and `expires_in 0` instead of modifying this? Would be clearer. You can call it `mark_cacheable!` or something.

Gargron (changes requested):

ClearlyClaire (approved):

Gargron (approved):

PR #9063, opened 2018-10-23
Redis 5.0.0 is out.

shleeable:
I'm going to test the straight up replacement.. might be worth waiting for 5.0.1, but the Mastodon usecase isn't using any of the more complex caching procedures.

Should be able to be swapped out easy enough.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
 
   redis:
     restart: always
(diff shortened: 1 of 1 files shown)


Gargron:
I vaguely heard about some deployment issues with the new Redis, which is why I am hesitating to merge this until there is more info.

Gargron (approved):

PR #9075, opened 2018-10-24
Fix RTL layout of status display names

### LINKED ISSUE #2350: RTL support for post contents in profile pages and individual posts view

Gargron:
Fix #2350

### DIFF
--- a/app/javascript/styles/mastodon/rtl.scss
+++ b/app/javascript/styles/mastodon/rtl.scss
@@ -130,17 +130,6 @@ body.rtl {
     float: left;
   }
 
-  .activity-stream .detailed-status.light .detailed-status__display-name > div {
-    float: right;
-    margin-right: 0;
-    margin-left: 10px;
-  }
-
-  .activity-stream .detailed-status.light .detailed-status__meta span > span {
-    margin-left: 0;
-    margin-right: 6px;
-  }
-
   .status__action-bar {
 
     &__counter {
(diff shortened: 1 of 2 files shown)


ykzts (approved):

app/javascript/styles/mastodon/rtl.scss
```diff
@@ -174,6 +163,10 @@ body.rtl {
     margin-right: 0;
   }
 
+  .detailed-status__display-name .display-name {
+    text-align: right;
```
mabkenar: I don't think this is needed here. If rtl styles are loaded, then the text here will be dir=rtl (by inheritance) and align=left (I assume by CSS convention). No?

app/views/stream_entries/_simple_status.html.haml
```diff
@@ -15,6 +15,7 @@
         %span.display-name
           %bdi
             %strong.display-name__html.p-name.emojify= display_name(status.account, custom_emojify: true, autoplay: autoplay)
+          &nbsp;
```
mabkenar: This `&nbsp;` forces that there always be at least one space character between display name and the account address, so I can accept is as a good-enough solution.

However, for **L**TR styles, it'll cause _two_ spaces between name and address. Moreover, for RTL styles, the original space at the right side of Display name would remain.

I don't know how to fix the above two (minor) problems. But I am already happy with your solution. Thanks!

PR #9085, opened 2018-10-24
Migrate all old direct messages to new conversations schema

Gargron:
First it creates a conversation record for every local DM's author (so everyone can keep track of DMs they sent). Then it iterates over notifications about DMs to create corresponding conversation records. This allows us to not mess with NotifyService filters on the fly, because if a notification exists, it can be shown.

### DIFF
--- a/db/migrate/20181024224956_migrate_account_conversations.rb
+++ b/db/migrate/20181024224956_migrate_account_conversations.rb
@@ -1,0 +1,43 @@
+class MigrateAccountConversations < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    say ''
+    say 'WARNING: This migration may take a *long* time for large instances'
+    say 'It will *not* lock tables for any significant time, but it may run'
+    say 'for a very long time. We will pause for 10 seconds to allow you to'
+    say 'interrupt this migration if you are not ready.'
+    say ''
+
+    10.downto(1) do |i|
+      say "Continuing in #{i} second#{i == 1 ? '' : 's'}...", true
+      sleep 1
+    end
+
+    local_direct_statuses.find_each do |status|
+      AccountConversation.add_status(status.account, status)
+    end
+
+    notifications_about_direct_statuses.find_each do |notification|
+      AccountConversation.add_status(notification.account, notification.target_status)
+    end
+  end
+
+  def down
+  end
+
+  private
+
+  def local_direct_statuses
+    Status.unscoped
+          .local
+          .where(visibility: :direct)
+          .includes(:account, mentions: :account)
+  end
+
+  def notifications_about_direct_statuses
+    Notification.joins(mention: :status)
+                .where(activity_type: 'Mention', statuses: { visibility: :direct })
+                .includes(:account, mention: { status: [:account, mentions: :account] })
+  end
+end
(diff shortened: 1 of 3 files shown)


ykzts (approved):

db/migrate/20181024224956_migrate_account_conversations.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    local_direct_statuses.find_each do |status|
+      AccountConversation.add_status(status.account, status)
+    end
+
+    notifications_about_direct_statuses.find_each do |notification|
+      AccountConversation.add_status(notification.account, notification.target_status)
+    end
+  end
```
ThibG: What about statuses being added out-of-order?

db/migrate/20181024224956_migrate_account_conversations.rb
```diff
@@ -0,0 +1,43 @@
...
+          .local
+          .where(visibility: :direct)
+          .includes(:account, mentions: :account)
+  end
+
+  def notifications_about_direct_statuses
+    Notification.joins(mention: :status)
+                .where(activity_type: 'Mention', statuses: { visibility: :direct })
+                .includes(:account, mention: { status: [:account, mentions: :account] })
+  end
```
ThibG: What about deleted notifications (of statuses that would still be in the DM column)?

db/migrate/20181024224956_migrate_account_conversations.rb
Gargron: Batch query enforces own order and it's based on id and i think it's ascending.

db/migrate/20181024224956_migrate_account_conversations.rb
Gargron: The alternative is not worth it

db/migrate/20181024224956_migrate_account_conversations.rb
ThibG: Any local toot will still be added before any remote toot, regardless of their actual order, right?
This means that if the last toot of a conversation was a local toot, it won't actually show, and that removing toots from the conversation will lead to strange behaviors.
It also means that for servers running on master or the previous RC, old toots will pop up on top of existing conversations.

db/migrate/20181024224956_migrate_account_conversations.rb
ThibG: I don't know, we currently have a way to clear the notifications, so people might do that.

db/migrate/20181024224956_migrate_account_conversations.rb
Gargron: You forget about the status_ids are sorted before being saved, so inserting out of order should be fine. Updates will be streamed though but only if you are connected to the direct stream and I think even then it's a minor one-off issue.

db/migrate/20181024224956_migrate_account_conversations.rb
Gargron: No I mean the alternative is running the NotifyService code which performs a lot of checks that are not optimized for n+1. It's a lot faster to iterate over already existing notifications. Also if you deleted the notifications it might be you don't wanna see those DMs anymore anyway. Overall I think it's the most optimal choice for the migration.

db/migrate/20181024224956_migrate_account_conversations.rb
ThibG: I'm sorry, I can't find where they are sorted.

Gargron: https://github.com/tootsuite/mastodon/blob/master/app/models/account_conversation.rb#L99-L102

PR #9086, opened 2018-10-25
Using Auto NSFW Option

S-H-GAMELINKS:
Auto NSFW option(Using Google Cloud Vision API)

Using like [this] (https://s3-us-east-2.amazonaws.com/creatodon-storage.com/media_attachments/files/000/439/203/original/99be7ec8ffe127ec.mp4)


nightpool:
1. putting complicated logic into the controller like this is directly against the MVC design principles of rails

2. the way this code is written right now is extremely fragile, you should be using the url that's built into the media attachment instead of trying to guess it on your own.

3. Mastodon will not add code that's this heavily dependent on a single cloud service provider, and this feature in general runs counter to mastodon's design principles—it delegates the safety and moderation of sensitive imagery to an unknowable algorithm. Users *need* to be responsible for the content they post.

4. this feature as implement has really bad UX—users set a sensitive or not sensitive state when they post the status, but it gets overridden by the opaque algorithmic determination, and their NSFW status is silently discarded. This is really, really bad—what if the status is about current events, or something specific to instance policy that the cloud vision API has no way of detecting?

I don't believe this feature can be accepted into mainline mastodon in this form.

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -83,4 +86,45 @@ def status_params
...
+
+    return paths
+  end
+
+  def check_nsfw(paths)
+    keys = JSON.parse(File.open(ENV['VISION_KEYFILE']).read).to_h
+
+    vision = Google::Cloud::Vision.new project: keys['project_id']
+
+    paths.each do |path|
```
nightpool: `paths = MediaAttachment.where(id: media_ids).map { |attachment| full_asset_url(attachment.file.url(:original))`

S-H-GAMELINKS: thanks!

S-H-GAMELINKS:
Thanks, review. 

Close this PR

PR #9093, opened 2018-10-25
Allow inbox owner to view implicitly targeted ActivityPub payload

### LINKED ISSUE #9091: Delivering to the personal inbox should add this person as receiver
### Expected behaviour

When an activity is send to a personal inbox (and not to the shared one), that person should be added automatically to the list of receivers. This is important for delivering content via ```bcc```.

According to the specifications, ```bcc``` must be discarded upon delivery. So delivering to the personal inbox is the only option so that the activity will reach it's intended destination. See also the discussion here: https://github.com/w3c/activitypub/issues/326

### Actual behaviour

The activity is not delivered to the receiver.

### Specifications

v2.6.0rc1

Gargron:
Fix #9091 

Please review thoroughly... Does this logic make sense?

### DIFF
--- a/app/controllers/activitypub/inboxes_controller.rb
+++ b/app/controllers/activitypub/inboxes_controller.rb
@@ -36,6 +36,6 @@ class ActivityPub::InboxesController < Api::BaseController
   end
 
   def process_payload
-    ActivityPub::ProcessingWorker.perform_async(signed_request_account.id, body.force_encoding('UTF-8'))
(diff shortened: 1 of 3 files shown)


ClearlyClaire (approved):
Still not 100% sure about tying audience with mentions (silent or not), but this seems consistent with that.

It could be extended to other activities (e.g., `Like`), but we don't have the underlying mechanisms for that and I'm not sure it's worth it.

PR #9095, opened 2018-10-25
Fix some fonts were not loaded due to CSP

gomasy:
screenshot from 2018-10-25 21-45-56


Gargron:
What kind of font is that though?.. Is it loaded by rellax.js?

Gomasy:
No, pure Mastodon UI seems to have this error.

Gargron:
It does not happen on mastodon.social which is vanilla 2.6.0rc1

Do you have a browser extension that inserts those fonts into the page, maybe?

Gomasy:
I am sorry.
It was a problem with the extended function, sorry for taking your time.

Issue #9088, opened 2018-10-25
Problem with profile setting in some languages when the display name is 28 characters


mayaeh:
### Actual behaviour

When Display name is 28 characters, if you change the Interface language to Čeština in the Preferences page, the Edit profile can't be open.

This also happens in the Cymraeg and Slovenščina.

The error log at this time is as follows:

Čeština: https://gist.github.com/mayaeh/3a5d7aeecc01e7a5f7710144a2372abe

Cymraeg: https://gist.github.com/mayaeh/7fca9a1a0070a9c2984b13bd3338ab0c

Also, when setting to Български, it seems that the character count in profile setting seems does not move.

This also happens in Hrvatski and Ido.

### Steps to reproduce the problem

Change display name to 28 characters at Settings -> Edit profile.

You access to Settings -> Preferences and change the Interface language to Čeština or Cymraeg or Slovenščina, press SAVE CHANGES.

Then, you access to Settings -> Edit profile, you get 500 error page from the server.
( To fix it please press the back button of the browser and return the Interface language to the original. )

### Specifications

equivalent to master 13e049d7729425ade92a6800357bfd6224a722b8
( https://github.com/mayaeh/mastodon/tree/7d1c9edd3d25a6a51d7f1b6d6390468edc9daaf3 )

Gargron:
Why was this not detected by `i18n-tasks missing-plural-keys`? :thinking:

PR #9100, opened 2018-10-25
Remove character counter from edit profile

Gargron:
HTML in translation is bad enough given that many translators on Weblate cannot code, but here it's also pluralized, creating issues like #9088. Instead, why not simply limit the input with `maxlength`? It's less upfront, but I also think that editing display name / bio comes up less often.


ykzts (approved):

nightpool:
hmm. I think this was a bad idea. we need to fix plurals on weblate, and now we have a check for them. there's no reason to think that avoiding plurals alltogether will help

PR #9116, opened 2018-10-26
Fix ActivityPub::Activity::Create

mayaeh:
Fix regression from #9093 .

Fixed an issue that cannot be delivered from the misskey.


mayaeh:
Sorry,  this is the same as #9114 .

PR #9123, opened 2018-10-27
When searching for an emoji with multiple separators, consider the full input

ClearlyClaire:
e.g., typing “blob_cat_p” used to search for “blob” and “cat”, but not
“blob_cat_p”, which means “blob_cat_patpat” is very unlikely to show up,
although it is likely what the user wanted to type in the first place.


ThibG:
This is not the right fix, as it computes the intersection of those values, not the union.

PR #9129, opened 2018-10-28
Ensure replied-to is a status not a boost

### LINKED ISSUE #8794: You can reply to boosts
see https://mastodon.social/users/earthlover/statuses/100792879813955373

* * * *

valerauko:
## Issue 
Fixes #8794

## Outline
If the given in-reply-to status is a boost, use the ID of the original status for replying.

## To do
- [x] tests for this specific case

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -434,6 +434,9 @@ class Status < ApplicationRecord
   end
 
   def set_conversation
+    # clients might send the ID of the boost itself instead of the boosted status
+    self.thread = thread.reblog if thread&.reblog?
+
     self.reply = !(in_reply_to_id.nil? && thread.nil?) unless reply
 
     if reply? && !thread.nil?
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -434,7 +434,6 @@ class Status < ApplicationRecord
   end
 
   def set_conversation
-    # clients might send the ID of the boost itself instead of the boosted status
     self.thread = thread.reblog if thread&.reblog?
 
     self.reply = !(in_reply_to_id.nil? && thread.nil?) unless reply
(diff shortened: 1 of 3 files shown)


puckipedia:
it's not the same thing. I can still send a s2s `Create` of a `Note` that is `inReplyTo`  a boost. I think it's better in the model, because it fixes it in all cases, not just this single edge case.

valerauko:
That's my lack of understanding then. Gonna adjust.

valerauko:
Moved to model.

Gargron:
@nightpool @ThibG Thoughts? I'm not sure replying to boosts is a bug, we certainly aren't taking advantage of it in the UI, but I'm not sure if we never will.

nightpool:
hmm. I'm worried that by merging this, we'd be encouraging client developers to be loose about boost/original IDs in other places as well.

nightpool:
however, since we already do this for boosting posts, maybe there's an argument that we should do it for replies for symmetry

valerauko:
I don't have any strong feelings about this, but I'd like to make some points:

* validating user input doesn't mean encouraging invalid input
* the current behaviour results in unexpected UX on the front
* it's literally a one-line change so if you ever want to implement separate conversation threads for boosts it's no effort to revert

ThibG:
I'm fine with this. We aren't storing AP objects anyway, it makes sense to change this in the model for our purposes.

app/models/status.rb
```diff
@@ -434,6 +434,9 @@ def set_visibility
   end
 
   def set_conversation
+    # clients might send the ID of the boost itself instead of the boosted status
```
Gargron: I think this comment is unnecessary

Gargron (approved):

PR #9132, opened 2018-10-28
Fix FetchAtomService  content type handling

### LINKED ISSUE #8773: FetchAtomService does not properly support AS2 mime type
From quick researching turns out: the HTTP gem actually strips anything after the semicolon from the mime type. This causes things to break horribly on newer Kroeg, as it doesn't recognize "application/ld+json" as appropriate AS2 json, since it wants a profile.

* * * *

valerauko:
## Issue
Fixes #8773 

## Changes
* add profile to json+ld in the Accept header as required by the spec
* use `headers['Content-type']` instead of `mime_type`
  The latter strips out the `profile` part of the content type header, so compliant ActivityPub services don't get handled properly. `headers['Content-type']` has the "raw" content type with the profile, so use that for checking.

### DIFF
--- a/app/services/fetch_atom_service.rb
+++ b/app/services/fetch_atom_service.rb
@@ -37,9 +37,11 @@ class FetchAtomService < BaseService
   def process_response(response, terminal = false)
     return nil if response.code != 200
 
-    if response.mime_type == 'application/atom+xml'
+    response_type = response.headers['Content-type']
+
+    if response_type == 'application/atom+xml'
       [@url, { prefetched_body: response.body_with_limit }, :ostatus]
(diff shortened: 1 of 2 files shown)


Gargron (approved):

Gargron:
This broke remote resource resolving because the "raw" header contains stuff like `; charset=utf-8` so of course none of the comparisons match. Also, application/ld+json is a superset of the one with the profile, so I do not follow why it would matter. If I pass an Accept header without the profile part, the responding server is still free to include the profile part in the response, and I am free to ignore it. There aren't many purely ld+json things out there in the wild for that check to be important.

I missed this bug during review and I am very regretful about it. I need to revert this patch.

valerauko:
Sorry about the trouble caused.

PR #9143, opened 2018-10-29
Add healthcheck commands to docker-compose.yml

BenLubar:
This will allow Docker to be automatically check the health of services.

Docker won't do anything other than showing the state in the output of `docker-compose ps` by default, but some management tools may watch for container health events.

Here's what my local instance looks like right now:

```
          Name                        Command                  State                     Ports
-------------------------------------------------------------------------------------------------------------
mastodon_db_1              docker-entrypoint.sh postgres    Up (healthy)
mastodon_es_1              /usr/local/bin/docker-entr ...   Up (healthy)
mastodon_redis_1           docker-entrypoint.sh redis ...   Up (healthy)
mastodon_redis_cache_1     docker-entrypoint.sh redis ...   Up (healthy)
mastodon_sidekiq_1         /sbin/tini -- bundle exec  ...   Up             3000/tcp, 4000/tcp
mastodon_streaming_1       /sbin/tini -- yarn start         Up (healthy)   3000/tcp, 127.0.0.1:4000->4000/tcp
mastodon_web_1             /sbin/tini -- bash -c rm - ...   Up (healthy)   127.0.0.1:3000->3000/tcp, 4000/tcp
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,8 @@ services:
     image: postgres:9.6-alpine
     networks:
       - internal_network
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
     volumes:
       - ./postgres:/var/lib/postgresql/data
 
@@ -14,6 +16,8 @@ services:
     image: redis:4.0-alpine
     networks:
       - internal_network
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
     volumes:
       - ./redis:/data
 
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -6,6 +6,8 @@ services:
     image: postgres:9.6-alpine
     networks:
       - internal_network
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
```
nightpool: why should we specify these explicitly? shouldn't these be part of the postgres docker image?

nightpool:
why was sidekiq excluded?

Gargron (approved):

Gargron:
@nightpool I don't know how you could check Sidekiq health from a script. It's not like it accepts connections...

PR #9144, opened 2018-10-29
Fix push notifications including HTML-encoded apostrophes.

### LINKED ISSUE #8248: Push notifications and HTML encoded entities
I received a push notification saying:

> That`&apos;`s so cool

The original text says

> That`'`s so cool

The push notification came to Firefox for Android 61.0.2 running on Oreo.

### Screenshot
f2a80323ee78b29c

### Toot which caused the notification
https://mastodon.social/@naomialderman/100565975103807068

* * * *

BenLubar:
The notifications already do not have encoded <, >, ", or & characters.

Fixes #8248.

### DIFF
--- a/app/javascript/mastodon/service_worker/web_push_notifications.js
+++ b/app/javascript/mastodon/service_worker/web_push_notifications.js
@@ -73,7 +73,7 @@ const formatMessage = (messageId, locale, values = {}) =>
   (new IntlMessageFormat(locales[locale][messageId], locale)).format(values);
 
 const htmlToPlainText = html =>
-  unescape(html.replace(/<br\s*\/?>/g, '\n').replace(/<\/p><p>/g, '\n\n').replace(/<[^>]*>/g, ''));
+  unescape(html.replace(/<br\s*\/?>/g, '\n').replace(/<\/p><p>/g, '\n\n').replace(/<[^>]*>/g, '').replace(/&apos;/g, '\''));
 
 const handlePush = (event) => {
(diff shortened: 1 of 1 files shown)


nightpool:
why are we encoding `&apos;` when we're not encoding other types of characters? this sounds like it's a fix that should be made server-side

SuperSandro2000:
Can we still merge this? I find this really annoying.

nightpool:
Actually, wouldn't unescape already handle this?

SuperSandro2000:
Clearly it doesn't cause otherwise I wouldn't get the wrong notifications.

fwenzel:
https://lodash.com/docs/4.17.15#unescape

Gargron:
> Note: No other HTML entities are unescaped. To unescape additional HTML entities use a third-party library like he.

Okay, then we should do that I guess

fwenzel:
This also exists: https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/utils/html.js , looks like we can consolidate solutions.

There's also a DOMParser API (cf. https://stackoverflow.com/a/34064434) which we could just use if present and fall back to cruddy incomplete entity replacement on the off chance the browser is old.

fwenzel:
Oh, good catch, clearly I did not pay attention.

SuperSandro2000:
Bump

SuperSandro2000:
bump

fwenzel:
fairly certain this is unfixed

PR #9147, opened 2018-10-29
Do not hide boost notifications from followed people with hidden boosts

ClearlyClaire:
Not displaying boosts from a followed user in the Home timeline and not
having notifications when they reblog your own content are two very
separate concerns, tying them together seem counter-intuitive and unwanted.


Gargron:
Yeah I just looked and that behaviour was not mentioned in the original PR either.

Gargron (approved):

PR #9150, opened 2018-10-30
Accept the same payload in multiple inboxes and deliver

Gargron:


app/lib/activitypub/activity/create.rb
```diff
@@ -99,6 +104,19 @@ def process_audience
     @params[:visibility] = :limited
   end
 
+  def postprocess_audience_and_deliver
+    return if @status.mentions.find_by(account_id: @options[:delivered_to_account_id])
+
+    delivered_to_account = Account.find(@options[:delivered_to_account_id])
+
+    @status.mentions.create(account: delivered_to_account, silent: true)
+    @status.update(visibility: :limited) if @status.direct_visibility?
```
ThibG: I'm not completely sure what the “limited” visibility in Mastodon encompasses, but apart from that, looks good to me.

ClearlyClaire (approved):

PR #9188, opened 2018-11-01
Support Page type in search input

mbajur:
This one is related to #9121 

***

Right now, search input only resolves URLs that are one of `Note`, `Article`, `Image` or `Video` type. While Mastodon now supports a `Page` object type, it would be great to add such support to the search input as well.

ps. sorry for not including this in #9121


arjenpdevries (commented):
Looks right to me :-)

ida (approved):

mkljczk (approved):

Gargron (approved):

PR #9189, opened 2018-11-01
HTTP Signatures: relaxed Date verification

vitalyster:
`Thu, 1 Nov 2018 09:02:46 GMT` is not accepted by `Time.httpdate` as `not RFC 2616 compliant date`, possible because it require leading zeros in day part. While RFC2616 compliance is questionable, `Time.parse` get correct results from it.

### DIFF
--- a/app/controllers/concerns/signature_verification.rb
+++ b/app/controllers/concerns/signature_verification.rb
@@ -95,7 +95,8 @@ module SignatureVerification
 
   def matches_time_window?
     begin
-      time_sent = Time.httpdate(request.headers['Date'])
+      date_header = request.headers['Date']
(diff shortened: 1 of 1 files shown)


vitalyster:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html#RFC_1123_DATE_TIME - the source of problem

o01eg:
RFC2616 requires fixed-length date format so it allows only 2DIGIT for day subset of RFC1123: https://tools.ietf.org/html/rfc2616#section-3.3.1

Gargron (changes requested):

PR #9200, opened 2018-11-04
Temporarily hold timeline if mouse moved recently (fixes #8630)

### LINKED ISSUE #8630: Stop the timeline from moving out from under my mouse when I'm about to click
Similar to one of my biggest peeves with TweetDeck - when I try to click something in a busy timeline in Mastodon, it often slides out from under my pointer and I end up interacting with something else.

It would be really nice if the timeline scrolling could pause for a significant period (say 1000ms) as I'm trying to interact with it.

This is basically a re-submission of both #1004 and #7278. 

Looks like #1004 was fixed in 4df831078f3af47588ea66e8b48d4ec26f4872eb, but then reverted in 7e0aed398f7e3e74a3c15e426082f88cf8b79bcf

Then #7278 was fixed in 58852695c8ec490239ed3812f82971f8c1e6c172 but then reverted in e06fbc4fcfbaaa2d420cbbace37d7c228a8bfc5b

I wouldn't mind finally getting around to scratching my own itch and submit a fix. But it might help to have some details as to why two different whacks at fixing it were reverted so I don't make the same mistakes.

* * * *

lmorchard:
- On recent mouse movement, hold timeline position so statuses remain in
  place for interactions in progress.

- If the timeline had been scrolled to the top before mouse movement,
  restore scroll on mouse idle.

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -9,6 +9,8 @@ import { List as ImmutableList } from 'immutable';
 import classNames from 'classnames';
 import { attachFullscreenListener, detachFullscreenListener, isFullscreen } from '../features/ui/util/fullscreen';
 
+const MOUSE_IDLE_DELAY = 300;
+
 export default class ScrollableList extends PureComponent {
 
   static contextTypes = {
@@ -37,6 +39,8 @@ export default class ScrollableList extends PureComponent {
 
   state = {
     fullscreen: null,
+    mouseMovedRecently: false,
+    scrollToTopOnMouseIdle: false,
   };
 
   intersectionObserverWrapper = new IntersectionObserverWrapper();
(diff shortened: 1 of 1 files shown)


lmorchard:
> You should know I have implemented and then reverted a similar patch.

Yeah, I figured I'd give it one more shot, because mis-clicking on the wrong status *really* bothers me. If this attempt gets rejected, I'll understand. 😅 

For what it's worth, I used a timer because the `_recentlyMoved` logic wouldn't be checked until the next mouse move or other column state update (e.g. new statuses).

I think a timer makes this more responsive - because one of the things I've got it doing is scrolling back to the top of the column on mouse idle, so that it doesn't actually stay stuck in the past.

lmorchard:
Pushed a small tweak to better clear the timer, since I realized the property never returned to null after a clearTimeout

Gargron:
>I think a timer makes this more responsive - because one of the things I've got it doing is scrolling back to the top of the column on mouse idle, so that it doesn't actually stay stuck in the past.

I have to ask - what if I intentionally scroll down to visually mark my position on something. This would pull me back to top?

Consider an alternative to this is disabling updates on mouseDown, since the click event is fired on mouseUp, the most annoying things are when you click on something and then it moves from under your mouse when you release, right?

lmorchard:
>I have to ask - what if I intentionally scroll down to visually mark my position on something. This would pull me back to top?

No, [if you're not at `scrollTop === 0`, scroll to top on idle doesn't happen](https://github.com/tootsuite/mastodon/pull/9200/files#diff-895b4ae1a3e9388ceff7e589a182a348R93)

> Consider an alternative to this is disabling updates on mouseDown, since the click event is fired on mouseUp, the most annoying things are when you click on something and then it moves from under your mouse when you release, right?

I don't think pause-on-mouseDown would help. For me, the issue feels most often like the timeline has moved in the span of approaching, landing on, and then clicking on a UI element. The mouseDown / mouseUp span is the shortest part of all that.

Gargron:
Okay, so there's kind of only one way to find out. I will merge this, it will get some use from people running master, and then we'll know if it's a good change or if the behaviour will be disturbing. Fingers crossed but if it'll turn out worse I'll revert it, just a heads up :pray:

Gargron (approved):

lmorchard:
> Fingers crossed but if it'll turn out worse I'll revert it, just a heads up 🙏

Yay, thanks for the consideration!

Gargron:
@lmorchard Could you please investigate #9217 as I am now getting complaints from people on servers running master. It seems that this patch interferes with manual scrolling. If it's the result of a minor bug, it would be great to find it, otherwise I am preparing to revert this.

lmorchard:
Sure, I will take a look in a few hours. I think I see what's going on

Gargron:
As I said I am prepared to revert, but I can wait until you investigate.

PR #9201, opened 2018-11-04
readme writing changes

kyotee:
I am a computer science student and Ruby on Rails enthusiast. I have come across your repo and I am definitely going to try out mastodon in the mere future. After reading the ReadMe file, I have added small writing changes in the introduction to improve readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@
 [weblate]: https://weblate.joinmastodon.org/engage/mastodon/
 [docker]: https://hub.docker.com/r/tootsuite/mastodon/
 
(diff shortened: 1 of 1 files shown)


nightpool:
thank you for your enthusiasm! however, I'm not sure I agree with all of these changes. particularly, following and publishing are in separate sentences for clarity and emphasis. Combining them makes the sentence nicer as prose, but worse for the casual user who can get confused by compound sentence structures while skimming.

 "click the gif below to learn more in a video" just kinda sounds weird, even if it's a more accurate description of what's happening. were you confused by the original language, or is this just a preemptive change? if it was the former, then we should probably consider restructuring the sentence, but if it was the latter then I don't think there's a problem

however, I'm just an engineer as well, so I'm not very experienced in this area. I'm going to forward this pull to a writer I know who has experience in marketing copy and let him have the final say.

Gargron:
Hey! Thank you for the contribution but I prefer the current text.

PR #9205, opened 2018-11-04
Add configuration option to filter replies in lists

ClearlyClaire:
There are three possible settings:
- Don't show replies (same behavior as Mastodon 2.6)
- Show all replies to followed users (same behavior as Mastodon < 2.6)
- Show only replies to users in the list (default)

Changing is done by ~~clicking on the link, which cycles between the settings (yeah, that's not great UI)~~ a set of radio buttons.

~~Note that only “show all replies” shows mentions to yourself. That could probably be changed.~~
All settings show replies to yourself.

### DIFF
--- a/app/javascript/mastodon/features/list_timeline/index.js
+++ b/app/javascript/mastodon/features/list_timeline/index.js
@@ -19,9 +19,9 @@ import Icon from 'mastodon/components/icon';
 const messages = defineMessages({
   deleteMessage: { id: 'confirmations.delete_list.message', defaultMessage: 'Are you sure you want to permanently delete this list?' },
   deleteConfirm: { id: 'confirmations.delete_list.confirm', defaultMessage: 'Delete' },
-  all_replies:   { id: 'lists.replies_policy.all_replies', defaultMessage: 'to any followed user' },
-  no_replies:    { id: 'lists.replies_policy.no_replies', defaultMessage: 'none' },
-  list_replies:  { id: 'lists.replies_policy.list_replies', defaultMessage: 'only to list' },
+  all_replies:   { id: 'lists.replies_policy.all_replies', defaultMessage: 'any followed user' },
+  no_replies:    { id: 'lists.replies_policy.no_replies', defaultMessage: 'no one' },
+  list_replies:  { id: 'lists.replies_policy.list_replies', defaultMessage: 'members of the list' },
 });
 
 const mapStateToProps = (state, props) => ({
@@ -134,15 +134,14 @@ class ListTimeline extends React.PureComponent {
     }));
   }
 
-  handleRepliesPolicyClick = () => {
-    const { dispatch, list } = this.props;
+  handleRepliesPolicyChange = ({ target }) => {
+    const { dispatch } = this.props;
     const { id } = this.props.params;
-    const replies_policy = { 'all_replies': 'no_replies', 'no_replies': 'list_replies', 'list_replies': 'all_replies' }[list.get('replies_policy')];
-    dispatch(updateList(id, undefined, false, replies_policy));
+    dispatch(updateList(id, undefined, false, target.value));
   }
 
   render () {
-    const { shouldUpdateScroll, hasUnread, columnId, multiColumn, list, intl } = this.props;
+    const { shouldUpdateScroll, hasUnread, columnId, multiColumn, list } = this.props;
     const { id } = this.props.params;
     const pinned = !!columnId;
     const title  = list ? list.get('title') : id;
@@ -187,9 +186,21 @@ class ListTimeline extends React.PureComponent {
             </button>
 
             { replies_policy !== undefined && (
-              <button className='text-btn column-header__setting-btn' tabIndex='0' onClick={this.handleRepliesPolicyClick}>
-                <i className='fa fa-reply' /> <FormattedMessage id='lists.replies_policy.title' defaultMessage='Show replies: {policy}' values={{ policy: intl.formatMessage(messages[replies_policy]) }} />
-              </button>
+              <div>
+                <div className='column-settings__row'>
+                  <fieldset>
+                    <legend><FormattedMessage id='lists.replies_policy.title' defaultMessage='Show replies to:' /></legend>
+                    { ['no_replies', 'list_replies', 'all_replies'].map(policy => (
+                      <div className='setting-radio'>
+                        <input className='setting-radio__input' id={['setting', 'radio', id, policy].join('-')} type='radio' value={policy} checked={replies_policy === policy} onChange={this.handleRepliesPolicyChange} />
+                        <label className='setting-radio__label' htmlFor={['setting', 'radio', id, policy].join('-')}>
+                          <FormattedMessage {...messages[policy]} />
+                        </label>
+                      </div>
+                    ))}
+                  </fieldset>
+                </div>
+              </div>
             )}
           </div>
         </ColumnHeader>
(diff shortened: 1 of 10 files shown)


ThibG:
I kept messing up the condition :weary:
But I'm confident it is correct now, and I added specs covering most cases.

UI-wise, it would probably be better to have a dropdown, but the only dropdown of this kind is the privacy dropdown, and it's not easily reusable.

ThibG:
Would anyone actually make use of the “show only replies to users in the list” option? This seems to cover most—if not all—needs, and it would make the code and UI simpler if it were the only option.

cbayerlein:
I think it'sfine as it is now, although it's always good to provide options to the users.

trwnh:
I would use those 3 and also a "show all replies" option, ideally. It depends on the list.

nightpool:
@trwnh can you be more specific?

nightpool:
regardless, I absolutely agree with gargron that we shouldn't be doing this
filtering server side

ThibG:
> regardless, I absolutely agree with gargron that we shouldn't be doing this filtering server side

Why? (Also, uh, it would be nice if said opinion of gargron was actually in the issue?)

BaptisteGelez:
I would personnaly like to use both the "Show only replies to users in the list" and "Show all replies to followed users" options.

ThibG:
Got some feedback from @nightpool on discord (I'll paraphrase them, and quotes will be from them):
- Doing this server-side would be confusing, “because users intuitively expect most filtering changes to be retroactive”
- They do not see value in having multiple choices and believe the correct behavior would be to only show replies to members of the list.

I personally think there is value in having the choices I described, and I think there is value in doing the filtering server-side (having the same behavior across clients without needing special support, avoiding useless data traffic), but I also understand the point about the filter not applying retroactively being confusing.

ThibG:
About the filtering not being retroactive, the same can be said about list creation itself: only toots received after the list has been created will appear there. If you think changing filtering after creation is too confusing, it could be possible to choose the type of filtering upfront and not providing any UI to change it (although I would find it problematic).

Of course, we could also not filter at all in the backend and do it in the front-end instead, but this may lead to inconsistent behavior across clients if they are not updated to deal with a new API, and it will also cause increased traffic, as unneeded toots may be streamed for every list.

I'm open to implementing client-side filtering, although I prefer server-side filtering for the above reasons (and also because I've already done the work, tbh).

benjaminbillet:
I don't really understand why this is not merged yet. The new list system makes no sense, what is the point of using lists if you also have to scroll your home to see if you don't miss anything? :s

ThibG:
This has been merged into glitch-soc, together with a change to the UI that I just ported.
This looks like this:

screenshot_2018-11-30 dev instance

ThibG:
@trwnh I will consider this, but as this would show *more* thing than the Home timeline we currently have, this seems a bit counter-intuitive. Also, this is slightly more involved as toots delivered to lists are pre-filtered with the same criteria as for the Home timeline.

trwnh:
@ThibG I guess that's true. The reason I'm interested in it is because I miss the functionality from Tweetdeck circa 2011 where you could show all replies in both home timelines and in lists. That might be a separate request, though... I should probably look into a separate issue for that.

kit-ty-kate:
@trwnh i think this can be added later, I would also like to have this option for both the home timeline and lists

ThibG:
I've personally been using a few lists, one of which is set to “any followed user” while the others are using the default “members of the list” option. It is extremely useful to me, and I've seen multiple people ask for something like this.

Can this PR be reconsidered? Would changing it to do the filtering client-side help?

kit-ty-kate:
Can this be reopened?

wryk:
I think this PR is still relevant. Without replies in lists users need to scroll the home timeline (miss the point of using lists) or peek at profile pages (not viable with large lists)

ThibG:
Reopening, but I doubt Gargron changed his mind on this :/

ykzts (approved):

shleeable:
RIP.

shleeable:
RIP.

SuperSandro2000:
Bump

db/migrate/20181127165847_add_show_replies_to_lists.rb
```diff
@@ -0,0 +1,23 @@
...
+class AddShowRepliesToLists < ActiveRecord::Migration[5.2]
+  include Mastodon::MigrationHelpers
+
+  disable_ddl_transaction!
+
+  def up
+    safety_assured do
+      add_column_with_default(
+        :lists,
+        :replies_policy,
```
Gargron, line 12: This is the first enum in our code we call `_policy`, I don't want to bikeshed too much but is this a fully appropriate name?

Gargron (approved):
LGTM but wondering about API naming

app/javascript/mastodon/features/list_timeline/index.js
```diff
@@ -10,14 +10,17 @@ import { addColumn, removeColumn, moveColumn } from '../../actions/columns';
...
-import { fetchList, deleteList } from '../../actions/lists';
+import { fetchList, deleteList, updateList } from '../../actions/lists';
 import { openModal } from '../../actions/modal';
 import MissingIndicator from '../../components/missing_indicator';
 import LoadingIndicator from '../../components/loading_indicator';
 
 const messages = defineMessages({
   deleteMessage: { id: 'confirmations.delete_list.message', defaultMessage: 'Are you sure you want to permanently delete this list?' },
   deleteConfirm: { id: 'confirmations.delete_list.confirm', defaultMessage: 'Delete' },
+  all_replies:   { id: 'lists.replies_policy.all_replies', defaultMessage: 'any followed user' },
```
Gargron, line 21: Actually, these should start with a capital letter, I believe.

app/javascript/mastodon/features/list_timeline/index.js
```diff
@@ -156,6 +166,24 @@ class ListTimeline extends React.PureComponent {
...
             </button>
+
+            { replies_policy !== undefined && (
+              <div>
+                <div className='column-settings__row'>
+                  <fieldset>
+                    <legend><FormattedMessage id='lists.replies_policy.title' defaultMessage='Show replies to:' /></legend>
+                    { ['no_replies', 'list_replies', 'all_replies'].map(policy => (
+                      <div className='setting-radio'>
+                        <input className='setting-radio__input' id={['setting', 'radio', id, policy].join('-')} type='radio' value={policy} checked={replies_policy === policy} onChange={this.handleRepliesPolicyChange} />
```
Gargron, line 177: Plain radio buttons look rather unsightly in there, can we use the toggles we use in other column settings?

Gargron (changes requested):
Actually upon seeing the screenshot down thread I have a few more concerns

app/javascript/mastodon/features/list_timeline/index.js
ThibG: Someone pointed out that we already have radio buttons somewhere else in the interface (directory settings), I'll have a look to reuse that styling.

db/migrate/20181127165847_add_show_replies_to_lists.rb
ThibG: This has been in glitch-soc for almost two years, so I'd need to check if any client uses that API, but otherwise I'm fine with changing the name if you have a better suggestion.

ThibG:
Updated the radio button design to look like the directory column design:
image

Gargron (approved):

PR #9210, opened 2018-11-05
Fix property name for styled scrollbars in Firefox

sorin-davidoi:
https://webplatform.news/issues/2018-10-26

### DIFF
--- a/app/javascript/styles/mastodon/reset.scss
+++ b/app/javascript/styles/mastodon/reset.scss
@@ -54,8 +54,7 @@ table {
 }
 
 html {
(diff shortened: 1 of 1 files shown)


Gargron (approved):

nolanlawson:
Out of curiosity, why change the track color from `rgba($base-overlay-background, 0.1)` to `transparent`?

sorin-davidoi:
I don't have Mastodon set up anymore and submitted this change after tweaking the CSS in the browser - `rgba($base-overlay-background, 0.1)` might be a better solution.

PR #9220, opened 2018-11-06
Optimize the process of following someone

Gargron:
Summary:

- Eliminate unnecessary "SELECT ... FROM accounts" from FollowService
- Create local follow (request) notifications asynchronously
- Do not re-resolve ActivityPub accounts from FollowService
- Optimistically update UI to show followed/unfollowed state before request completes


ThibG:
Not too sure about all the optimism going on in the front-end, but the backend changes LGTM.

PR #9227, opened 2018-11-07
Increase default column width from 330px to 350px

Gargron:


ashleyhull-versent:
been running on 360px for a while.. I approve the width increase

mayaeh (approved):

Issue #7699, opened 2018-06-01
Add thread marker


renatolond:
As a user, I would like to visual indicator that there's more replies from the toot author, specially when boosting a thread since I can't quote the toot indicating it's a thread.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #9228, opened 2018-11-07
Add "Show thread" link to self-replies

### LINKED ISSUE #4716: add an hint that a message is part of a thread
There is no hint that either a message is part of a thread reply or not. It lead to several disagreement and a real confusion about the meaning of a reply in the last case. the confusion eventually was solved. 

So my idea would be to add some kind of hint that a message is a reply. Actually an even better UX would be to show a part of the thread as twitter does, but I'm quite sure that this last option is a giant issue, so just an hint would be already great.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate. (as far as I could)
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this). (not in mastodon.social at least)

Gargron:
Removes the "reply-all" icon variant, as basically nobody noticed the difference. When a toot contains mentions it's pretty obvious that it's part of a reply chain. It is only not obvious on self-reply chains, therefore, on self-reply chains, show an obvious "show thread" link in the style of "read more"

Fix #4716

image

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -67,6 +67,7 @@ class Status extends ImmutablePureComponent {
     unread: PropTypes.bool,
     onMoveUp: PropTypes.func,
     onMoveDown: PropTypes.func,
+    showThread: PropTypes.bool,
   };
 
   // Avoid checking props that are functions (and whose equality will always
@@ -168,7 +169,7 @@ class Status extends ImmutablePureComponent {
     let media = null;
     let statusAvatar, prepend, rebloggedByText;
 
-    const { intl, hidden, featured, otherAccounts, unread } = this.props;
+    const { intl, hidden, featured, otherAccounts, unread, showThread } = this.props;
 
     let { status, account, ...other } = this.props;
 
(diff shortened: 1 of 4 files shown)


renatolond:
Fixed #7699 ?

nightpool:
mmm. feels pretty surprising/nonobvious/confusing that we show this for self replies but not other types of replies.....

renatolond:
For other replies you have other visual indicators, like the @ of the user the answer is in reply to (going up), or the number of replies (going down)

jeroenpraat:
I see that this in the code is localized, but it wasn't synced with Weblate. Therefore untranslated.

renatolond:
If you reply to a thread without mentioning a user (in Mastodon, not the case for other fediverse softwares) if I'm not mistaken the user is not notified, so the behavior is slightly different from a common reply, though I agree it lacks any indication of having something before.

Coming from Twitter, I find the idea of having "show thread" for things that are not replies to one own's posts weird. I would prefer to follow a similar way to Twitter in this sense (of showing "in reply to xxx").
I don't know how it feels for new users that don't come from Twitter, but to people coming from Twitter this is regular behavior, threads are only considered (and signaled) as threads if they are replies to your own statuses.

Issue #9223, opened 2018-11-07
"422Validation failed: You cannot follow more than 2280.3 people"


Angular-Angel:
### Expected behaviour

I try and follow someone, I get to follow them.

### Actual behaviour

I get a wierd error message with what looks like a floating point error.

### Steps to reproduce the problem

Try and follow someone on anticapitalist.party with an account that already follows a load of people.

### Specifications

Account in question: https://anticapitalist.party/@Angle

nightpool:
the error is correct, although we should probably round it for avoidance of
confusion.....

nightpool:
for more information see https://github.com/tootsuite/mastodon/pull/8807

PR #9230, opened 2018-11-07
Fix follow limit validator reporting lower number past threshold

Gargron:


nightpool:
also `.round` per #9223

PR #9233, opened 2018-11-08
Cancel list scroll reset after mouse move on wheel scroll

### LINKED ISSUE #9217: "Scratch" may occur when scrolling the timeline
### Actual behaviour

After #9200 , "scratch" may occur when scrolling the timeline.

In addition, an error occurs because the account timeline, favorite timeline, etc have not read status_list_container.

The hold timeline function is a useful feature when I "favorite" or "boost", but I think that the problem is great at present.

Reference video is [here](https://www.dropbox.com/s/3fu1cvb9hiyq91b/2018-11-06_mastodon_scroll_error.mp4?dl=0).

Also, in Chrome for Windows, the scrolled timeline from the top may unexpectedly return to the top.

It sometimes happens if you try to scroll the timeline from the top.

A JavaScript error occurs when you click the account name and display the account timeline and move the mouse.

This also happens in such as the favorite timeline, DM timeline and block list, that don’t read the status_list_container.

### Specifications

mastodon master 535ce848962c5040c11d3dd6dc4944e9501ff6f5

Firefox for mac 63.0.1, Chrome for mac 70.0.3538.77

Chrome for Windows 70.0.3538.77

lmorchard:
Also fixes an error introduced in #9200 caused by attempting to use an
undefined prop.

Fixes #9217

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -39,8 +39,6 @@ export default class ScrollableList extends PureComponent {
 
   state = {
     fullscreen: null,
-    mouseMovedRecently: false,
-    scrollToTopOnMouseIdle: false,
   };
 
   intersectionObserverWrapper = new IntersectionObserverWrapper();
@@ -65,6 +63,8 @@ export default class ScrollableList extends PureComponent {
   });
 
   mouseIdleTimer = null;
+  mouseMovedRecently = false;
+  scrollToTopOnMouseIdle = false;
 
   clearMouseIdleTimer = () => {
     if (this.mouseIdleTimer === null) {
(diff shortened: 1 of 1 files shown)


lmorchard:
I *think* this should fix #9217 - looked like the main issue was that I missed wheel scrolling as a source of intentional scrolling.

mayaeh:
In my environment, I seems that scrolling the timeline with the Firefox for mac (63.0.1) is heavier than before merging this PR's code.

I seems that there is no problem with the Safari for mac (12.0.1) and the Chrome for mac (70.0.3538.77).

As for Windows, I can not test it until I return from work... sorry.

I will try it in another environment after I return from work.

Gargron:
I wonder if these things need to be in state if they are never rendered. Using donor variables might work better and improve performance.

mayaeh:
I tried this on the Chrome for Windows (70.0.3538.77) and the Firefox for Windows (63.0.1).
I think that it was fixed within the range that I care about an issue.

Gargron:
What's the extent of this issue if you are manually scrolling with page up/down, the scrollbar, or arrow keys? (Because the fix seems to depend on the wheel event specifically)

lmorchard:
And one more tweak, as I just realized my call to `this.props.scrollToTop` was redundant because the scroll event gets fired by setting `node.scrollTop` and `handleScroll` takes care of it anyway.

Gargron (approved):

PR #9248, opened 2018-11-09
Add setting to not aggregate reblogs

### LINKED ISSUE #9222: Add an option to opt-out from boosts aggregation
### Pitch

The idea is to have an option to see all individual boosts from your followers and not just the first boost to appear on your timeline on your current session. This option would not be activated by default of course.

### Motivation

Some people prefer to know who boosted a toot and which toots are more important than others (i.e. with this feature such a toot would appear multiple times on your timeline) even if this means getting a lot more activity on your timeline and getting redundant posts.

ClearlyClaire:
Fixes #9222

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -27,7 +27,7 @@ class FeedManager
   end
 
   def push_to_home(account, status)
-    return false unless add_to_feed(:home, account.id, status, account.user.aggregates_reblogs?)
+    return false unless add_to_feed(:home, account.id, status, account.user&.aggregates_reblogs?)
     trim(:home, account.id)
     PushUpdateWorker.perform_async(account.id, status.id, "timeline:#{account.id}") if push_update_required?("timeline:#{account.id}")
     true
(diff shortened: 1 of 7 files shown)


app/lib/feed_manager.rb
```diff
@@ -226,11 +226,11 @@ def phrase_filtered?(status, receiver_id, context)
   # added, and false if it was not added to the feed. Note that this is
   # an internal helper: callers must call trim or push updates if
   # either action is appropriate.
-  def add_to_feed(timeline_type, account_id, status)
+  def add_to_feed(timeline_type, account_id, status, aggregate_reblogs = true)
     timeline_key = key(timeline_type, account_id)
     reblog_key   = key(timeline_type, account_id, 'reblogs')
 
-    if status.reblog?
+    if status.reblog? && (aggregate_reblogs.nil? || aggregate_reblogs)
```
ThibG: This is a bit awkward, but basically none of the tests pass an actual local account (with an associated `User`) to those functions, so…

kit-ty-kate:
@Gargron any opinion on this?

kit-ty-kate:
FYI: I'm using this PR on top of `2.6.2` on my instance [lesbiab.space](https://lesbiab.space) and it works perfectly so far

kedamaDQ:
I often see the actions which are boost and mention. But boosted toot is not visible in my home timeline and only mentioning is visible.
I think to this is a very good function if frequently need to open the account timeline for check to "what is mentioned?".

Otherwise,  behavior remains as it is unless actively opt-outed, I like it.

mayaeh:
I think this is useful.

In Japanese, especially in my visible range, I often see that they boosting post and then post impressions ( eg: I like this).
Then I will search for what he is talking about.

This PR can omit this procedure.

However, there are times when I do not want to see the same post many times.

Also, it may be better to show that this setting does not affect posts received before changing this setting.

Also, since the space between the hint and the next setting item is too narrow, I think that adjustment is necessary.

ThibG:
Updated the spacing and hint text

screenshot_2018-12-05 preferences - dev instance 1

mayaeh (approved):

kedamaDQ:
thank you! thank you for merging!!

PR #9253, opened 2018-11-10
Add support for SVG custom emoji

ClearlyClaire:
Add support for SVG custom emoji. SVG are always rasterized to provide an animation-less version.
In addition, unless the `ALLOW_UNSAFE_UPLOADS` environment variable is set to `true`, the “original” version is rasterized too, losing potential animations and other advantages of SVG (but accepting SVG emojis issued by other software).

In case `ALLOW_UNSAFE_UPLOADS` is set to `true` (which should only be done by and admin after they have reviewed their config to avoid potential XSS attacks), the original version of the emoji is copied as is, enjoying vector graphics and potential SVG animations.

### DIFF
--- a/.env.production.sample
+++ b/.env.production.sample
@@ -230,3 +230,8 @@ STREAMING_CLUSTER_NUM=1
 # http_proxy=http://gateway.local:8118
 # Access control for hidden service.
 # ALLOW_ACCESS_TO_HIDDEN_SERVICE=true
+
+# Allow unsafe uploads such as SVG files. Only do that if you are confident your
+# media server uses a sufficiently secure configuration (Content-Security-Policy
+# disallowing scripts execution, ideally using a separate domain name.
+# ALLOW_UNSAFE_UPLOADS=true
--- a/app/models/custom_emoji.rb
+++ b/app/models/custom_emoji.rb
@@ -29,11 +29,11 @@ class CustomEmoji < ApplicationRecord
 
   has_one :local_counterpart, -> { where(domain: nil) }, class_name: 'CustomEmoji', primary_key: :shortcode, foreign_key: :shortcode
 
-  has_attached_file :image, styles: { static: { format: 'png', convert_options: '-coalesce -strip' } }
+  has_attached_file :image, styles: { static: { format: 'png', source_file_options: '-channel rgba -background "rgba(0,0,0,0)"', convert_options: '-coalesce -strip' } }
 
   before_validation :downcase_domain
 
-  validates_attachment :image, content_type: { content_type: 'image/png' }, presence: true, size: { less_than: LIMIT }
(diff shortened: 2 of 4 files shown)


ThibG:
Marked WiP as I suspect ImageMagick will happily rasterize SVG to huge PNG images.

ThibG:
Limited rendered static file to 200×200px max. This is arbitrary and probably a bit much tbh, but custom emojis are displayed in multiple sizes (at least 20px, 24px and 25px) throughout the UI, and we may also accommodate screens with higher pixel density I guess.

So, upon receiving a custom emoji, it will:
- Be rejected if over `MAX_MATRIX_LIMIT`
- Rendered as a static png of max 200×200px for the non-animated views
- Used directly in the UI if animated emojis are enabled, used as the raster version defined above otherwise

nightpool:
in an ideal world, we should rasterize to the exact sizes we use, instead
of trusting the downscaler to "behave well"

ThibG:
The issue with this is that we would need as many variants as different sizes we use, and should we change those sizes, we would have to regenerate new variants.

nightpool:
Also, SVGs are extremely unsafe and hard to whitelist—there are tons of
possible scripting elements and similar constructions. For security, we
should probably just use the rasterized version all the time.

nightpool:
I wonder how hard it is to DOS imagemagick by sending it a malicious SVG?

ThibG:
Used in an img tag, scripting elements are not evaluated. Also, we have content security policies.

ThibG:
DOSing ImageMagick via a malicious SVG file is a valid concern, and I don't have an answer to that (although I'm not confident it handles any other type of media particularly well)

nightpool:
i would still be somewhat wary of hosting a malicious SVG file, even if we
"assume" everyone will be accessing it through an img tag.

the problem with SVG vs other types of media is that SVG has a much larger
attack surface

ThibG:
Thinking more about it, I think it is safe to use in the WebUI and public pages (as images), **but** a malicious SVG could still cause some damage if someone were to open it in their browser (right click, “View image”). If media are stored on the same host as the WebUI, this could lead to XSS.

ThibG:
So we could:
- Rasterize no matter what (safer, still open to potential ImageMagick/rsvg exploits, loose most benefits of using SVGs but remain compatible with federation)
- Store untrusted SVGs on a different domain and/or with a restrictive enough CSP. Unfortunately, that's not something we can enforce from Mastodon's code itself…

ThibG:
Managed to get PaperClip to change the extension of the stored file to png whenever it is rasterized.

What I mean by “sufficiently-secure” reverse-proxy config is a web server serving /system/custom_emojis/images with a CSP disallowing script executions, and possibly on a separate domain name. An example of such CSP would be:

```
Content-Security-Policy "default-src 'none'; style-src 'unsafe-inline'; media-src data:; img-src 'self' data:"
```

PR #9267, opened 2018-11-12
Change the password form order

mayaeh:
Change the order of password change form on Preferences -> Security page.

I think that in the many password change form, the current password -> new password is in order.

With this change, I think that we can reduce the case of failing to change the password.

Also fixes the problem where translation of "New password" is not applied to label.

Before:

After:

### DIFF
--- a/app/javascript/mastodon/locales/ja.json
+++ b/app/javascript/mastodon/locales/ja.json
@@ -321,7 +321,7 @@
   "status.show_less_all": "全て隠す",
   "status.show_more": "もっと見る",
   "status.show_more_all": "全て見る",
-  "status.show_thread": "続きを読む",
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/locales/ja.json
```diff
@@ -321,7 +321,7 @@
   "status.show_less_all": "全て隠す",
   "status.show_more": "もっと見る",
   "status.show_more_all": "全て見る",
-  "status.show_thread": "続きを読む",
+  "status.show_thread": "スレッドを表示",
```
TheKinrar: This is not related to the PR.

mayaeh: You're right.
Thanks.

Gargron (approved):

PR #9271, opened 2018-11-12
Swap avatar and header input fields in profile page

### LINKED ISSUE #9211: confusing header vs avatar upload button ordering in edit profile
In v2.6.1, the edit profile page has a confusing order for the buttons to upload the user's avatar and header.  Either the header and avatar images on the left should be swapped, or the two buttons and accompanying text on the right should be swapped (probably better to swap the buttons, as the header is above the avatar in user page view).

confusing-header-vs-avatar-button-order

ClearlyClaire:
Fixes #9211

### DIFF
--- a/app/views/settings/profiles/show.html.haml
+++ b/app/views/settings/profiles/show.html.haml
@@ -14,9 +14,10 @@
       = render 'application/card', account: @account
 
     .fields-row__column.fields-group.fields-row__column-6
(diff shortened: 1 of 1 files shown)


mayaeh:
I think this is necessary for usability improvement.

mayaeh (approved):

PR #9272, opened 2018-11-12
Fix race condition causing shallow status with only a "favourited" attribute

### LINKED ISSUE #9231: Favourite a toot while being deleted crashes the webpage
### Expected behaviour

Do nothing and still work.

### Actual behaviour

Getting a blank "grey" page.
Partial log from the console (sorry I couldn't get much more from a failed copy/past):
```
TypeError: "undefined has no properties"
TypeError: "t is null"
```

On the web application, open a toot, fav it at the same time that it gets deleted, watch the webapp crashing.

### Specifications

Mastodon version: 2.6.1
Web browser: Firefox 63.0.1

ClearlyClaire:
Fixes #9231

### DIFF
--- a/app/javascript/mastodon/reducers/statuses.js
+++ b/app/javascript/mastodon/reducers/statuses.js
@@ -38,11 +38,11 @@ export default function statuses(state = initialState, action) {
   case FAVOURITE_REQUEST:
     return state.setIn([action.status.get('id'), 'favourited'], true);
   case FAVOURITE_FAIL:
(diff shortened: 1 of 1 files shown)


ThibG:
A race condition remains, where the favorite API call succeeded and the toot was deleted in the meantime. In this case, the toot is still in the redux store, but at least nothing crashes.

ykzts (approved):

rapidocuento (approved):

rapidocuento (approved):

PR #9274, opened 2018-11-12
 Set mimetype for .gz manually

### LINKED ISSUE #9273: BackupService throws Aws::S3::Errors::SignatureDoesNotMatch when configured with Google Cloud Storage(S3 compatible API)
```
Aws::S3::Errors::SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.
```

unstabler:
fixes #9273

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -2,6 +2,10 @@
 
 Paperclip.options[:read_timeout] = 60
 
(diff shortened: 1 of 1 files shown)


mayaeh:
This will cause 404 errors in the OpenStack Swift environment.
I will investigate a little later.

PR #9278, opened 2018-11-13
Better WebUI error handling

ClearlyClaire:
- Do not crash when a column type does not exist (this is a rare occasion but it happens on glitch-soc since we have both front-end available and different column types)
- Add an Error Boundary component to display an error message and some hopefully helpful links when the WebUI crashes, instead of having an empty page


ThibG:
Added some styling and reworked the messages, it now looks like this:

screenshot_2018-11-14 dev instance

I wanted to add the stack trace for easy bug reporting, but that looks like to be a massive pain.

ThibG:
Added a way to display the stack trace on supported browsers:

screenshot_2018-11-14 dev instance 1

screenshot_2018-11-14 dev instance 2

ThibG:
As you mentioned in discord that you didn't want the elephant friend here, I updated it as follows:

screenshot_2018-11-27 dev instance

PR #9293, opened 2018-11-16
Touch account on successful response, change char shown when culled

renatolond:
Just the color is not enough change since not everyone uses colored
terminals.
Touching the account makes it so that the account is not in the
threshold window in case of running again

### DIFF
--- a/lib/mastodon/accounts_cli.rb
+++ b/lib/mastodon/accounts_cli.rb
@@ -242,8 +242,9 @@ module Mastodon
           end
 
           culled += 1
-          say('.', :green, false)
+          say('+', :green, false)
         else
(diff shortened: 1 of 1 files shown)


ClearlyClaire (approved):
I'm just slightly uneasy about touching the accounts during dry run, but since this is for accounts that are reachable, I guess it's ok.

renatolond:
My rationale behind it was this: I wanted to run a dry-run to check how many accounts/domains would be reported. Problem is that the culling takes some hours over here and then those accounts that were reachable would all be checked again if the results were acceptable.
I think creating yet-another-mode would be a bad idea, but indeed changing stuff on dry-run is kinda of a lie (since it's not really a dry-run anymore).

PR #9295, opened 2018-11-16
Move account counters to separate table

### LINKED ISSUE #7908: Split tables based on column hotness

Gargron:
`statuses_count` is the most updated column probably among all tables, because it goes up every time you post a new status. By extracting it to a table with only a few columns, updates can run faster and the dead tuples left behind until next vacuum are smaller.

Fix #7908 

Instructions:

1. Run `SKIP_POST_DEPLOYMENT_MIGRATIONS=true rails db:migrate` first
2. Restart Mastodon
3. Run rails db:migrate

### DIFF
--- a/app/controllers/admin/instances_controller.rb
+++ b/app/controllers/admin/instances_controller.rb
@@ -31,7 +31,7 @@ module Admin
     end
 
     def subscribeable_accounts
-      Account.with_followers.remote.where(domain: params[:by_domain])
+      Account.remote.where(protocol: :ostatus).where(domain: params[:by_domain])
     end
 
     def filter_params
--- a/app/models/concerns/account_counters.rb
+++ b/app/models/concerns/account_counters.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module AccountCounters
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :account_stat, inverse_of: :account
+    after_save :save_account_stat
+  end
+
+  delegate :statuses_count,
+           :statuses_count=,
+           :following_count,
+           :following_count=,
+           :followers_count,
+           :followers_count=,
+           :increment_count!,
+           :decrement_count!,
+           to: :account_stat
+
+  def account_stat
+    super || build_account_stat
+  end
+
+  private
+
+  def save_account_stat
+    return unless account_stat&.changed?
+    account_stat.save
+  end
+end
--- a/app/controllers/api/v1/accounts/follower_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/follower_accounts_controller.rb
@@ -25,7 +25,7 @@ class Api::V1::Accounts::FollowerAccountsController < Api::BaseController
   end
 
   def default_accounts
-    Account.includes(:active_relationships).references(:active_relationships)
+    Account.includes(:active_relationships, :account_stat).references(:active_relationships)
   end
 
   def paginated_follows
--- a/app/controllers/api/v1/accounts/following_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/following_accounts_controller.rb
@@ -25,7 +25,7 @@ class Api::V1::Accounts::FollowingAccountsController < Api::BaseController
   end
 
   def default_accounts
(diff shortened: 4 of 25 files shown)


app/models/concerns/account_counters.rb
```diff
@@ -0,0 +1,56 @@
...
+  def followers_count
+    account_stat&.followers_count || 0
+  end
+
+  def followers_count=(val)
+    (account_stat || build_account_stat).followers_count = val
+  end
+
+  def increment_count!(key)
+    update_account_stat!(key => public_send(key) + 1)
```
nightpool: i think this should maybe be `attributes[key]`? or something similar, i can't remember the precise syntax offhand.

app/controllers/admin/instances_controller.rb
```diff
@@ -31,7 +31,7 @@ def ordered_instances
     end
 
     def subscribeable_accounts
-      Account.with_followers.remote.where(domain: params[:by_domain])
+      Account.remote.where(protocol: :ostatus).where(domain: params[:by_domain])
```
Gargron: Note: This query has been wrong for a long time. I think it hasn't been updated since ActivityPub was introduced. It would send PuSH subscriptions to ActivityPub accounts. It also would *not* correctly filter accounts that had local followers. So this is a net improvement.

app/models/concerns/account_counters.rb
Gargron: That's if the counter was on the same table as before. We *want* to call the explicitly defined `followers_count` method here.

app/models/concerns/account_counters.rb
```diff
@@ -0,0 +1,59 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :account_stat, inverse_of: :account
+    after_initialize :set_account_stat
+    after_save :save_account_stat
+  end
+
+  def statuses_count
+    account_stat&.statuses_count || 0
```
nightpool: should be `delegate :statuses_count, :statuses_count=, :following_count, :following_count=, :followers_count, :followers_count=, to: :account_stat`

and then handle the case of nil account_stat in the `set_account_stat` hook

nightpool (changes requested):

app/models/concerns/account_counters.rb
nightpool: `%w(statuses following followers).flat_map { |i| [:"#{i}_count", :"#{i}_count=" }` is even a little cuter, your call.

app/models/concerns/account_counters.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def decrement_count!(key)
+    update_account_stat!(key => [public_send(key) - 1, 0].max)
+  end
+
+  private
+
+  def set_account_stat
+    self.account_stat = build_account_stat if account_stat.nil?
```
nightpool: per our conversation, this should handle migrating the counts from the existing record if there's no account_stat record.

app/models/concerns/account_counters.rb
```diff
@@ -0,0 +1,42 @@
...
+           :following_count=,
+           :followers_count,
+           :followers_count=,
+           to: :account_stat
+
+  def account_stat
+    super || build_account_stat
+  end
+
+  def increment_count!(key)
```
nightpool: i think these should live on AccountStat and be delegated also

app/models/concerns/account_counters.rb
```diff
@@ -0,0 +1,42 @@
...
+    update_account_stat!(key => public_send(key) + 1)
+  end
+
+  def decrement_count!(key)
+    update_account_stat!(key => [public_send(key) - 1, 0].max)
+  end
+
+  private
+
+  def update_account_stat!(attrs)
```
nightpool: and then we don't need to care about this method

nightpool (approved):

PR #9296, opened 2018-11-16
Sort self-replies to the top of the thread

### LINKED ISSUE #6463: [Feature request] If there are replies-to-self in a thread, group them all at the top
Behold:

screen shot 2018-02-11 at 20 08 42

If someone replies before I'm done with my thread it makes the thread hard to read. I think it would be cool if all the replies-to-self could be grouped together at the top, even if some people replied before all the posts were posted, just for continuity of a thread.

* * * *

Gargron:
Fix #6463

### DIFF
--- a/app/models/concerns/status_threading_concern.rb
+++ b/app/models/concerns/status_threading_concern.rb
@@ -86,6 +86,9 @@ module StatusThreadingConcern
 
     # Order ancestors/descendants by tree path
     statuses.sort_by! { |status| ids.index(status.id) }
+
+    # Bring self-replies to the top
+    statuses.sort_by! { |status| status.in_reply_to_account_id == status.account_id ? -1 : 0 }
   end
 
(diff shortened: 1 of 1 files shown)


ykzts (approved):

app/models/concerns/status_threading_concern.rb
```diff
@@ -86,6 +86,9 @@ def find_statuses_from_tree_path(ids, account)
 
     # Order ancestors/descendants by tree path
     statuses.sort_by! { |status| ids.index(status.id) }
+
+    # Bring self-replies to the top
+    statuses.sort_by! { |status| status.in_reply_to_account_id == status.account_id ? -1 : 0 }
```
ThibG: How reliable is that (is there any guarantee to respect the previous order)? Also, what happens for self-replies after someone else replied? (e.g, A replies to themselves in a reply to B, which replied to the original post from A)

PR #9298, opened 2018-11-17
Simplify hashtag query service, perf increase for 'any' tags

gdpelican:
For #9279 

This allows us to avoid an unsavoury `or` join and instead do a simple query beforehand to get the ids of any tags we're interested in in each hashtag query

### DIFF
--- a/app/services/hashtag_query_service.rb
+++ b/app/services/hashtag_query_service.rb
@@ -2,15 +2,14 @@
 
 class HashtagQueryService < BaseService
   def call(tag, params, account = nil, local = false)
-    any  = tags_for(params[:any])
+    tags = tags_for(Array(tag.name) | Array(params[:any])).pluck(:id)
     all  = tags_for(params[:all])
     none = tags_for(params[:none])
 
-    @query = Status.as_tag_timeline(tag, account, local)
-                   .tagged_with_all(all)
-                   .tagged_with_none(none)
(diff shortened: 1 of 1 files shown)


gdpelican:
^^ (cc @Gargron)

app/services/hashtag_query_service.rb
```diff
@@ -2,15 +2,14 @@
...
     all  = tags_for(params[:all])
     none = tags_for(params[:none])
 
-    @query = Status.as_tag_timeline(tag, account, local)
-                   .tagged_with_all(all)
-                   .tagged_with_none(none)
-    @query = @query.distinct.or(self.class.new.call(any, params.except(:any), account, local).distinct) if any
-    @query
+    Status.distinct
+          .as_tag_timeline(tags, account, local)
```
Gargron: What happens if `tags` is nil when passed here?

gdpelican: Hmm, well I was assuming here that 'tag' would always be passed into the method, so
```
tags = tags_for(Array(tag.name))
```
_should_ always be present.

If it did end up being empty (if nil were passed in as the `tag` param, for example), I believe tags would go into `Status#tagged_with`,
```
  scope :tagged_with, ->(tag) { joins(:statuses_tags).where(statuses_tags: { tag_id: tag }) }
```
which should return an empty relation

Gargron (approved):

PR #9299, opened 2018-11-17
Improvements

indie:
We need to be able to write intelligently about topics; this cannot be done in 500 characters for these topics.  Updating the max chars to a few over 1300 per status


indie:
ACK sorry.  I am trying to merge this change into indie:mastodon, not the tootsuite one.

PR #9310, opened 2018-11-19
Update Nginx config for Nanobox apps

danhunsaker:
The Nanobox files have gotten out of sync, a touch, with what Masto needs for Nginx settings. This PR updates them accordingly.

### DIFF
--- a/nanobox/nginx-local.conf
+++ b/nanobox/nginx-local.conf
@@ -38,7 +38,7 @@ http {
 
         root /app/public;
 
-        client_max_body_size 8M;
+        client_max_body_size 80M;
 
         location / {
             try_files $uri @rails;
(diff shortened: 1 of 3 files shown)


Gargron:
Oops, I forgot they were there! There are now example configs in dist/, could they be re-used here somehow?

danhunsaker:
Unfortunately not. Routing between components (web vs streaming) is handled before any component's Nginx service sees anything, so at best the bits about routing between components would be superfluous. More importantly, though, SSL is handled at the load balancer (each app gets its own LB), so each component's Nginx service needs to listen via HTTP on port 8080; Nanobox web components never see incoming connections on port 80 _or_ 443. That means the 80→443 redirect needs to be rewritten to check the standard proxy header `X-Forwarded-Proto` and redirect if the value isn't `https`. It also means the SSL directives need to be removed entirely. Throw in the fact that the on-disk paths are different.... It's simpler to maintain these separately, in my experience.

Gargron (approved):

PR #9325, opened 2018-11-21
Allow more than 4 attachments per status in the backend

ClearlyClaire:
This commit allows more than 4 attachments per status in the backend. Such attachments were already processed, but not attached to any status.

Marked as WiP because changes to the WebUI would be needed to make more sense:
- [ ] Add a “and more” or something in the inline media gallery when there are more than 4 pics (the modal would allow seeing them all)
- [ ] Add a way to access all media when there are more than one and the first is a video

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -67,7 +67,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
         visibility: visibility_from_audience,
         thread: replied_to_status,
         conversation: conversation_from_uri(@object['conversation']),
-        media_attachment_ids: process_attachments.take(4).map(&:id),
+        media_attachment_ids: process_attachments.map(&:id),
       }
     end
   end


Gargron:
Even if they get processed, they get cleaned out later if they are not attached. I don't think this is a good idea

ThibG:
I don't think silently dropping attachments is a good idea either?

Gargron:
1. Storing more media attachments translates directly into hosting costs for admins, so limiting how much can be stored is very important.
2. Especially since the UI is designed around displaying 4 attachments at most, having hidden but stored attachments is worse than simply cutting off attachments, because the invisible files could be big (or illegal)
3. If you relax the processing limitations, people will *absolutely* take advantage of it. See: ridiculous character limits. It would be that, but worse, because of (1)

Gargron:
We could limit the number of processed attachments too

trwnh:
Perhaps the "and more" link would load the original post on the original domain? That could prevent Mastodon from having to cache infinite media while still signalling that more than 4 attachments exist.

giromide:
To ThibG's first point just above, if someone wants to drop ten related images into their Mastodon instance, I'd much prefer to see them in one post than three.

Gargron:
Yes it's no secret to me that given absolute freedom people would dump 100-image albums, ZIP archives, ISO images and whatnot into their posts. What's good for the poster isn't necessarily good for the reader, or the hoster. The system is designed around these constraints. I say, chop it off after 4 processed attachments.

trwnh:
>given absolute freedom, people would dump 100-image albums [...] and whatnot into their posts

Surely ignoring the freedom of other systems entirely is a bad move? Maybe vanilla Mastodon doesn't need to locally cache more than 4, but some admins will want to cache more, or host more, etc. At the very least, the existence of attachments should be acknowledged, even if it is not supported on the local instance.

MirceaKitsune:
I'd like to see this change as well. Only 4 attachments per post is pretty limiting: By default I'd go with allowing something like 16 by default, while of course letting only first 4 show up in the preview. If more than 4 attachments exist on the post, a text over the previews can indicate how many more... for instance if there are 7 attachments it would say "+3" (4 + 3 = 7).

A semi-related concern: Are mixed attachments officially supported? Can you attach both images, video, and audio in the same post? Ideally this can be allowed too with multiple attachments.

app/lib/activitypub/activity/create.rb
```diff
@@ -67,7 +67,7 @@ def process_status_params
         visibility: visibility_from_audience,
         thread: replied_to_status,
         conversation: conversation_from_uri(@object['conversation']),
-        media_attachment_ids: process_attachments.take(4).map(&:id),
+        media_attachment_ids: process_attachments.map(&:id),
```
Gargron: Could you turn this into a PR that does not process more than 4 attachments? Most clients won't be able to display more than 4 right now, and the important thing is not storing stuff we don't use.

Gargron (changes requested):

PR #9329, opened 2018-11-22
Fix connect timeout not being enforced

Gargron:
The loop was catching the timeout exception that should stop execution, so the next IP would no longer be within a timed block, which led to requests taking much longer than 10 seconds.

Regression from #6813

Now, the total timeout is 1 second for DNS look-up, and 10 seconds for socket opening, wherein 2 IPs are tried one after the other within those 10 seconds. So if there is only one IP, we wait 10 seconds, if there's two IPs, we wait 5 seconds each.

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -139,16 +139,21 @@ class Request
   class Socket < TCPSocket
     class << self
       def open(host, *args)
-        return super host, *args if thru_hidden_service? host
+        return super(host, *args) if thru_hidden_service?(host)
+
         outer_e = nil
+
         Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
           begin
-            raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
-            return super address.ip_address, *args
+            raise Mastodon::HostValidationError if PrivateAddressCheck.private_address?(IPAddr.new(address.ip_address))
+            return super(address.ip_address, *args)
+          rescue HTTP::TimeoutError => e
+            raise e
           rescue => e
             outer_e = e
           end
         end
+
         raise outer_e if outer_e
       end
 
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -146,7 +146,7 @@ class Request
         time_slot = 10.0 / addresses.size
 
         addresses.each do |address|
-          ::Timeout::timeout(time_slot, HTTP::TimeoutError) do
+          ::Timeout.timeout(time_slot, HTTP::TimeoutError) do
             begin
               raise Mastodon::HostValidationError if PrivateAddressCheck.private_address?(IPAddr.new(address.ip_address))
               return super(address.ip_address, *args)
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -146,13 +146,14 @@ class Request
         time_slot = 10.0 / addresses.size
 
         addresses.each do |address|
-          ::Timeout.timeout(time_slot, HTTP::TimeoutError) do
-            begin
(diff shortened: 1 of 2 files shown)


Gargron:
cc @rtucker @akihikodaki

ThibG:
This raises an exception and stops trying to connect at the first IP that timeouts, right?
If so, that negates any benefit from #6813

Gargron:
@Thibg The alternative is that it moves to the 2nd IP and then never times out (measured 120 seconds in a test before something times out somehow (on 2nd IP after first, specifically))

Gargron:
We can put a Timeout::timeout in the foreach block, but then I think we should also limit the maximum number of iterated IPs, and scale individual timeouts such that the total does not exceed 10s.

ThibG:
@Gargron outch. In this case, yeah, that's probably better for now, but we ought to find a better way.

akihikodaki (approved):

rtucker:
Yeah, doing it on the full block is probably best.  The downside of this PR is that it's fixing reachability to one set of broken instances while breaking reachability to another set of broken instances.

I'd oppose merging this until it can be properly fixed, but, well, too late now.

nightpool:
@rtucker it hasn't been merged yet.


I would generally oppose merging this but I think that on balance it's better to have stricter timeouts then it is to have perfect reachability, especially if it's going to be dependent on stuff like DNS order anyways.

Gargron:
`Timeout::timeout` spawns a thread so the less we use it, the better... So if I move the timeout to within the loop, I'd want to make sure the `HTTP::Timeout#connect` method doesn't also call it. I remember reading that `Timeout::timeout` calls the block directly if passed a 0, so that's probably what I will have to do...

app/lib/request.rb
```diff
@@ -94,7 +94,7 @@ def key_id
   end
 
   def timeout
-    { connect: 10, read: 10, write: 10 }
+    { connect: nil, read: 10, write: 10 }
```
Gargron: `Timeout.timeout(nil)` yields the block immediately, without spawning any threads

app/lib/request.rb
```diff
@@ -139,16 +139,23 @@ def body_with_limit(limit = 1.megabyte)
...
-        Addrinfo.foreach(host, nil, nil, :SOCK_STREAM) do |address|
-          begin
-            raise Mastodon::HostValidationError if PrivateAddressCheck.private_address? IPAddr.new(address.ip_address)
-            return super address.ip_address, *args
-          rescue => e
-            outer_e = e
+        return super(host, *args) if thru_hidden_service?(host)
+
+        outer_e   = nil
+        addresses = Addrinfo.getaddrinfo(host, nil, nil, :SOCK_STREAM).take(2)
```
Gargron: Why 2: AAAA, A is the most common use case. Spawning threads is not cheap, `Timeout.timeout` spawns threads, so we don't want to spawn more than 2 (one more than before)

rtucker:
I've had coffee now, and yeah, this seems reasonable.  :)

ClearlyClaire (approved):

staticsafe:
Hey y'all,

I noticed that this change is causing a change in address family preference, I noticed mastodon.social switching over to IPv4 for outbound federation as opposed to using IPv6 preferred if available.

148.251.134.157 - - [22/Nov/2018:22:57:32 +0000] "POST /inbox HTTP/1.1" 202 36 "-" "http.rb/3.3.0 (Mastodon/2.6.1; +https://mastodon.social/)"

That seems to be a regression, is this some sort of default behaviour in Resolv::DNS, if so can we change this behaviour?

Eriner:
@Gargron, any chance I can be credited in the release notes for reporting this issue?

Eriner:
This issue (and #9381) have been assigned CVE-2018-21018, which was resolved with Mastodon release v2.6.3.

PR #9348, opened 2018-11-25
check_boxes label should use display: inline-block

takayamaki:
Check boxes labels of chosen language in `/settings/preferences` are too wide.
It cause accidentally click.
Users said to me about this, "Public timeline does not work."

|before|after|
|:-:|:-:|
|2018-11-25-232203_1920x983_scrot|2018-11-25-232133_1920x983_scrot|

### DIFF
--- a/app/javascript/styles/mastodon/forms.scss
+++ b/app/javascript/styles/mastodon/forms.scss
@@ -266,7 +266,7 @@ code {
         font-family: inherit;
         font-size: 14px;
         color: $primary-text-color;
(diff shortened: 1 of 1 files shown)


Gargron (approved):

mayaeh (approved):

fvh-P:
This modification is not bad, however, first of all, it should stop putting a lot of check boxes.
Now the LANGUAGES preference is taking up a lot of space in the preferences page...

Issue #7429, opened 2018-05-09
No Volume slider for videos


szernex:
There doesn't seem to be a volume slider for HTML5 videos, you can only mute/unmute the video, where it blasts at full volume. Tested on Chrome 66.0.3359.139 (Official Build) (64-bit) and Vivaldi 1.15.1147.36 (Stable channel) (32-bit).

Undigon:
It's the same on Mozilla Firefox 60.

imbsky:
Is there any update on this?

no-boot-device:
Having a volume slider has been requested by many users.

PR #9366, opened 2018-11-27
Volume sliders for videos #7429

sumdog:
This is a patch to add volume support to videos and change the default volume to 50%

Questions:

* I didn't see any selenium or web browser tests. Are there tests I need to update for this component?
* I didn't add in Intl/messages for the volume slider since there weren't any for the seek slider. Is that alright or should I add them and to what component?

I've tested this manually on Firefox desktop, Chrome desktop, Firefox mobile and chrome mobile. It matches the behaviour of the seek slider.


Gargron (approved):
:100: :tada: :ok_hand:

PR #9375, opened 2018-11-27
Do not leak existence of e-mail addresses during login attempt in French translation

### LINKED ISSUE #9374: Discovery attack on login form
### Expected behaviour

When attempting to login, the use of bad credentials should provide a single and consistent error message saying credentials are invalid, without finer granularity.

### Actual behaviour

When using an incorrect email address, the login form actually states the email is wrong, instead of the credentials pair.
That means email enumeration of registered users is possible ans consists itself in a form of discovery attack.

1. Reach the login form of any mastodon instance
2. Try to login with non-registered email address

### Specifications

The instance that behavior has been noticed on was v2.6.1

mkljczk:
Fix #9374

### DIFF
--- a/config/locales/devise.fr.yml
+++ b/config/locales/devise.fr.yml
@@ -8,7 +8,7 @@ fr:
     failure:
       already_authenticated: Vous êtes déjà connecté⋅e.
       inactive: Votre compte n’est pas encore activé.
-      invalid: "%{authentication_keys} incorrect."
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Berbe:
Wow that is unexpected: how are translations able to tell one case (non-existent email) from the other (invalid credentials)?
Sounds like a design flaw.

PR #9378, opened 2018-11-27
Change the icon "reply" to "reply-all" when reply to other accounts

mayaeh:
After #9228 , the reply icon is unified to "reply" and the "Show thread" message is displayed in the self reply.

I agree with the display of the "Show thread" message.

However, after #9228, I think that it has become difficult to distinguish whether reply to other people is reply or call.

So I changed the icon to “reply-all” when replies to others.

How about this?

before:

after:

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -148,6 +148,7 @@ class StatusActionBar extends ImmutablePureComponent {
 
     let menu = [];
     let reblogIcon = 'retweet';
+    let replyIcon;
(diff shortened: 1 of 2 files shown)


ClearlyClaire (approved):
I agree, this conveys more information than the thread link thingy.

rixx:
Shouldn't the icon still be "reply" if I'll end up replying to only one person, as is the case when that person mentioned only me?

joyeusenoelle:
No; the doubled arrow represents more than one person in the conversation, not more than one recipient.

PR #9382, opened 2018-11-27
Add database statement timeout of 60s

Gargron:
I have not observed any queries with average duration of over 60 seconds. Having a statement timeout should ensure that database deadlocks, if they occur, will not remain forever until manual restart. However, **this has absolutely no effect when using pgBouncer**.

When using pgBouncer, the only way to enforce a statement timeout is setting it on the database directly:

    ALTER DATABASE mastodon_production SET statement_timeout = '60s';

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,8 @@ default: &default
   pool: <%= ENV["DB_POOL"] || ENV['MAX_THREADS'] || 5 %>
   timeout: 5000
   encoding: unicode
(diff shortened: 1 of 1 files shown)


tateisu:
Is this work with migration?

Gargron:
That's a very good question. I don't think Rails is smart enough to omit this for migrations, so I need to include something extra that runs before migrations to reset it...

PR #9384, opened 2018-11-27
Add option to auto-delete old statuses

### LINKED ISSUE #1295: Content auto expire / delete for users
As an end user I would appreciate a setting in mastodon similar to https://www.tweetdelete.net/, which deletes all posts after a certain period of time

Implementation proposal:

- `content_expiration_days` on `user`
    - a profile UI setting for this option
    - `content_expiration_days==None` user option unset
    - `content_expiration_days<=0` user opts out of content deletion
    - `content_expiration_days>0` user has set a content deletion time
- an admin setting `DEFAULT_CONTENT_EXPIRATION`, in `.env.production` used when the user expiration value is unset
- a daily runner that deletes all statuses where `status.created_at + content_expiration_days > datetime.now`

* * * *

Gargron:
Fix #1295 

**Concerns**: Deletes are relatively expensive, and filling the Sidekiq queue with deletes can cause other kinds of content to stall, and user experience to be impacted. This PR limits the impact by scheduling at most 100 deletes per user during a run.

> Why don't you queue a `RemovalWorker` for a later date when a status is created, like ephermal.glitch.social does?

Sidekiq queues are stored in Redis, which is a RAM-based database (it dumps to disk only for persistence reasons). Assuming that a large instance has everyone opt into the auto-delete feature with a long delay, the Sidekiq queue for future deletes will grow at the same pace as all new statuses that are created, exhausting available RAM. It is not a good idea to use the Sidekiq queue for long-term storage.

Therefore, this PR queries old statuses periodically instead.

### DIFF
--- a/app/workers/deletion_schedule_worker.rb
+++ b/app/workers/deletion_schedule_worker.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class DeletionScheduleWorker
+  include Sidekiq::Worker
+
+  sidekiq_options unique: :until_executed, retry: 0
+
+  def perform(account_id, delay)
+    RemovalWorker.push_bulk(status_ids(account_id, snowflake_at(Time.now.utc - delay.seconds)))
+  end
+
+  private
+
+  def snowflake_at(time)
+    Mastodon::Snowflake.id_at(time)
+  end
+
+  def status_ids(account_id, cut_off_id)
+    Status.where(account_id: account_id)
+          .where(Status.arel_table[:id].lt(cut_off_id))
+          .reorder(id: :asc)
+          .limit(100)
+          .pluck(:id)
+  end
+end
--- a/app/models/deletion_schedule.rb
+++ b/app/models/deletion_schedule.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+# == Schema Information
+#
+# Table name: deletion_schedules
+#
+#  id         :bigint(8)        not null, primary key
+#  user_id    :bigint(8)
+#  delay      :integer          default(604800), not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class DeletionSchedule < ApplicationRecord
+  belongs_to :user
+
+  validates :delay, presence: true, numericality: { only_integer: true, greater_than: 7.days.seconds }
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -70,6 +70,7 @@ class User < ApplicationRecord
 
   has_many :applications, class_name: 'Doorkeeper::Application', as: :owner
   has_many :backups, inverse_of: :user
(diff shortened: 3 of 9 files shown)


renatolond:
The only thing I thought about is what you already commented on your concern, how a huge number of deletes would affect the queues.
What happens if a user deletes a status that has been inserted in the queue and has not been processed yet?

kit-ty-kate:
I'm sorry to ask this but is sidekiq a priority queue?

nightpool:
@kit-ty-kat sidekiq had the concept of queue priorities, but not job priorities, as I understand it. you can read more about sidekiq here: https://github.com/mperham/sidekiq

nightpool:
@renatoland I'm not sure I understand your comment, since there is no "queue" of statuses to be deleted in the code.......

app/workers/deletion_schedule_worker.rb
```diff
@@ -0,0 +1,25 @@
...
+  sidekiq_options unique: :until_executed, retry: 0
+
+  def perform(account_id, delay)
+    RemovalWorker.push_bulk(status_ids(account_id, snowflake_at(Time.now.utc - delay.seconds)))
+  end
+
+  private
+
+  def snowflake_at(time)
+    Mastodon::Snowflake.id_at(time)
```
nightpool: I think time+1 would be more correct, but less readable

renatolond:
@nightpool I meant if a RemovalWorker job to delete was inserted to remove status XXX and status XXX was removed by the user in the mean time, but I just saw that RemovalWorker already treats that now

nightpool (approved):
I think this makes sense from an implementation perspective, although obviously the controller and UI code have yet to be written. 

left a couple small nits

nightpool:
@renatoland nothing about that has changed by adding this, it's completely identical to the case of the user hitting the delete button twice

Cassolotl:
Sorry if this is implied somehow in the first post and I'm just not understanding it, but:

Is this a user-side feature that lets people set their posts to automatically expire after a certain amount of time, and if I boost someone's post it might be deleted and removed from my profile? Or is it a thing that deletes old posts without interactions (boosts, replies, stars) to save server space, and people can opt out of that?

nightpool:
@cassolotl the user-side code hasn't been written yet, but this is a mirror
of codl's forget and ephemeral.glitch.social, in that it deletes all posts
a user has made after a certain time period.

p0sty:
just want to say if this makes for a more default 'snapchat' like option for mastodon that is bloody excellent and very welcome.

sternenseemann:
Why the close?

Also I wonder whether it would be possible to reduce the cost of expiration by making it an attribute of a toot. This way the deletion wouldn't need to be federated because the expiration date would already have been federated when posting, so every instance only has to delete expired toots only for itself. Correct me if I'm wrong on anything here.

PR #9387, opened 2018-11-28
Add moderation API

### LINKED ISSUE #7143: Proposal: Moderation APIs
Some of the things that'd be good to implement for moderation really fall outside of what should necessarily be in the Mastodon core (for instance, crowdsourced moderation where you can browse a set of images, mark all those that are sensitive or against policy and submit for review)

In order to do this, there should be a few moderation APIs, currently, I'm thinking of these endpoints:
- get all media that's not marked as sensitive
- get all profile pictures that are not the default (ordered by recently uploaded)
- get all header pictures (ordered by recently uploaded)
- webhook for reports; use cases
  - figuring out who to notify based on an on-call schedule when a report comes in because simply emailing every moderator all the time does not scale and leads to burn out
  - integrating with third-party moderation providers
- fetch one or more reports as JSON

The first three would allow us to crowdsource the moderation of media, without needing people to manually report from the app or UI each offence — e.g., you could see a page with 10 images, and report multiple violations all at once (I know this is the method that Tastebuds.fm used for moderating its content, they also gamified it)

The webhook would allow us to reduce email and notification fatigue on our moderators, which enables happier moderators. However, I'm pretty sure this shouldn't necessarily be built in, as not ever mastodon instance operates at a scale where this is needed.

Longer team ideas:
- an endpoint for bulk updates; use cases:
  - responding to crowdsourced profile/header picture reports in-bulk
  - for integrating with the third-party moderation services where they are actually allowed to take action
- webhook or stream for all media uploads or content posts
  - integrating with third-party moderation providers
  - automating reporting with machine learning services which auto-detect abuse

* * * *

### LINKED ISSUE #8580: Allow moderation through API
Moderators of Mastodon instances can currently take several actions through the web UI that they cannot accomplish through the public API:
- Delete toots
- Silence users
- Silence instances
- Resolve/unresolve reports
- etc.

This issue exists to suggest an API be created that would allow moderators to take every action that they can currently take through the web UI.  This would enable third-party tools to be built to help with the moderation process, which would magnify the effectives of moderators.  This, in turn, would help strengthen one of Mastodon's key selling points: that it provides a better-moderated environment than other social networks.

Potential use-cases for a moderation API include:
- Tools to enable easier moderation from a mobile device
- Tools to bulk-resolve certain reports
- Tools to block spam domains based on reported spam
- Tools to manage moderation workload between multiple mods
- Tools to automatically take certain moderation action in extreme cases (e.g., racial slurs)
- Many others that we cannot think of right now.

I believe that Mastodon user @balrogboogie@ceilidh.space is working on a related API, but I don't know the status of that work or whether it would close this issue.

* * * *

Gargron:
- Fix #8580
- Fix #7143

|Method|Description|
|-------|------------|
|`GET /api/v1/admin/accounts`|Get accounts, with params `local`, `remote`, `by_domain`, `active`, `pending`, `disabled`, `silenced`, `suspended`, `username`, `display_name`, `email`, `ip`, `staff`|
|`GET /api/v1/admin/accounts/:id`|Get a specific account|
|`POST /api/v1/admin/accounts/:id/enable`|Re-enable account|
|`POST /api/v1/admin/accounts/:id/approve`|Approve pending account|
|`POST /api/v1/admin/accounts/:id/reject`|Reject pending account|
|`POST /api/v1/admin/accounts/:id/unsilence`|Unsilence account|
|`POST /api/v1/admin/accounts/:id/unsuspend`|Unsuspend account|
|`GET /api/v1/admin/reports`|Get reports, with params `resolved`, `account_id`, `target_account_id`|
|`GET /api/v1/admin/reports/:id`|Get a specific report|
|`POST /api/v1/admin/reports/:id/assign_to_self`|Assign report to self|
|`POST /api/v1/admin/reports/:id/unassign`|Unassign report from self|
|`POST /api/v1/admin/reports/:id/reopen`|Re-open report|
|`POST /api/v1/admin/reports/:id/resolve`|Close report|
|`POST /api/v1/admin/accounts/:id/action`|Perform a moderation action on an account, with params `type`, `report_id`, `warning_preset_id`, `text`, `send_email_notification`|

JSON returned from accounts APIs includes `id`, `username`, `domain`, `created_at`, `email`, `role`, `ip`, `confirmed`, `suspended`, `silenced`, `disabled`, `approved`, and `account` which renders the standard account data used in the rest of the application. Additionally, there are `locale`, `invite_request`, `created_by_application_id` and `invited_by_account_id`

JSON returns from reports APIs includes `id`, `action_taken`, `comment`, `created_at`, `updated_at`, `account_id`, `target_account_id`, `assigned_account_id`, `action_taken_by_account_id`, and `statuses` which renders the standard status data used in the rest of the application.

### DIFF
--- a/app/controllers/admin/accounts_controller.rb
+++ b/app/controllers/admin/accounts_controller.rb
@@ -127,6 +127,7 @@ module Admin
         :by_domain,
         :active,
         :pending,
+        :disabled,
         :silenced,
         :suspended,
         :username,
--- a/app/controllers/api/v1/admin/account_actions_controller.rb
+++ b/app/controllers/api/v1/admin/account_actions_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class Api::V1::Admin::AccountActionsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :'admin:write', :'admin:write:accounts' }
+  before_action :require_staff!
+  before_action :set_account
+
+  def create
+    account_action                 = Admin::AccountAction.new(resource_params)
+    account_action.target_account  = @account
+    account_action.current_account = current_account
+    account_action.save!
+
+    render_empty
+  end
+
+  private
+
+  def set_account
+    @account = Account.find(params[:account_id])
+  end
+
+  def resource_params
+    params.permit(
+      :type,
+      :report_id,
+      :warning_preset_id,
+      :text,
+      :send_email_notification
+    )
+  end
+end
--- a/app/controllers/api/v1/admin/accounts_controller.rb
+++ b/app/controllers/api/v1/admin/accounts_controller.rb
@@ -1,0 +1,128 @@
+# frozen_string_literal: true
+
+class Api::V1::Admin::AccountsController < Api::BaseController
+  include Authorization
+  include AccountableConcern
+
+  LIMIT = 100
+
+  before_action -> { doorkeeper_authorize! :'admin:read', :'admin:read:accounts' }, only: [:index, :show]
+  before_action -> { doorkeeper_authorize! :'admin:write', :'admin:write:accounts' }, except: [:index, :show]
+  before_action :require_staff!
+  before_action :set_accounts, only: :index
+  before_action :set_account, except: :index
+  before_action :require_local_account!, only: [:enable, :approve, :reject]
+
+  after_action :insert_pagination_headers, only: :index
+
+  FILTER_PARAMS = %i(
+    local
+    remote
+    by_domain
+    active
+    pending
+    disabled
+    silenced
+    suspended
+    username
+    display_name
+    email
+    ip
+    staff
+  ).freeze
+
+  PAGINATION_PARAMS = (%i(limit) + FILTER_PARAMS).freeze
+
+  def index
+    authorize :account, :index?
+    render json: @accounts, each_serializer: REST::Admin::AccountSerializer
+  end
+
+  def show
+    authorize @account, :show?
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  def enable
+    authorize @account.user, :enable?
+    @account.user.enable!
+    log_action :enable, @account.user
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  def approve
+    authorize @account.user, :approve?
+    @account.user.approve!
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  def reject
+    authorize @account.user, :reject?
+    SuspendAccountService.new.call(@account, including_user: true, destroy: true, skip_distribution: true)
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  def unsilence
+    authorize @account, :unsilence?
+    @account.unsilence!
+    log_action :unsilence, @account
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  def unsuspend
+    authorize @account, :unsuspend?
+    @account.unsuspend!
+    log_action :unsuspend, @account
+    render json: @account, serializer: REST::Admin::AccountSerializer
+  end
+
+  private
+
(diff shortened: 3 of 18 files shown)


stom79:
@Gargron  is it possible to add a parameter in Account entity like "is_moderator" to display moderation features only if needed?

Gargron:
This is still far from everything that admins can do through the website. However, it should be possible to automate the most pressing tasks with this: Approval of pending accounts and auto-modding of reports. What is not yet possible but could be important as part of auto-modding applications is modifying statuses, such as updating `sensitive` or removing them.

bclindner:
Great stuff, thanks! This should help with Fedi's still-at-large spam problem.

One thing I can think of that could be added is the ability to get recent accounts and their posts, so that an automated script could pull them and screen their posts to ensure they aren't following common spam patterns.

joyeusenoelle:
Can `pending` be added as a parameter to `GET /api/v1/admin/accounts`? Or is there another way to get a list of pending accounts that I don't see?

Gargron:
> Can `pending` be added as a parameter to `GET /api/v1/admin/accounts`? Or is there another way to get a list of pending accounts that I don't see?

It's supported I forgot to list it.

> One thing I can think of that could be added is the ability to get recent accounts and their posts

Accounts returned from the accounts API are ordered by recent-first. The non-admin REST API could be used to retrieve those accounts' posts, I believe.

slice:
Endpoints to manage federation (e.g. suspending instances) would be nice to have, if they aren't already a thing.

bclindner:
I concur, but would argue that's not something most people would want to be doing from a mobile app or automated script, at least with the current state of the Fediverse.

slice:
@bclindner I'm sure it would be nice to have regardless; what I had in mind when suggesting that was the ability to write a small CLI script to let me suspend an instance from my terminal window instead of having to jump through hoops in the web interface.

Gargron:
@slice I'm sure a new `tootctl` command would suffice for that

slice:
@Gargron Yeah, `tootctl` seems like it would be more appropriate for this, actually.

Kjwon15:
moderation note missing?

stom79:
@Gargron does ` /accounts/verify_credentials` that returns an Account will also return the role like in `/api/v1/admin/accounts/:id/action`?
That's an important element for displaying admin features inside applications. Also what are the possible values for role: user, moderator, admin?

stom79:
It's easy to get the entity Account for the reported one (ie via statuses->getAccount) but for the account who reports, it needs an extra call from account_id. Would it be possible to attach Account entities inside the report entity?

Gargron:
Okay, that will make the API a bit slower, though. I was even considering GraphQL for this to avoid sending unnecessary attributes and entities but I guess it'd be kinda weird to add GraphQL just for the admin parts of the API.

Gargron:
Okay, that information is now included in the reports API. This kind of makes sense too because it means you can give an app access to reports only and it will be able to see data of reported accounts only.

ykzts (approved):

ThisIsMissEm:
@gargron If you are interested in developing a GraphQL API to mastodon, please drop me a message, and I'd be happy to help with the schema design & implementation.

(Though it could also be a standalone component like the real-time service, which would also then support GraphQL Subscriptions)

ThisIsMissEm:
@gargron, does this support marking media as sensitive or removing images or other media? (I'm looking at this thinking it can't be that hard to add, but it's been a while since I've contributed)

aakritichoudhary:
@Gargron , are there APIs to DELETE and SENSITIVE the post(s) attached to the reported user? These actions can be done from the web UI but I could not find an API for this. 
POST /api/v1/admin/accounts/:id/action - I believe this only works for user level actions and not toot level actions.
Also can we fetch list of all the moderator(s) on an instance via an API call? If so, please specify the API.

ThisIsMissEm:
@aakritichoudhary there isn't currently. There's also not an API to fetch all moderators on a server.

Issue #9359, opened 2018-11-26
npm backdoored package (event-stream) part of Mastodon dependency tree


rtucker:
FYI:

https://github.com/dominictarr/event-stream/issues/116

It looks like npm-run-all depends on ps-tree which depends on event-stream.  I'm not sure if it pulls in an affected version because I'm not sure how NPM stuff works, but it looks like it very possibly does.

Note: it's currently believed that the malicious payload targets crypto-currency-related things and thus would not be activated on a typical Mastodon install, so I don't believe there's a realistic threat beyond the usual "sub-sub-sub-dependency gets taken over by malicious actor" situation w/ npm stuff.  I am not qualified to assess this, however.

Gargron:
```
└─┬ npm-run-all@4.1.3
  └─┬ ps-tree@1.1.0
    └─┬ event-stream@3.3.6
      └── flatmap-stream@0.1.0
```

0.1.0 is not an affected version of flatmap-stream, as I understand.

nightpool:
we're not using an affected version, but we should probably pull the dependency anyway for safety, because now we're only one dependabot upgrade away from having a bad time.

the relevant pull request was #5925

nightpool:
I'm still trying to figure out why that package got upgraded in that pull request.... it doesn't look like any of the dependent packages changed.....

Kjwon15:
flatmap-stream is down on npm registry (https://www.npmjs.com/package/flatmap-stream)
So I cannot build on heroku/dokku system.

PR #9392, opened 2018-11-30
Downgrade event-stream

tribela:
Due to flatmap-stream backdoor, event-stream 3.3.6 is gone.
Related: #9359


nightpool:
it's viable for us to just pull npm-run-all entirely and use && in the
relevant package.json invocation, so we should just do that

nightpool:
it's viable for us to just remove npm-run-all entirely and use && in the relevant package.json invocation, so we should just do that

PR #9412, opened 2018-12-02
Preload common JSON-LD contexts

### LINKED ISSUE #9411: Common JSON-LD contexts should be bundled with Mastodon
In many cases, remote JSON-LD documents have to be loaded to parse/canonicalize JSON-LD documents.
This is the case for outbound statuses, which require at least `https://w3id.org/security/v1`, `https://www.w3.org/ns/activitystreams`, and `https://w3id.org/identity/v1`.

The last one is currently unavailable, so any instance not having it in cache is currently screwed.

Such documents used to be bundled with Mastodon, but that was reverted in d40ef1129e6df5098e150ff3f6ee4eda9b7efcf8. I don't think saving a few MB of RAM was worthwhile here.

ClearlyClaire:
Fixes #9411

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,7 @@ gem 'webpacker', '~> 3.5'
 gem 'webpush'
 
 gem 'json-ld', '~> 2.2'
+gem 'json-ld-preloaded', '~> 2.2'
 gem 'rdf-normalize', '~> 0.3'
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -290,6 +290,10 @@ GEM
     json-ld (2.2.1)
       multi_json (~> 1.12)
       rdf (>= 2.2.8, < 4.0)
+    json-ld-preloaded (2.2.3)
+      json-ld (>= 2.2, < 4.0)
+      multi_json (~> 1.12)
+      rdf (>= 2.2, < 4.0)
     jsonapi-renderer (0.2.0)
     jwt (2.1.0)
     kaminari (1.1.1)
@@ -693,6 +697,7 @@ DEPENDENCIES
   idn-ruby
   iso-639
   json-ld (~> 2.2)
+  json-ld-preloaded (~> 2.2)
   kaminari (~> 1.1)
   letter_opener (~> 1.4)
   letter_opener_web (~> 1.3)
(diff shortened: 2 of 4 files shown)


ThibG:
@nightpool It doesn't go through the `load_jsonld_context` codepath for those for the affected JSON-LD contexts, so it should be ok.

nightpool (changes requested):
i want to make sure this gets tested on an affected instance (can just be locally after clearing redis) before merging.

ThibG:
It doesn't try to get the cache anymore when I send a toot, so I'm pretty confident it fixes it.

kit-ty-kate:
I'm testing it right now. Give my 30 minutes to rebuild the images and restart the services.

Gargron:
This is counter to #7301

nightpool:
@Gargron yes, because of urgent federation problems that some instances are experiencing.

ThibG:
@Gargron I would argue that not depending on HTTP requests to a potentially unreachable remote JSON-LD document that isn't supposed to change is probably worth a slight increase in initial RAM usage

Gargron:
OK, I've read the problem description. Thoughts:

Context is not neccessary for HTTP signatures, so why is federation broken when the context is unavailable, even though only linked data signatures should be broken? We should contain that error within that functionality, and make sure it doesn't break everything.

I have originally replaced the preloads with a cache mechanism because there is no guarantee that an incoming document will use any of the preloaded contexts, which would mean an arbitrary HTTP request. The preloads are a false sense of security in a way. Perhaps even more importantly, using cache instead of preloads allows even older instances to understand signatures from newer instances if the context has changed. However, I can't deny that currently things are bad because we don't have preloads as a fallback anymore.

ThibG:
@Gargron yes, we could decouple those two things (signing the payload, and sending a potentially unsigned payload), but I don't think there is much sense in that if we can make sure signing the payload is always ok.

My commit doesn't remove the loading and caching of unknown remote contexts. I don't think such “core” contexts are supposed to change, are they? Also, I think preloading them has performances advantages over re-parsing and re-loading them from cache.

nightpool:
@Gargron i agree, but we should get this fix out urgently and then work on making LDsigs fail-fast later.

@ThibG 

>I don't think such “core” contexts are supposed to change, are they?

contexts are assumed to be mutable by default, although i believe w3c does provide versioned activitystreams contexts now.

Gargron (approved):

nightpool (approved):

kit-ty-kate:
I confirm that the fixes works on my instance (sorry for the delay, the build failed halfway through)

Issue #9422, opened 2018-12-04
Add dates for notifications in Notification column


pawelngei:
Suggestion. Based on #675 , which makes it possible.

### Pitch

Add a visible date for each notification. Example for a toot posted 11th of November, shared 5h ago.

![notificationdate](https://user-images.githubusercontent.com/4021653/49438663-59128400-f7bf-11e8-9c5d-2e2009266ea7.png)

The date can be put on the same line as the notification title (pictured above), aligned right in case of longer lines:

![long_name_date](https://user-images.githubusercontent.com/4021653/49439121-83b10c80-f7c0-11e8-9832-f71c92d3c95f.png)

### Motivation

Right now there is no way to tell when an action happened based on the information in the Notifications column. It took me a few days to realize that the date displayed now is the date of the original post (boosted / favourited), not the action itself. I imagine it is similarly confusing for other newcomers.

ThibG:
Fixed by #9423

PR #9423, opened 2018-12-04
Add visible dates for notifications in Notification column

pawelngei:
UI. Implements #9422 .

Right now there is no way to tell when an action happened based on the information in the Notifications column. It took me a few days to realize that the date displayed now is the date of the original post (boosted / favourited), not the action itself. I imagine it is similarly confusing for other newcomers.

notification_dates_done

I would argue that the dates are really useful, but if you have a better idea _where_ to put them, I'm open.


Gargron:
This looks alright, we can give it a go

Gargron (approved):

PR #9427, opened 2018-12-05
Add profile directory

### LINKED ISSUE #5578: Profile hashtags
In profile descriptions, hashtags should link to page with list of other people using the same hashtag, that would be nice way to find interesting people. It seems to be easy to implement, but does it make sense?

* * * *

Gargron:
Fix #5578 

- Opt-in setting to be listed on a directory page ("Explore Mastodon"). Uses hashtags from profile bio
- Local-only for now, so each server will only list its own opted-in users
- Min. 10 followers required to appear on the directory to prevent spam
- Sorts by most recent toot by default
- Admin UI to see which hashtags are on the directory, and hide them if needed

Directory page:

image

Directory opt-in checkbox in "Edit profile":

image

Admin UI for hashtag moderation:

image

This should answer the question "how do I find worthwhile users from x without signing up and lurking there?" as well as "how do i find people interested in y?"

Toot asking for initial feedback on the idea: https://mastodon.social/@Gargron/101168486745315315

### DIFF
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Admin
+  class TagsController < BaseController
+    before_action :set_tags, only: :index
+    before_action :set_tag, except: :index
+    before_action :set_filter_params
+
+    def index
+      authorize :tag, :index?
+    end
+
+    def hide
+      authorize @tag, :hide?
+      @tag.account_tag_stat.update!(hidden: true)
+      redirect_to admin_tags_path(@filter_params)
+    end
+
+    def unhide
+      authorize @tag, :unhide?
+      @tag.account_tag_stat.update!(hidden: true)
+      redirect_to admin_tags_path(@filter_params)
+    end
+
+    private
+
+    def set_tags
+      @tags = Tag.discoverable
+      @tags.merge!(Tag.hidden) if filter_params[:hidden]
+    end
+
+    def set_tag
+      @tag = Tag.find(params[:id])
+    end
+
+    def set_filter_params
+      @filter_params = filter_params.to_hash.symbolize_keys
+    end
+
+    def filter_params
+      params.permit(:hidden)
+    end
+  end
+end
--- a/app/controllers/api/v1/accounts/credentials_controller.rb
+++ b/app/controllers/api/v1/accounts/credentials_controller.rb
@@ -21,7 +21,7 @@ class Api::V1::Accounts::CredentialsController < Api::BaseController
   private
 
   def account_params
-    params.permit(:display_name, :note, :avatar, :header, :locked, :bot, fields_attributes: [:name, :value])
+    params.permit(:display_name, :note, :avatar, :header, :locked, :bot, :discoverable, fields_attributes: [:name, :value])
   end
 
   def user_settings_params
--- a/app/controllers/directories_controller.rb
+++ b/app/controllers/directories_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class DirectoriesController < ApplicationController
+  layout 'public'
+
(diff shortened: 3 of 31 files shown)


ThibG:
Haven't actually read it, but there appears to be some leftover from your other PR regarding `deletion_schedules`.

Otherwise, I have two main worries:
- I wonder how useful that thing really is without federating. I fear it would only be useful for big instances and further encourage them.
- There were concerns about trending hashtags being used for harassment campaigns. Since this feature doesn't federate and you're planning to add ways for admins to blacklist hashtags, this seems like a lesser concern here, but I'd still ask around to get the people that were concerned by the trending hashtags feature before implementing this.

ykzts (approved):

ThisIsMissEm:
> First of all, the friends-of-friends, "who to follow" stuff is not something I want to go back into right now, so this feature is not intended to touch that topic at all. It's not about personalized results, but about finding people in general.

That's fair, I know personalization of results is a very contentious topic, and one that'll spark much debate. However, I think it's something people expect, especially populations from mainstream social networks. Is there an issue for this, or should we start one?

> Nor is it about getting a feel for what a server is like. I want to reduce the need to sign up on different servers to experience their community (since that is a side-effect of the local timeline feature anyway).

Do people really do this? To me, it seems that people pick an instance and roll with it, especially given the promotion of certain instances within certain communities (e.g., switter.at and sex workers on twitter)

I'm speaking only from personal experiences here, but I don't know anyone who's signed up to another instance just "to experience their community", as you can't experience a community without being part of it (to some degree) — I'm sure anthropologists have a phrase for this.

> It's about you knowing that cybre.space exists

How does the new user of Mastodon know that cybre.space exists in the first place? Even when viewing a profile from a remote instance, it's not particularly highlighted in anyway. (Other than the "this information may be incorrect" warning)

> and wanting to get some of the cybre.space vibes into your home feed on another server. How do you know who to follow? Here, look at this directory of cybre.space users who want to be found.

Would it perhaps make sense to surface the ability to add the local timeline of other instances to your own set of timelines? This seems like it'd probably be more useful than the full federated timeline of all instances.

I think we've still a lot of education to do with regards to instances and how they work for the mainstream internet population.

Gargron:
>Would it perhaps make sense to surface the ability to add the local timeline of other instances to your own set of timelines?

I know this is really desired by many but there's no good way to do it. All current Mastodon features relating to other servers use ActivityPub. Viewing another instance's local timeline would require using the Mastodon REST API, which would play weird with Pleroma, Misskey, and any potential future project. Third-party apps can afford to do this because they are Mastodon apps and their decisions are very localized. If the Mastodon server started to use its API for federation, that would be worse. That's the showstopper for that feature in my eyes...

ThisIsMissEm:
I'm not saying doing this via server implementation, just via the client; or perhaps just add a way to get a filtered federated timeline by instances? i.e., only federated toots are available via the instance timeline?

This is just an idea, but really we've first to address the education about instances & introduce people to the fact that not all people are on the same instance as you.

Cassolotl:
>>I want to reduce the need to sign up on different servers to experience their community (since that is a side-effect of the local timeline feature anyway).
>
>Do people really do this?

I have done this a lot!

hakaba-hitoyo:
I love this feature but against `MIN_FOLLOWERS_DISCOVERY = 10`. Spammers can easily and automatically register 10 users, but lonely newcomers rarely find their first 10 friends.

deutrino:
> That's fair, I know personalization of results is a very contentious topic, and one that'll spark much debate. However, I think it's something people expect, especially populations from mainstream social networks. Is there an issue for this, or should we start one?

https://github.com/tootsuite/mastodon/issues/4870#issuecomment-399818962 :)

ChameleonScales:
This is a great new feature, however it is very hidden for those who don't follow the release notes (and the future new comers). I suggest to have it turned on by default.

PR #9434, opened 2018-12-05
Add timeout to web push workers and stop retries when getting 4xx

Gargron:


ThibG:
1-3s is a bit on the low end, no? Why not something closer to the timeouts we use for basically everything else?

Gargron:
Because webpush endpoints are either Mozilla or Google or Apple, or an app developer, i.e. I can expect it to not be under heavy load or to be able to handle big loads. 1 second for a request is really long! It's also not that critical.

ThibG:
Hm, the instance issuing the notifications may itself have connectivity issues/be under load.

Are those tasks really causing delays? If so, those changes are probably stopping legitimate notifications from being delivered, right?

Issue #9343, opened 2018-11-25
HTML is not sanitized on profile edit preview


tribela:
### Expected behaviour

![image](https://user-images.githubusercontent.com/5047683/48976693-a52e3c00-f0ce-11e8-8afb-6ca077869ac1.png)

### Actual behaviour
![image](https://user-images.githubusercontent.com/5047683/48976680-71531680-f0ce-11e8-9751-5a1fee7b4de5.png)

### Steps to reproduce the problem

Set display name as `<script>alert(1)</script>a` and preview

After saving it, it shows normally, But only in preview

pawelngei:
I can implement a fix tomorrow.

nightpool:
I'm not seeing a difference between the two screenshots?

nightpool:
oh, it's in the profile card. pretty minor issue, but we'd be happy to accept a fix (should be as simple as using textContent instead of innerHtml)

pawelngei:
It's not just `textContent`, it seems to be caused by `emojify(escapeTextContentForBrowser(displayName), emojiMap)`, since it occurs only when typing. If the profile page loads with `<script>` in the variable, it's displayed as text.

pawelngei:
It's in Ruby, `_card.html.haml`. Now I'll learn some more Ruby and I'll fix this :P

pawelngei:
Okay, I can honestly say I know Ruby to little to fix it. More data:

`_card.html.haml` contains `%strong.emojify.p-name= display_name(account, custom_emojify: true)`

`stream_entries_helper.rb`:
```
  def display_name(account, **options)
    if options[:custom_emojify]
      Formatter.instance.format_display_name(account, options)
    else
      account.display_name.presence || account.username
    end
  end
```

and finally `formatter.rb`:
```
  def format_display_name(account, **options)
    html = encode(account.display_name.presence || account.username)
    html = encode_custom_emojis(html, account.emojis, options[:autoplay]) if options[:custom_emojify]
    html.html_safe # rubocop:disable Rails/OutputSafety
  end
```

I don't know which element causes HTML not to be sanitized when typed in dynamically and sanitized on initial page render.

nightpool:
again, it's not in the ruby. ruby runs on the server, JavaScript runs on
the client. the live preview discrepancy is solely caused by not escaping
things on the client side

pawelngei:
Okay, I communicated this badly, I'm sorry.

[...]

PR #9446, opened 2018-12-06
Escape HTML in profile name preview in profile settings

pawelngei:
Addresses #9343 . Additionally falls back to the default profile name if user deletes their custom one.

Right now the data about the default name is set in a `display: none` span, but it can be provided to JavaScript in some other way.

Personally I'd rewrite this view to React some day, but we should be good for now.

### DIFF
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -135,7 +135,7 @@ function main() {
     const name = document.querySelector('.card .display-name strong');
 
     if (name) {
-      name.innerHTML = emojify(target.value);
+      name.textContent = emojify(target.value);
     }
   });
 
--- a/app/javascript/packs/public.js
+++ b/app/javascript/packs/public.js
@@ -1,3 +1,4 @@
+import escapeTextContentForBrowser from 'escape-html';
 import loadPolyfills from '../mastodon/load_polyfills';
 import ready from '../mastodon/ready';
 import { start } from '../mastodon/common';
@@ -135,7 +136,7 @@ function main() {
     const name = document.querySelector('.card .display-name strong');
     if (name) {
       if (target.value) {
-        name.textContent = emojify(target.value);
+        name.innerHTML = emojify(escapeTextContentForBrowser(target.value));
       } else {
         name.textContent = document.querySelector('#default_account_display_name').textContent;
       }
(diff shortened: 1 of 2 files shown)


app/views/application/_card.html.haml
```diff
@@ -9,6 +9,7 @@
         = image_tag account.avatar.url, alt: '', width: 48, height: 48, class: 'u-photo'
 
       .display-name
+        %span{:id=>"default_account_display_name", :style=>"display:none;"}= display_name(account, custom_emojify: true)
```
Gargron: Use `id: ""` style, the `=>` style is outdated

Gargron (changes requested):

app/javascript/packs/public.js
```diff
@@ -133,9 +133,12 @@ function main() {
 
   delegate(document, '#account_display_name', 'input', ({ target }) => {
     const name = document.querySelector('.card .display-name strong');
-
     if (name) {
-      name.innerHTML = emojify(target.value);
+      if (target.value) {
+        name.textContent = emojify(target.value);
```
nightpool: textContent = emojify doesn't work, since emojify returns img tags for custom emoji

app/views/application/_card.html.haml
nightpool: since the default display name is always just the username, we should display that instead. this will give incorrect results for users trying to unset their display name

nightpool (dismissed):

app/javascript/packs/public.js
pawelngei: :+1:

pawelngei: Okay, I think there may be a wider bug, but even on `master` I cannot see the emojis in this field. I checked out https://writing.exchange/ and https://chaos.social/ and it doesn't work there, either.

pawelngei:
I've started looking at https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/features/emoji/emoji.js and there's something _seriously_ wrong with it. Does it work for anyone? I tested it on my local instance, chaos.social and writing.exchange and it doesn't emojify the profile.

From what I see the extremely convoluted `emojify` function never gets past the first `while`.

I'd like to refactor that function, but first I'd like to know if it works for anybody, anywhere..

pawelngei:
Okay, after several hours of debugging I have no idea how `emojify` works and how to test custom emojis. I welcome you to try fixing that and testing what kind of escaping works properly.

pawelngei:
Managed to handle the emojified unicode, but I have no idea if this will work for custom icons as well or how to test them.

Gargron (approved):

PR #9450, opened 2018-12-07
Add account hashtags to ActivityPub actor JSON

Gargron:
Follow-up to #9427

### DIFF
--- a/app/serializers/activitypub/actor_serializer.rb
+++ b/app/serializers/activitypub/actor_serializer.rb
@@ -105,7 +105,7 @@ class ActivityPub::ActorSerializer < ActiveModel::Serializer
   end
 
   def virtual_tags
-    object.emojis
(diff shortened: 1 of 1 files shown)


app/serializers/activitypub/actor_serializer.rb
```diff
@@ -119,6 +119,24 @@ def moved_to
...
+    include RoutingHelper
+
+    attributes :type, :href, :name
+
+    def type
+      'Hashtag'
+    end
+
+    def href
+      explore_hashtag_url(object)
```
ThibG: Not too sure statuses hashtags and account hashtags should be considered as two entirely different things?

ClearlyClaire (approved):
Not sure statuses hashtags and contents hashtags should be considered as two different things, but otherwise, LGTM

PR #9470, opened 2018-12-08
Move #set_user to Admin::BaseController

ysksn:


ThibG:
The only issue I have with this is that `app/controllers/admin/two_factor_authentications_controller.rb` features a different `set_user`, so that may be a bit confusing.

ysksn:
How about renaming `Admin::TwoFactorAuthenticationsController#set_user` to `#set_target_user`, `#set_disabled_user` or `#set_user_will_be_disabled` ?

ThibG:
`set_target_user` would be fine I guess

ClearlyClaire (approved):

Issue #9351, opened 2018-11-26
Custom emojis - Domain is case sensitive


csmall:
The domain filter for the Custom emojis admin screen is case sensitive, it should be case insensitive.

Administration ➡️ Custom Emojis
In "Domain" box type the domain
Click "Search" button

Mastodon.social shows 0 hits
mastodon.social shows many emojis

Esteth:
Fixed as of 7d00e4e

PR #9474, opened 2018-12-08
Make custom emoji domains case insensitive #9351

Esteth:
Make custom emoji domains case insensitive to fix #9531

Postgres supports CITEXT, but migrating the database from TEXT to CITEXT could require a whole table lock or an interim migration with dual writes.

Instead, it should be fine to migrate existing rows to lowercase one-by-one, input only lowercase input in future, and do comparisons against lowercase input.

This is my first pull request for mastodon and I haven't worked in Ruby in a long time - please let me know if I'm missing something!

### DIFF
--- a/db/migrate/20181207011115_downcase_custom_emoji_domains.rb
+++ b/db/migrate/20181207011115_downcase_custom_emoji_domains.rb
@@ -1,5 +1,7 @@
 class DowncaseCustomEmojiDomains < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
   def change
-    CustomEmoji.update_all('domain = lower(domain)')
+    CustomEmoji.in_batches.update_all('domain = lower(domain)')
   end
 end
--- a/db/migrate/20181207011115_downcase_custom_emoji_domains.rb
+++ b/db/migrate/20181207011115_downcase_custom_emoji_domains.rb
@@ -1,0 +1,5 @@
+class DowncaseCustomEmojiDomains < ActiveRecord::Migration[5.2]
+  def change
+    CustomEmoji.update_all('domain = lower(domain)')
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_12_04_215309) do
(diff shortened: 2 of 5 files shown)


db/migrate/20181207011115_downcase_custom_emoji_domains.rb
```diff
@@ -0,0 +1,5 @@
+class DowncaseCustomEmojiDomains < ActiveRecord::Migration[5.2]
+  def change
+    CustomEmoji.update_all('domain = lower(domain)')
```
Gargron: Use `in_batches.update_all` to prevent table lock ups

db/migrate/20181207011115_downcase_custom_emoji_domains.rb
```diff
@@ -0,0 +1,5 @@
+class DowncaseCustomEmojiDomains < ActiveRecord::Migration[5.2]
```
Gargron: Add `disable_ddl_transaction!` to use no locking around the updates.

db/schema.rb
```diff
@@ -336,9 +336,9 @@
   create_table "mutes", force: :cascade do |t|
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "hide_notifications", default: true, null: false
```
Gargron: Not sure what this is about, the version string update is OK but this seems unnecessary, you can manually switch it around.

Gargron (changes requested):

db/migrate/20181207011115_downcase_custom_emoji_domains.rb
Esteth: Thanks, I'm learning a lot about ActiveRecord migrations :)

db/migrate/20181207011115_downcase_custom_emoji_domains.rb
Esteth: Awesome, I didn't know this existed!

db/schema.rb
Esteth: Oops, this is the result of a bad merge. My bad for not seeing it!

Reverted

Esteth (commented):
Thanks Gargron - I've forgotten so much about rails!

Gargron (approved):

PR #9480, opened 2018-12-09
Add audio uploads

### LINKED ISSUE #4827: Allow users to post audio files
It is currently possible to attach up to four pictures or one video to a toot. However, it's not possible to add an audio file to a post.

I suggest we add the possibility for users to attach audio files to their status.

* * * *

Gargron:
Fix #4827

image

- [x] Server-side changes to accept audio files
- [x] Audio player component

### DIFF
--- a/app/javascript/mastodon/features/status/components/detailed_status.js
+++ b/app/javascript/mastodon/features/status/components/detailed_status.js
@@ -11,6 +11,7 @@ import { FormattedDate, FormattedNumber } from 'react-intl';
 import Card from './card';
 import ImmutablePureComponent from 'react-immutable-pure-component';
 import Video from '../../video';
+import Audio from '../../audio';
 
 export default class DetailedStatus extends ImmutablePureComponent {
 
@@ -53,6 +54,12 @@ export default class DetailedStatus extends ImmutablePureComponent {
     if (status.get('media_attachments').size > 0) {
       if (status.get('media_attachments').some(item => item.get('type') === 'unknown')) {
         media = <AttachmentList media={status.get('media_attachments')} />;
+      } else if (status.getIn(['media_attachments', 0, 'type']) === 'audio') {
+        const audio = status.getIn(['media_attachments', 0]);
+
+        media = (
+          <Audio src={audio.get('url')} duration={audio.getIn(['meta', 'original', 'duration'])} />
+        );
       } else if (status.getIn(['media_attachments', 0, 'type']) === 'video') {
         const video = status.getIn(['media_attachments', 0]);
 
--- a/app/models/media_attachment.rb
+++ b/app/models/media_attachment.rb
@@ -22,14 +22,16 @@
 class MediaAttachment < ApplicationRecord
   self.inheritance_column = nil
 
-  enum type: [:image, :gifv, :video, :unknown]
+  enum type: %i(image gifv video unknown audio)
 
-  IMAGE_FILE_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif'].freeze
-  VIDEO_FILE_EXTENSIONS = ['.webm', '.mp4', '.m4v', '.mov'].freeze
+  IMAGE_FILE_EXTENSIONS = %w(.jpg .jpeg .png .gif).freeze
+  VIDEO_FILE_EXTENSIONS = %w(.webm .mp4 .m4v .mov).freeze
+  AUDIO_FILE_EXTENSIONS = %w(.ogg .oga .mp3).freeze
 
-  IMAGE_MIME_TYPES             = ['image/jpeg', 'image/png', 'image/gif'].freeze
-  VIDEO_MIME_TYPES             = ['video/webm', 'video/mp4', 'video/quicktime'].freeze
-  VIDEO_CONVERTIBLE_MIME_TYPES = ['video/webm', 'video/quicktime'].freeze
+  IMAGE_MIME_TYPES             = %w(image/jpeg image/png image/gif).freeze
+  VIDEO_MIME_TYPES             = %w(video/webm video/mp4 video/quicktime).freeze
+  VIDEO_CONVERTIBLE_MIME_TYPES = %w(video/webm video/quicktime).freeze
+  AUDIO_MIME_TYPES             = %w(audio/ogg video/ogg audio/mpeg).freeze
 
   IMAGE_STYLES = {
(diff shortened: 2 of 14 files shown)


app/javascript/mastodon/features/status/components/detailed_status.js
```diff
@@ -53,7 +53,7 @@ export default class DetailedStatus extends ImmutablePureComponent {
     if (status.get('media_attachments').size > 0) {
       if (status.get('media_attachments').some(item => item.get('type') === 'unknown')) {
         media = <AttachmentList media={status.get('media_attachments')} />;
-      } else if (status.getIn(['media_attachments', 0, 'type']) === 'video') {
+      } else if (['audio', 'video'].indexOf(status.getIn(['media_attachments', 0, 'type'])) !== -1) {
```
sphinxc0re: I'm not actually part of the project, but isn't there `Array.prototype.includes()`? 😁
Here is some documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

Alamantus: `Array.includes()` is only in ES2015+ JS standard, so unless Mastodon is using Babel or some such transposer, it won't work for many browsers. That said, for all I know, it could already be using Babel—I haven't checked the source like I should have...

app/models/media_attachment.rb
```diff
@@ -22,14 +22,16 @@
...
   self.inheritance_column = nil
 
-  enum type: [:image, :gifv, :video, :unknown]
+  enum type: %i(image gifv video unknown audio)
 
-  IMAGE_FILE_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif'].freeze
-  VIDEO_FILE_EXTENSIONS = ['.webm', '.mp4', '.m4v', '.mov'].freeze
+  IMAGE_FILE_EXTENSIONS = %w(.jpg .jpeg .png .gif).freeze
+  VIDEO_FILE_EXTENSIONS = %w(.webm .mp4 .m4v .mov).freeze
+  AUDIO_FILE_EXTENSIONS = %w(.ogg .oga .mp3).freeze
```
koyuawsmbrtn: What about adding support for FLAC and WAV files?

trwnh: Also .opus would be appreciated!

Deiru2k: https://caniuse.com/#feat=flac
https://caniuse.com/#search=wav

Can't see why not.

Gargron: See, this is the can of worms I wanted to avoid by not messing with audio. FLAC and WAV are fucking huge, and converting them takes precious CPU cycles. Makes me question whether this whole PR is worth it.

ThibG: To be honest, we already convert videos, which is probably way more CPU-intensive than converting audio files.

koyuawsmbrtn: I would say the people who are uploading audio files already know how to convert them. Anything else will mess up the code even more. This feature is nice and I would really like to get it merged. Everyone else who needs FLACs, WAVs or OPUS can upload theirs to Pleroma or whatever. That's why these things exist.

Gargron: @ThibG For reference we convert those because we have no choice in regards to mobile devices. But no urgent need exists for audio.

ykzts (approved):

koyuawsmbrtn:
I would suggest to merge this, because we can implement support for other file types later :p

MirceaKitsune:
:+1: for this! I was thinking of opening an issue about this exact limitation. Very glad a PR is in place and seemingly ready, hope this will be in soon.

aka-rin:
Can't believe to see this! Please merge it!

I agree with koyuawsmbrtn's comment. Although mp3 is dead, it's still the most popular format of audio files with the best compatibility. Auto-transcoded to OPUS can be set as an optional function for admins. More formats can be supported in the future.

I'm just curious about the possibility of creating podcasts in Mastodon:

1. What is the exact file limitation of audio files? If the size is too small, it may be hard for podcasts with a medium duration. Maybe 20 minutes is enough?
2. Will Mastodon provide an [RSS/Atom interface for podcasts](https://github.com/gpodder/podcast-feed-best-practice/blob/master/podcast-feed-best-practice.md)? Maybe like lists or favorite toots, users can create a channel with a collection of toots and share it to others.

AirDur:
Why was it closed ? It is a good and requested feature.

koyuawsmbrtn:
@AirDur what I was told is that Eugen wasn't quite happy with the UI and functions this PR provided. I can understand this for sure.

Sylvhem:
@koyuawsmbrtn We still have the broken DM interface nobody asked for and the butchered lists that were requested by three people. So I think implementing audio supports in a less than perfect way would have been fine.

MirceaKitsune:
I thought this was merged at first. I'm saddened it was not, and hope the team will come back to this and merge it before Mastodon 2.8.0 is released.

koyuawsmbrtn:
@Sylvhem I can agree, the DM user interface is a piece of garbage. It's really hard to keep track e.g. who you were talking to the night before. I might open up a new issue for that one if it doesn't exist.

Gargron:
@koyuawsmbrtn Are you on the 2.8 release candidate? I changed the DMs to always display other participants instead of your own name, regardless of who the last message belongs to. I'm not 100% happy with that interface yet and I have to admit Twitter's is smoother, but I can't relate to the old list of every direct message from/to you in a single stream being easier to find something in. The new interface involves less scrolling...

koyuawsmbrtn:
@Gargron I'm still on 2.7.4, but I can try it out

novadeviator:
Many musicians and sound artists would be grateful for audio upload support. I read about all the difficulties and reservations about it and I can understand them, but there's IS an urgent need for it as much as it is for video and image support. There are number of communities around music-making and audio and they need this feature (smusi.ch, metalhead.club, sonomu.club), folks doing podcasts, and blind people as well. Please make this feature happen - also, support OPUS and FLAC which are the open standards that are here to stay.

jaredcwhite:
Darn, I was really hoping to launch a themed instance for a particular music genre soon. Thought this was almost ready to roll out! Does anyone think there's enough solid code for me to create a private fork and just use this for one instance?

PR #9489, opened 2018-12-10
Improve e-mail MX validator and add tests

Gargron:
The problem with the old validator code wasn't so much that it had a bug, but the usage was a bit misleading for admins. MX records are hostnames, so if you blacklisted an IP that a MX record pointed to, but not the MX hostname itself, it would still let accounts through.

Now the validator fetches MX records, and then A records for both the original domain as well as MX records, and then checks both the hostnames and the IPs against the blacklist, so if you have blacklisted either of them, no accounts would be let through.

A future improvement may be offering the admins a choice to automatically create blacklist records for the IPs when they enter a domain name.


renatolond (approved):
Lgtm

Issue #8529, opened 2018-08-30
Open toot when clicking on the blank zone below profile picture


roipoussiere:
When a user posts a toot containing only a picture, it's annoying to open the toot, because when you click on it, it just opens the picture.

![1](https://user-images.githubusercontent.com/1665542/44851009-45ef9580-ac5f-11e8-9c6e-a679d081993b.png)
*User: hmm, where I'm supposed to click to open this toot?*

Of course you can open it by clicking on the three-dot button, then open toot. But many users don't know this trick and it's two clicks instead one: maybe we can improve the usability for this scenario.

My suggestion is to open the toot when the user clicks on the blank zone below the profile picture:

![2](https://user-images.githubusercontent.com/1665542/44851053-64559100-ac5f-11e8-9f73-012e5a4bc977.png)

In this way it's still possible to open any toot, with or without text, with or without image.

![3](https://user-images.githubusercontent.com/1665542/44851273-0d9c8700-ac60-11e8-88a4-13822cc6dd9e.png)
![4](https://user-images.githubusercontent.com/1665542/44851275-0f664a80-ac60-11e8-8b79-f1590d35706a.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

zcdunn:
This would especially be an improvement on mobile, where it's nearly impossible to open some posts with one tap.

Cassolotl:
This is a duplicate of *image-only toots don't have any space to click on to load them in the 4th column #564*, which was closed. It's also related to *On toots with only URLs, there is nowhere in the toot area to click to expand the toot #4314*, which I opened because I didn't like that the first one was closed!

roipoussiere:
Ok, thanks, I didn't find these issues.

But here I propose a concrete suggestion.

Down10:
I would note that the post date (in the upper right corner) is a reliable place to open the toot's permalink page. Maybe make the hitbox for that link a little larger?

Cassolotl:
That opens the toot in a new tab, which you can't interact with, rather than expanding it in the last column

trwnh:
One problem: the current "blue area" you highlighted is for selecting and focusing a certain toot already. If clicking in that area is made to open the toot instead, then navigating the column by keyboard shortcuts would become a lot harder, as you would have to click Tab 23 different times just to select the top toot in the Home column, and then who knows how many times down to reach some arbitrary position.

yaomtc:
[...]

PR #9492, opened 2018-12-11
Make clicks on the status action bar expand toots

Dar13:
Possible fix for #8529, this makes clicks on the status action bar (the area underneath toots where the various buttons are) expand the current toot. Leaves the larger area under the profile picture alone, to keep previous behavior that lets users focus the toot without expanding it. Any suggestions are welcome to cover other use-cases.

The image has the new clickable zones on the status action bar.
clickable_zones_changed

### DIFF
--- a/app/javascript/mastodon/components/status_action_bar.js
+++ b/app/javascript/mastodon/components/status_action_bar.js
@@ -248,7 +248,7 @@ class StatusActionBar extends ImmutablePureComponent {
     );
 
     return (
-      <div className='status__action-bar'>
+      <div className='status__action-bar' role='presentation' onClick={this.handleOpen}>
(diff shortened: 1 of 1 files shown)


Dar13:
Seems the linter isn't like that I made the `div` interactive. Guess I could try using an `area` element instead? I'm kind of out of my normal area of programming, not sure what the path forward would be.

Gargron:
>Seems the linter isn't like that I made the div interactive

FWIW the linter is wrong here because there's so many other ways to open a status, if this element isn't accessible to screen readers it shouldn't be a problem

crockwave:
This should be a good solution.  Following is what we tell our users on how to expand toots with only a link or image and no comment:

On toots with only a link or image, to expand the toot, click on the ellipses and select  "Expand this status" to allow you to see the replies.  On toots with just a link and no comment, it is technically possible to hover just to the right of the link ellipses and click to expand the status.

Dar13:
Resolved the merge conflicts, are there objections to this PR being merged? Pinging @Gargron

Dar13:
Obsoleted with the merging of https://github.com/tootsuite/mastodon/pull/10766

PR #9494, opened 2018-12-11
Restore profile links in /admin/accounts

### LINKED ISSUE #9431: Provide quick access to users' public profiles or web view from admin panel
Since 73faadad28b897cd61cc1d6c8bee47e7d72a51a1, there is no link to listed user's profile from the public panel.

The admin view of a user (`/admin/accounts/id`) is not a complete replacement as it is missing at least:
- their header image
- their bio & fields
- an easy way to contact them

ClearlyClaire:
Fixes #9431 

Before
--------

screenshot_2018-12-11 accounts - dev instance

After
------

screenshot_2018-12-11 accounts - dev instance 1

### DIFF
--- a/app/views/admin/accounts/index.html.haml
+++ b/app/views/admin/accounts/index.html.haml
@@ -41,6 +41,7 @@
         %th= t('admin.accounts.role')
         %th= t('admin.accounts.most_recent_ip')
         %th= t('admin.accounts.most_recent_activity')
+        %th
     %tbody
       = render @accounts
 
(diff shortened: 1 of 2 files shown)


ykzts (approved):

app/views/admin/accounts/index.html.haml
```diff
@@ -41,6 +41,7 @@
         %th= t('admin.accounts.role')
         %th= t('admin.accounts.most_recent_ip')
         %th= t('admin.accounts.most_recent_activity')
+        %th= t('admin.accounts.profile_links')
```
Gargron: Would rather avoid adding a string for this

Gargron (changes requested):

Gargron (approved):

PR #9497, opened 2018-12-11
Add instance-wide setting to disable profile directory

### LINKED ISSUE #9496: Make profile directory optional
### Pitch

Instance admins should be able to disable the profile directory feature from site settings.

### Motivation

This feature uselessly clutters the interface for single-user or instances with few users.

Furthermore, it introduces more avenues for spammers, which may be more work for moderators.

ClearlyClaire:
Fixes #9496

When the profile directory is disabled:
- The “discoverable” setting is hidden from users
- The “profile directory” link is not shown on public pages
- /explore returns 404

### DIFF
--- a/app/controllers/directories_controller.rb
+++ b/app/controllers/directories_controller.rb
@@ -3,22 +3,25 @@
 class DirectoriesController < ApplicationController
   layout 'public'
 
+  before_action :check_enabled
   before_action :set_instance_presenter
   before_action :set_tag, only: :show
   before_action :set_tags
   before_action :set_accounts
 
   def index
-    return not_found unless Setting.profile_directory
     render :index
   end
 
   def show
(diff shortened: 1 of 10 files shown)


app/controllers/directories_controller.rb
```diff
@@ -9,10 +9,12 @@ class DirectoriesController < ApplicationController
   before_action :set_accounts
 
   def index
+    return not_found unless Setting.profile_directory
```
Gargron: This could be a before_action

Gargron (approved):

33b5e5:
Might I suggest also disabling the directory when `SINGLE_USER_MODE=true` in `.env.production`. There is no point in a profile directory when you are running a single user instance.

PR #9510, opened 2018-12-12
 Ignore low-confidence CharlockHolmes guesses when parsing link cards

### LINKED ISSUE #9466: PreviewCard does not respect charset
### Expected behaviour

Cards should parse the `Content-Type` HTTP header and `<meta http-equiv="Content-Type">` element and use the specified charset to parse the `<title>`.

### Actual behaviour

The wrong charset is used and the link preview is garbage.

Link to https://playground.jomo.tv/windows-1251.html

Content of that site, encoded in windows-1251 (pipe it to `iconv -f utf8 -t windows-1251`):

```html
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=windows-1251" />
  <title>сэмпл текст</title>
</head>
<body>
  <p>сэмпл текст</p>
</body>
</html>
```

### Specifications

Tested on a Mastodon v2.6.5 instance

ClearlyClaire:
Fixes #9466 by ignoring low-confidence `CharlockHolmes` charset guesses.

High-confidence guesses will still override server or document-specified charset, and low confidence guesses won't be used at all even if there is no server or document-specified charset, so this might not be the best way to proceed.

Note also that the “hint” given to `CharlockHolmes` as very little importance, and feeding it the correct charset may not prevent it from outputing an incorrect, low-confidence guess.

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -137,7 +137,7 @@ class FetchLinkCardService < BaseService
     detector.strip_tags = true
 
     guess      = detector.detect(@html, @html_charset)
-    encoding   = guess&.fetch(:confidence, 0) > 60 ? guess&.fetch(:encoding, nil) : nil
+    encoding   = guess&.fetch(:confidence, 0).to_i > 60 ? guess&.fetch(:encoding, nil) : nil
     page       = Nokogiri::HTML(@html, nil, encoding)
     player_url = meta_property(page, 'twitter:player')
 
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -137,7 +137,8 @@ class FetchLinkCardService < BaseService
     detector.strip_tags = true
 
     guess      = detector.detect(@html, @html_charset)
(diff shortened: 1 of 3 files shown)


ThibG:
Weird, the tests pass in my dev environment :thinking:

app/services/fetch_link_card_service.rb
```diff
@@ -137,7 +137,8 @@ def attempt_opengraph
     detector.strip_tags = true
 
     guess      = detector.detect(@html, @html_charset)
-    page       = Nokogiri::HTML(@html, nil, guess&.fetch(:encoding, nil))
+    encoding   = guess&.fetch(:confidence, 0) > 60 ? guess&.fetch(:encoding, nil) : nil
```
Gargron: I think `nil > 60` will throw no method error

Gargron (changes requested):

app/services/fetch_link_card_service.rb
ThibG: Indeed! Addressed.

Gargron (approved):

jomo:
Thank you, @ThibG!

PR #9515, opened 2018-12-13
Make avatar and headers respect `auto_play_gif`

ClearlyClaire:
Also fix avatar URLs in some public pages when data is stored on a different host

### DIFF
--- a/app/views/accounts/_header.html.haml
+++ b/app/views/accounts/_header.html.haml
@@ -1,15 +1,9 @@
 .public-account-header{:class => ("inactive" if account.moved?)}
   .public-account-header__image
-    - if current_account&.user&.setting_auto_play_gif
-      = image_tag account.header.url, class: 'parallax'
-    - else
-      = image_tag account.header_static_url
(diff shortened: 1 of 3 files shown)


app/views/accounts/_header.html.haml
```diff
@@ -1,6 +1,9 @@
 .public-account-header{:class => ("inactive" if account.moved?)}
   .public-account-header__image
-    = image_tag account.header.url, class: 'parallax'
+    - if current_account&.user&.setting_auto_play_gif
```
Gargron: Why using blocks here if you just need to switch the URL. Ternary operator should do

Gargron (changes requested):

app/views/accounts/_header.html.haml
ThibG: For consistency with the code for avatars that is just a few lines below:
https://github.com/tootsuite/mastodon/pull/9515/files#diff-282ac2735bdc933e88272b729a2faf8bL6

Gargron: OK but that code is bad too. I think ideally it would be extracted to a helper method altogether.

Gargron (approved):

PR #9519, opened 2018-12-14
Add moderation warnings

Gargron:
Moderation warnings are meant to formalize the process of warning users about their misconduct prior to harsher measures, so that other staff can track that process. At the same time, the warning system is meant to solve the issue of absolute intransparency about moderation actions concerning your account, i.e. if you are banned, you won't know when and why you were banned.

However, sometimes moderation actions *need* to be covert, and as such, you can opt to generate no warnings/e-mails when perfming said actions.

In terms of moderation UI, whereas previously, "disable" and "silence" were one-click actions, now they are unified into a form that allows you to enter warning text (or choose from custom presets to save time). That form likewise oversees the "suspend" action, which previously was on a separate page that required you to enter the username of the account for confirmation (that is no longer so).

That new form is not only unified between various actions, but also between actions that are part of a report, and actions taken directly from the account page. Of course, warnings can only be sent to local users.

___

Action UI:

image

E-mail the user receives:

image

Warning presets UI:

image

___

Here are the e-mail contents for various action types:

|Action|Subject|Message (above optional custom text)|
|------|-------|------------------------|
|none| Warning for %{acct}||
|disable|Your account %{acct} has been frozen|While your account is frozen, your account data remains intact, but you cannot perform any actions until it is unlocked.|
|silence|Your account %{acct} has been limited|While your account is limited, only people who are already following you will see your toots on this server, and you may be excluded from various public listings. However, others may still manually follow you.|
|suspend|Your account %{acct} has been suspended|Your account has been suspended, and all of your toots and your uploaded media files have been irreversibly removed from this server, and servers where you had followers.|

Below this explanation, you can enter custom text, or choose one of the preset options. You can add presets.

The e-mail also contains a link to "review server policies" at /about/more


Packbat:
I think it would make sense to include account and instance in the subject line and message. Tentative suggestion:

Action | Subject | Message (above optional custom text)
-- | -- | --
none | Warning for [account]@[instance] |  
disable | Your account [account]@[instace] has been disabled | You have been locked out of your account [account]@[instance]. Your login sessions and  app tokens are still intact, but you cannot perform any actions until  your account is unlocked.
silence | Your account [account]@[instance] has been quarantined | While your account [account]@[instance] is quarantined, only people who are already following you will see your toots on this server.
suspend | Your account [account]@[instance] has been deleted | Your account [account]@[instance] and all of your toots and media attachments have been removed.

joyeusenoelle:
Suggest additional text after the custom text, along the lines of:

*disable*: "Please contact <admin email address> to discuss re-enabling your login."
*silence*: "Please contact <admin email address> to discuss removing the limit on your account."
*suspend*: "Your account cannot be recovered. If you return to Mastodon, please choose another server here: https://joinmastodon.org ."

Kjwon15:
What's mean "none"?
Will send email at just reported? or send at mark as solved?

Gargron:
None means no action against the account, allows you to just send a warning e-mail. Sending a warning with no action when coming from a report will mark that particular report as closed.

Kjwon15:
I suggest to send this kind of emails to the reporters.
When I using Twitter's one, It was very not satisfied which not giving any feedback for reportings.

app/controllers/admin/reports_controller.rb
```diff
@@ -13,75 +13,42 @@ def show
       authorize @report, :show?
 
       @report_note  = @report.notes.new
-      @report_notes = (@report.notes.latest + @report.history).sort_by(&:created_at)
+      @report_notes = (@report.notes.latest + @report.history + @report.target_account.targeted_account_warnings.latest.custom).sort_by(&:created_at)
```
krainboltgreene: Might be time to create a view.

app/mailers/user_mailer.rb
```diff
@@ -78,4 +78,16 @@ def backup_ready(user, backup)
...
   end
+
+  def warning(user, warning)
+    @resource = user
+    @warning  = warning
+    @instance = Rails.configuration.x.local_domain
+
+    I18n.with_locale(@resource.locale || I18n.default_locale) do
+      mail to: @resource.email,
+           subject: I18n.t("user_mailer.warning.subject.#{@warning.action}", acct: "#{user.account.username}@#{@instance}"),
```
krainboltgreene:
```
"#{user.account.username}@#{@instance}"
```

Kinda surprised this isn't a function?

app/models/account_warning.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class AccountWarning < ApplicationRecord
+  enum action: %i(none disable silence suspend), _suffix: :action
+
+  belongs_to :account, inverse_of: :account_warnings
+  belongs_to :target_account, class_name: 'Account', inverse_of: :targeted_account_warnings
+
+  scope :latest, -> { order(created_at: :desc) }
+  scope :custom, -> { where.not(text: '') }
+end
```
krainboltgreene: Validate the enum and create a constant for future reference?

krainboltgreene (approved):

app/mailers/user_mailer.rb
Gargron: You are right, I forgot about `Account#local_username_and_domain`

app/models/account_warning.rb
Gargron: Validate the enum how? I believe the enum can't be set to a value that isn't defined

app/controllers/admin/reports_controller.rb
Gargron: In most cases this is only 4-5 items at most in total. That being said, how do I query a view in Rails?

Cassolotl:
Is "Notify the user per email" the same meaning as "Notify the user by email"? I've not seen it written that way before.

danhunsaker:
I tend to see/use "via" rather than "by", but agree that "per" seems unusual, and may in fact have a different meaning than intended. ("Per" generally means "in accordance with".)

Gargron:
It should be "by", my mistake

ThibG:
Nothing in the admin accounts UI explains that those records are warnings.
screenshot_2018-12-23 tmi - mastodon instance perso de thibg

PR #9534, opened 2018-12-16
Do not ignore federated reports targetting already-reported accounts

ClearlyClaire:


ThibG:
So, currently, only the first remote report from a given account targeting a local account is processed.

This means a high-quality report sent after a low-quality one will be lost. It also means that a repeat offender will only get reports from their first offense (indeed, new reports will be ignored if there is an existing report, *resolved or not*).

This is made worse by the fact that all reports from a same instance will typically be sent from the same account.

nightpool (approved):
Thibg's reasoning is sound here, I would like to see this merged. I didn't even realize that the reports I was seeing might be incomplete or not reflect the actual quantity of reports. For example, if someone gets reported for not marking their media as NSFW, and that issue is resolved, then I would never see a report when 6 months later they decide to go on a racist tirade and i need to ban them. This is super exploitable by bad actors.

PR #9537, opened 2018-12-16
Remove PostgreSQL statement timeout

Gargron:
Revert #9382


ClearlyClaire (approved):
Slow requests can happen under load, or when running migration scripts or similar long-running queries. The main reason for introducing this in the first place was to work around deadlocks which should be fixed now. Already rolled back in glitch-soc.

PR #9549, opened 2018-12-17
Default to a non-animated version of oops.gif

### LINKED ISSUE #6060: Request: could the 500 page become non-blinking?

ClearlyClaire:
Fixes #6060 

The new png is the first frame from the animated gif, with the orange bits removed since they only make sense in the animation.

screenshot_2018-12-17 this page is not correct - dev glitch social


ashleyhull-versent:
image

ashleyhull-versent:
RIP.

koyuawsmbrtn:
What about trying to get the value from the settings saying "allow autoplay" and then animate it if the user wishes so? It's also disabled by default so it shouldn't bother anyone.

ThibG:
I guess we could, for the non-static part. But most of the time, such pages will show up because the server itself is down, and that lookup cannot be done in such occasions.

ThibG:
Updated the PR to respect the autoPlayGif setting when it's available (and default to the still image otherwise)

koyuawsmbrtn (approved):
Looks pretty good, I don't know if it's ready to merge, but I would approve it

renatolond (approved):

danhunsaker (approved):

Kjwon15:
Noooo.... I loved that smashing friend

PR #9552, opened 2018-12-17
Sanitize and sandbox toot embeds

ClearlyClaire:
Up until now, we blindly included oEmbed replies from remote toots in the embed modal.
Thankfully, since we introduced a Content Security Policy, this is not a security issue in Mastodon itself.

It however causes Content Security Policy warnings, and we may still provide users with malicious instructions.

The proposed change sanitizes the oEmbed content before suggesting it to the user, and also sandboxes the code to prevent executing scripts in a same-origin context.

This does cause some functionality loss: the generated code for embedding remote toots will not include the JS snippet needed to adjust the iframe's height.

### DIFF
--- a/app/controllers/api/web/embeds_controller.rb
+++ b/app/controllers/api/web/embeds_controller.rb
@@ -10,6 +10,7 @@ class Api::Web::EmbedsController < Api::Web::BaseController
     render json: status, serializer: OEmbedSerializer, width: 400
   rescue ActiveRecord::RecordNotFound
     oembed = FetchOEmbedService.new.call(params[:url])
+    oembed[:html] = Formatter.instance.sanitize(oembed[:html], Sanitize::Config::MASTODON_OEMBED) if oembed[:html].present?
 
     if oembed
       render json: oembed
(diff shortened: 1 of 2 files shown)


ThibG:
Another possibility, to avoid presenting the user untrusted code, is to accept serving remote toots with the embed controller, but it would require more changes.

app/controllers/api/web/embeds_controller.rb
```diff
@@ -10,6 +10,7 @@ def create
     render json: status, serializer: OEmbedSerializer, width: 400
   rescue ActiveRecord::RecordNotFound
     oembed = FetchOEmbedService.new.call(params[:url])
+    oembed[:html] = Formatter.instance.sanitize(oembed[:html], Sanitize::Config::MASTODON_OEMBED) if oembed[:html].present?
```
Gargron: This *will* break Mastodon embeds because they need a `<script>` for iframe resizing, and this ruleset strips out script tags.

ThibG: As I said:

> This does cause some functionality loss: the generated code for embedding remote toots will not include the JS snippet needed to adjust the iframe's height.

But I prefer that to instructing users to include untrusted code (note that embeds from your own instance don't go through that codepath and the javascript tag is still included).

ThibG: What I mean is, it's okayish to instruct people to include scripts that we control, but I don't think it's ok at all to instruct them to include scripts from a remote instance we don't control.

Gargron (approved):

PR #9553, opened 2018-12-17
Makes drawer sticky when horizontal scroll

ronane:
This allows the drawer to be sticky and stay on top of other columns when user scrolls horizontally

image

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -1925,6 +1925,10 @@ a.account__display-name {
   display: flex;
   flex-direction: column;
   overflow-y: hidden;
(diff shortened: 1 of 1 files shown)


jeroenpraat:
I would suggest to make this optional with some kind of non-intrusive pin icon.

hook54321:
> I would suggest to make this optional with some kind of non-intrusive pin icon.

or through something in settings, which would probably be the least intrusive.

ronane:
Well, at last i've some reviews, advices, thanks. ^^

I wanted to make this CSS only to have the least impact, but yeah, a checkbox in settings could be great.
Though i've no idea how to do this :\

Any help is welcomed!

PR #9560, opened 2018-12-18
Add frontend validation when registration

k-nasa:
## background issue
https://github.com/tootsuite/mastodon/issues/9346

I looked at this issue and added validation on the registration form.
I'm a beginner of React. So I'm sorry if there is a funny code

I will do another PR verification of user name duplication etc


m4sk1n:
I think using property names in messages isn’t best UX practice…

nightpool (changes requested):
`admin@localhost:3000` is a valid email address, your validation code is broken.

I don't know if we should be using another react view here—we try to keep react code to an absolute minimum where possible. Down to be convinced either way on this one though

I don't think we should be adding a dependency for something as simple as "has more then 8 characters"

k-nasa:
@Gargron I have only one question.
What do you think is a smart way to do reactive validation at the front end?

danhunsaker:
As was mentioned above, HTML5 supports regex-based validation of field values without touching JavaScript. Use that for reactive validation.

As to password lengths, the regex `/.{8,}/` would ensure 8 characters or more. Emails shouldn't be validated as anything more than `/.+@.+/` - one or more characters, an @, and one or more characters. Username max of 30 characters? `/.{1,30}/`. Easy as that.

k-nasa:
@danhunsaker Thank you for teaching carefully!

nightpool:
@danhunsaker if my email is nightpool@cybre.space and I want to sign up for an account on cybre.space, it *should* be valid to just put "nightpool" as my email.

(i'm unsure if this actually works in practice for most common configurations :P, but it's a hole in "emails should be validated as `/.+@.+/`")

danhunsaker:
This sort of "implied host" address is not actually an email address at all, but rather an implicit assumption made by certain software (and not even the majority of those). It seems valid for localhost, if you're running a *nix system, but only because these are Unix mail addresses, not email addresses. SMTP, POP3, and IMAP all require the host part, and servers/agents that allow those to be omitted are obscuring the spec for the sake of convenience (they add the implied host to all addresses before processing them).

At the end of the day, Mastodon isn't set up (nor likely to be) to fill in these implied hosts, so the `@` is, in fact, mandatory.

nightpool:
@danhunsaker ah, good to know! still, usability trumps correctness >:3 (and in this case, helping people figure out when they've "probably" filled out the form wrong is much better then like, optimizing for people filling out local addresses in a form where that doesn't even make a ton of sense)

PR #9572, opened 2018-12-19
Add REST API for creating an account

Gargron:
The method is available to apps with a token obtained via the client credentials grant. It creates a user and account records, as well as an access token for the app that initiated the request. The user is
unconfirmed, and an e-mail is sent as usual.

The method returns the access token, which the app should save for later. The REST API is not available to users with unconfirmed accounts, so the app must be smart to wait for the user to click a
link in their e-mail inbox.

The method is rate-limited by IP to 5 requests per 30 minutes.

**Motivation:** Multiple app developers have requested this. Currently an end-user cannot just install a mobile app and start using Mastodon, they must visit the website first. This, on the other hand, would allow a fully immersive sign-up process. The mobile app can implement a server picker, etc...

**Risks:** The API can be used to create spam bots. However, it is just as easy for a dedicated spammer to use the HTML form to create spam bots, and this API method actually has a higher rate limit (5/30 min here vs 25/30 min). E-mail confirmation is still required, so nothing changes there...

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -68,7 +68,7 @@ class Api::BaseController < ApplicationController
   end
 
   def require_user!
-    if current_user && !current_user.disabled?
+    if current_user && !current_user.disabled? && current_user.confirmed?
       set_user_activity
     elsif current_user
       render json: { error: 'Your login is currently disabled' }, status: 403
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -76,7 +76,7 @@ class Api::V1::AccountsController < Api::BaseController
   end
 
   def account_params
-    params.permit(:username, :email, :password)
+    params.permit(:username, :email, :password, :agreement)
   end
 
   def check_enabled_registrations
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -5,11 +5,12 @@ class Api::V1::AccountsController < Api::BaseController
   before_action -> { doorkeeper_authorize! :follow, :'write:follows' }, only: [:follow, :unfollow]
   before_action -> { doorkeeper_authorize! :follow, :'write:mutes' }, only: [:mute, :unmute]
   before_action -> { doorkeeper_authorize! :follow, :'write:blocks' }, only: [:block, :unblock]
-  before_action -> { doorkeeper_authorize! }, only: [:create]
+  before_action -> { doorkeeper_authorize! :write, :'write:accounts' }, only: [:create]
 
   before_action :require_user!, except: [:show, :create]
   before_action :set_account, except: [:create]
   before_action :check_account_suspension, only: [:show]
+  before_action :check_enabled_registrations, only: [:create]
 
   respond_to :json
 
@@ -76,5 +77,9 @@ class Api::V1::AccountsController < Api::BaseController
 
   def account_params
     params.permit(:username, :email, :password)
+  end
+
+  def check_enabled_registrations
+    forbidden if single_user_mode? || !Setting.open_registrations
   end
 end
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -1,19 +1,30 @@
 # frozen_string_literal: true
 
 class Api::V1::AccountsController < Api::BaseController
-  before_action -> { authorize_if_got_token! :read, :'read:accounts' }, except: [:follow, :unfollow, :block, :unblock, :mute, :unmute]
+  before_action -> { authorize_if_got_token! :read, :'read:accounts' }, except: [:create, :follow, :unfollow, :block, :unblock, :mute, :unmute]
   before_action -> { doorkeeper_authorize! :follow, :'write:follows' }, only: [:follow, :unfollow]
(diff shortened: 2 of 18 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -68,7 +68,7 @@ def current_user
   end
 
   def require_user!
-    if current_user && !current_user.disabled?
+    if current_user && !current_user.disabled? && current_user.confirmed?
```
Gargron: This condition was unneccessary before because unconfirmed users had no way to obtain an access token.

app/controllers/api/v1/accounts_controller.rb
```diff
@@ -16,6 +17,16 @@ def show
     render json: @account, serializer: REST::AccountSerializer
   end
 
+  def create
+    token    = AppSignUpService.new.call(doorkeeper_token.application, account_params)
+    response = Doorkeeper::OAuth::TokenResponse.new(token)
```
Gargron: This mirrors the usage of this class by Doorkeeper controllers.

nolanlawson:
This looks great. The only downside I can foresee is that (as you pointed out), the username and password are exposed to the app. But I don't see any way around that.

Having the app show a temporary screen saying "please go click a link in your inbox" is fine, but it's a bit odd that that link will go to the main Mastodon frontend, and there's no way for the app to intercept it or at least be notified that the user is now verified.

Maybe the API could accept a redirect URL as well, and the Mastodon frontend could display some text like, "Thanks for confirming your email. Application <strong>FooApp</strong> would like you to continue your signup <a href="http://example.com">here</a>."

Gargron:
>Maybe the API could accept a redirect URL as well, and the Mastodon frontend could display some text like, "Thanks for confirming your email. Application FooApp would like you to continue your signup here."

Hm... So far it was possible to implement this PR without changing data structures or requiring apps to re-register, but I don't think an extra redirect could be made with existing parts... Need to save information that a user was created by an app to be able to do that... Plus add an extra redirect URI to apps... Unless it would be OK to redirect users to the already saved OAuth redirect URI...

nolanlawson:
Hmmm, redirecting to the already saved OAuth URL would probably be fine. The app can just poke the API at `verify_credentials` to confirm that the user is confirmed, so no other info is required.

Gargron:
Done

mayaeh:
I think that we need to check the rules and terms and the conditions of it instance before creating an account. What do you think?

Also, how about making it possible to switch by setting ON/OFF of this API?

Gargron:
>I think that we need to check the rules and terms and the conditions of it instance before creating an account. What do you think?

An app could implement a web view to /about/more. Since rules are HTML, with potentially complex things, images, iframes, you would end up needing to render a web page anyway.

> Also, how about making it possible to switch by setting ON/OFF of this API?

Maybe. Apps must already deal with this API missing from older versions, so there is no consistency to lose.

mayaeh:
> An app could implement a web view to /about/more. Since rules are HTML, with potentially complex things, images, iframes, you would end up needing to render a web page anyway.

Fediverse have also instances for specific topics.
It is unfortunate that a mismatch occurs that "I created an account from an application, but it was not a place to accept me".

I think that it is heavy to let app developers handle all the measures to prevent them.


This is one idea:
1. When requesting creation of an account with this API, the server first returns a link (/about etc.) and a unique ID.
2. Then, the account registration is accepted only within a certain period after receiving the flag of "agree" and its ID with this API.

(I think that it is necessary to delete the registration unique ID expired periodically by cron job etc.)

I think it is in vain to send it link (/about, etc.) because it is already known.
But I hope they will read it before registering.

Gargron:
>When requesting creation of an account with this API, the server first returns a link (/about etc.) and a unique ID.

There is no guarantee that an app will show anything to the user, it could just send the ID back straight away. Likewise, even without the ID an app could make the user read through that page and click "Agree"

> Also, how about making it possible to switch by setting ON/OFF of this API?

I have been thinking about the on/off switch for this feature, and I think the admin settings UI has become too complex. Everything is in one form, that doesn't work well. I wouldn't want to add one more checkbox to it. :(

Neustrashimy:
Could you implement Boolean "confirmed to terms" parameter for creation API?
I want to get confirm that new user has read  my instance's custom terms/rules and agreed.
thanks.

mayaeh:
> There is no guarantee that an app will show anything to the user, it could just send the ID back straight away. Likewise, even without the ID an app could make the user read through that page and click "Agree"

I agree with that.
Besides, current flow of account creation API is consistent with WebUI.

However, WebUI has the following notation:
`By clicking "Sign up" below you agree to follow the rules of the instance and our terms of service.`

I think that if there is no equivalent to this API, if a registered user causes trouble, it will be considered "This response is unjust because you are not showing such a document".

I think that this complaint will be directed to the instance administrator, not to the application developer.


> I have been thinking about the on/off switch for this feature, and I think the admin settings UI has become too complex. Everything is in one form, that doesn't work well. I wouldn't want to add one more checkbox to it. :(

I think that there are really many items...

Gargron:
Okay, I have added required `agreement` param to the API.

Neustrashimy:
thanks a lot!

mayaeh:
I confirmed that I can create an account using this API.
There is no error as far as I can see.

I think that it is necessary to carefully write a document, such as handling "agreement" flag later.

mayaeh (approved):

PR #9573, opened 2018-12-19
Allow unauthenticated REST API access to GET /api/v1/accounts/:id/statuses

### LINKED ISSUE #7087: Make user statuses available when not authenticated

Gargron:
Fix #7087

The same data is available over the ActivityPub outbox, RSS, and Atom, so
there is little benefit to keeping it limited in this method.

### DIFF
--- a/app/controllers/api/v1/accounts/statuses_controller.rb
+++ b/app/controllers/api/v1/accounts/statuses_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Api::V1::Accounts::StatusesController < Api::BaseController
-  before_action -> { doorkeeper_authorize! :read, :'read:statuses' }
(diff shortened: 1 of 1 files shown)


ykzts (approved):

tateisu:
How about /api/v1/accounts/:id, /api/v1/accounts/:id/followers, /api/v1/accounts/:id/following ? 
All of these are viewable on the AP.

Gargron:
/api/v1/accounts/:id is already available without token.

PR #9583, opened 2018-12-19
Limit maximum visibility of local silenced users to unlisted

### LINKED ISSUE #9580: Silenced local users should not send public messages to other instances
### Pitch

If a local user is silenced, their outbound posts should be marked at most unlisted, and not public.

### Motivation

Local users are frequently silenced when they don't follow instance rules such as “no public untagged sensitive content”. Silencing them may appear to solve the issue, but this is not a sufficient way to ensure that instance rules are respected towards remote ones.

ClearlyClaire:
Fixes #9580

### DIFF
--- a/app/services/post_status_service.rb
+++ b/app/services/post_status_service.rb
@@ -23,13 +23,16 @@ class PostStatusService < BaseService
     status = nil
     text   = options.delete(:spoiler_text) if text.blank? && options[:spoiler_text].present?
 
+    visibility = options[:visibility] || account.user&.setting_default_privacy
+    visibility = :unlisted if visibility == :public && account.silenced
+
     ApplicationRecord.transaction do
       status = account.statuses.create!(text: text,
                                         media_attachments: media || [],
                                         thread: in_reply_to,
                                         sensitive: (options[:sensitive].nil? ? account.user&.setting_default_sensitive : options[:sensitive]) || options[:spoiler_text].present?,
                                         spoiler_text: options[:spoiler_text] || '',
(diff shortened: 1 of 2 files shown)


ykzts (approved):

app/services/post_status_service.rb
```diff
@@ -23,13 +23,16 @@ def call(account, text, in_reply_to = nil, **options)
...
+    visibility = :unlisted if visibility == :public && account.silenced
+
     ApplicationRecord.transaction do
       status = account.statuses.create!(text: text,
                                         media_attachments: media || [],
                                         thread: in_reply_to,
                                         sensitive: (options[:sensitive].nil? ? account.user&.setting_default_sensitive : options[:sensitive]) || options[:spoiler_text].present?,
                                         spoiler_text: options[:spoiler_text] || '',
-                                        visibility: options[:visibility] || account.user&.setting_default_privacy,
+                                        visibility: visibility,
```
Gargron: This does give away to the user that they are silenced. I thought you want to hook into the ActivityPub-facing JSON to force it there. Although you could argue that they can look at the ActivityPub JSON likewise to find out, so maybe this is the simplest solution...

ThibG: They can already to this by looking at the public timeline of their instance…?
Also, I'm not sure I see the point of silently silencing a local user. Shouldn't a decision about a *local* user be known to them?

Gargron:
Okay, now with #9519 communicating "your account is limited" to end-users it makes more sense for this change to limit post visibility in a way that's obvious to those end-users.

PR #9590, opened 2018-12-20
Skip mailer job retries when a record no longer exists

### LINKED ISSUE #8666: Displayed retry job at the Sidekiq: ActiveJob::DeserializationError
When after #8438 , if "mention" is deleted immediately after transmission, the error of the next image will be displayed in Sidekiq of the received instance.
This also occurs when you cancel "follow" immediately after "follow".
The same is true for "boost", and more.

I think that it need to search the Notification id at the notification_mailer.rb.
But.. I did not understand how to do it.

* * * *

Gargron:
Fix #8666

### DIFF
--- a/config/initializers/delivery_job.rb
+++ b/config/initializers/delivery_job.rb
@@ -1,0 +1,3 @@
+ActionMailer::DeliveryJob.class_eval do
+  discard_on ActiveJob::DeserializationError
+end


ThibG:
I'm concerned this may be too broad? But on the other hand, I can't see any occurrence where a job would fail with this error then work on a next attempt, hm…

Gargron:
@ThibG This is only within the ActiveMailer job, not any other Sidekiq workers.

ykzts (approved):

PR #9603, opened 2018-12-21
Add account creation date to admin user details UI

chr-1x:
A helpful piece of information to have when viewing a local user is their registration date. This PR adds that to the table of admin-visible info. It also rearranges the "Most recent" info so that the most recent time comes first (to line up with the registration date info)

### DIFF
--- a/app/views/admin/accounts/show.html.haml
+++ b/app/views/admin/accounts/show.html.haml
@@ -66,8 +66,10 @@
               = t('admin.accounts.enabled')
               = table_link_to 'lock', t('admin.accounts.disable'), disable_admin_account_path(@account.id), method: :post if can?(:disable, @account.user)
         %tr
-          %th= t('admin.accounts.most_recent_ip')
-          %td= @account.user_current_sign_in_ip
+          %th= t('admin.accounts.account_created')
+          %td
+            %time.formatted{ datetime: @account.created_at.iso8601, title: l(@account.created_at) }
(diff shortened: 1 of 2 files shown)


app/views/admin/accounts/show.html.haml
```diff
@@ -66,8 +66,10 @@
               = t('admin.accounts.enabled')
               = table_link_to 'lock', t('admin.accounts.disable'), disable_admin_account_path(@account.id), method: :post if can?(:disable, @account.user)
         %tr
-          %th= t('admin.accounts.most_recent_ip')
-          %td= @account.user_current_sign_in_ip
+          %th= t('admin.accounts.account_created')
```
Gargron: Is this a new string?

chr-1x: It seems to already have an english translation, so I assumed it was a string that either existed previously or is used currently somewhere else

chr-1x: grepping around for it and finding nothing, so I have no idea how it's resolving to anything meaningful in my dev instance. Added to the relevant section of `en.yml` (does anything else need to happen for new strings?)

chr-1x: I see, rails tries to be smart and auto-converts the variable name to a space-delimited phrase if there's no corresponding translation. ¯\\\_(ツ)\_/¯ ok

Issue #9611, opened 2018-12-23
Problems with patching/updating profile custom fields


skddc:
I found a couple of issues while trying to update a custom field of my profile via the API:

### Expected behaviour

1. `PATCH /api/v1/accounts/update_credentials` with parameter `fields_attributes[1][value]=Austria` should update the second custom field value with "Austria".
2. `PATCH /api/v1/accounts/update_credentials` with parameters `fields_attributes[1][name]=Location&fields_attributes[1][value]=Austria` should update the second custom field's name and value accordingly.

### Actual behaviour

1. All custom fields are deleted entirely (name + value), and a 200 response is returned with no errors and an empty custom field array.
2. All custom fields are deleted entirely, and replaced with only a single custom field with the new name and value, returned in the 200 response.

### Steps to reproduce the problem

Configure custom fields, try to update one.

### Specifications

Version 2.6.5

hongthang152:
@skddc Did you pass these parameters on the URL or in the request body ? 
If it is in the request body, I am having no issue with it so far. 
If it is in the URL, the controller might misinterprets the nested array index as the hash key.

skddc:
In the request body. I can reproduce it both with our deployed version and locally with the exact steps I described.

hongthang152:
Thats weird. Let me look further into this

Gargron:
You need to put the entire fields array there, you can't selectively update just a part of it

danhunsaker:
> You need to put the entire fields array there, you can't selectively update just a part of it

That sounds like the expected behavior of a PUT, rather than a PATCH. Under the HTTP spec, PATCH "applies partial modifications to a resource".

Gargron:
Well, PUT and PATCH are mapped to the same method in Rails. Think of the fields attribute like a single item, not a nested structure it's returned as. You can't PATCH to update the 3rd letter in a display name.

danhunsaker:
Always love it when frameworks violate the standards they are meant to support. 😕

nightpool:
🙄

nightpool:
`fields_attributes` is a single part of the larger account resource, and nobody follows the HTTP spec anyway—there are even some arguments that, by letter of the spec, [sending field values alone without instructions on how to reverse the changes is a violation](https://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/)

skddc:
[...]

PR #9619, opened 2018-12-24
Fix update field_attributes for Account Model

hongthang152:
#9611 
Remove checking for Hash object, change to checking for Array object.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -234,7 +234,6 @@ class Account < ApplicationRecord
       if !attr.is_a?(Hash) || attr[:name].blank?
         raise ArgumentError, 'One of custom field attributes has invalid format'
       end
-      
       previous = old_fields.find { |item| item['value'] == attr[:value] }
 
       if previous && previous['verified_at'].present?
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -230,8 +230,8 @@ class Account < ApplicationRecord
     fields     = []
     old_fields = self[:fields] || []
 
-    if attributes.is_a?(Array)
-      attributes.each do |attr|
+    Array.wrap(attributes).each do |attr|
+      if attr.is_a?(Hash)
         next if attr[:name].blank?
 
         previous = old_fields.find { |item| item['value'] == attr[:value] }
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -230,8 +230,8 @@ class Account < ApplicationRecord
     fields     = []
     old_fields = self[:fields] || []
 
-    if attributes.is_a?(Hash)
-      attributes.each_value do |attr|
+    if attributes.kind_of?(Array)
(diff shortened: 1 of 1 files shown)


app/models/account.rb
```diff
@@ -230,8 +230,8 @@ def fields_attributes=(attributes)
     fields     = []
     old_fields = self[:fields] || []
 
-    if attributes.is_a?(Hash)
-      attributes.each_value do |attr|
+    if attributes.is_a?(Array)
```
Gargron: This change isn't right either, because we get a hash from a standard form submission. So it can be a hash or an array.

Gargron (changes requested):

app/models/account.rb
hongthang152: Change to 
`Array.wrap(attributes).each do |attr|`

We wrap the array around a single element to allow iteration.

hongthang152: @Gargron You can check the latest commit

app/models/account.rb
```diff
@@ -230,18 +230,16 @@ def fields_attributes=(attributes)
     fields     = []
     old_fields = self[:fields] || []
 
-    if attributes.is_a?(Hash)
-      attributes.each_value do |attr|
-        next if attr[:name].blank?
+    Array.wrap(attributes).each do |attr|
+      next if !attr.is_a?(Hash) || attr[:name].blank?
```
nightpool: we should throw a validation error here instead of passing silently, to resolve #9611

hongthang152: It raises error now when it fails the validation.

hongthang152:
```
     if !attr.is_a?(Hash) || attr[:name].blank?
        raise ArgumentError, 'One of custom field attributes has invalid format'
      end
```

app/models/account.rb
```diff
@@ -230,18 +230,17 @@ def fields_attributes=(attributes)
...
-        next if attr[:name].blank?
-
-        previous = old_fields.find { |item| item['value'] == attr[:value] }
-
-        if previous && previous['verified_at'].present?
-          attr[:verified_at] = previous['verified_at']
-        end
+    Array.wrap(attributes).each do |attr|
+      if !attr.is_a?(Hash) || attr[:name].blank?
+        raise ArgumentError, 'One of custom field attributes has invalid format'
```
Gargron: I'm sorry, I disagree with @nightpool on this... Raising ArgumentError will lead to error 500 in the API, so it won't be very useful. It's possible to raise ActiveRecord::RecordInvalid, but I am not 100% sure if this is the right place to do it (this is in a setter, not in the validation cycle...)

Until there is a better validation solution, silently ignoring seems better. Sorry for long delay.

Gargron (changes requested):

PR #9620, opened 2018-12-24
Make backend code counts endline as 1 char instead of 2, like front-end

### LINKED ISSUE #4419: Server- and client-side bio and display name count endlines and emoji combinations inconsistently

ariasuni:
Fix #4419

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -112,11 +112,6 @@ class Account < ApplicationRecord
 
   delegate :chosen_languages, to: :user, prefix: false, allow_nil: true
 
-  # Convert newlines so validation doesn't count them as two characters
-  def note=(val)
-    self['note'] = val.gsub("\r\n", "\n")
-  end
-
   def local?
     domain.nil?
   end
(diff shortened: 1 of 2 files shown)


danhunsaker (approved):
This is the way I was going to suggest doing it if you hadn't already, so +100!

app/models/account.rb
```diff
@@ -112,6 +112,11 @@ class Account < ApplicationRecord
 
   delegate :chosen_languages, to: :user, prefix: false, allow_nil: true
 
+  # Convert newlines so validation doesn't count them as two characters
+  def note=(val)
+    self['note'] = val.gsub("\r\n", "\n")
```
Gargron: I suggest:

```ruby
before_validation :normalize_note

private

def normalize_note
  self.note = note.gsub("\r\n", "\n")
end
```

Worth noting I do not remember something like this for status text...

ariasuni: This code is used for counting a status text:

https://github.com/tootsuite/mastodon/blob/bf70e5cfdaef9b7a3a927548a05203a864d5bea9/app/validators/status_length_validator.rb#L18

I couldn’t make a test triggering `before_validation`, so I didn’t go for you suggestion or for using `grapheme_length`.

ariasuni:
Thanks.

Issue #8003, opened 2018-07-12
Inform followers that your account has moved


garbados:
Migration tooling should proactively inform a user's followers that the migration has occurred.

Here are some feature suggestions that meet this requirement:

- [ ] After indicating that your account has moved, the instance from which you are migrating should send a notification to all your followers that you have moved which includes a link to the new profile, where a follower can opt to re-follow the user at their new instance. This serves as a deprecation notice for the old account.
- [ ] For a time (ex: six months), every time a migrated user toots from their new account, the old account boosts it with a header including a button to follow the user at their new location. This serves as an ongoing deprecation notice for the old account.

Here are some backend suggestions to make it work:

- [ ] API endpoint to broadcast notifications to all your followers.
- [ ] Backend hook to create and delete rules indicating that an account should automatically boost posts from another account. (In this case, specifically for the old account to boost the new account)
- [ ] New UI component for the "header message" which accompanies these automated boosts and includes a follow button.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

elomatreb:
I wonder if a "regular" notificatoin in the Notifications column could possibly be a better choice. Something like:

"User username has moved accounts. Do you want to follow their new account? [Yes] [Dismiss]"

Would avoid the need for self-boosting toots, which feel a little awkward if I'm honest (because even if it is repeated, it won't reliably reach all followers).

hoodiek:
sending the notification again could work too but that is maybe too intrusive

RobertRence:
Is it possible this could be exploited with circular references between accounts to create a DDOS across a set of federations?

As a potential solution, you might want to make it so it only allows a single boost.

nightpool:
duplicate of #177, i think?

jfmcbrayer:
I think it's a narrower, more incremental approach to #177. Just like delete-and-redraft is a narrower, more incremental approach to editing. It's also more concrete.

paralithode:
This seems like a great solution to the issue, one I recently encountered myself. I moved instances & have been experimentally boosting a post explaining that I've moved, or a post from my new account, every few days. I consistently garner a few refollows at my new account every time I do this, so evidently it's very easy to miss someone having moved in the current system.

trwnh:
re: #177, yeah, this has been discussed there, but it's also more of an improvement to the already-existing "account moved" setting and not at all to do with migrating your account -- it's about moving to a new account.

[...]

Issue #6955, opened 2018-03-29
Suggest following new account whenever a followed account moves


ThibG:
Since 2.1.0, Mastodon displays and federates information about migrated accounts.
However, end-user followers of a migrated account are not notified of such changes.

Followers should be notified in some way, suggesting them to follow the new account and unfollowing the old one. But I currently have no idea how that should be presented to the user.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
The idea is to make the unfollowing & re-following automatic. The challenge is that the process needs to be authorized by both old and new account and followers need to be able to verify this. So like having two signatures or something like that...

Gargron:
Something like: old account generates Move activity and sends it to new account. New account sends back Accept->Move (signed). Old account adds own signature to that, and sends it to followers..? Or would it have to become Announce->Accept->Move? It's getting a bit weird. Paging @cwebber

ThibG:
While a completely automatic process is a worthwhile goal, it's going to be a massive pain to get right, specify and implement properly.

On the other hand, if we can figure out the correct interface for it, the suggestion approach can be deployed right now, take care of accounts that have already been migrated, require no further specification, and have the advantage of letting the user know what's going on and give them a chance to check that the account migration isn't fishy.

ThibG:
That would be possible and that seems reasonable. However:
- That's a new notification type, which may cause issues with clients
- That would require a rake task or something to issue notifications for previously-migrated accounts

Sylvhem:
Is the last point really mandatory? If it's too much of a burden to implement, I don't know if it's really worth it.

ThibG:
Nah, it would just be better to not miss out on the old migrations. And I don't think that point would be very difficult to write.

I may try to work on this if @Gargron doesn't oppose to the idea.

ThibG:
@Gargron I still think we need something that doesn't require additional protocols, even if it's not entirely automatic as a result. Indeed, what about accounts that already moved on? Accounts moving from an unmaintained server that won't support the new protocol? There are many cases where the ideal fully-automatic idea isn't applicable.

I think we need to implement either a new type of notifications (but that may break existing clients depending on how they handle notifications) or a new API endpoint entirely (but I'm not sure how to display the information then…)

[...]

PR #9629, opened 2018-12-26
Add handler for Move activity

Gargron:
See: #177, #6955, #8003

### Traffic

My guesses about the traffic patterns of this feature:

Given N as the number of followers on the original account, and M (M <= N) as the number of separate servers those followers are on.

1. The Move activity is distributed to M inboxes
2. M servers access the new account via account resolution
3. N accounts send unfollow activity to the old server
4. N accounts send follow activity to the new server
5. N accept-follow activities are distributed from the new server to the new followers

Categorizing the workload by servers goes as such:

- The old server has to make inbox deliveries M times and process N incoming activities (Simplified: 2N)
- The new server has to answer M queries and process N incoming activities, then send out N activities more (Simplified: 3N)
- Each server that has any followers of the old account must process 1 incoming activity and send out 2 \* (x / N) activities. In some cases this includes the old and new servers, so this could add on top of their workloads as well.

Let's imagine a large account decides to move - me. At the time of writing, that's roughly N=200,000. That means the old server has to process 400,000 Sidekiq jobs, the new server has to process 600,000 Sidekiq jobs.

That is to say, that's a lot of Sidekiq jobs, and a lot of traffic. For this reason I believe that the UX around the account migration feature should be *incredibly strict*, to prevent multiple uses by the same person (across servers, too) as much as possible.

### Risks

It allows potentially amassing a large following for a brand new account from multiple smaller parts. Kind of like that large Mastodon power ranger mech assembling from the smaller ones, or that game where you are a bubble that eats other bubbles to grow larger. Amass a number of followers, then sell your account to someone who is willing to pay for them. (Consider that I primarily view this risk not from the perspective of high follower counts being desireable, but from the perspective of being able to put a message on many people's home feeds).

### Risk mitigation

On the old server side, it is easy to add a cooldown (perhaps permanent!) on sending out Move activities. On the new server side, it is required to add the old account's URI to an "also known as" attribute, otherwise the Move activities won't validate anywhere (i.e. both old and new accounts have to point at each other). So it is possible to put a cooldown on updating that attribute, as well.

However, those precautions can be patched out by a malicious admin (or non-Mastodon software), i.e. the risk of massive follower accumulation from different sources remains.

### End notes

Just as a disclaimer since it may not be obvious, this PR, in any case, does not include any UIs for performing Move activities in any capacity. For this feature to be effective, the UIs cannot be released until most of the Mastodon servers in the wild have upgraded to a release that includes this Move handler.

### DIFF
--- a/app/lib/activitypub/activity.rb
+++ b/app/lib/activitypub/activity.rb
@@ -50,6 +50,8 @@ class ActivityPub::Activity
         ActivityPub::Activity::Add
       when 'Remove'
         ActivityPub::Activity::Remove
+      when 'Move'
+        ActivityPub::Activity::Move
       end
     end
   end
--- a/app/lib/activitypub/activity/follow.rb
+++ b/app/lib/activitypub/activity/follow.rb
@@ -6,7 +6,7 @@ class ActivityPub::Activity::Follow < ActivityPub::Activity
 
     return if target_account.nil? || !target_account.local? || delete_arrived_first?(@json['id']) || @account.requested?(target_account)
 
-    if target_account.blocking?(@account) || target_account.domain_blocking?(@account.domain)
+    if target_account.blocking?(@account) || target_account.domain_blocking?(@account.domain) || target_account.moved?
       reject_follow_request!(target_account)
       return
     end
--- a/app/lib/activitypub/activity/move.rb
+++ b/app/lib/activitypub/activity/move.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class ActivityPub::Activity::Move < ActivityPub::Activity
+  PROCESSING_COOLDOWN = 7.days.seconds
+
+  def perform
+    return if origin_account.uri != object_uri || processed?
+
+    mark_as_processing!
+
+    target_account = ActivityPub::FetchRemoteAccountService.new.call(target_uri)
+
+    return if target_account.nil? || !target_account.also_known_as.include?(origin_account.uri)
+
+    # In case for some reason we didn't have a redirect for the profile already, set it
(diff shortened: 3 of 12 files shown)


Gargron:
>If you could use help beta testing

Unfortunately, you won't be able to test this until all the servers where you have followers will have this patch, by which point testing it will be useless.

trwnh:
>3. N accounts send unfollow activity to the old server

I'm concerned this means that old followers-only content will become inaccessible. Unless content is also moved over, or if the privacy code checks for relations with the `alsoKnownAs`/`movedTo`, maybe?

Gargron:
Unfollowing the old account is a crucial safety mechanism. Otherwise there's gonna be a business of selling followers, not just once to one person, but to as many people as are willing to pay.

ykzts (approved):

ThibG:
I understand your concerns, but I'm really uneasy with the unfollow part… especially with the code attempting to unfollow, then attempting to follow, possibly failing in the process.

I wonder if we really need fully-automated migration? What about some kind of notification giving user the choice to follow the new account and (separately) unfollow the old one?

nightpool:
Yeah, after looking closely at this, a fully-automated unfollowing that removes access to previously accessible follower-only content is going to be really surprising for most people. Let's fix this in a follow-up before release.

Kjwon15:
I think that unfollowing part should changed.
1. Not every server is Mastodon.
If follower is on non-mastodon server, unfollowing is never happened.
2. If original account concerns about privacy, Use Block-Unblock(or something) to force remove followers. And it should be checkboxed option for migrator.

PR #9630, opened 2018-12-26
Fix scope latest of ReportNote

ysksn:


Gargron (approved):

ThisIsMissEm:
Why was the sort order changed here? I don't think having the newest note at the top, and the oldest note above the comment field makes sense, especially when Report Notes was intended to allow discussion between admins about reports.

PR #9645, opened 2018-12-28
Improve admin UI for domain (block) management

Gargron:
- Retire legacy "resubscribe" function from known instances page
- Unite "known instances" and "domain blocks" into "federation"
- Dashboard for any known domain

___

image

image

image


jeroenpraat:
Good idea to combine those two. I also like the UI.

I advise to change the title 'Federation'. This title doesn't explain enough what to find here. Maybe 'Other instances' or 'Other domains' is better.

CrazyJake88:
The date a domain was limited should be included next to it. We also need a text field to add a note like when moderating individual reports as to why a domain was limited.

nightpool:
Im not sure im a 100% convinced that conflating federation and Known Accounts is a good idea? Unless the label is misleading and it actually lists all accounts that follow or are being followed by local accounts (in which case i think its better to split those for clarity)

nightpool:
Maybe we should replace silenced with "Hide" and "Suspend" with "Ban", to be more intuitive?

ClearlyClaire (approved):

PR #8653, opened 2018-09-08
Styled scrollbars in Firefox Nightly

sorin-davidoi:
You currently need to enable `layout.css.scrollbar-colors.enabled` in `about:config` in Firefox
Nightly.

image

PR #9649, opened 2018-12-28
Style column scrollbars

pascalduez:
Style column scrollbars with latest spec properties.  
See: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars  

This will work for Firefox >= 64 (current), as well as spec compliant browsers.

I reused the colors from the `::-webkit-scrollbar` rules in `reset.scss`.  

### Before
screenshot_2018-12-28 la quadrature du net - mastodon - media federe 1

### After 
screenshot_2018-12-28 la quadrature du net - mastodon - media federe

NB: screenshot on a Gnome env.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -1861,6 +1861,13 @@ a.account__display-name {
 
   > .scrollable {
     background: $ui-base-color;
+    scrollbar-width: thin;
+    scrollbar-color: lighten($ui-base-color, 4%) rgba($base-overlay-background, 0.1);
+    transition: scrollbar-color 0.2s;
+
+    &:hover {
+      scrollbar-color: lighten($ui-base-color, 6%) $ui-base-color;
+    }
   }
 }
 


ykzts (approved):

ThibG:
Wait, isn't it already supposed to work?

pascalduez:
> Wait, isn't it already supposed to work?

There's some `::-webkit-scrollbar` in `reset.css` rules, but they are none standard, and applies only for Chrome, Webkit browsers.

Gargron:
@pascalduez No, #8653 and #9210

pascalduez:
@Gargron #8653 are outdated properties. #9210 update them.

I need to investigate why it's not inherited by the columns then.

Gargron:
On my Firefox 64 they are inherited. Both on Windows and Linux.

pascalduez:
Yup, I edited my comment.

So to sum up, it's up to you. 

I think this change still have it's place, the `html` declaration act as a nice default, while the one on `.column > .scrollable` are more tailored.

Gargron:
I tried this out and save for a small transition effect it's identical to what we already have, except much more narrow. Difficult to target with the mouse. Sorry, I don't think I will merge this. Thank you for your time.

PR #9652, opened 2018-12-29
Replace Ruby 2.3 with 2.6 in CircleCI

shleeable:
* Added Ruby 2.6 and Redis 5
* Dropped Ruby 2.3 due to upcoming EOL
* Updated Docker Images with most recent versions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,13 +116,6 @@ jobs:
         environment: *ruby_environment
     <<: *install_ruby_dependencies
 
-  install-ruby2.3:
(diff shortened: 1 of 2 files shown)


Gargron (approved):

ThibG:
I am very surprised this works :|
(And it does not work in glitch-soc)

The reason for this is that `build` and `test-i18n` tasks depend on `install-2.5` which installs dependencies for Ruby 2.5, but the environment it then uses is… ruby 2.6, since that's the new default.

I'll open a PR for this.

PR #9654, opened 2018-12-29
Skip full text search if failed

tribela:
Sometimes, ES server are dead while main server is live.
We should give general search results(accounts, hashtags) even if ES server is dead.

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -14,8 +14,12 @@ class SearchService < BaseService
         results.merge!(url_resource_results) unless url_resource.nil?
       elsif query.present?
         results[:accounts] = perform_accounts_search! if account_searchable?
-        results[:statuses] = perform_statuses_search! if full_text_searchable?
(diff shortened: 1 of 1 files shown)


Gargron (changes requested):

app/services/search_service.rb
```diff
@@ -62,7 +62,12 @@ def url_resource_symbol
 
   def full_text_searchable?
```
Gargron: That's not the right method. Also, def/end blocks can have a rescue clause without extra begin/end.

Gargron (changes requested):

app/services/search_service.rb
Kjwon15: like this?
```ruby
def method
  do_something
rescue Exception
  fallback
end
```

Gargron: Yes, like that. :+1:

Gargron (approved):

PR #9656, opened 2018-12-29
Dockerfile upgraded docker images for node 8.15/ruby 2.5.3

shleeable:
We need to test this - I believe the sigfault with sidekiq was resolved months ago.

If you're running 2.5 and sidekiq container has not crashed after 48 hours.... please comment


Reverite:
@ashleyhull-versent Ruby 2.5.3 still crashes for me.

PR #9661, opened 2018-12-30
Use Contact User as Relay, Report, Subscribe.

yukimochi:
Following relay, sending report and subscribing are used user selected by `Account.local.where(suspend: false)` in Mastodon.
This method may give responsibility someone who does not know anything.
Use Contact User if set.

### DIFF
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -5,7 +5,7 @@ module AccountFinderConcern
 
   class_methods do
     def representative!
-      find_remote(Setting.site_contact_username.gsub(/\A@/, ''), nil) || raise(ActiveRecord::RecordNotFound)
+      find_local!(Setting.site_contact_username.gsub(/\A@/, ''))
     end
 
     def find_local!(username)
@@ -17,7 +17,7 @@ module AccountFinderConcern
     end
 
     def representative
-      find_remote(Setting.site_contact_username.gsub(/\A@/, ''), nil)
+      find_local(Setting.site_contact_username.gsub(/\A@/, ''))
     end
 
     def find_local(username)
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -13,7 +13,7 @@ module AccountFinderConcern
     end
 
     def representative
-      find_local(Setting.site_contact_username.gsub(/\A@/, ''))
+      find_local(Setting.site_contact_username.gsub(/\A@/, '')) || Account.local.find_by(suspended: false)
     end
 
     def find_local(username)
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -4,10 +4,6 @@ module AccountFinderConcern
   extend ActiveSupport::Concern
 
   class_methods do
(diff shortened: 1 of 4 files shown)


app/models/relay.rb
```diff
@@ -68,7 +68,7 @@ def unfollow_activity(activity_id)
   end
 
   def some_local_account
-    @some_local_account ||= Account.local.find_by(suspended: false)
+    @some_local_account ||= Account.find_local(Setting.site_contact_username.gsub(/\A@/, '')) || Account.local.find_by(suspended: false)
```
Gargron: Perhaps create `Account.representative` method?

Gargron (changes requested):

app/models/relay.rb
yukimochi: OK.
Sorry for many force-pushes, all done.

app/models/concerns/account_finder_concern.rb
```diff
@@ -4,6 +4,10 @@ module AccountFinderConcern
   extend ActiveSupport::Concern
 
   class_methods do
+    def representative!
+      find_remote(Setting.site_contact_username.gsub(/\A@/, ''), nil) || raise(ActiveRecord::RecordNotFound)
```
nightpool: `find_local!`

app/services/subscribe_service.rb
```diff
@@ -43,7 +43,7 @@ def subscription_params
   end
 
   def some_local_account
-    @some_local_account ||= Account.local.where(suspended: false).first
+    @some_local_account ||= Account.representative || Account.local.where(suspended: false).first
```
nightpool: no reason to use Account.representative here.

yukimochi: Certainly, that's right.

nightpool:
I'm a little nervous about using some_local_account or Account.representative interchangeably—if we're saying relay subscription confers responsibility in some way, then the silent fallback could be dangerous. Certainly federated reports do confer responsibility. But at the same time, we can't expect that all instances have a contact_username set. 

I think we should resolve this by having a "server" actor as suggested a couple of times when the relay and federated reports were being developed—the combined usecases of relays and federated reports make this make sense, even if we wouldn't want the complexity for either alone.

app/models/concerns/account_finder_concern.rb
```diff
@@ -4,6 +4,10 @@ module AccountFinderConcern
   extend ActiveSupport::Concern
 
   class_methods do
+    def representative!
```
Gargron: Since this is not used anywhere, I do not think it is required.

app/models/concerns/account_finder_concern.rb
```diff
@@ -12,6 +16,10 @@ def find_remote!(username, domain)
       find_remote(username, domain) || raise(ActiveRecord::RecordNotFound)
     end
 
+    def representative
+      find_local(Setting.site_contact_username.gsub(/\A@/, ''))
```
Gargron: I believe the `|| Account.local.find_by(suspended: false)` logic should be included here, since it is repeated everywhere this method is called.

Gargron (changes requested):

Gargron (approved):

PR #9677, opened 2018-12-31
Add CSV export for lists and domain blocks

### LINKED ISSUE #6893: Ability to export Lists
I have defined a few follower Lists, but now that I go to export one of my Lists to another account, I find that there isn't a UI for it. It's not one of the options in the Data Export panel, and not in the archive tarball.

* * * *

### LINKED ISSUE #9268: Make domain blocks importable/exportable
### Pitch

At the moment it's already possible to export local blocklists from a user account and import them back into another account. It would be useful to have the same functionality for the domain blocks (the ones added manually by admins).

### Motivation

There are a lot of public "bad actor" domain blocklists for instances that have illegal content or contain harassing users. Unfortunately, adding these domains to the blocklist is still a manual process that in the case of some lists can both take quite a while and needs to be manually monitored for updates.

This would greatly benefit admins that wish to use said public blocklists as this would let them add these blocklists and add new domains to it as needed. While it does not remove the manual update factor, it would greatly reduce the time needed to block new domains or add a new blocklist.

Gargron:
Fix #6893
Fix #9268

### DIFF
--- a/app/controllers/settings/exports/blocked_domains_controller.rb
+++ b/app/controllers/settings/exports/blocked_domains_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Settings
+  module Exports
+    class BlockedDomainsController < ApplicationController
+      include ExportControllerConcern
+
+      def index
+        send_export_file
+      end
+
+      private
+
+      def export_data
+        @export.to_blocked_domains_csv
+      end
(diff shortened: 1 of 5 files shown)


ClearlyClaire (approved):
A rspec test for each of the new methods would be nice, but LGTM.

PR #9695, opened 2019-01-02
Fix CircleCI

ClearlyClaire:


.circleci/config.yml
```diff
@@ -186,17 +186,17 @@ workflows:
       - install-ruby2.6:
           requires:
             - install
-            - install-ruby2.5
       - install-ruby2.5:
           requires:
             - install
+            - install-ruby2.6
```
Gargron: Hm why do 2.5 and 2.4 have to wait until after 2.6 installs?

ThibG: Same reason other versions had to wait on 2.5, it's to enforce ordering whe a task inherits its workspace from multiple tasks.
We added that in #8228

Gargron (approved):

PR #9706, opened 2019-01-03
Add scheduled statuses

### LINKED ISSUE #340: Scheduled Toots
The ability to schedule toots is very useful for content creators who want to send toots at a time more people will be on the service.

(Related: #286 Time-delayed reblogging.)

Gargron:
Allows users to schedule a toot (with media attachments) to be published at a certain future date. This PR only includes REST API changes, no UI.

The scheduled date must be at least 5 minutes into the future. At most, 300 toots can be scheduled at the same time. Only 50 toots can be scheduled for any given day.

Fix #340

### API changes

- `POST /api/v1/statuses` new param `scheduled_at` (timestamp string). When present, and more than 5 minutes into the future, the API method will return a Scheduled Status instead of Status. The Scheduled Status will be turned into a real Status roughly at the scheduled time.
- `GET /api/v1/scheduled_statuses`
- `GET|PUT|DELETE /api/v1/scheduled_statuses/:id`

### Refactoring

The PR includes a refactor of the PostStatusService class to allow passing all necessary options from a single hash, which allows encoding it more easily in the scheduled_statuses table.

### Overview

The high-level overview of the PR: When scheduled_at option is present, instead of creating a status, we only run status validation, and if it passes, we create an entry in scheduled_statuses which encodes the status attributes. Media attachments now also track scheduled_status_id to prevent clearing out the ones that are going to be attached later. Every 5 minutes, a scheduler iterates over the scheduled_statuses table to fetch the ones due in the next 5 minutes, and push them into a more precise Sidekiq queue. In Sidekiq, the individual statuses are created, with media attachments being unassigned from the scheduled status and assigned to the real one.

### DIFF
--- a/app/controllers/api/v1/scheduled_statuses_controller.rb
+++ b/app/controllers/api/v1/scheduled_statuses_controller.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+class Api::V1::ScheduledStatusesController < Api::BaseController
+  include Authorization
+
+  before_action -> { doorkeeper_authorize! :read, :'read:statuses' }, except: [:update, :destroy]
+  before_action -> { doorkeeper_authorize! :write, :'write:statuses' }, only: [:update, :destroy]
+
+  before_action :set_statuses, only: :index
+  before_action :set_status, except: :index
+
+  after_action :insert_pagination_headers, only: :index
+
+  def index
+    render json: @statuses, each_serializer: REST::ScheduledStatusSerializer
+  end
+
+  def show
+    render json: @status, serializer: REST::ScheduledStatusSerializer
+  end
+
+  def update
+    @status.update!(scheduled_status_params)
+    render json: @status, serializer: REST::ScheduledStatusSerializer
+  end
+
+  def destroy
+    @status.destroy!
+    render_empty
+  end
+
+  private
+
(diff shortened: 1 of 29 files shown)


giromide:
Apologies for this not being a code review, but this is another way to set Mastodon apart from that other service with the birds.

giromide:
My suggestion was to limit scheduled posts to five, if only to match the minimum amount of time one could schedule. Five posts. Five minutes.

tateisu:
Is there a way to list scheduled statuses and delete one?

deutrino:
This is essential for curators. Thanks for doing this work. CRUD would be really helpful, yes.

ykzts (approved):

jeroenpraat:
@Gargron Any change this will result in a UI for the web-app somewhere in the (near) future?

tateisu:
please add check scheduled_at is after 5min or later, because the wall clock of client machine may not correct, and customized instance may have different limit.

schiessle:
Is there already a issue/feature request for the web UI to create and manage scheduled toot's I could follow?

schiessle:
As I couldn't find a follow-up issue I created now one https://github.com/tootsuite/mastodon/issues/10088

PR #9718, opened 2019-01-04
Bump version to 2.7.0rc1

Gargron:


Quenty31:
Please update the translations for every RC :pray:

PR #9722, opened 2019-01-05
Enable immutable caching for S3 objects

### LINKED ISSUE #9721: No Cache-Control header in S3 objects
### Expected behaviour

When requesting an object uploaded to S3, the Cache-Control response header should be some kind of long-lived value, to indicate to the browser that it doesn't need to do revalidation requests (since the URLs contain a checksum). This would save money for admins and increase performance for users.

My recommended Cache-Control header would be:

    Cache-Control: public, max-age=31536000, immutable

(See [immutable caching](https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/) for details.)

Note that these headers would also propagate to CloudFront, as it just uses whatever headers S3 serves up.

### Actual behaviour

If you request an object from S3, there are no Cache-Control headers.

Request any object from S3, e.g. an avatar image.

### Specifications

Mastodon 2.6.5

nolanlawson:
Fixes #9721 

I also added "public" here, as I can't think of a good reason not to add it. Perhaps it has some marginal benefit in that ISPs (or other proxies) can cache it for all users. The assets are certainly publicly available and the same for all users.

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -25,7 +25,7 @@ if ENV['S3_ENABLED'] == 'true'
     s3_protocol: s3_protocol,
     s3_host_name: s3_hostname,
     s3_headers: {
-      'Cache-Control' => 'max-age=315576000',
+      'Cache-Control' => 'public, max-age=315576000, immutable',
     },
     s3_permissions: ENV.fetch('S3_PERMISSION') { 'public-read' },
     s3_region: s3_region,


Gargron (approved):

nolanlawson:
It just occurred to me that this may be a moot point for HTTPS anyway.  I can't seem to find any info on how proxies handle `cache-control: public` for HTTPS requests.

Then again, some admins may be using S3 with HTTP rather than HTTPS. And in either case I don't think there's any harm with adding `public`, yeah.

nolanlawson:
BTW for the benefit of Mastodon admins reading this issue who may want to add this header to existing S3 assets, you can use this `s3cmd` command:

```bash
s3cmd --recursive modify --add-header="Cache-Control:public, max-age=315576000, immutable" s3://my-bucket-name/
```

Note that you will want the latest version of `s3cmd` from GitHub because apparently there are connection timeout bugs in the Debian version:

```bash
git clone https://github.com/s3tools/s3cmd.git
sudo apt install python-dateutil
./s3cmd/s3cmd --help
```

PR #9730, opened 2019-01-06
Remove useless argument in RegenerationWorker#perform

yusuko:
# Description

`RegenerationWorker.perform_async` is used with only one argument now.
So I remove useless argument ` _ = :home` for readability.


Gargron:
The risk with modifying this is that you can't predict what version people will upgrade from. If it's from a version where the job was still enqueued with two arguments, it will lead to errors...

yusuko:
Thank you for the reply👍
It's true. I'm sorry for an easy mistake...

PR #9745, opened 2019-01-07
Move profile directory link in web UI from footer to discover section

Gargron:


ThibG:
That would be the first “getting started” column link to actually lead outside of the WebUI, right?

Gargron:
@ThibG No, there is a settings link that appears in single-column layout.

PR #9753, opened 2019-01-08
Not to skip executable specs

ysksn:


Gargron:
We skip those because they are so much slower :(

ysksn:
It still takes 1/4 of the original, if you don't like the last commit, please close this PR. Thanks :)
@Gargron

Gargron (approved):

Issue #8237, opened 2018-08-17
[Accessibility] Hovertext for image descriptions is insufficient


Cassolotl:
> image descriptions should be made available in the same formatting as the text of posts.
> 
> on-hover is not good enough, not everyone who needs an image description uses a screen reader and not everyone who needs an image description can read hover text.

[Source](https://sins.center/@Fidgetcetera/100565022574290523)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Fidgetcetera:
we'll expand on this idea a bit since sometimes when we point the issue out people ask us for details, but basically - hover text is hard to invidually style for people who need custom stylings for their browser (for example, overriding fonts, making text larger, etc. hover text is also very difficult (almost impossible?) to view without a mouse.

a potential nice way to handle this could be a "view image description(s)" button underneath a post, that can be optional (if turned off, the image descriptions just are written beneath the images). ideally the image description text would be very similar to the post's general text (same font and size and everything).

hughrun:
This seems like a pretty reasonable request to me, just wondering about implementation - I wouldn't want to see this become "image description is always visible to all users". 

@Fidgetcetera what do you mean by 'beneath'? Do you mean on a new line below the image, or rendered on a layer below the image so it's not visible?

Is the proposal that the button would basically be a toggle to show the description in place of the image? Could it (additionally?) be a user setting so that they always see the description by default rather than the image?

LogicalDash:
When you click an image, currently, you get a modal with the image displayed at full size. How about simply displaying the image description there as well?

Fidgetcetera:
@hughrun 
"rather than" is not a great idea, since for some people they _can_ see the image, but cannot make out or discern all the details, and would prefer to see both at the same time. we meant on a new line, beneath image(s). we aren't really UI designers here, so we're not sure if the best way would be placed beneath each individual image of multi-image posts, or just at the end of the post after the images, or what.

Fidgetcetera:
@LogicalDash that might be a bit better, but an option to view it just alongside the normal post text would also be great.

hughrun:
Multi-image posts can (should) have a different description for each image, so it would either need to be a description under each image or multiple descriptions at the end of the post something like 
`Image 1: tabby cat`<br>
`Image 2: black cat`

trwnh:
Tumblr also showed image descriptions in their lightbox. Mastodon should consider doing the same.

[...]

PR #9754, opened 2019-01-08
Add alt text to the lightbox for accessibility.

Kansattica:
Regarding issue #8237, I put together a simple proof of concept for how alt text could be added to the lightbox in images. This solution would require images with alt text to be scaled down slightly smaller, both for aesthetic reasons and to prevent images with long alt text from being pushed off the screen. I don't understand the code well enough to implement the scaling here, but I hope it helps.

image

### DIFF
--- a/app/javascript/mastodon/features/ui/components/zoomable_image.js
+++ b/app/javascript/mastodon/features/ui/components/zoomable_image.js
@@ -133,6 +133,7 @@ export default class ZoomableImage extends React.PureComponent {
         ref={this.setContainerRef}
         style={{ overflow }}
       >
+	<figure>
         <img
           role='presentation'
           ref={this.setImageRef}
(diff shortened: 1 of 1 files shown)


scottaohara (changes requested):
Unfortunately, while it's understandable as to why this might seem like the right thing to do, `figure`s which contain images that are hidden to screen readers (`alt=''` or `role='presentation'`) can [cause certain screen reader / browser pairings to ignore `figure` & `figcaption` patterns all together](https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html#figures-and-screen-readers).  

Ideally a few things would change about this to make the images accessible:
1. there shouldn't be a `role=presentation`.  Even with an `alt` set, `role=presentation` will hide the image to screen readers.
2. The use of `title` and `alt` on the image is redundant and can cause duplicate announcements with Chrome.  Use one or the other, but ideally `alt`.
3. Remove the `figure` and `figcaption` elements. Unfortunately, in this context they're not appropriate and can cause for reduced accessibility :(

PR #9755, opened 2019-01-08
Add hashtag filter to profiles

Gargron:
screenshot_2019-02-03 admin admin localhost 3000

Allow browsing a profile by a specific hashtag. Useful for artists.

Public profile change: `GET /@:username/tagged/:hashtag`
REST API change: `GET /api/v1/accounts/:id/statuses?tagged=:hashtag`

___

TODO:

- [x] Public profile page route
- [x] REST API
- [x] List of browserable hashtags in sidebar on public profile

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -57,6 +57,7 @@ class AccountsController < ApplicationController
 
   def filtered_statuses
     default_statuses.tap do |statuses|
+      statuses.merge!(hashtag_scope)    if tag_requested?
       statuses.merge!(only_media_scope) if media_requested?
       statuses.merge!(no_replies_scope) unless replies_requested?
     end
@@ -78,12 +79,15 @@ class AccountsController < ApplicationController
     Status.without_replies
   end
 
+  def hashtag_scope
+    Status.tagged_with(Tag.find_by(name: params[:tag].downcase)&.id)
+  end
+
   def set_account
     @account = Account.find_local!(params[:username])
   end
 
   def older_url
-    ::Rails.logger.info("older: max_id #{@statuses.last.id}, url #{pagination_url(max_id: @statuses.last.id)}")
     pagination_url(max_id: @statuses.last.id)
   end
 
@@ -92,7 +96,9 @@ class AccountsController < ApplicationController
   end
 
   def pagination_url(max_id: nil, min_id: nil)
-    if media_requested?
+    if tag_requested?
+      short_account_tag_url(@account, params[:tag], max_id: max_id, min_id: min_id)
+    elsif media_requested?
       short_account_media_url(@account, max_id: max_id, min_id: min_id)
     elsif replies_requested?
       short_account_with_replies_url(@account, max_id: max_id, min_id: min_id)
@@ -107,6 +113,10 @@ class AccountsController < ApplicationController
 
   def replies_requested?
     request.path.ends_with?('/with_replies')
+  end
+
+  def tag_requested?
+    request.path.ends_with?("/tagged/#{params[:tag]}")
   end
 
   def filtered_status_page(params)
--- a/app/controllers/settings/featured_tags_controller.rb
+++ b/app/controllers/settings/featured_tags_controller.rb
@@ -6,22 +6,28 @@ class Settings::FeaturedTagsController < Settings::BaseController
   before_action :authenticate_user!
   before_action :set_featured_tags, only: :index
   before_action :set_featured_tag, except: [:index, :create]
+  before_action :set_most_used_tags, only: :index
 
   def index
-    @featured_tag   = FeaturedTag.new
-    @most_used_tags = Tag.most_used(current_account).where.not(id: @featured_tags.map(&:id)).limit(10)
+    @featured_tag = FeaturedTag.new
   end
 
   def create
     @featured_tag = current_account.featured_tags.new(featured_tag_params)
     @featured_tag.reset_data
-    @featured_tag.save
 
-    redirect_to settings_featured_tags_path
+    if @featured_tag.save
+      redirect_to settings_featured_tags_path
+    else
+      set_featured_tags
+      set_most_used_tags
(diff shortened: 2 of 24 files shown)


MirceaKitsune:
Will this also work for the search bar? For instance you could use `#sometag@username@domain` to only list posts containing that tag from the user in cause. Additionally just `#sometag@domain` would be a very welcome syntax, allowing you to search for a tag on a specific instance only.

app/controllers/accounts_controller.rb
```diff
@@ -78,12 +79,15 @@ def no_replies_scope
     Status.without_replies
   end
 
+  def hashtag_scope
+    Status.tagged_with(Tag.find_by(name: params[:tag].downcase)&.id)
```
krainboltgreene: Not sure what the project's preference is, but I've been seeing a lot of people move from `#[]` to `#fetch()` since it tells the reader a lot more about the intent. For example, in this method we *EXPECT* `:tag` to be a key, thus `params.fetch(:tag)`, but if we didn't `params.fetch(:tag, nil)` and of course we get all the niceness of `#fetch`'s other default api: `fetch(:key) {dynamicvalue}`.

Either way, seems like what you want here could also be expressed as: `Status.tagged_with(Tag.where(name: params[:tag]).pluck(:id).first)`, though if `tagged_with` allows an array you can just drop `.first`.

app/controllers/settings/featured_tags_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+class Settings::FeaturedTagsController < Settings::BaseController
+  layout 'admin'
+
+  before_action :authenticate_user!
+  before_action :set_featured_tags, only: :index
+  before_action :set_featured_tag, except: [:index, :create]
+
+  def index
+    @featured_tag   = FeaturedTag.new
+    @most_used_tags = Tag.most_used(current_account).where.not(id: @featured_tags.map(&:id)).limit(10)
```
krainboltgreene: Probably want `@featured_tags.pluck(:id)` to avoid pulling all of those records into memory.

krainboltgreene: Also, depending on `Tag`'s database size, you probably want to do `WHERE NOT EXIST IN (SELECT 1 FROM "featured_tags" WHERE "featured_tags"."id" = "tags"."id")`, see: https://sqlperformance.com/2012/12/t-sql-queries/left-anti-semi-join

app/controllers/settings/featured_tags_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def create
+    @featured_tag = FeaturedTag.new(featured_tag_params)
+    @featured_tag.account = current_account
```
krainboltgreene: Not sure of current project practice, but you can also do this as `@featured_tag = current_account.featured_tags.new(featured_tag_params)`

app/controllers/settings/featured_tags_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @featured_tag.statuses_count = current_account.statuses.tagged_with(@featured_tag.tag).count
+    @featured_tag.last_status_at = current_account.statuses.tagged_with(@featured_tag.tag).maximum(:created_at)
+    @featured_tag.save
```
krainboltgreene: Probably want to `save!` so you can rescue from invalid records, yes?

app/controllers/settings/featured_tags_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    redirect_to settings_featured_tags_path
+  end
+
+  def destroy
+    @featured_tag.destroy
```
krainboltgreene: Probably want `destroy` so you get an exception you can rescue from if it fails, yes?

app/models/featured_tag.rb
```diff
@@ -0,0 +1,24 @@
...
+#
+
+class FeaturedTag < ApplicationRecord
+  belongs_to :account, inverse_of: :featured_tags, required: true
+  belongs_to :tag, inverse_of: :featured_tags, required: true
+
+  delegate :name, to: :tag, allow_nil: true
+
+  def name=(str)
+    self.tag = Tag.find_by(name: str.delete('#').mb_chars.downcase.to_s)
```
krainboltgreene: Downcase should only work on a string, so `.to_s` seems redundant?

db/migrate/20190203180359_create_featured_tags.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFeaturedTags < ActiveRecord::Migration[5.2]
+  def change
+    create_table :featured_tags do |t|
+      t.references :account, foreign_key: { on_delete: :cascade }
+      t.references :tag, foreign_key: { on_delete: :cascade }
```
krainboltgreene: These references should probably be `null: false`, yes?

app/models/featured_tag.rb
Gargron: mb_chars returns a different kind of string that I believe caused issues with ActiveRecord before

db/migrate/20190203180359_create_featured_tags.rb
Gargron: I don't believe any of them are. There is no appropriate default non-null value here.

app/controllers/settings/featured_tags_controller.rb
Gargron: `@featured_tags` is already loaded and it's expected to be a small array.

krainboltgreene: Can you point to where it's already loaded? I don't believe it is.

app/models/featured_tag.rb
krainboltgreene: Ah, okay, that kinda makes sense. Resolved!

krainboltgreene (approved):

PR #9780, opened 2019-01-10
Add attachment list fallback to public pages

### LINKED ISSUE #6714: Removal of media links makes media viewing more difficult
As per #6672, there is now no requirement at all to have media hyperlinks within the text of a toot.  It's nice and all that "ugly" interface elements are being removed, but this change means there is no way to view static images without using a browser that supports javascript, and it makes it more difficult to view media directly since there's no standard hyperlink to click on.

Sure sounds like an accessibility issue to me.

* * * *

ClearlyClaire:
Fixes #6714

I'm not too familiar with HAML, so I'm not sure if the fallback-as-a-block pattern makes a lot of sense.
Also, the code for rendering the list should probably be moved into another fragment or a helper.

Looks like this when the JS part hasn't loaded yet / if it doesn't load:
screenshot_2019-01-10 thib thib mastodev sitedethib com

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -69,8 +69,12 @@ module ApplicationHelper
     tag(:meta, content: content, property: property)
   end
 
-  def react_component(name, props = {})
-    content_tag(:div, nil, data: { component: name.to_s.camelcase, props: Oj.dump(props) })
+  def react_component(name, props = {}, &block)
+    if block.nil?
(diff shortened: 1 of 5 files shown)


Gargron:
Something is not right in the screenshot. Spacing to the icon as well as the color of the icon.

Gargron:
I think it should be possible to reuse the CSS of the AttachmentList component instead of redefining a copy.

ThibG:
Oops, updated it, it looks like this now:
screenshot_2019-01-11 thib thib mastodev sitedethib com

ykzts (approved):

PR #9796, opened 2019-01-13
Update Dockerfile for Ruby 2.6

shleeable:
I'd love some testing before this merged.

Planning on running this config once RC2 is released for testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM node:8.14.0-alpine as node
-FROM ruby:2.4.5-alpine3.8
+FROM node:8.15-alpine as node
+FROM ruby:2.6-alpine3.8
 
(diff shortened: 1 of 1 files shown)


Reverite:
@Gargron @ashleyhull-versent 

The Mastodon Docker instances of NodeJS & Ruby have been running at 10.14/15 & 2.6.0 respectively for 6 days, so if there's some sort of bug that might possibly cause us to revert to Node 8 or 2.4, I'm not getting it yet. Looks like the things that caused 2.5 to crash aren't in 2.6. I'm using this on a Mastodon instance with ~400 users.

So I'd actually suggest:

```
FROM node:10.15-alpine as node
FROM ruby:2.6.0-alpine
```

ykzts (approved):

Issue #687, opened 2017-03-26
[BUG] Cannot dismiss "drag to upload" dialog


yiskah:
If you accidentally drag anything (including a link) while your mouse is over a mastodon window, it will bring up the "drag and drop to upload" dialog. If it is not a valid file to upload (for instance, if i click a button and drag) or if after moving the object over the window you then do not drop the object, the dialog stays open and you are unable to interact with the page or dismiss the dialog unless you refresh the page or find a photo on your computer and drag it onto the window so it has something to upload

nclm:
Yes, this is quite annoying. I came here to open the same bug.

To illustrate the issue, two examples of cases when it happens: when dragging some text from Mastodon to look it up (which happens to me all the time), and when dragging a unrelated file purposelessly over Mastodon.

![ezgif-3-40fe195c53](https://cloud.githubusercontent.com/assets/923008/25064150/aee2f2e0-21f4-11e7-82e6-4406b5cbb34f.gif)

![ezgif-3-9296d2b824](https://cloud.githubusercontent.com/assets/923008/25064170/ee96d33e-21f4-11e7-89bf-b32d4de1d81a.gif)

Technowix:
This issue is still here on v1.2

fwenzel:
The easiest way to fix this is probably hooking up a keypress handler for Esc on the upload dialog. That won't auto-dismiss, but at least you can do something that's *not* reloading.

stephenburgess8:
@yiskah @nclm Dragging text should no longer trigger the upload modal. Even when dragging files, this should be triggered (less) often. I'm still able to reproduce this behavior in Firefox when dragging from the window into the URL area, because for some reason neither dragEnd nor dragLeave is triggered. ¯\_(ツ)_/¯ Since I wasn't able to completely get rid of it, I added the escape listener. I confirmed this works in Chrome, Safari, and Firefox. Let me know if there's any further changes you think would help.

PR #9797, opened 2019-01-13
make draggingOver state off correctly

hinaloe:
Sometimes, `.upload-area` could not close when dragging leave to out of window.

image

## How to reproduce
1. **drag and drop** *any* contents (Image, Text, and other) on Mastodon window (to same window).
1. Try drag and drop any file from any file manager to Mastodon window, and cancel (non drop, or drop in original or other location).

## Tested on
Chrome and Firefox (macOS, Linux)

**Edge and Safari** has the same problem reproducibility, ~~but it does not seem to work well.~~
-> I found the simple way.


hinaloe:
This seems to be related to #687

ykzts (approved):

PR #9806, opened 2019-01-14
Fix hashtag timeline on mobile

ClearlyClaire:
Instead of forcing 90vh for the scrollable part of the timeline's height as
in #9781 (which may be off), use `display: flex`, exactly like on desktop.

### DIFF
--- a/app/javascript/styles/mastodon/about.scss
+++ b/app/javascript/styles/mastodon/about.scss
@@ -1041,10 +1041,6 @@ $small-breakpoint: 960px;
 
     .scrollable {
       height: 400px;
-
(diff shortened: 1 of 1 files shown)


ThibG:
Note: not using `100vw` for the width causes a little of padding on the right, at least on Firefox.
That is because of the body's scrollbar. Using `100vw` for the width causes the body's scrollbar to show over the timeline's scrollbar, at least on Firefox, which is confusing.
It would be better to disable the body's scrollbar, but it requires more work.

screenshot_2019-01-14 parakeet

Gargron (approved):

PR #9811, opened 2019-01-14
Reject existing Follow in addition to sending a Block

ClearlyClaire:
Mastodon expects remote servers to remove follow relationships upon receiving
a Block. However, the spec only evokes Block activities in a C2S context, never
in a S2S context.

This PR, in addition to federating the Block, explicitly sends a Reject for any
affected follow relationship, which makes a bit more sense with regards to the
spec.

~~Marked Work in Progress as I have to check the generated Activity.~~

### DIFF
--- a/app/services/unfollow_service.rb
+++ b/app/services/unfollow_service.rb
@@ -20,6 +20,7 @@ class UnfollowService < BaseService
 
     follow.destroy!
     create_notification(follow) unless @target_account.local?
+    create_reject_notification(follow) if @target_account.local? && !@source_account.local?
     UnmergeWorker.perform_async(@target_account.id, @source_account.id)
     follow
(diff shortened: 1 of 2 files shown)


app/services/unfollow_service.rb
```diff
@@ -42,6 +43,13 @@ def create_notification(follow)
     end
   end
 
+  def create_reject_notification(follow)
+    # Rejecting an already-existing follow request
+    if follow.account.activitypub?
```
Gargron: Use `return unless` to avoid indentation

Gargron (changes requested):

ThibG:
(Addressed comments)

Gargron (approved):

PR #9815, opened 2019-01-15
Reduce chances of race conditions when processing deleted toots

ClearlyClaire:
I have noticed that sometimes, toots are not correctly deleted across instances (despite a direct Follow relationship).

One of those cases may be because of a race condition between the processing of `Create` and `Delete` when a toot is quickly created then deleted.

This PR reduces chances of such a race condition happening, but does not eliminate it. It should be possible to avoid this race condition completely by moving the `delete_arrived_first?` check into the locked block, and lock the `delete_later!` call as well, but I'm not sure about the performances impact.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -5,10 +5,12 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   CONVERTED_TYPES = %w(Image Video Article Page).freeze
 
   def perform
-    return if delete_arrived_first?(object_uri) || unsupported_object_type? || invalid_origin?(@object['id'])
+    return if unsupported_object_type? || invalid_origin?(@object['id'])
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
+        return if delete_arrived_first?(object_uri)
+
         @status = find_existing_status
 
         if @status.nil?
--- a/app/lib/activitypub/activity/delete.rb
+++ b/app/lib/activitypub/activity/delete.rb
@@ -21,10 +21,10 @@ class ActivityPub::Activity::Delete < ActivityPub::Activity
   def delete_note
     return if object_uri.nil?
 
+    delete_later!(object_uri)
+
     @status   = Status.find_by(uri: object_uri, account: @account)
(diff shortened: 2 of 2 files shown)


ThibG:
Added a commit which uses a lock, ensuring that a deleted status has either been processed or won't be processed at all. I am not too sure about the performance implications when processing `Delete`.

Gargron (approved):

app/lib/activitypub/activity/create.rb
```diff
@@ -5,10 +5,12 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
   CONVERTED_TYPES = %w(Image Video Article Page).freeze
 
   def perform
-    return if delete_arrived_first?(object_uri) || unsupported_object_type? || invalid_origin?(@object['id'])
+    return if unsupported_object_type? || invalid_origin?(@object['id'])
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
+        return if delete_arrived_first?(object_uri)
```
ThibG: CodeClimate complains, but I find it clearer to just return here. Or maybe there's something I don't understand about Ruby blocks?

app/lib/activitypub/activity/delete.rb
```diff
@@ -21,11 +21,13 @@ def delete_person
   def delete_note
     return if object_uri.nil?
 
+    RedisLock.acquire(lock_options) do |lock|
+      delete_later!(object_uri)
```
ThibG: CodeClimate complains about us not using `lock`, but we can't leave it out.
We don't use it because we want to call `delete_later!` even in the unlikely case where we couldn't acquire the lock, it's not really an issue.

Gargron: We can totally leave it out though? Or just name it `_lock`

app/lib/activitypub/activity/create.rb
Gargron: I believe the code does what you want it to do, but in most cases return from a block is a mistake.

app/lib/activitypub/activity/delete.rb
ThibG: We can't leave it out because Redislock checks for that parameter (see https://circleci.com/gh/tootsuite/mastodon/39022?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I will name it `_lock`.

PR #9821, opened 2019-01-16
Expose the expand_spoilers user setting via the credentials API

Tak:


connyduck:
This is an important feature for client apps, please merge it!

Gargron:
We need a preferences API, sticking more and more properties into the source attribute feels increasingly unsustainable.

Tak:
Ok, has there been any planning for that, or do you have guidance about an implementation you would approve?

Gargron:
I have a vague idea of what it would look like. Something like /api/v1/preferences returning a hash of key/values. I imagine the keys would be namespaced similarly to OAuth scopes, via colons, and the names of the properties would not necessarily correspond to their in-database name, rather, we should aim for more understandable names and try to avoid underscores if possible.

PR #9823, opened 2019-01-16
Use summary as summary for converted ActivityPub objects

### LINKED ISSUE #8609: Articles summary displayed as CW
I recently implemented subtitles for articles in Plume. They are federated using the `summary` field, which makes Mastodon display them as CW. Would it be possible to make a special case for `Article`s to display the `summary`/CW between the title and the link to the content? If the person wants to read what is behind the CW, they will click the link instead of the current button, and the presentation will be nicer and more coherent.

* * * *

Gargron:
For non-Note objects, `summary` has a slightly different semantic meaning, and it doesn't make sense to use it as a content warning hiding the title and link to the article. So, for converted object types, display the title, then summary, then the link to the article.

Fix #8609

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -61,7 +61,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
         account: @account,
         text: text_from_content || '',
         language: detected_language,
-        spoiler_text: text_from_summary || '',
+        spoiler_text: converted_object_type? ? '' : (text_from_summary || ''),
         created_at: @object['published'],
         override_timestamps: @options[:override_timestamps],
         reply: @object['inReplyTo'].present?,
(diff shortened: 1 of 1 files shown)


ClearlyClaire (approved):

ThibG:
LGTM. This means such objects can't have a Content Warning though. Maybe use the title as a CW?

Gargron:
I see no good reason to make people click show more, since both the title and the summary of a blog post are typically oriented towards being displayed in public view, while what would be actually hidden by the CW is the main text which is not displayed anyway (link).

app/lib/activitypub/activity/create.rb
```diff
@@ -61,7 +61,7 @@ def process_status_params
         account: @account,
         text: text_from_content || '',
         language: detected_language,
-        spoiler_text: text_from_summary || '',
+        spoiler_text: converted_object_type? ? '' : (text_from_summary || ''),
```
nightpool: this should be object[:sensitive] or something equivalent, so converted object types can still opt in to CWs

nightpool:
aren't there other converted object types besides articles though?

Gargron:
If a video or image is marked sensitive, it will be hidden. Since they don't have other text, it doesn't make sense to to make people click "show more" for nothing.

PR #9824, opened 2019-01-16
Fix public hashtag timeline width on mobile, fix scrollbar width compensation

ClearlyClaire:
Fixes regression regarding timeline width on very narrow screens introduced in #9806.

Also computes scroll width automatically to avoid discrepancies when compensating for scrollbars.


Gargron:
Consider #9822

ThibG:
It doesn't conflict with it, and fixes other pages as well (although the window/screen has to be really narrow for this to be a problem).

Gargron (approved):

PR #9830, opened 2019-01-17
Add tombstones for remote statuses

ClearlyClaire:
Record deleted remote statuses so that they can't be re-injected.

### DIFF
--- a/app/lib/activitypub/activity/delete.rb
+++ b/app/lib/activitypub/activity/delete.rb
@@ -21,8 +21,9 @@ class ActivityPub::Activity::Delete < ActivityPub::Activity
   def delete_note
     return if object_uri.nil?
 
-    RedisLock.acquire(lock_options) do |_lock|
-      delete_later!(object_uri)
+    unless invalid_origin?(object_uri)
+      RedisLock.acquire(lock_options) { |_lock| delete_later!(object_uri) }
+      Tombstone.find_or_create_by(uri: object_uri, account: @account)
     end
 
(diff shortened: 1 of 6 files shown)


app/lib/activitypub/activity/delete.rb
```diff
@@ -21,8 +21,11 @@ def delete_person
   def delete_note
     return if object_uri.nil?
 
-    RedisLock.acquire(lock_options) do |_lock|
-      delete_later!(object_uri)
+    unless invalid_origin?(object_uri)
+      RedisLock.acquire(lock_options) do |_lock|
+        delete_later!(object_uri)
```
Gargron: Maybe use inline block here since it's a single line.

Gargron: Maybe use inline block here since it's a single line.

app/services/activitypub/process_account_service.rb
```diff
@@ -33,6 +33,7 @@ def call(username, domain, json, options = {})
 
     after_protocol_change! if protocol_changed?
     after_key_change! if key_changed? && !@options[:signed_with_known_key]
+    clear_tombstones! if key_changed?
     unless @options[:only_key]
```
Gargron: Empty newline before next block please

Gargron (changes requested):

ThibG:
(Addressed the comments)

Gargron (approved):

PR #9851, opened 2019-01-18
Add `visibility` param to reblog REST API

### LINKED ISSUE #7322: Ability to boost to followers only ("private boost")
**Currently, we can toot privately, and boost publicly. But not boost to follower-only (a kind of "private boost").**
This has a **privacy advantage**: you don't show publicly that you boosted [some toot] with [some content], but only to your followers.
But to fully hide this public data, it should also not be counted in the list of people who boosted the post (or just not shown in the list, but counted ?). If not, it would just hide it from public profile page (that might be interesting too ?).

Concerning the **UI**, it could be implemented as a drop-down menu on the boost button, for instance.

*****

**The current workaround** is to reply (privately, with or without mentioning the author) to the toot, saying "Hey, check this toot here ↑, that talks about [this topic]". It's doable, but not convenient.
Or to boost a post, and "deboost" it (~cancel the boost). It works, but only for people currently connected, and if they doesn't refresh the page. And it doesn't show who boosted the toot (and create a bit of confusion, when a new person's toot integrate the TL without any boost origin mention).
*****

This might be some kind of extension of https://github.com/tootsuite/mastodon/pull/6157 (boosting own-private toots).

Or to this proposal: https://github.com/tootsuite/mastodon/issues/422 (lists, mutuals only, and eliminating the need for multiple accounts)
For instance, in addition to posting to a specific list of users, we could boost only for them.
In addition to the privacy advantage (for instance: toots boosted only to family/friends, instead of all followers), **it's also a way to publish/boost toots to groups based on their interest**. Let's say that you're talking to 2 groups of people, one interested in topic X, the other in topic Y. You could publish content related to X to one group, and related to Y to the other, without "spamming" them with unrelevant content.

* * * *

Note: I've searched if there was another issue talking about this, but I found more general topics about post/boost privacy, like https://github.com/tootsuite/mastodon/issues/5723 (that doesn't really talk about the same thing) or https://github.com/tootsuite/mastodon/issues/422 (as said above).

Gargron:
Fix #7322 

- Allow setting visibility for each reblog
- Changes the default to reblog with the default chosen visibility setting instead of public
- Use async worker for creating reblog notification to improve performance

### DIFF
--- a/app/controllers/api/v1/statuses/reblogs_controller.rb
+++ b/app/controllers/api/v1/statuses/reblogs_controller.rb
@@ -9,7 +9,7 @@ class Api::V1::Statuses::ReblogsController < Api::BaseController
   respond_to :json
 
   def create
-    @status = ReblogService.new.call(current_user.account, status_for_reblog)
+    @status = ReblogService.new.call(current_user.account, status_for_reblog, reblog_params)
     render json: @status, serializer: REST::StatusSerializer
   end
 
@@ -31,5 +31,9 @@ class Api::V1::Statuses::ReblogsController < Api::BaseController
 
   def status_for_destroy
     current_user.account.statuses.where(reblog_of_id: params[:status_id]).first!
+  end
+
+  def reblog_params
+    params.permit(:visibility)
   end
 end
--- a/app/services/reblog_service.rb
+++ b/app/services/reblog_service.rb
@@ -7,8 +7,9 @@ class ReblogService < BaseService
   # Reblog a status and notify its remote author
   # @param [Account] account Account to reblog from
   # @param [Status] reblogged_status Status to be reblogged
+  # @param [Hash] options
   # @return [Status]
-  def call(account, reblogged_status)
+  def call(account, reblogged_status, options = {})
     reblogged_status = reblogged_status.reblog if reblogged_status.reblog?
 
     authorize_with account, reblogged_status, :reblog?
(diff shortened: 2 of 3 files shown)


ClearlyClaire (dismissed):

app/controllers/api/v1/statuses/reblogs_controller.rb
```diff
@@ -32,4 +32,8 @@ def status_for_reblog
   def status_for_destroy
     current_user.account.statuses.where(reblog_of_id: params[:status_id]).first!
   end
+
+  def reblog_params
+    params.permit(:visibility)
```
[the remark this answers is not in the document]
Gargron: Agree direct and limited make no sense. Limited might make sense if explicit group addressing is added to the API, but right now, it should be excluded. Amended. Public/unlisted is fine. Us excluding stuff from the PTLs is a readability decision rather than a spec thing. A different platform might want to show all public boosts in a stream.

ykzts (approved):

app/services/reblog_service.rb
```diff
@@ -17,7 +18,7 @@ def call(account, reblogged_status)
 
     return reblog unless reblog.nil?
 
-    reblog = account.statuses.create!(reblog: reblogged_status, text: '')
+    reblog = account.statuses.create!(reblog: reblogged_status, text: '', visibility: options[:visibility] || account.user&.setting_default_privacy)
```
ThibG: On the one hand, I'm still uneasy with that change in behavior when we don't have UI to tell the visibility of a boost nor UI to specify it when boosting.
On the other hand, I think it makes more sense than the current behavior…

Gargron: Posted a poll: https://mastodon.social/@Gargron/101749616287239685

lapineige:
Sorry if it's not the place to ask, but will only be activated for new boosts in >=2.8, right ?
It doesn't change the status of previous (<2.8) boosts, correct ?

Gargron:
Yes, correct

lapineige:
And if we want to boost publicly some toot, we have to change the setting, boost, switch back to private visibility by default, and that would work ?

PR #9854, opened 2019-01-18
Update to current Node LTS for Dockerfile

ladyisatis:
Ruby will use latest Alpine release, to match Node using the latest. This will need testing, but I have been using Node 10 for about 2 weeks, and uWS will build on the LTS.

Implementing what I commented in the PR on #9796. My current thinking is that this needs testing but nothing is breaking with streaming, and Webpack seems to respond just fine. There will also be no breaking changes to the Node 10 LTS (semantic versioning).

(Also, Ruby 2.6 itself seems stable on Docker, I'm not getting segfaults still. Yay!)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM node:8.15-alpine as node
-FROM ruby:2.6-alpine3.8
+FROM node:10-alpine as node
+FROM ruby:2.6-alpine
 
 LABEL maintainer="https://github.com/tootsuite/mastodon" \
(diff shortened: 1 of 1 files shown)


Gargron:
To confirm, uWS builds on Node 10? I heard this was a problem previously...

Reverite:
Answered on Discord, but: yeah, it builds successfully here via Docker Hub.

Gargron:
The structure of the Dockerfile has been changed significantly so it's best to re-open a new PR.

PR #9856, opened 2019-01-18
Allow multiple files upload through web UI, including drag & drop

### LINKED ISSUE #5556: Feature request: Ability to upload more than one image at once
As far as I'm aware, it's only possible to upload one image at a time at the moment.

I don't know if it's possible, but I'd love to be able to drag and drop more than one image at a time into the window to upload, or to select more than one image in the file-picker when you click the camera icon. :)

* * * *

- [Y] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [N/A] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

tmm576:
Maintains maximum limit of four files uploaded. Possible opportunity to abstract to instance setting.

Fix #5556

### DIFF
--- a/app/javascript/mastodon/actions/alerts.js
+++ b/app/javascript/mastodon/actions/alerts.js
@@ -22,7 +22,7 @@ export function clearAlert() {
   };
 };
 
-export function showAlert(title, message) {
+export function showAlert(title = messages.unexpectedTitle, message = messages.unexpectedMessage) {
   return {
     type: ALERT_SHOW,
     title,
@@ -44,6 +44,6 @@ export function showAlertForError(error) {
     return showAlert(title, message);
   } else {
     console.error(error);
-    return showAlert(messages.unexpectedTitle, messages.unexpectedMessage);
+    return showAlert();
   }
 }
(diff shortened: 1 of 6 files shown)


Gargron:
I have to ask, does this actually work? AJAX requests are asynchronous so making them from a loop does not guarantee sequential execution, and upload progress is a one-for-all state subtree so multiple concurrent requests for different files would be overwriting it all the time

tmm576:
@Gargron This update keeps the single upload progress bar, but is calculated on the total of all files being uploaded.

Gargron (approved):

Gargron:
Perhaps it would make sense to mark the file input as allowing multiple files, so that multiple files can be selected from the normal dialog, as well? Then you could put "Fix #5556" in your commit message

tmm576:
Fix for #5556 
I added in some user feedback for file limits so wanted to get your opinion on using alerts this way from the action.

Gargron:
Yes, I believe it is appropriate :+1:

PR #9871, opened 2019-01-20
i18n: Update Japanese translations

mayaeh:
I translated Japanese untranslated parts and more.
Several translations are applied from Weblate's proposal.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -106,7 +106,6 @@ en:
       enabled: Enabled
       feed_url: Feed URL
       followers: Followers
-      followers_local: "(%{local} local)"
       followers_url: Followers URL
       follows: Follows
       header: Header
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -106,6 +106,7 @@ en:
       enabled: Enabled
       feed_url: Feed URL
       followers: Followers
+      followers_local: "(%{local} local)"
```
Gargron: This is not used

Gargron (changes requested):

config/locales/en.yml
mayaeh: It is added by executing "bundle exec i18n-tasks add-missing".
Is it safe to delete it?

Gargron (approved):

PR #9879, opened 2019-01-20
Add WebP support

### LINKED ISSUE #8165: webp support
Supporting webp files would allow people to send and download images faster and those who store pictures in this format wouldn't be forced to convert it before posting.

acid-chicken:
Resolves #8165


JanJastrow:
@mayaeh the next version of Firefox 56 will likely include WebP support.

app/models/media_attachment.rb
```diff
@@ -25,10 +25,10 @@ class MediaAttachment < ApplicationRecord
 
   enum type: [:image, :gifv, :video, :unknown]
 
-  IMAGE_FILE_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif'].freeze
+  IMAGE_FILE_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif', '.webp'].freeze
```
Gargron: How about this: Keep the support but remove the changes to the tooltip. Then it's future-proof without encouraging users to upload images that still aren't supported by Safari/iOS

Gargron (changes requested):

app/models/media_attachment.rb
acid-chicken: @Gargron Thank you for the review! I've removed the changes at 993e3e059dde767a1fc3795434823e47a2fba087. Is this alright with you?

Gargron (approved):

PR #9894, opened 2019-01-21
Fix scheduled toot with media immediately creating a toot

ClearlyClaire:


mayaeh:
There seems to be no problem in the range I tried.
I think that [this problem](https://github.com/tootsuite/mastodon/issues/9893#issuecomment-456097822) was fixed.
Thank you !

Gargron (approved):

PR #9897, opened 2019-01-21
Make displaying application used to toot opt-in

ClearlyClaire:
This PR introduces a setting to control whether an the application used to post toots is ~~stored and~~ displayed, and makes it opt-in (thus changing the default behavior).

The reasoning for this is that this information has little legitimate use, but may be used to get private information about the user (are they posting from their computer or a phone app? Do they use a very little-used app? If so, can we identify them across accounts?)

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -48,6 +48,7 @@ class Settings::PreferencesController < Settings::BaseController
       :setting_theme,
       :setting_hide_network,
       :setting_aggregate_reblogs,
+      :setting_show_application,
       notification_emails: %i(follow follow_request reblog favourite mention digest report),
       interactions: %i(must_be_follower must_be_following)
     )
--- a/app/lib/user_settings_decorator.rb
+++ b/app/lib/user_settings_decorator.rb
@@ -32,6 +32,7 @@ class UserSettingsDecorator
     user.settings['theme']               = theme_preference if change?('setting_theme')
     user.settings['hide_network']        = hide_network_preference if change?('setting_hide_network')
     user.settings['aggregate_reblogs']   = aggregate_reblogs_preference if change?('setting_aggregate_reblogs')
+    user.settings['show_application']    = show_application_preference if change?('setting_show_application')
   end
 
   def merged_notification_emails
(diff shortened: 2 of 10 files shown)


Gargron:
I have disagreements for multiple reasons:

- Malicious app that modifies preference to hide the source of malicious messages
- Makes harder to implement any mute-by-source functionality in the future
- App developers depend on this for app discovery (to some degree)

ThibG:
Hm, the first two reasons kind of make sense, but I'm really skeptical about the third point.
I do think that exposing the application used is a liability regarding privacy.

ThibG:
I changed it so that the information is still stored and displayed to the author. I believe this addresses all your concerns.

Gargron (approved):

rinsuki:
@ThibG Why did you change default behavior?
I think, this should be an opt-out...

ThibG:
I changed it because I think most users do not care about displaying such information, and I think we should err on the side of not leaking info that might turn out to be private, at least by default. Furthermore, nothing currently tells users that such information will be displayed.


Le 6 février 2019 17:31:17 GMT+01:00, rinsuki <notifications@github.com> a écrit :
>@ThibG Why did you change default behavior?
>I think, this should be an opt-out...
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/tootsuite/mastodon/pull/9897#issuecomment-461089490

-- 

Gargron:
I didn't realize you made it default to false after my concerns. I agree with @rinsuki the default behaviour should not be changed.

>I think most users do not care about displaying such information

This is true, but I think there are generally a lot of ideological questions users do not care about. Whether you are using Tusky or Mastalab does not endanger anyone in the majority of cases, and we need to help the app ecosystem, because users generally don't think about an app ecosystem after they've found an app that works for them. Not everyone leaves a 5-star review on the app store, so do you think they will look for an option they do not know is there?

rinsuki:
I think that this change will cause malfunction of https://distsn.org/mastodon-apps.html and some sites.

I understand that you are concerned about privacy, but I think that changing default behavior is not favorable.

As an alternative to opt-in, I propose to display some message in "/oauth/authorize".
e.g. "Note: anyone can see the application you used to toot. You can overwrite this behavior in settings."

PR #9898, opened 2019-01-22
Added underlines to links in toots

aardrian:
Link text color is #f1ebff, surrounding text is #ffffff. They have a contrast ratio of 1.2:1. Not only is hard to see links that are so close in color to the text, they also require a pointing device (mouse) to quickly scan for links. Instead of having to meet a 4.5:1 contrast ratio, adding the underline will allow this to pass a WCAG check.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -670,7 +670,7 @@
 
   a {
     color: $secondary-text-color;
-    text-decoration: none;
+    text-decoration: underline;
 
     &:hover {
       text-decoration: underline;


rugk:
IMHO, I'd rather choose a better color than adding underlines, as these do aesthetically not look that nice. And the theme could work better with a little more contrast.

Gargron:
Links in toots practically always look like example.com/foo, not like other text, so the homogenity of color and style with other text is a feature, not a bug.

aardrian:
This reference (which is a collection of references) may be useful: [On Link Underlines](http://adrianroselli.com/2016/06/on-link-underlines.html)

@rugk 

>  I'd rather choose a better color

That is an option, but you will want to target a 4.5:1 contrast ratio with the background and a 3:1 contrast ratio with the surrounding text.

> these do aesthetically not look that nice

That is a matter of opinion. I think they look fine with underlines. More importantly, I can _use_ it with underlines. I think usability trumps aesthetics.

@Gargron 

> so the homogenity of color and style with other text is a feature, not a bug.

I am not sure I understand. Are you saying that the link text matching the surrounding text is a feature?

Gargron:
Yes. Readers do not have to guess whether a piece of text is a link by color or underline, because it never looks like other text, it looks like a URL. The slight color difference is a minor help in skimming, but it is not fundamental in my opinion. Keeping visual clutter down is important.

rugk:
> This reference (which is a collection of references) may be useful: On Link Underlines

Thanks, that is actually interesting, because it states that (G182) this is actually a visual clue to make links identifiable even without color vision. Thus, it addresses a slightly different case than the contrast. It also explains, you can use other clues, like a different font size or so.

aardrian:
@Gargron 

> Yes. Readers do not have to guess whether a piece of text is a link by color or underline, because it never looks like other text, it looks like a URL.

The is demonstrably not true. I tooted a11y.reviews yesterday. No sub-page, and it does not use a common TLD.

> The slight color difference is a minor help in skimming, but it is not fundamental in my opinion. Keeping visual clutter down is important.

The slight color difference is so slight as to be imperceptible except under ideal conditions (no glare on screen, good screen, good vision, etc.).

What you perceive as visual clutter is making it harder for some users (or all users in some conditions) to easily visually identify links.

Now that you know the interface is harder for some users (which is also backed up by contrast ratios and best practices as outlined by WCAG), are you comfortable telling those users that your preference for "less clutter" overrides their ability to use it?

@rugk 

> Thus, it addresses a slightly different case than the contrast.

It does. Which is why I added underlines. Instead of solving for contrast, underlines address it better for all sighted users.

> It also explains, you can use other clues, like a different font size or so.

Yup, but after years of testing this, underlines work best as they result in the least page reflow and weird spacing.

scottaohara:
Regarding all the above, I'd definitely prefer underlines on links.  it's by far the easiest way to differentiate them from other text.

mayaeh:
Even now, when you hover the mouse over the link, the underline will be displayed.
Why is not it enough?

nightpool:
@scottaohara i don't have any dog in this fight, since I use a custom theme with a much more noticeable link contrast (and there is a high-contrast theme that I believe is included by default in mastodon for those that would like to use it), but it's absolutely false to say that there are no focus styles for keyboard users when it comes to links:

image

aardrian:
@mayaeh 

> Even now, when you hover the mouse over the link, the underline will be displayed.
Why is not it enough?

Per my statements above:

- not everyone is a mouse user (touch screen users, keyboard users, power users, people who lost their mouse to their cat)
- per WCAG, another affordance is needed.

And @scottaohara beat me to it.

I am going to attach this image, however, showing how non-standard TLDs, slash-separated lists of items, and non-linked URLs are confusing and _require_ me to go tapping or mousing or tabbing to see what is what.

Oh, I also kinda came to Mastodon because Twitter has been so historically crap at supporting accessibility in any meaningful way. I would like to see it do better, and since I can contribute I am doing just that. Trying to make it better.

screenshot_20190122-203031

aardrian:
@mayaeh 

> but it's absolutely false to say that there are no focus styles for keyboard users when it comes to links:

There is no `:focus` style. You can see it in the inspector or the CSS files. What you are seeing is the browser's default focus ring. So Scott is demonstrably correct, but you have keyed into a common misunderstanding about browser defaults versus, well, doing better than browser defaults.

nightpool:
@aardrian I'm not sure where you took that screenshot, but it seems to be using a custom theme. Links look like this on mastodon:

image

This isn't great, and I agree it should probably be changed, but it's much better then the artificially darkened screenshot you posted.

> What you are seeing is the browser's default focus ring

I know exactly what I'm seeing. Whether the style is applied by the user-agent or by the application, there is still a style that's applied on `:focus`. In fact, nearly every accessibility guideline you will ever read will tell you not to override the brower's default `:focus` styles, since the primary goal is to be consistent across different applications. Framing it as not doing "better" then the browser defaults is very dangerous and misleading

Gargron:
There is a high-contrast theme in Mastodon.

mayaeh:
@scottaohara @aardrian I was forgotten about touch screen.
umm... I am thinking about the importance of making the link easier to distinguish.

aardrian:
@nightpool

The screen shot is off my mobile, where I mostly read Mastodon. I am using the Toot Cafe site theme. I have also browsed other Mastodon instances with different themes. It's part of why I suggested an underline instead of trying to find every color combination for every theme that meets WCAG minimums.

> In fact, nearly every accessibility guideline you will ever read will tell you not to override the browser's default :focus styles, since the primary goal is to be consistent across different applications. Framing it as not doing "better" then the browser defaults is very dangerous and misleading

I would like to see these guidelines you reference. Most of the accessibility practitioners I know (including me) recommend against leaving the defaults as they can be difficult to see on some backgrounds.

Further reading: [Avoid Default Browser Focus Styles](http://adrianroselli.com/2017/02/avoid-default-browser-focus-styles.html)

aardrian:
@Gargron 

> There is a high-contrast theme in Mastodon.

It is not sufficient because it also suffers from the same contrast issues and lack of underline. I attached a screen shot from my mobile using the high contrast theme.

screenshot_20190122-204937

nightpool:
G165 specifically addresses using the default focus indicator, and is a sufficient technique for WCAG-AA:

> [G165: Using the default focus indicator for the platform so that high visibility default focus indicators will carry over](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G165)

anyway, @aardrian we should

scottaohara:
@nightpool I'll be more clear, there are no defined `:focus` styles.

What you showed was a screen shot from what I assume is Chrome. 

Here is a screen shot showing a focus outline from Firefox on macOS.  Which is what happens when you don't define custom focus styles on heavily styled (dark theme) web documents/applications

screen shot from firefox on macOS



> In fact, nearly every accessibility guideline you will ever read will tell you not to override the browser's default :focus styles, since the primary goal is to be consistent across different applications. Framing it as not doing "better" then the browser defaults is very dangerous and misleading

>>I would like to see these guidelines you reference. Most of the accessibility practitioners I know (including me) recommend against leaving the defaults as they can be difficult to see on some backgrounds.

I would also be interested to see the documentation you're referring to that says to not change default focus styles.  Also being an accessibility practitioner, I'd be interested in knowing of these documents and then reaching out to these people to ask them to stop providing misleading information...

nightpool:
WCAG G165 specifically addresses using the default focus indicator:

> [G165: Using the default focus indicator for the platform so that high visibility default focus indicators will carry over](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G165)

They address the concerns you have about visibility but say that it's sufficient for users to be able to customize the default focus indicator using assistive technology. But, again, discussions of the focus indicator are off topic. 

It does look like there's a bug in the high contrast theme here. We should fix the high-contrast theme.

aardrian:
@nightpool 

> G165 specifically addresses using the default focus indicator, and is a sufficient technique for WCAG-AA:

You are correct. I forget about that because, frankly, nobody who actually does this stuff pays attention to that one. In fact, many actively counsel against it (see my link above). There are a few Sufficient Techniques that (since WCAG 2.0 was published) we have found to not be truly sufficient after a decade of testing. It's one of the reasons automated accessibility checkers are meh and trying to only meet WCAG at the minimum is frustrating to watch.

That being said, it doesn't address the issue I raised as that only satisfies for when a control has focus, not conveying its affordance (clickability in this case) before that interaction.

> It does look like there's a bug in the high contrast theme here. We should fix the high-contrast theme.

Agreed.

In the meantime, can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

nightpool:
I would prefer to solve this with high-contrast links rather then an underline. Specifically I would prefer to solve this with high-contrast links in both the main style and the high-contrast styles—preferably with a 3:1 contrast for the main style and a higher contrast for the high-contrast styles.

If we do choose to go the underline route we should specifically exclude hashtags and mentions, since those are already delineated by a sigil and should not be set off from running text in the same way that links generally can be.

Gargron:
>So can't you just include this change (underline) in the hard-contrast theme?

That was my point as well, I think that is the best solution.

aardrian:
> > So can't you just include this change (underline) in the hard-contrast theme?
>
> That was my point as well, I think that is the best solution.

I am all for seeing the high contrast theme actually be high contrast.

But the issue I have raised is for links in the main theme / all themes. They all either need an underline (easiest) or better contrast (trickier because each theme uses different colors).

Telling users who cannot see links to switch to the high-contrast theme still does not address all use cases. It assumes that when in the sun or using the phone dimmed at night that users know there is a high-contrast theme and can easily get into the preferences to change it.

Again, are they are any arguments against this other than aesthetic preferences of the project maintainers?

I think I have identified plenty of cases for adding underlines. They are grounded in WCAG, US case law, international law, best practices, years of user research, and usability.

kleinfreund:
As a user, I should not need to understand URI syntax in order to discover links. I should be able to rely on visual or acoustic cues ensuring the discoverability of interactive elements. One of the seven fundamental principles of design as stated in “The Design of Everyday Things” by Don Norman says:

“Discoverability. It is possible to determine what actions are possible and the current state of the device.” (Source: Donald A. Norman, 2013. The Design of Everyday Things. Page 72. Basic Books)

Several reasons for why the discoverability of links in the default theme cannot be ensured have been named. In my opinion, there haven’t been arguments against underlined links which would outweigh these reasons in importance. Maintaining a clean aesthetic is *not* more important than ensuring the operability of a web page. An interactive element that one cannot perceive is not operable.

mayaeh:
Underline seems to apply also to the hashtag and the name of the reply destination.
Umm... I can not judge whether this is correct.

_2019-01-24_11_55_30_re
Larger size is [here](https://taruntarun.net/media_files/Mastodon/media_attachments/files/000/315/172/original/0920a7f5f70380a6.png)

kleinfreund:
> Underline seems to apply also to the hashtag and the name of the reply destination.
> Umm... I can not judge whether this is correct.

That’s what I would expect. A tweet’s content is subject to change and as such, you don’t always know which elements are interactive. Again, the fundamental question at hand is “Do I know whether a thing is interactive?”. The answer to this question should be clear and not based on guessing. Underlining links allow a very clear answer to this question.

Gargron:
Links on Twitter do not have underlines either (albeit a more distinct color). I believe links on Facebook and Instagram do not have underlines, either. As it stands, underlines are visually unappealing and would *most definitely* put Mastodon at a further disadvantage compared to those platforms (e.g. "Twitter looks more polished than Mastodon"). I believe that adding underlines to the high-contrast theme which is aimed at people with impaired vision would solve the issue.

rugk:
> Adding underlines at the very least opens the platform up to 10% more market share (using global and US disability stats),

Wherever you get that stat from. I only get 2.3% in the US on the linked site [as per this PDF, page 5](http://www.disabilitystatistics.org/StatusReports/2017-PDF/2017-StatusReport_US.pdf?CFID=19464425&CFTOKEN=8a7d2af4c19e576c-2B7D73B6-C311-0FD6-4E6A7FBE4EADCE24) (2017). However, this includes all visual disabilities, this change here is, as we noticed, is only relevant to people without color vision, which is not separately listed.
As such, please choose your arguments wisely and back them up by (real) facts.

aardrian:
@rugk 

> As such, please choose your arguments wisely and back them up by (real) facts.

My fault. I pointed you to a resource that uses self-reported data from surveys, which is traditionally under-reported (folks who wear glasses or use a magnifying glass in a restaurant do not generally consider themselves vision impaired, for example).

@kleinfreund pointing you to WHO is better (and what we generally use with clients and in business plans anyway) because it looks at medical studies and patient care. Sadly, I was not aware I needed to be a trained physician to convince you that there a lot of people with low vision. I kind of figured the presence of WCAG would be proof in itself.

Related, you can also read how the W3C's Web Accessibility Initiative came up with the contrast ratio numbers and how good contrast is more than a color blindness issue in [Understanding Success Criterion 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) (along with the math they use for the ratio). I excerpted one part to demonstrate how these numbers account for vision that is more common than the self-reported numbers in the Cornell info:

> The contrast ratio of 4.5:1 was chosen for level AA because it compensated for the loss in contrast sensitivity usually experienced by users with vision loss equivalent to approximately 20/40 vision.

Not all people with 20/40 vision even wear glasses and they are still allowed to drive in the US. Perhaps that can drive home (pun intended) just how many people this change benefits.

Can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

Heydon:
@mayaeh 

> Even now, when you hover the mouse over the link, the underline will be displayed.
Why is not it enough?

Because not everyone uses a mouse, and if you can't see what is a link and what isn't you don't know where to hover.

Gargron:
**We could change the color of the links to be more distinct from other text.**

Notice that on GitHub as well links do not have underlines. There are designs out there where underlines look good, but it depends on the design. Platforms with similar information density as Mastodon, such as Twitter, TweetDeck, Facebook, Instagram, YouTube, Discord and even GitHub do not mark links with underlines until hovered.

Accessibility is important, which is why Mastodon does not play GIFs by default and why it has hotkey based navigation and screen reader support. Knowing that a particular string in a toot is a clickable link *through an underline* is not critical: As I mentioned before, links follow a syntax, links are hoverable, some links have large previews, they are focusable by screen readers. And we can improve the color.

P. S. Please do not dismiss the importance visual attraction and clean design. Only developers and people with impairments typically consider accessibility as a topic; the average person won't think "this is ugly, but I respect why they had to do that", the average person will just think "this is ugly"

kleinfreund:
> […] links follow a syntax,

Irrelevant. URI syntax is unknown to the general user and cannot be relied upon for identifying an interactive element.

> links are hoverable, […]

Irrelevant. Not all users use a mouse. Sometimes, I even use my phone.

---

But all that is besides the point. There are users who, due to visual impairments, are unable to distinquish between UI elements by color alone. This point is being ignored.

> Please do not dismiss the importance visual attraction and clean design.

Visual attraction is not useful if an interactive element cannot be interacted with. Then, it is just attractive and serves no other purpose than to please the eye and with that, we’re not even talking about all eyes. As for clean design, there has yet to be substantial proof as to how underlined links are opposed to the goal of achieving clean design. Also, the term design should not be confused with the term aesthetics. They are different.

aardrian:
@Gargron 

> We could change the color of the links to be more distinct from other text.

Yes, you can. If you want to guarantee the necessary contrast requirements and take that extra effort, you can do that. Per my [note above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-456419500), underlines are easier and this PR already has the needed change.

> Notice that on GitHub as well links do not have underlines.

I actually filed an issue with some of the GitHub people a year or two ago, but there is no official repo. Just as I also created an [accessible contributions chart](http://adrianroselli.com/2018/02/github-contributions-chart.html) and submitted it to a GitHub person and again to Microsoft's accessibility team after the acquisition.

But per [another comment of mine above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457261838), let's not argue for the documented accessibility failures of another platform as an excuse to also be less accessible.

> As I mentioned before, links follow a syntax, links are hoverable, some links have large previews, they are focusable by screen readers. And we can improve the color.

See my [comment above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-456632998) about how hovering is useless for touch or keyboard users. Previews are useless for multiple links. You can improve the color; see my comment at the top of this very reply.

> P. S. Please do not dismiss the importance visual attraction and clean design.

I have not. If you consider yourself a designer then rise to the challenge of the design constraints I have outlined above and prove it by creating a better underline style or coming up with all the needed colors across themes that match the required contrast ratios. I refer you to this, [which I linked above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-456394924): [On Link Underlines](http://adrianroselli.com/2016/06/on-link-underlines.html)

> Only developers and people with impairments typically consider accessibility as a topic; the average person won't think "this is ugly, but I respect why they had to do that", the average person will just think "this is ugly"

Once again I challenge you to prove that. I also challenge you to define what is "average".

You have only raised arguments that I or others have already addressed. You have not raised any new information to suggest this is demonstrably a _bad_ idea. You cannot even prove it without deploying it and measuring it.

Can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

kleinfreund:
> coming up with all the needed colors across themes that match the required contrast ratios

This should be emphasized. Don’t undererstimated how hard that task is if you are to support multiple themes.

mayaeh:
I think that it is also necessary to make it easy for anyone to distinguish between text and link.
However, I think the main theme does not need to make the link stand out beyond the main text.

By the way, underline is also added to hashtags and reply names.
I think the link is not so easy to distinguish because it is full of underline.

I think that Apple, Google, GitHub are also not forcing to attach underline to their website links.
However, I think that they are making it easy to distinguish by selecting distinctive color scheme or by adding a character (eg: >) Indicating a link.

Rather, I think that it is easier to see if we change the link character color to the same blue color as "Show thread" on the High contrast theme.

SteveOSigma:
This discussion is odd.

@Gargron I am a visual designer of many years. _Form **follows** function_ is a basic tenet of design. In this case form = good practice digital design for maximum usability = underline the links.

I'm a designer, not a dev, or an impaired user. Accessibility is most definitely a topic.

Why is everyone arguing with specialists who are just trying to improve the usability, and broaden the user base, of Mastodon?

hyperpress:
So, the fact that this is the discussion is covering everything from contrast, to focus, to multiple themes and browser variations, high contrast themes, and more, overwhelmingly supports the initial solution by @aardrian. 

The underline works, is less complicated, with better usability. Honestly, the style preference (aesthetic) is the only counter to the underline solution. It's a far less complex and more universal accessibility win to use underlines.

aardrian:
@mayaeh 

> However, I think the main theme does not need to make the link stand out beyond the main text.

I disagree. That is exactly where it needs to live. For all the reasons I have cited (more audience reach, supports users across a variety of browsing contexts, easier/consistent management, …)

> By the way, underline is also added to hashtags and reply names.

Yes. That is intentional. I said as much above. They are all actionable controls, they are all links.

> I think the link is not so easy to distinguish because it is full of underline.

Between web page links versus handles versus hashtags or between something else?

> I think that Apple, Google, GitHub are also not forcing to attach underline to their website links.

If you scroll up you will see that I repeat a couple times that arguing for known accessibility failures on other platforms. These are also terrible examples because they ubiquitous interfaces used regularly by familiar audiences. Far more than Mastodon.

> However, I think that they are making it easy to distinguish by selecting distinctive color scheme or by adding a character (eg: >) Indicating a link.

Please, for the love of FSM, read the section _Styling Options_ in my post that I keep linking (I've linked right to that section this time): [On Link Underlines](http://adrianroselli.com/2016/06/on-link-underlines.html#styling)

> Rather, I think that it is easier to see if we change the link character color to the same blue color as "Show thread" on the High contrast theme.

If you choose to do a color change, that is your call (and one against which I have recommended more than once here). But you cannot limit it to the high contrast theme. Again, for all the reasons I have (repeatedly) outlined above.

Can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

clarcharr:
Eugen, you're being bombarded with tonnes of actual, cited sources explaining why this extremely small change is important. You're responding with vague, uncited claims of "average users" without any studies to back it up.

If there is data to support this, please share it. Otherwise, stop being an ass and merge this literal one-word change.

bobstechsite:
As mentioned above, WCAG 2.0 compliance would require that colour is *not* used as the sole method of conveying content and visual elements on a webpage: https://www.w3.org/TR/WCAG20/#visual-audio-contrast-without-color

Colour-blind users and organisations complying with legal requirements for internally-hosted instances would benefit from underlining links in toots by default.

enterprisey:
I don't think clean design is incompatible some sort of line under the links. Regular CSS underlines look really messy to me, since they're the same color as the links themselves. On the other hand, Wired's border-bottom approach looks good, at least in my testing. (I'd advocate for text-decoration-color, but that [doesn't work on some browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color#Browser_compatibility).) So we could just do the same link color, but with only 70% alpha, or something. That way we get underlines but they don't have the same weight in the interface.

clarcharr:
@enterprisey I would strongly recommend text-decoration-color over border-bottom. It's necessary for text that covers multiple lines and underlines have actual typographical requirements to not look terrible, which should be handled by the font and not CSS.

enterprisey:
@clarcharr Sure, it just depends on whether we want to support IE (and Edge? I haven't checked).

If there's documentation on which browsers are officially supported by the Mastodon front-end and IE isn't among them, we're good.

clarcharr:
@enterprisey I wouldn't really consider a browser not supporting coloured underlines a reason to replace it with a hack that makes underlines work worse for everyone. Just let some people deal with slightly differently coloured underlines while people with newer browsers get a different colour.

jmfcodes:
User story: new user signs up for mastodon, skims the pop-ups at first login, looks at the public TLs, can’t really differentiate links or has trouble mousing over due to color blindness and decreased dexterity. New user gets frustrated. New user doesn’t know high contrast theme exists (even if it’s also an accessibility fail), because it requires navigating inaccessible UI, logs out, and gives up on mastodon entirely.

Masto Dev Story: once again, maintaining the (inaccessible) status quo.

enterprisey:
Ok, cool. @aardrian would you consider adding another rule that sets text-decoration-color to `rgba($secondary-text-color, 0.7)`?

aardrian:
@enterprisey 

> Ok, cool. @aardrian would you consider adding another rule that sets `text-decoration-color` to `rgba($secondary-text-color, 0.7)`?

I would. Do we know the secondary color also has sufficient contrast against the background?

I would rather do it as a separate PR. After all, not all theme colors may contrasty enough to guarantee the secondary color at 70% has sufficient contrast. So testing would be needed. And probably documentation for new theme developers to ensure sufficient contrast for new colors.

In short, similar challenges to just changing the color but now affecting only those on newer browsers.

benlk:
Though I advocate underlines, I can think of one (1) situation where underlines are inappropriate. That situation is [when the text is in Chinese](https://en.wikipedia.org/wiki/Underline#Underlines_in_non-Latin_scripts), because underlines have a semantic meaning in Chinese that is not held in other languages.

As far as I'm aware, Mastodon/Tootsuite don't track language on a per-toot basis, so [per-language link styles](https://www.w3.org/International/questions/qa-css-lang) would need to be applied based on the UI language (`<html lang="cn">`) , not based on an individual toot's language (`<article aria-posinset="1" aria-setsize="20" data-id="" tabindex="0" lang="cn">`).

Through [this site](https://www.chenhuijing.com/blog/chinese-web-typography/#%F0%9F%8E%AE) I found[ a typographic resource for use with Chinese characters](https://hanzi.pro/manual/); the link styles used on that site are to use a contrasting color, with color change and colored dotted underline on `:hover` (but we've already discussed how `:hover` is insufficient on devices without fine pointers).

I don't speak Chinese, so take this with many grains of salt. If there's anyone who knows more about link styling in Chinese, I yield the floor to them.

It may be better to leave `text-decoration-style` unset, and rely on user-agent styles for link text decoration.

aardrian:
@benlk 

> It may be better to leave text-decoration-style unset, and rely on user-agent styles for link text decoration.

You raise a really good point, and one I forgot. For background, I also had a localization company for a decade or so and we did lots of work localizing content for SE Asia, China, Japan, etc.

If we use the code as I submitted it we are in good shape. User agents will adapt it to the locale settings. If we start trying other styles (on default underlines or in place of) then we will need to do testing.

So probably safest to just approve the PR as it stands and then figure out the internationalization bits after gathering some data.

Amolith:
As a long-time Mastodon user, I've always found the lack of noticeable contrast between links and the surrounding text annoying and so has everyone I've personally talked to. I don't really understand why _increasing_ the contrast is a debate. It's really hard for people to distinguish when skimming _(we all do it, don't lie)_

_How the contrast should be implemented_, however, is something that does need discussion. I do think that underlines disrupt the flow and make the UI look worse. Changing the text colour is a much better solution in my opinion. What the colour should be changed to, though, I don't know. 

I have somewhere to be in a few minutes but I wanted to throw my opinion out there. When I get home, if a decision hasn't yet been made, I'm going to see about coming up with a better colour.

aardrian:
@Amolith 

> _How the contrast should be implemented_, however, is something that does need discussion.

Check out colors for one theme [as identified up above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-456463569).

Also bear in mind that it would need to address all themes, and provide guidance for theme developers, [as outlined above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457642821) (in a few other spots too).

If you read through this thread you will see the general consensus (barring the single argument against) is standard underlines are the most effective, least labor intensive solution right now. But if you want to take a stab at it, go for it.

In the meantime, @Gargron , can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

Amolith:
@aardrian 

I certainly agree that underlines are the most effective and easiest solution and maybe even that they should go on and be implemented :man_shrugging: I just think that it would be a good idea to _eventually_ use a more contrasting colour rather than underlines because, again, _in my opinion_, it does clutter up the UI and make it less appealing.

@Gargron, I would also like to say that I love the UI of Mastodon and think it's _very_ well designed other than this one issue.

dolljoints:
I would like to add that the "but it's ugly" argument also made real world places inaccessible for a lot of people before high contrast railings, braille on elevator buttons and signs, tactile paving and so forth became required.
 I am deeply deeply troubled by the exact same arguments being done for online spaces. Twitter is also really bad to use with a screen-reader, you can't say "but Twitter doesn't do it" as an argument against doing it. 
9/10 of my disabled friends (disabled in many different ways) tried Mastodon and gave up on it after a few days, for many of the reasons people keep saying in this thread. I don't buy the idea that making Mastodon more accessible will somehow cause the userbase to plummet, nor do I buy into the ideology that expansion at any cost is morally better than ensuring true multi-dimensional accessibility.

deraw (approved):

Amolith:
Just had a thought to throw out there. Hashtags have an octothorpe (#) preceding them. Handles have an arobase (@) preceding them. We've all been talking about underlines vs colours. Why should links not have a symbol preceding them as well? Maybe add  to the front? I'm actually working on implementing the same thing on my website because I don't like underlines either, bold isn't enough to differentiate them, and I like the text being a uniform colour as well.

It has a visual element that noticeably sets it apart, looks pleasant (especially on a new line), and I don't think that it detracts from the flow much, if at all. It stands out if you're _looking_ for links but isn't distracting when you're trying to read a toot either.

 **mastodon.social**

I just grabbed the glyph from [Font Awesome](https://fontawesome.com/icons/link?style=solid).

clarcharr:
@Amolith main issue with this is that that's the link is already an emoji ( :link: ) and isn't actually required in the markup, and could be confusing. Otherwise, it's a decent suggestion. Underlines still seem to be the best solution here, at least until a better one is found.

aardrian:
@Amolith 

> Why should links not have a symbol preceding them as well? Maybe add  to the front?

I talked about other styling options and even linked to more information [up above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457676462): [On Link Underlines](http://adrianroselli.com/2016/06/on-link-underlines.html#styling)

To quote:

> One of the studies tried using icons with some success, but without an icon at the start and end of the text, there was no way to indicate to a user when the link ended.
> 
> If you rely on icons to indicate off-site links, downloads, secure pages, email links, etc., then this approach is probably not a good fit. Another catch is that there is likely an additional HTTP request to get the asset (as using icon fonts will confuse screen readers), and the need to account for Windows High Contrast Mode, some sort of description, and maintaining quality and alignment as the text is scaled.

The research is linked in the post.

Now we have to consider a few other things.

* How will it be announced by screen readers?
* Are you willing to insert even more HTML to support screen reader users properly (hint, see below)?
* Are you prepared to sub-set the font?
* Are you ok with FOUT?
* What about users who block third-party content (heck, even my ad blocker does that in some cases)?
* Will it have enough contrast?
* Will it add too much visual noise to the toot, considering hashtags and octothorpes?
* How will you denote a handle or octothorpe is clickable (as that is still a WCAG requirement since it is an actionable control)? Perhaps an underline?

FontAwesome has a [page about accessibility](https://fontawesome.com/v4.7.0/accessibility/). The advice for strictly decorative icon use it to follow this HTML pattern:

`<i class="fa fa-fighter-jet" aria-hidden="true"></i>`

But now you have to test that. That is a bigger code change, not just CSS. There is a whole process you need to follow, additional payload for end users, and still a chance the icon never appears (just as it does not in your comment here).

Still, that snippet above is your safest bet. But if you want to start using them elsewhere (FontAwesome is a slippery slope) the HTML gets more complicated:

```
<a href="path/to/shopping/cart" aria-label="View 3 items in your shopping cart">
  <i class="fa fa-shopping-cart" aria-hidden="true"></i>
</a>
```

Now you have a localization challenge. Google Translate does not recognize `aria-label`. So then you will need to have localized text strings coming out of your database for all the languages you support.

There is a reason developers are moving away from fonts to SVGs.

Now if you want to start _that_ discussion, I have a pile of reading for you before you even come back here with it:

* [Using ARIA to enhance SVG accessibility](https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/) (2013)
* [Tips for Creating Accessible SVG](https://css-tricks.com/tips-creating-accessible-svg/) (2014)
* [Accessible SVGs](https://css-tricks.com/accessible-svgs/) (2016)
* [Making SVG accessible: London Web Standards](https://developer.paciellogroup.com/blog/2017/05/slides-london-web-standards-making-svg-accessible/) 2017
* [Accessible SVG Icons with Inline Sprites](https://www.24a11y.com/2018/accessible-svg-icons-with-inline-sprites/) (2018)
* and then I stopped because I just want to go to bed.

Please don't come back with SVG suggestions without doing the legwork.

Anyway, this crap is all off the top of my head. At practically midnight. Within 15 minutes of the font icons suggestion. This is _what I do_. I don't mean to be cocky, but this is just tiring. Take the experiments to replace underlines and do them on your own time. Read up on the options, do your research, make a wonderful design, and issue your own PR.

Please, @Gargron , can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

aardrian:
I hit the wrong button. It is late. Re-opened.

aardrian:
I think the problem is that I never told you that underlines are objectively quite attractive.

I like the crunch in my molars when I click on them. The smell of ozone, the taste of pennies. They bring me the joy of Christmas morning, the calm of falling asleep in a library.

The confidence of the line, shape of the graceful rectangle, together have the beauty of a Stradivarius and the soul of wit (unlike brevity).

They are as close to perfection as the face of the owner of His Noodly Appendage.

So I think that settles the aesthetics issue you raised.

With all the evidence I cited above and my sincere belief underlines are pure poetry of form, let’s do this, @Gargron ! Approve this PR and let’s make something beautiful together!

hughrun:
"Links should look like links" is a basic tenet of web design, whether "accessible" or not. This is about the design of *the standard out-of-the-box* Mastodon. If @Gargron or anyone else prefers to retain less accessible URLs on their own instance for personal aesthetic reasons, then they have the custom CSS section available to remove the underline.

trwnh:
One point I'd like to note -- not necessarily for or against, but also not based purely in aesthetic -- is that from my experience, I expect underlines to be applied to links _on hover_ rather than by default. I'm not entirely sure why I have this assumption, and I can't really break down the percentage of websites that do this, but this is my gut assumption and what I instinctively expect after 20-odd years of browsing the web. I can't say for sure, but I'd warrant that most people can agree with the following due to prevailing design patterns:
- Links are blue.
- Visited links are purple.
- Underlines should be applied sometimes.
These three things are roughly constant across most websites and therefore reasonable expectations in the absence of significant theme overrides.

With respect to creating contrast in an absolute sense, though, one should stick to only one paradigm -- color or shape. I personally think links should take the accent color and have an underline applied on hover or focus, and that's how I personally do it in my Masto Flat CSS theme. But this could easily be reversed so that links are underlined and take a color on hover instead. Regardless of which ends up being chosen, the current contrast is way too low, even for the non-high-contrast theme.

yatil:
@trwnh You assume that everyone can perceive color. The default in browsers is that links are always underlined. That there is an anti-pattern that was established when people didn’t care about accessibility/usability much, informed by the designers who only knew print, is not an argument to continue this user-hostile design trend. 

- - -

People who cannot perceive color, who don’t use the mouse pointer, who use the mouse but have a hard time pointing it precisely, elderly people, people who use the site in the sun and everyone else profits from underlines. 

I’m flabbergasted that open source communication projects don’t want to reach the widest audience possible. Just add underlines. It really isn’t a big deal.

trwnh:
@yatil I don't disagree, nor do I assume everyone can perceive color.

aburone (approved):

bobstechsite:
If it's of interest to anyone, I've raised a feature request about making it easier for individual users and instance admins to apply their own custom CSS themes and share them with others.

I think the discussion about default themes is an important one to have (as I stated above, I'm in support of adding underlines on the grounds WCAG 2.0 suggests colour alone isn't sufficient), but giving people more of an ability to override the default behaviour in future releases may be one way of resolving this discussion.

https://github.com/tootsuite/mastodon/issues/9919

scottaohara:
@trwnh some edits to your bullets about links which would make the points indisputable:

* Links are blue (by default).
* Visited links are purple (by default).
* Underlines (are applied by default, and CSS can be written to remove them. Without extra care/design considerations, removing underlines can have a negative impact on allowing people to be able to quickly identify links, based on visual and/or situational impairments).

It honestly matters little of what other websites do with their links.  There are *tons* of inaccessible and/or poor UX ant-patterns that other websites employ.   Poor design patterns does not excuse poor design patterns. 

Adding underlines to links, by default, would help provide a more inclusive experience to Mastodon users.  @bobstechsite's feature request is a good idea, but is a much bigger ask than a single line edit to the default CSS theme.  But please, do not interpret that feature request as thinking its OK to put the onus on the user to make their theme more accessible  (not saying that's what you're suggesting bob. but it's often path people take to disavow themselves of responsibility).  


All this circular discussion boils down to this though:

You can either accept this PR to make links more inclusive to Mastodon users (and then take up additional work for an alternate accessible solution, if you so choose).

Or you can continue to value opinionated aesthetics over cited usability concerns.

dustinwilson:
I am quite surprised to see a thread as long as this one over accessibility of hyperlinks especially considering Mastodon has for quite some time now included accessibility features but refuses to right now do the most basic of changes that would benefit the accessibility of a large amount of people.

I have 10 years of experience in the Web browser realm, working as a user experience designer first on contract for Opera Software and then later full time. The W3C created guidelines for all of this stuff starting years ago, a good bit of it worked out by many of my colleagues at Opera. Thanks to the miracle of eyeglasses and contacts I have perfect 20/20 vision, and I sometimes have difficulty discerning the difference between plain text and hyperlinks on Mastodon. This is because the contrast ratio between the white text of a toot and its hyperlinks is an abysmal 1.32:1, and factors such as varied lighting in my own environment and glare on my screens can contribute to impaired vision. The W3C recommends at least a 3:1 ratio between hyperlinks without an underline and the text around them, and that is for *normal* vision. At the very least please change the color to something even people with normal vision can see. This applies to colors of icons as well.

High contrast themes are a great thing, but when users must navigate an inaccessible-for-them user interface looking for a UI widget that they might not even know exists to make the page display in a higher contrast it in of itself is not accessible. A Web application intended to be used by the mass public should at least conform to WCAG AA and its high contrast theme conform to WCAG AAA or more. This allows for people who aren't blind enough to need screen readers to navigate your Web application and even find where to turn on the high contrast theme if needed.

There have been extensive usability testing done over nearly three decades concerning hyperlinks, and it has been determined time and time again that underlined links are much faster for users to identify and use. Users are so accustomed to scanning for underlines that any use of underlines except for hyperlinks on the Web is strongly discouraged. Usability tests have also shown that text that is colored significantly differently from text around it are also identified by most users as hyperlinks, but the time to identification is significantly higher than with underlined hyperlinks. Citing other websites' non-usage of underlined hyperlinks isn't a valid argument considering the other side has scientific proof through decades of usability testing to back itself up. It's not realistic in a web application to make all links underlined because of hierarchy of information and links-that-really-are-buttons. At the very least consider making hyperlinks in toots underlined in the high contrast theme.

tl;dr Please increase the contrast ratio on hyperlinks to a 3:1 ratio against the text that surrounds it and add underlines at least to hyperlinks within toots to the high contrast theme.

aardrian:
@Gargron , have you ghosted? You have not offered any new information on this PR in 4 days but your GitHub contributions chart shows you have been active each day since (including today).

I made a screen shot showing how the underlines look in the toot.cafe theme. Note that now links are visible in the notifications panel (where the contrast is reduced even more). Note how they ease scanning. Note [how beautiful the underlines are](http://adrianroselli.com/2019/01/underlines-are-beautiful.html).

The PR in the SCSS file compiles to this simple declaration:

```
.status__content a, .reply-indicator__content a { text-decoration: underline; }
```

It is so simple, it even fits in a bookmarklet:

```
javascript:(function(){var a=document.createElement('style'),b;document.head.appendChild(a);b=a.sheet;b.insertRule('.status__content a, .reply-indicator__content a{text-decoration:underline}',0);})()
```

So, again, please approve this PR. Let's provide a better experience for all of Mastodon's users.

Amolith:
I'm going to change the opinions I've mentioned above and say I'd prefer underlining links rather than increasing contrast. I was thinking that this PR would look like a screenshots shown in a [comment above](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457058606), with links under every, well, link. However, as @aardrian shows [here](https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457932479), the change only applies to . . . "normal" links and I think that *does* look quite nice. While it admittedly disrupts the "flow" a ***little***, I think it makes the UI look bit cleaner and I honestly prefer it to coloured URLs.

nickcolley:
I was wondering if the underlines looked aesthetically bad but given @aardrian 's screenshot it looks really tidy and clear.

Mastodon has the opportunity to lead and be the most inclusive platform available. 

This is something we can be proud of that commercial platforms have little incentive to do.

dariusk:
@Amolith Both of the screenshots you link seem to be showing the same implementation. The only difference is that @aardrian's screenshot shows posts that don't have hashtags and @-mentions in them.

1. https://user-images.githubusercontent.com/766076/51653631-933a1e00-1fd7-11e9-8965-bd5cb5a4745f.png
2. https://user-images.githubusercontent.com/1376607/51803727-4730e880-2226-11e9-931c-a59950d55e7a.png

You can see in image 1 that there are underlines under hashtags, @-mentions, and regular links. There are not underlines under the headers show user info in each toot.

In image 2, there are underlines under hashtags (you can see it in `#a11y` at the top of the notifications column), there are no @-mentions present, and regular links are underlined.

In short, they seem to be the same (there are no @-mentions present in the 2nd image which makes it seem cleaner). @aardrian's screenshot isn't terribly representative of my usual timelines, which have a lot of @-mentions and not a lot of hashtags.

nightpool:
The underline looks pretty good in aardian's screenshot, but it's on the
custom toot.cat theme, which has different link colors. It might look
better with the new high-contrast links, but it definitely doesn't look
good on stock mastodon link colors, even when you exclude the mentions.

Amolith:
@dariusk you're right :man_facepalming: :sweat_smile:

dolljoints:
that screenshot looks really nice, it's wild how much less cognitive effort
it is to parse, especially in the notifications.
i thought i didn't have much personal desire to use it, but thanks to you
posting the bookmarklet, i'm gonna give it a go with the cybre theme.
that's a really nice line thickness, and i think it does look better
overall honestly!
(at the very least this should be some option in the visual appearance
section, you could even choose what you want underlined.)
also: a lot of the time i forget to type an instance sometimes and it just
says @user but isn't clickable. the underline highlighting clickable things
can be nice to clarify if you've got it wrong, or if someone is @ing a
nonexistent account in shitposting
so, yeah, i think this is really neat. i hope the option gets added soon
but am really grateful for the bookmarklet letting me make the option
exist. thanks javascript, thavascript

aardrian:
Screen shots with default Masto theme with toots pulled from the federated and local timeline (I could not find ones with lots of hashtags or handles).

But again, the aesthetics are arbitrary. The base should be underlined, with specific instances having the ability to remove underlines or otherwise hobble their UI.

benlk:
Another screenshot showing the default theme, with underlines via the change that this PR would make.

The difference between this screenshot and @aardrian's screenshot is that my browser, the current Chrome, has [`text-decoration-skip-ink: auto;`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink) set in the user-agent stylesheet. This means that where various glyphs' descenders cross the underline, there's a gap in the underline to accommodate the descender.

You'll also notice that underlines in this screenshot are thicker than in the other screenshot; that's probably because I'm either browsing at 90% zoom or because of browser-dependent rendering of underlines.

 screen shot 2019-01-28 at 9 50 10 am

Here's a screenshot allegedly using the "High Contrast" theme, also with underlines not disabled:

screen shot 2019-01-28 at 9 58 40 am

clarcharr:
Very much appreciate this combined with the contrast increase in the other PR. Increasing colour contrast does not negate the need for underlines; both are very appreciated.

deraw (approved):

ashleyhull-versent:
This is the best bike shed I have ever seen.. You drive your bike up to the bike shed, take a ticket and a robot arm reaches up from the ground and positions the bike onto a conveyor belt that takes the bike unto the underground storage compartment to protect the bikes from theft and the elements?!

Wow.. such a bike shed.. please approve this PR

nightpool:
(I've deleted a comment from this thread for being overly hostile and dismissive. Everyone on both sides is reminded to be civil, respectful, and assume good faith when discussing this issue)

aardrian:
Does anyone here have the authority to add the _Accessibility_ label to this PR?

aardrian:
Thanks, @sorin-davidoi !

deutrino:
One problem among several with the "people can totally tell links are links because they're always a URL" claim: Mastodon isn't going to be the biggest fish in the pond for long, and other ActivityPub software may well start emitting links that aren't in URL form.

dolljoints:
rather than "may well start", it is already here. some of my friends who
run a single user pleroma instance for themselves sometimes will end up
posting a thing where the link is just contained in the word "this" or it
is a filename like "photo.jpg" and on hover, you realise you can click it.
(or, with that javascript code to add underlines, then i can tell without
having to). i don't know what exactly causes that but that's how it ends up
looking for me in masto web.

aardrian:
With @dolljoints' comment about only the word _this_ being linked, I believe we definitely cannot rely on a URL as a link signifier.

Can we approve the link underline here? I haven't heard any arguments against that go beyond aesthetics.

u1-liquid:
Is there any reason why this PR has been left for so long?
Well explained, easy to identify link, and it  definitely not ugly, isn't it?

aardrian:
For comparison, this same change (adding underlines to links) went from PR to approved and deployed in under 2 hours on the ECMA TC39 site (the body that standardizes the JavaScript language): https://github.com/tc39/tc39.github.io/pull/141

jeremyredhead:
Just a rambling rando but, here's my take: 
Whether or not underlines look good on links not only depends on the person, but the site itself.
When I tried enabling underline on all links here on github, I found myself agreeing with @Gargon &mdash; visual clutter. 
But when I tried disabling underlines on @aardrian's site for example, I found myself preferring the classier underlines.

Ultimately however, it should be up to the __user__, neither Gargon nor the instance owner, as to how links are styled, and probably different options at that for outgoing.links vs @people or #hashtags.
(checkboxes for bold and underline would probably be sufficient). Of course, by default _at least_ normal links should have underlines; maybe bold for @people like here on github?

I'm not actually a mastodon user (yet), so take this all with a grain of salt (does masto even support settings like i described currently?). If you think I'm a crank, don't @ me just delete comment and move on.

wxcafe:
hey how about we merge this? links in the default UI *are* hard to read and there already are non-url links everywhere, this is a good accessibility measure and letting it rot will not make the problem disappear. Make a toggle if you want, but this PR is *good* for the community

dolljoints:
thank you wxcafé!

wxcafe:
well I mean, I didn't really do anything apart from commenting, gargron doesn't seem to want to do anything except ignore this, so once again mastodon is blocked on a basic accessibility issue

dolljoints:
yeah, i just like seeing people be vocal about this. it feels validating.
obviously you just said you think it should happen too but. it's nice to
see ppl talking about it.

benlk:
This PR adds an underline to `.status__content a` and `.reply-indicator__content a` across the board. 

The PR https://github.com/tootsuite/mastodon/pull/9949 does the same, but more extensively, and only in the "high contrast" theme. It was opened after this PR, and it has been merged.

rugk:
Ah, so this is fixed? So this here can be closed?

trwnh:
@rugk it's not fixed for the standard themes, which are de facto "low contrast". i think there was another pr to change the color of links and achieve color-based contrast, but that one was closed unceremoniously... #9928

aardrian:
@rugk & @benlk 

> Ah, so this is fixed? So this here can be closed?

No. I addressed why setting underlines on the high contrast themes is insufficient in my comment above: https://github.com/tootsuite/mastodon/pull/9898#issuecomment-456811331

The high contrast change in #9949 would have been unnecessary had this PR been merged. As it is, I think the high contrast theme still fails to pass contrast checks.

Regardless, as long as this PR sits unmerged, Mastodon default themes are still in violation of WCAG. And unusable for many.

Please, @Gargron , can we approve the link underline here? I haven't heard any arguments against that are more than a personal preference.

rugk:
Well… yeah… the default theme has a bad contrast anyway. That's also already bad for accessibility…  :thinking: Don't know whether underlines in links are the biggest problem there… :wink:

wxcafe:
it not being the biggest problem doesn't mean it's a problem and maybe we can start with applying the solutions we have to fix the problems we can before moving on to other problems, rather than sitting around waiting for everything to be magically solved at once.

aardrian:
@Gargron, can we approve the PR, please?

All the arguments against it have been raised and addressed. Copious amounts of evidence have been provided. Benefits have been demonstrated. Related issues have only addressed narrow cases. Let's help users across contexts _and_ make it satisfy WCAG.

weboverhauls:
Please merge.

aarongustafson:
::bump:: @aburone @deraw

evanminto:
I am 100% in support on @aardrian’s changes as-is, but I’d like to propose a slight amendment to them that might make them more palatable to those concerned about aesthetics.

Adding `text-decoration-color: hsla(0, 0%, 100%, 0.5);` (or a version of it using a variable to switch out the color based on the link color) fades out the underline a little bit. Using this property it’s possible to reduce the underline contrast while still keeping it above the 4.5:1 ratio. Combined with `text-decoration-skip-ink: auto;`, it looks like this:

Screen Shot 2019-10-15 at 8 41 40 AM

`text-decoration-color` is supported in Chrome, Safari, Firefox, Samsung Internet, and Opera. `text-decoration-skip-ink` is supported in Chrome, Opera, and the upcoming version of Firefox. The fallback behavior in IE/non-Chromium Edge/etc. is the less attractive legacy underline without the descender skipping and custom color, so those browsers will also maintain accessible styles.

aardrian:
@evanminto, the catch there is different themes may not guarantee sufficient contrast. See my comment (https://github.com/tootsuite/mastodon/pull/9898#issuecomment-457730947) on a similar suggestion above. Though that might be a good addition for the individual themes.

Gargron:
The color of links has been changed to be distinguishable via contrast; those who want to see underlined links can switch to the high-contrast theme (unfortunately some users had to switch away from the high-contrast theme after the underlines were added because it made it less legible for them). This PR will not be merged. Thank you.

PR #9907, opened 2019-01-22
Update en.yml with minor language and grammar fixes

jazmichaelking:
- Consistent use of "email" vs. "e-mail" (arbitrarily chose "email" based on existing usage count)
- Consistent use of "added" and "deleted" (vs. the occasional created/destroyed)
- Consistent use of "toot" vs. "status" (where status means toot not status)
- Consistent use of "can" vs. "may" - (arbitrarily chose "can", little more friendly)
- Successfully removed "successfully" from success messages!
- "Profile directory" is now "Profile Directory"
- "Back to Mastodon" is now "Back to Timelines"
- Minor grammar updates, removed some ambiguous wording

Open to comments, feedback. Some of these changes were driven by the past four months of translating Mastodon to Welsh; my hope is some of these edits will help future translators better understand context or intended meaning. Other edits are simply for a sleeker read, redundant redundancies,  and consistent usage to better communicate to the end user. As with all things grammar and language, nothing is absolute, but I made some arbitrary choices that could go either way and still be "correct".

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2,12 +2,12 @@
 en:
   about:
     about_hashtag_html: These are public toots tagged with <strong>#%{hashtag}</strong>. You can interact with them if you have an account anywhere in the fediverse.
-    about_mastodon_html: Mastodon is a social network based on open web protocols and free, open-source software. It is decentralized like e-mail.
+    about_mastodon_html: Mastodon is a social network based on open web protocols and free, open-source software. It is decentralized like email.
     about_this: About
     administered_by: 'Administered by:'
     api: API
     apps: Mobile apps
-    closed_registrations: Registrations are currently closed on this instance. However! You can find a different instance to make an account on and get access to the very same network from there.
+    closed_registrations: Registrations are currently closed on this instance. However! You can register on a different instance to and get access to the very same network from there.
     contact: Contact
     contact_missing: Not set
     contact_unavailable: N/A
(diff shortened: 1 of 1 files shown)


Gargron:
By adding an extra variable to the english translation of some keys, you have invalidated their translations in other languages. It is easiest to simply undo those particular changes and avoid new variables.

config/locales/en.yml
```diff
@@ -25,14 +25,14 @@ en:
       within_reach_body: Multiple apps for iOS, Android, and other platforms thanks to a developer-friendly API ecosystem allow you to keep up with your friends anywhere.
       within_reach_title: Always within reach
     generic_description: "%{domain} is one server in the network"
-    hosted_on: Mastodon hosted on %{domain}
+    hosted_on: This Mastodon instance hosted by %{domain}
```
joycem137: I thought Mastodon is moving away from "instance" and moving specifically to "server" as the general language?
```suggestion
    hosted_on: This Mastodon server hosted by %{domain}
```

config/locales/en.yml
```diff
@@ -606,22 +606,22 @@ en:
       title: Add new filter
   followers:
     domain: Domain
-    explanation_html: If you want to ensure the privacy of your statuses, you must be aware of who is following you. <strong>Your private statuses are delivered to all instances where you have followers</strong>. You may wish to review them, and remove followers if you do not trust your privacy to be respected by the staff or software of those instances.
+    explanation_html: If you want to ensure the privacy of your toots, you must be aware of who is following you. <strong>Your private statuses are delivered to all instances where you have followers</strong>. You may wish to review them, and remove followers if you do not trust your privacy to be respected by the staff or software of those instances.
```
joycem137: If you're changing "statuses" to "toots" you missed one here.

```suggestion
    explanation_html: If you want to ensure the privacy of your toots, you must be aware of who is following you. <strong>Your private toots are delivered to all instances where you have followers</strong>. You may wish to review them, and remove followers if you do not trust your privacy to be respected by the staff or software of those instances.
```

config/locales/en.yml
mynameismonkey: I completely forgot about this, I need to redo with Eugen's direction and I'll factor in your comments as well, thanks! (I'm unaware of an "official" move from instance to server, but will happily factor that in)

mynameismonkey:
Updated new PR https://github.com/mynameismonkey/mastodon/pull/2

PR #9928, opened 2019-01-27
Change color of links in toots to have higher contrast

Gargron:
Alternative to #9898. Improve ability to distinguish links from other text.

|With color|Grayscale|
|------------|--------------|
|links-color|links-grayscale|

Unlike underlines, it keeps visual noise down in an otherwise busy interface. Unlike text links in articles, links in toots are associated with specific syntax (@ for mentions, # for hashtags, example.com for URLs), and their primary function is to be read in the text, rather than clicked on. Contrast ensures the distinction works even for colorblind people (tested with [this tool](https://www.color-blindness.com/coblis-color-blindness-simulator/))

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -669,15 +669,11 @@
   }
 
   a {
-    color: $secondary-text-color;
+    color: lighten($ui-highlight-color, 12%);
     text-decoration: none;
 
     &:hover {
       text-decoration: underline;
-
-      .fa {
-        color: lighten($dark-text-color, 7%);
-      }
     }
 
     &.mention {
@@ -689,13 +685,10 @@
         }
       }
     }
-
-    .fa {
-      color: $dark-text-color;
-    }
   }
 
   .status__content__spoiler-link {
+    color: $secondary-text-color;
     background: $action-button-color;
 
     &:hover {
@@ -731,7 +724,7 @@
   display: block;
   font-size: 15px;
   line-height: 20px;
-  color: lighten($ui-highlight-color, 8%);
+  color: lighten($ui-highlight-color, 12%);
   border: 0;
   background: transparent;
   padding: 0;
@@ -1065,6 +1058,10 @@
 
   a {
     color: $lighter-text-color;
+
+    &:not(.mention) {
+      color: $lighter-text-color;
+    }
   }
 }
 


ThibG:
I think the link color change is very welcome. I am very unsure about not changing hashtags and mentions, though.

Gargron:
All the colors involved are in the same palette, present through the UI. Mentions and hashtags are more common than URLs, I would say.

Gargron:
I've stumbled across the problem that not all posts use the same class names to mark up mentions/hashtags, so keeping the colors different is more inconsistent than I originally thought. So perhaps changing all of them would be more appropriate.

hyperpress:
This is pretty dissapointing. If you want to be Tw

nolanlawson:
I appreciate that this PR automatically fixes the color contrast even for custom themes. Here it is with Toot Cafe's default purple, the "Paris mon Amour" theme on mastodon.technology, and the "Photon Dark" theme also on toot.cafe.

Screenshot of three different themes with good color contrast

dariusk:
@aardrian, you say "contrast alone is insufficient to achieve WCAG compliance" -- I thought that [WCAG 2.0 states that color contrast is in fact sufficient](https://www.w3.org/TR/WCAG20-TECHS/G183.html) if it's a 3:1 contrast ratio, as long as a hover also causes underlining, which it does already.

And I'm trying to figure out what the WCAG position is on this, btw. You seem _far_ more the accessibility expert than I am and I'd appreciate any corrections you could offer.

Heydon:
@dariusk You want 1.4.1 https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html

aardrian:
@dariusk , @Heydon has already linked it, but for more context and discussion of G183 and how it relates to 1.4.1, I have a section in my post (which I wrote for use with clients in this same situation): [On Link Underlines](http://adrianroselli.com/2016/06/on-link-underlines.html#accessibility) (this points to the Accessibility section)

dariusk:
@aardrian, your article quotes this:

> If there are not a large number of links in the block of text, underlines are recommended for links

You're emphasizing the part after the comma, those who disagree are emphasizing the part before the comma. I do take your point that these guidelines were written before touch interface was widely available.

For the record, I'm in favor of underlines as default but I'm trying to understand where everyone is coming from here. I'm merging your original PR to my instance today and I'll see what my users think (or if they even notice).

aardrian:
@dariusk 

> You're emphasizing the part after the comma, those who disagree are emphasizing the part before the comma.

Link density is a fair concern. If an entire block is all links then that is a usability issue, but the lack of affordance (underlines in particular) is still an accessibility issue. So while it might be more unusable to some in those cases, it is still inaccessible to many in all cases (whether permanent disability, or temporary/situational impairment).

clarcharr:
Colour contrast change is much appreciated, although you should include hashtags and mentions as well. And, add underlines, as mentioned in #9898.

mayaeh:
I feel that it is not necessary to change the link of the main theme.
The name and the link become more conspicuous than the main text.

But accessibility is also important.
I will accept if this change is welcomed.

Issue #9937, opened 2019-01-28
instance? server?


mayaeh:
### Pitch

I like new tutorials.
However, the expression of "instance" that the user sees has increased to "instance" and "server".

Should we unify term from instance to server?
Like "[About this instance](https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/locales/en.json#L227)", we have many places to express it as "instance".

### Motivation

I received the following question from an instance administrator.

> **I'm writing the term "instance" in /about/more.**
> **What should I do now?**

/about/more is the next page moved from the link "About this instance".
Therefore, "server" will lose its sense of unity.

Should we change everything to "server"? 
Or, should we change only a part of that?

mayaeh:
Thank you very much for your quick reply.

PR #9938, opened 2019-01-28
Rename from instance to server

mayaeh:
Change the expression from "instance" to "server".
Please refer to #9937 .
What do you think?

There may still be leaks.
Should I also change the definition name such as %{instance}?

~~I have not tested this change yet.~~
I tried a simple operation test on my test server.
I executed "yarn manage:translations en", "i18n-tasks missing en" and "i18n-tasks unused en".


Gargron (approved):

mayaeh:
Thank you for merging the master branch.

PR #9949, opened 2019-01-30
Fix link color in high-contrast theme, add underlines

Gargron:
Improve sorting of default themes in the dropdown

image


Gargron:
What I've encountered in practice is that the classes used by Pleroma/Pixelfed/Misskey and others don't quite match up, so either I make all links look as similar as possible, or it's gonna be really jarring when you see a non-Mastodon post, which I would prefer to avoid.

PR #9956, opened 2019-01-31
Upgrade Ruby to 2.6.1

zunda:
I'm not sure if the [fixed bug](https://www.ruby-lang.org/en/news/2019/01/30/ruby-2-6-1-released/) affects us but just in case.

> `Net::Protocol::BufferedIO#write` raises `NoMethodError` when sending large multi-byte string has been fixed.


ykzts (approved):

angristan:
LGTM. No issue so far on 2.6.1 for me

Gargron:
People complain whenever `.ruby-local` is updated and we just updated it in 2.7.0, so I just think this is too soon. I don't believe the mentioned bug affects our code.

angristan:
Ok 👌

angristan:
It turns out elasticsearch-rails has a bug in 2.6.1: https://github.com/elastic/elasticsearch-rails/issues/866

Another PR has been made to update the Ruby version: #10083

PR #9962, opened 2019-02-01
Add option to overwrite imported data

### LINKED ISSUE #7465: Allow user to choose whether to merge or replace imported CSV data
From what I understand by reading code and experimenting, the userlist importers in the settings page currently only merge entries into the user's existing list.

Adding an option to replace the old list with the imported one would allow a way to undo big mistakes, whether they built up over time or were a result of a bad import — i.e. export the list, fix it in a text editor, and re-import. That's not perfect but it's much easier than clicking remove buttons one by one in the web interface, or writing a thing to do it via the API.

(If added, this would also need a warning when it's done on the follow list, due to the risk of unintentionally unfollowing protected accounts)

* * * *

Gargron:
Fix #7465

image

Also add import for domain blocks

### DIFF
--- a/app/models/account_domain_block.rb
+++ b/app/models/account_domain_block.rb
@@ -12,6 +12,7 @@
 
 class AccountDomainBlock < ApplicationRecord
   include Paginable
+  include DomainNormalizable
 
   belongs_to :account
   validates :domain, presence: true, uniqueness: { scope: :account_id }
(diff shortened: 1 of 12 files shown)


Cassolotl:
I feel like the wording for "overwrite" is a bit ambiguous. How about "remove current records before adding new ones"?

ykzts (approved):

PR #9968, opened 2019-02-04
[UI] Fix whitespace being applied to div instead of p

trwnh:
Before: image

After: 
image

`white-space: pre-wrap` is an unnecessary rule, and removing it fixes the issue without introducing any noticeable side effects. The large gap was occurring mainly in posts from Pleroma, which would add the white space *and* also add `margin-bottom: 20px`, which led to effectively two line breaks for every new `p` element.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -661,6 +661,7 @@
 
   p {
     margin-bottom: 20px;
+    white-space: pre-wrap;
 
     &:last-child {
       margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


Gargron:
Did you check how the ascii art posts look?

trwnh:
Hmm, I had not. I guess adjusting white-space is the wrong strategy, then. But I can't exactly figure out why this seems to be happening -- there seems to be an issue with line height somehow, as this element is exactly 2 lines high despite only containing a 1-line `p`:

image

image

trwnh:
A better demonstration of why adding `white-space: pre-wrap` to the `div` is problematic: it means you can't use whitespace in your source code at all.

Mastodon's source: 
image

Pleroma's source: 
image

Removing white-space from inside the `div` yields this: 
image

Adding more white-space yields this: 
image

So applying the rule to the div means that this will take multiple lines:
```
<div>
    <p>test</p>
</div>
```

And so this is the only way to get a single line:
```
<div><p>test</p><div>
```

Gargron (approved):

PR #9984, opened 2019-02-08
Add option to disable real-time updates in web UI

### LINKED ISSUE #7913: Disable real-time timeline
Currently, when viewing a timeline (local or federated) and being on the top of it, it will update automatically and therefore move on. 
This makes it very hard to explore timelines, especially when being disabled in some way (for example, the movement alone makes my head dizzy).

Possible solutions could include: 
- having a "view new toots" button or something similar
- including a toggle in the settings to change the functionality of timelines to not scroll until the user chooses to scroll manually (it can also be impairing or just annoying for some to constantly having to press a button for timeline updates)
- being able to set an interval for automatic updates (every X seconds/minutes) 

This issue was already opened here: https://github.com/tootsuite/mastodon/issues/240
It was closed around 16 hours ago, even though it is **not** solved. It is not reopened yet so I'm opening a new issue because it **is** important for many. 
If the old issue is reopened this one can be seen as unnecessary and can get closed.

### LINKED ISSUE #9031: [Feature request] Allowing the user to disable timeline streaming
There should be an option in the settings menu that lets people disable timeline streaming. It would be extremely helpful for people who follow a bunch of other people and it gets extremely annoying when you interact with the wrong post just because someone just sent another one.

Gargron:
Fix #9031
Fix #7913

### DIFF
--- a/app/javascript/mastodon/actions/timelines.js
+++ b/app/javascript/mastodon/actions/timelines.js
@@ -1,6 +1,8 @@
 import { importFetchedStatus, importFetchedStatuses } from './importer';
-import api, { getLinks } from '../api';
+import api, { getLinks } from 'mastodon/api';
 import { Map as ImmutableMap, List as ImmutableList } from 'immutable';
+import compareId from 'mastodon/compare_id';
+import { usePendingItems as preferPendingItems } from 'mastodon/initial_state';
 
 export const TIMELINE_UPDATE  = 'TIMELINE_UPDATE';
 export const TIMELINE_DELETE  = 'TIMELINE_DELETE';
@@ -10,10 +12,15 @@ export const TIMELINE_EXPAND_REQUEST = 'TIMELINE_EXPAND_REQUEST';
 export const TIMELINE_EXPAND_SUCCESS = 'TIMELINE_EXPAND_SUCCESS';
 export const TIMELINE_EXPAND_FAIL    = 'TIMELINE_EXPAND_FAIL';
 
-export const TIMELINE_SCROLL_TOP = 'TIMELINE_SCROLL_TOP';
+export const TIMELINE_SCROLL_TOP   = 'TIMELINE_SCROLL_TOP';
+export const TIMELINE_LOAD_PENDING = 'TIMELINE_LOAD_PENDING';
+export const TIMELINE_DISCONNECT   = 'TIMELINE_DISCONNECT';
+export const TIMELINE_CONNECT      = 'TIMELINE_CONNECT';
 
-export const TIMELINE_CONNECT    = 'TIMELINE_CONNECT';
-export const TIMELINE_DISCONNECT = 'TIMELINE_DISCONNECT';
+export const loadPending = timeline => ({
+  type: TIMELINE_LOAD_PENDING,
+  timeline,
+});
 
 export function updateTimeline(timeline, status, accept) {
   return dispatch => {
@@ -27,6 +34,7 @@ export function updateTimeline(timeline, status, accept) {
       type: TIMELINE_UPDATE,
       timeline,
       status,
+      usePendingItems: preferPendingItems,
     });
   };
 };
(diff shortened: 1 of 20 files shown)


ClearlyClaire (changes requested):
I'm afraid this will lead to inconsistent behavior with non-websocket updates (`TIMELINE_EXPAND_SUCCESS`, `NOTIFICATIONS_EXPAND_SUCCESS`), and it doesn't set the “unread” flag nor does it clean up old toots (the 40 items limit).

Furthermore, while it should reduce CPU and RAM usage, and avoid some of the “timeline jump” issues, it will fail to eliminate those coming from preview cards appearing, those coming from Content Warnings being unfolded, and will move the others to the time of scrolling.

Gargron:
>Furthermore, while it should reduce CPU and RAM usage, and avoid some of the “timeline jump” issues, it will fail to

The goal of this PR is not eliminating some code-related issues, but simply resolving a long-standing user request.

Gargron:
I have added cleaning up of old toots and the unread flag.

I can't progress on the expand behaviour. It doesn't help that the polling fallback is indistinguishable from first load or returning to a column after leaving it. It also does these ID comparisons and gap handling which are very finnicky.

I suggest we merge the feature as it is, so that it works with streaming, and then fix the polling behaviour separately.

ThibG:
Those changes looks good to me, but I'm still uncomfortable with the polling fallback, which might break everything at the slightest connectivity issue…

Gargron:
I have made it work with polling updates and even disconnects.

app/javascript/mastodon/actions/timelines.js
```diff
@@ -71,18 +78,26 @@ export function expandTimeline(timelineId, path, params = {}, done = noOp) {
...
-    if (!params.max_id && !params.pinned && timeline.get('items', ImmutableList()).size > 0) {
-      params.since_id = timeline.getIn(['items', 0]);
+    if (!params.max_id && !params.pinned && (timeline.get('items', ImmutableList()).size + timeline.get('pendingItems', ImmutableList()).size) > 0) {
+      const a = timeline.getIn(['pendingItems', 0]);
+      const b = timeline.getIn(['items', 0]);
+
+      if (a && b && compareId(a, b) > 0) {
+        params.since_id = a;
+      } else {
+        params.since_id = b;
```
ThibG: If there are no known items yet but items are pending (can that happen?), `params.since_id` will be 0, won't it?
Why not default to '0' and compare the ids anyway?

ClearlyClaire (changes requested):
In addition to the id comparison thing, there is the fact that disabling the setting will push new toots below the possibly-pending ones. Disabling the setting should probably insert the pending toots right away.

Something seems to also be preventing loading older toots when switching back and forth between modes, but it appears to be working in both modes as long as no switching is involved.

Finally, this should be extended to list timelines as well.

Gargron:
Turns out managing the setting for lists and hashtags and the fact that you can have multiple copies of the same timeline pinned as columns is not trivial at all, and frankly, *having* to toggle the setting for each and every context might be an anti-feature so how about I make it a single global setting instead

ClearlyClaire (approved):
The name “Slow mode” really doesn't make it obvious what it does. We should find a better name or at least have a useful hint string. I'm not completely sure about having it as a list item that looks like the usual gaps and filtered items either, I wonder if it could make sense to have it in the column header instead?
Other than that, looks good to me!

chresan:
Hi, I am a new user of mastodon and wanted to give some feedback on this topic regarding it's usability. Current default is automatic pushing of messages into the timeline and moving already present messages out of the visible area. I want to point out that as a user the intention is in most cases that he wants to take a look at the messages in the local timeline to explore what is interesting. However the default setting prevents exactly this by repeatedly interrupting the reading flow.

Also it makes the webapp feel clunky that it is necessary to manually enable slow mode for basically the most common use case.

I do understand the the want to flex/see how many messages are being pushed on a server but I feel this is usually interesting only to developers and the important use case may be in the blind spot of a developer.

My suggestion is please enable slow mode by default to correspond with the most sensible use case.

PR #9998, opened 2019-02-09
Alternative handling of private self-boosts

ClearlyClaire:
- Serialize the original toot when sending self-boosts over ActivityPub
- Whenever an Announce is received with valid object from the same author, process it instead of fetching it

### DIFF
--- a/app/serializers/activitypub/activity_serializer.rb
+++ b/app/serializers/activitypub/activity_serializer.rb
@@ -3,8 +3,8 @@
 class ActivityPub::ActivitySerializer < ActiveModel::Serializer
   attributes :id, :type, :actor, :published, :to, :cc
 
-  has_one :proper, key: :object, serializer: ActivityPub::NoteSerializer, unless: :announce?
-  attribute :proper_uri, key: :object, if: :announce?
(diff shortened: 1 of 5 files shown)


app/serializers/activitypub/activity_serializer.rb
```diff
@@ -42,4 +42,8 @@ def atom_uri
   def announce?
     object.reblog?
   end
+
+  def owned_announce?
+    announce? && object.account == object.proper.account
```
Gargron: I am thinking, probably not worth wasting traffic for public items that can be fetched? On the other hand, it does spare the traffic from fetching... But then, status are commonly already known before being boosted, so it's definitely a waste in those cases...

ThibG: I don't know, including the toot is usually pretty cheap compared to performing an HTTPS request. I changed it to only embed private toots for now, but we may benefit from serializing it anyway in general. Indeed, in cases where the toot is not known to remote instances, it would be *much* cheaper to serialize it.

Gargron (approved):

PR #10001, opened 2019-02-10
Fix timeline jumps

ClearlyClaire:
This PR attempts to fix most “timeline jumps”, that is, situations in which the scrolling viewport of a timeline move to different toots without user intervention.

Timeline jumps are caused whenever a toot above the current scrolling position changes height after its first rendering, or gets deleted.

Height changes are currently due to:
1. `MediaGallery`, `Video` player and `Card` components being rendered in multiple steps, because of computations depending on their width.
2. `Card` components being displayed while they weren't at first because the status got updated at a later point, at which point the card has been fetched server-side.
3. Content warnings being folded and unfolded.

I have chosen not to fix 3. for now, as it is caused by user interaction, and thus is a little less jarring than the other issues.

This PR implements two main fixes:
- Avoid multiple-step rendering by caching the width of `MediaGallery`, `Video` and `Card` items into the `ScrollableList` component holding them, since all of those will have the same width. This mostly fixes 1. (the first rendering may cause a small jump if the default value turns out to be wrong due to CSS tweaks for instance).
- Compensate scroll position whenever we know the height changes (fixes 2., `Card` appearing, or a toot getting deleted). We want to avoid doing that when not necessary because it may cause the scroll to “stutter” a bit. I think my changes perform this as infrequently as possible.

Those are pretty big changes, so they should be thoroughly tested, which I haven't done yet.

### DIFF
--- a/app/javascript/mastodon/components/media_gallery.js
+++ b/app/javascript/mastodon/components/media_gallery.js
@@ -194,6 +194,8 @@ class MediaGallery extends React.PureComponent {
     height: PropTypes.number.isRequired,
     onOpenMedia: PropTypes.func.isRequired,
     intl: PropTypes.object.isRequired,
+    defaultWidth: PropTypes.number,
+    cacheWidth: PropTypes.func,
   };
 
   static defaultProps = {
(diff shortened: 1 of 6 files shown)


ykzts (approved):

ThibG:
Tested for a bit, I haven't seen any regression.

PR #10005, opened 2019-02-10
Filter incoming Create activities by relation to local activity

Gargron:
Reject messages:

- From accounts with no local followers
- From relays that are not enabled
- Except if:
  - The message addresses local accounts
  - Or it's a reply to an account that does have local followers

The goal of this feature is to minimize the surface area through which a server's database could be filled up through deliveries without consent.


ClearlyClaire (approved):
Seems fine to me overall. However, adding tests would be useful. In addition, it would make sense to check if relay software not relying on LD-Signatures would still work after those changes.

Finally, this is only a first step: at least `Announce` handling need similar restrictions, otherwise bypassing those `Create` restrictions is just a matter of `Announce`ing the object.

PR #10017, opened 2019-02-12
Don't focus spiler input when disabled spoiler

hinaloe:
When disabled CW, user can focus spiler-input with <kbd>Shift</kbd>+<kbd>Tab</kbd>. (since #7508?)
This PR prevent it.


Gargron (approved):

ThibG:
Wouldn't setting `disabled` be better?

Gargron:
Okay, that would be a cleaner solution (maybe? You can't tab-focus disabled input fields, can you?) but I think behaviour-wise there is no difference.

PR #10026, opened 2019-02-12
Save IP address used for sign-up, not only sign-in

### LINKED ISSUE #9995: Show "Most recent IP" for accounts that aren't confirmed yet
### Pitch

When an account is created but not yet confirmed, the IP of the person behind it is not shown. It should be possible and, IMO, a good idea to show it for every accounts

### Motivation

Spam bots create a lot of accounts and don't confirm them all at the same time. However I can't recognize patterns in the IP address if most of the accounts created by the spam bot haven't been confirmed.

ClearlyClaire:
Fixes #9995

### DIFF
--- a/app/controllers/auth/registrations_controller.rb
+++ b/app/controllers/auth/registrations_controller.rb
@@ -28,6 +28,7 @@ class Auth::RegistrationsController < Devise::RegistrationsController
     resource.invite_code = params[:invite_code] if resource.invite_code.blank?
     resource.agreement   = true
 
+    resource.current_sign_in_ip = request.remote_ip if resource.current_sign_in_ip.nil?
     resource.build_account if resource.account.nil?
   end
 


app/controllers/auth/registrations_controller.rb
```diff
@@ -28,6 +28,8 @@ def build_resource(hash = nil)
     resource.invite_code = params[:invite_code] if resource.invite_code.blank?
     resource.agreement   = true
 
+    resource.last_sign_in_ip = request.remote_ip if resource.last_sign_in_ip.nil?
```
Gargron: I'm not sure there is any situation when `resource.last_sign_in_ip` is not nil here? Also, current_sign_in_ip is probably enough, don't see a need for last.

ThibG: Removed the assignment to `last_sign_in_ip`.

Gargron (approved):

PR #10036, opened 2019-02-13
Improve image description user experience

BenLubar:
- Add image descriptions to searchable post content.
- Allow multi-line image descriptions.

### DIFF
--- a/app/chewy/statuses_index.rb
+++ b/app/chewy/statuses_index.rb
@@ -31,7 +31,7 @@ class StatusesIndex < Chewy::Index
     },
   }
 
-  define_type ::Status.unscoped.without_reblogs do
+  define_type ::Status.unscoped.without_reblogs.includes(:media_attachments) do
     crutch :mentions do |collection|
       data = ::Mention.where(status_id: collection.map(&:id)).pluck(:status_id, :account_id)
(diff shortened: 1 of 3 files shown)


app/chewy/statuses_index.rb
```diff
@@ -50,7 +50,7 @@ class StatusesIndex < Chewy::Index
     root date_detection: false do
       field :account_id, type: 'long'
 
-      field :text, type: 'text', value: ->(status) { [status.spoiler_text, Formatter.instance.plaintext(status)].join("\n\n") } do
+      field :text, type: 'text', value: ->(status) { [status.spoiler_text, Formatter.instance.plaintext(status)].concat(status.media_attachments.map(&:description)).join("\n\n") } do
```
Gargron: I have to admit I don't remember how eager loading works in this class but you need to make sure this doesn't create n+1 queries. While the batch import is only used once, usually, the faster it completes, the better...

Cassolotl:
This sounds great! Could we see a screenshot please, since it's an accessibility UI thing?

Gargron:
There are no visual changes as far as I can tell

Gargron (approved):

PR #10037, opened 2019-02-13
Change robots.txt to exclude some URLs

Gargron:
- Exclude static assets
- Exclude uploaded files
- Exclude alternate versions of the profile page
- Exclude media proxy URLs


ClearlyClaire (approved):

nightpool:
@Gargron what was the impetus of this change? this is really bad for archiving crawlers like archive.org that respect robots.txt

PR #10038, opened 2019-02-13
Change robots.txt to exclude only media proxy URLs

nightpool:
Reverts tootsuite/mastodon#10037

This change as written prevents archive.org from effectively archiving or displaying archived mastodon sites.

If the concern is googlebot, then we should add specific googlebot rules.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,5 +1,5 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 #
 # To ban all spiders from the entire site uncomment the next two lines:
-# User-agent: *
-# Disallow: /
+User-agent: *
+Disallow: /media_proxy/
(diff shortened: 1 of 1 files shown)


Gargron:
I am not sure many people would be thrilled about being archived by archive.org in the first place. Making this specific to Googlebot is not a good idea, because there is Bing, Yandex, etc etc. I guess it would be easier to add a whitelist entry for archive.org instead.

nightpool:
@Gargron what is the actual concern that motivated this change?

Gargron:
Okay, @nightpool provided compelling arguments about why artists might expect their art to show up in Google Image Search, and why the followers/following pages should be excluded via a noindex meta tag instead

nightpool:
(updated)

Gargron (approved):

PR #10040, opened 2019-02-14
Fix hashtag column not subscribing to stream on mount

### LINKED ISSUE #9895: Hashtag columns are not streaming
### Actual behaviour

The streaming connection of the WebUI hashtag columns seems to be disconnected immediately.
Perhaps I think that it has happened since #8904 and related fixes.

I am sorry that the report was delayed.

1. Click on any hashtag in WebUI and pinned column.
2. Make a post with that hashtag.

### Specifications

Mastodon v2.7.0

Gargron:
Fix #9895

### DIFF
--- a/app/javascript/mastodon/features/hashtag_timeline/index.js
+++ b/app/javascript/mastodon/features/hashtag_timeline/index.js
@@ -41,15 +41,19 @@ class HashtagTimeline extends React.PureComponent {
 
   title = () => {
     let title = [this.props.params.id];
+
     if (this.additionalFor('any')) {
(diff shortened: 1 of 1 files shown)


mayaeh:
I tried this on my test server and it seemed to be fine.
Thank you 🙏

mayaeh:
However after a while the disconnected mark (...) was displayed at hashtag column, so I check back for a while.

mayaeh:
> It seems that a disconnected mark "..."

Sorry for the ambiguous expression.
It is "load-more" "load-gap" class's button.

And even though this is displayed the column has been updated.
The streaming connection seems to be continuing.

mayaeh (approved):
#9895 's problem seems to be solved by this, and I don't see errors in particular, so I approve this.

PR #10041, opened 2019-02-14
Filter incoming Announce activities by relation to local activity

Gargron:
Reject if announcer is not followed by local accounts, and is not
from an enabled relay, and the object is not a local status

Follow-up to #10005


ClearlyClaire (approved):
I'd still allow fetching `Announce` since other software might expose the `Announce` URI rather than that of the announced object itself.

Otherwise it looks fine to me.

PR #10042, opened 2019-02-14
Add tight rate-limit for API deletions

Gargron:
Deletions take a lot of resources to execute and cause a lot of federation traffic, so it makes sense to decrease the number someone can queue up through the API.

30 per 30 minutes


ykzts (approved):

saper:
https://mastodon.social/@Gargron/101588449409740014 for more background information

PR #10043, opened 2019-02-14
Fix favourite notification without status

### LINKED ISSUE #63: 被お気に入り通知において、お気に入りされたtootの内容がまれに表示されない
https://imastodon.net/@aqua_cat/3007723

image

- 被お気に入り通知において、お気に入りされたtootの内容がまれに表示されない
- お気に入りされたtoot( https://imastodon.net/@aqua_cat/2990044 )は現在も存在する
- その場での通知では表示されるが、後から開いた場合には表示されない

状況証拠からすると、

- /api/v1/streaming/notificationsを通じて取得した場合は問題なく表示され、
- /api/v1/notifications から取得した場合に元tootの取得に失敗する場合がある

ものと思われる

### LINKED ISSUE #2865: Empty favourite notification received on mastodon.social (local->local)
The last toot from the thread:

https://mastodon.social/@applecandy/5138248

has been favourited by myself ([@saper@mastodon.social](http://mastodon.social/@saper)) few hours later.

It appeared as an empty notification on the toot author's side:

Empty fav notification received

The client used above was [Amaroq](https://github.com/ReticentJohn/Amaroq) (by @ReticentJohn)

From the web it is more confusing:

Mixed up fav notification

I didn't favourite that one at all.

* * * *

- [X] This bug happened on mastodon.social 1.3.3 (as of `Sun May  7 00:39:29 UTC 2017`) between two local users
- [X] Permission to show other users' screens and toots granted

rinsuki:
Fix this.
image

close https://github.com/imas/mastodon/issues/63 and maybe fix https://github.com/tootsuite/mastodon/issues/2865


app/services/favourite_service.rb
```diff
@@ -14,10 +14,12 @@ def call(account, status)
 
     return favourite unless favourite.nil?
 
-    favourite = Favourite.create!(account: account, status: status)
+    Favourite.transaction do
```
Gargron: In NotifyService, we have a check if the activity still exists, I believe it serves the same purpose as wrapping it in a transaction. It should be noted that putting more code into a transaction is not great for performance. ReblogService uses an async worker for NotifyService, and for consistency it might make sense to do the same here instead of using a transaction.

SuperSandro2000:
bump

app/services/favourite_service.rb
rinsuki: Sorry for not answering for a long time.
I tried to do the same thing with async worker several times, but it didn't work.
I think this issue is a timing issue as shown in the image below.
image

I believe this issue can only be resolved by either:

- Combine favorite creation process and notification creation process into one transaction to prevent users from deleting favorites until notification is created
- Add a foreign key constraint on the database side, and fail to insert notification to deleted favorites

And the latter method requires a (potentially large) migration to the notifications table, which can take a long time on a large instance.

Therefore, I think that the former method has no choice but to solve.

Please let me know if there is anything I missed.

PR #10057, opened 2019-02-15
Fix mutes, blocks, domain blocks and follow requests not paginating

Gargron:
Regression from #9581


nightpool:
are there any other uses of the ScrollableList component that rely on the old behavior?

Gargron:
No, I did a search and modified all.

ClearlyClaire (approved):

PR #10059, opened 2019-02-15
Change mention and hashtag links to have target=_blank by default

### LINKED ISSUE #8490: Hashtag links inside embeds don't work
Hashtag links try to open inside the iframe which just leaves it empty. They should open in a new window instead.

* * * *

Gargron:
Fix #8490

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -290,10 +290,10 @@ class Formatter
   end
 
   def hashtag_html(tag)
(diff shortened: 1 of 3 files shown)


nightpool:
not sure this is the right place to make this change. why not rewrite the links on the client side? this won't fix pleroma hashtags, for example

Gargron:
You are kind of right, but at the same time, remote statuses are not embeddable right now.

ykzts (approved):

PR #10063, opened 2019-02-16
Add `tootctl statuses remove` to sweep unreferenced statuses

### LINKED ISSUE #1554: Sweep up old federated toots from local DB
#875 talks about removing old toots for users, but this is more about keeping down DB size on instances. Would it be sensible to sweep up old toots from the fediverse on a regular basis? 

Perhaps any toot older than X days (configurable) that doesn't mention a local user, or was boosted or replied to by a local user, could be removed by a rake task similar to `rake mastodon:feeds:clear`.

Any thoughts?

* * * *

Gargron:
Query by @tateisu

Fix #1554

For discussion:

- Name of command
- Interface of command, e.g. options
- Verbosity of command
- The query itself

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -6,6 +6,7 @@ require_relative 'mastodon/emoji_cli'
 require_relative 'mastodon/accounts_cli'
 require_relative 'mastodon/feeds_cli'
 require_relative 'mastodon/settings_cli'
+require_relative 'mastodon/statuses_cli'
 require_relative 'mastodon/domains_cli'
 require_relative 'mastodon/version'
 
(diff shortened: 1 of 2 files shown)


nightpool:
needs to handle statuses fetched by threadresolveworker (i.e. root statuses with replies from remote followed accounts).

tateisu:
even it does not have reply id,it will not deleted.
because it will contains mention to local user.
also toots from remote followed user.

nightpool:
user A (not followed by anybody) posts
user B replies to A (followed by local user C)

user A's status is fetched by thread resolve worker, but not excluded from this query.

tateisu:
` .where('id NOT IN (SELECT statuses1.in_reply_to_id FROM statuses AS statuses1 WHERE statuses.id = statuses1.in_reply_to_id)')  ` is excludes all statuses that have replies.

tateisu:
https://gist.github.com/tateisu/c3c13f2580961e86f79889a17f57260a original post about this query

ykzts (approved):

BenLubar:
A description more precise than `Remove statuses` might make more admins run it. I only learned what this actually does because I was curious, and I wouldn't have ever run it if I didn't see the query.

jelutz77:
Just ran this command, and it removed 10GB of about 70GB of data. I have 2 active users on my instance and it's only been running about a month. I spend entire days without interacting, and so does the other user. It seems to me that the command should free up MUCH more space than this?

PR #10065, opened 2019-02-17
Fix Announce activities of unknown statuses not fetching those statuses

Gargron:
Regression from #9998


ClearlyClaire (approved):
Why did I do that… the comment even disagreed with the code =_=
Sorry about that. This fix looks good to me.

PR #10086, opened 2019-02-20
correct opt-out showing application

hinaloe:
in #9994, api response of status was not correct opted-out _without save settings_. (`/@user/:status` is correct.)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -245,7 +245,7 @@ class User < ApplicationRecord
   end
 
   def shows_application?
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron:
I am confused. `User#shows_application?` calls the wrong variable?

hinaloe:
Oh, that right. User#shows_application? looks call wrong var.

I reverted change of serializer and fix User model.

Gargron (approved):

PR #10097, opened 2019-02-21
Make the column header of toot/thread view look like the others

ariasuni:
It makes the toot/thread column like the others column. The biggest benefit I see is that the back button is always at the same place.

screenshot_2019-02-21 mastodon dev

Edit: also for profile:

screenshot_2019-02-21 mastodon dev 2

### DIFF
--- a/app/javascript/mastodon/features/account/components/profile_column_header.js
+++ b/app/javascript/mastodon/features/account/components/profile_column_header.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ColumnHeader from '../../../components/column_header';
+import { FormattedMessage, injectIntl, defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  profile: { id: 'column_header.profile', defaultMessage: 'Profile' },
+});
+
+export default @injectIntl
+class ProfileColumnHeader extends React.PureComponent {
+
+  static propTypes = {
+    intl: PropTypes.object.isRequired,
+  };
+
+  render() {
+    const { intl } = this.props;
+
+    return (
+      <ColumnHeader
+        icon='user-circle'
+        title={intl.formatMessage(messages.profile)}
+        showBackButton
+      >
+      </ColumnHeader>
+    )
+  }
+}
--- a/app/javascript/mastodon/features/account/components/profile_column_header.js
+++ b/app/javascript/mastodon/features/account/components/profile_column_header.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ColumnHeader from '../../../components/column_header';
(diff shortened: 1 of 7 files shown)


ClearlyClaire (approved):
Looks good to me!
Something similar should be done for the account profile column.

ClearlyClaire (approved):

Gargron:
I am not a fan... I think the "Profile" and "Toots and replies" labels are useless information in those cases, and so are the icons :confused: 

The back button already stretches to the exact same spot as in the proposed design, so what is gained? In other words, you can already click in the same place in all column headers.

ThibG:
I think it makes it a little more explicit what kind of TL you're looking at (it might be good for newcomers and/or when scrolling long threads or accounts who boost a lot of stuff). Not the most useful information, but I think it can still help, and it's not really taking more space here. The main benefit is consistency.

You're right that you could already click in the same place in all column headers… but I actually hadn't noticed it.

lanodan:
> The back button already stretches to the exact same spot as in the proposed design

Well, it doesn’t *visually* appears to be:
screen
screen
screen

Also the indication between user and thread is actually useful if you have scrolled down (and the profile nor the detailed status is shown).

And consistency in design is something that should be pursued.

RX14:
I actually always assumed the back button being in a different location was a feature, because it is to me. The back button being in a different location allows you to quickly click back through a possibly lengthy series of opened posts and user profiles, stopping on the list/timeline you originally opened in the right column. Otherwise you have to click and make sure you haven't overshot and landed on "Getting started" accidentally. It's a small detail, but one I thought for years was rather intelligent design. But it seems it's accidental and has been gotten rid of (in glitch-fe, which is why I've spent an hour using git blame to find who to complain to, hopefully this helps show how much this feature means to me)

RX14:
To be clearer on my workflow: I use glitchfe on a vertical monitor, with 3 columns. The righermost "unpinned" column is also used as a flexible column which is showing a list most of the time, but is also obviously used for expanding posts in and navigating threads and user profiles. So this column pulls dual-duty.

After expanding a post from my timeline, reading the thread or replies, possibly clicking on a profile and otherwise navigating several levels deep in the "back" hierarchy, I'd like to exit out and return back to the rightermost column viewing a list. Before this commit, I can press on the back button in the left-hand side of the column repeatedly and the back hierarchy will be returned to the list view, before the back button moves to the right-hand side, causing future clicks to miss the back button. After this I'm forced to click slowly, and check the column title to make sure I don't "overshoot" the list column.

Gargron:
In conclusion, I stand by my comment of the labels conveying no new information. About consistency, there are columns that have a wide back button, and columns that have a header, depending on whether the header has useful information, and I believe that *is* consistent. Sorry. Thank you for your contribution @ariasuni, it is appreciated nonetheless.

PR #10100, opened 2019-02-23
Switch to Ubuntu for docker

Sir-Boops:
* Based on my image [Here](https://git.sergal.org/Sir-Boops/docker-mastodon/src/branch/master/Dockerfile)
* 200~ish MB smaller then the current build ( When built right )
* Build with jemalloc for slightly less ram usage
* /Should/ be an exact drop in replacement ( I've been told by other that switched to using this image that's a complete drop in replacement and I don't see any reason it wouldn't be but another confirmation would be nice :smile:  )
* Only real downfall of this image is it has to be built using --squash ( docker beta feature ) or the image size will be huge from all the individual RUNs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,89 +1,126 @@
-FROM node:8.15-alpine as node
-FROM ruby:2.6-alpine3.9
+FROM ubuntu:18.04 as build-dep
 
-LABEL maintainer="https://github.com/tootsuite/mastodon" \
-      description="Your self-hosted, globally interconnected microblogging community"
+# Use bash for the shell
+SHELL ["bash", "-c"]
 
+# Install Node
+ENV NODE_VER="8.15.0"
+RUN	echo "Etc/UTC" > /etc/localtime && \
+	apt update && \
+	apt -y dist-upgrade && \
+	apt -y install wget make gcc g++ python && \
+	cd ~ && \
+	wget https://nodejs.org/download/release/v$NODE_VER/node-v$NODE_VER.tar.gz && \
+	tar xf node-v$NODE_VER.tar.gz && \
+	cd node-v$NODE_VER && \
+	./configure --prefix=/opt/node && \
+	make -j$(nproc) > /dev/null && \
+	make install
+
+# Install jemalloc
+ENV JE_VER="5.1.0"
+RUN apt -y install autoconf && \
+	cd ~ && \
+	wget https://github.com/jemalloc/jemalloc/archive/$JE_VER.tar.gz && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,89 +1,85 @@
...
-RUN addgroup -g ${GID} mastodon && adduser -h /mastodon -s /bin/sh -D -G mastodon -u ${UID} mastodon \
- && mkdir -p /mastodon/public/system /mastodon/public/assets /mastodon/public/packs \
- && chown -R mastodon:mastodon /mastodon/public
-
-COPY . /mastodon
-
-RUN chown -R mastodon:mastodon /mastodon
-
-VOLUME /mastodon/public/system
+FROM sirboops/nodejs:8.15.0 as node
```
Gargron: Why sirboops namespace instead of official images?

Sir-Boops: Just so everything is built using the same libs

Official images afaik use alpine/Deb not Ubuntu

Sir-Boops: If you want everything built here I can switch it over to (this)[https://git.sergal.org/Sir-Boops/docker-mastodon/src/commit/f5faec868dfdd3c23c17d9bb1fbb6c701325db77/Dockerfile]

Just means alot longer build times

Gargron: In for a penny, in for a pound. Let's switch to that. Just to have more confidence over long-term availability of these images.

Sir-Boops:
Updated to build everything locally

I haven't tested this exact build on a live server yet but everything built and assets compiled just fine locally :+1:

Dockerfile
```diff
@@ -1,89 +1,124 @@
...
+# Add more PATHs to the PATH
+ENV PATH="${PATH}:/opt/ruby/bin:/opt/node/bin:/opt/mastodon/bin"
+
+# Create the mastodon user
+RUN apt update && \
+	echo "Etc/UTC" > /etc/localtime && \
+	ln -s /opt/jemalloc/lib/* /usr/lib/ && \
+	apt -y dist-upgrade && \
+	apt install -y whois && \
+	addgroup --gid 991 mastodon && \
```
Gargron: Missing UID and GID as args, hardcoded 991

Gargron (changes requested):

Dockerfile
Sir-Boops: Added and set the default to 991

Gargron (approved):

ashleyhull-versent:
Thanks for the work @Sir-Boops. I agree with the local building method over prebaked images :)

PR #10106, opened 2019-02-24
Improved remote thread fetching

ClearlyClaire:
So far, Mastodon only fetches *ancestors* of newly-discovered toots, which means in order to make sure a thread is known by all your followers, you would need to boost the *last* status of the thread, instead (or in addition to) the first one.

This PR aims to change that by fetching a limited number of same-server toots for every newly-discovered remote toot, and announcing a limited number of public self-replies on toots. This way, provided that a thread is already complete, boosting its first status will cause remote Mastodon servers to fetch its self-replies, and so on, fetching the whole thread.

### DIFF
--- a/app/serializers/activitypub/note_serializer.rb
+++ b/app/serializers/activitypub/note_serializer.rb
@@ -13,6 +13,8 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
   has_many :media_attachments, key: :attachment
   has_many :virtual_tags, key: :tag
 
+  has_one :replies, serializer: ActivityPub::CollectionSerializer
+
   def id
     ActivityPub::TagManager.instance.uri_for(object)
   end
@@ -31,6 +33,13 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
 
   def content_map
     { object.language => Formatter.instance.format(object) }
+  end
+
+  def replies
+    ActivityPub::CollectionPresenter.new(
+      type: :unordered,
+      items: object.self_replies(5).map(&:uri)
+    )
   end
 
   def language?
--- a/app/serializers/activitypub/note_serializer.rb
+++ b/app/serializers/activitypub/note_serializer.rb
@@ -38,7 +38,11 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
   def replies
     ActivityPub::CollectionPresenter.new(
       type: :unordered,
-      items: object.self_replies(5).map(&:uri)
+      first: ActivityPub::CollectionPresenter.new(
+        type: :unordered,
+        page: true,
+        items: object.self_replies(5).map(&:uri)
+      )
     )
   end
 
--- a/app/serializers/activitypub/note_serializer.rb
+++ b/app/serializers/activitypub/note_serializer.rb
@@ -41,7 +41,7 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
       first: ActivityPub::CollectionPresenter.new(
         type: :unordered,
         page: true,
-        items: object.self_replies(5).map(&:uri)
+        items: object.self_replies(5).pluck(:uri)
       )
     )
   end
--- a/app/services/activitypub/fetch_replies_service.rb
+++ b/app/services/activitypub/fetch_replies_service.rb
@@ -3,13 +3,16 @@
 class ActivityPub::FetchRepliesService < BaseService
   include JsonLdHelper
 
-  def call(parent_status, collection_or_uri)
+  def call(parent_status, collection_or_uri, allow_synchronous_requests = true)
     @account = parent_status.account
+    @allow_synchronous_requests = allow_synchronous_requests
 
     @items = collection_items(collection_or_uri)
     return if @items.nil?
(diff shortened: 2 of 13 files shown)


app/serializers/activitypub/note_serializer.rb
```diff
@@ -33,6 +35,13 @@ def content_map
     { object.language => Formatter.instance.format(object) }
   end
 
+  def replies
+    ActivityPub::CollectionPresenter.new(
+      type: :unordered,
+      items: object.self_replies(5).map(&:uri)
```
Gargron: I guess it's odd to hardcode limit this to 5. For example, there is this self-replies thread of every "good morning" since January 1st: https://mastodon.technology/@ashfurrow/101341748880807263 And it's a lot of items... But only listing 5 of them seems inaccurate, without a way to indicate that there's more.

app/services/activitypub/fetch_replies_service.rb
```diff
@@ -0,0 +1,51 @@
...
+      collection['orderedItems']
+    end
+  end
+
+  def filtered_replies
+    # Only fetch replies to the same server as the original status to avoid
+    # amplification attacks.
+
+    # Also limit to 5 fetched replies to limit potential for DoS.
+    @items.map { |item| value_or_id(item) }.reject { |uri| invalid_origin?(uri) }.take(5)
```
Gargron: Same comment here. I understand the DoS concern, and I don't know what to do here. 5 seems unrealistically low though. Perhaps I should run a SQL query to check average self-replies length on mastodon.social?

app/workers/activitypub/fetch_replies_worker.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ActivityPub::FetchRepliesWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: 'pull', retry: 3
+
+  sidekiq_retry_in do |count|
+    15 + 10 * (count**4) + rand(10 * (count**4))
```
Gargron: If this is shared between multiple workers, maybe extract it to a concern?

ThibG:
The number of immediate self-replies is usually low (and it is, in the example you've given me). The other toots in the thread will progressively be discovered when fetching each of the (up to 5) children

Le 25 février 2019 01:42:25 GMT+01:00, Eugen Rochko <notifications@github.com> a écrit :
>
>
>
>> +  def collection_items(collection)
>+    return unless collection.is_a?(Hash)
>+    case collection['type']
>+    when 'Collection', 'CollectionPage'
>+      collection['items']
>+    when 'OrderedCollection', 'OrderedCollectionPage'
>+      collection['orderedItems']
>+    end
>+  end
>+
>+  def filtered_replies
>+    # Only fetch replies to the same server as the original status to
>avoid
>+    # amplification attacks.
>+
>+    # Also limit to 5 fetched replies to limit potential for DoS.
>+    @items.map { |item| value_or_id(item) }.reject { |uri|
>invalid_origin?(uri) }.take(5)
>
>Same comment here. I understand the DoS concern, and I don't know what
>to do here. 5 seems unrealistically low though. Perhaps I should run a
>SQL query to check average self-replies length on mastodon.social?
>
>-- 
>https://github.com/tootsuite/mastodon/pull/10106#pullrequestreview-207183887

-- 

app/serializers/activitypub/note_serializer.rb
ThibG: Maybe it would be better to add something making explicit that the collection is incomplete, indeed.
But for the purposes of this PR, it doesn't need to list all replies, especially not replies that are not immediate replies.

app/services/activitypub/fetch_replies_service.rb
ThibG: For the purpose of thread-fetching, I think 5 is fine, as long as immediate self-replies are listed first. Indeed, if we take https://mastodon.technology/@ashfurrow/101341748880807263, the first status will list https://mastodon.technology/@ashfurrow/101358323548336310, which will itself list https://mastodon.technology/@ashfurrow/101358324411505893 and https://mastodon.technology/@ashfurrow/101364619725098480, and so on.

app/serializers/activitypub/note_serializer.rb
ThibG: I can't find a better attribute than `replies` for this, and I can't find an existing `Collection` attribute to indicate it's incomplete. But I think it's fine this way, if not completely satisfying.

ThibG: Hm, while I do understand your suggestion of using a paginated `Collection` not restricted to a hardcoded limit of 5 self-replies, I think this would make the code much more complex for no real benefit.
Indeed, going with your logic, `replies` should not be limited to self-replies either, so we would need to have up to 5 self-replies in the first page, *then* have the following page include both self-replies and other replies, excluding the 5 self-replies, etc. That would also require a new URL scheme, etc.
The `replies` collection is never going to be truly exhaustive anyway, we might exclude muted users, private posts, etc.

Gargron (approved):

PR #10109, opened 2019-02-24
Add a preferences API so apps can share basic behaviours

Gargron:
Resolve #9821

`GET /api/v1/preferences` example output:

```json
{
  "posting:default:visibility": "public",
  "posting:default:sensitive": false,
  "posting:default:language": null,
  "reading:expand:media": "default",
  "reading:expand:spoilers": false
}
```

### DIFF
--- a/app/controllers/api/v1/preferences_controller.rb
+++ b/app/controllers/api/v1/preferences_controller.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class Api::V1::PreferencesController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:accounts' }
+  before_action :require_user!
+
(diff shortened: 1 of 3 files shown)


ykzts (approved):

ykzts:
oops. this PR was draft

Gargron:
@Tak @connyduck Is this sufficient?

Tak:
@Gargron Sorry, it's not clear to me from the diff - does this also allow apps to _write_ these preferences?

A minor comment on the output format - if the keys are meant to be hierarchical, why not represent them that way in the json?
```"posting": {
  "default": [
    "visibility",
    "sensitive",
    "language"
  ]
},
"reading": { //...```

Tak:
Cool, good discussion.

PR #10111, opened 2019-02-25
Add polls

### LINKED ISSUE #1629: Polls
On twitter, I have used polls to write an interactive story together with @blinry. This is currently not possible with Mastodon.

I would like to start working on this though I am not sure if I will be able to finish.

* * * *

<details>
<summary>
I searched or browsed the repo’s other issues to ensure this is not a duplicate:
</summary>
screenshot
</details>

Gargron:
Fix #1629 

___

TODO:

- [x] Database schema
- [x] Attach to statuses model
- [x] Serialize to REST API
- [x] REST API for creating
- [x] REST API for voting
- [x] Prevent double-voting
- [x] REST API for retrieving results
- [x] Serialize poll to ActivityPub
- [x] Serialize voting to ActivityPub
- [x] Parse remote polls from ActivityPub
- [x] Handle incoming votes
- [x] Fetch latest results of remote polls from ActivityPub on demand
- [x] Web UI component for displaying polls
- [x] Tests

___

TODO in separate PRs:

- Web UI component for creating polls


tateisu:
How do we see this from the ActivityPub?

There is a version of friends.nico as an existing implementation.
In the case of friends.nico, voting has time limit. The result was output as another post.

Gargron:
>How do we see this from the ActivityPub?

In my understanding, when a poll is attached to a status, the status should have type `Question` instead of `Note`, and contain `oneOf`/`anyOf`/`closed` properties. I do not know yet how exactly the vote counts would be represented. Perhaps the items inside `oneOf`/`anyOf` would have a `replies` property containing a `Collection` with `totalItems`...

> In the case of friends.nico, voting has time limit.

Represented here by `expires_at` property (optional). Although perhaps it should not be optional, who ever needs infinite polls?

app/serializers/activitypub/note_serializer.rb
```diff
@@ -164,4 +189,34 @@ def name
 
   class CustomEmojiSerializer < ActivityPub::EmojiSerializer
   end
+
+  class OptionSerializer < ActiveModel::Serializer
+    class RepliesSerializer < ActiveModel::Serializer
+      attributes :type, :total_items
```
ThibG: The ActivityStreams spec is actually not clear about that, but in all examples, the options are just `Object`s identified by their `name`, and voting is simply done by sending an `Object` with the same `name` as the option `inReplyTo` the `Question`.
I think we should do the same.
Also, we should work with Pleroma to make sure we are interoperable, as they are currently investigating about polls as well.

app/serializers/activitypub/note_serializer.rb
```diff
@@ -13,12 +13,17 @@ class ActivityPub::NoteSerializer < ActiveModel::Serializer
   has_many :media_attachments, key: :attachment
   has_many :virtual_tags, key: :tag
 
+  has_many :poll_loaded_options, key: :one_of, if: :poll_and_not_multiple?
+  has_many :poll_loaded_options, key: :any_of, if: :poll_and_multiple?
+
+  attribute :closed, if: :poll_and_expires_at?
```
ThibG: Shouldn't this only be defined if the poll is *already closed*, not only if it has an expiration date set?

app/serializers/activitypub/note_serializer.rb
Gargron: Apparently Misskey already implemented federated polls, so that's what we should match. The voting is implemented like you said.

app/serializers/activitypub/note_serializer.rb
Gargron: We need to know in advance when it closes so users don't submit votes in vain. I think future date is fine, it is not a boolean after all

ThibG: It *can* be a boolean, and the spec says “Indicates that a question has been closed, and answers are no longer accepted.” https://www.w3.org/TR/activitystreams-vocabulary/#dfn-closed
It really seems weird to me to put a future date here :/

Gargron: Well, that sucks. How are we supposed to serialize this data then?

Gargron: `endTime`

app/serializers/activitypub/note_serializer.rb
```diff
@@ -164,4 +196,34 @@ def name
...
+      end
+    end
+
+    attributes :type, :name
+
+    has_one :replies, serializer: ActivityPub::NoteSerializer::OptionSerializer::RepliesSerializer
+
+    def type
+      'Note'
+    end
```
ThibG: I'm not sure why answer options would be Notes?

Gargron: It's in the example here: https://www.w3.org/TR/activitystreams-vocabulary/#dfn-question

app/serializers/activitypub/note_serializer.rb
Gargron: Should be noted Misskey serializes vote results as `_misskey_votes` property. I think using `replies` with `totalItems` is more in the spirit of the protocol so I will go with that :stuck_out_tongue:

app/lib/activitypub/activity/create.rb
```diff
@@ -209,6 +210,41 @@ def process_attachments
...
+      multiple: multiple,
+      expires_at: expires_at,
+      options: items.map { |item| item['name'].presence || item['content'] },
+      cached_tallies: items.map { |item| item.dig('replies', 'totalItems') || 0 }
+    )
+  end
+
+  def poll_vote?
+    return false if replied_to_status.nil? || replied_to_status.poll.nil? || !replied_to_status.local? || !replied_to_status.poll.options.include?(@object['content'])
+    replied_to_status.poll.votes.create!(account: @account, choice: replied_to_status.poll.options.index(@object['content']))
```
ThibG: Having `poll_vote?` actually process the vote is confusing.

app/serializers/activitypub/note_serializer.rb
```diff
@@ -181,4 +213,34 @@ def name
...
+  class OptionSerializer < ActiveModel::Serializer
+    class RepliesSerializer < ActiveModel::Serializer
+      attributes :type, :total_items
+
+      def type
+        'Collection'
+      end
+
+      def total_items
+        object.votes
```
ThibG: `object.votes_count`?

app/serializers/activitypub/vote_serializer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class ActivityPub::VoteSerializer < ActiveModel::Serializer
+  class NoteSerializer < ActiveModel::Serializer
+    attributes :id, :type, :content, :attributed_to,
+               :in_reply_to, :to
+
```
ThibG: The examples in the spec use `name` and not `content`. It also doesn't set a `type`: https://www.w3.org/TR/activitystreams-vocabulary/#dfn-question

I guess one upshot of doing that is to make sure you do not flood software that does not support this with vote `Note`s.

app/services/activitypub/fetch_remote_poll_service.rb
```diff
@@ -0,0 +1,44 @@
...
+        @json['anyOf']
+      else
+        @json['oneOf']
+      end
+    end
+
+    poll.update!(
+      expires_at: expires_at,
+      cached_tallies: items.map { |item| item.dig('replies', 'totalItems') || 0 }
+    )
```
ThibG: Same remark as earlier, what about fetching collections that aren't inlined?
Furthermore, if the poll *has* changed since it was initially fetched, this code will produce inconsistent tallies.

app/validators/poll_validator.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class PollValidator < ActiveModel::Validator
+  MAX_OPTIONS      = 4
+  MAX_OPTION_CHARS = 25
```
ThibG: Those two limits seem quite arbitrary. `4` for `MAX_OPTIONS` sounds a bit low to be honest.
Overall this should be fine for local statuses, but when interacting with others instances, this could be an issue.
We should at the very check what limits other implementations have, if any.

app/validators/poll_validator.rb
```diff
@@ -0,0 +1,19 @@
...
+  MAX_EXPIRATION   = 7.days.freeze
+  MIN_EXPIRATION   = 1.day.freeze
```
ThibG: Same goes for theses, tbh.

app/lib/activitypub/activity/create.rb
ThibG: Also, we could, at this point, send an `Update` of the polls' replies collection to other instances (with the same rules we use to forward replies, except we don't require LDSignatures) to update the tally.

ClearlyClaire (commented):
Overall, I like this PR, but I have some concerns. Mainly, we make a few arbitrary choices that may not work well with other implementations:
- as far as I can see, we don't allow for changing votes, that is, neither deleting existing votes, nor overwriting existing ones. If other implementations allow this, we don't have a correct error path to notify them that their vote is invalid. This may lead to inconsistent user experience.
- polls and media being mutually exclusive kind of makes sense with the current UI, but other software might have different interfaces and not have that limit
- the number of replies etc. are quite arbitrary and might also be an issue

An interesting thing is that polls being tied to a `Note`, they are only visible to the audience of that `Note`. However, the voting code does not seem to enforce that only the audience can vote. It might be worth tightening those restrictions.

Finally, we could send and support `Update` of the poll/its collections to update the tallies without the need for fetching changes.

app/controllers/api/v1/statuses_controller.rb
```diff
@@ -73,7 +73,6 @@ def set_status
     @status = Status.find(params[:id])
     authorize @status, :show?
   rescue Mastodon::NotPermittedError
-    # Reraise in order to get a 404 instead of a 403 error code
```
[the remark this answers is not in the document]
Gargron: Because it looks messy and doesn't add much information, the code itself already says that we replace a NotPermitted with a NotFound error.

app/lib/activitypub/activity/create.rb
```diff
@@ -6,7 +6,7 @@ def perform
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
-        return if delete_arrived_first?(object_uri)
+        return if delete_arrived_first?(object_uri) || poll_vote?
```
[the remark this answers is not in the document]
Gargron: We don't receive votes for remote polls. The delete-related method is not relevant for votes so it doesn't matter in which order they are executed.

app/lib/activitypub/activity/create.rb
```diff
@@ -209,6 +210,41 @@ def process_attachments
...
 
+  def process_poll
+    return unless @object['type'] == 'Question' && (@object['anyOf'].is_a?(Array) || @object['oneOf'].is_a?(Array))
+
+    expires_at = begin
+      if @object['closed'].is_a?(String)
+        @object['closed']
+      elsif @object['closed'].is_a?(TrueClass)
+        Time.now.utc
+      else
```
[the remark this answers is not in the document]
Gargron: I guess you want me to turn it into `elseif !@object['closed'].is_a?(FalseClass)` so that anything except a literal `false` marks the poll as closed?

app/lib/activitypub/activity/create.rb
```diff
@@ -209,6 +210,41 @@ def process_attachments
...
+      multiple = false
+      items    = @object['oneOf']
+    end
+
+    Poll.new(
+      account: @account,
+      multiple: multiple,
+      expires_at: expires_at,
+      options: items.map { |item| item['name'].presence || item['content'] },
+      cached_tallies: items.map { |item| item.dig('replies', 'totalItems') || 0 }
```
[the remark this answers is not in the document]
Gargron: I don't think it's worth it..?

app/controllers/api/v1/statuses_controller.rb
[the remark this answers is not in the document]
ThibG: Alright

app/models/poll.rb
```diff
@@ -0,0 +1,89 @@
...
+#  hide_totals     :boolean          default(FALSE), not null
+#  votes_count     :bigint(8)        default(0), not null
+#  last_fetched_at :datetime
+#  created_at      :datetime         not null
+#  updated_at      :datetime         not null
+#
+
+class Poll < ApplicationRecord
+  belongs_to :account
+  belongs_to :status, optional: true
```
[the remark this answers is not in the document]
Gargron: I am not sure which way is better. I used the flow we already established for media attachments, but admittedly, since no extra uploads are involved, it could just be part of PostStatusService

@nightpool @ykzts Thoughts?

app/lib/activitypub/activity/create.rb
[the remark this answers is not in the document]
ThibG: You can't know whether you receive vote for remote polls. You don't send them, but nothing forbids other implementations to send them.

I was talking about returning before entering the lock, but now I understand that `poll_vote?` also performs side effects (which I don't think it should do)

app/lib/activitypub/activity/create.rb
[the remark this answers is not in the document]
ThibG: I'm not 100% sure of it yet, but yeah, I think that would make the most sense.

app/lib/activitypub/activity/create.rb
Gargron: Wasted effort, in my opinion. Here is why: We can't guarantee that an Update will be sent to everyone who ever came across the poll. We do this for normal replies because previously they were missing altogether, and if some branches are not visible, that's not that big of a deal. However, in case of polls, if some votes are missing, that's a gross misrepresentation of the poll and makes it useless/harmful. So, either way, we need to go and download the latest tallies from the origin. Since we need to do that anyway, I see messing with this whole Update business as wasted effort.

app/serializers/activitypub/vote_serializer.rb
Gargron: I believe Misskey sends them as content? cc @syuilo

Also your upshot is also a downshot. Since the content is the poll option text, it works for the human observers even if it's not officially interpreted as a vote.

app/services/activitypub/fetch_remote_poll_service.rb
Gargron: >Furthermore, if the poll has changed since it was initially fetched, this code will produce inconsistent tallies.

I don't understand what you mean. This is fetched from the origin, for remote polls. The origin has the most correct count of votes.

app/validators/poll_validator.rb
Gargron: 4 is the max on Twitter, and I am inclined to agree. Anything else turns your home feed into a survey.

@syuilo I can't find validations in Misskey, what limits are there?

app/services/activitypub/fetch_remote_poll_service.rb
ThibG: Sorry, I meant that if the list of options changed (which nothing prevents), the updated tallies are going to be garbage.

app/validators/poll_validator.rb
Gargron: They make sense to me.

app/services/activitypub/fetch_remote_poll_service.rb
Gargron: I don't think updating the options of published polls should be allowed. That invalidates the results.

app/lib/activitypub/activity/create.rb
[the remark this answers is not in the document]
Gargron: Well, it kind of massively sucks that poll votes don't have a dedicated type in ActivityPub. We can check if the replied-to post is known and has a poll, but if the replied-to post is unknown, we can't treat it like a vote.

app/validators/poll_validator.rb
acid-chicken: In Misskey, we can create up to 10 choices which have 50 characters or less. However, there is no limitations like that in notes which created by ActivityPub (maybe. I'll test about that).

app/serializers/activitypub/vote_serializer.rb
acid-chicken: Misskey is using `name` for now, but they don't have `type` property. Should we add `type` property or use `content` instead of `name`?

<details>
<summary>sample json</summary>
<pre><code>{
  "@context": [
    { "Hashtag": "as:Hashtag" }
  ],
  "type": "Question",
  "id": "https://emoji.msky.work/questions/5c7bfbc0d4ab42731132b980",
  "actor": "https://emoji.msky.work/users/5be70e1033638b16ae9c6469",
  "content": "1 choice poll test.",
  "oneOf": [
    { "name": "test", "_misskey_votes": 6 }
  ]
}
</code></pre>
</details>

Gargron: @acid-chicken I mean poll votes, do the votes use content or name?

app/validators/poll_validator.rb
Gargron: Okay, I think I will have soft limits for local polls (4) and higher hard limits (10) for remote ones.

app/lib/activitypub/activity/create.rb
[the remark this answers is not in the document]
ThibG: Yeah, that's also one of the reasons for which I would like to avoid the `Note` with `content` solution.

app/lib/activitypub/activity/create.rb
ThibG: If we get an `Update` of a poll with updated tallies and a `closed` attribute, we don't have to fetch anything.

ThibG: That being said, such an update mechanism doesn't have to be part of this PR, we can always add it later.

app/services/activitypub/fetch_remote_poll_service.rb
ThibG: You could reorder them or add requested options, I don't know. I think we should at least update tallies based on `name`, not on the option's position in the serialization.

app/validators/poll_validator.rb
ThibG: The thing is, why I can see why allowing many options may look bad in the UI, I can't see a reason to enforce such limitations on long-running polls or short-running ones.

ThibG:
I'm also thinking that, in the case we can't manage getting tallies from the remote server (because there is nothing standardized for that and the suggested solution is the `Question`'s `replies` thing, which we do not want), we should offer to go to the remote public page to see the results.

Also, aren't we missing code to display the poll on public pages?

app/serializers/activitypub/vote_serializer.rb
acid-chicken: @Gargron Voting activities are represented as Note, and they're using `content` in Misskey. (I'm sorry for miss reading!)

<details>
<summary>sample json</summary>
<pre><code>{
  "id": "https://misskey.m544.net/notes/5c7c152d34671559e850a1f8/activity",
  "actor": "https://misskey.m544.net/users/5bf0009a2d98942caa53e39e",
  "type": "Create",
  "published": "2019-03-03T17:55:57.525Z",
  "object": {
    "id": "https://misskey.m544.net/notes/5c7c152d34671559e850a1f8",
    "type": "Note",
    "attributedTo": "https://misskey.m544.net/users/5bf0009a2d98942caa53e39e",
    "summary": null,
    "content": "&lt;p&gt;&lt;span&gt;1&lt;/span&gt;&lt;/p&gt;",
    "_misskey_content": "1",
    "published": "2019-03-03T17:55:57.525Z",
    "to": [
      "https://emoji.msky.work/users/5be70e1033638b16ae9c6469"
    ],
    "cc": [],
    "inReplyTo": "https://emoji.msky.work/notes/5c7c1521ec5f6f5c1af0597e",
    "attachment": [],
    "sensitive": false,
    "tag": [
      {
        "type": "Mention",
        "href": "https://emoji.msky.work/users/5be70e1033638b16ae9c6469",
        "name": "@acid_chicken@emoji.msky.work"
      }
    ]
  },
  "to": [
    "https://emoji.msky.work/users/5be70e1033638b16ae9c6469"
  ],
  "cc": []
}
</code></pre>
</details>

app/models/status.rb
```diff
@@ -250,6 +255,8 @@ def decrement_count!(key)
   before_validation :set_conversation
   before_validation :set_local
 
+  before_save :set_poll_id
```
ThibG: We can use a normal association instead of this, now, can't we?

db/migrate/20190226003449_add_poll_id_to_statuses.rb
```diff
@@ -0,0 +1,5 @@
+class AddPollIdToStatuses < ActiveRecord::Migration[5.2]
+  def change
+    add_column :statuses, :poll_id, :bigint
```
ThibG: Same goes here, we can use a foreign key.

Gargron: I think foreign keys are an increasingly bad idea. It requires having an index on that column, otherwise DELETE or NULLIFY cascades take an incredible amount of time. Any index on statuses would also be massive.

app/models/status.rb
Gargron: We have this throwaway inverse reference to polls so that during eager-loading we can skip queries on the polls table when no statuses have polls, which I believe would be the most common case.

ClearlyClaire (approved):
I think `poll_vote?` is still very confusing, fallback for remote implementations not using our `replies` scheme should be considered, limits could be tweaked, and we could investigate sending `Update`s and fetching the size of `replies` collections identified by an URI instead of being inlined, but that's all stuff that can be done later.

mattcoxonline:
Do you have a design for the creation of polls in the web UI, @Gargron ? Want me to create some mocks?

Gargron:
@mattcoxonline I do not have a design. You can give it a go...

PR #10125, opened 2019-02-27
Fix mention processing for unknwon accounts on incoming ActivityPub Notes

ClearlyClaire:
`::FetchRemoteAccountService` is not `ActivityPub::FetchRemoteAccountService`,
its second argument is the pre-fetched body. Passing `id: false` actually passed
a `Hash` as the prefetched body, instead of properly resolving unknown remote
accounts.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -159,7 +159,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     return if tag['href'].blank?
 
     account = account_from_uri(tag['href'])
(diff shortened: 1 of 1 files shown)


ThibG:
Another solution would be to revert ef69c655cc I guess. I'm not sure exactly what it solves?

Gargron:
> I'm not sure exactly what it solves?

href is not id, therefore it's not necessarily the verbatim URI of the account. At least that's what I think it was about. Or maybe I was getting errors about an unknown constant so I fixed it but picked the wrong namespace?

Gargron (approved):

PR #10131, opened 2019-03-01
Fix public timelines being broken by new toots when they are not mounted

Gargron:
Regression from #6876 

The condition was returning true even when the timeline has not been loaded once before. Once a toot was posted, it would then be inserted into the timeline, and when the timeline is then opened, that toot would be sitting there alone, making it so the *actual* timeline loading would fetch only toots older than that toot, creating unnatural gaps.

This was a bad look for new users especially.

### DIFF
--- a/app/javascript/mastodon/actions/compose.js
+++ b/app/javascript/mastodon/actions/compose.js
@@ -158,7 +158,9 @@ export function submitCompose(routerHistory) {
       // into the columns
 
       const insertIfOnline = timelineId => {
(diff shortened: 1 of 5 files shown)


ThibG:
It certainly reduces issues, but what about when streaming fails, the timeline is updated, and the `null` placeholder is thus removed? There will be a gap between the last fetched status and a newly-posted one.
Maybe we should insert it if and only if the streaming works, otherwise trigger an `expandTimeline` if the timeline is mounted, and otherwise do nothing.

ClearlyClaire (approved):

PR #10141, opened 2019-03-03
Correctly make polls and media mutually exclusive

ClearlyClaire:


Gargron (approved):

JamesABDL:
What's the reason for polls and media to be mutually exclusive?

Had a [user question it](https://abdl.link/@Kiddyscord/101906135798041645), and I couldn't find anything on the original issue which explains why.

PR #10155, opened 2019-03-04
Add non-JS fallback for polls on public pages

ClearlyClaire:
# When not voted

screenshot_2019-03-04 thib thib mastodev sitedethib com

# When voted

screenshot_2019-03-04 thib thib mastodev sitedethib com 1


ykzts (approved):
nice 👍

PR #10158, opened 2019-03-04
Store remote votes URI

ClearlyClaire:
This is useful if we ever plan to support remote users deleting their votes.

### DIFF
--- a/app/lib/activitypub/tag_manager.rb
+++ b/app/lib/activitypub/tag_manager.rb
@@ -35,6 +35,8 @@ class ActivityPub::TagManager
       account_status_url(target.account, target)
     when :emoji
       emoji_url(target)
+    when :vote
+      [uri_for(target.account), '#votes/', target.id].join
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/lib/activitypub/tag_manager.rb
```diff
@@ -35,6 +35,8 @@ def uri_for(target)
       account_status_url(target.account, target)
     when :emoji
       emoji_url(target)
+    when :vote
+      [uri_for(target.account), '#votes/', target.id].join
```
Gargron: Well, if they are going to be stored, might generate a real path for them?

ThibG: we could, but I'm not sure if that is useful, much like follows

Gargron: I don't disagree, but the follows &co are not in this method :confused: Perhaps you could move them here then?

ThibG: Changed it to use the same logic as `Follow` and cie

Gargron (approved):

PR #10160, opened 2019-03-04
When serializing polls over OStatus, serialize poll options to text

ClearlyClaire:


app/lib/formatter.rb
```diff
@@ -19,6 +19,10 @@ def format(status, **options)
 
     raw_content = status.text
 
+    if status.poll && options[:inline_poll_options]
```
Gargron: The other way around is better

ThibG: You mean, checking the option then the poll? Or have it on by default? Changed it for the former.

Gargron (approved):

PR #10161, opened 2019-03-04
Ensure only people allowed to see the poll can actually vote

ClearlyClaire:


app/policies/poll_policy.rb
```diff
@@ -2,6 +2,7 @@
 
 class PollPolicy < ApplicationPolicy
   def vote?
+    return false unless StatusPolicy.new(current_account, record.status).show?
```
Gargron: Just put it in the condition below with `&&` :stuck_out_tongue:

Gargron (approved):

PR #10163, opened 2019-03-04
Assume remote polls expire in a year when expiration date not set

ClearlyClaire:
This allows compatibility with remote polls without a set expiration date
without changing our internal logic.


Gargron:
Save for that `hide_totals` is incompatible with an infinite expiration date, what are we gaining here? What if they *are* infinite polls?

ThibG:
yeah, it's not useful if `hide_totals` is true… but it can't be, we aren't federating that information…

Gargron:
But that would be reflected in `closed`... Also while `hide_totals` does not federate, aren't the `totalItems` set to 0 when it's true?

ThibG:
Oh nevermind, I thought we required `expires_at` even for remote polls. That isn't the case.
The logic for displaying polls without an expiration date is broken though, but that's another issue.

PR #10184, opened 2019-03-05
Add UI for creating polls

Gargron:
grafik


mayaeh:
Is there a possibility that it will fail if it is five minutes?
422 error returned.

Gargron:
@mayaeh Fixed by #10186

mayaeh:
Thank you.
Other action seems to be no problem in the range I tried.
There seems to be no error.

mayaeh (approved):

PR #10186, opened 2019-03-06
Fix poll validation issues

Gargron:
- Fix missing interpolation argument in PollValidator
- Fix PollValidator rejecting exact allowed min/max durations


lib/active_record/validations.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ActiveRecord::Validations::ClassMethods
+  def validates_associated(association, options = {})
```
Gargron: Reference: https://content.pivotal.io/blog/alias-method-chain-validates-associated-informative-error-message

It's from 2008, but seems to work.

PR #10196, opened 2019-03-06
Add optimistic lock to avoid race conditions when handling votes

ClearlyClaire:
I'm not sure this is the most efficient solution to this issue, but it should work…


app/models/poll_vote.rb
```diff
@@ -32,5 +32,7 @@ def object_type
   def increment_counter_cache
     poll.cached_tallies[choice] = (poll.cached_tallies[choice] || 0) + 1
     poll.save
+  rescue ActiveRecord::StaleObjectError
+    retry
```
Gargron: Does this reload the tallies?

ThibG: Not sure. Updated it.

Gargron (approved):

PR #10197, opened 2019-03-06
i18n: 🇵🇱

mkljczk:


config/locales/pl.yml
```diff
@@ -757,6 +758,16 @@ pl:
...
     prev: Poprzednia
     truncate: "&hellip;"
+  polls:
+    errors:
+      already_voted: Już zagłosowałeś(-aś) w tym głosowanie
+      duplicate_options: zawiera powtarzające się opcje
+      duration_too_long: jest zbyt odległa
+      duration_too_short: jest zbyt bliska
+      expired: To głosowanie już zakończyło się
+      over_character_limit: nie mogą składać się z więcej niż %{MAX} znaków
```
Aditoo17:
```suggestion
      over_character_limit: nie mogą składać się z więcej niż %{max} znaków
```
%{MAX} was recently changed to %{max}, please change this line.

config/locales/pl.yml
```diff
@@ -757,6 +758,16 @@ pl:
...
+      too_few_options: musi zawierać przynajmniej dwie opcji
+      too_many_options: nie może zawierać więcej niż %{MAX} opcji
```
Aditoo17:
```suggestion
      too_many_options: nie może zawierać więcej niż %{max} opcji
```
See above.

Aditoo17 (changes requested):

config/locales/pl.yml
m4sk1n: thanks

Aditoo17 (approved):

m4sk1n:
tbh i’m unable to normalize it, could somebody do it for me?

Gargron:
Can you submit the changes over Weblate instead?

m4sk1n:
ok

Aditoo17:
I saw that you added some strings which aren't in the `app/javascript/mastodon/locales/en.js` file yet. Might that be the cause?

m4sk1n:
afaik `check-i18n` is for /config/locales/ files

PR #10209, opened 2019-03-07
Support pushing and receiving updates to poll tallies

ClearlyClaire:
- Accept `Update` activities on `Question` objects, treating them the same way as polling those `Question`s
- After a vote to a local poll with `hide_totals` unset has been processed, schedule a job to issue an `Update` to the audience of the poll (every mentioned user, everyone who has boosted the toot, everyone who has voted, and, if the poll is not a direct message, every follower). Only one job at a time can exist.
- At the creation of a poll, schedule a task to notify its owner and send an `Update` to the people listed above.

screenshot_2019-03-08 dev instance 1

### DIFF
--- a/app/javascript/mastodon/actions/notifications.js
+++ b/app/javascript/mastodon/actions/notifications.js
@@ -92,7 +92,7 @@ export function updateNotifications(notification, intlMessages, intlLocale) {
 const excludeTypesFromSettings = state => state.getIn(['settings', 'notifications', 'shows']).filter(enabled => !enabled).keySeq().toJS();
 
 const excludeTypesFromFilter = filter => {
-  const allTypes = ImmutableList(['follow', 'favourite', 'reblog', 'mention']);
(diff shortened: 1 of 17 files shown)


ThibG:
Just noticing that this PR doesn't address sending the final `Update` when `hide_totals` is set.
Also, it might make sense to explicitly include all voters.

Cassolotl:
Does this mean that if you boost or vote in a poll you get a notification of the results when it closes? :o If so, I love that.

ThibG:
@Cassolotl no, it only notifies the author of the poll. We *could* extend that to voters or people who boosted it, but I'm not sure we should do that (and it would be slightly trickier).

Also, we need to change the code so that the new notification type isn't sent to unsupporting clients (it breaks some of them).

ThibG:
Extended the notification to apply to all users who have voted, should handle both local and remote polls

Gargron (approved):

PR #10210, opened 2019-03-07
config: add DB_SSLMODE for managed/remote PG

ghost:


Gargron (approved):

Gargron:
Presumably streaming/index.js needs support for the same variable?

sascha-sl:
I did notice that too now; it actually defaulted to `off`/`none`.

`prefer` should be a much nicer default, though I don't have an unencrypted DB to test if there's any node weirdness involved.

PR #10212, opened 2019-03-07
Re-organize Dockerfile to improve incremental builds.

BenLubar:
- Always run apt update before any other apt command. (This fixes
  incremental builds failing if a remote package is updated.)
- Only copy dependency lists before installing dependencies. (This means
  editing code doesn't force all dependencies to be re-downloaded.)
- Delete cache in the same layer that it is created. (Otherwise,
  deleting cache *increases* the size of the image on non-squashed
  builds.)
- Move the installation of some static dependencies to *before* Mastodon
  code is imported to Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,8 @@ RUN	echo "Etc/UTC" > /etc/localtime && \
 
 # Install jemalloc
 ENV JE_VER="5.1.0"
-RUN apt -y install autoconf && \
+RUN apt update && \
+	apt -y install autoconf && \
 	cd ~ && \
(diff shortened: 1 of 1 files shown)


Gargron:
I believe this is the kind of back-and-forth I was warning against @Sir-Boops #10100

BenLubar:
This runs all the same code, just in a different order that's friendlier to the build cache.

Sir-Boops:
I question the weird way this COPYs the masto src around but other then that It's a nice cleanup UwU


I don't know what you mean about 'back-and-forth'

Gargron (approved):

PR #10219, opened 2019-03-08
Parse query string for the database URL

zunda:
Make it possible for the streaming process to connect to Postgres with SSL enabled when `DATABASE_URL` has `?ssl=true` query.


zunda:
Taken care of with #10210. Closing

PR #10225, opened 2019-03-09
Fix SSL configuration regression in streaming API

### LINKED ISSUE #10223: Timeline streaming stopped after #10210
### Actual behaviour
Streaming stopped because connection from node to postgresql become SSL even if DB_SSLMODE is not set. (regression from #10210)

Start streaming service with postgresql which works non ssl connections. I got error log, and streaming stopped.

```
ERR! Error: The server does not support SSL connections
```

There is not DB_SSLMODE in my .env.production. As far as I investigate, "ssl" option of pg.Pool() of node not has "sslmode" option. Probabry, just only add "ssl" key, and even if it is empty, connection to postgresql become SSL.

### Specifications
- mastodon: master (3f4a6d4)
- My server is all-in-one that is contains web, streaming, sidekiq, redis and postgresql in single.

Gargron:
Fix #10223

### DIFF
--- a/streaming/index.js
+++ b/streaming/index.js
@@ -45,8 +45,8 @@ const dbUrlToConfig = (dbUrl) => {
 
   const ssl = params.query && params.query.ssl;
 
-  if (ssl) {
-    config.ssl = ssl === 'true' || ssl === '1';
+  if (ssl && ssl === 'true' || ssl === '1') {
+    config.ssl = true;
   }
 
   return config;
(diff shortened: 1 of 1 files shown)


kedamaDQ:
Thanks! Thanks a lot!!

kedamaDQ:
Please wait. Just with the "ssl" key, node forcedly seemed to connect with ssl.

Gargron:
Even if `false`? :confused:

kedamaDQ:
I will check it.

kedamaDQ:
As far as I confirmed, the situation did not change. Just with the key "ssl", streaming stops, it seems that an error is output to the log.

I done in development environment:

```
# git fetch upstream pull/10225/head:check
# git checkout check
# foreman start
```

But situation did not change... :thinking:

Gargron:
@kedamaDQ I'm going to try passing `ssl: undefined`. Let me know if it works.

kedamaDQ:
Just a moment please. I am midnight in my country ... I'm sorry.

ykzts (approved):

kedamaDQ:
I'm sorry, I forgot to delete "DB_SSLMODE=prefer" from my .env ...
I deleted "DB_SSLMODE" and confirmed to solved the problem with either "false" or "undefined".

Thank you for your response!

kedamaDQ:
Oops...sorry I have probrems yet...
スクリーンショット 2019-03-10 15 23 33

Issue #10229, opened 2019-03-10
Suspending remote users do not prevent them from seeing private toots


ThibG:
### Expected behaviour

When a local user `@A` has remote followers `@B@remote` and `@C@remote` hosted on the same remote instance, and `@A`'s admins decide to suspend the remote account `@C@remote`, `@B@remote` should still see new follower-only toots by `@A` but `@C@remote` should not receive them anymore.

### Actual behavior

`@C@remote` is locally removed from `@A`'s list of followers, but the remote instance is not notified, thus still showing `@C@remote` toots from `@A` whenever `@B@remote` receives them.

Furthermore, the local instance having removed `@C@remote` from `@A`'s followers, `@A` is not aware that `@C@remote` still get their private toots and cannot act on it.

kaniini:
This is why I keep saying that the followers collection needs to be expanded in transit.

Gargron:
@kaniini That's not practical with 2,000, 5,000 or 210,000 followers :confused: Although if you expand only followers from the server you are delivering too, it reduces the number a little, but it can still be in the thousands. Now, if it was possible to send a *bloom filter*, then we're talking...

We definitely need to send out Reject-Follow for remote users on suspension though.

Gargron:
> Now, if it was possible to send a bloom filter, then we're talking...

To expand more on this, it doesn't solve the issue of how computationally taxing it would be to process individual followers into a rendered JSON array or a bloom filter (we operate mainly on shared_inbox_url so that saves a lot of workload currently), but at least a bloom filter is very compact, so the transmitted JSON would not be megabytes-big. And likewise, on the processing side, you merely need to check for each local follower (which already happens), whether that follower is definitely not in the bloom filter (quick operation).

But of course, none of that is in the ActivityPub spec, so if even if everyone thinks it's a good idea, we'd need to invent it ourselves.

kaniini:
I'm down for federating with bloom filters for addressing.  Will try to mock up a spec later this weekend.  Sorry, didn't catch this until now.

mjankowski:
Last update here was ~5 years ago ... is this still relevant to latest (4.3) version, and if so is there anything we can practically do to improve things? (or any related WIP PRs to check out)

trwnh:
#10230 partially fixes this but the real problem is sending non-public activities to sharedInbox. when you send to sharedInbox you are depending on the receiving server to know what to do with that activity. default spec behavior is to treat that as a collection of public activities.

[...]

PR #10230, opened 2019-03-10
Reject existing Follows when suspending a remote account

ClearlyClaire:
Partial fix to #10229 

This is only a partial fix because it addresses *new* remote user suspensions and not existing ones, for which the list of followers is already lost


ThibG:
(My dev environment is currently broken, so I haven't actually tested this)

Gargron (approved):
Looks correct to me

PR #10239, opened 2019-03-11
Squish username before validation

### LINKED ISSUE #10101: Trim whitespace in username
### Pitch

The username field during registration and login should allow a whitespace **postfix**.

This is important because many mobile input methods (on-screen keyboards) insert
a whitespace each time you complete a word, so sometimes a whitespace sneaks in
where it shouldn't (after your username).

The server could easily treat the string as if it didn't have a whitespace postfix.
Any user is almost guaranteed not to have meant anything by it. Whitespace is even
explicitly disallowed in usernames.

The thing is that after your keyboard has
automatically added a whitespace and you left the input field, you can't even
see that whitespace, so you will just get confused why mastodon isn't accepting
your new username.

aurelien-reeves:
Fix #10101

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -78,6 +78,10 @@ class Account < ApplicationRecord
   validates :note, note_length: { maximum: 160 }, if: -> { local? && will_save_change_to_note? }
   validates :fields, length: { maximum: 4 }, if: -> { local? && will_save_change_to_fields? }
 
+  before_validation(on: :create) do
+    self.username = username&.squish
+  end
+
   scope :remote, -> { where.not(domain: nil) }
   scope :local, -> { where(domain: nil) }
   scope :expiring, ->(time) { remote.where.not(subscription_expires_at: nil).where('subscription_expires_at < ?', time) }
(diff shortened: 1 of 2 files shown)


app/models/account.rb
```diff
@@ -78,6 +78,10 @@ class Account < ApplicationRecord
   validates :note, note_length: { maximum: 160 }, if: -> { local? && will_save_change_to_note? }
   validates :fields, length: { maximum: 4 }, if: -> { local? && will_save_change_to_fields? }
 
+  before_validation(on: :create) do
```
Gargron: Style-wise, it would be preferable to do it via a private method

`before_validation :prepare_username, on: :create`

I am also not sure what `squish` gives here over `strip`, would prefer to just use `strip`

Gargron (changes requested):

app/models/account.rb
aurelien-reeves: Indeed, I will move `before_validation` into a private method.

Regarding `squish`, actually it processes unicode whitespaces. `strip` does not.

```ruby
pry(main)> "\u3000test\u00a0".strip
=> " test "
pry(main)> "\u3000test\u00a0".squish
=> "test"
```

If you agree to use `squish` to process Unicode white-spaces, I will add a dedicated test for that, or at least I will add Unicode white-spaces into the test I have added.

Gargron (approved):

PR #10255, opened 2019-03-13
Add UI for enabling/disabling poll notifications

ClearlyClaire:


Aditoo17:
Could you also add poll notifications to the quick filter bar?

Gargron:
I'm wondering if the strings are understandable enough for the end-user. Polls - why polls? Maybe call it "Poll results" or something like that? What does the notification text say when you get a notification like that?

ThibG:
Fair point, “poll results” is probably a better name.

Here is what it looks like:
Screenshot_2019-03-13 Dev instance(1)

And looking at it, I realize it hasn't been updated since the point such notifications were only for your own polls…

Gargron:
Okay! So "Poll results", and also "A poll you voted in has ended"

ThibG:
Screenshot_2019-03-13 Dev instance(2)

Gargron (approved):

PR #10260, opened 2019-03-14
Disable real-time updates on public pages to improve readability

Gargron:
Better to only let the user scroll back in time than to constantly update masonry layout.


trwnh:
Could realtime updates for public pages be an opt-in toggle per session? I can imagine some people might want to preview public streams.

ClearlyClaire (approved):

PR #10261, opened 2019-03-14
Fix detailed poll validation errors not being returned in the API

Gargron:
No more "Owned poll is invalid"


ClearlyClaire (approved):
I'm not too familiar with ActiveRecord nested attributes, but LGTM.

PR #10268, opened 2019-03-14
Add relationship manager UI

Gargron:
image

Ultimately, the authorized followers page, listing domains of all followers, was not very useful. On the other hand, removing more than one follower through the web UI was very involved, because it requires a click-through to each profile, switching from the list of accounts and back.

The new "Follows and followers" page is more mature. It lists who you follow, or all followers individually, while letting them be filtered (e.g. show only mutuals, show people who have moved accounts), and more importantly, allowing batch actions to be applied. You can select multiple followers and remove them at once. Preserving the old functionality, you can also select one or more followers, and block the entire domains that they are from. You can also batch unfollow who you were following.

It also shows you when your follows/followers were last active.


ThibG:
Looks good! I think not being able to search or remove followers from a particular domain is an important regression, though.

Gargron:
If you select a follower from a domain and click the second batch button, it will block the entire domain of that follower. Also, it was hard to find a specific domain in the list before, too, so I wouldn't call it a regression. I wanted to add a search form, however, we don't have ready-made components for search for end-users. We have the admin design but it's a bit clunky...

On the other hand, I just noticed I am not putting filter params `relationship`, `status` and `by_domain` into the form, which is a mistake that will affect redirects to the right page.

ThibG:
Oh, I did not notice the “Remove all followers from the selected domains” thing. This is suboptimal but it does cover that functionality, so why not.

As for a search form, I believe I had made a PR for that ;)

Gargron:
Yes, I didn't merge it because I was planning to do this. And as I mentioned:

> We have the admin design but it's a bit clunky...

I.e. my concern is that the input and search buttons interrupt the page flow in a way that may be confusing to non-technical users, and it's not super pretty. I would expect the search to look more like what we have in the list editing modal in the web UI, or the search form in the web UI, but of course, it's complicated because that's in React, and this is outside React.

ThibG:
The only real change between what we have for admin views and what we have in React is that the page gets reloaded, right?

ykzts (approved):

PR #10269, opened 2019-03-14
Migrate existing `open_registrations` setting to the new `registrations_mode`

### LINKED ISSUE #10263: Updating to #10250 or later automatically opens a new registration
### Pitch

When a server closing a new registration is updated to #10250 or later, a new registration opens automatically.
I think it would be better to set Setting.registrations_mode with reference to Setting.open_registrations when we run db:migrate or next time we start server.

### Motivation

If run Mastodon for hosting or personal use, they may be shocked if they don't notice this.

ClearlyClaire:
Fixes #10263

### DIFF
--- a/db/migrate/20190314181829_migrate_open_registrations_setting.rb
+++ b/db/migrate/20190314181829_migrate_open_registrations_setting.rb
@@ -1,0 +1,15 @@
+class MigrateOpenRegistrationsSetting < ActiveRecord::Migration[5.2]
+  def up
+    open_registrations = Setting.find_by(var: 'open_registrations')
+    return if open_registrations.nil? || open_registrations.value
+    setting = Setting.where(var: 'registrations_mode').first_or_initialize(var: 'registrations_mode')
+    setting.update(value: 'none')
+  end
+
+  def down
+    registrations_mode = Setting.find_by(var: 'registrations_mode')
+    return if registrations_mode.nil?
+    setting = Setting.where(var: 'open_registrations').first_or_initialize(var: 'open_registrations')
+    setting.update(value: registrations_mode.value == 'open')
+  end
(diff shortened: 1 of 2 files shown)


db/migrate/20190314181829_migrate_open_registrations_setting.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateOpenRegistrationsSetting < ActiveRecord::Migration[5.2]
+  def up
+    open_registrations = Setting.find_by(var: 'open_registrations')
+    return if open_registrations.nil? || open_registrations.value
+    setting = Setting.where(var: 'registrations_mode').first_or_initialize(var: 'registrations_mode')
+    setting.update(value: 'none')
```
Gargron: I suggest `approved`

ThibG: Why? It would be confusing for people that really wanted to close their registrations…

Gargron: I think a lot of closed registrations are due to the spammer problem, or uncontrolled growth. For example, that instance that only allowed people to sign up during the full moon... With this option, people can sign up any time, but only be allowed in during the full moon.

I fear that otherwise this feature will not be discovered by those who it would benefit :confused:

ThibG: I think some instances really want closed registrations.

I think the problem for instances that have closed with Instances that closed registrations due to spam problems or uncontrolled growth will just shift to approval requests. And it'll be worse on the admins if you disrupt their current ad-hoc approval flow without prior notice.

Gargron (approved):

PR #10273, opened 2019-03-14
Fix a comma

marek-lach:


Aditoo17:
I don't think that's wrong.

Aditoo17:
I'm pretty sure that you can't add a comma before "or". My Grammarly addon tells me the comma shouldn't be there.

Cassolotl:
@Aditoo17 

>I'm pretty sure that you can't add a comma before "or".

It's called an Oxford comma, and there is much debate on its appropriateness in various situations! I think it adds clarity here though.

Gargron:
> Irreversible filtering only works with home, or notifications context

Makes it less readable... It's like a random pause in the middle for no reason, throws me off. I'm not a linguist but the Oxford comma in my understanding is about the "and" particle, and it doesn't apply in exactly those cases where two things are directly connected. "I bought strawberry and banana ice cream". "I bought strawberry, and banana ice cream". The latter sounds weird, like strawberry is a stand-alone thing rather than a flavour of the ice cream.

Aditoo17:
I've also read that an Oxford comma can be used in a list of three or more terms. Meaning the commas you added aren't really grammatically correct.

PR #10280, opened 2019-03-15
Avoid race condition when streaming deleted statuses

ClearlyClaire:
This relies on new redis locks and values so I'm not sure how efficient that is.
But I can't see other ways to prevent those race conditions (basically, `DistributeWorker` doing its job while a status has been/is being deleted, resulting in the status being pushed to timelines *after* it was deleted).


ThibG:
Ok, this was a bit trickier than expected, as `FanOutOnWriteService` does itself push updates asynchronously to redis, so I locked the whole unpush until delete thing into a redis lock.

ykzts (approved):

Gargron (approved):

PR #10281, opened 2019-03-15
fix: update to emoji-mart 2.10.0

nolanlawson:
emoji-mart 2.10.0 contains a lot of fixes, especially around a11y. I've updated while trying to match our current design as much as possible:

Screenshot from 2019-03-15 08-43-36

emoji-mart 2.10.0 also adds a new `dist-modern` build, which gives a slightly smaller size but doesn't support IE11, only the latest versions of Chrome/Firefox/Safari/Edge. I can't see why we wouldn't want to use this, since we don't support IE11 either.

To remove prop-types in emoji-mart, you now have to use the `babel-plugin-transform-react-remove-prop-types`, so I've added config for that as well.

As you can see, the picker itself is down to ~39.1kB:

Screenshot from 2019-03-15 09-05-26

### DIFF
--- a/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
+++ b/app/javascript/mastodon/features/compose/components/emoji_picker_dropdown.js
@@ -50,6 +50,7 @@ class ModifierPickerMenu extends React.PureComponent {
     active: PropTypes.bool,
     onSelect: PropTypes.func.isRequired,
     onClose: PropTypes.func.isRequired,
+    modifier: PropTypes.number,
   };
 
   handleClick = e => {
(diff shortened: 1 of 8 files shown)


nolanlawson:
Hmmm looks like something changed in `babel_external.js`. Seems it was removed in https://github.com/tootsuite/mastodon/pull/10277, I'll update

nolanlawson:
OK, that should fix the merge conflict.

BTW it may make sense to only run babel on a subset of `node_modules`; else this may take a long time. We probably only need it for those libraries that ship `prop-types`.

ykzts (approved):

nolanlawson:
OK, I went through and did a big cleanup on the CSS to make it look almost exactly the same as before.

Before:

before

After:

after

I also fixed some of the accessibility in the dropdown picker, although there is more work to do to handle focus styles, keyboard input, etc. following the [menu pattern](https://www.w3.org/TR/wai-aria-practices/#menubutton).

nolanlawson:
fixed the fonts:

Screenshot from 2019-03-16 11-53-14

Gargron (approved):

PR #10284, opened 2019-03-15
Do not show timeline preview or directory widgets if they are disabled

ClearlyClaire:


Gargron:
I'm not convinced why

ThibG:
(Removing the closed messages commit from this, it's still WiP and another subject entirely)

Well, I find it only confusing. Also, the directory still shows users who opted in even if the feature has been disabled.

Gargron:
> Also, the directory still shows users who opted in even if the feature has been disabled.

Technically, it doesn't show the users, it shows the avatars--and I think even if the directory cannot be browsed, it's not a bad detail to convince the reader that there are people on the server (not a "ghost town").

I understand that being greyed out as an indicator for being disabled is perhaps not a fully intuitive thing, but maybe there are ways to adjust the "disabled" design to make it more obvious. (The "disabled" cursor on mouse-over, for example?)

ThibG:
I'm still not sure what is the benefit of showing a feature that's not enabled.
I guess I could understand if it was some kind of checklist of features the instance offers to a logged in user, but that's not the case.

Gargron (approved):

Issue #3815, opened 2017-06-17
Confirmation dialogue for unboosting toots


clarfonthey:
There is the shift + click confirmation dialogue for boosting toots, but not unboosting.

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

harrybailey:
#10287 looks to be merged. Can this be closed?

trwnh:
unboost confirmation is currently tied to boost confirmation, but it does exist. see #7860 for tracking issue on confirmation dialogs

PR #10287, opened 2019-03-16
Add confirm modal for unboosting toots

aurelien-reeves:
[#3815]

Display the boost modal also when unboosting toots.


aurelien-reeves:
Should we add tests for that?

I did not update the preferences. I think the setting_boost_modal locale string could be updated?

Maybe we should add a dedicated option to be able to activate the confirmation independently when boosting and unboosting toots?

Gargron (approved):

PR #10288, opened 2019-03-16
perf: remove zopfli to speed up webpack

nolanlawson:
This PR removes zopfli compression, speeding up webpack compilation in production by **~7.5x** (6m36s -> 0m53s).

## Analysis

I did some profiling of the command:

```bash
RAILS_ENV=production bundle exec rails assets:precompile
```

after some digging, I found I could debug webpack using: 

```bash
RAILS_ENV=production NODE_ENV=production node --inspect-brk /home/nolan/workspace/mastodon/node_modules/.bin/webpack --config /home/nolan/workspace/mastodon/config/webpack/production.js
```

So I took a trace. I think the screenshot speaks for itself:

Screenshot from 2019-03-15 19-00-50

## Benchmark

My benchmark is the following:

```bash
rm -fr tmp/
RAILS_ENV=production bundle exec rails assets:precompile
```

Then I use the `time` command to time it.

Without the fix, this takes **6m36.437s** in real time on my Dell XPS 13. With the fix, it takes **0m52.958s** (**~7.5x** improvement).

## Explanation

Zopfli outputs a very compact gzip format, but it's slow. Removing it speeds things up considerably.

Even though nginx's default gzip output may be less efficient than zopfli, I don't think it's reasonable to force every instance admin to run this command which takes a very long time. Maybe some admins aren't even using gzip &ndash; maybe they're using Brotli. They can also configure the gzip options in nginx, or run zopfli themselves.

As a future improvement, I don't think generating `.gz` files is even strictly necessary. Given [this config](https://github.com/tootsuite/mastodon/blob/782b622f5f295fc9d91bb2461dc049f55766af9c/dist/nginx.conf#L37-L44), nginx should serve gzipped content no matter what &ndash; it's just that we don't necessarily control the gzip output.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -7,7 +7,6 @@ const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const OfflinePlugin = require('offline-plugin');
 const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 const CompressionPlugin = require('compression-webpack-plugin');
-const zopfli = require('@gfx/zopfli');
 const { output } = require('./configuration');
 const sharedConfig = require('./shared');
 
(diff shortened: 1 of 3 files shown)


Gargron (approved):

angristan:
Thanks a lot !

PR #10295, opened 2019-03-16
Add missing locales (bn, ga, hi)

Gargron:


Aditoo17:
`ml` and `zh` is used only for language filtering (`ml` doesn't have a translation file and`zh-CN`, `zh-HK` and `zh-TW` are used for localization instead of `zh`) and `hi` is an empty file.

ykzts (approved):

PR #10297, opened 2019-03-16
Add Keybase integration

Gargron:
Allow users to setup a cryptographic connection between their Mastodon profile and their Keybase profile. Keybase is extracted into a polymorphous `ProofProvider` interface to potentially allow other providers to be supported and minimize "hardcoding" a specific provider.

image

image

image

Performance notes:

- One synchronous HTTP request is performed when proof is being saved to check if the signature hash is actually valid on Keybase
- One asynchronous HTTP request is queued right after the proof is created to check if the proof went live on Keybase (and can therefore be showcased)
- One synchronous HTTP request is performed to retrieve the avatar URL from Keybase for each proof, both on the proof index as well as on new proof page
- One synchronous HTTP request is performed to check if the proof is still live and valid whenever the proof index is loaded

___

Resolve #10013


app/controllers/well_known/keybase_proof_config_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module WellKnown
+  class KeybaseProofConfigController < ActionController::Base
+    def show
+      render json: {}, serializer: ProofProvider::Keybase::ConfigSerializer
```
krainboltgreene: This seems strange?

app/lib/proof_provider.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
krainboltgreene: Why `app/lib` and not `app/providers`?

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class ProofProvider::Keybase
+  BASE_URL = 'https://keybase.io'
+
+  class Error < StandardError; end
+
+  class ExpectedProofLiveError < Error; end
+
+  class UnexpectedResponseError < Error; end
```
krainboltgreene: One practice I've taken up is having a global exception directory (and error): `app/errors` and `app/exceptions`, so that errors can be shared and to deliniate between business problems and developer problems (example: `ArgumentError` vs `NotEnoughFundsException`

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def initialize(proof = nil)
+    @proof = proof
+  end
```
krainboltgreene: Looks like this could benefit from being an `ActiveModel::Model`?

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def serializer_class
+    ProofProvider::Keybase::Serializer
+  end
+
+  def worker_class
+    ProofProvider::Keybase::Worker
+  end
```
krainboltgreene: I feel like this should be derived rather than explicit? AKA `ProofProvider.const_get(self.class.name)::Worker`

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def validate!
+    unless @proof.token&.size == 66
+      @proof.errors.add(:base, I18n.t('identity_proofs.errors.keybase.invalid_token'))
```
krainboltgreene: It feels really strange to reach into `@proof` to modify it's errorset. Why not keep errors here?

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
...
+      return
+    end
+
+    return if @proof.provider_username.blank?
```
krainboltgreene: This means we'll return nil, instead of a Boolean.

app/lib/proof_provider/keybase/config_serializer.rb
```diff
@@ -0,0 +1,69 @@
...
+  def check_path
+    ['signatures']
+  end
+
+  def avatar_path
+    ['avatar']
+  end
+
+  def contact
+    Setting.keybase_contacts.split(',').map(&:strip)
```
krainboltgreene: You can shorten this to `.split(/\s+,\s+/)` btw.

spec/controllers/api/proofs_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    stub_request(:get, 'https://keybase.io/_/api/1.0/sig/proof_valid.json?domain=cb6e6126.ngrok.io&kb_username=hidden_alice&sig_hash=222222222222222222222222222222222222222222222222222222222222222222&username=alice').to_return(status: 200, body: '{"proof_valid":true,"proof_live":true}')
+    stub_request(:get, 'https://keybase.io/_/api/1.0/sig/proof_live.json?domain=cb6e6126.ngrok.io&kb_username=hidden_alice&sig_hash=222222222222222222222222222222222222222222222222222222222222222222&username=alice').to_return(status: 200, body: '{"proof_valid":true,"proof_live":true}')
+  end
+
+  describe 'GET #index' do
+    describe 'with a non-existent username' do
+      it '404s' do
+        get :index, params: { username: 'nonexistent', provider: 'keybase' }
+
+        expect(response.status).to eq 404
```
krainboltgreene: You can use the rspec helper here: `expect(response).to have_http_status(:not_founnd)`

app/controllers/well_known/keybase_proof_config_controller.rb
Gargron: The serializer does not use a specific object, but uses various globals. See also: REST::InstanceSerializer

app/lib/proof_provider.rb
Gargron: The `ActivityPub` and `OStatus` classes are likewise in `app/lib`, so I followed a precedent. Also, the namespacing is intentional to compartmentalize the proprietary vendor in such a way as to make removing it or swapping it out easier (everything is loaded more or less based on the `provider` value in `AccountIdentityProof` model)

app/lib/proof_provider/keybase.rb
Gargron: We do have an `app/lib/errors.rb` file with that sort of thing. My intention here is to contain Keybase-specific errors in the Keybase-specific area of the code--the Rails controllers/models should not receive these exceptions.

app/lib/proof_provider/keybase.rb
Gargron: I was considering using metaprogramming, and this might be a fair critique, but since it's just two methods that do this, I opted to just write them out. I am generally more comfortable with less metaprogramming involved because it simplifies finding errors and searching code for usage.

app/lib/proof_provider/keybase.rb
Gargron: `AccountIdentityProof` calls this method from `validate`. It is essentially an ActiveRecord validator. I was thinking whether it should be its own class or not, but it's not too many lines, so I just put it here.

One problem is that I have to use `validate` with a method/lambda that calls `provider_instance`, instead of `validates_with SomeValidatorClass` because that class would depend on the dynamic `provider` value. Of course I could do something like `validates_with ProofProvider::Validator`, which then in its own `validate_each` method would resolve the right class and call it, but why do all those extra steps when this method works?

app/lib/proof_provider/keybase.rb
Gargron: This method does not have a return value, it just fills `@proof.errors` during `@proof`'s validation phase.

app/lib/proof_provider/keybase/config_serializer.rb
```diff
@@ -0,0 +1,69 @@
...
+    { min: 1, max: 30, re: Account::USERNAME_RE.inspect }
+  end
+
+  def prefill_url
+    params = {
+      provider: 'keybase',
+      token: '%{sig_hash}',
+      provider_username: '%{kb_username}',
+      user_agent: '%{kb_ua}',
+    }
```
BenLubar: The validator wants there to be a username parameter (presumably, Mastodon should check to make sure this matches the logged-in user's name).

```json
{
  "status": {
    "code": 100,
    "desc": "missing or invalid inputs {\"prefill_url\":\"Error: Input https://mastodon.example.com/settings/identity_proofs/new?provider=keybase&provider_username=%{kb_username}&token=%{sig_hash}&user_agent=%{kb_ua} missing %{username}\"}",
    "name": "INPUT_ERROR"
  }
}
```

```suggestion
      username: '%{username}',
    }
```

Gargron: @xgess This wasn't in the original PR--has the validation on your side changed? We don't use the `username` param on that page on our side...

app/lib/proof_provider/keybase/config_serializer.rb
```diff
@@ -0,0 +1,70 @@
...
+  def check_path
+    ['signatures']
+  end
+
+  def avatar_path
+    ['avatar']
+  end
+
+  def contact
+    Setting.keybase_contacts.split(/\s+,\s+/)
```
Gargron: @xgess Are you sure you don't want `Setting.site_contact_email` here, which will likely be set in more cases than specifically `Setting.keybase_contacts`?

app/controllers/settings/identity_proofs_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+class Settings::IdentityProofsController < Settings::BaseController
+  layout 'admin'
+
+  before_action :authenticate_user!
+  before_action :check_required_params, only: :new
+
+  def index
+    @proofs = AccountIdentityProof.where(account: current_account).order(provider: :asc, provider_username: :asc)
+    @proofs.each(&:refresh!)
```
Gargron: @xgess Is this necessary? n synchronous HTTP requests on page load is not great. I would think the worker queued after the proof is saved would take care of checking if the proof is live

app/lib/proof_provider/keybase/config_serializer.rb
Gargron: I added the param anyway for the time being

config/settings.yml
```diff
@@ -54,6 +54,7 @@ defaults: &defaults
     - moderator
   disallowed_hashtags: # space separated string or list of hashtags without the hash
   bootstrap_timeline_accounts: ''
+  keybase_contacts: 'admin@localhostmastodon, yournamehere@keybase'
```
ThibG: Should probably be an empty string. This file includes valid default settings, not examples.

app/controllers/settings/identity_proofs_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def new
+    @proof = current_account.identity_proofs.new(
+      token: params[:token],
+      provider: params[:provider],
+      provider_username: params[:provider_username]
```
ThibG: `resource_params`?

app/lib/proof_provider/keybase.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    if verifier.valid?
+      @proof.verified = true
+      @proof.live     = false
```
ThibG: Should ActiveRecord validation really have side-effects?

app/views/accounts/_bio.html.haml
```diff
@@ -1,14 +1,25 @@
+- proofs = account.identity_proofs.active
+- fields = account.fields
+
 .public-account-bio
-  - unless account.fields.empty?
+  - if !fields.empty? || !proofs.empty?
```
ThibG: `unless fields.empty? && proofs.empty?`?

app/controllers/settings/identity_proofs_controller.rb
ThibG: The `refresh!` do not seem to be synchronous (it spawns a worker?), but I question the need to trigger a refresh each time that page is visited.

ClearlyClaire (commented):
This review doesn't really cover the correctness of the implementation wrt. keybase's protocol itself, as I have not read it, but only code quality and some concerns.

It seems fine overall, but I'm a bit concerned with the various HTTP calls, some of which synchronous. This shouldn't be a major issue, but still.

Also, will this work for any instance, or do instance admins still need to get approved by keybase?

Wouldn't it make sense to have a setting to enable or disable this feature? (Especially if the admin needs to take action wrt. keybase for it to be usable)

app/controllers/settings/identity_proofs_controller.rb
Gargron: Ah no, the difference to resource_params is that these are like `?token=???` while resource_params after form submission are nested

app/lib/proof_provider/keybase.rb
Gargron: I suppose you could argue that `verified` could be set before_save which would not happen if validation doesn't pass, but I am not too unhappy with how it is

app/controllers/settings/identity_proofs_controller.rb
Gargron: It does `worker_class.new.perform(@proof)` which *is* synchronous

Gargron:
>Also, will this work for any instance, or do instance admins still need to get approved by keybase?
>
>Wouldn't it make sense to have a setting to enable or disable this feature? (Especially if the admin needs to take action wrt. keybase for it to be usable)

I was assured no proactive action would be necessary on the part of the instance admin

app/lib/proof_provider/keybase/config_serializer.rb
xgess: i trust your judgment on this

app/controllers/settings/identity_proofs_controller.rb
xgess: the main reason i had to do this was that there's nothing currently built on the mastodon side to recognize remotely revoked proofs. i thought it might be weird if a user revokes a proof in keybase, then days later still sees it as live in mastodon until the first refresh. we had it as a note to talk about / build something that might let keybase inform mastodon a proof is revoked, or build a rake task for mastodon to check and invalidate revoked proofs. 

i'm flexible on this though. if you want to change it to be async, i think that's reasonable.

app/lib/proof_provider.rb
krainboltgreene: Yeah, seems like these should be a ProofProvider type of class, IMO, but I can definitely understand the current implementation.

app/lib/proof_provider/keybase.rb
krainboltgreene: Understandable.

app/lib/proof_provider/keybase.rb
krainboltgreene: I guess. This violates the Law of Demeter, though, and IMO will create annoyances down the line.

app/lib/proof_provider/keybase.rb
krainboltgreene: Ah, I see.

app/lib/proof_provider/keybase/config_serializer.rb
krainboltgreene: Also, is it more likely that you want `/\s*,\s*/` to allow for no or any spaces?

krainboltgreene (approved):

witcheslive:
I have updated witches.live to the master branch which has support for this, do you need any help testing it, and if so how could I go about doing that?

cnf:
Is there any documentation on how to use this as a user? I can't seem to find it.

Gargron:
@cnf It is not live on Keybase's side yet. Other than that, there is nothing to do *on Mastodon*, you initiate the process from Keybase.

cnf:
Right, thanks!

PR #10312, opened 2019-03-18
Fix admin settings description

tribela:
- ~Remove `closed_registrations_message` which is no longer used~
- Change descriptions on `site_description`, `site_short_description`


ClearlyClaire (commented):
I stronlgy disagree with removing `closed_registrations_message`, I think it should be reintroduced back.

Gargron:
I have re-added closed registrations message. Could you re-submit only the description changes?

Gargron (approved):

PR #10322, opened 2019-03-19
Add support for custom emojis in poll options

ClearlyClaire:
I decided to list the used custom emoji in the REST serialization, but it's definitely not the most efficient way to do it. It was the easiest way with the current architecture though.
I can look into changing that if needed.

### DIFF
--- a/app/javascript/mastodon/components/poll.js
+++ b/app/javascript/mastodon/components/poll.js
@@ -44,6 +44,11 @@ const timeRemainingString = (intl, date, now) => {
   return relativeTime;
 };
 
+const makeEmojiMap = record => record.get('emojis').reduce((obj, emoji) => {
+  obj[`:${emoji.get('shortcode')}:`] = emoji.toJS();
+  return obj;
+}, {});
+
 export default @injectIntl
 class Poll extends ImmutablePureComponent {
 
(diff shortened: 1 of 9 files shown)


app/javascript/mastodon/components/poll.js
```diff
@@ -99,6 +104,12 @@ class Poll extends ImmutablePureComponent {
     const active             = !!this.state.selected[`${optionIndex}`];
     const showResults        = poll.get('voted') || poll.get('expired');
 
+    let titleEmojified = option.get('title_emojified');
+    if (!titleEmojified) {
+      const emojiMap = makeEmojiMap(poll);
+      titleEmojified = emojify(escapeTextContentForBrowser(option.get('title')), emojiMap);
```
Gargron: This might be slow now. Is it possible to move this to the normalizer?

ThibG: This is for the public pages, where we don't actually use the redux store and normalizer. Could still move it in the public-page specific code, before going through `fromJS` I guess.

Gargron: Oh right. Then it's fine. I completely forgot that's how it was structured.

Gargron (approved):

PR #10326, opened 2019-03-20
Do not try fetching keys of unknown accounts on a Delete from them

ClearlyClaire:


app/controllers/activitypub/inboxes_controller.rb
```diff
@@ -17,12 +20,21 @@ def create
 
   private
 
+  def unknown_deleted_account?
+    json = Oj.load(body, mode: :strict)
+    return false unless json['type'] == 'Delete' && json['actor'].present?
+    return false unless json['actor'] == value_or_id(json['object'])
+    !Account.where(uri: json['actor']).exists?
```
Gargron: I feel like these three boolean checks might be more readable in a single statement instead of separate returns

Gargron (approved):

PR #10335, opened 2019-03-21
Export and import `hide_notifications` alongside user mutes

ClearlyClaire:


Gargron:
I am concerned that this is a pandora's box because we also have follows with/without reblogs and whatnot. So far the CSV files were just lists of strings, very simple and understandable, now we are putting unexplained booleans in them. Maybe this should require the introduction of a header row.

ThibG:
Ah, yes, we could add a header, that would also help with potential future changes.
We need to be careful to chose header names that might not be interpreted as usernames by older versions though

ThibG:
Well, I'm afraid I have no idea why the test fails on 2.4 and 2.5 but not 2.6.

ThibG:
It appears it's a bug in the CSV module in older versions of Ruby when involving both headers and empty lines :weary:
I modified the test suite to not have such empty lines…

ThibG:
Updated to use more natural-looking headers such as “Account address” instead of “#account”

Gargron (approved):

PR #10337, opened 2019-03-21
Redesign profile column in web UI to match design on public pages

Gargron:
image

### Design notes

- Square avatars (common complaint that they had a shape inconsistent with the rest of the interface)
- Although I do miss the bouncy corners animation, that was sweet
- Header image no longer obscured by everything else
- Big obvious action button, instead of grey icon button that was blending with the backround
- Also further away from the back button
- Username now always showing full address even if local


trwnh:
the only thing i can suggest is to consider making the avatar not overlap the header (for consistency with the profile cards shown on following/followers pages), but i'm fine with it either way.

ThibG:
Overall this seems good, but I'm not super satisfied with the Toots / Follows / Followers stuff, which doesn't look super nice nor feels like tabs/buttons?

Gargron:
> the only thing i can suggest is to consider making the avatar not overlap the header (for consistency with the profile cards shown on following/followers pages), but i'm fine with it either way.

That's what I started with but it leads to either massive wasted space in the middle, or a pretty small avatar to be the same height as the buttons. I imagine people want to see what's on the avatar when they open a profile.

Gargron:
> Overall this seems good, but I'm not super satisfied with the Toots / Follows / Followers stuff, which doesn't look super nice nor feels like tabs/buttons?

It is what we've had on the public pages for a while.

Maybe making the entire text bolder would improve them?

Aditoo17:
You might also want to make the avatar on public pages square, to be consistent.

ThibG:
I mean, on the public pages, it looks like that, and I find it's way better?
Screenshot_2019-03-22 Eugen ( Gargron mastodon social)

Gargron:
@ThibG Resize the window to mobile width

ariasuni:
Shouldn’t be the three dot button be to the right of the «Edit profile» one, for consistency with toots and compose column?

ykzts (approved):

PR #10340, opened 2019-03-22
Bump version to 2.8.0rc1

Gargron:
TODO:

- [x] Re-introduce custom closed message to new landing page design
- [x] Hide identity proofs menu item unless the user has identity proofs
- [x] Fix emoji-mart search crash
- [x] Fix slowness in list of domain blocks
- [x] Add federation for identity proofs


ThibG:
Also: fixing the slowness in `/admin/instances`

PR #10351, opened 2019-03-23
Weblate translations (2019-03-23)

Gargron:


mayaeh:
@Gargron defaultMessages.json is trying to return to the previous translation state.

https://github.com/tootsuite/mastodon/pull/10351/files#diff-be68e2cbf3d70594c963caa9d60ac578

PR #10354, opened 2019-03-23
Fix HTML validation

ClearlyClaire:
~~A possible improvement would be to use `fragment.errors` to tell *why* validation failed.~~

### DIFF
--- a/app/validators/html_validator.rb
+++ b/app/validators/html_validator.rb
@@ -9,6 +9,7 @@ class HtmlValidator < ActiveModel::EachValidator
   private
 
   def valid_html?(str)
-    Nokogiri::HTML.fragment(str).to_s == str
+    fragment = Nokogiri::HTML.fragment(str)
+    fragment.errors.empty?
   end
 end
--- a/app/validators/html_validator.rb
+++ b/app/validators/html_validator.rb
@@ -3,13 +3,16 @@
 class HtmlValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.blank?
-    record.errors.add(attribute, I18n.t('html_validator.invalid_markup')) unless valid_html?(value)
+    errors = html_errors(value)
+    unless errors.empty?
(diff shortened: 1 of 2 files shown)


app/validators/html_validator.rb
```diff
@@ -3,12 +3,16 @@
 class HtmlValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.blank?
-    record.errors.add(attribute, I18n.t('html_validator.invalid_markup')) unless valid_html?(value)
+    errors = html_errors(value)
+    unless errors.empty?
+      record.errors.add(attribute, I18n.t('html_validator.invalid_markup', error: errors.first.to_s))
```
Gargron: The error message from Nokogiri won't be localized, so I think it might be worthless to show it to the end-user.

ThibG:
I think it's still worth showing, even if it's not localized. Most of the time, it will mostly consist of a tag name and a position, which people might understand even if they don't understand English.
I think this is better than just telling them it's invalid without telling people why, which would be a really frustrating experience.



Le 25 mars 2019 05:15:13 GMT+01:00, Eugen Rochko <notifications@github.com> a écrit :
>
>
>
>> @@ -3,12 +3,16 @@
> class HtmlValidator < ActiveModel::EachValidator
>   def validate_each(record, attribute, value)
>     return if value.blank?
>-    record.errors.add(attribute,
>I18n.t('html_validator.invalid_markup')) unless valid_html?(value)
>+    errors = html_errors(value)
>+    unless errors.empty?
>+      record.errors.add(attribute,
>I18n.t('html_validator.invalid_markup', error: errors.first.to_s))
>
>The error message from Nokogiri won't be localized, so I think it might
>be worthless to show it to the end-user.
>
>-- 
>https://github.com/tootsuite/mastodon/pull/10354#pullrequestreview-218139382

-- 

Gargron (approved):

Issue #10226, opened 2019-03-09
Impossible to report a user after muting or blocking them


KellerFuchs:
### Expected behaviour

- Block or mute user.
- Be able to report the behavior that prompted the block.

### Actual behaviour

- Block or mute user.
- UI instantly closes their profile (if it was opened in the third column), hides all their toots, and removes all mentions of their username.
- Be unable to report bad behaviour.

### Steps to reproduce the problem

See above.

### Specifications

Encountered on vulpine.club.

ThibG:
I experience something slightly different from described: blocking an user does not close their profile.
It indeed removes statuses from every timeline, including account timelines, but subsequently loading the timeline again (which does happen when firing up the report modal) allows you to report statuses.

So I guess the only thing we should do differently is to not remove statuses from that account's timeline? We could also offer to open the report dialog after blocking the account.

PR #10358, opened 2019-03-24
Display how many people blocked or muted users in the admin settings

ClearlyClaire:
I think this information could be useful. I am afraid this would make the “dashboard” a bit too noisy though.

Screenshot_2019-03-24 thib - Dev instance

Additionally, it could be interesting to be able to sort users by the number of blocks. This could be useful to let admins investigate accounts who are blocked a lot, even in the absence of actual reports.

### DIFF
--- a/app/views/admin/accounts/show.html.haml
+++ b/app/views/admin/accounts/show.html.haml
@@ -26,6 +26,18 @@
       .dashboard__counters__label= t '.targeted_reports'
   %div
     %div
+      .dashboard__counters__num= number_with_delimiter @account.muted_by.count
+      .dashboard__counters__label= t '.muted_by'
+  %div
+    %div
+      .dashboard__counters__num= number_with_delimiter @account.blocked_by.count
+      .dashboard__counters__label= t '.blocked_by'
+  %div
+    %div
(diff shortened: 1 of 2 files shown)


ThibG:
This is a feature I've seen requested, as there are a bunch of people who are blocked but not reported despite being repeat offenders of more minor offenses, such a feature would allow to spot that.

On the other hand, blocks may arise from personal issues without actual abuse from the blocked user, and blocks themselves do not tell anything about *why* the user might be blocked. So while this may help spotting troublesome users for further investigation, this feature might also misguide admins and lead to further unjustified isolation of some people if admins trust the number of blocks rather than investigate properly. In the case of an open-registration server, it would also allow harassers to massively create accounts and block their victim to turn them into a suspect.

Maybe a better option would be to encourage reporting instead: fixing #10226, adding a suggestion to report the blocked person in/after the block confirmation modal, implement proper feedback about reports…

Gargron:
I agree with the assessment--blocking is too personal, muting even more so, to make generalizations.

ThibG:
Yeah, gonna close this one

PR #10365, opened 2019-03-25
Fix an issue where the error message contained "Owned poll"

mayaeh:
before:
スクリーンショット 2019-03-22 19 11 43

after:
スクリーンショット 2019-03-25 8 55 38

### DIFF
--- a/config/locales/activerecord.en.yml
+++ b/config/locales/activerecord.en.yml
@@ -2,6 +2,9 @@
 en:
   activerecord:
     attributes:
+      owned_poll:
+        expires_at: The deadline
+        options: Choices
       status:
         owned_poll: Poll
     errors:
--- a/config/locales/activerecord.ja.yml
+++ b/config/locales/activerecord.ja.yml
@@ -2,6 +2,9 @@
 ja:
   activerecord:
     attributes:
+      owned_poll:
+        expires_at: 期限
+        options: 選択肢
       user:
         email: メールアドレス
     errors:


ykzts (approved):

Gargron:
Might it be worth it to switch places of the owned_poll and poll associations, if owned_poll is what comes up in user-facing validation messages? I am not sure that ActiveRecord stuff properly falls back on values defined in the English files, so these fixes might be necessary for all locales?

mayaeh:
I do not know if it is best to define in this place.
I tried the following referring to #10333 , but did not improve it.
スクリーンショット 2019-03-25 13 31 32
スクリーンショット 2019-03-25 13 29 19


Therefore I made this PR.

ActiveRecord fallback looks a bit unstable. 🤔 

ja:
スクリーンショット 2019-03-25 13 38 28

zh-CN:
スクリーンショット 2019-03-25 13 51 20

zh-HK:
スクリーンショット 2019-03-25 13 49 01


ko (activerecord.ko.yml seems not to exist) :
スクリーンショット 2019-03-25 13 45 46

Should I add this to "activerecord. *. yml" for all locales?

mayaeh:
I changed it to the following and tried, but "Owned poll" was output.

Gargron:
I mean swapping the `has_one`/`belongs_to` associations in `Status` model so that `owned_poll` would become just `poll`, then overriding locales would not be necessary, right?

hinaloe:
@Gargron Are you saying use `has_one :poll` instead of `has_one :owned_pill`? This may be one way to improve, but override locale is still necessary.

Gargron:
Sorry, I didn't mean to close this--I forgot to remove "resolve xxx" from the commit message. Can you re-submit the Japanese strings?

mayaeh:
@Gargron Don't worry! No problem. I will add these translations via the Weblate later.

Gargron:
Please submit them with a pull request! I think we should move away from Weblate due to lack of quality control over submissions there.

mayaeh:
Oh ... I read #10385 .
All right.

PR #10366, opened 2019-03-25
 Change icons of features on admin dashboard to remove bias

Gargron:
Fix sample accounts changing too frequently due to wrong query

### DIFF
--- a/app/views/admin/dashboard/index.html.haml
+++ b/app/views/admin/dashboard/index.html.haml
@@ -40,35 +40,17 @@
       %h4= t 'admin.dashboard.features'
       %ul
         %li
-          = link_to t('admin.dashboard.feature_registrations'), edit_admin_settings_path
-          - if @registrations_enabled
-            %span.pull-right.positive-hint= fa_icon 'check fw'
-          - else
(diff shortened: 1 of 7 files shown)


app/views/admin/dashboard/index.html.haml
```diff
@@ -42,33 +42,39 @@
         %li
           = link_to t('admin.dashboard.feature_registrations'), edit_admin_settings_path
           - if @registrations_enabled
-            %span.pull-right.positive-hint= fa_icon 'check fw'
+            %span.pull-right.positive-hint= fa_icon 'power-off fw'
```
ThibG: power-on, I guess?

Gargron: There is no power-on icon. They are the same thing

ThibG: Ah, I see. So the only difference is going to be the color? I'm afraid that it could be an accessibility issue

Gargron: Now they also say "yes"/"no"

ykzts (approved):

PR #10367, opened 2019-03-25
Add `tootctl self-destruct`

### LINKED ISSUE #10305: Content deletion when an instance shuts down
I was wondering about what happens in the case of an expected or unexpected shutdown of an instance.

- Will the content (including images) stay available on other instances?
- Does a user have the chance to delete his distributed content if the instance is already offline?

Edit: Accidentally hit send too early

Gargron:
Fix #10305

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,7 +80,7 @@ Rails/HttpStatus:
 Rails/Exit:
   Exclude:
     - 'lib/mastodon/*'
-    - 'lib/cli'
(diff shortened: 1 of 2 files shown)


ClearlyClaire (commented):
Wouldn't it make sense to also mark the accounts as suspended? For instance if the admin forgot to shut down the web services prior to running this.

Gargron:
@ThibG Done

ClearlyClaire (approved):

PR #10375, opened 2019-03-25
assorted proof-related followups

xgess:
* check the environment for some keybase-specific configs before using the current defaults. this is just nice for development.
* update the identity proof model to trigger the worker on updates as well as creates to handle the case where a user revokes and then re-proves the same identity
* in the case where a user is trying to prove a different mastodon account than the one they're logged into, catch that a little earlier and show a helpful error message. 
* 2 little tweaks to the hosted keybase configs that make them easier for keybase to consume
* fix a bug where the very long name of a mastodon instance or a proof provider can overflow in the new proof UI and look bad
* add active proofs to the account panel on the dashboard just like on the profile page
* add an option to post a new status during the identity proof creation flow.

here's what that last change looks like in the UI
Screen Shot 2019-03-25 at 4 42 52 PM

and then the post
Screen Shot 2019-03-25 at 4 49 01 PM

### DIFF
--- a/app/controllers/settings/identity_proofs_controller.rb
+++ b/app/controllers/settings/identity_proofs_controller.rb
@@ -18,7 +18,12 @@ class Settings::IdentityProofsController < Settings::BaseController
       provider_username: params[:provider_username]
     )
 
-    render layout: 'auth'
+    if current_account.username == params[:username]
+      render layout: 'auth'
+    else
+      flash[:alert] = I18n.t('identity_proofs.errors.wrong_user', proving: params[:username], current: current_account.username)
+      redirect_to settings_identity_proofs_path
+    end
   end
 
   def create
@@ -26,6 +31,7 @@ class Settings::IdentityProofsController < Settings::BaseController
     @proof.token = resource_params[:token]
 
     if @proof.save
+      PostStatusService.new.call(current_user.account, text: post_params[:status_text]) if publish_proof?
       redirect_to @proof.on_success_path(params[:user_agent])
     else
       flash[:alert] = I18n.t('identity_proofs.errors.failed', provider: @proof.provider.capitalize)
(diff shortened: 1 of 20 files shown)


app/serializers/rest/account_serializer.rb
```diff
@@ -20,6 +20,26 @@ def value
...
+    def profile_url
+      object.badge.profile_url
+    end
+
+    def provider
+      object.provider.capitalize
+    end
+  end
+
+  has_many :identity_proofs do
```
Gargron: We include a full account entity along with every status entity, so I believe this is not an option, performance-wise. It's an n+1 query and won't be used in most requests. I think it should be fetchable through a separate API endpoint.

app/views/settings/identity_proofs/new.html.haml
```diff
@@ -27,5 +27,10 @@
 
           %p= t('identity_proofs.i_am_html', username: content_tag(:strong, @proof.provider_username), service: @proof.provider.capitalize)
 
+    .connection-prompt__post
+      = f.input :post_status, label: "And toot this:", as: :boolean, wrapper: :with_label, :input_html => { checked: true, value: 'true' }
```
Gargron: Not localized, and it looks like the checkbox does not align well with the label for some reason (in the screenshot)...

app/serializers/rest/account_serializer.rb
xgess: yup. makes sense. i took a stab at it. lemme know what you think.

app/controllers/settings/identity_proofs_controller.rb
```diff
@@ -36,10 +42,22 @@ def create
...
   def check_required_params
-    redirect_to settings_identity_proofs_path unless [:provider, :provider_username, :token].all? { |k| params[k].present? }
+    redirect_to settings_identity_proofs_path unless [:provider, :provider_username, :username, :token].all? { |k| params[k].present? }
   end
 
   def resource_params
     params.require(:account_identity_proof).permit(:provider, :provider_username, :token)
   end
+
+  def publicize_new_proof
```
Gargron: This method should have a `?` at the end for readability. Could also just name it `publish_proof?` for simplicity

app/javascript/mastodon/features/account_timeline/containers/header_container.js
```diff
@@ -33,6 +34,7 @@ const makeMapStateToProps = () => {
 
   const mapStateToProps = (state, { accountId }) => ({
     account: getAccount(state, accountId),
+    identity_proofs: state.getIn(['identity_proofs', accountId.toString()], ImmutableList()),
```
Gargron: `accountId` is already a string

Gargron:
Save for the few comments I left, looks fine to me. You should be aware the web UI profile component changed in master (#10337) so you have merge conflicts right now.

xgess:
thanks @Gargron. conflict fixed. and i like those UI changes on master.

Gargron (approved):

PR #10378, opened 2019-03-26
Remove unused ActivityPub `@context` values depending on response

### LINKED ISSUE #8078: Use the @context that is appropriate for the response
While looking at the json-ld responses of various requests to `mastodon.technology`, which I recently joined, and noticed that each and every response includes the full `@context` definition with all Mastodon extensions, regardless of what is in the response body.

```json
{
  "@context": [
    {
      "manuallyApprovesFollowers": "as:manuallyApprovesFollowers",
      "sensitive": "as:sensitive",
      "movedTo": "as:movedTo",
      "Hashtag": "as:Hashtag",
      "ostatus": "http://ostatus.org#",
      "atomUri": "ostatus:atomUri",
      "inReplyToAtomUri": "ostatus:inReplyToAtomUri",
      "conversation": "ostatus:conversation",
      "toot": "http://joinmastodon.org/ns#",
      "Emoji": "toot:Emoji",
      "focalPoint": {
        "@container": "@list",
        "@id": "toot:focalPoint"
      },
      "featured": "toot:featured",
      "schema": "http://schema.org#",
      "PropertyValue": "schema:PropertyValue",
      "value": "schema:value"
    }
  ],
  "id": "https://mastodon.technology/users/rapnie/followers",
  "type": "OrderedCollection",
  "totalItems": 7,
  "first": "https://mastodon.technology/users/rapnie/followers?page=1"
}
```

The above message could just as well be written as:

```json
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "id": "https://mastodon.technology/users/rapnie/followers",
  "type": "OrderedCollection",
  "totalItems": 7,
  "first": "https://mastodon.technology/users/rapnie/followers?page=1"
}
```

While the `@context` + extensions does not contain that many bytes, it could add up on busy servers. So this feature request is for having contexts that are applicable to the response that is returned, and do not contain extensions that are never returned.

**Edi**t: I have since heard from @nightpool on Mastodon that implementing this would be hard, because:

>there's no real principle way for us to signal to the serializer which context we need, without writing a lot of fragile special case logic that would needlessly complicate the code-base. 
>
>The best I can see us doing would be collecting it all into a single context document, but even that comes with a lot of bandwidth/centralization tradeoffs that I don't really see as feasible for us.

Could be indeed hard to implement in Python - I am not familiar with it - but I don't see why it would lead to 'lot of fragile special case logic'. As I see it for every type of response the `@context` could be more or less fixed (but different wrt to another AP entity type).

So I decided to file this issue anyway.. feel free to close if this is too hard to implement in any case.

* * * *

Gargron:
Fix #8078

Each serializer states which contexts it uses, and which context extensions. The only useful named context right now is `security`. As for context extensions, they are categorized by features/useful property groups.

The adapter then assembles a `@context` dynamically from this information.

With less stuff in the `@context`, it saves a little bit of bandwidth, a little bit of storage in Sidekiq queues when the JSON is used as job arguments, and the payloads become more readable (glance value).

### DIFF
--- a/app/lib/activitypub/adapter.rb
+++ b/app/lib/activitypub/adapter.rb
@@ -1,30 +1,23 @@
 # frozen_string_literal: true
 
 class ActivityPub::Adapter < ActiveModelSerializers::Adapter::Base
-  CONTEXT = {
-    '@context': [
-      'https://www.w3.org/ns/activitystreams',
-      'https://w3id.org/security/v1',
+  NAMED_CONTEXT_MAP = {
+    activitystreams: 'https://www.w3.org/ns/activitystreams',
+    security: 'https://w3id.org/security/v1',
+  }.freeze
 
-      {
-        'manuallyApprovesFollowers' => 'as:manuallyApprovesFollowers',
-        'sensitive'                 => 'as:sensitive',
-        'movedTo'                   => { '@id' => 'as:movedTo', '@type' => '@id' },
-        'alsoKnownAs'               => { '@id' => 'as:alsoKnownAs', '@type' => '@id' },
-        'Hashtag'                   => 'as:Hashtag',
-        'ostatus'                   => 'http://ostatus.org#',
-        'atomUri'                   => 'ostatus:atomUri',
-        'inReplyToAtomUri'          => 'ostatus:inReplyToAtomUri',
-        'conversation'              => 'ostatus:conversation',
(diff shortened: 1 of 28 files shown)


nightpool (approved):
I still do not understand how associations are supposed to propagate but

PR #10390, opened 2019-03-26
Move the account dropdown menubar to the right, make sure it doesn't move

ClearlyClaire:
This moves the dropdown menubar on profile pages to the right, as suggested in https://github.com/tootsuite/mastodon/pull/10337#issuecomment-475949250

This improves consistency with other places such menus appear, and additionally ensures the button doesn't move while the relationships are loading.

## Before

Screenshot_2019-03-26 Dev instance(1)

before
(This one is an animated PNG)

## After

Screenshot_2019-03-26 Dev instance

after
(This one is an animated PNG)

### DIFF
--- a/app/javascript/mastodon/features/account/components/header.js
+++ b/app/javascript/mastodon/features/account/components/header.js
@@ -219,9 +219,9 @@ class Header extends ImmutablePureComponent {
             <div className='spacer' />
 
             <div className='account__header__tabs__buttons'>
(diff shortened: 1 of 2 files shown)


ThibG:
Fixed the margins

Screenshot_2019-03-26 Dev instance(2)

Gargron (approved):

Gargron (approved):

PR #10391, opened 2019-03-26
Add support for linking to OpenNIC (and peered networks) TLD domains

hook321:
This enables links to domains that use OpenNIC TLDs (and TLDs of networks that OpenNIC peers with) in toots.


Gargron (changes requested):
I believe I have said this over the Discord where you posted about this PR, but this file change does not lead to the result you intend. The JavaScript does not control what links show up. The linkifying happens in the backend, through the Formatter classes, and deeper still, through the twitter-text gem.

PR #10401, opened 2019-03-27
Rename :poll to :preloadable_poll and :owned_poll to :poll on Status

Gargron:
Also, fix some n+1 queries

### DIFF
--- a/app/helpers/stream_entries_helper.rb
+++ b/app/helpers/stream_entries_helper.rb
@@ -105,8 +105,8 @@ module StreamEntriesHelper
   end
 
   def poll_summary(status)
-    return unless status.poll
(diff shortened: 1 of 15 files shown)


mayaeh:
I tried this on my test server.

It is changed from "Owned poll", but "Poll options" and "Poll expired at" are displayed.

en:

ja:


I made #10365 because I wanted to be able to translate such messages on the Weblate.

Gargron:
@mayaeh When I try to post a poll with just one option, I get "Validation failed: Options can't be blank" (not "Poll options"). I am curious, why is the English variant different for you?

mayaeh:
Thank you very much. I agree.
I'm sorry to confuse you for my lack of explanation.

ykzts (approved):

PR #10412, opened 2019-03-28
Remove border around avatar on profile column

ClearlyClaire:
Borders do not play nice with transparent avatars.

## Before

Screenshot_2019-03-28 Dev instance(9)

## After

Screenshot_2019-03-28 Dev instance(10)

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -5308,7 +5308,7 @@ noscript {
       margin-left: -2px;
 
       .account__avatar {
-        border: 2px solid lighten($ui-base-color, 4%);
+        padding: 2px;
       }
     }
   }


Gargron:
I'm for background instead

ThibG:
I could not find a background color that actually looked good

Gargron:
The last option looks good, I think. I think the problem is that this particular avatar has an irregular shape and some of its own padding top/bottom. For others, I think it will look fine.

ThibG:
I mean, we could go for that last option… still, I'm not sure about the value of a border to begin with?

trwnh:
I think having a transparent avatar would be made much less attractive if a background was added. Not sure what's wrong with having no background and no border.

Gargron:
Did the background color fix (for a couple more elements too) in #10422

PR #10414, opened 2019-03-29
Add ActivityPub representation for identity proofs

Gargron:
I define one new property: `toot:IdentityProof`, which is an object type. The proofs go into the account's `attachment` property. The `name` represents the identity's (user)name, the `signatureAlgorithm` (from the security context) identifies the identity proof platform (e.g. `keybase`), and `signatureValue` (also from the security context) carries the cryprographic token used for verification. You could say that the signature algorithm "keybase" tells you how to verify the signature token.


ThibG:
Seems a bit weird to me to reuse `signatureAlgorithm` and `signatureValue` for that but otherwise this seems like a sensible approach.

nightpool:
are the other signature algorithms we use URI values? should this one be as well?

Gargron:
They are not URIs, no. Stuff like Rsa2017 or whatever.

PR #10420, opened 2019-03-29
Add indication that you have been blocked in web UI

Gargron:


ykzts (approved):

inmysocks:
This is just a gigantic gift to harassers at the expense of people being harassed. Who else would this help?

This violates some of the very basic principles for making a system that is in any way robust against harassment. 'Don't feed the trolls' as they say, here by blocking someone you are giving them a clear indication that they have had an effect and you are just encouraging them to find another channel for harassment. You are never supposed to give people who are harassing you any indication that you hear or don't hear them.

nightpool:
blocks have always been visible to the user blocked, since they stop a
blocked user from being able to see the other user's posts. we also have
the mute functionality, which is identical to blocks except the user you're
muting can't tell the difference.

inmysocks:
That rational is 'it isn't perfect so it is ok to make it worse'. People have been talking about how existing blocks and anti-harassment functionality could be improved. This is going in the opposite direction.

As I said, who does this help other than harassers? It just makes it easier for them to see that they have been blocked specifically instead of the account being made private or deleted.

bgcarlisle:
I am struggling to see who this helps, aside from abusers.

narFnarF:
I am a bit concerned too. I'd like to know what is the reasoning for this change. I see from nightpool's answer that there are some technical concerns too. 
I'm wondering if there would be other options.

nivex:
> blocks have always been visible to the user blocked, since they stop a blocked user from being able to see the other user's posts.

Fine, but did it explicitly *say* that the user had been blocked? I'll bet not since this PR exists. A cryptic error about not being able to see that profile or post gives plausible deniability as opposed to explicitly stating that a block exists.

If I follow a direct link to a follower-only post, I don't get an indication stating this, I only see an error message stating the post couldn't be found. Blocks should be the same way.

Concur with previous comments about abusers taking advantage of this. Please unmerge.

joshgiesbrecht:
Agreeing with the above concerns

kindfulkirby:
Wow, this is a *terrible* idea.
I thought one of the points of mastodon was to leave twitter's anti-features (like this one) behind!

ealgase:
I really like this change.  If someone's being harassed and they block the harasser, it acts as a wake up call to the harasser to stop.  If they make a new account, they'll get reblocked and banned from that instance pretty quickly.

Gargron:
The discussion seems to be happening here: #10433. Let's not split it up.

SelfsameSynonym:
I’m confused - is the goal with this to make mastodon worse than Facebook for letting users protect themselves?

Lana-chan:
this PR introduces a regression bug (#10433)

PR #10431, opened 2019-03-31
Improvements to image upload validation and creation

slice:
This PR adds behavior that checks if uploaded images aren't nil before creating SiteUpload models for them.

For example, on a fresh instance install, adding a hero image in `/admin/settings/edit` actually creates a nil server thumbnail because the key was provided in the form submission (with a nil value). This ends up causing the unauthenticated landing page to display `missing.png` instead of falling back to the hero image as expected, because it found the nil server thumbnail instead. This patch fixes that by checking if an image was actually uploaded, and refusing to create a `SiteUpload` model if so.


Gargron (approved):

Gargron:
SiteUpload should probably also validate presence of file... This shouldn't have happened...

slice:
I think this commit does the trick. (I don't have a lot of Rails experience.)

app/models/site_upload.rb
```diff
@@ -17,7 +17,7 @@
 class SiteUpload < ApplicationRecord
   has_attached_file :file
 
-  validates_attachment_content_type :file, content_type: /\Aimage\/.*\z/
+  validates_attachment_content_type :file, content_type: /\Aimage\/.*\z/, presence: true
```
Gargron: I don't think this will work. It needs to be an extra `validates :file, presence: true`

slice: Alright, should be fixed now.

slice:
I can't seem to setup a local testing environment. I just get these errors:

```
NameError:
  uninitialized constant Sidekiq::Testing
```

Not sure what's going on here.

Gargron (approved):

Gargron:
`spec/presenters/instance_presenter_spec.rb` is now broken because the tests don't attach files :weary:

slice:
Yup, gonna try to fix these without testing locally (because I can't get them to work 😩)

slice:
Added a commit that correctly fabricates `SiteUpload#file` with a photo of a [Utah teapot](https://en.wikipedia.org/wiki/Utah_teapot).

PR #10442, opened 2019-04-01
Hide blocking accounts from blocked users

ClearlyClaire:
- [x] Revert UI change displaying blocking information to blocked users
- [x] Revert API change making that information available
- [x] Hide blocking accounts from blocked accounts in search results
- [x] Filter blocking account from API requests such as follows, followers, favorites, boosts
- [x] Return a 410 gone in relevelant API calls

Blocking accounts still show up in mentions, so that gives away the account is known to the instance, and therefore the blocked user is blocked by them… but it is somewhat subtle, and fixing that would be pretty costly

### DIFF
--- a/app/controllers/api/v1/accounts/following_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/following_accounts_controller.rb
@@ -25,8 +25,7 @@ class Api::V1::Accounts::FollowingAccountsController < Api::BaseController
   end
 
   def default_accounts
-    blocked_by_ids = Block.where(target_account_id: current_account.id).pluck(&:account_id)
-    Account.where.not(id: blocked_by_ids).includes(:passive_relationships, :account_stat).references(:passive_relationships)
+    Account.without_blocking(current_account).includes(:passive_relationships, :account_stat).references(:passive_relationships)
   end
 
   def paginated_follows
--- a/app/controllers/api/v1/accounts/following_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/following_accounts_controller.rb
@@ -25,7 +25,8 @@ class Api::V1::Accounts::FollowingAccountsController < Api::BaseController
   end
 
   def default_accounts
-    Account.includes(:passive_relationships, :account_stat).references(:passive_relationships)
+    blocked_by_ids = Block.where(target_account_id: current_account.id).pluck(&:account_id)
+    Account.where.not(id: blocked_by_ids).includes(:passive_relationships, :account_stat).references(:passive_relationships)
   end
 
   def paginated_follows
--- a/app/javascript/mastodon/features/account/components/header.js
+++ b/app/javascript/mastodon/features/account/components/header.js
@@ -94,15 +94,15 @@ class Header extends ImmutablePureComponent {
     let menu        = [];
 
     if (me !== account.get('id') && account.getIn(['relationship', 'followed_by'])) {
-      info.push(<span key='followed_by' className='relationship-tag'><FormattedMessage id='account.follows_you' defaultMessage='Follows you' /></span>);
+      info.push(<span className='relationship-tag'><FormattedMessage id='account.follows_you' defaultMessage='Follows you' /></span>);
     } else if (me !== account.get('id') && account.getIn(['relationship', 'blocking'])) {
-      info.push(<span key='blocked' className='relationship-tag'><FormattedMessage id='account.blocked' defaultMessage='Blocked' /></span>);
+      info.push(<span className='relationship-tag'><FormattedMessage id='account.blocked' defaultMessage='Blocked' /></span>);
     }
 
     if (me !== account.get('id') && account.getIn(['relationship', 'muting'])) {
-      info.push(<span key='muted' className='relationship-tag'><FormattedMessage id='account.muted' defaultMessage='Muted' /></span>);
+      info.push(<span className='relationship-tag'><FormattedMessage id='account.muted' defaultMessage='Muted' /></span>);
     } else if (me !== account.get('id') && account.getIn(['relationship', 'domain_blocking'])) {
-      info.push(<span key='domain_blocked' className='relationship-tag'><FormattedMessage id='account.domain_blocked' defaultMessage='Domain hidden' /></span>);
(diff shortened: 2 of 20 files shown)


app/javascript/mastodon/features/account/components/header.js
```diff
@@ -94,15 +94,15 @@ class Header extends ImmutablePureComponent {
     let menu        = [];
 
     if (me !== account.get('id') && account.getIn(['relationship', 'followed_by'])) {
-      info.push(<span key='followed_by' className='relationship-tag'><FormattedMessage id='account.follows_you' defaultMessage='Follows you' /></span>);
+      info.push(<span className='relationship-tag'><FormattedMessage id='account.follows_you' defaultMessage='Follows you' /></span>);
```
Gargron: These changes should be kept to avoid React warning

app/services/search_service.rb
```diff
@@ -12,6 +12,8 @@ def call(query, account, limit, options = {})
     default_results.tap do |results|
       if url_query?
         results.merge!(url_resource_results) unless url_resource.nil?
+        results[:accounts].reject! { |item| item.blocking?(@account) }
+        results[:statuses].reject! { |status| StatusFilter.new(status, @account).filtered? }
```
Gargron: Note: We run StatusFilter on ElasticSearch results too, though I don't remember if it's in this class or the controller. Might be worth considering just running it once before returning results, instead of in each branch separately.

app/controllers/api/v1/accounts/following_accounts_controller.rb
```diff
@@ -25,7 +25,8 @@ def load_accounts
   end
 
   def default_accounts
-    Account.includes(:passive_relationships, :account_stat).references(:passive_relationships)
+    blocked_by_ids = Block.where(target_account_id: current_account.id).pluck(&:account_id)
+    Account.where.not(id: blocked_by_ids).includes(:passive_relationships, :account_stat).references(:passive_relationships)
```
Gargron: Worth turning into a scope?

app/services/search_service.rb
ThibG: It's done in each branch separately, we could refactor it, but we wouldn't gain much.

Gargron (approved):

ealgase:
At least a 410 is returned so apps that try to be as transparent as possible to users aren't affected by this change.

dn3s:
@ealgase what's this mean? does this status code uniquely identify the account as blocked? if so this sounds like a bug

ThibG:
> @ealgase what's this mean? does this status code uniquely identify the account as blocked? if so this sounds like a bug

No, this does not uniquely identify the account as blocking you. It can also mean it has been blocked by the instance admin.

The only other reason for an account not showing up would be that it does not exist, but that's something a blocked user can check really easily. Also, the “there is a federation bug” cause wouldn't fly here if they had previously interacted with that account.

So I think returning a 410 is the best thing we can do.

ealgase:
> > @ealgase what's this mean? does this status code uniquely identify the account as blocked? if so this sounds like a bug
> 
> No, this does not uniquely identify the account as blocking you. It can also mean it has been blocked by the instance admin.
> 
> The only other reason for an account not showing up would be that it does not exist, but that's something a blocked user can check really easily. Also, the “there is a federation bug” cause wouldn't fly here if they had previously interacted with that account.
> 
> So I think returning a 410 is the best thing we can do.

That's my view, at least.  If we don't want to show "you are blocked" and we don't want to allow users to view the profiles of people who have blocked them, returning 410 makes the most sense.

ThibG:
This approach has serious downsides that I did not anticipate:
- The account will still show up if the client knows about it (caches it, e.g. if you haven't closed the Web UI from before you were blocked), but it's the same for suspended accounts, really
- More importantly, it makes you unable to block someone who blocked you first, meaning they're the one to decide when the block ends, and they can still see your public posts (but well, those are public posts). I am not sure how we should deal with that :/

ealgase:
Yep, there are many issues that stem from assuming that blocked users are always in the wrong and shouldn't be catered to.  They're users of the software too, if we knew all blocked users were bad, we'd automatically delete their accounts.

dn3s:
none of these measures need imply unilateral wrongness on the part of a blocked user, simply an acknowledgement that some users need to be able to make themselves invisible to other users, and that that's a valid requirement. however, it does make sense to extend this requirement to both parties.

if we want to level the playing field to better enable mutual blocks, a couple easy options are:
- have a free-form input for blocking, allowing users to block accounts that don't necessarily exist.
- show a block button for any blocking or deleted user, without differentiating between the two.

There are probably other, possibly more user friendly approaches to this, but they all require a simple shift, from blocking _users_ to blocking _usernames_.

dn3s:
it would also be nice if blocking weren't framed as necessarily adversarial. most blocks probably wouldn't need any sort of stealth whatsoever, but i think most of us can think of at least a few situations in our lives, ranging from awkward to physically dangerous, where that ambiguity is important. blocking really does have 2 different uses, there's the general "i don't want to interact with this person any more", and the more serious and specific, "my safety requires me to put as much between myself and this person as possible". ideally we'd separate these into separate functions, and make the latter as strong (while also abuse-resistant) as possible.

Cassolotl:
>More importantly, it makes you unable to block someone who blocked you first, meaning they're the one to decide when the block ends, and they can still see your public posts (but well, those are public posts). I am not sure how we should deal with that :/

A way to block hypothetical `@username@instance`s in settings? Like, a list of blocked users, with an "add user" textbox and button at the top?

Laurelai:
While were making an add user" textbox it should allow "@username@*" because many people make accounts with the same name across instances. One block would catch them all and all future accounts made with that name.

ThibG:
> ideally we'd separate these into separate functions, and make the latter as strong (while also abuse-resistant) as possible

What is the stronger one? One that prevents interactions as much as possible, or one that is more plausibly deniable and don't draw as much attention?

To some extent, I think this is covered by mutes and blocks, where basically, mutes do exactly the same thing as blocks except everything that might signal the other user that you're blocking them (which means they can still see your public/unlisted posts). In the end we cannot really have both…

PR #10449, opened 2019-04-01
Fix admin validation being too strict about usernames

### LINKED ISSUE #10446: Validation for usernames in site settings is too strict
This is related to the validation implemented in #7492 and found in version 2.8.0rc1

### Expected behaviour

Since this has been functional in the past, we should be able to use the @ sign in front of the username and spaces between the comma separated lists for both the Contact Username and Default follows for new users.

i.e. Contact Username [@admin]
i.e. Default follows for new users [@announcements, @host]

Alternate solution: Improve the error message to indicate what the problem is if Mastodon is dropping support for @ and lists separated by spaces and commas.

### Actual behaviour

Usernames need to be free of the @ sign and in the Default follows list, no spaces are allowed.  Since this was functioning previously, this will cause upgraded instances to not be able to get the site settings page to work without discovering these new requirements.

On an administrator account, go to Settings -> Administration -> Site Settings and try to enter usernames in the indicated locations.

### Specifications

Found in vanilla Mastodon 2.8.0rc1.

Gargron:
Fix #10446

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -50,7 +50,7 @@ class HomeController < ApplicationController
       push_subscription: current_account.user.web_push_subscription(current_session),
       current_account: current_account,
       token: current_session.token,
-      admin: Account.find_local(Setting.site_contact_username),
+      admin: Account.find_local(Setting.site_contact_username.strip.gsub(/\A@/, '')),
     }
   end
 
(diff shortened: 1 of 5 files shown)


ThibG:
The `Account.find_local(Setting.…)` pattern is also used in `app/models/concerns/account_finder_concern.rb`, `app/controllers/home_controller.rb` and `app/controllers/shares_controller.rb`

Gargron:
@ThibG This still needs a :heavy_check_mark: even if you added some commits yourself

ClearlyClaire (approved):

PR #10457, opened 2019-04-02
Add instance actor

ClearlyClaire:
- [x] Basic actor AP object and HTML representation
- [x] Webfinger loopback
- [ ] Store keypair
- [x] Auto-reject follow requests
- [x] Sign outgoing fetch requests
- [ ] Add tests

### DIFF
--- a/app/controllers/instance_actors_controller.rb
+++ b/app/controllers/instance_actors_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class InstanceActorsController < ApplicationController
+  layout 'public'
+
+  before_action :set_cache_headers
+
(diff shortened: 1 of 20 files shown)


ThibG:
An alternative implementation, that would solve the key storage part (for now, I'm using the first registered user's), would be to have a regular `Account` representing the instance actor. I am afraid this would require a lot of special-casing, though, as there are lots of differences between the instance actor and regular accounts (we may not want to handle follows in the same way, we certainly do not want to have notifications created when the instance actor is mentioned, nor do we want to create a timeline for it, or have it have settings etc.)

ThibG:
Instance actors have been added in #11321

PR #10465, opened 2019-04-03
Add invite comments

ClearlyClaire:


Gargron:
Was someone asking for this?

ThibG:
Yeah

app/models/invite.rb
```diff
@@ -12,6 +12,7 @@
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #  autofollow :boolean          default(FALSE), not null
+#  note       :text
```
Gargron: So is it a `comment` or a `note`?

Gargron (changes requested):

ThibG:
I think the reason why this hadn't been merged was to not add a new translatable string just before an upcoming release? If so, this could be merged now

ThibG:
@Gargron can we get that in before the 3.0 call for translations?

Gargron (approved):

PR #10467, opened 2019-04-03
Cache featured collections, as well as outbox, followers and following

ClearlyClaire:


ThibG:
When a status is suddenly boosted or replied out of the usual audience of an account, many instances may try to resolve that account at the same time, fetching various collections (`outbox`, `followers`, `following`, `featuring`). While the first ones are usually just a “wrapper” with just the number of items and a link to the first page, and are thus rather cheap, it doesn't hurt to let the reverse-proxy handle them. The `featured` collection is potentially more expensive, caching it both on the reverse-proxy and at the Rails level makes sense.

Gargron (approved):

PR #10473, opened 2019-04-04
Add ability to silence or lower the volume of the notification sound

lfuelling:
This PR adds the ability to either silence or lower the volume of the notification sound. It's implemented by adding two new checkboxes to `Settings > Notifications`.

screenshot of the new checkboxes

The defaults are as before: 
- lower volume: `false`
- sound activated: `true`


Gargron:
The notifications sound is specific to the notifications column and can be disabled through those column's settings. Admittedly it doesn't allow you to *lower* its volume, but I am not sure that warrants adding two options on a separate page.

lfuelling:
I didn't even know that the columns have their own settings.  

IMO Settings > Notifications would be the perfect place for those as well.

PR #10478, opened 2019-04-05
Add machine readable feature hashtag and endorsement to public profile page

noellabo:
On the public profile page, the features provided for display are not machine readable.

These features are based on what is already provided for display. Therefore, if necessary, it can be obtained by scraping. However, it is much better to get it with json or rss.

### Featured tag

* Fetch "https://dtp-mstdn.jp/@noellabo" with application/activity+json
```
  "featuredTag": [
    {
      "type": "Hashtag",
      "href": "https://dtp-mstdn.jp/@noellabo/tagged/crystal",
      "name": "#crystal",
      "statusesCount": "10",
      "lastStatusAt": "2019-02-13T02:39:55Z"
    }
  ],
```

* Fetch "https://dtp-mstdn.jp/@noellabo/tagged/crystal.rss" to get rss.
* The hashtag that can be specified is not limited to the featured hashtag. All hashtags are targeted.

### Endorsement

* Fetch "https://dtp-mstdn.jp/@noellabo" with application/activity+json
```
  "following": "https://dtp-mstdn.jp/users/noellabo/following",
  "followers": "https://dtp-mstdn.jp/users/noellabo/followers",
  "endorsed": "https://dtp-mstdn.jp/users/noellabo/endorsed",
```

* Fetch "https://dtp-mstdn.jp/users/noellabo/endorsed" with application/activity+json
* This works like a list of followers.
* When accessed from a web browser, they are listed in card format. However, there is currently no link to this page.
* Even if it is set to "Hide your network", endorsement is displayed.
```
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "id": "/users/noellabo/endorsed",
  "type": "OrderedCollection",
  "totalItems": 2,
  "first": "/users/noellabo/endorsed?page=1"
}
```

* Fetch "https://dtp-mstdn.jp/users/noellabo/endorsed?page=1" with application/activity+json
```
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "id": "/users/noellabo/endorsed?page=1",
  "type": "OrderedCollectionPage",
  "totalItems": 2,
  "partOf": "/users/noellabo/endorsed",
  "orderedItems": [
    "https://dtp-mstdn.jp/users/yamo",
    "https://dtp-mstdn.jp/users/assause"
  ]
}
```


Gargron:
I am tentatively in favour of this (wanted to do something like this eventually), but this is adding some brand new JSON vocabulary (but it's not reflected in the context yet)

Gargron:
New properties:

- `featuredTag`
- `endorsed`
- `lastStatusAt`
- `statusesCount`

The first two seem fine, but I am not sure about the last two. Re-using existing semantics wherever possible is preferable. Need some input from other contributors on this

nightpool:
Definitely think it's good to federate these in some way, but as gargron said i'm  👎 on `lastStatusAt` and `statusesCount`

I think the endorsed collection and the featuredTag array is pretty good, but I wonder if there's something more general we could be doing. I think the best way to do this is to add both the accounts and hashtags to the `tags` array, with a `featured` property:

```json
tag: [
   {
     "type": "Hashtag",
     "href": "https://dtp-mstdn.jp/explore/illustrator",
     "name": "#illustrator"
   },
   {
      "type": "Hashtag",
      "href": "https://dtp-mstdn.jp/@noellabo/tagged/%E3%83%8F%E3%83%83%E3%82%B7%E3%83%A5%E3%82%BF%E3%82%B0%E3%83%AA%E3%83%AC%E3%83%BC",
      "name": "#ハッシュタグリレー",
      "featured": true
    },
{
      "type": "Account",
      "href": "https://cybre.space/@nightpool",
      "name": "@nightpool@cybre.space",
      "featured": true
    }
],
```

that removes 90% of the complexity from this PR, which is a huge improvement! since we already have code for rendering hashtags and accounts in this list, i think that's the best of both worlds.

noellabo:
@nightpool What a simple and smart way!
I agree with this consolidation proposal.

For the account, is not the id property (account uri) required?

Picking up 4 random endorsements is a great idea for a display page, but json thinks it should list all collections. However, endorsement may be very many. What about the limit of the number? The endorsed collection may be easier to handle.

nightpool:
@noellabo ah yeah sorry, my fault, that should be `id` not `href`

I would be interested in statistics on the number of endorsed blogs, but practically, since we're only serializing a very small payload for each, I don't think it's a huge deal. certainly not going to be a problem for numbers in the low hundreds.

app/serializers/activitypub/actor_serializer.rb
```diff
@@ -173,12 +164,30 @@ def name
...
+    end
+  end
+
+  class AccountSerializer < ActivityPub::Serializer
+    include RoutingHelper
+
+    attributes :type, :id, :name, :featured
+
+    def type
+      'Accoount'
```
Gargron: @nightpool Your suggestion is actually a bit problematic because `Account` is not an ActivityPub type. There is `Person`, `Service`, etc... `Mention` is what comes up in `tag` usually. Perhaps this is where `Endorsement` would appear as a type.

noellabo: It has been made to reflect the `actor_type` of the account.

app/controllers/accounts_controller.rb
```diff
@@ -99,6 +101,10 @@ def username_param
     params[:username]
   end
 
+  def rss_url
+    "#{TagManager.instance.url_for(@account)}#{Addressable::URI.parse("/tagged/#{params[:tag]}").normalize if params[:tag].present?}.rss"
```
Gargron: I am not super comfortable with how this RSS-related code is implemented, so perhaps it could be separated into a different PR, so we can get the ActivityPub stuff in quicker?

app/serializers/activitypub/actor_serializer.rb
```diff
@@ -147,6 +147,50 @@ def name
...
+
+    def href
+      short_account_tag_url(Account.find(object.account_id), object.name)
+    end
+
+    def name
+      "##{object.name}"
+    end
+
+    def featured
```
Gargron: @nightpool Wait, so `featured` used to be the link to the pinned posts collection. Is it okay to re-use it for a boolean value with a different semantic meaning? I have a feeling "maybe not"

app/controllers/accounts_controller.rb
noellabo: OK. Divide into another PR.

noellabo: It moved to https://github.com/tootsuite/mastodon/pull/10502

app/serializers/activitypub/actor_serializer.rb
trwnh: yeah pretty sure this is not going to be okay for any software doing proper json-ld. maybe 'promoted' or 'endorsed' or 'highlighted' or 'recommended' or some other similar word. i'm personally for 'recommended' as in 'i recommend you browse this hashtag of mine' or 'i recommend you check out this cool profile', but no strong feelings.

noellabo:
It is an alternative.

### featuredTag

* Need new properties but do not affect existing properties
* Featured tag collection is easy to use

```
"featuredTag": [
   {
      "type": "Hashtag",
      "href": "https://dtp-mstdn.jp/@noellabo/tagged/%E3%83%8F%E3%83%83%E3%82%B7%E3%83%A5%E3%82%BF%E3%82%B0%E3%83%AA%E3%83%AC%E3%83%BC",
      "name": "#ハッシュタグリレー",
    },
    {
      "type": "Person",
      "id": "https://cybre.space/users/nightpool",
      "name": "@nightpool@cybre.space",
    }
],
```

### featured

* It is necessary to extend the definition of `featured`, but the objects representing the same featured are gathered in one place

```
"featured": [
  "https://dtp-mstdn.jp/users/noellabo/collections/featured",
   {
      "type": "Hashtag",
      "href": "https://dtp-mstdn.jp/@noellabo/tagged/%E3%83%8F%E3%83%83%E3%82%B7%E3%83%A5%E3%82%BF%E3%82%B0%E3%83%AA%E3%83%AC%E3%83%BC",
      "name": "#ハッシュタグリレー",
    },
    {
      "type": "Person",
      "id": "https://cybre.space/users/nightpool",
      "name": "@nightpool@cybre.space",
    }
],
```

nightpool:
oh I forgot we had featured for pinned posts. that's even better let's put
them there.

nightpool:
(if that's compatible with existing impls, I don't think anyone uses pinned
posts except us, so it should be easy to check)

noellabo:
Well, which expression do you think is good?

Properties like `featured` are consistently considered collections. I think that way is easy to understand.

It seems that there are not many examples of properties treated as Boolean, but the meaning is clear.

What about a collection of values, such as `flag`, `mark`, `option`, and `subType`? In this case, we treat `featured` as the value. Other values ​​may also be specified.

PR #10483, opened 2019-04-06
Add `tootctl accounts reset-relationships`

noellabo:
Reset all followee and / or followers for a user.

* Clean up after too many follow imports.
* Reduce server load without suspending users who have not been used for a long time.
* Reset the followers of users who performed follower collection activities that violate the policy.


Gargron:
I am worried about the too wide meaning of "reset" for this command, since it is imaginable that "account reset" could mean removing posts, resetting password, etc. Maybe "reset-relationships"?

noellabo:
Oh, I couldn't think of that name.
Thank you for the good suggestions!
I will change the name.

Gargron (approved):

Issue #10323, opened 2019-03-20
Keybase integration validation error, username regular expression and no contacts


noellabo:
Errors may occur when performing a Config Validation following the instructions in Keybase's [Proof Integration Guide](https://keybase.io/docs/proof_integration_guide).

```bash
http GET https://mastodon.social/.well-known/keybase-proof-config > config.json
http POST "https://keybase.io/_/api/1.0/validate_proof_config.json" config="$(jq -c . < config.json)"
```

### Error in username regular expression

```json
{
    "status": {
        "code": 218,
        "desc": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 122",
        "fields": {
            "config": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 122"
        },
        "name": "GENERIC"
    }
}
```
`"/[a-z0-9_]+([a-z0-9_\\.-]+[a-z0-9_]+)?/i"`
(Account::USERNAME_RE.inspect in [config_serializer.rb](https://github.com/tootsuite/mastodon/blob/master/app/lib/proof_provider/keybase/config_serializer.rb#L36))

The escape part of `.` seems to be a problem.

### Required contacts have not been set up

```json
{
    "status": {
        "code": 100,
        "desc": "missing or invalid inputs {\"contact\":\"Error: wrong number in list\"}",
        "fields": {
            "config": "missing or invalid inputs {\"contact\":\"Error: wrong number in list\"}"
        },
        "name": "INPUT_ERROR"
    }
}
```

It will be an error if the contact is not set.
There are often such servers. Is there a good way?

ThibG:
I don't have the regexp error, but yeah, not having an email contact defined causes that. Maybe we shouldn't enable the feature at all when it's not set?

Gargron:
You might say the feature already isn't enabled when the config doesn't validate.

ThibG:
I mean, not exposing the routes and not showing the settings page

syui:
I have the same trouble as you have.

```sh
$ curl -X POST -d "config=`jq -c . < ./keybase-proof-config.json`" https://keybase.io/_/api/1.0/validate_proof_config.json

{"status":{"code":218,"desc":"Unable to parse JSON SyntaxError: Unexpected end of JSON input","fields":{"config":"Unable to parse JSON SyntaxError: Unexpected end of JSON input"},"name":"GENERIC"}}
```

noellabo:
[...]

PR #10487, opened 2019-04-06
Fix config serializer regex

noellabo:
The character class `.` does not need to be escaped, so delete `\`.

### DIFF
--- a/app/lib/proof_provider/keybase/config_serializer.rb
+++ b/app/lib/proof_provider/keybase/config_serializer.rb
@@ -34,7 +34,7 @@ class ProofProvider::Keybase::ConfigSerializer < ActiveModel::Serializer
   end
 
   def username
-    { min: 1, max: 30, re: '[a-z0-9_]+([a-z0-9_\.-]+[a-z0-9_]+)?' }
+    { min: 1, max: 30, re: '[a-z0-9_]+([a-z0-9_.-]+[a-z0-9_]+)?' }
   end
 
   def prefill_url


app/lib/proof_provider/keybase/config_serializer.rb
```diff
@@ -34,7 +34,7 @@ def description
   end
 
   def username
-    { min: 1, max: 30, re: '[a-z0-9_]+([a-z0-9_\.-]+[a-z0-9_]+)?' }
+    { min: 1, max: 30, re: '[a-z0-9_]+([a-z0-9_.-]+[a-z0-9_]+)?' }
```
nightpool: I feel like if this is causing issues in their JSON parsing or in our JSON serialization, that's a bigger issue and we shouldn't paper over it by removing this slash.

nightpool:
@ThibG indicated in #10323 that he didn't get the same regex error when validating the keybase config. Maybe this is something unique to your local environment or server? Let's figure that out before merging this change.

Gargron:
@xgess Is there an issue with the current regex?

noellabo:
Please try the following.

### escaped

escaped.json
```
cat <<'_EOL_' >escaped.json
{"version":1,"domain":"example.com","display_name":"example.com","username":{"min":1,"max":30,"re":"[a-z0-9_]+([a-z0-9_\\.-]+[a-z0-9_]+)?"},"brand_color":"#282c37","logo":{"svg_black":"https://example.com/black.svg","svg_full":"https://example.com/full.svg"},"description":"example.com","prefill_url":"https://example.com/settings/identity_proofs/new?provider=keybase&provider_username=%{kb_username}&token=%{sig_hash}&user_agent=%{kb_ua}&username=%{username}","profile_url":"https://example.com/@%{username}","check_url":"https://example.com/api/proofs?provider=keybase&username=%{username}","check_path":["signatures"],"avatar_path":["avatar"],"contact":["admin@example.com"]}
_EOL_
```

run keybase validator
```
http POST "https://keybase.io/_/api/1.0/validate_proof_config.json" config="$(jq -c . < escaped.json)" -b
```

result
```
{
    "status": {
        "code": 218,
        "desc": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 120",
        "fields": {
            "config": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 120"
        },
        "name": "GENERIC"
    }
}
```

### unescaped

unescaped.json
```
cat <<'_EOL_' >unescaped.json
{"version":1,"domain":"example.com","display_name":"example.com","username":{"min":1,"max":30,"re":"[a-z0-9_]+([a-z0-9_.-]+[a-z0-9_]+)?"},"brand_color":"#282c37","logo":{"svg_black":"https://example.com/black.svg","svg_full":"https://example.com/full.svg"},"description":"example.com","prefill_url":"https://example.com/settings/identity_proofs/new?provider=keybase&provider_username=%{kb_username}&token=%{sig_hash}&user_agent=%{kb_ua}&username=%{username}","profile_url":"https://example.com/@%{username}","check_url":"https://example.com/api/proofs?provider=keybase&username=%{username}","check_path":["signatures"],"avatar_path":["avatar"],"contact":["admin@example.com"]}
_EOL_
```

run keybase validator
```
http POST "https://keybase.io/_/api/1.0/validate_proof_config.json" config="$(jq -c . < unescaped.json)" -b
```

result
```
{
    "status": {
        "code": 0,
        "name": "OK"
    }
}
```

### from mastodon.social

```
http GET https://mastodon.social/.well-known/keybase-proof-config > config.json
```

run keybase validator
```
http POST "https://keybase.io/_/api/1.0/validate_proof_config.json" config="$(jq -c . < config.json)" -b
```

result
```
{
    "status": {
        "code": 218,
        "desc": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 121",
        "fields": {
            "config": "Unable to parse JSON SyntaxError: Unexpected token . in JSON at position 121"
        },
        "name": "GENERIC"
    }
}
```

----
In the character class, whatever character escapes is interpreted as that character as fallback, so it should work anyway.

However, key-based validators seem to be strict.

There are only four characters in the character class that have special meanings: `^` `-` `]` `\` There is no special meaning in `.`.

The trailing `-` and the `^` other than the beginning of a line are also characters without special meaning.

Although not included in this request, `Account::USERNAME_RE` and `Account::MENTION_RE` also have excessive escapes. However, because it works correctly, it is not necessary to fix it if it is only used internally. (It's refactoring)

noellabo:
There seems to be a problem with my `http` usage.
Attempting other verification methods did not result in an error.

I will organize later, but in any case there is no problem. This request is not required. Think of it as refactoring.

noellabo:
The correct calling method is as follows.
```
http POST "https://keybase.io/_/api/1.0/validate_proof_config.json" config=@config.json
```

or

```
curl -X POST https://keybase.io/_/api/1.0/validate_proof_config.json --data-urlencode config@config.json
```

or directly

```
http GET "https://keybase.io/_/api/1.0/validate_proof_config.json" config_url="https://mastodon.social/.well-known/keybase-proof-config"
```

Gargron (approved):

xgess:
took me a bit to verify this end-to-end. yes. this will work fine and i do believe it to be an improvement. despite the validation error that happens when this is parsed into and out of json (i.e. adding an extra `\` before getting it into keybase's servers), they both work fine on our end as long as it only actually gets persisted by us with at most one slash. but yes. i didn't know periods needn't be escaped inside the character class. 🆒 thanks for doing this.

Issue #10388, opened 2019-03-26
Animate GIF avatars on profile pages when hovered


dnaf:
Currently, animated avatars on profile pages are always paused. It'd be nice if they would animate when hovered, as they do in the web app.

PR #10493, opened 2019-04-07
Animate profile gifs on hover

hyenagirl64:
Quick fix for #10388

Replicates the animate on hover functionality of gifs on the main page for user profile pages.


ThibG:
I'm afraid this won't work due to the Content Security Policy: we don't allow inline scripts for security reasons. A better solution would probably be to have `data-` attributes for representing the static and animated images and have something in `app/javascript/packs/public.js` handle them.

hyenagirl64:
thanks for the tip.  I'll close the request and have another go.

PR #10495, opened 2019-04-07
Export and import show_reblogs together with following list

ClearlyClaire:


Gargron (approved):

hakaba-hitoyo:
Already merged, but I inform this PR breaks the interoperability towards [Misskey](https://github.com/syuilo/misskey/issues/4816), Pleroma, or other implementations. New data format has better to be optional.

Gargron:
Sorry for the trouble :frowning:

PR #10502, opened 2019-04-08
Add an RSS feed tagged to a public profile page

noellabo:


app/controllers/accounts_controller.rb
```diff
@@ -99,6 +101,10 @@ def username_param
     params[:username]
   end
 
+  def rss_url
+    "#{TagManager.instance.url_for(@account)}#{Addressable::URI.parse("/tagged/#{params[:tag]}").normalize if params[:tag].present?}.rss"
```
Gargron:
```ruby
if tag_requested?
  short_account_tag_url(@account, params[:tag], format: 'rss')
else
  short_account_url(@account, format: 'rss')
end
```

app/controllers/accounts_controller.rb
```diff
@@ -128,7 +134,7 @@ def replies_requested?
   end
 
   def tag_requested?
-    request.path.ends_with?(Addressable::URI.parse("/tagged/#{params[:tag]}").normalize)
+    request.path.split('.')[0].ends_with?(Addressable::URI.parse("/tagged/#{params[:tag]}").normalize)
```
Gargron: It is prettier to use `first` instead of `[0]`

app/serializers/rss/account_serializer.rb
```diff
@@ -5,12 +5,12 @@ class RSS::AccountSerializer
...
   include RoutingHelper
 
-  def render(account, statuses)
+  def render(account, statuses, tag)
     builder = RSSBuilder.new
 
     builder.title("#{display_name(account)} (@#{account.local_username_and_domain})")
            .description(account_description(account))
-           .link(TagManager.instance.url_for(account))
+           .link("#{TagManager.instance.url_for(account)}#{Addressable::URI.parse("/tagged/#{tag}").normalize if tag.present?}")
```
Gargron: Since `RoutingHelper` is already included, it is probably better to use route helpers:

```ruby
.link(tag.present? ? short_account_tag_url(account, tag) : short_account_url(account))
```

app/controllers/accounts_controller.rb
```diff
@@ -25,6 +25,7 @@ def show
         @statuses        = filtered_status_page(params)
         @statuses        = cache_collection(@statuses, Status)
 
+        @rss_url = rss_url
```
Gargron: Align assignment with `@statuses` (remove newline above) and add newline before conditional block.

Gargron (changes requested):

Gargron (approved):

PR #10508, opened 2019-04-08
Fix batch actions not working on pending accounts

ClearlyClaire:
`:update` works on individual collection members, not on the collection itself, so use a new route for batch actions.
Also make it clear when no accounts were selected.

### DIFF
--- a/app/controllers/admin/pending_accounts_controller.rb
+++ b/app/controllers/admin/pending_accounts_controller.rb
@@ -8,11 +8,11 @@ module Admin
       @form = Form::AccountBatch.new
     end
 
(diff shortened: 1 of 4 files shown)


Gargron:
Oh, right. I think it should have been a `resource` with `:show` and  `:update` :confused: That's how other batch actions are implemented

ThibG:
Actually, this is not handled in one single consistent way throughout the code. It's done using `:create`, sometimes on the collection itself, sometimes on a new endpoint. It's a bit weird tbh.

Gargron (approved):

PR #10513, opened 2019-04-08
Tweak filtering of newly-blocked/muted accounts

ClearlyClaire:
- Do not send `delete` events over the streaming server when retroactively filtering toots from the home timeline:
  This causes a regression in which, when several streaming clients are connected, only the one issuing the mute/block will filter already-loaded toots, in the same way unfollowing someone is currently handled.
  On the other hand, it allows fixing the fact that notifications from newly-muted users (for which notifications are still allowed) temporarily disappear.
- Retroactively filter toots from lists as well as the home timeline:
  Currently, the web client will filter blocked/muted users' toots from the timeline, but that filtering is not made server-side.
  A possible downside is a performance concern, as the size of a timeline is bounded, but the number of lists theoretically isn't.


ykzts (approved):

SuperSandro2000:
bump

PR #10515, opened 2019-04-08
i18n: Russian localization update

MasterGroosha:
It has been a long time since I last updated 🇷🇺 localization for Mastodon (sorry).
I'll try to update everything asap, translating file by file and checking autobuild.
Everything is tested manually on my local dev instance, so at least it works fine and looks nice (IMO).

So, changelog (will be updated eventually):
• Added translation for polls.
• Added translation for quick filters in notifications.
• Added translation for "Follows & Followers" section (which has a bug, will link later).
• Added translation for tutorial.
• Added translation for various dashboard sections.
• Added translation for various other settings.

• Changed "Joined date" format in profiles from, for example "Mar 2019" to "03.2019" because rails has incorrect "normal case" format for months. I guess it's not Mastodon's problem, but [Rails's](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/ru.yml)


MasterGroosha:
I'm not sure what's the conflict with `config/locales/ru.yml`. From what I see, in [this commit](https://github.com/tootsuite/mastodon/commit/9b0d8f74cbbb04c4b1227293137774db78b67896#diff-8a74326a0a6cc6a54fae6faf02937eb3) two strings were removed, though they're used in Web client (checked it myself).

Gargron (approved):

inexcode:
Just found a serious issue with this one: "Followed by them" is translated as "Blocked by them", which brings a lot of confusion.
[Line 350 of config/locales/ru.yaml](https://github.com/tootsuite/mastodon/blob/master/config/locales/ru.yml#L350).
Should be something like 
``` yaml
total_followed_by_them:  Их подписчиков
```

MasterGroosha:
@inexcode Oh, shame on me! You're right.
Since Mastodon uses Crowdin now, I fixed that string directly [there](https://crowdin.com/translate/mastodon/18/en-ru#1334)

PR #10518, opened 2019-04-09
Animate gif on profile page

hyenagirl64:
Second attempt to fix issue #10388

handling it via assigning event handlers in public.js


ThibG:
Using a custom opengraph tag to store that value seems weird to me. I think I would have used custom `data-` attributes on the image such as `data-original` and `data-static`. Not too sure about a toggle property but it would probably fit right in there too.

hyenagirl64:
well, if I'm adding custom attributes, I could store the information as to whether or not to do the hover effect on the image in another attribute, rather than relying on whether or not the toggle is the same as the source.  That would probably be a less bug-prone way to do it as well, since it would make the functionality more explicit.

PR #10534, opened 2019-04-10
Change HTML validator to ignore all errors except unmatched tags

Gargron:


ykzts (approved):

app/validators/html_validator.rb
```diff
@@ -1,18 +1,20 @@
...
+    record.errors.add(attribute, I18n.t('html_validator.invalid_markup', error: errors.first.to_s)) unless errors.empty?
   end
 
   private
 
   def html_errors(str)
-    fragment = Nokogiri::HTML.fragment(str)
-    fragment.errors
+    fragment = Nokogiri::HTML.fragment(options[:wrap_with] ? "<#{options[:wrap_with]}>#{str}</#{options[:wrap_with]}>" : str)
+    fragment.errors.select { |error| ERROR_RE =~ error.message }
```
ThibG: The `wrap_with` thing will introduce an offset in the reported line/column of the error, which might be confusing and we might want to take care of.

Kjwon15:
```html
After this commit, I cannot insert `<ul>` in server description.
It causes `ERROR: Unexpected end tag : p` even if I just entered `<ul></ul>`

salvadorpla:
I can't modify Site settings because three errors about 'ERROR: Unexpected end tag : p'
In Server description it claims : contains invalid HTML markup: 6:253: ERROR: Unexpected end tag : p
In Custom extended information it claims : contains invalid HTML markup: 2:308: ERROR: Unexpected end tag : p
And in Custom terms of service it claims : contains invalid HTML markup: 2:84: ERROR: Unexpected end tag : p.
The current situation gets better, my custom CSS is not working.

psydwannabe:
Your HTML Validator is claiming I have unmatched tags in HTML that clearly has matched tags.

I am no longer able to update my instance's server description or custom extended information as of release 2.8 because of this code.

Gargron:
@Kjwon15 @psydwannabe You are probably both encountering the issue that server description is pre-wrapped in a `<p>` tag, and those do not support other block-level tags within themselves. Prior to this validation you just didn't know it was happening. Let me know if that's the case. You can close the invisible p tag and then re-open it at the end to solve this issue.

PR #10544, opened 2019-04-10
compare usernames case-insensitively on new proof creation flow

xgess:
it's possible to get an error message if you attempt to prove the correctly logged in username but cased differently. the error message isn't the worst, but this now does the intuitive thing.

### DIFF
--- a/app/controllers/settings/identity_proofs_controller.rb
+++ b/app/controllers/settings/identity_proofs_controller.rb
@@ -18,7 +18,7 @@ class Settings::IdentityProofsController < Settings::BaseController
       provider_username: params[:provider_username]
     )
 
-    if current_account.username == params[:username]
+    if current_account.username.downcase == params[:username].downcase
       render layout: 'auth'
     else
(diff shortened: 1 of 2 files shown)


app/controllers/settings/identity_proofs_controller.rb
```diff
@@ -18,7 +18,7 @@ def new
       provider_username: params[:provider_username]
     )
 
-    if current_account.username == params[:username]
+    if current_account.username.downcase == params[:username].downcase
```
Gargron: `if current_account.username.casecmp(params[:username]).zero?`

app/controllers/settings/identity_proofs_controller.rb
```diff
@@ -18,7 +18,7 @@ def new
       provider_username: params[:provider_username]
     )
 
-    if current_account.username == params[:username]
+    if current_account.username.casecmp(params[:username]).zero?
```
xgess: +1

Gargron (approved):

PR #10546, opened 2019-04-10
increase allowable length of remote proof username

xgess:
keybase usernames can actually be up to 16 characters. i must have read it wrong from the beginning. i'm really sorry about this because the error message isn't great. 

and if i'm in here anyway, it's silly to bind this to keybase's username restrictions going forward, so i increased it to the mastodon allowable username character limit.


Gargron (approved):

xgess:
thanks!

PR #10547, opened 2019-04-10
Fix Keybase verification using wrong domain for remote accounts

Gargron:


Gargron:
cc @xgess

app/lib/proof_provider/keybase.rb
```diff
@@ -2,7 +2,7 @@
 
 class ProofProvider::Keybase
   BASE_URL = ENV.fetch('KEYBASE_BASE_URL', 'https://keybase.io')
-  DOMAIN = ENV.fetch('KEYBASE_DOMAIN', Rails.configuration.x.local_domain)
+  DOMAIN   = ENV.fetch('KEYBASE_DOMAIN', Rails.configuration.x.local_domain)
```
xgess:
```suggestion
  DOMAIN = ENV.fetch('KEYBASE_DOMAIN', Rails.configuration.x.local_domain)
```

Gargron: I like to align assigments :+1:

xgess: :cool:

PR #10553, opened 2019-04-11
Fixed Healthchecks in docker-compose.yml

### LINKED ISSUE #10554: docker healthcheck doesn't work after upgrade to v2.8.0
docker ps:
tootsuite/mastodon:v2.8.0    "/tini -- yarn start"    5 hours ago         Up 5 hours (unhealthy)

docker-compose.yml is original at healtcheck:
 healthcheck:
      test: ["CMD-SHELL", "wget -q --spider --header 'x-forwarded-proto: https' --proxy off localhost:4000/api/v1/streaming/health || exit 1"]

happens on web and streaming-container.

fabianonline:
The healthchecks in docker-compose.yml for streaming and web are not working. The `--proxy off` is interpreted as switch `--proxy` and `off` as URL, having wget make two requests with at least the one to `http://off/` failing. This will lead to the healthcheck always failing.

This PR just replaces `--proxy off` with `--proxy=off`, fixing this bug.

Resolve #10554 (edited by @ykzts)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,7 +43,7 @@ services:
       - external_network
       - internal_network
     healthcheck:
-      test: ["CMD-SHELL", "wget -q --spider --header 'x-forwarded-proto: https' --proxy off localhost:3000/api/v1/instance || exit 1"]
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Shleeble:
wow. I expected the healthcheck to cause endless restarts or something but just seems to mark as unhealthy. good catch tho

Gargron (approved):

PR #10555, opened 2019-04-11
Add emoji suggestions to CW and poll option fields

ClearlyClaire:
I am not very happy with the CSS, especially for poll options… but it seems to work

EDIT: I'm a bit happier with it now, but reviews are welcome


ThibG:
There's one thing missing compared for the CW and poll choices inputs compared to the compose textarea: putting the caret just right after the newly-inserted suggestion rather than at the end of the input.

Gargron (approved):

PR #10563, opened 2019-04-12
GlitchSocial-style content warnings (hidden content includes attachments)

BenLubar:
I don't have a vanilla Mastodon instance set up to test this with, so could someone else verify that content warnings don't crash or stop working or make demons fly out of your nose?

This also adds the icons to represent what type of content was hidden to the web interface but not to the public interface.

---

This change means that images and videos on posts with content warnings can no longer be made visible without first making the text visible, and that polls on posts with content warnings will be hidden unless the content warning is expanded.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -214,6 +214,7 @@ class Status extends ImmutablePureComponent {
 
   render () {
     let media = null;
+    let mediaIcon = null;
     let statusAvatar, prepend, rebloggedByText;
 
     const { intl, hidden, featured, otherAccounts, unread, showThread } = this.props;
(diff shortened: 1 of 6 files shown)


ThibG:
I (obviously) like the idea, and it's a good start, but it introduces several regressions:
1. video size/aspect ratio isn't set correctly (the same thing probably occurs for images)
2. clicking on a media will open the status in the dynamic column, in addition to the normal action

In addition, it might cause issues with regards to timeline jumps, but I'd have to check.

1. is caused by some components (`MediaGallery`, `Video` and `Card`) computing their width at the time they are mounted, but if they are put behind CWs, they are hidden, and their width is thus 0.
In glitch-soc, this is handled by updating their width in `componentDidUpdate`, but there are a few tricks with the `Video` player (I encourage you to look at glitch-soc's code)
2. is caused by the status content having a click handler while the media previously did not. This could probably be fixed by patching `StatusContent`'s `handleMouseUp`. Again, I encourage you to read glitch-soc's version of the code.

BenLubar:
Ah, forgot to update this issue with my results:

It seems that the public UI breaks if there are content warnings (images cannot be seen, but the HTML seems like it *should* work). Polls do show up correctly in the public UI.

PR #10575, opened 2019-04-13
Prevent silenced local users from notifying remote users not following them

ClearlyClaire:
This is an attempt to extend the local restrictions of silenced users to the
federation.

### DIFF
--- a/app/lib/activitypub/tag_manager.rb
+++ b/app/lib/activitypub/tag_manager.rb
@@ -65,7 +65,14 @@ class ActivityPub::TagManager
     when 'unlisted', 'private'
       [account_followers_url(status.account)]
     when 'direct', 'limited'
-      status.active_mentions.map { |mention| uri_for(mention.account) }
+      if status.account.silenced?
+        # Only notify followers if the account is locally silenced
+        account_ids = status.active_mentions.pluck(:account_id)
+        to = status.account.followers.where(id: account_ids).map { |account| uri_for(account) }
+        to.concat(FollowRequest.where(target_account_id: status.account_id, account_id: account_ids).map { |request| uri_for(request.account) })
+      else
+        status.active_mentions.map { |mention| uri_for(mention.account) }
+      end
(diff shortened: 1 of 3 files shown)


ThibG:
When a silenced local users, mentionned users that do not follow (or request to follow) them won't appear in the `to` and `cc` fields, but messages will still get delivered to mentioned user's inbox.
While other implementations may interpret this differently, this should allow mentioned users of existing Mastodon instances to be able to see the toots if they look for them while preventing them from appearing in the recipient's notifications.

ThibG:
I haven't actually tested it yet.
Also, I'm not super happy with the audience of a toot possibly changing over time when users get silenced/unsilenced, but that shouldn't be a problem in practice, and the same could be said of remote mentioned users disappearing or getting banned I guess

app/lib/activitypub/tag_manager.rb
```diff
@@ -86,7 +93,16 @@ def cc(status)
       cc << COLLECTIONS[:public]
     end
 
-    cc.concat(status.active_mentions.map { |mention| uri_for(mention.account) }) unless status.direct_visibility? || status.limited_visibility?
+    unless status.direct_visibility? || status.limited_visibility?
+      if status.account.silenced?
+        # Only notify followers if the account is locally silenced
+        account_ids = status.active_mentions.pluck(:account_id)
+        cc.concat(status.account.followers.where(id: account_ids).map { |account| uri_for(account) })
+        cc.concat(FollowRequest.where(target_account_id: status.account_id, account_id: account_ids).map { |request| uri_for(request.account) })
```
nightpool: what's up with this line?

ThibG: I'm not sure what you mean. It's to allow silenced local users to mention (and notify) remote users who have requested to follow them.
This is the same logic as in `NotifyService` for local notifications.

Gargron (approved):

PR #10605, opened 2019-04-19
Fix modal items cannot scroll on touch devices

kedamaDQ:
This PR fixes the actions-modal to allow scrolling items on touch devices which are short in height.

before:
cannot-scroll

after:
can-scroll

without-status:
can-scroll-without-status

different sizing:
can-scroll-diff-sizing


ClearlyClaire (approved):

kedamaDQ:
Thanks!

PR #10614, opened 2019-04-21
Do not retry processing ActivityPub jobs raising validation errors

ClearlyClaire:
Jobs yielding validation errors most probably won't ever be accepted,
so it makes sense not to clutter the queues with retries.

### DIFF
--- a/app/workers/activitypub/processing_worker.rb
+++ b/app/workers/activitypub/processing_worker.rb
@@ -7,5 +7,9 @@ class ActivityPub::ProcessingWorker
 
   def perform(account_id, body, delivered_to_account_id = nil)
     ActivityPub::ProcessCollectionService.new.call(body, Account.find(account_id), override_timestamps: true, delivered_to_account_id: delivered_to_account_id, delivery: true)
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.error "#{e.class}: #{e.message}"
+    Rails.logger.error e.backtrace.join("\n")
+    # Do not retry
   end
 end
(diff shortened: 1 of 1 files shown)


app/workers/activitypub/processing_worker.rb
```diff
@@ -7,5 +7,9 @@ class ActivityPub::ProcessingWorker
 
   def perform(account_id, body, delivered_to_account_id = nil)
     ActivityPub::ProcessCollectionService.new.call(body, Account.find(account_id), override_timestamps: true, delivered_to_account_id: delivered_to_account_id, delivery: true)
+  rescue ActiveRecord::RecordInvalid => e
+    Rails.logger.error "#{e.class}: #{e.message}"
+    Rails.logger.error e.backtrace.join("\n")
```
Gargron: I don't like the log dumping. Feels unnecessary...

ThibG: I think it is useful to at least know that something's up…?

Gargron: In other places we use the `debug` level, not `error`. Also, I am always uncomfortable with manually printing the backtrace anywhere.

ThibG: Honestly I would have preferred sidekiq to just handle it as an error, but without scheduling the job for retry. Unfortunately that doesn't appear to be possible, the only way to avoid a retry being to catch the exception, and thus have nothing printed.

I guess I can lower the log level to `debug`, but `debug` is already filled with so much stuff this would go unnoticed.

Gargron: Sorry, could we stay somewhat consistent with https://github.com/tootsuite/mastodon/blob/e1ec3a9f09add8a34b1a77ce217be8e51ebd4936/app/services/verify_link_service.rb#L13-L16? No other exception rescue prints the backtrace anywhere.

ThibG: afaik we do on `Mastodon::HostValidationError`

Gargron (approved):

PR #10630, opened 2019-04-23
Add blurhash

Gargron:
Blurhash encodes an image as a short string of ASCII characters that can be saved in the database and transmitted via the API. It allows displaying a blurred preview on the clients before the real image file loads. It's also blurred enough that it can be displayed instead of the black box when media is hidden.

grafik

Making the hidden media box a little more visually appealing should hopefully help the adult content creators on the network. But also beyond that, being able to show a nice placeholder is especially useful for media that has been uncached and that is being (slowly) reloaded through the media proxy.

It's also potentially useful for federation with servers from which you do not want to store/load media. The short ASCII code is safe to store and display to the end-user, allowing them to make the decision to click-through to the original resource.


Shleeble:
Excellent little UX tweak.

ykzts (approved):

PR #10631, opened 2019-04-24
Allow turning keybase off instance-wide

ClearlyClaire:


Nutomic:
Does this make keybase opt-out? I would honestly prefer if it was opt-in, seeing as it is a for-profit company.

ThibG:
Yeah, it makes it opt-out. I figured making it opt-in could be confusing with all the communication that hinted at it being automatic (which it currently *kinda* is?)

app/controllers/admin/dashboard_controller.rb
```diff
@@ -30,6 +30,7 @@ def index
       @trending_hashtags     = TrendingTags.get(7)
       @profile_directory     = Setting.profile_directory
       @timeline_preview      = Setting.timeline_preview
+      @keybase_integration   = Setting.enable_keybase
```
Gargron: Technically you are disabling the entire identity proofs feature, not just Keybase. I suggest renaming all variables accordingly.

Gargron (changes requested):

app/controllers/admin/dashboard_controller.rb
ThibG: Well, I did really intend to disable keybase itself, not the identity proofs feature which currently only implement keybase. I considered checking whether the provider was “keybase” in the proof creation thing but figured we would change it if/whenever we add other identity proof providers.

ThibG: I mean, I think it makes sense for the option to be specifically about keybase, as people might not have the same opposition towards other hypothetical identity proof providers.
Would you want me to explicitly check for the `keybase` provider in the code? Or change the option anyway to control the whole feature rather than keybase itself?

Nutomic:
still interested

SuperSandro2000:
bump

koyuawsmbrtn:
Since Keybase has been aqcuired by Zoom it'll be a good time to merge it under a new PR with mentioned changes

@ThibG @Gargron

anahata0108:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

this bot doesn't serve any useful purpose. Can we remove it from this project? All it does is close legitimate bugs / PRs.

patcon:
@anahata0108 I feel for you, but if you wish to have that convo, could you open a new issue? Thx :)

For cross-linking if you do that:
- https://github.com/tootsuite/mastodon/issues/12225
- https://github.com/tootsuite/mastodon/pull/12217
- https://github.com/tootsuite/mastodon/pull/12207

SuperSandro2000:
Bump

koyuawsmbrtn:
Ok thanks for nothing, stalebot

SuperSandro2000:
@stale 🖕

PR #10647, opened 2019-04-28
Add a keyboard shortcut to hide/show media

ClearlyClaire:


ykzts (approved):

Gargron:
Maybe appropriate time to ask if maybe opening media & text should be a single action and single button instead of having two separate hotkeys

ThibG:
Even with glitch-soc's way of hiding the media behind the CW, I prefer it being a two-step process, because the text content can help me better assess what the media may contain.

However, some people think differently, so there's an option in glitch-soc to automatically reveal media that was hidden behind a CW when unfolding the CW.

Issue #10578, opened 2019-04-14
Some posts are classified as Chinese while they are actually Japanese


carbon-twelve:
### Expected behaviour
No `lang="zh"` attribute on a div element holding the toot below (see Steps to reproduce the problem)

### Actual behaviour
The div element has a `lang="zh"` attribute.

### Steps to reproduce the problem
Post "来週土曜" with Mastodon posting language set to "Automatically detect language" and the browser's language set to "ja"

(I guess "来週土曜" doesn't make sense in Chinese)

### Specifications
Mastodon version: v2.8.0
OS: macOS High Sierra
Browser: Firefox 66.0.2

I understand this behavior won't happen when the posting language is "日本語 (Japanese)".
I'd say, however, Mastodon should respect the browser's language since it seems not a few users leave their posting language as "Automatically detect language" and that the precision of the language detector is questionable.
(It's *highly* challenging to distinguish Japanese from Chinese automatically given a really short text, which is a common case for Mastodon)

Gargron:
Unfortunately I do not know how we can use browser language to influence language detection. CLD3 does not take a "hint". Simply assuming the browser's language is the same as the language a person is going to post in might lead to a lot of wrong results (is there any data on this?). One idea would be to use CLD3's `FindTopNMostFreqLangs` function to return 3 or 4 guesses and compare them against the browser's language, however, @akihikodaki's CLD3 gem does not currently expose that CLD3 function and I personally don't even know how to return a C array into Ruby so I can't modify it.

ykzts:
Resolved with #12124

PR #10653, opened 2019-04-29
Revert "Change language detector threshold from 140 characters to 4 words (#10376)"

carbon-twelve:
This reverts commit e6cfa7ab897ac4fd6bf9bbcafe09fc42c4cc2c5d.

For the background of this pull request, please take a look at issue #10578.
This is meant to be a palliative fix until we contrive a better solution.

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -183,14 +183,14 @@ export default class StatusContent extends React.PureComponent {
       return (
         <div className={classNames} ref={this.setRef} tabIndex='0' style={directionStyle} onMouseDown={this.handleMouseDown} onMouseUp={this.handleMouseUp}>
           <p style={{ marginBottom: hidden && status.get('mentions').isEmpty() ? '0px' : null }}>
-            <span dangerouslySetInnerHTML={spoilerContent} lang={status.get('language')} />
+            <span dangerouslySetInnerHTML={spoilerContent} />
             {' '}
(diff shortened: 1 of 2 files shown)


Gargron:
This is the wrong commit to revert. Japanese and Chinese languages are affected by the line below, `RELIABLE_CHARACTERS_RE`. It's not possible to revert with git automatically, because some of the follow-up commits are good.

carbon-twelve:
`RELIABLE_CHARACTERS_RE` is not a thing here because this regular expression just decides if an input text should be passed to the language detector or not.
The problem lies after that decision: the language detector is not reliable enough.

So, I'd suggest we stop showing the results from the detector for now (this pull request), and also remove the detector completely or exclude Japanese at least later.

carbon-twelve:
What kind of solution do you have in your mind instead?

Gargron:
- Remove Han character set from `RELIABLE_CHARACTERS_RE` or
- Enforce minimum number of characters even for "reliable" character sets

carbon-twelve:
It makes sense to me to remove Han character set as well as Katakana and Hiragana ones.
(which means disabling automatic language detection for Chinese and Japanese. Do I get it right?)

> Enforce minimum number of characters even for "reliable" character sets
I wouldn't go this way.
From my observations, it's not rare the language detector fails to classify toots correctly even if they contain Hiragana or they're more than hundred characters long.

Gargron:
> It makes sense to me to remove Han character set as well as Katakana and Hiragana ones.
> (which means disabling automatic language detection for Chinese and Japanese. Do I get it right?)

Han is shared between multiple languages, but are Katakana and Hiragana?

carbon-twelve:
I suggested removing them as containing those characters doesn't necessarily mean the results of the language detector is reliable.
I'd say such situation is common if not often.

ykzts:
Resolve with #12124

PR #10655, opened 2019-04-30
Explicitly disable storage of REST API results

### LINKED ISSUE #10652: Don’t show old toots when restoring a Mastodon tab
### Expected behaviour

Restoring a Mastodon tab loads the same toots that when you open a new Mastodon tab.

### Actual behaviour

The same tabs that were open before unloading the tab/closing the browser are displayed.

Use an extension to unload tabs/use «Restore my previous session» settings in your browser and restart it.

### Specifications

Mastodon 2.8.0

ClearlyClaire:
Fixes #10652

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -9,6 +9,8 @@ class Api::BaseController < ApplicationController
   skip_before_action :store_current_location
   skip_before_action :check_user_permissions
 
+  before_action :set_cache_headers
+
   protect_from_forgery with: :null_session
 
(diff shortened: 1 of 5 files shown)


ykzts (approved):

Gargron (changes requested):
Is the idea that you skip the headers on all API responses that use `render_cached_json`? Because in that case you've missed `/api/v1/instance` as well as `/api/v1/instance/activity`

ThibG:
Oh, yeah, indeed. Fixed that.

Gargron (approved):

ariasuni:
Thanks a lot, this is so much better now!

PR #10660, opened 2019-05-01
Record account suspend/silence time and keep track of domain blocks

ClearlyClaire:
Currently, undoing an instance block will either unblock *every* blocked person on that instance or none (depending on the moderator's choice), because an instance block translates to an individual block for every user of that instance.

This change records the time an user was blocked, and ties the unblock part to blocks performed *exactly* at the same time as the instance block.

~~This is a work in progress because we have to figure out what to do with instances that were blocked before that PR. The code in the PR will *not* unblock them, which is probably not what we want to do.~~ The proposed code will currently unblock anyone that matches the severity of the domain block and who was individually blocked *before* the migration was run.

~~Also need to add tests and make sure the interface is clear.~~


ThibG:
Updated. The unblock screen won't ask whether you want to unblock *all* existing accounts of that instance anymore, and will instead give the count of accounts that would actually be unblocked.

The accounts affected are those whose block date matches that of the domain block or are `nil`. That is, those who were definitely blocked through the domain block as well as those who were blocked before the feature was merged. There's a few things we could do slightly differently in that regard, but not much.

Gargron (commented):
We could remove the booleans

ThibG:
We definitely could, I just wanted to avoid a costly and error-prone migration script.



Le 1 mai 2019 17:55:22 GMT+02:00, Eugen Rochko <notifications@github.com> a écrit :
>
>We could remove the booleans
>
>
>
>-- 
>https://github.com/tootsuite/mastodon/pull/10660#pullrequestreview-232629983

-- 

nightpool:
I think the idea here is good but removing the boolean columns now is going to save us a ton of headaches down the road wrt our data model.

If you really really don't want to remove the booleans, then we should encapsulate any logic that refers to the booleans or the dates into the account model so that we only have to care about the legacy data model in one place.

ThibG:
I have changed the migration script to set the `silenced_at` and `suspended_at` fields of existing accounts to the date of the corresponding domain block if there is one, and to the current time otherwise.
I have changed the code to not use the `suspended` and `silenced` fields anymore and updated the tests accordingly.
Finally, I have added a post-deployment migration script to remove those columns from the database.

I have **not** tried the migration yet.

nightpool (changes requested):
First pass. I think we can narrow down the amount of files touched by this change, especially spec files as noted below. We should make sure we're using the `suspend!` and `silence!` methods in specs going forward.

spec/controllers/application_controller_spec.rb
```diff
@@ -182,13 +182,13 @@ def invalid_authenticity_token
     end
 
     it 'does nothing if user who signed in is not suspended' do
-      sign_in(Fabricate(:user, account: Fabricate(:account, suspended: false)))
+      sign_in(Fabricate(:user, account: Fabricate(:account, suspended_at: nil)))
```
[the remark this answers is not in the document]
ThibG: I've tried to do that but I'm not quite sure how to do that, or if it is even possible

app/models/account.rb
```diff
@@ -165,25 +165,35 @@ def refresh!
     ResolveAccountService.new.call(acct)
   end
 
-  def silence!
-    update!(silenced: true)
+  def silenced?
+    !silenced_at.nil?
```
nightpool: prefer `.present?` to `!thing.nil?`

app/models/account.rb
```diff
@@ -165,25 +165,35 @@ def refresh!
...
+  end
+
+  def silence!(date = nil)
+    date = Time.now.utc if date.nil?
```
nightpool: `date ||= Time.now`

app/models/account.rb
```diff
@@ -165,25 +165,35 @@ def refresh!
...
+    update!(silenced_at: date)
   end
 
   def unsilence!
-    update!(silenced: false)
+    update!(silenced_at: nil)
+  end
+
+  def suspended?
+    !suspended_at.nil?
```
nightpool: .present per above

app/models/account.rb
```diff
@@ -165,25 +165,35 @@ def refresh!
...
   end
 
-  def suspend!
+  def suspend!(date = nil)
+    date = Time.now.utc if date.nil?
```
nightpool: ||= per above

app/services/resolve_account_service.rb
```diff
@@ -118,10 +118,10 @@ def handle_activitypub
   def create_account
     Rails.logger.debug "Creating new remote account for #{@username}@#{@domain}"
 
-    @account = Account.new(username: @username, domain: @domain)
-    @account.suspended   = true if auto_suspend?
-    @account.silenced    = true if auto_silence?
-    @account.private_key = nil
+    @account              = Account.new(username: @username, domain: @domain)
```
nightpool: I don't agree with this whitespace change but it's also not a huge deal. this line is semantically different from the lines that follow, so they shouldn't be grouped together

nightpool (changes requested):
now that we can undo domain blocks precisely, is there any argument for leaving in the ability to undo them and NOT unsilence/unsuspend affected users? it seems like an extra level of complexity that doesn't help anyone

ThibG:
Removed the `retroactive` option as suggested, making *all* unblocks/unsilences retroactive.

This removes the checkbox but not the text:

image

spec/services/block_domain_service_spec.rb
```diff
@@ -1,20 +1,14 @@
...
 RSpec.describe BlockDomainService, type: :service do
-  let(:bad_account) { Fabricate(:account, username: 'badguy666', domain: 'evil.org') }
-  let(:bad_status1) { Fabricate(:status, account: bad_account, text: 'You suck') }
-  let(:bad_status2) { Fabricate(:status, account: bad_account, text: 'Hahaha') }
-  let(:bad_attachment) { Fabricate(:media_attachment, account: bad_account, status: bad_status2, file: attachment_fixture('attachment.jpg')) }
+  let!(:bad_account) { Fabricate(:account, username: 'badguy666', domain: 'evil.org') }
+  let!(:bad_status1) { Fabricate(:status, account: bad_account, text: 'You suck') }
+  let!(:bad_status2) { Fabricate(:status, account: bad_account, text: 'Hahaha') }
+  let!(:bad_attachment) { Fabricate(:media_attachment, account: bad_account, status: bad_status2, file: attachment_fixture('attachment.jpg')) }
+  let!(:already_banned_account) { Fabricate(:account, username: 'badguy', domain: 'evil.org', suspended: true, silenced_at: 1.day.ago) }
```
nightpool: both `suspended` and `silenced_at`?

nightpool (approved):
LGTM % a minor question about the tests

JamesABDL:
I think this has broken the explore page:

```
method=GET path=/explore format=html controller=DirectoriesController action=index status=500 error='ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column accounts.suspended does not exist
```

Gargron:
Need to tootctl cache clear

Issue #10608, opened 2019-04-20
Domain blocks cannot be upgraded


joyeusenoelle:
### Expected behaviour

As a server admin, when I go to /admin/domain_blocks/new, I should be able to *change* the severity of a domain block (i.e. from Silenced to Suspended or vice versa).

### Actual behaviour

As a server admin, when I go to /admin/domain_blocks/new, attempting to change the severity of a domain block returns an error, with the text under the **Domain** field "has already been taken".

I have to go back to the list of blocked domains (at /admin/instances?limited=1), find the domain in question, *undo* the existing domain block, and then set a new block severity at /admin/domain_blocks/new .

### Steps to reproduce the problem

1. Silence an instance through the interface.
2. Attempt to suspend that instance through the interface.

### Specifications

This occurs in v2.8.0.

ThibG:
Unfortunately, there are conceptual issues with downgrading (or undoing) an instance block: an instance block works by individually suspending every user of that instance, as well as making sure that any newly-discovered user of that instance gets blocked too. Undoing the block will stop marking newly-discovered as suspended, and optionally, will unsuspend all users of that instance (no matter *why* they were suspended in the first place). Downgrading a suspend to a silence would cause the same problem of either requiring unsuspending everyone no matter why they were suspended, or only affect newly-discovered accounts. While that issue is not the one described here, I think that's a blocker.

I am also not sure how to design the interface to avoid accidentally downgrading an active domain block. But I guess we could at least link to the active domain block in the error message to avoid one of the hoops?

joyeusenoelle:
The interface already asks if you want to unsuspend/unsilence all of the affected accounts. See the attached image. It wouldn't be a stretch to add "change suspension to silence for all affected accounts", and it would be one line of SQL to perform that change.

![Screenshot from 2019-05-01 12-25-13](https://user-images.githubusercontent.com/4139417/57028387-f8239680-6c0c-11e9-90ab-a83c178a248e.png)

Alternately, it's reasonable to suggest that a new column be added to the `users` table: `manual_admin_action` or something similar, to mark, if a user is suspended or silenced, whether it was done through the admin interface or because of a domain block. That would obviate the problem entirely; just add `where manual_admin_action = 'f'` to the SQL statement.

As for accidentally downgrading an active domain block, yes, I think amending the error message is the way to go.

ThibG:
[...]

PR #10663, opened 2019-05-01
Provide a link to existing domain block when trying to block an already-blocked domain

ClearlyClaire:
This adds a link to the error message in an attempt to improve the situation described by #10608 a bit.

image

### DIFF
--- a/app/models/domain_block.rb
+++ b/app/models/domain_block.rb
@@ -29,4 +29,11 @@ class DomainBlock < ApplicationRecord
   def self.blocked?(domain)
     where(domain: domain, severity: :suspend).exists?
   end
+
+  def stricter_than?(other_block)
+    return true if suspend?
+    return false if other_block.suspend? && (silence? || noop?)
+    return false if other_block.silence? && noop?
+    (reject_media || !other_block.reject_media) && (reject_reports || !other_block.reject_reports)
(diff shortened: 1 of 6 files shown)


Gargron:
Why not just make the submission replace the previous domain block?

ThibG:
As stated in the related issue, what I'm afraid of is an admin accidentally lowering the severity of a block that's already in effect. Also, we currently offer a choice to undo the corresponding individual blocks or not, and replacing the block would not allow us to do that.



Le 1 mai 2019 17:56:41 GMT+02:00, Eugen Rochko <notifications@github.com> a écrit :
>Why not just make the submission replace the previous domain block?
>
>-- 
>https://github.com/tootsuite/mastodon/pull/10663#issuecomment-488323553

-- 

app/models/domain_block.rb
```diff
@@ -29,4 +29,11 @@ class DomainBlock < ApplicationRecord
   def self.blocked?(domain)
     where(domain: domain, severity: :suspend).exists?
   end
+
+  def stricter_than?(other_block)
```
Gargron: You could overload the `>` method here and that would be fun, but probably best not to

Gargron (approved):

PR #10669, opened 2019-05-02
Add toot source to delete result to ease Delete & Redraft

ClearlyClaire:
Currently, a client has to parse the formatted status when doing Delete & Redraft.
This is not trivial, and even the WebUI does not do it correctly for mention to users with weird URIs.

This PR returns the status in the `delete` reply with an additional `raw_content` field containing exactly the text that was initially input.

### DIFF
--- a/app/controllers/api/v1/statuses_controller.rb
+++ b/app/controllers/api/v1/statuses_controller.rb
@@ -65,7 +65,7 @@ class Api::V1::StatusesController < Api::BaseController
 
     RemovalWorker.perform_async(@status.id)
 
-    render_empty
+    render json: @status, serializer: REST::StatusSerializer, source_requested: true
   end
 
   private
--- a/app/javascript/mastodon/actions/statuses.js
+++ b/app/javascript/mastodon/actions/statuses.js
@@ -131,10 +131,11 @@ export function fetchStatusFail(id, error, skipLoading) {
   };
 };
 
-export function redraft(status) {
+export function redraft(status, raw_text) {
   return {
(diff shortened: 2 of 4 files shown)


Gargron:
I am a little concerned/unsure about the naming and the semantics of adding stuff to entities--I am not a fan of the `source` property in the Account entity either. One option is to simply return the `content` unformatted from that particular response, without adding it into a new property. Since the previous response is empty, it would not mess with current applications...

Another option, instead of creating a `raw_content` property, is to return a `text` property *instead* of the `content` property, since the real column name *is* `text` after all.

ThibG:
I'm weary of directly returning the content because it may be an issue for us down the road if we want to add stuff later on. I think having the complete status *could* be useful for clients as well, depending on how they handle various things… but it probably isn't necessary.

If you prefer `text` instead of `raw_content`, that's an easy change I can do right now. I can also remove the `content` property in that particular case, but I'm not too sure why that would be necessary.

ThibG:
> I am a little concerned/unsure about the naming and the semantics of adding stuff to entities--I am not a fan of the `source` property in the Account entity either. One option is to simply return the `content` unformatted from that particular response, without adding it into a new property. Since the previous response is empty, it would not mess with current applications...

Oh, I think I misunderstood what you meant. Returning the full status *but* with the `contents` as plaintext rather than formatted? I think this would be more confusing and error-prone, the nature of the field changing depending on the context does not sit right with me.

Gargron (approved):

PR #10702, opened 2019-05-04
Make sure the instance banner is never cropped

elegaanz:
I don't know if there was a real reason for this limit to be here, but it seems it can be removed without breaking anything.

Example on eldritch.cafe:

| Before            | After        |
|-------------------|--------------|
| Screenshot_2019-05-04 Eldritch Café(2) | Screenshot_2019-05-04 Eldritch Café(1) |

### DIFF
--- a/app/javascript/styles/mastodon/widgets.scss
+++ b/app/javascript/styles/mastodon/widgets.scss
@@ -4,7 +4,6 @@
 
   &__img {
     width: 100%;
-    height: 167px;
     position: relative;
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


ClearlyClaire (approved):
LGTM

Gargron:
I believe the original reason for the height value was the profile page sidebar. Can you check if it still looks alright there?

ThibG:
The fixed height value isn't used in the sidebar, the CSS.
It is actually cropped there too, but in a different way. We probably could change that too.

PR #10711, opened 2019-05-05
Fix some colors of high contrast theme

nzws:
- Fix "nothing here" text color because it is hard to see.
- Fix counter border color to highlight-color.

**Before:**

image

**After:**

image


Gargron (approved):

yuzulabo:
I think test-webui test seems to be passed but it is not displayed on github.
https://circleci.com/gh/yuzulabo/mastodon/122

Should I do something?

PR #10715, opened 2019-05-06
Bring back crossed eye icon on sensitive media

tribela:
About account media gallery:

On 2.8.0, there were crossed eye icon on sensitive media
Screenshot from 2019-05-06 13-37-49

But since 2.8.1, crossed eye icon is removed due to blurhash, But old media don't have blurhash. So it look like this:
Screenshot from 2019-05-06 13-38-01
This looks very ugly, And user cannot know that image is just broken or sensitive image

So bring back that crossed eye image to tell "this is a sensitive image, you have to click to view image thumbnail" like this:
image


FoxSays:
Yeah this definitely should be fixed.

ykzts (approved):

ClearlyClaire (approved):

PR #10717, opened 2019-05-06
Add visibility icon back in media gallery

ClearlyClaire:
image

One possible issue is that it's still always white while we don't know the color of the background.


ThibG:
Oops, I did not see  #10715

PR #10718, opened 2019-05-06
Add setting to hide toots from bots

ClearlyClaire:


Gargron:
Hm, I do believe we had a PR that chose this approach before... (#8314), and I've said this there: Would it not make more sense to make it a request option like `only_media` instead of another account-wide preference?

ThibG:
Oh, I had not seen that PR. It would make sense… kinda… how would this combine with `only_media` for instance? And how would it work for notifications or the Home timeline?

Gargron:
Until there is a way for mods to mark an account as a bot forcibly, I don't think there is any benefit from blocking notifications from bots. The annoying bots would simply not mark themselves as bots. The not-annoying ones usually only respond to user input.

ThibG:
Well, I did write a bot, flagged as bot, which mentions people it is told to mention, unless they have “#nobot” in their bio (and if the poster meets other requirements specific to what that bot does)

ThibG:
But yeah, apart from that, it looks like a better way to do it, more flexible and easier to implement. I will do that.

ThibG:
I was looking at implementing it how you've described, but there are several hurdles:
- the home timeline controller currently does not accept any such option
- the query string is not used for websocket, it subscribes to a different feed instead, so it's not clear how that would combine. I guess we could keep that as a query parameter in both cases, but it's a bit inconsistent

ThibG:
Ok I reworked the PR to handle a `no_bots` query parameter for public timelines, tag timelines, Home timelines, and all streaming endpoints.

This is currently untested and only a backend change. There are disparities between the tag/public timelines endpoints and the streaming server, similar to how they behave for mentioned blocked users:
- public timelines and tag timelines endpoints will filter out all bot toots regardless of whether the current user follows them, while the streaming server will only filter toots from bots that the current user does not follow
- public timelines and tag timelines endpoints will not handle boosts properly, but that's only an issue for some forks who show boosts in public timelines

I'm a bit dissatisfied with those disparities and the different places the filtering is applied…

ThibG:
Implemented it in the UI as a toggle on each column (public, local, home, and tags)

image

Changing the setting reloads the timeline and disconnects/reconnects but without changing identifiers. I changed the `online` property of timeline ids to be refcounting, which also solves a corner case occurring when pinning some columns.

I am not 100% sure about that design but it appears to work. Reviews welcome.

app/controllers/api/v1/timelines/home_controller.rb
```diff
@@ -44,7 +48,7 @@ def insert_pagination_headers
   end
 
   def pagination_params(core_params)
-    params.slice(:local, :limit).permit(:local, :limit).merge(core_params)
+    params.slice(:local, :limit, :no_bots).permit(:local, :limit, :no_bots).merge(core_params)
```
Gargron: Minor nitpick but if we have "exclude_replies" elsewhere it should be "exclude_bots" as well

Gargron (changes requested):
Looks correct aside from param name choice

app/models/status.rb
```diff
@@ -88,6 +88,7 @@ class Status < ApplicationRecord
   scope :including_silenced_accounts, -> { left_outer_joins(:account).where.not(accounts: { silenced_at: nil }) }
   scope :not_excluded_by_account, ->(account) { where.not(account_id: account.excluded_from_timeline_account_ids) }
   scope :not_domain_blocked_by_account, ->(account) { account.excluded_from_timeline_domains.blank? ? left_outer_joins(:account) : left_outer_joins(:account).where('accounts.domain IS NULL OR accounts.domain NOT IN (?)', account.excluded_from_timeline_domains) }
+  scope :not_from_bot, -> { left_outer_joins(:account).where.not('accounts.actor_type': %w(Application Service)) }
```
Gargron: Potentially slow query due to missing index

app/controllers/api/v1/timelines/home_controller.rb
```diff
@@ -19,7 +19,11 @@ def show
   private
 
   def load_statuses
-    cached_home_statuses
+    if truthy_param?(:exclude_bots)
+      cached_home_statuses.reject { |status| status.reblog? && status.reblog.account.bot && !current_account.following?(status.reblog.account.bot) }
```
Gargron: Can return empty page, resulting in early "end" of the home timeline

Gargron (approved):

ThibG:
I think this would still be useful, and the only reason it is stale is that it hasn't been merged despite being approved…

umonaca:
This feature is not as well-implemented as the `media only` filter. 
I know one instance (https://moe.cat) has merged the commits. However, they are not as responsive as the `media only` filter.
For `media only` filter, if the user chooses to hide media, the images disappered. Then if the user chooses to show media, It refreshes the timeline.
However, for `exclude bot` filter, the changes to the settings are not shown immediately. And many times users have to refresh the entire page for the change of setting to take efffect.

umonaca:
Update: tried cherry-pick in a local development environment. Does not work...

ThibG:
You need to restart the streaming server as well. Will check about the retroactive effect of the setting. Also will rebase and look into the issue you just reported.

ThibG:
Rebased and updated. There was an issue where it wasn't properly reloading public or community timelines when using the toggle, and another which made it not apply to the community timeline when polling (but should have applied to streaming).
The React warning shouldn't have caused any issue, but I fixed it too.

umonaca:
Thank you for your quick response. However, it still does not work in a local test environment.

### Home timeline: 
One bug and one suggestion.

#### Bug: 
Steps to reproduce:
1. In the home timeline, toggle on to hide. Nothing happens.
2. Then refresh the entire page. The option is still turned off.

#### Suggestion:
Some users, including myself are following bot accounts, but still want to have an option to turn on and off when they want to skip the toots from bots. The "clear timeline" function should also be applied to the home timeline as well.

### Local timeline and community timeline

#### Bug:
Steps to reproduce:
1. In a local development environment, login as the local admin. Randomly post something. You probably also want to set the site setting.
2. In terminal, run the following commands to add two user accounts:
```
bin/tootctl accounts create userA --email=usera@localhost --confirmed
bin/tootctl accounts create userBot --email=userbot@localhost --confirmed
```
3. Login each account from different browsers, or just use different Firefox Containers to login. Set userBot to a bot account from the web interface. Keep userA as a normal user. Randomly post something from all accounts.
4. Login as the local admin (admin@localhost:3000). In the local or federated timeline, toggle on to hide. 
5. Then all toots from everyone disappear, except the toots from the current user (admin@localhost:3000). However, what should happen is that only toots from userBot disappear.

Please consider fixing the bugs above. Thank you!

FYI:
I think this is a good feature, because some users from our instance are following bot actors which bridge RSS to Activitypub, Twitter to Activitypub, etc.

app/models/status.rb
```diff
@@ -90,6 +90,7 @@ class Status < ApplicationRecord
...
```
umonaca: When an account is created by using command line:
```
bin/tootctl accounts create normal --email=normal@localhost --confirmed
```
The `actor_type` will be `[null]` is stead of `Person`, `Service` or `Application`.

umonaca: You probably want something like the following:
```
{ left_outer_joins(:account).where('accounts.actor_type IS NULL OR accounts.actor_type NOT IN (?)', %w(Application Service)) }
```

umonaca:
Update:
Sometimes there is a weird bug like this: the hide-bot option is always turned on after a entire page refresh. I don't know how to reliably reproduce this bug, but it does happen in a local dev environment.

ThibG:
For the change setting+refresh “bug”, the settings are stored server-side, but the updated settings aren't uploaded as soon as you change them, there is a delay to avoid making too many requests. This isn't specific to that setting.

app/javascript/mastodon/features/community_timeline/index.js
```diff
@@ -82,6 +82,12 @@ class CommunityTimeline extends React.PureComponent {
       const { dispatch, onlyMedia, excludeBots } = this.props;
 
       this.disconnect();
+
+      if (prevProps.excludeBots !== excludeBots) {
+        dispatch(clearTimeline('community'));
+        dispatch(clearTimeline('community:media'));
```
umonaca: It seems that `community:media` is not found anywhere else in the source code?

ThibG:
community:media exists, but the string is built dynamically. See https://github.com/ThibG/mastodon/blob/dd467ba98338e5a49a40529fea50e8ed9738f4ca/app/javascript/mastodon/features/community_timeline/index.js#L135 for instance

app/models/status.rb
mason1900: Update:
Tried using the line that I provided here. It fixed "Local timeline and community timeline" bug that I reported. Home timeline bug remains unfixed.

umonaca:
Thank you for your reply.

To sum up:

### Home timeline
The home timeline filter does not work in a local dev environment. It won't filter posts from a bot account.

### Suggested changes
```
scope :not_from_bot, -> { left_outer_joins(:account).where('accounts.actor_type IS NULL OR accounts.actor_type NOT IN (?)', %w(Application Service)) }
```
It works in a local dev environment.
Also, maybe  "clear timeline" function should be applied to the home timeline as well.

### Resolved:
1. change setting+refresh “bug”
2. `community:media`

ThibG:
wrt. the Home timeline, it is supposed to filter bots *you do not follow* (that is, boosted toots from bots you don't follow), I guess that is confusing?

The actor type thing is interesting, I wonder if we should make sure this is set. Thanks for your suggested fix!

umonaca:
> wrt. the Home timeline, it is supposed to filter bots _you do not follow_ (that is, boosted toots from bots you don't follow), I guess that is confusing?

Thanks for your explanation. I was thinking something different.

Some users on our instance follow RSS to Activitypub news bots and image bots which cross-post or upload media files on a regular basis. They get their home timeline filled with automatic posts, which are mixed together with their friends' posts. However, the Mastodon "list" is not convenient as an alternative approach because the user has to *follow an account* in order to have that account in an list. 
Then users have to either leave their home timelines in an unusable state (and use lists instead), or unfollow all the bot accounts.

So I am suggesting that there are good reasons to filter out the posts from *the bots that the user follows*.

ThibG:
I guess lists could become a viable alternative to the Home field once we can add ourselves to them (which is the case in master). Another option some forks have adopted is to have “exclusive” lists which accounts do not appear in your home TL.

umonaca:
I think it makes more sense to do it in the reverse way: keep our friends in the home timeline and move the RSS news bots to the lists.
I don't know about the general user in the Fediverse. On our instance, many use the reply toggle to hide replies that they are not interested in; few people use the hide-boost toggle; I don't think anyone would use a toggle to "hide boosts from bots that they do not know".

umonaca:
New bug report:
### Local timeline and community timeline
An inconsistency in the filtering policy of local and community timelines.
As you have written in the posts above, currently hiding bots in local and community timelines will filter out all bots no matter the current account follows them or not. However, it is not true. After toggle on hiding, all bot posts disappear, but later, *new bot posts* from the bot accounts that the current account follows will continue to pop out.

#### Steps to reproduce:
1. In a local dev environment, create a bot account. Randomly post something from the bot account.
2. Use another account (suppose it is `admin`), toggle on hiding in the local timelime. The bot posts disappear.
3. Switch to the bot account in another browser, randomly post something.
4. Switch to the `admin` account browser window. You will see the *new* bot posts, although the switch for home timeline is still on.

#### Expected behaviour
The new bot posts should not appear in the local timeline.

umonaca:
image
The screenshot above shows what happens after turning on the home timeline switch. After a few seconds, there are inconsistent network activities. Sometimes `excludeBots` is true, and other times it is false.

umonaca:
Update:

Set the initial state of the toggle to be **off**. Wait for 5 secs and Refresh the entire page. Wait for a while then toggle **on** the home timeline filter.

I am not familiar with React, but by adding lots of `console.log()` in the source code, I got the following conclusion:

`refreshHomeTimelineAndNotification` is constantly being called with `false` value of `excludeBots` even the home timeline filter has been toggled on. `connectUserStream` is not constantly triggered by a component update, but it may has something to do with  `store.dispatch(connectUserStream({ excludeBots }))` in the `app/javascript/mastodon/containers/mastodon.js`. 

I guess the initial value of  `excludeBots` is somehow stored before and it won't change when the user toggle the switch. This explains why there are mixed type of network activities with both `excludeBots = false` and `excludeBots = true`.

umonaca:
Update:

I have seen [this pull request](https://github.com/tootsuite/mastodon/pull/11272) which seems to be very effective and bug-free on the hometime line. It consists of Javascript only, and only changes 4 files (other commits are just translations).

umonaca:
Update:
The refreshHomeTimelineAndNotification(excludeBots) is sent to `connectTimelineStream` as the `pollingRefresh`. However, when the switch is toggled in the home timeline, the previous polling with previous state of `excludeBots` does not stop. This is the reason for the home timeline bug that I reported.

PR #10721, opened 2019-05-07
Remove custom emojis on "tootctl domains purge"

tribela:
There is no reason to leaving custom emojis from purged instance.
Additionally, `DomainBlock` should not be removed on dry run

### DIFF
--- a/lib/mastodon/domains_cli.rb
+++ b/lib/mastodon/domains_cli.rb
@@ -34,8 +34,9 @@ module Mastodon
       say("Removed #{removed} accounts#{dry_run}", :green)
 
       custom_emojis = CustomEmoji.where(domain: domain)
-      say("Removing #{custom_emojis.count} custom emojis", :green)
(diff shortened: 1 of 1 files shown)


ClearlyClaire (approved):

lib/mastodon/domains_cli.rb
```diff
@@ -28,10 +28,14 @@ def purge(domain)
...
       end
 
-      DomainBlock.where(domain: domain).destroy_all
+      DomainBlock.where(domain: domain).destroy_all unless options[:dry_run]
 
       say
       say("Removed #{removed} accounts#{dry_run}", :green)
+
+      custom_emojis = CustomEmoji.where(domain: domain)
+      say("Removing #{custom_emojis.count} custom emojis", :green)
```
Gargron: The text above says "Removed x items" but this text says "Removing x items", I think it should be consistent?

PR #10723, opened 2019-05-07
Toot privacy language

robinmaypanpan:
The language around the differences between the different post privacy options can leave some users, especially new users, confused.  These changes to the privacy option descriptions make it clearer what the differences are between them.  Note that these are suggestions I have implemented on my instance and may not reflect the current direction of the project. Feel free to reject this PR if you don't like the suggested changes!

Rationale:
- Using the words "Broadcast" and "all" help suggest that Public posts may go to many different places.
- Explicitly mention followers and personal profile for Unlisted posts to ensure the user understands where they go.
- Use the word "Restrict" with Followers-only messages to ensure they are understood as limited. Also explicitly say "no boosting" to make it clear how this differs from Unlisted and Public posts.
- Use the word "Private message" with Direct to make it clear that these are for private conversations with other users and not "posts" in the same sense as the other privacy settings.


trwnh:
>Use the word "Private message" with Direct to make it clear that these are for private conversations with other users and not "posts" in the same sense as the other privacy settings.

This is incorrect. Direct posts are not private, and they *are* regular posts in the exact same sense as any other post. They are simply not addressed to the `followers` collection or to the `public` pseudo-object. With "Visible to mentioned users only", the "only" is important.

In any case, I think "post to" is more semantically correct, since the server literally POSTs your status creation activity to the actor inbox or sharedInbox. I'm not sure "restrict" and "broadcast" are appropriate or even accurate, although "restrict" is the less accurate of the two -- you can boost followers-only posts to the original audience.

However, it is a good idea to explain "unlisted" more clearly, so as to make it clear that an unlisted post is still "visible on your profile", and not "accessibly only by direct URL" as it is on e.g. YouTube.

joycem137:
> This is incorrect. Direct posts are not private, and they _are_ regular posts in the exact same sense as any other post. They are simply not addressed to the `followers` collection or to the `public` pseudo-object. With "Visible to mentioned users only", the "only" is important.

While this may be true, this isn't necessarily how end users relate to this concept.  The idea behind my language changes is to improve the user understanding of what these various settings do.  The current text does not convey the way they are used as effectively as it could, at least to new users, in my opinion.

> In any case, I think "post to" is more semantically correct, since the server literally POSTs your status creation activity to the actor inbox or sharedInbox. I'm not sure "restrict" and "broadcast" are appropriate or even accurate, although "restrict" is the less accurate of the two -- you can boost followers-only posts to the original audience.

I did a non-scientific poll around the word "Broadcast" for public posts and people seemed to really prefer that language, in their attempt to understand the difference between public, unlisted, and followers only.

Agree on restricted, but not sure how to improve on that myself.

> However, it is a good idea to explain "unlisted" more clearly, so as to make it clear that an unlisted post is still "visible on your profile", and not "accessibly only by direct URL" as it is on e.g. YouTube.

Agree.

Issue #10731, opened 2019-05-08
Moderation: deleted status should not be re-appeared


Kjwon15:
### Expected behaviour

Deleted(Redacted) status by moderated should not be re-appeared on timeline.

Maybe fixed with adding object uri to DB something like tombstone.

### Actual behaviour

Deleted status can be re-fetched via search API, boost

### Steps to reproduce the problem

- Post a status from another instance
- Click "Open this status in moderation UI"
- Delete that status
- Paste original status URI on search bar
- That status re-appeared

PR #10732, opened 2019-05-08
Record deleted(by mod) status to prevent re-appear

tribela:
Related: #10731 

Like Tombstone, deleted status by moderation should not be re-appeared on timeline.


ClearlyClaire (changes requested):
`db/schema.rb` is missing the new table, and `app/models/redacted_status.rb` is missing the annotations.

Also, I wonder if it wouldn't be better to add a flag in the `tombstones` table instead, to perform only one lookup instead of two.

Kjwon15:
Thanks for comment. I agree that using tombstone instead of making another table is performance better.
But that status is not actually deleted. So I am curious it is fine to add to Tombstone.

ThibG:
Yes, I think I'd add a field to `Tombstone` to mark that it was manually discarded by an admin rather than actually deleted.

Kjwon15:
Oh that's seems good.
Anyway, I am not rails expert. Can you tell me how to add annotation comments?

ThibG:
They are added when running the migration in the development environment. Usually I add them manually 🤷

Kjwon15:
Thanks!
looks fine now.

db/schema.rb
```diff
@@ -591,8 +591,8 @@
   create_table "status_pins", force: :cascade do |t|
     t.bigint "account_id", null: false
     t.bigint "status_id", null: false
-    t.datetime "created_at", default: -> { "now()" }, null: false
-    t.datetime "updated_at", default: -> { "now()" }, null: false
+    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "updated_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.index ["account_id", "status_id"], name: "index_status_pins_on_account_id_and_status_id", unique: true
```
ThibG: I am not sure why these changes were made?

ClearlyClaire (changes requested):
Also, it looks like the migration script is missing, now.

db/schema.rb
Kjwon15: IDK too.. But db:migrate makes that.

ClearlyClaire (approved):

PR #10734, opened 2019-05-09
Add strings for Slovak

marek-lach:


config/locales/sk.yml
```diff
@@ -340,18 +341,20 @@ sk:
         expired: Vypršalo
         title: Filtruj
       title: Pozvánky
+          pending_accounts:
```
ThibG: This line is incorrectly indented, it should be at the same level as `relays:` a few lines later!

Gargron (approved):

PR #10748, opened 2019-05-10
Change icon and label depending on whether media is marked as sensitive

ClearlyClaire:
Maybe a proper checkbox would be better though

## When not marked as sensitive

Screenshot_2019-05-10 Dev instance(1)

## When marked as sensitive

image


connyduck:
I think it should still be blue when set to sensitive. Also a little more space between text and icon would look better imho.

Gargron:
We don't change "CW" to "Undo CW". I think this might be very confusing. Maybe a checkbox would be better, the eye icon doesn't look very good anyway.

ThibG:
> We don't change "CW" to "Undo CW".

Maybe we should.

> I think this might be very confusing. Maybe a checkbox would be better, the eye icon doesn't look very good anyway.
> 
> Edit: If it's a checkbox the label should definitely not change

Yeah, a checkbox would probably be better, I'll try to do that.

ThibG:
What about this?

image

image

Gargron (approved):

PR #10779, opened 2019-05-18
add og:image:alt for media attachments in embeds

BenLubar:


app/views/stream_entries/_og_image.html.haml
```diff
@@ -7,6 +7,7 @@
       - unless media.file.meta.nil?
         = opengraph 'og:image:width', media.file.meta.dig('original', 'width')
         = opengraph 'og:image:height', media.file.meta.dig('original', 'height')
+      = opengraph 'og:image:alt', media.description
```
ykzts:
```suggestion
      - if media.description.present?
        = opengraph 'og:image:alt', media.description
```

I think that it is not necessary to append when the description is empty.

Gargron (approved):

PR #10783, opened 2019-05-19
Add post-deployment migration script to delete public-boosts-of-private boosts

ClearlyClaire:
~~I have *not* tried the migration script yet.~~

### DIFF
--- a/db/post_migrate/20190519130537_remove_boosts_widening_audience.rb
+++ b/db/post_migrate/20190519130537_remove_boosts_widening_audience.rb
@@ -1,0 +1,23 @@
+class RemoveBoostsWideningAudience < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    public_boosts = Status.find_by_sql(<<-SQL)
+      SELECT boost.id
+      FROM statuses AS boost
+      LEFT JOIN statuses AS boosted ON boost.reblog_of_id = boosted.id
+      WHERE
+        boost.id > 101746055577600000
+        AND (boost.local = TRUE OR boost.uri IS NULL)
+        AND boost.visibility IN (0, 1)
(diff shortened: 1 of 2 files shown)


db/post_migrate/20190519130537_remove_boosts_widening_audience.rb
```diff
@@ -0,0 +1,23 @@
+class RemoveBoostsWideningAudience < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    public_boosts = Status.find_by_sql(<<-SQL)
+      SELECT boost.id, boost.visibility, boost.reblog_of_id
```
Gargron: Nothing except `id` is actually used

db/post_migrate/20190519130537_remove_boosts_widening_audience.rb
```diff
@@ -0,0 +1,23 @@
...
+      LEFT JOIN statuses AS boosted ON boost.reblog_of_id = boosted.id
+      WHERE
+        boost.id > 101746055577600000
+        AND (boost.local = TRUE OR boost.uri IS NULL)
+        AND boost.visibility IN (0, 1)
+        AND boost.reblog_of_id IS NOT NULL
+        AND boosted.visibility = 2
+    SQL
+
+    RemovalWorker.push_bulk(public_boosts.pluck(:id))
```
Gargron: `find_by_sql` already loads full models, it does not return a relation, so `pluck` is equivalent to `map(&:id)` here

ThibG: Hm, ok, parts of the Mastodon code do that though. Is there something lighter we could do? We only need the id.

Gargron (approved):

PR #10785, opened 2019-05-19
fix `isSubmitting` prop case

hinaloe:
in #10613, compose_form's props' cases were changed to camel case. but, `isSubmitting` prop is not changed (looks like it was forgotten).

This was to allow duplicate postings.


Gargron:
> This was to allow duplicate postings.

I wonder why `Idempotency-Key` header did not prevent them. Race condition if two requests are processed at the same time by different threads?

hinaloe:
I don't know in detail, but it seems to be parallelized before the save is over.

Gargron (approved):

PR #10790, opened 2019-05-19
Change bio limit from 160 to 500

### LINKED ISSUE #10628: Bio limits too restrictive
### Rationale
100-ish characters is a bit too low for conveying more than a short sentence or two. One can use profile fields to fit more information, but only for non-essential supplementary things, because profile fields aren't as widely supported in alternative implementations.

### Pitch
Slightly increase the default character limit on bios. Maybe make it the same as statuses (500?) But IMO at least 420.

trwnh:
Fix #10628 

Just following https://github.com/tootsuite/mastodon/issues/10628#issuecomment-493780569

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -75,7 +75,7 @@ class Account < ApplicationRecord
   validates_with UniqueUsernameValidator, if: -> { local? && will_save_change_to_username? }
   validates_with UnreservedUsernameValidator, if: -> { local? && will_save_change_to_username? }
   validates :display_name, length: { maximum: 30 }, if: -> { local? && will_save_change_to_display_name? }
-  validates :note, note_length: { maximum: 160 }, if: -> { local? && will_save_change_to_note? }
+  validates :note, note_length: { maximum: 500 }, if: -> { local? && will_save_change_to_note? }
   validates :fields, length: { maximum: 4 }, if: -> { local? && will_save_change_to_fields? }
 
   scope :remote, -> { where.not(domain: nil) }
--- a/app/views/settings/profiles/show.html.haml
+++ b/app/views/settings/profiles/show.html.haml
@@ -7,7 +7,7 @@
   .fields-row
     .fields-row__column.fields-group.fields-row__column-6
       = f.input :display_name, wrapper: :with_label, input_html: { maxlength: 30 }, hint: false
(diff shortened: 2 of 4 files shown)


trwnh:
Looks like CI is failing because some of the tests expect longer than 160 to fail, but now they succeed -- not entirely sure how to fix that in CI.

Gargron (approved):

Gargron:
You need to adjust tests in:

- `spec/controllers/api/v1/accounts/credentials_controller_spec.rb:68`
- `spec/models/account_spec.rb:607`

trwnh:
looks like just the account_spec needed to be updated from 161 to 501. the api one just asserts that an unprocessable status returns an error instead of 200 ok. i'll wait for ci to recheck but hopefully that should do it

Gargron:
No, sorry, it still needs an adjustment in the other file.

trwnh:
I'm not sure how to fix the other file. Maybe it's a RUby syntax thing I'm not understanding, but it looks logically correct to me.

CI gives this:
```
1) Api::V1::Accounts::CredentialsController with an oauth token PATCH #update with invalid data returns http unprocessable entity
     Failure/Error: expect(response).to have_http_status(:unprocessable_entity)
       expected the response to have status code :unprocessable_entity (422) but it was :ok (200)
     # ./spec/controllers/api/v1/accounts/credentials_controller_spec.rb:68:in `block (5 levels) in <top (required)>'
```

Context around `./spec/controllers/api/v1/accounts/credentials_controller_spec.rb:68`:
```
      describe 'with invalid data' do
        before do
          patch :update, params: { note: 'This is too long. ' * 10 }
        end

        it 'returns http unprocessable entity' do
          expect(response).to have_http_status(:unprocessable_entity)
        end
      end
```

What should it be changed to? It seems to me like it's describing a test case that returns "unprocessable" and expecting it to have the "unprocessable" status code.

Gargron:
The test case was expecting an error, but it is getting success instead, because of the raised limit. `'This is too long. ' * 10` repeats the string 10 times. Which is below 500 characters.

trwnh:
Ahh, OK. I had no idea what exactly the *10 was doing, especially since the other test used a much less ambiguous `Fabricate.build(:account, note: Faker::Lorem.characters(501))`. Changed 10 to 30 so that the resulting string is 540 characters long.

trwnh:
OK, looks like it passes all tests now!

PR #10796, opened 2019-05-20
Add from account to notifications api

pwoolcoc:
This adds the ability to filter notifications by the account they originated from. Currently there is no UI for this, it only adds the ability to the API.

### DIFF
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -82,7 +82,7 @@ class Api::V1::NotificationsController < Api::BaseController
   end
 
   def from_account
-    params[:from_account]
+    params[:account_id]
   end
 
   def pagination_params(core_params)
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -44,7 +44,7 @@ class Api::V1::NotificationsController < Api::BaseController
   end
 
   def browserable_account_notifications
-    current_account.notifications.browserable(exclude_types)
+    current_account.notifications.browserable(exclude_types, from_account)
   end
 
   def target_statuses_from_notifications
@@ -79,6 +79,10 @@ class Api::V1::NotificationsController < Api::BaseController
     val = params.permit(exclude_types: [])[:exclude_types] || []
     val = [val] unless val.is_a?(Enumerable)
     val
+  end
+
+  def from_account
+      params[:from_account]
   end
 
   def pagination_params(core_params)
--- a/app/controllers/api/v1/notifications_controller.rb
+++ b/app/controllers/api/v1/notifications_controller.rb
@@ -82,7 +82,7 @@ class Api::V1::NotificationsController < Api::BaseController
   end
 
   def from_account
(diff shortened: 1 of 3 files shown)


app/models/notification.rb
```diff
@@ -41,9 +41,17 @@ class Notification < ApplicationRecord
...
   validates :activity_type, inclusion: { in: TYPE_CLASS_MAP.values }
 
-  scope :browserable, ->(exclude_types = []) {
+  scope :browserable, ->(exclude_types = [], from_account = nil) {
     types = TYPE_CLASS_MAP.values - activity_types_from_types(exclude_types + [:follow_request])
-    where(activity_type: types)
+    return where(activity_type: types) if from_account.nil?
+
+    username, domain = from_account.split('@')
+    account = if TagManager.instance.local_domain?(domain)
```
Gargron: I think you should definitely not reference the account by username/domain here and instead use the ID, that would simplify this scope significantly.

app/controllers/api/v1/notifications_controller.rb
```diff
@@ -81,6 +81,10 @@ def exclude_types
     val
   end
 
+  def from_account
+    params[:from_account]
```
Gargron: While the internal name is `from_account_id`, perhaps it would make more sense to name the query parameter simply `account_id`, since the "from" seems to be redundant in the context of how this will be used.

Gargron (changes requested):
I like the idea, just need some details changed

app/models/notification.rb
pwoolcoc: done

app/controllers/api/v1/notifications_controller.rb
pwoolcoc: done

app/models/notification.rb
```diff
@@ -41,9 +41,11 @@ class Notification < ApplicationRecord
   validates :account_id, uniqueness: { scope: [:activity_type, :activity_id] }
   validates :activity_type, inclusion: { in: TYPE_CLASS_MAP.values }
 
-  scope :browserable, ->(exclude_types = []) {
+  scope :browserable, ->(exclude_types = [], account_id = nil) {
     types = TYPE_CLASS_MAP.values - activity_types_from_types(exclude_types + [:follow_request])
-    where(activity_type: types)
+    return where(activity_type: types) if account_id.nil?
```
Gargron: This is a lambda, not a method, so the `return` keyword acts differently here. I suggest just an `if/else` block. Or you can change the scope to a class-level method:

```ruby
class << self
  def browserable(exclude_types = [], account_id = nil)
    types = TYPE_CLASS_MAP.values - activity_types_from_types(exclude_types + [:follow_request])
    return where(activity_type: types) if account_id.nil?
    where(activity_type: types, from_account_id: account_id)
  end
end
```

Gargron (changes requested):

app/models/notification.rb
pwoolcoc: done, though I thought `return` acted differently in procs not lambdas

Gargron (approved):

PR #10813, opened 2019-05-23
Move signature verification stoplight to the requests themselves

ClearlyClaire:
This avoids blocking messages from known keys for 5 minutes when only one fails…


ThibG:
Also, I would like to return something else than `401` in most cases so that remote instances retry without the need for #10812. But I'm not too sure what to return.

ThibG:
Also put the stoplight on the actual client address and not any eventual trusted reverse-proxy… thanks @0xa !

Gargron (approved):

PR #10685, opened 2019-05-03
Change poll options to alphabetic letters when status text is hidden

Gargron:
Fix #10569

PR #10817, opened 2019-05-24
Fix poll visibility on public pages

noellabo:
Fixed an issue where polls on public pages are all alphabetic letters.


Gargron:
I think maybe it is better to revert #10685? What do you think?

noellabo:
As you say,  #10685  doesn't seem to be appropriate.
I also think that revert is appropriate.

Gargron:
Can you include the revert in this PR? Thank you

Gargron (approved):

PR #10822, opened 2019-05-25
Docker: Remove dist-upgrade

SuperSandro2000:
Upgrading all packages is an anti pattern in docker. If there is no special reason to do that you should not do it. If any specific package is required to be update that individual package should be targeted.
The problem with that is that we potentially update many packages that increase the image size drastically and the end users needs to download more then necessary.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ SHELL ["bash", "-c"]
 ENV NODE_VER="8.15.0"
 RUN	echo "Etc/UTC" > /etc/localtime && \
 	apt update && \
-	apt -y dist-upgrade && \
 	apt -y install wget make gcc g++ python && \
 	cd ~ && \
 	wget https://nodejs.org/download/release/v$NODE_VER/node-v$NODE_VER.tar.gz && \
@@ -80,7 +79,6 @@ ARG GID=991
 RUN apt update && \
 	echo "Etc/UTC" > /etc/localtime && \
 	ln -s /opt/jemalloc/lib/* /usr/lib/ && \
-	apt -y dist-upgrade && \
 	apt install -y whois wget && \
 	addgroup --gid $GID mastodon && \
 	useradd -m -u $UID -g $GID -d /opt/mastodon mastodon && \


ykzts (approved):

Miaourt:
Just my humble opinion, but updating existent package aren't increasing the size by much (worst case a few kilobytes), don't take much time, and provide latest upstream security fixes.
I think this should be reverted, but replacing dist-upgrade with only upgrade might be interesting, since the first one isn't really pertinent in our objective of having only security issues patched.

SuperSandro2000:
Well, wrong. It increases the image by the package size which best case is 1MB and worst case hundreds of MBs. Docker updates every file apt update touches and puts them in a new layer.
The build time is already insane so every saving is appreciated.
The Ubuntu image is almost daily updated with the latest packages so there is no need to do this if the image is build on CI or the builder pulls the ubuntu image before building.
> with only upgrade might be interesting

Well, no. If you care that much about up to date run Arch and bare metal not docker.
Also this is against docker best practices which you should follow and only break if you have a valid reason to do so. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

If you're objective is to get this damn image to 2 GB you are on the best track. 1.4GB is already way to much for a docker image.

Miaourt:
It's not having bleeding edge stuff, it's having less security issue, docker is already a nightmare security-wise so it's quite important to do our best to have the very latest fixes.
The speed of built isn't the objective if this image. Faster the better, of course, but not on the detrimental of the security.
The very big size of this image is mostly due to the relatively big size of the mastodon software, relying on a bunch of gems and node packages, in addition to building all the styling at build time into the image for easier utilization, it's not a rush to save every bits can.

Also please try to stay courteous, we aren't here to attack each others.

SuperSandro2000:
> docker is already a nightmare security-wise 
Most docker container that are well build are way more up do date then the typical Ubuntu or Debian machine. Also you give an potential attacker way less attack surface out of the box and don't allow him to sneak into all production databases when he finds one rce. So it is more secure if you know what you are doing like always.

>The speed of built isn't the objective if this image.
It took bloody 47 minutes in circleci.....

>Faster the better, of course, but not on the detrimental of the security.
If there is an important security update it can always be installed via ``apt install``.
Also if the official Ubuntu base image is not up to date and there is an important update just nag the guys over here https://github.com/docker-library/official-images to update it asap.

>it's not a rush to save every bits can.
Doesn't hurt. Speeds up my dev cycle.

>Also please try to stay courteous, we aren't here to attack each others.
Please don't suggest anti patterns then.

Also apt-upgrade produces unpredictable builds and you can't reproduce an image in the future locally without pulling all updates. Good luck debugging a package issue with that.

Shleeble:
`docker-compose build --pull`

problem solved.

SuperSandro2000:
@Shleeble good joke. One of the most request features that one if the maintainers simply does not want to implement.

Sir-Boops:
I know this is thread is dead but @SuperSandro2000 

> If you're objective is to get this damn image to 2 GB you are on the best track. 1.4GB is already way to much for a docker image.

Is because the default docker build system is trash see [these](https://cloud.docker.com/repository/docker/sirboops/mastodon/tags) for what you can do using this dockerfile

> Docker updates every file apt update touches and puts them in a new layer.

See --squash

> It took bloody 47 minutes in circleci.....

That's because it builds nodejs, ruby, jemalloc, mastodon then finally compiles assets not from updates this was a choice made when this new dockerfile was added

SuperSandro2000:
The docker build system is not trash this just uses it wrong. I don't wanna use squash.

Why even compile nodejs and jemalloc? Why not use preconpiled versions?

Sir-Boops:
>  I don't wanna use squash.

What's wrong with squash?

> Why even compile nodejs and jemalloc? Why not use preconpiled versions?

Allows masto to use any version at any time and this way the docker image dosn't depend on too many other images :smile:

SuperSandro2000:
You know apt? It can install all those things and it saves us time updating those things and debugging build failures and build time and complexity.
And right now we use a pretty old version of node that isn't even the latest version of that minor.

Sir-Boops:
> It can install all those things and it saves us time

Build caching for local/dev servers, then who cares how long it takes to build the release version

> And right now we use a pretty old version of node that isn't even the latest version of that minor.

You build it in the container so you have version control even the slightly older version here it closer to current version then the version 18.04 is running [src](https://packages.ubuntu.com/bionic/nodejs) [src2](https://nodejs.org/download/release/latest-carbon/)

> And squash is not the ideal way to do something proper.

It shrinks image sizes massively why would you not build your release version using it

SuperSandro2000:
I still care about build time when I need to develop that stuff. If it is not necessary we don't need to do it. We don't really gain anything except more complexity and work and I would like to keep it simple and understandable.

I am not referring to the version in Ubuntu. I am referring to the the official node binary releases for Ubuntu. Same thing with Ruby and python. 

We can combine layers where it makes sense while building. This makes the docker file 100% comparable to the final image and makes debugging easier. It also does not add unnecessary CI complexity.

PR #10827, opened 2019-05-25
Docker: Use precompiled nodejs

SuperSandro2000:
I replaced the manual compile of nodejs, which is not necessary in my opinion with precompiled packages from https://github.com/nodesource/distributions.

Original it took almost 7.5 minutes for node to compile.
````
 => [build-dep 2/7] RUN ECHO "Etc/UTC" > /etc/localtime &&  apt update &&  apt -y dist-upgrade &&  apt -y install w  444.2s
````
Now it takes about 0.5 minute.
````
 => [build-dep 2/7] RUN ECHO "Etc/UTC" > /etc/localtime &&  apt update &&  apt -y install wget &&  wget -O - https:/  32.2s
````
Keep it mind that we need to double that cause we need to reinstall nodejs in the production container. We could copy over nodejs but this is the saver way. Also I moved ``gcc`` and ``make`` to the ``jemalloc`` installation step. This needs possible to move again depending if my other PR gets merged first or at all. This makes the metrics a bit less convincing but still we should save around 5 minutes on a typical developer machine.

Testing done on a i7-4790 @ 3.60 GHz, 32 GB RAM, Samsung 860 EVO 1 TB, 25 Mbit down connection

Edit: This also install node 8.16.X. As this is a minor update I do not expect it to break anything. I saw other can run mastodon on node 10.X so this should be totally fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,23 +4,16 @@ FROM ubuntu:18.04 as build-dep
 SHELL ["bash", "-c"]
 
 # Install Node
-ENV NODE_VER="8.15.0"
 RUN	echo "Etc/UTC" > /etc/localtime && \
 	apt update && \
-	apt -y dist-upgrade && \
-	apt -y install wget make gcc g++ python && \
-	cd ~ && \
-	wget https://nodejs.org/download/release/v$NODE_VER/node-v$NODE_VER.tar.gz && \
-	tar xf node-v$NODE_VER.tar.gz && \
-	cd node-v$NODE_VER && \
-	./configure --prefix=/opt/node && \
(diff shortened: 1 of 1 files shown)


clarfon:
I think it might be better to instead create a base image for Docker and then build using that-- i.e. a mastodon-base image which contains Ruby and Node.js already compiled-- which isn't updated as often. That way you can easily build changes from source without having to rebuild the entire image.

SuperSandro2000:
@clarfon Why do we even compile nodejs from source? There is no need to do so.

ykzts:
Resolve with #11768

PR #10828, opened 2019-05-25
CI: Test if docker image can be build

SuperSandro2000:
I added a circlci job that builds the docker image to see if changes in the Dockerfile or in mastodon would break the build process of the docker image.

This is my first time working with circleci. hope I did it right.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,15 @@ jobs:
       - run: bundle exec i18n-tasks missing -t plural
       - run: bundle exec i18n-tasks check-consistent-interpolations
 
+  build-docker-image:
+    <<: *defaults
+    docker:
+    machine: true
+    steps:
+      - *attach_workspace
+      - run: docker build -t tootsuite/mastodon:latest .
+      - run: docker images
+
 workflows:
   version: 2
   build-and-test:
@@ -215,3 +224,6 @@ workflows:
       - check-i18n:
           requires:
             - install-ruby2.6
+      - build-docker-image:
+          requires:
+            - test-ruby2.6


.circleci/config.yml
```diff
@@ -178,6 +178,15 @@ jobs:
...
       - run: bundle exec i18n-tasks check-consistent-interpolations
 
+  build-docker-image:
+    <<: *defaults
+    docker:
+    machine: true
+    steps:
+      - *attach_workspace
+      - run: docker build -t tootsuite/mastodon:latest .
+      - run: docker images
```
Gargron: Why run this line?

SuperSandro2000: Kind of habit and you can compare the image size with previous builds. [Right now it is 1.46GB](https://circleci.com/gh/tootsuite/mastodon/59636) which I think is quite a lot and we can compare if my other improvements not only decrease build time a lot but also image size.

Gargron:
My concern: I am on the free tier of CircleCI so only have one core to process all jobs. It already takes pretty long, and we get multiple PRs at the same time sometimes, plus once it's merged master also has to be tested. Should we maybe limit this to master? Although that defeats the benefit of checking if a PR breaks Docker *before* merging...

SuperSandro2000:
I just looked at the build time and at almost 50 minutes this is insane! On my local machine the worst case was 7.5 mins....
Do you know if circleci has a feature where we can limit this job to changes in the Dockerfile?
It wouldn't be to useful then but at least we get feeback if the image builds when we directly change it.

NARKOZ:
As an alternative, you can schedule the workflow to run once or twice a day https://circleci.com/docs/2.0/configuration-reference/#schedule

Gargron:
I don't think that's practical... The idea was to check within a PR if that PR breaks Docker builds, but doing that only once a day would make that operation quite useless. As for non-PR branches, master gets built by Docker Hub anyway. I think this change might not be worth it.

clarfon:
Yeah, unfortunately, Docker is massively slow with builds and there's absolutely no way this is going to be practical without a lot of dedicated CPU time, which costs a lot of money that isn't worth it in this case.

SuperSandro2000:
@clarfon docker is not significantly slower that real hardware. You either have just not good enough hardware or just a really bad docker image that builds everything from scratch where there is no need for it.

clarfon:
It wouldn't be that much slower than actual hardware if actual hardware were running the builds. Dockerhub and Circle-CI give abysmal amounts of CPU time for free plans and while I agree that work should be made to improve the build time of the docker image, I don't think it will ever be practical to test PRs' Docker builds without substantial caching.

PR #10829, opened 2019-05-25
Add a featured link

cutls:
## What's this request?

Add a featured link on the frontpage.  
It can be used for a link to a funding, a status page, a community wiki or something.  

## Screenshots
clink2  
clink1

### DIFF
--- a/app/models/form/admin_settings.rb
+++ b/app/models/form/admin_settings.rb
@@ -28,6 +28,10 @@ class Form::AdminSettings
     thumbnail
     hero
     mascot
+    featured_link
+    featured_link_address
+    featured_link_title
+    featured_link_desc
   ).freeze
 
   BOOLEAN_KEYS = %i(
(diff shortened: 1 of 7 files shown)


cutls:
Now, I'M checking i18n...

Gargron:
I like the idea! I like where it's positioned! But I wonder if it's taking the `Setting` model a bit too far, maybe a separate model is appropriate? After all, it's multiple values grouped together like a row in a table. Deferring to other contributors

nightpool:
could we just put a hash of {link, title, desc, icon} into the settings
table directly?

Gargron:
Technically, yeah, the settings gem encodes stuff as YAML. Not sure how elegant that is though.

cutls:
I think it is not best code, but I have no idea to improve it. I wish I had ability to write an elegant code...

PR #10831, opened 2019-05-25
Setup: Fix SMTP /w OpenSSL verify peer unable to get local issuer certificate

SuperSandro2000:
When I tried setup SMTP in the Docker container I kept getting this Error:
````
E-mail could not be sent with this configuration, try again.
SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)
````
I know for a fact that my SMTP configuration was correct as I use it for another service.
I found this fix https://stackoverflow.com/a/36526541/4446318 and this in the code https://github.com/tootsuite/mastodon/search?q=SMTP_CA_FILE&unscoped_q=SMTP_CA_FILE .

I am unsure if we should set this to a ``ENV`` and ask the user about the file? The last one would support self signed certificated on the SMTP server side but this is quite niche and could overwhelm the end user.

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -285,6 +285,7 @@ namespace :mastodon do
             domain:               env['LOCAL_DOMAIN'],
             authentication:       env['SMTP_AUTH_METHOD'] == 'none' ? nil : env['SMTP_AUTH_METHOD'] || :plain,
             openssl_verify_mode:  env['SMTP_OPENSSL_VERIFY_MODE'],
+            ca_file:              '/etc/ssl/certs/ca-certificates.crt',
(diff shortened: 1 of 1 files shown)


Gargron:
My first thought is: Does this have any side effects, does this break any existing configurations?

SuperSandro2000:
I am relative new to the code but as far as the desc [here](https://github.com/tootsuite/mastodon/blob/master/lib/tasks/mastodon.rake#L7) goes it only affects the setup script to get your instance ready for production. Also this portion of the code should only be used to send a test email to see if your provided settings are correct.

SuperSandro2000:
Bump

SuperSandro2000:
spam

PR #10836, opened 2019-05-26
Fix sanitizer making block level elements unreadable

### LINKED ISSUE #10834: Keep lists/headings in sanitizer, or convert to <p>
Semi-related: #8474

Stripping out block-level elements like `ul`, `ol`, `li`, `h1`-`h6`, etc. can cause a breakdown in structure, since their contents will be assumed to be part of one big element and this is incorrect.

### Example remote `content` field:
```
<h1>sample heading</h1>
<p>lorem ipsum dolor sit amet nunquam et titillandus. et qua non si dolorem qui va parfait...</p>
<ul>
<li>item 1</li>
<li>item 2</li>
<li>item 3</li>
```

Current sanitization would result in this:
```
sample heading <p>lorem ipsum dolor sit amet nunquam et titillandus. et qua non si dolorem qui va parfait...</p> item 1 item 2 item 3
```

This results in a noticeable mess when multiple unsupported block-level elements are used in succession, as commonly happens in lists, but can also happen with multiple headings in a row.

image

But preserving the elements actually doesn't look wrong, because basics.scss already contains overrides for everything:

image

Relevant CSS from basics.scss:
```
a, abbr, acronym, address, applet, article, aside, audio, b, big, blockquote, body, canvas, caption, center, cite, code, dd, del, details, dfn, div, dl, dt, em, embed, fieldset, figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6, header, hgroup, html, i, iframe, img, ins, kbd, label, legend, li, mark, menu, nav, object, ol, output, p, pre, q, ruby, s, samp, section, small, span, strike, strong, sub, summary, sup, table, tbody, td, tfoot, th, thead, time, tr, tt, u, ul, var, video {
  font: inherit
}
ol, ul {
  list-style: none
}
```

Semantically, it may be undesirable to maintain certain header levels -- I don't know enough about this, but if they are a problem, converting `h1`-`h6` to `p` should look exactly the same, anyway, so there's no *visual* difference whether the element is an `h1` or a `p`.

Gargron:
Fix #10834

### DIFF
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -19,6 +19,20 @@ class Sanitize
       node['class'] = class_list.join(' ')
     end
 
+    UNSUPPORTED_ELEMENTS_TRANSFORMER = lambda do |env|
+      return unless %w(h1 h2 h3 h4 h5 h6 blockquote pre ul ol li).include?(env[:node_name])
+
+      case env[:node_name]
+      when 'li'
+        env[:node].traverse do |node|
+          node.add_next_sibling('<br>') if node.next_sibling
+          node.replace(node.children) unless node.text?
+        end
+      else
+        env[:node].name = 'p'
+      end
+    end
+
     MASTODON_STRICT ||= freeze_config(
       elements: %w(p br span a),
 
@@ -40,6 +54,7 @@ class Sanitize
 
       transformers: [
         CLASS_WHITELIST_TRANSFORMER,
+        UNSUPPORTED_ELEMENTS_TRANSFORMER,
       ]
     )
 
--- a/spec/lib/sanitize_config_spec.rb
+++ b/spec/lib/sanitize_config_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require Rails.root.join('app', 'lib', 'sanitize_config.rb')
+
+describe Sanitize::Config do
+  describe '::MASTODON_STRICT' do
+    subject { Sanitize::Config::MASTODON_STRICT }
+
+    it 'converts h1 to p' do
+      expect(Sanitize.fragment('<h1>Foo</h1>', subject)).to eq '<p>Foo</p>'
+    end
+
+    it 'converts ul to p' do
+      expect(Sanitize.fragment('<p>Check out:</p><ul><li>Foo</li><li>Bar</li></ul>', subject)).to eq '<p>Check out:</p><p>Foo<br>Bar</p>'
+    end
+
+    it 'converts p inside ul' do
+      expect(Sanitize.fragment('<ul><li><p>Foo</p><p>Bar</p></li><li>Baz</li></ul>', subject)).to eq '<p>Foo<br>Bar<br>Baz</p>'
+    end
+
+    it 'converts ul inside ul' do
+      expect(Sanitize.fragment('<ul><li>Foo</li><li><ul><li>Bar</li><li>Baz</li></ul></li></ul>', subject)).to eq '<p>Foo<br>Bar<br>Baz</p>'
+    end
+  end
+end


ThibG:
I'd still prefer actual support for those tags, but this is an improvement on the current behavior.

ThibG:
Thinking about it, it's a good opportunity to translate inline `img` tags to a link to the original image + use the alt tag.

spec/lib/sanitize_config_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+describe Sanitize::Config do
+  describe '::MASTODON_STRICT' do
+    subject { Sanitize::Config::MASTODON_STRICT }
+
+    it 'converts h1 to p' do
+      expect(Sanitize.fragment('<h1>Foo</h1>', subject)).to eq '<p>Foo</p>'
+    end
+
+    it 'converts li to p' do
```
trwnh: did i misunderstand the job of UNSUPPORTED_ELEMENTS_TRANSFORMER? to me it read like (h1 h2 h3 h4 h5 h6 blockquote pre li) would be preserved, and anything else would be converted to p

the reason i ask is because converting li to p would insert margins below it, and that causes excessive spacing. essentially, an li element has only 1 visual line break, whereas a p element has 2 visual line breaks.

app/lib/sanitize_config.rb
```diff
@@ -19,6 +19,11 @@ module Config
       node['class'] = class_list.join(' ')
     end
 
+    UNSUPPORTED_ELEMENTS_TRANSFORMER = lambda do |env|
+      return unless %w(h1 h2 h3 h4 h5 h6 blockquote pre li).include?(env[:node_name])
```
trwnh: fwiw if this line does what i think it does, then pre should be converted instead of maintained (unless you add css along the lines of `overflow-x: auto` or `white-space: pre-wrap`.

headers, blockquote, and li are reset by basics.scss regardless. for that matter, ol and ul are also safely reset by the current css.

spec/lib/sanitize_config_spec.rb
Gargron: Yes, you probably misunderstood. They all get converted to `<p>`.

I feel like single line breaks can be very hard to notice within longer text, so while for single-line list items it would be best to use `<br />`, `<p>` is more fit for anything that has its own wrapping.

trwnh: ah, so it's like this then?
1. keep (p br span a)
2. convert (h1 h2 h3 h4 h5 h6 blockquote pre li) to p
3. strip all other elements

the "unless" was throwing me off.

i still think semantically if you were trying to map ol/ul and li, then li would be most like a br element. most formatting on the rest of the internet (and indeed most browsers' default css) treats lists as single-break, with double-break around the list due to margin css.

i'm still not sure what the point is in explicitly stripping out the list elements, since they render unobtrusively anyway. my idea of converting to p was more for unsupported elements like marquee. as far as i can tell this current code just removes marquee entirely, right? that would cause multiple marquees to become a run-on sentence.

ThibG: Actually, I think converting `li` to `p` elements might indeed be a problem. For a start, `li` elements can contain a `p` already. Also, regarding how it would be displayed:

## With my patch

image

## Converted as `p` tags

image

trwnh: likewise, this is what the following markup looks like without any patches:
```
<ul><li>item</li><li>item</li><li>item</li><li>item</li><li>item</li></ul>
```
image

no-boot-device:
For the record, I don't see why this is necessary when the user opinion has been massively in favor of full support.

ClearlyClaire (approved):
I would still prefer if we accepted those tags directly, but I think this is the best fallback we can have, even though it still loses information (nested lists for instance).

Also, this PR might be a good place for replacing incoming inlined images with links (as in https://github.com/glitch-soc/mastodon/pull/1071)

PR #10838, opened 2019-05-26
Fix null error in status component when determining showMedia state

Gargron:
Also update the `showMedia` value if the status passed to the component changes. Regression from #10647

Edit: Found another bug, the code for `showMedia` in the status component doesn't use the reblogged status properly, so reblogs of things with sensitive media get their media displayed. But it's 5:30am now, I can't fix it right now

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -46,6 +46,10 @@ export const defaultMediaVisibility = (status) => {
     return undefined;
   }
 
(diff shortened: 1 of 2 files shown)


ThibG:
The changes so far look fine to me, but I'm puzzled about the `sensitive` thing with reblogs. I mean, I'm looking at the previous code and it passed `this.props.status.get('sensitive')` as the `sensitive` prop to `MediaGallery`, so it should have been broken already :thinking:

ThibG:
I pushed two commits that should fix the issue and clean up the code a bit.

PR #10841, opened 2019-05-26
Improvement variable height of compose form in single column layout

noellabo:
This is an improvement to make the height of compose form in single column layout variable from fixed length.

When the input content exceeds the standard height, it expands to the location of Getting started footer, and when it is exceeded, the scroll bar is displayed.

scrolling_compose_form


app/javascript/styles/mastodon/components.scss
```diff
@@ -2204,10 +2204,37 @@ a.account__display-name {
...
+    top: 5px;
+    left: 234px;
+    z-index: 1;
+    width: 0;
+    height: 0;
+  }
+
+  .compose-form__autosuggest-wrapper {
+    overflow-y: auto;
+    background-color: white;
```
yuzulabo:
```suggestion
    background-color: $white;
```

I think this change will pass the test.
ref: https://codeclimate.com/github/tootsuite/mastodon/pull/10841

Gargron:
You have accidentally submitted your entire fork

noellabo:
I'm sorry.

And another bug was found that the suggestion was hidden and not visible. Close the request once.

marek-lach:
> I'm sorry.
> 
> And another bug was found that the suggestion was hidden and not visible. Close the request once.

You can create a new branch in your fork, then submit the changes by selecting that new branch in your fork and changing all the necessary files that way, so it's then easy to submit the PR here again.

PR #10843, opened 2019-05-26
In basic UI, bring 2-column view to the center

Hanage999:
Fix the layout of the new basic UI where columns are justified to the right when 2 columns are shown.

### DIFF
--- a/app/javascript/styles/mastodon/basics.scss
+++ b/app/javascript/styles/mastodon/basics.scss
@@ -136,5 +136,6 @@ button {
     align-items: center;
     justify-content: center;
     outline: 0 !important;
+    overflow-x: auto;
   }
 }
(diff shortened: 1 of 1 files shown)


Gargron:
Any side effects here? Does it affect how the multi-column UI looks?

Gargron:
Two side effects:

1. When the page loads, the width of the center element jumps
2. In multi-column layout, everything becomes centered

Hanage999:
> Two side effects:
> 
>     1. When the page loads, the width of the center element jumps
> 
>     2. In multi-column layout, everything becomes centered

2 looks like a good effect, and I would not mind 1, but it may well depend on how you use Web UI daily.

Hanage999:
Another side effect:

3. In multi-column layout, you cannot scroll to the leftmost column when a window is not wide enough to cover all UI elements.

At least I'm going to fix this.

trwnh:
It may be useful to add a class to the base body tag when "simple mode" is set or when "advanced mode" is set. This would allow CSS fixes to target only one of the two layouts.

Hanage999:
https://github.com/tootsuite/mastodon/pull/10915 should serve the same purpose without side effects, so I close this PR.

PR #10844, opened 2019-05-26
Fix React warning about legacy lifecycle calls

Gargron:


ThibG:
This will trigger an useless render, I guess we could use `getDerivedStateFromProps` instead.
Also, it just occurred to me that this will reset the state when doing stuff like faving or boosting the toot. We should compare the toot's id only, and not the full contents.

ClearlyClaire (changes requested):
This will trigger an useless render, I guess we could use getDerivedStateFromProps instead.
Also, it just occurred to me that this will reset the state when doing stuff like faving or boosting the toot. We should compare the toot's id only, and not the full contents.

We basically want to do this, I think: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-1-reset-uncontrolled-component-with-an-id-prop

PR #10847, opened 2019-05-26
Change default to single column UI

Gargron:
Keeps current setting for people who have been active within the month preceding the database migration, if they were relying on the default value.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -31,7 +31,7 @@ defaults: &defaults
   noindex: false
   theme: 'default'
   aggregate_reblogs: true
-  advanced_layout: true
+  advanced_layout: false
   notification_emails:
     follow: false
     reblog: false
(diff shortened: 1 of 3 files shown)


ykzts:
I think existing users will be puzzled by change. I think it is better to change only the default for new users.

marek-lach:
> I think existing users will be puzzled by change. I think it is better to change only the default for new users.

Agreed. This is why I think that the UI switch between single/advanced should also be in the Getting Started section, so that it's easily and quickly accessible and switch between for users.

StarryAster:
A request:
Just like we admins can change the default theme for new users, could we also be able to choose if we want the single or multi column layout as default, instead of just changing the default for everybody who updates their instance?

This can really help those who have themes that are heavily customized, tutorials that rely on the multi-column layout, or other things to be able to update those accordingly before choosing single-column as default, while individual users can still enjoy the single-column layout if they prefer it.

Gargron:
Updated PR to keep current setting for people active within the last month.

ykzts (approved):

PR #10852, opened 2019-05-27
Improve Preferences Page

robinmaypanpan:
To a novice user, the preferences page can seem like a huge wall of disorganized options.

This PR improves that with two major changes:
- With language options on top, the user is confronted with a huge wall of checkboxes, which most of the time they just have to scroll past to get to options they may want to change more often.  This moves those options to the bottom so they are out of the way, since they will likely only need to be changed once, on initial setup of your account.
- Each section in the preferences pane is now given header text to help clarify what settings are what and make it easier to identify.

I'm not too familiar with haml, so let me know if I'm doing anything funky with the haml here.

alt text

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -22,6 +22,7 @@
 #  application_id         :bigint(8)
 #  in_reply_to_account_id :bigint(8)
 #  poll_id                :bigint(8)
+#  local_only             :boolean
 #
 
 class Status < ApplicationRecord
--- a/app/views/settings/preferences/show.html.haml
+++ b/app/views/settings/preferences/show.html.haml
@@ -2,25 +2,19 @@
   = t('settings.preferences')
 
 %ul.quick-nav
-  %li= link_to t('preferences.languages'), '#settings_languages'
   %li= link_to t('preferences.publishing'), '#settings_publishing'
   %li= link_to t('preferences.other'), '#settings_other'
   %li= link_to t('preferences.web'), '#settings_web'
+  %li= link_to t('preferences.languages'), '#settings_languages'
   %li= link_to t('settings.notifications'), settings_notifications_path
 
 = simple_form_for current_user, url: settings_preferences_path, html: { method: :put } do |f|
   = render 'shared/error_messages', object: current_user
 
-  .fields-row#settings_languages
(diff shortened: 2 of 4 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_19_130537) do
+ActiveRecord::Schema.define(version: 2019_05_24_221356) do
```
Gargron: This file change is irrelevant to the PR

app/models/status.rb
```diff
@@ -22,6 +22,7 @@
 #  application_id         :bigint(8)
 #  in_reply_to_account_id :bigint(8)
 #  poll_id                :bigint(8)
+#  local_only             :boolean
```
Gargron: This change is irrelevant to the PR

app/views/settings/preferences/show.html.haml
```diff
@@ -2,33 +2,30 @@
...
-      = f.input :locale, collection: I18n.available_locales, wrapper: :with_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }, selected: I18n.locale
-    .fields-group.fields-row__column.fields-row__column-6
-      = f.input :setting_default_language, collection: [nil] + filterable_languages.sort, wrapper: :with_label, label_method: lambda { |locale| locale.nil? ? I18n.t('statuses.language_detection') : human_locale(locale) }, required: false, include_blank: false
-
-  .fields-group
-    = f.input :chosen_languages, collection: filterable_languages.sort, wrapper: :with_block_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }, required: false, as: :check_boxes, collection_wrapper_tag: 'ul', item_wrapper_tag: 'li'
-
   %hr#settings_publishing/
 
+  .heading
```
Gargron: `<h4>` tags would be better

Gargron (commented):
I have submitted #10977 that mainly focuses on extracting appearance-related options into a separate page. It also re-organizes some options on the general preferences page, similar to your PR. It is possible that that unfortunately makes this PR redundant. Let me know what you think

PR #10856, opened 2019-05-27
Increase default privacy settings for new users

robinmaypanpan:
New users to Mastodon may not realize that their toots may be indexed by search engines, nor that the list of who they choose to follow or who follows them will be made public.  This can lead to users that want this behavior not realizing they can change these options.

For those users that care about this, let's default to a higher level of privacy, so that they can be safer.  Users that want to expose their data can change these options later.

Also changed the "Other" section in Preferences to be named "Privacy" since that's closer to what these 3 options are.

### DIFF
--- a/app/views/settings/preferences/show.html.haml
+++ b/app/views/settings/preferences/show.html.haml
@@ -4,7 +4,7 @@
 %ul.quick-nav
   %li= link_to t('preferences.languages'), '#settings_languages'
   %li= link_to t('preferences.publishing'), '#settings_publishing'
-  %li= link_to t('preferences.other'), '#settings_other'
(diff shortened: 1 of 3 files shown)


marek-lach:
The title of this PR should probably be changed to ''Increase default privacy settings for new users'', or ''Stricter default privacy settings for new users'' ?

ykzts:
> I think this would be an unexpected default. Every other social media profile is findable through search engines. Specific communities on Mastodon might have a different expectation, and changing the instance-wide default should be easier for admins, but this repository serves more communities than those ones and I think overall this change would be considered an anti-feature.

https://github.com/tootsuite/mastodon/pull/11752#issuecomment-527659350

PR #10860, opened 2019-05-27
Improve rate limiting

ClearlyClaire:
- Use actual client IP for whitelisting and rate limiting when a reverse-proxy is involved
- Reduce allowed number of unauthenticated API requests
- Aggressively limit requests which involve paging


ThibG:
I'm not sure how much sense the limits make, and they'll probably affect search engine crawlers negatively.
One reason for this was to make life harder for people trying to archive public toots of an instance by crawling. I don't know how effective those limits would be.

Gargron (approved):

PR #10891, opened 2019-05-29
Change compose-panel to scrollable

mayaeh:
In my laptop environment, “TOOT” button may not fit on the screen via single column layout.
This may also occur in other small screen environments.

So I want to be able to scroll the "compose-panel".
What do you think about this?
スクリーンショット 2019-05-30 0 50 03

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -2186,6 +2186,7 @@ a.account__display-name {
   display: flex;
   flex-direction: column;
   height: 100%;
+  overflow: auto;
 
   .search__input {
     line-height: 18px;


Gargron:
We don't have many options... It's either this, or making the entire page scroll. One page wide scrollbar might be prettier though. Twitter makes the side panel scroll with JS when you scroll the whole page, for example

noellabo:
The advantage of this approach is that the compose panel and navigation links always stay within the screen. If you scroll the entire screen, you will have to go back to the top to post status or switch screens.

I like this approach.

The right navigation panel may also overflow. You should apply `overflow-y: auto;` to both side panels.

Also, the height of the post section is limited at present, but if you can scroll, it is better to delete `max-height: 200px;` of `.autosuggest-textarea__textarea`.

Gargron:
The problem is that scrollbars are an ugly UI element and the compose panel is not a visually boxed element so the scrollbar is hanging over empty space. I don't know how to deal with this.

mayaeh:
I agree. I realized that creating multiple scroll bars on a single column layout page is not good.
I think that it is difficult to deal with this problem.

Gargron:
It might be possible to bind to the wheel event and when scrolling down, ensure via JS that the bottom part of the compose-panel scrolls into view, and when scrolling up, return it to the top. That is roughly what Twitter does with the right panel that has the same issue. Although that is a bit improper as it is re-implementing scroll manually.

Alternatively, it could be possible to set compose-panel height to a pixel value, so that when it grows, it allows you to scroll the entire page (although it probably requires setting explicit height on the surrounding container elements as well)

mayaeh:
I will close this PR.
Because this PR breaks the idea of ​​single column layout.
I don't think that it is good that a lot of scroll bars line up.
Thank you for thinking about this PR.

trwnh:
For what it's worth, this is also an issue with the multi-column layout, IIRC. It seems like the best thing to do would be to set the max-height of the sidebar to 100vh or something slightly less to account for margins, then make the compose box overflow-y scrollable (because it is visually a box and therefore a scrollbar doesn't look too bad)

PR #10911, opened 2019-05-31
Whitelisted federation mode for academic institutions

### LINKED ISSUE #3880: [Feature Request] Instance-level whitelist
It would be nice for the admins of instances that are more concerned about privacy/leaking information to be able to choose between using the instance blacklist and an instance whitelist. This is already implemented for awoo.space, the code of which is publicly available.

https://github.com/vahnj/mastodon/commit/2cf7a096bf642ba75b3fe327038ee598f701db44
https://github.com/vahnj/mastodon/commit/2fca7e899aebc37fb45190951c1c2d46c72eb646

I am sure it needs to be cleaned up, have tests added, and so on, but I would love to see this feature more widely available for admins.

* * * *

Gargron:
Schools and universities want to teach students how to use social media in a controlled environment. For the safety of the students, data in- and output must be tightly controlled.

In "whitelist mode":

- [x] Public pages cannot be viewed without authentication
- [x] API access is not possible without a user-owned access token (no app-only access)
- [x] ActivityPub resources cannot be accessed without a HTTP signature \*
- [x] Accounts that do not belong to a whitelisted domain are not processed
- [x] ActivityPub actor JSON remains public, but returns the bare minimum without authentication

The risk of this mode is if it gains widespread adoption in non-academic contexts, thereby making entering the federation with a new server impractical. That would lead to the centralization around a few existing servers in the long term.

My risk mitigation is that the mode should have a "penalty" in the form of a clearly distinguishable branding that makes it clear to any visitor that they are looking at a standalone space and not at the fediverse / open web / Mastodon network.

Fix #3880

___

Pathways that can still be accessed without authentication:

- Webfinger
- Host-meta
- App creation API
- About, terms
- Login, sign up, forgotten password, e-mail confirmation
- ActivityPub actor JSON (bare minimum version)

___

TODO:

- [ ] Authenticate GETs of actor JSON
- [ ] Change frontpage to clearly communicate it's not part of the wider fediverse when in this mode
- [ ] Add tests

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -3,6 +3,7 @@
 class AboutController < ApplicationController
   layout 'public'
 
+  skip_before_action :check_access_requirements
   before_action :set_instance_presenter, only: [:show, :more, :terms]
 
   def show
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -5,6 +5,7 @@ class AccountsController < ApplicationController
 
   include AccountControllerConcern
 
+  skip_before_action :check_access_requirements, if: -> { request.format == :json }
   before_action :set_cache_headers
 
   def show
@@ -49,7 +50,7 @@ class AccountsController < ApplicationController
         mark_cacheable!
 
         render_cached_json(['activitypub', 'actor', @account], content_type: 'application/activity+json') do
-          ActiveModelSerializers::SerializableResource.new(@account, serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter)
+          ActiveModelSerializers::SerializableResource.new(@account, serializer: actor_serializer, adapter: ActivityPub::Adapter)
         end
       end
     end
@@ -136,6 +137,14 @@ class AccountsController < ApplicationController
       filtered_statuses.paginate_by_min_id(PAGE_SIZE, params[:min_id]).reverse
     else
       filtered_statuses.paginate_by_max_id(PAGE_SIZE, params[:max_id], params[:since_id]).to_a
+    end
+  end
+
+  def actor_serializer
+    if whitelist_mode? && !current_account
+      ActivityPub::BareActorSerializer
+    else
+      ActivityPub::ActorSerializer
     end
   end
 end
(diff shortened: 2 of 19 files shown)


marek-lach:
The label on such a profile could be something like:

''Access-controlled'' account, or
''Limited federation'' account...

Just throwing in an idea, in case you're not quite sure how to call it.

ThibG:
Just skimmed through the PR and I have a few questions/remarks:
- Are OStatus and RSS pathways properly disabled?
- As far as I can see, if an instance *switches* to whitelist mode, already-known remote accounts will still be able to interact. Same when removing instances from the whitelist.
- HTTP-level caching of statuses and actor data should also be disabled
- For this whole PR to be useful, Mastodon needs to sign every fetch, using an instance-wide actor, which isn't the case yet
- Finally, not caching the toots may have very serious performance implications

Otherwise, big :+1:

Gargron:
> As far as I can see, if an instance _switches_ to whitelist mode, already-known remote accounts will still be able to interact. Same when removing instances from the whitelist.

I do not expect this feature to be used so dynamically (enabling it on an already active instance)

> Finally, not caching the toots may have very serious performance implications

Unlikely to be a practical issue between 2-3 schools or universities. In fact, I suspect that most schools just won't federate with anyone at all.

> For this whole PR to be useful, Mastodon needs to sign every fetch, using an instance-wide actor, which isn't the case yet

Yes. On the TODO list. It can start with a random account for the first implementation.

> Are OStatus and RSS pathways properly disabled?

OStatus code goes through `ResolveAccountService`, which refuses to save accounts that are not whitelisted. So anything incoming would be ignored. All routes require authentication, except the few that have `skip_before_action`, so RSS/Atom should not be available to logged out visitors. There should be no way for a mention of a remote, not whitelisted account to be saved, so those won't trigger deliveries. WebSub remains, but the endpoint is likewise behind login. so no WebSub subscriptions should be creatable. Still, I wouldn't mind a detailed audit on this!

Laurelai:
> I do not expect this feature to be used so dynamically (enabling it on an already active instance)

Oh. You know for a fact people have been begging for this feature for a long time. You know people will want to use it. Switching to whitelist mode should purge all federation thats not whitelisted.

PR #10917, opened 2019-06-01
Improvement variable height in single column layout

noellabo:
This is an improvement to make the height of compose panel and navigation panel in single column layout variable from fixed length.

In the case where the compose panel has the longest vertical length, the following is assumed.

- Warning display of status visibility (WarningContainer)
- Display of reply source (ReplyIndicatorContainer)
- Field for spoiler text (AutosuggestInput)
- Compose form (AutosuggestTextarea)
- Attached image (UploadFormContainer) or poll form (PollFormContainer)

I made adjustments to make it possible to display this at the same time.

Also for the navigation panel, the vertical length gets longer depending on the number of lists, so we scroll only when necessary.

Check the animated GIF for behavior.
compose_form


noellabo:
Now that we have confirmed the operation on Microsoft Edge, remove the WIP.
Please comment if you need to fix it.

However, my changes have made AutosuggestTextarea go beyond its original role. This requires splitting, but I do not know how to write the appropriate code. Someone, please merge this PR and then refactor it.

Gargron (approved):

PR #10929, opened 2019-06-02
Feature ignore toot language

cutls:
# What's this?
Ignore `language` in status when WebUI rendering(rendering without HTML `lang` attr).  
When tooting in CJK(Chinese, Japanese, Korean), Mastodon sometimes detects wrong language.  
So, Japanese people often see  toots written in Japanese with Chinese fonts, which they are not used to.  
However, I think it is a feature only for Japanese...


cutls:
Sorry, I mistook. Close and reopen later.

PR #10930, opened 2019-06-02
Ignore toot language

cutls:
# What's this?
Ignore `language` in status when WebUI rendering(rendering without HTML `lang` attr).  
When tooting in CJK(Chinese, Japanese, Korean), Mastodon sometimes detects wrong language.  
So, Japanese people often see  toots written in Japanese with Chinese fonts, which they are not used to.  
However, I think it may be a feature only for Japanese...

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -36,6 +36,7 @@ class Settings::PreferencesController < Settings::BaseController
       :setting_default_privacy,
       :setting_default_sensitive,
       :setting_default_language,
+      :setting_ignore_toot_language,
       :setting_unfollow_modal,
       :setting_boost_modal,
       :setting_delete_modal,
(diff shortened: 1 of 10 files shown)


Gargron:
Should we just remove the `lang` attribute from web UI? I've only recently put it there. I'm not sure if it's truly useful.

yuzulabo:
If it has a `lang` attr, the font for that language takes precedence.
Japanese and Chinese use kanji, but the font is different.
It is strange that we look at other kanji fonts.

| lang=zh (Automatic detection) | none |
| -------- | ------- |
| image | image |
| image | image |

tateisu:
in case of CJK, setting HTML lang attribute for the readers region may be better than for the writer's language.

tateisu:
also, Google does not use the lang attribute, this is not helpful for SEO.  https://seopack.jp/seoblog/20161111-use-hreflang/

tateisu:
Web browser's behaviors for lang attribute is:
- change font
- change quotation mark of <q>
- there is CSS pseudo-class :lang

All behaviors are helpful for reader, not for writer's language.

PR #10946, opened 2019-06-03
Handle blank poll options more gracefully

ClearlyClaire:
Pleroma currently allows (erroneously imho) empty poll options, that is,
options with an empty (but existing) `name`.

In that case, `item['name'].presence || item['content']` returns `nil`, which
causes an error during validation. Using `item['name'] || item['content']`
instead will still use `item['content']` when `item['name']` is not set at all,
but return an empty string in Pleroma's case, which will nicely reject the
blank option without raising an error.


Gargron:
Wouldn't it be equally simple to add a `.compact` to the chain?

ThibG:
I guess so.

Gargron (approved):

PR #10960, opened 2019-06-04
Fix poll API not requiring authentication on non-public polls

### LINKED ISSUE #10959: Polls API doesn't require authentication
### Expected behaviour

Polls posted to followers or DM's require authentication.

### Actual behaviour

The poll API returns the data publicly.

Create a poll that's posted to followers only or DM, get the poll ID using the network inspector, go to the API page in an unauthenticated window.

Example: 
image
https://linuxrocks.online/api/v1/polls/2502 is the poll API URL, which works without authentication.

### Specifications

v2.8.4

Gargron:
That API does not reveal the content of the status, i.e. the question itself, nor who the author is, nor which status it belongs to, but it does reveal the poll options and how many answers they got

Fix #10959

### DIFF
--- a/spec/controllers/api/v1/polls_controller_spec.rb
+++ b/spec/controllers/api/v1/polls_controller_spec.rb
@@ -10,14 +10,26 @@ RSpec.describe Api::V1::PollsController, type: :controller do
   before { allow(controller).to receive(:doorkeeper_token) { token } }
 
   describe 'GET #show' do
-    let(:poll) { Fabricate(:poll) }
+    let(:poll) { Fabricate(:poll, status: Fabricate(:status, visibility: visibility)) }
 
     before do
       get :show, params: { id: poll.id }
     end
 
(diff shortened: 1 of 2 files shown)


ClearlyClaire (approved):
LGTM. Adding a test for that might be good.

spec/controllers/api/v1/polls_controller_spec.rb
```diff
@@ -10,14 +10,26 @@
...
+
+      it 'returns http success' do
+        expect(response).to have_http_status(200)
+      end
+    end
+
+    context 'when parent status is private' do
+      let(:visibility) { 'private' }
+
+      it 'returns http success' do
```
ThibG: “returns http unauthorized”

ClearlyClaire (changes requested):

ClearlyClaire (approved):

PR #10966, opened 2019-06-04
Refactor all ActivityPub deliveries to be serialized and signed through one concern

Gargron:


app/services/concerns/payloadable.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Payloadable
+  def serialize_payload(record, serializer, options = {})
+    signer    = options.delete(:signer)
+    sign_with = options.delete(:sign_with)
+    payload   = ActiveModelSerializers::SerializableResource.new(record, options.merge(serializer: serializer, adapter: ActivityPub::Adapter)).as_json
+
+    if (record.respond_to?(:sign?) && record.sign?) && signer && signing_enabled?
```
ThibG: small nitpick, those parenthesis do not help with readability

app/services/concerns/payloadable.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Payloadable
```
ThibG: Small nitpick: this name is weird, especially since it's used by services and not models…

ClearlyClaire (approved):
A few nitpicks but LGTM overall.
Also, is there an advantage using `Oj.dump` over `#to_json`?

Gargron:
> Also, is there an advantage using `Oj.dump` over `#to_json`?

I believe AMS was configured to use Oj in the background so there is no difference, except that the output of `serialize_payload` is sometimes used as a hash, so it's too early to convert it to a string there.

PR #10967, opened 2019-06-04
Do not misattribute inlined boosts if `attributedTo` isn't present

### LINKED ISSUE #10950: ActivityPub 'Announce' sometimes not displayed correctly as 'Boost'
I'm sending something like this:
```json
{"type":"Announce","id":"https://domain.tld/n/2944171-0#top","actor":"https://domain.tld/u/announcer","object":{"id":"https://domain.tld/n/2944171-0","published":"2019-06-04T07:51:33.779855Z","url":"https://domain.tld/m/2944171","to":["https://www.w3.org/ns/activitystreams#Public"],"content":"<p>text</p>","attributedTo":"https://domain.tld/u/poster","cc":["https://domain.tld/u/poster/followers/toc"],"type":"Note"},"type":"Announce","@context":["https://www.w3.org/ns/activitystreams","https://w3id.org/security/v1",{"Hashtag":"as:Hashtag"}]}
```

### Expected behaviour

It should be displayed as "@announcer boosted message (@poster@domain.tld: Text)"

### Actual behaviour

Sometimes it displayed correctly (I think its when @poster is already known to mastodon instance), sometimes it displayed as @announcer's(!) post, sometimes it displayed as "@announcer boosted @announcer's(!) post"

### Specifications

I'm discovering my @announcer on mastodon.social instance, both poster and announcer are from another ActivityPub service.

ClearlyClaire:
Fixes #10950

As of 6a5307a5733e7872e7827f32b27111434e0307c4, we accept boosts of inlined toots provided that they can be attributed to the actor boosting it. However, we assume that if no `attributedTo` field is present, then the object is to be attributed to the actor boosting it.

I guess that assumption might not be correct, although cases where that assumption would not be correct are cases where we cannot attribute the toot at all.

### DIFF
--- a/app/lib/activitypub/activity.rb
+++ b/app/lib/activitypub/activity.rb
@@ -143,7 +143,7 @@ class ActivityPub::Activity
 
     # If the boosted toot is embedded and it is a self-boost, handle it like a Create
     unless unsupported_object_type?
-      actor_id = value_or_id(first_of_value(@object['attributedTo'])) || @account.uri
+      actor_id = value_or_id(first_of_value(@object['attributedTo']))
 
       if actor_id == @account.uri
         return ActivityPub::Activity.factory({ 'type' => 'Create', 'actor' => actor_id, 'object' => @object }, @account).perform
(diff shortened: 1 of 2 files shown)


ThibG:
Note that I'm not 100% sure it is what caused the mentioned issue, as the example mentioned in the issue should not cause that. But it's the only code path I can see that uses the boosters' account when processing the boosted toot.

Gargron (approved):

PR #10970, opened 2019-06-05
Scroll to compose form when focus

tribela:
Scroll into left-most correctly when focusing compose textarea

### DIFF
--- a/app/javascript/mastodon/features/compose/components/compose_form.js
+++ b/app/javascript/mastodon/features/compose/components/compose_form.js
@@ -33,6 +33,11 @@ const messages = defineMessages({
 export default @injectIntl
 class ComposeForm extends ImmutablePureComponent {
 
+  constructor(props) {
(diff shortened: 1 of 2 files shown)


app/javascript/mastodon/features/compose/components/compose_form.js
```diff
@@ -33,6 +33,11 @@ const messages = defineMessages({
 export default @injectIntl
 class ComposeForm extends ImmutablePureComponent {
 
+  constructor(props) {
+    super(props);
+    this.composeForm = React.createRef();
```
Gargron: We have never done this. We just use `ref={this.setRef}` and `setRef = c => { this.foo = c; }`

Kjwon15: Thanks. I modified it.

Gargron (approved):

PR #10977, opened 2019-06-06
Change preferences page into appearance, notifications, and other

Gargron:
Notification and appearance preferences now inherit all methods from the preferences controller, merely overriding the redirect-after-save path.

Appearance preferences:

image

Other preferences:

image


ThibG:
The “recommended” stuff seems weird. If that's recommended, shouldn't that be the default? (In the case of gif autoplay I know why this isn't the default, and I guess I understand why you would want to nudge people to enable it, though). Also it's unclear why it's recommended. Especially stuff that does the user no good (e.g., disclosing the authoring app)

I'm not sure CW stuff and confirmation modals really fall into “appearance” (not sure about language either tbh)

Gargron:
> Also it's unclear why it's recommended. Especially stuff that does the user no good (e.g., disclosing the authoring app)

This is worded a bit unfairly. In the vast majority of cases disclosing the app used to make a post has no harmful effect on anyone, and it's good for the ecosystem by promoting good apps. That is why it is on by default. In fact, all the "recommended" items are the ones that are on by default, except GIF autoplay, which is off by default for accessibility reasons but is a good option if you don't suffer from motion sickness. In the case of "group boosts" for example, the option is a bit arcane. Who will know what it really means? So I think it's good to put a "recommended" label on that, as if to signal: This is on by default for a good reason. Leave it there.

> I'm not sure CW stuff and confirmation modals really fall into “appearance” (not sure about language either tbh)

I don't feel any doubt on this part.

ykzts (approved):

trwnh:
i still think "hide your network" and "opt out of search engine indexing" belong with the other "profile" settings. also, grouping boosts in timelines sounds like an appearance thing

PR #10979, opened 2019-06-06
Fix emoji picker being always displayed

noellabo:
Fixed emoji picker to remain displayed above SearchPopout and SearchResultsContainer.


noellabo:
Delete the content that was duplicated with https://github.com/tootsuite/mastodon/pull/10978 .

This is a bug that occurred in https://github.com/tootsuite/mastodon/pull/10917

Gargron (approved):

PR #10981, opened 2019-06-06
Fix display of alt-text when a media attachment is not available

ClearlyClaire:
This dates back to #5054, but I could not find out what that line was for, and I have seen no regression removing it.

Not perfect, but still an improvement I think.

## Before

image

## After

image


Gargron:
I think CodeClimate is not reading in `.sass-lint.yml` correctly :confused:

ykzts (approved):

PR #10983, opened 2019-06-06
Put poll options behind content warnings

ClearlyClaire:
This moves the whole poll component behind content warnings, in a similar way to what glitch-soc does, but it does not move media components there.

Public pages behave in the same way but don't have the poll icon added yet, because the event handling is more tricky.

## Before

image

## After

image

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -16,7 +16,6 @@ import { MediaGallery, Video } from '../features/ui/util/async-components';
 import { HotKeys } from 'react-hotkeys';
 import classNames from 'classnames';
 import Icon from 'mastodon/components/icon';
-import PollContainer from 'mastodon/containers/poll_container';
 import { displayMedia } from '../initial_state';
 
 // We use the component (and not the container) since we do not want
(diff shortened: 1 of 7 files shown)


ThibG:
Added the icon to the public pages

image

This also fixes the styling of the CW to be the same as in the WebUI (`button` VS `a`)

Gargron:
Realistically is there any point in adding a new icon inside the show more button? GlitchSoc has that convention because you hide images behind that as well, but in upstream this would be a new precedent and I'm not sure it's actually necessary...

ThibG:
Well, it's not necessary, but it provides meaningful information, as “the poll is an integral part of the status” (can't find it right now but you said something along those lines when we first discussed hiding them)

Gargron:
That is true, I just feel like that icon is way too tight in there. An alternative idea *could* be writing "Poll attached" in the same style as "Show thread", there is precedent for previewing hidden content on a separate line like with mentions.

ThibG:
I guess this is a possibility but that would be *in addition* to mentions, and that would probably be a pain to implement on public pages.

Gargron:
I'd be more comfortable merging this *right now* if it didn't change how the "show more" button looked. It can always be adjusted later if need arises.

ThibG:
ok, let me change that

Gargron (approved):

PR #10986, opened 2019-06-07
Media Timeline for the main page

Immediate-Zero:
On Pawoo I noticed they have a feature where you can view only Media that has been created or boosted by the people you follow.  Is there anyway a Media Timeline could be implemented?  

If this is the wrong place to put this, do forgive me.  I was having trouble finding a place to request features.


ykzts:
This is not Pull Request.

PR #10993, opened 2019-06-08
Change "Edit profile" menu item name to match page title

mayaeh:
This PR matches the "Edit Profile" menu item with the page title.

This change will confuse translators once.
However, I believe that this change will make it easier to translate "appearance" by #10977 and #10988.

before: 
スクリーンショット_2019-06-08_19_42_00

after: 
スクリーンショット_2019-06-08_19_43_24

### DIFF
--- a/config/navigation.rb
+++ b/config/navigation.rb
@@ -5,7 +5,7 @@ SimpleNavigation::Configuration.run do |navigation|
     n.item :web, safe_join([fa_icon('chevron-left fw'), t('settings.back')]), root_url
 
     n.item :profile, safe_join([fa_icon('user fw'), t('settings.profile')]), settings_profile_url do |s|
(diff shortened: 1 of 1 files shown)


Gargron:
Maybe "profile appearance"? I'm not sure why just "appearance" is not okay

mayaeh:
The page title for settings/profile is "Edit profile".
But its menu item name was "Appearance".
I think that it is better to unify the page title and the menu item name.

Should I change them to "Profile appearance"?

Cassolotl:
Page titles and menu titles being different *is* definitely confusing and makes it harder to find what you're looking for!

koyuawsmbrtn:
I would rather say

Profile
|_ Edit

Shorter and cleaner

PR #10998, opened 2019-06-08
i18n: Update Catalan translations

ghost:


Gargron (approved):

config/locales/ca.yml
```diff
@@ -498,6 +499,12 @@ ca:
       body: "%{reporter} ha informat de %{target}"
       body_remote: Algú des de el domini %{domain} ha informat sobre %{target}
       subject: Informe nou per a %{instance} (#%{id})
+  appearance:
+    advanced_web_interface: Interfície web avançada
+    advanced_web_interface_hint: 'Si vols fer ús de tota l'amplada de la teva pantalla, l'interfície web avançada et permet configurar diverses columnes per a veure molta més informació al mateix temps: Inici, notificacions, línia de temps federada i qualsevol número de llistes i etiquetes.'
```
Gargron: The syntax error probably comes from this line because you use `'` in the text and also surround the text in `'`

salvadorpla: I guess you are right. It's how is writen in Catalan -> l'amplada, l'interfície ...
Any action needed from my side?

Gargron: Wrap the string in `"` instead

ykzts (approved):

PR #10999, opened 2019-06-08
Add an icon denoting polls in the CW button

ClearlyClaire:
Follow-up to #10983. I know you're not a fan of this design, @Gargron, but still opening this PR to keep track of that feature.


Shleeble:
I support this icon. It's cool.

PR #11000, opened 2019-06-08
Specify gzip required in tootctl emoji help

dariusk:
I was using the bulk emoji import command in `tootctl`. I hit an error telling me that the tar archive needed to be gzipped, but it wasn't specified in the help text. This makes the help text clearer about that.


nightpool:
thanks darius!

nightpool (approved):

nightpool:
congrats on the sick PR number

PR #11061, opened 2019-06-12
Fix Serbian pluralization rules requiring a "many" key

Gargron:
According to Unicode there is no "many" key in Serbian: https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html


Gargron:
Opened upstream issue also: https://github.com/svenfuchs/rails-i18n/issues/853

ykzts (approved):

PR #11066, opened 2019-06-12
Fix ordering of keyboard access between CW field, textarea and emoji picker

ClearlyClaire:


noellabo:
With this change, single column EmojiPicker will scroll out.

Although the placement of EmojiPicker on the DOM may not be appropriate, EmojiPicker is a tool to support input, so it should not always be displayed?

(The Advanced UI EmojiPicker also scrolls out, but it should also be visible at all times)

Is there a good solution?

ThibG:
Hm, at least that's not a regression with regards to last release… I wonder how it could be achieved…

Gargron (approved):

Gargron (approved):

PR #11084, opened 2019-06-14
Fix check-i18n builds

Gargron:


ykzts (approved):

Gargron:
It seems like even though I updated our own pluralization rules for "sr" to not include the "many" key, i18n-tasks does not load that fix so it still complains about it missing. Maybe it is time to disable that check?

PR #11095, opened 2019-06-16
Add simple_form default for admin theme selector

alixrossi:
Fixes the "theme" label in Administration/Site settings not having a translation string, and thus falling back to `human_attribute_name`. As an example:
image


palindromordnilap:
... Well I screwed up. Gonna close this PR and make a new one.

PR #11099, opened 2019-06-17
Bring back mascot on single column

tribela:
Make mascot appear again!

It moves link footer to right panel and bring back mascot on left bottom corner

image

### DIFF
--- a/app/javascript/mastodon/features/ui/components/columns_area.js
+++ b/app/javascript/mastodon/features/ui/components/columns_area.js
@@ -20,6 +20,9 @@ import NavigationPanel from './navigation_panel';
 import detectPassiveEvents from 'detect-passive-events';
 import { scrollRight } from '../../../scroll';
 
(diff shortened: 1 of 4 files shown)


joycem137:
The mascot image isn't scaled in this version of the fix, so if you use a different size mascot than default, it'll appear at its native resolution and be huge or tiny.

Kjwon15:
oh. I'll find way to fix that

joycem137:
This will still break some mascots that are tall, since you're not constraining on the height.  Also, make sure to try really tall compose boxes to see how they interact with it.

Here's what I'm seeing on my dev instance with your PR:

Kjwon15:
@joycem137 Since it's floating from main UI, It's not possible to resize like advanced UI without js support.
What do you think about changing z-index to hide behind compose box?

Kjwon15:
I tried to use flex and `position: relative` to slide left, chrome doesn't support 'flex-basis: content` and `overflow-y: hidden` is malfunctioning..

PR #11112, opened 2019-06-18
Add message telling FTS is disabled when no toot can be found because of this

### LINKED ISSUE #11082: Add a meaningful message on the search results when the functionnality is disabled
### Pitch

Currently, admins can disable the search in toots and just keep the users/hashtags search.
If in the instance, the search feature is disabled, when a user looks for a word, the results list shows a message "0 results"
a screenshot of the results column

I suggest that with this message, in the column, another message warns the user that the search tool is disabled:
a screenshot showing a message "The search in toots content is disabled by your administrator." below the results found count

```html
<header class="search-results__header">
    <span aria-hidden="true" class="icon fa fa-search"></span><span>0 results</span>
    <br>
    <span style="font-style:italic;margin-top:5px;display:inline-block; font-size:12px">
        <span class="icon fa fa-info"></span>The search in toots content is disabled by your administrator.
    </span>
</header>
```

### Motivation

The fact that this is not explained is a problem because the user needs to ask to the admin why the search doesn't work to know why. Today I had to ask but it can be the same with everyone.

It's always a good thing to explain to the end user why something doesn't work, especially when it's a normal behavior of the app.

ClearlyClaire:
Fixes #11082

Made it so that it displays even if there are results, but only if the following conditions are met:
- There are no status results (as that can only result from URL fetching or FTS)
- FTS is actually disabled
- The search text would have been eligible for FTS

image

### DIFF
--- a/app/javascript/mastodon/actions/search.js
+++ b/app/javascript/mastodon/actions/search.js
@@ -48,7 +48,7 @@ export function submitSearch() {
         dispatch(importFetchedStatuses(response.data.statuses));
       }
 
-      dispatch(fetchSearchSuccess(response.data));
+      dispatch(fetchSearchSuccess(response.data, value));
       dispatch(fetchRelationships(response.data.accounts.map(item => item.id)));
     }).catch(error => {
       dispatch(fetchSearchFail(error));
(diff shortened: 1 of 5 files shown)


ykzts (approved):

Gargron:
I'm in favour of this but feel the wording could be better and not sure the "i" icon adds much.

Gargron:
"Searching toots by their contents is not enabled on this server of Mastodon"

ThibG:
Changed to “Searching toots by their content is not enabled on this Mastodon server.”
Removed the icon.

Gargron (approved):

PR #11121, opened 2019-06-19
Add label for admin theme selector

alixrossi:
Fixes the "theme" label in Administration/Site settings not having a translation string, and thus falling back to `human_attribute_name`. As an example:
image

### DIFF
--- a/app/views/admin/settings/edit.html.haml
+++ b/app/views/admin/settings/edit.html.haml
@@ -9,7 +9,7 @@
 
   .fields-row
     .fields-row__column.fields-row__column-6.fields-group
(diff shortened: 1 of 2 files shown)


Gargron (commented):
What about setting_theme? I am confused

Gargron:
Oh this is for the admin settings, I see. Wouldn't it be easier to make the form explicitly use setting_theme label though? To avoid adding another translation

palindromordnilap:
Oh, right. I considered it, but only as directly replacing :theme with :setting_theme and refactoring accordingly, I didn't think of just changing the label.

Gargron (approved):

PR #11123, opened 2019-06-19
Add audio uploads

### LINKED ISSUE #4827: Allow users to post audio files

Gargron:
Fix #4827

Accept uploads of OGG, WAV, FLAC, OPUS and MP3 files, and converts them to OGG. Media attachments get a new `audio` type. In the UI, audio uploads are displayed identically to video uploads.

___

This is a second take on #9480 because the audio player design was giving me grief and it was ambigous whether or not we need to parse, save, and edit special information about the files, like title, artist, album, etc. An audio player looks barren without those things, but they introduce more complexity, require a system-level dependency (`libtag1-dev`), require deciding on subset of tags to duplicate into our own schema, and require a way to store cover art within the media attachment model.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -7,6 +7,8 @@ class MediaController < ApplicationController
 
   before_action :set_media_attachment
   before_action :verify_permitted_status!
+  before_action :check_playable, only: :player
+  before_action :allow_iframing, only: :player
 
   content_security_policy only: :player do |p|
     p.frame_ancestors(false)
@@ -18,8 +20,6 @@ class MediaController < ApplicationController
 
   def player
     @body_classes = 'player'
-    response.headers['X-Frame-Options'] = 'ALLOWALL'
-    raise ActiveRecord::RecordNotFound unless @media_attachment.video? || @media_attachment.gifv? || @media_attachment.audio?
   end
 
   private
@@ -33,5 +33,13 @@ class MediaController < ApplicationController
   rescue Mastodon::NotPermittedError
     # Reraise in order to get a 404 instead of a 403 error code
     raise ActiveRecord::RecordNotFound
+  end
+
+  def check_playable
+    not_found unless @media_attachment.larger_media_format?
+  end
+
+  def allow_iframing
+    response.headers['X-Frame-Options'] = 'ALLOWALL'
   end
 end
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -19,7 +19,7 @@ class MediaController < ApplicationController
   def player
     @body_classes = 'player'
     response.headers['X-Frame-Options'] = 'ALLOWALL'
-    raise ActiveRecord::RecordNotFound unless @media_attachment.video? || @media_attachment.gifv?
+    raise ActiveRecord::RecordNotFound unless @media_attachment.video? || @media_attachment.gifv? || @media_attachment.audio?
   end
 
   private
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -334,16 +334,16 @@ class Status extends ImmutablePureComponent {
           />
         );
       } else if (['video', 'audio'].includes(status.getIn(['media_attachments', 0, 'type']))) {
-        const video = status.getIn(['media_attachments', 0]);
+        const attachment = status.getIn(['media_attachments', 0]);
 
         media = (
           <Bundle fetchComponent={Video} loading={this.renderLoadingVideoPlayer} >
             {Component => (
               <Component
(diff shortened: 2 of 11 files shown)


app/javascript/mastodon/components/status.js
```diff
@@ -333,7 +333,7 @@ class Status extends ImmutablePureComponent {
             media={status.get('media_attachments')}
           />
         );
-      } else if (status.getIn(['media_attachments', 0, 'type']) === 'video') {
+      } else if (['video', 'audio'].includes(status.getIn(['media_attachments', 0, 'type']))) {
         const video = status.getIn(['media_attachments', 0]);
```
nightpool: mb rename video to something more generic?

nightpool: attachment?

app/views/stream_entries/_detailed_status.html.haml
```diff
@@ -27,7 +27,7 @@
           = render partial: 'stream_entries/poll', locals: { status: status, poll: status.preloadable_poll, autoplay: autoplay }
 
   - if !status.media_attachments.empty?
-    - if status.media_attachments.first.video?
+    - if status.media_attachments.first.video? || status.media_attachments.first.audio?
```
nightpool: let's put this into a method on MediaAttachment, like `use_video_player?`

app/views/stream_entries/_simple_status.html.haml
```diff
@@ -31,7 +31,7 @@
           = render partial: 'stream_entries/poll', locals: { status: status, poll: status.preloadable_poll, autoplay: autoplay }
 
   - if !status.media_attachments.empty?
-    - if status.media_attachments.first.video?
+    - if status.media_attachments.first.video? || status.media_attachments.first.audio?
```
nightpool: as above

app/services/post_status_service.rb
```diff
@@ -100,7 +100,7 @@ def validate_media!
 
     @media = @account.media_attachments.where(status_id: nil).where(id: @options[:media_ids].take(4).map(&:to_i))
 
-    raise Mastodon::ValidationError, I18n.t('media_attachments.validations.images_and_video') if @media.size > 1 && @media.find(&:video?)
+    raise Mastodon::ValidationError, I18n.t('media_attachments.validations.images_and_video') if @media.size > 1 && @media.find { |x| x.video? || x.audio? }
```
nightpool: this should be `larger_media_format?` right?

app/controllers/media_controller.rb
```diff
@@ -19,7 +19,7 @@ def show
   def player
     @body_classes = 'player'
     response.headers['X-Frame-Options'] = 'ALLOWALL'
-    raise ActiveRecord::RecordNotFound unless @media_attachment.video? || @media_attachment.gifv?
+    raise ActiveRecord::RecordNotFound unless @media_attachment.video? || @media_attachment.gifv? || @media_attachment.audio?
```
nightpool: can/should this also be `larger_media_format?`

nightpool (changes requested):

nightpool (approved):
LGTM

koyuawsmbrtn:
Sick! :tada:

tn5421:
Is there any way we can simply embed the file as-is, without transcoding lossy source files to another format?

Gargron:
No, it won't be playable on all platforms then

tn5421:
Perhaps an option could be included in a users settings that allows them to override this default setting, with a warning given to the tune of your reply.

if I upload a 96kbps opus file and it gets transcoded to 192kbps mp3 it's going to sound like a 96kbps mp3 regardless of the extra file size

Gargron:
We use variable bitrate so I don't think the file will be bigger if the source isn't higher quality

koyuawsmbrtn:
Suggestion: Make the audio player smaller like in the previous PR

trwnh:
@tn5421 96kbps opus and 96kbps mp3 are not the same at all -- the 96kbps opus will sound vastly better since the opus codec is more efficient than mp3. also as stated, with variable bitrate the bitrate of a section of audio is allowed to dip down. there is still a quality loss with lossy->lossy transcode, but not so much that it makes the audio unusable.

what could be done is, if you are uploading an mp3 and it is already below a certain size or bitrate, then the transcode is skipped.

Gargron:
> what could be done is, if you are uploading an mp3 and it is already below a certain size or bitrate, then the transcode is skipped.

Nothing needs to be done. That's how it is

PR #11139, opened 2019-06-21
Remove expensive counters from federation page in admin UI

Gargron:


nightpool:
is this necessary? it's a pretty niche endpoint and the numbers are useful

Gargron:
I think it's annoying how long the domain blocks page loads. I added the numbers because I initially thought the same as you, but domain blocks are visited when you have a goal of either confirming something is blocked or adding a block and you don't want to take longer than necessary on either of those tasks.

Gargron:
Mind that for the "list of all known servers", the count has no effect on performance because the original query does a seqscan on the whole table anyway. So in that view, I keep the account counter, but replacing the "avatar stack" with it. The avatar stack was pretty but not worth it for the extra time it took. Yeah, it was cached, but you don't usually return to that page often so it's the first load that matters.

nightpool:
oh I missed that the account counter was still there.

nightpool (approved):

PR #11140, opened 2019-06-21
Compatibility with less friendly activitypub setups

throwaway-boop:
Some activity pub instances have seen fit to mess with the contexts. Since Mastodon fails to reject these obviously broken objects that do not have valid contexts, I've added a quick patch to check for them. [Example of a broken one, note the context field](https://develop.gab.com/users/robcolbert.json)

### DIFF
--- a/app/helpers/jsonld_helper.rb
+++ b/app/helpers/jsonld_helper.rb
@@ -40,7 +40,7 @@ module JsonLdHelper
   end
 
   def supported_context?(json)
-    !json.nil? && equals_or_includes?(json['@context'], ActivityPub::TagManager::CONTEXT)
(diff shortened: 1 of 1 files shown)


throwaway-boop:
Ah heck, I appear to have bungled the check. Oh well, I assume someone can get what I'm getting at and fix it for me. My Ruby is rusty.

m4sk1n:
Don’t think hardcoding domains is a good idea… And I don’t think they will merge anything related to Gab XD

Gargron:
There is no `gab` property in the ActivityPub spec

trwnh:
You could try to validate the `@context` by expanding and compacting JSON-LD, then noting if certain properties no longer resolve correctly due to improper namespacing. :thinking:

nightpool:
you don't need to expand and then compact, just a normal compact operation
as described by JSON-LD 1.1 should be enough. i'm tempted to do it just to
piss off gab. unfortunately it would only break pinned posts and custom
emoji :(

PR #11162, opened 2019-06-24
Remove flags from version.rb; just use pre

clarfonthey:
It appears that the `pre` variable is never used, and that the `flags` variable is used for RC suffixes. This removes the flags variable and uses `pre` for this purpose instead.

### DIFF
--- a/lib/mastodon/version.rb
+++ b/lib/mastodon/version.rb
@@ -20,16 +20,12 @@ module Mastodon
       nil
     end
 
-    def flags
-      ''
-    end
-
     def to_a
       [major, minor, patch, pre].compact
     end
 
     def to_s
-      [to_a.join('.'), flags].join
+      [[major, minor, patch].join('.'), pre].join
     end
 
     def repository


nightpool:
`#flags` is used by a lot of forks, removing it sounds like needless merge conflicts

clarfon:
Examples of fork like this? Glitch adds a separate suffix variable because `flags` is used for RCs, and this would basically use the `pre` variable for this purpose instead of `flags`.

Although if you can find a fork that uses both `pre` and `flags` it would make sense to keep it.

nightpool:
hm, you're right, i'm mis-remembering. it still seems weird to remove it and not `pre` though

clarfon:
The main reason is because `pre` is only used for RCs in tootsuite, and I'd rather have forks add their own variables (like Glitch does with the `suffix` variable) that they can mess with than modify the version numbers that are used in the vanilla version uses.

Gargron:
> The main reason is because `pre` is only used for RCs in tootsuite

I'm confused, we never used `pre`, the `"rc1"` string is put inside `flags`

clarfon:
Sorry, I should clarify:

I checked the git history, and `pre` was originally for RCs, and only RCs. This was changed to `flags` because it doesn't add an extra dot (i.e. `2.9.0rc1` instead of `2.9.0.rc1`). The original `pre` variable was kept.

Separate forks seem to have followed the route that glitch.social uses, which is adding separate variables for their own version changes rather than using `flags` directly. This avoids merge conflicts when updating to the latest Mastodon version.

Because of this, it seems incredibly confusing to have both a `flags` and a `pre` variable. This PR opts to combine these two into a single variable, `pre`, which does not add an extra dot in the version string.

This is, of course, under the assumption that `flags` is only used for RCs in the latest Mastodon versions, which seems to be the case.

Gargron:
Maybe we could add `flavour` method (default nil) that would get appended to the version string with a `-` when present? So when fully filled out it would generate `2.10.0rc1-foo`? I'm not sure if `pre` or `flags` is the more appropriate name to describe what `rc1` is.

clarfon:
Right now glitch uses a `suffix` method so perhaps that would be more appropriate? But I'm also for that as well. We could have `pre` be the RC and `suffix` include the rest.

Basically, I've been trying to think of how to standardise this for the Florence fork and wanted to come up with a reasonable way to detect flavours without having to mess too much with the code here.

PR #11179, opened 2019-06-25
Fix unnecessary SQL query performed on unauthenticated requests

Gargron:


ykzts (approved):

app/controllers/application_controller.rb
```diff
@@ -91,11 +91,15 @@ def use_seamless_external_login?
...
   def current_account
-    @current_account ||= current_user.try(:account)
+    return @current_account if defined?(@current_account)
+
+    @current_account = current_user&.account
   end
 
   def current_session
-    @current_session ||= SessionActivation.find_by(session_id: cookies.signed['_session_id'])
+    return @current_session if defined?(@current_session)
```
nightpool: why `if defined?` rather then `||=`? shouldn't they be equivalent?

Gargron: `||=` gets re-run if the previous value is nil, such as when the session was not found

PR #11187, opened 2019-06-26
Add support for M4V

mayaeh:
Some MP4 movie ( H.264/AAC )'s MIME type seem to be recognized as `video/x-m4v`.
In v2.9.2, it can't upload with the following error message:
`422 Validation failed: File content type is invalid, File is invalid.`

Before #11151 that format was also an uploadable format.
This change makes it possible to upload files of that type again.

I haven't yet conducted other formats of testing.

Please see also: https://en.wikipedia.org/wiki/M4V

### DIFF
--- a/app/models/media_attachment.rb
+++ b/app/models/media_attachment.rb
@@ -31,7 +31,7 @@ class MediaAttachment < ApplicationRecord
   AUDIO_FILE_EXTENSIONS = ['.ogg', '.oga', '.mp3', '.wav', '.flac', '.opus'].freeze
 
   IMAGE_MIME_TYPES             = ['image/jpeg', 'image/png', 'image/gif', 'image/webp'].freeze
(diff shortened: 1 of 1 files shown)


Gargron:
From the Wikipedia article:

> Some other video players can also recognize and play M4V files if the file extension is changed from ".m4v" to ".mp4".

Perhaps we need a special case for this format to ensure its mime type and extension remain mp4?

mayaeh:
Sorry, I don’t know what to do.

I was told from a friend, x-m4v does not appear on [the IANA’s media types list](https://www.iana.org/assignments/media-types/media-types.xhtml).

We may be better not to support x-m4v.

Should I close this PR?

Gargron:
This is what I meant: #11210

mayaeh:
Thank you !

PR #11190, opened 2019-06-26
Improve support of Audio, Video and Image AS objects

ClearlyClaire:
- If a `content` is available, display it after the link (typically, Peertube video descriptions)
- If the object's `url` points to some media we can use as attachment, do so

For Peertube videos, this does not change anything for videos above 40MB.
For videos smaller than that, it will download them and display them inline. I am not completely sure this is what we want, or if the 40MB cutoff is what we really want.

e.g, for https://tube.otter.sh/videos/watch/5dab27d6-1fff-46fb-9617-a16d26dba5c6 it renders
Screenshot_2019-06-26 Mastodon(2)

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -67,7 +67,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
         visibility: visibility_from_audience,
         thread: replied_to_status,
         conversation: conversation_from_uri(@object['conversation']),
-        media_attachment_ids: process_attachments.take(4).map(&:id),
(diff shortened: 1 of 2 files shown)


app/lib/activitypub/activity/create.rb
```diff
@@ -185,6 +185,43 @@ def process_emoji(tag)
...
+    as_array(@object['url']).each do |attachment|
+      if attachment.is_a?(String)
+        media_type = nil
+        size       = nil
+        href       = attachment
+        blurhash   = nil
+        focus      = nil
+      else
+        media_type = attachment['mediaType']
+        size       = attachment['size']&.to_i
```
ThibG: Note that `size` is definitely not standard, but it is what Peertube uses, and it lets us ignore bigger media easily

Issue #10679, opened 2019-05-03
[Feature request] Could polls we've voted in show which option we voted for?


Cassolotl:
### Pitch

![Screenshot 2019-05-03 at 12 03 27](https://user-images.githubusercontent.com/6607512/57133691-7f9f1000-6d9b-11e9-86d0-6192edf842b1.png)

Twitter shows how we voted with a little checkmark.

Yesterday and today I've been playing with polls, and [this happened](https://fosstodon.org/@RyuKurisu/102031458200017926):

![Screenshot 2019-05-03 at 12 04 27](https://user-images.githubusercontent.com/6607512/57133731-a3faec80-6d9b-11e9-8ae1-f23c1a897df0.png)

A checkmark would let people know that their vote has been registered, to avoid confusion like this happening again.

### Motivation

To avoid confusion over how one has voted and whether the vote has been correctly registered. It would be reassuring and avoid bug reports.

Cassolotl:
Thanks @ThibG! :)

PR #11195, opened 2019-06-27
Show user what options they have voted

ClearlyClaire:
I'm not sure about the API (it could make sense to have the property in the options itself) nor the interface design (I'm a bit worried about the extra spacing, especially when the user hasn't voted).

image


Cassolotl:
This fixes #10679. :) Thanks ThibG!

ThibG:
Updated to match the latest changes in the poll component design, and made it work for public pages as well.

Gargron (approved):

PR #11204, opened 2019-06-28
Fix expiration date of filters being set to “Never” when editing them

### LINKED ISSUE #9506: Custom filter edition page always show that the edited filter never expires
### Expected behaviour

When saving a filter with an expiry other than "never", clicking to edit the filter should show the original expiry that was selected when first saving the filter.

### Actual behaviour

Clicking to edit a filter shows the expiry "never" every time, no matter which expiry you selected and saved originally.

1) Make new filter with the expiry "one week".
2) Save.
3) In the filtered words list, click "edit" next to the filter you just saved.
4) Expiry is "never".

### Specifications

Vanilla Mastodon 2.6.1, https://queer.party.

fliter timing doesn t stick

ClearlyClaire:
When editing a custom filter, select the shortest preset duration that
still covers the remaining time of that filter.

Fixes #9506

### DIFF
--- a/app/models/custom_filter.rb
+++ b/app/models/custom_filter.rb
@@ -35,6 +35,13 @@ class CustomFilter < ApplicationRecord
   before_validation :clean_up_contexts
   after_commit :remove_cache
 
+  def expires_in
+    return @expires_in if defined?(@expires_in)
+    return nil if expires_at.nil?
+
+    [30.minutes, 1.hour, 6.hours, 12.hours, 1.day, 1.week].find { |expires_in| expires_in.from_now >= expires_at }
+  end
+
   private
 
   def clean_up_contexts


app/models/custom_filter.rb
```diff
@@ -35,6 +35,13 @@ class CustomFilter < ApplicationRecord
   before_validation :clean_up_contexts
   after_commit :remove_cache
 
+  def expires_in
+    return @expires_in if defined?(@expires_in)
+    return nil if expires_at.nil?
+
+    [30.minutes, 1.hour, 6.hours, 12.hours, 1.day, 1.week].filter { |expires_in| expires_in.from_now >= expires_at }.first
```
Gargron: Use `find` instead of `filter` and `first` (also we prefer `select` to `filter` (they are the same thing))

Gargron (changes requested):

Gargron (approved):

app/models/custom_filter.rb
```diff
@@ -35,6 +35,13 @@ class CustomFilter < ApplicationRecord
   before_validation :clean_up_contexts
   after_commit :remove_cache
 
+  def expires_in
+    return @expires_in if defined?(@expires_in)
```
mjankowski, line 39: I realize this PR is 5 years old, but bumped into this while doing something else ... it seems like this memoization is never actually used? (ie, this ivar is never assigned to anything) or am I missing something here?

ClearlyClaire: I don't think you're missing anything, I think this line is useless.

PR #11208, opened 2019-06-29
Fix hidden statuses losing focus

ClearlyClaire:
I am not too sure of the performance implications of keeping the `HotKeys` component though

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -278,12 +278,27 @@ class Status extends ImmutablePureComponent {
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
Why should a hidden status keep focus?

ThibG:
On chrome, you cannot keep scrolling with keyboard if the item looses focus. For some reason on Firefox this works, though.

ThibG:
Also, the status not being selectable makes the column number keyboard shortcuts not scroll to the very top, but to a seemingly-random point of the timeline (actually the first non-hidden status)

PR #11211, opened 2019-06-29
Optimize makeGetStatus

ClearlyClaire:
Because `ImmutableList.filter` always returns a new object and `createSelector`
memoizes based on object identity, the selector returned by `makeGetStatus`
would *always* execute.

To avoid that, we wrap `getFilters` into a new memoizer that memoizes based on
deep equality, thus returning the same object as long as the filters haven't
changed, allowing the memoization of `makeGetStatus` to work.

Furthermore, we memoize the compiled regexs instead of recomputing them each
time the selector is called.

### DIFF
--- a/app/javascript/mastodon/actions/notifications.js
+++ b/app/javascript/mastodon/actions/notifications.js
@@ -11,7 +11,7 @@ import { saveSettings } from './settings';
 import { defineMessages } from 'react-intl';
 import { List as ImmutableList } from 'immutable';
 import { unescapeHTML } from '../utils/html';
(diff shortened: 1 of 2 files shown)


ThibG:
Wait, no, this isn't actually what we want, because `contextType` isn't specific to the place where `makeGetFiltersRegex` is called. Ideally, `getFiltersRegex` should memoize on `contextType`.

ThibG:
Wrote a selector by hand, which memoize on server context type, thus ensuring optimal memoization of the filter regexps.

Gargron (approved):

PR #11214, opened 2019-06-29
Remove deprecated REST API `POST /api/v1/notifications/dismiss?id=:id`

Gargron:


ThibG:
Hm, why?

Gargron:
See changelog of 2.6.0. This API was a mistake from the start, it's supposed to be `POST /api/v1/notifications/:id/dismiss`

ThibG:
Fair enough!

PR #11219, opened 2019-07-01
Change domain block behaviour to prevent creation of accounts from suspended domains

Gargron:


app/services/activitypub/process_account_service.rb
```diff
@@ -15,6 +15,8 @@ def call(username, domain, json, options = {})
     @domain      = domain
     @collections = {}
 
+    return if auto_suspend?
```
ThibG: Maybe we should bail out only if the account isn't known already, and not when processing updates to already-known accounts? Tbh I'm not sure about the implications of this.

ClearlyClaire (approved):
LGTM. Not sure about refusing to process already-known accounts though

app/services/activitypub/process_account_service.rb
Gargron: If an account is suspended there is nothing on it to update. In fact it might make sense to delete account records from suspended domains (maybe except the ones that were suspended manually, to make sure those decisions are remembered after an undo).

PR #11241, opened 2019-07-05
Fix BackupService crashing when an attachment is missing

ClearlyClaire:
For various reasons such as admin error or out-of-sync media and
database backups, it might be possible for local attachments to be lost.

This commit allows the BackupService to continue its work even if some media
file is missing.

### DIFF
--- a/app/services/backup_service.rb
+++ b/app/services/backup_service.rb
@@ -142,5 +142,7 @@ class BackupService < BaseService
         io.write(buffer)
       end
     end
+  rescue Errno::ENOENT
+    Rails.logger.warn "Error making a backup for attachment #{attachment}"
   end
 end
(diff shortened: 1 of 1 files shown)


app/services/backup_service.rb
```diff
@@ -142,5 +142,7 @@ def download_to_tar(tar, attachment, filename)
         io.write(buffer)
       end
     end
+  rescue Errno::ENOENT
+    Rails.logger.warn "Error making a backup for attachment #{attachment}"
```
Gargron: Not sure if the string output of `attachment` is useful information..

ThibG: Hm, changed to display the filename (which really is the file path) instead

Gargron (approved):

PR #11245, opened 2019-07-06
Fix support for HTTP proxies

ClearlyClaire:
The `Request::Socket` class implements its own logic for connection timeouts, to give us a bit more control, and because `http.rb` uses the ill-advised `Timeout::timeout`. That latter mechanism is bypassed entirely. Therefore, *not* using our `Socket` class means we do not have proper handling of connection timeouts.

This PR changes that by using our custom class for both direct connections and connections through a proxy—in which case the socket class is used to connect to the *proxy*.

Additionally, since the socket is used to connect to the *proxy*, it will never get used with a hidden service as `host`, so removed the additional check.

Also added support for IP addresses in addition to hostnames.

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -181,47 +181,46 @@ class Request
         outer_e = nil
         port    = args.first
 
-        Resolv::DNS.open do |dns|
-          dns.timeouts = 5
-
-          addresses = []
-          begin
-            addresses = [IPAddr.new(host)]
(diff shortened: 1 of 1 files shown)


ThibG:
I haven't tested hidden services, but I have tested access with and without an HTTP proxy, as well as reverse-proxies failing to connect (this indeed fixed hanging processes).

Gargron (approved):

PR #11249, opened 2019-07-07
Refactor controllers for statuses, accounts, and more

Gargron:
- Remove `mark_cacheable!` as useless abstraction
- Use Active Model Serializers caching for ActivityPub JSON instead of custom solution
- Add `AccountOwnedConcern` to unify logic for account-owned controllers
- Replace repeated checks for public/unlisted visibility with `distributable?`
- Bug fix: Some controllers did not return 404 for accounts that have not yet been approved
- Bug fix: Remove link to Atom feed from account link header
- Bug fix: Make sure lock is displayed for limited visibility statuses

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -3,11 +3,11 @@
 class AboutController < ApplicationController
   layout 'public'
 
-  before_action :set_instance_presenter, only: [:show, :more, :terms]
+  before_action :set_body_classes, only: :show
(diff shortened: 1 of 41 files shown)


ykzts (approved):

Hanage999:
This commit prevents toots created in my server (mastodon.crazynewworld.net) from federating. BTs federate without problem. Would you take a look into this?

PR #11251, opened 2019-07-08
Refactor fetching of remote resources

Gargron:
- Remove OStatus-related logic from `FetchAtomService`
- Make `FetchAtomService` catch the same exceptions as `FetchLinkCardService`
- Make `ActivityPub::DeliveryWorker` and `JsonLdHelper` handle HTTP error codes the same way
- Bug fix: SSL errors not being caught by `FetchLinkCardService`
- Rename `FetchAtomService` to `FetchResourceService`


app/services/fetch_resource_service.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class FetchResourceService < BaseService
+  include JsonLdHelper
+
+  ACCEPT_HEADER = 'application/activity+json, application/ld+json; profile="https://www.w3.org/ns/activitystreams", application/atom+xml, text/html'
```
ThibG: Still requesting atom+xml?

spec/services/fetch_remote_account_service_spec.rb
```diff
@@ -4,6 +4,8 @@
   let(:url) { 'https://example.com/alice' }
   let(:prefetched_body) { nil }
   let(:protocol) { :ostatus }
+  let!(:representative) { Fabricate(:account) }
+
```
ThibG: Is that required? Isn't that leftover from when this PR included signing?

ClearlyClaire (approved):
LGTM, with two minor remarks

PR #11267, opened 2019-07-08
Changes to tootctl statuses remove

ClearlyClaire:
- Fix incorrect comments
- Keep statuses favourited by local users
- Do not exempt toots from being deleted just because they are replies
- Skip reblogs of local statuses instead of skipping *all* reblogs
- Use `Status.local` and `Status.uri` to check if toots are local instead of checking their author

There are most definitely more optimizations possible, and I'd like whole threads to be cleaned instead of just the leaves. I am not too sure how to do that efficiently (maybe doing the query in descending order on the ids would help?)

### DIFF
--- a/lib/mastodon/statuses_cli.rb
+++ b/lib/mastodon/statuses_cli.rb
@@ -36,11 +36,11 @@ module Mastodon
 
       Status.remote
             .where('id < ?', max_id)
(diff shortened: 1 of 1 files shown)


ThibG:
Well, I'm not sure if the optimization thing is really an optimization since there's no index on that, hm…

ThibG:
The rationale for boosts is that we probably aren't interested in boosts from people we aren't interested in anymore (no followers). Additionally, while current Mastodon versions will now reject them, Peertube used to forward every boost they received.

ThibG:
Also, I don't think a blanket exemption for toots from followed users is very useful either…

On the other hand, about things that get deleted and probably shouldn't, recent boosts of old toots will get affected too…

ykzts (approved):

PR #11269, opened 2019-07-08
Add ActivityPub secure mode

Gargron:
- Controlled by `SECURE_MODE=true`
- Requires incoming GET requests to ActivityPub resources to be signed
- Actors can still be fetched without signature, but return only technical attributes
- Bug fix: Handle inbox requests without request body gracefully

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -40,7 +40,7 @@ class AccountsController < ApplicationController
       end
 
       format.json do
-        expires_in 3.minutes, public: true unless secure_mode_enabled? && signed_request_account.present?
+        expires_in 3.minutes, public: !(authorized_fetch_mode? && signed_request_account.present?)
         render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter, fields: fields_selection
       end
     end
@@ -135,7 +135,7 @@ class AccountsController < ApplicationController
   end
 
   def fields_selection
-    if signed_request_account.present? || !secure_mode_enabled?
+    if signed_request_account.present? || !authorized_fetch_mode?
       # Return all fields
     else
       %i(id type preferred_username inbox public_key endpoints)
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -41,7 +41,7 @@ class AccountsController < ApplicationController
 
       format.json do
         expires_in 3.minutes, public: !(authorized_fetch_mode? && signed_request_account.present?)
-        render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter, fields: fields_selection
+        render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter, fields: restrict_fields_to
       end
     end
   end
@@ -134,7 +134,7 @@ class AccountsController < ApplicationController
     end
   end
 
-  def fields_selection
+  def restrict_fields_to
     if signed_request_account.present? || !authorized_fetch_mode?
       # Return all fields
     else
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -4,6 +4,7 @@ class AccountsController < ApplicationController
   PAGE_SIZE = 20
 
   include AccountControllerConcern
+  include SignatureAuthentication
 
   before_action :set_cache_headers
   before_action :set_body_classes
@@ -39,8 +40,8 @@ class AccountsController < ApplicationController
       end
 
       format.json do
-        expires_in 3.minutes, public: true
-        render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter
+        expires_in 3.minutes, public: true unless secure_mode_enabled? && signed_request_account.present?
+        render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter, fields: fields_selection
       end
     end
   end
@@ -130,6 +131,14 @@ class AccountsController < ApplicationController
       filtered_statuses.paginate_by_min_id(PAGE_SIZE, params[:min_id]).reverse
     else
       filtered_statuses.paginate_by_max_id(PAGE_SIZE, params[:max_id], params[:since_id]).to_a
+    end
+  end
+
+  def fields_selection
+    if signed_request_account.present? || !secure_mode_enabled?
+      # Return all fields
+    else
+      %i(id type preferred_username inbox public_key endpoints)
     end
   end
 end
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -135,7 +135,7 @@ class AccountsController < ApplicationController
   end
 
   def restrict_fields_to
-    if signed_request_account.present? || !authorized_fetch_mode?
+    if signed_request_account.present? || public_fetch_mode?
       # Return all fields
     else
       %i(id type preferred_username inbox public_key endpoints)
--- a/app/controllers/activitypub/collections_controller.rb
+++ b/app/controllers/activitypub/collections_controller.rb
@@ -4,12 +4,13 @@ class ActivityPub::CollectionsController < Api::BaseController
   include SignatureVerification
   include AccountOwnedConcern
 
+  before_action :require_signature!, if: :secure_mode_enabled?
   before_action :set_size
   before_action :set_statuses
   before_action :set_cache_headers
 
   def show
-    expires_in 3.minutes, public: true
+    expires_in 3.minutes, public: true unless secure_mode_enabled?
     render json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, skip_activities: true
   end
 
--- a/app/controllers/activitypub/collections_controller.rb
+++ b/app/controllers/activitypub/collections_controller.rb
@@ -4,13 +4,13 @@ class ActivityPub::CollectionsController < Api::BaseController
   include SignatureVerification
   include AccountOwnedConcern
 
-  before_action :require_signature!, if: :secure_mode_enabled?
+  before_action :require_signature!, if: :authorized_fetch_mode?
   before_action :set_size
   before_action :set_statuses
   before_action :set_cache_headers
 
   def show
-    expires_in 3.minutes, public: true unless secure_mode_enabled?
+    expires_in 3.minutes, public: !authorized_fetch_mode?
     render json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, skip_activities: true
   end
 
--- a/app/controllers/activitypub/collections_controller.rb
+++ b/app/controllers/activitypub/collections_controller.rb
@@ -10,7 +10,7 @@ class ActivityPub::CollectionsController < Api::BaseController
   before_action :set_cache_headers
 
   def show
-    expires_in 3.minutes, public: !authorized_fetch_mode?
+    expires_in 3.minutes, public: public_fetch_mode?
     render json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, skip_activities: true
   end
 
--- a/app/controllers/activitypub/inboxes_controller.rb
+++ b/app/controllers/activitypub/inboxes_controller.rb
@@ -5,23 +5,24 @@ class ActivityPub::InboxesController < Api::BaseController
   include JsonLdHelper
   include AccountOwnedConcern
 
+  before_action :skip_unknown_actor_delete
+  before_action :require_signature!
+
   def create
-    if unknown_deleted_account?
-      head 202
-    elsif signed_request_account
-      upgrade_account
-      process_payload
-      head 202
-    else
-      render plain: signature_verification_failure_reason, status: 401
-    end
+    upgrade_account
+    process_payload
+    head 202
   end
 
   private
 
+  def skip_unknown_actor_delete
+    head 202 if unknown_deleted_account?
+  end
+
   def unknown_deleted_account?
     json = Oj.load(body, mode: :strict)
-    json['type'] == 'Delete' && json['actor'].present? && json['actor'] == value_or_id(json['object']) && !Account.where(uri: json['actor']).exists?
+    json.is_a?(Hash) && json['type'] == 'Delete' && json['actor'].present? && json['actor'] == value_or_id(json['object']) && !Account.where(uri: json['actor']).exists?
   rescue Oj::ParseError
     false
   end
@@ -32,8 +33,12 @@ class ActivityPub::InboxesController < Api::BaseController
 
   def body
     return @body if defined?(@body)
-    @body = request.body.read.force_encoding('UTF-8')
+
+    @body = request.body.read
+    @body.force_encoding('UTF-8') if @body.present?
+
     request.body.rewind if request.body.respond_to?(:rewind)
+
     @body
   end
 
--- a/app/controllers/activitypub/outboxes_controller.rb
+++ b/app/controllers/activitypub/outboxes_controller.rb
@@ -11,7 +11,7 @@ class ActivityPub::OutboxesController < Api::BaseController
   before_action :set_cache_headers
 
   def show
-    expires_in page_requested? ? 0 : 3.minutes, public: !authorized_fetch_mode?
+    expires_in(page_requested? ? 0 : 3.minutes, public: !authorized_fetch_mode?)
     render json: outbox_presenter, serializer: ActivityPub::OutboxSerializer, adapter: ActivityPub::Adapter, content_type: 'application/activity+json'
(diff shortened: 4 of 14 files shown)


Gargron:
One thing I am not sure about is the naming of the environment variable and derivative methods. Secure mode sounds too generic. On the other hand, "require http signatures for activitypub resources" sounds too specific (and long).

kaniini:
cwebber and i have been working on integrating OCAP into ActivityPub for the past several months, and we are getting close.  i think it makes better sense to use OCAP to authorize the fetches instead of signatures.  no objection to allowing signatures as a temporary bandaid though.

thebaer:
So I understand what this is for, is the idea just to restrict data access? I.e. a signature requirement means an instance can choose whether or not to grant accounts read privileges?

Either way, is there a way to discover whether this is enabled or not from the outside world? Or should platforms like WriteFreely always send GET requests with a signature now?

schmittlauch:
> One thing I am not sure about is the naming of the environment variable and derivative methods. Secure mode sounds too generic. On the other hand, "require http signatures for activitypub resources" sounds too specific (and long).

What about `AUTHORIZED_FETCH`?

Gargron:
> Either way, is there a way to discover whether this is enabled or not from the outside world? Or should platforms like WriteFreely always send GET requests with a signature now?

I'm not sure on the first part, and on the second, even Mastodon is not currently signing all outgoing requests, so there are definitely considerations about how to proceed without breaking stuff, which is why it's going to be behind an off-by-default environment variable for now. There are actually two problems here:

1. Current servers not signing requests
2. High-performance proxy caching not being available for authorized fetches

That being said... Including a signature header with a request is basically free (public resources that don't check HTTP signatures will just ignore it) so I don't see a disadvantage to WriteFreely doing that.

BenLubar:
> 2. High-performance proxy caching not being available for authorized fetches

This is actually already the case for me, because Cloudflare doesn't support `Vary` headers other than `Accept-Encoding`. If I have public caching mode enabled, Cloudflare will send HTML responses to ActivityStreams requests and vice versa.

Gargron:
@BenLubar However, nginx supports it, and you probably still have nginx between puma and cloudflare! This is what would no longer be possible

BenLubar:
> @BenLubar However, nginx supports it, and you probably still have nginx between puma and cloudflare! This is what would no longer be possible

sending `Vary: Signature` should be enough to make nginx handle this gracefully

Gargron:
It would be pointless though, each signature is different. By definition you can't use the same one multiple times (past a threshold time window) to prevent replay attacks.

app/controllers/activitypub/collections_controller.rb
```diff
@@ -4,12 +4,13 @@ class ActivityPub::CollectionsController < Api::BaseController
...
   include AccountOwnedConcern
 
+  before_action :require_signature!, if: :secure_mode_enabled?
   before_action :set_size
   before_action :set_statuses
   before_action :set_cache_headers
 
   def show
-    expires_in 3.minutes, public: true
+    expires_in 3.minutes, public: true unless secure_mode_enabled?
```
BenLubar: `expires_in` can still be set if public is set to false because `private` means "no shared caches"

app/controllers/activitypub/inboxes_controller.rb
```diff
@@ -5,23 +5,24 @@ class ActivityPub::InboxesController < Api::BaseController
...
+    upgrade_account
+    process_payload
+    head 202
   end
 
   private
 
+  def skip_unknown_actor_delete
+    head 202 if unknown_deleted_account?
+  end
```
ThibG: Will this prevent other hooks and methods from running and returning an error?

ThibG: (Also, everything in this file is orthogonal to the purpose of this PR)

ThibG:
Seems good to me overall, although I'm very interested with the token-based approach discussed by kaniini, as that would make it far easier for a caching reverse-proxy to cooperate.

Also, we should make sure to properly document the new setting in `.env.production.sample` at the very least.

app/controllers/activitypub/inboxes_controller.rb
Gargron: When a `before_action` returns a response (e.g. `redirect_to` or `render` or `head`) it stops the execution chain

app/controllers/activitypub/collections_controller.rb
Gargron: Okay

Gargron:
I see no point in adding it to `.env.production.sample` until signing is implemented

kaniini:
In Pleroma, we plan on using an instance-specific actor to represent the fetches for this mitigation.

Of course, we also plan to use bearcaps for the long term, but that's something we talk about later :)

Point is, for now, will an instance-wide actor work?

Gargron:
> Point is, for now, will an instance-wide actor work?

Sure. I'm not sure why you think it wouldn't, if it's just a typical actor.

kaniini:
My main question is will the access control be enforced directly (fetching an object requiring a signature of someone authorized to fetch the object) or indirectly (any actor on the same instance will work)?

Gargron:
Ah. Well, if a post is public, then any actor can fetch it, provided the actor is not domain-blocked. If a post is private, then only an authorized follower can fetch it.

app/controllers/accounts_controller.rb
```diff
@@ -132,4 +133,12 @@ def filtered_status_page(params)
       filtered_statuses.paginate_by_max_id(PAGE_SIZE, params[:max_id], params[:since_id]).to_a
     end
   end
+
+  def fields_selection
+    if signed_request_account.present? || !authorized_fetch_mode?
```
nightpool: this might be more clear with early returns?

```ruby
return unless authorized_fetch_mode?
return if signed_request_account.present?
%i(id type preferred_username inbox public_key endpoints)
```

app/controllers/accounts_controller.rb
```diff
@@ -132,4 +133,12 @@ def filtered_status_page(params)
       filtered_statuses.paginate_by_max_id(PAGE_SIZE, params[:max_id], params[:since_id]).to_a
     end
   end
+
+  def fields_selection
```
nightpool: `restrict_fields_to`

app/controllers/accounts_controller.rb
nightpool: also i'm unclear what the point of returning a limited object is... mind explaining?

app/controllers/accounts_controller.rb
```diff
@@ -39,8 +40,8 @@ def show
       end
 
       format.json do
-        expires_in 3.minutes, public: true
-        render json: @account, content_type: 'application/activity+json', serializer: ActivityPub::ActorSerializer, adapter: ActivityPub::Adapter
+        expires_in 3.minutes, public: !(authorized_fetch_mode? && signed_request_account.present?)
```
nightpool: we should add `Vary: Signature` to the response too right? Otherwise you might get served a `public` response for a signed request. Do you know the best way to do that? Might be best to just make this `!authorized_fetch_mode?`

app/controllers/activitypub/collections_controller.rb
```diff
@@ -4,12 +4,13 @@ class ActivityPub::CollectionsController < Api::BaseController
...
   include AccountOwnedConcern
 
+  before_action :require_signature!, if: :authorized_fetch_mode?
   before_action :set_size
   before_action :set_statuses
   before_action :set_cache_headers
 
   def show
-    expires_in 3.minutes, public: true
+    expires_in 3.minutes, public: !authorized_fetch_mode?
```
nightpool: might be nice to have a readable helper method for `!authorized_fetch_mode?`

app/controllers/activitypub/outboxes_controller.rb
```diff
@@ -6,12 +6,12 @@ class ActivityPub::OutboxesController < Api::BaseController
...
   include AccountOwnedConcern
 
+  before_action :require_signature!, if: :authorized_fetch_mode?
   before_action :set_statuses
   before_action :set_cache_headers
 
   def show
-    expires_in 1.minute, public: true unless page_requested?
-
+    expires_in page_requested? ? 0 : 3.minutes, public: !authorized_fetch_mode?
```
nightpool: while personally I appreciate ruby's lack of punctuation, others might find `expires_in(page_requested? 0 : 3.minutes, public: !authorized_fetch_mode?)` easier to read

app/controllers/concerns/account_controller_concern.rb
```diff
@@ -11,7 +11,7 @@ module AccountControllerConcern
     layout 'public'
 
     before_action :set_instance_presenter
-    before_action :set_link_headers
+    before_action :set_link_headers, if: -> { request.format.nil? || request.format == :html }
```
nightpool: this might be clearer/more maintainable if it was an after_action so it could read the actual response type? I don't think I've ever actually seen someone use an after action so I have no clue whether this is a good idea or not but it seems better

app/controllers/follower_accounts_controller.rb
```diff
@@ -17,9 +19,9 @@ def index
       end
 
       format.json do
-        raise Mastodon::NotPermittedError if params[:page].present? && @account.user_hides_network?
+        raise Mastodon::NotPermittedError if page_requested? && @account.user_hides_network?
 
-        expires_in 3.minutes, public: true if params[:page].blank?
+        expires_in page_requested? ? 0 : 3.minutes, public: !authorized_fetch_mode?
```
nightpool: what's the point of changing the expires_in when a page is requested?

app/controllers/follower_accounts_controller.rb
```diff
@@ -2,7 +2,9 @@
 
 class FollowerAccountsController < ApplicationController
   include AccountControllerConcern
+  include SignatureVerification
 
+  before_action :require_signature!, if: -> { request.format == :json && authorized_fetch_mode? }
```
nightpool: feels deeply weird to me that we would allow people to scrape this info from the html but refuse to show it in a machine-readable format.

nightpool (changes requested):
- consider allowing authenticated local users to view activitypub resources as well (in the browser or perhaps using an api client)
 - consider whether we're comfortable showing in the browser things we're not presenting in a machine readable format. Maybe this should affect unauthenticated viewing of public profile pages/status pages as well?

app/controllers/accounts_controller.rb
Gargron: Assuming that your goal is to prevent blocked instances from obtaining any data from yours, profile info is just as important as individual statuses. Bio text, header image, profile picture--all of that is quite personal. The attributes we do return here are the bare minimum, and technical in nature, to allow a two-way cryptography to be established.

app/controllers/accounts_controller.rb
Gargron: I did do that and then @BenLubar told me I should do this. If the cache header says "private" it's not supposed to be shared anyway..?

app/controllers/accounts_controller.rb
nightpool: ah, right, a chicken and an egg sort of problem. we need to return the limited object so that we can authenticate to people who don't know who we are. hmm.

app/controllers/follower_accounts_controller.rb
Gargron: When no page is required, what we return is essentially a counter--and these counters are requested when we're fetching actor info (so that endpoint can be hammered quite a bit). When a page is requested, it returns actual data, and I guess we don't want actual data to be outdated.

app/controllers/accounts_controller.rb
nightpool: Right, but if the sequence of requests goes like this:

- client A requests the account without authentication and gets the limited account response (public: true)
- The intermediary cache caches this response publicly
- client B requests the account *with* authentication and the cache responds with the limited account response.

then that's bad

app/controllers/follower_accounts_controller.rb
Gargron: Stuff doesn't federate over HTML pages.

app/controllers/concerns/account_controller_concern.rb
Gargron: This is a minor detail and it works as it is though

app/controllers/accounts_controller.rb
Gargron: Also no I would find two early returns harder to read in this case. "If we're signed in, or we don't require sign in, do this" reads more naturally to me.

app/controllers/activitypub/outboxes_controller.rb
Gargron: Done

app/controllers/activitypub/collections_controller.rb
Gargron: I'm not sure I agree a simple negation warrants that. `not_authorized_fetch_mode?` would just be longer.

app/controllers/accounts_controller.rb
Gargron: Okay, Vary header added.

nightpool (approved):

app/controllers/accounts_controller.rb
```diff
@@ -132,4 +133,12 @@ def filtered_status_page(params)
       filtered_statuses.paginate_by_max_id(PAGE_SIZE, params[:max_id], params[:since_id]).to_a
     end
   end
+
+  def fields_selection
+    if signed_request_account.present? || !secure_mode_enabled?
```
kaniini: This makes me nervous.  Pleroma's Service actors aren't known via WebFinger.  Will fetches have to be signed by an actor that Mastodon can WebFinger?  If so, is it possible to change that to be pure AP?

Gargron: Did you mean to comment on this particular line? It controls what JSON attributes *we* expose

Separately from this, we *do* require webfinger to process an actor/key, currently. We need to investigate what it would mean to not require that in terms of uniqueness/authority guarantees.

BenLubar: does Pleroma at least support webfinger for service actors via URLs (that is, `https:`, not `acct:`)?

app/controllers/follower_accounts_controller.rb
BenLubar: There's no way to have people able to access HTML pages without auth and also prevent scraping. Mastodon already supports not showing followers publicly, and unless federated followers lists becomes a thing in Mastodon, the HTML won't need a separate setting.

app/controllers/accounts_controller.rb
trwnh: >we do require webfinger to process an actor/key, currently. We need to investigate what it would mean to not require that in terms of uniqueness/authority guarantees.

wouldn't it be the same? webfinger queries a domain for a resource, but having a direct url should effectively skip the webfinger step, since you already know the resource.

Gargron: Webfinger is host-wide by spec while there is no guarantee the URL path space of a host is not shared by independent entities that could try to impersonate each other... At least I believe that was the original argumentation

kaniini: the security model that makes most sense for the fediverse is that whatever software is running at the root of a domain has control over all keys associated with that domain.  while it is possible that some implementations don't work like this, those implementations don't presently exist (at least not in the mainstream).

thusly, it does not really matter if entities in the path space can impersonate each other, obviously they can, because the software itself controls the keys anyway.

the reason why Pleroma does not publish it's service actors over WebFinger is because we do not want people interacting with them.  they are meant to be used for internal transactions (like relaying presently), and for cases such as handling whatever authorization token is required for fetching remote objects (presently signatures, but ideally bearcaps in the near future) :)

kaniini: as for why i selected this line, it's just the line that brought this thought up about how the signatures will be enforced.

it is pointless for mastodon to try to enforce access control based on what key belonging to a remote host is used to fetch the object: once the remote peer holds the object, it can either respect your wishes, or not.  thusly, i believe it makes most sense to simply never have the metadata exposure in the first place and sign *all* fetches with a designated instance-wide actor (such as `Account.representative`).  this makes the implementation simpler and more robust and doesn't make the implementation try to do things that are ultimately irrelevant anyway.

the question is not whether the software trusts a given user on an instance, but whether the software trusts the instance itself.

ThibG:
That is true in practice, however:
1. I really dislike relying on this instance-centric view in non-obvious ways, especially as this is completely alien to the current specs
2. There *are* cases we want private toots to be fetchable, and trying to figure out whether the instance has an actor that is allowed to access it seems less obvious and natural than just checking the requesting author

kaniini:
I'm only talking about public posts in this context.

But if you want to go *there*, *those* posts should also be fetched using a negotiated bearcap, which could be issued in the `Accept Follow` message sent back to the user when a follow gets ACKed.

I see this signature stuff strictly as a bandaid, so I don't think it really matters that the bandaid is alien to what is in the current spec.  It's a mitigation, with a limited life cycle.

kaniini:
(and to be clear if you want to fetch private posts from Pleroma, we can talk about doing that, but I only want to talk about doing that with a bearcap.  again, I see this as a mitigation and only a mitigation.)

kaniini:
a thought I had on removing the webfinger requirement for signatures would be to add an `actor` field to the signature itself.  then the actor can be fetched and the signature can be checked against keys belonging to the actor.  this solves impersonation nicely: actor A says he signed it, that attestation is provable with the signature.

anotak:
hey, is this documented anywhere? what are the drawbacks of having this on? i assume there are some, right, because why is it not on by default?

PR #11271, opened 2019-07-09
Fix activity being rendered within activity due to caching

### LINKED ISSUE #11270: Toots stopped showing up on remote instances
### Expected behaviour

Public toots and replies appear timelines for an account on a remote server who follows the source account.

### Actual behaviour

After merging #11249, toots no longer appears on remote servers. Local timelines are not affected.

1. Merge #11249 and deploy it to Server A.
1. Login to Server A and post a toot to an account on a remote server, e.g., `@user-b@server-b hello`
4. Login to Server B as user-b, check the timeline

### Specifications
Versions:
- NG: 1e7187f2a8e0b9ffe4e7d6b06e9f70674c50471e
- OK: 1e7187f2a8e0b9ffe4e7d6b06e9f70674c50471e with 63c7fe8e4892b22e80c015bf0ecb04496318623b reverted

Looking through app logs, it seems that the toots are POSTed successfully to the remote server. @mayaeh [seems to have found](https://mastodon.crazynewworld.net/@hanage999/102410535013768500) that the toot is `bundle: Rejected Create activity` by the remote server receiving toots.

Gargron:
Fix #11270

### DIFF
--- a/app/serializers/activitypub/activity_serializer.rb
+++ b/app/serializers/activitypub/activity_serializer.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class ActivityPub::ActivitySerializer < ActivityPub::Serializer
-  cache key: 'activity', expires_in: 3.minutes
-
   attributes :id, :type, :actor, :published, :to, :cc
 
   has_one :proper, key: :object, serializer: ActivityPub::NoteSerializer, if: :serialize_object?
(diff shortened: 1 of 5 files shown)


nightpool:
image

zunda:
I`ve merged this to  1e7187f and seems to have fixed the issue - public toot and a reply showed up on a remote server :tada:

ThibG:
Doesn't that mean that if we don't rely on a caching reverse proxy, we do not have any cache whatsoever for public statuses? Wouldn't that badly affect performances?

PR #11272, opened 2019-07-09
Add statuses from bots filtering

mkljczk:
i think this simple change makes a lot of sense


Cassolotl:
Plz 😥 Merrrrge

m4sk1n:
🤔

ykzts (approved):

Cassolotl:
I just wanted to comment to try and bump this up the attention ladder, because I really want to hang out in the federated timeline and boost cool stuff out to other instances but I always end up getting frustrated and giving up, because on my smaller instance it's like 50% bots and I have to either sit there muting them all individually (and they're like the hydra) or I have to pay more attention to mentally filter out things that weren't posted by real people.

umonaca:
Seems like a duplicate of pull request #10718

mkljczk:
Not really, mine is just a front-end change, which is enough, I think

umonaca:
The drawback of this PR is that the reblogged post from bot accounts that the current user does not follow are not properly filtered out. Suppose A is a normal user account, and B is another normal user account. A follows B, and A does not follow C, which is a bot account. Then B reblogged C. That bot post will appear in A's home timeline.

umonaca:
Another problem:
When the timeline contains many toots, it takes a long time to process when switching off the button, even on a PC browser.

umonaca:
Another bug:
If the user switches off the button, when the next page only contains toots from bots, the `load more` button will not work. I think this bug may exist in the upstream `show replies` and `shoot boosts` as well.

jordankittle:
I would like the option to filter bots as well

PR #11283, opened 2019-07-10
[security] lodash issue identified by snyk.io

ghost:
SNYK-JS-LODASH-450202

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,414 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.13.5
+ignore: {}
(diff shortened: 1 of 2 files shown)


Gargron:
Thanks but we use dependabot, merged with #11287

LemonAndroid:
@Gargron "Thanks, we don't need lemons, we eat apples"

LemonAndroid:
@Gargron How does the security feature of dependabot.com work?

From what I could see, snyk.io is keeping a database of known vulnerabilities and looking if any of the packages described in the `Gemfile`, `package.json` *or* any of the dependencies of those packages are affected.

PR #11284, opened 2019-07-10
Add HTTP signatures to all outgoing ActivityPub GET requests

Gargron:
This was a lot simpler than I thought it would be. All ActivityPub fetches are using `fetch_resource` so that's the only place where we need to add a signing account, besides `FetchResourceService` which works with HTML pages as well.

I'm removing the code for attempting to fetch without signature if fetch with signature fails because if #11269 will be enabled in the long-term it will be a waste of time.

Change default `keyId` format from `acct` to `uri`

### DIFF
--- a/app/helpers/jsonld_helper.rb
+++ b/app/helpers/jsonld_helper.rb
@@ -77,19 +77,12 @@ module JsonLdHelper
   end
 
   def fetch_resource_without_id_validation(uri, on_behalf_of = nil, raise_on_temporary_error = false)
+    on_behalf_of ||= Account.representative
+
     build_request(uri, on_behalf_of).perform do |response|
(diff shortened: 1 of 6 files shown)


kaniini:
Picking a random account to sign the requests with is poor metadata hygeine.  It would be desirable to ensure that `Account.representative` is a special account that represents the instance actor.

Gargron:
Instance actor is in the works, but this will do for now. It's only a random account when the "contact account" is not configured. But in either case it's the same account every time. The only risk/downside to this approach vs dedicated instance actor account is that personal accounts may be suspended on the remote end for personal reasons--or, in the case of reports, the remote admin may misinterpret the actions as personal rather than automated.

ClearlyClaire (approved):
Looks ok to me. Even though I agree an instance actor would be way cleaner, for this particular use case it seems like a good interim solution.

PR #11291, opened 2019-07-11
Add whitelist mode

### LINKED ISSUE #11237: Whitelist mode
Previous issues: #3880, #4296

Schools and universities want to teach students how to use social media in a controlled environment. For the safety of the students, data in- and output must be tightly controlled.

In "whitelist mode":

- Public pages cannot be viewed without authentication
- API access is not possible without a user-owned access token (no app-only access)
- ActivityPub resources cannot be accessed without a HTTP signature
- Accounts that do not belong to a whitelisted domain are not processed
- ActivityPub actor JSON remains public, but returns the bare minimum without authentication

The risk of this mode is if it gains widespread adoption in non-academic contexts, thereby making entering the federation with a new server impractical. That would lead to the centralization around a few existing servers in the long term.

My risk mitigation is that the mode should have a "penalty" in the form of a clearly distinguishable branding that makes it clear to any visitor that they are looking at a standalone space and not at the fediverse / open web / Mastodon network.

Gargron:
Fix #11237

- Authorized fetch mode
- Only process accounts from allowed domains
- Disable LD-Signatures
- Disable app-only and unauthenticated REST API access
- Admin UI for whitelisting domains
- Disable activity and peers APIs
- Disable /about and /about/more pages
- Disable profile directory
- Disable unauthenticated access to public account and status pages as well as feeds

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -9,6 +9,7 @@ class Api::BaseController < ApplicationController
   skip_before_action :store_current_location
   skip_before_action :require_functional!
 
+  before_action :require_authenticated_user!, if: :disallow_unauthenticated_api_access?
   before_action :set_cache_headers
 
   protect_from_forgery with: :null_session
@@ -69,6 +70,10 @@ class Api::BaseController < ApplicationController
     nil
   end
 
+  def require_authenticated_user!
+    render json: { error: 'This API requires an authenticated user' }, status: 401 unless current_user
+  end
+
   def require_user!
     if !current_user
       render json: { error: 'This method requires an authenticated user' }, status: 422
@@ -93,5 +98,9 @@ class Api::BaseController < ApplicationController
 
   def set_cache_headers
     response.headers['Cache-Control'] = 'no-cache, no-store, max-age=0, must-revalidate'
+  end
+
+  def disallow_unauthenticated_api_access?
+    authorized_fetch_mode?
   end
 end
--- a/app/controllers/concerns/account_owned_concern.rb
+++ b/app/controllers/concerns/account_owned_concern.rb
@@ -4,6 +4,7 @@ module AccountOwnedConcern
   extend ActiveSupport::Concern
 
   included do
+    before_action :authenticate_user!, if: -> { whitelist_mode? && request.format != :json }
     before_action :set_account, if: :account_required?
     before_action :check_account_approval, if: :account_required?
     before_action :check_account_suspension, if: :account_required?
--- a/app/helpers/domain_control_helper.rb
+++ b/app/helpers/domain_control_helper.rb
@@ -12,6 +12,14 @@ module DomainControlHelper
       end
     end
 
-    DomainBlock.blocked?(domain)
+    if whitelist_mode?
+      !DomainAllow.allowed?(domain)
+    else
+      DomainBlock.blocked?(domain)
+    end
+  end
+
+  def whitelist_mode?
+    Rails.configuration.x.whitelist_mode
   end
 end
--- a/app/services/concerns/payloadable.rb
+++ b/app/services/concerns/payloadable.rb
@@ -14,6 +14,6 @@ module Payloadable
   end
 
   def signing_enabled?
-    ENV['AUTHORIZED_FETCH'] != 'true'
+    ENV['AUTHORIZED_FETCH'] != 'true' && !Rails.configuration.x.whitelist_mode
   end
 end
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -3,6 +3,7 @@
 class AboutController < ApplicationController
   layout 'public'
 
+  before_action :require_open_federation!, only: [:show, :more]
   before_action :set_body_classes, only: :show
   before_action :set_instance_presenter
   before_action :set_expires_in
@@ -18,6 +19,10 @@ class AboutController < ApplicationController
   def terms; end
 
   private
+
+  def require_open_federation!
+    not_found if whitelist_mode?
+  end
 
   def new_user
     User.new.tap do |user|
--- a/app/controllers/activitypub/base_controller.rb
+++ b/app/controllers/activitypub/base_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ActivityPub::BaseController < Api::BaseController
+  skip_before_action :require_authenticated_user!
+
   private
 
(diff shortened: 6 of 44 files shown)


app/services/concerns/payloadable.rb
```diff
@@ -14,6 +14,6 @@ def serialize_payload(record, serializer, options = {})
   end
 
   def signing_enabled?
-    true
+    !Rails.configuration.x.whitelist_mode
```
ThibG: This should also be done for `AUTHORIZED_FETCH`, see https://github.com/tootsuite/mastodon/pull/11295

ClearlyClaire (commented):
Seems mostly good, obviously still missing UI.
Also, switching an existing instance to whitelist mode should be considered. As far as I can see, the current state of this PR will still allow remote followers from non-whitelisted instances to receive local toots, but not send interactions back.

kpcyrd:
Wouldn't this discourage self hosted instances and instead push everybody into the few instances that are too big to fail, like mastodon.social?

ThibG:
@kpcyrd I understand the concern, but I think a whitelist mode remain useful both for some small communities and for stuff like schools or other organizations. In both cases, “big instances that are too big to fail” such as mastodon.social are probably *not* what those whitelist-based instances would be interested in federating with in the first place.

app/controllers/concerns/account_owned_concern.rb
```diff
@@ -4,6 +4,7 @@ module AccountOwnedConcern
   extend ActiveSupport::Concern
 
   included do
+    before_action :authenticate_user!, if: :whitelist_mode?
```
ThibG: Doesn't it break ActivityPub stuff? Plus you call that again in pretty much everything including `AccountOwnedConcern`

app/helpers/domain_control_helper.rb
```diff
@@ -12,6 +12,14 @@ def domain_not_allowed?(uri_or_domain)
       end
     end
 
-    DomainBlock.blocked?(domain)
+    if whitelist_mode?
+      DomainAllow.allowed?(domain)
```
ThibG: Missing negation?

ClearlyClaire (changes requested):
> - Disable activity and peers APIs
> - Disable profile directory
> - Disable unauthenticated access to public account and status pages as well as feeds

Those are all things I am not sure about and that can (or should imho, for the last one) be changed independently.

> - Disable /about and /about/more pages

I am pretty concerned about this one, people may want to have approval-based signup on a whitelisted instance, it makes sense for them to have `/about` and `/about/more`.

Also, the case of switching from a blocklist-based to a whitelist-based approach on an existing instance will not remove existing known accounts from non-whitelisted instances from search results, and it will not stop them from receiving toots they are mentioned in. I think it also lets people fetch statuses from already-known remote accounts that are not whitelisted?

Gargron:
> I am pretty concerned about this one, people may want to have approval-based signup on a whitelisted instance, it makes sense for them to have `/about` and `/about/more`.

I just don't think that "who runs it" and "how many users are there" should be published in this mode, the instance is essentially for private use only. Anything that must still be public can be put on the `/terms` page.

> Also, the case of switching from a blocklist-based to a whitelist-based approach on an existing instance will not remove existing known accounts from non-whitelisted instances from search results, and it will not stop them from receiving toots they are mentioned in. I think it also lets people fetch statuses from already-known remote accounts that are not whitelisted?

Well, how do you imagine it *should* work? Do you want me to purge the accounts at boot time? That could really backfire if an admin doesn't realize it's what this environment variable does. I would put it into a `tootctl` command instead.

app/controllers/concerns/account_owned_concern.rb
Gargron: Good catch

app/controllers/api/base_controller.rb
```diff
@@ -69,6 +70,10 @@ def current_user
     nil
   end
 
+  def require_authenticated_user!
+    render json: { error: 'This API requires an authenticated user' }, status: 401 unless current_user
+  end
+
```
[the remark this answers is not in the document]
Gargron: Extracted it to another method for easier future modification

ThibG:
> I just don't think that "who runs it" and "how many users are there" should be published in this mode, the instance is essentially for private use only. Anything that must still be public can be put on the `/terms` page.

I am not completely convinced. For instance, https://awoo.space/about is a whitelist-based instance and does have an about page (and also user directory).

> I would put it into a `tootctl` command instead.

Yes, that is what I would suggest as well.

app/controllers/api/base_controller.rb
```diff
@@ -9,6 +9,7 @@ class Api::BaseController < ApplicationController
   skip_before_action :store_current_location
   skip_before_action :require_functional!
 
+  before_action :require_authenticated_user!, if: :disallow_unauthenticated_api_access?
```
ThibG: Unless I'm mistaken, this will break any kind of federation as the ActivityPub-specific controllers (including the inbox controller) inherit from `Api::BaseController`

ClearlyClaire (changes requested):

app/controllers/api/base_controller.rb
Gargron: They shouldn't really be inheriting from that, but good catch...

Gargron:
> I am not completely convinced. For instance, https://awoo.space/about is a whitelist-based instance and does have an about page (and also user directory).

Awoo.space can simply keep using whatever they're using. They don't have to use this code. This PR's objective is not to create more of awoo.space or counter.social, but to help private networks like schools and universities. Networks that do *not* attract random people from the internet.

Gargron:
`tootctl domains purge --whitelist-mode` will remove accounts from domains that aren't whitelisted

app/controllers/api/base_controller.rb
```diff
@@ -94,4 +99,8 @@ def authorize_if_got_token!(*scopes)
   def set_cache_headers
     response.headers['Cache-Control'] = 'no-cache, no-store, max-age=0, must-revalidate'
   end
+
+  def disallow_unauthenticated_api_access?
+    whitelist_mode?
```
ThibG: I would probably tie that to the `authorized_fetch` mode more than the whitelist mode

ClearlyClaire (changes requested):
`ActivityPub::InboxesController` also inherits from `Api::BaseController` and should skip `:require_authenticated_user!`

ClearlyClaire (approved):

PR #11294, opened 2019-07-13
Stop reporting the total number of users/statuses in instance

ghost:
The total number of users/statuses are bad criteria for the activity of the instance, especially after gab has been federated. [The Vice says Gab is the biggest instance of Mastodon](https://www.vice.com/en_us/article/mb8y3x/the-nazi-free-alternative-to-twitter-is-now-home-to-the-biggest-far-right-social-network), but that is awfully misleading because he compares just the number of users in the instances.

This PR stops reporting the total number of users/statuses completely. The `/about` and `/about/more` pages show the number of monthly active users instead. For backward interoperability, `user_count` and `status_count` in the `/api/v1/instance` are filled with just the `0`.

You know we can use `/api/v1/instance/activity` instead, which has the weekly active users and the weekly number of statuses.


hakaba-hitoyo:
Sorry, [check-i18n](https://circleci.com/gh/tootsuite/mastodon/74047?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) is failed. I fix it later.

noellabo:
I agree that the total number of users / statuses is often used as a false indicator. Also, it may be easy to over-report or under-report, and it may be used for inappropriate campaigns.

In most cases, what really helps is the number of active users and posts.

However, the total information is still useful if it is used as a correct indicator. Comparison of total and active is a particularly interesting indicator.

Displaying it proudly in the total number has the effect of impressing that it is better for people to concentrate in one place. For Mastodon's design, it may be desirable to promote moderate dispersion.

The method of deleting the total number makes the comparison on the total number inaccurate, especially if a server with a large total number is adopted. But I think it is not a good way.

We need ideas to guide people to use activity as an indicator.

hakaba-hitoyo:
All checks have passed. Ready to merge.

hakaba-hitoyo:
@noellabo Monthly active users seems good enough for deleting total number of users. It may also a good idea to add monthly number of posts.

marek-lach:
> The total number of users/statuses are bad criteria for the activity of the instance, especially after gab has been federated. [The Vice says Gab is the biggest instance of Mastodon](https://www.vice.com/en_us/article/mb8y3x/the-nazi-free-alternative-to-twitter-is-now-home-to-the-biggest-far-right-social-network), but that is awfully misleading because he compares just the number of users in the instances.
> 
> This PR stops reporting the total number of users/statuses completely. The `/about` and `/about/more` pages show the number of monthly active users instead. For backward interoperability, `user_count` and `status_count` in the `/api/v1/instance` are filled with just the `0`.
> 
> You know we can use `/api/v1/instance/activity` instead, which has the weekly active users and the weekly number of statuses.

Vice's reporting is deliberately misleading, and we should not cripple Mastodon's functionality because of their orchestrated smears.

hakaba-hitoyo:
@marek-lach Gab is not the first instance which has too large nominal number of users than its real activity. For example, `horiedon.com` had over 3,500 nominal users, who registered just for their curiosity, and rarely posted in the instance. Nominal number of users is a misleading criteria, not only for Ben Makuch of VICE, but also for any potential newcomers.

tateisu:
What is active users?
Weekly login number in /api/v1/instance/activity is "count of the users who use web UI or REST API with access token". 
But many mobile app has periodically check for notification via REST API even if the user does not open the app while a period. These are counted in weekly login number.

hakaba-hitoyo:
@tateisu Thank you for informing the precise specification of the `/api/v1/instance/activity`. It sounds good enough approximation for the actual activity of the instance.

PR #11295, opened 2019-07-14
Disable LDSigning when AUTHORIZED_FETCH is set to true

ClearlyClaire:
Otherwise, replies and toots through relays will still end up on blocked instances

### DIFF
--- a/app/services/concerns/payloadable.rb
+++ b/app/services/concerns/payloadable.rb
@@ -14,6 +14,6 @@ module Payloadable
   end
 
   def signing_enabled?
-    true
+    ENV['AUTHORIZED_FETCH'] != 'true'
   end
 end


Gargron:
Does the relay code check for signature before sending the payload? If not, it should. Likewise the reply forwarding bits.

Reply forwarding has become an important part of Mastodon, in that it really substantially improves the readability of conversations. While it's okay to disable it in authorized fetch mode while it's optional, if the long-term goal is making authorized fetch mode the default, we need to figure out how reply forwards could work without LD-Signatures.

ThibG:
It does check, and do not try to forward unsigned toots. That wouldn't be new behavior, that would be akin to receiving messages from pleroma (which does not LDSign anything).

Yes, reply forwarding is somewhat important, but LDSigning is not compatible with AUTHORIZED_FETCH

Gargron:
@ThibG No I mean on our (sending to relay) end.

ThibG:
Oh, I don't believe it does check them. Most relays handle that fine by wrapping it in an Announce, though (even if that's far less efficient than using LDSigning)

Gargron (approved):

PR #11296, opened 2019-07-14
Add setting to disable the anti-spam

ClearlyClaire:


ykzts (approved):

Gargron:
I would prefer if the feature was perfected in such a way that hard-disabling it would not be necessary. Although as I understand it this action is pre-emptive as nobody has complained about too many false positives yet. But either way: This method makes disabling it inaccessible to masto.host (and similar) users.

ThibG:
Good point, going to change that to an admin panel setting.

ThibG:
(It has been changed)

config/locales/en.yml
```diff
@@ -390,6 +391,9 @@ en:
       activity_api_enabled:
         desc_html: Counts of locally posted statuses, active users, and new registrations in weekly buckets
         title: Publish aggregate statistics about user activity
+      antispam_enabled:
+        desc_html: Mastodon can auto-silence and auto-report accoutns based on measures such as detecting accounts who send repeated unsolicited messages. There may be false positives.
```
Gargron: Typo in accounts

app/controllers/admin/dashboard_controller.rb
```diff
@@ -30,6 +30,7 @@ def index
       @trending_hashtags     = TrendingTags.get(7)
       @profile_directory     = Setting.profile_directory
       @timeline_preview      = Setting.timeline_preview
+      @antispam_enabled      = Setting.antispam_enabled
```
Gargron: If the Ruby class is called "spam_check" would it not be more consistent to name it "spam_check_enabled"? That being said it probably doesn't matter.

PR #11298, opened 2019-07-14
Add public blocks to /about/blocks

ClearlyClaire:
Add an option to display domain blocks on /about/blocks.

image

This setting has three options:
- Show to no one (disabled, the default)
- Show to logged-in users only
- To everyone

It comes with a similar setting to show the rationale.

## Without rationale

image

## With rationale

image

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -23,15 +23,7 @@ class AboutController < ApplicationController
   def blocks
     @show_rationale = Setting.show_domain_blocks_rationale == 'all'
     @show_rationale |= Setting.show_domain_blocks_rationale == 'users' && !current_user.nil? && current_user.functional?
-    @blocks = DomainBlock.with_user_facing_limitations
-
-    @blocks = begin
-      if @show_rationale
-        @blocks.order('(CASE severity WHEN 0 THEN 1 WHEN 1 THEN 2 WHEN 2 THEN 0 END), reject_media, public_comment, domain')
-      else
-        @blocks.order('(CASE severity WHEN 0 THEN 1 WHEN 1 THEN 2 WHEN 2 THEN 0 END), reject_media, domain')
-      end
-    end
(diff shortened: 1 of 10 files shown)


ThibG:
Note that it lacks the rationale (as there is currently none included in the database), and it doesn't support censoring domain names.

koyuawsmbrtn:
I need such a feature! It's a real pain in the ass to manually maintain a public blocklist.

Kjwon15:
I am more interested on that "Sandbox" severity. (But it's more like soft-block)

Gargron:
> I am more interested on that "Sandbox" severity. (But it's more like soft-block)

That's just a more commonly understandable phrasing. Sandbox means "silenced". The text and format of the page is taken from mastodon.social's published blocklist on GitHub. Since that was a standalone thing I felt comfortable using different terminology with explanations, but within Mastodon, it might be worthwhile to either a) use "silence" consistently or b) consistently rename all mod actions from "silence" to "sandbox".

Kjwon15:
@Gargron is "silence" also prevent notification from stranger too? if that current hint message from admin UI below should be changed.
> <strong>Silence</strong> will make the account's posts invisible to anyone who isn't following them.

ThibG:
Changed “Sandbox” to “Silence” and rebased on latest master.

BenLubar:
Grouping domains by string-equal rationale would be good when adding rationale is supported.

clarfon:
I feel like rationales should be discussed in another PR. What makes sense is to just have a set list of tags admins can control and allow filtering by tag.

ThibG:
Edited to add support for optionally showing the public comment (rationale).

As for showing the blocklist at all, it can be set to show the rationale to anyone, logged in users, or nobody.

ykzts (approved):

ThibG:
Rebased, and redesigned it so it better handles long domain names and rationales:
image

ThibG:
Reworded the text a bit and added a disclaimer about suspended servers still being able to retrieve public content, whenever the instance is in `public_fetch_mode?`.

image

ThibG:
@Gargron I think the text is fine, can this be merged?

PR #11300, opened 2019-07-14
Fix leaking private statuses the admin account follows

ClearlyClaire:
Now that the request is signed, it can return private toots. Do not leak them.


Gargron:
At some point you wanted to sign such requests with the searching user, now would be the time to do it.

ThibG:
Yes, but given the privacy implications, this should remain optional, so that added check is useful either way

Gargron (approved):

PR #11301, opened 2019-07-15
Fix trying to deliver to relays/forward replies when signing is disabled

Gargron:


nightpool:
why make this change here instead of fixing the ingress logic to do resolve the status when it gets a forwarded status it can't authenticate?

Gargron:
I'm not sure I understand? This is a fix for redundant egress

nightpool:
It just seems weird to me to cut the egress back now when we know we want to fix reply forwading and relays in the future, even under secure mode. Why not just look up the status like we do for unknown Announces?

Gargron:
Well, these forwards aren't announces so I feel like it's a bit harder to know that something even needs to be done in that regard. Like without an LD-Signature they just look like misattributed posts (and I believe they should get rejected because of that)

ThibG:
I'm not too sure about the reply forwarding part, but the relay part is definitely working with relays made for Pleroma. Pleroma never had any LDSigning, so their relays work by `Announce`ing activities they receive. They are widely in use and work with both Mastodon and Pleroma.

On the other hand, if it is to check signatures at every request and not cache status, I'm unsure Mastodon would ever manage to cope with near-simultaneous requests from all subscribers to a relay.

PR #11303, opened 2019-07-15
Remove WebSub subscriptions

Gargron:


ThibG:
Is there way we can cleanly remove them on the other end during the migration script?

Gargron:
That's not needed is it? The model represents the sending side, and we are free to just stop sending whenever.

ykzts (approved):

PR #11320, opened 2019-07-15
Add `tootctl preview_cards remove`

mayaeh:
This changes will be change the remove thumbnails of previews from automatic execution by the Sidekiq scheduler to manual execution by the `tootctl` command.

After #11304, object storage’s rate-limit is triggered when `Scheduler::PreviewCardsCleanupScheduler ` is executed in some object storage environments.

スクリーンショット 2019-07-15 22 43 31

スクリーンショット_2019-07-15_22_46_12

This is similar to #8187.

So I made `tootctl preview_cards remove` command referring to `tootctl media remove`.

There are multiple `if` statements.
I think that it is not a good writing style.

Please let me know if there is a better way of writing.

----
TODO
- [x] Remove unused file
- [x] Fix if/elsif blocks
- [x] Add output to confirm when the number of days is less than 2 weeks


lib/mastodon/preview_cards_cli.rb
```diff
@@ -0,0 +1,91 @@
...
+      With the --link option, delete only link-type preview cards.
+    DESC
+    def remove
+      time_ago  = options[:days].days.ago
+      queued    = 0
+      processed = 0
+      size      = 0
+      dry_run   = options[:dry_run] ? '(DRY RUN)' : ''
+
+      if options[:link] && options[:background]
```
Gargron: I think the number of if/elsif blocks could be reduced to just two if the scope for the query was in a variable.

config/sidekiq.yml
```diff
@@ -27,9 +27,6 @@
   ip_cleanup_scheduler:
     cron: '<%= Random.rand(0..59) %> <%= Random.rand(3..5) %> * * *'
     class: Scheduler::IpCleanupScheduler
-  preview_cards_cleanup_scheduler:
-    cron: '<%= Random.rand(0..59) %> <%= Random.rand(3..5) %> * * *'
-    class: Scheduler::PreviewCardsCleanupScheduler
```
Gargron: You should delete the `Scheduler::PreviewCardsCleanupScheduler` file too.

Gargron (changes requested):

Gargron:
> After #11304, object storage’s rate-limit is triggered when `Scheduler::PreviewCardsCleanupScheduler ` is executed in some object storage environments.

It might be possible to add a rate-limit system to Sidekiq jobs. While `Scheduler::PreviewCardsCleanupScheduler` is an obvious culprit here, at certain number of Sidekiq threads/workers and certain workloads (a lot of incoming posts with media?) the object-storage rate-limit could be triggered by other jobs as well. Something to think about.

lib/mastodon/preview_cards_cli.rb
mayaeh: Thank you very much.
I think I was able to put it together using the scope.

config/sidekiq.yml
mayaeh: I removed that file.

mayaeh:
> While Scheduler::PreviewCardsCleanupScheduler is an obvious culprit here, at certain number of Sidekiq threads/workers and certain workloads (a lot of incoming posts with media?) the object-storage rate-limit could be triggered by other jobs as well. 

I think that it isn't happening on my server so far.
But, as you say, I think it can happen.

lib/mastodon/preview_cards_cli.rb
```diff
@@ -0,0 +1,94 @@
...
+      dry_run   = options[:dry_run] ? '(DRY RUN)' : ''
+      link      = options[:link] ? 'link-type ' : ''
+      scope     = PreviewCard.where.not(image_file_name: nil)
+      scope     = scope.where.not(image_file_name: '')
+      scope     = scope.where(type: :link) if options[:link]
+      scope     = scope.where('updated_at < ?', time_ago)
+
+      if time_ago > 2.weeks.ago
+        prompt.say "\n"
+        prompt.say('The preview cards less than the past two weeks will not be re-acquired even when needed.')
```
mayaeh: Is this text easy to understand? I do not have any confidence.

mayaeh: Is this text easy to understand? I do not have any confidence.

Gargron: Media remove does not ask for confirmation. Do you think it's necessary here?

Gargron (approved):

lib/mastodon/preview_cards_cli.rb
mayaeh: Yes, I think it is necessary.

In preview cards, I think that only those older than 2 weeks will be re-acquired.
https://github.com/tootsuite/mastodon/blob/master/app/services/fetch_link_card_service.rb#L25

On the other hand, I can not find the date limit for re-acquisition of attached media.

Therefore, I think that it is necessary to output a warning if the number of days less than 2 weeks is specified in this command.

lib/mastodon/preview_cards_cli.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    option :days, type: :numeric, default: 180
+    option :background, type: :boolean, default: false
+    option :verbose, type: :boolean, default: false
+    option :dry_run, type: :boolean, default: false
+    option :link, type: :boolean, default: false
+    desc 'remove', 'Remove preview cards'
+    long_desc <<-DESC
+      Removes locally thumbnails for previews.
```
angristan: What does `locally thumbnails` means?

PR #11321, opened 2019-07-15
Add instance actor (second take)

ClearlyClaire:
#10457 but using an actual `Account` with the following tricks:
- use `instance.tld` as username, to make sure we're not hitting an existing username
- use `Application` as `actor_type` and `0` as `id` to uniquely
- only serialize a subset of the `Actor` fields
- use its own route and controller to make it unique and avoid issues with dots in paths

It seems to work, I am not sure what may be missing but I haven't actually tested federating with it yet.


app/controllers/instance_actors_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+  def set_account
+    @account = Account.find(-99)
+  end
+
+  def restrict_fields_to
+    %i(id type preferred_username inbox public_key endpoints icon image url manually_approves_followers)
```
Gargron: What's the point of icon/image/url/manually_approves_followers?

app/serializers/activitypub/actor_serializer.rb
```diff
@@ -39,10 +39,12 @@ def shared_inbox
   delegate :moved?, to: :object
 
   def id
-    account_url(object)
+    object.instance_actor? ? instance_actor_url : account_url(object)
   end
 
   def type
+    return 'Application' if object.instance_actor?
```
Gargron: Feel this would be more readable in an if/elsif/else block.

app/views/instance_actors/show.html.haml
```diff
@@ -0,0 +1,48 @@
+- content_for :page_title do
```
Gargron: I can't really tell how this is going to look from code alone but do we *need* a HTML page for this?

config/routes.rb
```diff
@@ -28,6 +28,10 @@
   get 'intent', to: 'intents#show'
   get 'custom.css', to: 'custom_css#show', as: :custom_css
 
+  resource :instance_actor, path: 'instance-actor', only: [:show] do
```
Gargron: Maybe just `actor`? There's no overlap either way. Seems redundant.

db/migrate/20190715164535_add_instance_actor.rb
```diff
@@ -0,0 +1,9 @@
+class AddInstanceActor < ActiveRecord::Migration[5.2]
+  def up
+    Account.create!(id: -99, actor_type: 'Application', locked: true, username: Rails.configuration.x.local_domain)
```
Gargron: You need to add this to `seed.rb` as well for new installations.

Gargron (changes requested):

Gargron:
You should probably make `Account.representative` return this record

app/controllers/instance_actors_controller.rb
ThibG: `icon` and `image` were set to banner and hero or something like that in the other PR, but I have not done that, so it's not super useful, I'll remove them.

`manually_approves_followers` is kinda here to not confuse people when follows do not work

`url` is used to point at an explanation of what the actor is.

app/serializers/activitypub/actor_serializer.rb
ThibG: Alternatively, I considered using `object.actor_type` directly. I think this *should* be fine, but I'm not completely sure all the contexts it is used in. Will do an if/elsif else block for now.

app/views/instance_actors/show.html.haml
ThibG: The page is useful to let people know what this actor is about, as it may be confusing. Having it on a separate page allows people to know at a glance that it is a special user.

It looks like this:
image

config/routes.rb
ThibG: For the controller name? The path? I thought having “instance” in the name would make it clearer that it is meant to represent the instance itself.

db/migrate/20190715164535_add_instance_actor.rb
ThibG: Oops. I will do that!

ThibG:
> You should probably make `Account.representative` return this record

Yes, I will do that, I just need to think about how to do cleanly, and I'd have to adjust the tests.

app/views/instance_actors/show.html.haml
Gargron: Should the `url` maybe just point to the `/about` page then? What I'm trying to avoid is the maintenance cost of another static page design because they've tended to be redesigned quite a few times so far.

config/routes.rb
Gargron: For the path! Seems cleaner that way for two reasons: Our ActivityPub JSON doesn't use the word "instance" anywhere, and other multi-word URLs use underscores instead of dashes, so there would be a consistency issue.

ThibG: Ok, changed.

app/views/instance_actors/show.html.haml
ThibG: I don't think so, we really want to explain what this account is for. I should probably reword that description btw.

Gargron (approved):

ThibG:
Changed the instance actor profile URL to `/about/more?instance_actor=true` which displays a flash message:
image

This has the added benefit of always returning json for the actor URI, thus not having to worry about the `Vary` header for caching.

Gargron (approved):

Pizzacus:
> It is used for federation purposes and should not be blocked unless you want to block the whole **instance**

So far, the English version of Mastodon has been VERY consistent about using the word "server" instead of "instance".

This would actually be, as far as I know, the first time the word "instance" is used on the actual UI.

This should be addressed should it not?

PR #11326, opened 2019-07-16
Make puma bind address configurable with BIND env var

zunda:
This makes it possible to keep `puma` running on Heroku with setting config var `BIND=0.0.0.0` like #11302 for streaming.


Gargron:
Can't we modify Procfile to add `-b 0.0.0.0`?

zunda:
I guess we could do that (not tested) but I feel it is better to provide similar interface for both server processes.

Gargron (approved):

PR #11332, opened 2019-07-16
Extend AUTHORIZED_FETCH mode to user blocks as well

ClearlyClaire:
We might hold off merging that until we see what cwebber has in mind, but I think it makes sense to make sure instances blocked by an user don't get to see their toots.

### DIFF
--- a/app/controllers/statuses_controller.rb
+++ b/app/controllers/statuses_controller.rb
@@ -64,7 +64,6 @@ class StatusesController < ApplicationController
   def set_status
     @status = @account.statuses.find(params[:id])
     authorize @status, :show?
-    raise Mastodon::NotPermittedError if current_account&.domain&.present? && @account.domain_blocking?(current_account.domain)
   rescue Mastodon::NotPermittedError
     raise ActiveRecord::RecordNotFound
   end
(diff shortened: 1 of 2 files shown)


app/controllers/statuses_controller.rb
```diff
@@ -64,6 +64,7 @@ def set_link_headers
   def set_status
     @status = @account.statuses.find(params[:id])
     authorize @status, :show?
+    raise Mastodon::NotPermittedError if current_account&.domain&.present? && @account.domain_blocking?(current_account.domain)
```
Gargron: This should just be part of `StatusPolicy` imo

Gargron (changes requested):

app/controllers/statuses_controller.rb
ThibG: Could be, but it's really the only point where it can be called with a `current_account` that is of another domain

Gargron: It would be more clean and more future-proof. Also you're raising an exception just for it to be caught and re-raised as a different one straight away.

Gargron (approved):

PR #11333, opened 2019-07-17
Add (back) rails-level JSON caching

Gargron:
This is my idea for #11330 with prettier (and less) code, replacing `render_cached_json` with `render_with_cache`.

The `:key`, unless supplied explicitly, is generated from the controller name, controller action, the cache key of the object being serialized, as well as fields chosen. For example, it could look like this:

    accounts/show:accounts/1:id,type,preferred_username,public_key,endpoints

When no fields are passed (such as when all fields should be rendered), it's a lot shorter:

    accounts/show:accounts/1

When there's no record, the key is even simpler:

    api/v1/instances/show

The method supports being passed `:json` a record or hash, a method name as a symbol, or a block. The latter two will only get evaluated if there is no cache. Example:

```ruby
def show
  render_with_cache json: :foo
end

private

def foo
  # Heavy computing here
end
```

This means invocations of `render_with_cache` look almost identical to just using `render` and you don't need to think about manually specifying cache keys.

### DIFF
--- a/app/controllers/concerns/cache_concern.rb
+++ b/app/controllers/concerns/cache_concern.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module CacheConcern
+  extend ActiveSupport::Concern
+
+  def render_with_cache(**options)
+    raise ArgumentError, 'only JSON render calls are supported' unless options.key?(:json) || block_given?
+
+    key        = options.delete(:key) || [[params[:controller], params[:action]].join('/'), options[:json].respond_to?(:cache_key) ? options[:json].cache_key : nil, options[:fields].nil? ? nil : options[:fields].join(',')].compact.join(':')
+    expires_in = options.delete(:expires_in) || 3.minutes
+    body       = Rails.cache.read(key, raw: true)
+
+    if body
+      render(options.except(:json, :serializer, :each_serializer, :adapter, :fields).merge(json: body))
+    else
+      if block_given?
+        options[:json] = yield
+      elsif options[:json].is_a?(Symbol)
+        options[:json] = send(options[:json])
+      end
+
+      render(options)
+      Rails.cache.write(key, response.body, expires_in: expires_in, raw: true)
+    end
+  end
+
+  def set_cache_headers
+    response.headers['Vary'] = public_fetch_mode? ? 'Accept' : 'Accept, Signature'
+  end
+
+  def cache_collection(raw, klass)
+    return raw unless klass.respond_to?(:with_includes)
+
+    raw                    = raw.cache_ids.to_a if raw.is_a?(ActiveRecord::Relation)
+    cached_keys_with_value = Rails.cache.read_multi(*raw).transform_keys(&:id)
+    uncached_ids           = raw.map(&:id) - cached_keys_with_value.keys
+
+    klass.reload_stale_associations!(cached_keys_with_value.values) if klass.respond_to?(:reload_stale_associations!)
+
+    unless uncached_ids.empty?
+      uncached = klass.where(id: uncached_ids).with_includes.each_with_object({}) { |item, h| h[item.id] = item }
+
+      uncached.each_value do |item|
+        Rails.cache.write(item, item)
+      end
+    end
+
+    raw.map { |item| cached_keys_with_value[item.id] || uncached[item.id] }.compact
+  end
+end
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/cache_concern.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module CacheConcern
+  extend ActiveSupport::Concern
+
+  def render_with_cache(**options)
+    raise ArgumentError, 'only JSON render calls are supported' unless options[:json].present? || block_given?
+
+    key        = options.delete(:key) || request.url
```
Gargron: This is the part I don't really like--using `request.url` allows cache-busting too easily, and we don't differentiate between bare/full actor with this. My criteria however is to keep `render_with_cache` calls as minimal as possible instead of manually supplying a long key definition...

app/controllers/concerns/cache_concern.rb
```diff
@@ -0,0 +1,45 @@
...
+    expires_in = options.delete(:expires_in) || 3.minutes
+    body       = Rails.cache.read(key)
+
+    if body
+      render(options.except(:json, :serializer, :each_serializer, :adapter, :fields).merge(json: body))
+    else
+      options[:json] = yield if block_given?
```
Gargron: I was considering allowing calls like `render_with_cache json: -> { heavy_stuff_here }` initially but allocating Procs is slightly heavier than using yield, plus it's not that pretty either. However, supporting `render_with_cache json: :method_name_here` might be worth it..?

app/controllers/concerns/cache_concern.rb
ThibG: That is definitely an issue, at least wrt. the bare/full actor thing.

app/controllers/concerns/cache_concern.rb
ThibG: I don't know Ruby enough to know exactly how this would work, but yeah, that might be worth it.

app/controllers/concerns/cache_concern.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module CacheConcern
+  extend ActiveSupport::Concern
+
+  def render_with_cache(**options)
+    raise ArgumentError, 'only JSON render calls are supported' unless options[:json].present? || block_given?
```
ThibG: `options[:json].present?` returns false for an existing but empty Hash, thus erroring out for the instance info controller as all the logic is in the serializer

ClearlyClaire (changes requested):

app/controllers/concerns/cache_concern.rb
Gargron: Could use controller+action+cache_key of the object, if the object responds to cache_key

ThibG: That still wouldn't help, would it? The bare/full actor thing is a difference in the `fields` option, not the actor itself. You could pass the key explicitly for that case.

app/controllers/concerns/cache_concern.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module CacheConcern
+  extend ActiveSupport::Concern
+
+  def render_with_cache(**options)
+    raise ArgumentError, 'only JSON render calls are supported' unless options.key?(:json) || block_given?
+
+    key        = options.delete(:key) || [[params[:controller], params[:action]].join('/'), options[:json].respond_to?(:cache_key) ? options[:json].cache_key : nil, options[:fields].nil? ? nil : options[:fields].join(',')].compact.join(':')
```
ThibG: The cache key being dependent on the controller's action would mean that `StatusesController#show` and `StatusesController#activity` would have different cache keys, which isn't optimal. Not sure that would really be a problem in practice, though. What is `StatusesController#activity` used for anyway?

Gargron: That would be the correct behavior because those actions return different json. Activity is used for the resolveable id of activities.

ThibG: Do they really return different json?
Hm, also… that's not directly related, but `#activity` doesn't seem to require signatures even if `authorized_fetch_mode?` is true

Gargron: The show returns a Note object, the activity returns a Create or Announce object.

ThibG: Oh, right! Sorry about that.

ClearlyClaire (approved):
I'm not super comfortable with how it moves the cache key logic in one place which tries to figure out a lot of cases by itself, but that's just my opinion. Otherwise it looks fine.

PR #11343, opened 2019-07-17
Fix `alerts` booleans not being typecast correctly in push subscription

### LINKED ISSUE #10789: [API] Push subscription type inconsistency
`GET /api/v1/push/subscription` returns a `Push subscription` object containing an `alerts` dict which specifies which alert types are currently active. It’s a boolean value but they are presented as strings instead of true booleans. In other places the API correctly uses the boolean type to represent such values. I would expect this to be the case here also.

Edit: Also the `alert` object is currently undocumented.

### Expected behaviour
```json
{"id":1,"endpoint":"https://example.com","alerts":{"follow":true,"favourite":true,"reblog":true,"mention":true},"server_key":"foobar"}
```

### Actual behaviour
```json
{"id":1,"endpoint":"https://example.com","alerts":{"follow":"true","favourite":"true","reblog":"true","mention":"true"},"server_key":"foobar"}
```

GET /api/v1/push/subscription

### Specifications

Tested using Mastodon 2.8.2

Gargron:
Fix #10789

### DIFF
--- a/app/serializers/rest/web_push_subscription_serializer.rb
+++ b/app/serializers/rest/web_push_subscription_serializer.rb
@@ -4,7 +4,7 @@ class REST::WebPushSubscriptionSerializer < ActiveModel::Serializer
   attributes :id, :endpoint, :alerts, :server_key
 
   def alerts
-    object.data&.dig('alerts') || {}
+    (object.data&.dig('alerts') || {}).each_with_object({}) { |(k, v), h| h[k] => ActiveModel::Type::Boolean.new.cast(v) }
   end
 
   def server_key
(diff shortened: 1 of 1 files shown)


ClearlyClaire (changes requested):
Seems good apart from what CodeClimate is complaining about

app/serializers/rest/web_push_subscription_serializer.rb
```diff
@@ -4,7 +4,7 @@ class REST::WebPushSubscriptionSerializer < ActiveModel::Serializer
   attributes :id, :endpoint, :alerts, :server_key
 
   def alerts
-    object.data&.dig('alerts') || {}
+    (object.data&.dig('alerts') || {}).each_with_object({}) { |(k, v), hash| hash.merge(k => ActiveModel::Type::Boolean.new.cast(v)) }
```
ThibG: Why not just `hash[k] = ActiveModel::Type::Boolean.new.cast(v)`?
The point of `each_with_object` is to mutate the object.

ClearlyClaire (approved):

PR #11345, opened 2019-07-17
Fix only one middle dot being recognized in hashtags

### LINKED ISSUE #10934: Strange behaviour with U+00B7 MIDDLE DOT in hashtags
### Expected behaviour

U+00B7 MIDDLE DOT should either be allowed any number of times in hashtags, or else disallowed completely.

### Actual behaviour

From my [informal tests](https://weirder.earth/@onethousandtwentyfour/102205959763922491), the *first* U+00B7 MIDDLE DOT to be used in a hashtag is included, but later ones are not. I canʼt imagine why this would be intentional; it is extremely confusing for a character to only *sometimes* terminate hashtags. There may be a larger bug with the hashtag detection mechanism not limited to just this character.

(For clarity, if a user enters the text `#one·two·three`, the generated hashtag is `#one·two`, stopping before the second U+00B7 MIDDLE DOT, but including the first.)

### Specifications

The above link is on an older version of Mastodon but I tested on fandom.ink (running 2.8.4) and the behaviour still seems to be present.

Gargron:
Fix #10934

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -17,7 +17,7 @@ class Tag < ApplicationRecord
   has_many :featured_tags, dependent: :destroy, inverse_of: :tag
   has_one :account_tag_stat, dependent: :destroy
 
-  HASHTAG_NAME_RE = '[[:word:]_]*[[:alpha:]_·][[:word:]_]*'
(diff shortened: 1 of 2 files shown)


ykzts (approved):

ThibG:
This allows purely numeric hashtags such as `#1234`. Is that intended? My understanding was that it wasn't.

PR #11348, opened 2019-07-18
Play animated custom emoji on hover

ClearlyClaire:


app/lib/formatter.rb
```diff
@@ -157,7 +153,12 @@ def encode_custom_emojis(html, emojis, animate = false)
         emoji     = emoji_map[shortcode]
 
         if emoji
-          replacement = "<img draggable=\"false\" class=\"emojione\" alt=\":#{encode(shortcode)}:\" title=\":#{encode(shortcode)}:\" src=\"#{encode(emoji)}\" />"
+          original_url, static_url = emoji
+          replacement = if animate
```
Gargron: I would use begin/end to make this assignment more readable, as I've been doing recently.

Gargron (changes requested):
There is a merge conflict.

app/lib/formatter.rb
ThibG: Done

Gargron (approved):

PR #11358, opened 2019-07-19
New audio player

koyuawsmbrtn:
### Description

New audio player to differentiate between audio and actual video.

### Preview
image


Gargron:
I closed #9480 because I wasn't confident about this look. Needless to say it doesn't look quite right to me. It might not be possible to introduce a good-looking audio player at all until/unless cover art and meta tags are added. The video player is functionally the same so is there any pressure to change the UI?

koyuawsmbrtn:
Not really, but most of my files (and I think some others ones too) don't have any meta data in them or are just small sound effects. It's kind of more about user experience, because I always expect a video, but instead I get fart sounds or something. Also the change can be reverted if someone somes up with a better idea, but for now this UI is making it look familiar. I still get your point though.

koyuawsmbrtn:
So some people voted for it and I also think this UI is actually fine until real work is being done to displaying file meta data. I always say release first, fix later. Bad approach, but users want features even if I don't declare this PR as unstable. Maybe someone like @ykzts could review it.

ThibG:
I think this does look better than the current `Video` player. It would be better still if it displayed the current time and total or remaining time, and if it used the `audio` HTML tag instead of the `video` one.

Previous/next buttons would also be good in case of multiple uploaded files.

koyuawsmbrtn:
@ThibG audio multi-uploads aren't a thing yet, but sure i could go and implement a timer

ThibG:
I have been toying with https://github.com/ThibG/mastodon/tree/features/audio-player
I can't get something that looks much better, it's basically the same in—I think—a little bit cleaner.
I do think that either that or this PR in the current state is better than what we have in Mastodon for now, it's very confusing having a “blank” “video”.

koyuawsmbrtn:
@ThibG thanks, I'll have a look at it once I'm back home

koyuawsmbrtn:
@ThibG Looks better than my PR to be honest. I will close this one and create a new one with your code as base repo.

koyuawsmbrtn:
Further discussion on #11635

PR #11359, opened 2019-07-19
Change single-column mode to scroll the whole page

### LINKED ISSUE #10840: Simple web interface: scroll from anywhere
The new simple one-column interface is very nice, but after using it a little bit there is one behaviour that I found eventually annoying: scrolling is constrained to the column only. Which means that if my cursor is anywhere on the page but the column, wheeling up and down is not going to do anything.

I can see how this has evolved from the fact that it is adapted from the multi-column interface, but scrolling should probably be rethought for this one, as that behaviour is very unexpected. In other one-column feeds like Twitter, Facebook or Instagram, scrolling happens on the whole page, even if the feed itself is laid out only on a central column.

So either it should be the whole page scrolling, with the left and right parts of the interface as well as the column header being sticky, or scrolling from outside the column should be captured somehow and applied to the column.

**Edit:** By the way, this is not only about wheel scrolling, but also autoscrolling (pressing middle-click then going up and down) and keyboard scrolling (pressing up arrow, down arrow, page up, page down, page start, page end keys). They all don’t work if the cursor is not precisely on the column.

### LINKED ISSUE #11044: Add a class that represents the current UI like a theme
### Pitch

Add `ui-advanced` or `ui-simple` class to the body element.

Single column UI has been implemented in 2.9.0 (rc). This UI is very different in elements from the past Mastodon. But is there a way to determine this in CSS?

For example, if it is a theme, I can determine it by looking at the class attached to the body element.

`theme-default` is a dark theme and `theme-mastodon-light` is a light theme. This allows one CSS to apply a color that suits the user's one.

### Motivation

Custom CSS can provide styles according to the UI.

Gargron:
Fix #10840 

Side-effects:

- Removed useless overlapping background paint
- Calculate scrollbar width dynamically to prevent jump when media modal opened
- Fix #11044: Add `layout-single-column` and `layout-multiple-columns` to body

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -35,6 +35,7 @@ export default class ScrollableList extends PureComponent {
     alwaysPrepend: PropTypes.bool,
     emptyMessage: PropTypes.node,
     children: PropTypes.node,
+    bindToDocument: PropTypes.bool,
   };
 
   static defaultProps = {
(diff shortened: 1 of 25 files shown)


ykzts (approved):

wcpaez:
Having issues on Ubuntu 18.04 with Google Chrome Version 72.0.3626.109 (Official Build), the scroll is not displayed. On firefox is working great.

PR #11372, opened 2019-07-21
Modify external/localhost email server question

### LINKED ISSUE #11369: localhost question email confusion
Edit: it appears I've misunderstood the question. 

### Expected behaviour

When running the Docker setup, I'm prompted for whether I want to setup SMTP mailing or not. When saying no, I should not be prompted for configuration information relating to SMTP.

### Actual behaviour
When running the Docker setup, I'm prompted for whether I want to setup SMTP mailing or not. When saying no, I am prompted for configuration information relating to SMTP:
image

I was running the docker setup script as per the instructions using `docker-compose run --rm web bundle exec rake mastodon:setup`, using a the pre-built container images available from Docker Hub.

### Specifications

Using container image 2.9.2 for Mastodon, Ubuntu linux as the host. Docker version 18.09.8, build 0dd43dd87f. docker-compose version 1.24.1, build 4667896b

BNolet:
Question about localhost is misleading, modified question to make it clearer.

Would fix #11369

### DIFF
--- a/lib/tasks/mastodon.rake
+++ b/lib/tasks/mastodon.rake
@@ -231,6 +231,8 @@ namespace :mastodon do
       prompt.say "\n"
 
       loop do
+        break unless prompt.yes?('Do you want your instance to send e-mails?', default:true)
+        
         if prompt.yes?('Do you want to send e-mails from localhost?', default: false)
           env['SMTP_SERVER'] = 'localhost'
           env['SMTP_PORT'] = 25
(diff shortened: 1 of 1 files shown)


nightpool:
I don't think we want to support installing mastodon without the possibility of sending emails. Maybe we can reword the original question to "Do you want to send emails locally or use an external server?"

BNolet:
(although i'm not sure exactly how we would parse responses to that)

If this is the case then I would suggest the question "Do you want to use an external email server?" 

And if no, then do the same thing it already does when saying yes. I'll modify my PR to reflect this

PR #11376, opened 2019-07-21
fix a typo

partev:


Gargron:
The change won't be picked up unless it is submitted via: https://crowdin.com/project/mastodon

partev:
per comment moved this change to https://crowdin.com/project/mastodon

PR #11377, opened 2019-07-22
List columns within the method for a migration

zunda:
I tried creating a new server after a while and saw the exception below while running `rake db:migrate` for the first time. This change is to avoid this exception.

```
NoMethodError: undefined method `perform' for nil:NilClass
.../vendor/bundle/ruby/2.6.0/gems/strong_migrations-0.4.1/lib/strong_migrations/migration.rb:14:in `method_missing'
.../vendor/bundle/ruby/2.6.0/gems/activerecord-5.2.3/lib/active_record/migration.rb:604:in `method_missing'
.../db/migrate/20170918125918_ids_to_bigints.rb:69:in `<class:IdsToBigints>'
.../db/migrate/20170918125918_ids_to_bigints.rb:3:in `<top (required)>'
```

### DIFF
--- a/db/migrate/20170918125918_ids_to_bigints.rb
+++ b/db/migrate/20170918125918_ids_to_bigints.rb
@@ -5,70 +5,70 @@ class IdsToBigints < ActiveRecord::Migration[5.1]
 
   disable_ddl_transaction!
 
-  INCLUDED_COLUMNS = [
-    [:account_domain_blocks, :account_id],
-    [:account_domain_blocks, :id],
-    [:accounts, :id],
-    [:blocks, :account_id],
-    [:blocks, :id],
-    [:blocks, :target_account_id],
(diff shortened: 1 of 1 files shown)


ThibG:
It looks ok to me, but I'm not sure how that's changing anything?

zunda:
It seems to me that the exception is raised at `migration.rb:14.` in the definition of `method_missing(method, *args)` method:

```
      @checker.perform(method, *args) do
```

before the instance variable `@checker` is initialized. While I haven't fully followed it, it seems to me happening when `table_exists?(:deprecated_preview_cards)` is evaluated while the file is being loaded.

This change is to defer the evaluation until when `up` or `down` method is called so that `@checker` will have been initialized.

ClearlyClaire (approved):

PR #11378, opened 2019-07-22
Bind servers to 0.0.0.0 in Procfile

zunda:
Following #11302 and #11326, this change makes it easier to deploy Mastodon as a Heroku app.

The second line is to be used for a separate app to receive streaming requests.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
-web: bundle exec puma -C config/puma.rb
+web: BIND=0.0.0.0 bundle exec puma -C config/puma.rb
+#web: BIND=0.0.0.0 node ./streaming # for streaming app
 worker: bundle exec sidekiq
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -1,2 +1,3 @@
-web: bundle exec puma -C config/puma.rb
+web: BIND=0.0.0.0 bundle exec puma -C config/puma.rb
+#web: BIND=0.0.0.0 node ./streaming # for streaming app
```
Gargron: Why commented out, and why not call it "streaming"?

zunda: Heroku only routes requests from the Internet at ports 80 or 443 to "web" dynos.

To run the streaming server on Heroku, a separate app that shares Postgres and Redis with the main app is needed. On that app, the node process needs to run on "web" dynos.

Gargron: Okay. I would format it like this:

```Procfile
web: BIND=0.0.0.0 bundle exec puma -C config/puma.rb
worker: bundle exec sidekiq

# For the streaming API, you need a separate app that shares Postgres and Redis with
# the main one, and runs this dyno:
# web: BIND=0.0.0.0 node ./streaming 
```

zunda:
Cool. I'll try to come up with more through procedure while keeping it
short here.

Procfile
zunda: I pushed d7d15d4 so that the admin doesn't have to edit the `Procfile`. How does this look to you?

Gargron (approved):

PR #11393, opened 2019-07-23
Change account domain block to clear out notifications and follows

Gargron:
- Remove notifications from blocked domain in the UI
- Remove notifications from blocked domain in the backend
- Remove followed users from blocked domain
- Do not allow following users from a domain you blocked
- Remove DMs from muted/blocked accounts in the UI
- Remove suggestions from muted/blocked/domain blocked accounts in the UI

### DIFF
--- a/app/javascript/mastodon/reducers/conversations.js
+++ b/app/javascript/mastodon/reducers/conversations.js
@@ -8,6 +8,8 @@ import {
   CONVERSATIONS_UPDATE,
   CONVERSATIONS_READ,
 } from '../actions/conversations';
+import { ACCOUNT_BLOCK_SUCCESS, ACCOUNT_MUTE_SUCCESS } from 'mastodon/actions/accounts';
+import { DOMAIN_BLOCK_SUCCESS } from 'mastodon/actions/domain_blocks';
 import compareId from '../compare_id';
 
 const initialState = ImmutableMap({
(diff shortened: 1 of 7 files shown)


app/javascript/mastodon/reducers/conversations.js
```diff
@@ -74,6 +76,10 @@ const expandNormalizedConversations = (state, conversations, next, isLoadingRece
   });
 };
 
+const filterConversations = (state, accountIds) => {
+  return state.update('items', list => list.filterNot(item => item.get('accounts').some(accountId => accountIds.includes(accountId))));
+};
```
ThibG: Hm, I'm not still 100% sure about the “hide everything mentioning blocked users” logic (although it's not new).
People mentioning both you and someone you blocked might not know you've blocked that last person.
Also, not having notifications in those cases isn't new, but I think not being able to see the message at all is new? I'd have to check.

app/javascript/mastodon/reducers/conversations.js
```diff
@@ -96,6 +102,11 @@ export default function conversations(state = initialState, action) {
 
       return item;
     }));
+  case ACCOUNT_BLOCK_SUCCESS:
+  case ACCOUNT_MUTE_SUCCESS:
+    return filterConversations(state, [action.relationship.id]);
```
ThibG: Same as above.

ClearlyClaire (approved):
Front-end changes are consistent with the backend, so in that regard, LGTM.
However, I'm still iffy about hiding every toot merely *mentioning* someone you've blocked or muted, especially if that toot is addressed to you (and even more so, by a mutual). Maybe that should be configurable.
Also, I'm still iffy about domain mutes preventing you from having normal interactions with someone from the domain you mute, which is inconsistent with how instance-wide domain mutes work. (But that would be a different PR)

PR #11394, opened 2019-07-23
Change settings to show restricted accounts only what they have access to

ClearlyClaire:
Follow-up to 964ae8eee593687f922c873fa7b378bb6e3e39bb


Gargron:
I'm not sure about this, because maybe it's good to show new users what sorts of menu items there are that they will be able to access in the future. On the other hand, it might also be frustrating that they all don't work.

ThibG:
I guess we could add some styling and disable those options instead of outright hiding them, but that's more work and I'm not sure it's worth it.

Either way, I think it's definitely better not showing non-working links as if they were working

Gargron (approved):

PR #11405, opened 2019-07-25
Fix boosting & unboosting preventing a boost from appearing in the TL

ClearlyClaire:
Currently, if the person being the first to boost a toot in your TL unboosts it, that toot cannot appear in your TL anymore until the origin boost id falls of the 40 toots thing. This fixes that.

When aggregating boosts, if the boost that ended up in the TL is deleted, the code will reinsert the oldest boost known, but any streaming client won't know about it. I am not convinced this behavior makes sense, but it's not new, and I'm not sure how it should be handled exactly.

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -35,7 +35,7 @@ class FeedManager
   end
 
   def unpush_from_home(account, status)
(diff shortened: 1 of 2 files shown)


ThibG:
The code style issues raised by codeclimate were already there

Gargron (approved):

PR #11421, opened 2019-07-28
Prevent internet archiving when user set "noindex"

tribela:
When user set "Opt-out of search engine indexing", It doesn't prevent internet archive like _archive.org_.
This patch would prevent that kind of indexing


Gargron:
Shouldn't the archive abide by noindex?

Kjwon15:
> Shouldn't the archive abide by noindex?

@Gargron No, I checked that is not prevented by `noindex`.

https://web.archive.org/web/20190728054221/https://qdon.space/@jarm

Gargron (approved):

PR #11424, opened 2019-07-28
Revert "Remove conversation URI"

Gargron:
Reverts tootsuite/mastodon#11423


connyduck:
but why

Gargron:
Unfortunately @ThibG pointed out that this hacked-on non-standard attribute is the only way to make conversation mutes work beyond unavailable replies (blocked servers, private off-server posts) and out-of-order arrival of replies.

trwnh:
>hacked-on non-standard attribute

isn't it part of ostatus? and also de-facto bolted onto a lot of activitypub implementations -- pleroma, friendica, and honk, off the top of my head. (also https://github.com/w3c/activitystreams/issues/476 / https://github.com/w3c/activitystreams/issues/476#issuecomment-420720257 as well)

Gargron:
Yeah it was part of OStatus which never went past a draft stage and was never intended to be used in ActivityPub. There is no official JSON-LD context that provides that property, so we're hardcoding a hacky embedded definition with every payload. And yes, it seems a lot of implementations are doing the same. It is frustrating because it seems to be necessary for a very basic thing and yet that thing was never considered in the ActivityPub spec itself. Originally it was meant to bridge the gap when Mastodon was transitioning from OStatus to ActivityPub and all the old posts had a different format of IDs. I am quite frustrated that we might be stuck with it forever now.

trwnh:
i think pleroma and honk use the same value for both `conversation` and `context`, so if you're looking for something that's actually in activitypub then it's probably `context`. https://www.w3.org/TR/activitystreams-vocabulary/#dfn-context

personally i think "conversation" is more clear as to what the property does, and "context" is a bit too vague (intentionally so, but vague nonetheless).

PR #11426, opened 2019-07-28
Change links in webUI to rewrite misleading links

ClearlyClaire:
Prominently show link origin host whenever the text doesn't show it.

image

Styling should be improved, at the very least by using a color variable. Ideally, this should probably be opt-out. Also, relying merely on color might not be a good idea accessibility-wise.

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -40,7 +40,7 @@ const isLinkMisleading = (link) => {
   const targetURL = new URL(link.href);
 
   // The following may not work with international domain names
-  if (linkText === targetURL.origin || linkText === targetURL.host || linkText.startsWith(targetURL.origin + '/') || linkText.startsWith(targetURL.host + '/')) {
+  if (linkText === targetURL.origin || linkText === targetURL.host || 'www.' + linkText === targetURL.host || linkText.startsWith(targetURL.origin + '/') || linkText.startsWith(targetURL.host + '/')) {
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


no-boot-device:
It probably shouldn't be styled like that, perhaps something like reddit where it's grey and small? Therefore people can still see it, buuuut

ThibG:
Using the same color as for the handle in the status header:
image

no-boot-device:
Does this work for punycode? Should punycode be expanded or left alone?

ThibG:
As it relies on the `URL` API to do the URL parsing, it will expand punycode, so that will only work, so it will indeed show the domain… hm… maybe we shouldn't use that for parsing

Psycojoker:
> Using the same color as for the handle in the status header:
> image

I'm really expecting this one to be an accessibility issue for to low contrast :/ (there are some testers online but I don't have links right now, sorry x()

ThibG:
> > Using the same color as for the handle in the status header:
> > image
> 
> I'm really expecting this one to be an accessibility issue for to low contrast :/ (there are some testers online but I don't have links right now, sorry x()

Yeah, I have the same concerns, although if it is an accessibility issue, so is the display of the user's handle, it's the exact same color on exactly the same background. So if there's a fix needed, it can be applied to both instances.

ThibG:
There has been a suggestion to not change anything to the toot for links which text do not look like an auto-generated link (as it would not be *misleading*), and to entirely rewrite links which text do look like one.

This might help in reducing visual clutter, but I'm weary of rewriting the toot in non-obvious ways. Furthermore, it seems pretty difficult to tell what “looks like an auto-generated link”, as a domain name could also be a filename, for instance (but maybe that's an edge case that doesn't matter much).

ThibG:
Thinking a bit more about it, figuring out whether something “looks like an auto-generated link” sounds pretty difficult to do, as it should match all domain names and *stuff that looks like domain names*, including for instance dot homographs (․).

In a previous commit, I also allowed all international domain names, but it may make sense to use something like https://wiki.mozilla.org/IDN_Display_Algorithm to allow only punycode for some script combinations.

ThibG:
I implemented it so that it rewrites any link whose text can be confused with a (shortened) URL and doesn't match the target, and leave everything else as is.

Gargron (approved):

PR #11444, opened 2019-07-29
let bundler fetch http_parser.rb submodules

ghost:
Motavation for this change:

The http_parser.rb gem uses git submodules and fetches them during the build. In situations where the dependencies for mastodon are fetched upfront, and a network connection is intentionally unavailable during the build, this will make the http_parser.rb gem build fail.
This is the case when packaging Mastodon with the Nix package manager.
Adding the submodules attribute makes bundler fetch the submodules upfront instead of them being fetched during the build process.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -22,6 +22,7 @@ GIT
   remote: https://github.com/tmm1/http_parser.rb
   revision: 54b17ba8c7d8d20a16dfc65d1775241833219cf2
   ref: 54b17ba8c7d8d20a16dfc65d1775241833219cf2
+  submodules: true
   specs:
     http_parser.rb (0.6.1)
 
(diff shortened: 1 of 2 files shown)


Gargron:
No, it is related:

    The git source https://github.com/tmm1/http_parser.rb is not yet checked out. Please run `bundle install` before trying to start your application

petabyteboy:
It was failing with a different, unrelated error first (fuubar 2.4.0 could not be fetched). Now I rebased on master and it gives this different error. I will look into it later.

Gemfile.lock
```diff
@@ -22,6 +22,7 @@ GIT
   remote: https://github.com/tmm1/http_parser.rb
   revision: 54b17ba8c7d8d20a16dfc65d1775241833219cf2
   ref: 54b17ba8c7d8d20a16dfc65d1775241833219cf2
+	submodules: true
```
Shleeble: space here

petabyteboy: Thanks for catching that mistake!

petabyteboy:
The build failure was caused by the whitespace error @Shleeble found, now the build passes.

Gargron (approved):

PR #11446, opened 2019-07-29
Change adjacent notifications about the same status to display as one

### LINKED ISSUE #7698: Combine retoots and favorites in notification column
If you toot something that becomes popular, your notification column becomes full of share / favorite notification for this unique toot and it is very easy to miss an other important event.

Instead of the following:

- Liam liked 'some popular toot'
- Noah shared 'some popular toot'
- Emma liked 'some popular toot'
- William shared 'some popular toot'
- **Olivia answered to 'an other toot'**
- James shared 'some popular toot'
- Sophia shared 'some popular toot'
- Ava liked 'some popular toot'
- Logan liked 'some popular toot'

I find 3 main behaviors at this time to combine toots:

1. Combine notifications of same toot and same kind:

- Liam liked 'some popular toot'
- Noah shared 'some popular toot'
- Emma liked 'some popular toot'
- William shared 'some popular toot'
- **Olivia answered to 'an other toot'**
- James and Sophia shared 'some popular toot'
- Ava and Logan liked 'some popular toot'

2. Combine notifications of same toot, whatever the order of the notification kind:

- Liam and Emma liked 'some popular toot'
- Noah and William shared 'some popular toot'
- **Olivia answered to 'an other toot'**
- James and Sophia shared 'some popular toot'
- Ava and Logan liked 'some popular toot'

3. Combine notification with one big group, even if an event appears between popular notifications:

- James, Sophia and 2 other people shared 'some popular toot'
- Ava, Logan and 2 other people liked 'some popular toot'
- **Olivia answered to 'an other toot'**

Note that:
- considering the chronological order, `1.` is more accuracy than `2.` and `2.` is more accuracy than `3.`;
- considering the number of notifications, `1.` is more annoying than `2.` and `2.` is more annoying than `3.`;
- `3.` can be more difficult to implement because we need to know when the user last logged in.
- for both of the behaviors, I suggest here to display *and x other people* if there is more than 2 people, but maybe this number could me reduced to 1 in order to fit in the notification column;

* * * *

Gargron:
Fix #7698

### DIFF
--- a/app/javascript/mastodon/features/notifications/index.js
+++ b/app/javascript/mastodon/features/notifications/index.js
@@ -20,6 +20,22 @@ const messages = defineMessages({
   title: { id: 'column.notifications', defaultMessage: 'Notifications' },
 });
 
+const collapsibleNotifications = (a, b) => {
+  if (!['reblog', 'favourite'].includes(a.get('type'))) {
+    return false;
+  }
+
+  return a.get('type') === b.get('type') && a.get('status') === b.get('status');
+};
+
+const reduceNotifications = (list, notification) => {
+  if (!list.isEmpty() && collapsibleNotifications(list.last(), notification)) {
+    return list.update(list.size - 1, item => item.update('collapsed_account_ids', ImmutableList(), collapsed_account_ids => collapsed_account_ids.push(notification.get('account'))));
+  } else {
+    return list.push(notification);
+  }
+};
+
 const getNotifications = createSelector([
   state => state.getIn(['settings', 'notifications', 'quickFilter', 'show']),
   state => state.getIn(['settings', 'notifications', 'quickFilter', 'active']),
@@ -27,12 +43,13 @@ const getNotifications = createSelector([
   state => state.getIn(['notifications', 'items']),
 ], (showFilterBar, allowedType, excludedTypes, notifications) => {
   if (!showFilterBar || allowedType === 'all') {
-    // used if user changed the notification settings after loading the notifications from the server
+    // Used if user changed the notification settings after loading the notifications from the server
     // otherwise a list of notifications will come pre-filtered from the backend
     // we need to turn it off for FilterBar in order not to block ourselves from seeing a specific category
-    return notifications.filterNot(item => item !== null && excludedTypes.includes(item.get('type')));
+    return notifications.filterNot(item => item !== null && excludedTypes.includes(item.get('type'))).reduce(reduceNotifications, ImmutableList());
   }
-  return notifications.filter(item => item !== null && allowedType === item.get('type'));
+
+  return notifications.filter(item => item !== null && allowedType === item.get('type')).reduce(reduceNotifications, ImmutableList());
 });
 
 const mapStateToProps = state => ({
--- a/app/javascript/mastodon/features/notifications/components/notification.js
+++ b/app/javascript/mastodon/features/notifications/components/notification.js
@@ -17,6 +17,8 @@ const notificationForScreenReader = (intl, message, timestamp) => {
   return output.join(', ');
 };
 
+const appendDisplayNames = (base, accounts) => <span>{base} <abbr title={accounts.map(a => `@${a.get('acct')}`).join(', ')}><FormattedMessage id='notification.and_n_others' defaultMessage='and {count, plural, one {# other} other {# others}}' values={{ count: accounts.size }} /></abbr></span>;
+
 export default @injectIntl
 class Notification extends ImmutablePureComponent {
 
@@ -144,6 +146,10 @@ class Notification extends ImmutablePureComponent {
   renderFavourite (notification, link) {
     const { intl } = this.props;
 
+    if (notification.get('collapsed_accounts')) {
+      link = appendDisplayNames(link, notification.get('collapsed_accounts'));
+    }
+
     return (
       <HotKeys handlers={this.getHandlers()}>
         <div className='notification notification-favourite focusable' tabIndex='0' aria-label={notificationForScreenReader(intl, intl.formatMessage({ id: 'notification.favourite', defaultMessage: '{name} favourited your status' }, { name: notification.getIn(['account', 'acct']) }), notification.get('created_at'))}>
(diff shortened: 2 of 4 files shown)


ThibG:
If someone favs and boosts the same toot, it will display “user and 1 other boosted your status”, which is wrong. Furthermore, it doesn't solve the issue of two toots being fav/boosted alternatively.

Gargron:
> Furthermore, it doesn't solve the issue of two toots being fav/boosted alternatively.

I don't think that's an issue that's meant to be solved. It makes sense to me to merge similar notifications when they're next to each other, but just merging *everything* from the entire column into one notification seems excessive to me.

ThibG:
I mean it's frequent to have a few trending toots that are fav/boosted in succession (especially in threads), so whenever that is the case, we will lose any benefit from this PR.

Shleeble:
Thanks for looking into this. Could you take a screenshot of how this looks?

trwnh:
Could this be disabled?

Gargron:
We could just not merge it and close the issues asking for it. I don't mind individual notifications.

trwnh:
i'm sure a lot of people would appreciate having combined notifications, even if you or i are not part of those people

app/javascript/mastodon/features/notifications/index.js
```diff
@@ -20,19 +20,36 @@ const messages = defineMessages({
   title: { id: 'column.notifications', defaultMessage: 'Notifications' },
 });
 
+const collapsibleNotifications = (a, b) => {
+  if (!['reblog', 'favourite'].includes(a.get('type'))) {
+    return false;
+  }
+
+  return a.get('type') === b.get('type') && a.get('status') === b.get('status');
```
ThibG: Works, although now, a single toot being alternatively boosted and favourited (which is very frequent for trending toots) will prevent collapsing.

ThibG:
> We could just not merge it and close the issues asking for it. I don't mind individual notifications.

This is difficult to do, but I definitely think done correctly it would be a massive improvement for people choosing to display favs/boosts without making their notifications useless whenever a few of their toots are trending.

Shleeble:
My original view was bundling all favs and all boosts as separate notifications, but joining both of those things together sounds interesting.

Cassolotl:
This is a significant change for average users, and average users are told that if they want to have input on development they should come here to the Github list, but when we do that and ask for screenshots so that we can be involved we are outright ignored.

I second @Shleeble's request for screenshots.

rtucker:
> We could just not merge it and close the issues asking for it. I don't mind individual notifications.

This is why we want you and other white techbros to fuck off back to Twitter Junior and let us drive, Gargron.

noiob:
I really don't care that much about the chronology of favs and boosts if a toot gets very popular, but I care about my notifications columns staying readable. I like the way Twitter does it (bundling favs and boosts separately every now and then). Right now the only way I can still read replys while a toot is "going viral" is when I completely hide boosts and favs.

Also this could totally be a checkbox in the notification column settings

Gargron:
When there's something to show i add a screenshot without prompting. There is nothing to see here. You get one notification instead of multiple in a row. It says "alice and 123 others boosted your toot". It's the same text as before just with "and 123 others" added in the middle.

Cassolotl:
It's not the lack of screenshot in the first post that I object to, it's ignoring people who ask questions.

Gargron:
I don't think that combining everything about the same toot in a single notification without interruptions is a good idea. If the goal is for notifications to be readable when something is going viral, you have the quick tab bar to switch to mentions-only. I think all of this grouping is sterilizing the feeling you get from interacting with other people. Like it becomes just a rising number instead of a log of real people showing you that they care.

noiob:
I don't think having my column spammed to the point where I have to ignore favs/ boosts to be able to read replies makes my social media interactions more meaningful in any way, but go on

Twitter shows a row of profile pics you can hover to check who interacted with your toot.

Shleeble:
Sorry for the lack of spec/expections: In my head, grouped notifications such as `"alice and 123 others boosted your toot` is perfect.... but there must be an easy way to listing/viewing those 123 people?

Could we have a modal to popup and list all likes/boosts? or maybe a link back to the toot itself and we can view them easily

https://aus.social/web/statuses/102529608531895992/favourites
https://aus.social/web/statuses/102529608531895992/reblogs

for example

Shleeble:
@Gargron any option of this being opt-in? if you dont wanna enable by default?

arturo182:
f-ing hell, if this is how y'all respond to someone submitting a change that is simple, could be opt-in, and would be useful for many people, then mastodon is dead on arrival

olivierlacan:
@Gargron I think your points about the emotional impact of grouping multiple notifications as a "viral count" are actually very good. 

I'm here because I just had a post reblogged/boosted suddenly quite a lot and it made my ability to navigate my notifications much harder as a result. So while I understand your reticence to accidentally incentivize virality by giving people dopamine slot-machines (posting in hopes of getting reblogged), there's an undeniable poster experience side-effect to consider.

Mastodon is growing, more than it has ever grown. You will start to see cracks in the day-to-day experience of users manifest as power users and technically savvy folks requesting features like this. I think it's reasonable to:
- allow instance admins to disable notification grouping by default (perhaps as part of a future suite of feature all designed to tame attention-seeking impulses)
- default all notifications to ungrouped regardless of instance admins allowing it, simply so that newcomers aren't subtly given the hint that not enough people are liking/boosting their posts because they see `2 people reboosted: ` instead of discrete notifications
- whenever multiple notifications for the same status are displayed in the UI above an arbitrary threshold (let's say 5), display a small (forever dissmissable) notice indicating that users *can* group notifications temporarily if they wish

I can confidently predict this will make life easier for folks who occasionally say things that capture broad attention. Sure, we don't necessarily want to encourage this but having a useless notifications should be something we can solve while keeping in mind its potential impacts.

Last note, since grouped counts (as opposed to long lists as IIRC Tumblr displays interactions) are displayed at the bottom of posts, it only makes sense to make that notifications would also allow grouping.

image

PS: I would gladly help working on first adding the configuration support for this proposed feature, prior to enabling it.

PR #11449, opened 2019-07-31
Fix jumping of toot date when clicking spoiler button

ariasuni:
Fix the weird behavior you can see in this gif:

Peek 31-07-2019 02-53

The date of the toot moves 2px down and up uselessly when clicking the spoiler button.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -731,7 +731,7 @@
     white-space: pre-wrap;
 
     &:last-child {
-      margin-bottom: 0px;
+      margin-bottom: 0;
     }
   }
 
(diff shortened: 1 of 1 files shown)


ThibG:
This 2px margin was introduced by you in #10821
So the fix would probably be to somehow add that 2px margin when the toot is hidden as well

ariasuni:
I do not have the issue with the cut letters anymore, and I felt weird about adding a margin for that in the first place.

Can you confirm that letters aren’t cut anymore with:
```css
.reply-indicator__content p:last-child, .status__content p:last-child {
    margin-bottom: 0;
}
```

and that my previous workaround is not necessary anymore?

ThibG:
I'm indeed unable to reproduce the issue fixed by #10821 without that rule, hmm

ClearlyClaire (approved):

PR #11459, opened 2019-08-01
Add uploaded files size to admin dashboard

### LINKED ISSUE #11401: admin dashboard should show allocated media size
### Pitch

Admin dashboard (and submenus) give a good overview of the instance. 
There is, for example, `SPACE USAGE`, which summarizes Psql and Redis.
However I hadto realise (testing a new instance) that 90% of the space was actually used up by `…/live/public/system/media_attachments` directory, and it's size isn't indicated anywhere I could find.

### Motivation

When the disk space is gone the admin shall be shown why. It is obvious by using console and doing a full `du` but maybe it's not optimal.

Gargron:
Fix #11401

Fix redis-namespace deprecation warning about administrative commands


ykzts (approved):

Gargron:
Three `SUM` calls on big tables might make the dashboard a bit too slow to load... Caching would help but only on repeated loads which is not very helpful. What should I do?

ykzts:
How about not to execute the process until the admin requests it? (For example, use API in JavaScript)

PR #11461, opened 2019-08-01
Add query_track to find slow SQL queries

kirillshevch:
I suggest adding a gem `query_track` which will be show in the console info about SQL queries which slower than *1s* (for example, this is configurable). The reason similar to using `pghero`. Developers will be easier to profile the database and improve performance.

Example of log (from another project):
console


Gargron:
You will likely not encounter slow queries during development because such issues are not apparent with small tables. In production, such log output will likely not be monitored by most people. PGHero has the advantage of providing a tabular view of historic data instead of a live feed. I'm not sure this is a useful addition...

nightpool:
agree that pg_hero is probably better for this use-case.

PR #11469, opened 2019-08-02
Add nightly recounting of tag scores and account counters

### LINKED ISSUE #11403: Incorrect followers count (local account)
### Expected behaviour

Followers count = actual number of followers

### Actual behaviour

A local profile shows a low number of followers while it has in fact way more followers.

See https://mstdn.io/users/d1cor/followers. A few days ago the account showed "1 follower". Currently it shows 2 followers, while the account has 107 followers including local and foreign accounts.

### Specifications

```
mastodon@mstdn:~/live$ git status
HEAD detached at v2.9.2
nothing to commit, working tree clean
```

I'm not sure if this is a bug or something is wrong with my configuration/database. Feel free to ask me to run commands in the rake console or the database.

FYI, I remember this happening for other accounts with the toot count for example.

Gargron:
In a daily scheduler, put the number of unique users of each hashtag from the past 7 days into the hashtags' score. Use a formula that uses both this score and the length difference between the input and the given result so that a high score could rank higher than a super short match.

Add `last_status_at` to hashtags and update it once per day to only recount the ones that had activity in the last 7 days.

Also, recount the number of followers, statuses, and follows for all active users to fix #11403.

### DIFF
--- a/app/workers/scheduler/counting_scheduler.rb
+++ b/app/workers/scheduler/counting_scheduler.rb
@@ -13,7 +13,7 @@ class Scheduler::CountingScheduler
   private
 
   def recount_tag_scores!
-    Tag.find_each do |tag|
+    Tag.active.find_each do |tag|
       tag.score = Redis.current.pfcount(*(0..7).map { |i| "activity:tags:#{tag.id}:#{i.days.ago.beginning_of_day.to_i}:accounts" })
       tag.save if tag.changed?
     end
--- a/app/workers/scheduler/counting_scheduler.rb
+++ b/app/workers/scheduler/counting_scheduler.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class Scheduler::CountingScheduler
+  include Sidekiq::Worker
+
+  sidekiq_options unique: :until_executed, retry: 0
+
+  def perform
+    recount_tag_scores!
+    recount_account_counters!
+  end
+
+  private
+
+  def recount_tag_scores!
+    Tag.find_each do |tag|
+      tag.score = Redis.current.pfcount(*(0..7).map { |i| "activity:tags:#{tag.id}:#{i.days.ago.beginning_of_day.to_i}:accounts" })
+      tag.save if tag.changed?
+    end
+  end
+
+  def recount_account_counters!
+    Account.joins(:user).merge(User.active).includes(:account_stat).find_each do |account|
+      account.account_stat.following_count = account.active_relationships.count
+      account.account_stat.followers_count = account.passive_relationships.count
+      account.account_stat.statuses_count  = account.statuses.where.not(visibility: :direct).count
+
+      account.account_stat.save if account.account_stat.changed?
+    end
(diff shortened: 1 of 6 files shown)


app/workers/scheduler/counting_scheduler.rb
```diff
@@ -0,0 +1,17 @@
...
+  sidekiq_options unique: :until_executed, retry: 0
+
+  def perform
+    recount_tag_scores!
+  end
+
+  private
+
+  def recount_tag_scores!
+    Tag.find_each { |tag| tag.update(score: Redis.current.pfcount(*(0..7).map { |i| "activity:tags:#{tag.id}:#{i.days.ago.beginning_of_day.to_i}:accounts" })) }
```
ThibG: Isn't that going to be a *lot* of queries?
On my small single-user instance alone, there are 37626 known tags.

Gargron: While there will be 37626 fetches from Redis, in the new code an update will only be issued if the numbers differ. If both are 0, such as in the likely majority of cases, the only tags that will be updated will be the ones currently in use.

mayaeh: I tried this code on my instance (taruntarun.net).
(The result of Tag.count is 56109.)

I ran CountingScheduler by pressing the “Enqueue now” button on Sidekiq.
The first run took 3 minutes to finish.
The second run was 1 minute.
However, CPU usage increased about 50% during that time.

Gargron: I think 3 minutes of 50% CPU is acceptable? I suspect that it's the Redis pfcount-with-merge operation causing the CPU spike, but I do not have a solution for it.

mayaeh: (Personal opinion) I think it is acceptable.
This is less than my server scheduled `pg_dump`'s CPU usage.

mayaeh (changes requested):

config/sidekiq.yml
```diff
@@ -30,6 +30,9 @@
   backup_cleanup_scheduler:
     cron: '<%= Random.rand(0..59) %> <%= Random.rand(3..5) %> * * *'
     class: Scheduler::BackupCleanupScheduler
+  counting_scheduler:
+    cron: '<%= Random.rand(0..59) %> <%= Random.rand(20..0) %> * * *'
```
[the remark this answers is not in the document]
mayaeh: @Gargron What about this PR?
Was my operation wrong?

Maybe changing `20..0` to `20..24` will work.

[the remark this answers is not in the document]
Gargron: Sorry, I forgot about this! Yes, my mistake. @ThibG said that he wasn't happy with the nightly recounting approach though, and we are not sure what else can be done instead.

angristan:
In the meantime, maybe we could have a tootctl command to manually recount everything? Which we would only use after a follow-bot-instance ban

PR #11473, opened 2019-08-03
Add account bio to account admin view

ClearlyClaire:
image

### DIFF
--- a/app/views/admin/accounts/show.html.haml
+++ b/app/views/admin/accounts/show.html.haml
@@ -3,8 +3,33 @@
 
 = render 'application/card', account: @account
 
-.admin-account-info
-  = render 'accounts/bio', account: @account, admin_view: true
+- account = @account
+- proofs = account.identity_proofs.active
+- fields = account.fields
+- unless fields.empty? && proofs.empty? && account.note.blank?
+  .admin-account-bio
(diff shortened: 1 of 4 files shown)


app/views/admin/accounts/show.html.haml
```diff
@@ -1,8 +1,14 @@
 - content_for :page_title do
   = @account.acct
 
+- proofs = @account.identity_proofs.active
+- fields = @account.fields
```
ykzts: need?

ThibG: It's not needed. Sorry, this was leftover from a previous attempt at that PR!
Removed it

ykzts (approved):

ThibG:
@Gargron ping

Gargron:
It's really stretched out to the point it's hard to read, can we do something about that? I'm not sure it's necessary since I always click to just open the real profile

ThibG:
Good point, I'll try thinking about a better layout

ThibG:
Updated so it looks like this:

image

Sylvhem:
Yes! Yes, pretty please. I constantly have to open users' profiles to see if their bio is not full of sponsored links. This would save me so much time and half the clicks.

PR #11489, opened 2019-08-05
Bump faker from 1.9.6 to 2.1.0 and update faker api

unasuke:
Faker released v2 that has several breaking changes.

https://github.com/stympy/faker/releases/tag/2.0

Then, original patch ( #11485 ) is failing because using deprecated api.

I update faker api to use namred paramater (v2 style).


Gargron (approved):

unasuke:
Hmm, why doesn't the test run?

unasuke:
Thanks!

PR #11490, opened 2019-08-05
Change admin UI for hashtags and add back whitelisted trends

### LINKED ISSUE #271: Popular or trending hashtags
I think it would be cool to be able to see some of the most popular hashtags of the last 24 hours. It could fit nicely at the top of the "public" panel.

Gargron:
Fix #271

Add back the `GET /api/v1/trends` API with the caveat that it does not return tags that have not been allowed to trend by the staff.

When a hashtag begins to trend (internally) and that hashtag has not been previously reviewed by the staff, the staff is notified.

image

The new admin UI for hashtags allows filtering hashtags by where they are used (e.g. in the profile directory), whether they have been reviewed or are pending reviewal, they show by how many people
the hashtag is used in the directory, how many people used it today, how many statuses with it have been created today, and it allows fixing the name of the hashtag to make it more readable.

grafik

The disallowed hashtags feature has been reworked. It is now controlled from the admin UI for hashtags instead of from the file `config/settings.yml`


ykzts (approved):

rigow:
This MUST be configurable for instances. It may be scalable for a corporate mastodon server, but it is certainly not scalable for public instances like mamot.fr. Instead there should be an option for a user NOT to see toots with a certain hashtag. Because the end-user counts in public servers. While a corporate server may have a fulltime moderator to do all that, a public instance will just bottleneck the flow of information and people will find alternatives to hashtags which sensibly distorts the UX

PR #11491, opened 2019-08-05
Improve dropdown menu keyboard navigation

ClearlyClaire:
- Handle space bar in addition to `Enter`
- Fix modals closing immediately
- Trap Tab/shift+Tab
- Give back focus when closing the menu


ThibG:
The dropdown actually gives focus back to the dropdown icon when it closes rather than to the element which had it previously.
The fix would be the same as in #11492 so I'll wait for that PR to be merged before fixing that.

ykzts (approved):

PR #11492, opened 2019-08-05
Improve keyboard navigation in privacy dropdown

ClearlyClaire:


ThibG:
Hm, the focus isn't given back to the element that had it before the click, but to the icon button… I should probably fix that, but it will be a bit more involved

Gargron (approved):

PR #11493, opened 2019-08-05
Trap tab in modals

ClearlyClaire:
I am not totally sure about this one: I'm not sure the order is correct, and ideally, we probably would want to only act on the first and last items in the tab order


Gargron (approved):

ThibG:
There's actually one regression in the report modal with this on. Focus is attempted on polls' hidden input, which fails, and so focus won't cycle past this… I am not too sure how to fix that

ThibG:
It now works in any modal I can think of, although I'm not super happy with manually checking this kind of things.

PR #11499, opened 2019-08-06
Fix image uploads being perfectly white when canvas read access is blocked

### LINKED ISSUE #11496: Images larger than 1280² pixels are downsized to a white image when using the Tor Browser
### Expected behaviour

When an user uploads an image larger than 1280² pixels it should be downsized, on Firefox, Chrome and the Tor Browser.

### Actual behaviour

When an user uploads an image larger than 1280² pixels using the Tor Browser, the image is downsized to a white image.

_Note: This is about servers on the normal web, but accessed via the Tor Browser (so not hosted as a hidden service). I didn't tested this on a Mastodon server as a hidden service._

This is tested on Todon.nl and Mastodon.social. On regular browsers like Firefox or Chrome (using the normal web) the images are uploaded fine.

Install the most recent Tor Browser and upload a large image (e.g. 2592x1952). The image will be white.

### Specifications

- Mastodon v2.9.2 (Todon.nl + Mastodon.social)
- Tor Browser 8.5.4 on Ubuntu Linux 19.04

ClearlyClaire:
Fixes #11496

Note that some versions of the Tor Browser fail earlier in `loadImage`, but I couldn't fix it.

### DIFF
--- a/app/javascript/mastodon/utils/resize_image.js
+++ b/app/javascript/mastodon/utils/resize_image.js
@@ -67,6 +67,14 @@ const processImage = (img, { width, height, orientation, type = 'image/png' }) =
 
   context.drawImage(img, 0, 0, width, height);
 
+  // The Tor Browser and maybe other browsers may prevent reading from canvas
+  // and return an all-white image instead. Assume reading failed if the resized
+  // image is perfectly white.
+  const imageData = context.getImageData(0, 0, width, height);
+  if (imageData.every(value => value === 255)) {
+    throw 'Failed to read from canvas';
+  }
+
   canvas.toBlob(resolve, type);
 });
 


Gargron:
It's not impossible that someone might upload a totally white image at some point. I think reading all the pixels one more time is also not great. I suggest using a "canary" check instead. Like try creating a hardcoded all-black, small canvas and reading from it, then check if all the pixels are white. This can also be done once per page load to confirm whether the canvas method is available, instead of on each image upload.

ThibG:
This will stop at the first non-255 component, which is very likely to occur very fast except for some specific computer graphics. Also, this check is performed on the resized image, which is at most 1280×1280 in size.

It is true that someone may want to upload a totally white image, in which case it will upload it, just the non-resized version of it…

Doing it just once per page instead of more might make sense, but also, this means the resizing code wouldn't be called ever again even if the user accepts canvas access (as on Tor Browser and Firefox with appropriate privacy settings, it will return an all-white image as long as the user hasn't accepted it—the user prompt does not block the call)

Gargron (approved):

PR #11515, opened 2019-08-07
Add domain block notes

ClearlyClaire:
This PR adds a way to write comments regarding the reasons of instance blocks.
It also improves handling of upgrading an instance block, and allows editing (even downgrading) an instance block as long as it is triggered from the new “edit domain block” button.

It does not add anything to the admin API, as that API doesn't seem to cover domain blocks.

It does not actually add a way to display reasons publicly, as that would be dependent on #11298

# Creating a new domain block

image

# Displaying a domain block

image

# Editing an existing domain block

image

### DIFF
--- a/app/controllers/admin/domain_blocks_controller.rb
+++ b/app/controllers/admin/domain_blocks_controller.rb
@@ -53,7 +53,7 @@ module Admin
     end
 
     def resource_params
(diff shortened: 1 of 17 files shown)


Gargron:
Closed by accident?

ThibG:
Yeah, I didn't even notice I closed it :|

Gargron (approved):

PR #11519, opened 2019-08-07
Add GIF and WebP support for custom emojis

### LINKED ISSUE #11466: Custom gif emojis from Pleroma are not federated
### Pitch
Pleroma uses gif for custom emoji but Mastodon only uses apng, as a result, gif emojis from Pleroma remain in full text (shortcode) instead of displaying the image.

### Motivation
This currently brings a poor UX and users think that applications are broken. It could be nice to have at least non animated images.

Gargron:
Fix #11466

### DIFF
--- a/app/models/custom_emoji.rb
+++ b/app/models/custom_emoji.rb
@@ -28,6 +28,8 @@ class CustomEmoji < ApplicationRecord
     :(#{SHORTCODE_RE_FRAGMENT}):
     (?=[^[:alnum:]:]|$)/x
 
+  IMAGE_MIME_TYPES = %w(image/png image/gif image/webp).freeze
+
   belongs_to :category, class_name: 'CustomEmojiCategory', optional: true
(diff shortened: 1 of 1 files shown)


app/models/custom_emoji.rb
```diff
@@ -28,14 +28,16 @@ class CustomEmoji < ApplicationRecord
...
+
   belongs_to :category, class_name: 'CustomEmojiCategory', optional: true
   has_one :local_counterpart, -> { where(domain: nil) }, class_name: 'CustomEmoji', primary_key: :shortcode, foreign_key: :shortcode
 
   has_attached_file :image, styles: { static: { format: 'png', convert_options: '-coalesce -strip' } }
 
   before_validation :downcase_domain
 
-  validates_attachment :image, content_type: { content_type: 'image/png' }, presence: true, size: { less_than: LIMIT }
+  validates_attachment :image, content_type: IMAGE_MIME_TYPES, presence: true, size: { less_than: LIMIT }
```
ThibG: `{ content_type: IMAGE_MIME_TYPES }` I believe

ClearlyClaire (approved):

PR #11522, opened 2019-08-08
Fix “read more” button behing hidden (regression from #11404)

ClearlyClaire:
This has the side-effect of putting the “Read more” button below possibly
trunctated polls instead of putting the poll below the “Read more”

### DIFF
--- a/app/javascript/mastodon/components/status_content.js
+++ b/app/javascript/mastodon/components/status_content.js
@@ -166,11 +166,6 @@ export default class StatusContent extends React.PureComponent {
     }
   }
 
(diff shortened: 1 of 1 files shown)


ThibG:
Also, I think I misunderstood this functionality, as clicking that button opens the toot in detailed view, while I thought it would simply unfold the toot in place. Especially since there's code for that, in `handleCollapsedClick`, but it's dead code. It should either be removed or used by the `readMoreButton` instead of `this.props.onClick`.

Gargron (approved):

PR #11525, opened 2019-08-08
Fix account search always returning exact match on paginated results

### LINKED ISSUE #11365: /api/v2/search will always return ones own account, no matter the offset
TL;DR searching for your own account breaks pagination

### Expected behaviour

When I search for my own username with Search v2 api and a really high offset, I get no results.

```https://mastodon.social/api/v2/search?type=accounts&q=connyduck&resolve=true&limit=20&offset=300&following=false```

```
{
    "accounts": [],
    "statuses": [],
    "hashtags": []
}
```

### Actual behaviour

```https://mastodon.social/api/v2/search?type=accounts&q=connyduck&resolve=true&limit=20&offset=300&following=false```

```
{
    "accounts": [
        {
            // my own account
        }
    ],
    "statuses": [],
    "hashtags": []
}
```

My own account is always returned, no matter what offset I send in.

Send a search request for your own username and any offset. 

### Specifications

Could reproduce on multiple mastodon instances, all running v2.9.2.
Problem does not occur when the search query does not return my own account.

Gargron:
Fix #11365

### DIFF
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -21,18 +21,22 @@ class AccountSearchService < BaseService
     if resolving_non_matching_remote_account?
       [ResolveAccountService.new.call("#{query_username}@#{query_domain}")].compact
     else
-      search_results_and_exact_match.compact.uniq.slice(0, limit)
+      search_results_and_exact_match.compact.uniq
     end
   end
 
   def resolving_non_matching_remote_account?
-    options[:resolve] && !exact_match && !domain_is_local?
+    offset.zero? && options[:resolve] && !exact_match? && !domain_is_local?
   end
 
   def search_results_and_exact_match
-    exact = [exact_match]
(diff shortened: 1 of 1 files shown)


ThibG:
Won't the first request return up to `limit + 1` results, and clients would have to use that as `offset`, but that would skip one database result? And later down the line, the exact match could appear a second time

Gargron:
Okay, first request should now return exactly `limit` results. To fix the exact match possibly re-appearing later, the account search query methods need to refactored to be composable.

ykzts (approved):

ThibG:
I think this still skips the first result of the second pagination page? Say limit is 3, with one exact match, the service would return, for the first query: the exact match and 2 other results (results 0 and 1 from SQL)

when performing a second query, the client would use an offset of 3, that would return results 3 and 4 from the SQL query, missing the second one

PR #11527, opened 2019-08-08
Add cashtags

### LINKED ISSUE #7634: Support dollar sign as toot tags.
This is used by many cryptocurrency users on Twitter, who would like to use tags like `$BTC`, `$ETH`, `$XMR`, `$ETC`, etc.

* * * *

Gargron:
This may be unnecessary but it's rather easy to add and not disruptive... Cashtags are hashtags in the format `$ABC`, they only work for uppercase ascii letters and only exactly 3 characters long, mostly because those were the examples given in the feature request.

Fix #7634


Kiyonari:
Genuine question: What would it be used for ?

PR #11532, opened 2019-08-10
EO trad: correct lines related to polls + add missing translation

Vanege:


Vanege:
Actually, should I redo this edits using Crowdin to simplify the workflow?

Vanege:
Actually I would rather see merged the edits on Crowdin, so I will close the PR

PR #11534, opened 2019-08-11
Fix some emojis in profile metadata labels are not emojified.

kedamaDQ:
To use emojis in the profile, the emojis returned by /api/v1/accounts/:id must contain it.


Gargron (approved):

kedamaDQ:
thanks for merging!

PR #11552, opened 2019-08-12
Include max image dimensions in error

raboof:
Untested as I don't have my dev machine sufficiently set up yet - but presumably CI would touch this? :)

### DIFF
--- a/app/models/concerns/attachmentable.rb
+++ b/app/models/concerns/attachmentable.rb
@@ -43,7 +43,7 @@ module Attachmentable
 
       width, height = FastImage.size(attachment.queued_for_write[:original].path)
 
-      raise Mastodon::DimensionsValidationError, "#{width}x#{height} images are not supported" if width.present? && height.present? && (width * height >= MAX_MATRIX_LIMIT)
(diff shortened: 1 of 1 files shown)


app/models/concerns/attachmentable.rb
```diff
@@ -43,7 +43,7 @@ def check_image_dimensions
 
       width, height = FastImage.size(attachment.queued_for_write[:original].path)
 
-      raise Mastodon::DimensionsValidationError, "#{width}x#{height} images are not supported" if width.present? && height.present? && (width * height >= MAX_MATRIX_LIMIT)
+      raise Mastodon::DimensionsValidationError, "#{width}x#{height} images are not supported, w*h must be under #{MAX_MATRIX_LIMIT}" if width.present? && height.present? && (width * height >= MAX_MATRIX_LIMIT)
```
Gargron: There has to be a more readable way to communicate the matrix limit than `w*h`. Perhaps: `...images are not supported, must be below #{MAX_MATRIX_LIMIT} sqpx` (square pixels)?

Gargron (changes requested):

app/models/concerns/attachmentable.rb
raboof: either is an improvement, updated as suggested.

Gargron (approved):

PR #11556, opened 2019-08-12
Require at least 1 accepted follow from the remote domain for AUTHORIZED_FETCH mode.

BenLubar:
This means that remote instances will at least be *visible* to the admin.

### DIFF
--- a/app/controllers/concerns/signature_verification.rb
+++ b/app/controllers/concerns/signature_verification.rb
@@ -61,6 +61,12 @@ module SignatureVerification
       return
     end
 
+    unless Follow.where(account: Account.where(domain: account.domain)).exists?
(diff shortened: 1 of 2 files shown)


TemporaryUsername12481:
This effectively kills discoverability, should be an option disabled by default if implemented at all.

ClearlyClaire (changes requested):
That would at the very least require a configuration option.

Also, as is, the change would effectively break federation, as it would reject the follow requests to begin with.

ThibG:
Oh, and remote instances *are* visible to the admin with `AUTHORIZED_FETCH`: each instance signs stuff with an actor, so that actor is visible in the list of remote actors.

Gargron:
A random follower on some local account is not anymore visibile to the admin than the existence of a remote account in the local database, as is already the case. And this would be a really breaking change. So I do not think this PR should be merged.

PR #11560, opened 2019-08-12
Add indicator of unread content to window title when web UI is out of focus

### LINKED ISSUE #1288: Change the title  and/or the favicon when getting new notifications
Hello there,

I've been trying Mastodon for a few days now, and one feature I really wish I had would be to make new content/notifications more obvious. I know there is already a subtle sound when you get a notification, but it'd be great if the page title and/or the favicon could also update.

This is especially useful when using pinned tab, because those will get highlighted on Firefox when the title changes (probably on Chrome too, haven't tried it). Slack and Gmail have this very simple feature, and it makes it super easy to know if you should even check out the tab. Twitter does something similar but it's much less useful since it happens for every single tweet on the feed.

HTML5 notifications would be nice too, but I've never really used them outside of slack.

* * * *

Gargron:
Fix #1288

### DIFF
--- a/app/javascript/mastodon/actions/app.js
+++ b/app/javascript/mastodon/actions/app.js
@@ -1,0 +1,10 @@
+export const APP_FOCUS   = 'APP_FOCUS';
+export const APP_UNFOCUS = 'APP_UNFOCUS';
+
+export const focusApp = () => ({
+  type: APP_FOCUS,
+});
+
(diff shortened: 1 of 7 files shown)


ykzts (approved):

harrybailey:
Looks like this has since been removed.
So this would now be a suggestion, or can be closed.

PR #11563, opened 2019-08-14
Add media editing modal

### LINKED ISSUE #6713: Make adding image descriptions more comfortable
There are two things that I've been thinking about:

– Make the option more visible. I've heard from a few people that they didn't realise they could add a caption until someone told them, because they never hovered over their images.

– Make it more comfortable to write a description that is longer than the input field width. When I, for example, post a picture of a poster that has some text on it, I tend to switch back and forth between my browser, and a different window where I look at the picture to copy the text. But each time I switch back, only the beginning of the image description is visble and I have to use the arrow key to get back to the end.

It would be cool to either have it remember the position (the way it /sometimes/ does when you're only switching tabs or smth, couldn't quite figure it out) or make the field get bigger, so the whole text is visible (that would be really nice when doing corrections as well, it really helps to see the whole thing), and make it stay big when you switch around, too.

### LINKED ISSUE #8320: Give the alternate text field its own modal window
Currently, after having uploaded a media file, users have the possibility to add an alternative text to it.
Sadly, there is, in my opinions, three problems with the current implementation:
1. The field used for the input is really small, which make it difficult to use.
2. The field isn't a true field text, and spell checking isn't available as a result.
3. When you click outside the field and then go back to it, the cursor isn't positioned at the end but at the begging of the text. This is both quite annoying and impractical.

I would suggest that, instead of having the field right on the image, we use an icon and a small snippet of text (something like "Add alternate text") that, when clicked, open a modal containing the image and a proper text field. In that way, it would be similar to the "Crop" function we already have.
I think something along the lines of Twitter's implementation of this feature (see below) is what we should aim at.

Twitter's image description modal window.

* * * *

Gargron:
Move media description input to a modal and unite that modal with the focal point modal. Add a hint about choosing focal points, as well as a preview of a 16:9 thumbnail. Enable the user to watch the video next to the media description input.

image

Fix #8320
Fix #6713

### DIFF
--- a/app/javascript/mastodon/features/compose/components/upload.js
+++ b/app/javascript/mastodon/features/compose/components/upload.js
@@ -4,16 +4,11 @@ import PropTypes from 'prop-types';
 import Motion from '../../ui/util/optional_motion';
 import spring from 'react-motion/lib/spring';
 import ImmutablePureComponent from 'react-immutable-pure-component';
-import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
+import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 import Icon from 'mastodon/components/icon';
 
-const messages = defineMessages({
-  description: { id: 'upload_form.description', defaultMessage: 'Describe for the visually impaired' },
-});
-
-export default @injectIntl
-class Upload extends ImmutablePureComponent {
+export default class Upload extends ImmutablePureComponent {
 
   static contextTypes = {
     router: PropTypes.object,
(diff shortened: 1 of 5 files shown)


app/javascript/mastodon/features/compose/components/upload.js
```diff
@@ -55,69 +36,21 @@ class Upload extends ImmutablePureComponent {
...
-
-    this.setState({ focused: false, dirtyDescription: null });
-
-    if (dirtyDescription !== null) {
-      this.props.onDescriptionChange(this.props.media.get('id'), dirtyDescription);
-    }
-  }
-
   render () {
     const { intl, media } = this.props;
```
ykzts: I think `intl` is no longer needed.

ykzts: https://github.com/tootsuite/mastodon/pull/11563/files#diff-251ca9aae04b4c9b0b9722b6698c5d9eL11 is the same

ykzts (changes requested):

ykzts (approved):

ThibG:
This is less than optimal on mobile
Screen Shot 2019-08-14 at 11 09 02

The preview text and image prevent you from seeing the image appropriately or clicking to set the focus point, and there's a scrollbar

PR #11569, opened 2019-08-15
Fix trending counter width

brawaru:
Trending counter used to be constant 100px in width, which caused
issues in languages like Russian, where because of that, "talking"
text was cut to the size where actual count is not visible at all:

> 6 people talking
> Популярно у...

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -6006,12 +6006,12 @@ noscript {
 
     &__current {
       flex: 0 0 auto;
-      width: 100px;
       font-size: 24px;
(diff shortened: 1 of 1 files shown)


Sasha-Sorokin:
Hope did everything right :worried:

#### Screenshots for reference:

| Before | After |
|:------:|:-----:|
| Screenshot_20190815_092935 | Screenshot_20190815_093041 |

Tested with Firefox:

<details>
<summary>CSS changes</summary>

```patch
.trends__item__current {
- width: 100px;
+ margin-left: 5px;
}
```
</details>

ykzts (approved):

PR #11572, opened 2019-08-15
Change title bar to only show unread notifications

Gargron:


ykzts (approved):

witcheslive:
yes, please! an option to disable it entirely would also be nice

PR #11537, opened 2019-08-11
Add more accurate account search

Gargron:
When ElasticSearch is available, a more accurate search is implemented:

- Using edge n-gram index for acct and display name
- Using asciifolding and cjk width normalization on display names
- Using Gaussian decay on account activity for additional scoring (recency)
- Using followers/friends ratio for additional scoring (spamminess)
- Using followers number for additional scoring (size)

Additionally, the previous behaviour is also kept:

- Prioritizing people you follow (closeness)

The exact match precedence only takes effect when the input conforms to the username format and the username part of it is complete, i.e. when the user started typing the domain part.

**What does this mean?** Searching for people who have unicode characters in their display names, such as accents and umlauts, has become possible without entering those characters (asciifolding); likewise, you no longer need to use the exact flavour of CJK characters (full-width/half-width) to get the same results; accounts that haven't been active in a long time are less likely to appear near the top; accounts that are follow-botting are less likely to appear near the top (spamminess), and more established accounts (by followers number) are more likely to appear higher (but mind, that's within specific searches). Like before, the biggest factor is whether you're following someone.

___

### Evaluations

I decided to run different queries for the same search term to see the differences between results and how they are scored. The queries did not consider follow relationships because a user that is followed by the searching user will universally appear at the top of the results.

#### Search term "garg"

Expecting to find: "Gargron"

Here is the status quo, search results from PostgreSQL for comparison:

|Document score|Completion|
|--------------|----------|
|0.583333|Gargy|
|0.583333|gargosoft@instance.business|
|0.583333|Gargo@kinky.business|
|0.583333|gargath@canislupus.im|
|0.583333|gargantia@peertube.mastodon.host|
|0.583333|gargoyle@witchcraft.cafe|
|0.583333|gargamel@botsin.space|
|0.583333|Gargantia|
|0.583333|gargronbot|
|0.5|AnvitZero|

The first result is a dead account, the rest are mostly bots.

Now, let's begin with a simple query without any additional scoring:

|Document score|Completion|
|--------------|----------|
|26.223595|garg|
|24.804735|Gargy|
|23.99578|gargan|
|23.249702|Gargron|
|22.583332|gargle|
|22.343393|gargrant|
|22.343393|Gargabob|
|21.893581|Gargoyle|
|21.849594|Garg0yle|
|21.849594|gargirl_|

The third result is an active account, the rest, not so much. They're also all local accounts.

Same query with follower ratio affecting the score:

|Document score|Completion|Follower ratio|
[...]

PR #11579, opened 2019-08-15
Add more accurate hashtag search

Gargron:
Similar to #11537, use edge n-grams to autocomplete tags. Additional scoring is provided through:

- Gaussian decay is used on `last_status_at`
- Last 7 days of unique uses

Only hashtags that have been reviewed and are listable can appear in searches.


ykzts (approved):

mayaeh:
Maybe it works strangely if the hashtag contains numbers or languages ​​other than English

Gargron:
Okay, I do not want to merge it until I understand why that happens

PR #11582, opened 2019-08-16
Fix multiple issues with replies collection for pages further than self-replies

### LINKED ISSUE #11568: Publish valid replies in "replies" collection [JSON][ActivityPub]
### Pitch

Currently only self-replies are included. I propose including all valid replies.

### Motivation

The current thread construction behavior is flawed, and intersects with a lot of issues that have presented long-standing UX issues: #8565, #9409, #10916, #11503 (particularly https://github.com/tootsuite/mastodon/issues/11503#issuecomment-518748342 as it interfaces with #9409), #10865, and possibly more.

Including all valid replies is the first step to being able to query `replies` (which is needed for the above issues).

ClearlyClaire:
Fixes #11568

### DIFF
--- a/app/controllers/activitypub/replies_controller.rb
+++ b/app/controllers/activitypub/replies_controller.rb
@@ -38,7 +38,7 @@ class ActivityPub::RepliesController < ActivityPub::BaseController
       type: :unordered,
       part_of: account_status_replies_url(@account, @status),
       next: next_page,
-      items: @replies.map { |status| status.local ? status : status.id }
+      items: @replies.map { |status| status.local ? status : status.uri }
     )
 
     return page if page_requested?
--- a/app/controllers/activitypub/replies_controller.rb
+++ b/app/controllers/activitypub/replies_controller.rb
@@ -27,7 +27,7 @@ class ActivityPub::RepliesController < ActivityPub::BaseController
   end
 
   def set_replies
-    @replies = page_params[:other_accounts] ? Status.where.not(account_id: @account.id) : @account.statuses
+    @replies = page_params[:only_other_accounts] ? Status.where.not(account_id: @account.id) : @account.statuses
     @replies = @replies.where(in_reply_to_id: @status.id, visibility: [:public, :unlisted])
     @replies = @replies.paginate_by_min_id(DESCENDANTS_LIMIT, params[:min_id])
   end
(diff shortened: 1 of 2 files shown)


app/controllers/activitypub/replies_controller.rb
```diff
@@ -38,7 +38,7 @@ def replies_collection_presenter
       type: :unordered,
       part_of: account_status_replies_url(@account, @status),
       next: next_page,
-      items: @replies.map { |status| status.local ? status : status.id }
+      items: @replies.map { |status| status.local ? status : status.uri }
```
Gargron: For consistency shouldn't this use TagManager?

app/serializers/activitypub/note_serializer.rb
```diff
@@ -55,7 +55,7 @@ def replies
         type: :unordered,
         part_of: ActivityPub::TagManager.instance.replies_uri_for(object),
         items: replies.map(&:second),
-        next: last_id ? ActivityPub::TagManager.instance.replies_uri_for(object, page: true, min_id: last_id) : nil
+        next: last_id ? ActivityPub::TagManager.instance.replies_uri_for(object, page: true, min_id: last_id) : ActivityPub::TagManager.instance.replies_uri_for(object, page: true, other_accounts: true)
```
Gargron: I feel like `other_accounts` is an unhelpful param name. I have no idea what it does. Initially I assumed it would be an array of accounts!

Gargron (changes requested):

app/controllers/activitypub/replies_controller.rb
ThibG: I'm not sure I see the use for that. I don't think calling TagManager for the remote case would make things clearer. The local case returns the full status inlined, not only its identifier.

app/serializers/activitypub/note_serializer.rb
ThibG: What should it be, then? When it isn't set, the endpoint will return self-replies only, when it is set, it will return only replies from other accounts.

Gargron: I think `only_self_replies` could be a good alternative

app/controllers/activitypub/replies_controller.rb
Gargron: Mh okay then

app/serializers/activitypub/note_serializer.rb
ThibG: `only_other_accounts` to not change the logic?

Gargron: Okay.

Gargron (approved):

Issue #11536, opened 2019-08-11
Edit profile page crashes when trying to set WebP file as profile image


imbsky:
### Expected behaviour

If we really support WebP file, we should be able to set it as profile image.

### Actual behaviour

It shouldn't crash.

### Steps to reproduce the problem

After set WebP file as avatar or header image, press the save changes button.

### Specifications

- Mastodon version: latest master
- OS: macOS 10.14.6
- Browser: Chrome 76.0.3809.87

angristan:
The edit profile page says: 

> PNG, GIF or JPG. At most 2 MB. Will be downscaled to 400x400px

So yeah, it should either really support webp or reject it without crashing 🤔

angristan:
It crashes beaucase:

```
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae] Command :: file -b --mime '/tmp/1c48b7459dc9291c12b8a93b4857a41820190816-363-1gs8gux.webp'
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae] Command :: identify -format %m '/tmp/1c48b7459dc9291c12b8a93b4857a41820190816-363-aca78l.webp[0]'
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae] method=PUT path=/settings/profile format=html controller=Settings::ProfilesController action=update status=500 error='Paperclip::Error: There was an error running `identify` for 1c48b7459dc9291c12b8a93b4857a41820190816-363-aca78l' duration=424.26 view=0.00 db=3.61
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae]
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae] Paperclip::Error (There was an error running `identify` for 1c48b7459dc9291c12b8a93b4857a41820190816-363-aca78l):
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae]
Aug 16 22:39:39 mstdn bundle[209]: [de489f24-8ac7-4a83-966b-3980c02be6ae] lib/paperclip/lazy_thumbnail.rb:17:in `make'
[...]

PR #11589, opened 2019-08-17
Remove WebP support

angristan:
Add discussed in #11536, webp support is not stable (i.e. webp does not work at all), even when adding missing dependencies (https://github.com/tootsuite/mastodon/issues/11538).

Paperclip/ImageMagick don't handle webp well, which results in errors and even crashes. I tested webp support on multiple instances so it is not an isolated case.

Until we have a fully working webp support, it is best to remove it. 

This PR reverts https://github.com/tootsuite/mastodon/pull/9879 and half of https://github.com/tootsuite/mastodon/pull/9879.


Gargron (approved):

tateisu:
some admins edit policy.xml of ImageMagick, it restrict delegate process..

PR #11591, opened 2019-08-17
Add HTTP signature keyId to request log

Gargron:
Also, the stoplight was not handling `OpenSSL::SSL::SSLError` exceptions.


ykzts (approved):

ThibG:
When would this actually log anything?
Am I right to understand this would only log tracebacks and such, which you appear to have removed in #11590?

Gargron:
No, it's part of the request log--path, format, time spent in db etc.

Issue #11360, opened 2019-07-19
Mass remote follow being used as attack vector


fsnk:
![maliciousintent](https://user-images.githubusercontent.com/3524203/61521778-488ee100-aa11-11e9-8f8b-702439b4eb95.png)

The command being used is:
![remote_follow_DOS](https://user-images.githubusercontent.com/3524203/61521333-47a97f80-aa10-11e9-855a-3709c74e1a7e.png)

(Screenshot taken from the attacker's feed; I haven't personally tried it :) )

Presumably the initial intent was just to artificially inflate follower numbers for specific accounts, but since posting this has been actively used against several servers. 

Unclear whether simply suspending the remote server is sufficient to prevent impact as the admin interface states "The domain block will not prevent creation of account entries in the database" or whether firewall rules must also be employed.

mason1900:
The attacker seems to be using the following commands:
```bash
RAILS_ENV=production bin/tootctl accounts follow anyone@site.tld
RAILS_ENV=production bin/tootctl accounts create name --email e@ma.il --condirmed
```

kaniini:
the feature itself is useful, but I think it should probably be scoped to local accounts only ;)

MasTorDon:
never mind him look at this https://gab.com/WeedIsLife/posts/102486741169444123

it appears there is a bigger threat. someone has been DDoSing him. they could come after any of us next

z3usy:
https://FreeFediFollowers.ga/ is exploiting this in the wild. A suspend seems to slowly clear all followers from all accounts.

rlywtf:
https://FreeFediFollowers.ga/ is actively exploiting this in the wild. I recommend fellow fediverse nodes add a domain block for them.

p.s. Adding a domain suspension (also enabling both block options under silence prior to selecting suspend in the moderation panel) will kick off sidekiq jobs and slowly remove all follows from all accounts. 😀

PR #11592, opened 2019-08-17
Limit "tootctl accounts follow" to local accounts

angristan:
To (somewhat) limit mass remote follow: #11360.

I don't know if the code is good, I figured it's the simplest way to prevent remote follows.

But after all, the code is already out there. So "attackers" can modify this file themselves, make their own fork, not update, revert this commit, etc. It does not actually fix this mass follow issue. 

But is this feature actually useful in the first place? I mean it is useful for local accounts, but I don't see why an admin would force a remote follow.

Anyway I made this PR for fun so feel free to discard it ✌️


angristan:
Oops, gonna fix those quotes.

Gargron (approved):

PR #11595, opened 2019-08-18
Add featured hashtags as an ActivityPub collection

Gargron:
Resolve #10478

New collection `featuredTags` containing `Hashtag` objects.

### DIFF
--- a/app/controllers/activitypub/collections_controller.rb
+++ b/app/controllers/activitypub/collections_controller.rb
@@ -12,7 +12,7 @@ class ActivityPub::CollectionsController < ActivityPub::BaseController
 
   def show
     expires_in 3.minutes, public: public_fetch_mode?
-    render_with_cache json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, skip_activities: true
+    render_with_cache json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter
   end
 
   private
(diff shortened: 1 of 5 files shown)


ykzts (approved):

ThibG:
While from a protocol standpoint this sounds somewhat fine to me, I'm pretty sure this will break any current Mastodon instance.

Indeed, Mastodon currently assume every item is either a `Note`'s URI, or an object with a `Note`'s URI:

https://github.com/tootsuite/mastodon/blob/4e921832272425352d28cad550bfc4dffd6d0e78/app/services/activitypub/fetch_featured_collection_service.rb#L25-L30

It may be best to have different collections for this

Gargron:
@thibG Since the Mention and Hashtag objects do not have an `id` property and are not strings, the initial map will return `nil` for those items. Then the nil will be passed to `ActivityPub::FetchRemoteStatusService`, which should, in turn, return nil, which would finally be removed by `compact`. Isn't that fine? It means older versions won't break?

Gargron:
Correction: Passing nil into `ActivityPub::FetchRemoteStatusService` throws an `ArgumentError` from the `Request` class. Very unfortunate. It would have worked if not for that one bit. :(

Gargron:
I've decided to rework this to use a separate collection after all. I do *not* like making up another ActivityPub vocabulary term but maybe it's for the best :man_shrugging:

app/controllers/activitypub/collections_controller.rb
```diff
@@ -12,25 +12,17 @@ class ActivityPub::CollectionsController < ActivityPub::BaseController
 
   def show
     expires_in 3.minutes, public: public_fetch_mode?
-    render_with_cache json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter, skip_activities: true
+    render_with_cache json: collection_presenter, content_type: 'application/activity+json', serializer: ActivityPub::CollectionSerializer, adapter: ActivityPub::Adapter
```
ThibG, line 15: `skip_activities` doesn't do anything anymore? when have we removed it?

app/controllers/activitypub/collections_controller.rb
```diff
@@ -12,25 +12,17 @@ class ActivityPub::CollectionsController < ActivityPub::BaseController
...
-          []
-        else
-          cache_collection(@account.pinned_statuses, Status)
-        end
-      end
+      @items = for_signed_account { cache_collection(@account.pinned_statuses, Status) }
+    when 'tags'
+      @items = for_signed_account { @account.featured_tags }
     when 'devices'
       @items = @account.devices
```
ThibG, line 27: I guess this case could use `for_signed_account` too although it's not really a big deal either.

ClearlyClaire (approved):
I have made a few inline comments, but none of them prevent this PR from being merged.

app/controllers/activitypub/collections_controller.rb
Gargron: I can't even find where it was used... It doesn't seem to appear anywhere else in the code.

PR #11603, opened 2019-08-18
Add TLS v1.3 support

ghost:
Transport Layer Security (TLS) 1.3 was released in August 2018 and has several improvements over TLS 1.2 (faster handshakes and security improvements).

This would allow to maintain TLS v1.2 compatibility (might want to drop this later) and add support for TLS v1.3.

In order to use TLS v1.3 it is necessary to have Nginx 1.13.0 (or greater) built against OpenSSL 1.1.1 (or greater).

Instances would most likely keep communicating between each other via TLS 1.2 because of Ruby's http.rb gem, but the connection between the instance and its users is going to be faster and more secure.

This change would only affect new installations unless the nginx steps of the configuration process are repeated by the admins of existing instances.

EDIT: I have now been running with this change for 5 days on my instance and I have encountered no issues.

### DIFF
--- a/dist/nginx.conf
+++ b/dist/nginx.conf
@@ -19,7 +19,7 @@ server {
   listen [::]:443 ssl http2;
   server_name example.com;
 
-  ssl_protocols TLSv1.2;
+  ssl_protocols TLSv1.2 TLSv1.3;
   ssl_ciphers HIGH:!MEDIUM:!LOW:!aNULL:!NULL:!SHA;
   ssl_prefer_server_ciphers on;
   ssl_session_cache shared:SSL:10m;


angristan:
Does the `HIGH` cipher contain TLS 1.3 ciphers?

ichi-i:
> Does the `HIGH` cipher contain TLS 1.3 ciphers?

It should as far as I know. I am no SSL expert, but after reading docs and running tests on https://www.ssllabs.com the following TLS v1.3 cipher suites are being returned:
- TLS_AES_256_GCM_SHA384 (0x1302)
- TLS_CHACHA20_POLY1305_SHA256 (0x1303)
- TLS_AES_128_GCM_SHA256 (0x1301)

angristan:
Indeed:

```sh
~# openssl version
OpenSSL 1.1.1  11 Sep 2018
~# openssl ciphers -V 'HIGH:!MEDIUM:!LOW:!aNULL:!NULL:!SHA' | grep TLSv1.3
          0x13,0x02 - TLS_AES_256_GCM_SHA384  TLSv1.3 Kx=any      Au=any  Enc=AESGCM(256) Mac=AEAD
          0x13,0x03 - TLS_CHACHA20_POLY1305_SHA256 TLSv1.3 Kx=any      Au=any  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0x13,0x01 - TLS_AES_128_GCM_SHA256  TLSv1.3 Kx=any      Au=any  Enc=AESGCM(128) Mac=AEAD
```

angristan (approved):

umonaca:
I have tested this on our production server and saw no side-effects. Firefox is showing TLS 1.3 instead of TLS 1.2. Federation still works because TLS 1.2 is in the config as fallback.

Gargron (approved):

PR #11605, opened 2019-08-19
Check if document.getElementById('tabs-bar__portal') is null

zunda:
While viewing LTL or FTL in multi-column mode, Javascript raises TypeError while narrowing the window:

shonbori

While I'm not fully sure what I'm doing, this change seems to prevent the error.

### DIFF
--- a/app/javascript/mastodon/components/column_header.js
+++ b/app/javascript/mastodon/components/column_header.js
@@ -178,7 +178,9 @@ class ColumnHeader extends React.PureComponent {
     if (multiColumn || placeholder) {
       return component;
     } else {
-      return createPortal(component, document.getElementById('tabs-bar__portal'));
+      var e = document.getElementById('tabs-bar__portal');
+      if (e === null) return component;
+      return createPortal(component, e);
     }
   }
 


ThibG:
I am afraid of two things with this change:
- That instead of crashing, it would just display the column header within the column instead of above it (which admittedly is better than crashing, but would still lead ton inconsistent experience and be hard to debug)
- That it may also happen for the `ColumnBackButton` component

ThibG:
The root issue is that the portal container and its content can be created in the same React render cycle, so whenever that happens, the target of the portal isn't in the DOM yet. This is obviously an issue, and I guess there ought to be a better way to achieve this, but I can't think of one.

Your patch does indeed prevent a crash and it does indeed make the header scroll within the column as opposed to stay on top like it should normally do.

One workaround I can think of is to add something to force the component to render again as long as the portal target doesn't exist, with something like:
```javascript
this.setState({ forceReRender: !this.state.forceReRender });
```

app/javascript/mastodon/components/column_header.js
```diff
@@ -178,7 +178,9 @@ class ColumnHeader extends React.PureComponent {
     if (multiColumn || placeholder) {
       return component;
     } else {
-      return createPortal(component, document.getElementById('tabs-bar__portal'));
+      var e = document.getElementById('tabs-bar__portal');
+      if (e === null) return component;
```
ThibG: This branch should force the component to re-render, as otherwise, the column header will not scroll properly.

ClearlyClaire (changes requested):

zunda:
@ThibG thanks so much for looking into this! It seems that the problem is far too much for my knowledge around React. May I close this PR and create an Issue so that you'd be able to take time to come up with a better solution?

ThibG:
Or I could add the workaround I suggested in this PR, if you don't mind?

zunda:
Reported as an Issue at #11630 and closing this. Cheers!

PR #11622, opened 2019-08-19
Fix statuses being rejected because of invalid hashtag names

### LINKED ISSUE #11618: Don't recognize PeerTube posts with hashtags containing invalid characters
Some of PeerTube's posts are no longer recognized by Mastodon again.
This is due to invalid characters in the hashtag.

Example: Blender's `blender 3d` tag
https://video.blender.org/search?tagsOneOf=blender%203d

### Expected behaviour

Invalid hashtags are ignored.
https://github.com/tootsuite/mastodon/issues/6261

### Actual behaviour

422 error occurs.

Enter the following video URL in the search field.
https://video.blender.org/videos/watch/5053eaea-93bb-43d6-9320-d7ebfeeefe2b

### Specifications

This problem was recurred by https://github.com/tootsuite/mastodon/pull/11416

ClearlyClaire:
Fixes #11618

This was caused by `Tag#find_or_create_by_names` using `create` and not `create!` as previously.
I could also have changed that, but `find_or_create_by_names` is used in more places I haven't investigated yet.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -149,7 +149,7 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     return if tag['name'].blank?
 
     Tag.find_or_create_by_names(tag['name']) do |hashtag|
-      @tags << hashtag unless @tags.include?(hashtag)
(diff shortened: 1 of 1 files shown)


Gargron:
Uh, #11621?

PR #11623, opened 2019-08-19
Add soft delete for statuses for instant deletes through API

Gargron:
- Fix bug where same status was fetched multiple times in the unreblog API
- Fix bug where reblogs of sensitive statuses were showing media without spoiler in admin UI


ThibG:
What are the performance implications of that?
Also, is it possible that `SuspendAccountService` may skip deleting discarded statuses?

Gargron:
> What are the performance implications of that?

Not entirely sure, unfortunately. PG picks different query plans depending on table size so getting real performance estimates in development is hard.

> Also, is it possible that `SuspendAccountService` may skip deleting discarded statuses?

Yes, but because discarding and queueing removal worker happens at the same time, I feel like we don't need to care about those.

Gargron:
I have updated the index for the slowest query, getting an account's statuses. I believe the queries for public timelines will not benefit from other changes.

db/migrate/20190820003045_update_statuses_index.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateStatusesIndex < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured { add_index :statuses, [:account_id, :id, :visibility, :updated_at], where: 'deleted_at IS NULL', order: { id: :desc }, algorithm: :concurrently, name: :index_statuses_20190820 }
+    remove_index :statuses, name: :index_statuses_20180106
+  end
```
ThibG: I don't fully understand this, so I can't comment on that, and I don't understand the performance implications, which I am worried about.

Other than that, looks fine to me

Gargron: So, the index is made smaller by limiting it to rows that have a null deleted_at. Because the index is mainly used for queries that ask for rows where the deleted_at column is null (as that is the default scope), it will work like before, and not need to add a further filter for deleted_at in the query planner.

ykzts (approved):

PR #11624, opened 2019-08-19
Polish translation fix

101010pl:
Swapped GDPR and COPPA.


Gargron:
Please use https://crowdin.com/projects/mastodon, otherwise the change will be overwritten on next update.

PR #10079, opened 2019-02-20
Implement NodeInfo endpoint

ghost:
This addresses #6739 

This is my attempt to implement the node info protocol found here https://github.com/jhass/nodeinfo .  I have also followed pleroma's lead and tried to add similar metaData.

You can see this changes in action on my instance
**https://banana.dog/.well-known/nodeinfo**
```json
{
  "links": [
    {
      "rel": "http://nodeinfo.diaspora.software/ns/schema/2.0",
      "href": "https://banana.dog/nodeinfo/2.0"
    },
    {
      "rel": "http://nodeinfo.diaspora.software/ns/schema/2.1",
      "href": "https://banana.dog/nodeinfo/2.1"
    }
  ]
}
```

**https://banana.dog/nodeinfo/2.1**
```json
{
  "version": "2.1",
  "usage": {
    "users": {
      "total": 58,
      "activeHalfyear": 26,
      "activeMonth": 14
    },
    "localPosts": 11932
  },
  "software": {
    "version": "2.7.1.cerberus",
    "name": "mastodon",
    "repository": "https://github.com/usbsnowcrash/mastodon"
  },
  "services": {
    "outbound": [],
    "inbound": []
  },
  "protocols": [
    "ostatus",
    "activitypub"
  ],
  "openRegistrations": true,
  "metaData": {
    "nodeName": "Banana.dog",
[...]

PR #11628, opened 2019-08-20
Add nodeinfo endpoints

ClearlyClaire:
Displays basic nodeinfo metadata, with optional domain block info.

Based on https://github.com/florence-social/mastodon-fork/pull/134 with a few fixes and adaptations


nightpool (changes requested):
my comments on https://github.com/tootsuite/mastodon/pull/10079 have not changed. nodeinfo is an unspecified mess, and no-one has provided a compelling use case for it.

>Having something that a lot of servers implement is nice, but I wouldn't want to cargo cult a bad standard just because it's popular. nodeinfo is basically haphazardly structured json with a couple of arbitrary properties—of the servers that you implement it, i would love to see what selection of properties each actually implements. For example, I know pleroma has a ton of information in their metadata field that's not actually standardized by nodeinfo—and your proposed solution does as well. How is that any better then the existing state of the world?

> At least in the current state of the world you can never get confused because mastodon's nodeName means a slightly different thing then pleroma's nodeName

app/serializers/node_discovery_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+  attributes :links
+
+  def links
+    [
+      {
+        rel: 'http://nodeinfo.diaspora.software/ns/schema/2.0',
+        href: node_info_schema_url('2.0'),
+      },
+      {
+        rel: 'http://nodeinfo.diaspora.software/ns/schema/2.1',
```
nightpool: From #10079: there is no `http://nodeinfo.diaspora.software/ns/schema/2.1`. there is no official diaspora nodeinfo 2.1 spec, and the unofficial draft spec should not use a diaspora namespace without consent from diaspora (which i have not seen) 

https://github.com/tootsuite/mastodon/pull/10079#discussion_r258334251

app/serializers/node_info_serializer.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def version
+    object.adapter.serializer.instance_options[:version]
+  end
+
+  def usage
+    {
+      users: {
+        total: instance_presenter.user_count,
+        activeHalfyear: instance_presenter.active_user_count_month,
```
nightpool: should pass `months: 6` for clarity

app/presenters/instance_presenter.rb
```diff
@@ -24,6 +24,10 @@ def active_user_count
     Rails.cache.fetch('active_user_count') { Redis.current.pfcount(*(0..3).map { |i| "activity:logins:#{i.weeks.ago.utc.to_date.cweek}" }) }
   end
 
+  def active_user_count_month(months: 6)
```
nightpool: defaulting to 6 makes the call site confusing, I would prefer if this took no default value

app/serializers/node_info_serializer.rb
```diff
@@ -0,0 +1,87 @@
...
+    blocks = DomainBlock.with_user_facing_limitations
+    feds = {
+      reject_media: [],
+    }
+    blocks.each do |block|
+      feds[block.severity] = [] if %w(silence suspend).include?(block.severity) && !feds.keys.include?(block.severity)
+      feds[block.severity] << block.domain
+      feds[:reject_media] << block.domain if block.reject_media
+    end
+    feds
```
nightpool: `each_with_object` is more idiomatic then creating a temporary variable and then returning it.

```ruby
DomainBlock.with_user_facing_limitations.each_with_object(Hash.new { |h, k| h[k] = [] }) do |block, federation_payload|
  federation_payload[block.severity] << block.domain if %w(silence suspend).include?(block.severity)
  federation_payload[:reject_media] << block.domain if block.reject_media
end
```

app/serializers/node_info_serializer.rb
```diff
@@ -0,0 +1,87 @@
...
+    %w(mastodon_api mastodon_api_streaming)
+  end
+
+  def federation
+    blocks = DomainBlock.with_user_facing_limitations
+    feds = {
+      reject_media: [],
+    }
+    blocks.each do |block|
+      feds[block.severity] = [] if %w(silence suspend).include?(block.severity) && !feds.keys.include?(block.severity)
```
nightpool: why not just add silence: and suspend: to the initial value? the fact that `reject_media` and severity are treated differently makes the code harder to read

nightpool: actually, we can clean this whole thing up with Hash.new

trwnh:
if nodeinfo is no good, then can the equivalent information be added to /api/v1/instance?

app/serializers/node_discovery_serializer.rb
ThibG: Oh, indeed, I haven't thought to check that. I think I'll just remove nodeinfo 2.1.

ThibG:
Thanks for your review. I think the parts covering software version, name, usage stats, whether registrations and open, and what protocols are supported are pretty well-defined and are already useful, if only for stuff like https://fediverse.network/

You're right that the `metadata` part is unspecified, and it's somewhat disappointing. But machine-readable data for domain blocks could be useful, and nodeinfo is how they are already being displayed by other software.

At least the following software already make use of it: Pleroma, PeerTube, PixelFed, Friendica, Hubzilla, Misskey, and Plume, just to name a few. (I'm not sure which of these use nodeinfo to share block infos, though. I suspect only Pleroma does that?)

nightpool:
since exposing block information progrmatically is a big change, I would like more input from other developers before I'm comfortable merging that part of the PR.

Separately, on the nodeinfo side I would like a more thorough survey of the existing nodeinfo implementations for incompatibilities and perhaps a very quick spec document (doesn't need to be more then a couple of sentences per property) for the unspecified metadata values. I'm happy to do this myself, but my bandwidth is pretty limited

ThibG:
Exposing block information programmatically is already a thing since #11298 has been merged (but the point is to now export it in a machine-readable format).

**metadata**, is, well, free-form. Of the properties we set, here are what they seem to mean and which other implementations use them:

- **nodeName:** free-form human-readable name of the service (not necessarily its domain name), used at least by Pleroma, PeerTube, PixelFed, Friendica, Hubzilla, Diaspora
- **nodeDescription:** free-form human-readable description of the service, used at least by Pleroma and PeerTube
- **nodeTerms:** actually haven't seen that one used, but the name makes it pretty obvious it's free-form human-readable text to describe the terms of service of the service
- **siteContactEmail:** actually haven't seen that one used, but seems pretty obvious
- **domain_count:** haven't seen this one in use either
- **features:** array of strings used by Pleroma to list features such as support for the Mastodon API, whether media files are proxied, etc.
- **federation:** used by Pleroma to describe MRF rules currently in use, in addition to other moderation rules such as quarantined domains, rejected domains etc.
- **invitesEnabled:** boolean to tell whether invites are enabled, used by Pleroma

Misskey uses **name** in place of **nodeName**. Plume and Misskey use **description** in place of **nodeDescription**.

PR #11632, opened 2019-08-20
Restore hashtag suggestions from local tag history

ClearlyClaire:
I'm not happy about mixing `ImmutableJS` and native lists/objects here, but it seems like the best way to make it work both for when the REST API hasn't replied yet and for results are gotten from it…

~~EDIT: also not super happy about it claiming there has been 0 uses in the week when the tag data isn't from API results~~

Example when `#CatPosting` isn't approved (but has been used locally) but `#cat` is:
Peek 21-08-2019 13-13

### DIFF
--- a/app/javascript/mastodon/reducers/compose.js
+++ b/app/javascript/mastodon/reducers/compose.js
@@ -210,7 +210,7 @@ const mergeLocalHashtagResults = (suggestions, prefix, tagHistory) => {
   prefix = prefix.toLowerCase();
   if (suggestions.length < 4) {
     const localTags = tagHistory.filter(tag => tag.toLowerCase().startsWith(prefix) && !suggestions.some(suggestion => suggestion.type === 'hashtag' && suggestion.name.toLowerCase() === tag.toLowerCase()));
-    return localTags.slice(0, 4 - suggestions.length).toJS().map(tag => ({ type: 'hashtag', name: tag, history: [] })).concat(suggestions);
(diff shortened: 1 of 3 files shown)


app/javascript/mastodon/reducers/compose.js
```diff
@@ -210,7 +210,7 @@ const mergeLocalHashtagResults = (suggestions, prefix, tagHistory) => {
   prefix = prefix.toLowerCase();
   if (suggestions.length < 4) {
     const localTags = tagHistory.filter(tag => tag.toLowerCase().startsWith(prefix) && !suggestions.some(suggestion => suggestion.type === 'hashtag' && suggestion.name.toLowerCase() === tag.toLowerCase()));
-    return localTags.slice(0, 4 - suggestions.length).toJS().map(tag => ({ type: 'hashtag', name: tag, history: [] })).concat(suggestions);
+    return suggestions.concat(slice(0, 4 - suggestions.length).toJS().map(tag => ({ type: 'hashtag', name: tag })));
```
ykzts: `slice` ?

ThibG: Sorry, messed that one up ^^'
fixed

Gargron: Hold up what is it slicing now?

ykzts (approved):

Gargron:
> ~~EDIT: also not super happy about it claiming there has been 0 uses in the week when the tag data isn't from API results~~

It should be possible to fetch the data for those tags

PR #11642, opened 2019-08-23
Fix b, i, strong tag styling in content

tribela:
Because of 'reset.css' `b`, `i`, `strong` tags are styled incorrectly.
This commit fixes that

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -824,6 +824,16 @@
     color: lighten($ui-highlight-color, 8%);
   }
 
+  b,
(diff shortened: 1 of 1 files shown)


Gargron:
Upstream Mastodon does not allow these tags in HTML sanitization...

nightpool:
@Gargron /about/more maybe?

Kjwon15:
For `about/more`, `<sup>` need to be fixed too.
It's `reset.css`'s problem.

ThibG:
> @Gargron /about/more maybe?

The changes here are for statuses though

PR #11646, opened 2019-08-23
Make API rate limits configurable (updated from #9257)

sable-starflower:
Hello, I wanted to use the changes in https://github.com/tootsuite/mastodon/pull/9257 so I updated them to work with current master, since the original PR seemed to be abandoned.

#### Related issues

- https://github.com/tootsuite/mastodon/issues/11156
- https://github.com/tootsuite/mastodon/issues/10557
- https://github.com/tootsuite/mastodon/issues/3664

Resolve #9257


.env.production.sample
```diff
@@ -77,6 +77,18 @@ SMTP_FROM_ADDRESS=notifications@example.com
 #SMTP_ENABLE_STARTTLS_AUTO=true
 #SMTP_TLS=true
 
+# Rate limits for the API
+# The period of all limits is 5 minutes
```
nightpool: this comment is incorrect

nightpool:
We do not currently recommend confusing the rate limits on deletions, they're in place to prevent systems from flooding the network accidentally. This does mean it might take a couple of days to delete all toots on very active accounts (for example, I've been extremely active on mastodon, and it might take 2 weeks to delete all of my posts). In special cases, I could see admins individually lifting the rate limits, but it's easy enough to do without this pull request—just edit the file. Or just enqueue the deletes manually from the console.

for continuous use, these rate limits are required for network stability and we don't recommend changing them, so I don't think we should make them configurable.

sable-starflower:
Okay, not wanting to make the delete rate configurable makes sense. But I think the problem (and why there have been multiple issues opened about deletions in particular) is that the current rate is lower than someone browsing through their timeline and deleting posts one by one, even at a leisurely pace. I understand why a low limit would be needed to keep bots under control, but it seems like it could safely be raised a bit to avoid affecting users. Can I ask how the current limit was picked? Were there any measurements performed or was it estimated?

I have a couple of suggestions:

1. Increase the default rate limit, and/or lower the reset timer. Being able to delete one status per 30 seconds or per 15 seconds average instead of 1 minute would greatly increase usability and decrease frustration while still preventing automated abuse.
1. In the UI, inform the user of how long until they can retry a delete. Right now when throttling occurs, there is no indication to a nontechnical user why this error occurred or what they can do to resolve it. Knowing how long they have to wait would stop users from rage clicking and having a bad UX.

I would be willing to do the work for either of these if you would accept the patch.

nightpool:
(1) is pretty simple, I'm down to fine tune those numbers pending @gargron
weighing in. (2) is definitely more involved—let's hold off on doing that
for now unless we're REALLY certain it's going to be necessary with lowered
ratelimits.

ykzts (approved):

SuperSandro2000:
bump

PR #11668, opened 2019-08-26
Wrap long text in placeholder widget

koyuawsmbrtn:
This makes longer text appear better on some screen sizes.

### DIFF
--- a/app/javascript/styles/mastodon/widgets.scss
+++ b/app/javascript/styles/mastodon/widgets.scss
@@ -554,6 +554,7 @@ $fluid-breakpoint: $maximum-width + 20px;
     text-decoration: none;
     font-weight: 500;
     color: $ui-highlight-color;
+    display: inline-block;
 
     &:hover,
     &:focus,


ThibG:
Hm, this doesn't seem to change anything at all for me?

koyuawsmbrtn:
If the placeholder widget for hashtags exceeds a specific limit the text should be wrapped instead of some characters just being misplaced. It's a bug for some translations with long word names e.g. German.

ThibG:
Alright, but I cannot really see how it can help :/

koyuawsmbrtn:
Currently if the text in the placeholder widget is being too long some letters are floating around and making it hard to read. This PR fixes that issue.

ThibG:
By reading the diff, the only thing I can see changing is the “Add new” being on its own line. I can't really see how it would affect wrapping in general.

Can you show screenshots? Sorry to ask so much…

koyuawsmbrtn:
grafik

is going to be

grafik

I think it'll look a lot nicer

ThibG:
Oh yeah, for this particular case, this is nicer.

ClearlyClaire (approved):

PR #11673, opened 2019-08-26
Add federation support for the "hide network" preference

ClearlyClaire:
Listing an URL that will error out is not helpful.
Furthermore, Pleroma uses this to figure out whether someone's followers should be displayed or not.

~~(Marking as draft because I'm working on a second part, which would be to check that field and hide the followers accordingly in API queries)~~

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -46,6 +46,8 @@
 #  silenced_at             :datetime
 #  suspended_at            :datetime
 #  trust_level             :integer
+#  hide_followers          :boolean
+#  hide_following          :boolean
 #
 
 class Account < ApplicationRecord
@@ -320,6 +322,14 @@ class Account < ApplicationRecord
     self.header = nil
 
     save!
+  end
+
+  def hides_followers?
+    hide_followers || user_hides_network?
+  end
+
+  def hides_following?
+    hide_following || user_hides_network?
   end
 
   def object_type
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -324,11 +324,11 @@ class Account < ApplicationRecord
   end
 
   def hides_followers?
-    hide_collections || user_hides_network?
+    hide_collections? || user_hides_network?
   end
 
   def hides_following?
-    hide_collections || user_hides_network?
+    hide_collections? || user_hides_network?
   end
 
   def object_type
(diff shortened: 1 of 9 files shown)


db/migrate/20190826161737_add_hide_followers_and_following_to_accounts.rb
```diff
@@ -0,0 +1,6 @@
+class AddHideFollowersAndFollowingToAccounts < ActiveRecord::Migration[5.2]
+  def change
+    add_column :accounts, :hide_followers, :boolean
+    add_column :accounts, :hide_following, :boolean
```
Gargron: I am not entirely sure what these two separate options could be useful for. I understand that other software *could* use them separately, but it's probably a waste on our side. The accounts table is big already. So how about just one column? And transfer the local user's setting into it?

ThibG: Would adding two booleans instead of one really make a difference?
Using such booleans instead of the user setting for local accounts would make sense, but it would involve more work and a migration. I will investigate this, but I don't think that's a blocker for this PR, as any migration of this kind wouldn't be easier to do when introducing those columns than later down the road.

ThibG:
Updated to use a single column. Using that column instead of the `hide_network` setting would make sense, but the migration is going to be a mess, I'm not sure it's worth it.

db/migrate/20190826161737_add_hide_followers_and_following_to_accounts.rb
ThibG: Up

app/models/account.rb
```diff
@@ -322,6 +323,14 @@ def save_with_optional_media!
     save!
   end
 
+  def hides_followers?
+    hide_collections || user_hides_network?
```
Gargron: For boolean attributes, the getter with `?` is generated automatically and is the preferred access method. It should be `hide_collections?`

app/services/activitypub/process_account_service.rb
```diff
@@ -166,26 +167,36 @@ def mismatching_origin?(url)
   end
 
   def outbox_total_items
-    collection_total_items('outbox')
+    collection_info('outbox')[0]
```
Gargron: I guess `.first` and `.last` would be better here

Gargron (changes requested):

Gargron (approved):

PR #11674, opened 2019-08-26
Add API endpoint to list followed users whose reblogs are hidden

ClearlyClaire:


ThibG:
This may be useful to know which users you're hiding boosts from, as there is currently no way to know that except by manually checking every follower.

Gargron:
Eh, that seems kind of too specific? A whole API just for a single option?

ThibG:
Well, seeing your favourites is also one API endpoint for a single option… the server must communicate those informations somehow… the alternative, I guess, would be an option to `/api/v1/following_accounts` but that is very awkward because the option would only apply to when the requested account is the account of the person logged in

Gargron:
The difference is that "favourites" is a primary function while "hide reblogs" is an option on the primary "follows" function. Perhaps it would make sense within the larger goal of making the "relationship manager" become possible through the API.

PR #11677, opened 2019-08-26
Add special alert for throttled requests

ClearlyClaire:


app/javascript/mastodon/actions/alerts.js
```diff
@@ -33,13 +35,18 @@ export function showAlert(title = messages.unexpectedTitle, message = messages.u
...
+    const { data, status, statusText, headers } = error.response;
 
     if (status === 404 || status === 410) {
       // Skip these errors as they are reflected in the UI
       return { type: ALERT_NOOP };
     }
 
+    if (status === 429 && headers['x-ratelimit-reset']) {
+      const reset_date = new Date(headers['x-ratelimit-reset']);
+      return showAlert(messages.rateLimitedTitle, [messages.rateLimitedMessage, { 'retry_time': reset_date }]);
```
Gargron: The array thing seems unnecessary. Wouldn't it be just as easy to have a 3rd argument on the function and check if that argument is not undefined?

Gargron (changes requested):

app/javascript/mastodon/actions/alerts.js
ThibG: Pushed a commit to address that, not sure how much better that is, though

app/javascript/mastodon/actions/alerts.js
```diff
@@ -23,23 +25,29 @@ export function clearAlert() {
   };
 };
 
-export function showAlert(title = messages.unexpectedTitle, message = messages.unexpectedMessage) {
+export function showAlert(title = messages.unexpectedTitle, message = messages.unexpectedMessage, message_values = undefined) {
```
Gargron: There is really no need for a default value if the default value is `undefined`, JavaScript doesn't care how many arguments you pass into a function, if you pass fewer the other ones are undefined.

Gargron (approved):

PR #11693, opened 2019-08-29
Revert bcrypt to 3.1.12, again

ClearlyClaire:
Same reasons as in #11555


Gargron:
Looks like this time it's from a Devise update #11667

ThibG:
Still, I think we can apply this PR?
bcrypt 3.1.13 doesn't bring anything meaningful to us

Gargron (approved):

nanaya:
FWIW, bcrypt 3.1.12 (and earlier iirc) isn't usable on recent FreeBSD.

PR #11707, opened 2019-08-30
Update NodeJS source for Docker Image

shleeable:
Node v12 was supported recently in PR #11706 -- and Node v8 was in maintenance LTS, so worth upgrading.

This requires additional testing before it should be pulled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu:18.04 as build-dep
 SHELL ["bash", "-c"]
 
 # Install Node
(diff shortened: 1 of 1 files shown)


Shleeble:
This requires considerable testing before it should be pulled.. I will consider splitting these into three parts.

Jemalloc/Ruby should be safe... NodeJS is uncertain.

Shleeble:
First Shake looks fine... Waiting for other approvals.

Gargron (approved):

PR #11712, opened 2019-09-01
Change trending hashtags to not disappear instantly after midnight

Gargron:
When a hashtag is used, update its counters, and mark it down as used on that day for easy retrieval. Do *not* calculate the score right then, instead, add a regular scheduler. The scheduler goes over all used hashtags, and hashtags that were previously in the trends, and recalculates their scores. If a score peaks, record the time at which it peaked. Use the peak score, but decay it from the point at which it peaked with a half-life of 6 hours, for the sorting in the trends.

This removes the need to fully reset the trends at midnight, while ensuring that newer trends take precedence over older ones.

### DIFF
--- a/app/models/trending_tags.rb
+++ b/app/models/trending_tags.rb
@@ -7,6 +7,8 @@ class TrendingTags
   THRESHOLD            = 5
   LIMIT                = 10
   REVIEW_THRESHOLD     = 3
+  MAX_SCORE_COOLDOWN   = 3.days.freeze
+  MAX_SCORE_HALFLIFE   = 6.hours.freeze
 
   class << self
     include Redisable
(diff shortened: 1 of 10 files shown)


app/models/trending_tags.rb
```diff
@@ -16,14 +18,75 @@ def record_use!(tag, account, at_time = Time.now.utc)
...
+      end
+
+      users_for_review = User.staff.includes(:account).to_a.select(&:allows_trending_tag_emails?)
+
+      # Second pass to notify about previously unreviewed trends
+
+      tags.each do |tag|
+        current_rank              = redis.zrevrank(KEY, tag.id)
+        needs_review_notification = tag.requires_review? && !tag.requested_review?
+        rank_passes_threshold     = current_rank.present? && current_rank <= REVIEW_THRESHOLD
```
ThibG: Doesn't that mean that no notifications will be generated for tags that would be trending if there are too many spammy hashtag trends above? This isn't new though

app/models/trending_tags.rb
```diff
@@ -16,14 +18,75 @@ def record_use!(tag, account, at_time = Time.now.utc)
...
+        decaying_score = max_score * (0.5**((at_time.to_f - max_time.to_f) / MAX_SCORE_HALFLIFE.to_f))
+
+        if decaying_score.zero?
+          redis.zrem(KEY, tag.id)
+        else
+          redis.zadd(KEY, decaying_score, tag.id)
+        end
+      end
+
+      users_for_review = User.staff.includes(:account).to_a.select(&:allows_trending_tag_emails?)
```
ThibG: The name is slightly misleading, I was confused this was about reviewing users, and not users that should receive a notification about trending hashtags.

config/sidekiq.yml
```diff
@@ -9,6 +9,9 @@
   scheduled_statuses_scheduler:
     every: '5m'
     class: Scheduler::ScheduledStatusesScheduler
+  trending_tags_scheduler:
+    every: '5m'
+    class: Scheduler::TrendingTagsScheduler
```
ThibG: I'm not sure how much expensive the whole thing would be, I'm afraid that running it every 5 minute might be too expensive.

ClearlyClaire (approved):
Seems good overall, but I have a few reservations (see inline comments).

app/models/trending_tags.rb
Gargron: If a hashtag is disallowed from trending it won't even make it to this point. So this is all happening for hashtags that are either already allowed to trend, or hashtags that might be allowed to trend, so they can't be discounted.

config/sidekiq.yml
Gargron: It takes milliseconds to run. While the usage buckets are daily, we *do* have an interest in picking up trends as close to real-time as possible, because that's when they're at their most interesting.

app/models/trending_tags.rb
ThibG: I mean, if the 3 top trending tags are unreviewed and spammy, this won't notify of legitimate ones.
But I approached that with the belief that the task would run once per day, if it runs once per 15 minutes, this is probably ok?

config/sidekiq.yml
ThibG: ok, fine then

PR #11715, opened 2019-09-01
Add "not available" label on unknown media attachments

Gargron:
Since unhiding "sensitive media" doesn't actually unhide anything, just always show a "not available" message. This should bring some clarity when e.g. a domain is reject-media-blocked. It would be better to say as much in more words but I don't see a way to add an accessible hint to that interface.


ClearlyClaire (approved):
This is an improvement overall, but I find it pretty weird to list unsupported attachments in the `MediaGallery` layout in the first place?

PR #11717, opened 2019-09-01
Fix display of long poll options

ClearlyClaire:
### Web UI

image

### Public pages

image

### Web UI

image

### Public pages

image

### DIFF
--- a/app/javascript/styles/mastodon/polls.scss
+++ b/app/javascript/styles/mastodon/polls.scss
@@ -5,7 +5,6 @@
   li {
     margin-bottom: 10px;
     position: relative;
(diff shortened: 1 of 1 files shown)


ykzts (approved):

ThibG:
The reason why this doesn't get merged is that Gargron thinks the multi-line options do not look as nice as single-line options and is worried that supporting longer options would encourage people to actually use longer options.

Gargron:
> The reason why this doesn't get merged is that Gargron thinks the multi-line options do not look as nice as single-line options and is worried that supporting longer options would encourage people to actually use longer options.

#11837 :thinking:

PR #11749, opened 2019-09-03
Fix admin interface showing superfluous reject media/reports on suspended blocks

ClearlyClaire:


Gargron (approved):

Shleeble:
@Gargron @ThibG considering the existing domain blocks show these additional options on suspend mode... is there any logic bugs I've introduced on the blocks I have added now?

or should I delete and readd the domains to just show suspend?

PR #11752, opened 2019-09-03
noindex=true

todoneunl:
Following the discourse in the fediverse about search indexing, I think a lot of people would welcome it to change noindex to true by default. Of course it should be easier to change this to noindex=false in the admin interface

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -28,7 +28,7 @@ defaults: &defaults
   reduce_motion: false
   show_application: true
   system_font_ui: false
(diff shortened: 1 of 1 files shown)


Gargron:
I think this would be an unexpected default. Every other social media profile is findable through search engines. Specific communities on Mastodon might have a different expectation, and changing the instance-wide default *should* be easier for admins, but this repository serves more communities than those ones and I think overall this change would be considered an anti-feature.

PR #11755, opened 2019-09-04
Fix wrong sum of usage in hashtag component in web UI

Gargron:
1 + 1 = 11


ykzts (approved):
oops... I missed this.

PR #11757, opened 2019-09-04
Attempt to concurrently connect to remote IP addresses

ClearlyClaire:
Currently, Mastodon tries to connect one IP at a time, which means it will waste time if the first IP address it tries is unreachable.

Furthermore, it should address the following scenario:
- instance A has two IP addresses
- the first one is unreachable to instance B, the second is reachable to instance B
- A issues a request to B, who needs to fetch a resource at A before replying
- B timeouts on the first address, and thus can only reply after the timeout duration
- which means A timeouts before allowing B to reply

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -191,6 +191,9 @@ class Request
           end
         end
 
+        socks = []
+        addr_by_socket = {}
+
         addresses.each do |address|
           begin
             check_private_address(address)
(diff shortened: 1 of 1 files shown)


app/lib/request.rb
```diff
@@ -202,28 +205,46 @@ def open(host, *args)
...
-                rescue
-                  sock.close
-                  raise
-                end
-              else
-                sock.close
-                raise HTTP::TimeoutError, "Connect timed out after #{Request::TIMEOUT[:connect]} seconds"
-              end
+              socks << sock
+              addr_by_socket[sock] = sockaddr
```
ThibG: I am not super happy about having a socket as a hash key, but I'm not sure how else to make the sockaddr corresponding to the socket when doing the `IO.select` bit. Should be fine in practice.

Gargron (approved):

Issue #6882, opened 2018-03-23
Colour profiles being stripped from JPEGs?


HywelPhillips:
[
I'm very sorry if I'm doing this wrong, I am a newbie running a hosted instance.

It seems like coloursync profiles are stripped from JPEGs when posting, is this correct?

I posted the same pic to birdsite and via Mastodon, and the Mastodon one had its profile stripped away when I saved the file. This seems to be causing colour shifts for non-sRGB JPEGs (eg stuff in ProphotoRGB). 
]

* * * *

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

rhaamo:
IIRC all exif tags are stripped after upload, the coloursync profiles may be too.

cangyuyao:
There should be a better solution to the colour profiles. Every photo uploaded from iPhone 7 or above distorted in colour, which is really annoying.

StudioPetrikas:
I would like to add my 2 cents about this issue. 
I have listed some points as to why stripping color profiles is bad practice in mareklipka's pull request here:

https://github.com/mastodon/mastodon/pull/11758#issuecomment-1310671267

PR #11758, opened 2019-09-04
Prevent stripping color profiles from JPEGs

mareklipka:
ref #6882

Use `+profile exif` instead of `-strip` to keep color profiles in image files.

### DIFF
--- a/app/models/concerns/account_avatar.rb
+++ b/app/models/concerns/account_avatar.rb
@@ -18,7 +18,7 @@ module AccountAvatar
 
   included do
     # Avatar upload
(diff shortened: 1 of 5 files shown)


Gargron:
Can anyone confirm if this still strips all the privacy-related metadata?

StudioPetrikas:
Color profiles continue to get stripped from JPGs. I would like this issue to get reopened. 

Especially in these time where many monitors come with AdobeRGB profiles as default, and ALL new macOS devices are at least Display P3 or 'wider'. 

- Every image has to have a color profile attached. Untagged images are bad practice. 
- Some of the web browsers treat untagged images as monitor's profile, which, in turn, return an unintended image. 
- Untagged images are also bad for archiving/journaling purposes, as there is no instruction - just an assumption - on how to read it.

Please reconsider this as a big problem that will only get bigger with time.
Thank you in advance.

PR #11762, opened 2019-09-04
Add timeline read markers API

### LINKED ISSUE #4093: Endpoint for last read notification

### LINKED ISSUE #7418: Endpoint for last read toot
(Somewhat similar idea to #4093)

It would be nice to have a synced timeline across clients (web or not) connected to an instance. This functionality suggestion is directly inspired by the (discontinued) [Tweet Marker](https://github.com/manton/tweetmarker) 3rd party service for Twitter.

It would allow clients to set & get the last viewed toot on the instance itself, thus allowing them to know where to continue from when displaying the timeline. When using multiple clients it would make switching between them easier since each of the clients could show the same point in the instance's timeline.

I cannot give implementation suggestions but maybe the [Tweet Marker API description](https://github.com/manton/tweetmarker/blob/master/documentation/v2.md) could be a helpful starting point.

* * * *

Gargron:
Fix #4093, fix #7418, close #6756

|Method|Description|
|-----|----|
|`GET /api/v1/markers`|Get markers specified by `timeline` array param. Response is a map with timeline names as keys|
|`POST /api/v1/markers`|Create new marker(s) with `last_read_id` by submitting a map with timeline names as keys. The keys can be `home` or `notifications`|

JSON returned includes `last_read_id`, `updated_at` and `version` (which increments on every update).

### DIFF
--- a/app/javascript/mastodon/actions/markers.js
+++ b/app/javascript/mastodon/actions/markers.js
@@ -1,0 +1,30 @@
+export const submitMarkers = () => (dispatch, getState) => {
+  const accessToken = getState().getIn(['meta', 'access_token'], '');
+  const params      = {};
+
+  const lastHomeId         = getState().getIn(['timelines', 'home', 'items', 0]);
+  const lastNotificationId = getState().getIn(['notifications', 'items', 0, 'id']);
+
+  if (lastHomeId) {
+    params.home = {
+      last_read_id: lastHomeId,
+    };
+  }
+
+  if (lastNotificationId) {
+    params.notifications = {
+      last_read_id: lastNotificationId,
+    };
+  }
+
+  if (Object.keys(params).length === 0) {
+    return;
+  }
+
+  const client = new XMLHttpRequest();
+
+  client.open('POST', '/api/v1/markers', false);
+  client.setRequestHeader('Content-Type', 'application/json');
+  client.setRequestHeader('Authorization', `Bearer ${accessToken}`);
+  client.send(JSON.stringify(params));
+};
--- a/app/controllers/api/v1/markers_controller.rb
+++ b/app/controllers/api/v1/markers_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class Api::V1::MarkersController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:statuses' }, only: [:index]
+  before_action -> { doorkeeper_authorize! :write, :'write:statuses' }, except: [:index]
+
+  before_action :require_user!
+
+  def index
+    @markers = current_user.markers.where(timeline: Array(params[:timeline])).each_with_object({}) { |marker, h| h[marker.timeline] = marker }
+    render json: serialize_map(@markers)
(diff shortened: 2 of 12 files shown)


app/javascript/mastodon/actions/markers.js
```diff
@@ -0,0 +1,41 @@
...
+    params.notifications = {
+      last_read_id: lastNotificationId,
+    };
+  }
+
+  // At least in Firefox, the request does not send unless
+  // console.log is used like this
+
+  const beacon = sendBeacon(accessToken, params);
+  console.log('Saving timeline markers...', beacon);
```
ykzts:
```suggestion
  console.log('Saving timeline markers...', beacon); // eslint-disable-line no-console
```

app/javascript/mastodon/actions/markers.js
```diff
@@ -0,0 +1,41 @@
+import api from '../api';
```
ykzts: unused?

ykzts (changes requested):

ykzts (approved):

nolanlawson:
I have a few questions about this implementation:

1. It seems the Mastodon webapp will submit the markers in the unload event. So if a user keeps their tab open in their browser and never closes it, another client will never know what the last read position is? It seems like it may be better to do this lazily (e.g. every few minutes), as well as during the unload event and when a page goes into a background tab.
2. This is only for the home/notifications timelines right now, if I understand correctly?
3. Is the assumption that, on first load, a client will immediately scroll to the saved position in the timeline if a marker is found? So if a user has not checked Mastodon in several days and there are 100s of statuses, should the app load all 100+ of them and then scroll to the position? Or give up and reset the position? Or lazily show the new content?

/cc @charlag

charlag:
I agree with @nolanlawson on the first one.
I think when you open it on another client you just load:
 1. The latest ones
 2. Statuses with max_id = marker and 
and put a loadable gap between them (I think we do something like that already).
cc @connyduck

Gargron:
> It seems like it may be better to do this lazily (e.g. every few minutes), as well as during the unload event and when a page goes into a background tab.

Well, what if you have multiple clients open? And they all try to update the position at the same time? If they all use the streaming API I suppose they would arrive at similar or identical results but it seems wasteful to just keep updating it back and forth between them all. I'm not sure what the point would be of saving and loading a marker while a client is already actively open, the other app would then load the top of the timeline anyway..?

charlag:
It could be done in the way that we send it to the server and server writes it only when thr client stops communicating but that's complicated I guess.
I dot think anyone will load marker when the app is already open, only on startup. If two clients open, the last used one is likely to win (which makes sense to me).

PR #11768, opened 2019-09-06
Use prebuilt node.js on Docker image

ykzts:
Uses too much CPU resources and memory when building Node.js from source code. This may cause the Docker image build to fail in an environment with insufficient resources.

Since Node.js has an official prebuilt binary, change it to use it.


Shleeble:
works for me.. Hopefully we migrate to flodgatt sooner than later and we can drop the nodejs.

Gargron (approved):

PR #11770, opened 2019-09-06
Add healthcheck endpoint for web

ykzts:
Add a health check to check DB communication. The health check is connected with an internal request, so it does not force connection with HTTPS.

[This gem also uses GitLab](https://gitlab.com/gitlab-org/gitlab-ce/blob/v12.2.4/config/initializers/health_check.rb).


nightpool:
there's some kind of logical conflict between this PR and the one you opened right before it: https://github.com/tootsuite/mastodon/pull/11769/files

ykzts:
Yes, I will resolve the conflict based on what was merged earlier.

Gargron (approved):

Shleeble:
Excellent. I'll use that for my pingdom style external monitoring :)

PR #11775, opened 2019-09-08
Add retry for failed media downloads and `tootctl media refresh`

Gargron:
We download media attachments synchronously when processing a new remote status, so that we have all the information we need *before* the status is distributed to end-users through the streaming API &co. However, sometimes the download fails due to high load on the origin server, or Wasabi being an awfully unstable service. In those cases, we want to retry downloading specifically that attachment at a later time.

With this, a separate Sidekiq job will be queued with a random delay between 30 seconds and 10 minutes that attempts to re-download that attachment at most 3 times with exponential backoff.

Separately, there's a fix for not processing more than 4 media attachments for a status, since they will never be used within Mastodon (close #9325).

Furthermore, a new tootctl command, `tootctl media refresh` allows to re-download attachments for a particular status, a particular account, or a particular domain.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -189,22 +189,25 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
     media_attachments = []
 
     as_array(@object['attachment']).each do |attachment|
-      next if attachment['url'].blank?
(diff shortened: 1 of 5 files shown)


ykzts (approved):

Gargron:
Because of changes in #11776, this should be refactored if #11776 is merged.

PR #11776, opened 2019-09-08
Change tootctl to use inline parallelization instead of Sidekiq

Gargron:
- Add progress bar instead of printing `.`
- Skip over any individual errors
- Add `-c`/`--concurrency` flag
- Remove `--background` flag

Commands affected by change (unless specified otherwise, progress bar and parallelization):

- `tootctl accounts rotate` (not parallelized)
- `tootctl accounts cull`
- `tootctl accounts refresh`
- `tootctl accounts follow`
- `tootctl accounts unfollow`
- `tootctl accounts reset-relationships` (not parallelized)
- `tootctl cache recount`
- `tootctl domains purge`
- `tootctl domains crawl`
- `tootctl feeds build`
- `tootctl media remove`
- `tootctl preview_cards remove`

Examples:

`tootctl media remove --days=60 -c 20` will remove media older than 60 days, 20 at a time


ThibG:
Won't removing the workers cause issues if someone updates their instance while such tasks are queued?

Gargron:
> Won't removing the workers cause issues if someone updates their instance while such tasks are queued?

Since the workers are *only* used by the tootctl commands such timing is unlikely. I would prefer not to keep dead code around!

lib/mastodon/accounts_cli.rb
```diff
@@ -304,76 +282,53 @@ def refresh(username = nil)
...
-    LONG_DESC
-    def unfollow(acct)
-      target_account = Account.find_remote(*acct.split('@'))
-      processed      = 0
-      failed         = 0
+    option :concurrency, type: :numeric, default: 5, aliases: [:c]
+    option :verbose, type: :boolean, aliases: [:v]
+    desc 'unfollow USERNAME', 'Make all local accounts unfollow account specified by USERNAME'
+    def unfollow(username)
+      target_account = Account.find_local(username)
```
ThibG: You're restricting this to local accounts while it worked for remote accounts as well.

lib/mastodon/media_cli.rb
```diff
@@ -7,64 +7,41 @@
...
-        MediaAttachment.where.not(remote_url: '').where.not(file_file_name: nil).where('created_at < ?', time_ago).reorder(nil).find_in_batches do |media_attachments|
-          media_attachments.each do |m|
-            size += m.file_file_size || 0
-            Maintenance::UncacheMediaWorker.new.perform(m) unless options[:dry_run]
-            options[:verbose] ? say(m.id) : say('.', :green, false)
-            processed += 1
-          end
-        end
-      end
+      processed, aggregate = parallelize_with_progress(MediaAttachment.cached.where('created_at < ?', time_ago)) do |media_attachment|
```
ThibG: the `where.not(remote_url: '')` is lost in this query. Not sure it was useful though?

lib/mastodon/cli_helper.rb
```diff
@@ -7,3 +7,48 @@
 ActiveJob::Base.logger       = dev_null
 HttpLog.configuration.logger = dev_null
 Paperclip.options[:log]      = false
+
+module Mastodon
+  module CLIHelper
+    def create_progress_bar(total = nil)
+      ProgressBar.create(total: total, format: '%c/%u |%b%i| %e')
```
ThibG: I think we better add an explicit dependency on ruby-progressbar in the `Gemfile`?

lib/mastodon/cli_helper.rb
```diff
@@ -7,3 +7,48 @@
...
+          ActiveRecord::Base.connection_pool.with_connection do
+            begin
+              progress.log("Processing #{item.id}") if options[:verbose]
+
+              result = yield(item)
+              aggregate.increment(result) if result.is_a?(Integer)
+            rescue => e
+              progress.log pastel.red("Error processing #{item.id}: #{e}")
+            ensure
+              progress.increment
```
ThibG: Is `ProgressBar` thread-safe?

lib/mastodon/cli_helper.rb
```diff
@@ -7,3 +7,48 @@
...
+
+      progress  = create_progress_bar(scope.count)
+      pool      = Concurrent::FixedThreadPool.new(options[:concurrency])
+      futures   = []
+      aggregate = Concurrent::AtomicFixnum.new(0)
+
+      scope.reorder(nil).find_each do |item|
+        progress.total = futures.size + 1 if progress.total < futures.size + 1
+
+        futures << Concurrent::Future.execute(executor: pool) do
```
ThibG: From what I understand, even though `find_each` work in batches, this code will go through *all* the items in the query and put them in the queue, which might not be what we want, wrt. RAM usage.

Would it be possible to reap finished jobs as soon as possible and block whenever the pool is busy?

ClearlyClaire (changes requested):

lib/mastodon/accounts_cli.rb
Gargron: To make it consistent with the `follow` command. There is no reason to use these on remote accounts. They are intended for e.g. changing default followed account.

lib/mastodon/cli_helper.rb
Gargron: The futures start processing as soon as they are created, the map over values at the end is just to block execution until they all finish. What is stored are the futures and the return values, which are numbers in our case, not the ActiveRecord records, so it should help with memory usage.

lib/mastodon/accounts_cli.rb
```diff
@@ -215,63 +221,33 @@ def backup(username)
 
       Accounts that have had confirmed activity within the last week
       are excluded from the checks.
-
-      Domains that are unreachable are not checked.
-
-      With the --dry-run option, no deletes will actually be carried
-      out.
```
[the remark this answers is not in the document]
Gargron: There are now a few options that appear in a lot of commands. Manually maintaining copies of help text between them all doesn't seem good. The option appears automatically when executing `--help`, that should suffice.

lib/mastodon/cli_helper.rb
Gargron: I cannot find any information on this including any complaints on their bug tracker so I assume it's not going to be a problem.

lib/mastodon/accounts_cli.rb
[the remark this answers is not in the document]
ThibG: fair enough

lib/mastodon/accounts_cli.rb
ThibG: I am not entirely convinced :man_shrugging:

Gargron: Okay, restored.

lib/mastodon/cli_helper.rb
ThibG: I'm still worried about it being too much if the tasks are slow. I guess it could be interesting to see how it runs for pruning old remote media on m.s

ClearlyClaire (approved):

PR #11778, opened 2019-09-08
Add featured tags API

noellabo:
Enable to set and operate featured tags from API such as client application.

### DIFF
--- a/app/controllers/api/v1/featured_tags/suggestions_controller.rb
+++ b/app/controllers/api/v1/featured_tags/suggestions_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class Api::V1::FeaturedTags::SuggestionsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:featured_tags' }, only: [:index]
+
+  before_action :require_user!
+
+  respond_to :json
+
+  def index
+    most_used_tags = Tag.most_used(current_account).where.not(id: current_account.featured_tags).limit(10)
+    render json: most_used_tags, each_serializer: REST::TagSerializer
+  end
+end
--- a/app/controllers/api/v1/featured_tags/suggestions_controller.rb
+++ b/app/controllers/api/v1/featured_tags/suggestions_controller.rb
@@ -1,14 +1,20 @@
 # frozen_string_literal: true
 
 class Api::V1::FeaturedTags::SuggestionsController < Api::BaseController
-  before_action -> { doorkeeper_authorize! :read, :'read:featured_tags' }, only: [:index]
+  before_action -> { doorkeeper_authorize! :read, :'read:accounts' }, only: :index
 
   before_action :require_user!
+  before_action :set_most_used_tags, only: :index
 
   respond_to :json
 
   def index
-    most_used_tags = Tag.most_used(current_account).where.not(id: current_account.featured_tags).limit(10)
-    render json: most_used_tags, each_serializer: REST::TagSerializer
+    render json: @most_used_tags, each_serializer: REST::TagSerializer
+  end
+
+  private
+
+  def set_most_used_tags
+    @most_used_tags = Tag.most_used(current_account).where.not(id: current_account.featured_tags).limit(10)
   end
 end
--- a/app/controllers/api/v1/featured_tags_controller.rb
+++ b/app/controllers/api/v1/featured_tags_controller.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class Api::V1::FeaturedTagsController < Api::BaseController
-  before_action -> { doorkeeper_authorize! :read, :'read:featured_tags' }, only: [:index, :show]
-  before_action -> { doorkeeper_authorize! :write, :'write:featured_tags' }, except: [:index, :show]
+  before_action -> { doorkeeper_authorize! :read, :'read:accounts' }, only: :index
+  before_action -> { doorkeeper_authorize! :write, :'write:accounts' }, except: :index
 
   before_action :require_user!
   before_action :set_featured_tags, only: :index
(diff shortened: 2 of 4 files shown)


app/controllers/api/v1/featured_tags/suggestions_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Api::V1::FeaturedTags::SuggestionsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:featured_tags' }, only: [:index]
```
Gargron: Since adding new scopes for existing apps is a bit cumbersome, I wonder if maybe this should fall under `read:accounts` because it's part of an account "profile"

app/controllers/api/v1/featured_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class Api::V1::FeaturedTagsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :read, :'read:featured_tags' }, only: [:index, :show]
+  before_action -> { doorkeeper_authorize! :write, :'write:featured_tags' }, except: [:index, :show]
```
Gargron: Same thing here, `read:accounts`, `write:accounts`

app/controllers/api/v1/featured_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def show
+    render json: @featured_tag, serializer: REST::FeaturedTagSerializer
+  end
+
+  def create
+    @featured_tag = current_account.featured_tags.new(featured_tag_params)
+    @featured_tag.reset_data
+    @featured_tag.save
```
Gargron: This should use `save!` so it raises exceptions on failure.

app/controllers/api/v1/featured_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  private
+
+  def set_featured_tag
+    @featured_tag = current_account.featured_tags.find(params[:id])
+  end
+
+  def set_featured_tags
+    @featured_tags = current_account.featured_tags.order(statuses_count: :desc).reject(&:new_record?)
```
Gargron: I don't think the `reject` is necessary? Nothing I see would add unsaved records here

config/routes.rb
```diff
@@ -380,6 +380,11 @@
         resource :accounts, only: [:show, :create, :destroy], controller: 'lists/accounts'
       end
 
+      namespace :featured_tags do
+        get :suggestions, to: 'suggestions#index'
+      end
+      resources :featured_tags, only: [:index, :create, :show, :update, :destroy]
```
Gargron: Prefer newline after `end`

app/controllers/api/v1/featured_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def create
+    @featured_tag = current_account.featured_tags.new(featured_tag_params)
+    @featured_tag.reset_data
+    @featured_tag.save
+    render json: @featured_tag, serializer: REST::FeaturedTagSerializer
+  end
+
+  def update
```
Gargron: What is there to update? I think changing a hashtag should require deleting and creating a new one.

app/controllers/api/v1/featured_tags_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  before_action :require_user!
+  before_action :set_featured_tags, only: :index
+  before_action :set_featured_tag, except: [:index, :create]
+
+  def index
+    render json: @featured_tags, each_serializer: REST::FeaturedTagSerializer
+  end
+
+  def show
```
Gargron: I think `show` is probably not going to be used by apps.

app/controllers/api/v1/featured_tags/suggestions_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  before_action :require_user!
+
+  respond_to :json
+
+  def index
+    most_used_tags = Tag.most_used(current_account).where.not(id: current_account.featured_tags).limit(10)
```
Gargron: For code style, move this assignment to a `before_action`

Gargron (changes requested):
:+1:

app/controllers/api/v1/featured_tags/suggestions_controller.rb
noellabo: Change to `read:accounts`

app/controllers/api/v1/featured_tags_controller.rb
noellabo: Change to `read:accounts`, `write:accounts`

app/controllers/api/v1/featured_tags_controller.rb
noellabo: I see!

app/controllers/api/v1/featured_tags_controller.rb
noellabo: This was not necessary. Remove.

config/routes.rb
noellabo: Correct the code style.

app/controllers/api/v1/featured_tags_controller.rb
noellabo: There is certainly no use for it. Remove.

app/controllers/api/v1/featured_tags_controller.rb
noellabo: This is also true.

app/controllers/api/v1/featured_tags/suggestions_controller.rb
noellabo: Correct the code style.

noellabo:
Thank you for the review. Fixed them.

Gargron (approved):

PR #11793, opened 2019-09-09
Add batch actions and categories to admin UI for custom emojis

### LINKED ISSUE #11402: Admin UI for categorizing custom emojis

Gargron:
image

Fix #11402


ykzts:
`./spec/controllers/admin/custom_emojis_controller_spec.rb` needs to be changed.

app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -2,19 +2,22 @@
 
 module Admin
   class CustomEmojisController < BaseController
-    before_action :set_custom_emoji, except: [:index, :new, :create]
-    before_action :set_filter_params
-
     include ObfuscateFilename
+
+    before_action :set_custom_emoji, except: [:index, :new, :create, :batch]
```
ThibG: Is there an action for which it is used? If so, wouldn't `only` be more fitting than `except`?

app/views/admin/custom_emojis/index.html.haml
```diff
@@ -31,18 +33,54 @@
...
+        .fields-row
+          .fields-group.fields-row__column.fields-row__column-6
+            .input.select.optional
+              .label_input
+                = f.select :category_id, options_from_collection_for_select(CustomEmojiCategory.all, 'id', 'name'), prompt: t('admin.custom_emojis.assign_category'), class: 'select optional', 'aria-label': t('admin.custom_emojis.assign_category')
+
+          .fields-group.fields-row__column.fields-row__column-6
+            .input.string.optional
+              .label_input
+                = f.text_field :category_name, class: 'string optional', placeholder: t('admin.custom_emojis.create_new_category'), 'aria-label': t('admin.custom_emojis.create_new_category')
```
ThibG: I think those would be better as a single field with autocompletion… that can be left for later though.

app/controllers/admin/custom_emojis_controller.rb
```diff
@@ -31,57 +34,13 @@ def create
...
-
-    def disable
-      authorize @custom_emoji, :disable?
-      @custom_emoji.update!(disabled: true)
-      log_action :disable, @custom_emoji
-      flash[:notice] = I18n.t('admin.custom_emojis.disabled_msg')
-      redirect_to admin_custom_emojis_path(page: params[:page], **@filter_params)
+    def batch
+      @form = Form::CustomEmojiBatch.new(form_custom_emoji_batch_params.merge(current_account: current_account, action: action_from_button))
+      @form.save
```
ThibG: Shouldn't we have a flash message to tell the changes were processed?

spec/controllers/admin/custom_emojis_controller_spec.rb
```diff
@@ -52,64 +52,4 @@
...
-  describe 'POST #disable' do
-    let(:custom_emoji) { Fabricate(:custom_emoji, shortcode: 'test', disabled: false) }
-
-    before { post :disable, params: { id: custom_emoji.id } }
-
-    it 'enables custom emoji' do
-      expect(response).to redirect_to admin_custom_emojis_path
-      expect(custom_emoji.reload).to have_attributes(disabled: true)
-    end
-  end
```
ThibG: Specs that cover the same use case with the new actions should probably be written.

ClearlyClaire (approved):
LGTM overall, but see inline comments

PR #11798, opened 2019-09-09
Fix display of instance actor in reports UI

Gargron:


ThibG:
In which context does it show up?

Gargron:
Automatic reports

ThibG:
oh, right

ClearlyClaire (approved):

PR #11803, opened 2019-09-10
Fix duplicate HTML ids in /about

### LINKED ISSUE #11329: Duplicate HTML element IDs when navigating to the homepage
When navigating to the homepage of any Mastodon instance when logged-out, i.e. https://mastodon.social, there are warnings about duplicate IDs on HTML elements:

image
<sub>*User registration and login page on [mastodon.xyz](https://mastodon.xyz)*</sub>
These are duplicate IDs on fields with IDs `#new_user`, `#user_email` and `#user_password`.

As per the [HTML specification, section 3.2.6](https://html.spec.whatwg.org/multipage/dom.html#global-attributes):

> When specified on HTML elements, the `id` attribute value must be unique amongst all the IDs in the element's tree and must contain at least one character. The value must not contain any ASCII whitespace.

This makes the page fail [HTML validation](https://validator.w3.org/nu/?doc=https%3A%2F%2Fmastodon.social%2Fabout), and may give another minor problems when using the Web interface.

---
* [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
* [x] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) (2.9.x), and not on the `master` branch. (If you're a user, don't worry about this.)

ClearlyClaire:
Fixes #11329

### DIFF
--- a/app/views/about/_login.html.haml
+++ b/app/views/about/_login.html.haml
@@ -1,4 +1,4 @@
-= simple_form_for(new_user, url: user_session_path) do |f|
+= simple_form_for(new_user, url: user_session_path, namespace: 'login') do |f|
   .fields-group
     - if use_seamless_external_login?
       = f.input :email, placeholder: t('simple_form.labels.defaults.username_or_email'), input_html: { 'aria-label' => t('simple_form.labels.defaults.username_or_email') }, hint: false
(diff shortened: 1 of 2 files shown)


ykzts (approved):
good catch!

PR #11804, opened 2019-09-10
Add admin setting to opt out of search engine indexing by default

### LINKED ISSUE #11750: Instance-wide search engine opt-out
### Pitch

It’d be useful if instance admins could flip a toggle to opt the entire instance out of search engine indexing through robots.txt.

### Motivation

It’s already for individual users to do this through their settings. Often, an instance admin can change their nginx config to use a custom robots.txt, but many don’t have the technical expertise or don’t have shell access (e.g. of running their instance through a third-party provider).

Now that we’re seeing fediverse search engines (https://search.social/), some instances may want to opt out. There’s currently no way for an entire instance hosted on e.g. masto.host to opt out — instead, every single user has to do so.

ghost:
fixes #11750.

this is my first time working with both ruby and rails so i'm not quite sure where to put a test for this. i'd appreciate it if someone could point me in the right direction for that :)

testing the view for the `noindex` meta tag might be nice, but as far as i can tell things aren't really set up for that kind of integration testing?

### DIFF
--- a/app/lib/settings/scoped_settings.rb
+++ b/app/lib/settings/scoped_settings.rb
@@ -4,6 +4,7 @@ module Settings
   class ScopedSettings
     DEFAULTING_TO_UNSCOPED = %w(
       theme
+      noindex
     ).freeze
 
     def initialize(object)
--- a/app/models/form/admin_settings.rb
+++ b/app/models/form/admin_settings.rb
@@ -32,6 +32,7 @@ class Form::AdminSettings
     trends
     show_domain_blocks
     show_domain_blocks_rationale
+    noindex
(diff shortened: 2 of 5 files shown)


Gargron (approved):

Shleeble:
Why is opt-in the default? Shouldn't we opt out?

Gargron:
No, the default expectation for social media profiles is that they are findable. If I type in "eugen rochko linkedin" into Google I expect to see my profile rather than go and enable it separately in the innards of LinkedIn's settings area. If you think that that is not the expectation for your own community, that is why you now have this admin setting to control it.

PR #11806, opened 2019-09-11
Change spam check to apply to local accounts and add a threshold

Gargron:
Instead of detecting spam on first duplicate message, add a threshold of 5 such messages to reduce false positives, since the check will now affect local accounts as well.


ykzts (approved):

ClearlyClaire (commented):
Seems good (and actually better to catch bots opening accounts on legitimate Mastodon instances) overall, although I'm still worried about false positives given that the automatic action is quite drastic.

I'll run that on my instance for a while to get a sense of it, although I'm a single-user instance, so…

PR #11813, opened 2019-09-12
Add missing locale file for ga and add rake task to check for it

Gargron:


lib/tasks/repo.rake
```diff
@@ -76,4 +76,19 @@ namespace :repo do
...
   end
+
+  task check_locales_files: :environment do
+    pastel = Pastel.new
+
+    missing_yaml_files = I18n.available_locales.reject { |locale| File.exist?(Rails.root.join('config', 'locales', "#{locale}.yml")) }
+    missing_json_files = I18n.available_locales.reject { |locale| File.exist?(Rails.root.join('app', 'javascript', 'mastodon', 'locales', "#{locale}.json")) }
+
+    if missing_json_files.empty? && missing_yaml_files.empty?
+      puts pastel.green("OK")
```
ykzts:
```suggestion
      puts pastel.green('OK')
```

Prefer single-quoted strings when you don't need string interpolation or special symbols.

Gargron:
I'm not sure why the new rake command is failing in CircleCI :confused:

ykzts:
I think you need to add `- *install_system_dependencies` to `steps` of `check-i18n`.

Gargron:
Oh, I wasn't aware it wasn't in the same environment as the other tests.

ykzts (approved):

PR #11814, opened 2019-09-12
Change rate limit for media proxy

ykzts:
With 30 per 30 minutes, the limit is reached as soon as you tap the remote user's media tab.


ClearlyClaire (approved):
Those are still pretty expensive, but yeah, the limit was low enough to impede basic usage…

PR #11817, opened 2019-09-13
Added account featured tags API

noellabo:
Only works with local accounts until https://github.com/tootsuite/mastodon/pull/11595 is merged.

- `GET /api/v1/accounts/:id/featured_tags`
- returns id of `Tag`, not `FeaturedTag`
- Since it can be obtained from the public profile without authentication, the API is also not authenticated

```
[{"id":"606","name":"#distsns","url":"https://fedibird.com/@noellabo/tagged/distsns"},{"id":"1069","name":"#リレーの話","url":"https://fedibird.com/@noellabo/tagged/%E3%83%AA%E3%83%AC%E3%83%BC%E3%81%AE%E8%A9%B1"},{"id":"611","name":"#test","url":"https://fedibird.com/@noellabo/tagged/test"},{"id":"1043","name":"#fedibird","url":"https://fedibird.com/@noellabo/tagged/fedibird"}]
```

### DIFF
--- a/app/controllers/api/v1/accounts/featured_tags_controller.rb
+++ b/app/controllers/api/v1/accounts/featured_tags_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class Api::V1::Accounts::FeaturedTagsController < Api::BaseController
+  before_action :set_account
+  before_action :set_featured_tags
+
+  respond_to :json
+
+  def index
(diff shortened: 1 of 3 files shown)


noellabo:
Would it be better to use just `name` without using tag `id`?

Gargron:
Not stale

Gargron (approved):

PR #11818, opened 2019-09-13
Change unlisted custom emoji to not appear in autosuggestions

### LINKED ISSUE #11669: Unlisted custom emojis show up on autocompletion
### Expected behaviour

Unlisted custom emojis should not be shown on autocompletion panel.

### Actual behaviour

Unlisted custom emojis shows in the emoji autocompletor.

1. Unlist already added custom emoji.
2. Try to autocomplete it
3. It shows up

### Specifications

- tag: v2.9.3 (twingyeo.kr)

- firefox 70.0a1 (2019-08-24) (x64), Edge

I'm trying to rename non self-explanatory custom emojis to more self-explanatory one, by unlist old emojis and add new one. But after working on a few number of such emojis I relized that unlisted custom emoji shows up on the emoji autocompletion dialog.

image
image

Gargron:
Fix #11669

### DIFF
--- a/app/controllers/api/v1/custom_emojis_controller.rb
+++ b/app/controllers/api/v1/custom_emojis_controller.rb
@@ -7,6 +7,6 @@ class Api::V1::CustomEmojisController < Api::BaseController
 
   def index
     expires_in 3.minutes, public: true
-    render_with_cache(each_serializer: REST::CustomEmojiSerializer) { CustomEmoji.local.where(disabled: false).includes(:category) }
+    render_with_cache(each_serializer: REST::CustomEmojiSerializer) { CustomEmoji.listed.includes(:category) }
   end
 end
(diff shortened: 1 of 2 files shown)


ykzts (approved):

ClearlyClaire (approved):

trwnh:
How are unlisted emoji supposed to be discovered, then?

Gargron:
At least two different admins from different servers showed confusion about the previous behaviour. It seems like unlisted emoji are *not supposed* to be discovered. The main practical difference between an emoji being disabled and unlisted is that a disabled emoji will not render, while an unlisted will. The use case of unlisted emoji is therefore:

- Emoji for regulars who are "in the know" but which would be too noisy to include in the picker
- "Renaming" emoji without breaking old posts

PR #11821, opened 2019-09-13
Fix hashtags being split by ZWNJ character

### LINKED ISSUE #11761: ZWNJ is not supported in hashtags
### Expected behaviour

I expect ZWNJ to be included in hashtag. I need whole of "نرم‌افزار" be a hashtag.

### Actual behaviour

By now, Hashtag only includes non-ZWNJ characters. So in the given example above, only "نرم" gets converted to link.

You just add #نرم‌افزار and see how it only includes نرم as hashtag.

### Specifications

It is a general bug in Mastodon

Gargron:
Fix #11761

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -25,8 +25,9 @@ class Tag < ApplicationRecord
   has_many :featured_tags, dependent: :destroy, inverse_of: :tag
   has_one :account_tag_stat, dependent: :destroy
 
-  HASHTAG_NAME_RE = '([[:word:]_][[:word:]_·]*[[:alpha:]_·][[:word:]_·]*[[:word:]_])|([[:word:]_]*[[:alpha:]][[:word:]_]*)'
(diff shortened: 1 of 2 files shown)


ykzts (approved):

app/models/tag.rb
```diff
@@ -25,8 +25,9 @@ class Tag < ApplicationRecord
   has_many :featured_tags, dependent: :destroy, inverse_of: :tag
   has_one :account_tag_stat, dependent: :destroy
 
-  HASHTAG_NAME_RE = '([[:word:]_][[:word:]_·]*[[:alpha:]_·][[:word:]_·]*[[:word:]_])|([[:word:]_]*[[:alpha:]][[:word:]_]*)'
-  HASHTAG_RE = /(?:^|[^\/\)\w])#(#{HASHTAG_NAME_RE})/i
+  HASHTAG_SEPARATORS = "_\u00B7\u200c"
```
ThibG: I think the · would be more obvious written as is, but otherwise ok

ClearlyClaire (approved):

PR #11823, opened 2019-09-13
Remove deprecated `GET /api/v1/search` API

Gargron:
Use `GET /api/v2/search` instead


app/controllers/api/v2/search_controller.rb
```diff
@@ -1,8 +1,32 @@
...
+    SearchService.new.call(
+      params[:q],
+      current_account,
+      limit_param(RESULTS_LIMIT),
+      search_params.merge(resolve: truthy_param?(:resolve))
+    )
+  end
+
+  def search_params
+    params.permit(:type, :offset, :min_id, :max_id, :account_id)
```
ykzts: https://brakemanscanner.org/docs/warning_types/mass_assignment/

Gargron: Not mass assignment, so that's a false positive. Dismissed in CodeClimate.

ykzts: hm

ykzts (approved):

PR #11827, opened 2019-09-13
Add ability to pin thread columns

### LINKED ISSUE #11467: CW / Pin visible in all headers, even where inapplicable
mastodon.social master -- can suddenly see CW reveal icon in header, even for threads that have no CW. Can also see menu icon with "pin" option, although clicking this just makes the column disappear.

image

Gargron:
Fix #11467

### DIFF
--- a/app/javascript/mastodon/features/status/index.js
+++ b/app/javascript/mastodon/features/status/index.js
@@ -33,6 +33,7 @@ import {
 } from '../../actions/statuses';
 import { initMuteModal } from '../../actions/mutes';
 import { initReport } from '../../actions/reports';
+import { addColumn, removeColumn, moveColumn } from '../../actions/columns';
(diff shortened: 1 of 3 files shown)


ThibG:
I am not sure how useful that is without subscribing to updates from the thread

Gargron:
You get auto-updates if they arrive via notifications / home feed. Yes, it's not super useful, but then again how many people even know about the "pin" button? It feels like this would satisfy a few people without any side effects.

ThibG:
Hm, it seem to me that the use case would be to have quick access to some toots, more than having them displayed at all times. Such an use case could be addressed by bookmarks.

Otherwise OK, but every other pinned column does get update, this would introduce a confusing discrepancy

Gargron:
Quick access to some toots is also solved by favourites, I think the idea is having them always displayed.

PR #11829, opened 2019-09-13
Add search and sort functions to hashtag admin UI

mayaeh:
Add search and sort features to the hashtag page of the Admin UI.

In the "ORDER BY" field "MOST RECENT", "MOST POPULAR", and "LAST ACTIVE" do not know if this name is best.
Also, is this order appropriate?

They are sorted for the following reasons:

- MOST RECENT
  Sorted in order from the newly received hashtag by the server

- MOST POPULAR
  Sort in order from biggest "max_score"

- LAST ACTIVE
  Sorted in order from newest "last_status_at"

I think I should write a test too.
However, there are a lot of things I don't understand yet and I haven't built a test environment.

### DIFF
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -2,7 +2,6 @@
 
 module Admin
   class TagsController < BaseController
-    before_action :set_tags, only: :index
     before_action :set_tag, except: [:index, :batch, :approve_all, :reject_all]
     before_action :set_usage_by_domain, except: [:index, :batch, :approve_all, :reject_all]
     before_action :set_counters, except: [:index, :batch, :approve_all, :reject_all]
@@ -10,6 +9,7 @@ module Admin
     def index
       authorize :tag, :index?
 
+      @tags = filtered_tags.page(params[:page])
       @form = Form::TagBatch.new
     end
 
@@ -48,10 +48,6 @@ module Admin
 
     private
 
-    def set_tags
-      @tags = filtered_tags.page(params[:page])
-    end
-
     def set_tag
       @tag = Tag.find(params[:id])
     end
@@ -78,11 +74,14 @@ module Admin
       scope = scope.unreviewed if filter_params[:review] == 'unreviewed'
       scope = scope.reviewed.order(reviewed_at: :desc) if filter_params[:review] == 'reviewed'
       scope = scope.pending_review.order(requested_review_at: :desc) if filter_params[:review] == 'pending_review'
-      scope.order(max_score: :desc)
+      scope = scope.matches_name(filter_params[:name].to_s.strip) unless filter_params[:name].nil?
+      scope = scope.order('max_score DESC NULLS LAST') if filter_params[:order] == 'popular'
+      scope = scope.order('last_status_at DESC NULLS LAST') if filter_params[:order] == 'active'
+      scope.order(id: :desc)
     end
 
     def filter_params
-      params.slice(:context, :review, :page).permit(:context, :review, :page)
+      params.slice(:context, :review, :page, :name, :order).permit(:context, :review, :page, :name, :order)
     end
 
     def tag_params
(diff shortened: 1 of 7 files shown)


app/controllers/admin/tags_controller.rb
```diff
@@ -78,11 +74,14 @@ def filtered_tags
       scope = scope.unreviewed if filter_params[:review] == 'unreviewed'
       scope = scope.reviewed.order(reviewed_at: :desc) if filter_params[:review] == 'reviewed'
       scope = scope.pending_review.order(requested_review_at: :desc) if filter_params[:review] == 'pending_review'
-      scope.order(max_score: :desc)
+      scope = scope.matches_name(filter_params[:name].to_s.strip) unless filter_params[:name].nil?
```
mayaeh: Are there security concerns about the entered value?

config/locales/simple_form.en.yml
```diff
@@ -131,6 +131,8 @@ en:
         must_be_follower: Block notifications from non-followers
         must_be_following: Block notifications from people you don't follow
         must_be_following_dm: Block direct messages from people you don't follow
+      invite:
+        comment: Comment
```
mayaeh: This translation has nothing to do with this PR.
However, this is necessary for translating invitation link comments.
I would like to include this translation in this PR if possible.

app/views/admin/tags/index.html.haml
```diff
@@ -19,6 +19,26 @@
...
+    - Admin::FilterHelper::TAGS_FILTERS.each do |key|
+      = hidden_field_tag key, params[key] if params[key].present?
+
+    - %i(name).each do |key|
+      .input.string.optional
+        = text_field_tag key, params[key], class: 'string optional', placeholder: I18n.t("admin.tags.#{key}")
+
+    .actions
+      %button= t('admin.accounts.search')
+      = link_to t('admin.accounts.reset'), admin_tags_path, class: 'button negative'
```
mayaeh: These borrowed from existing translations.
Should I have these newly prepared?

app/controllers/admin/tags_controller.rb
ThibG: Seems fine to me.

app/views/admin/tags/index.html.haml
ThibG: It's fine, we do that for other views with the same buttons.

ClearlyClaire (approved):
Looks fine to me overall, but other parts use a model to model the filters, e.g. the `InstanceFilter` model. Maybe it would be more consistent to do it that way too?

mayaeh:
> Looks fine to me overall, but other parts use a model to model the filters, e.g. the `InstanceFilter` model. Maybe it would be more consistent to do it that way too?

Thank you very much! I will try to do that later.

app/models/tag_filter.rb
```diff
@@ -0,0 +1,62 @@
...
+  def scope_for(key, value)
+    case key.to_s
+    when 'context'
+      Tag.discoverable if value == 'directory'
+    when 'name'
+      Tag.matches_name(value)
+    when 'order'
+      scope_for_order(value)
+    when 'review'
+      scope_for_review(value)
```
mayaeh: I'm not sure if these method names are okay...
I'm also not sure if this is the best way to call these methods.
Is there a better idea?

app/models/tag_filter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    scope.order(id: :desc)
+  end
+
+  private
+
+  def scope_for(key, value)
+    case key.to_s
+    when 'context'
+      Tag.discoverable if value == 'directory'
```
Gargron: What happens if `scope.merge!(nil)`?

app/models/tag_filter.rb
Gargron: This seems OK.

app/models/tag_filter.rb
mayaeh: Oh… thank you, I will fix it.

PR #11830, opened 2019-09-13
Do not add margin-right when opening modal on mobile

ykzts:
There is hasn't scrollbar on mobile, so there is no need to add a margin-right for the scrollbar.


Gargron:
Wouldn't the scrollbar test return `0` on mobile? Width-based mobile detection includes small browser windows that do have non-zero scrollbars...

ykzts:
If `-webkit-scrollbar` is specified, scrollbars will be added to elements with `overflow: scroll` even on mobile.

Gargron:
But do webkit scrollbars take up width in the container? At least visually they are overlaid over the content rather than pushing it.

ykzts:
https://github.com/tootsuite/mastodon/blob/c00ab575b7450f133946fb94d7367b5d810a2d89/app/javascript/styles/mastodon/components.scss#L2132

The container scrollbar is not displayed on mobile. This is controlled by CSS.

Gargron:
What I am confused about is: Doesn't `getActualScrollbarWidth()` return 0 *anyway* when scrollbar is not displayed on mobile?

ykzts:
I feel something is wrong (browser bug?)

| iPhone | Android |
|--------|---------|
| iphone | android |

https://codepen.io/ykzts/pen/BaBqMdp

Gargron (approved):

PR #11835, opened 2019-09-13
Fix expiring polls not being displayed as such in the WebUI

ClearlyClaire:


Gargron (changes requested):
This is very likely to run into issues when components are re-used by the renderer and passed completely different props.

ThibG:
Good point (although honestly that's not something that should happen :weary:), I'll look into that

app/javascript/mastodon/components/poll.js
```diff
@@ -32,8 +32,36 @@ class Poll extends ImmutablePureComponent {
 
   state = {
     selected: {},
+    expired: this.props.poll.get('expired') || (new Date(this.props.poll.get('expires_at'))).getTime() < this.props.intl.now(),
```
Gargron: Set this to null.

Create `_setupTimer` method. Call this method from `componentDidMount`. Call this method from `componentWillReceiveProps` if `nextProps.poll.get('expired') !== this.props.poll.get('expired')` etc.

Gargron (changes requested):

app/javascript/mastodon/components/poll.js
ThibG: Setting it to null and then changing it in `componentDidMount` would incur an useless extra render cycle for every poll.

Gargron: Well, realistically, there's no reason to store `expired` in state, for example `RelativeTimestamp` component stores *current timestamp* in the state, and then the render method takes care of calculating the difference. That's probably the cleanest approach.

ThibG: I'd say there's no reason to store the current timestamp either, tbh.
Changed it to not use the deprecated `componentWillReceiveProps` and refactor it, it's cleaner imho, and should not do useless renders. However, it will re-fire the timer each time a poll gets updated even if its expiration date does not change. This is infrequent and cheaper than a render, though.

Gargron (approved):

PR #11837, opened 2019-09-13
Increase poll option length limit to 100 chars, number of options to 5

ClearlyClaire:


Gargron:
You've pointed out that even at 25 characters the options are sometimes longer than the bar. Previously that led to one-letter ellipsis, but now it leads to line breaks and thicker bars. Increasing that to 100 characters will ensure that the problem will become widespread and the uglier multiline inconsistent-height bars will be a common occurence. As I've mentioned that looks untidy and unprofessional and we should avoid showing it when possible.

ThibG:
Would something like the following make you reconsider?
Peek 16-09-2019 12-29

Alternative design
image

Gargron:
@ThibG Could clicking on the poll open the poll in a modal?

ThibG:
@Gargron honestly that would just be worse…? The modal would display it in a way similar to the proposed solutions and it would not be more intuitive or natural in any way?

PR #11839, opened 2019-09-13
improve UX of Direct Messages column

InspectorCaracal:
This commit aims to preserve the "conversation view" intent of the current DM column while making correct attribution easier to identify, using the same UI solutions as with Boosts.

The primary icon is the author of the visible status, with the conversation composite as an overlay.

The display name is the author of the visible status, with the conversation list as a status prepend.

image

### DIFF
--- a/app/javascript/mastodon/components/avatar_composite.js
+++ b/app/javascript/mastodon/components/avatar_composite.js
@@ -6,6 +6,7 @@ import { autoPlayGif } from '../initial_state';
 export default class AvatarComposite extends React.PureComponent {
 
   static propTypes = {
+    account: ImmutablePropTypes.map.isRequired,
     accounts: ImmutablePropTypes.list.isRequired,
     animate: PropTypes.bool,
     size: PropTypes.number.isRequired,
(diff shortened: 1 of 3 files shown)


trwnh:
I think it would be better to leave the "last status" part of the conversation the same as a regular status, and instead add the "conversation" ui elements as a header.

InspectorCaracal:
> I think it would be better to leave the "last status" part of the conversation the same as a regular status, and instead add the "conversation" ui elements as a header.

It sounds like you're under the impression that I added all the conversation stuff to a normal status display, but that's not the case.

The existing UI looks like this (with the same example DMs as the original post):

image

As you can see, there is currently no visual indication of who wrote the displayed status from this view.

This PR attempts to remedy that while altering the existing concept as little as possible and using already-established UX elements. Instead of using the composite icon as the status icon, it moves it to the overlay position and includes the status author itself as the base icon. Similarly, it replaces the list of account names with the name of the person who wrote the status, and moves the list to a status prepend.

This follows the same UI concepts as a boost: the original author's name is displayed in the author name location, and the original author's icon is displayed in the base avatar location. The previous conversation-view elements are then moved to the "booster" locations, as they're the source but not the author: the composite avatar is in the overlay position, and the list of account names is in the status prepend.

Gargron:
I was planning to redesign this UI at a future point. For example, I think on single-column mode there is enough space to have side-by-side list of conversations and the actual conversation, like in an e-mail client.

I also think that me re-using the status element was a mistake entirely and most of the buttons are a waste of space. DMs need their own menu, with "Mark as read" and "Delete conversation". And there is no reason to show media attachments or link previews on that screen.

trwnh:
Honestly there's still the UX mismatch with DMs technically thread-based instead of room-based. I think the whole DM system needs an overhaul with that in mind. It's annoying to have multiple "conversations" spawned from offshoots of the same thread, and I find myself missing when direct posts would simply appear in Home.

IMO the DM "Conversation" timeline should be conceived of as a separate, room-based feature.

InspectorCaracal:
@Gargron I think a complete overhaul is a *great* idea! If that's not expected to happen before the next version update, however, I'd like to suggest this PR be merged as a stop-gap measure until then.

PR #11847, opened 2019-09-15
Change dashboard to short number notation

noellabo:
The number on the dashboard overflows on the smartphone.

It is especially useful if the number of digits in interactions this week is large and you can switch to a short notation.

### DIFF
--- a/app/helpers/admin/dashboard_helper.rb
+++ b/app/helpers/admin/dashboard_helper.rb
@@ -9,10 +9,6 @@ module Admin::DashboardHelper
   end
 
   def short_number_format(number)
-    if Setting.short_number_enabled
-      number_to_human number, units: { unit: '', thousand: 'K', million: 'M', billion: 'B' }
-    else
-      number_with_delimiter number
-    end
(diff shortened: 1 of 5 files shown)


Gargron:
I think you can skip the setting part. Make it always use short notation with full notation in `title`.

noellabo:
I agree. Let's make a simple change.

app/helpers/admin/dashboard_helper.rb
```diff
@@ -7,4 +7,8 @@ def feature_hint(feature, enabled)
 
     safe_join([feature, content_tag(:span, indicator, class: class_names)])
   end
+
+  def short_number_format(number)
+    number_to_human number, units: { unit: '', thousand: 'K', million: 'M', billion: 'B' }
```
Gargron: Why is this necessary? Followers count on profile pages already uses short number format

noellabo: It existed for switching between short and full notation. Match the way you write your profile page.

Gargron (approved):

PR #11859, opened 2019-09-16
Fix “slow mode” issues

ClearlyClaire:
- Do not scroll weirdly when clicking the “load pending items” button
- Treat pending items as unread items

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -261,6 +261,13 @@ export default class ScrollableList extends PureComponent {
   handleLoadPending = e => {
     e.preventDefault();
     this.props.onLoadPending();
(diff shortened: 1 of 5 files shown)


ThibG:
The “X new items” click was interfering with the mouse movement scroll prevention behavior, which resulted in very weird scroll behavior. I fixed that.
By doing so, I decided to *not* scroll to top, which I guess is debatable.

Gargron:
Is the mouse movement thing still necessary if scrolling down will add new stuff to pending items instead of pushing the feed down?

ThibG:
yes, if you're still on top

Gargron (approved):

ThibG:
Pushed another commit preventing a scroll discrepancy I hadn't seen: when scrolled down and the “x new items” button appears, it pushed everything down and that wasn't corrected

PR #11867, opened 2019-09-16
Change timelines to add new items as pending when scrolled down

ClearlyClaire:
This makes the timelines use the same behavior as the “slow mode” whenever items get in and the timeline isn't scrolled to top.

Note that there remains small inconsistencies in the single column mode, in which a timeline won't work the same way if it was switched out while scrolled on top (it won't add things to the “pending” items list, but add them directly) or not (it will add them as pending), and the notifications (will always add to pending when not displayed)

### DIFF
--- a/app/javascript/mastodon/reducers/notifications.js
+++ b/app/javascript/mastodon/reducers/notifications.js
@@ -35,11 +35,11 @@ const notificationToMap = notification => ImmutableMap({
 });
 
 const normalizeNotification = (state, notification, usePendingItems) => {
-  if (usePendingItems) {
+  const top = state.get('top');
+
+  if (usePendingItems || !top || !state.get('pendingItems').isEmpty()) {
     return state.update('pendingItems', list => list.unshift(notificationToMap(notification))).update('unread', unread => unread + 1);
   }
-
-  const top = state.get('top');
 
   if (!top) {
     state = state.update('unread', unread => unread + 1);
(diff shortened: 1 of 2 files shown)


ThibG:
(I am not sure exactly which way those discrepancies should be solved)

Gargron (approved):

ykzts:
autopager is not working properly (maybe).

ThibG:
Autopager?

mayaeh:
@ThibG Please watch [this video](https://taruntarun.net/media_files/Mastodon/media_attachments/files/000/565/533/original/21bf5e52bb39ab17.mp4).
(This video is not in slow mode.)
It seems that "Load more" at the bottom edge is also working at the top edge.

zunda:
I see similar difficulty here.

Clicking the "* more toots" (30件の新着) link at the top of HTL loads older toots (toot from 16 minutes ago on top of one from 3 minutes ago) like below. Sometimes I see toots in expected order.

top-of-htl


Clicking the "Load more" (もっと見る) link at the bottom of HTL doesn't change what are shown in the timeline.

bottom-of-htl

Movies are cropped to the top and bottom of HTL from Mastodon in multi-column mode on Firefox 69.

ThibG:
oh, I see, going to fix that!

zunda:
Fixed with #11883. THANK YOU ❤️

PR #11876, opened 2019-09-17
Fix audit log error when custom emoji is copied from remote server

highemerly:
### Expected behaviour

When I copy custom emojis from remote server, 
- Custom emoji can be copied.
- This action is recorded to audit logs.

### Actual behaviour

When I copy custom emojis from remote server, 
- Custom emoji can be copied.
- This action is NOT recorded to audit logs.

In admin/custom_emojis pages, check some custom emoji and click "copy" button.

### Specifications

master branch

### DIFF
--- a/app/models/custom_emoji.rb
+++ b/app/models/custom_emoji.rb
@@ -63,7 +63,7 @@ class CustomEmoji < ApplicationRecord
   def copy!
     copy = self.class.find_or_initialize_by(domain: nil, shortcode: shortcode)
     copy.image = image
-    copy.save!
+    copy.tap(&:save!)
   end
 
   class << self


app/models/form/custom_emoji_batch.rb
```diff
@@ -90,8 +90,8 @@ def copy!
     custom_emojis.each { |custom_emoji| authorize(custom_emoji, :copy?) }
 
     custom_emojis.each do |custom_emoji|
-      copied_custom_emoji = custom_emoji.copy!
-      log_action :create, copied_custom_emoji
+      custom_emoji.copy!
+      log_action :create, custom_emoji
```
Gargron: I need to check why output of `custom_emoji.copy!` doesn't work, but this would be wrong, since it will log the creation of the *remote* emoji, and not the local one...

highemerly: I understand your concern. It seems that the output of `custom_emoji.copy!` is just true/false. Following changes is acceptable for you?

```
diff --git a/app/models/custom_emoji.rb b/app/models/custom_emoji.rb
index edb1bec..0dacaf6 100644
--- a/app/models/custom_emoji.rb
+++ b/app/models/custom_emoji.rb
@@ -63,7 +63,7 @@ class CustomEmoji < ApplicationRecord
   def copy!
     copy = self.class.find_or_initialize_by(domain: nil, shortcode: shortcode)
     copy.image = image
-    copy.save!
+    copy.tap(&:save!)
   end
```

Gargron: Oh yes, that is a good fix! I've run into that mistake a few times before

highemerly: Sent new fix.

Gargron (approved):

PR #11877, opened 2019-09-17
Fix TOTP codes not being filtered from logs during enabling/disabling

Gargron:
Not a serious issue because they are meaningless past single use


ykzts (approved):

ykzts:
oh sorry, I was mistaken.

PR #11878, opened 2019-09-17
Add challenge to 2FA settings, e-mail notifications

### LINKED ISSUE #3961: 2FA recovery codes shouldn’t be regenerable

Gargron:
Fix #3961

Adds a "challenge" page which asks for current password if the user has a password. If the challenge is passed, that is remembered for the next hour so the user is not asked again.

This is added on top of:

1. Enabling 2FA
2. The 2FA recover codes re-generation

image

There are a few other places that ask for confirmation using different UX, most notably account settings (e-mail/password) and account deletion, but it probably makes sense to keep those as they are now.

Changes "Generate recovery codes" link from looking like a text link to looking like a button to make it clearer that it's an action rather than just a page to be viewed.

image

Adds e-mail notifications about changes to 2FA settings and recovery code re-generation.

image


app/models/user.rb
```diff
@@ -264,17 +264,20 @@ def invite_code=(code)
   end
 
   def password_required?
-    return false if Devise.pam_authentication || Devise.ldap_authentication
+    return false if external?
+
```
ThibG: This makes the tests fail. However, I actually do not understand the current tests. Where does “pam_user1” come from?

ThibG: (This specific test pass now that you updated it, but I'm still curious at how it works)

Gargron: Yeah it's not obvious but the fake PAM data is defined in config/environments/test.rb

ClearlyClaire (approved):

PR #11885, opened 2019-09-18
Add table of contents to about page

Gargron:
Generate a table of contents out of the extended site description HTML.

Move the public listing of domain blocks (when enabled) to the about page. Instead of displaying them in a table with jargon words such as "suspend" or "silence", everything is presented in a format readable by the average person, with focus on what impact it has on that person.

image

Also, improve vertical flow of `.rich-formatting` elements, i.e. improve CSS for headers, lists, inline elements and more.


ykzts (approved):

ThibG:
I think we can keep the ToC thing and have the block information directly on `/about/more`, but can we please move back to the older design (maybe with the reworded severity explanations, and maybe with a link to the explanation in the severity level)?

We went from this:
image

To this:
image

Pardon my words, but this looks untidy and unprofessional

0xa:
yeah i have to agree, i like the helpful text but i'm very worried about how it would look with longer reasons and eighty entries.

the domain/severity table is clearly more readable and browsable. i think it would be better to take the good parts of both solutions and keep the table with the domain and a very short severity summary, with entries that unfold to show the full severity description and the full reason, however long both might be.

having a dedicated page also might be a better option, since about pages tend to already have a lot of content, and blocklists tend to be long and grow with time. maybe it could be linked in the table of content anyway, i'm sure there is a way to cleanly indicate that it's a separate page

PR #11896, opened 2019-09-20
Change conversations UI

### LINKED ISSUE #9860: possibility to delete received direct messages
Sometimes I receive really stupid direct messages. These then stay in my direct message column for ever and for a really long time on the first page as I use direct messages only rarely.

I think it would be a nice enhancement if i could delete direct messages from my "direct message"-column. Direct messages are something more personal, 1-to-1, like email, messengers. Think it would be good if people could manage them a little bit more like that.

### LINKED ISSUE #10434: Making it more visible to whom a user is actually talking to in the DMs
### Expected behaviour*

image

### Actual behaviour*

image

*Usernames and profile pictures are pixelated for privacy reasons

Sending someone a DM

### Specifications

Mastodon 2.7.4 (stable)
Vivaldi 2.4.1488.35 (64bit) on Arch Linux (Kernel 5.0.5)

### LINKED ISSUE #11414: DM display names can be linked to wrong user
image

1. Hover over a DM where you were the last participant

### Expected behaviour
- The avatar/name and link should match

### Actual behaviour
- The avatar/name do not match the link

### Notes
I actually don't think there's a clear fix for this because if you weren't the last participant, the behavior somewhat makes sense. Also, it makes less sense for group DMs overall. I would say that the DM conversation component needs to be redesigned to better account for this. The status model doesn't exactly line up 1:1 with the way conversations are displayed currently; with the old status-only direct timeline, it wasn't really an issue since the statuses were displayed directly, but with the conversations timeline, there needs to be more separation from the "latest status" and "participants". this might also help other points of confusion, e.g. #10900 #9992 #10434 #10675 (although personally i'd rather just have the direct timeline back).

Gargron:
Fix #11414, fix #9860, fix #10434

image

There was a bug preventing showing more than 2 avatars. Now 4 avatars can be displayed at once, with anything extra marked with a number label on top. Every participant is listed above the DM individually, and each name is clickable. Attachments are always displayed in list form. The action bar now includes DM-specific items such as deleting the whole conversations or marking it as read (without opening). An option to mute the conversation is more prominent.

Resolve #11839

### DIFF
--- a/app/javascript/mastodon/features/direct_timeline/components/conversation.js
+++ b/app/javascript/mastodon/features/direct_timeline/components/conversation.js
@@ -2,9 +2,28 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import ImmutablePureComponent from 'react-immutable-pure-component';
-import StatusContainer from '../../../containers/status_container';
+import StatusContent from 'mastodon/components/status_content';
+import AttachmentList from 'mastodon/components/attachment_list';
+import { defineMessages, injectIntl, FormattedMessage } from 'react-intl';
+import DropdownMenuContainer from 'mastodon/containers/dropdown_menu_container';
+import AvatarComposite from 'mastodon/components/avatar_composite';
+import Permalink from 'mastodon/components/permalink';
+import IconButton from 'mastodon/components/icon_button';
+import RelativeTimestamp from 'mastodon/components/relative_timestamp';
+import { HotKeys } from 'react-hotkeys';
 
-export default class Conversation extends ImmutablePureComponent {
+const messages = defineMessages({
+  more: { id: 'status.more', defaultMessage: 'More' },
+  open: { id: 'conversation.open', defaultMessage: 'View conversation' },
+  reply: { id: 'status.reply', defaultMessage: 'Reply' },
+  markAsRead: { id: 'conversation.mark_as_read', defaultMessage: 'Mark as read' },
+  delete: { id: 'conversation.delete', defaultMessage: 'Delete conversation' },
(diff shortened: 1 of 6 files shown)


app/javascript/mastodon/features/direct_timeline/components/conversation.js
```diff
@@ -42,22 +74,89 @@ export default class Conversation extends ImmutablePureComponent {
...
+      menu.push(null);
+    }
+
+    menu.push({ text: intl.formatMessage(messages.delete), action: this.handleDelete });
+
+    const names = accounts.map(a => <Permalink to={`/accounts/${a.get('id')}`} href={a.get('url')} key={a.get('id')} title={a.get('acct')}><bdi><strong className='display-name__html' dangerouslySetInnerHTML={{ __html: a.get('display_name_html') }} /></bdi></Permalink>).reduce((prev, cur) => [prev, ', ', cur]);
+
+    const handlers = {
+      reply: this.handleReply,
+      favourite: this.handleFavourite,
```
ThibG: slightly weird to have a handler for that when there's nothing in the UI to show the fav state nor do it with something else than the keyboard shortcut

ClearlyClaire (approved):
Apart from the fav handler thing, LGTM

app/javascript/mastodon/features/direct_timeline/components/conversation.js
Gargron: Oh oops, it isn't hooked up to anything, actually. I'm not sure if the fav button should be there. Don't want it to look too similar to a status. It also doesn't fav the conversation, but the most recent status, so... Not sure.

ThibG: yeah, I think it's best without it at all (I notice that glitch-soc's slight redesign of the conversations UI has the same issue: no button—on purpose—but the handler is still on)

PR #11900, opened 2019-09-20
Bump version to 3.0.0rc1

Gargron:


Gargron:
I'll write the changelog in a separate PR. It's time to bump the version on servers running master.

Issue #9880, opened 2019-01-20
Give unified media attachment URL and Give 301 redirect for old system URL


Kjwon15:
Some instances have been moved their media storage from system storage or another object storage.
Toots before moving media storage are not correctly rendered on remote instances because media URL on ActivityPub protocol is changed and not refetched.

Here is my suggestion for solving this.

## 1. Use media proxy url or some other unified URL for media attachment on ActivityPub
* before: `domain.tld/system/media/000/000/001/abc.png`
* after: `domain.tld/media_proxy/1.png` → `media.domain.tld/system/media_attachments/000/000/000/abc.png` (302 Temporary redirect)

## 2. Give 301 redirect for old URL
`domain.tld/system/media/000/000/001/abc.png` → `domain.tld/media_proxy/1.png`(301 permanent redirect)  → `media.domain.tld/system/media_attachments/000/000/001/abc.png` (302 temporary redirect)

Kjwon15:
I see that database call is very slow. But it(301) can be cached on nginx. isn't it?
Or, We just notice that the administrator should setup redirect manually.

PR #11915, opened 2019-09-22
Note that media can be broken on remote when not to set S3_ALIAS_HOST

tribela:
## Reason

Since Mastodon doesn't use media proxy for attachments (#9880) on AP protocol, Remote instances cannot re-fetch attachment after Object Storage provider has changed.
So, Take a note to set `S3_ALIAS_HOST` properly and permanently.

>@kaniini i suspect there is a regression
without using a bounce URL, it is impossible to change CDN providers
> @kaniini
several instances changed from wasabi recently, and broke their old media in the process (as observed by Pleroma instances anyway)
if a CNAME is used for the bucket, it would be safe indeed


Gargron:
I'm not sure people look into that file anymore? The setup is automatic through rake mastodon:setup, so new users would not see this comment. This information probably belongs in documentation.

PR #11916, opened 2019-09-22
Fix ugly TOC when title is too long

tribela:
When the title is too long, TOC and content looks ugly

## Before
Screenshot from 2019-09-22 19-03-37

## After Screenshot from 2019-09-22 19-09-36

### DIFF
--- a/app/javascript/styles/mastodon/containers.scss
+++ b/app/javascript/styles/mastodon/containers.scss
@@ -169,7 +169,7 @@
 .grid-4 {
   display: grid;
   grid-gap: 10px;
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Kjwon15:
I think there are more proper way to fix this

Kjwon15:
OK `grid-template-columns: repeat(4, minmax(0, 1fr))` is the proper way to fix this problem.
IDK why minmax(0, 1fr) fixes them, But it works

https://css-tricks.com/preventing-a-grid-blowout/

PR #11917, opened 2019-09-22
Add voters count support

ClearlyClaire:
For newly-created polls, count voters in addition to the number of votes.
Present them in the REST API as `voters_count` and over ActivityPub as `toot:votersCount`.

When that information is available, the text now says “X voters” instead of “X votes”, and the tallies use that for percentages instead of the total number of votes (which is far less useful for `anyOf` polls) 

Marked as a draft because it probably could use more tests, and I'm very unsure about the attribute it should use.

image

### DIFF
--- a/app/javascript/mastodon/components/poll.js
+++ b/app/javascript/mastodon/components/poll.js
@@ -102,10 +102,11 @@ class Poll extends ImmutablePureComponent {
 
   renderOption (option, optionIndex, showResults) {
     const { poll, disabled, intl } = this.props;
-    const percent = poll.get('votes_count') === 0 ? 0 : (option.get('votes_count') / poll.get('votes_count')) * 100;
-    const leading = poll.get('options').filterNot(other => other.get('title') === option.get('title')).every(other => option.get('votes_count') >= other.get('votes_count'));
-    const active  = !!this.state.selected[`${optionIndex}`];
-    const voted   = option.get('voted') || (poll.get('own_votes') && poll.get('own_votes').includes(optionIndex));
+    const pollVotesCount  = poll.get('voters_count') || poll.get('votes_count');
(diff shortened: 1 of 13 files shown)


app/javascript/mastodon/components/poll.js
```diff
@@ -101,7 +101,8 @@ class Poll extends ImmutablePureComponent {
 
   renderOption (option, optionIndex, showResults) {
     const { poll, disabled } = this.props;
-    const percent            = poll.get('votes_count') === 0 ? 0 : (option.get('votes_count') / poll.get('votes_count')) * 100;
+    const pollVotesCount         = poll.get('voters_count') || poll.get('votes_count');
```
nightpool: alignment.

Gargron:
I'm hesitant to add another custom property. Is the problem being solved the confusion around submitting another multiple-choice vote -> number of votes going up? That doesn't seem wrong to me per se.

ThibG:
It's not just that. 40 people voting just one option each is indistinguishable from 10 people all voting all the options, although those have very different meanings ultimately

trwnh:
instead of `votersCount` i would propose something like using `totalItems` of the `replies` Collection. whether the `Question` is `oneOf` or `anyOf` should be irrelevant; each reply should contain an array of the selected answers.

ThibG:
@trwnh no, that would mean listing all the votes, and each vote is for one option only, so that would just be equivalent to what we currently do

trwnh:
Wait, so one vote creates multiple responses? That doesn't make sense...

ThibG:
Yes, a vote is a reply object with the same name as the option it is for, voting multiple options require multiple such objects.

ThibG:
Marked as ready for review as there seem to be an agreement that the AS vocabulary isn't enough for this and pleroma devs are ok with the chosen attribute name.

Gargron (approved):

PR #11943, opened 2019-09-24
Fix authentication before 2FA challenge

Gargron:
Regression from #11831

Thankfully a master-only, unreleased bug.

It is my mistake but the Devise and Warden code around that functionality is very misleading: The `sign_in` call in the `Devise::SessionsController` is a lie, because `warden.authenticate!` the controller calls earlier already sets the session. Not only that, if you remove the `warden.authenticate!` call from the controller altogether, there is still another part of Devise that will call `warden.authenticate` which will succeed when the correct username/password are in the request params and also set the session. To add insult to injury, `:database_authenticatable` also has a side-effect, by setting the rememberable token in a cookie. So as it turns out there was no way to keep the code DRY and run configured strategies without any side effects just to find the correct user record to check for 2FA setting.

Instead we have to duplicate the functionality to perform the 2FA setting check, i.e. finding the user from LDAP, PAM or database depending on configuration in our own sessions controller. It is also important that all the Warden strategies fail for users who have 2FA enabled, so that the aforementioned fuckery does not get in the way.


lib/devise/pam_authenticatable.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Devise
+  module Strategies
+    class PamAuthenticatable
```
nightpool: needs to inherit from Devise:: Strategies::Base

nightpool (changes requested):

nightpool (approved):

PR #11944, opened 2019-09-24
Cast multipart threshold to integer

ykzts:


Gargron (commented):
A header will be transmitted as string anyway. Perhaps you want the cast on the megabytes object?

ykzts:
Yes, I thought so. However, this is handled as a number internally by aws-sdk-s3.

Also `Numeric#megabytes` returns `Integer`.

https://github.com/rails/rails/blob/v5.2.3/activesupport/lib/active_support/core_ext/numeric/bytes.rb#L30

Gargron (approved):

PR #11954, opened 2019-09-25
Fix search error when ElasticSearch is enabled but not available

tribela:
Like #9654, It fallback to database search when ES is not available.

### DIFF
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -43,7 +43,7 @@ class AccountSearchService < BaseService
 
     @search_results ||= begin
       results = from_elasticsearch if Chewy.enabled?
-      results = from_database if results.nil?
+      results ||= from_database
       results
     end
   end
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -43,7 +43,7 @@ class AccountSearchService < BaseService
 
     @search_results ||= begin
       results = from_elasticsearch if Chewy.enabled?
-      results = from_database if results.blank?
+      results = from_database if results.nil?
       results
     end
   end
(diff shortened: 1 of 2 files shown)


app/services/account_search_service.rb
```diff
@@ -92,6 +90,8 @@ def from_elasticsearch
     ActiveRecord::Associations::Preloader.new.preload(records, :account_stat)
 
     records
+  rescue Faraday::ConnectionFailed, Parslet::ParseFailed
```
Gargron: I agree with this, but not the other change, because for 0 results it will double the work load

Gargron (changes requested):

app/services/account_search_service.rb
Kjwon15: Then, how about return nil to prevent double work?
ES connection failed → return nil → do Database search
ES returned 0 results → Just use that

Kjwon15: Or just return empty result without fallback
How is your opinion?
I like to use first one to make giving results as possible

koyuawsmbrtn:
+1 for this feature, es sucks a bunch of resources, because java likes to leak memory

app/services/account_search_service.rb
```diff
@@ -42,11 +42,9 @@ def search_results
     return [] if limit_for_non_exact_results.zero?
 
     @search_results ||= begin
-      if Chewy.enabled?
-        from_elasticsearch
-      else
-        from_database
-      end
+      results = from_elasticsearch if Chewy.enabled?
+      results = from_database if results.nil?
```
Gargron: OK, just write it like `results ||= from_database`

Gargron (approved):

PR #11962, opened 2019-09-26
Fix hashtag batch actions not redirecting back with right filters

Gargron:
Regression from #11829


ykzts (approved):

mayaeh:
I'm sorry I missed it, thank you

PR #11963, opened 2019-09-26
Fix relays UI being available in whitelist/secure mode

Gargron:
Fix relays UI referencing relay that is not functional

I have re-named the dashboard entry for `authorized_fetch_mode` to "Secure mode" because that's what the PR was called and how it's called in the changelog. I'm not 100% sure on this but fear that "Authorized fetch mode" might just be too jargony.

### DIFF
--- a/app/controllers/admin/relays_controller.rb
+++ b/app/controllers/admin/relays_controller.rb
@@ -3,6 +3,7 @@
 module Admin
   class RelaysController < BaseController
     before_action :set_relay, except: [:index, :new, :create]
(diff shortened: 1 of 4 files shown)


ykzts (approved):

ThibG:
It makes sense to keep the feature available for authorized fetch mode, though (and to a lesser extent, for whitelist mode also)

Basically every relay software but the one you wrote relies on `Announce`s rather than LDSig, so authorized fetch mode wouldn't change anything for them.

PR #11965, opened 2019-09-26
Fix overflow on conversations

cutls:
# What?
Conversations(DMs) layout not to overflow.
## Before
image
## After
image  
Use Flexbox instead of using float around name and timestamp.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -6418,13 +6418,17 @@ noscript {
     flex: 1 1 auto;
     padding: 10px 5px;
     padding-right: 15px;
+    word-break: break-all;
+    overflow: hidden;
 
     &__info {
       overflow: hidden;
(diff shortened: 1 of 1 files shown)


Gargron:
I had to use float instead of flexbox because with flexbox if the name gets too long, it pushes the width of the container out.

cutls:
Ah, although using float, long-name make overflow.( I'm preparing the screenshot )  
Only display:flex, of course, they overflow. 
However, use flex-shrink to avoid overflowing name and timestamp.

cutls:
## Long Name on mastodon.social
Although contents is short, it overflow
image
## Fixed
image

cutls:
To avoid overflow when content is long,
only to use ` word-break: break-all;`

Gargron (approved):

Gargron:
The word break rule affects many common words, which makes them harder to read. That is not optimal...

PR #11967, opened 2019-09-26
Fix public list of domain blocks being too verbose on about page

Gargron:
image


ThibG:
I dislike the idea of having domains listed twice, it's confusing and things would be better in one single place.

Also, it's not immediately clear what applies to what here, maybe the different types should be separated by subsections.

In any case, this is *way* better than what is in master.

ykzts (approved):

PR #11970, opened 2019-09-27
Add translation strings for AdminUI custom emojis

mayaeh:
This PR enables these translations:


Gargron (approved):

mayaeh:
I'm sorry, there was still a string that could not be translated.
I will make a PR it with other matters.

PR #11975, opened 2019-09-27
Change silences to always require approval on follow

### LINKED ISSUE #11958: Change domain silence to quarantine inbound follow requests instead of drop them
### Pitch

After discussing with @Gargron in Discord about my plans to extend Pleroma's quarantine system to optionally require all follow requests from a given instance to be explicitly approved or rejected, Gargron expressed interest in adjusting domain silences to do the same.

### Motivation

Right now, there are scenarios where domain silence feature in Mastodon is too harsh.  Pleroma uses a different strategy for handling these situations which is more granular, but in both cases, API support and UX changes are needed to facilitate the changes, because the original API assumes a UX where follow requests are only approved explicitly if an account is locked.

### Getting There From Here

We need to add `pending_follow_requests` to the Account entity rendered at `/api/v1/verify_credentials`.

If `pending_follow_requests` is greater than zero, then the follow requests UI should be visible.  Alternatively if `locked` is true, then the follow requests UI should also be visible.

ClearlyClaire:
So far, silenced (mutes decided by an instance moderator) and user-muted accounts can follow local users just like any other account. The only difference being that they would not generate notifications when doing so.

Having silenced accounts silently follow users may be misleading and undesirable.

This PR changes the following:
- **Silenced accounts** (whether individually or through domain silence) will always generate a follow request, which won't be auto-accepted, regardless of whether the target account is locked. The rationale for this can be found in #11958. One downside though is that “follow spam” would not be prevented by silencing an instance.
- ~~**User-muted accounts** will always generate a follow request, which won't be auto-accepted, regardless of whether the target account is locked. Indeed, the question “should a muted user be able to follow the person muting them?” has been brought up repeatedly, and both answers to that have disadvantages. But requesting approval for such follows would solve unexpected follows while not changing the current behavior too drastically. In case a muted users sends repeated follow requests, #11562 would enable silently discarding them.~~ Removed to avoid leaking user mutes
- Add `follow_requests_count` to `/api/v1/verify_credentials` (fixes #11958)
- Make the WebUI check for pending follow requests regardless of whether the account is locked
- Make the API result for follows not assume the follow went through if the following account is locally silenced

### DIFF
--- a/app/controllers/api/v1/accounts_controller.rb
+++ b/app/controllers/api/v1/accounts_controller.rb
@@ -33,7 +33,7 @@ class Api::V1::AccountsController < Api::BaseController
   def follow
     FollowService.new.call(current_user.account, @account, reblogs: truthy_param?(:reblogs))
 
-    options = @account.locked? ? {} : { following_map: { @account.id => { reblogs: truthy_param?(:reblogs) } }, requested_map: { @account.id => false } }
+    options = @account.locked? || current_user.account.silenced? ? {} : { following_map: { @account.id => { reblogs: truthy_param?(:reblogs) } }, requested_map: { @account.id => false } }
 
     render json: @account, serializer: REST::RelationshipSerializer, relationships: relationships(options)
   end
--- a/app/javascript/mastodon/features/getting_started/index.js
+++ b/app/javascript/mastodon/features/getting_started/index.js
@@ -77,16 +77,14 @@ class GettingStarted extends ImmutablePureComponent {
   };
 
   componentDidMount () {
-    const { myAccount, fetchFollowRequests, multiColumn } = this.props;
+    const { fetchFollowRequests, multiColumn } = this.props;
 
     if (!multiColumn && window.innerWidth >= NAVIGATION_PANEL_BREAKPOINT) {
       this.context.router.history.replace('/timelines/home');
       return;
     }
 
-    if (myAccount.get('locked')) {
-      fetchFollowRequests();
-    }
+    fetchFollowRequests();
   }
 
   render () {
@@ -134,7 +132,7 @@ class GettingStarted extends ImmutablePureComponent {
 
     height += 48*3;
 
-    if (myAccount.get('locked')) {
+    if (myAccount.get('locked') || unreadFollowRequests > 0) {
       navItems.push(<ColumnLink key={i++} icon='user-plus' text={intl.formatMessage(messages.follow_requests)} badge={badgeDisplay(unreadFollowRequests, 40)} to='/follow_requests' />);
       height += 48;
     }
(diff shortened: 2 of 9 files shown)


ThibG:
While the new notification type would be useful, it is not strictly needed for our purposes, and it may break some clients. I think that can be postponed.

Gargron:
> API results for locally-silenced accounts will still pretend the follow to be immediate when trying to follow an open-follow user. This is mainly to avoid leaking the fact that the user has been silenced (although it can still be found out easily by refreshing and figuring out that no follow gets through immediately)

This is no longer necessary as the "account status" section reveals if your account is "limited".

Gargron:
> - add follow request notification type
> - handle follow request notifications in UI

As you mentioned, it would make follow spam something you have no escape from. The advantage of locked accounts was, until now, that follow requests were not spamming your notifications and you could address them at your own leisure. What we need, indeed, is a more prominent way of showing follow requests in the UI even if the account is not locked, but I do not think that putting follow requests into notifications will help. Consider Twitter's DMs UI: You have one tab for normal DMs, and one tab for DM "requests", i.e. from people you haven't accepted. There is similar UX with Facebook messages. I think something like that would make sense. One place for normal notifications, and one place for follow requests. It's *kind of* like that already except that the follow requests UI is far away from notifications visually and the count doesn't load on page load unless your account is locked.

ThibG:
I used “notification” pretty liberally here, the idea just being to push the update through the streaming server and be picked up by the front-end without having to poll the list.

ThibG:
> > API results for locally-silenced accounts will still pretend the follow to be immediate when trying to follow an open-follow user. This is mainly to avoid leaking the fact that the user has been silenced (although it can still be found out easily by refreshing and figuring out that no follow gets through immediately)
> 
> This is no longer necessary as the "account status" section reveals if your account is "limited".

Alright, changing that!

realityfabric:
For silenced users / instances, could requests be auto dropped (as opposed to auto accepted)? That seems like the expected behavior (and would stop follow spam, right?)

ThibG:
> For silenced users / instances, could requests be auto dropped (as opposed to auto accepted)? That seems like the expected behavior (and would stop follow spam, right?)

That sounds too harsh, as users from those instances will then be completely unable to follow local users. Up until now the model was basically “hide those remote users, but let them interact with people who know them”, this would prevent it. This PR removes the auto-accept and moves to pending accepts instead (this doesn't change anything for people with locked accounts).

ThibG:
Hm, in fact there is one thing I'm not too sure about: this could potentially leak that one person mutes another, if the follow doesn't get through immediately…

Gargron:
Maybe mutes should continue working like before. We're specifically interested in domain silences here.

ThibG:
> Maybe mutes should continue working like before. We're specifically interested in domain silences here.

Alright, makes sense. We should definitely mention the follow bit in the confirmation dialog, though (which is addressed, among other things, in #11562)

Gargron (approved):

PR #11980, opened 2019-09-27
Fix account migration not affecting followers on origin server

Gargron:


ykzts (approved):

app/workers/move_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+    true
+  end
+
+  private
+
+  def rewrite_follows!
+    @source_account.passive_relationships
+                   .where(account: Account.local)
+                   .in_batches
+                   .update_all(target_account: @target_account)
```
ThibG: Won't that mess up the follow counts?

PR #11988, opened 2019-09-28
Various updates for Nanobox deploys

danhunsaker:


.env.nanobox
```diff
@@ -11,24 +11,14 @@ DB_NAME=gonano
...
-
-# ImageMagick optimizations
-MAGICK_TEMPORARY_PATH=/app/tmp
-MAGICK_MEMORY_LIMIT=128MiB
-MAGICK_MAP_LIMIT=64MiB
-MAGICK_TIME_LIMIT=15
-MAGICK_AREA_LIMIT=16MP
-MAGICK_WIDTH_LIMIT=8KP
-MAGICK_HEIGHT_LIMIT=8KP
+BIND=0.0.0.0
```
Gargron: Because Puma config is read before dotenv-rails is loaded, the BIND variable cannot be used from .env files :confused:

Gargron: I'm not sure if you need to bind to 0.0.0.0 anyway if you are running nginx in the same container / on the same machine?

Gargron:
Alright, it's probably not going to work the way you expect it to, but I suspect you may not even need to bind to 0.0.0.0, so I am just going to merge this as is.

Gargron (approved):

PR #11992, opened 2019-09-28
Add explanation to mute dialog, refactor and clean up mute/block UI

ClearlyClaire:
Move some stuff out of #11562

image


ykzts (approved):

Gargron:
I believe there should be space between the "heading", which is the confirmation prompt, and the "explanation" element. Also, I am now noticing that it's odd to mix a left-aligned toggle element with a center-aligned text. Either both should be centered, or they all should be left-aligned.

Gargron:
"This will hide posts from them and posts mentioning them, but it will still allow them to see your posts and follow you."

ThibG:
Updated to this:
image

ThibG:
Updated again, everything is left-aligned now

image

PR #11995, opened 2019-09-29
Change @ icon in notification filter to reply-all

koyuawsmbrtn:
This would make icon thickness more consistent, especially when using custom fonts

### DIFF
--- a/app/javascript/mastodon/features/notifications/components/filter_bar.js
+++ b/app/javascript/mastodon/features/notifications/components/filter_bar.js
@@ -64,7 +64,7 @@ class FilterBar extends React.PureComponent {
           onClick={this.onClick('mention')}
           title={intl.formatMessage(tooltips.mentions)}
         >
-          <Icon id='at' fixedWidth />
+          <Icon id='reply-all' fixedWidth />
         </button>
         <button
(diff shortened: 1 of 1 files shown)


Gargron:
The reply button is what we use on the reply function but the @ symbol is universally understood as "mentions". Would this not be misleading? I don't really ever look at that quick bar so I don't have a strong opinion but I don't understand the thickness issue either.

koyuawsmbrtn:
I don't think this is misleading, because most mentions are replies anyway and mentions by just mentioning people is not happening too much. Also let me attach some before and after screenshots to make my point more understandable.

before:
grafik

after:
grafik

As you can see I was using custom fonts on both of these and the @ glyph has the same font thickness as any other font not giving the focus on other elements.

Also to come back on why using `reply-all` here is that the user also can be prompted to reply to a message so I think this is less misleading then?

Gargron:
The @ icon isn't actually the `@` character, it's a glyph just like the other icons, so it should not be affected by font width or custom fonts. That is just how it looks in Font Awesome. But yeah, okay, that looks fine.

Gargron (approved):

PR #12002, opened 2019-09-29
Add a nodeinfo endpoint

### LINKED ISSUE #6739: Expose instance metadata as json-ld
In the context of https://github.com/TheKinrar/mastodon-instances and https://github.com/libresh/catalog it would be a great feature to be able to scrape information in an automated manner.

As an admin, I already fill these metadata (description, name...) in the admin interface.

Exposing them as `json-ld` would ease the automation around listing instances and hosters that offer mastodon.

Gargron:
Since I can't push to the original branch on another repository, re-submitting as new PR.

Fix #6739, close #11628


ClearlyClaire (approved):
Block info would still be useful, but :man_shrugging:

PR #12019, opened 2019-09-30
Remove quotes in `LDAP_SEARCH_FILTER` example

madmath03:
:wrench: Remove quotes in LDAP search filter

Keeping the double quotes in the `.env.production` would generate an invalid search filter exception.


Gargron:
Does it try to print Bash variables because of the double quotes?

Gargron (approved):

madmath03:
@Gargron The exception message was not very explicit, just saying that the search filter syntax was invalid, but I do believe the quotes were printed (docker does not interpret quotes in environment variables).

Thanks for merging this :wink:

Issue #10760, opened 2019-05-13
Remote status isn't indexed into ES well


tribela:
### Expected behaviour

As written at commit 3ebc0ad4d, Remote status that mentioned me is supposed to searchable by me.
```
    * Add full-text search for authorized statuses

    - Search API will return statuses that match the query
    - Only for logged in users
    - Only if you are author of the status,
    - Or you were mentioned in it
    - Or you favourited or reblogged it
    - Configuration over `ES_ENABLED`, `ES_HOST`, `ES_PORT`, `ES_PREFIX`
    - Run `rails chewy:deploy` to create & populate index

```

### Actual behaviour

Remote status is not index to ES server until local user fav/boost it

### Steps to reproduce the problem

1. From user A on server X, Mention user B on server Y with some random message
2. On user B's side, search that random message
3. That status is not searchable (Even if manually queried to ES server directly)
4. After some user on server Y fav or boost that status, it is searchable

### Specifications

On mastodon 2.8.2 stable release

Kjwon15:
Additionally, unfav, unboost isn't update chewy record.

PR #12024, opened 2019-09-30
Fix records not being indexed sometimes

Gargron:
It's possible that after commit callbacks were not firing when exceptions occurred in the process. Also, the default Sidekiq strategy does not push indexing jobs immediately, which is not necessary and could be part of the issue too.


ykzts (approved):

Kjwon15:
Is #10760 not related on this commit?

Gargron:
Maybe, it is difficult to debug because the problem seems to occur only in live environment with concurrency and networking.

PR #12033, opened 2019-10-01
Add error description and button to copy stack trace to web UI

Gargron:
image

### DIFF
--- a/app/javascript/mastodon/components/error_boundary.js
+++ b/app/javascript/mastodon/components/error_boundary.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import illustration from '../../images/elephant_ui_disappointed.svg';
+import { FormattedMessage } from 'react-intl';
+import { version, source_url } from 'mastodon/initial_state';
 
 export default class ErrorBoundary extends React.PureComponent {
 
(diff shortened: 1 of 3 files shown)


Shleeble:
Another PR, another killed mascot.

witcheslive:
Now we all know how the Mastodon went extinct 😭

ykzts (approved):

Gargron:
I find that it's awkward to position the mascot next to the error message, but fine, here is another attempt.

Gargron:
The current error boundary is just the mascot, so I guess we'll just leave it at that. I can't currently position that graphic next to anything in a way that doesn't make me uneasy.

Gargron:
I think ultimately it's more important to say in clear words that the page crashed and what can be done about it than to show a cartoon elephant. The elephant graphic is shaped irregularly which makes it very hard to position next to text in a way that looks clean, and its color scheme is not meant to be next to red error text, as the graphic was originally designed for in-column errors accompanied by greyed out, short phrases.

witcheslive:
Just giving you a hard time mostly, I absolutely do agree that the current behavior is not ideal and somewhat confusing, though it would be nice to have a mascot on there as it is somewhat calming and also fairly common. Even on this fine web sight there's the git cat making joke about the star track: [https://github.com/l;asjdf;'asjf](https://github.com/l;asjdf;'asjf)

I don't know what the possibility or budget is for perhaps doing a new mascot that fits an error page better, but it should be considered probably.

koyuawsmbrtn:
I think this looks very clean and professional. This way normies really know that the site crashed and not that their computer might went bonkers.

PR #12036, opened 2019-10-01
Allow multiple domains for purge

tribela:
This PR allows to do `tootctl domains purge a.com b.tld c.social`
Because tootctl's bootup is too slow, it is better than `for domain in {a.com,b.tld,c.social}; do tootctl domains purge $domain; done`

### DIFF
--- a/lib/mastodon/domains_cli.rb
+++ b/lib/mastodon/domains_cli.rb
@@ -17,7 +17,7 @@ module Mastodon
     option :verbose, type: :boolean, aliases: [:v]
     option :dry_run, type: :boolean
     option :whitelist_mode, type: :boolean
-    desc 'purge [DOMAIN]', 'Remove accounts from a DOMAIN without a trace'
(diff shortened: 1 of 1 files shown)


lib/mastodon/domains_cli.rb
```diff
@@ -27,14 +27,14 @@ def self.exit_on_failure?
...
+    def purge(*domains)
       dry_run = options[:dry_run] ? ' (DRY RUN)' : ''
 
       scope = begin
         if options[:whitelist_mode]
           Account.remote.where.not(domain: DomainAllow.pluck(:domain))
-        elsif domain.present?
-          Account.remote.where(domain: domain)
+        elsif domains.present?
+          Account.remote.where('domain in (?)', domains)
```
Gargron: You don't have to write SQL, if you use `where(domain: domains)` it will do an `IN` query.

Gargron (changes requested):

PR #12038, opened 2019-10-01
Exclude statuses from index that is not searchable by local user

tribela:
Currently, `searchable_by` field contains all related accounts that contains remote user.
But remote user cannot perform a search on local instance.

This PR makes that ES index only contain statuses that searchable by at least one **local** user.
This slightly increase DB query at insert a status, But decrease ES volume massively(because most of remote status doesn't containing a mention to local user)

### DIFF
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -141,7 +141,7 @@ class Status < ApplicationRecord
       ids += preloaded.reblogs[id] || []
     end
 
-    ids.uniq
+    Account.where('id in (?)', ids.uniq).filter(&:local?).map(&:id)
   end
 
   def reply?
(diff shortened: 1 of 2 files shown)


app/models/status.rb
```diff
@@ -141,7 +141,7 @@ def searchable_by(preloaded = nil)
       ids += preloaded.reblogs[id] || []
     end
 
-    ids.uniq
+    Account.where('id in (?)', ids.uniq).filter(&:local?).map(&:id)
```
Gargron: This is an n+1 query which is why the indexing becomes slower. It also loads all account objects into memory before filtering them with Ruby instead of only selecting local accounts with SQL.

Instead, you can change the `crutch` definitions in `chewy/statuses_index.rb` to filter for local accounts in the first place.

For example:

```ruby
data = ::Mention.where(status_id: collection.map(&:id)).where(account: Account.local).pluck(:status_id, :account_id)
```

Gargron (changes requested):

PR #12039, opened 2019-10-01
Fix bad word breaking on conversations

cutls:
#11965 #12022 

> The word break rule affects many common words, which makes them harder to read. That is not optimal...
https://github.com/tootsuite/mastodon/pull/11965#issuecomment-536862447

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -6486,6 +6486,7 @@ noscript {
         }
       }
     }
+    
     a {
       word-break: break-word;
     }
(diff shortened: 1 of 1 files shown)


Gargron:
I am not quite sure if this is fixing anything...

The status component suffers from the same problems: a potentially too long display name at the top, relative timestamp, and potentially unbreakable text. But we made it work, somehow. The conversation component should be no different. This is why I tried copying the `float` approach at first. Maybe I didn't copy enough?

Gargron:
Or maybe I am wrong, the fix seems to work...

Gargron (approved):

PR #12051, opened 2019-10-02
Add parallelization to `tootctl search deploy`

noellabo:
Parallelization makes index generation very fast.


Kjwon15:
On rake job, there are `chewy:deploy` and `chewy:parallel:deploy`.
Parallel is so fast, But sometimes consume CPU very much to service unstable.
How about make it optional like `tootctl search deploy -p=4`?

lib/mastodon/search_cli.rb
```diff
@@ -15,7 +15,7 @@ class SearchCLI < Thor
       changed since the last run.
     LONG_DESC
     def deploy
-      processed = Chewy::RakeHelper.upgrade
+      processed = Chewy::RakeHelper.upgrade parallel: true
       Chewy::RakeHelper.sync(except: processed)
```
Gargron: This method also accepts `parallel` argument

Gargron: I agree with @Kjwon15 that it would make sense to expose the `p` option for the command and then pass it as `parallel` to both of these methods

Gargron: We could default it to 2, and make sure `parallel: false` when `-p=1` (or 0)

noellabo: How about this?

With the --parallel option, Execute tasks in parallel. The default is 2, which specifies the number of processors. If true is specified, it is automatically derived.

noellabo: 'auto' may be more appropriate than 'true'.

Gargron: Agree

lib/mastodon/search_cli.rb
```diff
@@ -6,17 +6,28 @@
...
     def deploy
-      processed = Chewy::RakeHelper.upgrade
+      processed = Chewy::RakeHelper.upgrade parallel: parallel
       Chewy::RakeHelper.sync(except: processed)
     end
+
+    private
+
+    def parallel
+      options[:parallel] == 'auto' ? true : Integer(options[:parallel], exception: false) || 1
```
Gargron: If parallel=1, doesn't that mean it's not parallel at all?

Gargron: Maybe Chewy is smart enough to handle that case though...

lib/mastodon/search_cli.rb
Gargron: `sync` is probably the heaviest part of the task, it also needs `parallel` argument

lib/mastodon/search_cli.rb
noellabo: 1 if an invalid value other than 'auto' or a number is specified.
I think `false` is appropriate, so change it.

noellabo: Chewy may accept nil. Should we remove false?

Gargron: It's probably fine to use false. However, the `Chewy::RakeHelper.sync(except: processed)` method is still not being passed `parallel` param.

Gargron:
Hm, I wonder if `--parallel` should actually be `--processes`, because that's what the number controls through chewy/parallelize.

Gargron (approved):

PR #12053, opened 2019-10-02
✨ Add an LDAP Mail attribute config

madmath03:
:sparkles: Add an LDAP Mail attribute config

This is just a small improvement that allows to set the LDAP mail attribute.
Also, I suggest a default search filter that allows LDAP users to login from their UID or their email address (just like internal users).


Gargron (approved):

Gargron:
Sorry for the delay. Please rebase.

madmath03:
Hi @Gargron,

No problem.
This should be okay now.

Gargron:
The changeset is now very large, I'm worried something else might get reverted by accident if I merge. Could you please do `git reset --hard HEAD~1` to remove the latest commit ("Merge branch..."), `git fetch` to make sure the info is up to date, then `git rebase origin/master`?

madmath03:
My bad.
I did not notice that when I pushed. It should be all good now.

Gargron (approved):

PR #12068, opened 2019-10-03
Remove invite comments from UI

Gargron:
Due to UX confusion and insufficient time to fix it (ref: https://mastodon.social/@Gargron/102899381413079502)


trwnh:
Might be better to rename it to something more explicit instead of removing it entirely. Something like "Describe this invite link (optional)" or add a small subheading to explain what it does. To me it seems clear once you see the table below it has a column for "comment"...

Gargron:
This would require changing strings and thereby delaying the 3.0 release by more days. Besides, the form still doesn't look good, it looks too complicated now, like you're filling out a questionnaire. Fixing that will also take time.

PR #12069, opened 2019-10-03
Change image upload limit from 10 MB / 4096 sqpx to 40 MB / 8192 sqpx

Gargron:
- Since videos already allow 40 MB, the 10 MB are not protecting the backend
- We want to support more uploads from newer generations of mobile devices
- The 16 MB -> 67 MB ImageMagick RAM usage increase is tolerable
- Images are still downsized to 1280 sqpx in the end so this should not affect storage capacity


tateisu:
1px means 3,4,6,or 8 bytes.

PR #12071, opened 2019-10-03
Add basic support for group actors

noellabo:
Add basic support for working with federated group actors.

- Show badge on group actor in WebUI
- Do not notify in case of `announce` by following group actor
- If you mention group actor, also mention group actor followers
- ~~Relax characters that can be used in username (same as Application)~~

A badge is displayed on the group actor so that it can be recognized.

Group actors often deliver posts via `announce`, resulting in excessive notification. Suppress boost notifications from group actors only if you are following.

Mentions to group actors are mentions to group members. We consider group followers as members, and include followers in mentions for proper delivery.

~~It is useful for group actors to be able to reserve names by using `username`, which contains characters that are not allowed for ordinary users. The same is permitted for the application.~~
It's premature, so I'll re-suggest it when needed.


app/helpers/statuses_helper.rb
```diff
@@ -4,6 +4,82 @@ module StatusesHelper
   EMBEDDED_CONTROLLER = 'statuses'
   EMBEDDED_ACTION = 'embed'
 
+  def display_name(account, **options)
```
Gargron: PR possibly needs a rebase? I think these helper functions were previously moved out of this file

app/models/account.rb
```diff
@@ -76,7 +76,7 @@ class Account < ApplicationRecord
   validates :username, format: { with: /\A#{USERNAME_RE}\z/i }, if: -> { !local? && will_save_change_to_username? }
 
   # Local user validations
-  validates :username, format: { with: /\A[a-z0-9_]+\z/i }, length: { maximum: 30 }, if: -> { local? && will_save_change_to_username? && actor_type != 'Application' }
+  validates :username, format: { with: /\A[a-z0-9_]+\z/i }, length: { maximum: 30 }, if: -> { local? && will_save_change_to_username? && !%w(Application Group).include?(actor_type) }
```
Gargron: Maybe better to define `application_or_group?` method instead of doing a raw comparison inside a lambda.

Also this local-only validation change may be too early... It won't help with backwards-compatiblity later (since it's local only anyway) and we don't know how groups will be created/used UX-wise yet. For example, if groups will need to be mentionable, then we would need to conform to the same username regex. If groups will instead have special UI for addressing, then username won't matter.

Gargron (changes requested):

app/helpers/statuses_helper.rb
noellabo: It was a merge mistake. Remove it.

app/models/account.rb
noellabo: This change does not seem to be necessary yet. Let's re-suggest along with other features. Revert it.

Gargron (approved):

PR #12085, opened 2019-10-06
Fix S3 adapter retrying failing uploads with exponential backoff

Gargron:
The default limit of 10 retries with exponential backoff meant
that if the S3 server was timing out, you would be stuck with it
for much, much longer than the 5 second read timeout we expect.

The uploading happens within a database transaction, which means
a failing S3 server could negatively affect database performance

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -40,6 +40,7 @@ if ENV['S3_ENABLED'] == 'true'
       http_open_timeout: 5,
       http_read_timeout: 5,
       http_idle_timeout: 5,
+      retry_limit: 0,
     }
   )
 


ykzts (approved):

smiba:
I know this is an old change, but could this in any way be the reason some people (me included) experience media sometimes being not available (as in #13739 #16520). 

Running `tootctl media refresh` on the posts fixes it, suggesting it's a temporary failure. There doesn't seem to be any attempt from mastodon to retry these posts though, they will stay at "Not Available" forever (until I run said command).

I'll have to dig a bit deeper in how mastodon handles it's media_cache in s3, but it's a recurring issue for me that doesn't seem to be logged either.

smiba:
I want to confirm that setting the retry_limit to `2` has severely reduced my image failures to basically 0%.
I've yet to see any, tootctl media refresh barely does anything nowadays :)

What to do with this? Should we slightly increase the retry_limit until we have developed a proper solution for S3 upload failures? @Gargron

PR #12117, opened 2019-10-08
Remove auto-silence behaviour from spam check

### LINKED ISSUE #12113: Improved anti-spam behavior
### Pitch

I think it's better to change the anti-spam feature so that it doesn't automatically silence the person.

~~I think it is better to divide it into an optional setting to silence automatically if possible (default off).~~
EDIT: I'm sorry, I think we shouldn't do this. I think that if some servers malfunction, the overall impression of Mastodon will deteriorate.

### Motivation

Because this feature performs wrong BAN action in the following cases:

- People who repeat mentions using messages formatted for greetings
- People who repeatedly sends the same content to the same person (for example, a joke)

I think it is important to introduce a function to combat spam. I agree with it.

But even if it's spam, I think we should avoid doing BAN actions for their account automatically.
I think it should be handled manually after the admin or moderator receives the report.

Gargron:
Fix #12113

### DIFF
--- a/app/lib/spam_check.rb
+++ b/app/lib/spam_check.rb
@@ -44,7 +44,6 @@ class SpamCheck
   end
 
   def flag!
-    auto_silence_account!
     auto_report_status!
   end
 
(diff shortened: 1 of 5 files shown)


ThibG:
It will keep reporting spammers over and over, as `already_flagged?` just checks for silenced status

ykzts (approved):

PR #12122, opened 2019-10-08
Add admin setting to auto-approve hashtags

Gargron:
And change some no longer accurate labels on admin settings

- "Timeline preview": No longer actually a "preview" embedded on the landing page, it is now a separate page. However, *also*, this setting has been expanded to controlling whether the `GET /api/v1/timelines/public` is accessible without authentication (but I don't think the setting affects the *streaming* API access to the public timeline... Nor does it affect the hashtag timeline API)
- "Show known fediverse on timeline preview": Same as above about the "timeline preview" concept, likewise, "known fediverse" is no longer what the federated timeline is called so it's better to just say "federated content". This setting also has no affect on whether the API can only return local content or not, and has no effect on the streaming API, it is purely cosmetical.

### DIFF
--- a/app/models/form/admin_settings.rb
+++ b/app/models/form/admin_settings.rb
@@ -30,6 +30,7 @@ class Form::AdminSettings
     mascot
     spam_check_enabled
     trends
+    trendable_by_default
     show_domain_blocks
     show_domain_blocks_rationale
     noindex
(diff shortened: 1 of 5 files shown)


ykzts (approved):
nice 👍

PR #12123, opened 2019-10-09
Fix S3 errors preventing processing of ActivityPub data

Gargron:
Instead of letting S3 networking errors propagate to job execution and forcing the jobs to fail and retry, catch them locally. For media attachments, a separate job will be queued to retry; custom
emoji will be skipped (but this will allow a future status containing them to redownload them). Avatars and headers will be skipped.

Change the behaviour of remotable concern. Previously, it would skip downloading an attachment if the stored remote URL is identical to the new one. Now it would not be skipped if the attachment is not actually currently stored by Paperclip.


Gargron:
Please wait... If S3 is not enabled then Seahorse is undefined

Gargron:
I have added a fallback definition

PR #12124, opened 2019-10-09
Remove `lang` attribute from individual statuses

Gargron:
Fix #10930


ykzts (approved):

marrus-sh:
I understand why this is necessary for CJK rendering, but it may also confound screenreaders, to the detriment of bilingual users which depend on them. (Screenreaders depend on the `lang` tag in order to know… what language to speak the text in.) It might be a good idea to add a user setting to re­‑enable this?

Issue #9582, opened 2018-12-19
Display square (and up to 1:1 ratio) medias without being cropped


nclm:
### Pitch

Currently, media attachments (both images and videos) that are higher than a ratio of what seems to be 16:9 get either cropped (images) or scaled down and letterboxed (videos) when shown in the interface. I think that maybe media up to 1:1 should be allowed to be displayed at their native sizes.

It could work like this:
1. Medias lower than 16:9 behave however they work now (I haven’t tried it, are they kept at their ratio or cropped to 16:9?).
2. Medias that are 16:9 obviously stay at 16:9.
3. Medias inbetween 16:9 and 1:1 (included) would now keep their ratio.
4. Medias higher than 1:1 do get cropped/scaled down to 16:9 exactly like they would now. I fully agree that we need at one point to limit the vertical space a toot can take.

Alternatively for number 3, instead of allowing all the variations, a set of common ratios (4:3, 1:1…) could be set, and the media  would be cropped/scaled to the closest.

### Motivation

1:1 ratio is quite common on social media, being used for graphics, illustrations — and also photography (see Instagram). Other social networks allows it so it might be expected by users that Mastodon would too. Additionally, the upcoming federation of Pixelfed might show the arrival of regular square pictures on the Fediverse, so for followers from Mastodon, not having to click on each pictures would be a more ideal experience.

4:3 can also appear from time to time, especially for screenshots and extracts from pre-mid-2000s media (TV shows, computer screens). In the landscape of social networks, Dribbble is notable for having it as its only ratio.

Gargron:
The reason its always 16:9 in timelines is I want visual consistency between different posts. It also prevents tall images from taking up too much space.

nclm:
Yes that I understand! The idea here is that when someone post something that is 1:1 or 4:3, it is often I think quite on purpose and expected to displayed in full. And they are both relatively common ratios to be maybe treated first class.

Everything else can get cropped down do 16:9 — I do entirely agree on limiting the tallness of toots, plus 1:1 is not _that_ tall. I also agree that consistency is important, but it can be “broken” from time to time at the off-chance someone post something precisely this size. I think the user experience would be improved by seeing these images at their size.

[...]

PR #12126, opened 2019-10-09
Add setting for whether to crop images in unexpanded toots

### LINKED ISSUE #9916: Change Thumbnail strategy to show whole image
### Expected behaviour

Posts containing an image should show the whole image content without needing to show the single article view. This is how other social media platforms present such content.

### Actual behaviour

The image is cropped and only the middle part is visible.
mastodon cropped thumb
screenshot_2019-01-24 norden social

Scroll trough timeline and look for image posts like memes, infographs, ... that usually need an full height to become easy to read.

### Specifications
FF 64
Mastodon 2.7

duxovni:
Fixes #9916 and #9582. Users can now select in the "Appearance" preferences menu whether they want to keep the old cropping behavior (the default) or show all images uncropped.

### DIFF
--- a/app/controllers/settings/preferences_controller.rb
+++ b/app/controllers/settings/preferences_controller.rb
@@ -57,6 +57,7 @@ class Settings::PreferencesController < Settings::BaseController
       :setting_use_blurhash,
       :setting_use_pending_items,
       :setting_trends,
+      :setting_crop_images,
       notification_emails: %i(follow follow_request reblog favourite mention digest report pending_account trending_tag),
       interactions: %i(must_be_follower must_be_following must_be_following_dm)
     )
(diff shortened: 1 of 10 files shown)


Gargron (approved):

JamesABDL:
Deployed this to my instance, I notice it only works on posts with a single image. Is that intended? Posts with two or more images are still cropped.

duxovni:
Yes, it's deliberate that this only applies to posts with a single image; it's less clear what the correct behavior should be for posts with multiple images. I guess we could display them all uncropped in a single column?

JamesABDL:
Ah okay, well the description in preferences ("Crop **images** in non-expanded toots to 16x9") implies that it will apply to all images in a post. I know if I told my users about this feature they would assume it was buggy.

I think there should be an option for both behaviours. Uncropped in a single column wouldn't look great, but some users / instances want that.

marians:
Better late then never: I just came across this setting in Mastodon and find that it improves my experience significantly. Thanks for this contribution!

PR #12137, opened 2019-10-10
Fix preloaded JSON-LD contexts not being used

Gargron:
Regression from #11316

Related: ruby-rdf/json-ld-preloaded#4


ykzts (approved):

Gargron:
Closing in favour of #12138

Issue #9638, opened 2018-12-26
unliking a post does not update like count


connyduck:
### Expected behaviour

when I unlike a post, the like count is updated as it is when liking a post

### Actual behaviour

The like count stays the same. Refreshing the page resolves the problem.
Does not happen everytime, but almost every time.

![video](https://user-images.githubusercontent.com/10157047/50457380-54269780-095b-11e9-97ac-01af69ff1956.gif)

### Steps to reproduce the problem

open the detail view of a toot, like it if it isn't already, unlike it

### Specifications

Happens on all instances I have accounts on, e.g. mastodon.social which is on 2.6.5 
Happens with both Tusky & web interface

chuckyz:
I got curious so I looked it up; it looks like this gets queued up for later deletion in a redis key.

https://github.com/tootsuite/mastodon/blob/9311430ed7022e2e02fbd0b453adfc43356fdf9d/app/lib/activitypub/activity.rb#L120

Can you check some ~6 hours after unliking to see if the count goes down, or perhaps look under redis, specifically the (**do not run this in a production instance**) `KEYS delete_upon_arrival:*` command via redis-cli?

nightpool:
@chuckx that's a completely different part of the code, and totally unrelated to this bug report, which is only about the front end (since it fixes itself after refreshing)

trwnh:
cannot reproduce in latest versions

PR #12140, opened 2019-10-10
Unliking a post updates like count on front end

NimaBoscarino:
Fixes the issue outlined by #9638.

Previous behaviour:

New behaviour:


Gargron:
I'm sorry, but there is a reason why an asynchronous worker is used. Yes, your code returns the real state, but does so by running a lot of tasks (including networking) within the request/response cycle. If connection drops in the middle, the unfavourite operation could miss some instructions. It would be far easier to update the favourites count in the Redux reducer on the client-side.

NimaBoscarino:
Okay, I did it through the reducer on the client-side! Does this look better?

app/javascript/mastodon/reducers/statuses.js
```diff
@@ -37,6 +38,9 @@ export default function statuses(state = initialState, action) {
     return importStatuses(state, action.statuses);
   case FAVOURITE_REQUEST:
     return state.setIn([action.status.get('id'), 'favourited'], true);
+  case UNFAVOURITE_SUCCESS:
+    const favouritesCount = action.status.get('favourites_count');
+    return state.setIn([action.status.get('id'), 'favourites_count'], favouritesCount - 1);
```
Gargron: A more concise way of writing this is:

```js
return state.updateIn([action.status.get('id'), 'favourites_count'], x => Math.max(x - 1, 0));
```

Gargron (approved):

PR #12154, opened 2019-10-12
Add ability to filter reports by target account domain

ClearlyClaire:
image
image


config/locales/en.yml
```diff
@@ -401,6 +401,7 @@ en:
       are_you_sure: Are you sure?
       assign_to_self: Assign to me
       assigned: Assigned moderator
+      by_target_domain: Domain of report target
```
Gargron: Instead, use "Domain of reported account"

Gargron (changes requested):

Gargron (approved):

PR #12155, opened 2019-10-12
Change misleading metrics in admin instance view dashboard

ClearlyClaire:
“Followed by them” was the number of follow relationships from them to local
users. Changed it to be the number of local users followed by at least one
of their users.

“Followed by us” was the number of follow relationships from local users
to them. Changed it to be the number of their users followed by at least one
of our local users.

“Blocked by us” was the number of user blocks from local users targetting
one of their users. Changed it to be the number of their users blocked by
at least one local user.

Also, change “… by us” strings to make it explicit that this is “by local users”, in particular for blocks where it could mislead admins into thinking it's about people they suspended.

## Before

image

## After

image

### DIFF
--- a/app/controllers/admin/instances_controller.rb
+++ b/app/controllers/admin/instances_controller.rb
@@ -15,10 +15,10 @@ module Admin
     def show
       authorize :instance, :show?
 
-      @following_count = Follow.where(account: Account.where(domain: params[:id])).count
-      @followers_count = Follow.where(target_account: Account.where(domain: params[:id])).count
+      @following_count = Follow.where(account: Account.where(domain: params[:id])).distinct.count(:target_account_id)
+      @followers_count = Follow.where(target_account: Account.where(domain: params[:id])).distinct.count(:target_account_id)
       @reports_count   = Report.where(target_account: Account.where(domain: params[:id])).count
-      @blocks_count    = Block.where(target_account: Account.where(domain: params[:id])).count
+      @blocks_count    = Block.where(target_account: Account.where(domain: params[:id])).distinct.count(:target_account_id)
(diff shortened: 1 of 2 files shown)


nightpool:
I feel like i would love to have both sets of numbers available, the four of 1) local users following 2) remote users following 3) local users followed and 4) remote users followed, and also the pair of 1) local users blocking 2) remote users blocked

ThibG:
I think all those metrics would indeed be useful, but I was aiming at keeping the same number of metrics to start with.

ThibG:
Maybe something like “Blocked by X local users”

trwnh:
There's also value in knowing the total relations and not just total users. Consider an instance with 1 user that has 100,000 connections to your instance's users. Yes, it's 1 user, but they're a well-connected user.

The harder question is how to present that info without it being too dense to be useless...

Gargron:
The labels may not be worded well but I think the original numbers are valuable. It answers the question "How many followers will my users lose if I block this domain"

ThibG:
Even then, I'm not sure how useful that is, it doesn't give the max number of lost followers, nor an average or something, it's not really super easy to reason about. Anyway, the label is misleading.

ThibG:
What about keeping the current metrics but renaming the labels, then?
- Follows from remote users to local users
- Follows from local users to remote users
- Number of blocks from local users to remote users (this sounds more awkward than with follows, tbh)

ykzts (approved):

trwnh:
How about:

- "Incoming follow relations"
- "Outgoing follow relations"
- "Outgoing block activities"

ThibG:
The first two suggestions seem good, less sure about “outgoing block activities” (especially since I want to give the option *not to* have outgoing block activities…)

trwnh:
Oh, right. Maybe just "total blocks"?

Tagadda:
This seems good to merge.

Issue #4038, opened 2017-07-02
Swipe feature on mobile version too sensitive


ThomasLeister:
Hi,

thanks to @sorin-davidoi for the new swipe feature in Mastodon 1.4.7! I like the idea. Unfortunately this feature currently drives me crazy because I constantly change the column view unintentionally when scrolling. :-( In my opinion the trigger for a view change is way too sensitive. 

Maybe there is a way to make it less sensitive? A view change should only be triggered on long-distance swipes, but not on tiny horizontal movements. 

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [X] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

Gargron:
Duplicate of #4035, resolved

Issue #4035, opened 2017-07-02
Too easy to trigger column swiping


sorin-davidoi:
I find myself accidentally swiping columns when scrolling up / down. `react-swipeable` has a delta which configures how long the swipe has to be before it is considered a swipe. The default is 10px.

This issue is mostly to see if this affects other people. We can increase the `delta` if this is the case.

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- [ ] This bug happens on a [tagged release](https://github.com/tootsuite/mastodon/releases) and not on `master` (If you're a user, don't worry about this).

kyzh:
I have an issue that sounds exactly like what you describe here.
It is really frustrating to say the least and I'm glad this could be the answer to it.

When I was trying to put words on what I was experiencing, I found 2 main behaviours:
- going laterally column wise 
- the change to happen at content/column level before the top icon.

I actively tried to swipe and the column content change before the icon at the top show the focus.
I therefore think my issue and yours is the same.

Speaking for myself, I'd love to be able to disable swiping all together as it is slower and less practical than clicking the icon at the top.
It take longer than clicking the icon at top (by an order of magnitude I'd say)

sorin-davidoi:
We are working on making swiping columns faster - it is indeed annoying that the content and the icon at the top are not in sync for ~1 second. Thanks for commenting, will make a pull request to increase the delta - 25 / 30 pixels should do it.

sorin-davidoi:
Regarding disabling it, this was a highly requested feature, we just need to make sure it is not triggered by accident.

nightpool:
I think the bigger issue here is the lack of visual feedback. It's impossible to tell when you MIGHT be triggering a swipe, so it ends up coming out of nowhere when it does happen.

jalefkowit:
I'm not sure why this issue has been closed; I still run into accidental column changes via swipe all the time on multiple devices (iPad Air 2 in Safari, Galaxy S6 in Firefox). Despite my going out of my way to try and keep my scrolling gestures perfectly vertical, it happens frequently enough to make the mobile web version of Mastodon unusable.

nightpool:
@jalefkowit These problems were fixed in the upcoming version of Mastodon, which isn't out to most instances yet. What version is your instance using?

jalefkowit:
@nightpool I'm on octodon.social, which as of this writing is running 1.4.7.

PR #12168, opened 2019-10-15
Improve swipe experience

umonaca:
Currently the swipe is too sensitive on mobile devices. Related issues: #4038 and #4035.
The `react-swipeable-views` changed a lot since 2017 so the previous commit no longer works.

Changes:
`threshold={15}` sets the speed threshold to be three times less sensitive than default to be identified as a quick-swipe. The default value is 5, which is too sensitive according users' feedback.
`hysteresis={0.2}` sets the delta threshold to be 20% of width to swipe for a slow swipe. By default this value is 0.6, which means users have to swipe 60% of the width to switch to another page during slow-swipes. The default value does not make sense.

I have tested these values and saw no side-effect.

Reference:
[1] `react-swipeable-views` API document: https://react-swipeable-views.com/api/api/


Gargron (approved):

umonaca:
Well, GitHub shows waiting for CI status forever...

umonaca:
Well, GitHub shows waiting for CI status forever...

PR #12170, opened 2019-10-15
Add Sidekiq Prometheus Metrics Endpoint

mhamzahkhan:
There are probably better ways of doing this, and I was hoping if there is, people can advise how.

I am exposing metrics on /sidekiq_metrics, and using nginx to control who can access the page.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -155,3 +155,5 @@ end
 
 gem 'concurrent-ruby', require: false
 gem 'connection_pool', require: false
+
+gem 'sidekiq-prometheus-exporter', '~> 0.1.12'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -585,6 +585,8 @@ GEM
       redis (>= 3.3.5, < 5)
     sidekiq-bulk (0.2.0)
       sidekiq
+    sidekiq-prometheus-exporter (0.1.12)
+      sidekiq (>= 3.3.1)
     sidekiq-scheduler (3.0.0)
       redis (>= 3, < 5)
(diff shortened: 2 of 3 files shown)


Gargron:
I am afraid that exposing internal metrics at an endpoint that has to be explicitly hidden by modifying nginx configuration is not acceptable as a lot of admins will miss the memo. Also, since the desire for Prometheus metrics will likely not stop at Sidekiq defining a Sidekiq-only metrics endpoint is not future-proof.

ykzts:
I'm using [statsd_exporter](https://github.com/prometheus/statsd_exporter). Does this not solve your request?

mhamzahkhan:
Can statsd export sidekiq metrics?
Also, I'm not sure how it can be used when running mastodon in docker

ykzts:
> Can statsd export sidekiq metrics?

Yes! It looks like this:

screenshot on Grafana

> Also, I'm not sure how it can be used when running mastodon in docker

If the UDP host and port provided by statsd_exporter are given to the Mastodon container in the form of the environment variable STATSD_ADDR, it works correctly.

travnewmatic:
> If the UDP host and port provided by statsd_exporter are given to the Mastodon container in the form of the environment variable STATSD_ADDR, it works correctly.

statsd-exporter can listen listen on TCP and UDP.  Perhaps it would be nice to mention that mastodon is going to attempt to communicate via UDP?

spent a while troubleshooting until i determined that I had my statsd-exporter listening for the wrong protocol :)

PR #12172, opened 2019-10-15
Drop filtered messages if the filter is “irreversible”

ClearlyClaire:
… instead of adding them to the timelines and then not showing them.
This fixes timelines showing new items when the only new items are “irreversibly” filtered toots, as well as such items taking space in the client-side lists.

This is actually an edge case in Mastodon itself, but the normal case in Pleroma, which can be used with a Mastodon front-end but does not apply filters server-side.

### DIFF
--- a/app/javascript/mastodon/actions/importer/normalizer.js
+++ b/app/javascript/mastodon/actions/importer/normalizer.js
@@ -10,6 +10,12 @@ const makeEmojiMap = record => record.emojis.reduce((obj, emoji) => {
   return obj;
 }, {});
 
+export function searchTextFromRawStatus (status) {
+  const spoilerText   = status.spoiler_text || '';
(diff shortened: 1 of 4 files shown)


Gargron:
> This is actually an edge case in Mastodon itself

If I'm not mistaken it *never* happens in Mastodon..?

ThibG:
Thinking a bit more about it, it probably never occurs in the `updateTimeline` case in Mastodon, but can occur in `expandTimeline` (for the same reason we have those filters client-side in the first place: filters not applying retroactively)

So the usefulness is pretty limited in Mastodon itself, as it only prevents TL lists to fill up with old dropped toots, the only user-facing consequence of that being old toots cleaned up “too early” (the 40 items thing) if space in-between was occupied by dropped toots.

Issue #12076, opened 2019-10-04
Download button on audio player (possibly video player as well)


jackwilliambell:
### Pitch

I would like to have a download button on audio player to allow people to download the audio file. This can be automatic (for every file) or by selecting an option when or after the file is uploaded.

### Motivation

I am planning to upload some music I created to my Mastodon instance with a Creative Commons license. It's nice that people can play the audio and share the link, but I would like to allow them to easily download the file as well so they can add it to their music library. I believe other audio file creators would feel the same way, although some (for example podcasters) might prefer to not have downloads.

### Current State

The uploaded file is available in the HTML for the audio player, so it appears the audio player is not using some kind of re-encoding scheme. Since that is the case it wouldn't increase the server load and a scraping program (or a savvy user) can already get the audio file name and download it the hard way.

NimaBoscarino:
This sounds like a great idea to me! In case there’s interest, I’ve made a PR for it.

ryliejamesthomas:
I'm also really keen on it. Works fine with the default audio player, so weird to see it absent here.

NimaBoscarino:
This has been added, so is this issue good to close?

ykzts:
Resolve with #12179

PR #12179, opened 2019-10-17
Add download button to audio and video players

NimaBoscarino:
Implements the following: #12076 

The download button is an \<a> tag with the "download" property, and styled similarly to the other existing buttons.

Download button for audio player

Download button for video player

### DIFF
--- a/app/javascript/mastodon/features/audio/index.js
+++ b/app/javascript/mastodon/features/audio/index.js
@@ -219,7 +219,7 @@ class Audio extends React.PureComponent {
               </span>
             </div>
 
-            <div className='video-player__buttons'>
+            <div className='video-player__buttons right'>
               <button type='button' aria-label={intl.formatMessage(messages.download)}>
                 <a className='video-player__download__icon' href={this.props.src} download>
                   <Icon id={'download'} fixedWidth />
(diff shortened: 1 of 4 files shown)


Cassolotl:
Does this come with controls for the creator to be able to turn it off?

NimaBoscarino:
Not at the moment, but I can definitely add that. I'd imagine you'd get to choose that when you upload your file?

Cassolotl:
Cool, that sounds good!

marrus-sh:
Mastodon doesn't implement DRM, so there is nothing to prevent people from downloading any file. If you want to stop people from downloading a file, the best option is to post it on SoundCloud or a video site which implements DRM.

Given that “whether to show that a file is downloadable” is something which would need an extension to ActivityPub in order to federate, I'm not sure it's worth disambiguating on a per-toot basis? And it's especially strange since there is no such restriction on images (any user can download any image at any time). Why would audio and video be different?

Gargron:
I agree with @marrus-sh. All files are already one click away from download. It’s even one of the selling points compared to competitors that put video behind obfuscation.

NimaBoscarino:
That's a good point! I guess for now it's enough to have the download button by default. Do you think it would be useful to have the download button for images as well?

Gargron:
In the audio screenshot, it looks like the button has a lot of unnecessary margin to the right?

NimaBoscarino:
Good catch! That should be fixed now.

Gargron (approved):

PR #12185, opened 2019-10-18
Fix incoming federation in whitelist mode

ClearlyClaire:
… posting to the AP inbox required a logged-in local user…


Gargron:
I think I'm continually blindsided by the inheritance of the inboxes controller from the API base controller :confused:

ThibG:
Well, in this case it's not because of the API base controller, but because of the `AccountOwnedConcern`

Gargron (approved):

PR #12189, opened 2019-10-20
microformat mentions can have an implicit url property

nightpool:
See the first example here: http://microformats.org/wiki/microformats2#hyperlinked_person

> Simple hyperlinked person reference
> `<a class="h-card" href="http://benward.me">Ben Ward</a>`
> 
> Parsed JSON:
> ```json
> {
>   "items": [{ 
>     "type": ["h-card"],
>     "properties": {
>       "name": ["Ben Ward"],
>       "url": ["http://benward.me"]
>    }
>   }]
> }

see http://microformats.org/wiki/microformats2-parsing#parsing_for_implied_properties for a more detailed list of implicit property parsing rules.

We might also want to change the HTML mastodon generates at some point, since having an extra span element feels awkward to me.

### DIFF
--- a/app/services/fetch_link_card_service.rb
+++ b/app/services/fetch_link_card_service.rb
@@ -84,7 +84,7 @@ class FetchLinkCardService < BaseService
 
   def skip_link?(a)
     # Avoid links for hashtags and mentions (microformats)
-    a['rel']&.include?('tag') || a['class']&.include?('u-url') || mention_link?(a)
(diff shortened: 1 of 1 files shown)


ykzts (approved):

Gargron:
There is an extra span for a better underline effect on hover that does not include the at sign, we won’t remove that

ClearlyClaire (approved):
Approved. But I think we could also skip URLs for hashtags and mentions listed in the ActivityStreams `tag` attribute.

PR #12193, opened 2019-10-21
Fix the pangram

ClearlyClaire:


trwnh:
Technically it's already a valid pangram. You only need to use the word "the" once.

ThibG:
Oh you're actually right >_<
I found it weird because Crowdin suggested a translation for the more usual variant, and I didn't actually check those letters were already used.

Nevermind then.

Issue #455, opened 2017-01-10
No notifications in notifications tab for follow requests


nikkisquared:
If your account is private, and somebody sends you a follow request, there is no notification for this happening in the notifications tab. Having emails turned on for follow requests works fine, but it's strange for follow requests to not give a notification in Mastodon, especially when follows on a non-private account do.

Gargron:
This has been a deliberate choice, at the time welcomed by users. The idea is that if your account is locked, you don't want to be bothered, and having follow requests in a place where you can look when *you* want it, rather than being demanded to look at it, is better.

If there is a significant movement for changing this, it would be simple enough to do, since currently follow requests are *deliberately* filtered out of the notifications column.

nightpool:
maybe make this opt-in? So follow requests are by default filtered out (client-side, using the existing notif filter stuff) but can be enabled?

nightpool:
I'm not entirely sold on the idea that this is the best default though, so maybe it should just be opt-out.

Cassolotl:
How about a little number next to "Follow Requests" in the Getting Started menu? The same colour as the rest of the text, just a little number in brackets - so you only see it when your fourth column is empty, and it's fairly discreet but you can see it without having to click anything?

![screen_shot_2017-07-07_at_12_03_19](https://user-images.githubusercontent.com/6607512/27955363-62a802ea-630c-11e7-934f-676f4a6c0132.png)

wxcafe:
it should *deffo* be opt-out imo and it really needs to exist yeah

rakiru:
I completely missed that follow requests were filtered out from notifications. I considered this surprising behaviour, since everything else appears to show a notification, including regular follows, and smaller actions such as favouriting toots, as well as there being toggles to filter each of these things out. I also find it inconvenient as I have to check it manually every so often, whereas every other notification I get live in the pinned notifications column.

I find most of Mastodon UI intuitive, but this is one of a few things that I only stumbled on by chance and never would have guessed.

Cassolotl:
Yeah that's a good point - if users want to be able to not see follow requests it can be a notification, and then in the notifications toggle menu people should be able to toggle them off.

dariusk:
[...]

PR #12198, opened 2019-10-22
Add follow_request notification type

ClearlyClaire:
The notification type already existed in the backend but was never pushed
to the front-end. This also means translation strings were also available
for the backend, from the notification mailer.

Unlike other notification types, these are off by default, to match what
I remember of Gargron's view on the topic: that follow requests should not
clutter notifications and should instead be reviewed at the user's own
leisure in the dedicated column.

Since follow requests have their own column, I've deemed it unnecessary to
add a specific tab for them in the notification quick filter.

## Settings

image

## Notification in column

image

I am not completely sure about this design tbh

### DIFF
--- a/app/controllers/api/v1/push/subscriptions_controller.rb
+++ b/app/controllers/api/v1/push/subscriptions_controller.rb
@@ -51,6 +51,6 @@ class Api::V1::Push::SubscriptionsController < Api::BaseController
 
   def data_params
     return {} if params[:data].blank?
-    params.require(:data).permit(alerts: [:follow, :favourite, :reblog, :mention, :poll])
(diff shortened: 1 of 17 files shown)


ThibG:
The notification could possibly be changed to show Accept/Reject buttons

Gargron:
I suggest using the typical `<Account />` component (or copying it) with the added benefit of the avatar/name being aligned with all the others, and then having a checkmark/times icons in place of what usually is the "follow" button on that component.

ThibG:
Updated accordingly:
image

Gargron (approved):

Phroneris:
Is this fixing #455 ?

ThibG:
Yes! Unlike with other notification types, it isn't enabled by default though

Issue #6053, opened 2017-12-16
Use inert polyfill


sorin-davidoi:
Since https://github.com/tootsuite/mastodon/pull/4405/ we are using the `inert` attribute for accessibility reasons in the modals. However, this is currently only supported in Google Chrome.

There are two polyfills which we can use:
 - [The official one](https://github.com/WICG/inert) from WICG - may incur a performance penalty
 - [The Google one](https://github.com/GoogleChrome/inert-polyfill) - aims to be more lightweight but not 100% correct

I would prefer the official one but we need to look into the performance penalty.

SuperSandro2000:
#12209 closed this issue.

PR #12209, opened 2019-10-25
Use inert polyfill

gabrielrumiranda:
this is a PR for #6053 issue


Gargron:
> codeclimate — 113 issues to fix 

It seems you are using a completely different code style.

gabrielrumiranda:
sorry @Gargron i forgot to config my text editor

package-lock.json
```diff
@@ -0,0 +1,12 @@
+{
```
Gargron: What is the purpose of this file? We use yarn.lock

Gargron (approved):

PR #12249, opened 2019-10-28
Fallback to Create audience when object has no defined audience

### LINKED ISSUE #11137: Support objects without to/cc
Some hubzilla objects (in particular, images) federate over without an explicit audience. However, by AP spec, we should only consider the audience on the Create activity for delivery. This does leave the visibility of the object itself a little uncertain, and it makes it harder to decide on the visibility for objects that get referenced from other activities without a Create. (boosts, etc)

<details><summary>Original issue description</summary>
Support of the ActivityStreams "Create=>Image" activity

### Pitch

In order to support sharing images creation, Mastodon need to implement this ActivityPub part and understand ActivityStreams "Create=>Image" activity.

### Motivation

People shares images creation and don't understand why this doesn't shows up on Mastodon.
</details>

ClearlyClaire:
Fixes #11137

I'm not sure of possible side-effects.

### DIFF
--- a/app/lib/activitypub/activity/create.rb
+++ b/app/lib/activitypub/activity/create.rb
@@ -25,6 +25,14 @@ class ActivityPub::Activity::Create < ActivityPub::Activity
 
   private
 
+  def audience_to
+    @object['to'] || @json['to']
+  end
+
+  def audience_cc
+    @object['cc'] || @json['cc']
+  end
+
   def process_status
     @tags     = []
     @mentions = []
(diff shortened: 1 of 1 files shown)


Gargron (approved):

Gargron:
This looks okay but I do not understand the semantics of having the audience on the activity. So I send the "Create" to one audience but a "Delete" to another?? It makes more sense on "Announce" which, frankly, starts looking more and more like an object rather than activity, but that's what we're working with.

ThibG:
I am not sure why one would send a Create Activity and the corresponding Delete to different audiences, but that's entirely possible, yes. The ActivityPub spec itself only talks about *objects* having an audience in C2S when they are used as a shorthand for a `Create`. Otherwise, ActivityPub doesn't specify what to do with objects' audiences.

PR #12259, opened 2019-10-30
make it not return http 400 when passing and empty source argument

jglauche:
I described my issue here:
https://github.com/tootsuite/mastodon/issues/12257


jglauche:
I've added an rspec. Got confused at first at it as the empty hash gets ignored if you don't add as: :json 
This added rspec fails before my patch and succeeds with my patch.

app/controllers/api/v1/accounts/credentials_controller.rb
```diff
@@ -26,6 +26,7 @@ def account_params
 
   def user_settings_params
     return nil unless params.key?(:source)
```
Gargron: This line can be removed because of the new one

Gargron (changes requested):

app/controllers/api/v1/accounts/credentials_controller.rb
jglauche: params[:source] will be nil if the parameter is not supplied by the requester, checking it #empty? on it will result in 
nil.empty? which will cause it to error.

Gargron: `nil.empty?` is `true`

Gargron:
```ruby
return nil if params[:source].nil? || params[:source].empty?
```

Gargron: Another option is to use `blank?` instead of `empty?` because it works on nil and hash, then only one check is needed.

jglauche:
I compacted it; I personally prefer to check for both nil and empty? as nil.blank? is a rails specific thing. 
Please note if you want me to use blank? instead.

jglauche:
alright, the code quality bot wants me to use .blank? instead. There you go.

Gargron (approved):

PR #12262, opened 2019-10-30
Fix remote media descriptions being cut off at 420 chars

### LINKED ISSUE #12258: Image Alt Text is Cut at 420 Characters when Federated
### Expected behaviour
Federated versions of a toot should include the full image alt text of the original toot.

### Actual behaviour
Federated versions of a toot include only the first 420 (nice) characters, truncating anything beyond.

1. Create a toot with an image
1. Set the alt text for that image to something longer than 420 characters
1. View that toot from another instance

### Specifications

Screenshot showing full image alt text on original toot:
A screenshot of a toot about Bernie Sanders with a text screenshot as the image. The full text of the screenshot is shown in the alt text of the image as shown by the developer tools window to the right of the page with the HTML image tag highlighted.

Screenshot showing boosted toot's truncated image alt text on a user profile page.
A screenshot of a toot as viewed from the public profile page of a user from another instance who boosted the toot- the alt text is truncated at 420 characters.

**Testing Details**
We were able to reproduce the issue between the following server versions:

Originating Version | Federated Version
------------ | -------------
Mastodon 3.0.1 | Mastodon 3.0.1
Mastodon 3.0.1 | v3.0.1+glitch+vulpine-prod-20191017-02
v3.0.1+glitch+vulpine-prod-20191017-02 | Mastodon 3.0.1

We tested and verified the issue in multiple browsers:
Firefox, Waterfox, Chrome and Opera.

And on multiple (API based) clients:
Tusky, Pinafore

ClearlyClaire:
Fixes #12258

### DIFF
--- a/spec/models/media_attachment_spec.rb
+++ b/spec/models/media_attachment_spec.rb
@@ -136,10 +136,10 @@ RSpec.describe MediaAttachment, type: :model do
   end
 
   describe 'descriptions for remote attachments' do
-    it 'are cut off at 140 characters' do
+    it 'are cut off at 1500 characters' do
       media = Fabricate(:media_attachment, description: 'foo' * 1000, remote_url: 'http://example.com/blah.jpg')
 
-      expect(media.description.size).to be <= 420
+      expect(media.description.size).to be <= 1_500
     end
   end
 end
--- a/app/models/media_attachment.rb
+++ b/app/models/media_attachment.rb
@@ -26,6 +26,8 @@ class MediaAttachment < ApplicationRecord
 
   enum type: [:image, :gifv, :video, :unknown, :audio]
 
+  MAX_DESCRIPTION_LENGTH = 1_500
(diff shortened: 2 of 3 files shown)


redtoaster:
Looks like another spec needs to be updated: 

https://github.com/tootsuite/mastodon/blob/master/spec/models/media_attachment_spec.rb#L138

ThibG:
indeed, thanks!

spec/models/media_attachment_spec.rb
```diff
@@ -139,7 +139,7 @@
     it 'are cut off at 140 characters' do
```
redtoaster: Might update the test's description too.

larry-perry (approved):
Looks good.  One nitpick that I wouldn't blame you for ignoring.

Gargron (approved):

PR #12266, opened 2019-10-30
Fix filtered out items being counted as pending items

ClearlyClaire:


ThibG:
Note that it doesn't prevent the “active” state from turning on (blue header text and glow). This would require changing the reducer itself, which is more awkward. It's probably worth fixing anyway, but it's a different, much older bug.

Gargron (approved):

PR #12283, opened 2019-11-02
Add `tootctl media lookup` command

thehonkinggoose:
This pull request is in reference to https://github.com/tootsuite/mastodon/issues/12282.

**Thought process for my implementation:**
We take a media url from the user and remove the `http://` (also need to edit this to edit `https://` which is more common) part and then split by the slash. We access from 1, onwards (not sure about Ruby's array syntax and will have to test, `[1, -1]`? Once we have a path to the file, we lookup from the MediaAttachment model using `file_file_name:` and pass a String with the file path we came up with. From there we create a url with the source toot.

**Initial snippet**
```ruby
desc 'lookup', 'Lookup a toot by passing a media URL'
    def lookup
      say("Please enter a URL to the media to lookup:")
      media_url = gets.chomp.gsub("http://", "").split("/")[1]
      attachment = MediaAttachment.where(file_file_name: media_url)
      mastodon_hostname=ENV['LOCAL_DOMAIN']
      status_id = attachment.status_id
      say("The source toot is: https://#{mastodon_hostname}/web/statuses/#{status_id}")
    end
```


nsuchy:
This should pass the checks and be ready for a maintainer to review and provide feedback 😄

lib/mastodon/media_cli.rb
```diff
@@ -113,5 +113,24 @@ def usage
...
+
+    desc 'lookup', 'Lookup a toot by passing a media URL'
+    def lookup
+      prompt = TTY::Prompt.new
+
+      exit(1) unless (url = prompt.ask('Please enter a URL to the media to lookup:', required: true))
+
+      id = url
+           .split('/')[0..-2]
+           .select do |key|
```
nightpool: reject { matches? } is clearer to read then select ... scan ... empty.

lib/mastodon/media_cli.rb
```diff
@@ -113,5 +113,24 @@ def usage
...
+             key.scan(/\D/).empty?
```
nightpool: personally I've always found the \d / \D thing a little hard to keep track of, and besides that I think it unnecessarily inverts the logic here. what about just using a pattern like `\A\d+\Z`?

lib/mastodon/media_cli.rb
```diff
@@ -113,5 +113,24 @@ def usage
       say("Imports:\t#{number_to_human_size(Import.sum(:data_file_size))}")
       say("Settings:\t#{number_to_human_size(SiteUpload.sum(:file_file_size))}")
     end
+
+    desc 'lookup', 'Lookup a toot by passing a media URL'
+    def lookup
+      prompt = TTY::Prompt.new
+
+      exit(1) unless (url = prompt.ask('Please enter a URL to the media to lookup:', required: true))
```
nightpool: what's the point of the exit(1) here? you'll never get a nonempty string back from `required: true`

lib/mastodon/media_cli.rb
```diff
@@ -113,5 +113,24 @@ def usage
...
+
+      id = url
+           .split('/')[0..-2]
```
nightpool: do we know the media url is always mounted on root? is there a way to configure s3 with a prefix?

lib/mastodon/media_cli.rb
```diff
@@ -113,5 +113,24 @@ def usage
...
+           .select do |key|
+             key.scan(/\D/).empty?
+           end
+           .join('')
+
+      attachment = MediaAttachment.find(id)
+
+      prompt.say('The corresponding media object you referenced could not be found, perhaps the toot was deleted?') unless attachment
```
nightpool: these two lines should be an if/else statement, there's no reason to display the url if the attachment doesn't exist

nightpool (dismissed):

lib/mastodon/media_cli.rb
nsuchy: We don't know it's mounted on root. This check is folder agnostic. It decides the URL into parts and pulls the parts with numbers only to determine an attachment ID. To the best of my knowledge whether it's mounted to root of the hostname or not does not affect this check.

lib/mastodon/media_cli.rb
nsuchy: You're right that's unnecessary, modifying my code to account for that.

lib/mastodon/media_cli.rb
nsuchy: Will update things.

lib/mastodon/media_cli.rb
nsuchy: Could you show me an example of how this might work in code? I'm not sure I follow how to do this with reject.

nsuchy:
One other thing I should work out is the different between a Toot and an Avatar Media Object, passing an avatar won't work since there is no status id associated with it. So I'll need to determine whether it's an avatar or not before outputting a response.

nsuchy:
> One other thing I should work out is the different between a Toot and an Avatar Media Object, passing an avatar won't work since there is no status id associated with it. So I'll need to determine whether it's an avatar or not before outputting a response.

This issue has been resolved with recent commits.

Gargron (approved):

PR #12302, opened 2019-11-04
Simplify SQL query when performing account search amongst followings

ClearlyClaire:


Gargron (commented):
Isn’t this going to search only among one’s own followers, rather than followed people?

ThibG:
I don't think so? The left outer join is just to compute the score: it's here in the query not restricted to followers too, and its goal is to bump the score along follow relationships. But when searching people you follow, *all* of them are by definition followed by you, so it's not useful to match those follow relations, they'd bump the score the same way for every account.

Gargron (approved):

PR #12390, opened 2019-11-15
LDAP & PAM added to OAuth password grant strategy

### LINKED ISSUE #7999: OAuth fails with LDAP Accounts
First of all, thanks for all your good work.

The current Mastodon fails to provide OAuth tokens for LDAP accounts via POST to /oauth/token:

```
# Local account works

$ cat local-account.json
{"password": "pass...", "client_secret": "cs...", "grant_type": "password", "redirect_uri": "urn:ietf:wg:oauth:2.0:oob", "client_id": "ci...", "scope": "read write follow push", "username": "mail@example.com"}

$ curl -X POST https://example.com/oauth/token -d @local-account.json --header "Content-Type: application/json"
{"access_token":"ac...","token_type":"bearer","scope":"read write follow push","created_at":1531335851}

# LDAP account does not work, tried account, account@example.com and even email address of that account. Triple checked password...

$ cat ldap-account.json
{"password": "pass...", "client_secret": "cs...", "grant_type": "password", "redirect_uri": "urn:ietf:wg:oauth:2.0:oob", "client_id": "ci...", "scope": "read write follow push", "username": "account"}

$ curl -X POST https://example.com/oauth/token -d @ldap-account.json --header "Content-Type: application/json"
{"error":"invalid_grant","error_description":"The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client."}

```

However, OAuth generally works with LDAP accounts.
Verified with: https://takahashim.github.io/mastodon-access-token/

Thanks in advance.

ntl-purism:
Resolves #7999

When authenticating via OAuth, the resource owner password grant
strategy is allowed by Mastodon, but (without this PR), it does not
attempt to authenticate against LDAP or PAM. As a result, LDAP or PAM
authenticated users cannot sign in to Mastodon with their
email/password credentials via OAuth (for instance, for native/mobile
app users).

This PR fleshes out the authentication strategy supplied to doorkeeper
in its initializer by looking up the user with LDAP and/or PAM when
devise is configured to use LDAP/PAM backends. It attempts to follow the
same logic as the Auth::SessionsController for handling email/password
credentials.

Note #1: Since this pull request affects an initializer, it's unclear
how to add test automation.

Note #2: The PAM authentication path has not been manually tested. It
was added for completeness sake, and it is hoped that it can be manually
tested before merging.

### DIFF
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -8,8 +8,20 @@ Doorkeeper.configure do
   end
 
   resource_owner_from_credentials do |_routes|
-    user = User.find_by(email: request.params[:username])
-    user if !user&.otp_required_for_login? && user&.valid_password?(request.params[:password])
+    if Devise.ldap_authentication
+      user = User.authenticate_with_ldap({ :email => request.params[:username], :password => request.params[:password] })
+    end
+
+    if Devise.pam_authentication
+      user ||= User.authenticate_with_ldap({ :email => request.params[:username], :password => request.params[:password] })
+    end
+
+    if user.nil?
+      user = User.find_by(email: request.params[:username])
+      user = nil unless user.valid_password?(request.params[:password])
+    end
+
+    user if !user&.otp_required_for_login?
   end
 
   # If you want to restrict access to the web interface for adding oauth authorized applications, you need to declare the block below.


ntl:
Yay! The tests pass now!

Gargron (approved):

PR #12391, opened 2019-11-15
Add Vietnamese translation files

taicv:


Gargron:
Hello, can you import these files on https://crowdin.com/project/mastodon ? If I import them here, they won't be picked up by Crowdin.

taicv:
> Hello, can you import these files on https://crowdin.com/project/mastodon ? If I import them here, they won't be picked up by Crowdin.

Hi Gargron,
How could I import in Crowdin? I have account there already but don't know how.
I think we have to create Vietnamese translation first then I import files after.

ykzts:
@taicv https://crowdin.com/project/mastodon/vi

PR #12428, opened 2019-11-19
Encrypt direct messages with Tanker

dmerejkowsky:
Hello, I've been using Mastodon since April 2017. One thing that always bugs me is that Mastodon administrators have access to direct messages between users.

I work at [Tanker.io](https://tanker.io) where we develop an end-to-end encryption solution and I recently decided to try and use it inside Mastodon. I wrote [a series on dev.to](https://dev.to/tanker/taking-mastodon-security-to-the-next-level-part-1-encrypt-your-toots-2p00) about it that you can read for more information.

This pull request is just a proof-of-concept and should not be merged as is, but it can be a good starting point for further discussion.

Feel free to browse the changes, comment on the patch or ask questions.

Cheers!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,11 @@ gem 'cld3', '~> 3.2.4'
 gem 'devise', '~> 4.7'
 gem 'devise-two-factor', '~> 3.1'
 
+# Warning: need to install libsodium for this gem to work
+# More info:
+# https://github.com/crypto-rb/rbnacl/wiki/Installing-libsodium
+gem 'tanker-identity', git: 'https://github.com/TankerHQ/identity-ruby'
+
 group :pam_authentication, optional: true do
   gem 'devise_pam_authenticatable2', '~> 9.2'
 end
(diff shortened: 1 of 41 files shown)


Gargron:
We were just discussing our plans for a DMs overhaul in the develepment Discord. However, I must say straight away: We cannot use a 3rd party service. If we go with E2EE DMs, we will need to build our own key exchange. Tanker could close shop or be compromised in the future and requiring external sign-ups/confirmations is also not viable from a UX perspective.

dmerejkowsky:
> Requiring external sign-ups/confirmations is also not viable for us from a UX perspective

Indeed. External sign-up and identity confirmation are required when using Tanker, but it's possible to do it without impacting the UX that much. This was not done in this patch because it is just a proof of concept.

> We were just discussing our plans for a DMs overhaul in the development Discord

Great! I've left a message there so we can continue the conversation there.

> We cannot use a 3rd party service [and] we will need to build our own key exchange

Understood. For information, we've documented how we did it in a [public specification](https://github.com/tankerhq/spec), and even if the server remains close source for now, 
note that source code for all the clients SDK is available on GitHub as well.

Cheers!

PR #12457, opened 2019-11-23
Avoid creating invalid username for instance account

zunda:
If database is set up before setting `LOCAL_DOMAIN`, the instance account may be created with an invalid username:

```
> SELECT username FROM accounts WHERE id=-99;
    username
-----------------
 localhost:45208
(1 row)
```

This is to avoid such username with removing the port number part from Rails' local domain.

### DIFF
--- a/db/migrate/20190715164535_add_instance_actor.rb
+++ b/db/migrate/20190715164535_add_instance_actor.rb
@@ -1,6 +1,6 @@
 class AddInstanceActor < ActiveRecord::Migration[5.2]
   def up
-    Account.create!(id: -99, actor_type: 'Application', locked: true, username: Rails.configuration.x.local_domain)
+    Account.create!(id: -99, actor_type: 'Application', locked: true, username: Rails.configuration.x.local_domain.gsub(/:\d*/, ''))
   end
 
   def down
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,6 +1,6 @@
 Doorkeeper::Application.create!(name: 'Web', superapp: true, redirect_uri: Doorkeeper.configuration.native_redirect_uri, scopes: 'read write follow push')
 
-domain = ENV['LOCAL_DOMAIN'] || Rails.configuration.x.local_domain
+domain = (ENV['LOCAL_DOMAIN'] || Rails.configuration.x.local_domain).gsub(/:\d*/, '')
 account = Account.find_or_initialize_by(id: -99, actor_type: 'Application', locked: true, username: domain)
 account.save!
 


Gargron:
Wouldn't "localhost" by itself also present a problem in a production environment?

zunda:
As far as I see, `username` being `localhost` doesn't present a problem because the instance account is always found by its `id`, i.e., with the code `Account.find(-99)`.

Gargron:
Maybe we could make the migration (and seeds.rb) raise an error if LOCAL_DOMAIN environment is not present.

zunda:
Another option here. Might it be possible to hard-code the `username` to something like `instance_actor`? This maybe better than making migrations fail because it pauses less prerequisites.

As I wrote above, the `username` column for the instance actor is not used internally. I've just checked that the actor as seen from outside shows the `username` as `preferredUsername` property. Currently, it seems to be stored in the `username` column but, again, doesn't seem to be referred to.

From a local production installation:

```
> SELECT username, domain, actor_type FROM accounts WHERE id=-99;
    username    | domain | actor_type
----------------+--------+-------------
 localhost:3000 |        | Application
(1 row)
```

The request `curl -s http://localhost:3000/actor#main-key | jq .` shows:

```json
{
  "@context": [
    {
      "manuallyApprovesFollowers": "as:manuallyApprovesFollowers",
      "toot": "http://joinmastodon.org/ns#",
      "featured": {
        "@id": "toot:featured",
        "@type": "@id"
      },
      "alsoKnownAs": {
        "@id": "as:alsoKnownAs",
        "@type": "@id"
      },
      "movedTo": {
        "@id": "as:movedTo",
        "@type": "@id"
      },
      "schema": "http://schema.org#",
      "PropertyValue": "schema:PropertyValue",
      "value": "schema:value",
      "IdentityProof": "toot:IdentityProof",
      "discoverable": "toot:discoverable"
    }
  ],
  "id": "http://localhost:3000/actor",
  "type": "Application",
  "inbox": "http://localhost:3000/actor/inbox",
  "preferredUsername": "localhost:3000",
  "url": "http://localhost:3000/about/more?instance_actor=true",
  "manuallyApprovesFollowers": true,
  "publicKey": {
    "id": "http://localhost:3000/actor#main-key",
    "owner": "http://localhost:3000/actor",
    "publicKeyPem": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAsfdnbo3c+AB9hMHyA5aZ\n0j/FyYXJNL0kPnHVHszA//72IRvlSaBuAiNK6Iwy/FhF6ZccryWOz5i756DjNNqk\nnzFoc541gSM67Ec4GqBjEb3JUXweNlG73/be0eK9pev/2uTmV4h7ihwotpGxmAvB\nJHVB+1sdWwZexIUhbtzdGoIJUIMGY9Huu8+H5qMhD9sI8fcCfC5/BRvxDnxLCeAc\njZNtNg7wHZ3NzSFmF/rB4ZT1DeHGDTNn846O1PS2613GJ2aVtHe6U2tc7Unwuu8e\nFvKwhWuGGzRVPE8ZUyypxA2sMN45dydy7I3ezPPrxRmkNW7G3SmBM5wE0lmUA84W\nzwIDAQAB\n-----END PUBLIC KEY-----\n"
  },
  "endpoints": {
    "sharedInbox": "http://localhost:3000/inbox"
  }
}
```

My server already has some instance accounts with `username` different from `domain`:

```
> SELECT username, domain FROM accounts WHERE actor_type='Application' LIMIT 10;
          username           |           domain
-----------------------------+-----------------------------
 mstdn.kemono-friends.info   | mstdn.kemono-friends.info
 thefreestate.xyz            | thefreestate.xyz
 internal.fetch              | carrot.army
 internal.fetch              | pleroma.mobi
 internal.fetch              | ia.cafe
 pachyder.me                 | pachyder.me
 mastodon.wearefamily.africa | mastodon.wearefamily.africa
 mastodon.otherreality.net   | mastodon.otherreality.net
 internal.fetch              | pleroma.travnewmatic.com
 internal.fetch              | jpop.club
(10 rows)
```

zunda:
I'm closing this for now. This can be worked around with setting `LOCAL_DOMAIN` environment variable before running `rake db:migrate`.

PR #12461, opened 2019-11-24
:sparkles: Convert LDAP username

### LINKED ISSUE #12021: [FEATURE] Convert LDAP username to valid format
### Pitch

When enabling LDAP auth, if my LDAP UID is considered invalid by Mastodon (cf. #12020), I would like Mastodon to (at least try to) convert my UID to something valid.

For instance, if my LDAP UID is `firstname.lastname`, I would like Mastodon to replace invalid chars by an underscore and generate the username `firstname_lastname`.

This feature could be enabled or disabled with an environment variable:
* `LDAP_UID_CONVERSION_ENABLED=true`

We could also add another environment variable to configure the list or regex of characters to replace and maybe the replacement character too:
* `LDAP_UID_CONVERSION_SEARCH=., -`
* `LDAP_UID_CONVERSION_REPLACE=_`

### Motivation

This would help a few people:
* LDAP admins would not need to create a new entry in users' LDAP info just to store a _valid_ mastodon username
* LDAP users could easily connect to Mastodon with their same username as any other services

madmath03:
:sparkles: Convert LDAP username to fix #12021


app/models/concerns/ldap_authenticable.rb
```diff
@@ -14,10 +14,19 @@ def authenticate_with_ldap(params = {})
     end
 
     def ldap_get_user(attributes = {})
-      resource = joins(:account).find_by(accounts: { username: attributes[Devise.ldap_uid.to_sym].first })
+      safe_username = attributes[Devise.ldap_uid.to_sym].first
+      # tootsuite/mastodon#12021: Convert LDAP username to valid format
+      if (Devise.ldap_uid_conversion_enabled)
```
Gargron: The paranthesis is unnecessary

.env.nanobox
```diff
@@ -183,6 +183,10 @@ SMTP_FROM_ADDRESS=notifications@${APP_NAME}.nanoapp.io
 # LDAP_BIND_DN=
 # LDAP_PASSWORD=
 # LDAP_UID=cn
+# tootsuite/mastodon#12021: Convert LDAP username to valid format
```
Gargron: I would remove the repeating comments referencing the PR, nothing else does that

app/models/concerns/ldap_authenticable.rb
madmath03: Parenthesis removed

.env.nanobox
madmath03: Removed all comments referencing the PR

Gargron (approved):

PR #12480, opened 2019-11-25
Allow muted users to appear in lists

### LINKED ISSUE #10346: Muted Users Continue To Appear In List
### Pitch
When user "A" mutes user "B", user A will not see user B in Home/Federated Timeline but can continue to view user's B content in the List Panel.

This can (once implemented) be done by muting user A then adding user B onto user's A list, where user's B content can be viewed by user A

### Motivation

Anyone who's interested in an user whom posts frequently but does not wish to see it "bloat" their Home/Federated timeline

This is my case with @monero@framapiaf.org who posts frequently, but I do not find joy seeing my Timelines filled by their content, instead I would rather mute @monero@framapiaf.org from my Timelines, then add @monero@framapiaf.org to my "Cryptocurrencies" List.

deansheather:
If you add a user to a list and mute them their posts will continue to show in the list. Previously, if you muted an account and put them in a list their posts would not appear in the list.

This is helpful to allow users you don't want in your home feed to have their own separate feed. I use this on my instance to have a "Bots" list containing muted bot accounts so they don't appear in my home feed.

Adds a feed manager filter function specifically for lists which does the same thing as the home filter minus the "is muted" check. Since the home filter and list filter are mostly the same, the home filter just checks for blocks/mutes and calls the list filter.

Closes #10346

### DIFF
--- a/app/lib/feed_manager.rb
+++ b/app/lib/feed_manager.rb
@@ -186,7 +186,7 @@ class FeedManager
 
     return true if check_for_blocks.any? { |target_account_id| crutches[:blocking][target_account_id] || crutches[:muting][target_account_id] }
 
-    return filter_from_list?(status, receiver_id, crutches)
+    filter_from_list?(status, receiver_id, crutches)
   end
 
   def filter_from_list?(status, receiver_id, crutches)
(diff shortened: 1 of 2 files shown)


ThibG:
Another approach that has been implemented in at least one fork is to have “exclusive” lists of accounts that won't appear in the Home timeline. I think this may be a better solution, since you don't want muted or blocked people to appear anyway in lists via boosts.

PR #12486, opened 2019-11-26
Change audio/video playback to stop playback when out of view

Gargron:
- Change video player to not loop, since the audio player doesn't
- Change playback and mute buttons to feel snappier

### DIFF
--- a/app/javascript/mastodon/features/audio/index.js
+++ b/app/javascript/mastodon/features/audio/index.js
@@ -37,15 +37,14 @@ class Audio extends React.PureComponent {
     volume: 0.5,
   };
 
-  // hard coded in components.scss
-  // any way to get ::before values programatically?
-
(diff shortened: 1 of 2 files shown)


ThibG:
I personally would oppose to this for two main reasons:
- actually, I prefer media to keep playing when scrolling, especially when it's mostly audio and I'm in a thread view. I hate what Twitter does in that regard
- this would be inconsistent with media from preview cards, which we can't control… (well, we *could* unload them completely, but I'd find it even more annoying)

Gargron:
The intersection observer or the reducer do unload the media when it's scrolled away, it's just not immediate. I feel like this way is cleaner, makes it a more consistent behaviour than "keep scrolling and at some point it'll just stop playback, or maybe not until new toots come in from top"

ykzts (approved):

PR #12489, opened 2019-11-27
Use override_csp options on pghero

### LINKED ISSUE #8966: PgHero Connections graphs broken on master
When testing from master today (at 9d4541c) it seems the graphs on the PgHero / Connections page (ex: https://mastodon.social/pghero/connections) are broken:

image

This only started happening recently, somewhere between 2.5.0 and now. I confirmed the same regression on a second instance. Graphs load on 2.5.0, but don't when running from master.

It doesn't seem very impactful since only admins can access pghero.

* * * *

ykzts:
Resolve #8966

see https://github.com/ankane/pghero/commit/3d9428e8d79f59c381e93bab56f5c75ee94ca1f4

### DIFF
--- a/config/pghero.yml
+++ b/config/pghero.yml
@@ -1,0 +1,41 @@
+databases:
+  primary:
+    # Database URL (defaults to app database)
+    # url: <%= ENV["DATABASE_URL"] %>
+
+  # Add more databases
+  # other:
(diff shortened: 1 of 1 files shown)


Gargron (approved):

ThibG:
This breaks PgHero when assets are served from a different domain.

PR #12502, opened 2019-11-30
Split relationships page strings

### LINKED ISSUE #10863: Create separate strings for "Follows and followers" section
Currently (since first introduced) "Follows and Followers" section in settings uses strings from main interface, which is bad for translations (these strings should differ).
For example:

image
image

In both cases there should be "Followers" string with "many" quantifier (currently it uses quantifier based on the actual count, looks ugly)
I guess the same goes for other strings in this section.

brawaru:
Before this moment relationships managing page was using strings from other context — from counters. But in order for translators to be able to translate it relatively to the page (as a tab there), it must use separate strings.

| English | Russian |
|:-------:|:-------:|
| Screenshot_20191130_131252 | Screenshot_20191130_131316 |
| Strings shown normally (*kinda*) | Strings irrelevantly “pluralized” |

I've split the strings for "Following" and "Followers" and put them to "relationships" keyset in localization file. This should solve this issue.

Fixes #10863

### DIFF
--- a/app/views/relationships/show.html.haml
+++ b/app/views/relationships/show.html.haml
@@ -8,8 +8,8 @@
   .filter-subset
     %strong= t 'relationships.relationship'
     %ul
-      %li= filter_link_to t('accounts.following', count: current_account.following_count), relationship: nil
-      %li= filter_link_to t('accounts.followers', count: current_account.followers_count), relationship: 'followed_by'
+      %li= filter_link_to t('relationships.following'), relationship: nil
+      %li= filter_link_to t('relationships.followers'), relationship: 'followed_by'
       %li= filter_link_to t('relationships.mutual'), relationship: 'mutual'
 
   .filter-subset
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -924,6 +924,8 @@ en:
   relationships:
     activity: Account activity
     dormant: Dormant
+    followers: Followers
+    following: Following
     last_active: Last active
     most_recent: Most recent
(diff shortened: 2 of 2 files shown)


app/views/relationships/show.html.haml
```diff
@@ -8,8 +8,8 @@
   .filter-subset
     %strong= t 'relationships.relationship'
     %ul
-      %li= filter_link_to t('accounts.following', count: current_account.following_count), relationship: nil
-      %li= filter_link_to t('accounts.followers', count: current_account.followers_count), relationship: 'followed_by'
```
Gargron: Maybe we could just access `accounts.followers.other` directly? Just thinking about reducing translation overhead

Sasha-Sorokin: Unfortunately no, they are similar in English, but may differ in other languages (they do in Russian actually). So I think it's the best solution for us to have separate strings based on context, that's what issue #10863 was exactly in the first place.

---

Also, `accounts.following` seems to be incorrect, because in original context it is used as plural, but in translation file it is a string:

```haml
-# .../accounts/_header.html.haml
%span.counter-number= number_to_human account.following_count, strip_insignificant_zeros: true
%span.counter-label= t('accounts.following', count: account.following_count)
```

```yml
# en.yml
    followers:
      one: Follower
      other: Followers
    following: Following
#   ^^^^^^^^^
```

Want me to fix that in this or another PR?

Gargron (approved):

PR #12508, opened 2019-12-01
Return `discoverable` via REST API (fix #12507)

### LINKED ISSUE #12507: `discoverable` not returned via API on Account entity
`discoverable` flag for inclusion in the profile directory was introduced in 2.7.0. `/api/v1/accounts/update_credentials` allows setting this parameter, but the parameter is not included in any responses. I tested the following:

- `/api/v1/accounts/update_credentials` does not return `discoverable`
- `/api/v1/accounts/verify_credentials` does not return `discoverable`
- `/api/v1/accounts/:id` does not return `discoverable`

...so I'm not sure where a client may discover whether the user has currently opted into the directory simply from knowing the account.

trwnh:
I think this should fix it?

https://github.com/tootsuite/mastodon/blob/master/app/serializers/rest/credential_account_serializer.rb is another place that it could be returned instead, but I think in the long-term it is valuable to know whether an account is discoverable even if it is not your own account.

### DIFF
--- a/app/serializers/rest/account_serializer.rb
+++ b/app/serializers/rest/account_serializer.rb
@@ -3,7 +3,7 @@
 class REST::AccountSerializer < ActiveModel::Serializer
   include RoutingHelper
 
-  attributes :id, :username, :acct, :display_name, :locked, :bot, :created_at,
+  attributes :id, :username, :acct, :display_name, :locked, :bot, :discoverable, :created_at,
              :note, :url, :avatar, :avatar_static, :header, :header_static,
              :followers_count, :following_count, :statuses_count, :last_status_at
 


Gargron:
Maybe this also resolves #7822?

trwnh:
I would be fine if we said toot:discoverable was also intended as an indexing flag, yes

ThibG:
As you said in your last comment on that PR, those seem to be different concerns…

Gargron:
Discoverable is a better indicator though because it federates. The noindex setting is local-only so it would never be correct for remote accounts in the API. Its use case is also so narrow that it's not worth federating, while "discoverable" could be used for multiple things that stem from one desire to be discovered or to not be discovered.

trwnh:
if you consider the profile directory as a sort of index, then opting out of indexing would imply opting out of the directory, and vice versa would be similar enough to the point that maybe two different properties aren't needed.

Gargron (approved):

PR #12537, opened 2019-12-02
Fix batch actions on non-pending tags in admin interface

ClearlyClaire:
Currently, the Hashtag admin page shows batch action checkboxes on every page but only actually show action buttons on the “pending review” tab. This is confusing and unnecessary, and frustrating when one checks every tag then figures out they can't act on them.

This PR changes that to:
- Allow batch actions on unreviewed tags
- Hide checkboxes when batch actions aren't available (“all” and “reviewed” tabs)

I wonder about that last part: should we allow to override the state even for reviewed tags? If so we could just show the batch actions in all cases.

image


ThibG:
oops, I had inadvertently added a temporary file. Removed it and rebased the PR

ykzts (approved):

PR #12538, opened 2019-12-02
Fix poll options not being selectable via keyboard

### LINKED ISSUE #12384: Accessibility issues with polls that are single-option
### Expected behaviour

Polls that have radio buttons should be accessible via the keyboard and with screen readers.

### Actual behaviour

Polls expose the options only as text, not as radio buttons. Keyboard focus  cannot land on them, and the keyboard cannot be used to interact with them. Screen readers do not know which option they have selected.

With NVDA and Firefox on Windows, or VoiceOver and Safari on the Mac:
1. Find a toot that has a poll in it where you can only select one option out of several. Or create one for testing.
2. Try to interact with it using the tab and arrow keys. Focus will land on the section that is the container for the question and list of answers, but arrow keys won't do anything. NVDA or VoiceOver also won't read anything. If exploring the answers, they'll be exposed as text, not as radio buttons, and the selected state of one, the one with the check mark, will not be communicated.

#### Analysis

The inner HTML for an option looks like this:

```
<label class="poll__text selectable"><input name="vote-options" type="radio" value="2"><span class="poll__input"></span><span>NOTA ; see my comment</span></label>
```

There are several problems here:

1. The actual input isn't visible, it is not rendered in the accessibility tree. It is probably due to some CSS rule marking it as display:none; so the span next to it can be used for styling instead.
2. The span that displays the selection does not communicate this fact semantically.
3. Because the input is visually hidden, it can also not accept keyboard focus or any other keyboard input.

Two possible solutions:

1. Make the input properly visible, and interactable.
2. Implement the [Radio Group design pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton) for these elements, using roles, states, and JS for keyboard input to make these actionable and properly exposed to screen readers.

Both have their amount of work, the first if styling is the issue, two because it needs to implement a lot from scratch that the native input would otherwise provide natively.

### Specifications

Mastodon both 3.01 and latest commit running on mastodon.social.

Firefox + NVDA on Windows, Chrome + NVDA on Windows, Safari or Chrome on MacOS with VoiceOver.

ClearlyClaire:
Fixes #12384

Also allow switching from single-option to multiple-option poll with keyboard and add a tooltip explaining that.

Changes the styling of hovered/focused/active checkboxes/radios as following:
Peek 02-12-2019 18-56

### DIFF
--- a/app/javascript/mastodon/components/poll.js
+++ b/app/javascript/mastodon/components/poll.js
@@ -87,9 +87,9 @@ class Poll extends ImmutablePureComponent {
     this._toggleOption(value);
   };
 
-  handleOptionKeyPress = (optionIndex, e) => {
+  handleOptionKeyPress = (e) => {
     if (e.key === 'Enter' || e.key === ' ') {
-      this._toggleOption(optionIndex);
+      this._toggleOption(e.target.getAttribute('data-index'));
       e.stopPropagation();
       e.preventDefault();
     }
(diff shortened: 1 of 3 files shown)


Gargron (approved):

ThibG:
Fixed some coding style issues, but I'll keep the `.bind` if there's no good reason for doing otherwise: this seems far simpler this way

ThibG:
Nevermind, updated it to avoid that, since it can have negative performance implications.

Gargron (approved):

PR #12434, opened 2019-11-20
upgrade/replace websocket.js to @gamestdio/websocket

hinaloe:
## Summary
[websocket.js](https://www.npmjs.com/package/websocket.js) has deprecated, and that renamed to [@gamestdio/websocket](https://www.npmjs.com/package/@gamestdio/websocket).

This replace to new one.

## Note
default backoff has [replaced](https://github.com/gamestdio/websocket/commit/717a90b13c86bea55e7e315829e05c0dd5416934#diff-1fdf421c05c1140f6d71444ea2b27638R16) fibonacci to exponential.

PR #12542, opened 2019-12-04
Updated node/ruby version requirements

shleeable:
Hey, nodejs v8 and Ruby 2.4 should be considered EOL. So I've bumped them up to more recent releases.

Node v13 / Ruby 2.7 will be stable soon but are untested.

This has the benefit of allowing blocked PRs #12521 #11727 #12434 etc


Shleeble:
lol. Forgot to delete the CI.

Shleeble:
Ready to merge.

Shleeble:
Please review @ThibG @trwnh @ykzts @Gargron.

README.md
```diff
@@ -65,12 +65,12 @@ Mastodon acts as an OAuth2 provider so 3rd party apps can use the REST and Strea
...
+**Requirements: (Recommended)**
 
-- **PostgreSQL** 9.5+
-- **Redis**
-- **Ruby** 2.4+
-- **Node.js** 8+
+- **PostgreSQL** 9.5+ (10.6+)
+- **Redis** 4+ (5+)
+- **Ruby** 2.5+ (2.6+)
+- **Node.js** 10.13+ (12.13+)
```
ThibG: Is there a specific reason for those recommendations?

ClearlyClaire (commented):
Seems fine, but I'm not sure there is any rush yet. Node 8's EOL is at the end of the month, but Ruby 2.4's EOL is in a few more months.

I see no strong reason to reject this change, but it can probably wait after the next Mastodon release (assuming there's one shortly, which is what I'm advocating for)

README.md
Shleeble: Ruby 2.7 will be out in a month or so - That'll likely be around the time of the EOL for the others too.

and the recommendations are just the versions used in Docker - so they're tested and slightly more bleeding edge.

Shleeble:
FYI Dropping Ruby 2.4 unblocks #12584 
"sprockets-4.0.0 requires ruby version >= 2.5.0"

README.md
ThibG: ok, recommending the versions used in the Docker image seem reasonable to me (but maybe that reason should be explicitly written there as well), I'm just not sure it's worth the burden of having to maintain that information when such versions change

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@tootsuite/mastodon",
   "license": "AGPL-3.0-or-later",
   "engines": {
-    "node": ">=8.12 <13"
+    "node": ">=10.13 <13"
```
abcang: The reason I set the lower limit of node version to 8.12 is to match the lower limit version of the package being used. I think it is better to set it to 10 if there is no special reason to specify the minor version.

Shleeble: Similar reasons for mine. #12521 is 10.13 for that package.

abcang: ok, I understand.

README.md
SuperSandro2000: I think it should be written out, too. It is quite confusing if you look at it the first time.

Shleeble:
I'll rejig this in the next few weeks.

PR #12566, opened 2019-12-07
Admin setting to disable default follows

ashpieboop:
I wanted to disable auto follows on my instance, so I implemented a setting.

### DIFF
--- a/app/services/bootstrap_timeline_service.rb
+++ b/app/services/bootstrap_timeline_service.rb
@@ -5,7 +5,7 @@ class BootstrapTimelineService < BaseService
     @source_account = source_account
 
     autofollow_inviter!
-    autofollow_bootstrap_timeline_accounts!
+    autofollow_bootstrap_timeline_accounts! if Setting.enable_bootstrap_timeline_accounts
   end
 
   private
--- a/app/javascript/packs/admin.js
+++ b/app/javascript/packs/admin.js
@@ -45,7 +45,25 @@ const onDomainBlockSeverityChange = (target) => {
 
 delegate(document, '#domain_block_severity', 'change', ({ target }) => onDomainBlockSeverityChange(target));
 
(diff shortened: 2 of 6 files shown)


app/services/bootstrap_timeline_service.rb
```diff
@@ -28,7 +28,13 @@ def autofollow_bootstrap_timeline_accounts!
   def bootstrap_timeline_accounts
     return @bootstrap_timeline_accounts if defined?(@bootstrap_timeline_accounts)
 
-    @bootstrap_timeline_accounts = bootstrap_timeline_accounts_usernames.empty? ? admin_accounts : local_unlocked_accounts(bootstrap_timeline_accounts_usernames)
+    accounts = []
+
+    if Setting.enable_bootstrap_timeline_accounts || Setting.enable_bootstrap_timeline_accounts.nil?
```
ArisuOngaku: Please particularly pay attention to this line; I tried to make it enabled by default, is this good?

ArisuOngaku:
Ok I think circleci isn't triggering because I actually set it up on the PR's repo, hold on I'm creating another one

app/services/bootstrap_timeline_service.rb
ThibG: Default values are set in `config/settings.yml`. Furthermore, I wouldn't edit the `bootstrap_timeline_accounts` method, but simply prevent `autofollow_bootstrap_timeline_accounts!` from running unless `Setting.enable_bootstrap_timeline_accounts` is true.

ClearlyClaire (changes requested):

ArisuOngaku:
@ThibG thanks for your review and for teaching me about config/settings.yml. I think you're right, this is a way better solution and I should have looked this through.

ClearlyClaire (approved):

PR #12568, opened 2019-12-08
Add `tootctl media remove-orphans`

Gargron:
> Scans file storage for files that do not belong to existing media attachments. Because this operation
> requires iterating over every single file individually, it will be slow.
>
> Please mind that some storage providers charge for the necessary API requests to list objects.

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -89,7 +89,7 @@ else
   Paperclip::Attachment.default_options.merge!(
     storage: :filesystem,
     use_timestamp: true,
-    path: ENV.fetch('PAPERCLIP_ROOT_PATH', ':rails_root/public/system') + '/:class/:attachment/:id_partition/:style/:filename',
+    path: File.join(ENV.fetch('PAPERCLIP_ROOT_PATH', File.join(':rails_root', 'public', 'system')), ':class', ':attachment', ':id_partition', ':style', ':filename'),
     url: ENV.fetch('PAPERCLIP_ROOT_URL', '/system') + '/:class/:attachment/:id_partition/:style/:filename',
   )
 end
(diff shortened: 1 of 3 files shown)


config/initializers/paperclip.rb
```diff
@@ -89,7 +89,7 @@
   Paperclip::Attachment.default_options.merge!(
     storage: :filesystem,
     use_timestamp: true,
-    path: ENV.fetch('PAPERCLIP_ROOT_PATH', ':rails_root/public/system') + '/:class/:attachment/:id_partition/:style/:filename',
+    path: File.join(ENV.fetch('PAPERCLIP_ROOT_PATH', File.join(':rails_root', 'public', 'system')), ':class', ':attachment', ':id_partition', ':style', ':filename'),
```
nightpool: `*%w(:rails_root public system)` can clean up these arg lists

lib/mastodon/media_cli.rb
```diff
@@ -44,6 +44,88 @@ def remove
...
+      Scans file storage for files that do not belong to existing media attachments. Because this operation
+      requires iterating over every single file individually, it will be slow.
+
+      Please mind that some storage providers charge for the necessary API requests to list objects.
+    LONG_DESC
+    def remove_orphans
+      progress        = create_progress_bar(nil)
+      reclaimed_bytes = 0
+      removed         = 0
+      dry_run         = options[:dry_run] ? ' (DRY RUN)' : ''
```
nightpool: maybe we should make this :destructive so it's opt in instead of opt out?

nightpool: or maybe we should gather all the files to delete, ask for confirmation, and then delete them

nightpool (approved):
code lgtm % comments

lib/mastodon/media_cli.rb
Gargron: I wouldn't do that because of memory implications if there are a lot of files as well as being able to interrupt the process while having some kind of tangible progress.

angristan:
Thanks for this

PR #12569, opened 2019-12-08
Use different strings on exports page

brawaru:
Currently the page re-uses strings from other contexts which doesn't fit very well - strings incorrectly lowercase-d and pluralized, when they don't need to be, because it's a table.

This PR changes page to re-use `accounts.toots.other` directly, and `admin.accounts` for "Following" and "Follows". This should look more aesthetically pleasing.

I cannot set up local testing Mastodon instance, so please **carefully review the changes** or test them on your local instance. Thanks!

### DIFF
--- a/app/views/settings/exports/show.html.haml
+++ b/app/views/settings/exports/show.html.haml
@@ -9,11 +9,11 @@
         %td= number_to_human_size @export.total_storage
         %td
       %tr
-        %th= t('accounts.posts', count: @export.total_statuses)
+        %th= t('accounts.posts_tab_heading')
         %td= number_with_delimiter @export.total_statuses
         %td
       %tr
-        %th= t('exports.follows')
+        %th= t('admin.accounts.follows')
         %td= number_with_delimiter @export.total_follows
         %td= table_link_to 'download', t('exports.csv'), settings_exports_follows_path(format: :csv)
       %tr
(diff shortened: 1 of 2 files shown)


app/views/settings/exports/show.html.haml
```diff
@@ -9,19 +9,19 @@
         %td= number_to_human_size @export.total_storage
         %td
       %tr
-        %th= t('accounts.posts', count: @export.total_statuses)
+        %th= t('accounts.posts.other')
         %td= number_with_delimiter @export.total_statuses
         %td
       %tr
-        %th= t('exports.follows')
```
Gargron: Why replace a string that seems to exist specifically for this page, though?

Sasha-Sorokin: It's kind of looks off with that “You” prefix being alongside with the ones above and below. And for prefix in general, I think it's redundant because the page itself is suggesting that it's your data to export, probably can remove it from the ones below too.

Sasha-Sorokin:
Hm, why did CI fail with complaint about unused key?

Sasha-Sorokin:
Okay, I have found `accounts.posts_tab_heading`, which is used on user profiles before “Toots and Replies” tab. I'm not sure this is a good idea to use strings from *such different* contexts, but this is the only option to avoid string duplication, plus from search it looks safe to re-use. I'll give it a try in a minute.

Gargron (approved):

PR #12570, opened 2019-12-08
Use codeblock for native redirect URI

brawaru:
This PR changes how `doorkeeper.applications.help.native_redirect_uri` string is being formatted, to use `<code>` tag for `native_redirect_uri` placeholder. This makes the URI look more distinguishable:

| How it looks now | How it supposed to look |
|:----------------:|:-----------------------:|
| Screenshot of actual page | Screenshot of page edited using developer tools |

This only changes how string is generated, so there is no need to change any of locale files.

Before merging, please **carefully review and test** this to ensure it works as expected. Thanks!

### DIFF
--- a/app/views/settings/applications/_fields.html.haml
+++ b/app/views/settings/applications/_fields.html.haml
@@ -7,7 +7,7 @@
 .fields-group
   = f.input :redirect_uri, wrapper: :with_block_label, label: t('activerecord.attributes.doorkeeper/application.redirect_uri'), hint: t('doorkeeper.applications.help.redirect_uri')
 
-  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: Doorkeeper.configuration.native_redirect_uri)
+  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: content_tag(:code, Doorkeeper.configuration.native_redirect_uri)).html_safe
 
 .field-group
   .input.with_block_label


app/views/settings/applications/_fields.html.haml
```diff
@@ -7,7 +7,7 @@
 .fields-group
   = f.input :redirect_uri, wrapper: :with_block_label, label: t('activerecord.attributes.doorkeeper/application.redirect_uri'), hint: t('doorkeeper.applications.help.redirect_uri')
 
-  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: Doorkeeper.configuration.native_redirect_uri)
+  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: "<code>#{Doorkeeper.configuration.native_redirect_uri}</code>")
```
Gargron: That's going to get escaped because this string is not marked as html-safe.

Sasha-Sorokin: By Googling, I should use `.html_safe` at the end of string in order for this to work, yes?

app/views/settings/applications/_fields.html.haml
```diff
@@ -7,7 +7,7 @@
 .fields-group
   = f.input :redirect_uri, wrapper: :with_block_label, label: t('activerecord.attributes.doorkeeper/application.redirect_uri'), hint: t('doorkeeper.applications.help.redirect_uri')
 
-  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: Doorkeeper.configuration.native_redirect_uri)
+  %p.hint= t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: "<code>#{Doorkeeper.configuration.native_redirect_uri}</code>".html_safe)
```
Gargron: Unfortunately, it doesn't work that smart. The escaping happens on the outermost level. The simpliest thing is just adding `.html_safe` to the `t` call. Also it would be better to use `content_tag(:code, Doorkeeper.configuration.native_redirect_uri)` instead of writing out the HTML.

Gargron: The alternative is renaming the string to `native_redirect_uri_html` which will skip escaping it, but that would require updating all locales.

Sasha-Sorokin: Okay, thanks for telling. I will try the first method and corresponding fixes a little later then.

Sasha-Sorokin: Went through nine circles of hell to deal with and install this heckin' Ruby! Took me no less than 1.5 hours. Just to install testing Mastodon instance because I'm tired of these blind changes and I guess you are tired of seeing these too.

Your method actually working:
Screenshot

*btw I also got more context as translator. That was effective :pray:*

Gargron (approved):

PR #12589, opened 2019-12-09
Avoid using pluralize on moderation pages

brawaru:
Pluralize function from Rails framework does not work with other languages than English, moreover it does not even work properly with English <sup>**[\[1\]][note_1]**</sup>. Not that the latest applies to this context, it's just a sign that we best to avoid this function, especially when there are more reliable ways.

This PR changes how reports pages generated in order to avoid usage of pluralize function, replacing it with default translation function, called with given counter. On top of that, we have to make strings pluralize-able, so have to change locale files.

[note_1]: https://medium.com/@anna7/b3927de2ca8e#6a60

<p align="center">Screenshot of report page, showing notes and reports counters for both reporter and reported user</p>

### DIFF
--- a/app/views/admin/reports/index.html.haml
+++ b/app/views/admin/reports/index.html.haml
@@ -28,7 +28,7 @@
     .report-card__profile
       = account_link_to target_account, '', size: 36, path: admin_account_path(target_account.id)
       .report-card__profile__stats
-        = link_to pluralize(target_account.targeted_moderation_notes.count, t('admin.reports.account.note')), admin_account_path(target_account.id)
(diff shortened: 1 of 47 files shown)


Sasha-Sorokin:
By the way, `i18n-tasks` still tries to do [this](https://github.com/tootsuite/mastodon/pull/12497#pullrequestreview-324486311) on normalization and changes to strings. Must we then finally accept these changes in another PR, so it won't bork anything and require manually reviewing each change done to localization files?

ykzts (approved):

ClearlyClaire (approved):

PR #12592, opened 2019-12-10
Fixing inconsistent language

ghost:


config/locales/en.yml
```diff
@@ -11,7 +11,7 @@ en:
     apps: Mobile apps
     apps_platforms: Use Mastodon from iOS, Android and other platforms
     browse_directory: Browse a profile directory and filter by interests
-    browse_local_posts: Browse a live stream of public posts from this server
+    browse_local_posts: Browse a live stream of public posts from this instance
```
Gargron: We use "server", not "instance", since #9938

Gargron (changes requested):

config/locales/en.yml
```diff
@@ -900,7 +900,7 @@ en:
           billion: B
           million: M
           quadrillion: Q
-          thousand: K
+          thoudand: K
```
ykzts:
```suggestion
          thousand: K
```

Is there any reason to change it

ykzts (changes requested):

Issue #12572, opened 2019-12-09
If the account “Media storage” usage exceeds 2GB, “Data export” will fail


mayaeh:
When the total amount of “Media storage” used by the account exceeds 2GB, “Data export” does not seem to run properly.

The backup file size is defined as `integer` in [backup.rb](https://github.com/tootsuite/mastodon/blob/master/app/models/backup.rb).
The `integer` does not seem to be able to handle more than 2GB.
See also: https://github.com/thoughtbot/paperclip/commit/34ec355e43e91c63288aab956a604f17471d4e59

What about measures to split the export file?

### Steps to reproduce the problem

I'm sorry, I haven't actually run “Data export” to confirm.
My main account disk usage is close to 5GB.
I can't try it because I think it will put a heavy load on the server.

However, I have received reports that running it on an account larger than 2GB failed.

Gargron:
Good catch, I feel like we had a similar problem before...

mayaeh:
Yes, I spoke about this at Discord about a year ago.
However, my misunderstanding was included at that time. And I forgot to discuss this matter.

PR #12602, opened 2019-12-12
Increase max backup size

scd31:
Changes backup size from integer to bigint. This solves #12572

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -653,8 +653,8 @@ ActiveRecord::Schema.define(version: 2019_12_12_003415) do
   create_table "status_pins", force: :cascade do |t|
     t.bigint "account_id", null: false
     t.bigint "status_id", null: false
-    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
-    t.datetime "updated_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "created_at", default: -> { "now()" }, null: false
+    t.datetime "updated_at", default: -> { "now()" }, null: false
     t.index ["account_id", "status_id"], name: "index_status_pins_on_account_id_and_status_id", unique: true
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_10_31_163205) do
+ActiveRecord::Schema.define(version: 2019_12_12_003415) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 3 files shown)


ClearlyClaire (approved):

db/schema.rb
```diff
@@ -653,8 +653,8 @@
   create_table "status_pins", force: :cascade do |t|
     t.bigint "account_id", null: false
     t.bigint "status_id", null: false
-    t.datetime "created_at", default: -> { "now()" }, null: false
-    t.datetime "updated_at", default: -> { "now()" }, null: false
+    t.datetime "created_at", default: -> { "CURRENT_TIMESTAMP" }, null: false
```
Gargron: What's going on here?

db/schema.rb
```diff
@@ -703,30 +703,6 @@
     t.index ["tag_id", "status_id"], name: "index_statuses_tags_on_tag_id_and_status_id", unique: true
   end
 
-  create_table "stream_entries", force: :cascade do |t|
```
Gargron: This is okay but shouldn't have been there in the first place...

db/schema.rb
scd31: Seems that `CURRENT_TIMESTAMP` was added to Rails 5. I'm not sure why it decided to change in this PR(and I can change it back if needed). This seems like the more rails-y way to do it, as it is database-agnostic.

Gargron: Yes, I'd like to minimize the changes as much as possible. We don't need to be database-agnostic anyway because we use postgres-specific query features.

scd31: This should be fixed now

ClearlyClaire (approved):

zunda:
ActiveRecord complains about this migration. Might there be a workaround?

```
Migrating to IncreaseBackupSize (20191212003415)
== 20191212003415 IncreaseBackupSize: migrating ===============================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

=== Dangerous operation detected #strong_migrations ===

Changing the type of an existing column requires the entire
table and indexes to be rewritten. A safer approach is to:

1. Create a new column
2. Write to both columns
3. Backfill data from the old column to the new column
4. Move reads from the old column to the new column
5. Stop writing to the old column
6. Drop the old column

/app/db/migrate/20191212003415_increase_backup_size.rb:3:in `up'
  :
```

scd31:
@zunda Well that sucks. I don't have time to fix that(Exam season is right now for me) but someone else should submit a PR to fix that. Otherwise I will have a PR up in a few days.

noellabo:
Since the number of data in the `backups` table is usually very small and the target field is not indexed, operations that change the type of this field seem to be safe in practice.

Display the actual number of records and make sure it is small enough.

```
bin/rails r 'puts Backup.count'
```

If it looks okay, run migration with `SAFETY_ASSURED=1`.

```
SAFETY_ASSURED=1 RAILS_ENV=production bundle exec rails db:migrate
```

You can also change the migration code.

```
safety_assured { change_column :backups, :dump_file_size, :bigint }
```

Gargron:
Since the schema version was updated in the PR, I assumed `db:migrate` has been run successfully. It's strange that this warning didn't come up. In practice I think that this table is so small it's okay to update the column live.

There is I believe also a migration helper to do this the safe way.

scd31:
@Gargron I ran db:migrate, however my database was a clean slate(I had only just installed mastodon for development).
I agree that `safety_assured { change_column :backups, :dump_file_size, :bigint }` is the way to go.

PR #12606, opened 2019-12-13
Ignore development specific env file

berkes:
Developers will often use a `.env.development` as adviced in [dotenv loading order](https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use). 

To avoid accidentally committing keys or other secrets, this should probably be ignored by default.

This PR does that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ public/packs
 public/packs-test
 .env
 .env.production
+.env.development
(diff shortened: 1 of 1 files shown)


nightpool:
thanks for the PR! I absolutely love being able to merge small QoL PRs like this.

nightpool (approved):

PR #12642, opened 2019-12-17
Include capistrano tasks that reload the services after deploying.

berkes:
After a successfull deploy, I need the services to switch over to use the
new code. I need to (re)start the services. Also, often, I need to inspect whether the services are running or why they crashed.

The capistrano tasks in this PR do that.

I'm not certain whether this is useful to mastodon at all. 

1. It requires the server to be set up using these services AND
2. Requires the "deploy" user to be allowed to run those commands with passwordless sudo.

Please let me know what you think about this PR. Is it useful to Mastodon? Would it be useful if I make some changes? Or should it simply be closed?

It might be more appropriate in the specific environment-deploy files for example: `config/deploy/production.rb`. Otherwise, I probably need to add documentation (but where?) and a PR in the ansible-repo to set this up properly.

On my server, using ansible, I configure the deploy user to be allowed to restart services. Without such a configuration, the deploy will fail with this setup.

    - name: "Allow deploy-user to reload the services"
      lineinfile:
        dest: /etc/sudoers
        state: present
        line: "%{{ deploy_user }} ALL=NOPASSWD: /bin/systemctl reload {{ item }}.service"
        validate: 'visudo -cf %s'
      with_items:
        - mastodon-web
        - mastodon-streaming
        - mastodon-sidekiq

    - name: "Allow deploy-user to check the service"
      lineinfile:
        dest: /etc/sudoers
        state: present
        line: "%{{  deploy_user }} ALL=NOPASSWD: /bin/systemctl status {{ item }}.service"
        validate: 'visudo -cf %s'
      with_items:
        - mastodon-web
        - mastodon-streaming
        - mastodon-sidekiq

EDIT: additional detail, is that it depends on the service and its setup whether it can be reloaded.  I have set up my units on my server such that they can be reloaded but with the service-examples in `/dist` of this repo, for example, Sidekiq cannot be reloaded, only restarted.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -12,3 +12,21 @@ set :migration_role, :app
 
 append :linked_files, '.env.production', 'public/robots.txt'
 append :linked_dirs, 'vendor/bundle', 'node_modules', 'public/system'
+
+namespace :systemd do
+  %i[sidekiq streaming web].each do |service|
+    %i[reload restart status].each do |action|
+      desc "Perform a #{action} on #{service} service"
+      task "#{service}:#{action}".to_sym do
+        on roles(:app) do
+          # runs e.g. "sudo restart mastodon-sidekiq.service"
+          sudo :systemctl, action, "#{fetch(:application)}-#{service}.service"
+        end
+      end
+    end
+  end
+end
+
+after 'deploy:publishing', 'systemd:web:reload'
+after 'deploy:publishing', 'systemd:sidekiq:restart'
+after 'deploy:publishing', 'systemd:streaming:restart'


berkes:
These tasks are added to capistrano:
Schermafdruk van 2019-12-17 12-05-45

Gargron:
> I have set up my units on my server such that they can be reloaded but with the service-examples in /dist of this repo, for example, Sidekiq cannot be reloaded, only restarted.

Can you submit those? Although I'm not sure reloading Sidekiq makes a difference since it's not user-facing and is not accepting connections. Curious how you got the streaming server to reload.

berkes:
> Can you submit those? Although I'm not sure reloading Sidekiq makes a difference since it's not user-facing and is not accepting connections. Curious how you got the streaming server to reload.

Certainly! 

Do you want them in a PR, fully integrated, or do you want them for reference, attached? Because in order to make them a full PR, I'll have to rewrite their structure and logic from my setup to that for mastodon.

>Although I'm not sure reloading Sidekiq makes a difference since it's not user-facing

Well, Sidekiq is using the codebase, database structure and so on, so you'll want to reload after a new deploy. E.g. to get new mail-templates, locales or to use migrated models.

Gargron:
> > Although I'm not sure reloading Sidekiq makes a difference since it's not user-facing
> 
> Well, Sidekiq is using the codebase, database structure and so on, so you'll want to reload after a new deploy. E.g. to get new mail-templates, locales or to use migrated models.

Sorry, you misunderstood. I'm talking about `reload` specifically, which is usually done through a SIGHUP and meant to restart a program without completely stopping and then starting it. Puma for example supports `reload` for zero-downtime code changes, i.e. it continues processing and accepting requests while it is loading new code.

config/deploy.rb
```diff
@@ -12,3 +12,30 @@
...
+  end
+
+  def systemctl(action, service)
+    # runs e.g. "sudo restart mastodon-sidekiq.service"
+    sudo :systemctl, action, "#{fetch(:application)}-#{service}.service"
+  end
+end
+
+after 'deploy:publishing', 'systemd:web:reload'
+after 'deploy:publishing', 'systemd:sidekiq:reload'
```
Gargron: Basically, only `web` has a `reload`. The others need a `restart`.

Gargron (changes requested):

config/deploy.rb
berkes: Understood. 

I've changed and refactored the code.

Rebased and force pushed too.

Gargron (approved):

PR #12643, opened 2019-12-17
Remove unused translatable strings

ClearlyClaire:


ykzts (approved):

Gargron:
In the future, please only update English locales. Crowdin will remove the strings when it detects they have been removed from the source locale and this will prevent massive merge conflicts.

Issue #11979, opened 2019-09-27
Following accounts from user-blocked domain is inconsistent


trwnh:
Possible overlap with #7792, related to #8186

### Steps to reproduce
1. Follow an account A from a domain X.
2. "Hide domain" on that domain X.
3. Attempt to follow an account B from that domain X.

### Actual behavior
- I am unable to follow account B, and I receive a 403 error from my client.
- I am still following the account A and receiving posts from them in my home timeline. Boosts of account B do not show in my home timeline.

### Expected behavior
I expect some level of consistency in what "hide domain" actually does. Either let me follow accounts from a hidden domain, remove my follows and not just my followers (and warn me of this), or add more scopes of varying severity to domain filtering like we have with user filtering.

Issue #8186, opened 2018-08-13
Overhaul of moderation/blocking tools needed


ThibG:
I think there is a need for an overhaul of the moderation and user-side blocking tools, making them more flexible and more consistent.

On the admin side of things, we currently can (correct me if I'm wrong):
1. Silence an account, which hides their toots from the public timelines and prevent them from getting in the notifications of people who do not follow them. Still allows them to see toots of any user.
2. Block accounts, which makes them unable to interact with anyone on the instance, or read any of the toots (unless using the public pages of the instance)
3. Silence instances, which has the same effect as silencing all users of that instance
4. Block instances, which has the same effect as blocking all users of that instance (and do not deliver anything to that instance, so following/follower desyncs have no consequence)
5. Optionally, reject/delete media files from an instance

I think this is pretty good, although I would argue that muting someone (or an instance) from public timelines and preventing them from getting in the notifications of people who do not follow them should be two different settings and the same could be said about instances (see #7153).

The terminology could also be changed to be less confusing, or alternatively, get rid of that terminology entirely by using checkboxes: an user/instance could be “blocked” or a combination of “hidden from public TLs” and “blocked unless followed” (see also #5694).

For end-users, now, things get a bit more confusing and inconsistent with “user domain blocks”: they roughly correspond to blocking a whole instance with the exception that you can still follow users there.
Like with admin instance silencing, hiding from public timelines and hiding notifications should be two different settings (see #7792). Furthermore, force-removing followers should be optional.

In addition to all these changes, whether a block is visible to the blocked user should be considered more thoroughly: indeed, I can see the point of not making the blocked user aware, but it is difficult or even impossible to achieve, and leads to more confusing software behaviors and make things harder to debug, especially in a federated setting.

To sum up, I would like to make at least these changes:
- Split “hide from timelines” and “hide from notifications (unless followed)” for both user and instance silences (admin)
- Split “domain hiding” (user) into domain block (removes any kind of relationship, rejects notifications and follow requests), and domain silence (with “hide from timelines” and “hide from notifications unless followed” as two independent options) with also the option to unfollow everyone on that instance (and maybe notifying the remote instance in the process, unlike what is currently done)

deutrino:
Related bug: #9028

ThisIsMissEm:
[...]

PR #12644, opened 2019-12-17
Change the wording of the confirmation dialog for domain block

ClearlyClaire:
This is a port of https://github.com/florence-social/mastodon-fork/pull/125, updated and applied to all occurrences of the source string

This changes the dialog from:

> Are you really, really sure you want to block the entire {domain}? In most cases a few targeted blocks or mutes are sufficient and preferable. You will not see content from that domain in any public timelines or your notifications. Your followers from that domain will be removed.

to:

> Blocking means you will not see content from {domain} in any public timelines or your notifications. Any followers from {domain} will also be removed.

Indeed, there is no need for the “in most cases a few targeted blocks or mutes are sufficient and preferable”, which is very context-dependent and might simply not be true.

### DIFF
--- a/app/javascript/mastodon/containers/domain_container.js
+++ b/app/javascript/mastodon/containers/domain_container.js
@@ -18,7 +18,7 @@ const makeMapStateToProps = () => {
 const mapDispatchToProps = (dispatch, { intl }) => ({
   onBlockDomain (domain) {
     dispatch(openModal('CONFIRM', {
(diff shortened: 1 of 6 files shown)


trwnh:
Should mention also that it filters out boosts of users from that domain, at least as currently implemented. This is something that I learned the hard way when domain-blocking botsin.space and then no longer seeing my friends boost their ebooks bots. Maybe it should also note that any accounts from {domain} that you currently follow will remain followed, and you will continue to see posts from them in your home timeline, but you will be unable to follow any users from {domain} while the block is active. (#11979) (also #8186 maybe)

PR #12648, opened 2019-12-17
Fix manual scrolling issue on Firefox/Windows

### LINKED ISSUE #12607: Manual scroll doesn't work on Windows
### Expected behaviour

Scroll the columns with the mouse

### Actual behaviour

The scrollbar doesn't react

### Specifications

I'm using 3.0.1+ glitch, someone else reported the problem and is using 3.0.1

I'm using Firefox on windows 10
This doesn't seem to happen on Firefox on Ubuntu

Capture of the behaviour here: https://biscuit.town/@Sapphaos/103300086696062102

ClearlyClaire:
Fixes #12607

`will-change: transform` apparently makes manual scrolling impossible on
Firefox/Windows. While this should probably be considered a Firefox bug,
`will-change: transform` seem like a very aggressive performance hint that
may possibly make the browser consume more resources than needed, especially
in multiple-column mode.

This was originally added to improve scrolling performances on mobile, but
I think this isn't necessary anymore, because of the two following reasons:
- `contain: paint` (which is implied by `contain: strict`, which we apply
  whenever the browser supports grids) should have similar effects
- in single-column mode, the scrolling container is the root element, which
  I believe is optimized in at least Chromium

Keep in mind that I have not been able to make in-depth benchmarks, and
especially not been able to try on mobile, so performances should probably
be investigated further…

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -2512,7 +2512,6 @@ a.account__display-name {
   overflow-x: hidden;
   flex: 1 1 auto;
   -webkit-overflow-scrolling: touch;
-  will-change: transform; // improves perf in mobile Chrome
 
   &.optionally-scrollable {
     overflow-y: auto;


Gargron:
cc @nolanlawson 

> in single-column mode, the scrolling container is the root element, which
I believe is optimized in at least Chromium

Seems like it would in any case be safe to remove that property specifically in single-column mode

ThibG:
Yes, it should be safe, but the issue only occurs on “advanced” layout afaik

Gargron (approved):

PR #12652, opened 2019-12-18
Fix an error when ActivityPub::FetchRemoteStatusService url is called…

noellabo:
Fix an error when ActivityPub::FetchRemoteStatusService url is called with nil.

This is the fix for #12173.

### DIFF
--- a/app/services/fetch_remote_status_service.rb
+++ b/app/services/fetch_remote_status_service.rb
@@ -9,6 +9,6 @@ class FetchRemoteStatusService < BaseService
       resource_options = { prefetched_body: prefetched_body }
     end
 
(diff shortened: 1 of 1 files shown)


mayaeh (approved):

ThibG:
It's fine, but I don't understand how the error could have been introduced by that PR in the first place…

noellabo:
If FetchResourceService fails, resource_options will be nil instead of {}.

```
TypeError: no implicit conversion of nil into Hash
from /home/mastodon/live/app/services/fetch_remote_status_service.rb:12:in `call'
```

ThibG:
oh ok, I get it thanks!

PR #12661, opened 2019-12-21
Summary: fix slowness due to layout thrashing when reloading a large …

### LINKED ISSUE #12455: Pressing "back" takes 5 seconds
video: https://assets.octodon.social/media_attachments/files/007/882/867/original/25b66c7b66fb7b63.mp4

### Expected behaviour

Pressing "back" should take less than a second.

### Actual behaviour

Depending how far the timeline was scrolled down and how much it has updated since the status was opened, it can take up to a minute or two to fully load the timeline, or just crash the browser.

Scroll down a timeline some, open a status in detail view, and then press "back."

### Specifications

This is new to Mastodon version 3.

This is in iOS 12, either browser or app bookmark. It seems to work better on Android and desktop browsers.

panarom:
…set of status updates

in order to limit the maximum size of a status in a list view (e.g. the home timeline), so as to avoid having to scroll all the way through an abnormally large status update (see https://github.com/tootsuite/mastodon/pull/8205), the following steps are taken:
•the element containing the status is rendered in the browser
•its height is calculated, to determine if it exceeds the maximum height threshold.
Unfortunately for performance, these steps are carried out in the componentDidMount(/Update) method, which also performs style modifications on the element.  The combination of  height request and style modification during javascript evaluation in the browser leads to layout-thrashing, where the elements are repeatedly re-laid-out (see https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing & https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Performance_best_practices_for_Firefox_fe_engineers).
The solution implemented here is to memoize the collapsed state in Redux the first time the status is seen (e.g. when fetched as part of a small batch, to populate the home timeline) , so that on subsequent re-renders, the value can be queried, rather than recalculated.  This strategy is derived from https://github.com/tootsuite/mastodon/pull/4439 & https://github.com/tootsuite/mastodon/pull/4909, and should resolve https://github.com/tootsuite/mastodon/issues/12455.

Andrew Lin (https://github.com/onethreeseven) is thanked for his assistance in root cause analysis and solution brainstorming

### DIFF
--- a/app/javascript/mastodon/actions/statuses.js
+++ b/app/javascript/mastodon/actions/statuses.js
@@ -322,9 +322,10 @@ export function revealStatus(ids) {
   };
 };
 
-export function toggleStatusCollapse(id) {
+export function toggleStatusCollapse(id, isCollapsed) {
   return {
     type: STATUS_COLLAPSE,
     id,
+    isCollapsed,
   };
 }
--- a/app/javascript/mastodon/actions/statuses.js
+++ b/app/javascript/mastodon/actions/statuses.js
@@ -26,8 +26,9 @@ export const STATUS_UNMUTE_REQUEST = 'STATUS_UNMUTE_REQUEST';
 export const STATUS_UNMUTE_SUCCESS = 'STATUS_UNMUTE_SUCCESS';
 export const STATUS_UNMUTE_FAIL    = 'STATUS_UNMUTE_FAIL';
 
-export const STATUS_REVEAL = 'STATUS_REVEAL';
-export const STATUS_HIDE   = 'STATUS_HIDE';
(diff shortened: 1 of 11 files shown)


panarom:
hold off on merging this: I've introduced a bug when viewing a status thread where if the text is too long it's hidden by a too-small div

Gargron:
Please also check code style errors found by CodeClimate!

Gargron (approved):

impiaaa:
My instance is on 3.1 now, and it does seem better! Thanks!

ThibG:
This reintroduces a lot of issues with the TL moving on its own when a status is deleted etc.

PR #12662, opened 2019-12-22
Add announcements

### LINKED ISSUE #11006: Server annoucements
A lot of servers admins currently struggle making sure everyone is informed on what are new on their Fediverse node. Even if there is an admin account being followed by the whole instance, regular announcements from the admins, a lot of people can be left aside and don't understand the changes.

I would suggest a new feature for admins to be able to make announcements to their users, trough an API and trough the WebUI, like a little column you can open with "news" and persistent notification until read.

Code-wise I don't really know what would be the best for it, since it would take a lot of space to track down every announcements read by users I suggest having an simple number you increase, like there is 5 announcements made since the beginning, and your "number" is 4, it mean there is still one announcement you missed.
Of course if you read the announcements, it mark them all as read, because you can't track down individuals announces.

The list of announces would be retrievable trough an endpoint, and another endpoint to update the "read state".

- [ ] API endpoint to retrieves announcements
- [ ] API endpoint to create/edit/remove announcements
- [ ] API endpoint to get announcements list
- [ ] API endpoint to get announcements "size" (used to calculate if the user read it)
- [ ] API endpoint to get user's announcements "size"(used to calculate if the user read it)
- [ ] API endpoint to set user's announcements "size"

- [ ] Admin interface to create, delete, edit announcements
- [ ] User interface read announcements

Gargron:
Fix #11006

image

image

This PR introduces the ability for admins to publish announcements visible to local users only. The announcements are distributed through the streaming API (and available through the REST API). They can be scheduled for the future and be bound to a certain time range. Announcements can use the standard syntax for mentions, hashtags, URLs and custom emojis. The announcements, when present, are displayed atop the home feed. Users can add reactions to announcements in the form of unicode and custom emojis, and these reactions are distributed through the streaming API as well.


app/javascript/mastodon/features/ui/components/media_modal.js
```diff
@@ -211,7 +211,6 @@ class MediaModal extends ImmutablePureComponent {
             style={swipeableViewsStyle}
             containerStyle={containerStyle}
             onChangeIndex={this.handleSwipe}
-            onSwitching={this.handleSwitching}
```
ThibG: Unrelated, but ok.

app/lib/entity_cache.rb
```diff
@@ -8,7 +8,7 @@ class EntityCache
   MAX_EXPIRATION = 7.days.freeze
 
   def mention(username, domain)
-    Rails.cache.fetch(to_key(:mention, username, domain), expires_in: MAX_EXPIRATION) { Account.select(:username, :domain, :url).find_remote(username, domain) }
+    Rails.cache.fetch(to_key(:mention, username, domain), expires_in: MAX_EXPIRATION) { Account.select(:id, :username, :domain, :url).find_remote(username, domain) }
```
ThibG: Why this change?
Actually, I'm not sure what this line does at all, since `AccountFinderConcern#find_remote` seems to perform a search with its own scope?

app/serializers/rest/announcement_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def id
+    object.id.to_s
+  end
+
+  def content
+    Formatter.instance.linkify(object.text)
+  end
+
+  class AccountSerializer < ActiveModel::Serializer
```
ThibG: Is this serializer used for `mentions`? If so, I'm not sure I understand why/how. `REST::StatusSerializer`'s is called `MentionSerializer` so I'm not sure what's different here…?

ClearlyClaire (changes requested):
In addition to the inline comments, I have a few remarks/questions:
- unlike for TLs and notifications, it seems that if streaming stops working at some point, new announcements will be missed entirely (no polling, nor, unless I'm mistaken, reconnecting to the streaming server?)
- expired announcements seem to not disappear until the web app is restarted; some people typically never restart the web app, which would lead to an ever-growing list
- judging by the code (please add screenshots/screencaps to UI changes), the announcements are displayed just above trending hashtags on single-column, which means it will only be displayed to users using single-column on a sufficiently-large screen. Furthermore, people have complained that trending hashtags were already getting cut off by the window's height, so it might be an issue too.
- It seems to me that whenever an announcement appears, it will offset the selected announcement's index, which is a bit weird: showing neither the announcement that was being displayed, nor the newly-added one (unless the last displayed announcement was the most recent one)

app/controllers/api/v1/announcements_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Api::V1::AnnouncementsController < Api::BaseController
+  before_action :set_announcements
```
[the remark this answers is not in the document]
Gargron: I don't think `respond_to :json` actually does anything at all, definitely not when we call `render json:` everywhere anyway. I think those lines should be removed but in a separate refactor.

app/serializers/rest/announcement_serializer.rb
Gargron: Status model has a Mention association. Here, we use Account models directly.

app/lib/entity_cache.rb
Gargron: Without this the retrieved model doesn't have an `id` attribute. Scopes are composable.

app/javascript/mastodon/features/ui/components/media_modal.js
Gargron: Yes, turns out this was never defined/used.

Gargron:
> It seems to me that whenever an announcement appears, it will offset the selected announcement's index, which is a bit weird: showing neither the announcement that was being displayed, nor the newly-added one (unless the last displayed announcement was the most recent one)

Should the view always slide to the newest announcement when it arrives? Or should it always reset to the first one. Or keep showing the one that was previously being shown?

Gargron:
> unlike for TLs and notifications, it seems that if streaming stops working at some point, new announcements will be missed entirely (no polling, nor, unless I'm mistaken, reconnecting to the streaming server?)

Would it really be worth it to poll the announcements API at the same rate as home/notifications?

ThibG:
> > unlike for TLs and notifications, it seems that if streaming stops working at some point, new announcements will be missed entirely (no polling, nor, unless I'm mistaken, reconnecting to the streaming server?)
> 
> Would it really be worth it to poll the announcements API at the same rate as home/notifications?

No, it would definitely not be worth polling at the same rate. Home/notifications are polled pretty often, announcements do not need to be refreshed nearly as often. I'd guess something as low as once every 15 minutes should be fine.

> > It seems to me that whenever an announcement appears, it will offset the selected announcement's index, which is a bit weird: showing neither the announcement that was being displayed, nor the newly-added one (unless the last displayed announcement was the most recent one)
> 
> Should the view always slide to the newest announcement when it arrives? Or should it always reset to the first one. Or keep showing the one that was previously being shown?

Hm, I'd say either keep to the current announcement and have some way to highlight there is a new one, or, just display the one that got pushed, as this shouldn't happen often anyway.

app/controllers/api/v1/announcements_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Api::V1::AnnouncementsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :write, :'write:accounts' }, only: :dismiss
+  before_action :require_user!, except: :index
```
ThibG: Hm, it could be interesting to have announcements visible only to logged-in users.
Is it useful to give access to unauthenticated users, anyway?

app/models/announcement.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
ThibG: It's lacking the usual schema description other models have.

ClearlyClaire (commented):
Overall this seems ok except for the following facts:
- once discarded, it doesn't seem possible to bring back announcements at all?
- there is no interface or even API to create announcements, the only way is using the rails console…

ThibG:
Also, announcements in advanced mode look less obvious without the mascot
Screenshot_2020-01-14 Mastodon

app/models/announcement.rb
Gargron: It stopped auto-generating them for some reason

app/controllers/api/v1/announcements_controller.rb
Gargron: I thought it might be useful in the same way an instance's custom emojis are publicly available, but I'm not attached to that idea.

ThibG: I'd actually vote against that, since it may not be obvious for admins that announcements could be accessed by anonymous users, and in the case of small instances with trusted users, announcements may possibly contain sensitive information.

app/models/announcement.rb
Gargron: Fixed. They changed config.

app/models/announcement_reaction.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  validates :name, presence: true
+  validates_with ReactionValidator
+
+  before_validation :set_custom_emoji
+
+  private
+
+  def set_custom_emoji
+    self.custom_emoji = CustomEmoji.local.find_by(disabled: false, shortcode: name) if name.present?
```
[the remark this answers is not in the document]
Gargron: This is a `before_validation` action

app/controllers/api/v1/announcements/reactions_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Api::V1::Announcements::ReactionsController < Api::BaseController
+  before_action -> { doorkeeper_authorize! :write, :'write:accounts' }
```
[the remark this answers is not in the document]
Gargron: I tend to see `write:accounts` as anything modifying your own account, and hiding which announcements you see fits in there in my opinion.

app/models/announcement.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      if account.nil?
+        scope.select('name, custom_emoji_id, count(*) as count, false as me')
+      else
+        scope.select("name, custom_emoji_id, count(*) as count, exists(select 1 from announcement_reactions r where r.account_id = #{account.id} and r.announcement_id = announcement_reactions.announcement_id and r.name = announcement_reactions.name) as me")
+      end
+    end
+
+    ActiveRecord::Associations::Preloader.new.preload(records, :custom_emoji)
+    records
```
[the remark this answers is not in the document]
Gargron: For sure, but this is a pre-optimization implementation and it might even end up being fine given the low number of announcements.

app/controllers/api/v1/announcements/reactions_controller.rb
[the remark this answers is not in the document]
ThibG: It doesn't control that, though, only reactions to announcements.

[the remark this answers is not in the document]
Gargron: Oh right, I thought this was about the announcement "dismiss" action. You have a point with `write:favourites`.

ThibG:
Some of the remarks I have made earlier remain:
- There is currently no way for anyone to write announcements, so while this PR is fine, merging it without the corresponding announcement authoring part would make little sense and only encourage implementing API support for a partially-implemented feature
- it doesn't appear to refresh announcements when the websocket connection fails, while arguably this could be a perfect occasion to do so, if the websocket connection fails because of a server-side issue, per instance
- expired announcements do not seem to be removed from the client state, meaning announcements would just pile up (they can be dismissed though so that's a very minor issue)
- it seems newer announcements will offset the currently-viewed announcement, which is rather odd and won't necessarily inform a user that there are new announcements
- there seem to be no way to view or bring back dismissed announcements

app/controllers/api/v1/announcements_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  def dismiss
+    AnnouncementMute.create!(account: current_account, announcement: @announcement)
+    render_empty
+  end
+
+  private
+
+  def set_announcements
+    @announcements = Announcement.published
+    @announcements = @announcements.without_muted(current_account) if user_signed_in?
```
ThibG: This can be simplified since a user is required to be signed in.

app/models/announcement_filter.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      scope.merge!(scope_for(key, value.to_s.strip)) if value.present?
+    end
+
+    scope.chronological
+  end
+
+  private
+
+  def scope_for(key, value)
```
ThibG: I guess this file could be simplified, especially wrt. the `value` argument here

Gargron:
> * expired announcements do not seem to be removed from the client state, meaning announcements would just pile up (they can be dismissed though so that's a very minor issue)
> * it seems newer announcements will offset the currently-viewed announcement, which is rather odd and won't necessarily inform a user that there are new announcements
> * there seem to be no way to view or bring back dismissed announcements

I would rather wait for real user feedback before tackling any of these, since I don't really know what people would expect. I have never wanted to bring a GitHub announcement back to look at it again, so this could just be wasting time unless there's real demand for it.

Sasha-Sorokin:
> expired announcements do not seem to be removed from the client state, meaning announcements would just pile up (they can be dismissed though so that's a very minor issue)

When creating announcement, maybe there should be field for duration or something like that, I actually would like to come back to Mastodon after some time and see what changed on my lovely instance and with Mastodon itself (not for all the time all I was off (who'd be happy to see “1/9999…99”?), but at least some part of it), but announcements about planned maintenance have no practical sense for me.

> there seem to be no way to view or bring back dismissed announcements

There is no point to bringing back specific announcement, but I see point in having announcements log, like blog, but log (haha, sorry…). Like for example, GitHub has [changelog page](https://github.blog/changelog/) where you can see all the recent changes and it is very simple, so must be announcements log, just another view (by view I mean… the same thing as if you clicked “Blocked users” menu item) and recent announcements of certain age (*define age*) or paginated, or infinitely scrollable (?). Link to it can be within the announcements panel, replacing instance domain, like “All announcements” or “View announcements”, or simplified “View all”.

Would that work?

Gargron:
> When creating announcement, maybe there should be field for duration or something like that

There's `ends_at`

This can all be adjusted later

app/javascript/mastodon/actions/notifications.js
```diff
@@ -188,6 +188,7 @@ export function expandNotificationsFail(error, isLoadingMore) {
     type: NOTIFICATIONS_EXPAND_FAIL,
     error,
     skipLoading: !isLoadingMore,
+    skipAlert: !isLoadingMore,
```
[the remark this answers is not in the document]
Gargron: When we're polling because streaming is down, we first poll home, then notifications, then announcements. If the server is down or network is unavailable, this would mean 3 error alerts. We can skip this one. Home and notifications are always either polling or streaming, so unless we are paginating (`isLoadingMore`) we would not be refreshing notifications outside of that.

ThibG:
> > * expired announcements do not seem to be removed from the client state, meaning announcements would just pile up (they can be dismissed though so that's a very minor issue)
> > * it seems newer announcements will offset the currently-viewed announcement, which is rather odd and won't necessarily inform a user that there are new announcements
> > * there seem to be no way to view or bring back dismissed announcements
> 
> I would rather wait for real user feedback before tackling any of these, since I don't really know what people would expect. I have never wanted to bring a GitHub announcement back to look at it again, so this could just be wasting time unless there's real demand for it.

I do think being able to see dismissed announcements (at least for those that are still active) is needed, because should an announcement have some important or actionable information, users will be forced to either keep it (which does eat some significant screen space), or dismiss it and remember the important information anyway. Furthermore, this is kinda at odds with the reactions feature, since you'd be interested in seeing the reactions, but would also want to dismiss the announcement to reclaim some screen space.

ClearlyClaire (approved):

mayaeh:
I found some behavior that seemed to be malfunctioning. Please wait a bit for the merge due to questions.

mayaeh:
I like this feature!

I applied this PR to the test server and tried it.
Then I found the following problem:

1. If the "Schedule publication" date and time are not set, "Published" appears to be displayed in UTC. ("Time range" is not UTC) 
1. When "Time range" is set, the format displayed on the WebUI is not user-friendly depending on the language setting of server. ([sample](https://media.taruntarun.net/Mastodon/media_attachments/files/000/695/350/original/0c1bd7f72797ca69.png))
1. Even if you create a notification that specifies a future date and time in "Begin of event", it will be displayed once on the WebUI. But if you reload after a few minutes, it disappears.
1. When deleting announcement with added reaction, `PublishAnnouncementReactionWorker: ActiveRecord::RecordNotFound: Couldn't find Announcement with 'id'=xx` error is added to Sidekiq. 
1. When the total number of Announcements to be displayed becomes 4, the close button and the next button disappear from the WebUI. And when it reaches 5, the back button and total number disappear. ([sample1](https://media.taruntarun.net/Mastodon/media_attachments/files/000/695/351/original/6340fbae52ea6d3c.png)) ([sample2](https://media.taruntarun.net/Mastodon/media_attachments/files/000/695/352/original/3b2cbfa263fb329e.png)) 
1. You can create an announcement with All-day event checked without setting "Time range".

 
The following is future request:

1. The title displayed on the WebUI is "Update", but is "Announce" better?
1. I want a reset button to re-display the Announcement that was deleted by pressing the close button. I think that it is better to be able to cancel because there is a possibility of accidentally pressing.
1. If you press the close button to update the content of a deleted announcement, do you think it is better to re-display it?
1. If you try to add or delete a "reaction" to an expired or deleted announcement, we should return an error message telling you that the announcement has been deleted.
1. I want a screen that lists "Announcements". I also want a menu item to display that screen to "Getting-started".
1. If possible, also stream announcement's updates and deletions.

app/workers/scheduler/scheduled_statuses_scheduler.rb
```diff
@@ -6,14 +6,38 @@ class Scheduler::ScheduledStatusesScheduler
...
+  def due_announcements
+    Announcement.unpublished.where('scheduled_at IS NOT NULL AND scheduled_at <= ?', Time.now.utc + PostStatusService::MIN_SCHEDULE_OFFSET)
+  end
+
+  def unpublish_expired_announcements!
+    expired_announcements.in_batches.update_all(published: false)
+  end
+
+  def expired_announcements
+    Announcement.published.where('ends_at IS NOT NULL AND ends_at >= ?', Time.now.utc)
```
ThibG: The comparison is wrong, resulting in the scheduler unpublishing any published announcement with a future end date.

ClearlyClaire (dismissed):

PR #12667, opened 2019-12-23
Use prettier

smorimoto:


Gargron:
This is difficult to review and merge because of the size... We need to coordinate a time when all current PRs touching JS are merged, then re-apply this. I also find some of these style changes uncomfortable, like the multi-line function arguments...

PR #12684, opened 2019-12-23
Avoid using uppercase text-transform

brawaru:
One user suggested that the loading indicator should not be written ALL CAPS, at first it was thought this change is very minor, but then a few other people asked agreed on the same thing - variant without caps looks better. It may be related that it is harder to read or just looks too "catchy". Moreover, I asked [@rf@mastodonsocial.ru](https://mastodonsocial.ru/@rf) community what they think of that, and [82% of 22 people agreed on this change](https://mastodon.social/@sasha_sorokin/103323161047099469).

This PR removes all usage of text-transform: uppercase, where the font size specified, it changes the value by one pixel larger, so we still keeping the "designed" size of the labels but without using CAPS.

### DIFF
--- a/app/javascript/styles/mastodon/_mixins.scss
+++ b/app/javascript/styles/mastodon/_mixins.scss
@@ -34,9 +34,8 @@
   box-shadow: 2px 4px 15px rgba($base-shadow-color, 0.4);
 
   h4 {
-    text-transform: uppercase;
     color: $light-text-color;
-    font-size: 13px;
+    font-size: 14px;
     font-weight: 500;
     margin-bottom: 10px;
   }
--- a/app/javascript/styles/mastodon/about.scss
+++ b/app/javascript/styles/mastodon/about.scss
@@ -719,9 +719,8 @@ $small-breakpoint: 960px;
 
     h4 {
       padding: 10px;
-      text-transform: uppercase;
       font-weight: 700;
-      font-size: 13px;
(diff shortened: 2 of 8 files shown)


ykzts (approved):

Gargron:
I agree about the loading indicators, but are you sure about all the other stuff? I mean, I think it looks good, at least in English...

Sasha-Sorokin:
Actually, I used own userstyle which overrides most of the uppercase stuff since almost month ago and I'd say it looks more cleaner to me and now I'd agree that this is a good thing to change.

Turning off userstyle, with uppercase it really feels like shouting (like “Toot!” is just exciting, but “TOOT!” is like yelling at darn machine [to do thing!!!](https://mastodon.social/oops.gif)…) and/or it just drafts too much attention, like headers in settings already have line below, color, bold font to distinguish that they are headers and we aren't in plain text to uppercase them.

I think uppercase may be working in certain design cases, but I've never seen a good one anywhere. You may run a silent study on mastodon.social and after some time ask users with poll if they noticed and whether they agree on the change. The final decision is up to you :)

*Also, if “Toot!” button looks lonely now, I had some [nice userstyle thing](https://i.imgur.com/D4tHRUH.gif) for it too, guess it could be somehow better be integrated by Mastodon itself.*

Gargron (approved):

PR #12685, opened 2019-12-23
fix: make Procfile compatible with herokuish

ghost:
rel: https://github.com/gliderlabs/herokuish/issues/326#issuecomment-365117882


bin/heroku-web
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
SuperSandro2000:
```suggestion
#!/usr/bin/env bash
```

and you are missing a new line.

aisk: Should we added a `set -e` below this line?

SuperSandro2000: `set -eou pipefail` if not commands fail.

SuperSandro2000:
bump

Gargron (approved):

PR #12697, opened 2019-12-27
Circleci - Upgrade to Buster images

shleeable:
The upcoming Ruby 2.7 Docker image and thus CircleCi images seem to only built with Buster.

This is to prepare for the other version upgrades related to dropping Ruby 2.4 in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 aliases:
   - &defaults
     docker:
-      - image: circleci/ruby:2.6-stretch-node
+      - image: circleci/ruby:2.6-buster-node
         environment: &ruby_environment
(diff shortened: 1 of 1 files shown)


Shleeble:
> LoadError: libicudata.so.57: cannot open shared object file: No such file or directory - /home/circleci/projects/mastodon/vendor/bundle/ruby/2.6.0/gems/charlock_holmes-0.7.7/lib/charlock_holmes/charlock_holmes.so

Charlock doesn't seem to like buster because libicudata 57 is not available in buster. https://packages.debian.org/buster/libicu-dev

Currently figuring out if this icu is set on our side or theirs. https://github.com/brianmario/charlock_holmes

PR #12705, opened 2019-12-29
Circleci - Upgrade to Buster Images

shleeable:
This should be safe to merge anytime.

The upcoming Ruby 2.7 Docker image and thus CircleCi images seem to only built with Buster. This is in preparation for the future 2.7 upgrade.

Note: I've fixed two dependence issues with ICU and Protobuf in Buster Debian.
1. libicudata.57 is not available in buster (63). https://packages.debian.org/buster/libicu-dev
2. Protobuf.10 is not avaiable in buster (17). https://packages.debian.org/buster/libprotobuf-dev

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 aliases:
   - &defaults
     docker:
-      - image: circleci/ruby:2.6-stretch-node
+      - image: circleci/ruby:2.6-buster-node
         environment: &ruby_environment
           BUNDLE_APP_CONFIG: ./.bundle/
           DB_HOST: localhost
@@ -39,7 +39,6 @@ aliases:
     steps:
       - checkout
       - *attach_workspace
-
       - restore_cache:
           keys:
             - v1-node-dependencies-{{ checksum "yarn.lock" }}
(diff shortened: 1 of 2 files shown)


Shleeble:
Notes: 

> LoadError: libicudata.so.57: cannot open shared object file: No such file or directory - /home/circleci/projects/mastodon/vendor/bundle/ruby/2.6.0/gems/charlock_holmes-0.7.7/lib/charlock_holmes/charlock_holmes.so

Charlock doesn't seem to like buster because libicudata 57 is not available in buster (63). https://packages.debian.org/buster/libicu-dev

Gargron (approved):

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 aliases:
   - &defaults
     docker:
-      - image: circleci/ruby:2.6-stretch-node
+      - image: circleci/ruby:2.6-buster-node
```
xiaoronglv, line 6: what's the difference between stretch and buster?

xiaoronglv, line 6: what's the difference between stretch and buster?

Gargron: https://www.debian.org/releases/stretch/

> Debian 9 has been superseded by Debian 10 ("buster"). Security updates have been discontinued as of July 6th, 2020.

xiaoronglv: Gargron, thanks for your explanation!

PR #12715, opened 2019-12-30
Fix IDN mentions not being processed, IDN domains not being rendered

### LINKED ISSUE #7812: IDN domains don't render correctly in usernames
IDN domains in usernames are displayed as punycode in usernames in the web interface, which isn't the intended result.

screenshot from 2018-06-15 22-49-35

here, for example, "@TransPosees@xn--pnibles-bya.xn--transposes-i7a.eu" should be displayed as "@TransPosees@pénibles.transposées.eu"

* * * *

### LINKED ISSUE #12246: Mentions of users with unicode domain names are not converted to punycode
I'm trying to mention my other account (`@sneak@hæresiar.ch`) on my own instance `@hæresiar.ch` from the public `@tilde.zone` instance but the mention fails.

### Expected behaviour

The mention's domain-part should be converted to punycode.

### Actual behaviour

The mention failed.

Mention a user @ an IDN

### Specifications

This is from the `@tilde.zone` (not run by me) instance which is running 3.0.0 afaik patched only to increase toot length.

Gargron:
This changes the REST API to return unicode domains in the `acct` attribute instead of punycode, and to render unicode instead of punycode on public HTML pages as well.

Fix #7812, fix #12246

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -13,7 +13,7 @@ module AccountsHelper
     if account.local?
       "@#{account.acct}@#{Rails.configuration.x.local_domain}"
     else
-      "@#{account.acct}"
+      "@#{account.pretty_acct}"
     end
   end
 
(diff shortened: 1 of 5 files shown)


ykzts (approved):

tigransimonyan:
Hello,
I think the current version of Mastodon has the same problem. Please check the issue: https://github.com/mastodon/mastodon/issues/16602

PR #12733, opened 2019-12-30
Hide blocked users from more places

### LINKED ISSUE #12732: Blocked users should be hidden from more places
### Pitch

Blocked users (and maybe muted or user currently blocking us) should be filtered from:
- list of boosts and favourites for a toot
- list of follows and followers for an account

### Motivation

Blocking an user usually means one doesn't want to see them or be reminded about them, but there are still places where blocked people can pop up.

ClearlyClaire:
Fixes #12732

Because it seemed easier, it also hides muted users and users who blocked us, but whether to do that is debatable.

- [x] hide blocked users from toots' fav list
- [x] hide blocked users from toots' reblogged list
- [x] hide blocked users from account followers (API)
- [x] hide blocked users from account followers (public pages, when logged in)
- [x] hide blocked users from account following (API)
- [x] hide blocked users from account following (public pages, when logged in)

### DIFF
--- a/app/controllers/api/v1/accounts/follower_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/follower_accounts_controller.rb
@@ -21,7 +21,9 @@ class Api::V1::Accounts::FollowerAccountsController < Api::BaseController
   def load_accounts
     return [] if hide_results?
 
-    default_accounts.merge(paginated_follows).to_a
+    scope = default_accounts
+    scope = scope.where.not(id: current_account.excluded_from_timeline_account_ids) unless current_account.nil?
+    scope.merge(paginated_follows).to_a
   end
 
   def hide_results?
--- a/app/controllers/api/v1/accounts/following_accounts_controller.rb
+++ b/app/controllers/api/v1/accounts/following_accounts_controller.rb
@@ -21,7 +21,9 @@ class Api::V1::Accounts::FollowingAccountsController < Api::BaseController
   def load_accounts
     return [] if hide_results?
 
-    default_accounts.merge(paginated_follows).to_a
+    scope = default_accounts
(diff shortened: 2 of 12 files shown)


Gargron (approved):

trwnh:
i'm uneasy with the fact that muted users are hidden too, because when i mute users i only do it because i don't want to see their statuses (and in fact, i make heavy use of the feature that allows to mute a user without muting notifications from them).

i'm also not sure about this because i may want to know whether someone i've blocked is following or followed by someone.

ThibG:
> i'm uneasy with the fact that muted users are hidden too, because when i mute users i only do it because i don't want to see their statuses (and in fact, i make heavy use of the feature that allows to mute a user without muting notifications from them).

I tend to agree, but:
1. The current code leverages existing caches, and I don't know if it's worth changing just for that
2. Until #11562 or similar is merged, we would have no way to completely hide someone without also sending them information about the block

> i'm also not sure about this because i may want to know whether someone i've blocked is following or followed by someone.

I don't think that's big of a deal, because that information was already only available through the web interface if either account was on the local instance and the account you were looking did not “hide their network”. Looking at the public pages without logging in is still possible, and I think it's better that this requires that extra step rather than pushing people your purposely blocked on your face.

Also, the same point could have been made about “I want to know if that person boosts that other person I have blocked/muted”

trwnh:
It's more about the scope of things, really -- mute to me includes content,
and blocking adds followship.

Really, I think the underlying issue here is that mutes and blocks need a
major rework (although i forgot the issue number for that). Muting i guess
is more similar to filtering, but by account instead of by keyword. I have
some thoughts about that but I should develop them a bit more before
sharing -- I'll do so on the appropriate issue.

PR #12735, opened 2019-12-30
fix root element of IntersectionObserver

panarom:
was previously using the <ScrollableList element, rather than the browser viewport, meaning
• hideIfNotIntersecting was never called
• when going <Back from a status update to a list view (e.g. home timeline), handleIntersection was called on ALL statuses (however many of them) at once, since they were all intersecting with the <ScrollableList element (rather than just the few in the browser viewport)

### DIFF
--- a/app/javascript/mastodon/components/scrollable_list.js
+++ b/app/javascript/mastodon/components/scrollable_list.js
@@ -210,10 +210,13 @@ export default class ScrollableList extends PureComponent {
   }
 
   attachIntersectionObserver () {
-    this.intersectionObserverWrapper.connect({
+    let nodeOptions = {
       root: this.node,
       rootMargin: '300% 0px',
-    });
+    };
+
+    this.intersectionObserverWrapper
+      .connect(this.props.bindToDocument ? {} : nodeOptions);
   }
 
   detachIntersectionObserver () {


app/javascript/mastodon/components/scrollable_list.js
```diff
@@ -209,13 +209,6 @@ export default class ScrollableList extends PureComponent {
     this.setState({ fullscreen: isFullscreen() });
   }
 
-  attachIntersectionObserver () {
-    this.intersectionObserverWrapper.connect({
-      root: this.node,
```
Gargron: Now wait... You're fixing it for single-column mode but breaking it for advanced layout, I think.

panarom: I tested it in both modes; is there some specific functionality I missed?

panarom: Ok, I've also discovered a discrepancy: in single-column mode, the column is of arbitrary size (extending indefinitely downwards into the browser, causing the behavior this PR is supposed to fix); whereas, in column mode, each column stops at the bottom (well, slightly above) of the viewport, and is scrollable (hideIfNotIntersecting is indeed called).  If you'd prefer, I can try to make a CSS change to make the column-layout consistent across advanced/basic interface modes (thereby hopefully achieving the same behavior for both); but I'm still currently of the opinion that the viewport is the correct root element for intersection calculation, regardless of interface.

Gargron: Well, in single-column mode, the scrollbar is on the document body itself, while in advanced layout, the scrollbar is on ScrollableList, which is why this code, which originates from when only advanced layout existed, uses `this.node` as `root`. Seems like it should be conditional as either `document.body` or `this.node` depending on layout (`this.props.bindToDocument`)

panarom: ok, I'll try introducing the conditional and seeing what happens; will update with results

panarom: this seems to work

Gargron (approved):

PR #12736, opened 2019-12-31
Add translation project promotion link

brawaru:
This PR adds promotional notice on appearance settings about translation project if any other locale than English is used. It allows users to learn and contribute translations to Mastodon.

<p align=center>Example screenshot</p>

**Step ahead**, in this commit one unusual string is added — link to a guide. By default it refers to Crowdin project itself, but if any of Mastodon localization teams established their own guide, they can refer it. Or, if Crowdin supports localized domain for language, it can also be put there (e.g. https://fr.crowdin.com/...).

### DIFF
--- a/app/views/settings/preferences/appearance/show.html.haml
+++ b/app/views/settings/preferences/appearance/show.html.haml
@@ -8,6 +8,10 @@
     .fields-group.fields-row__column.fields-row__column-6
       = f.input :setting_theme, collection: Themes.instance.names, label_method: lambda { |theme| I18n.t("themes.#{theme}", default: theme) }, wrapper: :with_label, include_blank: false, hint: false
 
+  - unless I18n.locale == :en
+    .flash-message{ style: "text-align: unset; color: unset" }
+      #{t 'appearance.localization.body'} #{content_tag(:a, t('appearance.localization.guide_link_text'), href: t('appearance.localization.guide_link'), target: "_blank", rel: "noopener", style: "text-decoration: underline")}
+
   %h4= t 'appearance.advanced_web_interface'
 
   %p.hint= t 'appearance.advanced_web_interface_hint'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -581,6 +581,10 @@ en:
     animations_and_accessibility: Animations and accessibility
     confirmation_dialogs: Confirmation dialogs
     discovery: Discovery
+    localization:
+      body: Mastodon is translated by volunteers.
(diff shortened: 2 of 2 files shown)


app/views/settings/preferences/appearance/show.html.haml
```diff
@@ -8,6 +8,10 @@
     .fields-group.fields-row__column.fields-row__column-6
       = f.input :setting_theme, collection: Themes.instance.names, label_method: lambda { |theme| I18n.t("themes.#{theme}", default: theme) }, wrapper: :with_label, include_blank: false, hint: false
 
+  - if I18n.locale.to_s != 'en'
+    .flash-message{ style: "text-align: unset; color: unset" }
+      #{content_tag(:span, t('appearance.localization.body'))} #{content_tag(:a, t('appearance.localization.guide_link_text'), href: t('appearance.localization.guide_link'), target: "_blank", rel: "noopener", style: "text-decoration: underline")}
```
Sasha-Sorokin: > `#{content_tag(:span, t('appearance.localization.body'))}`

Oops, leftover. Will remove.

Sasha-Sorokin:
Can someone reassure me, in production it will fallback to English translations if some strings not translated, right? Because on testing instance it makes spans instead, so it worries me a little.

app/views/settings/preferences/appearance/show.html.haml
```diff
@@ -8,6 +8,10 @@
     .fields-group.fields-row__column.fields-row__column-6
       = f.input :setting_theme, collection: Themes.instance.names, label_method: lambda { |theme| I18n.t("themes.#{theme}", default: theme) }, wrapper: :with_label, include_blank: false, hint: false
 
+  - if I18n.locale.to_s != 'enn'
```
Gargron:
```haml
- unless I18n.locale == :en 
```

Gargron (changes requested):

Gargron:
> Can someone reassure me, in production it will fallback to English translations if some strings not translated, right?

Correct. The spans only appear in development mode.

app/views/settings/preferences/appearance/show.html.haml
Sasha-Sorokin: Oh. I used `enn` for testing, but stashed this change, not sure how it end up getting here. And thanks for the tip about `unless`, I already forgot about it since Jekyll. Will fix in a minute.

Gargron (approved):

PR #12759, opened 2020-01-04
Gemfile Updates - fixes warnings

shleeable:


Shleeble:
This should be ready to commit.

Gargron (approved):

ThibG:
What does it fix wrt. faraday? It seems elasticsearch-transport 7.3.0 is in fact incompatible with faraday 1 (it is explicit in elasticsearch-transport 7.5.0's Gemfile, and while there is no release yet supporting farday 1, master is supposed to do)

PR #12760, opened 2020-01-04
parser gem test

shleeable:
Figure out why this fails.


Gargron:
> i18n-tasks: Error scanning app/models/concerns/paginable.rb: undefined method `to_ast' for false:FalseClass
> Did you mean?  to_s

PR #12764, opened 2020-01-05
Gemfile Update - minor upgrades

shleeable:
Updated gemfile/gemfile.lock


Shleeble:
This should be ready to commit.

Gargron:
We have dependabot in charge of keeping dependencies up-to-date. This allows us to review changelogs of the dependency updates and make an educated decision about them. These mass-update PRs are not necessary and will not be accepted unless there is a very good and clear reason.

Shleeble:
Understood.

PR #12765, opened 2020-01-05
[WIP] proposal to publish source codes automatically

masarakki:
Create zip file contains source codes that **running** on booting rails,
it is simple way to satisfy AGPL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ public/system
 public/assets
 public/packs
 public/packs-test
+public/source.zip
 .env
 .env.production
 .env.development
--- a/Gemfile
+++ b/Gemfile
@@ -100,6 +100,7 @@ gem 'webpush'
 gem 'json-ld', git: 'https://github.com/ruby-rdf/json-ld.git', ref: 'e742697a0906e74e8bb777ef98137bc3955d981d'
 gem 'json-ld-preloaded', '~> 3.0'
 gem 'rdf-normalize', '~> 0.3'
+gem 'rubyzip'
 
 group :development, :test do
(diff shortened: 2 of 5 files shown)


ThibG:
hm, I'm not super sure about that, I have two issues with that:
1. Doing it while starting Rails might incur some delay when starting/restarting the server
2. As far as I understand, the `Dir.glob` thing will call the block for each and every file, and the filter happens later. That means it will walk the wall `node_modules` and `public/system`, which potentially contains hundreds of gigabytes of small media files.
3. Such an archive isn't complete, as things like `public` contain source files
4. There is also a risk of including extra files an admin could have in those directories (e.g., some backup config with passwords, etc.)

Gargron:
Agree with @ThibG, I run 3 servers with Puma for mastodon.social, this would run ZIP 3 times on the project directory on every restart, and on 2 of those the static files are not directly served so it's completely pointless. Would be better as part of assets:precompile, but still questionable.

noellabo:
This proposal is intended to automatically comply with the AGPL if you make changes to Mastodon without using a public repository such as Github. Therefore, if `SOURCE_BASE_URL` or `GITHUB_REPOSITORY` is set, it is better to skip the archive.

masarakki:
Even if owner publishes on github, we can not believe that it is real or fake.
so I think it is better to make server vorbose.

masarakki:
I agree this code make server busy, and have risk of inclusion secrets.

tcitworld:
> Even if owner publishes on github, we can not believe that it is real or fake.

And they can fake the exported code through this method as well. You will always need to trust the server administrator anyway.

PR #12789, opened 2020-01-06
refactor relative-time link to point to local-instance thread

panarom:
previous behavior was for relative-time posted link on a status to reference the origin URL (i.e. on the poster's home instance).  This change refactors that link to point to the user's instance's thread.  This change is to streamline the use-case where the home thread is being browsed, and a particular status/thread is opened in a new browser tab, in order to peruse/respond at a later time.

### DIFF
--- a/app/javascript/mastodon/components/status.js
+++ b/app/javascript/mastodon/components/status.js
@@ -114,6 +114,10 @@ class Status extends ImmutablePureComponent {
     }
   }
 
+  static getStatusURL(status) {
+    return `/statuses/${status.getIn(['reblog', 'id'], status.get('id'))}`;
+  }
+
   handleToggleMediaVisibility = () => {
     this.setState({ showMedia: !this.state.showMedia });
   }
@@ -128,8 +132,7 @@ class Status extends ImmutablePureComponent {
       return;
     }
 
-    const { status } = this.props;
-    this.context.router.history.push(`/statuses/${status.getIn(['reblog', 'id'], status.get('id'))}`);
+    this.context.router.history.push(this.constructor.getStatusURL(this.props.status));
   }
 
   handleExpandClick = (e) => {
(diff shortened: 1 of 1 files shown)


ThibG:
Hm, I don't really like this change, I think the original URL makes more sense for sharing and viewing the toot in its full context (including replies the local instance might not know about)

As for marking toots as to be replied to later… I hope the recent bookmark feature answers this use case

Gargron:
I'm afraid I would not accept this change. Permalinks serve an important function for sharing and viewing and have traditionally been always connected to timestamps in all blog systems. Furthermore we do not want to incentivize opening multiple tabs of the web UI, the public page of a single post is a lot more efficient to render than a whole web UI.

panarom:
@ThibG the bookmarking feature is inappropriate for this usecase because it adds significant overhead (bookmarking, searching/scrolling through the bookmarks, reading/replying to the thread, then removing the temporary bookmark) for what is already handled elegantly by the tabbed interface of the web-browser (open in new tab, read/respond, close tab).  To elaborate further on the impetus, I tend to only scroll the timeline once or twice a day; and my preference is to scroll the whole interim timeline, opening potentially interesting conversations in additional tabs as I go to peruse later.

@Gargron the most significant problem this PR attempts to solve is specifically when the user wants to respond to the permalinked post.  This is only an easy workflow if one goes from timeline to conversation to reply immediately; if the permalink is opened, or the timeline is scrolled away, the only way I know of to get to back to a reply-able state is to search for the author, click on their avatar, and scroll through their history, manually looking for the thread; then clicking on the post to open the thread.  What do you think of passing a "reply-to" parameter in the permalink URL, so that when it's opened on its canonical server, it can render a "reply" button that will return the user to their own (logged-in) home-view on their instance?  Alternatively/additionally, perhaps a blank reply-to input could be placed at the bottom of the permalinked conversation, where a user from another instance could put in their @<name>@<instance> if they wanted to reply, and then be sent to their instance's view of the conversation.   That would also handle the case where a link to a conversation of interest was propagated through a medium other than mastodon itself (such as email or text message).

ThibG:
You can reply yo a post without expanding it, so you can reply from the bookmarks column. I agree that's a few more clicks though, especially with the bookmark/unbookmark actions being hidden in the dropdown menu.

Permalinks can be pasted in the search bar, directly bringing the linked toot in the search results (only works for public toots right now, but private toots you've already seen and you have access to will also pop up there in the next release).

PR #12791, opened 2020-01-06
Require nodejs 10.13 as minimum release.

shleeable:
nodejs 8 is EOL.

This can be merged as is now whenever the Big G feels the need.


ykzts (approved):

mstdn:
Can we just upgrade NodeJS with "curl -sL https://deb.nodesource.com/setup_10.x | bash -"?

Gargron:
I guess so

mstdn:
> I guess so

Works like a charm!

lcotonea:
On my version I'm using Node 12 which is also a LTS. Some deprecation warning, but it works like a charm too.

PR #12793, opened 2020-01-07
Little improvements to filters settings page

### LINKED ISSUE #11020: "Add new filter" button should be at top
Adding multiple filters on the /filters page is harder than it should be when there are multiple filters already. Each new filter makes the list longer and pushes the button down further, necessitating more scrolling.

I propose moving the button to the top, either immediately above the table, or next to the header on wider layouts.

image

### LINKED ISSUE #12790: Suggestion / Feature request: Filters > move the "ADD NEW FILTER" button
Add_New_Filter
### Pitch

I would like to suggest this: move the ADD NEW FILTER button here upwards, above the added filters.
https://mastodon.social/filters

### Motivation

The reason behind this is: the more filters I add, the more I have to scroll that page in order to add another filter, which is quite uncomfortable, as you can see in the attached picture.

brawaru:
When you have many filters, it may be hard for you to reach the button to create yet another one. This PR moves creation button to the heading, leaving the page just for the list.

On the other hand, when there are no filters, page looks kind of strange with the empty table. So text stating obvious fact that user has no filters was added in this PR too.

<p align=center>
Demo screenshot
</p>

Closes #11020
Closes #12790

### DIFF
--- a/app/views/filters/index.html.haml
+++ b/app/views/filters/index.html.haml
@@ -1,20 +1,25 @@
 - content_for :page_title do
   = t('filters.index.title')
 
-.table-wrapper
-  %table.table
-    %thead
-      %tr
-        %th= t('simple_form.labels.defaults.phrase')
-        %th= t('simple_form.labels.defaults.context')
-        %th
-    %tbody
-      - @filters.each do |filter|
-        %tr
-          %td= filter.phrase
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Cassolotl:
This is such a nicely written PR comment! :)

PR #12809, opened 2020-01-09
Fix some MP3 files not being recognized as such

### LINKED ISSUE #12693: Error on uploading audio with embedded image
When uploading an mp3 or oga/ogg file with an embedded image (e.g. an album cover), Mastodon gives this on screen error:

`422 Validation failed: File has contents that are not what they are reported to be`

When removing the image (with e.g. EasyTag) from the audio file, uploading is successful.

Tested it on todon.nl and mastodon.social.

### Specifications

Mastodon v3.0.1

ClearlyClaire:
Work around libmagic mistakenly identifying some obscure font format for some
files, by not stopping on the first guess, and taking the first that is *not*
that file format.

Fixes #12693

### DIFF
--- a/app/models/concerns/attachmentable.rb
+++ b/app/models/concerns/attachmentable.rb
@@ -8,6 +8,8 @@ module Attachmentable
   MAX_MATRIX_LIMIT = 16_777_216 # 4096x4096px or approx. 16MB
   GIF_MATRIX_LIMIT = 921_600    # 1280x720px
 
+  BOGUS_MIME_TYPES = %w(application/x-font-gdos).freeze
+
   included do
     before_post_process :obfuscate_file_name
     before_post_process :set_file_extensions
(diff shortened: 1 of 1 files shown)


koyuawsmbrtn:
Big issue for me. Would help a lot to get it merged before 3.1

cc @Gargron

saper:
Is there an upstream bug? It is nice to have a temporary workaround but this is certainly not the right solution here (big thanks @ThibG for getting it sorted out).

app/models/concerns/attachmentable.rb
```diff
@@ -63,7 +63,12 @@ def appropriate_extension(attachment)
   end
 
   def calculated_content_type(attachment)
-    content_type = Paperclip.run('file', '-b --mime :file', file: attachment.queued_for_write[:original].path).split(/[:;\s]+/).first.chomp
+    content_types = Paperclip.run('file', '-b -k -r --mime :file', file: attachment.queued_for_write[:original].path).split("\n- ")
+    content_types -= ['application/x-font-gdos']
```
Gargron, line 67: I would feel better if you aligned the `=` and put the mime type into a constant variable with a self-explaining name like `BOGUS_MIME_TYPES` or something.

Gargron (changes requested):

app/models/concerns/attachmentable.rb
ThibG: Addressed this just in case, but I think this PR isn't needed anymore, as the underlying bug in `file` has been fixed along with security fixes a while ago.

Gargron: Ah, in that case the PR can be closed. The workaround is still concerning to me because what if someone tried to upload a real font file, it would then be accepted and treated as MP3...

ThibG: It probably wouldn't, as the font file would probably not be detected as mp3 (the bug happens because both the mp3 magic and the font thing are detected, and the font thing used to take precedence even though it was unreliable and uses of that font format are incredibly rare).
Closing.

PR #12815, opened 2020-01-10
Add CircleCI task to test migrations

ClearlyClaire:
While this probably won't help with migrations depending on certain data values, it should at least catch most migrations failing due to `strong_migrations` updates, or some model changes.


ThibG:
Adequately fails because #12787 hasn't been merged

ykzts (approved):

PR #12819, opened 2020-01-11
WIP: Flodgatt on Docker Testing

shleeable:
This is my configuration on aus.social using the precompiled binary. 
If we want to compile from source instead, I'll make a fork to download the Rust compiler.

Notes:
* the Dockerfile `chmod +x` seems to not work. Seems to prompt access denied without the additional chmod in the docker-compose command. - FIX THAT

### DIFF
--- a/Dockerfile.flodgatt
+++ b/Dockerfile.flodgatt
@@ -1,0 +1,133 @@
+FROM ubuntu:18.04 as build-dep
+
+# Use bash for the shell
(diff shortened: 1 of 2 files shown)


nightpool:
hmm, that doesn't seem to have actually made it a draft PR. @Shleeble if this isn't suitable for merging, would you mind marking it as such in github's UI? (using the "This is a draft" checkbox)

PR #12820, opened 2020-01-11
WIP: Flodgatt on Docker Testing

shleeable:
This is my configuration on aus.social using the precompiled binary.
If we want to compile from source instead, I'll make a fork to download the Rust compiler.

Notes:
* the Dockerfile will likely be split again with the official Flodgatt docker container.
* the Dockerfile chmod +x seems to not work. Seems to prompt access denied without the additional chmod in the docker-compose command. - FIX THAT


Gargron:
Separate `docker-compose.yml.flodgatt` seems unnecessary? Wouldn't it just be a commented-out container just like `es`?

Gargron:
For that matter, why include `Dockerfile.flodgatt` in the repository that does not contain flodgatt code, when tootsuite/flodgatt should probably just publish an image to the Docker hub?

Shleeble:
> Separate `docker-compose.yml.flodgatt` seems unnecessary? Wouldn't it just be a commented-out container just like `es`?

> For that matter, why include `Dockerfile.flodgatt` in the repository that does not contain flodgatt code, when tootsuite/flodgatt should probably just publish an image to the Docker hub?

the `.flodgatt` is completely pointless. I'm just doing this incase people want to join me and test flodgatt quickly. I've merged the Dockerfile because it's actually won't break anything.

I'm going to maintain this until I'm happy with it over time...

Issue #12382, opened 2019-11-14
Seperate Media and Cache Buckets


irlcatgirl:
### Pitch

I would like an option to store local media and cached remote media in separate S3 buckets. This would be configurable in the `.env` files.

### Motivation

This would create smaller backup sizes for new instances and make them more manager (in stages) for larger ones.

Shleeble:
That's a wonderful idea.. Local bucket / remote bucket.

Gargron:
This would be great. Have to figure out how to a) use dynamic paperclip options based on model attributes and b) introduce this in already live systems without losing access to old files. :thinking: 

Conceptually similar to #6675

nsuchy:
For A) Could we do two separate Paperclip instances with a separate configuration or is it a Singleton class?

For B) We would need a special rake task to copy files over. It might be expensive for some instances.

Cordially,
Nathaniel Suchy (they/them)

luxtiasco:
I don't think it is solved, still unable to use different s3 buckets for cache and local media.

Issue #6675, opened 2018-03-07
[Feature] Separate media directories for local media and remote media


ThomasLeister:
**Feature request**: Store local media in different directory than remote media. 

Reason: This way you can easily back up only important local media by your own users and ignore remote media, which can be seen as less critical.  

Idea by @Sascha@bonn.social [Mastodon].

* * * *

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Gargron:
This would be nice but I don't see a way to introduce this change now.

ThomasLeister:
Hmmm, yes, I see the problem. Maybe someone comes up with a smart idea regarding this :-) If you don't mind we can keep this issue open for some more time?

abochmann:
Can someone provide a pointer on how to identify all local media files and their storage path in the _media_attachments_ table (or elsewhere)? Having that, it would be easy to create a script that just backs up those.

Gargron:
Fixed by #6675

osresearch:
> Fixed by [#6675](https://github.com/mastodon/mastodon/issues/6675)

Isn't this #6675? Was there a different commit that fixed it?

trwnh:
https://github.com/mastodon/mastodon/pull/12821

PR #12821, opened 2020-01-11
Add separate cache directory for non-local uploads

Gargron:
The following resources will be split into a top-level cache directory:

- custom emojis
- avatars
- headers
- media attachments
- preview cards

...when they don't originate from local accounts. With the exception of preview cards, which are *always* counted as not local. The change comes into effect for anything uploaded or downloaded after the upgrade, old files remain in place, so as to not incur massive object storage costs moving potentially terabytes of data.

Adds `tootctl upgrade storage-schema` to upgrade all existing files at once.

### DIFF
--- a/lib/mastodon/upgrade_cli.rb
+++ b/lib/mastodon/upgrade_cli.rb
@@ -1,0 +1,148 @@
+# frozen_string_literal: true
+
+require_relative '../../config/boot'
+require_relative '../../config/environment'
+require_relative 'cli_helper'
+
+module Mastodon
+  class UpgradeCLI < Thor
+    include CLIHelper
+
+    def self.exit_on_failure?
+      true
+    end
+
+    CURRENT_STORAGE_SCHEMA_VERSION = 1
+
+    option :dry_run, type: :boolean, default: false
+    option :verbose, type: :boolean, default: false, aliases: [:v]
+    desc 'storage-schema', 'Upgrade storage schema of various file attachments to the latest version'
+    long_desc <<~LONG_DESC
+      Iterates over every file attachment of every record and, if its storage schema is outdated, performs the
+      necessary upgrade to the latest one. In practice this means e.g. moving files to different directories.
+
+      Will most likely take a long time.
+    LONG_DESC
+    def storage_schema
+      progress = create_progress_bar(nil)
+      dry_run  = dry_run? ? ' (DRY RUN)' : ''
+      records  = 0
+
+      klasses = [
+        Account,
+        CustomEmoji,
+        MediaAttachment,
+        PreviewCard,
+      ]
+
+      klasses.each do |klass|
+        attachment_names = klass.attachment_definitions.keys
+
+        klass.find_each do |record|
+          attachment_names.each do |attachment_name|
+            attachment = record.public_send(attachment_name)
+
+            next if attachment.blank? || attachment.storage_schema_version >= CURRENT_STORAGE_SCHEMA_VERSION
+
+            attachment.styles.each_key do |style|
+              case Paperclip::Attachment.default_options[:storage]
+              when :s3
+                upgrade_storage_s3(progress, attachment, style)
+              when :fog
+                upgrade_storage_fog(progress, attachment, style)
+              when :filesystem
+                upgrade_storage_filesystem(progress, attachment, style)
+              end
+
+              progress.increment
+            end
+
+            attachment.instance_write(:storage_schema_version, CURRENT_STORAGE_SCHEMA_VERSION)
+          end
+
+          if record.changed?
+            record.save unless dry_run?
+            records += 1
+          end
+        end
+      end
+
+      progress.total = progress.progress
+      progress.finish
+
+      say("Upgraded storage schema of #{records} records#{dry_run}", :green, true)
+    end
+
+    private
+
+    def upgrade_storage_s3(progress, attachment, style)
+      previous_storage_schema_version = attachment.storage_schema_version
+      object                          = attachment.s3_object(style)
+
+      attachment.instance_write(:storage_schema_version, CURRENT_STORAGE_SCHEMA_VERSION)
+
+      upgraded_path = attachment.path(style)
+
+      if upgraded_path != object.key && object.exists?
+        progress.log("Moving #{object.key} to #{upgraded_path}") if options[:verbose]
+
+        begin
+          object.move_to(upgraded_path) unless dry_run?
+        rescue => e
+          progress.log(pastel.red("Error processing #{object.key}: #{e}"))
+        end
+      end
+
+      # Because we move files style-by-style, it's important to restore
+      # previous version at the end. The upgrade will be recorded after
+      # all styles are updated
+      attachment.instance_write(:storage_schema_version, previous_storage_schema_version)
+    end
+
+    def upgrade_storage_fog(_progress, _attachment, _style)
+      say('The fog storage driver is not supported for this operation at this time', :red)
+      exit(1)
+    end
+
+    def upgrade_storage_filesystem(progress, attachment, style)
+      previous_storage_schema_version = attachment.storage_schema_version
+      previous_path                   = attachment.path(style)
+
+      attachment.instance_write(:storage_schema_version, CURRENT_STORAGE_SCHEMA_VERSION)
+
+      upgraded_path = attachment.path(style)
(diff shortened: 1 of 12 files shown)


lib/paperclip/attachment_extensions.rb
```diff
@@ -14,6 +14,15 @@ def process_style?(style_name, style_args)
...
 
+    def storage_schema_version
+      instance_read(:storage_schema_version) || 0
+    end
+
+    def assign_attributes
+      super
+      instance_write(:storage_schema_version, 1)
+    end
+
```
ThibG, line 25: I'm not super comfortable with that part, it looks a bit hackish to me and I'm not sure about the implications.
I guess it *should* work though.

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,99 @@
...
+
+                if upgraded_path != previous_path
+                  FileUtils.mkdir_p(File.dirname(upgraded_path)) unless options[:dry_run]
+
+                  begin
+                    FileUtils.mv(previous_path, upgraded_path) unless options[:dry_run]
+                    upgraded += 1
+                  rescue Errno::ENOENT
+                    # The original file didn't exist for some reason
+                  end
```
ThibG, line 78: This potentially leaves empty directories.

ClearlyClaire (approved):
This is a very welcome change overall, although I have some minor concerns (see inline comments).

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,99 @@
...
+                upgraded_path = attachment.path(style)
+
+                if upgraded_path != object.key && object.exists?
+                  begin
+                    object.move_to(upgraded_path) unless options[:dry_run]
+                    upgraded += 1
+                  rescue => e
+                    progress.log(pastel.red("Error processing #{object.key}: #{e}"))
+                  end
+                end
```
noellabo, line 61: - If `object.move_to` fails, you should roll back `storage_schema_version`.
- If dry_run is specified, the `storage_schema_version` should not be updated.

lib/paperclip/attachment_extensions.rb
Gargron: The alternative approach is adding a `before_post_process` hook to every model (or one of the concerns), but this is quite neat in that `instance_write` doesn't do anything if the model doesn't have the relevant column. Also using `super` is a bit more clean than copy-and-pasting the original method in full, but it's understandably a bit hackick because the whole thing is a monkey-patch.

lib/mastodon/upgrade_cli.rb
Gargron: I need to think of how to clean it up. Just removing the directory isn't enough because (with `-p`) a deep directory structure is created and you need to recursively delete upwards but only if the directories are empty.

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,145 @@
...
+      # all styles are updated
+      attachment.instance_write(:storage_schema_version, previous_storage_schema_version)
+    end
+
+    def clean_up_empty_directories(path)
+      loop do
+        path = File.dirname(path)
+        FileUtils.rmdir(path)
+        break if File.exist?(path)
+      end
```
ThibG, line 140: `FileUtils.rmdir(path, parents: true)`?

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,145 @@
...
+
+        begin
+          unless dry_run?
+            FileUtils.mkdir_p(File.dirname(upgraded_path))
+            FileUtils.mv(previous_path, upgraded_path)
+          end
+        rescue => e
+          progress.log(pastel.red("Error processing #{previous_path}: #{e}"))
+          clean_up_empty_directories(upgraded_path) unless dry_run?
+        end
```
ThibG, line 126: empty directories should be cleaned up as well for the older path on successful migration

ClearlyClaire (changes requested):

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,136 @@
...
+      upgraded_path = attachment.path(style)
+
+      if upgraded_path != previous_path && File.exist?(previous_path)
+        progress.log("Moving #{previous_path} to #{upgraded_path}") if options[:verbose]
+
+        begin
+          unless dry_run?
+            FileUtils.mkdir_p(File.dirname(upgraded_path))
+            FileUtils.mv(previous_path, upgraded_path)
+            FileUtils.rmdir(previous_path, parents: true)
```
ThibG, line 122: I think `Errno::ENOTEMPTY` needs to be caught and ignored in any case.

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,136 @@
...
+        rescue => e
+          progress.log(pastel.red("Error processing #{previous_path}: #{e}"))
+          FileUtils.rmdir(upgraded_path, parents: true) unless dry_run?
+        end
+      end
+
+      # Because we move files style-by-style, it's important to restore
+      # previous version at the end. The upgrade will be recorded after
+      # all styles are updated
+      attachment.instance_write(:storage_schema_version, previous_storage_schema_version)
```
ThibG, line 133: Same, I think `Errno::ENOTEMPTY` needs to be caught and ignored in any case.

ClearlyClaire (changes requested):

lib/mastodon/upgrade_cli.rb
Gargron: Did you comment on the wrong line? I don't see why you would rescue it here.

ThibG: uh yeah i did :/

lib/mastodon/upgrade_cli.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        begin
+          unless dry_run?
+            FileUtils.mkdir_p(File.dirname(upgraded_path))
+            FileUtils.mv(previous_path, upgraded_path)
+            FileUtils.rmdir(previous_path, parents: true)
+          end
+        rescue => e
+          progress.log(pastel.red("Error processing #{previous_path}: #{e}"))
+          FileUtils.rmdir(upgraded_path, parents: true) unless dry_run?
```
ThibG, line 126: I was talking about this line, sorry

ClearlyClaire (changes requested):

ClearlyClaire (approved):

zunda:
This seems to have broken path to existing icons and custom emojis. E.g., my avatar which has been t https://s3.amazonaws.com/zundan-mastodon/accounts/avatars/000/000/001/original/89de63b40ca41c7e.gif is now rendered as https://s3.amazonaws.com/zundan-mastodon/system/accounts/avatars/000/000/001/original/89de63b40ca41c7e.gif_prefixaccounts/avatars/000/000/001/original/89de63b40ca41c7e.gif

`git revert c3ca3801f` fixed the problem for now. Not sure if related or not, my server is on ruby-2.7.1, makes direct use of S3 in us-east-1.

mayaeh:
> This seems to have broken path to existing icons and custom emojis.

This issue also occurs in my environment using Ruby 2.6.6 with OpenStack Swift.

mayaeh:
[This](https://github.com/mayaeh/mastodon/commit/d8d88895542468b3cd5352d18cd2458455b72b53) seems to fix it, but I don't understand why this problem occurs...

orlea:
Is it the same as the old issue below?
#6675
#12382

mashirozx:
Do you have any idea how to move the files from the original path to the cache path? I just tried `tootctl upgrade storage-schema`, which result in 404 errors. The media files are shown in a /cache/ URL, but the files still in their original path.

For example: https://hello.2heng.xin/web/statuses/104165895186766264

The media is shown as (404): https://s3-ap-shanghai.2heng.xin/aws_cached/mstd-1253826990/cache/media_attachments/files/000/065/646/original/4a81fedd03520d1b.png

But actually, they are still stored in:
https://s3-ap-shanghai.2heng.xin/aws_cached/mstd-1253826990/media_attachments/files/000/065/646/original/4a81fedd03520d1b.png

PR #12831, opened 2020-01-12
Fix Ruby 2.7 support

### LINKED ISSUE #12823: `http` 3.4 not compatible with Ruby 2.7
The `http` ruby gem is only compatible with Ruby 2.7 starting from 4.3.0.

This requires:
- a new release of the `goldfinger` gem (or using the git version)
- dropping the `ostatus2` gem (#12822), or updating it to use version 4.x of the `http` gem
- removing the `:per_operation` parameter of `timeout` in `app/lib/request.rb`

ClearlyClaire:
Fix #12823 

- update http gem to avoid errors
- update blurhash gem to avoid shared object loading error
- update goldfinger gem so the http gem could be updated
- update json gem to avoid warnings


Gargron (approved):

Shleeble:
Weird.. it's reporting this failed
https://circleci.com/gh/tootsuite/mastodon/110466 Failed

PR #12832, opened 2020-01-12
Drop support for Ruby 2.4/2.5

shleeable:


ThibG:
Is there a need to do that? Dropping 2.4 kinda makes sense, even though it hasn't reached EOL yet, at least to update Rails, but why would we drop 2.5?

PR #12835, opened 2020-01-12
Add option to keep evidence when suspending accounts

### LINKED ISSUE #547: Admin-side "Secure Evidence" option when banning users
After the recent couple of incidents of troll "raids" on a Mastodon instance I figured it might be useful to strengthen Mastodon's ability to collate evidence of illegal conduct (doxing, stalking, criminal harassment or hate speech, uploading illegal media etc.) in case this is needed for a handover to law enforcement agencies. 

Currently, banning an account wipes all its data from the database. While this is a sensible approach to take to protect server admins from inadvertently hosting illegal content, it also means that in the case of a request for information from law enforcement or if an admin wants to report illegal conduct themselves, they need to collate evidence manually, and if this isn't done there will be little evidence if e.g. a user requests an abuser be reported to law enforcement but the admin also wants to ban their account to protect their community.

Ideally it should be possible to collect evidence without requiring admins to sift through accounts potentially containing gore/shock images, hate speech etc. as this can have serious effects on their mental health (see e.g. https://www.theguardian.com/technology/2017/jan/11/microsoft-employees-child-abuse-lawsuit-ptsd).

With a few years of experience in working in child protection for an online service aimed at teenagers, I would propose a button or tickbox when banning a user to secure or collate the account's data in external files before wiping it from the database.

Ideally these files should NOT be stored on the server for obvious reasons. A secure approach to this would be to extract the banned user's data into memory or temporary files, pack it all up into a password-protected encrypted archive (ideally zip as this will be widely readable by law enforcement officers) and allow immediate download by the admin, displaying the password once and deleting the data from the server as soon as the admin confirms they have downloaded the file and noted down the password.

An admin could download the file directly to a USB stick for personal handover or temporarily save it to their computer to e-mail it to law enforcement without risking having potentially illegal content on their computer "in the clear", while at the same time the archive will make the collected evidence easier to handle by both the admin and LEOs.

To make the data more useful and readable, I would suggest structuring it as follows if possible (as a side note, something like this would also be a useful format for users requesting a backup of their data, e.g. if they choose to leave the service):

- profile picture and banner image
- a plain text file with account information - IP address on creation, creation timestamp, signup e-mail address, current profile information, bio, display name, current password hash + hashing info (cipher used and salt string)
- text posts as a CSV file with columns for timestamp, the user's current IP address at time of posting (if this is available), the actual post text, the post visibility setting, and any associated media IDs.
- media uploads as separate files in a dedicated subdirectory, with filenames matching their server-side media IDs so post content can be easily linked to files
- Any further information that might be useful for law enforcement to identify a user and their activity, e.g. timestamps of profile edits if available, browser User-Agent strings, etc.

As this is a lot of detail and includes sensitive information this should *never* be handed over to users but only to legitimate law enforcement agencies/officers, but it would be extremely helpful in case there ever is conduct that requires LEA involvement.

Gargron:
Fix #547

When selected, before the account's data is removed, some of it is denormalized into a separate, symmetrically-encrypted table. In particular:

- The e-mail
- All IPs used to access the account
- SHA256 fingerprints of all uploaded files
- URIs of accounts followed by or following the account
- URIs of accounts that were invited

___

This option has a cost in that it may conflict with the GDPR. It also has a technical cost in that retrieving and processing this data takes time and delays the removal of the data (in particular, uploaded files). For this reason, it is not checked by default.

### DIFF
--- a/app/controllers/admin/account_actions_controller.rb
+++ b/app/controllers/admin/account_actions_controller.rb
@@ -30,7 +30,7 @@ module Admin
     end
 
     def resource_params
-      params.require(:admin_account_action).permit(:type, :report_id, :warning_preset_id, :text, :send_email_notification, :include_statuses)
+      params.require(:admin_account_action).permit(:type, :report_id, :warning_preset_id, :text, :send_email_notification, :include_statuses, :create_account_summary)
     end
   end
 end
--- a/app/controllers/api/v1/admin/account_actions_controller.rb
+++ b/app/controllers/api/v1/admin/account_actions_controller.rb
@@ -26,7 +26,8 @@ class Api::V1::Admin::AccountActionsController < Api::BaseController
       :report_id,
       :warning_preset_id,
       :text,
-      :send_email_notification
+      :send_email_notification,
+      :create_account_summary
     )
   end
 end
--- a/app/controllers/settings/deletes_controller.rb
+++ b/app/controllers/settings/deletes_controller.rb
@@ -46,7 +46,7 @@ class Settings::DeletesController < Settings::BaseController
 
   def destroy_account!
     current_account.suspend!
-    Admin::SuspensionWorker.perform_async(current_user.account_id, true)
+    Admin::SuspensionWorker.perform_async(current_user.account_id, reserve_email: false)
     sign_out
   end
 end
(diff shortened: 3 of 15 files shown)


ClearlyClaire (changes requested):
I understand this is incomplete, as there is nothing to actually make use of that data yet, but I'm concerned this data could be kept longer than needed, or even forever. It should be obvious how to delete it, that it should be deleted at some point, and possibly it should auto-delete after a set time.

Also, regarding deletion, deleting rows may bypass the unique restrictions on IVs for an attacker that would have multiple snapshots of the database.

PR #12856, opened 2020-01-13
Document AUTHORIZED_FETCH mode and WHITELIST_MODE

ClearlyClaire:


Gargron:
Should this maybe be part of documentation instead? Since .env.production is auto-generated from mastodon:setup, a lot of people probably don't look into the sample file like they used to. It feels weird to document functionality in bash comments in an env file.

ThibG:
Well, summarily describing flags in the config file sample is helpful and common practice.
But I guess a shorter description of what they do *and* a proper entry in the documentation would be best.

Gargron:
A link to the anchor in documentation would be perfect, actually

ThibG:
Updated by shortening significantly the description and linking to the docs

Gargron (approved):

PR #12863, opened 2020-01-13
Add limit to how far back in time anonymous users can browse public TLs

### LINKED ISSUE #12858: Limit unauthenticated access to timelines
Please limit the timeline data returned by the API for unauthenticated requests. I can't think of a reason why an unauthenticated client should need more than the newest 50 or so statuses.

### Pitch

The server should never return any post older than the 50eth most recent to an unauthenticated client.

### Motivation

People are using the API to scrape all local public posts of instances. This is in most cases not wanted.

This is how researchers at the university of Milan did it:

> The spider exploits the instance list obtained from the previous step and makes a pool of requests to the instance endpoint which returns the latest toots of the local timeline. Since the timelines implement a pagination mechanism, the spider extracts the URL for the next request and repeat this procedure till it reaches the end of the timeline.

<https://aaai.org/ojs/index.php/ICWSM/article/download/3262/3130/>

ClearlyClaire:
Fixes #12858

This limits unauthenticated users to viewing up to the 40 newest toots
in the public TL + any newer than 2 hours ago.

### DIFF
--- a/app/controllers/api/v1/timelines/public_controller.rb
+++ b/app/controllers/api/v1/timelines/public_controller.rb
@@ -26,9 +26,22 @@ class Api::V1::Timelines::PublicController < Api::BaseController
   end
 
   def public_statuses
+    options = params_slice(:max_id, :since_id, :min_id)
+
+    unless current_user&.functional?
+      # Limit how far back unauthenticated users can read the public timeline
+      cutoff = (2.hours.ago.utc.to_i * 1000) << 16
+
+      # we're supposed to return toots immediately after `min_id`, but it's
+      # too far back, return nothing instead
+      return [] if (options[:min_id] || cutoff) < cutoff
+
+      options[:since_id] = [cutoff, options[:since_id] || 0].max if options[:max_id].present?
+    end
+
     statuses = public_timeline_statuses.paginate_by_id(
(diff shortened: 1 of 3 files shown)


ykzts (approved):

Gargron:
It might lead to visitors assuming there is nothing older on the network. How about showing a message like "There is more, but you need to sign up to see it" when the limit is reached?

ThibG:
Added the following message, which appears only when there's nothing more to load, and no user is logged on.

image

busshimoe:
Is there an option to disable this? I do not want this enabled on my instance.

Gargron:
@busshimoe This is not planned for 3.1.0, so there is time to think about this.

ThibG:
I still think this could be useful, even though it's only a very partial solution (to scrape, one would have to do that in realtime, which is definitely possible, but more difficult to do across a lot of instances).

umonaca:
I think it is useful too. I have seen some instances implemented this feature with other method.

umonaca:
This PR is useful for some minority groups worried about content security. Please reopen this issue.

ThisIsMissEm:
> Is there an option to disable this? I do not want this enabled on my instance.

@busshimoe could you clarify your comment here? Do you want to disable timeline previews full stop?

PR #12877, opened 2020-01-17
Close registration before self-destruct

tribela:
To prevent new join in self-destruct sequence is running, Close registration first.


Gargron (approved):

Shleeble:
Can you run the test again?

PR #12905, opened 2020-01-20
Add support for magnet: URIs

ClearlyClaire:
Generalize the XMPP URI parser function a bit and add magnet: URI support

### DIFF
--- a/app/lib/formatter.rb
+++ b/app/lib/formatter.rb
@@ -245,9 +245,9 @@ class Formatter
     end
 
     standard = Extractor.extract_entities_with_indices(text, options)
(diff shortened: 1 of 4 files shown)


wakest:
can this same thing be used for getting the other half of dat:// links working? currently only dat://domain.tld links work but dat://hash238u234892348 links do not.

ThibG:
Is there a spec for dat: URIs? I couldn't find any

Gargron (approved):

wakest:
@ThibG not sure how much more you need but there is: [DNS + DAT](https://beakerbrowser.com/docs/guides/use-a-domain-name-with-dat) and this also might help: [Parse DAT URL](https://github.com/pfrazee/parse-dat-url)

PR #12906, opened 2020-01-20
Upgrade .nvmrc to Node.js 12

mkody:
This should remove engine errors since #12791 by setting the latest LTS.


Gargron (approved):

Shleeble:
My commit pushed to nodejs 10. Do we want to push to packages to 12?

mkody:
Node.js 10 is still active and will reach its EOL in 2021. It should be fine to let users run that version until then.

I jumped right to 12 since it works and so that we (probably) don't need to upgrade to an another major version until April 2022.

Shleeble:
works for me.

PR #12926, opened 2020-01-23
Fixing the hero widget

koyuawsmbrtn:
Before:
before

After:
after

### DIFF
--- a/app/javascript/styles/mastodon-light/diff.scss
+++ b/app/javascript/styles/mastodon-light/diff.scss
@@ -675,7 +675,6 @@ html {
   .public-account-bio,
   .hero-widget__text {
     background: $account-background-color;
-    border: 1px solid lighten($ui-base-color, 8%);
   }
 
   .header {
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron (changes requested):
Please, instead of this, remove the border between those two elements in the light theme. Without the border, the spacing between the text elements is just right.

app/javascript/styles/mastodon-light/diff.scss
```diff
@@ -675,7 +675,6 @@ html {
   .public-account-bio,
   .hero-widget__text {
     background: $account-background-color;
-    border: 1px solid lighten($ui-base-color, 8%);
```
Gargron: You just need to set `border-bottom: 0`, I think. Just one side, not all of them.

Gargron (changes requested):

app/javascript/styles/mastodon-light/diff.scss
koyuawsmbrtn: This seems rather hacky and I'm not sure if it will even have an effect. On my side it kinda seems to want the whole border removed as the border styling has been done something like this

image

So no matter what I do removing the border completely does the job

SuperSandro2000:
bump

SuperSandro2000:
Bump

PR #12927, opened 2020-01-23
Change followers page to relationships page in admin UI

Gargron:
Allow browsing and filtering all relationships instead of just followers, unify the codebase with the user-facing relationship manager, add ability to see who the user invited


ykzts (approved):

Sasha-Sorokin:
It feels like this PR is broken? I wanted to check context, started a test instance (HEAD is c0006a0) and any filter shows only admin. I also invited a user with test account, completed registration using the invite, but it is still the same. And this all also applies to user's relationships (in user settings), it only shows admin there as well.

Screenshot of the demo account's relationships page with Invited filter selected, but despite showing admin account, which is clearly not who demo account would invite

PR #12995, opened 2020-01-28
Fix Gemfile.lock

ClearlyClaire:


hinaloe:
bundler said that `--deployment was deprecated` so should use `bundle config deployment`

Shleeble:
lgtm. Sorry again.

Gargron (approved):

PR #13012, opened 2020-01-31
Add JS IDE helper

Grawl:
In JetBrains IDE, like WebStorm or RubyMine, this import is not recognized as valid import:

```js
import Icon from 'mastodon/components/icon';
```

IDE [has Webpack support](https://www.jetbrains.com/help/idea/using-webpack.html#) but I don't know a way to set it up with complex Webpack configurations such as one used in this repository.

To let IDE know where is 'mastodon' alias is located, there is a trick: write fake Webpack config just for IDE, provide used aliases in form IDE understand, and select this file as Webpack config in IDE settings.

### DIFF
--- a/ide-helper.js
+++ b/ide-helper.js
@@ -1,0 +1,11 @@
+/*
+Preferences | Languages & Frameworks | JavaScript | Webpack | webpack configuration file
+jetbrains://WebStorm/settings?name=Languages+%26+Frameworks--JavaScript--Webpack
+*/
+module.exports = {
+  resolve: {
+    alias: {
+      'mastodon': path.resolve(__dirname, 'app/javascript/mastodon'),
+    }
+  }
+}
--- a/ide-helper.js
+++ b/ide-helper.js
@@ -1,3 +1,4 @@
+/* global path */
 /*
 Preferences | Languages & Frameworks | JavaScript | Webpack | webpack configuration file
 jetbrains://WebStorm/settings?name=Languages+%26+Frameworks--JavaScript--Webpack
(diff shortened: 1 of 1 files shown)


ide-helper.js
```diff
@@ -0,0 +1,11 @@
+/*
+Preferences | Languages & Frameworks | JavaScript | Webpack | webpack configuration file
+jetbrains://WebStorm/settings?name=Languages+%26+Frameworks--JavaScript--Webpack
+*/
+module.exports = {
+  resolve: {
+    alias: {
+      'mastodon': path.resolve(__dirname, 'app/javascript/mastodon'),
```
Sasha-Sorokin: CodeClimate raises an issue here that `path` is not defined, you probablly should add `const path = require('path')` at the top of the file.

Grawl: since it's ESLint error no need to require it, just let linter know it's Node global function

Grawl:
oh there is a new linter errors

Grawl:
> Your bundle is locked to httplog (1.4.0), but that version could not be found in
any of the sources listed in your Gemfile

I have the same error when I trying build Mastodon in Docker locally on Fedora 31

Sasha-Sorokin:
Try rebasing onto `upstream/master`? During rebase you can also squash your commits.

Grawl:
Can you squash on merge? I prefer to not squash on local.

Sasha-Sorokin:
Mastodon maintainers [always squash commits on merge](https://github.com/tootsuite/mastodon/blob/master/CONTRIBUTING.md#pull-requests). Squashing manually just allows to preserve sane commit message, automatic one can be misleading at times, but that's entirely one's choice.

Rebase worked, dependencies are installed and testing has just started.

Gargron (approved):

PR #13016, opened 2020-01-31
Search account domain in lowercase

abcang:
I noticed that the validation query before saving the remote account did not work with the index.

```sql
SELECT  1 AS one FROM "accounts" WHERE "accounts"."username" = 'abcang' AND "accounts"."domain" = 'pawoo.net' LIMIT 1
```

```sql
Limit  (cost=0.00..12043.92 rows=1 width=0) (actual time=336.947..336.948 rows=1 loops=1)
  ->  Seq Scan on accounts  (cost=0.00..12043.92 rows=1 width=0) (actual time=336.945..336.945 rows=1 loops=1)
        Filter: (((username)::text = 'abcang'::text) AND ((domain)::text = 'pawoo.net'::text))
        Rows Removed by Filter: 34062
Planning time: 0.191 ms
Execution time: 337.147 ms
```

Since the index of the account is as follows, it is necessary to compare it with `lower("accounts"." username")` and `lower("accounts"."domain")` to use index.

```rb
t.index "lower((username)::text), lower((domain)::text)", name: "index_accounts_on_username_and_domain_lower", unique: true
```

As a result of the fix, the index is now used correctly.

```sql
SELECT  1 AS one FROM "accounts" WHERE LOWER("accounts"."username") = 'abcang' AND LOWER("accounts"."domain") = 'pawoo.net' LIMIT 1
```

```
Limit  (cost=0.41..8.43 rows=1 width=0) (actual time=0.047..0.047 rows=1 loops=1)
  ->  Index Scan using index_accounts_on_username_and_domain_lower on accounts  (cost=0.41..8.43 rows=1 width=0) (actual time=0.046..0.046 rows=1 loops=1)
        Index Cond: ((lower((username)::text) = 'abcang'::text) AND (lower((domain)::text) = 'pawoo.net'::text))
Planning time: 5.180 ms
Execution time: 0.073 ms
```

Also, the remote account is validated with `case_sensitive: true`, but the actual unique index is not case-sensitive, so this was also fixed.

### DIFF
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -48,7 +48,7 @@ module AccountFinderConcern
     end
 
     def with_usernames
-      Account.where.not(username: '')
+      Account.where(Account.arel_table[:username].lower.not_eq '')
     end
 
     def matching_username
@@ -56,11 +56,7 @@ module AccountFinderConcern
     end
 
     def matching_domain
-      if domain.nil?
-        Account.where(domain: nil)
-      else
-        Account.where(Account.arel_table[:domain].lower.eq domain.to_s.downcase)
-      end
+      Account.where(Account.arel_table[:domain].lower.eq(domain.nil? ? nil : domain.to_s.downcase))
     end
   end
 end
--- a/app/models/concerns/account_finder_concern.rb
+++ b/app/models/concerns/account_finder_concern.rb
@@ -48,7 +48,7 @@ module AccountFinderConcern
     end
 
     def with_usernames
-      Account.where(Account.arel_table[:username].lower.not_eq '')
+      Account.where.not(Account.arel_table[:username].lower.eq '')
     end
 
     def matching_username
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -619,18 +619,18 @@ RSpec.describe Account, type: :model do
     end
 
     context 'when is remote' do
-      it 'is invalid if the username is not unique in case-sensitive comparison among accounts in the same normalized domain' do
+      it 'is invalid if the username is same among accounts in the same normalized domain' do
         Fabricate(:account, domain: 'にゃん', username: 'username')
         account = Fabricate.build(:account, domain: 'xn--r9j5b5b', username: 'username')
         account.valid?
         expect(account).to model_have_error_on_field(:username)
       end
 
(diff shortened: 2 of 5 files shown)


app/models/concerns/account_finder_concern.rb
```diff
@@ -48,19 +48,15 @@ def scoped_accounts
...
       Account.where(Account.arel_table[:username].lower.eq username.to_s.downcase)
     end
 
     def matching_domain
-      if domain.nil?
-        Account.where(domain: nil)
-      else
-        Account.where(Account.arel_table[:domain].lower.eq domain.to_s.downcase)
-      end
+      Account.where(Account.arel_table[:domain].lower.eq(domain.nil? ? nil : domain.to_s.downcase))
```
abcang: Even if the comparison target is nil, comparison with `lower("accounts"."domain")` is required to use the index.

Gargron (approved):

spec/models/account_spec.rb
```diff
@@ -619,18 +619,18 @@
...
         expect(account).to model_have_error_on_field(:username)
       end
 
-      it 'is valid even if the username is unique only in case-sensitive comparison among accounts in the same normalized domain' do
+      it 'is invalid if the username is not unique in case-insensitive comparison among accounts in the same normalized domain' do
         Fabricate(:account, domain: 'にゃん', username: 'username')
         account = Fabricate.build(:account, domain: 'xn--r9j5b5b', username: 'Username')
         account.valid?
-        expect(account).not_to model_have_error_on_field(:username)
+        expect(account).to model_have_error_on_field(:username)
```
abcang: In fact, this test was wrong. When saving the account without validating the username, an error occurred due to unique constraint of DB.

ClearlyClaire (commented):
Seems ok with current practices, but technically, the local part of acct URIs is supposed to be case-sensitive.

Gargron:
> technically, the local part of acct URIs is supposed to be case-sensitive.

I don't think that's true? `alice` is not supposed to be available when `Alice` exists, in fact I remember us having some issues and a migration to fix them around this.

ThibG:
@Gargron the specs say that the local part of acct: URIs are supposed to be case-sensitive. It's just that in Mastodon we explicitly decided to go against that, to accommodate with people changing the casing of already-existing accounts manually, iirc

Issue #13017, opened 2020-01-31
Mark announcement as read


renatolond:
### Pitch

With the new announcements feature, every time I reopen the interface, I get the number of active announcements highlighted. I think it would be nice for it to remember the announcements I already read instead.

### Motivation

It seems like there's a new announcement when there's not. It might give notification fatigue and users stop reading announcements.

PR #13020, opened 2020-02-01
Change how unread announcements are handled

ClearlyClaire:
- Change the effect of the “dismiss” API call to mark announcements as read instead of hiding them
- Display the badge with count of unread announcements even when the announcement box is open
- Display an “unread” or “new” marker the first time an announcement is shown
- Mark announcements as shown when they are displayed

Peek 01-02-2020 16-27


ThibG:
This is to fix #13017

Shleeble:
lgtm :)

Gargron (approved):

PR #13021, opened 2020-02-01
Update Twemoji to 12.1.5

DeeUnderscore:
## Twemoji
This updates Twemoji with the SVGs from the [Twemoji repo](https://github.com/twitter/twemoji) at `v12.1.5`.

Previous version in Mastodon was 12.1.3, so the most noticeable change with this one is the inclusion of gender-inclusive/gender-neutral designs, so that you'll see a difference between, for example, 🙋 without gender modifiers and 🙋‍♀️, with gender modifier (you won't see that difference here on the Github website).

`emoji_map.json` was also updated to Unicode 12, from 11.

## Emoji picker
I did not update the emoji picker, as that's an external dependency. While the spritesheet that the picker uses is distributed in this repo, the data for the picker comes with the picker package ([Gargron/emoji-mart](https://github.com/Gargron/emoji-mart)).

New data could be grabbed from the [upstream emoji-mart repo](https://github.com/missive/emoji-mart/tree/master/data), while new sheets could be grabbed from [iamcal/emoji-data](https://github.com/iamcal/emoji-data), which is what emoji-mart uses. 

This is something that should probably be eventually updated, especially after the Unicode 13 release and subsequent Twemoji updates, which will happen sometime starting in March.


moritzheiber:
I just noticed Mastodon doesn't have gender-neutral emoji yet .. which even the iPhone has now. How about merging this?

ykzts (approved):

Gargron:
Okay, I reproduced all changes, all clean.

PR #13031, opened 2020-02-03
Change how multiple choice poll percentages are calculated in web UI

Gargron:
Instead of calculating as `votes` / `unique people who voted`, calculate as `votes` / `total votes`, which is how it should have been all along and prevent issues where percentages don't add up to 100%.

Fix focused state erasing background of selected poll options

### DIFF
--- a/app/javascript/mastodon/components/poll.js
+++ b/app/javascript/mastodon/components/poll.js
@@ -113,7 +113,7 @@ class Poll extends ImmutablePureComponent {
 
   renderOption (option, optionIndex, showResults) {
     const { poll, disabled, intl } = this.props;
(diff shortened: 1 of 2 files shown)


ykzts (approved):

Gargron:
Wait, I messed up the commit message.

ThibG:
what?

trwnh:
Multiple choice polls are actually a series of yes/no boolean questions, so the only constraint is that each option has between 0-100%. It's misleading to use total votes instead of unique voters, although maybe I can see situations where the total share of votes might be interesting or relevant information? But in general I would be interested in percent voters, not percent votes.

PR #13037, opened 2020-02-03
Fix how links with unhandled schemes are rendered

ClearlyClaire:
Currently, if Mastodon encounters a link to a scheme it doesn't allow, it will strip the `href` attribute, making it unclickable, but link will *appear* clickable and, if formatted as Mastodon formats it, will not appear in full.

This PR changes it so that link styling as well as `ellipsis` and `invisible` classes only apply to links with a defined `href`.

For instance, without magnet support, Mastodon still show Mastodon-authored magnet links as follows:
Peek 03-02-2020 21-34

With this PR, they would show up like this (if they were still unsupported):
Screenshot_2020-02-03 Mastodon(1)

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -290,27 +290,29 @@
   position: absolute;
 }
 
-.invisible {
-  font-size: 0;
-  line-height: 0;
-  display: inline-block;
-  width: 0;
-  height: 0;
-  position: absolute;
+a[href] {
+ .invisible {
+    font-size: 0;
+    line-height: 0;
+    display: inline-block;
+    width: 0;
+    height: 0;
+    position: absolute;
 
-  img,
-  svg {
-    margin: 0 !important;
-    border: 0 !important;
-    padding: 0 !important;
-    width: 0 !important;
-    height: 0 !important;
+    img,
+    svg {
+      margin: 0 !important;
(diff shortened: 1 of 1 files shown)


Gargron:
Okay but, what purpose do `<a>` tags with no `href` serve? Should we not just render them as plain text? They're gonna have the different color and hover/focus effect but will do literally nothing when clicked?

ThibG:
It's not really about the purpose of `<a>` tags with no `href`. It's about Mastodon already producing them when encountering a remote toot with a link to an unsupported scheme.

When Mastodon 3.0.1 receives an xmpp or magnet link, it strips the `href` due to `Sanitize`. This produces a link that have the different color and hover/focus effect but do nothing when clicked.

That's what this PR fixes by applying the color and hover/focus effects only to `<a>` tags with a `href`. It also only applies `invisible` and `ellipsis` tags to only those, because so far, those can be applied outside of links too.

Stripping the `a` entirely is another option, but after a quick look at `Sanitize`'s documentation I couldn't find an easy way to do so. Furthermore, with `invisible` and `ellipsis` applying outside of links, it would have still displayed the shortened version, even without rendering links at all.

Gargron:
@ThibG I imagine stripped href-less anchors works the same way as our whitelisting of HTML elements, just check if the node has a href, if not, convert it to text.

PR #13039, opened 2020-02-03
Fix typos in nl translation of setting_crop_images

zopsicle:
There were two mistakes in this translation:

  1. The translation of to crop is bijsnijden. While besnijden is a Dutch word,
     its meaning is left as an exercise to the reader.

  2. To negate an adjective, a hyphen is needed between “niet” and the adjective.
     See https://taaladvies.net/taal/advies/vraag/1843.

### DIFF
--- a/config/locales/simple_form.nl.yml
+++ b/config/locales/simple_form.nl.yml
@@ -125,7 +125,7 @@ nl:
         setting_aggregate_reblogs: Boosts in tijdlijnen groeperen
         setting_auto_play_gif: Speel geanimeerde GIF's automatisch af
         setting_boost_modal: Vraag voor het boosten van een toot een bevestiging
-        setting_crop_images: Afbeeldingen tot 16x9 besnijden in niet uitgebreide toots
+        setting_crop_images: Afbeeldingen tot 16x9 bijsnijden in niet-uitgebreide toots
         setting_default_language: Taal van jouw toots
         setting_default_privacy: Standaardzichtbaarheid van jouw toots
         setting_default_sensitive: Media altijd als gevoelig markeren


ThibG:
Please use crowdin for that, as translations are pulled from there: https://crowdin.com/translate/mastodon/

chloekek:
I could not find a way to use Crowdin without running unfree JavaScript code. I will see if the API is usable for that.

chloekek:
The readme says to use Weblate, which is free software. Crowdin seems to be the current tool, but it requires unfree JavaScript, so I cannot use it. May I ask why this switch was made, especially considering Mastodon is a free software project?

Sasha-Sorokin:
I'm not sure why are you so worried of ‘non-free’ JavaScript, whether it's privacy issue, you can use Tor browser with Crowdin (though JavaScript needs to be enabled, Tor ensures your privacy is protected and browsers' script language is unable to interact with your system) and sing up using temporary e-mail (but confirmed e-mail ensures security of account and ability to reset password in future).

Being *free software project* doesn't obligate you in picking other free software when choosing tools. The truth is we have to rely on proprietary software from time to time, for the comfort of users and developers (GitHub used to host the code doesn't seen as much of a problem, is it?). This is the world we are living in now.

From what I learned, Crowdin was chosen for a number of reasons such as more reliable built-in tools and due to some issues with Weblate. Plus to that, Crowdin grants free of charge license to open source projects and doesn't needs manual hosting.

<details>
<summary>Personal experience</summary>

To add, I myself found Weblate very unsatisfying to use, as example it has no AJAX support and each action reloads the page completely, having bad internet connection or being far away from the server makes it difficult to use. The UI is also hard the navigate, making up the horrible UX experience.
</details>

Merging a PR like this will is highly likely to introduce sync conflicts when new translations are ready and will need manual intervention, only the source translation files should be changed (English, USA) and as for translations, they have to be updated on Crowdin going trough through the full process of submitted-voted-approved so they kept right and consistent.

I hope that clears some points. And about README, it may needs to be updated, the localized badge as well as CONTRIBUTING both point to Crowdin, so Weblate mention was probably mistakenly left there.

chloekek:
Alright, then I will not contribute to Mastodon.

Sasha-Sorokin:
I'm genuinely confused whether I somehow seemed mean with that detailed text or there's another reason why.

Well, at least, can I submit your fix from my account, leaving a comment with your PR contents and attributing you?

PR #13040, opened 2020-02-04
Fix rendering `<a>` without `href` when scheme unsupported

Gargron:
Close #13037

This replaces an `<a>` that has no `href` with its children. But maybe we don't want any children elements like `<span>` to stick around?

### DIFF
--- a/app/lib/sanitize_config.rb
+++ b/app/lib/sanitize_config.rb
@@ -2,7 +2,23 @@
 
 class Sanitize
   module Config
-    HTTP_PROTOCOLS ||= ['http', 'https', 'dat', 'dweb', 'ipfs', 'ipns', 'ssb', 'gopher', 'xmpp', 'magnet', :relative].freeze
+    HTTP_PROTOCOLS = %w(
+      http
+      https
+    ).freeze
+
+    LINK_PROTOCOLS = %w(
+      http
+      https
+      dat
+      dweb
+      ipfs
+      ipns
+      ssb
+      gopher
+      xmpp
+      magnet
+    ).freeze
 
     CLASS_WHITELIST_TRANSFORMER = lambda do |env|
       node = env[:node]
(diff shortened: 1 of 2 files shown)


ClearlyClaire (changes requested):
If the link has Mastodon-style formatting, it will hide the `invisible` parts and apply `ellipsis`, making the link unusable. So there needs to be some extra processing, or the CSS needs to be reworked.

app/lib/sanitize_config.rb
```diff
@@ -50,13 +76,12 @@ module Config
         },
       },
 
-      protocols: {
-        'a' => { 'href' => HTTP_PROTOCOLS },
-      },
+      protocols: {},
```
ThibG: Won't it strip away whatever isn't supported by default?

ThibG:
I'm not *opposed* to removing the `<a>` tag entirely, but it seems more complicated and error-prone than the purely-CSS solution.

Also, a `<a>` tag with no `href` attribute is semantically fine according to the HTML 5.1 specs (it says it must not have a `target` attribute though).

app/lib/sanitize_config.rb
Gargron: `spec/lib/sanitize_config_spec.rb:42` shows that it doesn't. `protocols` isn't by itself a whitelist, merely a definition of filters to be used. Nothing defined, no filters used.

ThibG: ok, this indeed works, I thought the config inherited from Sanitize's default one, but it's not the case

app/lib/sanitize_config.rb
```diff
@@ -19,19 +19,45 @@ module Config
...
+
+          if uri.relative?
+            :relative
+          else
+            uri.scheme
+          end
+        end
+      rescue Addressable::URI::InvalidURIError
+        nil
+      end
```
ThibG: This whole thing is done without fully parsing URLs in Sanitize
cf. https://github.com/rgrove/sanitize/blob/master/lib/sanitize/transformers/clean_element.rb#L143

It's probably safer (in terms of possible changed behaviors) and more CPU-efficient to do it the same way.

Btw, I think we should either reject URLs or rewrite them to be relative to the original status (an option I prefer, but is maybe not worth it).

ClearlyClaire (approved):
seems good overall, except for the inlined nitpicks

also, I'm not sure it makes sense to allow all the supported protocols (which shouldn't be called `HTTP_PROTOCOLS` anymore) in `embed` and `iframe` tags

ClearlyClaire (approved):

PR #13042, opened 2020-02-04
Fix malformed HTML causing uncaught error

Gargron:
Fix OEmbed preview API leaking existence of private statuses (see #12930)


ThibG:
Do you have an example of what's causing the ArgumentError?

Gargron:
@ThibG I do not have a *specific* example, some status(es) seem to be going around that cause it but I cannot see which ones, however in theory it's just `'<div>' * 1000`

ClearlyClaire (approved):
Hm, I agree with the OEmbed part… the `reformat` part, I'm not too sure about… it seems weird to just have an empty status or bio, even if the original content is probably not good any way… but otherwise, LGTM

PR #13043, opened 2020-02-04
Add stoplight for object storage failures, return HTTP 503

Gargron:
Reduce traffic congestion due to timeouts

### DIFF
--- a/lib/paperclip/attachment_extensions.rb
+++ b/lib/paperclip/attachment_extensions.rb
@@ -39,6 +39,23 @@ module Paperclip
     def default_url(style_name = default_style)
       @url_generator.for_as_default(style_name)
     end
+
+    STOPLIGHT_THRESHOLD = 10
+    STOPLIGHT_COOLDOWN  = 30
+
+    # We overwrite this method to put a circuit breaker around
+    # calls to object storage, to stop hitting APIs that are slow
+    # to respond or don't respond at all and as such minimize the
+    # impact of object storage outages on application throughput
+    def save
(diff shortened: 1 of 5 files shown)


nightpool:
can we make this opt in through an environment variable? it seems bad if people on reliable storage providers would see increased unavailability due to transient network errors. 

i.e., if I'm using Digital Ocean Spaces, and there's a one off timeout, I don't want that to trigger a stoplight, because I know it's probably not indicative of a more serious failure.

Gargron:
@nightpool Threshold is 3 and cooldown is 60 seconds by default. Can we find a combination that you would be happy with without requiring extra configuration? Even a cooldown of 10 seconds would greatly reduce the amount of timing out concurrent requests.

lib/paperclip/attachment_extensions.rb
```diff
@@ -24,6 +24,39 @@ def reprocess_original!
...
+
+        process = only_process
+        @queued_for_write.except!(:original) if process.any? && !process.include?(:original)
+
+        flush_writes
+      end
+
+      @dirty = false
+      true
+    end
```
ThibG: Would it make sense to somehow call the upstream method entirely wrapped in `circuit_break!` instead of redefining it?

ThibG:
One thing that this could have enabled is receiving and processing statuses with media attachments when the object storage is down. However, I think the error handling in `ActivityPub::Activity::Create#process_attachments` would actually not catch it, and abort the whole job?

ClearlyClaire (approved):
Seems fine to me although I'm a bit worried about temporary storage server failures causing to silently avoid downloading media attachments/emoji. It's probably a bit more useful overall, but I'm afraid this would make the behavior slightly more surprising and slightly more difficult to debug.

PR #13049, opened 2020-02-07
Fix unfiltered params error when generating ActivityPub tag pagination

Gargron:
`ActionController::UnfilteredParameters (unable to convert unpermitted parameters to hash)`

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -24,7 +24,7 @@ class TagsController < ApplicationController
       format.rss do
         expires_in 0, public: true
 
-        @statuses = HashtagQueryService.new.call(@tag, params.slice(:any, :all, :none)).limit(PAGE_SIZE)
+        @statuses = HashtagQueryService.new.call(@tag, filter_params).limit(PAGE_SIZE)
         @statuses = cache_collection(@statuses, Status)
 
         render xml: RSS::TagSerializer.render(@tag, @statuses)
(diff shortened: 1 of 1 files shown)


app/controllers/tags_controller.rb
```diff
@@ -57,10 +57,14 @@ def set_instance_presenter
...
-      id: tag_url(@tag, params.slice(:any, :all, :none)),
+      id: tag_url(@tag, filter_params),
       type: :ordered,
       size: @tag.statuses.count,
       items: @statuses.map { |s| ActivityPub::TagManager.instance.uri_for(s) }
     )
   end
+
+  def filter_params
+    params.slice(:any, :all, :none).permit(:any, :all, :none)
```
ThibG: Why use both `slice` and `permit`? Isn't `permit` enough?

Gargron: Permit isn't enough when you want to convert to hash and the hash contains params that are not permitted.

ThibG: As far as I understand it, both `slice` and `permit` return an `ActionController::Parameters`, sliced to the params, with `permit` marking them as permitted while `slice` doesn't. So `slice` alone isn't good for that use case, but `permit` is?

Gargron:
```
[1] pry(main)> p = ActionController::Parameters.new(foo: '1', bar: '2')
=> <ActionController::Parameters {"foo"=>"1", "bar"=>"2"} permitted: false>
[2] pry(main)> p.permit(:foo).to_h
Unpermitted parameter: :bar
=> {"foo"=>"1"}```

ThibG: for some reason i can't reproduce this here, but ok

ClearlyClaire (approved):

PR #13054, opened 2020-02-09
Upgrade Dockerfile to use latest nodejs 12 LTS - 12.14 to 12.16

shleeable:
Upgrade to new LTS


shleeable:
I believe this can be merged any time. Tested.

PR #13067, opened 2020-02-09
Enhanced french traduction

guigeekz:


ThibG:
Thanks for your contributions! However, translation work should be done on https://crowdin.com/translate/mastodon/

PR #13068, opened 2020-02-09
Add submit button to the top of preferences pages

### LINKED ISSUE #11026: Move the save to top of the settings
image

### Pitch
Move the save button higher.

### Motivation
the damn button is hidden default and the options don't autosave.

guigeekz:
This is to fix [#11026](https://github.com/tootsuite/mastodon/issues/11026)

submit_button

In order to harmonize the forms, I move the save button to the top for :

- preferences/appearance
- preferences/notifications
- preferences/other
- settings/profile
- admin/settings/edit

### DIFF
--- a/app/views/settings/preferences/appearance/show.html.haml
+++ b/app/views/settings/preferences/appearance/show.html.haml
@@ -1,7 +1,10 @@
 - content_for :page_title do
   = t('settings.appearance')
 
-= simple_form_for current_user, url: settings_preferences_appearance_path, html: { method: :put } do |f|
+- content_for :heading_actions do
+  = button_tag t('generic.save_changes'), class: 'button', form: 'edit_user'
+
+= simple_form_for current_user, url: settings_preferences_appearance_path, html: { method: :put, id: 'edit_user' } do |f|
   .fields-row
     .fields-group.fields-row__column.fields-row__column-6
       = f.input :locale, collection: I18n.available_locales, wrapper: :with_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }, selected: I18n.locale, hint: false
@@ -56,6 +59,3 @@
 
   .fields-group
     = f.input :setting_expand_spoilers, as: :boolean, wrapper: :with_label
-
-  .actions
-    = f.button :button, t('generic.save_changes'), type: :submit
--- a/app/views/settings/preferences/appearance/show.html.haml
+++ b/app/views/settings/preferences/appearance/show.html.haml
@@ -59,3 +59,6 @@
 
   .fields-group
     = f.input :setting_expand_spoilers, as: :boolean, wrapper: :with_label
+
+  .actions
+    = f.button :button, t('generic.save_changes'), type: :submit
--- a/app/views/settings/preferences/other/show.html.haml
+++ b/app/views/settings/preferences/other/show.html.haml
@@ -35,3 +35,6 @@
 
   .fields-group
     = f.input :chosen_languages, collection: filterable_languages.sort, wrapper: :with_block_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }, required: false, as: :check_boxes, collection_wrapper_tag: 'ul', item_wrapper_tag: 'li'
+
+  .actions
+    = f.button :button, t('generic.save_changes'), type: :submit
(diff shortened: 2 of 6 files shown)


app/views/settings/preferences/appearance/show.html.haml
```diff
@@ -56,6 +59,3 @@
 
   .fields-group
     = f.input :setting_expand_spoilers, as: :boolean, wrapper: :with_label
-
-  .actions
-    = f.button :button, t('generic.save_changes'), type: :submit
```
Sasha-Sorokin: Appearance settings page has many elements and long enough. I think it is a good idea to leave another button at the bottom too.

app/views/settings/preferences/other/show.html.haml
```diff
@@ -32,6 +35,3 @@
 
   .fields-group
     = f.input :chosen_languages, collection: filterable_languages.sort, wrapper: :with_block_label, include_blank: false, label_method: lambda { |locale| human_locale(locale) }, required: false, as: :check_boxes, collection_wrapper_tag: 'ul', item_wrapper_tag: 'li'
-
-  .actions
-    = f.button :button, t('generic.save_changes'), type: :submit
```
Sasha-Sorokin: Other settings page has many elements too, probably okay to leave button here too.

app/views/settings/profiles/show.html.haml
```diff
@@ -52,9 +55,6 @@
           %input{ type: :text, maxlength: '999', spellcheck: 'false', readonly: 'true', value: link_to('Mastodon', ActivityPub::TagManager.instance.url_for(@account), rel: 'me').to_str }
         %button{ type: :button }= t('generic.copy')
 
-  .actions
-    = f.button :button, t('generic.save_changes'), type: :submit
```
Sasha-Sorokin: …same as with others?

brawaru (commented):
When I first introduced “heading actions”, it was due to simplicity of the pages and frequency of action you probably want from them. If you look, apps and filter pages [where heading actions currently used] are single label when you don't have any of those and just tables when you have some. One exclusion was to report page, because it's only used by moderators.

Yet when looking at other settings pages — they have many elements which you change one-by-one and it's seems reasonable that you submit them only after going through them. I can agree that such save shortcut is nice to have for experienced users, but I don't agree on removing buttons at the bottom as that will require user which is going through the settings to scroll back to top just to submit changes.

<sub><i>Actually, I think, we need continuous saving on changes without having to press that button and reload page each time, but that's for another issue…</i></sub>

app/views/settings/preferences/appearance/show.html.haml
guigeekz: You're right, I just added again the button at the bottom of the page

app/views/settings/profiles/show.html.haml
guigeekz: Same here, I put back this button

app/views/settings/preferences/other/show.html.haml
guigeekz: Same here, I put back this button

brawaru (approved):
LGTM

Sasha-Sorokin:
Some tests are failing now. Probably because there are now two matching buttons to “Save changes”:

https://github.com/tootsuite/mastodon/blob/6c1ba513ee00443cba684adfe41f62567bd6bb21/spec/features/profile_spec.rb#L42

I'm not experienced in Capybara, but I guess `first('.btn[type=submit]').click` should do it [according to docs](https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Finders#find-instance_method).

guigeekz:
> Some tests are failing now. Probably because there are now two matching buttons to “Save changes”:
> 
> https://github.com/tootsuite/mastodon/blob/6c1ba513ee00443cba684adfe41f62567bd6bb21/spec/features/profile_spec.rb#L42
> 
> I'm not experienced in Capybara, but I guess `first('.btn[type=submit]').click` should do it [according to docs](https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Finders#find-instance_method).

Thanks @Sasha-Sorokin for the tips ! It's fixed.

Gargron (approved):

PR #13084, opened 2020-02-12
Do not suggest to login into the GitLab database

saper:
As pointed out on Discourse:

https://discourse.joinmastodon.org/t/obscure-wtf-error-message-running-migrations-for-3-1/2524

The message the administrators were getting was telling them
to log in to the "GitLab database" and also mentions MySQL


saper:
With https://github.com/tootsuite/mastodon/pull/9877 we have fixed the command, but still other problems with the message remain.

ClearlyClaire (approved):

Issue #1669, opened 2017-04-13
Hide a specific status


Geobert:
Feature to hide one status only (instead of muting a guy for one status)

* * * *

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

voidzero:
Would be happy if the priority could go up. Today I wanted to show Mastodon to someone, and luckily just before I did I opened the federated timeline and saw an image that's, let's say, not family friendly.

I know images can be hidden, but this is not persistent so closing and reopening the timeline just shows it again.

I don't want to mute or block people, but I want to hide statuses for this reason. And I'd like this to be a persistent setting, perhaps in my browser's local cache or something.

It would also be nice if hiding images could be made persistent - I'm not sure if it's ok to just ask for this in this issue, or to create a new issue for it.

magicalskye:
yes, sometimes i get triggered / annoyed by a toot and it just. has to go. but i don't want to silence the whole person because they're cool and they don't know that something innocuous is triggering to me. 

also sometimes the formatting of a toot or an image may be visually distracting and make it hard for me to focus on other things, or may cause visual overload.

so this is an accessibility issue.

wxcafe:
So uh since this has been linked on mastodon and is seeing more attention again, I'm wondering if the "mute conversation" that mastodon has had since 1.3.3 would answer this concern correctly? 
<img width="250" alt="screen shot 2017-06-17 at 20 05 28" src="https://user-images.githubusercontent.com/3009327/27255164-d8715e0a-5398-11e7-832e-0c14f1ca16bf.png">

Cassolotl:
@wxcafe Thank you! :)

Cassolotl:
I wanted to add that the X-Kit extension for Tumblr has this feature, and I use it surprisingly often:

![screen_shot_2018-08-26_at_12_31_10](https://user-images.githubusercontent.com/6607512/44627815-1de4f700-a92c-11e8-9fcd-b4ed94cbe3c3.png)

![screen shot 2018-08-26 at 12 31 24](https://user-images.githubusercontent.com/6607512/44627816-21787e00-a92c-11e8-81df-8df9ce93ab83.png)

bannisdale:
@joyeusenoelle Looks like you can safely start with this one now. ;-)

guigeekz:
@joyeusenoelle have you made any progress on this project? I'd like to work on it, but if there's already a branch that exists it might help. Thanks!

[...]

PR #13090, opened 2020-02-13
[WIP] Hide status

guigeekz:
Related to #1669


app/models/feed.rb
```diff
@@ -28,7 +28,7 @@ def from_redis(limit, max_id, since_id, min_id)
       unhydrated = redis.zrangebyscore(key, "(#{min_id}", '+inf', limit: [0, limit], with_scores: true).map(&:first).map(&:to_i)
     end
 
-    Status.where(id: unhydrated).cache_ids
+    Status.not_hidden(@account).where(id: unhydrated).cache_ids
```
guigeekz: Not sure it's the right place to call the scope **not_hidden**

ThibG:
(Note that it isn't a full review, I've only glanced at the PR so far, I intend to do a full review later on, when it's not marked WiP)

I'm interested in this feature. I was wary of implementing this because of the inability to find the status again if you wanted, and for possible performance reasons.

While the first point doesn't seem to be addressed, this probably isn't an issue anyway. Maybe the status could be known to the client in certain contexts though (e.g., looking at a thread) so it can tell you that a status is here, and tell you that it's only hidden (maybe offer you to unhide it). The downside of that is that clients not answering the introduced API would display the toot anyway in those contexts.

The performance implications seem reasonable.

Glancing at the PR, I'm not completely sure boosts are properly handled though.

guigeekz:
Hi @ThibG,

Thanks for your feedback!

In fact, I followed the specs that were proposed on #1669, and I figured that the action of hiding a status is irreversible for the user. Of course the status will always be present in the database, so if we finally find that the action of unhiding a status is useful, maybe we could incorporate it in a 2nd version of the feature ?

About the API, if we add the action to hide a status as well as the display of the attribute _hide_ on statuses GET calls, for you is it the responsibility of the API client or the back-end not to display/return hidden statuses ?

app/javascript/mastodon/locales/fr.json
```diff
@@ -368,6 +370,7 @@
   "status.embed": "Intégrer",
   "status.favourite": "Ajouter aux favoris",
   "status.filtered": "Filtré",
+  "status.hide": "Cacher le status",
```
Sasha-Sorokin: Changing translation files manually is generally a bad idea (*expect source files in English*). Translations should be added through Crowdin to avoid problems with synchronization and conflicts during translations update.

I once did that mistake where I deleted some strings from all locales and that introduced conflict in next translation update PR by Crowdin.

guigeekz: Ok @Sasha-Sorokin thanks for the feedback, in that case I will remove all french translations

app/controllers/api/v1/statuses/hidden_statuses_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    hide = HiddenStatus.find_or_create_by!(account: current_user.account, status: requested_status)
+
+    hide.status.reload
+  end
+
+  def requested_status
+    Status.find(params[:status_id])
+  end
+end
```
ThibG: The sibling controllers were refactored, it's not *critical*, but you should write it in the same style (set_status hook, etc.) as, say, the BookmarksController.

app/javascript/mastodon/actions/statuses.js
```diff
@@ -107,8 +111,7 @@ export function fetchStatus(id) {
       });
     }).then(() => {
       dispatch(fetchStatusSuccess(skipLoading));
-    }, () => api(getState).get(`/api/v1/statuses/${id}`).then(response => {
-      dispatch(importFetchedStatus(response.data));
+    }, () => api(getState).get(`/api/v1/statuses/${id}`).then(() => {
```
ThibG: Why this change? This doesn't look right to me.

app/javascript/mastodon/reducers/timelines.js
```diff
@@ -18,6 +18,8 @@ import {
 import { Map as ImmutableMap, List as ImmutableList, fromJS } from 'immutable';
 import compareId from '../compare_id';
 
+export const HIDE_SUCCESS = 'HIDE_SUCCESS';
```
ThibG: The symbol should be imported from the actions (see the above imports).

app/javascript/mastodon/locales/fr.json
ThibG: I concur, this is likely to cause merge issues.

app/javascript/mastodon/containers/status_container.js
```diff
@@ -49,6 +50,7 @@ const messages = defineMessages({
   replyConfirm: { id: 'confirmations.reply.confirm', defaultMessage: 'Reply' },
   replyMessage: { id: 'confirmations.reply.message', defaultMessage: 'Replying now will overwrite the message you are currently composing. Are you sure you want to proceed?' },
   blockDomainConfirm: { id: 'confirmations.domain_block.confirm', defaultMessage: 'Hide entire domain' },
+  hideStatusConfirm: { id: 'confirmations.hide_status.confirm', defaultMessage: 'Hide the status' },
```
ThibG: I'd use something shorter, like just “Hide”

app/javascript/mastodon/containers/status_container.js
```diff
@@ -111,6 +113,14 @@ const mapDispatchToProps = (dispatch, { intl }) => ({
     }
   },
 
+  onHide (status, history) {
+    dispatch(openModal('CONFIRM', {
+      message: <FormattedMessage id='confirmations.hide_status.message' defaultMessage='Are you sure to hide this status ?' />,
```
ThibG: It might make sense for the text to explain that a hidden status cannot be brought back.

app/javascript/mastodon/actions/statuses.js
```diff
@@ -188,6 +191,42 @@ export function deleteStatusFail(id, error) {
   };
 };
 
+export function hide(status) {
+  return function (dispatch, getState) {
+    dispatch(hideRequest(status));
+
+    api(getState).post(`/api/v1/statuses/${status.get('id')}/hide`).then(function (response) {
+      dispatch(importFetchedStatus(response.data));
+      dispatch(hideSuccess(status.get('id'), getState().get('statuses')));
```
ThibG: I'm not sure why you'd pass the entire list of statuses in that event.
Looking at the reducer code, you probably want to get the boosts instead (see deleteFromTimelines in app/javascript/mastodon/actions/timelines.js)

app/javascript/mastodon/reducers/timelines.js
```diff
@@ -158,6 +160,8 @@ export default function timelines(state = initialState, action) {
     return filterTimelines(state, action.relationship, action.statuses);
   case ACCOUNT_UNFOLLOW_SUCCESS:
     return filterTimeline('home', state, action.relationship, action.statuses);
+  case HIDE_SUCCESS:
+    return deleteStatus(state, action.id, action.accountId, action.statuses);
```
ThibG: The last parameter of deleteStatus is a list of (statusId, accountId) of toots that are boosts of the target status. It seems like (see comment for the actions file) that you are passing all the known statuses here instead.

app/models/hidden_status.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+#  id         :bigint(8)        not null, primary key
+#  account_id :bigint(8)        not null
+#  status_id  :bigint(8)        not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class HiddenStatus < ApplicationRecord
+  update_index('statuses#status', :status) if Chewy.enabled?
```
ThibG: This is probably not needed, as you don't want a hidden status to be added to the full-text search index for that account.

app/models/feed.rb
```diff
@@ -28,7 +28,11 @@ def from_redis(limit, max_id, since_id, min_id)
       unhydrated = redis.zrangebyscore(key, "(#{min_id}", '+inf', limit: [0, limit], with_scores: true).map(&:first).map(&:to_i)
     end
 
-    Status.where(id: unhydrated).cache_ids
+    Status.send(*status_scope).where(id: unhydrated).cache_ids
+  end
+
+  def status_scope
+    :all
```
ThibG: I think, if we are going to do something like this, it would be cleaner and more understandable to have something along the lines of “status_scope.where(...)” and have status_scope just be “Status”

app/models/home_feed.rb
```diff
@@ -10,4 +10,10 @@ def initialize(account)
   def regenerating?
     redis.exists("account:#{@id}:regeneration")
   end
+
+  protected
+
+  def status_scope
+    [:not_hidden, @account]
+  end
```
ThibG: As said for feed.rb, this here would then be “Status.not_hidden(@account)”

app/models/status.rb
```diff
@@ -92,6 +93,8 @@ class Status < ApplicationRecord
   scope :including_silenced_accounts, -> { left_outer_joins(:account).where.not(accounts: { silenced_at: nil }) }
   scope :not_excluded_by_account, ->(account) { where.not(account_id: account.excluded_from_timeline_account_ids) }
   scope :not_domain_blocked_by_account, ->(account) { account.excluded_from_timeline_domains.blank? ? left_outer_joins(:account) : left_outer_joins(:account).where('accounts.domain IS NULL OR accounts.domain NOT IN (?)', account.excluded_from_timeline_domains) }
+  scope :hidden, ->(account) { joins(:hidden_statuses).where('hidden_statuses.account_id': account.id) }
+  scope :not_hidden, ->(account) { where.not(id: hidden(account)) }
```
ThibG: I have a hard time thinking about the performance implications of that query, tbh.

app/models/status.rb
```diff
@@ -361,6 +368,7 @@ def permitted_for(target_account, account)
         scope = left_outer_joins(:reblog)
 
         scope.where(visibility: visibility)
+             .not_hidden(account)
```
ThibG: This is going to change a possibly expensive request, which we have an index tailored for. We should make sure this doesn't affect performances badly.

app/models/status.rb
```diff
@@ -384,7 +392,7 @@ def apply_timeline_filters(query, account, local_only)
     end
 
     def filter_timeline_for_account(query, account, local_only)
-      query = query.not_excluded_by_account(account)
+      query = query.not_excluded_by_account(account).not_hidden(account)
```
ThibG: Same as above regarding performances.

ClearlyClaire (changes requested):
I finally got to review this, sorry for the wait!

Overall, this is promising. See my inline comments, but overall I have two concerns:
- it seems the UI won't do much with the hidden attribute, and especially not hide toots in threads etc.
- I'm worried about the performance implications on the various timeline queries. This worry might be unfounded, I guess, but we should definitely investigate if there's any regression here. If this is an issue, we may want to do the filtering a posteriori (maybe even just client-side, as the PR already lets the client know whether something is filtered or not). A posteriori filtering might cause pagination issues, but I don't think it's a concern for this feature, which I believe would be used sparingly.

SuperSandro2000:
@guigeekz /unstale

PR #13092, opened 2020-02-14
Add documentation to several undocumented Services

berkes:
I was reading through the code in order to understand some concepts and data-flows.

While doing, I took some notes of the roles of some Services. I've turned those into documentation.


berkes:
This does not change any behaviour and does not touch any code. Just comments.

Anything I can do to help here?

SuperSandro2000:
Bump

PR #13097, opened 2020-02-16
Change config to raise errors on incorrect parameters instead of silently ignoring them

ClearlyClaire:


ThibG:
Hm, we use `permit` throughout the code to slice the params into sub-params, which isn't really how we should do it, I guess.
I guess it should be better to change that, but it's going to take a while and be somewhat error-prone.

app/controllers/api/v1/admin/account_actions_controller.rb
```diff
@@ -17,11 +17,20 @@ def create
   private
 
   def set_account
-    @account = Account.find(params[:account_id])
+    @account = Account.find(filtered_params[:account_id])
+  end
+
+  def filtered_params
+    params.require(:account_id).permit(:type,
```
Gargron: This isn't correct because `require` puts the subsequent code into a nesting, e.g. it will now expect `account_id[type]` to be a thing.

ThibG: Yeah, I have noticed that in the meantime…
Also the pattern of using `permit` and `require` to slice through the params for different uses seem to be encouraged… and at the same time it conflicts with using this configuration, so really idk what we should do here… there seem to be no simple way to raise an issue when given an object instead of a scalar or the opposite…

PR #13113, opened 2020-02-18
Fix NoMethodError in error handler

### LINKED ISSUE #13086: NoMethodError in LinkCrawlWorker
on v3.1.1

```
LinkCrawlWorker	
NoMethodError: undefined method `message' for "Net::ReadTimeout with #<TCPSocket:(closed)>":String

LinkCrawlWorker	
NoMethodError: undefined method `message' for "execution expired":String

LinkCrawlWorker	
NoMethodError: undefined method `message' for "end of file reached":String

LinkCrawlWorker	
NoMethodError: undefined method `message' for "Net::OpenTimeout":String
```

ClearlyClaire:
The aws-sdk-core gem can raise exceptions whose initializer expects
an exception, not a string. The error handling code was ignoring that,
causing the exception's initializer to call `#message` on the string object.

Also, this shortening of backtraces was introduced for the DeliveryWorker,
but affected other codepaths with more complex exception possibilities.

The other “rescue” block remains safe in principle as the exceptions it
can catch are limited to those raised by the http gem, but wrapping it may
be cleaner.

Fixes #13086

### DIFF
--- a/app/lib/request.rb
+++ b/app/lib/request.rb
@@ -73,8 +73,6 @@ class Request
       response.body_with_limit if http_client.persistent?
 
       yield response if block_given?
-    rescue => e
-      raise e.class, e.message, e.backtrace[0]
     ensure
       http_client.close unless http_client.persistent?
     end
--- a/app/workers/activitypub/delivery_worker.rb
+++ b/app/workers/activitypub/delivery_worker.rb
@@ -54,8 +54,9 @@ class ActivityPub::DeliveryWorker
       light.with_threshold(STOPLIGHT_FAILURE_THRESHOLD)
            .with_cool_off_time(STOPLIGHT_COOLDOWN)
            .run
-    rescue Stoplight::Error::RedLight => e
-      raise e.class, e.message, e.backtrace.first(3)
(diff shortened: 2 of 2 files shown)


app/lib/request.rb
```diff
@@ -73,8 +73,6 @@ def perform
       response.body_with_limit if http_client.persistent?
 
       yield response if block_given?
-    rescue => e
-      raise e.class, e.message, e.backtrace[0]
```
zunda: May we leave this block with changing the code like for `delivery_worker.rb` to keep logs less noisy?

```ruby
    rescue => e
      e.set_backtrace(e.backtrace.first(1))
      raise e
```

ThibG:
My fear is that if we do, it may make things harder to debug, since it will apply to *any* Request#perform call, including some that may be quite complex and which failure isn't caused by the Request failing in any way.

The original PR only mentioned DeliveryWorker, do you have any other place in mind causing excessive error logging?

zunda:
Got it! We'll just be able to trim the logs down later if needed.

zunda (approved):

PR #13114, opened 2020-02-18
Disable backtrace silencers

ClearlyClaire:
Backtrace silencers have made things confusing and difficult to debug more than once.


ykzts (approved):
👍

Gargron:
This has the potential of making all stack traces very long

Gargron:
I'm convinced this will be a mistake. Most bugs that we do have, happen in our code and not library code. Some sysadmins are already being mislead by the `set_locale` action wrapping around every error into thinking the error has something to do with locales, and that's our own action. By disabling backtrace silencers, every error will be wrapped in 20 more layers of middleware, increasing confusion. There's also the I/O performance aspect of outputting longer backtraces into logs, or having longer backtraces saved into Sidekiq.

SuperSandro2000:
Bump

PR #13116, opened 2020-02-18
Fix account JSON/RSS not being cacheable due to wrong mime type comparison

ClearlyClaire:
`request.format` is not a symbol but a `Mime::Type`, so the condition actually
never matched, and a session was created even for those requests, preventing
caching.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -9,7 +9,7 @@ class AccountsController < ApplicationController
   before_action :set_cache_headers
   before_action :set_body_classes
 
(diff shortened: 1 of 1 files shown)


ThibG:
Note that, the cache being broken results in weird and interesting behaviors when fetching an actor from an instance that had caching previously working: given sufficient requests to get puma busy, with the provided configuration, nginx will serve stale actor data with `X-Cached: UPDATING`. Since caching is broken, this stale data won't get updated.

Furthermore, we do not expose things like `updated_at` in the AP actor representation, so we can only assume the fetched version is newer, which in this case is false… see #13117

ykzts (approved):

PR #13125, opened 2020-02-21
Fix 404 when undoing a boost/bookmark that doesn't exist

ClearlyClaire:
Returning a 404 when trying to unboost or unbookmark a non-boosted or
non-bookmarked toot is arguably fine, but the Web UI (and possibly other
clients) error out and do not update the state of a status when such an
action fails.

### DIFF
--- a/app/controllers/api/v1/statuses/bookmarks_controller.rb
+++ b/app/controllers/api/v1/statuses/bookmarks_controller.rb
@@ -17,8 +17,8 @@ class Api::V1::Statuses::BookmarksController < Api::BaseController
     @status = requested_status
     @bookmarks_map = { @status.id => false }
 
-    bookmark = Bookmark.find_by!(account: current_user.account, status: @status)
-    bookmark.destroy!
+    bookmark = Bookmark.find_by(account: current_user.account, status: @status)
+    bookmark&.destroy!
 
     render json: @status, serializer: REST::StatusSerializer, relationships: StatusRelationshipsPresenter.new([@status], current_user&.account_id, bookmarks_map: @bookmarks_map)
   end
--- a/app/controllers/api/v1/statuses/reblogs_controller.rb
+++ b/app/controllers/api/v1/statuses/reblogs_controller.rb
@@ -14,12 +14,17 @@ class Api::V1::Statuses::ReblogsController < Api::BaseController
   end
 
   def destroy
-    @status = status_for_destroy.reblog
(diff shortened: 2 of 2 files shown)


app/controllers/api/v1/statuses/reblogs_controller.rb
```diff
@@ -14,12 +14,17 @@ def create
   end
 
   def destroy
-    @status = status_for_destroy.reblog
-    @reblogs_map = { @status.id => false }
+    begin
+      @status = status_for_destroy.reblog
```
Gargron: I don't think this can raise an error. It would just be nil.

app/controllers/api/v1/statuses/reblogs_controller.rb
```diff
@@ -14,12 +14,17 @@ def create
...
+    rescue ActiveRecord::RecordNotFound
+      @status = Status.find(params[:status_id])
+    else
```
Gargron: Also you can't chain `rescue` with `else` unless this is some super wild syntax thing I never encountered

app/controllers/api/v1/statuses/bookmarks_controller.rb
```diff
@@ -17,8 +17,12 @@ def destroy
     @status = requested_status
     @bookmarks_map = { @status.id => false }
 
-    bookmark = Bookmark.find_by!(account: current_user.account, status: @status)
-    bookmark.destroy!
+    begin
+      bookmark = Bookmark.find_by!(account: current_user.account, status: @status)
```
Gargron: It is easier to just use `find_by` without exclamation point and `bookmark&.destroy!`

Gargron (changes requested):

app/controllers/api/v1/statuses/reblogs_controller.rb
ThibG: `status_for_destroy` can

app/controllers/api/v1/statuses/reblogs_controller.rb
ThibG: I am pretty sure you can :thinking:

> The code in an else clause is executed if the code in the body of the begin statement runs to completion without exceptions. If an exception occurs, then the else clause will obviously not be executed. The use of an else clause is not particularly common in Ruby.
http://rubylearning.com/satishtalim/ruby_exceptions.html

ThibG:
Superseded by #13161

PR #13147, opened 2020-02-24
Fix duplicate accounts being created when fetching an account for its key only

### LINKED ISSUE #13136: 1 Account saved with web domain instead of local domain in remote instance
The `toot.kif.rocks` instance is configured with a local domain of  `kif.rocks` according to https://github.com/tootsuite/documentation/blob/archive/Running-Mastodon/Serving_a_different_domain.md This was the case since the first day this instance was online.

It seems one user on our instance has been duplicated on a remote instance but with the web domain instead of the local domain as domain see https://chaos.social/@LenaEine/103708312764209912 and attached discussion. As noted in: There are 2 users in the DB of the remote instance, one with the domain kif.rocks, and one with the domain toot.kif.rocks. Everything else, especially other users from kif.rocks, seem unaffected.

Would anyone have an idea how this might have happened, if a bug from around that time might explain this behaviour, or at least how to debug this? Furthermore, how should the chaos.social administrators best fix this issue? Thanks.

### Expected behaviour

No duplicate users.

### Actual behaviour

A duplicated user.

We have no idea.

### Specifications

kif.rocks is currently on 3.1, should have been on 3.0 around the time this happened. I would assume this applies to chaos.social as well.

ClearlyClaire:
Fixes #13136

When a user's canonical acct domain is different from its id's domain
(WEB_DOMAIN ≠ LOCAL_DOMAIN), two webfinger queries are required to find the
canonical domain from the URI. However, we skip webfinger queries when
updating only the key of a remote user, which led to the creation of a
duplicate account, using the URI's domain instead of the canonical acct: one.

### DIFF
--- a/app/services/activitypub/process_account_service.rb
+++ b/app/services/activitypub/process_account_service.rb
@@ -18,9 +18,10 @@ class ActivityPub::ProcessAccountService < BaseService
 
     RedisLock.acquire(lock_options) do |lock|
       if lock.acquired?
-        @account        = Account.find_remote(@username, @domain)
-        @old_public_key = @account&.public_key
-        @old_protocol   = @account&.protocol
+        @account          = Account.remote.find_by(uri: @uri) if @options[:only_key]
+        @account        ||= Account.find_remote(@username, @domain)
+        @old_public_key   = @account&.public_key
+        @old_protocol     = @account&.protocol
 
         create_account if @account.nil?
         update_account


Gargron (approved):

thegcat:
@Gargron @ThibG Thank you for the fix! Any pointers on how to deal with existing duplicate/erroneous entries in the DB? Should the remote admin just delete the users with the "wrong" domain?

thegcat:
Sorry, meant to ask this in the issue.

PR #13148, opened 2020-02-24
Fix text area above/right of emoji picker being accidentally clickable in web UI

### LINKED ISSUE #13135: Clicking above or right to the emoji button messes up cursor’s position
### Expected behaviour

Trying to click on the emoji button never changes the cursor position, unless you clicked under or to the left of the button in which case it’s pretty obvious you clicked in the textarea.

### Actual behaviour

Trying to click on the emoji button can change the cursor position because you can click above or to the right of the button, miss the button but still click in the textarea, which is confusing: to the user it looks like the click was missed completely, especially since no text could ever reach this place, and will often realize it changes the cursor’s position when succeeding on his/her click and inserting an emoji at the wrong place.

The button probably could have some padding, so that it fills the above and left spaces where you could still hit the textarea.

When trying to apply:

```css
.compose-form .emoji-picker-dropdown {
  top: 0px;
  right: 0px;
}
.emoji-button { padding: 5px };
```

I noticed that `.emoji-picker-dropdown` had two pixels (only on the left of the button) where it doesn’t trigger the button, but doesn’t click into the textarea either, even without my changes, but that’s a bigger problem with a padding.

Click above the emoji or to the right of the emoji button but before the end of the textarea.

### Specifications

Mastodon 3.1.1

ariasuni:
Fix #13135

I put `2px` to the left so that it doesn’t make putting the cursor at the end of the line with the mouse more difficult.

There is no more 2 dead pixels to the left of the emoji button, so trying to click between the textarea and the left of the emoji button either clicks in the textarea or trigger the emoji button, as expected.

### DIFF
--- a/app/javascript/styles/mastodon/components.scss
+++ b/app/javascript/styles/mastodon/components.scss
@@ -388,8 +388,8 @@
 
   .emoji-picker-dropdown {
     position: absolute;
-    top: 5px;
-    right: 5px;
+    top: 0;
+    right: 0;
   }
 
   .compose-form__autosuggest-wrapper {
@@ -4060,10 +4060,7 @@ a.status-card.compact:hover {
 
 .emoji-button {
   display: block;
-  font-size: 24px;
-  line-height: 24px;
-  margin-left: 2px;
-  width: 24px;
+  padding: 5px 5px 2px 2px;
   outline: 0;
   cursor: pointer;
 
@@ -4079,7 +4076,6 @@ a.status-card.compact:hover {
     margin: 0;
     width: 22px;
     height: 22px;
-    margin-top: 2px;
   }
 
   &:hover,


ykzts (approved):

ClearlyClaire (changes requested):
As is, the PR results in:
image

With `top: 0` and `right: 0`, this would look like:
image

The fact that it is so close to the right edge of the box feels a bit jarring, but I guess it's the whole point of the PR.

Maybe instead of a margin, this should be some padding in the `emoji-button` class?

ariasuni:
I took the time to properly test my change and now it’s working as expected. I reduce the padding under the emoji button, I think it looked weird to have more padding on the bottom than on the left.

ClearlyClaire (approved):

PR #13151, opened 2020-02-25
Fix error when searching for URLs that contain the mention syntax

### LINKED ISSUE #13150: Search for "@xyz" 500s
Discovered last October, [discussed with @ashfurrow et al on this small thread on mastodon.technology.](https://mastodon.technology/@snarfed/102973956125353721) Still happens, so I figured I'd file it. Thanks in advance!

### Expected behaviour

Returns search results

### Actual behaviour

HTTP 500 error. I haven't yet stood up a local Mastodon instance of my own to get logs, but that would probably be the next step.

Enter `"@xyz"` in the search box and press enter. You can replace `xyz` with anything, as long as it's at least one character long.

### Specifications

Reproduced on three v3.1.1 instances: mastodon.technology, mastodon.art, and mas.to.

I'm on Firefox 73, Mac OS X 10.15.3, but I doubt anything client side is related.

Thanks in advance for looking!

ClearlyClaire:
Fixes #13150

I fixed it to what I believe was the original intent of the code. Another possibility would be to get the match data from the original regexp to get the account URI to resolve.

### DIFF
--- a/app/services/account_search_service.rb
+++ b/app/services/account_search_service.rb
@@ -171,7 +171,7 @@ class AccountSearchService < BaseService
   end
 
   def username_complete?
-    query.include?('@') && "@#{query}" =~ Account::MENTION_RE
+    query.include?('@') && "@#{query}" =~ /\A#{Account::MENTION_RE}\Z/
   end
 
   def likely_acct?


Gargron (approved):

ThibG:
Anything holding off a merge?

PR #13156, opened 2020-02-26
Fix installation failing when Redis password contains special characters

### LINKED ISSUE #13154: database is not properly seeded
### Expected behaviour

When I try to enable a relay, i get a 404 error.

### Actual behaviour

"method=POST path=/admin/relays/1/enable format=html controller=Admin::RelaysController action=enable status=404 duration=38.96 view=1.49 db=3.02"

### Specifications
v.3.1.1

ClearlyClaire:
Fixes #13154

### DIFF
--- a/lib/mastodon/redis_config.rb
+++ b/lib/mastodon/redis_config.rb
@@ -14,9 +14,9 @@ def setup_redis_env_url(prefix = nil, defaults = true)
   ENV[prefix + 'REDIS_URL'] = if [password, host, port, db].all?(&:nil?)
                                 ENV['REDIS_URL']
                               else
-                                uri = Addressable::URI.parse("redis://#{host}:#{port}/#{db}")
-                                uri.password = password if password.present?
-                                uri.normalize.to_str
+                                Addressable::URI.parse("redis://#{host}:#{port}/#{db}").tap do |uri|
+                                  uri.password = password if password.present?
+                                end.normalize.to_str
                               end
 end
 
--- a/lib/mastodon/redis_config.rb
+++ b/lib/mastodon/redis_config.rb
@@ -14,7 +14,9 @@ def setup_redis_env_url(prefix = nil, defaults = true)
   ENV[prefix + 'REDIS_URL'] = if [password, host, port, db].all?(&:nil?)
                                 ENV['REDIS_URL']
                               else
-                                "redis://#{password.blank? ? '' : ":#{password}@"}#{host}:#{port}/#{db}"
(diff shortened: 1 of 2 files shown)


lib/mastodon/redis_config.rb
```diff
@@ -14,7 +14,9 @@ def setup_redis_env_url(prefix = nil, defaults = true)
   ENV[prefix + 'REDIS_URL'] = if [password, host, port, db].all?(&:nil?)
                                 ENV['REDIS_URL']
                               else
-                                "redis://#{password.blank? ? '' : ":#{password}@"}#{host}:#{port}/#{db}"
+                                uri = Addressable::URI.parse("redis://#{host}:#{port}/#{db}")
```
Gargron: I feel like this would benefit from using `tap` instead of variables.

app/javascript/styles/mastodon/variables.scss
```diff
@@ -12,7 +12,7 @@ $red-bookmark: $warning-red;
 $classic-base-color: #282c37;         // Midnight Express
 $classic-primary-color: #9baec8;      // Echo Blue
 $classic-secondary-color: #d9e1e8;    // Pattens Blue
-$classic-highlight-color: #2b90d9;    // Summer Sky
+$classic-highlight-color: #007ce6;    // Ziqkrr blue
```
nightpool: wrong PR?

ThibG: oops nvm

lib/tasks/mastodon.rake
```diff
@@ -336,7 +336,20 @@ namespace :mastodon do
...
+          if value.is_a?(String) && value =~ /[\s\#\\"]/
+            if value =~ /[']/
+              value = value.to_s.gsub(/[\\"\$]/) { |x| "\\#{x}" }
+              "#{key}=\"#{value}\""
+            else
+              "#{key}='#{value}'"
+            end
+          else
+            "#{key}=#{value}"
+          end
```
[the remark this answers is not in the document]
ThibG: we could, the reason i went against it is that `shellwords.escape` never makes use of single or double quotes, and instead always escape each and every “special” character, which is quickly unreadable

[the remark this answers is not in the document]
ThibG: also, this wouldn't have worked, because dotenv stops at `#` even if it's escaped, when it's not in a single or double-quoted string :weary:

Gargron (approved):

PR #13159, opened 2020-02-26
Code style improvements in JavaScript

berkes:
This fixes errors reported by `yarn run test:lint`.

Three linting errors are not fixed: 
```

xxx/app/javascript/mastodon/actions/announcements.js
  31:3  error  Expected catch() or return  promise/catch-or-return

xxx/app/javascript/mastodon/actions/notifications.js
  152:5  error  Expected catch() or return  promise/catch-or-return

xxx/app/javascript/mastodon/actions/timelines.js
  97:5  error  Expected catch() or return  promise/catch-or-return

✖ 3 problems (3 errors, 0 warnings)
```
I'm not confident enough with the intricacies of our promises here. There are already catch() clauses, but apparently the linter detects other promises that have no catch() clauses. I'm not certain which ones this would be (or if the linter is simply wrong).

### DIFF
--- a/app/javascript/mastodon/features/compose/components/poll_form.js
+++ b/app/javascript/mastodon/features/compose/components/poll_form.js
@@ -155,6 +155,7 @@ class PollForm extends ImmutablePureComponent {
         <div className='poll__footer'>
           <button disabled={options.size >= 4} className='button button-secondary' onClick={this.handleAddOption}><Icon id='plus' /> <FormattedMessage {...messages.add_option} /></button>
 
+          {/* eslint-disable-next-line jsx-a11y/no-onchange */}
           <select value={expiresIn} onChange={this.handleSelectDuration}>
             <option value={300}>{intl.formatMessage(messages.minutes, { number: 5 })}</option>
             <option value={1800}>{intl.formatMessage(messages.minutes, { number: 30 })}</option>
--- a/app/javascript/mastodon/actions/accounts.js
+++ b/app/javascript/mastodon/actions/accounts.js
@@ -106,7 +106,7 @@ export function fetchAccount(id) {
       dispatch,
       getState,
       db.transaction('accounts', 'read').objectStore('accounts').index('id'),
-      id
+      id,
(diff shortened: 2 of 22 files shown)


app/javascript/mastodon/features/compose/components/poll_form.js
```diff
@@ -155,7 +155,7 @@ class PollForm extends ImmutablePureComponent {
         <div className='poll__footer'>
           <button disabled={options.size >= 4} className='button button-secondary' onClick={this.handleAddOption}><Icon id='plus' /> <FormattedMessage {...messages.add_option} /></button>
 
-          <select value={expiresIn} onChange={this.handleSelectDuration}>
+          <select value={expiresIn} onBlur={this.handleSelectDuration}>
```
ThibG: This would introduce a bug (see #12596)

ClearlyClaire (changes requested):
I'm not convinced by the usefulness of this change, but it's ok save for the part I commented inline.

app/javascript/mastodon/features/compose/components/poll_form.js
berkes: I've reverted this, and added a line to configure the linter to ignore this.

berkes:
> I'm not convinced by the usefulness of this change

Currently there is an eslinter configured and set. Either that should go, or it should be honoured, IMO. It makes no sense to have the eslinter wired up, run in tests, but have it fail. That is, arguably, worse than having no linter at all.

Removing the linter would, IMO, be a good alternative if we're not following it anyway. But I'd say, since the amount of errors is small, keeping it in, and keeping running it, is the correct thing to do. 

My plan was to make a PR that enables the js tests in CI, once the linter passes (linting is the first part of running the tests, currently).

ClearlyClaire (approved):
I'm still somehow half-convinced by the need for that PR, but it looks good to me.

We mostly rely on CodeClimate for coding style, I'm not against removing the linter from the test suite, but I have no strong opinion either way.

PR #13164, opened 2020-02-27
Fix announcements with fully-qualified mention to local user crashing WebUI

ClearlyClaire:


Gargron (approved):

mstdn:
Thank you for this! <3

PR #13172, opened 2020-03-01
Add specific rate limits for posting and following

Gargron:
Add an application-level rate limit of 400 follows/follow-requests every 24 hours, and 300 statuses/reblogs every 3 hours. These numbers are adopted from Twitter. Application-level means these rate limits are only used when an action actually succeeds (whereas rack-level rate limits affect even requests that result in failure).

Rate limits checks are implemented at the model level because it's the most ergonomic way to react to creation and rollbacks, but are opt-in through a virtual attribute so that internal business logic would not get rate-limited. The resulting rate limit headers overwrite rack-level ones unless the rack-level ones are closer to being exhausted.

### DIFF
--- a/app/controllers/concerns/rate_limit_headers.rb
+++ b/app/controllers/concerns/rate_limit_headers.rb
@@ -3,6 +3,20 @@
 module RateLimitHeaders
   extend ActiveSupport::Concern
 
+  class_methods do
+    def override_rate_limit_headers(method_name, options = {})
+      around_action(only: method_name, if: :current_account) do |_controller, block|
+        begin
+          block.call
+        ensure
+          rate_limiter = RateLimiter.new(current_account, options)
+          rate_limit_headers = rate_limiter.to_headers
+          response.headers.merge!(rate_limit_headers) unless response.headers['X-RateLimit-Remaining'].present? && rate_limit_headers['X-RateLimit-Remaining'].to_i > response.headers['X-RateLimit-Remaining'].to_i
+        end
+      end
+    end
+  end
+
   included do
     before_action :set_rate_limit_headers, if: :rate_limited_request?
   end
@@ -44,7 +58,7 @@ module RateLimitHeaders
   end
 
   def api_throttle_data
-    most_limited_type, = request.env['rack.attack.throttle_data'].min_by { |_, v| v[:limit] }
+    most_limited_type, = request.env['rack.attack.throttle_data'].min_by { |_, v| v[:limit] - v[:count] }
     request.env['rack.attack.throttle_data'][most_limited_type]
   end
 
--- a/app/models/follow.rb
+++ b/app/models/follow.rb
@@ -15,6 +15,9 @@
 class Follow < ApplicationRecord
   include Paginable
   include RelationshipCacheable
+  include RateLimitable
+
+  rate_limit by: :account, family: :follows
 
   belongs_to :account
   belongs_to :target_account, class_name: 'Account'
--- a/app/models/status.rb
+++ b/app/models/status.rb
@@ -32,6 +32,9 @@ class Status < ApplicationRecord
   include Paginable
   include Cacheable
   include StatusThreadingConcern
+  include RateLimitable
+
+  rate_limit by: :account, family: :statuses
 
   self.discard_column = :deleted_at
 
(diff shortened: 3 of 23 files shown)


SuperSandro2000:
Is there a way to bypass this for admins or configure it higher?

Gargron:
@SuperSandro2000 We can discuss the numbers, for now I took the same limits and periods as Twitter. In what circumstance would you need to post more than 300 things in 3 hours? Most people don't post that much in day

SuperSandro2000:
I was thinking about bots who might send private messages to people and they could easily it 300 in 3 hours.

app/models/follow.rb
```diff
@@ -15,6 +15,9 @@
 class Follow < ApplicationRecord
   include Paginable
   include RelationshipCacheable
+  include RateLimitable
+
+  rate_limit by: :account, family: :follows
```
ThibG: This is somewhat difficult to reason about. From what I gather, this means if you try to create more `Follow` objects than the limit with the same `account`, this will raise an error. However, I can see at least two issues with that:
- an accepted follow request counts twice, afaict? once for the request, once for the accept
- the creation of such a `Follow` object might not be the direct result of an action from the rate-limited party: when accepting a follow request, you could get a confusing error because the potential follower is rate-limited

app/models/status.rb
```diff
@@ -32,6 +32,9 @@ class Status < ApplicationRecord
   include Paginable
   include Cacheable
   include StatusThreadingConcern
+  include RateLimitable
+
+  rate_limit by: :account, family: :statuses
```
ThibG: From what I understand, any scheduled status for an account that reached its limit will error out (in the publish worker). I am not sure this is how it *should* work.

app/models/follow.rb
ThibG: Also, it might cause the `Import::RelationshipWorker` and possibly other workers to fail, in a way that will push them back on the queue. Not sure it is what should happen.

app/controllers/concerns/rate_limit_headers.rb
```diff
@@ -3,6 +3,19 @@
 module RateLimitHeaders
   extend ActiveSupport::Concern
 
+  class_methods do
+    def override_rate_limit(method_name, options = {})
```
ThibG: maybe call that `override_rate_limit_headers`?

ClearlyClaire (changes requested):

halcy:
Do these apply on top of the previous rate limits or do they replace them? And for the headers, does this mean if I do a request to two different endpoints and my code is not aware that they have different rate limits, my tracking of the limits will break?

Gargron:
> 
> 
> Do these apply on top of the previous rate limits or do they replace them? And for the headers, does this mean if I do a request to two different endpoints and my code is not aware that they have different rate limits, my tracking of the limits will break?

Rate limits stack, the one with the fewer remaining requests takes precedence and is returned in the headers. Before this PR, stacking rate limits would be: API usage from an IP address (big), API usage of a specific authenticated user (smaller), API usage of the media upload endpoint (small).

ClearlyClaire (changes requested):
Seems ok to me, although still a bit difficult to reason about.

Also, is `app/services/concerns/rate_limitable.rb` a leftover?

Gargron:
Does it mean you would prefer rack-attack definitions over this?

ThibG:
hmmm… rack-attack would be more readable, I think, but it doesn't know as much about successful creation of statuses/follows… the requested changes were more about the empty (leftover?) `app/services/concerns/rate_limitable.rb`

Gargron:
Removed that file. Also moved the media upload rate limit to the new system so it only rate-limits successful uploads... Old way would rate limit even when there was some kind of issue uploading...

SuperSandro2000:
But that would allow to DOS Mastodon with not successful requests.

ThibG:
I agree with @SuperSandro2000, media that fails to transcode after some time (e.g., hitting a timeout, but it could be any error) can still consume resources, and moving the limit here would allow an attacker to send as many such file as they want.

Gargron:
Hm. The original intent of the rate limit was limiting the rate at which *disk space* is consumed, not necessarily being a DoS protection. Mostly I am bothered by how these API rate limits are now in different places and how hacky it is to make rack-attack limits apply to the right API method, I feel like there's room for error there and lots of repetition. But you have a point :confused:

ClearlyClaire (approved):

PR #13175, opened 2020-03-01
Refactor/cleanup TIMELINE_DELETE-related code

ClearlyClaire:


app/javascript/mastodon/actions/timelines.js
```diff
@@ -42,7 +42,7 @@ export function updateTimeline(timeline, status, accept) {
 export function deleteFromTimelines(id) {
   return (dispatch, getState) => {
     const accountId  = getState().getIn(['statuses', id, 'account']);
-    const references = getState().get('statuses').filter(status => status.get('reblog') === id).map(status => [status.get('id'), status.get('account')]);
+    const references = getState().get('statuses').filter(status => status.get('reblog') === id).map(status => status.get('id'));
```
Gargron: Isn't the information about the account used to remove deleted statuses from profile views?

ThibG: It doesn't appear to be the case. It is passed to `deleteStatus` in https://github.com/tootsuite/mastodon/blob/9cbbc50fcdfe2ec852107b1757d92d54b862a91a/app/javascript/mastodon/reducers/timelines.js#L92-L106 but it doesn't do anything with the `accountId` param…

Gargron (approved):

PR #13178, opened 2020-03-01
Change the string "Hide everything from …" to "Block domain …" in web UI

ClearlyClaire:
Blocking a domain is closer to blocking all its users than to a mute action.

Indeed, in addition to hiding every toot and notification from that domain, it breaks and prevents follow relationships. When `AUTHORIZED_FETCH` is enabled, it also does not let the blocked server fetch new toots from the blocking user.

It is not *exactly* like current user blocks, though, as it does not notify the blocked instance in any way (except for breaking following relationships).

Another possibility might be to use “Block server …” and “Unblock server …” instead of “Block domain …” and “Unblock domain …”.


trwnh:
i think "domain" is more correct than "server" since the block remains in effect even if the server goes down.

semi-related, i'd also like to see a "mute domain" option in the future for domains like botsin.space which i would want to hide but not fully block, like a user-level silence.

joyeusenoelle:
My perspective: The key difference between "domain" and "server" to me is whether servers are uniquely identified to each other *aside from* the domain name.

For example, let's say someone on mastodon.social decides to block my Mastodon instance at elekk.xyz. After that, I shut down my Mastodon instance and allow the domain registration to lapse. Someone else purchases the elekk.xyz domain and sets up a new Mastodon instance using that domain name. Will the mastodon.social user's block be in effect for the new Mastodon instance?

If the answer is yes, then keep "domain". If not, then change it to "server".

trwnh:
yes, the API accepts a string and will filter for that string

Gargron (approved):

